PR #1270, opened 2015-07-16
Fix change flag in edit account; Fix #1269

### LINKED ISSUE #1269: Account Settings
Just updated cockatrice yesterday and no matter what I do I cannot set my country/flag. On OSX 10.10.4.....It seems like I can only set the flag if I also set my gender pronouns to something other than neutral.

ctrlaltca:


ctrlaltca:
This is so trivial..

PR #1286, opened 2015-07-22
Prefer enabled sets when loading images

### LINKED ISSUE #1235: Disabled sets should be lowest priority for image downloads
This should be automatic without user intervention

ref #1072 #1231

ctrlaltca:
fix #1235

This PR changes the behavior used to determine the order sets will be used to load/download card images.
 * before: we used set order, making no distinction if sets were enabled or disabled;
 * after: we first use only enabled sets following their order, and then disabled sets following their order.

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -92,7 +92,7 @@ void CardSet::setIsKnown(bool _isknown)
     settings.setValue("isknown", isknown);
 }
 
-class SetList::CompareFunctor {
+class SetList::KeyCompareFunctor {
 public:
     inline bool operator()(CardSet *a, CardSet *b) const
     {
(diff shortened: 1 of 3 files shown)


ZeldaZach:
This seems to work as expected.
My only complaint is that the image cache is not flushed instantly when I disable/enable a set. I disabled MMA and the cards still showed the MMA image until I restarted the client.

ctrlaltca:
> My only complaint is that the image cache is not flushed instantly when I disable/enable a set

I can try to fix it, hang on

ctrlaltca:
@ZeldaZach can you test it again after the last commit?

ZeldaZach:
Yup, nice job!

Daenyth:
Awesome feature!

PR #1288, opened 2015-07-22
Hack to make travis always detect the installed qt5 version

ctrlaltca:
instead of having to update it manually every time homebrew updates its current version


ctrlaltca:
merging it since it fixes compilation on travis (osx+qt5)

PR #1293, opened 2015-07-24
Force black text color on deck list

### LINKED ISSUE #1289: Deck list is hard to read with dark UI theme
I use a GTK theme which has a dark background and light text. This, combined with the white, green, and gray coloring in the deck editor makes the deck list extremely hard to read.

One possible solution would be to ensure that the text in the deck list is always black, rather than just assuming that the default will be readable.

Another option would be to make the deck list colors change with the theme.

Screenshot:

ctrlaltca:
fix #1289

### DIFF
--- a/cockatrice/src/decklistmodel.cpp
+++ b/cockatrice/src/decklistmodel.cpp
@@ -108,6 +108,9 @@ QVariant DeckListModel::data(const QModelIndex &index, int role) const
                 int color = 90 + 60 * node->depth();
                 return QBrush(QColor(color, 255, color));
             }
+            case Qt::ForegroundRole: {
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Seems fine to me, but I'd like @skwerlman to possibly test this before we merge it

ctrlaltca:
I'd like @skwerlman to test this if possible, too.
I tested it with Xfce 4.12 using a dark theme, here's a screenshot:
black

ZeldaZach:
Based on your test I'm fine with merging this. I didn't have a dark-end to test with so it was awkward for me.

skwerlman:
Looks good!

Issue #453, opened 2014-11-21
Allow users to resize board states


ZeldaZach:
When you play in a multiplayer (3+ players) game, the board state gets very small and it's hard to see all of your cards. I'd like to propose a new layout option that would always give you (when you're the player) the largest client-sided screen so you can see your cards the best.

This would also allow your cards to appear bigger in hand, which I think is a plus.

woogerboy21:
Do we actually have the ability to do proper scaling with the current client?  So say a section is resized the scales of the various objects and images adjust accordingly?  @poixen may be more familiar as he has done a number of UI things.

ZeldaZach:
I believe the client has a scale system built into it already, just it needs a new definition of a layout for where to place each window

ZeldaZach:
We should for sure allow players to scale board states. If we keep aspect ratio the same, it should be trivial

PR #1294, opened 2015-07-24
tablezone refactor and cleanup

mildmongrel:
The tablezone code is pretty difficult to follow in its current state.  Looking ahead to features like #453 I think it made sense to tend to it a bit - more comments, more named constants, fewer magic numbers, refactored the large reorganizeCards() method, etc.

*Functionality should be unchanged* with one small exception:  The previous code was a little liberal with the amount of empty space being allocated to the right of the last card, so I reduced that.  That way the table area stays a little more compact and cards will stay just a little bit larger than before after scaling.

Tested by playing local games (Linux qt4) and by giving test builds to friends (Windows qt5) and playing some network games.  No reports of any unusual behavior. 

This code is pretty fundamental; please code review.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2389,6 +2389,7 @@ void Player::setMirrored(bool _mirrored)
 
 void Player::processSceneSizeChange(int newPlayerWidth)
 {
+    // Extend table (and hand, if horizontal) to accomodate the new player width.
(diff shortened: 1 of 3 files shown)


ctrlaltca:
Code seems fine, and works as expected :+1:

mildmongrel:
Thanks for reviewing and testing!

ZeldaZach:
Nice job. Seems to be fully functional :shipit:

Issue #344, opened 2014-10-02
"Save password" feature stores in clear text


woogerboy21:
When you use the "save password" check box the password for your account is stored in plain text.
Its a security risk.

tooomm:
For completeness (and because @woogerboy21 asked today), the passwords are stored in `servers.ini`

Vafthrudnir:
How about not storing the password at all, but a random hash instead? Kinda like how a "Remember me" cookie works:
The first time a user chooses the "Save password" option during login, after his/her credentials are verified the server could generate a random hash (token), and send it back to the client. This would be stored by the server among the user data, and by the client in the settings.
On subsequent logins, the server checks whether the user sent the password or the token. If either matches, the login is successful. If "Save password" was checked, a new token is generated, stored and sent back, otherwise it gets cleared on the server side. If the login failed due to wrong credentials, the token would also gets to be cleared.
This way the token would be either be updated or cleared with every login attempt. Do you think this solution would be sufficiently secure?

Daenyth:
I'll have to think about it. On first blush it seems reasonable. Can you
point to any existing security literature about this method?

Vafthrudnir:
After some research I found [this link](http://jaspan.com/improved_persistent_login_cookie_best_practice), which discusses an advanced version of the method I described, and also references the original author.

tooomm:
I didn't study your link @Vafthrudnir, but just from our explanation it sounds possible to simply copy the token from one's settings file and paste it into your own to log in.
So instead the pw in clear text, you would "spy" his token in clear text to login successfully...

It should still be an improvement, because you would realize on next login that your token isn't valid anymore and something is wrong. But that might be too late.
While it limits damage it doesn't fix the problem itself completely.

Vafthrudnir:
I haven't checked our login implementation yet, but if a unique client id is sent and stored on the server right now, then it would be quite easy to bind it to the token, and use the username-token-clientID-seriesID quartet on logins. That would add an additional layer of protection.
Although this isn't circumventable either: if someone modifies their own client to show the victims ID after stealing their token, or uses the same computer as the victim, there will be no way to detect it. Basically the whole feature is a security risk, and after a point there really isn't much we can do to protect the user.

skwerlman:
it's not like there's a huge incentive to steal cockatrice tokens, so as long as we make it difficult, it's probably good enough.

[...]

PR #1307, opened 2015-07-27
Previous server connection history

poixen:
newlogin

+ able to select previous servers from a list to connect to.
+ remove items from the drop down by selecting them and pressing delete.
+ new hosts will be added to the previous hosts.
+ remembers the last host connected to in the dropdown.
+ default hosts are "cockatrice.woogerworks.com", "vps.poixen.com" and "chickatrice.net"

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -1,10 +1,14 @@
 #include <QSettings>
 #include <QLabel>
 #include <QCheckBox>
+#include <QComboBox>
+#include <QRadioButton>
 #include <QGridLayout>
 #include <QHBoxLayout>
 #include <QDialogButtonBox>
 #include <QDebug>
+#include <QEvent>
+#include <QKeyEvent>
 #include <iostream>
 #include "dlg_connect.h"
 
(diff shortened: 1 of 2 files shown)


Daenyth:
Instead of previous host/ new host radio buttons, could we do a dropdown with an editable field for adding new items? Does Qt have such a widget?

Do we want to include chickatrice?

Daenyth:
How does password saving work with the host dropdown btw?

ZeldaZach:
I have a feeling that passwords are kept in plaintext if they're saved with the dropdowns.

Daenyth:
They've always been kept in plaintext, see #344

ZeldaZach:
Is that a bad thing?

Daenyth:
Arguably yes or no. Details are in that ticket.

ctrlaltca:
Do we want per-server user profile or just multiple server entries?

ZeldaZach:
Do you mean like if I have two accounts I'd have two options in the drop down menu?

ctrlaltca:
I meant: do we just want to save a list of recent servers, or do we want to have a list of profiles, where each profile has a different server name / user name / password saved?

ZeldaZach:
I like the idea of having profiles

poixen:
@Daenyth that was the first thing I thought of too. Whilst you can have an editable dropdown, it makes all of the items editable - its not good. I couldnt find anything else, so this was the next best thing.

Profiles would be great! If someone wants to expand on this they can. I wont be adding more to this PR  :/

poixen:
Oh and I though about adding chickatrice, but I couldnt connect. Are they reg only?

MassCraziness:
Yes, they are reg-only @poixen

poixen:
Added chickatrice. ready to be tested.

Daenyth:
Profiles were what I had in mind. Users should use different passwords on
different servers, and this encourages you to send your password for server
A to server B.

Daenyth:
poixen if you don't want to write that I'll try to find some time to work
on it myself, maybe this weekend or something

poixen:
@Daenyth go for it. You can use this as a base if you want.

poixen:
@Daenyth any updates on this?

Daenyth:
Haven't had time to work on it.

cockatrice/src/dlg_connect.cpp
```diff
@@ -109,3 +173,17 @@ void DlgConnect::actCancel()
     
     reject();
 }
+
+
+bool DeleteHighlightedItemWhenShiftDelPressedEventFilter::eventFilter(QObject *obj, QEvent *event)
```
Daenyth: That is quite the class name!

poixen: Ha! Forgot about that. It's some net code I found.

tooomm:
@Lachee any knews about this? Just saw it on your chickatrice.net site:
chickatrice net adress

Makes no sense to implement a dead address. ;)

Lachee:
I forgot I placed that. My domain seller.triednto scare me by saying the chickatrice.net was about to expire on the 3rd of August, but obviously it didn't. As I am a tad broke at the moment, I wasn't sure if I should continue the domain or buy voidpixel.com.an. 

I since learnt that the domain doesn't expire until the 13 of October or something like that, so I decided to put it of for a while and forgot to take down the messages.

I shall do that today, thank you for reminding me ;)

Cheers, Lachee

Date: Wed, 5 Aug 2015 07:39:54 -0700

ZeldaZach:
Is anything being done with this currently?

poixen:
This PR is good to go. Not sure if @Daenyth is working on another PR based on it.

ZeldaZach:
I'll let @Daenyth merge it if he wants then

Daenyth:
:shipit:, I'm not going to have time for a little bit and I don't want to hold this up. I'll get my enhancements in later.

PR #1310, opened 2015-07-28
Servatrice: do not disclose user emails to clients

ctrlaltca:
Currently servatrice sends the user email to all connected clients on join events.
This PR removes the the email address so that it's not disclosed to users.


Daenyth:
:+1:

PR #1312, opened 2015-07-29
[WIP] Servatrice Settings Interface 

woogerboy21:
So I am stumped on this one and could use some help from the higher powered coders that be.  Im attempting to make a new interface to the servatrice server to allow the lookup of the settings from the settings cache that is used by servatrice.  Currently the only way to pass the settings back and forth between the server and servatrice is either passing it as a static setting on startup or tunnel it through the database interface.  The database interface works but its ugly and just adds more confusion on top of what is already confusing to follow code.

Now I am just plain stumped on this one and could use some code review from others to tell me what I am missing / not understanding :disappointed:


woogerboy21:
im going to close this PR and maybe pick it up at a later time.  Possibly just throw an issue / recommendation out for someone that may have time and be more knowledgeable about it other than me.

Issue #145, opened 2014-06-28
Client timeout value change


woogerboy21:
Currently the client has a hard coded value for the maxTimeout value it expects to receive from the server during it's client->server keep a-lives.  The value is hard coded into the file:
> cockatrice/src/remoteclient.h

It would be nice to move that configuration option out of the header file and into the client configuration settings some were so users can adjust the setting to allow for slower speed connections.

woogerboy21:
@Daenyth So while adding the functionality by reading the setting yet not exposing it I may have come up with a concern on allowing the value to be exposed.  Since the client has no understanding of the settings on the server, if a client adjusts the ping value to a value higher than what the servers operator has set the user time out value, the users client wont check in as often as it would need to and the user would experience constant "server time outs" when the underlying issue is really the user is just contacting the server at to slow of an interval.  After #1313 is worked out, changing the polling interval will be real easy and if we ant to make a UI setting to adjust it thats fine, just as long as there is some type of warning or "hey if you do this , you could cause problems on your end" kind of warning.

Daenyth:
We could add a new field to the server handshake so that the server can inform the client of timeout settings perhaps

woogerboy21:
@Daenyth Should we consider this issue fixed with #1313 or should we keep it open due to the concern mentioned at https://github.com/Cockatrice/Cockatrice/issues/145#issuecomment-126491315 ?

Daenyth:
Can you file a specific ticket about that concern and close this one?

woogerboy21:
Well since the exposed UI setting isn't there and its mentioned that it can be added later (if need be) do we still want to open a new issue on a possibility that could potentially exist assuming the software gets updated to support it?  I don't foresee the UI change occurring anytime soon and not certain anyone is really to interested in doing it.  For me, the client ID communication time has been changed and the value moved to the settings cache which was the primary goal(s).

PR #1313, opened 2015-07-30
Update Client & Server ping time

woogerboy21:
Updated the client and server ping time periods from 1 second to 5 seconds.
Also changed the hard set value(s) for the timers to variables that are pulled from the settings cache respectively.  

One thing to note.  The client max ping time out is calculated based on the ping value in side the client settings cache were as the servers max time out value is not.  Its pulled from the servers config since that was there to begin with.  Should I update the value to calculate based on the operators default ping timer interval or is it safe to assume the operator can figure out that if they set the value of ping to be greater than the value of max it will cause time outs?

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -1,7 +1,7 @@
 #include <QTimer>
 #include <QThread>
 #include "remoteclient.h"
-
+#include "settingscache.h"
 #include "pending_command.h"
 #include "pb/commands.pb.h"
 #include "pb/session_commands.pb.h"
@@ -16,8 +16,10 @@ static const unsigned int protocolVersion = 14;
 RemoteClient::RemoteClient(QObject *parent)
     : AbstractClient(parent), timeRunning(0), lastDataReceived(0), messageInProgress(false), handshakeStarted(false), messageLength(0)
 {
+
+    int keepalive = settingsCache->getKeepAlive();
     timer = new QTimer(this);
-    timer->setInterval(1000);
+    timer->setInterval(keepalive * 1000);
     connect(timer, SIGNAL(timeout()), this, SLOT(ping()));
 
     socket = new QTcpSocket(this);
@@ -102,11 +104,11 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
 void RemoteClient::doLogin()
 {
     setStatus(StatusLoggingIn);
-    
+
     Command_Login cmdLogin;
     cmdLogin.set_user_name(userName.toStdString());
     cmdLogin.set_password(password.toStdString());
-    
+
     PendingCommand *pend = prepareSessionCommand(cmdLogin);
     connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(loginResponse(Response)));
     sendCommand(pend);
@@ -123,12 +125,12 @@ void RemoteClient::loginResponse(const Response &response)
     if (response.response_code() == Response::RespOk) {
         setStatus(StatusLoggedIn);
         emit userInfoChanged(resp.user_info());
-        
+
         QList<ServerInfo_User> buddyList;
         for (int i = resp.buddy_list_size() - 1; i >= 0; --i)
             buddyList.append(resp.buddy_list(i));
         emit buddyListReceived(buddyList);
-        
+
         QList<ServerInfo_User> ignoreList;
         for (int i = resp.ignore_list_size() - 1; i >= 0; --i)
             ignoreList.append(resp.ignore_list(i));
@@ -177,7 +179,7 @@ void RemoteClient::readData()
     QByteArray data = socket->readAll();
 
     inputBuffer.append(data);
-    
+
     do {
         if (!messageInProgress) {
             if (inputBuffer.size() >= 4) {
@@ -202,7 +204,7 @@ void RemoteClient::readData()
         }
         if (inputBuffer.size() < messageLength)
             return;
-        
+
         ServerMessage newServerMessage;
         newServerMessage.ParseFromArray(inputBuffer.data(), messageLength);
 #ifdef QT_DEBUG
@@ -210,9 +212,9 @@ void RemoteClient::readData()
 #endif
         inputBuffer.remove(0, messageLength);
         messageInProgress = false;
-        
+
         processProtocolItem(newServerMessage);
-    
+
         if (getStatus() == StatusDisconnecting) // use thread-safe getter
             doDisconnectFromServer();
     } while (!inputBuffer.isEmpty());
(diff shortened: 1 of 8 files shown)


woogerboy21:
Also , one thing to note.  With this change it should enable the ability for some one with greater UI skills than I to add something in the UI to adjust the value, kill the existing timer and restart a new one which would fix #145

ctrlaltca:
Clientside, it seems ok to me;
Serverside, i see some problems:
pingClockTimeout() was called every second and used to
 * disconnect the user if lastDataReceived > getMaxPlayerInactivityTime();
 * update the antiflood counters;
So, the "force disconnect of idle user"functionality is already there, and users that didn't send any message (not even a ping) for the last max_player_inactivity_time seconds (default to 15) are disconnected.
Adding a "clientkeepalive" (default 5) multiplier to the pingClockTimeout() timer will:
 * make the max_player_inactivity_time check happen every 5 seconds instead of very second; in the worst case the value will be stretched to 19 seconds (15+5-1); this is not a big problem; 
 * break the messageSizeOverTime, messageCountOverTime and commandCountOverTime antiflood counters. These are simple arrays with one integer per second, where each integer represents the number of commands received in that second, so the sum() of the array is the number of messages in the considered period. Every second an element=0 is prepended and the last one dropped; at every received command the current element is incremented by 0

woogerboy21:
> make the max_player_inactivity_time check happen every 5 seconds instead of very second; in the worst case the value will be stretched to 19 seconds (15+5-1); this is not a big problem;

This was some what addressed in my post about should we assume is safe to believe the operator is smart enough to set the max inactivity time accordingly.  If not I can change the calculations similar to the client side and make the max setting a calculation of the number of seconds defined rather than expecting the operator to calculate things on there own.

> break the messageSizeOverTime, messageCountOverTime and commandCountOverTime antiflood counters. These are simple arrays with one integer per second, where each integer represents the number of commands received in that second, so the sum() of the array is the number of messages in the considered period. Every second an element=0 is prepended and the last one dropped; at every received command the current element is incremented by 0

This one is a bit more tricky and will have to take a look when I have a moment.  First couple thoughts that come to my head is we can either implement a new timer for the flood control (not a fan of the idea) Or we could tinker with the amount that the arrays are incremented by given the time interval set by dividing ... maybe?

Thoughts?

ctrlaltca:
I can see a benefit for the client-size changes of reducing network traffic while keeping lag checking working.
What's the benefit of the server side changes, if we need to keep a timer firing every second to keep for antiflood anyway? I only see a solution in changing the antiflood to check intervals of clientkeepalive seconds.

woogerboy21:
I haven't had a chance to dig into what the pingTimeOut function is doing on the server side.  My thoughts were to try and reduce the over all work the server is constantly doing every second if we can assuming there is a significant performance increase in doing so.  Seeing as there would have to be a timer in place anyway (regardless of how long it takes the timer to fire) , is there anything in the function being called when the timer fires that is causing a significant performance impact?

ctrlaltca:
I need to check that.. profiling.

woogerboy21:
One thing to mention, if the flood control is a concern it still seems to be functioning at an acceptable level on the testing server with this change rolled in.

woogerboy21:
Wait.... the counting interval for the flood control is based on :
```
messageCountingInterval = settingsCache->value("security/message_counting_interval", 10).toInt();
```
So all that is happening is the ping time out counter fires and the function performs this:
```
void Server_ProtocolHandler::pingClockTimeout()
{
    int interval = server->getMessageCountingInterval();
    if (interval > 0) {
        messageSizeOverTime.prepend(0);
        if (messageSizeOverTime.size() > server->getMessageCountingInterval())
            messageSizeOverTime.removeLast();
        messageCountOverTime.prepend(0);
        if (messageCountOverTime.size() > server->getMessageCountingInterval())
            messageCountOverTime.removeLast();
    }

    interval = server->getCommandCountingInterval();
    if (interval > 0) {
        commandCountOverTime.prepend(0);
        if (commandCountOverTime.size() > server->getCommandCountingInterval())
            commandCountOverTime.removeLast();
    }

        prepareDestroy();
    ++timeRunning;
}

```

woogerboy21:
Lets give some examples (and feel free to correct me  if I am wrong).  So if you change the value of the timer from 1 second to 5, every 5 seconds the server will make sure the user has not exceeded the the flood control configuration with the exception of the first timer fire. Since the timer technically has to go twice for the arrays to be incremented and the second time is when its caught.  Seems ok to me...  but lets take an extreme example....


Lets change the timer value to 60 seconds.  Assuming the operator sets the max time out value accordingly so the users dont immediately time out.  Users could in theory flood the server for the entire minute before the flood control would kick in and stop it correct?  Is that the concern?

ctrlaltca:
Let's take the "message count per period" antiflood as an example. The default values are:
```
; Servatrice can avoid users from flooding rooms with large number of messages in an interval of time.
; This setting defines the length in seconds of the considered interval; default is 10
message_counting_interval=10

; Maximum number of messages in an interval before new messages gets dropped; default is 10
max_message_count_per_interval=10
```
So servatrice will drop messages if the user sends more than 10 messages in 10 seconds.
How does this work? Servatrice creates an array with 10 (message_counting_interval) items named messageCountOverTime, where each cell represents the number of messages sent by the user in that second:
```
[0, aka this sec][1 sec ago][2 sec ago] ... [9 sec ago]
```
Every time servatrice receives a message, it will increment the value stored in the first cell, aka  messageCountOverTime[0]++.
If servatrice receives 4 messages in this seconds, messageCountOverTime[0]=4;
Every time pingClockTimeout() triggers, a new empty cell ( = 0 ) gets prepended at the beginning, and the last one, aka messageCountOverTime[10], gets removed.
We can then sum() all the cells of the array to get the total number of messages in the last 10 seconds, and compare it to max_message_count_per_interval to decide if the antiflood must trigger.

If pingClockTimeout() triggers every 5 seconds, each cell in messageCountOverTime will represent the number of messages received in 5 seconds, so the entire array will represent the last 50 seconds.
The antiflood will then trigger like if we configured it for a max of 10 messages in 50 seconds.

woogerboy21:
Correct.  So when the flood does happen the user has to wait even longer to be allowed to make more messages.  So currently the user may flood the room and have to wait a possibility of 10 seconds before being able to retype things were as now its a power of time.  So for every additional second the user would have to wait 10 more seconds.
> I am assuming it will increase the amount of time a user is presented with the flood control message.
^Hense that statement (which maybe I wasnt clear when I made it

woogerboy21:
@ctrlaltca How's that?

woogerboy21:
Just an FYI.  I rolled this in on the woogerworks main cockatrice server and adjusted the server side time value to 5 seconds.  Clients will still be checking in every second but the server should be doing less polling checks.

PR #1324, opened 2015-08-03
Mod Clear Chat Globally

### LINKED ISSUE #506: Moderator Feature: Clear Chat
When we have users who do bad things, such as some of the guests who go on racial rampages, there should be a mod function to clear the main chat.

EDIT: Also, maybe a way to turn the chat off temporarily for certain ranks
- [ ] Disable for Guests
- [ ] Disable for Registered

ZeldaZach:
Fix #506.

This PR allows mods to clear the room's chat for all users in the room.
Server needs to be rebuilt in order for this change to work, as well as new clients.

screenshot 2015-08-02 21 42 15
screenshot 2015-08-02 21 42 17
screenshot 2015-08-02 21 42 20

I mainly did this as a learning experience, but I figured I'd throw it up here to see others' opinions

### DIFF
--- a/cockatrice/src/tab_room.cpp
+++ b/cockatrice/src/tab_room.cpp
@@ -28,6 +28,7 @@
 #include "pb/event_join_room.pb.h"
 #include "pb/event_leave_room.pb.h"
 #include "pb/event_room_say.pb.h"
+#include "pb/event_room_clear.pb.h"
 #include "pending_command.h"
 #include "dlg_settings.h"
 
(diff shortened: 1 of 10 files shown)


MassCraziness:
I really dislike this. A moderator should not have the power to erase chat messages for other people. Some may want to have the messages so they can continue a conversation. Just because someone interrupts it with something doesn't mean a mod should just wipe the whole chat out.

I do not like this at all.

ZeldaZach:
@MassCraziness Sorry you don't like it. I did it as a learning experience to figure out how to get protobuff calls to work initially and then went forward and finished it.

I'm curious what others do think, though.

poixen:
Would be nicer to remove individual lines. I can't see this going down well with the community.

Daenyth:
Yeah, I don't support this feature.

What's the rationale?

woogerboy21:
@Daenyth : the only real good reason I could see something like this being useful is if some one goes off spewing things in a chat room.  I've seen in the past users spew banners of hateful things.

Daenyth:
I'd rather just remove the person who did it. Anyone there will have seen
it, regardless of whether it's cleared. People who haven't joined yet won't
see it either way.

PR #1325, opened 2015-08-03
Cards Download in Appropriate Language

### LINKED ISSUE #841: allow to define a language for card picture downloads
A brazilian guy told me this... he said that this would be a very huge improvement since many people struggle with english in general (true for other languages/countries as well!). If they get the card pictures and/or text in the language of their choice that would help many foreign players and strengthens the international cockatrice community in general.
Everybody sees the pictures and/or text in their preferred languages, even in the same game. For many users this would be much more comfortable.

I found out after a check that magiccards.info and gatherer both provide card pictures (and -text) in different languages!

**Gatherer** (you need to click on languages to see all available (printed) languages)
Quiet Contemplation (en)
Quiet Contemplation (de)
Quiet Contemplation (chinese simplified)
As you see the multiverse id is different for each language, even for the same card.

**magiccards.info**
http://magiccards.info/ktk/en/50.html
http://magiccards.info/ktk/de/50.html
http://magiccards.info/ktk/cn/50.html
The direct link to the images are "magiccards.info/scans/**en**/ktk/50.jpg" etc.

So in theory this should be possible...

---

**Edit:**
In the meantime, mci redirects all requests to **Scryfall**, which has different arts, too as [it's part of scryfall now](https://scryfall.com/blog/scryfall-magiccards-info-190)!
So the above mci links point to this scryfall pages:
https://scryfall.com/card/ktk/50/en/
https://scryfall.com/card/ktk/50/de/
https://scryfall.com/card/ktk/50/zhs/

marcofernandezheras:
Fix #841

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -340,6 +340,71 @@ void PictureLoader::processLoadQueue()
     }
 }
 
+int PictureLoader::getIdFromHtml(QString html, QString lang)
+{
+    QStringList langs = html.split("class=\"cardItem");
+    for(int i = 1; i < langs.size(); ++i)
+    {
+        const QString langHtml = langs.at(i);
+        if(langHtml.contains(lang,Qt::CaseInsensitive))
+        {
+            //qDebug() << langHtml;
+            int posStart = langHtml.indexOf("multiverseid=") + 13;
+            int posEnd = langHtml.indexOf("\">",posStart);
+            QString sId = langHtml.mid(posStart, posEnd-posStart);
+            return sId.toDouble();
+        }
+    }
+    return 0;
+}
+
+QMap<QString,QString> langDict;
+
+int PictureLoader::getTranslatedId(int currentId)
+{
+    if(langDict.isEmpty())
+    {
+        langDict["cs"] =  "繁體中文";
+        langDict["de"] =  "Deutsch";
+        langDict["es"] =   "Español";
+        langDict["fr"] =   "Français";
+        langDict["it"] =   "Italiano";
+        langDict["ja"] =   "日本語";
+        langDict["ko"] =   "한국어";
+        langDict["pt_BR"] =  "Português";
+        langDict["pt"] =     "Português";
+        langDict["ru"] =     "русский язык";
+        langDict["zh-Hans"] = "简体中文";
+    }
+
+    QString lang = langDict.value(settingsCache->getLang(), QStringLiteral("English"));
+    if(lang == QStringLiteral("English"))
+        return 0;
+
+    QEventLoop eventLoop;
+
+       QNetworkAccessManager mgr;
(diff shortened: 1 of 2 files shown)


cockatrice/src/carddatabase.cpp
```diff
@@ -340,6 +340,71 @@ void PictureLoader::processLoadQueue()
...
+        langDict["es"] =   "Español";
+        langDict["fr"] =   "Français";
+        langDict["it"] =   "Italiano";
+        langDict["ja"] =   "日本語";
+        langDict["ko"] =   "한국어";
+        langDict["pt_BR"] =  "Português";
+        langDict["pt"] =     "Português";
+        langDict["ru"] =     "русский язык";
+        langDict["zh-Hans"] = "简体中文";
+    }
```
tooomm: where do you take this languages from?
are they gatherer related? are they related to our language settings or files in cockatrice?

marcofernandezheras: the Map key is language settings related ( returned by settingsCache->getLang() ) and the values are gatherer related, used to search for muid translation (it the "Translated Language" column of the gathere language page)

cockatrice/src/carddatabase.cpp
```diff
@@ -340,6 +340,71 @@ void PictureLoader::processLoadQueue()
...
 
+int PictureLoader::getIdFromHtml(QString html, QString lang)
+{
+    QStringList langs = html.split("class=\"cardItem");
+    for(int i = 1; i < langs.size(); ++i)
+    {
+        const QString langHtml = langs.at(i);
+        if(langHtml.contains(lang,Qt::CaseInsensitive))
+        {
+            //qDebug() << langHtml;
```
ZeldaZach: This line should be removed

cockatrice/src/carddatabase.cpp
```diff
@@ -16,7 +16,7 @@
 #include <QDebug>
 #include <QImageReader>
 #include <QMessageBox>
-
+#include <QEventLoop>
```
ZeldaZach: You should keep a line separator between the includes and the code to the file

ZeldaZach:
My initial review shows that there are a lot of spacing issues. Try to use only one space on either side of the `=` symbol(s). Also, you should really do this on a sub-branch on your fork instead of the master.

marcofernandezheras:
Moved to a new branch, making PR from there

cockatrice/src/carddatabase.cpp
tooomm: Ok. Thanks for the insight!

2 things:
- cs is Česky (Czech) ;)<br>
I guess you use it for Chinese Traditional? The correct (not yet in the client, but you should implement+support it already) is "**zh-Hant**"
- You let the client download Portuguese cards for people that use either Portuguese or Brazilian Portuguese as their client language? That should be fine, I talked to two Brazilians.

You might want to update that in your new PR. ;)

PR #1326, opened 2015-08-03
Card image translation

marcofernandezheras:


cockatrice/src/carddatabase.cpp
```diff
@@ -340,6 +341,69 @@ void PictureLoader::processLoadQueue()
...
+    QString lang = langDict.value(settingsCache->getLang(), "English");
+    if(lang == "English")
+        return 0;
+
+    QEventLoop eventLoop;
+
+    QNetworkAccessManager mgr;
+    QObject::connect(&mgr, SIGNAL(finished(QNetworkReply*)), &eventLoop, SLOT(quit()));
+
+    QNetworkRequest req( QUrl( "http://gatherer.wizards.com/Pages/Card/Languages.aspx?multiverseid="+QString::number(currentId,'f',0)) );
```
Daenyth: This work should happen in Oracle, I don't think we should do it on the fly.

We could expand cards.xml to support a `lang` attribute on card text, card name, etc, and then cockatrice could select which version to display based on that. We'd set a fallback language of english, and the english cardname would still be the master in terms of what the client sends to the server and what it searches for when getting a name from the server.

Does that make sense?

ZeldaZach: I like the idea of having this information added to the database XML instead of on the fly too, now thinking about it further.

marcofernandezheras:
Before tried translation on fly, i tried to bruteforce translate the AllSets.json... it took around 22 hours Y.Y
So, I don't see viable to generate XML database from AllSets.json + translation it, not at least on all clients...
Maybe one person will generate the XML & all other client to download it?

tooomm:
We could generate+host the translation xml the same way we do for tokens, let oracle download it if needed/wanted and update it for new editions?

Other option would be to make language parsing optional in oracle and let you choose your preferred/needed language (next to english). Would that help to narrow the time down to like 3 hours if's just 1 instead of 6 languages? But not helping much about the problem. ;)

Daenyth:
What I had in my head was that the user would automatically get their localized language downloaded. Eg if oracle's language is set to russian, download russian + english.

Maybe @aeosynth could help us out here? If he could add a new object to the json which has translated names, or perhaps some kind of allsets-lang files (`allsets-en.json`, `allsets-ru.json` etc)

tooomm:
I'm really looking forward to some kind of solution to this - would love this feature! :)

But there should be an option in settings as well... a tick box to swap between english and language setting or even better a second pull down to freely customize pictures language unrelated to client language (default could be the same for sure). I'm thinking of that guy used to his own language and wants to use software in his native language, but still ends up buying english product cards or just likes to have the original wording.


I added a comment to my line note in the old PR btw:
https://github.com/Cockatrice/Cockatrice/pull/1325#discussion_r36172392

ZeldaZach:
@Daenyth I think you're referring to @Sembiance with the JSON information. But nevertheless, having foreign language JSON files would be a tremendous help!

Daenyth:
Derp you're right, I meant @Sembiance

ZeldaZach:
So is our idea to not do anything with this PR and wait for @sembiance to make changes to MTGJSON in the future?

Daenyth:
I mean, I'd like to get some confirmation from him either way - we have a few options

- If @sembiance wants to add language support to the json himself awesome
- We could send a PR to mtgjson code to add support
- Failing that we'd have to fall back to something worse inside oracle/cockatrice

Sembiance:
Foreign language support for MTGJSON is tracked here: https://github.com/Sembiance/mtgjson/issues/40

This is not very high on my To-Do list. I would be surprised if I tackle it this year.

It's not really a matter of the work involved to do it, that's not so bad. It's the future maintenance required to continually keep it all correct, bug free and up to date. At this time I am not very excited about signing myself up for that much more maintenance work.

ZeldaZach:
@Sembiance are you referring to only storing the mID's or the entire card text and such?

Sembiance:
@ZeldaZach I was mostly referring to full foreign language text. If you are just looking for multiverseid's for the foreign languages, that is far far less work and wouldn't be very hard to add or maintain. That is tracked here: https://github.com/Sembiance/mtgjson/issues/12

Of course this would only work for the cards that exist on gatherer.

ZeldaZach:
With the revisions to MTGJSON, I'm going to close this PR and wait for a PR that will update how Oracle runs ([See the changelog](http://mtgjson.com/#changeLog))

Issue #1330, opened 2015-08-04
No shortcut for copying card text


doombringer777:
I often use Cockatrice on Windows when offline as a type of Gatherer to find exact rules text. However something that constantly annoys me is the inability to use Ctrl + C or some other shortcut to copy rules text to the clipboard that I have selected in the deck editor.

![screenshot 2015-08-04 15 02 17](https://cloud.githubusercontent.com/assets/1956705/9053588/3a4b5c60-3aba-11e5-970c-56d17aaa212b.png)

Instead as shown in the screenshot I have to right click on the text then select Copy.

Is it possible to have the program detect if you are highlighting text and allow you t then use Ctrl + C to copy the highlighted text.

marcofernandezheras:
The problem is that ctrl+c is bound to "Save deck to clipboard"...

Daenyth:
@poixen do you think it would be hard to modify the ctrl+c handler so that it checks for highlighted text before doing deck copy?

ZeldaZach:
@Daenyth You can use `qtextcursor` potentially (See: [qt forum](https://forum.qt.io/topic/24525/how-to-check-if-the-selected-word-is-bold-or-not))

doombringer777:
Any progress on this?

Daenyth:
No.

ctrlaltca:
With the new configurable shortcuts, you can just remove the ctrl+c shortcut from "copy deck to clipboard" and it will start working.

PR #1332, opened 2015-08-04
Fixes issue #1330 : Card description textedit copy shortcut

marcofernandezheras:
Fixes issue #1330 : Changed "Save deck to clipboard" shorcut to cut secuence because it shadows text edit copy shortcut


Daenyth:
See my comment in #1330 - there might be another way to make ctrl+c contextually work for both.

ZeldaZach:
Maybe we can turn this into a shortcut then add it to the shortcut menu (#1382)

marcofernandezheras:
Already on #1382  under "Deck Editor" > "Save deck (Clipboard)"

PR #1338, opened 2015-08-05
Client ID Generation

woogerboy21:
Implemented cockatrice client id generation and the storing of the id into the database.
The server does not require it, and doesn't really care if the client sends it at this point.  I dont intend on writing any server side expectations with this PR as that will be future development expansion to the functionality.

Also if some one could tell me how to properly get rid of all the commits listed below that would be helpful.  I tried and just ended up adding more.


woogerboy21:
Going to recreate PR to get rid of the extra fluff.

Issue #170, opened 2014-07-01
Add user suspicion level for moderators


Daenyth:
With large numbers of connected users, it's sometimes hard to figure out who is worth paying attention to.

Users are usually not suspicious when
- They registered a long time ago
- Their sessions tend tower less frequent but longer sessions (rather than constantly reconnecting)
- There aren't other users on the server sharing their ip address
- They have had few or no bans in the past
- They have lots of people who added them as a buddy
- They have few people who ignore them
- They are moderators or server admins

Servatrice should send back an extra field in the user info when talking to a moderator giving a number summarizing this.

Cockatrice can then take this number and apply a gradient to people in the user list, giving a quick visual indication when users are worth watching.

Moderators should also have the ability to manually tweak this number up and down.

cc @woogerboy21 @Beusapu

woogerboy21:
@Daenyth While working through some of the more recent idea's in the client additions I try and keep this in mind as it plays into a number of things I have in mind.  One question that has recently come up would be "At what point does the user suspicion level start to go down?".  Are we concerned about a users behavior forever?  Is after a period of time things start to drop?  Or think about it this way.  If a user starts to act out and the moderators get informed by the color change.  Then the user stops acting out and the suspicion level never drops.  Will that be a bad thing?  Would mods be constantly be notified of a user that is not really needing monitored?

ZeldaZach:
@woogerboy21 Maybe set an `expiration_date` field for each thing added to the `userSuspicion` table? Leave them there for reference, but stop counting them against the player after X time.

woogerboy21:
@Daenyth how often were you thinking this number should be calculated?

Daenyth:
On user login at most once per day should be plenty.

Issue #42, opened 2014-01-23
Additional Banning Functionality


woogerboy21:
I'm opening this issue in response to a request by Gavin.  Currently the banning functionality of with in the server side is very easy to circumvent by toxic users.  The current methods of banning include by username or by IP.  With out the ability to restrict the server to registered users only name banning is not possible and so IP banning must be used.  Sadly the problematic users can circumvent this ban by using proxy servers or simply resetting there modems.  Since the majority of users external IP address's are given to them by DHCP it is very easy for a user to change there IP address.  So what happens is you end up banning a bunch of IP's or usernames and the problematic user simply goes through a proxy, registers a new user name, or even worse yet simply logs in as an unregistered user.

My suggestion was to add an additional banning function to the server / client.  The thought I had would be that each client would generate a client id upon every startup of the application or possibly on every attempt to connect to a server.  This client id would need to be generated from a non-changing hardware specific device.  The MAC address of the computer that is being used to connect to the server is a good example (probably the real only example I can come up with).  Once there is a unique client id per client the server then would need the ability to ban the client by that identifier.  With this type of system in place it would prevent the current known methods of ban evasion.

Any other thoughts on how to prevent ban evasion?  Simply allowing only registered users would be a great leap in security at the very least.  At least having that ability would put some type of hurdle in place to prevent such simple ban evasions.

PR #1340, opened 2015-08-05
Client ID Generation

woogerboy21:
Re-roll of the PR for client side ID generation.  Currently the id is based on a hash of all the valid mac address's from across all the adapters in a PC.  We can add / change the logic for the client ID generation as it is just a function being called.  The client ID is then passed to the server and stored in the user database table with the account during login.

Right now the server does nothing with the ID other than store it.  This PR is designed to simply add design space into the management and tracking of users with the thought of adding additional banning functionality at a later date as well as possibly assist in the creation of things like suspicion level indication (https://github.com/Cockatrice/Cockatrice/issues/170)  or statistical data analysis.

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -1,4 +1,4 @@
-/***************************************************************************
+﻿/***************************************************************************
  *   Copyright (C) 2008 by Max-Wilhelm Bruker   *
  *   brukie@gmx.net   *
  *                                                                         *
@@ -32,6 +32,9 @@
 #include <QDesktopServices>
 #include <QDebug>
 #include <QSystemTrayIcon>
+#include "QtNetwork/QNetworkInterface"
+#include <QCryptographicHash>
+
 
 #include "main.h"
 #include "window_main.h"
@@ -95,6 +98,19 @@ bool settingsValid()
         !settingsCache->getDeckPath().isEmpty() &&
         QDir(settingsCache->getPicsPath()).exists() &&
         !settingsCache->getPicsPath().isEmpty();
+}
+
+void generateClientID()
+{
+    QString macList;
+    foreach(QNetworkInterface interface, QNetworkInterface::allInterfaces())
+    {
+        if (interface.hardwareAddress() != "")
+            if (interface.hardwareAddress() != "00:00:00:00:00:00:00:E0")
+                macList += interface.hardwareAddress() + ".";
+    }
+    QString strClientID = QCryptographicHash::hash(macList.toUtf8(), QCryptographicHash::Sha1).toHex().right(15);
+    settingsCache->setClientID(strClientID);
 }
 
 int main(int argc, char *argv[])
@@ -203,6 +219,9 @@ int main(int argc, char *argv[])
 
         QIcon icon(":/resources/appicon.svg");
         ui.setWindowIcon(icon);
+        
+        generateClientID();    //generate the users client id
+        qDebug() << "ClientID In Cache: " << settingsCache->getClientID();
 
         ui.show();
         qDebug("main(): ui.show() finished");
--- a/common/pb/response.proto
+++ b/common/pb/response.proto
@@ -1,58 +1,59 @@
 message Response {
-	enum ResponseCode {
-		RespNotConnected = -1;
-		RespNothing = 0;
-		RespOk = 1;
-		RespNotInRoom = 2;
-		RespInternalError = 3;
-		RespInvalidCommand = 4;
-		RespInvalidData = 5;
-		RespNameNotFound = 6;
-		RespLoginNeeded = 7;
-		RespFunctionNotAllowed = 8;
-		RespGameNotStarted = 9;
-		RespGameFull = 10;
-		RespContextError = 11;
-		RespWrongPassword = 12;
-		RespSpectatorsNotAllowed = 13;
-		RespOnlyBuddies = 14;
-		RespUserLevelTooLow = 15;
-		RespInIgnoreList = 16;
-		RespWouldOverwriteOldSession = 17;
-		RespChatFlood = 18;
-		RespUserIsBanned = 19;
-		RespAccessDenied = 20;
-		RespUsernameInvalid = 21;
-		RespRegistrationRequired = 22;
-		RespRegistrationAccepted = 23; // Server agrees to process client's registration request
-		RespUserAlreadyExists = 24; // Client attempted to register a name which is already registered
-		RespEmailRequiredToRegister = 25; // Server requires email to register accounts but client did not provide one
-		RespTooManyRequests = 26; // Server refused to complete command because client has sent too many too quickly
-		RespPasswordTooShort = 27; // Server requires a decent password
-		RespAccountNotActivated = 28; // Client attempted to log into a registered username but the account hasn't been activated
-		RespRegistrationDisabled = 29; // Server does not allow clients to register
-		RespRegistrationFailed = 30; // Server accepted a reg request but failed to perform the registration
-		RespActivationAccepted = 31; // Server accepted a reg user activation token
-		RespActivationFailed = 32; // Server didn't accept a reg user activation token
-		RespRegistrationAcceptedNeedsActivation = 33; // Server accepted cient registration, but it will need token activation
-	}
-	enum ResponseType {
-		JOIN_ROOM = 1000;
-		LIST_USERS = 1001;
-		GET_GAMES_OF_USER = 1002;
-		GET_USER_INFO = 1003;
-		DUMP_ZONE = 1004;
-		LOGIN = 1005;
-		DECK_LIST = 1006;
-		DECK_DOWNLOAD = 1007;
-		DECK_UPLOAD = 1008;
-		REGISTER = 1009;
-		ACTIVATE = 1010;
-		REPLAY_LIST = 1100;
-		REPLAY_DOWNLOAD = 1101;
-	}
-	required uint64 cmd_id = 1;
-	optional ResponseCode response_code = 2;
-	
-	extensions 100 to max;
+    enum ResponseCode {
+        RespNotConnected = -1;
+        RespNothing = 0;
+        RespOk = 1;
+        RespNotInRoom = 2;
+        RespInternalError = 3;
+        RespInvalidCommand = 4;
+        RespInvalidData = 5;
+        RespNameNotFound = 6;
+        RespLoginNeeded = 7;
+        RespFunctionNotAllowed = 8;
+        RespGameNotStarted = 9;
+        RespGameFull = 10;
+        RespContextError = 11;
+        RespWrongPassword = 12;
+        RespSpectatorsNotAllowed = 13;
+        RespOnlyBuddies = 14;
+        RespUserLevelTooLow = 15;
+        RespInIgnoreList = 16;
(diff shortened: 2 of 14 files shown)


tooomm:
Will it store every single ID that I ever connect with?
For example if I use several different computers (work + private for example) or buy a new one? Or get a new WiFi adapter?

woogerboy21:
It stores the last generated id of the client on the account at time of login and is overwritten upon next login. So if a user is logging in from different computers across the same account the user account will only show the last generated id from the last machine the user logged in from.

tooomm:
The client generates the id every start? Or just at installation?

I don't like to store any kind of this information. :)
I'm not even sure if there is a benefit at all or a serious improvement for the banning you mentioned.
I remember some older controversial chats on gitter with @Daenyth, you and some others about it...

woogerboy21:
The client generates the id on every start.  Its only stored locally during runtime in a settings cache location in the running client nor is the original client side hardware data (private information) sent.  Its a hash of the values collected and truncated so you wont even be able to reverse engineer any type of information from the data sent. The fact that when you log in your sourcing ip address is saved server side is a bigger type of security or personal information threat than this would be.

As far as it being an improvement, and strictly talking about banning capabilities.  With the current method of any banning we are relying on user name and source IP which is proven to be easily circumventable.  And the conversation that was controversial was sending raw data that is unique to a users machine (like the real mac address) as that may be used for malicious purposes if some one was smart enough to do so.

MassCraziness:
If this does improve banning functionality and can remove the necessity of storing IP Addresses while not being a bigger security risk, then I'm all for this.

I just want to be sure that's what you're trying to say. I think it is, though.

woogerboy21:
The unique user id generation has already been discussed through older issue's (#42 as an example).  There were others but im not certain they other issue cases have been closed rolling them into a larger issue.

With a unique user id generated based on something client side there is potential for the use of it as an identifier per user rather than there sourcing IP.  Keep in mind removing the storing and usage of source IP will be a bit of an undertaking but I had hoped to eventually get to that point.  Im also hoping to eventually hash the users email address's that get stored in the database so that if there is a security breach that information is not in plain text as well.  Currently with the way things are stored in the database if a security breach were to occur there is a table that has not only the user name but email , also there is a session table that links the users user name to a session table entry that includes the users source IP.   So if a crafted attack were to occur some one could potentially get the user name, email information, source ip information in plain text.  And given enough time could probably brute force there way into figuring out what the users account password is.  So if users use the same password for cockatrice and there email address username / password there is a good chance this information could be compromised very easily in the current state.

Generating a UUID has more benefits in my mind than negatives.

ZeldaZach:
I'm fine with this :shipit:

Daenyth:
I don't see ip address usage being removed

cockatrice/src/main.cpp
```diff
@@ -97,6 +100,19 @@ bool settingsValid()
         !settingsCache->getPicsPath().isEmpty();
 }
 
+void generateClientID()
```
Daenyth: I didn't get a chance to look at this before merge; in general I'd really prefer if this returned a QString without touching settings and then we store it outside the method. As it stands it doesn't really follow SRP and the inclusion of a side effect in the method means that we have to refactor it or work around it for testing

cockatrice/src/main.cpp
```diff
@@ -203,6 +219,9 @@ int main(int argc, char *argv[])
 
         QIcon icon(":/resources/appicon.svg");
         ui.setWindowIcon(icon);
+        
+        generateClientID();    //generate the users client id
```
Daenyth: This comment is useless

common/pb/response.proto
```diff
@@ -1,58 +1,59 @@
 message Response {
```
Daenyth: Reformatting the whole file here in the same commit as editing it makes it hard to review.

common/pb/response.proto
```diff
@@ -1,58 +1,59 @@
...
+        RespEmailRequiredToRegister = 25; // Server requires email to register accounts but client did not provide one
+        RespTooManyRequests = 26; // Server refused to complete command because client has sent too many too quickly
+        RespPasswordTooShort = 27; // Server requires a decent password
+        RespAccountNotActivated = 28; // Client attempted to log into a registered username but the account hasn't been activated
+        RespRegistrationDisabled = 29; // Server does not allow clients to register
+        RespRegistrationFailed = 30; // Server accepted a reg request but failed to perform the registration
+        RespActivationAccepted = 31; // Server accepted a reg user activation token
+        RespActivationFailed = 32; // Server didn't accept a reg user activation token
+        RespRegistrationAcceptedNeedsActivation = 33; // Server accepted cient registration, but it will need token activation
+        RespClientIDRequired = 34; // Server require's client to generate and send its client id before allowing access
```
Daenyth: `require's` -> `requires`

common/pb/session_commands.proto
```diff
@@ -1,157 +1,158 @@
...
-	}
-	optional uint32 room_id = 1;
+    extend SessionCommand {
+        optional Command_JoinRoom ext = 1015;
+    }
+    optional uint32 room_id = 1;
 }
 
 // User wants to register a new account
 message Command_Register {
```
Daenyth: A future extension could submit clientID with this message so that we can for bans before registering

common/server.cpp
```diff
@@ -168,6 +170,15 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString
     event.mutable_user_info()->CopyFrom(session->copyUserInfo(true, true, true));
     locker.unlock();
     
+    // check if client id exists (older client compatibility)
```
Daenyth: Comment could be reworded to be more useful.

common/server.cpp
```diff
@@ -168,6 +170,15 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString
...
+    if (clientid.isEmpty()){
+        // client id is empty, either out dated client or client has been modified
```
Daenyth: We should add a logger here or check server settings to see if the id is required.

common/server.cpp
```diff
@@ -168,6 +170,15 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString
...
+    }
+    else {
+        // update users database table with client id
+        databaseInterface->updateUsersClientID(name, clientid);
```
Daenyth: We should move to having one post-login call that updates all the "most recent" fields; this will increase DB load a bit for each connection.

common/server_protocolhandler.cpp
```diff
@@ -381,11 +381,12 @@ Response::ResponseCode Server_ProtocolHandler::cmdPing(const Command_Ping & /*cm
 Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd, ResponseContainer &rc)
 {
     QString userName = QString::fromStdString(cmd.user_name()).simplified();
+    QString clientid = QString::fromStdString(cmd.clientid()).simplified();
```
Daenyth: Shouldn't we check that this isn't empty? Also should be `clientId`

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -811,20 +811,33 @@ bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user, const
...
+void Servatrice_DatabaseInterface::updateUsersClientID(const QString &userName, const QString &userClientID)
+{
+    
+    if (!checkSql())
+        return;
+
+    QSqlQuery *query = prepareQuery("update {prefix}_users set clientid = :clientid where name = :username");
+    query->bindValue(":clientid", userClientID);
+    query->bindValue(":username", userName);
+    execSqlQuery(query);
```
Daenyth: We need to check for query errors here and log them

Daenyth:
For the future I'd strongly prefer that PRs including protocol extensions have a longer period for review

common/pb/session_commands.proto
woogerboy21: Are you expecting to be able to ban users based on rooms?

Daenyth: I'm commenting on the register command.

common/server.cpp
woogerboy21: I started to add the functionality for requiring the client ID (as you can see in the response protocol place holder) but decided to hold off till we decided what was to be done with the client IDs.

servatrice/src/servatrice_database_interface.cpp
woogerboy21: For that matter, we should update just about every database query, that does insertion at least, and do this :)

Daenyth: We probably should.

cockatrice/src/main.cpp
woogerboy21: Could you give me an example of what would need refactored to work around it?  Returning the value then immediately storing it makes almost no sense to me as all it does is require additional code line.  Its not that big a deal one way or the other IMO and I can see were you may want consistency in the code since the other functions through out the code do as youyou describe.

Daenyth: In a unit test that calls this method and nothing else, the test now depends on the global shared settings interface. Since this is global mutable state the test then has to remove the setting after running itself. Contrast with a version that returns QString without touching settings - we just call the method and check the return value. I know we don't have automated testing in place yet but it's one of my higher-priority goals for the project and I'd like to make the inclusion as smooth as I can.

common/server.cpp
woogerboy21: Are you saying you want to combine all the update calls to the DB into a single function / db call?  Cause simply calling the DB post logic with multiple function calls doesnt change things a whole lot in terms of performance and questionably would increase the amount of time the user table gets locked when the updates are occuring.

cockatrice/src/main.cpp
Daenyth: It still has a dependency on the network interface but we could work around that as well.

common/pb/session_commands.proto
woogerboy21: Ah ok, then yea that makes since.

common/server.cpp
Daenyth: I'm saying that whatever gets updated post-login should happen in one call
with 0 or 1 lock instead of X calls with a lock each

cockatrice/src/main.cpp
woogerboy21: Could you also explain to me how you would call the function and nothing else when the function is inside a file that includes/contains the 2 required components (network and settings)?  I can understand what your saying to a point and fully intend on fixing it,  but im just curious if your stretching things on this one a bit.

Daenyth: It would be enough to test that it generates the same id when called twice.

woogerboy21: Would you prefer the value be returned in a variable and then the setvalue function called to store it in the settings cache or do you care if I simply make it all a single line without requiring a variable to be defined?

common/server.cpp
woogerboy21: There is no lock being made when updating the client id on the user table?  No multiple locks are going on here.  Now you could say that maybe we should apply the client id inside the locking block but I thought you wanted to move away from locking the user table?

Daenyth: One line is fine.

And for updates you mentioned locks, I was going off that

PR #1345, opened 2015-08-08
Add more informative message on unrecognized login response

Daenyth:
When logging in and receiving a reply that isn't understood, show a more informative error message.

`tr() + tr()` for better translation flexibility.


woogerboy21:
:shipit:

tooomm:
The tr() split up is interesting... is that something we should go for in all client?
We have quite some large messages that could be split up!

Right now the problem is that if you change a tiny wording etc. all the message changes to untranslated again. With splitup at least the general term like `Unknown login error:` here stays translated.

Daenyth:
That was my idea

PR #1346, opened 2015-08-08
Reg only server cleanup

woogerboy21:
Moved the reg only server code functions from the database interface code block(s) into the proper server block of code.


ZeldaZach:
Seems good to me. Nice pickup on the spaces! :shipit:

PR #1344, opened 2015-08-07
Commit to resolve requests made by @Daenyth discussed after the commit.

woogerboy21:
See https://github.com/Cockatrice/Cockatrice/pull/1340 for details.
This should resolve everything mentioned except the following:

* Database update success check 
  * This should be updated in a future PR as additional database components outside the client ID PR should be updated as well and by implementing the check in this single commit goes against the way the other functions for handling database functionality has been written in the past.

PR #1350, opened 2015-08-09
Update Error 11 Clarity

woogerboy21:
Added case statement for the RespContextError return code when login failures occur due to updates in the protocol standards.

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -319,6 +319,9 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
         case Response::RespClientIdRequired:
             QMessageBox::critical(this, tr("Error"), tr("This server requires client ID's. Your client is either failing to generate an ID or you are running a modified client.\nPlease close and reopen your client to try again."));
             break;
+        case Response::RespContextError:
+            QMessageBox::critical(this, tr("Error"), tr("An internal error has occurred, please try closing and reopening your client and try again. If the error persists try updating your client to the most recent build and if need be contact your software provider."));
+            break;
         case Response::RespAccountNotActivated: {
             bool ok = false;
             QString token = QInputDialog::getText(this, tr("Account activation"), tr("Your account has not been activated yet.\nYou need to provide the activation token received in the activation email"), QLineEdit::Normal, QString(), &ok);


cockatrice/src/window_main.cpp
```diff
@@ -319,6 +319,9 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
         case Response::RespClientIdRequired:
             QMessageBox::critical(this, tr("Error"), tr("This server requires client ID's. Your client is either failing to generate an ID or you are running a modified client.\nPlease close and reopen your client to try again."));
             break;
+        case Response::RespContextError:
+            QMessageBox::critical(this, tr("Error"), tr("An internal error has occurred, please try closing and reopening your client and try again. If the error persists try updating your client to the most recent build and if need be contact your software provider."));
```
Daenyth: internal error where? In http-speak that would indicate a server-side error.

This error message doesn't help the user debug anything and it doesn't give us any information to use in ticket investigation

Daenyth: Looking at the usages of that response in the codebase I feel like we should add a debug value field to display in cases like this that the server can fill with a useful string.

woogerboy21: The error message only occurs in the login process and not any were else. Could there be a better description, sure.  Is it better than before. YES.

Daenyth: The server returns this response in a lot of places.

Daenyth: Under normal operation it should never happen, it indicates a bug

woogerboy21: Or mis/matching response containers due to older client compatibility ;)

Daenyth: We shouldn't have those kind of updates. Did we?

Generally adding a new proto item or adding a new field to an existing item is safe. It's only really unsafe if you remove existing fields or messages.

woogerboy21:
Your request to have the clientid check in the login container (#1344) forces a client update :confused:
When I rolled the change into the server I took the check out to allow older clients connectivity until the most recent release was published and at a later date when I upgrade the server again I'll keep it in.

cockatrice/src/window_main.cpp
woogerboy21: Your request to have the clientid check in the login container (#1344) forces a client update :confused:
When I rolled the change into the server I took the check out to allow older clients connectivity until the most recent release was published and at a later date when I upgrade the server again I'll keep it in.

Daenyth: I must not have reviewed the change enough. You can add new fields to
existing messages without breaking them

woogerboy21: Like I mentioned, if its alright with you ill probably put a PR in to remove the check from the protocol handler portion of the code since there is a check on the server side that the operator can require the clientid.  With the protocol handler check in place it kind of makes the server side check useless.

PR #1354, opened 2015-08-09
Fix #1351 - Server ignores only-reg if in "none" auth mode.

### LINKED ISSUE #1351: Due to user registration, it isn't possible to set up a server without a database
If a server is set up without a database, there is not such thing as a registered user.
The client only allows registered users to set up games which can be joined by non-registered users, therefore it's impossible to ever set up a game which another user can join.
Setting up a server without a database is useful for a quick home server which won't be exposed to the outside world.

Daenyth:
When in none auth mode, registered users don't exist; hence the setting
makes the game impossible to join.

Normally the setting would be greyed out when the user is unregistered,
but commit 475c54bf introduced a new edge case where the stored setting
would apply to the field even if the user is unregistered, making the
setting possible to apply in no-reg mode. Regardless, any client could
have done this, not just the default cockatrice.

The server side fix should prevent all issues in the future.

### DIFF
--- a/cockatrice/src/dlg_creategame.cpp
+++ b/cockatrice/src/dlg_creategame.cpp
@@ -114,7 +114,13 @@ DlgCreateGame::DlgCreateGame(TabRoom *_room, const QMap<int, QString> &_gameType
     descriptionEdit->setText(settingsCache->getGameDescription());
     maxPlayersEdit->setValue(settingsCache->getMaxPlayers());
     onlyBuddiesCheckBox->setChecked(settingsCache->getOnlyBuddies());
-    onlyRegisteredCheckBox->setChecked(settingsCache->getOnlyRegistered());
+    if (room && room->getUserInfo()->user_level() & ServerInfo_User::IsRegistered)
+    {
+        onlyRegisteredCheckBox->setChecked(settingsCache->getOnlyRegistered());
+    } else {
+        onlyBuddiesCheckBox->setEnabled(false);
+        onlyRegisteredCheckBox->setEnabled(false);
+    }
     spectatorsAllowedCheckBox->setChecked(settingsCache->getSpectatorsAllowed());
     spectatorsNeedPasswordCheckBox->setChecked(settingsCache->getSpectatorsNeedPassword());
     spectatorsCanTalkCheckBox->setChecked(settingsCache->getSpectatorsCanTalk());
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -636,7 +636,9 @@ Response::ResponseCode Server_ProtocolHandler::cmdCreateGame(const Command_Creat
     if (description.size() > 60)
         description = description.left(60);
 
-    Server_Game *game = new Server_Game(copyUserInfo(false), gameId, description, QString::fromStdString(cmd.password()), cmd.max_players(), gameTypes, cmd.only_buddies(), cmd.only_registered(), cmd.spectators_allowed(), cmd.spectators_need_password(), cmd.spectators_can_talk(), cmd.spectators_see_everything(), room);
+    // When server doesn't permit registered users to exist, do not honor only-reg setting
+    bool onlyRegisteredUsers = cmd.only_registered() && (server->permitUnregisteredUsers());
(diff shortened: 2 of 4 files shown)


Daenyth:
Need to fix:
```
/home/travis/build/Cockatrice/Cockatrice/common/server_protocolhandler.cpp: In member function ‘Response::ResponseCode Server_ProtocolHandler::cmdCreateGame(const Command_CreateGame&, Server_Room*, ResponseContainer&)’:
/home/travis/build/Cockatrice/Cockatrice/common/server_protocolhandler.cpp:640:66: error: ‘class Server’ has no member named ‘getAuthenticationMethod’
/home/travis/build/Cockatrice/Cockatrice/common/server_protocolhandler.cpp:640:95: error: ‘Servatrice’ has not been declared
make[2]: *** [common/CMakeFiles/cockatrice_common.dir/server_protocolhandler.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [common/CMakeFiles/cockatrice_common.dir/all] Error 2
make: *** [all] Error 2
```

woogerboy21:
So whats the idea behind making the change in the server protocol handler and not in the actual servatrice game server?

Daenyth:
Because that's where the game is created

woogerboy21:
I would suggest the check fall in at https://github.com/Cockatrice/Cockatrice/blob/master/common/server_game.cpp#L77 (but thats just my $0.02.)

Daenyth:
I don't think the game code should care about the server's auth settings. Ultimately it will work in either place; I think the protocol dispatcher is the right scope.

common/server_protocolhandler.cpp
```diff
@@ -636,7 +636,9 @@ Response::ResponseCode Server_ProtocolHandler::cmdCreateGame(const Command_Creat
     if (description.size() > 60)
         description = description.left(60);
 
-    Server_Game *game = new Server_Game(copyUserInfo(false), gameId, description, QString::fromStdString(cmd.password()), cmd.max_players(), gameTypes, cmd.only_buddies(), cmd.only_registered(), cmd.spectators_allowed(), cmd.spectators_need_password(), cmd.spectators_can_talk(), cmd.spectators_see_everything(), room);
+    // When server doesn't permit registered users to exist, do not honor only-reg setting
+    bool onlyRegisteredUsers = cmd.only_registered() && (server->getAuthenticationMethod() != Servatrice::AuthenticationNone);
```
woogerboy21: Looking into this I believe you can simply check for authState != UnknownUser 
(instead of trying to call the server->getauthmethod funnyness your doing).

ZeldaZach:
Is anything being done with this? I'd really like to see some kind of fix in place

Daenyth:
I need to fix the issues and then do some testing

woogerboy21:
Just curious, would it be safe to simply test for a DB connection and if not there force the value to be false rather than trying to pull the authentication method back?  (may be a quicker / easier solution).

Daenyth:
That would work under the current model but I think this is more correct. It's not even that much work, I just have to have time enough to sit down and do it.

Daenyth:
@woogerboy21 This should be ready to test

cockatrice/src/dlg_creategame.cpp
```diff
@@ -114,7 +114,13 @@ DlgCreateGame::DlgCreateGame(TabRoom *_room, const QMap<int, QString> &_gameType
     descriptionEdit->setText(settingsCache->getGameDescription());
     maxPlayersEdit->setValue(settingsCache->getMaxPlayers());
     onlyBuddiesCheckBox->setChecked(settingsCache->getOnlyBuddies());
-    onlyRegisteredCheckBox->setChecked(settingsCache->getOnlyRegistered());
+    if (room && room->getUserInfo()->user_level() & ServerInfo_User::IsRegistered)
+    {
+        onlyRegisteredCheckBox->setChecked(settingsCache->getOnlyRegistered());
+    } else {
+        onlyBuddiesCheckBox->setEnabled(false);
+        onlyRegisteredCheckBox->setEnabled(false);
```
ZeldaZach: add `setChecked(false)` for both of these please :)

Daenyth: It happens in `sharedCtor`

PR #1356, opened 2015-08-10
Username Completer in server room

### LINKED ISSUE #457: Add nickname autocompletion
In any chat window, I should be able to type @u and an autocomplete menu should come up with all users starting with `u` in the room. esc to cancel it, up/down to select a name if there's multiple, Tab/enter/space should complete to the name.

Fizztastic:
Fix #457
- Adds QCompleter in server room chat to assist in typing usernames.
- Adds a setting to enable/disable it under Settings > Chat.
- Implicit Shortcuts: Escape (cancel current completion), Enter/Tab/Space (insert current completion)

settings
chat

### DIFF
--- a/cockatrice/src/tab_room.cpp
+++ b/cockatrice/src/tab_room.cpp
@@ -11,6 +11,14 @@
 #include <QSplitter>
 #include <QApplication>
 #include <QSystemTrayIcon>
+#include <QCompleter>
+#include <QWidget>
+#include <QTextCursor>
+#include <QAbstractItemView>
+#include <QScrollBar>
+#include <QStringListModel>
+#include <QKeyEvent>
+#include <QFocusEvent>
 #include "tab_supervisor.h"
 #include "tab_room.h"
 #include "tab_userlists.h"
@@ -51,8 +59,9 @@ TabRoom::TabRoom(TabSupervisor *_tabSupervisor, AbstractClient *_client, ServerI
     connect(chatView, SIGNAL(showCardInfoPopup(QPoint, QString)), this, SLOT(showCardInfoPopup(QPoint, QString)));
     connect(chatView, SIGNAL(deleteCardInfoPopup(QString)), this, SLOT(deleteCardInfoPopup(QString)));
     connect(chatView, SIGNAL(addMentionTag(QString)), this, SLOT(addMentionTag(QString)));
+    connect(settingsCache, SIGNAL(chatMentionCompleterChanged()), this, SLOT(actCompleterChanged()));
     sayLabel = new QLabel;
-    sayEdit = new QLineEdit;
+    sayEdit = new CustomLineEdit;
     sayLabel->setBuddy(sayEdit);
     connect(sayEdit, SIGNAL(returnPressed()), this, SLOT(sendMessage()));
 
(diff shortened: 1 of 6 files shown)


tooomm:
Very cool!

Some thoughts/questions though:
- `Enter selects first entry`
Is the first entry (in the last screen) the one on top of the list or the one closest to the typing field?
Here in git the list appears beneath you and both applies (top of the list + closest to the typing field).

- Is the first entry highlighted, that you get some help what is going to be selected if you hit enter? It will show the user that you can interact with it, too.

- Can you use the arrow keys to navigate in the list of suggestions up/down?

- Is there a reason that this is just working in server room chat? (not in game chat e.g.)
If people play drafts and need to contact their opponent and tell them that the room is created this feature would help, too!

Fizztastic:
@tooomm 

* The first entry is the one on top of the list.
* The first entry doesn't need to be highlighted for it to be selected when you hit enter.
* The up/down arrow keys do work to navigate the suggestions.
* Currently this is only working in the Server room. I'll look into expanding it to private chat as well.

cockatrice/src/tab_room.cpp
```diff
@@ -104,12 +110,27 @@ TabRoom::TabRoom(TabSupervisor *_tabSupervisor, AbstractClient *_client, ServerI
...
+    }
     userList->sortItems();
 
     const int gameListSize = info.game_list_size();
     for (int i = 0; i < gameListSize; ++i)
         gameSelector->processGameInfo(info.game_list(i));
+
+    completer = new QCompleter(autocompleteUserList, sayEdit);
+    completer->setCaseSensitivity(Qt::CaseInsensitive);
+    completer->setCompletionRole(2);
```
ZeldaZach: This isn't needed as line 133 has the same thing and is overriding things

cockatrice/src/tab_room.cpp
```diff
@@ -259,3 +292,49 @@ void TabRoom::sendRoomCommand(PendingCommand *pend)
 {
     client->sendCommand(pend);
 }
+
+CustomLineEdit::CustomLineEdit(QWidget *parent) 
+    : QLineEdit(parent)
```
ZeldaZach: You can put this on one line. We split it up to two when it gets too long

ZeldaZach:
@Fizztastic Can you incorporate `tab` to be used as an `enter` key when the menu appears (Like gitter)?

ZeldaZach:
In addition, this list doesn't update when new users login to the client. So if I login and then `foo` logs in, he won't be in the `@f` popup list.

tooomm:
To quote @Daenyth in his initial issue (#457) about this:
> In any chat window, I should be able to type `@u` and an autocomplete menu should come up with all users starting with `u` in the room. **esc** to cancel it, up/down to select a name if there's multiple, **Tab/enter/space** should complete to the name.

ZeldaZach:
Nice find @tooomm I agree with everything said there

ZeldaZach:
Seems like whenever a new user joins the room the popup just disappears and won't re-appear unless you logout/login of the server.

Users online: `ZeldaZach`, `ZeldaZach3`:
screenshot 2015-08-11 19 30 09


User `ZeldaZach4` logs in:
screenshot 2015-08-11 19 30 25

ZeldaZach:
It seems that when a user logs on they are added to the list, but when they log out they aren't removed. Can you look into removing users who go offline from the mention list?

cockatrice/src/tab_room.cpp
```diff
@@ -259,3 +296,101 @@ void TabRoom::sendRoomCommand(PendingCommand *pend)
...
+    c = completer;
+    c->setWidget(this);
+    connect(c, SIGNAL(activated(QString)),this, SLOT(insertCompletion(QString)));
+}
+
+void CustomLineEdit::updateCompleterModel(QStringList completionList)
+{
+    if (!c)
+        return;
+
```
ZeldaZach: Can you combine these two into
```
if (!c || c->popup()->isVisible())
    return;
```

ZeldaZach:
Seems good to me. My only concern is I don't fully understand the functions in the `CustomLineEdit` class. Once those are reviewed by someone else and cleared that they're the most efficiently written, I'm fine with this. :shipit:

tooomm:
will this take into account all players on the server or just the players in the room you are chatting in?
should be the latter.

Fizztastic:
@tooomm It only takes into account the players that are in the same room.

woogerboy21:
:+1: So far this is working as expected for me on the testing client i've been used while logged in.

Issue #529, opened 2015-01-05
Make game replays tabs always accessible 


anoda9:
Currently, the deck storage and game replays tabs only appear when you connect to the server, but it would be better if they were accessible even when not connected. Users may still want to browse decks and replays stored on their local file system without being connected to the server.

tooomm:
I could imagine a grayed out right side (server storage) + including an overlay info message.
Maybe a button to connect to server, too

ZeldaZach:
I like the idea

Daenyth:
:+1: this would be nice.

ZeldaZach:
Deck storage doesn't make sense as decks are _stored on a server_. However, viewing downloaded replays seems reasonable. Relabeling ticket.

ebbit1q:
this is already a thing, you can open the replays that are stored on your file system by clicking the top left menu (Cockatrice) then "Watch replay"

PR #1361, opened 2015-08-10
Add option to servatrice to disable replay storage.

### LINKED ISSUE #380: Additional Configuration Options for Servatrice
Created by request of @Daenyth on #343.
It would be nice to add additional configuration options to the servatrice .ini that allows to enable/disable the functionality that servatrice does.  Things such as enable/disable the storing of replay data is an example that comes to mind.

With these options correctly defined in the settingsCache it would allow for a UI tab that could be created to allow for the features to be configured during operation of the server along with the ability for saving the options in a given state.

woogerboy21:
Fix #380
(i believe, but if not more can be added).
This allows the server operator to set servatrice to not store the replay data into the database.  This option would be useful for operators with small amounts of storage on their server.  With this setting the behavior is that a game is created by a user and recorded allowing for the client to review the game after a game is complete. But when the value is set on the server side to not store the data, the data never gets stored into the database allowing for review of the game up till the client closes.

### DIFF
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -193,6 +193,11 @@ roomlist\1\game_types\3\name="GameType3"
 ; default is 120
 max_game_inactivity_time=120
 
+; All actions during a game are recorded and stored in the database as a replay that all participants of
+; the game can go back to and review after the game is closed.  This can require a fairly large amount of
+; storage to save all the information.  Disable this option to prevent the storing of replay data in
+; the database.  Default value is true.
+store_replays=true
 
 [security]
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Seems fine to me, provided it was tested on a server. :shipit:

woogerboy21:
Yes it was tested on a local server I use for testing.

tooomm:
Will this disable/hide the replay tab as well?

Daenyth:
No

tooomm:
No change needed about hiding the tab with #529 in mind.

PR #1362, opened 2015-08-10
Removed clientid requirements from the protocol handler.

woogerboy21:
Remove the check for clientid with in the protocol handler since servatrice has the option to check if the clientid is required.


Daenyth:
:shipit:

PR #1366, opened 2015-08-11
Add script to validate/clear invalid country codes in the DB.

### LINKED ISSUE #768: check accounts for wrong/not allowed country codes (e.g. many use NATO)
Many users still have this weird white flags.
I think that's because they used to be able to use their own letters inside the country field.

Some messed around with it to get this "special looking white flag", others just don't know their right code and others used codes that are not working with our international norm we start using to pull the flags.

I checked some players during low-server times (just 500 users!) and still found all this different codes:
- un (**many many!**; unknown?)
- uk (**many!**; uk is United Kingdom (Great Britain) within the old NATO country codes);<br>
 to be honest, I think I never ever saw someone with a GB flag
- sw (**several**; sw is Sweden within the old NATO country codes)
- sp (**several**; sp is Spain within the old NATO country codes)
- ja (ja is Japan within the old NATO country codes)
- no (no is Norway within the old NATO country codes)
- en (en is Estonia within the old NATO country codes)
- ct (ct is Central African Republic within the old NATO country codes)
- wa (wa is Namibia within the old NATO country codes)
- rm (rm is Marshall Islands within the old NATO country codes)
- qc
- ti
- ip
- 29, N7 (??? :) )
<br>*(added more to the list in the meantime)*

**Important:**
As described in https://en.wikipedia.org/wiki/List_of_NATO_country_codes the NATO codes got changed to be the same as the ISO ones we use.
The old ones can be found here, e.g.: https://de.wikipedia.org/wiki/Liste_der_NATO-L%C3%A4ndercodes

I suggest checking the database for codes not conform to the ISO-3166-1 norm we use right now and reset these values. Some kind of script can do this? I wouldn't convert old NATO codes automatically.
@woogerboy21 @ZeldaZach @Daenyth your thoughts?

Our reference is the [ALPHA-2 code from ISO-3166-1](http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).

woogerboy21:
Fix #768 
Script to query the DB users table for distinct country code values, compare the values to images in the resources folder, then if set to correct will clear the accounts country columns containing invalid country code values.

### DIFF
--- a/servatrice/scripts/linux/maint_countrycodes
+++ b/servatrice/scripts/linux/maint_countrycodes
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# THIS SCRIPT EXPECTS TO BE EXECUTED FROM THE GITHUB SOURCE FOLDER PATH STRUCTURE
+# OTHERWISE, UPDATE THE 'COUNTRYCODEIMAGEPATH' TO POINT TO THE FOLDER CONTAINING THE COUNTRY CODE IMAGES
+# USE THIS SCRIPT TO COMPARE EXISTING USER ACCOUNTS TO VALID COUNTRY CODES AND CLEAR INVALID COUNTRY CODE DATA
+
+MODE="report"                       #set this to correct to fix invalid country codes, otherwise it only reports
+DBNAME="servatrice"                 #set this to the database name used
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Seems good to me. Nice job! :shipit:

woogerboy21:
This has already been run on woogerworks main server btw.

PR #1367, opened 2015-08-11
Added lower case country code on correction.

woogerboy21:
Not sure if this should be an issue post but apparently the country code is case sensitive.  Updated the script to change all values in the country column to lowercase if set to "correct".

### DIFF
--- a/servatrice/scripts/linux/maint_countrycodes
+++ b/servatrice/scripts/linux/maint_countrycodes
@@ -27,5 +27,10 @@ do
         fi 
     fi
 done
+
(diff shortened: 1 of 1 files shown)


tooomm:
You might want to add a hint about the problem as well?
Or let the script find uppercase and warn the user when just run with `report`?

ZeldaZach:
Works like this, from what I can tell:
```
US -> us
nF -> nf
de -> de
#N -> 
Un ->
```

Where US, NF, and DE are valid countries

PR #1369, opened 2015-08-11
remove max screen default

### LINKED ISSUE #1360: [Regression] Cockatrice now opens in a maximized window at all times, even when the last-closed instance was not a maximized window.
Before, Cockatrice would start with the exact window size I had it last when I closed it. As of the August 9, 2015 update, Cockatrice now starts maximized no matter what. I am unsure of whether or not this is simply an issue for me, however all I did was update the program, so that shouldn't be the case.

ZeldaZach:
Fix #1360 

Simple fix.


Daenyth:
@marcofernandezheras was there a specific reason you changed it?

PR #1374, opened 2015-08-13
Ban by client ID

### LINKED ISSUE #42: Additional Banning Functionality

woogerboy21:
Fix #42 

This adds the ability to ban users by there generated client id's.  This PR also includes the updated ban checking for client registration to check for a clientid in the ban table of the client trying to register a new account but has been banned under a different username.

The client id ban check function has been written to allow backwards compatibility for users that may have registered / used an older client to connect that does not pass the client id and thus not have the id stored in the database.

### DIFF
--- a/cockatrice/src/userlist.cpp
+++ b/cockatrice/src/userlist.cpp
@@ -36,11 +36,19 @@ BanDialog::BanDialog(const ServerInfo_User &info, QWidget *parent)
     ipBanCheckBox = new QCheckBox(tr("ban &IP address"));
     ipBanCheckBox->setChecked(true);
     ipBanEdit = new QLineEdit(QString::fromStdString(info.address()));
+    idBanCheckBox = new QCheckBox(tr("ban client I&D"));
+    idBanCheckBox->setChecked(true);
+    idBanEdit = new QLineEdit(QString::fromStdString(info.clientid()));
+    if (QString::fromStdString(info.clientid()).isEmpty())
+        idBanCheckBox->setChecked(false);
+
     QGridLayout *banTypeGrid = new QGridLayout;
     banTypeGrid->addWidget(nameBanCheckBox, 0, 0);
     banTypeGrid->addWidget(nameBanEdit, 0, 1);
     banTypeGrid->addWidget(ipBanCheckBox, 1, 0);
     banTypeGrid->addWidget(ipBanEdit, 1, 1);
+    banTypeGrid->addWidget(idBanCheckBox, 2, 0);
+    banTypeGrid->addWidget(idBanEdit, 2, 1);
     QGroupBox *banTypeGroupBox = new QGroupBox(tr("Ban type"));
     banTypeGroupBox->setLayout(banTypeGrid);
     
@@ -110,8 +118,8 @@ BanDialog::BanDialog(const ServerInfo_User &info, QWidget *parent)
 
 void BanDialog::okClicked()
 {
-    if (!nameBanCheckBox->isChecked() && !ipBanCheckBox->isChecked()) {
-        QMessageBox::critical(this, tr("Error"), tr("You have to select a name-based or IP-based ban, or both."));
+    if (!nameBanCheckBox->isChecked() && !ipBanCheckBox->isChecked() && !idBanCheckBox->isChecked()) {
+        QMessageBox::critical(this, tr("Error"), tr("You have to select a name-based, IP-based, clientId based, or some combination of the three to place a ban."));
         return;
     }
     accept();
@@ -125,6 +133,11 @@ void BanDialog::enableTemporaryEdits(bool enabled)
     hoursEdit->setEnabled(enabled);
     minutesLabel->setEnabled(enabled);
     minutesEdit->setEnabled(enabled);
+}
+
+QString BanDialog::getBanId() const
+{
+    return idBanCheckBox->isChecked() ? idBanEdit->text() : QString();
 }
 
 QString BanDialog::getBanName() const
--- a/servatrice/src/servatrice_database_interface.cpp
+++ b/servatrice/src/servatrice_database_interface.cpp
@@ -229,7 +229,7 @@ QChar Servatrice_DatabaseInterface::getGenderChar(ServerInfo_User_Gender const &
     }
 }
 
-AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_ProtocolHandler *handler, const QString &user, const QString &password, QString &reasonStr, int &banSecondsLeft)
+AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_ProtocolHandler *handler, const QString &user, const QString &password, const QString &clientId, QString &reasonStr, int &banSecondsLeft)
 {
     switch (server->getAuthenticationMethod()) {
     case Servatrice::AuthenticationNone: return UnknownUser;
@@ -247,7 +247,7 @@ AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_Prot
         if (!usernameIsValid(user, reasonStr))
             return UsernameInvalid;
 
-        if (checkUserIsBanned(handler->getAddress(), user, reasonStr, banSecondsLeft))
+        if (checkUserIsBanned(handler->getAddress(), user, clientId, reasonStr, banSecondsLeft))
             return UserIsBanned;
 
         QSqlQuery *passwordQuery = prepareQuery("select password_sha512, active from {prefix}_users where name = :name");
@@ -280,7 +280,7 @@ AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_Prot
     return UnknownUser;
 }
 
-bool Servatrice_DatabaseInterface::checkUserIsBanned(const QString &ipAddress, const QString &userName, QString &banReason, int &banSecondsRemaining)
+bool Servatrice_DatabaseInterface::checkUserIsBanned(const QString &ipAddress, const QString &userName, const QString &clientId, QString &banReason, int &banSecondsRemaining)
 {
     if (server->getAuthenticationMethod() != Servatrice::AuthenticationSql)
         return false;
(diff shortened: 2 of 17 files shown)


woogerboy21:
ban_by_clientid

cockatrice/src/userlist.cpp
```diff
@@ -110,8 +115,8 @@ BanDialog::BanDialog(const ServerInfo_User &info, QWidget *parent)
 
 void BanDialog::okClicked()
 {
-    if (!nameBanCheckBox->isChecked() && !ipBanCheckBox->isChecked()) {
-        QMessageBox::critical(this, tr("Error"), tr("You have to select a name-based or IP-based ban, or both."));
+    if (!nameBanCheckBox->isChecked() && !ipBanCheckBox->isChecked() && !idBanCheckBox->isChecked()) {
```
ZeldaZach: Should say a combination. Not all 3

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -291,11 +291,48 @@ bool Servatrice_DatabaseInterface::checkUserIsBanned(const QString &ipAddress, c
     }
 
     return
-        checkUserIsIpBanned(ipAddress, banReason, banSecondsRemaining)
-        || checkUserIsNameBanned(userName, banReason, banSecondsRemaining);
+        checkUserIsIpBanned(ipAddress, banReason, banSecondsRemaining) || checkUserIsNameBanned(userName, banReason, banSecondsRemaining) || checkUserIsIdBanned(clientId, banReason, banSecondsRemaining);;
```
ZeldaZach: Two ;;

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -291,11 +291,48 @@ bool Servatrice_DatabaseInterface::checkUserIsBanned(const QString &ipAddress, c
...
+                    " b.minutes <=> 0,"
+                    " b.visible_reason"
+                    " from {prefix}_bans b"
+                    " where"
+                    " b.time_from = (select max(c.time_from)"
+                    " from {prefix}_bans c"
+                    " where c.clientid = :id)"
+                    " and b.clientid = :id2");
+
+    idBanQuery->bindValue(":id", clientId);
```
ZeldaZach: Are you boy able to just call the "id" twice in the query? I don't get why we need "id2"

servatrice/src/servatrice_database_interface.cpp
woogerboy21: Two? Two what?

cockatrice/src/userlist.cpp
woogerboy21: Why?  You can choose all 3 if you like.  What you cant do is not select any.

servatrice/src/servatrice_database_interface.cpp
woogerboy21: The query is selecting bans based on a time frame.  So there is a start and end time frame that returns the results of the rows that contain the clientid in them.  You need 2 bind value(s) as 2 different resulting time frame queries are being calculated.

cockatrice/src/userlist.cpp
Nightfirecat: I think he means that it can be a name ban, ip ban, id ban, name+ip, name+id, etc etc. rather than only having one or all of them as options.

That is, it can be any combination such that `a || b || c === true`, not a combination such that only `a`, `b`, or `c` is true or `a && b && c` is true, like the text description seems to imply.

woogerboy21: Yeah im updating the message to be more descriptive.

woogerboy21: Dang, still missed an OR to remove in the update.

woogerboy21:
Im pretty sure this should be fixed up, let me know if there is anything else.
@Daenyth if you get a minute could you take a look and see if there is anything you want changed?

cockatrice/src/userlist.cpp
```diff
@@ -36,11 +36,19 @@ BanDialog::BanDialog(const ServerInfo_User &info, QWidget *parent)
     ipBanCheckBox = new QCheckBox(tr("ban &IP address"));
     ipBanCheckBox->setChecked(true);
     ipBanEdit = new QLineEdit(QString::fromStdString(info.address()));
+    idBanCheckBox = new QCheckBox(tr("ban I&D"));
+    idBanCheckBox->setChecked(true);
+    idBanEdit = new QLineEdit(QString::fromStdString(info.clientid()));
+    if (QString::fromStdString(info.clientid()).isEmpty())
+                idBanCheckBox->setChecked(false);
```
ZeldaZach: This is tabbed over a little too much, but not a big deal

cockatrice/src/userlist.cpp
ZeldaZach: Looks good :+1:

cockatrice/src/userlist.cpp
```diff
@@ -36,11 +36,19 @@ BanDialog::BanDialog(const ServerInfo_User &info, QWidget *parent)
     ipBanCheckBox = new QCheckBox(tr("ban &IP address"));
     ipBanCheckBox->setChecked(true);
     ipBanEdit = new QLineEdit(QString::fromStdString(info.address()));
+    idBanCheckBox = new QCheckBox(tr("ban I&D"));
```
Daenyth: "Ban Client ID"?

cockatrice/src/userlist.cpp
```diff
@@ -110,8 +118,8 @@ BanDialog::BanDialog(const ServerInfo_User &info, QWidget *parent)
 
 void BanDialog::okClicked()
 {
-    if (!nameBanCheckBox->isChecked() && !ipBanCheckBox->isChecked()) {
-        QMessageBox::critical(this, tr("Error"), tr("You have to select a name-based or IP-based ban, or both."));
+    if (!nameBanCheckBox->isChecked() && !ipBanCheckBox->isChecked() && !idBanCheckBox->isChecked()) {
+        QMessageBox::critical(this, tr("Error"), tr("You have to select a name-based, IP-based, clientId based, or some combination of the three to place a ban."));
```
Daenyth: You have to select at least one of... ?

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -247,7 +247,7 @@ AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_Prot
         if (!usernameIsValid(user, reasonStr))
             return UsernameInvalid;
 
-        if (checkUserIsBanned(handler->getAddress(), user, reasonStr, banSecondsLeft))
+        if (checkUserIsBanned(handler->getAddress(), user, reasonStr, banSecondsLeft, clientId))
```
Daenyth: Can you reorder the parameters to `checkUserIsBanned` so that the out params are at the end?

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -291,11 +291,48 @@ bool Servatrice_DatabaseInterface::checkUserIsBanned(const QString &ipAddress, c
     }
 
     return
-        checkUserIsIpBanned(ipAddress, banReason, banSecondsRemaining)
-        || checkUserIsNameBanned(userName, banReason, banSecondsRemaining);
+        checkUserIsIpBanned(ipAddress, banReason, banSecondsRemaining) || checkUserIsNameBanned(userName, banReason, banSecondsRemaining) || checkUserIsIdBanned(clientId, banReason, banSecondsRemaining);
```
Daenyth: We may actually want to not short circuit here - consider:
If a by-ip ban exists for 10 seconds, and a by-name ban exists for 20 seconds, you will tell the user they are banned for 20 seconds.

You don't need to address that in this PR I think, it's not a new behavior, but we should ticket it for future improvement

servatrice/src/servatrice_database_interface.cpp
Daenyth: I think what Zach means is, can we reuse `:id` so that it's `where c.clientid = :id and b.clientid = :id` and bind `:id` once?

I don't know the answer.

cockatrice/src/userlist.cpp
ZeldaZach: I'm fine with the change already done

servatrice/src/serversocketinterface.cpp
```diff
@@ -756,18 +756,33 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
     sqlInterface->execSqlQuery(query);
 
     servatrice->clientsLock.lockForRead();
     QList<ServerSocketInterface *> userList = servatrice->getUsersWithAddressAsList(QHostAddress(address));
     ServerSocketInterface *user = static_cast<ServerSocketInterface *>(server->getUsers().value(userName));
+
+    if (userName.isEmpty() && address.isEmpty()) {
+        QSqlQuery *query = sqlInterface->prepareQuery("select name from {prefix}_users where clientid = :client_id");
+        query->bindValue(":client_id", QString::fromStdString(cmd.clientid()));
+        sqlInterface->execSqlQuery(query);
```
Daenyth: I think we need to check for query errors here.

servatrice/src/serversocketinterface.cpp
```diff
@@ -756,18 +756,33 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
+        while (query->next()){
+            userName = query->value(0).toString();
+            ServerSocketInterface *user = static_cast<ServerSocketInterface *>(server->getUsers().value(userName));
+            if (user && !userList.contains(user))
```
Daenyth: If you convert the QList to a QSet then you don't need to do this - you can just `add` it whether or not it was already there.

servatrice/src/servatrice_database_interface.cpp
woogerboy21: Honestly the logic flow for banning as a whole could use refactoring.  But I do see your point in this.

servatrice/src/servatrice_database_interface.cpp
woogerboy21: Oh, good question.  I don't know either.  I've always done it with separate bind place holders.

servatrice/src/serversocketinterface.cpp
woogerboy21: And if error do.....?

woogerboy21:
@Daenyth I believe I have addressed everything (except the QList/QSet comment).  I did work in more logic in that block of code though.

poixen:
Jumping on this a bit late. If a user is banned and then updates to a new client, can they get around the ban?

Daenyth:
The client id should be stable between versions as long as we don't change
the algorithm. They could downgrade or use a modified client of course, not
much we can do about that.

woogerboy21:
The way I see it, all we can do is place road blocks to make trolls work harder to be trolls.  We will always have those 1%ers out there that will most likely be a head of the game.  So far all the techniques put in place (ban by name/ip, reg only server, now clientid's) are slowly weeding out the out the people that don't fall into that 1% mark.  Once we can get the pool of trolls down to just a few user tools such as the ignore user functionality will become much more viable and accepted.

There may be ways to even locate the people that hack the client, but not sure its worth the effort to work that in just yet since there are plenty of other area's that need work.

PR #1376, opened 2015-08-14
Add option to open customsets folder

Fizztastic:
untitled-1


ZeldaZach:
Tested and good to go on my end :shipit:

Daenyth:
Can we try using xdg-open for Linux? It probably shouldn't hold up the merge though.

We need to make a common function for desktop directory, we shouldn't have these ifdef all over. That should be another pr

tooomm:
if it doesn't hold up this pr --> get the stuff into issue(s), please! ;)

PR #1378, opened 2015-08-15
Add last_login column to user table and populate upon login

woogerboy21:
This will allow operators to keep a cleaner user base.  Even though the sessions table contains the users login/logout time the sessions table is commonly removed / cleared and should be considered volatile causing the loss of important last login information.  This resolves that issue.

This change is strictly a server side functionality addition and wont require any type of client side update.


ZeldaZach:
After reviewing the code base, it looks fine to me. Since this is server-only, no reason why to hold this up. :Shipit:

Issue #169, opened 2014-07-01
Add a moderator-only chatroom with chat history


Daenyth:
cc @woogerboy21 @beusapu

PR #1379, opened 2015-08-15
Add permission levels to rooms

woogerboy21:
This change adds definable permission levels to rooms allowing for rooms to be "registered only" or "moderator only" or "admin only" room access.  This will allow for moderator only type rooms mentioned in #169 .  It does not add chat history.  

Client side code updates are only required if the user wishes to see the message dialogue box that indicates why they are unable to join a room.  Otherwise older clients will just appear to do nothing when the join button on a room they do not have permissions to join is selected.

### DIFF
--- a/cockatrice/src/tab_server.cpp
+++ b/cockatrice/src/tab_server.cpp
@@ -25,7 +25,7 @@ RoomSelector::RoomSelector(AbstractClient *_client, QWidget *parent)
 {
     roomList = new QTreeWidget;
     roomList->setRootIsDecorated(false);
-    roomList->setColumnCount(4);
+    roomList->setColumnCount(5);
     roomList->header()->setStretchLastSection(false);
 #if QT_VERSION < 0x050000
     roomList->header()->setResizeMode(0, QHeaderView::ResizeToContents);
@@ -35,6 +35,7 @@ RoomSelector::RoomSelector(AbstractClient *_client, QWidget *parent)
 #else
     roomList->header()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
     roomList->header()->setSectionResizeMode(1, QHeaderView::Stretch);
+    roomList->header()->setSectionResizeMode(1, QHeaderView::ResizeToContents);
     roomList->header()->setSectionResizeMode(2, QHeaderView::ResizeToContents);
     roomList->header()->setSectionResizeMode(3, QHeaderView::ResizeToContents);
 #endif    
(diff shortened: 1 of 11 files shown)


woogerboy21:
roomslist

qmessage

common/server_protocolhandler.cpp
```diff
@@ -542,6 +542,24 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
     if (!r)
         return Response::RespNameNotFound;
 
+    QString roomPermission = r->getRoomPermission().toLower();
+    if (roomPermission != "none"){
+        if (roomPermission == "registered") {
```
ZeldaZach: This should be a switch > if statements imo

servatrice/servatrice.ini.example
```diff
@@ -172,6 +172,9 @@ roomlist\1\name="General room"
 ; Room description for the room number 1
 roomlist\1\description="Play anything here."
 
+; Rooms can restrict the level of user that can join.  Current supported options are none,registered,moderator,administrator
```
ZeldaZach: Should put spaces as such: `none, registered, moderator, administrator`

common/pb/serverinfo_room.proto
```diff
@@ -6,10 +6,11 @@ message ServerInfo_Room {
 	optional sint32 room_id = 1;
 	optional string name = 2;
 	optional string description = 3;
-	optional uint32 game_count = 4;
-	optional uint32 player_count = 5;
-	optional bool auto_join = 6;
-	repeated ServerInfo_Game game_list = 7;
-	repeated ServerInfo_User user_list = 8;
-	repeated ServerInfo_GameType gametype_list = 9;
+	optional string permissionlevel = 4;
```
ZeldaZach: Does bumping every other protocol's number like this impact anything?

common/server_protocolhandler.cpp
ZeldaZach: Also, you should probably save `userInfo->user_level()` as a variable since you call it multiple times in this block

common/pb/serverinfo_room.proto
woogerboy21: From what I can tell it shouldn't.  All the protocol parts are referenced/set by there descriptive name and not number.  The only thing that uses the number is the actual protocol buffer inside the generated code.
I actually debated on if this should have been added to.

common/server_protocolhandler.cpp
woogerboy21: Switch statements in C++ do not support strings (as we described in chat).  An enum may be feasible. With the way its currently written its only called once.  The sub comparison only occurs dependent on which roomPermission level is set. 

Want to get others opinions on the enum or if the if statements should be kept.

cockatrice/src/tab_server.cpp
```diff
@@ -130,10 +138,16 @@ void RoomSelector::joinClicked()
 
 void RoomSelector::joinFinished(const Response &r, const CommandContainer & /*commandContainer*/, const QVariant &extraData)
 {
-    if (r.response_code() != Response::RespOk)
-        return;
+    if (r.response_code() != Response::RespOk) {
```
ZeldaZach: I think you can add this as a case in the switch to reduce the number of lines and complexity

common/server_protocolhandler.cpp
Daenyth: Enum would be good, I agree. I think it's something we will likely revisit later so it's not such a big deal

Daenyth: Actually if you use userlevel like the user info does it might be cleaner.

woogerboy21: I started to do it with the userlevel but noticed through out the existing code it uses this method of comparison and figured I wanted to stick with the style that was already present.

Daenyth: K

PR #1381, opened 2015-08-16
Country to lower (fix win bug)

ZeldaZach:
Bug reported by @tooomm and @woogerboy21.

User country flag was sometimes set with the wrong case. This PR makes sure the country code is always lower case (except when you see it on somebody's profile, then it's upper).

### DIFF
--- a/cockatrice/src/dlg_edit_user.cpp
+++ b/cockatrice/src/dlg_edit_user.cpp
@@ -35,7 +35,7 @@ DlgEditUser::DlgEditUser(QWidget *parent, QString email, int gender, QString cou
     foreach(QString c, countries)
     {
         countryEdit->addItem(QPixmap(":/resources/countries/" + c + ".svg"), c);
-        if(c == country)
+        if (c == country)
             countryEdit->setCurrentIndex(i);
 
         ++i;
--- a/cockatrice/src/dlg_register.cpp
+++ b/cockatrice/src/dlg_register.cpp
@@ -62,9 +62,7 @@ DlgRegister::DlgRegister(QWidget *parent)
     countryEdit->setCurrentIndex(0);
     QStringList countries = settingsCache->getCountries();
     foreach(QString c, countries)
-    {
         countryEdit->addItem(QPixmap(":/resources/countries/" + c + ".svg"), c);
-    }
 
     realnameLabel = new QLabel(tr("Real name:"));
     realnameEdit = new QLineEdit();
--- a/cockatrice/src/pixmapgenerator.cpp
+++ b/cockatrice/src/pixmapgenerator.cpp
@@ -119,7 +119,7 @@ QPixmap CountryPixmapGenerator::generatePixmap(int height, const QString &countr
     if (pmCache.contains(key))
         return pmCache.value(key);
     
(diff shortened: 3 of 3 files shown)


Daenyth:
Do we have a central resource related class? You're repeating the same logic in multiple places, we could put it in a method

tooomm:
@woogerboy21 do we need/want a server side `.toLower()` as well?
Would prevent older clients from writing bad values into the db

ZeldaZach:
@tooomm that was my reasoning

cockatrice/src/dlg_edit_user.cpp
```diff
@@ -34,8 +34,8 @@ DlgEditUser::DlgEditUser(QWidget *parent, QString email, int gender, QString cou
     int i = 1;
     foreach(QString c, countries)
     {
-        countryEdit->addItem(QPixmap(":/resources/countries/" + c + ".svg"), c);
-        if(c == country)
+        countryEdit->addItem(QPixmap(":/resources/countries/" + c.toLower() + ".svg"), c);
```
ctrlaltca: I don't see the point in calling .toLower() here, since the values of ```c``` comes from ```SettingsCache::getCountries()```, that returns hardcoded lowercase values.

cockatrice/src/dlg_edit_user.cpp
```diff
@@ -34,8 +34,8 @@ DlgEditUser::DlgEditUser(QWidget *parent, QString email, int gender, QString cou
...
+        if (c == country)
```
ctrlaltca: Here the value of "country" arrives from the server, sowe probably need to ensure it's lowercase (or make the test case insensitive)

cockatrice/src/dlg_register.cpp
```diff
@@ -62,9 +62,7 @@ DlgRegister::DlgRegister(QWidget *parent)
     countryEdit->setCurrentIndex(0);
     QStringList countries = settingsCache->getCountries();
     foreach(QString c, countries)
-    {
-        countryEdit->addItem(QPixmap(":/resources/countries/" + c + ".svg"), c);
-    }
+        countryEdit->addItem(QPixmap(":/resources/countries/" + c.toLower() + ".svg"), c);
```
ctrlaltca: Same as before: ```c``` values are hardcoded lowercase, no need to lower them

woogerboy21: The issue that was experienced is the characters in the DB were being
written first character in upper. But I couldn't reproduce so I have no
idea how it was happening.

ZeldaZach: The fix provided does account for the issues that we can't reproduce... but I couldn't even find out why this was happening.

Daenyth: Could the upper values just be left over from when the db was text entry instead of a dropdown?

Daenyth:
:shipit: after ctrl's comments are addressed.

Daenyth:
ping @ctrlaltca

ctrlaltca:
This surely fixes flag images loading in the client if somehow the server sent an uppercase/mixed case country code (eg. a bad value is container in the database). :+1:

PR #1382, opened 2015-08-16
Customizable shortcuts

### LINKED ISSUE #130: Catalogue and overhaul keyboard shortcut system.
It's especially bad on OSX since a lot of the normal shortcut codes are taken by the OS

### LINKED ISSUE #691: Change Attach to Card shortcut (Ctrl+A) to something else.
**Platform:** Windows 7 64-bit.
**Version:** Every version.

This is not a regression, I'm sure I've never seen it work in multiplayer.

The shortcut is Ctrl+A, it's probably conflicting with the "Select All text" shortcut common to most Windows applications.

I don't know why it works in Local games only, but even then it won't work if the chat text box has focus. (Related #653)

### LINKED ISSUE #1278: Change reveal top card from ctrl + n to something not next to mull..

marcofernandezheras:
Fix #130 & Fix #1278 & Fix #691

shorcutsfinal

Added the possibility to customize shortcut. This should fix most of the shortcuts related issues.
Also added some new shortcuts.


woogerboy21:
This is cool, still need to review but the first thing that I would ask is what size is this screen?  Looks rather large which may cause issues on users machines with lower resolution.

marcofernandezheras:
After this commit, setting dialog is about 1060 width x 710 heigth

cockatrice/CMakeLists.txt
```diff
@@ -94,6 +94,9 @@ SET(cockatrice_SOURCES
     src/qt-json/json.cpp
     src/soundengine.cpp
     src/pending_command.cpp
+    src/shortcutssettings.cpp
+    src/sequenceEdit/secuenceedit.cpp
```
ZeldaZach: `Sequence` [original spelling is a bit off]

Daenyth:
Can you populate the setting boxes with the default binding if the user hasn't set anything? And is it possible with this to unbind completely?

marcofernandezheras:
Is the current behavior @Daenyth

cockatrice/src/sequenceEdit/secuenceedit.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <QLineEdit>
+#include <QPushButton>
+#include <QHBoxLayout>
+#include <QEvent>
+#include <QKeyEvent>
+#include <QToolTip>
+
+SecuenceEdit::SecuenceEdit(QString name, QWidget *parent) : QWidget(parent)
+{
+    this->shorcutName = name;
```
ctrlaltca: is "this->" needed? i can see the ctor in .h is defined as ``` SecuenceEdit(QString shorcutName, QWidget *parent = 0);```, but here the first parameter has been renamed as "name", so you should not have problems with the scope of the variable.
In other places in the code, i saw this pattern when dealing with such a case:
```
void Foo::bar(QString _variable) {
    variable = _variable;
}
```
or, in this specific case of a ctor: 
```
void Foo::Foo(QString _variable)
: variable(_variable)
 {
...
}
```

cockatrice/src/shortcutssettings.cpp
```diff
@@ -0,0 +1,248 @@
...
+                if(storedSecuence.contains(stringSecuences.at(j)))
+                    return false;
+            }
+        }
+    }
+    return true;
+}
+
+void ShortcutsSettings::fillDefaultShorcuts()
+{
```
ctrlaltca: I see default shortcuts are defined here, but used only when the "restore default shortcut" button is pressed, eg:
```
+    defaultShortCuts["MainWindow/aFullScreen"] = parseSecuenceString("Ctrl+F");
```
Then, in every call to ```settingsCache->shortcuts().getShortcut()``` the default shortcut is passed as the second parameter, eg:
```
+    aFullScreen->setShortcuts(settingsCache->shortcuts().getShortcut(
+                                  "MainWindow/aFullScreen", QKeySequence("Ctrl+F")));
```
Could the getShortcut() method use the defaults defined here? This would avoid the possible human error of adding/changing a default shortcut in one place forgetting to update the other place.

ctrlaltca:
I see minor issues, but the general idea is a big thumb up!

cockatrice/src/main.cpp
```diff
@@ -115,7 +115,7 @@ QString const generateClientID()
 int main(int argc, char *argv[])
 {
     QApplication app(argc, argv);
-
+    //app.setStyle("Fusion");
```
ctrlaltca: remove this?

tooomm:
this should be fixing #130 ?

ctrlaltca:
Related: #691 #1278

cockatrice/src/sequenceEdit/secuenceedit.cpp
```diff
@@ -0,0 +1,150 @@
...
+    keys = 0;
+    valid = false;
+
+    lineEdit = new QLineEdit(this);
+    clearButton = new QPushButton("", this);
+    defaultButton = new QPushButton("", this);
+
+    lineEdit->setMinimumWidth(100);
+    clearButton->setMaximumWidth(lineEdit->height());
+    defaultButton->setMaximumWidth(lineEdit->height());
```
ctrlaltca: These two lines cause a problem at least under osx; the buttons are overlapping:
schermata 2015-08-18 alle 22 46 08
Commenting out these lines fixes the problem:
schermata 2015-08-18 alle 22 47 00

ctrlaltca:
> This is cool, still need to review but the first thing that I would ask is what size is this screen? Looks rather large which may cause issues on users machines with lower resolution.

I agree with this. The problem is that, at least under osx, Qt forces very wide margins and spacing between widgets, so the windows doesn't fit in the screen. This is a screenshot from a Macbook pro 13" (1280x800), where the window is unusable:

schermata 2015-08-18 alle 22 51 48

Nightfirecat:
It might make the UI a little more click-heavy, but it might hel pto move each of the widgets to their own tab?

cockatrice/src/sequenceEdit/secuenceedit.cpp
ZeldaZach: Same issue here on a larger mac

Daenyth: What if each grouping of shortcuts was it's own tab inside that panel?

cockatrice/src/window_main.cpp
```diff
@@ -456,23 +456,54 @@ void MainWindow::retranslateUi()
...
+
     aSettings->setText(tr("&Settings..."));
+
+
+
+
     aExit->setText(tr("&Exit"));
 
+
+
```
ctrlaltca: why all these empty lines between line 459 and line 497? Just being picky...

ctrlaltca:
Window size is good now;
button overlap has been fixed, too;
shortcuts are updated on the fly when changed in the preferences pane.
Big thumb up for me :+1:

tooomm:
@marcofernandezheras as @ctrlaltca wrote, #1278 & #691 are just referenced for being related.
They don't get necessarily fixed with this. Just because people can change shortcuts for themselves, we still might want to have a better/other default.

If you keep them in your starting post they get automatically closed on merge and we miss the change to improve the default behavior of the client. (if needed, but that's up for discussion in the tickets)

ZeldaZach:
@Tooomm I added the closure of the tickets, not Marco. I think they're fine to be closed by this as people can set their own and it becomes less of a haste to fix in the future.

tooomm:
@ZeldaZach Oh. Haha. :)
I still think it's a huge difference if we let users fix a bad client behavior for themselves with a new tool we include, or if we provide a reasonable default experience. But this needs to be discussed in the related tickets. ;)
If many users end up changing a badly chosen shortcut, we should think about an alternative - no matter if people can easily change it anyway.

*For example the Ctrl+M/N ones seems to be interacting with each other in a bad way easily. And I personally would prefer to change one of them. Taking a mulligan by fault is always game breaking and you can't undo it! :(*

ZeldaZach:
@tooomm I personally think leaving the defaults as-is is best as people complain whenever things are changed.

While I do agree a mid-game mulligan ends the game essentially, maybe we can look into preventing the mulligan command from being called if there are cards in play? (I'll leave this on the other ticket)

cockatrice/src/sequenceEdit/ui_shortcutstab.h
```diff
@@ -0,0 +1,1597 @@
...
+    QLabel *lbl_Player_aDrawCards;
+    SequenceEdit *Player_aDrawCards;
+    QLabel *lbl_Player_aUndoDraw;
+    SequenceEdit *Player_aUndoDraw;
+    QLabel *lbl_Player_aAlwaysRevealTopCard;
+    SequenceEdit *Player_aAlwaysRevealTopCard;
+    QSpacerItem *verticalSpacer_3;
+    QWidget * tab_4;
+
+    void setupUi(QWidget *shortcutsTab)
```
ZeldaZach: My only concern with this is that it will be very difficult to change the layout in the future as it's all hand done

cockatrice/src/sequenceEdit/ui_shortcutstab.h
```diff
@@ -0,0 +1,1597 @@
...
+
+        tabWidget->addTab(tab_4, QString());
+
+        gridLayout_9->addWidget(tabWidget, 0, 0, 1, 1);
+
+
+        retranslateUi(shortcutsTab);
+
+        tabWidget->setCurrentIndex(0);
+
```
ZeldaZach: A lot of extra space, but no reason to hold the PR back

cockatrice/src/tab_game.cpp
```diff
@@ -148,6 +148,58 @@ void DeckViewContainer::updateSideboardLockButtonText()
...
+            case 9: temp->setShortcuts(settingsCache->shortcuts().getShortcut("Player/phase9")); break;
+            case 10:temp->setShortcuts(settingsCache->shortcuts().getShortcut("Player/phase10")); break;
+            default: ;
+        }
+    }
+
+    if (aNextPhase) {
+        aNextPhase->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aNextPhase"));
+    }
+    if (aNextTurn) {
```
ZeldaZach: Would changing these `if` statements to `else if` statements improve speed?

cockatrice/src/window_main.cpp
ZeldaZach: ^ I agree, but it's nothing to hold the PR back on

cockatrice/src/tab_game.cpp
marcofernandezheras: You can't change them to else if... those if are null check for each object (don't know if needed, C&P from the original code)

Daenyth:
Let's :shipit: and see what people think, we can keep tweaking as it goes

ZeldaZach:
This is good to go and will be merged shortly :shipit:

PR #1386, opened 2015-08-16
Fix linked cardwidget not showing PT/Loyalty

### LINKED ISSUE #1383: Power/Toughness when tagging a creature name?
I noticed that we don't have yet the viewing of a creature's Power/Toughness when someone tags a creature name. For example, if I type in the chat [[Dark Confidant]], besides the image itself you can't really tell what the power/toughness is on the creature (2/1). This would be useful because if gatherer or whatever database we use goes down, we won't see the card image but we will at least see all the characteristics of the card.

Fizztastic:
Fix #1383


ctrlaltca:
Works; we should probably refactor ```minimized``` as an enum, but by now this PR is sufficient to fix the issue.

PR #1389, opened 2015-08-18
Corrected miss-matched schema versions

woogerboy21:
Servatrice schema version should be 7 (not 6) for current commit level.  Not sure exactly why it wasn't updated.


ctrlaltca:
references 487cbb23cbddb9907f37575a444eca7909628087

Issue #190, opened 2014-07-05
In client reporting


ZeldaZach:
There should be a way to quickly report somebody from within the client instead of having to go to the woogerworks site, logging in, then filing the report

ZeldaZach:
I started working on the front end, but I don't know enough about the protocol to do much: https://github.com/Cockatrice/Cockatrice/compare/master...ZeldaZach:report_user

ZeldaZach:
cc: @woogerboy21

woogerboy21:
Far from finished , but here is the start of this for you @ZeldaZach 
Still needs:
* Client side code to specify report reason (not sure if you want simple or complicated window ui)
* Server side database table for storing data
* Server side storing of data in database
* Server side notification of reports (maybe server message similar to ban/warn?)
* Client side admin/mod functionality to view / edit / close reports
* ... possibly other things im overlooking.

PR #1391, opened 2015-08-19
Allows admins to promote users to mod and demote mod to user

### LINKED ISSUE #1377: Add an admin-level command to change user level.
Admins should be able to +mod registered users and -mod moderators.

woogerboy21:
Fix #1377 

This PR adds the ability for administrators to promote a user to moderator status, or demote a moderator to user status.  A new right click context menu item has been added when a moderator right clicks on the username from the user list (as long as the admin tools are  unlocked).

Things to point out about this change.
* A extension to the session event protocol has been put in place.
* Right click menu option evaluates users current position and only adds promote or demote
* When a user is promoted, the user is sent a notification about so mentioning they need to log out / in
* When a mod is demoted, the moderator is immediately removed from the server.
* Admin performing the command always gets notified of the commands results (success/fail)
* The new event notifications logic has been put in place to also accommodate #190.

(Items not pictured are the context menu items and failure popup boxes if the commands fail.)

Edit:
Please ignore the commit message, it mis-states "moderators to promote".  It should read "Admins to promote".  Moderators do not get the menu options and the new menu items are located in the processExtendedAdminCommands function.

### DIFF
--- a/cockatrice/src/tab_supervisor.cpp
+++ b/cockatrice/src/tab_supervisor.cpp
@@ -1,3 +1,6 @@
+#include <QDebug>
+#include <QPainter>
+#include <QMessageBox>
 #include <QApplication>
 #include "tab_supervisor.h"
 #include "abstractclient.h"
@@ -13,14 +16,12 @@
 #include "pixmapgenerator.h"
 #include "userlist.h"
 #include "settingscache.h"
-#include <QDebug>
-#include <QPainter>
-#include <QMessageBox>
 
 #include "pb/room_commands.pb.h"
 #include "pb/room_event.pb.h"
 #include "pb/game_event_container.pb.h"
 #include "pb/event_user_message.pb.h"
+#include "pb/event_notify_user.pb.h"
 #include "pb/event_game_joined.pb.h"
 #include "pb/serverinfo_user.pb.h"
 #include "pb/serverinfo_room.pb.h"
@@ -90,6 +91,7 @@ TabSupervisor::TabSupervisor(AbstractClient *_client, QWidget *parent)
     connect(client, SIGNAL(gameJoinedEventReceived(const Event_GameJoined &)), this, SLOT(gameJoined(const Event_GameJoined &)));
     connect(client, SIGNAL(userMessageEventReceived(const Event_UserMessage &)), this, SLOT(processUserMessageEvent(const Event_UserMessage &)));
     connect(client, SIGNAL(maxPingTime(int, int)), this, SLOT(updatePingTime(int, int)));
+    connect(client, SIGNAL(notifyUserEventReceived(const Event_NotifyUser &)), this, SLOT(processNotifyUserEvent(const Event_NotifyUser &)));
     
     retranslateUi();
 }
@@ -559,3 +561,13 @@ bool TabSupervisor::getAdminLocked() const
         return true;
     return tabAdmin->getLocked();
 }
+
+void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
+{
+    switch ((Event_NotifyUser::NotificationType) event.type()) {
+        case Event_NotifyUser::PROMOTED: QMessageBox::information(this, tr("Promotion"), tr("You have been promoted to moderator. Please log out and back in for changes to take effect.")); break;
+        default: ;
+    }
+
+}
+
--- a/cockatrice/src/user_context_menu.cpp
+++ b/cockatrice/src/user_context_menu.cpp
@@ -1,5 +1,6 @@
 #include <QAction>
 #include <QMenu>
+#include <QMessageBox>
 #include "user_context_menu.h"
 #include "tab_supervisor.h"
 #include "tab_userlists.h"
@@ -31,6 +32,8 @@ UserContextMenu::UserContextMenu(const TabSupervisor *_tabSupervisor, QWidget *p
     aRemoveFromIgnoreList = new QAction(QString(), this);
     aKick = new QAction(QString(), this);
     aBan = new QAction(QString(), this);
+    aPromoteToMod = new QAction(QString(), this);
+    aDemoteFromMod = new QAction(QString(), this);
 
     retranslateUi();
 }
@@ -46,6 +49,8 @@ void UserContextMenu::retranslateUi()
     aRemoveFromIgnoreList->setText(tr("Remove from &ignore list"));
     aKick->setText(tr("Kick from &game"));
     aBan->setText(tr("Ban from &server"));
+    aPromoteToMod->setText(tr("&Promote user to moderator"));
+    aDemoteFromMod->setText(tr("Dem&ote user from moderator"));
 }
 
 void UserContextMenu::gamesOfUserReceived(const Response &resp, const CommandContainer &commandContainer)
@@ -87,6 +92,27 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
     BanDialog *dlg = new BanDialog(response.user_info(), static_cast<QWidget *>(parent()));
     connect(dlg, SIGNAL(accepted()), this, SLOT(banUser_dialogFinished()));
     dlg->show();
+}
+
+void UserContextMenu::adjustMod_processUserResponse(const Response &resp, const CommandContainer &commandContainer)
+{
+
+    const Command_AdjustMod &cmd = commandContainer.admin_command(0).GetExtension(Command_AdjustMod::ext);
+
+    if (resp.response_code() == Response::RespOk) {
+        if (cmd.should_be_mod()) {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Success"), tr("Successfully promoted user."));
+        } else {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Success"), tr("Successfully demoted user."));
+        }
+
+    } else {
+        if (cmd.should_be_mod()) {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Failed"), tr("Failed to promote user."));
+        } else {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Failed"), tr("Failed to demote user."));
+        }
+    }
 }
 
 void UserContextMenu::banUser_dialogFinished()
@@ -132,6 +158,14 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
     if (!tabSupervisor->getAdminLocked()) {
         menu->addSeparator();
         menu->addAction(aBan);
+
+        menu->addSeparator();
+        if (userLevel.testFlag(ServerInfo_User::IsModerator) && (tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsAdmin)) {
+            menu->addAction(aDemoteFromMod);
+
+        } else if (userLevel.testFlag(ServerInfo_User::IsRegistered) && (tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsAdmin)) {
+            menu->addAction(aPromoteToMod);
+        }
     }
     bool anotherUser = userName != QString::fromStdString(tabSupervisor->getUserInfo()->name());
     aDetails->setEnabled(online);
@@ -143,6 +177,8 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
     aRemoveFromIgnoreList->setEnabled(anotherUser);
     aKick->setEnabled(anotherUser);
     aBan->setEnabled(anotherUser);
+    aPromoteToMod->setEnabled(anotherUser);
+    aDemoteFromMod->setEnabled(anotherUser);
 
     QAction *actionClicked = menu->exec(pos);
     if (actionClicked == aDetails) {
@@ -186,6 +222,7 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
     } else if (actionClicked == aKick) {
         Command_KickFromGame cmd;
         cmd.set_player_id(playerId);
+
         game->sendGameCommand(cmd);
     } else if (actionClicked == aBan) {
         Command_GetUserInfo cmd;
@@ -193,7 +230,24 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
 
         PendingCommand *pend = client->prepareSessionCommand(cmd);
         connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(banUser_processUserInfoResponse(Response)));
+        client->sendCommand(pend);
+    } else if (actionClicked == aPromoteToMod) {
+        Command_AdjustMod cmd;
+        cmd.set_user_name(userName.toStdString());
+        cmd.set_should_be_mod(true);
 
+        // client->sendCommand(client->prepareAdminCommand(cmd));
+        PendingCommand *pend = client->prepareAdminCommand(cmd);
+        connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(adjustMod_processUserResponse(Response,CommandContainer)));
+        client->sendCommand(pend);
+    } else if (actionClicked == aDemoteFromMod) {
+        Command_AdjustMod cmd;
+        cmd.set_user_name(userName.toStdString());
+        cmd.set_should_be_mod(false);
+
+        // client->sendCommand(client->prepareAdminCommand(cmd));
+        PendingCommand *pend = client->prepareAdminCommand(cmd);
+        connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(adjustMod_processUserResponse(Response,CommandContainer)));
         client->sendCommand(pend);
     }
 
--- a/cockatrice/src/user_context_menu.h
+++ b/cockatrice/src/user_context_menu.h
@@ -27,10 +27,12 @@ private:
     QAction *aAddToIgnoreList, *aRemoveFromIgnoreList;
     QAction *aKick;
     QAction *aBan;
+    QAction *aPromoteToMod, *aDemoteFromMod;
 signals:
     void openMessageDialog(const QString &userName, bool focus);
 private slots:
     void banUser_processUserInfoResponse(const Response &resp);
+    void adjustMod_processUserResponse(const Response &resp, const CommandContainer &commandContainer);
     void banUser_dialogFinished();
     void gamesOfUserReceived(const Response &resp, const CommandContainer &commandContainer);
 public:
--- a/common/pb/admin_commands.proto
+++ b/common/pb/admin_commands.proto
@@ -3,6 +3,7 @@ message AdminCommand {
 		UPDATE_SERVER_MESSAGE = 1000;
 		SHUTDOWN_SERVER = 1001;
 		RELOAD_CONFIG = 1002;
+		ADJUST_MOD = 1003;
 	}
 	extensions 100 to max;
 }
@@ -26,3 +27,12 @@ message Command_ReloadConfig {
 		optional Command_ReloadConfig ext = 1002;
 	}
 }
+
+message Command_AdjustMod {
+	extend AdminCommand {
+		optional Command_AdjustMod ext = 1003;
+	}
+	required string user_name = 1;
+	required bool should_be_mod = 2;
+}
+
--- a/common/pb/event_connection_closed.proto
+++ b/common/pb/event_connection_closed.proto
@@ -11,6 +11,7 @@ message Event_ConnectionClosed {
 		BANNED = 4;
 		USERNAMEINVALID = 5;
 		USER_LIMIT_REACHED = 6;
+		DEMOTED = 7;
 	}
 	optional CloseReason reason = 1;
 	optional string reason_str = 2;
--- a/common/pb/event_notify_user.proto
+++ b/common/pb/event_notify_user.proto
@@ -1,0 +1,14 @@
+import "session_event.proto";
+
+message Event_NotifyUser {
+
+	enum NotificationType {
+        PROMOTED = 1;
+    }
+
+	extend SessionEvent {
+		optional Event_NotifyUser ext = 1010;
+	}
+	optional NotificationType type = 1;
+
+}
--- a/common/pb/response_adjust_mod.proto
+++ b/common/pb/response_adjust_mod.proto
@@ -1,0 +1,7 @@
+import "response.proto";
+
+message Response_AdjustMod{
+	extend Response {
+		optional Response_AdjustMod ext = 1011;
+	}
+}
--- a/servatrice/src/serversocketinterface.cpp
+++ b/servatrice/src/serversocketinterface.cpp
@@ -49,6 +49,7 @@
 #include "pb/event_server_identification.pb.h"
 #include "pb/event_add_to_list.pb.h"
 #include "pb/event_remove_from_list.pb.h"
+#include "pb/event_notify_user.pb.h"
 #include "pb/response_deck_list.pb.h"
 #include "pb/response_deck_download.pb.h"
 #include "pb/response_deck_upload.pb.h"
@@ -309,6 +310,7 @@ Response::ResponseCode ServerSocketInterface::processExtendedAdminCommand(int cm
         case AdminCommand::SHUTDOWN_SERVER: return cmdShutdownServer(cmd.GetExtension(Command_ShutdownServer::ext), rc);
         case AdminCommand::UPDATE_SERVER_MESSAGE: return cmdUpdateServerMessage(cmd.GetExtension(Command_UpdateServerMessage::ext), rc);
         case AdminCommand::RELOAD_CONFIG: return cmdReloadConfig(cmd.GetExtension(Command_ReloadConfig::ext), rc);
+        case AdminCommand::ADJUST_MOD: return cmdAdjustMod(cmd.GetExtension(Command_AdjustMod::ext), rc);
         default: return Response::RespFunctionNotAllowed;
     }
 }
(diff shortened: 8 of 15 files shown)


woogerboy21:
screenshot from 2015-08-19 15 47 59
screenshot from 2015-08-19 15 48 08
screenshot from 2015-08-19 15 48 35
screenshot from 2015-08-19 15 48 39

cockatrice/src/tab_supervisor.cpp
```diff
@@ -559,3 +561,19 @@ bool TabSupervisor::getAdminLocked() const
         return true;
     return tabAdmin->getLocked();
 }
+
+void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
+{
+    QString notificationTitle("Notification");
+    QString notificationMessage;
+    switch ((Event_NotifyUser::NotificationType) event.type()) {
+        case Event_NotifyUser::PROMOTIONSUCCESS: QMessageBox::information(this, tr("Success"), tr("Successfully promoted user.")); break;
```
Daenyth: This isn't the right use for this. This data should be contained in the `Response` to the promotion action.

Look at the pr for adding registration protocol if you want an example of how to add new responses with commands.

common/pb/event_connection_closed.proto
```diff
@@ -11,6 +11,7 @@ message Event_ConnectionClosed {
 		BANNED = 4;
 		USERNAMEINVALID = 5;
 		USER_LIMIT_REACHED = 6;
+		DEMOTED = 7;
```
Daenyth: Reasonable.

common/pb/event_notify_user.proto
```diff
@@ -0,0 +1,18 @@
+import "session_event.proto";
+
+message Event_NotifyUser {
+
+	enum NotificationType {
+        PROMOTIONSUCCESS = 1;
+        PROMOTED = 2;
```
Daenyth: We might be able to not use this entire container if the server can just send the already connected user a new ServerInfo_User for themselves.

common/pb/admin_commands.proto
```diff
@@ -26,3 +27,13 @@ message Command_ReloadConfig {
 		optional Command_ReloadConfig ext = 1002;
 	}
 }
+
+message Command_AdjustMod {
+	extend AdminCommand {
+		optional Command_AdjustMod ext = 1003;
+	}
+	optional string admin_name = 1;
```
Daenyth: What's this used for? The server can tell who's sending the command already.

common/pb/admin_commands.proto
```diff
@@ -26,3 +27,13 @@ message Command_ReloadConfig {
...
+	optional string user_name = 2;
+	optional bool should_be_mod = 3;
```
Daenyth: These fields shouldn't be optional, the command doesn't make sense without them.

common/pb/admin_commands.proto
```diff
@@ -26,3 +27,13 @@ message Command_ReloadConfig {
 		optional Command_ReloadConfig ext = 1002;
 	}
 }
+
+message Command_AdjustMod {
```
Daenyth: This needs a new `Response_AdjustMod`

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,108 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
     settingsCache->sync();
     return Response::RespOk;
 }
+
+Response::ResponseCode ServerSocketInterface::cmdAdjustMod(const Command_AdjustMod &cmd, ResponseContainer & /*rc*/) {
+
+    QString adminName = QString::fromStdString(cmd.admin_name());
+    QString userName = QString::fromStdString(cmd.user_name());
+    QList<ServerSocketInterface *> userList;    //list is used to allow functionality of sending event messages to multiple people if need be.
```
Daenyth: How would you ever send a notification to multiple users when the command works on one user at a time?

cockatrice/src/tab_supervisor.cpp
woogerboy21: The reason it was written this way was to allow for the ability to be able to notify all staff members of the event (should we decide to change the code to do so).

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,108 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+            }
+        }
+        userList.clear();
+
+        ServerSocketInterface *admin = static_cast<ServerSocketInterface *>(server->getUsers().value(adminName));
+        userList.append(admin);
+
+        Event_NotifyUser sendtoadmin;
+        sendtoadmin.set_type(Event_NotifyUser::PROMOTIONSUCCESS);
+        if (!userList.isEmpty()) {
```
Daenyth: The `if` here is redundant because an empty `userList` will result in a `for` check of `0 < 0` and it will never enter the body

common/pb/admin_commands.proto
woogerboy21: This information was created to allow a notification system of more than just sender / server to have data passed to and from.

common/pb/admin_commands.proto
woogerboy21: Agree'd ill change this.

cockatrice/src/user_context_menu.cpp
```diff
@@ -193,8 +208,21 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
...
+        cmd.set_should_be_mod(true);
+
+        client->sendCommand(client->prepareAdminCommand(cmd));
+    } else if (actionClicked == aDemoteFromMod) {
+        Command_AdjustMod cmd;
+        cmd.set_admin_name(tabSupervisor->getUserInfo()->name());
+        cmd.set_user_name(userName.toStdString());
+        cmd.set_should_be_mod(false);
+
+        client->sendCommand(client->prepareAdminCommand(cmd));
```
Daenyth: These two branches could be combined like

```c++
if (actionClicked == aPromoteToMod || actionClicked == aDemoteFromMod) {
    Command_AdjustMod cmd;
    ...
    cmd.set_should_be_mod(actionClicked == aPromoteToMod)
    client->sendCommand(...)
}
```

cockatrice/src/user_context_menu.h
```diff
@@ -26,7 +27,7 @@ class UserContextMenu : public QObject {
     QAction *aAddToBuddyList, *aRemoveFromBuddyList;
     QAction *aAddToIgnoreList, *aRemoveFromIgnoreList;
     QAction *aKick;
-    QAction *aBan;
+    QAction *aBan, *aPromoteToMod, *aDemoteFromMod;
```
Daenyth: These should be on separate lines

servatrice/src/serversocketinterface.cpp
woogerboy21: I was thinking that the user may be logged in with different account names but the same client id.

cockatrice/src/tab_supervisor.cpp
Daenyth: I think we should omit that for now. Let's make the scope of this PR just to add the protocol ability for single actions + ui to do it

common/pb/event_notify_user.proto
woogerboy21: Could you explain more on this

servatrice/src/serversocketinterface.cpp
Daenyth: We should only inform the user being targeted, which is keyed by name.

Daenyth:
I think this isn't ready for merge yet; The protocol items aren't where I think they need to be. The server should inform the user performing an action of the result of that action using a Response container; Events are triggered by things other users do, not things you do yourself.

servatrice/src/serversocketinterface.cpp
woogerboy21: Regardless , if there is to ever be a notification to all the online staff members a list is needed.  That is the plans.

woogerboy21:
How do you intend on ever notifying the user then?  They take no action.

cockatrice/src/tab_supervisor.cpp
woogerboy21: Not to mention responses do not produce ui notifications.  Are you wanting that to change?

Daenyth:
Like I said, either by providing them with their updated server info and handling it seamlessly without restarting the app, or if that's too hard, (takes > 60 minutes) then we can use the notify method.

common/pb/admin_commands.proto
Daenyth: In that case you definitely should get the admin name via the connection. If you use this, I can lie and said that a different user performed the action.

woogerboy21: wait what?  are you saying your going to adjust the code to lie rather than adjust the code to change the protocol item?

cockatrice/src/tab_supervisor.cpp
woogerboy21: Wait a minute... this is just an event processing function on the client side.  This data is sent from the response of the promotion in the server socket handler.  Are you confusing something here?

servatrice/src/serversocketinterface.cpp
woogerboy21: Then why is it used else were in the cod?

cockatrice/src/user_context_menu.cpp
woogerboy21: No they cant, the should_be_mod container item is different.

common/pb/admin_commands.proto
woogerboy21: So are you basically wanting me to create its own response container and move the failed event's from above into the response container (in a nutshell)?  The successful events for the sending admin could be also moved , but that seems silly if the design of the notification for this type of command is more of a broadcast to staff members.

woogerboy21: Hold on a second.  This brings up a good point.  The response container is going to be processed client side.  Are you suggesting I then make a second server call to do a simply client notifiction?

woogerboy21:
@Daenyth 
Alright, I think I have adjusted everything requested.  The only thing not in place is the notification to the user.  Baring some type of automatic method of uping a user with out them logging out and back in, how would you like notifications to users on mod/admin based commands to flow?  There will need a way for this type of thing to occur.  I find it silly to expect the client to send a second command to the server requesting a notification to be sent to the client when the server should be the one sending events and calculating when those events should occur.

common/pb/admin_commands.proto
woogerboy21: Response container added (just saying since there is no diff to collapse this.)

Daenyth: The admin doing the command gets a response. Everyone else gets an event.

woogerboy21: Thats how it is now :)

Daenyth:
The notification sent to the mods who didn't do the promote/demote should include the name of the admin who did the action

Daenyth:
Your concerns with user info handling promotion automatically are very reasonable. Notification isnt out of line.

woogerboy21:
@Daenyth So I took out the event notification for everyone but the single user that is getting promoted.  The notification that occurs to the admin comes through the response at this point.  I removed the notification list for any other moderators as you had mentioned working this PR so that it was more a 1 to 1 type functionality

cockatrice/src/user_context_menu.cpp
```diff
@@ -46,6 +49,8 @@ void UserContextMenu::retranslateUi()
     aRemoveFromIgnoreList->setText(tr("Remove from &ignore list"));
     aKick->setText(tr("Kick from &game"));
     aBan->setText(tr("Ban from &server"));
+    aPromoteToMod->setText(tr("&Promote user to moderator"));
+    aDemoteFromMod->setText(tr("Dem&ote user from moderator"));
```
ZeldaZach: Do these "shortcuts" really do much? (jw)

cockatrice/src/user_context_menu.cpp
```diff
@@ -89,6 +94,27 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
...
+    if (resp.response_code() == Response::RespOk) {
+        if (cmd.should_be_mod()) {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Success"), tr("Successfully promoted user."));
+        } else {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Success"), tr("Successfully demoted user."));
+        }
+
+    } else {
+        if (cmd.should_be_mod()) {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Failed"), tr("Failed to promoted user."));
```
ZeldaZach: "Failed to **promote** user"

cockatrice/src/user_context_menu.cpp
woogerboy21: I never use them, but that doesn't mean others dont.

cockatrice/src/user_context_menu.cpp
```diff
@@ -89,6 +94,27 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
...
+        } else {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Failed"), tr("Failed to demoted user."));
```
ZeldaZach: "Failed to **demote** user"

cockatrice/src/user_context_menu.h
```diff
@@ -27,10 +27,12 @@ class UserContextMenu : public QObject {
     QAction *aAddToIgnoreList, *aRemoveFromIgnoreList;
     QAction *aKick;
     QAction *aBan;
+    QAction *aPromoteToMod, *aDemoteFromMod;
```
ZeldaZach: Not fully necessary, but do you want to put all the `QAction` vars into one line?

common/pb/admin_commands.proto
ZeldaZach: :+1:

cockatrice/src/user_context_menu.h
woogerboy21: @Daenyth told me to split them into separate lines.

common/pb/response_adjust_mod.proto
```diff
@@ -0,0 +1,7 @@
+import "response.proto";
+
+message Response_AdjustMod{
+	extend Response {
+		optional Response_AdjustMod ext = 1011;
```
ZeldaZach: Should I yell at you about tabs now or later? :stuck_out_tongue:

woogerboy21: If I recall correctly all the proto files are a different indent than the 4 space.  CLion tells me that all the time when im working with them.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,55 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+Response::ResponseCode ServerSocketInterface::cmdAdjustMod(const Command_AdjustMod &cmd, ResponseContainer & /*rc*/) {
+
+    QString adminName = QString::fromStdString(cmd.admin_name());
+    QString userName = QString::fromStdString(cmd.user_name());
+
+    if (cmd.should_be_mod()) {
+        logDebugMessage("Received admin command: promote user");
+        QSqlQuery *query = sqlInterface->prepareQuery(
+                "update {prefix}_users set admin = :adminlevel where name = :username");
+        query->bindValue(":adminlevel", "2");
```
ZeldaZach: Are we not able to just say `set admin = 2 where name...`? If so, i don't see a reason for the bind. If not, then this is fine.

common/pb/response_adjust_mod.proto
ZeldaZach: Whenever I work with them, I just readjust them to 4 space

cockatrice/src/user_context_menu.cpp
ZeldaZach: Do they work on OS X?

common/pb/response_adjust_mod.proto
woogerboy21: I think i may look into fixing the proto warnings that get thrown for every proto file.  If I can fix those ill fix the spacing at that time.

servatrice/src/serversocketinterface.cpp
woogerboy21: Ill fix this.  Ill keep the bind but change the value to an int (as that is what it should be).  Nice catch thanks.

cockatrice/src/user_context_menu.cpp
woogerboy21: Aren't you the mac guy?

common/pb/response_adjust_mod.proto
ZeldaZach: :100: kk

cockatrice/src/user_context_menu.h
ZeldaZach: :+1: Gotcha

cockatrice/src/user_context_menu.cpp
ZeldaZach: They never worked properly :disappointed:

common/pb/response_adjust_mod.proto
woogerboy21: FYI, i did find a fix for the proto warnings so Ill throw a PR in to fix the formatting as well as the warnings all at once.

common/pb/admin_commands.proto
```diff
@@ -26,3 +27,13 @@ message Command_ReloadConfig {
...
+	extend AdminCommand {
+		optional Command_AdjustMod ext = 1003;
+	}
+	required string admin_name = 1;
```
Daenyth: The server should get this information from the admins connection. If you use this field I could make a malicious client that would say that someone other than me performed the promotion

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,55 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+
+    if (cmd.should_be_mod()) {
+        logDebugMessage("Received admin command: promote user");
+        QSqlQuery *query = sqlInterface->prepareQuery(
+                "update {prefix}_users set admin = :adminlevel where name = :username");
+        query->bindValue(":adminlevel", 2);
+        query->bindValue(":username", userName);
+        if (!sqlInterface->execSqlQuery(query))
+            return Response::RespInternalError;
+
```
Daenyth: What if the user isn't online?

ZeldaZach: Good point, there should be a check to see if they're online (a simple query) and if not then just skip the remainder

cockatrice/src/user_context_menu.cpp
```diff
@@ -89,6 +94,27 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
     dlg->show();
 }
 
+void UserContextMenu::adjustMod_processUserResponse(const Response &resp, const CommandContainer &commandContainer)
```
Daenyth: I'm not sure the user menu is the right place to handle user level adjustment responses

servatrice/src/serversocketinterface.cpp
woogerboy21: Why would we check if the user is online to promote / demote?  Do you not want to be able to do so unless they are online?  That sounds like the admin will be waiting around for a while trying to get the time frames right between his schedule and there's.

Daenyth: If the user is offline I'm pretty sure you'll get a null pointer here and then use it crashing the server.

cockatrice/src/user_context_menu.cpp
woogerboy21: The user menu is were its fired from, why wouldnt the response be processed from it as well?  You quicker than lightning and going to close all your context menus before the server has time to process the command?

Daenyth:
I think this should wait a little more. I'll try to do more detailed review and testing when I'm at a computer

servatrice/src/serversocketinterface.cpp
woogerboy21: The command is sent from a menu that shows a clients name listed.  The only way the command could be fired from a menu with an offline client is from your buddy list.  Which will show the user as grey but still be able to pull the name from the list.

cockatrice/src/user_context_menu.cpp
Daenyth: You joke, but yes, I've done this. Bad connections exist

woogerboy21: So what happens.  The command is sent, the server processes it and the server tries to send you the response back but if you get disconnected the response container isnt sent at all and well it doesnt matter in the end.

servatrice/src/serversocketinterface.cpp
Daenyth: Don't make assumptions about what the client can do, someone can always
modify it. Code the server and protocol for the worst case.

Under normal operation I could perform the promote and before the server
gets my message the target could disconnect.so it should seamlessly handle
promoting offline users. Similarly we should be able to demote bad mods
when offline

woogerboy21: What are you talking about.  The server takes the command containers information and runs the sql query against it.  It doesnt care if the user is online/offline  .. taking a :poop:

cockatrice/src/user_context_menu.cpp
ZeldaZach: ^ Makes sense. If it fails to respond is that a big deal? Seems like that's how it's built now anyway

common/pb/event_connection_closed.proto
ZeldaZach: :+1:

common/pb/response_adjust_mod.proto
ZeldaZach: :+1:

servatrice/src/serversocketinterface.cpp
woogerboy21: So it looks like this is how its handled:
```
Response::ResponseCode Server_ProtocolHandler::cmdListUsers(const Command_ListUsers & /*cmd*/, ResponseContainer &rc)
{
    if (authState == NotLoggedIn)
        return Response::RespLoginNeeded;

    Response_ListUsers *re = new Response_ListUsers;
    server->clientsLock.lockForRead();
    QMapIterator<QString, Server_ProtocolHandler *> userIterator = server->getUsers();
    while (userIterator.hasNext())
        re->add_user_list()->CopyFrom(userIterator.next().value()->copyUserInfo(false));
    QMapIterator<QString, Server_AbstractUserInterface *> extIterator = server->getExternalUsers();
    while (extIterator.hasNext())
        re->add_user_list()->CopyFrom(extIterator.next().value()->copyUserInfo(false));

    acceptsUserListChanges = true;
    server->clientsLock.unlock();

    rc.setResponseExtension(re);
    return Response::RespOk;
}
```
The server actually takes the current listing of users and locks the list of changes prior to reading, iterates the list, and then returns the data if the user is online from that list.  Using the method that is used to collect a users data to send the event to shouldnt require you to check if the user is online as the socket handler does that for you.

woogerboy21:
>https://github.com/Cockatrice/Cockatrice/pull/1391#discussion_r37600212

The only other place that this would make any since to fall into is the tabSupervisor.  But I don't know I agree that another class should be processing responses from other classes.

cockatrice/src/user_context_menu.cpp
```diff
@@ -89,6 +94,27 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
...
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Success"), tr("Successfully promoted user."));
+        } else {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Success"), tr("Successfully demoted user."));
+        }
+
+    } else {
+        if (cmd.should_be_mod()) {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Failed"), tr("Failed to promote user."));
+        } else {
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Failed"), tr("Failed to demote user."));
```
Daenyth: For all these messages, can you include the name of the person I promoted? If I do multiple in a row I won't be able to tell the difference if the server responds slowly.

cockatrice/src/user_context_menu.cpp
```diff
@@ -193,7 +230,24 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
...
+        // client->sendCommand(client->prepareAdminCommand(cmd));
+        PendingCommand *pend = client->prepareAdminCommand(cmd);
+        connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(adjustMod_processUserResponse(Response,CommandContainer)));
+        client->sendCommand(pend);
+    } else if (actionClicked == aDemoteFromMod) {
+        Command_AdjustMod cmd;
+        cmd.set_user_name(userName.toStdString());
+        cmd.set_should_be_mod(false);
 
+        // client->sendCommand(client->prepareAdminCommand(cmd));
```
Daenyth: Extra commented lines here and in the block above can be removed.

cockatrice/src/user_context_menu.h
```diff
@@ -27,10 +27,12 @@ class UserContextMenu : public QObject {
...
 signals:
     void openMessageDialog(const QString &userName, bool focus);
 private slots:
     void banUser_processUserInfoResponse(const Response &resp);
+    void adjustMod_processUserResponse(const Response &resp, const CommandContainer &commandContainer);
```
Daenyth: I suppose this isn't terrible since ban response is handled the same

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,54 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
     settingsCache->sync();
     return Response::RespOk;
 }
+
+Response::ResponseCode ServerSocketInterface::cmdAdjustMod(const Command_AdjustMod &cmd, ResponseContainer & /*rc*/) {
+
+    QString userName = QString::fromStdString(cmd.user_name());
+
+    if (cmd.should_be_mod()) {
+        logDebugMessage("Received admin command: promote user");
```
Daenyth: It would be nice to include the name of the admin and name of the target in the debug line.

servatrice/src/serversocketinterface.cpp
Daenyth: If it just results in a timeout that's not so bad but before we merge this I'd like to use the python client I wrote for registration testing to test sending promotion/demotion of offline users and see how the server handles it.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,54 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+
+        ServerSocketInterface *user = static_cast<ServerSocketInterface *>(server->getUsers().value(userName));
+        Event_NotifyUser event;
+        event.set_type(Event_NotifyUser::PROMOTED);
+        SessionEvent *se = user->prepareSessionEvent(event);
+        user->sendProtocolItem(*se);
+        delete se;
+
+    } else {
+        logDebugMessage("Received admin command: demote user");
```
Daenyth: As above, logging names here would help.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,54 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+    QString userName = QString::fromStdString(cmd.user_name());
+
+    if (cmd.should_be_mod()) {
+        logDebugMessage("Received admin command: promote user");
+        QSqlQuery *query = sqlInterface->prepareQuery(
+                "update {prefix}_users set admin = :adminlevel where name = :username");
+        query->bindValue(":adminlevel", 2);
+        query->bindValue(":username", userName);
+        if (!sqlInterface->execSqlQuery(query))
+            return Response::RespInternalError;
```
Daenyth: We should log the query error.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,54 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+        user->sendProtocolItem(*se);
+        delete se;
+
+    } else {
+        logDebugMessage("Received admin command: demote user");
+        QSqlQuery *query = sqlInterface->prepareQuery("update {prefix}_users set admin = :adminlevel where name = :username");
+        query->bindValue(":adminlevel", 0);
+        query->bindValue(":username", userName);
+        if (!sqlInterface->execSqlQuery(query))
+            return Response::RespInternalError;
```
Daenyth: We should log the query error.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1000,3 +1002,54 @@ Response::ResponseCode ServerSocketInterface::cmdReloadConfig(const Command_Relo
...
+
+        QList<ServerSocketInterface *> userList;
+        ServerSocketInterface *user = static_cast<ServerSocketInterface *>(server->getUsers().value(userName));
+        userList.append(user);
+
+        if (!userList.isEmpty()) {
```
Daenyth: I don't see how userList can be non-empty here -- you just called `append` with `user`.

Skip the list maybe?

woogerboy21: Ah yes, I did change the promote to a single user target.  The demote needs adjusted as well.  But to answer your question, I try and stick to the same code style that is already in the code and this is what was originally used in the ban code that performs the same task.

Daenyth: An ip or id ban can apply to multiple online users, so it makes sense to have a list. This can only ever affect one user at a time though, so it doesn't make sense.

ZeldaZach:
This is now good to go. Tested and works fully as expected with the user online and offline :shipit:

PR #1399, opened 2015-08-21
Refactored settings to use settingscache & ini files - Possible Portable version

marcofernandezheras:
All setting around cockatrice refactored to use settingscache & ini files.
Include a translation from old setting system to the new system.
Also  tester needed to check if is possible to compile a portable version defining PORTABLE_BUILD


woogerboy21:
oh neat

Daenyth:
I suppose the best way to test this would be to run a build with these settings for a while

ctrlaltca:
Tested and working. I just noticed the ```GameFiltersSettings::setGameHashedTypeEnabled()``` method being unused. 
This will require a minor change in the NSIS installer: modification (or removal) of the "Update the paths in the application settings" option.

marcofernandezheras:
GameFiltersSettings::setGameHashedTypeEnabled() must be used at translate legacy settings... I'll fix it later :(

marcofernandezheras:
use -DPORTABLE=1 to build the portable version

tooomm:
Isn't it possible to combine the portable and normal windows version?
The NSIS installer asks the user on windows if he wants to install the program (with reg entries, settings in user folder, shortcuts etc.) or if he wants to use the program portable (unzip, settings in cockatrice folder etc.).

ctrlaltca:
This is currently hardcoded in the compiled binary. I suppose we could remove the #ifdefs and add a simple check eg. like checking if a file named "portable.xxx" exists alongside cockatrice.exe. The NSIS installer could then create this empty file if the user chose to installa  portable version.
I'm unsure if this really belongs to this PR or if we prefer to merge this as-is and plan the modified installer for a new PR.

ZeldaZach:
If I'm able to put the Cockatrice application on a USB stick and then use it on another computer with my same settings, then this PR does its purpose.

Conflicts must be resolved @marcofernandezheras and then I'd be fine with getting this in.

ZeldaZach:
This doesn't seem to work as expected for Macs.

I attempted to build the clients using both `DPORTABLE_BUILD=1` and `PORTABLE=1`, with neither having success in transferring my settings to the USB drive once I copy/paste the applications.

Also, it seems like the Oracle tool doesn't work here as the Continue button doesn't seem to work.

I also changed [The configs](https://github.com/Cockatrice/Cockatrice/pull/1399/files#diff-0538322454252454e4c452682aea843eR338) (Both trice & oracle) to "ON" but that didn't work either.

woogerboy21:
So I just checked this on windows.  A couple of questions.  Were is the settings supposed to be saving and are the default settings for things like deckpaths supposed to be different than the default appdata folder settings?

ZeldaZach:
To consider a client "portable" I feel the INI/setting files should be within the app folder (On mac, `cockatrice.app/Contents`, for example).

PR #1400, opened 2015-08-22
Log old session out of server if logging in from a second location

### LINKED ISSUE #1393: Extend login protocol to include ability to force close sessions
Add a new boolean flag to the protocol when logging in which will inform the server to kill existing sessions for that user.

woogerboy21:
Fix #1393 

This doesnt extend the session information, its just a change to the way the server handles the previous logged in session.  With no update to the protocol there is no need for a client side change.

image

### DIFF
--- a/common/server.cpp
+++ b/common/server.cpp
@@ -30,6 +30,7 @@
 #include "pb/event_user_left.pb.h"
 #include "pb/event_list_rooms.pb.h"
 #include "pb/session_event.pb.h"
+#include "pb/event_connection_closed.pb.h"
 #include "pb/isl_message.pb.h"
 #include <QCoreApplication>
 #include <QThread>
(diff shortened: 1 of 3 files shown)


common/server.cpp
```diff
@@ -126,9 +127,17 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString
...
-            databaseInterface->unlockSessionTables();
-            return WouldOverwriteOldSession;
+            qDebug("Session already logged in, logging old session out");
+
+            Event_ConnectionClosed event;
+            event.set_reason(Event_ConnectionClosed::LOGGEDINELSEWERE);
+            event.set_reason_str("You have been logged out due to logging in at another location.");
+            event.set_end_time(QDateTime::currentDateTime().toTime_t());
+
+            SessionEvent *se = getUsers().value(name)->prepareSessionEvent(event);
```
Daenyth: I think you can call `users.value` here instead of going through `getUsers`

Daenyth:
:+1: overall but squash the commits down.

woogerboy21:
Yea originally there was only a single commit, then i threw the translation line in and it borked things over only to find that the translation is done on the client side so a case statement will need added for that.  Ill get that put int and this cleaned up a little later.

woogerboy21:
:shipit: 
Rebased and tested.  Changes made were what @Daenyth as well as a translatable response line in the client so when the users eventually update the message can be translated.  Otherwise with older clients it will just display the message in english form.

PR #1414, opened 2015-08-23
Prevent bad flags

### LINKED ISSUE #1412: DB Country Code Issue
So there looks to be n issue with the way that the data in the database for the country code is being written.   I haven't tracked down if its with the registration process or the edit dialogue but when I user has the definition for "Undefined" (the blank box) translated, the insert that is being done is "Un".  I also see it for a couple of other countries like the following:

```
Un
미지
```

Im not to familiar with the country codes, but my guess is that the country code shouldnt be translated.  The countries name in the drop down box should be.

ZeldaZach:
Fix #1412 

This prevents the translation & invalid args being sent when registering for invalid flags.

### DIFF
--- a/cockatrice/src/dlg_register.h
+++ b/cockatrice/src/dlg_register.h
@@ -19,7 +19,7 @@ public:
     QString getPassword() const { return passwordEdit->text(); }
     QString getEmail() const { return emailEdit->text(); }
     int getGender() const { return genderEdit->currentIndex() - 1; }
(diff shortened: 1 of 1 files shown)


ctrlaltca:
This just removes the "undefined" text. The problem is here:
https://github.com/Cockatrice/Cockatrice/blob/18ec8e20b60f4ea0a1c299c3fe6e8f8b07ae20c6/cockatrice/src/dlg_register.h#L22
Its checks genderEdit instead of countryEdit. That's why the problem occurred so rarely: the user had to choose a gender but not a country to get the invalid value sent to the server.

PR #1416, opened 2015-08-23
Define URLs

### LINKED ISSUE #1411: Move url links in client to 'define'
Prototype: https://github.com/Cockatrice/Cockatrice/pull/1318
Links without `define`: https://github.com/Cockatrice/Cockatrice/pull/1213

Can't think of more right now...

ZeldaZach:
Fix #1411 

Defines the URLs in global vars instead of direct linking.


Daenyth:
:+1:

PR #1417, opened 2015-08-24
Extend the login protocol to send the client version (as optional)

woogerboy21:
Added the client version information into the PB item to be sent to the server from the client at login.


Daenyth:
:+1:

PR #1418, opened 2015-08-25
[WIP] Roles addition to user levels

woogerboy21:
This PR adds a new column to the user table allowing for user level roles.  This will allow for a much more granular level of permissions as well as more colored gummy's in the user lists.  I added a role called "Lead" with a custom purple based gummy for moderators assigned that role.

Being an additional value to the users information, older clients shouldnt be affected in any way.  They simply will not show the updated information in the UI till they upgrade.

image

image

image


ZeldaZach:
Do we plan on making the head moderators a different colored pawn? Just curious

cockatrice/src/userinfobox.cpp
```diff
@@ -116,6 +116,19 @@ void UserInfoBox::updateInfo(const ServerInfo_User &user)
         userLevelText = tr("Unregistered user");
     userLevelLabel3.setText(userLevelText);
 
+    switch (user.user_role()) {
+        case (0):
```
ZeldaZach: The `()` aren't really needed here

ZeldaZach:
A suggestion is instead of having the gummy pawns designated by "userLevel" alone, allow there to be an override:

```
if (!role.isEmpty())
{
    QSvgRenderer svg(QString(":/resources/userlevels/%1.svg").arg(role));
}
else
{
    QSvgRenderer svg(QString(":/resources/userlevels/" + levelString + .svg"));
}
```

Which could allow for a little more customization (maybe?).

What you have, I could see a donation and/or judge rank being set on the "registered" class as "registered1", "registered2", etc, but I'd like to hear your opinion on my suggestion.

woogerboy21:
@ZeldaZach 
The original idea for roles (for me that is) was to allow custom gummy images for users that have donated (if they so choose).  But since then the idea sprawled a little to include the head moderators as well.  And yes the idea was to give the head moderators a different gummy color + under the user information box users could see that the head moderators had a different role designation.

As far as the over ride idea, I actually had it coded in that way at first but decided that for ease of understanding in regards to which gummy svg would line up with which user role keeping the numerical value at the end of the svg would be easier.  If others feel the over ride should be in place I have no problem putting it back.

woogerboy21:
The only thing I would say Im not to happy with whats in the code is the block that is used to specify the role string for the user info box.  Does anyone know or have any good suggestions on a way to take and use the enum in the protobuf file to seach and declare a text string result?  I would like too simplify that block to do something like iterate the enum block and pull the designated user role value (as a string) from it to put in the user info box role box string location.

I could find a way to actually loop through the enum , but couldn't figure out how to convert the enum's text based designation to a string.

cockatrice/src/chatview.cpp
```diff
@@ -155,7 +155,7 @@ void ChatView::appendMessage(QString message, QString sender, UserLevelFlags use
         if (!sender.isEmpty() && tabSupervisor->getUserListsTab()) {
             const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
             QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
-            cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel, buddyList.contains(sender)).toImage());
+            cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel, buddyList.contains(sender), tabSupervisor->getUserInfo()->user_role()).toImage());
```
poixen: Doesnt the tab supervisor contain info about the current logged in user? You need to get the senders userlevel, not your own.

cockatrice/src/gamesmodel.cpp
```diff
@@ -86,7 +86,7 @@ QVariant GamesModel::data(const QModelIndex &index, int role) const
             case Qt::DisplayRole:
                 return QString::fromStdString(g.creator_info().name());
             case Qt::DecorationRole: {
-                    QPixmap avatarPixmap = UserLevelPixmapGenerator::generatePixmap(13, (UserLevelFlags)g.creator_info().user_level(), false);
+                    QPixmap avatarPixmap = UserLevelPixmapGenerator::generatePixmap(13, (UserLevelFlags)g.creator_info().user_level(), false, g.creator_info().user_role());
```
poixen: :+1:

cockatrice/src/chatview.cpp
woogerboy21: Yes, you are correct.  This needs fixed.

common/pb/serverinfo_user.proto
```diff
@@ -1,3 +1,4 @@
+syntax = "proto2";
```
Daenyth: What's this?

common/pb/serverinfo_user.proto
```diff
@@ -11,6 +12,10 @@ message ServerInfo_User {
         Male = 0;
         Female = 1;
     };
+    enum UserRole {
```
Daenyth: I'm not a fan of this - I think it makes the roles hard to extend. I'll post my concept on the main pr section

common/pb/serverinfo_user.proto
woogerboy21: This is what stops the warning messages from showing when you compile the .proto files.  it could use applied to all the .proto files.  I plan on putting a PR in to resolve all the ones that do not have the line in it.

common/pb/serverinfo_user.proto
woogerboy21: The enum?  Its not even used for any of the code.  I origiinally used it and it did make it hard to extend the user roles instead of a basic int so I left it in for human reference for what role is what.

Daenyth:
Here's the design I had in mind for roles:

Client side changes:
- Each user icon gets an enum value. The ServerInfo_User is extended to specify the icon. The existing user level values continue to be sent in there for backwards compatibility.

Server changes:
- New table, `roles`
- A role has a name, an icon (from the icon enum I mentioned above), an id, and an optional parent role.
- New table `role_permissions`
- Each role_permission entry refers to a role by id and permits a specific action.
- Users table changes to have a role column with a reference to their role id
- When a user tries to perform a protected action, the server checks the role that user has and then checks the set of role_permissions available to that role and that role's parent, recursively until there are no more parents. If the action they are trying to perform isn't in that set, they aren't allowed to do it.
- The server aids older clients which don't handle the custom icons by populating the old user levels as follows: Unregistered is obviously the same; If the user has permission to adjust mod levels, send admin. If the user has permission to ban, send moderator; otherwise send registered.

Here's an example default configuration under this model:

Role:

id | name | parent_id | icon
--- | --- | --------- | ----
1 | Registered | null | blue
2 | Moderator | 1 | black-white
3 | Admin | 2 | red

Role_permissions

id | permission | role_id
---- | ---- | ----
1 | ban_user | 2
2 | adjust_mod | 3
3 | shutdown_server | 3
4 | set_server_text | 3
5 | create_room | 2

With this example set, moderators can `ban_user`, `create_room`, and no more. Admins can `ban_user`, `create_room`, `shutdown_server`, etc.

The benefit of this approach is that it's infinitely extensible and the only protocol updates required are ones adding new icons. If the icon is a string instead of an enum we don't even need that. We'd have to have some fallback selection.

What do you think?

ZeldaZach:
I really like that idea Dae. Seems straight forward and easily changeable.

woogerboy21:
Initial thought, overly complicated. But would have to sit and think about scenario's as what your trying to accomplish .

Your point about backwards compatible clients and icons is a good point.  With no check in place right now the gummy for a new role wouldn't exist and user wouldn't see anything for a gummy.

So with your scenario, what happens when you want to give a user all permissions?  Would you have a role_permission row for each permission with that user id per row?

Daenyth:
> Initial thought, overly complicated
My approach is definitely more complicated but the advantage, I think, is that it keeps the protocol more future-proof. The more changes we can make without touching the protocol the more features we can enable without forcing people to change their clients.

> So with your scenario, what happens when you want to give a user all permissions?

You would define a role that has all permissions and mark the user as having that role.

> And what about new icons? How would one go about adding new gummy icons to roles? 
We could add a new protocol action. Something like `Request_RoleIcon` or something. It would ask the server to send the svg for a role name to the client.

Daenyth:
Another option which doesn't require the icon sending is to add a fallback question mark icon to use when we don't have one. The client would get the new icon by updating.

woogerboy21:
>You would define a role that has all permissions and mark the user as having that role.

So under the permissions column in the role permissions table, is there multiple permissions listed with a separator in the permissions column, or is the column designed only to have a single permission listed in the permission column requiring multiple rows with similar information for all other columns and different permissions defined requiring multiple select / join or iterations of the information out of the database table?

It seems to me that there is a step missing some were.  As if there needs to be a table that lists the role details, a table that lists the permissions details, and a table that combines the two to define what the role as a whole should look like along with possibly the user id linked in some way some were linking the user account to the information as a whole.

Or am I over thinking this combined with my borderline low blood sugar making me go crazy?

woogerboy21:
Ok, even more thinking on this.  I dont think we should be adjusting or changing the existing user level's that are in place as they are today.  While thinking scenario's through in my head I would like to see a role be a sub class of those different user levels.  That way when it comes to doing things like permission assignment or classifying data that should be sent to out through maybe situations like debug information we have a very easy way of saying "users" shouldnt get this but moderators and above should.  

@Daenyth   Thoughts?

ZeldaZach:
Any word on this?

woogerboy21:
@ZeldaZach I had been working on the ban history PR.  I want to roll in the official warnings functionality first and then get back to this.

Issue #1214, opened 2015-07-06
Client Version Check


woogerboy21:
Any idea how we could implement a client version check?  It would be nice to notify a user when they connect to a server if they are running an older client and to consider upgrading.  The first thing I thought was to have a DB entry (maybe in the schema table) and have the client pass a number to the server and compare it against. 

This may not be the best idea and am open to hearing other idea's.

ZeldaZach:
inside of the servatrice.ini file we should add a "min_requirement" and a "suggested_min" so the server can say "hey if you don't have a client that was made on date1 then you can't connect" and "hey you should update to this client as we're going to discontinue yours in the near future!"

woogerboy21:
ini might be better, i wasnt thinking they would be denied but i guess that could be an option.  i wonder if you can pull the commit date of the version being compiled and place it in the client some were as the commit date would work well as a check

Daenyth:
I'd support what @ZeldaZach says if we use it sparingly. If it's a weekly message people will ignore it.

woogerboy21:
@Daenyth well if its a .ini configuration option it would be at the server operators discretion .  Personally as an operator I would use it when I found major feature improvements were available.  Such as the in client account management. functionality that @ctrlaltca has created.  But thats not to say other server operators wouldnt choose to change there config on a more aggressive time frame.

ZeldaZach:
Nothing wrong with a server owner being more aggressive @woogerboy21. They want their players to have the best experience possible

woogerboy21:
@ZeldaZach Nope there isn't.  Just wanted to point out that it would be at the operators discretion since @Daenyth seemed hesitant

Daenyth:
I was imagining this would be hardcoded.

There's a question of whether to use build data or protocol version.

For this feature I think build date (read: date stamp of the git commit) is most appropriate.

ZeldaZach:
I was thinking of parsing the build date from the version we include in client, but I'd like to see this done in the INI over being hardcoded. We can always include a "github operator suggested min version" which is hardcoded, but allow server operators to make the requirements more strict
 + i.e. if we hardcode that clients made on `Jan 1, 2015` are still usable, the server operator can make the requirements `Jan 2, 2015` or newer from the INI

woogerboy21:
@ZeldaZach - 'I was thinking of parsing the build date from the version we include in client
Your not suggesting we have it hard coded some were a date that has to be constantly updated are you?

[...]

PR #1419, opened 2015-08-25
[WIP] Updated client notification

woogerboy21:
This is a fix / partial fix for #1214 

This PR adds the ability for client update checking based against the server that is being connected to.  How this works is, upon successful login a command is sent from the client to the server with the clients version string.  That string is then compared by the server to its own version string information and calculates if the servers version is newer than the clients version and if depending on the results sends a response back to the client containing the servers version string and true/false for a client upgrade.  The client then runs through its own checks to see if the expected result for update matches and if it does , it notifies the user that an updated client is available.

Both parties have the ability to disable the functionality preventing either update checks from occurring from the client or update true / server version string information from being returned to the client from the server.

The reason for this transaction type checking on both client and server parts is the future expectation for signed installers allowing for a trusted version of the code base.  If the client (or server) knows that they are in fact from a trusted source they can guarantee that there side calculations for if a client check is correct.  If that type of solution never comes to fruition the functionality will still operate, it just is doing more work than needed.

Another bit of functionality has been added to allow additional flexibility.  The client side check is designed to be a fall back type solution to update checking.  There is a variable created in the settings cache (TrustedSourceClientUpdateSuccess) that can be flipped to disable the client update check at login portion of the functionality leaving room for another update solution to occur prior to this version string type commit check to be performed.  With this functionality in place when another (non-version) type update solution comes around it can be worked in very easily and become the primary update check solution utilizing this as a fallback method.

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -1,16 +1,18 @@
 #include <QTimer>
 #include <QThread>
+#include <QDateTime>
 #include "remoteclient.h"
-#include "settingscache.h"
 #include "pending_command.h"
 #include "pb/commands.pb.h"
 #include "pb/session_commands.pb.h"
 #include "pb/response_login.pb.h"
 #include "pb/response_register.pb.h"
 #include "pb/response_activate.pb.h"
+#include "pb/response_update_check.pb.h"
 #include "pb/server_message.pb.h"
 #include "pb/event_server_identification.pb.h"
 #include "settingscache.h"
+#include "version_string.h"
 #include "main.h"
 #include "version_string.h"
 
@@ -37,6 +39,8 @@ RemoteClient::RemoteClient(QObject *parent)
     connect(this, SIGNAL(sigDisconnectFromServer()), this, SLOT(doDisconnectFromServer()));
     connect(this, SIGNAL(sigRegisterToServer(QString, unsigned int, QString, QString, QString, int, QString, QString)), this, SLOT(doRegisterToServer(QString, unsigned int, QString, QString, QString, int, QString, QString)));
     connect(this, SIGNAL(sigActivateToServer(QString)), this, SLOT(doActivateToServer(QString)));
+    connect(this, SIGNAL(serverUpdateClientCheckReceived()), this, SLOT(doServerUpdateClientCheck()));
+
 }
 
 RemoteClient::~RemoteClient()
@@ -140,6 +144,11 @@ void RemoteClient::loginResponse(const Response &response)
         for (int i = resp.ignore_list_size() - 1; i >= 0; --i)
             ignoreList.append(resp.ignore_list(i));
         emit ignoreListReceived(ignoreList);
+
+        if (settingsCache->getEnableClientUpdateCheck())
+            if (!settingsCache->getTrustedSourceClientUpdateSuccess())
+                emit serverUpdateClientCheckReceived();
+
     } else {
         emit loginError(response.response_code(), QString::fromStdString(resp.denied_reason_str()), resp.denied_end_time());
         setStatus(StatusDisconnecting);
@@ -249,6 +258,7 @@ void RemoteClient::doConnectToServer(const QString &hostname, unsigned int port,
     userName = _userName;
     password = _password;
     QString clientid = settingsCache->getClientID();
+    QString clientver = (VERSION_STRING);
     lastHostname = hostname;
     lastPort = port;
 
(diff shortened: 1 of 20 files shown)


woogerboy21:
Side note.  I may want to rename that fallback settings cache variable name to something more simple.

poixen:
Next up: Auto update clients. 

Chop chop!

cockatrice/src/remoteclient.cpp
```diff
@@ -348,3 +358,80 @@ void RemoteClient::disconnectFromServer()
...
+                 *  run the same type of date/commit calculations and verify the results are the same.  If
+                 *  there is a change in the logic here, make sure you update the logic in the server_protocolhandler.cpp
+                 *  file to match the clientUpdateAvailable results.
+                 */
+
+
+                QString serverVer = QString::fromStdString(resp.server_version()).simplified();
+                QString clientVer = QString::fromStdString(VERSION_STRING);
+                QDate clientDate = QDate::fromString(serverVer.mid(9, 10), "yyyy'-'MM'-'dd");
+                QDate serverDate = QDate::fromString(clientVer.mid(9, 10), "yyyy'-'MM'-'dd");
```
ZeldaZach: Did you get these `QDate`s backwards?

cockatrice/src/remoteclient.cpp
```diff
@@ -249,6 +258,7 @@ void RemoteClient::doConnectToServer(const QString &hostname, unsigned int port,
     userName = _userName;
     password = _password;
     QString clientid = settingsCache->getClientID();
+    QString clientver = (VERSION_STRING);
```
ZeldaZach: Are the `()` necessary?

cockatrice/src/remoteclient.cpp
```diff
@@ -348,3 +358,80 @@ void RemoteClient::disconnectFromServer()
...
+                QDate serverDate = QDate::fromString(clientVer.mid(9, 10), "yyyy'-'MM'-'dd");
+                bool clientUpdateAvailable = false;
+
+                if (clientVer.isEmpty()) {
+                    // client version not available OR client does not have code for sending the information
+                    clientUpdateAvailable = true;
+                } else if (serverVer.isEmpty()) {
+                    // server version not available OR server does not have the code for sending the information
+                    clientUpdateAvailable = false;
+                } else if (clientDate.operator==(serverDate)) {
```
ZeldaZach: Do we need the `()` around `serverDate`?

tooomm:
Didn't you said once you don't want to leak information about the server?
To prevent people from exploiting known bugs or cheating on unpatched servers?

> That string is then compared by the server to its own version string information and calculates if the servers version is newer than the clients version and if depending on the results **sends a response back to the client containing the servers version string and true/false for a client upgrade**.

Daenyth:
I think this isn't the best approach.

By adding the version string negotiation to the protocol, all possible server implementations need to keep their version strings in sync with servatrice, and if they don't, clients will think the server is outdated and doesn't support new features. All possible client implementations need to keep their version strings in sync with desktop cockatrice, or the server will tell them they are outdated even if they support all current features.

For example, suppose the most recent version of cockatrice is 2015-09-01. I'm running Robotrice for Android, version 2015-08-25. It supports all the same features that cockatrice does, but now the server tells me that my client is out of date.

Instead, I propose that we negotiate features in the protocol, not versions. Versions don't matter. If I make a custom compilation from a branch that merges something new in, my version string will be newer, but my feature set might not be different at all - could just be a UI change.

My proposal is instead to add a new field to the login action. The client will inform the server of the features it supports. The server will check this against its own feature set. If the client is missing features, the server will inform the client which features it is missing and the client can display a warning to the user. If the client supports features the server doesn't, the server will inform the client that those features may not work. For the official cockatrice/servatrice, we can add a constant somewhere in `common/`. It should contain only features that relate to the protocol; UI features, server updates, etc, won't matter here. We could potentially include the set into a protobuf type so that other client/server implementations can use the same api.

Example:
All features
- reveal_bottom_card
- register_accounts
- frobnicate_widgets
- send_client_id

Client -> Server
```
features: { reveal_bottom_card, register_accounts, frobnicate_widgets }
```

Server config:
```
{ reveal_bottom_card: optional
  send_client_id: required,
  register_accounts: optional}
```

Server -> client
```
{ client_missing_required_features: [ send_client_id ]
  server_does_not_support: [ frobnicate_widgets ] }
```

Client displays:
> Login unsuccessful. The server requires clients to support the `send_client_id` feature. Update your client.

User updates and client connects again, this time saying it supports `send_client_id`

Client displays:
> Warning: This server does not support the `frobnicate_widgets` feature.

woogerboy21:
How do you propose we maintain the list of features?

woogerboy21:
@tooomm 
I did originally but later on found that the information is already being sent back to the client upon login.

Daenyth:
When we add new protocol actions, we add a new feature to the list and mark it as either required or optional. The required/optional part could be up to server admin even, it could just be a QSet in common.

cockatrice/src/remoteclient.cpp
woogerboy21: Yes I did.... it was late.

cockatrice/src/remoteclient.cpp
woogerboy21: Was following the Qt page for syntax reference.  I could try it without.

woogerboy21:
So you want to pass all the features as a string that has some common delimiter in the protocol right?

Daenyth:
It could be a `repeated string` in the proto

PR #1424, opened 2015-08-26
SoundEngine overhaul

Fizztastic:
TODO: Switch sounds on mac to ~/Library/Application/Support/Cockatrice/Cockatrice
Related #1403 

- Added Qt5 support for .wav/.mp3/.mp4
- Added Qt4 support  for .wav
- Replace .raw sound files by .wav

List of actions that can have sound:
- Untap step, Upkeep step, Draw step, Main 1, Start Combat, _Attack step_, Block step, Damage step, End combat, Main 2, _End step_
- Draw card, Play card, _Tap_, Untap, Shuffle, Life change, Roll dice
- _Player join_, Player leave, Player disconnect, Player reconnect, Player concede
- Spectator join, Spectator leave
- Chat mention, @/all,  private message

Note: only the actions that are in _italic_ have sound out of the box.


ZeldaZach:
I think you should do something similar to the keyboard shortcuts page and revamp the sound editor to allow sounds to be loaded from elsewhere and be played in their spots (instead of uploading files with specific names to a specific location)

Daenyth:
In keeping with the GUI theme idea, we should have set filenames for the various actions and the setting should point to the parent folder of these sounds.

Fizztastic:
@Daenyth that's the current behavior.

ZeldaZach:
@Fizztastic GCC isn't happy with you. You might wanna look into it :)

Fizztastic:
@ZeldaZach you want the QT_VERSION check with 3 loops inside instead of one around and one inside?

cockatrice/src/soundengine.cpp
```diff
@@ -4,48 +4,76 @@
...
+            if(!file.exists())
+                continue;
+            file.open(QIODevice::ReadOnly);
+            file.seek(44);
+            audioCache.insert(fileNames[i], file.readAll());
+            file.close();
+        }
+    #else //QT5
+        for (int i = 0; i < fileNames.size(); ++i) {
+            for (int j = 0; j < ext.size(); ++j) {
```
Daenyth: Instead of the qt4 loop not having `j` why not make `ext` conditional on qt version and for qt4 only contain `wav`

cockatrice/src/soundengine.cpp
```diff
@@ -57,46 +85,157 @@ void SoundEngine::soundEnabledChanged()
...
+        return;
+    }
+
+    audioData[fileName]->setVolume(settingsCache->getMasterVolume());
+    audioData[fileName]->stop();
+    audioData[fileName]->setPosition(0);
+    audioData[fileName]->play();
+#endif
+}
+void SoundEngine::untapStep(){
```
Daenyth: I'm not sure I like the proliferation of methods here that just hide string params. This makes it harder to extend. The step names for example will be known via configuration rather than hardcoding once we start migrating to modular game system support

cockatrice/src/soundengine.cpp
```diff
@@ -57,46 +85,157 @@ void SoundEngine::soundEnabledChanged()
...
+void SoundEngine::spectatorJoin(){
+    playSound("spectator_joined");
 }
+
+void SoundEngine::spectatorLeave(){
+    playSound("spectator_leave");
+}
+
+void SoundEngine::chatMention(){
+    //Add timer
```
Daenyth: What's this about?

cockatrice/src/soundengine.h
```diff
@@ -12,24 +16,54 @@ class SoundEngine : public QObject {
...
+    QMap<QString, QByteArray> audioCache;
     QBuffer *inputBuffer;
     QAudioOutput *audio;
-    QDateTime lastTapPlayed;
-    QDateTime lastEndStepPlayed;
-    QDateTime lastAttackStepPlayed;
+    bool enabled;
+    QStringList fileNames;
+    
+	#if QT_VERSION > 0x050000
```
Daenyth: Weird spacing

cockatrice/src/soundengine.cpp
Fizztastic: Would changing the playSound method to public and eliminating all the intermediate methods to play sound acceptable?

cockatrice/src/soundengine.cpp
Fizztastic: Just forgot to remove the comments. The timer in not needed.

cockatrice/src/soundengine.cpp
Daenyth: I think that will be much easier.

ZeldaZach:
@Fizztastic Can you possibly look into the suggested changes? I'd really like to see this in the newest release :)

cockatrice/src/soundengine.cpp
```diff
@@ -57,46 +85,39 @@ void SoundEngine::soundEnabledChanged()
...
     }
+#else //QT5
+    if (settingsCache->getSoundEnabled()) {
+        qDebug("SoundEngine: enabling sound");
+        enabled = true;
+    }else{
+        qDebug("SoundEngine: disabling sound");
+        enabled = false;
+        for (int i = 0; i < fileNames.size(); ++i) {
+            audioData[fileNames[i]]->stop();
```
ZeldaZach: This seems to crash the client if the sound engine is disabled on launch

ZeldaZach:
So this seems to install to `cockatrice.app/Contents/Resources/sounds/*` instead of `~/Library/Application\ Support/Cockatrice/Cockatrice/sounds/*` on Macs. I'd like to see it in the Library instead to allow for easy editing.

It also seems like the sounds don't work on OS X at all. I've tried the test sound button, but it fails to work. Also, the other WAV files don't work either when doing the appropriate actions.

I've moved different sound files (.mp4, .mp3) to the folder and renamed them appropriate, but they still fail to play, even on client reboot.

PR #1428, opened 2015-08-27
text adjustments in oracle

tooomm:
- removed some info text since oracle imports all cards now and no longer offers sets selection
- added info about tokens in welcome message
- `<br/>` --> `\n` for better compatibility with Transifex
- capitalized `URL` term: http://www.oxforddictionaries.com/definition/english/URL

PR #1432, opened 2015-08-28
Remove docs in favor of wiki

ZeldaZach:
Fix #1427

PR #1434, opened 2015-08-28
Restore xsd files removed in #1428

Daenyth:
ಠ_ಠ


tooomm:
you mean #1432 ;)

What do we need them for?

Daenyth:
They provide API documentation and a method of automated validity checking. You can run a tool using the xsd and it can determine whether a cards.xml or tokens.xml file is valid according to the schema or not

Daenyth:
We could also adjust the card database so that instead of hand coded objects we could parse the xsd files to automatically generate the card classes.

Daenyth:
One low-hanging fruit is adding a headless mode to oracle. We could then run it from travis. The XSD could be used to assert that oracle generates valid card databases.

PR #1439, opened 2015-08-30
Added server/client feature set communication

woogerboy21:
Fix for #1214

This PR implements a feature set that gets communicated between the client and the server.  Upon login the server checks the clients features that are sent and if the operator has specified a particular feature be required, if the client is missing the feature they are denied access.  Also, if the server has more features than the client, the response from login includes the missing features and if the user has a new option checked under the settings is notified of a client update.

Clear as mud? :smile:


cockatrice/src/dlg_settings.cpp
```diff
@@ -246,6 +249,7 @@ void GeneralSettingsPage::retranslateUi()
     highQualityURLLabel.setText(tr("Custom Card Download URL:"));
     highQualityURLLinkLabel.setText(QString("<a href='%1'>%2</a>").arg(LINKING_FAQ_URL).arg(tr("Linking FAQ")));
     clearDownloadedPicsButton.setText(tr("Reset/Clear Downloaded Pictures"));
+    updateNotificationCheckBox.setText(tr("Notify when client updates are available"));
```
tooomm: This is a bit misleading I think.
The client won't inform you in general about a new update.
He will just inform you if you connect to a server running a newer "version" (or just if there are new features? no important bugfixes included?)

brainstorming:
`Allow a server you connect to, to inform you that he runs a newer version than your client`
`Enable notifications if a server supports more features (new version/update available)`
...

woogerboy21: This wording was put in place with the expectation that there would be some type of client update notification system worked in the client other than the feature set communication.  At least that's what @Daenyth was eluding to in the previous update check PR that I put in.  The wording can be changed but I was trying to keep it simple for the user and re-usable at a later date as the only functionality it permits/blocks is the notification message box of a newer version in general available.

cockatrice/src/window_main.cpp
```diff
@@ -795,3 +799,8 @@ void MainWindow::refreshShortcuts()
     aExit->setShortcuts(settingsCache->shortcuts().getShortcut("MainWindow/aExit"));
     aCheckCardUpdates->setShortcuts(settingsCache->shortcuts().getShortcut("MainWindow/aCheckCardUpdates"));
 }
+
+void MainWindow::notifyUserAboutUpdate()
+{
+    QMessageBox::information(this, tr("Information"), tr("A new version of the cockatrice client is available for download."));
```
ctrlaltca: I somewhat like this idea, but imho just telling the user that he needs to update without telling him how could cause confusion. What about adding a servatrice settings containing a configurable url for the "latest client version download page" and adding that link to the server response, so that the clickable link appear in the QMessageBox?
The url could point by default to an url we control, like a page on our wii explaining where releases can be found. 
I can see some security implications on this.. like a malicious server being able to serve an url pointing to malware.. but if you connect to a malicious server, i'm afraid we have a lot of other scenarios where bad things can happen.

woogerboy21: Im not sure I like the idea of presenting a URL, whats others thoughts?  Obviously we can do it.  Seems to me that the user of the client connecting should be smart enough to figure out were to go get there software updates from.

Daenyth: I think if the client itself is giving the url it's OK.

We can add an entry in the help menu to link to download page. We can also put a button on the dialog to go there.

I'm going to make a new repo under the org for weekly builds. We can change the build scripts to automatically publish releases to the github repo, and to upload and attach the builds. Then we'll have a centralized download page. Full releases can still happen on the cockatrice repo releases tab.

ctrlaltca: Right, this will address my initial concerns and avoid the possible security problems.

woogerboy21:
@Daenyth does this have everything you were thinking it needs or should I make some changes?

cockatrice/src/settingscache.h
```diff
@@ -47,6 +47,8 @@ class SettingsCache : public QObject {
     QByteArray mainWindowGeometry;
     QString lang;
     QString deckPath, replaysPath, picsPath, cardDatabasePath, tokenDatabasePath, themeName;
+    QString handBgPath, stackBgPath, tableBgPath, playerBgPath, cardBackPicturePath;
```
woogerboy21: I noticed that this line is no longer used and was possibly removed in the recent theme manager PR commit. @ctrlaltca can you confirm that I can remove this line?  I don't want to add it back if none of the variables are no longer needed.

ctrlaltca: yes, these variables are not used anymore.

cockatrice/src/dlg_settings.cpp
Daenyth: I think the most accurate wording here would be `Allow servers to inform you about unsupported features`.

Honestly I don't think it should be configurable at all - consider the case of reveal bottom card, where if you lack it, someone else can do game actions that you can't see - we shouldn't allow people to miss this information.

And if they're missing required and not optional features, then we have to tell them no matter what the setting.

cockatrice/src/remoteclient.cpp
```diff
@@ -140,6 +145,11 @@ void RemoteClient::loginResponse(const Response &response)
         for (int i = resp.ignore_list_size() - 1; i >= 0; --i)
             ignoreList.append(resp.ignore_list(i));
         emit ignoreListReceived(ignoreList);
+
+        if (resp.missing_features_size() > 0)
+            if (settingsCache->getNotifyAboutUpdates())
```
Daenyth: I'd combine these into one `if`

cockatrice/src/settingscache.h
Daenyth: If you remove the declarations and it compiles it's safe. :D

cockatrice/src/window_main.cpp
```diff
@@ -296,6 +297,9 @@ void MainWindow::serverTimeout()
 void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32 endTime)
 {
     switch (r) {
+        case Response::RespClientUpdateRequired:
+            QMessageBox::critical(this, tr("Error"), tr("Your client does not contain features that the server requires, please update your client and try again."));
```
Daenyth: For troubleshooting the inevitable support requests we should have the message here include the feature names missing.

cockatrice/src/dlg_settings.cpp
woogerboy21: @Daenyth  Well the required features deny access to the server and the user is notified which is not disable-able.  The client side setting to disable the notification only disables the popup if the client is missing non-required functionality informing the user that there is a possible update.

common/featureset.cpp
```diff
@@ -0,0 +1,56 @@
...
+    if (featureList.contains(featureName))
+        featureList.insert(featureName,false);
+}
+
+QMap<QString, bool> FeatureSet::addFeature(QMap<QString, bool> &featureList, QString featureName, bool isFeatureRequired){
+    featureList.insert(featureName,isFeatureRequired);
+    return featureList;
+}
+
+QMap<QString, bool> FeatureSet::identifyMissingFeatures(QMap<QString, bool> featureListToCheck, QMap<QString, bool> featureListToCompareTo){
```
Daenyth: I'd name the params something like `requiredFeatures` and `suppliedFeatures`

common/featureset.cpp
```diff
@@ -0,0 +1,56 @@
...
+    QMap<QString, bool>::iterator i;
+    for (i = featureListToCompareTo.begin(); i != featureListToCompareTo.end(); ++i) {
+        if (!featureListToCheck.contains(i.key())) {
+            missingList.insert(i.key(), i.value());
+        }
+    }
+    return missingList;
+}
+
+bool FeatureSet::isRequiredFeaturesMissing(QMap<QString, bool> featureListToCheck, QMap<QString, bool> featureListToCompareTo) {
```
Daenyth: Can we use something like `std::any_of` here?

http://en.cppreference.com/w/cpp/algorithm/all_any_none_of

Daenyth:
It looks like this expresses the behavior we were talking about. Following updates from this round of feedback I expect it'll be ready for merge after the next review.

We can include a `servatrice --list-features` cli option in another PR.

common/featureset.cpp
woogerboy21: Im not certain this is possible (or rather its already doing that).  Or maybe im miss-understanding the usage suggestion of the any_of.  Given the value(s) being compared are a qmap we first need to find out if any of the value(s) are required (which is the bool in the qmap) and then from there determine if any of those true value key's exist in the feature list of the client being passed.  which is what contains does.

cockatrice/src/window_main.cpp
woogerboy21: Are you certain you want to do this?  It sounds ok for maybe a few missing features but if you get a client that is WAY out of date your message box will be long and ugly.

cockatrice/src/window_main.cpp
woogerboy21: So this can be addressed by another PR when the other items are in place to support the suggestion?

cockatrice/src/window_main.cpp
Daenyth: I think that's fine.

common/featureset.cpp
```diff
@@ -0,0 +1,55 @@
...
+QMap<QString, bool> FeatureSet::identifyMissingFeatures(QMap<QString, bool> suppliedFeatures, QMap<QString, bool> requiredFeatures){
+    QMap<QString, bool> missingList;
+    QMap<QString, bool>::iterator i;
+    for (i = requiredFeatures.begin(); i != requiredFeatures.end(); ++i) {
+        if (!suppliedFeatures.contains(i.key())) {
+            missingList.insert(i.key(), i.value());
+        }
+    }
+    return missingList;
+}
```
Daenyth: (continuing previous discussion here)

I think what you have works, I was just looking for a way to make it clearer. I don't think C++/Qt gives us the tools to make it easy.

In scala I'd do something like...
```scala
val requiredFeatures: Set[String] = serverFeatures.filter { case (k, v) => v }.keySet
val missingFeatures: Set[String] = requireFeatures - clientFeatures // Set difference
if (missingFeatures.nonEmpty) ...
```

Doing it with `any` (`exists` in scala)
```scala
clientFeatures: Set[String] = ???
if (clientFeatures.exists(f => serverFeatures.getOrElse(f, false)) ...
```

The latter might translate (haven't tried to compile something like this)
```C++
QSet<String> clientFeatures;
if (std::any_of(clientFeatures.begin(), clientFeatures.end(), [](??? featureName) { serverFeatures.value(featureName, false) })) { ... }
```

Daenyth: I don't see what you're getting at - I'm not trying to construct a QMap, I'm just returning a bool.

woogerboy21: Right, but isn't your std:any_of expecting the begin / end values to be a std value type?

Daenyth: Oh, I'm not sure. I think it should work, the docs say it's an STL-style
iterator. I haven't tested it.

woogerboy21: Oh , one thing to point out with your example were your setting the differences and then doing something with those differences.  That is already done first by finding out if there is any missing features, and if there is only the list of missing features is passed into the isrequired function to make sure none of the missing features are a required feature.

To your point though, it is iterating through all the server features and if the features value is true (meaning required) it then performs the check against the missing features list that is passed into the function.

cockatrice/src/window_main.cpp
woogerboy21: So I started to look into this.  Can either @Daenyth , @ctrlaltca or @poixen tell me if you can pass an entire PB response container through a signal?  Right now the signal to the windows_main function that does the notification is only sending the response code, reason and end time pb container items and wasnt sure if its possible to send a repeated pb string in the container through a signal.  So Ive attempted to send the entire container but am having no luck.

Daenyth: I would translate the response container into an internal structure as early as possible. The method which generates the alert dialog could take a `QSet<QString>` of missing required features.

common/featureset.cpp
Daenyth: What you wrote works, which is what matters most. We can always change it later, I just wanted to bring up other approaches more for discussion than saying it needed to be changed.

cockatrice/src/window_main.cpp
woogerboy21: Yeah I could do that.

woogerboy21: Hows this:

image

woogerboy21: Only thing not pictured in that snapshot is the first line above the first missing feature now shows "Missing Features: " instead of being blank, and then each feature is listed on its own line.

Daenyth: Works for me.

woogerboy21:
https://files.gitter.im/Cockatrice/Cockatrice/2VKn/thumb/blob.png

ZeldaZach:
What would you like done with this from this point, @woogerboy21 ?

cockatrice/src/window_main.cpp
```diff
@@ -293,9 +294,23 @@ void MainWindow::serverTimeout()
...
+        case Response::RespClientUpdateRequired: {
+            QString formatedMissingFeatures;
+            formatedMissingFeatures = "Missing Features: ";
+            for (int i = 0; i < missingFeatures.size(); ++i)
+                formatedMissingFeatures.append(QString("\n     %1").arg(QChar(0x2022)) + " " + missingFeatures.value(i)   );
+
+            QMessageBox msgBox;
+            msgBox.setIcon(QMessageBox::Critical);
+            msgBox.setWindowTitle("Failed Login");
+            msgBox.setText("Your client does not contain features that the server requires, please update your client and try again.");
```
tooomm: support > contain ?

Daenyth:
Can you squash the two commits before merge?

:+1:

PR #1441, opened 2015-08-30
Fix ban issue

woogerboy21:
An issue while banning was identified recently.  If a moderator opens the ban window, clears the name and address and leaves the clientid field either blank or unchecked the server identified all users containing no client id as the target list of people to place the ban for.

This adds a check in to make sure that the clientid is not blank.


ZeldaZach:
:shipit: On travis complete

Daenyth:
I'm a bit embarrassed I didn't catch that in review...

woogerboy21:
@Daenyth the bug only really happened in a very specific scenario.  But had a pretty large impact when it did :cry:

PR #1444, opened 2015-08-31
Detect if a use is trying to place an invalid ban

woogerboy21:
If a user selects the boxes next to the type of ban to place yet clears the box value's and clicks ok a ban is created that has no username, no ip and no client id in the db table.  This PR will prevent the user from creating the ban by checking the value(s) if the check boxes are selected.

### DIFF
--- a/cockatrice/src/userlist.cpp
+++ b/cockatrice/src/userlist.cpp
@@ -122,6 +122,25 @@ void BanDialog::okClicked()
         QMessageBox::critical(this, tr("Error"), tr("You have to select a name-based, IP-based, clientId based, or some combination of the three to place a ban."));
         return;
     }
+
+    if (nameBanCheckBox->isChecked())
+        if (nameBanEdit->text() == ""){
+            QMessageBox::critical(this, tr("Error"), tr("You must have at value in the name ban when selecting the name ban checkbox."));
+            return;
+        }
+
+    if (ipBanCheckBox->isChecked())
+        if (ipBanEdit->text() == ""){
+            QMessageBox::critical(this, tr("Error"), tr("You must have at value in the ip ban when selecting the ip ban checkbox."));
+            return;
+        }
+
+    if (idBanCheckBox->isChecked())
+        if (idBanCheckBox->text() == ""){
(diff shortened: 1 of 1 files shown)


Daenyth:
I'd like to get the other mods feedback on whether it would be better to just silently ignore. If I enter an IP and blank name, will the error clear the ban box? I might lose the IP if so

woogerboy21:
This was put in place to fix the condition were some one check the checkbox but clears the text in the edit box and leaves the check in place when clicking Ok.  We could change it to silently ignore but thats not the current behavior and yes there has been mod's the past that have made blank bans (im sure by accident). :smile:

Daenyth:
The server should validate. If we only fix the client the bug is still
there for any new clients written

woogerboy21:
Agree'd and I understand were your going with this, but the reason it doesnt work the way you think it should is because when placing a ban there is no response code parsing.  The only dialogue notification that th user placing the ban see's is from the check done at the location that these were placed at.  

I could add the server side check as well.

woogerboy21:
Or another way of saying it is, the ban functionality expects a client side validation to be done since the ban command is actually being processed through child function calls made by the band command call from the client.  Resulting in a response container returning to the server and not directly to the client.

We now have ways to notify a user with an event for situations like this, but in the past there was none :smile:

Daenyth:
I agree we should validate on the client. It's better than handling a
response. Just making sure we get it server side (pretty sure we did)

ZeldaZach:
Instead of returning, why don't you just uncheck the box and continue with the ban?

woogerboy21:
@ZeldaZach 
Because the current behavior expects that if the box is checked the user wants to put a ban for that particular checked item in place.  If you want to silently clear it we can but that could lead to the expectation of a user placing the ban to place a ban outside there expected ban configuration.

woogerboy21:
@Daenyth so looking into this we do have checks in place on the server side code that checks to make sure that the command users name, ip address and client id dont end up to all be empty.

ZeldaZach:
:shipit:

PR #1448, opened 2015-08-31
Log inconsistent session / session table situation

woogerboy21:
If a user session is not successfully cleared from the db a condition can occur that the session does not exist at login to log out but the db table for session information still shows the old session data.  This adds a log entry when the condition occurs.

### DIFF
--- a/common/server.cpp
+++ b/common/server.cpp
@@ -133,6 +133,8 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString 
                 delete se;
 
                 users.value(name)->prepareDestroy();
+            } else {
+                qDebug() << "Active session and sessions table inconsistent, please validate session table information for user " << name;
             }
         }
 


ZeldaZach:
Seems fine to me as it's just an else case added with a logger. :shipit:

common/server.cpp
```diff
@@ -133,6 +133,8 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString
                 delete se;
 
                 users.value(name)->prepareDestroy();
+            } else {
+                qDebug() << "Active session and sessions table inconsistent, please validate session table information for user " << name;
```
Daenyth: Is there any other data we can log here? ids? timestamp?

woogerboy21: So the issue is the fact that a previous (now dead) session exists in the db session table.  So we can not query for any old session data from memory.  In order to query for that stale information I would have to create a function to query the data in the db sessions table.  Would you like that done?  Right now we do not have that type of functionality and im not so sure its really needed since the username is stored in the db sessions table and will be in the log message.

Daenyth: Fair enough. I'd just save the current timestamp along with it and
call it a day.

ctrlaltca: This will end up in server.log, where it already gets prepended with a timestamp. :+1:

Daenyth: :shipit:

PR #1469, opened 2015-09-06
Added the ability or moderation staff to request user ban history

### LINKED ISSUE #174: Add the ability for moderators to request a user's ban history

woogerboy21:
Fix #174 

This PR adds the right click option under the user context menu to request a user's ban history.

image

### DIFF
--- a/cockatrice/src/user_context_menu.cpp
+++ b/cockatrice/src/user_context_menu.cpp
@@ -11,12 +11,18 @@
 #include "gameselector.h"
 #include "pending_command.h"
 
+#include <QtGui>
+#if QT_VERSION >= 0x050000
+#include <QtWidgets>
+#endif
+
 #include "pb/commands.pb.h"
 #include "pb/session_commands.pb.h"
 #include "pb/moderator_commands.pb.h"
 #include "pb/command_kick_from_game.pb.h"
 #include "pb/response_get_games_of_user.pb.h"
 #include "pb/response_get_user_info.pb.h"
+#include "pb/response_ban_history.pb.h"
 
 UserContextMenu::UserContextMenu(const TabSupervisor *_tabSupervisor, QWidget *parent, TabGame *_game)
     : QObject(parent), client(_tabSupervisor->getClient()), tabSupervisor(_tabSupervisor), game(_game)
@@ -32,6 +38,7 @@ UserContextMenu::UserContextMenu(const TabSupervisor *_tabSupervisor, QWidget *p
     aRemoveFromIgnoreList = new QAction(QString(), this);
     aKick = new QAction(QString(), this);
     aBan = new QAction(QString(), this);
+    aBanHistory = new QAction(QString(), this);
     aPromoteToMod = new QAction(QString(), this);
     aDemoteFromMod = new QAction(QString(), this);
 
@@ -49,6 +56,7 @@ void UserContextMenu::retranslateUi()
     aRemoveFromIgnoreList->setText(tr("Remove from &ignore list"));
     aKick->setText(tr("Kick from &game"));
     aBan->setText(tr("Ban from &server"));
+    aBanHistory->setText(tr("View user's &ban history"));
     aPromoteToMod->setText(tr("&Promote user to moderator"));
     aDemoteFromMod->setText(tr("Dem&ote user from moderator"));
 }
@@ -92,6 +100,39 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
     BanDialog *dlg = new BanDialog(response.user_info(), static_cast<QWidget *>(parent()));
     connect(dlg, SIGNAL(accepted()), this, SLOT(banUser_dialogFinished()));
     dlg->show();
+}
+
+void UserContextMenu::banUserHistory_processResponse(const Response &resp) {
+    const Response_BanHistory &response = resp.GetExtension(Response_BanHistory::ext);
+    if (resp.response_code() == Response::RespOk) {
+
+        if (response.ban_list_size() > 0) {
+            QTableWidget *table = new QTableWidget();
+            table->setWindowTitle(tr("Ban History"));
+            table->setRowCount(response.ban_list_size());
+            table->setColumnCount(5);
+            table->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+
+            table->setHorizontalHeaderLabels(
+                    QString(tr("Ban Time;Moderator;Ban Length;Ban Reason;Visible Reason")).split(";"));
+
+            ServerInfo_Ban ban; for (int i = 0; i < response.ban_list_size(); ++i) {
+                ban = response.ban_list(i);
+                table->setItem(i, 0, new QTableWidgetItem(QString::fromStdString(ban.ban_time())));
+                table->setItem(i, 1, new QTableWidgetItem(QString::fromStdString(ban.admin_name())));
+                table->setItem(i, 2, new QTableWidgetItem(QString::fromStdString(ban.ban_length())));
+                table->setItem(i, 3, new QTableWidgetItem(QString::fromStdString(ban.ban_reason())));
+                table->setItem(i, 4, new QTableWidgetItem(QString::fromStdString(ban.visible_reason())));
+            }
+
+            table->resizeColumnsToContents();
+            table->setMinimumSize(table->horizontalHeader()->length() + (table->columnCount() * 5), table->verticalHeader()->length() + (table->rowCount() * 3));
+            table->show();
+        } else
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Ban History"), tr("User has never been banned."));
+
+    } else
+        QMessageBox::critical(static_cast<QWidget *>(parent()), tr("Ban History"), tr("Failed to collecting ban information."));
 }
 
 void UserContextMenu::adjustMod_processUserResponse(const Response &resp, const CommandContainer &commandContainer)
@@ -158,6 +199,7 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
     if (!tabSupervisor->getAdminLocked()) {
         menu->addSeparator();
         menu->addAction(aBan);
+        menu->addAction(aBanHistory);
 
         menu->addSeparator();
         if (userLevel.testFlag(ServerInfo_User::IsModerator) && (tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsAdmin)) {
@@ -177,6 +219,7 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
     aRemoveFromIgnoreList->setEnabled(anotherUser);
     aKick->setEnabled(anotherUser);
     aBan->setEnabled(anotherUser);
+    aBanHistory->setEnabled(anotherUser);
     aPromoteToMod->setEnabled(anotherUser);
     aDemoteFromMod->setEnabled(anotherUser);
 
@@ -238,6 +281,12 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
 
         PendingCommand *pend = client->prepareAdminCommand(cmd);
         connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(adjustMod_processUserResponse(Response, CommandContainer)));
+        client->sendCommand(pend);
+    } else if (actionClicked == aBanHistory) {
+        Command_GetBanHistory cmd;
+        cmd.set_user_name(userName.toStdString());
+        PendingCommand *pend = client->prepareModeratorCommand(cmd);
+        connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(banUserHistory_processResponse(Response)));
         client->sendCommand(pend);
     }
 
--- a/common/pb/moderator_commands.proto
+++ b/common/pb/moderator_commands.proto
@@ -2,6 +2,7 @@ syntax = "proto2";
 message ModeratorCommand {
     enum ModeratorCommandType {
         BAN_FROM_SERVER = 1000;
+        BAN_HISTORY = 1001;
     }
     extensions 100 to max;
 }
@@ -16,4 +17,11 @@ message Command_BanFromServer {
     optional string reason = 4;
     optional string visible_reason = 5;
     optional string clientid = 6;
+}
+
+message Command_GetBanHistory {
+    extend ModeratorCommand {
+        optional Command_GetBanHistory ext = 1001;
+    }
+    optional string user_name = 1;
 }
--- a/common/pb/response_ban_history.proto
+++ b/common/pb/response_ban_history.proto
@@ -1,0 +1,10 @@
+syntax = "proto2";
+import "response.proto";
+import "serverinfo_ban.proto";
+
+message Response_BanHistory{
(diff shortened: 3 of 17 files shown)


cockatrice/src/user_context_menu.cpp
```diff
@@ -94,6 +102,38 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
     dlg->show();
 }
 
+void UserContextMenu::banUserHistory_processResponse(const Response &resp) {
+    const Response_BanHistory &response = resp.GetExtension(Response_BanHistory::ext);
+    if (resp.response_code() == Response::RespOk) {
+
+        if (response.ban_time_size() > 0) {
+            QTableWidget *table = new QTableWidget();
+            table->setWindowTitle("Ban History");
```
tooomm: tr() missing

servatrice/src/serversocketinterface.cpp
```diff
@@ -747,6 +749,54 @@ Response::ResponseCode ServerSocketInterface::cmdReplayDeleteMatch(const Command
...
+{
+    if (!sqlInterface->checkSql())
+        return Response::RespInternalError;
+
+    QString userName = QString::fromStdString(cmd.user_name());
+
+    //lookup users ban history from database
+    QSqlQuery *query = sqlInterface->prepareQuery("select id_admin,time_from,minutes,reason,visible_reason from {prefix}_bans where user_name = :user_name");
+    query->bindValue(":user_name", userName);
+    sqlInterface->execSqlQuery(query);
```
ctrlaltca: The query gets executed twice: here and in the next line (with proper error checking). Remove this one maybe?

servatrice/src/serversocketinterface.cpp
```diff
@@ -747,6 +749,54 @@ Response::ResponseCode ServerSocketInterface::cmdReplayDeleteMatch(const Command
...
+    if (!sqlInterface->execSqlQuery(query))
+        return Response::RespInternalError;
+
+    //build response container information from results
+    QString adminID,adminName,banTime,banLength,banReason,visibleReason;
+    Response_BanHistory *re = new Response_BanHistory;
+    while (query->next()){
+
+        adminID = query->value(0).toString();
+        adminName = "Uknown";
```
ctrlaltca: typo

servatrice/src/serversocketinterface.cpp
```diff
@@ -747,6 +749,54 @@ Response::ResponseCode ServerSocketInterface::cmdReplayDeleteMatch(const Command
...
+        banTime = query->value(1).toString();
+        banLength = query->value(2).toString();
+        banReason = query->value(3).toString();
+        visibleReason = query->value(4).toString();
+
+        //lookup admin name based on user id from ban table
+        QSqlQuery *anamequery = sqlInterface->prepareQuery("select name from cockatrice_users where id = :user_id");
```
ctrlaltca: You could avoid making  second query if you join the two tables in the first query:
```
SELECT A.id_admin, A.time_from, A.minutes, A.reason, A.visible_reason, B. name AS name_admin FROM {prefix}_bans A LEFT JOIN {prefix}_users B ON A.id_admin=B.id WHERE A.user_name = :user_name
```

cockatrice/src/user_context_menu.cpp
```diff
@@ -49,6 +56,7 @@ void UserContextMenu::retranslateUi()
     aRemoveFromIgnoreList->setText(tr("Remove from &ignore list"));
     aKick->setText(tr("Kick from &game"));
     aBan->setText(tr("Ban from &server"));
+    aBanHistory->setText(tr("View users &ban history"));
```
Daenyth: > View *user's* ban history

cockatrice/src/user_context_menu.cpp
```diff
@@ -94,6 +102,38 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
...
+                table->setItem(i, 2, new QTableWidgetItem(QString::fromStdString(response.ban_length(i))));
+                table->setItem(i, 3, new QTableWidgetItem(QString::fromStdString(response.ban_reason(i))));
+                table->setItem(i, 4, new QTableWidgetItem(QString::fromStdString(response.visible_reason(i))));
+            }
+
+            table->resizeColumnsToContents();
+            table->setMinimumSize(table->horizontalHeader()->length() + (table->columnCount() * 5), table->verticalHeader()->length() + (table->rowCount() * 3));
+            table->show();
+        } else
+            QMessageBox::information(static_cast<QWidget *>(parent()), tr("Ban History"), tr("User has never been banned."));
```
Daenyth: This response handling is confusing to me. RespOk means that the server handled it without any errors.

Wouldn't no ban history just be RespOk with 0 ban rows?

common/pb/response_ban_history.proto
```diff
@@ -0,0 +1,14 @@
...
+message Response_BanHistory{
+    extend Response {
+        optional Response_BanHistory ext = 1012;
+    }
+    repeated string admin_id = 1;
+    repeated string admin_name = 2;
+    repeated string ban_time = 3;
+    repeated string ban_length = 4;
+    repeated string ban_reason = 5;
+    repeated string visible_reason = 6;
```
Daenyth: This is confusing. We should add a new `ServerInfo_Ban` type which contains all of the data, then this message just contains `repeated ServerInfo_Ban` instead of all these fields.

servatrice/src/serversocketinterface.cpp
```diff
@@ -747,6 +749,45 @@ Response::ResponseCode ServerSocketInterface::cmdReplayDeleteMatch(const Command
...
 
+Response::ResponseCode ServerSocketInterface::cmdGetBanHistory(const Command_GetBanHistory &cmd, ResponseContainer &rc)
+{
+    if (!sqlInterface->checkSql())
+        return Response::RespInternalError;
+
+    QString userName = QString::fromStdString(cmd.user_name());
+
+    //lookup users ban history from database
+    QSqlQuery *query = sqlInterface->prepareQuery("SELECT A.id_admin, A.time_from, A.minutes, A.reason, A.visible_reason, B.name AS name_admin FROM {prefix}_bans A LEFT JOIN {prefix}_users B ON A.id_admin=B.id WHERE A.user_name = :user_name");
```
Daenyth: Rather than a comment it would be more flexible to have a `getUserBanHistory` method. It's self-explanatory at that point and it provides us an easy method to work with when we get to automated testing.

servatrice/src/serversocketinterface.cpp
```diff
@@ -747,6 +749,45 @@ Response::ResponseCode ServerSocketInterface::cmdReplayDeleteMatch(const Command
...
+    query->bindValue(":user_name", userName);
+
+    if (!sqlInterface->execSqlQuery(query))
```
Daenyth: You already checked this in this method, above

servatrice/src/serversocketinterface.cpp
```diff
@@ -747,6 +749,45 @@ Response::ResponseCode ServerSocketInterface::cmdReplayDeleteMatch(const Command
...
+        banReason = query->value(3).toString();
+        visibleReason = query->value(4).toString();
+        adminName = query->value(5).toString();
+
+        re->add_admin_id(adminID.toStdString());
+        re->add_admin_name(adminName.toStdString());
+        re->add_ban_time(banTime.toStdString());
+        re->add_ban_length(banLength.toStdString());
+        re->add_ban_reason(banReason.toStdString());
+        re->add_visible_reason(visibleReason.toStdString());
```
Daenyth: See my comment in the pb file about working with a ServerInfo_Ban item instead.

cockatrice/src/user_context_menu.cpp
ctrlaltca: I think it's the way it's currently working.

common/pb/response_ban_history.proto
ctrlaltca: :+1:

servatrice/src/serversocketinterface.cpp
ctrlaltca: Maybe you're commenting on a previous version of this PR? this has been already solved in the last commit

cockatrice/src/user_context_menu.cpp
woogerboy21: That is the way it works.  It just has an information box that pops up saying there is no ban history if there is none rather than showing an empty ban history table.

servatrice/src/serversocketinterface.cpp
woogerboy21: Even if you made the actual response repeatable rather than each item this behavior wouldn't change.  You still need to add repeated response information.

servatrice/src/serversocketinterface.cpp
woogerboy21: The method being inside the socket interface?  I was discussing this with @ctrlaltca and the reason its done this way is the return response from the method would contain a fairly large amount of data.  If your returning one or two items no problem, but this is looking to grab 6 column items out of the table with the potential for a large repetitive number of those columns.  What type would you (or @ctrlaltca ) recommend I use to return such data?

common/pb/response_ban_history.proto
woogerboy21: Thinking through this, is this the behavior you want?  I dont think it is.  The server inst keeping any type of constant track of ban history.  The server info pb items are all based around objects that have specific sets of information that are stored and manipulated in memory requiring the information to be retained for periods of time during specific execution points when dealing with multi user bits of information.  So for example the serverinfo_card is keeping track of card information state during a game.  The ban history of a user doesn't really fall into that category.  We are looking for the data for a response from a command. The server won't be keeping any state change of the ban history nor will it be passed to anyone but the person making the command request.

Daenyth: I don't think we have to say that ServerInfo_Foo is only for things live in memory. It's info from the server. Even with your example of ServerInfo_User, that's a case where only working on online users is an accidental limitation rather than design goal.

Regardless, the point is to have the ban data inside a single logical unit. Keeping related data in unrelated arrays joined only by accidently sharing an index is not a good pattern to work with. It's error prone and makes the logic harder to follow.

cockatrice/src/user_context_menu.cpp
Daenyth: I see.

I think I'd rather see the table with 0 rows but if you think this is better UX then leave it. We can always come back to it later once people have used it a bit.

servatrice/src/serversocketinterface.cpp
Daenyth: It seems fine to return `QList<ServerInfo_Ban>`

servatrice/src/serversocketinterface.cpp
Daenyth: I think I misread, carry on.

servatrice/src/serversocketinterface.cpp
Daenyth: To make it more concrete; the DatabaseInterface method can return `QList<ServerInfo_Ban>` and the socket interface method can return the `Response`

woogerboy21: You can do that?  Going to have to try it.

common/pb/moderator_commands.proto
```diff
@@ -17,3 +18,10 @@ message Command_BanFromServer {
     optional string visible_reason = 5;
     optional string clientid = 6;
 }
+
+message Command_GetBanHistory {
+    extend ModeratorCommand {
+        optional Command_GetBanHistory ext = 1001;
+    }
+    optional string user_name = 1;
```
Daenyth: What do I get if I send `GetBanHistory` with no username? All users? That seems like it would be a good base behavior - in the future we will need to add more specific querying (ip, client id, etc)

common/pb/serverinfo_ban.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+message ServerInfo_Ban {
+    optional string admin_id = 1;
+    optional string admin_name = 2;
```
Daenyth: I think admin_id and name should be required; all bans have an admin who placed them.

common/pb/serverinfo_ban.proto
```diff
@@ -0,0 +1,9 @@
...
+    optional string ban_time = 3;
+    optional string ban_length = 4;
+    optional string ban_reason = 5;
+    optional string visible_reason = 6;
```
Daenyth: Could you add some basic doc to these formatted like this: https://github.com/estan/protoc-gen-doc#writing-documentation

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -923,3 +923,31 @@ void Servatrice_DatabaseInterface::updateUsersLastLoginData(const QString &userN
...
+    ServerInfo_Ban banDetails;
+
+    if (!checkSql())
+        return results;
+
+    QSqlQuery *query = prepareQuery("SELECT A.id_admin, A.time_from, A.minutes, A.reason, A.visible_reason, B.name AS name_admin FROM {prefix}_bans A LEFT JOIN {prefix}_users B ON A.id_admin=B.id WHERE A.user_name = :user_name");
+    query->bindValue(":user_name", userName);
+
+    if (!execSqlQuery(query))
+        return results;
```
Daenyth: Log the failure reason?

Can we add the logging-failed-queries behavior directly into `execSqlQuery`? That way we can't forget

Daenyth:
Once my most recent comments are addressed :shipit:

common/pb/moderator_commands.proto
woogerboy21: Hmm... good question, what behavior would you like to see if its blank?

Daenyth: I think returning all bans would be reasonable.

If that's too many, consider pagination keyed by the ban's autoinc id

If that's too much work, keep the proto as `optional` so we can extend the behavior later but have the server reject the command.

woogerboy21: I had never thought about it since the only method for calling to get some one's ban history is from the context menu and they being logged in so the field shouldn't ever end up blank.  But that's not to say some were else the command may be used in the future called else were.  I think my vote is to leave it as is and work the future behavior in when we get there. What was your acronym again?  Something about not writing code for the future or trying to predict the future.... i dont recall. :)

Daenyth: When working with the protocol it's important to consider what it makes possible, not just the current client behavior. In about 5 minutes I can write a client that can send any specific command with any variation of parameters. If some permutation isn't handled and leads to a crash, I've got a DOS script. Anyone could do this.

> What was your acronym again? Something about not writing code for the future or trying to predict the future.... i dont recall. :)
YAGNI - and I agree.

So let's keep the field as optional and then have the server reject the command if it's not present.

woogerboy21: Yeah we dont want to make it required as a ban could be placed with out the username.  Ill check out what happens if the name is blank in the current code.  Im expecting it to return bans that don't have username's associated with them (which can be made).  It wouldnt return all the bans.

Daenyth: if the Cmd username isn't present, returning bans where username == null is reasonable, if perhaps not exceptionally useful. Going with that is fine.

I don't really care about the behavior beyond two points:
- The server should handle it gracefully
- The field stays optional so we don't have to change the protocol later once we want better ban history searches

PR #1471, opened 2015-09-06
get cardback image from theme

ctrlaltca:
The deck editor was still using the hardcoded card back image instead of the themeable one.


Daenyth:
Looks fine :+1:

PR #1473, opened 2015-09-06
Added user analytics table to database

### LINKED ISSUE #629: Create new statistics db table
I would like to see a new table added to the DB that keeps track of any number of stat type information.  The first stat to monitor would be the last login date / time of users.  Another row column idea would be the number of times the kick function was used by a user.  

Any other thoughts to this?

woogerboy21:
Replace / Fix  #629

This PR adds a new table to the database that will be used to store long term analytic information.  It moves the "last_login" column from  the users table and also begins storing the client version that is being passed to the server.  A column for notes is also created to allow the storing annotations moderators / admins may want to place on users.  (note: the client does not have the ability to actually store the note information in client as of yet.)

### DIFF
--- a/servatrice/src/servatrice_database_interface.cpp
+++ b/servatrice/src/servatrice_database_interface.cpp
@@ -881,13 +881,45 @@ void Servatrice_DatabaseInterface::updateUsersClientID(const QString &userName, 
 
 }
 
-void Servatrice_DatabaseInterface::updateUsersLastLoginTime(const QString &userName)
-{
+void Servatrice_DatabaseInterface::updateUsersLastLoginData(const QString &userName, const QString &clientVersion) {
+
     if (!checkSql())
         return;
 
-    QSqlQuery *query = prepareQuery("update {prefix}_users set last_login = NOW() where name = :user_name");
-    query->bindValue(":user_name", userName);
-    execSqlQuery(query);
+    int usersID;
 
+    QSqlQuery *query = prepareQuery("select id from {prefix}_users where name = :user_name");
+    query->bindValue(":user_name", userName);
+    if (!execSqlQuery(query)) {
+        qDebug("Failed to locate user id when updating users last login data: SQL Error");
+        return;
+    }
+
+    if (query->next()) {
+        usersID = query->value(0).toInt();
+    }
+
+    if (usersID) {
+        int userCount;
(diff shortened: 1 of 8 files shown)


woogerboy21:
So for anyone really ... does anyone have a better way to do what Im doing at :
https://github.com/Cockatrice/Cockatrice/pull/1473/files#diff-e8469b96aeb5655c901ad95630a25753R902

Basically I am checking if the client id exists in the table and then placing the correct query (insert or update) but doing so requires an additional DB call.  Would there be a better way to do things OR should I just fire the update query and on failure fire the insert?

ctrlaltca:
In mysql you could use replace (https://dev.mysql.com/doc/refman/5.0/en/replace.html); it's like an insert statement, but if it will fail the insert due to a key collision, it will update the old record.
Anyway, this is a mysql-specific extension to sql. Postgres has UPSERT, sqlite has ON CONFLICT REPLACE, ...

woogerboy21:
I dont like the idea of using replace for a couple of reasons.  First should be obvious, its restricting the app's DB functionality to a single DB.  The second is the fact that a replace does a delete and insert if there is an existing entry that matches the primary key.  If the data in the table is to be used for historical analytical information the last thing I want to do is delete the row and insert a new one with only the log in data since there is additional columns that will not be inserted at the time of log in/out but wont want to be lost.

Daenyth:
We are already restricted to only mysql. It doesn't currently work on postgres or sqlite so I don't particularly care if we use more mysql-isms

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -881,13 +881,43 @@ void Servatrice_DatabaseInterface::updateUsersClientID(const QString &userName,
...
         return;
 
-    QSqlQuery *query = prepareQuery("update {prefix}_users set last_login = NOW() where name = :user_name");
+    int usersID;
+
+    QSqlQuery *query = prepareQuery("select id from {prefix}_users where name = :user_name");
     query->bindValue(":user_name", userName);
-    execSqlQuery(query);
+    if (!execSqlQuery(query))
+        return;
```
Daenyth: Check for error and log it if the query fails.

ZeldaZach:
Reviewing the code base, this seems fine to me.
As long as it was tested by @woogerboy21 i'm fine with this. :shipit:

woogerboy21:
:shipit:

PR #1474, opened 2015-09-07
Do not close message tabs on disconnect

woogerboy21:
This PR is to prevent the tab supervisor from closing private message tabs on disconnect from the server.  This will allow users with less stable connections to not loose there conversations.  

I thought there was an issue / request for this but I couldnt find it in the searching I did.

### DIFF
--- a/cockatrice/src/tab_supervisor.cpp
+++ b/cockatrice/src/tab_supervisor.cpp
@@ -281,11 +281,6 @@ void TabSupervisor::stop()
     while (replayIterator.hasNext())
         replayIterator.next()->deleteLater();
     replayTabs.clear();
-
-    QMapIterator<QString, TabMessage *> messageIterator(messageTabs);
-    while (messageIterator.hasNext())
-        messageIterator.next().value()->deleteLater();
-    messageTabs.clear();
     
     delete userInfo;
     userInfo = 0;


Daenyth:
Just delete the code, don't comment it out.

I think for a proper solution though, we need to disable the text fields so the user doesn't try to chat while disconnected. It should be restored only if they reconnect to the same server.

woogerboy21:
@Daenyth would you like me to make it an option under the settings that users can enable/disable?

Daenyth:
No, I'd make it the only behavior. If they don't want the tab open
when they exit they can close it.

ZeldaZach:
I agree with @daenyth on this subject by having it be the only behavior

ctrlaltca:
What about closing them only if the user manually forced a disconnect but not if the disconnection was due to a network problem? It's just an idea..

Daenyth:
I think we should still leave it open in that case. It harms little
for a normal user workflow. I don't think many people have large
numbers of chat tabs open

ZeldaZach:
@woogerboy21 If you're going to delete the code instead of commenting it out, then please do so at your earliest convenience. If not, then I'm ready to :shipit:

woogerboy21:
@ZeldaZach If @Daenyth wants it deleted ill just delete it.  Ill put the change in tonight after work.
@Daenyth about disabling the box, the reason I didnt do that is even if the user actually attempts to chat when disconnected nothing is displayed.  If your running the client in debug mode you see a debug message about being disconnected but since the way the chatting works is the command is sent and then an response / emit is done to both users with the message not being connected to the server shows nothing in the chat dialogue when not connected and typing is attempted.

PR #1479, opened 2015-09-08
Fix autoconnect

### LINKED ISSUE #1465: Auto-connect function broken
On 60ed9fd, OSX 10.10.5
It worked before, but for some reason it doesn't now. 

Edit: "before" as in - in previous builds. It has never worked for me on 60ed9fd.

ctrlaltca:
fix #1465


ZeldaZach:
Good to go :shipit:

PR #1498, opened 2015-09-10
Reworked SoundEngine overhaul

ctrlaltca:
based on @Fizztastic ’s #1424
This only support wavs, but drops the dependency to qt4 mobility.


ZeldaZach:
Works as expected on OS X.

My only issue with sounds is that they're not sym'd into the `~/Library/Application\ Support/Cockatrice/Cockatrice` directory, but that's an issue for another PR

:shipit:

tooomm:
if it's an issue, please file a ticket... it'll get lost here!

PR #1502, opened 2015-09-11
Fix server crash on banning

### LINKED ISSUE #107: make servatrice support correct user detail reporting for offline users and enable it again
Steps to reproduce

1. Friend someone
1. Have them go offline
1. From user tab right click and hit user details
1. Info box shows them as unreg with no info

woogerboy21:
Fix #107 

This PR also fixes the crash that can occur if the username field is passed as empty spaces for the name. There is a change to trim all the variables that could be used to place a ban and adds a check that if all 3 variables end up blank to not put an invalid ban table row.

This should fix the issue reported by JineteDV .
Was able to reproduce the issue on the testing server.

This also address's the issue reported that when an offline user data is queried no user information is returned.  The cmdGetUserInfo function is updated to query the users data from the user table and return it.  Information such as ip_address for an offline user is not returned but other details should be.

### DIFF
--- a/cockatrice/src/user_context_menu.cpp
+++ b/cockatrice/src/user_context_menu.cpp
@@ -210,7 +210,7 @@ void UserContextMenu::showContextMenu(const QPoint &pos, const QString &userName
         }
     }
     bool anotherUser = userName != QString::fromStdString(tabSupervisor->getUserInfo()->name());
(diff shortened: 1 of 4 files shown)


Daenyth:
Awesome, two fixes in one!

PR #1510, opened 2015-09-12
Added chat history to a room that is displayed on join

### LINKED ISSUE #169: Add a moderator-only chatroom with chat history

woogerboy21:
This PR adds room chat history that is presented on join. A new server info message type has been added to the protocol which is used for retaining room chat history that is now presented to users on join and will be used to also allow the moderator tool for looking up chat history to be developed later.

Currently the server only retains 100 messages on a per room basis.
Note:  Older clients will not have the historic messages formatting in place and the historic messages will be presented in the same text/color as the default room join message till they upgrade there clients.

This should be the last thing need to fix #169 but it may be mentioned some were else as something nice to have (chat room history).

One feature added to the room history is the ability for each room to have its own amount of history defined to it.  See the sample servatrice.ini to see how to define the number of rooms historic chat information to store.  A column in the database rooms table has also been added for those using a database to store room information.

image

### DIFF
--- a/cockatrice/src/chatview.cpp
+++ b/cockatrice/src/chatview.cpp
@@ -12,6 +12,7 @@
 #include "settingscache.h"
 #include "tab_userlists.h"
 #include "soundengine.h"
+#include "room_message_type.h"
 
 const QColor DEFAULT_MENTION_COLOR = QColor(194, 31, 47);
 const QColor OTHER_USER_COLOR = QColor(0, 65, 255); // dark blue
@@ -121,7 +122,7 @@ void ChatView::appendUrlTag(QTextCursor &cursor, QString url)
     cursor.setCharFormat(oldFormat);
 }
 
-void ChatView::appendMessage(QString message, QString sender, UserLevelFlags userLevel, bool playerBold)
+void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, QString sender, UserLevelFlags userLevel, bool playerBold)
 {
     bool atBottom = verticalScrollBar()->value() >= verticalScrollBar()->maximum();
     bool sameSender = (sender == lastSender) && !lastSender.isEmpty();
@@ -129,7 +130,7 @@ void ChatView::appendMessage(QString message, QString sender, UserLevelFlags use
     lastSender = sender;
     
     // timestamp
-    if (showTimestamps && !sameSender) {
+    if (showTimestamps && !sameSender && !sender.isEmpty()) {
         QTextCharFormat timeFormat;
         timeFormat.setForeground(QColor(SERVER_MESSAGE_COLOR));
         if (sender.isEmpty())
@@ -168,8 +169,17 @@ void ChatView::appendMessage(QString message, QString sender, UserLevelFlags use
     // use different color for server messages 
     defaultFormat = QTextCharFormat();
     if (sender.isEmpty()) {
-        defaultFormat.setForeground(Qt::darkGreen);
-        defaultFormat.setFontWeight(QFont::Bold);
+        switch (messageType) {
+            case Event_RoomSay::Welcome:
+                defaultFormat.setForeground(Qt::darkGreen);
+                defaultFormat.setFontWeight(QFont::Bold);
+                break;
+            case Event_RoomSay::ChatHistory:
+                defaultFormat.setForeground(Qt::gray);
+                defaultFormat.setFontWeight(QFont::Light);
+                defaultFormat.setFontItalic(true);
+                break;
+        }
     }
     cursor.setCharFormat(defaultFormat);
 
--- a/common/pb/event_room_say.proto
+++ b/common/pb/event_room_say.proto
@@ -5,6 +5,13 @@ message Event_RoomSay {
     extend RoomEvent {
         optional Event_RoomSay ext = 1002;
     }
+    enum RoomMessageType {
+        Welcome = 1;    // rooms welcome message
+        ChatHistory = 2;    // rooms chat history message
+    }
     optional string name = 1;
     optional string message = 2;
+    optional RoomMessageType message_type = 3;
+    optional uint64 time_of = 4;
+
 }
--- a/common/pb/serverinfo_chat_message.proto
+++ b/common/pb/serverinfo_chat_message.proto
@@ -1,0 +1,16 @@
+syntax = "proto2";
+/*
+ * Chat communication of a user to a target.
+ * Targets can be users or rooms.
+ * These communications are also stored in the DB log table.
+ */
+message ServerInfo_ChatMessage {
+    optional string time = 1; // time chat was sent
+    optional string sender_id = 2; // id of sender
+    optional string sender_name = 3; // name of sender
+    optional string sender_ip = 4; // ip of sender
+    optional string message = 5;  // message
+    optional string target_type = 6; // target type (room,game,chat)
+    optional string target_id = 7; // id of target
+    optional string target_name = 8; // name of target
+}
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -600,7 +600,18 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
 
     Event_RoomSay joinMessageEvent;
     joinMessageEvent.set_message(r->getJoinMessage().toStdString());
+    joinMessageEvent.set_message_type(Event_RoomSay::Welcome);
     rc.enqueuePostResponseItem(ServerMessage::ROOM_EVENT, r->prepareRoomEvent(joinMessageEvent));
+
+    ServerInfo_ChatMessage chatMessage; for (int i = 0; i < r->chatHistory.size(); ++i) {
+        chatMessage = r->chatHistory.at(i);
+        qDebug() << QString::fromStdString(chatMessage.message()).simplified();
+        Event_RoomSay roomChatHistory;
+        roomChatHistory.set_message(chatMessage.sender_name() + ": " + chatMessage.message());
+        roomChatHistory.set_message_type(Event_RoomSay::ChatHistory);
+        roomChatHistory.set_time_of(QDateTime::fromString(QString::fromStdString(chatMessage.time())).toMSecsSinceEpoch());
+        rc.enqueuePostResponseItem(ServerMessage::ROOM_EVENT, r->prepareRoomEvent(roomChatHistory));
+    }
 
     Response_JoinRoom *re = new Response_JoinRoom;
     r->getInfo(*re->mutable_room_info(), true);
--- a/common/server_room.cpp
+++ b/common/server_room.cpp
@@ -2,6 +2,7 @@
 #include "server_protocolhandler.h"
 #include "server_game.h"
 #include <QDebug>
+#include <QDateTime>
 
 #include "pb/commands.pb.h"
 #include "pb/room_commands.pb.h"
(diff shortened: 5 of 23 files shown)


cockatrice/src/chatview.cpp
```diff
@@ -168,8 +168,17 @@ void ChatView::appendMessage(QString message, QString sender, UserLevelFlags use
     // use different color for server messages 
     defaultFormat = QTextCharFormat();
     if (sender.isEmpty()) {
-        defaultFormat.setForeground(Qt::darkGreen);
-        defaultFormat.setFontWeight(QFont::Bold);
+        switch (messageType) {
+            case 1:
```
ctrlaltca: what does 0, 1, 2 means? Please create an enum with these values so that we can use labels instead of magic numbers.

common/pb/event_room_say.proto
```diff
@@ -5,6 +5,12 @@ message Event_RoomSay {
     extend RoomEvent {
         optional Event_RoomSay ext = 1002;
     }
+    enum RoomMessageType {
```
ctrlaltca: whoops, i see the enum is already here, so we just need to use it :)

common/server_room.cpp
```diff
@@ -232,6 +234,16 @@ void Server_Room::say(const QString &userName, const QString &s, bool sendToIsl)
     event.set_name(userName.toStdString());
     event.set_message(s.toStdString());
     sendRoomEvent(prepareRoomEvent(event), sendToIsl);
+
+    ServerInfo_ChatMessage chatMessage;
+    chatMessage.set_time(QDateTime::currentDateTime().toString().toStdString());
+    chatMessage.set_sender_name(userName.toStdString());
+    chatMessage.set_message(s.simplified().toStdString());
+
+    if (chatHistory.size() > 100)
```
ctrlaltca: Does this  deserve a "backlog_rows_number" option in servatrice.ini?

common/pb/serverinfo_chat_message.proto
```diff
@@ -0,0 +1,16 @@
...
+ */
+message ServerInfo_ChatMessage {
+    optional string time = 1; // time chat was sent
+    optional string sender_id = 2; // id of sender
+    optional string sender_name = 3; // name of sender
+    optional string sender_ip = 4; // ip of sender
+    optional string message = 5;  // message
+    optional string target_type = 6; // target type (room,game,chat)
+    optional string target_id = 7; // id of target
+    optional string target_name = 8; // name of target
```
ctrlaltca: It seems you are only using time, sender_name and message by now; Do you plan to use the other fields in a following PR?

woogerboy21: The others will be used when the moderator tool for pulling the db log data out and presenting to a moderator when doing chat history look ups.  I wanted to work in the protocol extension now with this PR but have the ability to use things later since the mod tool is a bit more work.

common/server_room.cpp
woogerboy21: I thought about this, the other area I thought it may fit was when the room is defined.  Thoughts to either?  It doesn't matter to me tbh.

common/pb/serverinfo_chat_message.proto
ctrlaltca: Cool!

common/server_room.cpp
ctrlaltca: Per-room is even better

common/pb/event_room_say.proto
woogerboy21: Done :+1:

woogerboy21:
@ctrlaltca Now there is a config per room options for chat history and enums used as requested..  If you want to double check I didnt bork things over. :)

common/server_protocolhandler.cpp
```diff
@@ -597,8 +597,18 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
...
     Event_RoomSay joinMessageEvent;
     joinMessageEvent.set_message(r->getJoinMessage().toStdString());
+    joinMessageEvent.set_message_type(Event_RoomSay::Welcome);
     rc.enqueuePostResponseItem(ServerMessage::ROOM_EVENT, r->prepareRoomEvent(joinMessageEvent));
 
+    ServerInfo_ChatMessage chatMessage; for (int i = 0; i < r->chatHistory.size(); ++i) {
+        chatMessage = r->chatHistory.at(i);
+        qDebug() << QString::fromStdString(chatMessage.message()).simplified();
+        Event_RoomSay roomChatHistory;
+        roomChatHistory.set_message("[" + chatMessage.time() + "] " + chatMessage.sender_name() + ":" + chatMessage.message());
```
ZeldaZach: Can you add a space after the `":"` -> `": "`

common/server_room.cpp
```diff
@@ -232,6 +234,18 @@ void Server_Room::say(const QString &userName, const QString &s, bool sendToIsl)
     event.set_name(userName.toStdString());
     event.set_message(s.toStdString());
     sendRoomEvent(prepareRoomEvent(event), sendToIsl);
+
```
ctrlaltca: if(chatHistorySize < 1) skip this block of code?

woogerboy21: If we did that, there would never be any chat history added would there not be?  Or an I miss-interpreting the line note?  If it always starts at 0 , then the block of code would be skipped all the time.

ctrlaltca: I mean: if the user set the "roomlist\1\chathistorysize" option to 0 because he doesn't want to retain any log, we don't need to create chatMessage, populate it, add it to chatHistory and the remove it. Just skip lines 238-248

woogerboy21: No I was correct.  So what is happening here is that a message is being sent by a user to the room.  The event above your noted line is the message that needs to make it to the room and the block of code below your note is the block that stores the message inside the PB message container and inserts it into the QList.  So skipping any portion of the code here would be a bad idea as you would either be not placing the event for the users message to make it to the room and since the history size starts at 0, skipping the lower block of code below the comment would never store a message in the history QList.

woogerboy21: Ah got it.

common/server_room.cpp
```diff
@@ -232,6 +234,22 @@ void Server_Room::say(const QString &userName, const QString &s, bool sendToIsl)
...
+    if (chatHistorySize != 0) {
+
+        ServerInfo_ChatMessage chatMessage;
+        QDateTime dateTime = dateTime.currentDateTime();
+        QString dateTimeString = dateTime.toString("MM/dd/yy hh:mm:ss");
+        chatMessage.set_time(dateTimeString.toStdString());
+        chatMessage.set_sender_name(userName.toStdString());
+        chatMessage.set_message(s.simplified().toStdString());
+
+        if (chatHistory.size() > chatHistorySize)
```
ctrlaltca: From my tests i always receive chatHistorySize+1 messages in the backlog (eg. if in the config i set chathistorysize=3, an user joining receives 4 messages). I think this is due to the fact that here you are using > instead of >=.

ctrlaltca:
Code looks fine and works as expected. There's a minor issue to be solved, then it's :+1: for me.

common/pb/serverinfo_chat_message.proto
```diff
@@ -0,0 +1,16 @@
...
+/*
+ * Chat communication of a user to a target.
+ * Targets can be users or rooms.
+ * These communications are also stored in the DB log table.
+ */
+message ServerInfo_ChatMessage {
+    optional string time = 1; // time chat was sent
+    optional string sender_id = 2; // id of sender
+    optional string sender_name = 3; // name of sender
+    optional string sender_ip = 4; // ip of sender
```
Daenyth: Whoa, why are we sharing the sender ip? This is restricted to mod users only right?

common/server_room.cpp
```diff
@@ -232,6 +234,22 @@ void Server_Room::say(const QString &userName, const QString &s, bool sendToIsl)
     event.set_name(userName.toStdString());
     event.set_message(s.toStdString());
     sendRoomEvent(prepareRoomEvent(event), sendToIsl);
+ 
+    if (chatHistorySize != 0) {
+
+        ServerInfo_ChatMessage chatMessage;
+        QDateTime dateTime = dateTime.currentDateTime();
+        QString dateTimeString = dateTime.toString("MM/dd/yy hh:mm:ss");
```
Daenyth: Is this format string used elsewhere? Maybe a chance to make it constant

servatrice/migrations/servatrice_0008_to_0009.sql
```diff
@@ -0,0 +1,6 @@
+-- Servatrice db migration from version 8 to version 9
+
+alter table cockatrice_rooms add chat_history_size int(4) not null after join_message;
+update cockatrice_rooms set chat_history_size = 100;
```
Daenyth: :+1: 

Can you add a message about this in the commit?

We need to notify server admins of the new configuration possibility

common/pb/serverinfo_chat_message.proto
woogerboy21: The protocol buffer message is designed around what data is stored in the chat log database table.  The message was created primarily for using the information when working with the log data stored in the db and the thought was for when we work with querying the data from the DB table for historic log lookup (mod only use).  The room history functionality uses the message container as a method to store the fields its needs into a qlist stored in memory but only assigns data to a couple of the message fields.  The qlist is then returned to the client by a room say message on join.

common/server_room.cpp
woogerboy21: It shouldn't be used any were else.  The MM/dd/yy format was added to show the historic date of the message for rooms that may have messages dating back in the days rather than minutes.  THe example I told @ZeldaZach was I can see a main room that may only have 20 messages that show up as history which the date wont really matter.  But in a mod only room we may  have 100+ messages all stated over a much longer day/week time span.

servatrice/migrations/servatrice_0008_to_0009.sql
woogerboy21: Done.

common/server_room.cpp
woogerboy21: Yup.  I can fix that.

common/pb/serverinfo_chat_message.proto
Daenyth: Gotcha, that makes sense.

I don't have a problem with using it this way, but it is always free to add new optional fields later.

common/server_room.cpp
ctrlaltca: Can i suggest Qt::ISODate ? mm/dd/yy is used only in the USA (https://en.wikipedia.org/wiki/Date_format_by_country)

woogerboy21: I actually had it set to show the originally formated QDateTime currentDateTime output.  But was asked by @ZeldaZach to change it to the format that you see now.  I don't really care myself.  I can change it to whatever.  What would everyone like?

ZeldaZach: How about `d mmm yy HH:ii:ss`? ("14 Sep 2015 13:33:22")

ZeldaZach:
After a discussion with Dae, a question emerged:

Is the server set in UTC and clients translate the time in client to their local times? And if not, is this a viable idea?

woogerboy21:
I could convert it to UTC time and then have the client do some type of formatting on its side to try and translate it out.  Sounds like a pain to me over something so small.

Daenyth:
I haven't dug into it yet, but what is the server storing if not UTC? On the client side that UTC needs to be translated to the local time, or you'll end up relaying false information to the users, or cases where the hour/minutes etc show up as times from the future (in local time perspective)

Daenyth:
Looks like the code stores server time, which should be UTC, in the `chatHistory` list. So the clients should treat the datetime they receive as utc and translate it to local time. Which still isn't entirely accurate since it refers to past events and the time zone definition may have changed between the event being recorded and being displayed, but it's close enough for our uses.

woogerboy21:
@Daenyth now passes the date of the room say message (for history) as int64 and client side converts it to local time

Daenyth:
:shipit:

PR #1512, opened 2015-09-13
sound themes

ctrlaltca:
This PR introduces the concept of sound themes. Sound themes works for sound files the same way themes works for images.
Two themes are included: Default (current theme) and Legacy (cuckoo!).


ZeldaZach:
This works as expected. Some ideas on it:
* It seems that different default themes have different sounds times (i.e. On "Default" there is sound for going to combat, while on "Legacy" there isn't) -> Maybe we can merge them or something?
* When you tap/untap multiple cards at once, the sound is played louder and louder until when you tap so many the sound just cracks and sounds bad. Maybe we can limit how many sounds play per action taken? (For mass tap/untapping)

ctrlaltca:
> It seems that different default themes have different sounds times (i.e. On "Default" there is sound for going to combat, while on "Legacy" there isn't) -> Maybe we can merge them or something?

I just created two "themes" that reflects the sounds actually present in cockatrice now and what used to be until a few months ago. I think nobody really cares a lot about sounds, but having a new sound theme covering all the available sounds slots would be nice.

> When you tap/untap multiple cards at once, the sound is played louder and louder until when you tap so many the sound just cracks and sounds bad. Maybe we can limit how many sounds play per action taken? (For mass tap/untapping)

Let me check

ZeldaZach:
@ctrlaltca Can you fix conflicts?

ctrlaltca:
sure; wait, i'll add a fix for #1514 aswell

ZeldaZach:
No problem, thanks!

ctrlaltca:
Ok, this should be ready for testing now.

Daenyth:
Can we squash commits down before merging next time? It makes the history easier to manage for other contributors + tools like bisect

ctrlaltca:
Yep, sorry.

Daenyth:
It's not a big deal :)

PR #1522, opened 2015-09-16
Added official warning messages moderators can use to inform users

### LINKED ISSUE #175: Official warnings
Right now, to warn someone you have to either send them a private message, tell them in the main chat, or create a 1-minute warning ban. None of those remain permanently, so if another mod sees the same user doing the same offense, he can't know if he had been warned before.

Some kind of warning list that could be seen in user details (only for mods, of course) would make it easier to know if someone is a repeated ofender. For example, you could look at someone's user details and see:
-(1) Spam warning.
-(2) Foul language warnings.

Or something like that. Warnings/infractions could expire in something like a month. If someone gets banned without a warning, that should create a warning for him too.

Users should also be able to see their active warnings when they log on, because right now, if someone logs off before you can talk to them, you have to create a long ban to make sure they get the message.

woogerboy21:
Fix #175

So here is a PR to add official warning messages that a moderator/admin can send to a user warning them of misconduct. The list of warnings are configurable by the server operator using the servatrice.ini

How it works - 
When the moderator wants to warn a user they can right click the users name in the userlist and choose warn user. A command is sent to the server that first collects the users information and once collected the couple of fields that is needed are extracted. Then a second command is sent to the server requesting the list of warnings that the operator has defined in the servers ini file. Once that is collected the warning dialog is presented listing the warnings that the operator has available on the server and allows the moderator to select the desired warning and issue it to the end user.

Once issued, the warning is then stored in the warnings table of the database and once successfully stored the warning event is sent to the user for the warning dialog to be shown. An event is also sent to all online moderator/admins that display a private message by the server to inform the online staff which user has been warned and for what reason.

A right click menu option to get the users warn history has also been added for moderators to be able to view a users warn history and determine if the user needs a harsher punishment.

Please let me know if there is any changes that need made to the list.
Please review heavily and make suggestions on changes. 

image


ZeldaZach:
My initial thoughts:

I don't really like how it says which moderator warned you, as you can then take a grudge out against them. Maybe leave that to internal only?

I'd like there to be an additional field for "internal notes" so when a mod sends a warning for "spamming" we can then internally record it as "This user was spamming with racist stuff and just general spam with letters" so we can have a better idea of the true offenses they committed and not just a canned reason internally.

Furthermore, I think warnings should include a link to the TOS of the server, which can be set in the config.ini for servatrice.

woogerboy21:
>I don't really like how it says which moderator warned you, as you can then take a grudge out against them. Maybe leave that to internal only?

The idea behind this was so that if the user had questions they knew whom to ask.  If others think it should be removed it can be done.

> I'd like there to be an additional field for "internal notes"

Whats everyone's thoughts here as well.  Personally a warning is a warning.  To your point there could be levels of severity that a a particular type of warning may have but in the end no matter if the user is being polite when they are misbehaving the warning should be treated as they are doing just that, misbehaving.  Once again I could be persuaded to add the option.

tooomm:
If there is a connection to the time and/or the place (room) the misconduct happened every mod can have a look into the logs to check for reasons/details on the warning when needed.

ZeldaZach:
@tooomm I don't believe so. Each instance is on its own, which is why I'd like to add a way for mods to write quick internal notes on the subject at hand.

woogerboy21:
@tooomm @ZeldaZach 
There is a way for the moderator to look historically at what the user said and were it was said.  They can review the logs.  Currently that functionality is not in client but I hope to have it soon.  Regardless of that can either one of you explain a scenario were that exactly matters?

The way I see it, as an operator the warnings are there as the very lowest form of discipline.  If I am operating a server (or servers) and a user is warned I don't really care if its in a private chat, a room, or game.  To say that you are making a decision on specific information about a previous warning leads me to believe there is inconsistent disciplinary actions being taken among the moderation team and is a problem.

Here's what I mean.  If there are 5 moderators on and some one is using foul language.  A moderator will make a decision on how harsh of an action is to be taken to discipline the user.  They will take into account maybe past number of warnings for foul language, or even past ban history.  But what of those are they looking at?  They should be looking at the reason. If they take into account things like notes on a ban and decide you know what, this time the user isn't doing anything as bad as last time (even though he's been banned 5 times in the past for similar reasons) I'll not ban him simply leads to confusion on the end uses part.  

Now they can get away with something because if I act really bad the first time and get banned, do a similar act but the act is not as bad as the last time I get warned even though I am still acting out seems like a problem in how the actions of the moderation staff are.  Then you end up in scenario's were players say "this mod is terrible and that mod isnt".  Or when XXX mod is around I can act one way but if YYY mod is around I can act another.

The correct way things should be handled are, the user is using foul language and is either spotted in a room chat or reported on.  The moderator reviews the reported information (log lookup) and if found to be true takes into account the severity of the action.  If its deemed to be a warning type offense review the prior warnings and if they have been warned numerous times for foul language a harsher action of punishment should occur.  It makes no difference is on Tuesday of last week in a room the user used the work fuck 5 times.  Or the specifics to the previous offense are stored in the warn/ban table.  The user is acting against the code of conduct and should be disciplined accordingly.

Now if there is a scenario you can explain or reason beyond what I am seeing I am all about hearing it.

ZeldaZach:
Based on what you've said @woogerboy21 I'm fine with leaving this as is and just moving forward.

That, and I don't have the time to write a 2 page response to yours, heh

woogerboy21:
>That, and I don't have the time to write a 2 page response to yours, heh

I would like to hear it if you eventually do get time.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -953,3 +953,46 @@ QList<ServerInfo_Ban> Servatrice_DatabaseInterface::getUserBanHistory(const QStr
...
+    if (!checkSql())
+        return false;
+
+    QSqlQuery *query = prepareQuery("insert into {prefix}_warnings (user_name,mod_name,reason,time_of,clientid) values (:user_name,:mod_name,:warn_reason,NOW(),:client_id)");
+    query->bindValue(":user_name", userName);
+    query->bindValue(":mod_name", adminName);
+    query->bindValue(":warn_reason", warningReason);
+    query->bindValue(":client_id", clientID);
+    if (!execSqlQuery(query))
+        return false;
```
ZeldaZach: Something tells me @Daenyth would want an error report generated if this fails

Daenyth:
I haven't dug into the code yet but is this configurable? We should allow this to be set by the server operator and produce a string directly to the users. I don't want to encode one server's practices into the code base and protocol. They won't apply to all servers.

woogerboy21:
@Daenyth 
That is not the way this works as it would break one of the primary goals.  One of the reasons for having a pre-canned / pre-defined list of warnings is so they can be translated.  Having the various settings in say something like the servers ini file would eliminate that possibility and make things work no differently than the ban functionality that exists today (with the exception of the booting from server part.)

Right now things are in the protocol to allow possible comparisons against the available warning reasons.  This in turn results in a need for a translation to be done from enum to string which is what I had mentioned in the original PR post about if we really wanted to keep it that way or not.  I can remove the enum and simply pass strings back and forth between client / server , which in turn would make life easier to add warnings in the future but eliminate the ability to do things like switch statements.  I'm ok with that if you want me to remove the enum, but I would prefer to not loose the translation portion.

Thoughts?

Daenyth:
Translation on the server side wouldn't work because the server doesn't (and shouldn't) know the configuration on the client. I don't think translation is valuable goal for ban messages because each server has its own primary language which operators use consistently and users are expected to understand.

At the end of the day I feel like these options are part of the woogerworks server rules, not all cockatrice servers. Additionally, adding it to the protocol imposes a burden on every future client which I think is unreasonable. I don't see the problem with the current banning system.

I'd be open to having canned responses that exist solely client-side for moderator use. Then the server operator can tell moderators what to use. The moderator could configure them to suit their needs.

I don't think baking your moderation guidelines into the protocol forever is reasonable. Further, once it's in it can't be changed, so updates moderation guidelines have to become even more protocol extensions. In the PR comment you even note this, that guideline changes need to become code changes.

We should strongly err on the side of using our existing tools to solve problems rather than introducing new tools which have to be maintained in perpetuity.

If the goal is to have consistent moderation messaging, then let's solve that using the tools we have - how about the ability to configure canned ban messages (and/or durations) on the client side. If we instantiate a second QSettings object we can even put that configuration in a dedicated ini file which you as the server owner could define once and share to mods. This reduces the maintainance burden significantly, and when your server rules evolve you don't need code changes.

Consider the huge long tail of upgrades we see - people are frequently on months-old clients. If the ban info is baked into the protocol they'll just perceive a crash and report it to you. And forcing them to update is always a bad UX and should be an absolute last resort.

woogerboy21:
I have no interests in changing the banning functionality. This PR is to address issue #175.  The translation portion is strictly for warning messages.  Warning a user should be something that the user can understand in there own native language as it leads to something more severe of a punishment.  The translations that are being done is in common in there own class which results in the translations happening on both sides (client and server) and is simply a solution to allowing a enum value to be converted to a string for display/storing purposes.  As mentioned, I can remove this type functionality and place things solely on the client side. 

Now as far as your idea on solving the problem client side.  Could you elaborate more?  With out having a custom client built for say woogerworks compared to some were like poixen how would some one go about getting any canned client side translatable messages? Anything in an external source file is not translatable.  In your example you suggest sharing a file with the moderators that would create a more consistent message that they could load into there client and that still does not accomplish one of the primary goals of canned warning notifications.

Daenyth:
Repeating some stuff I said in gitter chat: I'd like to keep protocol additions as flexible as possible. Adding an enum for something like that adds burden to all future clients and makes our system harder to develop for.

I want to get warnings in, even canned ones, but I want to keep the proto as lean as possible.

To that end, here's a proposal:

- Start with a generic "send warning to user" where the moderator can put in whatever warning is appropriate.
- Introduce a command and response for the client to request canned warning messages. These would be configurable, and we could provide a default set. That way if a server admin wants their own, they can. And they can provide them in the server's official language.
- Client could then have an ability to fetch these and present a list to the mod.

-----

That being said, from my moderation experience I don't think this is all that useful. People generally violate server guidelines because they don't care. The ones who are just uninformed correct their behavior after one brief PM chat. Further it's the chat that has the ability to really correct things on a larger scale. If someone is behaving inappropriately, an app-generated warning will just annoy them. A chat has the opportunity to bring their viewpoint around in a way canned responses absolutely cannot. So from a community health perspective I'd rather see more of norm setting and consensus building *which can then be enforced by the whole community and not just mods*, instead of authoritarian measures. And we can still fall back to bans as warnings, which I find totally appropriate.

If you're really dead set on the canned thing, how about the ability to have canned PMs we can send to users to kick off a conversation? It wouldn't need any protocol changes from here.

woogerboy21:
@Daenyth The problem with PM's is they can be (and will be) ignored.

woogerboy21:
Alright, things should be fixed up the way requested (but still will need review).  So the functionality now operates like this.  The server operator can now define a list of warning types in the servatrice.ini.  When the moderator wants to warn a user they can right click the users name in the userlist and choose warn user.  A command is sent to the server that first collects the users information and once collected the couple of fields that is needed is extracted.  Then a second command is sent to the server requesting the list of warnings that the operator has defined in the servers ini file.  Once that is collected the warning dialog is presented listing the warnings that the operator has available on the server and allows the moderator to select the desired warning and issue it to the end user.

The warning is then stored in the warnings table of the database and once successfully stored the warning event is sent to the user for the warning dialog to be shown.    An event is also sent to all online moderator/admins that display a private message by the server to inform the online staff which user has been warned and for what reason.

So the drawbacks at this point are:
* Translations do not occur for the end user, they must know English to read the dialog.

servatrice/servatrice.ini.example
```diff
@@ -56,6 +56,11 @@ requireclientid=false
 ; attempting to log in.  Example: "client_id,client_ver"
 requiredfeatures=""
 
+; You can define custom warnings that users are sent when the moderation staff uses the right client warn user
+; menu option.  This list is comma seperated that each item will appear in the drop down list for staff members
+; to choose from.  Example: "Flaming,Foul Language"
+officialwarnings=""
```
ZeldaZach: Maybe we can include some defaults?

servatrice/src/serversocketinterface.cpp
```diff
@@ -849,7 +917,7 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
     QList<QString> moderatorList = server->getOnlineModeratorList();
     QListIterator<QString> modIterator(moderatorList);
     foreach(QString moderator, moderatorList) {
-        QString notificationMessage = "A ban has been put in with the following details:";
+        QString notificationMessage = moderator + "has placed a ban with the following information:";
```
ZeldaZach: This doesn't look right... It seems like each mod will see that they placed the ban.

ZeldaZach:
This PR is up to my level of satisfactory. My review turned up nothing major, so I think this is ready to go.

Pending review by @Daenyth

servatrice/servatrice.ini.example
woogerboy21: What recommendations would you want for a default list?

ZeldaZach: Flamming, Spamming, Causing Drama, Abusive Language...

cockatrice/src/user_context_menu.cpp
```diff
@@ -102,6 +109,39 @@ void UserContextMenu::banUser_processUserInfoResponse(const Response &r)
...
 }
 
+void UserContextMenu::warnUser_processGetWarningsListResponse(const Response &r)
+{
+    const Response_WarnList &response = r.GetExtension(Response_WarnList::ext);
+
+    QString user = QString::fromStdString(response.user_name()).simplified();
+    QString clientid = QString::fromStdString(response.user_clientid()).simplified();
+
+    // The dialog needs to be non-modal in order to not block the event queue of the client.
```
Daenyth: Good comment!

common/pb/moderator_commands.proto
```diff
@@ -24,4 +27,30 @@ message Command_GetBanHistory {
...
+}
+
+message Command_GetWarnHistory {
+    extend ModeratorCommand {
+        optional Command_GetWarnHistory ext = 1003;
+    }
+    optional string user_name = 1;
+}
+
+message Command_GetWarnList {
```
Daenyth: I'm having a hard time understanding what this is for vs GetWarnHistory. Can you explain the difference? Comments on these would help

common/pb/serverinfo_warning.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto2";
+/*
+ * Historical warning information stored in the warnings table
+ */
+message ServerInfo_Warning {
```
Daenyth: :+1:

servatrice/migrations/servatrice_0009_to_0010.sql
```diff
@@ -0,0 +1,12 @@
+-- Servatrice db migration from version 9 to version 10
+
+CREATE TABLE IF NOT EXISTS `cockatrice_warnings` (
+  `user_name` varchar(255) NOT NULL,
```
Daenyth: These should be by id instead of by name, I'd think, since it's an FK to the users table

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -940,7 +940,7 @@ QList<ServerInfo_Ban> Servatrice_DatabaseInterface::getUserBanHistory(const QStr
         return results;
     }
 
-    QString adminID,adminName,banTime,banLength,banReason,visibleReason;
+    //QString adminID,adminName,banTime,banLength,banReason,visibleReason;
```
Daenyth: Delete this line?

servatrice/migrations/servatrice_0009_to_0010.sql
ZeldaZach: I see your point as users can change their usernames overtime and I do agree we should probably store user ID > username

servatrice/src/servatrice_database_interface.cpp
ZeldaZach: +1

common/pb/moderator_commands.proto
ZeldaZach: ^ Yes

Daenyth:
Once my points are addressed :shipit:

servatrice/src/servatrice_database_interface.cpp
woogerboy21: Yeah, it was left over by accident from some prior cleanup of the code work.

common/pb/moderator_commands.proto
woogerboy21: @Daenyth : yes that is correct

woogerboy21:
Alright, the code has been updated to store the clientid and also the lookup is done by client id now as well.  That way if the user changes there name the lookup will still show the previous warn history of the old username along with new username.

ZeldaZach:
:shipit:

Daenyth:
Wait, but client id can be changed, we need to be storing USER id.

This was not correct to merge. We need to create a new PR with a migration to repair the table now.

Issue #1334, opened 2015-08-04
settings window always saves changes immediatly


tooomm:
- go to "Settings"
- go to "Chat"
- mess around with the color code for mentions
- enjoy your new color - you'll never get your old back unless you remember ;)

The problem is here, that the saving does not happen while clicking "OK", it immediately happens after you made the change!
So if you want to revert a change or try out to preview a new color etc. you just can't click "X" and start over... (there is no "Cancel" button as well).
Other case is the language, you switch to a weird language by fault and might struggle to change it back.

Daenyth:
I'd like to keep the autosave but a "revert" button to reset the state of settings back to what they were when the window opened would be nice.

In general I prefer a make-forgiveness-easy workflow instead of an ask-permission workflow. "Are you sure" boxes are terrible, and having it autosave makes the most common case (making a change) more straightforward.

tooomm:
That's what I'm talking about... taken from [Mozilla Development Network](https://developer.mozilla.org/en-US/Add-ons/Overlay_Extensions/XUL_School/Handling_Preferences#Preference_windows)

>The general philosophy in non-Windows systems is that a change in a preference applies immediately. Preference windows don't have any buttons, or just an OK or Close button. On Windows, changing preferences don't apply until the user click on the OK button. The user can click on the Cancel button and none of the changes performed in the window will apply. This is why it makes sense to have Preference windows be modal on Windows. This way the user is urged to apply or discard any changes instead of being able to ignore the Preferences window.

Daenyth:
I'd much rather auto save with an undo button instead of confirming save,
because it streamlines the normal case

tooomm:
It'll confuse every single windows user though.
Autosave+revert will introduce a totally new concept to them as well! Windows user have the "Cancel" button to undo/reset their changes (aka exit without applying). ;)

They happen to be our biggest user group, and maybe even have the highest number of people which are not interested in tech as well. 

We could just name the "undo" button `Cancel` on windows as a workaround... or even just display it on win systems if it's not common on mac at all to have a "revert" button?!
Well, or the other way around. If it's easier to track made changes and revert them instead collecting them and applying them all together...

Daenyth:
I don't think it's that surprising. It's standard for web apps to behave
like this too.

I think the main thing is a way to not lose your starting settings, however
we end up doing it

ZeldaZach:
[...]

PR #1523, opened 2015-09-16
Restore shortcuts

marcofernandezheras:
resetshortcuts

Added Link to faq, reset & clear all shortcuts buttons

### DIFF
--- a/cockatrice/src/sequenceEdit/shortcutstab.cpp
+++ b/cockatrice/src/sequenceEdit/shortcutstab.cpp
@@ -26,30 +26,36 @@ ShortcutsTab::~ShortcutsTab()
 
 void ShortcutsTab::resetShortcuts()
 {
-    settingsCache->shortcuts().resetAllShortcuts();
+    if(QMessageBox::question(this,tr("Restore all default shortcuts"),
+                             tr("Do you really want to restore all default shortcuts?")) == QMessageBox::Yes)
+    {
+        settingsCache->shortcuts().resetAllShortcuts();
+    }
 }
 
 void ShortcutsTab::refreshEdits()
-{
+{    
     QList<SequenceEdit*> edits = this->findChildren<SequenceEdit*>();
     for(int i= 0; i < edits.length(); ++i)
     {
         edits.at(i)->refreshShortcut();
     }
-    QMessageBox::information(this,tr("Shortcuts reset"),tr("All shortcuts have been reset"));
 }
 
 void ShortcutsTab::clearShortcuts()
 {
-    settingsCache->shortcuts().clearAllShortcuts();
+    if(QMessageBox::question(this,tr("Clear all default shortcuts"),
+                             tr("Do you really want to clear all shortcuts?")) == QMessageBox::Yes)
+    {
+        settingsCache->shortcuts().clearAllShortcuts();
+    }
 }
 
 void ShortcutsTab::afterClear()
-{
+{    
     QList<SequenceEdit*> edits = this->findChildren<SequenceEdit*>();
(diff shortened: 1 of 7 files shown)


Daenyth:
What's the difference between clear and reset? As a user that's confusing. Does it unbind everything? When is that useful?

marcofernandezheras:
Clear is to delete all shortcuts, reset is to put all shortcuts to default.
Clear is usefull if you want a fresh start

Daenyth:
I see. That makes a little sense.

ZeldaZach:
I like it

tooomm:
`Reset all (shortcuts) to default` as label instead?

ZeldaZach:
So this works fully as expected. I don't really see an issue with the translation of the buttons as @tooomm but I'm open to hearing your opinion, @marcofernandezheras.

This is ready to go whenever we get wording done properly :shipit:

cockatrice/src/sequenceEdit/shortcutstab.cpp
```diff
@@ -16,3 +23,33 @@ ShortcutsTab::~ShortcutsTab()
...
+}
+
+void ShortcutsTab::refreshEdits()
+{
+    QList<SequenceEdit*> edits = this->findChildren<SequenceEdit*>();
+    for(int i= 0; i < edits.length(); ++i)
+    {
+        edits.at(i)->refreshShortcut();
+    }
+    QMessageBox::information(this,tr("Shortcuts reset"),tr("All shortcuts have been reset"));
```
tooomm: "All shortcuts have been reset **to their defaults**"

cockatrice/src/sequenceEdit/shortcutstab.cpp
```diff
@@ -16,3 +23,33 @@ ShortcutsTab::~ShortcutsTab()
...
+}
+
+void ShortcutsTab::afterClear()
+{
+    QList<SequenceEdit*> edits = this->findChildren<SequenceEdit*>();
+    for(int i= 0; i < edits.length(); ++i)
+    {
+        edits.at(i)->clear();
+    }
+    QMessageBox::information(this,tr("Shortcuts reset"),tr("All shortcuts have been cleared"));
```
tooomm: We have 2 messages boxes here with the same name/title description, but they do two completely different things.

Other languages use several terms for this different actions.
That will bring some trouble for translators and makes the UI less simple.

tooomm:
Is there some kind of confirmation dialog?

I don't like to miss click something and completely throw away my settings.
Since the settings window stores all changes immediatly (#1334) and doesn't wait until the `OK` button is clicked to close the settings dialog window AND save all changes, I like to see a confirmation dialog here for clearing/restoring shortcuts.

Daenyth:
If we don't have an "undo" action we should have a confirmation. I
prefer having no confirmation but with "undo" available, but that's a
bit harder to write.

tooomm:
I've barely seen a "undo" action in any application... ;P
A confirmation is not too embarrassing if you ask me and a nice tradeoff.

Nightfirecat:
I'm in agreement that there should be either a confirmation for resetting/deleting all shortcuts or a simple method to undo it. My personal preference is for a confirmation dialog, since it's more clear and doesn't require any considerations for UI/UX for pointing a user to the "undo" functionality.

marcofernandezheras:
Change button label to "Restore all default shortcuts"

ZeldaZach:
Very nice job @marcofernandezheras 

I don't really care for the popup after the fact ("Your shortcuts have been cleared/reset to default!"). I find is unnecessary.

Otherwise, this is perfect!

PR #1527, opened 2015-09-17
Moderator ban notifications + Server PM functionality

### LINKED ISSUE #668: Servatrice should inform mod users of all bans
When servatrice handles a ban command, it should send an event message to all connected mods so that we can put something in chat locally.

woogerboy21:
Fix #668 

Added server private message functionality to allow server based PM notifications.
Ban based notification to other online moderators is the first form implemented by this PR.

* The server now includes a public function to return a QList of online mod's/admins
* The servers socket interface has been extended to now allow PM's by the server (username Servatrice)
* The registration system restricts the registration of the Servatrice / servatrice username.
* When a private message is sent by the server, the response box is disabled to prevent user responding.

image

### DIFF
--- a/cockatrice/src/chatview.cpp
+++ b/cockatrice/src/chatview.cpp
@@ -130,7 +130,7 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
     lastSender = sender;
     
     // timestamp
-    if (showTimestamps && !sameSender && !sender.isEmpty()) {
+    if (showTimestamps && (!sameSender || sender.toLower() == "servatrice") && !sender.isEmpty()) {
         QTextCharFormat timeFormat;
         timeFormat.setForeground(QColor(SERVER_MESSAGE_COLOR));
         if (sender.isEmpty())
@@ -139,31 +139,35 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
         cursor.insertText(QDateTime::currentDateTime().toString("[hh:mm:ss] "));
     }
 
-    // nickname    
-    QTextCharFormat senderFormat;
-    if (tabSupervisor && tabSupervisor->getUserInfo() && (sender == QString::fromStdString(tabSupervisor->getUserInfo()->name()))) {
-        senderFormat.setForeground(QBrush(getCustomMentionColor()));
-        senderFormat.setFontWeight(QFont::Bold);
-    } else {
-        senderFormat.setForeground(QBrush(OTHER_USER_COLOR));
-        if (playerBold)
+    // nickname
+    if (sender.toLower() != "servatrice") {
+        QTextCharFormat senderFormat;
+        if (tabSupervisor && tabSupervisor->getUserInfo() &&
+            (sender == QString::fromStdString(tabSupervisor->getUserInfo()->name()))) {
+            senderFormat.setForeground(QBrush(getCustomMentionColor()));
             senderFormat.setFontWeight(QFont::Bold);
-    }
-    senderFormat.setAnchor(true);
-    senderFormat.setAnchorHref("user://" + QString::number(userLevel) + "_" + sender);
-    if (sameSender) {
-        cursor.insertText("    ");
-    } else {
-        if (!sender.isEmpty() && tabSupervisor->getUserListsTab()) {
-            const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
-            QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
-            cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel, buddyList.contains(sender)).toImage());
-            cursor.insertText(" ");
+        } else {
+            senderFormat.setForeground(QBrush(OTHER_USER_COLOR));
+            if (playerBold)
+                senderFormat.setFontWeight(QFont::Bold);
         }
-        cursor.setCharFormat(senderFormat);
-        if (!sender.isEmpty())
-            sender.append(": ");
-        cursor.insertText(sender);
+        senderFormat.setAnchor(true);
+        senderFormat.setAnchorHref("user://" + QString::number(userLevel) + "_" + sender);
+        if (sameSender) {
+            cursor.insertText("    ");
+        } else {
+            if (!sender.isEmpty() && tabSupervisor->getUserListsTab()) {
+                const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
+                QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
+                cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel,
+                                                                            buddyList.contains(sender)).toImage());
+                cursor.insertText(" ");
+            }
+            cursor.setCharFormat(senderFormat);
+            if (!sender.isEmpty())
+                sender.append(": ");
+            cursor.insertText(sender);
+        }
     }
 
     // use different color for server messages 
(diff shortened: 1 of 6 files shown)


servatrice/src/serversocketinterface.cpp
```diff
@@ -831,6 +846,19 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
     servatrice->clientsLock.unlock();
 
+    QList<QString> moderatorList = server->getOnlineModeratorList();
+    QListIterator<QString> modIterator(moderatorList);
+    foreach(QString moderator, moderatorList) {
+        QString notificationMessage;
+        if (!userName.isEmpty())
+            notificationMessage.append("\nA ban has been put in place for the user: " + userName);
+        if (!address.isEmpty())
+            notificationMessage.append("\nA ban has been put in place for the address: " + address);
```
ZeldaZach: "ip address"

cockatrice/src/chatview.cpp
```diff
@@ -130,7 +130,7 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType,
     lastSender = sender;
     
     // timestamp
-    if (showTimestamps && !sameSender && !sender.isEmpty()) {
+    if (showTimestamps && (!sameSender || sender == "Servatrice") && !sender.isEmpty()) {
```
ZeldaZach: Just in case, should we `sender.toLower()`?

ZeldaZach:
This looks good to me :shipit:

cockatrice/src/chatview.cpp
woogerboy21: Its up to you.  I don't think its lowered else were and since the server sends the message through  a function that pre-populates that field it should be consistent.

ZeldaZach: We probably could add a global variable `SERVER_ACCOUNT_NAME`

servatrice/src/serversocketinterface.cpp
```diff
@@ -863,6 +895,9 @@ Response::ResponseCode ServerSocketInterface::cmdRegisterAccount(const Command_R
         return Response::RespUsernameInvalid;
     }
 
+    if (userName == "Servatrice" || userName == "servatrice")
```
ZeldaZach: This should probably be swapped for
`if (userName.toLower() == "servatrice")` as there are multiple ways to write it out...

servatrice/src/serversocketinterface.cpp
```diff
@@ -831,6 +846,23 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
+    QList<QString> moderatorList = server->getOnlineModeratorList();
+    QListIterator<QString> modIterator(moderatorList);
+    foreach(QString moderator, moderatorList) {
+        QString notificationMessage = "A ban has been put in with the following details:";
+        if (!userName.isEmpty())
+            notificationMessage.append("\n    Username: " + userName);
+        if (!address.isEmpty())
+            notificationMessage.append("\n    IP Address: " + address);
+        if (!clientID.isEmpty())
+            notificationMessage.append("\n    ClientID: " + clientID);
```
ZeldaZach: maybe a space in Client ID?

woogerboy21: Nah, its shown as no space on the ban dialog.

ZeldaZach: I mean, we should add one :stuck_out_tongue:

tooomm:
>the response box is disabled to prevent user responding

we talked about that in https://github.com/Cockatrice/Cockatrice/pull/1474#issuecomment-138560389 as well.

@woogerboy21 
is it possible to reuse the code?
it would improve the user interface and experience if they just can't chat if they are offline and get visual support.

servatrice/src/serversocketinterface.cpp
```diff
@@ -863,6 +895,9 @@ Response::ResponseCode ServerSocketInterface::cmdRegisterAccount(const Command_R
         return Response::RespUsernameInvalid;
     }
 
+    if (userName.toLower() == "servatrice")
+        return Response::RespUsernameInvalid;
```
Daenyth: Is whitespace trimmed at this point? Could my username be `"servatrice "`?

servatrice/src/serversocketinterface.cpp
```diff
@@ -831,6 +846,23 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
     }
     servatrice->clientsLock.unlock();
 
+    QList<QString> moderatorList = server->getOnlineModeratorList();
+    QListIterator<QString> modIterator(moderatorList);
+    foreach(QString moderator, moderatorList) {
+        QString notificationMessage = "A ban has been put in with the following details:";
```
Daenyth: "A ban has been added:" IMO

servatrice/src/serversocketinterface.cpp
```diff
@@ -520,6 +521,20 @@ void ServerSocketInterface::deckDelDirHelper(int basePathId)
     sqlInterface->execSqlQuery(query);
 }
 
+void ServerSocketInterface::sendServerMessage(const QString userName, const QString message)
+{
+    ServerSocketInterface *user = static_cast<ServerSocketInterface *>(server->getUsers().value(userName));
+    if (user) {
```
Daenyth: Personally, (and this is minor code style), I find it more readable to do along the lines of

```
if (!something)
    return;
lotsOfStuff()
```
instead of
```
if (something) {
    lotsOfStuff()
}
```

common/server.cpp
```diff
@@ -262,6 +262,18 @@ void Server::removeClient(Server_ProtocolHandler *client)
     qDebug() << "Server::removeClient: removed" << (void *) client << ";" << clients.size() << "clients; " << users.size() << "users left";
 }
 
+QList<QString> Server::getOnlineModeratorList()
+{
+    QList<QString> results;
+    QReadLocker clientsLocker(&clientsLock);
+    for (int i = 0; i < clients.size(); ++i) {
+        ServerInfo_User *data = clients[i]->getUserInfo();
+        if (data->user_level() & ServerInfo_User::IsModerator || data->user_level() & ServerInfo_User::IsAdmin) //TODO: this line should be updated in the event there is any type of new user level created
```
Daenyth: We should (later) centralize all the permissions checking.

cockatrice/src/tab_message.cpp
```diff
@@ -116,6 +116,8 @@ void TabMessage::processUserMessageEvent(const Event_UserMessage &event)
         soundEngine->playSound("private_message");
     if (settingsCache->getShowMessagePopup() && shouldShowSystemPopup(event))
         showSystemPopup(event);
+    if (QString::fromStdString(event.sender_name()).simplified() == "Servatrice")
```
Daenyth: Elsewhere you compare lowercase, here not. Any reason?

Daenyth:
Overall :+1: 

(though I'd appreciate waiting a bit before merging so more than one person gets a look at it) :P

ZeldaZach:
We wanted to get it in for weekly builds and it didn't contain anything back breaking so I went forward

servatrice/src/serversocketinterface.cpp
ZeldaZach: Spaces aren't allowed in usernames by default, but I do agree it could be trimmed here.

servatrice/src/serversocketinterface.cpp
ZeldaZach: I'll change it shortly

common/server.cpp
ZeldaZach: +1

cockatrice/src/tab_message.cpp
ZeldaZach: missed case, thanks for pointing it out

Daenyth: The only other thing I might do is to make "servatrice" name a
constant shared across the app. Maybe even configurable so woods could
set it to "WoogerWorks Server" and poixen could do his own, etc.

I can imagine users being "Who is this servatrice user messaging me?"

PR #1544, opened 2015-09-21
Dockified the game view & replay review areas

woogerboy21:
This allows the content under the game tab to be docked / undocked from its tab and moved around.  Since the replay's also use the same game view they are dockable/undockable as well. 

This will allow for:
* Larger game play area's
* Multi monitor use for multiple game play
* Minor Customization layouts for games

### DIFF
--- a/cockatrice/src/shortcutssettings.cpp
+++ b/cockatrice/src/shortcutssettings.cpp
@@ -243,4 +243,5 @@ void ShortcutsSettings::fillDefaultShorcuts()
     defaultShortCuts["Player/phase9"] = parseSequenceString("F9");
     defaultShortCuts["tab_room/aClearChat"] = parseSequenceString("F12");
     defaultShortCuts["DlgLoadDeckFromClipboard/refreshButton"] = parseSequenceString("F5");
+    defaultShortCuts["Player/aResetLayout"] = parseSequenceString("");
 }
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -35,6 +35,7 @@
 #include "pictureloader.h"
 #include "replay_timeline_widget.h"
 #include "lineeditcompleter.h"
+#include "window_sets.h"
 
 #include <google/protobuf/descriptor.h>
 #include "pending_command.h"
@@ -66,6 +67,13 @@
 #include "pb/context_connection_state_changed.pb.h"
 #include "pb/context_ping_changed.pb.h"
 #include "get_pb_extension.h"
+
+#include <QtGui>
+#if QT_VERSION >= 0x050000
+#include <QtWidgets>
+
+
+#endif
 
 ToggleButton::ToggleButton(QWidget *parent)
     : QPushButton(parent), state(false)
@@ -202,6 +210,9 @@ void TabGame::refreshShortcuts()
     if (aCloseReplay) {
         aCloseReplay->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aCloseReplay"));
     }
+    if (aResetLayout) {
+        aResetLayout->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aResetLayout"));
+    }
 }
 
 void DeckViewContainer::loadLocalDeck()
@@ -310,11 +321,12 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor, GameReplay *_replay)
     sayLabel(0),
     sayEdit(0)
 {
+    MainWindow = new QMainWindow;
     setAttribute(Qt::WA_DeleteOnClose);
-    
+
     gameInfo.CopyFrom(replay->game_info());
     gameInfo.set_spectators_omniscient(true);
-    
+
     // Create list: event number -> time [ms]
     // Distribute simultaneous events evenly across 1 second.
     unsigned int lastEventTimestamp = 0;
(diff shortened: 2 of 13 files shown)


Daenyth:
Love the idea. Will dig into details soon

woogerboy21:
@ctrlaltca Can you try this one now on a mac and let me know if it crashes when you close a game/replay tab?  You pointed out that the log tab crashes a mac on disconnect.

Fizztastic:
I'd separate the PlayerWidget from the message log so you can move/close it individually.

woogerboy21:
@Fizztastic What OS are you using?

Fizztastic:
Win7 x64

woogerboy21:
I should probably disable the ability for the game area to be closed (don't you think)?

tooomm:
@Fizztastic :+1: 
If they are separated, you could group them in tabs "behind" each other... that would be awesome! I use that in the deck editor as well.
Saves space and you don't loose visibility when you want to check that deck hash or check the player count in the draft room...
grouped tabs

Is there a system to easily enable/disable single areas easily? Or that gives you feedback what is disabled right now? Like the buttons we have in the deck editor?

woogerboy21:
@tooomm There is a reset layout option under the games menu that will re-enable any docks that have been closed.  Is that sufficient or would everyone prefer individual options to show closed docks?  Also, should I disable any of the other docks abilities to close all together?

tooomm:
@woogerboy21 I like the individual options we have in the deck editor. But maybe it's not needed in the game view tab. I need to see and use it in the client first.

<br>
>Also, should I disable any of the other docks abilities to close all together?

Disable it for the `Message Log` maybe?

woogerboy21:
Here was my thoughts about things.  Currently there is no real way to customize the view in the game tab with out the PR.  I was trying to keep things as similar to the original functionality but add the ability to pull the various portions of the view out to allow for larger game area's or separate monitor support but keep things simple to begin with.  After that lets get some feed back from users that begin using it on the more advance tweaking like splitting sections up.

With that concept in mind maybe I should disable the ability to close any dock and remove any additional menu options that exist currently in this PR and we build from there?

ZeldaZach:
Some comments:
* When you join a game, the layout is always reset. We should store preferences of the layout.
* If you detach an element, you should be able to resize it vertically and/or horizontally, and the layout should adjust as required
* Users in the room and chat should be two separate elements

woogerboy21:
>If you detach an element, you should be able to resize it vertically and/or horizontally, and the layout should adjust as required

This does happen unless im mis-understanding

Fizztastic:
> If you detach an element, you should be able to resize it vertically and/or horizontally, and the layout should adjust as required

In the case of widget with multiple elements (the player widget, timer and message box), you need to be able to resize the elements individually because otherwise they will always fill the same portion (about half) of the widget window.

ZeldaZach:
screenshot 2015-09-22 15 52 38

When I detach the element, I'd like to be able to resize it via the corners, similar to any other application on a computer.

woogerboy21:
Alright, I have split the user layout from the message log (as requested).
I also removed the ability to close the dock's in the game tab but they can be closed in the replay tab.

What else was there to do?

cockatrice/src/shortcutssettings.cpp
```diff
@@ -243,4 +243,5 @@ void ShortcutsSettings::fillDefaultShorcuts()
     defaultShortCuts["Player/phase9"] = parseSequenceString("F9");
     defaultShortCuts["tab_room/aClearChat"] = parseSequenceString("F12");
     defaultShortCuts["DlgLoadDeckFromClipboard/refreshButton"] = parseSequenceString("F5");
+    defaultShortCuts["Player/aResetLayout"] = parseSequenceString("");
```
Daenyth: For clarity it may be good to define something like `const auto noShortcut = parseSequenceString("");` and then use that consistently in this listing, but it's not required.

cockatrice/src/tab_game.cpp
```diff
@@ -201,6 +207,9 @@ void TabGame::refreshShortcuts()
     if (aCloseReplay) {
         aCloseReplay->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aCloseReplay"));
     }
+    if (aResetLayout) {
+        aResetLayout->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aResetLayout"));
```
Daenyth: (on another pr) We should have a dedicated shortcut manager object.

cockatrice/src/tab_game.cpp
```diff
@@ -309,6 +318,7 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor, GameReplay *_replay)
     sayLabel(0),
     sayEdit(0)
 {
+    MainWindow = new QMainWindow;
```
Daenyth: Won't this leak? Does it become owned and destroyed by something else?

cockatrice/src/tab_game.cpp
```diff
@@ -376,52 +368,94 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor, GameReplay *_replay)
...
+    cardReplayBoxLayoutWidget->setLayout(cardVInfoLayout);
+    // end widget two
+
+    // begin widget three
+    playerListWidget = new PlayerListWidget(0, 0, this);
+    playerListWidget->setFocusPolicy(Qt::NoFocus);
+    messageLog = new MessageLogWidget(tabSupervisor, this);
+    // end widget two
+
+    // begin widget four
```
Daenyth: Does it make sense to split these blocks into sub-methods? One per dockified area?

Daenyth:
cc @poixen and @marcofernandezheras

cockatrice/src/tab_game.cpp
woogerboy21: I believe the life of the object only lasts the time in which the tab is created and will be destroyed once the tab is removed.  Correct me if I am wrong, but shouldn't on destruct of the tab the object be removed?

Daenyth: I don't know enough about the tools we're using to say.

@ctrlaltca is it possible to have valgrind target this line to check for leaks?

ctrlaltca: MainWindow is added to a layout in line 449, and then the layout is applied to TabGame.
As stated on http://doc.qt.io/qt-4.8/layout.html#tips-for-using-layouts , the layout will take care of reparenting all the widgets to TabGame.

woogerboy21: So is that a no, it wont cause a leak?

ctrlaltca: Yes; since MainWindow is a children object of TabGame, when TabGame will be deleted it will take care of deleting MainWindow, too; source: http://doc.qt.io/qt-4.8/objecttrees.html

cockatrice/src/tab_game.cpp
woogerboy21: I thought about this, but the reason I did not go that route is because the docks for the replay compared to the docks for the actual game play tab have different options allowing the docks to be closed (or not).  I could still go ahead and do it but I dont see were the code would end up being re-used.  At least not at this time.

It's up to you though, I can if you want me to.

Daenyth: It's fine if the reusability doesn't change right now - the readability will get better. And often I find that by factoring smaller I find more areas to reuse when I don't see them initially.

ZeldaZach:
Game tab doesn't have a label: screenshot 2015-10-01 14 23 32

Sidebar keeps becoming huge on joining games (layout should be saved in preferences, like the deck editor): 
screenshot 2015-10-01 14 26 08

Those are the two big ones right now

woogerboy21:
Thats odd, I have never had the side become "Huge".

woogerboy21:
So I cleaned up some of the code around were @Daenyth asked.  But I am stumped as to how to fix what @ZeldaZach wants.  I'll need some one with better UI knowledge to help me out on that.  Not sure how you would get widget sizes adjusted :cry:

ZeldaZach:
@woogerboy21 I suggest you look at the `settings/deckLayout.ini` file to get a better understanding of how to save/implement layout sizes. I'll see if there's anything within the deck layout code that you might not have for the game tab.

woogerboy21:
Ive looked at that code for about an hour now.

woogerboy21:
@ZeldaZach How's that?  (May have got it.)

ZeldaZach:
Seems like you took the main game play area and set hard minimums. The issue is I can't expand the other fields as far. If you did set minimums, then I suggest you lower the minimums a bit to see what happens.

Also, preferences still aren't saved.

woogerboy21:
As far as I can tell, this is the only way to control widgets starting size.
And I explained earlier why the settings are not saved.  At this point some one with more UI talent will need to step in and finish this up.  For now I am done with it.

ZeldaZach:
I believe #1753 takes over for what this set up. Thanks for the word Woods!

PR #1555, opened 2015-09-22
Fix crash on loading invalid cards.xml

### LINKED ISSUE #1553: Crash when loading cards.xml with outdated version number
https://drive.google.com/file/d/0B8XgYkguCQ0nczlIOGJzQ1otXzA/view?usp=sharing

is the custom set file, that cockatrice is not working with. I had a topic on reddit and this is the follow up.

Cockatrice crashes (I do get an error I think). It as not completely seeing the image method for me (this is a set attempted to be ported from MSE2.

ctrlaltca:
Additionally, move the custom database load after the LoadStatus check, so that they don't bring up the “fix your card database path” settings dialog.
Fix #1553

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -147,6 +147,7 @@ int main(int argc, char *argv[])
     rng = new RNG_SFMT;
     settingsCache = new SettingsCache;
     themeManager = new ThemeManager;
+    soundEngine = new SoundEngine;
     db = new CardDatabase;
 
     qtTranslator = new QTranslator;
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Good to go :shipit:

cockatrice/src/main.cpp
```diff
@@ -197,12 +191,18 @@ int main(int argc, char *argv[])
...
+        db->loadCustomCardDatabases(dataDir + "/customsets");
+    } else {
+        qDebug() << "Could not create " + dataDir + "/customsets folder.";
+    }
+
     if (settingsValid()) {
         qDebug("main(): starting main program");
 
-        soundEngine = new SoundEngine;
-        qDebug("main(): SoundEngine constructor finished");
```
Daenyth: We lost this debug message FWIW here.

ctrlaltca: Yep, i removed it since it was redundant with the messages introduced in a previous PR:
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/soundengine.cpp#L155
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/soundengine.cpp#L46

PR #1556, opened 2015-09-22
Reimplemented PictureLoader as a singleton

ctrlaltca:
This PR moves all the card picture loading code out of carddatabase.cpp and into a new pictureloader.cpp file. This should make maintenance easier.
Also, some hacks have been removed/reworked:
 * the card database was creating an empty "noCard" card item to cache the card back picture: this could lead to an empty card being saved in the token database. Hack removed.
 * removed Hq images handling; it was already disabled previously, and half hacked into being used for “custom” urls. Instead, give users the possibility to define both custom urls, primary (cards with muid) and fallback (cards without muid).

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -1,0 +1,450 @@
+#include "pictureloader.h"
+#include "carddatabase.h"
+#include "main.h"
+#include "settingscache.h"
+
+#include <QApplication>
+#include <QCryptographicHash>
+#include <QDebug>
+#include <QDir>
+#include <QFile>
+#include <QImageReader>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QNetworkRequest>
+#include <QPainter>
+#include <QPixmapCache>
+#include <QSet>
+#include <QSvgRenderer>
+#include <QThread>
+#include <QUrl>
+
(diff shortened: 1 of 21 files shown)


cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,450 @@
...
+            if(b->getEnabled())
+            {
+                // only b enabled
+                return false;
+            } else {
+                // both disabled: sort by key
+                return a->getSortKey() < b->getSortKey();
+            }
+        }
+    }
```
Daenyth: Ugh I wish we had good pattern matching.

ZeldaZach:
I'm under the assumption this is fine as it passed travis. I'll wait a bit still

Daenyth:
@ZeldaZach We don't have automated tests, you can't say that it's fine just because it compiles.

ctrlaltca:
Agreed. From a quick view it seems that i missed some point in the code that's happily dereferencing a pointer without checking if it's valid first. The removal of the "noCard" hack exposes these problems.

Daenyth:
@ctrlaltca you can make a new branch from current master, then `git revert  d55e44e` on it to keep testing

ctrlaltca:
> @ctrlaltca you can make a new branch from current master, then git revert d55e44e on it to keep testing

```git revert -m1 d55e44e```

Daenyth:
Probably

PR #1569, opened 2015-09-23
Updated cockatrice_warnings table to use id column as primary key

woogerboy21:
Per @Daenyth request, the cockatrice_warnings table has the primary keys set to user_name,time_of.  With this PR the schema is updated to use the user's id column and time_of.

### DIFF
--- a/servatrice/servatrice.sql
+++ b/servatrice/servatrice.sql
@@ -20,7 +20,7 @@ CREATE TABLE IF NOT EXISTS `cockatrice_schema_version` (
   PRIMARY KEY  (`version`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
 
-INSERT INTO cockatrice_schema_version VALUES(10);
+INSERT INTO cockatrice_schema_version VALUES(11);
 
 CREATE TABLE IF NOT EXISTS `cockatrice_decklist_files` (
   `id` int(7) unsigned zerofill NOT NULL auto_increment,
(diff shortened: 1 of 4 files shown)


Daenyth:
Doesn't this need a code update?

woogerboy21:
@Daenyth 
What do you mean a code update?  The PR added the id field, it just never set it as the primary key.

Daenyth:
The `id` field would the id of the warning - the schema doesn't include a `user_id` field at all.

woogerboy21:
Ah I see were the confusion is coming in.  The id field of the table is the id of the user.  There is no id field generated for each warning message.

https://github.com/Cockatrice/Cockatrice/blob/master/servatrice/src/servatrice_database_interface.cpp#L962

Are you wanting that changed?  The original request in the PR you stated you wanted the user id used rather than name so the id column was added that stores the user id.

Daenyth:
Yes, let's rename it. `id` as a column should only ever be an autoinc value specific to the table it's in. References to other tables should be `<table>_id`

woogerboy21:
@Daenyth Hows that?

servatrice/servatrice.sql
```diff
@@ -139,13 +139,13 @@ CREATE TABLE IF NOT EXISTS `cockatrice_bans` (
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
 
 CREATE TABLE IF NOT EXISTS `cockatrice_warnings` (
-  `id` int(7) unsigned NOT NULL,
+  `use_id` int(7) unsigned NOT NULL,
```
Daenyth: Typo

woogerboy21:
@Daenyth : Typo fixed.  Hows about this time? LoL

Daenyth:
:shipit:

woogerboy21:
Man when my blood sugar is high i make TONS of errors :laughing:

PR #1027, opened 2015-04-26
Enable hidpi (retina) support under osx

ctrlaltca:
* add to the application bundle a custom info.plist
 * fix a graphic glitch on the life counter

Fix #799

PR #1584, opened 2015-09-25
Enable Hidpi support

leematos:
Hey All,

I've made a tiny change that will allow HiDPI SVGs for _most_ things in the cockatrice interface. Tested on a Mac that built Cockatrice with QT5. It needs to be tested with QT4 and on other platforms to make sure it doesn't break things. I'll be digging in to figure out why all SVGs aren't automatically smoothed but this is a start and really helps. (Even the MenuBar Icon is HiDPI!!!)


woogerboy21:
Testing this out on a linux box with QT 5.4 and a standard video adapter running 1920x1080 seems to look fine.  I didnt test a whole lot out , just hit some cards up and jumped on the main server to watch some chat going on for a bit.

tooomm:
is it possible/necessary to enable this in oracle as well?
not sure if we use any svg's there right now, but we might in the future!

#1027 still needed with this?

Daenyth:
Can you squash commits down to one?

leematos:
@Daenyth yup, Totally doable.

@tooomm Yeah, both are needed. The Other fix I proposed will treat the QApplication as "HiDPI" and get smoother text (in most places), but it doesn't affect QPixmaps and some other things (I Think QLabel as well, testing now.). The other fix is also Mac Specific. This fix is to get QPixmaps HiDPI, which in theory, should work to make images better on all HiDPI platforms. That said, It will need to be tested on non-HiDPI platforms to make sure there are no regressions.

ZeldaZach:
I don't see any changes coming about from this PR, unless I'm looking in the wrong places.
screenshot 2015-09-28 20 58 01

Daenyth:
I'll test this when I get a chance (osx with retina screen)

Daenyth:
I see the same as Zelda.

I suspect it's an issue with the user details pane and not the PR. 

It looks fine, and in any case it shouldn't hurt anything.

Please squash to one commit and then I can merge

tooomm:
@leematos Can you please squash your commits down to one to let us finally merge your enhancement? ;)

If you need help, let us know.

Daenyth:
If I have time before he does it, I'll squash them myself and merge

ZeldaZach:
Any news on this?

Daenyth:
It seems I haven't had much time lately...

Daenyth:
See #1728

Issue #1562, opened 2015-09-23
Proposal: PRs need two signoffs before merge


Daenyth:
I propose that new PRs need :shipit: from two people other than the author before merge, given after the most recent commit. If the PR is security related or a fix for a severe crash, then it doesn't require any signoff and is instead merged at the author or first reviewer's discretion.

If the author has merge rights, they should merge it themselves.

If the author doesn't have merge rights, anyone on the project should feel free to merge it.

Person doing the merge should update the changelog. (For now; later I'd like to get changelogs auto-generated for the most part).

@Cockatrice/developers please :+1: or :-1: and leave comments.

poixen:
Love this idea. Though it can sometimes slow down prs as not everyone checks the prs. Pros and cons.

ctrlaltca:
I agree with this; i'd just province an exception for security PRs and hard crashes, especially if they're trivial patches (like oneliners).

Daenyth:
> I agree with this; i'd just province an exception for security PRs and hard crashes, especially if they're trivial patches (like oneliners).

Good point, I'll say that security and crash fixes get merged at author/reviewer's discretion.

woogerboy21:
Ill throw my $0.02 in here and mention that something like this sounds great, but with volunteered time things could potentially stay out for a very long time.  But even with that Im willing to try it with the understanding that it can always be changed in the future.  Obviously there is something that needs changed and with out finding out what it would be like any other way we will never know what way works best.

ZeldaZach:
I'm fine with trying this out. The only real issue is when things sit open for long period of time (4+ days) without any interactions

Daenyth:
Given the nature of recent support requests and general community response, I'm also going to add that new client/server features (as in ones listed in `featureset.cpp`) must be tested for at least 2 weeks prior to merge and public release.

Let's go with this strategy for now.

PR #1589, opened 2015-09-26
Added appveyor script

### LINKED ISSUE #1218: Set up Windows automated builds on Appveyor
Like travis but for windows

ctrlaltca:
It will test msvc builds and create windows packages for both 32 and 64 bits. Fix #1218.
If you're compiling windows packages (@woogerboy21), you could need to review these changes:
 * Forced /MD flag to the visual studio compiler (fixes linking)
 * removed the hardcoded "win32" text in the package name

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,78 @@
+version: 0.0.0.0.1-branch-{branch}-build-{build}
+cache:
+    - c:\protobuf
+    - c:\protoc
+    - c:\zlib
+environment:
+    matrix:
+        - vc_arch: amd64
+          choco_arch:
+          nuget_arch: x64
+          target_arch: x86_64
+          qt_ver: 5.5\msvc2013_64
+        - vc_arch: amd64_x86  # cross-compile from amd64 to x86
+          choco_arch: --x86
+          nuget_arch: Win32
+          target_arch: x86
+          qt_ver: 5.5\msvc2013
+install:
+    - systeminfo
+    # upgrade cmake in order to have c++11 support
+    - choco install cmake -y
+    - choco install nsis -y
+    - choco install jom -y
+    - ps: |
+        if (Test-Path c:\protoc) {
+            echo "using protoc from cache"
+        } else {
+            Invoke-WebRequest "https://github.com/google/protobuf/releases/download/v2.6.1/protoc-2.6.1-win32.zip" -OutFile c:\protoc-2.6.1-win32.zip
+            c:\cygwin\bin\bash -lc "cd /cygdrive/c; 7z x -y protoc-2.6.1-win32.zip -oc:\protoc"
+        }
+    - ps: |
+        if (Test-Path c:\protobuf) {
+            echo "using protobuf from cache"
+        } else {
+           nuget install protobuf-v120 -OutputDirectory c:\protobuf
+        }
+    - ps: |
+        if (Test-Path c:\zlib) {
+            echo "using zlib from cache"
+        } else {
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -0,0 +1,78 @@
...
+    - c:\protobuf
+    - c:\protoc
+    - c:\zlib
+environment:
+    matrix:
+        - vc_arch: amd64
+          choco_arch:
+          nuget_arch: x64
+          target_arch: x86_64
+          qt_ver: 5.5\msvc2013_64
```
Daenyth: Don't we want 5.4?

Is it possible for us to also test Qt4?

ctrlaltca: For windows 5.5 is fine; 5.5 is problematic only for osx (rpath enabled by default, problems in detecting clang features).
Appveyor currently only supports qt 5.3, 5.4 and 5.5. We could download and install qt4 manually, but that would take a lot of time.

woogerboy21: @Daenyth I've been using 5.5 for windows clients for some time now

woogerboy21:
@ctrlaltca Are you saving the builds off any were?

ctrlaltca:
Not yet; it's planned as step 2.

woogerboy21:
Is there a reason you dropped the win32 as well?  To me its pretty obvious which binary goes to what what OS but one thing you may want to add is the QT version some were in the binary name.  Otherwise you will end up with name collisions when creating both Qt4 & Qt5 packages.

ctrlaltca:
I dropped the hardcoded win32 because that would end up also in the name of x86_64 packages.
Adding the qt version is a nice idea.

woogerboy21:
I see you pull the proto files if dont exist :wink:

ctrlaltca:
last win32 package: https://ci.appveyor.com/project/ctrlaltca/cockatrice/build/job/jvbbvghk75xo3ibn/artifacts
last win64 package: 
https://ci.appveyor.com/project/ctrlaltca/cockatrice/build/job/ibkgcbssfmbgxmla/artifacts

woogerboy21:
Those are only Qt5 builds though correct?

woogerboy21:
@ctrlaltca Just curious, is this the script that will be used to test each PR put in?  And if so are we really wanting a binary saved each time?

ctrlaltca:
> @ctrlaltca Just curious, is this the script that will be used to test each PR put in? And if so are we really wanting a binary saved each time?

No, i just started working on it but forgot to do it on another branch; fixed

woogerboy21:
Is this safe to remove the PS content near the end that manipulates the final binary and use for testing windows compiling results?  I for one would be interested in seeing this implemented as a first go around to allow a better testing for windows compiling and later add the binary file functionality.

ctrlaltca:
The last PS line is just renaming the resulting exe, adding the arch it has been built for, and telling appveyor what files to conserve before trashing the worker vm. IMHO this PR is quite done.

woogerboy21:
:+1:   Wasnt sure if you had wanted to add more to the PS section.  But by your statement it does not appear to be the case for now.

woogerboy21:
Oh one other things, is there some were I could glance at the output of the run for a build in appveyor?  The links you provided earlier dont seem to show anything any longer.

ZeldaZach:
I'm assuming this is good to go then

woogerboy21:
Looks ok to me.

ctrlaltca:
As per #1562, having received two signoffs, i'm gonna merge this.

PR #1603, opened 2015-09-28
Fix volume control

### LINKED ISSUE #1528: Volume issues on Windows 7
Using the latest weekly (from ead6ab3), the sound is behaving strangely. For values of 0 and 1, the volume is entirely off &mdash; no sound can be heard. For values between 2 and 100, the volume is at what seems to be maximum.

ctrlaltca:
fix #1528

### DIFF
--- a/cockatrice/src/soundengine.cpp
+++ b/cockatrice/src/soundengine.cpp
@@ -89,7 +89,7 @@ void SoundEngine::playSound(QString fileName)
     inputBuffer->setData(audioData[fileName]);
     inputBuffer->open(QIODevice::ReadOnly);
 #if QT_VERSION >= 0x050000
-    player->setVolume(settingsCache->getMasterVolume());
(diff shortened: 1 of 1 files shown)


ZeldaZach:
The issue still seems to be present on OS X 10.10.5 for me

ZeldaZach:
This issue seems to be fixed on OS X 10.11 (El Capitan) so it was probably an issue in my build environment. :shipit:

Daenyth:
Can we make a packaged build for the folks in #1528 to try?

ctrlaltca:
I'm quite confident on this one.. it just fixes a miscalculation introduced here: https://github.com/Cockatrice/Cockatrice/pull/1498/files#diff-51b32c35f453129126156f5d53a775d5L72

Daenyth:
:shipit:

PR #1604, opened 2015-09-28
Re-Implemented the moderator notification upon user warning/ban

woogerboy21:
This PR re-introduces the notification that would happen when a user is given a warning or a ban.  Previously the user list was not correctly locked resulting in a server crash if the mod lookup function was called at the time a user logged in and a lock on the list was created.

This should go through some pretty good testing as it was known to cause crashes before.


ZeldaZach:
Are we looking into this still?

tooomm:
Initial PR was #1527, while #1593 removed the crashing part... the complete server pm code is still in the app?
Including Response functionality is disabled in private chat window, people are not allowed to register as `Servatrice` etc.?

So all online mods will receive a private message from the server?
Anyone tested this recently?

woogerboy21:
I synced this PR up with master and am in the process of bringing a number of test clients up to test this out and roll it in.  I wanted to have more than a single moderator account on for testing along with more than a single client ban/warn happen to make sure it doesn't crash things.  Hopefully have it testing in a few days.

woogerboy21:
Checked out and doesnt crash things.  Have 1 admin and 2 other moderator accounts logged in and 2 users.  Tested warnings and bans.  Both were sent correctly to all admin/mod's and didnt crash the server like it did during the first revision that was merged.

PR #1608, opened 2015-09-29
Travis changes for package creation

ctrlaltca:
* add a "Release" build
 * create a package for release builds;
 * mute osx's homebrew to avoid it filling travis's logfile and make the build fail
 * make the build matrix smaller, compiling only with gcc on linux and clang under osx


ZeldaZach:
I feel this is good to go as it's just a building script and I was able to get it passed. :shipit:

woogerboy21:
This should fix the failing clang issue's about file size being to big I believe correct?

ZeldaZach:
Yes @woogerboy21

ctrlaltca:
My only doubt is if it's ok for @Daenyth to remove the linux+clang and osx+gcc builds; i did it in order to lower the number of total builds from 12 to 6, speeding up a lot the time needed by travis to do his job.
Of course gcc is still tested under linux and clang is used under osx.

Daenyth:
I'm not sure how I feel about that.

How much time does it save? Especially once we're on the new containers

ctrlaltca:
I'm not that hopeful that we'll migrate on the new containers; it seems that something is not working in travis' current way of whitelisting packages.
Reducing the build matrix from 12 to 6 makes us save about 15 minutes on every build; it's not a real dealbreaker for us, it's a only a choice made to optimize the process trying to keep the number of different builds to a minimum, following an idea of parsimony.

Daenyth:
Looks like my email reply never came through.

I'm OK with reducing the builds

PR #1609, opened 2015-09-30
Implemented return response server feature set for in client functionality disabling

### LINKED ISSUE #1563: "Warn user" action should only be enabled if server supports it in featureset
If server doesn't support it, it should be greyed out in the menu

woogerboy21:
Fix #1563 

This PR implements a response protocol repeated string that is returned to the client that allows the client to compare features that the server has to local client functionality for disabling features that may not be supported by the server being connected to.

I stored the servers feature set that is returned in the client side settings cache as that seemed to be the most easily accessible location through out the client.  Wasn't sure if this was the best spot.


Daenyth:
I'll try to have a look at this tonight.

Storing the info on the server handle rather than the local settings cache may be more appropriate. Tab supervisor may need it passed in

cockatrice/src/remoteclient.cpp
```diff
@@ -150,6 +150,11 @@ void RemoteClient::loginResponse(const Response &response)
         if (resp.missing_features_size() > 0 && settingsCache->getNotifyAboutUpdates())
                 emit notifyUserAboutUpdate();
 
+        if (resp.server_features_size() > 0) {
+            for (int i = 0; i < resp.server_features_size(); ++i)
+                settingsCache->addServerFeatures(QString::fromStdString(resp.server_features(i)));
```
ctrlaltca: I see a problem here: the ```serverFeatures``` variable in settingscache is never cleared. So, if i first connect to a server supporting a specific feature "A", then disconnect and connect to a second server not supporting the feature, it will still be listed in the available server features.
We should either clear the serverFeatures list when disconnecting, or just making the ```serverFeatures``` variable a property of the current active server.

Daenyth: Yeah, I'm not liking this being on `settingsCache` - that's for local configuration regardless of what server you're on

woogerboy21: Keep in mind the cache would be changed based on the server your connecting
to.  It's some what of a grey area in my opinion.  Settings cache is just
that, it's not something that gets permanently written and reused later on
the client side after client unloading.

PR #1627, opened 2015-10-05
Fix appveyor builds

ctrlaltca:
it seems that jom is failing for some obscure reason.
This PR drops jom and use plain nmake.
Builds will take some more time to complete.. but at least they will.


Daenyth:
:shipit:

Issue #1628, opened 2015-10-06
Can't join the Main room! "Failed to join the room due to an unknown error."


VikkuViral:
I just installed, got the C++ install for Win 7, updated the card database, and registered to server: cockatrice.woogerworks.com, with port: 4747. Version d55e44e

Been trying to troubleshoot this and just can't figure it out. Am I just an idiot? I see about 1,000 people in the server so I know it's on my side!

Thanks and cheers!

woogerboy21:
This error has been mentioned before and we have yet to reproduce it. From
what we have seen the error is thrown when your username is already joined
to the room and is most likely caused by the client. It properly
disconnecting.  Let's start troubleshooting by making sure your client is
closed and you reboot your PC. This will ensure the client is completely
disconnected from the server.

Allow at least 15 minutes to pass before attempting to reopen and rejoin
the server and see if you can log in and join the room.

Also, is VikkiViral your username or are you using a different username? I
can see if there is anything on our side I. The logs that would give more
insight.

Thanks.

Daenyth:
So there's 3+ people so far who have reported this, very likely many more.

Anyone affected, please list:
- Version from "about" menu
- Does it happen 100% of the time or only sometimes?
- Any other details you've noticed about the behavior.

TFIN678123:
I have Version d55e44e.  This has happened only after I updated the version, and it has happened 100% of the times I tried to join a room.

Daenyth:
What's your username on the server?

TFIN678123:
TFIN

Daenyth:
Can you try right now? If you're on any version that isn't windows, [collecting debug output](https://github.com/Cockatrice/Cockatrice/wiki/Debugging#mac-osx) would help

TFIN678123:
Just tried and the same error occurred.  Should I try restarting?

I am using the Windows version.

Daenyth:
I don't see you online at all. Can you stay connected and PM me from the Account tab?

ctrlaltca:
I see him both online and inside the Main room now.

Daenyth:
Same

Daenyth:
@ctrlaltca Did anything server side change? I'm wondering if there's some odd condition that might make it not send the "you joined room x" message

TFIN678123:
thanks for the help!

Daenyth:
@VikkuViral When you get this, is the Main tab visible? @TFIN678123's issue was that the game list was empty.

[...]

PR #1637, opened 2015-10-07
Handle more Command_JoinRoom return values

ctrlaltca:
It could bring some light on #1628

### DIFF
--- a/cockatrice/src/tab_server.cpp
+++ b/cockatrice/src/tab_server.cpp
@@ -12,6 +12,7 @@
 #include "tab_server.h"
 #include "abstractclient.h"
 #include "userlist.h"
+#include "tab_supervisor.h"
 #include <QDebug>
 
 #include "pending_command.h"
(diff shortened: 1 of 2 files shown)


cockatrice/src/tab_server.cpp
```diff
@@ -139,8 +139,17 @@ void RoomSelector::joinClicked()
...
     switch (r.response_code()) {
-        case Response::RespOk: break;
-        case Response::RespUserLevelTooLow: QMessageBox::critical(this, tr("Error"), tr("You do not have the proper permission to join this room.")); return;
+        case Response::RespOk:
+            break;
+        case Response::RespNameNotFound:
+            QMessageBox::critical(this, tr("Error"), tr("Failed to join the room: it doesn't exists on the server."));
+            return;
+        case Response::RespContextError:
+            QMessageBox::critical(this, tr("Error"), tr("You are already inside the room."));
```
Daenyth: Instead of this, could we do a bit of validation on the client side? Check if the room tab is being displayed via the supervisor, and if it's not, show the error, if it is, focus it.

We could potentially have some other handling - if the server says we're in the room but we don't have it open locally, log an error, tell the server to leave the room, then once the "left room" response is handled, join the room again.

Daenyth: Now that the logic flow is changed, ignore my previous comment. As far as I can tell the only way this is reached is if the server thinks we are in the room but we are not rendering a room tab - which should not happen.

Assuming my assessment is correct (please tell me if I'm wrong), then I think we should change the message here to something like `The server thinks you are in the room but Cockatrice is unable to display it. Try restarting Cockatrice`

cockatrice/src/tab_server.cpp
```diff
@@ -178,3 +155,47 @@ void TabServer::processServerMessageEvent(const Event_ServerMessage &event)
...
+            return;
+        case Response::RespUserLevelTooLow:
+            QMessageBox::critical(this, tr("Error"), tr("You do not have the proper permission to join this room."));
```
Daenyth: > You do not have the required permission to join this room.

cockatrice/src/tab_server.cpp
```diff
@@ -178,3 +155,47 @@ void TabServer::processServerMessageEvent(const Event_ServerMessage &event)
...
+            return;
+        default:
+            QMessageBox::critical(this, tr("Error"), tr("Failed to join the room due to an unknown error."));
```
Daenyth: Can you display the `response_code` here so we can debug it easier? Maybe also a `qDebug`

Daenyth:
Squash commits and :shipit:

Issue #487, opened 2014-12-16
Expand gender/pronoun support.


StellaBiderman:
I'm playing in English.

It is currently using he/him/his pronouns, and despite changing the gender designation of my account, I can get it to neither use she/her/hers or they/them/their pronouns. I've tried both the "f" and "o" designators for gender on my account page and nothing happens.

JSBiderman:
Any progress?

Daenyth:
I think this actually doesn't impact translation a ton, though that's a nice to have. The `unknown` gender level already existing has the two-gender icon and uses "they" in English, which covers a lot of people.

I'm really uncertain how best to handle more complex cases though - do we want to support things like xe, and how can we direct translators? If we only add a proper name using the existing two-gender icon, what should we call it? Queer? That certainly doesn't cover everyone using "they" for their pronoun. This stuff is extremely complicated and I don't want to get it wrong for people, but at the same time I have very few ideas on getting it right.

Daenyth:
Note to self: https://github.com/Cockatrice/Cockatrice/blob/69e6aa355ff9caa016f066447b2d3aad5978dbb8/servatrice/src/servatrice_database_interface.cpp#L270

Daenyth:
I spoke with a friend of mine and these are their recommendations:

Change male/female to masculine/feminine
Add a top level neutral option, using the two gender icon

And I'll also add a proper unspecified option as the default, using the null gender icon. In English both the unspecified and neutral would be put with "they" pronouns but it would allow translators more information for use in their language.

@JSBiderman what's your opinion on that proposal?

JSBiderman:
That sounds wonderful! Thank you for caring :)

Daenyth:
WIP branch: https://github.com/Cockatrice/Cockatrice/tree/gender-updates

Daenyth:
@ideocl4st FYI expect a huge string update when this is done. I have no idea how korean handles gender neutral pronouns but it would be good to look into if you're uncertain

ideocl4st:
I don't mind huge string updates at all, but I do agree that I'll need some research as well.

Thanks for letting me know about it beforehand.

Daenyth:
It'll be four cases - masculine, feminine, unspecified (use whatever you would when you don't know the gender) and neutral (in english they'd have "they" as a neutral pronoun)

Daenyth:
[...]

PR #1638, opened 2015-10-08
added gender neutral and unspecified gender cases for each string

realityfabric:


Daenyth:
Github won't even show me the diff - I'll have a look locally on my machine when I can. I know this was a ton of work, so thanks up front!

realityfabric:
Should I have broken up the commits into smaller chunks so that GitHub would be able to manage it? I didn't realize that large commits would fail to display on GitHub

Daenyth:
Nope, one commit is correct - github just doesn't like to display diffs that are large in a single file. It wouldn't display them collectively if you split commits either.

Daenyth:
The idea being I think that generally changes like that are, for example, rewrites of svg files - the lines don't matter as much as the result.

realityfabric:
Okay, makes sense. Let me know if I missed anything, I just focused on all the places where there was an isFemale function call

Daenyth:
I think hitting that should cover them all.

tooomm:
linking #487

ZeldaZach:
Any word on the review of this? Travis said it failed, but I can't view the changes easily.

Daenyth:
I haven't had much time for anything this week. Will get to this as soon as
I can. You can view the diff manually locally

ctrlaltca:
The code seems ok; i also checked transifex, and it will show the gender disambiguation comment correctly. I see that the gender is correctly used to get a different translation for all the strings containing a pronoun (his/her/...).

It's also used in other strings where there's no pronoun involved (like line 43: You have joined game %1); Is this needed to support other languages, where the same sentence could have different translations depending on the gender of the user?
Some strings, instead, have no per-gender specific translation (eg. lines 695 and 698). Is this wanted or those lines have just been missed in the commit?

Another question: from a translator-specific point of view, what's the difference between "unspecified gender" and "gender neutral"?

Last, the logCreateArrow() handling is insane.. to bad qt offers no way to handle such cases.

Daenyth:
Unspecified is for when the person hasn't selected anything. Neutral is for when they've selected neutral pronouns (common in the queer community if the person doesn't identify strictly along binary lines)

ctrlaltca:
Thank you, i made some good readings on the queer topic and now understand the meaning of the neutral pronouns.

ctrlaltca:
As @tooomm pointed out, https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/cardzone.cpp#L47 needs to be fixed, too.

Psithief:
If you want to do this right, replace all instances of `female` with `feminine` and `male` with `masculine` as well.
`female`, `male` are sexes, not genders.

(I don't actually care about these issues, I just can't stop the QA)

Daenyth:
Yes, that would be better.

realityfabric:
While changing it to feminine and masculine are fine, and I don't want to get into heavy detail here because it's not really relevant to the project itself, Sex and Gender are not separate things, they are both social constructs, and there is no good reason to distinguish them from each other.

Psithief:
Letting users choose a gender was unnecessary in the first place. :+1:

Daenyth:
Yeah. It was added mostly for translation purposes I think. I wonder about
removing it entirely and just using neutral everywhere.

ZeldaZach:
Can we do anything with this right now?

Daenyth:
Honestly I'm wondering if it's not worth removing the option to set it and just go with neutral pronouns throughout as most software does. I haven't decided. Community feedback would be helpful perhaps.

ZeldaZach:
I'm fine with going to genderless in game. It's irrelevant for the most part (and we can put all translations to masculine)

Daenyth:
I'd leave it to the translators to decide how best to approach it at that point.

Daenyth:
In any case, leaving this open for now won't hurt anything

realityfabric:
It would definitely be easier, in English, to go genderless across the board. It would definitely cut down on the number of switches, and probably cut down on the amount of nested logic in general for messages.

ctrlaltca:
:+1: for genderless anywhere

tooomm:
it seems very reasonable to make things easier here. the benefit might not be worth the effort for such a small project. translations are bloated and difficult to handle for the ordinary player that just want to help a bit, too.

any problems we might run into with translations for other languages?
like special cases in some languages?

but in general, translators should be able to deal with language specialties:
>I'd leave it to the translators to decide how best to approach it at that point.

ZeldaZach:
So what's the plan with this?

Daenyth:
I think we can leave this open until either someone picks it up and updates it, or we decide on the path to removing selection in favor of all-neutral.

Daenyth:
I appreciate the work that went into the PR but at this point I think we've decided to remove the pronoun code entirely to streamline translations - we'll just use exclusively default language from here.

PR #1640, opened 2015-10-09
Fixes link list

theavish:
On line 5, fixes the Community Resources and Translation Status links to the full name of the header, to follow the naming convention present in the rest of the links.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,11 @@
 language: cpp
 env:
-  - QT4=1
-  - QT4=0
+  - QT4=1 BUILDTYPE=Debug
+  - QT4=0 BUILDTYPE=Debug
+  - QT4=0 BUILDTYPE=Release
 os:
   - linux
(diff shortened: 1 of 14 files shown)


Nightfirecat:
Looks like you didn't pull from master before making a PR.

I'd recommend updating your local copy of Cockatrice/Cockatrice, creating a new feature branch from there, and making your change there, rather than doing a master -> master PR.

tooomm:
mtom

Your PR is still from your master, you should create a new branch for your feature/change. Called `readme_toc_upate` for example and commit your changes *to this new branch*.
Then create a PR from that specific branch instead of your master branch. That's what @Nightfirecat mentioned.
Now yours is mixed up with other stuff as well.



The reason for cutting the names down was to not break the line.
With your change it would look like this:
toc

Daenyth:
I don't see the need for this.

PR #1600, opened 2015-09-28
Add dummy googletest

Daenyth:
Uses the built in `cmake -Dtest=ON`

It should download googletest on the fly if needed.

Adds a new make target, `make test`

Ref #167

PR #1641, opened 2015-10-09
CMake: Add "test" target

ctrlaltca:
Adds a new "test" target that can be enabled from cmake ``` -DWITH_TESTS=1```.
CMake downloads a vendored copy of googletest if not found on the system.
Tests can be run using ```make test```.
Also, update qt from 5.2 to 5.4.1 (from the same ppa)
Replaces and fixes #1600.


CMakeLists.txt
```diff
@@ -232,3 +232,10 @@ if(WITH_ORACLE)
     add_subdirectory(oracle)
     SET(CPACK_INSTALL_CMAKE_PROJECTS "release/oracle.app;oracle;ALL;/" ${CPACK_INSTALL_CMAKE_PROJECTS})
 endif()
+
+# Compile tests (default off)
+option(WITH_TESTS "build tests" OFF)
+if(WITH_TESTS)
```
Daenyth: Why use this over the built in `-Dtest=ON`?

ctrlaltca: Afaik no special "test" variable exists in cmake: https://cmake.org/cmake/help/v3.0/manual/cmake-variables.7.html
A special "test" target exists when CTest is enabled, but "-Dtest=ON" is nothing more than a normal #define.
So, tried to make it homogeneous with the other target-enabling definitions  (WITH_ORACLE, WITH_SERVER, ..).

Daenyth: I don't have a strong feeling about it.

ctrlaltca: Reverted to "TEST"

travis-compile.sh
```diff
@@ -9,13 +9,14 @@ if [[ $TRAVIS_OS_NAME == "osx" && $QT4 == 0 ]]; then
   prefix="-DCMAKE_PREFIX_PATH=`echo /usr/local/Cellar/qt5/5.*/`"
 fi
 if [[ $TRAVIS_OS_NAME == "linux" && $QT4 == 0 ]]; then
-  prefix="-DCMAKE_PREFIX_PATH=/opt/qt52/lib/cmake/"
+  prefix="-DCMAKE_PREFIX_PATH=`echo /opt/qt5*/lib/cmake/`"
 fi
-cmake .. -DWITH_SERVER=1 -DCMAKE_BUILD_TYPE=$BUILDTYPE -DWITH_QT4=$QT4 $prefix
 
 if [[ $BUILDTYPE == "Debug" ]]; then
```
Daenyth: :+1: I like the split.

Daenyth:
Can you update the README with the new cmake option and then :shipit:

PR #1642, opened 2015-10-10
Split PictureLoader out of CardDatabase

ctrlaltca:
This PR moves all the card picture loading code out of carddatabase.cpp and into a new pictureloader.cpp file. This should make maintenance easier.
Also, Hq images handling has been removed; it was already disabled previously, and half hacked into being used for “custom” urls. Instead, give users the possibility to define both custom urls, primary (cards with muid) and fallback (cards without muid).

This is basically #1556, minus the bits causing problems.

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -1,0 +1,462 @@
+#include "pictureloader.h"
+#include "carddatabase.h"
+#include "main.h"
+#include "settingscache.h"
+#include "thememanager.h"
+
+#include <QApplication>
+#include <QCryptographicHash>
+#include <QDebug>
+#include <QDir>
+#include <QFile>
+#include <QImageReader>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QNetworkRequest>
+#include <QPainter>
+#include <QPixmapCache>
+#include <QSet>
+#include <QSvgRenderer>
+#include <QThread>
+#include <QUrl>
+
+class PictureToLoad::EnabledAndKeyCompareFunctor {
+public:
+    inline bool operator()(CardSet *a, CardSet *b) const
+    {
+        if(a->getEnabled())
+        {
+            if(b->getEnabled())
+            {
+                // both enabled: sort by key
+                return a->getSortKey() < b->getSortKey();
+            } else {
+                // only a enabled
+                return true;
+            }
+        } else {
+            if(b->getEnabled())
+            {
+                // only b enabled
+                return false;
+            } else {
+                // both disabled: sort by key
+                return a->getSortKey() < b->getSortKey();
+            }
+        }
+    }
+};
+
+PictureToLoad::PictureToLoad(CardInfo *_card)
+    : card(_card), setIndex(0)
+{
+    if (card) {
+        sortedSets = card->getSets();
+        qSort(sortedSets.begin(), sortedSets.end(), EnabledAndKeyCompareFunctor());
+    }
+}
+
+bool PictureToLoad::nextSet()
+{
+    if (setIndex == sortedSets.size() - 1)
+        return false;
+    ++setIndex;
+    return true;
+}
+
+QString PictureToLoad::getSetName() const
+{
+    if (setIndex < sortedSets.size())
+        return sortedSets[setIndex]->getCorrectedShortName();
+    else
+        return QString("");
+}
+
+CardSet *PictureToLoad::getCurrentSet() const
+{
+    if (setIndex < sortedSets.size())
+        return sortedSets[setIndex];
+    else
+        return 0;
+}
+
+QStringList PictureLoader::md5Blacklist = QStringList()
+    << "db0c48db407a907c16ade38de048a441"; // card back returned by gatherer when card is not found
+
+PictureLoader::PictureLoader()
+    : QObject(0),
+      downloadRunning(false), loadQueueRunning(false)
+{
+    picsPath = settingsCache->getPicsPath();
+    picDownload = settingsCache->getPicDownload();
+
+    connect(this, SIGNAL(startLoadQueue()), this, SLOT(processLoadQueue()), Qt::QueuedConnection);
+    connect(settingsCache, SIGNAL(picsPathChanged()), this, SLOT(picsPathChanged()));
+    connect(settingsCache, SIGNAL(picDownloadChanged()), this, SLOT(picDownloadChanged()));
+
+    networkManager = new QNetworkAccessManager(this);
+    connect(networkManager, SIGNAL(finished(QNetworkReply *)), this, SLOT(picDownloadFinished(QNetworkReply *)));
+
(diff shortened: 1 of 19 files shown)


cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+#include <QNetworkReply>
+#include <QNetworkRequest>
+#include <QPainter>
+#include <QPixmapCache>
+#include <QSet>
+#include <QSvgRenderer>
+#include <QThread>
+#include <QUrl>
+
+class PictureToLoad::EnabledAndKeyCompareFunctor {
```
Daenyth: I know it's C++ convention to call this Functor, but can we please not? It's extremely confusing when you're used to using actual functors (from category theory).

`EnabledAndKeyCompareFunction` would be fine by me..

cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+public:
+    inline bool operator()(CardSet *a, CardSet *b) const
```
Daenyth: A little function doc here describing the usage would help. Something like

> Returns true if `a` has higher download priority than b

cockatrice/src/pictureloader.cpp
Daenyth: Actually why don't we call it `setDownloadPriorityComparator`

cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+PictureToLoad::PictureToLoad(CardInfo *_card)
+    : card(_card), setIndex(0)
+{
+    if (card) {
+        sortedSets = card->getSets();
+        qSort(sortedSets.begin(), sortedSets.end(), EnabledAndKeyCompareFunctor());
+    }
+}
+
+bool PictureToLoad::nextSet()
```
Daenyth: Can we do `hasNextSet`?

cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+                        << picsPath + "/downloadedPics/" + setName + "/" + correctedCardname;
+        }
+
+        QImage image;
+        QImageReader imgReader;
+        imgReader.setDecideFormatFromContent(true);
+        bool found = false;
+
+        //Iterates through the list of paths, searching for images with the desired name with any QImageReader-supported extension
+        for (int i = 0; i < picsPaths.length() && !found; i ++) {
```
Daenyth: I think it might make this more readable/maintainable to extract the logic out to a `bool cardImageExistsOnDisk` function (or maybe return `QString`/`nullptr` as result). Then in this function you can just `if cardImageExistsOnDisk(card, set)`

cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+        QUrl redirectUrl = reply->attribute(QNetworkRequest::RedirectionTargetAttribute).toUrl();
+        QNetworkRequest req(redirectUrl);
+        qDebug() << "following redirect:" << cardBeingDownloaded.getCard()->getName() << "Url:" << req.url();
+        networkManager->get(req);
+        return;
+    }
+
+    const QByteArray &picData = reply->peek(reply->size()); //peek is used to keep the data in the buffer for use by QImageReader
+
+    // check if the image is blacklisted
```
Daenyth: Rather than the comment I'd extract out to a method and `if imageIsBlackListed(picData)`

cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+    } else {
+        QPixmapCache::insert(card->getPixmapCacheKey(), QPixmap::fromImage(image));
+    }
+
+    card->emitPixmapUpdated();
+}
+
+void PictureLoader::clearPixmapCache(CardInfo *card)
+{
+    //qDebug() << "Deleting pixmap for" << name;
```
Daenyth: I'd either uncomment this or delete the line.

cockatrice/src/pictureloader.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+void PictureLoader::clearPixmapCache()
+{
+    QPixmapCache::clear();
+}
+
+void PictureLoader::cacheCardPixmaps(QList<CardInfo *> cards)
+{
+    QPixmap tmp;
+    // never cache more than 300 cards at once for a single deck
```
Daenyth: We should pull out a constant here.

doc/cards.xsd
```diff
@@ -29,12 +29,12 @@
                           <xs:extension base="xs:string">
                             <xs:attribute type="xs:int" name="muId" use="optional"/>
                             <xs:attribute type="xs:anyURI" name="picUrl" use="optional"/>
-                            <xs:attribute type="xs:anyURI" name="picUrlHq" use="optional"/>
```
Daenyth: Why remove this? It's still valid, no? Just because we don't have a source right now doesn't mean someone might not want to supply one manually.

oracle/src/oracleimporter.h
```diff
@@ -2,6 +2,7 @@
 #define ORACLEIMPORTER_H
 
 #include <QMap>
+#include <QVariant>
```
Daenyth: Why is this needed now if no other code changed in the file?

Daenyth:
This would be a great PR to start using googletest with, though I'd be ok with leaving it out.

We should _heavily_ test this before merge as it's pretty core behavior.

cockatrice/src/pictureloader.cpp
ctrlaltca: It's not exactly a ```bool hasNextSet()```, because it also increments the internal index to point to the next set (i'd expect a hasX() method to be object const).

doc/cards.xsd
ctrlaltca: Changes included in this PR removes che choice between SQ and HQ images; you can get them form the "default source" (currently, the gatherer) or a custom source.

oracle/src/oracleimporter.h
ctrlaltca: Before, ``` #include <carddatabase.h>``` would get in some includes for Qt objects that included <QVariant> themselves. With this PR a lot of includes have been removed from carddatabase.h, and i had to add this to fix compilation.

ctrlaltca:
Anyone caring to test?

ZeldaZach:
This seems to work fine to me.

I tested it through altering URLs, Making tokens (of real cards, tokens, and fake cards that don't exist), and trying to modify the game play. No crashes of any kind detected at this point.

When altering the URLs, if the first fails it goes to the secondary URL (as it should). When both are invalid we really should default to SOMEWHERE (probably gatherer).

I like it, but I'd like to see another test to confirm. :shipit:

Daenyth:
I'll try to look at this and signoff soon

PR #1648, opened 2015-10-17
Migrate travis to use ubuntu 14 (trusty)

ctrlaltca:
Travis (finally!) permits the use of ubuntu 14 as a platform to test builds (http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/).
This PR moves our tests to this new platform.
I'm unsure wether it could be useful to keep building on ubuntu 12, too.
From a "test the code" perspective is just the same, only without the hassle of having to install third party packages from PPAs.
From a "get a redistributable package built" perspective, i hope people migrated away from 12lts to 14lts long time ago.


Daenyth:
We should build both if possible. Ubuntu 14 is not an lts release - next one is 16.04

Daenyth:
But I'm ok with dropping auto builds for 12 if it makes things easier

ctrlaltca:
Closing by now; will include in another PR.

> Ubuntu 14 is not an lts release

huh? https://wiki.ubuntu.com/LTS

Daenyth:
Derp I got it mixed up with 15

PR #1654, opened 2015-10-20
Server-size fixes for card stacking

### LINKED ISSUE #467: Object stacking bug with attachments (Regression)
Previously (months ago), if an item had an attachment then auto-placing (double-click) the same card onto the play area would make its own stack. This no longer happens; cards are placed into stacks that include attachments, creating a visual glitch whereby the new item is hidden behind an existing item with an attached card.

## Example
1. Play a card from hand. (e.g. Plains)
2. Attach a card to it (e.g. Spreading Seas)
3. Play another of the card from hand (Plains)

The attached image shows that the second plains is appearing behind the first, yet above the attachment. It implies the first card is still considered to be in the correct position despite having an offset.
itemstack_bug

This also occurs when creating tokens.

## Suggestion:
* Reimplement the check to avoid stacking cards into stacks where at least one has an attachment.

### LINKED ISSUE #1652: Morph card issue
From #1651:
When you have a morph creature in play and you double-click a creature card from your hand that is the same as the face-down creature, it enters the battlefield on top of the morph, telling your opponent, that the morph is the same creature as the one you played from your hand.

ctrlaltca:
fix #1652: don’t stack same-name card if facedown
fix #467: don’t stack same-name card if attached card

### DIFF
--- a/common/server_cardzone.cpp
+++ b/common/server_cardzone.cpp
@@ -147,7 +147,12 @@ int Server_CardZone::getFreeGridColumn(int x, int y, const QString &cardName, bo
     if (x == -1) {
         if (!dontStackSameName && freePilesMap[y].contains(cardName)) {
             x = (freePilesMap[y].value(cardName) / 3) * 3;
-            if (!coordMap.contains(x))
+            
+            if(coordMap.contains(x) && 
+                (coordMap[x]->getFaceDown() || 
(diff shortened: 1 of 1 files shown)


ZeldaZach:
cc @woogerboy21

ctrlaltca:
Merging in order to get a broader audience to test it.

PR #1670, opened 2015-11-07
Having token set to not null breaks the use of requireemail=false

jedinerd:
Having token set to not null breaks the use of requireemail=false in the servatrice.ini as token will be null in this use case.

### DIFF
--- a/servatrice/migrations/servatrice_0011_to_0012.sql
+++ b/servatrice/migrations/servatrice_0011_to_0012.sql
@@ -1,0 +1,5 @@
+-- Servatrice db migration from version 11 to version 12
+
+alter table cockatrice_users modify token binary(16) NULL;
+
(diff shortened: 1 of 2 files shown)


Daenyth:
For a sql change we'll need a migration script and we'll have to update the db version. Have a look at https://github.com/Cockatrice/Cockatrice/pull/1569/files for a simple example.

Daenyth:
Other than that, this change makes sense.

cc @woogerboy21

Daenyth:
Actually we may also need changes to the servatrice code to handle null token values if the db doesn't require it

jedinerd:
I went ahead and wrote the migration and incremented the db version.

Daenyth:
I'll try to push this the rest of the way through.

ctrlaltca:
Tested and it actually fixes the problem :+1:

PR #1682, opened 2015-11-17
link renaming

tooomm:
Description is now more clear and fits the style and wording of `How to set custom shortcuts`.

- `Linking FAQ` --> `How to set a custom picture url`
- updated define accordingly

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -30,7 +30,7 @@
 #include "soundengine.h"
 #include "sequenceEdit/shortcutstab.h"
 
-#define LINKING_FAQ_URL "https://github.com/Cockatrice/Cockatrice/wiki/Custom-Download-URLs"
+#define WIKI_CUSTOM_PIC_URL "https://github.com/Cockatrice/Cockatrice/wiki/Custom-Download-URLs"
 
 GeneralSettingsPage::GeneralSettingsPage()
 {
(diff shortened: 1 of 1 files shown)


cockatrice/src/dlg_settings.cpp
```diff
@@ -247,7 +247,7 @@ void GeneralSettingsPage::retranslateUi()
     tokenDatabasePathLabel.setText(tr("Token database:"));
     pixmapCacheLabel.setText(tr("Picture cache size:"));
     highQualityURLLabel.setText(tr("Custom Card Download URL:"));
-    highQualityURLLinkLabel.setText(QString("<a href='%1'>%2</a>").arg(LINKING_FAQ_URL).arg(tr("Linking FAQ")));
+    highQualityURLLinkLabel.setText(QString("<a href='%1'>%2</a>").arg(WIKI_CUSTOM_PIC_URL).arg(tr("How to set a custom picture url")));
```
Daenyth: Can you use `Setting a custom url source for card images` as the link text?

tooomm: So you don't want to stick with the naming pattern we use on the `Shortcuts` tab in settings?
I also think the `How to` pattern helps users the most, because it clearly tells the user that he finds help there and people don't mix it up with some kind of feature/option they can click. That was one of the reasons that I renamed it in the first place. I thought "Linking FAQ" is not perfectly clear!

But I can use your other wording if you think that it's better:
`How to set a custom url source for card images`

PR #1688, opened 2015-12-01
Add script to validate schema changes.

Daenyth:
Basic assertions about version updates


ctrlaltca:
Rebased to master and added the "bc" package for debian builds

Issue #1693, opened 2015-12-04
An internal error has occurred, please try closing and reopening your client


isochron88:
On trying to connect to any server, I get an error:

"An internal error has occurred, please try closing and reopening your client and try again. If the error persists try updating your client to the most recent build and if need be contact your software provider."

I'm using Windows Vista, Cockatrice is the client from http://www.woogerworks.com the XP/Vista version.

Daenyth:
What's the version from the About menu? Try downloading from the website again and installing, there may be a newer version. What were you doing when it happened?

isochron88:
Version in About Menu is d55e44e (2015-09-23)
What I was doing.  Started the application, from the pull down menu "Cockatrice" selected "Connect"
In pop-up window labelled "Connect to server" the radio button "previous host" was selected with "cockatrice.woogerworks.com" as the host.  Port is "4747".  Player name and Password were left blank.  Pressed "ok" button. Error ""An internal error has occurred, please try closing and reopening your client and try again..." appeared. I tried closing and restarted Cockatrice as directed.  Problem still occurs.

Nightfirecat:
That version is several months out of date from the current version. I'd recommend downloading and installing the latest weekly build and trying again.

woogerboy21:
Wait, why is the username and password left blank? (just curious)

isochron88:
The user name and password were left blank because I have never used the application before and was trying to have a look at it before I signed up. Do I have to complile from source?  or can you supply a link to the latest build?

Daenyth:
Try using register to server from the main menu. We should make sure the
username field is there...

isochron88:
I successfully registered an account by using register to server.  After that the "Connect" pull down menu option works.  I no longer get "An internal error has occurred...", I connect as expected.  Maybe change the error message to inform new players that they need to create an account before attempting to connect.  Thank you for help.

PR #1698, opened 2015-12-11
Improved login error message

### LINKED ISSUE #1694: better info & handling of empty login credentials
see https://github.com/Cockatrice/Cockatrice/issues/1693#issuecomment-162168531 + whole issue

( @isochron88 )

ctrlaltca:
If the server is reg-only and an user tries to login with an empty username, he receives a generic error that’s really hard to interpret.

Behavior after this change:

Accepy registered users only:  true

 * Empty username : "Invalid username" popup
 * Inexistant username: "This server requires registration. Do you want to register now?" popup
 * Existant username, wrong or empry password: "Incorrect username or password" popup
 * Correct username and password : login

Accepy registered users only:  false 

* Empty username : "Invalid username" popup
* Inexistant username: login as unregistered user
* Existant username, wrong or empry password: "Incorrect username or password" popup
* Correct username and password : login

Fix #1694  and related ticket #1693

### DIFF
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -387,7 +387,7 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
     QString clientId = QString::fromStdString(cmd.clientid()).simplified();
     QString clientVersion = QString::fromStdString(cmd.clientver()).simplified();
 
-    if (userName.isEmpty() || (userInfo != 0))
+    if (userInfo != 0)
         return Response::RespContextError;
 
     // check client feature set against server feature set
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -176,7 +176,8 @@ bool Servatrice::initServer()
     bool registrationEnabled = settingsCache->value("registration/enabled", false).toBool();
     bool requireEmailForRegistration = settingsCache->value("registration/requireemail", true).toBool();
 
-    qDebug() << "Registration enabled: " << regServerOnly;
+    qDebug() << "Accept registered users only: " << regServerOnly;
+    qDebug() << "Registration enabled: " << registrationEnabled;
     if (registrationEnabled)
         qDebug() << "Require email address to register: " << requireEmailForRegistration;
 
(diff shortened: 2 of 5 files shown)


common/server_protocolhandler.cpp
```diff
@@ -387,7 +387,7 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
     QString clientId = QString::fromStdString(cmd.clientid()).simplified();
     QString clientVersion = QString::fromStdString(cmd.clientver()).simplified();
 
-    if (userName.isEmpty() || (userInfo != 0))
```
ctrlaltca: This avoid the generic error message when the username is empty

servatrice/src/servatrice.cpp
```diff
@@ -176,7 +176,8 @@ bool Servatrice::initServer()
     bool registrationEnabled = settingsCache->value("registration/enabled", false).toBool();
     bool requireEmailForRegistration = settingsCache->value("registration/requireemail", true).toBool();
 
-    qDebug() << "Registration enabled: " << regServerOnly;
```
ctrlaltca: On the server startup, some options are printed to the console.
This was reporting a wrong statement mixing up the description of one option with the value of another

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -124,6 +124,8 @@ bool Servatrice_DatabaseInterface::execSqlQuery(QSqlQuery *query)
 bool Servatrice_DatabaseInterface::usernameIsValid(const QString &user, QString & error)
 {
     int minNameLength = settingsCache->value("users/minnamelength", 6).toInt();
+    if(minNameLength < 1)
```
ctrlaltca: This is a sanity check to ensure the min length of a valid username is at least 1; it just defends servatrice from having a "0" written in its config file.

servatrice/src/servatrice.cpp
Daenyth: Nice catch

servatrice/src/servatrice.cpp
```diff
@@ -176,7 +176,8 @@ bool Servatrice::initServer()
...
+    qDebug() << "Accepy registered users only: " << regServerOnly;
```
Daenyth: Typo: `Accept`

ctrlaltca: Fixed, thank you

Daenyth:
Should we prevent the user from sending login with empty name as well?

ctrlaltca:
Do you mean that the client shouldn't permit the user to connect/register if the username field is empty? Sounds reasonable

Daenyth:
Yes, that's what I mean

Daenyth:
Regardless; I'd be happy to merge this as is

ctrlaltca:
Implemented client-side check for empty player name before connect/register.

Daenyth:
Seems good. :shipit: once compile passes.

PR #1700, opened 2015-12-11
Avoid crash on dereferencing null pointer

### LINKED ISSUE #1675: Crash when "Play"ing a group of cards from the library.
Choosing to play a group of cards from the library will crash the client. (Windows 7 build d55e44e)

- - -

Steps to reproduce this crash:

1. Enable play to stack.
2. Have a deck with two cards that *would be played to the stack*.
3. Start a local game with that deck.
4. Open the library.
5. Click & drag to select both cards.
6. Right-click and select 'Play'.
  * Either it crashes, or continue to step 7.
7. Right-click another of those selected cards without clearing the selection and right-click and 'Play'.

ctrlaltca:
The game's active card can be nulled when it's moved directly from the library to the play area. Properly check if the active card pointer is null before trying to dereference it.
Fix #1675

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2115,6 +2115,9 @@ void Player::actSetPT()
 
 void Player::actDrawArrow()
 {
+    if(!game->getActiveCard())
+        return;
+
     game->getActiveCard()->drawArrow(Qt::red);
 }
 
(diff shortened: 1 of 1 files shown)


Daenyth:
*sigh*

Really need to get some Option/Maybe type in here.

Code looks fine. :shipit: post compile

PR #1705, opened 2015-12-13
Add .txt to the common deck file types

### LINKED ISSUE #1683: .txt should be added to list of common deck file types
When using "load local deck", the file type always defaults to ".cod, .dec, .mwdeck". However, a lot of my decks are in the .txt format, and it's very annoying to have to keep changing the file type every time you want to load a deck. The load local deck function should remember the last selection - if the user changed it to "All files", this should be remembered. Alternatively, we could just add .txt to the list of common deck file types. 

github file type

ctrlaltca:
fix #1683


ctrlaltca:
Merging trivial change.

PR #1707, opened 2015-12-14
Add correct encrypted api key to deploy to bintray

### LINKED ISSUE #1703: Fix travis deploy to bintray
After #1650 got in, everything is in place for the automatic deploy of osx and ubuntu packages from travis to bintray. We just need to fix the credentials stored in .travis.yml following these instrucitons: https://docs.travis-ci.com/user/deployment/bintray (the current credentials are only valid for the branch i opened the PR from)

ctrlaltca:
With this key Travis should be authorized to deploy packages built from the Cockatrice/Cockatrice repo, master branch only, to bintray.

fix #1703

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,6 @@ deploy:
   file: "build/bintray_deploy.json"
   user: "ctrlaltca"
   key:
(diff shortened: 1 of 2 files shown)


ctrlaltca:
So, the only way to test if this works is to merge to master.

PR #1721, opened 2015-12-26
Misc speedups in card database

ctrlaltca:
Misc speedup patches for card database. Nothing really impressive.
 * Lazy loading of items in card database view 
 * cache each card’s set list instead of re-joining its sets names at every reload
 * remove unused pointer to card database on each card object
 * avoid double loading of card database on error

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -157,8 +157,7 @@ void SetList::guessSortKeys()
     }
 }
 
-CardInfo::CardInfo(CardDatabase *_db,
-                   const QString &_name,
+CardInfo::CardInfo(const QString &_name,
                    bool _isToken,
                    const QString &_manacost,
                    const QString &_cmc,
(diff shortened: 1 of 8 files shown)


cockatrice/src/carddatabase.cpp
```diff
@@ -611,23 +624,22 @@ LoadStatus CardDatabase::loadCardDatabase(const QString &path, bool tokens)
...
-        qDebug() << "loadCardDatabase(): Path = " << path << " Status = " << loadStatus;
-    }
 
+    qDebug() << "loadCardDatabase(): Path =" << path << "Tokens =" << tokens << "Status =" << loadStatus;
 
     return tempLoadStatus;
 }
 
-void CardDatabase::loadCardDatabase()
+LoadStatus CardDatabase::loadCardDatabase()
```
Daenyth: :+1: for non-void functions.

Daenyth:
Merge conflicts ahoy

ZeldaZach:
Should I go ahead with this?

PR #1723, opened 2015-12-27
Coin Flips

### LINKED ISSUE #1710: Special case 2-sided dice as coin flips
Include heads or tails

> PlayerTwo flipped a coin - tails (0)
> PlayerOne flipped a coin - heads (1)

Purely cosmetic, still done via the existing roll system.

ZeldaZach:
Close #1710

This adds the message of coin flips when you use a die with exactly 2 sides.

screenshot 2015-12-26 21 08 28
screenshot 2015-12-26 21 05 41

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -190,8 +190,15 @@ void MessageLogWidget::logShuffle(Player *player, CardZone *zone)
 
 void MessageLogWidget::logRollDie(Player *player, int sides, int roll)
 {
+    QString coinOptions[2] = {tr("Heads (1)"), tr("Tails (2)")};
     soundEngine->playSound("roll_dice");
-    if (isFemale(player))
+
+    if (sides == 2)
+        if (isFemale(player))
+            appendHtmlServerMessage(tr("%1 flipped a coin. It landed as %2.", "female").arg(sanitizeHtml(player->getName())).arg("<font color=\"blue\">" + coinOptions[roll - 1] + "</font>"));
+        else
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Looks good at a glance. We may want to abstract the male/female text out of the die roll scenario. That way we cut out one of the if statements, and make it easier for future statements to be added, since that is repeated less often.

ZeldaZach:
I agree @Nightfirecat done.

Nightfirecat:
Not quite what I meant... It's the same in English, but possibly not in other languages. I meant that it should be moved into a variable and the variable should be referenced instead of using an if statement to decide if "male" or "female" should be used.

ZeldaZach:
Ahh, now I get you. While I do agree moving it around to variables would be better, this is how it's been done for a long time and it becomes a pain to do

Nightfirecat:
Might be worth putting into a new PR then... Just seems like it'd help maintainability a lot down the line to move instances like this into variables. Might look into this myself later on.

ZeldaZach:
I'm going to leave it as is for now, but once it's merged (or if you get to it first, then I'll change it up myself before this merges), i'll look into changing how it is

cockatrice/src/messagelogwidget.cpp
```diff
@@ -190,8 +190,15 @@ void MessageLogWidget::logShuffle(Player *player, CardZone *zone)
 
 void MessageLogWidget::logRollDie(Player *player, int sides, int roll)
 {
+    QString coinOptions[2] = {"Heads (1)", "Tails (2)"};
```
ctrlaltca: This needs to be translated:  ```tr()```

Daenyth:
> Not quite what I meant... It's the same in English, but possibly not in other languages. I meant that it should be moved into a variable and the variable should be referenced instead of using an if statement to decide if "male" or "female" should be used.

You can't do that because the translation tools work on the raw text in the file to generate available strings. It has to be hardcoded like this.

Daenyth:
:shipit:

PR #1722, opened 2015-12-26
Implement reverse-related tag for cards

ctrlaltca:
close #1445
close #1446
close #1447

tested using: https://github.com/Psithief/Cockatrice-tokens/blob/reverse-related/tokens.xml

PR #1734, opened 2015-12-29
Token PT shown in create related dialog

ZeldaZach:
This addresses a concern I had before (in #1722) about tokens not showing their P/T in the dialog.

screenshot 2015-12-29 18 29 44
screenshot 2015-12-29 18 29 52
screenshot 2015-12-29 18 29 56
screenshot 2015-12-29 18 30 21
screenshot 2015-12-29 18 30 32

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -660,7 +660,12 @@ void CardDatabase::refreshCachedReverseRelatedCards()
         if(card->getReverseRelatedCards().isEmpty())
             continue;
 
-        QString relatedCardName = card->getName();
+        QString relatedCardName;
+        if (card->getPowTough().size() > 0)
+            relatedCardName = card->getPowTough() + " " + card->getName(); // "n/n name"
+        else
+            relatedCardName = card->getName(); // "name"
+
         foreach(QString targetCard, card->getReverseRelatedCards())
         {
             if (!cards.contains(targetCard))
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1135,7 +1135,14 @@ void Player::actCreateRelatedCard()
 
     // get the target card name
     QAction *action = static_cast<QAction *>(sender());
-    CardInfo *cardInfo = db->getCard(action->text());
+
+    // removes p/t from tokens (and leading space))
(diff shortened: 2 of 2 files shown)


ZeldaZach:
The only minute issue I see is spacing of the names as there might sometimes be a leading space for cards without a p/t

```
1/1 Soldier
 Emblem Name
```

tooomm:
The spacing issue will just be temporary until we can identify cards by cardID, sounds like a nice workaround for now.

Nightfirecat:
Looks good, but with it comes a thought&mdash;would it be beneficial to sort the options by P/T? Or in cases like Elspeth, Sun's Champion, sort with creature tokens above other types of tokens?

I feel the spacing issue is minor enough to ship, but is also minor enough that it ought to be fixed before merging. ;)

ZeldaZach:
@Nightfirecat I have no control over that. That falls into the hands of how the tokens.xml file is created.

I'll look for a fix to the spacing thing when I figure it out

ZeldaZach:
Ok I fixed the spacing issue @Nightfirecat. Updated images

ctrlaltca:
What if the "related card" contains a ```/``` character in its name? I find this string manipulation quite tricky.

ZeldaZach:
I haven't seen a token have a `/` in it's name yet, and I don't think I will ever as it's not a normal character you see in a word

ctrlaltca:
I was thinking of cards like ```Turn // Burn```, i was wrongly assuming we used the "related card" feature for them, too.

ZeldaZach:
Yeah, the dual cards aren't "tokens" and can't be made by anything. What do you think of this PR as a whole?

ctrlaltca:
The idea helpful but the implementation is tricky; the code itself is fine and we already know that this will be probably overhauled soon switching to the use of card ids, so we could get this in until we fix the mess in the card database.

ZeldaZach:
I'm fine with the temporary solution as it works with what we have right now. When something better comes along, we can change it up

Daenyth:
:shipit:

tooomm:
How is the sorting done btw?
It seems to be alphabetical, while ignoring the p/t values?

ZeldaZach:
I believe that's how it's done @tooomm

PR #1736, opened 2015-12-30
Token coordinates

### LINKED ISSUE #1735: "Create related card" places tokens incorrectly
Tokens created by this function should be placed onto the board using the same rules the `create token` dialog uses:

* Use the tablerow information associated with the card
* Don't place the token in a group of cards

---

Currently, the tablerow information is ignored; all the cards are being created in the top row.
If the card producing tokens is in the first position on the battlefield (top row, far left), the related tokens are placed on top of it, instead of in a group of like-cards.

Can more code be re-used from the create-token dialog?

ZeldaZach:
Fix #1735

This sets the x&y coordinates of the token so they don't bunch up and they take free space only.

Before:
screenshot 2015-12-29 23 41 36

After:
screenshot 2015-12-29 23 41 19

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1138,6 +1138,7 @@ void Player::actCreateRelatedCard()
     CardInfo *cardInfo = db->getCard(action->text());
     
     // get the target token's location
+    // TODO: Define this QPoint into its own function along with the one below
     QPoint gridPoint = QPoint(-1, table->clampValidTableRow(2 - cardInfo->getTableRow()));
 
     // create the token for the related card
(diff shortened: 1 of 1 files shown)


tooomm:
> Currently, the tablerow information is ignored; all the cards are being created in the top row.

Is this issue from #1735 also addressed?

ctrlaltca:
Yes, the tablerow information is fetched from the card (token).

cockatrice/src/player.cpp
```diff
@@ -1136,6 +1136,9 @@ void Player::actCreateRelatedCard()
     // get the target card name
     QAction *action = static_cast<QAction *>(sender());
     CardInfo *cardInfo = db->getCard(action->text());
+    
+    // get the target token's location
+    QPoint gridPoint = QPoint(-1, table->clampValidTableRow(2 - cardInfo->getTableRow()));
```
Daenyth: This seems worth wrapping in a function with a comment explaining what this is doing, and why `-1` and `2` are used.

Daenyth:
:+1: and can :shipit: after my comment is addressed.

PR #1747, opened 2016-01-06
[WAIT] added webpage to about window

tooomm:
Added gh pages url (https://cockatrice.github.io) to `About Cockatrice` dialog window.


Daenyth:
Can you squash the commits to one?

PR #1758, opened 2016-01-12
fix replay progress bar

### LINKED ISSUE #1525: replay issues with 59da62d
In the 59da62d build:
- top card of the library displays a blank card back instead of the magic card back
- (unreproducible bug): sometimes graveyards and exiles are not viewable (can't double click to pop out). I think the situation that causes this to happen is when the replay is so long that the progress bar goes past the limit on the right side. Once it reaches that point, you can't view graveyards or exile anymore.
  Edit: actually, this doesn't seem to be the triggering factor. I encountered a bug once watching a long replay where I couldn't view the gy or exile, but I'm not sure what caused it. 

Additional suggestion:
When a replay is opened, it should start playing immediately. 99% of the time when I open a replay, the first thing I do is click play. Having replays start playing automatically would save effort in the vast majority of cases.

ctrlaltca:
The replay timeline draws a green rectangle used as a progress bar.
On very long replays, the rectangle width’s calculation overflowed the integer data type, resulting in a negative width value.
As a result, after a specific time the green rectangle was nor visible anymore.

This fixes #1525, or at least a part of it.

### DIFF
--- a/cockatrice/src/replay_timeline_widget.cpp
+++ b/cockatrice/src/replay_timeline_widget.cpp
@@ -58,7 +58,8 @@ void ReplayTimelineWidget::paintEvent(QPaintEvent * /* event */)
     painter.fillPath(path, Qt::black);
     
     const QColor barColor = QColor::fromHsv(120, 255, 255, 100);
-    painter.fillRect(0, 0, (width() - 1) * currentTime / maxTime, height() - 1, barColor);
+    quint64 w = (quint64)(width() - 1) * (quint64) currentTime / maxTime;
+    painter.fillRect(0, 0, w, height() - 1, barColor);
 }
 
 QSize ReplayTimelineWidget::sizeHint() const


Daenyth:
Code looks reasonable. Let's test it and see.

ctrlaltca:
@anoda9 can you test if this fix works for you?

ZeldaZach:
How long is "long" exactly? I'm testing it and don't see any issues

ctrlaltca:
I tested using ```Bal_vs_hagbard_Game01.cor``` from http://www.mtgsalvation.com/forums/search?search-thread-id=579953&by-author=Bal
Around 4/5 of the game the problem appears.

tooomm:
I had a broken replay too (mentioned in 1753) --> working for me now! :+1:

tooomm:
This does not fix all 1525! Please reopen. :)

PR #1760, opened 2016-01-14
Remove 'syntax = proto2' from .proto files.

Daenyth:
That marker is only recognized by pb3-compatible compilers, meaning that
tools which are otherwise compatible with our files break on them.
Notably, this includes hprotoc (for haskell).

### DIFF
--- a/common/pb/admin_commands.proto
+++ b/common/pb/admin_commands.proto
@@ -1,4 +1,3 @@
-syntax = "proto2";
 message AdminCommand {
     enum AdminCommandType {
         UPDATE_SERVER_MESSAGE = 1000;
--- a/common/pb/card_attributes.proto
+++ b/common/pb/card_attributes.proto
@@ -1,4 +1,3 @@
-syntax = "proto2";
 enum CardAttribute {
     AttrTapped = 1;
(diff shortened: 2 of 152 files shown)


ctrlaltca:
Related: the original repo for hprotoc seems to be this one: https://github.com/k-bx/protocol-buffers, maybe we could ask them to recognize the ```syntax```` marker or just ignore it.

Daenyth:
Done. https://github.com/k-bx/protocol-buffers/issues/23

ctrlaltca:
Cool. From the google docs i read that:
```
The first line of the file specifies that you're using proto3 syntax: if you don't do this the protocol buffer compiler will assume you are using proto2. This must be the first non-empty, non-comment line of the file.
```
so this change is not supposed to break support for people using the protobuf v3 compiler. :+1:

woogerboy21:
@ctrlaltca 
If I recall correctly , what you say is right but it does throw a bunch of warnings in during compiling and is why the lines were added to clean up output.

Daenyth:
Well, it still does compile, so I'll take warnings here over compile errors elsewhere

PR #1761, opened 2016-01-14
mtgjson now secure

ZeldaZach:
MTGJson now uses Lets Encrypt, so they go by default on https. Just changing the defaults in the Oracle tool for ease of use.

### DIFF
--- a/oracle/src/oraclewizard.cpp
+++ b/oracle/src/oraclewizard.cpp
@@ -30,13 +30,13 @@
 #include "settingscache.h"
 
 #define ZIP_SIGNATURE "PK"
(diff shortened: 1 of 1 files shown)


Daenyth:
Code looks fine, let's just test the app.

ZeldaZach:
@ctrlaltca It appears Oracle won't allow the http to redirect to https, so it's glitching out and saying no data. I feel this is the same issue as with the card URLS being redirected in the past.

ctrlaltca:
Yep, oracle doesn't follow redirects (by now). The https link works anyway.

Daenyth:
oh boy, that's going to be a flood of support requests

ctrlaltca:
We should change the url of our bugtracker aswell ;)

Issue #1780, opened 2016-01-29
Problem when creating a room with saved setting as an unregistered user


ideocl4st:
![screen shot 2016-01-29 at 10 03 50 am](https://cloud.githubusercontent.com/assets/2484055/12663990/b4fb8a2a-c66f-11e5-89b1-84a5ae2e11db.png)

tl;dr - When creating a room as an unregistered user, if your prior room setting was for 'buddies only', you have no way of unchecking 'buddies only' and thus be stuck with no one being able to join your room.

Since woogerworks is down, I tried chickatrice.net as an alternative but found out that there was a small problem that prevented me from creating a playable room.

I would generally make a room for friends only, but when the saved setting carried over to a different server without an option to register, I was only able to create a room for 'buddies only' since the option was greyed out.

This might be a problem on chickatrice.net only, but I have no way of testing this further since this is the only server where you can play as an unregistered player.

Edit - Current workaround is to create a room after unchecking 'Remember settings' then creating another room, which would revert the settings back to default.

Daenyth:
I thought this was fixed some time ago? I think there's another ticket about this issue.

Daenyth:
ok, so this is related to, but not the same as #1351 

I think it should have a pretty easy fix along the same lines as #1354

tooomm:
This can be closed I think!

ideocl4st:
Yep, now it works as it should be :)

PR #1794, opened 2016-02-05
Fixes for create game dialogue.

zebington:
Fix for #1780. Also fixes settings clear as an unregistered user.


ZeldaZach:
Seems to work as expected

Daenyth:
Thanks for the contribution!

PR #1796, opened 2016-02-06
Add "Import custom set" function

### LINKED ISSUE #1752: dialog to import custom sets
requesting a dialog to import custom sets, instead of the current process which is manually opening your custom sets folder, pasting the file, restarting cockatrice, and finally seeing a 'would you like to add the new set' dialog.

zebington:
Adds a menu option underneath "open custom sets folder" to add a custom set via a file picker.
Closes #1752.


cockatrice/src/dlg_add_set.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <QVBoxLayout>
+
+DlgAddSet::DlgAddSet(QWidget *parent, bool success) : QDialog(parent) {
+    status = new QLabel(this);
+    restart = new QLabel(this);
+
+    if (success) {
+        setWindowTitle(tr("Success"));
+        status->setText(QString("Set added to Cockatrice."));
+        restart->setText(QString("You must restart to use your new set."));
```
tooomm: "You must restart **Cockatrice** to use *your* new set."
Otherwise some people might want to restart their PC... :)

I would prefer "the" over "your", too.

Daenyth: I'll try to review this tonight. Overall a needed change

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -45,6 +46,10 @@
 #include "cardframe.h"
 #include "filterbuilder.h"
 
+const QStringList TabDeckEditor::fileNameFilters = QStringList()
+    << QObject::tr("Cockatrice set format (*.xml)")
```
tooomm: There is no "set format" or sth, we just have a card databases...
People can import a xml database with cards from a whole block or any other combination of cards/sets as well.
(whatever kind of cards/sets are in there it needs to be a xml file)

`Cockatrice card database (*.xml)` might be more general and better suitable.


<br>
Thinking further... should the menu say something like `Add custom set(s) or card(s)` or more simple `Add custom sets/cards` ?

tooomm:
I just tested it with the challenge decks tokens file:
https://github.com/Cockatrice/Magic-Token

Working. And the file is properly placed in the `customsets` folder.
After restarting I get asked if the newly found set should be enabled. :+1: 

We need to update the [wiki](https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-%26-Sets) as well.

Psithief:
Shouldn't it happen without restarting?

zebington:
I haven't rewritten the code that checks for sets. All this does is copy the set file into the custom set folder.

Daenyth:
> Shouldn't it happen without restarting?

I'm fine with it needed a restart for now. It's less error prone

cockatrice/src/dlg_add_set.cpp
```diff
@@ -0,0 +1,37 @@
+#include "dlg_add_set.h"
+
+#include <QDialogButtonBox>
+#include <QPushButton>
+#include <QVBoxLayout>
+
+DlgAddSet::DlgAddSet(QWidget *parent, bool success) : QDialog(parent) {
```
Daenyth: This should probably be called AddCustomSet to be more clear.

cockatrice/src/dlg_add_set.cpp
```diff
@@ -0,0 +1,37 @@
...
+    status = new QLabel(this);
+    restart = new QLabel(this);
+
+    if (success) {
```
Daenyth: I'm not sure it makes sense to check this on instantiation. It seems like that should be a separate modal.

A workflow like

Settings panel > click add custom set button > open qt file picker > dialog for result makes the most sense.

Maybe I'm getting confused from the class name? Is this actually an AddCustomSetResult?

cockatrice/src/dlg_add_set.cpp
```diff
@@ -0,0 +1,37 @@
...
+        setWindowTitle(tr("Success"));
+        status->setText(QString("Sets/cards added to Cockatrice."));
+        restart->setText(QString("You must restart Cockatrice to use the new sets/cards."));
+    }
+    else {
+        setWindowTitle(tr("Failed"));
+        status->setText(QString("Sets/cards failed to import."));
```
Daenyth: We should pipe through the failure reason into the dialog so that the user has an actionable outcome. Showing the error with no suggested action is likely to confuse people.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -303,11 +311,12 @@ void TabDeckEditor::createMenus()
     dbMenu->addSeparator();
     dbMenu->addAction(aClearFilterOne);
     dbMenu->addAction(aClearFilterAll);
-#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
     dbMenu->addSeparator();
+#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
     dbMenu->addAction(aOpenCustomFolder);
```
Daenyth: Does anyone know why this isn't enabled on linux?

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +873,43 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
 
 }
 
+void TabDeckEditor::actAddCustomSet()
+{
+
+#if QT_VERSION < 0x050000
```
Daenyth: Personal preference; Can you check against [`QT_VERSION_CHECK`](http://doc.qt.io/qt-5/qtglobal.html#QT_VERSION_CHECK) instead of the hex string? I find it more readable.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +873,43 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+    QString dataDir = QDesktopServices::storageLocation(QDesktopServices::DataLocation);
+#else
+    QString dataDir = QStandardPaths::standardLocations(QStandardPaths::DataLocation).first();
```
Daenyth: You don't need to fix it for this PR, but we should make a standard convenience utility that just returns the QString so we don't have to inline the `#if` checks all over.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +873,43 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+    dialog.setNameFilters(TabDeckEditor::fileNameFilters);
+    if (!dialog.exec())
+        return;
+
+    QString fileName = dialog.selectedFiles().at(0);
+    QDir dir(dataDir.append("/customsets"));
+    QStringList files = dir.entryList();
+    int maxIndex = 0;
+    for (int i = 0; i < files.size(); ++i) {
+        int fileIndex = files.at(i).split(".").at(0).toInt();
```
Daenyth: I don't understand what the intent is here. Could you move it to an appropriately named function? Also, I'd strongly prefer more functional style `map` calls and stl iterator interactions instead of raw `for` loops. It's less likely to introduce errors and the declarative style conveys intent a lot more clearly, at least for me.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +873,43 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+        if (fileIndex > maxIndex)
+            maxIndex = fileIndex;
+    }
+    maxIndex++;
+    bool res = QFile::copy(
```
Daenyth: Can you check [`QFile::error`](http://doc.qt.io/qt-4.8/qfile.html#error) to see why it failed?

cockatrice/src/tab_deck_editor.cpp
zebington: I don't know, but that's how the code was originally, I've just left it as close to the original as possible.

cockatrice/src/dlg_add_set.cpp
zebington: Yes, it's supposed to be AddCustomSetResult.

cockatrice/src/tab_deck_editor.cpp
zebington: This grabs the next index for custom sets, which must start with a number code, zero-filled to two digits, at least according to the wiki, will make suggested changes.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -45,6 +46,10 @@
 #include "cardframe.h"
 #include "filterbuilder.h"
 
+const QStringList TabDeckEditor::fileNameFilters = QStringList()
+    << QObject::tr("Cockatrice card database (*.xml)")
```
Daenyth: The first of these becomes the default right?

cockatrice/src/tab_deck_editor.cpp
Daenyth: That's fine, more of a question for anyone else watching.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +870,55 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+    QFileDialog dialog(this, tr("Load sets/cards"));
+    dialog.setDirectory(dataDir);
+    dialog.setNameFilters(TabDeckEditor::fileNameFilters);
+    if (!dialog.exec())
+        return;
+
+    QString fileName = dialog.selectedFiles().at(0);
+
+    if (!QFile::exists(fileName)) {
+        DlgAddSetResult dlg(this, false, QString("Selected file cannot be found."));
```
Daenyth: Maybe `No file was found at $path` ?

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +870,55 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+    }
+
+    QDir dir(dataDir + "/customsets");
+    int nextPrefix = getNextCustomSetPrefix(dir);
+
+    bool res = QFile::copy(
+        fileName, dir.absolutePath() + "/" + (nextPrefix > 9 ? "" : "0") +
+        QString::number(nextPrefix) + "." + QFileInfo(fileName).fileName()
+    );
+
```
Daenyth: Can you add an

```c++
bool copySucceeded = QFile::copy(...)
if (!copySucceeded) {
    FileError error = QFile:error();
    QString errmsg = fileErrorToString(error) // implement this - qt docs tell what each enum value means
    DlgAddSetResult dlg(this, copySucceeded, errmsg)
} else {
    DlgAddSetResult dlg(this, copySucceeded, QString());
}
dlg.exec();
```

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +870,55 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+        DlgAddSetResult dlg(this, false, QString("Selected file cannot be found."));
+        dlg.exec();
+        return;
+    }
+
+    QDir dir(dataDir + "/customsets");
+    int nextPrefix = getNextCustomSetPrefix(dir);
+
+    bool res = QFile::copy(
+        fileName, dir.absolutePath() + "/" + (nextPrefix > 9 ? "" : "0") +
```
Daenyth: The target path logic here is long enough that it's worth a function.

```C++
bool copySucceeded = QFile::copy(
    fileName,
    nextAvailableCustomSetName(fileName, dir) // you can get nextPrefix inside this
)
```

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -863,6 +870,55 @@ void TabDeckEditor::actOpenCustomsetsFolder() {
...
+}
+
+int TabDeckEditor::getNextCustomSetPrefix(QDir dataDir) {
+    QStringList files = dataDir.entryList();
+    int maxIndex = 0;
+
+    QStringList::const_iterator filesIterator;
+    for (filesIterator = files.constBegin(); filesIterator != files.constEnd(); ++filesIterator) {
+        int fileIndex = (*filesIterator).split(".").at(0).toInt();
+        if (fileIndex > maxIndex)
```
Daenyth: It looks like there's an `std` function for getting max index; http://en.cppreference.com/w/cpp/algorithm/max_element

cockatrice/src/tab_deck_editor.cpp
zebington: Yes

cockatrice/src/tab_deck_editor.cpp
zebington: QFile::error() isn't a static function, it's not designed to be used like that. It also doesn't give you information on the copy, just returns a QFile::CopyError, which doesn't give us anymore information. I can make it say "File failed to copy" if you want. As it is, I don't see the copying failing as long as Cockatrice always has write permissions and there's space on the disk.

Daenyth: Derp, of course you're right - I missed that. Annoying it doesn't tell you why it failed. I can imagine permission error being a big one.

ZeldaZach:
Can you `rebase -i master` this? Then i'll look into a merge

Daenyth:
Commit history is in good shape, I don't think it needs to be squashed.
I'll do a final look tonight if I have time, we can merge after.

Daenyth:
Code looks good to go :shipit: 

@ZeldaZach are we happy with the testing for this so far? Go ahead and merge when you're ready to sign off.

@zebington Sorry for the delay, and thanks for contributing!

tooomm:
I tried the newest build as well.
It's working and properly importing, renaming and copying the file into the custom set folder.

Daenyth:
Thanks!

PR #1803, opened 2016-02-10
Update replay maintenance script

woogerboy21:
Update the replay maintenance script to allow for rolling replay maintenance along with cleanup of the two associated db tables (replay access and games).  This updated script will now allow players to see the past X days (8 by default) of replays rather than loosing all replays from the point in the the script gets run.

### DIFF
--- a/servatrice/scripts/linux/maint_replays
+++ b/servatrice/scripts/linux/maint_replays
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-# SCHEDULE WITH CRONTAB BASED ON TIME PERIOD REPLAYS SHOULD BE SAVED UNTIL (EX: SCHEDULE ONCE A WEEK TO KEEP A WEEKS WORTH OF REPLAYS IN THE DB)
-
+# SCHEDULE WITH CRONTAB DAILY
 DBNAME="servatrice"					#set this to the database name used
 TABLEPREFIX="cockatrice"			#set this to the prefix used for the table names in the database (do not inclue the _)
 SQLCONFFILE="./mysql.cnf" 			#set this to the path that contains the mysql.cnf file
-mysql --defaults-file=$SQLCONFFILE -h localhost -e 'truncate table ""$DBNAME"".""$TABLEPREFIX""_replays;truncate table ""$DBNAME"".""$TABLEPREFIX""_replays_access'
+mysql --defaults-file=$SQLCONFFILE -h localhost -e 'delete from servatrice.cockatrice_games where time_finished < DATE_SUB(now(), INTERVAL 8 DAY)'
+mysql --defaults-file=$SQLCONFFILE -h localhost -e 'delete from servatrice.cockatrice_replays where id_game NOT IN (select id from servatrice.cockatrice_games)'
+mysql --defaults-file=$SQLCONFFILE -h localhost -e 'delete from servatrice.cockatrice_replays_access where id_game NOT IN (select id from servatrice.cockatrice_games)'


Daenyth:
I'd like to see this as a .sql file rather than inline calls.

servatrice/scripts/linux/maint_replays
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
```
Daenyth: This script really needs a `set -e`

servatrice/scripts/linux/maint_replays
```diff
@@ -1,8 +1,9 @@
...
-# SCHEDULE WITH CRONTAB BASED ON TIME PERIOD REPLAYS SHOULD BE SAVED UNTIL (EX: SCHEDULE ONCE A WEEK TO KEEP A WEEKS WORTH OF REPLAYS IN THE DB)
-
+# SCHEDULE WITH CRONTAB DAILY
 DBNAME="servatrice"					#set this to the database name used
 TABLEPREFIX="cockatrice"			#set this to the prefix used for the table names in the database (do not inclue the _)
 SQLCONFFILE="./mysql.cnf" 			#set this to the path that contains the mysql.cnf file
-mysql --defaults-file=$SQLCONFFILE -h localhost -e 'truncate table ""$DBNAME"".""$TABLEPREFIX""_replays;truncate table ""$DBNAME"".""$TABLEPREFIX""_replays_access'
+mysql --defaults-file=$SQLCONFFILE -h localhost -e 'delete from servatrice.cockatrice_games where time_finished < DATE_SUB(now(), INTERVAL 8 DAY)'
+mysql --defaults-file=$SQLCONFFILE -h localhost -e 'delete from servatrice.cockatrice_replays where id_game NOT IN (select id from servatrice.cockatrice_games)'
+mysql --defaults-file=$SQLCONFFILE -h localhost -e 'delete from servatrice.cockatrice_replays_access where id_game NOT IN (select id from servatrice.cockatrice_games)'
```
Daenyth: This might not be needed if mysql lets us mark the key as a foreign key and automatically cascade deletes.

Postgres lets you do this, does mysql? That way when we delete a cockatrice_games row, the _replays and _games are automatically deleted.

woogerboy21: I don't think we can do this with the current db engine (MyISAM), I believe its possible with innodb though.  Which is something we may want to consider doing.

servatrice/scripts/linux/maint_replays
woogerboy21: Refresh my memory as to what set -e does again?

Daenyth: If any command exists non-zero, it aborts the script.

>              -e      Exit  immediately if a simple command (see SHELL GRAMMAR above) exits with a non-zero status.  The shell does not exit if the command that fails is part of the
                      command list immediately following a while or until keyword, part of the test in an if statement, part of a && or || list, or if the command's return value  is
                      being inverted via !.  A trap on ERR, if set, is executed before the shell exits.

servatrice/scripts/linux/maint_replays
Daenyth: I'm OK with doing it this script-based way for now but I'd like to move to an in-db option later.

How about 

delete_old_replays.sql
```sql
START TRANSACTION;
DROP TEMPORARY TABLE IF EXISTS old_games;
CREATE TEMPORARY TABLE old_games AS (SELECT id FROM cockatrice_games WHERE time_finished < DATE_SUB(now(), INTERVAL 8 DAY));

DELETE FROM cockatrice_replays r JOIN old_games g ON r.id_game = g.id;
DELETE FROM cockatrice_replays_access ra JOIN old_games g ON ra.id_game = g.id;
DELETE FROM cockatrice_games g JOIN old_games og ON g.id = og.id;

DROP TEMPORARY TABLE old_games;
COMMIT;
```

woogerboy21: Whats the advantage to doing it the way you suggest?  Only thing I can see is the single transaction?  Or is there something else I don't understand?  Seems to me like your actually increasing the amount of system resources being required by creating a temporary table, populating it, and then still deleting the contents from the original tables and then dropping the new temporary table.

Daenyth: Does doing the delete actually take a lot of resources?

The advantage is that you know you're not deleting anything else. With your script, a game row could be deleted and you'll delete the other things for it even though it wasn't part of the set of things you were intending to delete

woogerboy21: I can see your reasoning.  The question I guess I would have is "Should the replays and access table have entries in it that the games table does not have?"

I was approaching the problem with the mind set that if the games table does not contain the game information in it, the access and replay tables should be pruned / cleaned to match.  That being said the important query in the script is the first delete and the other two could technically be executed at any time.  It just so happens that they are executed consecutively.

Either solution is better than the original truncating of the tables IMO.

ctrlaltca: Imho:
 1. move the schema to use innodb tables (better locking, more resource usage but better scalability)
 2. add the necessary constraints (read : foreign keys on cascade delete) so that we can't have eg. a "replay player access" without a "replay"
 3. implement a migration script
 At this point we only need to delete the row in the "games" table to get everything cleaned up by mysql automatically.
We can then thinking on implementing a scheduled, configurable purge of games data older than "x" days directly from inside servatrice.

Daenyth: I totally agree with that plan.

@woogerboy21 if you want to use this script for now I think just do it locally until we do as ctrl says.

Daenyth:
Any objection to closing in favor of #1805 ?

woogerboy21:
Its up to you, if you think #1805 will be implemented in a timely fashion go for it.  Otherwise I would commit the updated script and then simply remove the script when #1805 gets implemented.  The original script works but doesnt clean all the tables accordingly.  The replays and replay access table gets truncated but the games table keeps the games information.

Daenyth:
That's reasonable. :shipit:

Issue #123, opened 2014-06-24
Chat widget is affected by KDE "open files with single click" setting


Daenyth:
When 'single click' is turned on, clicking a user in the chat list opens the default action for them.

When it's off, clicking a user selects their name first

ZeldaZach:
Doesn't seem to be an issue with latest releases

PR #1807, opened 2016-02-11
Correctly save and retrieve docks size in game tabs

### LINKED ISSUE #1777: Investigate better titlebar labels for dock elements in game
Extracted from @tooomm 's comment in #1753 : https://github.com/Cockatrice/Cockatrice/pull/1753#issuecomment-173601286.
Currently when in-game widgets are undocked, their titlebar is made up of the concatenation of the widget name and the game tab's label; this is sub-optimal, since the titlebar label can contain a very long text that always gets elided since the undocked widgets are quite small.

### LINKED ISSUE #1792: Game Tab Doesn't Save Configuration
When you  move elements around in a game tab, they don't store that config when you open a new game up. This gets annoying as the default isn't even "default" because if you hit reset layout it'll change.

ctrlaltca:
Fix #1792 
Fix #1777

### DIFF
--- a/cockatrice/src/settings/layoutssettings.cpp
+++ b/cockatrice/src/settings/layoutssettings.cpp
@@ -78,6 +78,39 @@ const QByteArray LayoutsSettings::getGamePlayAreaGeometry()
     return getValue("layouts/gameplayarea_geometry").toByteArray();
 }
 
+const QSize LayoutsSettings::getGameCardInfoSize()
+{
+    QVariant previous = getValue("layouts/gameplayarea_CardInfoSize");
+    return previous == QVariant() ? QSize(250,360) : previous.toSize();
+}
+
+void LayoutsSettings::setGameCardInfoSize(const QSize &value)
+{
+    setValue(value,"layouts/gameplayarea_CardInfoSize");
+}
+
+const QSize LayoutsSettings::getGameMessageLayoutSize()
+{
+    QVariant previous = getValue("layouts/gameplayarea_MessageLayoutSize");
+    return previous == QVariant() ? QSize(250,250) : previous.toSize();
+}
+
+void LayoutsSettings::setGameMessageLayoutSize(const QSize &value)
+{
+    setValue(value,"layouts/gameplayarea_MessageLayoutSize");
+}
+
+const QSize LayoutsSettings::getGamePlayerListSize()
+{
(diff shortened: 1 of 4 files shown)


ZeldaZach:
> Point 1 (widths on reset layout) should be fixed.
> I've not been able to fully reproduce point 2; only when i move 2 dock in the same place (eg. in your pic, the "filters" and "deck" docks are in the same place), then Qt will grow the dock size to make room for the bigger of the docks.
> Point 3: i know this is suboptimal, but the only other way i can think of is to require users to manually press a "save this layout as the default layout for new games" button, and i'm not sure if this would be considered easier or harder for users. 

Widths reset on layout is now fixed. :+1: 
Point 2 is no longer bugging on on me, so maybe it was something related? Either way :+1: 

I'm fine with it working as it is for Point 3. I see no reason to hold this up as it does everything needed. Can you possibly change how the popups work with the header text though? I know we discussed it in the past:

screenshot 2016-02-11 12 53 13

To just the room number? `#8104928592`

ctrlaltca:
Do you mean #1777? I'm looking for feedback before making the change, i already know @tooomm and you share this same idea, and i'm neutral; @Daenyth your opinion please?

ZeldaZach:
Yeah #1777 is what I mean. Didn't realize a ticket was open, oops

ZeldaZach:
I'll append my earlier comment here; The deck editor seems to glitch out if you disconnect / reconnect to a server with the deck editor closed. It re-added elements that I closed (like filters) and changed the layout. The game seems to be fine however

Daenyth:
@ctrlaltca I don't understand what feedback you'd like. The ticket description doesn't really have any open questions. I agree that it should be changed to contain only relevant info. Game number and a description of what the widget is. IE `Players | Game #123`

ctrlaltca:
> I agree that it should be changed to contain only relevant info

Sorry, i'm a bit tired and wrote "feedback" out of context. I just meant to ask your (and other's) opinion on the idea of changing the docks titlebar to only show the game id in order to avoid the titlebar text being truncated. You answered me anyway, so i'll go on making the requested change.

ctrlaltca:
Fix for #1777 added.

ctrlaltca:
Is this good to go?

ZeldaZach:
Yup good to go, just tested again :+1: :shipit:

PR #1808, opened 2016-02-11
Move schema to Innodb and add FKs

### LINKED ISSUE #1805: Change DB to innodb, use constraints for replays
> Imho:
 1. move the schema to use innodb tables (better locking, more resource usage but better scalability)
 2. add the necessary constraints (read : foreign keys on cascade delete) so that we can't have eg. a "replay player access" without a "replay"
 3. implement a migration script

>  At this point we only need to delete the row in the "games" table to get everything cleaned up by mysql automatically.
We can then thinking on implementing a scheduled, configurable purge of games data older than "x" days directly from inside servatrice.

[~ ctrlaltca](https://github.com/Cockatrice/Cockatrice/pull/1803#discussion_r52504123)

ctrlaltca:
This PR modifies servatrice’s database schema to use InnoDB instead of MyIsam; a migration script is included.
Two minor changes have been added to queries on the code:
 1. remove hardcoded table prefix from cockatrice_log
 2. accomodate the replay insert query to work with constraints

Needs testing on a server with more than a single user :)
Fix #1805

### DIFF
--- a/servatrice/migrations/servatrice_0012_to_0013.sql
+++ b/servatrice/migrations/servatrice_0012_to_0013.sql
@@ -1,0 +1,86 @@
+-- Servatrice db migration from version 12 to version 13
+
+-- WARNING: this is quite a big change, so you really, REALLY should
+-- backup your database before attempting to execute this migration.
+
+-- First move all the tables to the InnoDB engine
+ALTER TABLE `cockatrice_schema_version` ENGINE=InnoDB;
+ALTER TABLE `cockatrice_decklist_files` ENGINE=InnoDB;
+ALTER TABLE `cockatrice_decklist_folders` ENGINE=InnoDB;
+ALTER TABLE `cockatrice_games` ENGINE=InnoDB;
+ALTER TABLE `cockatrice_games_players` ENGINE=InnoDB;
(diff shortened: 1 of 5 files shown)


Daenyth:
Does this require any changes in install dependencies for server admins?

ctrlaltca:
The InnoDB storage engine should be bundled with any mysql-server package. MariabDB bundles XtraDB, which is their own enhanced version based on InnoDB.
IIRC debian used to disable InnoDB by default in old versions, but that was a really long time ago (around the mysql 3 to 4 move)

Daenyth:
:+1:

ctrlaltca:
Squashed commits

ZeldaZach:
I don't know how to properly test this without setting up an entire build environment with a server, so @woogerboy21 can you check this out?

Daenyth:
I thought I emailed to this, but I guess it didn't go through. I have servatrice+mysql setup instructions on the wiki; https://github.com/Cockatrice/Cockatrice/wiki/Setting-up-Servatrice#using-docker

Daenyth:
I can create a server using the full schema (migration isn't tested) with the docker setup, and connect to the server.

PR #1821, opened 2016-02-18
todo file - removed usermanual part

tooomm:
We copied the usermanual to the wiki some time ago and wanted to convert it to a getting started reference.

The aspects are outdated and not very relevant any longer at this location.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -10,13 +10,6 @@ Note that "improve" and "write" always also means: "document and comment"
 ##Scripts
 * Write example init script for servatrice.
 
-##Improve usermanual:
-* Improve Layout, Text, Screenshots; nearly everything is still quick n dirty copy and paste.
-* Add missing TODOs
-* Translate the Documentation
-* Rename the picture filenames to something more meaningful.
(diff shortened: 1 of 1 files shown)


ctrlaltca:
:+1: for removing old entries from the TODO file.
Where has the content of the usermanual been moved in the wiki? It seems to miss all the part about "how to play", eg.: https://github.com/Cockatrice/Cockatrice/commit/854e0a764c307c6637320d39b0c9cb1f075a4aef#diff-6c0ac1534818ae3732b739b1e705a937L522

tooomm:
We just copied the content from the old documentation to the wiki to make it easier accessible by the users and better maintainable by us.

The problem with the old version was that it was heavily outdated and containing some (maybe) problematic content. Therefore we removed all pictures and some parts that are not valid anymore or have their own wiki section already.

In comparison to the old file you linked to the wiki:
https://github.com/Cockatrice/Cockatrice/wiki/Getting-Started#finding-your-way-around
The only difference is the missing picture in this section.


But yes, in general it's still a big mess - missing many things and not too useful overall at the moment.

PR #1828, opened 2016-02-23
Fix libmysqlclient detection on 64-bit linux

### LINKED ISSUE #1827: Misleading CMake DB Error
When compiling from source using CentOS 7 and Qt5.5.1 users are presented with the following error message during the cmake command using the -DWITH_SERVER=1 option.

```
-- Mysql connector NOT FOUND: servatrice won't be able to connect to a mysql server
```

Even though the server compiles and DB connectivity to the mysql/mariadb app is available and functions correctly.

I noticed this as well when trying to compile using ubuntu but didnt get to testing if DB connectivity worked using that build/test.

ctrlaltca:
Added 64bit libraries each path for libmysqlclient.
Should fix #1827


Daenyth:
:+1:

woogerboy21:
Odd, this didnt seem to work on my test box.  Let me see what path things are in here.

woogerboy21:
For whatever reason the library file on this box is named libmysqlclient.so.18.0.0.  The updated /usr/lib64 path is correct though and this change should be made.  Just want to make sure we are not trying to search for alternate versions of the library name?  If its supposed to be than its still broke but if not its good.

ctrlaltca:
In my box  /usr/lib64/mysql/libmysqlclient.so is a symlink to the current installed version of the lib:
```
[root@xxx ~]# ls -l /usr/lib64/mysql/libmysqlclient*
lrwxrwxrwx 1 root root      17 19 dic 12.03 /usr/lib64/mysql/libmysqlclient_r.so -> libmysqlclient.so
lrwxrwxrwx 1 root root      20 19 dic 12.03 /usr/lib64/mysql/libmysqlclient.so -> libmysqlclient.so.18
lrwxrwxrwx 1 root root      24 19 dic 11.58 /usr/lib64/mysql/libmysqlclient.so.18 -> libmysqlclient.so.18.0.0
-rwxr-xr-x 1 root root 3133544 21 nov 16.32 /usr/lib64/mysql/libmysqlclient.so.18.0.0
```

woogerboy21:
Yea I had to create the symlink for the plain .so file name.(for whatever reason).
All I had was:
```
lrwxrwxrwx 1 root root   24 Jan 25 05:34 libmysqlclient.so.18 -> libmysqlclient.so.18.0.0
-rwxr-xr-x 1 root root 3.0M Nov 21 16:32 libmysqlclient.so.18.0.0

```

ctrlaltca:
Looking at the installed packages i saw i have both the -libs and -devel packages installed:
```
[root@xxx ~]# yum list installed | grep maria
mariadb.x86_64                      1:5.5.44-2.el7.centos            @base      
mariadb-devel.x86_64                1:5.5.44-2.el7.centos            @base      
mariadb-libs.x86_64                 1:5.5.44-2.el7.centos            @base      
mariadb-server.x86_64               1:5.5.44-2.el7.centos            @base
```

The /libs package seems to provide the library itself:
```
[root@xxx ~]# repoquery -l mariadb-libs | grep libmysqlclient
/usr/lib/mysql/libmysqlclient.so.18
/usr/lib/mysql/libmysqlclient.so.18.0.0
/usr/lib64/mysql/libmysqlclient.so.18
/usr/lib64/mysql/libmysqlclient.so.18.0.0
```

While the -devel package adds the symlinks you were missing:
```
[root@xxx ~]# repoquery -l mariadb-devel | grep libmysqlclient
/usr/lib/mysql/libmysqlclient.so
/usr/lib/mysql/libmysqlclient_r.so
/usr/lib64/mysql/libmysqlclient.so
/usr/lib64/mysql/libmysqlclient_r.so
```

Daenyth:
We should list that in the wiki as a dependency

PR #1832, opened 2016-02-24
move sounds+themes folder inside cockatrice

tooomm:
After talking to @Daenyth and @ctrlaltca this seems to be a better place for them.
They are just related to Cockatrice, but still they were placed inside the root directory of the repo.

`sounds` and `themes` moved into `cockatrice/`

moved subdirectory reference
`/cmakelists` to `/cockatrice/cmakelists`


Nightfirecat:
Would these files be retained after uninstallation/reinstallation for app upgrades?

tooomm:
@Nightfirecat I just changed the location in the github repo and told cmake where to look for them during compiling.

Everything else should be uneffected, like the same placement in the app folder as before and same uninstalling behavior etc.

Speaking of that, I don't know if they get removed (they are actually I guess, which makes sense since they are placed inside the main app folder), but they maybe should not since users can customize them and want to keep them while switching to a newer version of the app.

Moving them to the user directory like custom sets, deck files, settings etc. should be a different issue/pr anyway.
(or introducing a new folder set for custom themes and sounds there to keep the user profile as small as possible!)

Nightfirecat:
Ah, alright.

ctrlaltca:
Looks fine!

PR #1834, opened 2016-02-24
Warning/ban notifications: null check userInfo pointer before dereferencing it

ctrlaltca:
Should help avoiding crash caused by null derefefernce


ZeldaZach:
Seems good to me, nice catch!

Daenyth:
:shipit:

woogerboy21:
FYI, i just rolled this in to the code at woogerworks.  So far so good.

PR #1850, opened 2016-03-08
Fix mysql plugin deploy on windows

ctrlaltca:
Aka: make servatrice able to work with mysql under windows.

Fixes two separate issues:
 * qt’s mysql plugin was not being bundled in the package
 * appveyor builds were always bundling the 64-bit version of libmysql.dll. Now the x86 package correctly contains a 32-bit version of the library.

Note: servatrice+mysql will work only on OS >= Windows Vista. This is due to mysql not supporting anymore Windows Xp since mysql connector 6.1.2 (https://dev.mysql.com/doc/relnotes/connector-c/en/news-6-1-2.html)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,12 +11,18 @@ environment:
           target_arch: x86_64
           qt_ver: 5.5\msvc2013_64
           bintray_path: Win64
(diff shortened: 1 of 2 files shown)


woogerboy21:
:+1:   We should be able to remove the wiki section https://github.com/Cockatrice/Cockatrice/wiki/Compiling-Cockatrice#connecting-and-configuring-mariadb-for-use-with-servatrice  after this is committed making windows users a lot happier :smile:

PR #1866, opened 2016-03-15
Prevent trayIcon Null Crash

### LINKED ISSUE #1860: Cockatrice crashes when trying to show popup notifications
<b>OS:</b>
FreeBSD 10.2-STABLE

<b>Cockatrice version:</b>
cbea432 (2016-03-08)

___
<br>

I noticed that Cockatrice would segfault any time someone tried to PM me. Running it in a debugger, I found out that trayIcon is NULL in TabMessage::showSystemPopup, but the code assumes that it is valid. It looks like TabRoom::actShowMentionPopup would cause the same problem as well.

ZeldaZach:
Fix #1860 

Adds `trayIcon != NULL` to the checks prior to attempting to alter the trayIcon to prevent a crash.

### DIFF
--- a/cockatrice/src/tab_message.cpp
+++ b/cockatrice/src/tab_message.cpp
@@ -5,6 +5,7 @@
 #include <QAction>
 #include <QSystemTrayIcon>
 #include <QApplication>
+#include <QDebug>
 #include "tab_message.h"
 #include "abstractclient.h"
 #include "chatview.h"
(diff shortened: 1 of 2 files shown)


ctrlaltca:
```if(trayIcon)``` is enough, in c++ ```NULL == 0```, and zero already evaluates to boolean ```false```.

**Comment by ZeldaZach** [comment]
Hmm, didn't know that. Thanks for the enlightenment!!

File: cockatrice/src/tab_message.cpp [hunk 1]
```diff
@@ -131,9 +131,11 @@ bool TabMessage::shouldShowSystemPopup(const Event_UserMessage &event) {
 }
 
 void TabMessage::showSystemPopup(const Event_UserMessage &event) {
-    disconnect(trayIcon, SIGNAL(messageClicked()), 0, 0);
-    trayIcon->showMessage(tr("Private message from ") + otherUserInfo->name().c_str(), event.message().c_str());
-    connect(trayIcon, SIGNAL(messageClicked()), this, SLOT(messageClicked()));
+    if (trayIcon) {
```
Daenyth: Can we add a `qDebug` for the else case

ZeldaZach: Done

Daenyth:
:+1: once it compiles cleanly.

ctrlaltca:
```
/home/travis/build/Cockatrice/Cockatrice/cockatrice/src/tab_message.cpp: In member function ‘void TabMessage::showSystemPopup(const Event_UserMessage&)’:
/home/travis/build/Cockatrice/Cockatrice/cockatrice/src/tab_message.cpp:141:16: error: invalid use of incomplete type ‘struct QDebug’
```
You need to ```#include <QDebug>``` somewhere at the beginning of the file

PR #1873, opened 2016-03-17
Servatrice: Split requireemail into requireemail and requireemailaction

ctrlaltca:
> @woogerboy21:
 i see a potential need for having the ability to require registration + email but not send activation tokens (simply just activate the account) is it adds one more criteria that could be used for validating a users account really is that user when they "forget" there password and request a reset.
I am always hesitant to reset some one's password that sends me an email saying they forgot on an account that does not have an email address with it since it could be just some one randomly trying to hijack another users account
its really not that difficult to run a DB query to activate the accounts on a timed bases. the problem with using that method is the users are prompted to enter a token that never comes and most users end up sending a message to the server operator rather than simply retrying there login

This PR splits the "requireemail" servatrice setting into:
 * "requireemail": simple regexp validation on the email
 * requireemailactivation: send token email, require users to input their activation token

### DIFF
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -105,10 +105,13 @@ allowpunctuationprefix=false
 ; Enable this feature? Default false.
 ;enabled=false
 
-; Require users to provide an email address in order to register. Newly registered users will receive an
-; activation token by email, and will be required to input back this token on cockatrice at the first login
-; to get their account activated. Default true.
+; Require users to provide an email address in order to register. Default true.
 ;requireemail=true
+
+; Require email activation. Newly registered users will receive an activation token by email,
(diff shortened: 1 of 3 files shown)


Daenyth:
For new config options can we use camelCase to read a little easier? All lower case isn't great.

woogerboy21:
:+1:   This tested out great and worked as expected. Thanks for putting this in so quickly.

ctrlaltca:
> For new config options can we use camelCase to read a little easier? All lower case isn't great.

Qsettings should use case insensitive keys for INI files, so we should be able to cameCase all of them with no problem.

ctrlaltca:
Merging this. I'll change the setting keys to camelCase in a new PR

Issue #1671, opened 2015-11-10
Deck Editor should remember changes to column order


Psithief:
The deck editor layout doesn't persist through different resolutions, and some items don't persist at all.

Many examples:
- Resizing Cockatrice to a different size and back resets the layout. Maximising a smaller window leads to this waste of space:  
  ![editor_default](https://cloud.githubusercontent.com/assets/1124846/11054899/262611dc-87ac-11e5-9e05-2919d3c4759c.png)  Related: #1535
- The horizontal divider between the picture and the rules text (`Card Info -> Both`) doesn't save at all. (mentioned in #1359)
  - It resets when closing/opening Cockatrice
  - It doesn't persist when opening a new Deck Editor
  - It doesn't even persist when switching to `Card Info -> Image`/`Card Info -> Description` and back.  
    ![editor_divider](https://cloud.githubusercontent.com/assets/1124846/11054851/c2109ed8-87ab-11e5-9d8b-0ae1007eb552.png)
- Order, sort and widths of columns don't persist, for both card search and deck list.

ctrlaltca:
About the first issue, what's the best behavior?
 * resizing the 3 columns (carddb | info+filters | decklist ) proportionally, instead of just expanding the first column?
 * resizing the table columns of the carddb?

Psithief:
Perhaps the solution is a mix of better defaults, better interpolation and keeping track of a layout for different widths.

## Better defaults
The font size doesn't change, so the card db columns only need a certain width each.
On mine (Win7) @ 1920px screen size:
* **Name** is at most 163px (in black-border)
* **Sets** will take up as must room as it can (e.g. basics lands). If we pick something like 5 sets, that's 150px
* **Mana Cost** is at most 103px (in black-border: Reaper King's 2/W2/U2/B2/R2/G)
* **P/T** is at most 49px (1+*/1+*)
* **Card type** is at most 242px (in black-border)

Column width Sub-total: 707px  
Default card db width: 1,320px

The Card db should never expand past that smaller width unless the user drags it.

## Better interpolation

The card db columns don't react when their container gets narrower, but they should. And not linearly, because the useful width of a column is different depending on the column.
*E.g.:*
* **Name** ~140px
* **Sets** 65px is enough to show 2 sets (most of the time)
* **Mana cost** 52px is enough for most cards (though the column title is cut off)
* **P/T** 26px shows single digit power / toughness stats
[...]

PR #1877, opened 2016-03-19
Prevent PTColumn from being the last column shown.

Psithief:
P/T column is really narrow. It should go before Card Type, so Card Type can expand into the remaining space.

Improves #1671, assuming I got this correct.

Untested!

### DIFF
--- a/cockatrice/src/carddatabasemodel.h
+++ b/cockatrice/src/carddatabasemodel.h
@@ -12,7 +12,7 @@ class FilterTree;
 class CardDatabaseModel : public QAbstractListModel {
     Q_OBJECT
 public:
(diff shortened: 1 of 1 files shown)


Daenyth:
I don't think this will affect the visuals at all, but we can check

tooomm:
before
old-pt

after
new-pt


So it does change the ordering... but the first picture has the new change from ctrlaltca already in with a bigger name column. So not the best visual comparison. At least it reorders. :)
Someone else can make a comparison with against a normal build?

Daenyth:
Bump - what needs to still be done with this? Is it good to go?

ZeldaZach:
Seems good to me

ctrlaltca:
:+1:

PR #1879, opened 2016-03-20
User suspicion

woogerboy21:
So *hopefully* this is the beginning / partial implementation of issue #170 

First things first.  What this PR currently does.  The server now calculates a user suspicion level upon login and sends the information in the user_info pb container which in turn will change the background color of the userlist row if the suspicion level ever exceed's pre-defined values.  This functionality is turned on and off by lock/unlocking the admin tools in client.

Now as for what it doesn't do that I would like to see:
- When the admin/moderator hits the lock/unlock I would like to see the user list in rooms refreshed
- Better calculations for suspicion level (just threw a few in to start with)
- Fine tune ranges for background color adjustments

What thoughts do others have?  Keep in mind this is very very very rough.  I really had intended on implementing the basic's for the functionality to work and then later fine tune the calculations part.

Attached are a couple of pics showing the background color changes.
blob2
blob1

### DIFF
--- a/cockatrice/src/userlist.cpp
+++ b/cockatrice/src/userlist.cpp
@@ -352,10 +352,10 @@ void UserList::processUserInfo(const ServerInfo_User &user, bool online)
         item = new UserListTWI(user);
         users.insert(userName, item);
 
-		if (user.user_level() > 3) {  //NEED TO CORRECT THIS TO EITHER LOCATE IF YOUR ARE ADMIN/MOD AND SHOW OR BLOCK SUSPICION LEVEL FROM BEING RETURNED IF A REG USER REQUESTS USER DATA
+		if (!tabSupervisor->getAdminLocked()) {
 			if (user.suspicion() > 10) { // apply warning level due to user suspicion level being elevated
 				item->setBackground(0, *(new QBrush(Qt::yellow, Qt::Dense6Pattern)));
-				item->setBackground(1, *(new QBrush(Qt::yellow , Qt::Dense6Pattern)));
+				item->setBackground(1, *(new QBrush(Qt::yellow, Qt::Dense6Pattern)));
 				item->setBackground(2, *(new QBrush(Qt::yellow, Qt::Dense6Pattern)));
 			}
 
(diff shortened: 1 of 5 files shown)


woogerboy21:
Just an FYI.  I also plan on pulling the suspicion level color scale to the server side and only passing the color that the background should be to the client.  This way a server operator can set the scale in the servers ini and could be adjusted on the fly.

Daenyth:
So, I've been thinking about this and I think I actually want to hold off. I think the feature complexity vs utility doesn't justify it. I'd like to keep the protocol focused on gameplay at the moment, and on general utility functions.

So with that in mind, I'm trying to think about how we could add some of the utility of this without the protocol requirements

Daenyth:
What would it take to make this a client side modification? I think if we can improve the ban listing protocol we can compute this fully locally and not require it as a protocol item.

ctrlaltca:
Looking at the code, the number of queries needed to get the user suspicious level is enough to potentially provoke performance issues, since the calculation is triggered in a lot of places (at every user login, buddy/ignore list request, etc..)

woogerboy21:
So the two ways I can see it working is either client side calculations which limits the potential data points to pre existing data elements that are passed to the client or expanding the number of data elements in the form of additional queries potentially causing even further expansion of the protocol , more network usage and higher impact on the server through multiple database queries.

Or server side the calculations and shrink the protocol requirements .  One idea I had was more of a running calculation of a user suspicion level that extends the analytics table by one column and the users suspicion level is written at logout of the user.  That way a users suspicion level is calculated once if no value exists and then is updated in memory till logout.  Then written and saved till next login reducing the number of queries to only 1 and reducing the protocol expansion to a single bit of data.

But just my $0.02

ctrlaltca:
Yep, to save a cached copy of the value in the user table and updating it only when an event happens (the user is banned, etc) is probably the best way to overcome the possible performance problems, since you just remove the need for any additional query when looking up the user's info.

woogerboy21:
Server siding the code would also allow for easier expansion of data points
when either calculating newer users initial calculations or expanding what
calculations manipulate the suspicion level with out the need for protocol
creep or extra resource requirements on either client side or network side
that a server operator potentially has very little control over.

PR #1880, opened 2016-03-22
Remove "Ghost" Toolbar

leematos:
I'm working on a theme and realized we have a toolbar that used to be used for price, which we no longer display. I removed the code (and the unused price code action) as we weren't using either.


ctrlaltca:
The other code related to the price updater is currently commented out; maybe it's better to comment out these bits, too instead of removing them? @Daenyth

Daenyth:
We should delete the commented code. If needed we can extract it from git
later when appropriate but right now it's just clutter that adds no value

ctrlaltca:
Then, this is :+1:

Daenyth:
:shipit:

PR #1892, opened 2016-03-28
Add forgot password functionality

woogerboy21:
This PR adds the ability for users to reset there password in client if they forget.  The server admin has the ability to turn the functionality on or off completely.  And if on, the server can be set to verify that the user account email matches what the user inputs as well as check if the users last known client id matches the clientid that is being sent by the client requesting a password change.

If everything set to check for passes the password is updated.  If anything fails an "internal error" message is returned as a response so to not give any details away on a known user account.  The change password function also checks to see if the user making the request is being sent from an authenticated user and if the authenticated user is an admin or moderator it will permit the password update on the requested user account.  Keep in mind though the moderator/admin request functionality is not implemented anywhere but will be available for future functionality.


cockatrice/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ SET(cockatrice_SOURCES
     src/dlg_edit_password.cpp
     src/dlg_edit_tokens.cpp
     src/dlg_edit_user.cpp
+	src/dlg_forgotpassword.cpp
```
ctrlaltca: Currently all the code is indented using 4 spaces, this PR is using tabs instead.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -834,10 +836,12 @@ bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user, const
     if (!passwordQuery->next())
         return true;
 
-    const QString correctPassword = passwordQuery->value(0).toString();
-    if (correctPassword != PasswordHasher::computeHash(oldPassword, correctPassword.left(16)))
-        return true;
-
+	if (validateOldPasswordd) {
```
ctrlaltca: minor typo: "dd"

servatrice/src/serversocketinterface.cpp
```diff
@@ -974,6 +975,48 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
+	bool requireForgotPasswordClientID = settingsCache->value("users/forgotpasswordclientidrequired", true).toBool();
+
+	QString userName = QString::fromStdString(cmd.user_name());
+	QString password = QString::fromStdString(cmd.password());
+	QString userEmail = QString::fromStdString(cmd.email());
+	QString clientID = QString::fromStdString(cmd.clientid());
+
+	//if requesting user is a trusted user, change the password for the requested user
+	if (userInfo)
+		if (ServerInfo_User::IsModerator || ServerInfo_User::IsAdmin)
```
ctrlaltca: Bug: this will always evaluate to true.
Also, all admins are considered moderators, too, so no need to check for both.
Possibly correct code: ``if( userInfo->user_level() & ServerInfo_User::IsModerator )```

cockatrice/CMakeLists.txt
woogerboy21: Ah yes, I forgot to change that setting in VS when I reloaded it.  Ill roll through and change tab->spaces after all the reviewing is done before committing it so to not clutter up things during code review.

servatrice/src/serversocketinterface.cpp
woogerboy21: Are you indicating the userInfo check will evaluate true or the isMod/isAdmin line?

ctrlaltca:
```
message ServerInfo_User {
    enum UserLevelFlag {
        IsNothing = 0;
        IsUser = 1;
        IsRegistered = 2;
        IsModerator = 4;
        IsAdmin = 8;
    };
```
So, the line ```if (ServerInfo_User::IsModerator || ServerInfo_User::IsAdmin)``` equals to ``` if( 4 || 8 )```

common/pb/session_commands.proto
```diff
@@ -105,6 +106,16 @@ message Command_JoinRoom {
     optional uint32 room_id = 1;
 }
 
+message Command_ForgotPassword{
+	extend SessionCommand {
+        optional Command_ForgotPassword ext = 1021;
+    }
+	required string user_name = 1;
+	required string password = 2;
```
Daenyth: Is this for a desired new password? I'm not sure we should do that. We should probably have the server use a password reset process

servatrice/servatrice.ini.example
```diff
@@ -99,6 +99,15 @@ allowedpunctuation=_.-
 ; If a username can begin with punctuation defined in allowedpunctuation
 allowpunctuationprefix=false
 
+; Enable forgot password functionality
+enableForgotPassword=true
+
+; Validate users inputed email matches email on account before allowing forgot password reset
+forgotpasswordemailrequired=true
```
Daenyth: Can you `camelCase` this?

servatrice/servatrice.ini.example
```diff
@@ -99,6 +99,15 @@ allowedpunctuation=_.-
...
+
+; Validate users clientid matches clientid on account before allowing forgot password reset
+forgotpasswordclientidrequired=true
```
Daenyth: Is there a compelling reason to have this feature when we have the "require client id" global flag?

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -812,8 +812,10 @@ void Servatrice_DatabaseInterface::logMessage(const int senderId, const QString
     execSqlQuery(query);
 }
 
-bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword)
+bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword, const bool &validateOldPassword)
 {
+	//note: function returns false if password is successfully changed
```
Daenyth: That's really weird.

servatrice/src/serversocketinterface.cpp
```diff
@@ -974,6 +975,48 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
 
+Response::ResponseCode ServerSocketInterface::cmdForgotPassword(const Command_ForgotPassword &cmd)
+{
+
+	bool forgotPasswordEnabled = settingsCache->value("users/enableForgotPassword", true).toBool();
+	if (!forgotPasswordEnabled)
+		return Response::RespInternalError;
+
+	if (tooManyForgotPasswordAttempts(this->getAddress()))
+		return Response::RespInternalError;
```
Daenyth: This shouldn't be `RespInternalError` but a response specific to the action.

servatrice/src/serversocketinterface.cpp
```diff
@@ -974,6 +975,48 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
     return Response::RespOk;
 }
 
+Response::ResponseCode ServerSocketInterface::cmdForgotPassword(const Command_ForgotPassword &cmd)
+{
+
+	bool forgotPasswordEnabled = settingsCache->value("users/enableForgotPassword", true).toBool();
+	if (!forgotPasswordEnabled)
+		return Response::RespInternalError;
```
Daenyth: This should be a `RespForgotPasswordDisabled`

servatrice/servatrice.ini.example
ctrlaltca: when enabled, servatrice will check if the clientid of the cockatrice instance requesting the password reset matches the last-known clientid of the user the password reset was requested for.

servatrice/src/serversocketinterface.cpp
```diff
@@ -974,6 +975,48 @@ Response::ResponseCode ServerSocketInterface::cmdBanFromServer(const Command_Ban
...
+
+	bool requireForgotPasswordEmail = settingsCache->value("users/forgotpasswordemailrequired", true).toBool();;
+	bool requireForgotPasswordClientID = settingsCache->value("users/forgotpasswordclientidrequired", true).toBool();
+
+	QString userName = QString::fromStdString(cmd.user_name());
+	QString password = QString::fromStdString(cmd.password());
+	QString userEmail = QString::fromStdString(cmd.email());
+	QString clientID = QString::fromStdString(cmd.clientid());
+
+	//if requesting user is a trusted user, change the password for the requested user
```
Daenyth: I don't think "forgot password" should be what admin clients use to reset user passwords. They should use the "change password" function instead. An admin using this should just trigger the normal password reset process

Daenyth:
Please no one merge this until I get a chance to finish reviewing it.

servatrice/src/servatrice_database_interface.cpp
ctrlaltca: Yep, my fault in first trying to implement a POSIX-like return value (0 = no error, > 0 = errno), then not using it  and leaving such code there.

servatrice/src/serversocketinterface.cpp
ctrlaltca: :+1:

common/pb/session_commands.proto
woogerboy21: As in reset the password to something randomly generated then email the user?  Not sure that's any more secure and would be more troublesome for the user in the long run.    Whether we generate a new password or just apply what they send the validating of the user being who they say they are stays the same so the first part of the process wouldn't change resulting in just an annoyance for the end user to have to go to emails and try and copy/paste or reproduce some random string that depending on if there mail client is setup correctly may not display properly resulting in them contacting the server operating giving them less of a great user experience.

servatrice/src/serversocketinterface.cpp
woogerboy21: I could do this, but I was worried about returning to much information to a user that may not be the original user on the account.  I would hate for it to return something like "wrong email" or "incorrect client id" (just as an example of various area's in the function) and have the user that may be attempting to break into an account know what the exact problem might be.

Thoughts?

servatrice/src/serversocketinterface.cpp
woogerboy21: This does trigger the actual changeUserPassword DB functionality.  What other process are you referring to?  Take for example if the admin/mod wants to reset a different user account other than there own.  They would have to pass the other users information into a function that calls the DB functionality that changes the password.  That's basically what this does.  I have no problem removing the check for if there is a moderator/admin making the call and it ignoring the various other checks used to validate the user account is in fact the user making the call and work the functionality in some were else.  But logically it would be doing very similar work.

servatrice/servatrice.ini.example
woogerboy21: Sure.  I couldnt decide if it should be located under the users's section of the ini or if we should make a new section for it?

Thoughts?

servatrice/src/serversocketinterface.cpp
ctrlaltca: If i understood what @Daenyth means correctly, implementing the "change other users's password" functionality would better be implemented in the ```ServerSocketInterface::cmdAccountPassword``` method instead that in this one.

servatrice/servatrice.ini.example
tooomm: Many people tend to forget their passwords after not using the service for some time... so it might happen that they change their hardware in between.

They won't be able to reset the password and make the server send a new one to their linked mail address?

woogerboy21: Checked may be a better term to use and could resolve the confusion over the word "required".  As far as the idea on sending an email, the clientid should not change.  Even if the id does change I wouldn't want the user to be able to change there password in that case if I have the check for clientid as a required, otherwise i as an admin would just disable the check for client id option.

Plus there is always the elephant in the room about sending emails, what happens if I dont have an email server to use to send the emails?

servatrice/src/serversocketinterface.cpp
woogerboy21: So is the thought to scrap this function as a whole and blend it with the cmdAccountPassword function?

servatrice/src/serversocketinterface.cpp
woogerboy21: I would have to look at the current response codes but would something like RespDisabled be sufficient and reusable for other disabled functionality?

Daenyth:
Copying from gitter;

```
Daenyth 11:12
I'll do my best to get to it
but at a high level I have some issues with general structure relative to the reset process I was talking about before
I was hoping to discuss the workflow before we wrote code

woogerboy21 11:13
when you have time i wanted to run over the idea you had about hashing the host name in the clientid again

Daenyth 11:13
so that we can come up with the best case (with email on) and a degraded case with email off that still has good security
ah yeah we should really do that

woogerboy21 11:14
i was thinking about that the other day and was trying to rehash how having different client id's across different hosts would matter? I guess for things like forgot password?

Daenyth 11:14
If I connect to a malicious server, they have my client id. Then they can connect to your server spoofing my client id
and actions they take will look like I did them

woogerboy21 11:17
So the way the client id works is that the user's table is updated with the clientid on login. So a user would have to successfully login to the account in order to change the user table to associate the clientid to a username. Im trying to think through a scenario. So I as a user logs into a malicious server and now the malicious server has my clientid. They then could create an account and hard set the client id to match my username, log in with the second account and cause problems getting me banned? Is that the idea?
or change my password on another server , i guess that could be another scenario?
yeah that should be a pretty easy fix

Daenyth 11:31
And btw for the password reset thing I really think we should scope the initial feature set to only work for servers with email turned on
We can come up with an alternate method for servers without email later

woogerboy21 11:52
Just curious as to why?
i hate dealing with emails
if it doesnt send, its not received due to policies on email providers and if its received there is a chance the users email client is formating things incorrectly and they fail to be able to use what is sent to them

Daenyth 12:05
because you have no other way to reach the user
you cannot assume that the client sending you a reset request is the user
otherwise it will be totally open to abuse
you need some kind of out of band communication
I think the best approach for now is;
user resets are with email enabled only; moderators can force a reset
rather, moderators can directly set passwords

woogerboy21 12:08
So when the user account has a bogus email address on it and the users account is reset and the email goes no were then what?
cause that is a common practice
email !=reliable
my recommendation may be to add an email that is sent to the address on file saying the account password was reset .... maybe

Daenyth 12:09
You also don't send an email saying it was reset
that's a DOS attack against user logins
you send an email saying a reset was requested, provide the recipient a token they can use to finalize the reset
then it generates a new password and sends it back
if the email account is bogus, then there needs to be moderator intervention

woogerboy21 12:10
so let me understand your logic. you want to send an email to the user with a new password in it but you dont want to allow emails to be sent to users because its a DOS attack against logins?

Daenyth 12:10
Also, how is the email address bogus if account activation is enabled?
No
You're misunderstanding
you don't perform the reset upon initial request
If you do that, attacker can write a script that will continually send password resets for all users on the server
and then those users can't log in
hence DOS
the flow is:
Unknown untrusted user requests reset for account name foo or email foo@bar.com
Server sends either to email or looks up email by name and sends a short lived token
User who controls the email that was used to register the account uses the token to perform the reset
Server invalidates the token and generates a new password, emails it along with instructions on how to change it
the link I shared earlier had a flow chart

woogerboy21 12:14
Maybe your not understanding how the code logic works. The command is sent to the server from a client. the server checks to make sure various criteria is correct (email address on account matches, clientid on account matches and ive thought about adding last known ip address matches) and if all criteria matches then the account password is reset. No one can just write a script that makes requests and the server blindly resets passwords?

Daenyth 12:14
client id can be spoofed
you cannot assume that whoever sent you the request owns the account

woogerboy21 12:14
they all can be spoofed ... whats your point?

Daenyth 12:15
you need out of band confirmation or you are going to open a method to hijack accounts

woogerboy21 12:15
we dont collect enough data to avoid something like social engineering
lets take another approach ... say some one contacts you and says hey i need my password reset. what do you do?

Daenyth 12:16
I ask them what email they used and to email me from it to show they own it
That establishes that the user controls the email which was used to register (still doesn't prove it's the actual person)
but we can't get any better short of client side certs

woogerboy21 12:16
and when there account either has a) no email or b) a bogus email what happens?

Daenyth 12:17
If email activation is turned on, then they do have an email and it's not bogus
so that only applies when email is off
My suggested fallback is that the moderators should confirm
And at that point it's a judgement call based on how convincing the evidence is

woogerboy21 12:18
not true. there is an account with an email attached to it but doesnt mean its activated by using itts content

Daenyth 12:18
I'm not going to add something that's insecure by default - I'm OK with a less secure mode when it's what's configured by the admin
If the account isn't activated after a time limit it's deleted?

woogerboy21 12:18
and also, there is a method for accounts to be created with out email interaction

Daenyth 12:18
That's not good
Is it an actual problem in practice for you that you have users with bad/missing emails?
Let's fix that problem, we cannot add a feature that is easily abused to hijack accounts

woogerboy21 12:19
sure. people use things like 10minutemail.com all the time

Daenyth 12:19
That's a sign that the user is abusive anyway

woogerboy21 12:20
or they activate the account and then change the email address as they dont want it on file
or they are old accounts before emails were requireed

Daenyth 12:20
Then it would send to the new one
which should be activated

woogerboy21 12:20
so when i go in and update my account and put the email address as me@mail.com what do you do?

Daenyth 12:20
Mark it as inactive and send a token

woogerboy21 12:21
so you take an active users account that is logged in and gaming . mark there account as inactive so the next time they are f'd when trying to log in?

Daenyth 12:21
It sounds like there are workflow problems with email changing
It should send confirmation to the email on file BEFORE doing to chagne
change*
again, using a token for confirmation
```

woogerboy21:
My $0.02.  I don't like the idea of expecting the server owners/operators to have any additional infrastructure in place to operate a registered server .  In laymen terms, we should not expect the operators to own or use a mail server for functionality.

woogerboy21:
I would also like to point out that with places like 10minutemail.com (there are countless ones) getting an email address with no verification as to who you are or even the email address remaining for periods of time makes relying on the email process just as reliable as any of the other various forms of validating a user account is in fact the requesting users account.  There are methods to "spoof" or sidestep any of the methods.  We can only do our best as to validate the account and allow as many options to the server operator as to the method they want to use and trust.  Forcing an operator to do it your way or no way is not an ideal solution.

Daenyth:
It's important to distinguish between unreliable communication and spoofed communication. If it's unreliable, the correct user might not recover their account. If the data is spoofed by an attacker, an account can be hijacked. We want to minimize frustration from unreliable information and connections while maximizing things that block spoofing.

woogerboy21:
Email is unreliable.  Users simply do not receive notification emails for various reasons.  I have no intention on relying on email notifications nor expect anyone else to do so.  And to my original point, the server operator should have the choice to do so.  The functionality can be added in to do it, that's not a problem.  But forcing them to only use that method is a problem.

servatrice/src/servatrice_database_interface.cpp
Daenyth: We can clean it up later.

servatrice/src/serversocketinterface.cpp
Daenyth: I can't see how "too many reset attempts" as a response increases risk

servatrice/src/serversocketinterface.cpp
Daenyth: That's a good idea. I think the client can still figure out what is disabled because the response contains the command id

Daenyth:
I'd like to put this on hold until we fully flesh out the protocol portion. This is a critical API interaction and if we don't get it right up front we're going to have multiple years of headaches because of it.

So this isn't a "close as reject" but a "let's talk about the design more and then reopen"

PR #1894, opened 2016-03-29
Updated ClientID Functionality

woogerboy21:
ClientID is now generated on startup and stored in settings cache.  Then upon connect there is a new SrvClientID generated from the ClientID + servername being connected to.  This will allow for unique ID's to be generated per client and per server being connected to but not the same across all servers preventing a malicious server from collecting users client id's.

### DIFF
--- a/cockatrice/src/settingscache.cpp
+++ b/cockatrice/src/settingscache.cpp
@@ -2,6 +2,7 @@
 #include <QSettings>
 #include <QFile>
 #include <QApplication>
+#include <QCryptographicHash>
 
 #if QT_VERSION >= 0x050000
     #include <QStandardPaths>
@@ -612,4 +613,12 @@ void SettingsCache::setNotifyAboutUpdate(int _notifyaboutupdate)
 {
     notifyAboutUpdates = _notifyaboutupdate;
     settings->setValue("personal/updatenotification", notifyAboutUpdates);
+}
+
+QString SettingsCache::getSrvClientID(const QString _hostname)
+{
+    QString srvClientID = getClientID();
+    srvClientID += _hostname;
+    QString uniqueServerClientID = QCryptographicHash::hash(srvClientID.toUtf8(), QCryptographicHash::Sha1).toHex().right(15);
+    return uniqueServerClientID;
 }
--- a/cockatrice/src/settingscache.cpp
+++ b/cockatrice/src/settingscache.cpp
@@ -2,7 +2,6 @@
 #include <QSettings>
 #include <QFile>
 #include <QApplication>
-#include <QCryptographicHash>
 
 #if QT_VERSION >= 0x050000
     #include <QStandardPaths>
@@ -613,12 +612,4 @@ void SettingsCache::setNotifyAboutUpdate(int _notifyaboutupdate)
 {
     notifyAboutUpdates = _notifyaboutupdate;
     settings->setValue("personal/updatenotification", notifyAboutUpdates);
-}
-
-QString SettingsCache::getSrvClientID(const QString _hostname)
-{
-    QString srvClientID = getClientID();
-    srvClientID += _hostname;
-    QString uniqueServerClientID = QCryptographicHash::hash(srvClientID.toUtf8(), QCryptographicHash::Sha1).toHex().right(15);
-    return uniqueServerClientID;
 }
--- a/cockatrice/src/settingscache.h
+++ b/cockatrice/src/settingscache.h
@@ -176,7 +176,8 @@ public:
     bool getRememberGameSettings() const { return rememberGameSettings; }
     int getKeepAlive() const { return keepalive; }
     void setClientID(QString clientID);
-    QString getClientID() { return clientID; }    
+    QString getClientID() { return clientID; }   
+    QString getSrvClientID(const QString _hostname);
     ShortcutsSettings& shortcuts() const { return *shortcutsSettings; }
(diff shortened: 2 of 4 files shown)


woogerboy21:
A couple quick questions for everyone.  
1.) Should the function for generating the id now be in remoteclient or settingscache?
2.) Should I implement functionality server side to take and compare old id with new id and update the bans table with new id if they do not match?

cockatrice/src/settingscache.cpp
```diff
@@ -612,4 +613,12 @@ void SettingsCache::setNotifyAboutUpdate(int _notifyaboutupdate)
 {
     notifyAboutUpdates = _notifyaboutupdate;
     settings->setValue("personal/updatenotification", notifyAboutUpdates);
+}
+
+QString SettingsCache::getSrvClientID(const QString _hostname)
+{
+    QString srvClientID = getClientID();
+    srvClientID += _hostname;
+    QString uniqueServerClientID = QCryptographicHash::hash(srvClientID.toUtf8(), QCryptographicHash::Sha1).toHex().right(15);
```
Daenyth: Why do we take the `right` like this?

cockatrice/src/settingscache.h
```diff
@@ -176,7 +176,8 @@ class SettingsCache : public QObject {
     bool getRememberGameSettings() const { return rememberGameSettings; }
     int getKeepAlive() const { return keepalive; }
     void setClientID(QString clientID);
-    QString getClientID() { return clientID; }    
+    QString getClientID() { return clientID; }
```
Daenyth: This should become `private` to encourage correct usage

cockatrice/src/settingscache.cpp
woogerboy21: So one of the original concerning factors of the client-id functionality was that since the id was being generated off hardware devices on a persons machine that if we sent the hash to some were that some how some one may be able to reverse engineer the information and obtain the original senders hardware id.  So to mitigate the concern as much as possible we only took a portion of the generated hash and passed in hopes that the potential for some how reversing the hashed information back to some one's physical address list off there computer wouldnt be possible.

Also one thing to note, we have a varchar(15) in the db table so anything longer than that would just get truncated anyway.

cockatrice/src/settingscache.h
woogerboy21: I originally started out coding the unique server client id to be stored/called like the actual clientid value generated on startup so that it could be referenced later for who knows what ever reason down the road.  Later on while thinking about things it seems to me that the only time the function really should be called is prior to any remote client command which brought me to question 2 above.

If we want to go off that mentality the function should be moved into the remote client code base and made a private function.  Personally I like that idea so if you dont mind I will update it to that unless you have any thoughts on the matter?

woogerboy21:
@Daenyth Would you also like me to create the DB functionality to update the old client id's to new client id's in the ban's table.  It wouldn't be that hard to add but it would add extra overhead on resources server side on each login.  One advantage to having it in line in the code would be if some one does ever change there client id's later on (for whatever reason) the client id change would be corrected accordingly in other area's in the DB.

Daenyth:
I  think the best way to handle it is with a migration script actually. It's more work than it's worth to maintain migrating code forever when users will only upgrade once.

woogerboy21:
Im not certain how you would write any type of automated migration script for something like this.  Since its the actual data that is changing that relies on the clients input and there is no guarantee on when or even if  user will log in again how do you create some type of script around that?  Especially since the value's should be unique per client and is generated by the client themselves and told to the server?

That being said there may not be a migration script possible.

woogerboy21:
@Daenyth Moved the function to the remoteclient and privatized the function.  This is were it really should live as the custom clientid per server *should* only be called prior to connecting to any host.

ctrlaltca:
> That being said there may not be a migration script possible.

If i understand the new clientid alg correctly, here's some sql that could generate the new clientid starting from the old ones:
```
SELECT RIGHT(SHA1(CONCAT(clientid, 'cockatrice.woogerworks.com')), 15) FROM cockatrice_users WHERE clientid != '';
```
Of course if the user is connecting from a different server name (eg. xxx.info), the generated clientid will be different. This would be an easy way to circumvent a ban.
Maybe the server should send a nonce instead of us using its hostname?

woogerboy21:
@ctrlaltca Thanks for the SQL query.  Not sure I'm a fan of the nonce idea just yet but it would solve the issue of operators that have multiple name resolutions to the same server.  I doubt that is a big problem though and in a worst case scenario ban circumvention would most likely be temporary if the users behavior hasn't changed since they were banned.  So i'm not certain the development time on the nonce would be worth the benefits.  Maybe.  Others thoughts?

Daenyth:
I think the nonce idea isn't worth it yet. Let's do the simple version first and if it needs expansion do that later.

Migration in the current simpler model isn't strictly possible; the sql proposed above is roughly what I was going to suggest though. It should help provide some continuity between the versions, but definitely some bans will be lost. I think this isn't a big deal because if a user gets back on and acts up, they'll just get banned again.

woogerboy21:
@Daenyth so would you say this is safe to merge or do you want me to create a .sql migration file that contains the above query (with some notes) and add it to the PR?

ZeldaZach:
Looks fine to me, just tab->space it

ctrlaltca:
Code is fine and runs as expected :+1:

PR #1900, opened 2016-04-02
Save deck editor's database view columns state

### LINKED ISSUE #1897: Name column way too big, can't resize normally
f217551

deck_editor_name_column

The name column automatically expands to ridiculous size on 1920x1080 and the other columns are just one size. I can't modify the name column width, except by dragging the right-hand edge of the last column to the right (out of the table). That's a very obscure solution, one I don't think most users will understand.

The name text is way too far from the other text in the grid, so it's hard to read. It almost makes more sense to just have the name column as the far-right one now, bringing us back to the original problem.

I don't know the best solution, but the current one isn't a positive change for legibility.

In a portrait view, all the other columns are suddenly hidden. It becomes even more difficult to understand how to shrink the columns with name locked when the user has to scroll horizontally. The minimum name length in this situation is still much too wide.

deck_editor_name_column_portrait

ctrlaltca:
This PR add the save/restore of the state of the database views's columns in the deck editor.
You can resize the column, reorder them and choose which column is used for sorting the view.
These choices will be saved to ```layouts.ini``` and restored to the deck editor.

Hopefully fixes #1897

### DIFF
--- a/cockatrice/src/settings/layoutssettings.cpp
+++ b/cockatrice/src/settings/layoutssettings.cpp
@@ -58,6 +58,16 @@ void LayoutsSettings::setDeckEditorFilterSize(const QSize &value)
     setValue(value,"layouts/deckEditor_FilterSize");
 }
 
+const QByteArray LayoutsSettings::getDeckEditorDbHeaderState()
+{
+    return getValue("layouts/deckEditorDbHeader_state").toByteArray();
+}
+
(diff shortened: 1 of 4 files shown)


ZeldaZach:
Works for me. Nice job :shipit:

ctrlaltca:
@Psithief can you please test this one before it gets merged?

Daenyth:
Code looks fine.

PR #1901, opened 2016-04-02
CardDb: be permissive when loading xmls

ctrlaltca:
Background: after the last version was released, the new "reverse-related" tags has been added to the tokens file downloaded by default by oracle.
Users with an older oracle version were no more able to load tokens, since the card database will consider the unknown tag as an error and stop parsing the file. As a result only the first "Angel" token was present in the card database.

This PR makes such an error not-breaking, so that the card database will try to load anyway as much as it can.

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -463,6 +463,10 @@ void CardDatabase::loadSetsFromXml(QXmlStreamReader &xml)
                     setType = xml.readElementText();
                 else if (xml.name() == "releasedate")
                     releaseDate = QDate::fromString(xml.readElementText(), Qt::ISODate);
+                else if (xml.name() != "") {
+                    qDebug() << "[XMLReader] Unknown set property" << xml.name() << ", trying to continue anyway";
+                    xml.skipCurrentElement();
+                }
             }
 
             CardSet * newSet = getSet(shortName);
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Edited my token file with a new fake field. Doesn't seem to work as expected.

```
<card>
    <name>Angel </name>
    <set picURL="http://cdn.staticneo.com/w/mtg/b/bb/Angel3.jpg">APC</set>
    <color>B</color>
    <manacost></manacost>
    <type>Token Creature — Angel</type>
    <pt>3/3</pt>
    <tablerow>2</tablerow>
    <text>Flying</text>
    <token>1</token>
    <reverse-related>Haunted Angel</reverse-related>
	<dang>Yes</dang>
</card>
```
```
[XMLReader] Unknown card property "dang" , trying to continue anyway
[XMLReader] Unknown item "card" , trying to continue anyway
[XMLReader] Unknown item "name" , trying to continue anyway
```

screenshot 2016-04-02 17 56 17

ctrlaltca:
@ZeldaZach sorry, i pushed the wrong commits. Now it should be working.

ZeldaZach:
Tested and works perfectly. Nice job! :shipit: 

```
[XMLReader] Unknown card property "ding" , trying to continue anyway
[XMLReader] Unknown card property "walla" , trying to continue anyway
```

PR #1914, opened 2016-04-07
added filename to ISSUE_TEMPLATE.md

tooomm:
added `Alternatively the complete filename of the downloaded installer is also fine*


Daenyth:
How about `If you can't access this menu for any reason, include the full filename of the installer you used` ?

ctrlaltca:
Can you please squash this down to a single commit?

tooomm:
see #1926

PR #1921, opened 2016-04-08
Implemented !setnumber!

### LINKED ISSUE #1741: Custom Download URLs - !setnr
For some image repositories, the index of a card in its set is part of the static image URL.

This, for example, is the case with `magiccards.info`. The URL "http://magiccards.info/scans/en/v13/12.jpg" links to the card `"Gilded Lotus"` in the Set `"From the Vault: Twenty"` and has the set index `12`.

A token for that index is not yet implemented. Something like `"!setnr"` or `"!setindex"` could do well.

ctrlaltca:
With this PR, the new !setnumber! code can be used in picture urls.

Tested using ```http://magiccards.info/scans/en/!setcode_lower!/!setnumber!.jpg```

fix #1741

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -172,7 +172,8 @@ CardInfo::CardInfo(const QString &_name,
                    int _tableRow,
                    const SetList &_sets,
                    const QStringMap &_customPicURLs,
-                   MuidMap _muIds
+                   MuidMap _muIds,
+                   QStringMap _setNumbers
                    )
     : name(_name),
       isToken(_isToken),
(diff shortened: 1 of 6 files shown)


tooomm:
While documenting this into the wiki...
https://github.com/Cockatrice/Cockatrice/wiki/Custom-Download-URLs

I don't like the naming at all... it's very confusing!
Why don't we use `cardnumber`?
All placeholders with "set" are related to actual sets... this isn't. It's not a number of a set, it's a number of a card in a set.

Daenyth:
Yeah that's a reasonable point.

tooomm:
So how to proceed with this?

Daenyth:
Make a new ticket?

ZeldaZach:
The only issue I have with changing it is people will start having issues if they were using this and I'd rather leave it as is. The wiki clearly documents what the tags do and how to effectively use them

PR #1931, opened 2016-04-11
Updated Dockerfile and doc

ctrlaltca:
* Changed some libraries
 * added some text to the doc


Daenyth:
Seems fine. Might be good to link to the docker setup portion on the wiki from the readme

Daenyth:
:shipit:

PR #1949, opened 2016-04-16
changed issue link + new user explanation

tooomm:
- included `How to create a github ticket`
- changed issue link from immediate creation to issue list, so people hopefully check for duplicates fist

Or is that indirect `new issue` link too complicated? :)


Daenyth:
:shipit:

PR #1955, opened 2016-04-19
Make client gender neutral

leematos:
As a response to #1638, this pull request removes gender from the client to simplify translation. For protocol purposes, gender is set to -1 which is "GenderUnknown".

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -25,49 +25,27 @@ QString MessageLogWidget::cardLink(const QString &cardName) const
     return QString("<i><a href=\"card://%1\">%2</a></i>").arg(cardName).arg(cardName);
 }
 
-bool MessageLogWidget::isFemale(Player *player) const
-{
-    return player->getUserInfo()->gender() == ServerInfo_User::Female;
-}
-
-bool MessageLogWidget::userIsFemale() const
-{
-    return (tabSupervisor && tabSupervisor->getUserInfo() && (tabSupervisor->getUserInfo()->gender() & ServerInfo_User::Female));
-}
-
 void MessageLogWidget::logGameJoined(int gameId)
 {
     soundEngine->playSound("player_join");
-    if (userIsFemale())
-        appendHtmlServerMessage(tr("You have joined game #%1.", "female").arg("<font color=\"blue\">"+ QString::number(gameId) + "</font>"));
-    else
-        appendHtmlServerMessage(tr("You have joined game #%1.", "male").arg("<font color=\"blue\">"+ QString::number(gameId) + "</font>"));
+        appendHtmlServerMessage(tr("You have joined game #%1.").arg("<font color=\"blue\">"+ QString::number(gameId) + "</font>"));
 }
 
 void MessageLogWidget::logReplayStarted(int gameId)
 {
-    if (userIsFemale())
-        appendHtmlServerMessage(tr("You are watching a replay of game #%1.", "female").arg(gameId));
-    else
-        appendHtmlServerMessage(tr("You are watching a replay of game #%1.", "male").arg(gameId));
+        appendHtmlServerMessage(tr("You are watching a replay of game #%1.").arg(gameId));
 }
 
 void MessageLogWidget::logJoin(Player *player)
 {
     soundEngine->playSound("player_join");
-    if (isFemale(player))
-        appendHtmlServerMessage(tr("%1 has joined the game.", "female").arg(sanitizeHtml(player->getName())));
-    else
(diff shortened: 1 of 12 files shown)


ctrlaltca:
Code looks fine and works as intended 👍

tooomm:
- I found several leftover "male" statements which used to be male/female before:
For example these (there are many many more):
https://github.com/Cockatrice/Cockatrice/pull/1955/files#diff-ba223d707cfc4edecb1a981e30f71294R140
https://github.com/Cockatrice/Cockatrice/pull/1955/files#diff-ba223d707cfc4edecb1a981e30f71294R152
https://github.com/Cockatrice/Cockatrice/pull/1955/files#diff-ba223d707cfc4edecb1a981e30f71294R331
https://github.com/Cockatrice/Cockatrice/pull/1955/files#diff-ba223d707cfc4edecb1a981e30f71294R372
Turn them all to "neutral" or some real context help?

- Sometimes we use "nomative"
https://github.com/Cockatrice/Cockatrice/pull/1955/files#diff-572eb2766e70107b27267165f1474c31R52
while most of the time we have the string context there
https://github.com/Cockatrice/Cockatrice/pull/1955/files#diff-572eb2766e70107b27267165f1474c31R69
Should be uniform and supportive to the translator.


I checked transifex and the appended info like `"reveal zone, male owner"` gets passed to it, so it's a context help for the translators!
It'll display like this in the details of a string: `Context:    CardZone:reveal zone, male owner`
Same should be for "nominative" and "male". We need to make the changes with that in mind!

- I wasn't sure about some variable cases... are they still all needed?

Daenyth:
> Turn them all to "neutral" or some real context help?

I think we should make a general guideline for translators now, they should be using neutral everywhere.

leematos:
@tooomm Good catch! I can get those all uniform. @Daenyth @ctrlaltca, do we prefer:

neutral or normative in the string messages for translators?

ctrlaltca:
I'd remove all the translation hints (aka the second parameter of the tr() call), and add a big notice in the translation guidelines.

Daenyth:
@leematos 
In the line context we shouldn't call out neutral, neutral should be the assumption.

by "normative" do you mean heteronormative in contrast to gender neutral, or do you mean nominative? The latter should absolutely stay, it's about grammatical case in case. In the case of the former we should tell translators to prefer neutral when it's available ("they" for English, "hen" for Swedish, etc), but that if there's no acceptable method for it in their language to just use whatever makes sense.

leematos:
@Daenyth Doh! reading comments on a phone while not fully awake. (Don't do it.) Yeah, I meant Nominative. Should I be replacing all of what I currently wrote as  "neutral" (the second param of the translation tags) to nominative or go with @ctrlaltca's suggestion of removing the second param of the translation tags?

tooomm:
Yes, I think removing the "normative" labels (and "male"!) is the best way. We don't help translators if they see a "neutral" hint in the each and every string detail, we should have that in our [FAQ for Translators](https://github.com/Cockatrice/Cockatrice/wiki/Translation-FAQ).
Well, many translators don't read the FAQ maybe... but still.

What we should have though, are good hints in general that really support translators while contributing their time and knowledge!
We could either start right in this PR or make it a different issue/pr.

leematos:
@tooomm I went in and followed @ctrlaltca's advice, it made sense in the cases where it just said "male" to drop it. The cases where it says nominative, I left, but I'm happy to remove too if ya'll agree they are useless. Anything else needed for this PR to get merged?

cockatrice/src/messagelogwidget.cpp
```diff
@@ -359,69 +306,40 @@ void MessageLogWidget::logMulligan(Player *player, int number)
         return;
 
     if (number > -1) {
-        if (isFemale(player))
-            appendHtmlServerMessage(tr("%1 takes a mulligan to %n.", "female", number).arg(sanitizeHtml(player->getName())));
-        else
-            appendHtmlServerMessage(tr("%1 takes a mulligan to %n.", "male", number).arg(sanitizeHtml(player->getName())));
+        appendHtmlServerMessage(tr("%1 takes a mulligan to %n.", number).arg(sanitizeHtml(player->getName())));
```
ctrlaltca: This line borks the compilation. Change it to 
```
appendHtmlServerMessage(tr("%1 takes a mulligan to %n.", nullptr, number).arg(sanitizeHtml(player->getName())));
```

leematos: I actually, switched it slightly differently, which seemed to work and avoid the need for a null pointer.

ctrlaltca: That's fine, too 👍

PR #1969, opened 2016-05-04
remove numbers from the dice roll translatable strings

### LINKED ISSUE #1833: remove numbers from the dice roll translatable strings
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/messagelogwidget.cpp#L148

To prevent mistakes or make translators mix things up...
Basic dice functionality should be there in any case, and properly working!

tooomm:
fix #1833

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -145,7 +145,7 @@ void MessageLogWidget::logShuffle(Player *player, CardZone *zone)
 
 void MessageLogWidget::logRollDie(Player *player, int sides, int roll)
 {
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Looks good to me, will merge on completion

ctrlaltca:
Just in time to break translations for every language 😍

Daenyth:
I would have said leave this until after the release because of that. Too slow I suppose.

tooomm:
Well, I was not aware that the fix is that simple (and more or less exactly the code that I already provided in the issue over two month ago) ;P

PR #1974, opened 2016-05-08
fix updater from bintray api issue

### LINKED ISSUE #1972: BinTray Auto Updating Issues
Seems like BinTray's API may have changed as it no longer includes a "date" field (as seen [here]( https://github.com/Cockatrice/Cockatrice/blob/cbea43267ef32135d34099efbdf699703228547a/cockatrice/src/update_checker.cpp#L76))

I was successfully able to get it to work by changing "date" => "created" as they're very similar and contain the same type of dataset.

Once doing this, the program seems to work (on OSX) correctly. An issue I came across, however, is that [this](https://github.com/Cockatrice/Cockatrice/blob/cbea43267ef32135d34099efbdf699703228547a/cockatrice/src/dlg_update.cpp#L167) line on a mac doesn't open up the installer. I couldn't figure out how to do that, so I'm not opening a PR in favor of someone else looking into this.

This is a critical bug that needs to be addressed.

cc: @ctrlaltca

ZeldaZach:
Fix #1972 

This fixes the auto updater download issue. The issue I came across with mac last night was a weird glitch that is no longer affecting me so I'm going forward with this. (The file dmg attempted to load last night but didn't pop up; starts to popup now so ignoring the issue)

Will test on Windows 7 soon.

### DIFF
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -85,6 +85,7 @@ void DlgUpdate::gotoDownloadPage() {
 
 void DlgUpdate::downloadUpdate() {
     setLabel("Downloading update...");
+    enableOkButton(false);
     enableUpdateButton(false);
     uDownloader->beginDownload(updateUrl);
 }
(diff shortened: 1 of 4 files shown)


tooomm:
It's always helpful to explain what you actually changed, and why. ;) :+1: 
<br>
>The issue I came across with mac last night was a weird glitch that is no longer affecting me so I'm going forward with this.

You mean the "installer won't open and Cockatrice crashes after downloading" issue?

PR #1197, opened 2015-07-03
Update project to C++11 compilation.

Daenyth:
Fix #127

PR #1979, opened 2016-05-11
Drop Qt4 and other dependencies

### LINKED ISSUE #1809: Drop Qt4
Qt4 support by Qt ended on december 2014, but was extended for another year to get a 4.8.7 version out (http://blog.qt.io/blog/2014/11/27/qt-4-8-x-support-to-be-extended-for-another-year/).
All the windows packages built by Appveyor and the Osx, Linux "release" packages built by TravisCI are already qt5 only.
Removing qt4 we could
 * clean up a lot of dead code in ```#if QT_VERSION < 0x050000```
 * remove some dirty code (like servarice's own command line parser, see #1405)
 * remove the dependency on libgcrypt
 * (eventually) remove out bundled qt-json library and move to use qt5's own library

ctrlaltca:
Drop Qt4 usage and all the dependencies not needed anymore:
 * libgcrypt
 * qtmobility

Fix #1809


Daenyth:
We should update the wiki with this too.

tooomm:
I'll also update the io page tomorrow!

.github/CONTRIBUTING.md
```diff
@@ -4,16 +4,6 @@
 
 Cockatrice is written in C++-03, so do not use C++11 constructs such as `auto`.
```
tooomm: is this still true?
https://github.com/Cockatrice/Cockatrice/pull/1197

ctrlaltca: By now we enabled using a c++11 compiler, but there is no code using c++11 specific syntax yet.
The first thing i'd do after switching to Qt5 would be to refactor QObject::connect() calls in order to use use compile-time checking (instead of runtime checking) and lambdas where the code would benefit (https://wiki.qt.io/New_Signal_Slot_Syntax#Connecting_in_Qt5)

Daenyth: I think there's some `auto` stuff floating around already. We should definitely remove this line.

tooomm:
> I'll also update the io page after merge!

done

PR #1981, opened 2016-05-12
Create token dialog uses name as search

MarkyMarkMcDonald:
This pull request reuses the search functionality from the deck editor inside of the "Create token" dialog.

Here's a quick gif demo:

I'm pretty unfamiliar with C++ and Qt, so please be critical of anything I should change.

### DIFF
--- a/cockatrice/src/carddatabasemodel.cpp
+++ b/cockatrice/src/carddatabasemodel.cpp
@@ -237,8 +237,7 @@ TokenDisplayModel::TokenDisplayModel(QObject *parent)
 bool TokenDisplayModel::filterAcceptsRow(int sourceRow, const QModelIndex & /*sourceParent*/) const
 {
     CardInfo const *info = static_cast<CardDatabaseModel *>(sourceModel())->getCard(sourceRow);
-    
-    return info->getIsToken();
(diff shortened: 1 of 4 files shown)


Daenyth:
Code looks fine. We can test it and ideally get it into the next release. Thanks for the contribution!

@ZeldaZach do you have a little time to check it? @ctrlaltca want to review as well?

ctrlaltca:
Code looks fine 👍

tooomm:
I gave this a test. But I don't like it too much for some reasons.

Will come back with more details very soon, I'm busy at the moment. Idea is very nice though! :+1:

MarkyMarkMcDonald:
@tooomm Can you expand a bit? Do you think the search interaction needs more work, or the code? 

I could look at making a separate "search" input field above the token list for filtering and leave the name field as it was.

ctrlaltca:
I think the change is nice, maybe there's room for further improvement, but i see no point for not merging it right now. Maybe feeding the change to the few loyal testers, they'll came up with some ideas.

Daenyth:
@MarkyMarkMcDonald Thanks again for the contribution. It's very appreciated

PR #1987, opened 2016-05-22
fix broken collaborators link

tooomm:
Since we have around 80 people that helped with translations already, I exchanged the outdated and not used "translator hall of fame" wiki page for a always up-to-date "collaborators" list on transifex.
Matches more what we have for developers as well.

Little downside: you need to be logged in to see the list, but that's fine I guess


cockatrice/src/window_main.cpp
```diff
@@ -60,7 +60,7 @@
 #define GITHUB_PAGES_URL "https://cockatrice.github.io"
 #define GITHUB_CONTRIBUTORS_URL "https://github.com/Cockatrice/Cockatrice/graphs/contributors?type=c"
 #define GITHUB_CONTRIBUTE_URL "https://github.com/Cockatrice/Cockatrice#cockatrice"
-#define GITHUB_TRANSLATOR_RECOGNIZE_URL "https://github.com/Cockatrice/Cockatrice/wiki/Translators"
+#define TRANSIFEX_COLLABORATORS_URL "https://www.transifex.com/cockatrice/collaborators/?"
```
Daenyth: This brings me to a "manage contributors" screen if I'm logged in, or brings me to a login page if I'm logged out.

Does it show just a list if you're not a project admin?

cockatrice/src/window_main.cpp
```diff
@@ -286,7 +286,7 @@ void MainWindow::actAbout()
         + "<a href='" + GITHUB_CONTRIBUTORS_URL + "'>" + tr("Our Developers") + "</a><br>"
         + "<a href='" + GITHUB_CONTRIBUTE_URL + "'>" + tr("Help Develop!") + "</a><br><br>"
         + "<b>" + tr("Translators:") + "</b><br>"
-        + "<a href='" + GITHUB_TRANSLATOR_RECOGNIZE_URL + "'>" + tr("Recognition Page") + "</a><br>"
+        + "<a href='" + TRANSIFEX_COLLABORATORS_URL + "'>" + tr("Our Collaborators") + "</a><br>"
```
Daenyth: How about "Our Translators"?

cockatrice/src/window_main.cpp
tooomm: I'll check and come back to you after the weekend.

ZeldaZach:
This is forcing me to login, so this isn't too good for what we want...

tooomm:
What *do* we want?
I doubt it's what we have right now for several obvious reasons.
This is the only other possibility I can come up with which offers a complete solution that is always up to date and doesn't need maintenance...

**I think we are even better off just removing the link than having that "nonsense" wiki page.**

The wiki page only features two new translators since Dae took over btw, all others are from the early days of the app. Not to mention how much strings changed in between.
It's just heavily outdated and nobody is going to update it nor willing to maintain that page.

Both of the new translators mentioned there happened to contribute code here too. Coincidence? ;) What about the 80 other guys who might contributed way more translations wise?

Daenyth:
I emailed transifex asking about a public credit page. If they have a link we can use then I'll post it here, otherwise we should update the wiki page to be more generic and not need maintenance.

tooomm:
Cool. Let us know what they reply!

Daenyth:
> Hi Gavin, 
> 
> Thank you for contacting Transifex support and submitting your request. 
> 
> Regarding your case, first of all, I would like to clarify that pages that are related to the teams and the collaborators of your organization are not publicly accessible. 
> 
> So, the only way is to somehow export the information and share it. A suggestion I could make is to use the API call that is mentioned in our documentation guide here. So, in your case, the URL will be the following one: 
> 
> https://www.transifex.com/api/2/project/cockatrice/languages/
> (not publicly accessible as well)
> 
> So, you will get a list of all the contributors as follows: 
> 
> ```
> {
>         "coordinators": [
>             "Daenyth"
>         ], 
>         "language_code": "zh-Hans", 
>         "translators": [
>             "Pascal_Yu", 
>             "359763432"
>         ], 
>         "reviewers": []
>     }, 
>     {
>         "coordinators": [
>             "Daenyth"
>         ], 
>         "language_code": "cs", 
>         "translators": [
>             "Tomeno", 
>             "iameld3st", 
>             "LethalPigeon7", 
>             "irendie"
>         ], 
> ```
> 
> 
> The information you are looking for can also be found on the Manage Collaborators page. You can use the available filters but unfortunately, there is not any option to filter the results based on a specific project. 
> 
> Another solution would be to visit each target group page and find the translators that belong to it. But I think that this is a more time-consuming workaround. 
> 
> I hope that the above helps. 
> 
> Please feel free to reach out to me in case you need any further clarifications.

tooomm:
Ok, I updated the wording accordingly and reverted the link change.

So we go with that API call and some kind of script-generated wiki page?

Daenyth:
I think that's the best way to handle it eventually.

cockatrice/src/window_main.cpp
```diff
@@ -60,7 +60,7 @@
 #define GITHUB_PAGES_URL "https://cockatrice.github.io"
 #define GITHUB_CONTRIBUTORS_URL "https://github.com/Cockatrice/Cockatrice/graphs/contributors?type=c"
 #define GITHUB_CONTRIBUTE_URL "https://github.com/Cockatrice/Cockatrice#cockatrice"
-#define GITHUB_TRANSLATOR_RECOGNIZE_URL "https://github.com/Cockatrice/Cockatrice/wiki/Translators"
+#define GITHUB_TRANSIFEX_TRANSLATORS_URL "https://github.com/Cockatrice/Cockatrice/wiki/Translator-Hall-of-Fame"
```
Daenyth: 👎  changing the link url breaks the old versions already out there. If we do change the url we need to keep the old one but have it redirect to the new one.

tooomm: That's the original working link. We just have the [Translators Hall of Fame](https://github.com/Cockatrice/Cockatrice/wiki/Translator-Hall-of-Fame) to name them. It has always been like that. Zach created it like one year ago.

I'm not sure if I like that page name though...

Maybe if was broken all the time? Or Can't tell for sure.

Daenyth: The link in the old version of the code is broken. I guess no one ever
clicked it, or if they did, they didn't report anything.

ZeldaZach:
@tooomm Have you thought of another way to deal with this?

tooomm:
We could merge to finally get the link to the wiki working... (it used to be a dead link all the time)

Then ticket the api call and script based wiki page refresh/generation. It could be combined with a tx fetch/push translation script maybe.

Daenyth:
What's the status of this? Is this helpful still? If so we should get it cleaned up for merge. If not we should close it. If there's work to do before or after merging this, let's ticket it so we don't lose track

ZeldaZach:
Since users would have to login to see translation status and such for the "hall of fame" i'm fine with our current wiki on the subject.

Daenyth:
So close?

tooomm:
@ZeldaZach Did you even check what this is doing before closing? ;)

>What's the status of this? Is this helpful still? 

Yes! In the current state this **fixes a broken link** which has been in the client for like 2 years...
It also renames the link and the variable.

>Since users would have to login to see translation status and such for the "hall of fame" i'm fine with our current wiki on the subject.

This is not the case. I highly dislike our current wiki and still think it's pretty useless.
So I repeat my suggestion from the last post https://github.com/Cockatrice/Cockatrice/pull/1987#issuecomment-2575300270:
>We could merge to finally get the link to the wiki working... (it used to be a dead link all the time)
>
>Then ticket the api call and script based wiki page refresh/generation. It could be combined with a tx fetch/push translations script maybe.

PR #1991, opened 2016-05-25
Implementation of websockets in servatrice and test js client

### LINKED ISSUE #391: Websockets
Hi, me and some of my buddies are building a cockatrice web app! I've been in touch with Woogerworks on twitter making sure he'd be okay with it and he didn't object. We're now at a point where we need to start integrating with the server. Right now, because the server speaks via protocol buffers, the app cannot talk to it. We've come up with two options:

1) Build a dumb translation layer that will take JSON and turn it into the appropriate protocol buffer. This is proving to be more challenging than expected and also poses a problem with the number of clients that the server allows considering all the web clients would have to go through it.

2) Add websocket support to servatrice in order to allow the web clients direct access to the server, much like how the thick clients work today. This is our preferred solution.

In order to make 2 work, we would have to submit some PRs for servatrice to open itself to websocket connections and we want to make sure that's something you would be on board with. 

If you want to see how the web client has turned out so far, you clone it from https://github.com/thomasjmwb/bbbbbbbbbbbbbbb and run it locally. We hope to have this stood up on heroku in the very near future. Thanks so much for your time!

ctrlaltca:
This PR contains the first steps of opening up servatrice and its protocols to web clients.
The following changes are included for servatrice:
 * Servatrice now only works in threaded mode
 * Servatrice can listen on port 4748 for websockets connections
 * ServerSocketInterface is now an abstract class with 2 implementations: TcpServerSocketInterface and WebsocketServerSocketInterface to handle the different connection types.
 * The WebsocketServerSocketInterface directly exports the protobuf interfaces (no xml hack, no message buffering, etc..)

schermata 2016-05-25 alle 09 08 00

Included is also a working web client:
 * webclient.js is a library handling connection and message parsing, based on [protobuf.js](https://github.com/dcodeIO/protobuf.js).
 * index.html contains a test js client using webclient.js; the user interface is based on [jQuery](https://jquery.com/) and [jQueryUI](https://jqueryui.com/)

new

Note: "sql2" and "sql3" are two different rooms.

fix #391


tooomm:
Can we host the webclient on our webclient github repo and make it accessible as github page?
Or should that space be reserved for a new and nicer version?!

ctrlaltca:
> Can we host the webclient on our webclient github repo
The webclient needs to access the same protobuf files used by cockatrice and servatrice.
We could split the protobuf files to their own repo, servatrice to another repo, the webclient to yet another repo. I'm not sure this will make developing easier, anyway.

> and make it accessible as github page?
The webclient needs no serverside scripting suppor,t so it could be hosted on eg. github.io

> Or should that space be reserved for a new and nicer version?!
It surely needs improving before getting shipped to end users

tooomm:
I know we talked about splitting things up some time ago. Don't know what's the best way.

If we split every part up, there still needs to be cross-access between repos to allow Cockatrice and Servatrice to get the protobuff stuff from its own repo, if I understand you correctly?

But then the Cockatrice/WebClient should be able to access files in Cockatrice/Cockatrice repo right now as well?

woogerboy21:
> The webclient needs to access the same protobuf files used by cockatrice and servatrice.

Are there any particular instructions as to how to get the web client installed and working any were?   Im not personally familiar with java script and not sure what would be required to allow access to the protobuf files.  Its probably easier than what Im making it out to be.

ctrlaltca:
> If we split every part up, there still needs to be cross-access between repos to allow Cockatrice and Servatrice to get the protobuff stuff from its own repo, if I understand you correctly?

Yes. The protobuf files contains the "language" that is used to comunicate between servatrice and cockatrice/webclient, and they need to access those files to be sure they're able to speak the same "language".

> But then the Cockatrice/WebClient should be able to access files in Cockatrice/Cockatrice repo right now as well?

We could use [Git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules), but they add some complications in managing the repository and merge changes.
Eg. if we want to make a  change to the protocol, we'd need to open 4 different PRs on the 4 projects and ensure they get merged at the very same time, or we'll end up with a client that's not able to speak with the server or viceversa.
Splitting the project could be useful if we have a lot of developers and need fine grained permissions, but as long as it's just a few of us, imho it's just a complication.

If we want to better discuss this idea, better create another ticket.

ctrlaltca:
> Are there any particular instructions as to how to get the web client installed and working any were? Im not personally familiar with java script and not sure what would be required to allow access to the protobuf files. Its probably easier than what Im making it out to be.

The webclient directory contains a "pb" symlink to the original folder. You just need to ensure that the webclient will find the protobuf files on that path (eg. you can remove the symlink and copy the original directory, if you want).
About installation, you need a web server (any can do it: apache, nginx, ...); just copy the webclient directory in a browsable directory and you're done.

tooomm:
Do we want to stick to the naming pattern with e.g. `Webatrice` over `Webclient`?

Daenyth:
I don't particularly care

woogerboy21:
This is cool 👍 
The only thing (and correct me if I'm wrong @ctrlaltca  )  being only able to run in a single thread/thread pool scalability on the websocket side of the server wont be that great.  Though I still think this is an awesome addition.

ctrlaltca:
> The only thing (and correct me if I'm wrong @ctrlaltca ) being only able to run in a single thread/thread pool scalability on the websocket side of the server wont be that great.

Right.

> Is it possible to set a limit on the number of clients the websocket interface can allow in at a single time and make it a numerical value to define in the servers ini file to help server operators?

This is a nice idea, i'll see if i can split max_users_total in 2 different options for tcp and websocket connections.

servatrice/src/servatrice.cpp
```diff
@@ -98,14 +88,69 @@ void Servatrice_GameServer::incomingConnection(qintptr socketDescriptor)
...
+    for (int i = 0; i < connectionPools.size(); ++i) {
+        const int clientCount = connectionPools[i]->getClientCount();
+        if ((poolIndex == -1) || (clientCount < minClientCount)) {
+            minClientCount = clientCount;
+            poolIndex = i;
+        }
+        debugStr.append(QString::number(clientCount));
+    }
+    qDebug() << "Websocket pool utilisation:" << debugStr;
+    Servatrice_ConnectionPool *pool = connectionPools[poolIndex];
```
Daenyth: Can we split out that into a helper method so we can `Servatrice_ConnectionPool *pool = findLeastUsedConnectionPool()` ?

Daenyth:
What do you think about making Qt5.4 or 5.5 the minimum required version, you'd be able to get rid of some conditional compilation.

woogerboy21:
My vote would to **not** be to require something like the most recent version of Qt.

ctrlaltca:
Requested changes implemented, sorry for the delay

ctrlaltca:
Squashed down to a single commit

woogerboy21:
So is there anything currently holding this up?

ctrlaltca:
Imho this is ready for getting in.
The server part is fine, while the web client is still rough but working.

ctrlaltca:
@Daenyth are you ok with merging this or do you prefer to review it again?

Daenyth:
Let me double check but if I'm taking forever we can go ahead and push it since we can always edit on it after

TMiguelT:
Fantastic! The possibility of one (or many!) web clients is really exciting

woogerboy21:
Awesome.  I can finally get to updating the server on woogerwork's host :)

PR #1995, opened 2016-05-28
new tx logo

tooomm:
overall redesign of the page and their logo and colors
see https://www.transifex.com/blog/2016/new-transifex/


aeosynth:
seems good :+1:

PR #1999, opened 2016-05-31
Prevent no games from showing

### LINKED ISSUE #1476: After updating Cockatrice, have to click Clear Filter to see any matches
May be a mismatch between filter logic from a previous version - I assumed it was only me but another user experienced the same issue.

ZeldaZach:
Fix #1476 

When you first launch the client (the very first time) it will try to convert "legacy settings" you may have saved within the application in the past to the preferences file. If it didn't detect a value, -1 was inserted. This will prevent -1 from being inserted for max players and instead insert 99. By doing this, the game will show all games of all types once the client is launched for the first time.


Daenyth:
👍

ctrlaltca:
nice catch, :ship: :it:

tooomm:
Awesome fix!

PR #2000, opened 2016-06-01
Add set names to dialog

### LINKED ISSUE #1946: "new sets found" dialog addition - list newly found sets
I like how the [required featues](https://github.com/Cockatrice/Cockatrice/pull/1439#issuecomment-136877493) dialog shows features in a list right in the dialog box if you click on `details`.

This brought me to the idea that **it would be very helpful to have some kind of listing of all newly found sets in the `new sets found` dialog box as well.** At the moment you just see how many sets got found, not which ones. You need to go into the `edit sets` menu to see which new ones are available.

The required features dialog is just an positive example. It might be even better to have some kind of scrollable table with long name, short name and release date within the `new sets found` dialog.

ZeldaZach:
Fix #1946 

When you have new sets in the card database, you will now also be informed of their set codes in the popup message.

screenshot 2016-05-31 20 26 59

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -105,6 +105,18 @@ int SetList::getUnknownSetsNum()
     return num;
 }
 
+QStringList SetList::getUnknownSetsNames()
+{
+    QStringList sets = QStringList();
+    for (int i = 0; i < size(); ++i)
+    {
+        CardSet *set = at(i);
+        if(!set->getIsKnown())
+            sets << set->getShortName();
+    }
+    return sets;
+}
+
 void SetList::enableAllUnknown()
 {
     for (int i = 0; i < size(); ++i)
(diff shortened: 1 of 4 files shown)


ctrlaltca:
Nice addition, i'm just thinking if we should limit the max number of set names, to prevent an enormous message box from popping up in some edge cases.

ZeldaZach:
I haven't considered the corner cases, so if somebody wants to test it and give me a number to limit it to I can do that

tooomm:
Having a cap is not ideal.
Can we somehow reuse the code from #1439 to get a scrollable list, so that we don't need to have a limit and can provide complete information?


At best there would be a scrollable table with set codes, full names and release dates like in the `Edit sets` dialog - but that can be another ticket and is optional.

Daenyth:
Reusing the view style of the edit sets dialog would be nice

Daenyth:
We can incorporate the changes from feedback in another commit.

PR #2002, opened 2016-06-04
Check for x86/x86_64 update version on windows.

### LINKED ISSUE #1977: auto-updater - determine if the os is win32 or win64
The auto-updater should determine if the os is win32 or win64 to download correct file from bintray and doesn't force users on old 32bit windows to download (and break!) their Cockatrice installations with 64bit binaries.
The important information here really is the OS, not which version of Cockatrice is installed/running.

Interesting reading:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724340.aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/ms684139%28v=vs.85%29.aspx
Source: http://stackoverflow.com/questions/10529772/qt-get-system-32bit-or-64-bit-info

Daenyth:
Fix #1977


tooomm:
Why do we use "x86" in the 64bit as well and not just "x64"?
I now that it's used like that sometimes... It might add some confusion though.
Anyway, if it's technically the right/better way it's fine!

Why don't we check for dmg in osx build as well?
Because we do the other way around for win... (exe)

Daenyth:
That's the general name for the arch, we could change it. I realize
actually that there's an issue because x86_64 will match x86

tooomm:
Quoting [wikipedia](https://en.wikipedia.org/wiki/X86-64):
"Some in the industry, including Apple,[4][5][6] use x86-64 and x86_64, while others, notably Sun Microsystems/Oracle[7] and Microsoft,[8] use x64."

<br>
>I realize actually that there's an issue because x86_64 will match x86

So just use "x64" and avoid that conflict?

ctrlaltca:
> I realize actually that there's an issue because x86_64 will match x86
It seems that the files uploaded to bintray currently use the win32/win64 notation: https://api.bintray.com/packages/cockatrice/Cockatrice/Cockatrice/files

From a more generic PoV, checking if the file name contains a specific substring is fragile. Eg. if the next version is going to be called "Vortexes" or "AbosXylophones", the release name itself would contain "exe" or "osx", causing a wrong match. 
Matching using something like eg. `endsWith("_osx.dmg")` and `endsWith("_win64.exe")` is a bit better.
The best solution would be to add on bintray a property describing the os+arch on each the file, but unfortunately it seems like it's not possible on the single file, but only on the full "release".

Another note: `QSysInfo::buildAbi()` requires Qt 5.4, we need to bump the min required version in CMakeLists.txt

Daenyth:
I'll try to incorporate the feedback soon. I think we can test it using the "update anyway" button. It should select the right version.

Daenyth:
> Another note: QSysInfo::buildAbi() requires Qt 5.4, we need to bump the min required version in CMakeLists.txt

I only call it on windows, can we make the min version 5.4 on windows and not do that on other platforms? On OSX it's easy to get 5.6, but Ubuntu 14.04 only has 5.2. Maybe it's not such a concern with 16.04 out now.

Daenyth:
I updated the PR, it's now checking for "win32.zip", "win64.zip", "win32.exe" and so on.

cockatrice/src/update_checker.cpp
```diff
@@ -37,18 +34,33 @@ void UpdateChecker::check()
...
+    } else {
+        qWarning() << "Error checking for upgrade version: wordSize is" << wordSize;
+        return false;
+    }
+
+    auto fileName = build
+        .toMap()["name"]
+        .toString();
+    // Checking for .zip is a workaround for the May 6th 2016 release
+    auto zipName = arch + ".zip";
```
tooomm: @ZeldaZach changed one of the two .exe files to zip to default to 64bit downloads on all clients. it was a workaround because otherwise the updater was not working at all and most people are expected to run 64bit.

we could just remove it completely because we have a real differentiation in place now and can offer win32.exe and win64.exe again?!

Daenyth: I considered that but I thought this was relatively harmless to add, and it makes it work with the release that has .zip uploaded. Remember that people will be upgrading *to* this release by using the old version so we will have to .zip this release as well - if I didn't check for zip, this release wouldn't be able to detect itself for 32bit.

Daenyth: We should probably keep it for the next several versions actually

tooomm: Hmm?
We have the initial updater in Gatewatchers --> no update possible for sure because it checks for newer releases, which there are none at that time
Shadows release: people with Gatewatchers should've been able to update, but since bintray changed their api it wasn't working

We had the same updater code in Shadows, Zach then quickly resolved the bintray api change after realizing and testing the new official release (auto-update from gatewatchers was not possible) and re-released Shadows a second time shortly after with that fix.
In the same commit he also changed how checking works because we had a "qt5.exe" check inside, he just changed it to "exe" (https://github.com/Cockatrice/Cockatrice/pull/1974/files#diff-f3a13ff16ad53a26fc8911e00b6c543fL52)
Which itself leads to problems described by ctrlaltca for names like "Vort**exe**s", but also made the updater not work again on windows because it finds two exe files on bintray (which got separated before by qt4.exe and qt5.exe checks). So he introduced the zip workaround and changed one of the exe files on bintray to a zip file (containing the exe) to avoid the need of a code change again (and a 3rd Shadows release).

As far as I understand it, the only downside would be that windows users on Shadows release can just update properly to the upcoming next version as long as we keep the zip file as a workaround in bintray. I don't see the need for having the zip code in the app.
But please enlighten me. :)
There might be a problem at one point for some people which haven't updated for a longer time and try to jump from Shadows to the after next release once we go back to normal bintray hosting. But that's it.

They can update from shadows to the next release (which then has your fix in) and will safely bring them to the next release after as well - even with two exe files up again on bintray!

So we need to get your fix in (without zip code), release new version with just one exe on bintray, features features features, next release with two exe files on bintray. All good?!

Daenyth: That would work if we're OK with not letting the update work right for
people on older versions. I'm willing to inconvenience us a little to
provide better backwards compatibility for our users, and the cost is
pretty low right now

ctrlaltca:
Code looks fine, but i'm not sure about any possible failure due to the cmake change. The first `FIND_PACKAGE(Qt5Widgets 5.0.3 REQUIRED)` initializes cmake data used to compile files in `common/`, then the second `FIND_PACKAGE(Qt5Widgets 5.4.0 REQUIRED)` would make files in `cockatrice/` being compiled with a possibly different Qt version.
I'd move the check inside the main CMakeLists.txt like this:
```
IF(WIN32)
    FIND_PACKAGE(Qt5Widgets 5.4.0 REQUIRED)  # For QSysInfo::buildAbi()
ELSE()
    FIND_PACKAGE(Qt5Widgets 5.0.3 REQUIRED)
ENDIF()
```

Daenyth:
Thanks, I wasn't sure of that part especially. Once I make that change, :shipit: ?

PR #2014, opened 2016-06-17
properly differentiate between singular and plural

tooomm:
Or did you remove for a reason @ZeldaZach ?

https://github.com/Cockatrice/Cockatrice/pull/2000/files#diff-06fec504bab8f0aa370636aea5d74fdfL790


ctrlaltca:
We should really start handling plural forms the correct way: http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

tooomm:
Does the travis error carries over from https://github.com/Cockatrice/Cockatrice/pull/2011?
There was a change to the `shortcutstab`

PR #2021, opened 2016-06-21
Colors to deck editor

### LINKED ISSUE #1888: Add color column in deck editor.
In the Deck editor section where we can sort cards would it be hard to add "Color" to the list?

We currently have name/set/CMC/type/ and p.t   could we add Color here?

Also for DFC could we have a way to check the second side?

ZeldaZach:
Fix #1888 

This adds the color column to the list. Not much more to say.

### DIFF
--- a/cockatrice/src/carddatabasemodel.h
+++ b/cockatrice/src/carddatabasemodel.h
@@ -12,7 +12,7 @@ class FilterTree;
 class CardDatabaseModel : public QAbstractListModel {
     Q_OBJECT
 public:
-    enum Columns { NameColumn, SetListColumn, ManaCostColumn, PTColumn, CardTypeColumn, CMCColumn };
+    enum Columns { NameColumn, SetListColumn, ManaCostColumn, PTColumn, CardTypeColumn, ColorColumn };
     enum Role { SortRole=Qt::UserRole };
     CardDatabaseModel(CardDatabase *_db, QObject *parent = 0);
     ~CardDatabaseModel();
--- a/cockatrice/src/carddatabasemodel.cpp
+++ b/cockatrice/src/carddatabasemodel.cpp
@@ -1,7 +1,7 @@
 #include "carddatabasemodel.h"
 #include "filtertree.h"
 
-#define CARDDBMODEL_COLUMNS 5
+#define CARDDBMODEL_COLUMNS 6
 
 CardDatabaseModel::CardDatabaseModel(CardDatabase *_db, QObject *parent)
     : QAbstractListModel(parent), db(_db)
@@ -44,6 +44,7 @@ QVariant CardDatabaseModel::data(const QModelIndex &index, int role) const
             card->getManaCost();
         case CardTypeColumn: return card->getCardType();
         case PTColumn: return card->getPowTough();
+        case ColorColumn: return card->getColors().join("");
         default: return QVariant();
     }
 }
@@ -60,6 +61,7 @@ QVariant CardDatabaseModel::headerData(int section, Qt::Orientation orientation,
         case ManaCostColumn: return QString(tr("Mana cost"));
         case CardTypeColumn: return QString(tr("Card type"));
         case PTColumn: return QString(tr("P/T"));
+        case ColorColumn: return QString(tr("Color(s)"));
         default: return QVariant();
     }
 }


Psithief:
Piss off, @mention-bot.

Daenyth:
@Psithief I'll try to configure mention-bot to not mention you in the future. Right now it's having some issue with the configuration, but I'll see if I can get it working

tooomm:
I don't get the need for this. It adds information that is already present (with `Mana cost`)

If somebody needs to filter for the color key to build an EDH deck for example they should use the filters anyway... otherwise this just adds complexity and an unnecessary column for mostly all users.
´Mana cost` is the way more valuable information for the general user.

Is it just me questioning the benefit of this?

Nightfirecat:
There's a few exceptions&mdash;see: Pact cycle, Transguild Courier, cards with the Devoid keyword(?)

tooomm:
>There's a few exceptions—see: Pact cycle, Transguild Courier, cards with the Devoid keyword(?)

Well, cards with devoid have no color? They have colored mana cost, but no color.
Pacts and Courier are also fine. It'll correctly display the color of the cards?

Daenyth:
Even if this is fairly niche I don't see a problem with the feature unless
it causes performance issues

cockatrice/src/carddatabasemodel.h
```diff
@@ -12,7 +12,7 @@ class FilterTree;
 class CardDatabaseModel : public QAbstractListModel {
     Q_OBJECT
 public:
-    enum Columns { NameColumn, SetListColumn, ManaCostColumn, PTColumn, CardTypeColumn, CMCColumn };
+    enum Columns { NameColumn, SetListColumn, ManaCostColumn, PTColumn, CardTypeColumn, ColorColumn };
```
Daenyth: Don't replace cmc, add a new column

tooomm: Is there a cmc column at all? Have there ever been?

I even reinstalled the newest master build (e81a6d4) to check... there is no such column in the deck editor. :)

Daenyth: @ZeldaZach I'm pretty sure CMC has been asked for if it's not already supported. Want to throw that in too?

tooomm: Please don't mess the deck editor up with all possible columns **as long as the user is not able to enable/disable them as in other apps** (with a right click on one of the column headers)!

Daenyth: @tooomm I'm not sure I understand the objection to adding more columns? You can just scroll sideways, you won't even see this unless you want to. What's the problem here?

tooomm: Annoyance > Use

(Plus, we have very nice filters for more specific use cases, which extend the basics that we offer in a very clean and user-friendly way to the everyday/new user...)


Imagine somebody is playing a format like Tiny Leaders... Nobody would think about ordering by cmc and scrolling to find suitable cards in a list of which 70% of all cards are not needed. Now think about a format where you are just allowed to use cmc 5,6 and 9 for example... everybody would just use the filters because it's way more convenient!
Or if you want to seriously brew and deeply fine tune your decks curve for example... you're going to use filters anyway. ;)

Another one. Let's say you want to simply build a new deck. You start your normal deck building process to build a RB Aggro for example. Now you order by the new color column to find yourself with colors ordered for BGRUW or WURGB. But no way black and red cards are all in one single list next to each other. *(this is an assumption)*
You don't think about filters here? ;)

So there is very few this would actually help users with I guess.

---

We can have all columns in the world, but then it needs to be customizable imo.
Scrolling something away is a very bad workaround for a new feature of questionable use for the majority in the first place.

Example of a right click on a column in the mail list of Thunderbird:
columns

ZeldaZach: I removed CMC because it wasn't used anywhere else in the codebase, it was randomly added to the column enum at some point, but not referenced. Also, I really like @tooomm's idea a lot. Open a ticket on that and ping ctrlaltca

Daenyth: Fair enough. Leave it off then.

ZeldaZach:
Anything holding this up? I'd like to see a menu-like tab visibility added at some point, but for now i think this is ok

tooomm:
Any cocerns regarding this? Or a good way to handle it?

>
Edit:
If people mess around with the column ordering it will directly influence the results of searches as well:
search
You expect the perfect fit card `Angel` on top of the list for obvious reasons... but since I ordered for p/t it's not. 

<br>
>I'd like to see a menu-like tab visibility added at some point

I highly appreciate to see that, too.
Then we could add as much information as we have to the columns as well and handle all needs.

PR #2025, opened 2016-06-24
Log UI

### LINKED ISSUE #1867: In-app way to view debug/warning/info logs
By now if a non-fatal error happens when loading a custom card database or downloading a card's image, the error is logged in the console using ```qDebug()```. It would be nice if the user could see these messages, preferably in a non-blocking way (no modal messagebox), so that any issue is more easily debuggable.

Split from https://github.com/Cockatrice/Cockatrice/pull/1859#discussion_r56013944

ctrlaltca:
This PR adds a specialized class for logging of debug messages in cockatrice, and a dialog where the last log messages are shown.
It should help users without a decent terminal (= windows users) to report problems with card database loading, picture downloading, etc..

dat menu

dat dialog

fix #1867

### DIFF
--- a/cockatrice/src/logger.cpp
+++ b/cockatrice/src/logger.cpp
@@ -1,0 +1,62 @@
+#include "logger.h"
+
+#include <QDateTime>
+
+#define LOGGER_MAX_ENTRIES 128
+#define LOGGER_FILENAME "qdebug.txt"
+
+Logger::Logger()
+: logToFileEnabled(false)
+{
+}
+
+Logger::~Logger()
+{
+    closeLogfileSession();
+    logBuffer.clear();
+}
+
+void Logger::logToFile(bool enabled)
+{
+    if(enabled)
+        openLogfileSession();
+    else
+        closeLogfileSession();
+}
+
+void Logger::openLogfileSession()
+{
+    if(logToFileEnabled)
+        return;
+
+    fileHandle.setFileName(LOGGER_FILENAME);
+    fileHandle.open(QIODevice::WriteOnly | QIODevice::Truncate | QIODevice::Text);
+    fileStream.setDevice(&fileHandle);
+    fileStream << "Log session started at " << QDateTime::currentDateTime().toString() << endl;
+    logToFileEnabled = true;
+}
+
+void Logger::closeLogfileSession()
+{
+    if(!logToFileEnabled)
+        return;
+
+    logToFileEnabled = false;
+    fileStream << "Log session closed at " << QDateTime::currentDateTime().toString() << endl;
+    fileHandle.close();
+}
+
+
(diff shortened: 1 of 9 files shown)


tooomm:
Very nice Idea. This will help a lot with error reports!

Maybe name it more explicitly?
`View Debug Log`?

cockatrice/src/logger.h
```diff
@@ -0,0 +1,37 @@
...
+public:
+    static Logger& getInstance()
+    {
+        static Logger instance;
+        return instance;
+    }
+private:
+    Logger();
+    ~Logger();
+    // Don't implement
```
Daenyth: why

cockatrice/src/main.cpp
```diff
@@ -91,8 +85,9 @@ int main(int argc, char *argv[])
 {
     QApplication app(argc, argv);
 
+    qInstallMessageHandler(CockatriceLogger);
     if (app.arguments().contains("--debug-output"))
```
Daenyth: This might be awkward to use on windows. How can we best handle that?

cockatrice/src/logger.cpp
```diff
@@ -0,0 +1,48 @@
...
+{
+    logBuffer.append(message);
+    if(logBuffer.size() > LOGGER_MAX_ENTRIES)
+        logBuffer.removeFirst();
+
+    emit logEntryAdded(message);
+
+    if(logToFileEnabled)
+    {
+        fileStream << message << endl;
```
Daenyth: Can we include timestamp at the beginning of line?

cockatrice/src/logger.h
ctrlaltca: I'll add a more descriptive comment: `// Singleton - Don't implement copy constructor and assign operator`

cockatrice/src/main.cpp
ctrlaltca: This was my exact question when i started thinking about this PR. Most users are running Windows, and getting logs from there was quite impossible.
Please note that logs are always collected, the `--debug-output` option only enables writing the log to a file named `qdebug.txt` to mimic the old behavior.

Since Cockatrice is a UI app, now that the log is available from inside the UI i don't think that writing it to a file is so important, but i suppose that it can still be useful for automated testing purposes (run, collect logs, grep for errors).

Daenyth: > Please note that logs are always collected, the --debug-output option only enables writing the log to a file named qdebug.txt to mimic the old behavior.

Gotcha, makes sense.

ctrlaltca:
Implemented proposed changes

tooomm:
Timestamps sounds nice.

Thanks to this I could already add more details on another issue I was facing. Thank you! :)

PR #2036, opened 2016-06-29
Fix openssl detection on win64

### LINKED ISSUE #2031: Can't update card database over SSL
<b>OS:</b>
Windows 7

<b>Cockatrice version:</b>
c19f225 (2016-06-26)

-___
<br>

When I open cockatrice, a window says, "Cockatrice is unable to load the card database. Do you want to update your card database now? If unsure or first time user, choose "Yes"

I press "Yes". It brings me to language choice screen, I select english and press "next". Now is the source selection screen. The "Download URL:" option has "https://mtgjson/json/AllSets.json.zip" as the source. I click next. Then a new window immediately pops up saying, "Network error: Error creating SSL context ()."

How do I move past this?

ctrlaltca:
Fix #2031

### DIFF
--- a/cmake/FindWin32SslRuntime.cmake
+++ b/cmake/FindWin32SslRuntime.cmake
@@ -5,22 +5,37 @@ if (WIN32)
   # Get standard installation paths for OpenSSL under Windows 
 
   # http://www.slproweb.com/products/Win32OpenSSL.html
-  set(_OPENSSL_ROOT_HINTS
-    ${OPENSSL_ROOT_DIR}
-    "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\OpenSSL (32-bit)_is1;Inno Setup: App Path]"
-    "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\OpenSSL (64-bit)_is1;Inno Setup: App Path]"
-    ENV OPENSSL_ROOT_DIR
-    )
-  file(TO_CMAKE_PATH "$ENV{PROGRAMFILES}" _programfiles)
-  set(_OPENSSL_ROOT_PATHS
-    "${_programfiles}/OpenSSL"
-    "${_programfiles}/OpenSSL-Win32"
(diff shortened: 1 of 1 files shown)


tooomm:
Working for me now! :+1:

ctrlaltca:
Ready for merge when the CI turns green, then

cmake/FindWin32SslRuntime.cmake
```diff
@@ -5,22 +5,37 @@ if (WIN32)
...
+      "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\OpenSSL (64-bit)_is1;Inno Setup: App Path]"
+      ENV OPENSSL_ROOT_DIR
+      )
+    file(TO_CMAKE_PATH "$ENV{PROGRAMFILES}" _programfiles)
+    set(_OPENSSL_ROOT_PATHS
+      "${_programfiles}/OpenSSL-Win64"
+      "C:/OpenSSL-Win64/"
+      )
+    unset(_programfiles)
+  else( CMAKE_SIZEOF_VOID_P EQUAL 8 )
```
Daenyth: This is the same if clause as above?

```
if( CMAKE_SIZEOF_VOID_P EQUAL 8 )
else( CMAKE_SIZEOF_VOID_P EQUAL 8 )
```

cmake/FindWin32SslRuntime.cmake
```diff
@@ -33,6 +48,7 @@ set(_OPENSSL_ROOT_HINTS_AND_PATHS
     PATHS ${_OPENSSL_ROOT_PATHS}
     )
 
+# Even if the dll is 64bit, it's still suffixed as *32.dll
```
Daenyth: Goddamnit windows

cmake/FindWin32SslRuntime.cmake
Daenyth: oh I see it's else, not elseif..

ctrlaltca: They guys developing cmake originally made some weird syntax choices, like this one. We already define `CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS`, so we could remove the condition from else() and endif()

PR #2047, opened 2016-07-02
auto admin

poixen:
Something I have in a personal build.

When I log in as an admin I want to automatically be able to use the features.
From a usability standpoint, it is frustrating if you want to deal with a situation but first have to navigate to the admin window. It also means when the server kicks you, or if you close your laptop for a while, you need to re-enable it.

This makes my life a little easier.

+ Auto unlocks admin features when log in
+ Removed popup to ask if you are sure.

### DIFF
--- a/cockatrice/src/tab_admin.cpp
+++ b/cockatrice/src/tab_admin.cpp
@@ -87,6 +87,8 @@ TabAdmin::TabAdmin(TabSupervisor *_tabSupervisor, AbstractClient *_client, bool 
     QWidget * mainWidget = new QWidget(this);
     mainWidget->setLayout(mainLayout);
     setCentralWidget(mainWidget);
+
+    actUnlock();
 }
 
 void TabAdmin::retranslateUi()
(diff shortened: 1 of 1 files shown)


Daenyth:
mention-bot still isn't respecting the config..

cockatrice/src/tab_admin.cpp
```diff
@@ -125,14 +127,12 @@ void TabAdmin::actReloadConfig()
 
 void TabAdmin::actUnlock()
 {
-    if (QMessageBox::question(this, tr("Unlock administration functions"), tr("Do you really want to unlock the administration functions?"), QMessageBox::Yes | QMessageBox::No) == QMessageBox::Yes) {
```
Daenyth: 👍

Daenyth:
A reasonable change.

ZeldaZach:
I know wooger was personally against this at one point, but I'm fine with this. Thanks for the help! 👍

Issue #1823, opened 2016-02-19
"show buddies only games" toggle


LethalPigeon7:
> Daenyth commented on 18 Nov 2015
I don't see a compelling reason to set a limit of games you can create so low.
I could see hiding buddy games under "unavailable", or perhaps having an extra toggle specifically for it.
Maybe if we created a relevancy rating that was the default game sort. A combination of password, buddy only, game age, open slots._ - from a somewhat related post that was rightfuly closed

I think the last feature is far too complicated and ultimately useless, all I wish for is a "show buddies only games" toggle.  Seems like an extremely effective and simple fix for a pretty annoying problem.
Assuming this is already being considered I do apologize, but I can't assume that since there's no open issue to remind of this that I know of

Daenyth:
Don't worry about opening issues. If there's none existing it's fine and if there is we'll point it out.

Anyway I'm fine adding a "show games restricted to creator's buddies" filter

tooomm:
What are we talking about here exactly?

a feature/option
- to hide/show all games with the tag `buddy only`?
- to hide/show games that you can't join because you're not a buddy of that games creator

<br>
If I remember correctly the second would be best and very helpful, but is not possible right now.

LethalPigeon7:
The former. I suggested the latter previously and was told just that. Not possible right now. But I couldn't humanly care less, about the advatages of the only hiding games of people I haven't buddied personally, I'd be upset if time was wasted on implementing this pointless feature if anything. I have been playing for 3 years and I have not once looked in main to join a buddies only game. I just right click people's names.

tooomm:
I put some more thoughts into such a toggle and how it could look like...

What do you guys think about new filter buttons for the most simple and basic thinks... quickly accessible and visually easy to check:
![filter buttons](https://cloud.githubusercontent.com/assets/9874850/14348227/5c4fedda-fcbb-11e5-8ac3-a83301a5c92a.png)
- Password protected games `show = yellow lock` / `hide = grey or open lock`
- only Buddy games `show = colored budy icon` / `hide = grey or x'ed budy icon`

More advanced and custom stuff would still go to the `Filter games` dialog box.

There could also be a toggle for spectating (![spec](https://cloud.githubusercontent.com/assets/9874850/14348299/df5a12c8-fcbb-11e5-8d78-4fc1cb316d24.png)) or reg only games...


@poixen you're our button and icon specialist ;)

LethalPigeon7:
[...]

PR #2049, opened 2016-07-03
filter buddies only

poixen:
You may now filter games based on if they are for buddies only.

Addresses #1823

### DIFF
--- a/cockatrice/src/dlg_filter_games.cpp
+++ b/cockatrice/src/dlg_filter_games.cpp
@@ -16,6 +16,9 @@ DlgFilterGames::DlgFilterGames(const QMap<int, QString> &_allGameTypes, const Ga
       allGameTypes(_allGameTypes),
       gamesProxyModel(_gamesProxyModel)
 {
+    showBuddiesOnlyGames = new QCheckBox(tr("Show '&buddies only' games"));
+    showBuddiesOnlyGames->setChecked(gamesProxyModel->getShowBuddiesOnlyGames());
+
     unavailableGamesVisibleCheckBox = new QCheckBox(tr("Show &unavailable games"));
     unavailableGamesVisibleCheckBox->setChecked(gamesProxyModel->getUnavailableGamesVisible());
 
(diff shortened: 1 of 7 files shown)


cockatrice/src/dlg_filter_games.cpp
```diff
@@ -119,6 +123,16 @@ void DlgFilterGames::setUnavailableGamesVisible(bool _unavailableGamesVisible)
     unavailableGamesVisibleCheckBox->setChecked(_unavailableGamesVisible);
 }
 
+bool DlgFilterGames::getShowBuddiesOnlyGames() const
```
Daenyth: Should this be const there? What does that do in that position?

poixen: It means that this method is not allowed to change class state.

cockatrice/src/settings/gamefilterssettings.cpp
```diff
@@ -15,6 +15,17 @@ QString GameFiltersSettings::hashGameType(const QString &gameType) const
     return QCryptographicHash::hash(gameType.toUtf8(), QCryptographicHash::Md5).toHex();
 }
 
+void GameFiltersSettings::setShowBuddiesOnlyGames(bool show)
+{
+    setValue(show, "show_buddies_only_games", "filer_games");
```
tooomm: typo? `filter_games`

tooomm:
Looks like this:
buddy

The tick box is checked by default! I think that's ok since we want to display as much games as possible per default...

It works and successfully filters for the `buddy only` tag in games! :+1:

cockatrice/src/settings/gamefilterssettings.cpp
poixen: crap that is an important typo, thanks

poixen:
Just updated the formatting, all restriction related filters are now grouped.
screen shot 2016-07-03 at 20 14 42

cockatrice/src/dlg_filter_games.h
```diff
@@ -4,6 +4,7 @@
 #include <QDialog>
 #include <QSet>
 #include <QMap>
+#include <QCheckBox>
```
ctrlaltca: This line should be in the .cpp, not in the .h

ctrlaltca:
Looks fine to me. Please squash before merge 👍

PR #2053, opened 2016-07-05
[WIP] card completion

poixen:
TODO
- allow one space, so that you can search for cards with spaces https://github.com/Cockatrice/Cockatrice/pull/2053#issuecomment-231187517
- investigate in crashes on win https://github.com/Cockatrice/Cockatrice/pull/2053#issuecomment-234109832

---

Typing `[[` in the main chat now brings up a card auto complete, similar to how mentions work.

The code is rough as shit. Its just a duplication of the mention logic.
There are code duplication and better ways to format it.

screen shot 2016-07-05 at 23 17 28

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -649,7 +649,7 @@ void MessagesSettingsPage::retranslateUi()
     chatGroupBox->setTitle(tr("Chat settings"));
     highlightGroupBox->setTitle(tr("Custom alert words"));
     chatMentionCheckBox.setText(tr("Enable chat mentions"));
-    chatMentionCompleterCheckbox.setText(tr("Enable mention completer"));
+    chatMentionCompleterCheckbox.setText(tr("Enable mention mentionCompleter"));
     messageShortcuts->setTitle(tr("In-game message macros"));
     ignoreUnregUsersMainChat.setText(tr("Ignore chat room messages sent by unregistered users"));
     ignoreUnregUserMessages.setText(tr("Ignore private messages sent by unregistered users"));
--- a/cockatrice/src/lineeditcompleter.cpp
+++ b/cockatrice/src/lineeditcompleter.cpp
@@ -1,25 +1,20 @@
 #include <QStringListModel>
 #include <QLineEdit>
 #include <QCompleter>
-#include <QWidget>
-#include <QTextCursor>
 #include <QAbstractItemView>
 #include <QScrollBar>
-#include <QStringListModel>
 #include <QKeyEvent>
-#include <QFocusEvent>
 #include "lineeditcompleter.h"
 
 LineEditCompleter::LineEditCompleter(QWidget *parent)
-: QLineEdit(parent)
-{
+        : QLineEdit(parent) {
 }
 
-void LineEditCompleter::focusOutEvent(QFocusEvent * e){
+void LineEditCompleter::focusOutEvent(QFocusEvent *e) {
     QLineEdit::focusOutEvent(e);
-    if (c->popup()->isVisible()){
+    if (mentionCompleter->popup()->isVisible()) {
         //Remove Popup
-        c->popup()->hide();
+        mentionCompleter->popup()->hide();
         //Truncate the line to last space or whole string
         QString textValue = text();
         int lastIndex = textValue.length();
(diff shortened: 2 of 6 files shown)


poixen:
Bug:
Pressing space closes the popup and selects the highlighted entry.
Simple fix.

cockatrice/src/dlg_settings.cpp
```diff
@@ -649,7 +649,7 @@ void MessagesSettingsPage::retranslateUi()
     chatGroupBox->setTitle(tr("Chat settings"));
     highlightGroupBox->setTitle(tr("Custom alert words"));
     chatMentionCheckBox.setText(tr("Enable chat mentions"));
-    chatMentionCompleterCheckbox.setText(tr("Enable mention completer"));
+    chatMentionCompleterCheckbox.setText(tr("Enable mention mentionCompleter"));
```
ZeldaZach: I feel like you CTRL+F'd and replaced for this, or did you change it for another reason?

tooomm:
:heart_eyes: 

Will it just consider cards from within your enabled list of sets? Or prioritize them?
Or always check for names in all cards.xml?

poixen:
@tooomm It works just like the deck tab.
You need to have the cards in your xml (card db).
It all needs to be tested. I just literally threw it together.

cockatrice/src/dlg_settings.cpp
poixen: Yeah mistake.

ZeldaZach:
Seems like it causes issues on windows? #2055

ZeldaZach:
I'm not sure why this built to the main repo, but I deleted it from bintray. This PR has many issues on windows

cockatrice/src/lineeditcompleter.cpp
```diff
@@ -1,133 +1,205 @@
 #include <QStringListModel>
 #include <QLineEdit>
 #include <QCompleter>
-#include <QWidget>
-#include <QTextCursor>
 #include <QAbstractItemView>
 #include <QScrollBar>
-#include <QStringListModel>
 #include <QKeyEvent>
-#include <QFocusEvent>
```
ZeldaZach: I have a feeling removing all these includes is the reason windows is crashing

Daenyth: Removing includes either will cause no issues or compiling will fail, so
that's not it.

poixen:
Todo:
I dont know when Ill get round to finishing this. Its a good starting point for someone who wants to mess with string logic.

Here are some notes:

```
QString LineEditCompleter::cardCursorWord(const QString &line) const {
    // no space
    // '[[jace,'
    // cur pos = 7

    // no space found, use -1 + 1 = 0
    // 7 - (-1) - 1 = 7
    // mid (0, 7) = [[jace,

    // with space
    // '[[jace, '
    // cur pos = 8

    // space found, use 7 + 1 = 8
    // 8 - 7 - 1 = 0
    // mid (8, 0) = '' <--- // todo this is the issue, we need to allow a single space

    // get everything to the left of the cursor and find the last " " in that result

    // we need to allow a single space
    return line.mid(line.left(cursorPosition()).lastIndexOf(" ") + 1,
                    cursorPosition() - line.left(cursorPosition()).lastIndexOf(" ") - 1);
}
```

+ Allow an additional space, so that you can search for cards with spaces.

tooomm:
>@tooomm It works just like the deck tab.
You need to have the cards in your xml (card db).

Cards needs to be in your cards.xml database for sure... but do they need to be enabled in `Edit sets` and actually displayed in the deck editor as well?
Or could I disable all sets except SOI for example and still search for `Force of Will` and the autocompleter would catch it?

Daenyth:
Are there outstanding changes we want to make to this?

If it doesn't crash, doesn't cause performance issues, then I'm happy rolling it out as-is and doing further work in a separate PR.

Daenyth:
si

ctrlaltca:
@Daenyth you ok? 😄

tooomm:
@poixen mentioned that it's not complete yet:
> Allow an additional space, so that you can search for cards with spaces.

Daenyth:
Gotcha. Once that's fixed let's get it out there.

ZeldaZach:
@Daenyth This causes numerous crash issues on Windows, which is why I'm not rolling this in

poixen:
This still needs to be done:
https://github.com/Cockatrice/Cockatrice/pull/2053#issuecomment-231187517

I dont think it will be done by me though. I tried to get the space to work and I dont want to spend any more time on it. Its here for someone else to pick up if they want a challenge.

The windows thing is a mystery, maybe when the logic is fixed it will work.

Daenyth:
Got it, thanks for the update.

ZeldaZach:
@Fizztastic Want to take a look into this?

Fizztastic:
@ZeldaZach I'm gonna write the fix for the card names and I'll take a look at this.

tooomm:
@Fizztastic are you still up for this? there is a release upcoming and the more great features get included the better... ;)

PR #2064, opened 2016-07-08
Changes to recommended DL + chat link

tooomm:


aeosynth:
i blocked this bot but i still get pinged, what do

tooomm:
Sorry for that...
We know that it's not working as intended. it shouldn't do anything with `.md` files, too:
https://github.com/Cockatrice/Cockatrice/blob/master/.mention-bot

There is some problem on their side maybe?
Our configuration should be fine. What about implementation?

ZeldaZach:
I think we should just trash the mention bot

Daenyth:
I'll file a bug report on mention-bot, it seems like it completely ignores the config file.

PR #2071, opened 2016-07-10
Game Viewer Update

poixen:
## Description
Updated the `Game Created` column to be more accurate when a game a newer.
The overall appearance is less clutter, cleaner and more central.
 
+ New games (<2 mins) will be defined as `New`
+ Games <10 mins will show the time in mins.
+ Games over 10 mins will show `10+ min`, `20+ min` . . . all the way up to `50+ min`
+ Games over 1 hour will show the time in hours: `1+ h`, `2+ h` . . .

This makes 3 areas of games and precision:
+ High precision : Newest games
+ Medium precision : Game may be getting stagnant
+ Low precision: Game has high chance of being stagnant.

## Changelist

+ Changed `Game Created` to `Age`.  `Game` is not needed as all entries in the list fall under the category `Game` that is defined in the upper left of the table.

+ Changed `Game Type` to just `Type` for the same reason as above.

+ Removed `ago` from each timestamp. This is just clutter and is not needed.

+ Changed the units from `m` and `hr` to `min` and `h` as defined [here](https://en.wikipedia.org/wiki/Non-SI_units_mentioned_in_the_SI).

+ The timestamps are now centred to help draw the eye. 

+ `Players` is now centred.

## Old
screen shot 2016-07-10 at 15 16 09

## New
screen shot 2016-07-11 at 10 23 35

### DIFF
--- a/cockatrice/src/gameselector.cpp
+++ b/cockatrice/src/gameselector.cpp
@@ -36,7 +36,7 @@ GameSelector::GameSelector(AbstractClient *_client, const TabSupervisor *_tabSup
     gameListView->setAlternatingRowColors(true);
     gameListView->setRootIsDecorated(true);
     // game created width
-    gameListView->resizeColumnToContents(1);
+    gameListView->setColumnWidth(1, gameListView->columnWidth(2) * 0.7);
     // players width
     gameListView->resizeColumnToContents(6);
     // description width
--- a/cockatrice/src/gamesmodel.cpp
+++ b/cockatrice/src/gamesmodel.cpp
@@ -13,30 +13,21 @@ enum GameListColumn {ROOM, CREATED, DESCRIPTION, CREATOR, GAME_TYPE, RESTRICTION
 const QString GamesModel::getGameCreatedString(const int secs) const {
 
     QString ret;
-    if (secs < SECS_PER_MIN)
-        ret = tr("<1m ago");
-    else if (secs < SECS_PER_MIN * 5)
-        ret = tr("<5m ago");
-    else if (secs < SECS_PER_HOUR)
-        ret = tr("%1m ago").arg(QString::number(secs / SECS_PER_MIN));
-    else if (secs < SECS_PER_MIN * 90) {
-        ret = tr("1hr %1m ago").arg(QString::number((secs / SECS_PER_MIN) - 60));
-    } else if (secs < SECS_PER_HOUR * 4) {
-        unsigned int hours = secs / SECS_PER_HOUR;
-        if (secs % SECS_PER_HOUR >= SECS_PER_MIN * 30)
+    if (secs < SECS_PER_MIN * 2) // for first min we display "New"
+        ret = tr("New");
+    else if (secs < SECS_PER_MIN * 10) // from 2 - 10 mins we show the mins
+        ret = QString("%1 min").arg(QString::number(secs / SECS_PER_MIN));
+    else if (secs < SECS_PER_MIN * 60) { // from 10 mins to 1h we aggregate every 10 mins
+        int unitOfTen = secs / SECS_PER_TEN_MIN;
+        QString str = "%1%2";
+        ret = str.arg(QString::number(unitOfTen), "0+ min");
+    } else { // from 1 hr onward we show hrs
+        int hours = secs / SECS_PER_HOUR;
+        if (secs % SECS_PER_HOUR >= SECS_PER_MIN * 30) // if the room is open for 1hr 30 mins, we round to 2hrs
             hours++;
-        ret = tr("%1hr ago").arg(QString::number(hours));
-    } else
-        ret = tr("5+ hrs ago");
-
+        ret = QString("%1+ h").arg(QString::number(hours));
+    }
     return ret;
-
-    /*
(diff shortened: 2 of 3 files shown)


tooomm:
Looks so clean and nice, yet useful! :+1: 
<br>
>
    New games (<2 mins) will be defined as New
    Games <30 mins will show the time in mins.
    Games over 30 mins will show >30 min
    Games over 1 hour will show the time in hours.

What's the opinion from others on this?
I generally like it, but I'm not certain if we should have a cap for hours as well.
`>3 h` or `>5 h` as maximum for example

cockatrice/src/gamesmodel.cpp
```diff
@@ -13,30 +13,19 @@ enum GameListColumn {ROOM, CREATED, DESCRIPTION, CREATOR, GAME_TYPE, RESTRICTION
...
-    else if (secs < SECS_PER_MIN * 90) {
-        ret = tr("1hr %1m ago").arg(QString::number((secs / SECS_PER_MIN) - 60));
-    } else if (secs < SECS_PER_HOUR * 4) {
-        unsigned int hours = secs / SECS_PER_HOUR;
-        if (secs % SECS_PER_HOUR >= SECS_PER_MIN * 30)
+    if (secs < SECS_PER_MIN * 2) // for first 2 mins
+        ret = tr("New");
+    else if (secs < SECS_PER_MIN * 30) // after 3 mins we show total mins up to 30 mins
+        ret = tr("%1 min").arg(QString::number(secs / SECS_PER_MIN));
+    else if (secs < SECS_PER_MIN * 30) { // from 30 to 1hr we show >30
```
ctrlaltca: This else looks suspicious, as it's identical to the previous one. Shouldn't it be `secs < SECS_PER_MIN * 60` ?

ctrlaltca:
I like the new look and the overall the code is 👍 , except the line from my previous comment.

cockatrice/src/gamesmodel.cpp
poixen: Good find, updating.

Daenyth: The duration formatting method would be a good one to unit test since it's
easy to make static. I'll see if i can add it into the pr

ZeldaZach:
I'd like to see accurate time (to the minute) for the first hour. The word "new" just seems a bit weird to me, but I'll grow on it.  Otherwise I'm fine with this.

poixen:
Why do yo need accurate time for the first hour? Does it matter if a room is 39 mins old vs 55 mins?
Increasing the range increases the clutter, which this PR is trying to reduce.

tooomm:
After some adjustments on your site and some more tests I think we found a very nice working version covering all needs without losing the focus on clearness! Well done. :+1: 

Actually you were right with your first assumption, most games happen to be around 30-40min max:
ages
Sometimes there are some 1-3h games as well, but that's it. Exepct for the abandoned rooms, haha.
<br>
**We don't need to have the times translatable. Can you just wrap the "New" label in tr() tags, please?
Numbers aren't translated anyway, and with min and h we go with international standards and people are used to them.** No need to confuse our translators even more. :)

poixen:
Updated `tr`.

cockatrice/src/gameselector.cpp
```diff
@@ -36,7 +36,7 @@ GameSelector::GameSelector(AbstractClient *_client, const TabSupervisor *_tabSup
     gameListView->setAlternatingRowColors(true);
     gameListView->setRootIsDecorated(true);
     // game created width
-    gameListView->resizeColumnToContents(1);
+    gameListView->setColumnWidth(1, gameListView->columnWidth(2) * 0.7);
```
Daenyth: What's the 0.7?

tooomm: It should resize the default size of the `Age` column to 70%.

poixen: scaling down a bit. `1.0` is too wide.

setti93:
I think left alignment looks way better in the "Age" column.

tooomm:
We compared different versions at the beginning of this pr in gitter and mostly agreed on center alignment.
I don't know how left aligned would look like in the recent version since we did many changes in between.
Could you provide a picture please?

setti93:
I'm not a big fan of centered alignment when the content may have different length.
I think that will look good with numbers and text properly vertically aligned.

I mean something like this:
ssss

poixen:
@setti93 
Im not quite sure wht you posted there, but it looks really messy.
If you scroll up you can see everything in-line.

We talked as a group and found that left justification was too far away with the new text formatting, right was too cramped and that centered worked the best.

It seems like you understand some code though, you are free to run any local client you like.
So far this change is going through... Im not quite sure what we are waiting for ... @Daenyth ?

tooomm:
@setti93 I thought you have a working version running and could provide a real picture.
This is not helping too much, since we know what your talking about... I for myself just can't evaluate it properly as long as I don't see it in person. :)
But as @poixen said we mostly agreed in a discussion that this is ok as is.

Daenyth:
:shipit:

PR #2077, opened 2016-07-12
Some web client UI fixes

ctrlaltca:
Nothing big, just positioning elements in a way similar to cockatrice and hiding technical details from the user


Daenyth:
👍

PR #2078, opened 2016-07-14
fix typo in sample servatrice ini

skwerlman:
a `'` was used in place of a `;` at the beginning of a comment


ctrlaltca:
Thank you!

PR #2083, opened 2016-07-15
updated login

poixen:
## Description
Updated login screen. Previously the dlg would auto highlight the `port` number. Its more useful to have the username the first highlight item. I have also grouped the 2 different parts of the dlg.

## Changelist
 + `Username` is now the auto selected item.
 + Separation of Server and Login.
 + `New Host` label is faded when s`Previous Host` is selected.

## Old
screen shot 2016-07-15 at 15 33 22

## New
screen shot 2016-07-15 at 15 41 20


Daenyth:
👍

ctrlaltca:
Nice 👍

PR #2084, opened 2016-07-15
Shutdown timer changes

ZeldaZach:
Change shutdown timer from 10 minutes to 30 minutes.

The message will popup: when initialized, every 30 minutes after it was initialized, and every minute for the last 5 minutes

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -599,6 +599,7 @@ void Servatrice::scheduleShutdown(const QString &reason, int minutes)
 {
     shutdownReason = reason;
     shutdownMinutes = minutes + 1;
+    totalMinutes = shutdownMinutes;
     if (minutes > 0) {
         shutdownTimer = new QTimer;
         connect(shutdownTimer, SIGNAL(timeout()), this, SLOT(shutdownTimeout()));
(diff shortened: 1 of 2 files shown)


ctrlaltca:
The new math seems a bit too complicated to me.
I'd drop "isFirstShutdownMessage", rename "totalMinutes" to "nextShutdownMessageMinutes" and just make the check:
```
if (shutdownMinutes <= 5 || shutdownMinutes == nextShutdownMessageMinutes) {
nextShutdownMessageMinutes = shutdownMinutes / 2;
...
}
shutdownMinutes--;
```

ZeldaZach:
@ctrlaltca I made the changes you requested. I was trying to figure out the best way to do the math in my head and it was a bit complex :D

ctrlaltca:
I'm not sure it will work like that; at first shutdownMinutes == nextShutdownMessageMinutes, but at the beginning of `Servatrice::shutdownTimeout()` shutdownMinutes gets decreased by one.
That's why in my example i moved  `shutdownMinutes--;` after the check

ZeldaZach:
Changed @ctrlaltca

ctrlaltca:
Tested working as expected: 28 -> 14-> 7 -> 5 -> 4 -> 3 -> 2 -> 1 :+1

PR #2087, opened 2016-07-16
Client id nametoip fix

woogerboy21:
The way the clientid functionality works is it takes the name of the
host and combines it with the portion of the mac address to generate a
client id.  Using the name can allow for easy circumvention of the
clientid generation process by a user if they simply use a hostname file
to constantly change the name to ip resolution.  With this change the
name will no longer be used and the looked up ip address of the server
being connected to will be used.

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -1,7 +1,10 @@
+#include <QDebug>
 #include <QList>
 #include <QTimer>
 #include <QThread>
 #include <QCryptographicHash>
+#include <QHostInfo>
+#include <QHostAddress>
 #include "remoteclient.h"
 #include "settingscache.h"
 #include "pending_command.h"
@@ -365,7 +368,15 @@ void RemoteClient::disconnectFromServer()
 QString RemoteClient::getSrvClientID(const QString _hostname)
 {
     QString srvClientID = settingsCache->getClientID();
-    srvClientID += _hostname;
+    QHostInfo hostInfo = QHostInfo::fromName(_hostname);
+    if (!hostInfo.error()) {
+        QHostAddress hostAddress = hostInfo.addresses().first();
+        srvClientID += hostAddress.toString();
+    }
+    else {
+        qDebug() << "Warning: ClientID generation host lookup failure [" << hostInfo.errorString() << "]";
+        srvClientID += _hostname;
+    }
     QString uniqueServerClientID = QCryptographicHash::hash(srvClientID.toUtf8(), QCryptographicHash::Sha1).toHex().right(15);
     return uniqueServerClientID;
 }


ctrlaltca:
To play the devil's advocate:
A hostname change requires a reboot (at least on windows), while a change of the ip address is just a few clicks (at least for a common home connection with a router serving dhcp).
What's harder to fake/change? Maybe the gateway ip?

tooomm:
>What's harder to fake/change? Maybe the gateway ip?

Isn't the gateway ip also just a configurable router setting?

Daenyth:
If you mess with it you'll likely force all clients on the lan to get new dhcp leases.

Daenyth:
We shouldn't use it though because it's the same for many networks. >50% of the time gateway is just going to be 192.168.1.1

woogerboy21:
I'm not sure how the question really pertains to the issue.  Currently the way things work is the clients MAC address is used in combination with the host name of the server there connecting to.  That is what is used when generating the client ID.  Not the hostname of the client.

So if someone wants to easily circumvent the current implementation of the client ID generation all they have to do is edit the host file on there workstation and say me = IP address of cockatrice host.  Then next time change the "me" to "you" and so forth and a different client ID will be generated resulting in what we don't want.

There is no reboot or anything required on the users client machine.  Just a file edit before starting up the cockatrice client.

woogerboy21:
WAY back when me and @Daenyth had a discussion about this and he actually made the comment we could implement this type of change to resolve the hostname of the server the client is connecting to and use the IP address instead of the hostname.  Hence this PR.  It just took me a bit to get to with my eye surgery :(

ctrlaltca:
> I'm not sure how the question really pertains to the issue.  Currently the way things work is the clients MAC address is used in combination with the host name of the server there connecting to.  That is what is used when generating the client ID.  Not the hostname of the client.

My fault, i misunderstood the code of the PR. Sorry for that, please disregard my previous comment.

woogerboy21:
@ctrlaltca The resulting code change is pretty trivial.  Any opinion on if this is the easiest way or any other recommended methods?

cockatrice/src/remoteclient.cpp
```diff
@@ -365,7 +367,9 @@ void RemoteClient::disconnectFromServer()
 QString RemoteClient::getSrvClientID(const QString _hostname)
 {
     QString srvClientID = settingsCache->getClientID();
-    srvClientID += _hostname;
+    QHostInfo hostInfo = QHostInfo::fromName(_hostname);
+    QHostAddress hostAddress = hostInfo.addresses().first();
```
Daenyth: We probably need to check for failure here.

woogerboy21: What logic would you prefer if it does fail?  

Currently if it fails we append nothing to the end of the mac address section of the clientid and it gets hashed.  if it doesnt fail (success) we append the ip address found.  Fall back to name possibly?  IMO, if the lookup fails chances are the client will never connect to the cockatrice server they intend on connecting to anyway because Qt is using the underlying operating system functionality to perform the lookups so if the OS is failing to look things up then the app isnt getting to its end point in general.

Or are you thinking like a code handling type check?

Daenyth: well, we can check `QHostInfo::error()` and `QHostInfo::errorString()`. If there's an error on this step we probably can't connect to the server. I think it shouldn't attempt to connect if this fails.

woogerboy21: This block of code isn't even reached if the connection to the server fails.  The connection is made and then the doLogin code fires.  Are you saying that we should check for a lookup error on something that has already been looked up and successfully resolved?

woogerboy21:
Also, if you want a check that's fine but what behavior did you expect if the check fails?  Disconnect client? Process clientid's with hostname? No hostname just Mac? Play a sound? :)

Daenyth:
As a fallback, why don't we use the hostname as in the current version, if the lookup fails?

woogerboy21:
@Daenyth Is that what you had in mind?

Daenyth:
Sure, maybe just `qInfo` the `errorString()` so we can see it in the log viewer if needed?

woogerboy21:
@Daenyth 
Just curious, what difference is there between qInfo and qDebug?  And does it matter which one I would use in this case?  I used qDebug to stay consistent with other area's of the code currently but wasnt sure if you wanted me to use qInfo for sure.

Daenyth:
qDebug calls are ignored in Release build mode

woogerboy21:
Wait, why is this failing on the "trusty" travis tests?

ctrlaltca:
The error is:
```
/home/travis/build/Cockatrice/Cockatrice/cockatrice/src/remoteclient.cpp:376:15: error: ‘qInfo’ was not declared in this scope
         qInfo() << "Warning: ClientID generation host lookup failure [" << hostInfo.errorString() << "]";
```

You need to #include the correct header in order to use `qInfo()`, or you can just revert to `qDebug ()` (it's already included by some other file), since catch them also with #2025

woogerboy21:
@ctrlaltca 
qInfo is a part of QTGlobal which is already declared from what I can tell.  What exactly am I overlooking?  If it were a missing declaration wouldnt all the travis builds fail?

PR #2096, opened 2016-07-19
Fix infinite loop on local game close

### LINKED ISSUE #2091: CrashLog from Single Player Game
Crashlog location: https://gist.github.com/ZeldaZach/04ff77d4e0f68ce221fe159e950b21ce

ctrlaltca:
Since #1991 servatrice only works in multithread mode, so in`Server::prepareDestroy()` the code  dealing with client destruction was using proper locking.
Since the "local game" server in cockatrice is not multithreaded, while closing a game it will hang indefinitely waiting for the lock to be removed.
This PR removes locking when destroying a local game server, the same way it was implemented before #1991.

fix #2091

### DIFF
--- a/cockatrice/src/localserver.cpp
+++ b/cockatrice/src/localserver.cpp
@@ -11,6 +11,10 @@ LocalServer::LocalServer(QObject *parent)
 
 LocalServer::~LocalServer()
 {
+    // LocalServer is single threaded so it doesn't need locks on this
+    while (!clients.isEmpty())
+        clients.first()->prepareDestroy();
+
     prepareDestroy();
 }
 
--- a/common/server.cpp
+++ b/common/server.cpp
@@ -59,28 +59,6 @@ Server::~Server()
 
 void Server::prepareDestroy()
 {
-    // dirty :(
-    clientsLock.lockForRead();
-    for (int i = 0; i < clients.size(); ++i)
-        QMetaObject::invokeMethod(clients.at(i), "prepareDestroy", Qt::QueuedConnection);
-    clientsLock.unlock();
-
-    bool done = false;
-
-    class SleeperThread : public QThread
-    {
-    public:
-        static void msleep(unsigned long msecs) { QThread::usleep(msecs); }
-    };
-
-    do {
-        SleeperThread::msleep(10);
-        clientsLock.lockForRead();
-        if (clients.isEmpty())
-            done = true;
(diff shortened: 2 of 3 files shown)


common/server.cpp
```diff
@@ -57,29 +57,35 @@ Server::~Server()
 {
 }
 
-void Server::prepareDestroy()
+void Server::prepareDestroy(bool lockClients)
 {
     // dirty :(
```
Nightfirecat: Is this still a dirty way to do this?

cockatrice/src/localserver.cpp
```diff
@@ -11,7 +11,7 @@ LocalServer::LocalServer(QObject *parent)
 
 LocalServer::~LocalServer()
 {
-    prepareDestroy();
+    prepareDestroy(false);
```
Daenyth: What about making `prepareDestroy` virtual and overriding the behavior differently in real server and local server mode?

common/server.cpp
ctrlaltca: as dirty as before, so i didn't remove the comment yet.

ctrlaltca:
Reimplemented splitting methods to different classes as suggested by @Daenyth. A virtual method was not even needed.

cockatrice/src/localserver.cpp
```diff
@@ -11,6 +11,10 @@ LocalServer::LocalServer(QObject *parent)
 
 LocalServer::~LocalServer()
 {
+    // no need to lock clients, they live in the same thread
```
Daenyth: Does the `LocalServer` class even have any locks? If not the comment is misleading. If it owns some but they are unused from lack of necessity that's even more confusing.

ctrlaltca: `LocalServer` is just a mockup built around the abstract `Server` class, so it implements almost nothing. All the client handling is in `Server` too, except this part.
I thought that in the future someone reading this code would ask himself why clients are directly being destroyed here without proper locking, instead of in the async+lock way they are destroyed in servatrice; so i left this comment.

Daenyth: Maybe `// LocalServer is single threaded so it doesn't need locks on this` ?

ctrlaltca:
Comment updated

ZeldaZach:
Seems fixed to me. Good job! :shipit:

PR #2097, opened 2016-07-19
Remove additional minute in server shutdown

ctrlaltca:
With #2084 the server shutdown timer calculation has changed, and w don't need anymore to add one minute to the timer.


Daenyth:
Looks good

PR #2099, opened 2016-07-20
Servatrice rooms table respect Server ID

### LINKED ISSUE #1123: Servatrice rooms handling misses multi-server support
```
| cockatrice_rooms            |
| cockatrice_rooms_gametypes  |
```
Both these tables in the database do not contain a server id field and the server itself does not query only the rooms associated with its server id.  With out these each server stood up that share the same database will have to have the same rooms and game types in those rooms.

woogerboy21:
Fix #1123
This change add's the db column "id_server" to the rooms and rooms_gametypes table and updates servatrices queries to respect room/room information per server based on the server id.  This is the only portion of the database that I could find that doesn't respect a server id when the DB is queried for information and with the change will allow multiple servers to use the same DB.

Using the same DB will not of course allow for active game information to be shared but what it will allow for is geographically dispersed servers to replicate there DB information and keep a consistent copy amongst sites.

### DIFF
--- a/servatrice/migrations/servatrice_0014_to_0015.sql
+++ b/servatrice/migrations/servatrice_0014_to_0015.sql
@@ -1,0 +1,6 @@
+-- Servatrice db migration from version 14 to version 15
+
+alter table cockatrice_rooms add `id_server` tinyint(3) not null default 0;
+alter table cockatrice_rooms_gametypes add `id_server` tinyint(3) not null default 0;
+
+UPDATE cockatrice_schema_version SET version=15 WHERE version=14;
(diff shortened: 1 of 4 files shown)


woogerboy21:
@Daenyth  or @ctrlaltca 
Any idea why travis is complaining about the DB versions and failing out?  I updated the expected schema version in the database header and things compile and work on my end.  Not sure how travis works though.

ctrlaltca:
From a quick look it seems that you missed to update the version in servatrice.sql: https://github.com/Cockatrice/Cockatrice/blob/master/servatrice/servatrice.sql#L23

I added some docs recently: https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md#database-migrations

woogerboy21:
Ah yes, you are correct.  That is my bad.

woogerboy21:
@ctrlaltca / @Daenyth  / @ZeldaZach 
Any objection to this change?  Its another fairly simple one.

Daenyth:
Looks like it's currently failing with `Schema version 14 does not equal new version 15`

The changes look fine once the sql stuff is fixed.

Daenyth:
And we'll want to squash commits

PR #2100, opened 2016-07-21
Updated connect dlg

poixen:
## Description
Moved the check boxes into their relative sections to keep the dialog tidy, added a shortcut to open the connection dialog.

## Changelist

+ `Save password` has been moved into the `Login` group
+ `Auto connect` has been moved into the `Server` group
+ Renamed `Auto connect at start` to `Auto connect`.
+ Added tool tip to `Auto connect`: `Automatically connect to the most recent login when Cockatrice opens`
+ Added a default key binding of `Ctrl/Cmd + L` to open the connection dialog (L for Login).

## Old
screen shot 2016-07-21 at 11 49 48

## New
screen shot 2016-07-21 at 11 49 09

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -53,7 +53,9 @@ DlgConnect::DlgConnect(QWidget *parent)
     savePasswordCheckBox = new QCheckBox(tr("&Save password"));
     savePasswordCheckBox->setChecked(settingsCache->servers().getSavePassword());
 
-    autoConnectCheckBox = new QCheckBox(tr("A&uto connect at start"));
+    autoConnectCheckBox = new QCheckBox(tr("A&uto connect"));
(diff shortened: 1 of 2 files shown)


Daenyth:
:shipit:

tooomm:
- I'm wondering why you think that "auto connect" belongs more to the server group than the login group? I wouldn't expect it there. It's also confusing that a checkbox in the upper part gets unlocked by another one at the bottom.

- Why do we use a colloquial abbreviation like `Auto` ?
To name a different option it could be `Connect on launch" for example. 

- The tooltip kind of makes no sense?
`Automatically connect to the most recent login when Cockatrice opens`
"connect to the most recent login" --> ?
It should be "...login to the last used server..." I guess?


<br>
Well, sorry for being late. But since the PR was just open for a few hours and you easily lose track once it's closed...

I'm posting here instead opening an new issue on purpose.
Maybe you talked about all that before and it's *intended and correct*.

PR #2101, opened 2016-07-21
Remove news table (cleanup)

### LINKED ISSUE #2098: "News" table is un-used in the code base
Looks like the "news" table is created in the .sql script and referenced throughout the db upgrade scripts but the code base for the server never uses the table.  Seems like a line of development that may have been dropped at one point in time.  We should probably remove that table and create a db migration script to remove it from the db of server owners.

woogerboy21:
Fix #2098 
Remove of the unused news table


ZeldaZach:
Seems good to me :shipit:

PR #2103, opened 2016-07-21
More screen space

poixen:
## Description
More screen space for decks and in-game

## Changelist

+ Deck view now fills up as much room as it can.
+ Game view now fills up as much room as it can.
+ Renamed `Load local deck` to `Load deck...` (`...` is the standard for opening a load dialog)
+ Renamed `Load deck from server` to `Load remote deck...` (`...` is the standard for opening a load dialog)

## Old
Notice the grey wasted space around the edge.
screen shot 2016-07-21 at 19 28 11

## New
screen shot 2016-07-21 at 20 03 01

screen shot 2016-07-21 at 20 05 09

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -71,14 +71,14 @@
 #include "get_pb_extension.h"
 
 ToggleButton::ToggleButton(QWidget *parent)
-    : QPushButton(parent), state(false)
+        : QPushButton(parent), state(false)
 {
 }
 
 void ToggleButton::paintEvent(QPaintEvent *event)
 {
     QPushButton::paintEvent(event);
-    
+
     QPainter painter(this);
     QPen pen;
     pen.setWidth(3);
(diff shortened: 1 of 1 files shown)


poixen:
I would really like to see a before and after on a windows machine too. Any takers? @tooomm ?

Nightfirecat:
I'll get one up tonight if nobody else has.

ZeldaZach:
Tested on mac and seems to work as expected. I expect this to be well liked

cockatrice/src/tab_game.cpp
```diff
@@ -1185,16 +1188,16 @@ Player *TabGame::getActiveLocalPlayer() const
 {
     Player *active = players.value(activePlayer, 0);
     if (active)
-        if (active->getLocal())
-            return active;
-    
+    if (active->getLocal())
```
Daenyth: Can we oneline this check? `if (active && active->getLocal()) return active;` ?

Nightfirecat:
##New

Daenyth:
✨ :shipit: ✨

Issue #2112, opened 2016-07-25
Allow just one game room type at a time


tooomm:
[Idea taken from Gitter](https://gitter.im/Cockatrice/Cockatrice?at=579531ce32bd01d843ba67ab)

It's confusing to allow different types at a time, e.g. `standard, commander` or `commander, limited`.

check boxes --> radio buttons

CC: @poixen

poixen:
This is super easy to do client side. I'm not sure if we NEED to fix this server side, that will take more time and would require everyone update. 

I can have a look into this tomorrow. 

Best Regards

Jinxstands4Jinx:
This does make some sense as having people make modern,standard rooms makes no sense and for people like myself who only show certain formats it messes up our settings.

Daenyth:
I'd prefer a client side only fix; this keep the proto more flexible. Another game or client might want to use those fields for something where combinations do make sense.

tooomm:
One thing that just came to my mind... with radio buttons there needs to be one selection as default, right?
So we should make `Others` the pre-selection or introduce a new `None` field (a bit separated from the list maybe).
Once people start using it, it should remember the last selection though.

poixen:
You can lock down the `create` button until a selection has been made and have nothing selected to start with.

ZeldaZach:
I like that idea @poixen

ctrlaltca:
fixed by #2115

tritoch:
I'd like to see this be server managed.  There is a use case for allowing game with multiple types.  Pauper cube is Pauper & Limited and should be able to be advertised to players of both.

Daenyth:
There's no need to have it be server-managed. The motivating reason for this is streamlining the UI. If we can make a better UI for the games list I'd be very happy to re-add this. I used multi-format stuff in the same way you mention.

PR #2115, opened 2016-07-27
Radio buttons for game types

poixen:
## Description
Radio buttons are now used over select boxes. This will reduce rooms with multiple game types, which we have discussed in #2112 

## Changelist
+ Radio buttons are used instead of checkboxes for `Game Type`

## Image
screen shot 2016-07-27 at 09 16 04

### DIFF
--- a/cockatrice/src/dlg_creategame.cpp
+++ b/cockatrice/src/dlg_creategame.cpp
@@ -3,24 +3,21 @@
 #include <QCheckBox>
 #include <QPushButton>
 #include <QGridLayout>
-#include <QVBoxLayout>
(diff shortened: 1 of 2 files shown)


tooomm:
Pretty cool. `Clear` is not resetting the selection to *no selection*, though.
The value used for the last game persists.

poixen:
@tooomm I have updated the logic to not allow you to click `Ok` without having a type. Also clicking `Clear` now unchecks all.

tooomm:
Working as expected I would say. :+1:

ctrlaltca:
Confirmed working as expected

PR #2117, opened 2016-07-29
Webclient: misc improvements (was: parse html in server message)

### LINKED ISSUE #2088: Current web client server message formating problems
Looks like the current web client pulls the server message in but doesnt format the messages similar to how the actual client does.  If the server message contains html code structure (might be others as well) the web client simply puts the data in the field and doesnt follow formating rules.

image

ctrlaltca:
Fix #2088

Implements an htmlSanitize() method using a whitelist approach to only allows some specific html tags, attributes and protocols in anchors.

Minor: added favicon and clientfeatures set needed to connect to wooger's server

### DIFF
--- a/webclient/index.html
+++ b/webclient/index.html
@@ -2,6 +2,7 @@
 <html> 
 <head>
   <title>Cockatrice web client</title>
+  <link rel="shortcut icon" href="imgs/cockatrice.png" />
   <link rel="stylesheet" href="js/jquery-ui-1.11.4.css">
   <link rel="stylesheet" href="style.css">
 </head>
@@ -15,6 +16,7 @@ Loading cockatrice web client...
 </div>
 
 <div id="error-dialog" title="Error" style="display:none"></div>
+<div id="info-dialog" title="Information" style="display:none"></div>
 
 <div id="tabs" style="display:none">
   <ul>
@@ -26,7 +28,7 @@ Loading cockatrice web client...
   <div id="tab-login">
     <h3>Login to server</h3>  
     <label for="host">Host</label>
-    <input type="text" id="host" value="127.0.0.1" />
+    <input type="text" id="host" value="" />
     <br/>
     <label for="port">Port</label>
     <input type="text" id="port" value ="4748" />
@@ -54,11 +56,11 @@ Loading cockatrice web client...
   <div id="tab-account">
     <div class="buddies-container">
       <h3>Buddies</h3>
-      <select id="buddies" size="10"></select>
+      <ul id="buddies" size="10"></ul>
     </div>
     <div class="ignores-container">
       <h3>Ignores</h3>
-      <select id="ignores" size="10"></select>
+      <ul id="ignores" size="10"></ul>
     </div>
     <div class="userinfo-container">
       <h3>User info</h3>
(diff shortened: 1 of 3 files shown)


webclient/webclient.js
```diff
@@ -158,7 +158,7 @@ var WebClient = {
       "password" : this.options.pass,
       "clientid" : this.guid(),
       "clientver" : "webclient-0.1 (2015-12-23)",
-      "clientfeatures" : [ "client_id", "client_ver"],
+      "clientfeatures" : [ "client_id", "client_ver", "client_warnings", "feature_set", "mod_log_lookup", "room_chat_history", "user_ban_history"]
```
tooomm: Does the webclient support all these??

Daenyth: I don't think so

tooomm:
>Minor: added (...) clientfeatures set needed to connect to wooger's server

You can connect to http://cockatrice.woogerworks.com/play/ right now?
In the `Account` tab you see this:
missing_features

Daenyth:
I don't think those are mandatory features, just suggested. @woogerboy21  ?

webclient/webclient.js
ctrlaltca: i had to add them in order to make the client able to connect to wooger's server, that requires some of those  features in clients (i just added them all, didn't check them one by one).

tooomm: So they shouldn't be listed?

tooomm:
I didn't know that we differentiate between suggested and requiered!?

webclient/index.html
```diff
@@ -26,7 +27,7 @@
   <div id="tab-login">
     <h3>Login to server</h3>  
     <label for="host">Host</label>
-    <input type="text" id="host" value="127.0.0.1" />
+    <input type="text" id="host" value="cockatrice.woogerworks.com" />
```
ZeldaZach: Do we want this, or was this a test you were doing?

ctrlaltca: Which one is a better default? 127.0.0.1 is probably good only for local testing

woogerboy21:
IMO.  The code should not be adapted to fit one particular host over another.  I do understand why some of the things set with this change are the way they are since woogerworks is the most widely used host.  The question I have would be, how difficult is it for the provider of the service (web client in this case) to make adjustments for there preferences in order to make the web client work?  I already went through on the version I am hosting and made the changes rather easy but would there be an expectation for some one else to do so?

woogerboy21:
I dont know if i like the idea of putting features into a client that doesnt really support the features.

woogerboy21:
> I don't think those are mandatory features, just suggested. @woogerboy21 ?

The only two features I have set as required are: client_ver,client_warnings

It would be nice to see the web client have the ability to do client warnings at some point.  Only because right now some one could log into the web client, spew crap, a mod warn them and they would never see the warning.

webclient/index.html
woogerboy21: I didnt look, but if your going to expand the pre-defined host string you may want to expand the box character length as well. I know when I worked on the page at the woogerworks site I expanded the box length to fit the entire string without causing you to have to arrow over to see the right most characters.

ctrlaltca:
I'll update the PR adding  real support for client_warnings, removing the unneeded client features and implementing all the other suggested changes.

ctrlaltca:
Implemented changes:
 * Add an autocomplete with known server names to the connect dialog;
 * Added support for server messages (identification, shutdown, user warning and promotion);
 * Connection dialog’s inputs are now wider;
 * Only declare implemented client features;
 * Added check for protocol version on connect before login.

PR #2120, opened 2016-08-01
Italian translation update

setti93:
Updated italian translation file


ZeldaZach:
If you'd like to do translation updates, check out our project on https://www.transifex.com/. We have a page on the wiki about it and how you can do updates.

setti93:
I will, thanks :)

tooomm:
This is where you want to look: https://github.com/Cockatrice/Cockatrice/wiki/Translation-FAQ

https://www.transifex.com/cockatrice/cockatrice/ is the project on the transifex service which we use for managing the process. Check it out.
Note that @ctrlaltca needs to update the files to transifex first, translators work with them for some time and then they will make it back to the client. But in between there has maybe been new features and other strings changed too... so there is always transition and a delay!

setti93:
@tooomm I didn't know how translation process works. That's the first project i contribute to and i'm learning a lot these days.

tooomm:
Happy to see you contributing!
Always feel free to ask any questions here or the gitter chat! ;)

Daenyth:
@setti93 the way you did it (editing directly in git) was correct for our old process, but we now use transifex as a middle man because by doing that we don't force translators to use git - more people can translate. And they don't need to install any tools or understand xml.

PR #2125, opened 2016-08-09
Fix clipboard behavior in deck editor and game

### LINKED ISSUE #1330: No shortcut for copying card text

### LINKED ISSUE #1762: Some parts of the app doesn't allow to highlight text on Windows
I can't copy the text from in-game windows, such as the "about" menu or the "card info" tab both in game or in deck editor (both card text and other qualities like name or card type). Especially confusing with card text since it can randomly be selected unlike the others, but trying to copy it doesn't work.  
This bug isn't the end of the world, since it's simply a matter of inconvenience but it comes up a few times a day, it'd be swell if it were to be fixed.

ctrlaltca:
1. Change the default shortcut keys for "load/paste deck from clipboard" from `ctrl+c/v` to `ctrl+shift+c/v`, to avoid overriding the common shortcuts for copy/paste text provided by the OS
2. Make text selectable in the "card info" panel in both the deck editor and game
3. Make text form the "about dialog" selectable: people needs this to copypaste the cockatrice version number in reports.

Note: current shortcut keys in use are not automatically overwritten to avoid headaches to existing users.

fix #1762
fix #1330

### DIFF
--- a/cockatrice/src/cardinfotext.cpp
+++ b/cockatrice/src/cardinfotext.cpp
@@ -13,19 +13,25 @@ CardInfoText::CardInfoText(QWidget *parent)
     nameLabel1 = new QLabel;
     nameLabel2 = new QLabel;
     nameLabel2->setWordWrap(true);
+    nameLabel2->setTextInteractionFlags(Qt::TextBrowserInteraction);
     manacostLabel1 = new QLabel;
     manacostLabel2 = new QLabel;
     manacostLabel2->setWordWrap(true);
+    manacostLabel2->setTextInteractionFlags(Qt::TextBrowserInteraction);
     colorLabel1 = new QLabel;
     colorLabel2 = new QLabel;
     colorLabel2->setWordWrap(true);
+    colorLabel2->setTextInteractionFlags(Qt::TextBrowserInteraction);
     cardtypeLabel1 = new QLabel;
     cardtypeLabel2 = new QLabel;
     cardtypeLabel2->setWordWrap(true);
+    cardtypeLabel2->setTextInteractionFlags(Qt::TextBrowserInteraction);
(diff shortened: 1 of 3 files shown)


ctrlaltca:
@LethalPigeon7 as you are the original reporter of this problem, can you please test if the problem is fixed for you?
Download: 32bit
https://ci.appveyor.com/api/buildjobs/wwg9836bqjq871qe/artifacts/Cockatrice-0.0.1~git20160809.1c78598-x86_qt5.exe
Download: 64bit
https://ci.appveyor.com/api/buildjobs/92cjjny0a2ms4pem/artifacts/Cockatrice-0.0.1~git20160809.1c78598-x86_64_qt5.exe

Please note that you'll have to "restore default shortcuts" from the settings or, if you want to avoid losing any custom shortcut key you specified, manually restore only these 2 shortcut keys:

schermata 2016-08-09 alle 11 48 29

tooomm:
fixes #1330

LethalPigeon7:
Works here. The cursor doesnt change into the usual "I" when hovering over selectable text in the about section but that is hardly an issue, Both the about section and deck editor are functioning the way I'd hope they would when I started nearly 4 years ago.
Never set any personal shortcuts, after updating the shift was already default in clipboard shortcuts.

PR #2126, opened 2016-08-10
Improve deck export to text files

### LINKED ISSUE #601: Add Deck Name and Comments to Save to/Load from clipboard functionality
Saving the deck to clipboard results in the main deck and the sideboard being copied, however we lose the deck name and comments in this process.

### LINKED ISSUE #725: Deck Editor "copy to clipboard" should sort and add comments
When I ctrl+c from the deck editor, I get my deck file in the order I added cards.

It would be nice if instead it sorted it according to the same rules as the display, adding in the `// Creature` comments and so on.

ctrlaltca:
Related to #601:
 * Deck name and comments are added as comments in the exported file
 * When loading a deck, treat the first comment line as deck name and the following ones (up to the first card) as comment lines; note that this far from being perfect (who knows what's written in the file comments and how). Imho it's better to risk loading something more and then let the user decide to delete it, than to risk not to load all the comments.

Related to #725
When exporting a deck:
 * group cards by zone and card type, sort them by name;
 * write zone names and card type names as comments;
 * count card number subtotals for each card type/zone and write them;

Fix #601
Fix #725

### DIFF
--- a/cockatrice/src/deck_loader.cpp
+++ b/cockatrice/src/deck_loader.cpp
@@ -3,6 +3,8 @@
 #include <QDebug>
 #include "deck_loader.h"
 #include "decklist.h"
+#include "carddatabase.h"
+#include "main.h"
 
 const QStringList DeckLoader::fileNameFilters = QStringList()
     << QObject::tr("Common deck formats (*.cod *.dec *.txt *.mwDeck)")
@@ -107,4 +109,90 @@ DeckLoader::FileFormat DeckLoader::getFormatFromName(const QString &fileName)
         return CockatriceFormat;
     }
     return PlainTextFormat;
+}
+
+bool DeckLoader::saveToStream_Plain(QTextStream &out)
+{
+    saveToStream_DeckHeader(out);
+
+    // loop zones
+    for (int i = 0; i < getRoot()->size(); i++) {
+        const InnerDecklistNode *zoneNode =
+            dynamic_cast<InnerDecklistNode *>(getRoot()->at(i));
+
+        saveToStream_DeckZone(out, zoneNode);
+
+        // end of zone
+        out << "\n";
+    }
+
+    return true;
+}
+
+void DeckLoader::saveToStream_DeckHeader(QTextStream &out)
+{
+    if(!getName().isEmpty())
+        out << "// " << getName() << "\n\n";
+    if(!getComments().isEmpty())
+    {
+        QStringList commentRows = getComments().split(QRegExp("\n|\r\n|\r"));
+        foreach(QString row, commentRows)
+            out << "// " << row << "\n";
+        out << "\n";
+    }
+}
+
+void DeckLoader::saveToStream_DeckZone(QTextStream &out, const InnerDecklistNode *zoneNode)
+{
+   // group cards by card type and count the subtotals
+    QMultiMap<QString, DecklistCardNode*> cardsByType;
+    QMap<QString, int> cardTotalByType;
+    int cardTotal = 0;
+
+    for (int j = 0; j < zoneNode->size(); j++) {
+        DecklistCardNode *card =
+            dynamic_cast<DecklistCardNode *>(
+                zoneNode->at(j)
+            );
+
+        CardInfo *info = db->getCard(card->getName());
+        QString cardType = info ? info->getMainCardType() : "unknown";
+
+        cardsByType.insert(cardType, card);
+
+        if(cardTotalByType.contains(cardType))
+            cardTotalByType[cardType] += card->getNumber();
+        else
+            cardTotalByType[cardType] = card->getNumber();
+
+        cardTotal += card->getNumber();
+    }
+
+    out << "// " << cardTotal << " " << zoneNode->getVisibleName() << "\n";
+
+    // print cards to stream
+    foreach(QString cardType, cardsByType.uniqueKeys())
+    {
+
+        out << "// " << cardTotalByType[cardType] << " " << cardType << "\n";
+        QList <DecklistCardNode*> cards = cardsByType.values(cardType);
+
+        saveToStream_DeckZoneCards(out, zoneNode, cards);
+
+        out << "\n";
+    }
+}
+
+void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out, const InnerDecklistNode *zoneNode, QList <DecklistCardNode*> cards)
+{
+    // QMultiMap sorts values in reverse order
+    for(int i = cards.size() - 1; i >= 0; --i)
+    {
+        DecklistCardNode* card = cards[i];
+
+        if (zoneNode->getName() == "side")
+            out << "SB: ";
+
+       out << card->getNumber() << " " << card->getName() << "\n";
+    }
 }
--- a/cockatrice/src/deck_loader.h
+++ b/cockatrice/src/deck_loader.h
@@ -28,6 +28,14 @@ public:
     bool loadFromFile(const QString &fileName, FileFormat fmt);
     bool loadFromRemote(const QString &nativeString, int remoteDeckId);
     bool saveToFile(const QString &fileName, FileFormat fmt);
+
+    // overload
+    bool saveToStream_Plain(QTextStream &out);
+
+protected:
+    void saveToStream_DeckHeader(QTextStream &out);
+    void saveToStream_DeckZone(QTextStream &out, const InnerDecklistNode *zoneNode);
+    void saveToStream_DeckZoneCards(QTextStream &out, const InnerDecklistNode *zoneNode, QList <DecklistCardNode*> cards);
 };
 
 #endif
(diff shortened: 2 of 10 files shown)


ZeldaZach:
I gave this an attempt and it seems nice, but I found a small few issues:

Deck -> Export -> Import
The comments seem to have picked up some additional info

```
Test comment
```

to 
```
Test comment
60 Maindeck
2 Artifact
````

Also, is there a way to export without these new comments? They're a nice addition for sure, but they kind of seem a bit cluttered. Maybe additional linebreaks would help?

```
// 25 Land
4 Evolving Wilds
2 Forest
2 Foul Orchard
2 Geier Reach Sanitarium
4 Hissing Quagmire
4 Llanowar Wastes
7 Swamp

// 1 Planeswalker
1 Ob Nixilis Reignited

// 11 Sorcery
1 Collective Brutality
2 Dark Petition
2 Languish
2 Read the Bones
2 Ruinous Path
1 Seasons Past
1 Transgress the Mind
```

**Comment by ZeldaZach** [comment]
One other note; tappedout.net doesn't seem to like the "comments" and they need to be completely removed to import a deck there. Deckstats doesn't seem to care, which is a good thing

**Comment by ctrlaltca** [comment]
> tappedout.net doesn't seem to like the "comments"

Yep, problems with tappedout already popped up in #1311. They don't really support importing decks from a file in a predetermined format, but only manually writing card names in text fields.
Since the site is commonly used amongst players, we could add a "upload to tappedout" feature like the one that's already in place for deckstats.net.

**Comment by ctrlaltca** [comment]
I've added the additional linebreaks, and the exported file looks nicer now.
About tappedout, i've added a new feature "analyze deck on tappedout", that is surely nicer that having to copy the deck, edit it (#1311) and paste the result on the website.

**Comment by tooomm** [comment]
sweet :smiley: 
tappedout and new menu works for me.

can't tell about export to txt feature

File: cockatrice/src/deck_loader.cpp [hunk 1]
```diff
@@ -108,3 +110,72 @@ DeckLoader::FileFormat DeckLoader::getFormatFromName(const QString &fileName)
...
+    if(!getComments().isEmpty())
+    {
+        QStringList commentRows = getComments().split(QRegExp("\n|\r\n|\r"));
+        foreach(QString row, commentRows)
+            out << "// " << row << "\n";
+        out << "\n";
+    }
+
+    // loop zones
+    for (int i = 0; i < getRoot()->size(); i++) {
```
Daenyth: Does the class support foreach-style iteration? It's nicer to read and less prone to indexing errors than manual looping. If it can be used with stl stuff it might be nice code-wise to break this up so it can be applied in a functional way by mapping over the sequence with `std::transform`

cockatrice/src/deck_loader.cpp
```diff
@@ -108,3 +110,72 @@ DeckLoader::FileFormat DeckLoader::getFormatFromName(const QString &fileName)
...
+            else
+                cardTotalByType[cardType] = card->getNumber();
+
+            cardTotal += card->getNumber();
+        }
+
+        out << "// " << cardTotal << " " << node->getVisibleName() << "\n";
+
+        // print cards to stream
+        foreach(QString cardType, cardsByType.uniqueKeys())
```
Daenyth: IMO it's easier to test/edit the code if we split side effects from computation. I'd prefer splitting this into two functions. One that just processes the deck into the structure needed for export and another that exports the text. We could then reuse the processing function for different styles of export - plain text, tappedout, deckstats, test code, etc.

cockatrice/src/tappedout_interface.h
```diff
@@ -0,0 +1,28 @@
...
+#include "carddatabase.h"
+#include "decklist.h"
+#include <QObject>
+
+class QByteArray;
+class QNetworkAccessManager;
+class QNetworkReply;
+class DeckList;
+
+class TappedOutInterface : public QObject {
```
Daenyth: Class doc comments here would be good for other readers to know that this exists in order to support the "export to tappedout" feature.

common/decklist.cpp
```diff
@@ -600,14 +610,15 @@ bool DeckList::loadFromFile_Plain(QIODevice *device)
...
 
-    WriteToStream(QTextStream &_stream) : stream(_stream) {}
+    WriteToStream(QTextStream &_stream, bool _bMarkSB) : stream(_stream), bMarkSB(_bMarkSB) {}
 
     void operator()(
         const InnerDecklistNode *node,
         const DecklistCardNode *card
     ) {
-       if (node->getName() == "side") {
+       if (bMarkSB && node->getName() == "side") {
```
Daenyth: what does `bMarkSB` mean? Can we use a more meaningful name?

Daenyth:
I'm really excited for these features!

cockatrice/src/deck_loader.cpp
ctrlaltca: I typically like to use `foreach`, too. In this case it's possible, but it makes the code even harder to understand because the object returned by getRoot() is both a `InnerDecklistNode*` and a `QList<AbstractDecklistNode *>`, and we need to iterate over the QList and then `dynamic_cast` each returned abstract object back into a `InnerDecklistNode*` in order to loop over its childs:

```
    // loop zones
    InnerDecklistNode *rootNode = getRoot();
    foreach(AbstractDecklistNode *abstractNode, *rootNode)
    {
        const InnerDecklistNode *node =
            dynamic_cast<InnerDecklistNode *>(abstractNode);
```

cockatrice/src/tappedout_interface.h
ctrlaltca: Right, added

common/decklist.cpp
ctrlaltca: Right, fixed

cockatrice/src/deck_loader.cpp
Daenyth: Maybe we can use law of demeter to lead here a bit? We could add a getCountByCardType method to the decklist class itself so the logic doesn't need to be inside this method?

cockatrice/src/deck_loader.h
```diff
@@ -28,6 +28,9 @@ class DeckLoader : public DeckList {
     bool loadFromFile(const QString &fileName, FileFormat fmt);
     bool loadFromRemote(const QString &nativeString, int remoteDeckId);
     bool saveToFile(const QString &fileName, FileFormat fmt);
+
+    // overload
+    bool saveToStream_Plain(QTextStream &stream);
```
Daenyth: Is `stream` an out param here? We should document that in the header here. CLion picks up javadoc-style comments.

cockatrice/src/deck_loader.cpp
```diff
@@ -113,7 +113,24 @@ DeckLoader::FileFormat DeckLoader::getFormatFromName(const QString &fileName)
 
 bool DeckLoader::saveToStream_Plain(QTextStream &out)
```
Daenyth: Really nice refactoring 👍

cockatrice/src/tappedout_interface.cpp
```diff
@@ -28,9 +28,17 @@ void TappedOutInterface::queryFinished(QNetworkReply *reply)
 
     if(reply->hasRawHeader("Location"))
```
Daenyth: Can we qInfo log the response info - status code, location header, anything else relevant? This is useful information to capture into the user-visible debug log.

cockatrice/src/tappedout_interface.cpp
```diff
@@ -28,9 +28,17 @@ void TappedOutInterface::queryFinished(QNetworkReply *reply)
...
     {
+        /*
+         * If the reply contains a "Location" header, a relative URL to the deck on TappedOut
+         * can be extracted from the header. The http status is a 302 "redirect".
+         */
         QString deckUrl = reply->rawHeader("Location");
         QDesktopServices::openUrl("http://tappedout.net" + deckUrl);        
     } else {
+        /*
+         * Otherwise, the deck has not been parsed correctly. Error messages can be extracted
```
Daenyth: Is extracting the messages a future improvement item? Maybe we can log the full text with qWarn for now so it's in user log?

cockatrice/src/tappedout_interface.h
```diff
@@ -10,6 +10,12 @@ class QNetworkAccessManager;
 class QNetworkReply;
 class DeckList;
 
+/**
+ * TappedOutInterface exists in order to support the "Analyze on TappedOut" feature.
+ * An http POST request is sent and the result is retrieved from the reply. Parsing
+ * logic is implemented in TappedOutInterface::queryFinished().
```
Daenyth: 👍 👍

cockatrice/src/tappedout_interface.cpp
ctrlaltca: Sure

cockatrice/src/tappedout_interface.cpp
ctrlaltca: > Is extracting the messages a future improvement item?

It' already implemented a few lines of code later. The error messages are extracted from the html and shown to the user in a messagebox

tooomm:
@ZeldaZach can you check the export/import again?

ZeldaZach:
Very good, i love it!

PR #2132, opened 2016-08-17
Optimize card move between main/side in deck editor

### LINKED ISSUE #2129: Smoother sideboard <-> main movement
<b>OS:</b>
Win 7 SP1
<b>Cockatrice version:</b>
1c78598 (2016-08-09)
___
When I press "S" in the deck editor to move a card to or from the sideboard, the deckbuilder highlights the card in the new area (the deck when moving from the sideboard and vice versa).
This has been massively annoying any time I've ever wanted to use it (mostly constructed deckbuilding), and has lately been especially painful when I started building sealed decks in cockatrice. When I want to move multiple copies of a card I have to press S, scroll down to where I was, click the card and repeat for every copy. Same goes for single copies of multiple cards. 
Ideally when a card is moved from one part to the other, the original location would stay highlighted, and should all copies be used the card above/below the original card would be highlighted now.

ctrlaltca:
When a card is moved between main and side in the deck editor using the "S" shortcut key, change the currently selected card "following" its movement only if it's the last copy of that card in that zone.

To explain it a bit simpler, if you have 4 copies of the same card in the "main" zone, pressing S will move a card to the "side" but keep the selection on the 3 cards that are still in the "main" zone.
Only when the last card is moved to the "side" zone, the selection will follow the card into the new zone.
 
fix #2129

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -823,7 +823,7 @@ void TabDeckEditor::actSwapCard()
 
     QModelIndex newCardIndex = deckModel->addCard(cardName, otherZoneName);
     recursiveExpand(newCardIndex);
-    deckView->setCurrentIndex(newCardIndex);
+
     setModified(true);
 }
 


tooomm:
Working as described. The focus doesn't change as long as there are more of the same card left.
Is the last card moved the focus follows the card to the other zone.
This should improve the current situation!

(while in #2129 the proposal was to always keep the focus in the same group)

ctrlaltca:
> (in #2129 the proposal was to always keep the focus in the same group - do you prefer it this way?)

It looks more coherent with the old behavior to me, but i'm open to discussion.

ctrlaltca:
@LethalPigeon7 do you want to try this? 

32bit: https://ci.appveyor.com/api/buildjobs/scji6x7o52kanc66/artifacts/Cockatrice-0.0.1~git20160817.109f6bb-x86_qt5.exe
64bit: https://ci.appveyor.com/api/buildjobs/jraoq6tjmdf6a9mw/artifacts/Cockatrice-0.0.1~git20160817.109f6bb-x86_64_qt5.exe

LethalPigeon7:
It's going to be equally unusable if it still jumps for no reason after a stack is done. I play limited and singleton formats, the change is irrelevant to me, sorry.

Daenyth:
@LethalPigeon7 can you provide a description of an interaction that would be useful for you please?

LethalPigeon7:
When the last copy from a stack is moved to the other zone (sb/mb) the deck editor would highlight either the card which was either below or above the newly missing stack. Here's a visual representation to ideally make this more understandable and realistically even more confusing.
1  A            1 A         [1 A]    /   /    1 A   
[2 B]    ->  [1 B]    -> 3 C    /or/    [3 C]
3  C            3 C                 /    /
[highlited card]

Daenyth:
That's what I was thinking. My preference would be to highlight the card below your last position on moving the last copy.

@ctrlaltca is that reasonable?

ctrlaltca:
PR updated, now the selection never moves anymore; Qt will automatically select the card that will take its place (row) in the list.

LethalPigeon7:
Is a version including this available? I'd love to test and mostly just reap the benefits ^^

tooomm:
@LethalPigeon7 
64bit: https://ci.appveyor.com/api/buildjobs/11228bgnqnobosdd/artifacts/Cockatrice-0.0.1~git20160820.a3a6065-x86_64_qt5.exe
32bit: https://ci.appveyor.com/api/buildjobs/47ij50l07gklqu82/artifacts/Cockatrice-0.0.1~git20160820.a3a6065-x86_qt5.exe

tooomm:
>Qt will automatically select the card that will take its place (row) in the list.

Working... BUT it doesn't always highlight the next card in the list, it highlights the next element in the list.
If it's the last creature that you move for example, it'll highlight the next element in the list which would be the group headline for instants for example.
Same applies to the bottom card of the deck - "Sideboard" is highlighted afterwards.

LethalPigeon7:
Sorry if I'm missing something, just trying to help, but shouldn't this and my original issue be closed now, seeing as the problem is resolved and the fix is working?

tooomm:
This is a `Pull Request`. A so called PR just adds new, experimental code which can be tested and reviewed by others.

It's still not finalized and got not "merged" with the main code yet. We don't work all week on things 
here, so they hang around for some days - don't worry. ;)

There is still the issue that I pointed out, while it might be an improvement to the current situation we try to think things through and decide on common ground what's best.
Once this is done, the code is considered good and the feature working as expected and hopefully bug free. Issues are open for discussion and for awareness of others until "PR's" with new code and features are `merged`.

See the `fix #2129` line in the start post? That indicates that this particular issue will automatically get closed once we decide this Pull Request is complete.
So wait a bit more, give the guys a bit more time to tweak their code and let others review+test it - then it'll end up in the main *master* branch of the project and will be automatically distributed as development version to everybody who wants to run the newest features. (You'll find the download link [here](https://github.com/Cockatrice/Cockatrice#downloads) )
What we gave you in here was a unofficial beta version IN development to get some early feedback. ;)

ctrlaltca:
> Working... BUT it doesn't always highlight the next card in the list, it highlights the next element in the list.

So, i've investigated this possibility, but in order to get the "next element" from the list i'd have to rewrite the way sorting the list works, and i'm too lazy to do that. You could just move the selection using pressing an arrow, even if it's not exactly a solution.

LethalPigeon7:
It would be really handy if the other elements (creatures, lands, sideboard) weren't highlightable. Still would hide things when clicked on - which I've never used in my life but I can see it being useful and it's hardly a downside - but when using arrows to navigate it wouldn't select them. If that worked, removing two copies of a card wouldn't highlight them either - another thing that's only caused me inconvenience throughout my years of brewing. And it would fix this.

Daenyth:
So if I understand correctly I think this is in good shape to be added. @LethalPigeon7 your suggestions are good but I think it's better for us to bring this into the next release and then edit it to incorporate your suggestions separately, rather than holding this off longer to accomodate them.

LethalPigeon7:
That sounds like the only truly logical procedure. I was mostly looking for feedback on my last comment, rather than its immediate implementation, seeing as I'm essentially asking for the removal of a feature someone presumably put in on purpose.

Daenyth:
I'd really appreciate if you file that as a specific separate ticket; I
don't think there's one single comment that sums it up anywhere here, and
I'm a bit too distracted to reread the entire conversation this moment to
figure it all out again. I think you have a better picture of it

LethalPigeon7:
I was specifically referring to the one comment you replied to, but I'll make a dupe of sorts.

ZeldaZach:
I took a look at the PR and some of the comments and I don't fully understand why this change is needed. If I have 4x a card in the main and I click "S" I personally would like to see one copy moved at a time. Can somebody help correct me if this doesn't make sense?

ctrlaltca:
> Can somebody help correct me if this doesn't make sense?

It's not about moving more than one card at once.
When a card gets moved, which card to focus/select on the list?
Old behavior: select the card that just moved in its new position
New behavior: keep the selection on the old card (if at least one copy is still present) or to a nearby card.

ZeldaZach:
Ahh, that makes sense. In that case I'm good with this

PR #2134, opened 2016-08-19
Renamed !setnumber! to !collectornumber!

### LINKED ISSUE #1992: bad name choice for !setnumber!
see: https://github.com/Cockatrice/Cockatrice/pull/1921#issuecomment-220785141

>I don't like the naming at all... it's very confusing!
Why don't we use cardnumber?
All placeholders with "set" are related to actual sets... this isn't. It's not a number of a set, it's a number of a card in a set.

ctrlaltca:
Fix #1992

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -220,7 +220,10 @@ QString PictureLoaderWorker::getPicUrl()
     picUrl.replace("!cardid!", QUrl::toPercentEncoding(QString::number(muid)));
     if (set)
     {
-        picUrl.replace("!setnumber!", QUrl::toPercentEncoding(card->getSetNumber(set->getShortName())));
+        // renamed from !setnumber! to !collectornumber! on 20160819. Remove the old one when convenient.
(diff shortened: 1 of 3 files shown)


cockatrice/src/pictureloader.cpp
```diff
@@ -220,7 +220,10 @@ QString PictureLoaderWorker::getPicUrl()
     picUrl.replace("!cardid!", QUrl::toPercentEncoding(QString::number(muid)));
     if (set)
     {
+        // renamed from !setnumber! to !collectornumber! on 20160819. Remove the old one when convenient.
         picUrl.replace("!setnumber!", QUrl::toPercentEncoding(card->getSetNumber(set->getShortName())));
+        picUrl.replace("!collectornumber!", QUrl::toPercentEncoding(card->getSetNumber(set->getShortName())));
+
```
tooomm: Can you rename `getSetNumber` as well?

ctrlaltca: Done

ZeldaZach:
I'm fine with this :shipit:

PR #2164, opened 2016-09-16
Don't use corrected name when downloading card

### LINKED ISSUE #2160: Split Card picture download without MUID
If a split card does not have a muid, it will attempt to download as "LeftRight" instead of "Left // Right"

Debug follows

Trying to load picture (set:  "DGM"  card:  "WearTear" )
Picture NOT found, trying to download (set:  "DGM"  card:  "WearTear" )
starting picture download: "Wear // Tear" Url: QUrl("http://gatherer.wizards.com/Handlers/Image.ashx?name=WearTear&type=card")
Picture downloaded, but blacklisted, will consider it as not found
Picture NOT found, download failed, no more sets to try: BAILING OUT (oldset:  "DGM"  card:  "WearTear" )

ctrlaltca:
PictureLoader is using the card's corrected name (stripped of problematic chars like `/`) to load and save the picture file to disk.
When downloading the card, it should use the card "normal" name and not the corrected name.

fix #2160


ZeldaZach:
Failed due to protobuff it seems. It looks fine to me

ctrlaltca:
Rebased to master after the googletest fix, hopefully CI should succeed this time.

ctrlaltca:
@tritoch please can you check if the problem reported in #2160 is solved with this patch?

tritoch:
@ctrlaltca I don't have the capability at the moment.  Simplest test is to swap the fallback URL ( http://gatherer.wizards.com/Handlers/Image.ashx?name=!name!&type=card ) into the primary URL, then look up a split card.  

Previous behavior is it would search 
`http://gatherer.wizards.com/Handlers/Image.ashx?name=FireIce&type=card` 
correct is 
`http://gatherer.wizards.com/Handlers/Image.ashx?name=Fire // Ice&type=card` (without hanging the picdownloader)

ZeldaZach:
Can this be updated? And I'll try it out then.

ctrlaltca:
Updated.

PR #2166, opened 2016-09-17
Fix appearance and theme settings

ZeppowN:
Hello, First time on Github and I'm not really sure how to use it but I recently picked up cockatrice for mtg online games and well,

I cant stand that they're theme options are not updated. I know there is a theme setting where you can change the table top but you should be able to load your own images. And more importantly, change the color of the interface in cockatrice, not just the theme of the board.

I would like to see a "night-time" option or something where it inverts the color of the background interface to black and all the text to white, I'm not sure if I'm requesting this right but If someone could help out that would be amazing! Thanks.


skwerlman:
It looks like you meant to open an issue, not a PR.
Go [here](https://github.com/Cockatrice/Cockatrice/issues/new) to open one.

ctrlaltca:
Also, we already implemented themes a few months ago.
Please have a look at https://github.com/Cockatrice/Cockatrice/wiki/Themes , you will also find some user-built themes at the end of the page, including a dark one.

Issue #2173, opened 2016-09-20
New public server; separate page for public servers?


skwerlman:
I've got two (related) requests regarding cockatrice.github.io.

---

- [x] I've recently decided to make my (very small) personal server open to the public. Could it please be added to the website?

website: none yet (update: skwerlman.github.io)
host: mtg.tetrarch.co
port: 4747
no webclient support (waiting for it to mature a bit before allowing it)
in-client registration only

---

- [x] Add my server to the in-client list

---

- [x] Maybe have a separate page for the server list? The current list only works for servers which have their own websites attached, and does not really provide much info about where each server is, or info about the communities they support. I think there's a lot of room for improvement there.

I suggest having a table with info on each public server, like so:

| Server (+ link to website) | Location | Language | Aprox. Users | Hostname | Port | Webclient Port | Allows Unreg |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Tetrarch | US East | English | 10 | mtg.tetrarch.co | 4747 | N/A | No |
| [Woogerworks](http://www.woogerworks.com/) | Somewhere (EU?) | English | A bunch | cockatrice.woogerworks.com | 4747 | 4748 | ?? |

tooomm:
We could have a wiki page and link to it from the io page :+1: 
So server admins could adjust and add servers on their own. We just need to watch out for mistakes!

skwerlman:
That's a good solution. 

I'd be happy to get the page started, though I don't know much about the servers currently listed.

tooomm:
Start with a temp page again maybe. We'll bring @woogerboy21, @poixen and @Lachee on board as well.

Not sure which information are really needed and which should be thought over. We can figure that out later. I've some design ideas that i'll work into once you set the page up! :+1:

skwerlman:
Here's the list with all the info I have: https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers

woogerboy21:
Is there a link to this off the main wiki page side?

tooomm:
Not yet, we are gathering information and ideas at the moment.
Does everybody likes the idea of having a link on the github pages site to such a server wiki side? (instead of having a long list there and the subscribed downsides by skwerlman in his starting post here)

@skwerlman Do you request to get added with your server to the client as well?

Daenyth:
I like the wiki idea.

I think if we list a server on there we should put it in the client also

skwerlman:
[...]

PR #2175, opened 2016-09-20
add mtg.tetrarch.co server

tooomm:
Added `mtg.tetrarch.co` server hosted by @skwerlman to the in-client default list

This deals with Nr.2 in https://github.com/Cockatrice/Cockatrice/issues/2173

I can see this staying around for while to checkout how things develop (and finally don't forget about adding a new stable server to the list). ;)


tooomm:
Sadly this wasn't enough to make your server show up in the list @skwerlman .
It might show up for first time users with no config at all, I couldn't test that for now.

Issue #2161, opened 2016-09-12
Kaladesh Aether card images not functioning


samgc3:
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
2e3966a (2016-04-16)

---

<br>

I noticed while using custom set XML for Kaladesh that cards having Aether in their name aren't functioning properly. It converts the name from Aether to AEther and won't return the card image. I believe this is due to Wizards changing Æ to Ae and the former having been specially coded into Cockatrice.

Here is sample XML:

```
    <card>
      <name>Aether Hub</name>
      <set picURL="http://media-dominaria.cursecdn.com/avatars/126/947/636089260525055620.png">KLD</set>
      <color></color>
      <manacost></manacost>
      <cmc>0</cmc>
      <type>Land</type>
      <pt></pt>
      <tablerow>2</tablerow>
      <text>When Aether Hub enters the battlefield, you get e (an energy counter).

t: Add C to your mana pool.

t, Pay e: Add one mana of any color to your mana pool.</text>
    </card>
```

Daenyth:
Did oracle generate this xml or did you get it from somewhere? Oracle doesn't attach `picURL` any more, so I'm not clear.

Can you also attach the debug log from a session with the failure you're seeing?

samgc3:
I found that the cards function properly when you magnify them, but you cannot see the image in deck editor, hand, or on the battlefield. 

Someone created the XML, found here: http://www.mtgsalvation.com/forums/magic-fundamentals/magic-software/third-party-programs/754885-kaladesh-patch-for-cockatrice

samgc3:
Here's a sample of what I'm seeing. I'm not sure how to get the debug log.

http://imgur.com/a/SPJam

Daenyth:
From the application menu `Help > View debug log`

samgc3:
The issue seems to be that when loading a deck from a text file names with Aether are brought in as AEther. I can work around the issue by loading from a .cod instead.

Daenyth:
Generally speaking, .dec/.txt imports are for compatibility only and won't always be 100%. This is one thing it does for compatibility with some older cards, but we can probably remove it at this point.

If you have a deck in txt format, you can use the "load deck from clipboard" feature and save as .cod.

Jinxstands4Jinx:
You could use the Custom images Folder to overwrite this by saving the correct picture in there named Aether.

woogerboy21:
[...]

PR #2182, opened 2016-09-22
No "Ae" --> "AE" translate

tritoch:
Respect the 'Ae'

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -579,8 +579,6 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         cardName.replace(rx, "'");
         rx.setPattern("Æ");
         cardName.replace(rx, "AE");
-        rx.setPattern("^Aether");
-        cardName.replace(rx, "AEther");
         rx.setPattern("\\s*[|/]{1,2}\\s*");
         cardName.replace(rx, " // ");
 


Daenyth:
This should also fix #2161

common/decklist.cpp
```diff
@@ -579,8 +579,6 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         cardName.replace(rx, "'");
         rx.setPattern("Æ");
         cardName.replace(rx, "AE");
```
Psithief: Shouldn't this be `Ae` as well?

Daenyth: I'd prefer to leave that until we know it's an issue - this should only affect much older cards

Psithief: I suppose we need to know what the data source (mtgjson) plans to do about this first.

tooomm: I thought there are supposed to be no more `Æ` cards at all starting with Kaladesh?
(besides the ordinary errors on gatherer for sure...)

tritoch:
>no more AE cards on gatherer

That's correct (or should/will be) but this replacement is for deck editor.  It needs to handle people who made deck lists before gatherer was fixed or from other programs which may not have been updated.  The other replace should be 'Ae' I will update that asap.

tooomm:
Can you give a bit more details what your change is about in the PR next time, please?

Whats the problem, what does it fix, why is it needed... just like this: :P
>(...) **but this replacement is for deck editor. It needs to handle people who made deck lists before gatherer was fixed or from other programs which may not have been updated.**

ZeldaZach:
https://www.youtube.com/watch?v=1umfD-k2lww&feature=youtu.be was fixed with this PR, i assume?

PR #2185, opened 2016-09-23
Pending mtgjson/gatherer Æ-->Ae conversion

tritoch:
This is to prepare for when mtgjson converts all Æ cards to 'Ae'.  Formerly it was 'AE'.

I believe this portion of the code deals with the deck importer.  If a file to be imported and converted (read: not .cod) via clipboard or plaintext file has an Æ symbol in it, currently that is coverted to `AE` when they card is added to your Cockatrice deck editor list.  This will switch it to an `Ae` conversion if mtgjson moves forward with such a revision.

Do not merge until mtgjson updates to use `Ae` instead of `AE`

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -578,7 +578,7 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         rx.setPattern("’");
         cardName.replace(rx, "'");
         rx.setPattern("Æ");
-        cardName.replace(rx, "AE");
+        cardName.replace(rx, "Ae");
         rx.setPattern("\\s*[|/]{1,2}\\s*");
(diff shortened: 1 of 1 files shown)


ZeldaZach:
I tested this, but I don't think we need it as caps don't really matter, at least when using Gatherer....

Daenyth:
Casing matters for files on disk at minimum

ZeldaZach:
https://www.youtube.com/watch?v=1umfD-k2lww&feature=youtu.be gets fixed with this, right?

Psithief:
@ZeldaZach Actually I believe the previous merge about Æther fixes that. 

_This_ fixes card names with a ligature people might find on the Web and paste into Cockatrice/load from text format file.

PR #2194, opened 2016-09-28
[german-translation-fixes] Use formal you for consistency

Cat-Ion:
All the other translations already do, so I changed this one.

### DIFF
--- a/oracle/translations/oracle_de.ts
+++ b/oracle/translations/oracle_de.ts
@@ -134,7 +134,7 @@ Sie müssen eine URL oder einen Dateinamen als Quelle angeben.</translation>
     <message>
         <location filename="../src/oraclewizard.cpp" line="430"/>
         <source>Do you want to try to download a fresh copy of the uncompressed file instead?</source>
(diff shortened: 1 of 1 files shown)


ZeldaZach:
For translations we use transifex. Why don't you go commit the change there so we can implement it here :)

Daenyth:
@Cat-Ion Thank you for the contribution! As Zach mentioned, we manage the translations on Transifex rather than directly committing them with git.

Here's our landing page for translators: https://github.com/Cockatrice/Cockatrice/wiki/Translation-FAQ

It links to some Transifex documentation about getting started there. If anything is unclear we'd be happy to help you along with it.

Daenyth:
Specifically I think you want to be here: https://www.transifex.com/cockatrice/cockatrice/translate/#de/oracle/40511209

Cat-Ion:
Yup - I went to the wrong translations first (found two strings to correct there, though), and when I saw the one I'd created the pull request for it was already changed. Thanks :)

ZeldaZach:
Thanks for your help @Cat-Ion! Glad to see new faces around here :)

PR #2195, opened 2016-09-29
Prevent login error handler being incorrectly called

### LINKED ISSUE #2176: Server Full Multiple message box problems
When a server has reached its capacity the client shows 2 boxes.  The top box displays the proper message about the server being at capacity but the bow below it shows the unknown error -1 message box.  

Solution should be as simple as adding a case statement for the id in https://github.com/Cockatrice/Cockatrice/blob/da2eb97c7b587b1b49bc1d93d4d9f717eb4ec275/cockatrice/src/window_main.cpp#L326 but should be investigated to be certain.

wincinderith:
Fix #2176.

In some situations, when the server disconnects the client during a login,
multiple dialogs will be shown. This is undesirable behavior, and hence
`loginError` should only be called when the client is actually connected
(`response.response_code() != Response::RespNotConnected`).

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -152,7 +152,7 @@ void RemoteClient::loginResponse(const Response &response)
         if (resp.missing_features_size() > 0 && settingsCache->getNotifyAboutUpdates())
                 emit notifyUserAboutUpdate();
 
-    } else {
+    } else if (response.response_code() != Response::RespNotConnected) {
         QList<QString> missingFeatures;
         if (resp.missing_features_size() > 0) {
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Hey @wincinderith Thanks for your contribution!! I'll see what I can do to test this for you and get it into the main repo :)

woogerboy21:
Anyone actually test this change yet?

woogerboy21:
Looks like the idle client timeout PR that was merged caused a simple conflict that needs resolved before merge.  Other than that the change actually works as described.

wincinderith:
@woogerboy21 Conflicts have been resolved.

PR #2199, opened 2016-10-03
Disable "Load remote deck..." button in local game

### LINKED ISSUE #2109: `load remote deck` should be disabled in local game mode
The button should be greyed out. Indicating that you're playing a local game. ^^

wincinderith:
Fix #2109.

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -309,6 +309,7 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor, GameReplay *_replay)
           secondsElapsed(0),
           hostId(-1),
           localPlayerId(-1),
+          isLocalGame(_tabSupervisor->getIsLocalGame()),
           spectator(true),
           gameStateKnown(false),
           resuming(false),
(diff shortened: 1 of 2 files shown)


cockatrice/src/tab_game.cpp
```diff
@@ -100,6 +100,8 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
 {
     loadLocalButton = new QPushButton;
     loadRemoteButton = new QPushButton;
+    if (parentGame->getIsLocalGame())
```
Daenyth: Can you name this `isLocalGame`? `getIsLocalGame` is unnecessary.

cockatrice/src/tab_game.h
```diff
@@ -111,6 +111,7 @@ class TabGame : public Tab {
     QMap<int, QString> roomGameTypes;
     int hostId;
     int localPlayerId;
+    bool isLocalGame;
```
Daenyth: This can be const since it won't ever change while the tab is running.

cockatrice/src/tab_game.cpp
wincinderith: I was following the same naming scheme as in the `TabSupervisor` class to avoid confusion. We should probably keep the naming scheme as consistent as possible.

cockatrice/src/tab_game.h
wincinderith: [Done](https://github.com/Cockatrice/Cockatrice/pull/2199/commits/2376a7dbc01aeabef6f3b4e723f27225c5cf740c).

Daenyth:
I think the OSX build failure is not from your changes.

Daenyth:
@wincinderith Thank you for the contribution! :D

PR #2200, opened 2016-10-03
Disallow usernames that contain certain words and RegExp

### LINKED ISSUE #2135: As a server operator I would like user registration name filtering
## Description
As a server operator I would like to prevent the use of certain words in username by adding a filter. This filter can include words or special characters. This will prevent the creation of usernames not deemed appropriate for the server operator.

## Notes
A similar feature is already in place. Usernames can not have special chars that have been disallowed in the server conf.

wincinderith:
Fix #2135.

Check if any of the words in `disallowedwords` are contained in the username. If
so, return false like other checks.

NOTE: Needs testing.

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -404,7 +404,7 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 {
     QString out = tr("Invalid username.") + "<br/>";
     QStringList rules = in.split(QChar('|'));
-    if (rules.size() == 9)
+    if (rules.size() == 7 || rules.size() == 9)
     {
         out += tr("Your username must respect these rules:") + "<ul>";
 
@@ -418,11 +418,14 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 
         out += "<li>" + tr("first character can %1 be a punctuation mark").arg((rules.at(5).toInt() > 0) ? "" : tr("NOT")) + "</li>";
 
-        if (rules.at(7).size() > 0)
-            out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
+        if (rules.size() == 9)
+        {
+            if (rules.at(7).size() > 0)
+                out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
 
-        if (rules.at(8).size() > 0)
-            out += "<li>" + tr("can not match any of the following expressions: %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
+            if (rules.at(8).size() > 0)
+                out += "<li>" + tr("can not match any of the following expressions: %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
+        }
 
         out += "</ul>";
     }
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -407,11 +407,11 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 
         out += "<li>" + tr("first character can %1 be a punctuation mark").arg((rules.at(5).toInt() > 0) ? "" : tr("NOT")) + "</li>";
 
-		if (rules.at(7).size() > 0)
-			out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
+        if (rules.at(7).size() > 0)
+            out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
 
-		if (rules.at(8).size() > 0)
-			out += "<li>" + tr("can not match any of the following expressions: %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
+        if (rules.at(8).size() > 0)
+            out += "<li>" + tr("can not match any of the following expressions: %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
 
         out += "</ul>";
     }
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -393,7 +393,7 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 {
     QString out = tr("Invalid username.") + "<br/>";
     QStringList rules = in.split(QChar('|'));
-    if (rules.size() == 8)
+    if (rules.size() == 9)
     {
         out += tr("Your username must respect these rules:") + "<ul>";
 
@@ -409,6 +409,9 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 
 		if (rules.at(7).size() > 0)
 			out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
+
+		if (rules.at(8).size() > 0)
+			out += "<li>" + tr("can not match any of the following expressions: %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
 
         out += "</ul>";
     }
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -393,9 +393,9 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 {
     QString out = tr("Invalid username.") + "<br/>";
     QStringList rules = in.split(QChar('|'));
-    if (rules.size() == 7)
+    if (rules.size() == 8)
     {
-        out += tr("Your username must respect these rules:") + "<br><ul>";
+        out += tr("Your username must respect these rules:") + "<ul>";
 
         out += "<li>" + tr("is %1 - %2 characters long").arg(rules.at(0)).arg(rules.at(1)) + "</li>";
         out += "<li>" + tr("can %1 contain lowercase characters").arg((rules.at(2).toInt() > 0) ? "" : tr("NOT")) + "</li>";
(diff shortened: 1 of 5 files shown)


servatrice/servatrice.ini.example
```diff
@@ -107,6 +107,9 @@ allowedpunctuation=_.-
 ; If a username can begin with punctuation defined in allowedpunctuation
 allowpunctuationprefix=false
 
+; Disallow usernames containing these words
+disallowedwords="admin,administrator"
```
Nightfirecat: Wouldn't `disallowedwords="admin"` be identical to this line? (since "administrator" contains "admin" anyway)

wincinderith:
Extra little bit of info: I noticed the `Servatrice_DatabaseInterface::usernameIsValid` function was only called when using an SQL backend. Is this desirable? I think the username checks should be performed regardless.

servatrice/servatrice.ini.example
wincinderith: Ah yes, true. Totally flew past me. :laughing:

wincinderith: Should there be a default other than admin that I should add while I'm fixing that up?

tooomm: >Wouldn't disallowedwords="admin" be identical to this line? 

No adminimi, adminora... ?

wincinderith: They also contain the substring "admin"...

skwerlman:
It'd be nice if there were PCRE (or similar) support here.
That way I could block names like "user[0-9]+", but not block "confuser".

wincinderith:
I can look into implementing PCRE, however I don't know how the configuration file would deal with them if a comma was inside an expression. Maybe an escape sequence?

OK, I might open a separate PR for that, since it's a different feature. I'll keep this thread posted.

tooomm:
Is there some kind of useful error message for the user?

wincinderith:
@tooomm Not yet, working on it though.

I'll put this PR on hold until I get the username check working regardless of the backend.

wincinderith:
UPDATE: There is no simple way to perform the username check outside of registration. The original issue, #2135, only mentioned registration anyway, so maybe it's not worth the trouble.

wincinderith:
@skwerlman I have implemented RegExp in a new setting in `servatrice.ini`, called `disallowregexp`.

wincinderith:
@tooomm I have implemented a setting for matching PCRE. See the option `disallowedregexp` in `servatrice.ini` for docs.

Daenyth:
Actually I think PCRE support is a bad idea. PCRE uses backtracking, which means that a user can DOS the server by sending pathological usernames to the regex engine, causing exponential runtime for the regex check, and probably thrashing the cpu of the host. We should just do string contains matching.

servatrice/servatrice.ini.example
```diff
@@ -107,6 +107,14 @@ allowedpunctuation=_.-
 ; If a username can begin with punctuation defined in allowedpunctuation
 allowpunctuationprefix=false
 
+; Disallow usernames containing these words. This list is comma seperated
+disallowedwords="admin"
+
+; Disallow usernames matching these regular expressions. This list is comma
+; separated, hence you cannot use commas in your expressions. Backslashes must
+; be escaped, so `\w+\d+` becomes `\\w+\\d+`
+disallowedregexp=""
```
Daenyth: Let's remove this for the reasons I mentioned in the pr comment.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -132,14 +132,27 @@ bool Servatrice_DatabaseInterface::usernameIsValid(const QString &user, QString
...
 
     if (!allowPunctuationPrefix && allowedPunctuation.contains(user.at(0)))
         return false;
 
+    for (const QString &word : disallowedWords) {
+        if (user.contains(word, Qt::CaseInsensitive)) return false;
+    }
+
+    for (const QString &regExp : disallowedRegExp) {
+        if (QRegExp(regExp).exactMatch(user)) return false;
```
Daenyth: If we were going to keep the regex support in, we should store compiled regex objects rather than strings. You'd also want to log invalid-regex failures. It's a moot point though since we should take it out

skwerlman:
From what I've found, DOSing requires the attacker to be able to configure the regexp used by the server, unless the attacker already knows the regexp used to be vulnerable. Avoiding bad regexps, IMO, should be the responsibility of the server op.
For simple regexps like those likely to be used here, catastrophic backtracking is not likely to be an issue, anyway, since it (AFAIK) requires the regexp to contain nested quantifiers.

https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS
http://www.regular-expressions.info/catastrophic.html

wincinderith:
@Daenyth I can see why this would be an extremely useful feature though. Perhaps we can work around the issues with it?

wincinderith:
@skwerlman Brings up some good points. Perhaps compiling the QRegExp on runtime would be the entire fix then.

wincinderith:
@Daenyth How should I proceed?

Daenyth:
It's a good point that it's on the operator to keep the regex safe. I'd like a warning to that effect in the config. @woogerboy21 do you think the regex filter is worth it?

wincinderith:
Any tips on what exactly the warning should say?

woogerboy21:
@Daenyth If your asking me about if regex would be advantageous, sure it has its advantages but I doubt I would use it (maybe) however I am not the only server operator.  If your asking my opinion about if it should be the server owners responsibility for keeping it safe, my answer would be yes.

As far as what the config warning should have in it.  I would recommend that there be as detailed an explanation as possible with given examples and mentioning of what negative things could occur if rules are not followed. Not to the level of teaching an idiot something as I am a firm believer if you want to operate something you should understand how it works before doing so.   

 From my experience alot of people are gaining interest in standing up servers (mostly for personal play groups) but very few want to invest any time into understanding how the underlying code that makes functionality work resulting in potential for some pretty bad scenario's in situations like this.  Though most private type servers probably won't ever have to worry about being attacked.

cockatrice/src/window_main.cpp
```diff
@@ -393,9 +393,9 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 {
     QString out = tr("Invalid username.") + "<br/>";
     QStringList rules = in.split(QChar('|'));
-    if (rules.size() == 7)
+    if (rules.size() == 9)
```
ctrlaltca: It would be nice to make this more backwards compatible, so that when a client that includes this change connects to a server it would be able to handle both the old `rules.size() == 7` (old server not including this change) and the new `rules.size() == 9` (new server including this change).
You could just use to nested `if`s or a switch with fallthrough.

servatrice/servatrice.ini.example
```diff
@@ -107,6 +107,14 @@ allowedpunctuation=_.-
 ; If a username can begin with punctuation defined in allowedpunctuation
 allowpunctuationprefix=false
 
+; Disallow usernames containing these words. This list is comma seperated
+disallowedwords="admin"
```
ctrlaltca: I'd add a clearer example of the meaning of "containing", eg. that "ass" matches "grasshopper"

servatrice/servatrice.ini.example
ctrlaltca: I'd keep it, just adding a warning that adding a lot of regular expressions or some particularly complex regexp could lead to resource exhaustion on the server.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -132,14 +132,27 @@ bool Servatrice_DatabaseInterface::usernameIsValid(const QString &user, QString
     bool allowNumerics = settingsCache->value("users/allownumerics", true).toBool();
     bool allowPunctuationPrefix = settingsCache->value("users/allowpunctuationprefix", false).toBool();
     QString allowedPunctuation = settingsCache->value("users/allowedpunctuation", "_").toString();
-    error = QString("%1|%2|%3|%4|%5|%6|%7").arg(minNameLength).arg(maxNameLength).arg(allowLowercase).arg(allowUppercase).arg(allowNumerics).arg(allowPunctuationPrefix).arg(allowedPunctuation);
+	QString disallowedWordsStr = settingsCache->value("users/disallowedwords", "").toString();
+    QStringList disallowedWords = disallowedWordsStr.split(",", QString::SkipEmptyParts);
+    disallowedWords.removeDuplicates();
+	QString disallowedRegExpStr = settingsCache->value("users/disallowedregexp", "").toString();
+
+    error = QString("%1|%2|%3|%4|%5|%6|%7|%8|%9").arg(minNameLength).arg(maxNameLength).arg(allowLowercase).arg(allowUppercase).arg(allowNumerics).arg(allowPunctuationPrefix).arg(allowedPunctuation).arg(disallowedWordsStr).arg(disallowedRegExpStr);
```
ctrlaltca: This pipe-joined string sucks, my bad for having introduced it in the code some time ago. We'll have to keep it like this anyway..

servatrice/servatrice.ini.example
ctrlaltca: Also, please add an example for multiple items: is it `"word1", "word2"` or `"word1, word2"` ? As an user i'd have to look it up or test it to be sure, better to make it clear in the comment.

wincinderith:
@Daenyth @ctrlaltca All (current) issues have been rectified. Any final changes?

wincinderith:
@ctrlaltca Isn't there an option to squash and merge?

wincinderith:
https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits

ctrlaltca:
Re-tested changes one more time, everything's working fine: Merging.

ctrlaltca:
@wincinderith thank you for your contribution!

wincinderith:
@ctrlaltca Not a problem! Glad I could contribute to such a great project.

ZeldaZach:
Thanks for your help @wincinderith! I hope you stick around. If you'd ever like to chat with us, you can always find a member of the dev team online at gitter.im/cockatrice/cockatrice

PR #2209, opened 2016-10-09
Idle Client Timeout Functionality

### LINKED ISSUE #2208: Server side Inactivity logout
It would nice to see a feature (server side) that the client be logged out of a server after a set period of inactivity.  My initial thoughts were a client side set of code that does a simple counter that counts down and is reset in the event of an action is taken.  If the timer ever reaches 0, the client logs out.  Initial client side setting for timer is something high, maybe 2 hours.  

Anyone else have idea's or thoughts on something like this?

Edit:
First few of actions I can think of that would reset the timer would be:
- Send Chat
- Join Room
- Join Game
- Game Action (shuffle, draw etc)

woogerboy21:
Fix #2208 

Added the functionality client side to log users out of servers if they
are idle for more than 1 hour without joining either a game or room.
Sending a message (room/game/private). Pushing the play/pause button on a replay or performing a game action.

### DIFF
--- a/common/featureset.cpp
+++ b/common/featureset.cpp
@@ -20,6 +20,7 @@ void FeatureSet::initalizeFeatureList(QMap<QString, bool> &featureList) {
     featureList.insert("room_chat_history", false);
     featureList.insert("client_warnings", false);
     featureList.insert("mod_log_lookup", false);
+    featureList.insert("client_inactivetimeout", false);
 }
 
 void FeatureSet::enableRequiredFeature(QMap<QString, bool> &featureList, QString featureName){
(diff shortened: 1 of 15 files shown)


common/featureset.cpp
```diff
@@ -20,6 +20,7 @@ void FeatureSet::initalizeFeatureList(QMap<QString, bool> &featureList) {
     featureList.insert("room_chat_history", false);
     featureList.insert("client_warnings", false);
     featureList.insert("mod_log_lookup", false);
+    featureList.insert("client_inactivetimeout", false);
```
Daenyth: 👎  this feature isn't related to client-server communication so there's no need for the server to know about it.

woogerboy21:
I could see an argument to this.  As a server operator I would be interested in requiring clients to include the functionality to timeout if they are inactive.  But your statement is true as well.

PR #2213, opened 2016-10-11
[WIP] Add donator recognition

woogerboy21:
Added server side functionality to allow recognition of user accounts
that have donated. If the user is marked as a donator they are granted
server access regardless of maximum user settings.  In doing so I have
moved the max user check out of the server socket interface and moved it
into the protocol handler (were it should live honestly) and also
created a response to the server login request instead of triggering an
event for the server capacity being reached.

### DIFF
--- a/common/featureset.cpp
+++ b/common/featureset.cpp
@@ -21,6 +21,7 @@ void FeatureSet::initalizeFeatureList(QMap<QString, bool> &featureList) {
     featureList.insert("client_warnings", false);
     featureList.insert("mod_log_lookup", false);
     featureList.insert("client_inactivetimeout", false);
+    featureList.insert("donator_recognition", false);
 }
 
 void FeatureSet::enableRequiredFeature(QMap<QString, bool> &featureList, QString featureName){
--- a/common/pb/response.proto
+++ b/common/pb/response.proto
@@ -37,6 +37,7 @@ message Response {
         RespRegistrationAcceptedNeedsActivation = 33; // Server accepted cient registration, but it will need token activation
         RespClientIdRequired = 34; // Server requires client to generate and send its client id before allowing access
         RespClientUpdateRequired = 35; // Client is missing features that the server is requiring
+        RespUserLimitReached = 36; // Server has a user limit and it has been reached
     }
     enum ResponseType {
         JOIN_ROOM = 1000;
--- a/common/server.cpp
+++ b/common/server.cpp
@@ -84,7 +84,7 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString 
         name = name.left(35);
 
     Server_DatabaseInterface *databaseInterface = getDatabaseInterface();
-
+    
     AuthenticationResult authState = databaseInterface->checkUserPassword(session, name, password, clientid, reasonStr, secondsLeft);
     if (authState == NotLoggedIn || authState == UserIsBanned || authState == UsernameInvalid || authState == UserIsInactive)
         return authState;
@@ -92,8 +92,20 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString 
     ServerInfo_User data = databaseInterface->getUserData(name, true);
     data.set_address(session->getAddress().toStdString());
     name = QString::fromStdString(data.name()); // Compensate for case indifference
-
+    
     if (authState == PasswordRight) {
+        //check to see if user is full but allow users that donate access to server regardless of user count
+        if (!databaseInterface->isUserADonator(name)) {
+            //limit the number of total users based on configuration settings
+            if (getMaxUserLimitEnabled()) {
+                int userLimit = getMaxUsers();
+                int playerCount = (databaseInterface->getActiveUserCount() + 1);
+                if (playerCount > userLimit) {
+                    return UserLimitReached;
+                }
+            }
+        }
+
         if (users.contains(name) || databaseInterface->userSessionExists(name)) {
             if (users.contains(name)) {
                 qDebug("Session already logged in, logging old session out");
@@ -167,7 +179,7 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString 
     se = Server_ProtocolHandler::prepareSessionEvent(event);
     sendIsl_SessionEvent(*se);
     delete se;
-
+    
     return authState;
 }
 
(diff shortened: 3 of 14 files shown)


skwerlman:
Do admins/mods bypass the server limit?

woogerboy21:
If you mark them as a donator they would.  As of right now no one does regardless of rank.  Keep in mind donation can come in many forms.  It isn't just monitory .  Moderators for example are donating there time.

skwerlman:
In that case, could we call them "VIP" or similar? Marking myself as "Donator" feels odd.

I'm for this feature either way.

common/featureset.cpp
```diff
@@ -21,6 +21,7 @@ void FeatureSet::initalizeFeatureList(QMap<QString, bool> &featureList) {
     featureList.insert("client_warnings", false);
     featureList.insert("mod_log_lookup", false);
     featureList.insert("client_inactivetimeout", false);
+    featureList.insert("donator_recognition", false);
```
Daenyth: Does the client need to care about this?

common/pb/response.proto
```diff
@@ -37,6 +37,7 @@ message Response {
         RespRegistrationAcceptedNeedsActivation = 33; // Server accepted cient registration, but it will need token activation
         RespClientIdRequired = 34; // Server requires client to generate and send its client id before allowing access
         RespClientUpdateRequired = 35; // Client is missing features that the server is requiring
+        RespUserLimitReached = 36; // Server has a user limit and it has been reached
```
Daenyth: This really wasn't here already? Damn...

common/server.cpp
```diff
@@ -84,16 +84,28 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session, QString
...
         return authState;
 
     ServerInfo_User data = databaseInterface->getUserData(name, true);
     data.set_address(session->getAddress().toStdString());
     name = QString::fromStdString(data.name()); // Compensate for case indifference
-
+    
     if (authState == PasswordRight) {
+        //check to see if user is full but allow users that donate access to server regardless of user count
+        if (!databaseInterface->isUserADonator(name)) {
```
Daenyth: In the code I'd prefer to keep it agnostic - `isUserAllowedToBypassUserCountLimit` or something. I know it's verbose but I don't think the game server should care about donations. We can leave it to each server operator to determine how users are granted this bypass permission.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1091,3 +1091,23 @@ QList<ServerInfo_ChatMessage> Servatrice_DatabaseInterface::getMessageLogHistory
 
     return results;
 }
+
+bool Servatrice_DatabaseInterface::isUserADonator(const QString &user)
+{
+    if (!checkSql())
+        return false;
+    
+    QSqlQuery *query = prepareQuery("SELECT donator from {prefix}_users WHERE name = :user_name");
```
Daenyth: This is likely the wrong place to do it. Isn't there already a SELECT on _users when a person joins the server? Add the bypass column to that select instead of doing an extra query. It will be much more efficient in terms of db load.

servatrice/migrations/servatrice_0017_to_0018.sql
```diff
@@ -0,0 +1,5 @@
+-- Servatrice db migration from version 17 to version 18
+
+alter table cockatrice_users add `donator` tinyint(1) not null default 0;
```
Daenyth: Also here, `allow_max_user_bypass` or something instead.

common/featureset.cpp
Daenyth: I'm still not convinced it does - the worst that happens is that the client will get a 'user limit reached' rejection and be unable to properly display it. But that code path now has a bit of text saying something like 'this probably means your client is out of date and needs to be updated'

common/pb/response.proto
woogerboy21: Nope, in the past the lookup was done in the session creation step and thus an event was generated. The session creation happens before the actual user information checks occur and a response to login is not generated but an event is returned.  Having written the original code for this check and coming back to it later I see how implementing it in the portion of the code that does the login code portion makes more since than in the session creation portion.

common/server.cpp
woogerboy21: Im not so sure this is the correct solution.  The information being queried is then action-ed on.  The action to take shouldn't be what is queried from the DB.  There might be other things the information could be used for else were in the code (not really sure just a thought). Now we could say something like isUserAVIP as suggested by @skwerlman if the word "Donator" seems odd to everyone.

The general concept of the functionality is to mark a user as "important" but not "important enough" to create a new user role.  

@Daenyth 
Thoughts?

servatrice/src/servatrice_database_interface.cpp
woogerboy21: The query is to identify if the user is flagged as a "donator" (the naming could be changed).  The query is to be used to take some sort of action on the information.  It just so happens in this change it is for granting access to the user in the event the server is full.   Nothing is to say the information couldnt be used else were for other reasons.

The current way the login works:
connection is made -> db queried for capacity -> db is queried for password correct -> return response.

This change make it looks something like this:
connection is made -> session is created -> db is queried for user password correct -> db is queried for donator status -> db is queried for capacity -> return response. 

If you want I could probably combine the "db queried for password correctness & db is queried for donator status" and create a new AuthenticationResults and then action against that.  It would probably reduce the amount of queries by 1 (haven't looked so not sure).  Ill take a look.

servatrice/migrations/servatrice_0017_to_0018.sql
woogerboy21: see: https://github.com/Cockatrice/Cockatrice/pull/2213#discussion_r82812854

woogerboy21:
@skwerlman I can change the name to whatever, The underlying code functionality is the same :smile:

servatrice/src/servatrice_database_interface.cpp
Daenyth: That's not quite what I meant. I'm on my phone at the moment so it's hard
to type. I'll ping you on chat tonight and try to make my point more clear

common/featureset.cpp
ctrlaltca: As long as it's not a required feature i don't see a problem with this.

woogerboy21: Of course , that is unless we intend on creating a new user level with this type of change.  I don't think I would recommend that but maybe then it should be in the featureset list?

woogerboy21:
Per some conversations about the functionality of this PR I will be closing it.  Then opening a couple new PRs to separate out the two major things that it accomplishes.

Daenyth:
Sounds exciting!

PR #2215, opened 2016-10-11
Show card sets

### LINKED ISSUE #2210: Put the sets in the card description
I would like the card sets in the card description. 

Example, player plays a card you suspect to not be in a format. You can check it printings without the need to change tabs.

ZeldaZach:
Fix #2210. 

This PR will show the sets each card is in, according to the import rules from Oracle.

This change appears when you're using the deck editor, when a card is linked in chat via the [[ cardName ]] tag, and in game on the side.

It does slightly cut off some of the cards that have been in tons of sets, but I don't see much of an issue with that (ex: Basic Lands)

screen shot 2016-10-11 at 5 47 33 pm
screen shot 2016-10-11 at 5 47 43 pm
screen shot 2016-10-11 at 5 47 48 pm
screen shot 2016-10-11 at 5 48 09 pm

### DIFF
--- a/cockatrice/src/cardinfotext.cpp
+++ b/cockatrice/src/cardinfotext.cpp
@@ -31,6 +31,8 @@ CardInfoText::CardInfoText(QWidget *parent)
     powtoughLabel2->setTextInteractionFlags(Qt::TextBrowserInteraction);
     loyaltyLabel1 = new QLabel;
     loyaltyLabel2 = new QLabel;
+    setLabel1 = new QLabel;
+    setLabel2 = new QLabel;
     loyaltyLabel1->setTextInteractionFlags(Qt::TextBrowserInteraction);
 
     textLabel = new QTextEdit();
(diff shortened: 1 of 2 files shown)


Psithief:
What problem is this solving? Is it format-related?

Can it be disabled?

tooomm:
I don't like that either. Way to much interface consumption for a very narrow use case.

Is it really necessary to show more than the most recent set(s) at all?
The idea in 2210 was to quickly validate if a card is legal in e.g. standard or modern. For that purpose basically the most recent ~2 sets are already sufficient?!

Daenyth:
I don't see that this is a reasonable use of space. We already have a subset of users with smaller screens who have UI issues, this only makes it worse, and I literally can't recall in the past 5 years ever wanting this feature. 

I'd rather make the card search in the deck editor panel easier to use, which would solve this case pretty easily.

At most, a right click action in the card info section that would open up a more detailed panel with all card attributes.

Issue #2203, opened 2016-10-04
Missing window decorations


skwerlman:
<b>OS:</b>
Sabayon 16.11

<b>Cockatrice version:</b>
Version a7a0ea2 (2016-09-25)

---

<br>

(Might be related to #2174?)

When I open user information or the debug log, the window that pops up has no decorations.
In the case of the debug log, I cannot interact with any other cockatrice windows _(not sure if intentional? seems counterintuitive to me, at least)_, but I can close the window by hitting ESC.
The user information window can only be closed by right clicking on the title bar and selecting 'Close'.

ZeldaZach:
Same problem on mac that you cannot interact w/ the client when debug is open. I really don't agree with it currently and may consider a reversion. cc @ctrlaltca @Daenyth

tooomm:
@ZeldaZach It has always been like that, on windows as well. Same for many other windows like about or settings.

For debug though it would be very helpful!

ctrlaltca:
> (Might be related to #2174?)
Window decorations are drawn by the theme Qt is currently using.
Maybe the theme you are using doesn't support window decorations on those special kind of dialogs.

In #2221 i moved the debug log from a dialog to a tab, this should fix part of the problem.

ctrlaltca:
Can you check if the bug is caused by https://github.com/adapta-project/adapta-gtk-theme/issues/236 ?

skwerlman:
It's not quite the same, since their issue allowed them to click in the correct spot to close the window. I just tried that with no luck.
Trying again with a different theme, the 'X' _does_ appear for the user details window and log. I suppose its a bug in Adapta, then.

ctrlaltca:
Probably we are creating the dialog in a way that Adapta is not able to handle correctly.
Do you see the same problem in other dialogs, too? This could help us undestand what causes the problem and try to avoit it:
 * Debug log => YES
 * User information => YES
 * Connect dialog => ?
 * Register dialog => ?
 * Create token => ?
 * Create game => ?
 * Edit avatar => ?
 * Edit Password => ?
 * Edit token => ?
 * Edit user => ?
 * Game filters => ?
 * Load deck from clipboard => ?

skwerlman:
All dialogs listed are missing the close button.

ctrlaltca:
Ouch, i suppose we can't really workaround this if no dialog works..

skwerlman:
[...]

PR #2221, opened 2016-10-17
Move the debug log from a modal dialog to a tab

ctrlaltca:
Pro:
 * one less window open
 * the tab is not modal, so it doesn't interrupt the workflow
 * better usability: open the debug log, switch to the deck editor tab and try to load a picture, switch back tab and see why it failed

Related: #2203


Daenyth:
Would it be too hard to make it a fully separate window? I'd like to pull it off the app entirely.

ctrlaltca:
> Would it be too hard to make it a fully separate window?

I don't think it would be hard. I don't like that idea too much, i hate separate windows being lost behind other windows etc..

tooomm:
Is it an option to make it a widget?

ctrlaltca:
> My only concern is I don't want the inputs all stolen by the debug log. I want ./cockatrice > output.txt to work in order to debug launch crashes.

I don't think this is related to this specific change, anyway:
Cockatrice used to spam the terminal it was run from. With the introduction of the debug log i've redirected the output there, but you can still get it from outside Cockatrice running `./cockatrice --debug-log`.

> a fully separate window
> Is it an option to make it a widget?

Polling time: do we want a tab or a window?

Daenyth:
Ctrl is right, that change is unrelated to this one. There's no need to block updates to this. If there's a `--debug-log` command then all we need to do is update the wiki page on debugging

cockatrice/src/tab_debuglog.cpp
```diff
@@ -1,35 +1,48 @@
-#include "dlg_viewlog.h"
+#include "tab_debuglog.h"
 #include "logger.h"
 
 #include <QVBoxLayout>
 #include <QPlainTextEdit>
 
-DlgViewLog::DlgViewLog(QWidget *parent)
-: QDialog(parent)
+TabDebugLog::TabDebugLog(TabSupervisor *tabSupervisor, QWidget *parent)
```
Daenyth: Is there a smaller subset of TabSupervisor you could accept as an interface instead of the whole thing? I'd like to start chipping away at the tight coupling TabSupervisor has to everything.

Or is this because `Tab` requires it? We should think about how we can break that link

cockatrice/src/tab_debuglog.cpp
```diff
@@ -1,35 +1,48 @@
...
+{
+    emit debugLogClosing();
+}
+
+void TabDebugLog::closeRequest()
+{
+    deleteLater();
+}
+
+void TabDebugLog::retranslateUi()
```
Daenyth: Update the tab title for translations maybe?

ctrlaltca: It works the other way around: when the language gets changed TabSupervisor cycle all tabs, request their title and update the caption on the tab header.
The `getTabText()` method returning the tab title is already using `tr()` to return the translated title.

cockatrice/src/tab_debuglog.cpp
ctrlaltca: I agree they are too tightly coupled and it would be better to break this link.
Currently `Tab` requires a `TabSupervisor` in order to use it to get an handle to informations related to the server connection, eg:
 * `tabSupervisor->getIsLocalGame()`
 * `tabSupervisor->getUserListsTab()->getIgnoreList()->getUsers().contains(senderName)`
 * `TabRoom *r = tabSupervisor->getRoomTabs().value(game.room_id());`

Imho `TabSupervisor` should only handle the collection of tabs and, and not be the central dispatcher informations about the current server connection.
Reworking this in a saner way could also lead into removing some limitations, eg. not being able to open a local game when connected to a server.

ctrlaltca:
> Ctrl is right, that change is unrelated to this one. There's no need to block updates to this. If there's a --debug-log command then all we need to do is update the wiki page on debugging

Potentially good for another ticket:
 * add a filename parameter to `--debug-log`
 * add a new option `--debug` that prints logs to strout

ZeldaZach:
>Ctrl is right, that change is unrelated to this one. There's no need to block updates to this. If there's a --debug-log command then all we need to do is update the wiki page on debugging

I attempted to run `./cockatrice --debug-log` and `./cockatrice --debug` with neither printing out anything to the terminal or any file when redirected to do so. Can you explain?

ctrlaltca:
`./cockatrice --debug-log` will save a copy of the logs to a file named `qdebug.txt` in the current working directory where cockatrice is started from.

ZeldaZach:
@ctrlaltca That doesn't seem to happen. Maybe you can address that in another ticket?

Anyway, this seems good to me and I like the change.

tooomm:
>Polling time: do we want a tab or a window?

I highly dislike a tab.
The window was ok, not perfect though.

But what does the tab accomplish or helps you with that the windows doesn't?
You still need to switch focus to do/test something and switch back to check the log. You could just reopen the debug log window to end up with the same result.
This is not a developer version where every click matters and which is focuses around debugging. :)
Moreover I think it's way more confusing than having a normal window. The same as I think the `Settings` or `Edit sets` windows are way better off as windows instead tabs...

A reasonable change and real improvement to everybody would be if you could "use" the log in parallel or next to the app. So you see errors immediately while they happen. That's why I threw the widget idea in... alternatively it would perfectly work as well if the app could be used while the debug log window is open. (@Daenyth seems to be in that boat as well if you read the comments)
I know that some people had complains about that because it could lose focus. But I think the value that it adds is worth the ~2 people which we need to explain to look for a window in the background.
*Proposal for solution: We could bring the debug log window back to the front (+reset position) if it gets opened again while it's already open for example.*

woogerboy21:
> Polling time: do we want a tab or a window?

Makes no real difference to me personally.  I can work with either.  Can't really speak to how much it helps one way or another since I tend to not do much with the client side.

skwerlman:
> Polling time: do we want a tab or a window?

I like the idea of fully separate window for the log. I can't see the log window getting lost being a very large issue, since most users are familiar with the mechanics of windows anyway.

woogerboy21:
Wait... is this a dockable tab?

ctrlaltca:
Sounds like ax external window is the preferred solution, thus closing.

Daenyth:
I like that proposal

PR #2227, opened 2016-10-19
Update log path example when running under windows

woogerboy21:
Added example of log path syntax when running servatrice under windows.


tooomm:
I don't just want to delete branches on the repo that I'm not involved in...
If this is done and not needed anymore please do so @woogerboy21. Thanks.

PR #2228, opened 2016-10-20
Added user privilege level

woogerboy21:
Added a enum column in the users table named "privilevel" with the current values of "none", "vip", and "donator".  Also allowed anyone with a higher privilege level than "none" to log in even if the server
is set to limit the user total and the user limit is reached.  This change add's the new user information into the users container that gets populated and passed between client and server.

### DIFF
--- a/servatrice/migrations/servatrice_0017_to_0018.sql
+++ b/servatrice/migrations/servatrice_0017_to_0018.sql
@@ -1,0 +1,5 @@
+-- Servatrice db migration from version 16 to version 18
+
+alter table cockatrice_users add privlevel enum("NONE","VIP","DONATOR") NOT NULL;
+
+UPDATE cockatrice_schema_version SET version=18 WHERE version=17;
--- a/servatrice/migrations/servatrice_0017_to_0018.sql
+++ b/servatrice/migrations/servatrice_0017_to_0018.sql
@@ -1,4 +1,4 @@
--- Servatrice db migration from version 16 to version 18
+-- Servatrice db migration from version 17 to version 18
 
 alter table cockatrice_users add privlevel enum("NONE","VIP","DONATOR") NOT NULL;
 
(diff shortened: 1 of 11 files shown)


ZeldaZach:
Would you like to explain what VIP and DONOR mean exactly? Just making sure we have the right number and types of enums

woogerboy21:
The idea of the enum was that there could be additional levels added later as some have expressed interest in having them.  But the defaults that I added was VIP (for admin/moderators) and then Donators would be anyone that has provided time to either the project or has given a donation of some type to the servers owner / operator.

woogerboy21:
Should this be seen as a "feature"?  It is something both the client and server recognize, though the server is the only thing so far that does anything with the data really.

woogerboy21:
Waiting for @Daenyth to take a look (since he expressed interest in doing so).  Though the change is pretty basic.

woogerboy21:
FYI, I still hate git .....

servatrice/migrations/servatrice_0017_to_0018.sql
```diff
@@ -0,0 +1,5 @@
+-- Servatrice db migration from version 17 to version 18
+
+alter table cockatrice_users add privlevel enum("NONE","VIP","DONATOR") NOT NULL;
```
Daenyth: Shouldn't this be `ADD COLUMN`, not `ADD`?

servatrice/servatrice.ini.example
```diff
@@ -39,6 +39,8 @@ writelog=1
 
 ; Choose a name for the log file, if enabled; you can specify an absolute path or a path relative to
 ; the servatrice executable; the default file name is server.log (in the same path as servatrice)
+; Note: When running servatrice under windows you will need to use double backslashes between folder locations
```
Daenyth: 👍

servatrice/migrations/servatrice_0017_to_0018.sql
woogerboy21: This actually works both ways (with or without the word column).  At least in the version of mariadb that I am running.  Would you like me to add the word column to the script?

PR #2236, opened 2016-10-24
Fix dynamic user limit settings

woogerboy21:
PR #2220 removed the ability to be able to change the max user limit count while the server is running requiring a restart to make the settings change.  This PR reverts the behavior back to how it operated
prior to the PR.  It also corrects a minor camel hump naming of a function name involved with the setting.

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -234,11 +234,11 @@ bool Servatrice::initServer()
 
     qDebug() << "Store Replays: " << settingsCache->value("game/store_replays", true).toBool();
     qDebug() << "Client ID Required: " << clientIdRequired;
-    maxUserLimitEnabled = settingsCache->value("security/enable_max_user_limit", false).toBool();
+    bool maxUserLimitEnabled = settingsCache->value("security/enable_max_user_limit", false).toBool();
     qDebug() << "Maximum user limit enabled: " << maxUserLimitEnabled;
 
     if (maxUserLimitEnabled){
-        maxUserLimit = settingsCache->value("security/max_users_total", 500).toInt();
+        int maxUserLimit = settingsCache->value("security/max_users_total", 500).toInt();
         qDebug() << "Maximum total user limit: " << maxUserLimit;
         int maxTcpUserLimit = settingsCache->value("security/max_users_tcp", 500).toInt();
         qDebug() << "Maximum tcp user limit: " << maxTcpUserLimit;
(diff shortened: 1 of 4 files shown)


servatrice/src/servatrice.cpp
```diff
@@ -234,11 +234,11 @@ bool Servatrice::initServer()
 
     qDebug() << "Store Replays: " << settingsCache->value("game/store_replays", true).toBool();
     qDebug() << "Client ID Required: " << clientIdRequired;
-    maxUserLimitEnabled = settingsCache->value("security/enable_max_user_limit", false).toBool();
+    bool maxUserLimitEnabled = settingsCache->value("security/enable_max_user_limit", false).toBool();
```
ctrlaltca: We could just call `getMaxUserLimitEnabled()` here

servatrice/src/servatrice.cpp
```diff
@@ -234,11 +234,11 @@ bool Servatrice::initServer()
...
     qDebug() << "Maximum user limit enabled: " << maxUserLimitEnabled;
 
     if (maxUserLimitEnabled){
-        maxUserLimit = settingsCache->value("security/max_users_total", 500).toInt();
+        int maxUserLimit = settingsCache->value("security/max_users_total", 500).toInt();
```
ctrlaltca: We could just call `getMaxUserLimit()`here

servatrice/src/servatrice.cpp
```diff
@@ -725,3 +725,11 @@ void Servatrice::doSendIslMessage(const IslMessage &msg, int serverId)
...
     }
 }
+
+int Servatrice::getMaxUserLimit() const {
+    return settingsCache->value("security/max_users_total", 500).toInt();
+}
+
+bool Servatrice::getMaxUserLimitEnabled() const {
+    return settingsCache->value("security/enable_max_user_limit", false).toBool();
+}
```
ctrlaltca: These helper functions would be better suited as members of the `SettingsCache` class

woogerboy21: Or are you saying I could do something like this inside the header of servatrice?
```
int getMaxUserLimit() { settingsCache->getMaxUserLimit; }
```

woogerboy21:
His request will need a bit more work (and possibly his help) as I am not 100% certain I understand what he is suggesting.  Im pretty sure its the same concept as how the server is presented virtual functions.

PR #2242, opened 2016-10-28
change alert msg for missing features

ZeldaZach:


cockatrice/src/window_main.cpp
```diff
@@ -966,7 +967,7 @@ void MainWindow::refreshShortcuts()
 
 void MainWindow::notifyUserAboutUpdate()
 {
-    QMessageBox::information(this, tr("Information"), tr("Your client appears to be missing features that the server supports.\nThis usually means that your client version is out of date, please check to see if there is a new client available for download."));
+    QMessageBox::information(this, tr("Information"), tr("This server supports additional features that your client doesn't have.\nThis is most likely not a problem, but this message might mean there is a new version of Cockatrice available.\n\nTo update your client, go to Help &rarr; Update Cockatrice."));
```
ZeldaZach: It didn't encode correctly; I'll manually do it

cockatrice/src/window_main.cpp
```diff
@@ -340,11 +340,12 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
             formatedMissingFeatures = "Missing Features: ";
             for (int i = 0; i < missingFeatures.size(); ++i)
                 formatedMissingFeatures.append(QString("\n     %1").arg(QChar(0x2022)) + " " + missingFeatures.value(i)   );
-
+            formatedMissingFeatures.append("\nTo update your client, go to Help -> Update Cockatrice.");
```
Daenyth: typo: `formated` should be `formatted`

cockatrice/src/window_main.cpp
```diff
@@ -966,7 +967,7 @@ void MainWindow::refreshShortcuts()
 
 void MainWindow::notifyUserAboutUpdate()
 {
-    QMessageBox::information(this, tr("Information"), tr("Your client appears to be missing features that the server supports.\nThis usually means that your client version is out of date, please check to see if there is a new client available for download."));
+    QMessageBox::information(this, tr("Information"), tr("This server supports additional features that your client doesn't have.\nThis is most likely not a problem, but this message might mean there is a new version of Cockatrice available.\n\nTo update your client, go to Help -> Update Cockatrice."));
```
Daenyth: Add in `This can also happen if the server is running a custom or pre-release version`

Daenyth:
Can you squash commits also when you're done?

ZeldaZach:
Done @Daenyth

tooomm:
On windows 10 it looks like this:
optional-new-features

I don't like that it shows up for **every single** connection attempt though.
There is no way to say "Hey, good to know... but I'm ok. Please don't bother me with that (until there is another new feature or a strict requirement.)" or at least postpone the friendly reminder for some period of time (e.g. 1 month).
Also the user doesn't get any information (details) what the feature(s) is/are. I like how it's displayed for the required features ([see the pr for a picture](https://github.com/Cockatrice/Cockatrice/pull/1439#issuecomment-136877493)).
<br>
I'll bring up my wording proposal again:
>This server supports additional features that your client doesn't have.
This is most likely not a problem, **but there might be a new version of Cockatrice available or the server is running a custom or pre-release version.**
>
>**To check for client updates**, go to **"**Help -> Update Cockatrice**"**.

It's simpler and less forcing. It suggests to check for updates instead indicating *there is* an update.
I'm also not a fan of the unclear "...but this message might mean..." wording.

woogerboy21:
Do you want a client side setting to disable this popup?

tooomm:
Oh no, not a setting.

I was referring to one of this "don't remind me again" tick boxes in the message, or a "remind me in 1 month again" button next to `ok`. The tick box could also just be a postpone instead a infinite hiding.
But to have to deal with it every single login is a pain. I remember we removed some dialog boxes because they forced attention, disrupted workflow and such. Well this isn't any better... it's just a hint, nothing very important you *have to* deal with immediately.

woogerboy21:
Yeah..... i could add a checkbox in the settings at this point LoL.
More complicated interaction would take me a bit of time to think through and walk the spaghetti code 🤕

tooomm:
We can also just make it a new ticket, a new setting option is not a good way in this case I think.
Any more input? @Daenyth ?

Can we recycle the details like we have them in the required features dialog somehow?
Or would that involve more protocol stuff?

Besides that, what about the wording?

woogerboy21:
@tooomm this doesnt actually change the protocol.  its a change in a dialogue that already exists.  As for the option in the settings, all those pages do is flip Boolean values that are held in memory by the client so regardless of were the checkbox exists on what screen, the behavior under the hood would be the same.  If you were to introduce the UI setting on the dialogue that pops up that introduce alot of code changing for the dialogue since the current dialogue is just a default window with no custom defined layout components.

My recommendation (just my $0.02 though).  Let this PR commit to update the dialogue wording for clarity to end users, then add the options desired were in a later PR.  Nothing says you couldn't have both a box in the settings tab of your choice along with a checkbox on the popup dialogue that can turn the notice on and off.  There would need to be some type of thought process around working some time frame of disabling the notification but that can be worked through as well.

I believe a couple of the reasons that this popup was created was to 1) hint that there could be a client update available and 2) notify the user that the server they are connecting to is operating with additional features so if they pulled the client from say an "official" client source and they connect to a server with features not in that client they should understand the servers owner has modified the code in some way.

ZeldaZach:
This keeps bugging out, made a new PR.

tooomm:
@woogerboy21 I was unsure about the protocol, because I don't know how the required feature thing works exactly and if it requires an additional extension of the protocol to do the same for optional features. Obviously not. :)

The setting is just a flipping setting - sure. But it makes a huge difference where it's placed.
If we have it in normal settings dialog it'll lead to some people not seeing the information at all because it might just be still disabled. Plus it'll clutter the settings even more.
If it's in the dialog box it's directly connected to the context and will make people see it at least once!
But yeah, that's up for discussion [here](https://github.com/Cockatrice/Cockatrice/issues/2249). Thanks for your statements!

PR #2243, opened 2016-10-28
Remove client idle timer flag from feature set

Daenyth:
There's no protocol item at hand here so this isn't a protocol 'feature'


Daenyth:
Compiles locally and I don't want to wait for the CI build

PR #2244, opened 2016-10-28
Add client UI setting for idleclientdisconnect

woogerboy21:
Added disabled option in client UI for client disconnect.  With the way the code is configured the user can enable/disable the idle client timeout setting even while logged in on the fly allowing for users to decide if they want to set there client to timeout without requiring a restart of the client.

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -398,12 +398,16 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     annotateTokensCheckBox.setChecked(settingsCache->getAnnotateTokens());
     connect(&annotateTokensCheckBox, SIGNAL(stateChanged(int)), settingsCache, SLOT(setAnnotateTokens(int)));
 
+    idleClientTimeOutCheckBox.setChecked(settingsCache->getIdleClientTimeOutEnabled());
+    connect(&idleClientTimeOutCheckBox, SIGNAL(stateChanged(int)), settingsCache, SLOT(setIdleClientTimeOutEnabled(int)));
+
     QGridLayout *generalGrid = new QGridLayout;
     generalGrid->addWidget(&notificationsEnabledCheckBox, 0, 0);
     generalGrid->addWidget(&specNotificationsEnabledCheckBox, 1, 0);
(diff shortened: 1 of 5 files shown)


woogerboy21:
image

cockatrice/src/dlg_settings.cpp
```diff
@@ -438,6 +442,7 @@ void UserInterfaceSettingsPage::retranslateUi()
     annotateTokensCheckBox.setText(tr("Annotate card text on tokens"));
     animationGroupBox->setTitle(tr("Animation settings"));
     tapAnimationCheckBox.setText(tr("&Tap/untap animation"));
+    idleClientTimeOutCheckBox.setText(tr("Disconnect from server if sitting idle for an extended periods of time"));
```
Daenyth: `an extended periods` is not gramatically correct.

Let's go with `Disconnect from server if idle for (the exact time limit)`

cockatrice/src/dlg_settings.cpp
```diff
@@ -438,6 +442,7 @@ void UserInterfaceSettingsPage::retranslateUi()
     annotateTokensCheckBox.setText(tr("Annotate card text on tokens"));
     animationGroupBox->setTitle(tr("Animation settings"));
     tapAnimationCheckBox.setText(tr("&Tap/untap animation"));
+    idleClientTimeOutCheckBox.setText(tr("Disconnect from server if idle for 1 hour"));
```
tooomm: Hard-coded numbers should not be part of the translation, especially if they *could* change.
Is it possible to extract the value from settings as a variable?

(In case this is really just a temporary change and will be addressed/solved very soon - I don't care too much...)

Daenyth:
The time is hard coded too for now. When it's configurable the setting name
can be dynamic

tooomm:
I know, it needs to be updated both at the same time.
But I was concerned about problem with translations and discrepancies, or outdated versions.

Just checked... Transifex even complains and warns you if your translation doesn't match the digits. Cool 👍  Maybe I can even make it an error, because the translation is still saved at the moment.
And if we change it and the translations are not updated in time it will fallback to the english source string anyway. So the problem is not too big.

PR #2251, opened 2016-10-31
Recursively search CUSTOM folder for images (i.e. subdirectories)

### LINKED ISSUE #2250: Allow Subfolders in the CUSTOM image folder
<b>OS:</b>
Win 7 SP1

<b>Cockatrice version:</b>
Cockatrice d55e44e

---

<br>

When attempting to add a folder full of custom art for a custom set playtest I had all the images within a subdirectory (Cockatrice\pics\CUSTOM\Dreamscape), however the images within the Dreamscape folder are not detected, meaning I have to place them directly in CUSTOM. 

This is means if I want to have multiple groups of custom images I cannot easily delete or update a single group as all the images end up in a single directory.

I want to propose that sub-directories be supported by the CUSTOM image directory, allowing for easier management of custom content.

ZeldaZach:
Fix #2250.

This will allow for a recursive search through the CUSTOM folder until every image is checked to see if it matches the name of the card in question. If not found, continue like normal. 

Checks subdirectories alphabetically, so if you have 
```
.
├── Bar
│   └── example.jpg
└── Foo
    └── example.jpg
```

The `Bar/example.jpg` will be used.

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -19,6 +19,7 @@
 #include <QSvgRenderer>
 #include <QThread>
 #include <QUrl>
+#include <QDirIterator>
 
 // never cache more than 300 cards at once for a single deck
 #define CACHED_CARD_PER_DECK_MAX 300
@@ -167,9 +168,18 @@ bool PictureLoaderWorker::cardImageExistsOnDisk(QString & setName, QString & cor
     QImage image;
     QImageReader imgReader;
     imgReader.setDecideFormatFromContent(true);
+    QList<QString> picsPaths = QList<QString>();
+    QDirIterator it(customPicsPath, QDirIterator::Subdirectories);
 
-    //The list of paths to the folders in which to search for images
-    QList<QString> picsPaths = QList<QString>() << customPicsPath + correctedCardname;
+    // Recursively check all subdirectories of the CUSTOM folder
+    while (it.hasNext())
+    {
+        QString thisPath(it.next());
+        QFileInfo thisFileInfo(thisPath);
+
+        if (thisFileInfo.isFile() && thisFileInfo.baseName() == correctedCardname)
+            picsPaths << thisPath; // Card found in the CUSTOM directory, somewhere
+    }
 
     if(!setName.isEmpty())
     {


tooomm:
Should the same apply to `customsets` folder?

Maybe somebody wants to have folders for different games and such.

Daenyth:
We should use this for custom sets too. 

What happens if I put a symlink to the root directory inside custom? Will this then try to scan the whole file system?

ZeldaZach:
I'll update it to include the custom sets dir, and true about symlinks. I can have it ignore them

tritoch:
Does root take priority over subfolders?

Nightfirecat:
@Daenyth @ZeldaZach I'd rather Cockatrice not ignore symlinks; I can see a use case for storing images on a separate drive/mount and symlinking the directory. I think that most anyone who's savvy enough to create symlinks is also smart enough not to symlink `/` (or otherwise recursive directories like `../`) within the custom sets dir.

Nightfirecat:
That said, that could also be addressed by adding some kind of directory traversal limit&mdash;just stop searching deeper after some number of directories, say 10.

ZeldaZach:
I can try implementing a limit on how far down you can go @Nightfirecat, but do you think that's an optimal idea?

ZeldaZach:
I took a second look at what I've done and what I feel should be done.
I see no reason to have sub-folders for the `customsets` folder. They're all XML databases, why do they need to be sub'd? It doesn't really make sense, nor does it need to be done...

Second, for symlinks I feel that we should allow them, but limit the depth to 4 (`CUSTOM/symlink/dir1/dir2/dir3/image.jpg`) so symlinks that are linked incorrectly don't jump everywhere...

Finally, I feel the current selection algorithm (alphabetically by directory/filename) is the best choice, but I'm open to hearing other ideas on how we should sort the CUSTOM folder.

Daenyth:
I haven't had a chance to review yet, but I'd really like to see logs on load time so that someone can read their debug log and see what cockatrice tried to find, and the results

ZeldaZach:
@Daenyth Can you explain a little bit more on what you mean for the logs?

Daenyth:
As you iterate through directories, log them

ZeldaZach:
I can put a qDebug if you'd like, but it's a bit wasteful imo.

tooomm:
>I see no reason to have sub-folders for the customsets folder. They're all XML databases, why do they need to be sub'd? It doesn't really make sense, nor does it need to be done...

What about that players who do other game as well, or several ones? Pokémon, Yu-Gi-Oh!, Hearthstone...
They might want a folder in customsets for each game to not mess with single pre-release xls's for certain sets or other custom xml's etc. which show up for each game individually. They might even favor that, and don't use a cards.xml at all?
You could also simply move the folder of the game you're about to play to customsets, 

It doesn't hurt to have it anyway?

tooomm:
What happens to "qDebug" and "doing the same in customsets folder"?!

ctrlaltca:
If it's still relevant, open a new ticket describing the needed improvements; "doing the same in customsets folder" doesn't really belong to this PR anyway.
The PR was approved by @Daenyth and stuck for 38 days, i think that the user originally reporting the bug at #2250 will approve being finally able to use this feature.

Daenyth:
Agreed on all counts

PR #2255, opened 2016-10-31
Adds !corrected_name! and !corrected_name_lower! to customURL

Fizztastic:
Fixes a bug introduced in https://github.com/Cockatrice/Cockatrice/pull/2164 where split cards can no longer be downloaded using !name! in a customURL.


tooomm:
Can you document it please?
https://github.com/Cockatrice/Cockatrice/wiki/Custom-Download-URLs

Why is the wiki page named "custom download url" btw?
No hint in the name what its about... should contain pic(ture) or image?!

Fizztastic:
@tooomm I've made the changes to the wiki.

tooomm:
So the new "corrected name" specifically helps with split cards?
Just that? Or are there other effects/use cases as well?

tritoch:
It performs a few strips that can be a hassle for directory-based web servers.

`return result.remove(" // ").remove(':').remove('"').remove('?').replace('/', ' ');`

PR #2259, opened 2016-11-02
Rework idle timeout, now server side

### LINKED ISSUE #2208: Server side Inactivity logout

woogerboy21:
Fix #2208 

Initial commit for server side idle timeout counter.  This adds a new int value that is updated when room/game/mod/admin commands occur and is checked during the regular ping timout function that if the users new "idle" value exceeds the idleclienttimeout value defined in the servers configuration file the user is logged out.  The user will receive a warning at the 90% time frame mark about being idle.

This warning / disconnect only affects un-privileges users. Also settings the servers configuration file setting to 0 will disable the functionality.

image

### DIFF
--- a/cockatrice/src/tab_supervisor.cpp
+++ b/cockatrice/src/tab_supervisor.cpp
@@ -583,6 +583,7 @@ void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
 {
 
     switch ((Event_NotifyUser::NotificationType) event.type()) {
+        case Event_NotifyUser::UNKNOWN: QMessageBox::information(this, tr("Unknown Event"), tr("The server has sent you a message that your client does not understand.\nThis message might mean there is a new version of Cockatrice available or this server is running a custom or pre-release version.\n\nTo update your client, go to Help -> Update Cockatrice.")); break;
         case Event_NotifyUser::IDLEWARNING: QMessageBox::information(this, tr("Idle Timeout"), tr("You are about to be logged out due to inactivity.")); break;
         case Event_NotifyUser::PROMOTED: QMessageBox::information(this, tr("Promotion"), tr("You have been promoted to moderator. Please log out and back in for changes to take effect.")); break;
         case Event_NotifyUser::WARNING: {
--- a/cockatrice/src/tab_supervisor.cpp
+++ b/cockatrice/src/tab_supervisor.cpp
@@ -583,6 +583,7 @@ void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
 {
 
     switch ((Event_NotifyUser::NotificationType) event.type()) {
+        case Event_NotifyUser::IDLEWARNING: QMessageBox::information(this, tr("Idle Timeout"), tr("You are about to be logged out due to inactivity.")); break;
         case Event_NotifyUser::PROMOTED: QMessageBox::information(this, tr("Promotion"), tr("You have been promoted to moderator. Please log out and back in for changes to take effect.")); break;
         case Event_NotifyUser::WARNING: {
             if (!QString::fromStdString(event.warning_reason()).simplified().isEmpty())
--- a/cockatrice/src/tab_supervisor.cpp
+++ b/cockatrice/src/tab_supervisor.cpp
@@ -342,12 +342,10 @@ void TabSupervisor::gameJoined(const Event_GameJoined &event)
     connect(tab, SIGNAL(gameClosing(TabGame *)), this, SLOT(gameLeft(TabGame *)));
     connect(tab, SIGNAL(openMessageDialog(const QString &, bool)), this, SLOT(addMessageTab(const QString &, bool)));
     connect(tab, SIGNAL(openDeckEditor(const DeckLoader *)), this, SLOT(addDeckEditorTab(const DeckLoader *)));
-    connect(tab, SIGNAL(notIdle()), this, SLOT(resetIdleTimer()));
     int tabIndex = myAddTab(tab);
     addCloseButtonToTab(tab, tabIndex);
     gameTabs.insert(event.game_info().game_id(), tab);
     setCurrentWidget(tab);
-    emit idleTimerReset();
 }
 
 void TabSupervisor::localGameJoined(const Event_GameJoined &event)
@@ -391,7 +389,6 @@ void TabSupervisor::addRoomTab(const ServerInfo_Room &info, bool setCurrent)
     roomTabs.insert(info.room_id(), tab);
     if (setCurrent)
         setCurrentWidget(tab);
-    emit idleTimerReset();
 }
 
 void TabSupervisor::roomLeft(TabRoom *tab)
@@ -444,7 +441,6 @@ TabMessage *TabSupervisor::addMessageTab(const QString &receiverName, bool focus
     tab = new TabMessage(this, client, *userInfo, otherUser);
     connect(tab, SIGNAL(talkClosing(TabMessage *)), this, SLOT(talkLeft(TabMessage *)));
     connect(tab, SIGNAL(maximizeClient()), this, SLOT(maximizeMainWindow()));
-    connect(tab, SIGNAL(notIdle()), this, SLOT(resetIdleTimer()));
     int tabIndex = myAddTab(tab);
     addCloseButtonToTab(tab, tabIndex);
     messageTabs.insert(receiverName, tab);
@@ -594,9 +590,4 @@ void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
         default: ;
     }
 
-}
-
-void TabSupervisor::resetIdleTimer()
-{
-    emit idleTimerReset();
 }
--- a/common/pb/event_notify_user.proto
+++ b/common/pb/event_notify_user.proto
@@ -3,6 +3,7 @@ import "session_event.proto";
 message Event_NotifyUser {
 
     enum NotificationType {
+        UNKNOWN = 0;
         PROMOTED = 1;
         WARNING = 2;
         IDLEWARNING = 3;
--- a/common/pb/event_notify_user.proto
+++ b/common/pb/event_notify_user.proto
@@ -5,6 +5,7 @@ message Event_NotifyUser {
     enum NotificationType {
         PROMOTED = 1;
         WARNING = 2;
+        IDLEWARNING = 3;
     }
 
     extend SessionEvent {
--- a/common/pb/event_notify_user.proto
+++ b/common/pb/event_notify_user.proto
@@ -3,7 +3,7 @@ import "session_event.proto";
 message Event_NotifyUser {
 
     enum NotificationType {
-        UNKNOWN = 0;
+        UNKNOWN = 0;  // Default enum value if no "type" is defined when used
         PROMOTED = 1;
         WARNING = 2;
         IDLEWARNING = 3;
--- a/common/server.h
+++ b/common/server.h
@@ -64,7 +64,7 @@ public:
     virtual bool getClientIDRequiredEnabled() const { return false; }
     virtual bool getRegOnlyServerEnabled() const { return false; }
     virtual bool getMaxUserLimitEnabled() const { return false; }
-    virtual int getIdleClientTimout() const { return 0; }
+    virtual int getIdleClientTimeout() const { return 0; }
     virtual int getClientKeepAlive() const { return 0; }
     virtual int getMaxGameInactivityTime() const { return 9999999; }
     virtual int getMaxPlayerInactivityTime() const { return 9999999; }
--- a/common/server.h
+++ b/common/server.h
@@ -64,6 +64,7 @@ public:
     virtual bool getClientIDRequiredEnabled() const { return false; }
     virtual bool getRegOnlyServerEnabled() const { return false; }
     virtual bool getMaxUserLimitEnabled() const { return false; }
+    virtual int getIdleClientTimout() const { return 0; }
     virtual int getClientKeepAlive() const { return 0; }
     virtual int getMaxGameInactivityTime() const { return 9999999; }
     virtual int getMaxPlayerInactivityTime() const { return 9999999; }
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -19,6 +19,7 @@
 #include "pb/event_game_joined.pb.h"
 #include "pb/event_room_say.pb.h"
 #include "pb/serverinfo_user.pb.h"
+#include "pb/event_notify_user.pb.h"
 #include <google/protobuf/descriptor.h>
 #include "featureset.h"
 
@@ -32,7 +33,8 @@ Server_ProtocolHandler::Server_ProtocolHandler(Server *_server, Server_DatabaseI
       acceptsUserListChanges(false),
       acceptsRoomListChanges(false),
       timeRunning(0),
-      lastDataReceived(0)
+      lastDataReceived(0),
+      lastActionReceived(0)
 {
     connect(server, SIGNAL(pingClockTimeout()), this, SLOT(pingClockTimeout()));
 }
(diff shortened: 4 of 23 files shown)


woogerboy21:
Now for the bad news.  The current protobuf definition for "Notify_User" seems to be defaulting to the "PROMOTED" type definition on older clients.  So when users with older clients are sent the "IDLEWARN" event they see this:

image

woogerboy21:
Wait, why does the old XCode travis tests not like ceil?  Should I write standard integer logic to come up with a solution such as :

```
int roundUp(int numToRound, int multiple)
{
    if (multiple == 0)
        return numToRound;

    int remainder = abs(numToRound) % multiple;
    if (remainder == 0)
        return numToRound;

    if (numToRound < 0)
        return -(abs(numToRound) - remainder);
    else
        return numToRound + multiple - remainder;
}
```

Nightfirecat:
I'll be the first to admit that I'm totally C++ illiterate (closest experience I've got is multiple-years-old Java), but it looks like there's no `#include <math.h>` for `ceil()` to work?

Again, my apologies if I'm fully blind and haven't seen this elsewhere.

common/server.h
```diff
@@ -64,6 +64,7 @@ private slots:
     virtual bool getClientIDRequiredEnabled() const { return false; }
     virtual bool getRegOnlyServerEnabled() const { return false; }
     virtual bool getMaxUserLimitEnabled() const { return false; }
+    virtual int getIdleClientTimout() const { return 0; }
```
ctrlaltca: Typo: missing "e" in Timeout

common/server_protocolhandler.cpp
```diff
@@ -373,6 +383,20 @@ void Server_ProtocolHandler::pingClockTimeout()
 
     if (timeRunning - lastDataReceived > server->getMaxPlayerInactivityTime())
         prepareDestroy();
+
+    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none")
+        if ((timeRunning - lastActionReceived) == round(server->getIdleClientTimout() *.9)) {
```
ctrlaltca: Nice idea to add a warning, but i see a problem here.
This will only work if the config value of `clientkeepalive` is set to the default of 1, so the `pingClockTimeout()` method is called every second.
If a server admin sets that value to eg. 5 secs, there's only 20% probability that `pingClockTimeout()` gets executed exactly `server->getIdleClientTimout() *.9` seconds after the last action has been received.

A quick workaround could be to introduce the value of `clientkeepalive` in the equation, eg:
```
int pingClockInterval = settingsCache->value("server/clientkeepalive", 1).toInt();
int delta = timeRunning - lastActionReceived - round(server->getIdleClientTimout() *.9);
if(delta > 0 && delta < clientkeepalive) { ... }
```
I'm not exactly a fan of this solution, if you find something better go for it.

common/pb/event_notify_user.proto
```diff
@@ -5,6 +5,7 @@ message Event_NotifyUser {
     enum NotificationType {
         PROMOTED = 1;
         WARNING = 2;
+        IDLEWARNING = 3;
```
ctrlaltca: To avoid the problem of the message defaulting to "PROMOTED", you could set an explicit default value for NotificationType, see https://developers.google.com/protocol-buffers/docs/proto#optional
Right now since this is an `enum`, it defaults to its first value.

woogerboy21:
> I'll be the first to admit that I'm totally C++ illiterate (closest experience I've got is multiple-years-old Java), but it looks like there's no #include <math.h> for ceil() to work?

RawRawRaw... if this is it ... :poop:

common/server_protocolhandler.cpp
woogerboy21: Originally the runction used `ceil` rather than `round` and I prob be switching things back to that.  If we are introducing another counter would it make more since to add a bool into the equation and turn the bool on/off if the warning has been sent and have the bool clear at some point rather than making a best guess in the event the ping clock timer is set to some number that the math doesnt really work for?

ctrlaltca: Sure, that would work; just ensure to reset the bool every time you reset lastActionReceived

common/pb/event_notify_user.proto
woogerboy21: Yeah, Ill add an "UNKNOWN" enum value at the 0 location and add a client side switch statement so future clients will be able to show a message about the client being out of date.  Thanks.

ctrlaltca: Adding `UNKNOWN = 0` is a smart move 👍

Daenyth:
assigning for review purposes, I'll take a closer look when I'm off work

ctrlaltca:
> This means you broke the API in an incompatible way. There's going to be another way to define your api change that doesn't cause this problem, we need to find that instead. This isn't an acceptable api break

@Daenyth: his proposal of adding a "UNKNOWN" enum value with value 0 would fix this problem quite smartly.

woogerboy21:
Yes, regardless of if I make a default value for the type or define an enum 0 value when trying to use this protocol component for anything other than what is currently written will have this affect.  IMO its a perfect candidate for something like the featureset functionality.  The API is broken when using this event and it needs fixed in one way or another with the proposed solution.  Infact we probably should go through and make sure other declarations that have enum's have either a default value or an `unknown` catch all as the first value in the enum list for future break fix reasons.

woogerboy21:
The only other thing I can think of would be to fix the protocol declarations in this change and strip out the notification event and add that  in later after the client updates role out.

Daenyth:
What I mean though is that adding a new event type should not result in the clients showing anything at all - clients are supposed to ignore events they don't understand (events that fail to parse).

If the event this is sending gets parsed as some other kind of unrelated event then it's a problem with the event itself, unless I'm mistaken. I'll take a closer look at the code

Daenyth:
The behavior I'd expect for a current client (say the All Hallows one) connecting to a server with this:

- client is idle
- server sends notification
- client doesn't show anything to the user
- server disconnects client
- client tells user it was disconnected

woogerboy21:
It's an issue with the protobuf definitions  (proto file) .  Not how the event is declared in the code of the server.

woogerboy21:
So right now the reason you would see the wrong dialogue is because the older client has no default value in the proto file for event_notify user.  In that condition when an unknown enum event value is received by that client it defaults to the first enum value in the enum list.  There is no way around that till the event definitions are fixed in the proto file.  

Unless I create or use a different event but that wouldn't make since because the point of the notifyuser event is to do just that.

Daenyth:
Gotcha I understand now. That's unfortunate. I don't think there's a way we can fix this other than releasing a new version and not using the feature until the version is out for a while.

woogerboy21:
Yeah so the options I can see are:

1. Create functionality but drop notification till later (functionality includes proto fix)

2. Create separate PR for proto fix, then roll this in master at some time later.

3. Create separate PR to fix proto.  Remove notification from this PR and and make new PR later for notification. 

Any other ideas?

Daenyth:
How about we fix the protocol now (in this pr or another), and for the idle timeout can go out whenever, but we don't enable it until more people are on the new client

woogerboy21:
Alright, I have made the requested changes, feel free to take a look over the PR again.  

@Daenyth I have the proto file fixed in this PR.  I have no problem either allowing this one to commit and then not turning the functionality on OR if you would like, I can write another PR up to fix the proto file and then let this change sit for a period of time.  I personally dont like seeing PR's sit out in the list for long periods of time but I understand if you want to go that route.  I would hate for people to randomly get messages about being promoted to moderator :)

woogerboy21:
YaY, the build checks pass.  Man that Travis guy is picky.....

woogerboy21:
Just out of curiosity, should I add something to the client side debug log when a user is logged out from receiving an idle timeout disconnect?  That way if troubleshooting in some way you can see it in the log?

cockatrice/src/tab_supervisor.cpp
```diff
@@ -583,6 +583,8 @@ void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
 {
 
     switch ((Event_NotifyUser::NotificationType) event.type()) {
+        case Event_NotifyUser::UNKNOWN: QMessageBox::information(this, tr("Unknown Event"), tr("The server has sent you a message that your client does not understand.\nThis message might mean there is a new version of Cockatrice available or this server is running a custom or pre-release version.\n\nTo update your client, go to Help -> Update Cockatrice.")); break;
```
Daenyth: 👍 

We may want to consider pulling out this string to a reusable constant since we may use it in multiple places.

cockatrice/src/tab_supervisor.cpp
```diff
@@ -583,6 +583,8 @@ void TabSupervisor::processNotifyUserEvent(const Event_NotifyUser &event)
...
+        case Event_NotifyUser::IDLEWARNING: QMessageBox::information(this, tr("Idle Timeout"), tr("You are about to be logged out due to inactivity.")); break;
```
Daenyth: 👍

common/pb/event_notify_user.proto
```diff
@@ -3,8 +3,10 @@ import "session_event.proto";
 message Event_NotifyUser {
 
     enum NotificationType {
+        UNKNOWN = 0;
```
Daenyth: 👍  though a comment referencing that the first enum value is what gets selected for non-optional fields the parser doesn't understand might be useful

servatrice/servatrice.ini.example
```diff
@@ -71,6 +71,10 @@ requiredfeatures=""
 ; to choose from.  Example: "Flaming,Foul Language"
 officialwarnings="Flamming,Spamming,Causing Drama,Abusive Language"
 
+; Maximum time in seconds a player can stay connected but idle.  Default is 3600 (0 = disabled)
+; Clients will be notified at the 90% time period of pending disconnection if they do not take action.
+idleclienttimeout=3600
```
Daenyth: 👍

common/server_protocolhandler.cpp
```diff
@@ -373,6 +390,22 @@ void Server_ProtocolHandler::pingClockTimeout()
 
     if (timeRunning - lastDataReceived > server->getMaxPlayerInactivityTime())
         prepareDestroy();
+ 
+    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none")
```
Daenyth: Hiding this conditional under a `userMayBeTimedOutFromInactivity` would be good - it would give a place for a comment to live explaining that it only applies to non-{mod/admin/vip} get timed out

common/server_protocolhandler.cpp
```diff
@@ -373,6 +390,22 @@ void Server_ProtocolHandler::pingClockTimeout()
...
+        if ((server->getIdleClientTimeout() > 0) && (idleClientWarningSent == true))
+            if (timeRunning - lastActionReceived > server->getIdleClientTimeout())
+                prepareDestroy();
+
+    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none")
```
Daenyth: You're repeating this check twice in a row.

common/server_protocolhandler.cpp
```diff
@@ -373,6 +390,22 @@ void Server_ProtocolHandler::pingClockTimeout()
 
     if (timeRunning - lastDataReceived > server->getMaxPlayerInactivityTime())
         prepareDestroy();
+ 
+    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none")
+        if ((server->getIdleClientTimeout() > 0) && (idleClientWarningSent == true))
+            if (timeRunning - lastActionReceived > server->getIdleClientTimeout())
+                prepareDestroy();
```
Daenyth: I'd either wrap these conditional blocks with `{}` or put then inside one large `if` clause. The depth here with lack of scoping braces looks confusing.

common/server_protocolhandler.cpp
```diff
@@ -373,6 +390,22 @@ void Server_ProtocolHandler::pingClockTimeout()
...
+
+    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none")
+        if (((timeRunning - lastActionReceived) >= ceil(server->getIdleClientTimeout() *.9)) && (idleClientWarningSent == false)) {
```
Daenyth: `&& (!idleClientWarningSent)` would be a bit cleaner; comparing bool values to bool constants to get a resulting bool value is a bit excessive ;)

common/server_protocolhandler.cpp
```diff
@@ -171,6 +176,9 @@ Response::ResponseCode Server_ProtocolHandler::processRoomCommandContainer(const
     if (!room)
         return Response::RespNotInRoom;
 
+    lastActionReceived = timeRunning;
+    idleClientWarningSent = false;
```
Daenyth: These two lines of code are repeated a lot. It would be better to hide them under a `resetIdleTimer()` function

servatrice/src/servatrice.cpp
```diff
@@ -405,6 +405,8 @@ bool Servatrice::initServer()
         }
     }
 #endif
+
+    qDebug() << "Idle client timeout value: " << getIdleClientTimeout();
```
Daenyth: This provides no information about which user this relates to. That will make this log hard to use in any way for the admin

common/server_protocolhandler.cpp
woogerboy21: This condition check is used in a couple places at this point (during login for example) and now here.  I'm not so sure `userMayBeTimedOutFromInactivity` is the best description. how about something like `userIsUnderPrivileged` ?

servatrice/src/servatrice.cpp
woogerboy21: So you want a more descriptive log output basically?

Daenyth: The motivating scenario:

You're wooger. You run a server. You open your logs and see this line in them:
`DEBUG Idle client timeout value: 50`

What do you do?

cockatrice/src/tab_supervisor.cpp
woogerboy21: Are constants translatable?

common/server_protocolhandler.cpp
Daenyth: wnb

The former can call the latter - for now. One function encapsulating the mechanical check and another function named for the semantic intent.

It's only by current coincidence that those two function names would return the same value. One could change but not the other.

servatrice/src/servatrice.cpp
woogerboy21: I understand that my clients idle timeout is set to 50 seconds since i read the description in the servers ini file and realize that the clients will be warned for being idle at 90% of 50 seconds?  The output is there for troubleshooting value(s) on server startup.  Same way the output for the path to the servers ini file is listed on startup, or the value for if your using a db or not.

common/server_protocolhandler.cpp
woogerboy21: Some how I need to get an emo of an old guy swinging a cane....

Daenyth:
I was thinking it was per user, ignore me lol

woogerboy21:
@Daenyth I fixed most things so if you happen to want to take another look go for it.  The only thing I didnt change that you mentioned was making a function that returns if the user is under privileged.  I started to do so and couldn't think of a good way to do so without causing a lock on the user list since the function was written generically to pass a username in and would iterate the user list to make sure the user was actually logged in before trying to query the users info. There might be a better way to do so?

tooomm:
Should this pr take care of reverting the client side logout as well? client --> server

woogerboy21:
> Should this pr take care of reverting the client side logout as well? client --> server
I can or we could make a different PR to pull it

Daenyth:
Go for it

woogerboy21:
@tooomm Anything else?

PR #2260, opened 2016-11-02
Fix one to many zero's

woogerboy21:
Looks like I typo'd the default value for the number of seconds in 1 hour for the default client side idle time out.  This fixes it.


cockatrice/src/settingscache.cpp
```diff
@@ -162,7 +162,7 @@ SettingsCache::SettingsCache()
     notifyAboutUpdates = settings->value("personal/updatenotification", true).toBool();
     lang = settings->value("personal/lang").toString();
     keepalive = settings->value("personal/keepalive", 5).toInt();
-    idlekeepalive = settings->value("personal/idlekeepalive", 36000).toInt();
+    idlekeepalive = settings->value("personal/idlekeepalive", 3600).toInt();
```
Daenyth: For time related stuff like this I prefer using something like `60 * 60` to make it totally clear, but this is also fine

Issue #833, opened 2015-03-15
Move `list of all connected clients to this server` to the server tab


tooomm:
There are two types of user lists.
- **at "user lists" tab ("Benutzerlisten" in the pictures) on the right:** <br>shows all clients connected to the current server
- **at "Main" tab (used to be "Chat Room" in the older pictures) on the right:** <br>shows all clients connected to the current server, which are in this particular room

This is a bit confusing right now and they are misplaced a bit.
This issue is about moving the first one out of the "user lists" tab into the "server" tab.

Regarding the maybe upcoming possibilities of user created rooms it get's more important to have a list of all users at the server, not just the room you are in right now.
And you want to look for that at the server tab for sure...
The space gained at the "user lists" tab can be used for displaying and editing of account/profile details in future versions or additional buddy/ignore lists functionality.

server tab right now:
![server tab](https://cloud.githubusercontent.com/assets/9874850/6657051/adde3d04-cb3f-11e4-9937-b9acad196b07.png)

server tab with list of all clients/users connected to the server ( /= list of users in the room you are in!)
![server tab change](https://cloud.githubusercontent.com/assets/9874850/6657052/b05702a0-cb3f-11e4-974d-d99e080a3ec8.png)
(if you look closely, you see that there are 534 users in the "Main" room (used to be just "Chat Room"), but 543 users are connected to the server in total, some of them just left the "Main" room, but are still connected.)

ZeldaZach:
I was able to get the layout in a similar fashion to what was designed above, but I couldn't get it functional. https://stackoverflow.com/questions/3171965/widget-with-twoor-more-layout

tooomm:
Happy to see somebody working on this!
One step closer to a clearer user interface. :)

woogerboy21:
So what are your proposals for the "buddy lists" on the accounts tab?  The buddy lists are tied directly to the server in which you are connecting to?  Hense why the list of users on the server you are connected to lives in the accounts tab as well.

tooomm:
>The buddy lists are tied directly to the server in which you are connecting to?

Buddy lists are totally fine in the `Account` tab!

>Hense why the list of users on the server you are connected to lives in the accounts tab as well.

But the complete list of connected clients is "tied directly to the server". Exactly for that reason they should be in the `Server` tab, not the `Account` tab. The users on the server are not connected to your account. Your personal buddy and ignore lists are, but not the list of connected clients.

[...]

PR #2267, opened 2016-11-07
[WIP] Close Replay/Deck Tabs

ZeldaZach:
This is a primitive attempt at allowing all tabs to close, but there are some things to still be done/fixed up.

1) When you attempt to reopen tabs, they sometimes don't function as they should (i.e. the Server tab will throw an error)

2) You cannot close the Account tab or the program crashes b/c the user list within it is called upon elsewhere in the code

3) When you try to close the program/disconnect from a server, the `TabSupervisor::stop()` command does not work properly and segfaults if you close the original tabs in the client and/or add new tabs

4) There's no way to currently limit the menu from showing the 2 admin tabs 

If anybody would like to provide some feedback and/or a branch of my work on how to solve these 4 initial errors I've come across, please feel free to!

---

I decided to make this PR work for the tabs that seem to have no regressions on them (Game Replays and Deck Storage). I also allow the admin tabs to be closed (as logs isn't seeing much use until we can rewrite it, and you can always login again for the admin tab). This PR should be ready to go whenever tested.


tooomm:
>You cannot close the Account tab or the program crashes b/c the **user list within it** is called upon elsewhere in the code

See https://github.com/Cockatrice/Cockatrice/issues/833
I think it doesn't belong there anyway and should be moved to the `Server` tab!

ZeldaZach:
@ctrlaltca Wanna look into #833 ?

backendr:
why not just make the replays and deck storage closable?

ZeldaZach:
@tooomm 1) Actions -> Pick the tab (same to how deck editor is done now)

2) is not a problem for this PR as that's the current behavior for _any_ tab

cockatrice/src/tab_supervisor.cpp
```diff
@@ -263,19 +260,6 @@ void TabSupervisor::stop()
         localClients.clear();
         
         emit localGameEnded();
-    } else {
```
ZeldaZach: TODO: Rework this so new opened tabs close

tooomm:
> 2) is not a problem for this PR as that's the current behavior for any tab

No? If I disconnect *all* tabs close. Just `Deck Editor` stays open...

cockatrice/src/tab_deck_storage.cpp
```diff
@@ -112,6 +112,16 @@ TabDeckStorage::TabDeckStorage(TabSupervisor *_tabSupervisor, AbstractClient *_c
     setCentralWidget(mainWidget);
 }
 
+TabDeckStorage::~TabDeckStorage()
+{
+    emit deckStorageClosing(this);
```
Daenyth: 👎 

Let's do this reusably. Let's make a `ClosableTab` mixin class that adds an `emit tabClosing(this)` method. We don't need a differently-named copy of the same method across each tab class.

cockatrice/src/tab_supervisor.cpp
```diff
@@ -190,45 +190,42 @@ void TabSupervisor::start(const ServerInfo_User &_userInfo) {
...
     updatePingTime(0, -1);
 
     if (userInfo->user_level() & ServerInfo_User::IsRegistered) {
         tabDeckStorage = new TabDeckStorage(this, client);
-        connect(tabDeckStorage, SIGNAL(openDeckEditor(
-                                               const DeckLoader *)), this, SLOT(addDeckEditorTab(
-                                                                                        const DeckLoader *)));
-        myAddTab(tabDeckStorage);
+        connect(tabDeckStorage, SIGNAL(openDeckEditor(const DeckLoader *)), this, SLOT(addDeckEditorTab(const DeckLoader *)));
+        int ind = myAddTab(tabDeckStorage);
```
Daenyth: What's this for? I'm having a hard time following how it's used, especially because `ind` as a name doesn't tell me a whole lot. An index of some kind?

cockatrice/src/tab_supervisor.cpp
```diff
@@ -190,45 +190,42 @@ void TabSupervisor::start(const ServerInfo_User &_userInfo) {
...
+        addCloseButtonToTab(tabDeckStorage, ind);
```
Daenyth: We should make a way to have this not be explicit - ClosableTab should just have this by itself

PR #2271, opened 2016-11-09
Fix #2270

### LINKED ISSUE #2270: Server crash on register
<b>OS:</b>
Ubuntu 16.04

<b>Servatrice version:</b>
6962777

___
<br>

https://gist.github.com/skwerlman/eeaf79a9c89b7cc8c887c2aa126a920c

This seems to be caused by #2259

woogerboy21:
Fix #2270
Check userinfo first before trying to use it.

### DIFF
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -387,7 +387,7 @@ void Server_ProtocolHandler::pingClockTimeout()
     if (timeRunning - lastDataReceived > server->getMaxPlayerInactivityTime())
         prepareDestroy();
 
-    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none") {
+    if (!userInfo || QString::fromStdString(userInfo->privlevel()).toLower() == "none") {
         if ((server->getIdleClientTimeout() > 0) && (idleClientWarningSent)) {
             if (timeRunning - lastActionReceived > server->getIdleClientTimeout()) {
                 prepareDestroy();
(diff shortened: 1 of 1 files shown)


Daenyth:
I'd really like to know why this happens

Daenyth:
What's the series of events that triggers this?

ctrlaltca:
You connect to the server and issue no command at all (or register and then just wait).
The server sends a ping, then checks for ping timeout.
The method where this is implemented dereferences the pointer userInfo without checking if it's not null.

ctrlaltca:
> Shouldn't userInfo be not-null once you're connected?

Not the way it's currently implemented: `cmdLogin` is accepted only when userInfo is null and i-s filled using the user-provided data.

> Does the server ping before it checks auth?

The client pings the server every few seconds; the server only checks if the last ping has been received before then `max_player_inactivity_time` seconds ago and disconnects the client.
The server uses a single timer to check this for all connected clients, and the timer signal is connected to the method's slot as soon as the client object is created.

common/server_protocolhandler.cpp
```diff
@@ -387,7 +387,7 @@ void Server_ProtocolHandler::pingClockTimeout()
     if (timeRunning - lastDataReceived > server->getMaxPlayerInactivityTime())
         prepareDestroy();
 
-    if (QString::fromStdString(userInfo->privlevel()).toLower() == "none") {
+    if (!userInfo || QString::fromStdString(userInfo->privlevel()).toLower() == "none") {
```
ZeldaZach: I'm personally a fan of `if (userInfo && ...)` but this is fine.

ctrlaltca: `if (userInfo && ...)` would skip the check for not-logged in users.

PR #2275, opened 2016-11-11
Add missing feature rememberance to client

### LINKED ISSUE #2249: Don't show optional features info msg for EVERY login
<b>OS:</b>
Win 10 64bit

<b>Cockatrice version:</b>
3498b16 (2016-10-30) "All Hallow's Eve"

---

<br>

This is a follow up issue ticket from another PR, check there first for information and history:
https://github.com/Cockatrice/Cockatrice/pull/2242#issuecomment-257087739

We decided to get the wording change in quickly and postpone the proper settings handling to annoy users less. Other, a bit more complex, changes like e.g. a custom message window are kept track of here.

woogerboy21:
Fix #2249

Add the ability for the client to remember the missing features that it received from the server it last connected to and not display the "missing/optional feature" message at every login.  At each startup the client checks the known missing feature list and compares it to its feature set clearing any known features that have been added after a client upgrade.  The client will now only display the missing / optional feature window when a new missing feature is found upon connecting to a server only once.  Afterwords it will not display the message until an even newer feature the client is missing is found.

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -25,6 +25,7 @@ RemoteClient::RemoteClient(QObject *parent)
     : AbstractClient(parent), timeRunning(0), lastDataReceived(0), messageInProgress(false), handshakeStarted(false), messageLength(0)
 {
 
+    clearNewClientFeatures();
     int keepalive = settingsCache->getKeepAlive();
     timer = new QTimer(this);
     timer->setInterval(keepalive * 1000);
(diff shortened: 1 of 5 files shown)


Daenyth:
Awesome! I'll review as soon as I get a chance

Psithief:
So the user isn't allowed to find out what the missing feature is again from some esoteric window deep inside a help menu? :D

woogerboy21:
>So the user isn't allowed to find out what the missing feature is again from some esoteric window deep inside a help menu? :D

Im confused by this.  The user is never presented the missing features that there client has unless its a required feature on the server preventing them from logging in.  Otherwise there is a popup on login that indicates they are missing features that the server they are connecting to understands.  And before this change it pops up every single time the user logs in and there client is missing the feature.  

With this change the only difference in user experience is that the client will only present the popup message about missing features once as the client now remembers what those features are.  The experience for what features the users see's if denied access to the server from a missing feature doesnt change.

woogerboy21:
Was this ever looked at by anyone?  I know personally with work and the holiday stuff pretty much here I havent had a bunch of time myself to do much more project wise.

cockatrice/src/remoteclient.cpp
```diff
@@ -135,6 +136,13 @@ void RemoteClient::processConnectionClosedEvent(const Event_ConnectionClosed & /
 void RemoteClient::loginResponse(const Response &response)
 {
     const Response_Login &resp = response.GetExtension(Response_Login::ext);
+    
+    QString possibleMissingFeatures;
```
Daenyth: Why not a `QSet<QString>`? That would provide much easier checking, I'd think

woogerboy21: I actually started out by using a QSet<QString> but ran into an issue with doing so.  At this point though I don't recall what the issue was and it very well could have been something to do with me taking a different approach and then changing it after the failures occurred.

PR #2283, opened 2016-11-23
Fix Travis CI build

cotsog:
Hey @ZeldaZach,

I think I found the reason why your builds were failing.

The `set -e` statements fundamentally alters the way the script behaves when any command exits with nonzero status.

Hence, I would advise removing these statements or make sure to call `set +e` like  I did here.

Let me know if this helps.

Thanks!

### DIFF
--- a/servatrice/check_schema_version.sh
+++ b/servatrice/check_schema_version.sh
@@ -31,3 +31,5 @@ if ! grep -q "$expected_define" servatrice/src/servatrice_database_interface.h; 
     echo "servatrice_database_interface.h does not contain expected #define: $expected_define"
     exit 1
 fi
+
+set +e
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ matrix:
     - os: osx
       osx_image: xcode6.4
       env: BUILDTYPE=Release
-script: ./travis-compile.sh
-install: ./travis-dependencies.sh
+script: true
+install: true
 cache: apt
 notifications:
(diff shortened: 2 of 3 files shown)


ctrlaltca:
If you're trying to fix this problem: https://travis-ci.org/Cockatrice/Cockatrice/jobs/174493554#L1273 , have a  look at https://github.com/travis-ci/travis-ci/issues/6700

ZeldaZach:
@ctrlaltca Can you write up a change possibly if this isn't satisfactory?

servatrice/check_schema_version.sh
```diff
@@ -31,3 +31,5 @@ if ! grep -q "$expected_define" servatrice/src/servatrice_database_interface.h;
     echo "servatrice_database_interface.h does not contain expected #define: $expected_define"
     exit 1
 fi
+
+set +e
```
Daenyth: This doesn't make sense at all. There's no statements which follow this, so I don't see how this can have any effect. Can you elaborate on what this fixes specifically? If you could point me to any bash documentation that shows how this would affect things that would help too.

I guess we'll also see once the travis builds finish

Daenyth:
Actually this makes even less sense. The file modified in this PR isn't even being executed in the failing builds, the error is about a travis internal thing.

ctrlaltca:
Looks like Travis fixed their problems and now builds are working.
Thank you @cotsog anyway

PR #2294, opened 2016-12-04
Add enable/disable log query functionality

woogerboy21:
This change adds the ability for server owners to allow/disallow log querying
from inside the client.  In the event the functionality is not allowed a
result is returned indicating the functionality is disabled.

image


woogerboy21:
One thing to mention, the servers ini configuration option is changeable on the fly so allowing queries and then disabling access can happen without requiring the server to restart.

tooomm:
The info message is not translatable at the moment. Can you add that?

Daenyth:
Reasonable quick fix. This shouldn't go into the protocol

woogerboy21:
It doesn't?  It simply returns a hard coded string as a result as if the database returned the value and gets appended into the protocol's response container. 

Or were you just stating that it doesnt?

Daenyth:
I was stating that it doesn't and I approve of that

PR #2299, opened 2016-12-06
Update sessions table columns for consistency

### LINKED ISSUE #2276: ip_address in cockatrice_sessions is too short
Currently, the `ip_address` column in `cockatrice_sessions` is a `char(15)`, which is fine for storing IPv4 addresses, but it is too short to store IPv6 addresses, which can be as long as 39 chars.

In order to store IPv6 addresses, the `ip_address` column should be changed to a `varchar(39)`.

woogerboy21:
Fix #2276
This update changes two columns in the sessions table to match the
column declartations in the users table.

### DIFF
--- a/servatrice/migrations/servatrice_0018_to_0019.sql
+++ b/servatrice/migrations/servatrice_0018_to_0019.sql
@@ -1,0 +1,6 @@
+-- Servatrice db migration from version 18 to version 19
+
+alter table cockatrice_sessions modify column `user_name` varchar(35) NOT NULL;
+alter table cockatrice_sessions modify column `ip_address` varchar(255) NOT NULL;
+
+UPDATE cockatrice_schema_version SET version=19 WHERE version=18;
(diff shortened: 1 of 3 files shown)


woogerboy21:
Though this PR changes the column value(s) to match between the two tables and I have tested these changes on a test box I would like some one that is more well versed in databases to take a look at the column changes and verify that removing the formatting wont be to catastrophic.

woogerboy21:
If I recall correctly the failures for this are expected but @ctrlaltca would be better suited to answer it.  Since the DB version changes the build box DB needs updated to properly test the PR (if memory serves correctly).

Daenyth:
> Schema version 18 does not equal new version 19

You forgot to update one of the values.

Daenyth:
servatrice.sql

woogerboy21:
:+1:  Good catch.  Always seem to overlook that for whatever reason.

ZeldaZach:
Upon my review of literature I could not find a case where `char(X) -> varchar(Y), where Y>X` will cause loss of data.

PR #2300, opened 2016-12-06
Fix Store Replays Functionality

woogerboy21:
For whatever reason we have a variable in the servers ini to
enable/disable the storing of replays but there is no code that uses
that variable.  At one time there was but in light of it being removed
some were along the line, this change add's the ability back in for server
owners to disable the storing of game replay data.


woogerboy21:
im actually going to close this and change it else were.....

PR #2301, opened 2016-12-07
Fix store replays

woogerboy21:
Fixed the server side store replay enable/disable functionality.  Not sure when it was broken but we have a variable definition that allows owners to enable/disable the storing of replays but no underlying code that does the actual work.  This fixes that.

### DIFF
--- a/common/server_game.cpp
+++ b/common/server_game.cpp
@@ -153,8 +153,7 @@ void Server_Game::storeGameInformation()
     server->clientsLock.lockForRead();
     while (allUsersIterator.hasNext()) {
         Server_AbstractUserInterface *userHandler = server->findUser(allUsersIterator.next());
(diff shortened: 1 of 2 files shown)


common/server_game.cpp
```diff
@@ -156,12 +154,15 @@ void Server_Game::storeGameInformation()
     while (allUsersIterator.hasNext()) {
         Server_AbstractUserInterface *userHandler = server->findUser(allUsersIterator.next());
         if (userHandler)
-            userHandler->sendProtocolItem(*sessionEvent);
+            if (server->getStoreReplaysEnabled())
```
woogerboy21: I know you were going to suggest that.  The reason I didnt was due to me working on adding additional code in that block area that I ended up backing out.

PR #2304, opened 2016-12-08
New icons

ZeldaZach:
These are some new SVG icons for the client for when donor/vip is implemented.


ZeldaZach:
Will sub-merge into a @woogerboy21 PR.

PR #2305, opened 2016-12-08
New privuser gummys

woogerboy21:
This PR add's new gummy images based on the users privilege level.  For VIP's there is now a star in front of the gummy. And for registered user that are donor's, there is a purple gummy.   Sample image:

image

### DIFF
--- a/cockatrice/src/chatview.cpp
+++ b/cockatrice/src/chatview.cpp
@@ -160,7 +160,7 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
                 const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
                 QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
                 cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel,
-                                                                            buddyList.contains(sender)).toImage());
+                                                                            buddyList.contains(sender)).toImage(), QString::fromStdString(tabSupervisor->getUserInfo()->privlevel()));
                 cursor.insertText(" ");
             }
             cursor.setCharFormat(senderFormat);
--- a/cockatrice/src/chatview.cpp
+++ b/cockatrice/src/chatview.cpp
@@ -146,7 +146,8 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
             (sender == QString::fromStdString(tabSupervisor->getUserInfo()->name()))) {
             senderFormat.setForeground(QBrush(getCustomMentionColor()));
             senderFormat.setFontWeight(QFont::Bold);
-        } else {
+        }
+        else {
             senderFormat.setForeground(QBrush(OTHER_USER_COLOR));
             if (playerBold)
                 senderFormat.setFontWeight(QFont::Bold);
@@ -155,12 +156,14 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
         senderFormat.setAnchorHref("user://" + QString::number(userLevel) + "_" + sender);
         if (sameSender) {
             cursor.insertText("    ");
-        } else {
+        }
+        else {
             if (!sender.isEmpty() && tabSupervisor->getUserListsTab()) {
                 const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
                 QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
-                cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel,
-                                                                            buddyList.contains(sender)).toImage());
+                QMap<QString, UserListTWI *> userList = tabSupervisor->getUserListsTab()->getAllUsersList()->getUsers();
+                UserListTWI *vlu = userList.value(sender);
+                cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel, buddyList.contains(sender), QString::fromStdString(vlu->getUserInfo().privlevel())).toImage());
                 cursor.insertText(" ");
             }
             cursor.setCharFormat(senderFormat);
--- a/cockatrice/src/chatview.cpp
+++ b/cockatrice/src/chatview.cpp
@@ -146,8 +146,7 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
             (sender == QString::fromStdString(tabSupervisor->getUserInfo()->name()))) {
             senderFormat.setForeground(QBrush(getCustomMentionColor()));
             senderFormat.setFontWeight(QFont::Bold);
-        }
-        else {
+        } else {
             senderFormat.setForeground(QBrush(OTHER_USER_COLOR));
             if (playerBold)
                 senderFormat.setFontWeight(QFont::Bold);
@@ -156,8 +155,7 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
         senderFormat.setAnchorHref("user://" + QString::number(userLevel) + "_" + sender);
         if (sameSender) {
             cursor.insertText("    ");
-        }
-        else {
+        } else {
             if (!sender.isEmpty() && tabSupervisor->getUserListsTab()) {
                 const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
                 QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
--- a/cockatrice/src/chatview.cpp
+++ b/cockatrice/src/chatview.cpp
@@ -160,7 +160,7 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType, 
                 const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
                 QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
                 cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel,
-                                                                            buddyList.contains(sender)).toImage(), QString::fromStdString(tabSupervisor->getUserInfo()->privlevel()));
+                                                                            buddyList.contains(sender)).toImage());
                 cursor.insertText(" ");
             }
             cursor.setCharFormat(senderFormat);
--- a/cockatrice/src/pixmapgenerator.cpp
+++ b/cockatrice/src/pixmapgenerator.cpp
@@ -112,21 +112,29 @@ QPixmap CountryPixmapGenerator::generatePixmap(int height, const QString &countr
 
 QMap<QString, QPixmap> CountryPixmapGenerator::pmCache;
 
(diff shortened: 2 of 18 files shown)


ZeldaZach:
Fixed pawns in chat @ https://gist.github.com/ZeldaZach/19cb54e32534c5ff363c050b3d1e012b

cockatrice/src/pixmapgenerator.cpp
```diff
@@ -112,26 +112,35 @@ QPixmap CountryPixmapGenerator::generatePixmap(int height, const QString &countr
...
-QPixmap UserLevelPixmapGenerator::generatePixmap(int height, UserLevelFlags userLevel, bool isBuddy)
+QPixmap UserLevelPixmapGenerator::generatePixmap(int height, UserLevelFlags userLevel, bool isBuddy, QString privLevel)
 {
-
-    int key = height * 10000 + (int) userLevel + (int) isBuddy;
+    int privkeysize = privLevel.size();
+    if (privLevel.size() == 0)
+        privkeysize = 1;
+    
+    int key = height * 10000 + (int)userLevel + (int)isBuddy + privkeysize;
```
ZeldaZach: This can easily collide as userLevel, isBuddy, and privkeysize are all single numbers... Try changing to something like `int key = height * 10000 + (int)userLevel*88 + (int)isBuddy*54 + privkeysize*31;`

cockatrice/src/userlist.cpp
```diff
@@ -276,7 +276,7 @@ void UserListTWI::setUserInfo(const ServerInfo_User &_userInfo)
     userInfo = _userInfo;
 
     setData(0, Qt::UserRole, userInfo.user_level());
-    setIcon(0, QIcon(UserLevelPixmapGenerator::generatePixmap(12, UserLevelFlags(userInfo.user_level()), false)));
+    setIcon(0, QIcon(UserLevelPixmapGenerator::generatePixmap(12, UserLevelFlags(userInfo.user_level()), false, QString::fromStdString(userInfo.privlevel()))));
```
ZeldaZach: This won't work for offline users due to server configurations in the query ordering. I've fixed it in https://gist.github.com/ZeldaZach/d901c34d770d75618244a6489125b201

cockatrice/src/pixmapgenerator.cpp
```diff
@@ -112,26 +112,35 @@ QPixmap CountryPixmapGenerator::generatePixmap(int height, const QString &countr
...
-QPixmap UserLevelPixmapGenerator::generatePixmap(int height, UserLevelFlags userLevel, bool isBuddy)
+QPixmap UserLevelPixmapGenerator::generatePixmap(int height, UserLevelFlags userLevel, bool isBuddy, QString privLevel)
 {
-
-    int key = height * 10000 + (int) userLevel + (int) isBuddy;
+    int privkeysize = privLevel.size();
+    if (privLevel.size() == 0)
+        privkeysize = 1;
+    
+    int key = height * 10000 + (int)userLevel * 88 + (int)isBuddy * 54 + privkeysize * 31;
```
ctrlaltca: These magic numbers are quite hard to understand; also, using the length of `privLevel` as a key will cause collisions if in the future we'll add new levels.
Maybe a better idea is to change the `pmCache` map to use string keys:
```cpp
QMap<QString, QPixmap> UserLevelPixmapGenerator::pmCache;
```
and then generate the key like this:
```cpp
QString key = QString::number(height * 10000 + (int) userLevel + (int) isBuddy) + privLevel;
```

cockatrice/src/chatview.cpp
```diff
@@ -146,7 +146,8 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType,
             (sender == QString::fromStdString(tabSupervisor->getUserInfo()->name()))) {
             senderFormat.setForeground(QBrush(getCustomMentionColor()));
             senderFormat.setFontWeight(QFont::Bold);
-        } else {
+        }
+        else {
```
ctrlaltca: Check braces style: https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md#braces

cockatrice/src/chatview.cpp
```diff
@@ -155,12 +156,14 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType,
         senderFormat.setAnchorHref("user://" + QString::number(userLevel) + "_" + sender);
         if (sameSender) {
             cursor.insertText("    ");
-        } else {
+        }
+        else {
```
ctrlaltca: Check braces style: Check braces style: https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md#braces

cockatrice/src/chatview.cpp
```diff
@@ -155,12 +156,14 @@ void ChatView::appendMessage(QString message, RoomMessageTypeFlags messageType,
...
             if (!sender.isEmpty() && tabSupervisor->getUserListsTab()) {
                 const int pixelSize = QFontInfo(cursor.charFormat().font()).pixelSize();
                 QMap<QString, UserListTWI *> buddyList = tabSupervisor->getUserListsTab()->getBuddyList()->getUsers();
-                cursor.insertImage(UserLevelPixmapGenerator::generatePixmap(pixelSize, userLevel,
-                                                                            buddyList.contains(sender)).toImage());
+                QMap<QString, UserListTWI *> userList = tabSupervisor->getUserListsTab()->getAllUsersList()->getUsers();
```
ctrlaltca: This works fine, but it will need some refactoring in the future; right now we are retrieving data from the `QGroupBox` in the userlist tab, instead we should have a generic "connection" object that provides this data.
This is related to point 2 discussed in PR https://github.com/Cockatrice/Cockatrice/pull/2267

cockatrice/src/userlist.cpp
ctrlaltca: Can you explain what you mean?

woogerboy21: Since i'm not as skilled as most you guys with git/gist.  How would I compare what currently exists to what has changed here to see what you updated?

woogerboy21:
>> Can you explain what you mean?

Wonder why I can't reply to this?

cockatrice/src/userlist.cpp
woogerboy21: It looks like the evalUserQueryResult function has a method to pass in if you want all the user data or just some of it.  With the current way its written privlevel is only returned when all the data is requested.  Which doesnt happen when a user is offline.

woogerboy21: @ctrlaltca

woogerboy21: @ZeldaZach Even with this change I see the following when the user is offline:
image

cockatrice/src/pixmapgenerator.cpp
ZeldaZach: These were magic numbers added to avoid collisions, but a QString key should work too

cockatrice/src/chatview.cpp
ZeldaZach: I knew this would need changes. My first goal was to _make this work_ and from there try and figure out a more optimal solution.

cockatrice/src/userlist.cpp
ZeldaZach: The gist only re-ordered the query to have the privlevel called each time. Maybe I did something wrong. Can you manual the queries at hand and see what returns?

ctrlaltca: Now i understand the problem, thank you

woogerboy21: This should be fixed now with the most recent commit.

woogerboy21:
Please re-test this PR and make sure its working as expected before committing

cockatrice/src/userlist.cpp
ZeldaZach: Seems like I'm getting better at this ^ 👍

cockatrice/src/chatview.cpp
ZeldaZach: I'm fine with this as is until we can come up with a better solution in the future. No need to withhold this PR for it.

cockatrice/src/pixmapgenerator.cpp
```diff
@@ -112,32 +112,38 @@ QPixmap CountryPixmapGenerator::generatePixmap(int height, const QString &countr
...
-    else
+        if (privLevel.toLower() != "none")
+            levelString.append("_" + privLevel.toLower());
+    } else
         levelString = "normal";
 
     if (isBuddy)
         levelString.append("_buddy");
-
+
```
ZeldaZach: Can remove this extra space

cockatrice/src/chatview.cpp
Daenyth: I think that's more or less what he was saying

cockatrice/src/pixmapgenerator.cpp
```diff
@@ -112,32 +112,38 @@ QPixmap CountryPixmapGenerator::generatePixmap(int height, const QString &countr
 
 QMap<QString, QPixmap> CountryPixmapGenerator::pmCache;
 
-QPixmap UserLevelPixmapGenerator::generatePixmap(int height, UserLevelFlags userLevel, bool isBuddy)
+QPixmap UserLevelPixmapGenerator::generatePixmap(int height, UserLevelFlags userLevel, bool isBuddy, QString privLevel)
 {
-
-    int key = height * 10000 + (int) userLevel + (int) isBuddy;
+    
+    QString key = QString::number(height * 10000 + (int)userLevel + (int)isBuddy) + privLevel;
```
Daenyth: If `userLevel` can be 2 or 1, then this collides because `2 + 0 == 1 + 1`

Why not just `height + ":" + userLevel + ":" + isBuddy` or so?

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -505,38 +505,38 @@ ServerInfo_User Servatrice_DatabaseInterface::evalUserQueryResult(const QSqlQuer
...
             result.set_accountage_secs(accountAgeInSeconds);
         }
 
-        const QString email = query->value(8).toString();
+        const QString email = query->value(9).toString();
         if (!email.isEmpty())
             result.set_email(email.toStdString());
 
-        const QString clientid = query->value(9).toString();
+        const QString clientid = query->value(10).toString();
```
Daenyth: If there's a way in the future to get value from the column name of the query result we wouldn't have to keep bumping numbers like this. Like if you could `query->value("clientId")`. This might already be possible, I haven't checked the docs.

No change needed for now, maybe something to look at later

Daenyth:
To be clear, my 'approve' is on the general approach; we still need to fix the key collision

cockatrice/src/pixmapgenerator.cpp
woogerboy21: userLevel can be 0,1,2 and privelevel currently can be 3,4,7

ZeldaZach: I agree, this should be 
`QString key = QString::number(height * 10000) + ":" + (int)userLevel + ":" + (int)isBuddy + ":" + privLevel;`

servatrice/src/servatrice_database_interface.cpp
ZeldaZach: Sounds good to me

ZeldaZach:
Once the change asked for by Dae is added I'm ready to merge this in. I left a comment with what it should look like.

PR #2311, opened 2016-12-08
Document '--config' parameter

tritoch:
servatrice.ini can be passed to servatrice via the --config parameter.  I don't think this is documented anywhere.


ZeldaZach:
Maybe also document on the wiki?

ZeldaZach:
@tritoch You deleted your branch so I can't merge this.

PR #2312, opened 2016-12-09
Gold Fringe for VIP-Buddy

tritoch:
VIP+Another Classification currently has same image as Non-VIP.  Added gold fringe to VIP editions for differentiation.


ZeldaZach:
There's no real effective way to do this since the image is so small. I'd rather just leave buddy stars as they are.

PR #2320, opened 2016-12-16
Move related card creation out of sub menu

### LINKED ISSUE #2317: remove submenu "create related card"
Suggestion:

It's nice to see the "create related card"-feature implemented. It saved quite some of my lifetime already =). However I think it would be reasonable to optimize the menu a bit with a simple step.

I suggest, to remove the menuitem "create related card...". That means instead of a Menu looking like:

... 
Flip
Create related card
 - 1/1 Soldier Token
 - Elspeth Emblem
attach to a card...
...

The Menu should look like

... 
Flip
1/1 Soldier Token
Elspeth Emblem
attach to a card...
...

as most of these Menu have only one entry. Some have two. But only very few cards have more than that. This would be quite appropriate, as this menu entry is probably one of the most used ones.

ZeldaZach:
Fix #2317 

This changes how the sub-menu works relating to tokens. Instead of an additional sub-menu, there will now be additional components to the main menu that say "Token: n/n Example".

screenshot 2016-12-16 03 52 05
screenshot 2016-12-16 03 52 15

Works on both the stack and play ground.

PS: CLion 2016.3 started doing auto cmake builds, added that folder to gitignore.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -107,12 +107,12 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
 {
     userInfo = new ServerInfo_User;
     userInfo->CopyFrom(info);
-    
+
     connect(settingsCache, SIGNAL(horizontalHandChanged()), this, SLOT(rearrangeZones()));
     connect(settingsCache, SIGNAL(handJustificationChanged()), this, SLOT(rearrangeZones()));
-    
+
     playerArea = new PlayerArea(this);
-    
+
     playerTarget = new PlayerTarget(this, playerArea);
     qreal avatarMargin = (counterAreaWidth + CARD_HEIGHT + 15 - playerTarget->boundingRect().width()) / 2.0;
     playerTarget->setPos(QPointF(avatarMargin, avatarMargin));
@@ -126,7 +126,7 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
     HandCounter *handCounter = new HandCounter(playerArea);
     handCounter->setPos(base + QPointF(0, h + 10));
     qreal h2 = handCounter->boundingRect().height();
-    
+
     PileZone *grave = new PileZone(this, "grave", false, true, playerArea);
     grave->setPos(base + QPointF(0, h + h2 + 10));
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/player.cpp
```diff
@@ -1104,13 +1104,14 @@ void Player::actCreateRelatedCard()
     QAction *action = static_cast<QAction *>(sender());
 
     // removes p/t from tokens (and leading space))
-    QStringList spaces = action->text().split(" ");
+    // Added split for "Token:" due to change in PR fixing #2317
+    QStringList spaces = action->text().split(tr("Token: "))[1].split(" ");
```
tooomm: why not `Create Token: `?

ZeldaZach: Went with what Gavin said in a comment on the ticket.

Daenyth: I don't have strong feelings, but I suggested `token` for text width reasons

ZeldaZach:
@Daenyth It looks no different then normal. The "extra section" only gets added if reverse related cards are found. I'm traveling atm so I can't get screenshots, but you can trust that. I've manually tested with cards that make 1 token, cards that make numerous tokens and planeswalker flip cards. Additional tests are welcomed, but I believe I covered the spectrum. The code for the stack and table is the same, but testing on both wouldn't hurt.

I don't believe wiki changes are needed as this feature is just how to make tokens in client, but I'll take a look at remediate as necessary.

Daenyth:
:shipit:

cockatrice/src/player.cpp
tooomm: I don't think that 6 additional characters are a very big deal.

It could be just "Create 3/3 Elephant" (same as the new card text btw) as well... I think it helps the user a lot and ui/ux wise it's better if the feature label is self-explaining.
While writing this, I even think it's the superior way to go.

ZeldaZach: I can change it to "Create" if that's what everyone likes

tooomm:
>I can change it to "Create" if that's what everyone likes

So?! https://github.com/Cockatrice/Cockatrice/pull/2320#discussion_r92925918

How comes that you merge 4 minutes after your comment without any more interactions in the pr? 👎

tooomm:
>I can change it to "Create" if that's what everyone likes

So?! https://github.com/Cockatrice/Cockatrice/pull/2320#discussion_r92925918

How comes that you merge 4 minutes after your comment without any more interactions in the pr? 👎 

*Edit:
I don't want to be the bad guy here, but I don't see a reason nor a approval of the new input/proposal.*

ZeldaZach:
I asked in client for opinions and everybody who was surveys liked "token" better as it was more clear of what would be done. I can always directly change it in the future if there's a disagreement.

tooomm:
Well, I still don't prefer it. But that's ok. ;)

Just let us know next time, please! For traceability reasons.

Daenyth:
We can always change it later. This is good for now, no reason to hold up
the change just on the wording.

PR #2348, opened 2017-01-06
Condensed Card Counter Context Menu

poixen:
## Description
Context menu is used extensively, so should be as tight as possible. Currently _half_ of the menu is dedicated to the same 3 command for adding counters:
screen shot 2017-01-06 at 19 02 15
My solution condenses these colors into their own menus:
screen shot 2017-01-06 at 19 08 55
screen shot 2017-01-06 at 19 09 04
screen shot 2017-01-06 at 19 09 11

## Changelist

+ Cosmetic overhaul of the card counter context menu

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -698,13 +698,13 @@ void Player::retranslateUi()
     counterColors.append(tr("Green"));
 
     for (int i = 0; i < aAddCounter.size(); ++i){
-        aAddCounter[i]->setText(tr("&Add counter (%1)").arg(counterColors[i]));
+        aAddCounter[i]->setText(tr("%1").arg(counterColors[i]));
     }
     for (int i = 0; i < aRemoveCounter.size(); ++i){
-        aRemoveCounter[i]->setText(tr("&Remove counter (%1)").arg(counterColors[i]));
+        aRemoveCounter[i]->setText(tr("%1").arg(counterColors[i]));
     }
     for (int i = 0; i < aSetCounter.size(); ++i){
-        aSetCounter[i]->setText(tr("&Set counters (%1)...").arg(counterColors[i]));
+        aSetCounter[i]->setText(tr("%1...").arg(counterColors[i]));
     }
 
     aMoveToTopLibrary->setText(tr("&Top of library"));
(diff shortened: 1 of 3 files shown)


tooomm:
simple, yet smart... ❤️

cockatrice/src/player.cpp
```diff
@@ -698,13 +698,13 @@ void Player::retranslateUi()
...
-        aAddCounter[i]->setText(tr("&Add counter (%1)").arg(counterColors[i]));
+        aAddCounter[i]->setText(tr("%1").arg(counterColors[i]));
     }
     for (int i = 0; i < aRemoveCounter.size(); ++i){
-        aRemoveCounter[i]->setText(tr("&Remove counter (%1)").arg(counterColors[i]));
+        aRemoveCounter[i]->setText(tr("%1").arg(counterColors[i]));
     }
     for (int i = 0; i < aSetCounter.size(); ++i){
-        aSetCounter[i]->setText(tr("&Set counters (%1)...").arg(counterColors[i]));
+        aSetCounter[i]->setText(tr("%1...").arg(counterColors[i]));
```
tooomm: I'm wondering why exactly we have this `...` dots sometimes in the menus?
What's the intend/design behind them?

Because it's not completely uniform in all menus or submenus!

poixen: I agree. I think the idea is that `...` will open a new popup/menu.
If you compare to `Tap` that is a single command with no additional input.

ZeldaZach:
i like. Good job!

cockatrice/src/player.cpp
Daenyth: > the idea is that ... will open a new popup/menu.

Exactly

bjholmes3:
This new UI is very clean and organized, great job! I do have one thought, for some playstyles which make frequent use of counters (particularly suspends and cumulative upkeeps where setting and adding/removing counters happens multiple times per turn), it can be a lot of extra menuing to do these things. I propose perhaps adding one extra field, "Set Counter Color" so that Add, Remove, and Set Counters can go back to being options to be completed in the main right click menu.

Nightfirecat:
@bjholmes3 I'd personally recommend using keyboard shortcuts

Jinxstands4Jinx:
This was much better the way it was. please revert

Issue #2342, opened 2017-01-04
Adjust/append the 'user level' information in details accordingly


tooomm:
If you right click somebody, the `User details`  looks like this for a donator:
![user_infor](https://cloud.githubusercontent.com/assets/9874850/21647574/29aece94-d29b-11e6-91fb-070d92cf5968.png)

**It correctly displays for Admin and Moderator, but not for the new ranks.**

Since you needs to be registered for all ranks, I suggest to think about something like this:
Administrator / VIP (used to be Administrator)
Administrator
Moderator / VIP (used to be Moderator)
Moderator
Donator (used to be Registered user)
VIP (used to be Registered user)
Normal (used to be Registered user)
Unregistered

skwerlman:
AFAIK, this is because donator/VIP aren't ranks in the same sense as admin or mod. they instead are more like "rank modifiers".

I agree that donator/VIP should be displayed there instead of registered user.

woogerboy21:
> AFAIK, this is because donator/VIP aren't ranks in the same sense as admin or mod. they instead are more like "rank modifiers".

This is correct.  The user levels are still unreg/reg/mod/admin .  Maybe user level in the UI is slightly misleading at this point with the introduction of the new code functionality.

tooomm:
If it's even not that clear to people close to the project - and I hear some uncertainty in recent times - how can it be understood by the ordinary user, mods or other server admins (e.g. for smaller playgroups or even other public servers, whose owners doesn't happen to hang around here all the time)?

Maybe we should clear up things a bit for all stakeholders...

woogerboy21:
@tooomm 
Im not sure how its unclear.  It so far seems clear to me and @skwerlman .  It sounds to me like your just missing the understanding or trying to accomplish some goal im not understanding.  You wanted some indicator that the user is of a different user level (its there as registered user) and you would like to understand the users priv level (its there as the gummy icon).

The app has a rank (your terms not mine) of unregistered user, registered user, moderator, and admin.  And with in each of those there is a sub tier.  You have NONE, VIP, or DONATOR.  Not sure how unclear that is?

woogerboy21:
So you have something like this:

Administrator
  - VIP
  - DONATOR
  - NONE
Moderator
  - VIP
  - DONATOR
  - NONE

Registered User
  - VIP
  - DONATOR
  - NONE

Unregistered

[...]

PR #2349, opened 2017-01-06
Added privlevel to User Details

poixen:
Fix for #2342

Before:
screen shot 2017-01-06 at 20 33 18

After:
screen shot 2017-01-06 at 22 16 24

### DIFF
--- a/cockatrice/src/userinfobox.cpp
+++ b/cockatrice/src/userinfobox.cpp
@@ -110,6 +110,11 @@ void UserInfoBox::updateInfo(const ServerInfo_User &user)
         userLevelText = tr("Registered user");
     else
         userLevelText = tr("Unregistered user");
+
+    if (user.has_privlevel() && user.privlevel() != "NONE") {
+        userLevelText += " | " + tr("%1").arg(user.privlevel().c_str());
+    }
+
     userLevelLabel3.setText(userLevelText);
 
     QString accountAgeString = tr("Unregistered user");


Nightfirecat:
Would personally prefer to see `<userLevel> (<privLevel>)`, like **Registered user (DONATOR)**; at minimum, there should be spaces around the vertical bar, not just after it imo.

poixen:
Thanks for feedback @Nightfirecat!
@woogerboy21 @Daenyth and myself had a talk about it on the gitter page and came up with this version.

tritoch:
I think `Donor` is the proper English word versus `Donator`

woogerboy21:
http://www.thefreedictionary.com/donator

```
donator - noun
A person who gives to a charity or cause:
```

poixen:
@woogerboy21 I pull the string from the `ServerUser_Info`. So it can be whatever you like, dont you just update the DB?

woogerboy21:
> I pull the string from the ServerUser_Info. So it can be whatever you like, dont you just update the DB?

Yes.

cockatrice/src/userinfobox.cpp
```diff
@@ -110,6 +110,11 @@ void UserInfoBox::updateInfo(const ServerInfo_User &user)
         userLevelText = tr("Registered user");
     else
         userLevelText = tr("Unregistered user");
+
+    if (user.has_privlevel() && user.privlevel() != "NONE") {
+        userLevelText += tr(" | %1").arg(user.privlevel().c_str());
```
tooomm: Can you exclude " | " from tr() please? Actually there is nothing to translate at all here... ;P

tooomm:
Do you save them as all capital letter in the db @woogerboy21? --> DONATOR and VIP?
Just wondering. :)

Cool pr @poixen! Pretty busy today, he? 👍

woogerboy21:
> Do you save them as all capital letter in the db?

Yes they are stored in all caps in the DB.

PR #2350, opened 2017-01-06
Add email registration restriction

woogerboy21:
This change add's the ability for an operator to restrict the number of accounts that can be registered using any single email address.

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -444,7 +444,7 @@ void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quin
             QMessageBox::critical(this, tr("Registration denied"), tr("It's mandatory to specify a valid email address when registering."));
             break;
         case Response::RespTooManyRequests:
-            QMessageBox::critical(this, tr("Registration denied"), tr("Too many registration attempts from your IP address."));
+            QMessageBox::critical(this, tr("Registration denied"), tr("Too many registration attempts, please try again later or contact the server operator for further details."));
             break;
         case Response::RespPasswordTooShort:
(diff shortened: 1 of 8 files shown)


cockatrice/src/window_main.cpp
```diff
@@ -444,7 +444,7 @@ void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quin
             QMessageBox::critical(this, tr("Registration denied"), tr("It's mandatory to specify a valid email address when registering."));
             break;
         case Response::RespTooManyRequests:
-            QMessageBox::critical(this, tr("Registration denied"), tr("Too many registration attempts from your IP address."));
+            QMessageBox::critical(this, tr("Registration denied"), tr("Too many registration attempts, please try again later or contact the server operator for further details."));
```
tooomm: Should it say that the problem is the repeated use of the same mail address?

woogerboy21: I debated on adding a custom message stating to many email registration attempts but decided against it for 2 reasons.  The first is it would require a client update to understand the error properly.  This way the error on older clients would say they have to many attempts from there IP and once updated the error would revert to what you see.

The second reason is I wasn't sure how much details we wanted to give as to why the client was denied registering an account only because the registration functionality could be susceptible to  DOS attacks and to tell them why it was being denied would only enable whomever to continue to use that method for attack.

If its not a concern to everyone I can add a custom response stating email in the reasoning.

tooomm: Thanks for clarification. 👍 
What do the others think?

Daenyth: This is fine

Issue #1746, opened 2016-01-05
RPM Builds


TMiguelT:
I'm not sure if this has been considered, but should we do builds for Red Hat based OSs like Fedora? Would this require much effort?

Daenyth:
That would be difficult because the CI we use for the builds is on ubuntu. Possible, but probably not worth the effort. Anyone on an rpm-based distro should be comfortable with compiling.

TMiguelT:
But the CI we use also produces Windows and Mac builds, no? So surely it's possible?

Just to clarify I'm willing to put in the actual work to have RPM builds, I'd just need some pointers on how to do so. I'm on an RPM-based distro and I did build cockatrice but it would certainly be easier to have an automated update process!

ctrlaltca:
I've tried this in the past, but not being an user of an rpm-distro, i left the job half-cooked: https://build.opensuse.org/project/show/home:ctrlaltca:Cockatrice

Daenyth:
> But the CI we use also produces Windows and Mac builds, no

Travis has Ubuntu 14.04 and 12.04 build platforms, as well as OSX.
Appveyor has windows.

If you're aware of a CI that can do this I'm open to it. I suppose a last resort might be to compile it inside docker on travis, but the travis builds already take long enough...

Daenyth:
Can you make a specfile that just pulls in the git repo?

TMiguelT:
So is this BuildOpenSuse site a CI for rpm-using OSs ? I'll look into it if
that could work

TMiguelT:
@ctrlaltca it seems like this was almost finished? It just failed on xUbuntu and CentOS because they were missing some dependencies (which I suspect you can find anyway)

ctrlaltca:
Yep, it was mostly working, but totally untested.

TMiguelT:
Okay I should be able to test the fedora ones at least. Where are the .rpm downloads?

ctrlaltca:
Here, but they're surely outdated: http://download.opensuse.org/repositories/home:/ctrlaltca:/Cockatrice/Fedora_21/

TMiguelT:
Incidentally I suspect the package name is being set to the description. Also could we package the Cockatrice logo as well? The Software program sees it like this at the moment:

![fedora cockatrice](https://cloud.githubusercontent.com/assets/5019367/12165260/a5aee648-b51c-11e5-8924-5cb4effd1397.png)

[...]

PR #2351, opened 2017-01-07
allow specifying package generator on linux

skwerlman:
use like `cmake .. -DPACKAGE_GENERATOR=RPM`
defaults to DEB

I've never worked with cmake before, so there might be a better way to do this

valid generators are listed here: https://cmake.org/Wiki/CMake:CPackPackageGenerators

should help move #1746 forward


ctrlaltca:
By now we force the use of the "DEB" package generator under linux because we only implemented the CPACK_DEBIAN_PACKAGE_* variables needed to generate a debian package.
In order to get RPM packages to work, you'll need to:
 * add the needed CPACK_RPM_PACKAGE_* variables (https://cmake.org/cmake/help/v3.0/module/CPackRPM.html)
 * drop line 177 https://github.com/Cockatrice/Cockatrice/pull/2351/files#diff-af3b638bc2a3e6c650974192a53c7291L177
 * there's no need to ad a new "PACKAGE_GENERATOR" variable, you can just `cmake .. -DCPACK_GENERATOR=RPM`

So, this PR is mostly useless.

skwerlman:
Thanks for the info; I had no idea it needed those additional variables.
I'll fix and resubmit this.

ctrlaltca:
You can somewhat autodetect if the linux distro is using rpm by checking for the existence of a specific path, eg:
```cmake
    find_file(PACKAGE_TYPE_RPM Packages PATHS /var/lib/rpm/)
    mark_as_advanced(PACKAGE_TYPE_RPM)

    if (PACKAGE_TYPE_RPM)
        set(CPACK_PACKAGE_TYPE RPM)
        # set all the basic CPACK_RPM_* variables
    else()
        set(CPACK_PACKAGE_TYPE DEB)
        # move all the CPACK_DEB_* variables here
    endif()
```

Probably mirroring the existing CPACK_DEB_* variables into their CPACK_RPM_* counterpart is a good start in order to get working packages including basic informations like a name, a version and the needed dependencies.

skwerlman:
I'd like to avoid autodetection, especially in place of a manual flag, since its possible to have the tools for making both debs and rpms on one system.

Daenyth:
I agree. I'd pick a default and let it be manually overridden.

skwerlman:
re-submitted as #2373

PR #2352, opened 2017-01-07
Added registration email blacklist

### LINKED ISSUE #1862: Add ability to prevent certain mail domains for registration
Add the ability to setup domains in the servers .ini file to not allow use for registering accounts.  For example, I would like to prevent users from registering accounts using domains such as 10minutemail.com.

woogerboy21:
Fix #1862 
Added the ability to define email blacklist for user registration.  Now server operators can prevent users from registering accounts that contain providers they do not wish users to use.  This also adds a new response protocol item with new client popup response indicating the reason for denied registration.

image

### DIFF
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -144,6 +144,11 @@ disallowedregexp=""
 ; using the same email address. 0 = Unlimited number of accounts (default).
 ;maxaccountsperemail=0 
 
+; You can prevent users from using certain mail domains for registration.  This setting contains a
+; comma-seperated list of email provider domains that you would like to prevent users from using
+; during registration.  Example: "10minutemail.com,gmail.com,mail.com"
+emailblacklist=""
+
 [smtp]
 
 ; Enable the internal smtp client to send registration emails.  If you would like to
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -147,7 +147,7 @@ disallowedregexp=""
 ; You can prevent users from using certain mail domains for registration.  This setting contains a
 ; comma-seperated list of email provider domains that you would like to prevent users from using
 ; during registration.  Example: "10minutemail.com,gmail.com,mail.com"
-emailblacklist=""
+;emailproviderblacklist=""
 
 [smtp]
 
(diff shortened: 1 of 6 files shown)


servatrice/servatrice.ini.example
```diff
@@ -144,6 +144,11 @@ disallowedregexp=""
 ; using the same email address. 0 = Unlimited number of accounts (default).
 ;maxaccountsperemail=0 
 
+; You can prevent users from using certain mail domains for registration.  This setting contains a
+; comma-seperated list of email provider domains that you would like to prevent users from using
+; during registration.  Example: "10minutemail.com,gmail.com,mail.com"
+emailblacklist=""
```
ZeldaZach: Maybe change the name of the config value to `emailproviderblacklist` so it's a little more clear

woogerboy21:
@ZeldaZach Updated...

woogerboy21:
Sure, give me a bit and Ill see if I can stand something up this afternoon people can bounce tests off of.

ctrlaltca:
Code looks fine, the only note is about the use of "contains" when checking the provided email address against the blacklist: eg. if i blacklist "mail.com", it will implicitly block also "gmail.com", "whatevermail.com", "mail.computers.biz", etc..

I don't see this as a real problem as long as it's stated clearly in the comment before the setting, but the current comment mentions "10minutemail.com,gmail.com,mail.com" that is a misleading example as it contains "mail.com".

ZeldaZach:
@ctrlaltca Nice spotting on that. I don't like the wildcard effect of the blanket blocks (mail.com blocks gmail.com, etc). Can this PR be redone so it blocks _only_ that specific host @woogerboy21 ?

woogerboy21:
@ZeldaZach I would prefer to keep the wildcard affect myself.  Updating the description would be more of a solution that what your suggesting.

tooomm:
Haven't we something like that working here? https://github.com/Cockatrice/Cockatrice/pull/2200
Maybe grab the idea/concept?

woogerboy21:
@tooomm That is for the actual username, we could apply it to the email component of registration as well.  Regex always seems like overkill to me though.

woogerboy21:
@ZeldaZach I have a test box up.  Is there a particular email syntax you would like me prevent from registration on the box?  You should be able to get to the server using the host name testing.woogerworks.com and port 4747.  Its an open box at the moment, almost entirely blank db.  As well as a new DNS record so it might take a moment to replicate out.  Oh and the box is kind of slow so might take a moment to log you in / register.

woogerboy21:
@ZeldaZach Just an FYI on this one. I have email capabilities setup as well as changed the testing server information on the last post.  So when you get around to testing this make sure you have things correct.

PR #2336, opened 2017-01-02
Registered + VIP + Buddy Gummy Update

tritoch:
Made Reg/VIP/Buddy gummy unique (all-gold)

PR #2354, opened 2017-01-09
Gummy remap

tritoch:
This matches [suggestion](https://github.com/Cockatrice/Cockatrice/pull/2336#issuecomment-271054240) on #2336 to have buddy indicator be the gold star and gummies be solid (or b/w stripe for mods) colors.  VIP has gold color, so buddy star is green... not sure which way to go with that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ mysql.cnf
 .DS_Store
 .idea/
 *.aps
+cmake-build-debug/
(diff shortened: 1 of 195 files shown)


tooomm:
After clarifying with @tritoch..
This change introduces the 2nd table out of my suggestions in the linked pr.

Buddys will look like this:
chat-buddy

While I'm totally fine with that, and I would prefer that over big stars... most others seem to like the former big stars for buddys more?

PR #2359, opened 2017-01-11
updated mac wording for bintray deployment

tooomm:
MacOSX --> macOS

### DIFF
--- a/cmake/bintray_deploy.json.in
+++ b/cmake/bintray_deploy.json.in
@@ -22,7 +22,7 @@
     },
     "files": [
         { "includePattern": "build/(Cockatrice.*\\.deb)", "uploadPattern": "Ubuntu/$ENV{DIST}/$1" },
-        { "includePattern": "build/(Cockatrice.*\\.dmg)", "uploadPattern": "MacOSX/$1" }
+        { "includePattern": "build/(Cockatrice.*\\.dmg)", "uploadPattern": "macOS/$1" }
     ],
     "publish": true
 }


Daenyth:
Why?

Also we need to make sure this doesn't break the auto updater

tooomm:
Because it's the official term, because we use it everywhere else (e.g. wiki, wegpage)..?
And finally it's just correct. :)

But yeah, this will most likely kill the auto updater. 👎 That needs to be adressed.
Oh, and @ctrlaltca should double check if there are no other needed changes within bintray and such that might be affected.

ZeldaZach:
I'd rather not account for this change as it doesn't really matter and I don't want to mess with the updated script and risk breaking it again as we have no real way to test it.

ZeldaZach:
So I'd rather hold off on _any_ changes to the auto updater until we're working on the release/beta channel partition. At that point, we can possibly look into introducing something like this. Hope you understand Tom.

Daenyth:
If this change breaks the currently 'in the wild' auto updater I agree we should wait

tooomm:
>If this change breaks the currently 'in the wild' auto updater I agree we should wait

Why is this rumor still around?!
Seems like nobody put any effort into checking closer for 2mins... ;P

As I mentioned on gitter 7 days ago:
>am i totally wrong on #2359 ? the updater is not involved at all there... he checks a totally other bintray project (not cockatrice-git, where we have the os folder which i renamed):
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/update_checker.cpp#L8

No comments on that btw.

This'll change once the release channels are in place for sure (https://github.com/Cockatrice/Cockatrice/pull/2362) because then we check `cockatrice` (for release) AND `cockatrice-git` (for developer builds) on bintray.

PR #2361, opened 2017-01-15
Fix appveyor build names

### LINKED ISSUE #2358: Bintray deployment not working correctly
@woogerboy21 was wondering the other day that there are no dev builds for mac
I thought we had that once and started to search on bintray.

Check this, we used to have them all together...
https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/0.0.1%7Egit20161221.62dc6dc#files

but now it's mostly split up like this: (or sometimes no mac/linux at all)
https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/0.0.1%7Egit20161230.f7c8651d#files (win only)
https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/0.0.1%7Egit20161230.f7c8651#files (mac+trusty)

While for the most recent builds... it's just windows mostly all the time:
https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/0.0.1%7Egit20170108.f2481187#files
But there happen to be some builds that have mosty identical commit numbers, just a letter appended and they have both, but separated into Windows and Mac/Linux
https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/0.0.1%7Egit20170107.a85b0dc#files
https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/0.0.1%7Egit20170107.a85b0dcb#files 

Something is wonky here :)

<br>
Another thing that I realized is, that for some reason the vcs tag is not always there. We pass it here: https://github.com/Cockatrice/Cockatrice/blob/master/cmake/bintray_deploy.json.in#L20<br>
but it doesn't always show up. Check on of the split up links with just one letter appended for an example.
The mac/linux link has it, while the windows one hasn't.
The same goes for the description. mac/linux has it - windows doesn't.

ctrlaltca:
It seems that the `git` binary from Appveyor now defaults to 8 chars for git short commit hash, instead of 7 as currently used by github and travis.
This caused the cockatrice package to have a different name and version, so when uploaded to bintray it ended up in a different folder.
This PR forces the use of 7 chars.

Fix #2358

### DIFF
--- a/cmake/getversion.cmake
+++ b/cmake/getversion.cmake
@@ -1,7 +1,7 @@
 find_package(Git)
 if(GIT_FOUND)
 	execute_process(
-		COMMAND ${GIT_EXECUTABLE} log -1 --date=short "--pretty=%h"
+		COMMAND ${GIT_EXECUTABLE} log -1 --abbrev=7 --date=short "--pretty=%h"
 		WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
 		RESULT_VARIABLE res_var
 		OUTPUT_VARIABLE GIT_COM_ID


tooomm:
The vcs tag seems to be working correctly as well now! 👍

Daenyth:
Great find, thanks ctrl

Issue #2338, opened 2017-01-03
Infinite update loop


tooomm:
<b>OS:</b>
Win 10 Pro 64bit

<b>Cockatrice version:</b>
Cockatrice: All Hallow's Eve, Revision 1 - f7c8651d (2016-12-30)
**installed with in-client updater**

___
<br>

After installing the new release, I keep getting update notifications from `Help --> Check for updates`

![updater](https://cloud.githubusercontent.com/assets/9874850/21614510/fae4beaa-d1d9-11e6-9475-dcd3e6906be5.png)

woogerboy21:
Does anyone that works with the bintray stuff understand how the file at https://api.bintray.com/packages/cockatrice/Cockatrice/Cockatrice/files is generated?  If I am reading it correctly it shows a client version available from 01/01/2017 on the windows platform yet we dont have that client version anywere that I can see.

tooomm:
For clearness, the content of the bintray api call for the win64 build reads:
```
{
"name":"Cockatrice_All_Hallows_Eve_Revision1_win64.exe",
"path":"Cockatrice_All_Hallows_Eve_Revision1_win64.exe",
"repo":"Cockatrice",
"package":"Cockatrice",
"version":"2016-12-31_All_Hallows_Eve_Revision1",
"owner":"cockatrice",
"created":"2017-01-01T22:05:50.113Z",
(...)
}
```

>"version":"**2016-12-31**(...)"
vs.
"created":"**2017-01-01**(...)"


<br>
Question stands:

>Does anyone that works with the bintray stuff understand how the file (...) is generated?

Where does the date in "version" and the date for "created" exactly comes from?
How can we avoid differences? Or do we need to change the handling in the updater? Did bintray change something again?

woogerboy21:
@tooomm 
The code block for the updater for finding the client info from bintray looks like this:
```
QDate UpdateChecker::dateFromBuild(QVariant build)
{
    QString formatString = "yyyy-MM-dd";
    QString dateString = build.toMap()["created"].toString();
    dateString = dateString.remove(formatString.length(), dateString.length());

    return QDate::fromString(dateString, formatString);
}
```

[...]

Issue #2321, opened 2016-12-16
Add release channels to auto updater


Daenyth:
Currently if we host a build, the auto updater will download it. This means we can't effectively release beta or RC versions without having everyone start running it. We should add in release channels so that people who want to get beta versions will get them installed by the updater, and people who want stable can stay on full releases.

To start:
- Stable (our current releases and release process)
- Beta (manually released in the same way stable is, but not full releases)
- Unstable (nightly builds from master branch)

ZeldaZach:
I fully agree with this, and I'd love it if we can prioritize this for the release scheduled at the end of month. Beta/Unstable is a bit weird for us as we have 2 build channels; One that is manually updated by me with binaries, and one that is auto built every time a PR is merged. So we'd have to change the logic behind how stable vs beta is determined (right now, the way it parses bintray, i need to delete old builds for the updater to detect a change) then I'm all for it and we can easily sustain those 3 types of channels.

Daenyth:
I don't think this will get finished before then.

I don't think this is [API] since there's no server-side component here.

tooomm:
I like that! 👍 

Right now we have on bintray:
`cockatrice`: manually uploaded releases --> stable
`cockatrice-git`: automatic builds of master --> unstable

So we can go with these two at first maybe and extend the updater to offer downloads of unstable builds.

For the beta channel... would it be possible to separate the manual uploads in `cockatrice` like that:
"old/current releasename" --> stable
"new releasename + beta tag in name" --> beta
Once the beta gets stable, the beta tag is removed and the new release with its new name will get the beta label.
Then it might be possible to separate the two versions by two different queries.
If we separate both with an additional layer of folders (stable + beta) old clients with their update structure won't work any longer I guess?

**On the other hand, I'm not sure if we should offer unstable builds like that in-client at all. It adds a lot of traffic which might be problematic with bintray. And if people want to run experimental builds, they should deal with the effort of manually installing them. If they can't do that, they are most likely the first to complain and post issues about problems which are not too helpful just because they don't understand the meaning of unstable software in the first place.**

Daenyth:
It's reasonable to not auto-update unstable builds.

tooomm:
Oh, we are really speaking about real auto-updating?
Since we don't have that at all right now... all manual updates as of now. No automatic check and download.

Daenyth:
I thought the windows updater checked for new versions on startup?

tooomm:
[...]

PR #2362, opened 2017-01-15
Add different release channels for autoupdater

### LINKED ISSUE #2247: Wrong commit number in "new update available" dialog
<b>OS:</b>
Win 10 64bit

<b>Cockatrice version:</b>
4f5c2c3 (2016-10-29)

---

<br>

While the commit number of "All Hallow's Eve" is supposed to be 3498b16 ([Check release page](https://github.com/Cockatrice/Cockatrice/releases/tag/2016-10-30-Release)).
Cockatrice displays a different one:
updatemsg

ctrlaltca:
This is an initial implementation of #2321.
Currently two release channels are implemented: "stable releases" and "development snapshot", 	respectively pointing to our stable releases and git builds on bintray.

Users can choose their preferred release channel from the settings:
schermata 2017-01-15 alle 15 20 20

Their choice is reflected in the update dialog:
schermata 2017-01-15 alle 15 20 38

Also, remove the wrong commit hash from update dialog and show version name instead (fix #2247);
Still to be fixed / reworked, but maybe in another PR: don't use release date to determine if a newer version is available (aka avoid #2338)


ZeldaZach:
Just a thought to go with; `git rev-list --count HEAD` will give us a count of how many commits were made to the HEAD, which we can possibly use a confirmation. When a client starts up, it can run that commit and compare the counts. If they're not equal (aka HEAD > version string) then we show them update message.

```
$ git rev-list --count 784a75db
3539
$ git rev-list --count HEAD
3545
```

cockatrice/src/dlg_settings.cpp
```diff
@@ -269,8 +282,9 @@ void GeneralSettingsPage::retranslateUi()
     defaultUrlLabel.setText(tr("Primary download URL:"));
     fallbackUrlLabel.setText(tr("Fallback download URL:"));
     urlLinkLabel.setText(QString("<a href='%1'>%2</a>").arg(WIKI_CUSTOM_PIC_URL).arg(tr("How to set a custom picture url")));
-    clearDownloadedPicsButton.setText(tr("Reset/Clear Downloaded Pictures"));
-    updateNotificationCheckBox.setText(tr("Notify when new client features are available"));
+    clearDownloadedPicsButton.setText(tr("Reset/clear downloaded pictures"));
+    updateReleaseChannelLabel.setText(tr("Update channel"));
+    updateNotificationCheckBox.setText(tr("Notify when a new version is available"));
```
tooomm: Do we have any logic for automatic update checks yet?! First time I hear about that. :)

ctrlaltca: Not really; it still works that when you connect to a server and it tells you that you are missing some feature, we suppose that probably a new version supporting that feature is out.
I't just that imho the old text, even if formally correct, made no sense at all to an average user.

tooomm: I think the wording is still misleading...

Maybe add a "on connect" and rephrase a bit?
`"On connect, notify everytime a features supported by the server is missing in my client"`

tooomm:
@ctrlaltca 
>Also, remove the wrong commit hash from update dialog and **show version name instead** (fix #2247);

Can you show a picture, please?

I want to throw in https://github.com/Cockatrice/Cockatrice/pull/2359 again.
Now we should introduce the propper wording for the os on mac computers and scrap legacy terms.
We need to make sure that the app itself checks the new/right location though.

ctrlaltca:
> Just a thought to go with; git rev-list --count HEAD will give us a count of how many commits were made to the HEAD, which we can possibly use a confirmation. When a client starts up, it can run that commit and compare the counts. If they're not equal (aka HEAD > version string) then we show them update message.

That could work, unless if the user want to switch back from a development version to a stable version; just checking for HEAD != version string could be a simple workaround.
Anyway, the main problem is being able to push a value inside is the file that we load from bintray when checking for new versions (https://api.bintray.com/packages/cockatrice/Cockatrice/Cockatrice/files), so that Cockatrice is able to get this information. I'll dig into bintray's documentation, hopefully we can customize that file.
As an alternative, we could host a json file on cockatrice's repo.

> Can you show a picture, please?

Development snapshot update
schermata 2017-01-16 alle 21 17 58

Stable update:
schermata 2017-01-16 alle 21 18 22

> I want to throw in #2359 again.

Please, let me fix this once before we break it again ;)

cockatrice/src/dlg_settings.cpp
ctrlaltca: I've tried to change the sentence, but it's so long that it forces the settings window to be wider. So the candidates are:
```
# original wording
Notify when new client features are available
# short, misleading wording
Notify when a new version is available
# Precise but too long wording
On connect, notify everytime a feature supported by the server is missing in my client
# Other attempts
Notify if a feature supported by the server is missing in my client
Notify if the server reports that a feature is missing in my client
```
Others?

ZeldaZach:
I like the option with the dev/release channels; Looks good. If you're on a dev channel, will it allow you to go back to release and download the latest release with a notification or will you have to hit "download anyway?"

Having a JSON file tell which is the releases could be done, and we can host it on the /latest/release portion of the client as an uploaded file (github.com/cockatrice/cockatrice/releases/latest)

Daenyth:
I don't think we need to do a manual releases file, we should be able to
get the information from the releases page on github, or via the releases
api.

ctrlaltca:
> I don't think we need to do a manual releases file, we should be able to
> get the information from the releases page on github, or via the releases
> api.

How can we link the release information from github to the downloadable files from bintray?

Daenyth:
We could use a git tag with a specific structure. We talked about using
semver in another ticket, if we put that info into the git tags it would be
enough to determine new releases.

tooomm:
Regarding the release notes there is a `Release notes` tab on bintray, I'm pretty sure it used to display information from github releases the same as it's displaying information from the `README.md` file in the corresponding tab once. Not working anymore though?!

woogerboy21:
Compiled and tinkered a bit.  Any chance you could throw some info in the debug log when the update gets checked with some type of usable info?  Like maybe current version, versions found and results?  Something like that.

Daenyth:
Yes, that's exactly what I mean

tooomm:
>Probably someone copypasted them manually. Bintray can be configured to fetch (release notes) data from github, but imho we'd better redirect people on github, since all the informations are already here.

Link within the `release notes` tab of stable releases to https://github.com/Cockatrice/Cockatrice/releases/latest ?

ZeldaZach:
Any word on what we should do with this @Daenyth ?

ctrlaltca:
> Any word on what we should do with this @Daenyth ?

I'm still (slowly) working on implementing the changes

ctrlaltca:
Still work in progress, but updates for the development version are now implemented as discussed in https://github.com/Cockatrice/Cockatrice/pull/2362#issuecomment-273707379

ctrlaltca:
Both updates for the stable channel and the development channel have been implemented, and debug messages added as requested. This can be tested by reviewers.

On the stable channel it always download the last stable release available on bintray, since there's no way to actually link these files to the release on github, since both the release and the files are created manually; you may think of naming it the same and use this as a link, but this is fragile and a big con is the fact that on bintray the release can't be modified after the creation. We would risk to end up with the same problem of the last 2 releases where a not-matching date broke the update procedure.

woogerboy21:
Tinkered around some more.  Curious about something.  When running the most recent commit level with this PR rolled in and using the stable release option I have a popup that shows an update is available and lists the current stable release version that has been published.  Is this by design?  What I mean by this is since I am running a newer client than what is available in the stable version are we expecting the client to say "hey there is a more recent stable version than what your running (since your not running a stable release" or are we wanting it to say "hey this client is newer than the stable release available, do you want to install the stable release?"

Just want to make sure 1, the logic isnt actually failing to identify the correct versions and not informing people of updates when there is none and 2, expected behavior.

ctrlaltca:
This PR is not using the "release date" to comparing the current running version with the last version available for download, since we can't really control that information on bintray. This was the problem that caused the "update loop" in the two last releases.
The code from this PR is using the git commit hash embedded in cockatrice, comparing it to the commit hash obtained from github.
When running version from a release channel (stable or dev), if you decide to change to the other release channel you'll always find a different "last" version available, since it will surely have a different git commit hash.
There is no concept of a "newer" release, since dates are not involved in comparing the currently running version and the one available online.

cockatrice/src/dlg_settings.cpp
Daenyth: > Notify if a feature supported by the server is missing in my client

I really like this one but I don't have strong feelings - use your judgement

Daenyth:
@ctrlaltca What's the status of this at the moment? Are you waiting for code review, waiting for manual testing, or still making changes? I'd love to help push this through so if it's code review let me know and I'll get to it as I have time

ctrlaltca:
@Daenyth Waiting for review/test

cockatrice/src/dlg_update.cpp
```diff
@@ -108,40 +103,34 @@ void DlgUpdate::finishedUpdateCheck(bool needToUpdate, bool isCompatible, QVaria
...
         QMessageBox::information(this, tr("Cockatrice Update"), tr("Your version of Cockatrice is up to date."));
     }
 
+    if (isCompatible) {
+        //If there is an update, save its URL and work out its name
+        updateUrl = release->getDownloadUrl();
+        publishDate = release->getPublishDate().toString(Qt::DefaultLocaleLongDate);
+
+        QMessageBox::StandardButton reply;
+        reply = QMessageBox::question(this, "Update Available",
```
Daenyth: It seems like this is mixing UI and business logic together. An alternate approach might be to have the release updater not use any UI components but instead provide a function of `UpdateCheckResult checkForUpdates()` where `UpdateCheckResult` can be a tagged union with success+download_url, success-but-no-url-for-your-OS, you-are-up-to-date, or error-updating+the_error.

Does that make sense?

That would allow us to unit test in isolation from the UI.

Anyway I'm not going to block the PR on that change, but I'd like to hear your thoughts on it.

Daenyth: Also it seems like tagged unions are a goddamn PITA in cpp: http://en.cppreference.com/w/cpp/language/union

Thoughts on rust? I do have some docs on incorporating rust into a cpp app. I'd be perfectly willing to start extracting all the core bits to safer rust libraries and leave the Qt parts for just UI calling functions exposed from rust FFI.

cockatrice/src/releasechannel.cpp
```diff
@@ -0,0 +1,307 @@
...
+#define STABLEDOWNLOAD_URL "https://dl.bintray.com/cockatrice/Cockatrice/"
+#define STABLEMANUALDOWNLOAD_URL "https://bintray.com/cockatrice/Cockatrice/Cockatrice/_latestVersion#files"
+
+#define DEVRELEASE_URL "https://api.github.com/repos/Cockatrice/Cockatrice/commits/master"
+#define DEVFILES_URL "https://api.bintray.com/packages/cockatrice/Cockatrice/Cockatrice-git/files"
+#define DEVDOWNLOAD_URL "https://dl.bintray.com/cockatrice/Cockatrice/"
+#define DEVMANUALDOWNLOAD_URL "https://bintray.com/cockatrice/Cockatrice/Cockatrice-git/_latestVersion#files"
+#define GIT_SHORT_HASH_LEN 7
+
+ReleaseChannel::ReleaseChannel(int _index, QString _name)
```
Daenyth: Seems like the last 2 fields may want to be an UpdateResponse type object instead of the checker class itself.

cockatrice/src/releasechannel.cpp
```diff
@@ -0,0 +1,307 @@
...
+    QVariantMap resultMap = QtJson::Json::parse(tmp, ok).toMap();
+    if (!ok) {
+        qDebug() << "No reply received from the tag update server";
+        emit error(tr("No reply received from the tag update server."));
+        return;
+    }
+
+    if(!(resultMap.contains("object") && 
+        resultMap["object"].toMap().contains("sha"))) {
+        qDebug() << "Invalid received from the tag update server";
```
Daenyth: Is it easy to include the contents of `resultMap` here? Might help. Also probably should be `qWarning` at least

cockatrice/src/releasechannel.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+QString DevReleaseChannel::getManualDownloadUrl()
+{
+    return QString(DEVMANUALDOWNLOAD_URL);
+}
+
+void DevReleaseChannel::checkForUpdates()
+{
+    qDebug() << "Searching for updates on the dev channel: " << DEVRELEASE_URL;
+    response = netMan->get(QNetworkRequest(QString(DEVRELEASE_URL)));
```
Daenyth: This code is almost identical to the other class - it seems like url and log message are the only difference. Might be good to make it a method of the base interface they both extend and take the urls as parameters

cockatrice/src/releasechannel.cpp
```diff
@@ -0,0 +1,307 @@
...
+}
+
+void DevReleaseChannel::releaseListFinished()
+{
+    QNetworkReply *reply = static_cast<QNetworkReply *>(sender());
+    bool ok;
+    QString tmp = QString(reply->readAll());
+    reply->deleteLater();
+
+    // Errors are incapsulated in an object, check for them first
```
Daenyth: `encapsulated`

cockatrice/src/settingscache.cpp
```diff
@@ -162,7 +164,16 @@ SettingsCache::SettingsCache()
     if(!QFile(settingsPath+"global.ini").exists())
         translateLegacySettings();
 
+    // updates
+    dummy = QT_TRANSLATE_NOOP("i18n", "Stable releases");
+    dummy = QT_TRANSLATE_NOOP("i18n", "Development snapshots");
```
Daenyth: What's that do?

oracle/CMakeLists.txt
```diff
@@ -23,7 +23,9 @@ SET(oracle_SOURCES
     ../cockatrice/src/settings/layoutssettings.cpp
     ../cockatrice/src/thememanager.cpp 
     ../cockatrice/src/qt-json/json.cpp
- )
+    ../cockatrice/src/releasechannel.cpp
+     ${VERSION_STRING_CPP}
```
Daenyth: version cpp wasn't here already? How silly. We should probably make cockatrice's about menu reusable and call the same code from cockatrice and oracle. Not in this pr though

cockatrice/src/dlg_update.cpp
ctrlaltca: The business logic is mostly implemented in releasechannel.cpp, but these last checks ust be made in the UI since it's good to have the QMessageBox instanciated with a valid parent widget.
basically we are calling `settingsCache->getUpdateReleaseChannel()->checkForUpdates();` to start the check for updates and then waiting on the `DlgUpdate::finishedUpdateCheck()` slot for the result. It must be an async call anyway since the network is involved, so we cant just call a method and get the result directly.

cockatrice/src/releasechannel.cpp
ctrlaltca: I moved them inside the Release channels in the last update.

cockatrice/src/releasechannel.cpp
ctrlaltca: Done 👍

cockatrice/src/releasechannel.cpp
ctrlaltca: 👍  Done

cockatrice/src/releasechannel.cpp
ctrlaltca: Removed the comment altogether, it was a leftover from a copypaste

cockatrice/src/settingscache.cpp
ctrlaltca: It was used to force qt's l'update to find and extract these translatable strings.
I moved the strings inside the *ReleaseChannel classes and removed this trick.

cockatrice/src/dlg_update.cpp
```diff
@@ -47,11 +45,9 @@ DlgUpdate::DlgUpdate(QWidget *parent) : QDialog(parent) {
     //Check for SSL (this probably isn't necessary)
     if (!QSslSocket::supportsSsl()) {
         enableUpdateButton(false);
-        QMessageBox::critical(
-                this,
-                tr("Error"),
-                tr("Cockatrice was not built with SSL support, so cannot download updates! "
-                           "Please visit the download page and update manually."));
+        QMessageBox::critical(this, tr("Error"),
+            tr("Cockatrice was not built with SSL support, so cannot download updates! "
```
ZeldaZach: "Cockatrice was not build with SSL support, so you cannot download updates automatically! Please visit the download page to update manually."

cockatrice/src/releasechannel.cpp
```diff
@@ -0,0 +1,312 @@
...
+    QString myHash = QString(VERSION_COMMIT);
+    qDebug() << "Current hash=" << myHash << "update hash=" << shortHash;
+
+    bool needToUpdate = (QString::compare(shortHash, myHash, Qt::CaseInsensitive) != 0);
+    bool compatibleVersion = false;
+
+    foreach(QVariant file, resultList)
+    {
+        QVariantMap map = file.toMap();
+        // TODO: map github version to bintray version
```
ZeldaZach: any word on this?

cockatrice/src/releasechannel.h
```diff
@@ -0,0 +1,90 @@
...
+};
+
+class ReleaseChannel: public QObject {
+	Q_OBJECT
+public:
+    ReleaseChannel(int _index, QString _name);
+    ~ReleaseChannel();
+protected:
+	int index;
+	QString name;
```
ZeldaZach: tabs

cockatrice/src/releasechannel.cpp
ctrlaltca: Explained in the 2nd paragraph of https://github.com/Cockatrice/Cockatrice/pull/2362#issuecomment-281626414
Unfortunately i've not found a "perfect solution" yet, even though it's working anyway.

ZeldaZach: Sounds good for now, if we have to change uploads what will we do though? Have to publish a new revision?

cockatrice/src/releasechannel.h
ctrlaltca: Fixed, thank you

cockatrice/src/releasechannel.cpp
ctrlaltca: Publish a new revision to make cockatrice "find" the update.
Then, it will always download the "latest" stable release available on bintray.

ZeldaZach:
Builds seem to be failing @ctrlaltca

Daenyth:
Seems likely to be aws-related given that it's a connection error

Nightfirecat:
AWS has been facing a lot of issues today, most notably S3 being nearly fully out of operation for a few hours. That said, the error seems to be with a failing `Invoke-WebRequest` to download compiled protobuf 2.6.1 code: https://ci.appveyor.com/project/Daenyth/cockatrice-8k7y5/build/0.0.1-branch-master-build-565/job/2jfjveibhmpp2xof#L282, https://ci.appveyor.com/project/Daenyth/cockatrice-8k7y5/build/0.0.1-branch-master-build-565/job/df442gc7xk7rgavd#L282

ctrlaltca:
I've squashed down commits and pushed the PR again to force the builds to run again.
If everything's ok then i'll merge.

ZeldaZach:
I'll be looking into publishing a release on ~ March 14th (Pi Day :D)

ctrlaltca:
Looks like the travis-ci build ddin't trigger.. anyway it worked on the very same commit for my branch: https://travis-ci.org/ctrlaltca/Cockatrice/builds/206512853

PR #2363, opened 2017-01-15
Webclient: make account fields more user friendly

### LINKED ISSUE #2310: Extraneous / Debug fields on Web Client Account Tab
The following fields appear on the Account tab of web client (current version @ woogerworks, Windows 7 / Chrome):

Gender: GenderUnknown
id: -1
serverId: -1

Missing features
mod_log_lookup
user_ban_history

ctrlaltca:
Fix #2310

### DIFF
--- a/webclient/index.html
+++ b/webclient/index.html
@@ -301,7 +301,23 @@ Loading cockatrice web client...
                   break;
                 case 'gender':
                   $('#userinfo').append('Gender: ' + decodeGender(value) + '<br/>');
-                  break;                  
+                  break;
+                case 'name':
+                  $('#userinfo').append('Name: ' + value + '<br/>');
+                  break;
+                case 'realName':
+                  $('#userinfo').append('Real name: ' + value + '<br/>');
+                  break;
+                case 'country':
+                  $('#userinfo').append('Country: ' + value + '<br/>');
+                  break;
+                case 'privlevel':
+                  $('#userinfo').append('Privilege: ' + value + '<br/>');
+                  break;
+                case 'id':
+                case 'serverId':
+                  // don't output these fields
+                  break;
                 default:
                   $('#userinfo').append(key + ': ' + value + '<br/>');
                   break;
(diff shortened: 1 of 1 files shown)


woogerboy21:
Should the privlevel location mimic the non-web client syntax (userlevel | privlevel syntax) ?
I also didn't see anything addressing the missing features.  Was that an one of the issue in #2310 ?

ctrlaltca:
> Should the privlevel location mimic the non-web client syntax (userlevel | privlevel syntax) ?

I'm not a real fan of the `|` character, but more generally i have no preference on one or the other syntax;

> I also didn't see anything addressing the missing features. Was that an one of the issue in #2310 ?

How should i solve that? just hiding the missing features list?

woogerboy21:
> How should i solve that? just hiding the missing features list?

I don't care if they are showing myself.  But wasnt sure what others opinions are.  Maybe @tooomm might have a suggestion?  Maybe @tritoch since he opened the issue?

tritoch:
I don't have a strong opinion, hiding seems fine?

tooomm:
The gender concept is dropped, the field can be completely removed as we don't support it in the client anymore too --> https://github.com/Cockatrice/Cockatrice/pull/1955

The privlevel should be shown the same as in the client. 

Hmm, about missing features list... we don't have that information anywhere else in the webclient and the user don't get informed about it via popup as in the client.
Can we move that to the login tab maybe? Display current server related missing features there while connected?

tooomm:
I don't think so?

There is still an open question from ctrlatlca regarding the missing features list.
And the privlevel information is still displayed differently than in the normal client.

ctrlaltca:
Thank you @ZeldaZach for the commit.
About the missing feature list: i don't see a good reason to remove it, at least until the webclient reaches feature parity with cockatrice.
About the userlevel and privlevel label: the webclient is not meant not be a 1:1 clone of the cockatrice UI, and imho the current way the privlevel is shown in cockatrice is.. improveable.

I'll merge this as it is this right now, it's time to move on and start to work on fixing something more important than labels..

PR #2364, opened 2017-01-15
Fix use of uninitalized value

ctrlaltca:
Should only happen if the server accepts guest users.
Found using valgrind.


Daenyth:
👍

PR #2370, opened 2017-01-17
Add forgot password functionality

### LINKED ISSUE #1184: Add "Forgot password" functionality
We can register users but we can't reset passwords without outside intervention.
- [x] Cockatrice presents "forgot password" option with two modes. It accepts a username. If user has reset token, they can input it
- [x] Servatrice handles "forgot password" protocol action.
  - Reply to user with "Password reset request received. If the user has an activated email, they will be sent recovery instructions"
  - Check user exists, activated, has email.
  - Insert a row into password_reset table. It has a random token (like activation) and an expiry time set 24hrs after request. It records the requestor's ip address and time of request.
  - Send an email with "if you requested reset, paste this token to forgot password screen"
- [x] Servatrice handles forgot+token protocol action
  - Check username + token, generic error on failure
  - Check expiry time, if current time is after expiry time, reply with message saying it expired and to start over to get a new token
  - Set a randomized password and email it to the user.

woogerboy21:
Fix #1184

This PR adds "forgot user" functionality.  It currently works like this.  The connect dialog now has a button "forgot password".  If pressed, the user is presented with a dialog that asks for the server, port, username, and email.  The client then sends the following information to the server:

* filled in user name
* filled in email
* clients id

When the server receives the request it _can_ the checks the following:

* does users email match
* does the ip match from the users last login
* does the users clientid match last know clientid

If all criteria's pass the check, the server generates a new activation token and emails it to the user using the email address listed in the db associated with the user account.  It then adds and entry into a new audit table flagging the account and sends a request back to the client containing the following information:

* its defined server connectivity information
* username marked for password reset 

The user is then presented with another dialog that asks for the following information:

* generated token
* new password

The client then sends the inputed token and new password back to the server making the request.  When the server receives the new command it then checks for the following:

* is the user account marked for a forgot password request
* does the users token match what was inputted

If that criteria matches it then updates the users account password with the inputted new password, clears the flag in the audit table and sends the user a success.

If the entire process is successful, the user is now able to log in using the new password credentials.
If an account is flagged for forgot password reset and the user successfully logs in, the flag is cleared.
If the account is flagged for forgot password reset and another request is sent for forgot password functionality, the audit table is checked and a new entry is not put in place.

The server operator has the ability to enable/disable all 3 checks mentioned above as well as deactivate the account that has a forgot password request sent for it. Warnings are in place in both the servers ini as well as placed in the server log file if a server operator configures the server in an unsafe mode informing them that account deactivation with no other checks is a bad idea.

To prevent a user from trying to gain additional account information on a user account the server simply responds with RespOk if the forgot password request succeeds or RespInternalError if any of the various checks fail.


skwerlman:
This seems like it could be abused to lock someone out of their account, and requiring the IP and client ID to match feels like it would cause a lot of problems for people who travel a lot, have dynamic IPs, and people who haven't played in a long time. The client ID is especially troubling because even without switching computers, my client ID has changed several times in the last 12 months.

I think simply proving ownership of the email account (via the activation token) should be enough.

woogerboy21:
@skwerlman 
You can configure it that way.  Simply enable the functionality and disable the few checks you would prefer not to happen.  Though not comparing things would allow any user to type the username and email of another user and deactivate their account causing a password reset.  If thats how you want your server to operate you can configure it to do that.

skwerlman:
No, i meant that the only check should be the email, without the account ever getting deactivated. That way, if the reset was legit, the user can prove ownership of the account. If it was not legit, the account never gets locked, and the user gets notified of the attempt by the activation email.

woogerboy21:
Why on earth would this fail on just the trusty distro under travis?

woogerboy21:
> No, i meant that the only check should be the email, without the account ever getting deactivated. That way, if the reset was legit, the user can prove ownership of the account. If it was not legit, the account never gets locked, and the user gets notified of the attempt by the activation email.

So the account never gets deactivated, just flagged in some way.  Password never gets reset until they enter the new token and then the password is reset? Or the user is granted access and they can reset the password using the account tab? Something else...  What exactly logic flow did you have in mind?

ctrlaltca:
> Why on earth would this fail on just the trusty distro under travis (but not others)?

```
servatrice/src/serversocketinterface.cpp:959:24: error: unused parameter ‘rc’ [-Werror=unused-parameter]
 Response::ResponseCode AbstractServerSocketInterface::cmdForgotPassword(const Command_ForgotPassword &cmd, ResponseContainer &rc)
```

it fails only on that specific distro because it's configured to consider warnings as errors.

woogerboy21:
> it fails only on that specific distro because it's configured to consider warnings as errors.

Ah thank you.

woogerboy21:
So I've been brainstorming idea's on the auditing portion of the original #1184 issue.  Anyone have recommendations for the table?  My initial thoughts were to have a table that would encompass the use of more than just the password resets.  So maybe a table structure something like:

```
CREATE TABLE IF NOT EXISTS `cockatrice_audit` (
  `id` int(7) unsigned zerofill NOT NULL auto_increment,
  `user_name` varchar(35) NOT NULL,
  `type` varchar(255) NOT NULL,
  `email` varchar(255) NOT NULL,
  `incidentDate` datetime NOT NULL,
  `token` binary(16),
  `result` tinyint(1) NOT NULL,
  `details` varchar(255) NOT NULL,
  PRIMARY KEY  (`id`),
  KEY `token` (`user_name`),
  KEY `email` (`email`)
) ENGINE=INNODB  DEFAULT CHARSET=utf8;
```

Though is the token really needed?

woogerboy21:
@skwerlman 
Could you drop in a few concerns with the current flow of things to see if we can clear up any confusion and make sure we don't overlook anything.  Maybe give any idea's you may have outside what has already been discussed?

skwerlman:
> So the account never gets deactivated, just flagged in some way. Password never gets reset until they enter the new token and then the password is reset? Or the user is granted access and they can reset the password using the account tab? Something else... What exactly logic flow did you have in mind?

  1. Password reset is requested
  2. Account is flagged as wanting a reset
  3. User is sent email with activation token with an expiry of ~2h (should be configurable)
     1. If user successfully logs in with old password, server removes the reset flag, nullifying the activation token.
     2. If user tries to login with activation token as the password, they are presented with a dialog asking for a new password.
     3. If user does not attempt login before the token expires, the reset flag is removed.

> Im not sure I understand the concern as to me it sounds like your expecting the people trying to exploit the process to in some way not be able to identify whatever is sent in the clear text email that is being sent out yet imply that some how with the other checks they would be either able to mimic or not be in a condition with what information the server understands to verify the user is the user

My primary concern is that a successful reset request locks the user's account. What if a user loses access to their user account before being able to get the activation token? What if a user remembers their password shortly after requesting the reset?
The other concern is that some of the checks (client ID, IP) add restrictions on things that can and do change.

> One thing that could be added is a possible audit for the reset requests so that if some one is constantly trying to gain access to an account someone some were is notified in some way.... maybe.

I like this idea

> So maybe a table structure something like

I really like the idea of a generic auditing table; it'd make tracking things like failed logins, logins from new IPs, login from new client IDs, login to non-extant accounts easier.

woogerboy21:
(no idea how to quote and do the cool numbering thing you did so im just going to do it this way)

```
1. Password reset is requested
2. Account is flagged as wanting a reset
3. User is sent email with activation token with an expiry of ~2h (should be configurable)
     i. If user successfully logs in with old password, server removes the reset flag, nullifying the activation token.
    ii. If user tries to login with activation token as the password, they are presented with a dialog asking for a new password.
   iii. If user does not attempt login before the token expires, the reset flag is removed.
```

I'm not so sure I like the logic flow of this.  Primarily because what your asking is for a command to be created that allows for the password on an account to be reset without any type of account confirmation using account information with in the database.  Think of it this way, we have an app that sits on the open internet that listens to anyone that yells at it.  Currently we have layers of validation for interaction (login requirements etc) and your asking to for-go any level of those layers and simply ask the user to send a command and then a response to a return query for string information and if entered correctly blindly change the password on some ones account.  This screams "come hack me".  On top of that you also have to keep in mind that the password that is being sent to allow someone through the small hurdle put in place is being sent in clear text, to a most likely insecure account.  I can maybe buy into this type of password updating procedure if you had security questions as mentioned by @Daenyth in the original issue.

However, if everyone else believes this to be a "safe enough" solution then the code could be changed to accommodate this request.  I would have to think through how we could "flag" an account to allow the type of behavior you mention.  The users table is already growing rather large and I really dont like the idea of yet another table but that might be an option.  

Anyone else's opinion here would be helpful.

```
My primary concern is that a successful reset request locks the user's account. What if a user loses access to their user account before being able to get the activation token? 
```

If the user runs into any condition they would have to contact the server operator.  Which is what they already have to do for a password reset.  Personally I would prefer to lock the users account till it can be successfully verified that the true owner of the account is identified.  I'm just faulting on the side of safety.  If you would like I can adjust the code to accommodate the locking (or not) of an account on successful request.  I'm all about options 😄  But the notion of some one "spoofing" every check put in by the operator in order to even begin the account locking step isn't to much of an argument in my mind.  If you took the "anyone can spoof anything" approach we would never be able to come up with a solution. 

```
What if a user remembers their password shortly after requesting the reset?
```

In this case the user would simply use the new password information that is sent to them. One of those "oh shucks" moments.

```
The other concern is that some of the checks (client ID, IP) add restrictions on things that can and do change.
```

As mentioned, each check can be turned on or off by the server operator as they see fit.  I see no issue with including the functionality for data points available to the server operator and allowing them to decide how strict they would like things to be.

tooomm:
>The user's interface now has a menu option "forgot password" under the cockatrice menu.

Could there be a button in the `Connect` dialog instead? Sounds more natural over having a new menu entry.

Daenyth:
Related: https://www.owasp.org/index.php/Forgot_Password_Cheat_Sheet

I think we can skip the security question bit here. So that leaves us with a flow like:

1. User requests password change by username
2. Server emails the address on file with a time-limited token
3. User confirms reset by sending the token and a new desired password
4. Server confirms token is valid and assigns password to the new one
5. Server invalidates all current sessions for that username

All steps should also validate that the ip address and client id requesting the action are not banned.

The server should also validate that one ip address and client id are not requesting resets for many accounts in a short time frame.

All steps above should be logged into an audit table including IP address for every step, client id for every step, and timing

Daenyth:
BTW for changes like these we'd be well served by doing the design before writing code to minimize throwaway work

woogerboy21:
There isn't a lot of throw away to make it work the way you guys describe.  There will need to be an additional dialogue and command/response and adjustment to the audit table.

Daenyth:
summarizing from gitter

c->s reset $username plz
s->c sec question?
c->s sec answer
s->c ok, starting reset [email time limited token]
c->s confirm reset $username with $token
if $username has no question
skip those 2 steps

security question can be in the protocol and unimplemented on server side for now.

It should be a string question and answer in the protocol. We can use it for security questions as normal or later on with OTP two factor auth

woogerboy21:
> Could there be a button in the Connect dialog instead? Sounds more natural over having a new menu entry.

Done.  Though I will prob need help on figuring out how to size the button properly.  Now to add the server side logic for actually applying the new password.  The changes are *almost* done.

woogerboy21:
Alright, I think this PR is ready for re-review.  There is a large amount of updates / changes and prob needs some cleanup.  The only thing that I'm not so happy with right now is the size / layout of the buttons on the connect dialog.  I have no idea how to really adjust things so any help there would be thankful. 

Hopefully I have explained things out correctly in the first post.  It has been edited to reflect the way things function now.

@skwerlman 
@ctrlaltca 
@ZeldaZach 
@Daenyth 
@tooomm

tooomm:
>If all criteria's pass the check, the server generates a new activation token and emails it to the user using the email address listed in the db associated with the user account.

What happens if there is no mail server configured/available?
Will the tokens get stored so the admin can manually send them for example?

>Server invalidates all current sessions for that username

As Dae proposed, do we close all current sessions and boot all connected clients for that user name once the new password is set to force a new login with the updated credentials?
<br>
We have this? --> https://github.com/Cockatrice/Cockatrice/pull/2370#issuecomment-273960252

woogerboy21:
> What happens if there is no mail server configured/available?

Depending on the configuration of options it results could vary.  But in a nutshell the user would make a request, the request would generate the token and store it in the user's table and add an entry in the "to be emailed" table and then just sit there.  No email would be sent.  To the client they would receive a dialog saying "request made successfully, check your email" and it wouldnt come.  They would be presented with the "token/newpass" dialog and simply have to cancel out.  The generated token would expire in the number of minutes set by the operator in the .ini file (1 hour by default) and the process would have to start all over again in the event the user didn't get the token in time.  Once the user eventually successfully logs in the audit table is cleared of any "pending" forgot password requests.

> Will the tokens get stored so the admin can manually send them for example?

They are stored in the users table associated with the user account.  I don't believe there is any in client functionality that allows for anyone to look that information up at this time.

> As Dae proposed, do we close all current sessions and boot all connected clients for that user name once the new password is set to force a new login with the updated credentials?

If a user is logged in and some one were to make a "forgot password" request we do not boot there current logged in session.  But if someone completes the process of resetting the account password (or not really) and re-logins with the same username as the account already logged in.  The current session is forcefully closed and the new sessions is granted access.

tooomm:
>They would be presented with the "token/newpass" dialog and simply have to cancel out.

You can access to the same token dialog, if you close it and just start a new forgot password request?
If the time limit hasn't run out yet, the same token stays active and could be entered - let's say 30min - later?
<br>
In case one of you didn't run the pr-build, the button is placed like this right now:
fpw
Can you move it at the bottom of the `login` area?

woogerboy21:
> You can access to the same token dialog, if you close it and just start a new forgot password request?
If the time limit hasn't run out yet, the same token stays active and could be entered - let's say 30min - later?

Correct.  It is still valid till either the time limit the server operator has set (60min by default) OR the user logs in successfully using the account flagged. 

>  Can you move it to the bottom of the login area?

Ill see what I can do.  Not that great of a UI guy so it takes me a while to tinker with things like this.

woogerboy21:
@tooomm   Like this?

image

woogerboy21:
Alright, if anyone is interested in testing the current code functionality out.  I have a testing server up with this PR pulled in and things configured.  You can test using the following information:

server name:  testing.woogerworks.com
port: 4747

The server has the current users table imported into it from a backup done on 1/21 of the actual live woogerworks cockatrice server.  It also has registration and forgot password email functionality configured so you should get emails.

The settings are configured with the default settings for the forgot password code.

tooomm:
I just want to make sure that people, that aborted the process because the token mail is delayed (or send later manually), that they can reuse that token and are not stuck in the same problem for the second try. 👍 

I'm not committed to one or the other. :)
Another alternative placement for the button would be to have all 3 in a row (or maybe the new one a bit more separated to the left).

woogerboy21:
@tooomm 
How's this ->

image

tooomm:
Looks best so far!
Did you put a new group around them on purpose?
Only thing is that I tend to expect Ok/Cancle on the lower right. Might just be me.

woogerboy21:
> Did you put a new group around them on purpose?

Yes, the only way I could find to make the buttons sit in line was to format them this way.  The current ok/cancel buttons use a qdialogbuttonbox which has hard set buttons like ok/cancel/retry.  You can't add any custom buttons by the looks of things.  So I created stand alone ok/cancel/forgot password buttons then grouped them.  I may be able to hide the box around them or add a title to the box similar to the Login box.  Maybe a title like actions?

tooomm:
> I may be able to hide the box around them or add a title to the box similar to the Login box. Maybe a title like actions?

If you could hide it, give it a try. But no box name, please! :)

woogerboy21:
> If you could hide it, give it a try. But no box name, please! :)

Sadly my ui ninja skills are lacking.  I can't seem to figure out how to hide the border of the box.  I can remove the border of everything in that group box :(


image

tooomm:
Don't worry, the box is not too bad. 👍

tritoch:
Should the order of buttons be `OK | Forgot | Cancel` ?  I'm used to seeing 'cancel' as the rightmost option.

Daenyth:
The standards on button placement vary by OS.

woogerboy21:
I can place them in whatever order you like.  Just let me know what order and I can switch them around if need be.

tritoch:
Not sure which OS do which, but `Ok` should be changed to `Connect`

cockatrice/src/abstractclient.h
```diff
@@ -35,6 +35,8 @@ enum ClientStatus {
     StatusActivating,
     StatusLoggingIn,
     StatusLoggedIn,
+	StatusForgotPassword,
```
ZeldaZach: Tabs

cockatrice/src/dlg_connect.cpp
```diff
@@ -69,6 +71,19 @@ DlgConnect::DlgConnect(QWidget *parent)
 
     connect(savePasswordCheckBox, SIGNAL(stateChanged(int)), this, SLOT(passwordSaved(int)));
 
+	btnForgotPassword = new QPushButton(tr("Forgot password"));
+	btnForgotPassword->setFixedWidth(100);
```
ZeldaZach: Tabs + the FixedWidth looks bad on larger screens. The button will need to be larger I feel.

screenshot 2017-01-23 15 33 49

cockatrice/src/dlg_connect.cpp
```diff
@@ -79,6 +94,11 @@ DlgConnect::DlgConnect(QWidget *parent)
     connectionLayout->addWidget(portEdit, 4, 1);
     connectionLayout->addWidget(autoConnectCheckBox, 5, 1);
 
+	QGridLayout *buttons = new QGridLayout;
```
ZeldaZach: Tabs

cockatrice/src/dlg_connect.cpp
```diff
@@ -92,17 +112,16 @@ DlgConnect::DlgConnect(QWidget *parent)
     QGroupBox *loginGroupBox = new QGroupBox(tr("Login"));
     loginGroupBox->setLayout(loginLayout);
 
+	QGroupBox *btnGroupBox = new QGroupBox(tr(""));
```
ZeldaZach: Any reason you give it an empty name? I think you can just do `new QGroupBox()` and be done with it

cockatrice/src/dlg_connect.h
```diff
@@ -32,12 +36,14 @@ private slots:
     void passwordSaved(int state);
     void previousHostSelected(bool state);
     void newHostSelected(bool state);
+	void forgotYourPassword();
```
ZeldaZach: Tabs

cockatrice/src/dlg_forgotpassword.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <QHBoxLayout>
+#include <QDialogButtonBox>
+#include <QMessageBox>
+#include <QDebug>
+
+#include "dlg_forgotpassword.h"
+
+DlgForgotPassword::DlgForgotPassword(QWidget *parent)
+    : QDialog(parent)
+{
```
ZeldaZach: My question for this dialog is can we possibly re-use the server login portion with the dropdown menu so people don't always have to type things in constantly?

cockatrice/src/dlg_forgotpassword.cpp
```diff
@@ -0,0 +1,67 @@
...
+    portLabel = new QLabel(tr("Port:"));
+    portEdit = new QLineEdit();
+
+    playernameLabel = new QLabel(tr("Player Name:"));
+    playernameEdit = new QLineEdit();
+
+    emailLabel = new QLabel(tr("Email:"));
+    emailEdit = new QLineEdit();
+
+	QGridLayout *grid = new QGridLayout;
```
ZeldaZach: Tabs

cockatrice/src/dlg_forgotpassword.cpp
```diff
@@ -0,0 +1,67 @@
...
+	grid->addWidget(hostLabel, 0, 0);
+	grid->addWidget(hostEdit, 0, 1);
+	grid->addWidget(portLabel, 1, 0);
+	grid->addWidget(portEdit, 1, 1);
+	grid->addWidget(playernameLabel, 2, 0);
+	grid->addWidget(playernameEdit, 2, 1);
+	grid->addWidget(emailLabel, 3, 0);
+	grid->addWidget(emailEdit, 3, 1);
+
+    QDialogButtonBox *buttonBox = new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel);
```
ZeldaZach: When OK is pressed, the dialog will close no matter what and if you had something wrong in the details will have to start all over again... maybe persist the data or auto-reopen the dialog in case of failure?

cockatrice/src/dlg_forgotpasswordreset.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <QDebug>
+
+#include "dlg_forgotpasswordreset.h"
+
+DlgForgotPasswordReset::DlgForgotPasswordReset(QString host, int port, QString playerName, QWidget *parent)
+    : QDialog(parent)
+{
+    hostLabel = new QLabel(tr("Host:"));
+    hostEdit = new QLineEdit(host);
+	hostLabel->hide();
```
ZeldaZach: Why do we need a label if we're going to just hide it? Probably can be removed.

cockatrice/src/dlg_forgotpasswordreset.cpp
```diff
@@ -0,0 +1,101 @@
...
+    
+    newpasswordLabel = new QLabel(tr("New Password:"));
+    newpasswordEdit = new QLineEdit();
+	newpasswordEdit->setEchoMode(QLineEdit::Password);
+
+    confirmnewpasswordLabel = new QLabel(tr("Confirm New Password:"));
+    confirmnewpasswordEdit = new QLineEdit();
+	confirmnewpasswordEdit->setEchoMode(QLineEdit::Password);
+    
+	QGridLayout *grid = new QGridLayout;
```
ZeldaZach: Tabs

cockatrice/src/remoteclient.cpp
```diff
@@ -42,6 +43,8 @@ RemoteClient::RemoteClient(QObject *parent)
     connect(this, SIGNAL(sigConnectToServer(QString, unsigned int, QString, QString)), this, SLOT(doConnectToServer(QString, unsigned int, QString, QString)));
     connect(this, SIGNAL(sigDisconnectFromServer()), this, SLOT(doDisconnectFromServer()));
     connect(this, SIGNAL(sigRegisterToServer(QString, unsigned int, QString, QString, QString, int, QString, QString)), this, SLOT(doRegisterToServer(QString, unsigned int, QString, QString, QString, int, QString, QString)));
+	connect(this, SIGNAL(sigForgotPassword(QString, unsigned int, QString, QString)), this, SLOT(doForgotPassword(QString, unsigned int, QString, QString)));
```
ZeldaZach: Tabs

tooomm:
As suggested by tritoch, I vote for `Connect | Forgot | Cancel`

cockatrice/src/remoteclient.cpp
```diff
@@ -299,6 +343,30 @@ void RemoteClient::doRegisterToServer(const QString &hostname, unsigned int port
     setStatus(StatusRegistering);
 }
 
+void RemoteClient::doForgotPassword(const QString &hostname, unsigned int port, const QString &_userName, const QString &_email)
+{
+	doDisconnectFromServer();
+
+	userName = _userName;
+	email = _email;
```
ZeldaZach: Doesn't look like the email is used... I put a fake email in and it still emailed me at the correct email addr. Function should be it fails to send due to input_email != database_email. Whether we alert the user or not is up to you, but I disagree with having a user input an email addr if we're not using it for validation purposes

cockatrice/src/remoteclient.h
```diff
@@ -13,12 +13,16 @@ class RemoteClient : public AbstractClient {
     void serverTimeout();
     void loginError(Response::ResponseCode resp, QString reasonStr, quint32 endTime, QList<QString> missingFeatures);
     void registerError(Response::ResponseCode resp, QString reasonStr, quint32 endTime);
+	void processForgotPassword(Response::ResponseCode resp, QString requestingSrv, int requestingSrvPort, QString playerName);
```
ZeldaZach: Tabs

cockatrice/src/window_main.h
```diff
@@ -48,6 +49,8 @@ private slots:
     void serverTimeout();
     void loginError(Response::ResponseCode r, QString reasonStr, quint32 endTime, QList<QString> missingFeatures);
     void registerError(Response::ResponseCode r, QString reasonStr, quint32 endTime);
+	void processForgotPassword(Response::ResponseCode r, QString requestingSrv, int requestingSrvPort, QString playerName);
```
ZeldaZach: Tabs

cockatrice/src/window_main.cpp
```diff
@@ -474,6 +492,35 @@ void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quin
     actRegister();
 }
 
+void MainWindow::processForgotPassword(Response::ResponseCode r, QString requestingSrv, int requestingSrvPort, QString playerName)
+{
+	if (r == Response::RespOk) {
+			QMessageBox::information(this, tr("Success"), tr("Forgot password request successful, please check your email for your forgot passowrd activation token."));
```
ZeldaZach: "passowrd"

servatrice/servatrice.ini.example
```diff
@@ -144,6 +144,52 @@ disallowedregexp=""
 ; using the same email address. 0 = Unlimited number of accounts (default).
 ;maxaccountsperemail=0 
 
+[forgotpassword]
```
ZeldaZach: JW: Why do we have all the vars commented out in this section?

servatrice/src/servatrice.cpp
```diff
@@ -557,10 +585,38 @@ void Servatrice::statusUpdate()
...
+		query->bindValue(":email_type", "FORGOTPASS");
+		if (!servatriceDatabaseInterface->execSqlQuery(query)) {
+			qDebug() << "Failed to query users to send forgot password emails for: SQL Error";
+			return;
+		}
+
+		queryDelete = servatriceDatabaseInterface->prepareQuery("delete from {prefix}_activation_emails where name = :name and type = :email_type");
+
+		while (query->next()) {
+			const QString userName = query->value(0).toString();
```
ZeldaZach: Can we refer to the names of the values over an index?
`value("field_name")` instead of `value(0)`?

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -167,18 +167,18 @@ bool Servatrice_DatabaseInterface::usernameIsValid(const QString &user, QString
...
 
     QString passwordSha512 = PasswordHasher::computeHash(password, PasswordHasher::generateRandomSalt());
     token = active ? QString() : PasswordHasher::generateActivationToken();
 
-    QSqlQuery *query = prepareQuery("insert into {prefix}_users "
-            "(name, realname, gender, password_sha512, email, country, registrationDate, active, token) "
-            "values "
-            "(:userName, :realName, :gender, :password_sha512, :email, :country, UTC_TIMESTAMP(), :active, :token)");
+	QSqlQuery *query = prepareQuery("insert into {prefix}_users "
+		"(admin, name, realname, gender, password_sha512, email, country, registrationDate, active, token, avatar_bmp, clientid) "
```
ZeldaZach: 👍 on adding the default admin to insert

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -831,31 +832,35 @@ void Servatrice_DatabaseInterface::logMessage(const int senderId, const QString
     execSqlQuery(query);
 }
 
-bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword)
+bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword, bool force)
```
ZeldaZach: JW: Why did we change over all the true's to falses?

servatrice/src/servatrice_database_interface.h
```diff
@@ -71,13 +76,13 @@ public slots:
...
     bool registerUser(const QString &userName, const QString &realName, ServerInfo_User_Gender const &gender,
-        const QString &password, const QString &emailAddress, const QString &country, QString &token, bool active = false);
+        const QString &password, const QString &emailAddress, const QString &country, QString &token, const QString &clientid, bool active = false);
     bool activateUser(const QString &userName, const QString &token);
     void updateUsersClientID(const QString &userName, const QString &userClientID);
     void updateUsersLastLoginData(const QString &userName, const QString &clientVersion);
     void logMessage(const int senderId, const QString &senderName, const QString &senderIp, const QString &logMessage,
         LogMessage_TargetType targetType, const int targetId, const QString &targetName);
-    bool changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword);
+    bool changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword, bool force = false);
```
ZeldaZach: I took a look but didn't get what the "force" was for...

cockatrice/src/dlg_connect.cpp
woogerboy21: Yeah this was left over from a previous configuration of the dialog box and was missed when the buttons were redesigned.

cockatrice/src/dlg_connect.cpp
woogerboy21: Ill try it.  I dont recall if it threw errors or not when you left out the group box title name.

cockatrice/src/dlg_forgotpassword.cpp
woogerboy21: This was discussed in chat and it was mentioned to leave it as simple as possible at first and could be added to / improved later upon usage.

cockatrice/src/dlg_forgotpassword.cpp
woogerboy21: Isn't this the same idea as suggesting the dialog have similar functionality to the connect dialog?  My vote is to not store or retrieve any information for the forgot password dialog's.  IMO that verifys the user doesnt have a "whoops" moment.  They have to knowingly type in the required details.

cockatrice/src/dlg_forgotpasswordreset.cpp
woogerboy21: Nope.  The dialog's value is pulled to build the next command.  The dialog has line edits that are filled with data from the first response container of the server that was contacted to begin the process.  In a nutshell the entire process works like this:

client sends request to server for forgot password
server returns its host info and the value indicating its flagged the account for reset
client takes the response information and populates second dialog 
user fills in token (that was emailed) and new password sends it back to server
server resets user password

(keep in mind there are numerous checks that happen throughout the process)

cockatrice/src/remoteclient.cpp
woogerboy21: I would agree we should be checking email, but I was over-ruled.  (see the discussion above) . However , the email can be still be verified based on the configuration of the server operator (along with other checks).  In the default configuration of how the system is configured email verification was suggested to not happen.

servatrice/servatrice.ini.example
woogerboy21: Same reason the registration was set that way.  I just followed suit .

servatrice/src/servatrice_database_interface.cpp
woogerboy21: The original behavior of the function returned false if the results were successful.  Seemed back words to me.  The caller was sending a function call to "changeUserPassword" and it would return "false" if the check actually succeeded as the original variable name was "failedtochangepassword".  If some one else called the function for other uses with just "changeuserpassword" and expected a true for success or false for fail they wouldnt receive those results which was confusing.

servatrice/src/servatrice_database_interface.h
woogerboy21: The original use of the function was to change a users password ,  you would pass in the old password and new password.  The function would check if the old password matched and if so would reset the password, otherwise it wouldn't reset the user password.  To avoid building pretty much as similar function I simply appended the bool value (force) to this function and if the force option is passed the old password check is bypassed and the password is simply reset.  Since we dont know the users account old password during the password reset process we have no way to check against it :)

cockatrice/src/dlg_forgotpassword.cpp
ZeldaZach: Sure, but I still think this should be something we add in. Once we merge something like this, it might be a while until we touch it again

cockatrice/src/dlg_forgotpassword.cpp
ZeldaZach: Ok if you don't want to save the details, then at least re-open the dialog for forgotten password

cockatrice/src/dlg_forgotpasswordreset.cpp
ZeldaZach: I understand having `hostEdit` but I don't get why we need the `hostLabel`. That's what i'm asking

cockatrice/src/remoteclient.cpp
ZeldaZach: Well I disagree with not having the email verified. If a user has to type an email address it, it better be compared to the database email on file... otherwise I want to see this field removed from the asking question phase.

servatrice/src/servatrice_database_interface.cpp
ZeldaZach: Sounds good, thanks for fixing.

servatrice/src/servatrice_database_interface.h
ZeldaZach: Sounds good 👍

cockatrice/src/dlg_forgotpassword.cpp
woogerboy21: Im on @Daenyth side on this one (which generally doesn't happen).  Get the majority to agree with you and we can change it.  Im against the idea for now.  Im more in favor of making sure the user types the info in forcing a self aware verification type process.  But thats just my $0.02.

servatrice/servatrice.ini.example
woogerboy21: Should this be updated in any way or leave it commented?

servatrice/src/servatrice.cpp
woogerboy21: Maybe, once again though, followed suit with the registration block.

woogerboy21:
> Also, I don't like how we ask for an email addr, but don't use it for any kind of validation to see if the user entered the correct email. If they typed the wrong email, either alert them to this or not, but do NOT send an activation request.

As we discussed , the email check can be enabled/disabled by the server operator and it would be a very bad idea to report back to an un-verified users that the email they typed in for an account they are testing if the email on file is incorrect or not.  Think of someone trying to guess a users email address by typing in there username and sending the command over and over with different email address's....

servatrice/servatrice.ini.example
ZeldaZach: your call, not sure

servatrice/src/servatrice.h
```diff
@@ -191,8 +191,15 @@ public slots:
     bool getRequireEmailForRegistrationEnabled() const;
     bool getRequireEmailActivationEnabled() const;
     bool getEnableLogQuery() const;
+	bool getForgotPasswordEnabled() const;
```
ZeldaZach: tabs

ZeldaZach:
@woogerboy21 If that's the case, then report that 'If the email address you typed matches the email address on file, an activation email has been sent' and if not then don't send an email, but don't alert them to if it worked or not.

woogerboy21:
> If that's the case, then report that 'If the email address you typed matches the email address on file, an activation email has been sent' and if not then don't send an email, but don't alert them to if it worked or not.

Thinking through this, if I understand what your asking, you want me to update the dialog that notify's the user of a successful forgot password request.  Currently you see one of the two prompts:

Success -
```Forgot password request successful, please check your email for your forgot password activation token.```

Failure -
```Forgot password request failed, please contact the server operator to reset user account password.```

If your suggesting that we inform the user in some way that if the server is configured to check for matching passwords and inform them of something different based on server side configuration that wouldnt be a good idea either.  The wording is a client side dialog.  If we send some response back giving the reason why the user is getting a failure and try to disguise the results in a crafted dialog that doesnt really change the fact users information could be identified with custom clients.

woogerboy21:
@ZeldaZach 
I think I have a solution for your concern.  Plus it will help out in the future if questions are asked of the user.  I am going to add a chalange question step in the process.  So the user will submit there username for password reset.  the server will return a response requesting challenge at which point they will be prompted with whatever is enabled on the server for questions (such as users email) and then they submit that info.  then the server will mark the account ready if everything passes.  That should clear up any confusion and remove any un-needed inputs.

tooomm:
>So the user will submit there username for password reset. the server will return a response requesting challenge at which point they will be prompted with whatever is enabled on the server for questions (such as users email) and then they submit that info. 

It feels weird to just send the token without checking the mail or anything else. Even as option.

What about this?
check clientid -> if match with last known one --> info dialog + send token to known mail
no clientid match -> ask for mail --> on match: info dialog2 + send token to known+confirmed mail
(the second could be extended/exchanged to ask for other security question, 2FA... etc. later)

It also looks stupid to optionally force the same clientid... e.g. your at home all the time, pw is saved and now you want to show your friend the programm at his place; or you're on vacation and have to use a different pc. Of course you can't remember your saved pw because you never type it. Differet id --> no reset. :(

> does the ip match from the users last login

Is that check of any help? An ip is pretty random, changes frequently and if not it's most likely a connection like a university which has many different potential users.

woogerboy21:
@tooomm think of the process this way:

c->s (hey i want to reset the password for account user)
s-> check if user exists
s-> check if user banned
s-> check if users ip matches last login (optional check en/dis by operator)
if (en && y) - continue
s-> check if users clientid matches last login (optional check en/dis by operator)
if (en && y) - continue
s-> check if email validation is required (optional check en/dis by operator)
s-> build list of challenge questions
s->c send list of challenge required
c-> present user with list of challenge (if none continue)
c->s challenge question answer(s) (this is what the user said)
s-> is challenge answers correct
if (y)
s->c request successful, check email for token
c-> present user with dialog asking for token and new password
c-> send token + new pass
s-> is token correct
if (y)
s-> reset password
s->c password reset successfully
c-> present user with success

If at any time the "if's" are a no
s->c forgot password request failed
c-> present client with failure

ZeldaZach:
> @ZeldaZach
I think I have a solution for your concern. Plus it will help out in the future if questions are asked of the user. I am going to add a chalange question step in the process. So the user will submit there username for password reset. the server will return a response requesting challenge at which point they will be prompted with whatever is enabled on the server for questions (such as users email) and then they submit that info. then the server will mark the account ready if everything passes. That should clear up any confusion and remove any un-needed inputs.

So you're saying a server operator can request "challenge questions". And one of which will be "What is the email addr of this account?" And if it matches an email is sent and if it fails no email is sent. If this is the case, then I'm fine with this. I also think we should say "If the email addr you typed matches the email on file, we have sent an email to you to reset your password"

woogerboy21:
I am going to close this PR and resubmit in a simpler form.

PR #2376, opened 2017-01-19
Update registration user count check to include clientid

woogerboy21:
This PR updates the check that is done during registration for the number of accounts a user may have registered.  Originally the check only looked for accounts associated with the same email.  Now the check does email as well as clientid.

With the update in behavior I also updated the function names that get called for better description as well as the example.ini value used to set the limit.


tooomm:
Hmm. There is one limit for number of same email addresses AND clientid?

I think this are two separate things - while mail is more person related, a clientid is computer related. Imagine any shared computer within a family or public computers at uni/library/cybercafé...

woogerboy21:
Good point.  I'll switch it to a configurable option by the operator and mention some details in the ini .

woogerboy21:
Actually, I think I will close this PR and revisit it later once I get the email restriction in place and see how many people are really using multiple accounts on the same PC in a malicious fashion.

PR #2377, opened 2017-01-19
Added custom registration success/failure append messages

### LINKED ISSUE #1863: Add ability to append custom message to registration response
It would be nice to have a server ini option that would be appended to the dialogue that is presented when a user registered to a server .Might take a bit of thinking on how to do this.  Maybe in the response PB?

woogerboy21:
Fix #1863 

So this is my first attempt to implement the custom response message idea I was wanting to add a while back to allow server operators to add a more custom/descriptive reason as to why the server they operate may be denying registration or even an addition to the success message.

The original reasoning behind the idea was when registration emails do not immediately get sent to the end user they tend to nag operators.  And depending on the method an operator wishes to choose to activate an account an activation email may not be sent immediately upon registration request.  By allowing a more custom registration response message an operator may be able to explain a bit more in detail as to an acceptable waiting time for registration activation tokens.

The only thing I am questioning with the way its written in the PR is I am not certain how well translations would work mainly due to my lack of understanding how translations really work period.  I understand the that custom string read into the server will not translate but will the pre-appended "hard coded" words translate? Originally in the issue @Daenyth mentioned a custom response.  But that really isn't the goal of adding this functionality.  I didn't want to have to require the server operators to write updated code in order to send a custom message but more append a custom message to the already defined responses.

What are others thoughts in something like this?  To me the code seems a bit "dirty", but I couldn't think of a better way to implement it.  Maybe the solution should go more towards us creating custom responses and expecting operators to write code?

Thoughts?

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -187,7 +187,7 @@ void RemoteClient::registerResponse(const Response &response)
             doLogin();
             break;
         default:
-            emit registerError(response.response_code(), QString::fromStdString(resp.denied_reason_str()), resp.denied_end_time());
+            emit registerError(response.response_code(), QString::fromStdString(resp.denied_reason_str()), resp.denied_end_time(), QString::fromStdString(resp.custom_reason_str()));
             setStatus(StatusDisconnecting);
             doDisconnectFromServer();
             break;
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -433,32 +433,31 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
 
 void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quint32 endTime, QString customStr)
 {
-    qDebug() << "RECEIVED RESPONSE!";
-    QString titleStr = "Registration";
+    QString titleStr = "";
     QString explinationStr = "";
     switch (r) {
         case Response::RespRegistrationDisabled:
-            titleStr = titleStr + " denied";
-            explinationStr = "Registration is currently disabled on this server";
+            titleStr = tr("Registration denied");
+            explinationStr = tr("Registration is currently disabled on this server");
             break;
         case Response::RespUserAlreadyExists:
-            titleStr = titleStr + " denied";
-            explinationStr = "There is already an existing account with the same user name.";
+            titleStr = tr("Registration denied");
+            explinationStr = tr("There is already an existing account with the same user name.");
             break;
         case Response::RespEmailRequiredToRegister:
-            titleStr = titleStr + " denied";
-            explinationStr = "It's mandatory to specify a valid email address when registering.";
+            titleStr = tr("Registration denied");
+            explinationStr = tr("It's mandatory to specify a valid email address when registering.");
             break;
         case Response::RespTooManyRequests:
-            titleStr = titleStr + " denied";
-            explinationStr = "Too many registration attempts, please try again later or contact the server operator for further details.";
+            titleStr = tr("Registration denied");
+            explinationStr = tr("Too many registration attempts, please try again later or contact the server operator for further details.");
             break;
         case Response::RespPasswordTooShort:
-            titleStr = titleStr + " denied";
+            titleStr = tr("Registration denied");
             QMessageBox::critical(this, tr("Registration denied"), tr("Password too short."));
             break;
         case Response::RespUserIsBanned: {
-            titleStr = titleStr + " error";
+            titleStr = tr("Registration error");
             if (endTime)
                 explinationStr = tr("You are banned until %1.").arg(QDateTime::fromTime_t(endTime).toString());
             else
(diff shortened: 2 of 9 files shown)


cockatrice/src/window_main.cpp
```diff
@@ -431,46 +431,59 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
     return out;
 }
 
-void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quint32 endTime)
+void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quint32 endTime, QString customStr)
 {
+    qDebug() << "RECEIVED RESPONSE!";
```
Daenyth: 🔪

cockatrice/src/window_main.cpp
```diff
@@ -431,46 +431,59 @@ QString MainWindow::extractInvalidUsernameMessage(QString & in)
...
+    QString titleStr = "Registration";
+    QString explinationStr = "";
     switch (r) {
         case Response::RespRegistrationDisabled:
-            QMessageBox::critical(this, tr("Registration denied"), tr("Registration is currently disabled on this server"));
+            titleStr = titleStr + " denied";
```
Daenyth: This breaks the translation. The whole string to be translated needs to be in a `tr()` textually.

So what we want to do is actually keep the full `tr` calls in each switch case, and then *after* the switch case, append `\n\n` and the untranslated custom message. So like

```cpp
QString titleStr;
QString detailStr;
switch (r) {
  case Response::RespRegistrationDisabled:
    titleStr = tr("Registration denied");
    detailStr = tr("Registration is currently disabled on this server");
    break;
  case ...
  case default: ...
}
if (!customStr.isEmpty)
  detailStr += "\n\n" + tr("Note from server: ") + customStr;
QMessageBox::critical(this, titleStr, detailStr);
```

servatrice/src/servatrice.h
```diff
@@ -179,6 +179,8 @@ public slots:
     QString getRequiredFeatures() const;
     QString getAuthenticationMethodString() const;
     QString getDBTypeString() const;
+	QString getCustomRegErrorMessage() const;
```
Daenyth: Tab here should be space

servatrice/src/serversocketinterface.cpp
```diff
@@ -877,63 +877,77 @@ Response::ResponseCode AbstractServerSocketInterface::cmdRegisterAccount(const C
     QString clientId = QString::fromStdString(cmd.clientid());
     qDebug() << "Got register command: " << userName;
 
+    Response_Register *re = new Response_Register;
+    re->set_custom_reason_str(servatrice->getCustomRegErrorMessage().toStdString());
+
     bool registrationEnabled = settingsCache->value("registration/enabled", false).toBool();
-    if (!registrationEnabled)
+    if (!registrationEnabled) {
+        rc.setResponseExtension(re);
```
Daenyth: Is this needed? Did this branch not work right before?

servatrice/src/serversocketinterface.cpp
```diff
@@ -877,63 +877,77 @@ Response::ResponseCode AbstractServerSocketInterface::cmdRegisterAccount(const C
...
     if (!sqlInterface->usernameIsValid(userName, errorString))
     {
-        Response_Register *re = new Response_Register;
         re->set_denied_reason_str(errorString.toStdString());
         rc.setResponseExtension(re);
         return Response::RespUsernameInvalid;
     }
 
-    if (userName.toLower().simplified() == "servatrice")
+    if (userName.toLower().simplified() == "servatrice") {
```
Daenyth: lulz

servatrice/src/serversocketinterface.cpp
```diff
@@ -946,14 +960,21 @@ Response::ResponseCode AbstractServerSocketInterface::cmdRegisterAccount(const C
...
             QSqlQuery *query = sqlInterface->prepareQuery("insert into {prefix}_activation_emails (name) values(:name)");
             query->bindValue(":name", userName);
-            if (!sqlInterface->execSqlQuery(query))
+            if (!sqlInterface->execSqlQuery(query)) {
+                rc.setResponseExtension(re);
                 return Response::RespRegistrationFailed;
+            }
 
+            re->set_custom_reason_str(servatrice->getCustomRegSuccessMessage().toStdString());
+            rc.setResponseExtension(re);
```
Daenyth: This call gets duplicated all over. Can we move it out so it's only in the source once? Maybe right after allocating the response object?

cockatrice/src/remoteclient.cpp
```diff
@@ -187,7 +187,7 @@ void RemoteClient::registerResponse(const Response &response)
             doLogin();
             break;
         default:
-            emit registerError(response.response_code(), QString::fromStdString(resp.denied_reason_str()), resp.denied_end_time());
+            emit registerError(response.response_code(), QString::fromStdString(resp.denied_reason_str()), resp.denied_end_time(), QString::fromStdString(resp.custom_reason_str()));
```
Daenyth: `custom_reason_str` was already in the proto? Nice

servatrice/src/serversocketinterface.cpp
woogerboy21: This is the dirtiness factor I was referring to.  Since the response extension is created prior to the switch statement and the custom_reason_str gets set based on the results of the various tree outcomes you have to set the response code prior to returning the response.  I could possibly place a variable in the function that defines the type of response needs to be returned and set the response and send it below the switch statement.  Either way doesn't seem ideal to me.  I went this route as its seemed easier to understand through the various if tree's on were the code logic should end with a return for what the results being returned would be. (if that makes any since).

servatrice/src/serversocketinterface.cpp
woogerboy21: Whats so funny?

servatrice/src/serversocketinterface.cpp
woogerboy21: Branches like this simply returned a response value but didnt define any extended pb response items.  If the server operator has a custom string in the ini we now have to set the response extension results before returning the response and thus adding an additional line to the if statement requiring brackets around the beg/end of the logic tree.

cockatrice/src/remoteclient.cpp
woogerboy21: No it wasnt?  We had a reason string which was used for banned users.

cockatrice/src/window_main.cpp
ZeldaZach: Dae is correct here. The way translations are done this cannot work

common/pb/response_register.proto
```diff
@@ -7,4 +7,5 @@ message Response_Register {
     }
     optional string denied_reason_str = 1;
     optional uint64 denied_end_time = 2;
+	optional string custom_reason_str = 3;
```
Daenyth: Tab should be space

servatrice/src/serversocketinterface.cpp
woogerboy21: Also the custom_reason_str gets set either as a success or failure returning the appropriate string in the ini.  The server operator can define a "success" string and a "failure" string to return.

servatrice/src/serversocketinterface.cpp
woogerboy21: We actually use this username to send server messages and is why the name is hard denied for registration.

woogerboy21:
@Daenyth or @ZeldaZach 
I believe I have resolved the translation fixes for the block in the window_main discussed.  Take a look and let me know if it still isn't correct.

woogerboy21:
I am going to close this and use the approach mentioned in #2398 .  The more I think about this way it seems very limited in scope.  I would rather have a custom event that the server can send post any response notifying the user of additional information that the operator may want to inform the user of.

PR #2393, opened 2017-01-28
Simpler forgot password functionality

woogerboy21:
Here is a re-submission of the forgot password functionality.  This version is much simpler.  Here is how it works:

client->server (send request)
server-> verify functionality enabled
server-> validate user exists
server-> verify user isnt banned
server-> check if challenge questions enabled

if ( checks pass AND challenge enabled )
   server->client (request client to answer question)
   client->server (send challenge answer)
   server-> verify challenge questions
   if (challenge questions pass)
      server-> mark account for reset request
      server-> email client token
      server->client (request token + new pass)
      client->server (send token+pass)
      server-> check token
      if (token correct)
         server-> update pass
         server->client (notify of success)

if ( checks pass AND challenge disabled )
   server-> mark account for reset request
   server-> email client token
   server->client (request token + new pass)
   client->server (send token+pass)
   server-> check token
   if (token correct)
      server-> update pass
      server->client (notify of success)

Currently, if at any time the checks fail, the user is informed of a failure.  Now I know at one time or another there was some discussion about possibly changing the behavior on how things are presented to the client.  If things need adjusted they should be pretty easily changed.  If I remember correctly the only behavior change that may need to be made is if the user types the email in wrong that instead of telling them "fail" do a "succeed" so if that's needed let me know (its just a change on which signal to emit during the response processing).

Some other things to note,  unlike the last revision the client now handles storing the inputted information on the dialog and stores it throughout the process in new settings that are cleared upon success.  With this configuration dialog(s) can be closed and reopened later and the process continue.

Check it out and let me know what things need "rejiggered".

### DIFF
--- a/cockatrice/src/dlg_forgotpasswordchallenge.cpp
+++ b/cockatrice/src/dlg_forgotpasswordchallenge.cpp
@@ -6,10 +6,10 @@
 #include <QMessageBox>
 #include <QDebug>
 
-#include "dlg_forgotpasswordreset.h"
+#include "dlg_forgotpasswordchallenge.h"
 #include "settingscache.h"
 
-DlgForgotPasswordReset::DlgForgotPasswordReset(QWidget *parent)
+DlgForgotPasswordChallenge::DlgForgotPasswordChallenge(QWidget *parent)
     : QDialog(parent)
 {
 
@@ -26,7 +26,7 @@ DlgForgotPasswordReset::DlgForgotPasswordReset(QWidget *parent)
 
 	if (settingsCache->servers().getFPHostname().isEmpty() && settingsCache->servers().getFPPort().isEmpty() && settingsCache->servers().getFPPlayerName().isEmpty())
 	{
-		QMessageBox::warning(this, tr("Forgot Password Reset Warning"), tr("Opps, looks like something has gone wrong.  Please re-start the forgot password process by using the forgot password button on the connection screen."));
+		QMessageBox::warning(this, tr("Forgot Password Challenge Warning"), tr("Opps, looks like something has gone wrong.  Please re-start the forgot password process by using the forgot password button on the connection screen."));
 		actCancel();
 	}
 
@@ -42,19 +42,9 @@ DlgForgotPasswordReset::DlgForgotPasswordReset(QWidget *parent)
     playernameEdit = new QLineEdit(lastfpplayername);
     playernameLabel->setBuddy(playernameEdit);
 
-	tokenLabel = new QLabel(tr("Token:"));
-	tokenEdit = new QLineEdit();
-	tokenLabel->setBuddy(tokenLabel);
-
-	newpasswordLabel = new QLabel(tr("New Password:"));
-	newpasswordEdit = new QLineEdit();
-	newpasswordLabel->setBuddy(newpasswordEdit);
-	newpasswordEdit->setEchoMode(QLineEdit::Password);
-
-	newpasswordverifyLabel = new QLabel(tr("New Password:"));
-	newpasswordverifyEdit = new QLineEdit();
-	newpasswordverifyLabel->setBuddy(newpasswordEdit);
-	newpasswordverifyEdit->setEchoMode(QLineEdit::Password);
+	emailLabel = new QLabel(tr("Email:"));
+	emailEdit = new QLineEdit();
+	emailLabel->setBuddy(emailLabel);
 
 	if (!settingsCache->servers().getFPHostname().isEmpty() && !settingsCache->servers().getFPPort().isEmpty() && !settingsCache->servers().getFPPlayerName().isEmpty()) {
 		hostLabel->hide();
@@ -72,12 +62,8 @@ DlgForgotPasswordReset::DlgForgotPasswordReset(QWidget *parent)
     grid->addWidget(portEdit, 1, 1);
     grid->addWidget(playernameLabel, 2, 0);
     grid->addWidget(playernameEdit, 2, 1);
-	grid->addWidget(tokenLabel, 3, 0);
-	grid->addWidget(tokenEdit, 3, 1);
-	grid->addWidget(newpasswordLabel, 4, 0);
-	grid->addWidget(newpasswordEdit, 4, 1);
-	grid->addWidget(newpasswordverifyLabel, 5, 0);
-	grid->addWidget(newpasswordverifyEdit, 5, 1);
+	grid->addWidget(emailLabel, 3, 0);
+	grid->addWidget(emailEdit, 3, 1);
 
     QDialogButtonBox *buttonBox = new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel);
     connect(buttonBox, SIGNAL(accepted()), this, SLOT(actOk()));
@@ -88,34 +74,16 @@ DlgForgotPasswordReset::DlgForgotPasswordReset(QWidget *parent)
     mainLayout->addWidget(buttonBox);
     setLayout(mainLayout);
 
-    setWindowTitle(tr("Forgot Password Reset"));
+    setWindowTitle(tr("Forgot Password Challenge"));
     setFixedHeight(sizeHint().height());
     setMinimumWidth(300);
 }
 
-void DlgForgotPasswordReset::actOk()
+void DlgForgotPasswordChallenge::actOk()
 {
-    if(playernameEdit->text().isEmpty())
-    {
-        QMessageBox::critical(this, tr("Forgot Password Reset Warning"), tr("The player name can't be empty."));
-        return;
-    }
-
-	if (tokenEdit->text().isEmpty())
+	if (emailEdit->text().isEmpty())
 	{
-		QMessageBox::critical(this, tr("Forgot Password Reset Warning"), tr("The token can't be empty."));
-		return;
-	}
-
-	if (newpasswordEdit->text().isEmpty())
-	{
-		QMessageBox::critical(this, tr("Forgot Password Reset Warning"), tr("The new password can't be empty."));
-		return;
-	}
-
-	if (newpasswordEdit->text() != newpasswordverifyEdit->text())
-	{
-		QMessageBox::critical(this, tr("Forgot Password Reset Warning"), tr("The passwords do not match."));
+		QMessageBox::critical(this, tr("Forgot Password Challenge Warning"), tr("The email address can't be empty."));
 		return;
 	}
 
@@ -126,7 +94,7 @@ void DlgForgotPasswordReset::actOk()
     accept();
 }
 
-void DlgForgotPasswordReset::actCancel()
+void DlgForgotPasswordChallenge::actCancel()
 {
     reject();
 }
--- a/cockatrice/src/dlg_forgotpasswordchallenge.cpp
+++ b/cockatrice/src/dlg_forgotpasswordchallenge.cpp
@@ -26,7 +26,7 @@ DlgForgotPasswordChallenge::DlgForgotPasswordChallenge(QWidget *parent)
 
 	if (settingsCache->servers().getFPHostname().isEmpty() && settingsCache->servers().getFPPort().isEmpty() && settingsCache->servers().getFPPlayerName().isEmpty())
 	{
-		QMessageBox::warning(this, tr("Forgot Password Challenge Warning"), tr("Opps, looks like something has gone wrong.  Please re-start the forgot password process by using the forgot password button on the connection screen."));
+		QMessageBox::warning(this, tr("Forgot Password Challenge Warning"), tr("Oops, looks like something has gone wrong.  Please restart the forgot password process by using the forgot password button on the connection screen."));
 		actCancel();
 	}
 
--- a/cockatrice/src/dlg_forgotpasswordchallenge.cpp
+++ b/cockatrice/src/dlg_forgotpasswordchallenge.cpp
@@ -13,22 +13,22 @@ DlgForgotPasswordChallenge::DlgForgotPasswordChallenge(QWidget *parent)
     : QDialog(parent)
 {
 
-	QString lastfphost; QString lastfpport; QString lastfpplayername;
-	lastfphost = settingsCache->servers().getHostname("cockatrice.woogerworks.com");
-	lastfpport = settingsCache->servers().getPort("4747");
-	lastfpplayername = settingsCache->servers().getPlayerName("Player");
+    QString lastfphost; QString lastfpport; QString lastfpplayername;
+    lastfphost = settingsCache->servers().getHostname("cockatrice.woogerworks.com");
+    lastfpport = settingsCache->servers().getPort("4747");
+    lastfpplayername = settingsCache->servers().getPlayerName("Player");
 
-	if (!settingsCache->servers().getFPHostname().isEmpty() && !settingsCache->servers().getFPPort().isEmpty() && !settingsCache->servers().getFPPlayerName().isEmpty()) {
-		lastfphost = settingsCache->servers().getFPHostname();
-		lastfpport = settingsCache->servers().getFPPort();
-		lastfpplayername = settingsCache->servers().getFPPlayerName();
-	}
+    if (!settingsCache->servers().getFPHostname().isEmpty() && !settingsCache->servers().getFPPort().isEmpty() && !settingsCache->servers().getFPPlayerName().isEmpty()) {
+        lastfphost = settingsCache->servers().getFPHostname();
+        lastfpport = settingsCache->servers().getFPPort();
+        lastfpplayername = settingsCache->servers().getFPPlayerName();
+    }
 
-	if (settingsCache->servers().getFPHostname().isEmpty() && settingsCache->servers().getFPPort().isEmpty() && settingsCache->servers().getFPPlayerName().isEmpty())
-	{
-		QMessageBox::warning(this, tr("Forgot Password Challenge Warning"), tr("Oops, looks like something has gone wrong.  Please restart the forgot password process by using the forgot password button on the connection screen."));
-		actCancel();
-	}
+    if (settingsCache->servers().getFPHostname().isEmpty() && settingsCache->servers().getFPPort().isEmpty() && settingsCache->servers().getFPPlayerName().isEmpty())
+    {
+        QMessageBox::warning(this, tr("Forgot Password Challenge Warning"), tr("Oops, looks like something has gone wrong.  Please restart the forgot password process by using the forgot password button on the connection screen."));
+        actCancel();
+    }
 
     hostLabel = new QLabel(tr("&Host:"));
     hostEdit = new QLineEdit(lastfphost);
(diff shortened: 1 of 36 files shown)


woogerboy21:
Also, this should be progress (if not a fix) for #1184

woogerboy21:
If users would like to test this PR I have a server up and running with the codebase on it.
Host:  testing.woogerworks.com
Port: 4747

tooomm:
For comparison, [the former C-S communication concept in the old pr](https://github.com/Cockatrice/Cockatrice/pull/2370#issuecomment-274895456)!
Changes mostly dropping the clientID and ip check and restructuring the challenge questions?

We don't always want to validate the reset request in one way or the other?
Just asking for the username (public information) seems odd and could be abused, allows for user harassment (token spamming) or server stress. Are you not concerned here?

woogerboy21:
> We don't always want to validate the reset request in one way or the other?
Just asking for the username (public information) seems odd and could be abused, allows for user harassment (token spamming) or server stress. Are you not concerned here?

The functionality direction given was to accept a request from a user for forgot password, allow for challenge question functionality (not requiring it) and then validate users forgot password request with a generated token sent via email.  This version accomplishes those tasks.  The operator has the ability to turn on / off the challenge question (in this case prompts the user for the email on file but can be expanded in the future) and the server flags the user account for password reset, generating a token that is only valid for the period of time indicated by the operator and emails the users email on file the token.  During each step of validation (either challenge or token verification) the check for if the user account is flagged for password reset is verified.  If the account is not flagged it doesn't perform any actions and returns a failure to the user making the request.

I would agree there should be more checks in place, but from my understanding thats the "expand on in the future" part of the concept.  What to me isn't clear is how / when the user should be told "there is an error" or "move on to the next step".

As far as server stress, right now there is nothing to stop anyone from performing the actions you say for any of the command request.  Some one could simply spam the server over and over with buddy list queries or registration requests and the server does nothing about it.

woogerboy21:
FYI - I also intend on making an audit table PR in a different change since it was of interest through the last discussion but wanted to keep it separate at this time for clarity.

woogerboy21:
I committed some of the other PR's that were approved and then re-based this one.  It took one minor conflict resolution so I recompiled the test box if anyone is wanting to retry things on this.  Otherwise if no comments are made over the next couple days I'll go ahead and merge (unless some one else wishes to merge it).

ZeldaZach:
Just remember to tabs->spaces your work.

cockatrice/src/dlg_forgotpasswordchallenge.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include "dlg_forgotpasswordchallenge.h"
+#include "settingscache.h"
+
+DlgForgotPasswordChallenge::DlgForgotPasswordChallenge(QWidget *parent)
+    : QDialog(parent)
+{
+
+	QString lastfphost; QString lastfpport; QString lastfpplayername;
+	lastfphost = settingsCache->servers().getHostname("cockatrice.woogerworks.com");
```
Daenyth: Why is this server name hardcoded? This should come from the server info list

cockatrice/src/dlg_forgotpasswordchallenge.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+	if (!settingsCache->servers().getFPHostname().isEmpty() && !settingsCache->servers().getFPPort().isEmpty() && !settingsCache->servers().getFPPlayerName().isEmpty()) {
+		lastfphost = settingsCache->servers().getFPHostname();
+		lastfpport = settingsCache->servers().getFPPort();
+		lastfpplayername = settingsCache->servers().getFPPlayerName();
+	}
+
+	if (settingsCache->servers().getFPHostname().isEmpty() && settingsCache->servers().getFPPort().isEmpty() && settingsCache->servers().getFPPlayerName().isEmpty())
+	{
+		QMessageBox::warning(this, tr("Forgot Password Challenge Warning"), tr("Opps, looks like something has gone wrong.  Please re-start the forgot password process by using the forgot password button on the connection screen."));
```
Daenyth: typo: `Opps` should be `Oops`. `re-start` should be `restart`

cockatrice/src/dlg_forgotpasswordrequest.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "dlg_forgotpasswordrequest.h"
+#include "settingscache.h"
+
+DlgForgotPasswordRequest::DlgForgotPasswordRequest(QWidget *parent)
+    : QDialog(parent)
+{
+
+	QString lastfphost; QString lastfpport; QString lastfpplayername;
+	lastfphost = settingsCache->servers().getHostname("cockatrice.woogerworks.com");
```
Daenyth: Hardcoded server here again.

cockatrice/src/dlg_forgotpasswordreset.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include "dlg_forgotpasswordreset.h"
+#include "settingscache.h"
+
+DlgForgotPasswordReset::DlgForgotPasswordReset(QWidget *parent)
+    : QDialog(parent)
+{
+
+	QString lastfphost; QString lastfpport; QString lastfpplayername;
+	lastfphost = settingsCache->servers().getHostname("cockatrice.woogerworks.com");
```
Daenyth: Also here

common/pb/session_commands.proto
```diff
@@ -160,3 +163,30 @@ message Command_AccountPassword {
...
+    }
+    optional string user_name = 1;
+    optional string clientid = 2;
+}
+
+message Command_ForgotPasswordReset {
+    extend SessionCommand {
+        optional Command_ForgotPasswordReset ext = 1022;
+    }
+    optional string user_name = 1;
```
Daenyth: Why are these all optional? At least user_name makes sense to be required, probably token also and new_password

servatrice/migrations/servatrice_0020_to_0021.sql
```diff
@@ -0,0 +1,12 @@
+
+
+CREATE TABLE IF NOT EXISTS `cockatrice_forgot_password` (
+  `id` int(7) unsigned zerofill NOT NULL auto_increment,
+  `name` varchar(35) NOT NULL,
+  `requestDate` datetime NOT NULL default '0000-00-00 00:00:00',
```
Daenyth: Don't put a default value in here, that doesn't make sense. It should just be NOT NULL.

Can you also add columns for requestor IP address and client id?

servatrice/servatrice.ini.example
```diff
@@ -142,7 +142,37 @@ disallowedregexp=""
...
+; Default: 60
+; tokenlife=60
+
+; Servatrice can challenge users that are making forgot password requests to answer
+; questions in regards to their account to help validate they are the true owner of the account.
+; Should this feature be enabled? Default: false
+; enablechallenge=false
+
+; Sender email address: the "from" email address for forgot password emails
+email=root@localhost
```
Daenyth: Should this be a separate setting? Why not reuse the sender "from" address used in registration? I think the server should just have a single "from" address

servatrice/src/servatrice.cpp
```diff
@@ -543,27 +549,51 @@ void Servatrice::statusUpdate()
     servatriceDatabaseInterface->execSqlQuery(query);
 
     // send activation emails
```
Daenyth: Comment is outdated now I think. But this kind of comment hints that the code following may want to be extracted to a method with a similar name as the comment text

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -175,10 +175,10 @@ bool Servatrice_DatabaseInterface::registerUser(const QString &userName, const Q
     QString passwordSha512 = PasswordHasher::computeHash(password, PasswordHasher::generateRandomSalt());
     token = active ? QString() : PasswordHasher::generateActivationToken();
 
-    QSqlQuery *query = prepareQuery("insert into {prefix}_users "
-            "(name, realname, gender, password_sha512, email, country, registrationDate, active, token) "
-            "values "
-            "(:userName, :realName, :gender, :password_sha512, :email, :country, UTC_TIMESTAMP(), :active, :token)");
+	QSqlQuery *query = prepareQuery("insert into {prefix}_users "
+		"(name, realname, gender, password_sha512, email, country, registrationDate, active, token, admin, avatar_bmp, clientid, privlevel, privlevelStartDate, privlevelEndDate) "
```
Daenyth: Does this fix mean we can insert to users in strict mode?

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -831,41 +831,43 @@ void Servatrice_DatabaseInterface::logMessage(const int senderId, const QString
...
-        return true;
+        return false;
 
-    QSqlQuery *passwordQuery = prepareQuery("select password_sha512 from {prefix}_users where name = :name");
-    passwordQuery->bindValue(":name", user);
-    if (!execSqlQuery(passwordQuery)) {
-        qDebug("Change password denied: SQL error");
-        return true;
-    }
+	QSqlQuery *passwordQuery = prepareQuery("select password_sha512 from {prefix}_users where name = :name");
```
Daenyth: Can you reuse the "check password" code path from the login code rather than writing the password again here? If that's not easy, what you have here is fine for now

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1128,3 +1130,91 @@ int Servatrice_DatabaseInterface::checkNumberOfUserAccounts(const QString &email
 
     return 0;
 }
+
+bool Servatrice_DatabaseInterface::addForgotPassword(const QString &user)
```
Daenyth: Store the requestor ip address and client id

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1128,3 +1130,91 @@ int Servatrice_DatabaseInterface::checkNumberOfUserAccounts(const QString &email
...
+
+bool Servatrice_DatabaseInterface::updateUserToken(const QString & token, const QString &user)
+{
+	if (!checkSql())
+		return false;
+
+	if (token.isEmpty() || user.isEmpty())
+		return false;
+
+	QSqlQuery *query = prepareQuery("update {prefix}_users set token = :token where name = :user_name");
```
Daenyth: Can you tell me how this update method is used? I'm a little leery of reusing the users token field - I thought that was used only for non-activated accounts? It seems like we should store the token in the forgot password table, and then only check there when we get the reset-with-token proto message.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1054,14 +1057,93 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountPassword(const C
...
+
+Response::ResponseCode AbstractServerSocketInterface::cmdForgotPasswordReset(const Command_ForgotPasswordReset &cmd, ResponseContainer &rc)
+{
+	Q_UNUSED(rc);
+	qDebug() << "Received forgot password reset from user: " << QString::fromStdString(cmd.user_name());
+
+	if (!sqlInterface->doesForgotPasswordExist(QString::fromStdString(cmd.user_name())))
+		return Response::RespFunctionNotAllowed;
+
+	// if (!sqlInterface->validateUserToken(QString::fromStdString(cmd.token()), QString::fromStdString(cmd.user_name())))
```
Daenyth: Don't leave in commented code

servatrice/src/serversocketinterface.cpp
```diff
@@ -1054,14 +1057,93 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountPassword(const C
...
+
+Response::ResponseCode AbstractServerSocketInterface::cmdForgotPasswordChallenge(const Command_ForgotPasswordChallenge &cmd, ResponseContainer &rc)
+{
+	Q_UNUSED(rc);
+	qDebug() << "Received forgot password challenge from user: " << QString::fromStdString(cmd.user_name());
+
+	if (sqlInterface->doesForgotPasswordExist(QString::fromStdString(cmd.user_name())))
+		return Response::RespOk;
+
+	//if (sqlInterface->validateUserEmail(QString::fromStdString(cmd.email()), QString::fromStdString(cmd.user_name())))
```
Daenyth: 🔪

servatrice/src/smtpclient.cpp
```diff
@@ -82,6 +82,54 @@ bool SmtpClient::enqueueActivationTokenMail(const QString &nickname, const QStri
...
+		return false;
+	}
+
+	QxtMailMessage message;
+	message.setSender(name + " <" + email + ">");
+	message.addRecipient(recipient);
+	message.setSubject(subject);
+	message.setBody(body.replace("%username", nickname).replace("%token", token));
+
+	int id = smtp->send(message);
```
Daenyth: Is there an error state you can check for after calling `send`? If one exists we should make sure it's OK before returning `true`

servatrice/migrations/servatrice_0020_to_0021.sql
woogerboy21: I can add the two, though I had intended on creating an audit PR next that would record that information.  Since the functionality was designed to mimic the registration process and it doesn't record any of that information I didnt include it in the forgot password.

servatrice/servatrice.ini.example
woogerboy21: I can, given this is different logical functionality I didn't want to re-use parts of something else that appeared to be separated in its own since.  If you want something like a single "from" address should the "from" ini configuration setting be moved out of the registration into the smtp section?

servatrice/src/servatrice_database_interface.cpp
woogerboy21: Yes, that is what this was for.  I can't say I looked else were the strict mode may be affecting but this does resolve the issue when user account information is added in strict mode.

servatrice/src/servatrice_database_interface.cpp
woogerboy21: See comment : https://github.com/Cockatrice/Cockatrice/pull/2393#discussion_r100698083
Since everyone seemed really interested in a table for auditing , the very next PR I intend to submit (but left it out of this PR for clarity) is to create an audit table that will start with auditing information on registration, forgot password, and log queries done by admins that I can see the possibility of expanding to other aspects at some point.  Have to start some were though.

servatrice/src/servatrice_database_interface.cpp
woogerboy21: Currently the token field is used to store activation tokens generated for users when registering.  Once the account is active the code is no longer used any were else in the system.  Since the token is no longer needed after activation I see no issue with generating a new token, storing it here and sending it to the user for forgot password reasons.  Any reason you can think of?

servatrice/src/smtpclient.cpp
woogerboy21: For starters, this works the exact same way the registration works (so not sure why we want it to work differently), the other thing to mention is all this does is add the email to the que that is later processed at the next timer countdown.  I could look to see if the function in the smtp client returns something different or a value to check but I'm not sure if that should really hold the process up.  The end result would be the user looks for an email that simply never comes and then contacts the server operator for help.  Inconvenience sure, but its not really any different than the current behavior when a user goes to register.

common/pb/session_commands.proto
woogerboy21: I can make any of them required that you like, I didnt want to cause any issue's with future protocol manipulations on this process that's all.  I'll make user_name for sure.

cockatrice/src/dlg_forgotpasswordchallenge.cpp
woogerboy21: This line of code was simply pulled from the registration dialog.

cockatrice/src/dlg_forgotpasswordrequest.cpp
woogerboy21: See: https://github.com/Cockatrice/Cockatrice/pull/2393#discussion_r100698496

cockatrice/src/dlg_forgotpasswordreset.cpp
woogerboy21: See: https://github.com/Cockatrice/Cockatrice/pull/2393#discussion_r100698496

woogerboy21:
I'll make the changes @Daenyth, check out my responses when you get a chance @Daenyth .  Pretty sure when I remove tabs it will squash some of the questions asked in the thread.

servatrice/migrations/servatrice_0020_to_0021.sql
woogerboy21: Oh also, if you dont specify a value yet set it to "NOT NULL" you are returning to the strict mode problem you hate?

servatrice/src/servatrice.cpp
woogerboy21: I waffle with this, I can see your point but the code itself lies inside a function already (the statusUpdate()) .  Unless the recommended practices have changed (as they do over times and I am old), I was taught to not call functions from other functions as jumping through functions to finish tasks is troublesome to troubleshoot *unless* the function is used else were in the code (in this case it is not).

servatrice/src/serversocketinterface.cpp
woogerboy21: Yup, this was commented out when I built the challenge questions.  I'll remove them.

servatrice/src/serversocketinterface.cpp
woogerboy21: Yup, this was commented out when I built the challenge questions.  I'll remove them.

cockatrice/src/dlg_forgotpasswordchallenge.cpp
woogerboy21: Done. 👍

servatrice/src/servatrice_database_interface.cpp
woogerboy21: This code block is the block used for when users request a password change on the profiles tab.  I simply re-used the function of code adding a "force" flag allowing the function to be called and password update performed with out knowing the old password (since the user won't know it with the forgot password functionality).  I didn't re-invent the wheel 😄 

By the looks of things it already uses the same code that checks the password on line 859 (13 lines below the one you tagged) .

woogerboy21:
Alright, changes made (with the exception to the questions I posed back to  @Daenyth for things like the dialog.  The testing server mentioned above in the thread has been updated to have the new code on it so feel free to test away.

cockatrice/src/dlg_forgotpasswordchallenge.cpp
woogerboy21: BTW,  I havent looked at the code for the getHostname function but the behavior I am seeing in my windows client is the dialog's present the last host I connected to and if there is none listed it returns the server name "cockatrice.woogerworks.com".  Not sure if that is what the expected behavior is nor do I know if something like a mac client behaves differently.

servatrice/migrations/servatrice_0020_to_0021.sql
ZeldaZach: The audit table can contain that information at a later time. I also agree that not having a default here could be troublesome. This default timestamp is fine.

ZeldaZach:
I've done my review and I do plan on over-ruling @Daenyth on this case. We will have an audit table added in a future PR (Which is not in the so far future) so there's no need to do it here. Repetitive data is not fully needed, so if @woogerboy21 agrees that this is good to go I'm ready to :shipit:

woogerboy21:
The only thing I can see is maybe the questions to this stuff:  https://github.com/Cockatrice/Cockatrice/pull/2393#discussion_r100694127

I don't believe it needs changed but would like confirmation of behavior on other OS's (like mac).

PR #2414, opened 2017-02-15
Add cockatric.es server

tritoch:
Add cockatric.es server hosted by @tritoch to default server list


woogerboy21:
👍 

Just out of curiosity.  I think I registered cockatrice.us a LONG time ago (may have lost its registration by now).  Are we wanting as a group to start using the cockatrice.{location} type naming?

Issue #2425, opened 2017-02-25
Adding a keyboard shortcut for creating related tokens


MarkyMarkMcDonald:
### Feature Suggestion
A keyboard shortcut for creating all related cards of the selected card.

For cards with one related card, the shortcut text would be added to the single related card.
For cards with more than one related card, another menu item would be added saying "Create all related cards <shortcut-text>"

I think the default shortcut could be Cmd + Shift + T (Cmd + T being create token on OSX).

------
### Development
I'm working on a PR for this. My plan is to add a new method to Player.cpp that sends a create token command to the server for each related card (extracting a re-useable method from the create related card action).

One thing that has come up is that `relatedCards` and `reverseRelatedCards2Me` are both present in the related cards menu. Is it ok to create a token for every item in both lists? 

Cheers

Daenyth:
Is "Create all related cards" something you ever really want to do though? Maybe it should do nothing, or prompt?

MarkyMarkMcDonald:
Good point. I was thinking of cards like `Wurmcoil Engine` that always generate multiple kinds of tokens, but not cards like `One Dozen Eyes` which give the player a decision between different kinds of tokens.

I was curious about how many cards created different kinds of tokens and how many created a subset of different kinds so I created a quick search for card texts that create multiple tokens and scanned the list:

`cat AllCards.json | jq '.[].text | strings' | egrep "(create|put).*token.*(create|put|and|or).*token"`

Here are cards that give multiple kinds of tokens:
- Wurmcoil Engine
- Trostani's Summoner
- Sylvan Offering

Cards that produce multiple kinds at different times or produce a subset:
- One Dozen Eyes
- Hold the Perimeter

Cards that sometimes put all into play, but other times produce a subset:
 - Fable of Wolf and Owl

It seems like there is a mix of both. I think it's better to give power users a hotkey that creates both and then select + remove the one you don't want than to prompt or do nothing. If users find creating all of them when they only need one annoying they can still open up the card menu and choose the one they need.

tritoch:
Your search missed a few cards that have multiple related because of flip, transform, or multiple abilities.
ex:
- Huntmater of the Fells
- Garruk, Primal Hunter
- Bestial Menace

That said, I think a shortcut would be a great option for 99% of cards.  I agree it's easier to destroy a token than go back in to the menus and make the ones you haven't made.

Daenyth:
That's a solid argument. I'm sold

[...]

PR #2426, opened 2017-02-26
Add keybaord shortcut for creating related cards

MarkyMarkMcDonald:
Finishes #2425.

create all related cards example

Once again, c++ isn't my forte, so let me know if I can clean anything up.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1103,9 +1103,14 @@ void Player::actCreateRelatedCard()
     // get the target card name
     QAction *action = static_cast<QAction *>(sender());
 
+    const QString &cardName = action->text();
+    createCard(sourceCard, cardName);
+}
+
+void Player::createCard(const CardItem *sourceCard, const QString &cardName) {
     // removes p/t from tokens (and leading space))
     // Added split for "Token:" due to change in PR fixing #2317
-    QStringList spaces = action->text().split(tr("Token: "))[1].split(" ");
+    QStringList spaces = cardName.split(tr("Token: "))[1].split(" ");;
     if (spaces.at(0).indexOf("/") != -1) // Strip space from creatures
         spaces.removeFirst();
     CardInfo *cardInfo = db->getCard(spaces.join(" "));
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2355,9 +2355,11 @@ void Player::updateCardMenu(CardItem *card)
 
                 if(card->getInfo())
                 {
-                    QStringList relatedCards = card->getInfo()->getRelatedCards();
-                    QStringList reverserelatedCards2Me = card->getInfo()->getReverseRelatedCards2Me();
-                    if(relatedCards.size() || reverserelatedCards2Me.size())
+                    QStringList relatedCards = * new QStringList();
+                    relatedCards.append(card->getInfo()->getRelatedCards());
+                    relatedCards.append(card->getInfo()->getReverseRelatedCards2Me());
+
+                    if(!relatedCards.empty())
                     {
                         cardMenu->addSeparator();
                         for (int i = 0; i < relatedCards.size(); ++i) {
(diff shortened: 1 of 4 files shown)


tooomm:
I'm not sure if this is the best/right approach.
I'll leave a more detailed comment for discussion in the issue soon.

cockatrice/src/player.cpp
```diff
@@ -1095,20 +1095,34 @@ void Player::actCreatePredefinedToken()
...
+
+void Player::actCreateAllRelatedCards()
+{
+    CardItem * sourceCard = game->getActiveCard();
+    if(!sourceCard)
+        return;
+
+    QStringList relatedCards = * new QStringList();
+    relatedCards.append(sourceCard->getInfo()->getRelatedCards());
+    relatedCards.append(sourceCard->getInfo()->getReverseRelatedCards2Me());
```
ctrlaltca: While creating a new QStringList makes the code more readable, the second "append" forces Qt to make a deep copy of the list, with all the overhead involved in memory allocation, copy, etc..
This is not a big deal since the "related cards" list would only contain a few hundreds of bytes at max..

cockatrice/src/player.cpp
```diff
@@ -1095,20 +1095,34 @@ void Player::actCreatePredefinedToken()
...
+
+    for (int i = 0; i < relatedCards.size(); i++)
```
ctrlaltca: You can use `foreach` for QStringList:
```
foreach (const QString &tokenName, relatedCards)
{
  createCard(sourceCard, dbNameFromTokenDisplayName(tokenName));
}
```

cockatrice/src/player.cpp
```diff
@@ -1177,6 +1191,21 @@ void Player::setCardAttrHelper(const GameEventContext &context, CardItem *card,
     }
 }
 
+// token names take the form of "<Descriptors> <Power>/<Toughness> <Card Name>   " or "<Card Name>   ".
+// dbName for tokens should take the form of "<Card Name>   ".
+// trailing whitespace is significant; it is hacked on at the end as an additional identifier in our single key database
+QString Player::dbNameFromTokenDisplayName(const QString &tokenName) {
```
ctrlaltca: 👍  Nice refactoring, it makes clear the meaning of this code.

cockatrice/src/player.cpp
```diff
@@ -2425,6 +2424,44 @@ void Player::updateCardMenu(CardItem *card)
     }
 }
 
+void Player::addRelatedCardActions(const CardItem *card, QMenu *cardMenu) {
```
ctrlaltca: 👍  Another nice refactoring, less duplicated code

cockatrice/src/player.h
```diff
@@ -4,6 +4,7 @@
 #include <QInputDialog>
 #include <QPoint>
 #include <QMap>
+#include <QRegExp>
```
ctrlaltca: You can move this into the .cpp file, since it's not used inside the .h code.
It will avoid the inclusion of `<QRegExp>` in all the other files #including `player.h` but not using it.

cockatrice/src/player.cpp
```diff
@@ -2286,7 +2315,14 @@ void Player::actPlayFacedown()
 void Player::refreshShortcuts()
 {
     if(shortcutsActive)
+    {
         setShortcutsActive();
+
+        for (int i = 0; i < table->getCards().size(); ++i) {
```
Daenyth: Prefer using `foreach` here also as ctrl said

cockatrice/src/player.cpp
Daenyth: To be safe I think we should have nullptr checks on both arguments.

MarkyMarkMcDonald:
Thanks for the feedback - I think I handled everything except for the deep copy part. To me the deeper issue is that the this is the only place in the code base outside of importing from oracle that we refer to either of `reverseRelatedCards2Me` or `relatedCards`, and we combine those lists together in both cases. I'm not sure what the history is behind adding `getReverseRelatedCards2Me` - I'd rather clean it up at some other point and move `reverseRelatedCards2Me` into `relatedCards` at the data import level.

ZeldaZach:
I'm happy with the change. nice job!

ctrlaltca:
Merged, thank you!

Daenyth:
> I'd rather clean it up at some other point and move reverseRelatedCards2Me into relatedCards at the data import level.

You're right - only the card db should care about the difference. The rest of the code would rather just one list. I'd happily accept a PR with such a refactor. Thanks for helping improve the health of the codebase! It's grown really organically over time and needs some aggressive pruning and decoupling

PR #2432, opened 2017-03-02
fix folder names

### LINKED ISSUE #2429: [Bug] OnlineDeck Storage names with '/' are unusable
<b>OS:</b>
Linux arch , widnows 10
<b>Cockatrice version:</b>
on Arch b9cd942(compiled from master)
on windows dab7316

___
<br>
OnlineDeck Storage names with '/' are unusable

Folder with this name can't be deleted

The folder 'legacy/modern' is unusable, can't add deck to it neither delete it. No message of error or anything

ZeldaZach:
Fix #2429 

This should prevent folders from getting a troubling "/" character. Will replace with a backslash, which my testing showed worked fine.

### DIFF
--- a/cockatrice/src/tab_deck_storage.cpp
+++ b/cockatrice/src/tab_deck_storage.cpp
@@ -271,7 +271,7 @@ void TabDeckStorage::actNewFolder()
     if (folderName.isEmpty())
         return;
 
-	// '/' isn't a valid filename character on *nix so we're choosing to replace it with a different arbitrary character.
+    // '/' isn't a valid filename character on *nix so we're choosing to replace it with a different arbitrary character.
     std::string folder = folderName.toStdString();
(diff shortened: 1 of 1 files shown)


cockatrice/src/tab_deck_storage.cpp
```diff
@@ -270,6 +270,8 @@ void TabDeckStorage::actNewFolder()
     QString folderName = QInputDialog::getText(this, tr("New folder"), tr("Name of new folder:"));
     if (folderName.isEmpty())
         return;
+    std::string folder = folderName.toStdString();
+    std::replace(folder.begin(), folder.end(), '/', '\\'); // Fix #2429
```
Daenyth: You can remove the comment; include that info in the git commit. A good comment here might be
```
// '/' isn't a valid filename character on *nix, replace it to avoid problems storing on disk
```

PR #2433, opened 2017-03-02
Revert "Condensed Card Counter Context Menu"

### LINKED ISSUE #2402: Revert new nested counter menu
Ok this may not be explained correctly, Bear with me. Recently there was a build with the Token creation process changed. Specifically counters. You would select your token, and then put counters on.  Now you have to select color and counters seperately adding a 3rd step to this process... This is not an improvement more like a step backwards. because of this people are using tokens less actually and its very annoying and rude to not use the Actual tokens. We should revert this back to the way it was.

ZeldaZach:
Reverts Cockatrice/Cockatrice#2348
Fix #2402

Based on initial feedback, this feature was not well received. As such, we are reverting back. May consider adding an option to "condense menus" at some point, but it won't be today unfortunately :(

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -26,9 +26,6 @@ CardItem::CardItem(Player *_owner, const QString &_name, int _cardid, bool _reve
     cardMenu = new QMenu;
     ptMenu = new QMenu;
     moveMenu = new QMenu;
-    addCounterMenu = new QMenu;
-    removeCounterMenu = new QMenu;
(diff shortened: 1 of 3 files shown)


tooomm:
this fixes #2402

Jinxstands4Jinx:
so when will this be implemented or must you redownload something?

Nightfirecat:
It'll be in the next release, or you can download a build of that commit if you want to start using it sooner.

skwerlman:
i will look forward to the condense option in future builds, then

PR #2434, opened 2017-03-02
fix buddy checkbox bug

### LINKED ISSUE #2159: Buddies only filter buggy
<b>OS:</b>Win 7 SP1
<b>Cockatrice version:</b>a3a6065 (2016-08-20)

---

If I restart cockatrice it remembers that I had buddies only games filtered out last time I was logged on, and it shows the tickbox as unticked, but still shows buddies only games. Fixed by re-ticking the box twice.
This is a very minor issue that hasn't really affected me in any notable way, but I thought I might as well report it.
In case my point got lost in poor choice of words (happens too often, really), here's [overly specific screenshots](http://imgur.com/a/9LnH2)

### LINKED ISSUE #2399: Filter Games option forgets "Show Buddies Only Games" status
<b>OS:</b>
Windows 10

<b>Cockatrice version:</b>
F7C8651D
___
<br>

I find that Cockatrice occasionally forgets what I have selected for "Show 'buddies only' games". I always have it turned off because I'm a crusty hate filled individual who never plays in buddies only games, but sometimes it turns back on. I've even had it on a few rare occasions show me buddies only games when it's been unchecked.
Can't offer any steps to reproduce sorry, just thought I'd mention it to document it.

ZeldaZach:
fix #2399 
fix #2159

The buddy legacy setting was never translated over to the startup operation, which might be why it flickers from time to time. This should address that problem.


tooomm:
also fixing #2159?

PR #2435, opened 2017-03-02
add player numbers to in game

### LINKED ISSUE #2417: while in-game no way to tell the player count
*Especially relevant for draft or other bigger multiplayer games*

While you're in a game lobby and you're waiting for enough other players to join, you can't tell how much players are in the game or how many are missing.
You have to either
(1) switch to the main tab and find your game to check (must fit filter settings for sure),
(2) right click a player in your game lobby and check `show this users games`
(3) manually count the players in the player list

For like 4 player EDH you immediately know just by looking, but for drafts, other big games or people with eye restraints this might be an issue.
All options are suboptimal in this case, especially for draft games which need 8 people.

In a draft lobby you have to wait quite some time to find enough people because others also leave in between for example.
So what do you do while waiting? You're either chatting with the other players or you tab out and randomly check how it's going. But you always want to quickly+easily know how much people are missing.

The player count information, which we track already, just needs to be displayed somewhere in the game room, for example in the name of the `player list` widget.
It could also be in the name of the tab for the game, but it's already low on space and might lead to more truncation of other information there.

Maybe I even suggested something similar earlier, not sure
It just immediately popped to my face while I did a first draft game in a long while... really annoying.

ZeldaZach:
Fix #2417

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -69,6 +69,7 @@
 #include "pb/context_connection_state_changed.pb.h"
 #include "pb/context_ping_changed.pb.h"
 #include "get_pb_extension.h"
+#include "pb/serverinfo_game.pb.h"
 
 ToggleButton::ToggleButton(QWidget *parent)
         : QPushButton(parent), state(false)
@@ -448,7 +449,8 @@ TabGame::~TabGame()
 void TabGame::retranslateUi()
 {
     QString tabText = " | " + (replay ? tr("Replay") : tr("Game")) + " #" + QString::number(gameInfo.game_id());
-    QString userCountInfo = " " + QString::number(gameInfo.player_count()) + "/" + QString::number(gameInfo.max_players());
+    QString userCountInfo = QString(" %1/%2").arg(gameInfo.player_count()).arg(gameInfo.max_players());
+    qDebug() << "User Count Info: " << userCountInfo;
 
     cardInfoDock->setWindowTitle(tr("Card Info") + (cardInfoDock->isWindow() ? tabText : QString()));
     playerListDock->setWindowTitle(tr("Player List") + userCountInfo + (playerListDock->isWindow() ? tabText : QString()));
@@ -1032,7 +1034,10 @@ void TabGame::eventJoin(const Event_Join &event, int /*eventPlayerId*/, const Ga
         messageLog->logJoin(newPlayer);
     }
     playerListWidget->addPlayer(playerInfo);
+    qDebug() << "eventJoin retranslate w/ count = " << gameInfo.player_count();
+    retranslateUi();
     emitUserEvent();
+
 }
 
 void TabGame::eventLeave(const Event_Leave & /*event*/, int eventPlayerId, const GameEventContext & /*context*/)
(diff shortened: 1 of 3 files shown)


tooomm:
Works very nice on game room join and displays the count in the title of the widget or undocked window. Also works in tabbed mode.

But if players join it doesn't update at all and keeps the initial value:
untitled

So what's the difference, that it instantly updates in the game view list but not here?

cockatrice/src/tab_game.cpp
```diff
@@ -1029,16 +1034,30 @@ void TabGame::eventJoin(const Event_Join &event, int /*eventPlayerId*/, const Ga
     } else {
         Player *newPlayer = addPlayer(playerId, playerInfo.user_info());
         messageLog->logJoin(newPlayer);
+		playerCountInRoom++;
     }
     playerListWidget->addPlayer(playerInfo);
+	
+    qDebug() << "eventJoin retranslate w/ count = " << playerCountInRoom;
+    retranslateUi();
```
Daenyth: I think `emitUserEvent` causes a retranslateUi to fire already - can you check?

cockatrice/src/tab_game.cpp
```diff
@@ -420,6 +421,8 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor, QList<AbstractClient *> &_client
 
     this->installEventFilter(this);
     QTimer::singleShot(0, this, SLOT(loadLayout()));
+	
+	playerCountInRoom = gameInfo.player_count();
```
ctrlaltca: Tabs => spaces

cockatrice/src/tab_game.cpp
```diff
@@ -448,9 +451,11 @@ TabGame::~TabGame()
 void TabGame::retranslateUi()
 {
     QString tabText = " | " + (replay ? tr("Replay") : tr("Game")) + " #" + QString::number(gameInfo.game_id());
+    QString userCountInfo = QString(" %1/%2").arg(playerCountInRoom).arg(gameInfo.max_players());
+    qDebug() << "User Count Info: " << userCountInfo;
 
     cardInfoDock->setWindowTitle(tr("Card Info") + (cardInfoDock->isWindow() ? tabText : QString()));
-    playerListDock->setWindowTitle(tr("Player List") + (playerListDock->isWindow() ? tabText : QString()));
+    playerListDock->setWindowTitle(tr("Player List") + userCountInfo + (playerListDock->isWindow() ? tabText : QString()));
```
ctrlaltca: You can get the number of players from the PlayerListWidget, just adding a method like:
```cpp
int PlayerListWidget::getPlayersCount() { return players.size(); }
```

ZeldaZach:
@ctrlaltca Can you take over this PR? I'm not sure how to get it to work properly as nothing I've done suffices to get the correct data.

ctrlaltca:
> @ctrlaltca Can you take over this PR? I'm not sure how to get it to work properly as nothing I've done suffices to get the correct data.

Please look https://github.com/ZeldaZach/Cockatrice/pull/3

ZeldaZach:
This works as expected now. Thanks @ctrlaltca.
Can somebody review?

ctrlaltca:
> Care to explain why your method works?

The TabGame object already takes care of maintaining a list of the actual players and spectators. I just get the information from it when needed, instead of maintaining another counter.

tooomm:
Working now. It also just counts real players, no spectators! 👍

PR #2437, opened 2017-03-02
Enable 'Show Games' option in Self Context Menu

### LINKED ISSUE #2381: Can't "Show this user's games" on myself
<b>OS:</b>Win 7 SP 1
<b>Cockatrice version:</b> f7c8651d (2016-12-30)
___
The option to RMB myself and show the list of my games is disabled, why? Some things such as whether the room is buddies only, registered only, password protected or spectator rights aren't visible from inside the room. Sometimes, when no one is joining for a long time for example, I want to check that I didn't leave the room as buddies only on accident, and finding it in main is often much slower and less convenient than checking my own games.

tritoch:
Fixes #2381


tritoch:
Is there even a context menu when offline?  I guess `true` does the same thing?

Daenyth:
You can be disconnected from ping and such, and still have usernames to right click in the log, so `online` is right I think

PR #2440, opened 2017-03-04
make sure xml file is uploaded

### LINKED ISSUE #2065: `add custom set/cards` doesn't error on non .xml files
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
e05e66ae (2016-07-03)

___
<br>

If you try to import any other file than an `.xml` with the `Card Database --> Add custom set/cards` option it won't give you any error. It won't work either since we can just work with a valid xml file there.

 - we should check the file format
 - we should validate the xml against our [xsd](https://github.com/Cockatrice/Cockatrice/blob/master/doc/cards.xsd)

ZeldaZach:
Addresses bulletpoint 1 of #2065 

This will now require an XML file to be uploaded. We _dont_ check for well formatted XML files, which I find to be a problem. That would probably fix some other tickets I saw earlier (That @tooomm will link shortly haha).

We should incorporate a well-formated check for all XML files in Cockatrice on initial load(). Just something for another PR later.

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -1030,7 +1030,8 @@ void MainWindow::actAddCustomSet()
         QMessageBox::warning(this, tr("Load sets/cards"), tr("Selected file cannot be found."));
         return;
     }
-    else if (QFileInfo(fileName).suffix() != "xml") { // fileName = *.xml
+
+    if (QFileInfo(fileName).suffix() != "xml") { // fileName = *.xml
         QMessageBox::warning(this, tr("Load sets/cards"), tr("You can only import XML databases at this time."));
         return;
     }
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -1030,6 +1030,10 @@ void MainWindow::actAddCustomSet()
         QMessageBox::warning(this, tr("Load sets/cards"), tr("Selected file cannot be found."));
         return;
     }
+    else if (QFileInfo(fileName).suffix() != "xml") { // fileName = *.xml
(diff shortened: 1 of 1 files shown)


cockatrice/src/window_main.cpp
```diff
@@ -1030,6 +1030,10 @@ void MainWindow::actAddCustomSet()
         QMessageBox::warning(this, tr("Load sets/cards"), tr("Selected file cannot be found."));
         return;
     }
+    else if (QFileInfo(fileName).suffix() != "xml") { // fileName = *.xml
```
ctrlaltca: Just a minor remark: no need to `else`, since the previous `if` block contains a `return`.

tooomm:
I can't come up with other tickets at the moment. :P
But I found a very similar one which is referenced in the initial issue now.

I updated the pr/issue to keep track of the leftovers.^^

cockatrice/src/window_main.cpp
```diff
@@ -1031,6 +1031,11 @@ void MainWindow::actAddCustomSet()
         return;
     }
 
+    if (QFileInfo(fileName).suffix() != "xml") { // fileName = *.xml
```
Daenyth: Could we prevent non-xml files from being shown to the user at all by using `dialog.setNameFilters` ?

ZeldaZach: People can still override that, it's just a suggestion the opener can recommend (like Finder will say I can do "XML" or "All Files" )

ctrlaltca:
Do you plan to add the code suggested by @Daenyth or should we go with this PR as it is?

ZeldaZach:
I'm going with what is here for now

PR #2441, opened 2017-03-04
add additional features and remove unneeded functions to set editor

### LINKED ISSUE #2009: Shift-clicking in `Edit Sets`
<b>OS:</b>
Sabayon 16.05 Cinnamon/MDM

<b>Cockatrice version:</b>
0baccd3

---

<br>

Currently when installing Cockatrice for the first time, you are presented with an 'Edit Sets' window. When I am setting up my sets, I find myself disabling entire types of sets (e.g. Reprint sets).

It would be nice if I could shift-click to select multiple sets in this window, since clicking individually to disable each and every set I don't want gets tiring fast, especially since I reinstall my OS fairly often.

ZeldaZach:
Fix #2009 

This will allow for multi-set selection on the Editor. You can use the standard shift/command keys to make your selections (in range, single elements, etc).

In addition, I changed how the buttons operated to be toggles now. You can toggle selections or the entire board. I feel this change is better then our current system and I hope you like! I learned a lot from this PR about the different QT classes and how they interact :)


cockatrice/src/setsmodel.cpp
```diff
@@ -132,6 +132,12 @@ void SetsModel::toggleRow(int row, bool enable)
     emit dataChanged(index(row, EnabledCol), index(row, EnabledCol));
 }
 
+bool SetsModel::getToggleStatus(int row)
+{
+    CardSet *temp = sets.at(row);
+    return enabledSets.contains(temp) ? true : false;
```
ctrlaltca: Just a minor remark: [QSet::contains()](http://doc.qt.io/qt-5/qset.html#contains) already returns a `bool`, so the `? true : false` part is not needed.

tooomm:
- There is a problem with `Toggle all sets`.
If you hit the button for the first time all sets get disabled (if at least one set is already enabled). The second hit will enable all sets. After that the toggle behaves as intended and correctly toggles every status.
But it should actually properly toggle on first use correctly as well. Not just disable all sets.<br>
But for general purpose I liked the `enable and disable all sets` buttons, because most people either enable all or limited players for example disable all before selecting 1-2 relevant sets.
Since we magically somehow have this feature right now with the "broken" `toggle all sets` 2-in1 button... why not just rename it to `Disable/Enable all sets`? It perfectly describes what it's doing and is very useful the way it works actually. :)

 - Another problem is that if you select several sets and use the up/down button things get messy because just one of the selected sets moves!

cockatrice/src/setsmodel.cpp
ZeldaZach: Good spotting, I do my work late at night so sometimes things slip by me

ZeldaZach:
@tooomm There's nothing I can do about your first issue. People can clearly see which sets are enabled/not so I see no harm in it needing to be clicked twice sometimes. It all depends on the value of the first index, and Toggle is a shorter word for "Disable/Enable"

Also, the built in arrow keys aren't really needed as you can use a scroll wheel or something similar to view the entire list, so I'm not really going to say much about it. I find the changes made sufficient up to this point, but if you still disagree please let me know.

cockatrice/src/setsmodel.cpp
```diff
@@ -132,6 +132,12 @@ void SetsModel::toggleRow(int row, bool enable)
     emit dataChanged(index(row, EnabledCol), index(row, EnabledCol));
 }
 
+bool SetsModel::getToggleStatus(int row)
```
Daenyth: 👎

cockatrice/src/window_sets.cpp
```diff
@@ -147,32 +147,18 @@ void WndSets::selectRow(int row)
...
-        return;
-
-    model->toggleRow(rows.first().row(), false);
-}
-
-void WndSets::actEnableAll()
-{
-    model->toggleAll(true);
+    foreach(QModelIndex i, rows)
+        model->toggleRow(i.row(), !model->getToggleStatus(i.row()));
```
Daenyth: You should be able to add a helper function such that `model->toggleRow(i.row())` is correct.

tooomm:
>@tooomm There's nothing I can do about your first issue. People can clearly see which sets are enabled/not, so I see no harm in it needing to be clicked twice sometimes. It all depends on the value of the first index, and Toggle is a shorter word for "Disable/Enable"

You miss my point here, see:
- `Toggle set` works as a proper toggle for a single set and multiple selected sets - it switches all selected ones no matter of their state to their other state (like a toggle should do...). All enabled ones get disabled and the other way around, **even when you select several sets with different statuses at the same time!**
- Exactly that doesn't work (as the name toggle suggests) for `toggle all sets`. **It'll always disable all sets first, no matter of their current state - a second click then enables all. But it should actually swap each state for every set, not just disable them all first**.<br>I actually like that behavior, but the wording `Disable/Enable sets` better matches this behavior and makes that more clear imo.

Check for yourself.^^
<br>

>Also, the built in arrow keys aren't really needed as you can use a scroll wheel or something similar to view the entire list, so I'm not really going to say much about it. I find the changes made sufficient up to this point, but if you still disagree please let me know.

The arrow buttons are not needed for scrolling! :)
You need them to adjust the order of elements in your list, which is an essential feature for many people. You need to adjust the position of particular sets if you want to see special pictures or prefer full art lands for example etc.
This is connected with picture downloading a lot and the only way to influence it right now!

Expected behavior with the new possibility to select multiple sets at the same time:
Move the complete group relative to their current position.
- +/- one position for each selected set
- all selection to top/bottom (respect the order within the selection)

woogerboy21:
Hopefully you have better responses from the community than Poixen did switching things around such as this ;)

cockatrice/src/setsmodel.cpp
```diff
@@ -124,28 +124,39 @@ void SetsModel::toggleRow(int row, bool enable)
...
 
     emit dataChanged(index(row, EnabledCol), index(row, EnabledCol));
 }
 
-bool SetsModel::getToggleStatus(int row)
+void SetsModel::toggleRow(int row)
 {
-    CardSet *temp = sets.at(row);
-    return enabledSets.contains(temp);
+    CardSet *tmp = sets.at(row);
```
Daenyth: What happens if I pass `-1` or `999999999` as the row? Need a nullptr check?

cockatrice/src/setsmodel.h
```diff
@@ -43,11 +43,11 @@ class SetsModel : public QAbstractTableModel {
     QStringList mimeTypes() const;
     void swapRows(int oldRow, int newRow);
     void toggleRow(int row, bool enable);
-    void toggleAll(bool enable);
+    void toggleRow(int row);
+    void toggleAll(int row);
```
Daenyth: I do't understand what this interface means? I toggle all sets but I'm telling it a row? Can you help me understand the intent here?

ZeldaZach: toggleAll will take a row and change all sets to copy that row's status (So if you pass it a row that's enabled, it will enable everything and if you pass it a disabled row, it will disable everything)

cockatrice/src/setsmodel.h
```diff
@@ -43,7 +43,8 @@ class SetsModel : public QAbstractTableModel {
     QStringList mimeTypes() const;
     void swapRows(int oldRow, int newRow);
     void toggleRow(int row, bool enable);
-    void toggleAll(bool enable);
+    void toggleRow(int row);
+    void toggleAll(int rowToCopyStatusFrom);
```
ZeldaZach: I made the variable more defined now

ctrlaltca:
I have to say that @tooomm has a point here.
When selecting a few sets and toggling them, it works as expected:
toggle1

When using the "toggle all" button, it won't really toggle each set as expected (enabling the disabled ones and viceversa), but instead it will enable or disable all the sets depending on the current status of the first set in the list:
toggle2

Maybe it would be easier to understand if the "toggle all" button works exactly as the "toggle selected" button, but on all elements.

tooomm:
>Maybe it would be easier to understand if the "toggle all" button works exactly as the "toggle selected" button, but on all elements.

👍 

But I prefer a simple `Disable all sets / Enable all sets` combo-button here all the time!
In many cases you just want to enable or disable all sets... and manually cherry-pick a few afterwards.

That's why I'm still for renaming the button and adjusting the behavior to consistently switch all sets off first (if there are any enabled ones), then enable all on the second click.
Or directly enable all, if you manually deselect all others before for example.

With the proper `toggle all sets` button you're actually not achieving too much. I can't think of any situation where I want to do exactly that actually...

tooomm:
So far so good.

The `++/+/-/--` buttons still don't work with multi-selection.
Just the upper set from the selection is moved accordingly, not all selected sets.

ZeldaZach:
Do you think this is good to go then tom?

tooomm:
I'm not happy with that "issue" around since it's something that people just expect to work.
They select their most favorite legacy sets all at once (which is now possible thanks to this pr 👍 ) and want to put them on the top to get the images for example. It will end up in some confusion, and searching around for which sets finally got moved etc.

But we could say it's just minor, skip it and open an issue about it to keep track. It's not a big deal breaker I think.

Daenyth:
> They select their most favorite legacy sets all at once (which is now possible thanks to this pr 👍 ) and want to put them on the top to get the images for example. 

Unless this used to work and this PR breaks it (sounds like not the case) we should merge and then do what you suggest as a new ticket

tooomm:
Of course it wasn't working before, because you just couldn't select several entries at once... ;P

But what was working with one set (and everybody is used to) now is working different for several sets. So the buttons didn't adopt to the new "select several at once" feature, and behave different for one selected set and multi selection which isn't ideal at all.

So fix now, or later - but fix please! :)

ZeldaZach:
@tooomm I updated this to now make the buttons work for multi-selected

cockatrice/src/setsmodel.cpp
```diff
@@ -124,22 +124,39 @@ void SetsModel::toggleRow(int row, bool enable)
...
     else
         enabledSets.remove(temp);
 
     emit dataChanged(index(row, EnabledCol), index(row, EnabledCol));
 }
 
-void SetsModel::toggleAll(bool enable)
+void SetsModel::toggleRow(int row)
 {
+    CardSet *tmp = sets.at(row);
```
Daenyth: `if (tmp == nullptr) return`

cockatrice/src/setsmodel.cpp
```diff
@@ -124,22 +124,39 @@ void SetsModel::toggleRow(int row, bool enable)
...
+    else
+        enabledSets.insert(tmp);
+
+    emit dataChanged(index(row, EnabledCol), index(row, EnabledCol));
+}
+
+void SetsModel::toggleAll(int row)
+{
+    bool enabled = true;
+    CardSet *tmp = sets.at(row);
```
Daenyth: nullptr

cockatrice/src/setsmodel.h
Daenyth: I still think it's weird behavior. Toggle all should do

if any row enabled:
  disable all
else:
  enable all

But as we talked about it can happen in another pr if you want. Now that the release is out though I don't see the rush, so it probably makes sense to just do it now

ZeldaZach: The current behavior:
`if (row1 isEnabled) { disableAllRows() } else { enableAllRows() }`

skwerlman:
Personally, I think `Enable all`/`Disable all` buttons would be more useful than a single `Toggle all` which just inverts each checkbox, since I can't imagine a case where I'd want to invert my selections.

tooomm:
> Personally, I think `Enable all` / `Disable all` buttons would be more useful than a single Toggle all which just inverts each checkbox, since I can't imagine a case where I'd want to invert my selections.

Actually you are 100% right. 👍 The real feature in this PR is the multi selection.
Inverting check boxes isn't helping in the normal way you interact with the interface when trying to accomplish something with sets.

But since we can now select several sets the `enable` and `disable` buttons would need to handle correctly the case if you select two sets - one disabled and the other enabled.

ZeldaZach:
@Daenyth Is this good to go?

Daenyth:
Looks like you're still doing `rowToCopyStatusFrom` which we're all saying is not helpful and should be changed.

tooomm:
Well, @skwerlman and my recent point is to revert the buttons all back to one `Enable selected sets` and one `Disable selected sets`, since the toggle stuff is fun, but not really helpful.

The initially requested and indeed helpful feature addressed in this PR is the multi selection with <kbd>Strg</kbd>!

ZeldaZach:
@Daenyth I do `rowToCopyStatusFrom` to just copy the first row's status to see should I enable all the rows or Disable all the rows, that's it...

should I be doing it another way?

skwerlman:
> Well, @skwerlman and my recent point is to revert the buttons all back to one Enable selected sets and one Disable selected sets button

Actually, my thought was to have 4 buttons: `Enable All` `Enable Selected` `Disable Selected` `Disable All`

tooomm:
Yeah sure... everything else makes not so much sense. I guess I mixed things up while writing a ton of things today. :D

I'm sure the buttons could be somehow smart as well, to stay with just two:
No set selected --> `Enable all` / `Disable all`
One or several sets selected --> `Enable selected set(s)` / `Disable selected set(s)`

ZeldaZach:
@tooomm I implemented the feature you suggested with smart buttons.
screenshot 2017-03-18 20 34 55
screenshot 2017-03-18 20 34 58

ZeldaZach:
@Daenyth care to recheck?

skwerlman:
I like the current version, but I don't think the "hints:" title is necessary; just use the bullet points.

ZeldaZach:
@tooomm It works somewhat, it's not so easy... I'll try again later or sometime during the week as i'm very busy now

tooomm:
All buttons are not working as expected for me on win.
I'm using latest Appveyor build (2h old) linked at the bottom of the pr: 0bb100c (2017-03-21)

Do we have a different understanding of the buttons? Is there some wrong hosted build?
Basically all my points from last review are still valid. I can't imagine that you didn't realized these problems - so I assume it's a windwos only problem?
@skwerlman can you please give this a try on Linux and tell me if you experience the same issues described here: https://github.com/Cockatrice/Cockatrice/pull/2441#pullrequestreview-27790717 ?

tooomm:
Proposal to get things unstuck here...

If several sets are selected, just disable the buttons on the upper left and just allow to move single sets around for the time being. More or less the same as it used to be regarding moving sets.
Then we open a ticket and look into proper "multi-selection moving" later. 👍

Issue #742, opened 2015-02-15
Feature Request: Known Servers with Online Status


woogerboy21:
So @ZeldaZach and I were discussing some of the confusion there is for clients to know when a server is online or offline and one of the idea's that was thought of was to add a drop down listing option on the login screen that lists a list of "known" servers that the user can configure under there preferences and also have some type of indicator on if the server is online or offline (such as a small red/green globe near the name some were).

How much effort would something like this take and what are others thoughts?

Daenyth:
I've been meaning to do a drop down list of servers. For online status we
could implement the ability to ping while offline or something

tooomm:
Connected to  #1307

Online/offline indicators or ping checks are left.

Daenyth:
Rough layout courtesy of asciiflow.com

```
+-----------------------------------------------------------------------------------------------+
| +-----------------------------------------------+                                             |
| |    Server 1                                   |                                             |
| |      Host: cockatrice.woogerworks.com         |      Username: ____                         |
| |      Port: 4747                               |      Password: ____                         |
| |                                               |      Autologin: [ ]                         |
| +-----------------------------------------------+      Remember password: [ ]                 |
| |                                               |                                             |
| |    Server 2                                   |                                             |
| |                                               |                                             |
| |                                               |                                             |
| |                                               |                                             |
| |                                               |                                             |
| |                                               |   +-----------------+   +---------------+   |
| |                                               |   |   Connect       |   |   Register    |   |
| |                                               |   |                 |   |               |   |
| |                                               |   +-----------------+   +---------------+   |
+-+-----------------------------------------------+---------------------------------------------+
```

tooomm:
I extended your mockup and added online/offline indicators in front of the servers.
There should also be an option to add new servers to the list!

```
+-----------------------------------------------------------------------------------------------+
| +-----------------------------------------------|                                             |
| |    Server 1                                | ||                                             |
| | ++   Host: cockatrice.woogerworks.com      | ||      Username: ____                         |
| | ++   Port: 4747                            | ||      Password: ____                         |
| |                                            | ||      Autologin: [ ]                         |
| +--------------------------------------------+ ||      Remember password: [ ]                 |
| |    Server 2                                | ||                                             |
| | ++                                         | ||                                             |
| | ++                                         | ||                                             |
| |                                            | ||                                             |
| +--------------------------------------------+ ||                                             |
| |    Server 3                                | ||                                             |
[...]

PR #2452, opened 2017-03-07
Server Connect Remade

### LINKED ISSUE #2070: "connect to server" list misses ports
<b>OS:</b>
win 10

<b>Cockatrice version:</b>
archives full release

---

<br>

On first fresh start the `Connect...` dialog looks like this:
firststart_connect

If you try to use one of the pre-made servers it won't have the info on `port` and you can't use it:
firststart_connect2
<br>
We should store a default & working port number next to the host address!
(People can still adjust it if they run into trouble because they are on managed networks etc. - the server needs to allow connections on different ports first anyway - poixen doesn't for example)

for woogerworks it's: 4747
for poixen it's: 4747
for chickatrice it's: 4747

ZeldaZach:
## Related Ticket(s)
- Fixes #2070
(If you can find others please update this list)

The dropdown menu had only one set of configurations when we had multiple servers, leading users to want to have same passwords across numerous servatrices.

This changes the behavior of the connection dialog to only save one set of settings for all servers to have multiple configurations that are changed when the dropdown menu is changed. You can either save the server settings by pressing the button OR when you connect the settings are automatically updated. Everything is based around Menu Name as a key, so if you try to add an element that's the same name, it will instead update the old key values.

## Screenshots
screenshot 2017-03-14 17 32 02
screenshot 2017-03-14 17 32 06
screenshot 2017-03-14 17 32 40

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -14,6 +14,7 @@
 #include <QPushButton>
 #include "dlg_connect.h"
 #include "settingscache.h"
+#include "userconnection_information.h"
 
 DlgConnect::DlgConnect(QWidget *parent)
     : QDialog(parent)
@@ -22,21 +23,16 @@ DlgConnect::DlgConnect(QWidget *parent)
     previousHosts = new QComboBox(this);
     previousHosts->installEventFilter(new DeleteHighlightedItemWhenShiftDelPressedEventFilter);
 
-    QStringList previousHostList = settingsCache->servers().getPreviousHostList();
-    if (previousHostList.isEmpty()) {
-        previousHostList << "cockatrice.woogerworks.com";
-        previousHostList << "chickatrice.net";
-        previousHostList << "mtg.tetrarch.co";
-        previousHostList << "cockatric.es";
-    }
-    previousHosts->addItems(previousHostList);
-    previousHosts->setCurrentIndex(settingsCache->servers().getPrevioushostindex());
+    rebuildComboBoxList();
 
     newHostButton = new QRadioButton(tr("New Host"), this);
     
+    saveLabel = new QLabel(tr("Menu Name:"));
+    saveEdit = new QLineEdit(settingsCache->servers().getSaveName());
+    saveLabel->setBuddy(saveEdit);
+
     hostLabel = new QLabel(tr("&Host:"));
-    hostEdit = new QLineEdit();
-    hostEdit->setPlaceholderText(tr("Enter host name"));
+    hostEdit = new QLineEdit(settingsCache->servers().getHostname());
     hostLabel->setBuddy(hostEdit);
 
     portLabel = new QLabel(tr("&Port:"));
@@ -58,11 +54,13 @@ DlgConnect::DlgConnect(QWidget *parent)
     autoConnectCheckBox = new QCheckBox(tr("A&uto connect"));
     autoConnectCheckBox->setToolTip(tr("Automatically connect to the most recent login when Cockatrice opens"));
 
-    if(savePasswordCheckBox->isChecked())
+    if (savePasswordCheckBox->isChecked())
     {
         autoConnectCheckBox->setChecked(settingsCache->servers().getAutoConnect());
         autoConnectCheckBox->setEnabled(true);
-    } else {
+    }
+    else
+    {
         settingsCache->servers().setAutoConnect(0);
         autoConnectCheckBox->setChecked(0);
         autoConnectCheckBox->setEnabled(false);
@@ -85,11 +83,13 @@ DlgConnect::DlgConnect(QWidget *parent)
     connectionLayout->addWidget(previousHostButton, 0, 1);
     connectionLayout->addWidget(previousHosts, 1, 1);
     connectionLayout->addWidget(newHostButton, 2, 1);
-    connectionLayout->addWidget(hostLabel, 3, 0);
-    connectionLayout->addWidget(hostEdit, 3, 1);
-    connectionLayout->addWidget(portLabel, 4, 0);
-    connectionLayout->addWidget(portEdit, 4, 1);
-    connectionLayout->addWidget(autoConnectCheckBox, 5, 1);
+    connectionLayout->addWidget(saveLabel, 3, 0);
+    connectionLayout->addWidget(saveEdit, 3, 1);
+    connectionLayout->addWidget(hostLabel, 4, 0);
+    connectionLayout->addWidget(hostEdit, 4, 1);
+    connectionLayout->addWidget(portLabel, 5, 0);
+    connectionLayout->addWidget(portEdit, 5, 1);
+    connectionLayout->addWidget(autoConnectCheckBox, 6, 1);
 
     QGridLayout *buttons = new QGridLayout;
     buttons->addWidget(btnOk, 0, 0);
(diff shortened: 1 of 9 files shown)


cockatrice/src/dlg_connect.cpp
```diff
@@ -22,12 +23,19 @@ DlgConnect::DlgConnect(QWidget *parent)
...
     previousHosts->installEventFilter(new DeleteHighlightedItemWhenShiftDelPressedEventFilter);
 
+	connect(previousHosts, SIGNAL(currentIndexChanged(int)), this, SLOT(updatePorts(int)));
+	
+	QMap<QString, UserConnection_Information> tmp = settingsCache->servers().getAllInfo();
+	
+	
+	
+
     QStringList previousHostList = settingsCache->servers().getPreviousHostList();
```
Daenyth: I think you don't need this list at all - the map should contain all the information necessary to populate the dialog

cockatrice/src/dlg_connect.cpp
```diff
@@ -145,6 +153,16 @@ void DlgConnect::previousHostSelected(bool state) {
     }
 }
 
+void DlgConnect::updatePorts(int num) {
```
Daenyth: We should key all operations off the server configuration name - let's have 

```
QMap<ServerName, ConnectionProfile>
where
ServerName = QString
ConnectionProfile = UserConnection_Information
```

cockatrice/src/dlg_connect.cpp
```diff
@@ -145,6 +153,16 @@ void DlgConnect::previousHostSelected(bool state) {
...
+    auto tmp = previousHosts->itemData(previousHosts->currentIndex()+1, Qt::DisplayRole);
```
Daenyth: Don't name things `tmp` - maybe `previousPort` ?

cockatrice/src/userconnection_information.cpp
```diff
@@ -0,0 +1,45 @@
...
+QMap<QString, UserConnection_Information> UserConnection_Information::getServerInfo()
+{
+    QMap<QString, UserConnection_Information> tmp = new QMap<QString, UserConnection_Information>();
+    
+    int size = settings.beginReadArray("server_details");
+    for (int i = 0; i < size; i++)
+    {
+        settings.setArrayIndex(i);
+        
+        QString server_name = settings.value("server").toString();
```
Daenyth: It seems like you're using server name as both the hostname and as the name of the profile - I think it would be better UX to let the users provide a name like

```
Server:
  Name: Woogerworks
  Host: cockatrice.woogerworks.com
  Post: 4747
  ...
```

Also this allows a very important thing - you can change the hostname of the server (for example using an ip address or an alternate dns name) without breaking the settings. Using hostname for both the key and the connection info means that changing the host will lose all associated settings

cockatrice/src/userconnection_information.cpp
```diff
@@ -0,0 +1,45 @@
...
+        QString port_num = settings.value("port").toString();
+        QString user_name = settings.value("username").toString();
```
Daenyth: Local variables are formatted as `camelCase`, not `snake_case`

ZeldaZach:
I made some more changes @Daenyth If you'd like to look

ZeldaZach:
I made some more changes @Daenyth If you'd like to look
cc: @ctrlaltca

ZeldaZach:
Current issue experienced:
```
[100%] Linking CXX executable cockatrice.app/Contents/MacOS/cockatrice
Undefined symbols for architecture x86_64:
  "UserConnection_Information::addNewServer(QString, QString, QString, QString, QString, bool)", referenced from:
      DlgConnect::DlgConnect(QWidget*) in dlg_connect.cpp.o
  "UserConnection_Information::getServerInfo(QString const&)", referenced from:
      DlgConnect::updateDisplayInfo(QString const&) in dlg_connect.cpp.o
  "UserConnection_Information::getServerInfo()", referenced from:
      DlgConnect::DlgConnect(QWidget*) in dlg_connect.cpp.o
  "UserConnection_Information::UserConnection_Information()", referenced from:
      DlgConnect::DlgConnect(QWidget*) in dlg_connect.cpp.o
      DlgConnect::updateDisplayInfo(QString const&) in dlg_connect.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

cockatrice/src/userconnection_information.cpp
```diff
@@ -0,0 +1,72 @@
+#include "userconnection_information.h"
```
ctrlaltca: You need to add this file to CMakelists.txt in this list: https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/CMakeLists.txt#L117 for it to be picked up and compiled.

cockatrice/src/userconnection_information.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+UserConnection_Information::UserConnection_Information()
+{
+}
+
+void UserConnection_Information::addNewServer(QString _saveName, QString _serv, QString _port, QString _username, QString _password, bool _savePassword)
+{
+    QSettings settings("Cockatrice", "server_details");
```
ctrlaltca: You may want to prefer serversettings instead of using QSettings directly.
Server settings are currently managed here: https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/settings/serverssettings.h , just adapt it to your needs

cockatrice/src/userconnection_information.cpp
```diff
@@ -0,0 +1,72 @@
...
+        
+        serverList.insert(saveName, userInfo);  
+    }
+
+    return serverList;
+}
+
+QStringList UserConnection_Information::getServerInfo(const QString &find)
+{
+    QStringList server = new QMap<QString, UserConnection_Information>();
```
ctrlaltca: The declared types don't match: a `QStringList` is a `QList<QString>`, not a `<QMap>`;
Also, when you call "new" you are allocating a new object to memory, so the result is a *pointer

cockatrice/src/userconnection_information.h
```diff
@@ -0,0 +1,32 @@
+#ifndef USERCONNECTION_INFORMATION_H
+#define USERCONNECTION_INFORMATION_H
+
+#include <QSettings>
+#include <QFile>
+#include <QDir>
+#include <QDebug>
+#include <QApplication>
+#include <QStandardPaths>
```
ctrlaltca: Always prefer to add `#include`s to .cpp instead of .h

cockatrice/src/dlg_connect.cpp
```diff
@@ -22,15 +23,25 @@ DlgConnect::DlgConnect(QWidget *parent)
...
+    connect(previousHosts, SIGNAL(currentIndexChanged(const QString &)), this, SLOT(updateDisplayInfo(const QString &)));
+    
+    UserConnection_Information uci;
+    QMap<QString, UserConnection_Information> savedHostList = uci.getServerInfo();
+    
+    if (savedHostList.isEmpty())
+    {
+        uci.addNewServer("Woogerworks Cockatrice", "cockatrice.woogerworks.com", "4747", "", "", false);
+        uci.addNewServer("Chickatrice Cockatrice", "chickatrice.net", "4747", "", "", false);
+        uci.addNewServer("Cockatric.es Cockatrice", " cockatric.es", "4747", "", "", false);
```
ctrlaltca: minor: trim space in server name

ZeldaZach:
This PR is now in a state that it can be re-reviewed. cc @Daenyth @ctrlaltca

cockatrice/src/userconnection_information.h
```diff
@@ -0,0 +1,32 @@
...
+        QString saveName;
+        QString server;
+        QString port;
+        QString username;
+        QString password;
+        bool savePassword;
+    
+    public:
+        UserConnection_Information();
+		UserConnection_Information(QString, QString, QString, QString, QString, bool);
```
Daenyth: Tabs :P

ZeldaZach:
@Daenyth Updated PR status

ZeldaZach:
Do you all think we need the "Save Server Details" as it auto does that on connect? Also, should I clear the fields if a user goes from pre-defined to new host? Those were the last 2 questions I had on this issue.

Daenyth:
Are we punting on the UI redesign for now? https://github.com/Cockatrice/Cockatrice/issues/742#issuecomment-268402235

If this makes foundational changes that will make the above easier, and we plan to do the above after this is merged, then (once I check code), that's fine to ship.

If this doesn't move us closer to #742 we should re-examine and figure out which changes help (the per-server settings absolutely) and which don't.

Based on my last read I think this should help, I'll see with the current version

cockatrice/src/dlg_connect.cpp
```diff
@@ -132,24 +132,85 @@ DlgConnect::DlgConnect(QWidget *parent)
...
         previousHosts->setDisabled(false);
     }
 }
 
+void DlgConnect::updateDisplayInfo(const QString &saveName)
+{
+    UserConnection_Information uci;
+    QStringList data = uci.getServerInfo(saveName);
+
+    if (saveEdit == NULL)
```
Daenyth: Prefer `nullptr` over `NULL` in the future. This is fine for now though, not worth triggering CI again for it

ZeldaZach:
I feel this brings us closer to a UI redesign, but it will not be completed in this PR and will have to be looked at again in the future

ZeldaZach:
:shipit:

PR #2455, opened 2017-03-10
create related token if you press the 'Token:' button

### LINKED ISSUE #2453: Tokens created through reverse-related interface are not created again with "Create another token"
<b>OS:</b>
*Windows 7 SP1*

<b>Cockatrice version:</b>
*dab7316 (2017-01-19)*

___
<br>

The "Create another token" cannot create tokens created through the reverse-related token production menu item. Two possible reproduce cases:

Greyed out menu item:

1. Place a card with reverse-related token(s) in play (Voice of Resurgence used in this example)
2. Create one of its tokens (create an Elemental)
3. Right-click the battlefield; the "Create another token" menu item is greyed out&mdash;*expected behavior is for this menu item to be available*

Incorect token produced:

1. Create any token (Thopter used in this example)
2. Place a card with reverse-related token(s) in play (Voice of resurgence used in this example
3. Create one of its tokens (create an Elemental)
4. Right-click the battlefield and click "Create another token". This creates a Thopter token instead of an Elemental token. *expected behavior is for the created token to be a second Elemental token*

"Create another token" bug demonstration

ZeldaZach:
Fix #2453 

This will set the create additional token button if you use the "Token: TokenName" feature.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1103,6 +1103,19 @@ void Player::actCreateRelatedCard()
     QAction *action = static_cast<QAction *>(sender());
     const QString &actionDisplayName = action->text();
     createCard(sourceCard, dbNameFromTokenDisplayName(actionDisplayName));
+
+   /*
+    * If we made a token via "Token: TokenName"
+    * then lets allow it to be created via create another
+    */
+    CardInfo *cardInfo = db->getCard(dbNameFromTokenDisplayName(actionDisplayName));
+    lastTokenName = cardInfo->getName();
+    lastTokenColor = cardInfo->getColors().isEmpty() ? QString() : cardInfo->getColors().first().toLower();
+    lastTokenPT = cardInfo->getPowTough();
+    lastTokenAnnotation = settingsCache->getAnnotateTokens() ? cardInfo->getText() : "";
+    lastTokenTableRow = table->clampValidTableRow(2 - cardInfo->getTableRow());
+    lastTokenDestroy = true;
+    aCreateAnotherToken->setEnabled(true);
 }
 
 void Player::actCreateAllRelatedCards()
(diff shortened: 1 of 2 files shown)


Nightfirecat:
So does this only work if you use the "Create all related cards" when there is only one related card? Seems like `Player::actCreateAllRelatedCards()` should also inherit this behavior, unless I'm missing something.

ZeldaZach:
This will operate if you press either the "Token: TokenName" button _OR_ there is only one related card

Nightfirecat:
It looks like it misses the case of choosing "Token: TokenName" when there are multiple related cards? There's no action taken in `actCreateRelatedCard()`, [which is triggered in this scenario](https://github.com/Cockatrice/Cockatrice/blob/b2f2e9db8aa964fcf73a437d857f3e4badcd277a/cockatrice/src/player.cpp#L2448)

ZeldaZach:
Nice spotting @Nightfirecat I made another change. Care to try it out and lmk if it works as expected now?

Nightfirecat:
I'll try to wrestle through the Windows compilation steps again this weekend and see if I can't give this a test. At a glance, this looks like it should do the job.

On code style, this may be a good candidate for a function, given the mostly-identical duplicated code? Could be useful in the token dialog as well.

```
void setLastToken(CardItem *cardInfo)
{
    lastTokenName = cardInfo->getName();
    lastTokenColor = cardInfo->getColors().isEmpty() ? QString() : cardInfo->getColors().first().toLower();
    lastTokenPT = cardInfo->getPowTough();
    lastTokenAnnotation = settingsCache->getAnnotateTokens() ? cardInfo->getText() : "";
    lastTokenTableRow = table->clampValidTableRow(2 - cardInfo->getTableRow());
    lastTokenDestroy = true;
    aCreateAnotherToken->setEnabled(true);
}
```

ZeldaZach:
@Nightfirecat Changed that up

tooomm:
>I'll try to wrestle through the Windows compilation steps again this weekend and see if I can't give this a test. 

@Nightfirecat Why don't you just simply use the Appveyor CI build linked on the buttom of the PR? :)

Nightfirecat:
That sure would make things easier, wouldn't it? :arrow_up: 

Will get to testing it tonight or tomorrow.

Nightfirecat:
Looking good! :shipit: unless there's other code concerns.

fixed last-token logic

PR #2457, opened 2017-03-11
[WIP] Trying to change Host<->Creator

### LINKED ISSUE #2108: Update `Creatore`

ZeldaZach:
Fix #2108 

Wip on this, as I'm not sure how the protocol network works exactly.

### DIFF
--- a/common/pb/serverinfo_game.proto
+++ b/common/pb/serverinfo_game.proto
@@ -16,6 +16,7 @@ message ServerInfo_Game {
     optional bool spectators_need_password = 12;
     optional bool spectators_can_chat = 13;
     optional bool spectators_omniscient = 14;
+	optional ServerInfo_User host_info = 15;
     optional uint32 player_count = 30;
     optional uint32 spectators_count = 31;
     optional bool started = 50;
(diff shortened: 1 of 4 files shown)


ctrlaltca:
Looks like it currently works like this:
When a player joins a game, if his name matches the one of the room creator, he becomes the "host" of the game (https://github.com/Cockatrice/Cockatrice/blob/master/common/server_game.cpp#L452)
When a player is removed from the game (because he parts or gets kicked out), if the player is the current "host" then the first non-spectator available becomes the new "host" (https://github.com/Cockatrice/Cockatrice/blob/master/common/server_game.cpp#L495)
These changes are already sent to the game players using the `Event_GameHostChanged` event.
Cockatrice only uses this information to keep track of the current game "host" and enable the "kick user" menu if we're the host: https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/user_context_menu.cpp#L281
So, if you're inside the game lobby, you already get the information.

A different problem is the list of games in the room, as it currently only contains the creator info.
Adding a new host_info field in the ServerInfo_Game message as you did is correct, but you need to add that field also to the `Server_Game` class when it gets created (just duplicate how `creatorInfo` is handled), update it when the host changes (see the two links to server_game.cpp above) and then emit the `gameInfoChanged` signal to propagate this information to the clients.

common/pb/serverinfo_game.proto
```diff
@@ -16,6 +16,7 @@ message ServerInfo_Game {
     optional bool spectators_need_password = 12;
     optional bool spectators_can_chat = 13;
     optional bool spectators_omniscient = 14;
+	optional ServerInfo_User host_info = 15;
```
Daenyth: taaaabs

PR #2461, opened 2017-03-12
Cmake support for openssl 1.1 on windows; fix #2424

### LINKED ISSUE #2424: CMake can't find openssl 1.1 on windows
So far it looks like when trying to compile the application on windows using the current instructions the newest version of the OpenSSL package can not be found by CMAKE as well as the current location for the mysql DB library files are becoming stale.  The locations could use some updating.

ctrlaltca:
## Related Ticket(s)
- Fixes #2424

See #2424; In openssl 1.1 the library names changed, so our custom cmake script used to find and bundle them in the installer is not able to find them anymore.

Support for the new library names has been added.
Please note that Qt still doesn't support openssl 1.1 yet, so it's not usable anyway.

## Screenshots
N/a

### DIFF
--- a/cmake/FindWin32SslRuntime.cmake
+++ b/cmake/FindWin32SslRuntime.cmake
@@ -48,9 +48,10 @@ set(_OPENSSL_ROOT_HINTS_AND_PATHS
     PATHS ${_OPENSSL_ROOT_PATHS}
     )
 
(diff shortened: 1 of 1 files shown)


tooomm:
+1 for using the template! ❤️

PR #2462, opened 2017-03-12
rephrase missing feature setting

tooomm:
## Related Ticket(s)
- https://github.com/Cockatrice/Cockatrice/pull/2362#pullrequestreview-16717035

Bad and confusing label in settings
This change got missed in the initial PR.

Better description of what the setting changes/influences


ctrlaltca:
I remember discussing this, but i probably missed to include it in the PR.

PR #2464, opened 2017-03-12
Added privileged level only room

### LINKED ISSUE #2463: Allow Server Rooms to be based on user permissionLevels
The idea is we currently can set a room to be "open for all", "moderators only", "admin only".

We'd like to add a way to make a room "VIP only", "Donator only", "VIP/Donator only". These rooms would be an interesting addition. These rooms would be accessible by VIPs and/or Donators & Moderators/Admins of the server.

Areas to look at `tab_server.*`, Protocol

woogerboy21:
Fix #2463 

Added the ability for registered user - VIP/DONATOR/PRIVILEGED room.

Since the users privilege level does not have any type of tier concept to it, we can only compare the actual literal string value of the privilege level and check to see if a user is a part of the given privilege level to allow or deny them into a given room.   

As explained when the privilege level concept was being introduced, we have admin level which is a bit comparison to allow the concept of admin > mod > reg > non-reg, but privilege level doesn't work like that.  Its a sub category.   So something like VIP !> DONATOR and vice versa.

With the logic in this PR a DONATOR for example can not enter a VIP room, nor can a VIP enter a DONATOR room, but we do compare admin level value(s) so regardless of admin/mod privilege levels they can enter the rooms. How ever an unregistered user is not allowed to enter the room if defined with the "vip" , "donator" or "privileged"  database table privlevel"column value.

Clear as mud?  

This PR now adds a "privlevel" to the room so we can do easy comparisons between a rooms definition of "privilege" to a users definition of "privilege" and allow/deny access based on the combinations of definitions rather than a hard written code block for every single privlevel added to the db table.  This way new privilege levels can be added into the database columns for the enum value(s) in the users/rooms table columns and their should be no additional coding required to restrict room access based on the new enum definitions.

### DIFF
--- a/cockatrice/src/localserver.cpp
+++ b/cockatrice/src/localserver.cpp
@@ -6,7 +6,7 @@ LocalServer::LocalServer(QObject *parent)
     : Server(parent)
 {
     setDatabaseInterface(new LocalServer_DatabaseInterface(this));
-    addRoom(new Server_Room(0, 0, QString(), QString(), QString(), false, QString(), QStringList(), this));
+    addRoom(new Server_Room(0, 0, QString(), QString(), QString(), QString(), false, QString(), QStringList(), this));
 }
 
 LocalServer::~LocalServer()
--- a/cockatrice/src/tab_server.cpp
+++ b/cockatrice/src/tab_server.cpp
@@ -71,11 +71,18 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
     const int roomListSize = event.room_list_size();
     for (int i = 0; i < roomListSize; ++i) {
         const ServerInfo_Room &room = event.room_list(i);
+
+        /*
+        * A room can have a permission level and a privilege level. How ever we want to display only the necessary information 
+        * on the server tab needed to inform users of required permissions to enter a room. If the room has a privilege level
+        * the server tab will display the privilege level in the "permissions" column in the row however if the room contains
+        * a permissions level for the room the permissions level defined for the room will be displayed.
+        */
+
         QString roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();
-        if (QString::fromStdString(room.permissionlevel()).toLower() != "none" && QString::fromStdString(room.privilegelevel()).toLower() == "none")
+        if (QString::fromStdString(room.permissionlevel()).toLower() != "none" || roomPermissionDisplay == "")
             roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
-        if (roomPermissionDisplay == "")
-            roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
+
         for (int j = 0; j < roomList->topLevelItemCount(); ++j) {
               QTreeWidgetItem *twi = roomList->topLevelItem(j);
             if (twi->data(0, Qt::UserRole).toInt() == room.room_id()) {
--- a/cockatrice/src/tab_server.cpp
+++ b/cockatrice/src/tab_server.cpp
@@ -74,7 +74,8 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
         QString roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();
         if (QString::fromStdString(room.permissionlevel()).toLower() != "none" && QString::fromStdString(room.privilegelevel()).toLower() == "none")
             roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
-       
+        if (roomPermissionDisplay == "")
+            roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
         for (int j = 0; j < roomList->topLevelItemCount(); ++j) {
               QTreeWidgetItem *twi = roomList->topLevelItem(j);
             if (twi->data(0, Qt::UserRole).toInt() == room.room_id()) {
--- a/cockatrice/src/tab_server.cpp
+++ b/cockatrice/src/tab_server.cpp
@@ -80,8 +80,10 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
         */
 
         QString roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();
-        if (QString::fromStdString(room.permissionlevel()).toLower() != "none" || roomPermissionDisplay == "")
+        if (QString::fromStdString(room.permissionlevel()).toLower() != "none")
             roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
+        if (roomPermissionDisplay == "")    // catch all for misconfigured .ini room definitions 
+            roomPermissionDisplay = "none";
 
         for (int j = 0; j < roomList->topLevelItemCount(); ++j) {
               QTreeWidgetItem *twi = roomList->topLevelItem(j);
--- a/cockatrice/src/tab_server.cpp
+++ b/cockatrice/src/tab_server.cpp
@@ -71,7 +71,10 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
     const int roomListSize = event.room_list_size();
     for (int i = 0; i < roomListSize; ++i) {
         const ServerInfo_Room &room = event.room_list(i);
-        
+        QString roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();
+        if (QString::fromStdString(room.permissionlevel()).toLower() != "none" && QString::fromStdString(room.privilegelevel()).toLower() == "none")
+            roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
+       
         for (int j = 0; j < roomList->topLevelItemCount(); ++j) {
               QTreeWidgetItem *twi = roomList->topLevelItem(j);
             if (twi->data(0, Qt::UserRole).toInt() == room.room_id()) {
@@ -80,7 +83,7 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
                 if (room.has_description())
                     twi->setData(1, Qt::DisplayRole, QString::fromStdString(room.description()));
                 if (room.has_permissionlevel())
-                    twi->setData(2, Qt::DisplayRole, QString::fromStdString(room.permissionlevel()).toLower());
+                    twi->setData(2, Qt::DisplayRole, roomPermissionDisplay);
                 if (room.has_player_count())
                     twi->setData(3, Qt::DisplayRole, room.player_count());
                 if (room.has_game_count())
@@ -95,7 +98,7 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
         if (room.has_description())
             twi->setData(1, Qt::DisplayRole, QString::fromStdString(room.description()));
         if (room.has_permissionlevel())
-            twi->setData(2, Qt::DisplayRole, QString::fromStdString(room.permissionlevel()).toLower());
+            twi->setData(2, Qt::DisplayRole, roomPermissionDisplay);
         twi->setData(3, Qt::DisplayRole, room.player_count());
         twi->setData(4, Qt::DisplayRole, room.game_count());
         twi->setTextAlignment(2, Qt::AlignRight);
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -637,6 +637,8 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
                 return Response::RespUserLevelTooLow;
         }
 
+        // TODO: POSSIBLY EXTEND THE ROOM OBJECT TO HAVE PRIV LEVEL SETTINGS AND COMPARE TO USER PRIVLEVEL ?
+
         if (roomPermission == "vip") {
             if (!(userInfo->user_level() & ServerInfo_User::IsRegistered))
                 return Response::RespUserLevelTooLow;
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -645,6 +645,15 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
                 if (QString::fromStdString(userInfo->privlevel()).toLower() != "vip")
                     return Response::RespUserLevelTooLow;
         }
+
+        if (roomPermission == "donator") {
+            if (!(userInfo->user_level() & ServerInfo_User::IsRegistered))
+                return Response::RespUserLevelTooLow;
+
+            if (!((userInfo->user_level() & ServerInfo_User::IsAdmin) && (userInfo->user_level() & ServerInfo_User::IsModerator)))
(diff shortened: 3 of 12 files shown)


woogerboy21:
Is this the approach we should take or should we also add a value to the rooms to co-inside with the userinfo container for privlevel and do comparison's accordingly?  this solution will work if you have no interest in doing something like an admin-mod/vip-donator only room.  But if some one for-see's the requirement of having such a restriction then we should probably extend the room container.

ZeldaZach:
Just a thought on this; maybe make a way for a room to be for both VIP/Donators too?

common/server_protocolhandler.cpp
```diff
@@ -627,16 +627,36 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
...
         }
-    }
+
+        // TODO: POSSIBLY EXTEND THE ROOM OBJECT TO HAVE PRIV LEVEL SETTINGS AND COMPARE TO USER PRIVLEVEL ?
+
+        if (roomPermission == "vip") {
+            if (!(userInfo->user_level() & ServerInfo_User::IsRegistered))
+                return Response::RespUserLevelTooLow;
+            
+            if (!((userInfo->user_level() & ServerInfo_User::IsAdmin) && (userInfo->user_level() & ServerInfo_User::IsModerator)))
```
ZeldaZach: Shouldn't this be `||` instead? `If ( ! (isAdmin || isModerator) )`

woogerboy21: That's what it is doing... 😕

woogerboy21:
@ZeldaZach Should be all fixed up now

tritoch:
Any way to make this an auto-join for qualified users (without having disqualified users get a message saying unable to join)?

ZeldaZach:
Hmm, the idea of auto joining all rooms you're eligible for... That's an idea, but it can get messy if a mod or smth logs in and joins like 5 rooms automatically...

woogerboy21:
> Any way to make this an auto-join for qualified users (without having disqualified users get a message saying unable to join)?

Currently no.  Not even sure how we would approach a concept like this.

tritoch:
>Hmm, the idea of auto joining all rooms you're eligible for... That's an idea, but it can get messy if a mod or smth logs in and joins like 5 rooms automatically...

Doesn't even have to be all rooms.  There's already a per-room toggle in the `ini`.  But currently if you set it on a moderator room (the only privileged room available) all non-mod/admins will get a message about being unable to join.

woogerboy21:
@tritoch 
Correct, I know how the current behavior acts.  Not sure how we would even begin to change the way the code works to not show this on first login OR change the code logic completely to avoid even trying to join the rooms at login.

Daenyth:
> But currently if you set it on a moderator room (the only privileged room available) all non-mod/admins will get a message about being unable to join.

Bug - it shouldn't attempt to join rooms they don't have privileges for

Daenyth:
Before I dive into the code I wanted to talk about expected behavior..

Is it useful to distinguish between vip and donator? I would think we want the ability to create a room that either vip or donators could join, rather than only one or the other. Does the PR allow this?

woogerboy21:
> Bug - it shouldn't attempt to join rooms they don't have privileges for

This isnt a bug, what they are proposing is to make users attempt to join all the rooms they have privileges to and the only way to do so right now is set them as autojoin and if they are set this way with the current code base they get the message for the rooms they do not have permissions to.  The code base has not been modified (even in this PR) to have any intelligence on having a client not attempt to join a room which they dont have privileges to.

woogerboy21:
> Is it useful to distinguish between vip and donator? I would think we want the ability to create a room that either vip or donators could join, rather than only one or the other. Does the PR allow this?

This PR adds code to allow you to create a room that both can join, or either can join (privileged [both] OR vip [only] OR donator [only]) providing the most flexibility.  But to your point is this wanted/expected behavior and/or is the approach in the code base the way you want it taken?  We can either do something like the way its written now to accomplish the current base,  OR extend the room object to have a "privilege" setting and allow for easier expansion later on.  I've never really grasped the groups thought process on "crossing that bridge later" when its come up.

Daenyth:
I wasn't saying it's a bug with the PR, I was saying it's a bug with the existing implementation before your changes. You don't need to fix it in this PR

Daenyth:
> This PR adds code to allow you to create a room that both can join, or either can join (privileged [both] OR vip [only] OR donator [only]) providing the most flexibility. 

This sounds good 👍

common/server_protocolhandler.cpp
```diff
@@ -627,16 +627,45 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
...
+
+        if (roomPermission == "donator") {
+            if (!(userInfo->user_level() & ServerInfo_User::IsRegistered))
+                return Response::RespUserLevelTooLow;
+
+            if (!((userInfo->user_level() & ServerInfo_User::IsAdmin) || (userInfo->user_level() & ServerInfo_User::IsModerator)))
+                if (QString::fromStdString(userInfo->privlevel()).toLower() != "donator")
+                    return Response::RespUserLevelTooLow;
+        }
+}
```
Daenyth: Wrong indent - this should be aligned with the opening `if`

common/server_protocolhandler.cpp
```diff
@@ -627,16 +627,45 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
                 return Response::RespUserLevelTooLow;
         }
 
-        if (roomPermission == "moderator"){
+        if (roomPermission == "moderator") {
```
Daenyth: Can you encapsulate every roomPermission check into its own function?

```cpp
bool userIsPermittedToJoin(QString roomPermission, ServerInfo_User userInfo)
```

Then inside this function we can just
```cpp
if (!userIfPermittedToJoin(roomPermission, userInfo)
    return Response::RespUserLevelTooLow

```

If possible I think that function could be static - this would make it easier to unit test later

woogerboy21:
Over the next little bit I am going to change the way the code logic works for the permission/privilege level blocks is written.  I think we should do this in a different direction so it allows for easier expansion of both permission & privilege levels .  Hopefully won't take to long.

servatrice/migrations/servatrice_0022_to_0023.sql
```diff
@@ -0,0 +1,5 @@
+-- Servatrice db migration from version 22 to version 23
+
+alter table cockatrice_rooms add column privlevel enum("NONE","VIP","DONATOR") NOT NULL;
```
ZeldaZach: Might want "PRIVILEGED" too

servatrice/servatrice.sql
```diff
@@ -90,6 +90,7 @@ CREATE TABLE IF NOT EXISTS `cockatrice_rooms` (
   `name` varchar(50) NOT NULL,
   `descr` varchar(255) NOT NULL,
   `permissionlevel` varchar(20) NOT NULL,
+  `privlevel` enum("NONE","VIP","DONATOR") NOT NULL,
```
ZeldaZach: Might want "PRIVILEGED" too

servatrice/migrations/servatrice_0022_to_0023.sql
woogerboy21: I matched the wording for the column name in the users table in an attempt to keep confusion levels low.

servatrice/servatrice.sql
woogerboy21: I matched the wording for the column name in the users table in an attempt to keep confusion levels low.

woogerboy21:
@ZeldaZach  @Daenyth 
Now this PR is ready for review.  I may need to fix tab->spaces (at the very least)

ZeldaZach:
Unfortunately if it will show the incorrect message, (i.e. none when it's not actually none), I'll want a featureset added. Whether you choose to make it a comparison or not is up to you for login (making it required) but since we just released I'd be hesitant atm to do that.

common/server_protocolhandler.cpp
```diff
@@ -620,23 +620,9 @@ Response::ResponseCode Server_ProtocolHandler::cmdJoinRoom(const Command_JoinRoo
...
-            if (!(userInfo->user_level() & ServerInfo_User::IsModerator))
-                return Response::RespUserLevelTooLow;
-        }
-
-        if (roomPermission == "administrator"){
-            if (!(userInfo->user_level() & ServerInfo_User::IsAdmin))
-                return Response::RespUserLevelTooLow;
-        }
-    }
+    if (!(userInfo->user_level() & ServerInfo_User::IsModerator || userInfo->user_level() & ServerInfo_User::IsModerator))
```
ZeldaZach: This is a double moderator check. You probably want if (NOT ( isAdmin || isModerator ))

cockatrice/src/tab_server.cpp
```diff
@@ -71,7 +71,11 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
     const int roomListSize = event.room_list_size();
     for (int i = 0; i < roomListSize; ++i) {
         const ServerInfo_Room &room = event.room_list(i);
-        
+        QString roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();
```
ZeldaZach: I _really_ do not like how this block of code here looks. Please add a block comment explaining how this check fully operates. My suggestion on a rewrite (might not be as functional as yours...)
```
/*
 * A room can only have a permission level or a 
 * privilege level. If a room has both, then only
 * the permission level will be checked.
 */
QString roomPermissionDisplay;

if (QString::fromStdString(room.permissionlevel()).toLower() != "none")
	roomPermissionDisplay = QString::fromStdString(room.permissionlevel()).toLower();
else if (QString::fromStdString(room.privilegelevel()).toLower() != "none")
	roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();

```

cockatrice/src/localserver.cpp
```diff
@@ -6,7 +6,7 @@ LocalServer::LocalServer(QObject *parent)
     : Server(parent)
 {
     setDatabaseInterface(new LocalServer_DatabaseInterface(this));
-    addRoom(new Server_Room(0, 0, QString(), QString(), QString(), false, QString(), QStringList(), this));
+    addRoom(new Server_Room(0, 0, QString(), QString(), QString(), QString(), false, QString(), QStringList(), this));
```
ZeldaZach: is there a header file to update with such a change?

common/server_protocolhandler.cpp
```diff
@@ -765,3 +751,24 @@ void Server_ProtocolHandler::resetIdleTimer()
     lastActionReceived = timeRunning;
     idleClientWarningSent = false;
 }
+
+bool Server_ProtocolHandler::isUserPermittedToJoinRoom(const QString roomPermission, const QString roomPrivilegeLevel)
```
ZeldaZach: This function seems to counteract what's done above slightly. My comment above says that only permission OR roompriv is going to be set... How will a line look if I choose to set room isAdmin only && donator only for example? Should the room say `admin, donator` or something?

common/server_protocolhandler.cpp
woogerboy21: That is what its doing.  There is no double check.  See how the IsModertor || isAdmin is contained in () with an outer ! ().  Its sayi is not moderator or admin.

common/server_protocolhandler.cpp
woogerboy21: There is no such thing as an admin/moderators.  admin and moderators are allowed to enter all rooms regardless of their privilege level.

cockatrice/src/tab_server.cpp
woogerboy21: Yes, you can't get away with your typical if , followed by if, followed by if method you generally enjoy doing.  We want to avoid testing conditions based on combinations of permissions and privileges and avoid others when certain conditions are met.  I'll see about adding comments to make things more clear.

cockatrice/src/localserver.cpp
woogerboy21: No it does not.  Check out the localserver.h  I can't say I have dug into why this is for the local server.

woogerboy21:
@ZeldaZach Are there additional area's you would like comments added for clarity?

cockatrice/src/tab_server.cpp
```diff
@@ -71,7 +71,20 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
...
-        
+
+        /*
+        * A room can have a permission level and a privilege level. How ever we want to display only the necessary information 
+        * on the server tab needed to inform users of required permissions to enter a room. If the room has a privilege level
+        * the server tab will display the privilege level in the "permissions" column in the row however if the room contains
+        * a permissions level for the room the permissions level defined for the room will be displayed.
+        */
+
+        QString roomPermissionDisplay = QString::fromStdString(room.privilegelevel()).toLower();
```
ZeldaZach: So this will either return "none", "moderator", or "admin"?

common/server_protocolhandler.cpp
ZeldaZach: I think I understand

cockatrice/src/tab_server.cpp
woogerboy21: It will set the roomPermissionDisplay to what the privlevel definition of the room is (none,privileged,vip,donator at this time).  The next line then trumps it if there is a permission value defined.

common/server_protocolhandler.cpp
```diff
@@ -765,3 +751,24 @@ void Server_ProtocolHandler::resetIdleTimer()
...
+{
+    if (roomPermission.toLower() == "administrator" || roomPermission.toLower() == "moderator")
+        return false;
+
+    if (roomPermission.toLower() == "registered" && !(userInfo->user_level() & ServerInfo_User::IsRegistered))
+        return false;
+
+    if (roomPrivilegeLevel.toLower() != "none")
+    {
+        if (roomPrivilegeLevel.toLower() == "privileged")
```
ZeldaZach: Can you rework this block so it's more defined? The else seems a little out of place w/o brackets for example

woogerboy21: No

cockatrice/src/tab_server.cpp
ZeldaZach: Ahh ok, makes sense

common/server_protocolhandler.cpp
ZeldaZach: harsh man

cockatrice/src/tab_server.cpp
Daenyth: This could be refactored `const QString roomPermissionDisplay = getRoomPermissionDisplay(room)`

common/server_protocolhandler.cpp
Daenyth: Can you make this method take `userInfo` and mark it static so it's more easily tested with unit tests?

common/server_protocolhandler.cpp
woogerboy21: OK formatting changed...

woogerboy21:
Closing this as #2484 is a better way of doing things.  It will also work in our favor with getting auto join rooms on login working smoother.

PR #2468, opened 2017-03-13
Fix issues with new updater; fix #2465

### LINKED ISSUE #2465: several issues with new release channels
<b>OS:</b>
Win 10 64bit

<b>Cockatrice version:</b>
02bb226 (2017-03-12)

___
<br>

Feature initially introduced in https://github.com/Cockatrice/Cockatrice/pull/2362

## Issue 1:
Seems to be Windows only, since it seems to work for @ZeldaZach on macOS.
I realized that it's not possible to switch update channels. If you change the channel in settings and reopen the settings window it'll show the old value. The updater also fetches the old channel.
@woogerboy21 tracked it down more in detail - taken from gitter chat:
>no matter what i choose always seems to be set back to the value it originally was and i dont see any updates happening to the global config file
which always locks my client into the development branch

>if i manually change the value in the settings file the UI will reflect the change properly. its acting like the change in value in the UI isnt saving the config setting properly

>no matter what i set the value to in the settings dialogue it always checks for updates for the value in the settings file, which never changes even if i change the value in the settings ui under windows

<br>

## Issue 2:
In the info dialog for the `development snapshot` branch, the commit number is not shortened to the usual length:
development

<br>

## Issue 3:
Also related to that same info dialog, there is a link to the release notes of the last main release.
For development builds that makes not too much sense and a link to the commit on github, or even better the corresponding PR would be more helpful. Not sure what is possible here, maybe there are some other useful information that we could link to there instead...
So basically, we would need a different dialog message for each channel.

Actually, 2 and 3 are described in the PR but ended up different: https://github.com/Cockatrice/Cockatrice/pull/2362#issuecomment-273707379

If I read [this and the following comments](https://github.com/Cockatrice/Cockatrice/pull/2362#issuecomment-282577355) correctly, there is no date included and the updater just checks if the available commit for the selected channel is different from the installed one. 
Is there some way that prevents users from downgrading or otherwise installing a wrong/old/... version with the updater? Right now we are all dependent on bintray or the commits not beeing messed up?

<br>
CC: @ctrlaltca

ctrlaltca:
## Related Ticket(s)
- Fixes #2465

 - See #2465

 - Fixed release channel selector under windows (issue 1): Msvc initialized the items in the wrong order, so their indexes would mix up; fixed by adding items one by one in the list.
 - dev channel: show the short commit hash instead of the long one (issue 2);
 - dev channel: link to the compare between the old hash and the new hash instead of just the last commit (issue 3)
 - properly return when no update is found (thx @ZeldaZach)

## Screenshots
N/a

### DIFF
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -108,6 +108,7 @@ void DlgUpdate::finishedUpdateCheck(bool needToUpdate, bool isCompatible, Releas
         //If there's no need to update, tell them that. However we still allow them to run the
         //downloader themselves if there's a compatible build
         QMessageBox::information(this, tr("Cockatrice Update"), tr("Your version of Cockatrice is up to date."));
+        return;
     }
 
     if (isCompatible) {
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -111,10 +111,10 @@ void DlgUpdate::finishedUpdateCheck(bool needToUpdate, bool isCompatible, Releas
         return;
     }
 
+    publishDate = release->getPublishDate().toString(Qt::DefaultLocaleLongDate);
     if (isCompatible) {
         //If there is an update, save its URL and work out its name
         updateUrl = release->getDownloadUrl();
-        publishDate = release->getPublishDate().toString(Qt::DefaultLocaleLongDate);
 
         QMessageBox::StandardButton reply;
         reply = QMessageBox::question(this, "Update Available",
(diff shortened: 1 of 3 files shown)


ZeldaZach:
I'm happy with this :shipit:

ctrlaltca:
Fixed another minor issue: the "published on" date didn't show up when no package was available for your os/arch.

tooomm:
Here are the missing screens to properly get everybody on the same topic ;P

Release channel + new version available:
rel-new version

Development snapshop + no new version available:
dev-no new version

All mentioned fixes work as described. 👍 

I think we could improve the layout and display of the actual information. But I'll leave that for another ticket.

ctrlaltca:
Travis is experiencing problems, the osx builds are stuck.
This has been compiled and tested under osx by both m and @ZeldaZach , i assume it is safe to merge.

PR #2472, opened 2017-03-14
make settings size vary based on sizeHint()

### LINKED ISSUE #2471: Problem with preferences window resizing
<b>OS: macOS 10.12.2</b>

<b>Cockatrice version: 6e723b2 (2017-03-14)</b>

After opening the preferences window it's already oversized and goes beneath the screen boundaries. I'm not able to resize the height of the window when I mouse over the edge of it. I can move the window left or right, but when I try to lower it when the left click is released it jumps back up to the top of the screen.
___
<br>

screen shot 2017-03-14 at 9 31 28 am

ZeldaZach:
## Related Ticket(s)
- Fixes #2471 

The dialogs can sometimes be too large to handle on a screen

This adds a size minimum and calls the dialog to adjustSize() in order for it to fit the screen better. The size cannot go above 2/3'rds the size of the screen by default according to the function, so this may be helpful.

Downside is sometimes it's "smushed" and you'll have to manually readjust the size in order for you to see everything correctly... Suggestions can be taken! 

This is an initial test of just the settings dialog as that seems to be the main issue, but it can be expanded to any and all dialogs in the program.

## Screenshots
*(simply drag & drop image files directly into this description!)*

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -752,6 +752,9 @@ void SoundSettingsPage::retranslateUi() {
 DlgSettings::DlgSettings(QWidget *parent)
     : QDialog(parent)
 {
+    this->setMinimumSize(500,500);
+    this->adjustSize();
+
     connect(settingsCache, SIGNAL(langChanged()), this, SLOT(updateLanguage()));
     
     contentsWidget = new QListWidget;


tooomm:
Did somebody test this? :)
Introduces huge problems for me:
Running a8a4557 (2017-03-16) on Win 10 with 1366x768

PR #2480, opened 2017-03-14
fix resize

### LINKED ISSUE #1623: Bigger or resizable create token dialog
The create token dialog is too small and not resizable. It's a pain to search for tokens.

ZeldaZach:
## Related Ticket(s)
- Fixes #1623

Dialog size might be too small

This will allow the dialog to expand to a reasonable size.

## Screenshots
*(simply drag & drop image files directly into this description!)*


Daenyth:
If the size ends up being 1,1 can you actually interact with it? Maybe 100x100?

ZeldaZach:
Changed @Daenyth

tooomm:
Did somebody test this? :)
Introduces huge problems for me:
Running a8a4557 (2017-03-16) on Win 10 with 1366x768

PR #2484, opened 2017-03-15
Smarter room

### LINKED ISSUE #2463: Allow Server Rooms to be based on user permissionLevels

woogerboy21:
Fix #2463

Added the ability for registered user - VIP/DONATOR/PRIVILEGED room.

Since the users privilege level does not have any type of tier concept to it, we can only compare the actual literal string value of the privilege level and check to see if a user is a part of the given privilege level to allow or deny them into a given room.

As explained when the privilege level concept was being introduced, we have admin level which is a bit comparison to allow the concept of admin > mod > reg > non-reg, but privilege level doesn't work like that. Its a sub category. So something like VIP !> DONATOR and vice versa.

With the logic in this PR a DONATOR for example can not enter a VIP room, nor can a VIP enter a DONATOR room, but we do compare admin level value(s) so regardless of admin/mod privilege levels they can enter the rooms. How ever an unregistered user is not allowed to enter the room if defined with the "vip" , "donator" or "privileged" database table privlevel"column value.

Clear as mud?

This PR adds a "privlevel" to the room so we can do easy comparisons between a rooms definition of "privilege" to a users definition of "privilege" and allow/deny access based on the combinations of definitions rather than a hard written code block for every single privlevel added to the db table. This way new privilege levels can be added into the database columns for the enum value(s) in the users/rooms table columns and their should be no additional coding required to restrict room access based on the new enum definitions.


common/server_room.cpp
```diff
@@ -36,6 +36,37 @@ Server_Room::~Server_Room()
     usersLock.unlock();
 }
 
+bool Server_Room::userMayJoin(const ServerInfo_User & userInfo)
+{
+    qDebug() << "USERPRIV: " << QString::fromStdString(userInfo.privlevel());
+
+    ServerInfo_Room roomInfo;
```
ZeldaZach: You create a roomInfo object, but never use it in the function, nor call getRoomPermissionDisplay()

servatrice/migrations/servatrice_0022_to_0023.sql
```diff
@@ -0,0 +1,6 @@
+-- Servatrice db migration from version 22 to version 23
+
+alter table cockatrice_rooms modify column permissionlevel enum('NONE','REGISTERED','MODERATOR','ADMINISTRATOR');
```
ZeldaZach: `"` over `'` since the other line uses `"`? 🥇

common/server_room.cpp
Daenyth: You wouldn't call get..Display here, but this line doesn't need to be here, you're right

servatrice/migrations/servatrice_0022_to_0023.sql
Daenyth: For sql you use `'` for strings; `"` is used to quote table and column names

woogerboy21: Yeah they should be ' , not sure how one line came out one way and the other another.  I will fix this.

common/server_room.cpp
woogerboy21: Whoops left what looks like a debug line in, I'll 🔪 it.  As for the function, its called in the tab_server.cpp file to pull the room permissions to display in the table on the tab.  🍝

woogerboy21:
Should be all fixed up now.  I'll see about throwing test box up for you guys to play with tomorrow.  Heading to 🛏

common/server_room.cpp
```diff
@@ -36,6 +36,33 @@ Server_Room::~Server_Room()
     usersLock.unlock();
 }
 
+bool Server_Room::userMayJoin(const ServerInfo_User & userInfo)
+{
+    
+    ServerInfo_Room roomInfo;
```
ZeldaZach: Is there a reason this is here if we never use it? Or do you use it implicitly and I didn't notice?

woogerboy21: I swear I just answered this.... its used on line 624 of the protocol handler.
https://github.com/Cockatrice/Cockatrice/pull/2484/files#diff-d3b638a03e1222b79c2b7b5b8935096fR624

Daenyth: But that's a different function and this one isn't in scope there. Also it's a different type, this one is Server_Room and the other is ServerInfo_Room. I think you're mixed up here

woogerboy21: Oh you mean the declaration of the roomInfo on line 42?

Daenyth: Yes, line 42 :)

woogerboy21:
Did you manage to get a test box up and play with it?  I had planned on doing so for you to try but life jumped in the way today.
On 3/16/2017 10:10:41 PM, Zach H <notifications@github.com> wrote:
@ZeldaZach approved this pull request.

ZeldaZach:
I approved the code based on syntax and how it appears to function, i didn't get a chance to test it on a box. Will want to test before merging ofc

woogerboy21:
@ZeldaZach 
I finally had time to stand up a box with this PR rolled in.  People can test things out by using the host name of testing.woogerworks.com and port 4747 .  You should be able to use the username/password you have on the main woogerworks server.  I don't recall when the users table was pulled in on the testing server so if you recently changed your password the testing servers password might be your previous password.

ZeldaZach:
Tested and this seems to function as it should. :shipit:

cockatrice/src/tab_server.cpp
```diff
@@ -109,6 +109,24 @@ void RoomSelector::processListRoomsEvent(const Event_ListRooms &event)
...
 }
 
+QString RoomSelector::getRoomPermissionDisplay(const ServerInfo_Room & room)
+{
+    /*
+    * A room can have a permission level and a privilege level. How ever we want to display only the necessary information
+    * on the server tab needed to inform users of required permissions to enter a room. If the room has a privilege level
+    * the server tab will display the privilege level in the "permissions" column in the row however if the room contains
+    * a permissions level for the room the permissions level defined for the room will be displayed.
+    */
```
Daenyth: great comment

PR #2427, opened 2017-02-26
Removed font-size cap for on-card text (fixes bug: #2368)

Steven--:
Fix #2368

PR #2488, opened 2017-03-17
allow max font size to be defined by user

ZeldaZach:
## Related Ticket(s)
Reverts #2427 to make it better

Font size had a cap in the past, but then we removed it. Some users had fonts WAY too large. 

Users can now define their own font size caps. This will allow users who want smaller fonts to have such an option

## Screenshots
screenshot 2017-03-16 22 06 19
screenshot 2017-03-16 22 06 31
screenshot 2017-03-16 22 06 39

### DIFF
--- a/cockatrice/src/abstractcarditem.cpp
+++ b/cockatrice/src/abstractcarditem.cpp
@@ -68,7 +68,7 @@ QSizeF AbstractCardItem::getTranslatedSize(QPainter *painter) const
 void AbstractCardItem::transformPainter(QPainter *painter, const QSizeF &translatedSize, int angle)
 {
     const int MAX_FONT_SIZE = settingsCache->getMaxFontSize();
-    const int fontSize = std::max(9, std::min<int>(MAX_FONT_SIZE, round(translatedSize.height() / 8)));
+    const int fontSize = std::max(9, MAX_FONT_SIZE);
 
     QRectF totalBoundingRect = painter->combinedTransform().mapRect(boundingRect());
     
--- a/cockatrice/src/abstractcarditem.cpp
+++ b/cockatrice/src/abstractcarditem.cpp
@@ -3,6 +3,7 @@
 #include <QCursor>
 #include <QGraphicsSceneMouseEvent>
 #include <cmath>
+#include <algorithm>
 #ifdef _WIN32
 #include "round.h"
 #endif /* _WIN32 */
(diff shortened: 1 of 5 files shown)


cockatrice/src/abstractcarditem.cpp
```diff
@@ -80,6 +82,9 @@ void AbstractCardItem::transformPainter(QPainter *painter, const QSizeF &transla
     int fontSize = round(translatedSize.height() / 8);
     if (fontSize < 9)
         fontSize = 9;
+    if (fontSize > MAX_FONT_SIZE)
+        fontSize = MAX_FONT_SIZE;
```
Daenyth: You can consolidate all these lines with

```cpp
const int fontSize = std::max(9, std::min(MAX_FONT_SIZE, round(translatedSize.height() / 8)));
```

The reason to prefer the one-liner is because you can make the int const and the rest of the function can't accidentally modify it. By making things immutable and declarative it's easier to reason about at a larger scope.

cockatrice/src/dlg_settings.cpp
```diff
@@ -392,6 +399,7 @@ void AppearanceSettingsPage::retranslateUi()
     tableGroupBox->setTitle(tr("Table grid layout"));
     invertVerticalCoordinateCheckBox.setText(tr("Invert vertical coordinate"));
     minPlayersForMultiColumnLayoutLabel.setText(tr("Minimum player count for multi-column layout:"));
+    maxFontSizeForCardsLabel.setText(tr("Maximum size font for displaying card attributes"));
```
tooomm: Isn't it `font size` ?!

I would also consider adding `on the table`, or maybe better: rephrase to `information displayed on cards`.

ZeldaZach: Changed to `Maximum font size for information displayed on cards`

ZeldaZach:
@tooomm Good?

tooomm:
Working! 👍 
BUT, the initial value in settings is the max value of 100...

Notes:
- changes to card text just updates once you click on a card or move them around
- for me, the max font size is at around 20, higher values don't do anything (might be different on high res screens)

ZeldaZach:
There is an upper bound set internally by the system, I could remove that check and leave it solely on the user

ZeldaZach:
Removed system limits. Font size can now be [9-100]. Looks good and works fine. Plan on merging shortly.

ZeldaZach:
I can't find a way to globally update all cards, so i'll look into that in a future PR.

PR #2500, opened 2017-03-18
file filenames so window updater works

ZeldaZach:
Full release files end in arch name (win64.exe or win32.exe) while debug builds end in a QT string (-x86_64_qt5.exe or -x86_qt5.exe)

This fixes the auto-updater on windows for development builds


tooomm:
`debugEnd` is a weird and confusing wording... `devSnapshotEnd` or something instead?

ZeldaZach:
@tooomm meant to use `devSnapshotEnd` as name, forgot it wasn't debugging. Changed.

PR #2503, opened 2017-03-18
fix errors in new connect dialog

### LINKED ISSUE #2492: Can't edit host/port for saved servers
<b>OS:</b>
Win 10 64bit

<b>Cockatrice version:</b>
a8a4557 (2017-03-16)

___
<br>

Introduced with #2452

I encounter this issues:
- ~~Passwords are always saved, even if you don't check the corresponding option~~ fixed in #2503

- You can't change the name of a server profile<br>
As @ZeldaZach told me you're not able to edit the Profile name from within the UI right now
~~**possible workaround:** grey out the `Name` field if `previous host` is selected~~ done in #2503
**possible solution:** make profile names editable as well, the same way as `Host` and `Port`<br>
I think the name change is a nice way to allow people to customize things and prevent users from adding new profile duplicates. Especially since they can't delete the old ones from within the interface right now. <br>
*Another related problem here is, that if you try to change the name in the name field (which is possible right now) you can properly connect, but any changes you do to the profile like a new port won't be saved. Everything will reset and you're presented with the default view in the connect dialog for the next time!*

<br>

CC: @ZeldaZach

ZeldaZach:
## Related Ticket(s)
Fixes mostly #2492 
> Passwords are always saved, even if you don't check the corresponding option

Fixed

> You can't change the name of a server profile

They are a key, so I can't really do that right now. Maybe in the future

> Grey out the Name field if previous host is selected

Done

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -14,6 +14,8 @@
 #include "settingscache.h"
 #include "userconnection_information.h"
 
+#define PUBLIC_SERVERS_URL "https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers"
+
 DlgConnect::DlgConnect(QWidget *parent)
     : QDialog(parent)
 {
@@ -52,6 +54,14 @@ DlgConnect::DlgConnect(QWidget *parent)
     autoConnectCheckBox = new QCheckBox(tr("A&uto connect"));
     autoConnectCheckBox->setToolTip(tr("Automatically connect to the most recent login when Cockatrice opens"));
 
+    publicServersLabel = new QLabel(tr("Looking for a list of public servers? <a href=\"%1\">Click here</a>!").arg(PUBLIC_SERVERS_URL));
+    publicServersLabel->setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Fixed);
+    publicServersLabel->setWordWrap(true);
+    publicServersLabel->setTextFormat(Qt::RichText);
+    publicServersLabel->setTextInteractionFlags(Qt::TextBrowserInteraction);
+    publicServersLabel->setOpenExternalLinks(true);
+    publicServersLabel->setAlignment(Qt::AlignCenter);
+
     if (savePasswordCheckBox->isChecked())
     {
         autoConnectCheckBox->setChecked(settingsCache->servers().getAutoConnect());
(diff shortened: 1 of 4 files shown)


tooomm:
Very nice. Working great! 👍 
(I kept the original issue for profile editing)

One tiny issue though:
If you hit `new host` the fields are not cleared anymore... ;)

cockatrice/src/dlg_connect.cpp
```diff
@@ -155,10 +153,10 @@ void DlgConnect::rebuildComboBoxList()
...
     if (savedHostList.size() == 1)
     {
-        settingsCache->servers().addNewServer("Woogerworks Cockatrice", "cockatrice.woogerworks.com", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Chickatrice Cockatrice", "chickatrice.net", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Cockatric.es Cockatrice", "cockatric.es", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Tetrarch Cockatrice", "mtg.tetrarch.co", "4747", "", "", false);
+        settingsCache->servers().addNewServer("Woogerworks", "cockatrice.woogerworks.com", "4747", "", "", false);
+        settingsCache->servers().addNewServer("Chickatrice", "chickatrice.net", "4747", "", "", false);
+        settingsCache->servers().addNewServer("Cockatric.es", "cockatric.es", "4747", "", "", false);
+        settingsCache->servers().addNewServer("Tetrarch", "mtg.tetrarch.co", "4747", "", "", false);
```
tooomm: Do we add `cockatrice` to every server name for a reason?
It looks like unneeded information which just clutters things?

Regarding https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers
`cockatrice.es` is branded in lower case

Can we please have a link to public serves wiki in that dialog as well? In case somebody runs an older software or we don't include changes or new servers quickly enough...

ZeldaZach: I added "Cockatrice" just because I could, I had full control over it and I saw it took up a lot of unnecessary room, so I removed it.

Will fix his server name.

Will put servers wiki in it

ZeldaZach:
@tooomm All changes have been made.
screenshot 2017-03-18 15 04 05

tooomm:
Fields are resetting now. 👍 

Public server link is also working. Not sure where to put it at best though.
Other ideas:
- I thought about putting it in brackets behind the `New Host` label.<br>--> `New Host ("List of Public Servers")`
- Or on the bottom of the Server group: <br>`In case you're experiencing problems, double check`<br> `your server details with the "Public Server List"`

The `Previous Host` should also be renamed to `Known Hosts`, with the several predefined profiles etc. it's a better wording?

The servers are no longer ordered alphabetically now. :)

@woogerboy21 what's your proper server branding? On the public server page it's Woogerworks, on your site and twitter it's WoogerWorks.

woogerboy21:
@tooomm Personally I think I like Woogerworks , but im like windows.... not case sensitive 😄

ZeldaZach:
screenshot 2017-03-19 14 59 33
Final version

PR #2505, opened 2017-03-18
rename from update to check

### LINKED ISSUE #2498: "Update Cockatrice" in menu should be "Check for updates"
"Update Cockatrice" doesn't match what it actually does - which is check.

ZeldaZach:
## Related Ticket(s)
Fix #2498

### DIFF
--- a/cockatrice/src/tab_room.cpp
+++ b/cockatrice/src/tab_room.cpp
@@ -185,12 +185,12 @@ QString TabRoom::sanitizeHtml(QString dirty) const
 
 void TabRoom::sendMessage()
 {
-    if (sayEdit->text().isEmpty()){
+    if (sayEdit->text().isEmpty()) {
         return;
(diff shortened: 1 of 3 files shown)


cockatrice/src/tab_room.cpp
```diff
@@ -185,12 +185,17 @@ QString TabRoom::sanitizeHtml(QString dirty) const
 
 void TabRoom::sendMessage()
 {
-    if (sayEdit->text().isEmpty()){
+    if (sayEdit->text().isEmpty())
+    {
```
Daenyth: I'm guessing your clion style config is wrong; make sure it's set to open braces same line as 'if'.

settings > editor > code style > c++

screen shot 2017-03-18 at 7 50 19 am

Specifically 'other' controls this

ZeldaZach: Fixed

PR #2506, opened 2017-03-18
additional null checks & parsing

### LINKED ISSUE #2466: Can't transform `Liliana, Heretical Healer`.
<b>OS: Linux: Arch Linux</b>

<b>Cockatrice version: Version b2f2e9d (2017-03-07)</b>

___
<br>

Can't transform `Liliana, Heretical Healer`. That is, when I right click, and select `Token: Liliana, Defiant Necromancer` nothing happens.
If I try to transform `Jace, Vryn's Prodigy` it works just fine.
The card/token database is updated.

I checked the xml for the card database, and the information is there, and seems fine.

ZeldaZach:
## Related Ticket(s)
- Fixes #2466 

Parsing issues to to regex

Adds an additional check to make sure "Token:"  was parsed out properly. If not, do it.
Also, additional nullptr checks to prevent crashes.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1063,6 +1063,9 @@ void Player::actCreateToken()
 
 void Player::actCreateAnotherToken()
 {
+    if (lastTokenName.isEmpty())
+        return;
+
     Command_CreateToken cmd;
     cmd.set_zone("table");
     cmd.set_card_name(lastTokenName.toStdString());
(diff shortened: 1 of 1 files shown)


cockatrice/src/player.cpp
```diff
@@ -2535,6 +2541,9 @@ void Player::processSceneSizeChange(int newPlayerWidth)
 
 void Player::setLastToken(CardInfo *cardInfo)
 {
+    if (cardInfo == nullptr)
```
Daenyth: Might want to check the caller of this too to see that it's not using the cardinfo expecting it to be not null

ZeldaZach: Dealt with.

PR #2507, opened 2017-03-18
temporarly remove dev build downloads

tooomm:
no more hosted dev builds on bintray due to agb problems

- remove link


ZeldaZach:
That doesn't really help as we also have the stable releases linked to bintray, maybe consider changing things up on that too?

tooomm:
I think releases are just temporary blocked because of the dev build violation in our account?!
As long as we don't decide to move away from bintray for sure, and have an alternative setup, I would leave it.

Right now, I expect main releases to work again soon.

Daenyth:
Let's hold off on this until we know exactly what we're changing. I don't feel like triggering more CI to wait on for this

Issue #2341, opened 2017-01-04
Release Management Improvements


poixen:
The current release cycle feels like it could be improved. Here are some thoughts:

1. Establish a regular release cycle
   - Users know when a new build will be delivered
   - Developers can align tasks to release date
   - Testing period can be planned in advance
   - Organisational consistency
2. Consistent versioning
   - Current versioning system is split between 2 naming conventions:
     - Client hash (not human readable)
     - Arbitrary named releases (hard to keep track of)
   - Confusing for users / developers / management
   - Hard to distinguish between major and minor changes and releases
   - We should standardise the version naming convention 

Would be good to discuss with the rest of the team, so add feedback in the comments.

woogerboy21:
My $0.02...

> Establish a regular release cycle

Sounds great, though since we dont have consistent developers or really any type of development focus not sure you would end up with situations were the release cycle would come to date and nothing has really changed.  

> Consistent versioning

Drop names, stick with numbers generated by git or whatever product is used for code base management.

tooomm:
>>Establish a regular release cycle

>Sounds great, though since we dont have consistent developers or really any type of development focus not sure you would end up with situations were the release cycle would come to date and nothing has really changed.

This doesn't sound too bad to me. If there is a new release with just a few fixes... it's ok. Otherwise the users would wait even longer for fixes!
And if there is no development at all for one year for example there is nobody releasing either. :P

It could be like with Ubuntu, twice a year, with the year and month as numbers in the name. 15.10,16.04,16.10, ...

I highly like some kind of cycle!

woogerboy21:
> And if there is no development at all for one year for example there is nobody releasing either. 

Your trying to set a precedence on when users should expect improvements and all I'm saying is we aren't a business and thus there might never be.  Personally I'm really ok with trying anything. 

One of the larger issue's I see is the fact that we just don't have enough developers/testers/etc to really implement any code work / testing concept (if you ask me).  Take for example the report of replay watching being broken.  That seems like a pretty big deal to me, one could argue if we had proper testing before deployment it should have never happened, but we have no testing scripts or steps.  And if you ignore that fact there has been no real involvement with the fix.  I jumped on it but haven't seen or heard from anyone about the PR fix I put in place.  No idea if the user has done testing, no idea if anyone reviewed any of the code... nada.  And we are going on how long?

Any idea's on that front?

tooomm:
[...]

Issue #1825, opened 2016-02-21
[gh pages] exchange php script in favor of js/ajax code to provide download links


tooomm:
_(just a reminder to keep track)_
<br>

Right now we use this [php script](https://github.com/Cockatrice/cockatrice.github.io/blob/master/scripts/cockatrice_bin.php) hosted on a external site because github pages are just static and don't allow for server executable code.
The best solution would be to use JS/AJAX for that instead, implemented into the site.
Unfortunately this isn't supported by the Bintray API yet.

TMiguelT even wrote the support about that:
![bintray ajax](https://files.gitter.im/Cockatrice/Cockatrice/Jh2f/bintray-ajax.png)
Gitter Source: :point_up: [January 7, 2016 4:20 AM](https://gitter.im/Cockatrice/Cockatrice?at=568dd98084fa46770b25014a)

@TMiguelT Any news on that?

<br>
They might announce new features on their blog: http://blog.bintray.com/

TMiguelT:
No news from me, but I'm sure I can help fix this once they implement CORS. Feel free to email them yourself or watch the blog until then.

tooomm:
Update from support:

> The ETA for this is Q2 2016.

ZeldaZach:
Awesome!

ZeldaZach:
Any word on this?

tooomm:
Not yet. I'll ask them in a few weeks for an update.

tooomm:
At the moment we don't use the php script anymore, we use static links that we need to update every release instead.

It seems like there are options to query the github api to somehow accomplish the same and dynamically link to the newest release files on github instead of bintray on our gh webpage:
http://stackoverflow.com/questions/24987542/is-there-a-link-to-github-for-downloading-a-file-in-the-latest-release-of-a-repo
http://stackoverflow.com/questions/28236859/download-link-for-latest-release-of-github-project?rq=1

Is something pretending us from using this instead for the download links on the webpage?

tooomm:
@TMiguelT Calling for help here.
We switched to host our releases on github instead of bintray... and their API supports CORS!

Do you want to have a look?? 😃

tooomm:
There is a pretty cool page to check the download count of releases:
http://www.somsubhra.com/github-release-stats/

It uses js to query the github api as well, source code available here:
https://github.com/Somsubhra/github-release-stats/blob/master/js/main.js
We just need to adapt it to get the information that we are looking for. 👍

PR #2508, opened 2017-03-18
[WIP] release management todo list

tooomm:
TODO
- rephrase
- include new handling https://github.com/Cockatrice/Cockatrice/pull/2570#discussion_r110082650 --> https://github.com/Cockatrice/Cockatrice/pull/2508#discussion_r112707155
- cmake version bump https://github.com/Cockatrice/Cockatrice/pull/2613 (also link to file/line, increment y value) --> new: https://github.com/Cockatrice/Cockatrice/pull/2859
*is right way of doing things? we increase the version number and have a new one out without a hint for beta or pre-release etc., `2.1.18-custom` is more related to 2.1.17 with custom changes - than a future final 2.1.18 version. `-custom` doesn''t indicate a downgrade, it symbolises more of an update/change to something existent*
use github api call to get latest tag and work from there? in cmake file only a dummy default used for custom builds (overwritten by tags anyway): https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt#L25
- include info that the target for the release should be `master` branch, not a single commit! (git tag command can't target branches? --> github help)
- link to semver with it's "rules"
- don't upload newly build strings to transifex --> just pr/merge in repo, transifex will auto-update daily!
- rough procedure mentioned by Daenyth: https://github.com/Cockatrice/Cockatrice/issues/2341#issuecomment-270726881
- properly take feature relases (minor) into account and adjust semver not always on the patches digit!
actually all releases (except `revision`) are "major" or "minor", and revisions "patches"
- not only delete former beta releases, also corresponding tags
- also announce in cockatrice public gitter chat channel (+link to release)

<br>

deployment / binary naming
- we basically agreed on [this](https://github.com/Cockatrice/Cockatrice/issues/2341#issuecomment-270947017) naming convention for artifacts/binaries, we still use different
should the code name be included? but version number first for sure for uniform chronological ordering
- os tag also needs updates (+uniform 32/63bit naming convention for win)
- make custom tag with commit hash clickable in github issues again
used to be so before changing to current naming scheme with "_", if it's in ( or [ brackets it works when pasted to github, or we have to stripe and print it different to the about dialog window
Example (as is now): "Version 2.3.18-custom_a01c515 (2017-09-27)"
Example (clickable): "Version 2.3.18-custom[a01c515] (2017-09-27)"
- different naming scheme for binaries of different operating systems:
win has `qt5` extension and `win` tag, while mac and ubuntu compiled binaries have no none
--> drop qt hint + add os hint to all? (also stresses that it might only work with that old ubuntu version)
(what is exactly needed to find a new version in the updater? only file extension? also os string for win? ..? 32/64bit is checking installed version, NOT os compatibility)

<br>

Idea:
- is it possible to include a release code name into cmake as well, and only print it to about window (and file names?) next to version number on tagged releases? similar than we handle version numbers in cmake
would prevent [this](https://github.com/Cockatrice/Cockatrice/pull/2508/files#diff-8cc7b2b0d78dd2501610391c086a8516R235) and removing after release

---

The aim is to document the process and make it easier for everybody involved to not miss things.

Open for discussion!
Do you see any changes needed? Did I forget about something? ...

This also adds a simple TOC on top of the file for easier navigation.

cc: @ZeldaZach


.github/CONTRIBUTING.md
```diff
@@ -216,3 +222,24 @@ from Transifex to the source code and vice versa.
...
+
+
+# Release Management #
+
+### ToDo list around new main releases ###
+
+ - write a `Release Draft` on github with all changes since last main release
+ - build newest strings and upload them to transifex
+ - give translators some time --> pull newest translations from transifex and PR them back in
+ - **Final commit:** push a PR which adds the `Release Name` and `Version Number` in the `About` window
```
Daenyth: Not a fan of this dance. Let's think about this some more.

ZeldaZach:
@tooomm If you wanna make changes to this again, I'll re-review at some point :)

.github/CONTRIBUTING.md
tooomm: I don't know of any alternative or automation here... maybe for the version number something is possible because we are going to have them in the release tag on github at some point?

tooomm:
@ZeldaZach The earliest change I hope to make is depended on https://github.com/Cockatrice/Cockatrice/issues/1825 to get rid of the updating process on the gh page with new download links.

.github/CONTRIBUTING.md
tooomm: @Daenyth Actually with https://github.com/Cockatrice/Cockatrice/pull/2570 (outlined in https://github.com/Cockatrice/Cockatrice/pull/2613) this seems to be even more complex now :D
But I think that's how we have to do it?

Just realized while I wanted to update this pr...

cc: @ctrlaltca

.github/CONTRIBUTING.md
```diff
@@ -217,6 +223,26 @@ from Transifex to the source code and vice versa.
 
 Please have a look at the specific [FAQ for translators](https://github.com/Cockatrice/Cockatrice/wiki/Translation-FAQ).
 
+
+# Release Management #
+
+### ToDo list around new main releases ###
```
ZeldaZach: TODO, reLeases

Nightfirecat: I agree on `TODO`, but the font is what makes it look like `reLeases`; it's actually just `releases`.

skwerlman:
could you rebase this pr please? it's still using the old .gitlab-ci.yml which isn't branch filtered, so it's triggering unneeded pipelines on my server for each commit...

PR #2509, opened 2017-03-18
removed debug leftovers from #2500

tooomm:
## Related Ticket(s)
- cleans #2500 up

- removes unclear variable name leftovers from #2500
- additionally this removes the no longer needed .zip file workaround for a wrong release


Daenyth:
I've already got this in my branch for the updater

PR #2510, opened 2017-03-18
better null checking decklist

### LINKED ISSUE #2397: OS X after the new update, details inside
<b>OSX 10.10.5:</b>

<b>dab7316 (2017-01-19):</b>

The client crashes after editing a deck for more than 15 seconds. OS X prompts me asking if Id like to send a report to Apple, obviously I decline.

___
<br>

*Explain your Issue/Request/Suggestion in detail here!*

ZeldaZach:
## Related Ticket(s)
- Fixes #2397 

Crash in deck editor

Better null checking

### DIFF
--- a/cockatrice/src/decklistmodel.cpp
+++ b/cockatrice/src/decklistmodel.cpp
@@ -300,8 +300,9 @@ QModelIndex DeckListModel::addCard(const QString &cardName, const QString &zoneN
 
 QModelIndex DeckListModel::nodeToIndex(AbstractDecklistNode *node) const
 {
-    if (node == root)
+    if (node == nullptr || node == root)
         return QModelIndex();
+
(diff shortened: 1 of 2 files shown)


cockatrice/src/decklistmodel.cpp
```diff
@@ -300,8 +300,9 @@ QModelIndex DeckListModel::addCard(const QString &cardName, const QString &zoneN
 
 QModelIndex DeckListModel::nodeToIndex(AbstractDecklistNode *node) const
 {
-    if (node == root)
+    if (node == nullptr || node == root)
```
Daenyth: Have a look at any code that calls this method too? See if the node they pass being null would break other things

ZeldaZach: I took a look and saw nothing that should mind it. It wants an empty list unless the list is defined already

Issue #2502, opened 2017-03-18
All download links are broken as "forbidden"


jimivyas:
<b>OS:</b>
*running windows 10 pro build 14393.953 but it doesn't really matter.*

<b>browser version: </b>
*Chrome Version 56.0.2924.87 (64-bit)*

___
<br>

All three download links on https://cockatrice.github.io/ are currently broken with the error "Forbidden!"

___
<br>

Daenyth:
We should hopefully have this fixed by tomorrow

Daenyth:
Scratch that - bigger issue at hand. Apparantly bintray (what we use to host downloads) forbids uploading development builds. We'll have to change our CI system to not automatically upload, and delete them. Then we'll also have to do some code changes on the "development" auto update channel for the new storage location.

@ZeldaZach s3 might work.

cc @ctrlaltca

ZeldaZach:
@Daenyth We can probably still go on the upload path, just we need to only have X versions available at once and delete the rest. They have a REST API available to do such actions

Daenyth:
https://bintray.com/docs/terms_of_service.html

> 4. Use Restrictions
> ...
> You may not [...]  (xxvii) in connection with Contributions, upload and/or store on the Site continuous integration artifacts or nightly builds, **(i.e. you may only upload and store content in direct connection to official software releases).**

We can't host dev builds on bintray, period.

ZeldaZach:
Ok, so this brings back the question of going back to hosting locally, which used to be done on WoogerWorks servers... is that still an option, or should we now have to set up our own host services?

If we do it our own way, we can establish our own ways for the auto-updater to work better

Daenyth:
We can still build on CI, we don't need to build locally, we just need to upload the results elsewhere. I'll come up with a plan tomorrow

Daenyth:
Got this automated email

> Hi Cockatrice,
> Your Bintray account is currently rate-limited due to overuse.
> Customer Support should contact you at the earliest to remove the limits from your account. You may also contact them directly at support@jfrog.com.
> Regards,
> The Bintray Team

woogerboy21:
If you guys need a box I can host builds for a bit.

Daenyth:
I think we can just put them in S3 automatically

Daenyth:
Travis even has built in support for this https://docs.travis-ci.com/user/deployment/s3/

Daenyth:
[...]

PR #2515, opened 2017-03-19
Change stable channel downloads to use only github api and links

Daenyth:
## Related Ticket(s)
- Fixing #2502

- We don't hit bintray at all for stable channel updates
- Uses fewer network calls by using the assets list from github api

## Screenshot
screen shot 2017-03-19 at 9 11 35 am

### DIFF
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -8,7 +8,6 @@
 #include <QLabel>
 #include <QProgressBar>
 #include <QApplication>
-#include <QtGlobal>
 
 #include "dlg_update.h"
 #include "releasechannel.h"
@@ -164,7 +163,7 @@ void DlgUpdate::downloadSuccessful(QUrl filepath) {
     if (QDesktopServices::openUrl(filepath))
     {
         QMetaObject::invokeMethod((MainWindow*) parent(), "close", Qt::QueuedConnection);
-        qInfo() << "Opened downloaded update file successfully - closing Cockatrice";
+        qDebug() << "Opened downloaded update file successfully - closing Cockatrice";
         close();
     } else {
         setLabel(tr("Error"));
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -8,6 +8,7 @@
 #include <QLabel>
 #include <QProgressBar>
 #include <QApplication>
+#include <QtGlobal>
 
 #include "dlg_update.h"
 #include "releasechannel.h"
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -163,6 +163,7 @@ void DlgUpdate::downloadSuccessful(QUrl filepath) {
     if (QDesktopServices::openUrl(filepath))
     {
         QMetaObject::invokeMethod((MainWindow*) parent(), "close", Qt::QueuedConnection);
+        qInfo() << "Opened downloaded update file successfully - closing Cockatrice";
         close();
     } else {
         setLabel(tr("Error"));
--- a/cockatrice/src/releasechannel.cpp
+++ b/cockatrice/src/releasechannel.cpp
@@ -3,14 +3,11 @@
 #include "version_string.h"
 
 #include <QNetworkReply>
-#include <QNetworkAccessManager>
 #include <QMessageBox>
 
 #define STABLERELEASE_URL "https://api.github.com/repos/Cockatrice/Cockatrice/releases/latest"
 #define STABLETAG_URL "https://api.github.com/repos/Cockatrice/Cockatrice/git/refs/tags/"
-#define STABLEFILES_URL "https://api.bintray.com/packages/cockatrice/Cockatrice/Cockatrice/files"
-#define STABLEDOWNLOAD_URL "https://dl.bintray.com/cockatrice/Cockatrice/"
-#define STABLEMANUALDOWNLOAD_URL "https://bintray.com/cockatrice/Cockatrice/Cockatrice/_latestVersion#files"
+#define STABLEMANUALDOWNLOAD_URL "https://github.com/Cockatrice/Cockatrice/releases"
 
 #define DEVRELEASE_URL "https://api.github.com/repos/Cockatrice/Cockatrice/commits/master"
 #define DEVFILES_URL "https://api.bintray.com/packages/cockatrice/Cockatrice/Cockatrice-git/files"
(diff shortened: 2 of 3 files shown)


cockatrice/src/releasechannel.cpp
```diff
@@ -68,18 +65,13 @@ bool ReleaseChannel::downloadMatchesCurrentOS(QVariantMap build)
...
 
-    auto fileName = build["name"].toString();
-    // Checking for .zip is a workaround for the May 6th 2016 release
-    auto zipName = arch + ".zip";
     auto exeName = arch + ".exe";
-    auto zipDebugName = devSnapshotEnd + ".zip";
     auto exeDebugName = devSnapshotEnd + ".exe";
-    return (fileName.endsWith(exeName) || fileName.endsWith(zipName) ||
-        fileName.endsWith(exeDebugName) || fileName.endsWith(zipDebugName));
+    return (fileName.endsWith(exeName) || fileName.endsWith(exeDebugName));
```
tooomm: why 2x Debug?

Daenyth:
Currently broken - hitting 'yes' crashes and gives me this file opened
```
{"url":"https://api.github.com/repos/Cockatrice/Cockatrice/releases/assets/3389752","id":3389752,"name":"Cockatrice_Thopter_Pie_Network_osx.dmg","label":null,"uploader":{"login":"ZeldaZach","id":7460172,"avatar_url":"https://avatars3.githubusercontent.com/u/7460172?v=3","gravatar_id":"","url":"https://api.github.com/users/ZeldaZach","html_url":"https://github.com/ZeldaZach","followers_url":"https://api.github.com/users/ZeldaZach/followers","following_url":"https://api.github.com/users/ZeldaZach/following{/other_user}","gists_url":"https://api.github.com/users/ZeldaZach/gists{/gist_id}","starred_url":"https://api.github.com/users/ZeldaZach/starred{/owner}{/repo}","subscriptions_url":"https://api.github.com/users/ZeldaZach/subscriptions","organizations_url":"https://api.github.com/users/ZeldaZach/orgs","repos_url":"https://api.github.com/users/ZeldaZach/repos","events_url":"https://api.github.com/users/ZeldaZach/events{/privacy}","received_events_url":"https://api.github.com/users/ZeldaZach/received_events","type":"User","site_admin":false},"content_type":"application/x-diskcopy","state":"uploaded","size":35024018,"download_count":87,"created_at":"2017-03-14T04:00:28Z","updated_at":"2017-03-14T04:00:39Z","browser_download_url":"https://github.com/Cockatrice/Cockatrice/releases/download/2017-03-14-Release/Cockatrice_Thopter_Pie_Network_osx.dmg"}
```

Daenyth:
Ah, it's not a crash, it's expected behavior (stupid)

`//Try to open the installer. If it opens, quit Cockatrice`

tooomm:
Can you include a short description of how the update process works now?
It's a very critical function of the app^^

Daenyth:
Same as before if you're on stable channel

Daenyth:
Fixed downloading json instead of the file.
screen shot 2017-03-19 at 9 27 13 am

tooomm:
So the internal update check works also the same?
(that was the question... not the steps for the user)

Daenyth:
- Get latest release from github api
- Look for an os-appropriate file from that release's assets
- Get tag name -> commit id from github api. Compare those same as before to see if update is needed

Basically the same process except the old version had an extra step where it would take the commit and look for a file on bintray - now finding the file is part of the first request handler

cockatrice/src/releasechannel.cpp
ZeldaZach: it doesn't really matter what the variable name is called, it's pretty clear what it's use is

cockatrice/src/dlg_update.cpp
```diff
@@ -163,6 +164,7 @@ void DlgUpdate::downloadSuccessful(QUrl filepath) {
     if (QDesktopServices::openUrl(filepath))
     {
         QMetaObject::invokeMethod((MainWindow*) parent(), "close", Qt::QueuedConnection);
+        qInfo() << "Opened downloaded update file successfully - closing Cockatrice";
```
ZeldaZach: So this showing out of scope on the linux builds is a bug in QT < 5.4. Do we compile against that or lower for the linux builds?

Daenyth:
> If i'm on a dev build and switch to stable, it will tell me to "update" to Pie Network, even though, for me, that is a downgrade

I think that's fine for now. When we do semver it will work better since the snapshot can be something like x.y.z-git01234abcd, or whatever we want to make it > x.y.z

>Linux builds are failing to due a bug, so this is fine to go whenever you're ready on it.

Annoying; I think the next-next release should get travis building on modern ubuntu with modern qt via docker. PITA but we have to do it eventually.

> Do you think we should upload this version to bintray as a new "release" and then plan around github releases? I think I'd like to sort out the dev snapshots before we do that, but that's just me.

Yes I want to do that but as you say, not until we figure out dev snapshots

Daenyth:
`/home/travis/build/Cockatrice/Cockatrice/cockatrice/src/releasechannel.cpp:153:11: error: ‘qInfo’ was not declared in this scope`

I think I can fix this with an extra include

Daenyth:
Zach what's the bug? If those travis builds won't work we lose trusty deb builds

cockatrice/src/releasechannel.cpp
tooomm: proper name > bad name only because its use is clear

tooomm:
Another thing I run into while bintray wasn't working...

I always received two stacked error dialogs - is this still the case?
Maybe somebody can adjust their local builds and force an error.

update error since bintray prob

ZeldaZach:
Status of this?

Daenyth:
I haven't had time to address feedback yet. If you want to throw some commits on the branch go for it, otherwise it'll have to wait until I have time

woogerboy21:
> I always received two stacked error dialogs - is this still the case?

@tooomm 
Are you building on your own or pulling from appveyor?

tooomm:
>Are you building on your own or pulling from appveyor?

Appveyor all the time!

ZeldaZach:
Since i fixed the popup issue, what's left on this @tooomm ?

tooomm:
https://github.com/Cockatrice/Cockatrice/pull/2515#pullrequestreview-27797792

`Open download page` (if on release channe)l should have the direct link to latest release.

ZeldaZach:
Fixed @tooomm . I'm ready to merge this.

PR #2522, opened 2017-03-20
allow auto connect (prevHost always checked by default)

ZeldaZach:
Should always default to having the prev host button enabled


tooomm:
sounds logical, after we provide a default list now 👍

PR #2532, opened 2017-03-23
use list over vector

ZeldaZach:
I went to Stackoverflow's guide and checked around to see why QVector caused SIGABRT so frequently. It looks like it was common issue and a workaround seems to be to use QList instead. Similar functionality and we don't use this elsewhere, so lets throw it up!

### DIFF
--- a/cockatrice/src/dlg_viewlog.cpp
+++ b/cockatrice/src/dlg_viewlog.cpp
@@ -24,7 +24,7 @@ DlgViewLog::DlgViewLog(QWidget *parent)
 
 void DlgViewLog::loadInitialLogBuffer()
 {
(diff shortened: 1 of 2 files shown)


Daenyth:
I'm not convinced this will change anything. You'll get sigabrt if you're using the data structure incorrectly. This changes which structure we use but not how we use it. Chances are this is still wrong. It's not like QVector is somehow magically broken.

Let's get a reproducable stack trace and figure out the bug. CLion's debugger can catch the SIGABRT and jump to a stack trace to help with this

Daenyth:
Please? Try using 'debug' in clion, it will give a lot more info

PR #2534, opened 2017-03-24
cancel downloads from updater

### LINKED ISSUE #2518: Download of update installer can't be cancelled once started
<b>Cockatrice version:</b>
piday+some commits
screen shot 2017-03-19 at 9 27 13 am

There should be a "Cancel download" button

ZeldaZach:
## Related Ticket(s)
- Fixes #2518 
- Fixes https://github.com/Cockatrice/Cockatrice/pull/2515#issuecomment-287709757

Allows you to cancel a download mid-way the correct way.

### DIFF
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -3,7 +3,7 @@
 #include <QDesktopServices>
 #include <QMessageBox>
 #include <QVBoxLayout>
-#include <QDialogButtonBox>
+
 #include <QPushButton>
 #include <QLabel>
 #include <QProgressBar>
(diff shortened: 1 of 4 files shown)


cockatrice/src/dlg_update.cpp
```diff
@@ -142,6 +147,12 @@ void DlgUpdate::enableUpdateButton(bool enable) {
     manualDownload->setEnabled(enable);
 }
 
+void DlgUpdate::addStopDownloadAndRemoveOthers(bool enable) {
+    enable ? buttonBox->addButton(stopDownload, QDialogButtonBox::ActionRole) : buttonBox->removeButton(stopDownload);
```
Daenyth: Don't use ternary expressions as statements; only use them when you're using the value that the ternary returns. Use an explicit if/else here

tooomm:
stop button
Looks good. 👍 

Two issues Zach: 😘 
- Once I hit `Stop Download` I receive two confusing errors. Expected are none and just going back to the `Check for Updates` dialog
error msg
Maybe related to https://github.com/Cockatrice/Cockatrice/pull/2515#issuecomment-287709757?
Having the `stop` selected as default is better than `close` i think.

- ~~If you hit 'Open Download Page' it now somehow points you to `https://bintray.com/cockatrice/Cockatrice/Cockatrice/_latestVersion#files` again.~~ Scrap that, Dae's new PR is not yet in...

cockatrice/src/dlg_update.cpp
```diff
@@ -24,17 +24,21 @@ DlgUpdate::DlgUpdate(QWidget *parent) : QDialog(parent) {
...
     progress = new QProgressBar(this);
 
-    QDialogButtonBox *buttonBox = new QDialogButtonBox(this);
+    buttonBox = new QDialogButtonBox(this);
+    buttonBox->setFixedWidth(350);
+
     ok = new QPushButton("Close", this);
     manualDownload = new QPushButton(tr("Reinstall"), this);
-    enableUpdateButton(false); //Unless we know there's an update available, you can't install
+    stopDownload = new QPushButton(tr("Stop Download"), this);
```
tooomm: Exchange for `Cancel Download` as a more familiar wording?

cockatrice/src/update_downloader.cpp
```diff
@@ -14,10 +15,14 @@ void UpdateDownloader::beginDownload(QUrl downloadUrl) {
     response = netMan->get(QNetworkRequest(downloadUrl));
     connect(response, SIGNAL(finished()), this, SLOT(fileFinished()));
     connect(response, SIGNAL(downloadProgress(qint64, qint64)), this, SLOT(downloadProgress(qint64, qint64)));
-    connect(response, SIGNAL(error(QNetworkReply::NetworkError)), this, SLOT(downloadError(QNetworkReply::NetworkError)));
+    //connect(response, SIGNAL(error(QNetworkReply::NetworkError)), this, SLOT(downloadError(QNetworkReply::NetworkError)));
```
Daenyth: 🔪

tooomm:
- `cancel download` button should be the default if you hit enter, not the close button
- once I hit `Cancel Download` I receive ~~two~~ one confusing error. As a user I expected no error if I follow a intended workflow and use a feature. It should just cancel the download and go back to the `Check for Updates` dialog.
The error should be "Operacion **canceled by user**" and go to the logs instead poping up.

btw right now there is no entry in logs about cancel, i just get this:
>Found compatible version url= "https://dl.bintray.com/cockatrice/Cockatrice/Cockatrice_Thopter_Pie_Network_win64.exe"

ZeldaZach:
@tooomm Your first bullet point, I'm fine with it as is and I don't intend on changing that. Your second one, it's a QT thing and the message is fine and a user should understand they triggered it.. so I think this is ready to go.

PR #2541, opened 2017-03-26
better description for issue template

tooomm:
## Related Ticket(s)
- Related https://github.com/Cockatrice/Cockatrice/issues/2527

On mac the about window is placed differently than on windows, might also be different on linux...

- better description to not confuse people
- little updates to the rest of the text

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,10 +1,10 @@
 <b>OS:</b>
-*Override this line with the exact operating system you are running!
-For example: "Win 7 SP1", "macOS 10.12.3", "Ubuntu 16.10" etc.*
+*Put your operating system here! For example: "Win 7 SP2", "macOS 10.12.3", "Ubuntu 16.10" etc.*
 
 <b>Cockatrice version:</b>
 *Put your Cockatrice version number & build date here! For example: "2d53ce9 (2016-02-18)"
-You find them inside the app in the `About Cockatrice` window. If you can't access this menu for any reason, please include the full filename of the installer you used.*
+You find them inside the app under `Help → About Cockatrice` (Window & Linux), or `Cockatrice → About Cockatrice` (macOS).
+If you can't access this menu for any reason, please include the full filename of the installer you used.*
 
 __________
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,11 +1,12 @@
 <b>OS:</b>
-*Override this line with the exact operating system you are running! (e.g. "Win 7 SP1", "OS X 10.8.5", "Ubuntu 15.10" ...)*
+*Override this line with the exact operating system you are running!
```
Daenyth: How about simpler since users often ignore this:

> *Put your operating system here*: for example ....

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,11 +1,12 @@
...
+For example: "Win 7 SP1", "macOS 10.12.3", "Ubuntu 16.10" etc.*
 
 <b>Cockatrice version:</b>
-*Put your Cockatrice version number & build date here! You find them inside the app: `Help` --> `About Cockatrice` (e.g. "2d53ce9 (2016-02-18)"). If you can't access this menu for any reason, please include the full filename of the installer you used.*
+*Put your Cockatrice version number & build date here! For example: "2d53ce9 (2016-02-18)"
+You find them inside the app in the `About Cockatrice` window. If you can't access this menu for any reason, please include the full filename of the installer you used.*
```
Daenyth: > You can find the version under `Help > About Cockatrice` on windows and linux and `Cockatrice > About Cockatrice` on macOS

tooomm:
Do we need that phrase about the installer name at the end?
I guess most people can't provide that information... Also, it should be a rare edge case that `About` is not accessible at all.

I like a simple and clean version more here. 👍

ZeldaZach:
This looks fine to me

tooomm:
@ZeldaZach so you think we should keep this part or just remove it since it's of not too much help?
>If you can't access this menu for any reason, please include the full filename of the installer you used.

ZeldaZach:
Just ask for OS / Version if they can find it, no need for extra words.

tooomm:
This is ready.

Issue #2519, opened 2017-03-19
Dev channel builds broken post-bintray


Daenyth:
follow up #2502

Need to decide where to host builds from CI - probably S3 since it's cheap and integration is easy.

Need to decide how we will manage an index file so the client can tell what dev release is the latest. Hopefully this doesn't require code but if needed we could make something like an AWS lambda that's triggered on S3 upload which can update a static json file

ZeldaZach:
I'm personally looking at DigitalOcean for hosting as I think it would be cheaper then S3...

And I'm all for a JSON file that contains the necessary information (build name, release time, commit hash, download URL)

Daenyth:
I'd rather not manage a full server if we can avoid it. That's why I like s3+lambda, there's 0 management involved. Nothing to keep up to date. And I really do think money cost will be tiny

Daenyth:
I made an s3 bucket. Just need to get travis config in order now

tooomm:
Interesting read:
http://mediagoom.com/blog/How-to-automate-a-@GitHub-release-using-@travisci-and-@appveyor

PR #2516, opened 2017-03-19
remove travis deployment to bintray

tooomm:
## Related Ticket(s)
- Related #2502 
- Expands #2511 

## Short roundup of the initial problem
continuous development build deployment not allowed on bintray

## What will change with this Pull Request?
- stop automated travis deployment to bintray

PR #2550, opened 2017-03-27
Travis/Appveyor publish to github pre-release using tags

ZeldaZach:
This PR allows Travis and Appveyor to publish dev builds with tags to github pre-release. If we set a standard tag to the master branch with a number (i.e. 2.3.14 :wink:) we can then publish constantly. The programs will both overwrite old files if we publish to the same tag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,40 @@
 language: cpp
+sudo: required
 matrix:
   fast_finish: true
   include:
-    # linux debug build, trusty + gcc + qt5
-    - os: linux
-      dist: trusty
-      env: BUILDTYPE=Debug DIST=trusty
-    # osx debug build, osx + clang + qt5
-    - os: osx
-      osx_image: xcode6.4
-      env: BUILDTYPE=Debug
-    # linux trusty release build, precise + gcc + qt5
-    - os: linux
-      dist: trusty
-      env: BUILDTYPE=Release DIST=trusty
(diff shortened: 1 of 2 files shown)


tooomm:
article about travis+appveyor deployment to github: https://github.com/Cockatrice/Cockatrice/issues/2519#issuecomment-288498921

PR removing former deployment from travis to bintray: https://github.com/Cockatrice/Cockatrice/pull/2516

tooomm:
Why is `sudo` needed here?

Can you please explain how the process would look like to work with development releases on github now? 

Does this effect the way we publish normal main github releases in any way? Or can we benefit from that? Maybe some automatic version number or release name taken from the gh tag, that we bake into the release with a skript etc.?

What needs to change in the updater? @Daenyth @ctrlaltca

ZeldaZach:
@tooomm 
Sudo is needed because our build scripts use sudo for apt-get and I couldn't configure the new boxes correctly. I'll just leave it how it is and change it in the future.

This will actually improve our current build cycle as "tags" can either be created via the CLI _or_ by publishing a release on github, so it will automate the release cycle as well.

The updater just needs to check for the latest "pre-release" build and call that the "dev build"

ctrlaltca:
Code looks fine, if it works as expected 👍

ZeldaZach:
@ctrlaltca Gonna have to review it officially :P

PR #2557, opened 2017-03-31
dev release channel

ZeldaZach:
This fixes the dev build update cycle. The dev builds will now come from GitHub and take the latest release posted. If the latest release is a pre-release, for example, it will take that. If it's a full release, it will download that.

The following images will be seen if you're on the dev build only:

If a full release is available, this should be seen:
screenshot 2017-03-30 21 56 25

If a dev release is available, this should be seen:
screenshot 2017-03-30 21 57 13

### DIFF
--- a/cockatrice/src/releasechannel.cpp
+++ b/cockatrice/src/releasechannel.cpp
@@ -238,7 +238,7 @@ void DevReleaseChannel::releaseListFinished()
      */
     QVariantMap resultMap = array.at(0).toObject().toVariantMap();
 
-    if (resultMap.size() == 0) {
+    if (array.size() == 0 || resultMap.size() == 0) {
         qWarning() << "No reply received from the release update server:" << QString(jsonData);
         emit error(tr("No reply received from the release update server."));
         return;
--- a/cockatrice/src/releasechannel.cpp
+++ b/cockatrice/src/releasechannel.cpp
@@ -273,10 +273,10 @@ void DevReleaseChannel::releaseListFinished()
         << "commit=" << lastRelease->getCommitHash()
         << "date=" << lastRelease->getPublishDate();
 
-    QString devbuilddownload_url = resultMap["assets_url"].toString();
+    QString devBuildDownloadUrl = resultMap["assets_url"].toString();
 
-    qDebug() << "Searching for a corresponding file on the dev channel: " << devbuilddownload_url;
-    response = netMan->get(QNetworkRequest(devbuilddownload_url));
+    qDebug() << "Searching for a corresponding file on the dev channel: " << devBuildDownloadUrl;
+    response = netMan->get(QNetworkRequest(devBuildDownloadUrl));
     connect(response, SIGNAL(finished()), this, SLOT(fileListFinished()));
 }
 
--- a/cockatrice/src/releasechannel.cpp
+++ b/cockatrice/src/releasechannel.cpp
@@ -51,7 +51,7 @@ bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
 #include <QSysInfo>
 
 bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
-{VERSION_STRING_NUMS
+{
     QString wordSize = QSysInfo::buildAbi().split('-')[2];
     QString arch;
     QString devSnapshotEnd;
--- a/cockatrice/src/releasechannel.cpp
+++ b/cockatrice/src/releasechannel.cpp
@@ -257,14 +257,14 @@ void DevReleaseChannel::releaseListFinished()
         return;
     }
 
-    if (!lastRelease)
+    if (lastRelease == nullptr)
         lastRelease = new Release;
 
     lastRelease->setCommitHash(resultMap["target_commitish"].toString());
     lastRelease->setPublishDate(resultMap["created_at"].toDate());
 
     QString shortHash = lastRelease->getCommitHash().left(GIT_SHORT_HASH_LEN);
-    lastRelease->setName(resultMap["tag_name"].toString());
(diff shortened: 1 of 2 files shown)


cockatrice/src/releasechannel.cpp
```diff
@@ -49,7 +51,7 @@ bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
 #include <QSysInfo>
 
 bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
-{
+{VERSION_STRING_NUMS
```
Daenyth: ?

cockatrice/src/releasechannel.cpp
```diff
@@ -219,63 +221,75 @@ QString DevReleaseChannel::getReleaseChannelUrl() const
...
-        qWarning() << "No reply received from the release update server:" << tmp;
+    QJsonDocument doc = QJsonDocument::fromJson(jsonData);
+    QJsonArray array = doc.array();
+
+    /*
+     * Get the latest release on GitHub
+     * This can be either a pre-release or a published release
+     * depending on timing. Both are acceptable.
+     */
+    QVariantMap resultMap = array.at(0).toObject().toVariantMap();
```
Daenyth: `array.at(0)` can fail if we ever deleted the releases or if the github api for some reason returned bad json.

Does the app crash in that case?

cockatrice/src/releasechannel.cpp
```diff
@@ -290,20 +304,21 @@ void DevReleaseChannel::fileListFinished()
     foreach(QVariant file, resultList)
     {
         QVariantMap map = file.toMap();
-        if(!map.contains("version"))
-            continue;
-        if(!map["version"].toString().endsWith(shortHash))
+
+        // File name _MUST_ contain the build hash
```
Daenyth: Quite awkward. Do we have to have it this way? Can't we just go off the tag name + file?

Dev channel updater should release stable releases if that's the latest upload

cockatrice/src/releasechannel.cpp
ZeldaZach: Did not account for this will re-review

cockatrice/src/releasechannel.cpp
ZeldaZach: I agree it's awk, will look at a change

cockatrice/src/releasechannel.cpp
ZeldaZach: Copy/pasted wrong

cockatrice/src/releasechannel.cpp
ZeldaZach: If array size is 0, will return as a bad input.

If bad json is found, that is checked for below

ZeldaZach:
@Daenyth QJsonDocument was added in QT5.0.0, so it's fine to use.

ZeldaZach:
Works on Mac as expected, will need Windows check done to verify.

cockatrice/src/releasechannel.cpp
ctrlaltca: `array.at(0)` will return a `QJsonValue` of type `Undefined`; the following call to `toObject()` will return an empty `QJsonObject`, and `toVariantMap()` will then return an empty `QVariantMap`.
Each one of these calls could fail for some reason, but it should be safe anyway.

cockatrice/src/releasechannel.cpp
```diff
@@ -219,63 +221,75 @@ QString DevReleaseChannel::getReleaseChannelUrl() const
...
-    qDebug() << "Got reply from release server, size=" << tmp.size()
+    qDebug() << "Got reply from release server, size=" << resultMap.size()
         << "name=" << lastRelease->getName()
         << "desc=" << lastRelease->getDescriptionUrl()
         << "commit=" << lastRelease->getCommitHash()
         << "date=" << lastRelease->getPublishDate();
 
-    qDebug() << "Searching for a corresponding file on the dev channel: " << QString(DEVFILES_URL);
-    response = netMan->get(QNetworkRequest(QString(DEVFILES_URL)));
+    QString devbuilddownload_url = resultMap["assets_url"].toString();
```
ctrlaltca: minor: please don't mix different variable name styles. Everything else is camelCased, this should be devBuildDownloadUrl or similar.

cockatrice/src/releasechannel.cpp
```diff
@@ -66,8 +68,8 @@ bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
     }
 
     auto exeName = arch + ".exe";
-    auto exeDebugName = devSnapshotEnd + ".exe";
-    return (fileName.endsWith(exeName) || fileName.endsWith(exeDebugName));
+    auto exeDevName = devSnapshotEnd + ".exe";
+    return (fileName.endsWith(exeName) || fileName.endsWith(exeDevName));
```
tooomm: ... ❤️

cockatrice/src/releasechannel.cpp
```diff
@@ -219,63 +221,75 @@ QString DevReleaseChannel::getReleaseChannelUrl() const
...
 
-    if(!(resultMap.contains("commit") && 
-        resultMap.contains("html_url") && 
-        resultMap.contains("sha") &&
-        resultMap["commit"].toMap().contains("author") &&
-        resultMap["commit"].toMap()["author"].toMap().contains("date"))) {
-        qWarning() << "Invalid received from the release update server:" << tmp;
+    // Make sure resultMap has all elements we'll need
+    if (!resultMap.contains("assets") ||
+        !resultMap.contains("author") ||
```
tooomm: We don't use `author` anymore?

Can't find `assets` either.

ZeldaZach:
Since this works as expected, on mac/windows, will merge when passes

PR #2561, opened 2017-04-01
fix c style casting

### LINKED ISSUE #2558: Remove c-style pointer casts
https://www.codacy.com/app/Cockatrice/Cockatrice/issues?bid=4363391&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6WyJDUFAiXX0seyJpZCI6IkNhdGVnb3J5IiwidmFsdWVzIjpbIkNvZGUgU3R5bGUiXX0seyJpZCI6IkxldmVsIiwidmFsdWVzIjpbbnVsbF19LHsiaWQiOiJQYXR0ZXJuIiwidmFsdWVzIjpbMjEzMl19LHsiaWQiOiJBdXRob3IiLCJ2YWx1ZXMiOltudWxsXX0seyJ2YWx1ZXMiOltdfV0=

See https://www.securecoding.cert.org/confluence/display/cplusplus/EXP05-CPP.+Do+not+use+C-style+casts

ZeldaZach:
## Related Ticket(s)
- Fixes #2558 

Changes c style casting to C++ style

### DIFF
--- a/cockatrice/src/filtertree.cpp
+++ b/cockatrice/src/filtertree.cpp
@@ -27,7 +27,7 @@ int FilterTreeBranch<T>::childIndex(const FilterTreeNode *node) const
 
     /* to do the dynamic cast to T we will lose const'ness, but we can
      * trust QList::indexOf */
-    unconst = (FilterTreeNode *) node;
+    unconst = const_cast<FilterTreeNode *>(node);
     downcasted = dynamic_cast<T>(unconst);
     if (downcasted == NULL)
         return -1;
--- a/cockatrice/src/filtertree.cpp
+++ b/cockatrice/src/filtertree.cpp
@@ -25,8 +25,6 @@ int FilterTreeBranch<T>::childIndex(const FilterTreeNode *node) const
     FilterTreeNode *unconst;
     T downcasted;
 
-    /* to do the dynamic cast to T we will lose const'ness, but we can
-     * trust QList::indexOf */
     unconst = const_cast<FilterTreeNode *>(node);
     downcasted = dynamic_cast<T>(unconst);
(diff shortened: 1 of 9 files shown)


cockatrice/src/qt-json/json.cpp
```diff
@@ -269,8 +269,7 @@ QVariant Json::parseObject(const QString &json, int &index, bool &success)
         Json::nextToken(json, index);
 
         //Loop through all of the key/value pairs of the object
-        bool done = false;
-        while(!done)
+        while(true)
```
Daenyth: Did you mean to do this?

cockatrice/src/filtertree.cpp
```diff
@@ -27,7 +27,7 @@ int FilterTreeBranch<T>::childIndex(const FilterTreeNode *node) const
 
     /* to do the dynamic cast to T we will lose const'ness, but we can
```
Daenyth: I think the comment can go away now that we are using `const_cast`

cockatrice/src/qt-json/json.cpp
ZeldaZach: I changed it intentionally. The review site said that done was never changed to true anywhere in the code and I looked so I just changed it to true to shut them up

Nightfirecat: Maybe the solution would be to use a function rather than an infinite loop to get this return value?

Daenyth: Let's not change it. We can tell codacy to ignore the thing. This is code we vendored in - I can actually set up an ignore on the whole directory

PR #2570, opened 2017-04-05
Change version in about window and installer name

### LINKED ISSUE #1036: ascending versioning instead of random hashes
You can never tell, just from viewing the hash in the about dialog, how old a client is.
This makes bug reports etc. more difficult. Users can't easily compare their versions or know when they should upgrade, too.

Did some quick research and found these links, hope they help:
- http://brianmilco.blogspot.de/2012/11/cmake-automatically-use-git-tags-as.html
- http://eatmyrandom.blogspot.de/2010/06/automate-version-numbering-using-git.html
- http://pukkaone.github.io/2010/12/19/build-number-git-repository.html
- http://stackoverflow.com/questions/1435953/how-can-i-pass-git-sha1-to-compiler-as-definition-using-cmake

An alternative/addition would be to add a build date next to/below our actual version hash.

---
- [x] Added **commit date** to `About` dialog next to the commit hash (#1046)
- [ ] implement some kind of version numbering, style could be: www.semver.org

### LINKED ISSUE #2495: NSIS installer doesn't show build name anymore
Feature introduced in https://github.com/Cockatrice/Cockatrice/pull/1062

I realized a problem with this some time ago, but I thought it's just on development builds...
But actually it's also no longer working for normal releases!
The name always shown is `Cockarice-0`, tried with recent `Thopter Pie Network` release:
nsis_name

A former working version can be viewed here: https://github.com/Cockatrice/Cockatrice/pull/1169

cc: @ctrlaltca

### LINKED ISSUE #2556: Change "About Cockatrice" Menu to show SemVar
Current: (The version in the "about" window is git sha1 + date of build)
```
Version 54911ee (2017-03-24)
```

Future:
```
Version 2.3.15 (xxxx-xx-xx)
Version 2.3.16-betaXXX (xxxx-xx-xx)
```

ctrlaltca:
## Related Ticket(s)
- Fixes #1036
- Fixes #2556
- Fixes #2495
- Related: #2341 

Cockatrice development packages are currently versioned as 0.0.1 plus git commit date and hash
Cockatrice stable packages have no version at all, but only a nickname and an hash

This PR updates Cockatrice's current version from a dummy 0.0.1 to 2.3.15
Version number and wether to add the git commit hash to the commit hash for prerelease versions is controlled from the main CMakeFile.txt:
```
set(PROJECT_NAME "Cockatrice")
set(PROJECT_VERSION_MAJOR "2")
set(PROJECT_VERSION_MINOR "3")
set(PROJECT_VERSION_PATCH "15")
# Set to true to append git commit hash to version number
set(PROJECT_IS_PRERELEASE false)
```
Prelease version examples:
Example installer name: Cockatrice-2.3.15 ~ git20170405.657e1ac.dmg
Version string in about window: `Version 2.3.15~git_657e1ac (2017-04-05)`
Servatrice version reported on startup: `Servatrice 2.3.15~git_657e1ac (2017-04-05) starting.`

Release version examples: 
Example installer name: `Cockatrice-2.3.15.dmg`
Version string in about window: `Version 2.3.15 (2017-04-05)`
Servatrice version reported on startup: `Servatrice 2.3.15 (2017-04-05) starting.`


tooomm:
Also fixes https://github.com/Cockatrice/Cockatrice/issues/2495 ! Yeah 👍 

nsis_name

<br>
Are there any changes needed in the way the updater looks for available updates now? Regarding the new file names...

CMakeLists.txt
```diff
@@ -25,6 +25,11 @@ if(POLICY CMP0064)
 endif()
 
 set(PROJECT_NAME "Cockatrice")
+set(PROJECT_VERSION_MAJOR "2")
+set(PROJECT_VERSION_MINOR "3")
+set(PROJECT_VERSION_PATCH "15")
+# Set to true to append git commit hash to version number
+set(PROJECT_IS_PRERELEASE True)
```
Daenyth: How should we configure travis to do full releases? Right now we rely on the builds that CI produces to populate the release page, but with this the compiled ouput will say it's prerelease

ctrlaltca: The idea was to:
 * update the version number and set PROJECT_IS_PRERELEASE to false just before the release,
 * then tag the release in order to get it built by travis/appveyor
 * if everything is in order, publish it as a release
 * after the release, update the version number to "next dev version" and set PROJECT_IS_PRERELEASE back to True

Daenyth: Yeah, that's not entirely unreasonable.

ctrlaltca:
Any required change? should i bump version to 2.3.16? 17?

ZeldaZach:
Is there a way to make it auto build betas? Like, when I want to publish a beta snapshot, I name the release "Cockatrice-MAJ.MIN.PATCH-betaXXX" and I'd like it to show in the game menu "Cockatrice Version MAJOR.MIN.PATCH-betaXXX" where the Patch is +1 from what the current release is (so atm it would say 2.3.16-beta002"

Daenyth:
There's scripts to do that kind of thing for maven jar snapshots. The same
pattern would work here

ctrlaltca:
I saw that running `git describe --exact-match --tags $(git log -n1 --pretty='%h')` i can get a result like 
 * `cockatrice-2.3.16-beta002` if the current commit is tagged, or
 * `fatal: no tag exactly matches 'b1515223fbc0e6146547e04d25f8986934018873'` if the current commit is not tagged.

This information can be extracted by cmake and used to change the version in the "about" window and the installer name, but we should decide a precise syntax for tags to ensure they would produce meaningful informations.

tritoch:
>Example installer name: Cockatrice-2.3.15.dmg

Can we add the operating system name in the installer file name?

ZeldaZach:
@tritoch The endings are how we've designated the OS'. 
```
.deb => ubuntu 14.04
.dmg => os x 10.11 +
_64.exe => win 64 bit
.exe => win 32 bit
```

skwerlman:
> `.deb => ubuntu 14.04`

That's not a very good way to specify the OS. I can think of a dozen Linux distros that use .deb files in incompatible ways. As an example, .deb files built for Ubuntu 14.04 are not compatible with .deb files for Debian Wheezy.

Daenyth:
We can get the LSB release info to make sure it's the right kind of ubuntu http://stackoverflow.com/a/6315735/350351

skwerlman:
Personally, I think just appending `-ubuntu1404` to the version would be enough.

Daenyth:
Yes, but Cockatrice needs to know what OS it's running on to know that is the right file.

skwerlman:
Usually, an incompatible deb will just fail to install

Daenyth:
It shouldn't be downloaded at all. It will confuse people

skwerlman:
OH! you meant in the auto-updater. Wouldn't it be possible to include a text file created at build time stating what OS it was built for in the deb? That way you wouldn't have to deal with inconsistent lsb_release usage across distros.

Daenyth:
That precludes cross compiling, which isn't ideal.

If getting lsb_release info fails, we know it's not a supported distro, so we can tell the user so

ctrlaltca:
> Full: YYYY-MM-DD-Release-MAJ.MIN.PATCH
> DEV: cockatrice-MAJ.MIN.PATCH-betaXYZ

I have a few doubts about this:
 * The full release installer should be named 2017-04-11-Release-2.3.15.exe/dmg?
 * Where does betaXYZ comes from in the dev release? There are already 3 numbers, why not using them? Is it something like http://semver.org/#spec-item-9 ?
 * Why do we insist on naming things different between stable and dev releases? Is it just to confuse users?

Also how should it deal when the version defined in the cmake scripts is different from the version tagged in git?
We could think of removing the version from cmake, but imho we should be able to build from source even if when not working on a git clone (think of source distributions), so all the informations needed to build the version string should be available in the source code itself (or, like in this PR, inside he cmake build scripts).

ZeldaZach:
The full release has been how it was since we started doign releases around here. We can change dev build to match if it makes you happy to anything you'd like. Please take liberty changing what you find necessarily and I"ll make the changes around elsewhere to comply with all changes.

ctrlaltca:
Summing up what has been discussed on gitter:

The filename format for installers will be:
`<program name>[-ReleaseName]-MAJ.MIN.PATCH[-prerelease_label]`
where <tags> are mandatory and [tags] are optional.

`<program name>` comes from cmakelists.txt, `PROJECT_NAME` (aka Cockatrice).
We'll then check if we are building a tag (that is, a release or prerelease)

If we are _not_ building a tag, we'll use the MAJ.MIN.PATCH values from cmakelists.txt and append "custom" as the prerelease label, ending up with a filename like `Cockatrice-2.3.23-custom_win32.exe`

If we are building a tag, we'll get the tag name using `git` and parse it using `cmake`.
Tag names will use this format:
```
YYYY-MM-DD-Release-MAJ.MIN.PATCH
YYYY-MM-DD-Development-MAJ.MIN.PATCH-betaXYZ
```
If the git tag name contains "Development", we'll extract MAJ.MIN.PATCH and use betaXYZ as the prerelease_label.

If the git tag name contains "Release", we'll extract MAJ.MIN.PATCH and get the release name from github using the "get release info by tag" api, eg: https://api.github.com/repos/Cockatrice/Cockatrice/releases/tags/2017-03-31-Release-2.3.15

```
...
  "name": "Cockatrice: Thopter Pie Network, Revision 1",
...
```
The release name string will need to be cleaned a bit:
 * removing "Cockatrice"
 * removing all punctuation
 * replacing spaces with underscores
to end up with a string like "Thopter_Pie_Network_Revision_1"

Daenyth:
That sounds like a great plan

ctrlaltca:
So, i had the time to implement this.
Tested on a basic commit:
```
-- Commit is not a release or prerelease (no git tag found)
-- Project version: 2.3.15-custom_dd36187
-- Friendly project version: 2.3.15-custom_dd36187 (2017-04-16)
-- Project version filename: Cockatrice-2.3.15-custom_dd36187
```
I added the git commit to the "custom" label to ensure that people having a lot of different builds in the same directory (eg. developers) are able to understand what is what.

Tested on a release commit:
```
-- Commit is a release or prerelease, git tag: 2017-04-15-Release-2.3.16
-- Fetching tag informations from https://api.github.com/repos/Cockatrice/Cockatrice/releases/tags/2017-04-15-Release-2.3.16
-- [download 2% complete]
...
-- [download 100% complete]
-- Project version: 2.3.16
-- Friendly project version: 2.3.16 (2017-04-15)
-- Project version filename: Cockatrice-Thopter_Pie_Network_Revision_2-2.3.16
```

As you can see it works, even if i ave to say that probably adding "revision 2" to the name of a release is making hard to read the version number (and basically breaking semver).
I suppose major version "2" is cockatrice since xml was dropped and protobuf introduced,
minor version "3" is Thopter_Pie_Network
patch version "16" should be what is called "revision" instead.

ctrlaltca:
Rebased on master and pushed again to get the osx compilation fix #2575 in

Daenyth:
> I suppose major version "2" is cockatrice since xml was dropped and protobuf introduced,
> minor version "3" is Thopter_Pie_Network
> patch version "16" should be what is called "revision" instead.

Yes. The reason patch is so high is that we decided thopter pie was `2.3.14`, releasing for pi day (3/14)

ctrlaltca:
> Yes. The reason patch is so high is that we decided thopter pie was 2.3.14, releasing for pi day (3/14)

So revision 2 is 14+2=16, now it makes sense :)

ZeldaZach:
@ctrlaltca I made a mistake and somehow overwrite your changes, can you re-push? Sorry for that!

ctrlaltca:
> @ctrlaltca I made a mistake and somehow overwrite your changes, can you re-push? Sorry for that!

Done, no problem!

tooomm:
I found out about automatic bumping of the last version number with "APPVEYOR_BUILD_VERSION" and "UpdateBuild". You can specify a version scheme in the yml: https://www.appveyor.com/docs/appveyor-yml/
Is that of any help for automating most version increments?

Maybe there is also some way of using the info from tags: https://github.com/appveyor/ci/issues/691#issuecomment-217652469

ctrlaltca:
@tooomm we already get the tag information from github, and that works for both windows (appveyor) and anything else (travis, gitlab..)

(Un)fortunately appveyor/travis/etc have no access to modify the source code here on github, so they can't really update the numerical values in CMakeLists.txt.
We could drop them, but then any cockatrice compiled from a source package (tarball, zip downloaded from github, ...) would lose the informations about the version number.

skwerlman:
I don't think version numbers in source package builds is worth the effort of having to update CMakeLists.txt for each release.

Besides, I'm willing to bet that most manual builds are done using git, so we can use `git describe --abbrev=0 --tags` to get the version info even for builds from master.

Daenyth:
> I don't think version numbers in source package builds is worth the effort of having to update CMakeLists.txt for each release.

It's pretty important for people developing prebuilt packages for linux distributions.

cmake/getversion.cmake
```diff
@@ -1,40 +1,215 @@
...
+	endif()
+
+	string(REPLACE "\n" "" GIT_COM_DATE "${GIT_COM_DATE}")
+	set(GIT_COMMIT_DATE_FRIENDLY "${GIT_COM_DATE}" PARENT_SCOPE)
+
+	string(REPLACE "-" "" GIT_COM_DATE "${GIT_COM_DATE}")
+	set(GIT_COMMIT_DATE "${GIT_COM_DATE}" PARENT_SCOPE)
+endfunction()
+
+function(clean_release_name name)
```
Daenyth: Sometimes people complain about `sbt` build definitions being scala code.

But that is so much better than this :(

Daenyth: To be clear, this is correct and fine for trice. More a lamentation at the state of C++ build/packaging/distribution tools in general.

ctrlaltca: I know, some aspects of the cmake language are horrific.. but it's still the best cross-platform language by popular choice. Way better than autoconf/automake at least :D

tooomm:
@ctrlaltca can you please take care of the contribution file as well? https://github.com/Cockatrice/Cockatrice/pull/2570#discussion_r110082650

cmake/getversion.cmake
```diff
@@ -201,7 +201,7 @@ if(PROJECT_VERSION_LABEL)
 	set(PROJECT_VERSION "${PROJECT_VERSION}-${PROJECT_VERSION_LABEL}")
 endif()
 
-set(PROJECT_VERSION_FRIENDLY "${PROJECT_VERSION} (${GIT_COMMIT_DATE_FRIENDLY})")
+set(PROJECT_VERSION_FRIENDLY "${PROJECT_VERSION}<br>(${GIT_COMMIT_DATE_FRIENDLY})")
```
ctrlaltca: a-hem:
```
fab$ ./release/servatrice.app/Contents/MacOS/servatrice 
Using configuration file:  "servatrice.ini"
Servatrice 2.3.17-custom_51a9899<br>(2017-04-26) starting.
-------------------------
```

tooomm: I guess that's also the reason for this faulty display:
untitled

ctrlaltca: Yep

PR #2565, opened 2017-04-03
updated download section

tooomm:
## Related Ticket(s)
- Related #2550, #2536

## What will change with this Pull Request?
- remove placeholder-link and link to general release page with most recent release (normal or pre-release) on top

## Screenshots
https://github.com/Cockatrice/Cockatrice/tree/tooomm-readme_downloads#downloads

PR #2584, opened 2017-04-11
Fix Pre-release / Dev build link

tritoch:
Link to dev builds not valid.

- Link now points to releases section


Daenyth:
> Fixes #IssueNumber

?

skwerlman:
I don't think there is an associated issue.

Daenyth:
Then remove the boilerplate :P

tooomm:
Thanks for picking up my change @tritoch 

But I guess the same complains apply to this pr @Daenyth?! https://github.com/Cockatrice/Cockatrice/pull/2565#issuecomment-291116456

PR #2604, opened 2017-04-20
clear focus only if needed

### LINKED ISSUE #2593: Viewing replays causes program to crash
<b>OS: Win10 Home (x64) - inc. all updates</b>

<b>Cockatrice version: dc6c375 (2017-04-15)</b>

__________

Cockatrice crashes (with notification box) soon after starting to view a game replay. Without exception (so to speak ;)). I am not familiar with this issue at all from previous versions, but I might in fact have missed a few of those...

Otherwise, the program seems stable. And all other programs I use (many) seem fine, fwiw.

### LINKED ISSUE #2602: Spec mode chat
Spectator mode should never disengage from chat

ZeldaZach:
## Related Ticket(s)
Fix #2593 
Fix #2602

This will fix the replay problem as sayEdit might not be defined. Also, we will now only clear the focus if the user's text field is empty (they're not actively typing anything, if they were, we should keep focus no matter what).

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -755,7 +755,7 @@ void TabGame::processGameEventContainer(const GameEventContainer &cont, Abstract
                 case GameEvent::GAME_CLOSED: eventGameClosed(event.GetExtension(Event_GameClosed::ext), playerId, context); break;
                 case GameEvent::SET_ACTIVE_PLAYER: eventSetActivePlayer(event.GetExtension(Event_SetActivePlayer::ext), playerId, context); break;
                 case GameEvent::SET_ACTIVE_PHASE:
-                    if (sayEdit && sayEdit->text().size() == 0)
+                    if (sayEdit && sayEdit->text().isEmpty())
                         sayEdit->clearFocus();
                     eventSetActivePhase(event.GetExtension(Event_SetActivePhase::ext), playerId, context);
                     break;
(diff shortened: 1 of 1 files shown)


cockatrice/src/tab_game.cpp
```diff
@@ -755,7 +755,8 @@ void TabGame::processGameEventContainer(const GameEventContainer &cont, Abstract
                 case GameEvent::GAME_CLOSED: eventGameClosed(event.GetExtension(Event_GameClosed::ext), playerId, context); break;
                 case GameEvent::SET_ACTIVE_PLAYER: eventSetActivePlayer(event.GetExtension(Event_SetActivePlayer::ext), playerId, context); break;
                 case GameEvent::SET_ACTIVE_PHASE:
-                    sayEdit->clearFocus();
+                    if (sayEdit && sayEdit->text().size() == 0)
```
ctrlaltca: Equivalent: sayEdit->text().isEmpty()

backendr:
**This needs to be reverted**

It was not tested enough and this cripples communication between players.
You upset everyone at the cost of a minority who  use the replay featue.

@ZeldaZach @Daenyth @tooomm @tritoch @ctrlaltca

Daenyth:
@backendr Please don't just come in and make demands of people. I appreciate that you're trying to contribute to the application, but there are more constructive ways.

PR #2605, opened 2017-04-20
disable join if spec disabled

### LINKED ISSUE #2599: Disable Join as Spectator button
Suggestion:

Disable the "Join as Spectator" button if the highlighted game does not allow spectators.

:)

ZeldaZach:
## Related Ticket(s)
- Fixes #2599 

Join button should be disabled if the game doesn't allow spectators. This does so.

## Screenshots
screenshot 2017-04-19 23 31 47

### DIFF
--- a/cockatrice/src/gameselector.cpp
+++ b/cockatrice/src/gameselector.cpp
@@ -98,6 +98,7 @@ GameSelector::GameSelector(AbstractClient *_client, const TabSupervisor *_tabSup
 
     connect(joinButton, SIGNAL(clicked()), this, SLOT(actJoin()));
     connect(spectateButton, SIGNAL(clicked()), this, SLOT(actJoin()));
+    connect(gameListView, SIGNAL(clicked(const QModelIndex &)), this, SLOT(updateButtonChoices(const QModelIndex &)));
     connect(gameListView, SIGNAL(activated(const QModelIndex &)), this, SLOT(actJoin()));
 }
 
@@ -144,7 +145,7 @@ void GameSelector::checkResponse(const Response &response)
     if (createButton)
         createButton->setEnabled(true);
     joinButton->setEnabled(true);
-    spectateButton->setEnabled(true);
+
 
     switch (response.response_code()) {
         case Response::RespNotInRoom: QMessageBox::critical(this, tr("Error"), tr("Please join the appropriate room first.")); break;
@@ -157,6 +158,9 @@ void GameSelector::checkResponse(const Response &response)
         case Response::RespInIgnoreList: QMessageBox::critical(this, tr("Error"), tr("You are being ignored by the creator of this game.")); break;
         default: ;
     }
+
+    if (response.response_code() != Response::RespSpectatorsNotAllowed)
+        spectateButton->setEnabled(true);
 }
 
 void GameSelector::actJoin()
@@ -211,4 +215,18 @@ void GameSelector::retranslateUi()
 void GameSelector::processGameInfo(const ServerInfo_Game &info)
 {
     gameListModel->updateGameList(info);
+}
+
+void GameSelector::updateButtonChoices(const QModelIndex &index)
+{
+    if (!index.isValid())
+        return;
+
+    const ServerInfo_Game &game = gameListModel->getGame(index.data(Qt::UserRole).toInt());
+    bool overrideRestrictions = !tabSupervisor->getAdminLocked();
+
+    if (game.spectators_allowed() || overrideRestrictions)
+        spectateButton->setEnabled(true);
+    else
+        spectateButton->setEnabled(false);
 }
--- a/cockatrice/src/gameselector.h
+++ b/cockatrice/src/gameselector.h
@@ -23,6 +23,7 @@ private slots:
     void actCreate();
     void actJoin();
     void checkResponse(const Response &response);
+    void updateButtonChoices(const QModelIndex &);
 signals:
     void gameJoined(int gameId);
 private:


cockatrice/src/gameselector.h
```diff
@@ -23,6 +23,7 @@ private slots:
     void actCreate();
     void actJoin();
     void checkResponse(const Response &response);
+    void updateButtonChoices(const QModelIndex &);
```
backendr: Why aren't you following the `actXxxx` naming conventions?

cockatrice/src/gameselector.cpp
```diff
@@ -212,3 +216,17 @@ void GameSelector::processGameInfo(const ServerInfo_Game &info)
...
     gameListModel->updateGameList(info);
 }
+
+void GameSelector::updateButtonChoices(const QModelIndex &index)
+{
+    if (!index.isValid())
+        return;
+
+    const ServerInfo_Game &game = gameListModel->getGame(index.data(Qt::UserRole).toInt());
+    bool overrideRestrictions = !tabSupervisor->getAdminLocked();
```
backendr: Why is this check being performed?

cockatrice/src/gameselector.cpp
```diff
@@ -212,3 +216,17 @@ void GameSelector::processGameInfo(const ServerInfo_Game &info)
...
+
+    if (game.spectators_allowed() || overrideRestrictions)
```
backendr: Simplify to `spectateButton->setEnabled(game.spectators_allowed() || overrideRestrictions)`?

cockatrice/src/gameselector.cpp
ZeldaZach: Admins/Mods have full authority on a server

cockatrice/src/gameselector.cpp
ZeldaZach: True, might change it to this at some point in the future

cockatrice/src/gameselector.h
ZeldaZach: Didn't realize there was a convention

backendr: Can you change it?

ZeldaZach: If you wanna make a Pr to fix all the styles you outlined here, feel free to and I'll approve it

backendr: Are you able to do it? its your PR

cockatrice/src/gameselector.cpp
backendr: If you want to check for permissions, isnt it better to do it through the user `ServerInfo_User *getUserInfo() const { return userInfo; }` from `TabSupervisor`, rather than checking if a tab is locked?

cockatrice/src/gameselector.h
ctrlaltca: It has already been merged, so no more changes are possible in this PR.

cockatrice/src/gameselector.cpp
ctrlaltca: The "supervisor" tab has a button that locks/unlocks admin superpowers.
The line is making the admin able to override the no-spectators-allowed constraint only when the superpowers are unlocked.

cockatrice/src/gameselector.cpp
ctrlaltca: 👍

tooomm:
Another thing... why don't we do the same with `Join` for full games?

cockatrice/src/gameselector.cpp
backendr: Ok, so you need to click a button to enable a superuser permission.
This line also assigns an inverse result. It returns the `getAdminLocked()` and then `!` it. This doesnt feel right. The `!` should be placed in the check.

cockatrice/src/gameselector.h
backendr: I feel this was merged too soon. One of the reviewers @Daenyth hasnt replied yet.

backendr: @ctrlaltca There is nothing in the contributing.md about peer review processes.

ctrlaltca: Sorry, i don't get what you mean. I just said that right now there is no code convention defined about naming actions `actXXX`, but that as you suggested we should try to start using it and formalize it in the CONTRIBUTING.md file, that contains code best practices.

backendr: Is there any documentation on peer review processes?

cockatrice/src/gameselector.cpp
ctrlaltca: Imho having a variable named `overrideRestrictions` helps understanding what is actually checked in the `if()`, while just putting `!tabSupervisor->getAdminLocked()` would make it more obscure, as you originally noticed by not getting why the check was performed.

cockatrice/src/gameselector.h
ctrlaltca: Nope; afaik github actually has no way to force review by external peers.
It typically boils down to th peer leaving a 👍  on the PR or commenting it.

cockatrice/src/gameselector.cpp
backendr: `bool adminLocked = tabSupervisor->getAdminLocked()`
Return if the admin is locked.
`if (!adminLocked)`
If admin is not locked, then proceed. 
You dont need a custom `overrideRestrictions` name when the original variable says all it needs to.

ctrlaltca: It was just imho, but if you know to have the truth, probably you are right.

ZeldaZach:
@tooomm Not a bad idea. can you open a ticket and i'll look into that later?

backendr:
@ZeldaZach are you able to make a new PR addressing the feedback in this PR?

ZeldaZach:
I can, yes, but it's not too high on my priority list right now as I have other tickets I'd rather address. If you'd like to make a PR I'd be more then glad to review it and merge it in

tooomm:
@ZeldaZach will do later, yes. Leaving home now.

PR #2606, opened 2017-04-20
Change cards from sb<->mb with double-click

### LINKED ISSUE #2594: [sideboarding] double clicking a card should swap between main+side
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
dc6c375 (2017-04-15)

__________

Once you join a game room, loaded a deck and unlock your sideboard between games to sideboard...

You need to drag and drop cards to make them change zones, which is not ideal. I would like to be able to double click cards to swap them to the sideboad and vice versa.

ZeldaZach:
## Related Ticket(s)
Fix #2594

This PR allows a person to double click on a card to change it from sideboard to mainboard and vice versa. This will not impact tokens (if you imported them to the deck) and will work in the exact same context as the drag/drop method.

### DIFF
--- a/cockatrice/src/deckview.cpp
+++ b/cockatrice/src/deckview.cpp
@@ -1,5 +1,6 @@
 #include <QApplication>
 #include <QGraphicsSceneMouseEvent>
+#include <QMouseEvent>
 #include <math.h>
 #include "deckview.h"
 #include "decklist.h"
@@ -114,6 +115,39 @@ void DeckViewCard::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
         scene()->addItem(drag);
     }
     setCursor(Qt::OpenHandCursor);
+}
+
+void DeckView::mouseDoubleClickEvent(QMouseEvent *event)
+{
+    if (static_cast<DeckViewScene *>(scene())->getLocked())
+        return;
+
+    if (event->button() == Qt::LeftButton)
+    {
+        QList<MoveCard_ToZone> result;
+        QList<QGraphicsItem *> sel = scene()->selectedItems();
+
+        for (int i = 0; i < sel.size(); i++) {
+            DeckViewCard *c = static_cast<DeckViewCard *>(sel.at(i));
+            DeckViewCardContainer *zone = static_cast<DeckViewCardContainer *>(c->parentItem());
+            MoveCard_ToZone m;
+            m.set_card_name(c->getName().toStdString());
+            m.set_start_zone(zone->getName().toStdString());
+
+            if (zone->getName() == "main")
+                m.set_target_zone("side");
+            else if (zone->getName() == "side")
+                m.set_target_zone("main");
+            else // Trying to move from another zone
+                m.set_target_zone(zone->getName().toStdString());
+
+            result.append(m);
+        }
+
+        deckViewScene->applySideboardPlan(result);
+        deckViewScene->rearrangeItems();
+        emit deckViewScene->sideboardPlanChanged();
+    }
 }
 
 void DeckViewCard::hoverEnterEvent(QGraphicsSceneHoverEvent *event)
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -277,7 +277,7 @@ void DeckViewContainer::sideboardPlanChanged()
     Command_SetSideboardPlan cmd;
     const QList<MoveCard_ToZone> &newPlan = deckView->getSideboardPlan();
     for (int i = 0; i < newPlan.size(); ++i)
-        cmd.add_move_list()->CopyFrom(newPlan[i]);
+        cmd.add_move_list()->CopyFrom(newPlan.at(i));
     parentGame->sendGameCommand(cmd, playerId);
 }
 
--- a/cockatrice/src/deckview.h
+++ b/cockatrice/src/deckview.h
@@ -83,7 +83,6 @@ private:
     qreal optimalAspectRatio;
     void clearContents();
     void rebuildTree();
-    void applySideboardPlan(const QList<MoveCard_ToZone> &plan);
 public:
     DeckViewScene(QObject *parent = 0);
     ~DeckViewScene();
@@ -95,6 +94,7 @@ public:
     void updateContents();
     QList<MoveCard_ToZone> getSideboardPlan() const;
     void resetSideboardPlan();
+    void applySideboardPlan(const QList<MoveCard_ToZone> &plan);
 };
 
 class DeckView : public QGraphicsView {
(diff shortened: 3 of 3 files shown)


cockatrice/src/deckview.cpp
```diff
@@ -1,5 +1,6 @@
 #include <QApplication>
 #include <QGraphicsSceneMouseEvent>
+#include <QDebug>
```
backendr: Is `QDebug` needed in the PR?
I see that it outputs `"LEFT"`, though I assume this was for your own testing.

cockatrice/src/deckview.cpp
```diff
@@ -116,6 +117,34 @@ void DeckViewCard::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
     setCursor(Qt::OpenHandCursor);
 }
 
+void DeckView::mouseDoubleClickEvent(QMouseEvent *event)
+{
+    if (event->button() == Qt::LeftButton)
+    {
+        qDebug() << "LEFT";
```
backendr: This can be removed. See above comment.

cockatrice/src/deckview.cpp
```diff
@@ -116,6 +117,34 @@ void DeckViewCard::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
...
+        QList<MoveCard_ToZone> *result = new QList<MoveCard_ToZone>();
+        QList<QGraphicsItem *> sel = scene()->selectedItems();
+
+        for (int i = 0; i < sel.size(); i++) {
+            DeckViewCard *c = static_cast<DeckViewCard *>(sel.at(i));
+            MoveCard_ToZone m;
+            m.set_card_name(c->getName().toStdString());
+            m.set_start_zone(c->getOriginZone().toStdString());
+
+            if (c->getOriginZone() == "main")
```
backendr: Aren't there some public const variables for these named zones?

cockatrice/src/tab_game.cpp
```diff
@@ -274,13 +275,26 @@ void DeckViewContainer::sideboardLockButtonClicked()
 
 void DeckViewContainer::sideboardPlanChanged()
 {
+    qDebug() << "Standard Sideboard";
```
backendr: Is this needed? See other comments.

cockatrice/src/tab_game.cpp
```diff
@@ -274,13 +275,26 @@ void DeckViewContainer::sideboardLockButtonClicked()
...
     Command_SetSideboardPlan cmd;
     const QList<MoveCard_ToZone> &newPlan = deckView->getSideboardPlan();
     for (int i = 0; i < newPlan.size(); ++i)
         cmd.add_move_list()->CopyFrom(newPlan[i]);
     parentGame->sendGameCommand(cmd, playerId);
 }
 
+void DeckViewContainer::sideboardPlanChanged(QList<MoveCard_ToZone> *newPlan)
+{
+    qDebug() << "Advanced Sideboard" << newPlan->size();
```
backendr: Is this needed? See other comments.

cockatrice/src/deckview.cpp
```diff
@@ -116,6 +117,34 @@ void DeckViewCard::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
     setCursor(Qt::OpenHandCursor);
 }
 
+void DeckView::mouseDoubleClickEvent(QMouseEvent *event)
+{
+    if (event->button() == Qt::LeftButton)
+    {
+        qDebug() << "LEFT";
+        QList<MoveCard_ToZone> *result = new QList<MoveCard_ToZone>();
```
backendr: This resource needs to be deallocated.

cockatrice/src/deckview.cpp
ZeldaZach: This is a WIP, so I have debugging information in here for my purposes. Will all be removed at the end.

cockatrice/src/deckview.cpp
ZeldaZach: Left it allocated on purpose until I could get it to work, then would find a safe place to de-allocate

cockatrice/src/deckview.cpp
ZeldaZach: Nope

Daenyth: Good catch.

Can you add some constants for the special zones?

ZeldaZach: Yeah i'll add that to the PR

cockatrice/src/tab_game.cpp
```diff
@@ -274,13 +275,26 @@ void DeckViewContainer::sideboardLockButtonClicked()
...
+    Command_SetSideboardPlan cmd;
+    for (int i = 0; i < newPlan->size(); ++i)
+        cmd.add_move_list()->CopyFrom(newPlan->at(i));
+    parentGame->sendGameCommand(cmd, playerId);
+
+    // TODO: Need to figure out how to re-paint the deck view
```
ctrlaltca: [DeckViewScene::applySideboardPlan()](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/deckview.cpp#L333) is probably a good starting point.
You probably want to handle this inside `DeckView::mouseDoubleClickEvent`, get the old and new `DeckViewCardContainer` zones from the scene()->cardContainers list, and then move the card from one zone to another the same way it's implemented in [DeckViewCardDragItem::handleDrop](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/deckview.cpp#L36)

ZeldaZach:
So the current WIP has some issues still, but it's a step in the right direction.
- Cards move on double click now to an opposite zone. Issue with this is if I move a card from main to side, then if I click on any copy of that card in either it will force it to the side and cannot be returned to main.
- When this move happens, the server isn't updated with the settings and the deck, when you start the game, still shows you as the 60 card main deck (even if you made changes)

tooomm:
Oh wow, codacy even does reviews of open pr's with comments and info!? sweeet

ZeldaZach:
Any further help @ctrlaltca @Daenyth ?

cockatrice/src/deckview.cpp
ctrlaltca: The main problem now is that you are using `getOriginZone()` to determine where a card actually is and where it should move to, but the "original zone" is never updated, so if you move a card from "main" to "side" its original zone will still be "main"; double clicking it again will try to move it again from the main to the side.
You don't want to che the card "original zone", but its "current zone" instead; you should be able to get this information from the card's [parentItem()](http://doc.qt.io/qt-5.8/qgraphicsitem.html#parentItem)

ZeldaZach:
This PR is ready to be reviewed.

ZeldaZach:
@tooomm I fixed it to work server-side too

PR #2613, opened 2017-04-20
Update CONTRIBUTING.md

ctrlaltca:
## Related Ticket(s)
- Related: #2570

Update CONTRIBUTING.md adding a note about the hardcoded version numbers in CMakeLists.txt and how/when to update them

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -254,5 +254,5 @@ git tag -d $TAG_NAME
 The preferred flow of operations is:
  * just before a release, update the version number in CMakeLists.txt to "next release version";
  * tag the release following the previously described syntax in order to get it built by CI;
- * wait for CI upload the binaries, double check if everything is in order
- * after the release is complete, update again the version number to "next development version", typically increasing `PROJECT_VERSION_PATCH` by one.
+ * wait for CI to upload the binaries, double check if everything is in order
+ * after the release is complete, update the version number again to "next development version", typically increasing `PROJECT_VERSION_PATCH` by one.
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -249,3 +249,10 @@ git tag -d $TAG_NAME
 ```
 
 **NOTE:** Unfortunately, due to the method of how TravisCI and Appveyor work, to publish a full release you will need to make a copy of the release notes locally and then paste them into the GitHub GUI once the binaries have been uploaded. These build sites will automatically change the name of the release (to $TAG_NAME from whatever it was), the status of the release (to pre-release), and the body (to "Dev build of Cockatrice").
+
+**NOTE 2:** In the first lines of https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt there's an hardcoded version number used when compiling custom (not tagged) versions. While on tagged versions these numbers are overriden by the version numbers coming from the tag title, it's a good practice to keep them aligned with the real ones.
+The preferred flow of operations is:
+ * just before a release, update the version number in CMakeLists.txt to "next release version";
+ * tag the release following the previously described syntax in order to get it built by CI;
+ * wait for CI upload the binaries, double check if everything is in order
```
tooomm: Some natives please help with the correct phrasing here, I guess this is not correct!
"wait until the ci uploads..."
"...the ci uploading..."

Nightfirecat: "wait for CI upload" should be "wait for CI **to** upload"
"update again the version number" should be "update **the version number again**"

Otherwise, it seems fine.

Issue #2490, opened 2017-03-17
OSX: Keybinds disabled after sideboard


francisalbar:
<b>OS:</b>
*macOS Sierra 10.12.4 Beta (16E183b)*

<b>Cockatrice version:</b>
*Version 6e723b2 (2017-03-14)*

___
<br>

After sideboarding, keybinds (such as life total cmd+L or mulligan cmd+M) no longer work. Can get around issue by moving to another cockatrice tab, then going back to game tab.

Daenyth:
I wonder if something in the tab is stealing focus and eating the keybinds. Maybe the chat window. I think we had an issue with that before.

ZeldaZach:
I'm OS 10.12.3 and I don't have this issue... this is weird

ZeldaZach:
Change of plans, upon playing I've noticed this too. Weird bug

serrano65:
Hi! I am new at Cockatrice and I have the same issue. Have you solve it? 

Thank you :)

Daenyth:
It's not fixed yet - we've been focusing on some crash bugs with the new release.

Daenyth:
Let's consider re-testing this with #2583 reverted and try to pinpoint the exact spot so we can make a surgical fix, given that 2583 has had unexpected negative impact on other functions.

Really if anyone heroic wants to take a swing at refactoring the game tab from scratch you'd be a hero too, and probably improve this as well as many things.

ctrlaltca:
#2583 reverted in #2619, this is good to be triaged again

tooomm:
#2619 also reverted #2604 ^^

ctrlaltca:
The last good version is "All Hallow's Eve, Revision 2", the first broken is "Thopter Pie Network".
Trying to bisect the bug lead to nowhere, as the problem seems related to the switch from Qt 5.7 to 5.8.
I've just configured travis in order to build using Qt 5.7, so that we can test if it resolves the issue for the moment.

ctrlaltca:
The issue is currently fixed for me in the prerelease https://github.com/Cockatrice/Cockatrice/releases/tag/2017-04-25-Development-2.3.17-beta002 , it would be nice if someone could confirm

ctrlaltca:
Closing, see #2647

PR #2619, opened 2017-04-21
revert clear focus

ZeldaZach:
Will Revert the first attempts at clearing focus. Will look harder into the root cause


backendr:
Excellent work. We will find an alternative solution to the replays.

tooomm:
Replays are no problem, since the replay problem got introduced with the fix for something else related to this.
The initial issue is this: https://github.com/Cockatrice/Cockatrice/issues/2490 (shortcuts on mac don't work after sideboarding)

We are happy if you can share any ideas or possible solutions in the related ticket!

PR #2626, opened 2017-04-22
Handle 'aftermath' cards like 'split' cards

### LINKED ISSUE #2625: 'Aftermath' cards not being combined by Oracle
The new card type in Amonkhet 'Aftermath' ([Start // Finish](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=426917&part=Start)) is not combined by Oracle.  MTGJSON has given them the layout type `aftermath`.

tritoch:
## Related Ticket(s)
- Fixes #2625 

New card type in new set called 'aftermath' should be handled like old card type 'split'.  

- Aftermath cards will show as `TopCard // BottomCard` in Cockatrice card database like split cards.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -165,7 +165,7 @@ int OracleImporter::importTextSpoiler(CardSet *set, const QVariant &data)
         if(layout == "token")
             continue;
 
-        if(layout == "split")
+        if(layout == "split" || layout == "aftermath")
         {
             // Enqueue split card for later handling
             cardId = map.contains("multiverseid") ? map.value("multiverseid").toInt() : 0;


Daenyth:
Gatherer represents aftermath cards as `Cut /// Ribbons` - `///` and not `//`.

Should we handle this? If we don't, it means people pasting in deck lists won't get the right card. If we do, it's more work in oracle and possibly elsewhere

tritoch:
>`///`

That was just how they did it in the spoil gallery.  On gatherer they're same-as-split `//`

tooomm:
Yep! See http://gatherer.wizards.com/Pages/Search/Default.aspx?text=+[heaven]

Daenyth:
Of course they did.

PR #2633, opened 2017-04-24
Specify reason when on game leave; fix #2624

### LINKED ISSUE #2624: If a player gets kicked tell so, not just "leaves game"

ctrlaltca:
## Related Ticket(s)
- Fixes #2624

When an user leaves a game the reason is not stated, leading to misinterpretations:
 * the game host can quietly kick an user saying that he leaved by himself;
 * a user disconnected because of a temporary internet connection problem will seems to have leaved the game.

The reason why a user leaved the game is passed by the server to clients following @Daenyth's suggestion of keeping the protocol backwards compatible:
https://github.com/Cockatrice/Cockatrice/issues/2624#issuecomment-296640397

Please check the reason strings i chose, they need to be short but correct english.

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -1047,13 +1047,13 @@ QString TabGame::getLeaveReason(Event_Leave::LeaveReason reason)
     switch(reason)
     {
         case Event_Leave::USER_KICKED:
-            return tr("kicked");
+            return tr("kicked by game host or moderator");
             break;
-        case Event_Leave::USER_LEAVED:
-            return tr("player leaved the game");
+        case Event_Leave::USER_LEFT:
+            return tr("player left the game");
             break;
         case Event_Leave::USER_DISCONNECTED:
-            return tr("disconnected");
+            return tr("player disconnected from server");
             break;
         case Event_Leave::OTHER:
         default:
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -891,12 +891,12 @@ void TabGame::eventSpectatorSay(const Event_GameSay &event, int eventPlayerId, c
     messageLog->logSpectatorSay(QString::fromStdString(userInfo.name()), UserLevelFlags(userInfo.user_level()), QString::fromStdString(userInfo.privlevel()), QString::fromStdString(event.message()));
 }
 
-void TabGame::eventSpectatorLeave(const Event_Leave & /*event*/, int eventPlayerId, const GameEventContext & /*context*/)
+void TabGame::eventSpectatorLeave(const Event_Leave & event, int eventPlayerId, const GameEventContext & /*context*/)
 {
     QString playerName = "@" + QString::fromStdString(spectators.value(eventPlayerId).name());
     if (sayEdit && autocompleteUserList.removeOne(playerName))
         sayEdit->setCompletionList(autocompleteUserList);
-    messageLog->logLeaveSpectator(QString::fromStdString(spectators.value(eventPlayerId).name()));
+    messageLog->logLeaveSpectator(QString::fromStdString(spectators.value(eventPlayerId).name()), getLeaveReason(event.reason()));
     playerListWidget->removePlayer(eventPlayerId);
     spectators.remove(eventPlayerId);
 
@@ -1042,7 +1042,26 @@ void TabGame::eventJoin(const Event_Join &event, int /*eventPlayerId*/, const Ga
     emitUserEvent();
 }
 
-void TabGame::eventLeave(const Event_Leave & /*event*/, int eventPlayerId, const GameEventContext & /*context*/)
+QString TabGame::getLeaveReason(Event_Leave::LeaveReason reason)
+{
+    switch(reason)
+    {
+        case Event_Leave::USER_KICKED:
+            return tr("kicked");
+            break;
+        case Event_Leave::USER_LEAVED:
+            return tr("player leaved the game");
+            break;
+        case Event_Leave::USER_DISCONNECTED:
+            return tr("disconnected");
+            break;
+        case Event_Leave::OTHER:
+        default:
+            return tr("reason unknown");
+            break;
+    }
+}
+void TabGame::eventLeave(const Event_Leave & event, int eventPlayerId, const GameEventContext & /*context*/)
 {
     Player *player = players.value(eventPlayerId, 0);
     if (!player)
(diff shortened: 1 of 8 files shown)


common/pb/event_leave.proto
```diff
@@ -5,4 +5,11 @@ message Event_Leave {
     extend GameEvent {
         optional Event_Leave ext = 1001;
     }
+    enum LeaveReason {
+        OTHER = 1;
+        USER_KICKED = 2;
+        USER_LEAVED = 3;
```
Daenyth: Should be `USER_LEFT` ; `leaved` isn't correct English

cockatrice/src/tab_game.cpp
```diff
@@ -1042,7 +1042,26 @@ void TabGame::eventJoin(const Event_Join &event, int /*eventPlayerId*/, const Ga
     emitUserEvent();
 }
 
-void TabGame::eventLeave(const Event_Leave & /*event*/, int eventPlayerId, const GameEventContext & /*context*/)
+QString TabGame::getLeaveReason(Event_Leave::LeaveReason reason)
+{
+    switch(reason)
+    {
+        case Event_Leave::USER_KICKED:
+            return tr("kicked");
```
Daenyth: `kicked by game host or moderator` ?

cockatrice/src/tab_game.cpp
```diff
@@ -1042,7 +1042,26 @@ void TabGame::eventJoin(const Event_Join &event, int /*eventPlayerId*/, const Ga
...
+            break;
+        case Event_Leave::USER_LEAVED:
+            return tr("player leaved the game");
```
Daenyth: `player left the game`

cockatrice/src/tab_game.cpp
```diff
@@ -1042,7 +1042,26 @@ void TabGame::eventJoin(const Event_Join &event, int /*eventPlayerId*/, const Ga
...
+            break;
+        case Event_Leave::USER_DISCONNECTED:
+            return tr("disconnected");
```
Daenyth: `player disconnected from server`

ctrlaltca:
Strings updated, thank you

common/pb/event_leave.proto
```diff
@@ -5,4 +5,11 @@ message Event_Leave {
     extend GameEvent {
         optional Event_Leave ext = 1001;
     }
+    enum LeaveReason {
+        OTHER = 1;
+        USER_KICKED = 2;
+        USER_LEFT = 3;
+        USER_DISCONNECTED = 4;
```
tooomm: When a user closes the game it'll show "left"?
When a user closes the app it'll show "left" or "disconnect"?
When a user restarts his PC or his Internet connection drops, what will be shown?

ctrlaltca:
> When a user closes the game it'll show "left"? (other than game --> leave game)

When a user closes the game, he gets asked "do you really want to leave the game?" and if he accepts, cockatrice informs the server that the user left the game. So: "left"

> When a user closes the app it'll show "left" or "disconnect"?

If you "menu->disconnect" from the server, or just close the whole cockatrice app, it will show "disconnect".

> When a user restarts his PC or his Internet connection drops, what will be shown?

"disconnect"

tooomm:
What will the message be if I kill the `Cockatrice` process?

Do we somehow want to (can?) distinguish between disconnecting intentionally and unintentionally or indirectly?
Intentionally beeing hitting disconnect.
Indirectly beeing app close.
Unintentionally being restart or internet drop.

ctrlaltca:
> What will the message be if I kill the Cockatrice process?

"disconnect" once the server hits timeout.

> Do we somehow want to (can?) distinguish between disconnecting intentionally and unintentionally or indirectly?

If the server loses connection with a player, it has no way to know if the user wanted to disconnect or it was a connection problem.

tooomm:
> If the server loses connection with a player, it has no way to know if the user wanted to disconnect or it was a connection problem.

Sure sure, we can't know if they killed the process or restarted their pc.
But we could tell if the app was closed intentionally (hitting "x" or "Cockatrice --> Exit"), same as we know if a player hit "leave game".

ctrlaltca:
> But we could tell if the app was closed intentionally (hitting "x" or "Cockatrice --> Exit"), same as we know if a player hit "leave game".

We could make cockatrice send a "command_leave" to the server for each open game before it disconnects or closes down.

Daenyth:
> But we could tell if the app was closed intentionally (hitting "x" or "Cockatrice --> Exit"),

Currently that's not true, I'm pretty sure closing is just closing and doesn't inform the server.

There's actually perhaps a good reason to keep it that way - you can close the app then reopen, and you will be reconnected to any in-progress games, whereas if you send an explicit disconnect I believe it would close the game

ctrlaltca:
> There's actually perhaps a good reason to keep it that way - you can close the app then reopen, and you will be reconnected to any in-progress games, whereas if you send an explicit disconnect I believe it would close the game

Didn't think of that, it seems reasonable to keep it as it is.

tooomm:
>Currently that's not true

Ok, I didn't know.

If @ctrlaltca's idea (We could make cockatrice send a "command_leave" to the server for each open game before it disconnects or closes down.) prevent's you from reconnecting that would be bad, yes.

Only "problem" left is that rage quitters might tend to just hit "x" and go on, while you might think there is some serious problem and you keep waiting for nothing.

A new command `command_closeapp` brings up new problems? Old clients wouldn't just be able to send it...

ctrlaltca:
> Only "problem" left is that rage quitters might tend to just hit "x" and go on, while you might think there is some serious problem and you keep waiting for nothing.

If they "X" the game, you'll know they left it. If they "X" cockatrice and quit, they solved the problem for you every other player :)

tooomm:
>If they "X" cockatrice and quit, they solved the problem for you and all the other players :)

Just kind of. :)
Because they could just hit a forced daily disconnect for example... if you are in the middle of a game and you enjoy it so far you really want to wait in that case. If somebody just hits "x" without telling you before, you have some additional information to evaluate on.
For example if he doesn't manage to rejoin within 30s in that case because he closed the app on fault he might never will. The problem with raging is that they don't tell you before, so you can't know. :)

Different for a forced internet 24h disconnect by your provider, he will come back within like 5min!

ShrewMaster:
I don't know what's the current status of this, but I thought I should leave a comment.

So far, all the games I've done with the most recent release (2.3.17-beta004 (2017-04-27)) have shown a "reason unknown" when players leave. It still doesn't distinguish the reasons why players leave (in other words, getting kicked, leaving the game or leaving the server all show the same thing). Is this supposed to be happening? If not, could it be because players are running different versions of Cockatrice?

ctrlaltca:
I suppose the server has not been updated yet, so it's not passing the information back to clients.

Daenyth:
cc @woogerboy21 next maintenance window, can you upgrade?

ZeldaZach:
I spoke with Woods a few days ago and this is correct; the server wasn't updated yet. I'll work on a full release at some point

PR #2638, opened 2017-04-24
Fix unknown tokens being drawn as card back

### LINKED ISSUE #2600: Display card names on cards not having a picture bug
If you have the setting mentioned in the title disabled and try to create a numbered token, it doesnt show the card name.

A time when you dod this is during EDH, you create a token with a number to show what you rolled.

It shows an image of a card back with nothing on it. I presume that its seeing the card back as an image and not showing the token name. I think this should be fixed.

ctrlaltca:
## Related Ticket(s)
- Fixes #2600

Unknown tokens now are drawn as card back images.

Unknown tokens will be drawn as rectangles filled with the chosen color.
The token name will be shown also when the "Display card names on cards having a picture" is disabled.

## Screenshots
schermata 2017-04-24 alle 18 03 08

### DIFF
--- a/cockatrice/src/abstractcarditem.cpp
+++ b/cockatrice/src/abstractcarditem.cpp
@@ -95,7 +95,7 @@ void AbstractCardItem::paintPicture(QPainter *painter, const QSizeF &translatedS
     if(facedown)
     {
         // never reveal card color, always paint the card back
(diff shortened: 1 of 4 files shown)


tooomm:
@backendr want to give this a try?

PR #2648, opened 2017-04-25
Edit sets: sort by enabled; fix #2580

### LINKED ISSUE #2580: Edit Sets > sort by enabled does nothing
version snapshot-april-8

ctrlaltca:
## Related Ticket(s)
- Fixes #2580

Sort be "enabled" was not working in the "edit sets" dialog

Sort be "enabled" works in the "edit sets" dialog

### DIFF
--- a/cockatrice/src/setsmodel.cpp
+++ b/cockatrice/src/setsmodel.cpp
@@ -30,10 +30,21 @@ QVariant SetsModel::data(const QModelIndex &index, int role) const
 
     CardSet *set = sets[index.row()];
 
-    if ( role == Qt::CheckStateRole && index.column() == EnabledCol )
-        return static_cast< int >( enabledSets.contains(set) ? Qt::Checked : Qt::Unchecked );
+    if (index.column() == EnabledCol)
+    {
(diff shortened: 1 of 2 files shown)


cockatrice/src/setsmodel.cpp
```diff
@@ -30,10 +30,21 @@ QVariant SetsModel::data(const QModelIndex &index, int role) const
...
+            case Qt::CheckStateRole:
+                return static_cast< int >( enabledSets.contains(set) ? Qt::Checked : Qt::Unchecked );
+            case Qt::DisplayRole:
+            default:
+               return QVariant();
+        }
+    }
 
-    if (role != Qt::DisplayRole)
+    if (role != Qt::DisplayRole && role != SortRole)
```
ZeldaZach: Do we need this here anymore?

ctrlaltca: yep, there are a lot of other roles that we don't implement

ZeldaZach: Sounds good

PR #2652, opened 2017-04-26
Add "Random card from graveyard"

### LINKED ISSUE #2481: Reveal random card from graveyard
Link #2266 

Reveal random card from graveyard (ex. Deadbridge chant etc..) seems like a needed feature.

ZeldaZach:
## Related Ticket(s)
- Fixes #2481 

Couldn't randomly share a card from the graveyard, is a game issue for some cards in the card game MTG.

Add ability to randomly share graveyard with others.
screenshot 2017-04-26 00 47 51

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -287,6 +287,15 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
 
     graveMenu = playerMenu->addMenu(QString());
     graveMenu->addAction(aViewGraveyard);
+
+    if (local) {
+        mRevealRandomGraveyardCard = graveMenu->addMenu(QString());
+        QAction *newAction = mRevealRandomGraveyardCard->addAction(QString());
+        newAction->setData(-1);
+        connect(newAction, SIGNAL(triggered()), this, SLOT(actRevealRandomGraveyardCard()));
+        allPlayersActions.append(newAction);
+        mRevealRandomGraveyardCard->addSeparator();
+    }
     grave->setMenu(graveMenu, aViewGraveyard);
 
     rfgMenu = playerMenu->addMenu(QString());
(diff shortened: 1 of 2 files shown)


Nightfirecat:
I wonder if there's some way to add a generic function across all zones that can contain cards to reveal a random card from it.

ZeldaZach:
The entire game tab is really bad in itself. It needs a rewrite at some point

ctrlaltca:
Appveyor failure is due to an internal appveyor failure, not related to the code from this PR.

cockatrice/src/player.cpp
```diff
@@ -892,6 +902,18 @@ void Player::actViewGraveyard()
...
 
+void Player::actRevealRandomGraveyardCard()
+{
+    Command_RevealCards cmd;
+    QAction *action = static_cast<QAction *>(sender());
+    const int otherPlayerId = action->data().toInt();
+    if (otherPlayerId != -1)
+        cmd.set_player_id(otherPlayerId);
+    cmd.set_zone_name("grave");
+    cmd.set_card_id(-2);
```
ctrlaltca: Note to self: server code in `Server_Player::cmdRevealCards()` considers the card id value of `-2` as "choose a random card from the zone"

ctrlaltca:
Code looks fine and the feature works as expected.
I agree with @Nightfirecat, we should consider the idea of a more generic solution applying to every zone that can contain cards (cardzone). This is something outside the scope of this PR, of course.

cockatrice/src/player.cpp
Daenyth: Dear god please make that a constant :|

That's the worst kind of magic value

ZeldaZach: @Daenyth Want me to do it in this PR ?

cockatrice/src/player.cpp
```diff
@@ -892,6 +902,18 @@ void Player::actViewGraveyard()
...
 }
 
+void Player::actRevealRandomGraveyardCard()
+{
+    Command_RevealCards cmd;
+    QAction *action = static_cast<QAction *>(sender());
+    const int otherPlayerId = action->data().toInt();
+    if (otherPlayerId != -1)
+        cmd.set_player_id(otherPlayerId);
+    cmd.set_zone_name("grave");
```
Daenyth: Here you can make this function generic by accepting the zone name as an argument

ZeldaZach: Not easily, I"d rather leave it as is for now

Daenyth: Then can you ticket the fix and then do it next? Please, we need to make the codebase cleaner so we can start reducing the number of bugs we have.

Daenyth: Alright, that sounds reasonable

PR #2655, opened 2017-04-26
[Don't merge] Grey out unused options in player menu; fix #2640

### LINKED ISSUE #2640: grey out not useful options in right click menu
Same like `Create another token`...
If there are no single Token cards in your Deck, grey out the `Create predefined token` entry, since it won't show anything anyway.
predef_token

Same for `Say` by the way. If there are no macros set up --> grey out.

ctrlaltca:
## Related Ticket(s)
- Fixes #2640

Some player menu options were active even if useless

The following options will be greyed out:
 * "Say" when no macro has been set;
 * "Create predefined token" when the deck has no predefined token
 * "Card" when no card is currently selected

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -330,6 +330,7 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
         aCreateAnotherToken->setEnabled(false);
 
         createPredefinedTokenMenu = new QMenu(QString());
+        createPredefinedTokenMenu->setEnabled(false);
 
         playerMenu->addSeparator();
         countersMenu = playerMenu->addMenu(QString());
@@ -346,6 +347,7 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
         initSayMenu();
 
         aCardMenu = new QAction(this);
+        aCardMenu->setEnabled(false);
         playerMenu->addSeparator();
         playerMenu->addAction(aCardMenu);
 
@@ -825,6 +827,7 @@ void Player::initSayMenu()
     sayMenu->clear();
 
     int count = settingsCache->messages().getCount();
+    sayMenu->setEnabled(count > 0);
 
     for (int i = 0; i < count; i++) {
         QAction *newAction = new QAction(settingsCache->messages().getMessageAt(i), this);
@@ -842,10 +845,15 @@ void Player::setDeck(const DeckLoader &_deck)
     aOpenDeckInDeckEditor->setEnabled(deck);
 
     createPredefinedTokenMenu->clear();
+    createPredefinedTokenMenu->setEnabled(false);
     predefinedTokens.clear();
     InnerDecklistNode *tokenZone = dynamic_cast<InnerDecklistNode *>(deck->getRoot()->findChild("tokens"));
 
     if (tokenZone)
+    {
+        if(tokenZone->size() > 0)
+            createPredefinedTokenMenu->setEnabled(true);
+
         for (int i = 0; i < tokenZone->size(); ++i) {
             const QString tokenName = tokenZone->at(i)->getName();
             predefinedTokens.append(tokenName);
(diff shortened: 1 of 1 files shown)


ZeldaZach:
screenshot 2017-04-26 15 06 44
screenshot 2017-04-26 15 19 02

The dropdown menu still doesn't have the correct blockouts

tooomm:
Hmm, in a local game on windows it works:
untitled

Why do you have several Card items at the end of that list?? Even a different count?

ctrlaltca:
> Card also works, but I'm wondering why we have that at all?!

The "card" menu will contains the submenu of the currently selected card. If no card is currently selected, the entry is greyed out.

> The dropdown menu still doesn't have the correct blockouts

Right, this needs to be checked

> Why do you have several Card items at the end of that list?? Even a different count?

This needs to be checked, too

ctrlaltca:
So, after a few hours debugging why sometimes the menu entries are broken in the main menu, it seems that this only affects macOS and only affects the "global menu bar", and not menus inside the application.

> The dropdown menu still doesn't have the correct blockouts

This is completely random: putting a card on the stack and selecting it / deselecting it, the "card menu" submenu gets correctly added / removed, and the "card menu" action (read: menu entry) gets correctly enabled / disabled, but sometimes the global menu bar forgets to update the "enabled" status of the menu entry.

> Why do you have several Card items at the end of that list?? Even a different count?

This is even better, i've been able to reproduce on a local 4 players game selecting/deselecting cards: it seems that under some conditions the global menubar forgets that the "cards" item already exists, and adds another over and over again.

Moreover, it seems that disabling the native menubar (`menuBar()->setNativeMenuBar(false);`) everything works..

Daenyth:
> Moreover, it seems that disabling the native menubar (menuBar()->setNativeMenuBar(false);) everything works..

That's going to piss off about everyone on OSX though. Can we find another way?

tooomm:
Is the duplication of menu items also mac only?

Maybe we don't need this "feature" at all? I can't come up with a real use case (not even talking about that completely weird and unusual behavior on right clicks --> #2662)

Daenyth:
We shouldn't have to drop it to implement these changes

ZeldaZach:
Do you think we should get this in and work on a mac fix at a later point?

Daenyth:
I'm concerned that if it's adding menu items there's risk of a memory leak or crash

ZeldaZach:
@ctrlaltca What's your opinion?

ctrlaltca:
This fixes something on win/linux, but breaks it a little more on osx; I'd avoid merging it by now.
I could investigate if trying to use another way to create menu popups avoids this problem, but it would take some time to rewrite how all popup works.

Daenyth:
We should probably close this then and revisit when the code is in better
shape

tooomm:
--> WIP instead and add a TODO to the PR description on top?

Daenyth:
Whichever. But if this is going to require refactoring to complete correctly, then it's going to have extensive merge conflicts and need to be rewritten from scratch anyway.

ZeldaZach:
@Daenyth Can we re-eval what's happening with this PR? I see no issue holding it back, as when we made the PR it just helped expose another bug. This doesn't add/change the bug

tooomm:
>I see no issue holding it back, as when we made the PR it just helped expose another bug. This doesn't add/change the bug

Are you certain that the behavior in the client on macOS is the same right now as well @ZeldaZach ?
Even without this PR?

If the problems are identical it looks like no new issues are added with this PR.
It still is an improvement for Linux/Win, and there should be an issue created to track the macOS things down!

cockatrice/src/player.cpp
```diff
@@ -2495,7 +2504,10 @@ void Player::addRelatedCardActions(const CardItem *card, QMenu *cardMenu) {
 void Player::setCardMenu(QMenu *menu)
 {
     if (aCardMenu)
+    {
+        aCardMenu->setEnabled(menu != nullptr);
         aCardMenu->setMenu(menu);
```
Daenyth: This could execute `setMenu(nullptr)` potentially. Does that make sense?

tooomm:
What was your intention to silently close this @ZeldaZach?
Are you sure you want to throw this away?

It looks more like a potential merge candidate... that was at least the latest statement you made.

PR #2661, opened 2017-04-26
remove qt5 tag

tooomm:
All autobuild files for windows on appveyor had a `_qt5` tag in the file.

Since we don't use anything else besides qt5 and don't have this tag on other platforms --> remove.

<br>
Can somebody please double check that this isn't interfering with our updater checking for files?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -77,9 +77,9 @@ build_script:
     - c:\cygwin\bin\ls -l
     - ps: |
         $exe = dir -name *.exe
-        $new_name = $exe.Replace(".exe", "-${env:target_arch}_qt5.exe")
+        $new_name = $exe.Replace(".exe", "-${env:target_arch}.exe")
         Push-AppveyorArtifact $exe -FileName $new_name
-        $cmake_name = $exe.Replace(".exe", "-${env:target_arch}_qt5.cmake.txt")
+        $cmake_name = $exe.Replace(".exe", "-${env:target_arch}.cmake.txt")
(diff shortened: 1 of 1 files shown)


ZeldaZach:
```
master u=!cockatrice> grep "_qt5"
appveyor.yml
80:        $new_name = $exe.Replace(".exe", "-${env:target_arch}_qt5.exe")
82:        $cmake_name = $exe.Replace(".exe", "-${env:target_arch}_qt5.cmake.txt")

cockatrice/src/releasechannel.cpp
61:        devSnapshotEnd = "-x86_64_qt5";
64:        devSnapshotEnd = "-x86_qt5";
```

ZeldaZach:
Personally, I don't feel like breaking the updater yet again for such a minor change, as such I don't think this PR should be continued.

tooomm:
That's a bummer.
Why do we do such bad checks... if at all, the qt5 tag should have been *all the time* before the architecture. --> check consistently for `arch.file_ending`

Daenyth:
We can change the updater first to prep for this and then fix it.

tooomm:
Jep, let's do that. We then have to upload two naming schemes for a period of time to prevent problems?

What should it then check for?
`arch.file_ending`? Or is some more general and future proof thing possible?

It can also unify the used extension for win on our manual main release files and the automated build ones from appveyor. x86_64 vs. win64.
I'll open a new issue about it on the weekend.

Daenyth:
Thanks

PR #2666, opened 2017-04-27
Allow placement below top X cards

### LINKED ISSUE #2665: Putting cards X from the top of the library shortcut
<b>OS:</b>
macOS 10.12.3

<b>Cockatrice version:</b>
Version dab7316

__________

I would like there to be a shortcut in a game to put a card X cards from the top of the library. Due to the new cards Commit // Memory and Approach of the second sun, along with older cards like Unexpectedly Absent, it is very tedious to put facedown the cards from the top of the library one at a time. I would like to request an option where the card can be placed X cards from the top of the deck.

ZeldaZach:
## Related Ticket(s)
- Fixes #2665 
**NOTE:** This PR also addresses an option in the server that if we sent it (-inf, -2] as X, it would crash the user. This just prevents a crash and is backwards compatible with prior client/servers.

Couldn't place a card X cards from top easily into the deck.

## Screenshots
screenshot 2017-04-27 20 27 07
screenshot 2017-04-27 20 27 17
screenshot 2017-04-27 20 27 23

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -95,6 +95,7 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
       game(_parent),
       shortcutsActive(false),
       defaultNumberTopCards(1),
+      defaultNumberTopCardsToPlaceBelow(1),
       lastTokenDestroy(true),
       lastTokenTableRow(0),
       id(_id),
@@ -417,6 +418,7 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
     aMoveToTopLibrary->setData(cmMoveToTopLibrary);
     aMoveToBottomLibrary = new QAction(this);
     aMoveToBottomLibrary->setData(cmMoveToBottomLibrary);
+    aMoveToXfromTopOfLibrary = new QAction(this);
     aMoveToGraveyard = new QAction(this);
     aMoveToHand = new QAction(this);
     aMoveToHand->setData(cmMoveToHand);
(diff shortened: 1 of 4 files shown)


cockatrice/src/player.cpp
```diff
@@ -723,6 +725,7 @@ void Player::retranslateUi()
     }
 
     aMoveToTopLibrary->setText(tr("&Top of library"));
+    aMoveToXfromTopOfLibrary->setText(tr("Below X cards..."));
```
Daenyth: "X cards from the top..."

cockatrice/src/player.cpp
```diff
@@ -1972,6 +1975,45 @@ bool Player::clearCardsToDelete()
...
+    int startPlayerId = cardList[0]->getZone()->getPlayer()->getId();
+    QString startZone = cardList[0]->getZone()->getName();
+
+    Command_MoveCard *cmd = new Command_MoveCard;
+    cmd->set_start_player_id(startPlayerId);
+    cmd->set_start_zone(startZone.toStdString());
+    cmd->mutable_cards_to_move()->CopyFrom(idList);
+    cmd->set_target_player_id(getId());
+    cmd->set_target_zone("deck");
+    cmd->set_x(number);
```
Daenyth: What happens if I set `number = 100` in the UI but I only have 60 cards in the deck?

ZeldaZach: `if number > deck.size() { number = deck.size() }`

cockatrice/src/player.cpp
ZeldaZach: Changed

ForPsionics:
Thanks guys! This is awesome.

cockatrice/src/player.cpp
```diff
@@ -1972,6 +1976,49 @@ bool Player::clearCardsToDelete()
     return true;
 }
 
+void Player::actMoveCardXCardsFromTop()
+{
+    bool ok;
+    int number = QInputDialog::getInt(0, tr("Place card X cards from top library"), tr("How many cards from the top of the deck should this card be placed:"), defaultNumberTopCardsToPlaceBelow, 1, 2000000000, 1, &ok);
```
tritoch: Is this X cards ABOVE our card or X-1?  Unexpectedly Absent says "just beneath the top X cards of your library" whereas Oust says `second from the top` which is clarified as `6/15/2010: If the targeted creature’s owner has one or more cards in his or her library, that creature is put into that library directly under the top card.`

ZeldaZach: 2nd from top = means put 2 in input.
Beneath top X cards = X+1 is your input

tritoch: Should be `top of library` not `top library`

PR #2668, opened 2017-04-29
fix custom build + displaying name

tooomm:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/pull/2570#discussion_r114050968

Introduced in https://github.com/Cockatrice/Cockatrice/pull/2570/commits/b6ecbcc32633f6d1bd19ca5aaba5edb9095892d1
- custom builds had the `<br>` tag in their name
- html code in translation string on transifex

- removed `<br>` tags
The styling is imo also better without them


tooomm:
If we think that `Version 2.3.17-beta004 (2017-04-27)` is too large in one line we can use the suggested way of github: `v2.3.17-beta004 (2017-04-27)` or drop "Version" completely!

By the way, it was annoying to need to fix the linebreaks when writing an issue because the copy and paste from `About` also had them. ;)

cmake/getversion.cmake
```diff
@@ -201,7 +201,7 @@ if(PROJECT_VERSION_LABEL)
 	set(PROJECT_VERSION "${PROJECT_VERSION}-${PROJECT_VERSION_LABEL}")
 endif()
 
-set(PROJECT_VERSION_FRIENDLY "${PROJECT_VERSION}<br>(${GIT_COMMIT_DATE_FRIENDLY})")
+set(PROJECT_VERSION_FRIENDLY "${PROJECT_VERSION}(${GIT_COMMIT_DATE_FRIENDLY})")
```
ctrlaltca: There should be a space here between the version and the date

cockatrice/src/window_main.cpp
```diff
@@ -285,7 +285,7 @@ void MainWindow::actAbout()
 {
     QMessageBox mb(QMessageBox::NoIcon, tr("About Cockatrice"), QString(
         "<font size=\"8\"><b>Cockatrice</b></font><br>"
-        + tr("Version %1").arg(VERSION_STRING)
+        + tr("Version") + QString(" %1").arg(VERSION_STRING)
```
tooomm: Here I can use `tr("Version") + VERSION_STRING` instead?

ctrlaltca: That would miss a space and require an implicit cast from a const char* to QString.

tooomm: Ok, so it's better the way I did it right now.

PR #2504, opened 2017-03-18
rewording in updater

ZeldaZach:
## Related Ticket(s)
- Fixes partly #2497 
- Maybe related #2517

## Short roundup of the initial problem
Wording was ambiguous 

## Screenshots
screenshot 2017-03-18 02 14 54
screenshot 2017-03-18 02 15 06
screenshot 2017-03-18 14 45 46

PR #2672, opened 2017-04-29
update msg rework

tooomm:
## Short round-up of the initial problem
- some strings used to be messed up in transifex:
tx strings
- https://github.com/Cockatrice/Cockatrice/pull/2504/files#diff-68c33782ee685fda6ebc3cc66405a16aR114 was not displayed (Label inside QMessageBox)

- added missing tr() string
- split content and formatting (get rid of translation strings with html tags, formatting and arguments),
modularized the strings into smaller parts as a result
- rephrased messages for better visibility and readability,
tried to uniform messages and still stress the important parts for the user

`\n` is no problem and gets actually recognized as linebreak by transifex. It also tells users to use `Return` in this spot.

## Screenshots
Some example screenshots:
- before:
old-available
after (development snapshots):
new-version
after (stable releases):
new-available

### DIFF
--- a/cockatrice/src/dlg_update.cpp
+++ b/cockatrice/src/dlg_update.cpp
@@ -115,10 +115,10 @@ void DlgUpdate::finishedUpdateCheck(bool needToUpdate, bool isCompatible, Releas
         //If there's no need to update, tell them that. However we still allow them to run the
         //downloader themselves if there's a compatible build
         QMessageBox::information(this, tr("Cockatrice Update"),
-                 tr("Cockatrice is up to date!") + QString ("<br><br>")
-                 + tr("You are already running the latest version available in the chosen release channel.") + QString("<br>")
-                 + tr("Current version") + QString(": %2<br>")
-                 + tr("Current release channel") + QString(": %1").arg(settingsCache->getUpdateReleaseChannel()->getName().toUtf8()).arg(VERSION_STRING));
+                 tr("Cockatrice is up to date!") + "<br><br>"
+                 + tr("You are already running the latest version available in the chosen release channel.") + "<br>"
+                 + tr("Current version") + ": %2<br>"
+                 + tr("Current release channel") + ": %1").arg(settingsCache->getUpdateReleaseChannel()->getName().toUtf8()).arg(VERSION_STRING);
         return;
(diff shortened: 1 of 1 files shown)


tooomm:
Question:
- I'm not sure if both show up here... I guess it's just the QMessageBox title.
Can/should we remove setLabel there?
https://github.com/Cockatrice/Cockatrice/blob/e04010f00b18ac2c7d099d697d3afe29a09b4ffa/cockatrice/src/dlg_update.cpp#L176-L177
- Same for some other locations in that file. L195+196 or 181+183 for example

ZeldaZach:
@tooomm The setLabel() in those situations sets the header of the dialog that pops up

tooomm:
I'm not so sure:
updateerror

setLabel is `Error`. But the headline which show up is `Update Error` and it's defined within " QMessageBox".

ZeldaZach:
Weird, I"m not sure. We can address it in a later PR

tooomm:
We can also just fix it now. :P

It just needs very little input from somebody with better qt knowledge (or sb who understands the qt docs). Not even 5mins of work I guess.
"A stitch in time saves nine."

I'm pretty sure if we don't do it here it'll stay around for a very long time and translators need to deal with dead strings. ;)

ctrlaltca:
Is this ready to go?

tooomm:
There are two titles set for the dialogs, but it looks like just the one in QMessageBox is used, so we end up with unneeded translation strings.
https://github.com/Cockatrice/Cockatrice/pull/2672#issuecomment-298369030

Do you know if it's necessary to do it twice? @ctrlaltca

ctrlaltca:
> setLabel is Error. But the headline which show up is Update Error and it's defined within "QMessageBox".

`setLabel("Error");` is called on the update dialog, not on the popup message. It updates the status label in the middle of the update dialog: https://github.com/Cockatrice/Cockatrice/blob/e04010f00b18ac2c7d099d697d3afe29a09b4ffa/cockatrice/src/dlg_update.cpp#L171

That setLabel call misses the "tr()", so it needs to be fixed anyway.

tooomm:
done and rdy now

PR #2687, opened 2017-05-05
Removed server tab icons

backendr:
The `Server` tab Icon has 2 potential icons. A green dot that is ping/server status(?), and a blue info icon.
From my use, and I assume a large majority of others, I rarely click on the server tab, so it will remain as the info icon.
This PR removes the clutter. Its annoying to constantly have an alert icon when its not needed.

Extra note: When you remove _only_ the info icon, the ping will be a black dot (`updatePingTime(0, -1);`).
It will _only_ change to green if you click on the tab, making me this if this is even working at all.
Further reason I feel that this is not needed.

- Adds a `virtual` function to the `Tab` class to determine if is should update the tab icon
- Remove the green ping dot and info icon from the server tab

### DIFF
--- a/cockatrice/src/tab.h
+++ b/cockatrice/src/tab.h
@@ -33,6 +33,7 @@ public:
     virtual void retranslateUi() = 0;
     virtual void closeRequest() { }
     virtual void tabActivated() { }
+    virtual bool updateIcon() { return true; }
 };
 
 #endif
--- a/cockatrice/src/tab_server.h
+++ b/cockatrice/src/tab_server.h
@@ -51,6 +51,7 @@ public:
     TabServer(TabSupervisor *_tabSupervisor, AbstractClient *_client, QWidget *parent = 0);
     void retranslateUi();
     QString getTabText() const { return tr("Server"); }
+    bool updateIcon() { return false; }
 };
 
 #endif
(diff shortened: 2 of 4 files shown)


cockatrice/src/tab.h
```diff
@@ -33,6 +33,7 @@ protected slots:
     virtual void retranslateUi() = 0;
     virtual void closeRequest() { }
     virtual void tabActivated() { }
+    virtual bool updateIcon() { return true; }
```
backendr: It is used in line `479` in this PR:
```
if (tab->updateIcon()) {
...
}
```

Daenyth: I think I'm going to disable the "unused function" lint since it tends to have many false positives, especially with Qt's signal/slots system

backendr:
If this feature was working, why does the status remain black, when you remove the info icon?

cockatrice/src/tab.h
Daenyth: It is I believe technically correct - you aren't using the *virtual* method implementation but rather the concrete one.It is I believe technically correct - you aren't using the *virtual* method implementation but rather the concrete one.

Daenyth:
I'm not sure, but it certainly worked as of a few weeks ago.

Daenyth:
Just checked, it still works.

screen shot 2017-05-05 at 5 31 30 pm

I saw it turn black for a moment as the icons changed; presumably it's black when being reattached, and then once the next PING proto message comes through it updates itself

backendr:
Edit your code to remove the inclusion of the info icon and you will see it as always black.
I think if you want to keep this status, then keep _only_ the status. Dont obfuscate it with the info icon.

I can see if I can remove the info icon and have the ping icon drawing correctly.

PR #2688, opened 2017-05-05
Prevent emitting initial server message event

backendr:
## Related Ticket(s)
- #2687

When connecting to the server, the `i` will always show and cover the ping status.
To see the ping status, you have to click on the tab to remove the `i`.
The reason the `i` shows is because the client receives the server tab message.

- Initial message will not emit the `i`


tooomm:
Oh wow... I like this tiny change a lot! 👍

ZeldaZach:
Ok, so the black dot has nothing to do with this PR and it was there the entire time under the "(!)" symbol. We should look into a fix in the future, but this is fine for now.

PR #2689, opened 2017-05-06
[WIP] Add ability to place cards at bottom of library randomly

### LINKED ISSUE #673: Cascade can't be performed.
Currently the acceptable way to cascade in Cockatrice is to drag cards from the top of the deck onto the table until you find a card that fits the requirements. The annoying part is dragging each and every card individually to fill the table up with stacks of three cards each, then separating the cards that were originally on the table from the cards that I just revealed while putting those back on the deck.

Here's the alternative I'm proposing:
- Replaces the "view top cards of library..." option and adds a "reveal top cards of library..." option
- Instead of taking you to a menu where you choose the number of cards, then spawning a zone with that number of cards revealed, the option now takes you directly to the zone.
- The zone contains a button and dropdown which allows you to reveal n more cards whenever the button is activated.

I'd be happy to implement if others think the design is sound.

ZeldaZach:
## Related Ticket(s)
- Fixes #673 

Cascade wouldn't work as it should as you cannot randomly set to bottom.

Allow cards to be randomized before being placed back into the deck.
*NOTE:* Currently the log will tell you the order the cards were placed in, even though it is randomized. I couldn't figure out how to tell the server to "forget" the names of the cards before they get placed back into the deck.

## Screenshots
screenshot 2017-05-05 23 53 37
screenshot 2017-05-05 23 53 45

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -418,6 +418,8 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
     aMoveToTopLibrary->setData(cmMoveToTopLibrary);
     aMoveToBottomLibrary = new QAction(this);
     aMoveToBottomLibrary->setData(cmMoveToBottomLibrary);
+    aMoveToBottomLibraryRandomly = new QAction(this);
+    aMoveToBottomLibraryRandomly->setData(cmMoveToBottomLibraryRandomly);
     aMoveToXfromTopOfLibrary = new QAction(this);
     aMoveToGraveyard = new QAction(this);
     aMoveToHand = new QAction(this);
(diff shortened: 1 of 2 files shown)


backendr:
There is a card variable on the server which stores if the card has been revealed. Once it has the server will return the correct names for each. Your logic would have to also be server side.

Whilst this PR adds the function to place them randomly, it doesnt conceal the location/ordering.

I think it needs more work.

Daenyth:
Yeah, this isn't usable - it basically would be cheating by the game rules.

Daenyth:
I'm pretty sure this requires a new protocol item, which is incredibly messy, since clients who don't have the message won't be able to tell what you did. And if the server isn't updated it won't understand the request. It would be really hard to avoid forcing an update on people in that case

ZeldaZach:
I was worried it would require a new protocol item, which is why I was annoyed when I made this.

However, I feel it _is_ an improvement on the current situation as currently you know the order you place them back in as it tells you, but now we can randomize it.

While I agree this PR isn't complete, I feel it makes an incremental change towards supporting the Cascade ability.

tooomm:
Does this has too much room for manipulation if it's not server side?

backendr:
I think most of the player base just drag over a group then move to bottom (cmd+b is a nice shortcut). I have never heard anyone complain about this. Has this been requested from somewhere? Do we want to further clutter the context menus?

Why not make context menus customizable?
I am sure players would appreciate being able to add/remove contexts that they use.

ZeldaZach:
>Why not make context menus customizable?
I am sure players would appreciate being able to add/remove contexts that they use.

That would be awesome. Maybe in another PR we can start working on that.

Daenyth:
> I have never heard anyone complain about this. Has this been requested from somewhere? 

Yes, multiple times. You are literally unable to play one card legally as printed. That's a problem.

> While I agree this PR isn't complete, I feel it makes an incremental change towards supporting the Cascade ability.

Why don't we add the protocol item with no UI for now? That way we can add it as a server-aware feature, and in a few months add the UI portion. By that point, a forced update will be less painful to the user base.

We can even write the UI code now and hide it behind an IFDEF so that we're able to test it but normal players won't have access to it until the full user base does.

PR #2690, opened 2017-05-06
Moved load buttons

backendr:
## Description of feature
I have some custom branches with some tweaks that I like.
This is a simple one, it moves the load deck buttons to the bottom of the screen.
Lifts the deck viewer up. Can be argued its more comfortable to look at.

- Screen layout for deck view

## Screenshots
screen shot 2017-05-06 at 09 36 48

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -125,8 +125,8 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
     connect(deckView, SIGNAL(sideboardPlanChanged()), this, SLOT(sideboardPlanChanged()));
 
     QVBoxLayout *deckViewLayout = new QVBoxLayout;
-    deckViewLayout->addLayout(buttonHBox);
     deckViewLayout->addWidget(deckView);
+    deckViewLayout->addLayout(buttonHBox);
     deckViewLayout->setContentsMargins(0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Maybe we can make it an option of where to put the bar instead of forcing this change? A setting would be cool

backendr:
I am not sure you need to clutter the (already messy) settings with more switches.
Perhaps play with the branch for a day or so, then switch back. See which feels more comfortable.
I found that going from buttons at the bottom, then to top the most jarring.

backendr:
To add, this also follows the "design pattern" from the lobby, where the buttons are below an interface.

Daenyth:
I'm not against it in principle. I definitely don't want a config switch.

Let's poll the community and see which version they prefer, since it's basically arbitrary

ctrlaltca:
Pro: more consistent, looks better.
Con: People are used to see them on top, so be ready for complaints.

Imho :+1

tooomm:
Looks like this on windows - spaces are a bit too small on the bottom of the app I think:
buttons_button

Could lead to miss-clicks. Increase that a little?

ZeldaZach:
I'm moving forward with closing this PR as it would be a chance that I'd rather not force upon users (without a config option) and it doesn't increase UX for users that have used the client for so long.

Daenyth:
I'd like to revisit this when the UI layout is more configurable

PR #2698, opened 2017-05-07
Fix room chat history

woogerboy21:
Looks like an addition to the table shifted things.  We probably  should look into some how use the actual column names (if it can be done).  But for now this should fix the server room chat history.

## Related Ticket(s)
- Fixes broken chat history in rooms (not sure if there is a ticket number)

When users would join rooms previous chat was presented, this was broke by a change in the table / room layout.

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -297,7 +297,7 @@ bool Servatrice::initServer()
             while (query2->next())
                 gameTypes.append(query2->value(0).toString());
                 addRoom(new Server_Room(query->value(0).toInt(),
-                                    query->value(6).toInt(),
+                                    query->value(7).toInt(),
                                     query->value(1).toString(),
                                     query->value(2).toString(),
                                     query->value(3).toString().toLower(),


woogerboy21:
I haven't had time to check out if the functionality is broken using the config to define rooms.  Maybe some one else could do so?  @ctrlaltca  / @ZeldaZach 

This fixes the chat history being broken when the rooms are defined by the db table.

Jinxstands4Jinx:
TBH i liked having the option to see previous chat as it helped when you got randomly kicked via bad weather or dumb pets.

woogerboy21:
@ZeldaZach I looked into the way the server handles creating rooms from the config file.  It actually does string based comparisons to the settings value(s) and as such should work perfectly fine with this change.

If anyone is interested in knowing were the servers block of code is for creating rooms based on configuration its:  
https://github.com/Cockatrice/Cockatrice/blob/master/servatrice/src/servatrice.cpp#L287

In regards to my original statement about if we can use column headers instead of numerical value(s) for definitions.  The way the server handles room creating from db definitions is some what odd.  It runs a first query to pull the information about the rooms definitions out and then from that query runs a second to find out the game types defined for the room and then passes all that information to the room object.  Not sure if you can reference column headers from query results with the way things are written.  Maybe we can?  Some one with a bit more know-how might be able to clear things up.

PR #2670, opened 2017-04-29
Improved token loading, removed card price code

ctrlaltca:
## Related Ticket(s)
- Fixes #472
- Fixes #1599

## Short roundup of the initial problem
When loading a "plain text" deck from a file or clipboard, tokens were not recognized and added to the main deck

## What will change with this Pull Request?
Tokens will be correctly added to the "tokens" deck zone.
Additionally, there were a lot of zone codes ("main", "side", "tokens") scattered around the code. They have been replaced by three new constants `DECK_ZONE_*`.

PR #2706, opened 2017-05-09
Format insensitive cards for loader & Blank Delim for sideboard

### LINKED ISSUE #2701: Suggestion: remove case-sensitivity in "Import from Clipboard" in Deck Editor
OS and version irrelevant.

__________

The idea should be obvious. Case sensitivity doesn't help at all, and creates a lot of unnecessary retyping. Also would fix the "AEther" "Aether" issue.

ZeldaZach:
## Related Ticket(s)
- Fixes #2701 

Cards input to the deck editor weren't validated correctly and, if not typed _exactly_ as they are defined in an XML file, would be added to the "unknown" section.

The sideboard wouldn't begin if there was a line break in a file and insisted on a "Sideboard" word or "SB:" in front of cards.

This PR adds a validation against the installed databases to find which card the inputs match up to and gives them the correct name attribute before loading them to the deck editor, effectively allowing case insensitivity and grammatical insensitivity. 

This PR also addresses an issue when importing decks from MTGGoldFish where they would have the sideboard split up with one blank line. This is now an added feature to check for a blank line and then start the sideboard.

~~It is possible (in the future?) that we should re-address how the loader works and make it "smarter" to check if there is the word "Sideboard" in the file, and if it is there then treat blank lines like nothing.~~

I added some new parts to the parser to make it "smarter," per say. It will now cut out all blank lines at the start of the paste and see how many blank lines there are. If there's only 1, assume it's the sideboard split point. If not, just assume the old logic was correct.

### DIFF
--- a/cockatrice/src/chatview/chatview.cpp
+++ b/cockatrice/src/chatview/chatview.cpp
@@ -93,7 +93,7 @@ void ChatView::appendCardTag(QTextCursor &cursor, const QString &cardName)
     anchorFormat.setAnchor(true);
     anchorFormat.setAnchorHref("card://" + cardName);
     anchorFormat.setFontItalic(true);
-    
+
     cursor.setCharFormat(anchorFormat);
     cursor.insertText(cardName);
     cursor.setCharFormat(oldFormat);
--- a/cockatrice/src/deck_loader.cpp
+++ b/cockatrice/src/deck_loader.cpp
@@ -200,9 +200,19 @@ void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out, const InnerDecklis
 QString DeckLoader::getCardZoneFromName(QString cardName, QString currentZoneName)
 {
     CardInfo *card = db->getCard(cardName);
-    if(card && card->getIsToken())
+    if (card && card->getIsToken())
         return DECK_ZONE_TOKENS;
 
     return currentZoneName;
 }
 
+QString DeckLoader::getCompleteCardName(const QString cardName) const
+{
+    if (db) {
+        CardInfo *temp = db->getCardBySimpleName(cardName);
+        if (temp)
+            return temp->getName();
+    }
+
+    return cardName;
+}
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -3,6 +3,8 @@
 #include <QVariant>
 #include <QCryptographicHash>
 #include <QDebug>
+#include <QVector>
+#include <algorithm>
 #include "decklist.h"
 
 SideboardPlan::SideboardPlan(const QString &_name, const QList<MoveCard_ToZone> &_moveList)
@@ -282,18 +284,18 @@ void AbstractDecklistCardNode::writeElement(QXmlStreamWriter *xml)
 QVector<QPair<int, int> > InnerDecklistNode::sort(Qt::SortOrder order)
 {
     QVector<QPair<int, int> > result(size());
-    
+
     // Initialize temporary list with contents of current list
     QVector<QPair<int, AbstractDecklistNode *> > tempList(size());
     for (int i = size() - 1; i >= 0; --i) {
         tempList[i].first = i;
         tempList[i].second = at(i);
     }
-    
+
     // Sort temporary list
     compareFunctor cmp(order);
     qSort(tempList.begin(), tempList.end(), cmp);
-    
+
     // Map old indexes to new indexes and
     // copy temporary list to the current one
     for (int i = size() - 1; i >= 0; --i) {
@@ -318,7 +320,7 @@ DeckList::DeckList(const DeckList &other)
       deckHash(other.deckHash)
 {
     root = new InnerDecklistNode(other.getRoot());
-    
+
     QMapIterator<QString, SideboardPlan *> spIterator(other.getSideboardPlans());
     while (spIterator.hasNext()) {
         spIterator.next();
@@ -336,7 +338,7 @@ DeckList::DeckList(const QString &nativeString)
 DeckList::~DeckList()
 {
     delete root;
-    
+
     QMapIterator<QString, SideboardPlan *> i(sideboardPlans);
     while (i.hasNext())
         delete i.next().value();
@@ -358,7 +360,7 @@ void DeckList::setCurrentSideboardPlan(const QList<MoveCard_ToZone> &plan)
         current = new SideboardPlan;
         sideboardPlans.insert(QString(), current);
     }
-    
+
     current->setMoveList(plan);
 }
 
@@ -394,7 +396,7 @@ void DeckList::write(QXmlStreamWriter *xml)
 
     for (int i = 0; i < root->size(); i++)
         root->at(i)->writeElement(xml);
-    
+
     QMapIterator<QString, SideboardPlan *> i(sideboardPlans);
     while (i.hasNext())
         i.next().value()->write(xml);
@@ -463,32 +465,84 @@ bool DeckList::saveToFile_Native(QIODevice *device)
     return true;
 }
 
+void DeckList::removeEmptyLinesFromStart(QVector<QString> &inputs)
+{
+    // https://stackoverflow.com/questions/4642339/trimming-a-vector-of-strings
+    inputs.erase(inputs.begin(), std::find_if(inputs.begin(), inputs.end(), [](const QString& s) {return !s.isEmpty();}));
+}
+
 bool DeckList::loadFromStream_Plain(QTextStream &in)
 {
     cleanList();
+    bool inSideboard = false, titleFound = false, isSideboard;
+    int okRows = 0, blankLines = 0;
+    QVector<QString> inputs;
 
-    bool inSideboard = false, isSideboard = false;
+    // Lets convert input QTextStream => QVector
+    while (!in.atEnd())
+        inputs.push_back(in.readLine().simplified().toLower());
 
-    int okRows = 0;
-    bool titleFound = false;
-    while (!in.atEnd()) {
-        QString line = in.readLine().simplified();
+    // "Trim" the inputs of all blank lines at the start & end of paste
+    removeEmptyLinesFromStart(inputs);
+    std::reverse(inputs.begin(), inputs.end()); // Reverse vector (to remove blanks from end)
+    removeEmptyLinesFromStart(inputs);
+    std::reverse(inputs.begin(), inputs.end()); // Vector now in correct order
 
-        // skip comments
+    // This will "trim" multiple blank lines in a row to just one blank
+    // Ex: ("Card1", "Card2", "", "", "", "Card3") => ("Card1", "Card2", "", "Card3")
+    while (true) {
+        auto i1 = std::adjacent_find(inputs.begin(), inputs.end());
+
+        if (i1 != inputs.end()) {
+            if (i1 != QString()) {
+                // So we have a duplicate row like (..., "1 card1", "3 card1", ...)
+                // To deal with this, we will combine these rows together to form (..., "4 card1", ...)
+                // And replace the newest row (aka we'd edit "3 card1" in our example) with the new value
+                int combinedTotal = (*i1).mid(0,1).toInt() + (*i1+1).mid(0,1).toInt();
+                inputs.replace(static_cast<int>(std::distance(inputs.begin(), i1+1)), QString::number(combinedTotal) + (*(i1+1)).mid(1));
+            }
+
+            inputs.erase(inputs.begin() + std::distance(inputs.begin(), i1));
+        }
+        else
+            break;
+    }
+
+    // See if sideboard is split by a single blank line
+    // Or if the word "Sideboard" appears, then we just skip this check
+    for (auto iter = inputs.begin(); iter != inputs.end(); iter++) {
+        QString line = *iter;
+
+        if (line.isEmpty())
+            blankLines++;
+
+        if (line.contains("Sideboard", Qt::CaseInsensitive)) {
+            blankLines = -1;
+            break;
+        }
+    }
+
+    for (auto iter = inputs.begin(); iter != inputs.end(); iter++)
+    {
+        QString line = *iter;
+
+        // This is a comment line, ignore it
         if (line.startsWith("//"))
         {
-            if(!titleFound)
-            {
+            if (!titleFound) { // Set the title to the first comment
                 name = line.mid(2).trimmed();
                 titleFound = true;
-            } else if(okRows == 0) {
+            } else if (okRows == 0) { // We haven't processed any cards yet
                 comments += line.mid(2).trimmed() + "\n";
             }
             continue;
         }
 
-        // check for sideboard prefix
-        if (line.startsWith("Sideboard", Qt::CaseInsensitive)) {
+        // If we have a blank line and it's the _ONLY_ blank line in the paste
+        // Then we assume it means to start the sideboard section of the paste.
+        // If we have the word "Sideboard" appear on any line, then that will
+        // also indicate the start of the sideboard.
+        if ((line.isEmpty() && blankLines == 1) || line.startsWith("Sideboard", Qt::CaseInsensitive)) {
             inSideboard = true;
             continue;
         }
@@ -505,19 +559,19 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         line.remove(rx);
         rx.setPattern("\\(.*\\)");
         line.remove(rx);
-        //Filter out post card name editions
+
+        // Filter out post card name editions
         rx.setPattern("\\|.*$");
         line.remove(rx);
-        line = line.simplified();
+        line = line.simplified().toLower();
 
         int i = line.indexOf(' ');
         int cardNameStart = i + 1;
 
         // If the count ends with an 'x', ignore it. For example,
         // "4x Storm Crow" will count 4 correctly.
-        if (i > 0 && line[i - 1] == 'x') {
+        if (i > 0 && line[i - 1] == 'x')
             i--;
-        }
 
         bool ok;
         int number = line.left(i).toInt(&ok);
@@ -525,7 +579,8 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
             continue;
 
         QString cardName = line.mid(cardNameStart);
-        // Common differences between cockatrice's card names
+
+        // Common differences between Cockatrice's card names
         // and what's commonly used in decklists
         rx.setPattern("’");
         cardName.replace(rx, "'");
@@ -537,20 +592,26 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         // Replace only if the ampersand is preceded by a non-capital letter,
         // as would happen with acronyms. So 'Fire & Ice' is replaced but not
         // 'R&D' or 'R & D'.
-        //
-        // Qt regexes don't support lookbehind so we capture and replace
-        // instead.
+        // Qt regexes don't support lookbehind so we capture and replace instead.
         rx.setPattern("([^A-Z])\\s*&\\s*");
-        if (rx.indexIn(cardName) != -1) {
+        if (rx.indexIn(cardName) != -1)
             cardName.replace(rx, QString("%1 // ").arg(rx.cap(1)));
-        }
 
-        // Look for the correct card zone
-        QString zoneName = getCardZoneFromName(cardName, isSideboard ? DECK_ZONE_SIDE: DECK_ZONE_MAIN);
+        // We need to get the name of the card from the database,
+        // but we can't do that until we get the "real" name
+        // (name stored in database for the card)
+        // and establish a card info that is of the card, then it's
+        // a simple getting the _real_ name of the card
+        // (i.e. "STOrm, CrOW" => "Storm Crow")
+        cardName = getCompleteCardName(cardName);
 
-        ++okRows;
+        // Look for the correct card zone of where to place the new card
+        QString zoneName = getCardZoneFromName(cardName, isSideboard ? DECK_ZONE_SIDE : DECK_ZONE_MAIN);
+
+        okRows++;
         new DecklistCardNode(cardName, number, getZoneObjFromName(zoneName));
     }
+
     updateDeckHash();
     return (okRows > 0);
 }
@@ -676,7 +737,7 @@ bool DeckList::deleteNode(AbstractDecklistNode *node, InnerDecklistNode *rootNod
         rootNode = root;
         updateHash = true;
     }
-    
+
     int index = rootNode->indexOf(node);
     if (index != -1) {
         delete rootNode->takeAt(index);
@@ -732,6 +793,6 @@ void DeckList::updateDeckHash()
                     + (((quint64) (unsigned char) deckHashArray[3]) << 8)
                     + (quint64) (unsigned char) deckHashArray[4];
     deckHash = (isValidDeckList) ? QString::number(number, 32).rightJustified(8, '0') : "INVALID";
-    
+
     emit deckHashChanged();
 }
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -498,18 +498,18 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
     }
 
     /*
-     * Removes all blank lines at end of inputs
-     * Ex: ("Card1", "Card2", "", "", "") => ("Card1", "Card2")
+     * Removes  blank line at end of inputs (if applicable)
+     * Ex: ("Card1", "Card2", "") => ("Card1", "Card2")
+     * NOTE: Any duplicates were taken care of above, so there can be
+     * at most one blank line at the very end
      */
-     for (auto iter = inputs.end()-1; iter > inputs.begin(); iter--)
-     {
-         if (*iter == QString())
-             inputs.erase(iter);
-         else
-             break;
-     }
+    if (inputs.last() == QString())
+    {
+        blankLines--;
+        inputs.erase(inputs.end() - 1);
+    }
 
-    // If "Sideboard" line appears, then blank lines mean nothing
+    // If "Sideboard" line appears in inputs, then blank lines mean nothing
     if (inputs.contains("sideboard"))
         blankLines = 2;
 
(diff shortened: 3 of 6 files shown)


common/decklist.cpp
```diff
@@ -4,6 +4,8 @@
 #include <QCryptographicHash>
 #include <QDebug>
 #include "decklist.h"
+#include "../cockatrice/src/carddatabase.h"
```
ctrlaltca: This can't work, since decklist.cpp is a common file used by both cockatrice and servatrice.
I recently had to face the same problem in this code to fix token loading (see #2670, commit 36f6907fa36b60bcc03e93ca08fd7b3e2da56533).
The idea is that since cockatrice's DeckLoader is a subclass of the common class DeckList, you can create a new virtual method in DeckList and reimplement it in DeckLoader in order to add the needed functionality.
In my example i added the getCardZoneFromName method to get the correct zone for a card from the db.
In your code, you could do the same to add a method that renames the card with its correct name.

ZeldaZach: Well that stinks, I'll try to figure it out.

common/decklist.h
```diff
@@ -125,6 +125,7 @@ class DeckList : public QObject {
     InnerDecklistNode *getZoneObjFromName(const QString  zoneName);
 protected:
     virtual QString getCardZoneFromName(QString /* cardName */, QString currentZoneName) { return currentZoneName; };
+    virtual QString getCompleteCardName(QString cardName) { return cardName; };
```
ZeldaZach: Oh go away, I used it.

Nightfirecat:
Side thought: If the codebase has a good place to implement tests, decklist parsing seems like a very good candidate for some unit tests, particularly with these new changes.

tooomm:
@MarkyMarkMcDonald added some tests recently. Want to have a look here as well?

ZeldaZach:
The newest change I just pushed addresses in the inputs of all separate types. Will create testing soon. I added significant comments to help explain what's going on.

1) Trim all blank lines from start/end of input
    - ["", "", "c", "", "d", ""] => ["c", "", "d"]
2) Trim excessive blanks from interior
    - ["c", "", "", "", "d", "e"] => ["c", "", "d", "e"]
3) Check if there is a _single_ blank line in the file, and if so use it as the split point for sideboard
4) If multiple blank lines found, look for "Sideboard" to appear for split point.

tooomm:
>If multiple blank lines found, look for "Sideboard" to appear for split point.

Your comment in the code is the other way around, check for "Sideboard" word, then blank lines.

You don't check for "SB:" anymore? Should we scrap that?

cockatrice/src/deck_loader.cpp
```diff
@@ -200,9 +200,19 @@ void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out, const InnerDecklis
...
 {
     CardInfo *card = db->getCard(cardName);
-    if(card && card->getIsToken())
+    if (card && card->getIsToken())
         return DECK_ZONE_TOKENS;
 
     return currentZoneName;
 }
 
+QString DeckLoader::getCompleteCardName(QString cardName)
```
backendr: Isnt this method name misleading? Why do you mutate the original?
`QString DeckLoader::getCompleteCardName(const QString cardName) const` may be a better way to write a getter; that or change the name?

ZeldaZach: I have to do this in order to access the database from the common folder. It's based on [Fabio's Comment](https://github.com/Cockatrice/Cockatrice/pull/2706#discussion_r115417049)... Please read all parts of the discussion

ZeldaZach:
>You don't check for "SB:" anymore? Should we scrap that?
@tooomm We still check for this, just I explicitly wrote down in my comment what I changed. Feel free to read the actual code

cockatrice/src/deck_loader.cpp
backendr: This doesn't explain why you wrote a mutating method that implies its a getter. Compared to `getCardZoneFromName` from the same header file which does not mutate parameters.

ctrlaltca: Well, @backendr is right, you should `return temp->getName();` if the card is found or `return cardName` if it-s not found, but not modify the contents of cardName itself.

ZeldaZach: Sounds good

backendr: To further enforce in the method signature your intentions, its best to change to `QString DeckLoader::getCompleteCardName(const QString cardName) const`

ZeldaZach: Sure @backendr. Changed

cockatrice/src/deck_loader.cpp
```diff
@@ -200,9 +200,19 @@ void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out, const InnerDecklis
...
 {
     CardInfo *card = db->getCard(cardName);
-    if(card && card->getIsToken())
+    if (card && card->getIsToken())
         return DECK_ZONE_TOKENS;
 
     return currentZoneName;
 }
 
+QString DeckLoader::getCompleteCardName(QString cardName) const
```
backendr: The passed in `QString cardName` can be made `const QString cardName`. This ensures that the passed in param is not mutated. 
The trailing method `const` only states this method will not mutate class state.

common/decklist.cpp
```diff
@@ -466,29 +468,68 @@ bool DeckList::saveToFile_Native(QIODevice *device)
 bool DeckList::loadFromStream_Plain(QTextStream &in)
 {
     cleanList();
+    bool inSideboard = false, titleFound = false, isSideboard;
+    int okRows = 0, blankLines = 0, linesToSkip = 0;
```
backendr: `linesToSkip` is never used

common/decklist.cpp
```diff
@@ -466,29 +468,74 @@ bool DeckList::saveToFile_Native(QIODevice *device)
 bool DeckList::loadFromStream_Plain(QTextStream &in)
 {
     cleanList();
+    bool inSideboard = false, titleFound = false, isSideboard;
+    int okRows = 0, blankLines = 0;
```
backendr: Its best to keep the usage of variables as close to their usage as possible. 
The bools and `okRows` can be moved to line `517`, `blankLines` to `505`

backendr:
Having an additional linebreak in either the MB or SB will result in a 75 card MB.

```
2 Aether Hub
2 Archangel Avacyn
2 Canyon Slough
4 Concealed Courtyard
1 Cut
4 Fatal Push
4 Gideon, Ally of Zendikar
4 Heart of Kiran
4 Inspiring Vantage
1 Mountain
4 Plains
4 Scrapheap Scrounger
2 Shambling Vent
4 Spire of Industry
2 Swamp
2 Thalia, Heretic Cathar
4 Thraben Inspector
4 Toolcraft Exemplar
4 Unlicensed Disintegration
2 Walking Ballista

1 Anguished Unmaking
1 Cut
1 Fumigate
1 Magma Spray
2 Nahiri, the Harbinger
2 Oath of Liliana
1 Painful Truths

2 Release the Gremlins
2 Sorin, Grim Nemesis
2 Transgress the Mind

```

common/decklist.cpp
```diff
@@ -537,20 +585,25 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
...
         rx.setPattern("([^A-Z])\\s*&\\s*");
-        if (rx.indexIn(cardName) != -1) {
+        if (rx.indexIn(cardName) != -1)
             cardName.replace(rx, QString("%1 // ").arg(rx.cap(1)));
-        }
 
-        // Look for the correct card zone
-        QString zoneName = getCardZoneFromName(cardName, isSideboard ? DECK_ZONE_SIDE: DECK_ZONE_MAIN);
+        // We need to get the name of the card from the database,
+        // but we can't do that until we sanitize the name
```
backendr: when does sanitization happening? I cant see a usage of it in `getCompleteCardName` or its stack.

common/decklist.h
```diff
@@ -125,6 +125,7 @@ class DeckList : public QObject {
     InnerDecklistNode *getZoneObjFromName(const QString  zoneName);
 protected:
     virtual QString getCardZoneFromName(QString /* cardName */, QString currentZoneName) { return currentZoneName; };
+    virtual QString getCompleteCardName(QString cardName) const { return cardName; };
```
backendr: `virtual QString getCompleteCardName(const QString cardName) const { return cardName; };`

ZeldaZach:
>Having an additional linebreak in either the MB or SB will result in a 75 card MB.

Expected behavior. I'm trying to make the parser smarter in a sense as it didn't account for blank lines in the past for a single line, but if you have multiple blank lines it's not my fault and you'll have to make your deck list more concise. I have no way to appropriately parse that information if there are blank lines like that everywhere.

common/decklist.cpp
ZeldaZach: Bad wording, fixed

common/decklist.h
ZeldaZach: 👍

common/decklist.cpp
ZeldaZach: I'm fine leaving it where it is

cockatrice/src/deck_loader.cpp
ZeldaZach: 👍

ZeldaZach:
So it seems like my changes fail travis due to the function `QVector::rbegin()` being introduced in QT 5.6 and we compile against 5.0.3 and 5.4 (if I recall). @ctrlaltca Any alternatives you can suggest?

Daenyth:
We should generalize the Dockerfile for servatrice to be able to compile and package everything outputting to a volume, then it doesn't matter how crappy travis's build architecture is

ctrlaltca:
> the function QVector::rbegin() being introduced in QT 5.6

No reverse iterators before 5.6, sorry. You'll have to modify line https://github.com/Cockatrice/Cockatrice/pull/2706/files#diff-166b29bd05b87f762761b55ce34db82dR481 to use a backwards `for` loop or something similar.

ZeldaZach:
@ctrlaltca Is there a way to require a later QT? Can we somehow bundle it?

common/decklist.cpp
backendr: Why?

ctrlaltca:
Requiring 5.6 for the whole project would mean cutting out some linux distributions.
You could `#ifdef` the code to be used only if at least Qt 5.6 is found, but probably rewriting it is a easier solution.

ZeldaZach:
I created some new tests for the loader. If anyone else has some suggestions please let me know!

common/decklist.cpp
ctrlaltca: It's quite a minor detail, but declaring variables just before their use eases to read and understand the code.
Sometimes creating scopes and declaring variables inside the new scope helps even more, because you'll be sure that nothing outside the scope can touch and modify the variable.
Again, it's quite a minor thing and cockatrice has probably a lot of places where code could be made easier to understand.

common/decklist.cpp
```diff
@@ -463,32 +465,84 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+
+    // Lets convert input QTextStream => QVector
+    while (!in.atEnd())
+        inputs.push_back(in.readLine().simplified().toLower());
+
+    // "Trim" the inputs of all blank lines at the start & end of paste
+    removeEmptyLinesFromStart(inputs);
+    std::reverse(inputs.begin(), inputs.end()); // Reverse vector (to remove blanks from end)
+    removeEmptyLinesFromStart(inputs);
+    std::reverse(inputs.begin(), inputs.end()); // Vector now in correct order
```
ctrlaltca: Lines 482-489 and the `removeEmptyLinesFromStart` method looks over engineered to me.
Empty lines at start/end could just be dropped inside the first loop, and even multiple empty lines could be condensed in the same loop, and the remaining empty lines counted.
This would avoid the need of add all the lines them to the vector, reverse the vector twice, call the obscure std-iterators-with-callback-loop-whatever at line 471 and them loop the vector again at line 493, and again at line 513.
Anyway, as long as it works.. Yeah, well, that's just, like, my opinion, man (cit.)

common/decklist.cpp
```diff
@@ -463,32 +465,84 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+
+    // This will "trim" multiple blank lines in a row to just one blank
```
ctrlaltca: I usually see the english term "condense" for such operations, maybe it fits better?

common/decklist.cpp
```diff
@@ -463,32 +465,84 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+    while (true) {
+        auto i1 = std::adjacent_find(inputs.begin(), inputs.end());
+
+        if (i1 != inputs.end()) {
+            if (i1 != QString()) {
+                // So we have a duplicate row like (..., "1 card1", "3 card1", ...)
+                // To deal with this, we will combine these rows together to form (..., "4 card1", ...)
+                // And replace the newest row (aka we'd edit "3 card1" in our example) with the new value
+                int combinedTotal = (*i1).mid(0,1).toInt() + (*i1+1).mid(0,1).toInt();
+                inputs.replace(static_cast<int>(std::distance(inputs.begin(), i1+1)), QString::number(combinedTotal) + (*(i1+1)).mid(1));
```
ctrlaltca: Is there a good reason to do this? From the comment above this loop was meant to only delete double empty lines.
Also, what happens if my deck does't contain card numbers, or has more that 9 cards (that is, a number with more than a single digit) with the same name?
Example 1:
```
Kitten
Kitten
Kitten
Kitten
```

Example 2:
```
12 Kitten
13 Kitten
```

common/decklist.cpp
```diff
@@ -463,32 +465,84 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+    // Or if the word "Sideboard" appears, then we just skip this check
+    for (auto iter = inputs.begin(); iter != inputs.end(); iter++) {
+        QString line = *iter;
 
-    int okRows = 0;
-    bool titleFound = false;
-    while (!in.atEnd()) {
-        QString line = in.readLine().simplified();
+        if (line.isEmpty())
+            blankLines++;
```
ctrlaltca: Yet another loop :(

common/decklist.cpp
ZeldaZach: I can revoke this part as I just added it for personal adds

common/decklist.cpp
backendr: This is what McConnell says in his must-read book Code Complete 2:

Ideally, declare and define each variable close to where it’s used. A declaration establishes a variable’s type. A definition assigns the variable a specific value. In languages that support it, such as C++ and Java, variables should be declared and defined close to where they are first used.

He also recommends keeping variables alive for as short a time as possible, and to minimize scope.

ZeldaZach:
Ok I made another set of changes. I hopefully increased the efficiency of the program. Personally, the way I see it, this piece of code is run every once in a while, one time, by a user. It doesn't have to be the _most_ efficient thing in the client.

common/decklist.cpp
```diff
@@ -466,29 +468,74 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
     int okRows = 0;
-    bool titleFound = false;
-    while (!in.atEnd()) {
-        QString line = in.readLine().simplified();
+    for (auto iter = inputs.begin(); iter != inputs.end(); iter++)
+    {
+        QString line = *iter;
 
-        // skip comments
+        // This is a comment line, ignore it
```
backendr: Can be removed

cockatrice/src/chatview/chatview.cpp
```diff
@@ -93,7 +93,7 @@ void ChatView::appendCardTag(QTextCursor &cursor, const QString &cardName)
     anchorFormat.setAnchor(true);
     anchorFormat.setAnchorHref("card://" + cardName);
     anchorFormat.setFontItalic(true);
-    
+
```
ctrlaltca: Unrelated change

common/decklist.cpp
```diff
@@ -466,31 +468,76 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+        QString line = in.readLine().simplified().toLower();
+
+        /*
+         * Removes all blank lines at start of inputs
+         * Ex: ("", "", "", "Card1", "Card2") => ("Card1", "Card2")
+         *
+         * This will also concise multiple blank lines in a row to just one blank
+         * Ex: ("Card1", "Card2", "", "", "", "Card3") => ("Card1", "Card2", "", "Card3")
+         */
+        if (line == QString()) {
```
ctrlaltca: `if(line.isEmpty()) {`

common/decklist.cpp
```diff
@@ -466,31 +468,76 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+        inputs.push_back(line);
+    }
+
+    /*
+     * Removes  blank line at end of inputs (if applicable)
+     * Ex: ("Card1", "Card2", "") => ("Card1", "Card2")
+     * NOTE: Any duplicates were taken care of above, so there can be
+     * at most one blank line at the very end
+     */
+    if (inputs.size() && inputs.last() == QString())
```
ctrlaltca: `.. && inputs.last().isEmpty()`

common/decklist.cpp
```diff
@@ -466,31 +468,76 @@ bool DeckList::saveToFile_Native(QIODevice *device)
...
+        blankLines = 2;
 
+    bool inSideboard = false, titleFound = false, isSideboard;
     int okRows = 0;
-    bool titleFound = false;
-    while (!in.atEnd()) {
-        QString line = in.readLine().simplified();
+    for (auto iter = inputs.begin(); iter != inputs.end(); iter++)
+    {
+        QString line = *iter;
```
ctrlaltca: Alternative: `foreach(QString line, inputs) {`

common/decklist.cpp
```diff
@@ -500,32 +547,36 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
             isSideboard = true;
```
ctrlaltca: 2 lines above:
The line content is now surely lowercase, so: `if(line.startsWith("sb:")) {`

common/decklist.cpp
```diff
@@ -500,32 +547,36 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
...
         }
 
+        if (line.trimmed().isEmpty())
+            continue; // The line was "    " instead of "\n"
```
ctrlaltca: Can you explain this a little bit more? Is it a check for empty SB: lines like
```
SB:       
```
Isn't line supposed to have already be trimmed by `simplified()` above?

common/decklist.cpp
```diff
@@ -500,32 +547,36 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
...
         line.remove(rx);
         rx.setPattern("\\(.*\\)");
         line.remove(rx);
-        //Filter out post card name editions
+
+        // Filter out post card name editions
         rx.setPattern("\\|.*$");
         line.remove(rx);
-        line = line.simplified();
+        line = line.simplified().toLower();
```
ctrlaltca: Isn't line supposed to have already be `simplified()` and `toLower()`-ed above?

cockatrice/src/chatview/chatview.cpp
ZeldaZach: No idea how to undo this, it's just spacing anyway, doesn't rlly matter

common/decklist.cpp
ZeldaZach: While testing, for some reason entries as "______" started appearing randomly, and this stops it. There were no underscores in the inputs, and it was interpreting extra blank lines as this format for some reason. This just prevents that issue.

common/decklist.cpp
ZeldaZach: Yeah...

cockatrice/src/chatview/chatview.cpp
ctrlaltca: If this were a big project with a dozen people working on it, such a change would be a problem:
 1. because it could introduce merge conflicts
 2. because the git log will store anyway the fact that you are the author that  last changed that line of code (think of git blame, etc..)

In our case I agree it's not important, you can leave it as it is.

common/decklist.cpp
ctrlaltca: Ok, so it's a sanity check on the line contents before starting to parse card name and qty.

cockatrice/src/chatview/chatview.cpp
ZeldaZach: Thanks for the tip!

common/decklist.cpp
ZeldaZach: dealt with

common/decklist.cpp
ZeldaZach: Yes

cockatrice/src/chatview/chatview.cpp
Daenyth: Tip: When making changes, use `git add -p` to prepare your commit diff by diff. That way you can sculpt specific commits, and do the behavior changes in one, and formatting/refactoring in another

cockatrice/src/deck_loader.cpp
```diff
@@ -200,9 +200,19 @@ void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out, const InnerDecklis
 QString DeckLoader::getCardZoneFromName(QString cardName, QString currentZoneName)
 {
     CardInfo *card = db->getCard(cardName);
-    if(card && card->getIsToken())
+    if (card && card->getIsToken())
```
Daenyth: I really need to get clang-format set up. It's not hard to run, the hard part is just tuning its configuration flags to be as close as possible to current style, so as not to rewrite the whole codebase

common/decklist.cpp
```diff
@@ -466,66 +466,111 @@ bool DeckList::saveToFile_Native(QIODevice *device)
 bool DeckList::loadFromStream_Plain(QTextStream &in)
 {
     cleanList();
+    QVector<QString> inputs; // QTextStream -> QVector
 
-    bool inSideboard = false, isSideboard = false;
+    bool priorEntryIsBlank = true, isAtBeginning = true;
```
Daenyth: For the future let's not do same-line declarations

common/decklist.cpp
```diff
@@ -466,66 +466,111 @@ bool DeckList::saveToFile_Native(QIODevice *device)
 bool DeckList::loadFromStream_Plain(QTextStream &in)
```
Daenyth: This function is massive, it would be good to break it up into smaller units that are more easily composed and tested in isolation

tests/loading_from_clipboard/loading_from_clipboard_test.cpp
```diff
@@ -39,28 +39,33 @@ struct DecklistBuilder {
...
         DeckList *deckList = fromClipboard(new QString("Sideboard"));
-        ASSERT_TRUE(deckList->getCardList().isEmpty()) << "Deck should be empty";
+        ASSERT_TRUE(deckList->getCardList().isEmpty());
     }
 
     TEST(LoadingFromClipboardTest, QuantityPrefixed) {
         QString *clipboard = new QString(
                 "1 Mountain\n"
-                "2x Island\n"
+                        "2x Island\n"
```
Daenyth: Extra spacing

tests/loading_from_clipboard/loading_from_clipboard_test.cpp
```diff
@@ -114,12 +123,142 @@ namespace {
         DecklistBuilder decklistBuilder = DecklistBuilder();
         deckList->forEachCard(decklistBuilder);
 
-        CardRows expectedMainboard = CardRows({{"CardThatDoesNotExistInCardsXml", 1}});
+        CardRows expectedMainboard = CardRows({
+                                                      {"cardthatdoesnotexistincardsxml", 1}
```
Daenyth: what's with the spacing here?

common/decklist.cpp
ZeldaZach: Future endeavors

cockatrice/src/deck_loader.cpp
ZeldaZach: Yeah....

common/decklist.cpp
ZeldaZach: Isn't it more efficient to do so?

tests/loading_from_clipboard/loading_from_clipboard_test.cpp
ZeldaZach: CLion did it and i couldn't change it no matter what I did... maybe you can help?

common/decklist.cpp
backendr: Its not inefficient, though it can be less readable, as one line is doing more than one thing. You gain saving a line over readability. I prefer readability.

common/decklist.cpp
backendr: Do you have a timescale on when? Its better to "eat the frog" now than procrastinate on it. Its too easy to get a PR merged and either not come back to it, or leave more work for someone else.

tests/loading_from_clipboard/loading_from_clipboard_test.cpp
backendr: Try editing the file in your terminal

common/decklist.cpp
ZeldaZach: I don't plan on coming back to this personally, it'll be someone else small project

tests/loading_from_clipboard/loading_from_clipboard_test.cpp
ZeldaZach: That's only a temporary fix, there's probably a CLion configuration that I can edit just not sure which

common/decklist.cpp
Daenyth: It's strictly a readability thing

Daenyth:
Is this waiting on anything? Looks good to go

ctrlaltca:
This is good to go, except for the extra spacing you noticed.

tooomm:
And https://github.com/Cockatrice/Cockatrice/pull/2706#discussion_r116356241 maybe?

PR #2709, opened 2017-05-09
Message Log Widget changes

backendr:
- removed unused imports
- alphabetised methods
- refactoring
	- code reads down in list format, easier to read, less likely to get lost, tidy
           our eyes are more comfortable when reading tall and narrow columns of text
- removed pass by ref, `QString` uses implicit sharing / C-O-W
- removed logging when moving card from ~`grave` -> `grave`, and~ `rfg` -> `rfg`
- removed join game logging
- construct on first use idiom for const zone names
- conceded/connection lost/connection restored now bold

screen shot 2017-05-09 at 13 41 12

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -291,8 +291,7 @@ void MessageLogWidget::logDoMoveCard(LogMoveCard &lmc)
     // do not log if moved within the same zone
     if ((startZone == tableConstant() && targetZone == tableConstant() && !ownerChanged)
             || (startZone == handConstant() && targetZone == handConstant())
-               || (startZone == exileConstant() && targetZone == exileConstant())
-                  || (startZone == graveyardConstant() && targetZone == graveyardConstant()))
+               || (startZone == exileConstant() && targetZone == exileConstant()))
         return;
 
     QString cardName = lmc.cardName;
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -508,7 +508,7 @@ void MessageLogWidget::logRevealCards(
 
     if (cardId == -1) {
         if (otherPlayer)
-            appendHtmlServerMessage(QString(tr("%1 reveals %2 to %3."))
+            appendHtmlServerMessage(tr("%1 reveals %2 to %3.")
                                             .arg(sanitizeHtml(player->getName()))
                                             .arg(zone->getTranslatedName(true, CaseRevealZone))
                                             .arg(sanitizeHtml(otherPlayer->getName())));
@@ -518,7 +518,7 @@ void MessageLogWidget::logRevealCards(
                                             .arg(zone->getTranslatedName(true, CaseRevealZone)));
     } else if (cardId == -2) {
         if (otherPlayer)
-            appendHtmlServerMessage(QString(tr("%1 randomly reveals %2%3 to %4."))
+            appendHtmlServerMessage(tr("%1 randomly reveals %2%3 to %4.")
                                             .arg(sanitizeHtml(player->getName()))
                                             .arg(cardStr)
                                             .arg(fromStr)
@@ -531,16 +531,16 @@ void MessageLogWidget::logRevealCards(
     } else {
         if (faceDown && player == otherPlayer) {
             if (cardName.isEmpty())
-                appendHtmlServerMessage(QString(tr("%1 peeks at face down card #%2."))
+                appendHtmlServerMessage(tr("%1 peeks at face down card #%2.")
                                                 .arg(sanitizeHtml(player->getName()))
                                                 .arg(cardId));
             else
-                appendHtmlServerMessage(QString(tr("%1 peeks at face down card #%2: %3."))
+                appendHtmlServerMessage(tr("%1 peeks at face down card #%2: %3.")
                                                 .arg(sanitizeHtml(player->getName()))
                                                 .arg(cardId)
                                                 .arg(cardStr));
         } else if (otherPlayer)
-            appendHtmlServerMessage(QString(tr("%1 reveals %2%3 to %4."))
+            appendHtmlServerMessage(tr("%1 reveals %2%3 to %4.")
                                             .arg(sanitizeHtml(player->getName()))
                                             .arg(cardStr)
                                             .arg(fromStr)
@@ -727,7 +727,7 @@ void MessageLogWidget::logSetPT(Player *player, CardItem *card, QString newPT)
     if (currentContext == MessageContext_MoveCard)
         moveCardPT.insert(card, newPT);
     else
-        appendHtmlServerMessage(QString(tr("%1 sets PT of %2 to %3."))
+        appendHtmlServerMessage(tr("%1 sets PT of %2 to %3.")
                                         .arg(sanitizeHtml(player->getName()))
                                         .arg(cardLink(card->getName()))
                                         .arg(QString("<font color=\"blue\">%1</font>")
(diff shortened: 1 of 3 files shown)


tooomm:
>removed logging when moving card from `grave` -> `grave`, and `rfg` -> `rfg`

👍 

>removed join game logging

The game# will still show on start of replays?

>conceded/connection lost/connection restored now bold

Do you have a picture handy for the new bold messages?

Daenyth:
The graveyard is an ordered zone, you need to log when people change the
ordering of cards.

tooomm:
True for legacy formats... http://mtg.gamepedia.com/Graveyard#Order_of_cards_in_the_graveyard

backendr:
@tooomm forgot to add one. Added to initial message.

backendr:
@Daenyth added graveyard logging back

cockatrice/src/messagelogwidget.cpp
```diff
@@ -3,190 +3,87 @@
...
+        int position,
+        bool ownerChange) const
 {
     bool cardNameContainsStartZone = false;
     QString fromStr;
-    QString startName = zone->getName();
+    QString zoneName = zone->getName();
 
-    if (startName == "table")
+    if (zoneName == tableConstant())
```
ZeldaZach: Instead of declaring `tableConstant()` and the other constants as functions, define them as full cap constants at the top of the file maybe?
```
#define TABLE "table"
#define GRAVEYARD "grave"
etc
```

cockatrice/src/messagelogwidget.cpp
```diff
@@ -219,278 +116,379 @@ QPair<QString, QString> MessageLogWidget::getFromStr(CardZone *zone, QString car
...
+
+void MessageLogWidget::logDoMoveCard(LogMoveCard &lmc)
+{
+    QString startZone = lmc.startZone->getName();
+    QString targetZone = lmc.targetZone->getName();
+    bool ownerChanged = lmc.startZone->getPlayer() != lmc.targetZone->getPlayer();
+
+    // do not log if moved within the same zone
+    if ((startZone == tableConstant() && targetZone == tableConstant() && !ownerChanged)
+            || (startZone == handConstant() && targetZone == handConstant())
```
ZeldaZach: owner changed still might be a thing here, as I _could_ drag a card from my hand to my opponents hand (no idea why, but some games might do that?)

cockatrice/src/messagelogwidget.cpp
backendr: http://stackoverflow.com/questions/4103086/defining-constants-in-c

cockatrice/src/messagelogwidget.cpp
```diff
@@ -508,139 +506,357 @@ void MessageLogWidget::logRevealCards(Player *player, CardZone *zone, int cardId
...
-                    str = tr("%1 reveals %2 to %3.");
-            appendHtmlServerMessage(str.arg(sanitizeHtml(player->getName())).arg(zone->getTranslatedName(true, CaseRevealZone)).arg(sanitizeHtml(otherPlayer->getName())));
-        } else {
-            appendHtmlServerMessage((
-              tr("%1 reveals %2.")
-            ).arg(sanitizeHtml(player->getName()))
-             .arg(zone->getTranslatedName(true, CaseRevealZone)));
-        }
+        if (otherPlayer)
+            appendHtmlServerMessage(QString(tr("%1 reveals %2 to %3."))
```
ZeldaZach: TR will return a QString itself, no need to wrap it (https://doc.qt.io/qt-5/qobject.html#tr)

cockatrice/src/messagelogwidget.cpp
backendr: This is not currently possible in the master branch.

cockatrice/src/messagelogwidget.cpp
```diff
@@ -508,139 +506,357 @@ void MessageLogWidget::logRevealCards(Player *player, CardZone *zone, int cardId
...
-    connect(player, SIGNAL(logAttachCard(Player *, QString, Player *, QString)), this, SLOT(logAttachCard(Player *, QString, Player *, QString)));
-    connect(player, SIGNAL(logUnattachCard(Player *, QString)), this, SLOT(logUnattachCard(Player *, QString)));
-    connect(player, SIGNAL(logDumpZone(Player *, CardZone *, int)), this, SLOT(logDumpZone(Player *, CardZone *, int)));
-    connect(player, SIGNAL(logStopDumpZone(Player *, CardZone *)), this, SLOT(logStopDumpZone(Player *, CardZone *)));
-    connect(player, SIGNAL(logDrawCards(Player *, int)), this, SLOT(logDrawCards(Player *, int)));
-    connect(player, SIGNAL(logUndoDraw(Player *, QString)), this, SLOT(logUndoDraw(Player *, QString)));
-    connect(player, SIGNAL(logRevealCards(Player *, CardZone *, int, QString, Player *, bool)), this, SLOT(logRevealCards(Player *, CardZone *, int, QString, Player *, bool)));
-    connect(player, SIGNAL(logAlwaysRevealTopCard(Player *, CardZone *, bool)), this, SLOT(logAlwaysRevealTopCard(Player *, CardZone *, bool)));
+    connect(player, SIGNAL(logSay(Player *, QString)),
```
ZeldaZach: I kind of liked how the connects were all in line, feels like waste of space now

cockatrice/src/messagelogwidget.cpp
ZeldaZach: In the header file we can put them as constant public variables. I've never seen it done as a function like this before and it just feels a bit off

cockatrice/src/messagelogwidget.cpp
ZeldaZach: Darn, a feature to be done in the future

cockatrice/src/messagelogwidget.cpp
backendr: This is the construct on first use idiom.
http://stackoverflow.com/questions/335369/finding-c-static-initialization-order-problems

I dont feel adding "global" consts here is a good idea as you suggest. Instead it might be better to start a `constants.h/cpp` that can be accessed from anywhere.

tooomm:
Can you please combine your code comments in one review @ZeldaZach ?
You can also just use the review to comment, in case you don't want to request changes or approve things.

I just got hammered with mails :)

cockatrice/src/messagelogwidget.cpp
ZeldaZach: I'd be fine with a global constants file somewhere

backendr: OK. That is not part of this PR though.

cockatrice/src/messagelogwidget.cpp
```diff
@@ -508,139 +506,357 @@ void MessageLogWidget::logRevealCards(Player *player, CardZone *zone, int cardId
...
+    appendHtmlServerMessage(str.arg(sanitizeHtml(player->getName()))
+                                    .arg(cardLink(card->getName())));
+}
+
+void MessageLogWidget::logSetPT(Player *player, CardItem *card, QString newPT)
+{
+    if (currentContext == MessageContext_MoveCard)
+        moveCardPT.insert(card, newPT);
+    else
+        appendHtmlServerMessage(QString(tr("%1 sets PT of %2 to %3."))
```
ctrlaltca: QString() around tr() can be removed.

tooomm:
Since we have now highights for several (more) important messages...
I feel like we missed `deck load` and should also print them in bold.

PR #2711, opened 2017-05-09
limit gitlab builds to master

skwerlman:
## Related Ticket(s)
- N/A

Since getting Gitlab CI set up, my instance has built ~400 binaries, many of which were for small changes outside the master branch, which is essentially wasted work.

- Gitlab CI builds will only trigger for changes on the master branch
- My power bill will go down

## Screenshots
2017-05-09-171217_1329x281_scrot

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,6 +13,15 @@ stages:
     paths:
       - cache/
 
+.branches: &branches
+  only:
+    - master
+
+.tags: &tags
+  tags:
+    - linux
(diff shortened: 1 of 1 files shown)


tooomm:
Perfectly fine with that, I would even say it might be sufficient if you build on (pre)release tags only for example.

On demand we can provide others if needed.

skwerlman:
I can make it build debug builds for all commits on master, and release builds for the tags if y'all'd prefer.

PR #2714, opened 2017-05-12
counter menu context sensitive

### LINKED ISSUE #2677: Disable "Remove counter" context item when not useful
I feel that the counters menu can be more context based and only present options based on the current card state.

For example; disable, or even remove, the `Remove Counter(Colour)` if that counter is not on the card. 

screen shot 2017-05-01 at 09 50 53

backendr:
## Related Ticket(s)
- Fixes #2677

When right clicking on a card, the "remove counters" option would show, even when no counters were on the card.

Context menus for cards are created at the same time the card comes into play. This update will update when the context menu opens.
- Added `updateCardMenu` call when opening context menu on a single card
- Context menu will only add "remove counters" option if a counter of that type is on the card

## Screenshots

screen shot 2017-05-13 at 00 05 53

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -341,8 +341,10 @@ void CardItem::mouseReleaseEvent(QGraphicsSceneMouseEvent *event)
 {
     if (event->button() == Qt::RightButton) {
         if (cardMenu)
-            if (!cardMenu->isEmpty())
+            if (!cardMenu->isEmpty()) {
+                owner->updateCardMenu(this);
                 cardMenu->exec(event->screenPos());
+            }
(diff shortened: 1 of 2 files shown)


backendr:
Caveat:
Mass selecting cards with counters on them, and then opening the menu on a card _without_ that counter will not give you the option to remove the counter, as the context menu only cards about the current card you are selecting.

tooomm:
Why don't we just grey out the right now not useful menu options instead and stay in line with #2655?
I also prefer that because the right click menu wouldn't change in size and position of options, which players are used to. Makes it easier and more uniform to find the wanted options.

Daenyth:
Yeah, I do think greying out would be better than removing entirely.

PR #2716, opened 2017-05-13
Better scrolling in reveal zones

backendr:
Scrolling in a "zone" has always been painfully slow.

- Removed unused imports
- Better scroll bar height and stepping
    - Faster and less scrolling

## Screenshots

screen shot 2017-05-13 at 22 33 30


ZeldaZach:
Nice change 👍 

Since you're working on the Reveal Zones section, maybe you can also add better resizing. I can resize horizontally (only to shrink it) but vertically doesn't seem to want to resize at all.

PR #2718, opened 2017-05-14
[WIP] string reworks

tooomm:
- outdated update description
- `\n` at start of translation strings
- `%1` in translation strings
- possible html injections, because of `<br>` tags in messages from the server

- correct update description
- clean-up translation strings, split up text and formatting
- easier for translators
- better string readability in the code
- always use `\n` instead of html (`<br>`) to prevent possible code injection (if no other html is needed + wanted anyway)


cockatrice/src/window_main.cpp
```diff
@@ -337,7 +337,7 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
             formattedMissingFeatures = "Missing Features: ";
```
tooomm: How do I add the missing tr() here?

`formattedMissingFeatures = "tr("Missing Features:) + QString:(" ")";` ?

cockatrice/src/window_main.cpp
```diff
@@ -107,12 +107,15 @@ void MainWindow::processConnectionClosedEvent(const Event_ConnectionClosed &even
...
+    QMessageBox::critical(this, tr("Connection closed"), tr("The server has terminated your connection.") + QString("<br>")
+                                                        +tr ("Reason:") + QString(" %1").arg(reasonStr));
 }
 
 void MainWindow::processServerShutdownEvent(const Event_ServerShutdown &event)
 {
-    serverShutdownMessageBox.setInformativeText(tr("The server is going to be restarted in %n minute(s).\nAll running games will be lost.\nReason for shutdown: %1", "", event.minutes()).arg(QString::fromStdString(event.reason())));
+    serverShutdownMessageBox.setInformativeText(tr("The server is going to be restarted. All running games will be lost.") + QString("<br><br>")
+                                                + tr("Reason for shutdown:") + QString(" %1<br>").arg(QString::fromStdString(event.reason()))
+                                                + tr("Time until shutdown:") + QString(" %n ", "", event.minutes()) + tr("minutes"));
```
tooomm: Can somebody double-check this please?
Not sure with the `""` and `%n event` here.

cockatrice/src/window_main.cpp
```diff
@@ -494,16 +499,20 @@ void MainWindow::activateError()
...
     actConnect();
 }
 
 void MainWindow::protocolVersionMismatch(int localVersion, int remoteVersion)
 {
     if (localVersion > remoteVersion)
-        QMessageBox::critical(this, tr("Error"), tr("You are trying to connect to an obsolete server. Please downgrade your Cockatrice version or connect to a suitable server.\nLocal version is %1, remote version is %2.").arg(localVersion).arg(remoteVersion));
+        QMessageBox::critical(this, tr("Error"), tr("You are trying to connect to an obsolete server. Please downgrade your Cockatrice version or connect to a suitable server." + QString("<br>")
+                                                    + tr("Local version:") + QString(" %1<br>").arg(localVersion)
+                                                    + tr("Remote version:") + QString(" %1").arg(remoteVersion));
```
tooomm: `Local version` is the version of the client --> Cockatrice?
`Remote version` is the version the server is running --> Servatrice?

So go with `Server version:` for clarity?

cockatrice/src/window_main.cpp
```diff
@@ -107,12 +107,15 @@ void MainWindow::processConnectionClosedEvent(const Event_ConnectionClosed &even
         case Event_ConnectionClosed::LOGGEDINELSEWERE: reasonStr = tr("You have been logged out due to logging in at another location."); break;
         default: reasonStr = QString::fromStdString(event.reason_str());
     }
-    QMessageBox::critical(this, tr("Connection closed"), tr("The server has terminated your connection.\nReason: %1").arg(reasonStr));
+    QMessageBox::critical(this, tr("Connection closed"), tr("The server has terminated your connection.") + QString("<br>")
+                                                        +tr ("Reason:") + QString(" %1").arg(reasonStr));
```
Daenyth: `QString("%1").arg(reasonStr)` is redundant, `reasonStr` is already a `QString`

cockatrice/src/window_main.cpp
Daenyth: That looks about right

cockatrice/src/window_main.cpp
```diff
@@ -375,10 +378,12 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
             }
             break;
         case Response::RespClientIdRequired:
-            QMessageBox::critical(this, tr("Error"), tr("This server requires client ID's. Your client is either failing to generate an ID or you are running a modified client.\nPlease close and reopen your client to try again."));
+            QMessageBox::critical(this, tr("Error"), tr("This server requires client ID's. Your client is either failing to generate an ID or you are running a modified client.\n"
```
Daenyth: Why are you splitting these? It helps translators when it's one thing in context. They should also control the spacing and line breaks since the lines can be very different lengths from english.

ctrlaltca:
While i see the good motivations on trying to remove format chars from the translated strings, please remember that not every language follow the same syntax rules, so forcing a specific part of the string to be in a specific position can creates problems for translators, example:

> EN: You have %1 candies left.
> IT: Ti rimangono %1 caramelle.
> JP: %1個のキャンディーが残っている (hopefully right)

If you break the english sentence to `tr("You have") + "%1" + tr("candies left")`, you make it impossible for translators to do their job without using workarounds.

tooomm:
@ctrlaltca I think I avoided that because I used less real sentences (which always bring problems to translations), and used statements instead, like:
`What:` `How much`, where how much is the variable translators don't have to deal with any longer.
For the most part I just splittet `a: %1` into "`a:` and the variable" or created several single strings out of a big complex one. So no problem there.

Can you point me to problematic spots that I might missed?

cockatrice/src/window_main.cpp
tooomm: It's the same tr() wrapping.^^
So it's still one string for translators, but it's easier to read in the code.
Saw it at another place in the code and liked it.

cockatrice/src/window_main.cpp
ctrlaltca: [QObject::tr()](http://doc.qt.io/qt-5/qobject.html#tr) supports disambiguation strings ("") and different strings to handle plurals (%n). You can just use %1 in your refactoring.

cockatrice/src/window_main.cpp
```diff
@@ -95,7 +95,7 @@ void MainWindow::processConnectionClosedEvent(const Event_ConnectionClosed &even
         case Event_ConnectionClosed::BANNED: {
             reasonStr = tr("Banned by moderator");
             if (event.has_end_time())
-                reasonStr.append("\n" + tr("Expected end time: %1").arg(QDateTime::fromTime_t(event.end_time()).toString()));
+                reasonStr.append("\n" + tr("Expected end time:" + QString(" %1").arg(QDateTime::fromTime_t(event.end_time()).toString()));
```
ctrlaltca: `QString(" %1").arg(....toString())` is redundant

cockatrice/src/window_main.cpp
```diff
@@ -337,7 +340,7 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
             formattedMissingFeatures = "Missing Features: ";
             for (int i = 0; i < missingFeatures.size(); ++i)
                 formattedMissingFeatures.append(QString("\n     %1").arg(QChar(0x2022)) + " " + missingFeatures.value(i)   );
-            formattedMissingFeatures.append("\nTo update your client, go to Help -> Check for Updates.");
+            formattedMissingFeatures.append(QString("<br>") + tr("To update your client, go to \"Help -> Check for Client Updates\"."));
```
ctrlaltca: This message was using `\n` to avoid its text being rendered as html.
Since the popup contains text coming from the server, rendering it as html could be problematic if you don-t escape it. Think of a malicious server admin adding a new bogus "feature" using html to inject code into user's popups.
Please have a look at http://doc.qt.io/qt-5.8/qmessagebox.html#rich-text-and-the-text-format-property

cockatrice/src/window_main.cpp
```diff
@@ -396,44 +401,44 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
             break;
         }
         default:
-            QMessageBox::critical(this, tr("Error"), tr("Unknown login error: %1").arg(static_cast<int>(r)) + tr("\nThis usually means that your client version is out of date, and the server sent a reply your client doesn't understand."));
+            QMessageBox::critical(this, tr("Error"), tr("Unknown login error:") + QString(" %1<br>").arg(static_cast<int>(r)) + tr("This usually means that your client version is out of date, and the server sent a reply your client doesn't understand."));
```
ctrlaltca: +1 for moving the `\n` out of the translatable string, but using `<br>` instead can be problematic (see previous comment)

cockatrice/src/window_main.cpp
```diff
@@ -396,44 +401,44 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
...
 
-        out += "<li>" + tr("is %1 - %2 characters long").arg(rules.at(0)).arg(rules.at(1)) + "</li>";
+        out += "<li>" + tr("number of characters:") + QString("%1 - %2").arg(rules.at(0)).arg(rules.at(1)) + "</li>";
         out += "<li>" + tr("can %1 contain lowercase characters").arg((rules.at(2).toInt() > 0) ? "" : tr("NOT")) + "</li>";
         out += "<li>" + tr("can %1 contain uppercase characters").arg((rules.at(3).toInt() > 0) ? "" : tr("NOT")) + "</li>";
         out += "<li>" + tr("can %1 contain numeric characters").arg((rules.at(4).toInt() > 0) ? "" : tr("NOT")) + "</li>";
 
         if (rules.at(6).size() > 0)
-            out += "<li>" + tr("can contain the following punctuation: %1").arg(rules.at(6).toHtmlEscaped()) + "</li>";
+            out += "<li>" + tr("can contain the following punctuation:") + QString(" %1").arg(rules.at(6).toHtmlEscaped()) + "</li>";
```
ctrlaltca: Redundant `QString().arg()`

cockatrice/src/window_main.cpp
```diff
@@ -396,44 +401,44 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
...
 
         out += "<li>" + tr("first character can %1 be a punctuation mark").arg((rules.at(5).toInt() > 0) ? "" : tr("NOT")) + "</li>";
 
         if (rules.size() == 9)
         {
             if (rules.at(7).size() > 0)
-                out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
+                out += "<li>" + tr("can not contain any of the following words:") + QString(" %1").arg(rules.at(7).toHtmlEscaped()) + "</li>";
```
ctrlaltca: Redundant `QString().arg()`

cockatrice/src/window_main.cpp
```diff
@@ -396,44 +401,44 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
...
 
             if (rules.at(8).size() > 0)
-                out += "<li>" + tr("can not match any of the following expressions: %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
+                out += "<li>" + tr("can not match any of the following expressions:") + QString(" %1").arg(rules.at(8).toHtmlEscaped()) + "</li>";
```
ctrlaltca: Redundant `QString().arg()`

cockatrice/src/window_main.cpp
```diff
@@ -396,44 +401,44 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
...
         }
 
         out += "</ul>";
     }
     else
     {
-        out += tr("You may only use A-Z, a-z, 0-9, _, ., and - in your username.");
+        out += tr("You may only use this characters in your username:") + QString("\"A-Z\", \"a-z\", \"0-9\", \"_\", \".\", \"-\"");
```
ctrlaltca: these?

cockatrice/src/window_main.cpp
```diff
@@ -494,16 +499,20 @@ void MainWindow::activateError()
 
 void MainWindow::socketError(const QString &errorStr)
 {
-    QMessageBox::critical(this, tr("Error"), tr("Socket error: %1").arg(errorStr));
+    QMessageBox::critical(this, tr("Error"), tr("Socket error:") + QString(" %1").arg(errorStr));
```
ctrlaltca: Redundant `QString().arg()`

cockatrice/src/window_main.cpp
```diff
@@ -480,7 +485,7 @@ void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quin
             QMessageBox::critical(this, tr("Error"), tr("Registration failed for a technical problem on the server."));
             break;
         default:
-            QMessageBox::critical(this, tr("Error"), tr("Unknown registration error: %1").arg(static_cast<int>(r)) + tr("\nThis usually means that your client version is out of date, and the server sent a reply your client doesn't understand."));
+            QMessageBox::critical(this, tr("Error"), tr("Unknown registration error:") + QString(" %1<br>").arg(static_cast<int>(r)) + tr("This usually means that your client version is out of date, and the server sent a reply your client doesn't understand."));
```
ctrlaltca: `\n` => `<br>`

cockatrice/src/window_main.cpp
```diff
@@ -95,7 +95,7 @@ void MainWindow::processConnectionClosedEvent(const Event_ConnectionClosed &even
         case Event_ConnectionClosed::BANNED: {
             reasonStr = tr("Banned by moderator");
             if (event.has_end_time())
-                reasonStr.append("\n" + tr("Expected end time: %1").arg(QDateTime::fromTime_t(event.end_time()).toString()));
+                reasonStr.append("\n" + tr("Expected end time:" + QString(" ") + QDateTime::fromTime_t(event.end_time()).toString());
```
ZeldaZach: Tr tag needs a )

tooomm:
I addressed all feedback.
Travis keeps failing though... so I messed something up most likely.

Can somebody check and help please?

cockatrice/src/window_main.cpp
```diff
@@ -334,10 +337,10 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
     switch (r) {
         case Response::RespClientUpdateRequired: {
             QString formattedMissingFeatures;
-            formattedMissingFeatures = "Missing Features: ";
+            formattedMissingFeatures = "tr("Missing Features:") + QString:(" ")";
```
ctrlaltca: There's an excess quotation mark before tr() and one at the end of the string.

tooomm: Thanks. That should be ok now. 👍 
But well, Travis/Appveyor throws a lot of other stuff too...

tooomm:
Builds now, a lot of unrelated warnings left in both CI's though.

`[-Wmissing-declarations]` for example in travis

ctrlaltca:
> [-Wmissing-declarations] for example in travis

Don't worry, those messages are not related to your PR.

cockatrice/src/window_main.cpp
```diff
@@ -107,12 +107,15 @@ void MainWindow::processConnectionClosedEvent(const Event_ConnectionClosed &even
         case Event_ConnectionClosed::LOGGEDINELSEWERE: reasonStr = tr("You have been logged out due to logging in at another location."); break;
         default: reasonStr = QString::fromStdString(event.reason_str());
     }
-    QMessageBox::critical(this, tr("Connection closed"), tr("The server has terminated your connection.\nReason: %1").arg(reasonStr));
+    QMessageBox::critical(this, tr("Connection closed"), tr("The server has terminated your connection.") + QString("<br>")
```
Daenyth: Won't the `<br>` in here allow a server operator to inject html+javascript on you as ctrl mentioned? We shouldn't use `<br>` with any dynamic strings

tooomm: I don't know which ones are problematic and which not...
But yes, the `reason` comes from the server so there is a vector. Good point.

So basically if there is one single html piece in all `QMessageBox` or 
`serverShutdownMessageBox.setInformativeText` etc. - even if the part containing it (here a `QString("<br>")`) has no arguments itself inside where the code could pop up - it's an potential problem?
So if one single subpart/substring of e.g. `QMessageBox` has html, the complete `QMessageBox` will be treated as such? @ctrlaltca 

Should we just exchange every single `<br>` in general?
And also add it to contributing guidelines with explanation?

cockatrice/src/window_main.cpp
```diff
@@ -334,10 +337,10 @@ void MainWindow::loginError(Response::ResponseCode r, QString reasonStr, quint32
     switch (r) {
         case Response::RespClientUpdateRequired: {
             QString formattedMissingFeatures;
-            formattedMissingFeatures = "Missing Features: ";
+            formattedMissingFeatures = tr("Missing Features:") + QString(" ");
             for (int i = 0; i < missingFeatures.size(); ++i)
                 formattedMissingFeatures.append(QString("\n     %1").arg(QChar(0x2022)) + " " + missingFeatures.value(i)   );
-            formattedMissingFeatures.append("\nTo update your client, go to Help -> Check for Updates.");
+            formattedMissingFeatures.append(QString("\n") + tr("To update your client, go to \"Help -> Check for Client Updates\"."));
```
tooomm: Because it should mirror the exact phrasing of our menu item firstly.

We don't use this style of formatting in a uniform way at the moment... but as any other program (Word, Photoshop...) it's the way you label things in menus, names for windows, headlines etc.
It's called "title case" or "title capitalization", every newspaper uses it as well for captions for example.

tooomm:
@Daenyth @ctrlaltca can you both please revisit this, please. And approve+merge if it's ok.

Psithief:
I would like to see https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md updated with what to avoid when creating new translatable strings.

tooomm:
I'll see what I can do. Not here though, low on time atm.
But yes, totally reasonable! Will create an issue at least once this is finally merged.

tooomm:
linked wrong comment... dismiss because of https://github.com/Cockatrice/Cockatrice/pull/2718#discussion_r116386367 + revisit please @Daenyth

ctrlaltca:
Back at looking this PR, i see a few things that doesn't really looks like an improvement to me

### Splitting a translatable string in multiple strings
```diff
- tr("Users in this room: %1");
+ tr("Users connected to server:") + QString(" %1");
```
This change is making the assumption that the "value" will always be at the end of the string. This is not true for every language out there. An example from a language i know (italian) where the value position is different from english:
```
It is now fabio's turn. (It is now %1's turn.)
Adesso è il turno di fabio. (Adesso è il turno di %1.)
```

Some guidance at the end of this page: http://doc.qt.io/qt-5/i18n-source-translation.html

tooomm:
@ctrlaltca you mentioned this doubts earlier in this pr already: https://github.com/Cockatrice/Cockatrice/pull/2718#issuecomment-301325188
I think we are ok in doing so.

>This change is making the assumption that the "value" will always be at the end of the string. This is not true for every language out there.

Your example `It is now fabio's turn. (It is now %1's turn.)` is not of the type `What: How much`, it looks like a normal sentence for which translators need to understand variables and place them correctly in each language regarding its grammar.
But if you restructure and split strings up, the "value" will always be behind its "explanation", that's part of the nature of colons and should be the same between languages.
That's one of the advantages of this structure over real sentences. It only works for simpler constructions though.

<br>

Appveyor/**Branch** failling because outdated cmake in this branch. Building PR against master just fine.

cockatrice/src/dlg_update.cpp
```diff
@@ -142,8 +142,8 @@ void DlgUpdate::finishedUpdateCheck(bool needToUpdate, bool isCompatible, Releas
             tr("A new version of Cockatrice is available!") + "<br><br>"
             + "<b>" + tr("New version") + QString(":</b> %1<br>").arg(release->getName())
             + "<b>" + tr("Released") + QString(":</b> %1 <a href=\"%2\">(").arg(publishDate, release->getDescriptionUrl()) + tr("Changelog") + ")</a><br><br>"
-            + tr("Unfortunately there are no download packages available for your operating system. \nYou may have to build from source yourself.") + "<br><br>"
-            + tr("Please check the download page manually and visit the wiki for instructions on compiling."));
+            + tr("Unfortunately there could no download package for your operating system be found. Please check the download page manually.") + "<br><br>"
```
Nightfirecat: ???

tooomm: ???

skwerlman: `Unfortunately there could no download package for your operating system be found.` appears to be a typo or copy/paste error. Maybe change it to `No download package for your operating system could be found.`?

tooomm: https://github.com/Cockatrice/Cockatrice/pull/2718/commits/212af075d24e44af432a9e247a490dc2ec197162 should addresses this. What do you think @skwerlman ?

Are this the same complains you had @Nightfirecat ?
I wasn't sure what you try to point out.
Make sure to review/comment on the most recent code next time - and give me some hint, please. :P

skwerlman: 212af07 does indeed fix the issue i saw :+1:

Nightfirecat: Yes, that does fix it. 👍

tooomm: Ok, so thanks for reporting back. And poking in the first place! ;)

Issue #2702, opened 2017-05-08
Suggestion: add clone and create tokens to context menus for cards in graveyard and exile zones


aeschenkarnos:
OS and version number irrelevant.

__________

Currently a card has to be *on the battlefield* to bring up the "Clone" and "create token" right-click context menu items. This is not currently possible from the graveyard, library, exile, stack or hand zones. I suggest that there are occasional uses for this functionality and it should not be difficult to add it in.

- [x] Add "clone" to other zones
- [x] Add "create token" to other zones

tooomm:
Token creation also works on the stack... so in case for embalm for example you should pull the card out of your gy anyway to show what you are doing and hat your opponent can follow you. Technically you are casting it from the gy, so it should go over the stay anyway. From there you create the token and exile the original spell.

What else do you want/need to create tokens for from the gy? Or what do you want to clone from the exile?

tritoch:
You don't want to drag your `Bridge from Below` out of the graveyard every time it triggers.

tooomm:
Ok, so there is indeed a usecase. Sounds very helpful then! 👍
If we have it on gy, it should be the same for exile. Stack has tokens already - so we are just missing clone there

Daenyth:
Yes, this is totally reasonable. And should be an easy change too.

Nightfirecat:
While we're at it, would it be too complex to just add this context menu to cards in all zones? Another use case would be using Create Related for instants and sorceries which produce tokens, such as `Raise the Alarm`, et. al.

ZeldaZach:
@Nightfirecat I'll check this out

tooomm:
@Nightfirecat you can create tokens from instants and sorceries (all type of cards actually) from the stack already? :)
Right click --> `Token: 1/1 Soldier` for Raise the alarm

Nightfirecat:
> @Nightfirecat you can create tokens from instants and sorceries (all type of cards actually) from the stack already? :)
> Right click --> `Token: 1/1 Soldier` for Raise the alarm

Ah, right you are! If we could get this to be more generic, applying to cards in any zone, that might be helpful.

tooomm:
Other missing right click options related to zones: https://github.com/Cockatrice/Cockatrice/issues/2068

tooomm:
Actually, the missing token creation option should not be named `Create`, it should show the 
`Token: 2x 1/1 Soldier` option directly at the bottom of the right click options like it's done on the stack:
[...]

PR #2719, opened 2017-05-14
context menu fix

backendr:
## Related Ticket(s)
- Fixes #2714
- Support for #2702

#2714 introduced the issue of not being able to open a context menu on your opponents cards.

- Able to clone from zones apart from table #2702
- Removed unused import
- Context menu has been re-added for opponents cards
- Context menu for opponents cards has only 2 options:
    - draw arrow
    - clone

## Screenshots

screen shot 2017-05-14 at 11 11 28

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2079,7 +2079,7 @@ void Player::cardMenuAction()
                 }
                 case cmClone: {
                     Command_CreateToken *cmd = new Command_CreateToken;
-                    cmd->set_zone(card->getZone()->getName().toStdString());
+                    cmd->set_zone("table");
                     cmd->set_card_name(card->getName().toStdString());
                     cmd->set_color(card->getColor().toStdString());
                     cmd->set_pt(card->getPT().toStdString());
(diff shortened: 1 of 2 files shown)


backendr:
You are now able to clone cards from stack, meaning players that reveal hands to the stack can have them copied; which should prevent arguments.

tooomm:
Trice crashes if spectating a game if I run this commit.
With last full release everything works. Might be not related to this though...

backendr:
Fixed the crash. Try now.

tooomm:
Crash fixed 👍 

You don't see big right click menus on opponent cards any more. Since you couldn't use most of them anyway that's really cool. Just `Clone` and `Draw Arrow` left - do we miss something else here?

Old menu:
right_click_enemy

<br>

>Able to clone from zones apart from table #2702 

Note: Now able to clone enemy cards, so cards in their zones.
Still missing for your own zones, which 2702 particularly asks about.

cockatrice/src/player.cpp
```diff
@@ -2079,7 +2079,7 @@ void Player::cardMenuAction()
                 }
                 case cmClone: {
                     Command_CreateToken *cmd = new Command_CreateToken;
-                    cmd->set_zone(card->getZone()->getName().toStdString());
+                    cmd->set_zone("table");
```
ZeldaZach: can we make "table", "hand", etc into constants?

cockatrice/src/player.cpp
```diff
@@ -2519,6 +2519,13 @@ void Player::updateCardMenu(const CardItem *card)
             }
         } else
             cardMenu->addMenu(moveMenu);
+    } else {
+        if (card->getZone()
+            && card->getZone()->getName() != "hand") {
```
ZeldaZach: "hand" => some constant?

PR #2735, opened 2017-05-26
Webclient overhaul

ctrlaltca:
Migrate from protobuf.js 5.x to 6.x (remove long.js and bytebuffer.js)
Upgrade jQuery from 1.x to 3.x
Upgrade jQueryUI to 1.12.x
Use minimized version of js libraries
Disable debug messages
Fix default value for Event_RoomSay’s RoomMessageType field


tooomm:
When do we plan on splitting the webclient to it's own repo which has been around for quite some time now?

ZeldaZach:
@tooomm When do we plan on splitting Servatrice? I think we'll just one repo them for now in sub folders.

ctrlaltca:
At least the `common/pb` folder containing protobuf definitions needs to be shared between cockatrice, servatrice an the webclient. If we split the webclient into its own repo, it will need to include the files in that folder.
We could just copy the files to the new repo (but they could desync) or create a "protobuf" sub repo and include it in both cockatrice and webclient, but that's probably more hassle than it's worth by now.
Also, this discussion probably belongs to a new ticket, not this PR.

skwerlman:
I like the idea of having the protobuf files in their own repo. Updating the current repo to use the new protobuf repo would be done by adding a submodule, meaning the only real change would be adding a `--recursive` flag to each `git clone` in the build docs and CI scripts.

PR #2745, opened 2017-06-01
Make deck list appear if deck loaded

### LINKED ISSUE #2713: On deck load make sure "Deck" widget is visable
related #2478 

In case the widget is closed - show it again.
No reason you don't want to see your deck if you load it! ;)

ZeldaZach:
## Related Ticket(s)
- Fixes #2713 

If deck window was closed, then when you loaded a deck nothing really happened visually...

- On load of deck, deck list will become visible

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -937,6 +937,10 @@ void TabDeckEditor::setDeck(DeckLoader *_deck)
     PictureLoader::cacheCardPixmaps(db->getCards(deckModel->getDeckList()->getCardList()));
     deckView->expandAll();
     setModified(false);
+
+    // If they load a deck, make the deck list appear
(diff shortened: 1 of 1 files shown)


ctrlaltca:
When the deck widget gets hidden, its space gets redistributed between the other visible widgets.
if you make the deck widget visible again, it should reclaim its previous space.
If, while a widget is hidden, you change the other widget's geometry, it won't be able to recover exactly its previous space.
You can try to add some constraints to widget sizes (min/max width, height, ratio, ..) in order to make their default behaviour more sane, but this will probably cause accessibility issues for other users (eg. netbooks with small height).

> But this might be a worth an new issue for a better handling of the docked objects in general?

Keeping memory of this problem is a good idea, since it is already present when hiding/showing widgets manually.
Reworking the docked objects handling is a good idea, since each tab currently handles them separately and this means a lot of duplicated code..

PR #2746, opened 2017-06-01
make "not set to untap normally" normally more distinct

### LINKED ISSUE #2728: Suggestion: mark cards set not to untap
Currently there is no way of knowing which cards are set not to untap during their controller's untap step other than untapping everything and see what doesn't go along. Some way to mark these cards would be helpful. For example, border them in a different color.

ZeldaZach:
## Related Ticket(s)
- Fixes #2728 

Looked like a popular issue and it's something I feel should have been addressed a long time ago. This will add a magenta border to all cards that are set not to untap normally.

Also some cleanup with brackets to make it easier to understand.

screenshot 2017-06-01 00 27 09

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -90,7 +90,8 @@ void CardItem::paint(QPainter *painter, const QStyleOptionGraphicsItem *option, 
     
     int i = 0;
     QMapIterator<int, int> counterIterator(counters);
-    while (counterIterator.hasNext()) {
+    while (counterIterator.hasNext())
+    {
         counterIterator.next();
         QColor color;
         color.setHsv(counterIterator.key() * 60, 150, 255);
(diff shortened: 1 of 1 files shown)


Nightfirecat:
Code looks fine as far as I can tell. I'd like to consider some alternate ideas of how non-untapping cards should look before merging. What does this same effect look like in other clients? (MTGO, XMage, Untap) What other ways could this be displayed?

One consideration I had was to give the card a wide transparent grey shutter overlay. ([something like this, but with much larger blocks](http://thumb9.shutterstock.com/thumb_large/673339/246966916/stock-vector-vector-seamless-pattern-modern-stylish-texture-repeating-geometric-tiles-with-diagonal-lines-in-246966916.jpg)) Perhaps some kind of icon overlay to make it more clear what the effect actually is? Not necessarily the tap symbol, since that would infringe on copyright, but something in the same vein?

ZeldaZach:
I'm going to add this for now, but if we come up with a better/more interesting way to designate these cards are set to not untap, I'll be glad to make changes again in the future.

PR #2751, opened 2017-06-04
Model col widths off deck editor & disable button if no tokens in deck

ZeldaZach:
## Related Ticket(s)
- Fixes #2646 

- The widths of the token editor were random at best.
- Tokens in deck would show everything if no tokens were added to deck.

- Will now model exactly after the deck editor's layout (So if you set col1.width=100, tokensCol1.width will be 100 too). It's an exact copy, but you do not have the ability to edit the sizes in the token editor. (_If you make a change to deck editor width, token editor width will change, but not vice versa._)
- Button for tokens in deck will disable if no tokens were added to deck.

## Screenshots
screenshot 2017-06-03 22 56 05


tooomm:
Note: The column width in the `Edit tokens...` window is also affected by this!

PR #2760, opened 2017-06-09
More descriptive reason for "to many requests" when registering

woogerboy21:
Currently the only way for a user to get a response of "to many registration attempts"  is by the servers configuration to restrict the number of registration attempts but the error message presented to the user is very vague as to why they have been denied.  This PR updates the reasoning.  I'm open for suggestions on any clarity recommendations.

## Related Ticket(s)
I don't believe there is any.

I would say well over 95% of messages I receive from users comes from the vagueness of this message.  Just looking to clarify it a bit more.

The user dialogue wording that is presented if they attempt to register to many times.

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -455,7 +455,7 @@ void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quin
             QMessageBox::critical(this, tr("Registration denied"), tr("The email address provider used during registration has been blacklisted for use on this server."));
             break;
         case Response::RespTooManyRequests:
-            QMessageBox::critical(this, tr("Registration denied"), tr("It appears you are attempting to register a new account on this server yet you already have an account registered and the server restricts the number of accounts a user can register.  Please contact the server operator for further assistance or to obtain your credential information."));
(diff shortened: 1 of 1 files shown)


cockatrice/src/window_main.cpp
```diff
@@ -455,7 +455,7 @@ void MainWindow::registerError(Response::ResponseCode r, QString reasonStr, quin
             QMessageBox::critical(this, tr("Registration denied"), tr("The email address provider used during registration has been blacklisted for use on this server."));
             break;
         case Response::RespTooManyRequests:
-            QMessageBox::critical(this, tr("Registration denied"), tr("Too many registration attempts, please try again later or contact the server operator for further details."));
+            QMessageBox::critical(this, tr("Registration denied"), tr("It appears you are attempting to register a new account on this server yet you already have an account registered and the server restricts the number of accounts a user can register.  Please contact the server operator for further assistance or to obtain your credential information."));
```
Daenyth: > yet you already have an account registered 

Does that mean account with the same email address? If so, say that, maybe:

> ... but you already have an account registered to that email address. If you are trying to activate that account, <do x>. If you need to reset your password <do y>. Otherwise, contact the server operator for further assistance.

woogerboy21: How concerned with security are you?  I was trying to be as descriptive as possible without giving away to much information.  If we put something specific like "already registered with that email" then there is the potential for someone to hammer away at the registration process throwing email accounts at it and seeing if they get a response saying that is valid against an account.

Daenyth: They can already do that by checking whether or not they get an activation email or this error

PR #2787, opened 2017-06-29
Slightly more accurate XSD.

Psithief:
The previously pushed XSD does not enforce at least one appearance any element that was required. It's not practical to implement this without a sequence with XML1.0 for this many elements.

### On an unimportant note
Also, cockatrice is writing `<related>` tags after the `<color>` tags in `cards.xml`. That seems like an odd place for the output.

### DIFF
--- a/doc/cards.xsd
+++ b/doc/cards.xsd
@@ -1,4 +1,13 @@
 <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:complexType name="relatedType">
+    <xs:simpleContent>
+        <xs:extension base="xs:string">
+            <xs:attribute type="xs:string" name="count" use="optional"/>
+            <xs:attribute type="xs:string" name="exclude" use="optional"/>
+            <xs:attribute type="xs:string" name="attach" use="optional"/>
+        </xs:extension>
+    </xs:simpleContent>
+  </xs:complexType>
   <xs:element name="cockatrice_carddatabase">
     <xs:complexType>
       <xs:all>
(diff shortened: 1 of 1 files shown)


doc/cards.xsd
```diff
@@ -23,54 +32,37 @@
...
+                                                    <xs:attribute type="xs:int" name="muId" use="optional" />
+                                                    <xs:attribute type="xs:anyURI" name="picURL" use="optional" />
+                                                    <xs:attribute type="xs:string" name="num"  use="optional" />
+                                                    <xs:attribute type="xs:string" name="rarity" use="optional" />
+                                                </xs:extension>
+                                            </xs:simpleContent>
+                                        </xs:complexType>
+                                    </xs:element>
+                                    <xs:element type="xs:string" name="color" minOccurs="0" maxOccurs="5" />
+                                    <xs:element type="relatedType" name="related" minOccurs="0" maxOccurs="unbounded" />
```
Psithief: This is a cockatrice output quirk.

doc/cards.xsd
```diff
@@ -23,54 +32,37 @@
...
-                                            <xs:extension base="xs:string">
-                                                <xs:attribute type="xs:string" name="count" use="optional"/>
-                                                <xs:attribute type="xs:string" name="exclude" use="optional"/>
-                                                <xs:attribute type="xs:string" name="attach" use="optional"/>
-                                            </xs:extension>
-                                        </xs:simpleContent>
-                                    </xs:complexType>
-                                </xs:element>
-                            </xs:choice>
+                            <xs:sequence>
```
Psithief: XML 1.1 allows an `<xs:all>` that isn't restricted to 0 or 1 instances for every element.

Psithief:
> @Psithief you mentioned to switch to xml 1.1 the other day? what would be the advantage?
please open an issue if it's worth the change!

@tooomm I'm using Eclipse and they never implemented XML 1.1 support, ha.

PR #2794, opened 2017-07-03
Portable mode for windows

### LINKED ISSUE #1616: include portable option in NSIS win installer
Right now we have a normal installer file and a portable option that is more or less _unused_.

I would like to see them included as an option into the normal win installer (like other programs do), that would finally utilize the great option we have.
Even more with automated building and central hosting of the NSIS created file in the future.

https://gist.github.com/kevinhikaruevans/a46be6ad593e9fa3f7d2
http://stackoverflow.com/questions/13777988/use-nsis-to-create-both-normal-install-and-portable-install
http://stackoverflow.com/questions/23389007/nsis-install-like-in-teamviewer-portable-mode-option

example from [foobar2000](http://www.foobar2000.org) (sadly no open source):
nsis portable integration
_The "admin-rights-needed-shield" on the `Next` button will disappear if you select `Portable installation`_

### LINKED ISSUE #2789: Everything wants to go on the wrong drive
<b>OS: Windows 10</b>

<b>Cockatrice version: 2.3.17 (2017-05-05)</b>

__________

I'm trying to install Cockatrice on Drive D since that's my hard drive but even setting the install path there, creating a custom folder for it, and then doing the same for the card database, it still creates a database folder on my solid state Drive C and it STILL insists I need to update. How do I fix this?

ctrlaltca:
## Related Ticket(s)
- Fixes #1616
- Fixes #2789

Cockatrice already supported portable mode, but had to be compiled with a special flag. Since we didn't provide portable mode builds, this was impossible for users to use.

Portable mode is now activated when a spacial file named "portable.dat" is present in the same folder of cockatrice.exe.
The NSIS installer offers a choice between normal and portable install; normal installs works as before, while for portable installs there are a few differences:
 * the portable.dat file is created by NSIS during the install;
 * the default install directory is the user's desktop\Cockatrice;
 * no registry keys are touched: no uninstall entry is created, and no check for a previous installation is executed;
 * the "components" dialog is not shown; the "application" is installed and the "start menu" component is not installed.

While running in portable mode, Cockatrice will log "Portable mode enabled" in the debug log and disable the "edit" button for paths selection in the settings dialog.

Additional changes:
 * ask for administrator permissions when installing in "normal" mode;
 * support command line flags (portable mode, silent install, destination dir)
## Screenshots

schermata 2017-07-03 alle 20 00 34

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,6 @@ The following flags can be passed to `cmake`:
 - `-DWITH_SERVER=1` Whether to build the server (default 0 = no).
 - `-DWITH_CLIENT=0` Whether to build the client (default 1 = yes).
 - `-DWITH_ORACLE=0` Whether to build oracle (default 1 = yes).
-- `-DPORTABLE=1` Build portable versions of client & oracle (default 0 = no).
 - `-DCMAKE_BUILD_TYPE=Debug` Compile in debug mode. Enables extra logging output, debug symbols, and much more verbose compiler warnings (default `Release`).
 - `-DUPDATE_TRANSLATIONS=1` Configure `make` to update the translation .ts files for new strings in the source code. Note: Running `make clean` will remove the .ts files (default 0 = no).
 - `-DTEST=1` Enable regression tests (default 0 = no). Note: needs googletest, will be downloaded on the fly if unavailable. To run tests: ```make test```.
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -111,9 +111,9 @@ nsDialogs::Create 1018
 Pop $0
 ${NSD_CreateLabel} 0 10u 100% 24u "Select install mode:"
 Pop $0
-${NSD_CreateRadioButton} 30u 50u -30u 8u "Normal install"
+${NSD_CreateRadioButton} 30u 50u -30u 8u "Normal installation"
 Pop $1
-${NSD_CreateRadioButton} 30u 70u -30u 8u "Portable (all files in a single folder)"
+${NSD_CreateRadioButton} 30u 70u -30u 8u "Portable mode (all files in a single folder)"
 Pop $2
 ${If} $PortableMode = 0
     SendMessage $1 ${BM_SETCHECK} ${BST_CHECKED} 0
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -5,7 +5,7 @@ OutFile "@CPACK_TOPLEVEL_DIRECTORY@/@CPACK_OUTPUT_FILE_NAME@"
 
 !define UNINSTKEY "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Cockatrice"
 !define DEFAULTNORMALDESTINATON "$ProgramFiles\Cockatrice"
-!define DEFAULTPORTABLEDESTINATON "$Desktop\Cockatrice"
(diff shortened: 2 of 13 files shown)


tooomm:
This is awesome! 👍 

My observations/questions/suggestions with this are:
- Admin mode still requiered on file start, not just on "normal mode" installation (`Next >`)
Maybe you are aware of this and this is subject to change regarding your part at the end - wasn't quite sure.
- How do I update a portable version? is the internal installer dealing with a portable update correctly?
- Default location should be `desktop\Cockatrice portable` instead with **portable** branding
- The path in settings have a weird double slash `//` for one directory
port_paths
- I think as well that just disabling the edit button feels weird, the user things it's a normal input field and try to click or mark to do changes. It should be visible, but grayed out on top!
- The `Paths` headline for that settings group could additionally extend to: `Paths (editing disabled in portable mode)`
- There is a weird first run interaction if you hit the `settings` button on the dialog which asks you to run Oracle because there is no database found and you can't edit the paths. Cockatrice tries to offer you to change the paths but that doesn't work for sure.
- Phrasing `install` as noun is more colloquial I guess. should be `Portable installation` or `Install in portable mode` or just `Portable mode`. Same for the other option: `install normally` or `Standard installation`/`Normal installation`
Maybe also add a small explanation line like in the foobar example in #1616 
- `data` folder should be something like `user data` for clearness that that's the place where user can/should edit stuff. Normally it's also placed within the "user" folder, so...

ctrlaltca:
> Admin mode still requiered on file start, not just on "normal mode" installation (Next >)
> Maybe you are aware of this and this is subject to change regarding your part at the end - wasn't quite sure.

You can't elevate to admin rights in the middle of an application, but only when you start it.
Anyway, if you are an administrator on your pc, it will ask for admin elevation, and will permit you to install both in normal and in portable mode.
If you are not an administrator on your pc, it won't ask for admin elevation, but it will only permit you to install in portable mode:
schermata 2017-07-05 alle 09 19 25

> How do I update a portable version? is the internal installer dealing with a portable update correctly?

Re-installing the application over the same directory. I've added a "cleaning" routine in fbab7f0 that triggers when we are installing in portable mode on a directory where a previous portable mode installation is found.

> Default location should be desktop\Cockatrice portable instead with portable branding

Looking at https://portableapps.com/ as an example, they use "NamePortable" without spaces. I'd stick to their habits. Addressed in 60ce787

> The path in settings have a weird double slash // for one directory

This also happens in normal installations, i'll have a look in another PR.

> I think as well that just disabling the edit button feels weird, the user things it's a normal input field and try to click or mark to do changes. It should be visible, but grayed out on top!

Agreed, fixed in da3ca60

> The Paths headline for that settings group could additionally extend to: Paths (editing disabled in portable mode) or Paths (you can't change the locations in portable mode)

Agreed, fixed in da3ca60

> There is a weird first run interaction if you hit the settings button on the dialog which asks you to run Oracle because there is no database found and you can't edit the paths. Cockatrice tries to offer you to change the paths but that doesn't work for sure.

Yep, this is annoying, but i didn't come up with a good solution.. other that disable that button, doesn't really seems a good solution.

> Phrasing install as noun is more colloquial I guess. should be Portable installation or Install in portable mode or just Portable mode. Same for the other option: install normally or Standard installation/Normal installation

Addressed in 7568c27

> data folder should be something like user data for clearness that that's the place where user can/should edit stuff. Normally it's also placed within the "user" folder, so...

Sticking with https://portableapps.com/, they use "Data", and i don't see a good reason to change it.

PR #2796, opened 2017-07-04
Automatic Spoilers Season

ZeldaZach:
## Related Ticket(s)
- Fixes (anything?)

In the past, you had to manually download spoiler files and import them and go through the hassle and such.

- On cockatrice launch, download a small text file (that we host) to see if we're in spoiler season. If we aren't, just exit. Otherwise...
- Start a timer that is linked to settings. Every X minutes (configurable), we will download the spoilers file from the github page we have and replace the old one. This will also trigger the reloading of the card database and, if any new sets are found, that popup will appear as well.

## Screenshots

screenshot 2017-07-03 21 18 04
screenshot 2017-07-03 21 19 03

I spent a good few days on this, hopefully it's up to par with what we're looking for!

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -116,6 +116,7 @@ SET(cockatrice_SOURCES
     src/logger.cpp
     src/releasechannel.cpp
     src/userconnection_information.cpp
+    src/SpoilerBackgroundUpdaterThread.cpp
     ${VERSION_STRING_CPP}
 )
 
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -309,8 +309,3 @@ Data = Resources\")
         install(FILES ${WIN32SSLRUNTIME_LIBRARIES} DESTINATION ./)
     endif()
 endif()
-#Compile a portable version, default off
-option(PORTABLE "portable build" OFF)
-IF(PORTABLE)
-add_definitions(-DPORTABLE_BUILD)
-endif()
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -116,7 +116,7 @@ SET(cockatrice_SOURCES
     src/logger.cpp
     src/releasechannel.cpp
     src/userconnection_information.cpp
-    src/spoilerbackgroundupdaterthread.cpp
+    src/spoilerbackgroundupdater.cpp
     ${VERSION_STRING_CPP}
 )
 
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -116,7 +116,7 @@ SET(cockatrice_SOURCES
     src/logger.cpp
     src/releasechannel.cpp
     src/userconnection_information.cpp
-    src/SpoilerBackgroundUpdaterThread.cpp
+    src/spoilerbackgroundupdaterthread.cpp
     ${VERSION_STRING_CPP}
 )
 
--- a/cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
+++ b/cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
@@ -1,0 +1,233 @@
+#include <QDateTime>
+#include <QDebug>
+#include <QUrl>
+#include <QNetworkReply>
+#include <QMessageBox>
+#include <QFile>
+#include <QApplication>
+#include <QDesktopWidget>
+#include <QtConcurrent>
+#include "SpoilerBackgroundUpdaterThread.h"
+#include "settingscache.h"
+#include "carddatabase.h"
+#include "main.h"
+
+SpoilerBackgroundUpdaterThread::SpoilerBackgroundUpdaterThread(QObject *apParent) : QObject(apParent)
+{
+    // Determine if we're in spoilers season or not. If we're not, just kill this thread
+    downloadSpoilerSeasonStatusFile();
+    if (! mbIsActiveSpoilerSeason)
+    {
+        qDebug() << "Spoiler Timer disabled. NOT Spoilers Season";
+        return;
+    }
+
+    // Determine if the thread should be active in downloading/updating spoilers
+    mbIsActiveThread = settingsCache->getDownloadSpoilersStatus();
+    connect(settingsCache, SIGNAL(downloadSpoilerStatusChanged()), this, SLOT(changeActiveStatus()));
+
+    mpTimerForSpoilers = new QTimer(this);
+
+    // When timer ends, cause download of Spoilers XML file
+    connect(mpTimerForSpoilers, SIGNAL(timeout()), this, SLOT(timeoutOccurredTimeToDownloadSpoilers()));
+
+    // If there is a change in the settings, update the timer so we know when to update correctly
+    connect(settingsCache, SIGNAL(downloadSpoilerTimeIndexChanged()), this, SLOT(handleNewTimeInterval()));
+
+    // Start the timer sequence and the domino effects
+    runTimer();
+}
+
+SpoilerBackgroundUpdaterThread::~SpoilerBackgroundUpdaterThread()
+{
+    delete mpTimerForSpoilers;
+}
+
+void SpoilerBackgroundUpdaterThread::changeActiveStatus()
+{
+    mbIsActiveThread = (settingsCache->getDownloadSpoilersStatus());
+    qDebug() << "Spoiler Timer running status has been changed to" << mbIsActiveThread;
+}
+
+void SpoilerBackgroundUpdaterThread::timeoutOccurredTimeToDownloadSpoilers()
+{
+    qDebug() << "Spoiler Timer has finished";
+
+    settingsCache->setDownloadSpoilerLastUpdateTime(QDateTime::currentMSecsSinceEpoch());
+
+    // If this thread is active, then initiate download
+    if (mbIsActiveThread)
+    {
+        qDebug() << "Timer thread is active, so attempting to download spoilers";
+        downloadSpoilersFile(QString(GITHUB_SPOILERS_DOWNLOAD_URL));
+    }
+
+    // Now that we've downloaded the spoiler, reset the timer.
+    // Take it from the top, Jenkins!
+    runTimer(true);
+}
+
+void SpoilerBackgroundUpdaterThread::handleNewTimeInterval()
+{
+    // Cause the timer to stop then restart w/ new interval
+    qDebug() << "Spoiler Timer has new time interval established";
+    runTimer(true);
+}
+
+void SpoilerBackgroundUpdaterThread::runTimer(bool lbStopAndRestart)
+{
+    // Cause a restart of the timer as if we just restart via start() a timeout() will be
+    // thrown which will trigger a download... which we don't want off time!
+    if (lbStopAndRestart)
+    {
+        mpTimerForSpoilers->stop();
+    }
+
+    // Determine how long to wait between updates (Gets the # of hours from settingsCache / convert Mins -> Millis)
+    float lnTimeToWaitBetween = settingsCache->getDownloadSpoilerTimeMinutes() * 60000;
+
+    // How much time is needed until the next update (NOW() - Last Update Time == Time Remaining)
+    long lnLastUpdateTime = settingsCache->getDownloadSpoilerLastUpdateTime();
+    long lnTimeUntilNextUpdate = QDateTime::currentMSecsSinceEpoch() - lnLastUpdateTime;
+    if (lnTimeUntilNextUpdate <= 500 || lnLastUpdateTime == -1)
+    {
+        // The update just occurred less then half a sec ago, so we will restart the timer from the beginning (time zero)
+        mpTimerForSpoilers->start(static_cast<int>(lnTimeToWaitBetween));
+    }
+    else if (lnTimeUntilNextUpdate < lnTimeToWaitBetween)
+    {
+        // If we have checked for an update in the past and it's not yet time to check for another update
+        // Start the clock to count down until lnTimeUntilNextUpdate
+        mpTimerForSpoilers->start(static_cast<int>(lnTimeToWaitBetween - lnTimeUntilNextUpdate));
+    }
+    else
+    {
+        // It's been too long since we last checked (above threshold), so we will start now (time inf)
+        mpTimerForSpoilers->start(0);
+    }
+
+    qDebug() << "Spoiler Timer has started and will finish in" << mpTimerForSpoilers->interval()/60000. << "minutes";
+}
+
+void SpoilerBackgroundUpdaterThread::downloadSpoilersFile(QUrl acUrl)
+{
+    qDebug() << "Starting to download spoiler on NAM";
+    if (mpNetworkAccessManager == nullptr)
+    {
+        mpNetworkAccessManager = new QNetworkAccessManager(this);
+    }
+
+    QNetworkReply *lpNetworkReply = mpNetworkAccessManager->get(QNetworkRequest(acUrl));
+
+    connect(lpNetworkReply, SIGNAL(finished()), this, SLOT(actDownloadFinishedTokensFile()));
+}
+
+void SpoilerBackgroundUpdaterThread::downloadSpoilerSeasonStatusFile()
+{
+    qDebug() << "Checking spoiler season status";
+    if (mpNetworkAccessManager == nullptr)
+    {
+        mpNetworkAccessManager = new QNetworkAccessManager(this);
+    }
+
+    QNetworkReply *lpNetworkReply = mpNetworkAccessManager->get(QNetworkRequest(QString(GITHUB_SPOILERS_STATUS_URL)));
+
+    // Force the calling of actSeeIfSpoilerSeasonIsActive() in sequence instead of parallel
+    // as we need to get that result before we start this timer thread up.
+    QEventLoop lcSubLoop;
+    connect(lpNetworkReply, SIGNAL(finished()), this, SLOT(actSeeIfSpoilerSeasonIsActive()));
+    connect(lpNetworkReply, SIGNAL(finished()), &lcSubLoop, SLOT(quit()));
+    lcSubLoop.exec();
+}
+
+void SpoilerBackgroundUpdaterThread::actDownloadFinishedTokensFile()
+{
+    QNetworkReply *lpNetworkReply = static_cast<QNetworkReply *>(sender());
+    QNetworkReply::NetworkError leErrorCode = lpNetworkReply->error();
+    QWidget *lpTempWidget = new QWidget();
+    QString lsSpoilerSavePath = settingsCache->getSpoilerSavePath();
+
+    // Center the popup(s) that are to come in the near future
+    lpTempWidget->move(QApplication::desktop()->screen()->rect().center() - lpTempWidget->rect().center());
+    
+    if (leErrorCode != QNetworkReply::NoError)
+    {
+        QMessageBox::critical(lpTempWidget, tr("Error"), tr("Network error: %1.").arg(lpNetworkReply->errorString()));
+    }
+    else
+    {
+        int lnStatusCode = lpNetworkReply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+        if (lnStatusCode == 301 || lnStatusCode == 302)
+        {
+            // Was a redirect, will need to start over!
+            QUrl lcRedirectUrl = lpNetworkReply->attribute(QNetworkRequest::RedirectionTargetAttribute).toUrl();
+            downloadSpoilersFile(lcRedirectUrl);
+        }
+        else
+        {
+            // Read downloaded file into buffer
+            mcDownloadedFileData = lpNetworkReply->readAll();
+
+            if (saveSpoilersToFile(lsSpoilerSavePath))
+            {
+                // Reload the card database to import the spoilers
+                QtConcurrent::run(db, &CardDatabase::loadCardDatabases);
+                QMessageBox::information(lpTempWidget, tr("Success"), tr("Spoiler database updated and saved successfully to\n%1").arg(lsSpoilerSavePath));
+            }
+            else
+            {
+                QMessageBox::critical(lpTempWidget, tr("Error"), tr("Spoiler database updated but could not be saved to %1").arg(lsSpoilerSavePath));
+            }
+        }
+}
+
+lpNetworkReply->deleteLater();
+lpTempWidget->deleteLater();
+}
+
+void SpoilerBackgroundUpdaterThread::actSeeIfSpoilerSeasonIsActive()
+{
+    QNetworkReply *lpNetworkReply = static_cast<QNetworkReply *>(sender());
+    QNetworkReply::NetworkError leErrorCode = lpNetworkReply->error();
+    QWidget *lpTempWidget = new QWidget();
+    QString lsSpoilerSavePath = settingsCache->getSpoilerSavePath();
+
+    // Center the popup(s) that are to come in the near future
+    lpTempWidget->move(QApplication::desktop()->screen()->rect().center() - lpTempWidget->rect().center());
+
+    if (leErrorCode != QNetworkReply::NoError)
+    {
+        // If there's a network issue
+        mbIsActiveSpoilerSeason = false;
+        QMessageBox::critical(lpTempWidget, tr("Error"), tr("Network error: %1.").arg(lpNetworkReply->errorString()));
+    }
+    else
+    {
+        // If the file reads "enabled" then we're in spoiler season (true)
+        // If the file reads anything else then we're NOT in spoiler season (false)
+        QByteArray laFileData = lpNetworkReply->readAll();
+        mbIsActiveSpoilerSeason = QString(laFileData).contains("enabled", Qt::CaseInsensitive);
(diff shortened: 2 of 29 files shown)


cockatrice/CMakeLists.txt
```diff
@@ -116,6 +116,7 @@ SET(cockatrice_SOURCES
     src/logger.cpp
     src/releasechannel.cpp
     src/userconnection_information.cpp
+    src/SpoilerBackgroundUpdaterThread.cpp
```
Psithief: This is the only file that's not lowercase in the entire program.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include <QtConcurrent>
+#include "SpoilerBackgroundUpdaterThread.h"
+#include "settingscache.h"
+#include "carddatabase.h"
+#include "main.h"
+
+SpoilerBackgroundUpdaterThread::SpoilerBackgroundUpdaterThread(QObject *apParent) : QObject(apParent)
+{
+    // Determine if we're in spoilers season or not. If we're not, just kill this thread
+    downloadSpoilerSeasonStatusFile();
```
Psithief: This really needs to be _after_ the settings check. If someone doesn't want spoilers then there's an unnecessary request being made here.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include <QFile>
+#include <QApplication>
+#include <QDesktopWidget>
+#include <QtConcurrent>
+#include "SpoilerBackgroundUpdaterThread.h"
+#include "settingscache.h"
+#include "carddatabase.h"
+#include "main.h"
+
+SpoilerBackgroundUpdaterThread::SpoilerBackgroundUpdaterThread(QObject *apParent) : QObject(apParent)
```
Psithief: Some calls are from a QTimer and some aren't. This naming is confusing.

Psithief:
I have to recommend keeping everything in the main system and only spawning a thread for the actual download.
Just because you use QTimer doesn't mean you need a thread.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+{
+    // Cause a restart of the timer as if we just restart via start() a timeout() will be
+    // thrown which will trigger a download... which we don't want off time!
+    if (lbStopAndRestart)
+    {
+        mpTimerForSpoilers->stop();
+    }
+
+    // Determine how long to wait between updates (Gets the # of hours from settingsCache / convert Mins -> Millis)
+    float lnTimeToWaitBetween = settingsCache->getDownloadSpoilerTimeMinutes() * 60000;
```
ctrlaltca: Why float? maybe qint64

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    // How much time is needed until the next update (NOW() - Last Update Time == Time Remaining)
+    long lnLastUpdateTime = settingsCache->getDownloadSpoilerLastUpdateTime();
+    long lnTimeUntilNextUpdate = QDateTime::currentMSecsSinceEpoch() - lnLastUpdateTime;
```
ctrlaltca: As far as i understand, `lnTimeUntilNextUpdate` should ne named "lnTimeSinceLastUpdate"

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+    // If there is a change in the settings, update the timer so we know when to update correctly
+    connect(settingsCache, SIGNAL(downloadSpoilerTimeIndexChanged()), this, SLOT(handleNewTimeInterval()));
+
+    // Start the timer sequence and the domino effects
+    runTimer();
+}
+
+SpoilerBackgroundUpdaterThread::~SpoilerBackgroundUpdaterThread()
+{
+    delete mpTimerForSpoilers;
```
ctrlaltca: you should take care of deleting `mpNetworkAccessManager ` here, too, if it has been instanciated (!= nullptr).

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+{
+    qDebug() << "Starting to download spoiler on NAM";
+    if (mpNetworkAccessManager == nullptr)
+    {
+        mpNetworkAccessManager = new QNetworkAccessManager(this);
+    }
+
+    QNetworkReply *lpNetworkReply = mpNetworkAccessManager->get(QNetworkRequest(acUrl));
+
+    connect(lpNetworkReply, SIGNAL(finished()), this, SLOT(actDownloadFinishedTokensFile()));
```
ctrlaltca: Maybe rename `actDownloadFinishedTokensFile` to `actDownloadFinishedSpoilersFile`

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    QNetworkReply *lpNetworkReply = mpNetworkAccessManager->get(QNetworkRequest(QString(GITHUB_SPOILERS_STATUS_URL)));
+
+    // Force the calling of actSeeIfSpoilerSeasonIsActive() in sequence instead of parallel
+    // as we need to get that result before we start this timer thread up.
+    QEventLoop lcSubLoop;
+    connect(lpNetworkReply, SIGNAL(finished()), this, SLOT(actSeeIfSpoilerSeasonIsActive()));
+    connect(lpNetworkReply, SIGNAL(finished()), &lcSubLoop, SLOT(quit()));
+    lcSubLoop.exec();
```
ctrlaltca: This looks like an hack, and should better be refactored.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+lpNetworkReply->deleteLater();
+lpTempWidget->deleteLater();
+}
+
+void SpoilerBackgroundUpdaterThread::actSeeIfSpoilerSeasonIsActive()
+{
+    QNetworkReply *lpNetworkReply = static_cast<QNetworkReply *>(sender());
+    QNetworkReply::NetworkError leErrorCode = lpNetworkReply->error();
+    QWidget *lpTempWidget = new QWidget();
```
ctrlaltca: Is this widget created only to pass a parent to QMessageBox? It should work the same if you pass 0 or nullptr.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+    if (! lcFileToWriteTo.open(QIODevice::WriteOnly))
+    {
+        qDebug() << "File open (w) failed for" << asFileName;
+        return false;
+    }
+
+    if (lcFileToWriteTo.write(mcDownloadedFileData) == -1)
+    {
+        qDebug() << "File write (w) failed for" << asFileName;
+        return false;
```
ctrlaltca: you should try to close the file even if write failed before returning.

cockatrice/src/settingscache.h
```diff
@@ -133,12 +142,18 @@ class SettingsCache : public QObject {
...
+    QString getSpoilerSavePath() const { return msSpoilerSavePath; }
     bool getPicDownload() const { return picDownload; }
     bool getNotificationsEnabled() const { return notificationsEnabled; }
+    bool getDownloadSpoilersStatus() const { return mbDownloadSpoilers; }
     bool getSpectatorNotificationsEnabled() const { return spectatorNotificationsEnabled; }
     bool getNotifyAboutUpdates() const { return notifyAboutUpdates; }
     ReleaseChannel * getUpdateReleaseChannel() const { return releaseChannels.at(updateReleaseChannel); }
     QList<ReleaseChannel*> getUpdateReleaseChannels() const { return releaseChannels; }
+    int getDownloadSpoilerTimeIndex() const { return mnDownloadSpoilersTimeIndex; }
+    float getDownloadSpoilerTimeMinutes() const { return msDownloadSpoilersTimeMinutes; }
```
ctrlaltca: Why do you save in the settings file both the minutes and the entry index in the combobox? You should be able to derive one value from the other.
The problem is that if we ever ned to change the values on the list and keep them ordered, the mapping between indexes and minutes would change and this could cause odd effects when loading old settings in a new client version.

cockatrice/src/window_main.cpp
```diff
@@ -19,6 +19,7 @@
  ***************************************************************************/
 #include <QMenu>
 #include <QAction>
+#include <iostream>
```
ctrlaltca: Planning to use `cout`? Please remove if unused

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,233 @@
...
+        mbIsActiveSpoilerSeason = false;
+        QMessageBox::critical(lpTempWidget, tr("Error"), tr("Network error: %1.").arg(lpNetworkReply->errorString()));
+    }
+    else
+    {
+        // If the file reads "enabled" then we're in spoiler season (true)
+        // If the file reads anything else then we're NOT in spoiler season (false)
+        QByteArray laFileData = lpNetworkReply->readAll();
+        mbIsActiveSpoilerSeason = QString(laFileData).contains("enabled", Qt::CaseInsensitive);
+    }
```
ctrlaltca: To avoid the hack code at lines 135-140, you could start the timer here instead of in the constructor.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: At one point i was using floats, didn't change back to int. Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Thought I did this, must've stashed it by accident. Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: This 'hack' currently forces the file to be downloaded and parsed _before_ the constructor can continue execution. We need to know if we're in spoiler season or not before continuing, at least that was my idea...

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Moved, not sure if it'll do much though

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Addressed

cockatrice/src/window_main.cpp
ZeldaZach: Addressed

cockatrice/src/settingscache.h
ZeldaZach: Addressed

cockatrice/CMakeLists.txt
ZeldaZach: Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Addressed

ZeldaZach:
I have pushed a change that addresses all the comments brought up.
- The program will NOT check for a spoilers byte file on github if you don't have the feature enabled. 
- If you choose to enable it while the program is running, it will auto-download at that moment and then restart the timer from that point.
- Fixed up file name
- More stuff (see the changelog).

Please provide additional feedback at this time :)

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: I'm not sure what you want from here...

cockatrice/src/dlg_settings.cpp
```diff
@@ -464,25 +464,154 @@ void UserInterfaceSettingsPage::retranslateUi()
...
+    if (! lsPath.isEmpty())
+    {
+        mpSpoilerSavePathLineEdit->setText(lsPath + "/spoiler.xml");
+        settingsCache->setSpoilerSavePath(lsPath + "/spoiler.xml");
+    }
+}
+
+void DeckEditorSettingsPage::setDownloadSpoilerTime(QString asValue)
+{
+    // Sets the downloadspoilerstimeHours field (stored in HOURS (float))
```
ZeldaZach: haha... minutes, int

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,249 @@
+#include <QDateTime>
```
ctrlaltca: This file needs to be renamed in order to be all lowercase, or compilation will fail on case sensitive filesystems (eg. Linux)

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,249 @@
...
+void SpoilerBackgroundUpdaterThread::actDownloadFinishedSpoilersFile()
+{
+    QNetworkReply *lpNetworkReply = static_cast<QNetworkReply *>(sender());
+    QNetworkReply::NetworkError leErrorCode = lpNetworkReply->error();
+    QString lsSpoilerSavePath = settingsCache->getSpoilerDatabasePath();
+    
+    if (leErrorCode != QNetworkReply::NoError)
+    {
+        QMessageBox::critical(nullptr, tr("Error"), tr("Network error: %1.").arg(lpNetworkReply->errorString()));
+    }
```
ctrlaltca: Just a personal preference, take it as a "soft" suggestion:
In methods like this where you have a lot of if/then/else branches, try to keep the logic flow as simple as possible. For example, here adding a 
```
lpNetworkReply->deleteLater();
return;
```
you can remove the `else` branch and de-indent everything of one step.
The same can be done at line 185.
So instead of having three levels of nested `if`s, you can get an easier to follow flow like this:
```
if(!condition1)
  return;
if(!condition2)
  return;
get_the_job_done();
```

cockatrice/src/dlg_settings.cpp
```diff
@@ -464,25 +464,154 @@ void UserInterfaceSettingsPage::retranslateUi()
...
+    updateDownloadTimer(lbVisibility);
+}
+
+void DeckEditorSettingsPage::updateDownloadTimer(bool abIsEnabled)
+{
+    // Update the display so the user sees when the next download time is or nothing if disabled
+    if (abIsEnabled)
+    {
+        long lnTimeUntilNextUpdate = QDateTime::currentMSecsSinceEpoch() - settingsCache->getDownloadSpoilerLastUpdateTime();
+        float lnTimeToWaitBetween = settingsCache->getDownloadSpoilerTimeMinutes() * 60000;
```
ctrlaltca: Why float? Use long, long long (or, better, qint64) and avoid the static_cast<long> in the following lines

cockatrice/src/dlg_settings.cpp
```diff
@@ -464,25 +464,154 @@ void UserInterfaceSettingsPage::retranslateUi()
...
+    lpSpoilerGrid->setColumnStretch(0, 2);
+    lpSpoilerGrid->setColumnStretch(1, 4);
+    lpSpoilerGrid->setColumnStretch(2, 1);
+
+    lpGeneralGrid->addWidget(&mcGeneralMessageLabel, 0, 0);
+
+    // On a change to the check box, hide/unhide the other fields
+    // On a change to the combo box, send a signal to the timer threads
+    connect(&mcDownloadSpoilersCheckBox, SIGNAL(stateChanged(int)), settingsCache, SLOT(setDownloadSpoilerStatus(int)));
+    connect(&mcDownloadSpoilersCheckBox, SIGNAL(stateChanged(int)), this, SLOT(setSpoilersEnabled(int)));
```
ctrlaltca: You may want to bind the `toggled(bool)` signal instead to avoid the `static_cast<bool>` in the receiving code
(Also, the int returned by stateChanged has more values than just 0-unchecked 1-checked)

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: I swear I did... I guess github doesn't recognize case changes

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "main.h"
+
+SpoilerBackgroundUpdaterThread::SpoilerBackgroundUpdaterThread(QObject *apParent) : QObject(apParent)
+{
+    // User doesn't want to download spoilers... for now :)
+    mbIsActiveThread = settingsCache->getDownloadSpoilersStatus();
+    connect(settingsCache, SIGNAL(downloadSpoilerStatusChanged()), this, SLOT(changeActiveStatus()));
+
+    // When timer ends, cause download of Spoilers XML file
+    connect(mpTimerForSpoilers, SIGNAL(timeout()), this, SLOT(timeoutOccurredTimeToDownloadSpoilers()));
```
ZeldaZach: Addressed

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ZeldaZach: Dealt with

cockatrice/src/dlg_settings.cpp
ZeldaZach: Addressed

cockatrice/src/dlg_settings.cpp
ZeldaZach: Addressed

ZeldaZach:
I have made all the requested changes.

cockatrice/src/SpoilerBackgroundUpdaterThread.cpp
ctrlaltca: If you are working on a Mac, the filesystem is case insensitive, so it won't pick up changes to the case of file names.. you can probably workaround it by renaming the file to something else, committing, then rename again to the correct name and re-commit.

cockatrice/src/settingscache.h
```diff
@@ -4,6 +4,7 @@
 #include <QObject>
 #include <QSize>
 #include <QStringList>
+#include <QTimer>
```
ctrlaltca: Probably unneeded

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,241 @@
...
+        return;
+    }
+
+    // User wants spoilers, so lets see if we're in spoiler season or not.
+    downloadSpoilerSeasonStatusFileAndStartTimer();
+}
+
+SpoilerBackgroundUpdater::~SpoilerBackgroundUpdater()
+{
+    delete mpTimerForSpoilers;
```
ctrlaltca: If the user disabled spoiler checking, mpTimerForSpoilers  will never be created, and trying to delete it will probably cause a crash.
Anyway, i saw that when you instanciate mpTimerForSpoilers and mpNetworkAccessManager you always pass "this" as the object parent, so you can just skip deleting these 2 objects since Qt will take care of deleting them when their parent gets deleted.
Sorry, i know i first asked you to explicitly delete mpNetworkAccessManager, i overlooked the code.

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,241 @@
...
+    {
+        // If the file reads "enabled" then we're in spoiler season (true)
+        // If the file reads anything else then we're NOT in spoiler season (false)
+        QByteArray laFileData = lpNetworkReply->readAll();
+        mbIsActiveSpoilerSeason = QString(laFileData).contains("enabled", Qt::CaseInsensitive);
+
+        if (mbIsActiveSpoilerSeason)
+        {
+            // Start the timer for spoilers downloads
+            mpTimerForSpoilers = new QTimer(this);
```
ctrlaltca: runTimer() already takes care of creating the timer if it doesn't exists yet, no need to do it here
(Also, this was not checking if the timer already existed, so it could create a duplicate)

cockatrice/src/settingscache.h
ZeldaZach: Addressed

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: Addressed

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: Addressed

ZeldaZach:
When I try to click the checkbox on windows:
```
Problem signature:
  Problem Event Name:	APPCRASH
  Application Name:	cockatrice.exe
  Application Version:	0.0.0.0
  Application Timestamp:	5960b087
  Fault Module Name:	Qt5Network.dll
  Fault Module Version:	5.6.2.0
  Fault Module Timestamp:	57e77689
  Exception Code:	c0000005
  Exception Offset:	00000000000252e6
  OS Version:	6.1.7601.2.1.0.256.48
  Locale ID:	1033
  Additional Information 1:	8355
  Additional Information 2:	83551fe3c4bc47bb701562fbf1154230
  Additional Information 3:	0e93
  Additional Information 4:	0e9318370a9d9370baef69ec5f67ab73

Read our privacy statement online:
  http://go.microsoft.com/fwlink/?linkid=104288&clcid=0x0409

If the online privacy statement is not available, please read our privacy statement offline:
  C:\Windows\system32\en-US\erofflps.txt

```

ZeldaZach:
@ctrlaltca Some debugging stuff:
screenshot 2017-07-08 12 43 26
screenshot 2017-07-08 12 43 43

Brings me to https://github.com/Cockatrice/Cockatrice/pull/2796/files#diff-2c069c6d2a8e071f48b42363bb52741aR137 in the code base

ctrlaltca:
schermata 2017-07-09 alle 10 22 21

As you can see in the bottom left panel, both `mpNetworkAccessManager` and  `lpNetworkReply` have a weird value of `0xcccccccc`. At this point in the code you would expect `mpNetworkAccessManager` to have been instanciated a few lines before, while `lpNetworkReply` should be an invalid pointer, since we're just creating it.
Instead, they both contain an invalid value; this is due to the fact that `mpNetworkAccessManager` is not initialized to `nullptr` when the class is instanciated, so it will contain random garbage from the memory.

cockatrice/src/spoilerbackgroundupdater.h
```diff
@@ -0,0 +1,41 @@
...
+    void handleNewTimeInterval();
+
+private slots:
+    void actDownloadFinishedSpoilersFile();
+    void actSeeIfSpoilerSeasonIsActive();
+    void changeActiveStatus();
+
+private:
+    QTimer *mpTimerForSpoilers;
+    QNetworkAccessManager *mpNetworkAccessManager;
```
ctrlaltca: When you add a pointer member to a class, always ensure you initialize it in the class constructor, either instnciating the object or setting it to nullptr

Daenyth:
What happens with the spoiler file when season is over? Is there anything that prevents double entries, when you have a spoiler file + oracle run after the sets are out?

I'm concerned about user education. Oracle is already hard to understand for users. This would make it seem like they get the new cards automatically but then when season is over, they need to run oracle manually. Thoughts? I don't think that's a blocker for adding this - I think it's an awesome feature addition. I'm just wondering how we can best educate users.

I haven't looked at the code yet, but one thing I'd like to make sure is that all the steps around this have good debug log output so we can help people troubleshoot.

ZeldaZach:
@ctrlaltca You still working on a sub-PR for this?

PR #2797, opened 2017-07-04
Ignore incomplete unknown sets

### LINKED ISSUE #2712: Adding set code to tokens.xml creates set list entry
If you add a set code to tokens.xml, an entry is created in set list and you are prompted with `New set detected. Do you want to enable it?`

This behavior is unexpected. During spoiler season, the token information is published considerably before gatherer is updated. A user will run oracle, be prompted if they want to enable the new set, say yes, and expect to see the new cards. In addition, when the set is finally added to gatherer/mtgjson (which can take 2 weeks after token information is published) and the user re-runs Oracle, the user is *not* prompted again to enable to new set, because they already added tokens.

Psithief:
## Related Ticket(s)
- Fixes #2712

When checking unknown sets, consider sets with an empty long name, 
set type and release date as 'incomplete'/'partial' sets.

Do not mark partial sets as known, or ask the user to enable them.
Instead, silently enable partial sets.
When a partial set becomes complete, the user is prompted to enable the
set as before. If they choose not to enable, those partial sets are 
disabled, as we can no longer assume they are disabled by default.

### Detail
In practice, a missing `<sets>…<set>…</set>…</sets>` for example set `HOU` will be considered equivalent to a set with 3 zero-length details:
```xml
<set>
    <name>HOU</name>
    <longname></longname>
    <settype></settype>
    <releasedate></releasedate>
</set>
```

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -99,7 +99,7 @@ int SetList::getUnknownSetsNum()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(!set->getIsKnown())
+        if(!set->getIsKnown() && !set->getIsKnownIgnored())
             ++num;
     }
     return num;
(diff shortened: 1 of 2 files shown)


Psithief:
This commit isn't really trying to solve that problem. It's trying to solve the tokens problem.
Improvements to spoilers are a larger task that @ZeldaZach is already taking a look at, apparently.

PR #2799, opened 2017-07-05
enable caching for ci

tooomm:
- apt cache on travis was never working, see https://github.com/travis-ci/travis-ci/issues/5876
- no caching on appveyor used

- Travis: use ccache over apt (on mac also cache Homebrew)
to check our travis cache go to: https://travis-ci.org/Cockatrice/Cockatrice/caches
- Appveyor: **TODO**

>Hope to speed up thinks considerably!
>**Build time before:** Travis: linux (5-6min), macos (8-11min) / Appveyor: win (6-7min)
>**Build time after:** Travis: linux (1-2min), macos (5-8min) / Appveyor: win (not enabled yet)
>
>Overall improvement: Travis: ~30min --> ~16min (**-50%**) / Appveyor: **tbd**

<br>

---

Travis caching on osx: https://docs.travis-ci.com/user/caching/#ccache-on-OS-X **done**
Homebrew caching: https://stackoverflow.com/questions/39930171/cache-brew-builds-with-travis-ci **done**

Same should be possible on appveyor: https://www.appveyor.com/docs/build-cache/ **TODO**

cc: @ctrlaltca

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
     env: BUILDTYPE=Debug
     cache:
       ccache: true
+      branch: no-md5deep
       directories:
         - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
   - os: osx
@@ -23,6 +24,7 @@ matrix:
     env: BUILDTYPE=Release
     cache:
       ccache: true
+      branch: no-md5deep
       directories:
         - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
     env: BUILDTYPE=Debug
     cache:
       ccache: true
+      timeout: 360 # 6min timeout for cache uploading instead default 3
       directories:
         - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
   - os: osx
@@ -23,6 +24,7 @@ matrix:
     env: BUILDTYPE=Release
     cache:
       ccache: true
+      timeout: 360 # 6min timeout for cache uploading instead default 3
       directories:
         - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,19 +12,17 @@ matrix:
     dist: trusty
     env: BUILDTYPE=Release DIST=trusty
   - os: osx
-    osx_image: xcode6.4
+    osx_image: xcode7.3
     env: BUILDTYPE=Debug
     cache:
       ccache: true
-      branch: no-md5deep
       directories:
         - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
   - os: osx
-    osx_image: xcode6.4
+    osx_image: xcode7.3
     env: BUILDTYPE=Release
     cache:
       ccache: true
-      branch: no-md5deep
       directories:
         - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ matrix:
     env: BUILDTYPE=Release
 script: "./travis-compile.sh"
 install: "./travis-dependencies.sh"
-cache: apt
+cache: ccache
```
ctrlaltca: We are not using ccache either.

tooomm: triggered a second run here manually:
https://travis-ci.org/Cockatrice/Cockatrice/builds/250241804

ctrlaltca: So, just adding the line `cache: ccache` it seems to affect the travis environment:
Without:
```
$ export CXX=g++
$ export CC=gcc
....
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
```

With `cache: ccache`:
```
$ export CXX=g++
$ export CC=gcc
$ export PATH=/usr/lib/**ccache**:$PATH
cache.1
Setting up build cache
....
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/lib/**ccache**/gcc
-- Check for working C compiler: /usr/lib/**ccache**/gcc -- works
```

This is actually quite cool!
It doesn't seems to work for OSX builds, anyway.

tooomm: Please check my link above: https://docs.travis-ci.com/user/caching/#ccache-cache

We have to add this to make it work on mac. Wasn't sure where to put it or if there some mac specific lines are needed. Feel free to add it! (that's why i cc'ed you :P)

travis-dependencies.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 if [[ $TRAVIS_OS_NAME == "osx" ]] ; then
+  brew install ccache                       # enable caching on mac
+  PATH=/usr/local/opt/ccache/libexec:$PATH  # https://docs.travis-ci.com/user/caching/#ccache-on-OS-X
```
ctrlaltca: I suppose this is not correct, what we need is `export PATH="/usr/local/opt/ccache/libexec:$PATH"` for it to actually have an effect on the following commands

Also, can you move these two lines after the "brew update" step?

tooomm: Do we need brew update for sure? It has a *huge* impact on build time (2min!).
https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-OS-X

ctrlaltca: You can try to remove it, let's see if it works :)

.travis.yml
```diff
@@ -10,21 +12,28 @@ matrix:
...
+    osx_image: xcode7.3
     env: BUILDTYPE=Debug
+    cache:
+      ccache: true
+      timeout: 360 # 6min timeout for cache uploading instead default 3
+      directories:
+        - $HOME/Library/Caches/Homebrew   # cache Homebrew ressources
   - os: osx
-    osx_image: xcode6.4
+    osx_image: xcode7.3
```
ctrlaltca: This change will drop support for macOS 10.10, but it avoids a few problems with homebrew and will be inevitable anyway in a few months when the next macOS gets released.
@Daenyth is it ok for you?

ZeldaZach: I'm fine with this move

Daenyth: If you're running 10.10 I'm pretty sure you have security exploits present, so I'm fine with this

Issue #2068, opened 2016-07-09
ability to `attach to` from stack


tooomm:
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
641fd4a (2016-07-09)

---

<br>

While playing an enchantment you put it on the stack and point to your target creature for example.
It should be possible to immediately right click and choose `Attach to card...` (while the card is on the stack) instead of putting it to the battlefield first and right click then.

![attachstack](https://cloud.githubusercontent.com/assets/9874850/16707736/7aa71440-45d9-11e6-8753-d1c6cb1b6c69.png)

ZeldaZach:
I started preliminary working on this, you can check my branch (https://github.com/ZeldaZach/Cockatrice/tree/fix_2068). it seems to glitch out sometimes, though so I'm not 100% certain what's up.

ZeldaZach:
This should be doable with the current protocol. I'd like this added

PR #2813, opened 2017-07-30
Clone in exile graveyard

jjcard:
## Related Ticket(s)
- Fixes narrow scope of initial problem in #2702

 - Playing a game that required cloning a lot of cards from graveyard and exile required pulling them out individually back to the table to clone them.

- Add ability to clone cards from graveyard and exile
- small formatting enhancement per style guide

## Screenshots
exileclone
graveyardclone

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2555,14 +2555,10 @@ void Player::updateCardMenu(const CardItem *card)
         }
     }
 
-    if (revealedCard)
-    {
+    if (revealedCard) {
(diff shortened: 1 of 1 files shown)


Daenyth:
Nice addition! Thanks :)

tooomm:
Small leftover... You added `Clone` to "gy" and "exile", but you missed to add it to "stack" as well!

Can you add that as well, please? Then, the first bullet point of #2702 can be checked. 👍 
(#2068 requests to add another right click menu on items on the stack btw)

PR #2815, opened 2017-08-01
Double click token

### LINKED ISSUE #2754: Double click on a token in list should create it
Double click on a token in `Create token` dialog should create it.

Right now you are forced to hit the `Create` button or <kbd>Enter</kbd>.
Should be basically the same as with a game in the game list. You can join on double click after https://github.com/Cockatrice/Cockatrice/pull/454, and don't need to use the `Join Game` button anymore.

jjcard:
## Related Ticket(s)
- Fixes #2754

- Want the ability to create Tokens faster, same as joining game.

- Adds ability to create token when it is double-clicked in the list in "Create Token" dialog


Daenyth:
Oh boy I've wanted this for a while.

jjcard:
Oops, missed that. I have updated tabs to 4 spaces

tooomm:
Thanks, very nice! :)

PR #2816, opened 2017-08-03
Disabling in set window moves row to bottom of list

### LINKED ISSUE #2806: Card art preference failure
<b>OS: Windows 10</b>

<b>Cockatrice version: 6e723b2 (2017-03-14)</b>

__________

Hi all. I have organized my card art preference in ascending order of release date, and unchecked some sets. For most cards, older art is successfully preferred by Cockatrice, but for others, card art from unchecked sets is being used. In particular, I hate the Amonkhet invocations art but it's still being used despite the box being unchecked in "Edit Sets...". I have updated the art database and retried, but to no avail.

Viomi:
## Related Ticket(s)
- Fixes #2806 

Wasn't user-friendly to leave disabled sets in current location- Makes more sense to move them to bottom automatically.

- Disabling a set in the set selection window will now also move it to the bottom of the list.
- This applies if the set is disabled in any way, including setting multiple at once.

### DIFF
--- a/cockatrice/src/setsmodel.cpp
+++ b/cockatrice/src/setsmodel.cpp
@@ -135,10 +135,17 @@ void SetsModel::toggleRow(int row, bool enable)
 {
     CardSet *temp = sets.at(row);
 
-    if (enable)
+    int numRows = rowCount();
+
+    if (enable) {
         enabledSets.insert(temp);
-    else
+        swapRows(row, 0);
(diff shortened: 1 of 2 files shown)


Daenyth:
Like I mentioned, I'd strongly prefer this this behavior not be used, and instead just process the enable/disable during download. I don't see a compelling case for this

PR #2817, opened 2017-08-05
Debug Log Qstring warnings

jjcard:
 - Noticed "QString::arg: Argument missing:" warnings in Debug Log

- In logDoMoveCard method in messagelogwidget.cpp, made it only conditionally add the 4th argument that was making the warning happen.

## Screenshots
- Before
debugqstringbefore
- After
debugqstringafter

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -317,6 +317,7 @@ void MessageLogWidget::logDoMoveCard(LogMoveCard &lmc)
     }
 
     QString finalStr;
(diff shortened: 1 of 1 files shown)


ZeldaZach:
I don't think having the log spew QPainter info like that is useful, maybe you can change it up instead?

jjcard:
I think that might be an artifact of me compiling it locally, That was happening both before and after my change, but running the release version I have installed doesn't seem to spew it. 
 And googling the log seems to suggest that it was a Qt issue that has been fixed...

PR #2822, opened 2017-08-06
split up string and argument

tooomm:
## Related Ticket(s)
- small addition to https://github.com/Cockatrice/Cockatrice/pull/2818 (cc: @jjcard)

- `%1` in translation string

- translators don't have to deal with confusing strings in transifex


jjcard:
I was wondering why I was seeing QStrings that were just arguments. Makes sense now.

tooomm:
The idea was to prevent that translators could break things and to make it simpler for them in general.
We split html code for layouting as well for example. There is still one ticket open which covers a lot of these cases.

PR #2826, opened 2017-08-08
more repo cleanup

tooomm:
- license file had unusual name (maybe ancient leftover from brukie days?)
- appveyor config file also supports same naming scheme as other services with starting dot: https://www.appveyor.com/docs/build-configuration/#yaml-file-alternative-naming


tooomm:
Can I change the `Dockerfile` like that? No idea how they work...

skwerlman:
The name `COPYING` is normal for GPLd programs: https://stackoverflow.com/a/5678716

Daenyth:
I don't think this is especially useful. COPYING is standard as skwerlman says

tooomm:
Well, the internet mostly speaks about both if you search for gnu boilerplate...
Standard? I haven't seen too many `COPYING` files in other projects in general, most are `LICENSE`.
And you created a `LICENSE` file for our spoiler repo (with gpl) for yourself @Daenyth. ;)
So you chose the maybe more obvious name, or it's even GitHub's "standard" and proposal...

Anyway, we should just do it the same everywhere - what do we do for the org?
https://github.com/Cockatrice/Cockatrice/blob/master/COPYING (GPLv2)
https://github.com/Cockatrice/Magic-Spoiler/blob/master/LICENSE (GPLv3)
https://github.com/Cockatrice/Magic-Token/blob/master/LICENSE (The Unlicense)

skwerlman:
I agree with switching for the sake of consistency

tooomm:
I mean we can change all others instead, but it should be the same within the org. One way, or the other.

I realized there is no license yet for the gh page.

skwerlman:
While we're at it, is there a compelling reason to stay with GPLv2 for the main repo instead of GPLv3? Having a different license for each repo could wind up being confusing...

tooomm:
It would be great to have them all the same, I agree 💯 

But I've no idea if (or how) that would be possible.
Changing isn't that easy, but maybe an upgrade v2 to v3 is?

The "gh page" and "webclient" repo could more easily be adjusted maybe.
Our "token" repo has also very few contributors in case we have to ask every single one.

skwerlman:
Upgrading from v2 to v3 is extremely easy, since each version of GPL explicitly allows using later versions of the license at-will:
```
    This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation; either version 2 of the License, or
    (at your option) any later version.
```

There are lots of good reasons to switch away from the unlicense if we can: https://softwareengineering.stackexchange.com/a/147120/124258

Daenyth:
Agree about unlicense. Avoid using anything that hasn't been vetted by actual lawyers.

tooomm:
>Agree about unlicense. Avoid using anything that hasn't been vetted by actual lawyers.

@Daenyth How would we change that license? Is there like a step-by-step checklist?
Did anybody walk trough this process already?

Quickly reading https://opensource.stackexchange.com/questions/33/how-can-a-project-be-relicensed tells me that we might have to ask all contributors.
So that's just like 1-2 persons outside of the org... would we create an issue or pr for that purpose and collect `+1`'s or `PR Approval`'s as "permissions to change" and that's it?

Do we need to think about a CLA for future contributions in general? (and to avoid this problem again in the future) --> https://softwareengineering.stackexchange.com/questions/168020/how-signing-out-a-cla-prevents-legal-issues-in-open-source-projects

Is there a simple way to handle Contributor License Agreements?
Found several projects (basically all on github as well):
- https://cla-assistant.io/
- https://cla.github.com/
- https://www.clahub.com/
- https://colineberhardt.github.io/cla-bot/
- https://opensource.google.com/docs/cla/github/
- ...

Daenyth:
> would we create an issue or pr for that purpose and collect +1's or PR Approval's as "permissions to change" and that's it?

I think this is a reasonable path.

> Do we need to think about a CLA for future contributions in general? 

I've thought about it. I think for cockatrice specifically I don't want it - the entanglement of having contributors still own their contributions is a good thing imo. It prevents me or anyone else from changing to a non-free license in the future without a great effort.

I think in general as long as pick a reasonable FOSS license from the start for things, we don't need to worry about CLAs

skwerlman:
i personally think we should collect explicit approvals of the license change instead of +1's to avoid any ambiguities or re-interpretations later on.

something along the lines of

> I, \<username>, hereby allow all code and contributions previously made by me to this repository to be relicensed under the terms of the \<name of license>, and agree to use this license for all future code and contributions intended for this repository.

tooomm:
@Psithief should join the discussion here, as "main token repo guy". :)
Basically you might want start reading there: https://github.com/Cockatrice/Cockatrice/pull/2826#issuecomment-321342351

Psithief:
I don't even consider tokens.xml to _be_ software, that's why I applied the bizarre "unlicense" to it in the first place. It is an index that is parsed _by_ software. It may as well be a bitmap.

As for this repo:
----
I, an anonymous as-much-as-possible contributor, hereby waive whatever rights governments have decided someone using my accounts has to, or granted by, the following contributions to this project:

- https://github.com/Cockatrice/Cockatrice/commit/b05ed0376a88a0f553ddc2fcfcf7c20881a5577d
- https://github.com/Cockatrice/Cockatrice/commit/61893faf3b5359a2cce5158748e446e8c2255c54
- https://github.com/Cockatrice/Cockatrice/commit/03bb02f75b83f83b11ebeab1c4bc399ae84eef28
- https://github.com/Cockatrice/Cockatrice/commit/ae15f5df0684b6197d096853613c00999d4e01fb
- https://github.com/Cockatrice/Cockatrice/commit/f217551f5f67384caa43fc3eb41f0ce2132d65ad
- https://github.com/Cockatrice/Cockatrice/commit/30531683f4bc9efdd88d5ff2cf82b638e09cdefa

I would like to mention that the following commit is not authored by me, it is only something I found. The author is @tritoch 
- https://github.com/Cockatrice/Cockatrice/commit/bfcc48f5bf6f2cac07ac5b6dcd6774d09b4e929a

----

I don't think these declarations mean much without proof of identification, and that's not forthcoming.

PR #2829, opened 2017-08-13
link adjustments needed after #2826

tooomm:
## Related Ticket(s)
- follow up to #2826

old file names, links not working

- copying --> license


Daenyth:
We might look at using https://www.npmjs.com/package/markdown-link-check for validation in the future

PR #2834, opened 2017-08-16
update cards schema for new changes

ZeldaZach:
Seems our schema file was a bit out of date. This addresses it and validates against cards.xml generated from oracle.


ZeldaZach:
Seems like I had a bad file, new one works fine.

PR #2843, opened 2017-08-28
[WIP] Fix idle client disabling

woogerboy21:
Updated the idle client timeout value check to properly disable the  idle client checking routines during ping interval timeouts.  

Prior to this the code inside the loops were always entered and inner loops checked if time out value was set to 0.  This results in delayed server ping responses at higher user loads.

### DIFF
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -387,14 +387,14 @@ void Server_ProtocolHandler::pingClockTimeout()
     if (timeRunning - lastDataReceived > server->getMaxPlayerInactivityTime())
         prepareDestroy();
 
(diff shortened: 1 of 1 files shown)


skwerlman:
why was this closed?

woogerboy21:
I could re-open it if you like.  It simply moved the check for
clientidletimeout value (from the config) to the outer look rather than
having it in the two inner loops.  The actual underlying issue with delays
was a result of the 1 second client keep alive checks and user loads.

tooomm:
Can this branch be deleted @woogerboy21?

woogerboy21:
Yes it can.

PR #2858, opened 2017-09-21
add colourless keyword to filter

tooomm:
## Related Ticket(s)
- extends #2737 for missing term

british english was missing

- add `colourless` as option

<br>

Can somebody have a look at the AppVeyor builds, please?
There are some cmake errors due to deprecation and cmake_policy.


tooomm:
After https://github.com/Cockatrice/Cockatrice/pull/2859 we are good here? @ctrlaltca

PR #2859, opened 2017-09-22
Cmake: remove old unneeded POLICIES (fix Appveyor)

ctrlaltca:


ctrlaltca:
Nothing fancy here, but the last version of cmake used on Appveyor is complaining about our CMakeLists.txt containing old policies used by cmake version < 3. Since we require 3.1.2, we can safely remote them.

ctrlaltca:
CI succeded (travis's osx is currently broken)

PR #2872, opened 2017-09-29
add link to project roadmap

tooomm:
no way to find roadmap

- easy link in readme

Is there some kind of linking we can do without writing permissions?
Just realized that like this, everybody should be able to make suggestions to the file, I think we don't want to mess it up that easily.
If we can have a view-only link i would prefer to put that here instead @Daenyth.


Nightfirecat:
You'll need to use the "Publish to web" functionality on the doc.
Source: Fairly old article, but I believe it's still valid: http://sachachua.com/blog/2013/12/sharing-google-docs-one-link-to-edit-one-link-to-view/

PR #2885, opened 2017-10-04
Appveyor updates

ctrlaltca:
A lot of updates to appveyor builds:
 * now that we have a version number, use it in place of 0.0.1
 * use MS Visual Studio 2015 (was 2013)
 * use msbuild instead of nmake
 * download, compile and install protobuf and zlib instead of using old versions from nuget (cache the build results to avoid to recompile them at every commit)
 * Use Qt 5.9 for the 64bit package; the 32bit package still sticks with Qt5.6 since it's the last version compatible with Windows XP and Vista
 * removed external program usage (cygwin, 7zip, ...) and migrated to powershell

The new structure should make it easier to upgrade in the future, just changing the version number of packages.

Note: moving to a different version of the MSVC toolset will lead to people needing to install the 2015 version of the "vcredist" package.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,44 +1,33 @@
-version: 0.0.1-branch-{branch}-build-{build}
+version: 2.3.18-branch-{branch}-build-{build}
+image: Visual Studio 2015
 cache:
-    - c:\protobuf
-    - c:\protoc
+    - c:\protobuf-release
     - c:\zlib
 environment:
     matrix:
-        - vc_arch: amd64
-          choco_arch:
-          nuget_arch: x64
+        - nuget_arch: x64
           target_arch: x86_64
-          qt_ver: 5.6\msvc2013_64
-          bintray_path: Win64
-          MYSQL_DRIVER_URL: https://dev.mysql.com/get/Downloads/Connector-C/mysql-connector-c-6.1.6-winx64.zip
-          MYSQL_DRIVER_ARCHIVE: mysql-connector-c-6.1.6-winx64.zip
-          MYSQL_DRIVER_NAME: mysql-connector-c-6.1.6-winx64
-        - vc_arch: amd64_x86  # cross-compile from amd64 to x86
-          choco_arch: --x86
-          nuget_arch: Win32
+          qt_ver: 5.9\msvc2015_64
+          protobuf_ver: 3.4.1
+          cmake_generator: Visual Studio 14 2015 Win64
+          cmake_toolset: v140,host=x64
+        - nuget_arch: Win32
           target_arch: x86
-          qt_ver: 5.6\msvc2013
-          bintray_path: Win32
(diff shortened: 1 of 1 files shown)


Daenyth:
> Use Qt 5.9

Was there no qt issue after all?

tooomm:
I highly like the split you chose in appveyor.
One up-to-date build with Qt 5.9 on 64bit.
And a compatibility build with Qt 5.6 on 32bit. 👍 

>Note: moving to a different version of the MSVC toolset will lead to people needing to install the 2015 version of the "vcredist" package.

Isn't it more likely that newer versions of Windows come with the more recent versions packaged as well?

.appveyor.yml
```diff
@@ -1,80 +1,54 @@
-version: 0.0.1-branch-{branch}-build-{build}
+version: 2.3.18-branch-{branch}-build-{build}
```
tooomm: Do we have to hard code the version number here? :(
That's not too much of a benefit, right?

Since we would have to update that on every release, next to the version number update in the cmake file... Can't appveyor pull it from there?

.appveyor.yml
```diff
@@ -1,80 +1,54 @@
...
-          qt_ver: 5.6\msvc2013_64
-          bintray_path: Win64
-          MYSQL_DRIVER_URL: https://dev.mysql.com/get/Downloads/Connector-C/mysql-connector-c-6.1.6-winx64.zip
-          MYSQL_DRIVER_ARCHIVE: mysql-connector-c-6.1.6-winx64.zip
-          MYSQL_DRIVER_NAME: mysql-connector-c-6.1.6-winx64
-        - vc_arch: amd64_x86  # cross-compile from amd64 to x86
-          choco_arch: --x86
-          nuget_arch: Win32
+        - qt_ver: 5.6\msvc2015 # last version compatible with Windows XP, Vista
+          protobuf_ver: 3.4.1
```
tooomm: Travis still uses protobuf 2 instead?

.appveyor.yml
ZeldaZach: It's only used to display in AppVeyor so it's not a big deal

.appveyor.yml
ZeldaZach: Doesn't really matter, we're still on p2 syntax

tooomm: I know. I was just aiming for consistency between our ci build environments...

.appveyor.yml
tooomm: Especially then we should question it!
Nobody will take the effort - especially you Zach when you do a release - to update that number, just for some displaying purpose in one of the CI's...

And if we know beforehand that it will show a wrong/outdated number sooner than later, it might be better to just scrape it right away.
Results in better and a more helpful information displayed, and less confusing for all viewer.

If you look at appveyor, there are other issues with that version tag in general as is, too...

PR #2888, opened 2017-10-05
AppVeyor: prevent unnecessary builds

tooomm:
AppVeyor CI always started builds, also on completely unrelated changes.

- prevent builds on .md changes, like a README.md update
- prevent builds on config changes for other services, like travis, transifex, gitlab or github

Results in faster check marks for quicker merging. Also, important builds don't queue up and have to wait for such unrelated build to finish.
For example, travis is building this change to a appveyor config file, a total waste and delay!

Am I correct, that we don't want AppVeyor to build changes made in the webclient as well?
Did I miss anything?

---

~~This also removes the `qt5` extension for the windows builds. We don't use such a tag for our macOS and Linux builds.~~
Sadly this is needed as of now:
https://github.com/Cockatrice/Cockatrice/blob/ff6b0f86ec6d3bf76d6667003b82ce24479c0d15/cockatrice/src/releasechannel.cpp#L59-L68
We need a workaround for this to finally remove it in future versions. At best **before** releasing a new version. Just add an additional check for `-x86_64` and `-x86` without `qt5` for the transition period?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,9 +3,9 @@ version: 2.3.18-branch-{branch}-build-{build}
 # Skipping commits affecting specific files (GitHub only). More details here: /docs/appveyor-yml
 skip_commits:
   files:
-    - .github/*
-    - .travis/*
-    - .tx/*
+    - '.github/*'
+    - '.travis/*'
+    - '.tx/*'
     - webclient/*
     - .*ignore
     - .gitlab-ci.yml
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -69,9 +69,9 @@ build_script:
     - msbuild PACKAGE.vcxproj /p:Configuration=Release
     - ps: |
         $exe = dir -name *.exe
-        $new_name = $exe.Replace(".exe", "-${env:target_arch}.exe")
+        $new_name = $exe.Replace(".exe", "-${env:target_arch}_qt5.exe")
         Push-AppveyorArtifact $exe -FileName $new_name
-        $cmake_name = $exe.Replace(".exe", "-${env:target_arch}.cmake.txt")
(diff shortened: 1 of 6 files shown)


tooomm:
Tested and working with changes to these files:

- [.github/CONTRIBUTING.md](https://github.com/Cockatrice/Cockatrice/pull/2888/files#diff-8cc7b2b0d78dd2501610391c086a8516)
- [.tx/config](https://github.com/Cockatrice/Cockatrice/pull/2888/files#diff-cb86fa52440c3b29e9ad70ce1a82fe14)
- [.travis.yml](https://github.com/Cockatrice/Cockatrice/pull/2888/files#diff-354f30a63fb0907d4ad57269548329e3)
- [.dockerignore](https://github.com/Cockatrice/Cockatrice/pull/2888/files#diff-f7c5b4068637e2def526f9bbc7200c4e)
- [README.md](https://github.com/Cockatrice/Cockatrice/pull/2888/files#diff-04c6e90faac2675aa89e2176d2eec7d8)

All of these correctly didn't trigger AppVeyor builds! 👍 
Only untested is a (nested) change in `webclient/` folder, all other rules are working.

tooomm:
There seems to be a problem with the GitHub status check and how it expects results for `Required` builds.
It looks like this makes GitHub expect these builds to always happen, therefore the "waiting for pending" status unrelated if there is or going to be a build at all.

This looks to be a very minor issue, given the benefit.
Admins can still merge such PR's, or we could loosen the `Required` condition and make it a general policy instead a platform controlled enforcement.
But I'm also in contact with AppVeyor support to find a solution.

ctrlaltca:
This seems to work: the last build has been skipped.

tooomm:
Linking AppVeyor support ticket and discussion: https://github.com/appveyor/ci/issues/1848

PR #2901, opened 2017-10-23
Sorting by P/T numerically

### LINKED ISSUE #2884: p/t sorting in deck editor not correct
From Gitter:
>I don't know if this has already been mentioned or fixed but just noticed that when sorting cards by p/t they are ordered alphabetically rather than numerically as-in I believe the p/t for cards are being read as a string rather than an integer. ex. cards with power 10 being placed with cards of power 1 because they both start with a 1

Source:  :point_up: [September 3, 2017 11:24 AM](https://gitter.im/Cockatrice/Cockatrice?at=59abca59614889d475b67513)

<br>

Verified with Version: 2.3.18-custom_a01c515 (2017-09-27) on Win 10.

Some screenshots:

jjcard:
## Related Ticket(s)
- Fixes #2884

- Sorting of P/T was by Strings, so 10's were right after 1s, etc.

- have sorting by P/T field by trying to parse out the Power and toughness from the string and sorting by it.
  - ended up being a little more complicated of code then I want, but it's the best I could do over the weekend.

## Screenshots
ptsorting1
ptsorting2


Nightfirecat:
How are decimal numbers handled? How about cases where only the toughness is "X+*"? (are there any?) The screenshots I see look good!

cockatrice/src/carddatabasemodel.cpp
```diff
@@ -181,9 +181,90 @@ bool CardDatabaseDisplayModel::lessThan(const QModelIndex &left, const QModelInd
...
+            else {
+                //power equal, check toughness
+                return lessThanNumerically(leftList.at(1), rightList.at(1), &equal);
+            }
+        }
+    }
     return QString::localeAwareCompare(leftString, rightString) < 0;
 }
 
+bool CardDatabaseDisplayModel::lessThanNumerically(const QString &left, const QString&right, bool *equal) const {
```
Daenyth: I don't think we should use pointer to bool as an "out" parameter. I believe it's idiomatic to return a value in `{-1, 0, 1}` for LT, EQ, GT respectively.

In general, prefer using a pure functional approach.

cockatrice/src/carddatabasemodel.h
```diff
@@ -61,6 +61,7 @@ class CardDatabaseDisplayModel : public QSortFilterProxyModel {
     int rowCount(const QModelIndex &parent = QModelIndex()) const;
 protected:
     bool lessThan(const QModelIndex &left, const QModelIndex &right) const;
+    bool lessThanNumerically(const QString &left, const QString&right, bool *equal) const;
```
Daenyth: Can we make this a static method? It will make it easier to add unit tests for

cockatrice/src/carddatabasemodel.cpp
jjcard: ah ok. That would have been the way I did it in Java, wasn't quite sure what was 'normal' way here. I'll fix that up.

Daenyth:
BTW, whoever deleted the comments from the codacy bot (I didn't do that) - please don't do that again

tooomm:
> BTW, whoever deleted the comments from the codacy bot (I didn't do that) - please don't do that again

Isn't it just you with access there?
It has been a while since the bot commented here, I expected it to be on purpose?

But I realized that it always links to Zach's Codacy from within the status check `Details` for me instead yours (or a Cockatrice one): `codacy.com/app/ZeldaZach/Cockatrice/...`
Config seems to be out of place then?

cockatrice/src/carddatabasemodel.cpp
```diff
@@ -181,9 +181,101 @@ bool CardDatabaseDisplayModel::lessThan(const QModelIndex &left, const QModelInd
...
+    if (left == right) {
+        return 0;
+    }
+
+    bool okLeft, okRight;
+    float leftNum = left.toFloat(&okLeft);
+    float rightNum = right.toFloat(&okRight);
+
+    if (okLeft && okRight) {
+        //return leftNum - rightNum;
```
Daenyth: 🔪 this can be removed

jjcard:
I can look at adding some unit tests, never done it in C++ before. Would that be in the tests/ folder at the root of the project?

Daenyth:
Yes, we're using googletest as the framework and they're located there. If you run into a specific question we're happy to help, and like I said this is ready to merge in any case.

tooomm:
Do you still want to look into tests @jjcard?
Let us know so we can merge it without them otherwise! Happy for the fix anyway. 👍 

Not saying we would not appreciate to have them. ;)

Daenyth:
We can merge this as-is and leave tests for a separate PR I think

PR #2909, opened 2017-10-30
translation fixes for transifex

tooomm:
Weird translation strings at transifex, like single `The` strings or color codes `U`, `B` etc.

- easier and clearer strings for translators
- some cleanup

---

I did this changes while talking to @ctrlaltca.
Since he seems to be overly busy recently but there are still a few questions open I still wanted to open this.
Hopefully somebody else can help out:
- [x] Why is there a `tr()` inside `.arg`:
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/dlg_update.cpp#L24
- [ ] I need help here: https://github.com/Cockatrice/Cockatrice/pull/2909#pullrequestreview-72810488


cockatrice/src/messagelogwidget.cpp
```diff
@@ -434,15 +434,15 @@ void MessageLogWidget::logKicked()
 void MessageLogWidget::logLeave(Player *player, QString reason)
 {
     soundEngine->playSound("player_leave");
-    appendHtmlServerMessage(tr("%1 has left the game (%2).")
-                                    .arg(sanitizeHtml(player->getName()), sanitizeHtml(reason)), true);
+    appendHtmlServerMessage(tr("%1 has left the game").arg(sanitizeHtml(player->getName()))
+							+ QString(" (%1).").arg(sanitizeHtml(reason)), true);
```
tooomm: Can somebody please double check this `true` statement?
Is the structure still correct after my change?

ctrlaltca: > Why is there a tr() inside .arg:

The name of the release channel is a translatable string, too, but is not part of the main string.

> Can somebody please double check this true statement?
> Is the structure still correct after my change?

Please don't split arguments out of translatable strings. you can't be sure they will be in the exact place for every language out there, and it makes it hard for translators to understand the context of the strings.

tooomm: >The name of the release channel is a translatable string, too, but is not part of the main string.

👍 

>Please don't split arguments out of translatable strings. you can't be sure they will be in the exact place for every language out there, (...)

That makes totally sense for a complete sentence. The word order ABC within one sentence might be BAC or CBA in different languages due to different grammar rules.
But I did not touch the sentence, only the attached explanation in brackets.
So all language specific grammar is still possible - the argument for player name can freely be placed to fit any language needs.
I only force the explanation in brackets (which only exists out of a single argument) to be placed behind that sentence. With the benefit that a translator has to deal with one argument less.

>(...) and it makes it hard for translators to understand the context of the strings.

That's what I'm trying to do here - make life easier for translators.
I'm pretty sure `%1 has left the game` is easier to translate than `%1 has left the game (%2).`, especially out of context!
Transifex just gives very limited details on context and if you're not into coding they're not very helpful. Also, the page doesn't help a translator in understanding what "%2" stands for, or how he needs to position it properly regarding his language grammar rules. If you don't know what it is, you can't place it correctly.

ZeldaZach:
@tooomm Can you either fix this up or close it? Just trying to prune open PRs a bit :)

PR #2915, opened 2017-11-03
Even out db collation to utfmb4_unicode_ci; fix #2835 ; fix #2218

### LINKED ISSUE #2218: Use `utf8mb4`, not `utf8` in Servatrice
I haven't had a chance to test how this affects the server yet, but it looks to me like Servatrice is using the wrong Unicode encoding in MySQL.

Basically, MySQL's `utf8` encoding supports all Unicode code points which are 3 bytes or smaller. Unicode, however, has a number of 4-byte code points which cause problems here.
To fix this, MySQL added the `utf8mb4` encoding in version 5.5.3, which allows for 4-byte code points, and is backwards compatible with the older `utf8` encoding.

More info on this: https://mathiasbynens.be/notes/mysql-utf8mb4

### LINKED ISSUE #2835: Mysql collation is inconsistent.
The collation between mysql tables is not consistent. 
cockatrice_sessions uses `utf8_unicode_ci`
while cockatrice_users uses `utf8_general_ci`

This causes errors while trying to run queries like
`SELECT COUNT(s.id) FROM cockatrice_sessions s INNER JOIN cockatrice_users u ON s.user_name = u.name WHERE u.admin != 0 AND s.end_time is NULL and id_server = 1`

quick workaround is to use COLLATE in the query
`SELECT COUNT(s.id) FROM cockatrice_sessions s INNER JOIN cockatrice_users u ON s.user_name = u.name COLLATE utf8_unicode_ci WHERE u.admin != 0 AND s.end_time is NULL and id_server = 1`

The quickest way to repair would be to either drop collate statements from table creation or add them to tables that do not have them.

ctrlaltca:
## Related Ticket(s)
- Fixes #2835
- Fixes #2218

Some columns and tables in the servatrice database used different collation, making it hard to execute queries.

All the tables (and their columns) are converted to utf8mb4_unicode_ci

### DIFF
--- a/servatrice/migrations/servatrice_0023_to_0024.sql
+++ b/servatrice/migrations/servatrice_0023_to_0024.sql
@@ -2,10 +2,30 @@
 
 SET FOREIGN_KEY_CHECKS=0;
 
--- short the ip_address columns to 45 chars (max length of an ipv6 address)
--- to ensure the field can beused as a key on mysql < 5.7
+-- short the "ip address" columns to 45 chars (max length of an ipv6 address)
+-- to ensure the field can be used as a key on mysql < 5.7
+-- (not all fields are actually keys, but better keep them uniform)
 ALTER TABLE `cockatrice_sessions` MODIFY COLUMN `ip_address` varchar(45) NOT NULL;
 ALTER TABLE `cockatrice_bans` MODIFY COLUMN `ip_address` varchar(45) NOT NULL;
+ALTER TABLE `cockatrice_log` MODIFY COLUMN `sender_ip` varchar(45) NOT NULL;
+ALTER TABLE `cockatrice_audit` MODIFY COLUMN `ip_address` varchar(45) NOT NULL;
+
+-- short the "user name" columns to 35 chars (current max length in servatrice)
+-- to ensure the field can be used as a key on mysql < 5.7
+-- (not all fields are actually keys, but better keep them uniform)
+ALTER TABLE `cockatrice_bans` MODIFY COLUMN `user_name` varchar(35) NOT NULL;
+ALTER TABLE `cockatrice_warnings` MODIFY COLUMN `user_name` varchar(35) NOT NULL;
+ALTER TABLE `cockatrice_warnings` MODIFY COLUMN `mod_name` varchar(35) NOT NULL;
+ALTER TABLE `cockatrice_games` MODIFY COLUMN `creator_name` varchar(35) NOT NULL;
+ALTER TABLE `cockatrice_games_players` MODIFY COLUMN `player_name` varchar(35) NOT NULL;
+ALTER TABLE `cockatrice_donations` MODIFY COLUMN `username` varchar(35) NOT NULL;
+
+-- remove the FK from cockatrice_activation_emails (it will be created again later)
+-- the key name should end with _1, but multiple run of the 0012_to_003 migration
+-- can lead to a different name or even multiple keys. In this case you must remove
+-- all of them before continue.
+-- Use "show create table cockatrice_activation_emails" to see the key names.
+ALTER TABLE cockatrice_activation_emails DROP FOREIGN KEY `cockatrice_activation_emails_ibfk_1`;
 
 -- unify tables and columns collation to utf8mb4_unicode_ci
 ALTER TABLE `cockatrice_schema_version` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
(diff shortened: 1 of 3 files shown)


skwerlman:
this also fixes #2218, right?

ctrlaltca:
> this also fixes #2218, right?

Right, i missed that ticket.

woogerboy21:
Ill be setting up a new server here in a day or so and report back on this.  Initially I don't see a problem and not expecting any issue's.

ctrlaltca:
The only potentially problematic point is that the migration script needs to temporary disable foreign key checks in order to convert tables to the new format, one by one.
From my tests once the foreign key checks are reenabled they are correctly working, but i made my tests on a local server with only a few users, so i take my test results with a grain of salt.

woogerboy21:
I plan on restoring my entire userbase into this new server so I'll let you know how it goes on a larger number of users.  I wouldn't think some one will be attempting an upgrade without having the app shutdown?

woogerboy21:
So I attempted to run the script against a restore of the userbase of the woogerworks server and it threw this error:

```
ERROR 1071 (42000) at line 22: Specified key was too long; max key length is 767 bytes
```

servatrice/migrations/servatrice_0023_to_0024.sql
```diff
@@ -0,0 +1,35 @@
...
+ALTER TABLE `cockatrice_rooms_gametypes` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_games` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_games_players` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_replays` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_replays_access` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_servers` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_uptime` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_servermessages` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_sessions` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+ALTER TABLE `cockatrice_bans` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
woogerboy21:
```
ERROR 1071 (42000) at line 22: Specified key was too long; max key length is 767 bytes
```

servatrice/servatrice.sql
```diff
@@ -204,7 +204,7 @@ CREATE TABLE IF NOT EXISTS `cockatrice_bans` (
   KEY `time_from` (`time_from`,`ip_address`),
   KEY `ip_address` (`ip_address`),
   FOREIGN KEY(`id_admin`) REFERENCES `cockatrice_users`(`id`)  ON DELETE CASCADE ON UPDATE CASCADE
-) ENGINE=INNODB DEFAULT CHARSET=utf8;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci;
```
woogerboy21:
```
ERROR 1071 (42000) at line 194: Specified key was too long; max key length is 767 bytes
```

ctrlaltca: Looks like the index on `ip_address` can't be created because 255*4 bytex exceeds the max key length of 767 bytes; this is a limit of mysql < 5.7. I'll lave a look at this soon.

woogerboy21: If I recall correctly this index was added to assist with slow queries during login when looking up users previous ban history.  If you wanted to keep the same collation settings there may be a way to improve that query in general and drop the key.

ctrlaltca: The max length of an ipv6 address is 46 chars, so we could just reduce the column size in order to make it fit. Can you please check if your database contains records where the ip_address length is greater than 46?

woogerboy21:
```
MariaDB [servatrice]> select max(length(ip_address)) from cockatrice_bans;
+-------------------------+
| max(length(ip_address)) |
+-------------------------+
|                      26 |
+-------------------------+
1 row in set (0.13 sec)
```

But by the looks of things they are current IPv4 address scheme.  There doesnt appear to be any IPv6 address's in this DB.

woogerboy21: Look-e there , I learned something new today.  The ```::ffff:``` is actually used on an IPv4 network to allow for IPv6 socket requests to be responded to.  Interesting.

ctrlaltca:
The last commit should fix the problem @woogerboy21 experienced.
Sorry for the delay.

woogerboy21:
Just managed to get around to testing this, I am still getting the error even after the update.
```
ERROR 1071 (42000) at line 27: Specified key was too long; max key length is 767 bytes
```

Looks like you added a line in the migration script:
```
-- short the ip_address columns to 45 chars (max length of an ipv6 address)
-- to ensure the field can beused as a key on mysql < 5.7
ALTER TABLE `cockatrice_sessions` MODIFY COLUMN `ip_address` varchar(45) NOT NULL;
ALTER TABLE `cockatrice_bans` MODIFY COLUMN `ip_address` varchar(45) NOT NULL;
```

ctrlaltca:
I'll take a look at the further needed changes, thank you

woogerboy21:
Not a DB guy, but if the issue is we have columns that are varchar(255) and that size exceeds a limit that column type is across a number of table columns with different reasons.

ctrlaltca:
Sorry again for the long delay, i fixed and tested the migrations again this time on mysql 5.5.

ZeldaZach:
Any news on this @woogerboy21 ?

woogerboy21:
Never had a chance to test this since the update, give me a bit to bounce a second test off it.

PR #2935, opened 2017-12-02
readme: always center toc

tooomm:
## Screenshots
https://github.com/Cockatrice/Cockatrice/blob/tooomm-patch-1/README.md


tooomm:
Will include in another readme pr at a later point

PR #2944, opened 2017-12-08
Fix for Aftermath and Split Cards when sending to Tappedout; fix #2942

### LINKED ISSUE #2942: send deck to tappedout.net: aftermath cards not displayed correclty
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
Client Version: 2.4.1-custom_c8e843d (2017-12-05)

__________

Looking at this deck (http://tappedout.net/mtg-deckpaste/05-12-17-akh-pre-release-deck/) you see that `Mouth // Feed` is displayed as `Mouth / Feed` and is correctly linked.
It looks like tappedout uses this naming convention for all aftermath cards: http://tappedout.net/mtg-card/mouth-feed/ instead the double slash "//" to separate both halves.

If you send a deck with a aftermath card to them though, it will show up only with the first half's name!
I had to manually edit the list on their page to find out about their naming convention.
If you check for the card `Start // Finish` in the same decklist, you'll see and understand. It only shows as `Start` and links to only that part of the card (the picture is correct though).
In the edit mode it also only shows `Start`.

**So for tappedout.net we need to send aftermath cards with a slighlty different name. `/` instead `//`**
Or maybe we just send the first name at all?

This might be an general split card issue. Did not test with former ones.

KBGrammer:
## Related Ticket(s)
- Fixes #2942 

Split cards show up as only the first half of the card when a list is uploaded to tappedout due to an inconsistency in card naming. Tappedout uses "Side1 / Side2" instead of "Side1 // Side2".

- This makes a few changes to how a card list is generated when it's pushed to Tappedout and adds a boolean flag that causes "//" to be replaced with "/" if the flag is true. The changes below fix the issue for all split black bordered cards. The only card that remains broken is the silver card "Who//What//When//Where//Why", which causes an "Unable to analyze the deck" error.

## Screenshots
This change generates the following list: http://tappedout.net/mtg-deckpaste/07-12-17-fij-deck-paste/

The test list is as follows (every black bordered split card):
```
// 62 Maindeck
// 32 Instant
1 Leave // Chance
1 Commit // Memory
1 Failure // Comply
1 Consign // Oblivion
1 Refuse // Cooperate
1 Stand // Deliver
1 Destined // Lead
1 Bound // Determined
1 Odds // Ends
1 Far // Away
1 Farm // Market
1 Fire // Ice
1 Dead // Gone
1 Heaven // Earth
1 Hide // Seek
1 Hit // Run
1 Spite // Malice
1 Night // Day
1 Order // Chaos
1 Prepare // Fight
1 Profit // Loss
1 Protect // Serve
1 Ready // Willing
1 Illusion // Reality
1 Research // Development
1 Reduce // Rubble
1 Start // Finish
1 Struggle // Survive
1 Wear // Tear
1 Trial // Error
1 Turn // Burn
1 Onward // Victory

// 30 Sorcery
1 Alive // Well
1 Appeal // Authority
1 Assault // Battery
1 Reason // Believe
1 Boom // Bust
1 Beck // Call
1 Claim // Fame
1 Crime // Punishment
1 Cut // Ribbons
1 Armed // Dangerous
1 Dusk // Dawn
1 Life // Death
1 Supply // Demand
1 Driven // Despair
1 Down // Dirty
1 Breaking // Entering
1 Flesh // Blood
1 Give // Take
1 Grind // Dust
1 Insult // Injury
1 Spring // Mind
1 Mouth // Feed
1 Pain // Suffering
1 Rags // Riches
1 Catch // Release
1 Never // Return
1 Rise // Fall
1 Pure // Simple
1 Toil // Trouble
1 Rough // Tumble
```

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -647,7 +647,7 @@ struct WriteToStream {
 	   stream << QString("%1 %2\n").arg(
 	       card->getNumber()
 	   ).arg(
-               card->getName().replace("//", "/").replace("Who / What / When / Where / Why", "Who/What/When/Where/Why")
+               card->getName().replace("//", "/")
            );
        }
     }
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -647,7 +647,7 @@ struct WriteToStream {
 	   stream << QString("%1 %2\n").arg(
 	       card->getNumber()
 	   ).arg(
-               card->getName().replace("//", "/")
+               card->getName().replace("//", "/").replace("Who / What / When / Where / Why", "Who/What/When/Where/Why")
            );
        }
     }
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -638,18 +638,17 @@ struct WriteToStream {
            stream << "SB: ";
        }
        if (!slashTappedOutSplitCards) {
+           stream << QString("%1 %2\n").arg(
+               card->getNumber()
+           ).arg(
+               card->getName()
+	   );
+       } else {
 	   stream << QString("%1 %2\n").arg(
 	       card->getNumber()
-	     ).arg(
-		   card->getName()
-	   );
-       }
-       else {
-	   stream << QString("%1 %2\n").arg(
-	       card->getNumber()
-	     ).arg(
-		   card->getName().replace("//", "/")
-	   );
+	   ).arg(
+               card->getName().replace("//", "/")
+           );
        }
     }
 };
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -626,8 +626,9 @@ bool DeckList::loadFromFile_Plain(QIODevice *device)
 struct WriteToStream {
     QTextStream &stream;
     bool prefixSideboardCards;
+    bool slashTappedOutSplitCards;
 
-    WriteToStream(QTextStream &_stream, bool _prefixSideboardCards) : stream(_stream), prefixSideboardCards(_prefixSideboardCards) {}
+  WriteToStream(QTextStream &_stream, bool _prefixSideboardCards, bool _slashTappedOutSplitCards):  stream(_stream), prefixSideboardCards(_prefixSideboardCards), slashTappedOutSplitCards(_slashTappedOutSplitCards) {}
 
     void operator()(
         const InnerDecklistNode *node,
@@ -636,32 +637,41 @@ struct WriteToStream {
        if (prefixSideboardCards && node->getName() == DECK_ZONE_SIDE) {
            stream << "SB: ";
        }
-       stream << QString("%1 %2\n").arg(
-           card->getNumber()
-       ).arg(
-           card->getName()
-       );
+       if (!slashTappedOutSplitCards) {
+	   stream << QString("%1 %2\n").arg(
+	       card->getNumber()
+	     ).arg(
+		   card->getName()
+	   );
+       }
+       else {
+	   stream << QString("%1 %2\n").arg(
+	       card->getNumber()
+	     ).arg(
+		   card->getName().replace("//", "/") //kbg
+	   );
+       }
     }
 };
 
-bool DeckList::saveToStream_Plain(QTextStream &out, bool prefixSideboardCards)
+  bool DeckList::saveToStream_Plain(QTextStream &out, bool prefixSideboardCards, bool slashTappedOutSplitCards)
 {
-    WriteToStream writeToStream(out, prefixSideboardCards);
+    WriteToStream writeToStream(out, prefixSideboardCards, slashTappedOutSplitCards);
     forEachCard(writeToStream);
     return true;
 }
 
-bool DeckList::saveToFile_Plain(QIODevice *device, bool prefixSideboardCards)
+bool DeckList::saveToFile_Plain(QIODevice *device, bool prefixSideboardCards, bool slashTappedOutSplitCards)
 {
     QTextStream out(device);
-    return saveToStream_Plain(out, prefixSideboardCards);
+    return saveToStream_Plain(out, prefixSideboardCards, slashTappedOutSplitCards);
 }
 
-QString DeckList::writeToString_Plain(bool prefixSideboardCards)
+QString DeckList::writeToString_Plain(bool prefixSideboardCards, bool slashTappedOutSplitCards)
 {
     QString result;
     QTextStream out(&result);
-    saveToStream_Plain(out, prefixSideboardCards);
+    saveToStream_Plain(out, prefixSideboardCards, slashTappedOutSplitCards);
     return result;
 }
 
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -648,7 +648,7 @@ struct WriteToStream {
 	   stream << QString("%1 %2\n").arg(
 	       card->getNumber()
 	     ).arg(
-		   card->getName().replace("//", "/") //kbg
+		   card->getName().replace("//", "/")
 	   );
        }
     }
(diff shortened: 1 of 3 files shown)


tooomm:
>The changes below fix the issue for all split black bordered cards. 

Basically it looks like you've all `Aftermath`, `Fuse` and `normal split cards` covered.

If I check gatherer for [cards which contain "//"](http://gatherer.wizards.com/Pages/Search/Default.aspx?name=+[//]) it gives me 126, devided by 2 (assuming each half shows up separatly) we have 63 in total. I guess only one missing in your list is [`Wax // Wane`](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=220502&part=Wane) then.
I found it totally on random... and it puzzled me to try to understand which cards your fix might not cover until I read [this article](https://mtg.gamepedia.com/Split_card).
I doubt that there are any other cards besides the mentioned ones?

Will install the client and try the code. Report back tomorrow! 👍 

Thanks for opening a PR @KBGrammer! Very much appreciated.

KBGrammer:
I somehow lost Wax / /Wane along the way, but that card works for me as well.

I was able to grab the html output with Who/What/When/Where/Why in the list. Tappedout says that it can't find "Who / What / When / Where / Why" and it returns html from the request but it doesn't pass the status check and it fails. The problem is just that all of the other cards are "Side1 / Side2" and that particular one is has no spaces in it. It all works if I make a special case for that card. Sending the other cards as "Side1/Side2" causes a "Card not found" error, but the silver card is found.

All of the black border ones plus that silver one work with this for line 650 of decklist.cpp:
`card->getName().replace("//", "/").replace("Who / What / When / Where / Why", "Who/What/When/Where/Why")`
instead of
`card->getName().replace("//", "/")`

I think that's just what has to be done because it's an inconsistency on Tappedout's end. I've changed that line in my fork, so it will be present when you test it tomorrow.

common/decklist.cpp
```diff
@@ -636,32 +637,40 @@ struct WriteToStream {
...
+           stream << QString("%1 %2\n").arg(
+               card->getNumber()
+           ).arg(
+               card->getName()
+	   );
+       } else {
+	   stream << QString("%1 %2\n").arg(
+	       card->getNumber()
+	   ).arg(
+               card->getName().replace("//", "/").replace("Who / What / When / Where / Why", "Who/What/When/Where/Why")
```
Daenyth: Instead of hardcoding a specific card name, what about doing `.replace("//", "/").replace(" / ", "/")`? Would that work?

Daenyth:
> the real name of the card is "Who/What/When/Where/Why".

I think you're right. Maybe this is just a bug in mtgjson and they can fix it so the card name comes out correct. I don't love the idea of having a hack in place for a single card, though we've done similar before

common/decklist.cpp
tooomm: I don't think so, because we have it as `Who // What // When // Where // Why` with double `//` in our database, the same as all other split cards.

Check my other post for more details @Daenyth!

tooomm:
>Maybe this is just a bug in mtgjson and they can fix it so the card name comes out correct.

Had the same thought and double checked.
As by mtgjson design they don't have split cards by their complete name, they include each side or part as single card and give them the split layout key. Additionally they include the other names.

From allsets.json:
```
      {
        "artist": "Dan Frazier",
        "cmc": 1,
        "colorIdentity": [
          "W",
          "R",
          "U",
          "B",
          "G"
        ],
        "colors": [
          "White"
        ],
        "id": "b71a0a7c75747bec5727284eb2a3536ca38ff17e",
        "imageName": "who what when where why",
        "layout": "split",
        "manaCost": "{X}{W}",
        "mciNumber": "120a",
        "multiverseid": 74358,
        "name": "Who",
        "names": [
          "Who",
          "What",
          "When",
          "Where",
          "Why"
        ],
        "number": "120a",
        "rarity": "Rare",
        "text": "Target player gains X life.",
        "type": "Instant",
        "types": [
          "Instant"
        ]
      },
```

It then is actually Oracle combining them to the single card we have in `cards.xml` with the combined name. For all split cards we use `side1 // side 2`. That also applies to the one from Unhinged, so it ends up as "Who // What // When // Where // Why" in our database instead [`Who/What/When/Where/Why`](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=74358):
```
<card>
            <name>Who // What // When // Where // Why</name>
            <set rarity="" muId="74358" num="120a">UNH</set>
            <color>W</color>
            <color>R</color>
            <color>U</color>
            <color>B</color>
            <color>G</color>
            <manacost>XW // 2R // 2U // 3B // 1G</manacost>
            <cmc>1 // 3 // 3 // 4 // 2</cmc>
            <type>Instant // Instant // Instant // Instant // Instant</type>
            <tablerow>3</tablerow>
            <text>Target player gains X life.

---

Destroy target artifact.

---

Counter target creature spell.

---

Destroy target land.

---

Destroy target enchantment.</text>
        </card>
```
If you correct the card database entry manually to "Who/What/When/Where/Why" the card is send to tappedout correctly because the have the right name too: http://tappedout.net/mtg-card/whowhatwhenwherewhy/

<br>

All that makes me wonder why
>All of the black border ones plus that silver one work with this for line 650 of decklist.cpp:
`card->getName().replace("//", "/").replace("Who / What / When / Where / Why", "Who/What/When/Where/Why")`
instead of
`card->getName().replace("//", "/")`

should work. We don't know about a card named "Who / What / When / Where / Why" in our database...

---

<br>

Conclusion:
 - we should fix normal split cards in the propper way (replacing `//` with `/`, keeping spaces)
 - think about how to handle the unh card
      - should it be correctly in our database?
      - where do we want to have the exception for that card? (if any)
      - what are other followup problems?

*Right now that card obviously causes the same problems when `send to deckstats.net`, too*

KBGrammer:
Since it's a name problem and that deckstats has a similar issue, I think it makes the most sense to correct the name in the database. I can look into providing a fix for that as well, if that's the course of action you both support and revert this one to the simpler line 650. If the problem is fixed in Oracle, there would be a backwards compatibility annoyance for anyone with a decklist saved with the wrong card name. The upshot is that anyone who pastes a deck from tappedout would no longer have to correct that card, and I expect they would have to correct it now.

I think a single card hack is inevitable, and it's a matter of where that hack goes.

Daenyth:
I'm not really concerned with breaking decklists for people using that card
- most folks don't play un-sets on here because there's some effects that
are simply impossible

KBGrammer:
That's fair. I'll revert this fix so that the silver card remains broken, but everything else works.

KBGrammer:
Line 650 has been reverted to the previous case, with no special case for Who/What/When/Where/Why:
`card->getName().replace("//", "/")`

The full list of 63 black bordered split/aftermath cards looks like this when imported to tappedout: [http://tappedout.net/mtg-deckpaste/09-12-17-FHK-deck-paste/](http://tappedout.net/mtg-deckpaste/09-12-17-FHK-deck-paste/)

ZeldaZach:
Thanks for your contribution!

tooomm:
One note for the future though:
You always ever want to create a separate branch for changes you're working on and sending PR's from @KBGrammer.
For this PR you simply edited your own master branch of the fork, that's not ideal. ;)

Cheers, and thanks again! 👍

KBGrammer:
Thank you very much for the tip. I will make sure I do that next time.

Daenyth:
The reason is to avoid ugly merge issues in your fork if things change.
Easiest to keep your forks master in line with upstream.

@zeldazach can you make sure we have something like that in the contributor
guide?

PR #2949, opened 2017-12-12
Fixing split card from UNH with wrong name in database (fix #2947)

### LINKED ISSUE #2947: split card from UNH has wrong name in cards.xml database
This got spotted in a recent PR (https://github.com/Cockatrice/Cockatrice/pull/2944) by @KBGrammer:

Oracle assembles the card [`Who/What/When/Where/Why`](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=74358) from its several pieces/sides out of mtgjson.
But it uses the same schema as for all other split cards, that's why it ends up in the *cards.xml* database as
`Who // What // When // Where // Why`:

```
<card>
            <name>Who // What // When // Where // Why</name>
            <set rarity="" muId="74358" num="120a">UNH</set>
            <color>W</color>
            <color>R</color>
            <color>U</color>
            <color>B</color>
            <color>G</color>
            <manacost>XW // 2R // 2U // 3B // 1G</manacost>
            <cmc>1 // 3 // 3 // 4 // 2</cmc>
            <type>Instant // Instant // Instant // Instant // Instant</type>
            <tablerow>3</tablerow>
            <text>Target player gains X life.

---

Destroy target artifact.

---

Counter target creature spell.

---

Destroy target land.

---

Destroy target enchantment.</text>
        </card>
```

This leads to follow up problems like the `Send to online service --> Analyze deck (tappedout.net)` not recognizing the card and simply returning an error for all request for example.
Every other service you use to copy your deck list to, most likely has the same problem if they use the cards real name.

KBGrammer:
## Related Ticket(s)
- Fixes #2947, related to #2942 and #2944

The card [Who/What/When/Where/Why](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=74358) has the wrong name when it is generated in the database. Oracle defaults to the separator " // " for all split cards, which produces the wrong name for this card alone. This caused problems when exporting a list containing that card to Tappedout or Deckstats, with Tappedout failing due to the unknown card and Deckstats generating a link but with an unknown card in it.

The card name generation is modified for split/aftermath cards with more than 2 components to use the separator "/" instead of " // ". Deckstats and Tappedout generation now works for this card (and the 63 other split cards). The fix would affect any possible future split card with 3 sub-cards, but in principle such a situation would probably require this piece of code to be revisited anyway and adjusting this UNH card fix might not be an issue.

## Screenshots
[Tappedout](http://tappedout.net/mtg-deckpaste/11-12-17-ONc-deck-paste/)
[Deckstats](https://deckstats.net/deck-14709113-25762c17e9d77ae5f8fe175418ae8f2c.html)

### DIFF
--- a/cockatrice/src/tappedout_interface.cpp
+++ b/cockatrice/src/tappedout_interface.cpp
@@ -78,8 +78,8 @@ void TappedOutInterface::getAnalyzeRequestData(DeckList *deck, QByteArray *data)
     QUrl params;
     QUrlQuery urlQuery;
     urlQuery.addQueryItem("name", deck->getName());
-    urlQuery.addQueryItem("mainboard", mainboard.writeToString_Plain(false));
-    urlQuery.addQueryItem("sideboard", sideboard.writeToString_Plain(false));
+    urlQuery.addQueryItem("mainboard", mainboard.writeToString_Plain(false, true));
+    urlQuery.addQueryItem("sideboard", sideboard.writeToString_Plain(false, true));
     params.setQuery(urlQuery);
     data->append(params.query(QUrl::EncodeReserved));
 }
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -647,7 +647,7 @@ struct WriteToStream {
 	   stream << QString("%1 %2\n").arg(
 	       card->getNumber()
 	   ).arg(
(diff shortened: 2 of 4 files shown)


tooomm:
Seems to work for both services!

Thanks again for a quick fix with a very detailed description. 👍

PR #2958, opened 2017-12-14
Filters phrase *cmdr* when loading from clipboard (helps with TappedOut decks)

JoshuaLamusga:
## Related Ticket(s)
- Not brought up in tickets.

If a user loads their deck from clipboard and it shares the same format that TappedOut does for EDH / commander decks, then one of their cards has \*cmdr* written after the card name. Previously, that card would not be loaded correctly and would show up as "_their card name_ \*cmdr*" under the unknown category.

The phrase \*cmdr* is removed from cards loaded from the clipboard so that EDH / commander decks can be imported easier.

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -551,6 +551,10 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         rx.setPattern("\\(.*\\)");
         line.remove(rx);
 
+        // Filter out the phrase *cmdr* for commander decks
+        rx.setPattern("\\s*\\*cmdr\\*\\s*");
+        line.remove(rx);
+
         // Filter out post card name editions
         rx.setPattern("\\|.*$");
         line.remove(rx);


Daenyth:
I really think in that case a user should just fix their decklist. I don't think this change is useful for the majority of people, nor is it really in line from removing mtg-specific code in favor of generalized behavior for multiple card games

ZeldaZach:
I'd disagree @Daenyth I find this change to be useful. It's not hurting anyone or anything for having this here, but maybe instead we can create a global card game list to remove misc text from lines to be more linear in our approach

Daenyth:
If you strongly feel it's beneficial I suppose it's not the worst thing we've added, but I am leery of expanding the scope of how the application behaves. I feel like we're currently underwater with insufficient developer bandwidth to address some of the severe issues, so I'm really reluctant to add stuff that IMO has a low benefit.

It looks like we can't reopen this PR since the source branch has been rebased since, but if you want to resubmit and merge then go ahead I suppose

JoshuaLamusga:
Since other notations we remove are common across multiple sites and the \*cmdr* notation isn't, I agree with @Daenyth it's best not to include it.

PR #2959, opened 2017-12-16
allow backspace/delete button to remove a filter

### LINKED ISSUE #2950: Deck builder should delete filters by pressing delete
It would be nice if filters in the deck builder could be removed by pressing delete. It would be more consistent with the behavior of the deck list, which allows the user to press delete to remove a card.

Interestingly, if a card in the deck list is selected and the user clicks a filter and presses delete to remove it, the highlighted card is deleted. Someone may need to remove focus from the deck list when the filter list gains focus (and vice versa).

ZeldaZach:
Fixes #2950 

If you use the backspace/delete key on a filter, it will remove that component of the filter. Extends the button press to the key bind, so there should be no averse results.


tooomm:
Works for <kbd>Backspace</kbd>, but not <kbd>Delete</kbd>!

PR #2960, opened 2017-12-16
Allow non-db cards to be moved around

### LINKED ISSUE #2954: Cannot sideboard unrecognized cards
Loading a deck from the clipboard can create invalid cards like `2x Example`, which is useful so you know which cards failed to import. I don't know if it's relevant for custom cards, but trying to put an unknown card like `Example` in the sideboard will cause it to cease to exist instead.

To reproduce:
1. Copy a deck with bogus card names to the clipboard.
2. Open a deck builder tab and press Ctrl + Shift + V to make a deck from it.
3. Try to sideboard one of the cards (double-clicking it or by selecting it and pressing S).

ZeldaZach:
Fixes #2954

Before we could not move cards around if they weren't in the DB b/c if the  card was not in the db, the program would discard the entry. There is now an overload variable that tells the program to pretend it's in the database with default values and continue as directed.

____

Load from Clipboard:
screenshot 2017-12-15 23 51 43

Highlight the row and press "S":
screenshot 2017-12-15 23 51 48

### DIFF
--- a/cockatrice/src/decklistmodel.cpp
+++ b/cockatrice/src/decklistmodel.cpp
@@ -264,11 +264,44 @@ QModelIndex DeckListModel::findCard(const QString &cardName, const QString &zone
     return nodeToIndex(cardNode);
 }
 
-QModelIndex DeckListModel::addCard(const QString &cardName, const QString &zoneName)
+QModelIndex DeckListModel::addCard(const QString &cardName, const QString &zoneName, bool anAddAnyway)
 {
     CardInfo *info = db->getCard(cardName);
     if (info == nullptr)
-        return QModelIndex();
+    {
+        if (anAddAnyway)
+        {
+            // We need to keep this card added no matter what
+            // This is usually called from tab_deck_editor
+            // So we'll create a new CardInfo with the name
+            // and default values for all fields
+            info = new CardInfo(
+                    cardName,
+                    false,
+                    nullptr,
+                    nullptr,
+                    "unknown",
+                    nullptr,
+                    nullptr,
+                    QStringList(),
+                    QList<CardRelation *>(),
+                    QList<CardRelation *>(),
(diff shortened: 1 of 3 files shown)


tooomm:
I like that.

What do you think about having the card name in red in the decklist for such cards?
Also, instead default values... it could have a hint in the `text` or ` name` field to explain why it has no values (or why it's displayed in red), like:
`This card does not exist in your card database!` or simply `Unknown card`

empty

Daenyth:
I like the idea of painting the background with a css style (so it's themable) defaulting to red when we don't have the card.

That can be a separate PR

cockatrice/src/decklistmodel.cpp
```diff
@@ -264,11 +264,44 @@ QModelIndex DeckListModel::findCard(const QString &cardName, const QString &zone
     return nodeToIndex(cardNode);
 }
 
-QModelIndex DeckListModel::addCard(const QString &cardName, const QString &zoneName)
+QModelIndex DeckListModel::addCard(const QString &cardName, const QString &zoneName, bool anAddAnyway)
```
Daenyth: what's the `an` in `anAddAnyway`?

Maybe rename to `addIfMissingFromDb`?

ZeldaZach: an = argument variable, n = integer... should be a b = boolean. Will fix this to be
`abAddAnyway`

tooomm:
I guess we are good here then after the var rename @Daenyth ?

cockatrice/src/decklistmodel.cpp
Daenyth: We don't use hungarian notation, please don't do that. We can tell it's an argument because it's in the argument list and we can tell it's a boolean because it's labelled `bool`.

PR #2961, opened 2017-12-16
clear search bar if all filters cleared

### LINKED ISSUE #2953: clear filters button resets card list even if there is a search entered
Confusing and problematic part is that the search isn't resetted or reapplied.

So it should either:
- a) apply the same entered search to the changed card list (because filters changed)
- b) or clear the search bar as well to make clear that there is no search applied any longer

ZeldaZach:
Fixes #2953

When you clear all filters, this will also clear the search bar. I consider that a shortcut filter in itself for names, so I feel it should be cleared.

If there is a disagreement, I can look at clearing filters then enforcing what was typed in the search bar.

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -823,6 +823,7 @@ void TabDeckEditor::actAnalyzeDeckTappedout()
 void TabDeckEditor::actClearFilterAll()
 {
     databaseDisplayModel->clearFilterAll();
+    searchEdit->setText("");
 }
 
 void TabDeckEditor::actClearFilterOne()


tooomm:
It looks like `Delete selected` and `Clear all filters` don't behave consistent regarding that.
Will check again and report back in 2 hours.

tooomm:
Jep, there is a small miss-alignment.
- `Delete selected filter`: will remove the selected filter + refresh card list based on leftover filters + refresh card list based on the former entered search ( = display cards which apply to filter+search)
- `Clear all filters`: will remove all filters + clear search ( = display all cards)

To reproduce:
- enter search for `of`
- filter for name `abandon`
- filter for color `r`
- delete filter for color `r`<br>
Then do the same but "clear all filters" to see the difference.

I feel like the first is more right.
If I delete all or a single filter, that doesn't imply that I want to purge a search.
Maybe I just mistyped "instantt" and want to delete and correct the filter. I still want to search for "lightning".

Search /= Filter

What do you think?

ZeldaZach:
I did look into how that interacted, and I feel it works as expected. If you clear only one filter then it will delete only that one. If you clear all, we will reset the field.

tooomm:
Reading your comment in the issue, I understand your thought process now:
>**I consider the search bar a "shortcut" filter**, so I'll fix this and just flush that bar

Considering that, the behavior makes sense that way.

PR #2962, opened 2017-12-16
Update Rarity Filters & Refactoring

### LINKED ISSUE #2952: rarity filter improvements
The **rarity filter** right now only works if you type the full rarity name without misspelling to make it find cards:
`common`, `uncommon` , ` rare`, `mythic rare` (or `special`)

Same like with color filter, where you can [type <kbd>c</kbd> to search for colorless](https://github.com/Cockatrice/Cockatrice/pull/2737), it should allow to search rarity with only `c`, `u`, `r`, `m`/ `mr` (and `s`)!

---

Something similar could apply to the card type filter as well maybe (not subtypes, only main):
`a`rtifact, `i`nstant, `c`reature, `s`orcery, `l`and, `e`nchantment, `p`lanswalker ...

ZeldaZach:
Fixes #2952 

This will implement rarity filter updates.

+ COMMON: c, common
+ UNCOMMON: u, uncommon
+ RARE: r, rare
+ MYTHIC RARE: m, mr, mythic rare

I also refactored the file I was working with as CLion said it was bad... Fixed up the colorless stuff as well so it's more in line with what the filter code written does.

I'll be fixing up contribution docs at some point this month.

### DIFF
--- a/cockatrice/src/filtertree.cpp
+++ b/cockatrice/src/filtertree.cpp
@@ -7,7 +7,7 @@
 template <class T>
 FilterTreeNode *FilterTreeBranch<T>::nodeAt(int i) const
 {
-    return ((childNodes.size() > i)? childNodes.at(i) : NULL);
+    return (childNodes.size() > i) ? childNodes.at(i) : nullptr;
 }
 
 template <class T>
@@ -22,21 +22,18 @@ void FilterTreeBranch<T>::deleteAt(int i)
 template <class T>
 int FilterTreeBranch<T>::childIndex(const FilterTreeNode *node) const
 {
-    FilterTreeNode *unconst;
-    T downcasted;
-
-    unconst = const_cast<FilterTreeNode *>(node);
-    downcasted = dynamic_cast<T>(unconst);
-    if (downcasted == NULL)
-        return -1;
-    return childNodes.indexOf(downcasted);
+    FilterTreeNode *unconst = const_cast<FilterTreeNode *>(node);
+    T downcasted = dynamic_cast<T>(unconst);
+    return (downcasted) ? childNodes.indexOf(downcasted) : -1;
 }
 
 template <class T>
 FilterTreeBranch<T>::~FilterTreeBranch()
 {
     while (!childNodes.isEmpty())
+    {
         delete childNodes.takeFirst();
+    }
 }
 
 const FilterItemList *LogicMap::findTypeList(CardFilter::Type type) const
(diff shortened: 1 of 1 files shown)


tooomm:
This seems to consider all available rarities for a card. 
So if there are different rarities for the same card in different sets it's found by both. 👍

cockatrice/src/filtertree.cpp
```diff
@@ -232,59 +256,71 @@ bool FilterItem::acceptCmc(const CardInfo *info) const
...
+     * Conditional statement will exit once a case is successful in
+     * replacement OR we go through all possible cases
+     */
+    for (int i = 0; converted_term.length() <= 3 && i <= 6; i++)
+    {
+        switch (i)
+        {
+            case 0: converted_term.replace("mr", "mythic rare", Qt::CaseInsensitive); break;
+            case 1: converted_term.replace("m r", "mythic rare", Qt::CaseInsensitive); break;
+            case 2: converted_term.replace("m", "mythic rare", Qt::CaseInsensitive); break;
```
tooomm: >MYTHIC RARE: m, mr, mythic rare

One might search for `mythic` only too.

cockatrice/src/filtertree.cpp
```diff
@@ -22,46 +22,51 @@ void FilterTreeBranch<T>::deleteAt(int i)
...
+    FilterTreeNode *unconst = const_cast<FilterTreeNode *>(node);
+    T downcasted = dynamic_cast<T>(unconst);
+    return (downcasted) ? childNodes.indexOf(downcasted) : -1;
 }
 
 template <class T>
 FilterTreeBranch<T>::~FilterTreeBranch()
 {
     while (!childNodes.isEmpty())
+    {
```
Daenyth: Can you adjust the style guide and remove the instruction to skip braces for 1-liner blocks?

ZeldaZach: I'll get around to updating all the files this week

ZeldaZach:
Fixed @tooomm

cockatrice/src/filtertree.cpp
JoshuaLamusga: It seems every new dev skips braces on 1-liners until they discover it leads to errors later. Glad to hear that's not the desired style.

Daenyth: Went that way originally because it's very nice in scala. Of course scala is expression-oriented so a typical 'missing brace' code error will pretty much always make compiling fail instead of silently being weird. Doesn't make sense for C++

PR #2963, opened 2017-12-16
add checkbox option to clear log

### LINKED ISSUE #2939: debug log is purged on log window close
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
2.4.1-custom_10ff266 (2017-12-04)

__________

If you open the `Debug Log` window it shows startup details and other stuff logged since you opened the app.
Once you close that window and reopen the dialog again... it is empty and only the client version is shown. But for sure you don't want to throw all log away on close.

It could be handy to have a `Clear` button (or better a `hide old log output / show former log output` toggle button) if you want to catch/document a certain error in a clean log.

ZeldaZach:
Fixes #2939

This will add a checkbox option (that is persistent) to the logs. If checked (default), logs will be cleared when closing the popup. Otherwise, they will be saved.

screenshot 2017-12-16 02 52 55

### DIFF
--- a/cockatrice/src/dlg_viewlog.cpp
+++ b/cockatrice/src/dlg_viewlog.cpp
@@ -1,17 +1,25 @@
 #include "dlg_viewlog.h"
 #include "logger.h"
+#include "settingscache.h"
 
 #include <QVBoxLayout>
 #include <QPlainTextEdit>
 
-DlgViewLog::DlgViewLog(QWidget *parent)
-: QDialog(parent)
+DlgViewLog::DlgViewLog(QWidget *parent) : QDialog(parent)
 {
+
+
     logArea = new QPlainTextEdit;
     logArea->setReadOnly(true);
     
-    QVBoxLayout *mainLayout = new QVBoxLayout;
+    auto *mainLayout = new QVBoxLayout;
     mainLayout->addWidget(logArea);
+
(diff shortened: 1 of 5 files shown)


skwerlman:
>  If checked (default), logs will be saved when closing the popup.

The phrasing on the checkbox (`Clear log when closing`) seems to indicate the opposite behavior...

ZeldaZach:
@skwerlman Haha my PR wording was wrong. Intended behavior is to clear by default. WIll fix in PR.

tooomm:
> What about removing clear on close entirely and adding a button that clears? This behavior would be very explicit and understandable. People don't notice checkboxes much.

That would change to `keep` logs, instead `clear` as default behavior.

I can't really tell what is better from a users perspective or for troubleshooting if they report issues.
Collect all log until I close the app (or hit `clear`) looks straight forward and maybe more expected.

What is easier and more intuitive?

Daenyth:
I don't have strong opinions on this

ZeldaZach:
I'm fine with how this is and don't see a need to look into an alternative solution. There are other pending issues so for now I find this acceptable

tooomm:
Tested it in client and it feels good as is. 👍 
People just need to make sure to copy the log before they close if they need to share it.

PR #2964, opened 2017-12-16
remove untap option, combine w/ tap

### LINKED ISSUE #2932: Combine/merge Tap+Untap option in menu
Instead having a separate `Tap` and `Untap` right click menu entry all the time for every card, it should be smarter and cleaner:
- Show `Tap` if card is untapped.
- Show `Untap` if card is tapped.

toggle_tap_untap

ZeldaZach:
Fixes #2932

Combines Tap and Untap in the right click menu of cards. Also removes the shortcut from the shortcut menu/tab since there's no longer a need and it is combined w/ tap.

General cleanup / fixes also in this on the files.

screenshot 2017-12-16 03 15 27
screenshot 2017-12-16 03 15 15

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -379,9 +379,6 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
     aTap = new QAction(this);
     aTap->setData(cmTap);
     connect(aTap, SIGNAL(triggered()), this, SLOT(cardMenuAction()));
-    aUntap = new QAction(this);
-    aUntap->setData(cmUntap);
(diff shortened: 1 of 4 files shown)


cockatrice/src/player.cpp
```diff
@@ -905,13 +900,13 @@ void Player::actOpenDeckInDeckEditor()
 
 void Player::actViewGraveyard()
 {
-    static_cast<GameScene *>(scene())->toggleZoneView(this, "grave", -1);
```
Daenyth: Why?

ZeldaZach: CLion told me scene() could be a problem so we should dynamically cast it instead.

PR #2968, opened 2017-12-18
About dialog build architecture fix #2948

### LINKED ISSUE #2948: no easy way to check if installed version is 32 or 64bit
In the client updater we also check already for which architecture the package was compiled for I guess.
It would be nice to expose that information in the `About` window next to the version number, too.

Very helpful for troubleshooting instead trying to find other indicators to make sure.

KBGrammer:
## Related Ticket(s)
- Fixes #2948

The About dialog does not currently display the build architecture.

About dialog has a new item called "Build Architecture" that calls buildAbi() and returns the architecture that Qt was build for, which I expect is the same as the cockatrice application. There's another way to get this information about the cockatrice application itself by checking which of Q_PROCESSOR_X86_32 and Q_PROCESSOR_X86_64 is defined. The buildAbi() way is consistent with how this information is gathered for application updates (see releasechannel.cpp).

## Screenshots
about_dialog

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -283,7 +283,7 @@ void MainWindow::actExit()
 void MainWindow::actAbout()
 {
     QMessageBox mb(QMessageBox::NoIcon, tr("About Cockatrice"), QString(
-	"<font size=\"8\"><b>Cockatrice</b></font>(" + QString::fromStdString(BUILD_ARCHITECTURE) + ")<br>"
+	"<font size=\"8\"><b>Cockatrice</b></font> (" + QString::fromStdString(BUILD_ARCHITECTURE) + ")<br>"
         + tr("Version") + QString(" %1").arg(VERSION_STRING)
         + "<br><br><b><a href='" + GITHUB_PAGES_URL + "'>" + tr("Cockatrice Webpage") + "</a></b><br>"
         + "<br><br><b>" + tr("Project Manager:") + "</b><br>Gavin Bisesi<br><br>"
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -32,6 +32,7 @@
 #include <QApplication>
 #include <QtNetwork>
 #include <QtConcurrent>
+#include <QSysInfo>
 
 #include "main.h"
 #include "window_main.h"
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -32,7 +32,6 @@
 #include <QApplication>
 #include <QtNetwork>
 #include <QtConcurrent>
-#include <QSysInfo>
 
 #include "main.h"
 #include "window_main.h"
@@ -282,7 +281,6 @@ void MainWindow::actExit()
 
 void MainWindow::actAbout()
 {
-    QString arch = QSysInfo::buildAbi().split('-')[0];
     QMessageBox mb(QMessageBox::NoIcon, tr("About Cockatrice"), QString(
         "<font size=\"8\"><b>Cockatrice</b></font><br>"
         + tr("Version") + QString(" %1").arg(VERSION_STRING)
@@ -299,7 +297,14 @@ void MainWindow::actAbout()
         + "<a href='" + GITHUB_ISSUES_URL + "'>" + tr("Report an Issue") + "</a><br>"
         + "<a href='" + GITHUB_TROUBLESHOOTING_URL + "'>" + tr("Troubleshooting") + "</a><br>"
         + "<a href='" + GITHUB_FAQ_URL + "'>" + tr("F.A.Q.") + "</a><br>")
-        + "<br><b>" + tr("Build Architecture:") + "</b><br>" + arch + "<br>",
+        + "<br><b>" + tr("Build Architecture:") +
+#ifdef Q_PROCESSOR_X86_32
+        + "</b><br>X86_32<br>",
+#elif defined(Q_PROCESSOR_X86_64)
+       + "</b><br>X86_64<br>",
+#else
+       + "</b><br>unknown<br>",
+#endif
         QMessageBox::Ok, this
     );
     mb.setIconPixmap(QPixmap("theme:cockatrice").scaled(64, 64));
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -281,6 +281,7 @@ void MainWindow::actExit()
 
 void MainWindow::actAbout()
 {
+    QString arch = QSysInfo::buildAbi().split('-')[0];
     QMessageBox mb(QMessageBox::NoIcon, tr("About Cockatrice"), QString(
         "<font size=\"8\"><b>Cockatrice</b></font><br>"
         + tr("Version") + QString(" %1").arg(VERSION_STRING)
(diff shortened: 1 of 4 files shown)


KBGrammer:
This is failing in travis because `QSysInfo::buildAbi()` was added in Qt 5.4 and travis uses 5.2.1, so it's throwing an error that `‘buildAbi’ is not a member of ‘QSysInfo’`. This code compiles just fine on my version of Linux Mint. Identical code appears in `releasechannel.cpp` but it's inside an `#ifdef` wrapper that checks to see if it's building for Windows so travis never sees it.

What should I do next? I could put it inside a wrapper that checks for Windows, but then Linux and OSX need their own solution. It's possible that the `Q_PROCESSOR_X86_32` / `Q_PROCESSOR_X86_64` solution works instead. I'm pushing that version now.

KBGrammer:
I've been using cockatrice for a few years now, so I thought I might as well contribute something.

Not a problem on the location of this information on the About dialog. I figured it would be better in a place like that in the end but just demonstrating it works was a good first step.

In looking up how to do this, I came across a large list of `#defines` for various architectures. Adding something for ARM will not be a problem.

cockatrice/src/window_main.cpp
```diff
@@ -296,7 +296,15 @@ void MainWindow::actAbout()
         + "<b>" + tr("Support:") + "</b><br>"
         + "<a href='" + GITHUB_ISSUES_URL + "'>" + tr("Report an Issue") + "</a><br>"
         + "<a href='" + GITHUB_TROUBLESHOOTING_URL + "'>" + tr("Troubleshooting") + "</a><br>"
-        + "<a href='" + GITHUB_FAQ_URL + "'>" + tr("F.A.Q.") + "</a><br>"),
+        + "<a href='" + GITHUB_FAQ_URL + "'>" + tr("F.A.Q.") + "</a><br>")
+        + "<br><b>" + tr("Build Architecture:") +
+#ifdef Q_PROCESSOR_X86_32
+        + "</b><br>X86_32<br>",
```
Daenyth: Can you use "32 bit" and "64 bit" please? Our users will understand that better

tooomm:
The problem with missing dll's or not working installers because of architecture mismatches for some users are a windows-only problem. Linux and macOS versions of Cockatrice are always 64bit... so there should be nothing wrong with using the same code from the updater and wrapping it with `ifdef` to have it on Windows only.

**Or do we want to display it always for all operating systems?**
How do other Linux/macOS applications handle this?

<br>

>This is failing in travis because QSysInfo::buildAbi() was added in Qt 5.4 and travis uses 5.2.1

Manually updating Qt on travis is not possible for us unless we switch to the slower `sudo: enabled` build infrastructure. And downloading and installing the huge Qt package might slow things down even more.
But with the next upcoming travis update to 16.04 LTS they'll support Qt5.5.1 at least.
It's "around the corner" they say...

KBGrammer:
The `Q_PROCESSOR` version works in travis currently, at least. I'm exclusively using Linux, and the `Q_PROCESSOR` version works when I compile my copy locally. I did not find any suggestion that `Q_PROCESSOR` would work on Linux but not Windows or Mac, and I believe it's just part of a comprehensive list of architecture flags. There are other flags in the same family that would signify ARM and others, so I expect that going this route should provide a platform independent way to say "Cockatrice - <something>" at the top of the About dialog, and provide similar information in the debug logs.

I think a virtue of the `Q_PROCESSOR` version is that it's Qt's way of having an application directly convey this information about itself, whereas `buildAbi` is giving you that information about the Qt library that was used to build the application instead.

tooomm:
`Q_Processor` works on Windows, too:
- Q_PROCESSOR:
     - 32:
q32
    - 64:
q64

- buildAbi()
    - 32:
b32
     - 64: 
b64

<br>

>Also, if they're on an ARM chip we should have information for them as well.

Will Cockatrice run on ARM at all? I honestly can't tell. :D
I know that @ctrlaltca had some hacks ongoing to make a partially working app for Android...

KBGrammer:
The About dialog is updated as follows:
about

There's a new function in `logger.cpp` called `getBuildArchitecture` that adds three new lines to the logger dialog screen, command line, and the log file. When the logger dialog is closed and re-opened such that the text is cleared, these three lines are appended to the beginning of a new log just like the client version is currently. The top of the logger stream now looks like this:

> Client Version: 2.4.1-custom_accd39f (2017-12-17)
Client Operating System: Linux
Build Architecture: 64-bit
Qt Version: 5.5.1

Possible results for Client OS are Linux, macOS, Windows, Android, and "unknown". Possible results for Build Architecture are 64-bit, 32-bit, ARM, and "unknown". The Qt version is taken from the defined string `QT_VERSION_STR`, since I figured that might be useful information if someone were to share a log file if they're having problems. Other possibilities can be found here: http://doc.qt.io/qt-5/qtglobal.html, though none looked like they would be obviously useful.

cockatrice/src/window_main.cpp
```diff
@@ -281,8 +281,17 @@ void MainWindow::actExit()
 
 void MainWindow::actAbout()
 {
+#if   defined(Q_PROCESSOR_X86_32)
```
Daenyth: Shouldn't this section be invoking `getBuildArchitecture` instead of repeating the ifdef logic?

KBGrammer: In the same way that `getClientVersion` writes the string "Client Version: 2.4.1-custom_accd39f (2017-12-17)", `getBuildArchitecture` writes all three of the other lines. I think that having those #defines there is analogous to how the client version is added to the About dialog via `QString(" %1").arg(VERSION_STRING)`.

In the interest of having one set of `#ifdef` in one place for these, perhaps it makes more sense to add two more defined strings for `CLIENT_OS` and `BUILD_ARCHITECTURE` so that both `getBuildArchitecture` and `actAbout` are just returning whatever string is stored in those?

Daenyth: Yes please, I really prefer to avoid propagating ifdefs around the codebase because it usually means we are missing some encapsulation or abstraction that would make things more readable and maintainable

tooomm:
>Possible results for Client OS are Linux, macOS, Windows, Android, and "unknown"

`Q_OS_*` should only return the operating system during compile time, not on run time...
So "**Client** Operating System" is not correct and having it only display very general information isn't too helpful after all. Also, it returns no further details about the used version which is very much needed for troubleshooting, our main use case of the log.

Therefore I suggest to use [QSysInfo::prettyProductName()](http://doc.qt.io/qt-5/qsysinfo.html#prettyProductName) there instead.
It should return more useful details like `"Windows 10 (10.0)"` or `"Fedora 26 (26)"` as the **running client os**!
I know - same as all QSysInfo - it's Qt 5.4+ only. But we can include it for newer Qt versions at least, it's only the old, automated Ubuntu 14.04 version left out right now.
It's not too bad if we display nothing for that Ubuntu build with Qt5.2 right now, it has a very low user count and it'll get updated to 16.04 with Qt5.5 anyway soon... on all other recent platforms it works. 👍 
(Build architecture should stay q_processor!)

>Possible results for Build Architecture are 64-bit, 32-bit, ARM, and "unknown"

Question still stands: Do we need to cover ARM at all?

<br>

Good idea with showing off the Qt version in the logs, too! 👍 
I agree, it might be the only other useful one they have there.

KBGrammer:
`QSysInfo::prettyProductName()` would have to sit inside a Qt version check wrapper for now, but there's a `#define` for that. I'll put something together later today that consolidates to one set of `#ifdef` and works better for 5.4+.

KBGrammer:
There is not a single set of `#define` for 64, 32, and ARM as mentioned above. This same `#define` is used in the About dialog generation and the previous screen shot shows what this looks like. There's a new function that grabs `QSysInfo::prettyProductName()` if the Qt version is 5.4 or higher, and returns an empty string otherwise. If the returned string here is empty, then the OS information is skipped in the logger and it moves on to Build Architecture and the Qt version.

tooomm:
```
Client Version: 2.4.1-custom_61698b7 (2017-12-20)
Build Architecture: 64-bit
Qt Version: 5.9.2
```

Which os will it return for your "Linux"? @KBGrammer

KBGrammer:
I'm not on that machine right now, but it was something along the lines of "Ubuntu Linux 14.04.5 (LTS 14.04)".

tooomm:
That log info sounds way more helpful for troubleshooting! 👍 

Does it return something as useful for mac, too @ZeldaZach?

ZeldaZach:
screenshot 2018-01-10 13 42 28

```
Client Version: 2.4.1-custom_1b4a607 (2017-12-19)
Client Operating System: macOS 10.13
Build Architecture: 64-bit
Qt Version: 5.7.1
```

I'm fine with this

Daenyth:
And potentially we want to `tr` some stuff but again, we can fix that later if needed

ZeldaZach:
@Daenyth I can fix up the names/tr's myself and then merge this in later

tooomm:
@KBGrammer Thanks again! This should help with troubleshooting a lot. :)

PR #2969, opened 2017-12-18
Created migration script

Lachee:
Python3 script that enables automated migration to newer schema. Provides functionally to pause after each iteration and cancel or to continue uninterrupted. Designed to work on linux based systems. Specifically designed with Ubuntu 16 in mind.

## Usage
To use, first create a mysql configuration, with `mysql_config_editor set --login-path=<configname> --host=localhost --user=<username> --password`. This will create a safe configuration for the mysql and prevent annoying messages (also make passwords invisible in console).

Then run the script with `Python3 migrate.py`. The script does not accept any arguments. On screen, it will ask for --login-path (created previous setup), database (normally `servatrice`), starting schema (schema the database is currently on) and if a pause should be made between each migration (for error checking).

### DIFF
--- a/servatrice/migrations/migrate.py
+++ b/servatrice/migrations/migrate.py
@@ -1,0 +1,73 @@
+import os.path
+
+start = 1;
+mysql_config = "migrate";
+database = "servatrice";
+safe_mode = True;
+
+print("========================================");
+print("-- Python Mirgation Tool - By Lachee ---")
+print("In order to safely migrate into the database, a mysql config needs to be setup.");
+print("Please make sure you have created a temporary one by using the mysql_config_editor.");
+print("eg: mysql_config_editor set --login-path=<configname> --host=localhost --user=<username> --password");
+print("========================================");
+print();
+
+mysql_config = input("--login-path: ")
+database = input("database: ");
+start = int(input("current schema: "))
+do_safemode = input("pause after each migration? y | n: ");
+if (do_safemode == 'n'):
+    safe_mode = False;
+
+#Does all the big scary things here
+print("Migrating from schema version " + str(start) + "...");
+current = start;
+iterating = True;
+while (iterating):
+
+    #Prepare the name
+    start_fill = str(current).zfill(4);
+    next_fill = str(current + 1).zfill(4);
+    filename = "servatrice_" + start_fill + "_to_" + next_fill + ".sql";
+
+    #make sure the file exists
+    if (not os.path.exists(filename)):
+
+        #We havn't found the file, we must have reached the end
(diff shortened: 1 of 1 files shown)


Daenyth:
I suppose this is reasonable, but if we're bringing in python is there any reason not to just use alembic instead?

tooomm:
Would it be useful to have the little explanation on how to use it as comments on top of the file as well?

ZeldaZach:
Is this the most optimal version of the script? Can it be rewritten to be more efficient / compact?

skwerlman:
My thoughts on the script:
- `import os`, not `import os.path`; as is, the script _should_ fail at line 48
- make this work on both py2 and py3 since ubuntu refuses to let py2 die
- add a hashbang so i can just `./migrate.py`
- format the script so it passes `pycodestyle` without errors
- the main loop would be a lot clearer if the migration scripts were gathered and sorted before iterating
- all the inputs at the start should be arguments instead

Daenyth:
You can install python3 on ubuntu just fine, even 16.04 (with a ppa), so 👎 for python2 support - it's non-trivial effort over time, and I'd like to keep scope minimal.

So far no one has answered though, if we are making server operators use python, then why not use alembic to manage the migrations instead of all our hand rolled stuff?

ctrlaltca:
Bash oneliner:
```sh
for i in servatrice/migrations/servatrice_{0000..0099}_*.sql; do mysql -uuser -ppass servatrice < $i; done
```
Imho such a script could be useful if it makes some automated checks, eg:
 * it makes a full backup before even touching the db;
 * it automatically detects the current schema version (`mysql -uroot -p servatrice -sN -e 'select version from cockatrice_schema_version'`);
 * it runs all the migrations queries automatically one by one and it's able to detect if a single query  fails and stop before ruining the db;
 * it refuses to run a migration if it has already been run or if the needed previous ones have not been run.

Daenyth:
> it makes a full backup before even touching the db;

Alembic doesn't do this by default but it could be made to do so

> it automatically detects the current schema version (mysql -uroot -p servatrice -sN -e 'select version from cockatrice_schema_version');
> it runs all the migrations queries automatically one by one and it's able to detect if a single query fails and stop before ruining the db;
> it refuses to run a migration if it has already been run or if the needed previous ones have not been run.

Alembic does all this

Daenyth:
Alternately, skipping python entirely, https://flywaydb.org/

ZeldaZach:
So I'm going to close this PR for the time being. If you ever feel like working with the tools Gav suggested, please do so! Thanks for the time and effort @Lachee :)

PR #2974, opened 2017-12-18
appveyor: restore 32bit compatibility

tooomm:
AppVeyor release binary was not working on 32bit systems correctly.
All fonts were crossed out for some reason.

Report from user in Gitter (Dez 2017):
2 4 0 with xp

Go back to `msvc2013` for 32bit + XP-compatible build
 (switched to all msvc2013 in https://github.com/Cockatrice/Cockatrice/pull/2885)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -29,11 +29,11 @@ environment:
           cmake_generator: Visual Studio 14 2015 Win64
           cmake_toolset: v140,host=x64
           target_arch: x86_64
-        - qt_ver: 5.6\msvc2013 # last version compatible with Windows XP, Vista and 32bit systems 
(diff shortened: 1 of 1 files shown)


ctrlaltca:
The current build is using Qt 5.6 and MSVC 2015. It works fine under Windows 7 (32 bit), but under Windows XP font rendering is.. funny.
We used MSVC2013 in the past for a few releases, and saw no report about this issue, so getting back to that version should be fine.
I've tested this version to run fine on Windows 7 32bit. It would be nice if anyone with windows xp could test it.

Daenyth:
I see no reason for us to support windows XP when microsoft doesn't. We don't have the time to spend on it with so much backlog of changes we could be doing instead.

MS stopped supporting XP in 2014. It's time to move on.

Issue #2921, opened 2017-11-19
deployment from ci is tagging all relases as prerelease


tooomm:
Either appveyor or travis (both?) are resetting the release notes while updating the beforehand published release (uploading their release artifacts).

If you can provide more details I could have a look @ZeldaZach 
- Is all text purged? Is it replaced with some other text (like `Dev build of Cockatrice`)?
- Do you fully release the draft first? Or do you publish it as pre-release?

ZeldaZach:
All text is purged and replaced w/ Dev build phrase. I publish as a full release and it moves it to pre-release. After it's done with the uploads I re-upload the release notes and publish as full.

tooomm:
Ok, we can improve that.
Parts of it are still based on older release handling... I guess it was used for development snapshots as well regarding the phrase.
Also, travis and appveyor handle things a bit different. There might always be little conflicts because both have to deal with the same release on GitHub.

>All text is purged and replaced w/ Dev build phrase.

Shouldn't happen anymore with https://github.com/Cockatrice/Cockatrice/pull/2976
Let's test it with the next `-beta1` you're going to release sooner than later! ;)

>I publish as a full release and it moves it to pre-release. 

Right now it needs to be a bit generic because it has to work for stable and beta releases, they use the same deploy config. Therefore it's better to move both to `Pre-release` and manually correct stable releases than having all beta releases going to stable automatically.

<br>

To get it right, and have proper configuration options for both release types - `Beta` and `Stable` - we need to setup a second deploy provider and run specific commands based on different conditions.
I hope we can use the git tag, check if it contains `beta` for example and use the second deploy provider with its own config on that condition.
Name info for the git tag is already there with `APPVEYOR_REPO_TAG_NAME` and `TRAVIS_TAG`.
I'll try to check that.

PR #2976, opened 2017-12-19
ci: deployment config for beta+stable

tooomm:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/2921

All drafted changelog text + release name were purged and replaced with "Dev build" phrase and a wrong name for all kind of releases. Also, every release was tagged as *pre-release*.
In return one always has to adjust each published release, no custom config options based on release type were possible.

AS @ZeldaZach put it for stable releases:
```
--Build happens and replaces stuff--
Re-paste release notes
Re-paste release name
Re-publish to no longer be pre-release
```

With this PR all of above are no longer needed, and it allows for a lot more customization and other future tweaks.

- Utilize multiple deploy provider entries with custom conditions
- Separate config for `Beta` and `Stable` on both CI's
     - Betas will be published as *pre-release*, stable as *full release*
     - Stable releases shouldn't purge the beforehand drafted changelog any longer,
while Betas keep their beta phrase in the description and set a better name per default
- Which config to apply is taken from the GitHub `tag` in a env var (conditional deploy)
- The tag has to comply with [Semantic Versioning 2.0.0](https://semver.org/) schema:
   - `x.y.z` for stable, or `x.y.z-beta`/`x.y.z-beta.n` for beta
    - No leading 0 for single digit numbers
    - Beta appendix has to be `-beta` for the first (no `-beta.1` exists)
Follow ups have to be append a incrementing number, separated by `.` like `-beta.2` and up

The big advantage here is that we can handle both release types with unique configurations while the workflow is the same for both - we only need to tag a commit on GitHub.
The check for a correct semver prevents that a mistyped tag gets deployed and causes wrong file names for our binaries for example. Or version numbers being displayed wrongly in the `About` window and the windows installer for example! Our tag is the basis for all of this and that information gets parsed from it.
So one will recognize little errors before it's too late, and can re-tag correctly.

The ideal case would be that one only has to set a git tag and everything else is done automatically.
But to have it set a release name like `Cockatrice 2.4.1-beta` for example, our tag would need to be in a different form than it is right now.
Even automatically adding a changelog to the description in the future could be possible. It would require having a changelog file we maintain continuously instead the manual writeup of the past 6 month like Zach is doing it now...

I needed to adjust the cmake getversion file, too.

Possible future improvements:
- Have a good automated name for beta + stable releases on GitHub
- And maybe the changelog thing isn't ideal as is now...

<br>

>I'm sure the regex can be smarter, I didn't use *lookaround* for example.
**Improvements welcome!**

<br>

Link library:
https://docs.travis-ci.com/user/deployment/releases/#Advanced-options
https://www.appveyor.com/docs/deployment/github/
https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A
https://docs.travis-ci.com/user/conditional-builds-stages-jobs
https://regex101.com/
https://www.regular-expressions.info/quickstart.html
see https://github.com/Cockatrice/Cockatrice/pull/2976#issuecomment-354367340

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ deploy:
   file: "build/Cockatrice-*"
   overwrite: true
   draft: false
-  prerelease: false
+  prerelease: true
   on:
     tags: true
     repo: Cockatrice/Cockatrice
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ deploy:
     on:
       tags: true
       repo: Cockatrice/Cockatrice
-      condition: "$BUILDTYPE = Release && $TRAVIS_TAG =~ ([0-9]|[1-9][0-9])(.([0-9]|[1-9][0-9])){2}-beta(|.([2-9]|[1-9][0-9]))$"
+      condition: "$BUILDTYPE = Release && $TRAVIS_TAG =~ ([0-9]|[1-9][0-9])(.([0-9]|[1-9][0-9])){2}-beta(.([2-9]|[1-9][0-9]))?$"
       
 # Deploy configuration for "stable" releases
   - provider: releases
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ deploy:
 # Deploy configuration for "beta" releases
   - provider: releases
     api_key:
-      secure: "mLMF41q7xgOR1sjczsilEy7HQis2PkZCzhfOGbn/8FoOQnmmPOZjrsdhn06ZSl3SFsbfCLuClDYXAbFscQmdgjcGN5AmHV+JYfW650QEuQa/f4/lQFsVRtEqUA1O3FQ0OuRxdpCfJubZBdFVH8SbZ93GLC5zXJbkWQNq+xCX1fU="
+      secure: mLMF41q7xgOR1sjczsilEy7HQis2PkZCzhfOGbn/8FoOQnmmPOZjrsdhn06ZSl3SFsbfCLuClDYXAbFscQmdgjcGN5AmHV+JYfW650QEuQa/f4/lQFsVRtEqUA1O3FQ0OuRxdpCfJubZBdFVH8SbZ93GLC5zXJbkWQNq+xCX1fU=
     skip_cleanup: true
     name: "Cockatrice $TRAVIS_TAG"
     body: "Beta release of Cockatrice"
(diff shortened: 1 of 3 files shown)


tooomm:
I extended the PR and updated its description. 🎉 
Please check again!

.travis.yml
```diff
@@ -35,19 +35,41 @@ install: bash ./.travis/travis-dependencies.sh
...
+    body: "Beta release of Cockatrice"
+    file_glob: true
+    file: "build/Cockatrice-*"
+    overwrite: true
+    draft: false
+    prerelease: true
+    on:
+      tags: true
+      repo: Cockatrice/Cockatrice
+      condition: "$BUILDTYPE = Release && $TRAVIS_TAG =~ ([0-9]|[1-9][0-9])(.([0-9]|[1-9][0-9])){2}-beta(|.([2-9]|[1-9][0-9]))$"
```
ZeldaZach: You say this condition twice. Is that correct?

tooomm: What do you mean @ZeldaZach ?

- travis.yml
    - **beta config:** only on tags & on main repo & release build (not debug!) & **beta regex**
    - **stable: config:** only on tags & on main repo & release build (not debug!) & **stable regex**
- appveyor.yml
    - **beta config:** only on tags & **beta regex**
    - **stable: config:** only on tags & **stable regex**

The basic condition package always checks for general deploy rules, the same we had before.
But the deploy provider within each config file differentiate on the regex.

ctrlaltca:
The change looks fine, but i'm unsure wether there's a proper way to test it if not just merging the change and trying to tag a beta.

skwerlman:
two minor nitpicks about the regexes:
- each `.` should be escaped with a backslash, since `.` matches any character
- in the regex for beta tags, the last group should be `beta(\.([2-9]|[1-9][0-9]))?$`, not `beta(|\.([2-9]|[1-9][0-9]))$`

tooomm:
>each `.` should be escaped with a backslash, since `.` matches any character

@skwerlman for correct regex that is right, but the [travis yaml linter](https://yaml.travis-ci.org/) gave me en error, so I removed the escaping and kept the dot for the proper spacing at least:
`syntax error: (<unknown>): found unknown escape character while parsing a quoted scalar at line x column y`

A live run on Travis won't work too, just tested it:
```
ERROR: An error occured while trying to parse your .travis.yml file.
Please make sure that the file is valid YAML.
http://lint.travis-ci.org can check your .travis.yml.
The log message was: Build config file had a parse error: found unknown escape character while parsing a quoted scalar at line 46 column 18
```
Can I safely use something like `[.]` instead? What do you suggest?

(for AppVeyor with their regex escaping it seems to work - will re-add it there!)

Also 👍 for the better quantifier, haha. Got lost in brackets and alternations I guess. 🙈

skwerlman:
you should be able to double-escape the dot (like `\\.`)

tooomm:
@skwerlman that worked with no errors! Will use the same for `\d`...

Was fun to work with regex for the first time. :)

@ctrlaltca I had it running in my fork with the condition checks tested (but no deploy itself).
We would have to run it on master and observe for the first beta more closely, yes. 👍 
I also had to adjust the sanity check in cmake because the additional `.` in counting beta releases increases the max. allowed token count.
Can you please fix the `GIT_TAG_LABEL` in https://github.com/Cockatrice/Cockatrice/blob/master/cmake/getversion.cmake#L143-L148 that we grab the correct `Project version` etc. with the complete
pre-release-label (not only `beta`, but also `beta.3` and `beta.11`)?

tooomm:
I retested my fork with a tag again after changing to (escaped) `\d`,
Travis seems to really not like any variants of `\d`. Appveyor deploys just fine.

I'll have to revert to `[0-9]` otherwise for travis. Would do it for both for consistency reasons.
The expression will grow in size then, but well...

**Will report back**

tooomm:
`\d` is indeed not available.
Travis uses bash condition checks and POSIX doesn't suppport that. Tested the deployment checks again. 👍 

https://stackoverflow.com/questions/17968020/how-to-match-digits-in-regex
https://en.wikipedia.org/wiki/Regular_expression#Character_classes
https://www.regular-expressions.info/posix.html

---

Only this is left, I hope somebody can help out there:

>Can you please fix the `GIT_TAG_LABEL` in https://github.com/Cockatrice/Cockatrice/blob/master/cmake/getversion.cmake#L143-L148 that we grab the correct `Project version` etc. with the complete
pre-release-label (not only `beta`, but also `beta.3` and `beta.11`)?

Faulty result is now:
```
-- Commit is a release or prerelease, git tag: 2017-12-28-Development-10.40.1-beta.2
-- Project version: 10.40.1-beta
-- Friendly project version: 10.40.1-beta (2017-12-28)
-- Project version filename: Cockatrice-10.40.1-beta
```

ZeldaZach:
I think this is OK? I'm fine merging, pushing a beta build, and seeing what happens

PR #2987, opened 2018-01-02
card filter: alphabetical ordering for keywords

tooomm:
Keywords used to be unordered.
Because of the missing logical structure it was difficult to quickly find the needed filter.

- Display filter keywords in alphabetical order
- ~~Filter logic is ordered in the same way in the code now, too~~

## Screenshots
- before:
untitled1

- after:
untitled2

### DIFF
--- a/cockatrice/src/cardfilter.cpp
+++ b/cockatrice/src/cardfilter.cpp
@@ -20,25 +20,25 @@ const char *CardFilter::attrName(Attr a)
 {
     switch (a) {
         case AttrName:
-            return "name";
+            return "Name";
         case AttrType:
-            return "type";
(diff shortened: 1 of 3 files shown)


Nightfirecat:
Can these be sorted without reordering the cases of `CardFilter::attrName` in cardfilter.cpp? I don't mind them being alphabetically ordered in the source as well, but I'm also generally in favor of sorting with a `.sort()` call instead of doing it by hand&mdash;that way, future changes don't have to rigorously follow the same sorting, but would be automatically sorted.

ZeldaZach:
@nightfirecat I fully agree. This complicated change logs for no reason

PR #2989, opened 2018-01-03
fix tr call for card filter keywords

### LINKED ISSUE #2986: card filter: keywords are not translatable
The keywords such as `cmc`, `type` or `text` in the card filter are not translatable right now.

I guess it's only about the missing `tr()` wrappings here?
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/cardfilter.cpp#L22-L43

Same is true for the logical connectors like `and` and `or not` etc.

tooomm:
## Related Ticket(s)
- Fixes #2986

Translation of keywords in the card filter wasn't possible.

- tr() wrapping right next to the strings
- Use title case capitalization for UI

### DIFF
--- a/cockatrice/src/cardfilter.cpp
+++ b/cockatrice/src/cardfilter.cpp
@@ -20,25 +20,25 @@ const char *CardFilter::attrName(Attr a)
 {
     switch (a) {
         case AttrName:
-            return "name";
+            return tr("Name");
         case AttrType:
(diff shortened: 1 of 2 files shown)


tooomm:
Not as easy as initially thought: https://github.com/Cockatrice/Cockatrice/issues/2986#issuecomment-354913775 :)

Issue #2764, opened 2017-06-12
[RFC] Spoiler Handling


dev-id:
Spoilers are a type of custom set that is popular on Cockatrice servers. During the weeks leading up to a set release, Wizards and their affiliates provide images of cards for players to discuss, speculate, and test with. Many Cockatrice users like to get testing immediately and to facilitate this, custom XML files are created containing the spoiled cards.

In an effort to improve quality and centralize the spoilers, Cockatrice developers have created the [Magic-Spoiler](https://github.com/Cockatrice/Magic-Spoiler) repository. This repository is designed to scrape some of the sites that compile the spoilers as they are realized, as well as manually entering cards leveraging git and GitHub functionality to create a Cockatrice friendly XML file. (Some preliminary discussion in #2622)

This thread is to discuss and improve the way that Cockatrice handles these "unique" custom sets. Much of the custom set handling code is designed for user-created cards that are not frequently updated and do not expire (as spoilers effectively do when the set is released and added to mtgjson.)

Current possible issues with the custom set handling code in regards to spoilers (or overall):

- [x] 1. Using `Add custom sets/cards` creates duplicate set files instead of updating if set/spoiler exists
--> #2784
- [x] 2. Only way to 'clear' spoiler XML at end of season (mtgjson update) is to open folder and delete files via OS
- [x] 3. Not possible to see if spoiler file is newer than current
- [ ] 4. Customsets folder is not configurable
- [x] 5. If spoiler file exists for a set, running oracle and importing the same set from mtgjson will not prompt user to add set (related #2712)
- [ ] 6. If an improved (English, higher resolution or better quality) image for a card is found and updated in the spoiler file and this new file gets used in trice, Cockatrice still maintains the original image until deleted via entire image wipe or manual OS delete
- [ ] 7. After mtgjson updates, images from XML that exist are still used instead of gatherer or image source

tritoch:
Adding some meta data to the XML file will assist in some of these items:

1 & 3:
Should have a created date/time, and overwrite file if set code exists and created date is newer. Cockatrice can then check spoiler file against the same file in customsets for the created date/time.

2, 4, 5, 7:
A `<spoiler="True">` key would allow Cockatrice to check if file/set is a spoiler, and then if it exists in mtgjson, prompt (and recommend) you to delete the spoiler file and any downloaded pics, and then perform that action.

Daenyth:
[...]

Issue #1874, opened 2016-03-17
Add ability to add custom set from url (download it)


dev-id:
Would it be possible to have Cockatrice use a web-based XML file for its custom set?  

Use case: during spoiler season, it's possible to generate a new XML periodically for the newly-spoiled cards, but it has to be downloaded each time.

ctrlaltca:
I get your point, and we could expand the recently added feature "add custom set to cockatrice", that by now permits only to load local files, to also download remote files.
This way you'll have to
 * click on a menu: card database → add custom set
 * paste the XML url in there (only the first time, we can remember the last used url) and press Ok
 * have cockatrice download the file, add it to the custom sets and reload the card db on the fly

Bonus: if a custom set with the same name already exists, cockatrice should ask the user if he wants to overwrite the old file or create another.

radioactivehazards:
This is bad I think as most of those custom xlmls are bugged with errors. It's easier to test with Custom images and "writing cards" into your decks. A function that is Now glitched and NEEDS immediate Fixing!!!!!!!  People who do as you are suggesting and load a new xmls every day basically get viruses and many spelling errored cards and such.

ctrlaltca:
> most of those custom xlmls are bugged with errors

agreed

> A function that is Now glitched and NEEDS immediate Fixing!!!!!!!

This feature currently doesn't exists at all !!1!!eleven!

> load a new xmls every day basically get viruses

I'm not aware of any example of this,

> nd many spelling errored cards and such.

but i understand what you mean and somewhat agree that making people load files from untrusted sources could be troublesome. How to prevent such problems is an hard matter, but we could start a discussion right here.

radioactivehazards:
The Feature that you say does not exist Used to. I could make SOi cards into my decks without xmls very easily. Now its not possible.    Before I could find the card I want to test with and "Save image" to Custom ad then when i use the Save to clipboard/load from clipboard Features I can add in SOi card names. and when the deck is testing i get Soi cards with pictures Voila!. Now I am forced to use blank cards and use the spoiler to "refresh" what they all do.  Fixing the glitch with tokens would stop the need for Xmls every day as people would just do it the Same way that i used to as it's much easier and has less chance of errors

Daenyth:
You should be able to use custom images for cards not in the database in the way you described. If that doesn't work, it's a bug we need to fix. It's probably the same root problem as #1768

radioactivehazards:
[...]

PR #2991, opened 2018-01-04
Automatic Spoiler Season

ZeldaZach:
This PR contains a lot, so lets try and break it down a bit.

The goal of this PR is to allow users to automatically download and manage their spoilers through our spoiler service, Magic-Spoiler. In order to accomplish this, I've added components to the settings dialog.

screenshot 2018-01-08 23 52 23

If the user ticks the "Download Spoilers Automatically" button, the following gears internally will start turning:
+ On Cockatrice launch, check if the file "SpoilerSeasonEnabled" exists on the Magic-Spoiler repo. If it does not, then we know we are not in spoiler season and will continue with normal launch.

+ If the file does exist, then we will continue and now download spoiler.xml from the Magic-Spoiler repo. The system will then check if the downloaded copy matches the current local copy. If there is no local copy or the file hashes do not match, then we will replace the local copy with the downloaded copy and alert the user.
screenshot 2018-01-08 23 58 02

+ If the file hashes do match, then we know the spoilers are up to date and we discard the downloaded version and alert the user.
screenshot 2018-01-08 23 56 18

+ If the user presses the "Update Spoilers" button in the system settings, it will trigger bullet point 1 above with the SpoilerSeasonEnabled check and continue down the line.

+ If any errors occur, they will be properly logged in the debugger. 

___

In addition to the changes above, there is some general cleanup in the code base (as dictated by CLion's strict IDE policy of ensuring top quality code). Some examples include adding `explicit` to functions or `override`s. In addition, there are some bracket realignments and header realignments to meet the new coding policy (which will be updated in the near future).

Continuing on, the first attempt of this PR used Oracle to download the spoilers. Since I find that code to be useful, I have decided to leave it in the PR as an Easter Egg so it can be used in the future by another process (who knows, maybe we'll want to have Oracle do new things, and this is a great example of what it is capable of!)

Finally, I detected a race-time condition that can come about if the system attempts to reload the card database while a reload is currently in progress. I have added a Mutex to make the card database reload thread safe from now on.

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -2,12 +2,14 @@
 #include "pictureloader.h"
 #include "settingscache.h"
 #include "spoilerbackgroundupdater.h"
+#include "main.h"
 
 #include <QCryptographicHash>
 #include <QDebug>
 #include <QDir>
 #include <QFile>
 #include <QMessageBox>
+#include <QtConcurrent>
 
 const int CardDatabase::versionNeeded = 3;
 const char* CardDatabase::TOKENS_SETNAME = "TK";
@@ -688,6 +690,23 @@ void CardDatabase::loadSetsFromXml(QXmlStreamReader &xml)
             newSet->setReleaseDate(releaseDate);
         }
     }
+}
+
+/*
+ * If the database needs to be loaded/reloaded this method can be called
+ * and is thread safe. This solves the bug where if two instances of
+ * the reload were taking place, a race time condition would appear.
+ */
+void CardDatabase::threadSafeReloadCardDatabase()
+{
+    if (reloadDatabaseMutex == nullptr)
+    {
+        reloadDatabaseMutex = new QBasicMutex();
+    }
+
+    reloadDatabaseMutex->lock();
+    QtConcurrent::run(db, &CardDatabase::loadCardDatabases);
+    reloadDatabaseMutex->unlock();
 }
 
 void CardDatabase::loadCardsFromXml(QXmlStreamReader &xml)
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -14,7 +14,7 @@ const char* CardDatabase::TOKENS_SETNAME = "TK";
 
 static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardSet *set)
 {
-    if (! set)
+    if (set == nullptr)
     {
         qDebug() << "&operator<< set is nullptr";
         return xml;
@@ -78,9 +78,9 @@ class SetList::KeyCompareFunctor
     public:
         inline bool operator()(CardSet *a, CardSet *b) const
         {
-            if (!a || !b)
+            if (a == nullptr || b == nullptr)
             {
-                qDebug() << "SetList a or b is null";
+                qDebug() << "SetList::KeyCompareFunctor a or b is null";
                 return false;
             }
 
@@ -158,7 +158,7 @@ void SetList::enableAll()
     {
         CardSet *set = at(i);
 
-        if (!set)
+        if (set == nullptr)
         {
             qDebug() << "enabledAll has null";
             continue;
@@ -198,7 +198,7 @@ void SetList::guessSortKeys()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if (! set)
+        if (set == nullptr)
         {
             qDebug() << "guessSortKeys set is null";
             continue;
@@ -324,7 +324,7 @@ QString CardInfo::getCorrectedName() const
 
 void CardInfo::addToSet(CardSet *set)
 {
-    if (! set)
+    if (set == nullptr)
     {
         qDebug() << "addToSet(nullptr)";
         return;
@@ -387,9 +387,9 @@ const QChar CardInfo::getColorChar() const
 
 static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardInfo *info)
 {
-    if (! info)
+    if (info == nullptr)
     {
-        qDebug() << "operator<< (~376) info is nullptr";
+        qDebug() << "operator<< info is nullptr";
         return xml;
     }
 
@@ -574,7 +574,7 @@ void CardDatabase::addCard(CardInfo *card)
 
 void CardDatabase::removeCard(CardInfo *card)
 {
-    if (! card)
+    if (card == nullptr)
     {
         qDebug() << "removeCard(nullptr)";
         return;
@@ -939,7 +939,7 @@ bool CardDatabase::saveToFile(const QString &fileName, bool tokens)
     xml.writeStartElement("cockatrice_carddatabase");
     xml.writeAttribute("version", QString::number(versionNeeded));
 
-    if (!tokens)
+    if (! tokens)
     {
         xml.writeStartElement("sets");
         QHashIterator<QString, CardSet *> setIterator(sets);
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -526,7 +526,7 @@ static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardInfo *info)
 
 CardDatabase::CardDatabase(QObject *parent) : QObject(parent), loadStatus(NotLoaded)
 {
-    connect(settingsCache, SIGNAL(cardDatabasePathChanged()), this, SLOT(loadCardDatabases()), Qt::QueuedConnection);
+    connect(settingsCache, SIGNAL(cardDatabasePathChanged()), this, SLOT(threadSafeReloadCardDatabase()));
 }
 
 CardDatabase::~CardDatabase()
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -526,7 +526,7 @@ static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardInfo *info)
 
 CardDatabase::CardDatabase(QObject *parent) : QObject(parent), loadStatus(NotLoaded)
 {
-    connect(settingsCache, SIGNAL(cardDatabasePathChanged()), this, SLOT(loadCardDatabases()));
+    connect(settingsCache, SIGNAL(cardDatabasePathChanged()), this, SLOT(loadCardDatabases()), Qt::QueuedConnection);
 }
 
 CardDatabase::~CardDatabase()
@@ -699,14 +699,33 @@ void CardDatabase::loadSetsFromXml(QXmlStreamReader &xml)
  */
 void CardDatabase::threadSafeReloadCardDatabase()
 {
+    QtConcurrent::run(db, &CardDatabase::reloadCardDatabases);
+}
+
+void CardDatabase::reloadCardDatabases()
+{
     if (reloadDatabaseMutex == nullptr)
     {
         reloadDatabaseMutex = new QBasicMutex();
     }
 
-    reloadDatabaseMutex->lock();
-    QtConcurrent::run(db, &CardDatabase::loadCardDatabases);
-    reloadDatabaseMutex->unlock();
+    // Try to obtain the lock.
+    // If it fails, then the database is already
+    // being reloaded and we don't need to reload
+    // again as no results will change.
+    if (reloadDatabaseMutex->tryLock())
+    {
+        qDebug() << "CardDatabase::reloadCardDatabases Lock Acquired";
+
+        db->loadCardDatabases();
+        reloadDatabaseMutex->unlock();
+
+        qDebug() << "CardDatabase::reloadCardDatabases Lock Released";
+    }
+    else
+    {
+        qDebug() << "CardDatabase::reloadCardDatabases Failed to Acquire Lock";
+    }
 }
 
 void CardDatabase::loadCardsFromXml(QXmlStreamReader &xml)
@@ -1003,8 +1022,6 @@ LoadStatus CardDatabase::loadCardDatabase(const QString &path)
 
 LoadStatus CardDatabase::loadCardDatabases()
 {
-    qDebug() << "CardDatabase::loadCardDatabases start";
-
     clear(); // remove old db
 
     loadStatus = loadCardDatabase(settingsCache->getCardDatabasePath()); // load main card database
@@ -1035,11 +1052,10 @@ LoadStatus CardDatabase::loadCardDatabases()
     if (loadStatus == Ok)
     {
         checkUnknownSets(); // update deck editors, etc
-        qDebug() << "CardDatabase::loadCardDatabases success";
     }
     else
     {
-        qDebug() << "CardDatabase::loadCardDatabases failed";
+        qDebug() << "CardDatabase::loadCardDatabases Reload Failed";
         emit cardDatabaseLoadingFailed(); // bring up the settings dialog
     }
 
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -1,6 +1,7 @@
 #include "carddatabase.h"
 #include "pictureloader.h"
 #include "settingscache.h"
+#include "spoilerbackgroundupdater.h"
 
 #include <QCryptographicHash>
 #include <QDebug>
@@ -734,7 +735,9 @@ LoadStatus CardDatabase::loadCardDatabase(const QString &path)
 {
     LoadStatus tempLoadStatus = NotLoaded;
     if (!path.isEmpty())
+    {
         tempLoadStatus = loadFromFile(path);
+    }
 
     qDebug() << "[CardDatabase] loadCardDatabase(): Path =" << path << "Status =" << tempLoadStatus << "Cards =" << cards.size() << "Sets=" << sets.size();
 
@@ -744,12 +747,13 @@ LoadStatus CardDatabase::loadCardDatabase(const QString &path)
 LoadStatus CardDatabase::loadCardDatabases()
 {
     qDebug() << "CardDatabase::loadCardDatabases start";
-    // clean old db
-    clear();
-    // load main card database
-    loadStatus = loadCardDatabase(settingsCache->getCardDatabasePath());
-    // laod tokens database
-    loadCardDatabase(settingsCache->getTokenDatabasePath());
+
+    clear(); // remove old db
+
+    loadStatus = loadCardDatabase(settingsCache->getCardDatabasePath()); // load main card database
+    loadCardDatabase(settingsCache->getTokenDatabasePath()); // load tokens database
+    loadCardDatabase(settingsCache->getSpoilerCardDatabasePath()); // load spoilers database
+
     // load custom card databases
     QDir dir(settingsCache->getCustomCardDatabasePath());
     foreach(QString fileName, dir.entryList(QStringList("*.xml"), QDir::Files | QDir::Readable, QDir::Name | QDir::IgnoreCase))
@@ -763,25 +767,25 @@ LoadStatus CardDatabase::loadCardDatabases()
     SetList allSets;
     QHashIterator<QString, CardSet *> setsIterator(sets);
     while (setsIterator.hasNext())
+    {
         allSets.append(setsIterator.next().value());
+    }
     allSets.sortByKey();
 
     // resolve the reverse-related tags
     refreshCachedReverseRelatedCards();
 
-    if(loadStatus == Ok)
+    if (loadStatus == Ok)
     {
-        // check for unknown sets
-        checkUnknownSets();
-        // update deck editors, etc
+        checkUnknownSets(); // update deck editors, etc
         qDebug() << "CardDatabase::loadCardDatabases success";
-    } else {
-        // bring up thr settings dialog
+    }
+    else
+    {
         qDebug() << "CardDatabase::loadCardDatabases failed";
-        emit cardDatabaseLoadingFailed();
+        emit cardDatabaseLoadingFailed(); // bring up the settings dialog
     }
 
-    // return the loadstatus of the main card database.
     return loadStatus;
 }
 
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -6,9 +6,7 @@
 #include <QCryptographicHash>
 #include <QDebug>
 #include <QDir>
-#include <QDirIterator>
 #include <QFile>
-#include <QTextStream>
 #include <QMessageBox>
 
 const int CardDatabase::versionNeeded = 3;
@@ -16,6 +14,12 @@ const char* CardDatabase::TOKENS_SETNAME = "TK";
 
 static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardSet *set)
 {
+    if (! set)
+    {
+        qDebug() << "&operator<< set is nullptr";
+        return xml;
+    }
+
     xml.writeStartElement("set");
     xml.writeTextElement("name", set->getShortName());
     xml.writeTextElement("longname", set->getLongName());
@@ -69,12 +73,19 @@ void CardSet::setIsKnown(bool _isknown)
     settingsCache->cardDatabase().setIsKnown(shortName,_isknown);
 }
 
-class SetList::KeyCompareFunctor {
-public:
-    inline bool operator()(CardSet *a, CardSet *b) const
-    {
-        return a->getSortKey() < b->getSortKey();
-    }
+class SetList::KeyCompareFunctor
+{
+    public:
+        inline bool operator()(CardSet *a, CardSet *b) const
+        {
+            if (!a || !b)
+            {
+                qDebug() << "SetList a or b is null";
+                return false;
+            }
+
+            return a->getSortKey() < b->getSortKey();
+        }
 };
 
 void SetList::sortByKey()
@@ -84,24 +95,28 @@ void SetList::sortByKey()
 
 int SetList::getEnabledSetsNum()
 {
-    int num=0;
+    int num = 0;
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(set->getEnabled())
+        if (set && set->getEnabled())
+        {
             ++num;
+        }
     }
     return num;
 }
 
 int SetList::getUnknownSetsNum()
 {
-    int num=0;
+    int num = 0;
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(!set->getIsKnown() && !set->getIsKnownIgnored())
+        if (set && !set->getIsKnown() && !set->getIsKnownIgnored())
+        {
             ++num;
+        }
     }
     return num;
 }
@@ -112,8 +127,10 @@ QStringList SetList::getUnknownSetsNames()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(!set->getIsKnown() && !set->getIsKnownIgnored())
+        if (set && !set->getIsKnown() && !set->getIsKnownIgnored())
+        {
             sets << set->getShortName();
+        }
     }
     return sets;
 }
@@ -123,12 +140,12 @@ void SetList::enableAllUnknown()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(!set->getIsKnown() && !set->getIsKnownIgnored())
+        if (set && !set->getIsKnown() && !set->getIsKnownIgnored())
         {
             set->setIsKnown(true);
             set->setEnabled(true);
         }
-        else if (set->getIsKnownIgnored() && !set->getEnabled())
+        else if (set && set->getIsKnownIgnored() && !set->getEnabled())
         {
             set->setEnabled(true);
         }
@@ -140,8 +157,18 @@ void SetList::enableAll()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(!set->getIsKnownIgnored())
+
+        if (!set)
+        {
+            qDebug() << "enabledAll has null";
+            continue;
+        }
+
+        if (!set->getIsKnownIgnored())
+        {
             set->setIsKnown(true);
+        }
+
         set->setEnabled(true);
     }
 }
@@ -151,12 +178,12 @@ void SetList::markAllAsKnown()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
-        if(!set->getIsKnown() && !set->getIsKnownIgnored())
+        if(set && !set->getIsKnown() && !set->getIsKnownIgnored())
         {
             set->setIsKnown(true);
             set->setEnabled(false);
         }
-        else if (set->getIsKnownIgnored() && !set->getEnabled())
+        else if (set && set->getIsKnownIgnored() && !set->getEnabled())
         {
             set->setEnabled(true);
         }
@@ -171,11 +198,21 @@ void SetList::guessSortKeys()
     for (int i = 0; i < size(); ++i)
     {
         CardSet *set = at(i);
+        if (! set)
+        {
+            qDebug() << "guessSortKeys set is null";
+            continue;
+        }
+
         QDate date = set->getReleaseDate();
-        if(date.isNull())
-            set->setSortKey(aHundredYears);
+        if (date.isNull())
+        {
+            set->setSortKey(static_cast<unsigned int>(aHundredYears));
+        }
         else
-            set->setSortKey(date.daysTo(distantFuture));
+        {
+            set->setSortKey(static_cast<unsigned int>(date.daysTo(distantFuture)));
+        }
     }
 }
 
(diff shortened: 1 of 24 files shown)


Rendroc:
Is this active yet? What is the time interval?

ZeldaZach:
@Rendroc This is a pull request for a development feature. Until it's merged into the master branch it should be considered volatile.

The timer interval is configurable as seen in https://github.com/Cockatrice/Cockatrice/pull/2991/files#diff-59668b0ce2925fb7c66498412bdadfc1R174

dev-id:
Tested and working on Windows 7 (portable install)

Looks good. Finally a way to update spoilers automatically. Obsoletes a lot of #2764 . Somewhat handles #1874

tooomm:
As said in gitter, Cockatrice crashes immediately after/during updating on my Windows 10.
Again: Experienced it after the spoiler set was the only enabled set, then tried again with spoiler (RIX) plus Ixalan with same result, it crashes shortly after the app starts.
There is no way to stop the update process to eventually prevent a crash and disable the setting for example. One can't open `Edit sets...` and disconnecting from the internet also doesn't help.

I need to retest with a clean install and fresh settings now.

<details><summary>Crash Log</summary>

```
Version=1
EventType=APPCRASH
EventTime=131598206188288723
ReportType=2
Consent=1
UploadTime=131598206233483354
ReportStatus=100
ReportIdentifier=5b59d192-9826-40ff-8ba8-50148c417f07
IntegratorReportIdentifier=e94ba42d-8dfc-4e63-921a-55927d4923f7
Wow64Host=34404
NsAppName=cockatrice.exe
AppSessionGuid=000030a8-0001-000b-b15b-a20adf87d301
TargetAppId=W:0006849d33b216b6d95d4a2059e13c208d310000ffff!00002eaf785429b5fbd29b955fd9c6b27503cd46bed6!cockatrice.exe
TargetAppVer=2018//01//06:06:54:03!0!cockatrice.exe
BootId=4294967295
TargetAsId=88245
UserImpactVector=3146544
Response.type=4
Sig[0].Name=Application Name
Sig[0].Value=cockatrice.exe
Sig[1].Name=Application Version
Sig[1].Value=0.0.0.0
Sig[2].Name=Application Timestamp
Sig[2].Value=5a50728b
Sig[3].Name=Fault Module Name
Sig[3].Value=StackHash_7104
Sig[4].Name=Fault Module Version
Sig[4].Value=10.0.16299.64
Sig[5].Name=Fault Module Timestamp
Sig[5].Value=493793ea
Sig[6].Name=Exception Code
Sig[6].Value=c0000374
Sig[7].Name=Exception Offset
Sig[7].Value=PCH_20_FROM_ntdll+0x00000000000A0994
DynamicSig[1].Name=OS Version
DynamicSig[1].Value=10.0.16299.2.0.0.256.48
DynamicSig[2].Name=Locale ID
DynamicSig[2].Value=1031
DynamicSig[22].Name=Additional Information 1
DynamicSig[22].Value=7104
DynamicSig[23].Name=Additional Information 2
DynamicSig[23].Value=710446759dc3b2c9c63dad0617ab82ee
DynamicSig[24].Name=Additional Information 3
DynamicSig[24].Value=5349
DynamicSig[25].Name=Additional Information 4
DynamicSig[25].Value=5349ceb3184632cac27c03c21438cce2
UI[2]=C:\Program Files (x86)\Cockatrice\cockatrice.exe
UI[3]=cockatrice.exe has stopped working
UI[4]=Windows can check online for a solution to the problem.
UI[5]=Check online for a solution and close the program
UI[6]=Check online for a solution later and close the program
UI[7]=Close the program
LoadedModule[0]=C:\Program Files (x86)\Cockatrice\cockatrice.exe
LoadedModule[1]=C:\WINDOWS\SYSTEM32\ntdll.dll
LoadedModule[2]=C:\WINDOWS\System32\KERNEL32.DLL
LoadedModule[3]=C:\WINDOWS\System32\KERNELBASE.dll
LoadedModule[4]=C:\WINDOWS\System32\ucrtbase.dll
LoadedModule[5]=C:\WINDOWS\System32\SHELL32.dll
LoadedModule[6]=C:\WINDOWS\System32\msvcrt.dll
LoadedModule[7]=C:\WINDOWS\System32\cfgmgr32.dll
LoadedModule[8]=C:\WINDOWS\System32\shcore.dll
LoadedModule[9]=C:\WINDOWS\System32\RPCRT4.dll
LoadedModule[10]=C:\WINDOWS\System32\combase.dll
LoadedModule[11]=C:\WINDOWS\System32\bcryptPrimitives.dll
LoadedModule[12]=C:\WINDOWS\System32\windows.storage.dll
LoadedModule[13]=C:\WINDOWS\System32\advapi32.dll
LoadedModule[14]=C:\WINDOWS\System32\sechost.dll
LoadedModule[15]=C:\WINDOWS\System32\shlwapi.dll
LoadedModule[16]=C:\WINDOWS\System32\GDI32.dll
LoadedModule[17]=C:\WINDOWS\System32\gdi32full.dll
LoadedModule[18]=C:\WINDOWS\System32\msvcp_win.dll
LoadedModule[19]=C:\WINDOWS\System32\USER32.dll
LoadedModule[20]=C:\WINDOWS\System32\win32u.dll
LoadedModule[21]=C:\WINDOWS\System32\kernel.appcore.dll
LoadedModule[22]=C:\WINDOWS\System32\powrprof.dll
LoadedModule[23]=C:\WINDOWS\System32\profapi.dll
LoadedModule[24]=C:\Program Files (x86)\Cockatrice\Qt5Multimedia.dll
LoadedModule[25]=C:\Program Files (x86)\Cockatrice\Qt5PrintSupport.dll
LoadedModule[26]=C:\WINDOWS\System32\COMDLG32.dll
LoadedModule[27]=C:\Program Files (x86)\Cockatrice\Qt5Network.dll
LoadedModule[28]=C:\WINDOWS\System32\WS2_32.dll
LoadedModule[29]=C:\WINDOWS\System32\CRYPT32.dll
LoadedModule[30]=C:\Program Files (x86)\Cockatrice\Qt5Widgets.dll
LoadedModule[31]=C:\WINDOWS\System32\MSASN1.dll
LoadedModule[32]=C:\Program Files (x86)\Cockatrice\Qt5Gui.dll
LoadedModule[33]=C:\Program Files (x86)\Cockatrice\Qt5Core.dll
LoadedModule[34]=C:\WINDOWS\System32\ole32.dll
LoadedModule[35]=C:\WINDOWS\SYSTEM32\MSVCP140.dll
LoadedModule[36]=C:\WINDOWS\SYSTEM32\VCRUNTIME140.dll
LoadedModule[37]=C:\WINDOWS\WinSxS\amd64_microsoft.windows.common-controls_6595b64144ccf1df_5.82.16299.125_none_88782a244abc4c60\COMCTL32.dll
LoadedModule[38]=C:\WINDOWS\SYSTEM32\WINSPOOL.DRV
LoadedModule[39]=C:\WINDOWS\SYSTEM32\DNSAPI.dll
LoadedModule[40]=C:\WINDOWS\SYSTEM32\IPHLPAPI.DLL
LoadedModule[41]=C:\WINDOWS\System32\NSI.dll
LoadedModule[42]=C:\WINDOWS\SYSTEM32\UxTheme.dll
LoadedModule[43]=C:\WINDOWS\SYSTEM32\dwmapi.dll
LoadedModule[44]=C:\WINDOWS\SYSTEM32\MPR.dll
LoadedModule[45]=C:\WINDOWS\SYSTEM32\VERSION.dll
LoadedModule[46]=C:\WINDOWS\SYSTEM32\WINMM.dll
LoadedModule[47]=C:\WINDOWS\SYSTEM32\bcrypt.dll
LoadedModule[48]=C:\WINDOWS\SYSTEM32\WINMMBASE.dll
LoadedModule[49]=C:\WINDOWS\System32\IMM32.DLL
LoadedModule[50]=C:\WINDOWS\SYSTEM32\CRYPTBASE.DLL
LoadedModule[51]=C:\Program Files (x86)\Cockatrice\Plugins\platforms\qwindows.dll
LoadedModule[52]=C:\WINDOWS\System32\OLEAUT32.dll
LoadedModule[53]=C:\Program Files\Common Files\microsoft shared\ink\tiptsf.dll
LoadedModule[54]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qgif.dll
LoadedModule[55]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qicns.dll
LoadedModule[56]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qico.dll
LoadedModule[57]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qjpeg.dll
LoadedModule[58]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qsvg.dll
LoadedModule[59]=C:\Program Files (x86)\Cockatrice\Qt5Svg.dll
LoadedModule[60]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qtga.dll
LoadedModule[61]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qtiff.dll
LoadedModule[62]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qwbmp.dll
LoadedModule[63]=C:\Program Files (x86)\Cockatrice\plugins\imageformats\qwebp.dll
LoadedModule[64]=C:\WINDOWS\system32\dwrite.dll
LoadedModule[65]=C:\WINDOWS\System32\MSCTF.dll
LoadedModule[66]=C:\WINDOWS\System32\clbcatq.dll
LoadedModule[67]=C:\WINDOWS\system32\dataexchange.dll
LoadedModule[68]=C:\WINDOWS\system32\d3d11.dll
LoadedModule[69]=C:\WINDOWS\system32\dcomp.dll
LoadedModule[70]=C:\WINDOWS\system32\dxgi.dll
LoadedModule[71]=C:\WINDOWS\system32\twinapi.appcore.dll
LoadedModule[72]=C:\WINDOWS\system32\RMCLIENT.dll
LoadedModule[73]=C:\WINDOWS\SYSTEM32\dhcpcsvc.DLL
LoadedModule[74]=C:\WINDOWS\SYSTEM32\dhcpcsvc6.DLL
LoadedModule[75]=C:\Program Files (x86)\Cockatrice\ssleay32.dll
LoadedModule[76]=C:\Program Files (x86)\Cockatrice\LIBEAY32.dll
LoadedModule[77]=C:\WINDOWS\SYSTEM32\MSVCR120.dll
LoadedModule[78]=C:\WINDOWS\SYSTEM32\CRYPTSP.dll
LoadedModule[79]=C:\WINDOWS\system32\rsaenh.dll
LoadedModule[80]=C:\WINDOWS\system32\winhttp.dll
LoadedModule[81]=C:\WINDOWS\SYSTEM32\SspiCli.dll
LoadedModule[82]=C:\WINDOWS\system32\napinsp.dll
LoadedModule[83]=C:\WINDOWS\system32\pnrpnsp.dll
LoadedModule[84]=C:\WINDOWS\system32\NLAapi.dll
LoadedModule[85]=C:\WINDOWS\System32\mswsock.dll
LoadedModule[86]=C:\WINDOWS\System32\winrnr.dll
LoadedModule[87]=C:\Program Files\Bonjour\mdnsNSP.dll
LoadedModule[88]=C:\Windows\System32\rasadhlp.dll
LoadedModule[89]=C:\WINDOWS\System32\TextInputFramework.dll
LoadedModule[90]=C:\WINDOWS\System32\CoreUIComponents.dll
LoadedModule[91]=C:\WINDOWS\System32\CoreMessaging.dll
LoadedModule[92]=C:\WINDOWS\SYSTEM32\ntmarta.dll
LoadedModule[93]=C:\WINDOWS\SYSTEM32\wintypes.dll
LoadedModule[94]=C:\WINDOWS\System32\fwpuclnt.dll
LoadedModule[95]=C:\Windows\System32\uiautomationcore.dll
LoadedModule[96]=C:\WINDOWS\SYSTEM32\sxs.dll
LoadedModule[97]=C:\Windows\System32\OLEACC.dll
LoadedModule[98]=C:\WINDOWS\WinSxS\amd64_microsoft.windows.common-controls_6595b64144ccf1df_6.0.16299.125_none_15cbcf8893620c09\comctl32.dll
LoadedModule[99]=C:\WINDOWS\SYSTEM32\WindowsCodecs.dll
LoadedModule[100]=C:\Windows\System32\thumbcache.dll
OsInfo[0].Key=vermaj
OsInfo[0].Value=10
OsInfo[1].Key=vermin
OsInfo[1].Value=0
OsInfo[2].Key=verbld
OsInfo[2].Value=16299
OsInfo[3].Key=ubr
OsInfo[3].Value=125
OsInfo[4].Key=versp
OsInfo[4].Value=0
OsInfo[5].Key=arch
OsInfo[5].Value=9
OsInfo[6].Key=lcid
OsInfo[6].Value=1031
OsInfo[7].Key=geoid
OsInfo[7].Value=94
OsInfo[8].Key=sku
OsInfo[8].Value=48
OsInfo[9].Key=domain
OsInfo[9].Value=0
OsInfo[10].Key=prodsuite
OsInfo[10].Value=256
OsInfo[11].Key=ntprodtype
OsInfo[11].Value=1
OsInfo[12].Key=platid
OsInfo[12].Value=10
OsInfo[13].Key=sr
OsInfo[13].Value=0
OsInfo[14].Key=tmsi
OsInfo[14].Value=95461
OsInfo[15].Key=osinsty
OsInfo[15].Value=3
OsInfo[16].Key=iever
OsInfo[16].Value=11.125.16299.0-11.0.49
OsInfo[17].Key=portos
OsInfo[17].Value=0
OsInfo[18].Key=ram
OsInfo[18].Value=16201
OsInfo[19].Key=svolsz
OsInfo[19].Value=99
OsInfo[20].Key=wimbt
OsInfo[20].Value=0
OsInfo[21].Key=blddt
OsInfo[21].Value=170928
OsInfo[22].Key=bldtm
OsInfo[22].Value=1534
OsInfo[23].Key=bldbrch
OsInfo[23].Value=rs3_release
OsInfo[24].Key=bldchk
OsInfo[24].Value=0
OsInfo[25].Key=wpvermaj
OsInfo[25].Value=0
OsInfo[26].Key=wpvermin
OsInfo[26].Value=0
OsInfo[27].Key=wpbuildmaj
OsInfo[27].Value=0
OsInfo[28].Key=wpbuildmin
OsInfo[28].Value=0
OsInfo[29].Key=osver
OsInfo[29].Value=10.0.16299.125.amd64fre.rs3_release.170928-1534
OsInfo[30].Key=buildflightid
OsInfo[30].Value=58DDBB00-947E-420A-B296-A2451AE6AC0B.2
OsInfo[31].Key=edition
OsInfo[31].Value=Professional
OsInfo[32].Key=ring
OsInfo[32].Value=Retail
OsInfo[33].Key=containerid
OsInfo[34].Key=containertype
OsInfo[35].Key=edu
OsInfo[35].Value=0
File[0].CabName=WERInternalMetadata.xml
File[0].Path=WER3C42.tmp.WERInternalMetadata.xml
File[0].Flags=327683
File[0].Type=5
File[0].Original.Path=\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER3C42.tmp.WERInternalMetadata.xml
FriendlyEventName=Stopped working
ConsentKey=APPCRASH
AppName=cockatrice.exe
AppPath=C:\Program Files (x86)\Cockatrice\cockatrice.exe
NsPartner=windows
NsGroup=windows8
ApplicationIdentity=4161E7B722011D8F6A569491735EE9F8
MetadataHash=385328636
```

</details>

oracle/src/main.cpp
```diff
@@ -3,6 +3,7 @@
 #include <QIcon>
 #include <QTranslator>
 #include <QLibraryInfo>
+#include <QCommandLineParser>
```
ZeldaZach: I'm leaving the changes in Oracle as an Easter egg that can be used in the future.

tooomm:
I really like that we have something working like this now 👍 

**Here is my list of thoughts:**
- Cockatrice should handle the key file `SpoilerSeasonStatus` beeing non-existent/deleted or simply not available due to whatever reason.

- What happens when the same set/cards appears in spoiler.xml and mtgjson (->cards.xml)? This could happen during a short overlap or if there are other circumstances and should be covered!

- I like this to be a bit smarter.
It always takes the newest file - even with no changes - and replaces the old. It'll also notifies you falsely about a successful update and new available spoilers.
For example: Right now everything for RIX is spoiled since a few days, just very few very tiny updates will happen. But Cockatrice doesn't know, so each and every launch (can be several times a day too!), and for like 10 days in a row the spoiler updater will do it's work, and show a useless message to the user. Actually the message is wrong in the sense that it says "Spoilers updated!" which is wrong most of the time.
The date could give you a hint, but it's more likely to be just confusing. Anyway, I only want notifications if they tell me something new/useful. ;)<br>
**How it could work imo**
During season, on launch:
    - Check timestamp
(from the fresh file against a value in settings, because a user could mess with the local file)
       - New
          *as is now*
       - Same timestamp, no update available
          Tell the user that there are no new spoiler updates available (notification)
          (It could also be no notification, but then we need to educate the user in the "settings", otherwise they could think something is not working. Also, if there is no notification, users might think the `Update Spoilers` button in settings isn't working, or that needs to always trigger something)

- The notification could be a bit more helpful and explaining to the user.
Not only is the timestamp in its raw form a bit cryptic, but there is room for interpretation and misunderstanding. Take the user by their hand and guide them. :)

If you add a hint and explain the data a bit it's more helpful.
Discarding some less useful information like day, year and seconds help too:
(overall size is actually more or less the same)
```
Spoilers have been updated!
Last change: Jan 06, 21:53 (UTC)
```
Original message:
```
Spoilers have been updated!
Sat, Jan 06 2018, 21:53:13 (UTC)
```
automatic spoiler season notification

- Are the spoiler 100% ready for this and can we guarantee data reliably? @dev-id 
Maybe we should get scryfall going as default first: https://github.com/Cockatrice/Magic-Spoiler/issues/170 (main parts)
There quality and speed is nice, also updating is really incredible. Within a very short time (can be 30min), if you use the on-page reporting directly, they fix typos etc.! Just experienced that yesterday again.

<br>

**Design:** (I'm sorry, but we never spoke about it before)
- I don't like the extra file this is dependent on and that needs to be maintained additionally, maybe we can automate that somehow @ spoiler project, though

- Did you think about using "file existence" rather than "file content"?
File name could be `SpoilerSeasonEnabled` or `SpoilerActive` instead some specific content of a `SpoilerSeasonStatus` file. That seems more straight forward and simpler at first. Now you have to download the season file all the time, maybe just checking a file location for its existence is better and faster. At least it would require less data transfer. :)
A file content might be easier to update manually, but we are definitely aiming at an automation here. And then file creation/deletion should be simpler than changing content of a existing file. We didn't investigated into that yet for sure.

- Why don't we use github releases as a source instead the workaround with hosting output files for downloading in a branch? That's really not ideal.
Releases would allow to check for a change date before downloading the file via [API](https://developer.github.com/v3/repos/releases/) too. So you could check all year for changes with no effort and we wouldn't need the additional `SpoilerSeasonStatus` content to check against at all. That sounds more straight forward to me. Also easier to maintain and no shenanigans for automation needed.
Actually, we could do a date check with your design as well. Checking against the timestamp in spoiler.xml all the time. During season it is downloaded anyway, and in off-season the difference in size for an spoiler.xml with no cards and the `SpoilerSeasonStatus` file can't be that big. But using proper releases as a download source and the API should be a more mature, future proof solution with several additions and more possibilities. We have good experiences with it for app updates already!

- What about our former spoiler handling?
We even included special code for spoilers in custom sets: https://github.com/Cockatrice/Cockatrice/pull/2784
That is no longer needed and can be removed I guess?

I retested with the latest commit btw and the reload problem is now also covered.

cockatrice/src/dlg_settings.cpp
```diff
@@ -486,25 +487,90 @@ void UserInterfaceSettingsPage::retranslateUi()
...
-    
-    generalGroupBox = new QGroupBox;
-    generalGroupBox->setLayout(generalGrid);
-    
-    QVBoxLayout *mainLayout = new QVBoxLayout;
-    mainLayout->addWidget(generalGroupBox);
-    
-    setLayout(mainLayout);
+    auto *lpGeneralGrid = new QGridLayout;
+    auto *lpSpoilerGrid = new QGridLayout;
```
tooomm: All `GeneralGrid` could be removed maybe?
It has been empty for a long time now, and most likely will for some more... I don't know of any feature in the planning that could end up here!

And it looks not really nice 😄 
settings

ZeldaZach:
@tooomm 

I will respond to your bullets with my own bullets.

+ Implemented. It will now check for file `SpoilerSeasonEnabled` on the Magic-Spoiler repo. If this file is found, spoiler season code will execute, otherwise it will not. The content of the file does not matter.

+ If spoiler season is disabled, the spoiler.xml file will be removed from the user system. Users are not to modify this file (as it says so in the file header) so it is our right to delete it. This will help avoid the conflict with Oracle when it has the spoilers on MTGJSON.

+ If spoiler season is enabled, the downloaded spoiler.xml file will have headers compared with the local version. If they match, then a popup will say "Spoilers up to date". If the user hits the update button, they will get the same response (provided they didn't delete their local copy before pressing the button).

+ Notification changed to say "Last changed" with the date stamp.

Daenyth:
The description in the main pr comment sounds like an excellent workflow. I might offer a tip that on error we could pop up a notification that instructs the user to check the debug log, but that's minor - overall the workflow sounds good now.

cockatrice/src/carddatabase.cpp
```diff
@@ -1,20 +1,25 @@
...
 #include <QFile>
-#include <QTextStream>
 #include <QMessageBox>
 
 const int CardDatabase::versionNeeded = 3;
 const char* CardDatabase::TOKENS_SETNAME = "TK";
 
 static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardSet *set)
 {
+    if (! set)
```
Daenyth: Can you explicitly check `nullptr` here rather than the implicit style please? It helps with clarity and intent

cockatrice/src/carddatabase.cpp
```diff
@@ -318,107 +378,151 @@ const QChar CardInfo::getColorChar() const
...
             return QChar('m');
-            break;
     }
 }
 
 static QXmlStreamWriter &operator<<(QXmlStreamWriter &xml, const CardInfo *info)
 {
+    if (! info)
+    {
+        qDebug() << "operator<< (~376) info is nullptr";
```
Daenyth: What is 376?

cockatrice/src/carddatabase.cpp
```diff
@@ -318,107 +378,151 @@ const QChar CardInfo::getColorChar() const
...
     const QStringList &colors = info->getColors();
     for (int i = 0; i < colors.size(); i++)
+    {
         xml.writeTextElement("color", colors[i]);
+    }
 
     const QList<CardRelation *> related = info->getRelatedCards();
-    for (int i = 0; i < related.size(); i++) {
-
+    for (auto i : related)
```
Daenyth: 👍

cockatrice/src/dlg_connect.cpp
```diff
@@ -124,12 +123,12 @@ DlgConnect::DlgConnect(QWidget *parent)
     QGroupBox *btnGroupBox = new QGroupBox(tr(""));
     btnGroupBox->setLayout(buttons);
 
-    QGridLayout *grid = new QGridLayout;
+    auto *grid = new QGridLayout;
```
Daenyth: I don't really see the point in auto-fying these but I guess it doesn't hurt

cockatrice/src/dlg_settings.cpp
Daenyth: I don't see harm in leaving it - let's not increase the amount of stuff being changed since this is already large

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,210 @@
...
+void SpoilerBackgroundUpdater::actDownloadFinishedSpoilersFile()
+{
+    // Check for server reply
+    auto *reply = dynamic_cast<QNetworkReply *>(sender());
+    QNetworkReply::NetworkError errorCode = reply->error();
+
+    if (errorCode == QNetworkReply::NoError)
+    {
+        spoilerData = reply->readAll();
+        reply->deleteLater();
```
Daenyth: Should this be done after `saveDownloadedFile`? If `readAll` doesn't transfer pointer ownership then I could see the save following this being a null issue, maybe

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,210 @@
...
+        // The user should run Oracle to get the latest card information
+        QString fileName = settingsCache->getSpoilerCardDatabasePath();
+        QFileInfo fi(fileName);
+        QDir fileDir(fi.path());
+        QFile file(fileName);
+
+        // Delete the spoiler.xml file as we're not in spoiler season
+        if (file.exists() && file.remove())
+        {
+            qDebug() << "Spoiler Season Offline, Deleting spoiler.xml";
```
Daenyth: Can you also pop a desktop notification telling the user the spoiler file is being removed and they need to run oracle? Otherwise checking for updates will remove cards from their db and leave them confused

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /*
+         * ALERT: Ensure two reloads of the card database do not happen
+         * at the same time or a racetime condition can/will happen!
```
Daenyth: We should probably have a mutex around db reloading to avoid that, but maybe in another pr unless you feel up to adding that. Might need refactoring

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,210 @@
...
+    QFileInfo fi(fileName);
+    QDir fileDir(fi.path());
+
+    if (!fileDir.exists() && !fileDir.mkpath(fileDir.absolutePath()))
+    {
+        return false;
+    }
+
+    // Check if the data matches. If it does, then spoilers are up to date.
+    if (getHash(fileName) == getHash(data))
```
Daenyth: Are you hashing the file name or the file contents? Maybe include the individual hash values in a qDebug?

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,210 @@
...
+    if (!file.open(QIODevice::WriteOnly))
+    {
+        qDebug() << "Spoiler Service Error: File open (w) failed for" << fileName;
+        return false;
+    }
+
+    if (file.write(data) == -1)
+    {
+        qDebug() << "Spoiler Service Error: File write (w) failed for" << fileName;
+        return false;
```
Daenyth: The file handle won't be closed in this branch I think?

cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+QByteArray SpoilerBackgroundUpdater::getHash(const QString fileName)
+{
+    QFile file(fileName);
+
+    if (file.open(QFile::ReadOnly))
+    {
+        // Only read the first 512 bytes (enough to get the "created" tag)
+        const QByteArray bytes = file.read(512);
```
Daenyth: Might be worth qDebugging this value

cockatrice/src/carddatabase.cpp
Daenyth: And further down in the file also

cockatrice/src/spoilerbackgroundupdater.cpp
Daenyth: nvm I see it's overloaded. That works

oracle/src/main.cpp
tooomm: >the first attempt of this PR used Oracle to download the spoilers. Since I find that code to be useful, I have decided to leave it in the PR as an Easter Egg so it can be used in the future by another process (who knows, maybe we'll want to have Oracle do new things, and this is a great example of what it is capable of!)

If I understand it correctly you added unused code, shouldn't you comment that accordingly in the code at least?
The reference in this pr will be forgotten and the next coder will wonder. :)

cockatrice/src/spoilerbackgroundupdater.cpp
tooomm: An empty spoiler.xml file looks like this: https://github.com/Cockatrice/Magic-Spoiler/blob/77ec169ec5ec36352d02db8ab1137a780ba7133e/spoiler.xml

It has 272 bytes, so only checking first 256bytes instead 512 is sufficient?
Adding a do-not-edit hint as requested by you, won't change too much since it could simply be placed behind the timestamp.

oracle/src/main.cpp
ZeldaZach: There's nothing to comment really. It's fairly obvious this code isn't used (as there are no calls to it).

cockatrice/src/carddatabase.cpp
ZeldaZach: Yeah I'll fix that up

cockatrice/src/carddatabase.cpp
ZeldaZach: Was referring to the line number, i'll be more explicit

cockatrice/src/dlg_connect.cpp
ZeldaZach: CLion says auto is better and more "modern" so I'm listening to it and fixing all the "warnings" i come across

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: Makes sense, yeah I can do that

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: Agreed. I'm going to hold off on that so we can get this PR in and i'll start working on it later.

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: Never has been, but I can make it close

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: @tooomm I read the files and put it through systems to get byte counts. Most services tell me i'll need ~400 bytes to be certain to capture the date. As such, I put a power of 2 in place. 512 bytes isn't that much :smile:

cockatrice/src/spoilerbackgroundupdater.cpp
ZeldaZach: I don't think it's necessary, but I'll appease you :)

tooomm:
Finished testing the second last iteration (c70dfc3)

Further thoughts:
- New notification body is to long on windows and wraps around the line end:
updates available msg
The "nothing available" one is good:
no updates available msg
Can't you tweak the timestamp as suggested earlier?
Maybe you want to check out [QDateTime](http://doc.qt.io/qt-5/qdatetime.html) and a simple *[fromString](http://doc.qt.io/qt-5/qdatetime.html#fromString)* --> *[toString](http://doc.qt.io/qt-5/qdatetime.html#toString)* (or some other smarter way) will do the trick to convert from
`ddd, MMM dd yyyy, hh:mm:ss` to simpler+shorter `MMM d, hh:mm`
Also, the notification is only displayed for 3s which is hardly enough to understand all numbers with the linebreak etc.

- In settings, having the last run and/or last changes date/time could be helpful

- If you disable spoiler in menu, the spoiler database file should be deleted as well

- Good thing is spoiler.xml seems to not get deleted if there is simply no connection. 👍 
But it still shows `Spoiler Service Online` in *Debug Log* which is false information. We just don't know because trice can't check when there is a network problem.
Simply assuming it is all up and running and telling the user so in logs isn't ideal.
Also, once there is a network issue or no internet connection, the notification will still pop up and tell the user that `Spoilers are up to date` which is misleading and probably wrong.

- Run into another issue/false error:
If you go to settings and hit `Update Spoiler` and close the settings shortly after after it still seems to check+inform you, but you'll see an error about database load and if you want to change its location (Hitting `Yes` here will do nothing btw):
false error

The updated pr description is is very helpful and detailed btw 👍

Daenyth:
That last error is kind of concerning. Can you reproduce it reliably?

ZeldaZach:
@tooomm I've resolved all bullets and have made the reload thread safe.

ZeldaZach:
So it seems the builds are failing because the carddatabase_test doesn't have the access it wants. When i try to force it to use the thread safe option, it doesn't recognize the database or lock from the main header file (cockatrice/src/main.h) so I'm not certain where to go from here with that. 

I don't believe the tests are that useful. We made them in the past as a crude mock up and I'm sure they can be redone in the future, but the way we're doing it now doesn't seem sufficient.

cc: @Daenyth @ctrlaltca

tooomm:
Nice, thank you!

<br>

>The last error is no longer an issue based on the refactoring I just completed.

Sadly I can still reproduce the error the same as before.
db error settings

And yes @Daenyth, it can easily and 100% reliable be reproduced:
Go to `Settings`, switch to `Deck Editor` submenu, hit `Update Spoiler`, close `Settings` again (hit *Ok* or close the settings dialog). The most normal workflow if you want to update spoilers without restarting the app. Even if the notification appears and you close then it'll pop up.
You literally have to wait until all the spoiler/db stuff silently finished in the background - let's say 5s. But you don't know when the hidden update is finished... simply waiting for the notification isn't enough for me.
Anyway, it's hard for users to understand that, and what's going on.  So trouble is sure to follow. :(
The msg content and the question if you want to *change your database location* is not really related too.

PR #2999, opened 2018-01-08
token creation context menu improvements

tooomm:
- better visual display and separation of card names that other cards get attached to
- make quantity of tokens more clear, the quantity digit and token strength digit visually merged easily 

## Screenshots
- before:
abbey-old
alarm-old
old

- after:
abbey
alarm
new
morph

Note that single numbered tokens and flexible quantity "X" won't change and aren't affected.
Also, nothing changes for translators.

Please double check code context and logic.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2676,7 +2676,7 @@ void Player::addRelatedCardActions(const CardItem *card, QMenu *cardMenu) {
             if (relatedCards.at(0)->getDoesAttach()) {
                 createRelatedCards = new QAction(tr("Token: ") + tr("Attach to ") + "\"" + relatedCards.at(0)->getName() + "\"", this);
             } else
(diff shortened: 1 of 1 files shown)


Psithief:
Perhaps not `x` but &times;?

tooomm:
You suggest to exchange the lower case "times" `x` like this?
`Token: 3x 2/2 Zombie` --> `Token: 3× 2/2 Zombie`

`Token: 2x 1/1 Soldier` --> `Token: 2× 1/1 Soldier`

<br>

If a card has both, a fixed and a variable quantity:
```
Token: 3× 2/2 Zombie
Token: X 2/2 Zombie
```

What do others think?

ZeldaZach:
I’m happy with how it is. That seems a bit frivolous

Psithief:
Frivolous to an English speaker, perhaps!
But you have Unicode support by default, why not use it where it's appropriate?

PR #3000, opened 2018-01-08
deck editor: add placeholder text to search

tooomm:
## Short roundup of the initial problem
No hint/description about card search

## What will change with this Pull Request?
- Let users know that the search will only considers a card's name

## Screenshots
search
Once you start typing the soft greyish placeholder text will vanish.

PR #3003, opened 2018-01-10
card filter: add clear button and placeholder text

tooomm:
- Adds clear button to card search text field
- Adds placeholder text to card search text field
- I rearranged my change from #3000 to find a better line for it

## Screenshots
It now looks like this:
- Before you start typing you see the placeholder text:
before typing
- During typing it disapears, but you'll see a clear button instead to delete all line at once:
during typing

### DIFF
--- a/cockatrice/src/filterbuilder.cpp
+++ b/cockatrice/src/filterbuilder.cpp
@@ -8,76 +8,80 @@
 #include "cardfilter.h"
 
 FilterBuilder::FilterBuilder(QWidget *parent)
-    : QWidget(parent)
+	: QWidget(parent)
 {
-    filterCombo = new QComboBox;
-    filterCombo->setObjectName("filterCombo");
-    for (int i = 0; i < CardFilter::AttrEnd; i++)
-        filterCombo->addItem(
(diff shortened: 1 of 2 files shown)


cockatrice/src/filterbuilder.cpp
```diff
@@ -8,76 +8,80 @@
...
+		);
+
+	QPushButton *ok = new QPushButton(QPixmap("theme:icons/increment"), QString());
+	ok->setObjectName("ok");
+	ok->setMaximumSize(20, 20);
+
+	edit = new QLineEdit;
+	edit->setObjectName("edit");
+	edit->setPlaceholderText(tr("Type your filter here"));
+#if QT_VERSION >= 0x050200
```
ZeldaZach: We build above 5.0.2, so these guards should be removed at this time.

tooomm: I mirrored them from another place in the app.
Will remove both then 👍

tooomm: postponed, see #3005

PR #3007, opened 2018-01-11
Fix deck loader with spaces

### LINKED ISSUE #2998: Cards from clipboard in deck builder without quantity given have first word replaced by quantity
<b>OS: Win 10 Home 64-bit</b>
<b>Cockatrice version: 2.4.1</b>

Open Cockatrice and try to add a card or list of cards from the clipboard in deck builder (Ctrl + Shift + V) without quantities specified. Like _Notion Thief_ instead of _1x Notion Thief_. They're supposed to become _1x Notion Thief_ but become _1x  Thief_ instead. The first word is eaten by the quantity prepended to the card name. This affects every card I tested; length of the first word in the card name doesn't matter.

ZeldaZach:
Fixes #2998 

Tested with 
```
Quicksilver Elemental
3x Craterhoof Behemoth
2 Elvish Mystic
Progenitus
2x Abrade
1 Absorb
```

Before:
screenshot 2018-01-10 19 59 07

After:
screenshot 2018-01-10 19 58 50

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -1,6 +1,5 @@
 #include <QFile>
 #include <QTextStream>
-#include <QVariant>
 #include <QCryptographicHash>
 #include <QDebug>
 #include "decklist.h"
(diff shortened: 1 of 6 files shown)


common/decklist.cpp
```diff
@@ -555,21 +613,37 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         rx.setPattern("\\|.*$");
         line.remove(rx);
 
+        // If the user inputs "Quicksilver Elemental" then it will cut it off
```
ZeldaZach: The actual fix is in this file. Cleanup was done on all files touched to meet new coding standards.

common/decklist.cpp
```diff
@@ -555,21 +613,37 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
...
-            i--;
+        if (i > 0)
+        {
+            // If the count ends with an 'x', ignore it. For example,
+            // "4x Storm Crow" will count 4 correctly.
+            if (line.at(i-1) == 'x')
+            {
+                i--;
+            }
+            else if (! line.at(i-1).isDigit())
```
ctrlaltca: For other readers: here's the main change of this PR

common/decklist.h
```diff
@@ -12,6 +12,7 @@
 // Required on Mac. Forward declaration doesn't work. Don't ask why.
 #include <QtCore/QXmlStreamReader>
 #include <QtCore/QXmlStreamWriter>
+#include <utility>
```
ctrlaltca: I failed to see where this is needed

PR #3008, opened 2018-01-11
Smarter Mana Cost sorting

### LINKED ISSUE #3006: card filter: mana cost search isn't case insensitive
So searching for `ww` won't find you `Celestial Flare` because it's cost are `WW`.

Also the mana cost filter is a exact match filter.
You can't filter for all cards with double blue in their costs with `UU`. It won't return a `4UU` priced card.

Problem there is that you have to keep the order as well: `UU4` won't find the `4UU` spell either, that can be tricky when you search for Abrupt Decay because you *have to* filter for `BG` as `GB` won't return anything.
We need more smartness here! :)

The color filter handles the case here for example:
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/filtertree.cpp#L198

ZeldaZach:
Fixes #3006

If you input a partial mana cost, all cards that meet that mana cost requirement will appear.

Sort the mana costs input as well so `UU1 == U1U == 1UU == uu1 == ...`

Examples:
screenshot 2018-01-10 20 17 21
screenshot 2018-01-10 20 17 33

### DIFF
--- a/cockatrice/src/filtertree.cpp
+++ b/cockatrice/src/filtertree.cpp
@@ -22,8 +22,8 @@ void FilterTreeBranch<T>::deleteAt(int i)
 template <class T>
 int FilterTreeBranch<T>::childIndex(const FilterTreeNode *node) const
 {
-    FilterTreeNode *unconst = const_cast<FilterTreeNode *>(node);
-    T downcasted = dynamic_cast<T>(unconst);
+    auto *unconst = const_cast<FilterTreeNode *>(node);
+    auto downcasted = dynamic_cast<T>(unconst);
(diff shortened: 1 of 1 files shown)


ZeldaZach:
@ctrlaltca Do you think using the Qt::CaseInsensitive check would be more efficient?

ctrlaltca:
> Do you think using the Qt::CaseInsensitive check would be more efficient?
Maybe a little bit, since the Qt implementation of indexOf() is comparing characters one by one, but since mana cost strings are typically really short, it won't be noticeable.
Of course i'd love to be proven wrong with some actual performance benchmark.

Issue #2183, opened 2016-09-22
clang-format stuff


Daenyth:
https://arcanis.me/en/2015/10/17/cppcheck-and-clang-format/

And the branch.

(this ticket is a reminder for me for later)

Daenyth:
wip style
```
BasedOnStyle : webkit

---
Language: Cpp
PointerAlignment: Right
ColumnLimit: 120
---
Language: Proto
DisableFormat: true
```

ZeldaZach:
Any news on this?

Daenyth:
No

tooomm:
Since all repo branches got lost...

Here a write up from @Daenyth about his initial process on gitter:
https://gitter.im/Cockatrice/Cockatrice?at=5a5f9b4d5ade18be398aa836
>idgaf about brace style as long as it's consistent
In fact if someone can please pick up my clang-format branch
I have a beginning of a clang-format config file
the way I did it was to run clang-format over the codebase, examine the diff, find something that got a big diff, and see if clang-format had a toggle to make that diff smaller, then reset the codebase and run with the new settings
tedious iterative process
that said, if someone wants to just pick a style and commit to it and do one huge diff of formatting the whole codebase
When this is done, it means we will never have another comment on code format in PR ever again
and I think we can all appreciate how nice that would be
especially for new contributors!

PR #3009, opened 2018-01-11
Contributing Doc Updates

ZeldaZach:


ctrlaltca:
The fixes are good, i only see two debatable items here:
1. wether to advice CLion: i know it's good, but if you tell people to use it then they'll expect you to support it and help them fixing their problems with CLion;
2. the brace change: one brace per line is not the currently used style, do we really want to change it?

My personal opinion is:
 * no to item 1, unless the working is changed to something softer like eg. "some developers like to use.."
 * indifferent to item 2, it's just a matter of convenience on what style the current developers prefer to maintain.

@Daenyth your opinion is definitely needed here (others, too).

ZeldaZach:
Updated based on feedback

ZeldaZach:
@Daenyth I'll start looking into clang-format soon

tooomm:
`clang-format` issue is https://github.com/Cockatrice/Cockatrice/issues/2183

PR #3018, opened 2018-01-16
Trim all inputs (except password)

### LINKED ISSUE #3016: stripe leading and trailing spaces from host names
As reported on gitter: ([January 14, 2018](https://gitter.im/Cockatrice/Cockatrice?at=5a5ae19fba39a53f1a11fd60))
>hi guys. when im logging in it says host not found
for the main server
son of a gun.  disregard.  **i had a space at the end of the host**

We should be ably to detect, disallow or even fix that for the user.
Could be an potential error from e.g. copy & paste nobody realizes.

Host names never contain any spaces and we can safely delete them?

ZeldaZach:
## Related Ticket(s)
- Fixes #3016

This will trim all inputs to the login details except for the password.


Nightfirecat:
Will the saved server/username details be trimmed as well, or would they be the same as the given input?

ZeldaZach:
@Nightfirecat I trim all fields except for passwords. They will save the trimmed versions.

PR #3025, opened 2018-01-19
Fix sideboard

### LINKED ISSUE #3022: Uniform disable behavior for "Sideboard locked" button
The `Sideboard locked` button is the only one not disabled as long as no deck is loaded.
Clicking on it does nothing at that point:

sideboard-button

ZeldaZach:
Fix #3022 

General code cleanup

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,29 +15,41 @@ skip_commits:
     - Dockerfile
     - LICENSE
 
-image: Visual Studio 2015
+image: Visual Studio 2017
 
 cache:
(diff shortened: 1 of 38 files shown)


ZeldaZach:
@tooomm No, I can't remove the red borders, sorry. They're hard coded in between red and green; adding another state would be confusing and not worth it imo

tooomm:
Actually, I'm seeing issues with this build.

Can you properly play a game after you load a deck @ZeldaZach ?
I can't draw or move cards.

PR #3030, opened 2018-01-19
contributing: add toc + little tweaks

tooomm:
- add toc with links for easy navigation
- fixed `steps` for translations
- little wording tweaks and updates

## Preview
https://github.com/Cockatrice/Cockatrice/blob/tooomm-patch_contributing/.github/CONTRIBUTING.md

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -231,10 +231,12 @@ It's now suggested to disable the parameter using:
 ```sh
 cmake .. -DUPDATE_TRANSLATIONS=OFF
 ```
-Now you are ready to propose your change. Once your change gets merged, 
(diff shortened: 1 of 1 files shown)


ctrlaltca:
> It starts with step 2, followed by step 5... :)
> Since you wrote that @ctrlaltca, can you check please and let me know?

It's probably a copy/paste fail.
I suppose the steps were written somewhere else before, and were something like:
1. add a translatable string
2. extract it (step 2)
3. push to transifex
4. translate on transifex
5. get translated language files from transifex into cockatrice (step 5)

tooomm:
Thanks @ctrlaltca.
I updated the starting post and added a commit to fix the `steps`.

PR #3034, opened 2018-01-20
Fix Bad Gameplay

ZeldaZach:
Will revert #3025 due to complications. This will need to be investigated in the future.

Puts the fix into place for sideboard button, so still functional change.


tooomm:
This mostly fixes the issue described in https://github.com/Cockatrice/Cockatrice/pull/3025#issuecomment-359028120 
I can draw/play/move cards again! 👍 

Only thing left is the misplaced battlefield:
misplaced

PR #3039, opened 2018-01-20
Update Servatrice MySQL Search Path

woogerboy21:
Removed the older versions of the mysql library install paths and updated to reflect the 5.7 product paths.
This will fix the cmake search failures for windows machines.

### DIFF
--- a/servatrice/CMakeLists.txt
+++ b/servatrice/CMakeLists.txt
@@ -82,7 +82,7 @@ if(UNIX)
         SET(MYSQLCLIENT_DEFAULT_PATHS "/usr/lib64" "/usr/local/lib64" "/usr/lib" "/usr/local/lib")
     endif()
 elseif(WIN32)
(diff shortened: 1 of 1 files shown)


tooomm:
@ctrlaltca Are you saying that servatrice might not run properly with the 32bit version we provide?

ctrlaltca:
It will run, but it probably won't be able to connect to a mysql server. Given that the number of users running a database-backed servatrice server instance under 32-bit windows tends to zero, we could just drop it.

woogerboy21:
The path listed will find both 32 but or 64 but depending on the installation?  If the MySQL server is installed on a 32bit version of Windows and using the 32bit version of the server the path to the libraries is under program files.  If it’s 64bit OS with 64bit server it’s still program files.  If it’s 64bit is with 32bit client it’s program files (86).

Not sure exactly what you mean by the number of users using 32bit versions of the server.  Both paths are represented in the CMake search strings.

The old lines simply had older numerical versions of the MySQL server paths that haven’t been used in some time.

Issue #2982, opened 2017-12-29
travis: properly implement ccache for mac


tooomm:
Travis does support ccache. But even when [following their docs](https://docs.travis-ci.com/user/caching/#ccache-on-OS-X) and adding `cache: ccache` and exporting a path as we do already, it's not working as expected.

I found this articles which explain how to do it right. This would finally bring mac build time down, too!
In their example it goes down from over 14 to under 8 minutes:
http://www.brianlheim.com/2017/08/20/supercollider-travisci.html (he referenced [this](https://crascit.com/2016/04/09/using-ccache-with-cmake))
>"for our two linux builds, using ccache in travis was super easy – all i had to do was add the value 
`ccache` to the `cache:` section of our .travis.yml file. i instantly saw fantastic build time speedup, which i’ll discuss in a bit. for macOS, the solution was a bit more complicated. [travis claims](https://docs.travis-ci.com/user/caching#ccache-cache) that you just need to add a few extra lines to your script to get ccache to work for macOS builds:
>
> *(code copy from travis docs)*
>
>but as far as i could tell that’s simply not true. reading the build logs, it was clear that xcode was still being invoked, and looking at the build times, it was very clear that nothing was changing. our build time was still hovering around 15-16 minutes."

It would be great to have a drastically lower build time for macOS with a proper configuration and ccache usage: It's now ~100s (linux) vs. ~850s (osx) basically...

With #3041 osx is down to ~750s at least. :)

tooomm:
More detailed observations and logs:

From a tavis linux build log with correctly recognized ccache:
```
...

$ bash ./.ci/travis-compile.sh
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/lib/ccache/gcc
-- Check for working C compiler: /usr/lib/ccache/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib/ccache/g++
-- Check for working CXX compiler: /usr/lib/ccache/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done

...
```

In comparison, the same build also triggered a osx job at travis:
```
...

$ bash ./.ci/travis-compile.sh
[...]

PR #3041, opened 2018-01-21
travis update (osx + linux)

tooomm:
- Caching all homebrew directory resulted in a very big cache file which took a long time to download, validate, check for changes and upload again (~50s longer than the one additional "brew update" run)

- Remove homebrew caching
- Update to last xcode image with 10.11 support (Xcode 8 GM):
https://docs.travis-ci.com/user/reference/osx/#OS-X-Version
- Rename `.travis` folder for scripts to `.ci` to place appveyor stuff there too
- Limit `BUILDTYPE=Release` builds to jobs from `master` branch which are no PR's to prevent unneeded builds and return a useful result more quickly in the GitHub interface.
Travis only allows a certain number of jobs that run in parallel, this reduces our own queue we produce.
[Release = Debug just without tests](https://github.com/Cockatrice/Cockatrice/blob/tooomm-travis_osx/.ci/travis-compile.sh#L21) -> so we should not miss out on something
    - Docs: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Jobs
- Little tweaks

<br>

While this slightly helps, the big speed improvement for mac builds has to come with https://github.com/Cockatrice/Cockatrice/issues/2982


.travis.yml
```diff
@@ -1,64 +1,48 @@
...
-        - qttools5-dev
-        - qttools5-dev-tools
-        - qtmultimedia5-dev
-        - libqt5multimedia5-plugins
-        - libqt5svg5-dev
-        - libqt5sql5-mysql
+    if: branch = master
   - os: osx
-    osx_image: xcode7.3
+    osx_image: xcode8
```
ctrlaltca: Updating to xcode 8 means cockatrice will only run on osx => 10.12 sierra

tooomm: As said above:
>Update to **last xcode image with 10.11 support** (Xcode 8 GM):
>https://docs.travis-ci.com/user/reference/osx/#OS-X-Version

from the [build log](https://travis-ci.org/Cockatrice/Cockatrice/jobs/331430456#L4):
`instance: 06841a6e-1629-4a9d-81d1-42a7e6d7b3d6 travis-ci-osx10.11-xcode8-1479248015`

Only starting with `xcode8.3` onwards it'll be osx =>10.12!

<br>

Is it worth to consider switchting to 10.12 yet? It cuts off some older hardware and 10.11 is just ~2 years old I think.

ctrlaltca: Reading Apple's docs it says that xcode 8 only installs the SDK for macOS >=10.12, but probably travis handles this different. Sorry for the noise.

tooomm:
Expanded the job conditions for `Release` builds to not run on PR's since every PR is build against master.
It'll still run on every merge to our master, or manually triggered builds on master like tagging or rerunning travis. Therefore deploying (which only happens from `Release` builds) is not affected.

Before:
before

After:
after

PR #3028, opened 2018-01-19
Clang-format

ctrlaltca:
## Related Ticket(s)
- Fixes #2183

## Short roundup of the initial problem
Correctly formatting the code is hard, documenting the style is even harder

## What will change with this Pull Request?
- Added a .clang-format config file: maybe it still needs some tweaking, comments are welcome
- Run clang-format on the linux-debug travis build (the same one that run test)
- unrelated: set a cmake policy to avoid garbage output on cmake 3.10

## Screenshots
win31-1-1

PR #3047, opened 2018-01-25
[WIP] ci: scripts improvements

tooomm:
- Cleanup CI config files
Move appveyor scripts into their own files (like we do it with Travis)
--> Results in a cleaner and more readable config file, and CI log 👍 
- ~~Includes cmake 3.10 fix from https://github.com/Cockatrice/Cockatrice/pull/3028:
`set a cmake policy to avoid garbage output on cmake 3.10`~~
- Move clang-format check to the start of Linux builds for faster feedback (Debug only)
- Make sure to have all scripts in `100755` mode
- Remove clang-format dependency for osx (we only run the clang check on linux)
- Remove old prefix needed for former non-bundled qt installs on linux
- Travis: Improve readability of conditional job runs
- Use the special clang-format bash script provided in the repo
- **TODO:** fix appveyor script calls (working on [/branch](https://ci.appveyor.com/project/Daenyth/cockatrice/build/2.4.1-branch-tooomm-appveyor_updates-build-891), but same commit fails on [/pr](https://ci.appveyor.com/project/Daenyth/cockatrice-8k7y5/build/2.4.1-branch-master-build-1252) build)
- **TODO:** Check https://github.com/Cockatrice/Cockatrice/pull/3047#discussion_r166948375


ctrlaltca:
Please don't split the powershell commands out of appveyor.xml. The env variables are declared in the main file but used around the other files, making it harder to follow the flow of the process.

tooomm:
The cmake fix is now a separate PR: https://github.com/Cockatrice/Cockatrice/pull/3048

I won't be around over the weekend starting today, we can talk about this later then. 👍

.ci/travis-dependencies.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 
+if [[ $TRAVIS_OS_NAME == "linux" ]] ; then
```
ZeldaZach: I see no reason to move this around

tooomm: It doesn't matter code wise, yes.

The idea is to have all travis scripts (travis-checks, -dependencies, -compile) with the same structure.
1. Check for `linux` (also first in travis build matrix)
2. Check for `osx`
3. Check for `Debug`
4. Check for `Release`

And don't have mac checks at the beginning of one file and at the end of the next.

Daenyth: 👍 consistency

.ci/travis-checks.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+if [[ $TRAVIS_OS_NAME == "linux" ]]; then
+  if [[ $BUILDTYPE == "Debug" ]]; then   ## TODO: generalize h cpp files search to top-directory? run bash script by zach here?
+    echo "Checking code style..."
+    clang-format -i \
```
ZeldaZach: We should probably update this to just run the clangify file?

ZeldaZach: Or, at least, copy what it does in terms of file coverage

tooomm: I defensively look into running your script there! 👍

ZeldaZach:
So i'm leaving this up to @ctrlaltca on what he wants to do with this

tooomm:
There is still the [WIP] label here for a reason. 😉 
The changes applied to this branch/pr are a bit broader now and not only limited to the initial appveyor script split off.

Anyway, I had no chance to chat with ctrlaltca yet.

.ci/travis-compile.sh
```diff
@@ -2,51 +2,32 @@
...
-    git diff --quiet || (
-      echo "*****************************************************";
-      echo "***  This PR is not clean against our code style  ***";
-      echo "***  Run clang-format and fix up any differences  ***";
-      echo "***  Check our CONTRIBUTING.md file for details!  ***";
-      echo "***                  Thank you ♥                  ***";
-      echo "*****************************************************";
-    )
-    git diff --exit-code
+    prefix="-DCMAKE_PREFIX_PATH=$(echo /opt/qt5*/lib/cmake/)"
```
ctrlaltca: I think this is not needed anymore, since linux builds are using the bundled qt version. Can you try `prefix=""`?

clangify.sh
```diff
@@ -12,7 +16,33 @@ if hash clang-format 2>/dev/null; then
...
+				echo "";
+				echo "*****************************************************";
+				echo "***  This PR is not clean against our code style  ***";
+				echo "***  Run clang-format and fix up any differences  ***";
+				echo "***  Check our CONTRIBUTING.md file for details!  ***";
+				echo "***                  Thank you ♥                  ***";
+				echo "*****************************************************";
+				)
+			fi
+			git diff --exit-code ## correct position? what is this used for? original script: https://github.com/Cockatrice/Cockatrice/blob/994a643d9c4e08dad1dfd7b13272598fdb42b196/.ci/travis-compile.sh
```
ctrlaltca: At the beginning of this script there's a `set -e` that means (man page content):
```
 -e  Exit immediately if a command exits with a non-zero status.
```
Now, `git diff --exit-code` is used to: (from: https://git-scm.com/docs/git-diff#git-diff---exit-code)
```
Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.
```
Basically, the idea is that if clang-format makes any change to the files, `git --diff` will notice it and exit with a status of 1, marking the script (and thus the build) as failed.
This probably won't work anymore since this check is now inside the `clangify.sh` subscript, and nothing is checking its exit code.

.ci/travis-compile.sh
tooomm: We generally do this a few lines above already:

clangify.sh
tooomm: TODO

ZeldaZach:
Feel free to re-open when you're able to address all of the comments

Issue #2226, opened 2016-10-18
Discuss: Fetch server list from online


ctrlaltca:
Currently cockatrice (and the webclient) bundles an hardcoded list of servers.
We could serialize the info present at https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers into a downloadable file and publish it into a new repository (cockatrice-servers).

When cockatrice's connect dialog opens, it will:
- fetch the server list from a predetermined url;
- try to ping each server name, and display the result;
- (advanced, needs proper design): connect without logging in and ask the server for basic informations (requires registration / accepts new registrations / has free slots for new users / current num of users and games / planned shutdown / whatever )

woogerboy21:
What idea's did you have on who would maintain the list and were would it be stored?

Daenyth:
We should maintain the official list as a json on github.

I'd like to wrap this into a more general login panel that introduces a protocol update which allows a client to (without a session) ask a server about itself - online status, user count, registration rules, etc.

Daenyth:
That would be #742 I believe

woogerboy21:
It should prob also include #2189 as well.

skwerlman:
> who would maintain the list

I would be fine maintaining the list of public servers as a separate repo on here.


> Should the "pre-determiined url" be client side editable?

I can see this becoming a source of very confusing problems for users

Daenyth:
I'll write up what I picture for this tonight (hopefully). at work atm so I can't get into detail. But we should store server info as a whole unit and let users add/remove, rather than individual unconnected text fields

woogerboy21:
I actually started to think about this in regards to implementing the protocol adjustments and code to make the requests and almost out the git-go (see what i did there) came up with some questions.  Seeing as ping is not something built in to any library (like qt libraries) we would have to write a custom set of functions to make it happen.  Is ping really something we want to do?   What is the thought process behind using ping?  I am assuming you want to test to see if a server is responding on the IP that the name resolves to and if it does yet the connection to the server isnt made we could say something like "server is up but not accepting connections? 

If this is the case how do we even know if the IP address resolved from the name is even a cockatrice server? Some one could easily fat finger a server name and this result could occur and the user would simply think "oh woogerworks is down" when in fact they had made the error by a bad copy/paste or mis-spelling.

Or are we simply interested in making a connection and if it succeeds request the data we are looking for and if it fails assume the server is offline/old?

[...]

PR #3051, opened 2018-01-27
Domain change

dev-id:
`.es` es no mas

Server and database from cockatric.es is now active on cockatrice.dr4ft.com

Wiki updated


ZeldaZach:
Unfortunately this doesn't flush the current saved version for all users, just people who download the client as new or manually flush their settings or do a full re-install.... Maybe something to improve upon with #2226

PR #3053, opened 2018-01-28
Replace CardInfo* and Carset* with smart pointers

ctrlaltca:
Currently most of the reported crashed are due to to the "game cleanup" routine, where cards (to be more precise, `CardInfo *`) are removed from the zones and deleted.
Moreover, it seems like any attempt to reload the card database while at least one card is used somewhere (in a game, in he deck editor, etc..) leads to a crash, eg. see #3042.
This PR tries to mitigate the problem switching from raw pointers to smart pointers. 
The idea is that we never "manually" delete the card objects, since each card will delete itself when nothing uses it anymore.
This is not a panacea for all memory related crashes, but i'm curious if it's enough to stop at least #3042; it's not meant to be merged right now, but only if it's actually proved that it prevents some crashes and doesn't create more problems (performance, other crashes, ..).
@ZeldaZach since you seems to able to trigger #3042, i'll be glad if you test it against this branch.

### DIFF
--- a/.ci/travis-compile.sh
+++ b/.ci/travis-compile.sh
@@ -21,7 +21,32 @@ if [[ $BUILDTYPE == "Debug" ]]; then
   cmake .. -DWITH_SERVER=1 -DCMAKE_BUILD_TYPE=$BUILDTYPE $prefix -DTEST=1
   make -j2
   make test
+
+  if [[ $TRAVIS_OS_NAME == "linux" ]]; then
+    cd ..
+    clang-format -i \
+      common/*.h \
+      common/*.cpp \
+      cockatrice/src/*.h \
+      cockatrice/src/*.cpp \
+      oracle/src/*.h \
+      oracle/src/*.cpp \
+      servatrice/src/*.h \
+      servatrice/src/*.cpp
+    
+    git clean -f
+    git diff --quiet || (
+      echo "*****************************************************";
+      echo "***  This PR is not clean against our code style  ***";
+      echo "***  Run clang-format and fix up any differences  ***";
+      echo "***  Check our CONTRIBUTING.md file for details!  ***";
+      echo "***                  Thank you ♥                  ***";
+      echo "*****************************************************";
+    )
+    git diff --exit-code
+  fi
 fi
+
 if [[ $BUILDTYPE == "Release" ]]; then
   cmake .. -DWITH_SERVER=1 -DCMAKE_BUILD_TYPE=$BUILDTYPE $prefix
   make package -j2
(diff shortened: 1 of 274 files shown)


ZeldaZach:
Ok, so this PR fixes the issues described. I pushed my own set of changes to fix the bad compilations. 

clang-format seems to be weird as I brew installed it then ran
```bash
clang-format -i -style=file cockatrice/src/*.cpp cockatrice/src/*.h
```

but the linux build doesn't seem to like it. Good enough here

ctrlaltca:
Thanks for the fixes! About clang-format, You may want to run it also on `oracle/src/*.cpp, *.h`, common/, servatrice/, etc..

ctrlaltca:
> Ok, so this PR fixes the issues described.

This is a really good news. Still 2 things needs to be checked:
1. if RAM is really freed when objects should be deleted (i.e. if RAM usage doesn't ramp up)
2. if it created any side effect, eg. pointer usage that i didn't consider to cards not being loaded or similar

Daenyth:
You should be able to run clang-format on just the directories, `clang-format cockatrice/ oracle/ servatrice/` or so. Maybe you need to give it an extra argument but I remember doing that in my testing.

tooomm:
We need a paragraph with explanations about how to install+apply clang-format in the CONTRIBUTING file, too!

ZeldaZach:
So I want to merge this in and publish a beta to get feedback from the user base on any potential increases in ram

ZeldaZach:
Ok so after testing, there seems to be a memory leak. Prior version, I ran and used 1.8%. When I launched this PR and tested 3 times, the Mem usage started around 1.8% and went up to 2.7% (Around 0.3% each run)

If you think this is an acceptable loss for now, we can go forward. Otherwise, We can see what's up with the clear component.

ZeldaZach:
[1517379957-trice.log](https://github.com/Cockatrice/Cockatrice/files/1680655/1517379957-trice.log)

a valgrind log I generated. Seems that on linux we had a 0.4% increase per run of the database reload. There appears to be no other additional leaks that are tolling the system, and I don't see this as a major problem as a reload of the client fixes it. How often will people reload without closing? once? twice? I think this is fine to roll in for now

ZeldaZach:
@tooomm Can you make a ticket

tooomm:
My understanding is that you took care of the merge - so you accept the responsibility to (instantly) open the requested ticket.
If you can't do that right away because you're on the road for example, you shouldn't merge and keep the reminder around for everybody to see. 😉

Who can be sure I'll ever read this or care?

ZeldaZach:
Well ok then, grumpy 😜

PR #3056, opened 2018-01-30
Fix mwDeck importing

### LINKED ISSUE #3038: Wrong format when loading .mwDeck
<b>OS:</b>
Win 7 SP2

<b>Cockatrice version:</b>
Version 2.4.0 (2017-11-19)
__________

This is what the .mwDeck original file looks like:

It is an old deck I saved in Magic Workstation.

When I load a deck of type .mwDeck, each card entry is loaded with an extra blank space in front of it and each card has only lower-case letters:

This means that Cockatrice doesn't recognize the card correctly. No image or rules text is displayed.

When I save the deck as a .cod file, each card gets a blank space in front of its name and it is saved in lower-case only:

Daenyth:
## Related Ticket(s)
Fix #3038

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -562,9 +562,9 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         }
 
         // Filter out MWS edition symbols and basic land extras
-        QRegExp rx("\\[.*\\]\s?]");
+        QRegExp rx("\\[.*\\]\s?");
         line.remove(rx);
-        rx.setPattern("\\(.*\\)\s");
+        rx.setPattern("\\\s(.*\\)");
         line.remove(rx);
 
         // Filter out post card name editions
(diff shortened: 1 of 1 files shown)


common/decklist.cpp
```diff
@@ -562,9 +562,9 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         }
 
         // Filter out MWS edition symbols and basic land extras
-        QRegExp rx("\\[.*\\]");
+        QRegExp rx("\\[.*\\]\s?");
         line.remove(rx);
-        rx.setPattern("\\(.*\\)");
+        rx.setPattern("\\\s(.*\\)");
```
tooomm: Needs the `\` within `\s` another `escape \` infront?
--> `\\s`?

tooomm: Actually, @dino572 fixed this and managed to created his own PR: https://github.com/Cockatrice/Cockatrice/pull/3057

Daenyth:
Closing for #3057

PR #3057, opened 2018-01-30
Fix mwDeck edition tag parsing on deck load

### LINKED ISSUE #3038: Wrong format when loading .mwDeck

dino572:
## Related Ticket(s)
- Fixes #3038 

MWS uses a special style for their .mwDeck format in decklists.
They add the set tag in square brackets before the card name (e.g. `[RIX]`), and append an additional number in round brackets (e.g. `(3)`) to specify a particular card if there are more than one with the same name in that set (several forest cards with different art for example).

The code in place to remove those additions left a space behind.

- RegEx update

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -562,9 +562,9 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         }
 
         // Filter out MWS edition symbols and basic land extras
-        QRegExp rx("\\[.*\\]\s?]");
+        QRegExp rx("\\[.*\\]\s?");
         line.remove(rx);
-        rx.setPattern("\\(.*\\)\s");
+        rx.setPattern("\\\s(.*\\)");
         line.remove(rx);
 
         // Filter out post card name editions
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -564,7 +564,7 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         // Filter out MWS edition symbols and basic land extras
         QRegExp rx("\\[.*\\]\\s?");
         line.remove(rx);
-        rx.setPattern("\\\\s(.*\\)");
+        rx.setPattern("\\(.*\\)");
         line.remove(rx);
 
         // Filter out post card name editions
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -562,9 +562,9 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         }
 
         // Filter out MWS edition symbols and basic land extras
-        QRegExp rx("\\[.*\\]");
+        QRegExp rx("\\[.*\\]\s?]");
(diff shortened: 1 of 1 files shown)


tooomm:
Happy to see you managed to successfully create your first pull request @dino572 ! ;)

Our automated CI tests pass, we will review and test the change now.

Since you are on windows, you can test it for yourself too without the need to compile the app for yourself.

Just click on `Show all checks` at the bottom and select `Details` for **AppVeyor**.
Then, at the appveyor page you select the 64bit (first one) or 32bit (second entry in the list) build. Click on it and select `Artifacts`. There you can download the installer with your change in!

Feel free to report back if it works for you. 👍

dino572:
@tooomm Thanks for the instructions! I now have two portable installations: One from the official release and one from this change. In the official release, parsing is wrong but in this one, parsing is right. The cards show up correctly when I load a .mwDeck. However, I still need to change back the second line I changed because it was working correctly before. Let's hope I make that change correctly this time!

dino572:
It's still not loading this type of line correctly:
    2 [ON] Forest (3)
    1 [XLN] Island (2)
    3 [MI] Plains (4)

It becomes:
2  forest
1  island
3  plains

common/decklist.cpp
```diff
@@ -564,7 +564,7 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
         // Filter out MWS edition symbols and basic land extras
         QRegExp rx("\\[.*\\]\\s?");
         line.remove(rx);
-        rx.setPattern("\\(.*\\)");
+        rx.setPattern("\\\\s?(.*\\)");
```
tooomm: This looks not correct.

>I am adding `\\s?` at the start of the regex.

You had the right thought, but it should look a bit different.
Did you try `rx.setPattern("\\s?\\(.*\\)");`?

dino572:
Now, it becomes
`20 forest (3)`
I'll try what tooomm said now.

dino572:
I think it works now. Lands also show up correctly now. Can this go into the official client? This will make it easier for me to transition from MWS to Cockatrice!

Daenyth:
@tooomm would you make new tickets for that? I expect the lowercasing is accidental, used for doing card name lookups

tooomm:
Thanks for your contribution and this fix @dino572!

Issue created at  #3064

Issue #2647, opened 2017-04-25
Osx: investigate usability of Qt versions after 5.7


ctrlaltca:
<b>OS:</b>
macOS

<b>Cockatrice version:</b>
Any

Some new issues arose when our macOS packages started bundling Qt5.8.
This happened quite transparently to us, since out packages are built on Travis and install Qt from homebrew. When homebrew updated to Qt5.7, we started using it.
When we noticed that Qt5.8 was the cause of such problems, we forced the use of the last known working Qt5.7 (see commits 7f057612bf575422c55d050ea88bd49c71c37f16 and 38ad71b06a55fd8915c589bf07bf498cc71767a9)

This issue is a collector for all issues related to macOS and Qt5.8 that needs to be checked before trying to update the Qt version we use in the future:

 * #2527 Crashes on QNetworkManager
 * #2490 Keybinds disabled after sideboard

skwerlman:
Weirdly, I don't experience any of those issues on Arch Linux with Qt5.8... Maybe its an issue with brew's specific build of Qt5.8?

ctrlaltca:
@skwerlman: MacOS only

ctrlaltca:
Just tested Qt 5.9.1 for bug #2490 (keyboards shortcuts after sideboard), and it's still broken :(
I'km afraid we'll have to workaround it

ZeldaZach:
Sounds like we'll be sticking with 5.7.1 for a while...

tooomm:
Did somebody retest with the recent Qt 5.10 yet?
Too bad that Qt 5.9 LTS has the same issues still according to @ctrlaltca.

tooomm:
Is there an issue at Qt for this? Are they aware of the problem?

tooomm:
`qt@5.7` is discontinued at homebrew: https://github.com/Homebrew/homebrew-core/pull/23165#issuecomment-362270835

We have to either use `qt@5.5` for now as a temporary solution, or look into Qt 5.10 with this issues.

NickBeeuwsaert:
For #2490, it looks like the problem isn't caused by side-boarding, but just opening a dialog will cause all sub-menu item shortcuts in the game menu to stop working (but shortcuts like for Next Turn will still be available). I can put together a minimum working example if anyone wants to look at it / see if they can find a workaround

Daenyth:
That would be helpful

NickBeeuwsaert:
Alright here you go:

```
[...]

PR #3066, opened 2018-02-01
travis: temporary downgrade mac builds to qt5.5

tooomm:
[`qt@5.7`](http://formulae.brew.sh/formula/qt@5.7) is no longer available at homebrew

- Use [`qt@5.5`](http://formulae.brew.sh/formula/qt@5.5) instead
- This is only a temporary workaround because `qt@5.5` won't be there forever
The general [`qt`](http://formulae.brew.sh/formula/qt) package defaults to Qt 5.10 at the moment which conflicts with https://github.com/Cockatrice/Cockatrice/issues/2647

<br>

Can somebody with a mac give this a run and test if the client works the same before we deploy dmg's based on that?

### DIFF
--- a/.ci/travis-dependencies.sh
+++ b/.ci/travis-dependencies.sh
@@ -2,7 +2,7 @@
 
 if [[ $TRAVIS_OS_NAME == "osx" ]] ; then
   brew update
-  brew install ccache clang-format protobuf qt@5.7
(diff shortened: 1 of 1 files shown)


ZeldaZach:
This should be fine as I've built on 5.5 in the past.

We do need to address the style problems of Qt5.8 and newer so we can support the entire qt package. This should be a high priority.

PR #3080, opened 2018-02-05
Relation checks added to relevant fields in card filter

### LINKED ISSUE #2205: Deck editor/card list filter: Add "Greater than / Less than" filters
Feature request: Allow greater than / less than in deck editor filters (for CMC, Loyalty, Power, and Toughness... and maybe rarity, if we're ambitious.)

I think the easiest way to implement is to allow and translate the characters > < and = in the data entry box, rather than adding new selectors.

Vafthrudnir:
## Related Ticket(s)
- Fixes #2205 - The issue mentions applying this to rarity as well, but the same method couldn't be used there, I think that would need a much larger modification

## Screenshots
filter_relations


tooomm:
>The issue mentions applying this to rarity as well, but the same method couldn't be used there, I think that would need a much larger modification

To not have this for "rarity" is totally fine 👍

cockatrice/src/filtertree.cpp
```diff
@@ -295,6 +295,27 @@ bool FilterItem::acceptRarity(const CardInfoPtr info) const
...
+        int termInt = term.mid(1).toInt();
+        if (term.startsWith('<')) {
+            result = (cardInfo < termInt);
+        }
+        else if (term.startsWith('>')) {
+            result = (cardInfo > termInt);
+        }
+        else if (term.startsWith('=')) {
+            result = (cardInfo == termInt);
+        }
```
ZeldaZach: We don't seem to have a default case here, which is a bit troublesome. Maybe default it to ==?

cockatrice/src/filtertree.cpp
```diff
@@ -295,6 +295,27 @@ bool FilterItem::acceptRarity(const CardInfoPtr info) const
...
 
+bool FilterItem::relationCheck(int cardInfo) const
+{
+    bool result, conversion;
+
+    result = (cardInfo == term.toInt(&conversion));
+    if (!conversion)
+    {
+        int termInt = term.mid(1).toInt();
+        if (term.startsWith('<')) {
```
ZeldaZach: Also thinking just a bit further... can this possibly be simplified to

`result = (cardInfo term.left(1) termInt)` or something along those lines?

Vafthrudnir: It would be really nice, but I'm not sure how could that work in C++. Can Qt somehow convert a char into a logical operator?

cockatrice/src/filtertree.cpp
```diff
@@ -295,6 +295,24 @@ bool FilterItem::acceptRarity(const CardInfoPtr info) const
     return false;
 }
 
+bool FilterItem::relationCheck(int cardInfo) const
+{
+    bool result, conversion;
+
+    result = (cardInfo == term.toInt(&conversion));
```
ZeldaZach: A comment here would be really useful

tooomm:
`<`, `>` and `=`... all work nicely! 👍 

What about `<=` and `>=`?

ZeldaZach:
Good point tom

tooomm:
Another highly appreciated change, thanks @Vafthrudnir!

(`<=` etc. are working great, too. 👏 )

PR #3083, opened 2018-02-06
clang-format full filesystem

ZeldaZach:


tooomm:
Hmm 4k lines again... did @ctrlaltca miss all these in #3028?!

ZeldaZach:
Possibly. He missed sub directories mainly. I ran it with the clang-format-all file I found on github

ctrlaltca:
I skipped the directories containing 3rd party libraries, like qt-json (mtgjson parsing) and qtsmtp (used to send email from servatrice).
The reason why i skipped them is that applying our modifications to those files makes it a lot harder to merge changes from upstream in the future.

tooomm:
>I ran it with the clang-format-all file I found on github

This one, right? https://github.com/eklitzke/clang-format-all

They also state on their site:
>**Note:** While clang-format is a powerful tool for enforcing a consistent coding standard, you should be aware that different versions of clang-format may format the same code differently. As one example that I know of, in C++ long vector or set literals (say, a dozen elements or more) in clang-format 3.7 are generally formatted with one entry per line, but under clang-format 3.5 the same literals may be formatted with multiple entries per line if the entries are short enough. Therefore if multiple people are contributing to the same project, you should strive to ensure that they're all using the same version of clang-format.

So we should add the [clang version we check against](https://github.com/Cockatrice/Cockatrice/blob/master/.travis.yml#L34) to the upcoming explanation in contributing too.

Did you run clang-format version 5, too? @ZeldaZach

Daenyth:
I think we should not run this on the vendor dependencies. Can we provide people a bash script that will run format on the files we care about?

ZeldaZach:
I'll look into a bash script quickly

PR #3085, opened 2018-02-06
Clang script

ZeldaZach:


clangify.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# This script will run clang-format on all non-3rd-party C++/Header files.
+
+set -e
+
+if hash clang-format 2>/dev/null; then
```
Daenyth: Nice trick, forgot about `hash`

PR #3088, opened 2018-02-07
Consistent use of 'beta' tag across the project

tooomm:
Before, we used terms like `development snapshots` in the UI and code to name our "nightly" builds back when we deployed all commits from master to bintray.

- Consistently use `Beta`
That's what it is, and we use all the time. (Pre-release names, tags, semver appendixes...)


tooomm:
commit squash went wrong somehow 😕

Issue #2546, opened 2017-03-27
Improve filter for split cards


dev-id:
When using filters in deck editor, split cards are difficult to locate.  Here is an example split card entry from cards.xml:

```
        <card>
            <name>Far // Away</name>
            <set rarity="" muId="369042" num="127a">DGM</set>
            <color>U</color>
            <color>B</color>
            <manacost>1U // 2B</manacost>
            <cmc>2 // 3</cmc>
            <type>Instant // Instant</type>
            <tablerow>3</tablerow>
            <text>Return target creature to its owner's hand.
Fuse (You may cast one or both halves of this card from your hand.)

---

Target player sacrifices a creature.
Fuse (You may cast one or both halves of this card from your hand.)</text>
        </card>
```

- [x] Filtering for CMC 2, 3 or 5 will not find the split card
- [ ] nor will manacost `1U` or `2B` (or `3UB` though that would not be expected.) 

`type` and `text` may not be well handled, and using `num` in image downloader url may not work as expected.

tooomm:
With amonkhet the rules even change and (all!) split cards have a different cmc now:
http://magic.wizards.com/en/articles/archive/news/amonkhet-split-card-rules-changes-2017-04-04

So as long as the card isn't played on the stack (well, one half of it) the cmc of the card are the combined costs of both.

So Oracle should account for that and properly display `cmc 5` in the example card above as well.
Or even `5 (2 // 3)`

tooomm:
#3090 fixes the first part, and improves the second...

But searching for `Mana Cost = 1U AND 2B` still won't find Far // Away.

For mana cost each "sub-card" should be treated on its on, no combination of all single values needed as with cmc I think.

PR #3090, opened 2018-02-07
Split card handling added to cmc filter

### LINKED ISSUE #3071: cmc filter for split/aftermath cards not working
<b>Info:</b>
Client Version: 2.5.0-custom(3372007) (2018-01-31)
Build Architecture: 64-bit
Qt Version: 5.9.3

__________

The filter for cmc in the deck editor isn't working for split and aftermath cards.
No matter if you filter for `4`, `6` or `10` cmc... [Commit // Memory](https://scryfall.com/card/akh/211) isn't found.

Vafthrudnir:
## Related Ticket(s)
- Fixes #3071 

CMC filter didn't work on split cards in the deck builder

## Screenshots
split_card_filter

### DIFF
--- a/cockatrice/src/filtertree.cpp
+++ b/cockatrice/src/filtertree.cpp
@@ -231,7 +231,21 @@ bool FilterItem::acceptManaCost(const CardInfoPtr info) const
 
 bool FilterItem::acceptCmc(const CardInfoPtr info) const
 {
-    return relationCheck(info->getCmc().toInt());
+    bool convertSuccess;
+    int cmcInt = info->getCmc().toInt(&convertSuccess);
+    if (!convertSuccess) {
+        int cmcSum = 0;
+        foreach (QString cmc, info->getCmc().split("//")) {
+            cmcInt = cmc.toInt();
+            cmcSum += cmcInt;
+            if (relationCheck(cmcInt)) {
+                return true;
+            }
+        }
+        return (cmcSum > 1 ? relationCheck(cmcSum) : false);
+    } else {
+        return relationCheck(cmcInt);
+    }
 }
 
 bool FilterItem::acceptPower(const CardInfoPtr info) const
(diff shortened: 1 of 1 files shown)


tooomm:
I found a very similar issue than #3071 --> #2546

Will test both later today.

Vafthrudnir:
Thanks @Daenyth !
Rarity filter still doesn't work on split cards, because rarity isn't saved for them to the cards.xml for some reason. The rest seemed to be ok.

cockatrice/src/filtertree.cpp
```diff
@@ -231,7 +231,22 @@ bool FilterItem::acceptManaCost(const CardInfoPtr info) const
 
 bool FilterItem::acceptCmc(const CardInfoPtr info) const
 {
-    return relationCheck(info->getCmc().toInt());
+    bool convertSuccess;
+    int cmcInt = info->getCmc().toInt(&convertSuccess);
+    // if conversion failed, check for the "//" separator used in split cards
+    if (!convertSuccess) {
+        int cmcSum = 0;
+        foreach (QString cmc, info->getCmc().split("//")) {
```
ZeldaZach: can you change this to `for(QString cmc : info->getCMC().split("//")) {` ? I'm trying to remove the foreach from the codebase as it's not proper C++

Vafthrudnir: sure

tooomm:
>Rarity filter still doesn't work on split cards, because rarity isn't saved for them to the cards.xml for some reason. The rest seemed to be ok.

Good that you mention something like that, I guess we were not aware of that as of now. Thanks :)
Issue created:  #3091

cockatrice/src/filtertree.cpp
Daenyth: We should see if we can add a custom lint rule to codacy

PR #3095, opened 2018-02-08
Add custom QStyle class to tab bar to fix render issue on macOS. 

### LINKED ISSUE #3070: Mac Style Issues with Qt >= 5.8.0
With recent changes to QT5, they have updated several style components. These changes make our tabs appear wonky and have extra space on them (See Deck, Server, Magic). We'll need to see how we can be more compatible with newer QT versions so we can advance the Mac builds. In addition, Qt5.7 was dropped by Homebrew so we'll have to start looking into being compatible with head (5.10) asap.

QT 5.7.1:
screenshot 2018-02-04 02 52 53

QT 5.10.0:
screenshot 2018-02-04 02 54 09

NickBeeuwsaert:
This fixes a rendering issue on macOS with qt5.10
## Related Ticket(s)
- Fixes #3070

### DIFF
--- a/cockatrice/src/tab_supervisor.cpp
+++ b/cockatrice/src/tab_supervisor.cpp
@@ -30,8 +30,7 @@
 
 QRect MacOSTabFixStyle::subElementRect(SubElement element, const QStyleOption *option, const QWidget *widget) const
 {
-    if(element != SE_TabBarTabText)
-    {
+    if (element != SE_TabBarTabText) {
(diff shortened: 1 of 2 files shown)


tooomm:
Looks like there is something wrong here.
The CI can't find the branch :)

ctrlaltca:
Yep, it looks like the second commit is confusing CI. Anyway, the change is working and safe.

tooomm:
So now is "only" #2647 left...

PR #3097, opened 2018-02-08
Hotfix/2950 delete filters with keyboard

### LINKED ISSUE #2950: Deck builder should delete filters by pressing delete

Vafthrudnir:
## Related Ticket(s)
- Fixes #2950 

`Del` key didn't clear the selected filter in the deckbuilder

- "Remove card" default shortcut removed from `Del`
- Setting up shortcuts to some important keys is made impossible
- Shortcut key validity is checked on startup as well: users won't be able to insert such invalid keys from a tampered `shortcuts.ini` file either

### DIFF
--- a/cockatrice/src/shortcutssettings.cpp
+++ b/cockatrice/src/shortcutssettings.cpp
@@ -17,14 +17,24 @@ ShortcutsSettings::ShortcutsSettings(QString settingsPath, QObject *parent) : QO
     if (exists) {
         shortCutsFile.beginGroup("Custom");
         const QStringList customKeys = shortCutsFile.allKeys();
-
+        QStringList invalidKeys = QStringList();
         for (QStringList::const_iterator it = customKeys.constBegin(); it != customKeys.constEnd(); ++it) {
             QString stringSequence = shortCutsFile.value(*it).toString();
-            QList<QKeySequence> SequenceList = parseSequenceString(stringSequence);
-            shortCuts.insert(*it, SequenceList);
+            // check whether shortcut is valid: neither forbidden keys nor used elsewhere
+            if (isValid(*it, stringSequence)) {
+                QList<QKeySequence> SequenceList = parseSequenceString(stringSequence);
+                shortCuts.insert(*it, SequenceList);
+            }
+            else {
+                invalidKeys.append(*it);
+            }
         }
-
         shortCutsFile.endGroup();
+
+        // set default shortcut where stored value was invalid
+        for (QString key : invalidKeys) {
+            setShortcuts(key, getDefaultShortcutString(key));
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


cockatrice/src/shortcutssettings.cpp
```diff
@@ -114,8 +113,15 @@ bool ShortcutsSettings::isValid(QString name, QString Sequences)
...
-        << "Return" << "Enter" << "Backspace" << "Esc");
+    QStringList forbiddenKeys = (QStringList() << "Del"
+                                               << "Down"
+                                               << "Up"
+                                               << "Left"
+                                               << "Right"
+                                               << "Return"
+                                               << "Enter"
+                                               << "Backspace"
+                                               << "Esc");
```
tooomm: What about `CapsLock` and `Menu` (or application) keys on windows?
No idea about mac...

Vafthrudnir: Good idea, haven't thought about those!
Could we make a list of all such keys so that I could insert them in one commit?

skwerlman:
Is the user warned when they try to set an invalid key for a shortcut? I think explaining to the user why they can't use certain keys is important.

When invalid shortcut keys are detected on startup, is the user told what's happening, or are they silently unbound?

cockatrice/src/shortcutssettings.cpp
skwerlman: Numlock and scrolllock should also be banned imo

tooomm: `Print` and `Pause` ?

Vafthrudnir:
When it's detected on startup it's happening silently - but you are right, a message window should pop up to inform the user.

When the user tries to set up an invalid key, the same message appears as when they try to setup a shortcut that is already used elsewhere: "Shortcut already in use". I can modify it to "Shortcut is invalid or already in use" message if you'd like.

skwerlman:
I think it definitely deserves its own message. "Shortcut already in use" feels like it would confuse a lot of people.

cockatrice/src/shortcutssettings.cpp
Vafthrudnir: If the user would like to use numlock or pause why should we forbid it?
I think this ban should only include keys that actually break the software in some way, e.g. shortcut on arrow keys would make tree navigation impossible with keyboard; `menu` key would pop up a menu whenever used; `del` would (and did!) prevent emitting some signals, etc.
Other keys that are just simply inconvenient should be fair game in my opinion.

Vafthrudnir:
Yeah, you are right. It will need some code refactoring, but I can see the advantage of the separate message. I will get to it later tonight.

ctrlaltca:
Space: it's supposed to press the focused button
Tab: it's supposed to switch between focusable widgets
Should we avoid also common combinations like ctrl+x, ctrl+c, ctrl+v? They have a special meaning right now in the deck editor, but i in a room/game they are useful to move text around.

Vafthrudnir:
Space and Tab should definitely be banned.
Although ctrl+c,v,x shouldn't be a problem: when a text field is in focus these combinations seem to have their text moving function as priority above their shortcut function. At least on Windows, I'm not sure about Linux or Mac.

Vafthrudnir:
- Message has been modified when a user tries to add an invalid shortcut
- Warning messagebox have been created that pops up when the saved data contain any not allowed shortcuts
- Since this problem only occurs in deck-builder window, the validity-check have been limited to those shortcuts related to it
- List of invalid keys have been expanded to everything every combination handled in `keysignals.cpp`

PR #3098, opened 2018-02-08
Mana cost filtering for split cards

### LINKED ISSUE #2546: Improve filter for split cards

Vafthrudnir:
## Related Ticket(s)
- Fixes #2546 

Mana cost filter didn't work for split cards

## Screenshots
split_card_filter_mana_cost

### DIFF
--- a/cockatrice/src/filtertree.cpp
+++ b/cockatrice/src/filtertree.cpp
@@ -219,14 +219,21 @@ bool FilterItem::acceptSet(const CardInfoPtr info) const
 bool FilterItem::acceptManaCost(const CardInfoPtr info) const
 {
     QString partialCost = term.toUpper();
-    QString fullManaCost = info->getManaCost();
 
-    // Sort the mana costs so they can be easily compared
+    // Sort the mana cost so it will be easy to find
     std::sort(partialCost.begin(), partialCost.end());
-    std::sort(fullManaCost.begin(), fullManaCost.end());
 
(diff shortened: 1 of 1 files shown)


Vafthrudnir:
I think it could be sped up significantly by storing whatever properties are possible as integers instead of strings (converted mana cost, loyalty, power/toughness). It must take a lot of performance to search substrings in a string for each card with each new filter - multiple times in some cases.

ZeldaZach:
We can look into a change like that in the near future hopefully :)

I"ll open a ticket about it

Issue #3094, opened 2018-02-07
Allow to rotate cards 180° (e.g. Kamigawa flip cards)


tooomm:
This is a split-off from #3033

Since parts of that issue are dependent on some other conceptional changes, this is the simple part:

> - [ ] Have a `Rotate card` option as new feature in the right click menu of cards
`Flip` is "rotate 180°" in Magic terms, but we use it for "turn over" in our interface
I wouldn't mirror that special mtg terminology, therefore add a `Rotate card 180°` to the UI to allow to play and represent e.g. "Kamigawa flip cards" properly on the battlefield
>
> - [x] Derived from the above: rename `Flip` to `Turn over` in right click menu of cards
That would be more general and has no confusion with flip cards from mtg (which do not turn over, but rotate 180°)
*Done in ~~#3100~~ #3149*

For the deck editor we could think about another solution to make such "flip cards" better readable there too, like right clicking the card preview picture and selecting `Rotate picture 180°` etc.

<br>
I'm open to any suggestions on terminology!

Daenyth:
I kind of don't like this, but I'm not really sure what the right thing is at all.

tooomm:
What don't you like exactly @Daenyth?

Renaming `Flip`?
Allow to `Rotate cards 180°`? (in addition to "Rotate 90°" aka "Tap")

Daenyth:
I dislike "rotate 180" because nobody says that. Rotate generally means 90. Flip usually means physically turn over the flat edge (like transform).

Really this is wotc's fault for using "flip" to mean rotate 180 :P

tbh there's no good solution here. I think the least bad is to leave "flip" for kami flipping and "turn over" for face up/down

tooomm:
>tbh there's no good solution here. I think the least bad is to leave "flip" for kami flipping and "turn over" for face up/down

Ok, `Turn over` it is! --> #3100

About recycling the term "flip" for a new "rotate 180" feature in the future, I'm not 100% sold yet.
There is unlikely a big confusion for players because we have the name change now and the new feature in like 2 releases maybe... so people are used to that by then.
But looking up more words and thesauri we might have other options there as well.

Would do you think about these?
`Flip around`, `Flip upside down`, `Turn around`, `Turn upside down` `Spin`, `Spin around`, `Upend`, `Reverse`, `Invert`, `Tip`...
It looks like explanations on the web tend to `Upend`. It might be other context, but never heard that before.

[...]

PR #3100, opened 2018-02-09
Rename 'flip'  to 'turn over'

tooomm:
## Related Ticket(s)
- Related to #3094

`Flip` was used as a potentially confusing term for "turning a card face up/down"

- Use `Turn over` instead


cockatrice/src/player.cpp
```diff
@@ -733,7 +733,7 @@ void Player::setShortcutsActive()
     aPlay->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aPlay"));
     aTap->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aTap"));
     aDoesntUntap->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aDoesntUntap"));
-    aFlip->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aFlip"));
+    aTurnOver->setShortcuts(settingsCache->shortcuts().getShortcut("Player/aTurnOver"));
```
tooomm: This string looks weird to me. What is exactly used for?

ctrlaltca: Settings are saved to an ini file; that's the "key" used to identify that specific setting in the file, eg:.
```
[Custom]
...
Player\aDrawCards=Ctrl+E
```

cockatrice/src/player.cpp
```diff
@@ -682,9 +682,11 @@ void Player::retranslateUi()
     aPlay->setText(tr("&Play"));
     aHide->setText(tr("&Hide"));
     aPlayFacedown->setText(tr("Play &Face Down"));
+    //: Turn sideways or back again
     aTap->setText(tr("&Tap / Untap"));
     aDoesntUntap->setText(tr("Toggle &normal untapping"));
-    aFlip->setText(tr("&Flip"));
+    //: Turn face up/face down
+    aTurnOver->setText(tr("&Turn Over"));
```
ZeldaZach: This won't work; We have 2 &T's now. I honestly dislike the windows automated shortcuts as they cause translations to be a PITA

tooomm: Good catch! Will change that.

Is it windows only? I think it's a general way to access menus on all platforms?

But wait... does this work for right click menus at all?
For the main menus at top sure, no problem. But here I didn't get it to work in a local test game.

Any ideas? Could we remove it from all right click menus?

tooomm:
>Is this going to break server protocol? 

I tested it live on the server (offline it works), you are right! @ZeldaZach. 😞 

The renamed `.proto` files are the problem I assume?

I can change that back for now, but at one point we need to gather such updates and the servers need a new feature requirement. Otherwise the code will fall apart more and more.
Or do we opt for the string change only? But still, that makes it difficult for new contributors...

common/pb/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ SET(PROTO_FILES
     command_delete_arrow.proto
     command_draw_cards.proto
     command_dump_zone.proto
-    command_flip_card.proto
+    command_turn_card_over.proto
```
Daenyth: Don't change the proto

tooomm: So do we keep track of such left out changes and do them at one point combined?

In addition to requesting the server admins to require that "feature" when connecting to their servers?
Or bumping Cockatrice to v3 as incompatible change?

common/pb/game_commands.proto
```diff
@@ -9,7 +9,7 @@ message GameCommand {
         ROLL_DIE = 1005;
         DRAW_CARDS = 1006;
         UNDO_DRAW = 1007;
-        FLIP_CARD = 1008;
+        TURN_CARD_OVER = 1008;
```
tooomm: These proto changes should be save because they are digit coded?

common/pb/CMakeLists.txt
Daenyth: Add a comment here, but otherwise leave all the proto files alone.

This should only be a client-facing string change.

ZeldaZach:
Honestly, changing all parts of the code (internal) is not really necessary. Provided we just change the visible string (external) I feel that is sufficient

ZeldaZach:
Do we have any plans with this PR? I don't fully agree with it and it's going to cause issues between what protocol is and what we think the protocol actually is with the namings

tooomm:
Ok. So https://github.com/Cockatrice/Cockatrice/pull/3100#discussion_r167368554 and https://github.com/Cockatrice/Cockatrice/pull/3100#issuecomment-364630117 it is? UI only change in the client with a code comment for the protocol that we are renaming internally with a v3 protocol rewrite?

I can submit the changes tomorrow.

tooomm:
see #3149

PR #3104, opened 2018-02-10
Tablerow fixed for non-mana generating artifacts

### LINKED ISSUE #2904: tablerow parsing error
>Doubling clicking on treasure map puts it in the land row instead of the noncreature and nonland row.

Looking at my `cards.xml` file I get this:
```
     <card>
            <name>Treasure Map</name>
            <set rarity="Rare" muId="435410" num="250a">XLN</set>
            <related attach="attach">Treasure Cove</related>
            <manacost>2</manacost>
            <cmc>2</cmc>
            <type>Artifact</type>
            <tablerow>0</tablerow>
            <text>{1}, {T}: Scry 1. Put a landmark counter on Treasure Map. Then if there are three or more landmark counters on it, remove those counters, transform Treasure Map, and create three colorless Treasure artifact tokens with &quot;{T}, Sacrifice this artifact: Add one mana of any color to your mana pool.&quot;</text>
      </card>
```

Which is a mistake in Oracle parsing of mtgjson I guess. Tablerow needs to be `1`.

Vafthrudnir:
## Related Ticket(s)
- Fixes #2904 

As discussed in the related topic, the quoted parts of artifact texts are not taken into consideration when deciding whether they are mana-producing artifacts or not.
The discussion mentioned three such artifacts, but in fact only Treasure Map is affected, because the other two are equipments. This makes the `if (cardType.endsWith("Artifact"))` condition fail on them, so they get `tableRow=1` by default - even without this modification.
See attached shortcut for diff between cards.xml generated with and without the modification.

## Screenshots
xml_diff

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -84,10 +84,14 @@ CardInfoPtr OracleImporter::addCard(const QString &setName,
 
         // detect mana generator artifacts
         bool mArtifact = false;
-        if (cardType.endsWith("Artifact"))
-            for (int i = 0; i < cardTextRows.size(); ++i)
-                if (cardTextRows[i].contains("{T}") && cardTextRows[i].contains("to your mana pool"))
+        if (cardType.endsWith("Artifact")) {
+            for (int i = 0; i < cardTextRows.size(); ++i) {
+                cardTextRows[i].remove(QRegularExpression("\\\".*?\\\""));
+                if (cardTextRows[i].contains("{T}") && cardTextRows[i].contains("to your mana pool")) {
                     mArtifact = true;
+                }
+            }
+        }
 
         // detect cards that enter the field tapped
         bool cipt =


oracle/src/oracleimporter.cpp
```diff
@@ -84,10 +84,14 @@ CardInfoPtr OracleImporter::addCard(const QString &setName,
 
         // detect mana generator artifacts
         bool mArtifact = false;
-        if (cardType.endsWith("Artifact"))
-            for (int i = 0; i < cardTextRows.size(); ++i)
-                if (cardTextRows[i].contains("{T}") && cardTextRows[i].contains("to your mana pool"))
+        if (cardType.endsWith("Artifact")) {
+            for (int i = 0; i < cardTextRows.size(); ++i) {
+                cardTextRows[i].remove(QRegularExpression("\\\".*?\\\""));
```
Nightfirecat: Isn't `.*?` redundant? Shouldn't it just be `.*`?

ZeldaZach: `.*?` means match any character any number of times, whereas `.*` means to match any character once... Not sure which is better here

Nightfirecat:
Oh, duh--been a while since I last regex'd. You're right @Daenyth, my bad.

PR #3106, opened 2018-02-11
Sending card name to chat on shift+click

### LINKED ISSUE #2629: Ctrl + left-click on a card should make its name appear in the chat
This was a feature previously present on Magic Workstation, an old software in which I used to play Magic, and it was very useful for targeting and naming cards. I think this even worked for players (you clicked on the avatar of a player and its name appeared in the chat). Where I believe this feature would be mostly useful is when targeting multiple cards (instead of having to drag an arrow several times) and searching an opponent's library. Because we can no longer drag cards from an opponent's library to the field like we used to, if we need to choose several cards because of an effect, it gets both easier and faster for us to tell that opponent what cards we are choosing.

Vafthrudnir:
## Related Ticket(s)
- Fixes #2629 

When a user shift+left clicks a card, its name will be added to his/her chat text field, enclosed by double brackets.
The issue asks for ctrl+click, but that combination is already used for selecting multiple cards.

### DIFF
--- a/cockatrice/src/abstractcarditem.cpp
+++ b/cockatrice/src/abstractcarditem.cpp
@@ -271,7 +271,9 @@ void AbstractCardItem::setFaceDown(bool _facedown)
 
 void AbstractCardItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
 {
-    if ((event->modifiers() & Qt::ControlModifier)) {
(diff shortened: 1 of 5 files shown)


Vafthrudnir:
Sorry, I haven't knew single click playing was an option - handled it in the new commit.
Although this brought up a new problem: shift+double click (or single click in this case) is already used to play a card face down.
Had to change the card linking combo to alt+click instead.

PR #3115, opened 2018-02-15
"Show related card" option added to ingame context menu

### LINKED ISSUE #1993: Reverse Side of Dual Face Card (DFC) name to xml
Currently, unless a card is on the battlefield/stack, there's no way to know what the name of the back side of the card is.  Add the name of the back/front of DFC to cards.xml. (eg Huntmaster says somewhere in the card Description in deck editor/chat/hand "Ravager of the Fells" and ravager says in the description "Huntmaster of the Fells"

Vafthrudnir:
## Related Ticket(s)
- Fixes #1993 

There was no way to know the details of a card's reverse side without actually creating that card in-game.

- "Show Card" option added to in-game right click menus in all zones and pop-up card lists (e.g. on top cards of library view)
- Each related card is a menu new menu item, when clicked it updates the card info widget with its details
- Context menu on cards without related cards haven't changed

## Screenshots
screenshot 2018-02-17 02 43 38
screenshot 2018-02-17 02 43 24
screenshot 2018-02-17 02 43 17
screenshot 2018-02-17 02 43 09

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -25,6 +25,7 @@
 #include <QMenu>
 #include <QPainter>
 #include <QRegExp>
+#include <QSignalMapper>
 
 #include "pb/command_attach_card.pb.h"
 #include "pb/command_change_zone_properties.pb.h"
@@ -2685,6 +2686,27 @@ void Player::updateCardMenu(const CardItem *card)
             cardMenu->addAction(aClone);
         }
     }
+    addRelatedCardView(card, cardMenu);
+}
+
+void Player::addRelatedCardView(const CardItem *card, QMenu *cardMenu)
+{
+    if (card == nullptr || cardMenu == nullptr || card->getInfo() == nullptr)
+        return;
+
+    QList<CardRelation *> relatedCards = card->getInfo()->getRelatedCards();
+    if (relatedCards.isEmpty()) {
+        return;
+    }
+
+    cardMenu->addSeparator();
+    auto *signalMapper = new QSignalMapper(this);
+    for (const CardRelation *relatedCard : relatedCards) {
+        QAction *viewCard = cardMenu->addAction("Show card: \"" + relatedCard->getName() + "\"");
+        connect(viewCard, SIGNAL(triggered()), signalMapper, SLOT(map()));
+        signalMapper->setMapping(viewCard, relatedCard->getName());
+    }
+    connect(signalMapper, SIGNAL(mapped(const QString &)), game, SLOT(viewCardInfo(const QString &)));
 }
 
 void Player::addRelatedCardActions(const CardItem *card, QMenu *cardMenu)
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2676,7 +2676,7 @@ void Player::updateCardMenu(const CardItem *card)
                 cardMenu->addAction(aClone);
                 cardMenu->addMenu(moveMenu);
             } else {
-                // Card is in hand
+                // Card is in hand or a custom zone specified by server
                 cardMenu->addAction(aPlay);
                 cardMenu->addAction(aPlayFacedown);
                 cardMenu->addMenu(moveMenu);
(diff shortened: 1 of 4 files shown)


Daenyth:
I'm concerned that "Show card" might be misleading to users. It's very similar to "reveal card"

Vafthrudnir:
Yeah, that really may be so. Do you have a suggestion what to rename it to?

Daenyth:
"Display card info"? Maybe? @ZeldaZach @tooomm thoughts?

cockatrice/src/player.cpp
```diff
@@ -2685,6 +2686,28 @@ void Player::updateCardMenu(const CardItem *card)
...
+
+    QList<CardRelation *> relatedCards = card->getInfo()->getRelatedCards();
+    if (relatedCards.isEmpty()) {
+        return;
+    }
+
+    cardMenu->addSeparator();
+    auto *signalMapper = new QSignalMapper(this);
+    for (const CardRelation *relatedCard : relatedCards) {
+        QAction *viewCard = cardMenu->addAction("Show Card: \"" + relatedCard->getName() + "\"");
```
ZeldaZach: `...->addAction(tr("View Card") + ":\"" + relatedCard->getName() + "\"");`

cockatrice/src/player.cpp
```diff
@@ -2685,6 +2686,28 @@ void Player::updateCardMenu(const CardItem *card)
             cardMenu->addAction(aClone);
         }
     }
+    addRelatedCardView(card, cardMenu);
```
ZeldaZach: While this location is good for cards in hand, it's really inconvenient for cards in play. Either we should make this feature only work in hand, or we should place it somewhere better when the card's in play. Similar issues on the stack.

screenshot 2018-02-16 02 35 21
screenshot 2018-02-16 02 35 25
screenshot 2018-02-16 02 35 51

cockatrice/src/player.cpp
Daenyth: Note especially the `tr`, that's important so the string gets translated for other languages

cockatrice/src/player.cpp
Daenyth: For cards in play I'd actually put it near the token menu, maybe above or below

ZeldaZach: My thought was to `view related card images >` submenu it

Vafthrudnir:
I left the option at the end of the context menu, but changed its name ("View related card info:") and added it as a submenu. Also added the `tr()` call.

cockatrice/src/player.cpp
```diff
@@ -2612,6 +2612,8 @@ void Player::updateCardMenu(const CardItem *card)
 
         if (card->getZone()) {
             if (card->getZone()->getName() == "table") {
+                // Card is on the battlefield
```
Daenyth: I'm not sure this comment adds any clarity; the line above says the same thing.

cockatrice/src/player.cpp
```diff
@@ -2657,22 +2660,27 @@ void Player::updateCardMenu(const CardItem *card)
...
                 addRelatedCardActions(card, cardMenu);
             } else if (card->getZone()->getName() == "rfg" || card->getZone()->getName() == "grave") {
+                // Card is in the graveyard or exile
                 cardMenu->addAction(aPlay);
                 cardMenu->addAction(aPlayFacedown);
                 cardMenu->addSeparator();
                 cardMenu->addAction(aClone);
                 cardMenu->addMenu(moveMenu);
             } else {
+                // Card is in hand
```
Daenyth: in hand or any custom zone that the server is using which the client isn't aware of, would be more accurate.

cockatrice/src/player.cpp
ZeldaZach: Added since I added to all of them

cockatrice/src/player.cpp
ZeldaZach: True, i'll make this change

ZeldaZach:
Screenshots Updated

ZeldaZach:
I see... Can you add it in for that case?

ZeldaZach:
I just have to say, with the tokens, it seems really weird to see the same card twice. Maybe we can move the create tokens to a sub dialog? What do y'all think?

Daenyth:
We probably shouldn't nest the menu too deeply, since that makes it more
annoying to use, and related cards are such a common interaction

Issue #3063, opened 2018-01-31
Add "Get Started" guide (widget) to the default UI


tooomm:
The idea is to have a new, additional widget which is displayed on default for clean installs.
On later runs the user could just close the widget because he most likely never needs it again. But if there is more help required one could display it again like all other widgets: `View` --> `Get Started` --> `visible`
This way veteran players are not annoyed at all while we can provide a better user experience to new players.

It should contain several most basic information and hints to guide a new, tech-unsavy person with no background knowledge about such programs to a pleasurable first run experience and a not complicated and intuitive setup until they can play their very first game connected to a server.

This ticket is about discussions on the idea itself, which information is needed/requiered and in which form it should be presented.

Suggestions and thoughts are highly welcome!
If you are a user, let us know which problems you run into at your first start of the app. What are your experiences?
If you are a moderator/server admin, which issues did new users report to you, or constantly ask questions about?
*It could be reasonable to open a thread at reddit as well to gather more feedback!*

---

I created a mockup how it could look like:
![first_start_view_with_help_widget_shown](https://user-images.githubusercontent.com/9874850/35627242-98de0adc-0698-11e8-8007-6e4be7b124de.png)

A first simple version could be text only with links to our wiki maybe.
And a later one can provide buttons that open `Oracle` or the `Register` dialog etc.

Daenyth:
I actually would put the "get started" guide on the otherwise-blank pre-connect screen. 

One of the reason I've been so on the idea of #652 is because it's low work for us and non-intrusive for folks who don't want it.

This is a good idea though!

tooomm:
>the otherwise-blank pre-connect screen

What's do you mean exactly?

ZeldaZach:
@Daenyth Maybe i'll think of something to build on TOTH

Daenyth:
By pre-connect screen I mean when you install cockatrice and open the app we just have a big grey empty window. You need to go to the menu and select "connect".

tooomm:
>By pre-connect screen I mean when you install cockatrice and open the app we just have a big grey empty window. 

I guess things changed in between, I get this very clumsy view on a clean first run with a popup as a first attempt to help new players getting the cards:
![first-run](https://user-images.githubusercontent.com/9874850/35650493-90d5dd08-06dc-11e8-92f7-0b30c1b0af68.png)


It won't even start in full screen... it looks randomly placed at the left.
[...]

Issue #652, opened 2015-01-31
Implement "Tip of the day" style feature


Daenyth:
One of the big UI problems we have is that we have a lot of features that aren't very discoverable. Something like intellij's tip of the day feature would help people find out about them
- Card arrows
- Buddy list
- `[[card]]`, `[card]`, and `[url]`
- Handling bottom deck cards
- How to play face down
- Confusion about when other player can see cardnames - drawing card, etc
- What the user shirts mean

Example: http://www.javaspecialists.eu/courses/concurrency/setup/idea12.png

Daenyth:
We can have a list of tips, a seen flag in the config (tips/seenCardArrowTip -> true) and cycle through them. Most useful/important tips first

poixen:
I like this. Especially since we are adding a lot of features now.

ZeldaZach:
I'm a fan, and users can opt out of if they'd like (at which points mods should just refuse their dumb questions relating to something we explain)

woogerboy21:
Were would the list of items to display to the user be pulled from?

ZeldaZach:
Server operator's Database

woogerboy21:
And pulled down to the client?  Or is it something presented to the user on connect?

ZeldaZach:
Pulled and presented after the user logs into the server

poixen:
You mean like a popup? That would get annoying real fast.

ZeldaZach:
Disable-able at any point with a little checkbox on the physical popup

poixen:
I would still want to get message of the day. Maybe it would be better built into the client. Maybe under the user list in each room? Or in the server tab.

Daenyth:
They would be hard coded in the client. That way we can use it to message
about new features. This doesn't need to have any server side knowledge

Daenyth:
I'll grab some screenshots of a version other software uses

woogerboy21:
I was thinking you wanted something client side, but the problem with that idea is the list of items would never change to display to the user till they upgraded there client.

Daenyth:
That's kind of the point though. It should inform them about what they can
do. The update alerting is a separate thing.

Daenyth:
<img width="558" alt="screen shot 2015-09-15 at 7 59 14 am" src="https://cloud.githubusercontent.com/assets/14644/9875869/b28681f6-5b7f-11e5-9f6f-65752d641fec.png">

[...]

PR #3116, opened 2018-02-15
left, right, and S shortcuts modified to use shift

### LINKED ISSUE #2937: Use Shift for deck builder shortcuts (Shift + S, Shift + arrows)
There is conflicting functionality in Cockatrice's deck builder.

Using the left arrowkey on a card will decrease the amount of that card by 1, but focus will be moved to the parent element. So if you use the left arrowkey twice, you will decrease the amount of the selected card and then close the list it's in. Using the right arrowkey increases the amount of a card by 1 without inconsistent effects.

Using S will move a card to or from the sideboard. If focus is on the deck list, you can start typing a card name in your deck to select it, which is very useful. However, every time you press S the current card will switch to/from sideboard, which ruins this feature.

Both of these inconsistent behaviors can be resolved by requiring Shift to be held. Shift + S to sideboard cards and Shift + left arrow or Shift + right arrow to decrease or increase the quantity of a card by 1; respectively.

**Additional issue:** Sometimes, using left arrow to reduce the only card in a category to 0 quantity will hide that card and category instead of removing it. Even adding another card to that category won't reveal the hidden card, and it's not reflected in the number of cards in the deck. It gets saved with the deck and appears when reloading the deck or playing. I'm not sure if this is related.

Vafthrudnir:
## Related Ticket(s)
- Fixes #2937 

'S', 'Left', and 'Right' keys were set as shortcuts in the decbuilder which conflicted with these keys' tree navigation purpose.

Modified to shortcuts to use respectively: 'Shift+S', 'Shift+Left', and 'Shift+Right' key combinations.

### DIFF
--- a/cockatrice/src/keysignals.cpp
+++ b/cockatrice/src/keysignals.cpp
@@ -21,11 +21,13 @@ bool KeySignals::eventFilter(QObject * /*object*/, QEvent *event)
 
             break;
         case Qt::Key_Right:
-            emit onRight();
+            if (kevent->modifiers() & Qt::ShiftModifier)
+                emit onShiftRight();
 
             break;
         case Qt::Key_Left:
-            emit onLeft();
+            if (kevent->modifiers() & Qt::ShiftModifier)
+                emit onShiftLeft();
 
             break;
         case Qt::Key_Delete:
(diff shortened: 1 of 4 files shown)


Vafthrudnir:
Some programs pop up a "What's changed?" window after every update, maybe we could do something similar.

ZeldaZach:
@Vafthrudnir Would be nice for that Tip of the Day thing. Related: #652 #3063

Daenyth:
I think in the past we've done some ad-hoc one-time-only alerts. We don't have a good solution in general though

ZeldaZach:
With #3118 I'm fine with merging this now.

PR #3118, opened 2018-02-19
Tip of the Day

### LINKED ISSUE #652: Implement "Tip of the day" style feature

Vafthrudnir:
## Related Ticket(s)
- Fixes #652

- Tip of the day feature added, which pops up once with each start.
- It can be enabled/disabled from the options menu
- Tips can be added with the `tips_of_the_day.xml` file, which is located in the `cockatrice/resources/tips` folder
- An image may be attached to each tip, which must be located in the `cockatrice/resources/tips/images` subfolder - these files also have to be added to the `cockatrice.qrc` file as new resources
- The tips are shown in order (based on the xml), starting with the first one. The program loads the last shown tip on start, and displays the next in order. 
- The `date` value in the xml is unused as of yet
- The xml is filled with dummy data and the attached screenshots are just for test purposes as well
- The window is resizable

## Screenshots
new_tip

screenshot 2018-02-19 01 18 12


cockatrice/cockatrice.qrc
```diff
@@ -341,6 +341,9 @@
         <file>resources/userlevels/admin_buddy.svg</file>
         <file>resources/userlevels/admin_vip.svg</file>
         <file>resources/userlevels/admin_vip_buddy.svg</file>
-
+        
+        <file>resources/tips/tips_of_the_day.xml</file>
+        <file>resources/tips/images/context_in_hand.png</file>
```
ZeldaZach: Ok so this image has to be removed asap. We cannot store any copyright information within the code base. Good example, just make sure to black out certain things :)

cockatrice/CMakeLists.txt
```diff
@@ -80,6 +81,7 @@ SET(cockatrice_SOURCES
     src/tab_userlists.cpp 
     src/tab_deck_editor.cpp
     src/tab_logs.cpp
+    src/tip_of_the_day.cpp
```
ZeldaZach: Can we put the two tip lines together in the CMake file? Just so they're a bit more organized

cockatrice/resources/tips/tips_of_the_day.xml
```diff
@@ -0,0 +1,14 @@
+<tips>
+	<tip>
+		<title>Tip #1</title>
```
ZeldaZach: Spaces > Tabs (4x space per tab)

cockatrice/src/dlg_tip_of_the_day.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <QPushButton>
+
+#include "dlg_tip_of_the_day.h"
+#include "settingscache.h"
+#include "tip_of_the_day.h"
+
+DlgTipOfTheDay::DlgTipOfTheDay(QWidget *parent) : QDialog(parent)
+{
+    successfulInit = false;
+    QString xmlPath = "theme:tips/tips_of_the_day.xml";
```
ZeldaZach: How does this path work if we're building with portable mode?

cockatrice/cockatrice.qrc
skwerlman: removing this should be done via a force push, since simply deleting it would still include it in history.

ZeldaZach: We'll be squashing anyway, so it shouldn't matter once merged

skwerlman: ah, i forgot about squash on merge haha

Nightfirecat:
I've often seen the order of those buttons be <kbd>Prev</kbd> <kbd>Next</kbd> <kbd>Ok</kbd> in other applications. Is it common for them to be ordered the way they are in your screenshots?

ZeldaZach:
@Nightfirecat I've uploaded the win/mac pics to the main PR.

Clion's TOTD looks like this
screenshot 2018-02-19 01 20 03

Vafthrudnir:
The previous and next buttons should be changed up, that would probably look a bit better

ZeldaZach:
cc: @tooomm about writing tips

cockatrice/src/dlg_tip_of_the_day.cpp
Vafthrudnir: I think it doesn't matter whether it's portable or not, because the xml has been added as resource and those are compiled to binary and get put inside the executable itself.

Daenyth:
I will look at the code as soon as I can but two key points

- I'd like the opt-out to be visible from the popup
- I'd like the first tip shown to be talking about the tip feature so people know what to expect.

ZeldaZach:
@Daenyth  Bullet 1 is solved already. Agreed with the 2nd one

Daenyth:
Excellent thanks. I'll try to get to the code today

PR #3120, opened 2018-02-25
Power/toughness indicator displayed after card flip

### LINKED ISSUE #2929: Flipping creature face-up doesn't show power / toughness
<b>OS: Windows 10 Home</b>
<b>Cockatrice version: 2.3.17 (2017-05-05)</b>
__________

Playing a creature card face down during gameplay, then flipping it face up will not display the power and toughness of that creature. To be clear, you have to play it from your hand face down; the bug does not occur if you play a creature and then flip it face-down.

Vafthrudnir:
## Related Ticket(s)
- Fixes #2929 

When a face down card was flipped on the table the p/t indicator didn't appear.

When a flip card command is received by the server, a set attribute event will also be generated to make the indicator appear.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -2205,6 +2205,11 @@ void Player::cardMenuAction()
                     cmd->set_zone(card->getZone()->getName().toStdString());
                     cmd->set_card_id(card->getId());
                     cmd->set_face_down(!card->getFaceDown());
+                    if (card->getFaceDown()) {
+                        CardInfoPtr ci = card->getInfo();
+                        if (ci)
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Can you possibly attach a few screenshots showing this in action? Not sure what goes on from reading the code

Vafthrudnir:
It's a bit hard to show it on a screenshot, so I'll just list the steps to reproduce the bug:. put down a creature card face-down from your hand, then flip it over. The p/t indicator in the lower right corner of the card will be missing. The PR fixes this bug. It works in either local or normal games.

PR #3122, opened 2018-02-27
clangify only the different files

ZeldaZach:
Speed up clangify by only running on modified files

### DIFF
--- a/clangify.sh
+++ b/clangify.sh
@@ -5,7 +5,8 @@
 set -e
 
 if hash clang-format 2>/dev/null && hash git 2>/dev/null; then
-	git diff --name-only | xargs -I{} find '{}' \( -name "*.cpp" -o -name "*.h" \) \
+	git diff --name-only $(git merge-base origin/master HEAD) | \
+	xargs -I{} find '{}' \( -name "*.cpp" -o -name "*.h" \) \
 		-not -path "./cockatrice/src/qt-json/*" \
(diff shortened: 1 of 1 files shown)


clangify.sh
```diff
@@ -1,11 +1,11 @@
...
 
-# This script will run clang-format on all non-3rd-party C++/Header files.
+# This script will run clang-format on all modified, non-3rd-party C++/Header files.
 
 set -e
 
-if hash clang-format 2>/dev/null; then
-	find . \( -name "*.cpp" -o -name "*.h" \) \
+if hash clang-format 2>/dev/null && hash git 2>/dev/null; then
+	git diff --name-only | xargs -I{} find '{}' \( -name "*.cpp" -o -name "*.h" \) \
```
Daenyth: `git diff --name-only $(git merge-base origin/master)..HEAD` I think

ZeldaZach: Looks to be `git diff --name-only $(git merge-base origin/master HEAD)`

clangify.sh
```diff
@@ -1,18 +1,22 @@
 #!/bin/bash
 
-# This script will run clang-format on all non-3rd-party C++/Header files.
+# This script will run clang-format on all modified, non-3rd-party, C++/Header files.
```
tooomm: It's an enumeration of two adjectives. It should be ok with no or one comma, but not two.

PR #3128, opened 2018-03-05
Colors on split cards sorted in WUBRG order

### LINKED ISSUE #3127: Colour field of split cards is not in WUBRG order in
<b>Cockatrice version:</b>
2.5.0
__________
Colour field should be in WUBRG order, and it random instead.

This is caused by Oracle: the two parts of a split card are stored separately in mtgjson, oracle just concatenates the two and removes duplicates without modifying the order.
Manual sorting should be added after this.

Screenshot

Vafthrudnir:
## Related Ticket(s)
- Fixes #3127

- Oracle will sort the colors of split cards and store them in WUBRG order
- Only the Color(s) field is affected

## Screenshot
image

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -299,6 +299,10 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
         }
 
         colors.removeDuplicates();
+        if (colors.length() > 1) {
+            sortColors(colors);
+        }
+
         // Fortunately, there are no split cards that flip, transform or meld.
         relatedCards = QList<CardRelation *>();
         reverseRelatedCards = QList<CardRelation *>();
@@ -316,6 +320,26 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
     }
 
     return cards;
+}
+
+void OracleImporter::sortColors(QStringList &colors)
+{
+    QMap<QString, unsigned int> colorOrder;
+    colorOrder["W"] = 0;
+    colorOrder["U"] = 1;
+    colorOrder["B"] = 2;
+    colorOrder["R"] = 3;
+    colorOrder["G"] = 4;
+    bool correctOrder = false;
+    while (!correctOrder) {
+        correctOrder = true;
+        for (unsigned int i = 0; i < colors.length() - 1; i++) {
+            if (colorOrder[colors[i]] > colorOrder[colors[i + 1]]) {
+                colors.swap(i, i + 1);
+                correctOrder = false;
+            }
+        }
+    }
 }
 
 int OracleImporter::startImport()
--- a/oracle/src/oracleimporter.h
+++ b/oracle/src/oracleimporter.h
@@ -93,6 +93,7 @@ public:
 
 protected:
     void extractColors(const QStringList &in, QStringList &out);
+    void sortColors(QStringList &colors);
 };
 
 #endif


oracle/src/oracleimporter.cpp
```diff
@@ -318,6 +322,26 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
     return cards;
 }
 
+void OracleImporter::sortColors(QStringList &colors)
+{
+    QMap<QString, unsigned int> colorOrder;
+    colorOrder["W"] = 0;
```
ZeldaZach: Can probably do this as:
`{"W", 0}, {"U", 1}, {"B", 2}, {"R", 3}, {"G", 4}`

Vafthrudnir:
Sure, will change it when I get home.

oracle/src/oracleimporter.cpp
```diff
@@ -318,6 +322,26 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
...
+    colorOrder["U"] = 1;
+    colorOrder["B"] = 2;
+    colorOrder["R"] = 3;
+    colorOrder["G"] = 4;
+    bool correctOrder = false;
+    while (!correctOrder) {
+        correctOrder = true;
+        for (unsigned int i = 0; i < colors.length() - 1; i++) {
```
ZeldaZach: Do we really want to bubble sort this?

oracle/src/oracleimporter.cpp
```diff
@@ -318,6 +322,26 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
...
+            if (colorOrder[colors[i]] > colorOrder[colors[i + 1]]) {
```
ctrlaltca: If i add a new fake color to the xml file (eg: `Z`), this is going to crash.

oracle/src/oracleimporter.cpp
ctrlaltca: Agreed

Vafthrudnir: I can't push it rn, but what about this solution?
```
QStringList sortColors(QStringList &colors) {
	QStringList orderedColors;
	QStringList colorOrder;
	colorOrder << "W" << "U" << "B" << "R" << "G";
	for (auto color : colorOrder) {
		int index = colors.indexOf(color);
		if (index != -1) {
			orderedColors << color;
			colors.removeAt(index);
		}
	}
	if (!colors.isEmpty()) {
		orderedColors.append(colors);
	}
	return orderedColors;
}
```
This was my original idea, but dismissed it because I thought bubble sort would be better when most of the multicoloured cards are maximum 2-3 colours, and sometimes in the correct colour order anyway.
Although I hadn't thought about the possibility of new colours, that would surely crush with the current code.

Daenyth: Doesn't stl have sorting algorithms? Surely there must be a way where we don't have to implement anything more than a `(char, char) -> int` function

Daenyth: https://www.codeproject.com/Articles/38381/STL-Sort-Comparison-Function#cpplambda

Vafthrudnir: Thanks for the link, it's much nicer that way!

oracle/src/oracleimporter.cpp
Vafthrudnir: That shouldn't be a problem, because this whole sorting happens before the xml is created, during the mtgjson processing. The only input we'll ever get in here is the output of the `OracleImporter::extractColors` method, which is guaranteed to be one of the five valid color codes.

oracle/src/oracleimporter.cpp
```diff
@@ -318,6 +322,13 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
     return cards;
 }
 
+void OracleImporter::sortColors(QStringList &colors)
+{
+    const QMap<QString, unsigned int> colorOrder{{"W", 0}, {"U", 1}, {"B", 2}, {"R", 3}, {"G", 4}};
+    std::sort(colors.begin(), colors.end(),
+              [&colorOrder](const QString a, const QString b) { return colorOrder[a] < colorOrder[b]; });
```
ZeldaZach: Does this work if we add a random new color?

Vafthrudnir: No, it will crash. But for that to happen, this new color would have to be added to the mtgjson file itself. 
And since the sorting only operates on the output of the `OracleImporter::extractColors` method, that would also have to be updated to process this new color before it can cause a crash. If you'd like I can add a few lines to check whether the color is among the supported color codes, but I can't really imagine a new color being added either short or long term.

Daenyth: Can you add a fallback just in case? I wouldn't rule out a bug in mtgjson generating bad colors in the future.

Vafthrudnir: Added it.
I was wrong though, it wouldn't have crashed with new colors. Turns out QMap returns a default value (0 in this case) when it encounters an unknown key. This would have made the new color the same priority as "W".

PR #3149, opened 2018-03-23
flip rename (client side)

tooomm:
Rename `flip` to `turn over` (UI only change).

Related to #3094.


cockatrice/src/player.cpp
```diff
@@ -683,9 +683,13 @@ void Player::retranslateUi()
...
     aHide->setText(tr("&Hide"));
     aPlayFacedown->setText(tr("Play &Face Down"));
+    //: Turn sideways or back again
     aTap->setText(tr("&Tap / Untap"));
     aDoesntUntap->setText(tr("Toggle &normal untapping"));
-    aFlip->setText(tr("&Flip"));
+    //: Turn face up/face down
+    aFlip->setText(tr("T&urn Over")); // Only the user facing names in client got renamed to "turn over"
+                                      // All code and proto bits are still unchanged (flip) for compatibility reasons
+                                      // A protocol rewrite with v3 should incorporate that, see #3100
```
Daenyth: `should` is a bit strong. `could` perhaps

PR #3145, opened 2018-03-22
Tip of the Day extra features

Vafthrudnir:
## Related Ticket(s)
- Fixes #3139 

## What will change with this Pull Request?
- Added link to open tip of the day from the help menu
- Option to enable/disable tips moved
- Once all the tips have been seen, the window is not shown again at startup. If new tips are added (and it haven't been disabled by the user) it is shown again, starting with the oldest unseen tip.

## Screenshots
Help menu:
image
Settings window:
image

PR #3152, opened 2018-03-24
Feature/3139 new features for tips

Vafthrudnir:
Rebase of PR #3145

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -93,15 +93,15 @@ GeneralSettingsPage::GeneralSettingsPage()
     personalGrid->addWidget(&pixmapCacheLabel, 2, 0);
     personalGrid->addWidget(&pixmapCacheEdit, 2, 1);
     personalGrid->addWidget(&updateNotificationCheckBox, 3, 0);
-    personalGrid->addWidget(&picDownloadCheckBox, 4, 0, 1, 3);
-    personalGrid->addWidget(&defaultUrlLabel, 5, 0, 1, 1);
-    personalGrid->addWidget(defaultUrlEdit, 5, 1, 1, 1);
-    personalGrid->addWidget(&defaultUrlRestoreButton, 5, 2, 1, 1);
-    personalGrid->addWidget(&fallbackUrlLabel, 6, 0, 1, 1);
-    personalGrid->addWidget(fallbackUrlEdit, 6, 1, 1, 1);
-    personalGrid->addWidget(&fallbackUrlRestoreButton, 6, 2, 1, 1);
-    personalGrid->addWidget(&showTipsOnStartup, 7, 0);
-    personalGrid->addWidget(&urlLinkLabel, 7, 1, 1, 1);
+    personalGrid->addWidget(&showTipsOnStartup, 4, 0);
+    personalGrid->addWidget(&picDownloadCheckBox, 5, 0);
+    personalGrid->addWidget(&urlLinkLabel, 5, 1);
+    personalGrid->addWidget(&defaultUrlLabel, 6, 0, 1, 1);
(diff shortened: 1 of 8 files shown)


Daenyth:
FYI you don't need to make a new branch - if you just force push to the original branch then the PR will update.

Vafthrudnir:
I haven't knew that, thanks!

tooomm:
I like the changes 👍 
- Settings reorder
- New menu item to open tips on demand
- And most important: once you've seen all tips, it won't start over with the old ones

Couldn't test if tips start showing up again if new ones are added.

In the initial issue we discussed how we have to add new tips to make the app realize there are some (and which ones). And how to handle deletion of outdated ones which are no longer valid etc.
There should be a hint in the xml (?) how one has to handle this.

Vafthrudnir:
You are right @tooomm, the last shown tip counting really is buggy.
Though fortunately this `lastShowTip` value is only used after all the tips were seen - in which case I think it is a bit useless. If someone saw all the tips I think this "show the one after the last seen tip" mechanic add no benefit to the user. The window could just open with the first tip and that wouldn't be any worse, because the user already saw all of the tips - why should it matter which they've seen last?

So I'd rather remove this feature altogether, it's only a remnant of the first implementation.

Daenyth:
@Vafthrudnir what happens in the case that we add new tips in a new release? Would they be skipped from the buggy counting? If they have tips enabled but have seen all the ones in v1, and we upgrade to v2, we want to show them the new tips

Vafthrudnir:
If new tips are added they'll always be shown first, this counting bug doesn't affect that.

tooomm:
Now the previous and next buttons are broken for me if I open the tips manually from the menu. 😄

Vafthrudnir:
Yeah, sorry about that - I accidentally removed a line I shouldn't have :/

tooomm:
Buttons are fixed now.

- If you open the tips dialog from the menu, it'll *always* open the tips dialog with `Tip #1`?
- For new users, will it still remember the last seen one and continues with the follow-up on next app start if tips are enabled?
- Same if we introduce new tips. Will it jump right there even if people manually open tips and review some old ones in between?

Vafthrudnir:
> If you open the tips dialog from the menu, it'll always open the tips dialog with `Tip #1`?

That's right, if you've already seen all the tips the window will open with the first tip.

> For new users, will it still remember the last seen one and continues with the follow-up on next app start if tips are enabled?

New users will first see the first tip. All tip indices that were already seen by the user are saved to the config file, and when the window opens again the first unseen tip will be shown.

> Same if we introduce new tips. Will it jump right there even if people manually open tips and review some old ones in between?

If new tips are introduced the first unseen tip will be shown, regardless of which was seen last. Same with manually opening the window: if there are any tips left unseen, those are shown first when the window opens.
If you'd like you can actually try this out by deleting the last few values from the `global.ini` config file `seenTips` key. That would simulate the experience a new user has.

tooomm:
Jep, deleting the key made me test it a bit better.

Another issue is that the `[x]` Show tips on startup` box in the dialog window is always checked.
If you uncheck and say ok, it'll be checked again if you open it again. In Settings it seems to be reflected correct and won't get checked again.

Vafthrudnir:
Yes, the checkbox in the tip window doesn't use the user settings to load its value, it is checked by default. I'll change it to reflect the settings when I get home.
Thanks for the thorough testing by the way!

tooomm:
The checkbox is now correctly mirrored with the option in settings and works as expected.

The tip opening is inconsistent though. Sometimes it counts up, sometimes the same is opened again etc.
Try to play around with several use cases and (re)opening the dialog. I was sure it worked before?!

Vafthrudnir:
@tooomm could you write a test case where this inconsitency can be observed? I tried several different cases, but it works properly at me.

tooomm:
It works for me now too after clearing all ini again. 👍

PR #3163, opened 2018-04-02
add rooster ranges, remove woogerworks

ZeldaZach:
Since WW is closing with the release of RR, I've changed all references over and added a clause to add it to the user's connection list if it does not already exist.

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -171,8 +171,7 @@ void DlgConnect::rebuildComboBoxList()
         settingsCache->servers().addNewServer("Chickatrice", "chickatrice.net", "4747", "", "", false);
         settingsCache->servers().addNewServer("dr4ft", "cockatrice.dr4ft.com", "4747", "", "", false);
         settingsCache->servers().addNewServer("Tetrarch", "mtg.tetrarch.co", "4747", "", "", false);
-    }
-    else if (! savedHostList.contains("Rooster Ranges")) // If Rooster Ranges not in the list
+    } else if (!savedHostList.contains("Rooster Ranges")) // If Rooster Ranges not in the list
     {
         settingsCache->servers().addNewServer("Rooster Ranges", "server.cockatrice.us", "4747", "", "", false);
     }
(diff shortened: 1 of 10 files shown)


Vafthrudnir:
If WW closes for good, shouldn't we remove it from the users' connection options btw?

ZeldaZach:
I wasn’t sure how to remove WW, want to add to this PR? ❤️

Vafthrudnir:
I'm not sure either, and working on an another issue rn. Maybe pull this PR as is, and create a new issue to remember to do it later?

ZeldaZach:
I’ll see what I can come up with tonight

ZeldaZach:
Figured out the remove and have implemented it. Feel free to re-test

tooomm:
1) Are there any more details or background information? @woogerboy21 
It looks like you took the server over and re-branded it, Zach?

2) Is there a backup server etc. for RR, too?

3) With the build from this PR, `Woogerworks` is still in my `Connect` list:
connect list
I added the RR server manually before updating, so no idea if it would have added it correctly.

Looking at https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers @dev-id changed his server, too.
I never realized that in the connect dialog though. In code it's there (https://github.com/Cockatrice/Cockatrice/pull/3163/files#diff-601846fadf10ec28443e4b7013530e18R172), but I can still not select the server from the drop down in 2.5.1-beta as realized in the initial PR: https://github.com/Cockatrice/Cockatrice/pull/3051
Also, his old server got never deleted.

skwerlman:
Why are the default servers hardcoded, and not stored in an XML? If they were in an XML, we could load them, then load known servers overwriting info from the XML in case of conflicts. We could have a field in the XML to mark a default as removed, which would cause the client to hide (but not forget!) settings from that site.
It'd have the same behavior from the user's perspective, but we wouldn't have to worry about all this fiddly stuff when one changes.

Daenyth:
As with most questions about "why" trice does something some way; it's arbitrary - that's how it was written in the first pass and it hasn't changed.

I'd accept a change making it load known servers from xml

Daenyth:
Or json. Please no more xml

ZeldaZach:
Merging so I can work on my other PR

PR #3164, opened 2018-04-02
Untranslated string comparison fixed

### LINKED ISSUE #2139: Impossibile to switch a card from main deck to sideboard by double clicking it in deckeditor
<b>OS:</b>
Win 10

<b>Cockatrice version:</b>
277d7e2 (2016-06-30)

---

<br>

Hi, as in the title me and my other two friends are having this same problem.

Vafthrudnir:
## Related Ticket(s)
- Fixes #2139 

Translated zone name was compared to untranslated "Maindeck" string, which caused errors with card swapping between main- and sidedecks on non-English version of the program.

`data()` calls with `EditRole` of the deckListModel returns now the universally defined `DECK_ZONE_MAIN` and `DECK_ZONE_SIDE` strings instead of language-dependent ones.

### DIFF
--- a/cockatrice/src/decklistmodel.cpp
+++ b/cockatrice/src/decklistmodel.cpp
@@ -106,7 +106,10 @@ QVariant DeckListModel::data(const QModelIndex &index, int role) const
                     case 0:
                         return node->recursiveCount(true);
                     case 1:
-                        return node->getVisibleName();
+                        if (role == Qt::DisplayRole)
+                            return node->getVisibleName();
+                        else
+                            return node->getName();
                     default:
                         return QVariant();
                 }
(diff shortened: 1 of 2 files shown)


ZeldaZach:
Welp... feels bad for introducing that one. Great find, as always!

Vafthrudnir:
Don't feel bad, it was easy to miss. I couldn't find it either until they pointed out that the language change solved the problem.

tooomm:
Nice one!

Xatrith:
Nice :) Is it possible to install the fix on our version of the soft ?

Daenyth:
@ZeldaZach can you cut a beta release so that @Xatrith and others can install?

@Vafthrudnir nice find, thank you!

Issue #2474, opened 2017-03-14
Spec for server status protocol


Daenyth:
From #742, we need a spec for a server's information

Add a new command able to be sent to a server when not logged in
- Propose `Command_GetServerInfo`

A new response container for the useful information
- Propose `Response_ServerInfo_ServerStatus

Response should have fields useful to the UI:
- require reg to connect?
- reg is enabled?
- Current number of online users
- configured max online users

Any further suggestions?

Daenyth:
#2201 becomes possible with this

Daenyth:
- Response can also have a short length-limited (256?) text field for a description of the server
- Server should include required + supported features to connect

Should it include the version of the server?

woogerboy21:
Any concerns with possible server attack vectors on something like this?

skwerlman:
I can't think of a reason to include the server version, but I don't think there's any reason not to.

I think it should include whether the server has a scheduled shutdown pending, so people don't log on only to be told that the server is shutting down right now in the middle of their game. It'd be nice to see how long until the shutdown as well as the reason, but I don't think that part's as important.

I don't think there's any risk of attack here, but IANAE.

ZeldaZach:
Some things I'd like to see this protocol have:

- server status (online, pending reboot, offline [aka no response])
- server requirements (clientId, mod log, registered user, etc)
- can you register on this server (independent of what the requirement field says)
- how many users are online this server
- how many users can this server support

Daenyth:
> Any concerns with possible server attack vectors on something like this?

That was my reasoning for not necessarily including the version. Also the version isn't really helpful to users - can be a custom version or newer or older than client and none of it matters if feature set matches

Daenyth:
Zach I think you just restated all the things I mentioned :P

woogerboy21:
Identifying the scheduled shutdown time remaining would be new to me.  Is there a way to query the amount of time left on a timer?

tooomm:
>#2201 becomes possible with this

Should be the same for #1702 then

>I think it should include whether the server has a scheduled shutdown pending, so people don't log on only to be told that the server is shutting down right now in the middle of their game. It'd be nice to see how long until the shutdown as well as the reason, but I don't think that part's as important.

Well, they get the `scheduled shutdown` message right after login right now I think?
So maybe not needed, still nice to have though...

ZeldaZach:
[...]

PR #3165, opened 2018-04-03
Server List from JSON

### LINKED ISSUE #2226: Discuss: Fetch server list from online

ZeldaZach:
Fixes #2226 

This PR will take the servers from the list on <strike>the wiki</strike> [GitHub Public Servers JSON](https://cockatrice.github.io/public-servers.json) and import them to the client, thus removing the dependency on hardcoded values.

When the user "refreshes the servers", we will re-download the list and clear any server not found on the public list.

screenshot 2018-04-08 23 47 24
screenshot 2018-04-08 23 47 28

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -3,6 +3,7 @@
 #include "userconnection_information.h"
 #include <QCheckBox>
 #include <QComboBox>
+#include <QDebug>
 #include <QDialogButtonBox>
 #include <QEvent>
 #include <QGridLayout>
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -263,7 +263,6 @@ void DlgConnect::preRebuildComboBoxList()
 
 void DlgConnect::rebuildComboBoxList()
 {
-    int prevHostSize = previousHosts->count();
     previousHosts->clear();
 
     UserConnection_Information uci;
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -21,6 +21,7 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
     previousHosts = new QComboBox(this);
     previousHosts->installEventFilter(new DeleteHighlightedItemWhenShiftDelPressedEventFilter);
 
+    // TODO: Remove the button at the 2.7.0 release
     hps = new HandlePublicServers(this);
     btnRefreshServers = new QPushButton(tr("Refresh Servers"), this);
     connect(hps, SIGNAL(sigPublicServersDownloadedSuccessfully()), this, SLOT(rebuildComboBoxList()));
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -26,7 +26,7 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
     hps = new HandlePublicServers(this);
     btnRefreshServers = new QPushButton(tr("Refresh Servers"), this);
     connect(hps, SIGNAL(sigPublicServersDownloadedSuccessfully()), this, SLOT(rebuildComboBoxList()));
-    connect(hps, SIGNAL(sigPublicServersDownloadedUnsuccessfully()), this, SLOT(rebuildComboBoxList()));
+    connect(hps, SIGNAL(sigPublicServersDownloadedUnsuccessfully(int)), this, SLOT(rebuildComboBoxList(int)));
     connect(btnRefreshServers, SIGNAL(released()), this, SLOT(downloadThePublicServers()));
 
     connect(this, SIGNAL(sigPublicServersDownloaded()), this, SLOT(rebuildComboBoxList()));
@@ -211,7 +211,7 @@ void DlgConnect::downloadThePublicServers()
 {
     previousHosts->clear();
     previousHosts->addItem(placeHolderText);
-    hps->downloadPublicServers(); // Refresh the servers
+    hps->downloadPublicServers();
 }
 
 void DlgConnect::preRebuildComboBoxList()
@@ -226,15 +226,18 @@ void DlgConnect::preRebuildComboBoxList()
     }
 }
 
-void DlgConnect::rebuildComboBoxList()
+void DlgConnect::rebuildComboBoxList(int failure)
 {
+    Q_UNUSED(failure);
+
     previousHosts->clear();
 
     UserConnection_Information uci;
     savedHostList = uci.getServerInfo();
 
     int i = 0;
-    for (UserConnection_Information tmp : savedHostList) {
+    for (auto pair : savedHostList) {
+        auto tmp = pair.second;
         QString saveName = tmp.getSaveName();
         if (saveName.size()) {
             previousHosts->addItem(saveName);
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -222,19 +222,16 @@ void DlgConnect::actFinishParsingDownloadedData()
             }
 
             serversOnWiki.append(serverName);
-            if (! savedHostList.contains(serverName))
-            {
+            if (!savedHostList.contains(serverName)) {
                 // Adds new servers to list (if exists, skip it)
                 settingsCache->servers().addNewServer(serverName, serverAddress, serverPort, "", "", false);
             }
         }
 
         // Now that we added the new servers, we can remove all servers that were not on the public page
-        for (auto server : savedHostList)
-        {
+        for (auto server : savedHostList) {
             QString serverName = server.getSaveName();
-            if (serversOnWiki.indexOf(serverName) < 0)
-            {
+            if (serversOnWiki.indexOf(serverName) < 0) {
                 settingsCache->servers().removeServer(serverName);
             }
         }
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -170,6 +170,9 @@ void DlgConnect::actSaveConfig()
 
 void DlgConnect::downloadPublicServers()
 {
+    previousHosts->clear();
+    previousHosts->addItem(placeHolderText);
+
     QUrl url(QString(PUBLIC_SERVERS_URL));
     auto *nam = new QNetworkAccessManager(this);
     QNetworkReply *reply = nam->get(QNetworkRequest(url));
@@ -199,27 +202,12 @@ void DlgConnect::actFinishParsingDownloadedData()
             QStringList serverValues = server.split(QRegExp("<td[^>]*>"));
             QString serverName, serverAddress, serverPort;
 
-            for (int fieldCheck = 1; fieldCheck < serverValues.size(); fieldCheck++) {
-                // The fields correspond with the headers on the server list (index starting at 1)
-                switch (fieldCheck) {
-                    case 1: // Server Name
-                        serverName =
-                            QString(serverValues.at(fieldCheck)).remove(QRegExp("<a[^>]*>")).split("</a>").at(0);
-                        break;
+            // The fields correspond with the headers on the server list (index starting at 1)
+            short serverNameField = 1, serverAddressField = 5, serverPortField = 6;
 
-                    case 5: // Server Address
-                        serverAddress =
-                            QString(serverValues.at(fieldCheck)).remove("<strong>").split("</strong>").at(0);
-                        break;
-
-                    case 6: // Server Port
-                        serverPort = QString(serverValues.at(fieldCheck)).remove("<strong>").split("</strong>").at(0);
-                        break;
-
-                    default:
-                        break;
-                }
-            }
+            serverName = QString(serverValues.at(serverNameField)).remove(QRegExp("<a[^>]*>")).split("</a>").at(0);
+            serverAddress = QString(serverValues.at(serverAddressField)).remove("<strong>").split("</strong>").at(0);
+            serverPort = QString(serverValues.at(serverPortField)).remove("<strong>").split("</strong>").at(0);
 
             serversOnWiki.append(serverName);
             if (!savedHostList.contains(serverName)) {
@@ -251,7 +239,6 @@ void DlgConnect::preRebuildComboBoxList()
     savedHostList = uci.getServerInfo();
 
     if (savedHostList.size() == 1) {
-        previousHosts->addItem(tr("Downloading..."));
         downloadPublicServers();
     } else {
         emit sigPublicServersDownloaded();
@@ -291,7 +278,7 @@ void DlgConnect::previousHostSelected(bool state)
 
 void DlgConnect::updateDisplayInfo(const QString &saveName)
 {
-    if (saveEdit == nullptr) {
+    if (saveEdit == nullptr || saveName == placeHolderText) {
         return;
     }
 
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -1,5 +1,4 @@
 #include "dlg_connect.h"
-#include "qt-json/json.h"
 #include "settingscache.h"
 #include "userconnection_information.h"
 #include <QCheckBox>
@@ -11,7 +10,6 @@
 #include <QKeyEvent>
 #include <QLabel>
 #include <QMessageBox>
-#include <QNetworkReply>
 #include <QPushButton>
 #include <QRadioButton>
 
@@ -157,40 +155,7 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
     playernameEdit->setFocus();
 }
 
-DlgConnect::~DlgConnect()
-{
-    hostLabel->deleteLater();
-    portLabel->deleteLater();
-    playernameLabel->deleteLater();
-    passwordLabel->deleteLater();
-    saveLabel->deleteLater();
-    publicServersLabel->deleteLater();
-    hostEdit->deleteLater();
-    portEdit->deleteLater();
-    playernameEdit->deleteLater();
-    passwordEdit->deleteLater();
-    saveEdit->deleteLater();
-    savePasswordCheckBox->deleteLater();
-    autoConnectCheckBox->deleteLater();
-    previousHosts->deleteLater();
-    newHostButton->deleteLater();
-    previousHostButton->deleteLater();
-    btnOk->deleteLater();
-    btnCancel->deleteLater();
-    btnForgotPassword->deleteLater();
-    btnRefreshServers->deleteLater();
-    newHostLayout->deleteLater();
-    connectionLayout->deleteLater();
-    buttons->deleteLater();
-    loginLayout->deleteLater();
-    grid->deleteLater();
-    newHolderLayout->deleteLater();
-    loginGroupBox->deleteLater();
-    btnGroupBox->deleteLater();
-    mainLayout->deleteLater();
-    restrictionsGroupBox->deleteLater();
-    hps->deleteLater();
-}
+DlgConnect::~DlgConnect() = default;
 
 void DlgConnect::actSaveConfig()
 {
@@ -209,6 +174,7 @@ void DlgConnect::actSaveConfig()
 
 void DlgConnect::downloadThePublicServers()
 {
+    btnRefreshServers->setDisabled(true);
     previousHosts->clear();
     previousHosts->addItem(placeHolderText);
     hps->downloadPublicServers();
@@ -249,6 +215,8 @@ void DlgConnect::rebuildComboBoxList(int failure)
             i++;
         }
     }
+
+    btnRefreshServers->setDisabled(false);
 }
 
 void DlgConnect::previousHostSelected(bool state)
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -10,6 +10,7 @@
 #include <QKeyEvent>
 #include <QLabel>
 #include <QMessageBox>
+#include <QNetworkReply>
 #include <QPushButton>
 #include <QRadioButton>
 
@@ -21,7 +22,8 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
     previousHosts = new QComboBox(this);
     previousHosts->installEventFilter(new DeleteHighlightedItemWhenShiftDelPressedEventFilter);
 
-    rebuildComboBoxList();
+    connect(this, SIGNAL(sigPublicServersDownloaded()), this, SLOT(rebuildComboBoxList()));
+    preRebuildComboBoxList();
 
     newHostButton = new QRadioButton(tr("New Host"), this);
 
@@ -156,7 +158,79 @@ void DlgConnect::actSaveConfig()
                                               passwordEdit->text(), savePasswordCheckBox->isChecked());
     }
 
-    rebuildComboBoxList();
+    preRebuildComboBoxList();
+}
+
+void DlgConnect::downloadPublicServers(QUrl url)
+{
+    auto *nam = new QNetworkAccessManager(this);
+    QNetworkReply *reply = nam->get(QNetworkRequest(url));
+    connect(reply, SIGNAL(finished()), this, SLOT(actFinishParsingDownloadedData()));
+}
+
+void DlgConnect::actFinishParsingDownloadedData()
+{
+    auto *reply = dynamic_cast<QNetworkReply *>(sender());
+    QNetworkReply::NetworkError errorCode = reply->error();
+
+    if (errorCode == QNetworkReply::NoError) {
+        // This will get all the data from the GitHub page and just give us the table data
+        QString stringData =
+            QString(reply->readAll()).remove(QRegExp(R"([\n\t\r])")).split("<tbody>").at(1).split("</tbody>").at(0);
+
+        // [Server1 Details], [Server2 Details], ...
+        QStringList serverList = stringData.split(QRegExp("<tr[^>]*>"));
+        serverList.removeAt(0);
+
+        for (const auto &server : serverList) {
+            QStringList serverValues = server.split(QRegExp("<td[^>]*>"));
+            QString serverName, serverAddress, serverPort;
+
+            for (int fieldCheck = 1; fieldCheck < serverValues.size(); fieldCheck++) {
+                // The fields correspond with the headers on the server list (index starting at 1)
+                switch (fieldCheck) {
+                    case 1: // Server Name
+                        serverName =
+                            QString(serverValues.at(fieldCheck)).remove(QRegExp("<a[^>]*>")).split("</a>").at(0);
+                        break;
+
+                    case 5: // Server Address
+                        serverAddress =
+                            QString(serverValues.at(fieldCheck)).remove("<strong>").split("</strong>").at(0);
+                        break;
+
+                    case 6: // Server Port
+                        serverPort = QString(serverValues.at(fieldCheck)).remove("<strong>").split("</strong>").at(0);
+                        break;
+
+                    default:
+                        break;
+                }
+            }
+
+            settingsCache->servers().addNewServer(serverName, serverAddress, serverPort, "", "", false);
+        }
+
+        reply->deleteLater();
+    } else {
+        // Respond that an error occurred
+        qDebug() << "ERROR DOWNLOADING PUBLIC SERVERS" << errorCode;
+    }
+
+    emit sigPublicServersDownloaded();
+}
+
+void DlgConnect::preRebuildComboBoxList()
+{
+    UserConnection_Information uci;
+    savedHostList = uci.getServerInfo();
+
+    if (savedHostList.size() == 1) {
+        previousHosts->addItem("Downloading...");
+        downloadPublicServers(QString(PUBLIC_SERVERS_URL));
+    } else {
+        emit sigPublicServersDownloaded();
+    }
 }
 
 void DlgConnect::rebuildComboBoxList()
@@ -164,14 +238,6 @@ void DlgConnect::rebuildComboBoxList()
     previousHosts->clear();
 
     UserConnection_Information uci;
-    savedHostList = uci.getServerInfo();
-
-    if (savedHostList.size() == 1) {
-        settingsCache->servers().addNewServer("Woogerworks", "cockatrice.woogerworks.com", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Chickatrice", "chickatrice.net", "4747", "", "", false);
-        settingsCache->servers().addNewServer("dr4ft", "cockatrice.dr4ft.com", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Tetrarch", "mtg.tetrarch.co", "4747", "", "", false);
-    }
     savedHostList = uci.getServerInfo();
 
     int i = 0;
@@ -232,7 +298,7 @@ void DlgConnect::newHostSelected(bool state)
         saveEdit->setPlaceholderText("New Menu Name");
         saveEdit->setDisabled(false);
     } else {
-        rebuildComboBoxList();
+        preRebuildComboBoxList();
     }
 }
 
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -16,8 +16,7 @@
 #include <QRadioButton>
 
 #define PUBLIC_SERVERS_URL "https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers"
-#define PUBLIC_SERVERS_JSON                                                                                            \
-    "https://raw.githubusercontent.com/Cockatrice/cockatrice.github.io/master/public-servers.json"
+#define PUBLIC_SERVERS_JSON "https://cockatrice.github.io/public-servers.json"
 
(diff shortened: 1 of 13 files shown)


Nightfirecat:
Not a fan of reading from wiki--anyone can create an account and vandalize that. Could we maintain a file in this repo (or a separate repo) to hold that information?

ancestral:
You can do it through the wiki, as long as the page has admin-only access.

Other sites do this sort of thing. [The Battle for Wesnoth](http://wesnoth.org) does this for their download links when they release updates, and the system works well, especially for maintainers who are not well-versed with command line functions.

ancestral:
Still, I think most pieces of software with update checks will look utilize REST APIs (for example, GET `http://cockatrice.us/api/servers` might return simple JSON with a list of servers) and when things change, they will just edit the static file on the server.

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +165,104 @@ void DlgConnect::actSaveConfig()
...
+                        break;
+
+                    case 6: // Server Port
+                        serverPort = QString(serverValues.at(fieldCheck)).remove("<strong>").split("</strong>").at(0);
+                        break;
+
+                    default:
+                        break;
+                }
+            }
```
Vafthrudnir: I looping really necessary here? All it does is increment an index variable, which is used in a switch - wouldn't directly indexing the `serverValues` list be nicer?

```
short nameIndex = 1;
short addressIndex = 5;
short portIndex = 6;
serverName = QString(serverValues.at(nameIndex)).remove(QRegExp("<a[^>]*>")).split("</a>").at(0);
serverAddress = QString(serverValues.at(addressIndex)).remove("<strong>").split("</strong>").at(0);
serverPort = QString(serverValues.at(portIndex)).remove("<strong>").split("</strong>").at(0);
...
```

ZeldaZach: Honestly, I wrote this on a whim and once it worked I put it up for review. Looking back, the loop is kind of pointless and I'm sure there's a better way

Vafthrudnir: By the way I don't really like that we get this list from a displayed source; if someone would change the formatting of the text (e.g. make it italic) the code would also have to be changed. It would be much better imo if we either had a dedicated file for the server list or used an API as @ancestral proposed.

tooomm:
>Still, I think most pieces of software with update checks will utilize REST APIs (for example, GET http://cockatrice.us/api/servers might return simple JSON with a list of servers) and when things change, they will just edit the static file on the server.

That's not working for us I think.
As long as that API is not part of our https://cockatrice.github.io/ webpage [hosted from gh pages](https://github.com/Cockatrice/cockatrice.github.io).

<br>

>You can do it through the wiki, as long as the page has admin-only access.

We would have to make all wiki admin-only. You can't do that for certain pages only I think?
Also that wouldn't help too much since all server owners would have to create an issue and somebody has to manually update the wiki.
If it's a file in a repo all owners can update for themselves and somebody just needs to push the merge button after checking. That's a simpler and better process and keeps the list a bit easier to maintain.

The wiki page was for pretty presentation purpose, its structure or content might change (like I added the up/down badges recently). The parsing code would need to reflect such changes. Or somebody could easily mess a change up without even noticing and break things for all users.
We are looking for a robust and simple data source here instead.
Others might not realize a bad change because no PR got created, nobody checked the change and no notifications about changes got send etc.

As spoken about in the initial issue, I prefer a github based file, too.
Placed in a separate repo, gh pages repo, cockatrice repo ...

<br>

>When the user "refreshes the servers", we will re-download the list from the wiki and **clear any server not found on the public list**.

We don't want to do that I guess? 😛 
Players still want to connect to their private or custom servers (testing, friends, lgs, pro team...).

Custom created servers should always be kept. Do we need two lists here then?
A official, public one that we maintain and update. Plus a custom one the user can add to.
Like with cards etc.

cockatrice/src/dlg_connect.cpp
ZeldaZach: I'd be fine with having a dedicated file; where would we store it?

ZeldaZach:
@tooomm Nothing I can do about the sorting issue; If we capitalize the server name on the page, it should work. We added manually and this is how QMaps work

ancestral:
@Daenyth As I said, you would make the page private, editable only by admins or certain people.

Nightfirecat:
@ancestral Far as I can tell, there is no "private, editable-only-by-admins" option for wiki pages. All pages are editable by any GitHub user who has access to the repo.

ancestral:
@Nightfirecat It might depend on the wiki software you are running, [but you can definitely add user groups in MediaWiki.](https://www.mediawiki.org/wiki/Manual:User_rights)

Nightfirecat:
I'm referring to GitHub wiki, [since that's where it's currently stored](https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers).

tooomm:
You can limit edits to org members in the github wiki too, but then all server admins can't manage their data anymore on their own as outlined above since they are not involved in development mostly. As well as the other drawbacks which come with using the wiki as source. IMO it's not the best solution here.

Daenyth:
I'd really prefer using git. It has a review process, it's public, and it's reliable. Even if we can lock the one wiki page to org members only.

tooomm:
You can only lock all wiki as far as I know^^

+1 for git

ZeldaZach:
Latest Commit: Will now use https://github.com/Cockatrice/cockatrice.github.io/blob/master/public-servers.json instead of the Wiki. This allows us to keep the server details in git, prevent issues with stylistic changes, and better determine the order shown in client.

cockatrice/src/dlg_connect.cpp
```diff
@@ -10,18 +11,25 @@
 #include <QKeyEvent>
 #include <QLabel>
 #include <QMessageBox>
+#include <QNetworkReply>
 #include <QPushButton>
 #include <QRadioButton>
 
 #define PUBLIC_SERVERS_URL "https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers"
+#define PUBLIC_SERVERS_JSON                                                                                            \
+    "https://raw.githubusercontent.com/Cockatrice/cockatrice.github.io/master/public-servers.json"
```
Nightfirecat: Looks like this file needs to be added?

ZeldaZach: Does the file not work for you? The formatting was clang-format

Nightfirecat: D'oh, I was looking in the wrong place. Never mind. :upside_down_face:

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +168,96 @@ void DlgConnect::actSaveConfig()
...
+
+        // If a server was removed from the public list,
+        // we will delete it from the local system
+        // as that server is no longer around.
+        for (auto server : savedHostList) {
+            QString serverName = server.getSaveName();
+            if (publicServersAvailable.indexOf(serverName) < 0) {
+                settingsCache->servers().removeServer(serverName);
+            }
+        }
```
ctrlaltca: I see a couple problematic points with this approach:
 * if a server gets updated/edited, the change won't be picked up
 * if the user added a private/custom server (eg. i usually connect to 127.0.0.1 for testing), the custom server will be removed.

The first point could be resolved just updating the server address/port if the name matches.
The second point is a bit more difficult: should we handle 2 different types of server, "official" servers and user-defined server?
We could just add an "official" field to each server and only update those?

Vafthrudnir: I think the problem with the "official" field would be that at those users who have a configuration generated by a previous version of the program we couldn't differentiate between an "official" and a "custom" server - we either have to delete both or neither.
But I think we could solve this problem if we were to add a tag to the json file storing whether a specific server is "dead" or not. If it is dead and the user's config contains it then it should be deleted, otherwise it should be added, or if it already exists updated. The only drawback would be that we mustn't remove obsolete servers from the json file in this case.

cockatrice/src/dlg_connect.cpp
tooomm: We can link to https://cockatrice.github.io/public-servers.json instead, too!

ctrlaltca: Right, that url is probably more well suited for access from a lot of users

tooomm:
Do we need all that additional details about server location, webpage, auth and reg etc. in the file?

You simply preserved all information from the wiki I assume, but we don't use that currently.
I want to keep the file as simple and clean as possible. Especially since we plan on querying each server for such information directly as I understand it (https://github.com/Cockatrice/Cockatrice/issues/2474, also https://github.com/Cockatrice/Cockatrice/issues/742).

@ctrlaltca also mentioned it in the issue you are about to close with this PR:
> (advanced, needs proper design): connect without logging in and ask the server for basic informations (requires registration / accepts new registrations / has free slots for new users / current num of users and games / planned shutdown / whatever )

The JSON mostly just needs `host` and `port` as basic information for this.
Everything else could come from the server response each owner maintains locally in the long run.
But for now we need to serve `name` too I guess.

backendr:
@ZeldaZach if you are going to use that json then you will need to removed woogerworks from it.

backendr:
Your code is self documented, I dont think you need all the comments.

ZeldaZach:
Latest Commit: Whenever a new host is added manually by the user (from now on), we will mark it as a custom server. Custom servers will not be purged by the refresh servers feature. This will purge old ones prior to this PR, but I'm fine with this.

In addition, server updates/changes will be applied. When refreshed, we will ensure all fields (minus the friendly name) are up to date with the JSON file.

I've added a new field,  bool "isInactive" to the JSON. if this is set to true, the server is no longer available. The only information necessary for an inactive server are ID, name, and isInactive. Hope this addresses your concern @backendr 

Also @backendr , having more comments is not a problem as I want to be as explicit as possible

@tooomm If you don't feel we need as much information in the JSON, feel free to edit it to cut down what we don't need. I'm only currently using name, host, port, and isInactive. I may find a use for the rest in the future, but if you think you'll be able to get around to the other tickets make any changes you find necessary

cockatrice/src/dlg_connect.cpp
ZeldaZach: Thanks for that, @tooomm . Addressed!

ZeldaZach:
> It would be nice to be able to use the server list also from the "register to server" and "password-reset/request/challenge" dialogs. Even that's not a task to be done in this PR, the code could still be moved inside an external file or structured so that it can be used by all those dialogs in the future.

Just saw this @ctrlaltca. I'll see about moving it to an external resource soon. Are we looking to have the entire download/update be moved or just the way to create the dropdown menu with the server information?

Daenyth:
I'm going to disagree with ctrl and suggest that this refactor doesn't need to happen inside this PR. I'd approve it if included, but I wouldn't *not* merge this because of it.

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +167,104 @@ void DlgConnect::actSaveConfig()
                                               passwordEdit->text(), savePasswordCheckBox->isChecked());
     }
 
-    rebuildComboBoxList();
+    preRebuildComboBoxList();
 }
 
-void DlgConnect::rebuildComboBoxList()
+void DlgConnect::downloadPublicServers()
```
Daenyth: I'm not sure this belongs in here - it seems appropriate to be inside of an object whose only responsibility is to download the server data and fire a signal on success/fail completion

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +167,104 @@ void DlgConnect::actSaveConfig()
...
+        }
+
+        reply->deleteLater();
+    } else {
+        qDebug() << "[CONNECTION DIALOG]"
+                 << "Error Downloading Public Servers" << errorCode;
+    }
+
+    // Signal that we're done with updating the combo box
+    emit sigPublicServersDownloaded();
```
Daenyth: What do you think about having separate signals for success/fail? Or if one signal, at least include a data object that gives info about the failure. I don't like swallowing it.

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +167,104 @@ void DlgConnect::actSaveConfig()
...
 
     if (savedHostList.size() == 1) {
-        settingsCache->servers().addNewServer("Rooster Ranges", "server.cockatrice.us", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Chickatrice", "chickatrice.net", "4747", "", "", false);
-        settingsCache->servers().addNewServer("dr4ft", "cockatrice.dr4ft.com", "4747", "", "", false);
-        settingsCache->servers().addNewServer("Tetrarch", "mtg.tetrarch.co", "4747", "", "", false);
-    } else if (!savedHostList.contains("Rooster Ranges")) {
-        settingsCache->servers().addNewServer("Rooster Ranges", "server.cockatrice.us", "4747", "", "", false);
-        settingsCache->servers().removeServer("Woogerworks");
+        downloadPublicServers();
```
Daenyth: What does this look like at first run time? We should have a label in there that tells the user the data is being downloaded.

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +167,104 @@ void DlgConnect::actSaveConfig()
...
+    } else {
+        emit sigPublicServersDownloaded();
```
Daenyth: If the saved host list != size 1, emit a signal saying we downloaded a server list? The logic doesn't make sense to me at all. What is this for?

cockatrice/src/dlg_connect.cpp
ZeldaZach: This can probably be refactored to an external file, and I can look into that if you think it's worth it

cockatrice/src/dlg_connect.cpp
ZeldaZach: The signal's job here is to just say "Hey we're done, go rebuild" It has no other meaning. If you want a 2nd signal possible to be emitted and all of them caught so be it

cockatrice/src/dlg_connect.cpp
ZeldaZach: It has a label, the DPS() adds it

cockatrice/src/dlg_connect.cpp
ZeldaZach: This signal just means continue the rebuilding process. If the size != 1 then we have info already stored locally and do not need to do a first time run install

cockatrice/src/dlg_connect.cpp
```diff
@@ -156,26 +204,33 @@ void DlgConnect::actSaveConfig()
...
-    rebuildComboBoxList();
+    preRebuildComboBoxList();
 }
 
-void DlgConnect::rebuildComboBoxList()
+void DlgConnect::downloadThePublicServers()
 {
     previousHosts->clear();
+    previousHosts->addItem(placeHolderText);
+    hps->downloadPublicServers(); // Refresh the servers
```
Daenyth: useless comment 🔪

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,92 @@
+//
+// Created by Zachary Halpern on 4/6/18.
```
Daenyth: git already knows this, comment isn't needed.

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,92 @@
...
+        savedHostList = uci.getServerInfo();
+
+        // List of public servers
+        // Will be populated below
+        QStringList publicServersAvailable;
+        QStringList publicServersToRemove;
+
+        // Downloaded data from GitHub
+        QString jsonData = QString(reply->readAll());
+        auto jsonMap = QtJson::Json::parse(jsonData).toMap();
```
Daenyth: What happens if the parse fails?

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+        // Servers available
+        auto publicServersJSONList = jsonMap["servers"].toList();
```
Daenyth: Can you split this behavior out into its own function taking jsonMap as a parameter? That would make this class far more testable because we can ignore all the network logic and focus on testing the json logic.

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,92 @@
...
+                publicServersToRemove.append(serverName);
+                continue;
+            }
+
+            QString serverAddress = serverMap["host"].toString();
+            QString serverPort = serverMap["port"].toString();
+
+            publicServersAvailable.append(serverName);
+
+            if (!savedHostList.contains(serverName)) {
```
Daenyth: What about checking for server url instead of name? That seems more likely to be accurate.

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,92 @@
...
+            if (!isCustom && publicServersToRemove.indexOf(serverName) != -1) {
+                settingsCache->servers().removeServer(serverName);
+            }
+        }
+
+        emit sigPublicServersDownloadedSuccessfully();
+    } else {
+        qDebug() << "[CONNECTION DIALOG]"
+                 << "Error Downloading Public Servers" << errorCode;
+        emit sigPublicServersDownloadedUnsuccessfully();
```
Daenyth: Include the error code?

cockatrice/src/handle_public_servers.h
```diff
@@ -0,0 +1,38 @@
+//
+// Created by Zachary Halpern on 4/6/18.
```
Daenyth: 🔪

cockatrice/src/handle_public_servers.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef COCKATRICE_HANDLE_PUBLIC_SERVERS_H
+#define COCKATRICE_HANDLE_PUBLIC_SERVERS_H
+
+#include "userconnection_information.h"
+#include <QNetworkReply>
+
+/**
+ * This class is used to update the servers.ini file and ensure
+ * the list of public servers has up-to-date information.
+ * Servers that are added manually by users are not modified.
```
Daenyth: Wonderful doc comment 👍

cockatrice/src/handle_public_servers.cpp
ZeldaZach: CLion added it, didn't realize

Daenyth:
Can you include some more defensive coding around the json handling? If we good and accidentally break the file or miss including a key, this won't work right - we should also emit debug logs for bad json

cockatrice/src/handle_public_servers.cpp
Daenyth: settings > new file template or so, to fix that for good

cockatrice/src/handle_public_servers.cpp
ZeldaZach: The host list is a QList<QString> of server names only... I'd rather not change that around if I can avoid it

cockatrice/src/handle_public_servers.cpp
ZeldaZach: Done

cockatrice/src/handle_public_servers.cpp
ZeldaZach: Addressed

cockatrice/src/handle_public_servers.cpp
ZeldaZach: Sure... doesn't really matter as qdebug says it, but i can include it

cockatrice/src/handle_public_servers.h
ZeldaZach: yep

cockatrice/src/handle_public_servers.cpp
Daenyth: std::pair it up?

cockatrice/src/handle_public_servers.h
```diff
@@ -0,0 +1,37 @@
+#ifndef COCKATRICE_HANDLE_PUBLIC_SERVERS_H
+#define COCKATRICE_HANDLE_PUBLIC_SERVERS_H
+
+#include "userconnection_information.h"
+#include <QNetworkReply>
```
ctrlaltca: More a hint than a comment.
Since this header file is meant to be included in other classes (eg. dlg_connect.cpp), and `#include` basically means "replace this line with the whole content of the included file", you may want to minimize the number of `#include`s that gets propagated.

In this case, since here you are not using the `QNetworkReply` class itself but only a pointer to it, you can just replace this line with a dummy
```
class QNetworkReply;
class QNetworkAccessManager;
```

and move the `#include <QNetworkReply>` in the `handle_public_servers.cpp` file, where the full class definition is really needed.

Using this hint everywhere in he code can speed up compilation a lot and also, in some rare cases, avoid obscure bugs due to double inclusion/definition of the same named resource.

cockatrice/src/userconnection_information.h
```diff
@@ -16,35 +16,40 @@ class UserConnection_Information
...
     UserConnection_Information();
-    UserConnection_Information(QString, QString, QString, QString, QString, bool);
-    QString getSaveName()
+    UserConnection_Information(QString, QString, QString, QString, QString, bool, bool _isCustom = false);
+    QString getSaveName() const
     {
         return saveName;
     }
-    QString getServer()
+    QString getServer() const
```
ctrlaltca: I love this kind of accuracy fixes ❤️

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,107 @@
...
+#define PUBLIC_SERVERS_JSON "https://cockatrice.github.io/public-servers.json"
+
+HandlePublicServers::HandlePublicServers(QObject *parent) : QObject(parent), nam(nullptr), reply(nullptr)
+{
+}
+
+void HandlePublicServers::downloadPublicServers()
+{
+    QUrl url(QString(PUBLIC_SERVERS_JSON));
+    nam = new QNetworkAccessManager(this);
```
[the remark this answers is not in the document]
tooomm: See https://github.com/Cockatrice/Cockatrice/pull/3165#pullrequestreview-109222816

With manual spoiler refresh the reason to disable the button and change its label until the process is finished was to also prevent request spamming and potential further problems as you described above for example.

The `Downloading...` text/hint in the drop down menu itself is a nice addition though. 👍

cockatrice/src/dlg_connect.cpp
```diff
@@ -144,6 +157,41 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
     playernameEdit->setFocus();
 }
 
+DlgConnect::~DlgConnect()
+{
+    hostLabel->deleteLater();
```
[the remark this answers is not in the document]
ZeldaZach: If you say so

cockatrice/src/dlg_connect.cpp
```diff
@@ -1,4 +1,5 @@
 #include "dlg_connect.h"
+#include "qt-json/json.h"
```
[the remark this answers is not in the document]
ZeldaZach: Yep

cockatrice/src/dlg_connect.cpp
```diff
@@ -10,6 +11,7 @@
 #include <QKeyEvent>
 #include <QLabel>
 #include <QMessageBox>
+#include <QNetworkReply>
```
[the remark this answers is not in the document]
ZeldaZach: Yep

cockatrice/src/dlg_connect.h
```diff
@@ -48,15 +55,23 @@ private slots:
...
     QLineEdit *hostEdit, *portEdit, *playernameEdit, *passwordEdit, *saveEdit;
     QCheckBox *savePasswordCheckBox, *autoConnectCheckBox;
     QComboBox *previousHosts;
     QRadioButton *newHostButton, *previousHostButton;
-    QPushButton *btnOk, *btnCancel, *btnForgotPassword;
-    QMap<QString, UserConnection_Information> savedHostList;
+    QPushButton *btnOk, *btnCancel, *btnForgotPassword, *btnRefreshServers;
+    QMap<QString, std::pair<QString, UserConnection_Information>> savedHostList;
+    HandlePublicServers *hps;
+    const QString placeHolderText = tr("Downloading...");
```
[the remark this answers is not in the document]
ZeldaZach: I see nothing online that states you _cant_ do this

cockatrice/src/handle_public_servers.cpp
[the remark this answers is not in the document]
ZeldaZach: Addressed

cockatrice/src/handle_public_servers.h
ZeldaZach: I see, never really understood the purpose. Thanks!

cockatrice/src/userconnection_information.h
ZeldaZach: :D

ZeldaZach:
@tooomm 
* Modifying a public server will not make it "custom". Only if you add via the "new host" method.
* Yes, Dae told me to change it to use the URL instead of nick name
* Because it's the easiest way for me to figure out what's going on for what to remove and what to keep. We only archive the name, so I'm fine with this.
* Sounds good!

ZeldaZach:
Ok, I've addressed everyone's comments... How's this looking now @Daenyth @ctrlaltca @tooomm ?

I've also added an option with this latest push to have a menu option that refreshes the servers. This will be the new default as of the 2.7.0 release (_next next_ release) where we will remove the GUI button.

ZeldaZach:
Any opinions on a one-time flush of the server list so it re-downloads the public? Or should we leave it as is? I just tested with an older .ini file and refreshed. While it added the new servers, the old ones were left behind (since they were "custom" and were old)... I'd like to flush them all first time through

tooomm:
Yes, I realized that the old woods server didn't get deleted, too.
But without warning the user/giving a choice, I don't like flushing all their personal lists.

Another thing I run into, is that refreshing the list kills your selection and jumps to last host used. So if you swapped servers and entered new login details but decide to refresh servers first, all your new input data is lost and the selection jumps back to the last connected server.

tooomm:
From Gitter:

>Zach H @ZeldaZach Apr 07 18:28
A thought. Should I move the refresh server button to the menu instead of on the GUI
>
>David Szabo @Vafthrudnir Apr 07 18:56
I think it's okay on the gui
Easily recognizable
>
>Zach H @ZeldaZach Apr 07 18:57
I'm gonna try both for now
Because it's not a button people will press often
>
>David Szabo @Vafthrudnir Apr 07 19:04
The main use-case will be IMO that when a user can't connect to their usual server,
They will click it and try again
And in this case it's better if the refresh button is right there instead of somewhere in a menu
>
>Zach H @ZeldaZach Apr 07 19:07

I kind of find this better tbh
>
>David Szabo @Vafthrudnir Apr 07 19:16
What happens when it's clicked? Does it just start the refresh process in the background?
>
>Zach H @ZeldaZach Apr 07 19:30
Yes, and it will launch a popup on failure/success
I think, for this release, I'll include the button. Afterwards I'll remove it

I have to disagree here. I highly prefer the option within the related dialog instead a separate option in the main app menu.
What we can do is to exchange the big text button for a small one right next to the drop down menu if you think it catches too much attention for a not so often needed feature. Instead a big text label it can only be a "refresh/reload circle" icon similar to the one we use in the custom shortcuts page in settings:
refresh-small
A clear and helpful button hover label should provide a bit more additional info to the user like "Refresh the server list from a online source with provided public ones"

ZeldaZach:
@tooomm I love that idea of re-using the "refresh" button from the shortcuts! I'll look into that today or tomorrow. If that's the case and we stick with it, I'll be removing the dialog option as it won't be needed.

cockatrice/src/dlg_connect.h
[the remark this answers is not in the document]
tooomm: You can build new strings locally, and test what happens when you add a translation to e.g. the italian string file. 👍

ZeldaZach:
Newest commit: Reverts the menu option, moves the button and uses the refresh icon next to the server list with a tool tip to let you know what it does. On successful run, it will ask the user if they want to clear their old list or not (Sorry @Daenyth  :P).

screenshot 2018-04-08 23 47 24
screenshot 2018-04-08 23 47 28

cockatrice/src/window_main.cpp
```diff
@@ -1205,7 +1227,7 @@ int MainWindow::getNextCustomSetPrefix(QDir dataDir)
 
 void MainWindow::actManageSets()
 {
-    auto *w = new WndSets;
+    w = new WndSets;
```
Vafthrudnir: I think a one-character variable name was okay in this local scope, but since now it's made into a data member of a quite large class it probably should get a more expressive name.

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        QString serverAddress = serverMap["host"].toString();
+        QString serverPort = serverMap["port"].toString();
+
+        publicServersAvailable.append(serverName);
+
+        bool serverFound = false;
+        for (const auto &iter : savedHostList) {
+            if (iter.first == serverName) {
```
ZeldaZach: TODO: Change this to match by URL

cockatrice/src/window_main.cpp
ZeldaZach: Addressed

cockatrice/src/handle_public_servers.cpp
```diff
@@ -0,0 +1,104 @@
...
+        if (serverMap["isInactive"].toBool()) {
+            publicServersToRemove.append(serverAddress);
+            continue;
+        }
+
+        QString serverName = serverMap["name"].toString();
+        QString serverPort = serverMap["port"].toString();
+
+        bool serverFound = false;
+        for (const auto &iter : savedHostList) {
```
Daenyth: This would be clearer with `std::find_if` but this is also fine 👍

cockatrice/src/settingscache.cpp
```diff
@@ -492,32 +493,32 @@ void SettingsCache::setMinPlayersForMultiColumnLayout(int _minPlayersForMultiCol
...
 {
-    chatMention = _chatMention;
+    chatMention = static_cast<bool>(_chatMention);
     settings->setValue("chat/mention", chatMention);
 }
 
 void SettingsCache::setChatMentionCompleter(const int _enableMentionCompleter)
 {
-    chatMentionCompleter = _enableMentionCompleter;
+    chatMentionCompleter = (bool)_enableMentionCompleter;
```
Daenyth: Why not `static_cast` here?

cockatrice/src/shortcutssettings.cpp
```diff
@@ -150,10 +150,7 @@ bool ShortcutsSettings::isKeyAllowed(QString name, QString Sequences)
                                                << "Shift+S"
                                                << "Shift+Left"
                                                << "Shift+Right");
-    if (forbiddenKeys.contains(checkSequence)) {
-        return false;
-    }
-    return true;
+    return !forbiddenKeys.contains(checkSequence);
```
Daenyth: 👍

Daenyth:
I really appreciate everyone going through the multiple rounds of feedback; this is going to be a great feature for our users.

Nice work @ZeldaZach @Vafthrudnir @ctrlaltca @Nightfirecat @tooomm :D

cockatrice/src/settingscache.cpp
ZeldaZach: Clion changed it to be like this. I'm assuming it has something to do with the const functionality and it might need a `const_cast` which is dangerous. C style probably avoids this. @ctrlaltca ?

cockatrice/src/handle_public_servers.cpp
ZeldaZach: Never heard of it, so I'd rather leave it as-is for now. Will look into the standard functions though for future reference!

Nightfirecat:
Agreed, great work all, and especially great work @ZeldaZach for implementing! :shipit:

cockatrice/src/settingscache.cpp
ctrlaltca: The member var `chatMentionCompleter` is a boolean, while the `_enableMentionCompleter` parameter is an integer. Clion just made the "int to bool" cast explicit; no const involved.

Daenyth: Dynamic cast then? I thought our style was to not use the c-style paren casts any more in favor of the cpp casts

tooomm:
@Lachee @dev-id 
FYI: the central file Cockatrice pulls its server data from is accessible trough [this url](https://cockatrice.github.io/public-servers.json), and stored/edited right in the github repo for our webpage:
https://github.com/Cockatrice/cockatrice.github.io/blob/master/public-servers.json

Please keep that updated in case of important changes (as well as the [wiki page for public servers](https://github.com/Cockatrice/Cockatrice/wiki/Public-Servers)).

(@Lachee: Also, check your gitter pm's by chance 😉)

PR #3167, opened 2018-04-04
allow multiple emails to be used by servatrice

ZeldaZach:


ZeldaZach:
I think I'll redo this to allow multiple options to be sent to avoid spam filters

servatrice/servatrice.ini.example
```diff
@@ -192,20 +192,26 @@ connection=tcp
 ; Accept all certificates: in ssl mode, enable this if your server is using an invalid/self signed certificate
 acceptallcerts=false;
 
+; The host, port, username, password, and email fields can be expanded with the next
+; sequential inter to have multiple accounts to choose from, which will be done
```
Daenyth: `inter` -> `integer` ?

Daenyth:
Email send limits are often per-ip, likewise with spam. Really a server operator should have DKIM and so on set up for helping with that kind of thing.

We actually should probably do the opposite of this - set a configurable cap on number of actions. Limit the number of emails that a particular user can send per time span and limit how many the server overall can send per time span.

Daenyth:
> The goal will be to give arrays of emails and phrases to send so you have a larger combination class instead of 1C1 * 1C1

Also this is literally exactly what spammers do, you are really more likely to get someone marked as spam by acting like a spammer.

PR #3184, opened 2018-04-09
Add a note about bug #3135

### LINKED ISSUE #3135: Beta channel doesn't find stable releases any longer
Starting with the update to 2.5.0 - released on 2nd march - (from a former 2.5.0-beta) via the build in updater, I realized that the "Beta Channel" no longer finds and suggests updates to full stable versions of the app:
beta channel

I'm pretty sure this used to be different. It makes no sense to only look for releases tagged with "pre-releases".

Users should get any most recent update, including normal releases, if opting to receive beta releases additionally. Otherwise there will be a weird delay and gap.

ctrlaltca:
## Related Ticket(s)
- Fixes #3135 ; not really a proper fix, but an acceptable workaround


ZeldaZach:
Great fix 👍 😛

PR #3204, opened 2018-04-16
datetime display fixed for OSs with non-English localization

### LINKED ISSUE #3140: Spoiler pop-up broken on windows
Client Version: 2.5.0 (2018-03-02)
Build Architecture: 64-bit
Qt Version: 5.9.3

__________

There is no date shown in the update notification for spoiler on windows:
spoiler-udpate_mistake

If I understood @ZeldaZach correctly it works fine on macOS. No idea about Linux.

Vafthrudnir:
## Related Ticket(s)
- Fixes #3140 

Time of last spoiler update was not displayed on OS-s where the localization was set to other than English. 

`QDateTime::fromString()` method tries to parse the given string based on the OS's default localization setting, which caused an invalid date, since the string was in English. Setting `QLocale` default to English and using it to parse dates instead of `QDateTime` solved the problem.

## Screenshots
image

### DIFF
--- a/cockatrice/src/spoilerbackgroundupdater.cpp
+++ b/cockatrice/src/spoilerbackgroundupdater.cpp
@@ -3,6 +3,7 @@
 #include <QDateTime>
 #include <QDebug>
 #include <QFile>
+#include <QLocale>
 #include <QMessageBox>
 #include <QNetworkReply>
 #include <QUrl>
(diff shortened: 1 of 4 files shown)


cockatrice/src/spoilerbackgroundupdater.cpp
```diff
@@ -165,11 +166,11 @@ bool SpoilerBackgroundUpdater::saveDownloadedFile(QByteArray data)
         QList<QByteArray> lines = data.split('\n');
 
         foreach (QByteArray line, lines) {
-            if (line.indexOf("created:") > -1) {
+            if (line.contains("created:")) {
```
Daenyth: 👍

Vafthrudnir:
Added the localization settings and cockatrice language to the logs. Localization is only checked on startup, and language is logged with every change.

tooomm:
```
Client Version: 2.5.0-custom(e00467c) (2018-04-16)
Build Architecture: 64-bit
Qt Version: 5.9.4
System Locale: de_DE
Theme changed: "Magic The Gathering"
SoundEngine: disabling sound
Sound theme changed: "Default"
SoundEngine: disabling sound
Language changed:  "en"
main(): starting main program
main(): MainWindow constructor finished
CardDatabase::loadCardDatabases start
main(): ui.show() finished
```

There is a double space between `Language changed:` and `"en"`, same in your example @Vafthrudnir.

Otherwise it works for me and the date is correctly shown again! Awesome 👍

PR #3207, opened 2018-04-19
log cleanup

tooomm:
People often only submitted half of the needed information

- introduce a separation line in the log to stress what is important head data
- clear log option set to false
- repeated system info logging removed
- remove tr() to not get any foreign translated logs submitted
- adjusted issue template

<pre>
Note:
Closing with clearing the Debug Log window will remove the dash line and
further logging will start right below the system details until you restart the app
</pre>

## Screenshots
untitled

Thanks @Vafthrudnir 👍


Vafthrudnir:
> Note:
There is one issue left... closing and clearing the Debug Log window will remove
the dash line and further logging will start right below the system details after that

I think we shouldn't implement a logic to always keep this part of the log and clear everything else - it would be much easier if the `Clear log when closing` checkbox would be unchecked by default. That would keep the log intact at most users.

PR #3211, opened 2018-04-21
connect dialog tweaks

tooomm:
- smaller refresh button
- disable host and port fields for known server
- add a webpage label next to the url (hidden for custom server)
- real url is displayed instead mirroring server name
- move box with server contact up
- more explaining placeholder text for adding new custom server

## Screenshots
new
new2

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -123,7 +123,7 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
     serverInfoLayout = new QGridLayout;
     serverInfoLayout->addWidget(serverIssuesLabel, 0, 0, 1, 3, Qt::AlignTop);
     serverInfoLayout->addWidget(serverContactLabel, 1, 0);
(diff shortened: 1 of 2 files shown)


cockatrice/src/dlg_connect.cpp
```diff
@@ -232,6 +234,8 @@ void DlgConnect::previousHostSelected(bool state)
 {
     if (state) {
         saveEdit->setDisabled(true);
+        hostEdit->setDisabled(true);
```
ZeldaZach: disagree with disabling these. If they typo a custom it goes for naught as they have to delete/remake it.

tooomm: I see that problem.
Though, if they mistype the server url they can't change that anyway because we update/compare based on that. Same for name since that field was disabled all the time. They could only fix a wrong port number from within the UI.

Changing the server url does nothing currently, it won't save a new one and doesn't update the existing. Without deleting/editing the ini file the user is basically stuck with what they entered, we never supported editing existing entries from within the UI.

What we could do is to disable url, but enable name to let them edit that.

PR #3215, opened 2018-04-25
log moderation stats in the uptime table

ZeldaZach:
Update the server stats logging to include mods / which mods are logged in. I find it useful as a server operator to know who on my staff is around.

### DIFF
--- a/common/server.cpp
+++ b/common/server.cpp
@@ -53,8 +53,6 @@ Server::Server(QObject *parent) : QObject(parent), nextLocalGameId(0), tcpUserCo
             Qt::QueuedConnection);
 }
 
-Server::~Server() = default;
-
 void Server::prepareDestroy()
 {
     roomsLock.lockForWrite();
--- a/common/server.cpp
+++ b/common/server.cpp
@@ -53,9 +53,7 @@ Server::Server(QObject *parent) : QObject(parent), nextLocalGameId(0), tcpUserCo
             Qt::QueuedConnection);
 }
 
-Server::~Server()
-{
-}
+Server::~Server() = default;
 
 void Server::prepareDestroy()
 {
@@ -146,8 +144,8 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session,
     users.insert(name, session);
     qDebug() << "Server::loginUser:" << session << "name=" << name;
 
-    data.set_session_id(
-        databaseInterface->startSession(name, session->getAddress(), clientid, session->getConnectionType()));
+    data.set_session_id(static_cast<google::protobuf::uint64>(
+        databaseInterface->startSession(name, session->getAddress(), clientid, session->getConnectionType())));
     databaseInterface->unlockSessionTables();
 
     usersBySessionId.insert(data.session_id(), session);
@@ -158,9 +156,9 @@ AuthenticationResult Server::loginUser(Server_ProtocolHandler *session,
     Event_UserJoined event;
     event.mutable_user_info()->CopyFrom(session->copyUserInfo(false));
     SessionEvent *se = Server_ProtocolHandler::prepareSessionEvent(event);
-    for (int i = 0; i < clients.size(); ++i)
-        if (clients[i]->getAcceptsUserListChanges())
-            clients[i]->sendProtocolItem(*se);
+    for (auto &client : clients)
+        if (client->getAcceptsUserListChanges())
+            client->sendProtocolItem(*se);
     delete se;
 
     event.mutable_user_info()->CopyFrom(session->copyUserInfo(true, true, true));
@@ -240,9 +238,9 @@ void Server::removeClient(Server_ProtocolHandler *client)
         Event_UserLeft event;
         event.set_name(data->name());
         SessionEvent *se = Server_ProtocolHandler::prepareSessionEvent(event);
-        for (int i = 0; i < clients.size(); ++i)
-            if (clients[i]->getAcceptsUserListChanges())
-                clients[i]->sendProtocolItem(*se);
+        for (auto &client : clients)
+            if (client->getAcceptsUserListChanges())
+                client->sendProtocolItem(*se);
         sendIsl_SessionEvent(*se);
         delete se;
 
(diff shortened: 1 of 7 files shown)


common/server.h
```diff
@@ -187,8 +187,9 @@ private slots:
     void sendIsl_GameCommand(const CommandContainer &item, int serverId, qint64 sessionId, int roomId, int playerId);
     void sendIsl_RoomCommand(const CommandContainer &item, int serverId, qint64 sessionId, int roomId);
 
-    void addExternalUser(const ServerInfo_User &userInfo);
-    void removeExternalUser(const QString &userName);
+    virtual void addExternalUser(const ServerInfo_User &userInfo) = delete;
+    virtual void removeExternalUser(const QString &userName) = delete;
```
ctrlaltca: These looks like leftovers as they have been implemented in `Server_Room`, maybe they can just be deleted.

common/server.cpp
```diff
@@ -53,9 +53,7 @@ Server::Server(QObject *parent) : QObject(parent), nextLocalGameId(0), tcpUserCo
             Qt::QueuedConnection);
 }
 
-Server::~Server()
-{
-}
+Server::~Server() = default;
```
ctrlaltca: Can this be removed and `= default` added in the .h?

servatrice/migrations/servatrice_0024_to_0025.sql
```diff
@@ -0,0 +1,6 @@
+-- Servatrice db migration from version 24 to version 25
+
+ALTER TABLE `cockatrice_uptime` ADD COLUMN `mods_count` int(11) NOT NULL DEFAULT 0;
+ALTER TABLE `cockatrice_uptime` ADD COLUMN `mods_list` BLOB;
```
ctrlaltca: If you are storing text and not binary data, `TEXT` is a better choice.

servatrice/src/servatrice.cpp
```diff
@@ -74,11 +74,11 @@ Servatrice_GameServer::~Servatrice_GameServer()
     }
 }
 
-void Servatrice_GameServer::incomingConnection(qintptr socketDescriptor)
+void Servatrice_GameServer::incomingConnection(const int &socketDescriptor)
```
ctrlaltca: The base Qt class is declaring a `qintptr` argument (http://doc.qt.io/qt-5/qtcpserver.html#incomingConnection), i see no reason to change this.

servatrice/src/servatrice.cpp
```diff
@@ -515,9 +514,9 @@ QList<AbstractServerSocketInterface *> Servatrice::getUsersWithAddressAsList(con
 {
     QList<AbstractServerSocketInterface *> result;
     QReadLocker locker(&clientsLock);
-    for (int i = 0; i < clients.size(); ++i)
-        if (static_cast<AbstractServerSocketInterface *>(clients[i])->getPeerAddress() == address)
-            result.append(static_cast<AbstractServerSocketInterface *>(clients[i]));
+    for (auto client : clients)
+        if (dynamic_cast<AbstractServerSocketInterface *>(client)->getPeerAddress() == address)
+            result.append(dynamic_cast<AbstractServerSocketInterface *>(client));
```
ctrlaltca: A `client` is a `Server_ProtocolHandler *`, and the`AbstractServerSocketInterface` is a subclass of `Server_ProtocolHandler`.
Since we can guarantee that the down-cast is safe, `static_cast` is a better choice.
`dynamic_cast` could be a safer choice if we are not sure about the type of `client`, but still we should check the cast result before dereferencing it.
Also, `dynamic_cast` has a runtime performance cost.

common/server.h
ZeldaZach: Gotcha

servatrice/migrations/servatrice_0024_to_0025.sql
ZeldaZach: Never knew the diff, thanks!

servatrice/src/servatrice.cpp
ZeldaZach: I didn't change this, which was weird... Clion didn't like the qintptr :P Will readdress this

servatrice/src/servatrice.cpp
ZeldaZach: CLion didn't agree with this being static for some reason, but I'll listen

ZeldaZach:
Done!

servatrice/migrations/servatrice_0024_to_0025.sql
```diff
@@ -0,0 +1,6 @@
+-- Servatrice db migration from version 24 to version 25
+
+ALTER TABLE `cockatrice_uptime` ADD COLUMN `mods_count` int(11) NOT NULL DEFAULT 0;
+ALTER TABLE `cockatrice_uptime` ADD COLUMN `mods_list` TEXT;
+
+UPDATE `cockatrice_schema_version` SET `version` = 25 WHERE `version` = 24;
```
ctrlaltca: The build script is checking for an exact string here; the check fails because the table and column names are surrounded by ` quotes.
Quick fix: remove them

ZeldaZach: This has been done with other sql migrations... oh well

common/server.h
```diff
@@ -56,8 +56,8 @@ private slots:
 
 public:
     mutable QReadWriteLock clientsLock, roomsLock; // locking order: roomsLock before clientsLock
-    Server(QObject *parent = 0);
-    ~Server();
+    Server(QObject *parent = nullptr);
+    Server::~Server() = default;
```
ctrlaltca: The `Server::` qualifier is not needed here

servatrice/migrations/servatrice_0024_to_0025.sql
```diff
@@ -0,0 +1,6 @@
+-- Servatrice db migration from version 24 to version 25
+
+ALTER TABLE cockatrice_uptime ADD COLUMN mods_count int(11) NOT NULL DEFAULT 0;
+ALTER TABLE cockatrice_uptime ADD COLUMN mods_list TEXT;
+
+UPDATE cockatrice_schema_version SET version = 25 WHERE version = 24;
```
ctrlaltca: The checker script expects the exact string:
```
UPDATE cockatrice_schema_version SET version=25 WHERE version=24;
```
Please remove spaces around the equal signs, we'll fix the script later.

ZeldaZach:
@ctrlaltca I'm not sure why linux is failing... the function `join()` was introduced in qt4 so it should be in qt5.2.1...

servatrice/src/servatrice.cpp
```diff
@@ -566,6 +565,11 @@ void Servatrice::statusUpdate()
         return;
 
     const int uc = getUsersCount(); // for correct mutex locking order
+
+    const QList<QString> mods_info = getOnlineModeratorList();
```
ctrlaltca: use `QStringList` instead of `QList<QString>`, add the #include if needed

ZeldaZach: Seems silly as they should be equiv... but w/e i'll get that fixed

PR #3108, opened 2018-02-14
Braces left in for split mana costs in xml

Vafthrudnir:
## Related Ticket(s)
- Fixes #3017

## Short roundup of the initial problem
Split mana costs were not displayed correctly: without the braces `{G/W}{G/W}` looked like `G/WG/W`, which can be easily misinterpreted.

## Screenshots
split_mana_costs

PR #3217, opened 2018-04-29
braces are only left for split cards

Vafthrudnir:
Due to a mistake in #3108 all mana costs are shown in braces instead of only split costs. This PR fixes that.

## Screenshots
image

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -83,7 +83,7 @@ CardInfoPtr OracleImporter::addCard(const QString &setName,
         QStringList symbols = cardCost.split("}");
         QString formattedCardCost = QString();
         for (QString symbol : symbols) {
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Thought this was a feature. Whoops haha

Vafthrudnir:
Well, the previous modification was only meant to put braces around split mana costs, but if you think it's better I can change it back to that version.

The code must be modified either way, because in that case it's quite useless to look for split costs and then do nothing special with them.

Vafthrudnir:
Thanks for noticing that @tooomm! Fixed it in the previous commit.

ZeldaZach:
screenshot 2018-05-02 18 08 23
screenshot 2018-05-02 18 08 28
screenshot 2018-05-02 18 08 49

Vafthrudnir:
Have you run oracle again? This same version show up as this at me:
image
image

PR #3223, opened 2018-05-03
Extract the xml parser from carddatabase

### LINKED ISSUE #3201: Card database: multiple parser engines
It would be nice to take the "xml file loader and parser" logic out of the card database.
Creating a generic interface with a few methods that a card parser should implement, we could add support for multiple parser engines.
We currently support only xml files in the `cockatrice_carddatabase version="3"` format, but supporting multiple parser engines we could easily add support for other formats like eg. json, a different xml format, a folder with one card per file etc..

A parser engine should implement at least these methods:
  * "can i parse this file?": a method  that receives the full file path and detects if the parser is able to load that file (checking file extension, cockatrice_carddatabase version, etc..). The card database would just poll each parser until it founds one capable of loading the file;
 * "parse this file": a method that actually parses the file

A few things that needs to be sorted out while developing this:
 * how to handle conflicts (duplicated card names?)
 * wether to make the parser directly call `cardDatabase->addCard()`, use some sort of generic callback, or let the card database fetch cards form the engine at the end of the parse step

ctrlaltca:
## Related Ticket(s)
- Fixes #3201

See #3201

The xml parser has been extracted from the card database.
The card database now supports querying multiple parsers; by now the only parser available is the "old" xml file parser.

There should be no user-noticeable difference in card parsing, loading or anything else.

### DIFF
--- a/cockatrice/src/carddatabase.h
+++ b/cockatrice/src/carddatabase.h
@@ -401,6 +401,7 @@ protected:
      * The available card database parsers
      */
     QVector<ICardDatabaseParser *> parsers;
+
 private:
     CardInfoPtr getCardFromMap(const CardNameMap &cardMap, const QString &cardName) const;
     void checkUnknownSets();
--- a/cockatrice/src/carddatabase.h
+++ b/cockatrice/src/carddatabase.h
@@ -2,7 +2,6 @@
 #define CARDDATABASE_H
 
 #include <QBasicMutex>
-#include <QDataStream>
 #include <QDate>
 #include <QHash>
 #include <QList>
@@ -15,6 +14,7 @@ class CardDatabase;
 class CardInfo;
 class CardSet;
 class CardRelation;
+class ICardDatabaseParser;
 
 typedef QMap<QString, QString> QStringMap;
 typedef QMap<QString, int> MuidMap;
@@ -58,15 +58,15 @@ public:
     {
         return releaseDate;
     }
-    void setLongName(QString &_longName)
+    void setLongName(const QString &_longName)
     {
         longName = _longName;
     }
-    void setSetType(QString &_setType)
+    void setSetType(const QString &_setType)
     {
         setType = _setType;
     }
-    void setReleaseDate(QDate &_releaseDate)
+    void setReleaseDate(const QDate &_releaseDate)
     {
         releaseDate = _releaseDate;
     }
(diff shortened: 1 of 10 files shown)


cockatrice/src/carddbparser/carddatabaseparser.h
```diff
@@ -0,0 +1,22 @@
...
+#include <QString>
+#include <QIODevice>
+
+#include "../carddatabase.h"
+
+class ICardDatabaseParser : public QObject
+{
+public:
+    virtual ~ICardDatabaseParser() {}
+    virtual bool getCanParseFile(const QString &name, QIODevice &device) = 0;
```
ZeldaZach: `= 0;` vs `= delete;` ?

ZeldaZach:
You'll want to run the `./clangify.sh` file

cockatrice/src/carddatabase.h
```diff
@@ -398,11 +397,11 @@ class CardDatabase : public QObject
 
     LoadStatus loadStatus;
 
+    /*
+     * The available card database parsers
```
Daenyth: 👎 IMO we should avoid "what is this code" style comments unless it's very hard. In this case naming the variable `availableParsers` would be enough.

cockatrice/src/carddbparser/cockatricexml3.h
```diff
@@ -7,21 +7,25 @@
...
 class CockatriceXml3Parser : public ICardDatabaseParser
 {
-Q_OBJECT
-Q_INTERFACES(ICardDatabaseParser)
+    Q_OBJECT
+    Q_INTERFACES(ICardDatabaseParser)
 public:
-	CockatriceXml3Parser() {};
-    ~CockatriceXml3Parser() {};
+    CockatriceXml3Parser(){};
```
ZeldaZach: =default; ?

cockatrice/src/carddbparser/carddatabaseparser.h
ctrlaltca: I want a pure virtual function in order to make the class abstract (non-instanceable).
As far as I understand, with `= delete` the method is not callable, but the class can still be instanced.

ZeldaZach: https://stackoverflow.com/questions/24958270/why-cant-i-declare-a-pure-virtual-function-with-delete Yeah you're right. Didn't realize we could create a fully abstract class in C++ :D

ctrlaltca: Yep, in any other language i'd use an `interface`. In C++ you need some trickery to achieve the same result.

cockatrice/src/carddbparser/cockatricexml3.h
ctrlaltca: Yay!

Issue #2845, opened 2017-09-01
Revert order of Set List in Edit Sets


dev-id:
In Set Editor, my preference is to disable all promo sets. The easiest way to do this is to sort by Set type. However if I have previously re-ordered my sets, sorting will clear this information.

It should be an option to revert/undo a sort after enable/disabling sets.

Desired functionality:

1. Re-order sets. Ex: Default sort, oldest first, but move Limited Edition Beta above Limited Edition Alpha
2. Sort sets by type.
3. Disable promo sets.
4. Revert the sort from step 2 - returns order to order from after step 1 (oldest first, LEB above LEA)

tooomm:
I might want to link one of my other issue here: https://github.com/Cockatrice/Cockatrice/issues/2740

If you have the ability to search within sets, that could just filter the displayed results (without affecting your ordering).
This should help here as well:
You could then search for promo, disable all of them, clear the search and go back to your former sorted list without altering your actual set order.

tooomm:
Done in #3229

PR #3229, opened 2018-05-09
Search in "Manage sets" window

### LINKED ISSUE #2740: possibility to search in sets
With the crowing and crowing number of sets in `Manage sets..` it would be very helpful to have a search possibility there to show only sets matching a "text filter".

Same like in deck editor with the search bar!

Vafthrudnir:
## Related Ticket(s)
- Fixes #2740 and #2845 

It wasn't possible to filter or search sets in the `Manage Sets` window before.

- Search textbox added to filter set selection
- Filtering searches the entered expression in the `Long name`, `Set code`, and `Set type` columns, if either matches the row is shown
- Sorting by a property will not modify the original model order anymore, only explicit drag & drop or moving by button operations are usable for order modification
- If the view has been sorted by clicking on one of the column headers the `Disable sorting` button can be used to retrieve the previous order and re-enable manual order-setting

## Screenshots
image
image
image

### DIFF
--- a/cockatrice/src/setsmodel.cpp
+++ b/cockatrice/src/setsmodel.cpp
@@ -259,39 +259,30 @@ SetsDisplayModel::SetsDisplayModel(QObject *parent) : QSortFilterProxyModel(pare
 {
 	setFilterCaseSensitivity(Qt::CaseInsensitive);
 	setSortCaseSensitivity(Qt::CaseInsensitive);
-
-	loadedRowCount = 0;
-}
-
-bool SetsDisplayModel::canFetchMore(const QModelIndex &index) const
-{
-	return loadedRowCount < sourceModel()->rowCount(index);
 }
 
 void SetsDisplayModel::fetchMore(const QModelIndex &index)
 {
-	int remainder = sourceModel()->rowCount(index) - loadedRowCount;
-	int itemsToFetch = qMin(100, remainder);
+	int itemsToFetch = sourceModel()->rowCount(index);
 
-	beginInsertRows(QModelIndex(), loadedRowCount, loadedRowCount + itemsToFetch - 1);
+	beginInsertRows(QModelIndex(), 0, itemsToFetch - 1);
 
-	loadedRowCount += itemsToFetch;
 	endInsertRows();
 }
 
-bool SetsDisplayModel::filterAcceptsRow(int sourceRow, const QModelIndex & /*sourceParent*/) const
+bool SetsDisplayModel::filterAcceptsRow(int sourceRow, const QModelIndex & sourceParent) const
 {
-	return true;
-}
+	QModelIndex index0 = sourceModel()->index(sourceRow, SetsModel::SetTypeCol, sourceParent);
+	QModelIndex index1 = sourceModel()->index(sourceRow, SetsModel::LongNameCol, sourceParent);
+	QModelIndex index2 = sourceModel()->index(sourceRow, SetsModel::ShortNameCol, sourceParent);
 
-int SetsDisplayModel::rowCount(const QModelIndex &parent) const
-{
-	return qMin(QSortFilterProxyModel::rowCount(parent), loadedRowCount);
+	return (sourceModel()->data(index0).toString().contains(filterRegExp()) ||
+		sourceModel()->data(index1).toString().contains(filterRegExp()) ||
+		sourceModel()->data(index2).toString().contains(filterRegExp()));
 }
 
 bool SetsDisplayModel::lessThan(const QModelIndex &left, const QModelIndex &right) const
 {
-	QString searchTerm = "";
 	QString leftString = sourceModel()->data(left, SetsModel::SortRole).toString();
 	QString rightString = sourceModel()->data(right, SetsModel::SortRole).toString();
 
--- a/cockatrice/src/setsmodel.cpp
+++ b/cockatrice/src/setsmodel.cpp
@@ -252,3 +252,61 @@ QStringList SetsModel::mimeTypes() const
 {
     return QStringList() << "application/x-cockatricecardset";
 }
+
+
+
+SetsDisplayModel::SetsDisplayModel(QObject *parent) : QSortFilterProxyModel(parent)
+{
+	setFilterCaseSensitivity(Qt::CaseInsensitive);
+	setSortCaseSensitivity(Qt::CaseInsensitive);
+
+	loadedRowCount = 0;
+}
+
+bool SetsDisplayModel::canFetchMore(const QModelIndex &index) const
+{
+	return loadedRowCount < sourceModel()->rowCount(index);
+}
+
+void SetsDisplayModel::fetchMore(const QModelIndex &index)
+{
+	int remainder = sourceModel()->rowCount(index) - loadedRowCount;
+	int itemsToFetch = qMin(100, remainder);
+
+	beginInsertRows(QModelIndex(), loadedRowCount, loadedRowCount + itemsToFetch - 1);
+
+	loadedRowCount += itemsToFetch;
+	endInsertRows();
+}
+
+bool SetsDisplayModel::filterAcceptsRow(int sourceRow, const QModelIndex & /*sourceParent*/) const
+{
+	return true;
+}
+
+int SetsDisplayModel::rowCount(const QModelIndex &parent) const
+{
+	return qMin(QSortFilterProxyModel::rowCount(parent), loadedRowCount);
+}
+
+bool SetsDisplayModel::lessThan(const QModelIndex &left, const QModelIndex &right) const
+{
+	QString searchTerm = "";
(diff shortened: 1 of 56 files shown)


cockatrice/src/window_sets.cpp
```diff
@@ -173,6 +193,31 @@ void WndSets::actRestore()
     close();
 }
 
+void WndSets::actResetSort()
+{
+    view->header()->setSortIndicator(-1, Qt::DescendingOrder);
```
ctrlaltca: A minor one: you could use a define a explicit name for this magic number, eg.
```
#define SORT_RESET -1
```
and use the name instead of the value.

Vafthrudnir:
@ctrlaltca 
- Yeah, I felt like this was not the best name for the button, but nothing else came to mind mind... Though I like this "Restore order", it really makes it clear what it does.
- That workflow won't be bothered by this PR. You still can sort the sets by date, and enable/disable them however you like. You can even filter them by typing "Expansion" to hide any non-standard sets. What won't be possible is rearranging the set order, while the list is sorted, like this:
sort_and_rearrange

cockatrice/src/setsmodel.cpp
```diff
@@ -252,3 +252,37 @@ QStringList SetsModel::mimeTypes() const
...
+    int itemsToFetch = sourceModel()->rowCount(index);
+
+    beginInsertRows(QModelIndex(), 0, itemsToFetch - 1);
+
+    endInsertRows();
+}
+
+bool SetsDisplayModel::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
+{
+    QModelIndex index0 = sourceModel()->index(sourceRow, SetsModel::SetTypeCol, sourceParent);
```
ZeldaZach: These names aren't too good...

cockatrice/src/window_sets.cpp
```diff
@@ -53,16 +55,32 @@ WndSets::WndSets(QWidget *parent) : QMainWindow(parent)
     connect(aBottom, SIGNAL(triggered()), this, SLOT(actBottom()));
     setsEditToolBar->addAction(aBottom);
 
+    // search field
+    searchField = new QLineEdit;
+    searchField->setObjectName("searchEdit");
+    searchField->setPlaceholderText(tr("Search by card name"));
```
tooomm: Copy paste error? 😉

Vafthrudnir: Yup, fixed it in next commit 😄

Vafthrudnir:
@tooomm 
Thanks for the test! Made the "Enabled" column fixed sized, now it won't change width during filtering. Also added the `(x)` button you requested.

tooomm:
After this change, reordering of sets (drag&drop and arrow buttons) is only permitted if you have restored the default order (hitting the button or sort the most recent set to the top). It puzzled me at first why the buttons on the left are disabled and I couldn't enable them when selecting a set...

In general the `Manage sets` dialog a complex window which even gets a bit more complex now, with the very much needed and important search/filter feature.
We should look into decoupling general set management (search + enable) from actual order setting and the connected image load behavior.
People are always confused about that second part and never really understand it at first.
The little explanation we provide at the buttom doesn't help to much with that.
It might be really worth to make that a separate dialog or include it into the settings at one point.

Can you add a explaining hover hint for the restore button? @Vafthrudnir

cockatrice/src/window_sets.cpp
```diff
@@ -53,16 +55,33 @@ WndSets::WndSets(QWidget *parent) : QMainWindow(parent)
...
 
+    // search field
+    searchField = new QLineEdit;
+    searchField->setObjectName("searchEdit");
+    searchField->setPlaceholderText(tr("Search by set name, code, or type"));
+    searchField->addAction(QPixmap("theme:icons/search"), QLineEdit::LeadingPosition);
+    searchField->setClearButtonEnabled(true);
+    setFocusProxy(searchField);
+
+    resetSortButton = new QPushButton(tr("Restore order"));
```
tooomm: What about `Back to default view`.
That could stress that it's only the displayed view which is going to change, it won't reset your personalized order or restore (=overwrite) changes you made.

It could also show the "mode" you are currently in and highlight that.
`Mode: Picture load ordering` for the disabled button, and `Adjust card art display` for the clickable button.

But that might not yet be thought-out enough... 😄

tooomm:
>Could you help me coming up with something useful to show there?

Sure, what about e.g. `Revert displayed set order to your customized picture load priority` or `Show default `?

Vafthrudnir:
Thanks for the suggestion @tooomm!
In the latest commit I added the hint text and a new button that restores the original timeline order (new on top).
When the filtering is active (there's something written in the searchbox) this button gets disabled because it wouldn't make much sense to reset set order on only a filtered subset of the sets.
image
My concern with this solution is that the function of these two buttons isn't clearly shown by their names. Do you have any suggestions how to solve this issue?

tooomm:
Haha, that dialog is a really getting problematic. :D
But making it two won't solve it immediately if you think about about it for a bit.
I still think it would be better most likely. Nothing for this PR though.


>My concern with this solution is that the function of these two buttons isn't clearly shown by their names. Do you have any suggestions how to solve this issue?

Since appveyor is failing currently (they updated two days ago, maybe that's why?), I can't download a test version.

Vafthrudnir:
Yeah, I agree :smile: 
Would it be feasible to remove the `Restore order` button, and add that functionality to the column headers? So when a header is clicked, it first orders the sets in descending order, on second click orders it in ascending order, and the third click would restore the original order.
This way a label could pop up somewhere showing why the manual order-changing buttons are disabled, and telling the users how to re-enable them again (click on the column header twice/once). 
Although I can't really see any good substitute for the `Default order` button, something with that functionality should be implemented. What do you think?

Daenyth:
> when a header is clicked, it first orders the sets in descending order, on second click orders it in ascending order, and the third click would restore the original order.

If you can do that it would be ideal

Vafthrudnir:
Removed the `Restore order` button, and added the same functionality to the column headers (needs to be clicked a third time to disable sorting). When the table is sorted a warning is shown like this:
image

tooomm:
This is indeed way better and combines both features of the sets dialog quite nicely! Well done.

A few minor things:
- Can you place the warning message right below the list (above the buttons), or below the buttons?
I assume right below the list will feel and look best when it appears, and should be easier to recognize too. Maybe check both and tell us what works best.
Also, having the warning label or all text in red or within a box to make it more obvious is an idea.

- When you sort for a column and hit `Default order` afterwards it looks like everything is back to default (also the sorting), but the warning is still displayed

- It would be nice if hitting `Default order` will bring you back to the top of the list

tooomm:
Further observation:

I ran into an issue in combination with automatic spoiler updates / spoiler season:
If a new spoiler file is found you see the "first time user" dialog:


Then, no matter if you click `x` or `OK`, an **empty** "Manage sets..." window opens:


With the current stable release (2.5.1) I don't even see the "first time user" dialog at all?! 😄

Vafthrudnir:
The warning message is definitely more recognizable when it is placed above the buttons, but I think it looks strange there:
image
I'd prefer it under the buttons, above the other text:
image
What do you think?
Regarding the icons, I think they are missing from the screenshots due to the program running in debug mode, though I'm not sure, there might be something wrong with my environment. Hasn't really bothered me before 😄 

I also checked the manual spoiler update stuff, but I couldn't reproduce the error on my side - it shows the "Spoilers already up to date" message in the bottom right corner, but doesn't show the first time user window. Seems really strange, are you sure this PR causes it?

tooomm:
>The warning message is definitely more recognizable when it is placed above the buttons, but I think it looks strange there:

Yeah right. Alternatively the warning could be placed above the set list. Or the buttons could be above and the warning below the list.

>I also checked the manual spoiler update stuff, but I couldn't reproduce the error on my side - it shows the "Spoilers already up to date" message in the bottom right corner, but doesn't show the first time user window. Seems really strange, are you sure this PR causes it?

I  tried again with Client Version: 2.5.2-custom(16e17a2) (2018-05-23), the last commit in this pr. Same result.
If you manually open the manage sets dialog after hitting the update spoiler button in settings and a moment after the notification appears... is the dialog then bugged as well?

skwerlman:
You could surround the warning in a box with a slightly different background color to make it stand out where it is

tooomm:
Maybe I'm experiencing #3262?

skwerlman:
> Maybe I'm experiencing #3264?

What you have screenshotted is definitely #3264

tooomm:
>What you have screenshotted is definitely #3264

It should not be caused or related by this PR. :+1:
Weird though, that @Vafthrudnir can't reproduce it.

Vafthrudnir:
How about this?
image

tooomm:
Can somebody give this a try and double check the code again, please?

PR #3230, opened 2018-05-09
Cmake scripts refactoring; fix #3225

### LINKED ISSUE #3225: Building Servatrice only, causes dependency on Qt5LinguistTools
https://github.com/Cockatrice/Cockatrice/blob/4e10ce7473b925b24bce81d53c63cd5ad4644ce5/CMakeLists.txt#L146

I'm not a cmake expert, so I'm not clear on how to fix this issue.  However, if one *only* builds servatrice, then we don't need to check/build/install Qt5LinguistTools AFAIK.

I found this while modding the FreeBSD Port version to install dependencies based on user options of "Cockatrice + Oracle" or "Servatrice".

ctrlaltca:
## Related Ticket(s)
- Fixes #3225

The cmake components detection could be improved to avoid false warning messages.

The cmake components detection has been improved to avoid false warning messages. Also, it's making use of a little more modern cmake dialects with package COMPONENTS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,29 +130,23 @@ OPTION(UPDATE_TRANSLATIONS "Update translations on compile" OFF)
 MESSAGE(STATUS "UPDATE TRANSLATIONS: ${UPDATE_TRANSLATIONS}")
 
 IF(WIN32)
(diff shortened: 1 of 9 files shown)


ctrlaltca:
Thank you!

Issue #3112, opened 2018-02-14
Add basic bold/italic to card text


Qriist:
Gentle request to add `<b>` and `<i> `tags to card text fields in the xml. 

Currently, attempts to add the tags break cockatrice's interpreting of the xml and.. that's it, no more loading. If you don't want to rewrite the xml parser to account for <> in the field (which I image would be a little bit of a pain) then some other symbol or tag could be used, such as :bold:**bolded text**:bold: that would switch the bold/italics on and off as it renders that text.

Example:
```
<text>♥  :bold:HP:bold: 80

☣ :bold:Evolves from::bold: Dratini

:bold:Attacks:bold:
⚔ :italic::Healing Light:italic: ⚔ 
{Colorless, Colorless}
Remove 1 damage counter from each of your Pokémon that has any damage counters on it.

⚔ :italic:Protective Wave:italic: ⚔ 
{Colorless, Colorless, Colorless}
Flip a coin. If heads, during your opponent's next turn, prevent all effects of attacks, including damage, done to Light Dragonair.

☯ :bold:Resistance & Weakness:bold:
Resistant to Psychic: -30

↯ :bold:Retreat Cost::bold: {Colorless, Colorless}</text>
```

Rendered example:
♥  **HP** 80

☣ **Evolves From:** Dratini

**Attacks**
⚔ _Healing Light_ ⚔ 
{Colorless, Colorless}
Remove 1 damage counter from each of your Pokémon that has any damage counters on it.

⚔ _Protective Wave_ ⚔ 
{Colorless, Colorless, Colorless}
Flip a coin. If heads, during your opponent's next turn, prevent all effects of attacks, including damage, done to Light Dragonair.

☯ **Resistance & Weakness**
Resistant to Psychic: -30

↯ **Retreat Cost** {Colorless, Colorless}

Qriist:
Oh, and maybe some basic colors while we are at it.

:red:
:blue:
:green:
etc

ctrlaltca:
You can already insert html code inside a card text.
You just need to properly escape it as needed by xml (hint: https://www.freeformatter.com/xml-escape.html)
Example:
```
<text>&lt;b&gt;Defender&lt;/b&gt;
Formidable — {G}: Glade Watcher can attack this turn as though it didn't have defender. Activate this ability only if creatures you control have total power 8 or greater.</text>

```
Result:
[...]

PR #3231, opened 2018-05-10
Update cockatricexml3.cpp

### LINKED ISSUE #3227: Allow case insensitive tags in Cockatrice set files
Cockatrice refuses to load cards from set files which contain uppercase tags. For example, `<I>` where italic text exists on a card will fail, and the rest of the cards from the file is not imported.

Could Cockatrice allow for uppercase tags, or allow this be case insensitive?

ancestral:
Add support for child elements in a card’s text. Any valid markup that may be present will not cause the parser to stop. Instead, Cockatrice will not style unknown elements.

## Related Ticket(s)
- Fixes #3227, #3112

The Cockatrice set parser was stopping when it hit an unknown element in the text field. This included any kind of markup that may be present, including tags like`<i>` or `<b>` that may end up being in a card.

The parser will not stop on childElements found inside a card’s text and simply take in the card. (Any kind of tags it finds will of course be unstyled.)

## Screenshots
Screenshots can be provided upon request.

## Test file
The test file contains three test markup elements: `<B>` in the card “Dust Bunny”, `<i>` in the card “Furious Fusillade” and `<TranquilExpanse>` in the card “Tranquil Expanse”.
[03.DAD.xml.zip](https://github.com/Cockatrice/Cockatrice/files/1990164/03.DAD.xml.zip)

### DIFF
--- a/cockatrice/src/carddbparser/cockatricexml3.cpp
+++ b/cockatrice/src/carddbparser/cockatricexml3.cpp
@@ -150,7 +150,7 @@ void CockatriceXml3Parser::loadCardsFromXml(QXmlStreamReader &xml)
                 } else if (xml.name() == "pt") {
                     pt = xml.readElementText();
                 } else if (xml.name() == "text") {
-                    text = xml.readElementText();
+                    text = xml.readElementText(QXmlStreamReader::IncludeChildElements);
                 } else if (xml.name() == "set") {
                     QXmlStreamAttributes attrs = xml.attributes();
                     QString setName = xml.readElementText();


ZeldaZach:
Do you have test XMLs?

ancestral:
Thanks, just added one.

Nightfirecat:
Might be out-of-scope for this PR, but this sounds like a great thing to add unit tests for.

ctrlaltca:
This is a non-solution to a non-problem. Skipping over broken xml files is not a solution, the solution is to properly encode those xml files.
Also, why skip only broken entries inside the `<text />` tag and not the ones inside other tags?

ZeldaZach:
While I do agree that malformed XMLs should not be loaded (and possible errors should be thrown to the user), Maybe we should look into a script that will encode the data requested?

ancestral:
I misunderstood what you folks wanted out of the schema, and I misattributed this to what are actually bugs in the new parser.

Issue #1207, opened 2015-07-05
validate custom sets


aeosynth:
custom sets should be validated to prevent problems such as https://github.com/aeosynth/spoiler/issues/1

ZeldaZach:
We have an [XSD validation file](https://github.com/Cockatrice/Cockatrice/wiki/Custom-Card-Images#i-modified-my-cardsxml-how-can-i-ensure-it-wont-create-problems) with Cockatrice, but it seems to be out of date. I'll work on getting that up to date.

tooomm:
@ZeldaZach here is the correct link: [cards.xsd](https://github.com/Cockatrice/Cockatrice/blob/master/doc/cards.xsd)
Is it still outdated?

There is also a deck.xsd file...


I highly support a validation of the custom sets files. Just tested the `BFZ.xml` file provided [on reddit](https://www.reddit.com/r/Cockatrice/comments/3kg2bv/bfz_xml_file/).

The file is "working", but the result looks aweful:
![custom sets bfz](https://cloud.githubusercontent.com/assets/9874850/9826450/b510eb16-58da-11e5-8e0f-0a8557c4e5fc.png)

The xml file has stuff like this inside which might cause the issue:
```
<card>
      <name>~Awaken Cancel</name>
      <type>Instant</type>
      <text>Counter target spell.&#xD;Awaken 3&#xE2;&#x80;&#x94;4UU (If you cast this spell for 4UU, also put three +1/+1 counters on target land you control and it becomes a 0/0 Elemental creature with haste. It's still a land.)</text>
      <manacost>1&#xD;
                &#xD;
                    U&#xD;
                &#xD;
                    U</manacost>
      <tablerow>3</tablerow>
      <set picURL="">BFZ</set>
</card>
```

Not sure though if it could be detected with a validation against the xsd.

aeosynth:
@tooomm i updated my spoiler generator, here's my xml: https://gist.github.com/aeosynth/9c23d2d6de733a0ec118

i did have an issue with an `&bull;` nuking everything past that point. also i had to check the wiki to see that custom sets are required to have a `.xml` extension; a warning dialog would have been nice (eg 'unsupported file format')

ebbit1q:
dupe #2065

PR #3238, opened 2018-05-10
Update cards.xsd

### LINKED ISSUE #3237: Card Schema should use `xs:all` instead of `xs:sequence`
In the schema, I don’t think we need `xs:sequence`, especially with card elements. Sequence is needed only when the order of the elements is important, Given that we are talking about attributes within a card, this seems a little rigid. `xs:all` should be fine.

ancestral:
In the card schema, the order of the set’s and card’s elements are not important, so we can use `xs:all` here.

## Related Ticket(s)
- Fixes #3237 

The schema uses `xs:sequence` which, when validating a set file, will fail if it’s not in a specific order (for example, `cmc` must come before `type`). The order of a card’s elements is not important, so it can be changed to `xs:all`.

- cards.xsd

## Test file
This test file mixes up the order of some card elements.
[cards.xml.zip](https://github.com/Cockatrice/Cockatrice/files/1992249/cards.xml.zip)

### DIFF
--- a/doc/cards.xsd
+++ b/doc/cards.xsd
@@ -13,7 +13,7 @@
       <xs:all>
         <xs:element name="sets" minOccurs="0">
             <xs:complexType>
-                <xs:sequence>
+                <xs:all>
                     <xs:element name="set" maxOccurs="unbounded" minOccurs="0">
                         <xs:complexType>
                             <xs:all>
@@ -24,15 +24,15 @@
                             </xs:all>
                         </xs:complexType>
                     </xs:element>
-                </xs:sequence>
+                </xs:all>
             </xs:complexType>
         </xs:element>
         <xs:element name="cards" minOccurs="0">
             <xs:complexType>
-                <xs:sequence>
+                <xs:all>
                     <xs:element name="card" maxOccurs="unbounded" minOccurs="0">
                         <xs:complexType>
-                            <xs:sequence>
+                            <xs:all>
                                     <xs:element type="xs:string" name="name" />
                                     <xs:element name="set" maxOccurs="unbounded" minOccurs="1">
                                         <xs:complexType>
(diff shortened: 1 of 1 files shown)


ctrlaltca:
xs:all around the card element means that a card database can have from 0 to 1 card (https://www.w3.org/TR/xmlschema11-1/#element-all), that's not probably what we want.

Daenyth:
Is that bad actually? What would that mean? It could define a set with no cards? I'm trying to think whether that's useful or interesting

ctrlaltca:
After the same change you can have a card with no tag at all (no text, no name, no set).
Do we really need this at all?

ancestral:
Yeah I can work on an edit to require the name. Which other elements do you think should be required?

Psithief:
Hmm. We already accepted a change to the XSD from choice to sequence in the commit. Why change it back?https://github.com/Cockatrice/Cockatrice/pull/2787/files/f15295230fff758fc8a9c0716ed949087a107ff7

Psithief:
I'm somewhat sure (it's late) this was changed because xs:choice does not do what many think it does, and the tools required to make it perfect aren't available in xml.
There simply is no method to allow multiple optional/repeatable tags in any order without wrapping them in child elements.

ancestral:
It may be possible to have the first element, `<name>` as a sequence, which is required, followed by any other mandatory elements, and then have the other optional card elements follow suit in a choice.

FWIW right now nothing is using the schema in the Cockatrice app.

tooomm:
>FWIW right now nothing is using the schema in the Cockatrice app.

We track that issue in https://github.com/Cockatrice/Cockatrice/issues/1207 and especially https://github.com/Cockatrice/Cockatrice/issues/2065.

Also, [Cockatrice/Magic-Spoiler](https://github.com/Cockatrice/Magic-Spoiler) uses it.

ZeldaZach:
Any info on what we want to do with this?

Psithief:
Close it, because these changes don't work as intended.

It is already as good as we can make it.

PR #3242, opened 2018-05-12
Servatrice: make listening host configurable; fix #3241

### LINKED ISSUE #3241: Servatrice will not start if IPv4 or IPv6 is missing
https://github.com/Cockatrice/Cockatrice/blob/73fb9ee03e0b5d8f901da4b8f1f12a47ab47500c/servatrice/src/servatrice.cpp#L425

I'm guessing that the fix here should be to try to open a v4 address and v6 address separately and fail if *both* are missing.  Right now, if there is no v6 address available, servatrice will fail with:

```
Fri May 11 10:36:02 2018 "[pool 0] Opening database..."
Fri May 11 10:36:02 2018 Starting server on port 4747
Fri May 11 10:36:02 2018 gameServer->listen(): Error: "Protocol type not supported"

```

I found this in a FreeBSD jail with no ipv6 address configured.  This would probably happen on linux in a docker container or whatever the equivalent would be.  Or, one could disable IPV6 alltogether to see this behavior.

I added a v6 address and everything was working correctly. 

http://doc.qt.io/qt-5/qhostaddress.html
Probably what is needed is something like:
```
bool listeningV4 = FALSE;
bool listeningV6 = FALSE;
listeningV4 = gameServer->listen(QHostAddress::AnyIPv4, getServerTCPPort());
listeningV6 = gameServer->listen(QHostAddress::AnyIPv6, getServerTCPPort());
if (listeningV4 || listeningV6) 
            qDebug() << "Server listening.";
else {
            qDebug() << "gameServer->listen(): Error:" << gameServer->errorString();
            return false;
        }
```

ctrlaltca:
## Related Ticket(s)
- Fixes #3241 

Servatrice will always listen on all available ipv4 and ipv6 addressed on the host it's started from.
This can cause problem on hosts with an ipv6 stack but no ipv6 address available (#3241).
This is also problematic for people that only want to make the connection possible from localhost or that want to bind the server to a specific ip address on hosts with more than one network inteface / alias.

Two new server/host and server/websocket_host confgiuration options have been added. Each option accepts one ipv4 or ipv6 address or the magic word "any" to listen on all interfaces.
Both options default to "any", to mimic the old behavior.

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -438,7 +438,8 @@ bool Servatrice::initServer()
                                                                  servatriceDatabaseInterface->getDatabase(), this);
         websocketGameServer->setMaxPendingConnections(1000);
         QHostAddress webSocketHost = getServerWebSocketHost();
-        qDebug() << "Starting websocket server on host" << webSocketHost.toString() << "port" << getServerWebSocketPort();
+        qDebug() << "Starting websocket server on host" << webSocketHost.toString() << "port"
+                 << getServerWebSocketPort();
         if (websocketGameServer->listen(webSocketHost, static_cast<quint16>(getServerWebSocketPort())))
             qDebug() << "Websocket server listening.";
         else {
@@ -929,7 +930,7 @@ int Servatrice::getNumberOfTCPPools() const
 QHostAddress Servatrice::getServerTCPHost() const
 {
     QString host = settingsCache->value("server/host", "any").toString();
-    if(host == "any")
+    if (host == "any")
         return QHostAddress::Any;
     else
         return QHostAddress(host);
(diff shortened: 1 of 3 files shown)


servatrice/src/servatrice.cpp
```diff
@@ -934,6 +946,15 @@ int Servatrice::getNumberOfWebSocketPools() const
     return settingsCache->value("server/websocket_number_pools", 1).toInt();
 }
 
+QHostAddress Servatrice::getServerWebSocketHost() const
+{
+    QString host = settingsCache->value("server/websocket_host", "any").toString();
+    if (host == "any")
+        return QHostAddress::Any;
+    else
+        return QHostAddress(host);
```
Daenyth: What happens if the string in the config file is not a valid address?

ctrlaltca: The `QHostAddress` [ctor](https://doc.qt.io/qt-5/qhostaddress.html#QHostAddress-6) decides how to parse the string. From my tests it looks like it's dropping anything that's not numeric or a valid separator `.:`, and then parses the rest as an address.
Anyway, if an invalid address is detected, the call to listen() fails and servatrice aborts:
```
Starting server on host "0.0.0.1" port 4747
gameServer->listen(): Error: "The address is not available"
```

Daenyth: That seems fine. The error message isn't great but it's at least fail fast

PR #3243, opened 2018-05-13
connect dialog: button reworks

tooomm:
Unusual dialog window style

- Switch to default button usage and a more natural, os-specific look
- Move forgot pw to an icon only button right next to the pw field
That also removes problems like a too small button for some languages
- Cleanup of old unused button code + consistent naming

## Screenshots

- old look:
connect-old

- new look:
connect-new


Daenyth:
for icons maybe have a look at feather? https://feathericons.com/

tooomm:
Interesting service!

There is no perfect match... but I tried with the `unlock` picture of an open lock:
(Also realize the different refresh icon)
grafik
I really like the clean style.

Thoughts?

tooomm:
I reverted the icon to use the simple lock that we already have.

So this pr focuses on removing the old manually added buttons and uses default os-independent layout for buttons.
The forgot pw button is moved in that process and the icons will change before the next release.

The exact look of the icons can get changed with the new icon retheme as recently discussed.
I'll look into the icons in a different pr once @Daenyth has access to the license.
(I added two additional potential looks in my post above)

---

macOS builds are failing on Travis --> https://github.com/Cockatrice/Cockatrice/issues/3257

tooomm:
I directly call `reject()` from the button, so we don't need `actCancel` if I'm not mistaken.
The saving on cancel didn't make sense anyway. :)

PR #3249, opened 2018-05-16
ensure column ordering is consistent

skwerlman:
## Related Ticket(s)
- N/A

A database that had been continually migrated would have a different column ordering for some tables compared to a new database. This could cause subtle bugs in tools that depended on the ordering of the columns in the database.

- All tables will have their column order corrected to match servatrice.sql
- A note has been added to CONTRIBUTING.md regarding column ordering

## Screenshots

N/A


skwerlman:
okay, so codacy is complaining because the tool they use to lint sql files doesn't understand the `MODIFY` keyword. I think i can get around that by using `CHANGE` instead; aside from being more verbose than `MODIFY`, they're identical

ctrlaltca:
Also, nevermind about codacy, his sql checker is not able to parse the mysql dialect correctly.

PR #3250, opened 2018-05-17
0 values in power/toughness filters won't match non-zero values anymore

### LINKED ISSUE #3212: 0 matches values other than 0 in filters
<b>System Information:</b>
   

   

Client Version: 2.5.1 (2018-04-16)
OS: Windows 10 64-bit

________________________________________________________________________________________

   

   

The filter `Power AND 0` matches cards with a power of 0, as well as cards with odd powers like `*`, `2+*`, `*²`, `½`, `3½`, `?`, or `∞`.

This also applies to toughness filters.

IMO `Power AND 0` should match exactly, like other filters. If the current functionality is desired, it should be a separate filter, like `Effective Power` or something

Vafthrudnir:
## Related Ticket(s)
- Fixes #3212 

- `0` input on power/toughness filters will only return `0` values
- `relationCheck` method will only get called after simple string comparison failed
- due to their similar structure `acceptPower` and `acceptToughness` methods has been merged


cockatrice/src/filtertree.cpp
```diff
@@ -272,16 +272,22 @@ bool FilterItem::acceptLoyalty(const CardInfoPtr info) const
...
-}
-
-bool FilterItem::acceptToughness(const CardInfoPtr info) const
-{
-    int slash = info->getPowTough().indexOf("/");
-    return (slash != -1) ? (relationCheck(info->getPowTough().mid(slash + 1).toInt())) : false;
+    if (slash == -1)
+        return false;
+    QString valueString;
+    if (attr == CardFilter::AttrPow)
```
ZeldaZach: missing brackets </3

Vafthrudnir: right, sorry about that 😄 
BTW, would it be possible to modify clang-format to fix these type of mistakes?

tooomm: >BTW, would it be possible to modify clang-format to fix these type of mistakes?

clang-format can't since it's about modifying, not adding/removing stuff.
but clang-tidy could...
https://stackoverflow.com/questions/26111162/can-clang-format-add-braces-to-single-line-if-statements-etc

It has an option for [readability-braces-around-statements](http://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html).

ctrlaltca:
Code looks fine 👍

PR #3251, opened 2018-05-18
Fix appveyor build

Vafthrudnir:
## Initial problem
Appveyor failed due to a recent visual studio update:
>CMake Error at C:/projects/cockatrice-8k7y5/build/cmake_install.cmake:36 (file):
file INSTALL cannot find "C:/Program Files (x86)/Microsoft Visual
Studio/2017/Community/VC/redist/MSVC/14.13.26020/vc_redist.x64.exe".

## Solution
Increased VS version number to fix building.


tooomm:
Is there support for `* ` globes in the file path? If only the version number is changing... 

I've no idea how recent changes looked like.
But I remember @ctrlaltca said they changed the pattern once.

PR #3266, opened 2018-05-29
travis: remove unneeded debug builds on tags for faster release deployment

tooomm:
```
ONLY merge once https://github.com/travis-ci/travis-conditions/pull/1 is merged

Update:
They put a doc pr up, too: https://github.com/travis-ci/docs-travis-ci-com/pull/1884
```

Even on tagged commits which we only use to automatically deploy releases from master from within the release build, there was a debug build created at travis. This delays the deployment process itself.
When the tag is added, the same commit which is receiving the tag was build as debug from master branch before anyway. So we are not skipping something.

- always create debug build, except for tags

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: cpp
 
 cache: ccache
 
+conditions: v1
+
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


Daenyth:
Let's wait until upstream is merged

tooomm:
Ok, PR updated.

tooomm:
https://github.com/travis-ci/travis-conditions/pull/1#issuecomment-399078651 is merged/released

ZeldaZach:
@tooomm Merge when  you think this is ready

PR #3270, opened 2018-05-31
Fix archlinux compilation by fixing error-warnings `ignored qualifiers` and `implicit fallthrough`

### LINKED ISSUE #3240: Compilation on Archlinux fails because of missing break statements
## Problem
Compiling the newest Cockatrice fails under Archlinux. These are the cmake options I have used:

`-DCMAKE_BUILD_TYPE=Debug -DWITH_SERVER=1`

If I DONT set the Build Type to "Debug" cockatrice and servatrice compile just fine. Also, using
the Debug Build Type is working on Ubuntu 16.04 (GCC 5), it may have to do with the brand new GCC Archlinux is using (GCC 8)

This is the error: https://hastebin.com/ajahasacow.rb , somehow the switch statements without breaks are not liked anymore. ([-Werror=implicit-fallthrough=], warning treated as error)

## System Info
OS: Archlinux x64 Kernel: 4.16.8-1
Cockatrice: Newest
cmake: 3.11.1
GCC: 8.1.0
Qt: 5.10.1
Protobuf: 3.5.1
Zlib: 1.2.11

berserkingyadis:
## Related Ticket(s)
- Fixes #3240

Compiling the newest Cockatrice fails under Archlinux. These are the cmake options I have used:

`-DCMAKE_BUILD_TYPE=Debug -DWITH_SERVER=1`

Output:
https://hastebin.com/ajahasacow.rb
https://hastebin.com/esigarajun.vbs

- The code will be fixed, so that these warnings will no longer occur.
- Archlinux will be able to compile in Debug mode

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -317,7 +317,7 @@ void CardItem::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
             2 * QApplication::startDragDistance())
             return;
         if (zone->getIsView()) {
-            const ZoneViewZone *const view = static_cast<const ZoneViewZone *const>(zone);
+            const ZoneViewZone *view = static_cast<const ZoneViewZone *>(zone);
             if (view->getRevealZone() && !view->getWriteableRevealZone())
                 return;
         } else if (!owner->getLocal())
--- a/cockatrice/src/gamesmodel.cpp
+++ b/cockatrice/src/gamesmodel.cpp
@@ -196,6 +196,8 @@ QVariant GamesModel::headerData(int section, Qt::Orientation /*orientation*/, in
                     return tr("Age");
                 case Qt::TextAlignmentRole:
                     return Qt::AlignCenter;
+                default:
+                    return QVariant();
             }
         }
         case DESCRIPTION:
@@ -212,6 +214,8 @@ QVariant GamesModel::headerData(int section, Qt::Orientation /*orientation*/, in
                     return tr("Players");
                 case Qt::TextAlignmentRole:
                     return Qt::AlignCenter;
+                default:
+                    return QVariant();
             }
         }
         case SPECTATORS:
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -1029,7 +1029,7 @@ void TabGame::eventPlayerPropertiesChanged(const Event_PlayerPropertiesChanged &
     playerListWidget->updatePlayerProperties(prop, eventPlayerId);
 
     const GameEventContext::ContextType contextType =
-        static_cast<const GameEventContext::ContextType>(getPbExtension(context));
+        static_cast<GameEventContext::ContextType>(getPbExtension(context));
     switch (contextType) {
         case GameEventContext::READY_START: {
             bool ready = prop.ready_start();


Daenyth:
How bad is it to fix `ignored-qualifiers` rather than ignoring it?

berserkingyadis:
Ill post the `ignored-qualifiers` warnings in here, dunno how many there are.

berserkingyadis:
There are 2 warnings each.

`implicit-fallthrough`:

```
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/gamesmodel.cpp: In member function ‘virtual QVariant GamesModel::headerData(int, Qt::Orientation, int) const’:
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/gamesmodel.cpp:194:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
             switch (role) {
             ^~~~~~
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/gamesmodel.cpp:201:9: note: here
         case DESCRIPTION:
         ^~~~
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/gamesmodel.cpp:210:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
             switch (role) {
             ^~~~~~
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/gamesmodel.cpp:217:9: note: here
         case SPECTATORS:
         ^~~~
```
(only 2 in the whole project?)

`ignored-qualifiers`:

```
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/carditem.cpp: In member function ‘virtual void CardItem::mouseMoveEvent(QGraphicsSceneMouseEvent*)’:
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/carditem.cpp:320:89: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
             const ZoneViewZone *const view = static_cast<const ZoneViewZone *const>(zone);
                                                                                         ^
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/tab_game.cpp: In member function ‘void TabGame::eventPlayerPropertiesChanged(const Event_PlayerPropertiesChanged&, int, const GameEventContext&)’:
/home/bers/Projects/contrib/Cockatrice/cockatrice/src/tab_game.cpp:1032:81: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
         static_cast<const GameEventContext::ContextType>(getPbExtension(context));
                                                                                 ^

```

Seeing that there are only 2 warnings each, one could fix these imo.

(all warnings when compiling: https://pastebin.com/tfxaitET)

berserkingyadis:
I do not have the knowledge to fix the `ignored-qualifiers` warnings, can somebody help?

berserkingyadis:
I fixed the 2 warnings for `ignored-qualifiers` and added an exception for `implicit-fallthrough`, since break-less switches are used intentionally (https://github.com/Cockatrice/Cockatrice/issues/3240#issuecomment-388550785).

cockatrice/src/carditem.cpp
```diff
@@ -317,7 +317,7 @@ void CardItem::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
             2 * QApplication::startDragDistance())
             return;
         if (zone->getIsView()) {
-            const ZoneViewZone *const view = static_cast<const ZoneViewZone *const>(zone);
+            const ZoneViewZone *view = static_cast<const ZoneViewZone *>(zone);
```
berserkingyadis: I removed the `const` from the pointer since the type of `zone` is non-const pointer. https://github.com/Cockatrice/Cockatrice/blob/89ab7f1f67aba119377312cc538d29c01662476c/cockatrice/src/carditem.h#L23 that should have been the issue.

ctrlaltca:
c++17 adds a `[[fallthrough]]` statement that can be used to inform the compiler that the fallthrough is intentional.
Since we currently only support c++11, only compiler-specific statements can be used. Here's how mozilla is dealing with these, maybe it can be an inspiration:
```
#ifndef __has_cpp_attribute
#  define __has_cpp_attribute(x) 0
#endif

#if __has_cpp_attribute(clang::fallthrough)
#  define MOZ_FALLTHROUGH [[clang::fallthrough]]
#elif __has_cpp_attribute(gnu::fallthrough)
#  define MOZ_FALLTHROUGH [[gnu::fallthrough]]
#elif defined(_MSC_VER)
    /*
     * MSVC's __fallthrough annotations are checked by /analyze (Code Analysis):
     * https://msdn.microsoft.com/en-us/library/ms235402%28VS.80%29.aspx
     */
#  include <sal.h>
#  define MOZ_FALLTHROUGH __fallthrough
#else
#  define MOZ_FALLTHROUGH /* FALLTHROUGH */
#endif
```

usage
```
switch(test)
{
	case 1:
		break;
	case 2:
		MOZ_FALLTHROUGH
	default:
		break;
}
```

berserkingyadis:
where should I add the MOZ_FALLTHROUGH macro? a new header file?

ctrlaltca:
> a new header file?

Yep, that's the idea, and then `#include "newfile.h"` only where needed

berserkingyadis:
This fallthrough business is very strange, when looking at the 2 warnings it causes, there are like 100 other `case` blocks without break and they do not cause the warning. What is so special about the 2 lines that cause the warning?

I tried using the macro but it throws an error, I must be using it wrong.

Code:
```c++
switch (section) {
        case ROOM:
            return tr("Room");
        case CREATED: {
            switch (role) { //<- warning is here
                case Qt::DisplayRole:
                    return tr("Age");
                case Qt::TextAlignmentRole:
                    return Qt::AlignCenter;
            }
            MOZ_FALLTHROUGH
        }
        case DESCRIPTION:
            return tr("Description");
```

Above the `switch(role)` doesnt work either, does the `case` block have to be empty like in your example?


error:
```
home/bers/Cockatrice/cockatrice/src/gamesmodel.cpp: In member function ‘virtual QVariant GamesModel::headerData(int, Qt::Orientation, int) const’:
/home/bers/Cockatrice/cockatrice/../common/switch_fallthrough.h:11:27: error: attributes at the beginning of statement are ignored [-Werror=attributes]
 #  define MOZ_FALLTHROUGH [[gnu::fallthrough]]
                           ^
/home/bers/Cockatrice/cockatrice/src/gamesmodel.cpp:195:13: note: in expansion of macro ‘MOZ_FALLTHROUGH’
             MOZ_FALLTHROUGH
             ^~~~~~~~~~~~~~~
```

berserkingyadis:
That should do it, the macro should be tested on other platforms (clang, osx, windows)

Both warnings are fixed in the code, CMakeLists.txt remains untouched.

cockatrice/src/gamesmodel.cpp
```diff
@@ -197,6 +198,7 @@ QVariant GamesModel::headerData(int section, Qt::Orientation /*orientation*/, in
                 case Qt::TextAlignmentRole:
                     return Qt::AlignCenter;
             }
+            SWITCH_FALLTHROUGH;
```
ctrlaltca: I'm sorry i didn't look at the code before bringing up the MOZ_FALLTHROUGH topic, but here a simple `break;` would have avoided the warning without the need of adding a new header file.

berserkingyadis: If the other switch statement is the same I could just add a break there too. Remove the header file?

ctrlaltca: yes, sorry again for the wasted time

berserkingyadis: adding a `break` in these places makes the compiler think that the control could reach the end of the function `error: control reaches end of non-void function [-Werror=return-type]`

since the switch defaults to `return QVariant();` I think ill add it at the end of the function

berserkingyadis: another idea would be adding default cases to both the switches returning QVariant(), this is done alot in the function above

ctrlaltca: or, you can add `return QVariant()` in place of the breaks

berserkingyadis:
ok, added the defaults to the switches

tooomm:
Can you update your starting post? @berserkingyadis

berserkingyadis:
updated, should reflect the changes now.

ctrlaltca:
Merged, thank you!

PR #3287, opened 2018-06-17
add feature list requirement

ZeldaZach:
This adds the requirement for users to be pushed to the newest version. It's a temporary addition while the WW->RR transition occurs, as users need a bit of a 'push' to get off the links and we can disassociate WW from the program.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-version: 2.5.2-branch-{branch}-build-{build}
+version: 2.6.0-branch-{branch}-build-{build}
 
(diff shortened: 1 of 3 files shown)


skwerlman:
i wouldn't mind seeing a single, more general "minimum version" feature instead of simply adding a new protocol feature every time we want to do something like this.

ZeldaZach:
This is a changeable field, which in itself I guess we can declare minimum version. I don't really have a lot of time to look into incorporating a full minimum version requirement setup and I want to do a release this weekend

Daenyth:
@skwerlman The reason we went with feature list instead is because version number gets awkward if/when we have more than one client implementation (like the web client)

ZeldaZach:
This good to merge?

Issue #3255, opened 2018-05-22
icons8 TODO


tooomm:
Checklist *icons8*

- [x] Signup an account at https://icons8.com/
- [x] Tell support the account mail address to receive license
- [ ] Decide which icons we want to use (#3256)
- [ ] :shipit: 

CC @Daenyth

Daenyth:
do I email support or should you @tooomm? I used my daenyth@gmail to sign up

tooomm:
I feel like this ticket is not really needed anymore, the other one should be enough for discussion on particular icons.

PR #3289, opened 2018-06-17
connect dialog: new icons

tooomm:
For a first version we recycled already existing icons, which don't fit that perfectly and have different meanings in different dialogs.
Now with our [icons8 license](https://github.com/Cockatrice/Cockatrice/issues/3255) we can overhaul the look of Cockatrice.

- new icons only dedicated for a single purpose and to better symbolize what will happen
- uniform, simple and modern style

## Screenshots

before:
old

after:
new

Can somebody please verify the style on mac and/or linux?


ZeldaZach:
Looks good to me

PR #3300, opened 2018-06-19
NSIS: misc fixes

### LINKED ISSUE #1856: 64bit windows NSIS installer should prevent 32bit users from installing
...and print a useful error.

Right now users can install 64bit binaries on a 32bit version of windows without problems, but the application won't start and is not usable!

http://stackoverflow.com/questions/13229212/how-to-detect-windows-32bit-or-64-bit-using-nsis-script
(link credit goes to dae)

This would also allow to remove https://github.com/Cockatrice/Cockatrice/wiki/Troubleshooting#not-a-valid-win32-application-0xc000007b and to slim down the wiki.

### LINKED ISSUE #2035: Cockatrice-x64 installs on the wrong folder on Win64
Currently the NSIS script only supports installation on 32bit systems.
As a result, even if cockatrice is compiled as a 64bit app, it's installed in `c:\program files (x86)\Cockatrice` and the installer is using the 32bit registry.

### LINKED ISSUE #3288: Delete "vc_redist" file after install?
It looks like `vc_redist.x64.exe` persists in the install directory of Cockatrice after both got correctly installed.
If there is a problem with that file or the installation of the vc tools the users are better off reinstaling or redownloading all Cockatrice setup to avoid other problems.

Do we want to delete this file at the end of NSIS setup? I mean, it's 15MB and never used again.

### LINKED ISSUE #3301: Cockatrice is not available in the overview of installed apps
<b>System Information:</b>
Client Version: 2.6.0 (2018-06-17)
Build Architecture: 64-bit
Qt Version: 5.9.5
System Locale: de_DE

For me, Cockatrice is not listed in the list of `Apps and Features`.
So getting app details and uninstalling from there is not possible!

win10_apps

What about mac? Is there some equivalent?

ctrlaltca:
## Related Ticket(s)
- partially deals with #3288: delete vc_redist after use
- Fixes #1856: block 64-bit installation on 32-bit windows
- Fixes #2035: installation folder on 64-bit windows
- Fixes #3301: fix installer for windows 64bit 

NOTE: please test the resulting artifact from AppVeyor.


ctrlaltca:
I just tested the appveyor artifacts on both 32bit and 64bit windows 7 and on win10 64bit, and they works for me.

ctrlaltca:
I've added another commit addressing all the win64 issues related to NSIS. Let's see if everything works as expected..

tooomm:
>   Fixes #3288: delete vc_redist after use

❌ 
Still doesn't work for me, with the new binary it installs to the correct path, but the vc_redist file persists there. It was the first install ever to that directory:
untitled

What is a possible difference at your system?

>   Fixes #1856: block 64-bit installation on 32-bit windows

Can't test that.

>   Fixes #2035: installation folder on 64-bit windows

✔️ 
>   Fixes #3301: fix installer for windows 64bit

❌ 
Cockatrice still doesn't show in the list of installed apps.

Daenyth:
If you want to merge as-is and do 3301 in another PR I'm happy with that

ZeldaZach:
I agree with Gav's idea of fixing 3301 later

tooomm:
Ok, so it is not me or my setup.
Did you use 32bit windows maybe to also test the fix for #1856? @ctrlaltca

ctrlaltca:
> Fixes #1856: block 64-bit installation on 32-bit windows

schermata 2018-06-22 alle 09 10 13

Hold on, i'll investigate #3288 and #3301 further.

ctrlaltca:
> Fixes #3301: fix installer for windows 64bit

It should be working now.
schermata 2018-06-22 alle 10 08 24


So, only "#3288: delete vc_redist after use" is left.
The problem is, it works for me on all the vms I have available, so I have no idea on how to debug it.

Daenyth:
We can punt on it. It's very minor IMO

ctrlaltca:
I'd go on and merge this as-is.
The only fix not (always) working is #3288: "delete vc_redist after use", it's a minor one and it seems that even if NSIS actually executed the "delete" command, the vcredist file is still locked and can't be deleted.

tooomm:
>Fixes #3301: fix installer for windows 64bit

It looks like 32bit installs they show up as well!

tooomm:
Another note/question @ctrlaltca:

The version only shows the x.y.z string.
If you have a beta installed, that info is dropped even though we have it in our version string. Is that a windows limitation?
beta

ctrlaltca:
No, i think we can fit whatever string in there.

PR #3305, opened 2018-06-21
Tips close with the main window

### LINKED ISSUE #3199: Closing main Cockatrice window with "Tip of the Day" open does not close the app
<b>System Information:</b>
```
Client Version: 2.5.0 (2018-03-02)
Client Operating System: Windows 7 SP 1 (6.1)
Build Architecture: 64-bit
Qt Version: 5.9.3
```
________________________________________________________________________________________

See title. When the Tip of the Day window is open, closing the main window does not close the app.

Vafthrudnir:
## Related Ticket(s)
- Fixes #3199 

Tip of the Day window was left open after the main window have been closed. This only happened with the initial tip window, the one opened from the Help menu blocked the main window during running.

- Main window will be available while a the Tip of The Day window is open (if opened from help menu)
- Tip of the Day window will close as the main window closes
- It was possible to have two opened tip windows at the same time, if one was opened automatically on program start, and one was opened manually from the menu. This bug has been fixed as well.

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -22,7 +22,6 @@
 #include "QtNetwork/QNetworkInterface"
 #include "carddatabase.h"
 #include "dlg_settings.h"
-#include "dlg_tip_of_the_day.h"
 #include "featureset.h"
 #include "logger.h"
 #include "pixmapgenerator.h"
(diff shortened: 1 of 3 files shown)


ZeldaZach:
when you have your delete tip lines, I see no validation ensuring there is no double frees. Tip has no default so deleting it may cause a seg fault. 

Or I am completely wrong and didn’t analyze hard enough on my phones screen 😄

Vafthrudnir:
You are right there's no validation, but I think segfault is not possible the way it's written. 

Tip dialogs either get freed by the main window destructor, or before the user opens a new tip window from the help menu.

If it gets freed in the destructor, it's not possible to free it again, because that happens right before the end of the whole program. The other one avoids double frees by having a new allocation right after the `delete` operation making it safe to free up again later.

ZeldaZach:
I'm just trying to be safe; I like the if block you have.

PR #3307, opened 2018-06-24
Fix cards loading bug; fix #3262

### LINKED ISSUE #3262: Can't create related tokens
<b>System Information:</b>
  
Client Version: 2.5.2-custom(81803e2) (2018-05-25)
Build Architecture: 64-bit
Qt Version: 5.11.0
System Locale: en_US

  
The "create related card" option does not appear for any card, although the tokens are in my card database.

ctrlaltca:
## Related Ticket(s)
- Fixes #3262

After the card database files gets loaded, a few steps are executed:
 * resolve reverse-related cards
 * check if no set is enabled or new sets are available
Due to a change introduced in #3223, under windows/linux these steps could happen to be executed before card databases are fully loaded, creating the following problems:
 * reverse related cards (mostly tokens) are not available;
 * at startup, the "no sets found" popup gets shown

This trivial change ensures that cards are added at the right time, before the final steps gets executed.

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -388,8 +388,8 @@ CardDatabase::CardDatabase(QObject *parent) : QObject(parent), loadStatus(NotLoa
     availableParsers << new CockatriceXml3Parser;
 
     for (auto &parser : availableParsers) {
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Fixed Win10 too, we're good

PR #3316, opened 2018-07-03
force 2.6.1

ZeldaZach:
When we do the forced upgrade soon, it will be to 2.6.1 (addressing issues in 2.6.0)


tooomm:
Would a user with any 2.6.1-beta be able to join such a server already?
Or would it need to be the final 2.6.1 stable release @ZeldaZach?

skwerlman:
i'm already enforcing this on tetrarch; any 2.6.1 version (beta or otherwise) can connect

Daenyth:
What was the motivation for this again?

Issue #3294, opened 2018-06-18
Update ubuntu support to supported versions


Daenyth:
Right now we only can install on 14.04, which is pretty bad since it's 4 years out of date. I know there's something about travis that was in the way before, but we need to get this resolved.

I think worst case we should be able to use an Ubuntu 18.04 docker image, compile inside it, and get the resulting .deb out via a volume.

Daenyth:
Somewhat related #2898

Arc676:
Any update on this? Installing the `deb` package on 18.04 fails due to dependencies.
```
dpkg: dependency problems prevent configuration of cockatrice:
 cockatrice depends on libprotobuf9v5; however:
  Package libprotobuf9v5 is not installed.
 cockatrice depends on libqt5websockets5 (>= 5.3.0); however:
  Package libqt5websockets5 is not installed.
```

ctrlaltca:
`libqt5websockets5` is available for 18.04, you just need to install the package.
`libprotobuf9v5` is not available for 18.04, its version was bumped to `libprotobuf10`.
I guess the best bet you have is to compile Cockatrice: https://github.com/Cockatrice/Cockatrice/wiki/Compiling-Cockatrice-(Linux)#ubuntu-16xx17xx18xx

Daenyth:
No update, this needs someone to work on it

tooomm:
If someone can have a chat with me on Gitter I'm sure we can set something up since the basic construct is working but I fail with the right commands to get compiling inside docker to work.

Daenyth:
@tooomm can you open a WIP PR?

ctrlaltca:
Looks like appveyor now supports building on Linux, too, and they have Ubuntu 18.04 Bionic Beaver images available: https://www.appveyor.com/docs/linux-images-software/

tooomm:
Yes, I mentioned that in Gitter already. Though, @Daenyth's preference was to not use it because it puts even more stress on AppVeyor (which builds already very slowly), and supports no parallel jobs for free unlike travis...

I've no idea if it would have 1x windows AND 1x linux in parallel maybe?

Daenyth:
I think we'd accept an update to the config if someone does it, but I am concerned about the automatic github release thing we use for auto-update. I know appveyor works differently than the others, so idk.

Xeddius:
As a fellow linux user I too would like this awesome program to be updated for 18.04 and Linux Mint 19 (also 18.04 based.).
[...]

PR #3320, opened 2018-07-11
travis: switch to xenial

tooomm:
## Related Ticket(s)
- Related #https://github.com/Cockatrice/Cockatrice/issues/3294

Travis uses an old build image with a very outdated os version containing old packages.
Since we use that for deployment and to provide a .deb download a more recent one is highly appreciated.

- use Ubuntu Xenial (16.04 LTS) over Trusty (14.04 LTS)
- brings Qt from 5.2.1 to 5.5.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ matrix:
   fast_finish: true
   include:
   - os: linux
-    dist: trusty
+    dist: xenial
     group: stable
     env: BUILDTYPE=Debug
     if: tag IS NOT present
   - os: linux
-    dist: trusty
+    dist: xenial
     group: stable
     env: BUILDTYPE=Release
     if: (branch = master AND NOT type = pull_request) OR tag IS present


Daenyth:
I think we'll also need to update the website once this is merged so we tell people the right version

ctrlaltca:
This is great, but did they announce support for it officially?
The last word I read is this: https://github.com/travis-ci/travis-ci/issues/7260#issuecomment-354350596
> As you saw, we are working on dist: xenial. It may or may not work, depending on what we are working on. It is not supported in any way, and we apologize in advance that you may experience problems from time to time. We will work on various issues as they come up, but we are unable to offer any support.

I'd suggest merging this anyway in order to test it.

ctrlaltca:
I guess Appveyor is stuck because the change only touched .travis.yml and we explicitly skip appveyor builds when those files are changed: https://github.com/Cockatrice/Cockatrice/blob/master/.appveyor.yml#L16

Merge forced.

tooomm:
>I think we'll also need to update the website once this is merged so we tell people the right version 

We should wait for the next release with that update, before there will be no downloadable files affected
--> https://github.com/Cockatrice/Cockatrice/issues/3322

<br>

>This is great, but did they announce support for it officially?

No official word on it still @ctrlaltca 
I realized unofficial support months ago already and I felt like waiting even longer makes not too much sense for us. It has been tested by a bunch of people for a while now and should work ok for us.
Let's try and check with the next beta. 👍 

<br>

>I guess Appveyor is stuck because the change only touched .travis.yml and we explicitly skip appveyor builds when those files are changed:

Exactly. I already talked to their support but the current behavior is intentional and unlikely subject to change.
We can work around that though with a little custom script that does the same check and exits the build with a code of our choice.
That way such builds are marked *ok* without the long waiting time of a finished appveyor build... especially if changes are unrelated. This would also remove the admin requirement for such merges! 👍 
If you are interested we can have a chat about something like that on gitter and how it could look/work.
I've some ideas and can even provide a powershell example.

Issue #3005, opened 2018-01-10
remove all qt version guards in the code


tooomm:
A lot of
```
#if QT_VERSION >= 0x050200
#if QT_VERSION >= 0x050300
#if QT_VERSION > 0x050300
#if QT_VERSION >= 0x050400
...
```

etc. exist in the code. (also double check `#ifdef`'s)

Version code explanation: https://doc-snapshots.qt.io/qt5-dev/qtglobal.html#QT_VERSION

Right now we build on
- Qt 5.9 for windows (appveyor)
- Qt 5.11 for mac (travis)
- Qt 5.2 for ubuntu (travis) (no newer image available, https://packages.ubuntu.com/trusty/qt5-default)

Once travis updates their 14.04 LTS (trusty) image to 16.04 (xenial) we can remove the guards in the code as [xenial contains Qt 5.5](https://packages.ubuntu.com/xenial/qt5-default).
([18.04 LTS Bionic would have 5.9.5](https://packages.ubuntu.com/bionic/qt5-default))

Also, update the minimum requirements for Qt then and document them [here](https://github.com/Cockatrice/Cockatrice/blob/master/README.md#build--).
Append to Qt dependency: `Qt (≥ v5.5.x)`

PR #3328, opened 2018-07-14
remove Qt guards

tooomm:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/3005

Outdated guards in code due to different Qt requirements

- define a reasonable, uniform base version
- remove guards
- hint for needed min. Qt version

### DIFF
--- a/cockatrice/src/logger.cpp
+++ b/cockatrice/src/logger.cpp
@@ -2,14 +2,13 @@
 #include "version_string.h"
 #include <QDateTime>
 #include <QLocale>
+#include <QSysInfo>
 #include <iostream>
 
 #define LOGGER_MAX_ENTRIES 128
 #define LOGGER_FILENAME "qdebug.txt"
 
-#if QT_VERSION >= 0x050400
-#include <QSysInfo>
-#endif
+
 
 Logger::Logger() : logToFileEnabled(false)
 {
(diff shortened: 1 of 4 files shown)


cockatrice/src/logger.cpp
```diff
@@ -104,10 +101,7 @@ QString Logger::getSystemArchitecture()
 
 QString Logger::getClientOperatingSystem()
 {
-#if QT_VERSION >= 0x050400
     return QSysInfo::prettyProductName();
-#endif
-
     return {};
```
ZeldaZach: Can be removed

tooomm:
Do we want to put all that min requirement hints in the README?
https://github.com/Cockatrice/Cockatrice/tree/tooomm-guards#build--

We didn't have it before and we would need to maintain them...
Aren't the people compiling aware of such things or doesn't CMake etc. throws helpful error messages if
- `cmake_minimum_required(VERSION 3.1)` 
https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt

- or `FIND_PACKAGE(Qt5Core 5.0.3 REQUIRED)`
https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt#L142

are not matched?

Issue #998, opened 2015-04-21
missing 'trice icon for several dialog windows


tooomm:
https://github.com/Cockatrice/Cockatrice/pull/840#issuecomment-81722310

Maybe other dialog windows than the listed ones are affected as well:
- `Show this user's games`
  (as poixen mentioned in his comment, if you open this and close Cockatrice, the user's games window won't close with the app)
- `Load deck from server`
- `Create token`
- `Game information`
- `Manage sets...`
- `Edit tokens...`
- `Reveal top cards of library` the one, where you put the amount of cards you want to reveal
  (maybe other reveal/show dialogs as well or where you get asked about how many (dice e.g.))

no icon:
![dialog window-noicon](https://cloud.githubusercontent.com/assets/9874850/7261257/96c03dea-e870-11e4-8712-8cc1bc385e6f.png)

should be:
![dialog window-withicon](https://cloud.githubusercontent.com/assets/9874850/7261258/9bf84f6e-e870-11e4-80cb-61e89d0170ab.png)

---
- `Die roll` got fixed: #1065

poixen:
The reason that there is no icon is that the window has no attachment to a parent. Usually in the window constructor you can pass a pointer to the parent. Where the code is though, it is hard to determine who the parent is. 

The lack of a parent causes 2 issues:

 + no icon
 + window will not close when parent does (if you open a parent-less window then close the app the window will remain)

tooomm:
Still exists in e50c201 (2016-05-31)
I was able to close the app itself while the `Edit sets...` dialog box was open.
Weird part is that if I enable a new set and hit `OK` I get th confirmation message and everything closes as usual. When Cockatrice gets started again for the next time the newly enabled set is enabled. So while the app itself is closed some components can continue to run and even write changes to files...

PR #3329, opened 2018-07-14
fix: cockatrice quits properly when the window is closed

amarzial:
Added a call to QApplication::quit() when all the program's windows are closed
This makes sure that the application exits as expected.

Solves an issue on Linux that after closing and opening cockatrice again, some processes of the first instance were still running.

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -88,6 +88,8 @@ int main(int argc, char *argv[])
 {
     QApplication app(argc, argv);
 
+    QObject::connect(&app, SIGNAL(lastWindowClosed()), &app, SLOT(quit()));
+
     qInstallMessageHandler(CockatriceLogger);
     if (app.arguments().contains("--debug-output"))
         Logger::getInstance().logToFile(true);


ctrlaltca:
Hi, and first thanks for the PR!

A Qt application is supposed to quit by itself when the last main (parentless) window is closed (unless [quitOnLastWindowClosed](http://doc.qt.io/qt-5/qguiapplication.html#quitOnLastWindowClosed-prop) is disabled).
This is complicated by the [Qt::WA_QuitOnClose](http://doc.qt.io/qt-5/qt.html#WidgetAttribute-enum) attribute, that is not set for some types of top level windows, as stated in the [QWidget::close()](http://doc.qt.io/qt-5/qwidget.html#close) documentation:

> By default this attribute is set for all widgets except transient windows such as splash screens, tool windows, and popup menus.

Back on your issue, there's something that is blocking the application from quitting, probably a (hidden?) top level window.
Qt's built-in [quit on last window closed functionality](http://code.qt.io/cgit/qt/qtbase.git/tree/src/widgets/kernel/qapplication.cpp#n2904) is similar to your fix, but your fix skips the check for `Qt::WA_QuitOnClose`, thus I'm guessing that the window that is blocking the application from quitting correctly doesn't have that specific attribute set.

Now, it would be nice to know what window (tip of the day dialog? a popup menu? somethign else?) is actually causing the issue and fix it by correctly closing it or adding the `Qt::WA_QuitOnClose` to it.
In the meanwhile, your PR is acceptable since it workaround the issue in  quite clean way.

ZeldaZach:
The issue is if you’re viewing a users profile, what games they’re in, or Tip of the day from my tests a few weeks ago

tooomm:
There are several dialogs that are not wired correctly I think...
The `Manage sets...` dialog for example stays open after closing the main window, too!

I assume it affects all windows that don't show the little Cockatrice app icon in their title bar.

Linking https://github.com/Cockatrice/Cockatrice/issues/998

tooomm:
Any news here? What's the status and what do we want to do with this PR?

Is it an improvement already and can be merged with further tweaks and discussions postponed to issues and later PR's? @Daenyth @ZeldaZach 

@amarzial Are you still up to make adjustments?

amarzial:
@tooomm I tried to set the ```Qt::WA_QuitOnClose``` for all the top level windows as @ctrlaltca suggested. 
I used ```QApplication::topLevelWidgets()``` to get all the top level windows, but it still doesn't work.

Daenyth:
If this is an improvement on the status quo we should merge it. If there's more to do still we can amend the ticket or make a new one

ctrlaltca:
> If this is an improvement on the status quo we should merge it

Agreed

PR #3331, opened 2018-07-16
NSIS: Show hint when user installs 32bit Cockatrice on 64bit Windows

### LINKED ISSUE #3330: NSIS: Include hint for 32bit installs on 64bit windows
With #3300 we have a warning that prevents users on 32bit windows to install 64bit versions.

Would it be possible to extend that/build up on that?
I'm thinking about something the other way around, so that we somehow get users to switch over to 64bit and eventually get to drop 32bit sooner or later.

Same like NSIS detects the OS type, and then showing a warning for 64bit users trying to install a 32bit version with a hint and link to our website for downloading the correct version for their system.

32bit os + 64bit Cockatrice binary --> Error (done)
64bit os + 32bit Cockatrice binary --> Hint + DL link

Most users simply use the client updater and not even realize they are using a "wrong" version since the updater keeps them on the same bit-ness and doesn't give a choice or even tell them.
With a hint + link users are way more likely to switch (if we don't want to force them with the updater as discussed once), and we finally get to know how many still use/need a 32bit build and when we can safely deprecate it in the future.

Having that logic within NSIS sounds better than a solution with the in-client updater to me because people might get the binary files elsewhere or people direct link to old releases or share outdated binaries... NSIS would still catch them.

tooomm:
## Related Ticket(s)
- Fixes #3330

Most users simply use the client updater and not even realize they are using a "wrong" version since the updater keeps them on the same bit-ness and doesn't give a choice or even tell them.

- Show a hint + name webpage url as download source when installing 32bit on 64bit system
- Other install variants are unchanged
- Exchange warning for error icon on other dialog

Having that logic within NSIS is actually better than a solution with the in-client updater, because people might get the binary files elsewhere or people direct link to old releases or share outdated binaries...
NSIS still catches these cases.

Now, we finally get a better understanding of how many user still use/need a 32bit build and when we can safely deprecate it in the future.

>My preferred solution would have been a button that brings you to our webpage if the users wants to (via `ExecShell "open" "https://cockatrice.github.io"`), but I couldn't manage to provide these two functions next to each other or come up with another approach:
>- User wants to cancel the installer and open our webpage to download the "right" 64bit version
>- User skips the hint and still wants to install the "wrong" 32bit version
>
> It looks like one needs to write a custom plugin to alter the button labels, have in-text clickable urls etc.

## Screenshots

v3

### DIFF
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -59,7 +59,8 @@ ${Else}   #NSIS 32bit
         We advise you to use the 64-bit installer instead." \
         IDOK ok IDCANCEL cancel
         ok:
-        cancel: Abort
+	    ExecShell "open" "https://cockatrice.github.io/"
+	    Abort
     ${EndIf}
     StrCpy $NormalDestDir "$ProgramFiles\Cockatrice"
 ${EndIf}
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -54,7 +54,9 @@ ${If} ${NSIS_IS_64_BIT} == 1
     SetRegView 64
 ${Else}   #NSIS 32bit
     ${If} ${RunningX64}
-        MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION "You are about to install a 32-bit version of Cockatrice on a 64-bit Windows system. \nWe advise you to use the 64-bit installer instead."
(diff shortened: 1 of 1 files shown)


ZeldaZach:
@tooomm If you've tested this on 32/64, feel free to merge

tooomm:
Yep, for me it works with both installers as expected.

PR #3258, opened 2018-05-23
Fix compilation on Qt 5.11; fix #3257

ctrlaltca:
## Related Ticket(s)
- Fixes #3257

## Short roundup of the initial problem
In the newly-released Qt version 5.11's cmake modules some deprecated methods used to link Qt libraries have been removed

## What will change with this Pull Request?
Switched to the new methods. I'm unsure whether this will work on on ancient Qt version.
**EDIT**: it works with CMake >= 2.8.11 and Qt >= 5.1, so we should be fine.
In CMakeLists.txt we require Qt 5.0.3, so we should probably raise that number, but i have no way to test all the old Qt versions on linux right now.

PR #3332, opened 2018-07-16
[DO NOT MERGE] CMake: minimum requirements

### LINKED ISSUE #3259: raise qt minimum requirement in CMakeLists.txt
From https://github.com/Cockatrice/Cockatrice/pull/3258#issue-189923329:
>In CMakeLists.txt we require Qt 5.0.3, so we should probably raise that number, but i have no way to test all the old Qt versions on linux right now.

<br>

Linking https://github.com/Cockatrice/Cockatrice/issues/3005

tooomm:
## Related Ticket(s)
- Fixes #3259
- Related https://github.com/Cockatrice/Cockatrice/pull/3328 and #3005

As mentioned in https://github.com/Cockatrice/Cockatrice/pull/3258#issue-189923329:
>Switched to the new methods. I'm unsure whether this will work on on ancient Qt version.
**EDIT:** it works with CMake >= 2.8.11 and Qt >= 5.1, so we should be fine.
In CMakeLists.txt we require Qt 5.0.3, so we should probably raise that number, but i have no way to test all the old Qt versions on linux right now.

- Qt min req: 5.5.0
(Xenial has our lowest Qt version currently @ 5.5.1 / mac @ 5.11.1 / win @ 5.9.5 [5.11.1 possible])
Feature wise our min. requirement should be Qt5.4

- list (and maintain!) Qt and CMake (others too?) min. req. in README
**That's up for discussion**

---

This needs tests on older Qt versions!


tooomm:
Former post in https://github.com/Cockatrice/Cockatrice/pull/3328

>Do we want to put all that min requirement hints in the README?
https://github.com/Cockatrice/Cockatrice/tree/tooomm-guards#build--
>
>We didn't have it before and we would need to maintain them...
Aren't the people compiling aware of such things or doesn't CMake etc. throws helpful error messages if
>- `cmake_minimum_required(VERSION 3.1)` 
https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt#L9
>
>- or `FIND_PACKAGE(Qt5Core 5.0.3 REQUIRED)`
https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt#L142
>
>are not matched?

ZeldaZach:
My suggestion is CMAKE 3.1, QT 5.5

Not sure what you're asking

tooomm:
>Not sure what you're asking

Do we want to put these min requirement hints in the README at all?

- We didn't have such an info before
- We need to maintain them which people will forget, and we end up with old/wrong info there again
- CMake tells the user at compilation time already what they need if they don't match it

tooomm:
#3375 merged instead

PR #3336, opened 2018-07-18
travis: more cleanup

tooomm:
Unneeded package installs listed in travis.yml

- *cmake* and *clang* (clang-format is part of it) are already available in the Travis Linux build image:
https://docs.travis-ci.com/user/build-environment-updates/2017-12-12/#Changed
The logs now also print their version numbers.
- *bc* used to be included in the newest version, too
https://travis-ci.org/Cockatrice/Cockatrice/jobs/405156717#L428
- *clang-format* is never used on mac builds (we only check on Linux), remove its installation
- Added links to the related build image changelogs for each ci service
- `conditions` config can be removed as it's the default now
- Add job names
- Remove `fast finish`, which only has effects when allowed failures are configured

I tried to remove the other apt installs which should be already available in the Xenial image regarding this list: https://packages.ubuntu.com/xenial/allpackages (long load time!)
But it didn't work out. 😃 

## Screenshots
job names

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -97,6 +97,7 @@ build_script:
         $protodir = "c:\protobuf-release"
         $protoc = "c:\protobuf-release\bin\protoc.exe"
         $mysqldll = "c:\Program Files\MySQL\MySQL Server 5.7\lib\libmysql.dll"
+        cmake --version
         cmake .. -G "$env:cmake_generator" -T "$env:cmake_toolset" "-DCMAKE_PREFIX_PATH=c:/Qt/$env:qt_ver;$protodir;$zlibdir;$openssldir" "-DWITH_SERVER=1" "-DPROTOBUF_PROTOC_EXECUTABLE=$protoc" "-DMYSQLCLIENT_LIBRARIES=$mysqldll"
     - msbuild PACKAGE.vcxproj /p:Configuration=Release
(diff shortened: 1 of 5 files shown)


ctrlaltca:
If `clang-format` is already available, you can also remove the lines used to import the apt source for the llvm repository: https://github.com/Cockatrice/Cockatrice/pull/3336/files#diff-354f30a63fb0907d4ad57269548329e3R36
`clang-format` is a bit picky, since different versions can need different config files and different report errors, so you happen to know what version in installed? Ubuntu should have 6.0-41~exp4 installed by default

tooomm:
>`clang-format` is a bit picky, since different versions can need different config files and different report errors, so you happen to know what version in installed? Ubuntu should have 6.0-41~exp4 installed by default

`clang 5.0.0` is installed, so it's clang-format 5.0.0, too:
https://travis-ci.org/Cockatrice/Cockatrice/jobs/411445789#L2101

I guess you are talking about a newer Ubuntu than the 16.04 travis still uses. ;P

ctrlaltca:
Oh, you are right, i always forget how outdated the travis arch is.

tooomm:
@ctrlaltca Can you revisit and merge this PR, please?

I want no other CI stuff to lie around before looking into further changes.

PR #3337, opened 2018-07-18
Manage sets window: "Enable all/selected" button toggling never disabled

### LINKED ISSUE #3314: Regression in Manage Set Dialog
<b>System Information:</b>
  
Client Version: 2.6.0 (2018-06-17)
Build Architecture: 64-bit
Qt Version: 5.9.5
System Locale: en_US
  
In the "Manage sets" dialog, you have the options to "Enable all sets" and "Disable all sets."  When you have multiple sets selected, these options change to "Enable selected sets" and "Disable selected sets."  This used to be true whether the sets were in default order or sorted by one of the columns.

However, with this release when you have the sets sorted by any of the columns, for example "Set Type," the buttons will no longer toggle between "all" and "selected" depending on whether you have sets highlighted.  Instead, if you had no sets highlighted before sorting by column it will stay at "Enable all sets" and "Disable all sets" even when you subsequently highlight one or more sets, and if you have one or more sets highlighted before sorting the opposite happens.

Vafthrudnir:
## Related Ticket(s)
- Fixes #3314 

When the view was sorted in the manage sets window, the "Enable/Disable All" buttons never toggled to "Enable/Disable selected" buttons, even when more sets were selected.

- Selected sets can be enabled/disabled in sorted view as well


tooomm:
Can somebody have a look at the code here, please?

Issue #3082, opened 2018-02-05
Remove former file name pattern from in-client updater code


tooomm:
There is our old file pattern still baked into the updater code:
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/releasechannel.cpp#L34-L73

File names are now always:
- Windows (64bit): `filename-win64.exe`
- Windows (32bit): `filename-win32.exe`
- macOS: `filename.dmg`
- Ubuntu: `filename.deb`

(in comparison to stable releases, *filename* of betas have a `-beta` or `-beta.X` tag attached)

So there is no need to check for `x86` and `x86_64` any longer!

tooomm:
Actually, is this `betaEnd` and `exeBetaName` needed alltogether?
See https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/releasechannel.cpp#L52-L72

We don't use such a pattern like `-x86_64_qt5` for any releases anymore.
Don't we differentiate via the download urls?
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/releasechannel.cpp#L11-L17

PR #3338, opened 2018-07-20
updater: remove check for now unused beta file endings

tooomm:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/3082

I went forward with my unanswered assumptions from #3083 as that's how I understand it:
>File names are now always:
>
>      Win (64bit) --> filename-win64.exe
>      Win (32bit) --> filename-win32.exe
>      macOS --> filename.dmg
>      Ubuntu --> filename.deb
>
>(in comparison to stable releases, `filename` part for betas only has a `-beta` or `-beta.X` tag attached)
>
>So there is no need to check for `x86` and `x86_64` any longer!

>Actually, is this `betaEnd` and `exeBetaName` needed alltogether?
>
>We don't use such a pattern like `-x86_64_qt5` for any releases anymore.
Don't we differentiate via the download urls between stable and beta? 

- remove unused beta naming mentions

### DIFF
--- a/cockatrice/src/releasechannel.cpp
+++ b/cockatrice/src/releasechannel.cpp
@@ -9,8 +9,8 @@
 #include <QNetworkReply>
 
 #define STABLERELEASE_URL "https://api.github.com/repos/Cockatrice/Cockatrice/releases/latest"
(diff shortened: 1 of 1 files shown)


ctrlaltca:
Looks like this is fine to merge, since the only affected users are those on the "beta" release channel that didn't upgrade since more than one year.

PR #3351, opened 2018-07-24
UI dialog tweaks: filter + create game(s)

tooomm:
Dialogs looked a bit weird and misaligned.

- add group boxes to both dialogs for general information
- `create game`: move remember check box very visible to the bottom
- `create game`: swap game type selection with spectator box to have all important options on the left

## Screenshots
- `Filter games` before:
before2

- `Filter games` after:
after2

- `Create game` before:
before

- `Create game` after:
after-new


tooomm:
You are right, `Filter games` is a bit off. I didn't realized that... will take a look and try to fix that. The little additional gap shouldn't be there.

tooomm:
Fixed @ctrlaltca 

after-new2

ctrlaltca:
Looks better now! 👍

PR #3363, opened 2018-08-10
[WIP] Fallback on 404

andrewzwicky:
## Related Ticket(s)
- Fixes #2479 

## Short roundup of the initial problem
URL resolution and fallback was not working as expected

## What will change with this Pull Request?
URL resolution will be updated to be more in line with user expectations for primary and fallback URLs.

## Screenshots
So far, the update is performing basic functions as expected

```
Trying to load picture for card: "Angelic Benediction"  from set: "DD3_DVD"
Picture not found, trying to download
Trying to download picture for card: "Angelic Benediction"  from set: "DD3_DVD" from url: "http://magiccards.info/scans/en/dd3_dvd/19.jpg"
following redirect: "Angelic Benediction" Url: "https://magiccards.info/scans/en/dd3_dvd/19.jpg"
Download failed: "Error transferring https://magiccards.info/scans/en/dd3_dvd/19.jpg - server replied: Not Found"
Possible picture at  "https://magiccards.info/scans/en/dd3_dvd/19.jpg"  could not be loaded
Trying to load picture for card: "Angelic Benediction"  from set: "DD3_DVD"
Picture not found, trying to download
Trying to download picture for card: "Angelic Benediction"  from set: "DD3_DVD" from url: "http://gatherer.wizards.com/Handlers/Image.ashx?multiverseid=394008&type=deck"
Possible picture at  "http://gatherer.wizards.com/Handlers/Image.ashx?multiverseid=394008&type=deck"  could not be loaded
Trying to load picture for card: "Angelic Benediction"  from set: "M13"
Picture not found, trying to download
Trying to download picture for card: "Angelic Benediction"  from set: "M13" from url: "http://magiccards.info/scans/en/m13/4.jpg"
following redirect: "Angelic Benediction" Url: "https://magiccards.info/scans/en/m13/4.jpg"
Image successfully downloaded from  "https://magiccards.info/scans/en/m13/4.jpg"
```

This is using:
`http://magiccards.info/scans/en/!setcode_lower!/!setnumber!.jpg` as my primary URL and `http://gatherer.wizards.com/Handlers/Image.ashx?multiverseid=!cardid!&type=deck` as my secondary URL, (with `=deck` on the end, so it will intentionally fail).

For the set that doesn't match, the download is attempted, but then it moves the secondary URL, then to the next set.  Code needs to be cleaned up and tested more thoroughly, but it's looking good so far.

Feedback welcome (keep in mind I have more to do on this before I'm ready to ask for it to be merged).

PR #3367, opened 2018-08-12
[WIP] Card image loading: Fallback on 404

### LINKED ISSUE #2479: Spec for card image download priority order
Currently we have a loosely-defined image url download priority order with fallback.

The goal of this ticket is to come up with a flexible, understandable design that helps users get the images they want.

At a high level, we currently have:
if card has muid:
- try main url for every enabled set containing that card in priority order
If card does not have muid:
- use fallback

I propose instead:
- Allow users to make a list of download urls to attempt, with a priority order.

When a card image is requested:
- For each enabled set in order, try all download url templates in order, skipping url templates which ask for information that the card doesn't have
  - eg if the template asks for muid, and card has none, skip that url.

In pseudocode:
```
QList<UrlTemplate> urls = get_urls_for_card(cardInfo, enabledSetConfiguration);
for (UrlTemplate url, urls):
    Url url = UrlTemplate.apply(cardInfo)
    response = download(url)
    if response.ok:
        save(response.image)
    else:
        continue
```
  
This ticket supersedes #1984 and #2346

andrewzwicky:
## Related Ticket(s)
- Fixes #2479

See #3363 

- Relative to #3363, the work has been updated to no longer use separately managed indices, and instead uses lists and takeFirst() to empty them.

- Method for substituting information into Urls now gives more feedback about which items are failing, and will also return empty string instead of leaving empty strings inside the return Url.

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -265,7 +265,8 @@ QString PictureToLoad::transformUrl(QString urlTemplate) const
                 /* This means the template is requesting information that is not populated
                  * in this card, so it should return an empty string, indicating an invalid Url.
                  */
-                qDebug() << "Requested information (" << prop << ") for Url template (" << urlTemplate << ") is not available for card:" << card->getName();
+                qDebug() << "Requested information (" << prop << ") for Url template (" << urlTemplate
+                         << ") is not available for card:" << card->getName();
                 return QString();
             }
         }
@@ -286,10 +287,8 @@ QString PictureToLoad::transformUrl(QString urlTemplate) const
                 /* This means the template is requesting information that is not populated
                  * in this card, so it should return an empty string, indicating an invalid Url.
                  */
-                qDebug() << "Requested information (" << prop <<
-                            ") for Url template (" << urlTemplate <<
-                            ") is not available for card (" << card->getName() <<
-                            ") in set (" << getSetName() << ")";
+                qDebug() << "Requested information (" << prop << ") for Url template (" << urlTemplate
+                         << ") is not available for card (" << card->getName() << ") in set (" << getSetName() << ")";
                 return QString();
             }
         }
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -161,20 +161,20 @@ void PictureLoaderWorker::processLoadQueue()
         QString setName = cardBeingLoaded.getSetName();
         QString cardName = cardBeingLoaded.getCard()->getName();
         QString correctedCardName = cardBeingLoaded.getCard()->getCorrectedName();
-        qDebug() << "Trying to load picture for card:" << cardName << " from set:" << setName;
+        qDebug() << "PictureLoader: Trying to load picture for card:" << cardName << " from set:" << setName;
 
         if (cardImageExistsOnDisk(setName, correctedCardName))
             continue;
 
         if (picDownload) {
-            qDebug() << "Picture not found on disk, trying to download";
+            qDebug() << "PictureLoader: Picture not found on disk, trying to download";
             cardsToDownload.append(cardBeingLoaded);
             cardBeingLoaded.clear();
             if (!downloadRunning)
                 startNextPicDownload();
         } else {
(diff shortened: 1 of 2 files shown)


andrewzwicky:
I'm still planning to do more targeted testing to make sure things work under all conditions, but early feedback is always helpful.

cockatrice/src/pictureloader.cpp
```diff
@@ -43,40 +43,80 @@ class PictureToLoad::SetDownloadPriorityComparator
...
+
+    foreach (QString urlTemplate, urlTemplates) {
+        QString transformedUrl = transformUrl(urlTemplate);
+
+        if (!transformedUrl.isEmpty()) {
+            currentSetUrls.append(transformedUrl);
+        }
+    }
+
+    if (!currentSetUrls.isEmpty()) {
```
ctrlaltca: This few lines are basically an initial call to `nextUrl()`

cockatrice/src/pictureloader.cpp
```diff
@@ -256,30 +319,32 @@ void PictureLoaderWorker::startNextPicDownload()
...
         networkManager->get(req);
     }
 }
 
 void PictureLoaderWorker::picDownloadFailed()
 {
-    if (cardBeingDownloaded.nextSet()) {
-        qDebug() << "Picture NOT found, download failed, moving to next set (newset: "
-                 << cardBeingDownloaded.getSetName() << " card: " << cardBeingDownloaded.getCard()->getName() << ")";
+    if (cardBeingDownloaded.nextUrl() || cardBeingDownloaded.nextSet()) {
```
ctrlaltca: Note to future myself: this works because the || operator is short-circuiting: if the first operand is true, the second operand is not evaluated.

cockatrice/src/pictureloader.cpp
```diff
@@ -43,40 +43,80 @@ class PictureToLoad::SetDownloadPriorityComparator
...
+
+    if (card && currentSet) {
+        QString setCustomURL = card->getCustomPicURL(currentSet->getShortName());
+
+        if (!setCustomURL.isEmpty()) {
+            currentSetUrls.append(setCustomURL);
+        }
+    }
+
+    foreach (QString urlTemplate, urlTemplates) {
```
ctrlaltca: The `urlTemplates` list contains a local copy of the unmodified picUrl strings from settingCache; I don't see it being useful in the present since we could just fetch the data from settingsCache here. Maybe you ahve an idea of using it to filter/modify the data coming from settingsCache?

andrewzwicky:
I can clean up those few things and add some comments.  It probably won't get done until mid next-week.

cockatrice/src/pictureloader.cpp
andrewzwicky: Yes, I am planning on updating the settingCache to contain a list of Urls instead of just two.  This is in line with what #2479 was proposing.  It will just make things easier to add in the future.

cockatrice/src/pictureloader.cpp
andrewzwicky: good catch, I'll update this as well.

andrewzwicky:
If you're good with that, I'm satisfied with it.

ctrlaltca:
Merged, thank you a lot for the code and your patience!

tooomm:
Thanks for the  feature! :)

@andrewzwicky please remove the `[WIP]` label from the PR title when you're done next time, please, and @ctrlaltca check such things before merging.

Otherwise it ends up in our commit list and is a bit confusing:
https://github.com/Cockatrice/Cockatrice/commits/master

I changed it now, but that won't change the history.

cockatrice/src/pictureloader.cpp
ZeldaZach: My goal was to remove foreach from the codebase. Will get back to fixing this at some point in the future

Daenyth: @ZeldaZach in favor of using `map` and `fold` functions? Nothing seems really terrible with foreach

andrewzwicky: @ZeldaZach I'm newish to C++, what's the problem / prefered alternative for `foreach`?

PR #3377, opened 2018-08-30
[NEEDS TESTING] Fix servatrice linking under FreeBSD - fix #3376

### LINKED ISSUE #3376: Issue in linking of servatrice
<b>FreeBSD 11.1-RELEASE. C++ compiler is Clang 4.0.0</b>
  

  
I'm struggling to make the executable "servatrice' on my FreeBSD server. When compiling, it's hitting the issue: 
```
[ 91%] Building CXX object /CMakeFiles/servatrice.dir/servatrice_autogen/mocs_compilation.cpp.o
[ 91%] Linking CXX executable servatrice
/bin/ld: undefined reference to symbol `backtrace_symbols_fd' (try adding -lexecinfo)
//usr/lib/libexecinfo.so.1: could not read symbols: Bad value
c++: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1

Stop.
make[2]: stopped in /usr/share/games/Cockatrice/build
*** Error code 1

Stop.
make[1]: stopped in /usr/share/games/Cockatrice/build
*** Error code 1

Stop.
make: stopped in /usr/share/games/Cockatrice/build
```

I've gone through the servatrice source and found the call for 'backtrace_symbols_fd' on line: 89 of signalhandler.cpp. Further research leads me to execinfo.h being the header file that has the reference to the function, but now I'm struggling to find exactly where the function is _defined_. Would you all be able to help me? What other information can I provide that would be of use?

ctrlaltca:
## Related Ticket(s)
- Fixes #3376

In servatrice e are currently using the `backtrace_symbols_fd()` call, that is specific to the GNU libc.
For FreeBSD an equivalent exists, but we need to check for `libexecinfoa and then link with `-lexecinfo`.


mikey0715:
Hi guys, 

Sorry for taking so long to get back to you. This fix is functioning for me, I followed your links to the CMakeLists in the other Github thread, and it compiled and I've been playing with friends ever since!

ctrlaltca:
@mikey0715 thank you for the feedback, I've added the proposed fix.

Issue #3378, opened 2018-08-30
Appveyor: 32bit build fails


ctrlaltca:
Error signature:
```
c:\qt\5.9\msvc2015\include\qtcore\qalgorithms.h(847): error C3615: constexpr function 'qCountLeadingZeroBits' cannot result in a constant expression [C:\projects\cockatrice\build\oracle\oracle.vcxproj]
  c:\qt\5.9\msvc2015\include\qtcore\qalgorithms.h(858): note: failure was caused by call of undefined function or one not declared 'constexpr'
  c:\qt\5.9\msvc2015\include\qtcore\qalgorithms.h(858): note: see usage of 'qPopulationCount'
```

This is due to the recent update of Visual Studio 2017 15.8.
Qt already knows the problem: https://bugreports.qt.io/browse/QTBUG-67259, a patch in Qt is needed to fix it: https://codereview.qt-project.org/#/c/236948/
We should hope in a quick release of the patched Qt version 5.9.7.

tooomm:
Once https://github.com/appveyor/ci/issues/2645 is deployed, this should finally be solved.

tooomm:
@ctrlaltca with the new update at appveyor Qt might be fixed, but we are seeing different errors now: https://ci.appveyor.com/project/Daenyth/cockatrice-8k7y5/builds/20209818

ctrlaltca:
The new errors are due to some other modifications that Apvveyor made to the build images.
I've already reported these problems [here](https://github.com/appveyor/ci/issues/2737), and [here](https://github.com/Cockatrice/Cockatrice/pull/3434)'s a PR to workaround them and restore wroking builds.

tooomm:
Thank for taking a look! Also for the upstream report and workaround PR. 👍

PR #3384, opened 2018-09-03
Disable saving of decks when the deck is empty

### LINKED ISSUE #3339: prevent saving of empty decks
<b>System Information:</b>
Client Version: 2.6.0-custom(168cc49) (2018-07-20)
Build Architecture: 64-bit
Qt Version: 5.9.5
System Locale: de_DE

Currently it's possible to save an empty deck.
The menu entry should be greyed out until there is something to save like an added name, comments or cards...

leestran1995:
## Related Ticket(s)
- Fixes #3339 

- Empty decks are able to be saved

- The "save" and "save as" options will be greyed out when the deck is empty
- Code for checking the size of the deck is borrowed from the actExportDeckDecklist() function

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -756,12 +756,10 @@ void TabDeckEditor::actLoadDeck()
     DeckLoader::FileFormat fmt = DeckLoader::getFormatFromName(fileName);
 
     auto *l = new DeckLoader;
-    if (l->loadFromFile(fileName, fmt))
-    {
+    if (l->loadFromFile(fileName, fmt)) {
         setSaveStatus(false);
         setDeck(l);
-    }
-    else
+    } else
         delete l;
 }
 
@@ -999,13 +997,9 @@ void TabDeckEditor::actRemoveCard()
     deckModel->removeRow(currentIndex.row(), currentIndex.parent());
 
     DeckLoader *const deck = deckModel->getDeckList();
-    QString decklistUrlString;
-    if (deck->isEmpty())
-    {
+    if (deck->isEmpty()) {
         setSaveStatus(false);
-    }
-    else
-    {
+    } else {
         setSaveStatus(true);
     }
     setModified(true);
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -989,7 +989,7 @@ void TabDeckEditor::actAddCardToSideboard()
 void TabDeckEditor::actRemoveCard()
 {
     const QModelIndex &currentIndex = deckView->selectionModel()->currentIndex();
-    if (!currentIndex.isValid() || deckModel->hasChildren(currentIndex)) 
+    if (!currentIndex.isValid() || deckModel->hasChildren(currentIndex))
         return;
     deckModel->removeRow(currentIndex.row(), currentIndex.parent());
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/tab_deck_editor.cpp
```diff
@@ -988,6 +992,17 @@ void TabDeckEditor::actRemoveCard()
     if (!currentIndex.isValid() || deckModel->hasChildren(currentIndex))
         return;
     deckModel->removeRow(currentIndex.row(), currentIndex.parent());
+
+    // Disable saving if the deck is empty
+    DeckLoader *const deck = deckModel->getDeckList();
+    QString decklistUrlString;
+    if (deck) {
+        decklistUrlString = deck->exportDeckToDecklist();
```
ctrlaltca: I know this is only run when a card is removed, but it looks to me like a lot of wasted cpu cycles.
Why not use `deck->isEmpty()`?

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -988,6 +994,17 @@ void TabDeckEditor::actRemoveCard()
...
         return;
     deckModel->removeRow(currentIndex.row(), currentIndex.parent());
+
+    DeckLoader *const deck = deckModel->getDeckList();
+    QString decklistUrlString;
+    if (deck->isEmpty()) {
+        aSaveDeck->setEnabled(false);
+        aSaveDeckAs->setEnabled(false);
+    }
+    else {
```
tooomm: Tiny code style error left ❤️

leestran1995: I'll get on all that stuff right away!

Speaking of style errors, I noticed a lot of if/else statements omitting braces for single-statements which goes against the code style guide. Would there be any utility to refactoring them to match the style guide?

tooomm: Normally clang-format should catch and report those. It can also fix them with a simple command, that's why we use it. :)
Speaking about that, we still have no description setup on how to properly use it...

Maybe somebody else know what's up with the if/else style you mentioned.
Config got added and discussed in https://github.com/Cockatrice/Cockatrice/pull/3028 by the way.

In the meantime can you point to 1-2 examples?

leestran1995: image

This if/else block contradicts the Braces section in the contributing guide in that the braces are not on their own line.

image

This also contradicts the Braces section of the Contributing guide since the braces are omitted.

I think maybe it's the contributing guide that should be updated to reflect the way the automatic code review bots enforce the code styling?

berserkingyadis:
What about print deck and send deck to online service? Shouldnt they be disabled too?

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -988,6 +995,13 @@ void TabDeckEditor::actRemoveCard()
     if (!currentIndex.isValid() || deckModel->hasChildren(currentIndex))
         return;
     deckModel->removeRow(currentIndex.row(), currentIndex.parent());
+
+    DeckLoader *const deck = deckModel->getDeckList();
+    if (deck->isEmpty()) {
```
ZeldaZach: Can simplify to `setSaveStatus(!deck->isEmpty())`

leestran1995: Of course! I'll commit that ASAP

ctrlaltca:
@tooomm are you ok with this? Shall we merge?

tooomm:
Can check tomorrow again.

tooomm:
After loading a deck, the options shouldn't be disabled.
Example workflows: Load deck --> `print` or `send to online service` or `save copy`

tooomm:
No worries, great! 🎉

Daenyth:
@ctrlaltca do you know what's up with the appveyor failures?

tooomm:
@Daenyth see https://github.com/Cockatrice/Cockatrice/issues/3378 for the failure on 32bit builds currently

Daenyth:
@ZeldaZach ready to go :)

ZeldaZach:
It technically works, so I'll get this in :)

PR #3401, opened 2018-09-29
Translate curly apostrophe to ASCII apostrophe (#3395)

### LINKED ISSUE #3395: Treat different apostrophes similarly ' ’
When searching for cards, some resources treat card titles with curly quotes `’` which can’t be searched with straight quotes `'`.

Could we implement support in the search bars for straight and curly quotes interchangeably? i.e. searching for `'` or `’` returns both.

Hydro666:
Treats the curly apostrophe as a straight apostrophe when searching in
the deck editor search bar.

## Related Ticket(s)
- Fixes #3395 

When users would paste card text into the deck editor search bar, curly apostrophes would not be interpreted as straight ASCII apostrophes.

- Curly apostrophes will now be interpreted as ASCII apostrophes in the deck editor search bar.

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -700,9 +700,7 @@ void TabDeckEditor::updateCardInfoRight(const QModelIndex &current, const QModel
 
 void TabDeckEditor::updateSearch(const QString &search)
 {
-    QString sanitizedInput = QString(search);
-    sanitizedInput.replace("’", "'");
-    databaseDisplayModel->setCardName(sanitizedInput);
+    databaseDisplayModel->setCardName(search);
     QModelIndexList sel = databaseView->selectionModel()->selectedRows();
     if (sel.isEmpty() && databaseDisplayModel->rowCount())
         databaseView->selectionModel()->setCurrentIndex(databaseDisplayModel->index(0, 0),
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -700,7 +700,9 @@ void TabDeckEditor::updateCardInfoRight(const QModelIndex &current, const QModel
 
 void TabDeckEditor::updateSearch(const QString &search)
 {
-    databaseDisplayModel->setCardName(search);
+    QString sanitizedInput = QString(search);
+    sanitizedInput.replace("’", "'");
+    databaseDisplayModel->setCardName(sanitizedInput);
     QModelIndexList sel = databaseView->selectionModel()->selectedRows();
     if (sel.isEmpty() && databaseDisplayModel->rowCount())
         databaseView->selectionModel()->setCurrentIndex(databaseDisplayModel->index(0, 0),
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Can you look at supporting all of the possible symbols used? Such as the fancy quotes for things like Kongming

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -700,7 +700,9 @@ void TabDeckEditor::updateCardInfoRight(const QModelIndex &current, const QModel
 
 void TabDeckEditor::updateSearch(const QString &search)
 {
-    databaseDisplayModel->setCardName(search);
+    QString sanitizedInput = QString(search);
+    sanitizedInput.replace("’", "'");
+    databaseDisplayModel->setCardName(sanitizedInput);
```
Daenyth: What do you think about putting this logic inside `setCardName`?

Hydro666: I wasn't sure how many other modules relied on `setCardName`, and that maybe the distinction between the curly apostrophe and the ASCII apostrophe was meaningful.

I did look and it seems like `dlg_crate_token.cpp:180` does use it, via a signal sent from a text box containing the default text "Token", however the function containing this statement doesn't appear to be used anywhere else in the program.

In this case, I think that it's safe to put this logic in `serCardName`.

Hydro666:
@ZeldaZach I created a map of type `QMap<QString, QString>` that will serve as the translation table. and placed it in the file `carddatabasemodel.cpp:23`. This way, if we want to add any more characters that we would like translated, we can add them to the map instantiation.

ctrlaltca:
Visual studio is reporting a few errors [here](https://ci.appveyor.com/project/Daenyth/cockatrice-8k7y5/build/build%201512/job/qu97l5367rsn0m63#L2248):
```
cockatrice\src\carddatabasemodel.cpp(21): error C2015: too many characters in constant
cockatrice\src\carddatabasemodel.cpp(22): error C2015: too many characters in constant
cockatrice\src\carddatabasemodel.cpp(23): error C2015: too many characters in constant cockatrice\src\carddatabasemodel.cpp(24): error C2015: too many characters in constant 
```

Hydro666:
Interesting, AppVeyor appears to be throwing an error due to: https://bugreports.qt.io/browse/QTBUG-67259 "Compilation error on MSVC with enabled constexpr on 32-bit target"

The bug was opened on 2018-03-22 for a noisy error of the form `error C3615: constexpr function 'qCountLeadingZeroBits' cannot result in a constant expression` but was closed because [MSVC did not yet implement the breaking change](https://bugreports.qt.io/browse/QTBUG-67259?focusedCommentId=396586&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-396586), but it has now been implemented as of [2018-08-15](https://bugreports.qt.io/browse/QTBUG-67259?focusedCommentId=415294&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-415294). Interestingly the bug assignee seems to think that this issue will not be present in QT 5.9, but afaik, cockatrice uses QT 5.9.5 so it seems weird that we would still have this issue.

ctrlaltca:
Looks fine now!
Thank you @Hydro666

Hydro666:
No problem! Happy to help :)

Issue #3402, opened 2018-09-30
Suggestion: add "reset P/T to base values" as an option in power/toughness context menu


aeschenkarnos:
Suggestion: under the right-click context menu for power / toughness, add the option "reset to base values". This is one of the more commonly used actions through that menu. 

2. It could also be useful to have a hotkey for this action.

3. "Set power and toughness" (hotkey ctrl-P) could pre-populate to the base values, rather than the current values.

ebbit1q:
Seems like a good first issue, I'll try this :)

tooomm:
Cool. Join our Gitter chat in case you've any questions! @ebbit1q

ebbit1q:
This issue has been resolved and can be closed.
#3412 added the ctrl+alt+0 hotkey and menu options for resetting p/t.

tooomm:
You are right, thanks for your hint (and the PR in the first place 😄)! @ebbit1q 
Just had a chance to test the change in the app, I like it. 👍 

Regarding the open issue, that slipped our attention because we are used to closing issues right from within the PR with keywords:
https://help.github.com/articles/closing-issues-using-keywords/

We have that also covered in our PR template.

PR #3412, opened 2018-10-20
add reset power and toughness option #3402

ebbit1q:
Adds option to reset power and toughness to context menu and keybinds.
See issue #3402

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -390,6 +390,8 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
     connect(aDecPT, SIGNAL(triggered()), this, SLOT(actDecPT()));
     aSetPT = new QAction(this);
     connect(aSetPT, SIGNAL(triggered()), this, SLOT(actSetPT()));
+    aResetPT = new QAction(this);
+    connect(aResetPT, SIGNAL(triggered()), this, SLOT(actResetPT()));
     aSetAnnotation = new QAction(this);
     connect(aSetAnnotation, SIGNAL(triggered()), this, SLOT(actSetAnnotation()));
     aFlip = new QAction(this);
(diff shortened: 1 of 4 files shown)


ZeldaZach:
Make sure to run the `clangify.sh` executable so your builds can be retried for actual failures :)

ebbit1q:
Ok, will do

ebbit1q:
clangify did nothing other than move one file back to otb style while the style guide tells you not to use it?

ebbit1q:
https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md#braces this is literally the opposite of what clang-format does?

cockatrice/src/player.cpp
```diff
@@ -2347,6 +2351,31 @@ void Player::actIncPT(int deltaP, int deltaT)
...
+        Command_SetCardAttr *cmd = new Command_SetCardAttr;
+        QString zoneName = card->getZone()->getName();
+        cmd->set_zone(zoneName.toStdString());
+        cmd->set_card_id(card->getId());
+        cmd->set_attribute(AttrPT);
+        QString ptString = info->getPowTough();
+        cmd->set_attr_value(ptString.toStdString());
+        commandList.append(cmd);
+
+        if (local) {
```
ZeldaZach: You set playerid at the top for no apparent reason to possibly modify it here. A suggestion could be to ternary case it, or if/else

ZeldaZach: Also, you're checking if it's a local game each time... can probably do it just once

ebbit1q: I could but then I'd have to change that everywhere else too.

ZeldaZach: Ohh.. hmm. Refactor at some point will be necessary, but this is good to go :)

tooomm: Is there a ticket for that refactor with more details?

tooomm:
> [/.github/CONTRIBUTING.md@`master`#braces](https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md#braces) this is literally the opposite of what clang-format does?

@ebbit1q Can you please open a ticket with the issues you found for further discussion. We can than fix it to make it clear to all other new contributors. Thanks!

ebbit1q:
I opened a ticket: #3413 but there is a lot of other tickets with contributing guide issues

tooomm:
Thanks for the ticket and the fix right after @ebbit1q 👍

PR #3415, opened 2018-10-25
Build ubuntu 18.04 package using appveyor

### LINKED ISSUE #3294: Update ubuntu support to supported versions

ctrlaltca:
## Related Ticket(s)
- Fixes #3294

- Appveyor compiles a "release" package for ubuntu 18.04;
- ~~some currently broken dependency packages have been updated to working versions (openssl, protobuf)~~ --> See https://github.com/Cockatrice/Cockatrice/pull/3420

EDIT
Notes:
 - ~~win32 is still broken; Qt 5.9.7 has been released 2 days ago, hopefully Appveyor will pick up the update soon~~ --> See https://www.appveyor.com/updates/2018/11/09/
 - the "deploy" part of the script can be been tested only inside Cockatrice's master branch.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -90,18 +90,7 @@ install:
         }
     # Linux
     - sh: sudo apt-get update
-    - sh: sudo apt-get install -y \
-        cmake \
-        libprotobuf-dev \
-        protobuf-compiler \
-        qt5-default \
-        qttools5-dev \
-        qttools5-dev-tools \
-        qtmultimedia5-dev \
-        libqt5multimedia5-plugins \
-        libqt5svg5-dev \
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Do we want to go through with this or wait for #3433  instead

Daenyth:
I think I'd prefer to go with the docker approach, mainly because I think it gives us a ton of options for doing even more release versions on linux later

PR #3419, opened 2018-10-28
Improve hints for manage sets dialog

tooomm:
- stress presentation and better fit to layout
- add group box with label

## Screenshots
- before:
before

- after:
after


tooomm:
The mac log is full of homebrew actions and exceeds the maximum limit, that's why it keeps failing:
`The job exceeded the maximum log length, and has been terminated.`

Any idea what's causing this?

PR #3020, opened 2018-01-16
Windows package updates

ctrlaltca:
## Related Ticket(s)
- Fixes #2994 
- Fixes #2031
- Relates to #2973

## Short roundup of the initial problem
The openssl libraries that we currently bundle depends on the Visual studio 2013 runtime. This means that while Cockatrice (and Oracle, Servatrice) depends on the Visual studio 2015 runtime to start, the Visual studio 2013 runtime is also needed in order to make Qt able to load the openssl libraries and handle https urls.

## What will change with this Pull Request?
 * Appveyor will use Visual Studio 2017 to compile protobuf, zlib and Cockatrice.
 * Protobuf has been updated to 3.5.1
 * Instead of using the appveyor-provided openssl package, we download and bundle a semi-official package from https://indy.fulgan.com (found here: https://wiki.openssl.org/index.php/Binaries ) than has no dependency on any Visual Studio runtime
 * a small compilation warning has been removed (leftover of Windows Xp compatibility)
 * during the "install" step, cmake will install a copy of the correct vcresit packae in the cockatrice tree that will be included in the NSIS package
 * during installation, NSIS will run the vcredist if available inside the package.

As a result of this last change, the package size almost doubled from  ~16 to ~30 MB.

## Screenshots
*(simply drag & drop image files directly into this description!)*

PR #3421, opened 2018-10-28
appveyor: use provided OpenSSL again

tooomm:
External OpenSSL was used because of some former issues with bundling, see https://github.com/Cockatrice/Cockatrice/pull/3020.
These problems seems to be solved.

- use OpenSSL from appveyor again to cut down on external dependencies: https://www.appveyor.com/docs/windows-images-software/#tools

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,7 +32,7 @@ cache:
 
 
 environment:
-    protobuf_ver: 3.6.0
+    protobuf_ver: 3.6.1
     zlib_ver: 1.2.11
 
     matrix:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,7 +24,6 @@ clone_depth: 50    #same as travis, see https://www.appveyor.com/blog/2014/06/04
 image: Visual Studio 2017
 
 cache:
-    - c:\openssl-release
     - c:\protobuf-release
     - c:\zlib-release
 # TODO: set dependency on ps skript file (in ./ci) / "-> appveyor.yml" maybe not ideal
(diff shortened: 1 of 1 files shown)


ctrlaltca:
I've tried downloading the resulting artifact from appveyor and looking at the imports of libeay32.dll and ssleay32.dll. They depends on msvcr120.dll, so they need the ms visual c++ 2013 redistributable, while the rest of cockatrice and its dlls depends on msvcr140.dll (msvc++ 2015/2017).
So, this is not going to work yet.

tooomm:
Hmm, did you try it?
I only checked that the appveyor build passed and the logs say OpenSSL was found. I then installed it locally to run Oracle. I was able to download cards and tokens via two https links which should only work with OpenSSL?
https://mtgjson.com/json/AllSets.json.zip
https://raw.githubusercontent.com/Cockatrice/Magic-Token/master/tokens.xml
No debugging info though, since Oracle doesn't print to the `Debug Log` in Cockatrice.

Client Version: 2.6.1-custom(552c37a) (2018-10-28)
oracle-552c37a

ctrlaltca:
Of course it works, it used to work also in the past.
It won't work only on computers where the needed msvcr120.dll is not available because nothing else installed that runtime version yet.

tooomm:
Oh sure, got you. We definitely want to stick with one single msvcr dependency!
Regarding all the issues user brought up to us with that. :)

Thank you for clarification @ctrlaltca!

tooomm:
Any idea if using the even newer `OpenSSL 1.1.0i ` Appveyor provides as well has the same issue? @ctrlaltca 

https://www.appveyor.com/docs/windows-images-software/#tools

PR #3422, opened 2018-10-29
Load plain improvements

### LINKED ISSUE #3064: On deck load "name" and "comments" have wrong information
<b>Info:</b>
Client Version: 2.5.0-custom(7fbff92) (2018-01-31)
Build Architecture: 64-bit
Qt Version: 5.9.3

__________

As spotted in a recent PR: https://github.com/Cockatrice/Cockatrice/pull/3057#pullrequestreview-92894335

- The text for both is all lower case
- There is another bug when you copy any deck to the clipboard from the client menu, and import the deck back from there, suddenly the comment field contains information like `4 maindeck` which are our own annotations we create

Example:
comments issue

ebbit1q:
## Related Ticket(s)
- Fixes #3064

Some weirdness in loading plaintext decklists resulted in the name and comments being added wrong when using the same format Cockatrice uses to export plaintext decklists.

- loading plaintext decklists will be completely rewritten.
- comments and names will no longer be all lowercase when imported.
- the first two comments with mainboard and cardtypes will no longer be seen as names or comments.

I've tested this new function extensively on some weird edgecase decklists but feel free to doublecheck.
Some things that are a bit weird:
- having two entries for the same card inserts them as separate entries, this is the same as the original functionality and might warrant a new issue.
- the original function checked for a pipe `|` and removed everything behind it for some reason, this caused split cards noted like "Fire | Ice" to not register correctly, this functionality didn't seem useful to me and has not been implemented.
- all the regexes are currently defined inside the function but as they are const they might be defined as private members, or as static variables inside the function (I prefer avoiding static), should this be changed?
- the new function regards every weird character that should never occur as the first character of a cardname as a comment, this is good imo but debatable as it might remove the start of a deck comment starting with such a character. (please tell me how many of your decknames start with #, $, or &; and if missing that first character when you plaintext copy it matters)

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -481,7 +481,7 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
     const QRegularExpression reCardLine("^\\s*[\\w\\[\\(\\{].*$", QRegularExpression::UseUnicodePropertiesOption);
     const QRegularExpression reEmpty("^\\s*$");
     const QRegularExpression reComment("[\\w\\[\\(\\{].*$", QRegularExpression::UseUnicodePropertiesOption);
-    const QRegularExpression reSBMark("^sb:\\s*(.+)", QRegularExpression::CaseInsensitiveOption);
+    const QRegularExpression reSBMark("^\\s*sb:\\s*(.+)", QRegularExpression::CaseInsensitiveOption);
     const QRegularExpression reSBComment("sideboard", QRegularExpression::CaseInsensitiveOption);
 
     // simplified matches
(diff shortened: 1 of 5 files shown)


ebbit1q:
I noticed that the checks test that if there are more than one empty lines all cards should be mainboard, are we sure we don't want it to simply take everything after the last empty line as sideboard?

ebbit1q:
I compiled the tests locally, there's a few things I didn't put into account.

ebbit1q:
this pr is now dependent on #3425

ebbit1q:
There simply are too many things broken in the previous version, can I simply write new checks that actually test properly for just this version?

ebbit1q:
Do you want me to squash the commits?

ZeldaZach:
Don’t worry about squashing, we auto do it on merge

ebbit1q:
cool

ebbit1q:
Hold on, I'm fixing the last problem with the capitalization of the card names. Currently unknown cards get decapitalized, this causes an issue if you would import a plaintext deck with cards you don't have yet because you forgot to update your sets.

Daenyth:
I really appreciate the thoroughness on the edge case handling. Great work :)

tooomm:
@ebbit1q You can always add a `[WIP]` tag to your PR title to indicate that it's not yet ready!

ebbit1q:
Well it could have been merged right now, it'd still have some bugs shared with the old version, but there would be no regression.

ebbit1q:
Nobody noticed I forgot header guards?

ebbit1q:
note, we also still use QT 5.5.0 on travis on the ubuntu 16.04 build

PR #3423, opened 2018-10-29
add more precise rules to CONTRIBUTING.md

ebbit1q:
Add extra rules that clang-format enforces, like the order of includes and the way single line comments should be formatted.
Correct some grammar ( what is up with the translation section ordering steps out of order?? ).
This is a web edit.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -38,25 +38,25 @@ Use header guards in the form of `FILE_NAME_H`.
 Simple functions, such as getters, may be written inline in the header file,
 but other functions should be written in the source file.
 
-Keep library includes and project includes grouped together. So this is okay:
+Group library includes after project includes, and in alphabetic order. Like this:
 ```c++
 // Good
-#include <QList>
-#include <QString>
-#include "card.h"
-#include "deck.h"
-
-// Good
 #include "card.h"
 #include "deck.h"
 #include <QList>
 #include <QString>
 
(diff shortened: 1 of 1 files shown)


ebbit1q:
The translation section lists five steps but when explaining the five steps they are listed as 1, 2, 3, 5, 4; likely to separate them over different roles (dev, maintainer, translator) but it's just confusing.

tooomm:
Ahh, got you now! And you are right in your assumption, that was the idea behind it.
If we order it by numbers the process for maintainers will be split up and the reader might think he is done already without - realizing after a step not relevant for him - sometime else has to be done.

But I'll look into improving that and at least add some linking.
Maybe I've some other idea, too. Any suggestion from your side?

ebbit1q:
I say ditch the step numbers in the explanation and see them as guides per target user, it should mention in the maintainer guide that translators do something in between the last two steps of course.

tooomm:
For now I added links to the ordered structure and a jumplist per "user type":
https://github.com/Cockatrice/Cockatrice/blob/tooomm-contributing/.github/CONTRIBUTING.md#translations

Will look further and consider your input in the next days. 👍

ebbit1q:
I created a different pr from the web ui to show how I think the translating chapter would be more readable #3424 , feel free to close it if you want to preserve the current step layout.

ZeldaZach:
Conflicts will need to be resolved b4 i can merge this

PR #3433, opened 2018-11-07
add docker compilation to travis

ebbit1q:
my attempt at simply adding the dockerfile then running compilation inside it

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu:bionic
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        bc \
+        build-essential \
+        clang-format \
+        g++ \
+        git \
+        cmake \
+        libprotobuf-dev \
+        libqt5multimedia5-plugins \
+        libqt5svg5-dev \
+        libqt5sql5-mysql \
+        libqt5websockets5-dev \
+        protobuf-compiler \
+        qt5-default \
+        qttools5-dev \
+        qttools5-dev-tools \
+        qtmultimedia5-dev \
+    && rm -rf /var/lib/apt/lists/*
--- a/.ci/travis-compile.sh
+++ b/.ci/travis-compile.sh
@@ -2,7 +2,7 @@
 
 # This script is to be used in .travis.yaml from the project root directory, do not use it from somewhere else.
 
-set -e
+set -ex
 
 while [[ "$@" ]]; do
   case "$1" in
@@ -68,7 +68,7 @@ fi
 
 # Add qt install location when using brew
 if [[ $(uname) == "Darwin" ]]; then
-  export PATH="/usr/local/opt/ccache/bin:$PATH"
+  PATH="/usr/local/opt/ccache/bin:$PATH"
   flags+=" -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5/"
 fi
 
--- a/.ci/travis-compile.sh
+++ b/.ci/travis-compile.sh
@@ -26,6 +26,14 @@ while [[ "$@" ]]; do
       MAKE_TEST=1
       shift
       ;;
+    '--debug')
+      BUILDTYPE="Debug"
+      shift
+      ;;
+    '--release')
+      BUILDTYPE="Release"
+      shift
+      ;;
     *)
       BUILDTYPE="$1"
       shift
@@ -55,12 +63,13 @@ fi
 mkdir -p build
 cd build
 
-# Add flags
+# Add cmake flags
 if [[ $MAKE_SERVER ]]; then
   flags+=" -DWITH_SERVER=1"
 fi
 if [[ $MAKE_TEST ]]; then
   flags+=" -DTEST=1"
+  BUILDTYPE="Debug" # test requires buildtype Debug
 fi
 if [[ $BUILDTYPE ]]; then
   flags+=" -DCMAKE_BUILD_TYPE=$BUILDTYPE"
--- a/.ci/travis-compile.sh
+++ b/.ci/travis-compile.sh
@@ -2,8 +2,9 @@
 
 # This script is to be used in .travis.yaml from the project root directory, do not use it from somewhere else.
 
-set -ex
+set -e
 
+# Read arguments
 while [[ "$@" ]]; do
   case "$1" in
     '--format')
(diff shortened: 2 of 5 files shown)


tooomm:
We would need travis to deploy the .dmg file from the 18.04 container to github releases as well.

ebbit1q:
I'm not sure what you mean, aren't dmg files for mac? Note that this direction only uses docker for the compilation, it literally runs .ci/travis-compile.sh on the local repo (not a copy but it has full access and will leave the local files as if compiled on 18.04).

ebbit1q:
It would be preferable to have the docker commands that are in compile.sh right now in travis.yaml (right now compile.sh with DOCKER_BUILDTYPE="Debug" will run compile.sh itself with BUILDTYPE="Debug" inside the docker container) but conditionally. Not to mention the travis container no longer needs dependencies itself, it only needs to build the docker image.

tooomm:
I meant .deb of course.. :)

ebbit1q:
Well it does.

ebbit1q:
I mean it does that for 16.04 doesn't it? This is no different. Where does it export it as .deb exactly? I can't find it.

.ci/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+        build-essential \
+        g++ \
+        cmake \
+        libprotobuf-dev \
+        protobuf-compiler \
+        qt5-default \
+        qttools5-dev \
+        qttools5-dev-tools \
+        qtmultimedia5-dev \
+        libqt5multimedia5-plugins \
```
Daenyth: Can you alphabetize the lines? It helps keep it manageable as it changes over time.

.ci/travis-compile.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+if [[ $DOCKER_BUILDTYPE ]]; then
```
Daenyth: Should this be `[[ -n $DOCKER_BUILDTYPE ]]` ? It seems to contain a string below

ebbit1q: -n is the default operation, it doesn't need to be specified in bash.

ebbit1q:
What is the -DCMAKE_PREFIX_PATH used for, and why is it set to a non existing location on linux? (I'll shield my eyes from the fact that it runs echo in a subshell)

ctrlaltca:
The CMAKE_PREFIX_PATH parameter is used to specify one or more paths that cmake should try first while searching for libraries and binaries. The `/opt/qt5*/lib/cmake/` folder is where some PPAs we used installed custom versions of Qt5; i guess it was the ubuntu 12.x time and the version bundled in the distro was unusable, so we used a more recent version from a PPA.

ebbit1q:
So it is no longer doing anything? What does it do on mac?

ebbit1q:
Maybe a simple -r test?

ctrlaltca:
https://github.com/Cockatrice/Cockatrice/blob/master/.ci/travis-compile.sh#L11
This is only used on travis, and it's the official way to check if you're running on osx.
Checking for "Darwin" is what you would normally do on a Makefile that's supposed to be run on everyone's box.

ebbit1q:
Well that's the thing, the script is now run on a docker container in this pr, making it no longer a travis only script (even though the container runs in travis it's just nicer)

ctrlaltca:
> the script is now run on a docker container in this pr

Right, now i get your point :)
Go on as you prefer, either define new env variables in travis, change how the script is working or even create a new compile script only used in docker builds, whatever fits you better

ebbit1q:
The idea is to have as much similarity between the docker build and the normal build as possible, making sure there's no mistakes there.

.ci/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:bionic
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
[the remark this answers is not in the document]
ebbit1q: Do we actually care about this?

ebbit1q:
separated clangify changes to separate pr: #3435

ebbit1q:
Happy travis, happy pr.

.travis.yml
```diff
@@ -4,50 +4,65 @@ cache: ccache
...
-    os: linux
-    dist: xenial
-    group: stable
-    env: BUILDTYPE=Debug
-  - name: Ubuntu (Release)
+    services: docker
+    before_install: docker build -t img .ci
+    script: docker run --mount "type=bind,source=$(pwd),target=/src" -w="/src" img
+          bash .ci/travis-compile.sh --server "Debug"
+  - name: Ubuntu Bionic (Debug)
```
ctrlaltca: I guess this is a typo, should be `(Release)`

.travis.yml
```diff
@@ -4,50 +4,65 @@ cache: ccache
...
     if: (branch = master AND NOT type = pull_request) OR tag IS present
+    services: docker
+    script: docker build -t img .ci &&
+        docker run --mount "type=bind,source=$(pwd),target=/src" -w="/src" img
+          bash .ci/travis-compile.sh --server "Release"
```
ctrlaltca: Do we need `--package` here, so that it can be uploaded it to the github release?

.travis.yml
```diff
@@ -4,50 +4,65 @@ cache: ccache
...
-
-script: bash ./.ci/travis-compile.sh
+    before_install:
+      - brew update
+      - brew update
+      - brew install ccache
+      - brew unlink python # protobuf python2 install requires this link to be removed
+      - brew install protobuf
+      - brew install qt
+    script: bash ./.ci/travis-compile.sh --server --install "Release"
```
ctrlaltca: Do we need --package here, too, so that it can be uploaded it to the github release?

.travis.yml
ebbit1q: Yes it is, I had to copy all the configs around, this is why we hate WET!

ebbit1q:
I've separated the changes to clangify.sh in a separate pr! Please look at #3435 first!

.ci/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
```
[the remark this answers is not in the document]
Daenyth: ^ I disabled that lint now, I disagree for this file

.ci/travis-dependencies.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-
-if [[ $TRAVIS_OS_NAME == "osx" ]] ; then
-  brew update
```
Daenyth: Why remove the OSX brew instructions?

Daenyth: I see, moved to travis config

tooomm: I think the idea with the external script was to keep the config file itself slimmer and cleaner. Ctrl introduced it back then.

ebbit1q:
Please review this pr.

ebbit1q:
There is no cp step required, this approach gives the docker container full access to the local files and builds in exactly the same location.

ctrlaltca:
> There is no cp step required, this approach gives the docker container full access to the local files and builds in exactly the same location.

Ohh, now i see the magic:
```
    script: docker run --mount "type=bind,source=$(pwd),target=/src" -w="/src" img
```

ebbit1q:
This also means there is no need to clone the repo twice :)

ctrlaltca:
@ebbit1q Make package on ubuntu failed: https://travis-ci.org/Cockatrice/Cockatrice/jobs/456005691#L3024
Looks like it's missing https://packages.ubuntu.com/bionic/file

ctrlaltca:
Also, on mac you can `brew install protobuf --without-python@2`, avoid the `- brew unlink python` and speed up a bit the build

ebbit1q:
Cool, do you want me to make a new pr?

ctrlaltca:
go for it 👍

ebbit1q:
fixed in #3441

PR #3438, opened 2018-11-12
add os version name to package filename

ebbit1q:
Add CMAKE_SYSTEM_VERSION to the PROJECT_VERSION_FILENAME.
Without this package names are the same for all systems using the same package type, causing them to be overwritten.
If this is added we'll be able to support both Ubuntu 16.04 and 18.04 for example.
See #3433 for the pr adding Ubuntu 18.04 compilation to travis ci.

### DIFF
--- a/.ci/travis-compile.sh
+++ b/.ci/travis-compile.sh
@@ -17,7 +17,8 @@ while [[ "$@" ]]; do
       ;;
     '--package')
       MAKE_PACKAGE=1
-      shift
+      PACKAGE_NAME="$2"
+      shift 2
       ;;
     '--server')
       MAKE_SERVER=1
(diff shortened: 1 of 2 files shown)


tooomm:
Am I correct that it'll be named
`Cockatrice-2.6.1-custom(f1af0a4)-4.15.0-1023-gcp` for Ubuntu 16 then?

That's not really great and user friendly...
It might also "brake" other places where we display the version.

ebbit1q:
Well I couldn't find a better variable representing this in CMake, also this keyword is used absolutely nowhere else, not even outside of cockatrice (search for cmake "PROJECT_VERSION_FILENAME" and the top result will be our getversion.cmake file)

ebbit1q:
Oh do we already do that? I guess that's a lot easier as we'll have the exact name of the system already set there (please don't try to extract dist info in a general way that works for all of unix, it's hell)

PR #3441, opened 2018-11-16
fix building packages on travis-docker

ebbit1q:
Fix building packages on travis-docker #3433.


ebbit1q:
Fixed, please review.

.ci/UbuntuBionic/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:bionic
```
ctrlaltca: Looks like you moved the Dockerfile to a subdirectory, but the old file in `.ci/` has not been deleted.

ebbit1q: strange, I guess `git mv` doesn't actually remove the old file?

ctrlaltca:
File fixed, ci succeded

PR #3443, opened 2018-11-18
travis: exclude mac debug build on tagged release

tooomm:
On tagged releases we only build `BUILDTYPE=release` for a speedy deployment.

This got left out in #3433 during the refactor.

PR #3444, opened 2018-11-19
set error message to be conditional

ebbit1q:
#3443 fails on the git merge base for "some reason" but at least this error message should be clearer.


ebbit1q:
this has been merged in #3449

PR #3446, opened 2018-11-20
add ccache to docker compilation

ebbit1q:
Adds ccache to the docker image and shares travis' ccache dir with the image.

### DIFF
--- a/.ci/UbuntuBionic/Dockerfile
+++ b/.ci/UbuntuBionic/Dockerfile
@@ -7,6 +7,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         file \
         g++ \
         git \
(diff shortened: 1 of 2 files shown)


ebbit1q:
This works, let the festivities begin.

ebbit1q:
image
👌

ctrlaltca:
Nice!

tooomm:
That's so great. Down from ~9 minutes! The docker build (bionic on trusty base) is only 15s slower than on xenial directly.
Thank you @ebbit1q 😍 

However I've this concern with the cache: https://github.com/Cockatrice/Cockatrice/pull/3447#issuecomment-440450479
Let me know what you guys think!

PR #3451, opened 2018-11-22
fix travis format checking again

ebbit1q:
Set the exit code properly, "$?" does not work after "if !".
Move "set -e" to after format checks.


ebbit1q:
This works, you can test it by "unformatting" a source file and then running the script `bash .ci/travis-compile.sh --format` while in the project root directory.

ebbit1q:
Note that currently when you do this while on the master branch the script will output nothing and immediately exit 2.

ebbit1q:
Please review this, without this fix format checking generates no output.

PR #3455, opened 2018-11-25
travis speedups

ebbit1q:
After some testing I got all travis tests to below 3 mins, I also made a bit nicer environment for the docker tests.
- Added a script for setting the docker environment in `.ci/travis-docker.sh`
- Stores and loads docker images between pull requests, this gives some speedup (20 - 40%) but causes longer build times if the image needs to be both built and stored.
- Disables updating the cache (including ccache) in prs, this would reupload the entire image again costing 1 to 1.5 minutes. (we can't cache things more selectively on travis)

Alternatively we could keep rebuilding images for all pull requests or save the image somewhere else than in travis' cache.

### DIFF
--- a/.ci/UbuntuBionic/Dockerfile
+++ b/.ci/UbuntuBionic/Dockerfile
@@ -18,4 +18,5 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         qttools5-dev \
         qttools5-dev-tools \
         qtmultimedia5-dev \
+    && apt-get clean
     && rm -rf /var/lib/apt/lists/*
--- a/.ci/UbuntuBionic/Dockerfile
+++ b/.ci/UbuntuBionic/Dockerfile
@@ -18,5 +18,5 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         qttools5-dev \
         qttools5-dev-tools \
         qtmultimedia5-dev \
-    && apt-get clean
+    && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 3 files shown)


ebbit1q:
Well, this clearly doesn't work, at least I made a nice bash script...

Daenyth:
Shoot me a review request once this is in working shape

ebbit1q:
This works, but needs improvements to be valuable at all, current issues are:
- ccache is not updated in between pull requests
- when the image needs to be reuploaded the build will be significantly slower
Solutions would be:
- further slimming of the docker images, this would improve the time travis spends downloading the images
- saving the images somewhere else than on travis, this would separate ccache from the image allowing us to only update ccache without reuploading the image
- still building the images but caching the dependencies outside the image, this would require some more mounting etc
One of the nice things of this is that it adds a nice script for the docker builds separating that from the travis config. You can review the script if you'd like.

ZeldaZach:
How's this looking?

ZeldaZach:
Do we know what we want to do with this @ebbit1q @ctrlaltca ?

tooomm:
Build times aren't too bad for the new docker builds.
Maybe this adds too much code and complexity which is not worth the relatively small improvement for now?

What are your opinions here?

ctrlaltca:
@ebbit1q please have a look at this; if you still want it merged, please resolve the conflict and request a review.

ebbit1q:
I'll close this for now and keep the branch on my fork. The script here has already proven useful for me to test travis compiling locally. It's not needed but useful.

PR #3458, opened 2018-11-27
[WIP] Basic mtgjsonv4 support

### LINKED ISSUE #3406: support for mtgjson v4 (we still use v2)
Advantages as far as I can tell (please correct me and expand the list):
- unique ids
- information on legality
- foreign language support
- token
- spoiler

ctrlaltca:
## Related Ticket(s)
- Fixes partially #3406

I quickly wrapped up the needed changes, here's what i found:
 * it looks like the json file size passed from about 31MB to a whooping 207MB, so it takes a lot more time to parse;
 * a lot of new sets are available, most of them are promo or other special sets;
 * set codes are lowercase, so we need to uppercase them;
 * the "aftermath" card layout has been merged in "split";
 * the "cmc" field is now named "convertedManaCost";
 * the "multiverseid" field is now named "multiverseId";
 * the "colors" field is now using single letter codes;
 
No support for card ids, foreign languages, legality, etc.
Some testing is welcome :)


ebbit1q:
Ignore the macos build, checks are passing.

tooomm:
Two fist notes:
- `Set type` in `Manage sets` window is all written in capital letters now.
It should only have the first letter in capital, unlike `set code`.

- Legalities and foreign names would be very nice additions from v4, yes!

oracle/src/oracleimporter.cpp
```diff
@@ -33,13 +33,10 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
...
         map = it.next().toMap();
-        edition = map.value("code").toString();
+        edition = map.value("code").toString().toUpper();
         editionLong = map.value("name").toString();
         editionCards = map.value("cards");
-        setType = map.value("type").toString();
-        // capitalize set type
-        if (setType.length() > 0)
-            setType[0] = setType[0].toUpper();
+        setType = map.value("type").toString().toUpper();
```
tooomm: `Set type` looks weird now:
set type

ctrlaltca: Fixed, thank you

Daenyth:
@ZeldaZach Given that this is json data, I expect xz or bz compression will yield far better results than gzip in terms of file size.

ZeldaZach:
I can offer bz compression if you would like that

Daenyth:
That would help. I know we have users in Australia where connection is very often metered for bandwidth, so file size savings on common stuff is worth doing

ZeldaZach:
MTGJSONv4 downloads are now available in the following formats:
```
XXX.json
XXX.json.zip
XXX.json.gz
XXX.json.bz2
```

tooomm:
What's the difference in file size @ZeldaZach ?

And I assume we would need to bundle particular software to use .bz2 in Oracle?

ZeldaZach:
```
-rw-r--r-- 1 mtgjson  77M Nov 26 18:03 AllCards.json
-rw-r--r-- 1 mtgjson  11M Nov 28 11:52 AllCards.json.bz2
-rw-r--r-- 1 mtgjson  18M Nov 28 11:51 AllCards.json.gz
-rw-r--r-- 1 mtgjson  19M Nov 28 11:51 AllCards.json.zip
-rw-r--r-- 1 mtgjson 199M Nov 26 18:03 AllSets.json
-rw-r--r-- 1 mtgjson  19M Nov 28 11:57 AllSets.json.bz2
-rw-r--r-- 1 mtgjson  33M Nov 28 11:55 AllSets.json.gz
-rw-r--r-- 1 mtgjson  34M Nov 28 11:54 AllSets.json.zip
```

The BZIP2s appear to be around 10% the size of the original, and half the size of the GZIP/ZIPs.

@ctrlaltca can we look into BZIP2 incorporation into Oracle for this update?

ebbit1q:
Just checked this branch out, seems like it working just fine but a few issues pop up:

A few sets have no name, nothing but a set code and often contain just a single token:
image
I don't know if these can get more descriptive names or could be merged into something like "tokens" but this seems like more of a thing for the json side.

The new version adds alternate arts with a letter between braces behind the card:
image
These are separate cards now, with different names. Some of them load images for the alternates correctly, some don't.

Cards that were only printed in a set with alternates will not be recognized if loaded in a deck with their normal cardnames:
image

Tokens are now cards and can be added to decks, I think these should be sorted beforehand. This isn't a real issue but if we want to get our tokens from here they should definitely be added to the tokens list, no?

I think we should address these before implementing this, extra features can come later.
Also if you compile and run the executable in the build location it'll be unable to find oracle, you can simply run oracle but it's interesting that this function in the main menu simply does not work if you didn't install first.

ZeldaZach:
Hmm, Maybe I should re-think how to do the alternative versions in a set... I'll prob change that with json4.1.2 to not have that issue, except with Un-sets

Daenyth:
@ZeldaZach I'd have a `printing_type` field perhaps

Daenyth:
What's the next step here, that this needs user testing?

ZeldaZach:
V4.1.2 has been released: https://mtgjson.com/v4/docs.html
Addresses the (a), (b) for _most_ sets. Left in place for Un-sets as they are weird.

tooomm:
>What's the next step here, that this needs user testing?

I think we need to have .bz2 support in Oracle before merging this.
Otherwise the file size and time consumed is just not working out for users, even on normal connections.

About supporting more fields, that can be a secondary step.
Especially those that relate to features which are not yet implemented like the rarity filter or foreign cards (and art).

But the meta data would be a very nice-to-have and useful addition right from the start! 👍

ebbit1q:
I tested this again with the new json version, all looks good but some arts aren't loading, is this a related issue?

ebbit1q:
I propose merging this and adding bz2 support separately instead, that way we can test quicker.

ZeldaZach:
I'm testing right now. I see no issues with image downloading for a variety of cards.

ebbit1q:
Well, so far I noticed all original unhinged cards do not load.

ebbit1q:
Correction all unhinged cards that have a (b-f) suffix are not loading. (I ofcourse immediately wanted to test the new suffixes)
They do work for unstable cards however.

ebbit1q:
```PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]: Trying to load picture
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]: Picture not found on disk, trying to download
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]: Trying to download picture from url: "http://gatherer.wizards.com/Handlers/Image.ashx?multiverseid=0&type=card"
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]:Picture downloaded, but blacklisted, will consider it as not found
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]: Trying to load picture
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]: Picture not found on disk, trying to download
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]: Trying to download picture from url: "http://gatherer.wizards.com/Handlers/Image.ashx?name=My First Tome %28b%29&type=card"
PictureLoader: [card:  "My First Tome (b)"  set:  "UNH" ]:Picture downloaded, but blacklisted, will consider it as not found
PictureLoader: [card:  "My First Tome (b)"  set:  "" ]:  Picture NOT found, download failed, no more url combinations to try: BAILING OUT
```

ZeldaZach:
A valid solution could be to use Scryfall's image API as our primary, as the data is directly linked. I tested it and it works for _all_ cards that were imported by JSONv4

`https://api.scryfall.com/cards/multiverse/!cardid!?format=image`

ZeldaZach:
In addition, I also recommend we move to using the token data provided by SF / included in the JSON4 files. Maybe not this PR, but a good step in the right direction

ebbit1q:
Scryfall's api is very complete.

ZeldaZach:
Another thing, we should support "UUID" from mtgjson for card searching reasons :)

ebbit1q:
Make an issue for it?

Daenyth:
I agree we want scryfall, but it 100% needs to be in another PR from this, along with format expansions etc.

ZeldaZach:
Update: Support for XZ has been added. `XXX.json.xz` :)

ctrlaltca:
> Update: Support for XZ has been added. `XXX.json.xz` :)

Cool! I'm already working on the decoder, it works but still has a major bug..

ZeldaZach:
@ctrlaltca Let me know if you need some assistance!

ctrlaltca:
@ZeldaZach i've just pushed the xz-related changes. It quite works, but the decompression fails at the last chunk, like a few kbs are missing from the file. There's probably something wrong on the way the file is read or the size s computed, i left some debug code but had no time to fix it yet.

ZeldaZach:
Hmm, if i have time this weekend I"ll give it a look. Also, we should probably still keep ZIP support around, in case the user doesn't have XZ but has ZIP ?

ebbit1q:
The lzma code doesn't need formatting, add it to the ignore list in clangify.sh.

ZeldaZach:
Note: Do we want to use JSONv4's token system or continue to use the manual one we have?

tooomm:
Why do you priotize `.xz` over `.bz2` btw?

I'm not sure if we need the zip as fallback, as the fallback can always be uncompressed @ZeldaZach ?

Token system can come later I think?

ZeldaZach:
@tooomm XZ is a better compression algorithm for text-based documents, like JSON. 
```bash
-rw-r--r-- 1 mtgjson  73M Nov 30 23:49 AllCards.json
-rw-r--r-- 1 mtgjson  11M Dec  6 23:44 AllCards.json.bz2
-rw-r--r-- 1 mtgjson  18M Dec  6 23:44 AllCards.json.gz
-rw-r--r-- 1 mtgjson 8.6M Dec  6 23:45 AllCards.json.xz
-rw-r--r-- 1 mtgjson  18M Dec  6 23:44 AllCards.json.zip
-rw-r--r-- 1 mtgjson 201M Nov 30 23:49 AllSets.json
-rw-r--r-- 1 mtgjson  19M Dec  6 23:43 AllSets.json.bz2
-rw-r--r-- 1 mtgjson  33M Dec  6 23:42 AllSets.json.gz
-rw-r--r-- 1 mtgjson  17M Dec  6 23:44 AllSets.json.xz
-rw-r--r-- 1 mtgjson  34M Dec  6 23:42 AllSets.json.zip
```

The uncompressed is a real heavy hitter, would like to see zip fallback still...

We can do tokens later, but if we have a PR up why not try?

README.md
```diff
@@ -87,7 +87,8 @@ Dependencies: *(for minimum requirements search our [CMake file](https://github.
 - [protobuf](https://github.com/google/protobuf)
 - [CMake](https://www.cmake.org/)
 
-Oracle can optionally use xz to load compressed files:
+Oracle can optionally use zlib adndxz to load compressed files:
```
tooomm: typo

It's actually `and/or` right? Or is it support for both or none?

tooomm:
There is one issue with clang-format left @ctrlaltca 

If I see things right, this would be our first basic support then, right?
No further changes needed here?

More feature can come in extra PR's to get things started with v4.
We should really have a release around xmas ready maybe!

ctrlaltca:
> There is one issue with clang-format left @ctrlaltca 
Thank you, I'll take care of them when i'm done with the code

> If I see things right, this would be our first basic support then, right?
> No further changes needed here?

The code should be ok, but packages builders are still missing the needed changes

ebbit1q:
There is a space too many on line 424 of oracle/src/oraclewizard.cpp triggering clang-format.

tooomm:
Did you solve the remaining issues @ctrlaltca ?

ctrlaltca:
Current status:
 * new mtgjson4 json strucure parsing works;
 * xz support has been added, zip is still included;
 * the needed libraries / dependencies have been added/declared to packages
 * the appveyor build script has been modified to use vcpkg instead of manually install specific version of the packages. This introduced a problem with openssl's dlls not being picked up correctly, that i'm working on.
To be done:
 * decide if we want to switch from cockatrice's tokens to mtgjson's tokens
 * work out problems with single cards like a/b variants
 * add support for new fields (eg. uuid)

tooomm:
I only see these two as critical and high priority in this PR for now.:

>the appveyor build script has been modified to use vcpkg instead of manually install specific version of the packages. **This introduced a problem with openssl's dlls not being picked up correctly, that i'm working on.**

>work out problems with single cards like a/b variants

What does this mean exactly? Will some cards not be parsed and not be included in our database?

ebbit1q:
a/b variants are limited to unsets right now, as of the latest mtgjson version. They work but if there's issues with them that should be solved on mtgjson not here.

.appveyor.yml
```diff
@@ -24,65 +24,31 @@ clone_depth: 50    #same as travis, see https://www.appveyor.com/blog/2014/06/04
...
-            echo "using zlib from cache"
+        if (Test-Path C:\Tools\vcpkg\installed\x64-windows) {
+          echo "using vcpkg packages from cache"
         } else {
-            Invoke-WebRequest "https://github.com/madler/zlib/archive/v$env:zlib_ver.zip" -OutFile c:\zlib-$env:zlib_ver.zip
-            Expand-Archive -Path c:\zlib-$env:zlib_ver.zip -DestinationPath c:\
-            Set-Location -Path C:\zlib-$env:zlib_ver
-            cmake . -G "$env:cmake_generator" -T "$env:cmake_toolset" -DCMAKE_INSTALL_PREFIX=c:/zlib-release
-            msbuild INSTALL.vcxproj /p:Configuration=Release
+          vcpkg install openssl:$env:vcpkg_arch-windows
```
ras0219-msft: `vcpkg install` is a no-op if the packages are already installed, so I would instead recommend that you always run ```
vcpkg remove --outdated --recurse
vcpkg install openssl protobuf liblzma zlib --triplet $env:vcpkg_arch-windows
```

The remove call will purge any existing libraries that don't match the current commit ID.

Comment (ctrlaltca) [reply .appveyor.yml]: Great advice, thank you!

**Comment by ZeldaZach** [comment]
Was speaking with HSJSON (cc @jleclanche ) and they recommended looking into brotli support instead of XZ

> regarding cockatrice, if they specifically asked for xz because of the file size, i do think its worth talking to them about brotli and supporting that, because:
> 1. its supported natively as an http encoding
> 2. cloudflare will do it for free

**Comment by jleclanche** [comment]
For reference: https://support.cloudflare.com/hc/en-us/articles/200168396

**Comment by jleclanche** [comment]
If cloudflare is already in front of mtgjson, you should already be able to request brotli-compressed streams by adding [`accept-encoding: br;q=1.0, gzip;q=0.8, *;q=0.1`](](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding)) as a header, in front of the plaintext json files. But I'm not sure if there's a size limit above which cloudflare stops compressing. It's possible the source file has to be compressed down first.

**Comment by ctrlaltca** [comment]
Did you check if Qt's QNetworkAccessManager has support for brotli? I don't think we want to include the whole webengine or implement it ourselves.

**Comment by ZeldaZach** [comment]
This is good to go now. Will merge on completed Travis

File: oracle/src/oracleimporter.cpp [hunk 4]
```diff
@@ -33,7 +33,7 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
 
     while (it.hasNext()) {
         map = it.next().toMap();
-        edition = map.value("code").toString();
+        edition = map.value("code").toString().toUpper();
```
tooomm: That's not needed anymore. See https://github.com/mtgjson/mtgjson/pull/196.

Is it worth removing this again?
Would it help with quicker parsing the json and creating our database? @ctrlaltca

tooomm:
New !uuid! also in the wiki now:
https://github.com/Cockatrice/Cockatrice/wiki/Custom-Picture-Download-URLs

ZeldaZach:
We also need to impl Scryfall uuid. Whoops

tooomm:
Is there a ticket with a compiled overview what is supported yet and what's still missing but can provide useful benefits?

PR #3463, opened 2018-11-29
add an ignore button to sort warning

### LINKED ISSUE #3359: Can't sort sets by columns anymore
I assume, given the "While set list is sorted by columns, custom setting is disabled" message, that this is intended behavior, but I'd like to ask you to reconsider, or at least add a toggle for this behavior. Up until now, I've sorted my sets by *oldest* first, so I can see the original art for every card, by sorting by release date and saving, but now I can't do that. And since the new default order is *newest* first, that means I have to hand-reverse every single set. Please, let me just save after sorting by column again.

ebbit1q:
## Related Ticket(s)
- Fixes #3359

The added "feature" allowing you to sort sets while not affecting the actual sorting didn't allow for ignoring this and setting the sorting like that anyway.

- I've simply added an ignore button next to the warning which will remove the warning, sort the sets like current and remove the current sorting view as the model is ordered like this now anyway.

## Screenshots
image

### DIFF
--- a/cockatrice/src/window_sets.cpp
+++ b/cockatrice/src/window_sets.cpp
@@ -139,13 +139,22 @@ WndSets::WndSets(QWidget *parent) : QMainWindow(parent)
     hintsGroupBox = new QGroupBox(tr("Hints"));
     hintsGroupBox->setLayout(hintsGrid);
 
-    sortWarning = new QLabel;
-    sortWarning->setWordWrap(true);
-    sortWarning->setText(
-        "<b>" + tr("Warning: ") + "</b><br>" +
+    sortWarning = new QGroupBox(tr("Warning"));
+    QGridLayout *sortWarningLayout = new QGridLayout;
+    sortWarningText = new QLabel;
+    sortWarningText->setWordWrap(true);
+    sortWarningText->setText(
         tr("While the set list is sorted by any of the columns, custom art priority setting is disabled.") + "<br>" +
         tr("To disable sorting click on the same column header again until this message disappears."));
-    sortWarning->setStyleSheet("QLabel { background-color:red;}");
+    sortWarningLayout->addWidget(sortWarningText, 0, 0);
+    sortWarningButton = new QPushButton;
+    sortWarningButton->setText(tr("Ignore"));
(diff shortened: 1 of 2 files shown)


Daenyth:
I'm not exactly clear on how this works. Can you elaborate on the UX for:
- User wants to sort images by set release date, oldest first, and then bump un-sets to highest priority after doing that?

For example?

ebbit1q:
In this example you'd:
- click the header for release date, causing the same warning to appear
- click the new ignore button, overwriting whatever order the cards were in previously
- filter and order the cards like you could before

Daenyth:
It sounds like the button effect is "allow clicking headers to sort set order" - could we maybe name the button `"Allow column sorting"` with the warning text along the lines of `Sorting sets by column will not currently change image priority order. To override this and set image priority by column sorting, (click button)`

ebbit1q:
I changed the looks of the warning message: 
image

tooomm:
How could the user revert to normal set priority/sorting when he activated the button?

ebbit1q:
Erm, what do you mean? The button doesn't change any behavior, it just takes the current sorting and applies it.

tooomm:
Oh, so the message stays or pops up again when you sort again etc.
I'll need to check in client.

ebbit1q:
Yeah, that the entire order isn't overthrown when sorting by column is intended to be a feature.
How often will you want to do this anyway?

tooomm:
No idea, the thing with the message was a workaround to improve the former situation.
It's still not great and this might help as well.

But a good solution needs a new approach I think.
Like split sorting sets and managing (enable/disable) sets completely since the combined functionality is complex and difficult to understand. No matter how often you use it

ebbit1q:
Alternatively we could simply add the card priority as a number, right next to the enable/disabled. This should be clear to users and we could allow editing it etc. That way we can simply sort the cards by card priority by default and add a separate button for adapting the current sort as the new priority numbering.

tooomm:
Hmm, so people could sort by that column as well and it would be more obvious or self-explaining. It could work nicely.

What do others think here?

ZeldaZach:
What do we want to do with this?

PR #3465, opened 2018-12-03
place the os nametag after the version

ebbit1q:
When i added the appended os nametags I didn't account for the dots inside the version, this will place the nametag before the last dot instead of after the first dot.


ebbit1q:
Review me! I'm a tiny pr! @Daenyth

ebbit1q:
Anyone! Review me!

PR #3472, opened 2018-12-12
fix rpm builds in travis

ebbit1q:
## Related Ticket(s)
- Fixes #I can't find the issue rn

Fedora builds in travis create debian packages instead of rpm ones

- the correct argument will be passed to cmake
- the missing dependency will be added to the image file


ebbit1q:
I tested it on fedora, the package works.

tooomm:
rpm deployment worked 👍

PR #3480, opened 2018-12-24
[WIP] Oracle: Optimize json parsing

ctrlaltca:
## Related Ticket(s)
- #3458

Since the move from Mtgson2 to Mtgjson4, the json file size increased a lot. This resulted in increased parsing time in oracle, eg. on my mac from 5-10 second to over a minute. We are not even reporting a progress indicator to users.
Back in the Qt4 days, Qt had no native support for Json, so Cockatrice is embedding a 3rd party library called qt-json: https://github.com/qt-json/qt-json
Nowadays Qt5 has a native solution to parse Json, but it's so optimized for speed on small files that it doesn't handle big json files at all. I've tried to use it, but it end up with a `too large document` error.

While investigating the speed of the json library, I've found out that a couple optimizations could speed up the parsing by about 3 times. The original patch was found here: https://github.com/qt-json/qt-json/issues/20
On my mac, the parsing time is reduced from 63 to 21 seconds.

I'm keeping this PR open a couple of days to investigate other possible optimizations.

### DIFF
--- a/cockatrice/src/qt-json/json.cpp
+++ b/cockatrice/src/qt-json/json.cpp
@@ -1,28 +1,28 @@
 /* Copyright 2011 Eeli Reilin. All rights reserved.
  *
- * Redistribution and use in source and binary forms, with or without 
+ * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
  *
- * 1. Redistributions of source code must retain the above copyright notice, 
+ * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  *
  * 2. Redistributions in binary form must reproduce the above copyright notice,
- *    this list of conditions and the following disclaimer in the documentation 
+ *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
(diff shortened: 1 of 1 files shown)


ebbit1q:
Isn't it faster to use regular expressions instead of indexof?

tooomm:
Parsing time is considerably faster now, which is awesome. 👍 

Could you still look into implementing a progress bar for that part, please? @ctrlaltca

ebbit1q:
regular expressions are several orders of magnitude slower :/

ebbit1q:
Have you looked at using qjson ( http://qjson.sourceforge.net/ ) instead? We don't have to maintain a json parser ourselves.
Also qt's parser can't be used, see this bug: https://bugreports.qt.io/browse/QTBUG-58652

ctrlaltca:
> Have you looked at using qjson ( http://qjson.sourceforge.net/ ) instead?
```
fab$ ./bin/cmdline_tester /Users/fab/Downloads/AllSets.json
Parsing of "/Users/fab/Downloads/AllSets.json" took 14377 ms
```
Looks like a good candidate, performance is comparable with the current implementation.

PR #3481, opened 2018-12-28
Minor translation issue fix

Caledor:
## Related Ticket(s)
No issue opened cause it's not worth it for such a simple problem

We have 2 identical "Draw" strings in ui_shortcutstab.h. One refers the action of drawing a card (first screenshot), the other to the Draw Step (second screenshot). This causes side effects when translating said strings.

It merely changes one of the two strings so that both can be translated separately: I went for the first one.

## Screenshots
cattura
cattura2


ZeldaZach:
Appreciate the fix. Keep up the great work! :)

PR #3487, opened 2019-01-03
Switch from (s) to plural form whenever possible in message log

Caledor:
## Related Ticket(s)
- No ticked opened

- 3 strings that appear in the game log don't use plural forms, and opt for "(s)" instead
- Various missing tr()s

- Straight up fix for "%1 draws %2 card(s)".
- "%1 is looking at the top %2 card(s) %3" will allow translating into both "at the top card" and "at the top x cards" by switching arguments 2 and 3 as mentioned in tr description.
- For "%1 places %2 %3 counter(s) on %4 (now %5)." i just moved "counter(s)" into argument %3 as plural form is already used there for each color.
- Add the tr()

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -354,10 +354,11 @@ void MessageLogWidget::logDumpZone(Player *player, CardZone *zone, int numberCar
                                     .arg(sanitizeHtml(player->getName()))
                                     .arg(zone->getTranslatedName(zone->getPlayer() == player, CaseLookAtZone)));
     else
-        appendHtmlServerMessage(tr("%1 is looking at the top %3 card(s) %2.", "top card for singular, top %3 cards for plural", numberCards)
-                                    .arg(sanitizeHtml(player->getName()))
-                                    .arg(zone->getTranslatedName(zone->getPlayer() == player, CaseTopCardsOfZone))
-                                    .arg("<font color=\"blue\">" + QString::number(numberCards) + "</font>"));
+         appendHtmlServerMessage(
+             tr("%1 is looking at the top %3 card(s) %2.", "top card for singular, top %3 cards for plural", numberCards)
+                 .arg(sanitizeHtml(player->getName()))
+                 .arg(zone->getTranslatedName(zone->getPlayer() == player, CaseTopCardsOfZone))
+                 .arg("<font color=\"blue\">" + QString::number(numberCards) + "</font>"));
 }
 
 void MessageLogWidget::logFlipCard(Player *player, QString cardName, bool faceDown)
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -156,7 +156,7 @@ void MessageLogWidget::logAlwaysRevealTopCard(Player *player, CardZone *zone, bo
 
 void MessageLogWidget::logAttachCard(Player *player, QString cardName, Player *targetPlayer, QString targetCardName)
 {
-    appendHtmlServerMessage(QString("%1 attaches %2 to %3's %4.")
+    appendHtmlServerMessage(tr("%1 attaches %2 to %3's %4.")
                                 .arg(sanitizeHtml(player->getName()))
                                 .arg(cardLink(cardName))
                                 .arg(sanitizeHtml(targetPlayer->getName()))
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -341,7 +341,7 @@ void MessageLogWidget::logDrawCards(Player *player, int number)
         mulliganPlayer = player;
     else {
         soundEngine->playSound("draw_card");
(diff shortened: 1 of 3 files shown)


tooomm:
Tried to fix the formatting, but it didn't work out :P

Caledor:
Giving this a closer look, the check wants one initial space less at each line marked with "!". I also found the check before your commit in my browser history and they match. Is it worth a try?

cockatrice/src/messagelogwidget.cpp
```diff
@@ -341,7 +341,7 @@ void MessageLogWidget::logDrawCards(Player *player, int number)
         mulliganPlayer = player;
     else {
         soundEngine->playSound("draw_card");
-        appendHtmlServerMessage(tr("%1 draws %2 card(s).")
+        appendHtmlServerMessage(tr("%1 draws %2 card(s).", "", number)
```
ctrlaltca: IIRC in order to work this string needs to contain a `%n` in place of `%2`.

cockatrice/src/messagelogwidget.cpp
```diff
@@ -354,10 +354,11 @@ void MessageLogWidget::logDumpZone(Player *player, CardZone *zone, int numberCar
                                     .arg(sanitizeHtml(player->getName()))
                                     .arg(zone->getTranslatedName(zone->getPlayer() == player, CaseLookAtZone)));
     else
-        appendHtmlServerMessage(tr("%1 is looking at the top %2 card(s) %3.")
-                                    .arg(sanitizeHtml(player->getName()))
-                                    .arg("<font color=\"blue\">" + QString::number(numberCards) + "</font>")
-                                    .arg(zone->getTranslatedName(zone->getPlayer() == player, CaseTopCardsOfZone)));
+        appendHtmlServerMessage(
+            tr("%1 is looking at the top %3 card(s) %2.", "top card for singular, top %3 cards for plural", numberCards)
```
ctrlaltca: I don't get why the arguments have been reordered 1 - 3 - 2

cockatrice/src/messagelogwidget.cpp
```diff
@@ -626,20 +627,20 @@ void MessageLogWidget::logSetCardCounter(Player *player, QString cardName, int c
...
+        finalStr = tr("%1 places %2 %3 on %4 (now %5).");
     else
-        finalStr = tr("%1 removes %2 %3 counter(s) from %4 (now %5).");
+        finalStr = tr("%1 removes %2 %3 from %4 (now %5).");
 
     QString colorStr;
     switch (counterId) {
         case 0:
-            colorStr = tr("red", "", delta);
+            colorStr = tr("red counter(s)", "", delta);
```
ctrlaltca: I guess this has been moved here because in some languages the adjective  also needs the singular/plural form (eg. un segnalino rosso / due segnalini rossi), am I correct?

Caledor: yes. also i'm italian and i wanted to talk with you about the italian translation since these days i've changed a lot of stuff (mostly to use official terms and whatnot). i tried using the suggestion feature on transifex but it would take forever, so i thought i'd just send you the ts file i'm working on

cockatrice/src/messagelogwidget.cpp
Caledor: nope, it works. i've compiled cockatrice and tested those changes.

tooomm: The documentation has %n, too:
http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

Maybe transifex needs it to properly display the various translations options?

cockatrice/src/messagelogwidget.cpp
tooomm: Right, since @ctrlaltca reviewed most strings they are fixed and normal translators can only send suggestions. Are there that many changes required?
I personally never used the suggest feature and don't know how cumbersome it is to use...

cockatrice/src/messagelogwidget.cpp
Caledor: I think transifex just reads the numerous and numerusform tags from the ts file to handle that but i don't know for sure

cockatrice/src/messagelogwidget.cpp
Caledor: >

cockatrice/src/messagelogwidget.cpp
ebbit1q: The `%n` can be used, but doesn't have to, all instances of `%n` would be converted to the value of `number`. We can simply pass the number a second time as an `.arg()`, that way we can easily pass the font color html tags as well.

cockatrice/src/messagelogwidget.cpp
ctrlaltca: This makes sense, thank you :)

ctrlaltca:
Thank you for the clarifications.
I've added you as a reviewer for the Italian team on Transifex, so you have the ability to change what's already reviewed.
If you want to me bulk upload a ts file, just send it to my nickname at gmail dot com.

PR #3490, opened 2019-01-06
Make deck editor filter translatable ; extracted new translations; Fix #3488

### LINKED ISSUE #3488: Strings missing in ts files
Need to manually add the strings returned by CardFilter::attrName (at https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/cardfilter.cpp) in the ts file, context FilterBuilder.

The corresponding tr() is already called at https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/filterbuilder.cpp#L15

ctrlaltca:
## Related Ticket(s)
- Fixes #3488

### DIFF
--- a/cockatrice/src/cardfilter.cpp
+++ b/cockatrice/src/cardfilter.cpp
@@ -1,47 +1,47 @@
 #include "cardfilter.h"
 
-const char *CardFilter::typeName(Type t)
+const QString CardFilter::typeName(Type t)
 {
     switch (t) {
         case TypeAnd:
-            return "AND";
+            return QString("AND");
         case TypeOr:
-            return "OR";
+            return QString("OR");
         case TypeAndNot:
-            return "AND NOT";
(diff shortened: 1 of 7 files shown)


cockatrice/src/cardfilter.cpp
```diff
@@ -1,47 +1,47 @@
 #include "cardfilter.h"
 
-const char *CardFilter::typeName(Type t)
+const QString CardFilter::typeName(Type t)
 {
     switch (t) {
         case TypeAnd:
-            return "AND";
+            return QString("AND");
```
ebbit1q: Should these not be translated?

ctrlaltca: I guess this is a good idea; maybe it's not easy to find short, meaningful terms for these logical operators for every language out there, but let translators decide is probably the best choice.

PR #3491, opened 2019-01-06
Fix shortcuts on mac; fix #3482

### LINKED ISSUE #3482: Shortcuts loosing focus
Cockatrice 2.6.2

Shortcuts seem to lose focus when certain actions take place (Rolling dice I believe is one of them).

Came about with the update to 2.6.2. Investigation is needed

ctrlaltca:
## Related Ticket(s)
- Fixes #3482

Due to recent changes to our travis script, we are using the old, known problematic version of Qt 5.9.3 in our builds. The original issues are described here: https://github.com/Cockatrice/Cockatrice/issues/2647 

As a first attempt, we let homebrew update itself to get a recent, working Qt version installed.
The brew update+install step requires about 7 mins, and the entire build time goes up to 13 mins.
As an alternative, we could try a more recent macOS image on Travis, but this means dropping support for macOS 10.12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,7 +85,6 @@ matrix:
     if: tag IS NOT present
     os: osx
     osx_image: xcode9.2
-    env: HOMEBREW_NO_AUTO_UPDATE=1
     cache: ccache
     before_install:
       - brew install ccache protobuf qt xz
@@ -95,7 +94,6 @@ matrix:
     if: (branch = master AND NOT type = pull_request) OR tag IS present
     os: osx
     osx_image: xcode9.2
-    env: HOMEBREW_NO_AUTO_UPDATE=1
     cache: ccache
     before_install:
       - brew install ccache protobuf qt xz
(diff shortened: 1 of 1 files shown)


ebbit1q:
Ah, 13 mins is always better than windows.

ZeldaZach:
Would rather have homebrew only update if it detects a build of master (i.e. release)

ebbit1q:
Instead of disabling homebrew updates we could simply use a newer image for faster builds, then use an older image for releases.

ctrlaltca:
CI failed for an unrelated problem in the Fedora build:
```
/usr/lib64/qt5/bin/lrelease: error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
```

ebbit1q:
:sweat:

ctrlaltca:
I'd go with this. Remember to self: ensure the release configuration is working on the next prerelease tag.

ZeldaZach:
Seems good to go for me

ZeldaZach:
New release incoming soon cc @ctrlaltca

PR #3494, opened 2019-01-08
Several download URLs have arrived!

ZeldaZach:
Not sure which (if any) tickets this solves.

Cockatrice Picture loader uses better defined URLs now. URLs are defined on the Card Properties tab. Instead of Primary/Backup, you can now define a list of URLs. List of URLs can be drag/dropped for priority ordering.

ALSO: Oracle now uses `scryfallId` over `mtgjsonUUID` for `!uuid!` in URLs.

Before:
screenshot 2019-01-07 20 24 14

After:
screenshot 2019-01-07 20 24 30
screenshot 2019-01-07 20 24 37

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -449,7 +449,9 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     urlList->setAlternatingRowColors(true);
     urlList->setDragEnabled(true);
     urlList->setDragDropMode(QAbstractItemView::InternalMove);
-    connect(urlList, SIGNAL(itemSelectionChanged()), this, SLOT(storeSettings())); // TODO: Make this signal better
+    connect(urlList->model(),
+            SIGNAL(rowsMoved(const QModelIndex, int, int, const QModelIndex, int)), this,
+            SLOT(urlListChanged(const QModelIndex, int, int, const QModelIndex, int)));
 
     for (int i = 0; i < settingsCache->downloads().getCount(); i++)
         urlList->addItem(settingsCache->downloads().getDownloadUrlAt(i));
@@ -589,6 +591,11 @@ void DeckEditorSettingsPage::storeSettings()
         qInfo() << "Priority" << i << ":" << urlList->item(i)->text();
         settingsCache->downloads().setDownloadUrlAt(i, urlList->item(i)->text());
     }
+}
+
+void DeckEditorSettingsPage::urlListChanged(const QModelIndex &parent, int start, int end, const QModelIndex &destination, int row)
+{
+    storeSettings();
 }
 
 void DeckEditorSettingsPage::updateSpoilers()
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -45,8 +45,6 @@ GeneralSettingsPage::GeneralSettingsPage()
             languageBox.setCurrentIndex(i);
     }
 
-    picDownloadCheckBox.setChecked(settingsCache->getPicDownload());
-
     // updates
     QList<ReleaseChannel *> channels = settingsCache->getUpdateReleaseChannels();
     foreach (ReleaseChannel *chan, channels) {
@@ -64,26 +62,14 @@ GeneralSettingsPage::GeneralSettingsPage()
     pixmapCacheEdit.setValue(settingsCache->getPixmapCacheSize());
     pixmapCacheEdit.setSuffix(" MB");
 
-    defaultUrlEdit = new QLineEdit(settingsCache->getPicUrl());
-    fallbackUrlEdit = new QLineEdit(settingsCache->getPicUrlFallback());
-
     showTipsOnStartup.setChecked(settingsCache->getShowTipsOnStartup());
 
-    connect(&clearDownloadedPicsButton, SIGNAL(clicked()), this, SLOT(clearDownloadedPicsButtonClicked()));
     connect(&languageBox, SIGNAL(currentIndexChanged(int)), this, SLOT(languageBoxChanged(int)));
-    connect(&picDownloadCheckBox, SIGNAL(stateChanged(int)), settingsCache, SLOT(setPicDownload(int)));
     connect(&pixmapCacheEdit, SIGNAL(valueChanged(int)), settingsCache, SLOT(setPixmapCacheSize(int)));
     connect(&updateReleaseChannelBox, SIGNAL(currentIndexChanged(int)), settingsCache,
             SLOT(setUpdateReleaseChannel(int)));
     connect(&updateNotificationCheckBox, SIGNAL(stateChanged(int)), settingsCache, SLOT(setNotifyAboutUpdate(int)));
-    connect(&picDownloadCheckBox, SIGNAL(clicked(bool)), this, SLOT(setEnabledStatus(bool)));
-    connect(defaultUrlEdit, SIGNAL(textChanged(QString)), settingsCache, SLOT(setPicUrl(QString)));
-    connect(fallbackUrlEdit, SIGNAL(textChanged(QString)), settingsCache, SLOT(setPicUrlFallback(QString)));
-    connect(&defaultUrlRestoreButton, SIGNAL(clicked()), this, SLOT(defaultUrlRestoreButtonClicked()));
-    connect(&fallbackUrlRestoreButton, SIGNAL(clicked()), this, SLOT(fallbackUrlRestoreButtonClicked()));
     connect(&showTipsOnStartup, SIGNAL(clicked(bool)), settingsCache, SLOT(setShowTipsOnStartup(bool)));
-
-    setEnabledStatus(settingsCache->getPicDownload());
 
     auto *personalGrid = new QGridLayout;
     personalGrid->addWidget(&languageLabel, 0, 0);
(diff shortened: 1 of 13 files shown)


Daenyth:
I'm a little concerned with

> ALSO: Oracle now uses scryfallId over mtgjsonUUID for !uuid! in URLs.

Did we ever ship `!uuid!` as a default value in any templates? As far as I know, it was only ever `muid` that we defaulted with. If that's the case, I'm fine with that change since it's unlikely to negatively affect anyone.

cockatrice/src/dlg_settings.cpp
```diff
@@ -515,9 +446,45 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+    urlList->setSelectionMode(QAbstractItemView::SingleSelection);
+    urlList->setAlternatingRowColors(true);
+    urlList->setDragEnabled(true);
+    urlList->setDragDropMode(QAbstractItemView::InternalMove);
+    connect(urlList, SIGNAL(itemSelectionChanged()), this, SLOT(storeSettings())); // TODO: Make this signal better
+
+    for (int i = 0; i < settingsCache->downloads().getCount(); i++)
+        urlList->addItem(settingsCache->downloads().getDownloadUrlAt(i));
+
+    auto aAdd = new QAction(this);
```
Daenyth: Man `auto` is nice :)

ZeldaZach:
We shipped it with 2.6.2 as mtgjsonUUID, which isn't as helpful since i don't have an image repo for it, nor will I. scryfallId is more useful, and it would allow for better compatibility with our URLs

cockatrice/src/dlg_settings.cpp
```diff
@@ -543,6 +510,91 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+/**
+ * If reset, this method contains the default URLs we will populate
+ */
+void DeckEditorSettingsPage::resetDownloadedURLsButtonClicked()
+{
+    urlList->clear();
+
+    urlList->addItem("https://api.scryfall.com/cards/!uuid!?format=image");
+    urlList->addItem("https://api.scryfall.com/cards/multiverse/!cardid!?format=image");
+    // urlList->addItem("http://gatherer.wizards.com/Handlers/Image.ashx?multiverseid=!cardid!&type=card");
```
Daenyth: I think we should leave the gatherer urls in as lowest priority fallbacks.

In any case they shouldn't be committed commented out.

cockatrice/src/dlg_settings.cpp
```diff
@@ -543,6 +510,91 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+    // urlList->addItem("http://gatherer.wizards.com/Handlers/Image.ashx?name=!name!&type=card");
+
+    storeSettings();
+    QMessageBox::information(this, tr("Success"), tr("Download URLs have been reset."));
```
Daenyth: We should `qInfo` the url list each time it's changed, I think.

cockatrice/src/dlg_settings.cpp
```diff
@@ -543,6 +510,91 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+}
+
+void DeckEditorSettingsPage::clearDownloadedPicsButtonClicked()
+{
+    QString picsPath = settingsCache->getPicsPath() + "/downloadedPics/";
+    QStringList dirs = QDir(picsPath).entryList(QDir::AllDirs | QDir::NoDotAndDotDot);
+    bool outerSuccessRemove = true;
+    for (int i = 0; i < dirs.length(); i++) {
```
Daenyth: Please use `foreach` over index iteration, or one of the `std` iterating approaches.

Also is there really no recursive delete option in Qt? That seems shocking that it isn't built in.

cockatrice/src/dlg_settings.cpp
```diff
@@ -543,6 +510,91 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+            bool success = QDir(picsPath).rmdir(dirs.at(i));
+            if (!success) {
+                qDebug() << "Failed to delete inner directory" << picsPath;
+            }
+        }
+    }
+    if (outerSuccessRemove)
+        QMessageBox::information(this, tr("Success"), tr("Downloaded card pictures have been reset."));
+    else
+        QMessageBox::critical(this, tr("Error"), tr("One or more downloaded card pictures could not be cleared."));
```
Daenyth: Please `qInfo` each file path as its deleted - then this message can prompt the user to check the log window for more information. As is, this isn't actionable.

cockatrice/src/dlg_settings.cpp
```diff
@@ -543,6 +510,91 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+}
+
+void DeckEditorSettingsPage::actAddURL()
+{
+    bool ok;
+    QString msg = QInputDialog::getText(this, tr("Add URL"), tr("URL:"), QLineEdit::Normal, QString(), &ok);
+    if (ok) {
```
Daenyth: Man Qt has some butt-ugly apis :(

cockatrice/src/dlg_settings.cpp
```diff
@@ -543,6 +510,91 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
...
+        urlList->addItem(msg);
+        storeSettings();
+    }
+}
+
+void DeckEditorSettingsPage::actRemoveURL()
+{
+    if (urlList->currentItem()) {
```
Daenyth: Is this right? You say if item, delete current row.

Looking at the docs, it seems like `currentItem` returns a pointer - can you do an explicit `nullptr` comparison instead of using it as bool-ish?

cockatrice/src/dlg_settings.cpp
```diff
@@ -689,20 +745,24 @@ MessagesSettingsPage::MessagesSettingsPage()
...
+    aRemove->setStatusTip("Remove URL");
     connect(aRemove, SIGNAL(triggered()), this, SLOT(actRemove()));
 
     auto *messageToolBar = new QToolBar;
     messageToolBar->setOrientation(Qt::Vertical);
     messageToolBar->addAction(aAdd);
     messageToolBar->addAction(aRemove);
     messageToolBar->addAction(aEdit);
-    messageToolBar->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::MinimumExpanding);
+    // messageToolBar->setSizePolicy(QSizePolicy::Preferred);
```
Daenyth: ?

cockatrice/src/dlg_settings.cpp
```diff
@@ -689,20 +745,24 @@ MessagesSettingsPage::MessagesSettingsPage()
...
+    aAdd->setStatusTip("Add New URL");
+
     connect(aAdd, SIGNAL(triggered()), this, SLOT(actAdd()));
     aEdit = new QAction(this);
     aEdit->setIcon(QPixmap("theme:icons/pencil"));
+    aEdit->setStatusTip("Edit URL");
     connect(aEdit, SIGNAL(triggered()), this, SLOT(actEdit()));
     aRemove = new QAction(this);
     aRemove->setIcon(QPixmap("theme:icons/decrement"));
+    aRemove->setStatusTip("Remove URL");
```
Daenyth: Missing `tr` all over the place here.

cockatrice/src/pictureloader.cpp
```diff
@@ -121,10 +95,8 @@ QString PictureToLoad::getSetName() const
     }
 }
 
-QStringList PictureLoaderWorker::md5Blacklist = QStringList()
-                                                << "db0c48db407a907c16ade38de048a441"; // card back returned
-                                                                                       // by gatherer when
-                                                                                       // card is not found
+// Card back returned by gatherer when card is not found
```
Daenyth: How does scryfall behave here? I'm presuming it gives a 404 instead of the cardback?

cockatrice/src/pictureloader.cpp
```diff
@@ -227,22 +202,22 @@ bool PictureLoaderWorker::cardImageExistsOnDisk(QString &setName, QString &corre
     // Iterates through the list of paths, searching for images with the desired
     // name with any QImageReader-supported
     // extension
-    for (int i = 0; i < picsPaths.length(); i++) {
-        imgReader.setFileName(picsPaths.at(i));
+    for (const auto &picsPath : picsPaths) {
```
Daenyth: ty

oracle/src/oracleimporter.cpp
```diff
@@ -97,7 +97,7 @@ CardInfoPtr OracleImporter::addCard(const QString &setName,
         bool mArtifact = false;
         if (cardType.endsWith("Artifact")) {
             for (int i = 0; i < cardTextRows.size(); ++i) {
-                cardTextRows[i].remove(QRegularExpression("\\\".*?\\\""));
+                cardTextRows[i].remove(QRegularExpression(R"(\".*?\")"));
```
Daenyth: Nifty

cockatrice/src/dlg_settings.cpp
ZeldaZach: Sure

cockatrice/src/dlg_settings.cpp
ZeldaZach: Sorry mate </3

cockatrice/src/dlg_settings.cpp
ZeldaZach: So I took this from the macros code exactly as it was. But sure, I can make it `!= nullptr` if that's what you'd prefer

cockatrice/src/dlg_settings.cpp
ZeldaZach: Haha sorry <3

cockatrice/src/dlg_settings.cpp
ZeldaZach: Whoops

cockatrice/src/pictureloader.cpp
ZeldaZach: That would be correct. This is kept here for Gatherer issues.

oracle/src/oracleimporter.cpp
ZeldaZach: Yep! Intellij is very smart :)

oracle/src/oracleimporter.cpp
```diff
@@ -218,18 +219,18 @@ int OracleImporter::importTextSpoiler(CardSetPtr set, const QVariant &data)
...
         QList<QVariantMap> maps = splitCards.values(muid);
         QStringList names;
         // now, reorder the cards using the ordered list of names
         QMap<int, QVariantMap> orderedMaps;
-        foreach (QVariantMap map, maps) {
+        for (const QVariantMap &inner_map : maps) {
             if (names.isEmpty())
-                names = map.contains("names") ? map.value("names").toStringList() : QStringList();
-            QString name = map.value("name").toString();
+                names = inner_map.contains("names") ? inner_map.value("names").toStringList() : QStringList();
```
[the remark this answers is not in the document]
ZeldaZach: I didn't really modify these, just did a replacement of the var name to reduce a warning. It's fine to leave as is for now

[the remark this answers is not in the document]
ebbit1q: Ok then...

ebbit1q:
What happens to users with these settings currently changed?

ZeldaZach:
@ctrlaltca Thanks for the ticket nums. Unfortunately I couldn’t figure out how to make it work with the drag and drop process, as I couldn’t find the right signal. If you have suggestions please let me know. 

I tried with a few methods but they all said “signal not found” even though they should have been there. Qt5.12

ebbit1q:
They lose them...

PR #3513, opened 2019-01-22
Disable darkmode on MacOS

basicer:

On MacOS, when compiling with Qt5.12, dark mode looks pretty bad.

- Dark mode will be prevented

## Screenshots
image
(What it looks like without this patch.  With it looks normal.)


tooomm:
Does it make sense to include this if we get qt@5.9 from homebrew soon and ship that instead 5.12?
I guess it’s a temporary bug that will be fixed when we switch to 5.12 again?

PR #3524, opened 2019-01-27
New json parser: fix split cards; Fix #3523

### LINKED ISSUE #3523: Split Cards Not Showing Up
<b>System Information:</b>
  
Client Version: 2.6.2-custom(37f9f62) (2019-01-25)
Client Operating System: Debian GNU/Linux buster/sid
Build Architecture: 64-bit
Qt Version: 5.11.3
System Locale: en_US
  
Hello everyone,

After getting the latest card database update for RNA, I can no longer use (or find) split cards. They come up under "Unknown" in the deck builder (when loading a pre-existing deck file with a split card) and have no card art loaded.

I updated to 2.6.2 *after* getting the new set, and the problem persisted. I re-downloaded and re-updated the card database a couple of times after this, each time the problem persisted.

If it helps any, grepping about in the ~/.local/share/Cockatrice/Cockatrice/cards.xml file, I was *not* able to find *any* split cards, even from previous expansions like Amonkhet. I am still able to find these split cards in the AllCards.json file from the URL https://mtgjson.com/json/AllSets.json.zip . I found their names are in the "names" field, which is a two-element list containing both portions of the split card name.

I am guessing something is being lost in translation from the json to xml.

Sincerely,

PySean

ctrlaltca:
## Related Ticket(s)
- Fixes #3523

Due to an error, not all split cards were imported correctly.

All split cards are imported correctly.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -311,8 +311,7 @@ int OracleImporter::importCardsFromSet(CardSetPtr currentSet, const QList<QVaria
                 }
             }
         }
-
-        CardInfoPtr newCard = addCard(name, text, isToken, properties, relatedCards, setInfo);
+        CardInfoPtr newCard = addCard(nameSplit, text, isToken, properties, relatedCards, setInfo);
         numCards++;
     }
 


ctrlaltca:
Fast-merging a fix-only trivial change

PR #3531, opened 2019-01-27
Judge mode

basicer:
- When judges enter a game to fix problems, it can be difficult to get the players to do the right thing.
- Some non-magic involve having a deal, banker, or game master that should be able to manipulate all the cards.
- When cockatrice to players, it can be helpful to just do some things for them while they are still learning.

- Adds a new flag to users on the server, "Judge".
- Admins can add/remove judges the same as moderators.
- Adds a new flag to players in a game "Judge".  Judes can be both players and spectators.
- A judge can compel a player to take an action by using that player's player area (this is similar to how offline multiplayer work).
- A judge has a special gavel indicator in the player list.
- If a judge uses this power, a log message is generated telling players the judge is using their special ability.
- Server Judges can join as a judge by holding down shift when the join a game.
- A host can create a game as a judge by holding down shift as they create the game, but only if the server has a special flag set (defaults to off).

## Screenshots
image


ZeldaZach:
I'm reading this over and the code seems decent, just have some questions.

1) Who are "judges?" Are they defined by the server operator, similar to the Donator, Vip, Moderator etc. stuff, or are the defined by the game creator?

2) The shift key thing is a bit awkward for me. Judges should be allowed in any room, and not be allowed to bypass passwords. Maybe have a checkbox on by the "Join Room" button that is similar to `[ x ] Judge Mode` would be more helpful

3) I see moderators can bypass with shift, which is fine by me as moderators are defined as more advanced users of the program. Up to you if you want to include the judge mode checkbox I defined above for them. Might be clever, and make sure to incorporate it into the mod powers enabled/disabled menu so if they are not normally a judge, it would grey out / uncheck if they are not on mod power mode (albeit this doesn't happen much anymore, but still)

Overall, this is awesome!! Keep up the great work, and welcome to Cockatrice development :)

basicer:
I'm actually not sure who judges are yet, which is why I didn't add the checkbox to the create_game dialog.  I see the primary users of this being the judges of MagicLeague if that ever gets going again or some future similar thing  

One thing I don't want for example is people creating their games with judge mode for no reason, and using it to cheat or troll players who don't yet understand judge mode.

ZeldaZach:
Sounds like we'll need to add a new rank to the system. Can you update the database in some way to allow for people to be a judge in addition to other titles. Currently we have an Admin/Moderator/None option and a VIP/Donator/None option

basicer:
I'm just wondering if it is ever the case you trust people enough to be a judge but not a moderator.

skwerlman:
> I'm just wondering if it is ever the case you trust people enough to be a judge but not a moderator.

a judge is a different role from a moderator. a moderator is delegated authority over an entire server; a judge should only be given authority over _the games on the server people want judges in_.

> Maybe have a checkbox on by the "Join Room" button that is similar to [ x ] Judge Mode would be more helpful

imo a "join game as judge" button that appeared for judges on to games that allow judges would be better.

> Adds a new flag to players in a game "Judge". Judes can be both players and spectators.

i'm not a fan of judges being allowed to both play and judge in the same game at the same time; it should really be an either-or kind of thing just to mitigate the potential for abuse

> Some non-magic involve having a deal, banker, or game master that should be able to manipulate all the cards.

I think this case deserves a separate descriptor, and should be per-game, rather than per-server. creating a game with "judge" permissions should be visibly distinct from a user the server op has explicitly granted judge authority to

ZeldaZach:
Seeing if there are any updates on this :)

basicer:
image

tooomm:
1) I think the little icons in the pawns are not ideal. They are really difficult to recognize.

2) Also, I had the idea to make the judge actions/log messages a different color in the log.

3) Do you want to stick with the hammer, or do you prefer that judge/law scale instead?

basicer:
> 1. I think the little icons in the pawns are not ideal. They are really difficult to recognize.

For sure.  We've been talking about reworking the pawn system completely.  Ideally we would unify the language, and do more things procedually (like how the gavel is burned in now).  

> 2. Also, I had the idea to make the judge actions/log messages a different color in the log.

What color were you thinking?

> 3. Do you want to stick with the hammer, or do you prefer that judge/law scale instead?

No preference, whatever looks better.

tooomm:
>What color were you thinking?

Maybe just black as all log is colored currently?

tooomm:
Can you post a screenshot on how the log and player area are looking in the current version, please?

I would also consider the "Scale" icon over the gavel: https://icons8.com/icon/61021/measurement-scale

While the gavel is a perfectly fine symbol for that purpose in itself, it looks too similar to the other weapon (sword) which is used for a "Ready Player" in my opinion.

basicer:
image

basicer:
image

common/server_player.cpp
```diff
@@ -797,6 +799,24 @@ Server_Player::cmdUnconcede(const Command_Unconcede & /*cmd*/, ResponseContainer
...
+        return Response::RespFunctionNotAllowed;
+
+    Server_Player *player = this->game->getPlayers()[cmd.target_id()];
+
+    ges.setForcedByJudge(playerId);
+    if (player == nullptr)
+        return Response::RespContextError;
+
+    for (int i = 0; i < cmd.game_command_size(); ++i) {
+        player->processGameCommand(cmd.game_command(i), rc, ges);
```
ctrlaltca: Not really a change request, but I was a bit in doubt about this; maybe you already checked it, but I thought it could be useful to point this out.

When servatrice runs multithreaded, multiple connection pools can live in different threads, so command sent by 2 different players on the same game can be handled in 2 different instances of `Server_ProtocolHandler` living in different threads.
Anyway, it seems like `Server_Game::addPlayer()` already takes care of this by ensuring that every `Server_Player` object lives in the same thread of the game.

common/server_player.cpp
```diff
@@ -1491,13 +1511,16 @@ Server_Player::cmdDelCounter(const Command_DelCounter &cmd, ResponseContainer &
...
     if (!game->getGameStarted())
         return Response::RespGameNotStarted;
-    if (conceded)
-        return Response::RespContextError;
+
+    if (!judge) {
+        if (spectator)
+            return Response::RespFunctionNotAllowed;
+
+        if (conceded)
```
ctrlaltca: Is it intentional that a judge can change phase/turn of the game of a player after he conceded?
Afaik all the player zones and their contents should have been zapped already (Note: this comment applies to the following, similar changes, too).
This would also simplify the change to a simple
```
if(spectator && !judge)
..
```
If it's intentional, can you elaborate on a use case? Thank you

common/server_protocolhandler.cpp
```diff
@@ -762,6 +762,11 @@ Server_ProtocolHandler::cmdCreateGame(const Command_CreateGame &cmd, Server_Room
         if (room->getGamesCreatedByUser(QString::fromStdString(userInfo->name())) >= server->getMaxGamesPerUser())
             return Response::RespContextError;
 
+    if (cmd.join_as_judge() && !server->permitCreateGameAsJudge() &&
```
ctrlaltca: This looks fishy to me.
If `permitCreateGameAsJudge` is true, anyone can create a game as a judge even without the needed "judge" permission level.
If `permitCreateGameAsJudge` is false, only judges can create a game as a judge.

Is this how you meant it to be implemented?
Please add the configuration entry and a short description to servatrice.ini.example

servatrice/src/serversocketinterface.cpp
```diff
@@ -1415,57 +1415,88 @@ Response::ResponseCode AbstractServerSocketInterface::cmdReloadConfig(const Comm
...
+            QString::fromStdString("Failed to demote user %1: %2").arg(userName).arg(query->lastError().text()));
+        return false;
+    }
+
+    AbstractServerSocketInterface *user =
+        static_cast<AbstractServerSocketInterface *>(server->getUsers().value(userName));
+    if (user) {
+        Event_ConnectionClosed event;
+        event.set_reason(Event_ConnectionClosed::DEMOTED);
+        event.set_reason_str("Your moderator status has been revoked.");
```
ctrlaltca: Your moderator / judge status has been revoked?

common/server_protocolhandler.cpp
basicer: That's right.  Some games like Werewolf you need the host to be able to do things, so for servers that want to support this type of thing, the creator of the game can start as a judge for this that game.  It's off by default to prevent abuse on servers that are mostly for MtG.

ctrlaltca: I got your point, and it makes sense. Thank you

cockatrice/src/playerlistwidget.cpp
```diff
@@ -111,8 +112,16 @@ void PlayerListWidget::updatePlayerProperties(const ServerInfo_PlayerProperties
...
     bool isSpectator = prop.has_spectator() && prop.spectator();
-    player->setIcon(1, isSpectator ? spectatorIcon : playerIcon);
-    player->setData(1, Qt::UserRole, !isSpectator);
+    if (prop.has_judge() || prop.has_spectator()) {
+        if (prop.has_judge() && prop.judge()) {
+            player->setIcon(1, judgeIcon);
+        } else if (isSpectator) {
+            player->setIcon(1, spectatorIcon);
+        } else {
+            player->setIcon(1, playerIcon);
```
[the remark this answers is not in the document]
basicer: I think what happens here is sometimes messages come in with no spectator / judge info even though they are currently a spectator.  In this case you don't want to remove their icon.

common/server_player.cpp
basicer: In this flow the judge is executing the command directly, not though another player, so they will never be conceded.

cockatrice/src/userlist.cpp
```diff
@@ -307,7 +307,7 @@ bool UserListTWI::operator<(const QTreeWidgetItem &other) const
 
     // Sort by user level
     if (data(0, Qt::UserRole) != other.data(0, Qt::UserRole))
-        return data(0, Qt::UserRole).toInt() > other.data(0, Qt::UserRole).toInt();
+        return (data(0, Qt::UserRole).toInt() & 15) > (other.data(0, Qt::UserRole).toInt() & 15);
```
ZeldaZach: Do we want to &15 instead of & role name?

servatrice/src/serversocketinterface.cpp
```diff
@@ -1457,7 +1457,7 @@ bool AbstractServerSocketInterface::removeAdminFlagFromUser(const QString &userN
     if (user) {
         Event_ConnectionClosed event;
         event.set_reason(Event_ConnectionClosed::DEMOTED);
-        event.set_reason_str("Your moderator status has been revoked.");
+        event.set_reason_str("Your moderator / judge status has been revoked.");
```
ZeldaZach: and/or

cockatrice/src/userlist.cpp
basicer: There is no role name for 15, it's all the roles below judge or'd together.

ZeldaZach: -facepalm-

ZeldaZach:
I'm somewhat ready to roll this in and see what happens. Worst comes to worst, 2.7.1 :)

cc @ctrlaltca

ctrlaltca:
Before merging, please document the new option in servatrice.ini.example

ZeldaZach:
Is there any database migration required for this? Or does it work with the current layout of the DB?

basicer:
admin is an int field so it fires right up.

PR #3534, opened 2019-01-30
Counter expressions

basicer:
- You cant get the middle click menu for counters with a track pad
- Math is hard.

- Shift+Clicking a counter will show the middle click menu.
- When setting a counter value, you can type a math expression in the box and the counter will be set to the result.


skwerlman:
Is it possible to reference the current value of the counter?

basicer:
Yup, the current value is x.

ebbit1q:
Err? Why can't you middle mouse button on trackpad? isn't it triple tap everywhere? I can do this just fine.....

ctrlaltca:
> Err? Why can't you middle mouse button on trackpad? isn't it triple tap everywhere? I can do this just fine.....

On macs the "three finger tap" is tied to a useless "lookup" function. The shift+click addition can be quite useful there.

tooomm:
Do we want to let the user know that he can do calculations?
There is no exposure of this feature in the interface yet, right?

basicer:
I suppose we could put it in the tips and tricks dialog.

tooomm:
Haven't thought about that one :)

That would go here:
https://github.com/Cockatrice/Cockatrice/tree/master/cockatrice/resources/tips

Do you want to include that in this PR as well?

ZeldaZach:
Should be included in the PR since it's such a big change. Re: tom's comment on new tip dialog

PR #3535, opened 2019-01-30
set default minumum player count for multi column layout to 4

### LINKED ISSUE #3533: Multi Column layout default should be 4, not 5.
Currently the default for multi-column layout is 5, it should be set at 4. 

Having non-multi column layout for 4 player games is a pain because it is very space inefficient, and while you can change it, new players don't know this and so end up starting their first games on a very space inefficient boardstate with tiny icons and many of them just leave because it is not very nice.

Anyone who knows about the multicolumn layout doesn't get much benefit out of it because if they are playing with other players that don't use it they have to constantly ask them to move stuff to the left or else they just pile their lands out to the right, not knowing that thier opponents are playing with mult-column layout on. 

The fix is to change the default to 4 instead of 5. (I actually use 3 as mine, but the benefit from 3 to 4 isn't as extreme). This would benefit both new and old players.

ebbit1q:
fixes #3533
does not change existing settings

### DIFF
--- a/cockatrice/src/settingscache.cpp
+++ b/cockatrice/src/settingscache.cpp
@@ -233,7 +233,7 @@ SettingsCache::SettingsCache()
     displayCardNames = settings->value("cards/displaycardnames", true).toBool();
     horizontalHand = settings->value("hand/horizontal", true).toBool();
     invertVerticalCoordinate = settings->value("table/invert_vertical", false).toBool();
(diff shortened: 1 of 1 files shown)


ebbit1q:
People don't like change, simply put.

Insertcleverphrasehere:
Zach's suggestion sound's like it would work fine.

PR #3536, opened 2019-01-30
fix #3500

### LINKED ISSUE #3500: Fix "reveal top cards" message in the log
When using "reveal top cards", the message shown in the log is always "reveals the top card" cause logRevealCards doesn't take the number of cards revealed `cardList.size()` as a parameter

https://github.com/Cockatrice/Cockatrice/blob/273d5d89b74afd2ad5cb88e7a14a6f0edf276fdb/cockatrice/src/player.cpp#L1870

ebbit1q:
fixes #3500
Please don't reveal the top 0 cards.

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -479,8 +479,12 @@ void MessageLogWidget::logRevealCards(Player *player,
     if (cardNameContainsStartZone) {
         cardStr = cardName;
     } else if (cardName.isEmpty()) {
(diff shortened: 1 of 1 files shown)


ctrlaltca:
> NOBODY SAW THAT

That missing semicolon :D

ebbit1q:
NOOOO

tooomm:
Too bad the 3rd party never knows how much cards are being revealed between the other players.
That would be the perfect, transparent solution between all players.

ebbit1q:
That would require changes on the server side, this is a good solution without changing that.

PR #3540, opened 2019-01-31
Fix up Dockerfile to use ubuntu:bionic and actually build

### LINKED ISSUE #3539: Dockerfile does not build
<b>System Information:</b>

AWS EC2 t2.micro instance running a free Ubuntu 18.04 amd64 image

Using `docker-ce` from Docker repos

Quite simply, the Docker build of Cockatrice (required for the Docker-based `servertrice` setup) does not work. The main issue is that not all necessary files transferred into the Docker container (`.git` is missing, and is needed by cmake). However, even doing this transfer manually leads to further issues.

I have a PR pending for fixing this, and it will be attached soon.

fsufitch:
## Related Ticket(s)
- Fixes #3539

The servertrice Docker build does not work.

- `Dockerfile` is revamped to use `ubuntu:bionic`, no PPAs, and transfer all the correct files to the container.
- Building the image specified by the `Dockerfile` now works.

## Screenshots
N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,12 +17,6 @@ RUN apt-get update && apt-get install -y\
 COPY . /home/servatrice/code/
 WORKDIR /home/servatrice/code
 
-<<<<<<< HEAD
-=======
-COPY . /home/servatrice/code/
-WORKDIR /home/servatrice/code
-
->>>>>>> 9565b8fa01261d75c26dfae313b05f775d076483
 RUN mkdir oracle
 
 WORKDIR build
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,19 @@
 FROM ubuntu:bionic
 MAINTAINER Zach Halpern <zahalpern+github@gmail.com>
 
-RUN apt-get update && apt-get install -y software-properties-common
 #RUN add-apt-repository -y ppa:smspillaz/cmake-master
 RUN apt-get update && apt-get install -y\
-  build-essential g++\
+  build-essential\
   cmake\
   git\
   libprotobuf-dev\
+  libqt5sql5-mysql\
   protobuf-compiler\
   qt5-default\
   qtbase5-dev\
   qttools5-dev-tools\
   qttools5-dev\
-  libqt5sql5-mysql
+  software-properties-common
 
 COPY . /home/servatrice/code/
 WORKDIR /home/servatrice/code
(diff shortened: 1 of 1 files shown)


ZeldaZach:
@ebbit1q Can you test this please?

fsufitch:
Tested and runs on an AWS Ubuntu Bionic amd64 image, and on a Fedora 29 Workstation laptop.

Also, updated the Docker instructions on the wiki to follow a process that actually works: https://github.com/Cockatrice/Cockatrice/wiki/Setting-up-Servatrice#using-docker

ebbit1q:
Looks fine, will test later.

Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM ubuntu:trusty
+FROM ubuntu:bionic
 MAINTAINER Zach Halpern <zahalpern+github@gmail.com>
 
 RUN apt-get update && apt-get install -y software-properties-common
-RUN apt-add-repository ppa:ubuntu-sdk-team/ppa
-RUN add-apt-repository -y ppa:smspillaz/cmake-master
+#RUN add-apt-repository -y ppa:smspillaz/cmake-master
```
tooomm: I think we can remove that too as 18.04 LTS finally includs a more recent cmake (3.10), see https://packages.ubuntu.com/bionic/cmake

The old one only had 2.8 and we require [3.1](https://github.com/Cockatrice/Cockatrice/blob/master/CMakeLists.txt#L9)

fsufitch: Removed that line.

Dockerfile
```diff
@@ -1,30 +1,29 @@
...
   qttools5-dev\
-  libqt5sql5-mysql
+  software-properties-common
 
-ENV dir /home/servatrice/code
-WORKDIR $dir
+COPY . /home/servatrice/code/
+WORKDIR /home/servatrice/code
+
+<<<<<<< HEAD
```
ZeldaZach: Nice try

ebbit1q: WTF

Dockerfile
```diff
@@ -1,30 +1,23 @@
...
+FROM ubuntu:bionic
 MAINTAINER Zach Halpern <zahalpern+github@gmail.com>
 
-RUN apt-get update && apt-get install -y software-properties-common
-RUN apt-add-repository ppa:ubuntu-sdk-team/ppa
-RUN add-apt-repository -y ppa:smspillaz/cmake-master
 RUN apt-get update && apt-get install -y\
-  build-essential g++\
+  build-essential\
   cmake\
```
tooomm: Do we still need this?
I think we use the bundled cmake now?

ebbit1q: We still need to actually install cmake to use cmake.

Issue #3537, opened 2019-01-30
Deprecate TCP Protocol, Switch Client to WebSockets


basicer:
Right now the server supports two protocols for incoming connections WebSockets and regular TCP connections.  Switching the default to be Websockets has many advantages:

- You could host the server for free on platforms like Heroku
- You could put cloudflare or another WAF in front of the server if you are having DDOS/Spam issues.
- We get encryption for more or less free via HTTPS, which is great as we currently send passwords in more or less plain text.
- Users in locked down networks that force all traffic though a transparent proxy have a better chance of working.
- Users that are required to use non SOCKS proxy servers will be better supported

ZeldaZach:
You bring up good points. I have Cloudflare enabled on the server already, but it's a pass-through for server connections. I'd approve of a change like this :)

ctrlaltca:
Now that both the server and the client support websockets, we have to plan how to move client users to websockets in order to deprecate and remove the tcp protocol in the future. Here's a draft todolist:

- [ ] ensure current public servers support websocket connections
- [ ] inform server operators that tcp support is going away soon (decide a date / next version, but let them know beforehand)
- [ ] make the client default to websocket connections for new installs
- [ ] make the server dialog updater default to websocket connections
- [ ] release a version with websocket support enabled and used by default
- [ ] inform users that the new release is out and that tcp support is going away
- [ ] check back with server operators if the translation is going smooth
- [ ] remove tcp support from client and server
- [ ] inform server operators that tcp support has been removed
- [ ] publish a new release without tcp support
- [ ] inform users that the new release is out and that tcp support has gone away.

ctrlaltca:
I'm tagging cockatrice public server admins to inform them about this proposal.
@ZeldaZach (Rooster Ranges)
@skwerlman (Tetrarch)
@Lachee (Chickatrice)
@dev-id (dr4ft)
Dear server admins, please have a look at the proposal of using websockets by default on cockatrice; we are eager to hear about your opinions, concerns and suggestions.

ctrlaltca:
So, first of all, i checked the current public servers.
Roosted Ranges and Chickatrice already supports websockets.
Tetrarch doesn't support websockets at the moment.
dr4ft is down.

skwerlman:
The proposal looks good to me. I'll enable websocket support later this week when I update my server

ZeldaZach:
I'm voting to remove dr4ft from the public servers list. They've been inactive for too long

[...]

PR #3545, opened 2019-01-31
Client Websockets

basicer:
## Related Ticket(s)
- Works toward #3537

- Client can't connect over websockets

- When selecting a websocket port (80, 443, 8080, 4748) . connect using webosckets instead of raw TCP.
- Servatrice can read the client IP out of an HTTP header to support reverse proxies in front of it.

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -18,12 +18,13 @@
 #include <QList>
 #include <QThread>
 #include <QTimer>
+#include <QWebSocket>
 
 static const unsigned int protocolVersion = 14;
 
 RemoteClient::RemoteClient(QObject *parent)
     : AbstractClient(parent), timeRunning(0), lastDataReceived(0), messageInProgress(false), handshakeStarted(false),
-      messageLength(0)
+      usingWebSocket(false), messageLength(0)
 {
 
     clearNewClientFeatures();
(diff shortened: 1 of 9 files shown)


ZeldaZach:
screenshot 2019-02-01 15 18 23

Could be a problem

cockatrice/src/remoteclient.cpp
```diff
@@ -301,21 +319,51 @@ void RemoteClient::readData()
...
+        buf.data()[1] = (unsigned char)(size >> 16);
+        buf.data()[0] = (unsigned char)(size >> 24);
+
+        socket->write(buf);
+    }
+}
+
+void RemoteClient::connectToHost(const QString &hostname, unsigned int port)
+{
+    usingWebSocket = port == 443 || port == 80 || port == 4748 || port == 8080;
```
ctrlaltca: maybe it's easier to think that anything but 4747 is using websockets?
`usingWebSocket = port != 4747;`

servatrice/src/serversocketinterface.cpp
```diff
@@ -1037,14 +1037,15 @@ Response::ResponseCode AbstractServerSocketInterface::cmdRegisterAccount(const C
         return Response::RespUserAlreadyExists;
     }
 
-    if (servatrice->getMaxAccountsPerEmail() &&
-        !(sqlInterface->checkNumberOfUserAccounts(emailAddress) < servatrice->getMaxAccountsPerEmail())) {
-        if (servatrice->getEnableRegistrationAudit())
-            sqlInterface->addAuditRecord(QString::fromStdString(cmd.user_name()).simplified(), this->getAddress(),
-                                         QString::fromStdString(cmd.clientid()).simplified(), "REGISTER_ACCOUNT",
-                                         "Too many usernames registered with this email address", false);
+    if ((sqlInterface->checkNumberOfUserAccounts(emailAddress) >= servatrice->getMaxAccountsPerEmail())) {
```
ctrlaltca: Is this change somewhat related/needed? It forces servatrice to make a query to lookup if an email address is already used even if the related settings is disabled.

cockatrice/src/remoteclient.h
```diff
@@ -85,28 +87,35 @@ private slots:
...
 public:
-    RemoteClient(QObject *parent = 0);
-    ~RemoteClient();
+    explicit RemoteClient(QObject *parent = nullptr);
+    ~RemoteClient() override;
     QString peerName() const
     {
-        return socket->peerName();
+        if (usingWebSocket) {
+            return socket->peerName();
```
ctrlaltca: bug: inverted cases. If `usingWebSocket` you should return `websocket->peerName()`

cockatrice/src/remoteclient.cpp
basicer: Was thinking the same thing.

cockatrice/src/remoteclient.h
basicer: Good catch!

servatrice/src/serversocketinterface.cpp
ZeldaZach: Fixed

PR #3546, opened 2019-02-01
Make server 12-factor compatible. 

basicer:
- Hosting systems like heroku and dokku don't do well with config files.  Allow important settings to be changed via environmental variables. 

- Some important variables can be set vie env variable.
- Add heroku compatible docker file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,12 @@ RUN apt-get update && apt-get install -y\
   qtbase5-dev\
   qttools5-dev-tools\
   qttools5-dev\
-  software-properties-common
+  software-properties-common\
+  libqt5sql5-mysql\
+  libqt5websockets5-dev
 
 COPY . /home/servatrice/code/
 WORKDIR /home/servatrice/code
-
-RUN mkdir oracle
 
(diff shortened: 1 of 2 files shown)


ctrlaltca:
The change looks fine; if we merge it, we should document the usable env variables or at least add an example of how to use them in the wiki.

ebbit1q:
There is a very clear advantage of sorting your package list, makes sure you don't have doubles :)

ZeldaZach:
Nice catch @ebbit1q

ZeldaZach:
@basicer When you get to documenting this on the wiki, ping me and i'll get this merged :)

basicer:
### Hosting servatrice on PaaS platforms like Heroku

Servatrice will respect the `PORT` and `DATABASE_URL` environmental variables, using them instead of a config file if present.  When the `DATABASE_URL` variable is present, servatrice also assumes the `cockatrice` table prefix as well as `sql` mode for authentication and room creation.   `DATABASE_URL` should be a full mysql connection string: ex. `mysql://user:pass@host.com/databse`

PR #3548, opened 2019-02-02
Add shortcut for next phase with action

basicer:
You have to click a lot on the phase tool bar while playing.
A shortcut that moved you to the next phase and preformed that phases default action will streamline things.

- Adds a new action (default shit+tab) to move to the next phase and preform the phase action.
- Preforming the action in the last phase automatically passes the turn before moving phases.


ebbit1q:
I'd prefer to have it do the default action (drawing a card) on pressing shift tab the second time, not immediately after switching. Overall I think this is a great addition. All good :+1:

PR #3549, opened 2019-02-02
Move to bottom of library in random order.

### LINKED ISSUE #673: Cascade can't be performed.

basicer:
## Related Ticket(s)
- Fixes #673 

- No way to cascade

- Allow shuffling a subset of a zone
- When moving cards to the bottom of library via right click, shuffle them after
- If you want to keep the order, drag them or right click one at a time.
- Process events in the correct order server-side (they were processed backward before).

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -742,6 +742,8 @@ void MessageLogWidget::logShuffle(Player *player, CardZone *zone, int start, int
         return;
     }
 
+    // start and end are indexes into the portion of the deck that was shuffled
+    // with negitive numbers counging from the bottom up.
     if (start == 0 && end == -1) {
         appendHtmlServerMessage(tr("%1 shuffles %2.")
                                     .arg(sanitizeHtml(player->getName()))
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -732,13 +732,32 @@ void MessageLogWidget::logSetTapped(Player *player, CardItem *card, bool tapped)
     }
 }
 
-void MessageLogWidget::logShuffle(Player *player, CardZone *zone)
+void MessageLogWidget::logShuffle(Player *player, CardZone *zone, int start, int end)
 {
     soundEngine->playSound("shuffle");
-    if (currentContext != MessageContext_Mulligan)
+    if (currentContext == MessageContext_Mulligan)
+        return;
+    if (start == 0 && end == -1) {
         appendHtmlServerMessage(tr("%1 shuffles %2.")
                                     .arg(sanitizeHtml(player->getName()))
                                     .arg(zone->getTranslatedName(true, CaseShuffleZone)));
+    } else if (start < 0 && end == -1) {
+        appendHtmlServerMessage(tr("%1 shuffles the bottom %3 cards of %2.")
+                                    .arg(sanitizeHtml(player->getName()))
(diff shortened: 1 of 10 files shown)


cockatrice/src/messagelogwidget.cpp
```diff
@@ -732,13 +732,32 @@ void MessageLogWidget::logSetTapped(Player *player, CardItem *card, bool tapped)
...
 }
 
-void MessageLogWidget::logShuffle(Player *player, CardZone *zone)
+void MessageLogWidget::logShuffle(Player *player, CardZone *zone, int start, int end)
 {
     soundEngine->playSound("shuffle");
-    if (currentContext != MessageContext_Mulligan)
+    if (currentContext == MessageContext_Mulligan)
+        return;
+    if (start == 0 && end == -1) {
```
ZeldaZach: Comments? I have no clue what these could mean.

common/server_cardzone.cpp
```diff
@@ -40,13 +40,25 @@ Server_CardZone::~Server_CardZone()
...
-void Server_CardZone::shuffle()
+void Server_CardZone::shuffle(int start, int end)
 {
     // Size 0 or 1 decks are sorted
     if (cards.size() < 2)
         return;
-    for (int i = cards.size() - 1; i > 0; i--) {
-        int j = rng->rand(0, i);
+
+    if (end < 0)
```
ZeldaZach: Comments?

common/server_player.cpp
```diff
@@ -854,18 +854,26 @@ Server_Player::cmdShuffle(const Command_Shuffle & /*cmd*/, ResponseContainer & /
     if (conceded)
         return Response::RespContextError;
 
-    Server_CardZone *deckZone = zones.value("deck");
-    deckZone->shuffle();
+    if (cmd.zone_name() != "deck")
```
ZeldaZach: I wish clang-format would add brackets :/ I'm a fan of brackets everywhere

basicer:
As it turns our the order isn't what is shown in the logs.  The shuffle happens after the cards are moved, but it shows before they are move in the log.  Why?  Not sure, but I think because we emit a bunch of stuff as we process the events, and then later the events happen in some other order, so no matter what the shuffle goes in the log first.

cockatrice/src/messagelogwidget.cpp
ctrlaltca: As far as i understand, a positive number is an offset from the top of the deck, a negative number is an offset from the bottom of the deck.
0 and -1 are special cases (since randomize the position of a single card make no sense), so they have the special meaning of first and last card in the deck.
Maybe we should add 2 constants/defines/enum/whatever:
```
LAST_CARD_IN_DECK = -1;
FIRST_CARD_IN_DECK=0;
```

ZeldaZach:
I'm a fan of merging this in, as it solves its purpose, with the note that we will look at #3553 at some point soon. Great job :)

PR #3556, opened 2019-02-05
Connect from commandline

basicer:
- When testing network options, it is a pain to get multiple test clients connected

- Add --connect command line option to connect the client
- Normalize argument parsing

## Screenshots
```
./cockatrice --help
Cockatrice

Options:
  -h, --help                           Displays this help.
  -v, --version                        Displays version information.
  -c, --connect <user:pass@host:port>  Connect on startup
  -d, --debug-output                   Debug to file
```

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -30,6 +30,7 @@
 #include "soundengine.h"
 #include "spoilerbackgroundupdater.h"
 #include "thememanager.h"
+#include "version_string.h"
 #include "window_main.h"
 #include <QApplication>
 #include <QCryptographicHash>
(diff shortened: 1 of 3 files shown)


basicer:
http://doc.qt.io/qt-5/qcoreapplication.html#applicationVersion-prop

tooomm:
@ebbit1q do you have the same issue with the version command at servatrice?
https://github.com/Cockatrice/Cockatrice/wiki/Servatrice-Command-Line-Reference

Also, are the new commands documented already?

ebbit1q:
No, servatrice always gave the correct version string.

cockatrice/src/main.cpp
```diff
@@ -114,6 +113,21 @@ int main(int argc, char *argv[])
...
+    QCommandLineParser parser;
+    parser.setApplicationDescription("Cockatrice");
+    parser.addHelpOption();
+    parser.addVersionOption();
+
+    parser.addOptions(
+        {{{"c", "connect"}, QCoreApplication::translate("main", "Connect on startup"), "user:pass@host:port"},
+         {{"d", "debug-output"}, QCoreApplication::translate("main", "Debug to file")}});
+
+    parser.process(app);
```
ZeldaZach: Should we error check to ensure this works?
```cpp
if (!parser.parse(QCoreApplication::arguments())) {
        *errorMessage = parser.errorText();
        return CommandLineError;
    }
```

basicer: .process() takes care of this, exiting and printing an error message if there is a problem.

PR #3560, opened 2019-02-05
keep wrong project domain for proper settings storage

tooomm:
Old brukie domain was left in code

- Add explaining hint why we have to keep a wrong value

Looks like it is used for storing and loading settings:
https://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -98,8 +98,10 @@ int main(int argc, char *argv[])
     app.addLibraryPath(app.applicationDirPath() + "/plugins");
 #endif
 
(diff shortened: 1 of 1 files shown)


ctrlaltca:
That values is only used by the settings loader/saver, this PR breaks loading of old config values.

ZeldaZach:
While cockatrice.de is wrong, we shouldn't change this value as it's used across the file systems for current installs and we would break everyone's settings. I vote to close.

tooomm:
We can add an explaining comment as of why it has to be wrong at least?

PR #3562, opened 2019-02-06
Add websocket feature

ctrlaltca:
## Related Ticket(s)
- Related #3537

A new "websocket" feature will be exposed by the server, the client and the webclient.
A server could make the feature mandatory and inform users that they need to update their trice to be able to connect in the future.

### DIFF
--- a/webclient/webclient.js
+++ b/webclient/webclient.js
@@ -86,7 +86,7 @@ var WebClient = {
 
   resetConnectionvars : function () {
     this.cmdId = 0;
-    this.pendingCommands = {};    
+    this.pendingCommands = {};
   },
 
   sendCommand : function (cmd, callback)
@@ -151,7 +151,7 @@ var WebClient = {
       WebClient.sendSessionCommand(sc, function() {
         WebClient.lastPingPending = false;
       });
-        
+
     }, WebClient.options.keepalive);
   },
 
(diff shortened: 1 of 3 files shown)


ZeldaZach:
I ask you also add a flag, `2.7.0_min_version` to the config as that will be useful in the near future and I'd rather do one PR modifying these files

ctrlaltca:
@ZeldaZach `2.7.0_min_version` added.

webclient/webclient.js
```diff
@@ -168,8 +168,9 @@ var WebClient = {
         "feature_set",
         "room_chat_history",
         "client_warnings",
-        "2.6.1_min_version",
         "websocket",
+        "2.6.1_min_version",
+        "2.7.0_min_version",
         /* unimplemented features */
```
tooomm: Unimplemented features? hmm?
We have forgot pw, and idle client timer?

ctrlaltca: I don't get the question, what do you mean?

tooomm: It looks like the list is somehow separated by a comment.
But the divider makes no sense to me:
```
      "clientfeatures" : [
        "client_id",
        "client_ver",
        "feature_set",
        "room_chat_history",
        "client_warnings",
        /* unimplemented features */
        "forgot_password",
        "idle_client",
        "mod_log_lookup",
        "user_ban_history"
]
```

See https://github.com/Cockatrice/Cockatrice/blob/master/webclient/webclient.js#L165-L176

tooomm: Scrape that, it's the webclient.... 🙈

ctrlaltca: The webclient only implements the features before the comment.
The others are still included only to be able to connect to servers requiring/declaring them.

PR #3570, opened 2019-02-08
Click replay timeline to seek.

basicer:
You can't seek replays, so if you want to rewind, you have to start over and fast forward.

Clicking on the histogram seeks the replay to that position.

### DIFF
--- a/cockatrice/src/replay_timeline_widget.cpp
+++ b/cockatrice/src/replay_timeline_widget.cpp
@@ -64,7 +64,7 @@ void ReplayTimelineWidget::paintEvent(QPaintEvent * /* event */)
 
 void ReplayTimelineWidget::mousePressEvent(QMouseEvent *event)
 {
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Seems to work full bar length for "shorter" games but for long ones it cuts off

ctrlaltca:
Work fine! I see you are basically restarting the replay from the start to the point where the user clicked.
In the message log the new events are added to the ones already present, maybe it would be better to clean the message log every time the user clicks on the histogram and the whole history gets replayed?

basicer:
> Works fine! I see you are basically restarting the replay from the beginning to the point where the user clicked.
> In the message log the new events are added to the ones already present, maybe it would be better to clean the message log every time the user clicks on the histogram and the whole history gets replayed?

It only starts over if you seek backwards, but I agree the log should be cleared in that case.

PR #3567, opened 2019-02-07
Make the xml parser less strict about xml problems

ctrlaltca:
## Related Ticket(s)
- Fixes #3236

## Short roundup of the initial problem
If an xml card file contains not properly escaped html inside an element text, eg:
```
<card>
  <text>card text with a <b>html bold tag</i> in it</text>
</card>
```
it will currently stop parsing the whole file.

## What will change with this Pull Request?
The xml parser will just ignore the html tag and load the text (see [QXmlStreamReader::IncludeChildElements](https://doc.qt.io/qt-5/qxmlstreamreader.html#ReadElementTextBehaviour-enum) )

PR #3571, opened 2019-02-08
Fix unbalanced save() restore() in cards painter.

basicer:
There are thousands of spam lines in console saying `QPainter::end: Painter ended with 2 saved states`

No more of that.


tooomm:
Hmm, the debug log dialog windows seems to be negatively effected...
Does it still work for you @basicer ?

ctrlaltca:
@tooomm i'm afraid the mess in the debug log is caused by #3567

tooomm:
I guess it is ok to fix it up here?

Or do you want to open a separate PR and update this pr from master? @ctrlaltca

ctrlaltca:
> Or do you want to open a separate PR and update this pr from master? @ctrlaltca

Done, #3572

PR #3573, opened 2019-02-09
new translations

ZeldaZach:


tooomm:
You did not build new strings... you only pulled translations from transifex which would be the second step.

We need the distribute the newly added strings to our translators first 🙈

PR #3577, opened 2019-02-10
Git clone depth 15

tooomm:
- lower clone depth to 15 for travis and appveyor (from 50)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,7 +18,7 @@ skip_commits:
     - LICENSE
 
 skip_branch_with_pr: true
-
+ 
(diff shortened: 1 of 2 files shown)


ZeldaZach:
The time saved isn't a lot, but there's no reason to clone so far down to begin with.

tooomm:
Do you think we should lower the number even more?
I wasn't sure if we do some git commands that might need it

ZeldaZach:
I don't believe we use git history at all in our builds, so theoretically 5 would be fine. But at the same time, this saves such a minuscule amount of time that it's not fully necessary

PR #3579, opened 2019-02-11
Tweaks to message log.

### LINKED ISSUE #3553: Rewrite message log to not use events to guarantee order

>Rob Blanckaert @basicer

>
>Zach H @ZeldaZach 20:51
That’s really dumb

basicer:
## Related Ticket(s)
- Partial addresses #3553

- Milled cards dont happen in the correct order
- Roll up shuffles into bottom message
- Remove unused variable

- Card milling will be in the IRL order.


basicer:
image

ebbit1q:
Don't we prefer squashed merges?

ZeldaZach:
Usually. I should update that setting

PR #3581, opened 2019-02-17
Updater for websockets

basicer:
## Related Ticket(s)
- #3537

- We need to force update to get users to switch to the websocket port

- The server updater won't erase passwords
- The server updater is triggered on client update.

### DIFF
--- a/cockatrice/src/settings/serverssettings.cpp
+++ b/cockatrice/src/settings/serverssettings.cpp
@@ -227,23 +227,41 @@ void ServersSettings::removeServer(QString servAddr)
 /**
  * Will only update fields with new values, ignores empty values
  */
-bool ServersSettings::updateExistingServerWithoutLoss(QString saveName,
-                                                      QString serv,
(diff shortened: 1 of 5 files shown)


cockatrice/src/settings/serverssettings.cpp
```diff
@@ -227,23 +227,41 @@ void ServersSettings::removeServer(QString servAddr)
...
+                                           QString username,
+                                           QString password,
+                                           bool savePassword,
+                                           QString site)
+{
+    int size = getValue("totalServers", "server", "server_details").toInt() + 1;
+
+    for (int i = 0; i < size; i++) {
+        if (serv == getValue(QString("server%1").arg(i), "server", "server_details").toString()) {
+            setValue(port, QString("port%1").arg(i), "server", "server_details");
```
ebbit1q: I understand why this is copied but being inconsistent on one function having the port be optional while the other doesn't just irks me.

PR #3582, opened 2019-02-18
Filter Strings for Deck Editor search

### LINKED ISSUE #215: New Feature: Sort and filter by Format
In the new Deck Editor, it allows you to have filters. Instead of having to type to say, lets go with standard, "RTR, GTC, DGM, THS, BTG, JOU, M14, M15," you should be able to click on the 2nd dropdown option "legality" and you can either then type "Modern, Standard, Legacy, Vintage, etc." to filter the cards accordingly.

basicer:
## Related Issues
- Fixes #215

Searching cards in the deck editor is hard.

- Entering a string in the search box with [:=<>] in it will try to parse the query in the MagicCards.info syntax.

## Screenshots
image


tooomm:
> MagicCards.info syntax

You mean scryfall syntax since they took over mci? https://scryfall.com/docs/syntax

If it's the old mci syntax and that's any different, I wouldn't use that since the site no longer exists and people can't check the syntax or new users are not used to it at all.

<br>

In general a very nice new feature though! 👍

ZeldaZach:
My question is: Why doesn't this auto-create the filters in the filter section once they're typed up? Would make a bit more sense as why would we have two types of filtering operations?

basicer:
I tried generating the filter tree in v0, but it turned out the filter tree wasn't expressive enough to capture the kinds of queries you can write.

basicer:
> > MagicCards.info syntax
> 
> You mean scryfall syntax since they took over mci? https://scryfall.com/docs/syntax
> 
> If it's the old mci syntax and that's any different, I wouldn't use that since the site no longer exists and people can't check the syntax or new users are not used to it at all.
> 
> 
> In general a very nice new feature though! 👍

What I really mean is http://magidex.com syntax : )

ebbit1q:
Scryfall syntax is almost exactly the same: https://scryfall.com/docs/syntax
Exception is the + and operation on sets (which is not needed)
Also Scryfall is more expansive, with more details going as far as what character is represented on the art of the card.

basicer:
The joke is I made magidex.com before scryfall ;)

cockatrice/src/filter_string.cpp
```diff
@@ -0,0 +1,217 @@
+#include "filter_string.h"
+#include "../../common/lib/peglib.h"
+
+#include <QByteArray>
+#include <QString>
+#include <cmath>
+#include <functional>
+
+peg::parser search(R"(
```
ZeldaZach: This long block of parser code just worries me... Does this change the editor's efficiency?

cockatrice/src/filter_string.cpp
```diff
@@ -0,0 +1,217 @@
...
+                return QString::fromStdString(sv.token(0));
+            }
+        };
+
+        search["NumericExpression"] = [](const peg::SemanticValues &sv) -> NumberMatcher {
+            auto arg = sv[1].get<int>();
+            auto op = sv[0].get<QString>();
+            return [=](int s) {
+                if (op == ">")
+                    return s > arg;
```
ZeldaZach: I remember in the past for a school problem I was able to convert the op string to a real op without switches... maybe I'm mistaken?

cockatrice/src/filter_string.cpp
```diff
@@ -0,0 +1,217 @@
...
+        };
+
+        search["GenericQuery"] = [](const peg::SemanticValues &sv) -> Filter {
+            StringMatcher matcher = sv[0].get<StringMatcher>();
+            return [=](CardData x) { return matcher(x->getName()); };
+        };
+
+        search["Color"] = [](const peg::SemanticValues &sv) -> char {
+            switch (sv.choice()) {
+                case 0:
```
ZeldaZach: Should we make a small `const map {W,U,B,R,G, }` and just return the index instead of the case switches?

cockatrice/src/filter_string.cpp
basicer: Nah it gets compiled to an intermediate form on app start.

cockatrice/src/filter_string.cpp
basicer: Well you know what an array of chars is ;)

cockatrice/src/filter_string.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+        search["Color"] = [](const peg::SemanticValues &sv) -> char { return "WUBRGU"[sv.choice()]; };
+        search["ColorEx"] = [](const peg::SemanticValues &sv) -> char {
+            return sv.choice() == 0 ? sv[0].get<char>() : *sv.c_str();
+        };
+    });
+
+    _error = QString();
+
+    if (ba.size() == 0) {
```
ZeldaZach:
```suggestion
    if (ba.isEmpty()) {
```

cockatrice/src/filter_string.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+NumericExpression <- NumericOperator ws? NumericValue
+NumericOperator <- [=:] / <[><!][=]?>
+NumericValue <- [0-9]+ 
+
+    )");
+
+std::once_flag init;
+
+FilterString::FilterString(const QString &expr)
```
ZeldaZach: Would _really_ like some comments in here

basicer: Should I add a tip of the day?

basicer:
> Seems fine to me. More of an Easter egg than anything, but has no detriment to the program so I'm fine having it.

Should I add a tip of the day?

ZeldaZach:
Good idea, forgot about that 😅

ebbit1q:
hold on, do we now have two separate filter methods? we already have a entire separate filter widget and all.

ctrlaltca:
I agree, having two different ways to do the same thing (filtering cards in the deck editor), both of them not exactly clear to use and not well documented is not really good from an user perspective.
To be honest, this implementation looks far better than the existing one; it would be nice to merge them into a single one or even drop the older, as long as we implement a visual way to discover the possible syntax that can be user in the new filter.

basicer:
Could add something like this to the search text widget that gives syntax help.

image

ebbit1q:
I don't like the old filter system either. this would solve multiple issues with it. Documenting it well is a must.

ctrlaltca:
That would be great. Also, a history of the last used search queries, so that a recent search pattern can be reused easily. That would totally surpass the old filters capabilities and make it obsolete.

tests/carddatabase/filter_string_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+  // void TearDown() override {}
+
+  CardData cat;
+};
+
+    Query(Empty, cat, "", true)
+    Query(Typing, cat, "t", true)
+
+    Query(NonMatchingType, cat, "t:kithkin", false)
+    Query(MathcingType, cat, "t:creature", true)
```
ebbit1q: I assume this should be MatchingType?

ebbit1q: Can you explain why it should be `MathcingType` instead?

tooomm:
With the recent commits, this should fix https://github.com/Cockatrice/Cockatrice/issues/215?

basicer:
> With the recent commits, this should address #215?

A blast from the past.

tooomm:
@basicer if this fixes #215 please make this pr auto-close that issue on merge.
You have to use `Fixes #215` in the starting post, not only link the issue please.

basicer:
> * `f:` and `banned:` aren't implemented, it's best to remove them from the syntax help for now

They should be, update your card database?

ebbit1q:
My bad, they work fine.

cockatrice/src/tab_deck_editor.cpp
```diff
@@ -700,7 +704,11 @@ void TabDeckEditor::updateCardInfoRight(const QModelIndex &current, const QModel
 
 void TabDeckEditor::updateSearch(const QString &search)
 {
-    databaseDisplayModel->setCardName(search);
+    if (search.contains(QRegularExpression("[:<>=\"!]"))) {
```
ebbit1q: Wouldn't it be a tad more efficient to store the regex instead? afaik qt does some optimization for them.

ebbit1q:
```
--- a/cockatrice/resources/help/search.md
+++ b/cockatrice/resources/help/search.md
@@ -23,7 +23,7 @@
 <dd>[c!w](#c!w) <small>(Cards that are only white)</small></dd>
 <dd>[c!wu](#c!wu) <small>(Cards that are only white or blue, or both)</small></dd>
 <dd>[c!wum](#c!wum) <small>(Cards that are only white and blue, and multicolored)</small></dd>
-<dd>[c=wubrg](#c%3Dwubrg) <small>(Cards that are all five colors)</small></dd>
+<dd>[c=wubrg](#c=wubrg) <small>(Cards that are all five colors)</small></dd>
 <dd>[c:m](#c:m) <small>(Any multicolored card)</small></dd>
 
 <dt><u>Pow</u>er, <u>Tou</u>ghness, <u>C</u>onverted <u>M</u>ana <u>C</u>ost:</dt>
@@ -41,7 +41,7 @@
 <dd>[e:lea](#e:lea)</dd>
 <dd>[e:lea,leb](#e:lea,leb) <small>(Cards that appear in Alpha or Beta)</small></dd>
 
-<dd>[e:lea+leb](#e:al,be) <small>(Cards that appear in Alpha and Beta)</small></dd>
-<dd>[e:lea,leb -e:lea+leb](#e:al,be -e:al,be) <small>(Cards that appear in Alpha or Beta but not in both editions)</small></dd>
+<dd>[e:lea+leb](#e:lea+leb) <small>(Cards that appear in Alpha and Beta)</small></dd>
+<dd>[e:lea,leb -e:lea+leb](#e:lea,leb -e:lea+leb) <small>(Cards that appear in Alpha or Beta but not in both editions)</small></dd>
```

cockatrice/src/tab_deck_editor.cpp
ZeldaZach: Agreed, having a const QRegex is better

ebbit1q:
Can you link the syntax help window so that it closes when you close the main window?

ebbit1q:
considering this will add legality to oracle, linking https://github.com/Cockatrice/Cockatrice/issues/3406

PR #3584, opened 2019-02-21
rework pt setting

### LINKED ISSUE #3544: Issues with P/T setting for face down cards
<b>System Information:</b>
Client Version: 2.6.3-custom(cf8c920) (2019-01-31)
Build Architecture: 64-bit
Qt Version: 5.9.7
System Locale: de_DE

- Once you play a card face down (unknown card) and set a P/T value, the numbers are shown in white even though modified values should always be orange.
Another thing is, that when you peak at such an card (unknown --> known card) the number changes to orange.
If you play a card face down from your hand (known card) and adjust the numbers, the P/T values are orange too.

<br>

- When adjusting power or toughness values for the first time, the resulting value is completely weird:
    - increasing power: 0/0
    - increasing toughness: -1/1
    - increasing power AND toughness: 0/1

This only happens when no value has been set before (e.g. creatures enter the battlefield with their correct P/T when played, or a morph card has been set to 2/2 etc.)

ebbit1q:
save pt as a string serverside

set the pt of cards that enter the battlefield to empty (was -1/0)

display old pt to messagelog

add flow pt actions and keybinds that increase while decreasing

put more braces everywhere

various refactors like adding consts and for loops

remove a single superfluous semicolon

changes the way pt is displayed client side to set any facedown card pt or infoless card pt or different from original pt to orange

fixes #3544 fully

Comparison table:
```
p/t << inserted | old setpt | new setpt
========================================
3/3 << 2/2      | 2/2       | 2/2
3/3 << 2        | 2/2       | 2
3/3 << +-2      | 1/1       | 1
3/3 << -2       |           | 1
3/3 << +-5      |           | -3
3/3 << +-1/+4   | 2/7       | 2/7
3/3 << -1/+4    | -1/7      | 2/7
3/3 << +4/-1    |           | 7/2
3/3 << text     | 0/0       | text
3/3 <<          |           |
3/3 << +1/+1/+1 | 4/0       | 4/4/1
3/3/3 << +1/+1  | N/A       | 4/4
3 << +1         | N/A       | 4
3/3 << /+1/+1   | 0/0       | +1/+1

```


cockatrice/src/player.cpp
```diff
@@ -2602,6 +2615,90 @@ void Player::actSetPT()
...
+        if (!card->getPT().isEmpty()) {
+            oldPT = card->getPT();
+        }
+    }
+    bool ok;
+    dialogSemaphore = true;
+    QString pt = QInputDialog::getText(nullptr, tr("Change power/toughness"), tr("Please enter the change to PT:"),
+                                       QLineEdit::Normal, oldPT, &ok);
+    dialogSemaphore = false;
+    if (clearCardsToDelete()) {
```
ZeldaZach: can simplify with the two if statements into one

cockatrice/src/player.cpp
```diff
@@ -2641,12 +2738,22 @@ void Player::actDecPT()
...
+}
+
 void Player::actSetAnnotation()
 {
     QString oldAnnotation;
-    QListIterator<QGraphicsItem *> i(scene()->selectedItems());
-    while (i.hasNext()) {
-        auto *card = static_cast<CardItem *>(i.next());
+    auto sel = scene()->selectedItems();
+    for (const auto &item : sel) {
```
ZeldaZach: can simplify further
```suggestion
    for (const auto &item : scene()->selectedItems()) {
```

cockatrice/src/player.cpp
```diff
@@ -2602,6 +2615,90 @@ void Player::actSetPT()
...
+    if (!ok) {
+        return;
+    }
+
+    bool addPow, addTou;
+    int pow, tou;
+    {
+        int sep = pt.indexOf('/');
+        QString p1 = pt.left(sep);
+        QString p2 = pt.mid(sep + 1);
```
ctrlaltca: This lacks proper user input validation. Just enter a single "/" in the dialog and it will crash.

cockatrice/src/player.cpp
ebbit1q: true, just read over that

cockatrice/src/player.cpp
```diff
@@ -2504,18 +2516,21 @@ void Player::cardMenuAction()
...
     QList<const ::google::protobuf::Message *> commandList;
-    QListIterator<QGraphicsItem *> j(scene()->selectedItems());
-    while (j.hasNext()) {
-        auto *card = static_cast<CardItem *>(j.next());
+    for (const auto &item : scene()->selectedItems()) {
+        auto *card = static_cast<CardItem *>(item);
+        QString pt = card->getPT();
+        int sep = pt.indexOf('/');
+        int pow = pt.left(sep).toInt(); // if not found this will be set to 0
+        int tou = pt.mid(sep + 1).toInt();
```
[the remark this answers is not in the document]
ebbit1q: this is exactly what the old behavior did but underwater.

cockatrice/src/player.cpp
ebbit1q: oh wow, you're right, I could have seen that immediately, I don't see where it actually checks that in the old version though.

ebbit1q: oh, I see, if isEmpty() return, why would I remove that while half asleep?

ebbit1q: actually the crash completely unrelated and is caused when drawing, and while fixing this I'll actually fix #3544 :)

cockatrice/src/player.cpp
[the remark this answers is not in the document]
ZeldaZach: The behavior here seems right. You can currently shortcut by `4` becoming `4/4` which I don't want to lose. happy with this. Only problem comes about if someone writes a word, but that tries to throw it to a number. WIl play around with it.

ebbit1q:
So current issue is that the set, P/T hover over menu is too crowded. I propose we remove the hover over menu and instead implement a popup with buttons to replace it.

ZeldaZach:
I left some comments in gitter, but i'll quickly re-iterate:
- Having Set and Change P/T options is confusing. Would rather have a combined method w/ backwards compatibility.
- Right click menu is very large, need to find a way to shrink it (future PR)

cockatrice/src/player.cpp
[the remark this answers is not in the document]
ebbit1q: Could consider streamlining it, but I'd actually like it to be more general than just applying solely to mtg pt, maybe rename it to a "stat line".

tooomm:
Does this fix the `nullptr` string on the shortcuts settings page btw? :)

ebbit1q:
Yes it does.

ebbit1q:
Change pt is nothing other than the old set pt, including all behavior and including weirdness.
The new set pt sets the pt to any string.
Change pt can do things that set pt can't: inputting +1/-1 modifies the current pt: `5/5 << +1/-1 = 6/4` also single digits get put on both sides: `5/5 << -3 = 2/2`.
Both have their place, change pt could use some improvements though.
As I suggested earlier it might be better to merge all pt editing functions into a single widget.

tooomm:
A single widget where you decide first if you want to enter numbers only or letters and the rest adopts sounds best. 👍

ebbit1q:
I'm not sure what you mean, consider the scenario where you'd play a game that has a single stat per card and you want to set it to 3, using change pt you'd set it to 3/3 instead.

ZeldaZach:
I'm willing to sacrifice the 3->3/3 module if that will help combine them

ebbit1q:
I'd like to keep the option to set the pt to any string without anything being done to it. If we were to have both options in a single dialog it'd have to be a special customized one that adds something as simple as a checkmark. While we're at it we might as well add the other options to it too. I'll take some time to make a little dialog I guess.

ebbit1q:
image

ZeldaZach:
You forgot to `git add` the two images for tip of the day

ebbit1q:
oh...

Issue #2657, opened 2017-04-26
Generic solution for all card zones (reveal etc.)


tooomm:
See https://github.com/Cockatrice/Cockatrice/pull/2652#issuecomment-297280172 and

>I agree with @Nightfirecat, we should consider the idea of a more generic solution applying to every zone that can contain cards (cardzone).

PR #3586, opened 2019-02-21
[WIP] Add Shared Zone

### LINKED ISSUE #2562: Load a subdeck
I am requesting the ability for the host (or all players, potentially) to load a subdeck at the normal deck loading screen. This is important for some formats such as Planechase and Archenemy, where the cards cannot be part of a player deck but are NEEDED for the fomat.

Perhaps a small amount of the stack space (at the bottom, where cards rarely get to, visually) can be allocated to tracking, interacting and accessing this zone.

### LINKED ISSUE #3416: Archenemy Scheme Decks
Is there any way to easily use extra decks, and if not, is there a plan to implement it? An extra deck zone would be difficult to implement, but is there any possibility of being able to shuffle cards on the battlefield so that an extra deck could be kept in the sideboard and brought in? Or is there another way (potential or currently-available) that a Scheme deck or the like could be used?

Thank you in advance for the help.

basicer:
## Related Tickets
Fix #3416 
Fix #2562

- Shared decks are a huge pain.

- Each player has an new zone named shared
- Other plays can always view and edit each others shared zones
- Other players can draw from each others shared zones.
- The owner of a shared zone can shuffle it.

## Screenshots

image

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -129,6 +129,9 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, bool _judge, T
     PileZone *sb = new PileZone(this, "sb", false, false, playerArea);
     sb->setVisible(false);
 
+    PileZone *shared = new PileZone(this, "shared", false, false, playerArea);
+    shared->setVisible(false);
+
     table = new TableZone(this, this);
     connect(table, SIGNAL(sizeChanged()), this, SLOT(updateBoundingRect()));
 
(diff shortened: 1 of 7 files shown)


ebbit1q:
fails on qt 5.5 only, don't worry about it

cockatrice/src/player.cpp
```diff
@@ -300,6 +303,57 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
     rfgMenu->addAction(aViewRfg);
     rfg->setMenu(rfgMenu, aViewRfg);
 
+    {
+        sharedMenu = playerMenu->addMenu(QString());
+
+        sharedMenu->addAction(tr("Draw card from shared"), [=]() {
```
ebbit1q: To be consistent with all the other actions I'd rather see these as separate functions than lambdas.

ZeldaZach:
You've accomplished @Daenyth's dream :)

This is a very breaking change, so I'm going to require a review from all core members.

ebbit1q:
it's not really that breaking, is it? this is something that's pretty nice though.

ZeldaZach:
> fails on qt 5.5 only, don't worry about it

We'll need to put a way to avoid this failing, as I'd rather not drop Xenial support yet

cockatrice/src/player.cpp
basicer: The thing about that is, you have to capture in the action which player the action references.  In the long term I think you could use this approach to abstract out a  makeMenuForZone function as in #2657.  When you do it this way you also don't have to have a sperate retranslate step.  All in all I think switching everything to this you could cut things down 400ish lines.

That said this way of doing things doesnt exist in 5.5, so Ill switch it over to the old way when Im happy with the way it works.

cockatrice/src/player.cpp
```diff
@@ -300,6 +303,57 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, TabGame *_pare
...
+            cmd.set_start_zone("shared");
+            cmd.set_start_player_id(this->getId());
+            cmd.set_target_zone("table");
+            cmd.mutable_cards_to_move()->add_card()->set_card_id(0);
+            sendGameCommand(cmd);
+        });
+        sharedMenu->addAction(tr("Move X card from your library to shared"), [=]() {
+            bool ok;
+            int number = QInputDialog::getInt(nullptr, tr("Place card X cards from library to shared"),
+                                              tr("How many cards from the top of the deck should this card be placed:"),
```
ebbit1q: This message is a bit misleading, something like "Amount of cards to move from the top of the deck to shared library:" would be clearer.

ZeldaZach:
I'm having some issues with this. For one, the menu doesn't have a title. It seems the "shared" deck of a player is one way: Player X can put into their own shared, and anyone can take out, but nobody else can put into it. The wording on moving a card to the "shared" zone is weirdly written. It also feels weird having multiple shared zones, one per player. Is it not one shared zone for the entire match?

Can't move cards into/out of shared zones from anywhere that's not the top menu bar. Any plans to add this option around?

screenshot 2019-02-25 15 43 40
screenshot 2019-02-25 15 43 36

ZeldaZach:
More: I like how you can add face down cards to a shared, but you can't tell if the cards you added were face down or not, only your opponents can.

screenshot 2019-02-25 15 47 33

ZeldaZach:
When removing a face-down card from shared, it doesn't update on the opponent's shared slot and if they try to move it, a randomly selected card gets moved instead.

ZeldaZach:
Do you have any time to consider looking into this again?

ZeldaZach:
ping @basicer 

Will be closing for now, but please do continue! Would love this in :)

PR #3591, opened 2019-02-27
Minor cleanup & set RR default for new clients

ZeldaZach:
For new clients, have them support Rooster Ranges (the official development server) as the primary target. Also, some minor cleanup and a fix for first builds that didn't set help websites

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -206,7 +206,7 @@ void DlgConnect::rebuildComboBoxList(int failure)
     }
 
     // If first run, set to RR
-    if (settingsCache->servers().getPrevioushostName().isEmpty()) {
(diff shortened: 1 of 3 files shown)


cockatrice/src/dlg_connect.cpp
```diff
@@ -220,6 +205,11 @@ void DlgConnect::rebuildComboBoxList(int failure)
         }
     }
 
+    // If first run, set to RR
+    if (settingsCache->servers().getPrevioushostName().isEmpty()) {
+        previousHosts->setCurrentIndex(1);
```
ebbit1q: Are we able to guarantee this index exists? What if it fails to get the servers?

ZeldaZach: Good point, will add a check

ebbit1q:
Note that this will have to be changed if we'd ever add a new official server.

PR #3610, opened 2019-03-04
Adjust fontsize up on highdpi displays

### LINKED ISSUE #3597: Incorrect font size for cards
<b>System Information:</b>
  

  
When using the latest version, 2.7.0, and it is run in macOS with a Retina display, the text is much too small to read legibly. In fact, it’s half as high as before!

A work around is doubling the text size, but the bug is that the text system seems to not understand what retina graphics means. 12pt in a window in low res mode should physically be the same size as 12pt in a window in retina mode.

Can you make it so the fonts that overlay onto card images are correctly sized at 12pt?

basicer:
Fix #3597

- Scale the font correctly.

### DIFF
--- a/cockatrice/src/abstractgraphicsitem.cpp
+++ b/cockatrice/src/abstractgraphicsitem.cpp
@@ -49,8 +49,11 @@ void AbstractGraphicsItem::paintNumberEllipse(int number,
 
     painter->restore();
 }
-void resetPainterTransform(QPainter *painter)
+
+int resetPainterTransform(QPainter *painter)
 {
     painter->resetTransform();
-    painter->setTransform(painter->deviceTransform().inverted());
+    auto tx = painter->deviceTransform().inverted();
+    painter->setTransform(tx);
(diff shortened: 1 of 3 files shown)


ebbit1q:
duplicate of https://github.com/Cockatrice/Cockatrice/issues/3597

ZeldaZach:
This is a pull request sir lol

ebbit1q:
I NEED SLEEP

cockatrice/src/abstractgraphicsitem.cpp
```diff
@@ -49,8 +49,11 @@ void AbstractGraphicsItem::paintNumberEllipse(int number,
 
     painter->restore();
 }
-void resetPainterTransform(QPainter *painter)
+#include <qDebug>
+int resetPainterTransform(QPainter *painter)
 {
     painter->resetTransform();
-    painter->setTransform(painter->deviceTransform().inverted());
-}
```
ebbit1q: I see you like line endings.

PR #3641, opened 2019-03-09
Shortcuts preference pane

### LINKED ISSUE #3448: Shortcut settings menu has too many entries per tab
Currently a lot of tabs in the shortcut settings menu are very crowded, prohibiting addition of additional shortcuts.
This can be improved by simply splitting the grouped tabs into separate tabs, with less entries per tab.

See my pr that adds 4 shortcuts while there isn't any space for them: #3439 
And @tooomm 's comment about the lack of space there: https://github.com/Cockatrice/Cockatrice/pull/3439#issuecomment-439696856
Quote:
> Just tested in client. Works for me +1
> 
> Maybe we should have more taps and don't group too many things together.
> I think it's easier to use with more tabs than all that things combined in a few:
> shortcuts
> 
> We are really running low on space within each tab, which affects usability and we could come up with better grouping to make it even better for the user and use improve naming since there is more available space.

ctrlaltca:
## Related Ticket(s)
- Fixes #3448

The shortcuts pane is too crowded, on small screens it looks more like a modern art installation.

The tabs have been replaced by a single table, with one entry per row, grouped by sections (main windows, deck editor, ...).
Table rows are readonly, and each row shows the action name and the associated shortcut.
When you select a row in the table, under the table you get the selected action name and the widget used to change the associated shortcut.

## Screenshots
Schermata 2019-03-09 alle 10 34 48


ctrlaltca:
Additional comments:
 * `ui_shortcutstab.h`has been removed; it was quite hard to maintain it: in fact, it missed some shortcuts and had some duplicated lines of code;
 * items are grouped as before; this means there are 5 (colors) + 3 (r/y/g counters) + 3 (life, storm, x) groups just for counters, maybe we can rearrange them better now that we have plenty of space for descriptions. The same applies to Power, Toughness, and "Power and Toughness" and maybe others.

tooomm:
Can the general buttons (restore all, clear all, how to) be moved below the group box?
All box content is then related to the selection in the table, and the rest is separated.

ebbit1q:
Another small issue: clicking restore or clear all shortcuts doesn't update the lineedit and will keep displaying the previous text but do nothing when interacted with.

ctrlaltca:
Addressed some comments:

> Can the general buttons (restore all, clear all, how to) be moved below the group box?

Done

> I'd like to see some more space for headers and some vertical bars to better differentiate entries.

I guess you meant alternating row colors like the deck editor: done.

> clicking restore or clear all shortcuts doesn't update the lineedit and will keep displaying the previous text but do nothing when interacted with.

Fixed

> Yet on my screen the settings menu gets super stretched resulting in everything looking weird (it is not resizable)

Actually the first column is forced to 2/3 of the available space; from your image, it looks like it's using ResizeToContents instead.. I guess this is being forced by the theme Qt is using on your DE?

> This should be improved by simply having text entry focus be glued to the line edit item

Added an event filter on the table, to that every pressed key is directly sent to the lineedit.
This means you can just click on the table row and press the key now, but navigation on the table (up/down keys) is not possible anymore. Do we ant to filter these keys out?

> putting the remove buttons next to the entry list (the button is absolutely tiny and hidden right now)

I'm afraid that to avoid the "click and point" kind of nightmare we would have to increase the table row height too much. I've made the current buttons larger and with text, they should look better now.

Current screenshot:
Schermata 2019-03-10 alle 10 27 24

tooomm:
Wow, the table is way better than before. And cleaner in the code! 👍 

As seen in your screen there can be a bit of confusion now between `Counter - G` and `Counter - Green`.
Can we double group items to give a bit of context and gather them? Like
```
Counter
> on Cards
  > Green
  > Red
  > Yellow
> on Players
  > B
  > G
  > Life
  > R
  > Storm
  > U
  > W
  > X
```

<br>

>This means you can just click on the table row and press the key now, but navigation on the table (up/down keys) is not possible anymore. Do we ant to filter these keys out?

I miss using arrow key for navigation, yes. :/
Arrow left/right should collapse/uncollapse the groups just like in the deck editor, too.

Also, can you add a placeholder text in the add new shortcut field please?
For example: `Hit the key/combination of keys you want to set for this action`

ctrlaltca:
> Can we double group items to give a bit of context and gather them? Like

Yep, i thought about that solution too, but the code was quite more convoluted and the final result looked more complicated to understand. Changing the group entry name to something more descriptive is probably easier, I've merged player counters, card counters and pt-related actions in 3 different groups, please have a look.

> Arrow left/right should collapse/uncollapse the groups just like in the deck editor, too.

Fixed

> Also, can you add a placeholder text in the add new shortcut field please?

Done

tooomm:
That's pretty cool now.

Did you thought about a search/filter? But that can be another PR maybe.

PR #3646, opened 2019-03-10
Permit use of up/down keys to increment/decrement counter value; Fix #3618

### LINKED ISSUE #3618: Arrow keys for life total
<b>System Information:</b>
  
ok so like ctrl L-> life total pop up
the prev version had the option to change it by pressing the upper and down button. 
That is missing now, can we have that back?

ctrlaltca:
## Related Ticket(s)
- Fixes #3618

#3534 introduced the possibility of using formulas in the "set counter" dialog.
Previously the dialog was only accepting numerical values, so the up/down arrow keys could be used to increment/decrement the value.
Now the dialog accepts a text value, so those keys can't be used anymore.

We explicitly listen for the up/down keypress and increment/decrement the value if the text is numerical.

### DIFF
--- a/cockatrice/src/abstractcounter.cpp
+++ b/cockatrice/src/abstractcounter.cpp
@@ -163,14 +163,11 @@ void AbstractCounter::incrementCounter()
 void AbstractCounter::setCounter()
 {
     dialogSemaphore = true;
-    QInputDialog *dialog = new AbstractCounterDialog(name, QString::number(value));
-    dialog->open(this, SLOT(setCounterAccepted(QString)));
-}
-
-void AbstractCounter::setCounterAccepted(QString expression)
-{
+    AbstractCounterDialog dialog(name, QString::number(value));
+    if(!dialog.exec())
+        return;
     Expression exp(value);
-    int newValue = static_cast<int>(exp.parse(expression));
+    int newValue = static_cast<int>(exp.parse(dialog.textValue()));
 
     if (deleteAfterDialog) {
         deleteLater();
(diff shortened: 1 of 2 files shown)


ebbit1q:
unused variable warnings

ctrlaltca:
> The old dialog blocked access to the main application and would always remain in the foreground,

The current dialog is modal, so it should already work like you described.. or it doesn't?

ebbit1q:
it does not

ctrlaltca:
Thank you for noticing, i saw it modal but probably it was specific to macOS.
Can you please try again now and see if it works as intended?

ebbit1q:
still the same

ctrlaltca:
Well, more than forcing the dialog to be application modal i'm not sure what i can do.. can you please  try to play with [this flags](https://github.com/Cockatrice/Cockatrice/pull/3646/files#diff-5ea2e975f4c9ba52fac62326cc3ce68fR190) on the dialog until it works for you?

ebbit1q:
WindowModal is the same, there are no other values.

ebbit1q:
After doing some investigating, the rabbit hole goes deep, the issue is that dialogs created from qgraphicsitems don't have their parent set correctly.
Somehow other dialogs hack around this by immediately switching focus back to the dialog when you try to focus the client.
I managed to return modality properly by setting the parent to the main window (with a lot of hoops `scene()->views()[0]->window()`) it would be better if the dialog could have access to a parent qwidget in a nicer way.

ctrlaltca:
maybe setting the parent to `qApp` works, too? `include<QApplication>` is probably required

ebbit1q:
QApplication is not a QWidget, for some reason QDialogs require a QWidget as a parent.

ctrlaltca:
So, i've tried to mimic the way Qt runs the dialog, using `exec()` instead of `open()`. This should make the dialog behave the same way as before, can you please have another check run @ebbit1q ?

PR #3663, opened 2019-03-16
include a list of priorities for maincardtypes in oracle

### LINKED ISSUE #3662: Tribal
<b>System Information:</b>
  

  
This may have already been fixed alongside Artifact Lands and Artifact Creatures/Enchantment Creatures, but Tribal spells are being displayed with the main card type "tribal" :thinking:

ebbit1q:
fix #3662

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -66,6 +66,16 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
     return true;
 }
 
+QString OracleImporter::getMainCardType(const QStringList &typeList)
+{
+    for (const auto &type : mainCardTypes) {
+        if (typeList.contains(type)) {
+            return type;
+        }
+    }
+    return typeList.first();
+}
+
 CardInfoPtr OracleImporter::addCard(QString name,
                                     QString text,
                                     bool isToken,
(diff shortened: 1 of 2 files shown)


oracle/src/oracleimporter.h
```diff
@@ -89,10 +89,13 @@ class OracleImporter : public CardDatabase
 {
     Q_OBJECT
 private:
+    const QStringList mainCardTypes = {"Planeswalker", "Creature", "Land",       "Sorcery",
```
ZeldaZach: spacing is awk

oracle/src/oracleimporter.cpp
```diff
@@ -66,6 +66,16 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
     return true;
 }
 
+QString OracleImporter::getMainCardType(const QStringList &typeList)
```
ZeldaZach: What will this do for "Artifact Creature" ? Reading it, it appears to put it with the Artifacts instead of the Creatures

ctrlaltca: It loops following the order of items in `mainCardTypes`, so it will match Creature before Artifact

ZeldaZach: Ahh, i was looking at the `.first()` return value and thought it was weird. Yeah, this makes sense :)

Issue #2122, opened 2016-08-02
Non-english card image/text support


Daenyth:
This is a meta-ticket centralizing information for non-english card support
- [ ] Download non-english card databases in some way #1876
- [ ] Use the client's language setting to select what language cards to show #841 
(should be a separate setting for card art language on its own)

kascesar:
http://www.codexruler.com/base-de-datos-de-magic-assistant-actualizada-al-castellano/  (base de datos a español, hasta tarkir)

tooomm:
mtgjson doesn't have image links, but foreign names and unique multiverse ids for some sets since v3. 
Check issue number 12 in their repo. But it's still work in progress I think...
We don't utilise v3 yet, though. @ctrlaltca was working on that once. 

No foreign card text in mtgjson as far as I can remember, but I'm not certain.

EtaoinWu:
[http://gatherer.wizards.com](MTG Gatherer) uses cookies to mark the language. Cookie's name is `CardDatabaseSettings` and its value is `1=zh-CN` (if I am using Chinese language).

Daenyth:
I think each foreign printing gets its own multiverse id, so that should be
enough. I'm pretty sure the cookie setting is for the web page contents

tooomm:
Potentially useful link, thanks @ZeldaZach 
https://scryfall.com/docs/api/cards/collector

Phantato:
I found a URL by using scryfall api that would help to get non-english card image:

`https://api.scryfall.com/cards/!set!/!set:num!/zhs?format=image&face=!prop:side!`

and one can simply change `zhs` for what language they want. But the thing is that this api need a lowercase set code rather than uppercase. I don't know whether the downloader could provide the lowercase set code

Hope this would help. @Daenyth

ebbit1q:
I knew this was an option but while scouring the api docs I never found it, thanks!

tooomm:
Here is a list of all language codes at scryfall: https://scryfall.com/docs/api/languages

ebbit1q:
I got it working, exactly as intended: `https://api.scryfall.com/cards/!setcode_lower!/!set:num!/zhs?format=image&face=!prop:side!`
![image](https://user-images.githubusercontent.com/36401181/54883034-2620f880-4e61-11e9-872f-c28bbfad92d2.png)

ZeldaZach:
[...]

PR #3670, opened 2019-03-24
add !sflang! property for card image urls

ebbit1q:
update translations
add translation for the sflang property to supported languages

## Related Ticket(s)
- Related #2122 

No support for translated images.

Translated images can now be fetched from scryfall using this source: `https://api.scryfall.com/cards/!setcode_lower!/!set:num!/!sflang!?format=image&face=!prop:side!`

## Screenshots
image

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -291,6 +291,10 @@ QString PictureToLoad::transformUrl(const QString &urlTemplate) const
         }
     }
 
+    // language setting
+    transformMap["!sflang!"] = QString(QCoreApplication::translate(
+        "PictureLoader", "en", "code for scryfall's language property, not available for all languages"));
+
     for (const QString &prop : transformMap.keys()) {
         if (transformedUrl.contains(prop)) {
             if (!transformMap[prop].isEmpty()) {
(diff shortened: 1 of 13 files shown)


tooomm:
This will take the selected client language for picture downloads when `!sflang!` is used in the url?

ebbit1q:
yes, what it actually does is translate it using the normal translate function, translations for each code in scryfall is included in the pr.

tooomm:
I see the problem here as many users want to have a separate setting for that and don't force them into one direction.
Client language and card art language should be two settings

ebbit1q:
If they don't want this they can change the url, this pr adds the ability to use the url, it doesn't actually set it.

ZeldaZach:
Would like more added to this PR: When a language is set/changed, we add that URL to the download list. It should probably be the top result to change the download language, but just so people know more about this. Right now, this appears to be an easter egg, and I'd like it to be more mainstream for those who are using the program in foreign languages

ebbit1q:
I was thinking of adding more later, first do some brainstorming on how we want that implemented.

tooomm:
Please document your addition in our wiki as well @ebbit1q 
https://github.com/Cockatrice/Cockatrice/wiki/Custom-Picture-Download-URLs

An example link at the bottom with a short explanation would be very nice as well.

ebbit1q:
That entire page was out of date, fixed it.

PR #3701, opened 2019-04-30
work around confusing new default badge label

tooomm:
Shields.io changed the default badge label. Since the badge for stable and release now both show `downloads@latest` but with different numbers this can be confusing.

- always use `downloads` as label
basically that is how it looked before

## Screenshots
https://github.com/Cockatrice/Cockatrice/tree/tooomm-patch-2#download-


ebbit1q:
appveyor broke

tooomm:
That's normal. Appveyor doesn't run on readme changes.
Sadly it can't tell github and since we require appveyor to pass an admin has to merge.

PR #3709, opened 2019-05-05
Split buttons in the deck editor

### LINKED ISSUE #3695: The deck minus/plus icons could perform a bit different more intelligently
<b>System Information:</b>
  
Client Operating System: macOS Mojave (10.14)
Build Architecture: 64-bit
Qt Version: 5.12.1
System Locale: en_US
Install Mode: Standard
  
The icon seems to work to add and remove a card you just included in the deck, however, if you dont have that card in the deck I believe it should truly remove or add that card in to your deck (the one you have highlighted). Ive always noticed that regardless of card is highlighted in the search area, if you minus a card it targets the highlighted card on the right side (the deck itself)

This could also be circumvented by revamping the menu a little bit

Left Side (Search)
`(Mainboard) ->  | (sideboard) -> ... targeting highlighted searched card`

Right Side (Deck)
`+ | - ... targeting highlighted deck card`

The outcome here is that the search area would only deal with adding cards to a deck, and the deck area would only deal with adding more or removing from a deck.

ctrlaltca:
## Related Ticket(s)
- Fixes #3695

The deck editor buttons originally lived in the middle section, where tree filters are now placed.
Now that they have been moved to the left, their usage is counter-intuitive because:
 * add card to deck / add card to sideboard target the card currently selected in the card database (left) pane
 * increment / decrement / remove target the card currently selected in the deck (right) pane

The buttons targeting the deck have been moved in the deck pane.
Other minor style changes have been made to accomodate the change.

## Screenshots
Schermata 2019-05-05 alle 13 51 51
Schermata 2019-05-05 alle 13 51 55

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -91,6 +91,24 @@ void TabDeckEditor::createDeckDock()
     commentsLabel->setBuddy(commentsEdit);
     connect(commentsEdit, SIGNAL(textChanged()), this, SLOT(updateComments()));
 
+    aIncrement = new QAction(QString(), this);
+    aIncrement->setIcon(QPixmap("theme:icons/increment"));
+    connect(aIncrement, SIGNAL(triggered()), this, SLOT(actIncrement()));
+    auto *tbIncrement = new QToolButton(this);
+    tbIncrement->setDefaultAction(aIncrement);
+
+    aDecrement = new QAction(QString(), this);
+    aDecrement->setIcon(QPixmap("theme:icons/decrement"));
+    connect(aDecrement, SIGNAL(triggered()), this, SLOT(actDecrement()));
+    auto *tbDecrement = new QToolButton(this);
+    tbDecrement->setDefaultAction(aDecrement);
+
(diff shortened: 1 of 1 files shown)


ebbit1q:
I'm not sure if I like this change, what do others think?

tooomm:
I like it a lot when playing with it in the client. 👍 
Better than the old placement, logic wise.

Was there a particular reason why you changed the style @ctrlaltca ?
I think the old buttons (a bit bigger & icon only style) are better!

Here are the two in comparison:

old
buttons_old
new
buttons_new

ctrlaltca:
> Was there a particular reason why you changed the style @ctrlaltca ?
> Is that due to the removing of `deckEditToolBar` QToolBar?
Exactly.
The extra space taken by the toolbar looked quite weird having only 2 buttons on it.
Anyway, i can revert to the old style if it looks better for you.

tooomm:
>The extra space taken by the toolbar looked quite weird having only 2 buttons on it.

Really? I think it looks **much** nicer: (I also like the little space between +/- and x)
qtoolbar

qtoolbar_no

PR #3710, opened 2019-05-05
Add v4 xsd

### LINKED ISSUE #3696: New xsd for card database version 4
The current doc/cards.xsd is for card database version 3.

There should be one for card database version 4 that reflects the actual minimum required elements and attributes for any card database xml file that, if not adhered to, will cause a complete parse failure for that file.

Should it be a new file, or just overwrite? 
Should a link to https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets be included in a comment?

Related issues: 
#2545
#3458

ctrlaltca:
## Related Ticket(s)
- Fixes #3696

see #3696

A new xds for card database version 4 has been added.

Unfortunately it's not perfect; the most problematic point is that some elements in our xml are not used as xml would like to:
 * multiple set-per-card definitions are not nested inside a parent `<sets>` element;
 * the same for card relations `<related>` and `<reverse-related>`;
This forces us into using `<xs:sequence>` instead of `<xs:all>`, that unfortunately also enforces the order of elements.

About card properties, it would be nice to be able to specify a few mandatory element names but also  permit any other element (for other games or other custom user-defined properties), but unfortunately we can't add a `<xs:any>` inside `<xs:all>`.

These are mostly XSD 1.0 limitations that could be lifted by using an XSD 1.1 validator, but afaik no decent XSD 1.1 validator is easily available on every os/platform.

### DIFF
--- a/doc/carddatabase_v3/cards.xsd
+++ b/doc/carddatabase_v3/cards.xsd
@@ -70,7 +70,7 @@
             </xs:complexType>
         </xs:element>
       </xs:all>
-      <xs:attribute type="xs:integer" name="version"/>
+      <xs:attribute type="xs:integer" name="version" fixed="3"/>
     </xs:complexType>
   </xs:element>
 </xs:schema>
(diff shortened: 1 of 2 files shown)


tooomm:
Cc: @Psithief

Psithief:
I raised all of these issues with the v3 XSD. It's a pity the v4 database format didn't take those into account.

Issue #3693, opened 2019-04-19
Use new v4 database format for TK.xml file


tooomm:
https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets#to-add-your-own-custom-cards-follow-these-steps

tooomm:
fixed in #3711

PR #3711, opened 2019-05-05
Token dialog fixes

### LINKED ISSUE #3045: "Edit tokens..." dialog needs update
- The dialog shouldn't contain any tokens from `tokens.xml` file since we don't want people to edit them. (Errors should get fixed in the repo for everybody, and custom tokens belong into `TK.xml`)
Confusingly they can't even get edited there... you can actually change values of provided tokens, they get picked up in the UI and suggest that you changed them for real.
But after a restarting the client everything is gone again...
edittokens

- Instead, rename the dialog to `Edit custom tokens...` and show what it is about:
Creating and editing of custom tokens that go to the `TK.xml` file in `/customsets`

ctrlaltca:
## Related Ticket(s)
- Fixes #3045, #3693

This PR fixes a few problems found during usage of the "edit tokens" dialog

 - #3045: the "edit tokens" dialog now only permits to edit tokens from the TK.xml file
 - #3693: use carddb v4 for TK.xml file
 - (unreported): the "edit tokens" dialog logic for displaying card properties was inverted
 - (unreported): cockatrice crashes when adding a new custom token


ebbit1q:
Is the edit token dialog even still useful?

ctrlaltca:
I don't think a lot of people uses it, but at least now it's not crashing anymore :)

ebbit1q:
removing it also stops it from crashing

ZeldaZach:
I like the fix vs the removal. Why rip out things when you dont have to?

ebbit1q:
because nobody needs them

tooomm:
In my issue post I also suggested to rename that dialog to `Edit custom tokens` to stress what it's limited to.

There could also be a helpful sentence on top of the dialog like
`Create and edit custom tokens from your TK.xml file in your "customsets" folder`

ebbit1q:
In that case rework it into a "custom card maker"

tooomm:
Interesting idea, that would make a lot of sense!

An URL field for pictures would be handy then as well.
And it should be CC.xml, instead TK.xml maybe.
Also, we have to make sure people don't use it to create all custom sets with that since it's a predefined file and have some hint at least. Or make the dialog even more powerful to allow easy handling of that case.

I guess we need a new issue and have some discussion there and go with the fixes (and rename of the dialog) in this PR for now?

ctrlaltca:
An editor for custom sets (customsets/*.xml) would definitely be a nice addition, but it deserves itw own issue (maybe one already exists).
About this PR, we can either merge it and fix the current dialog or close it and create a new PR for removing the dialog. This last option would piss off any user actually using that dialog, even if it's just a few of them, until we add the proposed alternative.

tooomm:
Can you still rename the dialog and menu entry?
I'm fine with merging this PR as long as we have nothing else working! 👍

tooomm:
Hmm, we need to check the influence on the `Create token` dialog in a game.
It looks like they share the same list and tokens from tokens.xml don't show up there now - only custom created ones from TK.xml (`Edit custom tokens` dialog).
I don't think that's a big problem because we have related cards feature and everybody uses that.
But if you add tokens to a deck they don't show up there any longer too.

ctrlaltca:
Restored the old in-game tokens dialog behavior of showing all tokens.
While the "related cards" feature is handy, manually creating tokens can still be a necessity for a lot of games.

tooomm:
Both issues are fixed now. 👍 

`Edit custom token` remains unchanged and only shows TK.xml data
`Create token` now shows tokens.xml & TK.xml data, and also recognizes tokens added to the deck

Issue #3508, opened 2019-01-17
Travis: Fedora docker builds broken


tooomm:
See the logs https://travis-ci.org/Cockatrice/Cockatrice/builds/479252067

ctrlaltca:
Cross-linking ticket on fedora bugtracker: https://bugzilla.redhat.com/show_bug.cgi?id=1668865

PR #3714, opened 2019-05-06
[WIP] update travis' docker target to fedora 30

### LINKED ISSUE #3712: Fedora 30 released / update travis docker image once reasonable
`Do not forget` `Keep track` issue

ebbit1q:
fix #3712 

if it builds, it builds, if it doesn't we patiently wait.

### DIFF
--- a/.ci/Fedora30/Dockerfile
+++ b/.ci/Fedora30/Dockerfile
@@ -10,7 +10,7 @@ RUN dnf install -y \
         hicolor-icon-theme \
         libappstream-glib \
         protobuf-devel \
-        qt5-{qttools,qtsvg,qtmultimedia,qtwebsockets}-devel-5.11.1-2.fc29 \
+        qt5-{qttools,qtsvg,qtmultimedia,qtwebsockets}-devel \
         rpm-build \
         sqlite-devel \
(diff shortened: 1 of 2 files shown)


ctrlaltca:
Looks like #3508 is striking again

ebbit1q:
So, we wait.

tooomm:
As reference: https://bugzilla.redhat.com/show_bug.cgi?id=1668865 is the filled bug we run into again.
We don't have to rush for another workaround again, let's wait a bit.

tooomm:
from gitter:

>the fedora guys say that the problem in #3714 is a docker problem (that happens only in fedora?), and that since it never worked in fedora 30, they don't consider it a regression, so they won't supply a patch for it

>can you look into a workaround again?
you used a special qt version in Cockatrice/Cockatrice#3512 back then

>on fedora 29 we use the "latest working" version
on fedora 30 no known working version exists
a fedora guy on irc told me we'd better drop using docker (and.. travis?) and use some fedora-specific package build env
i'd go for drop using fedora at all :)

ZeldaZach:
🤷‍♂ Not sure what to do with this for now.

PR #3731, opened 2019-05-29
View hand window

### LINKED ISSUE #3720: Reveal hand to SELF/ View Hand
Talking with people who regularly play EDH on cockatrice, there is a feature that they highly requested. 

When drawing a lot of cards (20 or more) it can be very difficult if not impossible to be able to read the cards in your hand and sort through them. One ad hoc fix that people often use is to dump their hand into their sideboard so that they can use the search sideboard functionality to have a nicely sorted hand of cards. This is not ideal, however, because other players don't necessarily know that the other player's sideboard was empty at the beginning of the game, and also can't then see how many cards are in the player's hand. 

Two fixes have been requested: 
-One is that the "reveal hand to Player" currently only has options of each other player or everybody, YOU can easily be added to this list so that you can reveal your own hand to yourself, which would allow you a much handier way to look at the cards in your hand. However, this is not ideal as the reveal window is not interactable, so digging through the actual hand when very full still becomes an issue
-Two is that an option to 'View Hand' should be added as a shortcut and to the Game-->Player "Name"-->Hand menu. This menu is also accessed when right-clicking the hand field. This would open an interactive window identical to the sideboard and library search windows, and would allow digging through very large and cluttered hands easily.

basicer:
## Related Ticket(s)
- Fixes #3720

- Sometimes its hard to see your hand, so why not open a window?

- Adds View Hand to the player hand menu.  Opens it like any zone.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -844,6 +844,7 @@ void Player::setShortcutsActive()
 
     aViewSideboard->setShortcut(shortcuts.getSingleShortcut("Player/aViewSideboard"));
     aViewLibrary->setShortcut(shortcuts.getSingleShortcut("Player/aViewLibrary"));
+    aViewHand->setShortcut(shortcuts.getSingleShortcut("Player/aViewHand"));
     aViewTopCards->setShortcut(shortcuts.getSingleShortcut("Player/aViewTopCards"));
     aViewGraveyard->setShortcut(shortcuts.getSingleShortcut("Player/aViewGraveyard"));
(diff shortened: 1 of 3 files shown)


basicer:
Are you sure? The shortcut pane is already crazy and this is really niche
command.

ctrlaltca:
> The shortcut pane is already crazy

Did you see the new shortcut pane? https://github.com/Cockatrice/Cockatrice/pull/3641

> this is really niche command.

Still, sooner or later an issue will come up asking for that ;)

tooomm:
At least there should be an option to configure a shortcut (same as with all other view windows), without a default value pre-assigned.

PR #3740, opened 2019-06-01
Misc startup improvement

### LINKED ISSUE #3738: First startup is painful
Clean install of trice, by default it starts the program:
Schermata 2019-06-01 alle 14 57 46

As soon as trice starts, i get a popup congratulating me for upgrading (even if i just clean installed), and proposes to start oracle:
Schermata 2019-06-01 alle 14 58 10

Then i get both oracle started and a second dialog, telling me that cockatrice is unable to load the card database. What am i supposed to do first? Why? (issue #3559)
Schermata 2019-06-01 alle 14 58 29

Whatever i do, even if I just close oracle, I get yet another dialog telling me that something completed and a reload is required (issue #3470):
Schermata 2019-06-01 alle 14 58 40

Ok, done with the dialogs, i found another window with some (spoiler: not) useful information about the fact that it exists and serves a purpose:
Schermata 2019-06-01 alle 14 59 09

I'm finally done with modal popups and all, I'm surely confused and there's still an high possibility that I didn't let Oracle complete and my card database is empty.

If a card database is found, you get a dialog welcoming you again
Schermata 2019-06-01 alle 16 44 02

Then the "edit sets"
Schermata 2019-06-01 alle 16 44 13

Once you press "ok", yet another confirmation
Schermata 2019-06-01 alle 16 44 25

Additionally, Oracle consists of 6 more wizard pages and 2 additional popup dialogs.

ctrlaltca:
## Related Ticket(s)
- Fixes #3738

See #3738

The startup routine has been simplified, removing a lot of blocking steps when running a fresh installed or a just updated installation:

 * detect if this is a fresh installation or an update, and show a different welcome dialog informing the user to run oracle; eg. for new install:
first
 * don't show tips window on first run: having a both a modal welcome dialog and a the tip window coming up at the same time is just too confusing; tip window will only show up on subsequent runs;
 * Tip window: ensure it's raised on top of the main window;
 * remove the "Update completed successfully. Cockatrice will now reload the card database." popup dialog; it's just redundant and you can't avoid it;
 * Remove the "The sets database has been saved successfully." popup when saving the sets dialog: it's redundant to inform the user of something that's expected to happen when confirming a dialog;
 * Remove the "The card database has been saved successfully to ..." popups in oracle. Instead, show the default path on the wizard page. Also, invert the "choose custom path" checkbox logic to make it more user-friendly:
Schermata 2019-06-01 alle 16 42 13

I made some tests on fresh install / update / subsquent installs, but I'd really like some external testing on this. Thank you

### DIFF
--- a/oracle/src/oraclewizard.cpp
+++ b/oracle/src/oraclewizard.cpp
@@ -198,8 +198,9 @@ void IntroPage::retranslateUi()
 void OutroPage::retranslateUi()
 {
     setTitle(tr("Finished"));
-    setSubTitle(tr("The wizard has finished.") + "<br/>" +
-                tr("You can now start using Cockatrice with the newly updated cards."));
+    setSubTitle(tr("The wizard has finished.") + "<br>" +
+				tr("You can now start using Cockatrice with the newly updated cards.") + "<br><br>" +
+				tr("If the card databases don't reload automatically, restart the Cockatrice client."));
 }
 
 LoadSetsPage::LoadSetsPage(QWidget *parent) : OracleWizardPage(parent)
@@ -247,7 +248,7 @@ void LoadSetsPage::initializePage()
 void LoadSetsPage::retranslateUi()
 {
     setTitle(tr("Source selection"));
-    setSubTitle(tr("Please specify a source for the list of sets and cards. "
+    setSubTitle(tr("Please specify a compatible source for the list of sets and cards. "
                    "You can specify a URL address that will be downloaded or "
                    "use an existing file from your computer."));
 
@@ -483,7 +484,7 @@ void LoadSetsPage::zipDownloadFailed(const QString &message)
     QMessageBox::StandardButton reply;
     reply = static_cast<QMessageBox::StandardButton>(QMessageBox::question(
         this, tr("Error"),
-        message + "<br/>" + tr("Do you want to try to download a fresh copy of the uncompressed file instead?"),
+        message + "<br>" + tr("Do you want to try to download a fresh copy of the uncompressed file instead?"),
         QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes));
 
     if (reply == QMessageBox::Yes) {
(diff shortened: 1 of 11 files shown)


oracle/src/oraclewizard.cpp
```diff
@@ -890,7 +890,8 @@ void SaveSpoilersPage::retranslateUi()
     setSubTitle(tr("The spoilers file has been imported. "
                    "Press \"Save\" to save the imported spoilers to the Cockatrice card database."));
 
-    defaultPathCheckBox->setText(tr("Save to the default path (recommended)"));
+    pathLabel->setText(tr("The spoiler database will be saved at") + "<br/>" + settingsCache->getTokenDatabasePath());
```
tooomm: `getTokenDatabasePath` is wrong here

oracle/src/oraclewizard.cpp
```diff
@@ -543,7 +545,9 @@ void SaveSetsPage::retranslateUi()
     setSubTitle(tr("The following sets has been imported. "
                    "Press \"Save\" to save the imported cards to the Cockatrice database."));
 
-    defaultPathCheckBox->setText(tr("Save to the default path (recommended)"));
+    pathLabel->setText(tr("The card database will be saved at") + "<br/>" + settingsCache->getTokenDatabasePath());
```
tooomm: `getTokenDatabasePath` is wrong here

oracle/src/oraclewizard.cpp
ctrlaltca: Fixed, thank you

oracle/src/oraclewizard.cpp
ctrlaltca: Fixed, thank you

tooomm:
Thanks for the path fixes, they are displayed correctly now. 👍 

I have three further suggestions though:
- It now simply closes automatically without a hint or comment after the token step. The user can't be sure if everything went well, something crashed or whatever. Maybe he realizes a reload, sometimes there might be a new set popup. But I think we can be more clear and helpful here. Especially for very new users.<br>
There could be a very last, new conclusion page in Oracle that gives you an overview of what happened so far, which steps succeeded, where files are stored and that the database update process is finished for example. Also giving the user the hint that Cockatrice will automatically reload, and that the user should restart the client if something isn't working as expected.

- Can you reorder the phrasing with the paths a bit? I think that is more expressive:
path reorder

- The token import has two steps. Since it's only a simple file download, no parsing and processing it could all simply be on one single page: (token source selection + tokens imported)
token

ctrlaltca:
> There could be a very last, new conclusion page in Oracle

Added.



>     * Can you reorder the phrasing with the paths a bit?

Done

>     * it could all simply be on one single page

Done for both tokens and spoilers.

PR #3752, opened 2019-06-08
remove fedora package support

tooomm:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/3712

Newer Fedora can't be supported due to a issue with docker they are not going to fix, and we rely on docker to build the package in our CI.

- Remove old Fedora 29 package with releases following 2.7.1
- There will be no newer Fedora packages supported


ZeldaZach:
I assume this is fine?

PR #3759, opened 2019-06-14
Message log: handle "shuffle after moving card to deck"; fix #3553

### LINKED ISSUE #3553: Rewrite message log to not use events to guarantee order

ctrlaltca:
## Related Ticket(s)
- Fixes #3553

The messagelogwidget contains a workaround that forces card moves to be always logged last.
This causes problems with logging of command that both moves a card and make some other actions (eg #3549, move+shuffle) since events are logged in the wrong order.

This is one attempt of fixing the problem by adding yet another workaround for this specific case.
Also, 2 minor fixes are included to avoid warning spam in the log.

### DIFF
--- a/cockatrice/src/abstractcarddragitem.cpp
+++ b/cockatrice/src/abstractcarddragitem.cpp
@@ -39,7 +39,6 @@ AbstractCardDragItem::AbstractCardDragItem(AbstractCardItem *_item,
 
 AbstractCardDragItem::~AbstractCardDragItem()
 {
-    qDebug("CardDragItem destructor");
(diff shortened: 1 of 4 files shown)


ctrlaltca:
Please also have a look at #3760 before rewieving this: this PR and #3760 are two different attempts to fix the same problem.
This one is completely backwards compatible with the old behavior of the logger, but it will probably cause problems like #3553 in the future.

ctrlaltca:
closing in favor of #3760

PR #3760, opened 2019-06-14
Message log: remove all workarounds; fix #3553

### LINKED ISSUE #3553: Rewrite message log to not use events to guarantee order

ctrlaltca:
## Related Ticket(s)
- Fixes #3553

Please read #3759 first.
This is another attempt at fixing the problem. All the workarounds from messagelogwidget are removed, so that  event logging is always in the correct order as it is expected.
Also, 2 minor fixes are included to avoid warning spam in the log.


ctrlaltca:
Please also have a look at #3759 before rewieving this: this PR and #3759 are two different attempts to fix the same problem.
This one changes a bit the behavior of the logger, but it eradicates the problem to ensure problems like #3553 won't happen again in the future.

PR #3770, opened 2019-06-30
properly capitalize set types like Duel Deck and From the Vault

### LINKED ISSUE #3766: set types are not displayed perfectly
Oracle imports `duel_deck` as `Dual_deck` etc.
So we capitalize the first letter as they are all lowercase in mtgjson (see https://mtgjson.com/structures/set/).

But what we should really do is to convert to `Dual Deck`, with a space as separator and capitalize all words.

Untitled

ebbit1q:
## Related Ticket
- Fixes #3766
## Screenshot

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,6 +96,7 @@ matrix:
         - protobuf
         - qt
         - xz
+        update: true
     script: bash ./.ci/travis-compile.sh --server --install --debug
 
   - name: macOS (Release)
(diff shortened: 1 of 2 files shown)


tooomm:
That's cool! Thanks for taking care.

Missing words in the list that don't get capitalized as well: `of, in, and, with, but, or`

ebbit1q:
I only included words that I thought might be included in the future, currently it's only needed for From the Vault

ebbit1q:
there, I included them

tooomm:
If it doesn't hurt to have them. 👍
But you are right that the possible list of set_types shouldn't change to much and often...

Issue #3059, opened 2018-01-31
AppVeyor doesn't use a compiler cache


tooomm:
Similar options to ccache on Linux/macOS could be these for Windows:
https://github.com/frerich/clcache
https://github.com/inorton/cclash

As seen with the recent macOS build change (#3052), enabling this kind of cache has a very huge impact on the time needed to complete a CI run!

Build time for the windows builds on appveyor are slow ~15min now each.

Daenyth:
https://www.appveyor.com/docs/build-cache/ might be helpful

tooomm:
Note:
@Daenyth enabled the setting `Save build cache for Pull Requests` for the `Cockatrice-PR` project at appveyor in between.

Fenrirthviti:
Seems like most of the recent CI work was being focused on Azure/GHA. As someone with extensive experience fighting AppVeyor's issues and knowing how nice Azure/GHA can be comparatively, strongly recommend focusing efforts there instead.

tooomm:
Do you have experience in setting up windows builds at GitHub Actions? @Fenrirthviti 
We happily take any help there...

@ebbit1q tried it already, he can give you more details on what did not work and link you to his test builds.

tooomm:
As it looks right now, the Windows builds are not faster at GitHub Actions and take the same amount of time as on AppVeyor before.

This issue shipped over.

PR #3771, opened 2019-06-30
travis: use identical osx environments again

tooomm:
Travis wasn't installing qt properly anymore

- update homebrew again
- use same osx environment for `Debug` and `Release` again

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ matrix:
   - name: macOS (Debug)
     if: tag IS NOT present
     os: osx
-    osx_image: xcode10.1
+    osx_image: xcode9.2
     cache: ccache
     addons:
       homebrew:
@@ -96,6 +96,7 @@ matrix:
         - protobuf
         - qt
         - xz
+        update: true
     script: bash ./.ci/travis-compile.sh --server --install --debug
 
   - name: macOS (Release)


ebbit1q:
14 mins, oof

tooomm:
Well, yours is 20min 😉: https://travis-ci.org/Cockatrice/Cockatrice/builds/552420906

The old macOS image we have to use to keep backwards compatibility, combined with homebrew update is one issue why it takes so much time.
But also the compiler cache doesn't fully work for our macOS builds I think...

At least there is some caching working.
At Appveyor, we have no compiler cache setup at all (https://github.com/Cockatrice/Cockatrice/issues/3059). The builds there could be way faster I guess.

ZeldaZach:
What do we want to do with this?

tooomm:
In https://github.com/Cockatrice/Cockatrice/pull/3770 (https://github.com/Cockatrice/Cockatrice/pull/3770/commits/a042d8da35991214eab0e2e701faf1c9dc035509) @ebbit1q enabled the homebrew update again to make travis pass, but that leads to long build times again.
Back then we switched to two different osx images to be able to skip the hombrew update for quicker build times in debug at least... now that we have to update again anyway we can use identical environments for debug and release to test with the same environment we deploy with.

tooomm:
Included in https://github.com/Cockatrice/Cockatrice/pull/3776

PR #3772, opened 2019-07-02
Enable tear-off on menus.

basicer:
- Menus are hard to use on tablets, so let users tear them off.

## Screenshots

image


ZeldaZach:
These seem to persist if I go into other tabs. Ex: Playing a game, popout a menu, then click Chat room. Now the popups are over the chat room. The popups are still functional while overlaying.

Popups (correctly) fade out if you close the game tab.

tooomm:
Do you want to make this a default off option in settings as talked about on Gitter? @basicer

ZeldaZach:
I'm fine putting this in as-is and we can address it further if needed at a later date.

PR #3773, opened 2019-07-07
replace old mulligan with new behavior

### LINKED ISSUE #3222: Suggestion: Change the ctrl+m shortcut default
No other 2 key shortcut can destroy a game state like an accidental mulligan. Generally I'm against getting in the way of people and stupid, but I also feel like changing the mulligan shortcut to ctrl+shift+m or ctrl+alt+m is a simple precaution that'll help against the most detrimental misclick. 

Most players won't ever change their shortcuts or even know where they are, or that there are even multiple tabs in their settings.

### LINKED ISSUE #3749: improve move to bottom of library from hand (new mulligan rule)
Announcing article: https://magic.wizards.com/en/articles/archive/news/london-mulligan-2019-06-03

With `CoreSet 2020` there will be the new `London Mulligan Rule`.
It requires people to put single cards (or more) from their hand to the bottom of the library.

Currently, users can right click a card in their hand (or select several), and chose
`Move to --> Bottom of library in random order`
While that works and accommodates the new mulligan rule, the title of that option is really weird and quite confusing. "Single card in random order"?
But also when you select several cards at once it's difficult to understand how the random part works and if it gets performed correctly due to https://github.com/Cockatrice/Cockatrice/issues/3553.

PR which introduced the feature: https://github.com/Cockatrice/Cockatrice/pull/3549

CC: @basicer

ebbit1q:
london mulligan is now the default
paris mulligan is still accessible through a setting
starting hand size is now configurable
the amount of cards drawn is now always mentioned in the message log

## Related Ticket(s)
- Fixes #3749
- Fixes #3222

## Screenshots
image
(large screen settings window sucks btw)

### DIFF
--- a/common/server_player.cpp
+++ b/common/server_player.cpp
@@ -90,7 +90,7 @@ Server_Player::Server_Player(Server_Game *_game,
                              bool _judge,
                              Server_AbstractUserInterface *_userInterface)
     : ServerInfo_User_Container(_userInfo), game(_game), userInterface(_userInterface), deck(nullptr), pingTime(0),
-      playerId(_playerId), spectator(_spectator), judge(_judge), initialCards(0), nextCardId(0), readyStart(false),
+      playerId(_playerId), spectator(_spectator), judge(_judge), nextCardId(0), readyStart(false),
       conceded(false), sideboardLocked(true)
 {
 }
@@ -165,8 +165,6 @@ void Server_Player::setupZones()
     addCounter(new Server_Counter(5, "g", makeColor(150, 255, 150), 20, 0));
     addCounter(new Server_Counter(6, "x", makeColor(255, 255, 255), 20, 0));
     addCounter(new Server_Counter(7, "storm", makeColor(255, 150, 30), 20, 0));
-
-    initialCards = 7;
 
     // ------------------------------------------------------------------
 
@@ -980,7 +978,7 @@ Server_Player::cmdShuffle(const Command_Shuffle &cmd, ResponseContainer & /*rc*/
 }
 
 Response::ResponseCode
-Server_Player::cmdMulligan(const Command_Mulligan & /*cmd*/, ResponseContainer & /*rc*/, GameEventStorage &ges)
+Server_Player::cmdMulligan(const Command_Mulligan &cmd, ResponseContainer & /*rc*/, GameEventStorage &ges)
 {
     if (spectator) {
         return Response::RespFunctionNotAllowed;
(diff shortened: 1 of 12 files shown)


ZeldaZach:
Updated to use Radio buttons (defaults to new):
Screenshot 2019-07-07 22 19 07

tooomm:
When I saw this PR yesterday, I immediatly thought: "Why not do it as drop down?"

Zach's change goes in that direction to show and name both options which is easier to understand as a user. But the presentation is not ideal. There is no headline or explanation what you are about to select.

Zach's solution in different style:
zach radio

Dropdown versions:
drop down
drop down2

I like the drop down version as it hides unneeded options completely as users probably never change this setting. It's also easier to expand without adding more and more to the settings interface.

ebbit1q:
Some users won't understand what a "vancouver" mulligan means, I liked the checkbox best.

ZeldaZach:
I like the drop down, and we can explain with toolTip or an extra label field below the selection. Want me to look into that tonight?

tooomm:
Yes. We should have simple explanations for all options in any case.

However, in the issue there was a new solution mentioned. A different approach.

common/server_player.cpp
```diff
@@ -994,27 +992,24 @@ Server_Player::cmdMulligan(const Command_Mulligan & /*cmd*/, ResponseContainer &
...
+    auto cardsToMove = QList<const CardToMove *>();
+    for (auto &card : hand->getCards()) {
         auto *cardToMove = new CardToMove;
-        cardToMove->set_card_id(hand->getCards().first()->getId());
-        moveCard(ges, hand, QList<const CardToMove *>() << cardToMove, deck, 0, 0, false);
-        delete cardToMove;
+        cardToMove->set_card_id(card->getId());
+        cardsToMove.append(cardToMove);
     }
+    moveCard(ges, hand, cardsToMove, deck, -1, 0, false);
```
ctrlaltca: Seems like `cardsToMove` leaks memory, qDeleteAll()?

ebbit1q: I was worried about that, seems like I forgot about it before pushing.

basicer:
It feels strange to me to mention the scry, or putting cards on the bottom here, as the mulligan command doesnt do either of those things.  I would change to to say "Redraw hand with one less card" or "Redraw hand with same number of cards"

ZeldaZach:
Screenshot 2019-07-09 22 46 03
Screenshot 2019-07-09 22 46 06

tooomm:
> I prefer to name the options `2019 London Mulligan` and `2015 Vancouver Mulligan`. The year numbers should help with understanding and selecting. 
>
Also, i think the vancouver mulligan option should be called  `1997 Paris Mulligan` as the additinal scry in the vancouver variant is not performed automatically. ~~Maybe that should be indicated somewhere, same for the put to bottom.~~
Or use a combined label: `1997 Paris / 2015 Vancouver Mulligan`

ZeldaZach:
Getting too complex and we don’t have the space for such long messages. The msg that populates next to the box should be sufficient imo. It explains what is happening well

ebbit1q:
or you know a bool as there are only two states

ebbit1q:
I have chosen to remove the entries from settings and instead use a dialog as suggested by this comment on the issue: https://github.com/Cockatrice/Cockatrice/issues/3749#issuecomment-509159693

ebbit1q:
This solves multiple problems:
-we no longer need to confuse users with settings, it will by default display 7 and you can just click ok
-this added confirmation step will avoid mulliganing by accident during a game
-power users can opt to set it to 0 or -1 to change it relative to their current hand size

The only downside is that it is an extra step that will slightly disrupt your game flow, the new key sequence to perform an old vancouver mulligan will be: ctrl-e 7 ret ctrl-m -1 ret ctrl w 1 ret, while for the new mulligan you can just use the default and go: ctrl-m ret, then proceed to move cards to bottom.

skwerlman:
please mark this as fixing #3222 as well, since the heart of that issue is solved by the action leading to a dialog rather than an instant action

Issue #3795, opened 2019-08-16
Feature request: turn order alteration


aeschenkarnos:
In light of the existence of [[Aeon Engine]] it seems that future cards may alter turn order. I suggest that a game owner should be able to arbitrarily reorder the turn order by dragging player names around, including a one-click "reverse all turn orders" which could be put under the kick/PM/friend context menu.

Some players of EDH seem to expect that their turns should go in the dice order rather than "winner first then standard order" which is current best practice. Letting the game owner re-order the turn order would implement this, for those who want to.

skwerlman:
duplicate #3062

aeschenkarnos:
I disagree, #3062 asks for *only* randomization of the order. Though in a 4-player game you could just repeatedly randomize the order and get the result you want after a dozen tries or so, it's more tedious than a simple rearrange. I do agree that randomization is also useful (though can be done with die rolls and then intentional arrangement).

ctrlaltca:
[[Aeon Engine]] fixed by #3802.
Being able to arbitrarily reorder the turn is not possible with the current way turn order is handled, but maybe will be implemented in the future if needed.

ZeldaZach:
I still want this feature, but it's a low priority. It would require a lot of changes to the game definitions system and player code, which is a lot of work. I think it'll happen one day, but just not today.

PR #3802, opened 2019-08-23
Add a new command to reverse turn order

ctrlaltca:
## Related Ticket(s)
- Related #3795

The effect of some cards can alter the turn order.
Currently servatrice's turn order id based on the players' ids, so we can't just reorder the players arbitrarily without breaking the id <-> player association (that would cause a lot of problems).

By now, the most requested feature is just to be able to just reverse turn order, that is implemeted in this PR as a new command / event.

## Screenshots
Schermata 2019-08-23 alle 12 22 53
Schermata 2019-08-23 alle 12 23 15

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -530,6 +530,13 @@ void MessageLogWidget::logRevealCards(Player *player,
     }
 }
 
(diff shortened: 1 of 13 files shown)


ZeldaZach:
Will review after GPLV

PR #3808, opened 2019-09-06
Update Dockerfile to inc missing lib, not build dbconv, and use ENTRYPOINT

dualmoon:
This commit updates the Dockerfile to do the following:

1. Include `libmysqlclient-dev` so that server will be built with SQL support
2. Not build `dbconverter` target because of compilation errors, and because it is not needed here
3. Swap `CMD` for `ENTRYPOINT` so that args on the command line when running the docker are carried over

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ RUN apt-get update && apt-get install -y\
   git\
   libprotobuf-dev\
   libqt5sql5-mysql\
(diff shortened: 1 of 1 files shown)


ctrlaltca:
Merged, thank you!

ZeldaZach:
Much appreciated!

PR #3810, opened 2019-09-08
Settings: default to current chosen directory while opening file dialogs

### LINKED ISSUE #3809: Saving pictures on a external hard drive unlinks on restart
<b>System Information:</b>
Win 7 SP1
Version 2.7.2 (2019-08-31) (this problem has persisted for a long time through many versions)_

I have my pictures folder on a usb hard drive, and whenever I restart cockatrice and click to change the location, it opens program files/Cockatrice. I think it still downloads to the correct location, just opens the wrong one when I try to change it. 
Note: I have cockatrice installed in a custom location on C:
I get pictures not working correctly every now and then even though I always disable all promos, and I thought it was linked to this, but I'm not sure how.
I didn't paste the logs, because I couldn't find any relevant info(related to card pictures) in them throughout the process of restarting, changing location, and loading up a card picture (only says "found on disk")
1: after restart
image
2: after clicking [...]
image

ctrlaltca:
## Related Ticket(s)
- Fixes #3809

In the settings window some buttons open a "choose file/directory" dialog.
These dialogs by default starts browsing the filesystem from the directory where cockatrice is started from.

The dialogs will open pointing to the directory currently chosen for the specific setting.

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -178,7 +178,7 @@ QString GeneralSettingsPage::languageName(const QString &qmFile)
 
 void GeneralSettingsPage::deckPathButtonClicked()
 {
-    QString path = QFileDialog::getExistingDirectory(this, tr("Choose path"));
+    QString path = QFileDialog::getExistingDirectory(this, tr("Choose path"), deckPathEdit->text());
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Thanks!

PR #3814, opened 2019-09-12
travis: update os name for mac

tooomm:
Apple rebranded their operating system and uses `macOS` since a while now.

- replace `osx` with `macos` for the user facing binary for installation


ZeldaZach:
Looks fin eto me

PR #3816, opened 2019-09-17
Oracle: fix crash on no card type; fix #3815

### LINKED ISSUE #3815: Oracle crashes on parsing sets
This issue has been replicated on both linux and windows versions since the most recent version. A change in mtgjson is most likely the problem considering oracle hasn't been changed then.
image
image
The issue occurs after downloading at the end of the parsing step.

ctrlaltca:
## Related Ticket(s)
- Fixes #3815

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -82,16 +82,17 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
 
 QString OracleImporter::getMainCardType(const QStringList &typeList)
 {
+    if (typeList.isEmpty()) {
+        return {};
+    }
+
     for (const auto &type : mainCardTypes) {
         if (typeList.contains(type)) {
             return type;
         }
     }
-    if (typeList.isEmpty()) {
-        return {};
-    } else {
-        return typeList.first();
-    }
+
+    return typeList.first();
 }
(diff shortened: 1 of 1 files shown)


ebbit1q:
Why not test if `card.value("types")` is empty beforehand?

ctrlaltca:
Afaik uou can't check `isEmpty()` on a `QVariant`, you need to convert it to a specific type.
Anyway, for a personal preference in coding style, I prefer to keep checks on a variable always as near as possible to the actual variable usage. This helps in minimizing the possibilities of introducing problems when in the future the function will be modified (and the dev will have to remember that the function expects the list not to be empty), or sometimes even subtle problems like TOCTOU.

ebbit1q:
Considering that we know that these will trigger on every run at the modern horizons art card fronts, I think the debug line isn't necessary. Also elsewhere we use return QString() instead of {}, but that doesn't really matter.

tooomm:
Is this obsolete now? https://github.com/Cockatrice/Cockatrice/issues/3815#issuecomment-532441193

ebbit1q:
No we still need this fix if in the future we get a json that doesn't contain card types pops up.

ZeldaZach:
I'm fine with this change. Thanks for the fix!

PR #3819, opened 2019-09-22
Travis: add Mojave deploy target and release a zipped .dmg

### LINKED ISSUE #3811: Updater on macOS broken (for non 10.14 systems)
After having a look at the updater code and talking to @ctrlaltca, we tested and yes, the updater finds the manually added `(...)mojave.dmg` first, downloads it and leaves all Mac users on older system with a message that the file can not be installed.

As a quick solution we should remove that manually added binary.

But how to deal with that best? I think this are basically our options:
1) Make 10.14 our default system and update the CI accordingly
2) Change the updater
3) Remove Mojave and make users who want best dark mode support compile for themselves or use binaries not officially maintained and hosted by the development team

Problem with 1) is that it's a cut the community, which tends to be not only tech savy people on a lot of older hardware, is not really ready for I guess.

The second makes us updating the code all the time and we have to maintain the updater as e.g. newer versions arrive. That ends in some users having trouble updating again because they stopped playing for a while and their outdated version doesn't find the correct binaries any more.
We use a general check for a `.dmg` file currently, which required no further logic.

Leaves us with 3)...
As a workaround, ctrlaltca had the idea to provide the mojave.dmg in a zip.
But is it worth it? I don't think it's a good practice to manually add files to the automated CI process in the first place.

Is it that difficult to compile on mac if somebody really wants dark mode?
Or maybe skwerlman or somebody else can host the binary and we link to it in the unofficial downloads section in the wiki (https://github.com/Cockatrice/Cockatrice/wiki/Unofficial-Developer-Builds).

What do you think? Ideas?

CC: @ZeldaZach

tooomm:
## Related Ticket(s)
- Fixes #3811

The macOS updater always found the manually added Mojave.dmg and broke updating for macOS <10.14.
We provide that file zipped now for manual installation.
But all compiled and uploaded by hand still.

- Automate the process in Travis CI

<br>

Open question:
- Do we need the separate Mojave debug build?
Or is it enough to have the Sierra build for that and only do an additional release build for Mojave?
- Update homebrew for Mojave? That adds a few minutes to the build but gives us always newest versions. Right now the image has Qt 5.13.1, Cmake 3.15.3, protobuf 3.9.1...


ZeldaZach:
Conflicts to resolve

tooomm:
>Open question:
>
>    - Do we need the separate Mojave debug build?
    Or is it enough to have the Sierra build for that and only do an additional release build for Mojave?
>    - Update homebrew for Mojave? That adds a few minutes to the build but gives us always newest versions. Right now the image has Qt 5.13.1, Cmake 3.15.3, protobuf 3.9.1...

CC @ctrlaltca

tooomm:
So we do want to update homebrew on purpse now for mojave? @ebbit1q

ebbit1q:
Yes, it's not worth the trouble.

tooomm:
Then we are ready to merge (and test the zip deploy) as having a debug build for both - high Sierra and Mojave - is not redundant either. At least I assume so because nobody complained.

ebbit1q:
We'll see how they fare, we can always remove thme if the new build takes too long because the queues for it look to be longer.

tooomm:
Im not sure anymore if they share the same queue or if several osx builds are allowed In parallel in the free tier.

PR #3822, opened 2019-09-25
[WIP] add counter permanency toggle

### LINKED ISSUE #3820: Add ability to add counters to cards when on the stack.
We should add the ability to add counters to cards that are on the stack in cockatrice, and have those counters stay on the cards when moving from battlefield and back... 
This would be useful for another more intuitive way of tracking commander tax (especially useful for decks that actually use their storm count, or with multiple commanders), but also would be particularly useful for suspend cards.
(And I guess for fringe cases like Skullbriar). 
I brought this up on the dev channel on discord and Zach agreed that it was a good idea.

ebbit1q:
## Related Ticket(s)
- Fixes #3820

Certain cards retain their cards while changing zones

- a toggle will be added to the counters menu that enables counter permanency
- because there are so many counters options they have been moved to their own menu
- the counters menu is now also accessible from the stack

This still has a list of required fixes, is this the way we want to implement this?


tooomm:
Quick notes and first discoveries:
- No message in chat log that counters are locked / mode changed for card x
- No visual indicator on the card/counters themselves that there is some different behaviour ("toggle untapping" has a purple border around the card for example)
- Is it possible to have a checkmark next to the right click menu option to make it's mode very clear? Like in the `View` main menu (same should go for toggle untapping)

ebbit1q:
I'm already aware of these issues, in addition: counters are not visible in other zones while they do exist and counters cannot be changed outside of play.
The question is, is this the right direction? Cockatrice is starting to become a ui mess with all the options, it's scaring new players.

ZeldaZach:
I'll play around with this later tonight and report back

ZeldaZach:
I tested this out. The counter permanency works as expected across zones. After reading the code, I see copying a card with counters and permanency enabled doesn't copy them. This (might) be a problem, but i could see a reason for not copying them as in Magic you usually don't copy counters when copying a creature or permanent. 

We've also heard from users in the past they don't like a "counters menu" as it adds additional clicks to playing the game. Maybe have this as a thing at the bottom of the normal menu?

I do agree we need to try and make the UI more in, for a future ticket

tooomm:
>I see copying a card with counters and permanency enabled doesn't copy them.

I think that is fine as you copy the card, not the attachments of the card (even if some are locked to the card).

>We've also heard from users in the past they don't like a "counters menu" as it adds additional clicks to playing the game

I'm pretty sure it was only one user. At least here.
I support that change and prefer a cleaner organized interface. On small screens the very big context menu can be a problem too. And there are still shortcuts.
As @ebbit1q said ("Cockatrice is starting to become a ui mess with all the options, it's scaring new players.") better organize and group them instead overloading the right click menu.
Otherwise half menu would be for counters only...

<br>

**Current list of known issues:**
- No message in chat log that counters are locked / mode changed for card x
- No visual indicator on the card/counters themselves that there is some different behaviour ("toggle untapping" has a purple border around the card for example - have a purple border around counters might work too?)
- Have a checkmark next to the right click menu toggle option to make it's mode clear.
Like the "Always reveal top card" right click option for the library (same should go for toggle untapping btw!)
- Counters are not visible in other zones while they do exist
- Counters cannot be changed outside of play
- I miss an option to clear/remove all counters at once. Maybe even for each color separately + all?

ZeldaZach:
cc: @ebbit1q any chance you can get back to this PR?

ZeldaZach:
@ebbit1q Want to do anything w/ this before our release?

ZeldaZach:
Closing due to lack of attempts, we might return to this if it becomes more relevant

PR #3826, opened 2019-10-01
add disable tearoffmenu option

### LINKED ISSUE #3780: New tear-off menus should be disabled by default + provide option to turn them on
Feature was introduced in https://github.com/Cockatrice/Cockatrice/pull/3772.
The requested addition of it being not the default and allowing users to enable it on demand in the settings is not provided yet.

Background: The feature is mostly useful for tablet/touch users, which are the big minority and hardly present at all.

CC: @basicer / @ZeldaZach

ebbit1q:
## Related Ticket(s)
- Fixes #3780
tearoffmenus were added but no option to disable them existed.

- a new convenient class that manages if a menu should be tearoff-able is added
- a checkbox in the interface settings is added for disabling tearoff is added

## Screenshots
image

### DIFF
--- a/cockatrice/src/abstractcounter.cpp
+++ b/cockatrice/src/abstractcounter.cpp
@@ -30,13 +30,12 @@ AbstractCounter::AbstractCounter(Player *_player,
     shortcutActive = false;
 
     if (player->getLocalOrJudge()) {
-        menu = new QMenu(name);
-        menu->setTearOffEnabled(true);
+        menu = new TearOffMenu(name);
         aSet = new QAction(this);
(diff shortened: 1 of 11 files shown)


ebbit1q:
While playing around with this I have also found that the tearoff menus are borderline unstable, some actions will delete the torn off menu to never be seen again.

ZeldaZach:
I like, nice change! We meant to do this originally, but I probably borked that one up

tooomm:
Note: tear-off menus are (still) the default

ebbit1q:
Well they were already implemented and turned on, less change more better.

PR #3827, opened 2019-10-01
add move top card to stack keybind

### LINKED ISSUE #3824: Cascade QOL improvement suggestion
Would it be possible to create  a command for Cascade? Like click on a card -> Command : Cascade -> Action : Puts cards on the stack from the top of the library until a lower cmc card is revealed?

ebbit1q:
## Related Ticket(s)
- Fixes #3824

Cascading is as difficult as it is in real life

- action and keybind to move top card to stack has been added

## Screenshots
image


ZeldaZach:
Code looks right, but I want to test first. Will check out tonight

PR #3831, opened 2019-10-02
Cockatrice Docker Image

alfoltran:
## Related Ticket(s)
Scripts for client image generation. I am using this image to run the Cockatrice on Fedora.

I don't know if this is the best place to store these scripts, but these are the scripts I used to create and maintain the images available in [Docker Hub](https://hub.docker.com/r/cockatrice/cockatrice).

I would also like to pass Cockatrice Docker Hub access credentials. So you can upload the image of the servatrice if you wish.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,8 +1,7 @@
 FROM ubuntu:18.04
-MAINTAINER Alfredo Foltran <alfoltran@gmail.com>
 
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update \
+RUN apt-get update -y --no-install-recommends \
     && apt-get install -y --no-install-recommends \
         ca-certificates \
         curl \
@@ -21,6 +20,8 @@ RUN apt-get update \
         libqt5websockets5 \
         libqt5widgets5 \
     && dpkg -i ./Cockatrice-Faerie_Tales-2.7.2-UbuntuBionic.deb \
-    && apt-get install -f
+    && apt-get install -f \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
 CMD cockatrice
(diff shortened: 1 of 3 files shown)


tooomm:
Did you see our [Dockerfile](https://github.com/Cockatrice/Cockatrice/blob/master/Dockerfile)? Because you created a new one

alfoltran:
> Did you see our [Dockerfile](https://github.com/Cockatrice/Cockatrice/blob/master/Dockerfile)? Because you created a new one

This is for the client, not for the server.

ZeldaZach:
Looks like a quality change to me. Uncertain if the location is best, but it's a good tool to have around. @tooomm We only provide a servatrice instance on Docker currently. This adds a cockatrice instance.

Would it make sense to merge the two setups together? (I don't know enough about docker to know if that's bad practice)

tooomm:
I assume having a dockerfile-servatrice and dockerfile-cockatrice is also not „allowed“?

tooomm:
@alfoltran we are lacking Docker experience here. Can you have a look again and help us, please?

About Docker Hub:
I think it would be easy to include into our deployment routine.

ZeldaZach:
Could be related to #3974, would love to see consistency across the docker operations we have

alfoltran:
> 
> 
> Could be related to #3974, would love to see consistency across the docker operations we have

There is no relationship here. However I can give my feedback here and say that it would not be a good idea to have two compose.

First, with a little work on the host, docker-compose.yml works on Windows without changes.

Second, if you want to make things easier, just use docker-compose.yml.windows, because it works on Linux too.

ZeldaZach:
If you're willing to get them combined into one file, would be greatly appreciated

ZeldaZach:
Hey @alfoltran just a bump on this PR. Looking to cleanup some PRs and wasn't sure if you were still looking to solve this issue! Reach out https://cockatrice.us/discord if you have any concern or would like to chat!

ZeldaZach:
Closing for now, due to lack of responses and interest. Feel free to re-file if you'd like to see this added

PR #3832, opened 2019-10-02
[WIP] add shortcuts for readying and sideboarding in the lobby

ebbit1q:
When I think of cockatrice I think of a sprawling mess of possibly conflicting shortcuts, this is why I think we need more shortcuts to pursue this vision further.

- configurable shortcuts for readying for a game and unlocking the sideboard will be added

Currently the lock shortcut only works once and I don't know why, it's been annoying me for a while. Does just changing the text throw off the shortcut?


ZeldaZach:
I like, nice work!

ebbit1q:
Should we either revert this merger of this pr that I marked WIP or should I open a ticket for the relock sideboard shortcut not working?

PR #3851, opened 2019-10-29
link to new AllPrintings file

tooomm:
MTGJSON changed their endpoint, see https://mtgjson.com/changelog/#_4-6-0-2019-10-28
But redirects should be in place, so nobody will be cut off.

- use new `AllPrintings` file name
- use new file location in `\files\` folder (used to be `\json\`)

Makes us independent from the redirect.

Tested with the windows build from appveyor. 👍


tooomm:
Travis is having issues with building mac currently, builds don't start... 🙈

Issue #473, opened 2014-12-03
Can't distinguish different tokens with the same name


Psithief:
Has anyone had any ideas how Cockatrice can improve tokens handling?
## Problems
- Tokens with the same name are differentiated only by increasing amounts of whitespace at the end of the token name. This whitespace does not survive copying to clipboard/loading from clipboard.
  - Tokens also get put in the main deck if loaded from clipboard. (#472)
- ~~Cockatrice users don't get an automatic tokens update as with Oracle, and there isn't a set of instructions on how to download and install a tokens database.~~
- Wizards of the Coast is also releasing some sets with variants of the same token, e.g. the two white soldiers in Theros and the two white Warriors in Khans of Tarkir.

ZeldaZach:
> Cockatrice users don't get an automatic tokens update as with Oracle, and there isn't a set of instructions on how to download and install a tokens database.

Now they do

Fenrirthviti:
Can confirm this is still a pain point, especially in EDH format where the board state can get pretty complicated. I'm going to add an additional point for clones, with the following example:

![img](https://pub.rachni.com/img/cockatrice_2020-11-24_03-33-10.png)

One of the cards here is a token copy that is a clone of the card. There is no way of telling which one is which, and can lead to annoyance/confusion when needing to, say, destroy one. One will disappear when destroyed, but the other goes to the graveyard, and knowing which needs to be chosen can be important to avoid having to roll back the board state if you accidentally choose the wrong one. We can use annotations as a workaround, but that becomes fairly tedious.

There should be a clear UI paradigm that indicates a clone is a token (maybe a colored border of some kind, or just adding `(Token)` to the card name).

Tracking copies of the same token by incrementing a number would be sufficient for most use cases I can think of.

ebbit1q:
@Fenrirthviti that is a separate issue, this issue is aimed at the database being indexed by name.

Fenrirthviti:
Ah! My misunderstanding. Is there a separate issue I should move those comments to? Or I can create a new one if one does not exist currently.

ebbit1q:
I can't find any open issues on this, you should probably make a new issue.

tooomm:
>this issue is aimed at the database being indexed by name.

Wouldn't this then be a dupe of #2545?

<br>

@Fenrirthviti 
[...]

PR #3870, opened 2019-11-23
Trim tokens as this is our most common issue

ZeldaZach:
Users seem to have issues with the tokens as extra spaces are accidentally included. This trims tokens before sending to server, alleviating this issue.

This is a big issue for my moderation team. Tested and works as expected with a new account registered.


ctrlaltca:
Won't this break all the tokens that have the same name but different stats? We currently use the "whitespace workaround" to distinguish them: #473

tooomm:
@ctrlaltca I think Zach is talking about the token for account activation.

ctrlaltca:
I should really have a coffee on the morning before starting to review code.. sorry

PR #3874, opened 2019-12-10
Refactoring file cockatrice/src/messagelogwidget.cpp

kopcion:
## Related Ticket(s)
- TODO found in code

Created separate class in order to define phase

- removed long switch-case from messagelogwidget.cpp file


ebbit1q:
Thank you for getting to fixing that, you should remove the separate sql changing commits :)

ebbit1q:
Also not marking the functions called "constant" in the code not const is sort of counterintuitive.

kopcion:
About those commits, I have noticed too late, that's why I have closed the PR

ebbit1q:
noone is going to merge it in the time that you are sorting that out, don't stress :)

PR #3875, opened 2019-12-10
Refactor messagelogwidget

kopcion:
## Related Ticket(s)
- TODO found in code

- Created separate class in order to define phases

- removed long switch-case from messagelogwidget.cpp file


ebbit1q:
This is great, I would have simply force pushed to the old pr but this works too.

You should be able to run the clangformat commands to automatically fix those style issues.

cockatrice/src/messagelogwidget.cpp
```diff
@@ -559,73 +559,16 @@ void MessageLogWidget::logSay(Player *player, QString message)
...
+    Phase phase = Phases::getPhase(phaseNumber);
+
+    if (phase.name != "Unknown Phase") {
+        soundEngine->playSound(phase.soundFileName);
     }
-    appendHtml("<font color=\"" + color + "\"><b>" + QDateTime::currentDateTime().toString("[hh:mm:ss] ") +
-               QString("%1").arg(phaseName) + "</b></font>");
+
+    appendHtml("<font color=\"" + phase.color + "\"><b>" + QDateTime::currentDateTime().toString("[hh:mm:ss] ") +
+               QString("%1").arg(tr(phase.name.toStdString().c_str())) + "</b></font>");
```
ctrlaltca: I see a possible improvement and an issue here:
 * improvement: use `phase.name.toUtf8().data()`
 * issue: since the string literals are no more inside explicit `tr()` calls, qt's `lupdate` won't find these anymore, so they won't end up in the file containing the translatable strings. You'll need to use `QT_TRANSLATE_NOOP ()`.

tooomm:
@kopcion Can you please look into the requested changes?
If you need help or more details let us know!

ZeldaZach:
Can someone try picking this feature up? Would love to see it get merged

cockatrice/src/messagelogwidget.cpp
ebbit1q: I have addressed these and moved the tr calls to phase.cpp

cockatrice/src/phase.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    if (0 <= phase && phase < Phases::phaseTypesCount) {
+        return phases[phase];
+    } else {
+        return unknownPhase;
+    }
+}
+
+const Phase Phases::unknownPhase(QT_TRANSLATE_NOOP("Phase", "Unknown Phase"), "black", "unknown_phase");
+const Phase Phases::phases[Phases::phaseTypesCount] = {
```
ZeldaZach, line 18: I feel like these should be in the header file

ebbit1q: Either works, didn't bother moving them from the original pr.

ebbit1q: actually they are static and should be in the implementation file.

tooomm:
This should be ready @ZeldaZach?

ZeldaZach:
Agreed

PR #3883, opened 2020-01-06
Hide games created/hosted by people on your Ignore List

### LINKED ISSUE #724: Feature Request: Filter to hide games created/hosted by people you ignore
Hi, I was wondering if it would be at all possible to add a feature to the ignore list that would prevent games created by people on that list from showing in the games list?

Obviously not a high priority as you can simply choose not to join their game. The benefit comes when you ignored someone a while ago and don't recall their specific name, instead of checking your ignore list, their game simply won't show.

As I said, it's not really a necessity, just a simple feature that would fit well I think.

Thanks

pmwheatley:
## Related Ticket(s)
- Fixes #724 

Implemented new filter on the Game List to allow hiding games created by users who are on your Ignore List.

- New checkbox in Filter Dialog to enable hiding games whose creator is ignored.
- Game Selector now responds to Ignore List Received and Add To/Remove From List signals, and refreshes the Game List accordingly.

## Screenshots
Screenshot from 2020-01-06 22-12-02

### DIFF
--- a/cockatrice/src/settings/gamefilterssettings.cpp
+++ b/cockatrice/src/settings/gamefilterssettings.cpp
@@ -48,6 +48,17 @@ bool GameFiltersSettings::isShowPasswordProtectedGames()
     return previous == QVariant() ? true : previous.toBool();
 }
 
+void GameFiltersSettings::setHideIgnoredUserGames(bool hide)
+{
+    setValue(hide, "hide_ignored_user_games", "filter_games");
+}
+
+bool GameFiltersSettings::isHideIgnoredUserGames()
+{
+    QVariant previous = getValue("hide_ignored_user_games", "filter_games");
+    return previous == QVariant() ? false : previous.toBool();
+}
+
 void GameFiltersSettings::setGameNameFilter(QString gameName)
 {
     setValue(gameName, "game_name_filter", "filter_games");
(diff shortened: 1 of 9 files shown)


cockatrice/src/settings/gamefilterssettings.cpp
```diff
@@ -48,6 +48,17 @@ bool GameFiltersSettings::isShowPasswordProtectedGames()
...
 
+void GameFiltersSettings::setHideIgnoredUserGames(bool hide)
+{
+    setValue(hide, "hide_ignored_user_games", "filter_games");
+}
+
+bool GameFiltersSettings::isHideIgnoredUserGames()
+{
+    QVariant previous = getValue("hide_ignored_user_games", "filter_games");
+    return previous == QVariant() ? false : previous.toBool();
```
ebbit1q: shouldn't we be using isNull() here and the rest of the uses?

pmwheatley: As far as I can tell, checking `isNull()` is not strictly equivalent to equality-checking invalid (i.e. unset) variants, as `QVariant` itself may be set with valid null values. I'm pretty new to Qt though, so may well have misinterpreted the docs here (https://doc.qt.io/qt-5/qvariant.html#QVariant).

ZeldaZach:
Thanks for the change, much appreciated!!

PR #3887, opened 2020-01-13
Docker-compose setup for Servatrice

pmwheatley:

Initial setup of Servatrice (even via docker) is more involved than necessary. Using docker-compose, you can manage the running of both the mysql server and a pre-configured Servatrice instance, with a minimum of setup.

- Added docker-compose specific `servatrice-docker.ini` file, containing the database setup for docker-compose.
- `docker-compose.yml` file to setup and run both containers/services (i.e. mysql and servatrice).
- The existing `Dockerfile` is unchanged in this PR as it is used in the CI, but there are a number of improvements that _could_ be made if the docker-compose approach proves useful.


ebbit1q:
Why not simply use the same default files?

pmwheatley:
@ebbit1q after a second look, the sql file was unnecessary, and rather than duplicate the entire example ini file, I've just extracted the database section which is specific to the docker-compose section. Hope that's better 😄!

ebbit1q:
it's great! I'll test it in a bit.

tooomm:
Can we include some testing for the docker and docker compose file in the ci?

Like check if they properly build, run, listen+reply to protobuf communication and if the database setup is working for example?

PR #3889, opened 2020-01-17
support mtgo .dek files

skwerlman:
they use plain text internally so we just need to reveal them in the load dialog

## Related Ticket(s)
- N/A

MTGO .dek files didnt show up in the "load deck" dialog

- .dek files should show up in the "load deck" dialog

### DIFF
--- a/cockatrice/src/deck_loader.cpp
+++ b/cockatrice/src/deck_loader.cpp
@@ -7,7 +7,7 @@
 #include <QStringList>
 
 const QStringList DeckLoader::fileNameFilters =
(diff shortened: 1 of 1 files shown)


tooomm:
What about saving the format, too?

skwerlman:
saving uses the same info as loading: https://github.com/Cockatrice/Cockatrice/blob/7bfefee073d0eb04e89136d5ce3a8aebf904c283/cockatrice/src/tab_deck_editor.cpp#L823

ebbit1q:
Great, we can always use more newlines at ends of files.

PR #3898, opened 2020-02-20
Add keyboard shorcuts to focus and unfocus chat

Xenos6666:
It isn't possible to focus and unfocus the in-game chat with keyboard shortcuts only

- A new shortcut has been added to focus the chat box in the game tab
- A new shortcut has been added that unfocus any line input in the main window

_(Unrelated: added include sorting to clang-format configuration)_


Xenos6666:
I'm not very familiar with the Qt library, and for some reason I needed to add the shortcut action to a menu for it to work. Is there a way to make it work that doesn't require to add it to a menu? This could also be useful for future keyboard control support I plan to work on

ebbit1q:
It should be, I prefer to just add it to a menu as well though.

ebbit1q:
Just checked, ctrl e after clicking chat doesn't work on master either. This is in a local game.

Xenos6666:
Unfocusing the chat worked for me when I tested it in a local game, I'll be testing it further later today when I have the time, I might also take a look at the Ctrl+e bug while I'm at it

Xenos6666:
I did some tests on the Ctrl+e error, it seem like keys that already have a function in chat don't trigger the associated shortcuts, which is the expected behavior (Ctrl+e goes to the end of line), but it's not really a problem, since the main aim of adding the escape behavior was to prevent these conflicts (it also allows people to have key bindings without Ctrl). However, by, fooling around, I found something else, it seems like shortcuts set by the player.cpp file don't work with multiple bindings on the same shortcut, the problem seems to come from the `setShortcutsActive` function, I'm not sure if this is intended behavior.

Xenos6666:
As for the Esc bug, it seems to work as long as the game hasn't started, but stops working after that. I think I'll try implementing it on each instances of LineEditCompleter to make it more stable and prevent problems with the focus being stuck in other parts of the application (for example in the deck editor tab's search bar)

Xenos6666:
Actually, I just noticed that Search bar isn't an instance of LineEditCompleter

Xenos6666:
I had 2 alternatives to have get the behavior I wanted : 
- Catching the Esc key press on the MainWindow, and checking the currently focused Widget
- Catching the Esc key press on the QLineEdit widget itself

The first one had the potential to create bugs and conflicts with other people's codes, while the second one is probably less error prone but required a minor restructuring of many disjointed places of the code, and could cause inconsistencies in future codes. I chose to go with the second one, but if you prefer the first option it wouldn't take me long to re-code it.

Xenos6666:
Wait, the fix didn't actually work, but now I know what causes it, the unfocusing doesn't work if a shortcut is set using the esc key

Xenos6666:
All issues should be fixed

cockatrice/src/customlineedit.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <QLineEdit>
+
+class QKeyEvent;
+class QWidget;
+class QString;
+
+// Should be used when the there is a risk of conflict between line editor
+// shortcuts and other shortcuts
+class CustomLineEdit : public QLineEdit
```
ebbit1q: CustomLineEdit is to vague in my opinion, this is specifically a lineedit that you can unfocus with a key.

cockatrice/src/lineeditcompleter.h
```diff
@@ -3,10 +3,11 @@
 
 #include <QFocusEvent>
 #include <QKeyEvent>
-#include <QLineEdit>
 #include <QStringList>
 
-class LineEditCompleter : public QLineEdit
+#include "customlineedit.h"
```
ebbit1q: project include above others, see https://github.com/Cockatrice/Cockatrice/blob/master/.github/CONTRIBUTING.md#header-files

cockatrice/src/tab_deck_editor.h
```diff
@@ -5,9 +5,9 @@
 #include "tab.h"
 #include <QAbstractItemModel>
 #include <QDir>
-#include <QLineEdit>
 
 #include "carddatabase.h"
```
ebbit1q: this one already was at the wrong spot

cockatrice/src/tab_logs.cpp
```diff
@@ -19,6 +14,13 @@
 #include <QtGui>
 #include <QtWidgets>
 
+#include "abstractclient.h"
```
ebbit1q: same

cockatrice/src/tab_message.cpp
```diff
@@ -6,11 +6,11 @@
 #include "soundengine.h"
 #include <QApplication>
 #include <QDebug>
-#include <QLineEdit>
 #include <QMenu>
 #include <QSystemTrayIcon>
 #include <QVBoxLayout>
 
+#include "customlineedit.h"
```
ebbit1q: oh dear, I guess we're not very consistent

cockatrice/src/window_sets.cpp
```diff
@@ -19,6 +16,11 @@
 
 #include <algorithm>
 
+#include "customlineedit.h"
```
ebbit1q: same

cockatrice/src/customlineedit.cpp
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    QLineEdit::keyPressEvent(event);
+}
+
+bool CustomLineEdit::eventFilter(QObject *watched, QEvent *event)
+{
+    if (event->type() == QEvent::ShortcutOverride) {
+        QKeyEvent *keyEvent = static_cast<QKeyEvent *>(event);
+        if (keyEvent->key() == Qt::Key_Escape) {
```
ebbit1q: Would it be possible to make this shortcut configurable?

ZeldaZach:
I'm pretty sure the review by @ebbit1q Can be solved by running the clang-format file on the codebase

Xenos6666:
Clang format wasn't configured to sort headers, I added a few lines to fix that. I also renamed CustomLineEdit to LineEditUnfocusable, and added a configurable shortcut for unfocusing, as requested. However, I couldn't manage to make a `refreshShortcuts()` function work on the LineEditUnfocusable for some reason, so I decided to check the shortcuts at each key presses done when the widget is focused, this seems very inefficient to me, but it's the only solution I found.

tooomm:
>Clang format wasn't configured to sort headers, I added a few lines to fix that

Very appreciated addition, really nice!
But please do that in a separate PR next time as it is not related to this at all. 👍

Xenos6666:
I'll keep that in mind, thanks! (It's the first time I've made a pull request, I'm learning how things work)

PR #3904, opened 2020-03-01
Keep stars but only sometimes, add scheme cards back

### LINKED ISSUE #3845: Scheme cards are not added
Many scheme cards are printed with a star in their collectors number as the only print, this means they are ignored.

ebbit1q:
## Related Ticket(s)
- Fixes #3845


ebbit1q:
```
this will correct #3706 and #3715 being a bit overzealous in removing
cards and thus fix scheme cards being removed in entirety
fix #3845
note that this causes a lot more cards to be added that are in promo
sets, if these promo sets should prove to be problematic they should be
disabled somehow as having them as an option is still nice.
```

PR #3906, opened 2020-03-06
update deprecated methods in qt5.14 and protobuf 3.4

ebbit1q:
debug builds showed up a bunch of deprecation warnings

- the deprecation warnings should be gone, yay for future builds


ctrlaltca:
Ubuntu Bionic test fails because `ByteSizeLong` has been introduced in protobuf 3.1.0, and the distro is shipping protobuf 3.0.
We should either add some `#ifdef`s or drop bionic, but it would be nice to wait for the next LTS to be released on april.

PR #3907, opened 2020-03-17
change all the numbers to msvc 2019

### LINKED ISSUE #3910: Windows builds are built without ssl support
Currently all windows builds, which are built on appveyor https://ci.appveyor.com/project/ZeldaZach/cockatrice , are broken and do not include SSL meaning they cannot connect to servers using it or download images over https or get card info.

Because of this issue we have retracted release 2.7.3! If you have downloaded 2.7.3 please install 2.7.2 while we resolve these issues.

ebbit1q:
change to msvs2019 to see if it fixes the ssl issue
possibly fixes #3910

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,7 +21,7 @@ skip_branch_with_pr: true
 
 clone_depth: 15
 
-image: Visual Studio 2017
+image: Visual Studio 2019
 
 cache:
     - c:\Tools\vcpkg\installed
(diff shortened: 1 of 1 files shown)


ebbit1q:
https://cmake.org/cmake/help/v3.16/generator/Visual%20Studio%2016%202019.html

ebbit1q:
the latest build: https://ci.appveyor.com/project/ZeldaZach/cockatrice/builds/31529272 , fails with an exception but doesn't say what, the warning above it might be the cause and is generated from this location: https://github.com/Cockatrice/Cockatrice/blob/master/cmake/FindVCredistRuntime.cmake#L26

ZeldaZach:
Gonna close for now and we can revisit

PR #3908, opened 2020-03-17
clangify everything with the new header sorting

ebbit1q:
#3898 changed the .clang-format rules, this pr simply runs clangify.sh --branch "" to make every file compliant.


tooomm:
Do we need to add that to our style guide as well?

ebbit1q:
What needs to be added? everything is already there

PR #3913, opened 2020-03-19
Fix bundling of openssl librearies on x64 windows

ctrlaltca:


ctrlaltca:
Retired, wrong branch

PR #3915, opened 2020-03-19
Make travis builds on fedora 31, debian 10 and arch

ebbit1q:
Because windows on travis is not working I made this instead.

Adds compile steps for building debian, fedora, and arch because we can.

That is provided they work without me testing them once.

### DIFF
--- a/.ci/ArchLinux/Dockerfile
+++ b/.ci/ArchLinux/Dockerfile
@@ -1,0 +1,12 @@
+from archlinux:latest
+
+RUN pacman --sync --refresh --sysupgrade --needed \
+    base-devel \
+    cmake \
+    git \
+    protobuf \
+    qt5-base \
+    qt5-multimedia \
+    qt5-svg \
+    qt5-tools \
+    qt5-websockets
--- a/.ci/ArchLinux/Dockerfile
+++ b/.ci/ArchLinux/Dockerfile
@@ -1,12 +1,13 @@
 from archlinux:latest
 
-RUN pacman --sync --refresh --sysupgrade --needed \
-    base-devel \
-    cmake \
-    git \
-    protobuf \
-    qt5-base \
-    qt5-multimedia \
-    qt5-svg \
-    qt5-tools \
-    qt5-websockets
+RUN pacman --sync --refresh --sysupgrade --needed --noconfirm \
+        base-devel \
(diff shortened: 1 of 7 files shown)


ebbit1q:
I tested everything locally now, if it doesn't work it's on travis.

ebbit1q:
Only the new builds fail with the following error:
```
RCC: Error in '/src/oracle/oracle.qrc': Cannot find file 'resources/oracle.svg'
make[2]: *** [oracle/CMakeFiles/oracle.dir/build.make:63: oracle/qrc_oracle.cpp] Error 1
make[1]: *** [CMakeFiles/Makefile2:491: oracle/CMakeFiles/oracle.dir/all] Error 2
```
This does not happen locally when running the scripts, it makes no sense.

ctrlaltca:
@ebbit1q https://stackoverflow.com/questions/51195528/rcc-error-in-resource-qrc-cannot-find-file-png
The old distros we build on Travis are using Qt < 5.10

ctrlaltca:
Updating travis to use the bionic build images should fix the issue:
`
dist: bionic
`

ebbit1q:
Thanks, I did not expect this bug to be from using an older docker version.

ZeldaZach:
Can you look into getting this ready for a release test? Would like to see all new OSes have a release target (Arch linux doesn't seem to have one, and it's optional right now?)

Looks to be passing all of the tests, which I'm thrilled with!

ebbit1q:
We are already packaged in arch! https://www.archlinux.org/packages/community/x86_64/cockatrice/

Anyway the reason why it's optional is that arch is a rolling release distro, this means it uses the most recent versions of any of the software it uses. The other builds have a specific version of the os they are using, a package built on that version of the distro will keep working on that distro. On arch however there are no "versions", we could get a non backwards compatible change in our dependencies and have the build stop working. This is why we should keep track of this and fix them as soon as possible but not have it block otherwise perfectly functional pull requests.

ZeldaZach:
Seems reasonable to me, thanks!

ebbit1q:
Anyway I think this is ready, I don't think there is any difference in like queue lengths on travis between using ubuntu 18.04 or 16.04 for the base image. It is kind of odd that we are using a docker image to run the exact same os however.

PR #3919, opened 2020-03-20
Travis: skip cleanup before deploy

ctrlaltca:
Maybe this will fix deploying of tagged releases


ebbit1q:
Can we test this without making superfluous releases?

ctrlaltca:
Unfortunately I'm not aware of any method to test this without making a tag.

tooomm:
I think this change is viable and solves our issue. Can respond later today in more detail.

PR #3922, opened 2020-03-21
Handle Release Channels better as we host more packages now

ZeldaZach:


ebbit1q:
Wouldn't it be better if we store these values inside the application while building instead of needing to check at runtime? What if something changes and the same working application can run on a system not checked by these constraints, I'd rather have it try to download the exact same package it was installed from instead of needing to check for it.

tooomm:
I guess with this pr we don't need to package the osx 10.14 release in a zip file anymore?
We used to do that for that package to make users on 10.13 not fail.

That leads me to the questions what happens when this is merged and we don't change the release files right away?
How do we manage the transition period best where users are still on old versions and some are on this release?

ZeldaZach:
Unfortunately there's no good solution for older clients other than to keep using ZIP for the time being.

PR #3934, opened 2020-03-29
Oracle / card xml improvements

### LINKED ISSUE #1610: Oracle and Cockatrice should produce xml schema tags in produced xml files
They can have `xsi:schemaLocation` link to `https://github.com/Cockatrice/Cockatrice/blob/<git_version_hash>/doc/cards.xsd` and `deck.xsd` respectively. This should help with tool validation.

### LINKED ISSUE #2679: cards/tokens.xml should have information about their source data
It's an important information in general, but especially for troubleshooting with this files.
Would be helpful to know where they fetched their data on creation from.

For example was somebody providing a local file? Or did he use @tritoch's allsets.json with spoiler info for example.

It's nice to have this kind of comprehensive information in general!

I suggest that Oracle put's the source download link of it's data into the files. (or the file location + name when a local file is provided)
Date+time of oracle run/xml creacion might also be useful.

Same for the Oracle  version (just use the one we have in Cockatrice `about` window) to quickly troubleshoot errors like with aftermath splitcards for example!

### LINKED ISSUE #3647: Add mtgjson version to oracle cards.xml db + update checker
Include the version (and date?) from mtgjson meta tag into our own card database.

This version can then be checked against https://mtgjson.com/json/version.json to see if new cards or updates are available. Oracle would then only fully run when really needed.

Check could be the first step in Oracle run and/or as `Card Database Update Available` notification in the cockatrice client to notify that the user should run an update.
A bit like how spoilers are done in client.

ctrlaltca:
## Related Ticket(s)
- Fixes #1610
- Fixes #2679
- Partially fixes #3647

See tickets

The xml schema has been updated to includ an optional `<info>` tag containing information about the author, version and date.
The source version will be fetched automatically from mtgjson when oracleis using the predefined url.
The xml now contains a proper schemaLocation attribute pointing to cards.xsd, so there-s no need to specify it anymore when running `xmllint`.

## Example
```xml
<?xml version="1.0" encoding="UTF-8"?>
<cockatrice_carddatabase version="4" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://raw.githubusercontent.com/Cockatrice/Cockatrice/master/doc/carddatabase_v4/cards.xsd">
    <info>
        <author>Cockatrice 2.7.3-custom(9c0f619) (2020-03-29)</author>
        <createdAt>2020-03-29T15:52:29Z</createdAt>
        <sourceUrl>https://www.mtgjson.com/files/AllPrintings.json.xz</sourceUrl>
        <sourceVersion>4.6.2+20200328</sourceVersion>
    </info>
    <sets>
        <set>
...
```

### DIFF
--- a/oracle/src/oraclewizard.cpp
+++ b/oracle/src/oraclewizard.cpp
@@ -39,6 +39,7 @@
 // Xz stream header: 0xFD + "7zXZ"
 #define XZ_SIGNATURE "\xFD\x37\x7A\x58\x5A"
 #define ALLSETS_URL_FALLBACK "https://www.mtgjson.com/files/AllPrintings.json"
+#define MTGJSON_VERSION_URL "https://www.mtgjson.com/files/version.json"
 
 #ifdef HAS_LZMA
 #define ALLSETS_URL "https://www.mtgjson.com/files/AllPrintings.json.xz"
@@ -315,6 +316,8 @@ bool LoadSetsPage::validatePage()
         wizard()->disableButtons();
         setEnabled(false);
 
+        wizard()->setCardSourceVersion("meh");
+
         downloadSetsFile(url);
     } else if (fileRadioButton->isChecked()) {
         QFile setsFile(fileLineEdit->text());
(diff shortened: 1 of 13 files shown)


tooomm:
Awesome! 👏 👏 

This should help with DB troubleshooting and allows for a way smoother update routine for our users.

Should some of this be added to the debug log as well on database load?

ebbit1q:
is it correct that this key is not used by the database loader?
I get a `[CockatriceXml4Parser] Unknown item "info" , trying to continue anyway`

ctrlaltca:
Yep, it's not used yet. Once the info key is in all the cards/token/spoiler files, i'll add proper support for it.

oracle/src/oraclewizard.cpp
```diff
@@ -315,6 +316,8 @@ bool LoadSetsPage::validatePage()
         wizard()->disableButtons();
         setEnabled(false);
 
+        wizard()->setCardSourceVersion("meh");
```
ZeldaZach: :joy:

PR #3946, opened 2020-04-07
Fix #3068: Add hint text about game list filters

### LINKED ISSUE #3068: Add text hint that filter are applied to the game list
Taken from https://github.com/Cockatrice/Cockatrice/issues/3067#issuecomment-362619646
The idea is to clarify to the user that the game list doesn't show all available games right now.

>There could be a simple little hint text **only showing** when filters are active:
filter hint
>
>I totally see people adding filters to find a suitable game, then close trice and come back a week later to wonder why there are just a few games shown. (Cockatrice remembers used filters, which is a nice feature!)

knitknit:
## Related Ticket(s)
- Fixes #3068 

Cockatrice remembers filters applied between sessions, but users were not alerted if any filters were altered from their default values. 

- Display hint text next to former "Clear filters" button, indicating either that default filters are applied or that filters have been altered from default values.
- Rename "Clear filters" to "Reset filters", but keep 'l' as the hotkey. This seems like a more accurate name, as we're not actually clearing all filters (for example, "show password protected games" is true by default and therefore checked in the UI).

## Screenshots

Hint text on initial load, assuming filters are set to default:

default-start

Hint text after altering filters:

applied

Hint text after resetting filters:

reset

### DIFF
--- a/cockatrice/src/gameselector.cpp
+++ b/cockatrice/src/gameselector.cpp
@@ -290,12 +290,13 @@ void GameSelector::actSelectedGameChanged(const QModelIndex &current, const QMod
     joinButton->setEnabled(game.player_count() < game.max_players() || overrideRestrictions);
 }
 
-void GameSelector::setAlteredFiltersText(const int numAlteredFilters) {
+void GameSelector::setAlteredFiltersText(const int numAlteredFilters)
+{
     if (alteredFiltersLabel != nullptr) {
         if (numAlteredFilters == 0) {
             alteredFiltersLabel->setText(tr("Default filters applied"));
         } else {
-            alteredFiltersLabel->setText(tr("%1 filter(s) altered").arg(numAlteredFilters));
+            alteredFiltersLabel->setText(tr("%1 filter(s) applied").arg(numAlteredFilters));
         }
     }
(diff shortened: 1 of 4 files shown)


cockatrice/src/gameselector.cpp
```diff
@@ -282,3 +289,13 @@ void GameSelector::actSelectedGameChanged(const QModelIndex &current, const QMod
     spectateButton->setEnabled(game.spectators_allowed() || overrideRestrictions);
     joinButton->setEnabled(game.player_count() < game.max_players() || overrideRestrictions);
 }
+
+void GameSelector::setAlteredFiltersText(const int numAlteredFilters) {
+    if (alteredFiltersLabel != nullptr) {
+        if (numAlteredFilters == 0) {
+            alteredFiltersLabel->setText(tr("Default filters applied"));
+        } else {
+            alteredFiltersLabel->setText(tr("%1 filter(s) altered").arg(numAlteredFilters));
```
ZeldaZach, line 298: Would use "applied" across the fleet, not "altered" :)

knitknit:
Woohoo! Happy to help :)

PR #3959, opened 2020-04-16
Add ability to filter games older than 2 hours

### LINKED ISSUE #3067: [Feature request] Add age filter for games (if we can't deal with outdated games)
Everyone knows there are outdated games hanging on the server all day long. Seems like we can't deal with them (probably cuz its hard to say if game is outdated or just takes way too long, or maybe its just tenth game in the same lobby), maybe we can at least add an option to filter them out on the age basis?

knitknit:
## Related Ticket(s)
- Fixes #3067

Users want to filter out old games. Starting with 2 hours per https://github.com/Cockatrice/Cockatrice/issues/3067#issuecomment-362580744.

Ability to filter out old games (where "old" == "started more than 2 hours ago") is added to game filtering.

### DIFF
--- a/cockatrice/src/dlg_filter_games.cpp
+++ b/cockatrice/src/dlg_filter_games.cpp
@@ -29,6 +29,9 @@ DlgFilterGames::DlgFilterGames(const QMap<int, QString> &_allGameTypes,
     hideIgnoredUserGames = new QCheckBox(tr("Hide '&ignored user' games"));
     hideIgnoredUserGames->setChecked(gamesProxyModel->getHideIgnoredUserGames());
 
+    hideOldGames = new QCheckBox(tr("Hide games &older than 2 hours"));
+    hideOldGames->setChecked(gamesProxyModel->getHideOldGames());
+
     gameNameFilterEdit = new QLineEdit;
     gameNameFilterEdit->setText(gamesProxyModel->getGameNameFilter());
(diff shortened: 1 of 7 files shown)


knitknit:
@tooomm @ZeldaZach right now I have this set up to be a checkbox, to filter games that are >2 hours old. Do you think we should allow users to specify an age cutoff instead?

ZeldaZach:
Sorry for the delay in response! A 2 hour checkbox seems reasonable, but giving them the option to filter by X hours seems better to me. Play around with it :)

tooomm:
Nice work, this looks really helpful! :)

While a custom value would be even better, this is already a great addition to the filter options.

Let us know as soon as you're done with your changes (hit the `ready for review` button). 👍

ZeldaZach:
Would like to see this enabled by default

I'd also recommend checking out https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/gamesmodel.cpp#L26 to make it better match what we're trying to do (It will show 1hr 31min games as "2+ hrs" which feels backwards compared to the filter you're working on)

knitknit:
Thanks both! I got turned around somewhere when trying to merge in the X/Y filter changes, I'm going to go ahead and close this PR and start fresh. Will let you know as soon as the new one is ready for review!

PR #3965, opened 2020-04-22
skip all duplicate promos instead of just stars

### LINKED ISSUE #3964: [Ikoria] Wrong picture for many cards
Hello,

Many pictures downloaded with oracle for the “Ikoria” extension are not the default one.
There is even one in Japanese: Mysterious Egg
Many other cards are getting alternate art instead of the original one’s example: auspicious Starrix; Brokkos, Apex of Forever; cloudpiercer; glowstone recluse; ....
I have not check them all. Can you help us fixing this automatically instead of manually through the custom images?

Sincerely,

Danyfire

ebbit1q:
fix #3964
see #3904 which did this for all unique numbers, now this has been replaced with unique names because that's what we care about anyway and cards that have the isPromo field are added to it as well.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -201,7 +201,7 @@ QString OracleImporter::getStringPropertyFromMap(QVariantMap card, QString prope
     return card.contains(propertyName) ? card.value(propertyName).toString() : QString("");
 }
 
-int OracleImporter::importCardsFromSet(CardSetPtr currentSet, const QList<QVariant> &cardsList, bool skipSpecialNums)
+int OracleImporter::importCardsFromSet(CardSetPtr currentSet, const QList<QVariant> &cardsList, bool skipSpecialCards)
 {
     static const QMap<QString, QString> cardProperties{
         // mtgjson name => xml name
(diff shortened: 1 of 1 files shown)


ebbit1q:
Note this removes cards like godzilla entirely, so people that want these cards will have to use custom images instead. An alternative solution to this would be to automatically generate an additional set to hold these duplicate cards but the custom image folder is a better system for these people anyways as they only want some of these promos and not all of them.

ebbit1q:
https://travis-ci.org/github/Cockatrice/Cockatrice/builds/678126222 ci competed but it's not reflected here :/

ZeldaZach:
I'd like to see the cards split up on Cockatrice's end into separate internal sets. Removing godzillas entirely is a bleh thing. Will still merge if it addresses issues, but I'd like that fixed

ebbit1q:
do you want godzilla to be an entirely separate card instead of an alt art for it's original card?

ZeldaZach:
This goes along w/ the plan of changing how we load in specific cards. Should have a way for people to specify their art, even if local only. Maybe with a dropdown menu per-card saving in the cache?

ebbit1q:
Apparently there is no original zilortha card so godzilla doesn't get removed, this pr only removes promos if they are duplicates of another card in the same set.

PR #3974, opened 2020-04-26
Docker-Compose support for Windows

mpalmujoki:
The existing docker-compose.yml works for Linux environment, but Windows needs additional consideration.

- Supplied an docker-compose file that will work in Windows environment.
- Set up an suitable directory for mysql-storage.


ZeldaZach:
Appreciate the input!!

Issue #3985, opened 2020-04-30
Cutomzing Appearance - Load two different tablezone.jpg


Assassinsquid:
When it comes to customzing the appearance of Cockatrice, You can only load one image for each zone (tablezone, stackzone...). So as a result, you see each zone twice (for you and your opponent), and it is not very beautiful. Is it possible to load two different images for each zone (one for you and one for your opponent) ? It would be a very great add. Thanks.

fdipilla:
If the approach I'm using looks ok, I would like to make similar changes for all the zones and up to 4 players.

ZeldaZach:
Fixed in the linked issues

tooomm:
@fdipilla do you still want to apply your approach to the other zones as well?? That would be awesome! 🚀

fdipilla:
Heck yeah!

PR #3990, opened 2020-05-07
Draft on Cockatrice/Cockatrice#3985 about using a different backgroun…

fdipilla:
…d for each player

## Related Ticket(s)
- Draft on  #3985

Give the user the possibility of using multiple background images for each player

If the file named tablezone2.jpg exists it will be used as the Player 2 table background

## Screenshots
Screenshot from 2020-05-07 19-22-21

### DIFF
--- a/cockatrice/src/tablezone.cpp
+++ b/cockatrice/src/tablezone.cpp
@@ -55,9 +55,8 @@ bool TableZone::isInverted() const
 
 void TableZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*option*/, QWidget * /*widget*/)
 {
-    QBrush brush;
+    QBrush brush = themeManager->getTableBgBrush();
 
-    brush = themeManager->getTableBgBrush();
     // If the player is other than Player 1
     if (player->getId() > 0) {
         // The palyer's id starts with 0 so in order to get the correct image we need to add 1
--- a/cockatrice/src/tablezone.cpp
+++ b/cockatrice/src/tablezone.cpp
@@ -59,7 +59,7 @@ void TableZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*opti
 
     // If the player is other than Player 1
     if (player->getId() > 0) {
-        // The palyer's id starts with 0 so in order to get the correct image we need to add 1
+        // The player's id starts with 0 so in order to get the correct image we need to add 1
         brush = themeManager->getExtraTableBgBrush(QString::number(player->getId() + 1));
     }
     painter->fillRect(boundingRect(), brush);
--- a/cockatrice/src/tablezone.cpp
+++ b/cockatrice/src/tablezone.cpp
@@ -58,8 +58,10 @@ void TableZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*opti
     QBrush brush;
 
     brush = themeManager->getTableBgBrush();
-    if (player->getId() == 1) {
-        brush = themeManager->getTable2BgBrush();
+    // If the player is other than Player 1
+    if (player->getId() > 0) {
+        // The palyer's id starts with 0 so in order to get the correct image we need to add 1
+        brush = themeManager->getExtraTableBgBrush(QString::number(player->getId() + 1));
     }
     painter->fillRect(boundingRect(), brush);
 
--- a/cockatrice/src/tablezone.cpp
+++ b/cockatrice/src/tablezone.cpp
@@ -55,7 +55,13 @@ bool TableZone::isInverted() const
 
 void TableZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*option*/, QWidget * /*widget*/)
 {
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Is there a way we can expand this PR to allow for PlayerN? For EDH/Kingdoms type games, that go up to 6 players, would be nice to support custom fields for each player

fdipilla:
I think this one does the trick, although I'm not sure about my function/variable naming

cockatrice/src/tablezone.cpp
```diff
@@ -55,7 +55,15 @@ bool TableZone::isInverted() const
 
 void TableZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*option*/, QWidget * /*widget*/)
 {
-    painter->fillRect(boundingRect(), themeManager->getTableBgBrush());
+    QBrush brush;
+
+    brush = themeManager->getTableBgBrush();
```
ZeldaZach, line 60: I'd combine the initializer and definition

cockatrice/src/thememanager.cpp
```diff
@@ -15,6 +15,7 @@
 #define PLAYERZONE_BG_NAME "playerzone"
 #define STACKZONE_BG_NAME "stackzone"
 #define TABLEZONE_BG_NAME "tablezone"
+#define TABLEZONE2_BG_NAME "tablezone2"
```
ZeldaZach, line 18: unneeded

cockatrice/src/tablezone.cpp
```diff
@@ -55,7 +55,15 @@ bool TableZone::isInverted() const
...
+    // If the player is other than Player 1
+    if (player->getId() > 0) {
+        // The palyer's id starts with 0 so in order to get the correct image we need to add 1
+        brush = themeManager->getExtraTableBgBrush(QString::number(player->getId() + 1));
```
ZeldaZach, line 64: What happens if this misses? (read: What if the images aren't found?)

fdipilla: I've tested it and it just uses the default images

cockatrice/src/tablezone.cpp
```diff
@@ -55,7 +55,14 @@ bool TableZone::isInverted() const
 
 void TableZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*option*/, QWidget * /*widget*/)
 {
-    painter->fillRect(boundingRect(), themeManager->getTableBgBrush());
+    QBrush brush = themeManager->getTableBgBrush();
+
+    // If the player is other than Player 1
+    if (player->getId() > 0) {
+        // The palyer's id starts with 0 so in order to get the correct image we need to add 1
```
ZeldaZach, line 62: typo

tooomm:
This should probably be documented in the customizing/theming section in the wiki.

tooomm:
Do you want to make any further changes? Otherwise hit that `Ready for review` button @fdipilla! 👍 

The addition for the wiki documentation should go here: https://github.com/Cockatrice/Cockatrice/wiki/Themes
As you probably know best... can you add an explanation sentence or two there, please? At best with an example! The wiki is free to edit.

fdipilla:
Yup!

tooomm:
@fdipilla I assume “player1” is always yourself?

ZeldaZach:
player0 is yourself (index starts at 0)

tooomm:
Ok, I'll adjust that in the wiki.

`tablezone.jpg` alters **all** tablezones then, and the ones with the number (e.g. `tablezone0.jpg`) change the default tablezone image only for that player?

fdipilla:
> `tablezone.jpg` alters **all** tablezones then, and the ones with a number attached (e.g. `tablezone0.jpg`) change the default tablezone image only for that player?

Yes, with the exception of player0 (Player 1 for the user). In that case the code will always be looking for `tablezone.jpg`

But now that you mention it, maybe it would be nice to allow a the 'Player 1' to have a different background other than the default `tablezone.jpg`  then it will be something like:

tablezone.jpg | default, all players
tablezone1.jpg | Player 1
tablezone2.jpg | Player 2
tablezone3.jpg | Player 3

> So, `tablezone2.jpg` would only be visible in multiplayer games for the 3rd player (aka "Player 3" from a user point of view, one does not call himself "Player 0" I assume smile)?

That's why I added the `player->getId() + 1`, so tablezone2.jpg -> "Player 2", tablezone3.jpg -> "Player 3", etc

ebbit1q:
In other words the current implementation allows you to change the background for players 2+ by placing the file tablezone2.jpg or higher in the theme. Your own tablezone will always be the default and if it can't find a tablezone for other players they will also be the default.

ebbit1q:
Actually player with id 0 is not always yourself, only if you are the creator of the lobby.

tooomm:
>But now that you mention it, maybe it would be nice to allow a the 'Player 1' to have a different background other than the default tablezone.jpg 

Agreed.


Thanks for the nailing phrasing and important hint @ebbit1q!
Damn, that's more difficult than one might think haha.

PR #3995, opened 2020-05-16
Capitalize player counters

### LINKED ISSUE #3992: Capitalize player counters
Player counters should be `W`, `U`, `B`, `R`, `G` and `C` (See https://github.com/Cockatrice/Cockatrice/issues/3991), as well as `Life` and `Storm`:

counter_cap

Rocangus:
## Related Ticket(s)
- Fixes #3992

Player counters should appear capitalized, not all lower case as current.

Counter names will appear capitalized in the client. No server-side changes have been included, including changing the name of x to c.

### DIFF
--- a/cockatrice/src/abstractcounter.cpp
+++ b/cockatrice/src/abstractcounter.cpp
@@ -27,12 +27,14 @@ AbstractCounter::AbstractCounter(Player *_player,
       useNameForShortcut(_useNameForShortcut), hovered(false), aDec(nullptr), aInc(nullptr), dialogSemaphore(false),
       deleteAfterDialog(false), shownInCounterArea(_shownInCounterArea), game(_game)
 {
+    QString displayName = QString(_name);
+    displayName[0] = displayName[0].toTitleCase();
     setAcceptHoverEvents(true);
 
     shortcutActive = false;
 
     if (player->getLocalOrJudge()) {
-        menu = new TearOffMenu(name);
+        menu = new TearOffMenu(displayName);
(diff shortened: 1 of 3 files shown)


cockatrice/src/messagelogwidget.cpp
```diff
@@ -677,13 +677,15 @@ void MessageLogWidget::logSetCardCounter(Player *player, QString cardName, int c
 
 void MessageLogWidget::logSetCounter(Player *player, QString counterName, int value, int oldValue)
 {
-    if (counterName == "life") {
+    QString counterDisplayName = QString(counterName);
```
ZeldaZach, line 680: There’s no reason to add the display string here. If the “if statement” works, then construct the display name. You’re excessively creating strings (expensive) that may never get used.

cockatrice/src/abstractcounter.cpp
```diff
@@ -27,12 +27,14 @@ AbstractCounter::AbstractCounter(Player *_player,
       useNameForShortcut(_useNameForShortcut), hovered(false), aDec(nullptr), aInc(nullptr), dialogSemaphore(false),
       deleteAfterDialog(false), shownInCounterArea(_shownInCounterArea), game(_game)
 {
+    QString displayName = QString(_name);
```
ZeldaZach, line 30: Construct this as close to the display as possible. In this case, right before you use it, inside the if statement

Rocangus:
Creating new (draft) PR for new branch on my fork

Issue #3978, opened 2020-04-27
Improve "Move to bottom" message (and menu name)


tooomm:
**Description of New Feature**
Moving a card to the bottom of the library (when performing a London Mulligan for example) via right click on a card in your hand ("move to --> bottom of library in random order") results in the weird message "x puts y from their hand into their library z cards from the top".

It should simply say "x puts y on the bottom of their library". Our current message is quite confusing.

Ideally, the menu option would also be named
`Bottom of library` when one card is selected, and
`Bottom of library in random order` when several cards are selected.

leestran1995-daugherty:
I'd like to give this a look. No promises I can deliver on it anytime soon though, so if anyone else wants to jump in and work on it too I won't be offended :-)

tooomm:
Hi @leestran1995-daugherty. Welcome :)
Sure, feel free to ask any questions here, at gitter or in the discord.

leestran1995-daugherty:
Customizing the name of the menu option is trickier than it seems. We (you?) populate that menu for each card when the game is started, not when the cards are selected or when the user right-clicks, so with that current implementation we can't customize that message to the number of cards the user has selected.

PR #4006, opened 2020-05-20
fix message when moving cards to bottom of library

LeeTranMN:
## Related Ticket(s)
- Part of #3978 

When placing cards on the bottom of a library, Cockatrice currently gives a message along the lines of "player x places card y z cards from the top of their library" instead of just saying "bottom of their library"

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -326,8 +326,8 @@ void MessageLogWidget::logMoveCard(Player *player,
     } else if (targetZoneName == deckConstant()) {
         if (newX == -1) {
             finalStr = tr("%1 puts %2%3 into their library.");
-        } else if (newX == targetZone->getCards().size() - 1) {
-            finalStr = tr("%1 puts %2%3 on bottom of their library.");
+        } else if (newX == targetZone->getCards().size()) {
+            finalStr = tr("%1 puts %2%3 onto the bottom of their library.");
         } else if (newX == 0) {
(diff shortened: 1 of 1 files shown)


ebbit1q:
I'm not a native speaker but onto the bottom sounds wrong to me, I think on the bottom is more suitable.

cockatrice/src/messagelogwidget.cpp
```diff
@@ -326,8 +326,8 @@ void MessageLogWidget::logMoveCard(Player *player,
     } else if (targetZoneName == deckConstant()) {
         if (newX == -1) {
             finalStr = tr("%1 puts %2%3 into their library.");
-        } else if (newX == targetZone->getCards().size() - 1) {
-            finalStr = tr("%1 puts %2%3 on bottom of their library.");
+        } else if (newX == targetZone->getCards().size()) {
```
ebbit1q, line 329: I think just having a >= is appropriate here, while the game allows you to put a card a million cards from the top, this is the result anyway :)

tooomm:
@ZeldaZach can you check the wording and code here, please?

Sorry for not being that responsive at all with your submitted change @LeeTranMN. :(

cockatrice/src/messagelogwidget.cpp
ZeldaZach, line 329: ``
```suggestion
        } else if (newX >= targetZone->getCards().size()) {
```

tooomm:
@LeeTranMN thanks for your first contribution to this project. Check out upcoming beta builds to see it live!

You can always check out the discord if you want to chat.

PR #4009, opened 2020-05-20
Made user information window resizable

### LINKED ISSUE #2874: Make the user information window resizable
<b>OS:</b>
Windows 10

<b>Cockatrice version:</b>
Version 2.3.18-beta004 (2017-06-28)
__________
Could the user information window be made resizable? The avatar image inside should zoom with it.
I use a magic card as my avatar, and the text is hard to read with the current window size.

awlangham:
## Related Ticket(s)
- Fixes #2874

Makes the user information window resizable by the user.

- Removed the call to setFixedSize (which locked the widget's size as its sizeHint) in UserInfoBox.

## Screenshots
Before resizing:
image

After resizing (example):
image

### DIFF
--- a/cockatrice/src/userinfobox.cpp
+++ b/cockatrice/src/userinfobox.cpp
@@ -27,11 +27,11 @@ UserInfoBox::UserInfoBox(AbstractClient *_client, bool _editable, QWidget *paren
     nameFont.setPointSize(nameFont.pointSize() * 1.5);
     nameLabel.setFont(nameFont);
 
-    avatarLabel.setMaximumWidth(400);
-    avatarLabel.setMaximumHeight(200);
+    avatarLabel.setScaledContents(true);
+    avatarLabel.setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
 
     QGridLayout *mainLayout = new QGridLayout;
-    mainLayout->addWidget(&avatarLabel, 0, 0, 1, 3, Qt::AlignCenter);
+    mainLayout->addWidget(&avatarLabel, 0, 0, 1, 3);
     mainLayout->addWidget(&nameLabel, 1, 0, 1, 3);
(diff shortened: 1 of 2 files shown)


ebbit1q:
It'd be nice if the image would scale with it, currently all this does is allow the text to get more spaced out. This should be possible by setting the size policies.

awlangham:
Sure, I'll look into that.

tooomm:
Just a thought, I did not test this...
Could it be that the fixed size had a purpose? What happens if somebody has a huge player image? With this change, opening his profile may result in a very big window which we might want to prevent. Do we need a reasonable default size on opening?

awlangham:
I changed it to give the avatar an expanding size policy. However, this doesn't preserve the aspect ratio of the picture, since it's implemented as a QPixmap placed in a QLabel. From what I've found online it seems that to preserve the aspect ratio one must either subclass QLabel or reimplement resizeEvent to adjust the QPixmap's size when the QLabel's size is changed (currently it is handled with setScaledContents allowing the QLabel to adjust the QPixmap size).

ebbit1q:
It is preferable to have the image retain its aspect ratio.

awlangham:
My apologies for how long it's taken for me to follow up on this. From everything that I could find online, including other Qt users trying to do something similar, using a QLabel with a QPixmap on top of it does not have a simple, layout-based way to preserve the aspect ratio of the pixmap. I reimplemented the widget's resizeEvent to control the aspect ratio of the pixmap (the alternative seems to be to subclass QLabel to add this behavior). Please let me know if this would be an acceptable approach to solve this.

ebbit1q:
Works great! Thanks for coming back to this. I have formatted your pr.

ZeldaZach:
Thanks for your contribution to my project! We greatly appreciate it :)

PR #4015, opened 2020-05-26
Move card db update menu entry

tooomm:
`Update Client` and `Update Cards` are two separate processes.
Users often don't find them, or are not aware that there are two as they only realize one in the menus.

- Put them next to each other in the `Help` menu
- I reordered some lines to make them match the appearance in client

## Screenshots

helpmenu

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -1,6 +1,6 @@
 /***************************************************************************
- *   Copyright (C) 2008 by Max-Wilhelm Bruker   *
- *   brukie@gmx.net   *
+ *   Copyright (C) 2008 by Max-Wilhelm Bruker                              *
+ *   brukie@gmx.net                                                        *
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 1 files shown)


tooomm:
`Check for Card Updates...` option to run Oracle could also be placed in the `Card Database` menu.
People are more likely to expect and find it there and might look there more regular to spot it because they edit sets.

Both are better places than the current one in my opinion.
The `Cockatrice` menu is barely used after registering once and setting auto-connect.

ebbit1q:
You're making some assumptions there, I'd prefer this placed in its own menu called Updates with an option to open up release notes. That way noone wonders which menu to use for updating.

tooomm:
>I'd prefer this placed in its own menu called Updates...

Hmm, I understand where you are going. But generally such update actions are often placed in the `Help` or `About` menu for apps with a normal menu (e.g. Adobe applications). I can only tell for Windows.

Furthermore, the menu would only keep two options for now. As soon as we get automatic update checks, all kind of update processes would probably be combined to one action which lists all available updates. Also with additional information hopefully.

>...with an option to open up release notes.

Users can open the changelog/release notes before updating already:
rnotes

<br>

I think this is an improvement over the split up entries, without rethinking all menu.
With the new update routine and automated checks we can see what makes most sense then?

tooomm:
How do we proceed here? @ZeldaZach

There is no reason a user has to search in different places in order to use the most recent client and data.

I see two potential options for improvements for now:
- Group both in the `Cockatrice` menu (aka move client updates there)
- Group both in the `Help` menu as other apps have it there, too (aka move card updates there) <kbd>That's what this PR proposes</kbd>

Ebbit suggested a third, which I'm not a fan of as outlined above.

ZeldaZach:
My suggestion would be to move them all to the Help menu, which this PR solves. As such, I'm fine going ahead with this change.

ebbit1q:
I see it does some reordering as well, do these not affect the order in which the menus are displayed?

tooomm:
I did the reorder exactly for that reason - appearance in code matches now the appearance in client. :)
Positions are defined in a different spot, so one could argue that the actions should be alphabetical then. It did not match either pattern and looked random, so I decided to go with the "order them the way they appear in client".

Do you want me to change that? @ebbit1q 
Is it better to have them in e.g. alphabetical order when working with the code and adding new stuff for example?

ebbit1q:
Order they appear in client, they look random because they were most likely added over time.

tooomm:
As a positive side effect, updating cards and tokens is now uniform on all platforms:
https://github.com/Cockatrice/Cockatrice/wiki/Frequently-Asked-Questions-(FAQs)#how-do-i-update-my-card-database-to-have-the-latest-cards

PR #4017, opened 2020-06-01
make sure no //es are added to maintypes

### LINKED ISSUE #3972: Double click instants/sorceries from stack to graveyard
**Similar Requests**

After double clicking an instant or sorcery, it goes to the stack. Double clicking it again, it goes to play. It should probably go to the graveyard?

**Additional Context**

ebbit1q:
instead of only avoiding this on adventures they should just be avoided
on any card so the maintype instant // sorcery isn't created for example,
the normal typeline isn't affected by this
    
fixes #3972


ebbit1q:
All spells go to the stack, this spell did that too, this change means that it now goes to the graveyard after.

PR #4027, opened 2020-06-15
qt 5.15 compatibility

ebbit1q:
- fixes deprecation warnings and moves of enums in qt 5.15
- removed old, unused qsrand call
- refactored logic for changing user image 

fixes: #4028

### DIFF
--- a/cockatrice/src/filtertreemodel.cpp
+++ b/cockatrice/src/filtertreemodel.cpp
@@ -4,6 +4,7 @@
 #include "filtertree.h"
 
 #include <QFont>
+#include <qnamespace.h>
 
 FilterTreeModel::FilterTreeModel(QObject *parent) : QAbstractItemModel(parent)
 {
@@ -174,11 +175,11 @@ Qt::ItemFlags FilterTreeModel::flags(const QModelIndex &index) const
     Qt::ItemFlags result;
 
     if (!index.isValid())
-        return 0;
+        return Qt::NoItemFlags;
 
     node = indexToNode(index);
     if (node == NULL)
-        return 0;
+        return Qt::NoItemFlags;
 
     result = Qt::ItemIsEnabled;
     if (node == fTree)
--- a/cockatrice/src/filtertreemodel.cpp
+++ b/cockatrice/src/filtertreemodel.cpp
@@ -4,7 +4,6 @@
 #include "filtertree.h"
 
 #include <QFont>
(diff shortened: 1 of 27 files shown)


servatrice/src/server_logger.cpp
```diff
@@ -57,7 +57,7 @@ void ServerLogger::logMessage(QString message, void *caller)
     // filter out all log entries based on values in configuration file
     bool shouldWeWriteLog = settingsCache->value("server/writelog", 1).toBool();
     QString logFilters = settingsCache->value("server/logfilters").toString();
-    QStringList listlogFilters = logFilters.split(",", QString::SkipEmptyParts);
+    QStringList listlogFilters = logFilters.split(",", Qt::SkipEmptyParts);
```
ctrlaltca, line 60: Qt::SkipEmptyParts was introduced in Qt 5.14, I'm afraid we'll need an #ifdef here

ebbit1q: I just noticed a lot more warnings as well :(

ebbit1q:
I fixed all of the easy ones, there are like 3 left that require some more consideration.

cockatrice/src/filtertreemodel.cpp
```diff
@@ -4,6 +4,7 @@
 #include "filtertree.h"
 
 #include <QFont>
+#include <qnamespace.h>
```
ctrlaltca, line 7: That looks a private header, maybe `#include <Qt>` will work?

ebbit1q: ah, yeah, that seems to have been autogenerated.

ebbit1q:
I removed the qsrand call, I'm pretty certain that is no longer needed since we use sfmt.

ebbit1q:
Also the refactored logic for changing user image should get some more testing.

ctrlaltca:
> I removed the qsrand call, I'm pretty certain that is no longer needed since we use sfmt.

I agree, qsrand() is used to seed the generator used by qrand(), but I can't find any usage of qrand() in the code

Btw, there are still some `#include <qnamespace.h>` in the patch

ebbit1q:
I err, don't know how those got there.

ebbit1q:
passed but doesn't update >.> https://travis-ci.org/github/Cockatrice/Cockatrice/builds/699346129

PR #4033, opened 2020-06-21
[WIP] Implement basic "Load deck from URL" feature.

### LINKED ISSUE #4032: Load Remote Deck from URL (in Deck Editor)
I would like the ability to Load a Remote Deck from a URL instead of just my filesystem, or the remote server. It should be implemented to allow any format Cockatrice already supports, defaulting to XML.

**Additional Context**
This could allow loading decks from a shared library (github/dropbox/drive/...) or from an API that returns a deck in XML, text or another supported format (based on Content-Type headers).

I'm looking forward to the upcoming [JumpStart](https://mtg.gamepedia.com/Jumpstart) and was thinking of creating a small API that would return a Deck in XML that is a random set of 2 or more of the 121 possible packs. Being able to "Load Remote Deck from URL" would be a an easy and quick way to play JumpStart in Cockatrice.

**Similar Requests**
I searched around and didn't seen any existing issues related or similar to this.

Omicron7:
This is a very basic first pass and pulls a lot from the clipboard and remote-deck loading code. It still needs some error reporting. It's been quite a while since I've done any C/C++, but wanted to give this a shot. Any feedback is appreciated.

## Related Ticket(s)
- Fixes #4032 

Load a deck from a URL without building the deck in the Deck Editor first. This would allow using APIs for random decks, decks from github/dropbox/pastebin, or possibly draft simulator providing URL of final picks.

- Added the ability to load a deck from a URL.

## Screenshots
Load deck from URL...

Load deck from URL dialog


Omicron7:
You can use tappedout.net to test. Just add `?fmt=txt` to any deck URL.
Example: `https://tappedout.net/mtg-decks/some-kind-of-tron-1/?fmt=txt`

tooomm:
I think adding some kind of explanation to the URL dialog itself can be helpful here.
Users probably don't know what can be entered and which input is expected.

ZeldaZach:
Hey @Omicron7 thanks so much for your contribution! I took a look at what you've presented, and have some suggestions on improvements.

My first idea is to have the system be more fault tolerant. You posted how we need to tell the external sources to present in a text way. Why not add several common sub-classes so we can parse on demand? Ex: TappedOut, Scryfall, DeckStats. That way, the user can post any type of URL from these services and have them auto loaded in. Better UX imo for this, even if it's a bit more work on our end.

Next, we should explain in more detail what formats we support. Right now, it's just a box asking for a URL. That could lead to a malicious operation if we're not careful. We should include in the dialog that we support X, Y, and Z natively, but can also support any text dump from a site. An idea could also be to restrict it to only certain service URLs, but that's probably not necessary.

Finally, maybe it's worth cleaning up how users can load a deck. Maybe change it from 3 distinct buttons to a dropdown of some kind? Just so we can expand more in the future without having UI bloat.

Again, thanks for helping! I hope we can get this change in soon :)

ZeldaZach:
Ebbit and I spoke about this, and agree current clipboard loader should suit these needs. We'd also have to maintain a long list of sites that handle decks. We're inclined to close for now, but if situations change please feel free to re-file!

In any case, I appreciate your PR and hope you choose to file new PRs in the future :)

PR #4038, opened 2020-06-29
Add exclude term to search hints

tooomm:
People might search for the more common (?) `Exclude` term, but can't find it within the explanations/hints.

- add `Exlude` next to `Inverse`

### DIFF
--- a/cockatrice/resources/help/search.md
+++ b/cockatrice/resources/help/search.md
@@ -49,7 +49,7 @@ The search bar recognizes a set of special commands similar to some other card d
 <dd>[e:lea,leb](#e:lea,leb) <small>(Cards that appear in Alpha or Beta)</small></dd>
 <dd><a href="#e:lea,leb -(e:lea e:leb)">e:lea,leb -(e:lea e:leb)</a> <small>(Cards that appear in Alpha or Beta but not in both editions)</small></dd>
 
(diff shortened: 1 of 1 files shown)


ebbit1q:
I'd prefer one word to be there, adding all the synonyms in the dictionary doesn't make it easier to read, compare scryfall's guide https://scryfall.com/docs/syntax#negating

tooomm:
I like how everything can be negated at scryfall, their explanations are also good.
But does it work the same for us?

Having a single hint that all commands can get negated/excluded would be easy to understand I guess. 👍

Maybe it's just me, but I prefer the terms exclude and negate over inverse. I like negate even more.
Feels more simple, straight forward and better known. What do other non-native speaker think?
What do other apps or documenations use?

<br>

*side note: looks like scryfall uses "!" for exact name matches @ebbit1q*

ebbit1q:
Then put negate there, it doesn't matter that much.

ZeldaZach:
I'm not seeing much of a purpose to this change, @tooomm is it critical?

tooomm:
I opted to use `Negate`, along the lines what scryfall uses in their docs as stated by ebbit.

ebbit1q (approved):
either works, this is fine I guess

PR #4046, opened 2020-07-14
Change method of opening directories to be the same for all oses, including linux

ebbit1q:
this uses QDesktopServices to open the url "file://[location]"
by default this is
"file://$HOME/.local/share/Cockatrice/Cockatrice/pics/CUSTOM"

any distro that has a file browser should have an accompanying mime type
specifying the file handler for the file:// protocol using the
inode/directory mime type

see https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html

if a user were to have removed their mime database this will not work and
it will fail with nothing but a log message, this would be rare and not
worth checking in my opinion

fixes a ticket that hopefully exists but I can't find

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -46,6 +46,7 @@
 #include "version_string.h"
 #include "window_sets.h"
 
+#include <QDesktopServices>
 #include <QAction>
 #include <QApplication>
 #include <QCloseEvent>
@@ -752,10 +753,8 @@ void MainWindow::createMenus()
     dbMenu->addAction(aManageSets);
     dbMenu->addAction(aEditTokens);
     dbMenu->addSeparator();
-#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
     dbMenu->addAction(aOpenCustomFolder);
     dbMenu->addAction(aOpenCustomsetsFolder);
-#endif
     dbMenu->addAction(aAddCustomSet);
 
     helpMenu = menuBar()->addMenu(QString());
@@ -1206,6 +1205,8 @@ void MainWindow::actOpenCustomFolder()
     QStringList args;
     args << QDir::toNativeSeparators(dir);
     QProcess::startDetached("explorer", args);
+#else // LINUX
+    QDesktopServices::openUrl(QUrl::fromLocalFile(dir));
 #endif
 }
 
@@ -1225,6 +1226,8 @@ void MainWindow::actOpenCustomsetsFolder()
     QStringList args;
     args << QDir::toNativeSeparators(dir);
     QProcess::startDetached("explorer", args);
+#else // LINUX
+    QDesktopServices::openUrl(QUrl::fromLocalFile(dir));
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


ctrlaltca:
Nice, this could probably also replace out current implementations for windows and macos.
Looks like internally it's using [ShellExecuteA](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecutea) on Windows and [NSWorkspace.shared.openFile](https://developer.apple.com/documentation/appkit/nsworkspace) on macOS

ebbit1q:
It totally could, but I don't have those systems and I read that file:// protocol is possibly handled by iexplorer on windows.

ctrlaltca:
I just tested both windows and macOS, a single call to `QDesktopServices::openUrl(QUrl::fromLocalFile(dir));` works fine on both

ebbit1q:
How about different versions of windows? How many are we supposed to support anyway?
I'd say it's safe to assume this works on any version of macos, but windows I'm not so sure about.

ctrlaltca:
I tested 7 and 10.

tooomm:
Win 7 is the oldest we support currently 👍

ebbit1q:
I made #4047 which allowed me to test some edge cases, the current method seems to be all that is needed because it will always be a valid directory or go back to default.

ebbit1q:
why does it not show my recent commit? https://github.com/ebbit1q/Cockatrice/commits/linux_open_dir

ctrlaltca:
That's.. funny, I guess?
I had to retarget the PR to another branch in order for Github to see the third commit; I then retargeted master.

ebbit1q:
What is this crap
```
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: 2 of 2 required status checks are expected. At least 1 approving review is required by reviewers with write access.
To https://github.com/Cockatrice/Cockatrice
 ! [remote rejected]   master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/Cockatrice/Cockatrice'
```

ebbit1q:
ci didn't trigger... github is broken burn it down!

tooomm:
Both of your recent commits show a "This commit does not belong to any branch on this repository." message. 🤔

ctrlaltca:
Looks like they are experiencing problems: https://twitter.com/githubstatus/status/1283426403782995974

ebbit1q:
Fine, I guess we'll wait.

ZeldaZach:
When GH status reports all green I'll see about re-triggering manually

ebbit1q:
https://www.githubstatus.com/ seems like the fire is gone now.

ctrlaltca:
Builds restarted

ebbit1q:
https://travis-ci.org/github/Cockatrice/Cockatrice/pull_requests travis did not trigger.

ctrlaltca:
I relaunched https://travis-ci.org/github/Cockatrice/Cockatrice/builds/708338451 , it completed fine but I guess that because of yesterday problems the status is not reported correctly. I'll just merge to avoid having to deal with it.

tooomm:
See the new build triggerd on master with the PR merge: https://travis-ci.org/github/Cockatrice/Cockatrice/builds/708666307

ebbit1q:
note: I noticed that even though if the user selects a location that is not a directory it'll cause the default directory to be used, it's even created if it doesn't exist. It is possible to replace the default location with something else. So if a user were boneheaded enough to replace pics/CUSTOM with say an image, the application will open that image instead of the file browser. This is not necessarily a problem, just something to be aware of.

PR #4047, opened 2020-07-15
make custom sets directory configurable

ebbit1q:
while testing #4046 I noticed how this isn't even a location that can be changed

fixes a ticket that I'm reasonably certain doesn't exist

screenshot:
image


ebbit1q:
Note that the custom image location will always be `[pics directory]/CUSTOM`, it's better that way and the pics directory is configurable.

cockatrice/src/dlg_settings.cpp
```diff
@@ -248,6 +268,7 @@ void GeneralSettingsPage::retranslateUi()
     replaysPathLabel.setText(tr("Replays directory:"));
     picsPathLabel.setText(tr("Pictures directory:"));
     cardDatabasePathLabel.setText(tr("Card database:"));
+    customCardDatabasePathLabel.setText(tr("Custom database directory:"));
```
tooomm, line 271: Can you swap ordering of the new option with the card database one?
That way we have directories on top, followed by file locations. While databases are still grouped together.

Also, we use the plural form for all others directories.

ebbit1q: I considered it ( I had it below replays first ) but I moved it down here because the main card database is more important, I'd consider moving it below tokens.
I made it singular because a database of databases is still a database, and with this now being the longest entry the place for long path names has been diminished for all entries :^)

PR #4052, opened 2020-07-25
Added horizontal layout and stretch for player icon

### LINKED ISSUE #4051: Issues with player information widget after resizable change
<b>System Information:</b>
Client Version: 2.7.4-custom(880ed8c) (2020-07-23)
Build Architecture: 64-bit
Qt Version: 5.12.6
System Locale: de_DE
Install Mode: Standard

  

#4009 introduced a problem to the Player Information widget.
The values for coutnry and account age are right aligned to the player picture. That way the reference to their label is lost and everything looks off.
You realize at best when there is a rectangular avatar and you increase the width further.

user-info

Those values should not move on window size changes at all.

CC: @awlangham

awlangham:
## Related Ticket(s)
- Fixes #4051 

Previous fix for #2874 via #4009 (making user information box resizable) had effect of separating the user info columns below the player icon.

- Added a horizontal layout in which there are 2 spacers which keep the player icon centered (while still allowing an expanding size policy)
- Changed the player icon's cell in the grid layout to occupy all three columns instead of just one; changed the stretch factor to be on the second column (as it was before #4009); added center alignment on avatar QLabel to center its QPixmap

## Screenshots
image

### DIFF
--- a/cockatrice/src/userinfobox.cpp
+++ b/cockatrice/src/userinfobox.cpp
@@ -48,7 +48,7 @@ UserInfoBox::UserInfoBox(AbstractClient *_client, bool _editable, QWidget *paren
     mainLayout->addWidget(&userLevelLabel1, 5, 0, 1, 1);
     mainLayout->addWidget(&userLevelLabel2, 5, 1, 1, 1);
     mainLayout->addWidget(&userLevelLabel3, 5, 2, 1, 1);
-    mainLayout->addWidget(&accountAgeLebel1, 6, 0, 1, 1);
+    mainLayout->addWidget(&accountAgeLabel1, 6, 0, 1, 1);
     mainLayout->addWidget(&accountAgeLabel2, 6, 2, 1, 1);
     mainLayout->setColumnStretch(2, 10);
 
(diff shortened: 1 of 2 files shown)


awlangham:
Hmm. I'm not sure why the AppVeyor build is failing. In messages it says "Build execution time has reached the maximum allowed time for your plan (60 minutes)."

ebbit1q:
That probably means something happened with our dependencies for windows builds, if it's not trivial we'll make a ticket for it.

ctrlaltca:
Build restarted, I hope it was a temporary problem

ebbit1q:
It looks like it wanted to update protobuf and recompile it and then errored out on some mismatched keys?

ebbit1q:
isn't there a binary for protobuf so we don't have to compile it with mingw?

ebbit1q:
The new build has the same problems... Did they change something at vcpkg?

ctrlaltca:
> isn't there a binary for protobuf so we don't have to compile it with mingw?

The problem with binary packges is that they will probably require a different vcredist version, so they will fail to work on pcs where those are not already installed.

PR #4055, opened 2020-07-30
Fix unresolved symbols when link tests to system libgtest-dev

olegshtch:

I've tried to build and test Cockatrice on Debian with `libgtest-dev 1.10.0-3` installed and run into unresolved `pthread` symbols errors.

- Adds `Threads::Threads` target to tests.
- Replace CMAKE_THREAD_LIBS_INIT with Threads::Threads in the server.


ZeldaZach:
Thanks for your contribution!!

Issue #4000, opened 2020-05-17
database tests fail on non debug builds


ebbit1q:
Building without the flag `-DCMAKE_BUILD_TYPE=Debug` but with `-DTEST=1` results in the tests crashing when run.

Minimal reproduction:
* Compile:
```
git pull https://github.com/Cockatrice/Cockatrice
cd Cockatrice
mkdir build
cd build
cmake .. -DTEST=1
make
```
* Run test:
```
./tests/carddatabase/carddatabase_test
```
* Output:
```
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from CardDatabaseTest
[ RUN      ] CardDatabaseTest.LoadXml
CardDatabase::loadCardDatabases start
Segmentation fault (core dumped)
```
* Traceback:
```
#0  0x000056133b549bb7 in CardDatabase::loadCardDatabases() ()
#1  0x000056133b541b58 in ?? ()
#2  0x00007f936853088b in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
   from /usr/lib/libgtest.so.1.8.1
#3  0x00007f9368526b2c in testing::Test::Run() () from /usr/lib/libgtest.so.1.8.1
#4  0x00007f9368526c89 in testing::TestInfo::Run() () from /usr/lib/libgtest.so.1.8.1
#5  0x00007f9368526d67 in testing::TestCase::Run() () from /usr/lib/libgtest.so.1.8.1
#6  0x00007f936852729d in testing::internal::UnitTestImpl::RunAllTests() () from /usr/lib/libgtest.so.1.8.1
#7  0x00007f9368530dfb in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) () from /usr/lib/libgtest.so.1.8.1
#8  0x00007f93685273cf in testing::UnitTest::Run() () from /usr/lib/libgtest.so.1.8.1
#9  0x000056133b53f7df in main ()
```

This issue directly causes https://github.com/Cockatrice/Cockatrice/issues/3639

ebbit1q:
Backtrace with the flag `-DCMAKE_BUILD_TYPE=RelWithDebInfo`:
```
[...]

PR #4056, opened 2020-07-30
Run tests on WIndows CI

olegshtch:
- Adds tests to Appveyor
- Synchronizes mock headers in the `dbconverter` and tests.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -45,7 +45,7 @@ install:
     - .\bootstrap-vcpkg.bat
     - cd %APPVEYOR_BUILD_FOLDER%
     - vcpkg remove --outdated --recurse
-    - vcpkg install openssl protobuf liblzma zlib --triplet %vcpkg_arch%-windows
+    - vcpkg install openssl protobuf liblzma zlib gtest --triplet %vcpkg_arch%-windows
 
 services:
     - mysql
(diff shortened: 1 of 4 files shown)


ebbit1q:
This is very useful, tests would probably behave differently on windows. Consider #4000 you'll most likely have to use the debug build type which'll give other issues :(

olegshtch:
I suppose MSVS doesn't accept mocks as the same symbols.

ebbit1q:
Is anyone even able to compile the tests locally on like a normal windows vm?

olegshtch:
Tests are failed because required libraries aren't in `PATH` but link issues are fixed and test results are published.

olegshtch:
@ebbit1q Looks like Windows also affected by #4000

ebbit1q:
I expected as much, would we be able to use the debug build type if it fixes the problem?

ebbit1q:
I see there is now 4 builds, do the release builds still run tests or only the debug ones?

tooomm:
>It doesn't fail build although.

@olegshtch Is that a general appveyor thing, or a setting?

<br>

Having all 4 builds run every time doubles the already long appveyor build time again and leads to very late results...

For travis we have "conditional builds" setup:
PR's build debug only, master branch builds debug+release and tagged commits build only release.
Maybe something similar is possible here, too?

ebbit1q:
on travis all the pr builds are debug, only the ubuntu 18.04 one is tests but it only adds like 3 seconds or so.

ebbit1q:
this is pretty much what I expected for debug builds on windows:
```
    4846 Warning(s)
    18 Error(s)
```

olegshtch:
> this is pretty much what I expected for debug builds on windows:

@ebbit1q Looks like mocks doesn't work for debug builds as even `dbconverter` is broken. I'll remove debug configuration for now.

> Maybe something similar is possible here, too?

Debug is hadrly broken here so I think it's too early to think about it. I want to try to parallelize build.

olegshtch:
It looks like in Debug build `mocks.obj` have own `QString SettingsCache::getCardDatabasePath() const` from `mocks.cpp` and `carddatabse.obj` have not inlined `QString SettingsCache::getCardDatabasePath() const` from `settingscache.h`. In the Release build second is inlined and doesn't lead to link conflict.

olegshtch:
#4000 was fixed so tests are passed now.

PR #4057, opened 2020-07-31
Enable parallel compilation.

olegshtch:

Appveyor is slow.

- Enables options to parallel compilation and building.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -57,8 +57,8 @@ build_script:
         $vcpkgbindir = "C:\Tools\vcpkg\installed\$env:vcpkg_arch-windows\bin"
         $mysqldll = "c:\Program Files\MySQL\MySQL Server 5.7\lib\libmysql.dll"
         cmake --version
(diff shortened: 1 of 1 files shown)


tooomm:
That seems to speed up the clock! :)

Old PR without the /MP flag took 10min longer to build: https://ci.appveyor.com/project/ZeldaZach/cockatrice/builds/34110709/job/u0o3i52mnb4gfm31

ebbit1q:
I always wondered if we could enable multithreading on those. it's faster than macos now https://travis-ci.org/github/Cockatrice/Cockatrice/builds/713817908

tooomm:
@olegshtch There is also https://github.com/Cockatrice/Cockatrice/issues/3059 slowing AppVeyor down... what's your opinion there? We never managed to make it work.

olegshtch:
@tooomm I've never seen any solution for MSVC caching.

tooomm:
@olegshtch Any idea why compilation time is back up to 22min again (See [here](https://ci.appveyor.com/project/ZeldaZach/cockatrice-k884w/builds/35463905)), even more than before this PR?!
It used to go down by 50% to 11min with your change! 😭

ebbit1q:
it's hard to say what's taking so long, the logs aren't timestamped like travis

tooomm:
I'm not aware of any changes to the build environment at appveyor or that they cut down resources for free projects etc.

PR #4058, opened 2020-08-01
Refactor SettingCache singleton access

olegshtch:
Allows to link mocked accessor instead of real one.

## Related Ticket(s)
- Part of solution for #4000

Trying to refactor access to singleton to be lesser depend on non-inlining mocked functions. It's still doesn't fix issue as carddatabase expects object of real SettingsCache memory layout.

- Uses Q_GLOBAL_STATIC to declare singleton.
- Uses accessor function to get instance instead of pointer.


ebbit1q:
Seems good :+1: just check the linter script and some of the codacy tips are right too.

tooomm:
For the record, #4063 superseded this?

olegshtch:
@tooomm Yes

PR #4059, opened 2020-08-03
travis: update macos 10.15 images

tooomm:
- use most recent (non-beta) Catalina (10.15) image available on travis
- high sierra is still on the latest, no changes needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,7 +163,7 @@ jobs:
   - name: macOS Catalina (Debug)
     if: tag IS NOT present
     os: osx
-    osx_image: xcode12
+    osx_image: xcode11.6
     cache:
     - ccache
     - directories:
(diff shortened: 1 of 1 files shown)


ctrlaltca:
You could even switch to catalina, it should support all the macs that mojave supports.

tooomm:
Will check, expected it to cut older macs off.

ZeldaZach:
I'm fine with the bump, as we don't lose anything

ebbit1q:
Which versions of macos are supported by this build?

tooomm:
With this PR we build on
- High Sierra: https://github.com/Cockatrice/Cockatrice/blob/master/.travis.yml#L120 
- Catalina: https://github.com/Cockatrice/Cockatrice/blob/master/.travis.yml#L162

So for releases, the deployed dmg is unchanged and still supports 10.13+ (should still run on 10.15, but without Dark Mode support).
The zipped dmg is now 10.15+ (instead 10.14+).

ebbit1q:
So we dropped dark mode support from 10.14?

ZeldaZach:
All machines that can run 10.14 _can also_ run 10.15, so they'd just have to upgrade

ebbit1q:
Yes, but they should know that they need to upgrade as well.

PR #4064, opened 2020-08-14
Update test card database to v4

olegshtch:
- Test card database will use v4 format


ebbit1q:
Backwards compatibility with v3 should still be tested, maybe keep the customsets one as is.

PR #4077, opened 2020-08-24
Remove release build for former Ubuntu 18.04 LTS

tooomm:
- Remove release build for 18.04 LTS, only keep 20.04 LTS
- Add hint and link why we have to keep 18.04 for debug builds for now (Qt bug)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,18 +30,6 @@ jobs:
     before_install: source .ci/travis-docker.sh --build
     script: RUN --server --debug --test
 
-  - name: Ubuntu Bionic (Release)
-    if: (branch = master AND NOT type = pull_request) OR tag IS present
-    os: linux
-    services: docker
-    language: minimal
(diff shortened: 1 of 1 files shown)


ebbit1q:
18.04 is fully supported still, even 16.04 is.

tooomm:
>18.04 is fully supported still, even 16.04 is.

That's what long term support releases are about. Hence we only offer downloads for them.

We never offered binaries for all available or possible variants, or several versions in support.
It was only the most recent LTS release. I think that's reasonable.

This change will only have an impact starting with our next release, considering that will most likely be in ~6 months I do not see any issues. 20.04 isn't super new either.

ebbit1q:
I don't see a reason to remove them still, they work fine and build quickly.

tooomm:
We can leave this open until we are more comfortable with this change...

ebbit1q:
closing in favor of #4164

PR #4089, opened 2020-09-07
Move "Games shown" text to top, to match user count list.

### LINKED ISSUE #3068: Add text hint that filter are applied to the game list

knitknit:
Fixes last part of #3068 that isn't already handled by #4088.

## Related Ticket(s)
- Fixes parts of #3068 

#3068 had a number of requests, including moving the "Games shown: X/Y" label to the title to match user count, rather than being next to the "Clear filters" button. This PR fixes that part of the request, and #4088 handles the rest of the requests in #3068 .

- "Games shown: X/Y" is at the top of the game selector instead of next to the "Clear filters" button.

## Screenshots
Screen Shot 2020-09-06 at 8 40 54 PM

### DIFF
--- a/cockatrice/src/gameselector.cpp
+++ b/cockatrice/src/gameselector.cpp
@@ -36,7 +36,6 @@ GameSelector::GameSelector(AbstractClient *_client,
 {
     gameListView = new QTreeView;
     gameListModel = new GamesModel(_rooms, _gameTypes, this);
-    filteredGamesLabel = new QLabel;
     if (showFilters) {
         gameListProxyModel = new GamesProxyModel(this, tabSupervisor);
         gameListProxyModel->setSourceModel(gameListModel);
@@ -90,7 +89,6 @@ GameSelector::GameSelector(AbstractClient *_client,
     if (showFilters) {
         buttonLayout->addWidget(filterButton);
         buttonLayout->addWidget(clearFilterButton);
-        buttonLayout->addWidget(filteredGamesLabel);
     }
     buttonLayout->addStretch();
     if (room)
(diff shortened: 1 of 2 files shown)


cockatrice/src/gameselector.cpp
```diff
@@ -296,11 +295,13 @@ void GameSelector::actSelectedGameChanged(const QModelIndex &current, const QMod
     joinButton->setEnabled(game.player_count() < game.max_players() || overrideRestrictions);
 }
 
-void GameSelector::setFilteredGamesLabel()
+void GameSelector::updateTitle()
 {
     if (showFilters) {
         const int totalGames = gameListModel->rowCount();
         const int shownGames = totalGames - gameListProxyModel->getNumFilteredGames();
-        filteredGamesLabel->setText(tr("Games shown: %1 / %2").arg(shownGames).arg(totalGames));
```
ebbit1q, line 304: Doesn't this mean the filtered games label will never contain text and can thus be safely removed?

knitknit: Yup! Done.

ebbit1q (approved):
:ok_hand:

tooomm (approved):
Happy to see you back with further improvements to the old project.

PR #4090, opened 2020-09-08
Allow Logs/Admin to be closed/opened

ZeldaZach:
This PR allows moderators and administrators to close/reopen the Logs and Administration tabs. By default, the Logs tab will show for mods & admins, whereas the Administration tab will only show for admins.

A moderator or administrator can press SHIFT + CONTROL + Click on the Cockatrice/Actions menu to reveal a hidden menu to reopen these tabs. If a normal users were to attempt at accessing the service tools, they would be unable to view any content, as the data is returned as a server sided check on the user account.

### DIFF
--- a/cockatrice/src/tab_admin.cpp
+++ b/cockatrice/src/tab_admin.cpp
@@ -91,6 +91,11 @@ TabAdmin::TabAdmin(TabSupervisor *_tabSupervisor, AbstractClient *_client, bool 
     actUnlock();
 }
 
+void TabAdmin::closeRequest()
+{
+    deleteLater();
+}
+
 void TabAdmin::retranslateUi()
 {
     updateServerMessageButton->setText(tr("Update server &message"));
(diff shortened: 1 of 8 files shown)


ebbit1q:
I don't like this ctrl shift hidden feature, either just make it a plain checkbox in settings or have it simply check when connecting if it needs to add these actions.

ZeldaZach:
@ebbit1q The menu doesn't take into account user accounts at all, and it doesn't have an easily modifyable menu. If you'd like to help bring that functionality in, I'd be down. This was the best I could come up with at the time of development.

ebbit1q:
as I suggested earlier you should instead simply hide the tabs instead of deleting them to save people from having to relog or similar.

ebbit1q:
I'd like this implemented in a simpler way that wouldn't introduce weird issues for edge cases, consider that this isn't something the average user sees in the first place. something that simply hides the tab would solve the issue just fine and doesn't have to deal with destroying and reopening it.

tooomm:
Can the source branch be deleted?

ebbit1q:
sure, github allows you to restore them anyway

PR #4091, opened 2020-09-08
Judges can talk in games

### LINKED ISSUE #4078: Judge role can't override no spectator chat
2.7.5

Judges should have the power to override the no spectator chat, while in judge mode. Similar to moderator/administrator access.

ZeldaZach:
Server & Client sided change. Will require a version bump to deploy officially.

Yes, the if statements are ugly. I left a comment block to indicate what's actually going on.

## Related Ticket(s)
- Fixes #4078

### DIFF
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -641,11 +641,6 @@ bool TabGame::isSpectator()
     return spectator;
 }
 
-bool TabGame::isJudge()
-{
-    return judge;
-}
-
 void TabGame::actGameInfo()
 {
     DlgCreateGame dlg(gameInfo, roomGameTypes, this);
@@ -1824,7 +1819,7 @@ void TabGame::createMessageDock(bool bReplay)
          * (b) the spectator is a moderator/administrator
          * (c) the spectator is a judge
          */
-        if (!(isSpectator() && (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || isJudge()))) {
+        if (!(spectator && (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || judge))) {
             sayLabel->hide();
             sayEdit->hide();
         }
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -1819,7 +1819,15 @@ void TabGame::createMessageDock(bool bReplay)
          * (b) the spectator is a moderator/administrator
          * (c) the spectator is a judge
          */
-        if (!(spectator && (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || judge))) {
+        bool canChat = true;
+        if (spectator) {
+            if (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || judge) {
+                canChat = true;
+            } else {
+                canChat = false;
+            }
+        }
+        if (!canChat) {
             sayLabel->hide();
             sayEdit->hide();
         }
--- a/cockatrice/src/tab_game.cpp
+++ b/cockatrice/src/tab_game.cpp
@@ -1814,14 +1814,17 @@ void TabGame::createMessageDock(bool bReplay)
         sayEdit->setCompleter(completer);
         actCompleterChanged();
 
-        /* Spectators can only talk if:
-         * (a) the game creator allows it
-         * (b) the spectator is a moderator/administrator
-         * (c) the spectator is a judge
-         */
-        if (spectator && !gameInfo.spectators_can_chat() && tabSupervisor->getAdminLocked() && !judge) {
-            sayLabel->hide();
-            sayEdit->hide();
+        if (spectator) {
+            /* Spectators can only talk if:
+             * (a) the game creator allows it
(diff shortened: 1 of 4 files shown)


cockatrice/src/tab_game.h
```diff
@@ -117,7 +117,7 @@ class TabGame : public Tab
 {
     Q_OBJECT
 private:
-    QTimer *gameTimer;
+    QTimer *gameTimer{};
```
ebbit1q, line 120: why add default initializers to all these?

ZeldaZach: Clangify did that, as it said some variables were missed. I'll revert this

cockatrice/src/tab_game.cpp
```diff
@@ -1814,10 +1819,16 @@ void TabGame::createMessageDock(bool bReplay)
         sayEdit->setCompleter(completer);
         actCompleterChanged();
 
-        if (spectator && !gameInfo.spectators_can_chat() && tabSupervisor->getAdminLocked()) {
+        /* Spectators can only talk if:
+         * (a) the game creator allows it
+         * (b) the spectator is a moderator/administrator
+         * (c) the spectator is a judge
+         */
+        if (!(isSpectator() && (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || isJudge()))) {
```
ebbit1q, line 1827: is there a specific reason to use (and add) these public functions while they're simply available as private members of this object?

ZeldaZach: Not really, I thought it was a bit cleaner. I can change that if you'd like

cockatrice/src/tab_game.cpp
```diff
@@ -1814,10 +1814,16 @@ void TabGame::createMessageDock(bool bReplay)
         sayEdit->setCompleter(completer);
         actCompleterChanged();
 
-        if (spectator && !gameInfo.spectators_can_chat() && tabSupervisor->getAdminLocked()) {
+        /* Spectators can only talk if:
+         * (a) the game creator allows it
+         * (b) the spectator is a moderator/administrator
+         * (c) the spectator is a judge
+         */
+        if (!(spectator && (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || judge))) {
```
ebbit1q, line 1822: I think the style this was in previously was a lot more readable, simply change this to && !judge?

ZeldaZach: That won't work, because you need to be a spectator and one of the three options. This was the cleanest way I could come up with.

`if NOT (A & (B OR (NOT C) OR D))`

ebbit1q: I got confused so I made a truth table, your code will hide the chat for everyone who isn't a spectator :(
`(a, b, c, d, not (a and (b or not c or d)), a and not b and (c or not d)) for a, b, c, d in ((bool(i & (1 << n)) for n in range(4)) for i in range(1 << 4))`

cockatrice/src/tab_game.cpp
```diff
@@ -1814,10 +1814,26 @@ void TabGame::createMessageDock(bool bReplay)
...
+            if (gameInfo.spectators_can_chat() || !tabSupervisor->getAdminLocked() || judge) {
+                canChat = true;
+            }
+        }
+
+        if (!canChat) {
             sayLabel->hide();
             sayEdit->hide();
         }
+
```
kodabb, line 1836: nit: unrelated cosmetic change

kodabb (approved):

cockatrice/src/tab_game.cpp
ebbit1q: looks a lot better though

cockatrice/src/tab_supervisor.cpp
```diff
@@ -621,11 +621,17 @@ void TabSupervisor::updateCurrent(int index)
...
 
+/**
+ * Determine if a user is a moderator/administrator
+ * By seeing if they have the admin or log tabs open
+ * @return if one of the two tabs are open
+ */
 bool TabSupervisor::getAdminLocked() const
 {
-    if (!tabAdmin)
+    if (!tabAdmin && !tabLog)
```
ebbit1q, line 631: this segfaults if tabadmin is uninitialized while tabLog is true

ZeldaZach: True. I'm kinda planning ahead for the other PR with this change, so I'll clean it up

cockatrice/src/tab_supervisor.cpp
```diff
@@ -621,6 +621,11 @@ void TabSupervisor::updateCurrent(int index)
         emit setMenu();
 }
 
+/**
+ * Determine if a user is a moderator/administrator
+ * By seeing if they have the admin tab open & unlocked
+ * @return if one of the two tabs are open
```
ebbit1q, line 627: this isn't valid anymore

ebbit1q (approved):
bikeshedder seal of approval(tm)(r)(c)

tooomm:
Does it make sense to include it into the featureset file?

ZeldaZach:
Not really, because the server owner just has to update. It's not a breaking change on the client side.

PR #4092, opened 2020-09-09
Add dropdown for game age filtering

knitknit:
Options include 5/10/30min, 1/2hr, or no filter. Contains many cosmetic and performance changes as well courtesy @ebbit1q (https://github.com/knitknit/Cockatrice/pull/2)

## Related Ticket(s)
- Fixes #3067
- Addresses age request in #2431

## Short roundup of the initial problem
Users don't want to see games that are 2-3 hours old, and it's hard to properly close out such games.

## What will change with this Pull Request?
- Adding a dropdown selection for game age filtering
- Updates logic that actually filters games, including logic to transition between the combo (selection dropdown) box and the intended age filtering
- Does not apply any age filtering by default

## Screenshots

Can see "no limit" is default, and "Clear filters" button remains disabled on cold load:
Screen Shot 2020-09-23 at 9 11 55 PM

Full list of options:
Screen Shot 2020-09-23 at 9 11 43 PM

Issue #4003, opened 2020-05-18
Use an english translation file for better pluralization


tooomm:
**Description of New Feature**

@ebbit1q pointed this out in the discord:
>what I'm talking about is translating the source strings to english similarly as we do all the others to have: to quote the help article here: "To handle plural forms in the native language, you need to load a translation file for this language, too. The lupdate tool has the -pluralonly command line option, which allows the creation of TS files containing only entries with plural forms."

Linking #2029

ebbit1q:
So zach delegated this to be by assigning me maintainership on transifex.
I want to move the source translation file to cockatrice_en@source.ts as painlessly as possible, without losing the existing translations on transifex.
Changing the file in the build scripts is a piece of cake but I don't want transifex to break :grimacing:

ebbit1q:
I err added the new en_US language on transifex, it's complaining that noone translates it? I guess I hsould try searching for (s) and replace them

ebbit1q:
![image](https://user-images.githubusercontent.com/36401181/94848906-f626be00-0424-11eb-97dc-215ab907faf6.png)
this warning is idiotic, can we disable it? I can see it preventing our translators from properly translating our project.

tooomm:
I can hopp on discord and we can have a look. But I’m busy until Sunday for sure. 😔

Thanks for looking into that @ebbit1q.
Curious how transifex handles this and everything works out in the end. Several questions come to my mind...

PR #4095, opened 2020-09-10
refactor getting game age

ebbit1q:
## Related Ticket(s)
- #4092

#4092 copied these static values for game ages which made me upset, they're added in 25dbfb3 based on 93ab9f9, which introduced these weird values while qt simply has these time classes available.

- games older than 24 hours will now show their age as tr("days")
- games older than 5/10/20 hours will now show their age as >5/10/20 tr("h")
- games older than 1 hour will show as n tr("h")
- games older than 5/10/20 minutes will show as >5/10/20 tr("min")
- games older than 2 minutes will show as n tr("min")
- games younger than that will show as tr("new")
- h and min can be translated with access to pluralization

image

### DIFF
--- a/cockatrice/src/gamesmodel.cpp
+++ b/cockatrice/src/gamesmodel.cpp
@@ -24,7 +24,7 @@ enum GameListColumn
     SPECTATORS
 };
 
-const QString GamesModel::getGameCreatedString(const int secs) const
+const QString GamesModel::getGameCreatedString(const int secs)
 {
     static const QTime zeroTime{0, 0};
     static const int wrapSeconds = zeroTime.secsTo(zeroTime.addSecs(-1));
@@ -36,20 +36,31 @@ const QString GamesModel::getGameCreatedString(const int secs) const
     }
 
     QTime total = zeroTime.addSecs(secs);
+    if (total.minute() < 2) { // games are new during their first minute
+        return tr("New");
+    }
+
+    QString unit;
+    int amount;
     if (total.hour()) {
         total = total.addSecs(halfHourSecs); // round up
-        return QString::number(total.hour()) + QLatin1String("+ h");
-    } else if (total.minute() < 2) { // games are new during their first minute
-        return tr("New");
+        amount = total.hour();
+        unit = tr("hr(s)", "short notation age in hours", amount);
     } else {
         total = total.addSecs(halfMinSecs); // round up
-        int tenMinutes = total.minute() / 10;
-        if (tenMinutes) { // aggregate multiples of ten
-            return QString::number(tenMinutes) + QLatin1String("0+ m");
-        } else {
-            return QString::number(total.minute()) + QLatin1String("+ m");
+        amount = total.minute();
+        unit = tr("min(s)", "short notation age in minutes", amount);
+    }
+
+    QString larger{};
+    for (int aggregate : {20, 10, 5}) { // floor to values in this list
+        if (amount >= aggregate) {
+            amount -= amount % aggregate;
+            larger = ">";
+            break;
         }
     }
+    return larger + QString::number(amount) + " " + unit;
 }
 
 GamesModel::GamesModel(const QMap<int, QString> &_rooms, const QMap<int, GameTypeMap> &_gameTypes, QObject *parent)
--- a/cockatrice/src/gamesmodel.cpp
+++ b/cockatrice/src/gamesmodel.cpp
@@ -7,6 +7,7 @@
 #include "userlist.h"
 
 #include <QDateTime>
+#include <QTime>
 #include <QDebug>
 #include <QIcon>
 #include <QStringList>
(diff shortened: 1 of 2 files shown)


cockatrice/src/gamesmodel.cpp
```diff
@@ -25,23 +26,30 @@ enum GameListColumn
...
+        return QString::number(total.hour()) + QLatin1String("+ h");
+    } else if (total.minute() < 2) { // games are new during their first minute
+        return tr("New");
+    } else {
+        total = total.addSecs(halfMinSecs); // round up
+        int tenMinutes = total.minute() / 10;
+        if (tenMinutes) { // aggregate multiples of ten
+            return QString::number(tenMinutes) + QLatin1String("0+ m");
+        } else {
+            return QString::number(total.minute()) + QLatin1String("+ m");
```
tooomm, line 50: Did you change from the standard used for time (`min`) to `m` for a particular reason?

tooomm:
I guess that makes things easier than using our own baked version. :)

Can we further simplify the UI by not having all values up to 10m shown individually in the process?
I think it is fully sufficient to have `New`, `5+ min`, `10+ min`, `20+ min` etc. - maybe even cut 5min as well?
It doesn't give away too much information to have different value for e.g. 2,3,4 minute old games. But in return makes the interface clearer in my opinion.

cockatrice/src/gamesmodel.cpp
ebbit1q: that's not intentional

ebbit1q:
sure

cockatrice/src/gamesmodel.cpp
```diff
@@ -23,25 +24,43 @@ enum GameListColumn
...
+    } else {
+        total = total.addSecs(halfMinSecs); // round up
+        amount = total.minute();
+        unit = tr("min(s)", "short notation age in minutes", amount);
+    }
+
+    QString larger{};
+    for (int aggregate : {20, 10, 5}) { // floor to values in this list
+        if (amount >= aggregate) {
+            amount -= amount % aggregate;
```
ebbit1q, line 58: note that this is only relevant in a single case right now, if >40 mins is not needed this would make more sense:
```suggestion
            amount = aggregate;
```

tooomm:
Hmm, I had some weird occasions where old games showed as `New`, even if they were listed in the bottom spots.
Some time later it was ok again. I try to make some screenshots.

tooomm:
Ok, that seems fixed. I still run into this:

- first
time1

- later
time2

- later
time3

ebbit1q:
it sorts them alphabetically?

ebbit1q:
this is unacceptable! :(

tooomm:
I assume if a game is >40min old and is about to transition to be a 60min game (=1h) it somehow gets shows as 0mins first, before some time later it is correctly shown as a 1h game. Look at the marked game and the one above in all three screenshots. The game list was all the time sorting by age from newest to oldest as the screen shows. So no way a 0min game ends up between 40 and 60min.
Not the case for you? Saw it plenty of times - nothing rare.

<br>

Other than that I thought about the time values. Using `>x` over `x+`... sure.
Going with `min(s)` over `min` and using `hr(s)` over `h` is not necessary I guess and we can save that bit of space. These "symbols" or "abbreviations" shouldn't need a plural as they are always used in one form - maybe they don't even require a translation.
They are standardised (not strictly part of SI though) and worldwide accepted: https://en.wikipedia.org/wiki/Non-SI_units_mentioned_in_the_SI
Probably "hr" is also used in some kind of context?

Not sure if [this](https://github.com/Cockatrice/Cockatrice/pull/4095/files#diff-09c63d3b0431cdbd5a905f746469009cR62) can lead to issues with arabic or other languages where they read from right to left for example.

I guess we are pretty save to assume all world understands those standard symbols and what they stand for, probably they are even used in non-latin characters languages as well? That way we can keep translations and code a bit simpler.

ebbit1q:
you're right

ebbit1q:
also I originally spelled out hours in the translate calls but decided to make it shorter again :/

ebbit1q:
about the (s): https://github.com/Cockatrice/Cockatrice/issues/4003

ZeldaZach:
I _really_ dislike the `(s)` added. I find it poor user experience. If we could have the system know when to add the s, that would be most idea.

ebbit1q:
@ZeldaZach please read the ticket I linked, it proposes to use qt's translation to eliminate all (s) in the application

tooomm:
Don't think about it as a short form for "minute" or "minutes"... it's nothing linguistic.
I see it as a unit to measure time. And the symbol used for that is simply "min". No matter if it's 1 minute or 42 minutes... it's written as 1 min and 42 min (maybe without the space?).

That's why I made the proposition to not even translate it.

cockatrice/src/gamesmodel.cpp
ebbit1q: it's now like this

tooomm:
Short feedback in between before I have more time next week...

As spotted on the Chickatrice server today, the `days` value is not working properly yet:
days

ebbit1q:
it intentionally doesn't bother to calculate the amount of days

ebbit1q:
I changed it to say >day to make it more consistent with the other values

cockatrice/src/gamesmodel.cpp
```diff
@@ -23,25 +24,37 @@ enum GameListColumn
...
     }
-    return ret;
+
+    QTime total = zeroTime.addSecs(secs);
+    QTime totalRounded = total.addSecs(halfMinSecs); // round up
+    QString form;
+    int amount;
+    if (totalRounded.hour()) {
+        amount = total.addSecs(halfHourSecs).hour(); // round up separately
+        form = tr("%1%2 h", "short age in hours", amount);
```
ZeldaZach, line 44: Would like this to be hr, to be in alignment with "min"
```suggestion
        form = tr("%1%2 hr", "short age in hours", amount);
```

cockatrice/src/gamesmodel.cpp
```diff
@@ -23,25 +24,37 @@ enum GameListColumn
...
-        int unitOfTen = secs / SECS_PER_TEN_MIN;
-        QString str = "%1%2";
-        ret = str.arg(QString::number(unitOfTen), "0+ min");
-    } else { // from 1 hr onward we show hrs
-        int hours = secs / SECS_PER_HOUR;
-        if (secs % SECS_PER_HOUR >= SECS_PER_MIN * 30) // if the room is open for 1hr 30 mins, we round to 2hrs
-            ++hours;
-        ret = QString("%1+ h").arg(QString::number(hours));
+    if (secs >= wrapSeconds) { // QTime wraps after a day
+        return tr(">day");
```
ZeldaZach, line 35:
```suggestion
        return tr(">1 day");
```

cockatrice/src/gamesmodel.cpp
ebbit1q: as mentioned by @tooomm the appropriate unit shorthand is h https://en.wikipedia.org/wiki/Non-SI_units_mentioned_in_the_SI

cockatrice/src/gamesmodel.cpp
ebbit1q: I refrained from mentioning a specific amount of days because there is no real need to check it in my opinion. Games older than 24 hours shouldn't even be necessary. Though we can't say with certainty it's just 1 day.

cockatrice/src/gamesmodel.cpp
ebbit1q: if you want it to be hr, you should add that to the english translation file instead: https://github.com/Cockatrice/Cockatrice/issues/4003

ZeldaZach (approved):

PR #4097, opened 2020-09-12
skip duplicate nonpromos

ebbit1q:
## Related Ticket(s)
- https://github.com/Cockatrice/Cockatrice/pull/3965

lotus cobra is shown as an alternative printing with fancy frame in cockatrice as number 307 because it does not have the isAlternative property set

note that for testing this you'd need to clear the image cache and restart cockatrice after running oracle

- all cards with the frameEffects property will be treated like promos and will be deferred to be added later together with the promos
- if a special set contains only promos and alt frame versions the "non promo" alt frame versions are no longer given priority over the promo versions and either of them might become the selected printing for that set (in practice highest collector number takes precedence afaik)

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -278,7 +278,10 @@ int OracleImporter::importCardsFromSet(CardSetPtr currentSet, const QList<QVaria
 
         if (skipSpecialCards) {
             // skip promo cards if it's not the only print
-            if (getStringPropertyFromMap(card, "isPromo") == "true" || card.contains("frameEffects")) {
+            if (allNameProps.contains(name)) {
+                continue;
+            }
+            if (getStringPropertyFromMap(card, "isPromo") == "true") {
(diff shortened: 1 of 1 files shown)


ZeldaZach (changes requested):
This is incorrect, based on the values a frameEffect can have.
See https://mtgjson.com/data-models/card/ for more details, but some examples include "devoid" and "miracle", which are clearly fine examples that shouldn't be delegated away

ebbit1q:
I considered parsing the frameeffects but opted to instead have all cards that are not "promo" or something like it retain the first copy instead of parsing all of them and keeping the last.
This means that cards from unstable like Secret Base will now show the first iteration instead of the last and cards like lotus cobra in znr will keep the lower collectors number because that's the order in which they appear from mtgjson

ebbit1q:
see https://github.com/mtgjson/mtgjson/issues/679

ZeldaZach (approved):
This seems okay as a workaround for now. When we integrate V5, we should have better handling hopefully

PR #4102, opened 2020-09-16
fix QByteArray::append(QString) deprecation in qt 5.15.1

ebbit1q:
explicitly adds conversion to qbytearray before appending of strings with toUtf8()

### DIFF
--- a/cockatrice/src/deckstats_interface.cpp
+++ b/cockatrice/src/deckstats_interface.cpp
@@ -52,8 +52,7 @@ void DeckStatsInterface::getAnalyzeRequestData(DeckList *deck, QByteArray *data)
     urlQuery.addQueryItem("deck", deckWithoutTokens.writeToString_Plain());
     urlQuery.addQueryItem("decktitle", deck->getName());
     params.setQuery(urlQuery);
-    QString queryString = params.query(QUrl::EncodeReserved);
-    data->append(queryString.toUtf8());
+    data->append(params.query(QUrl::EncodeReserved).toUtf8());
 }
 
 void DeckStatsInterface::analyzeDeck(DeckList *deck)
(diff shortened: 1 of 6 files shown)


cockatrice/src/deckstats_interface.cpp
```diff
@@ -52,7 +52,8 @@ void DeckStatsInterface::getAnalyzeRequestData(DeckList *deck, QByteArray *data)
     urlQuery.addQueryItem("deck", deckWithoutTokens.writeToString_Plain());
     urlQuery.addQueryItem("decktitle", deck->getName());
     params.setQuery(urlQuery);
-    data->append(params.query(QUrl::EncodeReserved));
+    QString queryString = params.query(QUrl::EncodeReserved);
+    data->append(queryString.toUtf8());
```
ZeldaZach, line 56: Can you not just say `data->append(param.query(QUrl::EncodeReserved).toUtf8());`?

ebbit1q: of course, it just becomes harder to read.

ZeldaZach: I find this a bit more confusing as-is honestly. we're setting a temp variable to just use it next line

ebbit1q: I disagree, it breaks up the long statement, making it easier to interpret

ZeldaZach (approved):

ebbit1q:
@ZeldaZach catalina build stalled https://travis-ci.org/github/Cockatrice/Cockatrice/jobs/728014132#L85

ZeldaZach:
Restarted

ebbit1q:
I opted to use toUtf8 in the client but toLatin1 in the qxt code, I'm not 100% sure about which is the correct option but we might consider replacing the old qxt code with something we don't have to maintain ourselves.

PR #4103, opened 2020-09-22
add fedora33 to travis

ebbit1q:
fedora33 is coming out soon but is not released yet, this adds fedora33 ahead of time in order to check for compatibility

### DIFF
--- a/.ci/Fedora33/Dockerfile
+++ b/.ci/Fedora33/Dockerfile
@@ -1,0 +1,20 @@
+FROM fedora:33
+
+RUN dnf install -y \
+        @development-tools \
+        ccache \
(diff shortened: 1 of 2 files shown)


ZeldaZach (approved):

tooomm:
Fedora 33 released two days ago.

I think we should replace 32 then. By the time we have a new release up users updated probably.
I'm afraid of an ever crowing number of jobs. Next there will be two Debian builds etc. 😆

ebbit1q:
we don't have to do the tests on every single version but having the releases is nice and it's not like they are very hard to set up

ebbit1q:
closing in favor of #4164

PR #4105, opened 2020-09-24
Fix show-buddies-only filter loading, and add handling for creator na…

### LINKED ISSUE #3282: Cockatrice ignores saved "Show buddies only games" setting
<b>System Information:</b>
  

  

Client Version: 2.5.2-beta (2018-06-10)
Build Architecture: 64-bit
Qt Version: 5.11.0
System Locale: en_US

  

  

Regardless of what filter I used previously, the saved setting for filtering buddies only games (`filter_games/show_buddies_only_games` in `gamefilters.ini`) is completely ignored, and instead is always false when i launch trice.

Additionally, trice does not save my preference at all when i close the app.

This was last fixed in #2434, but it seems to have been broken since.

knitknit:
…me filter storage/loading.

## Related Ticket(s)
- Fixes #3282

"Show buddies only" filter setting was not remembered properly. #4088 fixed it to load the proper default of true, but this PR actually reads the stored value.

- Read the stored value for "show buddies only" filter
- Store and read value for "creator name" filter


ebbit1q (approved):
works great :+1:

PR #4116, opened 2020-09-29
Fix spoiler update message

### LINKED ISSUE #3689: "Spoiler Updated" notification no longer appearing
<b>System Information:</b>
Client Version: 2.7.1-custom(be7eaa6) (2019-04-05)
Build Architecture: 64-bit
Qt Version: 5.9.7
System Locale: de_DE
Install Mode: Standard

There used to be a notification when new data entered the spoiler.xml file.
It doesn't display any longer for me.
So currently, I know there is new data when no notification is shown on start. :)

The `No Updates Available` notification works and gets displayed on app start correctly.

<br>

Can others confirm this? Is it a windows only problem?

tooomm:
## Related Ticket(s)
- Fixes #3689

The format at Magic-Spoiler changed at one point a long time ago to `Created At:`.
https://github.com/Cockatrice/Magic-Spoiler/blob/files/spoiler.xml#L3

- Update the way Cockatrice parses the file accordingly.

Sometimes having a look at the code doesn't hurt. :)

### DIFF
--- a/cockatrice/src/spoilerbackgroundupdater.cpp
+++ b/cockatrice/src/spoilerbackgroundupdater.cpp
@@ -167,8 +167,8 @@ bool SpoilerBackgroundUpdater::saveDownloadedFile(QByteArray data)
         QList<QByteArray> lines = data.split('\n');
 
         foreach (QByteArray line, lines) {
-            if (line.contains("created:")) {
-                QString timeStamp = QString(line).replace("created:", "").trimmed();
+            if (line.contains("Created At:")) {
+                QString timeStamp = QString(line).replace("Created At:", "").trimmed();
                 timeStamp.chop(6); // Remove " (UTC)"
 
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
I remember noticing this before but thought it didn't have an effect like this :sweat_smile: this comment parsing code is ridiculous though, this should be handled differently using xml

tooomm:
Jep, a dedicated key and parsing that would be way better...

Check https://github.com/Cockatrice/Magic-Spoiler/pull/251, there is something in the works (but pending reviews) for a while. Can you have a look maybe? @ebbit1q

ZeldaZach:
While I'm not overly thrilled with this fix, it's a fix nonetheless. I'd like this addressed in the future to look for distinct indicators, not awkward string comparisons

tooomm:
To do this right, the Spoiler repo needs an update to provide the file in v4 style (https://github.com/Cockatrice/Magic-Spoiler/issues/232)
Then the information and created information can get included which I already prepared (https://github.com/Cockatrice/Magic-Spoiler/issues/251).

PR #4120, opened 2020-10-01
[WIP] add english translation

### LINKED ISSUE #4003: Use an english translation file for better pluralization

ebbit1q:
## Related Ticket(s)
- Fixes #4003

this will add part 1 of the move, by moving the source files to a different location, the other part would be adding the translation to transifex and then pulling the new files from there.

- moves en.ts to en@source.ts in cockatrice and oracle
- updates translations


ZeldaZach:
Note: Push to master should trigger a re-build for cmake stuff and push to transifex

PR #4123, opened 2020-10-02
apply clang format to proto files

### LINKED ISSUE #3060: Expand clang-format for proto and js
While #3028 covered `Language: Cpp`

@Daenyth also suggested in https://github.com/Cockatrice/Cockatrice/issues/2183#issuecomment-249416472 to include a style for proto.
Clang-format could also be expanded to cover the webclient written in Javascript!

Interesting link with help/description regarding the different options availabel: https://clangformat.com
Official clang-format documentation: https://clang.llvm.org/docs/ClangFormat.html

ebbit1q:
## Related Ticket(s)
- Fixes #3060

proto files are not automatically formatted

- rules that currently define our proto files are added to .clang-format instructions
- clangify will now pick up the proto extension
- the existing files are formatted

### DIFF
--- a/.ci/travis-lint.sh
+++ b/.ci/travis-lint.sh
@@ -18,7 +18,7 @@ case $err in
 ***  Then commit and push those changes to this branch. ***
 ***   Check our CONTRIBUTING.md file for more details.  ***
 ***                                                     ***
-***                    Thank you ❤️                     ***
+***                    Thank you ❤️                      ***
 ***                                                     ***
 ***********************************************************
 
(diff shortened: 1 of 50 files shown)


ebbit1q:
I think I'm using a different version of clang-format :joy:

ctrlaltca:
The value of `AllowShortBlocksOnASingleLine` used to avcept only boolean values up to clang 9; from clang 10 it accepts Never/Empty/Always.
IIRC travis is stuck at clang 5.

ebbit1q:
well this displays another problem, if an internal error happens in the command `clang-format x | diff x -` the command diff will exit nonzero and hide the clang-format exit code even with set -o pipefail

ebbit1q:
also it displays the version, it's using version 7, a rare case of foresight on my part

Issue #2642, opened 2017-04-24
Missing game filter options


tooomm:
- Show available games
- Show games which do not allow spectators

In case you are looking to just spectate games the last one would be pretty handy.
The first one though, in combination with "Show unavailable games" would allow to just show unavailable games solely because you could disable available games.

Default should be to show available games and non-spectator games for sure.

Daenyth:
I'm thinking maybe we should kill unavailable and just have toggles for the
options it represents

tooomm:
Can you explain a bit more?

But yes, I need to think about the wording twice to make exactly sure all needed/wanted cases are working. Combining things under toggles might be easier to understand and use since you just have less check boxes to deal with.

Daenyth:
- Show full games
- Show games with passwords
- Show games for registered users only
- Show buddy only games

tooomm:
Your comment confuses me a bit @Daenyth. :)

While it would be nice to have the `registered user` option, this issue is about having filters to
- find `games for spectators` and
- to show `unavailable games` (or full games) only

Right now you can just decide of you want to display full games additionally, but you can't hide available games. Something which you want in case you want to spectate a running match and you don't need to see all this games which are waiting for players to join.

ZeldaZach:
Show full games
    Show games with passwords
    Show games for registered users only
    Show buddy only games

I can probably look into those changes @Daenyth 
Seems straight forward to me. Will also add  `Show games that allow spectators`

Daenyth:
@tooomm I didn't explain myself clearly, I was talking about removing the "Show unavailable games" option because it is confusing and instead replacing it with a switch for everything it does.

tooomm:
Oh right, so you extended my idea.

So we have something like this in the end? (wording could need some work)

```
not full games --> "Show games with places left"
full games --> "Show full games"
with pw --> "Show games with password protection"
reg only --> "Show games for registered users only"
buddy only --> "Show buddy only games"
allow spec --> "Show games that allow spectators"
```

tooomm:
https://github.com/Cockatrice/Cockatrice/issues/628 also requests a switch to separate and show reg/unreg games individually.

So it needs to be:
games for reg only --> "Show games for registered users only"
games for unreg only --> "Show games for unregistered users"
(otherwise you can't only display reg games... because unreg games would always show up)

I hope I got it right... this is quite confusing. 😕

ZeldaZach:
Dupe #2431

PR #4127, opened 2020-10-04
Add game filtering for spectator attributes

### LINKED ISSUE #388: additional filters for "filter games"
I really like watching games of other players. I usually run it in the background on my tv while I work on my laptop. What's kind of frustrating is that there are games listed that I cannot watch.

When a user turns on "show unavailable games", there should be a tick box that hides games that do not allow spectators.

There should also be the option to hide games of people that are not on your buddy list. This can be used to easily filter out people you cannot watch because it is buddy list only, or filter out players you don't want to play against because they aren't on your buddy list.

I also have a question about the buddy list lock. If there is a game being played that is "buddies only", do you have to be a buddy of the creator, or can you view it if you are buddies with the joining party?

### LINKED ISSUE #2431: Add more filters for game list in rooms
Currently you can filter rooms by everything except Age and Spectator Rights. It'd be nice if those were available too.  
I've never once filtered rooms by creator name, description or player count, nor will I ever need to, but I'd love it if I could filter out games that don't allow spectators (especially when looking for rooms to spectate) and games where the host is presumably afk (~30+min).  Other people might want to filter out games where specs can see hands in fear of cheating or filter by time even more restrictively to only see fresh rooms. 
To throw in a suggestion of specific implementation:  
Spec Rights:  4 tickboxes (show "specs banned" games / show "specs allowed" games / show "specs can chat" games / show "specs can see everything" games).  
Age: same as the current player count filter, just for time in minutes in place of headcount.

knitknit:
## Related Ticket(s)
- Fixes #2431
- Fixes #388 
- Splits "unavailable" into full/started per #2642 but does not yet split out registered/unregistered users

Users want to be able to filter games by spectator attributes (are spectators allowed, is password required, can spectators chat, can spectators see hands).

Add UI and settings plumbing for spectator attribute filtering, see screenshots below.

## Screenshots
Latest:
Screen Shot 2020-10-14 at 11 09 39 PM

### Old screenshots
For draft, initial settings filter to show only if spectators can watch (this is easily changed, just proof of concept).
Screen Shot 2020-10-03 at 9 18 33 PM

If show-only-if-spectators-can-watch is deselected, the remaining spectator-related filters are disabled in the UI. They are also ignored when actually filtering games.
Screen Shot 2020-10-03 at 9 18 52 PM

### DIFF
--- a/cockatrice/src/dlg_filter_games.cpp
+++ b/cockatrice/src/dlg_filter_games.cpp
@@ -113,23 +113,23 @@ DlgFilterGames::DlgFilterGames(const QMap<int, QString> &_allGameTypes,
     auto *restrictionsGroupBox = new QGroupBox(tr("Restrictions"));
     restrictionsGroupBox->setLayout(restrictionsLayout);
 
-    spectatorsCanWatch = new QCheckBox(tr("Show games &spectators can watch"));
-    spectatorsCanWatch->setChecked(true);  // DO NOT SUBMIT get from storage
-    connect(spectatorsCanWatch, SIGNAL(toggled(bool)), this, SLOT(toggleSpectatorCheckboxEnabledness(bool)));
+    showOnlyIfSpectatorsCanWatch = new QCheckBox(tr("Show games only if &spectators can watch"));
+    showOnlyIfSpectatorsCanWatch->setChecked(gamesProxyModel->getShowOnlyIfSpectatorsCanWatch());
+    connect(showOnlyIfSpectatorsCanWatch, SIGNAL(toggled(bool)), this, SLOT(toggleSpectatorCheckboxEnabledness(bool)));
 
-    spectatorsNeedPassword = new QCheckBox(tr("Show if spectators need password"));
-    spectatorsNeedPassword->setChecked(true);  // DO NOT SUBMIT get from storage
-    spectatorsCanChat = new QCheckBox(tr("Show if spectators can chat"));
-    spectatorsCanChat->setChecked(true);  // DO NOT SUBMIT get from storage
-    spectatorsCanSeeHands = new QCheckBox(tr("Show if spectators can see hands"));
-    spectatorsCanSeeHands->setChecked(true);  // DO NOT SUBMIT get from storage
-    toggleSpectatorCheckboxEnabledness(getSpectatorsCanWatch());
+    showSpectatorPasswordProtected = new QCheckBox(tr("Show spectator password p&rotected games"));
+    showSpectatorPasswordProtected->setChecked(gamesProxyModel->getShowSpectatorPasswordProtected());
+    showOnlyIfSpectatorsCanChat = new QCheckBox(tr("Show only if spectators can ch&at"));
+    showOnlyIfSpectatorsCanChat->setChecked(gamesProxyModel->getShowOnlyIfSpectatorsCanChat());
+    showOnlyIfSpectatorsCanSeeHands = new QCheckBox(tr("Show only if spectators can see &hands"));
+    showOnlyIfSpectatorsCanSeeHands->setChecked(gamesProxyModel->getShowOnlyIfSpectatorsCanSeeHands());
+    toggleSpectatorCheckboxEnabledness(getShowOnlyIfSpectatorsCanWatch());
 
     auto *spectatorsLayout = new QGridLayout;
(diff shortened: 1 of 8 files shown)


knitknit:
@ZeldaZach @ebbit1q @tooomm let me know what you think of this as a general approach, thanks!

cockatrice/src/settings/gamefilterssettings.cpp
```diff
@@ -129,3 +129,47 @@ bool GameFiltersSettings::isGameTypeEnabled(QString gametype)
...
+
+void GameFiltersSettings::setShowOnlyIfSpectatorsCanWatch(bool show)
+{
+    setValue(show, "show_only_if_spectators_can_watch", "filter_games");
+}
+
+bool GameFiltersSettings::isShowOnlyIfSpectatorsCanWatch()
+{
+    QVariant previous = getValue("show_only_if_spectators_can_watch", "filter_games");
+    return previous == QVariant() ? true : previous.toBool();
```
ebbit1q, line 141: I think this default is a bit harsh, people most likely still want to join games even if they don't allow spectators.

cockatrice/src/gamesmodel.cpp
```diff
@@ -27,6 +27,10 @@ const bool DEFAULT_UNAVAILABLE_GAMES_VISIBLE = false;
 const bool DEFAULT_SHOW_PASSWORD_PROTECTED_GAMES = true;
 const bool DEFAULT_SHOW_BUDDIES_ONLY_GAMES = true;
 const bool DEFAULT_HIDE_IGNORED_USER_GAMES = false;
+const bool DEFAULT_SHOW_ONLY_IF_SPECTATORS_CAN_WATCH = true;
```
ebbit1q, line 30: I think this should be false by default

cockatrice/src/dlg_filter_games.cpp
```diff
@@ -63,12 +63,12 @@ DlgFilterGames::DlgFilterGames(const QMap<int, QString> &_allGameTypes,
     generalGroupBox = new QGroupBox(tr("General"));
     generalGroupBox->setLayout(generalGrid);
 
-    QVBoxLayout *gameTypeFilterLayout = new QVBoxLayout;
+    auto *gameTypeFilterLayout = new QVBoxLayout;
     QMapIterator<int, QString> gameTypesIterator(allGameTypes);
     while (gameTypesIterator.hasNext()) {
```
ebbit1q, line 68: while you're at it with the refactoring, can you format this too?

ebbit1q:
I think now that this dialog is a lot bigger with the recent additions the qgroupbox on the right with the game types is a bit slim, I feel like it needs a min width or to prevent having it's entries spread so thin with a spacer.

cockatrice/src/dlg_filter_games.cpp
knitknit: Sure, what formatting would you like changed? Sorry if missing something obvious

knitknit:
> I think now that this dialog is a lot bigger with the recent additions the qgroupbox on the right with the game types is a bit slim, I feel like it needs a min width or to prevent having it's entries spread so thin with a spacer.

Reasonable! I pulled in a suggestion from duped #2642 (splitting up "unavailable" into full/started), and also pulled spectator over to the right underneath game types. Let me know what you think.

cockatrice/src/settings/gamefilterssettings.cpp
knitknit: Switched to false

cockatrice/src/gamesmodel.cpp
knitknit: Switched

ebbit1q:
the dialog is now a lot wider, also considering the list of formats is arbitrary it could actually be longer
image
this shows some inconsistency with resizing the dialog, the left column has a spacer added while the right one does not.
image

ebbit1q (approved):
as long as you don't resize the dialog it looks the best it can now, also the extra options are nice for wanting to see unstarted games as a spectator but I see no applications otherwise.

cockatrice/src/dlg_filter_games.cpp
ebbit1q: I prefer the `for` syntax instead of java style, but we don't even call attention to this in the CONTRIBUTING.md so just leave it as is.

PR #4131, opened 2020-10-05
Guess card names on plain text imports

ebbit1q:
## Related Ticket(s)
- Fixes part of https://github.com/dr4fters/dr4ft/issues/1242 but the deck hash will not match

plaintext import sometimes breaks when a card is using the wrong format for split cards

further suggestions for improvements would be to take the left hand side if the right hand side is given and to hide right hand sides in the deck editor

- modify up the simplifyCardName function to ignore right halves
- add guessCard function that prioritises full card names the simple ones
-  fix imports for misformatted split cards or double faced cards
- introduces a small concession: not completely formatted names with a
    shared name on the left side will get mixed up, eg "bind" but not "Bind"
  - this should be fine considering how this would fix a lot more cards

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -12,6 +12,7 @@
 #include <QDir>
 #include <QFile>
 #include <QMessageBox>
+#include <QRegularExpression>
 #include <algorithm>
 #include <utility>
 
@@ -291,22 +292,23 @@ void CardInfo::refreshCachedSetNames()
 
 QString CardInfo::simplifyName(const QString &name)
 {
-    QString simpleName(name);
+    static const QRegularExpression spaceOrSplit("(\\s+|\\/\\/.*)");
+    static const QRegularExpression nonAlnum("[^a-z0-9]");
+
+    QString simpleName = name.toLower();
+
+    // remove spaces and right halves of split cards
+    simpleName.remove(spaceOrSplit);
 
     // So Aetherling would work, but not Ætherling since 'Æ' would get replaced
     // with nothing.
     simpleName.replace("æ", "ae");
-    simpleName.replace("Æ", "AE");
 
     // Replace Jötun Grunt with Jotun Grunt.
     simpleName = simpleName.normalized(QString::NormalizationForm_KD);
 
-    // Replace dashes with spaces so that we can say "garruk the veil cursed"
-    // instead of the unintuitive "garruk the veilcursed".
-    simpleName = simpleName.replace("-", " ");
-
-    simpleName.remove(QRegExp("[^a-zA-Z0-9 ]"));
-    simpleName = simpleName.toLower();
+    // remove all non alphanumeric characters from the name
+    simpleName.remove(nonAlnum);
     return simpleName;
 }
 
@@ -435,6 +437,19 @@ QList<CardInfoPtr> CardDatabase::getCards(const QStringList &cardNames) const
 CardInfoPtr CardDatabase::getCardBySimpleName(const QString &cardName) const
 {
     return getCardFromMap(simpleNameCards, CardInfo::simplifyName(cardName));
+}
+
+CardInfoPtr CardDatabase::guessCard(const QString &cardName) const
+{
+    CardInfoPtr temp = getCard(cardName);
+    if (temp == nullptr) { // get card by simple name instead
+        temp = getCardBySimpleName(cardName);
+        if (temp == nullptr) { // still could not find the card, so simplify the cardName too
+            QString simpleCardName = CardInfo::simplifyName(cardName);
+            temp = getCardBySimpleName(simpleCardName);
+        }
+    }
+    return temp; // returns nullptr if not found
 }
 
 CardSetPtr CardDatabase::getSet(const QString &setName)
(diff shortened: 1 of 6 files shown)


ebbit1q:
updated op

cockatrice/src/deck_loader.cpp
```diff
@@ -288,13 +289,25 @@ QString DeckLoader::getCardZoneFromName(QString cardName, QString currentZoneNam
     return currentZoneName;
 }
 
-QString DeckLoader::getCompleteCardName(const QString cardName) const
+QString DeckLoader::guessCompleteCardName(const QString cardName) const
 {
+    static const QRegularExpression evenSimplerExp("(\\s+|\\/\\/.*)"); // matches whitespace or double names
```
ZeldaZach, line 294: Your regex is not intuitive, and the name doesn't help its cause

cockatrice/src/carddatabase.cpp
```diff
@@ -291,22 +292,22 @@ void CardInfo::refreshCachedSetNames()
 
 QString CardInfo::simplifyName(const QString &name)
 {
-    QString simpleName(name);
+    static const QRegularExpression nonAlnumspace("[^a-z0-9 ]");
```
ZeldaZach, line 295: can we come up with a better variable name

cockatrice/src/deck_loader.cpp
```diff
@@ -288,13 +289,25 @@ QString DeckLoader::getCardZoneFromName(QString cardName, QString currentZoneNam
...
     if (db) {
-        CardInfoPtr temp = db->getCardBySimpleName(cardName);
+        const CardInfoPtr temp = db->getCardBySimpleName(cardName);
         if (temp) {
             return temp->getName();
         }
+
+        // could not find the card, so simplify more
+        QString simpleCardName = CardInfo::simplifyName(QString(cardName).remove(evenSimplerExp));
+        for (const CardInfoPtr &card : db->getCardList()) {
```
ZeldaZach, line 304: This iteration looks like it's super slow, as it's iterating 50k cards and then applying a regular expression to each... is this the best approach?

cockatrice/src/deck_loader.cpp
ebbit1q: it simplifies the card names further than the simplify cardname function does, and it matches all whitespace or anything after and including a double slash

cockatrice/src/carddatabase.cpp
ebbit1q: it matches anything that is not alphanumeric or a space

cockatrice/src/deck_loader.cpp
ebbit1q: it's actually applying around 4 regexes, and it's only actually used if a card is not found in the database. I considered storing the table but it's important to update the table when it's reloaded but it's also not used often enough to warrant always creating this new table on a reload.

ebbit1q: I just tested it, it's quite slow actually, it takes like 2 seconds for only 12 cards

ebbit1q: that is 12 unknown cards, cards that are immediately recognized are snappy as normal

ebbit1q:
updated op

ZeldaZach (approved):

mixmix:
hey @ebbit1q I'm using cockatrice 2.8.0 and I'm seeing an example of an imported card with // in it's name **semi-working**

image
 
So the problem is it shows up, but the "related card" is not accessible, and a friend (perhaps running an older version of cockatrice?) couldn't see the card at all

For contrast see:
image
You'll notice this card has a lot more detail like :
- related card
- color identity
- card text
- layout

mixmix:
Could be a janky set definition for kaldheim, tried it with importing some other cards:
:heavy_check_mark: fire // ice (split)
:heavy_check_mark: Jace, Vryn's Prodigy // Jace, Telepath Unbound (transform)
:heavy_check_mark: Turntimber Symbiosis // Turntimber, Serpentine Wood (modal_dfc)

all totally fine!

ebbit1q:
this is because your darkbore pathway with the // in it is from an already existing deck that, perhaps because of the inconsistency with the generated spoilers, now has the deck saved with the // in it.
now the card image fetcher will try to find a matching image for it online, which it did successfully, however it is still an unknown card without textbox.
note that when importing modal dfcs from plain text some corrections get applied but that means that the created deck does not have a matching hash as the original.

mixmix:
Yeah interesting. So we need a global spec for generating hashes which doesn't know about magic or dual sided cards, and some careful rules and design about what the canonical name/ identifier for a card is.

Do other platforms like mtgo / arena have deck hashes?

ebbit1q:
no, at least, not ones that are visible to its users. consider why they were implemented (https://github.com/Cockatrice/Cockatrice/commit/0bb4ef8bb8ce7952122a2e26a7850bdc381a8cf4), they exist to allow both players to verify their opponents aren't cheating with the cards in their deck in a competitive setting. this simplifies something that'd have to be integrated through a lot of extra effort and protocols or through a third party checking decks manually (like in real life tournaments). it is not a perfect solution but rather a good enough one.

PR #4134, opened 2020-10-07
free qprocess on oracle update fail

ebbit1q:
reload the database whenever oracle exits
search for oracle in ../oracle so you can use it from the build dir

### DIFF
--- a/cockatrice/src/window_main.h
+++ b/cockatrice/src/window_main.h
@@ -116,11 +116,11 @@ private:
     void createTrayIcon();
     void createTrayActions();
     int getNextCustomSetPrefix(QDir dataDir);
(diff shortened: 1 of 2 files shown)


cockatrice/src/window_main.h
```diff
@@ -116,11 +116,11 @@ private slots:
     void createTrayIcon();
     void createTrayActions();
     int getNextCustomSetPrefix(QDir dataDir);
-    // TODO: add a preference item to choose updater name for other games
```
ebbit1q, line 119: I removed this comment because we are no longer planning to support this

ZeldaZach (approved):

PR #4139, opened 2020-10-12
userlists --> account

### LINKED ISSUE #1737: tab_userlists --> tab_account (code file rename required)
In #906 we changed the former `User lists` tab to `Account`.

However the code files didn't receive a renaming and are still named `tab_userlists.h` & `tab_userlists.cpp`. They need to get renamed as well I think. :)

The code may include many more old references related to that change that confuse new developers which want to work with the code!

---

other example:
https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/sequenceEdit/ui_shortcutstab.h#L1478
It should be "ClearFilterSelected" or sth., not "ClearFilterOne"
You are going to clear filter one? It's not precise...

tooomm:
## Related Ticket(s)
- Fixes #1737

Interface got updated, but code had still the old naming pattern.
It's easier now to navigate in the source files for new contributors.

- rename file
- update file calls

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -79,7 +79,7 @@ SET(cockatrice_SOURCES
     src/tab_replays.cpp
     src/tab_supervisor.cpp
     src/tab_admin.cpp
-    src/tab_userlists.cpp
+    src/tab_account.cpp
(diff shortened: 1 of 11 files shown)


ebbit1q (approved):
rather insignificant imo, but sure

PR #4140, opened 2020-10-12
Travis: fix lint msg

tooomm:
Toooo muuuch spaccce

### DIFF
--- a/.ci/travis-lint.sh
+++ b/.ci/travis-lint.sh
@@ -18,7 +18,7 @@ case $err in
 ***  Then commit and push those changes to this branch. ***
 ***   Check our CONTRIBUTING.md file for more details.  ***
 ***                                                     ***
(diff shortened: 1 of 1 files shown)


ebbit1q (changes requested):
this is font specific, maybe you have a font that uses two character widths here instead of one
image

tooomm:
Oh, maybe!
For me on Windows it's always wider, no matter if viewed with Firefox or Chromium

- Travis log:
travis

- GitHub
github

ZeldaZach (changes requested):
It was correct as-is, might be a font issue on your end

ebbit1q:
the :+1: on the other message is using two widths though, but my browser doesn't have a monospace version of that emoji in its fonts somehow so it's never aligned for me.

Issue #2491, opened 2017-03-17
Local game, 6 player, order of players...


mpalmujoki:
<b>OS:</b>
macOS Sierra 10.12.3

<b>Cockatrice version:</b>
Version dab7316 (2017-01-19)

___
<br>

The order of switching players is wrong in 6 player 3v3 team games:

I tend to do 3v3 games with one player in middle. I've noted that in Cockatrice, it switches from the player to the lower right to the player in middle on the left side, instead of the player in the left bottom:

![image](https://cloud.githubusercontent.com/assets/8876090/24060958/33fe5f94-0b5e-11e7-8d57-0342fcc1c54c.png)

I cant understand why the order goes like that. What I expected, is this:

![image](https://cloud.githubusercontent.com/assets/8876090/24061009/655850d6-0b5e-11e7-85f0-7053f32b69bc.png)

Because of the odd player switching logic, I have to double hop to get to the bottom player, and then hop 5 hops to get to the middle player I earlier hopped over... It gets tedious. 

Any support for this, please?

Daenyth:
The player order is the order in the player list from top to bottom. The problem is that we don't align and rotate the board layout appropriately for this, especially with 2 column layout. This should be a client-side-only ui fix I believe

skwerlman:
The ability to change the play order would be nice, though

Daenyth:
Probably not happening soon

PR #4144, opened 2020-10-15
Multiple background images on all zones

fdipilla:
## Related Ticket(s)
- Related to #3985

Now you may have a different tablezone background image for each player but not different handzones, playerzones or stackzones.

With this you can have a different handzone, playerzone and stackzone background image for each player.

### DIFF
--- a/cockatrice/src/handzone.cpp
+++ b/cockatrice/src/handzone.cpp
@@ -83,7 +83,6 @@ void HandZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*optio
         brush = themeManager->getExtraHandBgBrush(QString::number(player->getZoneId()), brush);
     }
     painter->fillRect(boundingRect(), brush);
-    // painter->fillRect(boundingRect(), themeManager->getHandBgBrush());
 }
 
 void HandZone::reorganizeCards()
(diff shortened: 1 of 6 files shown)


ebbit1q:
this seems to be working fine for me
image
when will this be ready for review?

fdipilla:
I was having a similar issue than what is shown in your screenshot, notice the mismatch between the 'Player 1' name and the zone number 2, the same with 'Player 2'.
I was hoping to take a look at that in the next few days, but maybe it's just me nitpicking.
If you think that the PR is good as is I will mark it as ready for review right ahead.

ebbit1q:
I think it's fine this way, it aligns with the way the background was alterated this entire time, https://github.com/Cockatrice/Cockatrice/issues/2491 is a separate issue btw

ebbit1q:
I propose we merge this and if you want you can make a separate pr for the turn order or background order issue.

fdipilla:
Totally missed that it has its own issue :)

ebbit1q (approved):
:+1: makes these consistent with the backgrounds, we could consider how we could prevent all this code duplication though

tooomm:
I removed the `[WIP]` tag from the PR title for you as you marked it as "ready" now @fdipilla.

The "Themes" wiki page is also updated already.

ZeldaZach:
Thanks for your hard work helping to improve my project. It's greatly appreciated. If you'd ever like to chat about other improvements, feel free to hit us up on Discord or Gitter :)

PR #4147, opened 2020-10-25
Travis: go back to macOS 10.14

tooomm:
## Related Ticket(s)
- Related #4059

As noticed in the last image update (https://github.com/Cockatrice/Cockatrice/pull/4059) users **need** to update their OS in order to keep using Dark Mode support with that change as the new 10.15 zip file will not work on 10.14 any longer! While the hardware requirements mostly don't prevent that, users on 10.14 suddenly still lost Dark Mode support when updating Cockatrice without knowing.
Our approach used to be to reach most users with our release binaries. Following that we should go back to 10.14 for our extra Dark Mode binary, users on Catalina can still use that one.

- Update the travis images to use macOS 10.14 again with the most recent Xcode 11.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -159,7 +159,7 @@ jobs:
     before_cache:
     - brew cleanup
 
-  #macOS Mojave
+  #macOS Mojave (additional build to offer Dark Mode support introduced in macOS 10.14)
   - name: macOS Mojave (Debug)
     if: tag IS NOT present
     os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -159,11 +159,11 @@ jobs:
     before_cache:
     - brew cleanup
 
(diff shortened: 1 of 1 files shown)


ebbit1q:
I specifically asked about this and it was argued that users using mojave should upgrade to catalina https://github.com/Cockatrice/Cockatrice/pull/4059#issuecomment-673017870

tooomm:
True. We still did not warn or educate users about the loss of functionality or how to get it back... we simply silently turned Dark Mode support off for all 10.14 users.

Now that the harm is done... :) 
...how much do we care and how big of an issue is it that we want to go back to macOS 10.14 for the next release?
I'm also fine with skipping it and closing this.

tooomm:
See #4164

Issue #4148, opened 2020-10-27
Resizing filter widget behaves weirdly


ebbit1q:
> the dialog is now a lot wider, also considering the list of formats is arbitrary it could actually be longer
> ![image](https://user-images.githubusercontent.com/36401181/96749941-bd4a8b00-13cb-11eb-8e65-491f380fce3d.png)
> this shows some inconsistency with resizing the dialog, the left column has a spacer added while the right one does not.
> ![image](https://user-images.githubusercontent.com/36401181/96750302-17e3e700-13cc-11eb-822b-a7cd05f5537b.png)

knitknit:
Sorry I didn't get around to this before the original PR was merged in - thanks @tooomm !

PR #4149, opened 2020-10-28
Fix resizing for game filter dialog

tooomm:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4148

Dialog behaved weird on resizing.

- Added a stretch element to the right "column" as well. Game types or spectator options will not stretch any longer
- As there is no reason in making the dialog longer, it's now always the suggested vertical size (same as the `Create game` dialog)

## Screenshots
- before:
image

- after:
with long list of game types
stretchfixed

    with short list of games types
    stretchfix2


knitknit (approved):
Thanks for the fixes!

PR #4152, opened 2020-10-30
Allow undo draw if card was moved within hand

### LINKED ISSUE #326: Bug: Undo draw function fails after rearranging cards
If you draw a card, then rearrange the freshly drawn card in your hand, the undo draw function will not work anymore. (presumably because it can't track the card in its new position.)

knitknit:
## Related Ticket(s)
- Fixes #326

User could not undo draw if drawn card was moved around within hand.

- Do not clear `lastDrawList` if the card is moved within hand, only clear once a recently drawn card is moved outside the hand
- Stop doing `lastDrawList.removeAt(lastDrawList.indexOf(card->getId()));` when `undoingDraw == true`, as the card was already popped off the end of the list in line 1086: https://github.com/Cockatrice/Cockatrice/blob/bec02b495299992b4210a8cc3724ca63679d7616/common/server_player.cpp#L1085-L1089

## Open question
How should "undo draw" work from a player/judge's perspective? Right now I think it's technically possible to get into a situation where say cards A, B, C are drawn; A is played, so list is cleared; user draws D; user plays B, which doesn't cause the list to be cleared; user undoes draw D. Should this be allowed?

## Screenshots
[Screen Recording 2020-10-29 at 10.08.02 PM.zip](https://github.com/Cockatrice/Cockatrice/files/5462608/Screen.Recording.2020-10-29.at.10.08.02.PM.zip)

### DIFF
--- a/common/server_player.cpp
+++ b/common/server_player.cpp
@@ -149,9 +149,9 @@ void Server_Player::setupZones()
     // ------------------------------------------------------------------
 
     // Create zones
-    Server_CardZone *deckZone = new Server_CardZone(this, "deck", false, ServerInfo_Zone::HiddenZone);
+    auto *deckZone = new Server_CardZone(this, "deck", false, ServerInfo_Zone::HiddenZone);
     addZone(deckZone);
-    Server_CardZone *sbZone = new Server_CardZone(this, "sb", false, ServerInfo_Zone::HiddenZone);
+    auto *sbZone = new Server_CardZone(this, "sb", false, ServerInfo_Zone::HiddenZone);
     addZone(sbZone);
     addZone(new Server_CardZone(this, "table", true, ServerInfo_Zone::PublicZone));
     addZone(new Server_CardZone(this, "hand", false, ServerInfo_Zone::PrivateZone));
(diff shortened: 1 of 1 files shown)


knitknit:
@ebbit1q what do you think of this change and the open question I have in the description? I don't think it's new or specific to this change, just something I realized as I was thinking about it.

ebbit1q:
if the same player doesn't play any card they should be able to undo their draw, moving cards around is fine and other players should be able to play cards in between even though that's pretty confusing.

ZeldaZach:
As long as playerA does no game actions involving moving cards from their hand to any other zone, I see no reason for the undo to not work.

ebbit1q:
theoretically moving cards into your hand wouldn't break it, actually the only thing that'd break it is moving the specific drawn card out of your hand, but it seems reasonable to disallow it to avoid issues where players can no longer track game actions.

ZeldaZach:
Once cards start leaving/entering the hand, I would call the hand "tainted" and we shouldn't allow a fix in that case. Also, I don't remember, but if I draw 3 cards one at a time, can I undo the draw 3 times? Or does it only work for the latest?

ZeldaZach (approved):
Tested and audited. This makes perfect sense to me and should be a great change. Will deploy to RR soon.

PR #4156, opened 2020-11-01
Fix racetime condition with token cloning

### LINKED ISSUE #2820: Crash on cloning a token that's no longer there
<b>OS: macOS 10.12.5</b> 

<b>Cockatrice version: 2.317</b>
__________

[CR3.txt](https://github.com/Cockatrice/Cockatrice/files/1202587/CR3.txt)

Right clicked on a token, opponent destroyed said token, right click menu for that was still open and I clicked clone. Client crashed.

ZeldaZach:
## Related Ticket(s)
- Fixes #2820

There was a racetime condition where an opponent would delete an object and the game tried to make a copy of it. The call to `this->setCursor()` triggered the crash because `this` was a null object by the time we got to that call. Removing that call seemed to fix up the issue.

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -48,22 +48,22 @@ CardItem::~CardItem()
 
 void CardItem::prepareDelete()
 {
-    if (owner) {
+    if (owner != nullptr) {
         if (owner->getCardMenu() == cardMenu) {
-            owner->setCardMenu(0);
-            owner->getGame()->setActiveCard(0);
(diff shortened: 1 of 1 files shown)


ebbit1q:
with this change the cursor will never open after clicking :(

ZeldaZach:
If you have a better idea, I'm down. This seems to be an annoying issue and it looks fine on the mac as you'll be doing other actions fairly quick anyway, such as rolling over another card

ebbit1q:
grabbing a card fixes it but this should be fixable with a simple check

PR #4157, opened 2020-11-01
Allow DeckName/Comments to be resized more

### LINKED ISSUE #566: Feature Request - Vertically Adjustable Comments and Deck Box Sizes In Deck Editor
Afternoon.
I've been thinking for a while how I wish I was able to adjust the size of my comment box in the deck editor. I don't know how the workflow is of other players, but I use the comments box pretty heavily with an assortment of notes. (combos and interactions, a maybe board for card ideas for the deck, sideboard notes etc) however the box size itself isn't really conducive to expanded note taking.

Without overhauling the deck editor UI to incorporate a more detailed notation area, I'm proposing making the boxes adjustable on a vertical axis. I don't know how difficult it is to adjust the UI with Qt, but since there isn't anything mapping to them (that I can tell), I don't think it's major. Moving a title box, adding a vertical adjust property etc. But I'm not a dev, so what do I know. Maybe I'm way off base.

Here's an (ugly) example of what I'm talking about: http://screencast.com/t/7Ly0sWqbb
Essentially, I think it would not only look cleaner to have the boxes be congruent in width, but it would serve as a practical purpose of creating more usable space for the end user.

ZeldaZach:
## Related Ticket(s)
- Fixes #566

The comments section had too large of a maximum height, and there was no way to get just pure deck viewing. The default is the same, but if a user wants to resize the limits can be seen below

Screen Shot 2020-11-01 at 4 25 59 PM

Screen Shot 2020-11-01 at 4 26 03 PM

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -26,8 +26,6 @@
 #include <QDir>
 #include <QDockWidget>
 #include <QFileDialog>
-#include <QGroupBox>
-#include <QHBoxLayout>
 #include <QHeaderView>
 #include <QLabel>
 #include <QLineEdit>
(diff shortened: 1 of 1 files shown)


tooomm:
Works! I like that one can grow the decklist even more now.

Did you thought about renaming the labels to `Name` and `Notes` to give more space to users who use comments extensively without requiring them to hide too much from the deck?

ZeldaZach:
Outside the scope of the ticket @tooomm. If that's something users want, we can address it in a ticket and PR, if necessary.

ebbit1q (approved):
very simple change, tested :+1:

PR #4160, opened 2020-11-08
add missing mysql connector dependencies to docker images

ebbit1q:
## Related Ticket(s)
- possibly #3954

debian based builds are missing the mysql connector dependency, see https://travis-ci.org/github/Cockatrice/Cockatrice/jobs/740707002#L2032

- adds the dependency to ubuntu and debian builds

### DIFF
--- a/.ci/ArchLinux/Dockerfile
+++ b/.ci/ArchLinux/Dockerfile
@@ -5,6 +5,7 @@ RUN pacman --sync --refresh --sysupgrade --needed --noconfirm \
         ccache \
         cmake \
         git \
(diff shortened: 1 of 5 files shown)


skwerlman:
do the wiki instructions for compiling also need to be updated?

ebbit1q:
@skwerlman almost none of the instructions on the wiki are aimed at building the server, any instruction that has -DWITH_SERVER=1 should have the missing dep.

ebbit1q:
Adding instructions for servers would help I guess, it's not properly documented in general.

PR #4169, opened 2020-11-18
refactor pingClockTimeout

ebbit1q:
try to see if it changes #3954

### DIFF
--- a/common/server_game.cpp
+++ b/common/server_game.cpp
@@ -169,26 +169,29 @@ void Server_Game::pingClockTimeout()
     GameEventStorage ges;
     ges.setGameEventContext(Context_PingChanged());
 
-    QList<ServerInfo_PlayerPing *> pingList;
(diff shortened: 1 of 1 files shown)


common/server_game.cpp
```diff
@@ -169,26 +169,29 @@ void Server_Game::pingClockTimeout()
...
+                newPingTime = -1;
+            }
+        }
 
-        if ((newPingTime != -1) && !player->getSpectator())
+        if ((newPingTime != -1) && !player->getSpectator()) {
             allPlayersInactive = false;
+        }
 
         if ((abs(oldPingTime - newPingTime) > 1) || ((newPingTime == -1) && (oldPingTime != -1)) ||
```
ebbit1q, line 196: this code is basically obfuscated, I'd like to fix this but I'm worried about breaking it

ZeldaZach (approved):

Issue #4171, opened 2020-11-22
unable to rejoin private room on reconnect


ebbit1q:
@ebbit1q Here's what I just did:

* My friend created a password protected room on Chickatrice
* I entered the room, placed some cards in the table
* Took off my internet connection, got this:

![image](https://user-images.githubusercontent.com/20914054/99887405-d6d92f80-2c22-11eb-8810-f0b6efe77bed.png)

* Reconnected, verified my internet is fine, and then clicked OK on the error dialog above.

* The following dialog appeared immediately:

![image](https://user-images.githubusercontent.com/20914054/99887450-4d762d00-2c23-11eb-9ea4-8d69635792bf.png)

I chose Chickatrice again, added the same username and password that I used in the first time, and then the following error message appeared:

![image](https://user-images.githubusercontent.com/20914054/99887506-b493e180-2c23-11eb-8585-9d774e19a241.png)

Also, at this moment the game room that was open (paused in the last state before I lost connection) closed itself automatically.

_Originally posted by @papb in https://github.com/Cockatrice/Cockatrice/issues/3939#issuecomment-731636267_

tooomm:
>I chose Chickatrice again, added the same username and password that I used in the first time, and then the following error message appeared:
>
>![image](https://user-images.githubusercontent.com/20914054/99887506-b493e180-2c23-11eb-8585-9d774e19a241.png)

Do you possibly use an unregistered account for joining Chickatrice? @papb

papb:
Hello! Thank you @ebbit1q for creating this for me, I missed the part in which you said to create a new ticket instead of commenting there.

Debug Log:

```
Client Version: 2.7.0 (2019-03-04)
Client Operating System: Windows 7 SP 1 (6.1)
Build Architecture: 64-bit
Qt Version: 5.9.7
System Locale: pt_BR
Install Mode: Portable
---------------------------------------------------------------------------
Portable mode enabled
Theme changed: "Default"
SoundEngine: disabling sound
Sound theme changed: "Default"
SoundEngine: disabling sound
Language changed: ""
main(): starting main program
Spoilers Disabled
CardDatabase::loadCardDatabases start
[CockatriceXml3Parser] Trying to parse:  "E:/Portable/Cockatrice/data/cards.xml"
[CockatriceXml3Parser] Parsing failed: wrong version 4
[CockatriceXml4Parser] Trying to parse:  "E:/Portable/Cockatrice/data/cards.xml"
[...]

PR #4173, opened 2020-11-23
Server room message: clearify wording

tooomm:
## Related Ticket(s)
- Related #4171

The `room` wording was not always clear. In some cases the user could not tell if it was `game room`  or `server room`.

- Added the term server to the mix.

Not all of these messages need it though as some only occur after a certain action and are linked to that in direct response. To stay uniform I changed them all.
Do you guys prefer to only change the problematic one that can occur without direct reference and can cause confusion because the client tries to auto join all available rooms?
`You do not have the required permission to join this server room.`


ebbit1q (approved):
sure

PR #4181, opened 2020-11-24
[WIP] GH Actions tests

tooomm:
The 10.13 environment isn't found.

- Try 10.14 instead.

### DIFF
--- a/.github/workflows/macos-builds.yml
+++ b/.github/workflows/macos-builds.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         target:
           - Debug
-          - ElCapitan
+          - HighSierra
           - Mojave
           - Catalina
           - BigSur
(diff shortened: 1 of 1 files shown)


ebbit1q:
I expected it to not find 10.14, instead it uses 10.15 which doesn't include this xcode version, it seems like we are dropping HighSierra support...

tooomm:
Interesting that it replaces 10.14 with 10.15, but shows an error message about not finding 10.13 altogether:
```
Check failure on line 1 in .github
@github-actions github-actions / build (ElCapitan)

.github#L1

No runner matching the specified labels was found: macos-10.13.

```

tooomm:
>it seems like we are dropping HighSierra support...

It looks like the lowest GH Actions can offer is macOS 10.14 Mojave (via Xcode 10.3 on macOS 10.15), yes.

ebbit1q:
it seems like macos 10.14 images don't exist as I already thought, I'm going to close this pr.

PR #4188, opened 2020-11-26
ui: Add shortcut for "Save deck as..."

### LINKED ISSUE #4174: Use [Ctrl+Shift+S] as a Shortcut to "Save deck as..."
Use Ctrl+Shift+S as a Shortcut to "Save deck as..."

**Additional Context**
Ctrl+S already is used for "Save deck"
We could also implement the other established saving-Shortcut. A lot of software uses [Ctrl+Shift+S] as a Shortcut to "Save as...". A few examples: Libreoffice, most other text-editors, Blender, a lot of video-editing software, etc.
- Most applications use this shortcut.
- It doesn't seem to be used for something else in the deck-editor

So why not use it?

Fenrirthviti:
## Related Ticket(s)
- Fixes #4174

No shortcut existed for saving as, simple single line change. 

- Ctrl+Shift+S now opens the "Save deck as..." dialog

Compiled and tested on Windows 10 x64 (2004) against Qt 5.15.2. Works as expected.

## Screenshots

img

### DIFF
--- a/cockatrice/src/shortcutssettings.h
+++ b/cockatrice/src/shortcutssettings.h
@@ -219,7 +219,7 @@ private:
                                                 parseSequenceString("Ctrl+S"),
                                                 ShortcutGroup::Deck_Editor)},
         {"TabDeckEditor/aSaveDeckAs", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Save Deck as..."),
-                                                  parseSequenceString(""),
+                                                  parseSequenceString("Ctrl+Shift+S"),
                                                   ShortcutGroup::Deck_Editor)},
         {"TabDeckEditor/aSaveDeckToClipboard",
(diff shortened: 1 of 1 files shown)


tooomm:
I've run into issues when testing your new shortcut... it's displayed for me in the menu and in the shortcut settings, but it doesn't seem to work.
Cockatrice doesn't recognize it. I added a card to a deck and tried the "save as" menu option - it works.
Using the shortcut does not, it's not captured somehow. I also tried focusing the deck widget, the main app window etc.

When trying to edit the shortcut in settings, I can set it to CTRL+Shift+E for example and it works then.
Manually setting it back to CTRL+Shift+S does not. The combination is not recognized. I have to manually reset it to defaults to bring it back to that.

Windows 10 64bit 20H2 here (I'm using the 64bit version for Windows compiled by AppVeyor with Qt 5.12.6)

ebbit1q (approved):
Tested, works as expected.

PR #4193, opened 2020-11-29
CI: Fix up Windows builds and add GitHub Actions for Windows

Fenrirthviti:
## Related Ticket(s)
- Related to #3059

Windows builds are somewhat neglected right now, this fixes up and provides repeatable viable builds in CI using GitHub Actions over AppVeyor

- Both 32bit and 64bit builds will be generated
- Artifacts are uploaded in the form of the installer packge.
  - This can be easily modified to provide just the program directories instead if desired, to avoid having to run an installer for test builds
- Adds vcpkg as a submodule to the repository
  - This is the recommended way of using vcpkg with a repo
- Fixes a bug with how the NSIS installer script was checking for 64bit builds
  - NOTE: This fix will break current Windows appveyor builds.

## Screenshots
n/a


Fenrirthviti:
As another note, this PR will need to be updated to define when it should run. I wasn't sure exactly when you would all want to trigger a build, so just let me know and I can update accordingly. Currently this only triggers on my test branch, so merging as-is will not do anything.

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,132 @@
+name: 'CI Windows On-Push'
+
+on:
+  push:
+    branches:
+      - github-actions
```
tooomm, line 6: >As another note, this PR will need to be updated to define when it should run.

We run our Linux and Mac builds currently on pushes to and PR's against master (ignoring md-only changes), as well as on published GitHub releases:

```suggestion
on:
  push:
    branches:
      - master
    paths-ignore:
      - '**.md'
  pull_request:
    branches:
      - master
    paths-ignore:
      - '**.md'
  release:
    types:
      - published
```

Fenrirthviti: Updated to reflect this, thanks!

Fenrirthviti:
Another note, I am actually working on better integrating vcpkg in to the project using a manifest rather than a responses file. I am on the fence about adding that to this PR, since it's modifying cmake and the entire build process on Windows significantly, as opposed to just the small streamlining that this does currently.

Should I hold that for a separate PR after this is merged, or include it here?

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,142 @@
...
+        uses: actions/checkout@v2.3.3
+        with:
+          submodules: 'recursive'
+      - name: 'Get Cockatrice git info'
+        shell: bash
+        working-directory: ${{ github.workspace }}
+        run: |
+          git fetch --prune --unshallow
+          echo "GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
+          echo "GIT_HASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
```
ebbit1q, line 38: isn't this already available as GITHUB_SHA?

ebbit1q:
If this works and replaces appveyor adequately I'd merge this in now and make the cmake changes a separate pr.

.github/workflows/windows-builds.yml
Fenrirthviti: GITHUB_SHA is not the same when dealing with PRs vs the main repo, apparently: <https://github.community/t/github-sha-not-the-same-as-the-triggering-commit/18286>

This is safer from what I understand. There's a couple other issues I've seen on the checkout actions repo as well referencing this. Doing it this way will ensure that the triggering PR hash is being used.

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,142 @@
...
+        working-directory: ${{ github.workspace }}
+        run: |
+          cd build32
+          msbuild /m /p:Configuration=Release PACKAGE.vcxproj
+          cp *.exe ../Cockatrice-${{ env.GIT_TAG }}-${{ env.GIT_HASH }}-32bit-installer.exe
+      - name: 'Publish'
+        if: success()
+        uses: actions/upload-artifact@v2.2.0
+        with:
+          name: 'Cockatrice-${{ env.GIT_TAG }}'
```
ebbit1q, line 141: it'd be preferable to have these not share their name so you can download both artifacts separately
```suggestion
          name: 'Cockatrice-${{ env.GIT_TAG }}-x86'
```

ebbit1q (approved):
As is this only adds the build artifacts for prs etc to the ci checks and doesn't upload them to github releases like the other builds do, I'd be up for quickly adding that in the same fashion we do on our other builds though.

Adding a submodule is something we avoided messing around with before but if it's the expected way to build on windows I guess it can just be ignored on other platforms.

.github/workflows/windows-builds.yml
Fenrirthviti: Sure, easy change I'll do that in a moment.

Fenrirthviti:
Automatic releases is easy enough to add in, I left it out as I wasn't sure if that was something that was desired.

Submodules are not the expected way to build on windows, but it's the expected way to build when relying on vcpkg specifically.  I am not sure who made the choice initially to use vcpkg, but it's honestly not even 100% required as you could build/compile all the deps yourself individually if you want, or get the deps from other locations. vcpkg streamlines this process a little bit, but if it's not desired to include in the project I can look in to other options. Personally, I found it pretty easy to work with, despite the submodule requirement for getting the most out of it.

.github/workflows/windows-builds.yml
Fenrirthviti: Added, but used 32/64bit nomenclature as that is more "natural" to us smooth-brain windows users. Hopefully that is acceptable :)

ebbit1q: of course it is :)

ebbit1q:
Can you just rip out the .appveyor.yml here so we don't have to do it later?

Fenrirthviti:
Going to mark this as draft for now while I add the tests. On appveyor they upload the results to appveyor directly for a simple output returned, but we don't have that option on GHA. I'm going to see if I can get it working in CI first, and then figure out a way to check for any test failures. If this is not a hard dependency to merging, feel free to remove from draft status and merge as is, I'll keep at this and submit a new PR when it's ready.

Fenrirthviti:
Ok, that took less time than expect, so this is ready to go again.

tooomm:
That looks sweet. Great to have all in one place finally!!


Can we please implement uploading to GitHub releases again before merging this? Our file deployment for windows is gone now without substitution. :)
Also, we should definitely test that with a new beta pre-release after merging as well...

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,212 @@
...
+        with:
+          vcpkgArguments: '@${{ github.workspace }}/vcpkg.txt'
+          vcpkgDirectory: '${{ github.workspace }}/vcpkg'
+          appendedCacheKey: ${{ hashFiles(env.vcpkgResponseFile) }}
+      - name: 'Configure Cockatrice 64-bit'
+        working-directory: ${{ github.workspace }}
+        run: |
+          New-Item build64 -type directory -force
+          cd .\build64
+          cmake -G "${{ env.CMAKE_GENERATOR }}" -A "x64" -T "host=x64" -DCMAKE_TOOLCHAIN_FILE="${{ github.workspace }}\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_PREFIX_PATH="${{ runner.workspace }}\Qt64\Qt\5.12.9\msvc2017_64;${{ github.workspace }}\vcpkg\installed\x64-windows\bin" -DCMAKE_BUILD_TYPE="Release" -DWITH_SERVER=1 -DTEST=1 ..
```
ZeldaZach, line 64:
```suggestion
          cmake .. -G "${{ env.CMAKE_GENERATOR }}" -A "x64" -T "host=x64" -DCMAKE_TOOLCHAIN_FILE="${{ github.workspace }}\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_PREFIX_PATH="${{ runner.workspace }}\Qt64\Qt\5.12.9\msvc2017_64;${{ github.workspace }}\vcpkg\installed\x64-windows\bin" -DCMAKE_BUILD_TYPE="Release" -DWITH_SERVER=1 -DTEST=1
```

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,212 @@
...
+        uses: lukka/run-vcpkg@v5
+        with:
+          vcpkgArguments: '@${{ github.workspace }}/vcpkg.txt'
+          vcpkgDirectory: '${{ github.workspace }}/vcpkg'
+          appendedCacheKey: ${{ hashFiles(env.vcpkgResponseFile) }}
+      - name: 'Configure Cockatrice 64-bit'
+        working-directory: ${{ github.workspace }}
+        run: |
+          New-Item build64 -type directory -force
+          cd .\build64
```
ZeldaZach, line 63:
```suggestion
          cd build64
```

ZeldaZach: This suggestion is b/c of https://github.com/Cockatrice/Cockatrice/pull/4193/files#diff-ca5dd169a0cae96b54f57acfa81f594ba1fde1358a8b8488a5aaeb619e985736R71

.github/workflows/windows-builds.yml
ZeldaZach: I prefer having the pathing known at the start, then flags after with cmake

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,212 @@
...
+        uses: lukka/run-vcpkg@v5
+        with:
+          vcpkgArguments: '@${{ github.workspace }}/vcpkg.txt'
+          vcpkgDirectory: '${{ github.workspace }}/vcpkg'
+          appendedCacheKey: ${{ hashFiles(env.vcpkgResponseFile) }}
+      - name: 'Configure Cockatrice 32-bit'
+        working-directory: ${{ github.workspace }}
+        run: |
+          New-Item build32 -type directory -force
+          cd .\build32
```
ZeldaZach, line 128:
```suggestion
          cd build32
```

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,212 @@
...
+          cmake -G "${{ env.CMAKE_GENERATOR }}" -A "Win32" -DCMAKE_TOOLCHAIN_FILE="${{ github.workspace }}\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_PREFIX_PATH="${{ runner.workspace }}\Qt32\Qt\5.12.9\msvc2017;${{ github.workspace }}\vcpkg\installed\x86-windows\bin" -DCMAKE_BUILD_TYPE="Release" -DWITH_SERVER=1 -DTEST=1 ..
```
ZeldaZach, line 129:
```suggestion
          cmake .. -G "${{ env.CMAKE_GENERATOR }}" -A "Win32" -DCMAKE_TOOLCHAIN_FILE="${{ github.workspace }}\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_PREFIX_PATH="${{ runner.workspace }}\Qt32\Qt\5.12.9\msvc2017;${{ github.workspace }}\vcpkg\installed\x86-windows\bin" -DCMAKE_BUILD_TYPE="Release" -DWITH_SERVER=1 -DTEST=1
```

.github/workflows/windows-builds.yml
```diff
@@ -0,0 +1,212 @@
...
+          submodules: 'recursive'
+      - name: 'Fetch git tags and generate file name'
+        shell: bash
+        run: |
+          git fetch --prune --unshallow
+          echo "GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
+          echo "GIT_HASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+          echo "GIT_TAG=$(git describe --tags --abbrev=0)" >> $GITHUB_ENV
+      - name: 'Checking if beta'
+        if: contains(env.GIT_TAG, 'beta')
```
ZeldaZach, line 166: can we not do an if/else here, instead of having two name operations?

ZeldaZach (commented):
Some small suggestions/comments, but overall if this works I'm thrilled to bring Cockatrice packaging to the next level!

.github/workflows/windows-builds.yml
Fenrirthviti: Sure, changed.

.github/workflows/windows-builds.yml
Fenrirthviti: Whoops, thought I removed this already. Looks like I did in the next step. Fixed.

.github/workflows/windows-builds.yml
Fenrirthviti: This is typically "safer" and won't really add any time to the overall workflow as one is always skipped.

vcpkg.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "cockatrice",
+  "version-string": "0.0.0-dev",
```
ZeldaZach, line 3: Do we want this version?

Fenrirthviti: It's not actually used anywhere, just required as part of the json and is the version of the json file itself.

ZeldaZach (approved):
Seems like a solid change, thanks for taking the time to do this!!

PR #4197, opened 2020-11-30
merge build workflows

ebbit1q:
Change the way mac os versions are compared to simply release <= current version.

Add github action grouping and error and warning messages to scripts.

Take the pretty formatted release name from the release title on github, eg "Blessed Sanctuary" and add it to the published files in the same manner we currently do by hand.


ebbit1q:
The ccache path on macOS is still not correct, it should be somewhere else.
Docker is running as a different user, causing it to be unable to move the output, we can just sudo mv or cp the file but this seems unnecessary. Possibly we can just move all this naming to cmake, I hope @Fenrirthviti can instead do it in a more reasonable manner in https://github.com/Fenrirthviti/Cockatrice/tree/cmake-updates 
making this a draft for now

ebbit1q:
with the recent changes to the build naming I simplified name_build.sh and merged everything into a single file with a separate job for preconfiguration/release creation

ZeldaZach (approved):

PR #4199, opened 2020-12-01
Address a handful of warnings from #6095

ZeldaZach:
Solved a handful of the warnings, but there are several we can't control with protobuf. Should turn those off within the protobuf headers.

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -138,9 +138,9 @@ void TabDeckEditor::createDeckDock()
     lowerLayout->addWidget(deckView, 1, 0, 1, 5);
 
     // Create widgets for both layouts to make splitter work correctly
-    QWidget *topWidget = new QWidget;
+    auto *topWidget = new QWidget;
     topWidget->setLayout(upperLayout);
-    QWidget *bottomWidget = new QWidget;
+    auto *bottomWidget = new QWidget;
     bottomWidget->setLayout(lowerLayout);
 
     auto *split = new QSplitter;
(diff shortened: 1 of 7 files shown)


cockatrice/src/tab_deck_editor.cpp
```diff
@@ -1248,7 +1249,7 @@ void TabDeckEditor::showSearchSyntaxHelp()
                .replace(QRegularExpression("^(##)(.*)", opts), "<h2>\\2</h2>")
                .replace(QRegularExpression("^(#)(.*)", opts), "<h1>\\2</h1>")
                .replace(QRegularExpression("^------*", opts), "<hr />")
-               .replace(QRegularExpression("\\[([^\[]+)\\]\\(([^\\)]+)\\)", opts), "<a href=\'\\2\'>\\1</a>");
+               .replace(QRegularExpression(R"(\[([^[]+)\]\(([^\)]+)\))", opts), R"(<a href='\2'>\1</a>)");
```
ebbit1q, line 1252: I don't think this is a readability improvement when all the others use the escaped backslashes

ZeldaZach: This wasn't readable to begin with, but the linter recommended I change it. It no longer has the incorrect escape sequence, which was lost in all the backslashes to begin with.

ebbit1q: you mean the escaped [?

ZeldaZach: Yeah, which wasn't necessary

ebbit1q (approved):
looks fine

PR #4204, opened 2020-12-04
cmake: Minor updates

Fenrirthviti:
## Related Ticket(s)
n/a

CMake was a bit messy

In order of change in the file:

CMakeLists.txt
- Add new cmake var for GIT_TAG_RELEASENAME which replaces the need to pull the release name from a web page
- Update Windows build paths to be more in line with expectations of other Windows projects (non-functional change)
- Cleaned up a comment that was 3 lines to 1 because it was bothering me
- Added missing check for arch for QTDIR (can optionally supply QTDIR32 or QTDIR64 if building both at once, or just QTDIR if only building on one arch), and fixed whitespace issue
- Added a quick message stating that protoc was found if it is. If the check is important enough to be doing, it's important enough to be logging success.
- Fixed typo in dbconverter build flag description

getversion.cmake
- Removed the need to curl a URL and parse the page title. CMake will no longer be making web calls every time it's being run with a valid release/pre-release tag.
- Set a few of the version defaults to blank to avoid potential confusion in third party builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,9 +38,15 @@ if(WIN32)
         message(WARNING "QTDIR variable is missing.  Please set this variable to specify path to Qt (e.g. C:/Qt/5.7/msvc2015_64)")
     endif()
 endif()
+
 # A project name is needed for CPack
 # Version can be overriden by git tags, see cmake/getversion.cmake
 PROJECT("Cockatrice" VERSION 2.7.6)
+
+# Set release name if not provided via env/cmake var
+if(NOT DEFINED GIT_TAG_RELEASENAME)
+    set(GIT_TAG_RELEASENAME "Blessed Sanctuary")
+endif()
 
 # Use c++11 for all targets
 set(CMAKE_CXX_STANDARD 11 CACHE STRING "C++ ISO Standard")
@@ -52,7 +58,6 @@ set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS true)
 # Search path for cmake modules
 SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
-# Retrieve git version hash
 include(getversion)
 
 # Create a header and a cpp file containing the version hash
@@ -95,7 +100,7 @@ if(UNIX)
         endif()
     endif()
 elseif(WIN32)
-    set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/release)
+    set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/rundir/${CMAKE_BUILD_TYPE})
 endif()
 
 # Treat warnings as errors (Debug builds only)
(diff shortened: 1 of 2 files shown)


Fenrirthviti:
Looks like 32bit build failed due to CI derping, should be fine otherwise if review passes.

CMakeLists.txt
```diff
@@ -52,7 +53,11 @@ set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS true)
 # Search path for cmake modules
 SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
-# Retrieve git version hash
+# Set release name if not provided via env var
+if(NOT DEFINED GIT_TAG_RELEASENAME)
+    set(GIT_TAG_RELEASENAME "Blessed Sanctuary")
```
ebbit1q, line 58: can this and the project version be placed in a separate place so we can easily keep track of them, either at the top or a separate file?

Fenrirthviti: Sure, shouldn't matter where in the file this gets set. I just put it next to the getversion include as that's where it's actually used, but shouldn't matter. I'll move it up to the top.

tooomm:
Hmm, second time now the 32bit windows build failed at the `Get Cockatrice git info` step.

<details>
<summary>Details</summary>

```
Run git fetch --prune --unshallow
  git fetch --prune --unshallow
  echo "GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
  echo "GIT_HASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
  echo "GIT_TAG=$(git describe --tags --abbrev=0)" >> $GITHUB_ENV
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
  env:
    QT_VERSION: 5.12.9
    CMAKE_GENERATOR: Visual Studio 16 2019
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: expected flush after ref listing
Error: Process completed with exit code 128.
```
</details>

Re-running the windows builds fixed it again. Nothing we can do about I guess.

Fenrirthviti:
ebbit said they were getting that locally as well, probably just a github issue more than CI. Annoying, but yeah not much we can do.

CMakeLists.txt
```diff
@@ -180,7 +189,9 @@ set(CMAKE_AUTOMOC TRUE)
 # Find other needed libraries
 FIND_PACKAGE(Protobuf REQUIRED)
 IF(NOT EXISTS "${Protobuf_PROTOC_EXECUTABLE}")
-  MESSAGE(FATAL_ERROR "No protoc command found!")
+    MESSAGE(FATAL_ERROR "No protoc command found!")
+ELSE()
+    MESSAGE(STATUS "Protoc found!")
```
ZeldaZach, line 194: Can you specify what version you found?

Fenrirthviti: Whoops, thought I had already pushed this, should be updated now.

PR #4220, opened 2021-01-13
Structure change

jchamish:
## Related Ticket(s)
- Fixes #IssueNumber

- This will change the structure of the react application.


webclient/copy_shared_files.ps1
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+robocopy /E ../common/pb/. ./public/pb/
```
seavor, line 2: Lets keep this file Mac-friendly.

webclient/package.json
```diff
@@ -37,8 +37,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject",
-    "postinstall": "./copy_shared_files.sh"
```
seavor, line 41: And let's keep it as part of the post-install.

seavor:
It looks like /ducks and /store are dupes?

webclient/copy_shared_files.ps1
jchamish: I left it Mac-Friendly with `copy_shared_files.sh` but just adding that file for windows.

webclient/package.json
jchamish: Ok, I can put this back

jchamish:
What's the different between store and ducks. I thought it was ducks because it's redux which was in the react template I was using. 

Reference of the react template I used https://github.com/sharetribe/ftw-daily/tree/master/src/ducks

seavor:
> What's the different between store and ducks? I thought it was ducks because it's redux which was in the react template I was using.
> 
> Reference of the react template I used https://github.com/sharetribe/ftw-daily/tree/master/src/ducks

My understanding was ducks are components that contain their own store. So if you have a Cart duck, it contains both the UI and state management inside the duck.

jchamish:
Will go forward with store and we can always rename the folder after doesn't matter.

seavor:
I see you embedded AppShell/common/services under websocket/instanceServices. These services are the application's access to WebSocket services, but are also a decoupling layer from the websocket itself. Is there a particular reason you made them apart of the Websocket?

jchamish:
Honestly wasn't really sure where to put them but figure it was easy to put them there for now and clean up later. Since all the web socket stuff was together.

ZeldaZach (approved):

PR #4238, opened 2021-01-31
Always look at top card

### LINKED ISSUE #3813: Feature Request: Always reveal top card to me.
Newer cards like Experimental Frenzy and Mystic Forge allow you to look at the top card of your library at any time, essentially meaning it is always revealed to you. The only way to handle this in the client currently is to do `Library > View top cards of library... > 1` every time the top card changes.

It would be nice for players to have an option to enable always revealing the top card to just themselves.

omegaula:
## Related Ticket(s)
- Fixes #3813

Adds option to look at top of library without revealing to anybody else.

- new option in deck menu to "Always look on top of library"
- matching changes in the server

## Screenshots
Screenshot from 2021-01-31 15-01-18

Screenshot from 2021-01-31 15-08-35
Screenshot from 2021-01-31 15-09-04
Screenshot from 2021-01-31 15-09-36


common/server_player.cpp
```diff
@@ -318,16 +318,34 @@ Response::ResponseCode Server_Player::drawCards(GameEventStorage &ges, int numbe
     ges.enqueueGameEvent(eventPrivate, playerId, GameEventStorageItem::SendToPrivate, playerId);
     ges.enqueueGameEvent(eventOthers, playerId, GameEventStorageItem::SendToOthers);
 
-    if (deckZone->getAlwaysRevealTopCard() && !deckZone->getCards().isEmpty()) {
+    revealTopCardIfNeeded(deckZone, ges);
+
+    return Response::RespOk;
+}
+
+void Server_Player::revealTopCardIfNeeded(Server_CardZone *zone, GameEventStorage &ges)
```
ebbit1q, line 326: a much needed refactor :+1:

common/server_player.cpp
```diff
@@ -1987,14 +1987,20 @@ Response::ResponseCode Server_Player::cmdChangeZoneProperties(const Command_Chan
...
-        if (!zone->getCards().isEmpty() && cmd.always_reveal_top_card()) {
-            Event_RevealCards revealEvent;
-            revealEvent.set_zone_name(zone->getName().toStdString());
-            revealEvent.set_card_id(0);
-            zone->getCards().first()->getInfo(revealEvent.add_cards());
+        revealTopCardIfNeeded(zone, ges);
 
-            ges.enqueueGameEvent(revealEvent, playerId);
+        return Response::RespOk;
+    } else if (cmd.has_always_look_at_top_card()) {
```
ebbit1q, line 1993: I don't see a problem with toggling both at the same time, the event should be queued only once of course.

common/server_player.cpp
```diff
@@ -592,21 +610,11 @@ Response::ResponseCode Server_Player::moveCard(GameEventStorage &ges,
...
                                   AttrPT, ptString);
             }
-            if (startzone->getAlwaysRevealTopCard() && !startzone->getCards().isEmpty() && (originalPosition == 0)) {
-                Event_RevealCards revealEvent;
-                revealEvent.set_zone_name(startzone->getName().toStdString());
-                revealEvent.set_card_id(0);
-                startzone->getCards().first()->getInfo(revealEvent.add_cards());
-
-                ges.enqueueGameEvent(revealEvent, playerId);
+            if (originalPosition == 0) {
```
ebbit1q, line 613:
```suggestion
            if (originalPosition == 0 || newX == 0) {
```

common/server_player.cpp
```diff
@@ -318,16 +318,34 @@ Response::ResponseCode Server_Player::drawCards(GameEventStorage &ges, int numbe
...
         Event_RevealCards revealEvent;
-        revealEvent.set_zone_name(deckZone->getName().toStdString());
+        revealEvent.set_zone_name(zone->getName().toStdString());
         revealEvent.set_card_id(0);
-        deckZone->getCards().first()->getInfo(revealEvent.add_cards());
+        zone->getCards().first()->getInfo(revealEvent.add_cards());
 
         ges.enqueueGameEvent(revealEvent, playerId);
     }
+    if (zone->getAlwaysLookAtTopCard()) {
```
ebbit1q, line 339: there is no point in revealing the card twice if both are enabled, preferably the client will disable the other flag when enabling one though.

ebbit1q (changes requested):
Tested working both online and local :+1: 
When both are enabled the same card will be revealed twice, preferably the other option should be disabled when enabling one.
Your opponents are not being told the name of the card but they aren't being told which one of your cards you are looking at, instead they get no information except the card zone, resulting in a log message like this "opponent reveals their library", preferably the other players should get the game event of them looking at the top card similar to scrying.
The player the card is being revealed to is not told whether this was because of the reveal to all players event or not, I propose the look at top cards event is used again, it should be possible to use the same method of revealing in the client as for the public event.

Great work, just a few changes needed :+1:

common/server_player.cpp
omegaula: In practice they will never true in the same event, since each checkbox sends it's own event with only one field filled in.

common/server_player.cpp
omegaula: Those need to be 2 separate conditions, because they affect different zones.

common/server_player.cpp
omegaula: If both are revealed I will send only more visible event.

common/server_player.cpp
ebbit1q: you're right

common/server_player.cpp
ebbit1q: it doesn't matter what the client does, if it is in the protocol it should be accounted for. we are soon to have more than one client.

omegaula: Updated, single event will be send even if both are set,

ebbit1q (approved):
I've tested this change and it works great with only a small addition to the protocol.
Servers can include this change while retaining backwards compatibility.
Other players without this change will only see the fact that the player on the newer version seems to consistently look at the top card of their library whenever it is needed.
Servers that do not have this change will simply ignore it without warning, this is to be expected.

ebbit1q:
Do note that this change also never sends the "stops looking at their library messages" this message is superfluous because this view never updates to contain more information anyway.

omegaula:
@ebbit1q Any idea why the ArchLinux test is failing here? It doesn't look related to my PR, is this a known issue? Should I sync?

ebbit1q:
The arch linux test is a test that is allowed to fail, there is nothing to worry about https://github.com/Cockatrice/Cockatrice/actions/runs/557093022/workflow#L101

ebbit1q:
I don't like that it is adding a little X to the commit though, it should show that it is inconsequential in my opinion.
Though if you look at the bottom of the page you see that it doesn't block merging.

omegaula:
I'm new to Github, what happens now?

ebbit1q:
I have approved the pull request and it will be merged into the master branch after other members approve as well, any amount of time afterwards we will make a 2.8.1 release which will provide the change to all users. This will take some time, we don't have a release schedule and tend to publish whenever the amount of changes seems worthwhile.

My advice is not to worry about it and allow people to take their time, we're all volunteers here :)

ebbit1q:
On the other hand though, there is nothing wrong with at least reminding them by pinging @ZeldaZach or @tooomm here or on discord.

tooomm:
@omegaula Also watch out for our next beta release after this is merged. That way it's accessible to everybody who wants to use it earlier.

Can't properly check atm, I leave this one to Zach.

common/server_player.cpp
```diff
@@ -318,16 +318,39 @@ Response::ResponseCode Server_Player::drawCards(GameEventStorage &ges, int numbe
...
-        deckZone->getCards().first()->getInfo(revealEvent.add_cards());
+        zone->getCards().first()->getInfo(revealEvent.add_cards());
 
         ges.enqueueGameEvent(revealEvent, playerId);
+        return;
     }
+    if (zone->getAlwaysLookAtTopCard()) {
+        Event_DumpZone dumpEvent;
+        dumpEvent.set_zone_owner_id(playerId);
+        dumpEvent.set_zone_name(zone->getName().toStdString());
```
ZeldaZach, line 343: This is a concerning line, because this value is set by the user (See https://github.com/Cockatrice/Cockatrice/pull/4238/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&file-filters%5B%5D=.proto#diff-efb87c85d1930eaec92f8dec6437407d8934e27318e68baca67f0011e4039159R1004) as it's a user passed string we're executing on. This could be modified, recompiled, and give a user an unfair advantage. Something to fix up for sure.

common/server_player.cpp
```diff
@@ -318,16 +318,39 @@ Response::ResponseCode Server_Player::drawCards(GameEventStorage &ges, int numbe
...
+        dumpEvent.set_number_cards(1);
+        ges.enqueueGameEvent(dumpEvent, playerId, GameEventStorageItem::SendToOthers);
 
-    return Response::RespOk;
+        Event_RevealCards revealEvent;
+        revealEvent.set_zone_name(zone->getName().toStdString());
```
ZeldaZach, line 348: Same potential concern here

ZeldaZach (approved):
Approved, thank you very much for the help! I spotted a vuln that is several years old and will file a ticket accordingly. :)

tooomm:
>I spotted a vuln that is several years old and will file a ticket accordingly. :)

Reminder. I think you did not get around to that yet.

ebbit1q:
this vulnerability does not exist, the player only has access to their own zone still, if we were to add a different library zone with hidden information it'd be handled the same way and still not cause any problems.

tooomm:
@ebbit1q Thanks for clarifying and dispelling the initial concerns. 👌

PR #4244, opened 2021-02-06
don't sort split card halves alphabetically

### LINKED ISSUE #4241: Adventure cards not registering as creatures
About 1/3 of creatures with an adventure are registering as having instant or sorcery as their primary card type, rather than creature. This isn't a huge issue in practice, but there are a few minor, odd things this does--if on the stack, clicking them again sends them to the graveyard, rather than the battlefield; the power & toughness entry while on the battlefield is listed as, for example, " // 5/5", rather than just "5/5", which in turn means that adding or subtracting toughness treats this score as being base 0/0. This is also a bit of a nuisance in the Deck Builder, where these cards aren't listed as creatures.
The following, at least in my client, are running into this issue:
Embereth Shieldbreaker
Giant Killer
Lovestruck Beast
Merchant of the Vale
Oakhame Ranger
Order of Midnight
Realm-Cloaked Giant
Reaper of Night
Rimrock Knight
Silverflame Squire
Smitten Swordmaster
Tuinvale Treefolk

ebbit1q:
Split cards are sorted alphabetically before merging, resulting in the lower value half being the top/left/first side.
This affects aftermath, adventure and normal split cards.
examples:
giant killer // chop down
failure // comply
smelt // herd // saw

Instead the order in which they appear in mtgjson is used, which happily coincides with the order in which they should be except for adventures which we reverse in order.

fixes #4241

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -200,7 +200,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet,
     static const QMap<QString, QString> identifierProperties{{"multiverseId", "muid"}, {"scryfallId", "uuid"}};
 
     int numCards = 0;
-    QMultiMap<QString, SplitCardPart> splitCards;
+    QMap<QString, QList<SplitCardPart>> splitCards;
     QString ptSeparator("/");
     QVariantMap card;
     QString layout, name, text, colors, colorIdentity, maintype, power, toughness, faceName;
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):

TheSilentE:
The reverse issue is happening to me now. All of the adventures except the ones mentioned above are sorted the wrong way round

tooomm:
Which version of Cockatrice are you using?

Did you also update your card database after reinstalling the app?

TheSilentE:
version 2.9.0 for Windows, and yeah I downloaded a fresh card database as well

ebbit1q:
this is a merged pull request, please make a new issue and link #4241

PR #4258, opened 2021-02-17
temporary workaround for arch libc version requiring upgraded host

ebbit1q:
see https://bugs.archlinux.org/task/69563#comment196582

Arch linux builds have some incompatibility when in containers with some host systems, while this isn't a problem for builds it just doesn't look very nice.

### DIFF
--- a/.github/workflows/ci-builds.yml
+++ b/.github/workflows/ci-builds.yml
@@ -78,13 +78,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distro: # these names correspond to the files in .ci/$distro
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Arch still seems unhappy

ebbit1q:
yeah, it wants the service to be reinitialized somehow but it intentionally disables it when I install the other runc

ebbit1q:
my inability to read prevented me from noticing I of course need to install an actually updated version of runc instead of the normal one

ZeldaZach (approved):

Issue #4259, opened 2021-02-17
View library with large decks


MotherofRuins:
When someone views their library, if the library is very large (eg, 4000 cards), it can take a couple minutes to display.  This action is blocking.  Generally with these decks we try to avoid F3 and just make a token with the  name of the card we would search for, but sometimes we hit F3 without thinking, and then Cockatrice freezes up for a while.  In a local game it's inconvenient but in an online game it annoys the other players.  One wants a way to break out.  ESC would be a typical abort key.

Daenyth:
Likely area of interest: [ZoneViewWidget](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/zoneviewwidget.cpp)

I'm betting it has something to do with sorting the view

ebbit1q:
making 4000 card deck I guess...

Daenyth:
Would be good to check on number of cards vs number of *distinct* cards, and how that impacts it

ebbit1q:
I err, had a way to automate this and now spent more time to find it than it would to make again...

ebbit1q:
here
```bash
cards="$HOME/.local/share/Cockatrice/Cockatrice/cards.xml"
pre='<?xml version="1.0" encoding="UTF-8"?>
<cockatrice_deck version="1">
    <deckname>big deck</deckname>
    <comments></comments>
    <zone name="main">'
end='    </zone>
</cockatrice_deck>'
linepre='        <card number="1" name="'
lineend='"/>'
first="<cards>"
regex='<name>(.*)<\/name>'

echo "$pre"
{
  while read "line" && [[ ! $line =~ $first ]]; do
    :
  done
  while read "line"; do
    [[ $line =~ $regex ]] && echo "$linepre${BASH_REMATCH[1]}$lineend"
  done
}<"$cards"
echo "$end"
```

MotherofRuins:
I can send you a 4000 card deck if you like.  Also it's singleton.

ebbit1q:
> I can send you a 4000 card deck if you like. Also it's singleton.

no need, the script up there automatically generates a 20.000 card list

MotherofRuins:
The big deck displays with F3 instantly, thank you.

ebbit1q:
please don't close the issue until this pull request is merged

PR #4262, opened 2021-02-20
implement a maximum amount of cardmenus generated for views

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4259

Opening a large deck in a view results in a long wait while it connects every single action on the card which is expensive af, the smart way to do this would be to only populate the menu when needed but this works for now....

- only the first 300 cards in card views will get card actions, other cards will have to be dragged to the top of the deck in order to populate the right click menu now
- pressing f3 with a deck containing every card in mtg will not take your soul
- starting the game with that deck will still take forever

also as a note, testing reveals all the bottleneck is concentrated on card menu creation, servers etc do not experience any load at all.

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -18,9 +18,15 @@
 #include <QMenu>
 #include <QPainter>
 
-CardItem::CardItem(Player *_owner, const QString &_name, int _cardid, bool _revealedCard, QGraphicsItem *parent)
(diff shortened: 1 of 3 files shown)


MotherofRuins:
This should meet my needs, thanks.

ebbit1q:
> This should meet my needs, thanks.

you can download this version and use it online right now, our build system automatically creates installers for every pull request: https://github.com/Cockatrice/Cockatrice/actions/runs/583207427

at the bottom of the page you can select the artifact matching your os

ZeldaZach (approved):

PR #4269, opened 2021-02-24
[webclient] remove duplicate services files

mixmix:
Reading through the repo I saw this : 

```
src/websocket
├── commands
├── events
├── index.ts
├── instanceService
├── ProtoFiles.tsx
├── services
│   ├── index.ts
│   ├── NormalizeService.tsx
│   ├── RoomService.tsx            <<<
│   └── SessionService.tsx         <<<
├── utils
├── WebClient
│   └── services
│       ├── RoomService.tsx        <<< duplicate?
│       └── SessionService.tsx     <<< duplicate?
└── WebClient.tsx
```

They appear to be accidental duplicates, and removing them doesn't seem to cause any problems

PR #4270, opened 2021-02-25
GitHub Actions: Exclude webclient changes from builds

tooomm:
## Related Ticket(s)
- Related #4269

Any changes on the Cockatrice webclient triggered our long CI build runs only targeting the Cockatrice desktop app.

- Exclude changes in the `webclient` folder

### DIFF
--- a/.github/workflows/ci-builds.yml
+++ b/.github/workflows/ci-builds.yml
@@ -6,6 +6,7 @@ on:
       - master
     paths-ignore:
       - '**.md'
+      - 'webclient/**'
     tags:
       - '*'
   pull_request:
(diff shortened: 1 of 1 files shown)


mixmix:
This is great!
I think it means we can also safely have other actions which pay attention to _only_ to the web client

The git histories will still be intermingled I think could get messy if we ever need to time travel tho.
I think monorepos could help with this though

tooomm:
>I think it means we can also safely have other actions which pay attention to only to the web client

Exactly.

>The git histories will still be intermingled I think could get messy if we ever need to time travel tho.
I think monorepos could help with this though

I think the reason to start with it in this repo was the `/common` folder with the shared protocol code.

PR #4273, opened 2021-02-26
disallow rich text in deck comments

### LINKED ISSUE #4255: Deck comments allow to paste text formatting
**System Information:**
Client Version: 2.8.1-custom(efc3f98) (2021-02-09)
Build Architecture: 64-bit
Qt Version: 5.12.9

Deck comments should transform any input to normal text, same like the deck name field works.
The issue here is that this format change locks in and any freshly typed input afterwards keeps it.
comments

**Steps to reproduce:**
 - Copy any non default layout text like bold and bigger font size to the clipboard
 - Paste it to your deck comments
 - Try to write normal text, it will all show in bold and big letters

ebbit1q:
## Related Ticket(s)
- Fixes #4255

comments allow rich text formatting when pasted into

- turns it off


tooomm (approved):
Thanks for the fix!

PR #4292, opened 2021-03-21
fix off-by-one maxUsers check on session init

zachreizner:
The returned number of users from `getUsersWithAddress` will include the already connected user. The predicate `>= maxUsers` is incorrectly assuming that the new user is not already counted by `getUsersWithAddress`. This change corrects this off-by-one error by only closing connections after their are strictly too many users.


ZeldaZach:
Ahh, nice catch! Thanks for the quick fix :)

PR #4293, opened 2021-03-24
add more info to dialogs

ebbit1q:
adds descriptive strings to the register, password reset request,
password reset challenge request, password reset token dialogs
adds tip to set manager to use ctrl a to select all sets
change sizes in set manager
moves default server info to settings instead of having it hardcoded in
each dialog

default size of sets manager is now this
image

### DIFF
--- a/cockatrice/src/dlg_manage_sets.cpp
+++ b/cockatrice/src/dlg_manage_sets.cpp
@@ -131,12 +131,12 @@ WndSets::WndSets(QWidget *parent) : QMainWindow(parent)
     labNotes->setOpenExternalLinks(true);
     labNotes->setText(tr("Use ctrl+a to select all sets in the view.") + "<br><b>" + tr("Deck Editor") + ":</b> " +
                       tr("Only cards in enabled sets will appear in the card list of the deck editor.") + "<br><b>" +
-                      tr("Card Art") + ":</b> " + tr("Image priority is decided in the following order:") + "<ol><li>" +
-                      tr("CUSTOM Folder") +
-                      " (<a href='https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-%26-Sets"
-                      "#to-add-custom-art-for-cards-the-easiest-way-is-to-use-the-custom-folder'>" +
-                      tr("How to use custom card art") + "</a>)</li><li>" + tr("Enabled Sets (Top to Bottom)") +
-                      "</li><li>" + tr("Disabled Sets (Top to Bottom)") + "</li></ol>");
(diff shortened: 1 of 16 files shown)


cockatrice/src/dlg_manage_sets.cpp
```diff
@@ -128,16 +129,17 @@ WndSets::WndSets(QWidget *parent) : QMainWindow(parent)
     labNotes->setWordWrap(true);
     labNotes->setTextInteractionFlags(Qt::TextBrowserInteraction);
     labNotes->setOpenExternalLinks(true);
-    labNotes->setText("<b>" + tr("Deck Editor") + ":</b> " +
-                      tr("Only cards in enabled sets will appear in the card list of the deck editor") + "<br><br>" +
-                      "<b>" + tr("Card Art") + ":</b> " + tr("Image priority is decided in the following order") +
-                      "<ol><li>" + tr("CUSTOM Folder") +
+    labNotes->setText(tr("Use ctrl+a to select all sets in the view.") + "<br><b>" + tr("Deck Editor") + ":</b> " +
```
tooomm, line 132:
```suggestion
    labNotes->setText(tr("Use CTRL+A to select all sets in the view.") + "<br><b>" + tr("Deck Editor") + ":</b> " +
```

<br>

Regarding all hintsGrid:
Don't you think the extra line breaks were useful?
Because of how lists are displayed you now have two blocks, but the second block being bullet points of the last element of the first block. The connection and context is not as clear now as seen in your screen.

Yours:


Compared to what is currently in client:
hints

ebbit1q: I wanted to make it smaller, it's taking up a lot of space

tooomm: I see, sure. Both not ideal. :)

ebbit1q:
image
I made the dialog hint a little smaller again

ebbit1q:
in order to make the list smaller I stopped disabled sets from being used for image art, this resulted in a >1% performance increase as shown with the added timer

ebbit1q:
I removed a few needless entries to the server dialog settings, among which a password entry that saves the password you used last in the register or recover dialog, whether you want it or not............
None of these entries are used anymore since the new login dialog was added, even though they all accounted for entries to be used by the old dialog they now do nothing.
Its possible to adjust these changes to try to mess with the new login dialog but imo it's not worth it because it doesn't in the current state.
I also fixed the remove function, which used to remove an entry but not move all other entries back by one.

ZeldaZach (approved):

PR #4295, opened 2021-03-25
remove arch workaround

ebbit1q:
workaround introduced in #4258 
the package used seems to no longer be live, first testing if it can be removed entirely before updating it.


ebbit1q:
the issue seems to have been resolved, this is ready to merge

Issue #4290, opened 2021-03-20
judges share hidden card information while taking actions on unknown cards


ebbit1q:
judges forcing players to take an action share results in the server giving the player the name of the hidden card

this happens when peeking at cards or when taking a card from their library

in addition, the judge taking the action isn't given any information at all

_______________________________________________________________________________________

**Steps to reproduce:**
 - create a server then create a game spectating as judge, this is easy to do by editing the server config to allow this since #4281 (joining as judge is done by holding shift)
 - have two players join and start a game, this can be done by just launching the client multiple times and having them join the local server
 - on the judge client, look at one of the players' libraries then grab a card from it face down by dragging it and holding shift
 - peek at the card face

both when the card is taken and when peeking, the card name is leaked

PR #4297, opened 2021-03-26
send hidden info to judge instead of player

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4290

judges can force an action to be taken by a player, this will tell the player instead of the judge any hidden information in generated event messages

- the private event message is instead sent to the judge

this has been tested with the judge as spectator.
this is a server only change, older clients are compatible, no part of the actual protocol is changed.
this change displays some glaring issues with the judge mode, such as how things are shown to players in the message log and in corner cases the judge's client does not correctly handle clearing memory and might crash.
I have decided these issues are client side and outside the scope of this change.

### DIFF
--- a/common/server_response_containers.cpp
+++ b/common/server_response_containers.cpp
@@ -55,9 +55,11 @@ void GameEventStorage::sendToGame(Server_Game *game)
 
     GameEventContainer *contPrivate = new GameEventContainer;
     GameEventContainer *contOthers = new GameEventContainer;
+    int id = privatePlayerId;
     if (forcedByJudge != -1) {
         contPrivate->set_forced_by_judge(forcedByJudge);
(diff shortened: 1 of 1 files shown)


ebbit1q:
reproduction steps for crash (caused on client side, I'll include this in a ticket after it this is merged)
1. create game with judge as spectator
2. have a player join, and start a game
3. the judge views the player's library
4. the player leaves the game
4. alternatively, the player concedes and starts the game again

this will lead to a crash in the code of the card view zone, it's therefore likely this is caused by earlier changes to card views https://github.com/Cockatrice/Cockatrice/pull/4263

ZeldaZach (approved):

PR #4302, opened 2021-04-01
Fix 4294

### LINKED ISSUE #4294: Prefill forgot password dialog with info from selected server
The `Forgot Password` dialog should know which server is currently selected and populate the related+stored server and user information to its dialog when opened to request a new password.
Currently, it's always pre-filled with the last successfully connected server, not the actual selection.

Join server A
Disconnect
Open Connect dialog
Server A will be preselected
Select server B
Join server B
Disconnect
Open Connect dialog
Server B will be preselected
Select server A
Click Forgot Password
It will be prefilled with the details from server B (last connected one) + your user name from server B

It should have all info from server A instead as that is your selection.
Users can not even copy and paste the server data from the connect dialog, one has to type them which is annoying and an easy source for errors.

This is probably a leftover when the connect dialog got changed ~2 years ago. Before, the `Forgot Password` button was more general and not as connected to the login fields which are now tied to the server selection as well.

ebbit1q:
## Related Ticket(s)
- Fixes #4294

restoring passwords is hard

- restore password will use the currently entered information in the connect dialog
- a menu item that starts the restore password dialog directly is added

### DIFF
--- a/cockatrice/src/dlg_connect.cpp
+++ b/cockatrice/src/dlg_connect.cpp
@@ -347,6 +347,11 @@ bool DeleteHighlightedItemWhenShiftDelPressedEventFilter::eventFilter(QObject *o
 
 void DlgConnect::actForgotPassword()
 {
+    ServersSettings &servers = SettingsCache::instance().servers();
+    servers.setFPHostName(hostEdit->text());
+    servers.setFPPort(portEdit->text());
+    servers.setFPPlayerName(playernameEdit->text().trimmed());
+
     emit sigStartForgotPasswordRequest();
     reject();
 }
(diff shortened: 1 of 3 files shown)


tooomm:
>    restore password will use the currently entered information in the connect dialog

👍 

>    a menu item that starts the restore password dialog directly is added

We have now two names in the client for the same thing.
The old button and the dialog is labelled `Forgot Password`, the new option is labelled `Restore Password`. Not ideal for users and translations.

Not so sure if that additional menu entry actually helps that much.
If you think the current button is difficult to find, it might be better to improve that or keep it in mind with the new and redesigned combined connect/register dialog.

ebbit1q:
> We have now two names in the client for the same thing.
> The old button and the dialog is labelled `Forgot Password`, the new option is labelled `Restore Password`. Not ideal for users and translations.
> 
> Not so sure if that additional menu entry actually helps that much.
> If you think the current button is difficult to find, it might be better to improve that or keep it in mind with the new and redesigned combined connect/register dialog.

"forgot password" on a button is not correct english, the button should express the action that is taken, not the situation

tooomm:
Do you sichert

ebbit1q:
either works

ebbit1q:
feel free to pr that

PR #4318, opened 2021-04-18
change number dialog defaults

ebbit1q:
all dialogs in the player now have a sane maximum and retain their last entered value.


tooomm (approved):
👍

PR #4324, opened 2021-04-26
Closes all library views when shuffled (#4261)

### LINKED ISSUE #4261: Shuffling invalidates all views but doesn't notify them
When you have a card view window open and shuffle you can still manipulate the view and grab cards that now have different ids that will be treated as the original card, eg pulling a card with a p/t set will set the p/t on the card that is pulled but because it is a different card now the p/t will not match.

**Steps to reproduce:**
 - Load this deck: 
 ```
1 savannah lions
100 plains
```
 - Start a local game with this deck
 - Press f3 to view the library
 - Press ctrl + s to shuffle
 - Drag savannah lions into play
 - 99% chance you will place a 2/1 plains into play
 - Note that the view will have updated and removed the lions, reopening it reveals it never left

marsnicholas:
## Related Ticket(s)
- Fixes #4261 

Library views would be invalidated when a player shuffled

The library views will be closed on when the library is shuffled


ebbit1q (approved):
works as intended.

does not send the close message to the server, this message is pointless anyway as clients can send this message whenever they'd like (I'm going to remove it)

all views are closed when cards that aren't viewed are randomized, this is less than ideal but not a dealbreaker but it should be possible to check which cards are viewed and which cards are being shuffled at that moment. (clients shuffle parts of their library when they move cards to the bottom of their library in a random order)

tooomm:
Does it make sense to add e.g. '.vc' to the gitignore file to prevent local detection?

ebbit1q:
> Does it make sense to add e.g. '.vc' to the gitignore file to prevent local detection?

yes, I asked them to do so through discord

PR #4327, opened 2021-04-30
open decks and replays from file system

djpiper28:
uses qlocalserer and qlocalpeer with filelocks to make clicked links in file system open in the active cockatrice instance. requires the mime type to be installed to work. linux only at the moment as windows requires registry changes to get it to be recognised as the default application for .cor/.cod files. all platforms should work but would need the user to manually select it.

## Related Ticket(s)
none

Replays and decks must be opened in the client, this adds a feature to open them from the file system.

- added qtlocalpeer which is the backend to the inter process communication
- added the code to parse the arguments that are parsed by xdg-open (separate from the current argument parser)
- the application instance manager is a short wrapper class from qbittorrent which makes the qtlocalpeer easier to work with
- added the mime type definition and the script to install the mimes on a linux system

## Screenshots

https://youtu.be/LMScvDbK4a8 is a video of it working


cockatrice/src/dlg_connect.cpp
```diff
@@ -332,6 +332,36 @@ QString DlgConnect::getHost() const
     return hostEdit->text().trimmed();
 }
 
+void DlgConnect::setServer(QString serverName, QString address, unsigned int port)
```
djpiper28, line 335: this code should not be here my bad

ebbit1q:
:eyes: looks like it actually compiled

djpiper28:
It does compile but I don't like using th home folders name for the instance id. But it does compile so I might not change it.

djpiper28:
> How's this coming along? Is it ready to be tested?

Yeah it works, however the file association has to be setup manually.

PR #4338, opened 2021-05-09
add werror flags when making a debug build using llvm

ebbit1q:
this would get bugs like
https://github.com/Cockatrice/Cockatrice/pull/4337
get signalled earlier to us

macos builds do not fail on warnings, this is a useful feature that can give an advance warning of a bug that would be exclusive to macos

- adds the same flags used for gcc to llvm, hoping they do the same thing


ebbit1q:
please, undo that

it didn't build correctly

PR #4345, opened 2021-05-10
do not edit the zone currently iterated on

ebbit1q:
## Related Ticket(s)
- reverts #4340 and  #4341 which were fixes that didn't work

unattaching cards invokes the moveCard function which changes the
internal cards list

this can cause the iterator to become invalidated which will crash but
because of the data not always being moved it will often still work as
intended, giving the idea that it is random

- cards will no longer be unattached when a player leaves play if both cards are controlled by the same player
  - this is unnoticeable to clients as in the same event that player will have all their cards deleted regardless


ZeldaZach (approved):
This makes sense

PR #4347, opened 2021-05-12
Fix for poor performance with large decks

### LINKED ISSUE #4284: improve performance on large decks
**Similar Requests**
https://github.com/Cockatrice/Cockatrice/issues/4259
https://github.com/Cockatrice/Cockatrice/issues/67

there is an opportunity to improve the performance of loading decks in the player by delaying or optimizing how card menus are generated.

**Additional Context**
by my estimate of just stopping the client to see what it is doing while loading decks and noticing the improvement when disabled in the code, card menus are responsible for more than 90% of the time spent when loading the deck.
this time is spent on allocating QAction objects and initializing them as each card has its own menu with a unique set of actions that the card can take.
this menu should instead be either shared between multiple cards or simply generated when the card is interacted with.
this isn't limited to loading the deck either, a new card object with a new set of qactions is created for each card in every cardview as well.
this issue isn't very noticeable until you load about a thousand cards but any improvement here could lead normal sized decks to load in a bit snappier as well, not to mention the fact this is basically just wasting a lot of cpu cycles.

for testing it does not matter if the cards are unique, simply making a deck of 1000 of the same card should suffice.
relevant code is found here https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/player.cpp#L3026

djpiper28:
-> The menus have the update menu thing emitted when they get triggered.
| -> works surprising well https://youtu.be/KOOmhxvHA2c is a demo on a 10000ish card deck

## Related Ticket(s)
- Fixes #4284

The client was slow when opening large decks for games or having large amounts of cards played. The bottleneck was the action menu being updated,

- Made the menus update when they get triggered.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -3301,8 +3301,7 @@ void Player::updateCardMenu(const CardItem *card)
 {
     /**
      * Regarding issue https://github.com/Cockatrice/Cockatrice/issues/4284
-     * The best fix is to add the QActions when a card is right clicked on,
-     * so this method is not needed
+     * The best fix is to add the QActions when a card is right clicked on
      */
     // If bad card OR is a spectator (as spectators don't need card menus), return
     if (card == nullptr || (game->isSpectator() && !judge)) {
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1946,7 +1946,6 @@ void Player::eventSetCardCounter(const Event_SetCardCounter &event)
 
     int oldValue = card->getCounters().value(event.counter_id(), 0);
     card->setCounter(event.counter_id(), event.counter_value());
-    updateCardMenu(card);
     emit logSetCardCounter(this, card->getName(), event.counter_id(), event.counter_value(), oldValue);
 }
 
@@ -3295,15 +3294,16 @@ void Player::refreshShortcuts()
 {
     if (shortcutsActive) {
         setShortcutsActive();
-
-        for (const CardItem *cardItem : table->getCards()) {
-            updateCardMenu(cardItem);
-        }
     }
 }
 
 void Player::updateCardMenu(const CardItem *card)
 {
+    /**
+     * Regarding issue https://github.com/Cockatrice/Cockatrice/issues/4284
+     * The best fix is to add the QActions when a card is right clicked on,
+     * so this method is not needed
+     */
     // If bad card OR is a spectator (as spectators don't need card menus), return
     if (card == nullptr || (game->isSpectator() && !judge)) {
         return;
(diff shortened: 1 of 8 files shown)


ebbit1q (changes requested):
this looks promising, while testing I noticed:

the menu for cards in play is suddenly used for cards on the stack

cards in views no longer have menus at all

ebbit1q:
also note that because this is so fast, there is no need anymore to limit the amount of menus as implemented here https://github.com/Cockatrice/Cockatrice/pull/4262

ebbit1q (changes requested):
this now works great on all the cards not in views but all cards in views still do nothing

ebbit1q (changes requested):
another problem is that all card actions are also bound to keybinds, if the action isn't initialised yet all keyboard actions on the card fail, a way to remedy this would be to initialise the menu when the card is interacted with in any way

ebbit1q:
ugh, I started looking into what is going on here and fixing things.

I found some bigger problems going on with card selection, expect me to push a change soon.

ebbit1q:
so this is a rather annoying bug that is related but not the focus of this pr, if this bug were fixed it would mean this pr would be finished.

however, it's not really the scope of this pr, I was thinking of making a separate ticket for it but I am worried that it might conflict with this change.

djpiper28:
Does that still update shortcuts when the selected card is Zone moved?

ebbit1q:
yes, it technically also does if it is moved inside a zone

ebbit1q:
I made a ticket for it https://github.com/Cockatrice/Cockatrice/issues/4372

cockatrice/src/player.cpp
```diff
@@ -3538,18 +3537,15 @@ void Player::addRelatedCardActions(const CardItem *card, QMenu *cardMenu)
...
     }
 }
 
 QMenu *Player::getCardMenu() const
 {
-    if (aCardMenu) {
-        return aCardMenu->menu();
-    }
-    return nullptr;
+    return aCardMenu->menu(); // can return nullptr
```
ZeldaZach, line 3548: I don't like the removal of a NP check

ebbit1q (approved):

cockatrice/src/player.cpp
```diff
@@ -3545,7 +3545,11 @@ void Player::setCardMenu(QMenu *menu)
 
 QMenu *Player::getCardMenu() const
 {
-    return aCardMenu->menu(); // can return nullptr
+    if (aCardMenu != nullptr) {
+        return aCardMenu->menu();
+    } else {
```
ZeldaZach, line 3550: Can just revert this function to what it was, no need for else

ZeldaZach (approved):
Tested and good to go!

PR #4351, opened 2021-05-14
Templates: Add Transifex link

tooomm:
## Short roundup of the initial problem
Not super easy to get involved.

## What will change with this Pull Request?
- Add link to our transifex project page which handles all strings and languages.

PR #4354, opened 2021-05-15
it is not possible to add a link to the wiki to this description

ebbit1q:
https://github.com/Cockatrice/Cockatrice/pull/4351 added this but it does not work, nor is there another way to do this it seems


ebbit1q:
@tooomm we should add a link to the wiki to our transifex page

tooomm:
@ebbit1q html tags also not working?

I can see if transifex allows custom links.

ebbit1q:
I guess it's something

PR #4356, opened 2021-05-15
properly enable notification checkboxes

### LINKED ISSUE #4355: Can't uncheck "notify in taskbar when users in buddy list connect" box
**System Information:** Mac OS X Big Sur

When attempting to disable these notifications, I found that the notification checkbox was greyed out.

image

Checking the other two boxes did not cause it to become checkable. I did manage to resolve the issue by checking both of the other two boxes, then restarting the application, at which point the third option ceased to be greyed out and I was able to uncheck it normally.

**Steps to reproduce:**
 - Do A
 - Do B
 - Do C ...

Unfortunately, I do not know how to reproduce this bug.

ebbit1q:
## Related Ticket(s)
- Fixes #4355

when unchecking and checking the enable notifications checkbox in the settings the checkbox for notifications for friends connecting misbehaved

- enable and disable the checkbox properly
- rename the function because it is horribly confusing
- uncheck both checkboxes when the main checkbox is unchecked

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -411,7 +411,7 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     notificationsEnabledCheckBox.setChecked(SettingsCache::instance().getNotificationsEnabled());
     connect(&notificationsEnabledCheckBox, SIGNAL(stateChanged(int)), &SettingsCache::instance(),
             SLOT(setNotificationsEnabled(int)));
-    connect(&notificationsEnabledCheckBox, SIGNAL(stateChanged(int)), this, SLOT(setSpecNotificationEnabled(int)));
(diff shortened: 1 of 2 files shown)


tooomm:
>uncheck both checkboxes when the main checkbox is unchecked

I'm not so sure if that is the expected behavior.
Just because the user disables the parent checkbox and turns all notification feature off, he probably doesn't expect any custom settings to get altered in the process. Now, when he decides to use notifications again he has to check and configure all related settings again.

Compare it to the sound setting:
If you use a custom master volume at 60/100. Nobody expects that de-selecting the `Enable sounds` checkbox to change that and make it go back to e.g. 100 when you select it again to enable sounds.

ebbit1q:
I'm aware of that but I don't think it is worth worrying about, this solution is clear to the user even if they have to expend the few extra clicks

tooomm (approved):
Other than that it does what it says and fixes an issue.

PR #4361, opened 2021-05-18
Cleanup and refactor

seavor:


ZeldaZach:
cc: @jchamish @mixmix @Shmiddty @ebbit1q

ZeldaZach:
cc: @jkachurek

webclient/src/components/ThreePaneLayout/ThreePaneLayout.css
```diff
@@ -22,12 +22,19 @@
   flex-shrink: 0;
 }
 
-.three-pane-layout .grid-main__top.fixedHeight {
-  height: 50%;
+.three-pane-layout .grid-main__bottom {
```
mixmix, line 25: This selector is already defined below. Merge with that

mixmix (approved):
Seems like a fairly straightforward renaming.

Not sure why there are style changes in here as well. For future reference that makes this harder to review - I've not spun this up to see how it looks

seavor:
> Seems like a fairly straightforward renaming.
> 
> Not sure why there are style changes in here as well. For future reference that makes this harder to review - I've not spun this up to see how it looks

Fair point, this is the result of a live coding session we did, so it has a few things going on.

ZeldaZach (approved):

PR #4378, opened 2021-06-03
add gtest from arch repos so it doesn't have to download it

ebbit1q:
arch builds are failing, even though they are set to optional they will still show a red cross when they fail which is confusing

but, it's easier to try to just fix the build?

- instead of downloading gtest, install it from repos which hopefully isn't broken

### DIFF
--- a/.ci/ArchLinux/Dockerfile
+++ b/.ci/ArchLinux/Dockerfile
@@ -5,6 +5,7 @@ RUN pacman --sync --refresh --sysupgrade --needed --noconfirm \
         ccache \
         cmake \
         git \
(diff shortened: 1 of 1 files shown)


ebbit1q:
so apparently the version in arch repos is more recent, it might be worth adding gtest as a dependency to all our images to avoid this issue in the future https://github.com/Cockatrice/Cockatrice/runs/2732742455#step:6:81 vs https://github.com/Cockatrice/Cockatrice/runs/2732742389#step:6:90

tooomm (approved):
Including gtest for Arch fixes its build.

PR #4379, opened 2021-06-03
Include known dropdown at login for web client

jchamish:
## Related Ticket(s)
- Fixes #IssueNumber
- https://github.com/Cockatrice/Cockatrice/projects/4#card-29964862

This is a new feature to the webpage.

- Will add the dropdown at the login

## Screenshots

Screen Shot 2021-06-03 at 12 14 19 AM


tooomm:
We have a central place where we update server information and the Cockatrice client pulls this information from: https://github.com/Cockatrice/cockatrice.github.io/blob/master/public-servers.json

Webatrice should use the same data source probably instead hard coding certain servers?
The data could have a field for `"webclient": true` if necessary.

Issue #1223, opened 2015-07-07
Add "Restore default location" button for path settings.


eigurrola:
I changed the card database path for testing purposes under the preferences menu. I wanted to go back to my original path
$HOME/Library/Application Support/Cockatrice/Cockatrice/cards.xml

however, when I press the ... button I can't find the Library directory (I am guessing Mac OS has it hidden) and therefore I can't set my original card database path. I tried setting the field manually but it seems the Qt field editing option has been disabled.

I tried looking for a "restore defaults" button but I had no success there.

Any way to fix this?
Thanks!

eigurrola:
I found out that this is something from the Mac OS, to solve this issue do the following:
1. Open a Finder Window
2. Go to you Home holder Go > Home (Shift + Command + H)
3. VIew > Show View Options (Command + J)
4. Check the "Show Library Folder" option near the end

It would be nice to have a restore defaults button though, or letting manual editing.
Have a nice day!

ZeldaZach:
I'll look into this later

tooomm:
It looks like the behavior didn't change as of 2.4.0:
![path_reset](https://user-images.githubusercontent.com/9874850/33320506-c655a1f6-d442-11e7-989e-37c58744bde7.png)

On Windows the `AppData` folder is hidden by default too. So same problem here for inexperienced users as with macOS.

Instead having a button with a "Reset" text label, it could be the same icon-only button we use in the shortcut settings to restore the old value.

ZeldaZach:
This is a major problem still, would love it solved

PR #4384, opened 2021-06-22
add reset paths button to settings

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/1223

there is no option to restore the default settings in the client, short of deleting the settings directory

- allow resetting paths and reloading database when needed
- add a button to the settings
- add a label to the settings that tells you it has been reset

 
## Screenshots

image


ZeldaZach (approved):
While I'd prefer to see individual path resets on each path, this does solve the issue at hand.

PR #4387, opened 2021-06-27
apply chat flood prevention in games next to rooms

ebbit1q:
users can send chat messages without any reasonable character limit to other users and in games, this will instead make it follow the same rules as the main rooms (1000 chars/10 seconds is the default)
this limit is shared over all games and all chat rooms and all their private messages at the same time and isn't very high, but this will stop users from sending too much text quite effectively.
this will sidestep the major performance complications associated with the way we parse for trigger words.

### DIFF
--- a/common/server_abstractuserinterface.h
+++ b/common/server_abstractuserinterface.h
@@ -38,6 +38,7 @@ public:
     }
 
     virtual int getLastCommandTime() const = 0;
+    virtual bool addSaidMessageSize(int size) = 0;
 
     void playerRemovedFromGame(Server_Game *game);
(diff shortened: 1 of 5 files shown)


ebbit1q:
mac debug builds seem to be hitting a similar issue to the 20.04 ubuntu builds where they are using a version of protobuf that triggers the unused parameter waruning https://github.com/Cockatrice/Cockatrice/runs/2927119377#step:8:322
this warning can be safely ignored but it's in a header file and we don't want to ignore the warning in the rest of the project

ZeldaZach (approved):

PR #4388, opened 2021-06-29
#4316: Wording change: CMC --> MV

### LINKED ISSUE #4316: Wording change: CMC --> MV
STX will introduce a rule change regarding terminology.
`Converted Mana Cost (CMC)` will be named `Mana Value (MV?)` starting in one week.

We use it in several places in the UI, as well as the card databases.

@ZeldaZach Will MTGJSON change as well? Scryfall API?
Do we want to change the Spoiler and Token files as well?

mstanaszak:
Add support for MV for UI, no backend changes

## Related Ticket(s)
- Fixes #4316

Simple change for UI and filters for wording change from CMC to MV

### DIFF
--- a/cockatrice/resources/help/search.md
+++ b/cockatrice/resources/help/search.md
@@ -33,7 +33,8 @@ The search bar recognizes a set of special commands similar to some other card d
 <dt><u>Pow</u>er, <u>Tou</u>ghness, <u>C</u>onverted <u>M</u>ana <u>C</u>ost:</dt>
 <dd>[tou:1](#tou:1) <small>(Any card with a toughness of 1)</small></dd>
 <dd>[pow>=8](#pow>=8) <small>(Any card with a power greater than or equal to 8)</small></dd>
-<dd>[cmc=7](#cmc=7) <small>(Any card with a converted mana cost equal to 7)</small></dd>
+<dd>[cmc=7](#cmc=7) <small>(Any card with a mana value equal to 7)</small></dd>
+<dd>[mv=7](#mv=7) <small>(Any card with a mana value equal to 7)</small></dd>
 
 <dt><u>R</u>arity:</dt>
 <dd>[r:mythic](#r:mythic) <small>(Any card that has the mythic-rare rarity)</small></dd>
--- a/cockatrice/resources/help/search.md
+++ b/cockatrice/resources/help/search.md
@@ -30,10 +30,9 @@ The search bar recognizes a set of special commands similar to some other card d
 <dd>[c:c](#c:c) <small>(Any colorless card)</small></dd>
 <dd>[ci:w](#ci:w) <small>(Any card that has white in it's color identity)</small></dd>
 
-<dt><u>Pow</u>er, <u>Tou</u>ghness, <u>C</u>onverted <u>M</u>ana <u>C</u>ost:</dt>
+<dt><u>Pow</u>er, <u>Tou</u>ghness, <u>M</u>ana <u>V</u>alue:</dt>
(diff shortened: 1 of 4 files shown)


cockatrice/resources/help/search.md
```diff
@@ -33,7 +33,8 @@ The search bar recognizes a set of special commands similar to some other card d
 <dt><u>Pow</u>er, <u>Tou</u>ghness, <u>C</u>onverted <u>M</u>ana <u>C</u>ost:</dt>
```
tooomm, line 33:
```suggestion
<dt><u>Pow</u>er, <u>Tou</u>ghness, <u>M</u>ana <u>V</u>alue:</dt>
```

tooomm (commented):
👍 

mv

cockatrice/resources/help/search.md
ZeldaZach: I'm not sure I agree, as we support both. Maybe include Mana Value in this, but leave CMC too

tooomm: It's just the headline of the group and doesn't alter or remove functionality.
In other groups (e.g. format or edition) we do also not list all possible synonyms and options:
cmc

I suggested to rename it here as that's the new default wording replacing cmc.
It's just a matter of time until the mana value term will be more popular and better known.

ebbit1q: mana value should now be the default option with cmc as a backup, we can safely remove cmc from the help file

tooomm: >mana value should now be the default option with cmc as a backup, we can safely remove cmc from the help file

I would not remove all notes of it right away as people might not all be aware or used to yet.
What do you think about removing it from the help file completely after the next main release? That should give our users enough transition time @ebbit1q.

ebbit1q: mana value was chosen because it is more intuitive, I don't think a user would be confused by the change as it converts this "CMC" to normal english

cockatrice/resources/help/search.md
```diff
@@ -33,7 +33,8 @@ The search bar recognizes a set of special commands similar to some other card d
...
 <dd>[tou:1](#tou:1) <small>(Any card with a toughness of 1)</small></dd>
 <dd>[pow>=8](#pow>=8) <small>(Any card with a power greater than or equal to 8)</small></dd>
-<dd>[cmc=7](#cmc=7) <small>(Any card with a converted mana cost equal to 7)</small></dd>
+<dd>[cmc=7](#cmc=7) <small>(Any card with a mana value equal to 7)</small></dd>
+<dd>[mv=7](#mv=7) <small>(Any card with a mana value equal to 7)</small></dd>
```
tooomm, line 37:
```suggestion
<dd>[mv=7](#mv=7) <small>(Any card with a mana value equal to 7)</small></dd>
```

cockatrice/resources/help/search.md
tooomm: >>>we can safely remove cmc from the help file
>>
>>What do you think about removing it from the help file completely after the next main release?
>>That should give our users enough transition time.
>
>I don't think a user would be confused by the change

Sure, accepting both code suggestions should do it then and remove CMC from the help file completely.

ebbit1q (approved):
tested, all good

tooomm:
Sorry for the big delay with this rather small UI change. PR's are piling up recently. 🙈

We are very happy about this adjustment and you taking the time to help! :) Thanks again. @mstanaszak 
Watch out for the next (beta) release with this included.

PR #4397, opened 2021-07-13
initial component setup

seavor:


webclient/src/forms/CardImportForm/CardImportForm.tsx
```diff
@@ -0,0 +1,169 @@
...
+          size={15}
+        />
+      )} />
+    </div>
+  );
+};
+
+const propsMap = {
+  form: FormKey.CARD_IMPORT,
+  onClose: Function
```
seavor, line 159: I couldn't figure out how to get this onClose method passed in from the parent dialog component. You'll notice above I've hijacked the onSubmit method that i'm able to pass in.

```
const { handleSubmit, onSubmit:onClose } = props;
```

webclient/src/services/CardImporterService.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        const tokens = Array.from(dom.querySelectorAll('card')).map(
+          (tokenElement) => this.parseXmlAttributes(tokenElement)
+        );
+
+        return tokens;
+      })
+  }
+
+  private parseXmlAttributes(dom: Element) {
```
seavor, line 45: The final data structure here could probably use some work. I'm trying to accommodate the possibility, but not guarantee, that some values like `reverse-related` or `set` might have 1 or more values. Embedding everything within a `{ value: attrValue}` structure is so that we have some consistency when tags might have attribute values, otherwise we end up with an inconsistent structure of strings and complex objects.

This was just my first pass blazing through the import wizard. The final structure can most certainly be improved.

webclient/src/components/Message/Message.tsx
```diff
@@ -0,0 +1,103 @@
...
+  );
+};
+
+const PlayerLink = ({ name, label = name }) => (
+  <NavLink className="link" to={generatePath(RouteEnum.PLAYER, { name })}>
+    {label}
+  </NavLink>
+);
+
+function parseChunks(chunk, index) {
```
seavor, line 59: This likewise is a bit of a hot mess, but I've done what i could to clean it up, and it's functional.

webclient/src/components/Message/Message.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <ParsedMessage message={message} />
+    </div>
+  </div>
+);
+
+const ParsedMessage = ({ message }) => {
+  const [messageChunks, setMessageChunks] = useState(null);
+  const [name, setName] = useState(null);
+
+  useMemo(() => {
```
seavor, line 30: As expensive as this component may be, at least with useMemo, parsing only ever happens once per component. I don't foresee a message ever being updated within the the lifecycle of a Message component.

webclient/src/components/Message/CardCallout.tsx
```diff
@@ -0,0 +1,87 @@
...
+  },
+}));
+
+const CardCallout = ({ name }) => {
+  const classes = useStyles();
+  const [card, setCard] = useState<CardDTO>(null);
+  const [token, setToken] = useState<TokenDTO>(null);
+  const [anchorEl, setAnchorEl] = useState<Element>(null);
+
+  useMemo(async () => {
```
seavor, line 28: We want to limit DB calls as much as necessary, so we should always wrap component calls to the DB inside a useMemo hook, and only make new calls when input params change.

seavor: I always want to investigate Dexie's useLiveQuery, which is a feature in their alpha release that gives you a query that triggers onChange when the results of that query change (i think).

webclient/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { CardDTO } from 'services';
+
+import './Card.css';
+
+interface CardProps {
+  card: CardDTO;
+}
+
+const Card = ({ card }: CardProps) => {
+  const src = `https://api.scryfall.com/cards/multiverse/${card?.identifiers?.multiverseId}?format=image`;
```
ZeldaZach, line 13: Should use ScryfallID imo as gatherer is not complete. Also, should give user an option, in the future, to pick their download provider

webclient/src/components/CardDetails/CardDetails.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+              {
+                !card.layout ? null : (
+                  <div className='cardDetails-attribute'>
+                    <span className='cardDetails-attribute__label'>Layout:</span>
+                    <span className='cardDetails-attribute__value'>{card.layout}</span>
+                  </div>
+                )
+              }
+            </div>
```
ZeldaZach, line 110: Missing Loyalty, Hand, etc. You can grab them all from  https://github.com/Cockatrice/Cockatrice/blob/141de4628e56f43f0d8d2b73f0589ce2928ec46c/cockatrice/src/game_specific_terms.h#L26

webclient/src/components/CardImportDialog/CardImportDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { CardImportForm } from 'forms';
+
+import './CardImportDialog.css';
+
+const CardImportDialog = ({ classes, handleClose, isOpen }: any) => {
+  const handleOnClose = () => {
+    handleClose();
+  }
+
+  return (
```
ZeldaZach, line 18: There's no clear/load default URL button, but not a priority rn

webclient/src/components/Message/Message.tsx
ZeldaZach: I'd like better variable names / comments as these are some complicated regular expressions

webclient/src/components/Message/Message.tsx
ZeldaZach: Comments are all that's necessary to understand

webclient/src/components/TokenDetails/TokenDetails.tsx
```diff
@@ -0,0 +1,86 @@
...
+              }
+
+              {
+                !props.type?.value ? null : (
+                  <div className='cardDetails-attribute'>
+                    <span className='cardDetails-attribute__label'>Type:</span>
+                    <span className='cardDetails-attribute__value'>{props.type.value}</span>
+                  </div>
+                )
+              }
```
ZeldaZach, line 66: Same list of properties as above comment

webclient/src/forms/CardImportForm/CardImportForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+};
+
+const mapStateToProps = () => ({
+  initialValues: {
+    cardDownloadUrl: 'https://www.mtgjson.com/api/v5/AllPrintings.json',
```
ZeldaZach, line 220: Should handle compression types in the future (XZ, BZ2, GZIP and/or ZIP)

webclient/src/services/DexieService.ts
```diff
@@ -0,0 +1,35 @@
...
+  [Stores.CARDS]: "name",
+  [Stores.SETS]: "code",
+  [Stores.TOKENS]: "name.value",
+};
+
+class DexieService {
+  private db: Dexie = new Dexie('Webatrice');
+
+  constructor() {
+    this.db.version(1).stores(StoreKeyIndexes);
```
ZeldaZach, line 19: You don't seem to do any kind of version checks (yet)

webclient/src/types/server.tsx
```diff
@@ -22,7 +22,7 @@ export enum KnownHost {
 }
 
 export const KnownHosts = {
-  [KnownHost.ROOSTER]: { port: 443, host: 'server.cockatrice.us', },
+  [KnownHost.ROOSTER]: { port: 4748, host: 'server.cockatrice.us', },
```
ZeldaZach, line 25: was this intentional?

ZeldaZach (commented):
A great start, thank you so much!!!

PR #4398, opened 2021-07-23
Prevent users from being able to upload super large files via Sockets

ZeldaZach:
…, which could inadvertently deny access to the server


Daenyth (approved):

ebbit1q (approved):
we should aim to set this as a setting but this can be done in a later pr

PR #4407, opened 2021-08-04
CI: Update Qt version (Windows)

tooomm:
- Update to Qt 5.15.2 and MSVC 2019
- Fix QTDIR var call

### DIFF
--- a/.github/workflows/ci-builds.yml
+++ b/.github/workflows/ci-builds.yml
@@ -333,7 +333,7 @@ jobs:
     runs-on: windows-latest
 
     env:
-      QT_VERSION: '5.12.9'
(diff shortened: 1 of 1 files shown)


ebbit1q:
we should update it to 5.15 if possible, 5.15 will be the longest supported qt5 version and has an important security issue fixed that'd affect you if you wanted to run a server on windows. https://github.com/Cockatrice/Cockatrice/pull/4398

tooomm:
Let's see how this works with msvc and 32bit.

ebbit1q:
I'm not able to test these

tooomm:
I have the 64bit version running without issues since two weeks

ZeldaZach (approved):

PR #4413, opened 2021-08-11
fix card moving on server

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/pull/4326

when looking at cards the server will provide previously unknown information to users without notifying any players

also, when moving cards the cards would be reversed in order

- when a previously unknown card is moved into a position where it would be shown in a private card view, it instead isn't revealed and appears as a blank card
- cards will preserve their order when moved and will have cards on the same row remain sequential instead of going column first

## Screenshots

image

### DIFF
--- a/common/server_cardzone.cpp
+++ b/common/server_cardzone.cpp
@@ -44,6 +44,8 @@ Server_CardZone::~Server_CardZone()
 
 void Server_CardZone::shuffle(int start, int end)
 {
+    cardsBeingLookedAt = 0;
+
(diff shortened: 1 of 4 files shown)


ebbit1q:
one issue I found is that when grabbing the second card you are looking at in a view into play it'll lose known information status, it used to do this correctly but I don't think this is a major issue, I'll try to fix it but this is available for testing to see if no more information is being leaked

ZeldaZach (approved):
Tested and works

PR #4420, opened 2021-08-24
Fixed layout on Deck Editor not using last layout sizes

rrtischler:
It was reseting layout on ctor.

When resizing card info, deck and filters on Deck Editor the application wasn't keeping the layout on app restart. 

- The app will keep layout of Deck Editor widgets on restart.


ZeldaZach (approved):

tooomm:
> >There is similar code in game tab, but it's never called on the ctor.
> 
> on my end it doesn't work properly on games but in the deck editor everything keeps its last size

That seems to be a Linux only issue then, ebbit?

I can confirm as well that changing the layout in a game (e.g. wider card info and messages/log widget) is preserved on Windows.

PR #4447, opened 2021-10-30
add fedora 35 to ci

ebbit1q:
remove fedora 33

fedora 35 will release on november 2nd, it'll be best to already have this building in preparation for its release


ebbit1q:
this didn't build at all, I made it not required instead so we can wait for it to be released that way

PR #4449, opened 2021-10-30
CI: Xcode update, fix Big Sur build

tooomm:
This should fix Big Sur builds.

Xcode 12.5 is discontinued since 2 weeks.

### DIFF
--- a/.github/workflows/ci-builds.yml
+++ b/.github/workflows/ci-builds.yml
@@ -190,7 +190,7 @@ jobs:
           - Debug
           - 10.14_Mojave
           - 10.15_Catalina
-          - 11.0_Big_Sur
+          - 11_Big_Sur
         include:
           - target: Debug # tests only
             os: macos-latest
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
thanks :+1:

tooomm:
Will try the new Xcode 13 as well. 12.5.1 worked 👍

ebbit1q:
hey, I was merging that

ebbit1q:
just pick the longest supported one

ebbit1q:
I see 13 didn't work, I've merged this so it works again on master, if there are further improvements to be made they can be added later

tooomm:
Xcode 13.0 is the default version on the `macos-11` image in between.
But Cockatrice is not compiling against it without changes: https://github.com/Cockatrice/Cockatrice/runs/4055314133?check_suite_focus=true

12.5.1 works for now. There are no real alternatives if we can not use 13 yet and want to stick to macOS 11.

I lost track which version we do want to offer as I never understood the need to provide a binary for macOS 11 in the first place. As far as I'm aware users can use the 10.15 build without limitations.

ebbit1q:
under my understanding building under macos 11 improves performance on new apple hardware

PR #4453, opened 2021-11-01
Support getting a user's password salt via initial websocket connection

ZeldaZach:
This change will allow clientside password hashing and storage, removing the need to store passwords in plaintext on a local user's machine (Ref #344) 

We still need a new hashed_password field on user login that will allow us to directly compare hashes instead of calculating the hash before comparing. Should be fairly simple, just outside the scope of this PR.

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -20,6 +20,7 @@
 #include <QList>
 #include <QThread>
 #include <QTimer>
+#include <QUrlQuery>
 #include <QWebSocket>
 
 static const unsigned int protocolVersion = 14;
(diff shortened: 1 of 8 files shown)


seavor:
Context:
```
Steps for retrieving user salt from server:

- OnConnect, pass username as part of query string
  -- `${protocol}://${host}:${port}?user=${user}`

- OnServerIdentification, return user salt to client
  -- const { serverName, serverVersion, protocolVersion, userSalt } = info;

- OnLogin, client will use the salt to hash the password and submit the login command.
  -- passHash = hash(password, userSalt)
```

ebbit1q:
why not add a message in between?

in my opinion your password salt should not be a part of the server ident, rather it should be requested on login with a new parameter to the login command and a new command that is sent by the server containing your salt. then the client will send the login command again with the hashed password.

having this be consistent is important if we want this feature to be maintainable.

seavor:
The thinking behind this approach is to limit the amount of commands un-authenticated users can make to the server. OnConnect and ServerIdentification are unavoidable, but we don't need to expose another endpoint like getSalt to connected but un-authenticated users. I understand your point, and it has merit. We should discuss the trade offs in a call and record what we decide back here.

PR #4455, opened 2021-11-01
Fixed issue #4332 - changed error message for replay folder download

### LINKED ISSUE #4332: trying the download button on a replay folder gives a confusing error message
**System Information:**
affects all systems and versions from march 2017 onwards

the server places remote replays into a folder structure with one replay per match placed together with all the others in that room. to download a replay users have to expand the folder in the treeview, select the replay, then click the download button. the client nor the server recognizes a command to download all replays in a directory or multiple at the same time. the botton is not disabled and users might try to download a folder still which might lead to confusion. https://github.com/Cockatrice/Cockatrice/pull/2442 introduced an error message to signify that downloading an entire folder is not supported but does not tell people to download the files individually. additionally it uses the phrasing "at this time" alluding to the future inclusion of this feature, which has not happened in 4 years. users interpret this error message as if something went wrong or that it might be a problem at the server while in fact it is simply unimplemented.

**Steps to reproduce:**
 - log in to a server that supports replay storage (all three public servers do on registered accounts)
 - go to the replay tab
 - select a folder of a recently played game and click the download button (second from left at the bottom)
   - if no games have been played recently you can simply create a game lobby, start a 1 person game, and then leave it to create one

gusmccallum:
## Related Ticket(s)
- Fixes #4332

Error message on user attempting replay folder download was unclear.

-Error message when user requests replay folder download has been changed to be more clear.

## Screenshots
image

### DIFF
--- a/cockatrice/src/tab_replays.cpp
+++ b/cockatrice/src/tab_replays.cpp
@@ -197,7 +197,7 @@ void TabReplays::actDownload()
 
     if (!curRight) {
         QMessageBox::information(this, tr("Downloading Replays"),
-                                 tr("You cannot download replay folders at this time"));
+                                 tr("Folder download is not yet supported. Please download replays individually."));
         return;
     }
 


ebbit1q (approved):
thanks for your contribution!

PR #4467, opened 2021-11-11
Respect device pixel ratio when scaling card imgs

### LINKED ISSUE #4466: Properly handle different device pixel ratios (high DPI)
This is most evident with card images which currently look pixelated on retina displays.

I did a bit of hacking around (I'm not a C++ developer so it was very trial and error), and it looks like we just need to multiply the card image size by the device pixel ratio when we generate the scaled image.

I think the right thing is to get the pixel ratio from the current screen.. but I'm not exactly sure what the pitfalls are here / what's best practice. There's a possibility that the user is using multiple displays with different pixel ratios and has some application windows on one and not the other. Maybe rather than using the primary screen we should be getting the screen that the relevant widget is being displayed on? Not sure how straightforward that would be..

We then also need to set the device pixel ratio on the pixmap.

After that it renders beautifully.

Where else do we need to be doing this? I've noticed that the "close" and "pop out window" icons have a similar issue. Not sure where they come from though.

https://github.com/Cockatrice/Cockatrice/issues/1426#issuecomment-962543120

johtso:
## Related Ticket(s)
- Fixes #4466

### DIFF
--- a/cockatrice/src/pictureloader.cpp
+++ b/cockatrice/src/pictureloader.cpp
@@ -6,6 +6,7 @@
 #include "thememanager.h"
 
 #include <QApplication>
+#include <QScreen>
 #include <QCryptographicHash>
 #include <QDebug>
 #include <QDir>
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):

ebbit1q:
thanks for looking into this!
I still have a question though, what happens if the user changes the pixel ratio while the application is running?

johtso:
@ebbit1q good question, I touched on the fact that there might be some unexpected behaviors, or more robust ways to check the right ratio for wherever the widget is being drawn. Unfortunately I don't have any external monitors as the moment to test. Just going based on the primary screen is a basic approach that probably works most of the time.

I _think_ it should fail gracefully though. If the image was incorrectly sized to 2x the necessary number of pixels, but was displayed on a normal dpi screen, it would just get scaled down when it's drawn. I _think_..

PR #4469, opened 2021-11-12
Support HashedPassword workflow for logins

ZeldaZach:


webclient/src/websocket/commands/SessionCommands.ts
```diff
@@ -2,13 +2,13 @@ import {StatusEnum} from 'types';
...
 import webClient from '../WebClient';
-import {guid} from '../utils';
+import {guid, hashPassword} from '../utils';
 import {WebSocketConnectReason, WebSocketOptions} from "../services/WebSocketService";
 import {
   AccountActivationParams,
   ForgotPasswordChallengeParams,
   ForgotPasswordParams,
-  ForgotPasswordResetParams,
+  ForgotPasswordResetParams, RequestPasswordSaltParams,
```
seavor, line 11: Probably from your auto-importer, but put RequestPasswordSaltParams on its own line.

webclient/src/websocket/commands/SessionCommands.ts
```diff
@@ -110,6 +115,31 @@ export class SessionCommands {
...
+    const sc = webClient.protobuf.controller.SessionCommand.create({
+      ".Command_RequestPasswordSalt.ext" : CmdRequestPasswordSalt
+    });
+
+    webClient.protobuf.sendSessionCommand(sc, raw => {
+      if (raw.responseCode === webClient.protobuf.controller.Response.ResponseCode.RespOk) {
+        const passwordSalt = raw[".Response_PasswordSalt.ext"].passwordSalt;
+        SessionCommands.login(passwordSalt);
+      } else if (raw.responseCode === webClient.protobuf.controller.Response.ResponseCode.RespRegistrationRequired) {
+          SessionCommands.updateStatus(StatusEnum.DISCONNECTED, "Login failed: Registration Required");
```
seavor, line 137: I'm a bit confused, does RespRegistrationRequired mean the user is not registered at all, or is awaiting account verification?

webclient/src/websocket/commands/SessionCommands.ts
```diff
@@ -110,6 +115,31 @@ export class SessionCommands {
...
+          SessionCommands.disconnect();
+      }
```
seavor, line 139: Let's do a final else block that just calls disconnect, just to harden against future changes.

webclient/src/websocket/events/SessionEvents.ts
```diff
@@ -124,7 +124,11 @@ function serverIdentification(info: ServerIdentificationData) {
   switch (webClient.options.reason) {
     case WebSocketConnectReason.LOGIN:
       SessionCommands.updateStatus(StatusEnum.LOGGING_IN, 'Logging In...');
-      SessionCommands.login();
+      if (serverOptions & webClient.protobuf.controller.Event_ServerIdentification.ServerOptions.SupportsPasswordHash) {
```
seavor, line 127: Let's leave a comment here identifying this line as an intentional use of a bitwise operator.

webclient/src/websocket/commands/SessionCommands.ts
ebbit1q: respregistration required means that this user is not recognized or eligible for login (they do not have a password in the database) and is not allowed to log in unregistered.

ebbit1q: this check here does not check if the account is awaiting verification, I didn't test that yet

seavor: So really the message we want here is "Invalid Username/Password". For security, you dont want to differentiate between username and password issues for failed logins, unless it's a special case like AccountActivationRequired or something, in which case the issue isnt with the username/password at all, but with the account status itself.

seavor: In other words, if someone attempts to login with an account that doesnt exist, we shouldnt expose that information to them. To the user, attempting to login with an account that doesnt exist is identical to attempting to login with a wrong password.

My initial thought is maybe the server should just return a dummy salt for unknown users.

seavor (approved):

webclient/src/websocket/commands/SessionCommands.ts
ebbit1q: I fully agree with you, yet this is the way it works currently on login, you can ask the server which account names exist anyway if you try to register their names, I don't think changing this is a priority

PR #4478, opened 2021-11-15
CI: Add webclient

tooomm:
The webclient was not part of the CI workflows yet.
This is a first implementation and POC that includes a lint check and test/build commands for a quick feedback in PR's.
It should only run on PR's (or pushes to master) including changes related to the webclient.
It should not run on tags/releases.

*Note:
For this PR, it will include non-web runs as well because I had to do some little updates in the related config files in the process.
Normally it would only include [`Code Style (TypeScript) / ESLint`](https://github.com/Cockatrice/Cockatrice/runs/4223799985?check_suite_focus=true), [`Build Web / React (Node 12)`](https://github.com/Cockatrice/Cockatrice/runs/4223799990?check_suite_focus=true) and [`Build Web / React (Node lts/*)`](https://github.com/Cockatrice/Cockatrice/runs/4223800074?check_suite_focus=true).*

Let me know how you think this can be improved!
Which Node.js versions should be targeted? ~~I just went with 14/16 as a test.~~

---~~

TODO

- [x] run against Node 12 and Node LTS (16)
- [x] update CI badges in README (x2)


ZeldaZach:
We're running Node12 right now, might target 14/16 in the future.

ZeldaZach:
Let me know if/when this gets done, looks good so far!

tooomm:
>Let me know if/when this gets done, looks good so far!

For now, I've no additional input.
While the recent updates to master fixed linting again ([It did broke in another recent PR](https://github.com/Cockatrice/Cockatrice/runs/4272654124?check_suite_focus=true)), the Node builds are still failing for some reason.

We can roll it in and see it in action before we fine tune further.

What do you think @seavor?

.github/workflows/web-build.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          node-version: ${{matrix.node_version}}
+          cache: 'npm'
+          cache-dependency-path: 'webclient/package-lock.json'
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Test app
+        run: npm test
```
seavor, line 51: I just opened a PR that creates an `npm run golden` command that I suggest we use here instead.

seavor, line 51: `npm run test`

PR #4483, opened 2021-11-21
Support Server requests for MFA, Render failed UI statuses to user, C…

ZeldaZach:
…onnect to KnownHosts component

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


webclient/src/components/ResetPasswordDialog/ChangePasswordDialog.tsx
```diff
@@ -10,7 +10,7 @@ import { ResetPasswordForm } from 'forms';
 
 import './ResetPasswordDialog.css';
 
-const ResetPasswordDialog = ({ classes, handleClose, isOpen, onSubmit }: any) => {
+const ChangePasswordDialog = ({ classes, handleClose, isOpen, onSubmit }: any) => {
```
seavor, line 13: I would prefer we keep calling this ResetPassword, ChangePassword will be a different bit of functionality that will require different fields and endpoints.

webclient/src/forms/RequestPasswordResetForm/RequestPasswordResetForm.tsx
```diff
@@ -9,21 +9,42 @@ import { InputField, KnownHosts } from 'components';
...
+  const [errorMessage, setErrorMessage] = useState(false);
+  const [isMFA, setIsMFA] = useState(false);
 
   const onHostChange = ({ host, port }) => {
     dispatch(change(FormKey.RESET_PASSWORD_REQUEST, 'host', host));
     dispatch(change(FormKey.RESET_PASSWORD_REQUEST, 'port', port));
   }
 
+  useReduxEffect(() => {
+    setErrorMessage(true);
```
seavor, line 26: You should probably set this to false onSubmit.

webclient/src/forms/ResetPasswordForm/ResetPasswordForm.tsx
```diff
@@ -9,18 +9,31 @@ import { InputField, KnownHosts } from 'components';
...
+
+
   const onHostChange = ({ host, port }) => {
     dispatch(change(FormKey.RESET_PASSWORD, 'host', host));
     dispatch(change(FormKey.RESET_PASSWORD, 'port', port));
   }
 
+  useReduxEffect(() => {
+    setErrorMessage(true);
+  }, ServerTypes.RESET_PASSWORD_FAILED, []);
```
seavor, line 28: We should inspect the error response object for a fail reason enum or string. Same with the other form failure effect. We can pass those into the action from the websocket layer and pick them up here in the component redux effect.

webclient/src/websocket/commands/SessionCommands.ts
```diff
@@ -326,8 +331,10 @@ export class SessionCommands {
 
     webClient.protobuf.sendSessionCommand(sc, raw => {
       if (raw.responseCode === webClient.protobuf.controller.Response.ResponseCode.RespOk) {
+        SessionCommands.updateStatus(StatusEnum.DISCONNECTED, 'Password successfully updated');
         SessionPersistence.resetPasswordSuccess();
       } else {
+        SessionCommands.updateStatus(StatusEnum.DISCONNECTED, 'Password update failed, please try again');
```
seavor, line 337: If we want to show the error message in the modal, we shouldn't also show it here in the server status. We can talk more about this as a follow up.

webclient/src/forms/ResetPasswordForm/ResetPasswordForm.tsx
ebbit1q: the error response object doesn't contain such information right now, the response is either respOk or respFunctionNotAllowed without details

seavor (approved):

PR #4484, opened 2021-11-24
Add beta hint to release template

tooomm:


ebbit1q:
I don't think this is necessary, either way it's not like our full releases are tested more than our beta releases :/

tooomm:
There is (should be) a difference though. We normally do not merge any new feature (especially bigger changes) and release a full release right after. It goes trough at least a beta first and will see at least some testing there.

Well, in theory. :)

But yeah, it was just a proposal to make people more aware. Also, I do not want to discourage them to use a beta...

ZeldaZach:
I don't find this overly helpful at the current moment, but if things change in the future why don't we revisit our message?

PR #4493, opened 2021-12-01
use hashed passwords in all commands

ebbit1q:
passwords are sent in plain text for the registration, change password and reset password commands

- add hashed password field to commands
- implement server side
- properly split force change password function
- implement client side
- make sure to always log in with hashed password if possible
- add warning when not using hashed password
- add password requirement to change email address when logged in using the hashed password

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -132,7 +132,8 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
             auto passwordSalt = PasswordHasher::generateRandomSalt();
             hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             cmdForgotPasswordReset.set_hashed_new_password(hashedPassword.toStdString());
-        } else {
+        } else if (!password.isEmpty()) {
+            qDebug() << "Warning: using plain text password to reset password";
             cmdForgotPasswordReset.set_new_password(password.toStdString());
         }
         PendingCommand *pend = prepareSessionCommand(cmdForgotPasswordReset);
@@ -161,7 +162,8 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
             auto passwordSalt = PasswordHasher::generateRandomSalt();
             hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             cmdRegister.set_hashed_password(hashedPassword.toStdString());
-        } else {
+        } else if (!password.isEmpty()) {
+            qDebug() << "Warning: using plain text password to register";
             cmdRegister.set_password(password.toStdString());
         }
         cmdRegister.set_email(email.toStdString());
@@ -231,7 +233,10 @@ void RemoteClient::doLogin()
         // TODO add setting for client to reject unhashed logins
         setStatus(StatusLoggingIn);
         Command_Login cmdLogin = generateCommandLogin();
-        cmdLogin.set_password(password.toStdString());
+        if (!password.isEmpty()) {
+            qDebug() << "Warning: using plain text password to log in";
+            cmdLogin.set_password(password.toStdString());
+        }
 
         PendingCommand *pend = prepareSessionCommand(cmdLogin);
         connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(loginResponse(Response)));
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -27,7 +27,7 @@ static const unsigned int protocolVersion = 14;
 
 RemoteClient::RemoteClient(QObject *parent)
     : AbstractClient(parent), timeRunning(0), lastDataReceived(0), messageInProgress(false), handshakeStarted(false),
-      usingWebSocket(false), messageLength(0)
+      usingWebSocket(false), messageLength(0), hashedPassword()
 {
 
     clearNewClientFeatures();
@@ -129,8 +129,8 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
         cmdForgotPasswordReset.set_clientid(getSrvClientID(lastHostname).toStdString());
         cmdForgotPasswordReset.set_token(token.toStdString());
         if (!password.isEmpty() && serverSupportsPasswordHash) {
-            const auto passwordSalt = PasswordHasher::generateRandomSalt();
-            const auto hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
+            auto passwordSalt = PasswordHasher::generateRandomSalt();
+            hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             cmdForgotPasswordReset.set_hashed_new_password(hashedPassword.toStdString());
         } else {
             cmdForgotPasswordReset.set_new_password(password.toStdString());
@@ -158,8 +158,8 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
         Command_Register cmdRegister;
         cmdRegister.set_user_name(userName.toStdString());
         if (!password.isEmpty() && serverSupportsPasswordHash) {
-            const auto passwordSalt = PasswordHasher::generateRandomSalt();
-            const auto hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
+            auto passwordSalt = PasswordHasher::generateRandomSalt();
+            hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             cmdRegister.set_hashed_password(hashedPassword.toStdString());
         } else {
             cmdRegister.set_password(password.toStdString());
@@ -188,13 +188,7 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
         return;
     }
 
-    if (!password.isEmpty() && serverSupportsPasswordHash) {
-        // TODO store and log in using stored hashed password
-        doRequestPasswordSalt(); // log in using password salt
-    } else {
-        // TODO add setting for client to reject unhashed logins
-        doLogin();
-    }
+    doLogin();
 }
 
 void RemoteClient::doRequestPasswordSalt()
@@ -226,21 +220,30 @@ Command_Login RemoteClient::generateCommandLogin()
 
 void RemoteClient::doLogin()
 {
-    setStatus(StatusLoggingIn);
-    Command_Login cmdLogin = generateCommandLogin();
-    cmdLogin.set_password(password.toStdString());
+    if (!password.isEmpty() && serverSupportsPasswordHash) {
+        // TODO store and log in using stored hashed password
+        if (hashedPassword.isEmpty()) {
+            doRequestPasswordSalt(); // ask salt to create hashedPassword, then log in
+        } else {
+            doHashedLogin(); // log in using hashed password instead
+        }
+    } else {
+        // TODO add setting for client to reject unhashed logins
+        setStatus(StatusLoggingIn);
+        Command_Login cmdLogin = generateCommandLogin();
+        cmdLogin.set_password(password.toStdString());
 
-    PendingCommand *pend = prepareSessionCommand(cmdLogin);
-    connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(loginResponse(Response)));
-    sendCommand(pend);
+        PendingCommand *pend = prepareSessionCommand(cmdLogin);
+        connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(loginResponse(Response)));
+        sendCommand(pend);
+    }
 }
 
-void RemoteClient::doLogin(const QString &passwordSalt)
+void RemoteClient::doHashedLogin()
 {
     setStatus(StatusLoggingIn);
     Command_Login cmdLogin = generateCommandLogin();
 
-    const auto hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
     cmdLogin.set_hashed_password(hashedPassword.toStdString());
 
     PendingCommand *pend = prepareSessionCommand(cmdLogin);
@@ -257,12 +260,13 @@ void RemoteClient::passwordSaltResponse(const Response &response)
 {
     if (response.response_code() == Response::RespOk) {
         const Response_PasswordSalt &resp = response.GetExtension(Response_PasswordSalt::ext);
-        QString salt = QString::fromStdString(resp.password_salt());
-        if (salt.isEmpty()) { // the server does not recognize the user but allows them to enter unregistered
-            password = "";    // the password will not be used
+        auto passwordSalt = QString::fromStdString(resp.password_salt());
+        if (passwordSalt.isEmpty()) { // the server does not recognize the user but allows them to enter unregistered
+            password = "";            // the password will not be used
             doLogin();
         } else {
-            doLogin(salt);
+            hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
+            doHashedLogin();
         }
     } else if (response.response_code() != Response::RespNotConnected) {
         emit loginError(response.response_code(), {}, 0, {});
@@ -451,6 +455,7 @@ void RemoteClient::doConnectToServer(const QString &hostname,
     password = _password;
     lastHostname = hostname;
     lastPort = port;
+    hashedPassword = "";
 
     connectToHost(hostname, port);
     setStatus(StatusConnecting);
@@ -473,6 +478,7 @@ void RemoteClient::doRegisterToServer(const QString &hostname,
     realName = _realname;
     lastHostname = hostname;
     lastPort = port;
+    hashedPassword = "";
 
     connectToHost(hostname, port);
     setStatus(StatusRegistering);
@@ -661,6 +667,7 @@ void RemoteClient::doSubmitForgotPasswordResetToServer(const QString &hostname,
     lastPort = port;
     token = _token.trimmed();
     password = _newpassword;
+    hashedPassword = "";
 
     connectToHost(lastHostname, lastPort);
     setStatus(StatusSubmitForgotPasswordReset);
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -267,7 +267,7 @@ void RemoteClient::passwordSaltResponse(const Response &response)
         const Response_PasswordSalt &resp = response.GetExtension(Response_PasswordSalt::ext);
         auto passwordSalt = QString::fromStdString(resp.password_salt());
         if (passwordSalt.isEmpty()) { // the server does not recognize the user but allows them to enter unregistered
-            password = "";            // the password will not be used
+            password.clear();            // the password will not be used
             doLogin();
         } else {
             hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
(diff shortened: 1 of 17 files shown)


ZeldaZach:
Haven't had a chance to review, but please make sure change password requires the original password and not the salt+hashed one. I don't want someone stealing a salt+hash and using it to change passwords

ebbit1q:
that's fair, I will remove that then

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -199,14 +199,18 @@ bool Servatrice_DatabaseInterface::registerUser(const QString &userName,
...
+                                                bool passwordNeedsHash,
                                                 const QString &emailAddress,
                                                 const QString &country,
                                                 bool active)
 {
     if (!checkSql())
         return false;
 
-    QString passwordSha512 = PasswordHasher::computeHash(password, PasswordHasher::generateRandomSalt());
+    QString passwordSha512 = password;
```
ZeldaZach, line 210: Might be more efficient to set this in an if/else?

common/pb/session_commands.proto
```diff
@@ -165,6 +166,8 @@ message Command_AccountPassword {
     }
     optional string old_password = 1;
     optional string new_password = 2;
+    optional string hashed_old_password = 3;
```
ZeldaZach, line 169: This shouldn't be here based on our discussion

servatrice/src/servatrice_database_interface.cpp
ebbit1q: this is simpler, less error prone

common/pb/session_commands.proto
ebbit1q: I havent taken the time to work on this because I'd like to discuss alternatives first, note that you can use the hashed password to log in and change the email address in the current implementation

ebbit1q:
this pr is now based on https://github.com/Cockatrice/Cockatrice/pull/4496 in order to make merging easier

cockatrice/src/remoteclient.cpp
```diff
@@ -27,7 +27,7 @@ static const unsigned int protocolVersion = 14;
 
 RemoteClient::RemoteClient(QObject *parent)
     : AbstractClient(parent), timeRunning(0), lastDataReceived(0), messageInProgress(false), handshakeStarted(false),
-      usingWebSocket(false), messageLength(0)
+      usingWebSocket(false), messageLength(0), hashedPassword()
```
ZeldaZach, line 30: It took me a moment to realize this was a string, maybe "" as default init?

cockatrice/src/remoteclient.cpp
```diff
@@ -127,7 +128,14 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
         cmdForgotPasswordReset.set_user_name(userName.toStdString());
         cmdForgotPasswordReset.set_clientid(getSrvClientID(lastHostname).toStdString());
         cmdForgotPasswordReset.set_token(token.toStdString());
-        cmdForgotPasswordReset.set_new_password(password.toStdString());
+        if (!password.isEmpty() && serverSupportsPasswordHash) {
+            auto passwordSalt = PasswordHasher::generateRandomSalt();
+            hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
+            cmdForgotPasswordReset.set_hashed_new_password(hashedPassword.toStdString());
+        } else if (!password.isEmpty()) {
+            qDebug() << "Warning: using plain text password to reset password";
```
ZeldaZach, line 136: use qWarning() instead of Debug?

cockatrice/src/remoteclient.cpp
```diff
@@ -150,7 +158,14 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
     if (getStatus() == StatusRegistering) {
         Command_Register cmdRegister;
         cmdRegister.set_user_name(userName.toStdString());
-        cmdRegister.set_password(password.toStdString());
+        if (!password.isEmpty() && serverSupportsPasswordHash) {
+            auto passwordSalt = PasswordHasher::generateRandomSalt();
+            hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
+            cmdRegister.set_hashed_password(hashedPassword.toStdString());
+        } else if (!password.isEmpty()) {
+            qDebug() << "Warning: using plain text password to register";
```
ZeldaZach, line 166: qWarning?

cockatrice/src/remoteclient.cpp
```diff
@@ -213,21 +222,33 @@ Command_Login RemoteClient::generateCommandLogin()
...
+            doRequestPasswordSalt(); // ask salt to create hashedPassword, then log in
+        } else {
+            doHashedLogin(); // log in using hashed password instead
+        }
+    } else {
+        // TODO add setting for client to reject unhashed logins
+        setStatus(StatusLoggingIn);
+        Command_Login cmdLogin = generateCommandLogin();
+        if (!password.isEmpty()) {
+            qDebug() << "Warning: using plain text password to log in";
```
ZeldaZach, line 237: qWarning

cockatrice/src/remoteclient.cpp
ebbit1q: well, I just copied this from somewhere in the same file

cockatrice/src/remoteclient.cpp
```diff
@@ -648,6 +672,7 @@ void RemoteClient::doSubmitForgotPasswordResetToServer(const QString &hostname,
     lastPort = port;
     token = _token.trimmed();
     password = _newpassword;
+    hashedPassword = "";
```
ZeldaZach, line 675: if we're just gonna set it empty a bunch of times here, i like my idea of the default "" above

cockatrice/src/remoteclient.cpp
ebbit1q: no, using the default initaliser for this is good

cockatrice/src/userinfobox.cpp
```diff
@@ -199,7 +201,22 @@ void UserInfoBox::actEditInternal(const Response &r)
 
     Command_AccountEdit cmd;
     cmd.set_real_name(dlg.getRealName().toStdString());
-    cmd.set_email(dlg.getEmail().toStdString());
+    if (client->getServerSupportsPasswordHash()) {
```
ZeldaZach, line 204: A client can be modified to remove this check -- Haven't gotten there, but does the server reject the change if the passwordCheck isn't set when the server users hashed passwords and the user fails to provide the plaintext one?

cockatrice/src/userinfobox.cpp
```diff
@@ -215,9 +232,42 @@ void UserInfoBox::actPassword()
...
+void UserInfoBox::changePassword(const QString &oldPassword, const QString &newPassword)
+{
     Command_AccountPassword cmd;
-    cmd.set_old_password(dlg.getOldPassword().toStdString());
-    cmd.set_new_password(dlg.getNewPassword().toStdString());
+    cmd.set_old_password(oldPassword.toStdString());
+    if (client->getServerSupportsPasswordHash()) {
+        auto passwordSalt = PasswordHasher::generateRandomSalt();
+        QString hashedPassword = PasswordHasher::computeHash(newPassword, passwordSalt);
+        cmd.set_hashed_new_password(hashedPassword.toStdString());
```
ZeldaZach, line 267: still user side, do we check to ensure hashed is used if the server supports hashing?

common/pb/session_commands.proto
```diff
@@ -149,6 +150,7 @@ message Command_AccountEdit {
     optional string email = 2;
     // gender = 3; // obsolete
     optional string country = 4;
+    optional string password_check = 100; // password is required to change sensitive information
```
ZeldaZach, line 153: Any reason 100?

common/pb/session_commands.proto
```diff
@@ -164,6 +166,8 @@ message Command_AccountPassword {
     }
     optional string old_password = 1;
     optional string new_password = 2;
+    // optional string hashed_old_password = 3; // we don't want users to steal hashed passwords and change them
```
ZeldaZach, line 169: This wasn't apart of the protocol yet, was it? Why do we comment out and remove 3 from service?

common/pb/session_commands.proto
ebbit1q: nope, it's not in the way and similar to other places we do this

common/pb/session_commands.proto
ebbit1q: in case we'd ever want it, I already implemented it at that point and the comment makes clear we don't want it right now.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -953,30 +979,24 @@ bool Servatrice_DatabaseInterface::changeUserPassword(const QString &user,
...
+    if (!passwordQuery->next())
+        return false;
 
-    passwordQuery = prepareQuery("update {prefix}_users set password_sha512=:password, "
-                                 "passwordLastChangedDate = NOW() where name = :name");
-    passwordQuery->bindValue(":password", passwordSha512);
-    passwordQuery->bindValue(":name", user);
-    if (execSqlQuery(passwordQuery))
-        return true;
+    const QString correctPassword = passwordQuery->value(0).toString();
```
ZeldaZach, line 990: should be correctPasswordSha512, for consistency

cockatrice/src/userinfobox.cpp
ebbit1q: 3 lines above this line...

servatrice/src/servatrice_database_interface.h
```diff
@@ -111,8 +112,12 @@ public slots:
                     LogMessage_TargetType targetType,
                     const int targetId,
                     const QString &targetName);
-    bool
-    changeUserPassword(const QString &user, const QString &oldPassword, const QString &newPassword, const bool &force);
+    bool changeUserPassword(const QString &user, const QString &password, bool passwordNeedsHash);
+    bool changeUserPassword(const QString &user,
+                            const QString &oldPassword,
+                            bool oldPasswordNeedsHash,
```
ZeldaZach, line 118: Any reason oldPWNeedsHash and newPWNeedsHash? Shouldn't one be enough? (I could be wrong)

cockatrice/src/remoteclient.cpp
ebbit1q: this is shorter than QString() which I'd use otherwise

cockatrice/src/userinfobox.cpp
ebbit1q: of course, it returns RespFunctionNotAllowed https://github.com/Cockatrice/Cockatrice/pull/4493/commits/5129f40461e10519e758c8b65e6927327a7f78c1#diff-16bda30986b55ad12ce94d6bd6d89fce0f84793010faa23a893d8194dbf84530R1294

servatrice/src/serversocketinterface.cpp
```diff
@@ -1257,24 +1263,75 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountEdit(const Comma
...
+            checkedPassword = true;
+        } else {
+            // the user already logged in with this info, the only change is their password
+            return Response::RespWrongPassword;
+        }
+    }
+    QString queryText = "update {prefix}_users set ";
+    bool changes = false;
+    if (cmd.has_real_name()) {
+        queryText += "realname=:realName, ";
```
ZeldaZach, line 1285: queries should not be built up like a standard string. If you dont have a value you need, use a null or default. I really don't like this.

servatrice/src/servatrice_database_interface.h
ebbit1q: there is no harm in having both even though the old one is not used right now

servatrice/src/serversocketinterface.cpp
ebbit1q: well, please explain to me how you'd prefer it then.

cockatrice/src/userinfobox.cpp
ZeldaZach: I meant server-side to check that set_hashed_new_password is required, if the server supports hashing

common/pb/session_commands.proto
ZeldaZach: I'm not sure of a situation where we'll ever use it, and the comment isn't explicit on why it's reserved

servatrice/src/servatrice_database_interface.h
ZeldaZach: If it's not used, i'm not sure why we need it exactly. Adds more complexity for no real value

servatrice/src/serversocketinterface.cpp
ZeldaZach: I'm thinking two queries, one to get the current values and one to update. If the user didn't override, then place the current value in the variable. Else use the new value. 

I really don't like query builders with raw strings, it's begging for an issue

ZeldaZach: We can also look into a query builder, maybe like https://github.com/MasterAler/qsqlbuilder ?

ZeldaZach: Spoke with a friend, they recommended using a QStringList to build up the parts, then a `list.join(',')` so we don't have to manually modify strings. Also, moving the query build component to the bottom could be useful so we can see the full query without having to scout around the file.  `UPDATE table SET {list.join(',')} WHERE conditionals` as a basic approach

cockatrice/src/userinfobox.cpp
ebbit1q: there is no method in place to stop users from using the old login, and I see no reason to stop them, clients can decide to use the less secure method if they wanted to

cockatrice/src/userinfobox.cpp
```diff
@@ -199,7 +201,22 @@ void UserInfoBox::actEditInternal(const Response &r)
 
     Command_AccountEdit cmd;
     cmd.set_real_name(dlg.getRealName().toStdString());
-    cmd.set_email(dlg.getEmail().toStdString());
+    if (client->getServerSupportsPasswordHash()) {
+        if (email != dlg.getEmail()) {
+            // real password is required to change email
+            bool ok = false;
+            QString password =
+                QInputDialog::getText(this, tr("Enter Password"),
```
ZeldaZach, line 209: This doesn't have a max size

cockatrice/src/userinfobox.cpp
```diff
@@ -215,9 +232,42 @@ void UserInfoBox::actPassword()
...
 
+    auto oldPassword = dlg.getOldPassword();
+    auto newPassword = dlg.getNewPassword();
+
+    if (client->getServerSupportsPasswordHash()) {
+        Command_RequestPasswordSalt cmd;
+        cmd.set_user_name(client->getUserName().toStdString());
+
+        PendingCommand *pend = client->prepareSessionCommand(cmd);
+        connect(pend,
```
ZeldaZach, line 243: What is this doing?

servatrice/src/serversocketinterface.cpp
```diff
@@ -1257,24 +1263,70 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountEdit(const Comma
...
 
-    QString realName = QString::fromStdString(cmd.real_name());
-    QString emailAddress = QString::fromStdString(cmd.email());
-    QString country = QString::fromStdString(cmd.country());
-
+    bool checkedPassword = false;
     QString userName = QString::fromStdString(userInfo->name());
+    if (cmd.has_password_check()) {
+        QString password = QString::fromStdString(cmd.password_check());
+        QString clientId = QString::fromStdString(userInfo->clientid());
```
ZeldaZach, line 1270: no size checks

cockatrice/src/userinfobox.cpp
ebbit1q: gees, give me time to do a rebase ok

cockatrice/src/userinfobox.cpp
ebbit1q: it asks the salt before using changePassword

servatrice/src/serversocketinterface.cpp
ebbit1q: getting the clientId doesn't need a check, this info comes from the server db

cockatrice/src/userinfobox.cpp
```diff
@@ -216,9 +235,42 @@ void UserInfoBox::actPassword()
...
+                // we need qoverload here in order to select the right version of this function
+                QOverload<const Response &, const CommandContainer &, const QVariant &>::of(&PendingCommand::finished),
+                this, [=](const Response &response, const CommandContainer &, const QVariant &) {
+                    if (response.response_code() == Response::RespOk) {
+                        changePassword(oldPassword, newPassword);
+                    } else {
+                        QMessageBox::critical(this, tr("Error"),
+                                              tr("An error occured while trying to update your user information."));
```
ebbit1q, line 254:
```suggestion
                                              tr("An error occurred while trying to update your user information."));
```

cockatrice/src/userinfobox.cpp
```diff
@@ -280,7 +335,7 @@ void UserInfoBox::processPasswordResponse(const Response &r)
         case Response::RespInternalError:
         default:
             QMessageBox::critical(this, tr("Error"),
-                                  tr("An error occured while trying to update your user informations."));
+                                  tr("An error occured while trying to update your user information."));
```
ebbit1q, line 338:
```suggestion
                                  tr("An error occurred while trying to update your user information."));
```

servatrice/src/serversocketinterface.cpp
```diff
@@ -1301,14 +1357,19 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountPassword(const C
...
 
     QString oldPassword = nameFromStdString(cmd.old_password());
-    QString newPassword = nameFromStdString(cmd.new_password());
-
-    if (!isPasswordLongEnough(newPassword.length()))
-        return Response::RespPasswordTooShort;
+    QString newPassword;
+    bool newPasswordNeedsHash = false;
+    if (cmd.has_new_password()) {
+        newPassword = nameFromStdString(cmd.new_password());
```
ZeldaZach, line 1363: Do we want to reject too long passwords? Instead of silently trimming?

servatrice/src/serversocketinterface.cpp
```diff
@@ -1301,14 +1357,19 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountPassword(const C
...
+        newPasswordNeedsHash = true;
+        if (!isPasswordLongEnough(newPassword.length()))
+            return Response::RespPasswordTooShort;
+    } else {
+        newPassword = nameFromStdString(cmd.hashed_new_password());
```
ZeldaZach, line 1368: again, should we look to reject if trimmed

servatrice/src/serversocketinterface.cpp
```diff
@@ -1421,7 +1482,16 @@ Response::ResponseCode AbstractServerSocketInterface::cmdForgotPasswordReset(con
         return Response::RespFunctionNotAllowed;
     }
 
-    if (sqlInterface->changeUserPassword(userName, "", nameFromStdString(cmd.new_password()), true)) {
+    QString password;
+    bool passwordNeedsHash = false;
+    if (cmd.has_new_password()) {
+        password = nameFromStdString(cmd.new_password());
```
ZeldaZach, line 1488: reject if trim

servatrice/src/serversocketinterface.cpp
```diff
@@ -1421,7 +1482,16 @@ Response::ResponseCode AbstractServerSocketInterface::cmdForgotPasswordReset(con
...
+        passwordNeedsHash = true;
+    } else {
+        password = nameFromStdString(cmd.hashed_new_password());
```
ZeldaZach, line 1491: reject if trim?

PR #4496, opened 2021-12-10
deprecate the gender property from the protocol entirely

ebbit1q:
this removes all mentions of it and leaves some comments to reserve those places


servatrice/servatrice.sql
```diff
@@ -28,6 +28,7 @@ CREATE TABLE IF NOT EXISTS `cockatrice_users` (
   `admin` tinyint(1) NOT NULL,
   `name` varchar(35) NOT NULL,
   `realname` varchar(255) NOT NULL,
+  -- this field is deprecated and will not be used, this could be removed in a database migration
   `gender` char(1) NOT NULL,
```
ZeldaZach, line 32: we'll want to remove it here, bump db version, and create migrate script in this pr

common/pb/session_commands.proto
```diff
@@ -148,7 +147,7 @@ message Command_AccountEdit {
     }
     optional string real_name = 1;
     optional string email = 2;
-    optional ServerInfo_User.Gender gender = 3;
+    // gender = 3; // deprecated
```
ZeldaZach, line 150: removed, not deprecated

servatrice/servatrice.sql
ebbit1q: I'm not sure how to do a db migration, it's some sql script?

ZeldaZach: Correct, we have a folder of them within the repo already

ZeldaZach (approved):
I confirmed no more instances of gender are in the code base

PR #4497, opened 2021-12-12
Webclient: Render account age in days and years

chrisfinigan:
- Render account age as days and years
- Fix double quotes for linter
- Add utils/convertTime and tests


seavor:
This actually looks really good, very lightweight and simple, plus i definitely appreciate the included unit tests.

However, timestamp formatting is one of those foundational pieces that I think i would prefer left to a library, both for robustness in functionality as well as keeping us from reinventing wheels. Your code here could easily be extended to account for minutes for game age, but it will fall short when we look to represent message timestamps w/ timezone considered.

MomentJS was recently deprecated, and based on this article, Luxon seems to be a winning alternative. I took at a look at their docs, its looks very promising and lightweight.

https://moment.github.io/luxon/#/
https://terodox.tech/migrating-away-from-momentjs-part1/

chrisfinigan:
makes sense, ill work on adding Luxon instead in a timeService

chrisfinigan:
Added prettier config for auto formating, husky hook to run "npm run golden" pre-push and use new timeService

webclient/package.json
```diff
@@ -40,7 +41,13 @@
     "eject": "react-scripts eject",
     "lint": "eslint \"./**/*.{ts,tsx}\"",
     "lint:fix": "eslint \"./**/*.{ts,tsx}\" --fix",
-    "golden": "npm run lint && npm run test"
+    "golden": "npm run lint && npm run test",
+    "prepare": "cd .. && husky install webclient/.husky"
```
seavor, line 45: Does this not work? Also, will this command work on windows?
```suggestion
    "prepare": "husky install ./.husky"
```

seavor: It might make sense to add this command to the env-specific pre-install copy files.

webclient/src/services/TimeService/TimeService.ts
```diff
@@ -0,0 +1,33 @@
+import { DateTime } from 'luxon';
+
+class TimeService {
+  renderAccountAge(seconds: number) {
```
seavor, line 4: Does this support `< 1 Day`? I dont think we care about minutes/hours/seconds for account age, tho i think game age starts with `< 1 minute`.

chrisfinigan: Currently, for less than one days, no text is returned. I am adding a test case to show this

webclient/package.json
```diff
@@ -30,8 +31,8 @@
     "typescript": "^4.4.4"
   },
   "scripts": {
-    "postinstall": "run-script-os",
-    "postinstall:windows": "powershell .\\copy_shared_files.ps1",
+    "postinstall": "run-script-os && npm run prepare",
+    "postinstall:windows": "powershell ./copy_shared_files.ps1",
```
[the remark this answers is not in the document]
chrisfinigan: I was having issues running postinstall with the forward slash. On windows powershell, it would fail with the double forward slash and pass with the change

webclient/package.json
chrisfinigan: I wanted to have husky in the webclient directory but am happy to move it to parent directory with the pre-push hook

chrisfinigan: husky install has to be ran in the same directory as the .git/ hence the cd. I have moved the .husky up to the parent directory. This is working on windows with powershell

seavor (approved):

webclient/package.json
```diff
@@ -30,8 +31,8 @@
     "typescript": "^4.4.4"
   },
   "scripts": {
-    "postinstall": "run-script-os",
-    "postinstall:windows": "powershell .\\copy_shared_files.ps1",
+    "postinstall": "run-script-os && npm run prepare",
```
seavor, line 34: I think the code to run whats in `prepare` should probably just go in the `copy_shared_files` (maybe change the name to something more relevant to the extended usage). We use `run-script-os` to run the prepare scripts that are tailored for each specific environment.

ebbit1q:
this seems to work as expected and shows the age the same as it does in the latest client version.

leap years correctly display as they should, similar to the client now.

tests should also test for leap years and preferably not for the generated string but for the numbers created ( otherwise it'll become hard to translate later ).
we have recently added some similar tests to the cockatrice client itself https://github.com/Cockatrice/Cockatrice/blob/master/tests/test_age_formatting.cpp

if we can also have accounts created on the same day display as 0 days we can merge this pr.

PR #4505, opened 2021-12-23
remove ccache from mac builds

ebbit1q:
ccache never was used and it's stopping the debug build now.


ebbit1q:
seems like I was a bit hastey, the homebrew protbuf formula has been updated only yesterday and it doesn't seem to be working yet...

ebbit1q:
so removing ccache actually had 0 effect on the problem but hey, the builds run now and only take 40 seconds extra for updating homebrew

PR #4507, opened 2021-12-28
Change GHA to store Zip of Build dir

ZeldaZach:


ebbit1q:
why do we need to upload this much?

PR #4512, opened 2022-01-06
missed this in #4496

ebbit1q:
this connect call doesn't prevent compilation so I missed it when removing all instances of the arg


ebbit1q:
accidentally linked wrong pr

Issue #4494, opened 2021-12-03
Pasting card name from Excel into search bar includes linefeed character


MotherofRuins:
When I copy a card name from an Excel cell and paste into the deck editor search bar, for some reason there is what appears to be a space character at the end.  It's actually a linefeed (ASCII 10).  That prevents the card from turning up in search results until I manually delete it.  Pasting into emacs adds the linefeed as well.  Pasting into TCGPlayer search works fine, they must strip the non-printing characters.  Request that CR, LF, and possibly other ascii control characters be stripped when pasted into deck editor search bar.

Reproduce:  Highlight and copy the following two lines.  Paste into search bar.  Text appears correct but search fails.

Abominable
Treefolk

ZeldaZach:
Bounty established

PR #4516, opened 2022-01-08
simplify search string

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4494

you can paste weird whitespace in the search and it'd break the search

- searches are "simplified", this function replaces all stretches of whitespace with single spaces


MotherofRuins:
Thanks!

ebbit1q:
you're welcome

PR #4517, opened 2022-01-08
use multiline input dialog for annotations

### LINKED ISSUE #4430: Line Breaks in Annotations
Would love to be able to type Shift-Enter or some such to put line breaks in annotations.

ebbit1q:
## Related Ticket(s)
- Fixes #4430

it's hard to add new lines to annotations

- changes the dialog from a text one to multiline text
- pressing enter now makes a new line instead

## Screenshots

image

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -3034,7 +3034,7 @@ void Player::actSetPT()
     }
     bool ok;
     dialogSemaphore = true;
-    QString pt = QInputDialog::getText(nullptr, tr("Change power/toughness"), tr("Change stats to:"), QLineEdit::Normal,
+    QString pt = QInputDialog::getText(game, tr("Change power/toughness"), tr("Change stats to:"), QLineEdit::Normal,
                                        oldPT, &ok);
     dialogSemaphore = false;
     if (clearCardsToDelete() || !ok) {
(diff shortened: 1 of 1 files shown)


MotherofRuins:
This is nice, thank you, but my preference would be to bind Enter to the OK button and Shift-Enter to add newline in text.  The dialog box can helpfully remind people "Please enter the new annotation.  Use Shift-Enter for newlines."

ebbit1q:
that will take more time to make, the built in dialog for multiline text is adequate, isn't it?

MotherofRuins:
It's adequate, thanks, but for the most common use case, one line, it would be better if the original interface remained unchanged.

ebbit1q:
you mean you want it reverted?

MotherofRuins:
By original interface, I meant:
Enter = OK should be the same whether in the single-line or multi-line dialog.

ebbit1q:
this is simply the behavior of the multiline input dialog, if we want to customize that we'd have to implement our own dialog from scratch.

MotherofRuins:
I imagine people who usually just write a word will complain when this goes live.  That's the case with 95% of my annotations.  Maybe multiline annotation could be a separate item in the context menus.

ebbit1q:
I think adding more items to the context menu is the last thing we want, it's already cluttered

tooomm:
Did you consider an extended version of the old dialog?
Instead of using one text input or one multi line input... it could be several text inputs that can be cycled trough with tab if needed (or mouse).
Enter should still submit the data in that case.

Mockup:
Untitled2

A more sophisticated version could add more fields on demand.

ebbit1q:
this would be a worse version while still making the dialog from scratch...

tooomm:
Probably better to revert this then until a better solution for the initial issue can be proposed?

tooomm:
As expected, people do not like the new behavior of the multi line dialog that much as it makes the default use case worse: [#4915](https://github.com/Cockatrice/Cockatrice/issues/4915)

Just that the issue/request itself is proposing a workaround, instead a solution to the root cause. (See also the linked PR #4929)

Issue #3941, opened 2020-04-04
Unable to Seek During Replays


D0nV170:
<b>System Information:</b>
  
Client Version: 2.7.4 (2020-03-20)
Client Operating System: Windows 10 (10.0)
Build Architecture: 64-bit
Qt Version: 5.12.6
  
_______________________________________________________________________________________

  
I am unable to seek past about 25% during any replay that is long. If I try to seek any further, it moves it directly to the end or starts it back from the beginning. I don't know what the exact number of minutes is since I can't tell how long a replay is but anything that is more than an average 1v1 60 card deck game length seems to give the issue. Others have reported this issue as well and it's never affected me until yesterday when I had to watch a 4 player EDH game all the way.  Other users have reported this same issue. It feels like the Replay Timeline is off in terms of scaling. Anything short, I can click right on the timeline and it seeks to that point in the game no problem. As for the longer ones, I can only seem to get the Timeline to seek into various parts of the first ~25% or to the very end.

n21lv:
I can confirm this, happens with my replays all the time, regardless of how long I let them to run.

tyler274:
Quite literally all my replays are unseekable as described.

ZeldaZach:
This used to work, and I'm not sure what happened. This is a high priority to fix. Maybe @ctrlaltca or @ebbit1q ya'll have some ideas?

ZeldaZach:
We can't reproduce this issue on Mac or Linux, so I'm assuming it's a Windods limitation.

tooomm:
It works for me with the 2.7.5 release on Windows 10 64bit.

n21lv:
I just updated to 2.7.5 release on Windoes 10 64 bit and tried to play and old replay file, and it still has this issue. @tooomm can you can upload the one you were testing somewhere so I could check if it works for me? I'm not sure when I will get to play next time.

tooomm:
The two replay i tried were 3y old... probably somebody else should check with more recent files. :D

pulverize:
![d0d2af41-fa12-4b58-a568-bed059edde22](https://user-images.githubusercontent.com/3393014/97756427-35eacf00-1ad1-11eb-9c3c-8eb1e0acb7f6.gif)

Windows 10

ebbit1q:
can someone make one of these replays that are broken for them available here? I seem to be unable to reproduce this whatever I do.

pulverize:
[...]

PR #4519, opened 2022-01-10
remove all instances of the type long

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/3941

the long type has different sizes across operating systems and should
not be used

in the timeline an overflow could occur if the width in pixels
multiplied by the total amount of milliseconds in the replay is larger
than 2^31 which is easy enough considering with only 500 pixels width
you'll reach this number with only 1.2 hours of replay (about 4 million
millis), note that this would be windows exclusive as *nix uses 64 bits

qt-json's own repo suggests using qt5's implementation instead, testing
revealed this is quite a bit faster, contrary to #3480

servatrice uses the qthread usleep function which used to be protected
but is now public

cockatrice is not compatible with qt4 and hasn't been for a while

### DIFF
--- a/cockatrice/src/replay_timeline_widget.cpp
+++ b/cockatrice/src/replay_timeline_widget.cpp
@@ -62,7 +62,7 @@ void ReplayTimelineWidget::paintEvent(QPaintEvent * /* event */)
 
 void ReplayTimelineWidget::mousePressEvent(QMouseEvent *event)
 {
-    int newTime = static_cast<int>((long)maxTime * (long)event->x() / width());
+    int newTime = static_cast<int>((qint64)maxTime * (qint64)event->x() / width());
     newTime -= newTime % 200; // Time should always be a multiple of 200
     if (newTime < currentTime) {
         currentTime = 0;
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -211,24 +211,12 @@ Servatrice::~Servatrice()
     clientsLock.unlock();
 
     // client destruction is asynchronous, wait for all clients to be gone
-    bool done = false;
-
-    class SleeperThread : public QThread
-    {
-    public:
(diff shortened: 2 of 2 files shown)


ebbit1q:
after further investigation in became clear that the changes to oracle would not be compatible with older versions of qt, I have split them into a separate branch for now: https://github.com/Cockatrice/Cockatrice/compare/master...ebbit1q:removeqtjson

ebbit1q:
I also made a conditional version https://github.com/Cockatrice/Cockatrice/compare/master...ebbit1q:conditionallyremoveqtjson

ebbit1q:
@n21lv confirmed that this fixed the issue

ZeldaZach:
I want to talk more about this, will DM you

tooomm:
>after further investigation in became clear that the changes to oracle would not be compatible with older versions of qt

Shouldn't the QJson Classes be available starting with Qt 5.0?

ebbit1q:
they can't handle our file sizes, only 5.15 can

tooomm:
I see. Our Debian and Ubuntu are the only ones not using 5.15 yet.
Debian 11 does support it, and Ubuntu 21.04 LTS and newer as well.

ebbit1q:
ubuntu 21.04 is not an lts release, current lts is 20.04

ebbit1q:
we should start packaging for debian 11

tooomm:
> ubuntu 21.04 is not an lts release, current lts is 20.04

You are right. The releases in April (even years) are the LTS ones, not the uneven. 🙈 
Next one will be 22.04 in 3 month.

ebbit1q:
this is a very simple change, can we merge this so we can make the next beta

ebbit1q:
> I want to talk more about this, will DM you

what do you still need?

PR #4522, opened 2022-01-13
implement max lengths for input dialogs that are sent to the server

ebbit1q:
people can enter really long messages where they shouldn't ever need that much

- add a shared header that contains some constants
- add these constants as a max value to every dialog with info that is sent to the server
- some refactors (this always happens doesn't it)

### DIFF
--- a/cockatrice/src/dlg_edit_avatar.cpp
+++ b/cockatrice/src/dlg_edit_avatar.cpp
@@ -1,5 +1,7 @@
 #include "dlg_edit_avatar.h"
 
+#include "stringsizes.h"
+
 #include <QBuffer>
 #include <QDebug>
 #include <QDialogButtonBox>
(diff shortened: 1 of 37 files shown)


cockatrice/src/player.cpp
```diff
@@ -3141,7 +3147,8 @@ void Player::actSetAnnotation()
     bool ok;
     dialogSemaphore = true;
     QString annotation = QInputDialog::getMultiLineText(game, tr("Set annotation"),
-                                                        tr("Please enter the new annotation:"), oldAnnotation, &ok);
+                                                        tr("Please enter the new annotation:"), oldAnnotation, &ok)
+                             .left(MAX_NAME_LENGTH);
```
ZeldaZach, line 3151: Is this what clangify does? Weird

ebbit1q: yes, there is no proper way to validate this type of widget so I just slapped left on it

ZeldaZach (approved):
Seems fine

cockatrice/src/dlg_edit_avatar.cpp
```diff
@@ -75,6 +77,13 @@ QByteArray DlgEditAvatar::getImage()
     QByteArray ba;
     QBuffer buffer(&ba);
     buffer.open(QIODevice::WriteOnly);
-    image.save(&buffer, "JPG");
-    return ba;
+    for (;;) {
```
ZeldaZach, line 80: any reason you do `for (;;)` instead of `while (true)`?

ebbit1q: I just prefer it, I think it looks nice

PR #4523, opened 2022-01-15
Address macOS issue where right-clicking a username in the main chat

ZeldaZach:
…(or game chat) areas would pop up a seemingly empty user profile. This is because the resize event is overridden and doesn't actually attempt to resize based on the size hint of the dialog. Now that we're explicit with the call, this resize should be forced and have comparable results to popping up user profile from the user list.

## Related Ticket(s)
- Fixes #IssueNumber

Before:
Screen Shot 2022-01-15 at 1 27 54 AM

After:
Screen Shot 2022-01-15 at 1 25 36 AM

### DIFF
--- a/cockatrice/src/userinfobox.cpp
+++ b/cockatrice/src/userinfobox.cpp
@@ -14,10 +14,6 @@
 #include <QLabel>
 #include <QMessageBox>
 
-const quint64 SIXTY = 60;
-const quint64 HOURS_IN_A_DAY = 24;
-const quint64 DAYS_IN_A_YEAR = 365;
-
 UserInfoBox::UserInfoBox(AbstractClient *_client, bool _editable, QWidget *parent, Qt::WindowFlags flags)
     : QWidget(parent, flags), client(_client), editable(_editable)
 {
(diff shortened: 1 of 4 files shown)


cockatrice/src/userinfobox.cpp
```diff
@@ -24,7 +24,7 @@ UserInfoBox::UserInfoBox(AbstractClient *_client, bool _editable, QWidget *paren
 {
     QFont nameFont = nameLabel.font();
     nameFont.setBold(true);
-    nameFont.setPointSize(nameFont.pointSize() * 1.5);
+    nameFont.setPointSize(std::ceil(nameFont.pointSize() * 1.5));
```
ebbit1q, line 27: qCeil?

cockatrice/src/userinfobox.cpp
```diff
@@ -131,7 +130,7 @@ void UserInfoBox::updateInfo(const ServerInfo_User &user)
         if (user.accountage_secs() == 0)
             accountAgeString = tr("Unknown");
         else {
-            qint64 seconds = user.accountage_secs();
+            auto seconds = static_cast<qint64>(user.accountage_secs());
```
ebbit1q, line 133: is this necessary?

cockatrice/src/userinfobox.cpp
```diff
@@ -305,5 +304,6 @@ void UserInfoBox::resizeEvent(QResizeEvent *event)
 {
     QPixmap resizedPixmap = avatarPixmap.scaled(avatarLabel.size(), Qt::KeepAspectRatio, Qt::SmoothTransformation);
     avatarLabel.setPixmap(resizedPixmap);
+    resize(sizeHint());
```
ebbit1q, line 307: resizing the window will result in an infinite recursion now....

cockatrice/src/userinfobox.cpp
ebbit1q: actually, why not use setPointSizeF?

ZeldaZach: Didn't know about it -- changed

cockatrice/src/userinfobox.cpp
ZeldaZach: It was a warning that was simple enough to handle as it's an unsigned to signed conversion

ebbit1q: why not use unsigned then

cockatrice/src/userinfobox.cpp
```diff
@@ -24,7 +23,7 @@ UserInfoBox::UserInfoBox(AbstractClient *_client, bool _editable, QWidget *paren
 {
     QFont nameFont = nameLabel.font();
     nameFont.setBold(true);
-    nameFont.setPointSize(nameFont.pointSize() * 1.5);
+    nameFont.setPointSizeF(nameFont.pointSize() * 1.5);
```
ebbit1q, line 26:
```suggestion
    nameFont.setPointSizeF(nameFont.pointSizeF() * 1.5);
```

ebbit1q:
wait, why do we do something like assigning SIXTY as a constant, like what

cockatrice/src/userinfobox.cpp
```diff
@@ -11,20 +11,19 @@
...
-#include <QHBoxLayout>
 #include <QLabel>
 #include <QMessageBox>
 
-const qint64 SIXTY = 60;
-const qint64 HOURS_IN_A_DAY = 24;
-const qint64 DAYS_IN_A_YEAR = 365;
+const quint64 SIXTY = 60;
+const quint64 HOURS_IN_A_DAY = 24;
+const quint64 DAYS_IN_A_YEAR = 365;
```
tooomm, line 19: >wait, why do we do something like assigning SIXTY as a constant, like what

I assume the intention was to name it `MINUTES_IN_AN_HOUR` when looking at the other constant names.

ebbit1q:
I made a pr that will address this #4526

PR #4526, opened 2022-01-15
use datetime for calculating account age

ebbit1q:
make translating easier by using tr multiples
automatically account for leap days
also adds creation date to the ui


ebbit1q:
image
problem

PR #4531, opened 2022-01-18
UI: Settings dialog layout fixes

tooomm:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -421,7 +421,7 @@ void AppearanceSettingsPage::retranslateUi()
     cardScalingCheckBox.setText(tr("Scale cards on mouse over"));
 
     handGroupBox->setTitle(tr("Hand layout"));
-    horizontalHandCheckBox.setText(tr("Display hand horizontally (wastes space)"));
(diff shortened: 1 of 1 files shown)


cockatrice/src/dlg_settings.cpp
```diff
@@ -805,6 +805,7 @@ MessagesSettingsPage::MessagesSettingsPage()
     connect(&invertHighlightForeground, SIGNAL(stateChanged(int)), this, SLOT(updateTextHighlightColor(int)));
 
     mentionColor = new QLineEdit();
+    mentionColor->setPlaceholderText(tr("Enter HEX code"));
```
ebbit1q, line 808:
```suggestion
    mentionColor->setPlaceholderText(tr("Enter hex code"));
```
hex is short for hexadecimal

tooomm: I've seen both.
When you search "color picker" in Google to access their tool, it uses HEX for example.

tooomm: Obsolete with #4535

PR #4524, opened 2022-01-15
UI: Improve alignment in user info

tooomm:
## Short roundup of the initial problem
User information is not displayed very pretty.

## What will change with this Pull Request?
- Align account age to the front, same as name
- Uniform capitalization
- Center icons
- Have little spacing between flag and country code

## Screenshots

- before:
before

- after:
after

PR #4533, opened 2022-01-19
UI: User info dialog updates

tooomm:
## Related Ticket(s)
- Follow up to #4524

Alignment was mixed across various lines.

- Reorder information to avoid alignment issues
- Move user "level" icon in front of the user name
- Flag icon a tiny bit smaller
- Some reordering and updated names

## Screenshots

- before
userinfo-old

- after
userinfo

<br>
Note: I think the new account creation date from #4526 looks a bit odd as is right now. 🤔


ebbit1q:
I do think the pawn is sorta part of the user level though, also I removed the hardcoded multiple from the user age part refer to https://github.com/Cockatrice/Cockatrice/pull/4120

ZeldaZach:
No objections to moving the pawn up by the username, looks neat. Would still like location to be 2nd tho

tooomm:
>I do think the pawn is sorta part of the user level though

I guess one could argue the pawn as visual representation of a player with their role is closely connected to the player and his user name as well. It feels natural to me to have it like that - the creator is listed exactly like that in the list of games. Basically all occurrences of the player name have the pawn close by.
The text explanation should be there nonetheless of course.

Both ways of placement are fine I guess, but this way it help with the alignment issue.

>No objections to moving the pawn up by the username, looks neat. **Would still like location to be 2nd tho**

I thought that might bring us in the same situation where there is an icon in one line with text below it and it would result in the same alignment issues this tries to circumvent.

Turns out, it looks ok as it's only one icon:
150208859-201725ad-2e3c-44d6-a5e9-3097f4afb64a vs. userinfo

I prefer the first one, too now. :D

<br>

Another thing:
The resizing behavior of the avatar is unchanged and feels a bit weird, but I found no quick solution for it.

ZeldaZach:
Remind me to change the User level to be commas instead of bars, and to make DONATOR/VIP as capital case first then lowercase

tooomm:
>Remind me to change the User level to be commas instead of bars, and to make DONATOR/VIP as capital case first then lowercase

The acronym VIP should stay all capital letters though.

tooomm:
@ZeldaZach Reminder

tooomm:
>>Remind me to change the User level to be commas instead of bars, and to make DONATOR/VIP as capital case first then lowercase
>
>The acronym VIP should stay all capital letters though.

@ZeldaZach I think it's ok to put a second reminder after one year.

Anyways, this can be merged I guess?

PR #4535, opened 2022-01-20
try a color picker instead

ebbit1q:
replaces hex code with native color picker

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -14,6 +14,7 @@
 #include <QApplication>
 #include <QCheckBox>
 #include <QCloseEvent>
+#include <QColorDialog>
 #include <QComboBox>
 #include <QDebug>
 #include <QDesktopServices>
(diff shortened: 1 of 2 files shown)


cockatrice/src/dlg_settings.cpp
```diff
@@ -868,15 +867,12 @@ MessagesSettingsPage::MessagesSettingsPage()
...
-    aAdd->setToolTip(tr("Add New Message"));
     connect(aAdd, SIGNAL(triggered()), this, SLOT(actAdd()));
     aEdit = new QAction(this);
     aEdit->setIcon(QPixmap("theme:icons/pencil"));
-    aEdit->setToolTip(tr("Edit Message"));
     connect(aEdit, SIGNAL(triggered()), this, SLOT(actEdit()));
     aRemove = new QAction(this);
     aRemove->setIcon(QPixmap("theme:icons/decrement"));
-    aRemove->setToolTip(tr("Remove Message"));
     connect(aRemove, SIGNAL(triggered()), this, SLOT(actRemove()));
```
tooomm, line 876: Hmm. Can you briefly explain why you move those?
The other tooltips for the add/remove/edit URL buttons of the card sources are placed similarly in the same file.

tooomm:
I like the idea of a color picker!

cockatrice/src/dlg_settings.cpp
ebbit1q: I move these to the retranslate function because otherwise they will not be changed when you change the language setting

ebbit1q:
the reset button doesn't work properly and the alert one I left as is, do you have any feedback on the interface changes?

cockatrice/src/dlg_settings.cpp
tooomm: Oh, great catch.

I recently added missing tooltips to the custom picture URL buttons in settings as well. The same issue does probably apply there as well. Can you have a look, please?
https://github.com/ebbit1q/Cockatrice/blob/4531_color_picker/cockatrice/src/dlg_settings.cpp#L568-L576

ebbit1q: yes, these are in the wrong spot, in fact anything that has the tr() function in it should be in the retranslate function

tooomm:
> the reset button doesn't work properly and the alert one I left as is, do you have any feedback on the interface changes?

The layout is in line with the old one and improves its usability. 👍 

For me the reset button switches the color to black. Isn't that the value you defined for chat mentions currently?
A good default value to reset to could be bright red `#ff0000`.

The text preview works well here. I guess foreground/background highlighting for mentions of your own user name is not needed and makes it simpler.

The custom alert words highlighting benefits from a color picker, too.
Foreground/background highlighting could be helpful there to better distinguish such highlights from user names ones?

Ideas:
- Shorter button label: `Select Color` or `Color Picker`?
- Tooltip for the reset button, e.g. `Restore Default Color`
- Reword the color preview to e.g. `Preview`, `@YourName`, `UserName` or `Player1`

cockatrice/src/dlg_settings.cpp
tooomm: Understandable, it takes focus away.
Probably good practice to just open issues to keep track for later and raise awareness. 👍

ZeldaZach:
Color picker works as expected, nice work.

However, there are a few issues I've observed.
- No ability to invert text color (or change the text color).
- Refresh button doesn't seem to work (yet). 
- Also, just changing the word "Color" to the color they pick isn't great UX or a11y for our users. Having a "live" demo (i.e the old system) gave a good feel on what the user would actually see w/o having to get a real alert. This approach doesn't seem to give that feeling.

PR #4543, opened 2022-01-23
create workflow for automatically updating the translation sources

ebbit1q:
runs lupdate daily and automatically pushes changes to master

### DIFF
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -3,8 +3,8 @@ name: Update translation source
 on:
   workflow_dispatch:
   schedule:
-    # runs every day
-    - cron: '0 0 * * *'
+    # runs once per month
+    - cron: '0 0 1 * *'
 
 jobs:
   translations:
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,66 @@
+name: Update translation source
+
+on:
+  workflow_dispatch:
+  schedule:
+    # runs every day
+    - cron: '0 0 * * *'
+
+jobs:
+  translations:
+    # Do not run the scheduled workflow on forks
+    if: github.event != 'schedule' || github.repository_owner == 'Cockatrice'
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install lupdate
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends qttools5-dev-tools
+
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Update cockatrice translations
+        shell: bash
+        run: |
+          shopt -s globstar # globstar is needed for recursive **
+          lupdate -version
+          echo "reading the following source files:"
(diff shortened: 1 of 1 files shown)


ebbit1q:
see this completed run https://github.com/ebbit1q/Cockatrice/runs/4910366045?check_suite_focus=true

ebbit1q:
and this run for no updates https://github.com/ebbit1q/Cockatrice/runs/4910394630?check_suite_focus=true

tooomm:
That's great, I love it.

- Workflow dispatch to easily trigger manual runs on demand 👍 
- Clickable link to commit as reference 👍 

<br>

I would tweak the scheduled runs a bit.
Daily doesn't accomplish too much IMO. We do not want to annoy translators with constant mails that a few strings are updated on the platform. Transifex automatically pulls new strings from the repo and sends notifications.
Batching them together and check weekly is probably better. Maybe even monthly or every 6 weeks is sufficient?
Additionally (or only?), we can run it on beta releases to make sure translations are up to date and translators get pinged before a potentially upcoming main release.

ebbit1q:
we often do commits to master in batches anyway, I'd rather have it push those changes to transifex right away

.github/workflows/translations.yml
```diff
@@ -0,0 +1,66 @@
+name: Update translation source
+
+on:
+  workflow_dispatch:
+  schedule:
+    # runs every day
+    - cron: '0 0 * * *'
```
tooomm, line 7: >we often do commits to master in batches anyway, I'd rather have it push those changes to transifex right away

Let's go with weekly then and try it first. We can adjust from there. I'm not really comfortable with any higher frequency.
I really feel bad for translators that get mails 3 times a week just because some PR's got merged and there are 2 strings updated each.
It's a sensitive setting that doesn't really change anything on the project side no matter if its daily or monthly considering our release schedule.

And the manual, one click run to enforce an immediate update if really needed works always.

```suggestion
on:
  workflow_dispatch:
  schedule:
    # runs every friday
    - cron: '0 0 * * 5'
  push:
    tags:
      - '*-beta*'
```

ebbit1q:
I assume you know more about translations than me, I changed it to once per month, I am not comfortable with adding it to beta tags, that might not interact nicely with our builds

tooomm (approved):

ebbit1q:
I forgot about the branch protection, maybe it'll be better to create automatic prs

tooomm:
```
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: At least 1 approving review is required by reviewers with write access.  
```

Seems like there are only clunky workarounds, PR creation sounds not too bad!

ZeldaZach:
No issue with updating once per month, seems nice! However, it doesn't seem like branch protection was addressed here. Is the PR-creation service a feature that is a WIP?

tooomm:
>However, it doesn't seem like branch protection was addressed here. Is the PR-creation service a feature that is a WIP?

That is dealt with in https://github.com/Cockatrice/Cockatrice/pull/4544. There are ready made actions available for PR creation.

PR #4544, opened 2022-01-24
CI: Add automatic PR creation for source string updates

tooomm:
## Related Ticket(s)
- Fixes issue encountered in #4543

Due to the enabled branch protection, GitHub Actions can not directly commit to `master`.
We need to open a PR with the output of the scheduled CI run instead.

- Uses https://github.com/marketplace/actions/create-pull-request action
  >**Action behaviour**
  >
  >The default behaviour of the action is to create a pull request that will be continually updated with new changes until it is merged or closed. Changes are committed and pushed to a fixed-name branch, the name of which can be configured with the `branch` input. Any subsequent changes will be committed to the same branch and reflected in the open pull request.
  >
  >How the action behaves:
  >
  >- If there are changes (i.e. a diff exists with the checked-out base branch), the changes will be pushed to a new branch and a pull request created.
  >- If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.
  >- If a pull request already exists and there are no further changes (i.e. no diff with the current pull request branch) then the action exits silently.
  >- If a pull request exists and new changes on the base branch make the pull request unnecessary (i.e. there is no longer a diff between the pull request branch and the base), the pull request is automatically closed. Additionally, if delete-branch is set to true the branch will be deleted.

- Save relevant parts of the lupdate command as step output to re-use the data in the PR body
    - ebbit outsourced that logic in a nice script and introduced a template file for the PR body
- Fix deprecation of `set-output`, see https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- Bump action versions
- Wording updates
- Do not trigger other CI runs when only the translation workflow is updated
- Reflect changes to the translation process in the CONTRIBUTING file
- Allow workflow to run in PR's that change it to see some results
- This does not take care of the strings in the webclient yet (#4562) (Note: See [i18next docs on string extraction](https://www.i18next.com/how-to/extracting-translations), e.g. [i18next-parser](https://github.com/i18next/i18next-parser))

<br>

@ebbit1q Can you please double check the bash parts?


ebbit1q:
instead of creating long outputs consider simply saving the output as a file with tee

I'd also like to have it not create a new pr if an old one already exists, that way we can run this more frequently but not accumulate more prs

another thing I'd like to know is if this create pr action correctly behaves when there are no changes to commit, it doesn't look like you ever add which files changed to it, I'd rather use more direct creation of the git commit and only optionally, is there a specific reason you removed the git diff step from the original version?

tooomm:
Action behaviour

>instead of creating long outputs consider simply saving the output as a file with tee

Thanks for pointing me! Will have a look 👍 

>I'd also like to have it not create a new pr if an old one already exists, that way we can run this more frequently but not accumulate more prs

It doesn't do that.
I thought that part is also in the "how it behaves" section, my bad. Copied that part from the docs now, too.
*"The default behaviour of the action is to create a pull request that will be continually updated with new changes until it is merged or closed. Changes are committed and pushed to a fixed-name branch, the name of which can be configured with the branch input. Any subsequent changes will be committed to the same branch and reflected in the open pull request."*

With an updated PR in place we can think about another schedule, yes.
I would also bring runs on beta tags on the table again. We don't have to go as crazy as running it on all pushes to master though.

>another thing I'd like to know is if this create pr action **correctly behaves when there are no changes to commit**, it doesn't look like you ever add which files changed to it, I'd rather use more direct creation of the git commit and only optionally, is there a specific reason you removed the git diff step from the original version?

It's the first time I'm working with this action, but I think the first is covered by the action and the second is not needed for it to work:
*"If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently."*
*"If a pull request already exists and there are no further changes (i.e. no diff with the current pull request branch) then the action exits silently."*

tooomm:
I ended up using step outputs again, the output is way shorter now and only one line.
Do you still think that's an issue? I also tried it with GitHub environment variables, but is that really better?

Any other suggestions to improve things? @ebbit1q

ebbit1q:
if you look at the captured output you'll see that lupdate creates 2 lines of output, only one line is captured by the step instead of the full output

tooomm:
I intentionally do not save the complete lupdate output to keep only the relevant data and nicely present it in the PR body.

tooomm:
> using a side effect of the output setting like that is not easy to see, someone editing this later might not be aware of this, removing all but the first line should be explicit

I'm not quite sure what you mean. I do not abuse the behaviour of the output setting? 🤔 
I use sed to manipulate the content of the file saved with tee before creating the output.

Check the [code of this test commit](https://github.com/Cockatrice/Cockatrice/pull/4544/commits/cac4adb67c1a2558106477431be5a82249936760#diff-a75e641518c1159cf35efe74d38212add0ebee51c513e91821d31a590ded7d47R61) and the [related actions run](https://github.com/Cockatrice/Cockatrice/runs/4937192846?check_suite_focus=true#step:5:23) for clarification.

All output from lupdate is completely printed to the console and fully saved to a file with tee:
```
lupdate ... | tee cockatrice_stats.txt
```

Then, I edit the "cockatrice_stats.txt" file (-i), delete the first line (1d) and remove leading whitespace (s/^[ \t]*//) with sed:
```
sed -i '1d; s/^[ \t]*//' cockatrice_stats.txt
```

Should I add a comment to the sed line?

ebbit1q:
I went ahead and made a script for it and ran some tests https://github.com/ebbit1q/Cockatrice/pull/3

I removed the run on changes to prs, it doesn't create a new pr and it's better to just run it on a fork and use the dispatch

.github/workflows/translations.yml
```diff
@@ -5,16 +5,13 @@ on:
...
-  pull_request:
-    paths:
-      - '.github/workflows/translations.yml'
 
 jobs:
   translations:
-    # Do not run the scheduled workflow on forks, and ensure the action runs only from master branch but still from any PR
-    if: (github.event != 'schedule' || github.repository_owner == 'Cockatrice') && (github.ref_name == 'master' || github.event_name == 'pull_request')
+    # Do not run the scheduled workflow on forks
+    if: (github.event != 'schedule' || github.repository_owner == 'Cockatrice')
```
tooomm, line 12: The idea to only run it from master branch was to ensure there are no manual dispatch runs against any other branch by fault that could lead to unexpected results and weird PR's based on false or outdated strings.

ebbit1q: I don't think that's a concern, the dispatch selects the master branch as the default and there is no reason to lock it away

tooomm:
>I went ahead and made a script for it and ran some tests ebbit1q#3

That works nicely. 👍 👍 
I like some changes you made on the way. I missed for example that the GitHub bot is not the default commit author in that action!

>I removed the run on changes by prs, it doesn't create a new pr and it's better to just run it on a fork and use the dispatch

It makes a lot of sense when proposing changes, to have a first test run right in the related PR. We do the same for all other workflows.  Why change the procedure for this particular one? Ofc it should not create a PR in such cases.

tooomm:
Looked at this again, still seems ok. Just updated the used actions to the most recent versions.

As the current action does not work at all, I suggest to merge this and see how it plays out with the PR creation.
Adding webatrice can then be a follow up.

.github/workflows/translations.yml
```diff
@@ -1,66 +1,87 @@
-name: Update translation source
+name: Update Translation Source
 
 on:
   workflow_dispatch:
   schedule:
-    # runs once per month
+    # runs at the start of each month (UTC)
```
ZeldaZach, line 6: Is once a month enough?

.github/workflows/translations.yml
```diff
@@ -1,66 +1,87 @@
...
 
     steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
       - name: Install lupdate
         shell: bash
         run: |
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends qttools5-dev-tools
```
ZeldaZach, line 28: Does it matter if we use qt5 or 6 here?

.github/workflows/translations.yml
```diff
@@ -1,66 +1,87 @@
...
+        with:
+          add-paths: |
+            cockatrice/cockatrice_en@source.ts
+            oracle/oracle_en@source.ts
+          commit-message: Update translation source strings
+          # author is the owner of the commit
+          author: github-actions <github-actions@github.com>
+          branch: ci-update_translations
+          delete-branch: true
+          title: '[Translations] Update source strings'
```
ZeldaZach, line 69: Maybe annotate it's automatically generated?

.github/workflows/translations.yml
```diff
@@ -1,66 +1,87 @@
-name: Update translation source
+name: Update Translation Source
 
 on:
   workflow_dispatch:
   schedule:
-    # runs once per month
+    # runs at the start of each month (UTC)
     - cron: '0 0 1 * *'
+  pull_request:
```
ZeldaZach, line 8: I don't see a need to do this via PRs, as cron is more than enough. I'd increase the frequency to every 2 weeks however

tooomm: It won't create a PR when the run is triggered from a PR (like this one).
The purpose is only to verify and check changes made to the workflow itself which would not be possible otherwise - that's why it's limited to only run on changes to that one config file.
I should add the related bash script there as well actually.

.github/workflows/translations.yml
ebbit1q: the body of the commit contains that and more, the template for that is added here as well

.github/workflows/translations.yml
ebbit1q: yes, we already discussed this

tooomm: Now, with the PR creation instead of a push to master (which is what we planned first, but did not end up working), I'm ok with something like 2 weeks. 👍 

We can always trigger a manual run on demand with one click though. That's what the "dispatch" is enabling.

ebbit1q: see https://github.com/Cockatrice/Cockatrice/pull/4543#issuecomment-1019732098

.github/workflows/translations.yml
ebbit1q: the run triggered by pr would be useless, if you're not testing the creation of the translation pr (preferably on a fork) what's the point?

.ci/update_translations_template.md
```diff
@@ -0,0 +1,11 @@
...
+
+<br>
+
+Updated source strings for translations based on {{ .commit }}:
+- {{ .cockatrice_output }} (Cockatrice)
+- {{ .oracle_output }} (Oracle)
+
+<br>
+
+*This PR is automatically generated and updated by the workflow at `.github/workflows/translations.yml`.*
```
ebbit1q, line 11: is there a specific reason for removing the location of the template here? it'd be useful to know where the template is stored right away.

.ci/update_translations_template.md
```diff
@@ -0,0 +1,11 @@
+After merging, all changes to the source language are available to translators at [Transifex](https://www.transifex.com/projects/p/cockatrice/) shortly.
+
+<br>
+
+Updated source strings for translations based on {{ .commit }}:
```
ebbit1q, line 5: I preferred when the {{.commit}} was its own line, this makes it easier to scan for when checking if the pr has been updated

.github/workflows/translations.yml
tooomm: That's a good question. Probably good to use Qt6 here as well to stay uniform in the repo with the recent changes.

.github/workflows/translations.yml
tooomm: The PR creation is only one step of the workflow.
The logs give useful information in many other areas, e.g. if switching from Qt 5 to 6 for lupdate would break something...
Isn't it just good practice to always check and verify changes that are made by people on the internet?

This 30s action running in a PR happens probably like every 100th PR that gets opened... 🙄

.ci/update_translations_template.md
tooomm: Hmm? It's both times one line of fixed text of the same length with an updating hash at the end?
`Updated source strings for translations based on {{ .commit }}:` vs. one additional line that says the same:
`The most recent update is based on commit: {{ .commit }}:`

I tried to reduce and combine information to make the template more light and focus on the relevant information without sacrificing useful content.

.ci/update_translations_template.md
tooomm: If it would be a an important file that needs frequent updates, I would have kept it that prominently.
As is, it won't see many changes and I think it's no useful information in most cases. The mentioned config file links to it in context as well, so it can still be find relatively well.

.ci/update_translations_template.md
ebbit1q: I separated the information that targets a different group of people, the simple list of how many changes happened at the top which is easy to understand and the more technical information like the hash number and where to change this template or workflow at the bottom

.github/CONTRIBUTING.md
```diff
@@ -321,20 +322,46 @@ https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals)
 
 If you're about to propose a change that adds or modifies any translatable
 string in the code, you don't need to take care of adding the new strings to
-the translation files.
-Every few days, or when a lot of new strings have been added, someone from the
-development team will take care of extracting all the new strings and adding
-them to the english translation files and making them available to translators
-on Transifex.
+the translation files.<br>
+We have an automted process setup to update our language source files accordingly
```
ebbit1q, line 326:
```suggestion
We have an automated process to update our language source files on a schedule
```

.github/CONTRIBUTING.md
```diff
@@ -321,20 +322,46 @@ https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals)
...
 
-When new translatable strings have been added to the code, a maintainer should
-make them available to translators on Transifex. Every few days, or when a lot
-of new strings have been added, a maintainer should take care of extracting all
-the new strings and add them to the english translation files.
+When new translatable strings have been added to the code, a maintainer has to
+make them available to translators on Transifex.
 
-To update the english translation files, re-run cmake enabling the appropriate
+To help with that, we have an automated CI run setup, that regularly looks at the
```
ebbit1q, line 335: github calls them workflows instead of CI run setup and I'd like to keep this a bit more consistent

.github/workflows/translations.yml
ebbit1q: of course it is good to verify changes, and you should do so in a proper fashion, a dry run like that will point out simple typos in the workflow but you should be testing the workflow with an automatically generated pr anyway, so I don't see a benefit to having this trigger like this because well, as you mentioned this will rarely happen.

.github/workflows/translations.yml
```diff
@@ -1,66 +1,87 @@
...
         run: |
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends qttools5-dev-tools
 
-      - name: Checkout repo
-        uses: actions/checkout@v2
-
-      - name: Update cockatrice translations
+      - name: Update Cockatrice translation source
+        id: cockatrice
```
ebbit1q, line 31: why add the id field? it's unused

.github/CONTRIBUTING.md
tooomm: Yep. That makes sense.

`setup` should have been `set up`. Will just remove it.

.ci/update_translations_template.md
tooomm: I see. Do you like the adjusted template?

.github/workflows/translations.yml
tooomm: There is use in having it (also no harm). Even if you see only little benefit.

I would much rather cut down on our huge general build matrix. Bionic accounts for 0,2% of all 2.8.0 downloads... 🙈

.ci/update_translations_template.md
```diff
@@ -1,13 +1,11 @@
...
+- {{ .oracle_output }} (Oracle)
 
 <br>
 
-The most recent update is based on commit: {{ .commit }}
+Last changes are based on commit {{ .commit }}.
 
-*This PR is automatically generated and updated by the workflow at `.github/workflows/translations.yml`. This template is located at `.ci/update_translations_template.md`.*
+---
+*After merging, all changes to the source language are available for translation at [Transifex](https://www.transifex.com/projects/p/cockatrice/) shortly.*<br>
```
ebbit1q, line 10: I say the text this pr is automatically generated should be the first part of this section

.github/workflows/translations.yml
ebbit1q: it doesn't matter how many people download bionic, we need to verify our minimum requirements with that build

tooomm: My point was a different one, even though the example was maybe a bad pick.

Anyway, not sure if the topic is basically worth staling a PR for one year.

tooomm:
Updated this old PR again to hopefully get it merged at one point...
50 strings are missing in our source by now and can not be translated since the job is not done manually (the current automation does not work).

I would like to see it in action and do further adjustments later if there is need. The PR and old discussions here make things a bit unclear by now.

Open, but not blocking question: https://github.com/Cockatrice/Cockatrice/pull/4544#discussion_r867552652

.github/workflows/translations.yml
tooomm: I tried to get `lupdate` to work in a quick test - no luck.
So Qt5 is good enough for now I guess

PR #4551, opened 2022-01-31
Webatrice: implement password length requirements

seavor:


ebbit1q (approved):
I deployed this to beta and it works as expected

PR #4555, opened 2022-02-01
Webatrice: hide account details on failed login

seavor:
As a security measure, failed login attempts shouldnt reveal whether the:

- password was incorrect, or
- the username was incorrect, or
- the username doesn't exist


seavor:
Looks like i have some unit tests to fix with these changes.

seavor:
Previous discussions have decided against this.

PR #4557, opened 2022-02-04
Webatrice: fix login bugs

seavor:
## Related Ticket(s)
- Fixes https://trello.com/c/sU6Omg6C/6
- Fixes https://trello.com/c/sU6Omg6C/7
- Fixes https://trello.com/c/q2jOCqLW/8
- Fixes https://trello.com/c/q2jOCqLW/9

### DIFF
--- a/webclient/src/websocket/events/SessionEvents.ts
+++ b/webclient/src/websocket/events/SessionEvents.ts
@@ -3,7 +3,7 @@ import { Room, StatusEnum, User, WebSocketConnectReason } from 'types';
 import { SessionCommands } from '../commands';
 import { RoomPersistence, SessionPersistence } from '../persistence';
 import { ProtobufEvents } from '../services/ProtobufService';
-import { passwordSaltSupported } from '../utils';
+import { generateSalt, passwordSaltSupported } from '../utils';
 import webClient from '../WebClient';
 
 export const SessionEvents: ProtobufEvents = {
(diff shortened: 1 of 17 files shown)


webclient/src/websocket/events/SessionEvents.ts
```diff
@@ -130,11 +130,8 @@ function serverIdentification(info: ServerIdentificationData) {
       }
       break;
     case WebSocketConnectReason.REGISTER:
-      if (passwordSaltSupported(serverOptions, webClient)) {
-        SessionCommands.requestPasswordSalt();
-      } else {
-        SessionCommands.register();
-      }
+      const passwordSalt = passwordSaltSupported(serverOptions, webClient) ? generateSalt() : undefined;
```
brentmclark, line 133: A small NIT (I use a Panda 🐼)

Would `null` be more appropriate here than `undefined`?  It _is_ defined, it's just a nil value.

webclient/src/websocket/services/WebSocketService.ts
```diff
@@ -60,7 +61,10 @@ export class WebSocketService {
     const socket = new WebSocket(url);
     socket.binaryType = 'arraybuffer';
 
+    const connectionTimer = setTimeout(() => socket.close(), this.keepalive);
```
brentmclark, line 64: Is `socket.error()` a more appropriate choice here?  I can't quite tell, but this implementation suggests to me that if the code is ever called it's unexpected and represents an unexpected error.

brentmclark (commented):
A NIT and an actual question for you.

webclient/src/websocket/events/SessionEvents.ts
seavor: Respectable NIT lol

webclient/src/websocket/services/WebSocketService.ts
seavor: From testing this out, calling `socket.close` before a connection is made results in an error being thrown anyway. This only gets called if the connection isnt made within the time limit. This was to prevent the 30+ second default timeout that webSockets limit themselves to, we basically just cancel the request ourself with `socket.close`.

brentmclark: Fair enough.  Thanks for the enlightenment.

brentmclark (approved):
LGTM :shipit:

PR #4560, opened 2022-02-05
Webatrice: tech debt

seavor:
## Related Ticket(s)
- Fixes https://trello.com/c/K6lSPgkV/6-turn-off-browser-autocomplete-for-text-fields
- Fixes https://trello.com/c/vGozi7yz/9-consider-trim-for-certain-fields
- Fixes https://trello.com/c/UXDVuXu3/10-support-servers-that-dont-require-account-passwords


webclient/src/forms/KnownHostForm/KnownHostForm.tsx
```diff
@@ -33,6 +33,14 @@ const KnownHostForm = ({ host, onRemove, onSubmit }) => {
     }
   };
 
+  const handleOnSubmit = ({ name, host, port, ...values }) => {
+    name = name.trim();
+    host = host.trim();
+    port = port.trim();
```
brentmclark: Not a huge deal obviously, but could the `port` input be set to `type="number"` to remove the need for a trim here?

webclient/src/forms/RegisterForm/RegisterForm.tsx
```diff
@@ -100,10 +105,10 @@ const RegisterForm = ({ onSubmit }: RegisterFormProps) => {
 
         return (
           <>
-            <form className="RegisterForm" onSubmit={handleSubmit} autoComplete="off">
+            <form className="RegisterForm" onSubmit={handleSubmit}>
               <div className="RegisterForm-column">
                 <div className="RegisterForm-item">
-                  <Field label="Player Name" name="userName" component={InputField} />
+                  <Field label="Player Name" name="userName" component={InputField} autoComplete="username" />
```
brentmclark, line 111: `autoComplete="username"`?

webclient/src/forms/RequestPasswordResetForm/RequestPasswordResetForm.tsx
```diff
@@ -26,9 +26,13 @@ const RequestPasswordResetForm = ({ onSubmit, skipTokenRequest }) => {
     setIsMFA(true);
   }, ServerTypes.RESET_PASSWORD_CHALLENGE, []);
 
-  const handleOnSubmit = (form) => {
+  const handleOnSubmit = ({ userName, email, ...values }) => {
     setErrorMessage(false);
-    onSubmit(form);
+
+    userName = userName.trim();
+    email = email.trim();
```
brentmclark: Similarly to `number` above, could this field be set to `type="email` to avoid an explicit trim?

webclient/src/components/InputField/InputField.tsx
```diff
@@ -38,6 +38,7 @@ const InputField = ({ input, meta: { touched, error, warning }, ...args }) => {
       ) }
 
       <TextField
+        autoComplete='off'
```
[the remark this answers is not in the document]
seavor: I think this is sufficient enough for us, i dont think we wanna try too hard to stop chrome from doing what it wants to.

webclient/src/forms/RegisterForm/RegisterForm.tsx
brentmclark: Is this meant to be the `autocomplete="randomString"` hack for Chrome perhaps?

brentmclark (approved):
Some open questions and possibly some smol changes, but nothing blocking.  This all looks great!

webclient/src/forms/RegisterForm/RegisterForm.tsx
seavor: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

webclient/src/forms/KnownHostForm/KnownHostForm.tsx
seavor: It is currently set to `number`, i wasnt sure if that input type would always protect against the need for trim, so i put this here as a failsafe. Do you think it's unnecessary?

webclient/src/forms/RequestPasswordResetForm/RequestPasswordResetForm.tsx
seavor: This one is missing the `type="email"` attribute, i can add it. Similarly, do you think this protects against the need for using trim?

@ZeldaZach any issue with adding the type=email attr? This will enforce the browsers email validation.

ebbit1q: why wouldn't we? I see no problem with that

seavor: We were both wrong about this, `number` does protect us from the need for trimming, but something with React is causing type email to not be validated, nor does it protect us from needing to trim.

webclient/src/forms/RegisterForm/RegisterForm.tsx
brentmclark: Neat!  TIL.

webclient/src/components/InputField/InputField.tsx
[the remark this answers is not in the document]
brentmclark: No worries on my end.  Worst case scenario is that Chrome users continue to report issues and we deal with it then.  We have more pressing issues IMO.

brentmclark:
The updates look great!  :shipit:

webclient/src/components/InputField/InputField.tsx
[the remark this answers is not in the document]
seavor: My personal thoughts on it are, browsers are free to augment the user experience as they see fit, such as with field history or password/address managers. Unless it's really interfering with our application, or there are security concerns with autofill (Bank of America clears my password field after chrome autofills it, but i can still focus on the field and get the autofill options to put it back), then i dont think we should overwork to stop them.

ebbit1q (approved):

PR #4563, opened 2022-02-06
Webatrice: fix saved password

seavor:
## Related Ticket(s)
- Fixes https://trello.com/c/8Mut2EHb/10-saved-password-is-removed-after-failed-login
- Fixes issue where `*saved*` label appeared instead of `Password` in focused mode
- Fixes issue where saved password was used after it was unchecked


ebbit1q:
I still see *saved* when logging in

ebbit1q:
if I have a saved password and uncheck save password then log in it will still use the saved password

ebbit1q:
it's now almost impossible to get rid of the saved password, even after logging in as a different user without saving the password it will retain the username and password of the first user

seavor:
Hmmm, I'll take another pass at it, good catch.

brentmclark:
> Hmmm, I'll take another pass at it, good catch.

Is it time to take a fresh look at the `react-final-form` mutators?  I'm happy to pair up on that or take it on as an upcoming task.

seavor:
> > Hmmm, I'll take another pass at it, good catch.
> 
> Is it time to take a fresh look at the `react-final-form` mutators? I'm happy to pair up on that or take it on as an upcoming task.

Might not be a bad idea, this bit of functionality (save/autoconnect) has gotten quite a bit icky.

ebbit1q:
if I save my password, then log out, the password field says \*saved\* yet I can enter text if I want to, it doesn't matter what text I add it will not forget my password and only use the save password, I can then try to log in with a different account without changing the password (probably failing) and then go back to the previous account and it will still have remembered the password, I think as soon as you change one of these fields and log in it should lose your password

seavor:
> the password field is no longer labeled password when displaying _SAVED_
> the username is not retained when the password is not saved

Expected

> when registering an account the user name of the previous user is not replaced but the password is

Will fix.

ebbit1q:
the password field not showing that it is a password field is rather confusing, instead have it say \* saved password \* or something so it's clear what is being saved

webclient/src/containers/Login/Login.tsx
```diff
@@ -166,7 +148,10 @@ const Login = ({ state, description }: LoginProps) => {
   };
 
   const handleAccountActivationDialogSubmit = ({ token }) => {
-    AuthenticationService.activateAccount({ token } as any);
+    AuthenticationService.activateAccount({
```
ZeldaZach, line 151: passing as any, do we want to still do this? Or make a better interface?

ebbit1q (approved):
deployed version to https://beta.cockatrice.us/webatrice works as expected
\* SAVED PASSWORD \* all caps is a bit aggressive but it's fine
after creating a new account with save password checked you are logged in automatically as expected but it will not save your password, this is a problem for later that might need some ui work

PR #4564, opened 2022-02-06
Webatrice: improve prebuild steps and add .env configs

seavor:
Creates an `.env` file for server configuration.


tooomm:
Take the version from package.json? The commit hash can be appended in brackets still?

ebbit1q:
does this also handle setting the homepage?

seavor:
> Take the version from package.json?

Autoincrementing the package.json::version is a bit more difficult, and we would still need to grab the value and place it in the .env file. Using the commit hash feels less user-error prone.

> does this also handle setting the homepage?

No, this is something i read today is required by create-react-app. I dont think there's anyway of knowing what the server's intended domain/homepage is.

brentmclark:
If I'm reading this correctly, the `.env` file would be overwritten every time `prebuild` is ran regardless of the environment (dev, test, prod, etc.).  Am I reading that correctly?

If yes, how does this interact with CRA's support for environment-specific `.env` files such as `.env.production` or `.env.test`?

ebbit1q:
instead of overwriting a .env file, wouldn't it make more sense to simply generate a `generated-version.json` file and .gitignore that?

seavor:
> If yes, how does this interact with CRA's support for environment-specific `.env` files such as `.env.production` or `.env.test`?

After reading the docs, `.env` serves as the default level, and more specific env files like `.env.production` will run after and overwrite any identical variables.

seavor:
> instead of overwriting a .env file, wouldn't it make more sense to simply generate a `generated-version.json` file and .gitignore that?

I think what I have here is fine for at least the first iteration, we can improve as the need arises.

brentmclark:
I'm deeply suspicious about this implementation.  I think it will bite us at some point, but I can't cite anything specific at this time.

Given that this entire app is in active development with no actual users, I'm open to merging this PR anyway in the interest of progress.  I hope we continue to look for an alternative that can return the `.env` file to normal usage soon.

seavor:
i have been moved by everyone's objections. Please see most recent changes.

ebbit1q (approved):
I like the new changes that merge the prebuild steps for the different systems into a crossplatform one

PR #4566, opened 2022-02-08
Webatrice: Registration toasts

brentmclark:
## Related Ticket(s)
- Completes [Show toast on successful account registration](https://trello.com/c/rFGFhJhL/18-show-toast-on-successful-account-registration)
- Completes [Show toast on successful password reset](https://trello.com/c/a0Hge38B/17-show-toast-on-successful-password-reset)
- Completes [show toast on successful registration validation](https://trello.com/c/LLnUy2At/19-show-toast-on-successful-registration-validation)

The only feedback a user gets indicating their registration was successful is that they become authenticated.  It's a bit jarring and could use a toast to help close the feedback look cleanly.

- A new `ToastContext` was added along with an accompanying `useToast` hook.  This ensures that the toast will remain on screen even if the containing component becomes unmounted.
- A new `registrationSuccess` event was piped through redux to allow an event to fire that ultimately renders the toast.

## Discussion Points
- I chose to use a local reducer in `ToastContext` instead of `redux`.  There's certainly a tradeoff between using the established global state management solution (`redux`) and using a localized alternative (`useReducer`).  I believe the tradeoff is worth it, though.  I place a high value on colocation, especially when the entire state management apparatus belongs to a sibling component.  I've seen this have a positive long-term impact on the maintainability of the component.  That said, I'm prepared to rip it out in exchange for `redux` if that's the preference.

- [`immutability-helper`](https://github.com/kolodny/immutability-helper) is worth considering for both `redux` _and_ any `useReducer` implementations.  It adds complexity in that a new DSL must be learned.  However, it's pretty lightweight and automatically handles a LOT of complexity w.r.t. immutability.  I opted against it for now, since it's not that big of a lift to ensure immutability manually in this reducer.

## Remaining Tasks
Since this a draft PR, I still have a few tasks to handle before this becomes ready for final review.
- [x] Toast on password reset
- [x] Toast on successful account validation

### DIFF
--- a/webclient/src/components/Toast/ToastContext.tsx
+++ b/webclient/src/components/Toast/ToastContext.tsx
@@ -1,11 +1,11 @@
-import * as React from 'react'
+import { createContext, FC, ReactChild, ReactNode, useContext, useEffect, useReducer } from 'react'
 
 import { ACTIONS, initialState, reducer } from './reducer';
 import Toast from './Toast'
 
 interface Toast {
   isOpen: boolean,
-  children: React.ReactChild,
+  children: ReactChild,
 }
 
 interface ToastContext {
@@ -16,7 +16,7 @@ interface ToastContext {
     removeToast: (key) => void,
 }
 
-const ToastContext: any = React.createContext({
+const ToastContext: any = createContext({
   toasts: new Map(),
   addToast: (key, children) => {},
   openToast: (key) => {},
@@ -24,9 +24,9 @@ const ToastContext: any = React.createContext({
   removeToast: (key) => {},
 });
 
-const ToastProvider: React.FC<React.ReactNode> = (props) => {
+export const ToastProvider: FC<ReactNode> = (props) => {
   const { children } = props
-  const [state, dispatch] = React.useReducer(reducer, initialState)
+  const [state, dispatch] = useReducer(reducer, initialState)
   const providerState = {
     toasts: state.toasts,
     addToast: (key, children) => dispatch({ type: ACTIONS.ADD_TOAST, payload: { key, children } }),
@@ -51,10 +51,15 @@ const ToastProvider: React.FC<React.ReactNode> = (props) => {
   )
 }
 
-function useToast({ key, children }) {
-  const { addToast, openToast, closeToast, removeToast } = React.useContext(ToastContext)
+export interface ToastHookOptions {
+  key: string,
+  children: ReactNode
+}
 
-  React.useEffect(() => {
+export function useToast<ToastHookOptions>({ key, children }) {
+  const { addToast, openToast, closeToast, removeToast } = useContext(ToastContext)
+
+  useEffect(() => {
     addToast(key, children)
   }, [])
 
@@ -63,9 +68,4 @@ function useToast({ key, children }) {
     closeToast: () => closeToast(key),
     removeToast: () => removeToast(key),
   }
-}
-
-export {
-  ToastProvider,
-  useToast
 }
--- a/webclient/src/components/Toast/ToastContext.tsx
+++ b/webclient/src/components/Toast/ToastContext.tsx
@@ -52,7 +52,7 @@ const ToastProvider: React.FC<React.ReactNode> = (props) => {
 }
 
 function useToast({ key, children }) {
-  const { addToast, openToast, closeToast, removeToast, toasts } = React.useContext(ToastContext)
+  const { addToast, openToast, closeToast, removeToast } = React.useContext(ToastContext)
 
   React.useEffect(() => {
     addToast(key, children)
--- a/webclient/src/components/Toast/ToastContext.tsx
+++ b/webclient/src/components/Toast/ToastContext.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react'
+
+import { ACTIONS, initialState, reducer } from './reducer';
+import Toast from './Toast'
+
+interface Toast {
+  isOpen: boolean,
+  children: React.ReactChild,
+}
+
+interface ToastContext {
+    toasts: Toast[],
+    addToast: (key, children) => void,
+    openToast: (key) => void,
+    closeToast: (key) => void,
+    removeToast: (key) => void,
+}
+
+const ToastContext: any = React.createContext({
+  toasts: new Map(),
+  addToast: (key, children) => {},
+  openToast: (key) => {},
+  closeToast: (key) => {},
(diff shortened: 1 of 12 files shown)


webclient/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const ToastContext: any = React.createContext({
+  toasts: new Map(),
+  addToast: (key, children) => {},
+  openToast: (key) => {},
+  closeToast: (key) => {},
+  removeToast: (key) => {},
+});
+
+const ToastProvider: React.FC<React.ReactNode> = (props) => {
```
seavor, line 27:
```suggestion
const ToastProvider: FC<ReactNode> = (props) => {
```

webclient/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { children } = props
+  const [state, dispatch] = React.useReducer(reducer, initialState)
```
seavor, line 29:
```suggestion
  const [state, dispatch] = useReducer(reducer, initialState)
```

webclient/src/components/Toast/index.ts
```diff
@@ -0,0 +1,8 @@
+import { useToast, ToastProvider } from './ToastContext';
+import Toast from './Toast';
+
+export {
+  Toast as default,
```
seavor, line 5: Toast is just an interface, lets just export it normal.

brentmclark: Great catch!  This was from a previous iteration and is only meant for use within the component itself.

Do you think it's worth the effort to expose a `ToastHookOptions` for the `useToast` hook in it's place?

webclient/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react'
```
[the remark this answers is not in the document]
brentmclark: FWIW I do this to disambiguate between the react hooks and the homegrown hooks at-a-glance.  It's not that big of a deal to me; I'll stop with the `React.{whatever}` namespacing and destructure everything at the import instead.

webclient/src/containers/Login/Login.tsx
```diff
@@ -68,6 +69,8 @@ const Login = ({ state, description }: LoginProps) => {
   });
   const [userToResetPassword, setUserToResetPassword] = useState(null);
 
+  const passwordResetToast = useToast({key: 'password-reset-success', children: 'Password Reset Successfully'})
```
seavor, line 72: I love how easy this is to use.

webclient/src/components/Toast/index.ts
seavor: For sure, more typing the better.

webclient/src/containers/Login/Login.tsx
brentmclark: Thanks!  I'm pretty happy with it too.  I've grown really enamored with this stateful-provider + hook pattern over the past few years.  The DX has been great.

webclient/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+interface ToastContext {
+    toasts: Toast[],
+    addToast: (key, children) => void,
+    openToast: (key) => void,
+    closeToast: (key) => void,
+    removeToast: (key) => void,
+}
+
+const ToastContext: any = createContext({
```
seavor, line 19:
```suggestion
const toastContext: ToastContext = createContext({
```

seavor: One of these `ToastContext` needs to be named something different.

webclient/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+  addToast: (key, children) => {},
+  openToast: (key) => {},
+  closeToast: (key) => {},
+  removeToast: (key) => {},
+});
+
+export const ToastProvider: FC<ReactNode> = (props) => {
+  const { children } = props
+  const [state, dispatch] = useReducer(reducer, initialState)
+  const providerState = {
```
seavor, line 30:
```suggestion
  const providerState: ToastContext = {
```

seavor: Also, why are these methods defined here in a separate object and not defined in the params passed to `createContext`? Those seem like implementations, but they're noops.

webclient/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import { ACTIONS, initialState, reducer } from './reducer';
+import Toast from './Toast'
+
+interface Toast {
+  isOpen: boolean,
+  children: ReactChild,
+}
+
+interface ToastContext {
```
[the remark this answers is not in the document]
brentmclark: `ToastState` it shall be.

webclient/src/components/Toast/ToastContext.tsx
brentmclark: `createContext` accepts a `defaultValue` which needs to match the "shape" that the consumers expect.  [The docs](https://reactjs.org/docs/context.html#updating-context-from-a-nested-component) cover this briefly.

In my own words, the default value needs to have all of the same properties and methods that the final state will have or we risk runtime errors.

webclient/src/containers/App/AppShell.tsx
```diff
@@ -23,15 +24,17 @@ class AppShell extends Component {
   render() {
     return (
       <Provider store={store}>
-        <CssBaseline />
-        <div className="AppShell" onContextMenu={this.handleContextMenu}>
-          <Router>
-            <Header />
+        <ToastProvider>
+          <CssBaseline />
```
seavor, line 28: I think we still want `<CssBaseline />` to come directy after `<Provider store={store}>`.

brentmclark: Seems reasonable.

ebbit1q:
@seavor do you have any remarks left on this pr?

seavor:
> @seavor do you have any remarks left on this pr?

nope, once the CssBaseline directive is moved this PR is good to go.

seavor (approved):

Issue #4540, opened 2022-01-22
profile picture downscaling does not work


ebbit1q:
if you currently try to upload a profile picture larger than the allowed max of MAX_FILE_LENGTH it'll upload a blank image instead

Issue #750, opened 2015-02-16
Server allows user profile images of ridiculous size


Daenyth:
Look at Ninja Ferret; black image in 7136x2992 resolution

ZeldaZach:
We can add a restriction to the uploader on the webUI service to, say, 300x300?

Daenyth:
I'd like to enforce something in the main code. Probably when we allow
profile modification in client

woogerboy21:
@ctrlaltca Was this worked in already?

ctrlaltca:
Not really; in-client profile editing forces the image to a max 400x200 jpeg. The problem is also that the server's database column is a BLOB, limited to 64kb.

ZeldaZach:
If it's forced to a maximum size, then it seems like this is kinda-fixed

tooomm:
So if somebody uploads a picture within the client it will get resized to 400x200?

I just tried it myself and it looks like the pictures are getting shrinked + they keep the original image ratio.
Do Cockatrice compress them as well to match the 64kb limit?

ctrlaltca:
> So if somebody uploads a picture within the client it will get resized to 400x200?

It will be shrunk (respecting keeping its original width/height ratio) to fit inside that size.

> Does Cockatrice compress them before uploading as well to match the 64kb limit?

No, the image is not compressed before uploading; the raw image ends up in the database, and if it's more that 64kb, it will just be badly truncated by the database. This is indeed a problem that needs to be fixed.

tooomm:
So the issue still standing here is:
>the raw image ends up in the database, and if it's more that 64kb, it will just be badly truncated by the database. This is indeed a problem that needs to be fixed.

Also, there is the different handling for image uploading in-client and the online form as decribed in https://github.com/Cockatrice/Cockatrice/issues/750#issuecomment-181257858

D0nV170:
So issues still persist with images. I was able to get a user to upload ~100kb image with no issues but when she tried to upload a larger image size, when she logged off it greyed out like the last 10% of the image. I saw the issue again on another user and asked how large his avatar image was and he told me ~500kb and almost all of it was greyed out. This seems only to happen at logoff and the image is fixed the next time it's uploaded.

ZeldaZach:
Temp fix was to change field from BLOB to MEDIUMBLOB. I'd still like to add reduction code to the service so it fits into a BLOB

ZeldaZach:
This is related to #793

PR #4567, opened 2022-02-08
fix image downscaling

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4540

the image was cleared instead of downscaled

- just make a new buffer instead of clearing it

the default size of the database is not 0xfffff but rather 0xffff, on rooster ranges this limit is increased to 0xffffff.
these limits correspond with the sql blob and mediumblob field types, if we want servers to accept the image size we currently enforce on pictures by default we have three options:
- create a smaller definition of profile images at 0xffff, meaning profile pictures will be smaller, we should consider more sophisticated downscaling code on the client side in this case
- make a database migration changing the field from blob to medium blob by default, this is what is currently on rooster ranges
- make a configurable limit that the server enforces, this would mean a protocol expansion to communicate the limit to clients

### DIFF
--- a/cockatrice/src/dlg_edit_avatar.cpp
+++ b/cockatrice/src/dlg_edit_avatar.cpp
@@ -74,13 +74,12 @@ QByteArray DlgEditAvatar::getImage()
         return QByteArray();
     }
 
-    QByteArray ba;
-    QBuffer buffer(&ba);
-    buffer.open(QIODevice::WriteOnly);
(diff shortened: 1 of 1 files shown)


ebbit1q:
related https://github.com/Cockatrice/Cockatrice/issues/750

ZeldaZach (approved):

PR #4576, opened 2022-02-14
Fix new spoiler notifications

tooomm:
## Related Ticket(s)
- https://github.com/Cockatrice/Magic-Spoiler/pull/251

Currently, as announced in the related PR, notifications for spoiler do currently not trigger notifications when new data is available and updated. The reason for this is the update to db v4 for the spoiler file and utilizing the new `<info>` key instead our former method to hide the update time stamp in a comment.

Old format (https://github.com/Cockatrice/Magic-Spoiler/blob/f347bfdbbfbb3667572ad43c1a926e9ee53595cb/spoiler.xml#L3)
``` 
  <!--
  Created At: Sat, Feb 05 2022, 00:30:45 (UTC)
  (...)
  -->
```

New format: (https://github.com/Cockatrice/Magic-Spoiler/blob/17588cea792e634bd3e2198a7034ef9cfd817f13/spoiler.xml#L8)
```
  <info>
    (...)
    <createdAt>2022-02-14 13:32:28 (UTC)</createdAt>
    (...)
  </info>
```

The new information is placed a few lines down compared to the older one, but is still captured when [only reading the first 512 bytes of the file](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/spoilerbackgroundupdater.cpp#L193-L214).

- Adapt for new placement and format

## Screenshot(s)
Untitled

<br>

**Note:** This is only a quick, simple fix, and still relies on the same approach to get the data.

**Note 2:** The "Last Change" value displayed in the settings dialog under `Card Sources` does display the last modification time of the spoiler file instead the actual time stamp of the data creation. Thus it does not align with the time information shown in the "new spoiler data available" notification shown. (https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/dlg_settings.cpp#L772)

### DIFF
--- a/cockatrice/src/spoilerbackgroundupdater.cpp
+++ b/cockatrice/src/spoilerbackgroundupdater.cpp
@@ -171,7 +171,7 @@ bool SpoilerBackgroundUpdater::saveDownloadedFile(QByteArray data)
                 QString timeStamp = QString(line).replace("<createdAt>", "").trimmed();
                 timeStamp.chop(18); // Remove " (UTC)</createdAt>"
 
-                auto utcTime = QLocale().toDateTime(timeStamp, "yyyy-mm-dd hh:mm:ss");
+                auto utcTime = QLocale().toDateTime(timeStamp, "yyyy-MM-dd hh:mm:ss");
                 utcTime.setTimeSpec(Qt::UTC);
 
                 QString localTime = utcTime.toLocalTime().toString("MMM d, hh:mm");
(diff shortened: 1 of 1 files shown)


ebbit1q:
this is not going to be included on older clients, I want the old comment back

tooomm:
>I want the old comment back

May I [quote you](https://github.com/Cockatrice/Cockatrice/pull/4116#pullrequestreview-498395584)?
*"this comment parsing code is ridiculous though, this should be handled differently using xml"*

<br>

>this is not going to be included on older clients

Correct. We can keep the comment next to the new values for a transition period if really wanted...

ebbit1q:
> Correct. We can keep the comment next to the new values for a transition period if really wanted...

this is what I mean

tooomm:
Done in https://github.com/Cockatrice/Magic-Spoiler/pull/292

ebbit1q:
thank you for re adding the previous system, I'll close this pr so we can work on properly interpreting the new value instead of continuing the same method without improvement

PR #4584, opened 2022-02-27
Webatrice: i18n login screen

seavor:
## Related Ticket(s)
- Implements https://trello.com/c/Zcu0xMLH/6-translation-support


webclient/src/common.i18n.json
```diff
@@ -1,6 +1,27 @@
...
+      "password": "Password",
+      "port": "Port",
+      "realName": "Real Name",
+      "saveChanges": "Save Changes",
+      "token": "Token",
+      "username": "Username"
+    },
+    "validation": {
+      "minChars_one": "Minimum of {{count}} character required",
+      "minChars_other": "Minimum of {{count}} characters required",
```
seavor, line 22: @ebbit1q here's your plural example.

ebbit1q: this does not look like the example in transifex's docs https://docs.transifex.com/formats/json

seavor: tl/dr it seems like we'll be fine, i dont see a way to reconcile i18n-next format with either ICU or Structured json formats in Transfex.

From their docs:
> Example:
>
> "files" : "{count, plural, one {You have {count} file.} other {You have {count} files.}}"
>
> For pluralized strings that follow a different structure like:
>
> "string" : "{number, plural, =1 {1 New}, =2 {# New}}"
>
> the Structured JSON parser will handle them as simple entries in the editor as follows:

Structured_JSON_Plurals_Support

brentmclark: What does this mean for us operationally?  Will translations need to be done twice per language?

brentmclark (commented):
Overall these changes make sense to me.  Nice working converting this big section over 💪 

One question I have is about the value of the sub-namespacing approach in the translation files.  e.g. `label`, `validation`, etc.  Can you tell me more about your approach there?

webclient/src/common.i18n.json
seavor: We'll need to provide both either way. `i18n-next` wants the following:
```
"key_one": "Minimum of {{count}} character required",
"key_other": "Minimum of {{count}} characters required",
```

Transifex wants this:
```
"key": "{prop, plural, one {{prop} translation} other {{prop} translations}}"
```

I belive Transifex with help the translators by showing them as group vs separate translations, which would be a bummer to lose, but not the end of the world.

seavor:
>  e.g. `label`, `validation`, etc. Can you tell me more about your approach there?

It's just a way of grouping pieces of the view in ways that are easier to find, as opposed to one large flat object.

webclient/src/common.i18n.json
seavor: Oh, it looks like `i18n-next` has a plugin for us to use the ICU format. I'll look into adding this in with this PR.

https://react.i18next.com/misc/using-with-icu-format

ebbit1q: we will see how it performs when this is merged, translations are picked up automatically

webclient/src/common.i18n.json
```diff
@@ -1,6 +1,27 @@
...
+      "password": "Password",
+      "passwordAgain": "Password Again",
+      "port": "Port",
+      "realName": "Real Name",
+      "saveChanges": "Save Changes",
+      "token": "Token",
+      "username": "Username"
+    },
+    "validation": {
+      "minChars": "Minimum of {count} {count, plural, one {character} other {characters}} required",
```
seavor, line 22: @ebbit1q here is the new plural example using the ICU format.

webclient/src/dialogs/KnownHostDialog/KnownHostDialog.i18n.json
```diff
@@ -0,0 +1,6 @@
+{
+ "KnownHostDialog": {
+    "title": "{mode, select, edit {Edit} other {Add}} Known Host",
```
seavor, line 3: @ebbit1q here is an example using the `select` operator, similar to the `plural` operator.

webclient/src/containers/Initialize/Initialize.i18n.json
```diff
@@ -0,0 +1,6 @@
+{
+  "InitializeContainer": {
+    "title": "DID YOU KNOW",
+    "subtitle": "<1>Cockatrice is run by volunteers</1><1>that love card games!</1>"
```
seavor, line 4: @ebbit1q here is an example using strings that have DOM components interspersed. See the `Trans` component in the controller.

ebbit1q (approved):
seems good

ZeldaZach (approved):
Makes sense to me

PR #4587, opened 2022-03-05
add fr and nl translations

seavor:


ebbit1q:
on beta the password field is untranslated also the country dropdown when registering is completely untranslated why do we have every name of the country in english?

ebbit1q:
correction, the password field is only translated when you reload the page

seavor:
> the country dropdown when registering is completely untranslated why do we have every name of the country in english?

We have a follow up task to translate the country dropdown.

ebbit1q:
do we really want to translate every country name into every language?

seavor:
> do we really want to translate every country name into every language?

Part of the hold up is i think some discussion is needed about what exactly we what to do there.

webclient/src/types/languages.ts
```diff
@@ -1,11 +1,13 @@
 import { CountryLabel } from './countries';
 
 export enum Language {
-  'en-US' = 'en-US',
-  'es-ES' = 'es-ES',
+  'en' = 'en',
```
ebbit1q, line 4: the client currently uses en-US but fr and de for most languages, we also have pt-BR and zh-hans as examples, yet on the client this is all internal, I recommend adding what the language calls itself as a translatable tag

ebbit1q (dismissed):
works on beta

webclient/src/types/languages.ts
seavor: region-based translations have been another TBD discussion item, lets make it a follow up.

webclient/src/components/LanguageDropdown/LanguageDropdown.tsx
```diff
@@ -36,7 +36,7 @@ const LanguageDropdown = () => {
             return (
               <MenuItem value={lang} key={lang}>
                 <div className="LanguageDropdown-item">
-                  <img className="LanguageDropdown-item__image" src={Images.Countries[country]} alt={CountryLabel[country]} />
```
seavor, line 39: alt text intentionally removed as it rendered on top of the label.

webclient/src/common.i18n.json
```diff
@@ -22,6 +22,258 @@
...
+      "CN": "China",
+      "CX": "Christmas Island",
+      "CC": "Cocos (Keeling) Islands",
+      "CO": "Colombia",
+      "KM": "Comoros",
+      "CG": "Congo",
+      "CD": "The Democratic Republic of the Congo",
+      "CK": "Cook Islands",
+      "CR": "Costa Rica",
+      "CI": "Côte d\"Ivoire",
```
ebbit1q, line 81: why \\" ? also why is ivory coast not translated?

webclient/src/components/CountryDropdown/CountryDropdown.tsx
```diff
@@ -3,15 +3,28 @@ import React, { useEffect, useState } from 'react';
...
 
 const CountryDropdown = ({ input: { onChange } }) => {
   const [state, setState] = useState('');
+  const [sortedCountries, setSortedCountries] = useState([]);
+  const { t, i18n } = useTranslation();
 
   useEffect(() => onChange(state), [state]);
 
+  useEffect(() => {
+    const collator = new Intl.Collator(i18n.language);
```
seavor, line 21: https://www.measurethat.net/Benchmarks/Show/12644/1/intlcollatorcompare-vs-localecompare

webclient/src/common.i18n.json
ebbit1q: hey, if you're renaming that one what about fixing saint helena? what about north macedonia? how many countries recognize kosovo (XK)?

ebbit1q (approved):
I guess fixing up the contents of the country list is out of scope, good job on the sorting, looks good even though the menu is ever so sluggish

seavor:
> I guess fixing up the contents of the country list is out of scope, good job on the sorting, looks good even though the menu is ever so sluggish

I created tickets for both of these issues

PR #4596, opened 2022-03-19
test connection UI

seavor:
## Related Ticket(s)
- Implements https://trello.com/c/mjcfYOF9/4-test-connection-in-add-edit-known-host-modal

Screen Shot 2022-03-19 at 3 52 28 PM


webclient/src/components/KnownHosts/KnownHosts.tsx
```diff
@@ -152,20 +193,31 @@ const KnownHosts = (props) => {
     );
   };
 
+  const testConnection = () => {
+    setTestingConnection(s => ({
+      testing: true,
```
ZeldaZach, line 198: Would an ENUM be better compared to three unique booleans? Is that something TS can support?

webclient/src/websocket/commands/SessionCommands.ts
```diff
@@ -25,6 +25,9 @@ export class SessionCommands {
       case WebSocketConnectReason.PASSWORD_RESET:
         SessionCommands.updateStatus(StatusEnum.CONNECTING, 'Connecting...');
         break;
+      case WebSocketConnectReason.TEST_CONNECTION:
+        webClient.testConnect({ ...options });
```
ZeldaZach, line 29: Would passing options be the same as what you're doing?

webclient/src/components/KnownHosts/KnownHosts.tsx
seavor: Not really, we'd then have to do value comparisons down where we use these booleans.

webclient/src/websocket/commands/SessionCommands.ts
seavor: Yes, it's just a convention to ensure immutability.

ebbit1q (approved):
this is working as expected, would like some input on icon choices still though

seavor:
> this is working as expected, would like some input on icon choices still though

We'll get some feedback from Lee on a follow up.

PR #4599, opened 2022-03-21
Added Chickatrice.net to Hosts

Lachee:
Chickatrice has sorted out its SSL for websockets, and webatrice can now connect to it without any issues.
This PR adds it to the known hosts.

Chickatrice was not present in the default hosts.

- Adds chickatrice to the list of default hosts and known hosts.

## Additional
I have placed Chickatrice alphabetically, but it is completely reasonable for this early stage of development to want Rooster ranges to be default since it is most likely to have the most compatible server. The Chickatrice entry can be easily moved down between Tetrarch and Rooster Beta.

### DIFF
--- a/webclient/src/types/server.tsx
+++ b/webclient/src/types/server.tsx
@@ -55,6 +55,12 @@ export class Host {
 
 export const DefaultHosts: Host[] = [
   {
+    name: 'Chickatrice',
+    host: 'mtg.chickatrice.net/servatrice',
+    port: '443',
+    editable: false,
+  },
+  {
     name: 'Rooster',
     host: 'server.cockatrice.us/servatrice',
     port: '4748',
@@ -93,11 +99,13 @@ export const getHostPort = (host: Host): { host: string, port: string } => {
 };
 
 export enum KnownHost {
+  CHICKATRICE = 'Chickatrice',
   ROOSTER = 'Rooster',
   TETRARCH = 'Tetrarch',
 }
 
 export const KnownHosts = {
+  [KnownHost.CHICKATRICE]: { port: 443, host: 'mtg.chickatrice.net', },
   [KnownHost.ROOSTER]: { port: 4748, host: 'server.cockatrice.us', },
   [KnownHost.TETRARCH]: { port: 443, host: 'mtg.tetrarch.co/servatrice' },
 }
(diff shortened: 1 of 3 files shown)


webclient/src/types/server.tsx
```diff
@@ -54,6 +54,12 @@ export class Host {
 }
 
 export const DefaultHosts: Host[] = [
+  {
```
seavor, line 57: So, politics time.. the dropdown will be ordered by this list. Rooster seems to be the main server, which is why I have it listed first, as it will be selected automatically by default.

Otherwise, this looks fine. You might want to play with the `localHost` value, and try host and localHost with and without /servatrice to see if it works locally with `ws`.

Lachee: every other server list has been alphabetical.
Yeah ill give that localHost a try

webclient/src/types/server.tsx
```diff
@@ -93,11 +99,13 @@ export const getHostPort = (host: Host): { host: string, port: string } => {
 };
 
 export enum KnownHost {
+  CHICKATRICE = 'Chickatrice',
```
seavor, line 102: Fairly certain these are no longer used, if you can be bothered, you can remove this enum and the below const entirely, and just make sure the build still runs fine.

webclient/src/types/server.tsx
```diff
@@ -54,6 +54,12 @@ export class Host {
 }
 
 export const DefaultHosts: Host[] = [
+  {
+    name: 'Chickatrice',
+    host: 'mtg.chickatrice.net/servatrice',
+    port: '443',
```
seavor, line 60: This seemed to work for me, testing on beta (wss) and a locally running instance (ws).

```suggestion
    host: 'mtg.chickatrice.net',
    port: '443',
    localPort: '4748
```

tooomm:
Why do we try to maintain a second list of servers if we have a central place for them already?
 --> https://github.com/Cockatrice/cockatrice.github.io/blob/master/public-servers.json

Lachee:
> Why do we try to maintain a second list of servers if we have a central place for them already? --> https://github.com/Cockatrice/cockatrice.github.io/blob/master/public-servers.json

I had assumed for ease of development at this early stage. It would be trivial to add a extra fetch for that public server list, but eitherway more information would need to be added to that list to state if it supports webatrice (I mean, all public servers do now)

ebbit1q:
tip for prs: it's preferred to use a separate branch instead of your master branch for pull requests, you'll find this more convenient if you were to make another pull request as you can split off and delete branches easily

tooomm:
Server got added in #4606 already.

PR #4603, opened 2022-04-01
CI: fix if condition

tooomm:
Condition was not correct, see https://github.com/Cockatrice/Magic-Spoiler/pull/285.
As seen in https://github.com/Cockatrice/Cockatrice/pull/4599, it [triggered for Lachee's fork](https://github.com/Lachee/Cockatrice/actions/runs/2074593563).

This is a fix taken from #4544, which is still lacking approval.


ebbit1q (approved):
ah, we forgot about that, thanks

PR #4607, opened 2022-04-13
Move to Qt6, Qt5.8.0, and Graceful Server Exit

ZeldaZach:
This moves Servatrice to Qt 6.3.0, which skips the memory leak issue found in 5.15.X

Also bumps min required Qt5 to 5.8.0 to support timing better

Also fixes some issues noticed while transitioning and allowing servatrice to exit for Valgrind logs


CMakeLists.txt
```diff
@@ -165,23 +171,32 @@ elseif(DEFINED ENV{QTDIR})
     list(APPEND CMAKE_PREFIX_PATH "$ENV{QTDIR}")
 endif()
 
-FIND_PACKAGE(Qt5Core 5.5.0 REQUIRED)
+# Detect Qt6 then Qt5, unless user forces Qt5
+# This is the ONLY place the versions of Qt are specified
+IF(NOT FORCE_USE_QT5)
+    FIND_PACKAGE(Qt6 6.2.3 COMPONENTS Concurrent Network Svg Widgets LinguistTools Core5Compat Gui WebSockets)
```
ZeldaZach, line 177: Changing to 6.3.0

CMakeLists.txt
```diff
@@ -165,23 +171,32 @@ elseif(DEFINED ENV{QTDIR})
...
+ENDIF()
+IF(Qt6_FOUND)
+    SET(QT_VERSION_MAJOR Qt6)
+ELSE()
+    SET(QT_VERSION_MAJOR Qt5)
+    FIND_PACKAGE(Qt5 5.8.0 COMPONENTS Concurrent Network Svg Widgets LinguistTools REQUIRED)
```
ZeldaZach, line 183: Need to call out in ticket

CMakeLists.txt
```diff
@@ -231,15 +246,23 @@ if(UNIX)
         # linux
         IF(CPACK_GENERATOR STREQUAL "RPM")
             set(CPACK_RPM_PACKAGE_LICENSE "GPLv2")
-            set(CPACK_RPM_PACKAGE_REQUIRES "protobuf, qt5-qttools, qt5-qtsvg, qt5-qtmultimedia")
+            IF(QT6_FOUND)
+                SET(CPACK_RPM_PACKAGE_REQUIRES "protobuf, qt6-qttools, qt6-qtsvg, qt6-qtmultimedia")
```
ZeldaZach, line 250: Need to confirm these are legit

CMakeLists.txt
```diff
@@ -231,15 +246,23 @@ if(UNIX)
...
             set(CPACK_RPM_PACKAGE_GROUP "Amusements/Games")
             set(CPACK_RPM_PACKAGE_URL "http://github.com/Cockatrice/Cockatrice")
         ELSE()
             set(CPACK_GENERATOR DEB)
             set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
             set(CPACK_DEBIAN_PACKAGE_SECTION "games")
             set(CPACK_DEBIAN_PACKAGE_HOMEPAGE "http://github.com/Cockatrice/Cockatrice")
-            set(CPACK_DEBIAN_PACKAGE_DEPENDS "libqt5multimedia5-plugins, libqt5svg5")
+            IF(QT6_FOUND)
+                set(CPACK_DEBIAN_PACKAGE_DEPENDS "libqt6multimedia6-plugins, libqt6svg6")
```
ZeldaZach, line 262: Need to confirm these are legit

cockatrice/src/dlg_settings.cpp
```diff
@@ -186,15 +190,21 @@ QStringList GeneralSettingsPage::findQmFiles()
...
 
 QString GeneralSettingsPage::languageName(const QString &qmFile)
 {
     QTranslator qTranslator;
-    qTranslator.load(translationPrefix + "_" + qmFile + ".qm", translationPath);
+
+    const auto fileName = translationPrefix + "_" + qmFile + ".qm";
+    const auto fileLoaded = qTranslator.load(fileName, translationPath);
+    if (!fileLoaded) {
+        qWarning() << "Unable to load translation file" << translationPath + fileName;
```
ZeldaZach, line 205: Log will look weird

cockatrice/src/main.cpp
```diff
@@ -67,10 +67,26 @@ void installNewTranslator()
...
+    const auto fileName0 = "qt_" + lang;
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    const auto dir0 = QLibraryInfo::path(QLibraryInfo::TranslationsPath);
+#else
+    const auto dir0 = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
+#endif
+
+    const auto fileLoaded0 = qtTranslator->load(fileName0, dir0);
+    if (!fileLoaded0) {
+        qWarning() << "Unable to load file" << dir0 + fileName0;
```
ZeldaZach, line 79: log will look weird

cockatrice/src/main.cpp
```diff
@@ -67,10 +67,26 @@ void installNewTranslator()
...
+    }
     qApp->installTranslator(qtTranslator);
-    translator->load(translationPrefix + "_" + lang, translationPath);
+
+    const auto fileName1 = translationPrefix + "_" + lang;
+    const auto fileLoaded1 = translator->load(fileName1, translationPath);
+    if (!fileLoaded1) {
+        qWarning() << "Unable to load translation file" << translationPath + fileName1;
```
ZeldaZach, line 86: log will look weird

common/server_player.cpp
```diff
@@ -1235,22 +1235,25 @@ Server_Player::cmdAttachCard(const Command_AttachCard &cmd, ResponseContainer &
...
-    while (playerIterator.hasNext()) {
-        Server_Player *p = playerIterator.next().value();
-        QList<Server_Arrow *> arrows = p->getArrows().values();
-        QList<Server_Arrow *> toDelete;
-        for (auto a : arrows) {
-            auto *tCard = qobject_cast<Server_Card *>(a->getTargetItem());
-            if ((tCard == card) || (a->getStartCard() == card)) {
-                toDelete.append(a);
+    {
+        QMutexLocker locker(&game->gameMutex);
```
ZeldaZach, line 1239: I don't think this fixed the issue, as I saw it appear again

dbconverter/CMakeLists.txt
```diff
@@ -13,21 +13,37 @@ SET(dbconverter_SOURCES
     ${VERSION_STRING_CPP}
 )
 
-# Qt5
-find_package(Qt5 COMPONENTS Network Widgets REQUIRED)
-set(dbconverter_QT_MODULES Qt5::Core Qt5::Network Qt5::Widgets)
+# Detect Qt6 then Qt5, unless user forces Qt5
+IF(NOT FORCE_USE_QT5)
```
ZeldaZach, line 17: Documentation update to show this flag

oracle/src/main.cpp
```diff
@@ -22,9 +22,25 @@ void installNewTranslator()
...
+    const auto fileName1 = "qt_" + lang;
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    const auto dir = QLibraryInfo::path(QLibraryInfo::TranslationsPath);
+#else
+    const auto dir = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
+#endif
+
+    const auto fileLoaded1 = qtTranslator->load(fileName1, dir);
+    if (!fileLoaded1) {
+        qWarning() << "Unable to load translation file" << dir + fileName1;
```
ZeldaZach, line 34: log will look weird

oracle/src/main.cpp
```diff
@@ -22,9 +22,25 @@ void installNewTranslator()
...
+    }
+
     qApp->installTranslator(qtTranslator);
-    translator->load(translationPrefix + "_" + lang, translationPath);
+
+    const auto fileName2 = translationPrefix + "_" + lang;
+    const auto fileLoaded2 = qtTranslator->load(fileName2, translationPath);
+    if (!fileLoaded2) {
+        qWarning() << "Unable to load translation file" << translationPath + fileName2;
```
ZeldaZach, line 42: log will look weird

oracle/src/oraclewizard.cpp
```diff
@@ -170,15 +170,19 @@ QStringList IntroPage::findQmFiles()
...
 }
 
 QString IntroPage::languageName(const QString &qmFile)
 {
     QTranslator qTranslator;
-    qTranslator.load(translationPrefix + "_" + qmFile + ".qm", translationPath);
+    const auto fileName = translationPrefix + "_" + qmFile + ".qm";
+    const auto fileLoaded = qTranslator.load(fileName, translationPath);
+    if (!fileLoaded) {
+        qWarning() << "Unable to load translation file" << translationPath + fileName;
```
ZeldaZach, line 184: log will look weird

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1359,6 +1360,7 @@ bool Servatrice_DatabaseInterface::updateUserToken(const QString &token, const Q
     return false;
 }
 
+// TODO
```
ZeldaZach, line 1363: Look into why i put this here

ZeldaZach:
Note to self: Update GHA to download the appropriate QT versions, and setup new boxes to support Qt5/6

CMakeLists.txt
ZeldaZach: Actually 6.2.4, since that's what ships with 22.04 linux

CMakeLists.txt
ZeldaZach: plugins was combined to libqt6multimedia6 it seems

CMakeLists.txt
ZeldaZach: Validated

.ci/release_template.md
```diff
@@ -8,18 +8,22 @@ git push -d origin --REPLACE-WITH-BETA-LIST--
...
- - <kbd>Ubuntu 20.10</kbd> ("Groovy Gorilla")</i>
- - <kbd>Ubuntu 21.04</kbd> ("Hirsute Hippo")</i>
+ - <kbd>Ubuntu 21.10</kbd> ("Hirsute Hippo")</i>
+ - <kbd>Ubuntu 21.10</kbd> ("Impish Indri")</i>
+ - <kbd>Ubuntu 22.04</kbd> ("Jammy Jellyfish")</i>
  - <kbd>Debian 10</kbd> ("Buster")</i>
+ - <kbd>Debian 11</kbd> ("Bullseye")</i>
  - <kbd>Fedora 33</kbd></i>
  - <kbd>Fedora 34</kbd></i>
+ - <kbd>Fedora 35</kbd></i>
```
tooomm, line 26:
```suggestion
 - <kbd>Fedora 34</kbd></i>
 - <kbd>Fedora 35</kbd></i>
 - <kbd>Fedora 36</kbd></i>
```

.ci/docker.sh
```diff
@@ -43,7 +47,7 @@ while [[ $# != 0 ]]; do
     *)
       if [[ $1 == -* ]]; then
         echo "unrecognized option: $1"
-        return 3
+        exit 3
```
ebbit1q, line 50: no, the interactive option is just extra, the normal use is to source the script

.ci/docker.sh
```diff
@@ -57,7 +61,7 @@ if ! [[ $NAME ]]; then
   return 3
 fi
 
-export IMAGE_NAME="${project_name,,}_${NAME,,}" # lower case
+export IMAGE_NAME="${project_name,,}"_"${NAME,,}" # lower case
```
ebbit1q, line 64: why?

CMakeLists.txt
```diff
@@ -106,10 +106,14 @@ endif()
...
 
+IF(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.12)
+    add_compile_definitions(QT_DISABLE_DEPRECATED_BEFORE=0x050F00)
+ENDIF()
+
 # Define proper compilation flags
 IF(MSVC)
     # Visual Studio: Maximum optimization, disable warning C4251
-    set(CMAKE_CXX_FLAGS_RELEASE "/Ox /MD /wd4251 ")
+    SET(CMAKE_CXX_FLAGS_RELEASE "/Ox /MD /wd4251 /Zc:__cplusplus /std:c++17 /permissive-")
```
ebbit1q, line 116: can you expand the comment above this with these changes?

README.md
```diff
@@ -117,6 +117,7 @@ The following flags can be passed to `cmake`:
 - `-DWARNING_AS_ERROR=0` Whether to treat compilation warnings as errors in debug mode (default 1 = yes).
 - `-DUPDATE_TRANSLATIONS=1` Configure `make` to update the translation .ts files for new strings in the source code. Note: Running `make clean` will remove the .ts files (default 0 = no).
 - `-DTEST=1` Enable regression tests (default 0 = no). Note: needs googletest, will be downloaded on the fly if unavailable. To run tests: ```make test```.
+- `-FORCE_USE_QT5=1` Skip looking for Qt6 before trying to find Qt5
```
ebbit1q, line 120: should this be -D?

.ci/release_template.md
```diff
@@ -8,18 +8,22 @@ git push -d origin --REPLACE-WITH-BETA-LIST--
...
 <pre>
 <b>Pre-compiled binaries we serve:</b>
- - <kbd>Windows 7/8/10 (32-bit)</kbd></i>
- - <kbd>Windows 7/8/10 (64-bit)</kbd></i>
+ - <kbd>Windows 7/8/10/11 (32-bit)</kbd></i>
+ - <kbd>Windows 7/8/10/11 (64-bit)</kbd></i>
  - <kbd>macOS 10.14</kbd> ("Mojave")</i>
  - <kbd>macOS 10.15</kbd> ("Catalina")</i>
  - <kbd>macOS 11.0</kbd> ("Big Sur")</i>
+ - <kbd>macOS 12.0</kbd> ("Monterey")</i>
```
ebbit1q, line 16: this isn't actually working

.ci/release_template.md
```diff
@@ -8,18 +8,22 @@ git push -d origin --REPLACE-WITH-BETA-LIST--
...
  - <kbd>Ubuntu 18.04</kbd> ("Bionic Beaver")</i>
  - <kbd>Ubuntu 20.04</kbd> ("Focal Fossa")</i>
- - <kbd>Ubuntu 20.10</kbd> ("Groovy Gorilla")</i>
- - <kbd>Ubuntu 21.04</kbd> ("Hirsute Hippo")</i>
+ - <kbd>Ubuntu 21.10</kbd> ("Hirsute Hippo")</i>
```
ebbit1q, line 19: we should deprecate older versions

CMakeLists.txt
```diff
@@ -106,10 +106,14 @@ endif()
 # Treat warnings as errors (Debug builds only)
 option(WARNING_AS_ERROR "Treat warnings as errors in debug builds" ON)
 
+IF(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.12)
+    add_compile_definitions(QT_DISABLE_DEPRECATED_BEFORE=0x050F00)
```
ebbit1q, line 110: we should remove this and see if it works properly

cmake/FindQtRuntime.cmake
```diff
@@ -0,0 +1,53 @@
+# Find a compatible Qt version
+# Sets COCKATRICE_QT_VERSION_NAME -- Example values: Qt5, Qt6
+
+FIND_PACKAGE(Qt6 6.2.3 COMPONENTS
+        Concurrent
```
ebbit1q, line 5: I don't like how we require all components here, we should just require the minimum for each compile target so you can avoid unnecessary things for your server etc

cmake/FindQtRuntime.cmake
```diff
@@ -0,0 +1,53 @@
...
+IF(${COCKATRICE_QT_VERSION_NAME}_POSITION_INDEPENDENT_CODE OR Qt6_FOUND)
+    SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
+ENDIF()
+
+# guess plugins and libraries directory
+SET(QT_PLUGINS_DIR "${${COCKATRICE_QT_VERSION_NAME}Core_DIR}/../../../plugins")
+GET_TARGET_PROPERTY(QT_LIBRARY_DIR ${COCKATRICE_QT_VERSION_NAME}::Core LOCATION)
+GET_FILENAME_COMPONENT(QT_LIBRARY_DIR ${QT_LIBRARY_DIR} PATH)
+
+MESSAGE(STATUS "Found Qt ${${COCKATRICE_QT_VERSION_NAME}_VERSION} at: ${${COCKATRICE_QT_VERSION_NAME}_DIR}")
```
ebbit1q, line 53: this should be VERSION_STRING

cockatrice/src/abstractclient.cpp
```diff
@@ -52,6 +52,8 @@ AbstractClient::AbstractClient(QObject *parent)
     FeatureSet features;
     features.initalizeFeatureList(clientFeatures);
 
+    qRegisterMetaType<PendingCommand *>("pendingCommand");
```
ebbit1q, line 55: what does this do?

cockatrice/src/carddbparser/cockatricexml3.cpp
```diff
@@ -72,24 +73,24 @@ void CockatriceXml3Parser::loadSetsFromXml(QXmlStreamReader &xml)
...
                     longName = xml.readElementText(QXmlStreamReader::IncludeChildElements);
-                } else if (xml.name() == "settype") {
+                } else if (xml.name().toString() == "settype") {
                     setType = xml.readElementText(QXmlStreamReader::IncludeChildElements);
-                } else if (xml.name() == "releasedate") {
+                } else if (xml.name().toString() == "releasedate") {
                     releaseDate =
                         QDate::fromString(xml.readElementText(QXmlStreamReader::IncludeChildElements), Qt::ISODate);
-                } else if (xml.name() != "") {
+                } else if (xml.name().toString() != "") {
```
ebbit1q, line 93: comparisons to "" are probably useless, .isEmpty() or something would make more sense

cockatrice/src/carddbparser/cockatricexml3.cpp
```diff
@@ -23,7 +23,7 @@ bool CockatriceXml3Parser::getCanParseFile(const QString &fileName, QIODevice &d
     QXmlStreamReader xml(&device);
     while (!xml.atEnd()) {
         if (xml.readNext() == QXmlStreamReader::StartElement) {
-            if (xml.name() == COCKATRICE_XML3_TAGNAME) {
+            if (xml.name().toString() == COCKATRICE_XML3_TAGNAME) {
```
ebbit1q, line 26: we call xml.name().toString() so often it's obnoxious, we should just split this to a variable

cockatrice/src/carditem.h
```diff
@@ -2,6 +2,7 @@
 #define CARDITEM_H
 
 #include "abstractcarditem.h"
+#include "server_card.h"
```
ebbit1q, line 5: do we need this include? we already have a forward declaration a few lines down

cockatrice/src/chatview/chatview.cpp
```diff
@@ -510,10 +510,17 @@ void ChatView::redactMessages(const QString &userName, int amount)
     }
 }
 
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
```
ebbit1q, line 513: this guard only needs to guard one line, not the entire function

cockatrice/src/deckview.cpp
```diff
@@ -238,7 +238,11 @@ int DeckViewCardContainer::getCardTypeTextWidth() const
     QFontMetrics fm(f);
 
     int maxCardTypeWidth = 0;
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    QMultiMapIterator<QString, DeckViewCard *> i(cardsByType);
+#else
     QMapIterator<QString, DeckViewCard *> i(cardsByType);
+#endif
     while (i.hasNext()) {
```
ebbit1q, line 246: this could just be for const auto &key : cardsByType.keys {

cockatrice/src/dlg_settings.cpp
```diff
@@ -1235,11 +1240,12 @@ void ShortcutSettingsPage::retranslateUi()
 
 DlgSettings::DlgSettings(QWidget *parent) : QDialog(parent)
 {
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 13, 0))
-    auto rec = qApp->primaryScreen()->availableGeometry();
-#else
-    auto rec = QApplication::desktop()->availableGeometry();
-#endif
+    //#if (QT_VERSION >= QT_VERSION_CHECK(5, 13, 0))
```
ebbit1q, line 1243: these comments aren't doing much

cockatrice/src/pilezone.cpp
```diff
@@ -44,7 +44,8 @@ void PileZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*optio
 void PileZone::addCardImpl(CardItem *card, int x, int /*y*/)
 {
     connect(card, SIGNAL(sigPixmapUpdated()), this, SLOT(callUpdate()));
-    cards.insert(x, card);
+
+    cards.insert(x < 0 ? 0 : x, card);
```
ebbit1q, line 48: should we check upper bound as well?

cockatrice/src/player.cpp
```diff
@@ -3066,7 +3067,13 @@ void Player::actSetPT()
             const auto oldpt = parsePT(card->getPT());
             int ptIter = 0;
             for (const auto &item : ptList) {
-                if (item.type() == QVariant::Int) {
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+                if (item.typeId() == QMetaType::Type::Int)
+#else
+                if (item.type() == QVariant::Int)
+#endif
+                {
```
ebbit1q, line 3075: why does this brace get its own line

cockatrice/src/remoteclient.cpp
```diff
@@ -41,8 +41,13 @@ RemoteClient::RemoteClient(QObject *parent)
     socket->setSocketOption(QAbstractSocket::LowDelayOption, 1);
     connect(socket, SIGNAL(connected()), this, SLOT(slotConnected()));
     connect(socket, SIGNAL(readyRead()), this, SLOT(readData()));
+
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    connect(socket, &QTcpSocket::errorOccurred, this, &RemoteClient::slotSocketError);
```
ebbit1q, line 46: does this even do the same thing at all?

cockatrice/src/setsmodel.cpp
```diff
@@ -274,9 +276,19 @@ bool SetsDisplayModel::filterAcceptsRow(int sourceRow, const QModelIndex &source
     auto nameIndex = sourceModel()->index(sourceRow, SetsModel::LongNameCol, sourceParent);
     auto shortNameIndex = sourceModel()->index(sourceRow, SetsModel::ShortNameCol, sourceParent);
 
-    return (sourceModel()->data(typeIndex).toString().contains(filterRegExp()) ||
-            sourceModel()->data(nameIndex).toString().contains(filterRegExp()) ||
-            sourceModel()->data(shortNameIndex).toString().contains(filterRegExp()));
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 12, 0))
+    auto filter1 = filterRegularExpression();
```
ebbit1q, line 280: these are all the same object

.ci/docker.sh
ZeldaZach: Fixed

.ci/docker.sh
ZeldaZach: Reverted

.ci/release_template.md
ZeldaZach: Yep removed

CMakeLists.txt
```diff
@@ -103,13 +118,10 @@ elseif(WIN32)
     set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/rundir/${CMAKE_BUILD_TYPE})
 endif()
 
-# Treat warnings as errors (Debug builds only)
-option(WARNING_AS_ERROR "Treat warnings as errors in debug builds" ON)
```
ebbit1q, line 107: why disable this? this is useful to catch many issues early

cockatrice/src/remoteclient.cpp
```diff
@@ -41,8 +41,13 @@ RemoteClient::RemoteClient(QObject *parent)
...
+#else
     connect(socket, SIGNAL(error(QAbstractSocket::SocketError)), this,
```
ebbit1q, line 48: `QObject::connect: No such signal QProcess::error(QAbstractSocket::SocketError)` did this change?

.ci/UbuntuBionic/Dockerfile
```diff
@@ -21,4 +21,5 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         qttools5-dev \
         qttools5-dev-tools \
     && apt-get clean \
+    && git config --global --add safe.directory /src \
```
ebbit1q, line 24: I fixed this in https://github.com/Cockatrice/Cockatrice/pull/4610

servatrice/src/smtp/qxtsmtp.cpp
```diff
@@ -23,39 +23,44 @@
...
     qxt_d().state = QxtSmtpPrivate::Disconnected;
     qxt_d().nextID = 0;
     qxt_d().socket = new QSslSocket(this);
     QObject::connect(socket(), SIGNAL(encrypted()), this, SIGNAL(encrypted()));
-    //QObject::connect(socket(), SIGNAL(encrypted()), &qxt_d(), SLOT(ehlo()));
+    // QObject::connect(socket(), SIGNAL(encrypted()), &qxt_d(), SLOT(ehlo()));
     QObject::connect(socket(), SIGNAL(connected()), this, SIGNAL(connected()));
     QObject::connect(socket(), SIGNAL(disconnected()), this, SIGNAL(disconnected()));
-    QObject::connect(socket(), SIGNAL(error(QAbstractSocket::SocketError)), &qxt_d(), SLOT(socketError(QAbstractSocket::SocketError)));
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
```
ebbit1q, line 57: the formatting changes here make this hard to figure out, is this the only change in the entire file?
is it worth reformatting the entire file that is borrowed from the old libqxt anyway?

.ci/docker.sh
```diff
@@ -28,6 +28,10 @@ while [[ $# != 0 ]]; do
       GET=1
       shift
       ;;
+    '--interactive')
```
ebbit1q, line 31: I have put this change into its own pr as well https://github.com/Cockatrice/Cockatrice/pull/4610

.github/workflows/desktop-build.yml
```diff
@@ -296,74 +305,72 @@ jobs:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: recursive
 
-      - name: Restore Qt ${{env.QT_VERSION}} ${{matrix.arch}}-bit from cache
+      - name: Restore Qt ${{matrix.qt_version}} (${{matrix.arch}}-bit) from cache
         id: cache-qt
-        uses: actions/cache@v2
+        uses: actions/cache@v1
```
ebbit1q, line 335: is there a specific reason for downgrading the cache action?

.github/workflows/desktop-build.yml
```diff
@@ -296,74 +305,72 @@ jobs:
...
         with:
-          key: ${{runner.os}}-QtCache-${{env.QT_VERSION}}-${{matrix.arch}}
-          path: ${{runner.workspace}}/Qt
+          key: ${{runner.os}}-QtCache-${{matrix.qt_version}}-${{matrix.arch}}
+          path: '${{github.workspace}}/Qt'
```
ebbit1q, line 338: shouldn't we store this in an out of the way temporary directory like runner.temp?

.github/workflows/desktop-build.yml
```diff
@@ -296,74 +305,72 @@ jobs:
...
-        uses: lukka/run-vcpkg@v6
+          version: ${{matrix.qt_version}}
+          arch: win${{matrix.arch}}_${{matrix.qt_arch}}
+          modules: ${{matrix.qt_modules}}
+      - name: Run vcpkg
+        uses: lukka/run-vcpkg@v10.2
         with:
-          vcpkgArguments: '@${{github.workspace}}/vcpkg.txt'
-          vcpkgDirectory: ${{github.workspace}}/vcpkg
-          appendedCacheKey: ${{hashFiles('**/vcpkg.txt')}}
```
ebbit1q, line 347: does this mean we no longer cache the vcpkg packages? how much space do these use?

CMakeLists.txt
```diff
@@ -252,7 +257,7 @@ elseif(WIN32)
 
     # Configure file with custom definitions for NSIS.
     configure_file(
-        ${CMAKE_MODULE_PATH}/NSIS.definitions.nsh.in
+        cmake/NSIS.definitions.nsh.in
```
ebbit1q, line 260: why change this? this variable is set in this file and sets the absolute location

CMakeLists.txt
```diff
@@ -231,15 +228,23 @@ if(UNIX)
         # linux
         IF(CPACK_GENERATOR STREQUAL "RPM")
             set(CPACK_RPM_PACKAGE_LICENSE "GPLv2")
-            set(CPACK_RPM_PACKAGE_REQUIRES "protobuf, qt5-qttools, qt5-qtsvg, qt5-qtmultimedia")
+            IF(QT6_FOUND)
```
ebbit1q, line 231: what is it QT6_FOUND or Qt6_FOUND?

CMakeLists.txt
```diff
@@ -231,15 +228,23 @@ if(UNIX)
...
+                SET(CPACK_RPM_PACKAGE_REQUIRES "protobuf, qt6-qttools, qt6-qtsvg, qt6-qtmultimedia")
+            ELSEIF(Qt5_FOUND)
```
ebbit1q, line 233: if neither of these is found an error happens, doing the double check here seems weird

cmake/FindQtRuntime.cmake
```diff
@@ -0,0 +1,44 @@
...
+endif()
+list(REMOVE_DUPLICATES COCKATRICE_QT_COMPONENTS)
+
+if(NOT FORCE_USE_QT5)
+    find_package(Qt6 6.2.3 COMPONENTS ${COCKATRICE_QT_COMPONENTS} QUIET HINTS ${Qt6_DIR})
+endif()
+if(Qt6_FOUND)
+    set(COCKATRICE_QT_VERSION_NAME Qt6)
+else()
+    list(REMOVE_ITEM COCKATRICE_QT_COMPONENTS Core5Compat) # Core5Compat is Qt6 Only
```
ebbit1q, line 26: why remove the item here if you could be adding it instead, resulting in less complexity?

cockatrice/CMakeLists.txt
```diff
@@ -153,30 +153,43 @@ if(APPLE)
     set(cockatrice_SOURCES ${cockatrice_SOURCES} ${CMAKE_CURRENT_SOURCE_DIR}/resources/appicon.icns)
 ENDIF(APPLE)
 
-# Qt5
-find_package(Qt5 COMPONENTS Concurrent Multimedia Network PrintSupport Svg WebSockets Widgets REQUIRED)
-set(COCKATRICE_QT_MODULES Qt5::Concurrent Qt5::Multimedia Qt5::Network Qt5::PrintSupport Qt5::Svg Qt5::Widgets Qt5::WebSockets)
+LIST(APPEND COCKATRICE_QT_MODULES ${COCKATRICE_QT_VERSION_NAME}::Concurrent ${COCKATRICE_QT_VERSION_NAME}::Multimedia ${COCKATRICE_QT_VERSION_NAME}::Network ${COCKATRICE_QT_VERSION_NAME}::PrintSupport ${COCKATRICE_QT_VERSION_NAME}::Svg ${COCKATRICE_QT_VERSION_NAME}::Widgets ${COCKATRICE_QT_VERSION_NAME}::WebSockets)
```
ebbit1q, line 156: if you moved all these to the findqt file, why are they duplicated here?

cockatrice/CMakeLists.txt
```diff
@@ -153,30 +153,43 @@ if(APPLE)
...
+IF (Qt6_FOUND)
+    LIST(APPEND COCKATRICE_QT_MODULES ${COCKATRICE_QT_VERSION_NAME}::Core5Compat ${COCKATRICE_QT_VERSION_NAME}::Gui)
+ENDIF()
 
-# Qt5LinguistTools
-find_package(Qt5LinguistTools)
-if(Qt5LinguistTools_FOUND)
-    list(APPEND COCKATRICE_LIBS Qt5::LinguistTools)
+# Optional Qt LinguistTools
+IF(${COCKATRICE_QT_VERSION_NAME}LinguistTools_FOUND)
```
ebbit1q, line 162: it doesn't seem like you ever run find_package(qtxLinuguistTools before this is checked, neither here or in the findqt file, this will probably not work at all

ebbit1q: it seems like it would silently fail too, so we have no idea if it does

CMakeLists.txt
ebbit1q: I say we stick to a consistent casing either all uppercase or all lowercase

cockatrice/CMakeLists.txt
ebbit1q: I just checked in the application, the language selector is nonfunctional because of this

cockatrice/src/abstractclient.cpp
ebbit1q, line 55: would you mind adding this to the list above it?

cockatrice/src/carddbparser/cockatricexml4.cpp
```diff
@@ -52,12 +52,13 @@ void CockatriceXml4Parser::parseFile(QIODevice &device)
                     break;
                 }
 
-                if (xml.name() == "sets") {
+                if (xml.name().toString() == "sets") {
```
ebbit1q, line 55: this has the same issue as the xml3 had

cockatrice/src/chatview/chatview.cpp
```diff
@@ -321,7 +321,7 @@ void ChatView::checkTag(QTextCursor &cursor, QString &message)
 
 void ChatView::checkMention(QTextCursor &cursor, QString &message, const QString &userName, UserLevelFlags userLevel)
 {
-    const QRegExp notALetterOrNumber = QRegExp("[^a-zA-Z0-9]");
+    const auto notALetterOrNumber = QRegularExpression("[^a-zA-Z0-9]");
```
ebbit1q, line 324: should this be static?

cockatrice/src/deckview.cpp
```diff
@@ -10,7 +10,7 @@
 #include <QGraphicsSceneMouseEvent>
 #include <QMouseEvent>
 #include <algorithm>
-#include <math.h>
+#include <cmath>
```
ebbit1q, line 13: um, where are we using cmath instead of qt? that might have portability concerns

cockatrice/src/dlg_connect.cpp
```diff
@@ -162,8 +162,7 @@ DlgConnect::DlgConnect(QWidget *parent) : QDialog(parent)
 
     previousHostButton->setChecked(true);
 
-    connect(previousHosts, SIGNAL(currentIndexChanged(const QString &)), this,
-            SLOT(updateDisplayInfo(const QString &)));
+    connect(previousHosts, SIGNAL(currentTextChanged(const QString &)), this, SLOT(updateDisplayInfo(const QString &)));
```
ebbit1q, line 165: why change this? the signal didn't disappear

ebbit1q: also.. conflicts with my pr...

ebbit1q: ah, I see, it takes an int now, I already changed this in my pr

cockatrice/src/pilezone.cpp
```diff
@@ -44,7 +44,15 @@ void PileZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*optio
 void PileZone::addCardImpl(CardItem *card, int x, int /*y*/)
 {
     connect(card, SIGNAL(sigPixmapUpdated()), this, SLOT(callUpdate()));
-    cards.insert(x, card);
+
+    if (0 <= x) {
```
ebbit1q, line 48: to reiterate on my previous comment here, this is not where we should be doing this check, we should be doing it in CardZone::addCard before we call this function which is implemented in all other card zones, that way the issue will not happen in those classes as well, alternatively also add this in all those other zones

cockatrice/src/replay_timeline_widget.cpp
```diff
@@ -62,7 +62,11 @@ void ReplayTimelineWidget::paintEvent(QPaintEvent * /* event */)
 
 void ReplayTimelineWidget::mousePressEvent(QMouseEvent *event)
 {
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    int newTime = static_cast<int>((qint64)maxTime * (qint64)event->position().x() / width());
```
ebbit1q, line 66: all this casting is very pointless, this could just be a qint64

cockatrice/src/pilezone.cpp
ZeldaZach: That makes sense. I've redone it to do the check upstream

cockatrice/src/replay_timeline_widget.cpp
ZeldaZach: I'm not fully certain why it does this, but I'm not sure of a better chance as it might be a larger change required across the system

cockatrice/src/gamescene.cpp
```diff
@@ -13,7 +13,7 @@
 #include <QGraphicsSceneMouseEvent>
 #include <QGraphicsView>
 #include <QSet>
-#include <math.h>
+#include <QtMath>
```
ebbit1q, line 16: these includes don't seem to be used at all!

.github/workflows/desktop-build.yml
ebbit1q: that sounds like a caveat we should document with a comment, for posterity...

CMakeLists.txt
ebbit1q: it most definitely should not be a comma separated list, we set it like this `SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)` and we use it in other places as well

cockatrice/src/tab_supervisor.cpp
```diff
@@ -58,11 +58,19 @@ QSize CloseButton::sizeHint() const
     return QSize(width, height);
 }
 
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+void CloseButton::enterEvent(QEnterEvent *event)
```
ebbit1q, line 62: this could just be one line again

cockatrice/src/window_main.cpp
```diff
@@ -932,7 +932,8 @@ void MainWindow::createTrayIcon()
 void MainWindow::iconActivated(QSystemTrayIcon::ActivationReason reason)
 {
     if (reason == QSystemTrayIcon::DoubleClick) {
-        if (windowState() != Qt::WindowMinimized && windowState() != Qt::WindowMinimized + Qt::WindowMaximized)
+        if (windowState() != Qt::WindowMinimized &&
+            windowState() != ((int)Qt::WindowMinimized + (int)Qt::WindowMaximized))
```
ebbit1q, line 936: this makes no sense, the original writer of this line probably meant to use &

ZeldaZach: Since the Qt ENUMS are key'd in 2 powers, adding and bit-wise ANDing should return the same value. It's not best practice, sure, but it does work. https://doc.qt.io/qt-5/qt.html#WindowState-enum

ebbit1q: adding and bitwise and do not do the same thing here, you're confused with bitwise or

what I mean is windowstate() & Qt::WindowMinimized == 0 replacing both these lines

cockatrice/src/dlg_connect.cpp
ebbit1q: I want this pr merged first

cockatrice/src/window_main.cpp
ZeldaZach: Meant to say OR, but yeah. I'll take a look at this

cockatrice/src/dlg_connect.cpp
ZeldaZach: Gotcha, so I'll resolve this comment for now

cockatrice/src/remoteclient.cpp
ZeldaZach: error signal became errorOccurred in 5.15+

oracle/src/oraclewizard.cpp
```diff
@@ -454,8 +458,10 @@ void LoadSetsPage::readSetsFromByteArray(QByteArray data)
             zipDownloadFailed(tr("Xz extraction failed."));
             return;
         }
+        const auto &outBufferData = outBuffer->data();
 
-        future = QtConcurrent::run(wizard()->importer, &OracleImporter::readSetsFromByteArray, outBuffer->data());
+        future = QtConcurrent::run(
+            [this, &outBufferData] { return wizard()->importer->readSetsFromByteArray(outBufferData); });
```
ebbit1q, line 464: handing this by reference to a different thread doesn't look very safe at all

common/server.h
```diff
@@ -57,8 +57,8 @@ private slots:
 
 public:
     mutable QReadWriteLock clientsLock, roomsLock; // locking order: roomsLock before clientsLock
-    Server(QObject *parent = nullptr);
-    ~Server() = default;
+    explicit Server(QObject *parent = nullptr);
+    virtual ~Server(){};
```
ebbit1q, line 61:
```suggestion
    virtual ~Server() = default;
```

oracle/src/oraclewizard.cpp
ZeldaZach: The pointer wasn't safe and led to access errors, but as an object it seems fine from my testing

cockatrice/src/window_main.cpp
```diff
@@ -932,9 +932,9 @@ void MainWindow::createTrayIcon()
 void MainWindow::iconActivated(QSystemTrayIcon::ActivationReason reason)
 {
     if (reason == QSystemTrayIcon::DoubleClick) {
-        if (windowState() != Qt::WindowMinimized && windowState() != Qt::WindowMinimized + Qt::WindowMaximized)
+        if ((windowState() & (Qt::WindowMinimized & Qt::WindowMaximized)) == 0) {
```
ebbit1q, line 935: um, I'm not sure what you're doing here but doing & on these flags will always be 0

ebbit1q:
```suggestion
        if (windowState() & Qt::WindowMinimized == 0) {
```

ZeldaZach: Ahh I meant to do `|`

ZeldaZach:
```suggestion
        if ((windowState() & (Qt::WindowMinimized | Qt::WindowMaximized)) == 0) {
```

ZeldaZach:
```suggestion
        if ((windowState() & Qt::WindowMinimized) == 0) {
```

Paren required for bitwise

oracle/src/oraclewizard.cpp
ebbit1q: does this duplicate everything then?

cockatrice/CMakeLists.txt
ZeldaZach: Resolved this

servatrice/src/smtp/qxtsmtp.cpp
ZeldaZach: The file auto-reformatted with clang. I ran through it manually and I only made one change to the file (What oyu see here). Sorry for the overlaps.

CMakeLists.txt
ZeldaZach: I wonder why I was having issues with this on Windows. I'll switch it back and see if the build still works.

oracle/src/oraclewizard.cpp
ZeldaZach: Not sure what you mean

servatrice/src/serversocketinterface.cpp
```diff
@@ -1905,11 +1902,11 @@ void WebsocketServerSocketInterface::initConnection(void *_socket)
     connect(socket, SIGNAL(disconnected()), this, SLOT(catchSocketDisconnected()));
 
     // Add this object to the server's list of connections before it can receive socket events.
-    // Otherwise, in case a of a socket error, it could be removed from the list before it is added.
+    // Otherwise, in case of a socket error, it could be removed from the list before it is added.
     server->addClient(this);
 
     logger->logMessage(
-        QString("Incoming websocket connection: %1 (%2)").arg(address.toString()).arg(socket->peerAddress().toString()),
+        QString("Incoming websocket connection: %1 (%2)").arg(address.toString(), socket->peerAddress().toString()),
```
ebbit1q, line 1909: these are not the same
```suggestion
        QString("Incoming websocket connection: %1 (%2)").arg(address.toString()).arg(socket->peerAddress().toString()),
```

cockatrice/src/window_main.cpp
```diff
@@ -932,9 +932,9 @@ void MainWindow::createTrayIcon()
 void MainWindow::iconActivated(QSystemTrayIcon::ActivationReason reason)
 {
     if (reason == QSystemTrayIcon::DoubleClick) {
-        if (windowState() != Qt::WindowMinimized && windowState() != Qt::WindowMinimized + Qt::WindowMaximized)
+        if ((windowState() & Qt::WindowMinimized) == 0) {
```
ebbit1q, line 935: need to check if this functions the same on multiple oses

oracle/src/oraclewizard.cpp
ebbit1q: it's oracle, we shouldn't put too much effort in it imo

ebbit1q (approved):
:pray:

.github/workflows/desktop-build.yml
```diff
@@ -208,28 +208,28 @@ jobs:
 
           - target: 10.14_Mojave
             os: macos-10.15 # runs on Catalina
-            xcode: 10.3 # allows compatibility with macOS 10.14
+            xcode: 11.3 # allows compatibility with macOS 10.14
```
tooomm, line 211: This will no longer target macOS 10.14, but 10.15

Unfortunately, Xcode 10.3 is the most recent (and only) option we have to keep support for 10.14

.github/workflows/desktop-build.yml
```diff
@@ -208,28 +208,28 @@ jobs:
...
             type: Release
             do_tests: 0
             make_package: true
 
           - target: 10.15_Catalina
             os: macos-10.15
-            xcode: 12.1
+            xcode: 12.4
```
tooomm, line 218: This will no longer target 10.15, but 11.1

You can bump to 12.1.1 though.

.github/workflows/desktop-build.yml
```diff
@@ -208,28 +208,28 @@ jobs:
...
             type: Release
             do_tests: 0
             make_package: true
 
           - target: 11_Big_Sur
             os: macos-11
-            xcode: 12.5.1
+            xcode: 13.2
```
tooomm, line 225: This will no longer target 11, but 12.1

Xcode 13.0 is the max there.

cmake/FindQtRuntime.cmake
```diff
@@ -0,0 +1,106 @@
...
+        ${_SERVATRICE_NEEDED}
+        ${_COCKATRICE_NEEDED}
+        ${_ORACLE_NEEDED}
+        ${_DBCONVERTER_NEEDED}
+        ${_TEST_NEEDED})
+list(REMOVE_DUPLICATES REQUIRED_QT_COMPONENTS)
+
+if(NOT FORCE_USE_QT5)
+    # Core5Compat is Qt6 Only, Linguist is now a component in Qt6 instead of an external package
+    find_package(Qt6 6.2.3
```
tooomm, line 40: Qt 6.3 ?

tooomm (changes requested):

cmake/FindQtRuntime.cmake
ZeldaZach: Unix distros vends 6.2.3/6.2.4 natively

.github/workflows/desktop-build.yml
ZeldaZach: I found the last supported version at https://developer.apple.com/support/xcode/ which led me to changing these values

tooomm: GitHub only lists the supported SDK's for their runners without the minimum OS (which are actually lower it seems):
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#installed-sdks
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#installed-sdks

Looks like those upgrades are fine then. Double checked again.
Thanks for sharing the link!! 👍

ZeldaZach:
I think I'm happy with the current state. Will move towards testing across all OSes

cockatrice/src/window_main.cpp
ZeldaZach: Feel free to check, but I think it should be fine

.github/workflows/desktop-build.yml
```diff
@@ -290,80 +309,83 @@ jobs:
           asset_name: ${{steps.package.outputs.name}}
           asset_content_type: application/octet-stream
 
+      - uses: actions/checkout@v2
+      - name: Setup tmate session
```
ebbit1q, line 313: debug stuff

.ci/compile.sh
```diff
@@ -145,12 +134,10 @@ if [[ $(uname) == "Darwin" ]]; then
...
-if [[ $USE_CCACHE ]]; then
-  flags+=("-DUSE_CCACHE=1")
-  if [[ $CCACHE_SIZE ]]; then
-    # note, this setting persists after running the script
-    ccache --max-size "$CCACHE_SIZE"
-  fi
+# Add cmake --build flags
+buildflags=(--config "$BUILDTYPE")
+if [[ $PARALLEL_COUNT ]]; then
+  buildflags+=(--parallel "$PARALLEL_COUNT")
```
ZeldaZach, line 140: --parallel was added in CMake 3.12, but our minimum is 3.10. This causes Bionic to fail

tooomm:
If found this issue regarding the parallel flag and cmake on Windows: [no /mp option set](https://stackoverflow.com/questions/10688549/how-do-i-configure-portable-parallel-builds-in-cmake/70102570#70102570)

.ci/compile.sh
ebbit1q: right, I knew this and forgot

dbconverter/CMakeLists.txt
```diff
@@ -87,10 +87,12 @@ if(WIN32)
 
     install(DIRECTORY "${CMAKE_BINARY_DIR}/${PROJECT_NAME}/${CMAKE_BUILD_TYPE}/" DESTINATION ./ FILES_MATCHING PATTERN "*.dll")
 
-    # qt5 plugins: platforms, sqldrivers/mysql
-    install(DIRECTORY "${QT_PLUGINS_DIR}/" DESTINATION ${plugin_dest_dir} COMPONENT Runtime
-        FILES_MATCHING REGEX "(platforms/.*)\\.dll"
-        REGEX ".*d\\.dll" EXCLUDE)
+    # Qt plugins: platforms
+    install(DIRECTORY "${QT_PLUGINS_DIR}/" DESTINATION ${plugin_dest_dir} COMPONENT Runtime FILES_MATCHING
+            PATTERN "platforms/qdirect2d.dll"
```
ebbit1q, line 92: does dbconverter really need this plugin?

.ci/compile.sh
ebbit1q: I've added a workaround

cockatrice/src/window_main.cpp
ebbit1q: it's not for me, but it never was outside of windows afaik, so this is out of scope

cmake/FindWin32SslRuntime.cmake
```diff
@@ -1,66 +1,47 @@
...
+            "C:/OpenSSL-Win64"
+            "C:/Tools/vcpkg/installed/x64-windows/bin"
+            "${_programfiles}/OpenSSL-Win64"
+            )
     unset(_programfiles)
-  else( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-    # target win32
-    set(_OPENSSL_ROOT_HINTS
-      ${OPENSSL_ROOT_DIR}
-      "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\OpenSSL (32-bit)_is1;Inno Setup: App Path]"
```
ebbit1q, line 27: what did this registry looking thing do before?

ZeldaZach: It was apart of hinting, which I removed for pathing only

cmake/FindWin32SslRuntime.cmake
```diff
@@ -1,66 +1,47 @@
...
+    set(_OPENSSL_ROOT_HINTS ${OPENSSL_ROOT_DIR} ENV OPENSSL_ROOT_DIR)
     file(TO_CMAKE_PATH "$ENV{PROGRAMFILES}" _programfiles)
     set(_OPENSSL_ROOT_PATHS
-      "C:/Tools/vcpkg/installed/x86-windows/bin"
-      "${_programfiles}/OpenSSL"
-      "${_programfiles}/OpenSSL-Win32"
-      "C:/OpenSSL/"
-      "C:/OpenSSL-Win32/"
-      )
+            "$ENV{VCPKG_PACKAGES_DIR}/x86-windows/bin"
```
ebbit1q, line 21: how does this behave when the env var is empty (env vars default to empty)

ZeldaZach: Would look at the root dir, which isn't a problem as it'll just be another dead path

cmake/FindWin32SslRuntime.cmake
```diff
@@ -1,66 +1,47 @@
...
+            "C:/OpenSSL"
+            "C:/Tools/vcpkg/installed/x86-windows/bin"
+            "${_programfiles}/OpenSSL"
+            "${_programfiles}/OpenSSL-Win32"
+            )
     unset(_programfiles)
-  endif( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-
+  else()
+    message(WARNING "Unknown OS_PATH: $ENV{OS_PATH}, Can't find OpenSSL correctly")
```
ebbit1q, line 31: what's OS_PATH?

ZeldaZach: Forgot to change it, fixing

cmake/FindWin32SslRuntime.cmake
```diff
@@ -1,66 +1,47 @@
...
+  endif()
 else ()
-  set(_OPENSSL_ROOT_HINTS
-    ${OPENSSL_ROOT_DIR}
-    ENV OPENSSL_ROOT_DIR
-    )
-endif ()
-
-set(_OPENSSL_ROOT_HINTS_AND_PATHS
```
ebbit1q, line 48: are you sure this part was unnecessary? you do seem to be setting the _ROOT_HINTS earlier

ZeldaZach: I was also removing them from setting. HINTS and PATHS are both optional, I'm only using Paths rn for it, which is fine

cmake/FindWin32SslRuntime.cmake
```diff
@@ -1,66 +1,47 @@
...
-else( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-  # target win32
-  FIND_FILE(WIN32SSLRUNTIME_LIBEAY NAMES libcrypto-1_1.dll libcrypto.dll libeay32.dll ${_OPENSSL_ROOT_HINTS_AND_PATHS})
-  FIND_FILE(WIN32SSLRUNTIME_SSLEAY NAMES libssl-1_1.dll libssl.dll ssleay32.dll ${_OPENSSL_ROOT_HINTS_AND_PATHS})
-endif( CMAKE_SIZEOF_VOID_P EQUAL 8 )
+  set(_OPENSSL_ROOT_HINTS ${OPENSSL_ROOT_DIR} ENV OPENSSL_ROOT_DIR)
+endif()
+
+message(STATUS "Looking for OpenSSL @ ${CMAKE_GENERATOR_PLATFORM} in ${_OPENSSL_ROOT_PATHS}")
+if("${CMAKE_GENERATOR_PLATFORM}" STREQUAL "x64")
```
ebbit1q, line 38: this part seems to be outside of the WIN32 check but doesn't seem platform agnostic at all, isn't this entire file only used on windows anyway?

ZeldaZach: I've changed it to ensure we're in a win set, but yeah it's only ever included on windows

PR #4610, opened 2022-04-18
fix the uid and gid of the user in the docker container

ebbit1q:
this fixes this error:
unsafe repository ('/src' is owned by someone else)
this caused the hash to go missing in the version number


ebbit1q:
this also adds a bunch of comments explaining these bash scripts and the --dir option to compile.sh and --interactive to docker.sh which are very helpful when debugging these locally

PR #4615, opened 2022-05-07
CI: Explore improved parallel builds on Win

tooomm:
See https://github.com/Cockatrice/Cockatrice/pull/4615#issuecomment-1925790981 for a wrap up of the most recent information as this PR was evolving over a long time.

#4998 is the follow up PR with a clean history


ebbit1q:
is there a specific reason to do it this way? did the --parallel flag not work?

tooomm:
> is there a specific reason to do it this way? did the --parallel flag not work?

As I tried to point out in https://github.com/Cockatrice/Cockatrice/pull/4607#issuecomment-1106344852, --parallel is not sufficient on Windows.
See also the link to an open cmake issue I left in the code: https://gitlab.kitware.com/cmake/cmake/-/issues/20564

We need to tweak the compile command further to do parallelization right in Windows.

ebbit1q:
so, is having job parallelization sufficient? did you try `--parallel 2 -- -p:UseMultiToolTask=true -p:EnforceProcessCountAcrossBuilds=true` in combination?

tooomm:
> so, is having job parallelization sufficient?

It's really confusing and some of the results are not expected either.
There is the official issue on cmake, but people seems to find different "solutions" for it. On the other hand there is the observation and test in https://github.com/scipp/scipp/issues/2078 which finds that not running with parallel options at all is considerably faster...

>did you try `--parallel 2 -- -p:UseMultiToolTask=true -p:EnforceProcessCountAcrossBuilds=true` in combination?

[`fb0e7cd` (#4615)](https://github.com/Cockatrice/Cockatrice/pull/4615/commits/fb0e7cdffd5b343eaa298f7ee93055376417145a) should be that: https://github.com/Cockatrice/Cockatrice/actions/runs/2287700911

<br>

Here is a full list of what I tried:
|  Command in compile.sh script  |  Commit  |  Time (64bit)  |  Notes  |
|---------------------------------|-----------|---------------|---------|
|  cmake --build . "${buildflags[@]}"  |  [`b02adcc`](https://github.com/Cockatrice/Cockatrice/commit/b02adccf8799944e65bde0a6a0c9bf7b1a8589ee)  |  [16m 26s](https://github.com/Cockatrice/Cockatrice/runs/6329554458?check_suite_focus=true)  |  master after Qt6 PR  |
||||
|  cmake --build . "${buildflags[@]}" -- -p:CL_MPcount="$PARALLEL_COUNT"  |  [`039c48d`](https://github.com/Cockatrice/Cockatrice/pull/4615/commits/039c48d4df212ecebe3ff0a6b03ff1f7bddd05be)  |  [16m 47s](https://github.com/Cockatrice/Cockatrice/runs/6337343049?check_suite_focus=true)  |   |
|  cmake --build . "${buildflags[@]}" -- -p:CL_MPcount=4  |  [`a073d13`](https://github.com/Cockatrice/Cockatrice/pull/4615/commits/a073d13ab1840eb600faad3a7bc6c4f18c59532a)  |  [16m 28s](https://github.com/Cockatrice/Cockatrice/runs/6339920245?check_suite_focus=true)  |    |
|  cmake --build . "${buildflags[@]}" -- -p:UseMultiToolTask=true -p:CL_MPcount=4  |  [`896d7e2`](https://github.com/Cockatrice/Cockatrice/pull/4615/commits/896d7e2c8c9e39c1a3b5cd5c84a3170a87568b04)  |  [14m 02s](https://github.com/Cockatrice/Cockatrice/runs/6340258826?check_suite_focus=true)  |   |
|  cmake --build . "${buildflags[@]}" -- -p:UseMultiToolTask=true -p:EnforceProcessCountAcrossBuilds=true  |  [`fb0e7cd`](https://github.com/Cockatrice/Cockatrice/pull/4615/commits/fb0e7cdffd5b343eaa298f7ee93055376417145a)   |  [18m 42s](https://github.com/Cockatrice/Cockatrice/runs/6337110222?check_suite_focus=true)  |   |
|  cmake --build . "${buildflags[@]}"  |  [`26aa92b`](https://github.com/Cockatrice/Cockatrice/pull/4615/commits/26aa92b2ee0530e94ee8cabd355968f3eff37381)  |  [19m 39s](https://github.com/Cockatrice/Cockatrice/runs/6340552401?check_suite_focus=true)  |  tried to not pass any parallel value to buildflags with `:`  |

But of course it could be a misimplementation of one of the options on my side as well!

Any other combinations or commands worth trying?

ebbit1q:
so you never tried without --parallel while setting -p:CL_MPcount= ?

tooomm:
I added two more variants to the test table above.

ebbit1q:
the latest commit looks substantially faster than before the machine had double cores, yet it seems like we don't have any windows builds completing successfully with the new core count so far at all making comparing this result difficult.

tooomm:
Ok, so what I got now...
- ~25% improved run time for windows builds (9/11 min)
- they succeed again after failing mysteriously with the recent CPU count changes

Since my initial try at parallelism a while ago, some things changed:
- [CMake used to pass an option when --parallel was used that actually decreased parallelism on windows](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7926), with v3.26.0 this changed and `/p:CL_MPCount=1` (to limit file-parallelism) is no longer set automatically. `/M` (for project-parallelism) is still set as I understand it. That should not help us too much though?
The GitHub runner uses CMake 3.27.9 by now, too.
- Since v16.3 there is a new [MTT](https://learn.microsoft.com/en-us/visualstudio/msbuild/multitooltask-task?view=vs-2019) option in Visual Studio 2019 to utilize parallel builds: 
     - https://devblogs.microsoft.com/cppblog/improved-parallelism-in-msbuild
     - Also interesting and helpful read with further suggestions: https://devblogs.microsoft.com/cppblog/cpp-build-throughput-investigation-and-tune-up

  The Windows 2019 runner is updated in the meantime and uses Visual Studio 2019 [16.11](https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md#visual-studio-enterprise-2019) currently.

<br>

I've now tried several options and combinations of [/M](https://learn.microsoft.com/en-us/visualstudio/msbuild/building-multiple-projects-in-parallel-with-msbuild?view=vs-2019), [/MP](https://learn.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=msvc-160) and MTT again. There is also an [MSBuild Resource Manager option](https://devblogs.microsoft.com/cppblog/cpp-build-throughput-investigation-and-tune-up/#msbuild-resource-manager) since 16.11, which I quickly gave a shot.

In the end it looks like going only with the new `UseMultiToolTask=true` works best.
Setting `-p:EnforceProcessCountAcrossBuilds=true` additionally had a small negative effect in the GitHub environment, so I removed it again.

Other notes:
- It seems often recommended to use the new MTT parameter, see [here](https://gitlab.kitware.com/cmake/cmake/-/issues/20564#note_731173) for example. Unless there are many small and simple C++ files involved, as stated by MS [here](https://devblogs.microsoft.com/cppblog/cpp-build-throughput-investigation-and-tune-up/#msbuild-resource-manager#multitooltask-mtt) (improvement to that available in VS2022).

- Though, if I understand the [guideline MS is giving](https://learn.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=msvc-160#the-msbuild-tool) correctly, in our case `/MP` would be even better (no mention of MTT though)
  >The guideline for deciding whether to use MSBuild or /MP technology is as follows:
  >
  >  - If there are **many projects with few files in each project**, use the MSBuild tool with the /maxcpucount option.
  >  - If there are **few projects with many files in each project**, use the /MP option.
  >  - If the **number of projects and files per project is balanced**, use both MSBuild and /MP. Initially set the /maxcpucount option to the number of projects to build and the /MP option to the number of processors on your computer. Measure performance and then adjust your settings to yield the best results. Repeat that cycle until you're satisfied with the total build time.

- **I could not make the `/MP` switch work correctly, maybe I did it wrong? I got also lost with `/p:CL_MPCount`.
Feel free to have a look and let me know the right place to enable this!**

<br><br>

>it seems like we don't have any windows builds completing successfully with the new core count so far at all making comparing this result difficult.

I still don't get why moving from `--parallel 2` to `--parallel 4` now magically fails mostly all windows builds...
Windows 7 was passing in the PR adding the parallel 4 option in [13 min](https://github.com/Cockatrice/Cockatrice/actions/runs/7574897745/job/20912373125?pr=4988) though.
Without any optimization on parallelism, builds run 15/17min on Windows.

Please check and let me know if you spot something that I missed in my research or have additional ideas. It gave me a hard time trying to get into it and I was not sure how to properly pass options and parameters here.

tooomm:
Not yet. Will do that in a separate PR after this one. Not sure what other implications might come with it. 

Don't want to postpone this further by running into new issues and rather fix (and speed up) the win builds on master. 👍

PR #4617, opened 2022-05-08
Prevent bad access potential for removals by checking bounds

### LINKED ISSUE #4616: Alt+Right Click and Drag Causes a Crash
**System Information:**
Client Version: 2.8.1-beta.7 (2022-05-06)
Build Architecture: 64-bit
Qt Version: 6.3.0
System Locale: en_US
Install Mode: Standard

Whenever I start a game, then hold Alt+Right Click and Drag or Shift+Right Click and Drag to create a blue arrow or a green arrow more than once, the game immediately crashes on me. 

It doesn't seem to have the same problem when using Ctrl+Right Click and Drag, with which I can make many different yellow arrows. Perhaps this can help narrow down what's causing this crash.

**Steps to reproduce:**
 - Put any permanents on both players' battlefields.
 - Hold down either Alt+ or Shift+Right Click and Drag multiple times to try to create several arrows between permanents on opposite battlefields.
 - The game crashes.

ZeldaZach:
Fix #4616

The crash came about because the index was -1 since the card object didn't exist. This PR addresses that issue (by ignoring -1 attempts) and safe handles all access points within the program that can lead to a similar situation.

### DIFF
--- a/common/server_cardzone.cpp
+++ b/common/server_cardzone.cpp
@@ -135,9 +135,7 @@ int Server_CardZone::removeCard(Server_Card *card, bool &wasLookedAt)
     if (wasLookedAt && cardsBeingLookedAt > 0) {
         cardsBeingLookedAt -= 1;
     }
-    if (index != -1) {
-        cards.removeAt(index);
-    }
+    cards.removeAt(index);
     if (has_coords) {
         removeCardFromCoordMap(card, card->getX(), card->getY());
     }
--- a/common/server_cardzone.cpp
+++ b/common/server_cardzone.cpp
@@ -135,7 +135,9 @@ int Server_CardZone::removeCard(Server_Card *card, bool &wasLookedAt)
     if (wasLookedAt && cardsBeingLookedAt > 0) {
         cardsBeingLookedAt -= 1;
     }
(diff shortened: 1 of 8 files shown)


common/server_cardzone.cpp
```diff
@@ -135,7 +135,9 @@ int Server_CardZone::removeCard(Server_Card *card, bool &wasLookedAt)
     if (wasLookedAt && cardsBeingLookedAt > 0) {
         cardsBeingLookedAt -= 1;
     }
-    cards.removeAt(index);
+    if (index != -1) {
```
ebbit1q, line 138: why not start this function with if index == -1 return -1?

ZeldaZach: It's possible the card might have coords, then we'd fail to remove it

ebbit1q: if the card has coords but is not present in this zone we have a different problem, actually just don't do this check at all, if this happens something went very wrong and we should stop it before it can do this

ZeldaZach: Hmm, you make a good point. I'll change it back

ebbit1q (approved):

PR #4625, opened 2022-05-17
Allow Copy from Cardlist in Deck Editor

### LINKED ISSUE #4041: Deck Editor: Allow Copy Card Name from card Database
When highlighting a card in my deck, I can Ctrl-C to copy the card name to the clipboard.  Very helpful!

I would also like to be able to Ctrl-C to copy the card name to the clipboard when highlighting a card name in the card database.  That would save me from having to add it to a deck, then copy from the deck.

Impyrical:
# Related Ticket(s)
Adds the functionality requested by Fix #4041

## Brief Description

Previously, you could ctrl-c to grab the name of a card from the deck list, but the same functionality did not exist for the card database. This adds that functionality. Now, pressing ctrl-c in the card database will copy the contents of the currently selected cell to the clipboard.

## Notes

I'm unsure if this is the best way to add this functionality as it exists in the deck list as a default feature of QList. Please suggest a cleaner way of doing this if it exists; I'm by no means an expert in Qt.

### DIFF
--- a/cockatrice/src/keysignals.cpp
+++ b/cockatrice/src/keysignals.cpp
@@ -61,6 +61,11 @@ bool KeySignals::eventFilter(QObject * /*object*/, QEvent *event)
                 emit onShiftS();
 
             break;
+        case Qt::Key_C:
+            if (kevent->modifiers() & Qt::ControlModifier)
+                emit onCtrlC();
+
+            break;
         default:
             return false;
     }
--- a/cockatrice/src/keysignals.h
+++ b/cockatrice/src/keysignals.h
@@ -20,6 +20,7 @@ signals:
     void onCtrlAltLBracket();
     void onCtrlAltRBracket();
     void onShiftS();
+    void onCtrlC();
 
 protected:
     virtual bool eventFilter(QObject *, QEvent *event);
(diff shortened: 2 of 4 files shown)


ZeldaZach (approved):
Neat, clean, and to the point. Nice work! Thanks for making a contribution to our project :)

ZeldaZach:
I did some looking to see if there was another way to do this, but it seems the QTreeView is limited in power. An alternative could be converting to a QTableView, but that seems like a lot of work to accomplish what we've done here.

The only thing that could be a tad confusing is you're only copying the exact column value that the user selected before copy/pasting, so if they picked Set or Card Type, it would be that value copied rather than the card name. Not opposed to this, just something I felt needed to be pointed out.

Thanks for your work, and we hope to see more awesome work from ya soon :)

Discord: https://cockatrice.us/discord

tooomm:
Works for me.
Indeed a bit confusing that it copies the mana cost or type of a card when I happen to click on that column when selecting a particular card/line. It behaves a bit like excel, rather than selecting a line it selects a cell, even though all line is highlighted.

Fixed the title. It worked in the decklist already, this adds copy to the card list in the deck editor.

PR #4627, opened 2022-05-19
rework formatting with cmake-format

### LINKED ISSUE #4621: cmakify mangles comments
our cmake files have comments that depend on their current formatting for clarity, the cmake-format script instead merges their formatting to become a single line, we should either make the comments not sensitive to autoformatting or change the settings of cmake-format.

It's important we do this before other changes to the cmake files as we will have to revert the cmake files to a previous version in order to restore their formatting.

ebbit1q:
## Related Ticket(s)
- Fixes #4621

the cmake-format behaved weird

- clangify.sh is now format.sh 
- format.sh --cmake additionally performs cmake-format on all edited cmake files
- cmakify.sh is has been removed
- the ci will format cmake files
- the ci has been updated to 22.04 beta in order to easily install cmake-format
- comments in cmake files are no longer formatted
- dbconverter has been added to the locations to format
- all files have been formatted with the new script, including some that have been missed previously


.ci/lint_cpp.sh
```diff
@@ -10,10 +10,10 @@ if ! git merge-base origin/master HEAD; then
   git fetch --unshallow
 fi
 
-# Check formatting using clangify
+# Check formatting using format.sh
 echo "Checking your code using clang-format..."
```
tooomm, line 14:
```suggestion
echo ""
echo "Checking your code using clang-format and cmake-format..."
```

.github/workflows/desktop-lint.yml
```diff
@@ -9,20 +9,20 @@ on:
...
-    - name: Install clang-format
+    - name: Install dependencies
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install -y --no-install-recommends clang-format
+        sudo apt-get install -y --no-install-recommends clang-format cmake-format
 
-    - name: Run clangify
+    - name: Format code
```
tooomm, line 26:
```suggestion
    - name: Check code styles
```

.github/workflows/desktop-lint.yml
```diff
@@ -9,20 +9,20 @@ on:
 
 jobs:
   clang-format:
```
tooomm, line 11:
```suggestion
  Check Formatting:
```

tooomm (commented):
Left some rewording suggestions.

Most comments use `# comment` style, few are changed to `#comment`.

New contributors might not be aware that they need to pass `--cmake` to the format script, when only seeing
`Please correct it or run the "format.sh" script`. The CONTRIBUTING file does also not mention this.

ebbit1q:
that's because we don't expect new contributors to edit the cmake files, it's fine as is, no need to make things complicated and the script has its own help function

tooomm:
Wouldn't it be useful to print the used version(s) for all runs?
It looks like those are only displayed on runs where a format diff is found.

ebbit1q:
which is fine

ZeldaZach (approved):

PR #4632, opened 2022-05-27
replace trayicon activation with menu actions

ebbit1q:
## Related Ticket(s)
- #4607

tray icon inconsistencies:
on mac the doubleclick is never triggered because we have a menu set: https://doc.qt.io/qt-5/qsystemtrayicon.html#ActivationReason-enum
on linux it's all over the place, some don't trigger anything at all.

this came to my attention in the qt6 conversion:
https://github.com/Cockatrice/Cockatrice/pull/4607#discussion_r864909937

- replace the activation of the tray icon with a more complete menu
- replace the effect of the effect of the show action with something a bit more sensible:
  - old behavior: if window is minimized: open and focus window, else minimize it (even if not in focus)
  - new behavior: if window is focused: minimize it, else open and focus window

## Screenshots
image

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -62,6 +62,7 @@
 #include <QSystemTrayIcon>
 #include <QThread>
 #include <QTimer>
+#include <QWindow>
 #include <QtConcurrent>
 #include <QtNetwork>
 
@@ -658,6 +659,9 @@ void MainWindow::retranslateUi()
     aUpdate->setText(tr("Check for Client Updates"));
     aCheckCardUpdates->setText(tr("Check for Card Updates..."));
     aViewLog->setText(tr("View &Debug Log"));
+
+    aShow->setText(tr("Show/Hide"));
+
     tabSupervisor->retranslateUi();
 }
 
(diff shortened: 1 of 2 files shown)


tooomm:
On Windows, clicking the tray icon does nothing. Same for double clicking.
`Show/Hide` brings Cockatrice to the front when minimized, but not in full screen and at changing locations which is weird...
Clicking `Show/Hide` does bring the app in focus when not minimized but covered. When in focus, it does not hide on activation as the label suggests.

Do we need the tray icon at all in the first place? 
It has not really too much benefit for the type of app Cockatrice is, and it got added back then by ctrlaltca to allow to have notifications on new dm's and game start etc. Back then it was not possible to show them otherwise on Windows. On Windows we use the native notification system by now. I guess there is no real added value on macOS and Linux either?

ebbit1q:
I can easily remove the tray icon, do you want me to create a pr for this?

tooomm:
> I can easily remove the tray icon, do you want me to create a pr for this?

As long as you want to support Windows 7, we need to keep it for the notifications to work at all there I guess. Looking at other recent PR's, it looks like the project wants to keep supporting it still...
I might be totally wrong here, but I do not think it's overly useful and expensive to maintain. It should also be EOL for a while besides paid enterprise support perhaps.

Maybe it's time to finally release a new version, get some fixes in a follow up revision some time later and start removing support for that old systems slowly. The next release following then won't happen in another year probably. 😉

PR #4638, opened 2022-06-05
Fixed reference to Qt6_FOUND function

dawidskorzewski:
Fixed reference to Qt6_FOUND function in main CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,7 +324,7 @@ if(TEST)
   add_subdirectory(tests)
 endif()
 
-if(Qt6Found AND Qt6_VERSION_MINOR GREATER_EQUAL 3)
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
Curious, translations actually worked while this was broken. Thanks for noticing anyway.

dawidskorzewski:
BTW this whole portion of code could probably be removed, via documentation (https://doc.qt.io/qt-6/qt-finalize-project.html), qt6_finalize_project() runs automatically on projects using CMake 3.19+

dawidskorzewski:
On the other hand minimum Cmake version for this project is 3.10, (and QT6 requires 3.16) so maybe it should be left to be on the safe side

PR #4642, opened 2022-06-15
fix volume slider on qt6

ebbit1q:
sound never worked on qt6, the volume setting goes from 0 to 1 instead of 0 to 100

### DIFF
--- a/cockatrice/src/soundengine.cpp
+++ b/cockatrice/src/soundengine.cpp
@@ -59,11 +59,12 @@ void SoundEngine::playSound(QString fileName)
     }
 
     player->stop();
+    int volumeSliderValue = SettingsCache::instance().getMasterVolume();
 #if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
-    player->audioOutput()->setVolume(SettingsCache::instance().getMasterVolume());
+    player->audioOutput()->setVolume(qreal(volumeSliderValue) / 100);
     player->setSource(QUrl::fromLocalFile(audioData[fileName]));
 #else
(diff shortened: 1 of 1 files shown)


ebbit1q:
while this does fix the volume slider there seems to be a different issue on my system, the qt6 version only plays sound after I compile it until I run any other version of cockatrice on the same system, and there is always some audio clipping not present on the qt5 build

tooomm:
Sound seems to work with the Win10 64bit file.

tooomm:
Hmm, after going to the beta.9 to verify the issue... going back to

```
Client Version: 2.8.1-custom(fa859ec) (2022-06-15)
Build Architecture: 64-bit
Qt Version: 5.15.2
System Locale: en_GB
Install Mode: Standard
```

gives me no sound now. 🤔

tooomm (approved):
I tested and confirm this again for Windows with Qt6. 👍 

On beta.10 adjusting the volume has no effect, e.g. volume slider at 2 is as loud as at 100.
Only enabling/disabling works via the check box. Or putting the slider to 0 (disabled) or non 0 (enabled).
>Client Version: 2.8.1-beta.10 (2022-09-01)
Build Architecture: 64-bit
Qt Version: 6.3.1
System Locale: en_GB
Install Mode: Standard

With the fix from this PR in `2.8.1-custom(0221b0e)`, the volume slider has expected functionality.


<br>

>there seems to be a different issue on my system, the qt6 version only plays sound after I compile it until I run any other version of cockatrice on the same system, and there is always some audio clipping not present on the qt5 build

I don't think I've those issues on my machine as far as I can tell. 🤔

PR #4648, opened 2022-07-04
add move cards from top of library until dialog

ebbit1q:
a bit of a hack, the client will use the play top card action and then
compare it with the propmpted expression, as if you were cascading
normally but really fast

the new keybind for this is ctrl shift y

I have ratelimited the action to 10 cards a second

## Related Ticket(s)
- the play top card action was added in #3827
- this is more in line with the request in #3824

just like ctrl y all cards are placed on the stack, you might want to move things off the stack first

cascade with this new action would use the following expression `-t:land mv<x`
I made the action more general so it could be useful for other cards as well:
many cards find creatures like this: `t:creature`
abundance is easy, either choose `t:land` or `-t:land`
cards like demonic consultation or divining witch will simply ask for the card name, note that you'll want to use quotes in order to match the exact name, eg `"Thassa's Oracle"`
heirloom blade you can simply give a list, note the use of or: `t:creature ( t:human or t:warrior )`
treasure hunt decks rejoice


mainman879:
To fix: Prevent people from spamming the command to speed up the search and possibly lag the server.

Possible names for the command:
Put onto stack until
Search cards until
Search for card until
Search for specific card
Cascade (its not only for cascade but cascade is a very recognizable term)

Suggestion: Add clarification to the popup that it uses scryfall/deck building filters.

EonTAS:
image

 the option added to the shortcuts menu for this option is just called stack which is already the normal draw from deck command

tooomm:
@mainman879 @EonTAS 
Please open new issues for your reports if still relevant.
This is now available in the newest beta.

PR #4650, opened 2022-07-16
Shuffle attachments command

CajunAvenger:
## Related Ticket(s)
None

Cockatrice's only way to randomize a set of cards is to put them on the bottom of the library in a random order, which is pretty inconvenient for mini-decks like Scheme or Contraption decks.

- Cards with attachments will have a "Shuffle attached cards" menu option.
- Using this will randomize the order of the cards attached to it, and log that the player has shuffled the card's attachments.

## Other notes
- This doesn't currently update shortcuts or translations.

## Screenshots

shuffle command
shuffled

### DIFF
--- a/cockatrice/src/abstractcarditem.cpp
+++ b/cockatrice/src/abstractcarditem.cpp
@@ -13,13 +13,12 @@
 #include <algorithm>
 
 AbstractCardItem::AbstractCardItem(const QString &_name, Player *_owner, int _id, QGraphicsItem *parent)
-    : ArrowTarget(_owner, parent), id(_id), name(_name), tapped(false), facedown(false), tapAngle(0),
+    : ArrowTarget(_owner, parent), id(_id), name(_name), tapped(false), facedown(false), obscured(false), tapAngle(0),
       bgColor(Qt::transparent), isHovered(false), realZValue(0)
 {
     setCursor(Qt::OpenHandCursor);
     setFlag(ItemIsSelectable);
     setCacheMode(DeviceCoordinateCache);
-    setObscured(false);
 
     connect(&SettingsCache::instance(), SIGNAL(displayCardNamesChanged()), this, SLOT(callUpdate()));
     cardInfoUpdated();
--- a/cockatrice/src/abstractcarditem.cpp
+++ b/cockatrice/src/abstractcarditem.cpp
@@ -19,6 +19,7 @@ AbstractCardItem::AbstractCardItem(const QString &_name, Player *_owner, int _id
     setCursor(Qt::OpenHandCursor);
     setFlag(ItemIsSelectable);
     setCacheMode(DeviceCoordinateCache);
+    setObscured(false);
 
     connect(&SettingsCache::instance(), SIGNAL(displayCardNamesChanged()), this, SLOT(callUpdate()));
     cardInfoUpdated();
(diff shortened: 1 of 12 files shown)


ebbit1q (changes requested):
this pr performs the shuffle client side! two clients will not see the same result from a shuffle! the shuffle needs to happen in common/server_player.cpp or in a function used in there somewhere otherwise clients will not be able to agree on the randomization.

CajunAvenger:
k, changed the shuffle function so that it syncs on the server. 
- Event_MoveCard now has `repeatable sint32 swap_indexes` instead of `bool shuffle_attached`. This generates the random indexes in server_player.cpp which will be swapped latter
- Player::eventMoveCard passes that event to carditem.h which shuffles as before with the prerolled numbers.
- Changed the menu item to only appear if there's 2+ attachments

future stuff
- want to try anonymize face down cards shuffled this way to removed the semi marked cards aspect of this for face-down libraries
- ebbit had mentioned maybe making this a different command instead of modding move_card
- not sure this is the cleanest setup? of note i think it may want to pass just the swap_indexes array somehow instead of the entire event
- this doesn't double check that the swap_indexes still make sense, there should be a verify step if there's any chance of cards being unattached between the event being built and being run

CajunAvenger:
updates
- Added a way to anonymize shuffled facedown cards by adding `bool obscured` to AbstractCardItem which is set to true when its shuffled and false when it's unattached. Facedown obscured cards are labeled # ??? instead of # id#.
- Now checks that swap_indexes_size <= attachedCards.size to avoid possible illegal swapItemsAt.
- Moved the carditem function to the cpp, it's now iterated in player.cpp and given the indexes instead of given the event and iterating itself.

think this is now good to open up to proper review

cockatrice/src/abstractcarditem.cpp
```diff
@@ -139,7 +139,7 @@ void AbstractCardItem::paintPicture(QPainter *painter, const QSizeF &translatedS
         painter->setBackgroundMode(Qt::OpaqueMode);
         QString nameStr;
         if (facedown)
-            nameStr = "# " + QString::number(id);
+            nameStr = "# " + (obscured ? "???" : QString::number(id));
```
ebbit1q, line 142: hiding this client side does not fix the issue, a user could modify their client and remove this in order to gain advantage, the solution I propose is to create new ids server side for each card, a new id is already created every time a card is moved to play face down.

common/pb/event_move_card.proto
```diff
@@ -16,4 +16,5 @@ message Event_MoveCard {
     optional sint32 y = 9 [default = -1];
     optional sint32 new_card_id = 10 [default = -1];
     optional bool face_down = 11;
+    repeated sint32 swap_indexes = 12;
```
ebbit1q, line 19: this will not be backwards compatible, old clients will be unable to represent this information, resulting in the shuffle not happening at all for them, possibly causing them to become desynchronized and having cards swap places depending on the viewer (moving cards between zones fixes such desynchronizations)

ebbit1q (changes requested):
this is almost in working order, if maintaining backwards compatibility is too cumbersome it's not required but a nice upside (it might require 2 events to be sent for each card but I'm not sure yet about this)

common/pb/event_move_card.proto
CajunAvenger: hrm, could we maybe use position? it's signed so we could associate this with a large negative number that would otherwise not be possible to get, then either encode the indexes in the int or have it be a random number that's then used as a rand seed. i don't know if that fully fixes the backwards compatibility issue though since you'd still need the new client to translate.

ebbit1q: that's just as much not backwards compatible (and more confusing), it's not about not adding new fields it's about the behavior.
my proposal is still to delete the set of old cards and create new ones in a random order, that way it's not possible for clients to track which card is which as they'll be given new ids.
alternatively we could switch the ids of each of the cards without telling the clients which is which until flipped, giving like a single new event telling them which cards are affected, new clients could forget which id is associated with each card name then, while old clients would still see the old card until it's flipped over I guess.
the part that's most important to fix is that clients are told about the outcome of the randomisation, clients shouldn't be given what would be unknown information at all, even if the client we provide would hide it.

ZeldaZach:
Just checking up on this

PR #4651, opened 2022-07-17
Attach and Unattach apply to entire selection

CajunAvenger:
## Related Ticket(s)
Doesn't look like anyone's ticketed this before.

If you select multiple cards and use the menu option to Attach them or Unattach them, only the active card is moved and the rest stay where they were.

- Added childArrows to AttachArrowItem and added them to the mouseEvents.
- Added drawAttachArrow to CardItem. This is mostly identical to drawArrow, but doesn't skip the childArrow that's == this. If that is skipped, the active card is still the only one moved as before.
- actAttach and actUnattach are changed so the menu action now applies to the entire selection.

## Screenshots

arrows
all in a row

### DIFF
--- a/cockatrice/src/arrowitem.cpp
+++ b/cockatrice/src/arrowitem.cpp
@@ -260,6 +260,11 @@ ArrowAttachItem::ArrowAttachItem(ArrowTarget *_startItem)
 {
 }
 
+void ArrowAttachItem::addChildArrow(ArrowAttachItem *childArrow)
+{
+    childArrows.append(childArrow);
+}
+
 void ArrowAttachItem::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
 {
     if (!startItem)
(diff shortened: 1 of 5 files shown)


ebbit1q (changes requested):
~~while testing this I've noticed a bug:~~
~~when you move multiple cards from one zone to another zone they stay selected, if you then immediately attach them to a card on the battlefield without right clicking them their card menu never gets updated, even if you right click them afterwards~~

~~eg:~~
~~start a local game with a deck~~
~~press ctrl y 5 times to move some cards to the stack~~
~~select all of them~~
~~drag them into play~~
~~use ctrl alt a to draw attachment arrows~~
~~select one of the cards to attach to~~
~~up to one card of the attached cards will have its menu updated, the other attached cards will still have the same menu they had while on the stack~~

it appears this is not related to attaching, it simply stays that way until you unselect the cards

CajunAvenger:
> it appears this is not related to attaching, it simply stays that way until you unselect the cards

This occurs when I build Cockatrice:master on its own (though indeed doesn't seem to be happening on my stable installations.)

ebbit1q:
yes, this is a regression introduced by https://github.com/Cockatrice/Cockatrice/pull/4347

ebbit1q (approved):
this seems to work as expected, it basically works exactly the same as drawing arrows does, it might be possible to refactor this to a common class but I don't think it's worth the effort.

this change is client side only and works normally on older servers and older clients, action ratelimiting works as expected and is rather quick to be honest

ZeldaZach (approved):
Not sure why we swapped all the loops around, as it clutters the PR, but it seems to work as expected and makes sense. Nice work!

ebbit1q:
I did that, see individual commits

PR #4653, opened 2022-07-17
Selecting game already open in a tab brings user to that tab.

toinetoine:
## Related Ticket(s)
_none_

When already in game X (with it as an open tab), selecting game X from the game list won't bring the user to that game tab. When spectating multiple games (with multiple game tabs open) it's quite frustrating when selecting a game already joined and not being able to open it.

- Before sending `Command_JoinGame` to the server, it does a check to see if the game is open in a tab, and if so sets that tab as `TabSupervisor`'s current tab. The request to join is still sent to the server. Only change is that client-side check.

## Screenshots
Before:
before

After:
after

### DIFF
--- a/cockatrice/src/gameselector.cpp
+++ b/cockatrice/src/gameselector.cpp
@@ -10,6 +10,7 @@
 #include "pb/serverinfo_game.pb.h"
 #include "pending_command.h"
 #include "tab_account.h"
+#include "tab_game.h"
 #include "tab_room.h"
 #include "tab_supervisor.h"
 
@@ -26,7 +27,7 @@
 #include <QVBoxLayout>
 
 GameSelector::GameSelector(AbstractClient *_client,
-                           const TabSupervisor *_tabSupervisor,
+                           TabSupervisor *_tabSupervisor,
                            TabRoom *_room,
                            const QMap<int, QString> &_rooms,
                            const QMap<int, GameTypeMap> &_gameTypes,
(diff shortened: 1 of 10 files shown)


toinetoine:
Ran `format.sh` to fix formatting then amended commit.

ebbit1q (changes requested):
this is a great idea but I'd like to see some improvements in the implementation, instead of getting the tabs out of the tab supervisor just create a method that asks the supervisor to switch to this game if able returning a bool and then skip sending the command! also the current implementation will ask for a game password even if the game is open already. note that the cast for the supervisor is unnecessary, please remove the const marker for it in the header instead.

toinetoine:
Thanks for reviewing, I applied the recommendations but by making `tabsupervisor` member non-constant in the GameSelector header I had to change the constructors of `MessageLogWidget`, `ChatView`, and `UserContext` (downstream from `GameSelector`) to not require their `Tabsupervisor` parameter as a constant, is this ok? Is there another way to avoid this? 
Thank you for mentioning about password prompt, I should of tested for that. I confirmed after applying your recommendations that the password is no longer prompted for already joined games.

ebbit1q (changes requested):
I didn't notice that because the user context menu created a game selector it now needs to have it's member not be const, it's for the best though, we don't want to cast a const member to something mutable.

the switchToGame function has a loop that continues even after finding a match, this is unnecessary, it can return immediately.

it still performs the cast from const tabgame to mutable qwidget, instead don't use const and remove the cast.

the check still happens at the end of the actJoin action, this means the password is still asked for first.

some tips about the use of const: const means that the value is immutable, it cannot be changed later, this is purely syntactic and it is still possible to cast the value to become mutable.
const is useful when passing something by reference without worrying about its value changing during, which you might have to check again after etc. (using const when not dealing with pointers to values is not popular, it has very little use as the original value will not be affected because it will simply be copied when passed by value)
casting a const to become mutable is highly frowned upon, it reverts the use of the const keyword, making const meaningless.
please check before casting if the cast is actually necessary, casting is a powerful tool that is not easy to verify at compile time, causing many runtime bugs or undefined behavior.

cockatrice/src/tab_supervisor.cpp
```diff
@@ -735,3 +735,18 @@ const ServerInfo_User *TabSupervisor::getOnlineUser(const QString &userName) con
 
     return nullptr;
 };
+
+bool TabSupervisor::switchToGameTabIfAlreadyExists(const int gameId)
+{
+    bool isGameTabExists = false;
+    foreach (const int tabGameId, gameTabs.keys()) {
```
ZeldaZach, line 742: Could shortcut this with `if (gameTabs.keys().index(gameId) == -1)` or smth

cockatrice/src/gameselector.cpp
```diff
@@ -26,7 +27,7 @@
 #include <QVBoxLayout>
 
 GameSelector::GameSelector(AbstractClient *_client,
-                           const TabSupervisor *_tabSupervisor,
```
ZeldaZach, line 29: I've been reading through and don't fully understand why we removed the const-ness of the tabSupervisor everywhere

ebbit1q: opening a user context menu has an option to show that user's games in a game selector

cockatrice/src/tab_supervisor.cpp
ebbit1q: you can simply replace the loop with `gameTabs.contains(gameId)`

cockatrice/src/gameselector.cpp
toinetoine: Since `GameSelector`'s `TabSupervisor`is const can only call const member functions of `TabSupervisor` but couldn't figure out how to make `switchToGameTabIfAlreadyExists` const because it uses the `setCurrentIndex` of `TabSupervisor`'s parent `QTabWidget` to switch to the tab. 

Because `MessageLogWidget`, `ChatView`, and `UserContextMenu` are downstream from `GameSelector` had to also make their `tab_supervisor` member non-constant.

toinetoine:
@ebbit1q Thanks for reviewing, I've applied your recommendations and squashed into 1 commit:

---

> password prompt

It's now checking if the game tab already exists before sending the room command to server so the password is no longer prompted for.
```c++
if (!tabSupervisor->switchToGameTabIfAlreadyExists(game.game_id())) {
    PendingCommand *pend = r->prepareRoomCommand(cmd);
    connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(checkResponse(Response)));
    r->sendRoomCommand(pend);
}
```

---

> the switchToGame function has a loop that continues even after finding a match, this is unnecessary, it can return immediately.

Now the loop is removed and it gets the game by id directly.
```c++
TabGame *tabGame = gameTabs[gameId];
const int gameTabIndex = indexOf(tabGame);
```

---

> it still performs the cast from const tabgame to mutable qwidget, instead don't use const and remove the cast.

I removed the const and cast.

cockatrice/src/gameselector.cpp
```diff
@@ -262,9 +263,11 @@ void GameSelector::actJoin()
         return;
     }
 
-    PendingCommand *pend = r->prepareRoomCommand(cmd);
-    connect(pend, SIGNAL(finished(Response, CommandContainer, QVariant)), this, SLOT(checkResponse(Response)));
-    r->sendRoomCommand(pend);
+    if (!tabSupervisor->switchToGameTabIfAlreadyExists(game.game_id())) {
```
ebbit1q, line 266: the password is still asked for, try moving this check to at least before the password dialog, 18 lines above this one, also try returning immediately, disabling the buttons is not necessary and they will never get enabled again because this happens after when the game gets joined normally.

toinetoine: Moved and confirmed no password prompt.

https://user-images.githubusercontent.com/1865818/197104909-df19a83a-2ece-4484-a4fd-aa3b628d4929.mp4

ebbit1q (approved):
thank you for your contribution.

PR #4654, opened 2022-07-20
CI: Update Qt setup on Windows

tooomm:
- Use most recent patch version automatically (instead pinning 6.3.0 --> 6.3.* to get 6.3.1 by now)
- Use integrated caching, see https://github.com/jurplel/install-qt-action#cache
- Install OpenSSL via Qt tools

~~Edit: Unlike the docs suggest, caching is not yet deployed to v3.~~

~~Edit2: Author stated that v3 is not yet officially released, but we need v3 for Qt6 compatibility.~~

~~Edit3: v3 did even break a few days ago, it's now broken on master too with~~
>~~`Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under 'D:\a\Cockatrice\Cockatrice\action'. Did you forget to run actions/checkout before running your local action?`~~

~~The author is aware and working on it.~~

<br>

~~The jurplel/install-qt-action action also allows to [install OpenSSL via "tools_openssl_x64"](https://github.com/jurplel/install-qt-action#tools).~~
Maybe that's not even needed anymore as [Qt >6.2 has basic SSL functionality](https://bugreports.qt.io/browse/QTBUG-82876) and Qt 5.15 with -schannel option might also work without installing OpenSSL separately. I did not test those yet.


ebbit1q:
adding `tools: 'tools_openssl_x64'` (x86 for 32 bit) is a great alternative to the current script

tooomm:
> adding `tools: 'tools_openssl_x64'` (x86 for 32 bit) is a great alternative to the current script

Before we look into adding a potential better alternative...

Is there a chance the `-schannel` flag gives us enough SSL support without integrating OpenSSL altogether? I'm not experienced enough to judge here.
The above linked Qt bug has some information: [[Windows]: Switch default builds to use Secure Channel for SSL sockets](https://bugreports.qt.io/browse/QTBUG-82876)

tooomm:
The [install-qt-action v3](https://github.com/jurplel/install-qt-action/releases/tag/v3.0.0) is officially released now.

Can somebody more experienced with Qt and OpenSSL please check the mentioned notes and test if we still need to install OpenSSL altogether or the integrated SSL support is sufficient for us?

tooomm:
Update: I tried the integrated way of installing OpenSSL, it seems to work pretty well and also decreases build time. I did not test Cockatrice in a fresh VM.

Somebody else also needs to test if the basic TLS features without OpenSSL via `-schannel` are sufficient for us.
Also dropping [Secure Sockets Layer (SSL) Classes | Qt Network 6](https://doc.qt.io/qt-6/ssl.html).

.github/workflows/desktop-build.yml
```diff
@@ -333,20 +333,15 @@ jobs:
...
       - name: Install Qt ${{matrix.qt_version}} for ${{matrix.target}}
         uses: jurplel/install-qt-action@v3
         with:
-          cached: ${{steps.cache-qt.outputs.cache-hit}}
+          cache: true
           version: ${{matrix.qt_version}}
           arch: win${{matrix.arch}}_${{matrix.qt_arch}}
+          setup-python: false
           modules: ${{matrix.qt_modules}}
+          tools: tools_openssl_${{matrix.arch == 64 && 'x64' || 'x86'}}  # ternary operator styled workaround to get correct arch for openssl
```
ebbit1q, line 344: this isn't as much a workaround, it'd be simpler to just place this value in the matrix or simply use the vcpkg_default_triplet which holds exactly the same value

tooomm: You don't like short-circuit evaluation here?

I reworked the values a bit to use more general terms and tools configuration is now done in the matrix section, too.
Is that better?

ebbit1q (approved):
I don't understand why it was necessary to rename arch to bit or to rename vcpkg_default_triplet only to exclusively use it for vcpkg still but it's not like it really matters.
This seems to work as expected and removes the need for the download openssl script, it downloads the latest version of openssl 1.1.1j and with the correct bitness.
The qt cache has become simpler and doesn't seem to perform differently than before too.
I recommend a beta to test the functionality of the new builds.

tooomm:
I tested this on a fresh device in the meantime and I had no issues with SSL connections.

PR #4662, opened 2022-08-25
Protobuf Flaggage

ZeldaZach:
https://github.com/protocolbuffers/protobuf/issues/10357

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,7 +179,7 @@ else()
   # other: osx/llvm, bsd/llvm
   set(CMAKE_CXX_FLAGS_RELEASE "-O2")
   if(WARNING_AS_ERROR)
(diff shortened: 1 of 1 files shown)


ebbit1q:
I don't like disabling this useful warning

tooomm:
Other option, but also not ideal: We can mark the debug build to be allowed to fail for now.

ebbit1q:
with the new mac builds this version of protobuf that failed building is no longer used, closing

ebbit1q:
nevermind, it still happens I just didn't see it because the build failed because the runner version got bumped to 12 https://github.com/Cockatrice/Cockatrice/actions/runs/4129708130/jobs/7135613996#step:5:468

ebbit1q:
it seems like they currently have a pr open that will address this, it's been stalled for cla reasons though

PR #4663, opened 2022-08-31
fix keepalive being multiplied in server timeout

ebbit1q:
a timeout happened after the client not receiving anything for `keepalive * keepalive * maxtimeout` (5 * 5 * 10) seconds instead of what you'd expect. seems to have been an oversight in https://github.com/Cockatrice/Cockatrice/pull/1313

it now only uses keepalive once instead of twice, this means it should now take 50 seconds to time out when disconnected, we could also change the value of the max timeout (currently 10) to something like 6 to time out in 30 seconds instead.

### DIFF
--- a/cockatrice/src/remoteclient.cpp
+++ b/cockatrice/src/remoteclient.cpp
@@ -541,10 +541,9 @@ void RemoteClient::ping()
         }
     }
 
-    int keepalive = SettingsCache::instance().getKeepAlive();
     int maxTime = timeRunning - lastDataReceived;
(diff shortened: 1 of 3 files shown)


ebbit1q:
I've decided to change it to 15 seconds instead, this means the server will be pinged almost twice as often now and the amount of times it will try is halved. we could change this distribution though.

ZeldaZach (approved):
I don't agree with how the ping() method is defined, but this change does improve things incrementally, so it should be okay for now.

PR #4664, opened 2022-09-01
CI: `macOS-10.15` environment is deprecated

tooomm:
The environment in question will be fully unsupported on GitHub actions by December 1st.
Probably reasonable to have a full release by then (and before merging this) to get macOS 10.14 support once in a release again.

- macOS 10.14 has to be dropped
- macOS 10.15 builds can still be done via macOS 11 with an old Xcode
- Correctly target our build for macOS 11
- Use most recent working Xcode versions (in respect to the targets)

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -190,14 +190,14 @@ jobs:
         include:
           - target: Debug # tests only
             os: macos-latest
-            xcode: 12.5.1
+            xcode: 13.0
             qt_version: 6
             type: Debug
             do_tests: 1
 
           - target: 10.15_Catalina
             os: macos-11
-            xcode: 11.7
+            xcode: 11.7 # allows using macOS 10.15 SDK
             qt_version: 6
             type: Release
             do_tests: 1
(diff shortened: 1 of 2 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -190,30 +190,22 @@ jobs:
...
+            xcode: 11.7 # allows using macOS 10.15 SDK
             qt_version: 6
             type: Release
             do_tests: 1
             make_package: 1
 
           - target: 11_Big_Sur
             os: macos-11
-            xcode: 13.2
+            xcode: 13.0
```
tooomm, line 208: The old value was actually using the macOS 12 SDK...

tooomm:
Status:
- ✔️ [CMake error](https://github.com/Cockatrice/Cockatrice/runs/8143196796?check_suite_focus=true)
Xcode 13 still fails for us when building on macOS 11, unchanged from [last year](https://github.com/Cockatrice/Cockatrice/pull/4449#issuecomment-955239270).
Xcode 12.5.1 is the only option then.

- ❌ [Tests failing](https://github.com/Cockatrice/Cockatrice/runs/8143604722?check_suite_focus=true)
The new build to keep macOS 10.15 support is building, but failing 6/7 tests.

ZeldaZach (approved):

ebbit1q:
this was merged while it was known the tests fail, and they still do, do we want to make a new issue for that? reverting seems pointless as they couldn't even install deps before

PR #4667, opened 2022-09-04
CI: Change Qt install method on macOS

tooomm:
Install via homebrew takes a while

- Use jurplel/install-qt-action, same as for Windows
- The actions has caching build in

---

<details>
<summary>It works for the Qt5 build for "macOS 10.14_Mojave"</summary>

https://github.com/Cockatrice/Cockatrice/runs/8176866997?check_suite_focus=true

</details>

<details>
<summary>It works not for the Qt6 build for "macOS 10.15_Catalina" and the other Qt6 builds</summary>

https://github.com/Cockatrice/Cockatrice/runs/8176867039?check_suite_focus=true#step:5:47

Qt is installed, but CMake doesn't find a suitable version:

>   CMake Error at cmake/FindQtRuntime.cmake:75 (message):
    **No suitable version of Qt was found**
  -- Configuring incomplete, errors occurred!
    CMakeLists.txt:214 (include)

Setting `QTDIR` to `/Users/runner/work/Cockatrice/Qt` didn't change that.

</details>

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -206,7 +206,7 @@ jobs:
           - target: 10.15_Catalina
             os: macos-10.15
             xcode: 12.4
-            qt_version: 6.*
+            qt_version: 6.3.*
             type: Release
             do_tests: 1
             make_package: 1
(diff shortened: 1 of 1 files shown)


tooomm:
Closing. Homebrew is good enough I guess and works.
In case we really need caching or more specific versions of Qt, we can have a look again.

I tried various versions of Qt6 with the same results.
Maybe it's just a CMake config in one of the scripts.

ebbit1q:
I'm happy with homebrew, it verifies the method that we expect people to compile cockatrice themselves with

Issue #4670, opened 2022-09-13
Look at top card, draw a card, drag face down will be revealed even though you dont know that card


MultiDragon:
**System Information:**
```
Client Version: 2.8.0 (2021-01-26)
Client Operating System: EndeavourOS
Build Architecture: 64-bit
Qt Version: 5.15.4
System Locale: en_US
Install Mode: Standard
```

_______________________________________________________________________________________
Looking at the top X cards of your library and then dragging cards on the battlefield face down reveals them, even if you already drew the X cards in question. This is fixed if you close your "looking at library" popup. A video of the bug is attached.

https://user-images.githubusercontent.com/52964431/189984985-7881671c-8f85-4d2b-b4d8-4dcce0b0ac06.mp4

_______________________________________________________________________________________

**Steps to reproduce:**
 - Press Ctrl+W, 1 to look at the top card
 - Press Ctrl+D to draw that card
 - Play top card face down
 - Top card is revealed in chatlog

MultiDragon:
I hope it's not a duplicate, I did search for all "Bugs with face down" and did not find but could be named differently

ebbit1q:
I'm amazed noone noticed this bug before, I will try to fix this asap

ebbit1q:
this is also partially fixed already in #4491, it was just when drawing that it still had the unintentional behavior

PR #4671, opened 2022-09-13
remove cards being looked at from the count on drawing

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4670

drawing cards did not trigger the count of cards being looked at getting reduced, unlike moving cards.
this bug seems to have been present since inception of the feature.

- drawing cards reduces the cards being looked at by that amount

### DIFF
--- a/common/server_cardzone.h
+++ b/common/server_cardzone.h
@@ -68,7 +68,7 @@ public:
     }
     void setCardsBeingLookedAt(int _cardsBeingLookedAt)
     {
-        cardsBeingLookedAt = _cardsBeingLookedAt;
+        cardsBeingLookedAt = qMax(0, _cardsBeingLookedAt);
     }
     bool isCardAtPosLookedAt(int pos) const;
     bool hasCoords() const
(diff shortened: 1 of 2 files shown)


MultiDragon:
Not 100% sure if that works btw since putting cards face down revealed them even beyond the first card in certain cases (the video has look at first card but play two face down)

ebbit1q:
have you tried the changed version? (note that this is a server side change you'd only notice it in local games or servers running this version) see this guide https://github.com/Cockatrice/Cockatrice/wiki/How-to-use-unreleased-versions-created-by-pull-requests

in my tests it now works as intended

MultiDragon:
sorry for not looking at this, I was just assuming things about the source which are in that case not true, I dont actually have a way to test the fix so 🤷🏼

ebbit1q:
you don't have a way to test the fix? did you read the article I linked?

ZeldaZach (approved):

PR #4672, opened 2022-09-14
Fix Qt5 detection when Qt6 is partially present

skwerlman:
## Related Ticket(s)
- N/A

Building cockatrice against Qt5 when only some of the needed sources for Qt6 were present caused Qt5 to be misconfigured, leading to compilation errors like this one:
```
[ 66%] Automatic MOC for target cockatrice_common
[ 66%] Built target cockatrice_common_autogen
[ 66%] Building CXX object common/CMakeFiles/cockatrice_common.dir/cockatrice_common_autogen/mocs_compilation.cpp.o
In file included from /home/sk/tmp/cockatrice/build/common/cockatrice_common_autogen/mocs_compilation.cpp:2:
/home/sk/tmp/cockatrice/build/common/cockatrice_common_autogen/EWIEGA46WW/moc_decklist.cpp:16:2: error: #error "This file was generated using the moc from 6.3.2. It"
   16 | #error "This file was generated using the moc from 6.3.2. It"
      |  ^~~~~
/home/sk/tmp/cockatrice/build/common/cockatrice_common_autogen/EWIEGA46WW/moc_decklist.cpp:17:2: error: #error "cannot be used with the include files from this version of Qt."
   17 | #error "cannot be used with the include files from this version of Qt."
      |  ^~~~~
/home/sk/tmp/cockatrice/build/common/cockatrice_common_autogen/EWIEGA46WW/moc_decklist.cpp:18:2: error: #error "(The moc has changed too much.)"
   18 | #error "(The moc has changed too much.)"
      |  ^~~~~
```

This is caused by `find_package(Qt6)` setting Qt configuration variables even when it doesnt successfully find Qt6.

- find_package(Qt6) will only set Qt configuration variables if it finds all required modules

### DIFF
--- a/cmake/FindQtRuntime.cmake
+++ b/cmake/FindQtRuntime.cmake
@@ -42,31 +42,31 @@ list(REMOVE_DUPLICATES REQUIRED_QT_COMPONENTS)
 
 get_cmake_property(_parent_vars VARIABLES)
 function(have_qt6)
-    # Core5Compat is Qt6 Only, Linguist is now a component in Qt6 instead of an external package
-    find_package(
-      Qt6 6.2.3
-      COMPONENTS Core5Compat ${REQUIRED_QT_COMPONENTS}
-      OPTIONAL_COMPONENTS Linguist
-      QUIET HINTS ${Qt6_DIR}
-    )
-    if(Qt6_FOUND)
-      # Copy all new variables to parent scope, as if we weren't in a function
-      # A bit of an ugly hack, but I don't think there's a nicer way
-      get_cmake_property(_vars VARIABLES)
-      foreach (_var ${_vars})
-          list(FIND _parent_vars ${_var} _idx)
-          # FIND gives the index which starts at 0, or -1 for not found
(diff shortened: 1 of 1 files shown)


ebbit1q (changes requested):
I think instead of doing a fallback to QT5 when the necessary QT6 components are unavailable it'd be preferable to fail with an informing error message. Only when QT6 isn't available at all should we use QT5 for older systems. Systems with QT6 installed can use -DFORCE_USE_QT5=1 to use the old version of QT. I'll add that to the wiki right now.

ZeldaZach:
Going to close in favor of ebbit’s correction. Thanks for your continued effort to support our program!!

skwerlman:
hi ye, sorry about the nonresponsiveness here; been busy with travelling. yea, i think ebbits solution makes sense here

ebbit1q:
I don't think the issue is solved though, I think it should just error out immediately if it finds partial qt6 stuff instead of failing in an unexpected way later.

skwerlman:
oh, it though it wound up in one of your other qt6 prs. i can take another look at this in a bit

PR #4682, opened 2022-10-08
fix crash when a cardmenu becomes an orphan

### LINKED ISSUE #4465: Crash when right-click context menu is open on card in graveyard or exile and that card changes zone
If you have a right-click context menu open on a card in either a graveyard or in exile, the client will crash when that card changes zone.

Steps to reproduce:
1. Put a card in User 1's graveyard.
2. Have User 2 open the graveyard/exile and then open the right-click context menu on that card.
3. Have User 1 move that card to the battlefield while the menu is still open.
4. Observe User 2's client crash to desktop

Cockatrice version 2.8.0
Reproduced on Windows 10.

ebbit1q:
when a cardmenu is closed the cursor on that card reverts to the open hand, this crashed the client when that card would be destroyed or moved the act of reverting to the open hand now happens as an emitted signal, this way it just doesn't exist anymore when the card is deleted.

fixes #4465

### DIFF
--- a/cockatrice/src/carditem.cpp
+++ b/cockatrice/src/carditem.cpp
@@ -33,6 +33,8 @@ CardItem::CardItem(Player *_owner,
     ptMenu = new QMenu;
     moveMenu = new QMenu;
 
+    connect(this, SIGNAL(resetCursorSignal()), this, SLOT(resetCursor()));
+
     retranslateUi();
 }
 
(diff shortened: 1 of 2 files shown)


ebbit1q:
sometimes we fix something trying to maintain the original behavior without realizing the original behavior was pointless, there is no need at all to reset the cursor, opening the cardmenu never changes the cursor in the first place.
the simple return here also skips the passthrough of the event to underlying items etc, this just makes sense imo.

ebbit1q:
would appreciate someone to test this on other platforms, maybe the cursor behavior is different there.

ZeldaZach (approved):

PR #4691, opened 2022-10-18
use qt6 in arch builds

ebbit1q:
updates the arch linux debug build we do to use qt6 instead of qt5

### DIFF
--- a/.ci/ArchLinux/Dockerfile
+++ b/.ci/ArchLinux/Dockerfile
@@ -8,7 +8,6 @@ RUN pacman --sync --refresh --sysupgrade --needed --noconfirm \
         gtest \
         mariadb-libs \
         protobuf \
-        qt6-5compat \
         qt6-base \
         qt6-multimedia \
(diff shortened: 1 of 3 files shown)


.ci/ArchLinux/Dockerfile
```diff
@@ -8,9 +8,11 @@ RUN pacman --sync --refresh --sysupgrade --needed --noconfirm \
         gtest \
         mariadb-libs \
         protobuf \
-        qt5-base \
-        qt5-multimedia \
-        qt5-svg \
-        qt5-tools \
-        qt5-websockets \
+        qt6-5compat \
```
ZeldaZach, line 11: Do we need compat here anymore

ZeldaZach (approved):

.ci/ArchLinux/Dockerfile
ebbit1q: not if I merge 4692 first

PR #4697, opened 2022-10-27
Added hint when drawing 0 cards

### LINKED ISSUE #3530: Add hint to "draws 0 cards" message in log
Drawing from an empty library leads to a `... draws 0 card(s).` message in the game log:
library empty

Having a hint/reason there could be helpful.
`... draws 0 card(s). Library is empty.` or
`... draws 0 card(s). No cards left.` for example.

Tobyclark:
## Related Ticket(s)
- Fixes #3530

When a player drew 0 cards it was unclear why

- Logging a player drawing 0 cards will now result in the message "player had no cards left to draw."

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -369,7 +369,7 @@ void MessageLogWidget::logDrawCards(Player *player, int number)
     if (currentContext == MessageContext_Mulligan) {
         logMulligan(player, number);
     } else {
-        if (deckSize == 0) {
+        if (deckSize == 0  && number == 0) {
             appendHtmlServerMessage(tr("%1 had no cards left to draw.", "")
                                     .arg(sanitizeHtml(player->getName())));
         }
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -368,7 +368,7 @@ void MessageLogWidget::logDrawCards(Player *player, int number)
     if (currentContext == MessageContext_Mulligan) {
         logMulligan(player, number);
     } else {
-        if (number == 0) {
+        if (number == 0 && player->getZones().find()) {
             appendHtmlServerMessage(tr("%1 had no cards left to draw.", "")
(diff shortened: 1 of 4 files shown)


ebbit1q (changes requested):
it should check first if there are actually no cards left in the library instead of assuming drawing zero means there are none left.

cockatrice/src/messagelogwidget.cpp
```diff
@@ -364,13 +364,20 @@ void MessageLogWidget::logMoveCard(Player *player,
...
 void MessageLogWidget::logDrawCards(Player *player, int number)
 {
+    int deckSize = player->getZones().value("deck")->getCards().size();
     soundEngine->playSound("draw_card");
     if (currentContext == MessageContext_Mulligan) {
         logMulligan(player, number);
     } else {
-        appendHtmlServerMessage(tr("%1 draws %2 card(s).", "", number)
+        if (deckSize == 0  && number == 0) {
+            appendHtmlServerMessage(tr("%1 had no cards left to draw.", "")
```
ebbit1q, line 373:
```suggestion
            appendHtmlServerMessage(tr("%1 has no cards left to draw.")
```
most of our log messages use current tense instead of past tense, I think it'd make more sense to reword this in a way that works better in current tense, the original ticket has a suggestion in the way of "player draws from an empty library" or perhaps "player tries to draw from empty library", @tooomm might have some suggestions

cockatrice/src/messagelogwidget.cpp
```diff
@@ -364,13 +364,20 @@ void MessageLogWidget::logMoveCard(Player *player,
 
 void MessageLogWidget::logDrawCards(Player *player, int number)
 {
+    int deckSize = player->getZones().value("deck")->getCards().size();
```
ebbit1q, line 367: considering the player itself is the only one to ever emit the signal this connects to I think it'd make more sense to simply add it to the signal and this function as an extra parameter.

ebbit1q (approved):
excellent, thank you for your contribution!

PR #4705, opened 2022-11-11
Add left nav

brentmclark:
- `<Header />` component renamed to `LeftNav` and used on the left side of the screen instead of the top
- An empty bottom bar was introduce for future enhancements
- `<Layout />` component added to allow for more flexibility in the layout.
- No nav elements (left or bottom) on the login screen

## Screenshots
image


webclient/src/containers/Layout/Layout.tsx
```diff
@@ -0,0 +1,20 @@
+
+import { Header } from 'components';
+
+function Layout(props:LayoutProps) {
+    const { children, className, showNav = true } = props;
+    return (
+        <div className={className}>
+            <Header showNav={showNav} />
```
brentmclark, line 8: @seavor this line raises the TS the error below.  Any insight?

```
Type '{ showNav: boolean; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<RouteComponentProps<any, StaticContext, unknown>, never>, any, any>> & Readonly<...> & Readonly<...>'.
  Property 'showNav' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<RouteComponentProps<any, StaticContext, unknown>, never>, any, any>> & Readonly<...> & Readonly<...>'.ts(2322)
```

seavor: Hmmm, not without playing with it. I'm not sure ive ever tried assigning default values within a deconstructor  though.

webclient/src/containers/Login/Login.tsx
```diff
@@ -223,7 +224,7 @@ const Login = ({ state, description }: LoginProps) => {
   };
 
   return (
-    <div className={'login overflow-scroll ' + classes.root}>
+    <Layout className={'login overflow-scroll ' + classes.root} showNav={false}>
```
seavor, line 227: I like where this is going, might prefer the component be called MainNavigation or something to that effect.

webclient/src/containers/Layout/Layout.tsx
seavor: Oh, probably because you're trying to assign a value to a readonly prop property.

brentmclark:
That seems like a reasonable thing to try.  Thanks.

webclient/src/containers/Layout/Layout.tsx
brentmclark: I'll move this to a `defaultProp` and see how it responds.  Thanks

webclient/src/containers/Login/Login.tsx
brentmclark: Can you expand on your concerns?  My mental model is that changing routes will generate a fresh component tree with its own periods for the layout component.  There shouldn't be any concern for stake components hanging around.

seavor: Lets have a quick chat about it tomorrow or this weekend, but dont treat it as a blocker.

webclient/src/containers/Layout/Layout.tsx
brentmclark: Nope, that didn't do it.  I was able to figure it out, though.  It's the use of `withRouter`; it doesn't play nice with passing down props.  I'll take a swing at converting this component to a functional component so we can take advantage of `useRouter`.

webclient/src/containers/Login/Login.tsx
brentmclark: Following up, we chatted.  Seavor's concern is valid in certain contexts, but not the way this was implemented.

ebbit1q:
I put this on the beta: https://beta.cockatrice.us/webclient

brentmclark:
Need to fix a smol bug in here; then it should be ready for merging.

https://trello.com/c/CS0pcSBE/21-bug-fix-issue-with-login-screen-ui-footer-still-shows

brentmclark:
This is ready for review again.

ZeldaZach (approved):

PR #4720, opened 2022-12-10
Display Welcome Message as Most Recent Message

### LINKED ISSUE #4110: server welcome message gets scrolled up by backlog messages
the server leaves a first message to welcome players to the server in the chat, unfortunately this will be obscured by the backlog of "missed messages" that are placed in the chat afterwards.
it'd be preferable to simply place the welcome message after the backlog so it's more visible

SlightlyCircuitous:
## Related Ticket(s)
- Fixes #4110 

Room welcome message is currently displayed as the first message in the chat and then immediately buried by any old messages from people talking in the room before the user joins.

- Room welcome message will display as the most recent message in the chat box, instead of the oldest

## Screenshots
Screenshots are from a locally built server, hence the sparsity.

Current code:
Screenshot from 2022-12-10 15-37-38

Proposed change:
Screenshot from 2022-12-10 15-39-21


ebbit1q (approved):
looks good

PR #4721, opened 2022-12-18
Some image sources require that a proper user agent is sent.

peterjschroeder:
Some image sources require that a proper user agent is sent.

Example:

https://www.hcrealms.com/units/aa001.jpg


ebbit1q:
I don't think we should hardcode this string in our sources, we could instead use a configuration option, or even just put it in the settings.ini to set if you'd like by editing it.

ZeldaZach:
I like the idea of settings.ini it personally, compared to a hardcode

ZeldaZach:
If you'd like to update this PR and follow our recommendations, I'd love to see this merged :) cc @peterjschroeder

PR #4726, opened 2023-01-02
Web client: building protobuf instead of loading

skybloo:
Web client currently includes all .proto files and loads them at runtime. This added reflection as a build step to do that ahead of time and compiled to static javascript for better load time and smaller size

for end users hopefully nothing


brentmclark (approved):
This is really great work; thank you very much for all the effort.

I would like to hear your thoughts on that optional param vs null union topic, but it's not a merge blocker for me.  If you decide to normalize the implementations a follow-up PR is just fine if you have time.  Otherwise, I'll do it as time permits.

Also, would it be possible on future PRs to have you change your code formatter settings to only change lines you touched instead of the entire file?  there are a LOT of stylistic changes in this PR, and although they are all warranted it was time-consuming to separate them from functional changes that are more meaningful to review.

I'll live if that's not something you can do, but it would help me review the PRs faster if there was less noise.

brentmclark:
Oh whoops, the builds are failing because they are expecting a `package-lock.json` file to be in this PR as well.  Can you add that @skybloo?

ZeldaZach:
Hey there, any update on this @skybloo :D

tooomm:
Now that the Webatrice project got moved into it's own repository, this PR should no longer target the main Cockatrice repo, but rather [Webatrice](https://github.com/Cockatrice/Webatrice)/[Sockatrice](https://github.com/Cockatrice/Sockatrice).

FYI @seavor

PR #4728, opened 2023-01-08
Fixed ptSeparation oracle issue

PakhomCh:
## Related Ticket(s)
- Fixes p

ptSeparation oracle issue is an issue which causes cards from JSON file to have a '/' symbol  between its power and an unexisting toughness value.

- Additional check will be added to see, if ptSeparator is ever needed.

## Screenshots
The issue examle:
image

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -334,7 +334,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet,
 
         power = getStringPropertyFromMap(card, "power");
         toughness = getStringPropertyFromMap(card, "toughness");
-        if (power.isEmpty() || power == "" || toughness.isEmpty() || toughness == ""){
+        if (power.isEmpty() || toughness.isEmpty()){
             ptSeparator = "";
         }
         if (!(power.isEmpty() && toughness.isEmpty())) {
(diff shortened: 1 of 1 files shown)


oracle/src/oracleimporter.cpp
```diff
@@ -334,6 +334,9 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet,
 
         power = getStringPropertyFromMap(card, "power");
         toughness = getStringPropertyFromMap(card, "toughness");
+        if (power.isEmpty() || power == "" || toughness.isEmpty() || toughness == ""){
```
ZeldaZach, line 337: .isEmpty() implicitly checks if it's "", so this can be simply reduced to `if (power.isEmpty() || toughenss.isEmpty())` 

I'd also consider doing this a different way possibly, as the next if statement handles a check to make sure they're both set properly. The if statement below seems to be wrong in its goals.

A rough idea is to just build up a final response
if power, then builderArray.append(power)
if toughness, then builderArray.append(toughness)
pt = builderArray.join("ptSeparator") as an example

PakhomCh:
```suggestion
        if (power.isEmpty() || toughness.isEmpty()){
```
The fact, that following if statement checks if there is at least one value among power and toughness, makes the following code executed each time it is possible without throwing an exception/error. However, just turning ptSeparator into an empty string when there is no value to separate feels like an easier way to go and being more optimal solution.

PR #4730, opened 2023-01-10
Find OpenSSL on Windows

ZeldaZach:


ebbit1q (changes requested):
as far as I know our binaries require version 1.1 of open ssl, the packages created by our ci with this change all ship version 3 x64 (including the 32 bit one) so this seems to break which files get included in the packaged installer.

ebbit1q:
it now correctly includes the right version of openssl, though for some reason it still includes v3 x64 in addition for all of them, would it be possible to have it only package the version we use (and is the right arch)

PR #4733, opened 2023-01-28
FIX #4665: Address missing sound on Qt5 Builds

### LINKED ISSUE #4665: No sound on Windows with (new) Qt5 build
**System Information:**
Client Version: 2.8.1-beta.10 (2022-06-08)
Build Architecture: 64-bit
Qt Version: 5.15.2
System Locale: en_GB
Install Mode: Standard

After testing some 64-bit Qt5 build on Windows 10, I realized that they all have no sounds for me.
I found out, that `2.8.1-beta.6` (which was the last release that was still all Qt5) does play sound.
So with the switch to Qt6 (even though Qt5 was reintroduced later again, #4639) sound is lost now for Qt5 users on Windows it seems.

@motherofruins Do you have sound on your Qt5 build that you run on Windows 7? Which version is it?
Can you also try the new 2.8.1-beta.10/Win7/64bit (Qt5), please?

This issue is not directly related to #4642, I tried a build before that change as well and had no sound either.

Edit: Same issue when running the 32-bit version of 2.8.1-beta.10 (also Qt5) on my 64-bit Windows 10.

ZeldaZach:


ebbit1q:
this is a windows only issue, I can't verify if this fixed that, @tooomm @MotherofRuins ?

ebbit1q:
so it now ships two more dlls in the installer? do we not know which it actually needs?

ZeldaZach:
Confirmed functional

PR #4734, opened 2023-01-28
Fix #4706: don't replace ampersands when loading from plain text

### LINKED ISSUE #4706: "&" is converted to // from text input
**System Information:**

Client Version: 2.8.1-beta.6 (2022-01-16)
Build Architecture: 64-bit
Qt Version: 5.15.2
System Locale: en_GB
Install Mode: Standard

when a card with "&" in it's name is loaded form a txt file or form clipboard the & is replaced with //
expor the following deck to clipboard: 
image
paste deck form clipboard window: 
image
decklist form menu:
```
// 1 Maindeck
// 1 Planeswalker
1 Minsc & Boo, Timeless Heroes
```
resulting decklist: 
image

step 1: create a deck with minsc & boo, timeless heroes
step 2: export this deck to clipboard/txt file
step 3: reload form txt file or clipboard
**Steps to reproduce:**
 - Do A
 - Do B
 - Do C ...

ZeldaZach:
Tested by loading in from clipboard

```
1 Minsc & Boo, Timeless Heroes
1 Fire & Ice
```

### DIFF
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -505,7 +505,8 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
     const QRegularExpression reBraceDigit(R"( ?\([\dA-Z]+\) *\d+$)");
     const QHash<QRegularExpression, QString> differences{{QRegularExpression("’"), QString("'")},
                                                          {QRegularExpression("Æ"), QString("Ae")},
-                                                         {QRegularExpression("æ"), QString("ae")}};
+                                                         {QRegularExpression("æ"), QString("ae")},
+                                                         {QRegularExpression(" ?[|/]+ ?"), QString(" // ")}};
 
     cleanList();
 
--- a/common/decklist.cpp
+++ b/common/decklist.cpp
@@ -505,9 +505,7 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
     const QRegularExpression reBraceDigit(R"( ?\([\dA-Z]+\) *\d+$)");
     const QHash<QRegularExpression, QString> differences{{QRegularExpression("’"), QString("'")},
                                                          {QRegularExpression("Æ"), QString("Ae")},
-                                                         {QRegularExpression("æ"), QString("ae")},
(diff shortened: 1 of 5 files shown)


ebbit1q:
I think this is the wrong way to go about this, searching the db is one of the most expensive operations the application can do and I'd like to minimise it, instead of always searching the db every single time a replacement is done maybe just don't do the & replacement, perhaps do it after it searches for the name (see db guessCard which also does up to three db searches per card and mangles it a bit).

tests/loading_from_clipboard/loading_from_clipboard_test.cpp
```diff
@@ -145,7 +146,11 @@ TEST(LoadingFromClipboardTest, EdgeCaseTesting)
 )");
 
     Result result("DeckName", "Comment 1\n\nComment [two]\n(test) Æ ’ | / (3)",
-                  {{"Aether Adept", 1}, {"Fire // Ice", 2}, {"Pain // Suffering", 3}, {"Forest", 4}},
+                  {{"Aether Adept", 1},
+                   {"Fire // Ice", 2},
+                   {"Minsc & Boo, Timeless Heroes", 1},
```
ebbit1q, line 151: these used to go 1234, what have you done zach :upside_down_face:

common/decklist.cpp
```diff
@@ -529,15 +528,15 @@ bool DeckList::loadFromStream_Plain(QTextStream &in)
     }
 
     // find sideboard position, if marks are used this won't be needed
-    int sBStart = -1;
+    qsizetype sBStart = -1;
```
ebbit1q, line 531: Qt 5.9.5 doesn't have the qsizetype typedef, the int was fine?

ebbit1q:
```suggestion
    int sBStart = -1;
```

ebbit1q (approved):
I like seeing const references, though I am still not sure how impactful they are for qt objects that are always pointers to their data anyway

PR #4737, opened 2023-02-01
Enable buttons for current game when receiving server response

### LINKED ISSUE #4698: join as spectator on a game that can't be joined as a player enables the join button unintentionally
**System Information:**
Client Version: 2.8.1-custom(2c31e04) (2022-10-20)
Client Operating System: Pop!_OS 22.04 LTS
Build Architecture: 64-bit
Qt Version: 6.2.4
System Locale: en_US
Install Mode: Standard

games that are eg full cannot be joined by more players and the client greys out the join button, making it inaccessible.
joining a game initially greys out both buttons then enables them again when the game is joined, unconditionally, this means that if it was greyed out before it is now available again.
clicking this button while the game is full will change you into a spectator instead, so this isn't very impactful.

**Steps to reproduce:**
 - find a game that is full and accepts spectators or create one on another account
 - click join game as spectator
 - leave the game again
 - the normal join button is now available until you select a different game in the list

Elarnon:
Previously, upon joining a game, we were unconditionally re-enabling the "Join" button in the lobby, even if it was not enabled in the first place, causing #4698. This could also lead to issues where if the user selects a different game after joining (which they can do in case of e.g. network connectivity issues), the "Join as spectator" button could get incorrectly disabled.

This fixes #4698 by re-enabling the buttons based on the state of the currently selected game at the time the response is received. This also avoids inconsistencies if a different game has been selected in between joining and receiving a response from the server.

## Related Ticket(s)
- Fixes #4698

### DIFF
--- a/cockatrice/src/gameselector.cpp
+++ b/cockatrice/src/gameselector.cpp
@@ -198,9 +198,19 @@ void GameSelector::actCreate()
 
 void GameSelector::checkResponse(const Response &response)
 {
-    if (createButton)
-        createButton->setEnabled(true);
-    joinButton->setEnabled(true);
+    // NB: We re-enable buttons for the currently selected game, which may not
+    // be the same game as the one for which we are processing a response.
+    // This could lead to situations where we join a game, select a different
+    // game, join the new game, then receive the response for the original
+    // join, which would re-activate the buttons for the second game too soon.
+    // However, that is better than doing things the other ways, because then
+    // the response to the first game could lock us out of join/join as
+    // spectator for the second game!
+    //
+    // Ideally we should have a way to figure out if the current game is the
+    // same as the one we are getting a response for, but it's probably not
+    // worth the trouble.
+    enableButtons();
 
     switch (response.response_code()) {
         case Response::RespNotInRoom:
@@ -229,9 +239,6 @@ void GameSelector::checkResponse(const Response &response)
             break;
         default:;
     }
-
-    if (response.response_code() != Response::RespSpectatorsNotAllowed)
-        spectateButton->setEnabled(true);
 }
 
 void GameSelector::actJoin()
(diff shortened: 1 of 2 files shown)


ebbit1q (changes requested):
with this change joining a game disables the create button permanently, you have to rejoin the room otherwise you can't create new games, instead the create button should always be reenabled after you joining a game gets a reply from the server.

also, I don't think race conditions are really the reason for the buttons needing to be disabled (this wouldn't completely fix that), I think it's just to stop people that start repeatedly pressing buttons if the server's response starts to take longer than normal (like on a slow connection or intermittent wifi signal, you need an indicator that your mouseclick was actually registered), there is probably a commit message from 2014 that says why this was done but I'll look it up when I have more time.

Elarnon:
> with this change joining a game disables the create button permanently, you have to rejoin the room otherwise you can't create new games, instead the create button should always be reenabled after you joining a game gets a reply from the server.

Woops that was a typo due to a last-minute refactoring, `createButton->setEnabled(false)` should be `createButton->setEnabled(true)` in `enableButtons()`. Should be fixed now, my apologies.

> also, I don't think race conditions are really the reason for the buttons needing to be disabled (this wouldn't completely fix that), I think it's just to stop people that start repeatedly pressing buttons if the server's response starts to take longer than normal (like on a slow connection or intermittent wifi signal, you need an indicator that your mouseclick was actually registered), there is probably a commit message from 2014 that says why this was done but I'll look it up when I have more time.

I believe we agree here, I'm not saying disabling the buttons is trying to fix a race condition. I thought that is was possible (but unlinkely) for the server response to be processed before `sendRoomCommand` returns, which would result in the buttons staying disabled after the response. It looks like I was mistaken about this, so I removed the comment mentioning race conditions.

ebbit1q (approved):
Much appreciated!

PR #4742, opened 2023-02-06
Add Oracle support for persistent & Fix persistent on reverse-related

CajunAvenger:
## Related Ticket(s)
- #4646 

This is the followup to #4646 to add persistent relations into Oracle now that MTGJson supports Conjure spellbooks. It also fixes reverse-related cardRelations not staying persistent.

- Oracle now checks MTGJson's `relatedCards:{spellbook:[]}` property and sets a persistent relation for each of those cards
- refreshCachedReverseRelatedCards now does getIsPersistent so persistent reverse-related cardRelations don't get reset to not persistent.

## Other
- Smaller sets json used for testing: [YDMU.zip](https://github.com/Cockatrice/Cockatrice/files/10660056/YDMU.zip)
- I recognize this might become redundant with #4736, but I assume we'd still need Oracle to make that xml in the first place.

### DIFF
--- a/cockatrice/src/carddatabase.cpp
+++ b/cockatrice/src/carddatabase.cpp
@@ -589,7 +589,8 @@ void CardDatabase::refreshCachedReverseRelatedCards()
 
             auto *newCardRelation = new CardRelation(card->getName(), cardRelation->getDoesAttach(),
                                                      cardRelation->getIsCreateAllExclusion(),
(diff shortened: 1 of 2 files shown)


thopp-tudelft (approved):
Excellent, works as intended.

ebbit1q (approved):

PR #4743, opened 2023-02-07
Allow revealing specific cards from hand and library

Elarnon:

Currently Cockatrice allows revealing the whole hand, or one card at random from the hand. Sometimes, a player needs to reveal a specific card from their hand instead, which is not supported. To achieve a similar effect, players usually move the corresponding card (or cards) to a public zone, then back to their hand. While this works, it is unsatisfactory (compared to a regular reveal, you can't keep the "revealed" window around, for one) and somewhat non-intuitive.

- When selecting a single card or a set of cards in your hand and some other zones (e.g. when viewing your library), there is a new "Reveal to..." menu. Using this menu, you can reveal the selected cards to one specific player or to all players.
- Old clients (without the patch) will correctly see the cards revealed by new clients using a new server (with the patch), but they won't be able to reveal their own cards.
- New clients connected to old servers (without the patch) will only be able to reveal cards from their hand one at a time. Trying to reveal multiple cards at once on an old server will only reveal one (unspecified) card amongst the selected cards.

## Technical details

To implement the functionality at the protocol level, the existing RevealCards command is extended to support revealing multiple specific cards. This is done by making `card_id` a non-packed repeated field in the `Command_RevealCards` and `Event_RevealCards` protobufs.  Using a non-packed repeated fields allows maintaining backwards compatibility: an empty optional field is encoded the same way as an empty non-packed list, an optional field with a value is encoded the same way as a one-element non-packed list, and when decoding a multi-elements non-packed list as an optional, only the last item in the list is read.

Since the RevealCards command already exists, and due to the compatible encodings, a new client connecting to an old server can reveal a single specific card from their hand. When trying to reveal multiple cards at once, the old server will only see the request for one of the cards to be revealed, and the player will have to reveal each card separately.

On the other hand, `Event_RevealedCards` already has an explicit list of cards revealed by the server, and the `card_id` field is only used when exactly one card has been revealed: thus, old and new clients will behave identically when receiving a new `Event_RevealedCards`. In particular, if a player using a new client reveals multiple cards from their hand on a new server, another player using an old client will correctly see all the revealed cards.

The approach used to build the "Reveal to..." menu is slightly different from the approach used to build other player selection menus. Because the "Reveal to..." menu is specific to each card, but must also be updated whenever a player is added to or removed from the game, I chose to re-create it on the fly whenever a card is clicked, as that seemed the safest way to avoid both memory leaks and inconsistent state given my understanding of the code.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -573,6 +573,8 @@ void Player::addPlayer(Player *player)
         newAction->setData(player->getId());
         connect(newAction, SIGNAL(triggered()), this, SLOT(playerListActionTriggered()));
     }
+
+    playersInfo.append(qMakePair(player->getName(), player->getId()));
 }
 
 void Player::removePlayer(Player *player)
@@ -588,6 +590,14 @@ void Player::removePlayer(Player *player)
                 playerList->removeAction(j);
                 j->deleteLater();
             }
+    }
+
+    for (auto it = playersInfo.begin(); it != playersInfo.end();) {
+        if (it->second == player->getId()) {
+            it = playersInfo.erase(it);
+        } else {
+            ++it;
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


ZeldaZach:
Thanks for all of your help! This looks solid. If you'd like to join our Dev group chat on Discord, join us at http://cockatrice.us/discord in the #dev channel :)

ZeldaZach (approved):

Elarnon:
Note sure how much I want to get involved, my contributions may stay limited to small QoL changes that I need for myself after picking Cockatrice up again recently. I'll check the Discord out though, thanks!

ebbit1q:
don't worry about it, we just thought it'd be nice to chat because we appreciate your prs :+1:

PR #4749, opened 2023-02-15
Improved Double-Faced Card Support with Dynamic Zones

### LINKED ISSUE #4081: Double-Faced Cards Rework
Hello, I was wondering if it's possible to make it so the double-faced cards have something akin to a "flip card" option. It would 
help save space on the board with how the cards work currently.

Elarnon:
## Related Ticket(s)
- Fixes #4081

Currenty, Double-Faced Cards are dealt with by creating tokens to represent the back of the card, and attaching the original card to the token. While this works, it is a not very ergonomic: it means transformed cards take more space than necessary on the board, creates confusing visual states when other cards are attached to the transformed card, and prevent transformed cards from being attachments themselves. Personally, I often end up stashing away the front face into exile.

The solution implemented here, and described in https://github.com/Cockatrice/Cockatrice/issues/4081#issuecomment-1424418693, is based on the "stashing the original card away" idea, but it automates it significantly. More precisely, the idea is to:

 - Create a temporary zone to hold the original card representing its front-side, and move the original card to that zone
 - Create a token to represent the back-side (or the flipped version, for Kamigawa flip cards) of the card
 - If the token is created in the same zone as the original card, put it on the same position the card used to occupy (to give the impression that it is actually transforming)
 - When the token is destroyed due to a zone change, move all the cards (usually just the front face, but e.g. for Meld cards there could be multiple things there) from the temporary zone to the zone that the token was going to, then destroy the temporary zone.

This is fairly straightforward, but requires a new feature: it should be possible to create (and destroy) dynamic zones on the fly during the course of the game, and to somehow "attach" these zones to cards. Unfortunately, existing clients do not support zones others than what the game starts with, and would ignore card movements involving such dynamic zones, leaving these clients in an inconsistent state.

This PR thus proceeds in three steps, and is comprised of several commits that I suggest to review in order (each commit also has a fairly descriptive description in the commit message):

 - The first 3 commits perform some small-ish refactorings that help make the following changes go more smoothly
 - The next 2 commits implement the actual feature in three steps:
    + Protocol changes to support dynamic zones, introducing new events CreateZone, AttachZone and DestroyZone, and augmenting the CreateToken command to create dynamic zones attached to the token. If the token is marked as "destroy on zone changes", the attached zone is destroyed when the token changes zones, and its content moved to wherever the token was going (or the battlefield, if the token was going to a zone that ended up destroyed in the process).
    + Client-side implementation of the feature, including a new `attach="transform"` database attribute and the creation of appropriate menu entries to transform a card into its back-face and the back-face back to its front-face
 - The last 2 commits are concerned with backwards-compatibility and introduce a new protocol feature for backwards-compatibility, which I call "alternative events". This allows an EventContainer to contain alternative event streams based on the features supported by the client. This protocol feature is itself fully backwards-compatible, and it is used to make dynamic zones compatible with legacy clients by emitting combinations of CreateToken and DestroyCard events instead of MoveCard events.

These are rather big changes, compared to alternative, simpler implementations of the "card transform" mechanic such as simply hiding the attached card, as suggested in https://github.com/Cockatrice/Cockatrice/issues/4081#issuecomment-1424935648. However, I think it is a rather elegant solution to the problem, and it opens up new exciting possibilities for Cockatrice: for instance, this allows implementing the "mini-deck" feature requested in #4650 as a client-side feature without further protocol change, by piggy-backing on the existing Shuffle command (I may have cheated in some very natural changes to make the "shufflability" of a zone configurable when designing the zone_config.proto for that very purpose).


Elarnon:
Fixed build issues on Qt 5 and Windows, should be good to review now :)

I do not understand the macOS failure, it seems to be related to a warning protobuf headers.

ebbit1q:
the macos failure has been an issue for almost a year now, it's because the generated protobuf files don't play nice with -Wall, specifically unused parameters, this is currently being worked on at protobuf but it might take a bit longer for it to get pushed to homebrew still. we've just been ignoring the build because it doesn't really matter.

ebbit1q:
While the idea of dynamic zones is pretty nice and might have lots of applications, is there any real reason to move back of the card there? why not simply remove it and add it back later? or why not make the hiding completely client side and simply have the hidden card have a little parameter "hide while attached" added to the protocol?

Elarnon:
> While the idea of dynamic zones is pretty nice and might have lots of applications, is there any real reason to move back of the card there? why not simply remove it and add it back later? or why not make the hiding completely client side and simply have the hidden card have a little parameter "hide while attached" added to the protocol?

Good questions :) I wrote a lot about the technical implementations but have fallen short in explaining my rationale. The reasoning follows from trying to answer the question: where do we keep information about the front face(s)?

 - It can be an attachment with an explicit "hidden" flag. That's the simplest solution, but also the poorest UX-wise for a transformed card, because when moving the token to the graveyard, the original card will pop up on the battlefield. Moreover, since attachments can't themselves have attachments, cards that transform into an Aura such as Binding Geist // Spectral Binding will get confusing to use: they will look like regular cards (because the attachment is hidden), but without being able to use them normally (they can't be attached to another card, because they have an attachment).
 - If the card is destroyed then recreated, we still need to keep track of the front face(s) somewhere. It could be the client's responsibility to recreate the card, but I think we'll agree that's brittle and probably not a good idea, so the server needs to have a list of "stashed away" cards on the token to be recreated at the appropriate time. Unlike an attachment, these cards would be completely hidden from the client: no way to view them (I don't believe MTG has different cards with the same back faces, but other games might) or alter them (for Meld, it'd be easier UX-wise to add the two halves in separate steps) without introducing more commands to the protocol.

From a UX perspective, having cards stashed away is better than having them as an attachment from the reasons outlined above. Using a zone is just the better version of stashing the cards away: it is the existing Cockatrice concept for representing a set of cards with shared properties, and it already has support for interactions in the protocol.

Elarnon:
BTW I made a mistake in my refactoring of `Server_Player::moveCards` --- I forgot that for cards in a hidden zone, `card->getId()` can't be used as an identifier. Will fix later today.

Elarnon:
> BTW I made a mistake in my refactoring of `Server_Player::moveCards` --- I forgot that for cards in a hidden zone, `card->getId()` can't be used as an identifier. Will fix later today.

Done.

I will also make a simplified version of this PR that does the "stashing cards away" only, without dynamic. I went away with the idea of dynamic zones since it seemed the right way to tackle the problem (I still think it is) and it felt natural to do it all at once, but I don't want anyone to feel like I'm forcing this idea. The dynamic zone stuff does add a bit of complexity and it'd be best to discuss it on its own.

Elarnon:
Opened #4753 with a simpler implementation that doesn't use dynamic zones but simply stashes the card away, and opened #4754 to discuss the idea of dynamic zones, so I'll close this now.

PR #4752, opened 2023-02-16
Corrected edition search syntax

### LINKED ISSUE #4508: filtering doesn't work as described by the infobox
image

Maybe the filtering method has changed and no one updated the infobox? Or maybe it is just broken?

SlightlyCircuitous:
## Related Ticket(s)
- Fixes #4508 

The Search Help pop-up incorrectly lists set names separated by "," as a valid way to specify "or" in Cockatrice. It also lists (e:lea e:leb) as a valid construction but two set searches in a row within a grouping is invalid in Cockatrice. Both of these constructions work in Scryfall search; however, which is likely where this incorrect syntax was copied from.

The Search Help pop up will only list constructions that are valid in Cockatrice. I checked all the other constructions listed and they seem to be valid.

Please let me know if I edited the correct file to do this (just took a guess based on the contents and the tab_deck_editor.cpp code) or if further changes are required outside of this file.

### DIFF
--- a/cockatrice/resources/help/search.md
+++ b/cockatrice/resources/help/search.md
@@ -46,8 +46,7 @@ The search bar recognizes a set of special commands similar to some other card d
 
 <dt><u>E</u>dition:</dt>
 <dd>[set:lea](#set:lea) <small>(Cards that appear in Alpha, which has the set code LEA)</small></dd>
-<dd>[e:lea,leb](#e:lea,leb) <small>(Cards that appear in Alpha or Beta)</small></dd>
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
thanks

Esnos33:
Hi, @ebbit1q, do you know if this are all possible commands, or are there some more? Particularly if there are special symbols other then {T} and if you can use regex.

ebbit1q:
commands are defined in using peglib, here https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/filter_string.cpp
special symbols are not treated differently at all, they simply exist in the source text field
there is no option for regex currently

Esnos33:
It makes sens. Thanks for answer.

PR #4753, opened 2023-02-17
Better support Double-Faced Cards

### LINKED ISSUE #4081: Double-Faced Cards Rework

Elarnon:
## Related Ticket(s)
- Fixes #4081
- This is a simplified version of #4749 that does not implement dynamic zones

Currenty, Double-Faced Cards are dealt with by creating tokens to represent the back of the card, and attaching the original card to the token. While this works, it is a not very ergonomic: it means transformed cards take more space than necessary on the board, creates confusing visual states when other cards are attached to the transformed card, and prevent transformed cards from being attachments themselves. Personally, I often end up stashing away the front face into exile.

This patch allows cards to be (virtually) transformed into other cards while preserving their state, essentially implemeting the MTG mechanic of the same name.

On the server side, this is implemented by allowing cards to be "stashed away". A card that is stashed away is not in any zone, but is instead owned by another card. When a token is destroyed due to a zone change, if it had a card stashed away, that card is placed in the target zone instead of the token.

On the database side, `attach="transform"` is used on `<reverse>` and `<reverse-related>` to indicate that the created token should be transformed this way. When it is present, the corresponding menu entry changes from "Token: Attach to XXX" into "Token: Transform into XXX".

Old servers ignore the new field in `Command_CreateToken` and will perform a regular attachment, as currently.

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1791,12 +1791,12 @@ void Player::createCard(const CardItem *sourceCard,
 
         case CardRelation::AttachTo:
             cmd.set_target_card_id(sourceCard->getId());
-            cmd.set_target_mode(Command_CreateToken::REVERSE_ATTACH);
+            cmd.set_target_mode(Command_CreateToken::ATTACH_TO);
             break;
 
         case CardRelation::TransformInto:
             cmd.set_target_card_id(sourceCard->getId());
-            cmd.set_target_mode(Command_CreateToken::TRANSFORM_FROM);
+            cmd.set_target_mode(Command_CreateToken::TRANSFORM_INTO);
             break;
     }
 
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1677,7 +1677,7 @@ void Player::actCreateAllRelatedCards()
                         dbName = cardRelationAll->getName();
                         bool persistent = cardRelationAll->getIsPersistent();
                         for (int i = 0; i < cardRelationAll->getDefaultCount(); ++i) {
-                            createCard(sourceCard, dbName, false, persistent);
+                            createCard(sourceCard, dbName, CardRelation::DoesNotAttach, persistent);
                         }
                         ++tokensTypesCreated;
                         if (tokensTypesCreated == 1) {
@@ -1692,7 +1692,7 @@ void Player::actCreateAllRelatedCards()
                         dbName = cardRelationNotExcluded->getName();
                         bool persistent = cardRelationNotExcluded->getIsPersistent();
                         for (int i = 0; i < cardRelationNotExcluded->getDefaultCount(); ++i) {
-                            createCard(sourceCard, dbName, false, persistent);
+                            createCard(sourceCard, dbName, CardRelation::DoesNotAttach, persistent);
                         }
                         ++tokensTypesCreated;
                         if (tokensTypesCreated == 1) {
@@ -1731,23 +1731,22 @@ bool Player::createRelatedFromRelation(const CardItem *sourceCard, const CardRel
             return false;
         }
         for (int i = 0; i < count; ++i) {
-            createCard(sourceCard, dbName, false, persistent);
+            createCard(sourceCard, dbName, CardRelation::DoesNotAttach, persistent);
         }
     } else if (cardRelation->getDefaultCount() > 1) {
         for (int i = 0; i < cardRelation->getDefaultCount(); ++i) {
-            createCard(sourceCard, dbName, false, persistent);
+            createCard(sourceCard, dbName, CardRelation::DoesNotAttach, persistent);
         }
     } else {
-        if (cardRelation->getDoesAttach()) {
-            createAttachedCard(sourceCard, dbName, persistent);
-        } else {
-            createCard(sourceCard, dbName, false, persistent);
-        }
+        createCard(sourceCard, dbName, cardRelation->getAttachType(), persistent);
     }
     return true;
 }
 
-void Player::createCard(const CardItem *sourceCard, const QString &dbCardName, bool attach, bool persistent)
+void Player::createCard(const CardItem *sourceCard,
+                        const QString &dbCardName,
+                        CardRelation::AttachType attachType,
+                        bool persistent)
 {
     CardInfoPtr cardInfo = db->getCard(dbCardName);
 
@@ -1786,16 +1785,22 @@ void Player::createCard(const CardItem *sourceCard, const QString &dbCardName, b
     cmd.set_x(gridPoint.x());
     cmd.set_y(gridPoint.y());
 
-    if (attach) {
-        cmd.set_target_card_id(sourceCard->getId());
+    switch (attachType) {
+        case CardRelation::DoesNotAttach:
+            break;
+
+        case CardRelation::AttachTo:
+            cmd.set_target_card_id(sourceCard->getId());
+            cmd.set_target_mode(Command_CreateToken::REVERSE_ATTACH);
+            break;
+
+        case CardRelation::TransformInto:
+            cmd.set_target_card_id(sourceCard->getId());
+            cmd.set_target_mode(Command_CreateToken::TRANSFORM_FROM);
+            break;
     }
 
     sendGameCommand(cmd);
-}
-
-void Player::createAttachedCard(const CardItem *sourceCard, const QString &dbCardName, bool persistent)
-{
-    createCard(sourceCard, dbCardName, true, persistent);
 }
 
 void Player::actSayMessage()
@@ -3550,6 +3555,7 @@ void Player::addRelatedCardActions(const CardItem *card, QMenu *cardMenu)
         CardInfoPtr relatedCard = db->getCard(cardRelation->getName());
         if (relatedCard == nullptr)
             continue;
+
         QString relatedCardName;
         if (relatedCard->getPowTough().size() > 0) {
             relatedCardName = relatedCard->getPowTough() + " " + relatedCard->getName(); // "n/n name"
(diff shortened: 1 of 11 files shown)


common/server_card.cpp
```diff
@@ -47,15 +57,25 @@ void Server_Card::resetState()
     setPT(QString());
     setAnnotation(QString());
     setDoesntUntap(false);
+    setFaceDown(false);
```
ebbit1q, line 60: this causes a major bug: cards have their state reset when moved between the battlefield and the deck, their facedown state is then checked afterwards to determine what event to show to other players. with this change moving a facedown card to your deck (unknown to unknown) will tell all your opponents (but not you) what card it was.

common/server_player.cpp
```diff
@@ -485,8 +515,18 @@ Response::ResponseCode Server_Player::moveCard(GameEventStorage &ges,
             event.set_card_id(static_cast<google::protobuf::uint32>(card->getId()));
             ges.enqueueGameEvent(event, playerId);
 
-            card->deleteLater();
-        } else {
+            if (Server_Card *stashedCard = card->takeStashedCard()) {
+                stashedCard->setId(newCardId());
+                ges.enqueueGameEvent(makeCreateTokenEvent(startzone, stashedCard, card->getX(), card->getY()),
+                                     playerId);
+                card = stashedCard;
```
ebbit1q, line 522: I don't see card being deleted before this, this will probably leak memory

common/server_player.cpp
```diff
@@ -1336,33 +1398,89 @@ Server_Player::cmdCreateToken(const Command_CreateToken &cmd, ResponseContainer
...
+                card->setParentCard(parentCard);
+
+                ges.enqueueGameEvent(makeAttachCardEvent(card, parentCard), playerId);
+            }
 
-    return cmdAttachCard(cmd2, rc, ges);
+            // Copy attachments
+            while (!targetCard->getAttachedCards().isEmpty()) {
+                Server_Card *attachedCard = targetCard->getAttachedCards().last();
+                targetCard->removeAttachedCard(attachedCard);
```
ebbit1q, line 1471: the reparenting of attached cards seems incomplete, the attached cards are never given to the newly created transformed card, leaking memory and allowing the card to be attached while it should have attached cards already

ebbit1q (changes requested):
I like how this approach seems to work really well even if playing together with older clients.

one issue I notice is that when destroying the created transformed token by moving it to a new zone it is never checked if the original card wasn't a token, this allows you to move transformable tokens anywhere. now in magic this doesn't actually ever come up, back sides of cards aren't copyable by clone effects so you'd never end up with a token of a card that you transform, yet cockatrice does allow this and we simply shouldn't allow people to shuffle tokens into their deck this way by accident.

the events that are generated are also a bit unclear to clients, instead of the card being destroyed the event should contain information of what is really happening with the card, simply adding a flag to the event when it's being transformed that says so would make this clear if only in the log. same for reverting a transform, which currently creates three separate events. adding a flag to the token event would allow us to create an indicator on the card that it can transform back, and perhaps it should have an action to do so (simply destroying the card).

ebbit1q:
while I like this approach, and how it looks cleaner for old client versions, I'd like to pitch my idea for having the effect be purely expressed client side again (https://github.com/Cockatrice/Cockatrice/pull/4749#issuecomment-1432241678) this would sidestep the memory management issues on the server side by having the client be responsible for not showing the original transformed card while it is attached like it is in the current behavior. it might be a bit non intuitive though, and it'd incur a lot of the same issues with reparenting cards like here.

ebbit1q:
have you also considered cases where you'd like to transform multiple cards, like the meld mechanic does?
your original pr had the option to store multiple cards in an attached zone theoretically.

cockatrice/src/player.cpp
```diff
@@ -1786,18 +1785,24 @@ void Player::createCard(const CardItem *sourceCard, const QString &dbCardName, b
...
 
-    if (attach) {
-        cmd.set_target_card_id(sourceCard->getId());
+    switch (attachType) {
+        case CardRelation::DoesNotAttach:
+            break;
+
+        case CardRelation::AttachTo:
+            cmd.set_target_card_id(sourceCard->getId());
+            cmd.set_target_mode(Command_CreateToken::REVERSE_ATTACH);
```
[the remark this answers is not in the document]
Elarnon: On one hand I agree, and had it set up this way at some point.
On the other hand when `CardRelation` relates card A to card B with `AttachTo`, card A should be attached to card B, but when we create token T with target card C, we don't attach T to C but C to T, hence the reversed names.

Nothing I can't solve with some precise documentation I guess (:

common/server_card.cpp
Elarnon: Oh, right. I think I changed that (or intended to anyways) in my refactored implementation of `::moveCard` in the other PR, but not here. Will revert

cockatrice/src/player.cpp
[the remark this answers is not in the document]
Elarnon: On one hand I agree, on the other hand CardRelation goes from card A to card B while for CreateToken the natural reading of the relation is from token to card… I can keep the CardRelation names and add some explicit documentation in CreateToken

common/server_card.cpp
Elarnon: Right, I believe I changed this in the other PR but not there… Reverting

common/server_player.cpp
Elarnon: `card->deleteLater()` did get incorrectly yanked from this branch.

common/server_player.cpp
Elarnon: Right, this should be `attacheCard->setParent(card)`

Elarnon:
> one issue I notice is that when destroying the created transformed token by moving it to a new zone it is never checked if the original card wasn't a token [..]

Right, that used to be dealt with automatically in the original PR because the token was moved to the attached zone and got destroyed, but it's not in this version. Will fix.

> the events that are generated are also a bit unclear to clients, instead of the card being destroyed the event should contain information of what is really happening with the card, simply adding a flag to the event when it's being transformed that says so would make this clear if only in the log. same for reverting a transform, which currently creates three separate events. adding a flag to the token event would allow us to create an indicator on the card that it can transform back, and perhaps it should have an action to do so (simply destroying the card).

The issue here is that if we want to have it working with other clients we have to generate the multiple events. We can add a flag for token creation + an appropriate CardTransform context for reverting. Alternatively, in my original PR I had a feature to allow sending different events depending on the features supported by the client (rather, all events are sent and the client picks the most precise it know how to handle). I can bring that back and use it here if you're interested.

Agreed on having an action for the client as well, my original PR had this but I figured adding an "destroy card" action was maybe too much. Note that currently you can transform again the transformed card to revert the transform, it will generate weird events but functionally it's the same as reverting back to the original card (if I didn't mess up, stashing a card that already has a stashed card will destroy the card and replace it with its stashed card for that purpose).

> while I like this approach, and how it looks cleaner for old client versions, I'd like to pitch my idea for having the effect be purely expressed client side again (https://github.com/Cockatrice/Cockatrice/pull/4749#issuecomment-1432241678) this would sidestep the memory management issues on the server side by having the client be responsible for not showing the original transformed card while it is attached like it is in the current behavior. it might be a bit non intuitive though, and it'd incur a lot of the same issues with reparenting cards like here.

I believe I already gave my thoughts on this approach in another thread, I am not a fan of the idea mostly because it breaks [t:aura is:dfc](https://scryfall.com/search?q=is%3Adfc+t%3Aaura) in a non-intuitive way.

> have you also considered cases where you'd like to transform multiple cards, like the meld mechanic does? your original pr had the option to store multiple cards in an attached zone theoretically.

My goal with this PR was to have the simplest implementation of "stashing a card" working and implemented. Selfishly, I use Transform cards but not Meld cards (and I suppose that's more frequent than the opposite), so I went with the simplest design that could get the feature in quickly, which meant leaving Meld out mostly due to UI qualms.

For Meld, my plan would depend on whether dynamic/attached-to-card zones get in. If they do, we can switch to something closer to my original implementation using a zone to store the stashed cards, and if we expose the zone to users, we get (basic) Meld support for free. If they don't, or they get in w/o the "attach-zone-to-card" feature, transform can be revisited to allow stashing multiple cards, at which point the UI details can be figured out. Does that sound reasonable?

ebbit1q:
With further testing I found a new issue: many cards are able to transform back and forth, these cards will have the transform card relation on both card sides in the database. when you transform a card while it already is transformed it'll delete the transformed token and replace it with a new token, this is fine and I wouldn't bother putting a code path in place to detect this as it still works the way the user would expect. the problem is that in this case the card's state is improperly transferred, counters disappear and attachments will fall off, getting them placed in the same coordinate unfortunately.

I have also deployed this pr to beta.cockatrice.us in order to facilitate testing.

Also, marking a card relation as both persistent and transform will result in the card permanently replacing it, which I don't consider counterintuitive but is something to note.

Elarnon:
> With further testing I found a new issue: many cards are able to transform back and forth, these cards will have the transform card relation on both card sides in the database. when you transform a card while it already is transformed it'll delete the transformed token and replace it with a new token, this is fine and I wouldn't bother putting a code path in place to detect this as it still works the way the user would expect. the problem is that in this case the card's state is improperly transferred, counters disappear and attachments will fall off, getting them placed in the same coordinate unfortunately.

Oh well of course, I delete the current card before copying its state…That was introduced to fix the "transformed tokens don't disappear when they change zone" bug. Changed it to check in `setStashedCard` that the stashed card is not a token. Just pushed a fix

> Also, marking a card relation as both persistent and transform will result in the card permanently replacing it, which I don't consider counterintuitive but is something to note.

I would say that it's correct? "Persistent transform" sounds like the transform should never end. If it ends up causing issues we can make the "Revert" command but there are subtle issues there I'd rather not touch before we decide on what we do with the dynamic zones proposal, as that would largely influence the implementation.

ebbit1q (approved):
I think this is ready, I'll update the beta server when this is merged.

Issue #4112, opened 2020-09-28
PictureLoader Sequential


ZeldaZach:
PictureLoader (https://github.com/Cockatrice/Cockatrice/blob/44297dcd1cdf041f9633df8bcfbb6cf0c51c4a6d/cockatrice/src/pictureloader.cpp#L129) is sequential. Should look into parallelizing it.

ebbit1q:
fixed in #5977 and followup tickets

Issue #3729, opened 2019-05-24
Slow loading card images for very large decks


MotherofRuins:
Client Version: 2.7.0 (2019-03-04)
_______________________________________________________________________________________

I often play with very large decks (960 cards +/-).  I have hi-res images in CUSTOM on local SSD.  Picture loader only loads about 1 per second, so it takes many minutes to find and load all images (this occurred to a lesser extent with the lo-res default images).  Request that image loader prioritize loading images in visible game zones.  This would make a very noticeable difference in performance.

Also, sometimes PictureLoader finds the card on disk, other times it seems to look online, even when the file exists in CUSTOM, as with New Horizons from the log below:

PictureLoader: [card:  "Pillory of the Sleepless"  set:  "A25" ]: Trying to load picture
PictureLoader: [card:  "Pillory of the Sleepless"  set:  "A25" ]: Picture found on disk.
PictureLoader: [card:  "New Horizons"  set:  "WAR" ]: Requested set property ( "muid" ) for Url template ( "https://api.scryfall.com/cards/multiverse/!set:muid!?format=image" ) is not available
PictureLoader: [card:  "New Horizons"  set:  "WAR" ]: Requested set property ( "muid" ) for Url template ( "http://gatherer.wizards.com/Handlers/Image.ashx?multiverseid=!set:muid!&type=card" ) is not available

ctrlaltca:
> Request that image loader prioritize loading images in visible game zones. This would make a very noticeable difference in performance.

When a deck is loaded in the lobby, all the card images are precached.
How much is an image is MB? Ensure you have enough memory for the whole deck images to be used into Settings -> Picture cache size, to avoid Cockatrice from continuously loading an image, dropping another from the memory, then reloading it again.

> Also, sometimes PictureLoader finds the card on disk, other times it seems to look online, even when the file exists in CUSTOM, as with New Horizons from the log below:

That log is not enough to help you, sorry. It's missing all the initial steps where Cockatrice is trying to load the image from the disk.
Try opening a fresh cockatrice, search for the specific card on the deck editor and click on it, then get the log.

MotherofRuins:
I have 16GB RAM.  Image cache is maxed out at 2047MB.  There are 6068 hi-res images in CUSTOM, ranging in size from 85 to 372 K with an average of 224 KB.  Total on disk is 1.36 GB.  Total time to copy the entire folder onto the same disk is 20 seconds, or 0.0033 seconds per image file.

[...]

PR #4756, opened 2023-02-18
PictureLoader: Replace downloadedPics cache with QNetworkCache

Elarnon:

Currently when the "Download card pictures on the fly" option is enabled, Cockatrice stores downloaded pictures into a downloadedPics sub-folder, keyed on set and card name. If a picture is found in that folder, we never try to download a picture for that card ever again (until it is reprinted in a more recent set, I guess).

This has the unfortunate consequence that if you change the URLs for downloading card images, the changes are not applied to cards that already have their picture downloaded. In particular, if you use localized card pictures (through !sflang!), you get a mix of cards in different languages depending on the currently configured language at the time each card was downloaded.

This patch removes that mechanism in favor of setting a QNetworkDiskCache on the QNetworkAccessManager used by the PictureLoader to download pictures. The QNetworkDiskCache caches the picture keyed on their URL: if the URL changes, a new request will be made. In particular, if you use picture URLs with !sflang! and change the language, pictures for the current language will be downloaded even for cards that already have a picture. The QNetworkDiskCache is configured with a maximum size of 4GB, which should be enough to hold one high-quality JPEG for each M:TG card in existence.

Note that this does not affect the existing mechanism for defining custom card art, either through the CUSTOM directory or the set-based one.  Cockatrice will still read existing cards in the downloadedPics directory as before, it will just no longer write into that directory (since pictures are cached by the QNetworkDiskCache instead). To fully switch to the new cache, users should use the "Delete Downloaded Images" button in the settings: it will clear the QNetworkDiskCache but also remove the downloadedPics directory.

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -621,8 +621,8 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     connect(&mcDownloadSpoilersCheckBox, SIGNAL(toggled(bool)), &SettingsCache::instance(),
             SLOT(setDownloadSpoilerStatus(bool)));
     connect(&mcDownloadSpoilersCheckBox, SIGNAL(toggled(bool)), this, SLOT(setSpoilersEnabled(bool)));
-    connect(networkCacheEdit, &QSpinBox::valueChanged, &SettingsCache::instance(),
-            &SettingsCache::setNetworkCacheSizeInMB);
+    connect(networkCacheEdit, SIGNAL(valueChanged(int)), &SettingsCache::instance(),
+            SLOT(setNetworkCacheSizeInMB(int)));
 
     mpGeneralGroupBox = new QGroupBox;
     mpGeneralGroupBox->setLayout(lpGeneralGrid);
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -3,6 +3,7 @@
 #include "carddatabase.h"
 #include "gettextwithmax.h"
 #include "main.h"
+#include "pictureloader.h"
 #include "releasechannel.h"
 #include "sequenceEdit/sequenceedit.h"
 #include "settingscache.h"
@@ -635,6 +636,11 @@ void DeckEditorSettingsPage::resetDownloadedURLsButtonClicked()
 
 void DeckEditorSettingsPage::clearDownloadedPicsButtonClicked()
 {
+    PictureLoader::clearNetworkCache();
+
+    // These are not used anymore, but we don't delete them automatically, so
+    // we should do it here lest we leave pictures hanging around on users'
+    // machines.
     QString picsPath = SettingsCache::instance().getPicsPath() + "/downloadedPics/";
     QStringList dirs = QDir(picsPath).entryList(QDir::AllDirs | QDir::NoDotAndDotDot);
     bool outerSuccessRemove = true;
@@ -662,6 +668,7 @@ void DeckEditorSettingsPage::clearDownloadedPicsButtonClicked()
     }
     if (outerSuccessRemove) {
         QMessageBox::information(this, tr("Success"), tr("Downloaded card pictures have been reset."));
+        QDir(SettingsCache::instance().getPicsPath()).rmdir("downloadedPics");
     } else {
         QMessageBox::critical(this, tr("Error"), tr("One or more downloaded card pictures could not be cleared."));
     }
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -599,14 +599,18 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     networkCacheEdit->setValue(SettingsCache::instance().getNetworkCacheSizeInMB());
     networkCacheEdit->setSuffix(tr(" MB"));
 
+    auto networkCacheLayout = new QHBoxLayout;
+    networkCacheLayout->addWidget(&networkCacheLabel);
+    networkCacheLayout->addWidget(networkCacheEdit);
+    networkCacheLayout->addWidget(&clearDownloadedPicsButton);
+    networkCacheLayout->addStretch();
+
     // Top Layout
     lpGeneralGrid->addWidget(&picDownloadCheckBox, 0, 0);
     lpGeneralGrid->addWidget(&resetDownloadURLs, 0, 1);
     lpGeneralGrid->addLayout(messageListLayout, 1, 0, 1, 2);
-    lpGeneralGrid->addWidget(&networkCacheLabel, 2, 0);
-    lpGeneralGrid->addWidget(networkCacheEdit, 2, 1);
+    lpGeneralGrid->addLayout(networkCacheLayout, 2, 0, 1, 2);
     lpGeneralGrid->addWidget(&urlLinkLabel, 3, 0);
-    lpGeneralGrid->addWidget(&clearDownloadedPicsButton, 3, 1);
 
     // Spoiler Layout
     lpSpoilerGrid->addWidget(&mcDownloadSpoilersCheckBox, 0, 0);
@@ -623,6 +627,7 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     connect(&mcDownloadSpoilersCheckBox, SIGNAL(toggled(bool)), this, SLOT(setSpoilersEnabled(bool)));
     connect(networkCacheEdit, SIGNAL(valueChanged(int)), &SettingsCache::instance(),
             SLOT(setNetworkCacheSizeInMB(int)));
+    connect(&SettingsCache::instance(), SIGNAL(networkCacheSizeChanged(int)), networkCacheEdit, SLOT(setValue(int)));
 
     mpGeneralGroupBox = new QGroupBox;
     mpGeneralGroupBox->setLayout(lpGeneralGrid);
@@ -801,9 +806,9 @@ void DeckEditorSettingsPage::retranslateUi()
                                 tr("Do not close settings until manual update is complete"));
     picDownloadCheckBox.setText(tr("Download card pictures on the fly"));
     urlLinkLabel.setText(QString("<a href='%1'>%2</a>").arg(WIKI_CUSTOM_PIC_URL).arg(tr("How to add a custom URL")));
-    clearDownloadedPicsButton.setText(tr("Delete Downloaded Images"));
+    clearDownloadedPicsButton.setText(tr("Clear", "Delete Downloaded Images"));
     resetDownloadURLs.setText(tr("Reset Download URLs"));
-    networkCacheLabel.setText(tr("Maximum image cache size:"));
+    networkCacheLabel.setText(tr("Size of the downloaded images directory:"));
 }
 
 MessagesSettingsPage::MessagesSettingsPage()
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -69,18 +69,9 @@ GeneralSettingsPage::GeneralSettingsPage()
     updateNotificationCheckBox.setChecked(settings.getNotifyAboutUpdates());
     newVersionOracleCheckBox.setChecked(settings.getNotifyAboutNewVersion());
 
-    // pixmap cache
-    pixmapCacheEdit.setMinimum(PIXMAPCACHE_SIZE_MIN);
-    // 2047 is the max value to avoid overflowing of QPixmapCache::setCacheLimit(int size)
-    pixmapCacheEdit.setMaximum(PIXMAPCACHE_SIZE_MAX);
-    pixmapCacheEdit.setSingleStep(64);
-    pixmapCacheEdit.setValue(settings.getPixmapCacheSize());
-    pixmapCacheEdit.setSuffix(" MB");
-
     showTipsOnStartup.setChecked(settings.getShowTipsOnStartup());
 
     connect(&languageBox, SIGNAL(currentIndexChanged(int)), this, SLOT(languageBoxChanged(int)));
-    connect(&pixmapCacheEdit, SIGNAL(valueChanged(int)), &settings, SLOT(setPixmapCacheSize(int)));
     connect(&updateReleaseChannelBox, SIGNAL(currentIndexChanged(int)), &settings, SLOT(setUpdateReleaseChannel(int)));
     connect(&updateNotificationCheckBox, SIGNAL(stateChanged(int)), &settings, SLOT(setNotifyAboutUpdate(int)));
     connect(&newVersionOracleCheckBox, SIGNAL(stateChanged(int)), &settings, SLOT(setNotifyAboutNewVersion(int)));
@@ -91,8 +82,6 @@ GeneralSettingsPage::GeneralSettingsPage()
     personalGrid->addWidget(&languageBox, 0, 1);
     personalGrid->addWidget(&updateReleaseChannelLabel, 1, 0);
     personalGrid->addWidget(&updateReleaseChannelBox, 1, 1);
-    personalGrid->addWidget(&pixmapCacheLabel, 2, 0);
-    personalGrid->addWidget(&pixmapCacheEdit, 2, 1);
     personalGrid->addWidget(&updateNotificationCheckBox, 3, 0, 1, 2);
     personalGrid->addWidget(&newVersionOracleCheckBox, 4, 0, 1, 2);
     personalGrid->addWidget(&showTipsOnStartup, 5, 0, 1, 2);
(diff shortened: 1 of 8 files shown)


ebbit1q:
I absolutely love this, it fixes a main issue with the image cache system, I have two concerns though:
- what about cached entries that become unused? one of the main issues that seems to remain here is that when a new reprint releases the client will download a new version of that card while the old version remains on disk, perhaps cleaning up old entries periodically would be a good solution?
- 4GB seems like a nice default for the size of the cache, though it might not be for everyone, I think having this parameter be available in the settings would be better.

considering neither of these options are currently in the client we can consider them out of scope but we should ticket them for later.

ebbit1q (changes requested):
this looks about ready, I tested performance and it seems to be working quite nicely.

the cache location is outside of the main data location, meaning it's shared across portable installations on one machine but isn't portable itself. this is both an upside and a downside as this means you won't be downloading multiple copies of cards sometimes but transferring cached images is now a manual process instead of everything being contained inside one folder, you'd now have to copy your cache to the other machine.

however during my performance testing I noticed something annoying: closing the client during a cache saving operation breaks that entry somehow and whenever the client tries to retrieve it, it will instead stop loading images altogether.

this is compounded by the fact that currently the downloader is singlethreaded, it intentionally waits for each download in turn and doesn't have a timeout or anything.

to address this I propose a timeout for the downloader when getting an image from the cache, after which it will purge that specific entry from the cache and add the to be downloaded card back to the end of the queue.

ebbit1q:
I tried to find if the image queue parallelization is ticketed but couldn't find anything, so here is some further detail:

one of the reasons this is done (or simply hasn't changed since ~2011) is as a lazy rate limiting function, but while this does sort of rate limit the amount of requests to download sources if you have a low latency connection you can certainly go over scryfall's 10 images per second guideline. (I can't test this myself but I expect this to currently be a source of problems for some users)

I'd say fixing this is out of scope for this pr, though I've been planning to address this at some point, ping me on discord to discuss this.

Elarnon:
> * what about cached entries that become unused? one of the main issues that seems to remain here is that when a new reprint releases the client will download a new version of that card while the old version remains on disk, perhaps cleaning up old entries periodically would be a good solution?

Agreed that this would be nice. Currently old pictures are only deleted once the cache size limit is reached; oldest URLs are deleted first. We can customize the behavior to start by using last access time instead by writing extending `QNetworkDiskCache`, that would make more sense for us. As you say, should be done in a different PR.

> * 4GB seems like a nice default for the size of the cache, though it might not be for everyone, I think having this parameter be available in the settings would be better.

Agreed on this as well — I'm happy to add a settings for this, even intended to, but something came up and it slipped through my mind.

> the cache location is outside of the main data location, meaning it's shared across portable installations on one machine but isn't portable itself. this is both an upside and a downside as this means you won't be downloading multiple copies of cards sometimes but transferring cached images is now a manual process instead of everything being contained inside one folder, you'd now have to copy your cache to the other machine.

I was not aware that Cockatrice had portable installs. I'm happy to set the cache location to `qApp->applicationDirPath() + "/cache"` for portable locations.

> however during my performance testing I noticed something annoying: closing the client during a cache saving operation breaks that entry somehow and whenever the client tries to retrieve it, it will instead stop loading images altogether.

Oh, that's not good. I did not encounter this in my own testing. Just to clarify, it's not loading *any* image, not just that specific image, and it's not fixed by restarting the client? Do you happen to have a debug log for this?

> to address this I propose a timeout for the downloader when getting an image from the cache, after which it will purge that specific entry from the cache and add the to be downloaded card back to the end of the queue.

I'd rather figure out why the cache hangs… it shouldn't be doing that at all. Are you sure it's specifically due to the cache?
Now that I think about it, that's a behavior I have seen sporadically with downloadedPics, sometimes it stops downloading altogether but it works again after restarting the client; I suspected either a bug in the PictureLoader or a very slow request.

> I tried to find if the image queue parallelization is ticketed but couldn't find anything, so here is some further detail:

I've seen #4112 and there is also https://github.com/Cockatrice/Cockatrice/issues/3729#issuecomment-503381587

> one of the reasons this is done (or simply hasn't changed since ~2011) is as a lazy rate limiting function

We should implement an actual rate limit then.

> you can certainly go over scryfall's 10 images per second guideline

I believe the 10 requests per second only applies to `api.scryfall.com`, not to the actual image files — but I guess we do go through the API server to resolve the image URL. Maybe we could have Oracle store the actual URLs (on `*.scryfall.io`) in the DB, so that we can get the image directly without using `api.scryfall.com` at all?
Still, having an actual rate limit would be nice.

> I'd say fixing this is out of scope for this pr, though I've been planning to address this at some point, ping me on discord to discuss this.

Happy to discuss on Discord, will ping you when I have some time. I actually wrote a rough patch to make the PictureLoader parallel using `QtConcurrent` at the same time as this PR but dropped it because I was not satisfied with the resulting code.

ebbit1q:
that's definitely a ci issue, one that I thought I fixed

cockatrice/src/pictureloader.cpp
```diff
@@ -114,8 +115,23 @@ PictureLoaderWorker::PictureLoaderWorker()
...
+            &PictureLoaderWorker::networkCacheSizeChanged);
 
     networkManager = new QNetworkAccessManager(this);
+    // We need a timeout to ensure requests don't hang indefinitely in case of
+    // cache corruption, see related Qt bug: https://bugreports.qt.io/browse/QTBUG-111397
+    // Use Qt's default timeout (30s, as of 2023-02-22)
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 15, 0))
+    networkManager->setTransferTimeout();
+#endif
+    auto cache = new QNetworkDiskCache(this);
```
ebbit1q, line 128: why not assign the cache as a member?

cockatrice/src/pictureloader.cpp
```diff
@@ -385,11 +384,11 @@ void PictureLoaderWorker::startNextPicDownload()
         picDownloadFailed();
     } else {
         QUrl url(picUrl);
-        QNetworkRequest req(url);
         qDebug() << "PictureLoader: [card: " << cardBeingDownloaded.getCard()->getCorrectedName()
                  << " set: " << cardBeingDownloaded.getSetName()
-                 << "]: Trying to download picture from url:" << url.toDisplayString();
-        networkManager->get(req);
+                 << QString("]: Trying to download picture from %1:").arg(picDownload ? "url" : "cache")
```
ebbit1q, line 389: this is always url, I don't think it's able to know if it's in the cache at this point

cockatrice/src/pictureloader.cpp
```diff
@@ -420,20 +419,42 @@ bool PictureLoaderWorker::imageIsBlackListed(const QByteArray &picData)
...
+    if (!picDownload) {
+        req.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::AlwaysCache);
+    }
+    return networkManager->get(req);
+}
+
 void PictureLoaderWorker::picDownloadFinished(QNetworkReply *reply)
 {
     if (reply->error()) {
         qDebug() << "PictureLoader: [card: " << cardBeingDownloaded.getCard()->getName()
```
ebbit1q, line 434: having it say a download failed while it failed to get it from a cache makes no sense, I'd move this below the if statement there

cockatrice/src/dlg_settings.cpp
```diff
@@ -785,6 +803,7 @@ void DeckEditorSettingsPage::retranslateUi()
     urlLinkLabel.setText(QString("<a href='%1'>%2</a>").arg(WIKI_CUSTOM_PIC_URL).arg(tr("How to add a custom URL")));
     clearDownloadedPicsButton.setText(tr("Delete Downloaded Images"));
     resetDownloadURLs.setText(tr("Reset Download URLs"));
+    networkCacheLabel.setText(tr("Maximum image cache size:"));
```
ebbit1q, line 806: This label is too easily confused with the internal pixmap cache setting (on the general page see `pixmapCacheLabel`), in order to make the difference clearer I think it'd be best to move the two next to each other, either both on general or on the download sources page and have this one say it's on disk or otherwise make clear that one is persistent while the other is for the running application.

ebbit1q:
I think the logging in general can be a bit clearer here, logging is essential for us to help users that are having issues with their pictures not showing up (the most common issue on cockatrice probably: 4735 4725 4634 4545 is just the ones I immediately found) the more detail we can get in the logging the better, I'd like it to include if the download was from cache and if the timeout was the reason it failed.

cockatrice/src/pictureloader.cpp
Elarnon: Didn't seem necessary, it's accessible through `networkManager->cache()` already. Do you want me to make it a member?

cockatrice/src/pictureloader.cpp
```diff
@@ -384,10 +384,11 @@ void PictureLoaderWorker::startNextPicDownload()
         picDownloadFailed();
     } else {
         QUrl url(picUrl);
-        qDebug() << "PictureLoader: [card: " << cardBeingDownloaded.getCard()->getCorrectedName()
-                 << " set: " << cardBeingDownloaded.getSetName()
-                 << QString("]: Trying to download picture from %1:").arg(picDownload ? "url" : "cache")
-                 << url.toDisplayString();
+        qDebug() << qPrintable(QString("PictureLoader: [card: %1 set: %2]: Trying to %3: %4")
```
thopp-tudelft, line 387: what's the rationale behind using qPrintable here? converting the qstring to a char * isn't necessary for qDebug

Elarnon: Printing `QString` directly adds quotes and escaping when printing, which I find quite ugly especially since I was using `.arg()`, so I figured I'd remove while I was there. I don't really care either way (and apparently the right thing to use would be qUtf8Printable and not qPrintable). Reverting…

ebbit1q: chaining into qDebug() by default adds spaces around each argument, use qDebug().nospace() in order to not do that.

ebbit1q: similarly noquote() will not add quotes

cockatrice/src/pictureloader.cpp
ebbit1q: it just seems simpler than using `networkManager->cache()` to me? it's just a pointer

cockatrice/src/dlg_settings.cpp
Elarnon: Changed the label to mention "image directory" (should make it clear it's on disk), moved pixmap cache setting to "card source" page, and added tooltips to the QSpinBoxes to clarify further the purpose of both settings.

cockatrice/src/pictureloader.cpp
ebbit1q: the networkmanager does take ownership of the cache, so we could be left with a dangling pointer, in practice we assign the network manager as a child here and never delete it otherwise so I'd say that isn't a concern either.

Elarnon: I got rid of the setNetworkCacheSize on the worker and connected the settings to `QNetworkDiskCache::setMaximumCacheSize` directly so we don't need it outside of the constructor now, seems cleaner anyways

Elarnon:
@ebbit1q I think I addressed all your comments - ready for round 2 :)

cockatrice/src/pictureloader.cpp
```diff
@@ -384,11 +384,9 @@ void PictureLoaderWorker::startNextPicDownload()
         picDownloadFailed();
     } else {
         QUrl url(picUrl);
-        qDebug() << qPrintable(QString("PictureLoader: [card: %1 set: %2]: Trying to %3: %4")
-                                   .arg(cardBeingDownloaded.getCard()->getCorrectedName())
-                                   .arg(cardBeingDownloaded.getSetName())
-                                   .arg(picDownload ? "download picture from url" : "load picture from cache")
-                                   .arg(url.toDisplayString()));
+        qDebug() << "PictureLoader: [card: " << cardBeingDownloaded.getCard()->getCorrectedName()
+                 << " set: " << cardBeingDownloaded.getSetName()
```
ebbit1q, line 388: `]: Trying to ` seems to have disappeared?

ebbit1q:
qDebug() will by default add spaces between all its inputs, either remove the spaces in the joining strings or use qDebug().nospace()

cockatrice/src/pictureloader.cpp
Elarnon: Indeed, put it back in

skwerlman:
how does this affect the offline brewing usecase? i know of at least one user who downloads all the card images manually by scrolling down the list in the deck editor so images are cached for each card so that they can use trice when they have poor or no internet connection
i also have written a tool which is able to download all card images ahead of time to remove the need to scroll down the list. this ofc depends on the on-disk format of the cache. will this still be possible after this change?

Elarnon:
> how does this affect the offline brewing usecase? i know of at least one user who downloads all the card images manually by scrolling down the list in the deck editor so images are cached for each card so that they can use trice when they have poor or no internet connection

This will still work. Cockatrice still looks for images in the old cache directory, it just does not write new images there anymore, so people that already have a bunch of images preloaded this way won't lose them. OTOH if you do this today, the images will be stored in the new cache directory, but it'll otherwise work roughly the same.

The main difference is that the new cache is limited in size to a default of 4GB (it's a setting), after which it will start removing old images to make space for the new ones. We have plans to improve the logic for selecting which cards get removed (see https://github.com/Cockatrice/Cockatrice/pull/4756#issuecomment-1436020863 ).

> i also have written a tool which is able to download all card images ahead of time to remove the need to scroll down the list. this ofc depends on the on-disk format of the cache. will this still be possible after this change?

Not only will this still be possible but your script should still work as-is. With this change, Cockatrice still looks for card imagery in all the places it used to, it just uses the new cache if the image is not found in any of the existing places. 

Note that if your script writes to the `downloadedPics` directory and not the toplevel `pics` directory, the `downloadedPics` directory will get cleared along with the new cache when the user clicks the "Delete Downloaded Images" button (as it did before).

ebbit1q:
the only real downside I see here is the small amount of extra io required, which we plan to mitigate entirely by allowing parallel downloads

ebbit1q:
as an added bonus, I think not writing a bunch of images to disk will stop cockatrice from getting flagged by antiviruses (though I hope people have stopped using avast :P )

ebbit1q (approved):
I've changed the debug messages a bit, I think this is ready for merge now.

ebbit1q:
we have the set manager for this, you can disable sets you do not want to download arts from.

of course you can readd this as an option yourself, we've decided there is no reason to keep it around in this pr though.

skwerlman:
> Also tokens are almost always not grabbed correctly. I have it set to grab the large pngs of scryfall, but it always grabs the lower jpgs instead, despite usually grabbing the higher pngs for the normal cards.

Tokens use hardcoded URLs in the tokens.xml which means they won't respect your art url settings, only the set ordering.

Elarnon:
> I use the platform as @skwerlman mentioned above by caching everything locally to make sure I am getting the images I want. The problem with modern sets and the card xml is that often times random frames/versions of cards get pulled. I prefer just the normal frames, but I'll end up with borderless, special treatments, etc.

I am not sure I understand exactly what is the difference in behavior you have between 2.8 and 2.9 — this change should *not* impact the way Cockatrice decides to download images; if it does, it is a bug. Is the main issue that "often times random frames/versions of cards get pulled"? I don't remember the code in detail but it is possible that the order of card versions within a set are not deterministic, which would explain the behavior and have been revealed by this change.

ebbit1q:
that behavior is the same between versions, the difference is that they can't go to the folder the image is in to replace it, a usecase we don't intend and instead have the custom image folder for

Elarnon:
Ah, gotcha — no possibility of manual override. To be fair, it *would* be nice to be able to choose a desired image for a given card instead of letting Cockatrice decide what's best! I believe I actually intended to tackle that after this PR at some point but life had other plans.

ebbit1q:
cockatrice has the option to choose an image, it's what we have the custom image folder for, you place the card image there, copy paste the card name and done.

ebbit1q:
> How much of a hassle would it be to make this a toggle? Defaulting to on is fine. I use the platform as @skwerlman mentioned above by caching everything locally to make sure I am getting the images I want. The problem with modern sets and the card xml is that often times random frames/versions of cards get pulled. I prefer just the normal frames, but I'll end up with borderless, special treatments, etc. Also tokens are almost always not grabbed correctly. I have it set to grab the large pngs of scryfall, but it always grabs the lower jpgs instead, despite usually grabbing the higher pngs for the normal cards.
> 
> I could just manually grab all the cards off scryfall, but formatting the names to what trice expects (especially the tokens with all the spaces) will be a PITA.

I've made a pr for you that should make this easier: https://github.com/Cockatrice/Cockatrice/pull/4931

ebbit1q:
@jaitken11 please check this guide on how to test the version in the pr https://github.com/Cockatrice/Cockatrice/wiki/How-to-use-unreleased-versions-created-by-pull-requests

PR #4765, opened 2023-03-04
Make cards rounded

Elarnon:

Magic cards have rounded corners, and playing cards tend to have rounded corners as well, but Cockatrice currently displays rectangular cards.

This can cause visual glitches when using image scans where the border does not extend in the corner, and for this reason Cockatrice always draws a (rectangular) border around the card to try and make it look a bit better.

In this patch I take a different approach: rather than try to make rounded pegs, er, cards, go into a square hole, the hole is now rounded. More precisely, the AbstractCardItem now has a rounded rectangular shape (with a corner of 5% of the width of the card, identical to that of modern M:TG physical cards).

As a side effect, the card drawing gets a bit simplified by getting rid of transformPainter() when drawing the card outline and using the QPainter::drawPixmap overloads that takes a target QRectF instead.  This means we no longer have to bother about card rotation when painting since that's taken care of by the Graphics View framework (which transformPainter() undoes).

As another side effects, full-art cards look cooler now.

## Screenshots
| Old | New |
| ------|--------|
| Old |New |

### DIFF
--- a/cockatrice/src/abstractcarddragitem.cpp
+++ b/cockatrice/src/abstractcarddragitem.cpp
@@ -45,6 +45,13 @@ AbstractCardDragItem::~AbstractCardDragItem()
         delete childDrags[i];
 }
 
+QPainterPath AbstractCardDragItem::shape() const
+{
+    QPainterPath shape;
+    shape.addRoundedRect(boundingRect(), 0.05 * CARD_WIDTH, 0.05 * CARD_WIDTH);
+    return shape;
+}
+
 void AbstractCardDragItem::paint(QPainter *painter, const QStyleOptionGraphicsItem *option, QWidget *widget)
 {
     item->paint(painter, option, widget);
--- a/cockatrice/src/abstractcarddragitem.cpp
+++ b/cockatrice/src/abstractcarddragitem.cpp
@@ -50,7 +50,7 @@ void AbstractCardDragItem::paint(QPainter *painter, const QStyleOptionGraphicsIt
     item->paint(painter, option, widget);
 
     // adds a mask to the card so it looks like the card hasnt been placed yet
-    painter->fillRect(boundingRect(), GHOST_MASK);
+    painter->fillPath(shape(), GHOST_MASK);
 }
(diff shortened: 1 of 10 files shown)


ebbit1q (approved):
this looks really neat, I can't imagine anyone would want to keep the old behavior. If anyone does they'll have to make a ticket requesting the corner rounding to be configurable.

ebbit1q:
I'll merge this when the format action is happy

Elarnon:
Format should be happy — I also added the rounded corners to the PileZone (library, graveyard, exile…) for consistency.

tooomm:
This looks awesome @Elarnon!

Two questions:
- Will this also change the presentation in the card preview/info widgets and the view in the deck editor?
- Is there a reason the rounded corners are different for the zones (deck pile etc.) than the actual cards e.g. in play? It appears like that in your screenshots.

Elarnon:
> This looks awesome @Elarnon!

Thanks :)

>     * Will this also change the presentation in the card preview/info widgets and the view in the deck editor?

No to the first one, yes to the second.
For card preview, I can add the clipPath to the CardInfoPicture.
For deck editor, it looks a bit weird upon close inspection due to the large borders (but I think those large borders already look weird anyways).

>     * Is there a reason the rounded corners are different for the zones (deck pile etc.) than the actual cards e.g. in play? It appears like that in your screenshots.

This is an omission, just fixed it. Here are updated screenshots:
DeckView
PileZone

tooomm:
Thanks for also including card info!

- before
image

- after
image

ebbit1q (approved):
image
alpha cards had different corners, I consider this acceptable though, they looked worse with the old borders imo

Elarnon:
Yeah, I think Alpha cards still look better than before. We could implement the Alpha corner depending on the set (which wouldn't work for CUSTOM pics, mind) but not sure it's worth the hassle.

ZeldaZach:
I like the change overall, even with the weird alpha case. I'm fine with it as-is

PR #4770, opened 2023-03-09
Look bottom of library

### LINKED ISSUE #4710: New library option
I think there should be a “look at the bottom cards of your library“ option upon right clicking on one’s library. This would help a lot if people send the wrong card to the bottom of their library. It would avoid having to reveal  that card by sending it to your graveyard or the battlefield which is the current way.

Tobyclark:
## Related Ticket(s)
- Fixes #4710 

Adds the ability to look at the bottom of your library


ebbit1q:
please reopen your existing pr instead of creating new ones in order to avoid clutter

PR #4771, opened 2023-03-09
Add 'persistent' as valid attribute of related

SlightlyCircuitous:
With the recent addition of 'persistent' as an attribute of related in cards.xml, trying to validate the xml using the current v4 cards.xsd file produces several 'fails to validate' messages all relating to persistent.

-Updates the v4 cards.xsd file to reflect that 'persistent' is now a valid attribute of related.

Please let me know if the cards.xsd file in the carddatabase_v3 folder should be updated similarly.


ebbit1q (approved):
I can't believe we forgot to add this, thanks for noticing

ebbit1q:
the v3 format did not receive this extension, if you wanted to add it it should be added in the code as well

ebbit1q:
linking https://github.com/Cockatrice/Cockatrice/pull/4646

PR #4772, opened 2023-03-10
Add 'persistent' as valid attribute in v3

SlightlyCircuitous:
Updates the v3 xsd file to match changes made in https://github.com/Cockatrice/Cockatrice/pull/4771


ebbit1q (changes requested):
persistent is not an attribute in v3, when it was added it was only added to v4, there is nothing stopping it from being added to v3 however

SlightlyCircuitous:
Oh, I am sorry. I misunderstood your comment on my other PR.

PR #4784, opened 2023-04-12
Add Ubuntu 23.04 Lunar Lobster Build

SlightlyCircuitous:
Ubuntu 23.04 (Lunar Lobster) will be officially released next Thursday (April 20) so users will probably want builds for it soon.

- Adds a dockerfile and a matrix entry for Lunar
- Adds Lunar to the Release template

The names of the packages for Qt6 SVG and websockets development files were changed in Lunar, so the docker file looks slightly different.

I put this together based on the previous PR that added Kinetic so please let me know if I've missed anything.

### DIFF
--- a/.ci/UbuntuLunar/Dockerfile
+++ b/.ci/UbuntuLunar/Dockerfile
@@ -1,0 +1,26 @@
+FROM ubuntu:lunar
+
+RUN apt-get update && \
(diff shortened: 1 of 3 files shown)


ebbit1q (approved):
cool beans

PR #4785, opened 2023-04-14
add fedora 38, scheduled to release next week

ebbit1q:
36 will become deprecated next month


tooomm:
There is always only 1 month overlapping support.
Why not remove 36 right away when adding the new one?

Issue #2256, opened 2016-11-01
Create DMG background image


Daenyth:
DMG installer can have a background image set

https://cmake.org/cmake/help/v3.0/module/CPackDMG.html

Here's an example of a style that would work for us:

![i](https://c2.staticflickr.com/2/1049/3163487641_1400ab02b4.jpg)

ZeldaZach:
@tooomm Wanna check this out?

SlightlyCircuitous:
This seemed like fun so did some work on it. I made a .dmg manually in macOS 10.13 (only mac environment I have access to) as a mock up that seems to work well and can install both Cockatrice and Oracle by dragging. 

This is the layout of my mock up. I tried a few different colors and I'm open to other suggestions. As you can see, the icon label text doesn't show up as well on some of them. I think I like the blue one the best currently. The font is Noto Sans, which matches the website, but I'm happy to change it if there's a different application font or anything.

![magentatrice](https://user-images.githubusercontent.com/71394296/208986138-239ede1f-eb26-429b-9f0b-41492d59e167.png)

![bluatrice](https://user-images.githubusercontent.com/71394296/208986619-ce7682f4-1d7e-42d1-b90d-61afa4268091.png)

![purpatrice](https://user-images.githubusercontent.com/71394296/208986660-692a0af0-fd09-43d6-808a-11369ca37ac2.png)

My next step is to get the building process automated and integrated, but I wanted to check in about the goals first. Do we just need a background picture and a .DS_store file to facilitate building carbon copies of a predefined layout with cmake Drag'n'Drop or do we want a full AppleScript like in the above cmake announcement that can be modified to build any layout .dmg from scratch? From reading the above links, it seems like we can do either, just depends on how much flexibility/future proofing we want in the build. Lemme know either way and I'll keep playing around with it.

ebbit1q:
I know there have been some changes with installing on later mac versions but I wouldn't know if this is affected, as for colors they all seem to have terrible fringing around the smaller text and it's the least bad with blue?
as for the method of automating it, ease of maintainability is most important, if we need an apple os ourselves to configure it that would make adjustments less easy.

SlightlyCircuitous:
I'll take another pass at the colors. I am not sure why that white fringe happens so it's been a little trial and error. Maybe it doesn't like the gradient.

[...]

PR #4787, opened 2023-04-16
CPack DMG Installer with Background Image

SlightlyCircuitous:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/2256

macOS 10.15/11 installer does not have a background image or drag to install functionality like a lot of modern packages for macOS.

- Modifies CPack to build an installer for macOS with a background and drag to install functionality.

See https://github.com/Cockatrice/Cockatrice/issues/2256#issuecomment-1510028562 for detailed explanation of how this works. Essentially adds a background image and AppleScript to cockatrice/resources and modifies the CMakeLists file to point to those additions. The background image and AppleScript can really go anywhere as long as the CMakeLists points to the right place so they can be moved as needed if they are in the wrong place. 

Screenshot of built installer on macOS 11:
Screen Shot 2023-04-15 at 9 01 22 PM
(Yes, I know servatrice is misspelled here, it's fixed in the PR, I just don't have a screenshot of the new build with that fixed.)


ebbit1q:
would it be possible to use a vector based image like svg for the background?

SlightlyCircuitous:
GIMP won't export to svg, but I've changed it to a .eps, which is also vector. I started three separate build actions while editing so sorry about that.

ebbit1q:
while eps can contain vector data it can also contain bitmaps, in this case gimp is only able to make bitmaps, if you're unable to it's fine as well.

SlightlyCircuitous:
I could probably convert it using Inkscape or something similar, but it looks kinda fiddly, especially with multi-color images. I'll keep looking into it, but if its OK as is then maybe that could be a future PR.

SlightlyCircuitous:
Checked the latest build and the background no longer shows up. Doubled checked that the file paths that I changed are correct so unless I missed something, it doesn't seem like .svg is an option.

SlightlyCircuitous:
Trying to make a disk image manually on macOS 11 confirms that the View Options window simply does not accept .svg files for the background. I will revert this all back to the original .tiff.

ebbit1q:
why is the image 3MB? that's rather big for such a simple image

SlightlyCircuitous:
I think .tif files are just bigger because of the way they handle color information per pixel. I can make it a png (250 Kb) or a jpg (60 Kb) if you prefer. I don't think that the added color fidelity (?) is necessary here; I was just using .tif because that's what the cmake applescript example used.

ebbit1q:
tif is a very flexible image format, it does not necessarily need to use 16 bit color, you can even use jpeg compression in tif files.

ebbit1q:
either png or jpg would also suffice. if you think the jpg looks good enough then the small size is preferred. otherwise png would not have any artifacting.

SlightlyCircuitous:
Ah, I see. I had missed GIMP's tif compression options previously. I've made one that is much smaller and it doesn't seem to have any artifacting on close inspection (the one I made using jpg compression did).

ebbit1q:
background is a bit unclear as a filename, could you make it more specific like dmg background?

SlightlyCircuitous:
I think I was overzealous on the rename. As far as I can tell from the documentation, the applescript/cpack expects the path to be .background:background.tif even if the filename is different. Seems like Cpack copies whatever is in CPACK_DMG_BACKGROUND_IMAGE to a new .backgrounds folder and then renames it. We'll see if I am right if the build works.

ebbit1q:
cool, this seems ready, @ZeldaZach can you test it and see if you like the way this looks now?

tooomm:
I like it 👍

Did you also try to have the "Cockatrice" title centered?

SlightlyCircuitous:
Yes, I've tried the title centered and I don't think it looks particularly good with the rest of the arrangement, especially since that would put it right under the top bar with a centered 'Cockatrice'. Both the cmake example and the linked firefox example have it aligned left and I think it's a good look for this kind of thing.

SlightlyCircuitous:
The raw .xcf that I used to create the background image is available [here](https://github.com/SlightlyCircuitous/image-storage) if anyone would like to edit the background in the future.

PR #4790, opened 2023-04-19
Remove macOS 10.15 Catalina Build

SlightlyCircuitous:
As far as I can tell, macOS 10.15 Catalina became unsupported by Apple on November 30, 2022 so the build can safely be dropped, unless we were hoping to get one more stable release out for 10.15 before removing it.

Should separate builds for macOS 12 and 13 be added or is the 11 build sufficient for all newer macOS?

- Drop build support for EOL OS

### DIFF
--- a/.ci/release_template.md
+++ b/.ci/release_template.md
@@ -11,8 +11,7 @@ include different targets -->
  - <kbd>Windows 7+ (32-bit)</kbd>
  - <kbd>Windows 7+</kbd>
  - <kbd>Windows 10+</kbd>
- - <kbd>macOS 10.15+</kbd> ("Catalina")
- - <kbd>macOS 11+</kbd> ("Big Sur")
+ - <kbd>macOS 11+</kbd> ("Big Sur" or higher)
  - <kbd>Ubuntu 18.04</kbd> ("Bionic Beaver")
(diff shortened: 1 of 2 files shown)


ebbit1q:
because of the way apple deprecates their hardware most people on older machines can only run these old versions of mac os, we like to keep them around as long as they still function

ebbit1q:
mac builds are forwards compatible so these will work on newer versions and unless we have a compelling reason to make a separate version for it because it has some new feature we don't need to use a newer version afaik, do you notice a difference between the versions we currently have?

SlightlyCircuitous:
Alright, that makes sense. I can close this, then.

SlightlyCircuitous:
And to answer you questions, I don't run macOS 12 or 13 so I don't have any compelling reasons. I was just curious since there is that commented out macOS 12 line in the desktop-build and it seemed like we had a lot of separate builds for a lot of different versions of the other OS's, but I guess mac works differently than I am used to.

ebbit1q:
yeah, also compare the package sizes, the mac and windows builds ship with qt included, the linux packages rely on the distro's package manager to install them (this also means you can't install them offline), that's why the linux versions are specific to that version of the os while mac and windows allow forwards compatibility.

PR #4791, opened 2023-04-20
save sets dialog size

### LINKED ISSUE #4778: Interface issue in `Manage sets...`
**System Information:**
Client Version: 2.8.1-beta.12 (2023-03-03)
Build Architecture: 64-bit
Qt Version: 6.3.2
System Locale: en_GB
Install Mode: Standard

When you change the order in the sets view, a hint is displayed, the buttons are pushed above the table and cover content:
grafik

**Steps to reproduce:**
 - Open `Card Database` --> `Manage sets...`
 - Change ordering of sets with clicking on one column
 - Note appears and layout breaks

ebbit1q:
## Related Ticket(s)
- Fixes #4778

the sets dialog can be resized wrong

- don't set a minimum size of the list
- set a minimum size for the dialog
- save the dialog size in the settings
- save the header state in the settings
- allow the headers to be resized

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ preferences
 compile_commands.json
 .vs/
 .vscode/
+.cache
+.gdb_history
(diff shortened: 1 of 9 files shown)


tooomm:
Btw, this was not an issue on 2.8.0

The result with your change seems ok for me and fixes the reported issue.

However, changes to ordering (like clicking on a column header to sort) are saved right away now, and confirming/clicking on the button is no longer needed.

ebbit1q:
I absolutely did not change the behavior of ordering, are you sure?

ebbit1q:
I noticed what was happening, the header state includes the _displayed_ sorting order, this might be a bit annoying. note that at no point this changed your actual saved sets order.

tooomm (approved):
Issue(s) solved for me

PR #4795, opened 2023-04-23
Enable Dependabot

tooomm:
Stuff is out of date

Check several services for updates automatically and propose PRs
- GitHub Actions
- npm

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,44 @@
+# Configuration options: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  # Enable version updates for submodules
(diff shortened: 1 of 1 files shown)


ebbit1q:
with submodules you mean just vcpkg right? vcpkg updates a lot and we don't need the latest version of stuff in it that we don't use, we only care about getting like protobuf through it.

ebbit1q:
we also have dependabot making prs for the webclient already https://github.com/Cockatrice/Cockatrice/commits?author=dependabot%5Bbot%5D

tooomm:
That's only for vulnerability patches I think.

There is no general setup available as far as I'm aware. No config file in the repo, too.

tooomm:
In order to move this PR forward, I did not include submodule updates by Dependabot for now.

PR #4805, opened 2023-05-07
Update macOS Monterey app icon

Zarel:

In macOS 12 (Monterey), Apple introduced iOS-style app icons, which are as of now basically universal. Any icons not in the modern format stick out a lot.

This updates Cockatrice icons to the modern Monterey style.

## Screenshots

BEFORE:
image
AFTER:
image


ebbit1q:
is this backwards compatible to macos 11?

Zarel:
Short answer: Yes.

Long answer: It'll still look like the new icons in macOS 11, but every other app looks like this now, so it's not like it'll look out of place or anything.

Zarel:
Short answer: Yes.

Long answer: It'll still look like the new icons in macOS 11, but every other app looks like this now, so it's not like it'll look out of place or anything.

ebbit1q:
@ZeldaZach can you check if these look right?

Issue #4803, opened 2023-05-06
Format language json files


tooomm:
**Description of New Feature**
Use basic formatting for the default language json file:
https://github.com/Cockatrice/Cockatrice/blob/master/webclient/src/i18n-default.json

**Additional Context**
In https://github.com/Cockatrice/Cockatrice/pull/4800 you see that since all json is in one line, any change - no matter how big - will end up in a +1/-1 change and one can not review what was changed/updated here in GitHub.

PR #4806, opened 2023-05-09
Add indentation to i18n-default.json

SlightlyCircuitous:
Improves readability

## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4803

The entire json tree is smooshed together into one line, impairing readability and also git's ability to generate useful diffs.

Indents the json to match common.i18n.json. I merely read it into Python and then wrote it out again with indents so nothing *should* have changed, but stranger things have happened.


tooomm:
Thank you! That should do it for now.

This will need to get considered when extracting strings from source as well and committing the results back to the repo.

SlightlyCircuitous:
Yes, I was wondering how it got like this as well. Is it automatically generated by some process that doesn't pretty print it?

tooomm:
We have no process integrated or documented. But looking in the [i18next docs](https://www.i18next.com/how-to/extracting-translations) and reading about string extraction, it looks like there are several tools available.
They all have config options on spaces or indentation to solve this.
I just opened https://github.com/Cockatrice/Cockatrice/issues/4808 the other day to try to understand how this works.

I prepared a PR once to automate this and keep the source string up to date for the desktop app automatically: https://github.com/Cockatrice/Cockatrice/pull/4544
It could easily extended to include the webclient as well in a later iteration.

PR #4807, opened 2023-05-10
add a way to replace the user facing list of disallowed words

ebbit1q:
## Related Ticket(s)
- Fixes ##4575

sorry about forgetting this for so long

- adds a new field displaydisallowedwords to the server ini, it overrides the user facing list in a partly backwards compatible way.

## Screenshots

The new message looks like this:
image
Old clients will have them run together like this:
image

### DIFF
--- a/cockatrice/src/window_main.cpp
+++ b/cockatrice/src/window_main.cpp
@@ -476,9 +476,22 @@ QString MainWindow::extractInvalidUsernameMessage(QString &in)
                "</li>";
 
         if (rules.size() == 9) {
-            if (rules.at(7).size() > 0)
-                out += "<li>" + tr("can not contain any of the following words: %1").arg(rules.at(7).toHtmlEscaped()) +
-                       "</li>";
+            if (rules.at(7).size() > 0) {
+                QString words = rules.at(7).toHtmlEscaped();
+                if (words.startsWith("\n")) {
+                    out += tr("no unacceptable language as specified by these server rules:",
+                              "note that the following lines will not be translated");
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+                    for (QString &line : words.split("\n", Qt::SkipEmptyParts)) {
+#else
+                    for (QString &line : words.split("\n", QString::SkipEmptyParts)) {
+#endif
+                        out += "<li>" + line + "</li>";
+                    }
+                } else {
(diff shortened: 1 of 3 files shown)


ebbit1q:
suggestions for changing the wording in the new message are welcome, remember that the list is set server side and won't be translated, so if a server sets it in one language or were to list them in multiple languages they'd always show up like that to the user no matter what their language setting is.

SlightlyCircuitous:
Great work, glad this is getting changed. 

Some thoughts:
- In the new client message, is it possible to make the 'no unacceptable language...' line its own bullet point with the three rules indented under it? ('staff reserves the right...' can stay its own unindented bullet point as its more general). Even just a newline between the technical guidelines and the etiquette guidelines would be good.

- Could the 'no unparlimentary language' be changed to something like 'no discriminatory, abusive, or otherwise hurtful language'? I'm not sure 'unparlimentary' is a particularly common phrase in English (at least US English) outside of an actual parliament or congress

- Could be good to expand the third rule to 'no controversial figures or events' 

- Maybe change "No unacceptable language as specified by these server rules" to "No unacceptable language, including but not limited to the following server rules" and add "at their sole discretion" to "staff reserves the right..." to make it clear that server staff have final say and aren't obliged to allow your username if we forgot a rule or something.

ebbit1q:
it looked weird when I did that. the last line is at the same priority as the others.

this is just an example, servers can use whatever they want here

ditto

ditto

SlightlyCircuitous:
Ah, ok, I see. guess it's all fine then.

ebbit1q:
To clarify, this text is added here https://github.com/Cockatrice/Cockatrice/pull/4807/files#diff-97790e92c403415a183a37d1ea997f908ffe11524a362a5a33ac3597ee05b0b4R482 and here 
https://github.com/Cockatrice/Cockatrice/pull/4807/files#diff-326226192060f2921b4990614e8d49828ad5ee0a6b7ad42a89bf140c440977baR136

As you see the line "no unacceptable language as specified by these server rules:" will be coded into the application and available to translators. The line "no attempts at impersonating staff\nno unparliamentary language\nno references to controversial figures\nstaff reserves the right to remove accounts deemed inappropriate" will be added as an example to the .ini.example file as a template for people to use when starting their own server.

I'm not exactly satisfied with the layout of the text either, and I know that with some effort you could make it look like anything you'd like, I'm just not sure if it's worth it for an error message.

PR #4811, opened 2023-05-18
CI: Remove Ubuntu 18.04 Bionic Beaver

SlightlyCircuitous:

Ubuntu 18.04 Bionic Beaver will [reach the end of standard support on May 31, 2023.](https://ubuntu.com/18-04) 

-Remove Bionic Beaver from CI build.
-Update release template accordingly

There is a workaround in [compile.sh](https://github.com/Cockatrice/Cockatrice/blob/2b42bee4249919488a0a3cb03a0ae119f8cc88c1/.ci/compile.sh#LL131C5-L131C48) that seems to exist solely to support building Bionic in the CI environment; can this also be removed or is it at all necessary for people who want to compile Cockatrice for Bionic themselves?


ebbit1q:
18.04 is our minimum requirements build, it's there to test if cockatrice still works on qt5.9

SlightlyCircuitous:
Ah, ok, didn't know that.

ebbit1q:
well if it's about to eol we might need to consider something else. not sure if it's still useful for us under esm.

SlightlyCircuitous:
how might esm change its usefulness for testing?

ebbit1q:
I don't know. As far as I know that means public support ends.

SlightlyCircuitous:
My understanding is that standard updates to the 18.04 repos end on May 31st; however, individuals can sign up for free Ubuntu Pro (successor to ESM) and get security updates to Main (system repo) and Universe (community free software repo) on up to five machines for five years. So, normal users should still be able to use 18.04 for a while if they don't have too many machines. However, I don't know what that means for Github's Bionic VM since I can't imagine Github will be paying for Enterprise level Pro to keep that updated.

ebbit1q:
we don't run bionic in the github workflow, we just run a docker container for everything to keep things consistent

SlightlyCircuitous:
ah, yes, sorry, I realized my mistake and edited my post to be about docker before you commented. I'm not sure we can count on that image either, but its hard to tell from their website.

ebbit1q:
we'll wait and see

tooomm:
If it's about testing, we should be able to skip the deployment for it and don't upload the binary. But still build and test qt5 as long as it works with docker and we need it.

tooomm:
> I assume that the Docker image will become unsupported as 18.04 is currently the oldest Ubuntu supported by Docker. I can't imagine they will continue support through the ESM/Pro period since they would still support 16.04 (until 2026) if they were in the habit of doing that.

That was exactly the case, with it going EOL, no more updates to the image since. May 30th was the last one: 
https://hub.docker.com/layers/library/ubuntu/bionic-20230530/images/sha256-dca176c9663a7ba4c1f0e710986f5a25e672842963d95b960191e2d9f7185ebe?context=explore

Do we finally reconsider streamlining the builds a bit and getting rid of outdated ones? There are 2 newer LTS builds we support for quite a while in between... not many reasons to still use it.

>18.04 is our minimum requirements build, it's there to test if cockatrice still works on qt5.9

Only the Windows 7 build is also still on Qt5 (but 5.15), because Qt6 is not available for old systems.
Do we even have an official min requirement listed/communicated? I'm not aware.

**By the way, Qt5.9 is EOL since over 3 years and even Qt 5.15 LTS ended together with 5-year-old Ubuntu 18.04 LTS in May this year.**

PR #4812, opened 2023-06-01
Add Debian 12 "Bookworm" Build

SlightlyCircuitous:

Debian 12 "Bookworm" is set to [release on June 10, 2023](https://wiki.debian.org/DebianBookworm) so users will probably want a build for it soon.

- Adds a dockerfile and a matrix entry for Bookworm
- Adds Bookworm to the Release template

The dependencies in the docker build are very different from Debian 11 as Debian 12 has Qt6 by default and also uses the new qt6-svg-dev and qt6-websockets-dev packages that Ubuntu recently switched over to. I have tested building with these dependencies in a VM of Debian 12 RC4 and it seemed to work well. I will test the Github Actions build there too if it succeeds.

### DIFF
--- a/.ci/Debian12/Dockerfile
+++ b/.ci/Debian12/Dockerfile
@@ -1,0 +1,26 @@
+FROM debian:12
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        build-essential \
+        ccache \
+        clang-format \
+        cmake \
+        file \
+        g++ \
+        git \
+        libgl-dev \
+        liblzma-dev \
+        libmariadb-dev-compat \
+        libprotobuf-dev \
+        libqt6multimedia6 \
+        libqt6sql6-mysql \
+        qt6-svg-dev \
+        qt6-websockets-dev \
+        protobuf-compiler \
+        qt6-l10n-tools \
+        qt6-multimedia-dev \
+        qt6-tools-dev \
+        qt6-tools-dev-tools \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
--- a/.ci/Debian12/Dockerfile
+++ b/.ci/Debian12/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bookworm
+FROM debian:12
 
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
--- a/.ci/Debian12/Dockerfile
+++ b/.ci/Debian12/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:12
+FROM debian:bookworm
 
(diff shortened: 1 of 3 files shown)


SlightlyCircuitous:
I think I was too early on this. It doesn't look like "debian:12" is a real tag on docker yet. "debian:bookworm" is, but I don't know if there is a difference between making this file with "FROM debian:bookworm" versus "FROM debian:12". The debian 11 build uses "FROM debian:11" so I am inclined to wait for consistancy. I will have this as a draft until the tag appears.

ebbit1q:
there isn't really a difference, I wouldn't know why :12 isn't there yet but :bookworm would be just fine, as long as that image stays under the same name

SlightlyCircuitous:
The .deb from Github Actions installs correctly on my bookworm test VM, Cockatrice opens without incident, and oracle runs appropriately on start up. I can make, save, and load decks and participate in local games as expected.

ebbit1q (approved):
cool

tooomm:
> I think I was too early on this. It doesn't look like "debian:12" is a real tag on docker yet. "debian:bookworm" is, but I don't know if there is a difference between making this file with "FROM debian:bookworm" versus "FROM debian:12". The debian 11 build uses "FROM debian:11" so I am inclined to wait for consistancy. I will have this as a draft until the tag appears.

The reason is that Debian 12 was not yet officially released, the available bookworm tags on the official Debian Docker Hub are linking to the "testing" release (between stable and unstable).
As the image officially got released today, it's only a matter of time (hours?) until the latest tag on Docker Hub moves to Debian 12 as well and the `12` tag will be available.

I prefer to use that speaking and very clear tag over a named one. Not only because it matches our folders and used naming otherwise for Debian. It also directly shows which is the older or newer release. Many not so experienced users have no clue what Bullseye or Bookworm stand for.

tooomm:
Tag "12" is available now

SlightlyCircuitous:
Alright, I have changed the tag to :12 as per tooomm's suggestion. Would you like me to install and test this new build or is the change not large enough to warrant doing that?

tooomm:
If you a test is not too much struggle for you, yes, please! 😊

Probably good to be on the more safe side when releasing user facing stuff.

SlightlyCircuitous:
Ok, I'll reinstall debian 12 in VM and let you know how it works

SlightlyCircuitous:
New .deb works as expected on freshly installed Debian 12 final release in VM. Performed the same checks as before:

> Cockatrice opens without incident, and oracle runs appropriately on start up. I can make, save, and load decks and participate in local games as expected.

SlightlyCircuitous:
Is it concerning that the macOS 10.15 build failed this time around? macOS 11 failed due to the known homebrew error (I think), and macOS debug simply doesn't work, but I'm not sure what happened with 10.15. Some kind of protobuf problem, it seems like?

PR #4826, opened 2023-07-07
Remove Ubuntu 22.10 Kinetic Kudu Build

SlightlyCircuitous:

Ubuntu 22.10 Kinetic Kudu will [go end of life on July 20, 2023](https://lists.ubuntu.com/archives/ubuntu-announce/2023-June/000291.html) so we don't need to build for it in future releases.

- Removes Kinetic build from CI
- Removes Kinetic from the release template


ebbit1q (approved):
thanks

PR #4870, opened 2023-09-11
README: Cleanup translations widget and links

tooomm:
Transifex does no longer provide a diagram with translation data in picture format for easy integration. Only javascript code is available.

- remove broken graphics
- fix links


ebbit1q (approved):
thanks

PR #4864, opened 2023-08-18
fix issues with #4648

ebbit1q:
resolves some of the issues that were still unsolved when #4648 was merged

PR #4871, opened 2023-09-11
Rename "stack until found" feature

tooomm:
## Related Ticket(s)
- Related #4648
- Related #4864

I find the selected "Take top cards until" wording of the recently added "move cards from top of library until" not great.

- Use "Put top cards on stack until" phrasing, to stay in line with the rest of the options and be more clear of the action:
  image
- Tried to improve the dialog

<br>

>Which "search syntax" is actually supported in the follow-up dialog?

E.g. `mv<4 -t:land` for cascading

### DIFF
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -1313,8 +1313,8 @@ void Player::actMoveTopCardsUntil()
     QString expr = previousMovingCardsUntilExpr;
     for (;;) {
         bool ok;
-        expr = QInputDialog::getText(game, "Put top cards on stack until", "Card name (or search expressions)", {}, expr,
-                                     &ok);
+        expr = QInputDialog::getText(game, "Put top cards on stack until", "Card name (or search expressions)", {},
+                                     expr, &ok);
         if (!ok) {
             return;
         }
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -771,7 +771,7 @@ void Player::retranslateUi()
         aMoveTopCardToExile->setText(tr("Move top card to e&xile"));
         aMoveTopCardsToGraveyard->setText(tr("Move top cards to &graveyard..."));
         aMoveTopCardsToExile->setText(tr("Move top cards to &exile..."));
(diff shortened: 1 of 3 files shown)


cockatrice/src/player.cpp
```diff
@@ -1313,8 +1313,8 @@ void Player::actMoveTopCardsUntil()
     QString expr = previousMovingCardsUntilExpr;
     for (;;) {
         bool ok;
-        expr =
-            QInputDialog::getText(game, "Take top cards until", "Select card (accepts search syntax)", {}, expr, &ok);
+        expr = QInputDialog::getText(game, "Put top cards on stack until", "Card name (accepts search syntax)", {},
```
ebbit1q, line 1316: the box is much more useful for stuff that isn't the name of the card, eg mv<4 -t:land for cascading, I'd prefer to avoid the use of "name"

ebbit1q: we might consider changing the text in the deck editor as well, it currently says "search by card name" before typing

tooomm: > the box is much more useful for stuff that isn't the name of the card, eg mv<4 -t:land for cascading, I'd prefer to avoid the use of "name"

Nice, I had no clue! And the dialog is not helping here, see my question in the opening post: 
"Which search syntax is actually supported in the follow-up dialog?"
I did only try with a placeholder "goblin *", and it did not work.

tooomm: I went with `Card name (or advanced syntax)` now and used the same phrasing in the deck editor search. 👍 
That way people hopefully get the link that it's the same search syntax.

I still think the advanced search help is a bit hidden though. E.g. a colored icon over a grayed out one would signal that it can be interacted with, and a hover label on the icon would be great to explain what it's about.

ebbit1q: I don't like the "advanced" label, we only have one search syntax, https://scryfall.com/docs/syntax doesn't call it advanced either, maybe just call them "search expressions" instead of syntax which might be a less used word.

tooomm: I like this suggestion!

ebbit1q (approved):
thanks

PR #4879, opened 2023-09-18
Update to new Protobuf

### LINKED ISSUE #4825: Failing to build on arch
**System Information:**
Archlinux
Qt 6.5.1
Protobuf 23.3
GCC 13.1.1

When building on arch, the build fails.

CMake emits this warning:
```
CMake Warning at /usr/share/cmake/Modules/FindProtobuf.cmake:524 (message):
  Protobuf compiler version 23.3 doesn't match library version 4.23.3
  CMakeLists.txt:219 (find_package)


-- Found Protobuf: /usr/lib/libprotobuf.so (found version "4.23.3") 
```

Then the build fails with this:
```
/usr/bin/ld: /usr/lib/libprotobuf.so: undefined reference to symbol '_ZN4absl12lts_2023012512log_internal17MakeCheckOpStringIllEEPNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEET_T0_PKc'
/usr/bin/ld: /usr/lib/libabsl_log_internal_check_op.so.2301.0.0: error adding symbols: DSO missing from command line
```

full build log: [gist](https://gist.github.com/skwerlman/f83f4b7625a72c1cb475353dba2587db)

edit: this also happens when building normally, so i have updated the issue

**Steps to reproduce:**
 - git clone and cd in
 - mkdir build
 - cmake ..
 - make -j1

rdrpenguin04:
## Related Ticket(s)
- Fixes #4825

Newer versions of Protobuf have different APIs. This updates the CMakeLists to find the new version correctly, and it updates the code to remove non-existent APIs.

- Updated `CMakeLists`
- Removed line length snipping in `common/debug_pb_message.cpp`

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -85,7 +85,7 @@ void Servatrice_GameServer::incomingConnection(qintptr socketDescriptor)
     Servatrice_ConnectionPool *pool = findLeastUsedConnectionPool();
 
     auto ssi = new TcpServerSocketInterface(server, pool->getDatabaseInterface());
-    connect(ssi, SIGNAL(incTxBytes), this, SLOT(incTxBytes));
+    connect(ssi, SIGNAL(incTxBytes(qint64)), this, SLOT(incTxBytes(qint64)));
     ssi->moveToThread(pool->thread());
     pool->addClient();
     connect(ssi, SIGNAL(destroyed()), pool, SLOT(removeClient()));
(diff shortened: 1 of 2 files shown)


skwerlman:
This fails to link for me on arch linux when running `cmake ..` with no extra flags. It does link successfully when checking out 7974f69, so it seems that dbda722 needs a revert since `CONFIG` is necessary to have cmake use the system's protobuf cmakelists instead of the (broken) cmake ones

cc: @ZeldaZach

PR #4884, opened 2023-10-02
Support MacOS 12 & 13. Support Protobuf 23. Deprecate MacOS 11.

### LINKED ISSUE #4825: Failing to build on arch

ZeldaZach:
Supersedes #4879 and Fixes #4825

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -113,7 +113,7 @@ jobs:
           - distro: UbuntuJammy
             package: DEB
             test: skip # running tests on all distros is superfluous
-            
+
           - distro: UbuntuLunar
             package: DEB
 
@@ -229,7 +229,7 @@ jobs:
             qt_version: homebrew
             type: Release
             do_tests: 1
-            make_package: 1
+            make_package: 0
 
     name: macOS ${{matrix.target}}
     needs: configure
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -251,8 +251,8 @@ jobs:
           install_qt: ${{matrix.qt_version}}
         run: |
           brew update
-          brew install protobuf@21
-          brew link --force protobuf@21
+          brew install protobuf
+          brew link --force protobuf
           if [[ $install_qt == homebrew ]]; then
             brew install qt --force-bottle
           else # for some reason the tests fail with the action installed qt?
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -215,21 +215,21 @@ jobs:
             do_tests: 1
             make_package: 1
 
-          - target: 11_Big_Sur
-            os: macos-11
-            xcode: 12.5.1
+          - target: 12_Monterey
+            os: macos-12
+            xcode: 13.3
             qt_version: homebrew
             type: Release
             do_tests: 1
             make_package: 1
 
-#          - target: 12_Monterey
-#            os: macos-12
-#            xcode: 13.3
-#            qt_version: homebrew
-#            type: Release
-#            do_tests: 1
-#            make_package: 1
+          - target: 13_Ventura
+            os: macos-13
+            xcode: 14.3.1
+            qt_version: homebrew
+            type: Release
+            do_tests: 1
+            make_package: 1
 
     name: macOS ${{matrix.target}}
     needs: configure
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,15 +220,9 @@ find_package(Protobuf CONFIG)
 if(NOT Protobuf_FOUND)
   find_package(Protobuf REQUIRED)
 endif()
-set(PROTOBUF_LIBRARY "protobuf::libprotobuf")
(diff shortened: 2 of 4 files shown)


common/debug_pb_message.cpp
```diff
@@ -43,12 +34,7 @@ void LimitedPrinter::PrintString(const std::string &val,
                                  ::google::protobuf::TextFormat::BaseTextGenerator *generator) const
 {
     auto length = val.length();
```
skwerlman: this line needs to be removed to fix CI

ZeldaZach: Yeah I'm still playing around with this PR, seeing what's up. I've done a bit of research on the front and I want to see if I can maintain compatibility with what we've done so far

CMakeLists.txt
```diff
@@ -216,11 +216,19 @@ include(FindQtRuntime)
...
-find_package(Protobuf REQUIRED)
-if(NOT EXISTS "${Protobuf_PROTOC_EXECUTABLE}")
+find_package(Protobuf CONFIG)
+if(NOT Protobuf_FOUND)
+  find_package(Protobuf REQUIRED)
+endif()
+set(PROTOBUF_LIBRARY "protobuf::libprotobuf")
+set(PROTOBUF_PROTOC_EXECUTABLE "protobuf::protoc")
+
+if(${Protobuf_VERSION} VERSION_LESS "21.0.0" AND NOT EXISTS "${Protobuf_PROTOC_EXECUTABLE}")
```
ebbit1q, line 226: I think it's useful to print where these have been found, see this patch for reference https://gitlab.archlinux.org/archlinux/packaging/packages/cockatrice/-/blob/c2972726ac2ae3fcd52303e9eaee4cee18dbd664/protobuf-23.patch

ebbit1q (changes requested):
I've fixed the regression regarding log truncation, this pr now works as intended on my arch box.

I wish you the best of luck making vcpkg and cmake on windows happy.

CMakeLists.txt
ZeldaZach: I'm not exactly sure what you're referring to

.github/workflows/desktop-build.yml
```diff
@@ -215,21 +215,21 @@ jobs:
...
             make_package: 1
 
-#          - target: 12_Monterey
-#            os: macos-12
-#            xcode: 13.3
-#            qt_version: homebrew
-#            type: Release
-#            do_tests: 1
-#            make_package: 1
+          - target: 13_Ventura
```
ebbit1q, line 226: is there a benefit to having a separate macos 13 build? does it perform differently than the forwards compatible macos 12 build?

ZeldaZach: I believe 12 is forward compatible

tooomm: All macOS builds should be forward compatible.
Is the 10.15 release not working on macOS 13? It should, no?

We used to only have binaries for the oldest macOS version we were supporting. There were never issues afaik.
Releasing multiple `.dmg`'s started when 10.14? offered dark mode support and we added a second release just for that feature.
Since then we kept having and maintaining multiple macOS releases in parallel.

Wouldn't it be easier to only ensure the oldest macOS version we support is working and offering users a single release file only? CI can build and test the next newer already as well of course.

skwerlman (approved):
builds and runs perfectly on my machine (cmake 3.27.7, protobuf 24.3)

.github/workflows/desktop-build.yml
```diff
@@ -251,8 +251,8 @@ jobs:
           install_qt: ${{matrix.qt_version}}
         run: |
           brew update
-          brew install protobuf@21
-          brew link --force protobuf@21
+          brew install protobuf
+          brew link --force protobuf
```
tooomm, line 255: This does break macOS 10.15 now.

Homebrew uses protobuf 24 which is not available for anything below macOS 12.

PR #4896, opened 2023-10-07
Oracle: Update MV (CMC) mapping & face data for split cards

### LINKED ISSUE #4271: CMC not correct for all special layouts
**System Information:**
Client Version: 2.8.1-custom(2b0a997) (2021-02-24)
Build Architecture: 64-bit
Qt Version: 5.12.9

Cards with a `flip` layout have no value set for `Converted Mana Cost` , same for `meld`.
`transform` front side is ok, back side is falsely `0` - instead the same as the front side. [There was a rules change here](https://magic.wizards.com/en/articles/archive/feature/shadows-over-innistrad-mechanics)
`aftermath` seems to work fine. I did not check other split variants.

cmc

The value for [Jushi Apprentice](https://scryfall.com/card/chk/70/jushi-apprentice-tomoya-the-revealer) should be `2`

The value for [Gisela, the Broken Blade](https://scryfall.com/card/emn/28/gisela-the-broken-blade) should be `4`
The value for [Brisela, Voice of Nightmares](https://scryfall.com/card/emn/15b/brisela-voice-of-nightmares) should be `11` (4+7)

tooomm:
MTGJSON to deprecate/rename fields

- When building our cards.xml card database in Oracle, pull data for split cards from a different field in the MTGJSON source.
- Formerly used [`faceConvertedManaCost`](https://mtgjson.com/data-models/card/card-atomic/#faceconvertedmanacost) is deprecated and replaced by [`faceManaValue`](https://mtgjson.com/data-models/card/card-atomic/#facemanavalue) according to the MTGJSON docs.
- Update default mapping for cmc.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -189,8 +189,8 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet,
 {
     // mtgjson name => xml name
     static const QMap<QString, QString> cardProperties{
-        {"manaCost", "manacost"}, {"convertedManaCost", "cmc"}, {"type", "type"},
-        {"loyalty", "loyalty"},   {"layout", "layout"},         {"side", "side"},
+        {"manaCost", "manacost"}, {"manaValue", "cmc"}, {"type", "type"},
+        {"loyalty", "loyalty"},   {"layout", "layout"}, {"side", "side"},
     };
 
     // mtgjson name => xml name
@@ -365,7 +365,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet,
 
             // add other face for split cards as card relation
             if (!getStringPropertyFromMap(card, "side").isEmpty()) {
-                properties["cmc"] = getStringPropertyFromMap(card, "faceConvertedManaCost");
+                properties["cmc"] = getStringPropertyFromMap(card, "faceManaValue");
                 if (layout == "meld") { // meld cards don't work
                     static const QRegularExpression meldNameRegex{"then meld them into ([^\\.]*)"};
                     QString additionalName = meldNameRegex.match(text).captured(1);


CTimmerman (approved):

oracle/src/oracleimporter.cpp
```diff
@@ -365,7 +365,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet,
 
             // add other face for split cards as card relation
             if (!getStringPropertyFromMap(card, "side").isEmpty()) {
-                properties["cmc"] = getStringPropertyFromMap(card, "faceConvertedManaCost");
+                properties["cmc"] = getStringPropertyFromMap(card, "manaValue");
```
ZeldaZach, line 368: This isn't correct. `faceManaValue` should be used, as this is a face product

ZeldaZach: https://mtgjson.com/data-models/card/card-set/#facemanavalue

tooomm: I'm totally aware that faceConvertedManaCost is deprecated and superseded by faceManaValue. See the PR description and discussions in the linked issue (as well as the mentioned MTGJSON issue).

As noted, `When testing with faceManaValue it did not solve the issue though.`
Which is no surprise as it hold the same data as faceConvertedManaCost.

I'm not saying this is the right solution here, and this needs a check by somebody more familiar with the code and logic around this.
But the result it gives, appears to be better and I did not spot issues yet.

tooomm:
I had another look and tried to ensure we save the right data to our XML card database to begin with.
This PR is now making sure we use no values from our MTGJSON source any longer that are marked as deprecated:
- Map our initial "cmc" value to "manaValue" instead "[convertedManaCost](https://mtgjson.com/data-models/card/card-atomic/#convertedmanacost)"
- For cards with a side value present, use the value from "faceManaValue" instead "[faceConvertedManaCost](https://mtgjson.com/data-models/card/card-atomic/#faceconvertedmanacost)" for cmc

This should not have any immediate effect, as the deprecated values are still present currently as far as I can tell, but ensures future compatibility.

That looks about right to me and a starting point.

---

The initial issue is not solved by this.
@ZeldaZach Can you check and confirm MTGJSON has all manaValue's and faceManaValue's for all kind of cards like (M)DFC etc. as intended?
There is some confusion currently when which data is available at MTGJSON, see https://github.com/mtgjson/mtgjson/issues/1121.

oracle/src/oracleimporter.cpp
tooomm: Going step by step now, ensuring we save the right data first.

But if MTGJSON has all data and fields correct (which still needs to be verified), then Cockatrice is probably displaying certain data not correctly.

CTimmerman: MTGJSON is inconsistent (not even manaValue is present on all cards like the first instances of Blex, or my parser needs work) and Cockatrice should use the back face MV for the back face. https://github.com/Cockatrice/Cockatrice/issues/4271#issuecomment-1849757496

tooomm: Let's find all data issues related to the problems in Cockatrice and report them at MTGJSON.

Then we can see how to fix stuff in Cockatrice.

tooomm:
Is there anything blocking this which I'm not aware of? @ZeldaZach 
We should stop using the deprecated field and use its successor.

Please check and merge.

PR #4907, opened 2023-10-13
copy arrows on transform

ebbit1q:
## Related Ticket(s)
- #4753 

transforming cards doesn't take into account the arrows pointing at the original card, these are now copied instead of these arrows going to the shadow realm.

- transforming cards inherit arrows

### DIFF
--- a/common/server_player.cpp
+++ b/common/server_player.cpp
@@ -1489,12 +1489,12 @@ Server_Player::cmdCreateToken(const Command_CreateToken &cmd, ResponseContainer 
                         startCard = card;
                     }
                     const auto *targetItem = arrow->getTargetItem();
-                    if (targetItem == targetCard){
+                    if (targetItem == targetCard) {
                         sendGameEvent = true;
                         arrow->setTargetItem(card);
                         targetItem = card;
                     }
-                    if (sendGameEvent){
+                    if (sendGameEvent) {
                         Event_CreateArrow event;
                         ServerInfo_Arrow *arrowInfo = event.mutable_arrow_info();
                         arrowInfo->set_id(arrow->getId());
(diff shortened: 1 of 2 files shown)


common/server_player.cpp
```diff
@@ -1475,6 +1475,47 @@ Server_Player::cmdCreateToken(const Command_CreateToken &cmd, ResponseContainer
...
+                    }
+                    if (sendGameEvent) {
+                        Event_CreateArrow event;
+                        ServerInfo_Arrow *arrowInfo = event.mutable_arrow_info();
+                        arrowInfo->set_id(arrow->getId());
+                        arrowInfo->set_start_player_id(player->getPlayerId());
+                        arrowInfo->set_start_zone(startCard->getZone()->getName().toStdString());
+                        arrowInfo->set_start_card_id(startCard->getId());
+                        const Server_Player *arrowTargetPlayer = qobject_cast<const Server_Player *>(targetItem);
+                        if (arrowTargetPlayer != nullptr) {
```
ZeldaZach, line 1505: I believe `!= nullptr` is implicit here

common/server_player.cpp
```diff
@@ -1475,6 +1475,47 @@ Server_Player::cmdCreateToken(const Command_CreateToken &cmd, ResponseContainer
                                      attachedCard->getZone()->getPlayer()->getPlayerId());
             }
 
+            // Copy Arrows
+            const QList<Server_Player *> &players = game->getPlayers().values();
```
ZeldaZach, line 1479: Was this code block copied / modified from somewhere else? Would it make sense to functionize that if that's the case?

ZeldaZach:
Do we want to revert #4904 with this change? Or is that fine to stay as-is?

ebbit1q:
it's not hurting anything to have

common/server_player.cpp
ebbit1q: what do you mean? it's important to check if it's a player or a card.

PR #4908, opened 2023-10-13
Handle Qt6.6 Deprecations

ZeldaZach:
- Use QSqlQuery.exec instead of QDatabase.exec
- Use QColor::fromString instead of QColor.setNamedColor


ebbit1q:
qt 6.6 eh? cool

PR #4910, opened 2023-10-14
Address /W4 compiler warnings for Windows

ZeldaZach:
I had time and felt like it. Mostly just remaining, but two things to call out

1) The RNG function has a branch that doesn't make sense as the method returns an unsigned integer, but it attempts to negate the value. It's annotated that it's there to prevent crashes, which shouldn't matter overall for us.

2) The ZIP function seems to have an extra return statement that makes Windows sad. I didn't want to remove it, so I guarded against non-Windows complication.

### DIFF
--- a/oracle/src/zip/zipglobal.cpp
+++ b/oracle/src/zip/zipglobal.cpp
@@ -95,10 +95,6 @@ int OSDAB_ZIP_MANGLE(currentUtcOffset)()
     const int utcOffset = -qRound(difftime(global_time_t, local_time_t));
     return tm_struct->tm_isdst > 0 ? utcOffset + 3600 : utcOffset;
 #endif // No UTC
-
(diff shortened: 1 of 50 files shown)


oracle/src/zip/zipglobal.cpp
```diff
@@ -96,7 +96,9 @@ int OSDAB_ZIP_MANGLE(currentUtcOffset)()
     return tm_struct->tm_isdst > 0 ? utcOffset + 3600 : utcOffset;
 #endif // No UTC
 
+#if !defined(Q_OS_WIN)
     return 0;
```
ebbit1q, line 100: this return statement is not reachable at all... look at the #ifs

ebbit1q:
I prefer value_ instead of _value

PR #4914, opened 2023-10-15
Use proto21 on macos11

ZeldaZach:


tooomm (approved):

tooomm:
Quite unfortunate, [proto21](https://formulae.brew.sh/formula/protobuf@21) just now also removed support for macOS <12 which makes the build fail again: https://github.com/Cockatrice/Cockatrice/actions/runs/6565187763

PR #4919, opened 2023-10-18
Include missing strings for translation

tooomm:
Those strings were not recognized and extracted as source strings for translations due to missing tr() wrappings. Transifex could never offer them to translators.

- Add tr()


ebbit1q (approved):
thanks, we clearly missed these

PR #4921, opened 2023-10-21
CI: Cleanup & updates

### LINKED ISSUE #4777: Incompatible Qt5 library error messages when starting Qt6 version
**System Information:**
Client Version: 2.8.1-beta.12 (2023-03-03)
Build Architecture: 64-bit
Qt Version: 6.3.2
System Locale: en_GB
Install Mode: Standard

```
---------------------------------------------------------------------------
Theme changed: "Default"
In C:/Program Files/Cockatrice/plugins/imageformats/qicns.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/plugins/imageformats/qtga.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/plugins/imageformats/qtiff.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/plugins/imageformats/qwbmp.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/plugins/imageformats/qwebp.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/Plugins/imageformats/qicns.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/Plugins/imageformats/qtga.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/Plugins/imageformats/qtiff.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/Plugins/imageformats/qwbmp.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
In C:/Program Files/Cockatrice/Plugins/imageformats/qwebp.dll:
  Plugin uses incompatible Qt library (5.15.0) [release]
SoundEngine: disabling sound
Sound theme changed: "Default"
SoundEngine: disabling sound
Unable to load qt translation "qt_" at "C:/Program Files/Cockatrice/Resources/translations"
Unable to load "cockatrice" translation "cockatrice_" at "C:/Program Files/Cockatrice/translations"
main(): starting main program
main(): MainWindow constructor finished
PictureLoader: cache fail for "_trice_card_back_226158"
main(): ui.show() finished
Startup: found config with current version
CardDatabase::loadCardDatabases start
...
```

tooomm:
## Related Ticket(s)
- Linking #4883
- Fixes #4777

No more 32bit offerd on Windows
Outdated versions

Windows:
- Simplifications for config file and binary file names
- Use Qt 6.5 LTS on Windows 10 (also fixes some Qt5 error messages for me)

macOS:
- Update Xcode versions to latest available for the respective macOS line


ebbit1q:
this will break the updater again by changing the name of the build artifact

tooomm:
> this will break the updater again by changing the name of the build artifact

It won't? The changed Windows binaries are still found by the updater as far as I understand [the code](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/releasechannel.cpp#L71).

ebbit1q:
that's true, the new name will still match our updater logic, thanks for verifying

ebbit1q:
we do sometimes get people asking why there is no win 11 built though, the + is supposed to signify it is also compatible with win11, people often don't read the + though, I'm not sure what to add to make clear to people that the win10 build is also for win11

tooomm:
It's not only the Windows files that are forward compatible. Same applies to macOS, which does not have a `+`.
I think any pattern we use should be uniformly applied across operating systems and released files to not be misleading.

I'm not sure a label on the binary is really helping, we [highlight the forward compatibility in the release](https://github.com/Cockatrice/Cockatrice/blob/master/.ci/release_template.md) already.

If we change the pattern again, it should be future proof.
There could be generic tags like `-macOS` and `-Win` that are moving targets for the lowest compatible of their kind. A solid default. Any other versions/tags (if needed) can be more specific. The updater does not always need to catch those and can fallback to the general one.

ebbit1q:
yeah, I think you're right, there isn't a good solution for this, people should just read the list but they just don't

ebbit1q (approved):

tooomm:
What do you think about moving to an updated, future proof binary pattern @ZeldaZach?

Now with all binaries being 64bit, it's a bit simpler.

PR #4930, opened 2023-10-31
make cards on the stack slightly overlap to stress order

### LINKED ISSUE #4872: Make cards on the stack slightly overlap to stress order
Cards on the stack are shown next to each other. There is no way to know if the cards more on the top are the more recent ones or the other way around when just looking at it.

It would help to just slightly overlap the cards to clearly visualize their order.
Any potential information that might be covered up can easily be reviewed with moues over or middle click.

ebbit1q:
## Related Ticket(s)
- Fixes #4872

the order of cards on the stack was difficult to understand

- dragging cards to the stack now places the card at the location it is dropped
- cards on the stack will overlap by 1/3rd always

## Screenshots

image

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -299,6 +299,7 @@ void GeneralSettingsPage::retranslateUi()
 
 AppearanceSettingsPage::AppearanceSettingsPage()
 {
+    SettingsCache &settings = SettingsCache::instance();
     QString themeName = SettingsCache::instance().getThemeName();
 
     QStringList themeDirs = themeManager->getAvailableThemes().keys();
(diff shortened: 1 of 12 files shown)


tooomm:
Looking good. Very helpful 👍 
image

ebbit1q:
if we'd like the overlap amount to be configurable I could add it to the settings, it is currently 1/3rd of a card height

tooomm:
Personally, I do not think that's needed.

I guess you tried other values like 1/4 or 1/5 as well, and decided 1/3 was the best?
The exact value should not matter too much as long as there is an overlap.

Probably smaller numbers are better than bigger - like 50% overlap even with only 2-3 cards on the stack would not be great.

ebbit1q:
I haven't tried other numbers

tooomm (commented):
1/3 looks good to me

ebbit1q:
do we want to do the same with the vertical hand layout? they look strange next to each other.

cockatrice/src/dlg_settings.cpp
```diff
@@ -413,6 +418,7 @@ void AppearanceSettingsPage::retranslateUi()
     cardsGroupBox->setTitle(tr("Card rendering"));
     displayCardNamesCheckBox.setText(tr("Display card names on cards having a picture"));
     cardScalingCheckBox.setText(tr("Scale cards on mouse over"));
+    verticalCardOverlapPercentLabel.setText(tr("Minimum overlap percentage of cards in vertical hand and stack"));
```
tooomm, line 421:
```suggestion
    verticalCardOverlapPercentLabel.setText(tr("Minimum overlap percentage of cards on the stack and in vertical hand"));
```

tooomm (changes requested):
Okok, you went for the settings option. 👍 
Changing it does not apply the new setting unless the zones are updated though.

I would also switch the wording in settings around to stress the stack option (always active) over the vertical cards one (optional).

tooomm:
>Changing (the overlapping value) does not apply the new setting unless the zones are updated though.

Do you plan to address this @ebbit1q?

ebbit1q:
I tried but for some reason it doesn't work, I decided to remove the code for it again.

tooomm (approved):
We can list it as improvement and look into it later, otherwise we have another PR with enhancements waiting for months.

ZeldaZach (approved):

PR #4933, opened 2023-11-04
Allow 2 <cards> elements in xml files

SlightlyCircuitous:
## Related Ticket(s)
- https://github.com/Cockatrice/Magic-Token/pull/225

The linked PR changed the tokens.xml file by adding a second `<cards>` element to accommodate the addition of Theros creature tokens to the xml. Under the current validation schema, attempting to validate tokens.xml with xmllint gives the impression that there are too many `<cards>` elements and that the xml is invalid. 

- Changes the validation schema to allow for up to 2 `<cards>` elements in the xml (up from the default value of 1). Using the new rules, both tokens.xml and cards.xml are viewed as perfectly valid by xmllint.

### DIFF
--- a/doc/carddatabase_v4/cards.xsd
+++ b/doc/carddatabase_v4/cards.xsd
@@ -75,7 +75,7 @@
                         </xs:sequence>
                     </xs:complexType>
                 </xs:element>
-                <xs:element name="cards" minOccurs="0">
+                <xs:element name="cards" minOccurs="0" maxOccurs="2">
                     <xs:complexType>
                         <xs:sequence>
                             <xs:element type="cardType" name="card" maxOccurs="unbounded" minOccurs="0" />


ebbit1q:
why would we want this though? I'd rather just remove the unnecessary card tags, they serve no purpose.

SlightlyCircuitous:
That would be fine with me too; really I just want the validation to work. I just got the impression from the token PR that the extra element separating the Theros tokens out were wanted so this seemed like the solution. Maybe I am wrong there.

tooomm:
@ebbit1q If you look in the linked PR, you can get a feeling for the why and purpose.

In the end, I do not care too much and if people think there is no value, that's also fine.

ebbit1q:
I don't see anything in that pr about the extra `<cards>` elements, they don't actually do anything at all, if the concern is with human readability a comment is more appropriate

tooomm:
Sorry, I'm not sure what you are getting at.

Comment can be found here: 🤔 
https://github.com/Cockatrice/Magic-Token/pull/225/files#diff-ef2c5b2e0216513ba7314de2c6637a3b48b09a5ea3ee44d13c951195fb293697R14644

This second element groups all special, non-creature spells from Theros Challenge Decks that are now included in the tokens file to separate them from our other tokens.

Obviously, it does not change anything for the end user in Cockatrice.

ebbit1q:
yes, the comment is already there, why do we need more `<cards>` tags in our file?

tooomm:
I understand, you don't like it and see no benefit.

Maybe you can imagine that it helps to keep this special "group" separated instead having those cards mixed with others at one point. 
As you know, we use alphabetical order of `card`s in the `cards` element.

ebbit1q:
we don't, we already have multiple subgroups that we sort individually within the cards element

tooomm:
>we already have multiple subgroups that we sort individually within the cards element

This is not documented in the repo and I don't see comments in the file about it.
Either way there is something to improve here.

1) Place all tokens in the same `<cards>` element and separate any subgroups simply by comments
2) Have several `<cards>` elements to more clearly distinguish such groups

I assume the later also helps with programmatically ensuring an order within each cards element?

As long as it's uniform and clear to everybody (aka documented) I'm fine. The resulting file should also stay easy to work with and validate of course.

SlightlyCircuitous:
I am going to close this in favor of going back to having only one `<cards>` element in the tokens file; see my tokens PR that tries to implement tooomm's first option.

tooomm:
Thanks for pushing this forward, @SlightlyCircuitous

PR #4944, opened 2023-11-18
Update default theme name

tooomm:
If no theme is selected, the default theme is used.
Cockatrice labels the default theme as `None`, which is a bit misleading. There is a standard theme used, not none. :)
image

- Name standard visual theme `Default`, same as basic sound theme

### DIFF
--- a/cockatrice/src/thememanager.cpp
+++ b/cockatrice/src/thememanager.cpp
@@ -9,7 +9,7 @@
 #include <QPixmapCache>
 #include <QStandardPaths>
 
-#define NONE_THEME_NAME "None "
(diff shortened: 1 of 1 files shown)


ebbit1q:
I like default more than "standard", I named it "none" because the theme for it is actually inside the app, not a separate folder.

tooomm:
>I like default more than "standard"

Perfect. Had the same thought and did change it already. 👍

>the theme for it is actually inside the app, not a separate folder.

Jep, this is different for the visual themes compared to the sound ones. But nontheless, I'm in favor of this change.

ebbit1q (approved):
ok

PR #4945, opened 2023-11-19
Fix spacing issue in message log strings

tooomm:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -72,13 +72,13 @@ MessageLogWidget::getFromStr(CardZone *zone, QString cardName, int position, boo
     QString zoneName = zone->getName();
 
     if (zoneName == tableConstant()) {
-        fromStr = tr(" from play");
+        fromStr = tr("from play");
     } else if (zoneName == graveyardConstant()) {
-        fromStr = tr(" from their graveyard");
+        fromStr = tr("from their graveyard");
     } else if (zoneName == exileConstant()) {
-        fromStr = tr(" from exile");
+        fromStr = tr("from exile");
     } else if (zoneName == handConstant()) {
-        fromStr = tr(" from their hand");
+        fromStr = tr("from their hand");
     } else if (zoneName == deckConstant()) {
         if (position == 0) {
             if (cardName.isEmpty()) {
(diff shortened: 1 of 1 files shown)


cockatrice/src/messagelogwidget.cpp
```diff
@@ -251,7 +251,7 @@ void MessageLogWidget::logCreateArrow(Player *player,
 
 void MessageLogWidget::logCreateToken(Player *player, QString cardName, QString pt)
 {
-    appendHtmlServerMessage(tr("%1 creates token: %2%3.")
+    appendHtmlServerMessage(tr("%1 creates token: %2 %3.")
```
ebbit1q, line 254: %3 can be empty, this is why the space isn't here by default

tooomm: Checking the arguments again, `%2%3` seems totally fine here and it's actually not connected to this issue.

Will revert. Thank you!

cockatrice/src/messagelogwidget.cpp
```diff
@@ -325,33 +325,33 @@ void MessageLogWidget::logMoveCard(Player *player,
     if (targetZoneName == tableConstant()) {
         soundEngine->playSound("play_card");
         if (card->getFaceDown()) {
-            finalStr = tr("%1 puts %2 into play%3 face down.");
+            finalStr = tr("%1 puts %2 into play %3 face down.");
         } else {
-            finalStr = tr("%1 puts %2 into play%3.");
+            finalStr = tr("%1 puts %2 into play %3.");
```
tooomm, line 330: Now that I'm reading this again... this two strings are the only ones that have a different placement.

I leave it to the native speakers to tell me if the wording is correct or should be similar to the others:
 `finalStr = tr("%1 puts %2 %3 into play.");`

<br>

"`Player 1` puts `Shock` `from their hand` into play."
over
"`Player 1` puts `Shock` into play `from their hand."`

SlightlyCircuitous: While both options are technically correct, the second option ("`Player 1` puts `Shock` into play `from their hand`") sounds much more natural in English, to me at least.

tooomm: More native speakers opinions, please. :)

If `2)` is the best choice, we should change the other strings accordingly as well. 👍

ZeldaZach: I'm good with either choice, but I do like 2 better

PR #4948, opened 2023-11-28
Support escaping single and double quotes in Deck Editor Search Regex Strings

### LINKED ISSUE #4946: Escape sequences in card search to allow for searching the double-quote character (") etc
Currently it appears that there is no way to do a search that includes the double quotes character `"` via the search bar.

Example:
`o:"creatures you control gain \"{T}: "`
where '\\' is used as an escape character

Note: It **is** currently possible to search for the quotation mark via filters.

ZeldaZach:
## Related Ticket(s)
- Fix #4946

- When searching on text, you couldn't look for quotes as the grammar didn't support this functionality

- Added support for escaping quotes in text searches

## Screenshots
image

### DIFF
--- a/cockatrice/src/filter_string.cpp
+++ b/cockatrice/src/filter_string.cpp
@@ -42,9 +42,10 @@ ColorQuery <- [cC] 'olor'? <[iI]?> <[:!]> ColorEx*
 
 FieldQuery <- String [:] RegexString / String ws? NumericExpression
 
-NonQuote <- !["].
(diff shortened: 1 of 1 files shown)


ebbit1q:
do we have to escape single quotes? to my understanding we only give special meaning to double quotes

ZeldaZach:
> do we have to escape single quotes? to my understanding we only give special meaning to double quotes

We don't _have_ to, but it was easy to add... so why not?

PR #4977, opened 2024-01-04
CI: Fix ignore pattern & highlight status of translation automations

tooomm:
One needs to check the detailed log to grab the related status and PR

- Show the PR status as result of the action run in the workflow overview
- Link to the PR

The results will only be seen when run from master after merge.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -16,7 +16,7 @@ on:
       - '**.md'
       - 'webclient/**'
       - '.github/workflows/web-*.yml'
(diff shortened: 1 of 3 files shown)


tooomm:
Interesting... this triggered the desktop builds, but not the lint one? 🤔 

There seems to be an issue with the `paths-ignore` for desktop-build:
https://github.com/Cockatrice/Cockatrice/blob/master/.github/workflows/desktop-build.yml#L7-L11

ebbit1q (approved):
we'll see what this looks like I guess?

PR #4586, opened 2022-03-04
new connect dialog

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/742 
- Fixes https://github.com/Cockatrice/Cockatrice/issues/344
- Fixes https://github.com/Cockatrice/Cockatrice/issues/2492
- Fixes https://github.com/Cockatrice/Cockatrice/issues/4099
- related https://github.com/Cockatrice/Cockatrice/pull/4464

## Short roundup of the initial problem
passwords are still stored in plain text even when using the new hashed passwords

the register and forgot password dialogs are annoying

the connect dialog closes before connection is established

## What will change with this Pull Request?
- new connect dialog, includes registration and requesting new passwords
- the dialog closes on successful connection and then saves the hashed password

## Screenshots
WIP

PR #4983, opened 2024-01-13
Remember user inputs after failed registration #4982

### LINKED ISSUE #4982: As a user registering for a server, when registration fails, I would like the registration input fields not to be cleared that I do not have to reenter everything

When using the registration feature, the user may have their registration rejected because they might be using invalid characters in a name, or their name might already be taken. When does happens, the user is sent back to the registration input flow. However, all of the fields except for the server name and the server port are cleared. It would be nice if they were not cleared, so that the user only had to fix the single input field that was invalid and caused the registration rejection in the first place. 

This new feature would simply keep the valid input fields' data.

candyapplecorn:
## Related Ticket(s)
- Fixes #4982

Registering, when being rejected, would reset all of the input fields. That makes the user type them all over again, when all they should need to do is type the one that was incorrect.

- now, the code will remember the values that the user entered, so all the user has to do is update the value that is incorrect. (see videos)

More specifically, I decided to make `DlgRegister` a friend class of `AbstractClient`. This way, the registration dialogue can access the protected members of `AbstractClient`. 

This solution keeps the `AbstractClient`'s protected information safe, as it is only allowing the registration dialogue (`DlgRegister`) to access this information. All other parts of the code base that could not access this protected information inside of `AbstractClient` still can't access it.

### BEFORE:

https://github.com/Cockatrice/Cockatrice/assets/5856981/1e9be8d4-8965-4447-acd9-cb5b16b3c36d

### AFTER:

_Notice that the user's information remains in the input forms of `DlgRegister`_

https://github.com/Cockatrice/Cockatrice/assets/5856981/094c98d4-7e56-4e65-bd37-6f4972181ee0

### DIFF
--- a/cockatrice/src/dlg_register.cpp
+++ b/cockatrice/src/dlg_register.cpp
@@ -13,7 +13,13 @@
 #include <QLabel>
 #include <QMessageBox>
 
-DlgRegister::DlgRegister(QWidget *parent, AbstractClient *abstractClient) : QDialog(parent)
+DlgRegister::DlgRegister(QWidget *parent,
+                         QString username,
+                         QString password,
+                         QString email,
+                         QString country,
+                         QString realname)
+    : QDialog(parent)
 
 {
     ServersSettings &servers = SettingsCache::instance().servers();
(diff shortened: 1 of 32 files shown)


cockatrice/src/dlg_register.cpp
```diff
@@ -30,29 +32,30 @@ DlgRegister::DlgRegister(QWidget *parent) : QDialog(parent)
...
     playernameLabel = new QLabel(tr("Player &name:"));
-    playernameEdit = new QLineEdit(servers.getPlayerName());
+    playernameEdit =
+        new QLineEdit(servers.getPlayerName().isEmpty() ? abstractClient->getUserName() : servers.getPlayerName());
     playernameEdit->setMaxLength(MAX_NAME_LENGTH);
     playernameLabel->setBuddy(playernameEdit);
 
     passwordLabel = new QLabel(tr("P&assword:"));
-    passwordEdit = new QLineEdit();
+    passwordEdit = new QLineEdit(abstractClient->password);
```
ebbit1q, line 41: why instead of all this messing around with accessing the client's private fields just pass the needed information to the dialog's constructor?

candyapplecorn: > why not instead of all this messing around with accessing the client's private fields just pass the needed information to the dialog's constructor?

sure, that would work. Pretty sure I wanted to do it that way as well, cannot remember exactly why I decided to stop and try a different way. I believe we would need to modify the remote client (or the abstract client it inherits from) so that the `window_main` file could access those values. 

So I think in the end we would still need to modify some class to give permission to another class to access its data

A better solution still would be to keep the dialogue open and not destroy it until the user has exited out of the flow by either canceling or quitting the program or successfully registering. I did not do this though because I frankly got intimidated by the code base. It's a little bit complicated. You have to follow how the signals are admitted and then find out where the slots are that are listening for the signals. So like somewhere, there are some signals that cause the dialogue register to close. There's gotta be a failure one and a success one and maybe one or two others for errors or who knows what. We would need to find all of those slots that are listening for those signals and modify them to not close the dialogue if the user encounters a registration error. I also think just hitting the submit button on the registration dialogue close is the registration dialogue, so that one would need to be modified as well. This would definitely be the best approach if I could pull it off though because then we wouldn't need to modify any of the accessibility of the class attributes.

ebbit1q: I've been meaning to do that yet still haven't found the time to sit down and get it all sorted, it needs a lot of signals and stuff moved https://github.com/Cockatrice/Cockatrice/pull/4586

my comment here is just regarding the arguments in the constructor of the dialog, no changes to the header of the client will be required, just pass the values when the dialog is instantiated.

candyapplecorn: Sounds like an easy change to make then @ebbit1q . RAII is good. I hope to get around to it not too long from now. Thanks for your advice

candyapplecorn: I've changed the constructor to accept arguments: https://github.com/Cockatrice/Cockatrice/pull/4983/files#diff-1b35d27e75120bfb88322ce520f3c057ab93b345875738aba24e04a8596b3285R16

cockatrice/src/window_main.cpp
```diff
@@ -200,7 +200,15 @@ void MainWindow::actConnect()
...
 void MainWindow::actRegister()
 {
-    DlgRegister dlg(this);
+    // Note the pre-existing form data so that if the user is bounced back to the dialogue, they
+    // won't have to re-enter all the information again.
+    QString username = client->getUserName();
+    QString password = client->password;
+    QString email = client->email;
+    QString country = client->country;
+    QString realname = client->realName;
```
candyapplecorn, line 209: @ebbit1q now `MainWindow` is the `friend class` instead of `DlgRegister`, and these variables are passed to the `DlgRegister` constructor

ebbit1q:
I've made a few changes to your pr:
- I've made the values used by the dialog simply be saved in the mainwindow, this seems like the simplest solution and is easier to make sense of.
- host and port are separately saved now so edits to them are reflected in subsequent registration attempts.
- the country search checks if the given country is empty before doing a list search.

saving the values in the main window isn't better or anything, I can also see some more accessors being added to the client for this part. our current style is to not use direct access to properties, instead being more similar to qt's style and making accessors for properties. for me placing the values in the main window for now does make it simpler to work with #4586 as it does less changes.

tooomm:
@candyapplecorn Did you have a chance to review the last changes?
Any comment? :) I can not really help here.

PR #4985, opened 2024-01-15
CI bug fix - unlink 2to3 before installing python3

candyapplecorn:
See the discord thread here: https://discord.com/channels/314987288398659595/401594186983407616/1196451948413595778

> the builds are failing because of devops stuff.
>I looked into the Mac build and it’s failing because Python is failing to symlink 2to3 since there are multiple Python versions being installed by homebrew and the Catalina image GitHub actions uses 
>Idk why the windows 10 build failed

## Related Ticket(s)

Related to #4982  because i found the issue there

https://github.com/pypa/pipenv/issues/3831

github's 2to3 will be unlinked so that the new 2to3 can be linked

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -239,7 +239,7 @@ jobs:
       # RE: https://discord.com/channels/314987288398659595/401594186983407616/1196451948413595778
       - name: Unlink 2to3
         if: matrix.target == '10.15_Catalina'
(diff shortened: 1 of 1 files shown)


candyapplecorn:
change unlink target

njelich:
This seems to still be an issues looking at the CI results

tooomm:
#4689 is the more recent version of this PR

PR #4986, opened 2024-01-15
CI bug fix - unlink 2to3 before installing python3

candyapplecorn:
See the discord thread here: https://discord.com/channels/314987288398659595/401594186983407616/1196451948413595778

> the builds are failing because of devops stuff.
>I looked into the Mac build and it’s failing because Python is failing to symlink 2to3 since there are multiple Python versions being installed by homebrew and the Catalina image GitHub actions uses 
>Idk why the windows 10 build failed

## Related Ticket(s)

Related to #4982  because i found the issue there

https://github.com/pypa/pipenv/issues/3831

github's 2to3 will be unlinked so that the new 2to3 can be linked

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -239,7 +239,7 @@ jobs:
       # RE: https://discord.com/channels/314987288398659595/401594186983407616/1196451948413595778
       - name: Unlink 2to3
         if: matrix.target == '10.15_Catalina'
-        run: brew unlink /usr/local/bin/2to3
+        run: rm '/usr/local/bin/2to3'
 
       - name: Install dependencies using homebrew
         shell: bash
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -236,13 +236,34 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      # Workaround to fix homebrew on macOS 10.15
       # RE: https://discord.com/channels/314987288398659595/401594186983407616/1196451948413595778
+      # See also PR #4986 and issue #5007, as well as pypa/pipenv#3831
       - name: Unlink 2to3
         if: matrix.target == '10.15_Catalina'
         run: |
           rm '/usr/local/bin/2to3'
           rm '/usr/local/bin/2to3-3.11'
           rm '/usr/local/bin/2to3-3.12'
+          rm '/usr/local/bin/idle3'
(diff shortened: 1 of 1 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -236,6 +236,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      # RE: https://discord.com/channels/314987288398659595/401594186983407616/1196451948413595778
+      - name: Unlink 2to3
+        if: matrix.target == '10.15_Catalina'
+        run: brew unlink /usr/local/bin/2to3
```
tooomm, line 242: The log is stating to remove it. Did you try that?

```suggestion
        run: rm '/usr/local/bin/2to3'
```

tooomm:
I tried to remove stuff as suggested in the logs, but it only leads to more and more other locations that would actually need to get removed. 
That's probably not the right solution to the problem.

@candyapplecorn any other idea to follow up here?

<br>

>the builds are failing because of devops stuff.
I looked into the Mac build and it’s failing because Python is failing to symlink 2to3 since there are multiple Python versions being installed by homebrew **and the Catalina image GitHub actions uses**

The GitHub Action uses a macOS 11 runner (Big Sur) by the way, just that we use an old Xcode SDK to get a build which is targeting 10.15 (Catalina) at the end.

tooomm:
I realized that the issue is solved in the meantime and macOS builds do complete. No tweaks needed from our side I guess.
Windows is also working with my recent change.

Updated your other PR - it builds now as well.

candyapplecorn:
thank you @tooomm

tooomm:
As @njelich just pointed out in your first PR (4985) that the linking issue magically reappeared now without changes on our end.

Opened #5007

tooomm:
Updated the PR with all "unlinks" as a workaround.
It works and the `homebrew` step finishes.

But now, the follow up Qt installation fails...

njelich:
From the error message:

This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-brew-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.

njelich:
Upstream issue:

https://github.com/actions/runner-images/issues/4020

ZeldaZach:
Superseded by 5033

PR #4988, opened 2024-01-18
CI: Utilize new CPUs with more cores for Win/Linux

tooomm:
Slow builds in CI

- Fully utilize available cores in GitHub runners and speed up builds
- Official announcement: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/
- It's not even listed in the [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) yet, but according to the blog post, it's already available since yesterday

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -158,7 +158,7 @@ jobs:
           distro: '${{matrix.distro}}'
         run: |
           source .ci/docker.sh
-          RUN --server --debug --test --ccache "$CCACHE_SIZE" --parallel 2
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
sounds neat, will this actually result in a speed up for windows though?

tooomm:
>will this actually result in a speed up for windows though?

No, it wont. Still wanted to update the number we use there.

Next to compiler cache not working on Windows (#3059),
this reminds that I tried to fix parallel builds and the underlying cmake issue on Windows a while ago (#4615).

PR #4993, opened 2024-01-25
make translation update script more specific in line selection

ebbit1q:
## Related Ticket(s)
- #4544
- conversation at #4973

there is a chance of having an extra line of output from lupdate which messes up our fragile output grabbing logic

also note, the create pull request will soon be updated to v6 in order to avoid the node16 warning:
https://github.com/peter-evans/create-pull-request/issues/2444#issuecomment-1909906572

- we will check for the output by checking more text from the line we are interested in other than just the brackets
- test of this change is here: https://github.com/ebbit1q/Cockatrice/pull/5

### DIFF
--- a/.ci/update_translation_source_strings.sh
+++ b/.ci/update_translation_source_strings.sh
@@ -45,12 +45,14 @@ if ! got="$(lupdate $DIRS -ts "$FILE" | tee /dev/stderr)"; then
 fi
 
 # trim output
-output="${got##*(}" # trim everything before last (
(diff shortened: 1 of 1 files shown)


tooomm (approved):

tooomm:
Thank you ebbit, it worked nicely for our open PR as well: #4973

Also good to point out the old peter-evans action with an outdated Node, dependabot will provide us with the update once available. 👍

ebbit1q:
you're welcome

PR #4995, opened 2024-01-30
Readme: Remove Gitter chat

tooomm:
Gitter chat is no longer used.
Zach removed it from the webpage some time ago already.

- Remove Gitter badge


ebbit1q (approved):
seems fair

PR #4999, opened 2024-02-10
CI: Use windows-2022 image with Visual Studio 17 2022

### LINKED ISSUE #4923: Build on Windows 2022 image
We are running on Windows Server 2019, which is running out of active support shortly.

https://github.com/actions/runner-images#available-images

<br>

Setting [`EnableClServerMode=true`](https://devblogs.microsoft.com/cppblog/cpp-build-throughput-investigation-and-tune-up/#multitooltask-mtt) in combination with MTT could help in speed up our builds.

tooomm:
## Related Ticket(s)
- Fixes #4923

`vcredist` can not be found by CMake/NSIS with VS 2022:
>    CPack: Create package using NSIS
    CPack: Install projects
    CPack: - Install project: Cockatrice [Release]
    CMake Error at D:/a/Cockatrice/Cockatrice/build/cmake_install.cmake:36 (file):
  Error:     file INSTALL cannot find "C:/Program Files/Microsoft Visual
>      Studio/2022/Enterprise/VC/Redist/MSVC/14.38.33135/vcredist_x64.exe": File
      exists.

VS 2022 uses a different naming for the file. Trying to rename in the NSIS file according to this comment did not solve it yet:
https://github.com/Cockatrice/Cockatrice/blob/1715bcb2166f2a505d35329b7e542c227076ba6f/cmake/FindVCredistRuntime.cmake#L12
See also https://learn.microsoft.com/en-us/cpp/windows/redistributing-visual-cpp-files?view=msvc-170#locate-the-redistributable-files

**I'm wondering why the error message still states `vcredist_x64.exe`, while it should be `vc_redist.x64.exe`**.
Not an CMake expert at all here... any pointer?

- this
- and this

### DIFF
--- a/cmake/FindVCredistRuntime.cmake
+++ b/cmake/FindVCredistRuntime.cmake
@@ -24,7 +24,7 @@ if(WIN32)
     get_filename_component(_path ${_path}/../../ ABSOLUTE)
 
     foreach(redist_file ${REDIST_FILE_NAMES})
-      if(EXISTS "${_path}/${REDIST_FILE}")
+      if(EXISTS "${_path}/${redist_file}")
         set(VCREDISTRUNTIME_FOUND "YES")
         set(VCREDISTRUNTIME_FILE ${_path}/${redist_file})
         break()
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -325,6 +325,8 @@ jobs:
       - name: Add msbuild to PATH
         id: add-msbuild
         uses: microsoft/setup-msbuild@v2
+        with:
+          msbuild-architecture: x64
 
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -317,9 +317,9 @@ jobs:
 
     name: Windows ${{matrix.target}}
     needs: configure
-    runs-on: windows-2019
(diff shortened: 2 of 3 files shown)


ebbit1q:
would these changes also mean we'd need to update the wiki instructions for windows?

ebbit1q:
at https://github.com/Cockatrice/Cockatrice/actions/runs/7862117250/job/21451239518#step:6:119 it says it found vcredist_x64.exe this means if(EXISTS "${_path}/${REDIST_FILE}") was true so this file should exist, why it fails to then find it later eludes me, maybe using the tmate step would help in debugging this live on the github runner

cmake/FindVCredistRuntime.cmake
```diff
@@ -24,7 +24,7 @@ if(WIN32)
     get_filename_component(_path ${_path}/../../ ABSOLUTE)
 
     foreach(redist_file ${REDIST_FILE_NAMES})
-      if(EXISTS "${_path}/${REDIST_FILE}")
+      if(EXISTS "${_path}/${redist_file}")
```
tooomm, line 27: Thanks for investigating and finding the simple fix, @ZeldaZach!

tooomm:
>would these changes also mean we'd need to update the wiki instructions for windows?

Zach amended it as well:
https://github.com/Cockatrice/Cockatrice/wiki/Compiling-Cockatrice-(Windows)

PR #5001, opened 2024-02-20
increase version number so we can create a beta

ebbit1q:
now that builds work again I think we should make a beta now but we need to increase the version number first


tooomm (approved):
Well, looks like macOS 10.15 now started to fail again...

PR #5002, opened 2024-02-23
Remove Ubuntu 23.04 Lunar Lobster Build

SlightlyCircuitous:

Ubuntu 23.04 Lunar Lobster [went end of life on January 25, 2024](https://wiki.ubuntu.com/Releases) so we will not need to build for it in future releases. 

Please let me know if you would like me to add a build for non-LTS 23.10 Mantic Minotaur to replace it or just wait for LTS 24.04 Noble Numbat to be released in April to add a new Ubuntu build.

- Removes Lunar build from CI
- Removes Lunar from the release template


ebbit1q (approved):
thank you, it might be nice to have the latest ubuntu but not required, it might like help one person

SlightlyCircuitous:
That's what I was thinking, especially since you can run the 23.04 build of 2.9.0 on 23.10 with no issues (like I am) so the only reason to build for it would be for a new beta. I'll just wait and put in a build for 24.04 once it's reasonably stable.

ebbit1q:
thank you

tooomm:
>Please let me know if you would like me to add a build for
`non-LTS 23.10 "Mantic Minotaur"` to replace it or just wait for 
`LTS 24.04 "Noble Numbat"` to be released in April to add a new Ubuntu build.

Your PR's inspired me to analyze our download numbers a bit (#5006).

I also think we don't need a 23.10 non-LTS build, it's only a short benefit of having the newest non-LTS version as it will quickly be outdated and has no relevance for users any longer.
Especially, when considering the frequency we release new stable versions.
--> 24.04 LTS should be added once available.

The download numbers also show older LTS builds are basically not used, hence I proposed there to only keep like the two newest ones.
Our builds are not targeting server users who might run supported old versions very long, but general end users who want the newer features as well and don't ignore main version updates for years.
Hardware or vendor limitations that prevent updating are basically not existing unlike on macOS or Windows.

<br>

>you can run the 23.04 build of 2.9.0 on 23.10 with no issues (like I am)

I had no clue that works on Ubuntu as well and always thought you have to run the build for your very specific version! Cool.

>I'll just wait and put in a build for 24.04 LTS once it's reasonably stable.

👍

SlightlyCircuitous:
> I had no clue that works on Ubuntu as well and always thought you have to run the build for your very specific version! Cool.

This is not always the case. The biggest sticking point for Ubuntu builds is generally which protobuf package is available in the system repos. 20.04 has libprotobuf17 and 22.04 has libprotobuf23 so you couldn't use those builds interchangeably. But 23.04, 23.10, and 24.04 (at least currently) all have libprotobuf32, so those should all work interchangeably.

PR #5014, opened 2024-03-27
WebClient: refactor protobuf method structure

seavor:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


webclient/src/websocket/commands/session/register.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  const CmdRegister = webClient.protobuf.controller.Command_Register.create(registerConfig);
+
+  const sc = webClient.protobuf.controller.SessionCommand.create({
+    '.Command_Register.ext': CmdRegister
+  });
+
+  webClient.protobuf.sendSessionCommand(sc, raw => {
+    if (raw.responseCode === webClient.protobuf.controller.Response.ResponseCode.RespRegistrationAccepted) {
```
ZeldaZach, line 35: Any reason we call out this one case first here? Could we just combine it into the switch below?

webclient/src/websocket/commands/session/resetPasswordChallenge.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const CmdForgotPasswordChallenge = webClient.protobuf.controller.Command_ForgotPasswordChallenge.create(forgotPasswordChallengeConfig);
+
+  const sc = webClient.protobuf.controller.SessionCommand.create({
+    '.Command_ForgotPasswordChallenge.ext': CmdForgotPasswordChallenge
+  });
+
+  webClient.protobuf.sendSessionCommand(sc, raw => {
+    if (raw.responseCode === webClient.protobuf.controller.Response.ResponseCode.RespOk) {
+      updateStatus(StatusEnum.DISCONNECTED, null);
```
ZeldaZach, line 25: This case is common, can move it before the if statement

webclient/src/websocket/commands/session/resetPasswordRequest.ts
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  const CmdForgotPasswordRequest = webClient.protobuf.controller.Command_ForgotPasswordRequest.create(forgotPasswordConfig);
+
+  const sc = webClient.protobuf.controller.SessionCommand.create({
+    '.Command_ForgotPasswordRequest.ext': CmdForgotPasswordRequest
+  });
+
+  webClient.protobuf.sendSessionCommand(sc, raw => {
+    if (raw.responseCode === webClient.protobuf.controller.Response.ResponseCode.RespOk) {
```
ZeldaZach, line 24: Would it make more sense to do a negative case if (code != okay) and do a quick exit? Would remove indentation and might clear things up to avoid so many nested conditionals

PR #5023, opened 2024-04-24
Add Ubuntu 24.04 "Noble Numbat" Build

SlightlyCircuitous:

Ubuntu 24.04 is set to release on April 25, 2024 and will be LTS for five years so it makes sense to build for it in future releases.

- Adds a dockerfile and a matrix entry for Noble
- Adds Noble to the Release Template

### DIFF
--- a/.ci/UbuntuNoble/Dockerfile
+++ b/.ci/UbuntuNoble/Dockerfile
@@ -1,0 +1,27 @@
+FROM ubuntu:noble
+
+RUN apt-get update && \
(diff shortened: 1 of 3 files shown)


SlightlyCircuitous:
The .deb from Github Actions installs correctly on my noble test VM, Cockatrice opens without incident, and oracle runs appropriately on start up. I can make, save, and load decks, participate in local games, and connect to servers as expected.

ebbit1q (approved):
Excellent, thank you!

PR #5024, opened 2024-04-24
Remove Fedora 38 Build and Add Fedora 40 Build

SlightlyCircuitous:

Fedora 38 will be end of life on May 14, 2024 so we will not need to build for it in future releases. Any Fedora users staying on 38 have an RPM of 2.9.0 available to them if needed.

Fedora 40 was released on April 23, 2024 and will be supported until May 13, 2025 so it makes sense to start building for it instead.

- Removes the dockerfile and matrix entry for Fedora 38
- Adds a dockerfile and a matrix entry for Fedora 40
- Revises the Release Template to reflect the above changes


SlightlyCircuitous:
The .rpm from Github Actions installs correctly on my Fedora 40 VM, Cockatrice opens without incident, and oracle runs appropriately on start up. I can make, save, and load decks, participate in local games, and connect to servers as expected.

ZeldaZach (approved):

PR #5025, opened 2024-04-29
Addresses crash scenario where too many cards were being loaded in a game

ZeldaZach:
- Too many iterations that fill up the server's memory trying to (1) render the deck hash and (2) building up a string of the deck contents


ebbit1q:
I think instead of checking in the hashing algorithm we should use limits.h in the earliest location, right after receiving a deck.

PR #5033, opened 2024-05-14
Drop MacOS 10.15/11 Support :(

### LINKED ISSUE #4924: macOS 10.15 failing
Our macOS 10.15 build is failing because Homebrew removed [protobuf](https://formulae.brew.sh/formula/protobuf) (and now also [protobuf@21](https://formulae.brew.sh/formula/protobuf@21)) bottles for it.

Maybe we have to drop that old (and unsupported) OS now?
I'm aware that some users might still run super old hardware and others don't update their OS for whatever reason. We have the recent v2.9.0 release ready for them to use, but self compiling could also be an option in the future.

### LINKED ISSUE #5007: CI: macOS 10.15 builds failing when installing protobuf dependencies
```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.11
Target /usr/local/bin/2to3-3.11
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.11'
...
```

More information can be found in #4986 where @candyapplecorn tried to fix this already.

In general this could be related to #4924 and others where we had issues already with protobuf builds no longer being available for older macOS versions on homebrew.

ZeldaZach:
## Related Ticket(s)
- Closes #5007
- Closes #4924

- GitHub Runners are removing MacOS 11 support next month, which removes our ability to generate MacOS 10.15 and MacOS 11 builds. 

- Update Mac Builds to use macOS 12,~~13~~,14

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -243,6 +243,7 @@ jobs:
         # neither of these works: mariadb-connector-c mysql-connector-c++
         env:
           install_qt: ${{matrix.qt_version}}
+          force_overwrite_python: ${{matrix.force_overwrite_python}}
           use_old_protobuf: ${{matrix.use_old_protobuf}}
           HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
         run: |
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -219,7 +219,7 @@ jobs:
             do_tests: 1
             make_package: 1
 
-    name: macOS ${{matrix.target}}
+    name: macOS_${{matrix.target}}
     needs: configure
     runs-on: ${{matrix.os}}
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
@@ -302,7 +302,7 @@ jobs:
             qt_tools: "tools_opensslv3_x64"
             qt_modules: "qtmultimedia qtwebsockets"
 
-    name: Windows ${{matrix.target}}
+    name: Windows_${{matrix.target}}
     needs: configure
     runs-on: windows-2022
     env:
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -249,7 +249,6 @@ jobs:
         run: |
           brew update
           if [[ $force_overwrite_python == 1 ]]; then
-            brew install --overwrite python@3.12
             brew link --force python@3.12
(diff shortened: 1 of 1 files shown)


tooomm:
The new M1 runners are fast!

We can bump the used XCode versions and still stay on the same SDK:
macOS 12 --> 14.0.1
macOS 13 --> 14.3.1
macOS 14 --> 15.4 (Currently, with 14.3.1 it's still macOS 13, at minimum 15.0.1 would be needed for a macOS 14 target)

macOS 14 image runner are ARM M1 only (no more Intel option available for free tiers). Does that mean the package would not work on older Intel Mac's running macOS 14?

But it would be a good reason to have a native ARM binary ready.
On the other hand, I see no reason to have a binary for macOS 13?
There is no downside in using the one targeting 12.

tooomm (approved):
Sure, builds are looking ok.

I'm not sure if the updater needs additional changes before considering a new release though, especially with the M1 binary.

fwcd:
> GitHub Runners are removing MacOS 11 support next month, which removes our ability to generate MacOS 10.15 and MacOS 11 builds.

Just FYI, you can still target older macOS versions (and also both x64/arm64) when building with the most recent Xcode SDK (for the App Store that's even a requirement). Off top of my head, it should be as simple as setting the deployment target to the oldest macOS version you wish to support and the compiler should handle the rest.

tooomm:
Can you explain a bit more?

Not a Mac expert... We always went by the official list of available SDKs and selected the Xcode version to compile with:
https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#installed-sdks

fwcd:
Sure, what I mean is that you should (in principle) be able to pick the latest macOS runner and latest Xcode, then set your [`CMAKE_OSX_DEPLOYMENT_TARGET`](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html) (or, given that you're building dependencies with vcpkg, [`VCPKG_OSX_DEPLOYMENT_TARGET`](https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_osx_deployment_target) in a custom triplet) to whichever oldest macOS version you wish to support.

At Mixxx (a FOSS DJ program), we have [custom triplets](https://github.com/mixxxdj/vcpkg/tree/2.5/overlay/triplets) for building our dependencies with a macOS 11 deployment target, even though our runners are newer.

Issue #4925, opened 2023-10-25
macOS 12/13 occasionally failing


tooomm:
macOS 12 example:
https://github.com/Cockatrice/Cockatrice/actions/runs/6614690847/job/17965213572?pr=4922
```
  CPack: - Run preinstall target for: Cockatrice
  CPack: - Install project: Cockatrice []
  CPack: Create package
  CPack Error: Error executing: osascript "/Users/runner/work/Cockatrice/Cockatrice/cmake/CMakeDMGSetup.script" "Cockatrice"
  CPack Error: Error executing custom script on disk image.
  osascript: can't open default scripting component.
  
  CPack Error: Problem compressing the directory
  CPack Error: Error when generating package: Cockatrice
  make: *** [package] Error 1
  Error: Process completed with exit code 2.
```

macOS 13 example:
https://github.com/Cockatrice/Cockatrice/actions/runs/6632768476/job/18019167204?pr=4871
```
  CPack: - Run preinstall target for: Cockatrice
  CPack: - Install project: Cockatrice []
  CPack: Create package
  CPack Error: Error executing: /usr/bin/hdiutil create -ov -srcfolder "/Users/runner/work/Cockatrice/Cockatrice/build/_CPack_Packages/OSX/DragNDrop/Cockatrice-2.9.0-custom(6f88419)" -volname "Cockatrice" -fs "HFS+" -format UDRW "/Users/runner/work/Cockatrice/Cockatrice/build/_CPack_Packages/OSX/DragNDrop/temp.dmg"
  CPack Error: Error generating temporary disk image.
  hdiutil: create failed - Resource busy
  
  CPack Error: Problem compressing the directory
  CPack Error: Error when generating package: Cockatrice
  make: *** [package] Error 1
  Error: Process completed with exit code 2.
```

<br>

A rerun can sometimes fix the issue.

tooomm:
Unfortunately, #4961 is not solving this just yet.

See the CI run on master after merging:
- First run failing: https://github.com/Cockatrice/Cockatrice/actions/runs/7229159667/attempts/1
- Second run failing: https://github.com/Cockatrice/Cockatrice/actions/runs/7229159667/attempts/2
- Third run succeeding: https://github.com/Cockatrice/Cockatrice/actions/runs/7229159667/attempts/3

Also see Elarnons note in his PR. I had to rerun it as well to get a successful build for others to test.

ZeldaZach:
[...]

PR #5040, opened 2024-06-03
CI: Update build matrix & clean naming

tooomm:
## Related Ticket(s)
- Related #5033

Outdated build matrix with failing macOS build.

- Uniform & clean job and artifact names across builds
- Rename builds to use intuitive version numbers over names for all operating systems
- Update builds
  - Remove EoL Ubuntu 18
  - Native "M-chip" build as default for Mac, Intel build for older hardware
  - Remove EoL macOS 12 (GitHub runner to be decommissioned next month)
    image
  - Runs based on `macos-13` image [fail sometimes](https://github.com/Cockatrice/Cockatrice/actions/runs/11539857625/job/32120004884), re-runs mostly succeed.
  We saw this earlier already, see https://github.com/Cockatrice/Cockatrice/pull/5040#issuecomment-2200718060
- Several small adjustments

<br>

<b>List of jobs & artifacts:</b>
| Currently     | After         |
| ------------- | ------------- |
| image | image |

<b>New base images for macOS:</b>
- <kbd>Intel (x86_64)</kbd>: macOS 12 --> macOS 13
- <kbd>Apple M (ARM)</kbd>: macOS 14 --> macOS 13


ebbit1q (commented):
these changes seem fine, when will this pr be ready for review?

ZeldaZach:
Note: Macos 14 images are broken

tooomm:
> Note: Macos 14 images are broken

Actually it's only https://github.com/Cockatrice/Cockatrice/issues/4925 I think, which still awaits an upstream fix, see the ticket:
https://gitlab.kitware.com/cmake/cmake/-/issues/25671

We can only wait I guess :)

<br>

> these changes seem fine, when will this pr be ready for review?

Wanted to have another look at it, double check and summarize the changes. Not managing currently, will let you know of course!

Also, since I'm not sure the updated still works with Zach's last changes, there is a chance to fix it here.
With the new beta releases out, people should be able to tell easily.

ebbit1q:
this pr seems to fix all the issues blocking a beta release, is this ready for review yet?

.ci/UbuntuBionic/Dockerfile
```diff
@@ -1,25 +0,0 @@
-FROM ubuntu:bionic
```
ebbit1q, line 1: did the bionic build stop working?

.ci/release_template.md
```diff
@@ -8,18 +8,17 @@ git push -d origin --REPLACE-WITH-BETA-LIST--
...
  - <kbd>Ubuntu 24.04 LTS</kbd> ("Noble Numbat")
- - <kbd>Debian 11</kbd> ("Bullseye")
+ - <kbd>Ubuntu 22.04 LTS</kbd> ("Jammy Jellyfish")
+ - <kbd>Ubuntu 20.04 LTS</kbd> ("Focal Fossa")
  - <kbd>Debian 12</kbd> ("Bookworm")
+ - <kbd>Debian 11</kbd> ("Bullseye")
+ - <kbd>Fedora 40</kbd> 
  - <kbd>Fedora 39</kbd>
- - <kbd>Fedora 40</kbd>
 <kbd>We are also packaged in Arch Linux's official community repository, courtesy of @FFY00</kbd></i>
```
ebbit1q, line 22:
```suggestion
<kbd>We are also packaged in Arch Linux's official "extra" repository, courtesy of @FFY00</kbd></i>
```
arch changed the name of their community repo to extra

.github/workflows/desktop-build.yml
```diff
@@ -244,16 +251,22 @@ jobs:
...
-        # macOS runner actually have only 3 cores
-        # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
-        run: .ci/compile.sh --server --parallel 3
+          PACKAGE_SUFFIX: '-macOS${{matrix.target}}_${{matrix.soc}}'
+        # macOS runner have 3 cores usually - only the macos-13 image has 4:
+        # https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
+        # https://github.com/actions/runner-images?tab=readme-ov-file#available-images
+        run: |
+          if [[ ${{ matrix.os }} == macos-13 ]]; then
+            .ci/compile.sh --server --parallel 4
```
ebbit1q, line 260: please put this in the build matrix instead

tooomm:
Looking at our [updater logic](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/client/network/release_channel.cpp#L44C1-L59C76), it probably matches something on macOS currently. But I'm not sure users on Intel Macs might get a version delivered targeting ARM silicon and vice versa after the last change.

With this PR, Mac releases use the same naming scheme as all other operating systems, which the updater code does not support yet.
Mac binaries did use the unique pattern `macOS-osversion_osname.dmg`, which did also not indicate which of the two incompatible CPU types they are targeting:
image
Now it's the uniform `macOS13_Apple.dmg`/`macOS13_Intel.dmg`.

Open change here would be to adjust the default dmg to target Apple CPU without having an appendix, and only the Intel one receives an `_Intel` label. The Intel binaries would be obsolete at one point.
Same as we do not append `-64bit` for Windows anymore, as it's the default. We only had a 32bit label for the time it was needed.

Since the updater function needs a rework for the Intel and Apple CPU variants anyway on Mac, we could
- Update the logic
  - Understand new naming scheme (no backwards compatibility needed)
    `macOS$VERSION.dmg` --> Apple M CPU
    `macOS$VERSION_Intel.dmg` --> Intel CPU
  - Check if running on Intel/Apple silicon
  - Pull locally matching versions only
  - (Windows check for 32bit can also be removed)
- Create a new full release
- Merge this PR & move on

Thoughts?

.ci/UbuntuBionic/Dockerfile
tooomm: Not sure.
But it's EOL for 19 months.

.github/workflows/desktop-build.yml
tooomm: You mean the comment/links on the available images?

.github/workflows/desktop-build.yml
```diff
@@ -198,24 +201,28 @@ jobs:
       fail-fast: false
       matrix:
```
ebbit1q, line 202: this is the build matrix

.github/workflows/desktop-build.yml
ebbit1q: https://github.com/Cockatrice/Cockatrice/pull/5040/files#r1820632104

ebbit1q: I specifically mean the core count differentiation

.github/workflows/desktop-build.yml
```diff
@@ -198,24 +201,28 @@ jobs:
...
             make_package: 1
 
-          - target: 14_Sonoma_Debug
+          - target: 13
+            soc: Apple
             os: macos-14
-            xcode: "15.4"
+            xcode: "14.3.1"
             type: Debug
+            make_package: 1
```
ebbit1q, line 223: we don't want to release these

PR #5046, opened 2024-06-07
Added devcontainer Dockerfile and json

aayu3:
## Related Ticket(s)
- N/A

Added devcontainer support for people who want to develop on windows. Will add readme if contributors agree this is good

- .devcontainer file in root of project

### DIFF
--- a/.devcontainer/devcontainer.Dockerfile
+++ b/.devcontainer/devcontainer.Dockerfile
@@ -1,0 +1,55 @@
+# Use an official Ubuntu as a parent image
+FROM ubuntu:22.04
+
+# Set the environment variable to avoid interactive prompts during package installation
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Update package lists and install dependencies
+RUN apt-get update && \
+    apt-get install --yes --no-install-recommends \
+        build-essential \
+        ccache \
+        clang-format \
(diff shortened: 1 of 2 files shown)


ebbit1q:
I don't see any issue with having this but I'm also not familiar with development on windows, it seems kind of strange to me to need a linux container in order to develop, we should be able to compile and use the application on windows natively.

How many developers would use this and what if this container becomes outdated?

aayu3:
> I don't see any issue with having this but I'm also not familiar with development on windows, it seems kind of strange to me to need a linux container in order to develop, we should be able to compile and use the application on windows natively.
> 
> How many developers would use this and what if this container becomes outdated?

Personally I'm used to developing in containers as best practice from classes at Uni. I think the main advantage is how quickly you can setup the dev container. Also dependency isolation is nice, I was having some issues when setting up locally since I think I messed up my PATH when setting up anaconda locally. The ubuntu version is specified, but I'll update the .devcontainer to specify the version for desktop-lite.

ZeldaZach:
Hey @aayu3 is this something you'd still like to see added? Do you plan on contributing to the project further if/once this is merged through? I'd love to have another hand around these parts, and I'd be happy to merge something like this through if it's a necessity for ya :)

PR #5047, opened 2024-06-07
Allow up to 100 dice to be rolled at a time

### LINKED ISSUE #4276: add option to throw multiple dice
Sometimes you'd like to throw multiple dice at the same time, without being ratelimited etc.
I propose an additional value to the roll die command and event that simply specifies an amount of them, this should then be interpreted either server or clientside to translate a single rand to multiple values. note that the maximum random number currently geneneratable is half an int64. see https://github.com/Cockatrice/Cockatrice/blob/master/common/rng_sfmt.cpp

**Additional Context**
While a rare effect in magic it certainly is possible to have many die or coin related abilities trigger at the same time, other games could benefit as well.
This would be a small addition to the protocol and maintain backwards compatibility.

ZeldaZach:
- Fix #4276
- With contributions from @aayu3


cockatrice/src/messagelogwidget.cpp
```diff
@@ -558,18 +558,21 @@ void MessageLogWidget::logReverseTurn(Player *player, bool reversed)
...
-        QString coinOptions[2] = {tr("Heads") + " (1)", tr("Tails") + " (2)"};
-        appendHtmlServerMessage(tr("%1 flipped a coin. It landed as %2.")
-                                    .arg(sanitizeHtml(player->getName()))
-                                    .arg("<font class=\"blue\">" + coinOptions[roll - 1] + "</font>"));
-    } else {
-        appendHtmlServerMessage(tr("%1 rolls a %2 with a %3-sided die.")
-                                    .arg(sanitizeHtml(player->getName()))
-                                    .arg("<font class=\"blue\">" + QString::number(roll) + "</font>")
-                                    .arg("<font class=\"blue\">" + QString::number(sides) + "</font>"));
+    for (const auto &rollString : rolls.split(",")) {
```
ebbit1q, line 563: I think it'd be a lot cleaner if we were to compound these messages, for one die keep it the same but for multiple I think something like "player rolls 3 20 sided dice: 11, 14, 20" the list would be sorted.

ZeldaZach: I'm not sure I like the idea of sorting the list, but I'm okay with compounding them

cockatrice/src/player.cpp
```diff
@@ -1962,7 +1963,16 @@ void Player::eventShuffle(const Event_Shuffle &event)
...
+    QStringList stringResults;
+    for (const auto &i : event.values()) {
+        stringResults.append(QString::number(i));
+    }
+    if (event.value()) {
+        // Backwards compatibility for old clients
+        stringResults.append(QString::number(event.value()));
+    }
+
+    emit logRollDie(this, static_cast<int>(event.sides()), stringResults.join(","));
```
ebbit1q, line 1975: why make this a string? we could just have a list so we don't need to parse this thing down the line again...

common/pb/command_roll_die.proto
```diff
@@ -5,4 +5,5 @@ message Command_RollDie {
         optional Command_RollDie ext = 1005;
     }
     optional uint32 sides = 1;
+    optional uint32 dice_to_roll = 2;
```
ebbit1q, line 8: I think the name "dice_to_roll" isn't a very clear identifier, why not just "amount" or "count"?

ebbit1q: also, might be nice to have a default = 1 here

common/server_player.cpp
```diff
@@ -1069,9 +1069,15 @@ Server_Player::cmdRollDie(const Command_RollDie &cmd, ResponseContainer & /*rc*/
...
 
+    const auto validatedSides = static_cast<int>(std::min(std::max(cmd.sides(), MINIMUM_DIE_SIDES), MAXIMUM_DIE_SIDES));
+    const auto validatedDiceToRoll =
+        static_cast<int>(std::min(std::max(cmd.dice_to_roll(), MINIMUM_DICE_TO_ROLL), MAXIMUM_DICE_TO_ROLL));
+
     Event_RollDie event;
-    event.set_sides(cmd.sides());
-    event.set_value(rng->rand(1, cmd.sides()));
+    event.set_sides(validatedSides);
+    for (auto i = 0; i < validatedDiceToRoll; ++i) {
```
ebbit1q, line 1078: would it be worth optimising this? one of these rand calls throws away a lot of the "entropy" for one roll, multiple dice could be randomised using just one rand call and a %

common/server_player.h
```diff
@@ -203,7 +208,7 @@ class Server_Player : public Server_ArrowTarget, public ServerInfo_User_Containe
     Response::ResponseCode cmdGameSay(const Command_GameSay &cmd, ResponseContainer &rc, GameEventStorage &ges);
     Response::ResponseCode cmdShuffle(const Command_Shuffle &cmd, ResponseContainer &rc, GameEventStorage &ges);
     Response::ResponseCode cmdMulligan(const Command_Mulligan &cmd, ResponseContainer &rc, GameEventStorage &ges);
-    Response::ResponseCode cmdRollDie(const Command_RollDie &cmd, ResponseContainer &rc, GameEventStorage &ges);
+    Response::ResponseCode cmdRollDie(const Command_RollDie &cmd, ResponseContainer &rc, GameEventStorage &ges) const;
```
ebbit1q, line 211: are all these other functions also const-able?

cockatrice/src/messagelogwidget.cpp
ebbit1q: sorting would help so much with parsing the results though, if the order of the dice were important then using multiple separate rolls is still an option

common/server_player.h
```diff
@@ -86,6 +86,11 @@ class Server_Player : public Server_ArrowTarget, public ServerInfo_User_Containe
     void revealTopCardIfNeeded(Server_CardZone *zone, GameEventStorage &ges);
 
 public:
+    static constexpr uint MINIMUM_DIE_SIDES = 2;
+    static constexpr uint MAXIMUM_DIE_SIDES = 1000000;
+    static constexpr uint MINIMUM_DICE_TO_ROLL = 1;
+    static constexpr uint MAXIMUM_DICE_TO_ROLL = 100;
```
ebbit1q, line 92: I think instead of here these should go in the limits.h file

common/pb/command_roll_die.proto
ZeldaZach: Defaults aren't allowed in proto3, so i'm gonna skip that. I feel the name is also correct as-is, as both "amount" and "count" are ambiguous.

common/server_player.cpp
ZeldaZach: Agreed, will do

common/server_player.h
ZeldaZach: Possibly, but out of scope for PR

common/pb/command_roll_die.proto
ebbit1q: I disagree, "dice to roll" is ambiguous, it could be amount of dice, the fact that it is going to roll the dice is evident from the fact that it's in a command literally called roll die

common/server_player.cpp
ebbit1q: did you change anything here?

ZeldaZach: I looked to change it up, and it made it overly complex. I'm not sure what you're looking for on this

common/pb/command_roll_die.proto
ZeldaZach: Fair enough. I'll move it to "count" then

cockatrice/src/dlg_roll_dice.cpp
```diff
@@ -0,0 +1,49 @@
+#include "dlg_roll_dice.h"
+
+#include <QDialogButtonBox>
+#include <QIntValidator>
+#include <QLabel>
+#include <QVBoxLayout>
+#include <QWidget>
+#include <server_player.h>
```
ebbit1q, line 8: this include should now no longer be needed

cockatrice/src/dlg_roll_dice.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <trice_limits.h>
+
+DlgRollDice::DlgRollDice(QWidget *parent) : QDialog(parent)
+{
+    numberOfSidesLabel = new QLabel(tr("Number of sides:"));
+    numberOfSidesEdit = new QLineEdit(QString::number(DEFAULT_NUMBER_SIDES_DIE));
```
ebbit1q, line 14: why aren't you using a qspinbox?

cockatrice/src/player.cpp
```diff
@@ -1601,15 +1602,15 @@ void Player::actUntapAll()
 
 void Player::actRollDie()
 {
-    bool ok;
-    int sides = QInputDialog::getInt(game, tr("Roll die"), tr("Number of sides:"), defaultNumberDieRoll, minDieRoll,
```
ebbit1q, line 1605: will the new dialog function the same as the old getInt dialog? ie focus the number of sides and accept on hitting enter?

cockatrice/src/dlg_roll_dice.cpp
ZeldaZach: Didn't know it existed, replaced

cockatrice/src/player.cpp
ZeldaZach: Added

ebbit1q (approved):
thank you

PR #5053, opened 2024-06-16
Allow Judges to see all information, regardless of room settings

ZeldaZach:


common/server_game.cpp
```diff
@@ -287,18 +287,19 @@ void Server_Game::sendGameStateToPlayers()
...
-            gec = prepareGameEvent(spectatorEvent, -1);
-        else {
+        if (player->getSpectator()) {
+            if (spectatorsSeeEverything || player->getJudge()) {
+                gec = prepareGameEvent(omniscientEvent, -1);
+            } else {
+                gec = prepareGameEvent(spectatorNormalEvent, -1);
+            }
+        } else {
             Event_GameStateChanged event;
```
ebbit1q, line 303: I like the way this is cleaner now, for consistency could we move this declaration to the place the other one is?

ebbit1q: actually no, this event is unique to the player isn't it?

common/server_response_containers.cpp
```diff
@@ -41,8 +41,10 @@ void GameEventStorage::setGameEventContext(const ::google::protobuf::Message &_g
 void GameEventStorage::enqueueGameEvent(const ::google::protobuf::Message &event,
                                         int playerId,
                                         GameEventStorageItem::EventRecipients recipients,
-                                        int _privatePlayerId)
+                                        int _privatePlayerId,
+                                        bool _overwriteOwnership)
```
ebbit1q, line 45: on looking at the way setForcedByJudge is implemented I'd like to see this implemented consistently, please remove _overwriteOwnership from the enqueGameEvent function and instead have cmdRevealCards use ges.setOverwriteOwnership

ebbit1q (approved):
very nice, please merge this after addressing my one remark (moving one line of code)

I think we should try to make a beta with this as soon as possible.

common/server_game.cpp
ZeldaZach: Correct this is unique to the player

PR #5060, opened 2024-06-17
Re-Add ability to share editable deck views

### LINKED ISSUE #625: Cannot Grab cards out of opponents library
A player can no longer grab cards out of opponents library.  This used to work.

ZeldaZach:
Allows cards like [Praetor's Grasp](https://scryfall.com/card/nph/71/praetors-grasp) to be functional

- Rolls back 68118191619b26ffa908c86679c51c4db0d61125
- Follow up to adbb607700908f43d52cfa988eed8101683db22e

## Related Ticket(s)
- Fixes #625

### DIFF
--- a/cockatrice/src/messagelogwidget.cpp
+++ b/cockatrice/src/messagelogwidget.cpp
@@ -481,7 +481,8 @@ void MessageLogWidget::logRevealCards(Player *player,
                                       QString cardName,
                                       Player *otherPlayer,
                                       bool faceDown,
-                                      int amount)
+                                      int amount,
+                                      bool isLentToAnotherPlayer)
 {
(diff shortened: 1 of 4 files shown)


ebbit1q (approved):
tested and seems to work excellent, backwards compatibility is good, only the messagelog is still reveal for them.

note that when a deck is lent to you you can only move cards to play on your side, for praetor's grasp you'd have to hold shift and drag so people don't see the card you take. next to this people will not be told the name of the card you take when face down but they will be told which card nth from the top is taken (the message log will only say bottom or top but the info is there) so before lending your deck don't forget to shuffle and make sure your opponent shuffles before showing.

PR #5079, opened 2024-07-13
[4191] fix: Move unfocusTextBox and aFocusChat shortcuts to Player group

### LINKED ISSUE #4191: The Shift + Return isn't being checked as a possible duplicate shortcuts
The Shift + Return isn't being checked as a possible duplicate shortcuts. Other combinations apparently are ok.
This caused a false impression that this shortcut wasn't working for a custom bind (as it was already bound to a different thing).

LumaddT:
## Related Ticket(s)
- Fixes #4191

The check for duplicate shortcuts is inconsistent when it comes to chat focus and unfocus. These actions can be performed during a game, but their shortcuts can also be assigned to game actions. A situation like the one in the screenshot is undesirable because, in game, `Ctrl+Alt+Shift+Q` will focus and unfocus the chat, but will not open the dialogue to set yellow counters.

The check for duplicates is done within groups of shortcuts. A shortcut can´t be applied to two actions within the same group, and no shortcut can be the same as any in the `MainWindow` family. The groups are based on in which tab the shortcut would do something. 4 of the groups (`DlgLoadDeckFromClipboard`, `tab_room`, `tab_game`, `Textbox`) contain a single action. The problem ones are `Textbox/unfocusTextBox` and `tab_game/aFocusChat`.

There's nothing special about the `Shift + Return` shortcut specifically.

The change eliminates the `Textbox` and `tab_game` families and moves those actions to the `Player` one, which includes all the shortcuts that can be used in game. Nothing seems to break, except for the fact that now focus and unfocus must be on two different shortcuts while before they could be assigned to the same.

## Screenshots
image


ebbit1q:
this would remove player's previously saved shortcut settings, is it possible to change the check for duplicates instead?

LumaddT:
Not as pretty, but it gets the job done and allows to have the same shortcut for chat focus and unfocus. The issue with this configuration is that now the default shortcuts are, technically, illegal: `Player/aCloseMostRecentZoneView` (Close Recent View) and `Textbox/unfocusTextBox` (Unfocus Text Box) are both bound to `Esc` by default. A player would not be able to do it if that wasn't the default configuration. It It's not a problem per se but it's awkward.

ebbit1q:
did that limitation not exist with your previous solution? if you think that moving the shortcuts is a better solution then I do propose we install a migration check as well, if the old keys are found in the ini fill we add the new ones so people can keep their previously assigned shortcuts.

ebbit1q:
also, I think there might already be some more conflicting shortcuts by default.

LumaddT:
> did that limitation not exist with your previous solution? 

Moving both to the Player category wouldn't allow to have that toggle effect. The illegal default configuration did indeed happen with the first solution as well. I just didn't think about it before.

> if you think that moving the shortcuts is a better solution then I do propose we install a migration check as well, if the old keys are found in the ini fill we add the new ones so people can keep their previously assigned shortcuts.

Depends on what the purpose of the categories is. Are they intended to be logical separations that make sense, or are they just meant to be groups of shortcuts that shouldn't conflict with each other? If it's the first, the first solution conflicts with it I think. If it's the second it works just fine.

I did think about a migration, but I wasn't sure when it should apply. The most logical is during installation but that probably requires touching the NSIS file. It seems doable looking at the documentation but I'm no NSIS expert. On startup works too and is probably easier, but while it's a very cheap operation it feels wasted to apply it every single time the program starts. Like bad practice. I guess I'm overthinking it.

I think that the best solution from the user point of view is the current one, because it doesn't take functionality away (i.e. one can still set the same shortcut to both actions). In terms of readability and maintainability the first one, with a migration in the installer, would probably be better long term.

ebbit1q:
adding such migrations to the installer is not portable, it has to be on startup, it's affect on startup is minimal, there are much more expensive things a graphical program does, not to mention the database loading.
(check the logs for this line sometime `[CardDatabase] loadCardDatabase(): Path = "/home/ebbit/.local/share/Cockatrice/Cockatrice/cards.xml" Status = 0 Cards = 29990 Sets = 760 "879ms"`)

tooomm:
The app does detect when it got updated and shows a hint on first new start to update the database if I'm not mistaken.

Only then it would need to run, and only when coming from an version older than 2.9.1

LumaddT:
Thank you @tooomm. Running it only after an update makes me feel a lot better about it.

Went back to moving both under the Player group, with a migration that runs when the client detects it has been updated. It works well, to the extent that I tested it.

LumaddT:
I can't reproduce the build issue on my machine so I assume the problems with Windows7, Debian11, UbuntuBionic and UbuntuFocal are because they use Qt 5. I read that including `QDebug` may be necessary. `window_main` does not include `QDebug` but it doesn't seem to be having issues.

ebbit1q:
this would be because qdebug might be included already in qt6 in another included header

ZeldaZach (approved):
LGTM, just needs a rebase

LumaddT:
> just needs a rebase

Not yet fully into the workflow of contributing to open source projects. Am I supposed to do the rebase or you?

tooomm:
We are very happy if you can take care of updating your PR.

But if support is needed - let us know!

LumaddT:
The rebase was less painful than expected. CLion managed everything on its own. The `macOS 13 Intel` build fails but that seems an issue with master or GitHub itself.

tooomm:
mac Intel build finished in our CI, probably the macOS builds are still flaky. I think there is an upstream ticket for this.

The migration will run for all future updates?

LumaddT:
> The migration will run for all future updates?

Yes. That's what the edit to to `window_main.cpp` does. Every time it detects an update it runs the migration function right after prompting the user to update the card database. If there's nothing to migrate the function doesn't to anything.

ebbit1q (approved):
excellent, thank you

PR #5092, opened 2024-08-04
add qtimageformats module

ebbit1q:
## Related Ticket(s)
- Fixes #IssueNumber

it seems like qt6 has moved the ability to load some image formats, notably webp to a separate module, our qt6 builds did not include this module and therefor could not read webp images

- adds module dependency to aqt install of windows 10 build
- adds qt6-image-formats-plugins to apt and rpm depends in cmakelists
- removes redundant openssl tool install through aqt, it is now installed through vcpkg

## Screenshots

image
shown are custom build contents for windows 7 and windows 10 builds with this change, compared to latest beta

I did not test this and only assumed it is working through checking which dlls are present, installing the module through the package manager on my machine allows cockatrice to read webp as expected


ebbit1q:
new ubuntu package tested, it is possible the rpm deps list could also use this addition

tooomm:
> new ubuntu package tested, it is possible the rpm deps list could also use this addition

@SlightlyCircuitous Could you test the rpm and let us know, please?

I did test Windows 7 and 10 builds:
10 works with WebP ✅ 
7 gave me an TLS error on the https URL (Oracle throws also an error on SSL)

```
PictureLoader: [card: "WebP Test" set: "test"]: Trying to load picture
PictureLoader: [card: "WebP Test" set: "test"]: No custom picture, trying to download
PictureLoader: [card: "WebP Test" set: "test"]: Trying to fetch picture from url "https://www.gstatic.com/webp/gallery/2.webp"
QSslSocket::connectToHostEncrypted: TLS initialization failed
QSslSocket::connectToHostEncrypted: TLS initialization failed
PictureLoader: [card: "WebP Test" set: "test"]: Download failed for url "https://www.gstatic.com/webp/gallery/2.webp" ("TLS initialization failed")
```

SlightlyCircuitous:
I tested this in a Fedora 40 VM with the 2.9.1-beta2 build. I needed to manually install `qt6-qtimageformats` in order to get a .webp image to work in the custom image folder, so I imagine that will need to be added to the rpm deps list.

ebbit1q:
@tooomm please compare the old build for win7 and try what happens when you copy the openssl and libcrypto dlls from one to the other (you can use portable installs for this)

tooomm:
`v2.9.1-beta.2` same issue
`v2.9.1-beta` same issue
`v2.9.0` works

One difference that I can spot right away is:
The working v2.9.0 has libssl (and crypto) 3 AND 1-1
image
The not working ones do only have 3
image

Copying over the libssl1-1 from 2.9.0 to this PR build does not solve it. Copying over the libcrypto1-1 as well make it work.

I always tried the Win7 installers (Qt5) only.
For v2.9.0 I did only consider 64bit variant.

We had SSL issues several times already.
Here, it looks like the Win7 version tries to use the old OpenSSL and not the new one.

tooomm (approved):

.github/workflows/desktop-build.yml
```diff
@@ -273,15 +273,13 @@ jobs:
       matrix:
         include:
           - target: 7
-            qt_version: 5.15.*
+            qt_version: 5.15.14
```
tooomm, line 276: If this change is reverted, we are good to merge IMHO. 👍 

We have a new issue to track the SSL incompatibility rooted in unsupported Qt5.

ebbit1q: ok

tooomm:
What about macOS btw?

ebbit1q:
didn't check at all

tooomm:
@ZeldaZach can you please confirm this change and the added webp capability on macOS before we merge this PR?

ebbit1q:
nothing changed for mac, if this lib is there it probably means this never was a problem for macos like for the qt5 builds

tooomm:
We do not know for sure if it works as is or an additional change like for rpm is needed.

Better confirm than assume 👍

PR #5094, opened 2024-08-06
replace cipt check with regex

### LINKED ISSUE #5093: oracle text errata breaks oracle cipt
the august 2nd bloomburrow release came with an expansive oracle text errata, all instances of "enters the battlefield" have been changed to just "enters".
this impacts oracle which check the oracle text for instances of the text "cardname enters the battlefield tapped" which now is shortened to "cardname enters tapped".
oracle now no longer marks any cards with the `<cipt>` tag in the xml because of this.
this is an excellent example of how https://github.com/Cockatrice/Cockatrice/issues/4109 would solve such issues in the future without needing users to update their client.

ebbit1q:
## Related Ticket(s)
- Fixes #5093

the august 2nd bloomburrow release came with an expansive oracle text errata, all instances of "enters the battlefield" have been changed to just "enters".
this impacts oracle which check the oracle text for instances of the text "cardname enters the battlefield tapped" which now is shortened to "cardname enters tapped".

- the cipt check in oracle is replaced with a regex that accounts for this

because the mtgjson upstream has been changed it's difficult to compare if this change works as intended, we would need an old version of the mtgjson file from before august 2nd to check if the cards that this code marks cipt are now the same and how it affects performance.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -141,9 +141,9 @@ CardInfoPtr OracleImporter::addCard(QString name,
     // DETECT CARD POSITIONING INFO
 
     // cards that enter the field tapped
-    bool cipt = text.contains(" it enters the battlefield tapped") ||
-                (text.contains(name + " enters the battlefield tapped") &&
-                 !text.contains(name + " enters the battlefield tapped unless"));
+    QRegularExpression ciptRegex("( it|" + QRegularExpression::escape(name) +
+                                 ") enters( the battlefield)? tapped(?! unless)");
+    bool cipt = ciptRegex.match(text).hasMatch();
 
     // table row
     int tableRow = 1;


tooomm:
Maybe @ZeldaZach can help with the old file.

Why keeping the old "battlefield" variant, ebbit?

ZeldaZach:
@ebbit1q Here's a MTGJSON build from 7/30 -- https://www.dropbox.com/s/ebfgbka7ffio9ig/AllPrintings_2024_07_30.json.xz?dl=1

ebbit1q:
this mtgjson build already uses the new phrasing, I thought this errata would have been effective only when bloomburrow released but I guess I was wrong.

ebbit1q:
even though I was unable to compare in detail with an old mtgjson build this is still ready for merge, checking with random samples seems to confirm that there are no regressions which is good enough.

PR #5096, opened 2024-08-11
CI: Update install-qt-action

tooomm:
Update to v4

Dependabot is not picking this one up, because a beta branch in the source repo did trigger v4 too early and we closed it.
https://github.com/Cockatrice/Cockatrice/pull/4922


ebbit1q (approved):
thanks for noticing

PR #5097, opened 2024-08-12
feat: set prioritization by set type

cameronhimself:
## Related Ticket(s)
- N/A

I was going to just open an issue for this, but it seemed pretty simple so I decided to go ahead and propose the change for discussion instead.
### The issue

As a side effect of prioritizing sets purely by release date, Cockatrice ends up displaying oddball promo art for most popular cards out of the box. In my opinion, this really hurts the new user experience.

For this reason, instead of prioritizing sets purely by their release date, "Core" and "Expansion" sets should be prioritized first, then all other sets after.

- Initial users will have Core and Expansion sets prioritized over others

## Screenshots

2024-08-12 13_33_01-Manage sets (Ubuntu)


ebbit1q:
this is a great idea and something I've been thinking about before, I always imagined it being something to add to oracle/the xml format instead of in the client itself though, we'd like to not add more things into the base client that are specific to mtg

ebbit1q:
it's kind of silly that this code will break in 2050 though, it probably by now has used 30% of the total expected life since it was written in 2011 or so

cameronhimself:
@ebbit1q I'll take a look at moving this fix elsewhere. Thanks for the feedback.

Funnily enough, I actually have a local commit for this that gets rid of the 2050 bug. I'll commit that as well when I have an update.

ebbit1q:
a fix for that bug would be highly appreciated! it's not 2050 soon but I think it'd be nice to be early on this one :)

cameronhimself:
Actually, having thought about it a bit, it will be a confusing user experience to have a "Priority" column that doesn't match whatever priorities they've manually assigned. Maybe the column could be called "Default Priority", or something. Or maybe it could be hidden from the user altogether.

ebbit1q:
excellent plan, this would be a great solution and it would work with other card sources than oracle as well.

the priority levels don't have to be visible in my opinion, they should be documented on the wiki and mentioned in a tooltip or something.

the levels are arbitrary of course, maybe we could have some level that is disabled by default, and some extra high levels for custom card makers.

there are some other considerations as well, what behavior do we want for card updates? should we just add them to the top? what priority do we give to sets that don't have this option set in the xml?

this is just an expansion of the current format so a new version wouldn't be necessary I think. remember to update the schema as well when updating the parser.

note that some users might still want the old behavior for whatever reason, we should document that it is still available through the use of the header in the sets manager and then clicking the button to use the sorting as set order.

cameronhimself:
I went ahead and pushed my changes. I am definitely not a C/C++ programmer, so there's some cargo culting. Criticism is welcome.

> what behavior do we want for card updates? 

Having looked a bit at the code it seems like the current behavior is to not set the sort key, which would mean putting new sets at the top? Correct me if I'm wrong. In any case, I think that's still the only viable approach. We could calculate the sort key based on where it would be in the default prioritized sorting, but that won't be relevant if the user has set their own priority preferences. It might get pushed down into a seemingly arbitrary spot deep in their set list which they're unlikely to notice, and might lead them to believe the set didn't get added successfully.

> what priority do we give to sets that don't have this option set in the xml?

Sets without a priority get assigned the fallback priority--priority 4 in my above comment, although the numbers used in my implementation have diverged from that.

cockatrice/src/carddatabase.cpp
```diff
@@ -194,23 +196,29 @@ void SetList::markAllAsKnown()
 
 void SetList::guessSortKeys()
 {
-    // sort by release date DESC; invalid dates to the bottom.
-    QDate distantFuture(2050, 1, 1);
-    int aHundredYears = 36500;
+    defaultSort();
     for (int i = 0; i < size(); ++i) {
         CardSetPtr set = at(i);
-        if (set.isNull()) {
```
ebbit1q, line 202: this null check might still be necessary

cockatrice/src/carddbparser/carddatabaseparser.cpp
```diff
@@ -1,5 +1,7 @@
 #include "carddatabaseparser.h"
 
+#include <QDebug>
```
ebbit1q, line 3: this header seems unused here

cockatrice/src/carddbparser/cockatricexml3.cpp
```diff
@@ -77,6 +77,7 @@ void CockatriceXml3Parser::loadSetsFromXml(QXmlStreamReader &xml)
         if (name == "set") {
             QString shortName, longName, setType;
             QDate releaseDate;
+            short priority;
```
ebbit1q, line 80: I think we can leave the old v3 parser as is

ebbit1q (commented):
looks good, will be able to test only later.

good idea to have some extra space in between the levels for flexibility. might want to simply have the highest level have the highest priority, either works, it just looks a bit strange to have the ternary setting zero to seventy.

thanks helping out, we appreciate it!

cockatrice/src/carddatabase.cpp
cameronhimself: Good catch, got lost in the shuffle. Fixed.

cockatrice/src/carddbparser/cockatricexml3.cpp
cameronhimself: Reverted.

cameronhimself:
>might want to simply have the highest level have the highest priority, either works, it just looks a bit strange to have the ternary setting zero to fifty.

I thought about that, but we'd still have the bit of strangeness either way since the fallback prio is going to end up somewhere in the middle regardless. That said, we could maybe push that bit of logic out of the `CardSet` model and leave it to downstream consumers to handle undefined priorities. Thinking about it now, that's probably the best approach since there's otherwise no way outside the class to distinguish between a set that has a priority 0 and a set that actually has an explicit priority of 40. I'll refactor.

cameronhimself:
> what behavior do we want for card updates?

Having thought about this a bit more, I think it will be worthwhile to have a special check to see if the user's sets are still in the default priority order, and add the new set to the correct prioritized place if so (otherwise leave it at the top). I suspect that if this set prioritization feature gets merged, there will be much less of a reason for users to customize their set priorities, so having that default use-case covered might end up benefiting a lot of users.

I'll work on that as a separate PR. Is there a good way to test a new set being added? Interacting with the cards JSON file to make a mock is a bit tricky since it's so large and unformatted.

ebbit1q:
with to test a new set being added you mean locally? you could simply remove a set from the relevant settings ini file. or you could make a simple custom set xml with just a sets entry. don't nock the source json, we have a few tests on simple xmls already.

cameronhimself:
@ebbit1q Thanks, that makes sense. I wasn't thinking about it correctly.

I think I'm going to put that feature in a separate PR, since I feel this feature is complete enough without it. Is there anything else needed for this PR to be approved?

ebbit1q:
hey, thanks for reminding me, I still need to get around to testing your changes, I'll make time for it as soon as possible

oracle/src/oracleimporter.h
```diff
@@ -7,11 +7,35 @@
...
+    {"core", CardSet::PriorityPrimary},
+    {"expansion", CardSet::PriorityPrimary},
+
+    {"commander", CardSet::PrioritySecondary},
+    {"starter", CardSet::PrioritySecondary},
+
+    {"archenemy", CardSet::PriorityReprint},
+    {"arsenal", CardSet::PriorityReprint},
+    {"box", CardSet::PriorityReprint},
+    {"draft_innovation", CardSet::PriorityReprint},
```
ebbit1q, line 21: modern horizons and jumpstart are counted as draft innovations, they contain new cards that preferably get a higher priority than promo sets

this also applies but less so to duel decks, which only rarely include new cards.

doc/carddatabase_v3/cards.xsd
```diff
@@ -31,6 +31,7 @@
                                 <xs:element type="xs:string" name="longname" minOccurs="0" maxOccurs="1" />
                                 <xs:element type="xs:string" name="settype" minOccurs="0" maxOccurs="1"/>
                                 <xs:element type="xs:string" name="releasedate" minOccurs="0" maxOccurs="1"/>
+                                <xs:element type="xs:integer" name="priority" minOccurs="1" maxOccurs="1"/>
```
ebbit1q, line 34: this can be removed now

cockatrice/src/setsmodel.cpp
```diff
@@ -80,6 +82,9 @@ QVariant SetsModel::headerData(int section, Qt::Orientation orientation, int rol
         case IsKnownCol:
             return QString(
                 "Is known"); /* no tr() for translations needed, column is just used for sorting --> hidden */
+        case PriorityCol:
```
ebbit1q, line 85: if the priority is never shown to users, and the sorting is moved to the setslist, what benefit does adding the priority column to the model have?

cockatrice/src/carddatabase.cpp
```diff
@@ -194,23 +196,35 @@ void SetList::markAllAsKnown()
...
-        QDate date = set->getReleaseDate();
-        if (date.isNull()) {
-            set->setSortKey(static_cast<unsigned int>(aHundredYears));
-        } else {
-            set->setSortKey(static_cast<unsigned int>(date.daysTo(distantFuture)));
+void SetList::defaultSort()
+{
+    std::sort(begin(), end(), [](const CardSetPtr &a, const CardSetPtr &b) {
+        QDate distantPast = QDate(1970, 1, 1);
+        QDate aDate = a->getReleaseDate().isValid() ? a->getReleaseDate() : distantPast;
```
ebbit1q, line 214: this function will get called a bunch for sorting, it might be worth optimising it just a bit by trying to return early and trying to save and reuse results of functions.

cockatrice/src/carddatabase.cpp
```diff
@@ -194,23 +196,35 @@ void SetList::markAllAsKnown()
...
 
-        QDate date = set->getReleaseDate();
-        if (date.isNull()) {
-            set->setSortKey(static_cast<unsigned int>(aHundredYears));
-        } else {
-            set->setSortKey(static_cast<unsigned int>(date.daysTo(distantFuture)));
+void SetList::defaultSort()
+{
+    std::sort(begin(), end(), [](const CardSetPtr &a, const CardSetPtr &b) {
+        QDate distantPast = QDate(1970, 1, 1);
```
ebbit1q, line 213: this could be static and const right?

cockatrice/src/carddatabase.cpp
```diff
@@ -194,23 +196,35 @@ void SetList::markAllAsKnown()
...
+        QDate aDate = a->getReleaseDate().isValid() ? a->getReleaseDate() : distantPast;
+        QDate bDate = b->getReleaseDate().isValid() ? b->getReleaseDate() : distantPast;
+        CardSet::Priority aPriority = a->getPriority() ? a->getPriority() : CardSet::PriorityFallback;
+        CardSet::Priority bPriority = b->getPriority() ? b->getPriority() : CardSet::PriorityFallback;
```
ebbit1q, line 217: setting the fallback to be pretty low causes most custom sets and spoilers to be sorted pretty late, it might be nice to have custom sets be sorted early, it's a more advanced feature though but most users of custom sets would like them prioritised. the spoilers being sorted can be added to https://github.com/Cockatrice/Magic-Spoiler

ebbit1q:
I've tested your changes and everything looks good, I have a few minor nitpicks in the comments and then this would be ready to merge.

@ZeldaZach do you have any comments on adding a priority field to the xml?

ZeldaZach:
No comments from my end, seems like a reasonable and welcomed change

oracle/src/oracleimporter.h
cameronhimself: Good catch. I moved those sets to the `PrioritySecondary` category. I thought about making a `PriorityTertiary` classification for them, but it didn't seem necessary.

cockatrice/src/carddatabase.cpp
cameronhimself: Hmm, could you show me how best to go about this? I'm sure it's just due to my unfamiliarity with the language, but I don't see any obvious ways to optimize. As a sort function, it has to return either `true` or `false`, so I don't see any opportunity to return early, and the only functions/methods that it calls are only called once, so storing their return values doesn't look like it would help to me.

cockatrice/src/carddatabase.cpp
cameronhimself: Thanks for the repo link for spoilers; I'll look into making a PR.

Regarding custom sets: having experimented with them a bit now, it looks like Cockatrice doesn't persist whether a set is custom or not, making it impossible to factor into sorting. That said, the default sort function _does_ respect the priority defined in the custom set's XML, including priorities with greater precedence than Core/Expansion. So, my proposal is to simply not distinguish custom sets, and document the priority field for custom set creators to incorporate it into their XML files. As you say, it's an advanced feature, so it doesn't seem like an unreasonable ask.

The other option I see would be to persist a flag that distinguishes custom sets from official sets, and then use that flag to push custom sets to the top, but I didn't want to go down that road before discussing it. Let me know what you think.

ebbit1q: with this pr all sets that don't have a priority set will be either custom sets or the spoiler xml, we can change the spoiler xml generation to fix the last one, so we do have a way to tell if a set is custom, it'd simply be all sets without priority set

ebbit1q: in case new set types get added, I propose we do change the oracle code for those to simply set them to a low type

ZeldaZach:
Hey hey, if you're able to rebase this @cameronhimself I'd be happy to merge it in :)

doc/carddatabase_v4/cards.xsd
```diff
@@ -15,6 +15,7 @@
             <xs:element type="xs:string" name="longname" minOccurs="0" maxOccurs="1" />
             <xs:element type="xs:string" name="settype" minOccurs="0" maxOccurs="1" />
             <xs:element type="xs:string" name="releasedate" minOccurs="0" maxOccurs="1" />
+            <xs:element type="xs:integer" name="priority" minOccurs="1" maxOccurs="1"/>
```
ZeldaZach, line 18: If this becomes a required field, we might need to bump the version. I'd rather not do that.

oracle/src/oracleimporter.h
```diff
@@ -7,11 +7,35 @@
 #include <utility>
 
 // many users prefer not to see these sets with non english arts
-// as a solution we remove the date property on these sets
-// that way they will be sorted last by default
-// this will cause their art to not get priority over english cards
-// users will still be able to find these sets and prioritize them manually
 const QStringList nonEnglishSets = {"4BB", "FBB", "PS11", "PSAL", "REN", "RIN"};
+const QMap<QString, CardSet::Priority> setTypePriorities{
```
ZeldaZach, line 11: Does this account for all types at this point?

cockatrice/src/carddatabase.cpp
ZeldaZach: I'm also not sure where you're trying to go with this @ebbit1q can you clarify?

ZeldaZach (changes requested):
A few small tweaks, but overall this looks pretty solid!

cameronhimself:
Apologies for my absence, I should be able to address comments and rebase this week.

oracle/src/oracleimporter.h
cameronhimself: To my knowledge, yes, unless additional types have been added since I wrote this.

cockatrice/src/carddatabase.cpp
cameronhimself: I made `PriorityFallback` 1, above the others.

cameronhimself:
@ebbit1q @ZeldaZach I've rebased and addressed all the comments I'm able to. Sorry for the delay.

ebbit1q:
don't worry about taking your time! thank you so much for helping make cockatrice!

cameronhimself:
Here's the companion PR for Magic-Spoiler: https://github.com/Cockatrice/Magic-Spoiler/pull/308

cockatrice/src/carddatabase.cpp
cameronhimself: To me it seems worthwhile to distinguish between a set with a priority that hasn't been evaluated yet (`PriorityUndefined`; zero; falsy), and a set that has been evaluated and found no matching priority (`PriorityFallback`; 1; truthy, like other prioritized sets).

cameronhimself: That said, it's not something I feel strongly about and am happy to change it to zero. It just made the most sense in the moment to preserve the distinction, since the comments were only regarding where `PriorityFallback` falls, not the possibility of collapsing `PriorityFallback` and `PriorityUndefined` together.

ebbit1q: what benefits would having fallback over undefined have? it always gets changed from one into the other and there don't seem to be any other checks for it.

cameronhimself: Removed `PriorityUndefined` and set `PriorityFallback` to 0.

cockatrice/src/carddatabase.cpp
ebbit1q: I've changed it a bit, it's much simpler now and does the same thing.

ebbit1q (approved):

ZeldaZach (approved):

PR #5103, opened 2024-08-30
#3945 deck list: Navigation keys (PageUp/Down, Home/End) 

### LINKED ISSUE #3945: page up/page down in "Deck" tab/window / search results for faster browsing
**Similar Requests**

**Additional Context**

PoltyPoltou:
## Related Ticket(s)
- Fixes #3945

No navigation possible when writing in the card search text box apart from going on card up/down.

When focusing the card list or the search text box in the deck editor:
- `PageDown`, `PageUp` scrolls one page down or up the card list respectively.
- `Home` scrolls to the top the complete list, `End` to the bottom

### DIFF
--- a/cockatrice/src/tab_deck_editor.cpp
+++ b/cockatrice/src/tab_deck_editor.cpp
@@ -54,11 +54,11 @@ void SearchLineEdit::keyPressEvent(QKeyEvent *event)
     static const QVector<Qt::Key> forwardToTreeView = {Qt::Key_Up, Qt::Key_Down, Qt::Key_PageDown, Qt::Key_PageUp};
     // forward only if the search text is empty
     static const QVector<Qt::Key> forwardWhenEmpty = {Qt::Key_Home, Qt::Key_End};
-
+    Qt::Key key = static_cast<Qt::Key>(event->key());
     if (treeView) {
-        if (forwardToTreeView.contains(event->key()))
+        if (forwardToTreeView.contains(key))
             QCoreApplication::sendEvent(treeView, event);
-        if (text().isEmpty() && forwardWhenEmpty.contains(event->key()))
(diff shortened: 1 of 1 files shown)


PoltyPoltou:
One thing to note, is that the Home/End lose their default behavior in the text widget. I see three solutions:
- It is rare that people will use Home/End on the search row, so Home/End will always go the card list.
- Home/End will work as intended if the search is not empty, but if empty it will trigger the Go to top/bottom of the card list.
- Use other binding for Go to top/bottom (Shift+PageUp/PageDown)

PoltyPoltou:
I put in bold the actual behavior of my fork

ebbit1q:
This pull request is excellent and does what the ticket says, I have one question though, what is the use of having end an home only work when there is no text? going to the end of the entire list unfiltered is not very useful, but might be useful when there is text, annoyingly this does interact with the way text is put into the search bar, @tooomm what was your intent when you made this suggestion?

one note, some debug builds are failing because of the implicit conversion here https://github.com/Cockatrice/Cockatrice/actions/runs/10633383622/job/29493815724?pr=5103#step:6:998 , to fix this make the conversion explicit.

PoltyPoltou:
I put it this way because of the issue infos and another reasoning is: it would keep default controls of the text input bar, ie. in a text input pressing Home makes your cursor go to the beginning.

I agree with the need of Home/End when the list is filtered, I guess very few people are using Home/End in text editing, so I feel like overriding is good enough (compared to rebinding).

PoltyPoltou:
Not sure if the static_cast is needed for the compilation fix, I guess `Qt::Key key = event->key()` works too. But I like the explicit cast int to enum.

ebbit1q (approved):
does what the original ticket asked, we can consider changing the behavior of the home and end keys in a later ticket.

PR #5107, opened 2024-09-02
Webclient: Pretty print translation source

tooomm:
In https://github.com/Cockatrice/Cockatrice/pull/5021 all translation files did changed their formatting.

The problem is, that for any updates after that, it's impossible to review or even spot changes in the diff - no hint what or how much changed, see https://github.com/Cockatrice/Cockatrice/pull/5045.

- Pretty print the source json


tooomm:
This got actually already fixed in #4806 once.

But re-introduced in #5014
@seavor do you have a local formatting setting not set maybe?

tooomm:
As seen in the recent PR updating the translations from transifex, there is no way to tell what actually changed since it's a single line change - no matter if it's 1 string or 20.

https://github.com/Cockatrice/Cockatrice/pull/5129/files#diff-7b9d421190bb25481a8fc8a9d23e0f6a970896990ed039e9d012abbcec2fd4f3

ZeldaZach (approved):

PR #5108, opened 2024-09-03
Update Dockerfile Ubuntu version to newest LTS

moryall:
## Issue:
Dockerfile uses a version of Ubuntu that has reached end-of-life for primary LTS support. 

## Changes:
1. Updated image to Jammy (22.04) as it is still in it's LTS window. Didn't go with Noble (24.04) as it released after latest version of cockatrice released and I wanted to change as little as possible.
2. Removed maintainer as the new Dockerfile format is to use labels instead.
3. Had to add new ARG for it to pass the prompts in Jammy.
4. No qt5-default library, so replaced with qt5-qmake.

### Note:
I deeply apologize if I have done something wrong in this process, just figured I'd update it since I try to keep my systems current and noticed Bionic was outdated.


ebbit1q:
this docker file is simply unmaintained as none of us use it, it doesn't even seem to use the new 4748 port and is still using qt5 instead of 6, I'd rather just delete it and if anyone was using it they can make a better one using our compilation instructions.

ZeldaZach:
If you're able to update the instructions to Qt6 and use web port 4748 @moryall i'd be happy to consider a merge on this!

ZeldaZach (changes requested):
Please update to Qt6 instructions

moryall (commented):
Sorry for the delay's and also if this pull was not modified in the right place. Requests completed.

Dockerfile
```diff
@@ -1,19 +1,28 @@
-FROM ubuntu:bionic
-MAINTAINER Zach Halpern <zahalpern+github@gmail.com>
+FROM ubuntu:noble
```
tooomm, line 1: Can you use the numbered version over the name, please?

moryall: Makes sense & done

tooomm:
I did not test to build and run an image from the file, but all requested changes are included.
Maybe update the PR description as well for better future reference.

As this image builds Servatrice, not Cockatrice Client, does it make sense to clarify this better?
I saw other projects using a naming scheme like `Dockerfile.Servatrice`.

Dockerfile
```diff
@@ -1,19 +1,28 @@
...
-  libprotobuf-dev\
-  libqt5sql5-mysql\
-  libmysqlclient-dev\
-  libqt5websockets5-dev\
-  protobuf-compiler\
-  qt5-default\
-  qtbase5-dev\
-  qttools5-dev-tools\
-  qttools5-dev
+  build-essential \
```
ebbit1q, line 6: the indentation on this is a bit weird

Dockerfile
```diff
@@ -1,19 +1,28 @@
...
+        ccache \
+        clang-format \
+        cmake \
+        file \
+        g++ \
+        git \
+        libgl-dev \
+        liblzma-dev \
```
ebbit1q, line 14: many of these dependencies are only relevant to our ci (I guess you copied them from there)
for compiling the server none of the gui things are needed either.
remove: ccache clang-format liblzma-dev libgl-dev libqt6multimedia6 qt6-svg-dev qt6-l10n-tools qt6-multimedia-dev

ZeldaZach (approved):

PR #5112, opened 2024-09-15
fix macro shortcuts so that Ctrl+1 is not double-assigned

AlexanderJChoi:
## Related Ticket(s)
- No related tickets

In the menu Cockatrice > Settings > Chat, you can add In-game message macros. The first several macros are automatically assigned to the shortcuts Ctrl+1 to Ctrl+0. When more than 10 macros are added, Ctrl+1 does not send a message associated to any macro. 

- When more than 10 macros are added, Ctrl+1 will now send the message associated with the first macro in the list.


ZeldaZach:
Thanks for the fix!

PR #5113, opened 2024-09-17
In-game message macros available immediately in active games

### LINKED ISSUE #4166: Edit in-game message macros during a game and make them available immediately
In the settings, under the chat tab, you can edit in-game message macros (thank you for implementing this feature!). Currently, these changes are not available immediately if you do this during a game. I believe this would be a useful feature to have in multiplayer games to facilitate the communication of repeating triggered abilities.

Rhystic Study is probably the most prominent example where this feature would be useful, but there are certainly others. If an opponent controls one, you can add or edit two macros, one with "I will not pay {1} for Rhystic Study" and another with "I pay {1} for Rhystic Study", and use them whenever you play a spell without having to constantly type the same message.

Lastly, thank you so much for all your contributions to the development of the Cockatrice project 😊

AlexanderJChoi:
## Related Ticket(s)
- Fixes #4166 

- Previously, changes made to in-game message macros were not made available in any in-progress games.

- When users modify the in-game message macros, these changes will now apply immediately to any in-progress games games.

### DIFF
--- a/cockatrice/src/dlg_settings.cpp
+++ b/cockatrice/src/dlg_settings.cpp
@@ -1001,6 +1001,7 @@ void MessagesSettingsPage::storeSettings()
     SettingsCache::instance().messages().setCount(messageList->count());
     for (int i = 0; i < messageList->count(); i++)
         SettingsCache::instance().messages().setMessageAt(i, messageList->item(i)->text());
+    emit SettingsCache::instance().messages().messageMacrosChanged();
 }
 
 void MessagesSettingsPage::actAdd()
--- a/cockatrice/src/player.cpp
+++ b/cockatrice/src/player.cpp
@@ -419,6 +419,7 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, bool _judge, T
 
     if (local) {
         sayMenu = playerMenu->addMenu(QString());
+        connect(&SettingsCache::instance().messages(), SIGNAL(messageMacrosChanged()), this, SLOT(initSayMenu()));
         initSayMenu();
     }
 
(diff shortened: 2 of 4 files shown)


ebbit1q:
the initsaymenu looks like it leaks memory, does clearing the menu deallocate the previous actions? it might reparent them when they are added to the menu maybe? could you check if it does this in the docs?

AlexanderJChoi:
It looks like QMenu::clear() will delete the QActions, but ONLY if it is the parent for those actions (and the actions not being used elsewhere). The sayMenu actions are currently being initialized with Player as the parent. They are NOT re-parented when added to the menu. 

We could initialize sayMenu actions with sayMenu as parent? And that should fix the leak. I'll try this tomorrow.

AlexanderJChoi:
Hmm I'm having trouble seeing the memory leak from initSayMenu (in commit 9200dd9) in valgrind, so maybe someone can help point me toward how to check it?

I compiled a debug version of cockatrice, following the wiki.

and I'm running it in valgrind with the following flags `valgrind --leak-check=full --show-leak-kinds=all --log-file=valgrind-out.txt cockatrice`

but I'm not seeing anything in the log file that has initSayMenu in the stack trace. Is there something I'm missing here?

ebbit1q:
it's unlikely to show in valgrind, if the parent is the player then it will be in the list of children on it, with pointer, so it will not be considered "lost". yet it will never be used by that player, it will be cleaned up when the player is destroyed though, this happens when the game is finished.

ebbit1q (approved):
thank you for this change!

PR #5137, opened 2024-10-20
Add backwards support Qt6.7's checkStateChanged on QCheckBoxes

ZeldaZach:


ebbit1q (approved):
I kind of like how the macros make this a bit simpler, also to read, the downside is that in the future when we somehow get to the point that the lowest version we support is 6.7 we'll have to change a few lines instead of delete version guards. I think that's entirely fine, thanks for fixing this!
the format for the declarations are whatever, if you prefer them to have a space before the {} we should change that in the formatting options for clang-format.

PR #5140, opened 2024-10-21
Add keyboard shortcuts for skipping forward/backward in replays

### LINKED ISSUE #5139: Add keyboard shortcuts for skipping slightly forward/backward in replays
**Similar Requests**
#1536

Now that we have keyboard shortcuts in replays, we should add a keyboard shortcut for skipping slightly forward/backward in the timeline. 

A good default for the keys would be `<--` and `-->`.

As for the amount of time to skip, either 5 seconds or 10 seconds seems like a good value to try. I'm not sure if we need to make that value configurable. Perhaps we could have two keys: one for small skip and one for larger skip.

RickyRister:
## Related Ticket(s)
- Fixes #5139

https://github.com/user-attachments/assets/a564cb59-a2ac-496d-bbf6-8aaf1935236d
- Pressing left/right arrow keys will now skip forwards/backwards in replays.
- Holding `Ctrl` while pressing the arrow keys will skip by a larger amount.
  - Changed default fast-forward shortcut to `.`, since it was conflicting with `Ctrl+Right` 
- All added shortcuts are rebindable.
  Screenshot 2024-10-20 at 10 10 15 PM
- Also fixed bug with cockatrice complaining on startup that `space` is an invalid shortcut for pause/unpause

**Known Issues:** If you go to the end of a long replay and spam `Left`, Cockatrice will eventually lag out because it has to do a rewind for every single backwards skip. This is already accomplishable by clicking on the timeline fast enough, but adding the shortcut makes doing it a lot easier.


cockatrice/src/client/tabs/tab_game.cpp
```diff
@@ -1699,12 +1711,33 @@ void TabGame::createPlayAreaWidget(bool bReplay)
 
 void TabGame::createReplayDock()
 {
+    // timeline widget
```
ZeldaZach, line 1714: Is this comment needed?

cockatrice/src/client/tabs/tab_game.h
```diff
@@ -142,11 +142,14 @@ class TabGame : public Tab
     QStackedWidget *mainWidget;
 
     // Replay related members
+    const int SKIP_AMOUNT_SMALL = 1000;
```
ZeldaZach, line 145: Can this magic number get a unit attached? _SECONDS? _MILLISECONDS? etc

cockatrice/src/settings/shortcuts_settings.h
```diff
@@ -590,11 +590,23 @@ class ShortcutsSettings : public QObject
...
+                                                ShortcutGroup::Replays)},
+        {"Replays/aSkipBackwardBig", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Skip Backward by a lot"),
+                                                 parseSequenceString("Ctrl+Left"),
+                                                 ShortcutGroup::Replays)},
         {"Replays/playButton", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Play/Pause"),
                                            parseSequenceString("Space"),
                                            ShortcutGroup::Replays)},
         {"Replays/fastForwardButton", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Toggle Fast Forward"),
-                                                  parseSequenceString("Ctrl+Right"),
+                                                  parseSequenceString("."),
```
ZeldaZach, line 609: I'm not sure I agree with bounding something to a non-control key like this. @ebbit1q what do you think?

ZeldaZach:
Thanks for your contribution! Left a few comments :)

cockatrice/src/settings/shortcuts_settings.h
ebbit1q: I think . is a bit of a strange choice, maybe just set it to ctrl f, for fast forward

cockatrice/src/client/tabs/tab_game.cpp
RickyRister: Probably not. But since I decided to put comments at the top of every grouping, I felt like it looks nicer if there's something on this line.

cockatrice/src/settings/shortcuts_settings.h
RickyRister: gah, just realized that `Ctrl+F` is already used for fullscreen

ebbit1q: yeah, there is no check for duplicates in the default shortcuts

ZeldaZach (approved):

ebbit1q (approved):
thank you

PR #5141, opened 2024-10-24
Buffer rewinds from backward skips in replays

RickyRister:
## Related Ticket(s)
- Follow up to #5140

While viewing a replay, you may want to hold down `Left` in order to continually back up in the timeline. However, this results in a lot of small backwards skips in succession. Every single one of those backward skips will force cockatrice to do a rewind, wasting a lot of processing. 

@ebbit1q suggested making cockatrice wait a bit after each backwards skip to make sure there isn't any more backwards skips coming in quick succession, before processing the rewind.

https://github.com/user-attachments/assets/aeb4d89f-f837-46a2-8a29-531f5887d3d4

When a backwards skip is caused by a keyboard shortcut, we don't process the new board state immediately. Instead, we start a one-shot timer. The rewind and reprocessing will be triggered by the timer finishing. If a timer was already going, then we just restart the timer instead of doing anything else. 

This makes it so the new board state isn't processed until the flurry of backward skips concludes, saving on the number of rewind and reprocess operations.

Backward skips caused by clicks aren't buffered and will still always cause a rewind. Clicks usually don't happen fast enough for buffering to be necessary. I'd rather that the responsiveness of clicks don't change.

`REWIND_BUFFERING_TIMEOUT_MS` wants to be set to a value that's low enough to where it still feels responsive if you're just doing a one-off key press, but high enough for the buffering to take effect in most scenarios where the user is holding or spamming the left key. Currently, I've just set it to a value that felt good enough to me, but it's entirely debatable as to what the final value should be.

Also cleaned up the `ReplayTimelineWidget` class as part of implementing this.

### DIFF
--- a/cockatrice/src/client/network/replay_timeline_widget.cpp
+++ b/cockatrice/src/client/network/replay_timeline_widget.cpp
@@ -118,8 +118,8 @@ void ReplayTimelineWidget::handleBackwardsSkip(bool doRewindBuffering)
 /// The timeout scales based on the current event number, up to a limit
 int ReplayTimelineWidget::calcRewindBufferingTimeout() const
 {
-    int extraTime = std::min(currentEvent / 100, 100);
-    return BASE_REWIND_BUFFERING_TIMEOUT_MS + extraTime;
+    int extraTime = currentEvent / 100;
+    return std::min(BASE_REWIND_BUFFERING_TIMEOUT_MS + extraTime, MAX_REWIND_BUFFERING_TIMEOUT_MS);
 }
 
 void ReplayTimelineWidget::processRewind()
--- a/cockatrice/src/client/network/replay_timeline_widget.cpp
+++ b/cockatrice/src/client/network/replay_timeline_widget.cpp
@@ -82,12 +82,11 @@ void ReplayTimelineWidget::skipToTime(int newTime)
 
     newTime -= newTime % 200; // Time should always be a multiple of 200
     if (newTime < currentTime) {
-        currentTime = 0;
         currentEvent = 0;
         emit rewound();
     }
-    currentTime = newTime - 200; // 200 is added back in replayTimerTimeout
-    replayTimerTimeout();
+    currentTime = newTime;
+    processNewEvents();
     update();
 }
 
@@ -104,6 +103,16 @@ QSize ReplayTimelineWidget::minimumSizeHint() const
 void ReplayTimelineWidget::replayTimerTimeout()
 {
     currentTime += 200;
+
+    processNewEvents();
+
+    if (!(currentTime % 1000))
+        update();
+}
+
+/// Processes all unprocessed events up to the current time.
+void ReplayTimelineWidget::processNewEvents()
+{
     while ((currentEvent < replayTimeline.size()) && (replayTimeline[currentEvent] < currentTime)) {
         emit processNextEvent();
         ++currentEvent;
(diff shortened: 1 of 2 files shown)


tooomm:
For forward skipping, this is not an issue?

RickyRister:
> For forward skipping, this is not an issue?

For forward skips, cockatrice just naturally processes through the events until it reaches the new current time. Only backward skips require cockatrice to rewind and reprocess all events from the beginning. So no, forward skips are not an issue and can just be handled naturally.

cockatrice/src/client/network/replay_timeline_widget.h
```diff
@@ -18,6 +18,8 @@ class ReplayTimelineWidget : public QWidget
 
 private:
     QTimer *replayTimer;
+    static const int REWIND_BUFFERING_TIMEOUT_MS = 180;
```
ebbit1q, line 21: in my testing with a long replay this seems a bit short still, cockatrice seems to take 50% single core if I tap back slower than 5 times per second

RickyRister: should we just make this value configurable so we don't have to settle on a hardcoded value?

ebbit1q: we could make it proportional to the total length of the replay, or how far into the replay it is, or how estimate based on the total length of the replay (note that not all commands take the same amount of cycles to process)

cockatrice/src/client/network/replay_timeline_widget.cpp
```diff
@@ -81,16 +86,53 @@ void ReplayTimelineWidget::skipToTime(int newTime)
...
+        // otherwise, process the rewind immediately
+        processRewind();
+    }
+}
+
+/// The timeout scales based on the current event number, up to a limit
+int ReplayTimelineWidget::calcRewindBufferingTimeout() const
+{
+    int extraTime = std::min(currentEvent / 100, 100);
+    return BASE_REWIND_BUFFERING_TIMEOUT_MS + extraTime;
```
ebbit1q, line 122: why have both a base and a minimum added to each other? why is the minimum not a constant?

RickyRister: My intention is to have the total rewind timeout start at the base timeout and then scale up based on the number of events, up to a maximum of +100. Is there a better way to write that? 

I could make 100 into a `MAX_EXTRA_TIMEOUT` constant i guess.

cockatrice/src/client/network/replay_timeline_widget.h
```diff
@@ -18,6 +18,9 @@ class ReplayTimelineWidget : public QWidget
 
 private:
     QTimer *replayTimer;
+    static constexpr int BASE_REWIND_BUFFERING_TIMEOUT_MS = 180;
+    static constexpr int MAX_REWIND_BUFFERING_TIMEOUT_MS = 280;
```
ZeldaZach, line 22: Any context on where these magic numbers came from?

ZeldaZach (approved):
LGTM, should address other reviewer concerns

cockatrice/src/client/network/replay_timeline_widget.h
RickyRister: I just fiddled with the numbers until it felt good enough

ebbit1q: they could become configurable, I'm good with merging this as is though

ebbit1q (approved):

PR #5143, opened 2024-10-25
Add option to open deck in new tab by default

### LINKED ISSUE #5142: Option to open new deck editor tab by default when loading a deck
Cockatrice currently supports having multiple deck editor tabs open. However, the only way to create a new deck editor tab is through `Actions` -> `Deck Editor`. 

There will be times when I'm working on a deck, and want to load up another deck in a new tab. What usually happens is
- click `Load Deck...`
- realize that I need to open the new tab first, so then back out of the deck selection window
- click `Actions` -> `Deck Editor`
- finally click `Load Deck...`

It would be a quality-of-life improvement if there was an option to open a new tab by default when loading a deck.

RickyRister:
## Related Ticket(s)
- Fixes #5142 

Added the new setting `Always open deck in new tab`. 
It's located in the new `Deck editor settings` section under `User Interface`.

Screenshot 2024-10-24 at 8 42 42 PM

When enabled, `New deck`, `Load deck...`, and `Load deck from clipboard` will open the deck in a new deck editor tab, instead of replacing the deck in the existing tab. Cockatrice won't display the unsaved changes dialogue window, since we're no longer overwriting anything.

https://github.com/user-attachments/assets/19de3e88-55d4-4876-a7d0-62a5ce79094c

### DIFF
--- a/cockatrice/src/settings/cache_settings.cpp
+++ b/cockatrice/src/settings/cache_settings.cpp
@@ -541,7 +541,7 @@ void SettingsCache::setTapAnimation(QT_STATE_CHANGED_T _tapAnimation)
 void SettingsCache::setOpenDeckInNewTab(QT_STATE_CHANGED_T _openDeckInNewTab)
 {
     openDeckInNewTab = static_cast<bool>(_openDeckInNewTab);
-    settings->setValue("editor/openDeckInNewTab", _openDeckInNewTab);
+    settings->setValue("editor/openDeckInNewTab", openDeckInNewTab);
 }
 
 void SettingsCache::setChatMention(QT_STATE_CHANGED_T _chatMention)
--- a/cockatrice/src/settings/cache_settings.cpp
+++ b/cockatrice/src/settings/cache_settings.cpp
@@ -243,7 +243,7 @@ SettingsCache::SettingsCache()
     invertVerticalCoordinate = settings->value("table/invert_vertical", false).toBool();
     minPlayersForMultiColumnLayout = settings->value("interface/min_players_multicolumn", 4).toInt();
     tapAnimation = settings->value("cards/tapanimation", true).toBool();
-    openDeckInNewTab = settings->value("editor/open_deck_in_new_tab", true).toBool();
+    openDeckInNewTab = settings->value("editor/openDeckInNewTab", true).toBool();
     chatMention = settings->value("chat/mention", true).toBool();
     chatMentionCompleter = settings->value("chat/mentioncompleter", true).toBool();
     chatMentionForeground = settings->value("chat/mentionforeground", true).toBool();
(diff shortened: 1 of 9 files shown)


cockatrice/src/settings/cache_settings.cpp
```diff
@@ -537,6 +538,12 @@ void SettingsCache::setTapAnimation(QT_STATE_CHANGED_T _tapAnimation)
     settings->setValue("cards/tapanimation", tapAnimation);
 }
 
+void SettingsCache::setOpenDeckInNewTab(QT_STATE_CHANGED_T _openDeckInNewTab)
+{
+    openDeckInNewTab = static_cast<bool>(_openDeckInNewTab);
+    settings->setValue("editor/open_deck_in_new_tab", _openDeckInNewTab);
```
ZeldaZach, line 544: this should be `openDeckInNewTab` (without the underscore)

ZeldaZach (approved):
LGTM, one small tweak

cockatrice/src/settings/cache_settings.cpp
tooomm: Unfortunately it's not uniform in the file.

Like
`settings->setValue("cards/tapanimation", tapAnimation);`
or
`settings->setValue("table/invert_vertical", invertVerticalCoordinate);`

cockatrice/src/settings/cache_settings.cpp
```diff
@@ -537,6 +538,12 @@ void SettingsCache::setTapAnimation(QT_STATE_CHANGED_T _tapAnimation)
     settings->setValue("cards/tapanimation", tapAnimation);
 }
 
+void SettingsCache::setOpenDeckInNewTab(QT_STATE_CHANGED_T _openDeckInNewTab)
+{
+    openDeckInNewTab = static_cast<bool>(_openDeckInNewTab);
+    settings->setValue("editor/openDeckInNewTab", _openDeckInNewTab);
```
ebbit1q, line 544:
```suggestion
    settings->setValue("editor/openDeckInNewTab", openDeckInNewTab);
```

cockatrice/src/settings/cache_settings.cpp
```diff
@@ -243,6 +243,7 @@ SettingsCache::SettingsCache()
     invertVerticalCoordinate = settings->value("table/invert_vertical", false).toBool();
     minPlayersForMultiColumnLayout = settings->value("interface/min_players_multicolumn", 4).toInt();
     tapAnimation = settings->value("cards/tapanimation", true).toBool();
+    openDeckInNewTab = settings->value("editor/openDeckInNewTab", true).toBool();
```
ebbit1q, line 246: I'm not sure if we want to default to the new behavior immediately.

ebbit1q:
I feel like when a decklist is empty there is no need to open a new tab. also, opening a new tab takes a solid couple of seconds for me, probably because of the huge card list. I'd prefer that as a default it could be a third option when it prompts you with the "the decklist has been modified" dialog. the setting could persist but not as a default.

RickyRister:
> I feel like when a decklist is empty there is no need to open a new tab. also, opening a new tab takes a solid couple of seconds for me, probably because of the huge card list. 

Good point. I'll add a check for empty decklist before opening a new tab.

> I'd prefer that as a default it could be a third option when it prompts you with the "the decklist has been modified" dialog. 

Good idea. We can add that in a follow-up PR.

> the setting could persist but not as a default.

👍

ebbit1q (approved):
this is some neat extra functionality and ready to merge rn, improvements can come in later prs

PR #5147, opened 2024-10-26
fix crashes in local games because of using uninitialised pointers

ebbit1q:
## Related Ticket(s)
- Fixes #IssueNumber

for some reason since #5140 something changed and the horrible design of multiple constructors in tab_game now causes a crash, in tab_game we have many variables only used by one of the two constructors and none of them got initialised, resulting in undefined behavior when they were used, for the longest time noone cared about this because undefined can mean zero, since this latest change it wasn't zero, at least for me.

- initialise all the pointers with keybinds so that when they get used they are zero.
- no more random crash when creating a game for me, which might not even be reproducable for others


ebbit1q:
I'd strongly prefer we'd split the multiple tab games into two different classes using inheritance for their shared functions, I have no idea why this wasn't the original design.
however, we need master to stop crashing so this is a band-aid solution.

PR #5151, opened 2024-10-31
Remove Fedora 39 Build and Add Fedora 41 Build

SlightlyCircuitous:
Fedora 39 will [reach end of life on November 12, 2024](https://fedorapeople.org/groups/schedule/f-39/f-39-key-tasks.html) so we will not need to build for it in future releases. 

Fedora 41 was released on October 29, 2024 and [will be supported until November 19, 2025](https://fedorapeople.org/groups/schedule/f-41/f-41-all-milestones-tasks.html) so it would be advantageous to our users to start building for it instead.

- Removes the dockerfile and matrix entry for Fedora 39
 - Adds a dockerfile and a matrix entry for Fedora 41
 - Revises the Release Template to reflect the above changes


SlightlyCircuitous:
Looking into this failure. Specifically, it seems to have built Cockatrice successfully and then failed to package it into an .rpm. Not sure why it succeeded in Fedora 40 and not 41, but I'm trying to look at that now.

Possibly related, possibly not, when I try to build on my Fedora 41 VM, running `make package` makes a .deb and seemingly ignores the fact that its running on Fedora.

ebbit1q:
this seems like an upstream issue in cmake's cpack

ZeldaZach:
Moved to draft b/c not passing build

ebbit1q (approved):
thanks @tooomm for updating this pr

ebbit1q:
I forgot that this thing didn't build...

tooomm:
Conflicts resolved, yes.
Builds fine, the rpm packaging error (cpack?) is the same still.

ebbit1q:
probably a bug in cmake, might be necessary to install a newer version in the build

ebbit1q:
we didn't need the 39 build anyway so it's not worth reverting, I'll make a ticket for 41 not building.

PR #5152, opened 2024-11-01
Application Display Name (#4767)

### LINKED ISSUE #4767: Notifications on Windows show the name of the process/file instead of the app
**System Information:**
Client Version: 2.8.1-custom(4ee1994) (2023-03-04)
Build Architecture: 64-bit
Qt Version: 6.3.2
System Locale: en_GB
Install Mode: Standard

image

de01dwc:
## Related Ticket(s)
- Fixes #4767 

- System toast notifications shows filename instead of application name

- application name will display "Cockatrice" instead of "cockatrice.exe"

## Screenshot
original
resolution

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -119,6 +119,7 @@ int main(int argc, char *argv[])
     QCoreApplication::setOrganizationName("Cockatrice");
     QCoreApplication::setOrganizationDomain("cockatrice.de");
     QCoreApplication::setApplicationName("Cockatrice");
+    QCoreApplication::setApplicationDisplayName("Cockatrice");
     QCoreApplication::setApplicationVersion(VERSION_STRING);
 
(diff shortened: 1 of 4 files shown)


tooomm:
Checking the Qt docs, I see that there is no `setApplicationDisplayName` in [QCoreApplication](https://doc.qt.io/qt-6/qcoreapplication.html), but [QGuiAppication](https://doc.qt.io/qt-6/qguiapplication.html) has it.

ZeldaZach (changes requested):
This seems to fail all tests. I don't see an ApplicationDisplayName property in QCoreApplication, as Tom pointed out. Looks like setting the applicationName should be all that's necessary, which is already set for the apps.

tooomm:
>Looks like setting the applicationName should be all that's necessary, which is already set for the apps.

Yet it does show the name of the executable, here in the notification on Windows 11:
image

PR #5156, opened 2024-11-03
New signal cardDatabaseLoadingFinished() for CardDatabase to mirror cardDatabaseLoadingFailed()

BruebachL:
## Related Ticket(s)
- None

The card database currently does not notify UI components when it has finished loading the database, only when it fails. This is problematic for widgets that may be displayed at application startup but may rely on a loaded database to display their content. The proposed new signal allows UI components to late initialize when database loading has finished.

- CardDatabase gains a new signal void cardDatabaseLoadingFinished()
-  It will emit this signal in loadCardDatabases(), mirroring the else branch where cardDatabaseLoadingFailed() is emitted.

### DIFF
--- a/cockatrice/src/game/cards/card_database.cpp
+++ b/cockatrice/src/game/cards/card_database.cpp
@@ -567,6 +567,7 @@ LoadStatus CardDatabase::loadCardDatabases()
     if (loadStatus == Ok) {
         checkUnknownSets(); // update deck editors, etc
         qDebug() << "CardDatabase::loadCardDatabases success";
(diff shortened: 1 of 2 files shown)


ebbit1q:
is this signal just added but unused?

ebbit1q (approved):

BruebachL:
Yes, I have three uncommitted pull requests which rely on this functionality. Rather than including it in each one, I have chosen to refactor it as a separate atomic commit so I may rebase my other branches on top of it.

ebbit1q (approved):

PR #5157, opened 2024-11-03
Do not open card reveal windows when skipping in replays

### LINKED ISSUE #5138: Do not open card reveal windows when skipping in replays
**Description of Problem**
When watching a replay, clicking on the timeline to skip to a section will cause all card reveal windows between the current time and the new time to open at once, requiring you to manually close all of them.

This issue becomes more pronounced when skipping backwards, since cockatrice needs to replay all events from the beginning.  For example, near the end of a game where plenty of hand attack happened, every time you try to skip slightly backwards, it will flood the play area with several hand reveal windows, covering up the board and requiring you to manually close all of them.

When skipping in replays, card reveal windows should not open.

**Additional Notes**
There are several tiers to the implementation.
- The simplest fix to do is to ignore the window creation from reveal events while rewinding. This removes the most egregious effect of the issue.
- Next up is ignoring card reveal windows when skipping forward. However, that might not be ideal, because knowing what reveals just happened might be important to the context of the game.
- The most advanced implementation would be to, when skipping, only open card reveal windows that occurred within the last X seconds of the target location. This would provide a good balance between still providing context when skipping and not spamming the screen with card reveal windows.

RickyRister:
## Related Ticket(s)
- Fixes #5138 

When watching a replay, skipping will cause all card reveal windows from events processed during the skip to open at once. This is extra apparent when skipping backwards, since cockatrice needs to reprocess all events from the start.

https://github.com/user-attachments/assets/0855a8c3-105e-494b-b8d6-6d47b9ca687b
- Create a new `EventProcessingOptions` `QFlag` that will be passed along in any signals that result in event processing.
  - only has `SKIP_REVEAL_WINDOWS` for now, but I'm making it a `QFlag` in case we want to add more options in the future.
- The event processing code in `player.cpp` for reveal events will skip creating a window if an enabled flag is passed in.
- Added logic to `ReplayTimelineWidget` for determining when the signal should have the flag enabled
  - When forward skipping, we still open the window if reveal event is within 10 seconds of the new time. 
    - This is to coincide with the length of the large forward skip shortcut.
    - If you're forward skipping with shortcuts, you'll still be able to see all card reveal windows as you go over them.

### DIFF
--- a/cockatrice/src/client/network/replay_timeline_widget.cpp
+++ b/cockatrice/src/client/network/replay_timeline_widget.cpp
@@ -157,7 +157,7 @@ void ReplayTimelineWidget::replayTimerTimeout()
 void ReplayTimelineWidget::processNewEvents()
 {
     while ((currentEvent < replayTimeline.size()) && (replayTimeline[currentEvent] < currentTime)) {
-        emit processNextEvent();
+        emit processNextEvent({});
         ++currentEvent;
     }
     if (currentEvent == replayTimeline.size()) {
--- a/cockatrice/src/client/network/replay_timeline_widget.cpp
+++ b/cockatrice/src/client/network/replay_timeline_widget.cpp
@@ -93,7 +93,7 @@ void ReplayTimelineWidget::skipToTime(int newTime, bool doRewindBuffering)
     if (isBackwardsSkip) {
         handleBackwardsSkip(doRewindBuffering);
     } else {
-        processNewEvents();
+        processNewEvents(FORWARD_SKIP);
     }
 
     update();
@@ -130,7 +130,7 @@ void ReplayTimelineWidget::processRewind()
     // process the rewind
     currentEvent = 0;
     emit rewound();
-    processNewEvents();
+    processNewEvents(BACKWARD_SKIP);
 }
 
 QSize ReplayTimelineWidget::sizeHint() const
@@ -147,17 +147,24 @@ void ReplayTimelineWidget::replayTimerTimeout()
 {
     currentTime += 200;
 
-    processNewEvents();
+    processNewEvents(NORMAL_PLAYBACK);
 
     if (!(currentTime % 1000))
         update();
 }
 
 /// Processes all unprocessed events up to the current time.
-void ReplayTimelineWidget::processNewEvents()
+void ReplayTimelineWidget::processNewEvents(PlaybackMode playbackMode)
 {
     while ((currentEvent < replayTimeline.size()) && (replayTimeline[currentEvent] < currentTime)) {
-        emit processNextEvent({});
+        Player::EventProcessingOptions options;
+
+        // backwards skip => always skip reveal windows
+        // forwards skip => skip reveal windows that don't happen within 10 seconds of the target
+        if (playbackMode == BACKWARD_SKIP || currentTime - replayTimeline[currentEvent] > 10000)
+            options |= Player::EventProcessingOption::SKIP_REVEAL_WINDOW;
+
+        emit processNextEvent(options);
         ++currentEvent;
     }
     if (currentEvent == replayTimeline.size()) {
--- a/cockatrice/src/game/player/player.h
+++ b/cockatrice/src/game/player/player.h
@@ -339,7 +339,7 @@ private:
     void eventDestroyCard(const Event_DestroyCard &event);
     void eventAttachCard(const Event_AttachCard &event);
     void eventDrawCards(const Event_DrawCards &event);
-    void eventRevealCards(const Event_RevealCards &event);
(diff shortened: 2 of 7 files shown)


ebbit1q:
this pr depends on #5141 right?

RickyRister:
> this pr depends on #5141 right?

Yes

ZeldaZach (approved):
LGTM

ebbit1q:
this pr needs to be rebased, it seems like it diverged from #5141

cockatrice/src/game/player/player.h
```diff
@@ -226,6 +226,13 @@ private slots:
 
     void initSayMenu();
 
+public:
+    enum EventProcessingOption
+    {
+        SKIP_REVEAL_WINDOW = 0x0001
```
ebbit1q, line 232: this all seems rather complicated, why not make this a simple bool? are there any special eventprocessingoptions you have in mind?

RickyRister: I mostly made this an enum because adding a `skipRevealWindow` bool param to every `processEvent` down the chain felt ugly and felt like it was mixing two different levels of abstraction. Maybe adding one bool param is fine, but that just encourages us to keep adding more bool params to it in the future. Making it a flag enum makes it cleaner to extend in the future.

There’s probably some more events that we can save processing effort on during replays. I’m looking into skipping tap animations when rewinding. Might also look into if we can do anything about draw arrow events.

ebbit1q: if the only case of needing things to be skipped is during replays, why not make it a bool similar to "rewindingReplay"

RickyRister: Well, more specifically, `SKIP_REVEAL_WINDOWS` is enabled when rewinding **or** when forward skipping but only if it's more than a certain amount. I suppose we could just have a single boolean represent that condition, but in the future we might want something to happen on a different combination of conditions.

For example, I'm currently looking into skipping the tap/untap animation when rewinding. I've found that it looks nicer if we don't skip tap animations when forward skipping. So now I want the `SKIP_TAP_ANIMATIONS` flag to only be active when rewinding and not when forward skipping. If we used a single bool, we wouldn't be able to do that.

An alternative option we could do is, instead of having a flag enum that represents specific event processing flags that are decided at the point of event generation, we could instead pass an object in the signal that contains info about the playback state (replay vs actual game, skipping vs normal playback, rewinding, etc). Then we move the decision-making logic inside `Player`. That way all the conditional logic is contained in `Player` instead of the event generators.

Thoughts?

ebbit1q: either would work, benefits of keeping stuff in the player is that this'll give more context to the player, perhaps making other things simpler, benefits of keeping things in the replay tab is that it stays more centralized.

RickyRister: I think I prefer going with the current option for now. 

Going with the other option requires us to decide on the structure of the object that we're passing, as well as all the relevant states that we might want to represent. We might want to look into option 2 in the future if we end up with repeated conditional logic in the event generation code in multiple places. For now though, I'm mostly working on replay-related stuff, so it's more convenient to keep the logic centralized in the replay code.

ebbit1q: sounds good

cockatrice/src/client/network/replay_timeline_widget.cpp
```diff
@@ -147,17 +147,24 @@ void ReplayTimelineWidget::replayTimerTimeout()
...
-void ReplayTimelineWidget::processNewEvents()
+void ReplayTimelineWidget::processNewEvents(PlaybackMode playbackMode)
 {
     while ((currentEvent < replayTimeline.size()) && (replayTimeline[currentEvent] < currentTime)) {
-        emit processNextEvent();
+        Player::EventProcessingOptions options;
+
+        // backwards skip => always skip reveal windows
+        // forwards skip => skip reveal windows that don't happen within 10 seconds of the target
+        if (playbackMode == BACKWARD_SKIP || currentTime - replayTimeline[currentEvent] > 10000)
```
ebbit1q, line 164: I'd like this 10 000 to be made into a constant with unit

RickyRister: I already have a PR in the works to move all the magic numbers in ReplayTimeWidget into constants, but it builds upon this PR and so needs it to be merged in first

ebbit1q: so you'd like me to merge this in before that?

RickyRister: yeah

ebbit1q (approved):
:+1:

Issue #486, opened 2014-12-11
Multiple Arts For A Deck


ZeldaZach:
A few individuals have requested making it possible to have multiple land arts in the deck.

My proposal is to use a system similar to TCGPlayer and how they handle lands:
`Island (#240)`
`Island (#241)`
etc

Any opinions on this?

Daenyth:
To select which printing to use in your own deck, right click the card in your decklist.

Deck hash code should not consider printing at all; two decklists with the same cards in any combination of different printings should not be any different in hash code. This could be used as a small UI feature of draft sites, providing a .cod file with printings specific to the set being drafted

Fenrirthviti:
Personally, I am a fan of the way tappedout handles it from a human-readable standpoint, where the set name is used to define the art, with an additional flag for foil. Example follows:

```
1x Island (ISD)
2x Island (MBS)
2x Island (NPH)
1x Island (SOM)
1x Island (AVR) *F*
1x Island (ISD) *F*
1x Island (MMQ) *F*
1x Island (PDS) *F*
1x Island (UNH) *F*
```

However, any way this is implemented would be fantastic.

ebbit1q:
linking https://github.com/Cockatrice/Cockatrice/issues/2545 https://github.com/Cockatrice/Cockatrice/issues/473

gndowns:
hey @ebbit1q  is this still being worked on? Would love this as a feature

ebbit1q:
if you'd like to contribute on this it would be much appreciated, nothing has changed in regards of the goal of this ticket

ZeldaZach:
Thinking more about this, and how we can handle it.

I'd like to see a change to how COD files are formatted. 

Current
```xml
<?xml version="1.0" encoding="UTF-8"?>
<cockatrice_deck version="1">
    <deckname></deckname>
    <comments></comments>
    <zone name="main">
        <card number="1" name="_____-o-saurus"/>
        <card number="2" name="+2 Mace"/>
        <card number="1" name="Abbey Griffin"/>
    </zone>
</cockatrice_deck>
```

Proposed
```xml
<?xml version="1.0" encoding="UTF-8"?>
<cockatrice_deck version="1">
    <deckname></deckname>
    <comments></comments>
    <zone name="main">
        <card number="1" name="_____-o-saurus" setCode="XYZ" collectorNumber="123"/> 
        <card number="2" name="+2 Mace" setCode="XYZ"/>
[...]

PR #5158, opened 2024-11-03
Change CardInfo's PixmapCacheKey to be the UUID of the card in the preferred set after database loading has finished. Otherwise, and if no UUID of a preferred set is available, default to the card name.

BruebachL:
## Related Ticket(s)
- Technical precursor to #486

Currently, we use a CardInfo's name as the key for the Pixmap Cache. This is problematic because it does not allow us to store different images for a card with the same name (i.e. different printings) in the cache because of a key collision.

- Use the more unique UUID of the preferred printing as a key for the Pixmap cache key.
- By default, CardInfo will still assign the CardName as the Pixmap cache key.
- On database load completion, the cache keys are refreshed with the uuid of the CardInfoPerSet of the preferred Set.
- Tokens or other cards where the CardInfoPerSet does not contain a UUID will fall back to the card name.
- The PictureLoader will prepend a set to the sortedSets it uses to populate file download urls if the pixmapCacheKey of the card corresponds with the uuid of the card in the set.

### DIFF
--- a/cockatrice/src/client/ui/picture_loader.cpp
+++ b/cockatrice/src/client/ui/picture_loader.cpp
@@ -178,7 +178,8 @@ void PictureLoaderWorker::processLoadQueue()
         qDebug().nospace() << "PictureLoader: [card: " << cardName << " set: " << setName
                            << "]: Trying to load picture";
 
-        if (CardDatabaseManager::getInstance()->isUuidForPreferredPrinting(cardName, cardBeingLoaded.getCard()->getPixmapCacheKey())) {
+        if (CardDatabaseManager::getInstance()->isUuidForPreferredPrinting(
+                cardName, cardBeingLoaded.getCard()->getPixmapCacheKey())) {
             if (cardImageExistsOnDisk(setName, correctedCardName)) {
                 continue;
             }
--- a/cockatrice/src/client/ui/picture_loader.cpp
+++ b/cockatrice/src/client/ui/picture_loader.cpp
@@ -47,8 +47,8 @@ PictureToLoad::PictureToLoad(CardInfoPtr _card)
         for (const auto &set : card->getSets()) {
             if (QLatin1String("card_") + QString(set.getProperty("uuid")) == card->getPixmapCacheKey()) {
                 long long setIndex = sortedSets.indexOf(set.getPtr());
-                CardSetPtr set_for_card_uuid = sortedSets.takeAt(setIndex);
-                sortedSets.prepend(set_for_card_uuid);
+                CardSetPtr setForCardUUID = sortedSets.takeAt(setIndex);
+                sortedSets.prepend(setForCardUUID);
             }
         }
         // The first time called, nextSet will also populate the Urls for the first set.
--- a/cockatrice/src/client/ui/picture_loader.cpp
+++ b/cockatrice/src/client/ui/picture_loader.cpp
@@ -43,6 +43,14 @@ PictureToLoad::PictureToLoad(CardInfoPtr _card)
             sortedSets << CardSet::newInstance("", "", "", QDate());
         }
         std::sort(sortedSets.begin(), sortedSets.end(), SetDownloadPriorityComparator());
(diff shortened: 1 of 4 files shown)


cockatrice/src/client/ui/picture_loader.cpp
```diff
@@ -43,6 +43,12 @@ PictureToLoad::PictureToLoad(CardInfoPtr _card)
             sortedSets << CardSet::newInstance("", "", "", QDate());
         }
         std::sort(sortedSets.begin(), sortedSets.end(), SetDownloadPriorityComparator());
+        // If the pixmapCacheKey corresponds to a specific set, we have to try to load it first.
+        for (const auto &set : card->getSets()) {
+            if (set.getProperty(QString("uuid")) == card->getPixmapCacheKey()) {
```
ZeldaZach, line 48: I believe "uuid" is sufficient, no need to wrap in a QString.

cockatrice/src/client/ui/picture_loader.cpp
```diff
@@ -43,6 +43,12 @@ PictureToLoad::PictureToLoad(CardInfoPtr _card)
...
+                sortedSets.prepend(sortedSets.takeAt(sortedSets.indexOf(set.getPtr())));
```
ZeldaZach, line 49: There's a lot to this line, maybe try breaking it down a bit to explain what's going on?

cockatrice/src/game/cards/card_database.cpp
```diff
@@ -576,6 +592,63 @@ LoadStatus CardDatabase::loadCardDatabases()
...
+
+    CardSetPtr preferredSet = nullptr;
+    CardInfoPerSet preferredCard;
+    SetPriorityComparator comparator;
+
+    for (auto &set : setMap) {
+        CardSetPtr currentSet = set.getPtr();
+        if (!preferredSet || comparator(currentSet, preferredSet)) {
+            preferredSet = currentSet;
+            preferredCard = set;
```
ZeldaZach, line 622: would a break make sense here?

ZeldaZach, line 622: a card equaling a set is a bit confusing to me, can we rename these?

cockatrice/src/game/cards/card_database.cpp
```diff
@@ -576,6 +592,63 @@ LoadStatus CardDatabase::loadCardDatabases()
...
+    if (preferredSet) {
+        return preferredCard;
+    }
+
+    return CardInfoPerSet(nullptr);
+}
+
+QString CardDatabase::getPreferredPrintingUUIDForCard(const QString &cardName)
+{
+    CardInfoPerSet preferred_set_card_info = getPreferredPrintingForCard(cardName);
```
ZeldaZach, line 635: camelCase across the method please

cockatrice/src/game/cards/card_database.cpp
```diff
@@ -576,6 +592,63 @@ LoadStatus CardDatabase::loadCardDatabases()
...
+    QString preferred_set_uuid = preferred_set_card_info.getProperty(QString("uuid"));
+    if (preferred_set_uuid == QString("")) {
```
ZeldaZach, line 637: isEmpty

ZeldaZach (changes requested):
Overall, I love the idea of moving to UUIDs over card names for the cache. However, there are a few tweaks we still need to do.

cockatrice/src/game/cards/card_database.cpp
BruebachL: I do not think so, since we have to compare through the whole list in case the last entry supersedes the current preferred set.

BruebachL:
I've resolved the changes and will push again once PR #5159 is merged so that I may rebase this PR on it.

I have also changed the key of the PixmapCache to include "card_" to be in line with the old behavior.

ZeldaZach (approved):
LGTM

BruebachL:
Please ignore the latest force push, some branches got tangled in my own repo version.

PR #5161, opened 2024-11-05
fix bug with phase highlighting in replays

### LINKED ISSUE #5160: Phase marker highlighting in replays glitches out after backwards skips
While watching a replay, after doing some backward skips, the phase marker will have multiple phases highlighted.
Screenshot 2024-11-05 at 1 30 03 AM

**Steps to reproduce:**
- Watch a replay
- Skip to end of replay
- Backwards skip a few time through the replay, making sure to skip back through multiple phase changes.

RickyRister:
## Related Ticket(s)
- Fixes #5160

When watching a replay, if you do backward skips, the phase button highlighting will bug out. Multiple phases will stay highlighted.

- The `rewound()` signal from `ReplayTimelineWidget` will now trigger `TabGame` to properly reset the phases.
- Added a `reset()` method to the phase toolbar classes that will immediately reset the `PhaseButton`, skipping any animations.
- Also did some minor refactoring of `PhaseButton::updateAnimation()` when fixing a new bug that I introduced while fixing this bug.

### DIFF
--- a/cockatrice/src/client/ui/phases_toolbar.cpp
+++ b/cockatrice/src/client/ui/phases_toolbar.cpp
@@ -84,6 +84,17 @@ void PhaseButton::updateAnimation()
     update();
 }
 
+/**
+ * @brief Immediately resets the button to the inactive state, without going through the animation.
+ */
+void PhaseButton::reset()
+{
+    activeAnimationTimer->stop();
+    active = false;
+    activeAnimationCounter = highlightable ? 0 : 9;
+    update();
+}
+
 void PhaseButton::mousePressEvent(QGraphicsSceneMouseEvent * /*event*/)
 {
     emit clicked();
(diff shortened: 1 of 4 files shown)


ZeldaZach (approved):

cockatrice/src/client/ui/phases_toolbar.cpp
```diff
@@ -74,13 +74,26 @@ void PhaseButton::updateAnimation()
...
+}
+
+/**
+ * @brief Immediately resets the button to the inactive state, without going through the animation.
+ */
+void PhaseButton::reset()
+{
+    activeAnimationTimer->stop();
+    active = false;
+    activeAnimationCounter = highlightable ? 0 : 9;
```
ebbit1q, line 96: I think this'd read better as a simple if statement
```suggestion
    if (highlightable) {
        activeAnimationCounter = 0;
    }
```

ebbit1q (approved):
thank you

PR #5166, opened 2024-11-09
move replay-related constants into ReplayTimelineWidget

RickyRister:
## Related Ticket(s)
- mentioned in #5157

`ReplayTimelineWidget` has several magic numbers that can better be represented as constants. Additionally, some of those magic numbers are spread across both the class itself as well as the classes that use it (such as `tab_game`). 

We should move the magic numbers into constants and consolidate the constants into one place.

- Moved the replay-related constants that appear in `tab_game` into public static consts in `ReplayTimelineWidget`. 
  - That way all replay-related constants are in one place.
- Moved some magic numbers in `ReplayTimelineWidget` into private static consts.


ebbit1q (approved):
excellent, thanks

PR #5168, opened 2024-11-09
skip tap animation when backward skipping in replays

### LINKED ISSUE #5167: Unnecessary tap/untap animations when skipping backwards in replay
When you backwards skip while watching a replay, all tapped permanents will play the tap animation as the skip happens.

Note that the opposite will happen with permanents that enter tapped. Backwards skipping to a part where that permanent is untapped will always cause it to play the untap animation

https://github.com/user-attachments/assets/05a03ca9-ad4a-4ab3-977a-4bfadc06678b

**Steps to reproduce:**
 - Turn on tap/untap animations
 - Load up any replay where permanents have been tapped
 - Skip to a part where there's tapped permanents that used to be untapped
 - Backwards skip a tiny bit to where the permanents are still tapped
   - You should see those permanents play the tap animation

RickyRister:
## Related Ticket(s)
- Fixes #5167 

Due to how rewinding works, Cockatrice will always play the tap animation when backwards skipping to a permanent that is tapped.

https://github.com/user-attachments/assets/c808b481-b6f3-4cb2-87e1-71479622b5c9

- Skip tap/untap animations when rewinding; the permanents will immediately snap to their present state
  - add `SKIP_TAP_ANIMATION` to `EventProcessingOptions` 
  - events sent during a rewind will have that flag enabled
  - `Player::eventSetCardAttr` will set animation to false if the flag is enabled
- still kept around the animations when forward skipping, because I honestly think the animations looks kinda nice

### DIFF
--- a/cockatrice/src/client/network/replay_timeline_widget.cpp
+++ b/cockatrice/src/client/network/replay_timeline_widget.cpp
@@ -162,6 +162,10 @@ void ReplayTimelineWidget::processNewEvents(PlaybackMode playbackMode)
         if (playbackMode == BACKWARD_SKIP || currentTime - replayTimeline[currentEvent] > BIG_SKIP_MS)
             options |= Player::EventProcessingOption::SKIP_REVEAL_WINDOW;
 
+        // backwards skip => always skip tap animation
+        if (playbackMode == BACKWARD_SKIP)
+            options |= Player::EventProcessingOption::SKIP_TAP_ANIMATION;
+
         emit processNextEvent(options);
         ++currentEvent;
     }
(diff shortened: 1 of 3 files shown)


ZeldaZach:
@RickyRister Do you have a before and after video/gif you can include?

ebbit1q:
it's pretty easy to see yourself when skipping through a replay, all the animations will play after a skip, as if all the actions happened at the same time instead of you skipping through, even when going backwards

RickyRister:
> @RickyRister Do you have a before and after video/gif you can include?

There's a video in the linked issue

ebbit1q (approved):
lgtm

PR #5169, opened 2024-11-11
Don't open deck in new tab if current tab is blank

RickyRister:
## Related Ticket(s)
- Followup to #5143

> I feel like when a decklist is empty there is no need to open a new tab. also, opening a new tab takes a solid couple of seconds for me, probably because of the huge card list.

https://github.com/user-attachments/assets/9d8b0c96-525c-42b2-876a-a08519834b03

When the option `Always open deck in new tab` is enabled, if you try to load a deck while the current deck editor tab is **blank**, the deck will load into the existing tab instead of a new tab.

**Blank**: The decklist is currently unmodified **and** is not saved somewhere.

`New Deck` will still always open a new tab, since if you're using it while the current tab is blank, then there's a good chance you intentionally want to open a new tab.

### DIFF
--- a/cockatrice/src/client/tabs/tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/tab_deck_editor.cpp
@@ -783,7 +783,7 @@ void TabDeckEditor::actNewDeck()
 
 void TabDeckEditor::actLoadDeck()
 {
-    bool openInNewTab = SettingsCache::instance().getOpenDeckInNewTab();
+    bool openInNewTab = SettingsCache::instance().getOpenDeckInNewTab() && !isBlankNewDeck();
 
     if (!openInNewTab && !confirmClose())
         return;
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -783,7 +783,7 @@ void TabDeckEditor::actNewDeck()
 
 void TabDeckEditor::actLoadDeck()
 {
-    bool openInNewTab = SettingsCache::instance().getOpenDeckInNewTab();
+    bool openInNewTab = SettingsCache::instance().getOpenDeckInNewTab() && !isBlankNewDeck();
```
ebbit1q, line 786: this new function is very similar to the confirmClose function, if it is true then confirmClose will also quickly return so it wouldn't be necessary to call both

RickyRister: `confirmClose` only checks for modified and can pop up a window. It's also already used in a few places, so I don't want to touch it right now.

It's probably fine to have a slight bit of redundant checking. I feel like it makes the logic flow better, and the checks aren't expensive anyways.

ebbit1q (approved):
thanks

PR #5173, opened 2024-11-16
Refactor Card picture Widgets and enhance functionality

BruebachL:

Widgets should be called Widgets in QT. Also, I don't like the directory the CardPicture widgets currently reside in. This refactor moves them to a new folder to easily group them and adds two new functionalities. One is the ability to display arbitrary text over a CardPicture and the over is a hover to zoom functionality.

- CardPicture classes are moved.
- CardPictureWithTextOverlayWidget is available
- ZoomToHover is available.


ebbit1q:
this seems nice, did this go somewhere else?

Issue #1740, opened 2016-01-01
Cockatrice Plugins


ZeldaZach:
What if we strip all game-related code from the client and move it to a plugin that users can download? We could then theoretically support additional game-specific features (like the command zone for magic) without directly incorporating it into the client. This would also allow for users to start making their own plugins to alter the client to their needs.

Any feedback on this?

Daenyth:
Yes, this is my high level goal for the app since a long time.

Related: https://github.com/Cockatrice/Cockatrice/issues/329#issuecomment-57204205

ctrlaltca:
I like the idea, and in order to implement it we could:
 - [ ] find and collect all the mtg-specific code hardcoded in cockatrice (eg. zones, counters, game behavior, ..)
 - [ ] for every found item, design a possible way to make the feature generic and configurable/overridable by a plugin. Probably looping through all the places we'll find a generic solution that can be applied on all the places.
 - [ ] Create an interface that exposes the configurable / overridable items and lets plugins re-implement them
 - [ ] People playing in a game should be using the same plugins, so we need to handle this per-game
 - [ ] Provide a set of plugins and let others write them (plugin repository?)

Daenyth:
Those all sound appropriate.

Daenyth:
Related to #478 also

Qriist:
I have written a program that scrapes, organizes, and spits out an xml that has YuGiOh information in a way that cockatrice likes. While it does fudge a few things as a concession to getting relevant info on the screen, (notably, LINK monsters have their link number for their default toughness) the framework is more or less complete and can be reconfigured to appropriate values once this ticket gets implemented.

Additionally, the Pokemon scraper will be done in a few days time. (Same basic framework, woo!)

Daenyth:
Wow @Qriist that's awesome! Do you have a link to the code up?

ZeldaZach:
Just a note, I've updated the Wiki to include a section for user-created sets: https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets

tooomm:
@Qriist the YuGiOh xml file you link to in the wiki is unavailable as of now.
I realized that when restyling and expanding the custom sets wiki page.

Do you maintain a page or did you do a reddit announcement for your file with further information?

Qriist:
Nope, just typo-ed the link. It's fixed now.

danyfire:
[...]

PR #5174, opened 2024-11-16
Add the ability to define starting life total during game creation.

### LINKED ISSUE #4693: Adding a Commander Mode
As a Commander player I find it quite troublesome to have my commander on the stack, it would be nice to add a command zone when we are starting a Commander game.

Also, having a more ergonomic way to count for Commander damages instead of using counters on the side that we are already using for counting floating mana and infect.

The last request isn't for commanders really but it would be nice to have a best way to manage our exile stack. Because technically we can have different exile types, like temporary exile with cascades. When I play cascade, to not mix my exiled card from other sources and my exiled cards from cascade I put the cascade exile on the battlefield but it's starting to get really big when I have to exile 70 cards before getting a proper card that I can cascade (Codie with Hypergenesis build when this one is at the bottom of my lib). So maybe having a way to separate exile stacks, and also putting counters on exiled cards (for effects like Karn)

Imo it would be really great to get something like that and would help to have a better experience in cockatrice

### LINKED ISSUE #4858: Automated game start actions. Set Life, Commander, dice,...
Hello, 

The feature idea would be some kind of customable start game setup actions, with profils maybe depending of game format.
The goal is to automatically set at the beginning of the game actions.

For exemple if you join a multi player commander game you can click a action bouton that will :

- Set your Life to 40
- Put Commander(s) in command zone
- Roll Dice 20 and create a token with result
- Draw first Hand

For a modern profile 
- Set Life to 20
- Roll Dice 20 and create a token with result
- Draw first Hand

Et cetera.
There could be default templates for known formats and customable ones.

What do you dev think about this feature idea?

Sincerely,
Danyfire

BruebachL:
## Related Ticket(s)
- Contradicts #1740 by hard-coding some logic into the server.
- Properly partially closes #4858, which was closed because it related too closely to #1740.
- Properly partially closes #4693, which was closed because it related too closely to #1740.

Currently, Cockatrice will always set the life total to 20 since this is the default life total for "Standard". However, most, or a large portion of our user base, consists of commander players. This small change allows the ServerPlayer to respect the defined game types selected during room setup. If this includes Commander, we instead set the Life Total to 40. Since this change is purely opt-in and highly requested, I propose to merge this despite the above mentioned hard-coding.

- ServerPlayer::setupZones is now passed the room->getGameTypes();
- ServerPlayer::setupZones now checks if the GameType String includes "Commander" and then sets the life total to 40 instead.

### DIFF
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -815,13 +815,14 @@ Server_ProtocolHandler::cmdCreateGame(const Command_CreateGame &cmd, Server_Room
     }
 
     QString description = nameFromStdString(cmd.description());
+    int startingLifeTotal = cmd.has_starting_life_total() ? cmd.starting_life_total() : 20;
 
     // When server doesn't permit registered users to exist, do not honor only-reg setting
     bool onlyRegisteredUsers = cmd.only_registered() && (server->permitUnregisteredUsers());
     Server_Game *game = new Server_Game(
         copyUserInfo(false), gameId, description, QString::fromStdString(cmd.password()), cmd.max_players(), gameTypes,
         cmd.only_buddies(), onlyRegisteredUsers, cmd.spectators_allowed(), cmd.spectators_need_password(),
-        cmd.spectators_can_talk(), cmd.spectators_see_everything(), cmd.starting_life_total(), room);
+        cmd.spectators_can_talk(), cmd.spectators_see_everything(), startingLifeTotal, room);
 
     game->addPlayer(this, rc, asSpectator, asJudge, false);
     room->addGame(game);
--- a/common/server_protocolhandler.cpp
+++ b/common/server_protocolhandler.cpp
@@ -821,7 +821,7 @@ Server_ProtocolHandler::cmdCreateGame(const Command_CreateGame &cmd, Server_Room
     Server_Game *game = new Server_Game(
         copyUserInfo(false), gameId, description, QString::fromStdString(cmd.password()), cmd.max_players(), gameTypes,
         cmd.only_buddies(), onlyRegisteredUsers, cmd.spectators_allowed(), cmd.spectators_need_password(),
-        cmd.spectators_can_talk(), cmd.spectators_see_everything(), room);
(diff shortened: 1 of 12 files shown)


ebbit1q:
gametypes aren't even defined anywhere, it's a string that can be set inside the server config, I'd much prefer an option that defines the starting life total on game creation.

ZeldaZach (changes requested):
Agree with ebbit on this one. I'd rather extend the database to include additional setup parameters rather than hard code components like this.

BruebachL:
How would you like to define this if not over the room creation options? Afaik that's the only time we allow the user to set any sort of options.

ebbit1q:
when creating the game seems like the most sensible option to me

BruebachL:
image
image

ZeldaZach (approved):
This makes more sense to me than the original PR. Not sure how many other fields users might want to configure pre-game, but this is a fine starting place. Can always expand the menu later as needed.

common/server_protocolhandler.cpp
```diff
@@ -821,7 +821,7 @@ Server_ProtocolHandler::cmdCreateGame(const Command_CreateGame &cmd, Server_Room
     Server_Game *game = new Server_Game(
         copyUserInfo(false), gameId, description, QString::fromStdString(cmd.password()), cmd.max_players(), gameTypes,
         cmd.only_buddies(), onlyRegisteredUsers, cmd.spectators_allowed(), cmd.spectators_need_password(),
-        cmd.spectators_can_talk(), cmd.spectators_see_everything(), room);
+        cmd.spectators_can_talk(), cmd.spectators_see_everything(), cmd.starting_life_total(), room);
```
ebbit1q, line 824: I don't think this is backwards compatible this way, it needs a default value in case the cmd does not have starting life total set like an old client would, by the use of cmd.has_starting_life_total

BruebachL: Adressed in [e068302](https://github.com/Cockatrice/Cockatrice/pull/5174/commits/e06830209c487f2e38130640ea82880068646069)

PR #5175, opened 2024-11-16
Add a button to swap the card between mainboard and sideboard to the DeckEditor

### LINKED ISSUE #4030: Deck Editor: add "swap between maindeck and sideboard" button near "+ - X" buttons
It'd be helpful to be able to swap a card between sideboard and main deck without having to search for the card name (possibly after removing filters), re-add it into main deck, and delete it from sideboard (or vice versa). 

I suggest adding a "swap" button, maybe using the "( )" curvy arrows symbol.

BruebachL:
## Related Ticket(s)
- Closes #4030

Some people wished for a button to easily swap cards between mainboard and sideboard. Since this is an accessibility issues, I've added one. Please let me know how to proceed with the icons, I've currently used:

https://www.svgrepo.com/svg/25529/up-and-down-arrows

which is [CC0 License](https://www.svgrepo.com/page/licensing/#CC0)d

- TabDeckEditor gains a new button to swap cards.

### DIFF
--- a/cockatrice/cockatrice.qrc
+++ b/cockatrice/cockatrice.qrc
@@ -29,6 +29,7 @@
         <file>resources/icons/search.svg</file>
         <file>resources/icons/settings.svg</file>
         <file>resources/icons/spectator.svg</file>
+        <file>resources/icons/swap.svg</file>
         <file>resources/icons/sync.svg</file>
         <file>resources/icons/tab_changed.svg</file>
         <file>resources/icons/update.png</file>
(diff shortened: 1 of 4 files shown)


ebbit1q (changes requested):
the resource also has to be added to the cockatrice.qrc file, right now it is an empty button for me, also the decklist needs to be extended in width by another column

in addition I noticed in some situations a sort of desync can happen in the list if an item gets swapped and makes a new row while a row remains:
image

after clicking swap:
image

it doesn't happen all the time though, I think the important thing is to first swap an entire row and then have it automatically select the next one, as soon as the first of the automatically selected row gets moved the bug happens, I suppose it'd be easy to fix with a forced call to update()

BruebachL:
> the resource also has to be added to the cockatrice.qrc file, right now it is an empty button for me, also the decklist needs to be extended in width by another column

Done and done. Good catches, thanks.

> in addition I noticed in some situations a sort of desync can happen in the list if an item gets swapped and makes a new row while a row remains: image
> 
> after clicking swap: image
> 
> it doesn't happen all the time though, I think the important thing is to first swap an entire row and then have it automatically select the next one, as soon as the first of the automatically selected row gets moved the bug happens, I suppose it'd be easy to fix with a forced call to update()

I simply reused TabDeckEditor::actSwapCard for this so if there is a desync, it's an existing issue unrelated to the button. I'm not saying it's not worth looking into, just that it's unrelated to this PR's changes.
I've added a forced update call at the end at it seems to re-sync the changes after a brief delay on my end.

ZeldaZach (approved):

PR #5183, opened 2024-11-20
add "open in new tab" button to decklist confirmation dialogue

RickyRister:
## Related Ticket(s)
- Followup to #5143

Screenshot 2024-11-20 at 1 38 22 AM

When the `Always open deck in new tab` setting is disabled, when opening a deck while the current deck is modified, the confirmation dialogue has an additional `Open in new tab` button.

- refactored to have a `confirmOpen` method as well as a `confirmClose` method
  - moved the settings checking logic to inside of `confirmOpen`
  - `confirmClose` is now only used strictly for closing a deck editor tab without trying to open a new deck
  - `actNewDeck`, `actLoadDeck`, and `actLoadDeckFromClipboard` now all use `confirmOpen`
- Added a "Open in new tab" button to `confirmOpen`'s window that does what you expect it to do.
  - The `confirmOpen` window is otherwise exactly the same as before

### DIFF
--- a/cockatrice/src/client/tabs/tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/tab_deck_editor.cpp
@@ -769,8 +769,9 @@ void TabDeckEditor::actNewDeck()
 {
     auto deckOpenLocation = confirmOpen(false);
 
-    if (deckOpenLocation == CANCELLED)
+    if (deckOpenLocation == CANCELLED) {
         return;
+    }
 
     if (deckOpenLocation == NEW_TAB) {
         emit openDeckEditor(nullptr);
@@ -789,8 +790,9 @@ void TabDeckEditor::actLoadDeck()
 {
     auto deckOpenLocation = confirmOpen();
 
-    if (deckOpenLocation == CANCELLED)
+    if (deckOpenLocation == CANCELLED) {
         return;
+    }
 
     QFileDialog dialog(this, tr("Load deck"));
     dialog.setDirectory(SettingsCache::instance().getDeckPath());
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -767,13 +767,15 @@ void TabDeckEditor::closeRequest()
 
 void TabDeckEditor::actNewDeck()
 {
-    if (SettingsCache::instance().getOpenDeckInNewTab()) {
-        emit openDeckEditor(nullptr);
+    auto deckOpenLocation = confirmOpen(false);
+
+    if (deckOpenLocation == CANCELLED)
```
ZeldaZach, line 772: General Nit: Please use brackets for if statements, even if one liners <3

ebbit1q:
does this deprecate the confirmClose function?

RickyRister:
> does this deprecate the confirmClose function?

No, we still need `confirmClose` for when we directly close the tab or exit cockatrice

ebbit1q:
hmm, that kinda duplicates that part, the confirmOpen dialog has the same text, I'd prefer to have one place where this almost the same dialog is made, maybe add an option for it to not make the new tab button?

ebbit1q:
also, do we have a fitting icon for the new button?
image

ebbit1q:
so these are so called, standard buttons, they have os dependent behavior and draw icons from the selected theme I think.

RickyRister:
> maybe ordering the functions to be closer to the ones they are similar to might be nice? (confirmOpen confirmClose?)

`confirmClose` is public and `confirmClose` is private

ebbit1q:
hmm, strange, not sure why that'd be

PR #5186, opened 2024-11-22
Cache redirects properly by implementing our own QSettings cache for urls.

### LINKED ISSUE #4989: url cache is not used when offline
**Similar Requests**
I searched for similar issues and there were some related ones (e.g. https://github.com/Cockatrice/Cockatrice/issues/4319) , that gave me an understanding of how I think the image downloading and caching currently works, but I could be mistaken.

Keep all the images downloaded on the fly, so they do not need to be ever redownloaded. A user can force redownload simply by deleting the image that he wants replaced.

**Additional Context**
Each of my Cockatrice sessions, the app redownloads all images of decks loaded into games. I can stop this behavior by downloading all images I need for my decks using an external script directly calling the scryfall API, storing these image files in the `../Cockatrice/pics/CUSTOM` directory, and disabling the `download on the fly` option in settings/Card Sources. But then any image I may have missed is not downloaded and not shown. If I enable downloading on the fly, all images are always downloaded again (determined by the slower loading time and in the debug log) and the files in CUSTOM are ignored.

I'm not entirely clear on how the caching currently works, but why isn't there an option to keep any image that was downloaded on the fly? Or is there an issue with my installation and is this feature request how the app is supposed to work? Some issues mention the folder `../Cockatrice/pics/downloadedPics` but I do not have that one. If I manually make one and populate it with images, they are still ignored if downloading on the fly is enabled.

BruebachL:
## Related Ticket(s)
-  Fixes #4989

We cache the final url of a request but if we get redirected, we don't cache the fact that the original url has lead us to a redirect. This leads to us having to issue a network request for every original url, only to fetch a redirect, realize that we've cached it, and then load the cached url. This change introduces our own cache where we store original urls and their redirection urls, so we can skip the network request and load straight from cache both times. This results in a significant speedup for loading cached pictures.

- PictureLoaderWorker gains a new redirect cache and cache related functions.

### DIFF
--- a/cockatrice/src/client/ui/picture_loader.cpp
+++ b/cockatrice/src/client/ui/picture_loader.cpp
@@ -136,6 +136,8 @@ PictureLoaderWorker::PictureLoaderWorker()
 #endif
     auto cache = new QNetworkDiskCache(this);
     cache->setCacheDirectory(SettingsCache::instance().getNetworkCachePath());
+    cache->setMaximumCacheSize(1024L * 1024L *
+                               static_cast<qint64>(SettingsCache::instance().getNetworkCacheSizeInMB()));
     // Note: the settings is in MB, but QNetworkDiskCache uses bytes
     connect(&SettingsCache::instance(), &SettingsCache::networkCacheSizeChanged, cache,
             [cache](int newSizeInMB) { cache->setMaximumCacheSize(1024L * 1024L * static_cast<qint64>(newSizeInMB)); });
--- a/cockatrice/src/client/ui/picture_loader.cpp
+++ b/cockatrice/src/client/ui/picture_loader.cpp
@@ -145,9 +145,12 @@ PictureLoaderWorker::PictureLoaderWorker()
     networkManager->setRedirectPolicy(QNetworkRequest::ManualRedirectPolicy);
     connect(networkManager, SIGNAL(finished(QNetworkReply *)), this, SLOT(picDownloadFinished(QNetworkReply *)));
 
-    cacheFilePath = "redirectCache.ini";
+    cacheFilePath = SettingsCache::instance().getRedirectCachePath() + "cacheSettings.ini";
     loadRedirectCache();
     cleanStaleEntries();
+
+    connect(QCoreApplication::instance(), &QCoreApplication::aboutToQuit, this,
+            &PictureLoaderWorker::saveRedirectCache);
 
     pictureLoaderThread = new QThread;
     pictureLoaderThread->start(QThread::LowPriority);
@@ -156,7 +159,6 @@ PictureLoaderWorker::PictureLoaderWorker()
 
 PictureLoaderWorker::~PictureLoaderWorker()
 {
-    saveRedirectCache();
     pictureLoaderThread->deleteLater();
 }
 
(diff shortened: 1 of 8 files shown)


cockatrice/src/client/ui/picture_loader.cpp
```diff
@@ -145,18 +145,20 @@ PictureLoaderWorker::PictureLoaderWorker()
     networkManager->setRedirectPolicy(QNetworkRequest::ManualRedirectPolicy);
     connect(networkManager, SIGNAL(finished(QNetworkReply *)), this, SLOT(picDownloadFinished(QNetworkReply *)));
 
-    cacheFilePath = "redirectCache.ini";
+    cacheFilePath = SettingsCache::instance().getRedirectCachePath() + "cacheSettings.ini";
```
ZeldaZach, line 148: I'd call it just "cache.ini" personally, since the directory is called redirect. Frankly, do we need to put this into a folder, can we call it "redirectCache" again like you had and just drop it in the /pics folder, next to CUSTOM?

cockatrice/src/client/ui/picture_loader.h
```diff
@@ -83,6 +83,9 @@ class PictureLoaderWorker : public QObject
     QList<PictureToLoad> loadQueue;
     QMutex mutex;
     QNetworkAccessManager *networkManager;
+    QHash<QUrl, QPair<QUrl, QDateTime>> redirectCache; // Stores redirect and timestamp
+    QString cacheFilePath;                             // Path to persistent storage
+    int CacheTTLInDays = 30;
```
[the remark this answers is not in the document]
ZeldaZach: User specified maybe? with 30d default?

cockatrice/src/client/ui/picture_loader.cpp
```diff
@@ -136,6 +132,8 @@ PictureLoaderWorker::PictureLoaderWorker()
 #endif
     auto cache = new QNetworkDiskCache(this);
     cache->setCacheDirectory(SettingsCache::instance().getNetworkCachePath());
+    cache->setMaximumCacheSize(1024L * 1024L *
```
ebbit1q, line 135: this is rather confusing to me, not to mention that the L for long int is a different size on different oses and will get cast up to 64bit anyway in the end

cockatrice/src/client/ui/picture_loader.cpp
```diff
@@ -447,11 +452,104 @@ bool PictureLoaderWorker::imageIsBlackListed(const QByteArray &picData)
...
+    settings.endArray();
+}
+
+void PictureLoaderWorker::cleanStaleEntries()
+{
+    QDateTime now = QDateTime::currentDateTimeUtc();
+
+    auto it = redirectCache.begin();
+    while (it != redirectCache.end()) {
+        if (it.value().second.addDays(SettingsCache::instance().getRedirectCacheTtl()) < now) {
```
ebbit1q, line 547: this seems suboptimal, why not compare to now - getredirectcachettl?

cockatrice/src/client/ui/picture_loader.h
```diff
@@ -83,6 +89,9 @@ class PictureLoaderWorker : public QObject
     QList<PictureToLoad> loadQueue;
     QMutex mutex;
     QNetworkAccessManager *networkManager;
+    QHash<QUrl, QPair<QUrl, QDateTime>> redirectCache; // Stores redirect and timestamp
+    QString cacheFilePath;                             // Path to persistent storage
+    static constexpr int CacheTTLInDays = 30;          // TODO: Make user configurable
```
ebbit1q, line 94: is this used still?

PR #5192, opened 2024-11-24
fix type mismatch from QT_STATE_CHANGED_T

RickyRister:

`QObject::connect: No such slot UserInterfaceSettingsPage::setNotificationEnabled(Qt::CheckState) in /Users/Ricky/GitHub/Cockatrice/cockatrice/src/dialogs/dlg_settings.cpp:448`

Looks like we missed one when we changed the types in #5137.

- fixed the type of the parameter for `setNotificationEnabled`

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -524,7 +524,7 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     setLayout(mainLayout);
 }
 
-void UserInterfaceSettingsPage::setNotificationEnabled(int i)
+void UserInterfaceSettingsPage::setNotificationEnabled(QT_STATE_CHANGED_T i)
 {
     specNotificationsEnabledCheckBox.setEnabled(i != 0);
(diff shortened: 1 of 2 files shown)


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -524,7 +524,7 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     setLayout(mainLayout);
 }
 
-void UserInterfaceSettingsPage::setNotificationEnabled(int i)
+void UserInterfaceSettingsPage::setNotificationEnabled(QT_STATE_CHANGED_T i)
 {
     specNotificationsEnabledCheckBox.setEnabled(i != 0);
```
ZeldaZach, line 529: If this isn't an int, how can we compare against 0?

RickyRister: i guess because c++ enums are just ints under the hood

ebbit1q: yes, they get cast automatically, most things are ints under the hood

PR #5194, opened 2024-11-24
clean up DownloadSettings

### LINKED ISSUE #5193: Deleting the last download url is not reflected in the ini files
**System Information:**
Example recorded using release 2.9 but the issue still persists in master.

In the download priority list (under `Preferences` -> `Card Sources` -> `URL Download Priority`), when the last download url is deleted, the change is not reflected in the settings file (in `Cockatrice/settings/downloads.ini`). `downloads.ini` will still have the final url. Upon restarting Cockatrice, the download priority list will once again contain that url.

Forcing a card image download while the list is empty *will* cause the change to be reflected. (By causing the download urls to be reset to the default gatherer/scryfall urls)

https://github.com/user-attachments/assets/5acbf383-f8a9-4a66-860a-488d21d17a5d

**Steps to reproduce:**
- Open `Preferences` -> `Card Sources` when you have one or more download urls in the download priority list
- Check `Cockatrice/settings/downloads.ini` to confirm that it contains all those download urls
- Start deleting the download urls in the download priority list
  - Confirm that after each deletion, `downloads.ini` is updated to reflect the change.
- Delete the last download url in the download priority list
  - Confirm that `download.ini` still has the last download url remaining.
- Restart Cockatrice 
  - The last download url should be in the download priority list again

RickyRister:
## Related Ticket(s)
- Fixes #5193 

In addition to the above bug, `DownloadSettings` (and the usage of it) feels really "beat around the bush" for no reason. 

- Refactored `DownloadSettings` and its usage to be more direct.
- **Functionality change:** automatic resetting of urls to default now only happens on first run after clean install instead of whenever the urls are empty.
  - It felt weird to me that having an empty url list will forcibly cause it to be reset to default. 
    - There *are* valid reasons to have an empty url list. For example, if you're only playing with custom images.
    - If you *do* want to reset the urls because you messed up, you can always just manually press the reset button.

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -588,8 +588,9 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     connect(urlList->model(), SIGNAL(rowsMoved(const QModelIndex, int, int, const QModelIndex, int)), this,
             SLOT(urlListChanged(const QModelIndex, int, int, const QModelIndex, int)));
 
-    foreach (QString url, SettingsCache::instance().downloads().getAllURLs())
+    for (QString url : SettingsCache::instance().downloads().getAllURLs()) {
         urlList->addItem(url);
+    }
 
     auto aAdd = new QAction(this);
     aAdd->setIcon(QPixmap("theme:icons/increment"));
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -588,8 +588,8 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     connect(urlList->model(), SIGNAL(rowsMoved(const QModelIndex, int, int, const QModelIndex, int)), this,
             SLOT(urlListChanged(const QModelIndex, int, int, const QModelIndex, int)));
 
-    for (int i = 0; i < SettingsCache::instance().downloads().getCount(); i++)
-        urlList->addItem(SettingsCache::instance().downloads().getDownloadUrlAt(i));
+    foreach (QString url, SettingsCache::instance().downloads().getAllURLs())
+        urlList->addItem(url);
 
     auto aAdd = new QAction(this);
     aAdd->setIcon(QPixmap("theme:icons/increment"));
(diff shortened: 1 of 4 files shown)


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -588,8 +588,8 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     connect(urlList->model(), SIGNAL(rowsMoved(const QModelIndex, int, int, const QModelIndex, int)), this,
             SLOT(urlListChanged(const QModelIndex, int, int, const QModelIndex, int)));
 
-    for (int i = 0; i < SettingsCache::instance().downloads().getCount(); i++)
-        urlList->addItem(SettingsCache::instance().downloads().getDownloadUrlAt(i));
+    foreach (QString url, SettingsCache::instance().downloads().getAllURLs())
```
ZeldaZach, line 591: Please don't use the Qt foreach, but instead the CPP for (auto i : list) approach

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -588,8 +588,9 @@ DeckEditorSettingsPage::DeckEditorSettingsPage()
     connect(urlList->model(), SIGNAL(rowsMoved(const QModelIndex, int, int, const QModelIndex, int)), this,
             SLOT(urlListChanged(const QModelIndex, int, int, const QModelIndex, int)));
 
-    for (int i = 0; i < SettingsCache::instance().downloads().getCount(); i++)
-        urlList->addItem(SettingsCache::instance().downloads().getDownloadUrlAt(i));
+    for (QString url : SettingsCache::instance().downloads().getAllURLs()) {
+        urlList->addItem(url);
```
ebbit1q, line 592: why do we use a loop here while at line 700 we use addItems to add all in a single line?

RickyRister: good point

cockatrice/src/settings/download_settings.cpp
```diff
@@ -5,53 +5,24 @@
...
-    return downloadURLs;
+    return getValue("urls", "downloads").toStringList();
 }
 
-void DownloadSettings::populateDefaultURLs()
+void DownloadSettings::resetToDefaultURLs()
 {
-    downloadURLs.clear();
+    auto downloadURLs = QStringList();
     downloadURLs.append("https://api.scryfall.com/cards/!set:uuid!?format=image&face=!prop:side!");
```
ebbit1q, line 23: I think making this into a static list would be a worthwhile refactor

ebbit1q (approved):

PR #5196, opened 2024-11-25
add menu option to reload card db

RickyRister:
- add action to reload card db under deck builder tab -> `Card Database`

## Screenshots
Screenshot 2024-11-24 at 11 40 50 PM


ebbit1q:
ah, this is neat, I used to open the set card database location dialog for this

PR #5199, opened 2024-11-27
leave some documentation on Zone classes

RickyRister:
I left some documentation as I was trying to figure out how the code works.

- left some documentation on a bunch of classes related to zones
- small refactor in `ZoneViewWidget` constructor

### DIFF
--- a/cockatrice/src/game/zones/view_zone_widget.cpp
+++ b/cockatrice/src/game/zones/view_zone_widget.cpp
@@ -17,6 +17,13 @@
 #include <QStyleOption>
 #include <QStyleOptionTitleBar>
 
+/**
+ * @param _player the player the cards were revealed to.
+ * @param _origZone the zone the cards were revealed from.
+ * @param numberCards num of cards to reveal from the zone. Ex: scry the top 3 cards. Use -1 to reveal the entire zone.
+ * @param _revealZone if false, the cards will be face down.
+ * @param _writeableRevealZone whether the player can interact with the revealed cards.
+ */
 ZoneViewWidget::ZoneViewWidget(Player *_player,
                                CardZone *_origZone,
                                int numberCards,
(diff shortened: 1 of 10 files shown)


cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -33,6 +40,7 @@ ZoneViewWidget::ZoneViewWidget(Player *_player,
     QGraphicsLinearLayout *vbox = new QGraphicsLinearLayout(Qt::Vertical);
     QGraphicsLinearLayout *hPilebox = 0;
 
+    // If the number is < 0 then it means that we can make the area sorted and we dont care about the order.
```
ebbit1q, line 43: when the entire deck is shown we still have the option to show it unsorted, it's sorted by default yeah

cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -57,13 +65,13 @@ ZoneViewWidget::ZoneViewWidget(Player *_player,
         QGraphicsProxyWidget *pileViewProxy = new QGraphicsProxyWidget;
         pileViewProxy->setWidget(&pileViewCheckBox);
         hPilebox->addItem(pileViewProxy);
-    }
 
-    if (_origZone->getIsShufflable() && (numberCards == -1)) {
```
ebbit1q, line 62: are you sure == -1 and <0 are the same in our code?

RickyRister: does the code use negative ints beyond just -1 to encode even more info?

ebbit1q: I'm not sure, we'd have to check

ebbit1q: if it does appear that we only use -1, I think it might be best for readability if we add a const with value -1 and == to that

RickyRister: seems like the constructor does get called with `-2` somewhere
https://github.com/Cockatrice/Cockatrice/blob/f2b0fa164e4ba7140f54f0633a3e27c1be8f79cc/cockatrice/src/game/game_scene.cpp#L172

ebbit1q: allright, important to keep track of that!

ebbit1q (approved):
lgtm

PR #5206, opened 2024-11-29
add sort options to card view window

### LINKED ISSUE #4417: option to sort views
It would be nice if you could change the order of how cards are displayed in a graveyard popup so you can bring together all cards targetable by something or things like that. Maybe a button for "Custom Sort" would be in order so you can't mess it up accidentally?

### LINKED ISSUE #4949: A checkbox to filter or sort deck window by mana value / converted mana cost for "Pod" Style decks
Description of Change

As the name implies, it would be nice to have a checkbox when viewing your deck to sort by decreasing/increasing MV/CMC as a QoL for decks utilizing the various Birthing Pod style effects. Would only really need to be in the deck window as I don't see a use case for it to be in the view hand window but could be a nice to have there if the code ports over cleanly for more efficient turns.

Would preferably also be associated with the following issue whereby the deck list (cards in deck, not the search results) could be sorted alphabetically and by mana value).

Similar Issue

https://github.com/Cockatrice/Cockatrice/issues/4869

RickyRister:
## Related Ticket(s)
- Fixes #4417 
- Fixes #4949

https://github.com/user-attachments/assets/ae5f8a89-a95e-46b9-9422-b98ba6d88542

The `sort by name` and `sort by type` checkboxes have been replaced by two dropdown menus: `Group by X` and `Sort by X`.

With pile view disabled, the cards are first sorted by the `group by` option, and any cards that have the same value for the `group by` property will be sorted within the group with the `sort by` property. If the cards still have the same value, then we break the tie by sorting by name.

With pile view enabled, this will result in the cards displayed in piles based on the `group by` property and sorted within each pile by the `sort by` property.

Setting `Sort by ---` will turn off sorting within each pile. Setting `Group by ---` will turn off grouping (and disable the pile view checkbox until a different option is selected). Set both to `---` if you want to view the deck in its original order.

Cockatrice will automatically prevent you from selecting the same option on both, since it makes no sense to double-sort. If `Group by` option is set to the same value as the `Sort by` option, the `Sort by` option will automatically be set to `Sort by Name`. If you try to set `Sort by` to the same value as `Group by`, it will instead set the value to `Sort by Name`.

The currently selected option for the two dropdown menus will be remembered between games.

## Technical changes
- Refactored `CardList` to have `SortProperty`s that can be passed in to determine what to sort by.
- Add dropdown menus and stuff so that the new properties can be selected and then used in the sort logic.
- Changed a bunch of config stuff because those cached settings got changed from bool to enum (which we store as an int and then cast to an enum as we retrieve it).

### DIFF
--- a/cockatrice/src/game/zones/view_zone_widget.cpp
+++ b/cockatrice/src/game/zones/view_zone_widget.cpp
@@ -141,11 +141,25 @@ void ZoneViewWidget::processGroupBy(int index)
     auto option = static_cast<CardList::SortOption>(groupBySelector.itemData(index).toInt());
     SettingsCache::instance().setZoneViewGroupBy(option);
     zone->setGroupBy(option);
+
+    // reset sortBy if it has the same value as groupBy
+    if (option != CardList::NoSort &&
+        option == static_cast<CardList::SortOption>(sortBySelector.currentData().toInt())) {
+        sortBySelector.setCurrentIndex(1); // set to SortByName
+    }
 }
 
 void ZoneViewWidget::processSortBy(int index)
 {
     auto option = static_cast<CardList::SortOption>(sortBySelector.itemData(index).toInt());
+
+    // set to SortByName instead if it has the same value as groupBy
+    if (option != CardList::NoSort &&
+        option == static_cast<CardList::SortOption>(groupBySelector.currentData().toInt())) {
+        sortBySelector.setCurrentIndex(1); // set to SortByName
+        return;
+    }
+
     SettingsCache::instance().setZoneViewSortBy(option);
     zone->setSortBy(option);
 }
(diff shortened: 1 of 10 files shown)


ZeldaZach (approved):

cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -46,27 +46,43 @@ ZoneViewWidget::ZoneViewWidget(Player *_player,
...
 
-        QGraphicsProxyWidget *sortByNameProxy = new QGraphicsProxyWidget;
-        sortByNameProxy->setWidget(&sortByNameCheckBox);
-        hFilterbox->addItem(sortByNameProxy);
-
-        QGraphicsProxyWidget *sortByTypeProxy = new QGraphicsProxyWidget;
-        sortByTypeProxy->setWidget(&sortByTypeCheckBox);
-        hFilterbox->addItem(sortByTypeProxy);
+        // groupBy options
+        groupBySelector.addItem(tr("Group by ---"), CardList::NoSort);
```
ebbit1q, line 50: if tr() is outside of retranslateui() then when a user changes the language in the settings while this string is out it will not be updated

RickyRister: Yeah. Unfortunately, I'm not sure how to get `retranslateUi` to work with strings that are already inside a `QComboBox`.  I checked around and I don't think we ever handle the combination of `retranslateUi` and `QComboBox`s

PR #5207, opened 2024-11-29
store sort option in settings as QComboBox index instead of enum value

RickyRister:
## Related Ticket(s)
- Followup to #5206

The settings cache now stores the sort option values as the `QComboBox` index instead of the `SortOption` enum value. That way we can add more `SortOption`s and it won't mess up existing settings if the order of the enum entries gets changed. 
(it also just makes the code cleaner)

### DIFF
--- a/cockatrice/src/settings/cache_settings.cpp
+++ b/cockatrice/src/settings/cache_settings.cpp
@@ -252,7 +252,7 @@ SettingsCache::SettingsCache()
     chatMentionColor = settings->value("chat/mentioncolor", "A6120D").toString();
     chatHighlightColor = settings->value("chat/highlightcolor", "A6120D").toString();
 
-    zoneViewGroupByIndex = settings->value("zoneview/groupby", 2).toInt();
+    zoneViewGroupByIndex = settings->value("zoneview/groupby", 1).toInt();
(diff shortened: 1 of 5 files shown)


cockatrice/src/settings/cache_settings.cpp
```diff
@@ -252,8 +252,8 @@ SettingsCache::SettingsCache()
     chatMentionColor = settings->value("chat/mentioncolor", "A6120D").toString();
     chatHighlightColor = settings->value("chat/highlightcolor", "A6120D").toString();
 
-    zoneViewGroupBy = settings->value("zoneview/groupby", 2).toInt();
-    zoneViewSortBy = settings->value("zoneview/sortby", 1).toInt();
+    zoneViewGroupByIndex = settings->value("zoneview/groupby", 1).toInt();
```
ebbit1q, line 255: does this change the default value?

ebbit1q (approved):
seems like a valid fix to the previous pr

cockatrice/src/settings/cache_settings.cpp
RickyRister: `SortByType` is at enum position 2 and `QComboBox` index 1, so the default is still `Sort by Type`

PR #5208, opened 2024-11-30
get retranslateUi to work with sort options

RickyRister:
## Related Ticket(s)
- Followup to #5206

`retranslateUi` doesn't work with sort options

`retranslateUi` now works with sort options.


ZeldaZach (approved):
LGTM works as expected

PR #5210, opened 2024-11-30
Move repeated text in sort options out of the QComboBox

RickyRister:
- The `Group by` and `Sort by` text is now a text label beside the `QComboBox` instead of being part of the entry.
- Forced the `QComboBox` to have a minimum width.
  - This does mean that zone views with small number of cards are now wider than before 
- Small refactors
  - reordered the `QComboBox` resetting in `retranslateUi` because I realized that it was in the opposite order from what we do everywhere else 

## Screenshots
Screenshot 2024-11-29 at 11 11 07 PM
Screenshot 2024-11-29 at 11 12 21 PM
Screenshot 2024-11-29 at 11 10 33 PM

### DIFF
--- a/cockatrice/src/game/zones/view_zone_widget.cpp
+++ b/cockatrice/src/game/zones/view_zone_widget.cpp
@@ -45,13 +45,37 @@ ZoneViewWidget::ZoneViewWidget(Player *_player,
     if (numberCards < 0) {
         QGraphicsLinearLayout *hFilterbox = new QGraphicsLinearLayout(Qt::Horizontal);
 
+        // groupBy label
+        groupByLabel.setFixedHeight(23);
+        groupByLabel.setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
+        groupByLabel.setAlignment(Qt::AlignLeft);
+
+        QGraphicsProxyWidget *groupByLabelProxy = new QGraphicsProxyWidget;
+        groupByLabelProxy->setWidget(&groupByLabel);
+        groupByLabelProxy->setZValue(2000000008);
+        hFilterbox->addItem(groupByLabelProxy);
+
         // groupBy options
+        groupBySelector.setFixedHeight(23);
+
         QGraphicsProxyWidget *groupBySelectorProxy = new QGraphicsProxyWidget;
         groupBySelectorProxy->setWidget(&groupBySelector);
         groupBySelectorProxy->setZValue(2000000008);
         hFilterbox->addItem(groupBySelectorProxy);
 
+        // sortBy label
+        sortByLabel.setFixedHeight(23);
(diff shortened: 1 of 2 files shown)


ZeldaZach:
Before:

image
image
image

cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -184,25 +215,28 @@ void ZoneViewWidget::retranslateUi()
...
+    sortByLabel.setText(tr("Sort by"));
 
-    {
+    { // We can't change the strings after they're put into the QComboBox, so this is our workaround
         int oldIndex = groupBySelector.currentIndex();
         groupBySelector.clear();
-        groupBySelector.addItem(tr("Group by ---"), CardList::NoSort);
-        groupBySelector.addItem(tr("Group by Type"), CardList::SortByType);
-        groupBySelector.addItem(tr("Group by Mana Value"), CardList::SortByManaValue);
+        groupBySelector.addItem(tr("---"), CardList::NoSort);
```
tooomm, line 224: "---" does not need translations. 

You did not want to put a text label like "None"?

ebbit1q: I agree --- does not properly convey that this is the deck's own order, perhaps have it say "unsorted" or "sort by position in deck"?

cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -43,20 +43,49 @@ ZoneViewWidget::ZoneViewWidget(Player *_player,
 
     // If the number is < 0, then it means that we can give the option to make the area sorted
     if (numberCards < 0) {
-        QGraphicsLinearLayout *hPilebox = new QGraphicsLinearLayout(Qt::Horizontal);
         QGraphicsLinearLayout *hFilterbox = new QGraphicsLinearLayout(Qt::Horizontal);
 
+        // groupBy label
+        groupByLabel.setFixedHeight(23);
+        groupByLabel.setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
```
ebbit1q, line 50: by giving the label a fixed width, does it work properly when the text in it changes with a retranslate?

RickyRister:
After some discussion, we decided that there's no need to move the text out of the combo box. The text still looks fine inside the combo box, and it's better for translations to not split the sentence.

ebbit1q:
what about changing --- to something more descriptive though?

RickyRister:
> what about changing --- to something more descriptive though?

I haven't done that yet because I'm indecisive about what to rename it to. `Unsorted` works for the `Sort by ---` option, but I'm unsure about what to rename the `Group by ---` option to. Is `Ungrouped` a word?

Anyways, if someone else can decide on what to rename it to, then they're free to make a PR for the change.

tooomm:
I think what you had in mind is indeed a good representation @rickyrister.

#5232

PR #5221, opened 2024-12-04
Address weird crash case

ZeldaZach:


cockatrice/src/game/player/player.h
```diff
@@ -222,7 +222,7 @@ private slots:
     void actFlowT();
     void actSetAnnotation();
     void actPlay();
-    void actHide();
+    void actHide() const;
```
ebbit1q, line 225: what's the benefit of making this const anyway? it modifies the selected card so I'd hardly consider its behavior const even though it doesn't modify the player object directly

ZeldaZach: Was a recommendation from Clion; more so these methods don't modify the player, so technically they're consts.

ebbit1q: ok, but this function calls CardZone::removeCard which then calls Player::deleteCard, a non const function. Marking this function const while it actually indirectly does non const things is misguiding.

ebbit1q (approved):

ebbit1q:
we can't have too many null checks now can we?

PR #5227, opened 2024-12-06
make rewind buffering timeout configurable

RickyRister:
## Related Ticket(s)
- Follow-up to #5141

- Created new setting for rewind buffering timeout
  - Located under `User Interface` -> `Replay settings`
  - allowed range is 0 to 9999 ms
- Removed the rewind buffering scaling behavior
  - the scaling behavior has always been a compromise to limit the amount of cpu use when spamming rewind near the end of a long replay, by guaranteeing that any reasonable spam frequency there will trigger the buffering.
  - I figured that now users can just set the rewind buffering to a value they're comfortable with, which should consistently make the buffering kick in when they spam the back arrow.

## Screenshots

Screenshot 2024-12-05 at 6 04 26 PM

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -559,7 +559,7 @@ void UserInterfaceSettingsPage::retranslateUi()
     deckEditorGroupBox->setTitle(tr("Deck editor settings"));
     openDeckInNewTabCheckBox.setText(tr("Always open deck in new tab"));
     replayGroupBox->setTitle(tr("Replay settings"));
-    rewindBufferingMsLabel.setText(tr("Amount of time to buffer a backwards skip from keyboard shortcut by:"));
+    rewindBufferingMsLabel.setText(tr("Buffer time for backwards skip via shortcut:"));
     rewindBufferingMsBox.setSuffix(tr(" ms"));
 }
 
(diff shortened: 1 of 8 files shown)


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -544,6 +558,9 @@ void UserInterfaceSettingsPage::retranslateUi()
     tapAnimationCheckBox.setText(tr("&Tap/untap animation"));
     deckEditorGroupBox->setTitle(tr("Deck editor settings"));
     openDeckInNewTabCheckBox.setText(tr("Always open deck in new tab"));
+    replayGroupBox->setTitle(tr("Replay settings"));
+    rewindBufferingMsLabel.setText(tr("Amount of time to buffer a backwards skip from keyboard shortcut by:"));
```
RickyRister, line 562: someone please think of a better name for this

tooomm: `Buffer time for backwards skip via shortcut` ?

ebbit1q:
this seems to not compile on qt 5

cockatrice/src/settings/cache_settings.cpp
```diff
@@ -245,6 +245,7 @@ SettingsCache::SettingsCache()
     minPlayersForMultiColumnLayout = settings->value("interface/min_players_multicolumn", 4).toInt();
     tapAnimation = settings->value("cards/tapanimation", true).toBool();
     openDeckInNewTab = settings->value("editor/openDeckInNewTab", false).toBool();
+    rewindBufferingMs = settings->value("replay/rewindBufferingMs", 200).toBool();
```
ebbit1q, line 248: to bool?

RickyRister: wow that was a bad typo

ZeldaZach (approved):

PR #5229, opened 2024-12-07
"Play top cards until" action now has option for number of hits

### LINKED ISSUE #5205: Add an option to repeat "Take top cards until..." command multiple times

When using the ctrl+shift+Y menu to reveal cards from the top of a player's deck until they reach a card that qualifies, it currently opens a dialogue box containing a single text box that accepts search syntax and nothing else.  My request is to include an option for the quantity of cards to search for (possibly via a separate text box in the same popup, or a "quantity" argument that becomes available in addition to the search syntax, or any similar means).  When used in this way, the process would continue to reveal cards until that many cards that match the search syntax have been revealed.  In most cases, this new variable would default to 1 and continue to function as normal, but this would allow a single command to be used for certain effects that would otherwise require multiple consecutive uses.

In practice, this would simply accept an integer (presumably capped at the user's deck size) and then repeat the typical "reveal cards until..." process that many times.

**Additional Context**

Examples of MTG cards that would benefit from this include Mind Grind, Empty the Laboratory, or Mirko Vosk, Mind Drinker, which each normally require repeating the command multiple times until one reveals the correct number of cards that match the necessary conditions.

RickyRister:
## Related Ticket(s)
- Fixes #5205

https://github.com/user-attachments/assets/0c08a92a-2433-4527-acaf-7c57fd4a7289

`Put top cards on stack until...` action now has an additional option for number of hits. The value can range from 1 to 99. 

Setting the value above one will cause the "put top cards on stack" action to continue until that number of matching cards has been hit, or the deck runs out, whichever one happens first.

The value is saved between uses in the same game, same as the existing behavior for the expr.

Screenshot 2024-12-07 at 3 35 26 AM

## Technical Changes
- Refactored dialogue box for `MoveTopCardsUntil` into separate class in order to allow box to accept two inputs.
- Added some more instance variables to `Player` to track the new stuff
- Refactor moveTopCardUntil stopping and resetting code to `stopMoveTopCardsUntil()` method


cockatrice/src/dialogs/dlg_move_top_cards_until.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <QSpinBox>
+#include <QString>
+#include <QVBoxLayout>
+#include <QWidget>
+
+DlgMoveTopCardsUntil::DlgMoveTopCardsUntil(QWidget *parent, QString _expr, uint _numberOfHits) : QDialog(parent)
+{
+    exprLabel = new QLabel(tr("Card name (or search expressions):"));
+
+    exprEdit = new QLineEdit(this);
```
ebbit1q, line 17: now that we have this class for it that constructs the dialog, it would be nice to add the search syntax help we have on the deck editor somehow (`TabDeckEditor::showSearchSyntaxHelp`), this would mean moving that somewhere else as well so it might be a bit more work, doesn't have to be in this pr of cours

cockatrice/src/dialogs/dlg_move_top_cards_until.h
```diff
@@ -0,0 +1,26 @@
...
+    Q_OBJECT
+
+    QLabel *exprLabel, *numberOfHitsLabel;
+    QLineEdit *exprEdit;
+    QSpinBox *numberOfHitsEdit;
+    QDialogButtonBox *buttonBox;
+
+public:
+    explicit DlgMoveTopCardsUntil(QWidget *parent = nullptr, QString expr = QString(), uint numberOfHits = 1);
+    [[nodiscard]] QString getExpr() const;
```
ebbit1q, line 22: I don't think there is a lot of benefit of adding `nodiscard` to a simple getter, while yeah you probably wouldn't want to discard the return value from this function, you also wouldn't call it if you didn't use the return value, if we were to use `nodiscard` like this we might as well add it to every return value which would just make the code harder to read, imo `nodiscard` is to be used only for functions that are called for their side effects that would return for example an important error value.

cockatrice/src/game/player/player.cpp
```diff
@@ -1330,16 +1331,19 @@ void Player::actMoveTopCardsToExile()
...
+        DlgMoveTopCardsUntil dlg(game, expr, numberOfHits);
+        if (!dlg.exec()) {
             return;
         }
+
+        expr = dlg.getExpr();
+        numberOfHits = dlg.getNumberOfHits();
+
         movingCardsUntilFilter = FilterString(expr);
         if (movingCardsUntilFilter.valid()) {
```
ebbit1q, line 1348: because we now have a custom dialog, the validation can now be moved into the dialog itself, that way we don't have to reopen the dialog when the user enters an invalid filter and we can show the error in the dialog

cockatrice/src/dialogs/dlg_move_top_cards_until.h
RickyRister: i just copied the code from `DlgRollDice` lol. But sure

cockatrice/src/dialogs/dlg_move_top_cards_until.cpp
RickyRister: yeah, something to keep in mind for the future

cockatrice/src/dialogs/dlg_move_top_cards_until.h
ebbit1q: hmm, I wonder who wrote that now

ZeldaZach: The nodiscard is a big recommendation from clion nowadays. Not sure why, but i've been adding it in as appropriate

cockatrice/src/game/player/player.cpp
RickyRister: as an aside, is it possible for the validation to ever fail? Currently (in both release and master branch cockatrice), if I try to put in an invalid search expression, it'll happily accept that (and then proceed to flip the entire deck onto the stack)

ZeldaZach: @RickyRister Might want to add your own "fail" case where you pre-search the deck, and if you find 0 hits mark it as failed? Not sure on this one

RickyRister: Wouldn't that give the player more information than they're legally entitled to, in cases where the contents of the deck aren't known (for example, if cards were exiled face-down)?

ZeldaZach: Sure, but if they enter an invalid filter and flip the entire deck... it's kind of the same issue wouldn't you say?

RickyRister: Here is an example: 
- you have a single card in your starting deck that matches the expression
-  some number of cards are exiled face down from your deck
- You enter the search expression
- The client does a pre-search of the deck
- The client now knows the full contents of the deck
- The client sees that there’s a matching card and starts flipping

The client now knows the full contents of the deck, even though legally it should only know the contents of the deck down to the matching card. This is relevant due to the face down exiled cards, as that means you can now discern the identity of those cards with 100% certainty.

RickyRister: Anyways, seeing as the expression validation for this action doesn’t work on release cockatrice either, we should probably open a separate issue for fixing that.

ebbit1q: > @RickyRister Might want to add your own "fail" case where you pre-search the deck, and if you find 0 hits mark it as failed? Not sure on this one

this really is not possible in the current way this works, this function is very simple, it does not use any new protocol, it simply moves a card to the stack, waits for the next card to be moved to the stack, does a comparison with it, then repeats if necessary

cockatrice/src/dialogs/dlg_move_top_cards_until.h
ebbit1q: I disagree with it, clion is adding things we don't need and it seems like clion is just pretending to be smart :(

cockatrice/src/dialogs/dlg_move_top_cards_until.cpp
```diff
@@ -0,0 +1,66 @@
...
+    setWindowTitle(tr("Put top cards on stack until..."));
+}
+
+void DlgMoveTopCardsUntil::validateAndAccept()
+{
+    auto movingCardsUntilFilter = FilterString(exprEdit->text());
+    if (movingCardsUntilFilter.valid()) {
+        accept();
+    } else {
+        QMessageBox::warning(this, tr("Invalid filter"), movingCardsUntilFilter.error(), QMessageBox::Ok);
```
ebbit1q, line 54: I know this currently doesn't work, but I'd like to instead have a message in the existing dialog instead of another dialog

RickyRister: Can you give me an example of what that should look like?

ebbit1q: a label with red text is most common I think, cockatrice doesn't do this anywhere currently afaik but it's extremely common in other apps and websites, feel free to ignore this for now and we can get back to it later

RickyRister: sure, let's get the validation to actually work first so we can actually see the validation error

ebbit1q (approved):
excellent work

cockatrice/src/dialogs/dlg_move_top_cards_until.cpp
ebbit1q: @RickyRister did we ever ticket this? I can't find it

Issue #2680, opened 2017-05-04
Playing multiple cards from popup doesn't work


LethalPigeon7:
<b>OS:</b>W7 SP1
<b>Cockatrice version:</b>2.3.17-beta004 (2017-04-27)
Drag-selecting multiple cards from a window that pops up when a player reveals a hidden zone, and then right clicking and selecting an action (hide) either  
a) Only affects one of the cards, leaving the rest selected. Whilst selected, right clicking and chosing an option from the drop-down doesn't do anything, requiring cards to be re-selected.
b) Crashes Cockatrice.
This has been a reoccurring problem for a long time (months, I only started using dev snapshots this week), but the crashes are so rare I never realised what was causing them, and the minor inconvenience of having to re-select was always something I forgot throughout the course of the game.  
Edit: [Screenshots](https://imgur.com/a/xgspC) attempting to depict a).

ZeldaZach:
Selecting multiple cards doesn't do the Play or Play Face Down actions. The other ones work as expected. Other parts of this ticket we can't reproduce.

ZeldaZach:
Linking #2707

PR #5233, opened 2024-12-09
hide action now applies to all selected cards

RickyRister:
## Related Ticket(s)
- Partially addresses #2680 

Hide action only applies to the single card instead of the whole selection

https://github.com/user-attachments/assets/d6ad6a1b-df52-4695-8962-bd8be3ab3b5d

Hide action now applies to entire selection. It will work with selections across different reveal zones, but will only affect selected cards that are in zones that would have the hide action.

- `actHide` now loop over the entire selection
  - checks that the card is in a reveal zone before removing the card

--- 
I'm implementing "`play` action to applies to all selected cards" separately, since card properties get messed up when you use to same sort of looping logic to try to play multiple cards at once from the deck.

### DIFF
--- a/cockatrice/src/game/player/player.cpp
+++ b/cockatrice/src/game/player/player.cpp
@@ -3433,6 +3433,18 @@ void Player::actCardCounterTrigger()
     sendGameCommand(prepareGameCommand(commandList));
 }
 
+/**
+ * @brief returns true if the zone is a unwritable reveal zone view (eg a card reveal window)
+ */
+static bool isUnwritableRevealZone(CardZone *zone)
+{
+    if (zone && zone->getIsView()) {
+        auto *view = dynamic_cast<ZoneViewZone *>(zone);
+        return view->getRevealZone() && !view->getWriteableRevealZone();
+    }
+    return false;
+}
+
 void Player::actPlay()
 {
     auto *card = game->getActiveCard();
@@ -3451,7 +3463,7 @@ void Player::actHide()
     }
 
     for (auto &card : selectedCards) {
-        if (card) {
+        if (card && isUnwritableRevealZone(card->getZone())) {
             card->getZone()->removeCard(card);
         }
     }
--- a/cockatrice/src/game/player/player.cpp
+++ b/cockatrice/src/game/player/player.cpp
@@ -3444,9 +3444,16 @@ void Player::actPlay()
 
 void Player::actHide()
 {
-    auto *card = game->getActiveCard();
-    if (card) {
-        card->getZone()->removeCard(card);
+    QList<CardItem *> selectedCards;
+    for (const auto &item : scene()->selectedItems()) {
+        auto *card = static_cast<CardItem *>(item);
+        selectedCards.append(card);
+    }
+
+    for (auto &card : selectedCards) {
+        if (card) {
+            card->getZone()->removeCard(card);
+        }
     }
 }
 
--- a/cockatrice/src/game/player/player.cpp
+++ b/cockatrice/src/game/player/player.cpp
@@ -3434,12 +3434,13 @@ void Player::actCardCounterTrigger()
 }
 
 /**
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/player/player.cpp
```diff
@@ -3433,6 +3433,18 @@ void Player::actCardCounterTrigger()
     sendGameCommand(prepareGameCommand(commandList));
 }
 
+/**
+ * @brief returns true if the zone is a unwritable reveal zone view (eg a card reveal window)
+ */
+static bool isUnwritableRevealZone(CardZone *zone)
+{
+    if (zone && zone->getIsView()) {
+        auto *view = dynamic_cast<ZoneViewZone *>(zone);
```
ZeldaZach, line 3442: What's the purpose of dynamic casting here vs static casting?

cockatrice/src/game/player/player.cpp
```diff
@@ -3433,6 +3433,18 @@ void Player::actCardCounterTrigger()
...
+        return view->getRevealZone() && !view->getWriteableRevealZone();
```
ZeldaZach, line 3443: Can view be nullptr? Might want to include that in the return statement

cockatrice/src/game/player/player.cpp
```diff
@@ -3444,9 +3456,16 @@ void Player::actPlay()
 
 void Player::actHide()
 {
-    auto *card = game->getActiveCard();
-    if (card) {
-        card->getZone()->removeCard(card);
+    QList<CardItem *> selectedCards;
+    for (const auto &item : scene()->selectedItems()) {
```
ZeldaZach, line 3460: Any reason we do this as two separate for loops? Is it for readability?

cockatrice/src/game/player/player.cpp
RickyRister: good point

cockatrice/src/game/player/player.cpp
RickyRister: I didn't know if deleting cards within the loop would cause the contents of `selectedItems()` to get modified as I iterated over it. If it doesn't cause any problems, then I'll get rid of the redundant loop.

ebbit1q: selectedItems returns a list of pointers by value, it does not get modified in the loop

cockatrice/src/game/player/player.cpp
```diff
@@ -3444,9 +3457,11 @@ void Player::actPlay()
 
 void Player::actHide()
 {
-    auto *card = game->getActiveCard();
-    if (card) {
-        card->getZone()->removeCard(card);
+    for (const auto &item : scene()->selectedItems()) {
+        auto *card = static_cast<CardItem *>(item);
+        if (card && isUnwritableRevealZone(card->getZone())) {
```
ebbit1q, line 3462: actHide would only exist on an unwritable reveal zone, we didn't check for this before, why do we need to check for it now?

RickyRister: https://github.com/user-attachments/assets/faa7361d-bd72-4f52-a0b0-ddc97623f20a

If you select cards across multiple zones, then you literally get to nuke cards from existence (client-side), causing really weird desyncs if you do it in a multiplayer game.

ebbit1q: oh, I didn't know it allowed selections across zones, good catch!

ebbit1q: could we maybe add a little check at the end here that closes the view if it is empty after?

ebbit1q: I can't think of a good reason to keep around a view after all of its cards have been removed

RickyRister: sure, i'll add that in a separate PR

ebbit1q (approved):

ZeldaZach (approved):

PR #5236, opened 2024-12-10
make card view window max initial height configurable

RickyRister:
## Related Ticket(s)
- Follow up to #5228

https://github.com/user-attachments/assets/4221f8c7-74f4-48ae-9b60-a94083366990

- Created new setting for card view window maximum initial height
  - Located at `User Interface` -> `General interface settings` -> `Card view window maximum initial height` 
  - Allowed range is 0 to 9999

## Screenshots

Screenshot 2024-12-09 at 11 47 09 PM

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -455,7 +455,7 @@ void AppearanceSettingsPage::retranslateUi()
     cardScalingCheckBox.setText(tr("Scale cards on mouse over"));
     verticalCardOverlapPercentLabel.setText(
         tr("Minimum overlap percentage of cards on the stack and in vertical hand"));
-    cardViewInitialRowsMaxLabel.setText(tr("Card view window maximum initial height:"));
+    cardViewInitialRowsMaxLabel.setText(tr("Maximum initial height for card view window:"));
     cardViewInitialRowsMaxBox.setSuffix(tr(" rows"));
 
     handGroupBox->setTitle(tr("Hand layout"));
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -346,7 +346,7 @@ AppearanceSettingsPage::AppearanceSettingsPage()
     connect(&verticalCardOverlapPercentBox, SIGNAL(valueChanged(int)), &settings,
             SLOT(setStackCardOverlapPercent(int)));
 
-    cardViewInitialRowsMaxBox.setRange(0, 9999);
+    cardViewInitialRowsMaxBox.setRange(1, 9999);
     cardViewInitialRowsMaxBox.setValue(SettingsCache::instance().getCardViewInitialRowsMax());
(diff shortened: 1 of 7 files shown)


RickyRister:
Which one looks better?

Screenshot 2024-12-10 at 12 26 51 AM

Screenshot 2024-12-09 at 11 47 09 PM

ebbit1q:
I feel like this is an appearance thing, but I guess it's kinda subjective.

ebbit1q:
it's missing a unit, "pixels" might not be as easy to understand for most players, so maybe we can use card height as a unit? `cockatrice/src/game/cards/abstract_card_item.h:10` cards are 102 pixels

ebbit1q:
also, cards in views are always overlapped at 1/3rd, so a view of 204 pixels would fit 4 cards

ZeldaZach:
Agreed w/ ebbit, we should give some more clear units. Number of cards in a column seems like a reasonable amoung

RickyRister:
I feel like using `card height` as the unit ends up being kinda unintuitive due to the vertical padding and card overlap. 

For reference,
Screenshot 2024-12-10 at 10 59 35 PM 1 card height 

Screenshot 2024-12-10 at 10 59 57 PM 2 card heights

Screenshot 2024-12-10 at 11 00 26 PM 3 card heights

RickyRister:
`rows` seems like a decent unit to use

Screenshot 2024-12-13 at 2 14 31 AM 3 rows

Screenshot 2024-12-13 at 2 14 45 AM 4 rows

Screenshot 2024-12-13 at 2 15 01 AM 5 rows

Screenshot 2024-12-13 at 2 17 58 AM  10 rows

Screenshot 2024-12-13 at 2 15 43 AM 14 rows

ebbit1q:
rows is kinda what I tried to convey with "cards", rows + 2

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -346,18 +346,18 @@ AppearanceSettingsPage::AppearanceSettingsPage()
     connect(&verticalCardOverlapPercentBox, SIGNAL(valueChanged(int)), &settings,
             SLOT(setStackCardOverlapPercent(int)));
 
-    cardViewInitialHeightBox.setRange(0, 9999);
-    cardViewInitialHeightBox.setValue(SettingsCache::instance().getCardViewInitialHeight());
-    connect(&cardViewInitialHeightBox, qOverload<int>(&QSpinBox::valueChanged), &SettingsCache::instance(),
-            &SettingsCache::setCardViewInitialHeight);
+    cardViewInitialRowsMaxBox.setRange(0, 9999);
```
ebbit1q, line 349: I'd argue we'd want at least 1 here, 0 is useless

RickyRister:
> rows is kinda what I tried to convey with "cards", rows + 2

Which one do you think is more intuitive? `cards` or `rows`?
`cards` has the problem that, due to overlap, you're showing more than the given number when the window is cut off.

For example, `cards = 5` will show 7 cards names in the column.
Screenshot 2024-12-13 at 6 00 59 PM

ZeldaZach (approved):

ebbit1q:
I think rows is great!

PR #5240, opened 2024-12-12
improve FilterString validation error message

### LINKED ISSUE #5238: Misleading debug message for FilterString validation error
Client version 2.9.0 (2023-09-14)

Looking at the code, the “put top card on stack until” action is suppose to validate the filter expression, and terminate the action and throw up an error window if invalid. However, currently the validation will always succeed. 

**Steps to reproduce:**
 - Open Cockatrice with logs enabled
 - Try entering an invalid expression in the deck builder search bar
   - Confirm that cockatrice logs a debug message about invalid expression
 - Open a game and use the `put top card on stack until` action and enter an invalid expression
   - Cockatrice will happily accept that expression and then proceed to flip the entire deck onto the stack

RickyRister:
## Related Ticket(s)
- Fixes #5238

FilterString validation failure debug message does not contain the expression itself

Debug message now contains the expression itself

## Screenshots
Before: `Filter string error "1:5: syntax error"`
After: `Filter string error for "asdf:" ; "1:5: syntax error"`

### DIFF
--- a/cockatrice/src/game/filters/filter_string.cpp
+++ b/cockatrice/src/game/filters/filter_string.cpp
@@ -361,7 +361,7 @@ FilterString::FilterString(const QString &expr)
     }
 
     search.log = [&](size_t ln, size_t col, const std::string &msg) {
-        _error = QString("line %1, col %2: %3").arg(ln).arg(col).arg(QString::fromStdString(msg));
+        _error = QString("Error at position %1: %2").arg(col).arg(QString::fromStdString(msg));
     };
 
     if (!search.parse(ba.data(), result)) {
--- a/cockatrice/src/game/filters/filter_string.cpp
+++ b/cockatrice/src/game/filters/filter_string.cpp
@@ -365,7 +365,7 @@ FilterString::FilterString(const QString &expr)
     };
 
     if (!search.parse(ba.data(), result)) {
-        qDebug() << "Filter string error" << _error;
+        qDebug() << "Filter string error for" << expr << ";" << _error;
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -361,11 +361,11 @@ FilterString::FilterString(const QString &expr)
     }
 
     search.log = [&](size_t ln, size_t col, const std::string &msg) {
-        _error = QString("%1:%2: %3").arg(ln).arg(col).arg(QString::fromStdString(msg));
+        _error = QString("line %1, col %2: %3").arg(ln).arg(col).arg(QString::fromStdString(msg));
```
ebbit1q, line 364: if we're going for user friendlyness, line will often be 0, so hide it from the user in that case, col for column is not clear to most users, consider variable width fonts which don't align by column are used by default in this box, instead we could try "character" or "key" or "symbol" or "glyph"

ebbit1q: just "position"?

ZeldaZach (approved):

cockatrice/src/game/filters/filter_string.cpp
ebbit1q: hey, what I meant with "in that case" is to make it not be there in case it is 0, I don't know in what cases it might not be 0 but it could potentially be very confusing

RickyRister: oops... Well, free feel to open up a PR to fix it

PR #5242, opened 2024-12-13
Support Windows Debug Builds

ZeldaZach:


ebbit1q:
note that this really seems to change the way fetchMore works, I don't think it worked at all since dynamic card loading was implemented so it's not actually noticeable.

PR #5244, opened 2024-12-14
Update PegLib, Fix Database Searching CFG

### LINKED ISSUE #3730: searching r:common returns uncommon cards
<b>System Information:</b>
Build Architecture: 64-bit
Qt Version: 5.12.1
System Locale: en_US
Install Mode: Standard
  
Searching for `r:common` in the deck editor returns cards with uncommon rarity.
A workaround is to search `r:common -r:uncommon`, but this is not immediately clear to new users, and isn't very intuitive.

### LINKED ISSUE #5200: Deck editor's card search function does no longer work with text after quotation marks
Client Version: 2.10.0-custom(ac85b36) (2024-11-21)
Client Operating System: Ubuntu 20.04.6 LTS
Build Architecture: 64-bit
Qt Version: 5.12.8
System Locale: en_US
Install Mode: Standard

**Description:**

Expected Behaviour:
- When using the search function of the deck editor in the Cockatrice client and writing expressions using quotation marks, the correct results show up. This includes searches as e.g.
  `o:"enchantment" o:"artifact"`, `o:"enchantment" or o:"artifact"`, `o:"equip" -o:"arti"`, `f:c -ci:rgb (o:"fly" or o:"skulk") -o:"defender"`
  and so on.
- Results on searching simple terms like `o:"enchantment" o:"artifact"` are the same as the results on searching `o:enchantment o:artifact`

Issues:
- When using quotation marks in a card search of the deck editor, instead of displaying fitting results nothing is shown as if the query was wrong or as if there were no fitting results.
- Searching without quotation marks for the same expressions where possible give correct results, whilst using quotation marks in any capacity does not give any results at all.
- This behaviour was not present in client versions 2.8.1-beta or prior.
CardEditorSearchQuotationsBug

**Steps to reproduce:**
- Have client version 2.9.1-beta or up. (Potentially also 2.9.0 or up.)
- Use the search function of the deck editor with any search expressions involving quotation marks and any other text or epxression after a space after a quotation mark (see examples above).

ZeldaZach:
- Update Peglib to C++20 version
- Fix #5200 
- Fix #3730 
- Move to the C++20 Standard (Which has been approved/published for over 4 years now)


ebbit1q:
any specific reason we need 20 for?

ZeldaZach:
We don't _need_ it, per say, but it could be a nice to have to meet higher versions of standards

ebbit1q:
putting 20 in our cmakelists does sorta make it required though, it'd mean that people might not be able to compile using 17 anymore

ebbit1q (approved):

PR #5246, opened 2024-12-14
Add action to select all cards in zone

### LINKED ISSUE #5245: Add action to select all cards in zone

We should add an action to the card menu that would select all cards in the same zone as that card. This would be a huge quality-of-life improvement for situations such as exiling a fully-stocked graveyard.

The shortcut can be `Ctrl+A`, to match what people are used to. `Ctrl+A` is currently the default shortcut for `Draw arrow`, but I'm pretty sure nobody actually uses that shortcut for drawing arrows.

RickyRister:
## Related Ticket(s)
- Fixes #5245

https://github.com/user-attachments/assets/bd6d4655-443f-4d5f-9e1c-8ce353a57ce0

Added a `Select All Cards in Zone` action to the card menu. This action will add all cards in that card's current zone to the current selection.

This action will appear in the card menu for all zones, since it is always possible for selecting all cards in a zone to be relevant for a zone. I put it below the `Move` actions in the right-click menu.

The default shortcut for this action is `Ctrl+A`, to match existing intuition. That shortcut belonged to `Draw arrow`, but I'm not sure if anyone even uses that shortcut for drawing arrows, so I just got rid of the default shortcut for `Draw arrow`.

This action uses `game->getActiveCard()` to decide which card to use for the zone. When multiple cards are already selected, the keyboard shortcut should be able to key off the latest card selected.

## Screenshots
Screenshot 2024-12-13 at 8 01 53 PM


RickyRister:
So one problem with adding this action to every card menu is that it now makes the menus that used to be narrow become a lot wider. 

Screenshot 2024-12-13 at 7 55 14 PM


Should I reword the text to make it less wide?

Screenshot 2024-12-13 at 8 25 51 PM

Screenshot 2024-12-13 at 7 56 23 PM

cockatrice/src/settings/shortcuts_settings.h
```diff
@@ -536,7 +536,7 @@ class ShortcutsSettings : public QObject
                                                 parseSequenceString(""),
                                                 ShortcutGroup::Move_bottom)},
         {"Player/aDrawArrow", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Draw Arrow..."),
-                                          parseSequenceString("Ctrl+A"),
+                                          parseSequenceString(""),
```
ZeldaZach, line 539: @ebbit1q Is this a concern to you?

ZeldaZach (approved):

cockatrice/src/settings/shortcuts_settings.h
ebbit1q: yes, while most people might draw arrows with the right mouse drag we want to offer a shortcut as well, some people might prefer it and some people might have disabilities that make it easier to use a single press instead of a drag with the mouse.

ebbit1q:
I suggest the new shortcut use ctrl shift a instead of overwriting the existing shortcut

RickyRister:
`Ctrl+A` is such a fitting shortcut though. Could we just rebind the default `draw arrow` shortcut instead?

cockatrice/src/game/player/player.cpp
```diff
@@ -836,6 +839,8 @@ void Player::retranslateUi()
         sayMenu->setTitle(tr("S&ay"));
     }
 
+    aSelectAll->setText(tr("&Select All in Zone"));
```
ZeldaZach, line 842: I don't think you intended to put the &S here, as that's an auto Windows hot-key.

cockatrice/src/game/player/player.cpp
```diff
@@ -1536,6 +1543,22 @@ void Player::actMoveBottomCardToTop()
...
 
+void Player::actSelectAll()
+{
+    const CardItem *card = game->getActiveCard();
+    if (!card) {
+        return;
+    }
+
+    if (const auto *zone = card->getZone()) {
+        for (auto &cardItem : zone->getCards()) {
```
ZeldaZach, line 1554: nit: Technically you should check if zone is a nullptr

cockatrice/src/settings/shortcuts_settings.h
```diff
@@ -536,7 +539,7 @@ class ShortcutsSettings : public QObject
                                                 parseSequenceString(""),
                                                 ShortcutGroup::Move_bottom)},
         {"Player/aDrawArrow", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Draw Arrow..."),
-                                          parseSequenceString("Ctrl+A"),
+                                          parseSequenceString("Alt+A"),
```
ZeldaZach, line 542: I'm okay with this change

cockatrice/src/game/player/player.cpp
RickyRister: doesn't `if (const auto *zone = card->getZone())` already do that?

ZeldaZach: I've never seen that annotation before tbf, it might?

cockatrice/src/game/player/player.cpp
```diff
@@ -839,7 +839,7 @@ void Player::retranslateUi()
         sayMenu->setTitle(tr("S&ay"));
     }
 
-    aSelectAll->setText(tr("&Select All in Zone"));
```
ebbit1q: here is some background on what the & does: when opening a menu it automatically adds a shortcut with alt + the following letter, eg if a card menu were opened one would be able to use alt s to select this action (or open a sub menu), if two ampersands precede the same letter then instead of immediately selecting that one it'll cycle through them, requiring another button press to select the one you want, this is why we sometimes have it in front of a random letter, on mac os this feature is disabled (similar to the showing of shortcuts, god forbid mac users figure out how to use the keyboard), we of course don't actually know how many people are aware or use this feature at all but if I remember correctly it makes the use of the menus more accessible to users that have difficulty with pointing devices.

RickyRister: Should I just put it in front of a random letter then? I'm not sure what happens if two actions have overlapping auto-hokeys or if they conflict with existing keyboard shortcuts

ebbit1q: either works, it doesn't hurt to have it overlap, it's just an extra keypress, imo some of the random keys we use make so little sense that they take more time to use, reading the text looking for the highlighted letter, than an extra keypress

ebbit1q: afaik the menu takes focus when it is open, eating all shortcuts

cockatrice/src/settings/shortcuts_settings.h
ebbit1q: people can learn a new shortcut but I think if someone has set this shortcut and we overwrite it some unexpected things might happen, this'd need some testing

ebbit1q: looks like this testing was not done, this conflicts with the accelerator key for opening the card database menu on windows and linux...

PR #5248, opened 2024-12-15
Add keyboard shortcut for "hide" action

### LINKED ISSUE #5247: Enable keyboard shortcut for "hide" action

The `hide` action doesn't have a keyboard shortcut. We should allow it to have a keyboard shortcut.

`Ctrl+H` conflicts with the default system shortcut for "hide window", so perhaps we can make the default shortcut `Alt+H`.

RickyRister:
## Related Ticket(s)
- Fixes #5247 

- Set a keyboard shortcut for the `hide` action
- Shortcut is located under `Playing Area` in shortcut settings
  - Default shortcut is `Alt+H`

## Screenshots

Screenshot 2024-12-14 at 10 16 23 PM

Screenshot 2024-12-14 at 10 16 35 PM


cockatrice/src/settings/shortcuts_settings.h
```diff
@@ -411,6 +411,9 @@ class ShortcutsSettings : public QObject
         {"Player/aNextTurn", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Next Turn"),
                                          parseSequenceString("Ctrl+Return;Ctrl+Enter"),
                                          ShortcutGroup::Game_Phases)},
+        {"Player/aHide", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Hide Card (in Reveal Window)"),
```
ebbit1q, line 414: would this label also work without the parens?

RickyRister: It would. I added the parens because it's not immediately obvious that `hide` is referring to hiding a revealed card.

ebbit1q: I mean removing the parens, keeping their contents, it's easier to understand to translators

RickyRister: I can change it to `Hide Card that was Revealed`

ebbit1q: both of those sound great!

ebbit1q: the last one might be interpreted as the other reveal option, reveal top card of deck

RickyRister: actually yeah in that case `Hide Card in Reveal Window` is probably the clearest

ebbit1q (approved):
thank you!

PR #5250, opened 2024-12-16
ability to directly attach from other zones

### LINKED ISSUE #2068: ability to `attach to` from stack

RickyRister:
## Related Ticket(s)
- Fixes #2068

https://github.com/user-attachments/assets/0e916b76-f898-49b7-86aa-680cec969d22

The `attach to` action now functions in zones other than the table. When attaching from a zone other than the table, the card will be first played to the table, then attached.

More specifically, the `attach to` action now functions in `hand`, `graveyard`, `exile`, and `stack`. It still does not currently function in `deck` or `sideboard` because those zones are really wonky when you try to play multiple cards at the same time from them.

The card will be played to the table with all the properties (tablerow, etb tapped, p/t etc.) that it would have if played to the table normally. If a card with tablerow 3 (the stack) is played this way, it will be played to tablerow 2.

Also added the `draw arrow` action to all card menus that I added `attach to` to, since those two actions always seem to appear together.

### Technical Changes
- Added `Player::playCardToTable` method.
  - Copy-pasted from `Player::playCard` except I removed the logic for deciding which zone the card goes to.
- refactored card attachment logic in `ArrowAttachItem` to new method so that I can early return if the target card is already attached
  - this is to prevent cards from being played to table and then not attached if you try to attach to an already attached card.
- added `aAttach` and `aDrawArrow` actions to a bunch more card menus
  - Not super sold on the placement of the new actions into the dropdown menu. Let me know what you think
  - Honestly, we might need to go back and reorganize the options in the dropdown menu one day

## Screenshots

Screenshot 2024-12-15 at 6 34 15 PM

Screenshot 2024-12-15 at 6 34 06 PM

Screenshot 2024-12-15 at 6 34 01 PM

### DIFF
--- a/cockatrice/src/game/board/arrow_item.cpp
+++ b/cockatrice/src/game/board/arrow_item.cpp
@@ -341,7 +341,9 @@ void ArrowAttachItem::mouseReleaseEvent(QGraphicsSceneMouseEvent *event)
     if (targetItem && (targetItem != startItem)) {
         auto startCard = qgraphicsitem_cast<CardItem *>(startItem);
         auto targetCard = qgraphicsitem_cast<CardItem *>(targetItem);
-        attachCards(startCard, targetCard);
+        if (startCard && targetCard) {
+            attachCards(startCard, targetCard);
+        }
     }
 
     delArrow();
--- a/cockatrice/src/game/board/arrow_item.cpp
+++ b/cockatrice/src/game/board/arrow_item.cpp
@@ -318,8 +318,14 @@ void ArrowAttachItem::mouseReleaseEvent(QGraphicsSceneMouseEvent *event)
         CardItem *targetCard = qgraphicsitem_cast<CardItem *>(targetItem);
         CardZone *targetZone = targetCard->getZone();
 
+        // move card onto table first if attaching from some other zone
+        if (startZone->getName() != "table") {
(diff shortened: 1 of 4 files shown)


cockatrice/src/game/board/arrow_item.cpp
```diff
@@ -307,25 +307,41 @@ void ArrowAttachItem::mouseMoveEvent(QGraphicsSceneMouseEvent *event)
...
 
+void ArrowAttachItem::attachCards(CardItem *startCard, const CardItem *targetCard)
+{
+    // do nothing if target is already attached to another card
+    if (targetCard->getAttachedTo()) {
+        return;
+    }
+
+    CardZone *startZone = startCard->getZone();
+    CardZone *targetZone = targetCard->getZone();
```
ZeldaZach, line 318: should check these two values are non-null

ZeldaZach (approved):

PR #5254, opened 2024-12-17
clicking to play can now play all selected

RickyRister:
## Related Ticket(s)
- Follow-up to #5252 (This PR requires #5252 to be merged first)

Double-clicking to play cards doesn't cause all selected cards to get played, because they use a different method from `Player` to play the card.

Added new setting under `User Interface` -> `General interface settings`. When that setting is enabled, clicking to play will also apply the action to all selected cards.

Clicking will apply the existing "click to play" behavior to all selected cards: Clicking while holding the shift key will play them face down. Clicking on cards in a card reveal menu will hide them.

- When the setting is enabled, `CardItem::playCard` now gets `this->getZone()->getPlayer()` and calls `player->act(Play|PlayFaceDown|Hide)` on it, which will act on all selected cards.
  - When the setting is off, the behavior is as before
- Also refactored `CardItem` while I was at it

## Screenshots

Screenshot 2024-12-16 at 8 31 39 PM

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -492,6 +492,10 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     connect(&doubleClickToPlayCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance(),
             &SettingsCache::setDoubleClickToPlay);
 
+    clickPlaysAllSelectedCheckBox.setChecked(SettingsCache::instance().getClickPlaysAllSelected());
+    connect(&clickPlaysAllSelectedCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance(),
+            &SettingsCache::setClickPlaysAllSelected);
+
     playToStackCheckBox.setChecked(SettingsCache::instance().getPlayToStack());
     connect(&playToStackCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance(),
             &SettingsCache::setPlayToStack);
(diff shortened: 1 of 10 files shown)


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -582,6 +587,7 @@ void UserInterfaceSettingsPage::retranslateUi()
 {
     generalGroupBox->setTitle(tr("General interface settings"));
     doubleClickToPlayCheckBox.setText(tr("&Double-click cards to play them (instead of single-click)"));
+    clickPlaysAllSelectedCheckBox.setText(tr("&Clicking plays all selected cards (instead of just the clicked card)"));
```
ZeldaZach, line 590: I don't like the wording, but I'm not sure what would be better. @ebbit1q any ideas?

cockatrice/src/game/cards/card_item.cpp
```diff
@@ -383,8 +383,45 @@ void CardItem::playCard(bool faceDown)
...
+
+/**
+ * @brief returns true if the zone is a unwritable reveal zone view (eg a card reveal window). Will return false if zone
+ * is nullptr.
+ */
+static bool isUnwritableRevealZone(CardZone *zone)
+{
+    if (zone && zone->getIsView()) {
+        if (auto *view = static_cast<ZoneViewZone *>(zone))
+            return view->getRevealZone() && !view->getWriteableRevealZone();
```
ZeldaZach, line 403: add bracket

ZeldaZach (approved):

ebbit1q:
in my testing alt a does actually conflict with an existing shortcut, the menu bar automatically registers alt + C, A, G, V, D to open their menus

PR #5257, opened 2024-12-17
Move Card Zones to an enum-like approach

ZeldaZach:
I tried enums, was more annoying. Will probably look again later.


webclient/src/i18n-default.json
tooomm, line 1: Please do not change unrelated files and keep pretty printed formatting for JSONs.

No way to tell about changes otherwise.

cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -673,12 +673,12 @@ void TabDeckEditor::retranslateUi()
     deckMenu->setTitle(tr("&Deck Editor"));
 
     cardInfoDock->setWindowTitle(tr("Card Info"));
-    deckDock->setWindowTitle(tr("Deck"));
+    deckDock->setWindowTitle(tr(ZONE_DECK));
```
tooomm, line 676: That would change the capitalisation of the displayed string.

webclient/src/i18n-default.json
ZeldaZach: Windows keeps auto-linting it. I planned on fixing that before submitting.

cockatrice/src/client/tabs/tab_deck_editor.cpp
ZeldaZach: Yeah good catch, this doesn't need to be replaced

PR #5260, opened 2024-12-17
double clicking the card view window now expands/unexpands it

RickyRister:
## Related Ticket(s)
- Follow-up to #5228

Expanding the card view window is really finicky to do. The collision in the corner to start a drag is *tiny*, and the window sometimes glitches out all over the place as you try to expand it.

https://github.com/user-attachments/assets/2241fade-28af-4efb-8dea-efc8f295d5bc

Double clicking the top of the card view window will now expand/unexpand it. The collision area for the double-click is the same as the collision for dragging the window around.

Double clicking the window while it's not at its max width and height will cause the window to be resized to its max width and height. Double clicking while it is will cause the window to reset back to the size that it would be at when it was resized due to a card view change.

- add override qualifiers to overriden methods in `ViewZoneWidget` because it wasn't compiling for me otherwise

### DIFF
--- a/cockatrice/src/game/zones/view_zone_widget.cpp
+++ b/cockatrice/src/game/zones/view_zone_widget.cpp
@@ -333,3 +333,16 @@ void ZoneViewWidget::initStyleOption(QStyleOption *option) const
     if (titleBar)
         titleBar->icon = QPixmap("theme:cockatrice");
 }
+
+void ZoneViewWidget::mouseDoubleClickEvent(QGraphicsSceneMouseEvent *event)
+{
+    if (event->pos().y() <= 0) {
+        // expand window to full height and width if not currently at full height and width
+        // otherwise shrink window to initial max height
+        if (size().height() < maximumHeight() || size().width() < maximumWidth()) {
+            resize(maximumSize());
+        } else {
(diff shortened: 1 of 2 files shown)


ZeldaZach (approved):

ebbit1q:
image
it seems like some weird stuff can happen when the window extends beyond the play area, also I think it still needs some sort of limit, extending to the full length of 60 rows will not fit on the screen.

RickyRister:
We couldn't figure out how to limit the height of the window to the height of the screen. The problem is that the window's position uses scene coordinates (which is scaled to the game board and affected by the game board zooming out) while the size of the window uses pixel coordinates. We couldn't figure out a way to map those coordinates to each other.

The position of the window being based on pixel coordinates leads to some other problems, such as the window being forced down by the black area on top the when the board is zoomed out. Perhaps we should look into fixing that eventually, and get the window position to be bounded by the outer frame instead of the game board.

Personally, I don't think the uncapped height is a dealbreaker. The user can just double click again to unexpand it. However, if it is indeed a must-fix, then I'll probably put this PR on hold until we get the card view window's position to know about the outer frame.

Issue #5256, opened 2024-12-17
Add a Button to Flip/Rotate Cards like in Archidekt


issa-puppi:
Would be nice if similar to "view related cards" we had a button to rotate battles and rooms sideways for easier reading

Would be nice if it would be both in card info and in deck editor.

BruebachL:
Duplicate of [#479](https://github.com/Cockatrice/Cockatrice/issues/479) but will be added very soon with PR [#5264](https://github.com/Cockatrice/Cockatrice/pull/5264).

PR #5264, opened 2024-12-18
Rotate split cards

### LINKED ISSUE #479: Planes: Display horizontally
Some cards, namely planes, should be displayed horizontally, instead of vertically.
Not just in the command zone (or currently, the battlefield), but in the popup window/enlarged card in the upper right as well. They currently are condensed into vertical cards, making the art hard to read.

I can imagine this feature could also be useful for some other card games.

If possible, they could be enlarged in size as well to reflect their real life counterparts (which is twice the size of regular cards. This applies to schemes, vanguards, planes, phenomena.

BruebachL:
## Related Ticket(s)
- Fixes #479, #3094, and #5256

Split cards, battles and planes are displayed upright which hinders legibility. 

- The pixmap of these cards is rotated in the CardInfoPictureWidget if their layout or main type matches these cards.

## Screenshots
image
image


RickyRister:
I know this is outside the scope of the change, but i wonder if it would be better to instead introduce a `isRotated` property to the card info, instead of cueing off a specific set of conditions.

That would make it better for the custom card people, since they make new rotated cards all the time, often with new types.

For backwards compatibility purposes, perhaps if the card info doesn't have a `isRotated` field, then we can fall back to the condition check.

BruebachL:
> I know this is outside the scope of the change, but I wonder if it would be better to introduce a `isRotated` property to the card info, instead of cueing off a specific set of conditions.
> 
> That would make it better for the custom card people, since they make new rotated cards all the time, often with new never-seen-before types.
> 
> For backwards compatibility purposes, perhaps if the card info doesn't have a `isRotated` field, then we can fall back to the condition check.

I would be willing to concede an additional layout check to the custom card people (e.g. if layout == "landscape" || "rotated") but I personally don't see the need for the isRotated property.

If you are willing to make that happen, I encourage you to do it, but this is about as much time and effort as I am willing to invest in this before our latest release.

RickyRister:
Do note that by keying off certain properties, that means this thing will break as soon as wizards releases a new rotated card that doesn't match one of the above. 

I remember some similar breakage happening when wizards changed `enters the battlefield tapped` to `enters tapped` and the parsing engine didn't recognize those cards as etb tapped anymore. Luckily that breakage was on the cards.xml generator side, so they were able to fix it without changing the cockatrice client.

By hardcoding this logic inside of the cockatrice client itself, we would need to release an update of the client to do any fixes. And cockatrice releases really infrequently. 

I'm just worried that by putting the logic in the cockatrice client instead of the thing generating the xml, it becomes a lot harder to patch if wizards releases something in the future that breaks it.

BruebachL:
> Do note that by keying off certain properties, that means this thing will break as soon as wizards releases a new rotated card that doesn't match one of the above.

It will "break" for that new card type, which is to be expected if this an entirely new, never before seen card type.

> I remember some similar breakage happening when wizards changed `enters the battlefield tapped` to `enters tapped` and the parsing engine didn't recognize those cards as etb tapped anymore. Luckily that breakage was on the cards.xml generator side, so they were able to fix it without changing the cockatrice client.

Does cockatrice even take such things into consideration? Could you link me to that?
 
> By hardcoding this logic inside of the cockatrice client itself, we would need to release an update of the client to do any fixes. And cockatrice releases really infrequently.
> 
> I'm just worried that by putting the logic in the cockatrice client instead of the thing generating the xml, it becomes a lot harder to patch if wizards releases something in the future that breaks it.

The thing generating the XML is essentially cockatrice as well. I don't think we can update the oracle importer without updating cockatrice as well (someone please correct me if this is wrong).

---

My biggest point is that the field you are looking for `isRotated` is essentially already there as the `layout` field, just with much more detail. I see the `isRotated` field as redundant, in that it doesn't offer us any additional information that we don't already have and it merely shifts the responsibility of decoding this logic from UI to backend.

If we are truly concerned about other code using this property in a similar way, I'd rather encapsulate the UI logic into a function of the cards itself (Check your own layout and then give me a decision as the caller) and then use that function in the UI code.

BruebachL:
I found the relevant code [here](https://github.com/Cockatrice/Cockatrice/blob/27e5d21b6b39b385063f18795d58408632d4f048/oracle/src/oracleimporter.cpp#L153) and [here](https://github.com/Cockatrice/Cockatrice/blob/27e5d21b6b39b385063f18795d58408632d4f048/cockatrice/src/game/player/player.cpp#L3574)

RickyRister:
>> I remember some similar breakage happening when wizards changed enters the battlefield tapped to enters tapped and the parsing engine didn't recognize those cards as etb tapped anymore. Luckily that breakage was on the cards.xml generator side, so they were able to fix it without changing the cockatrice client.
>
> Does cockatrice even take such things into consideration? Could you link me to that?

https://github.com/Cockatrice/Cockatrice/issues/5093
I may have been mistaken about being able to update oracle without doing a new release of cockatrice though.

ebbit1q:
updates of oracle are only available through an update of oracle, however oracle can download an xml that is "fixed" though, I'd really like to host xmls for people instead of having all people download a large json file to throw out most of its properties and waste a lot of cpu cycles doing the same actions.

BruebachL:
My original point still stands, I see no reason to introduce an additional field when the same can be achieved through a runtime evaluation of an already existing field.

This evaluation should be consolidated in the method of a card item, if anywhere.

ZeldaZach:
This is definitely a tough one. I can understand arguments on both sides of the issue. However, from my experience with things, I tend to be a bit more conservative with how things are handled in the UI and like the UI to be as "dumb" as possible. In this case, moving the change to Oracle seems like a better approach, as if (when) we move Oracle to a more download-and-play model, we can include these special "rotate" fields, or w/e other fields we'd like on the fly.

tooomm:
I'm confused now, is this included or not? 🤔 
What is the final outcome on this as it seems to not work for me in 2.10.0-beta.24:
image

The feature requests are closed still by this PR.

BruebachL:
> I'm confused now, is this included or not? 🤔 What is the final outcome on this as it seems to not work for me in 2.10.0-beta.24: image
> 
> The feature requests are closed still by this PR.

Have you updated your card database? Iirc cards gained a new attribute which might require a reimport. It works for me on latest master.

image

tooomm:
Yes, I did.

I'm seeing `<landscapeOrientation>1</landscapeOrientation>` in cards.xml for the card:

```
        <card>
            <name>Depose // Deploy</name>
            <text>Tap target creature.
Draw a card.

---

Create two 1/1 colorless Thopter artifact creature tokens with flying, then you gain 1 life for each creature you control.</text>
            <prop>
                <format-oathbreaker>legal</format-oathbreaker>
                <side>front</side>
                <format-gladiator>legal</format-gladiator>
                <cmc>6</cmc>
                <format-commander>legal</format-commander>
                <format-historic>legal</format-historic>
                <colors>WU</colors>
                <format-modern>legal</format-modern>
                <format-pioneer>legal</format-pioneer>
                <format-vintage>legal</format-vintage>
                <layout>split</layout>
                <format-timeless>legal</format-timeless>
                <format-legacy>legal</format-legacy>
                <coloridentity>WU</coloridentity>
                <maintype>Instant</maintype>
                <format-brawl>legal</format-brawl>
                <format-explorer>legal</format-explorer>
                <type>Instant</type>
                <format-duel>legal</format-duel>
                <manacost>1{W/U} // 2WU</manacost>
            </prop>
            <set uuid="0bda0b80-8a41-4001-9f60-06b1d7f9b1fa" num="185" muid="651920" rarity="uncommon">CLU</set>
            <set uuid="ab0ba4ef-9e82-4177-a80f-8fa6f6a5bd60" num="225" muid="457369" rarity="uncommon">RNA</set>
            <tablerow>3</tablerow>
            <landscapeOrientation>1</landscapeOrientation>
        </card>
```

The new setting is enabled under `Appearance` too:
image

BruebachL:
> Yes, I did.
> 
> I'm seeing `<landscapeOrientation>1</landscapeOrientation>` in cards.xml for the card:
> 
> ```
>         <card>
>             <name>Depose // Deploy</name>
>             <text>Tap target creature.
> Draw a card.
> 
> ---
> 
> Create two 1/1 colorless Thopter artifact creature tokens with flying, then you gain 1 life for each creature you control.</text>
>             <prop>
>                 <format-oathbreaker>legal</format-oathbreaker>
>                 <side>front</side>
>                 <format-gladiator>legal</format-gladiator>
>                 <cmc>6</cmc>
>                 <format-commander>legal</format-commander>
>                 <format-historic>legal</format-historic>
>                 <colors>WU</colors>
>                 <format-modern>legal</format-modern>
>                 <format-pioneer>legal</format-pioneer>
>                 <format-vintage>legal</format-vintage>
>                 <layout>split</layout>
>                 <format-timeless>legal</format-timeless>
>                 <format-legacy>legal</format-legacy>
>                 <coloridentity>WU</coloridentity>
>                 <maintype>Instant</maintype>
>                 <format-brawl>legal</format-brawl>
>                 <format-explorer>legal</format-explorer>
>                 <type>Instant</type>
>                 <format-duel>legal</format-duel>
>                 <manacost>1{W/U} // 2WU</manacost>
>             </prop>
>             <set uuid="0bda0b80-8a41-4001-9f60-06b1d7f9b1fa" num="185" muid="651920" rarity="uncommon">CLU</set>
>             <set uuid="ab0ba4ef-9e82-4177-a80f-8fa6f6a5bd60" num="225" muid="457369" rarity="uncommon">RNA</set>
>             <tablerow>3</tablerow>
>             <landscapeOrientation>1</landscapeOrientation>
>         </card>
> ```
> 
> The new setting is enabled under `Appearance` too: image

I can't really reproduce this. It seems to work for beta testers, as far as I know. The change is very small, you might try debugging it yourself to see where the issue lies.

tooomm:
Hi @BruebachL, I have an interesting observation...
When running into https://github.com/Cockatrice/Cockatrice/issues/5500, I used the new "reload DB" button to manually get my set changes to apply. Since doing so the cards display rotated for me as well!
image

Seems like some weird interaction. Since I updated the client and the card db in between, and restarting many times - but apparently that was not enough and a specific refresh was missing.

PR #5285, opened 2024-12-21
implement search bar in shortcuts menu

### LINKED ISSUE #3648: Include search/filter to the custom shortcuts settings table
After #3641, custom shortcuts changed to be a nice table.

It would be very handy if the table would be searchable just like in the manage sets dialog or cards in the deck editor...

RickyRister:
## Related Ticket(s)
- Fixes #3648

https://github.com/user-attachments/assets/e39e3451-389c-4e2e-b68d-9e10b40fb1ad

Shortcut settings menu now has a search bar.

Currently the search bar is a very primitive implementation. It searches for that exact string, including spaces. It doesn't do the thing where it splits on spaces and shows the row if it matches any word in the string.
Perhaps we can improve the functionality of the search bar later.

- Moved the Shortcuts table from a `QTreeWidget` into its own `QTreeView` subclass in a separate file
  - Also moved the view and model logic into the subclass.

## Screenshots
Screenshot 2024-12-21 at 2 53 12 PM

Screenshot 2024-12-21 at 3 03 02 PM

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -39,8 +39,6 @@
 #include <QStackedWidget>
 #include <QToolBar>
 #include <QTranslator>
-#include <QTreeWidget>
-#include <QTreeWidgetItem>
 
 #define WIKI_CUSTOM_PIC_URL "https://github.com/Cockatrice/Cockatrice/wiki/Custom-Picture-Download-URLs"
 #define WIKI_CUSTOM_SHORTCUTS "https://github.com/Cockatrice/Cockatrice/wiki/Custom-Keyboard-Shortcuts"
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -6,10 +6,12 @@
 #include "../client/sound_engine.h"
 #include "../client/ui/picture_loader.h"
 #include "../client/ui/theme_manager.h"
+#include "../deck/custom_line_edit.h"
 #include "../game/cards/card_database.h"
 #include "../game/cards/card_database_manager.h"
 #include "../main.h"
 #include "../settings/cache_settings.h"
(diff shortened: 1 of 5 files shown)


ZeldaZach (approved):

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -1251,13 +1251,24 @@ void SoundSettingsPage::retranslateUi()
 
 ShortcutSettingsPage::ShortcutSettingsPage()
 {
+    // search bar
+    searchEdit = new SearchLineEdit;
+    searchEdit->setObjectName("searchEdit");
+    searchEdit->setPlaceholderText(tr("Search by shortcut name"));
```
ebbit1q, line 1257: translateable text should go in retranslateUI

cockatrice/src/settings/shortcut_treeview.h
```diff
@@ -0,0 +1,34 @@
...
+
+private:
+    QStandardItemModel *shortcutsModel;
+    QSortFilterProxyModel *proxyModel;
+    void populateShortcutsModel();
+
+private slots:
+    void refreshShortcuts();
+
+protected:
```
ebbit1q, line 30: what does setting this as protected actually do for this class that isn't subclassed again and isn't part of a library we provide (we don't provide any)?

RickyRister: We’re overriding a protected method in the parent class

cockatrice/src/dialogs/dlg_settings.cpp
RickyRister: Oops… feel free to open a PR to fix it

ZeldaZach: Fixing

cockatrice/src/settings/shortcut_treeview.h
ebbit1q: ah, I see

PR #5286, opened 2024-12-22
Fix token name highlight on open

ZeldaZach:


cockatrice/src/dialogs/dlg_create_token.cpp
```diff
@@ -31,7 +31,11 @@ DlgCreateToken::DlgCreateToken(const QStringList &_predefinedTokens, QWidget *pa
     nameLabel = new QLabel(tr("&Name:"));
     nameEdit = new QLineEdit(tr("Token"));
     nameEdit->setMaxLength(MAX_NAME_LENGTH);
-    nameEdit->selectAll();
+    QTimer::singleShot(100, this, [=, this]() {
```
ebbit1q, line 34: this was never broken for me, however because of the printing selector, opening the tokens menu has become even slower

PR #5288, opened 2024-12-22
Fix multiple "Selected Cards" in Menu on MacOS

ZeldaZach:
Did a lot of testing to make sure shortcuts don't break and only one selected cards ever appears, across both Windows and MacOS


RickyRister:
Fixes this
Screenshot_2024-12-21_at_4 13 10_AM

PR #4963, opened 2023-12-15
Improve drag & drop behavior

Elarnon:
## Related Ticket(s)
- Fixes #4817 (partially)

## Short roundup of the initial problem

Drag & drop placement on the grid is unintuitive.

## What will change with this Pull Request?

This patch tweaks the drag & drop behavior (in particular, the grid placement) to be more intuitive. More precisely, with this patch the drag & drop will:

 - Only use the "hot spot" (i.e. position of the cursor on the card) for zones where the card is actually displayed around the cursor (in particular, not on the table where the card snaps to the grid).

 - Use better boundaries computed with respect to the center of the card (rather than its top left corner) for determining which grid cell a card should go to

 - Align behavior of the preview and the actual effect when overflow of the 3-card stacks occurs

 - Avoid visual glitches where the cursor ends up outside of the card or at incorrect offsets when moving the mouse too fast (which translates to overflows of the hot spot computation)

Note that this patch does not include fixes for the group moves covered [here](https://github.com/Cockatrice/Cockatrice/issues/4817#issuecomment-1854808313).

### Moving a single card

On current `master` 

https://github.com/Cockatrice/Cockatrice/assets/146210/4f0856be-3aae-4a3c-b43b-5640cc55dd1e

With this PR

https://github.com/Cockatrice/Cockatrice/assets/146210/bc59463b-27a5-4fbe-b9a6-bd8d0cd213fa

### Moving to an occupied position

On current `master`

https://github.com/Cockatrice/Cockatrice/assets/146210/f5634e72-361e-4cef-8135-ad18cf68c953

With this PR

https://github.com/Cockatrice/Cockatrice/assets/146210/8128cfd7-538c-4dd9-b631-724a00f9e189

### Moving to a position with an attachment

On current `master`

https://github.com/Cockatrice/Cockatrice/assets/146210/a4173483-11e6-4f0f-a3fd-ed758054d60b

With this PR

https://github.com/Cockatrice/Cockatrice/assets/146210/4887cb53-6e68-450f-abb4-f7d34899b955

PR #5289, opened 2024-12-22
Fix "ghosting" of cards sticking on invalid moves

ZeldaZach:
## Related Ticket(s)
- Follow up to #4963

When trying to move a stack of cards, if the starting position was occupied then the ghost cards would not disappear.

Scene gets cleaned up of ghosts on each iteration, regardless if the move is successful or not

## Screenshots
image


ebbit1q:
I've never seen this, can you give me steps to reproduce?

RickyRister:
> I've never seen this, can you give me steps to reproduce?

https://github.com/user-attachments/assets/2692709c-3a4b-412e-b731-3f58f30f71be

RickyRister:
seems like the fix worked 

https://github.com/user-attachments/assets/66276e36-1033-4731-b7ea-a817203d9229

PR #5319, opened 2024-12-24
add "open recent" menu option to deck editor tab

### LINKED ISSUE #5291: "Open Recent" menu option in deck editor
**Similar Requests**

#693

Cockatrice should provide a `Open Recent` menu option in the deck editor, similar to what other applications do.

**Additional Context**

Cockatrice should store the list of most recently opened decks, and provide a menu action under `Deck Editor` that lists the recent decks and gives the user a chance to open them.

Keeping with what other applications do, there should also be a `Clear Menu` option at the bottom to clear the history.

RickyRister:
## Related Ticket(s)
- Fixes #5291

https://github.com/user-attachments/assets/35876fe4-0291-4a12-bfb9-53226a963478

Added a `Open Recent` section to the deck editor tab's menu. It will list recently opened decks, and allow you to open those decks. Opening a recent deck will move it to the top of the list. If the list reaches the max number of entries, it will start evicting the oldest entry.

Max number of entries is currently hardcoded at 10.

Currently, this only tracks decks that are opened with the `Open Deck` option from the deck editor tab. It does not track local decks opened through the deck storage tab.

## Things to do in future PRs
- Allow maximum recent decks to be configurable
- ~~Remember local decks that are loaded through the deck storage tab~~

### DIFF
--- a/cockatrice/src/client/tabs/tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/tab_deck_editor.cpp
@@ -716,7 +716,7 @@ void TabDeckEditor::retranslateUi()
     aNewDeck->setText(tr("&New deck"));
     aLoadDeck->setText(tr("&Load deck..."));
     loadRecentDeckMenu->setTitle(tr("Load recent deck..."));
-    aClearRecents->setText(tr("Clear items..."));
+    aClearRecents->setText(tr("Clear"));
     aSaveDeck->setText(tr("&Save deck"));
     aSaveDeckAs->setText(tr("Save deck &as..."));
     aLoadDeckFromClipboard->setText(tr("Load deck from cl&ipboard..."));
--- a/cockatrice/src/client/tabs/tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/tab_deck_editor.cpp
@@ -292,7 +292,7 @@ void TabDeckEditor::createMenus()
     aLoadDeck = new QAction(QString(), this);
     connect(aLoadDeck, SIGNAL(triggered()), this, SLOT(actLoadDeck()));
 
-    openRecentsMenu = new QMenu(this);
+    loadRecentDeckMenu = new QMenu(this);
     connect(&SettingsCache::instance().recents(), &RecentsSettings::recentlyOpenedDeckPathsChanged, this,
             &TabDeckEditor::updateRecentlyOpened);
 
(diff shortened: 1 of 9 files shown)


cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -705,6 +715,8 @@ void TabDeckEditor::retranslateUi()
 
     aNewDeck->setText(tr("&New deck"));
     aLoadDeck->setText(tr("&Load deck..."));
+    openRecentsMenu->setTitle(tr("Open Recent..."));
```
ZeldaZach, line 718: I'd like to follow the style we had, and call it "Load recent deck..."

cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -705,6 +715,8 @@ void TabDeckEditor::retranslateUi()
...
+    aClearRecents->setText(tr("Clear Items..."));
```
ZeldaZach, line 719: Lowercase I in items i think

cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -852,6 +864,20 @@ void TabDeckEditor::updateHash()
     hashLabel->setText(deckModel->getDeckList()->getDeckHash());
 }
 
+void TabDeckEditor::updateRecentlyOpened()
+{
+    openRecentsMenu->clear();
```
ZeldaZach, line 869: Does it make sense to try and deleteLater/Disconnect the items in the list before we clear it out and add new pointers? I'm not sure if Qt is handling this destructions or not. Might need to valgrind look.

ZeldaZach:
How does this handle if a deck can't be found at a previous location (Say, someone deleted it?)

RickyRister:
> How does this handle if a deck can't be found at a previous location (Say, someone deleted it?)

Added an error popup

https://github.com/user-attachments/assets/a087dded-694e-44c5-b1e8-e7e9635dd764

cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -852,6 +864,20 @@ void TabDeckEditor::updateHash()
     hashLabel->setText(deckModel->getDeckList()->getDeckHash());
 }
 
+void TabDeckEditor::updateRecentlyOpened()
+{
+    loadRecentDeckMenu->clear();
```
ZeldaZach, line 869: Note to self: Check if this leaks the contents at some point

ZeldaZach (approved):

PR #5320, opened 2024-12-24
Attempt to allow users to temporarily modify preferences (until next card DB update)

BruebachL:
## Related Ticket(s)
- Fixes users asking for this feature in discord

People have to scroll (instead of using the search bar as intended) to find their favorite printings so we have to add this to reduce friction.

- Users will be able to temporarily mark a printing as "preferred"


ZeldaZach (changes requested):
Spoke in DMs, but we'd rather have a new settings file that stores these overrides rather than modifying the cards.xml -- Will wait for a refactor

PR #5329, opened 2024-12-25
Support Picking Select Art per Card Basis

ZeldaZach:
Follow up to #5320


RickyRister:
Seems like it works

https://github.com/user-attachments/assets/01735681-767b-4973-a978-2505aabd7ac0

PR #5332, opened 2024-12-26
Add QSet for faster lookups in CardDatabase

ZeldaZach:
Make loading databases faster by having a set for iterations (Based on FlameGraphs)


RickyRister:
Before
https://github.com/user-attachments/assets/75412142-e54a-406c-978e-e51a90160d24

After
https://github.com/user-attachments/assets/559fcd79-9352-4e6a-8092-dfc619bf27e6

Increase in memory usage is small enough that it gets drowned out by variance when I just look at the total memory of the application

PR #5338, opened 2024-12-26
Allow starting of games with fewer than max number of players

### LINKED ISSUE #434: Game creator should be able to start game, even if not full
This mainly applies to EDH players; if a group sets up for 4 and only has 3 and wants to start, the game creator should have the option to launch the game at any point

ZeldaZach:
## Related Ticket(s)
- Fixes #434

Since the beginning of time, games could only start if everyone was ready'd up and the lobby was full.

Game hosts can now "Force start" a game, which will only include players that have decks loaded and are ready'd up. Players who don't have a deck ready, and empty player slots, will be kicked/removed from the game.

Future improvement will be to support moving the user to a spectator instead of kicking them, but it's not apart of this PR.

## Screenshots

See videos below.

### DIFF
--- a/cockatrice/src/client/tabs/tab_game.cpp
+++ b/cockatrice/src/client/tabs/tab_game.cpp
@@ -106,11 +106,14 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
     loadRemoteButton = new QPushButton;
     readyStartButton = new ToggleButton;
     readyStartButton->setEnabled(false);
+    forceStartGameButton = new QPushButton;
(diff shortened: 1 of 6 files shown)


RickyRister:
### Force starting kicks all non-ready players

https://github.com/user-attachments/assets/a476e5e6-e61b-49b9-9d27-9fcc8738d225

https://github.com/user-attachments/assets/59e5bc76-e0ee-41f3-8933-349e8b34fdab

### Force start validates that the user is host

https://github.com/user-attachments/assets/3654b9eb-e2e7-4b86-989d-56d9a36d0d6c

PR #5348, opened 2024-12-27
Establish Card Ownership & Return on Player Leave

### LINKED ISSUE #1912: Recovering cards from dead players.
Can we input a way to recover a card from a player who has "left" a game. When a player concedes all they own is "taken" off the battlefeild. Is there a way to fix this? so cards taken by effects like Act of treason or Gilded drake can be returned to the owner.

ZeldaZach:
## Related Ticket(s)
- Fixes #1912

When a player conceded/left a game, any cards in their possession at that time would be lost. This includes "borrowed" cards from opponents, from controlling effects, as an example.

- Cards moved to an opponent's board will now display "Owner: Owner Name" in the card annotations, making it clear who owns the card.
  - This is done by allowing a raw card pointer to be passed in instead of depending on a zone lookup later in time. The card was in a weird limbo state where it wasn't technically in any zone, so this is the best way I've found to set the attribute.
- When a player concedes/leaves the game, any card with the Owner tag will attempt to be returned to its rightful owner.

### DIFF
--- a/common/server_player.cpp
+++ b/common/server_player.cpp
@@ -82,6 +82,7 @@
 #include "trice_limits.h"
 
 #include <QDebug>
+#include <QRegularExpression>
 #include <algorithm>
 
 struct MoveCardStruct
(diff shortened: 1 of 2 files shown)


BruebachL:
LGTM.

common/server_player.cpp
```diff
@@ -644,6 +645,16 @@ Response::ResponseCode Server_Player::moveCard(GameEventStorage &ges,
                                   AttrPT, ptString);
             }
```
RickyRister, line 647: leave comment about what this block is for

ebbit1q:
cards that are transferred ownership on concession have their new "ownership" tag set to the conceded player

PR #5182, opened 2024-11-19
New printing selector

BruebachL:
## Related Tickets
- Closes #2411 

## Short roundup of the initial problem
In the current state, users can use individual printings (i.e. their printings are displayed to their opponent) but the selection is locked to the default printing.

## What will change with this Pull Request?
- Add a new PrintingSelectorWidget to the TabDeckEditor to allow users to add/remove specific printings from their deck.

## Screenshots
[Provisional]
Sorted alphabetically:
image

Sorted by preference:
image

PR #5379, opened 2024-12-29
`Printing Selector` naming

tooomm:
## Related Ticket(s)
- Related to #5182

The new `Printing Selector` feature is the only one that does not use its widget title in the menu.

- Uniform naming across the client

If this is about string length, the only alternative I can think of is `Printings`, but then the widget should have the same display name obviously.
IMHO it could be more stressed that this is strictly related to the deck (building) feature, e.g. `Deck - Printings`

Happy to adjust the PR if we can agree on a naming.

<br>

@BruebachL was it intentional that resetting the layout does not show the new widget?

## Screenshots

image


ZeldaZach:
I don't see a name change as necessary, and it was intentionally hidden on startup as not everyone might care about this feature. If you right click a card, you can open this view as well.

tooomm:
I do not see a reason why `Printing Selector` is be the only view which is not correctly named in the menu: 🤷 
image

PR #5381, opened 2024-12-29
Simpler relase channel naming

tooomm:
We use unnecessary extensive names for the release channels.
That leads to issues for certain languages or when a translator does not pay close attention to context or details.

The German translation made me think.

I suggest to use a more simple and straight forward naming for the two release channels.

In the settings, the dropdown option is labelled and described with `Update channel`:
- "Stable Releases" --> `Default`
- "Beta Releases" --> `Beta`


ZeldaZach:
I'm good with this change

PR #5390, opened 2024-12-30
fix cards having the wrong printing if rejoin game before card db finishes loading

RickyRister:

If a player joins a game before their card db finishes loading, the cards won't have the correct printings

This can happen if a user closes cockatrice while they're still in a game and then immediately rejoins the game when they open cockatrice again 

https://github.com/user-attachments/assets/7a01cc44-e386-4d43-a5d7-a1f1736deb52

- rename `AbstractCardInfo::cardInfoUpdated` to `refreshCardInfo` and make it public
- call `refreshCardInfo` on all cards in zone when card db finishes loading


ZeldaZach (approved):
Tested and works Windows & Mac

PR #5416, opened 2025-01-05
improve shortcut search to split by word

RickyRister:

Shortcut search still requires the fixed string to exactly match a continuous section of either the source or the parent. It can't  only match only some words with gaps between, or have the match be split across levels.  

https://github.com/user-attachments/assets/e11ac8fa-bcf0-4e8c-9f19-f49b288f33fc

- Shortcut search now splits the search by word and inserts `.*` between each word
- Shortcut search now appends the parent and source together, so the single combined regex search works across both levels


ZeldaZach:
Fails to build Ubuntu 20.04

ZeldaZach (approved):

PR #5418, opened 2025-01-05
Remember last opened directory when loading decks

### LINKED ISSUE #5402: [macOS] Fails to persist most recently used directory when loading decks
Please remember the last used directory when loading decks, so that we don't have to manually renavigate back to that location over and over.

RickyRister:
## Related Ticket(s)
- Fixes #5402 

https://github.com/user-attachments/assets/bf517bf4-42cb-45a5-82c6-88e3261425cb

- Remember the last opened directory when loading decks and start at that directory when you `load deck` again
  - Setting is stored in `dirs/latestDeckDir` in `recents.ini`
- Affects `Load deck` action in both Deck Editor and Game Lobby

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -56,6 +56,7 @@ set(cockatrice_SOURCES
     src/dialogs/dlg_tip_of_the_day.cpp
     src/dialogs/dlg_update.cpp
     src/dialogs/dlg_view_log.cpp
+    src/dialogs/dlg_load_deck.cpp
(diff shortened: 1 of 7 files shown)


RickyRister:
Currently, there isn't a way to reset the latest dir, which could lead to user getting stuck in the middle of nowhere and not being able to navigate back to the Cockatrice deck folder if they open a deck in their downloads folder, for example. Where would be a good place to put a way to reset the latest dir?

ZeldaZach (approved):

PR #5451, opened 2025-01-12
Make all tabs closable; add tabs menu

RickyRister:
## Related Ticket(s)
- Fixes #1963

## What will change with this Pull Request?

All tabs are now closable. 

Added a `Tabs` menu to the main menuBar, containing the currently openable tabs. (It gets updated on connect/disconnect). 

Open tabs are marked with a checkmark. Tabs can be opened and closed from that menu. That menu will also update if tabs are closed with the X button.

The `Deck Editor` action has also been moved to that menu. It's unique in that it isn't checkable and will always open a new deck editor tab.

https://github.com/user-attachments/assets/36df5d30-4b06-41d7-a5d9-43b0a9776a65

- `Tab`
  - `Tab::closeRequest` calls `close` by default
  - `Tab` now emits `closed` signal on close
    - connecting to the `destroyed` signal in order to react to a tab closing was causing segfaults on exit, so I added a `closed` signal
- `TabSupervisor` 
  - Now creates its tabs through methods that are attached to actions 
  - always attaches close button to created tabs
  - nulls out tab widget and toggles menu check on closed signal
- `MainWindow`
  - removed deck editor action
  - The tabsMenu is created in `WindowMain` and passed into `TabSupervisor`'s constructor 
- Doesn't segfault! (I'm pretty sure, at least)

## Screenshots
Screenshot 2025-01-12 at 2 52 23 AM

PR #5427, opened 2025-01-06
Visual deck storage v2

BruebachL:
## What will change with this Pull Request?
- Restore some button states (ready/sideboard locked) to sensible defaults when unloading a deck.
- Properly replace the visual deck storage widget with a deckView when loading from local file or remote deck.
- DeckLists now store when they were last loaded by a user (otherwise the last loaded timestamp is not modified) [This is a boolean the deckLoader has on load].
- Added the ability to sort by Deck Name and Last Loaded Timestamp.
- Added a color identity widget below the deck preview widgets.

## Screenshots
image
image
image
image
image

PR #5453, opened 2025-01-13
Make Visual Deck Storage tab be managed by TabSupervisor

RickyRister:
## Related Ticket(s)
- Followup to #5451 
- Followup to #5427

Visual Deck Storage is currently like deck editor in that you can open new instances of it. It should probably be a single-instance tab.

https://github.com/user-attachments/assets/314429c9-63fa-436a-acce-cf4973e89106

- Make visual deck storage a single-instance tab that is managed by the `Tabs` menu.
- Move visual deck storage stuff from `WindowMain` into `TabSupervisor`
- `VisualDeckStorageWidget` now refreshes itself on db load
  - Tab now immediately opens on startup, instead of after db load finishes
  - Still respects the settings for opening visual deck storage on startup

## Screenshots

Screenshot 2025-01-12 at 8 07 42 PM

### DIFF
--- a/cockatrice/src/client/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/client/tabs/tab_supervisor.cpp
@@ -164,6 +164,9 @@ TabSupervisor::TabSupervisor(AbstractClient *_client, QMenu *tabsMenu, QWidget *
     resetTabsMenu();
 
     retranslateUi();
+
+    // open always-available tabs on startup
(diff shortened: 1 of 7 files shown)


ZeldaZach (approved):

BruebachL:
Seems reasonable, I am curious how this interacts with #5456

BruebachL:
Interacts well with #5456 and makes switching to the tab snappier but it's not easy to tell since it now opens before the deck editor. Please make that a settings option at least.

ZeldaZach (approved):

PR #5455, opened 2025-01-13
refactor DeckViewContainer into own file

RickyRister:

`DeckViewContainer` is starting to get big enough that it should be in its own file

- Moved `DeckViewContainer` into own file
  - File is in the same `tabs` directory as `tab_game`
- move `TabGame::refreshShortcuts` method to below the `TabGame` constructors


RickyRister:
Which folder should `deck_view_container` go in? I just put it in the same directory as `tab_game` but I feel like that's probably wrong.

ZeldaZach:
> Which folder should `deck_view_container` go in? I just put it in the same directory as `tab_game` for now but I feel like that's probably wrong.

Should probably go in src/game/deckview or something along those lines

RickyRister:
also moved `deck_view` into the new folder

PR #5462, opened 2025-01-14
Disable Force Start for host on load

### LINKED ISSUE #5444: `Force start` button active before deck selection
**System Information:**
Client Version: 2.10.0-beta.24 (2025-01-11)
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_DE
Install Mode: Standard

The `Force start` button is active but does nothing when just joined/created a game and no deck is selected (tested in local game):

image

Same as the other buttons, it should only be active once a deck is loaded.
In general it should only be active if the amount of players in the room/game is below the max player count.

**Steps to reproduce:**
 - Start local game
 - Do not select a deck
 - Button is active and can be clicked without action

ZeldaZach:
- Fix #5444


tooomm:
Windows builds are failing now

ZeldaZach:
Thanks, Tom. Fixed in a push to master.

PR #5467, opened 2025-01-14
Remember which tabs are open between sessions

### LINKED ISSUE #5442: Client always starting (delayed) in `Visual Deck Storage` view
**System Information:**
Client Version: 2.10.0-beta.24 (2025-01-11)
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_DE
Install Mode: Standard

Opening the client starts as usual in `Deck Editor`, but with a delay and after loading the DB, the new `Visual Deck Storage` tab is loaded and focused.

While this might be convenient for those working with deck files, others just use the card database of Cockatrice and tinker where they left off last time (that would be a new feature on its own), just use the DB to look something up or start a new brew.
There are also people who just play limited and do not care about deck building/loading in the editor.

I would like to see an option in setting on the default tab, possibly also allow to disable the opening of the deck storage view altogether.
Remembering the state (loaded decks and focused/available tabs in Deck Editor) when closing the client would also be nice. Or a combination of both.

RickyRister:
## Related Ticket(s)
- Fixes #5442
- Requires #5464 to be merged first

The open state of each managed tab is now persisted in settings. Cockatrice will remember which tabs are open between sessions, and between connect/disconnect.

A tab being forcibly closed will not update the persisted state. That means, for example, if you disconnect, then all server-only tabs that were open when you disconnected will be opened again when you reconnect.

- Added the new settings to `global.ini` under `tabs`
  - Removed the `Open Visual Deck Storage on load` setting, since it's redundant now.
- Slightly refactored how TabSupervisor manages the single-instance tabs and how the tab menu actions trigger, in order to facilitate the new changes
  - Moved `addCloseButtonToTab` closer to `myAddTab`
    - added new param to those two methods
    - For single-instance tabs, the close button's trigger will now uncheck the tab's menu action; the menu action will handle the closing of the tab.

### DIFF
--- a/cockatrice/src/client/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/client/tabs/tab_supervisor.cpp
@@ -281,7 +281,15 @@ void TabSupervisor::initStartupTabs()
     checkAndTrigger(aTabVisualDeckStorage, SettingsCache::instance().getVisualDeckStorageShowOnLoad());
 }
 
-int TabSupervisor::myAddTab(Tab *tab)
+/**
+ * Adds the tab to the TabSupervisor's tab bar.
+ *
+ * @param tab The Tab to add
+ * @param manager The menu action that corresponds to this tab, if this is a single-instance managed tab. Pass in
+ * nullptr if this is not a managed tab.
+ * @return The index of the added tab in the tab widget's tab menu
+ */
+int TabSupervisor::myAddTab(Tab *tab, QAction *manager)
 {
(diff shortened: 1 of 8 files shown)


ZeldaZach:
Close All Tabs (Esp Accounts) > Disconnect > Connect > Crash

Accounts Tab, if not created, won't let the program function. I'll have to wait for the refactor of the TabAccounts before considering a merge on this.

RickyRister:
~~What, you don’t think putting the user in an inescapable crash loop is pure comedy?~~

ZeldaZach (approved):

PR #5497, opened 2025-01-18
Give deckList a signal to emit when the tags change and hook up the display widget to that.

BruebachL:
Editing a decks tags causes all of the bannerCard widgets to get reconstructed.

- Give deckList a signal to emit when the tags change and hook up the display widget to that instead of refreshing all banner cards.
- DeckViewContainer now reloads the DeckPreviewWidgets deck from file when loading a deck to ensure it has the latest decklist if it has been modified.

### DIFF
--- a/cockatrice/src/game/deckview/deck_view_container.cpp
+++ b/cockatrice/src/game/deckview/deck_view_container.cpp
@@ -181,6 +181,7 @@ void DeckViewContainer::refreshShortcuts()
 void DeckViewContainer::loadVisualDeck(QMouseEvent *event, DeckPreviewWidget *instance)
 {
     Q_UNUSED(event);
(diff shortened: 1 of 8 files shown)


cockatrice/src/game/deckview/deck_view_container.cpp
```diff
@@ -181,6 +181,7 @@ void DeckViewContainer::refreshShortcuts()
 void DeckViewContainer::loadVisualDeck(QMouseEvent *event, DeckPreviewWidget *instance)
 {
     Q_UNUSED(event);
+    instance->deckLoader->loadFromFile(instance->filePath, DeckLoader::CockatriceFormat, false);
```
RickyRister, line 184: is there a way to make this call `DeckViewContainer::loadDeckFromFile`, instead of duplicating the code?

ZeldaZach (approved):

Issue #4744, opened 2023-02-07
Easier to View Double Sided Cards


xTrae:
Feature idea:
I propose a change to how double sided cards are viewed, or can be viewed. Maybe as a toggleable mode in settings, so you can still use the current method if you wanted.

When you hover over a card, you can see a large version of the card in the side window. However, you can only ever see the front face of the card there. You have to go through an awkward process to view the other side, especially if you want to view a lot of them. Or rather it's impossible to see and keep track of if you have 5+ modal double sided cards. I would absolutely love to be able to see both sides of cards there, similar to what MTGO or Magic Arena does, especially for non-Magic games. This isn't asking about ease of playing/using the cards, as there have been feature requests years in the past. I'm just talking about being able to view the cards.

Perhaps three different modes would be most helpful, Classic/Default/Front Only, View Both, View Back Only.

ebbit1q:
if we want to have both cards show up in the card info widget where is that extra space coming from? there is pretty little space over there already to properly show everything. showing all related cards with the middle mouse button popout sounds pretty feasible though.

xTrae:
A different kind of mouse clickable popup would be equally unhelpful for viewing many doubled sided cards at once. If I have multiple in hand, I should be able to hover over it and see those related cards.

I don't believe you need any extra space. You just show two cards in the same area, making both of them smaller. Which yes, may be unpreferable to some people, especially those who only playing Magic. But when you have a game where everything is double sided... there's just no game simulator platform with that kind of ease of play.

tooomm:
>showing all related cards with the middle mouse button popout sounds pretty feasible though.

👍 
I also think displaying card picture considerable smaller, just to show two next to each other is not really helpful. In the middle mouse button pop-out, this is probably no issue as ebbit noted.

<br>

Currently, in the card info widget it's not possible to interact with the card art (in deck editor and during gameplay). Right or left clicking do not have any effect or give additional options.

Maybe utilize mouse over or mouse clicks there for DFCs?
- Switch shown card face on left click?
- Switch image to card back on hover?

This should only account for DFC cards though, not related cards.
If you want to consider them as well, it could be switching to the back side on hover, and displaying a menu with related cards on right click (and different art/printing at one point)?

In general, DFCs could have an indicator on the card to make clear they have additional content on the back. This icon could then also be interacted with to display other faces or even related cards.

PR #5498, opened 2025-01-19
Add a button to easily view the transformed version of a card.

BruebachL:
## Related Ticket(s)
- Follow up to #4744

- Add a new button to the CardInfoFrameWidget that will be visible if the card has a cardrelation that doesAttach. When clicked, it sets the CardInfoFrameWidgets info to that card.

## Screenshots
image
image
image


RickyRister:
Slightly redundant with #5375, but doesn't hurt to have more options

PR #5499, opened 2025-01-19
Hide load from remote button in local games

### LINKED ISSUE #5495: Remove `Load remote deck...` button  in local game
**System Information:**
Client Version: 2.10.0-beta.31 (2025-01-18)

In a local game, the button has no use when playtesting on your own in the client.
Not sure there is a use at all?

Image

<br>

I'm also still not sure if the `Force start` button after loading a deck, behaves as it should in local games.

BruebachL:
## Related Ticket(s)
- Fixes #5495

The load from remote button is visible in local games which doesn't make much sense.

- Change the visibility of load from remote to only be visible if the parentGame is not a local game.

### DIFF
--- a/cockatrice/src/game/deckview/deck_view_container.cpp
+++ b/cockatrice/src/game/deckview/deck_view_container.cpp
@@ -65,6 +65,7 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
     connect(sideboardLockButton, SIGNAL(stateChanged()), this, SLOT(updateSideboardLockButtonText()));
 
     if (parentGame->getIsLocalGame()) {
+        loadRemoteButton->setVisible(false);
         loadRemoteButton->setEnabled(false);
     } else {
         connect(loadRemoteButton, SIGNAL(clicked()), this, SLOT(loadRemoteDeck()));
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/deckview/deck_view_container.cpp
```diff
@@ -65,6 +65,7 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
     connect(sideboardLockButton, SIGNAL(stateChanged()), this, SLOT(updateSideboardLockButtonText()));
 
     if (parentGame->getIsLocalGame()) {
+        loadRemoteButton->setVisible(false);
```
RickyRister, line 68: Actually, you can just remove this check and set in the constructor, since it's already going to be handled in `switchToDeckSelectView`.

BruebachL: Shrug. It's a constructor meant for defining default-behavior, the button is already being disabled in the existing check so it shouldn't be visible.
Anyway, this is sorta nitpicky for a one-time inexpensive function call to hide a widget in a constructor.

ZeldaZach (approved):

PR #5502, opened 2025-01-20
Add setting to auto close card view when empty

### LINKED ISSUE #1893: Auto close card view when empty
As a player:
When I scry 1 and move the card to the bottom, I want the window to auto close.
I also want this to happen any other time a similar window would have no cards in it.

RickyRister:
## Related Ticket(s)
- Fixes #1893

https://github.com/user-attachments/assets/e09006ea-a927-4740-aac2-e047c9d7e3bf

- Added setting `Close card view window when last card is removed` under `User Interface` 
- When settings is enabled, the card view window will automatically close when a card removal from the window causes the window to become empty.
  - You can still open empty card view windows; the automatic closing only triggers when a card removal causes the window to become empty

## Screenshots

Screenshot 2025-01-19 at 5 17 43 PM

## Known issues

- When only a single card is in the zone view, dragging the card into the same zone will cause the window to close.
  - Only happens when there is a single card in the zone view; two or more cards are handled fine
  - Not sure how relevant this actually is, since you usually don't have a reason to be dragging a single card into the same zone, unless it's by mistake or something

https://github.com/user-attachments/assets/deab1bd8-67ed-4fcb-8270-d65d95d8f31d

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -561,6 +561,10 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     connect(&playToStackCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance(),
             &SettingsCache::setPlayToStack);
 
+    closeEmptyCardViewCheckBox.setChecked(SettingsCache::instance().getCloseEmptyCardView());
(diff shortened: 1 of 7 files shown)


BruebachL:
Implemented exactly as the ticket agreed on, great work and thank you for doing it.

ZeldaZach (approved):

PR #5506, opened 2025-01-20
Convert lastLoaded timestamp into proper QDateTime for correct comparison.

BruebachL:
Stuff isn't sorting correctly because I forgot to convert the lastLoadedTimestamp back to a QDateTime on comparison (in my defense, lastModified just above works just fine (because the QFileInfo generates it :P)

- Sort VDS properly on last load property.


cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_sort_widget.cpp
```diff
@@ -67,10 +67,6 @@ QList<DeckPreviewWidget *> &VisualDeckStorageSortWidget::filterFiles(QList<DeckP
 {
     // Sort the widgets list based on the current sort order
     std::sort(widgets.begin(), widgets.end(), [this](DeckPreviewWidget *widget1, DeckPreviewWidget *widget2) {
-        if (!widget1 || !widget2) {
```
ZeldaZach, line 70: Any reason we removed this null check?

BruebachL: Copy and pasted some stuff from an earlier version, looks like I whoopsied and selected a bit too much.

ZeldaZach (approved):

PR #5510, opened 2025-01-22
Remove spacing in CardInfoWidget caused by invisible button

RickyRister:
Attempt 2, this time without creating a new class

## Related Ticket(s)
- Fixes issue introduced in #5498

Card Info Widget now has extra space at the bottom due to the addition of the `view transformation` button, even when the button isn't showing.

before

https://github.com/user-attachments/assets/cd6cbb8e-e418-47da-8ded-1ce84e560364

Entirely delete the view transformation `QButton` when it's not in use, instead of just hiding it.

- Refactor refreshing the layout to a separate method
- Delete the button entirely when it's not visible
  - keep `viewTransformationButton` as a nullptr when card has no transformation
  - Code now null checks `viewTransformationButton` whenever it needs to do something with it
  - When transformable card appears, button is created and then refresh layout is called
- view transformation button still works as before

### DIFF
--- a/cockatrice/src/client/ui/widgets/cards/card_info_frame_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/cards/card_info_frame_widget.cpp
@@ -12,7 +12,7 @@
 #include <utility>
 
 CardInfoFrameWidget::CardInfoFrameWidget(const QString &cardName, QWidget *parent)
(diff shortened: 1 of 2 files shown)


BruebachL:
Yes, I believe that's the proper way to do it. Thanks for having another stab at this :)

ZeldaZach (approved):

PR #5514, opened 2025-01-23
Add a dialog to prompt user to convert to .cod format if trying to apply tags to a .txt deck.

BruebachL:
Users can currently apply tags to .txt files, which causes them to be saved in .cod format but with a .txt filename, making them unusable. This adds a dialog to ask if the user would like to convert the deck to .cod format if trying to apply a tag. The dialog can also be set to remember the choice and apply it for all future operations without prompting. This can be re-enabled/disabled respectively in the settings.


cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_deck_format_conversion_dialog.cpp
```diff
@@ -0,0 +1,39 @@
+#include "deck_preview_tag_deck_format_conversion_dialog.h"
+
+#include <QCheckBox>
+#include <QDialogButtonBox>
+#include <QLabel>
+#include <QVBoxLayout>
+
+DeckPreviewTagDeckFormatConversionDialog::DeckPreviewTagDeckFormatConversionDialog(QWidget *parent) : QDialog(parent)
+{
+    setWindowTitle("Deck Format Conversion");
```
ZeldaZach, line 10: tr() and in refreshUi()

cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_deck_format_conversion_dialog.h
```diff
@@ -0,0 +1,29 @@
...
+#include <QDialogButtonBox>
+#include <QLabel>
+#include <QVBoxLayout>
+
+class DeckPreviewTagDeckFormatConversionDialog : public QDialog
+{
+    Q_OBJECT
+
+public:
+    explicit DeckPreviewTagDeckFormatConversionDialog(QWidget *parent = nullptr);
```
ZeldaZach, line 15: I don't feel this should support a null parent

cockatrice/src/deck/deck_loader.cpp
```diff
@@ -481,6 +481,52 @@ void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out,
     }
 }
 
+bool DeckLoader::convertToCockatriceFormat(QString fileName)
+{
+    // Change the file extension to .cod
+    QFileInfo fileInfo(fileName);
+    QString newFileName = fileInfo.path() + "/" + fileInfo.completeBaseName() + ".cod";
```
ZeldaZach, line 488: This might not be true on all OS'es, might want to use the fileInfo or QDir::toNativeSeparators

cockatrice/src/deck/deck_loader.cpp
```diff
@@ -481,6 +481,52 @@ void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out,
...
+    switch (getFormatFromName(fileName)) {
+        case PlainTextFormat:
+            // Save in Cockatrice's native format
+            result = saveToFile_Native(&file);
+            break;
+        case CockatriceFormat:
+            qCInfo(DeckLoaderLog) << "File is already in Cockatrice format. No conversion needed.";
+            result = true;
+            break;
+        default:
```
ZeldaZach, line 509: nit: no break from default

cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_deck_format_conversion_dialog.h
RickyRister, line 1: do we want to put this file in `src/dlg` instead?

cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_deck_format_conversion_dialog.h
BruebachL: It's a dialog, as a free floating widget, it always gets a nullptr parent.

cockatrice/src/deck/deck_loader.cpp
BruebachL: Don't really need to since the path ends there anyway but sure

ZeldaZach (approved):

PR #5529, opened 2025-01-26
refactor CardInfoPictureWidget::createRightClickMenu

RickyRister:

`CardInfoPictureWidget::createRightClickMenu` is a bit messy 

- Split the creation of each sub-menu into its own method
- use `std::any_of` instead of a for loop


BruebachL:
Very much in favor of this, thank you.

PR #5512, opened 2025-01-22
Edhrec tab

BruebachL:
## Short roundup of the initial problem
Deck building can be significantly enhanced if users have access to external APIs like EDHRec from within the client.
This change adds such a tab for the EDHRec API.

## What will change with this Pull Request?
- Card entries in the database display gain two new right click actions (View on EDHRec as card and View on EDHRec as commander, if they meet the requirements)
- This opens a tab which displays the returned EDHRec API json data.
- Users can add the cards to any open deck editors by using a new right click context menu action.

## Screenshots

https://github.com/user-attachments/assets/b024b405-190c-4e43-8040-5b1f1861a841

PR #5530, opened 2025-01-26
Disable add to deck submenu if no deck editor tabs are open

RickyRister:
## Related Ticket(s)
- Follow-up to #5512
- requires #5529 to be merged first

The "Add card to deck" menu is enabled even when no deck editor tabs are open; it just doesn't do anything when hovered over.

Screenshot 2025-01-26 at 2 11 19 AM

https://github.com/user-attachments/assets/a7d982a0-f2ea-477b-af98-72f264e775a2

- "Add card to deck" menu is disabled when no deck editor tabs are open
- Rename the menu options from `"Add card to" + tab name` to just `tab name` since it's already obvious that you're adding the card to a deck


BruebachL:
Looks sensible, seems to work great, approved. Thank you!

ZeldaZach (approved):

PR #5554, opened 2025-02-03
preserve image quality when scaling svg

RickyRister:

Currently, when we need a pixmap of an svg, we load it into the pixmap first and then scale the pixmap, which destroys the image quality.

- Scale the svg itself instead of the pixmap, so the image quality doesn't get destroyed

## Screenshots
Before

Screenshot 2025-02-02 at 5 27 32 PM

Screenshot 2025-02-02 at 5 28 02 PM

After

Screenshot 2025-02-02 at 5 28 20 PM

Screenshot 2025-02-02 at 5 28 47 PM


ZeldaZach (approved):

RickyRister:
fixed the colorless counter

Screenshot 2025-02-02 at 6 42 44 PM

Issue #5558, opened 2025-02-03
Increase Deck Size Limit Slightly


fluidvanadium:
Increase limit from 25,000 to 100,000

I have created a deck with 1 of every card ever printed. It actually plays
-if I remove every card starting with A or B
-if I remove every card starting with T or further

but If the deck contains more than 25,014 cards, loading causes an error:
'The selected file could not be loaded.'

PR #5559, opened 2025-02-03
Deck limit

fluidvanadium:
Hello, this is my first PR for Cockatrice.

Increased deck size limit
#5558


ZeldaZach:
Hey there, and welcome! I'm not certain if increasing the file size limit like this is the best idea as it opens up more avenues for DoS. Is there a way we can marginally increase the size slightly, compared to 16x-ing?

fluidvanadium:
Amended to a 2x increase.

ZeldaZach (approved):

PR #5586, opened 2025-02-09
Refactor Visual Deck Storage sorting

RickyRister:

Sorting code in Visual Deck Storage is unintuitive. The `VisualDeckStorageFolderDisplayWidget` gets passed into the `VisualDeckStorageSortWidget` and then the sort widget recursively goes through and sorts the `DeckPreviewWidget`s that are inside of the `VisualDeckStorageFolderDisplayWidget`.

`VisualDeckStorageFolderDisplayWidget` contains the `DeckPreviewWidget`s so the sorting should really be happening in that class.

- Sorting of widgets is done inside `VisualDeckStorageFolderDisplayWidget` instead of `VisualDeckStorageSortWidget`
  - `VisualDeckStorageFolderDisplayWidget::sortBy` takes a comparator and sorts all decks in it and all subfolders
  - `VisualDeckStorageSortWidget::getComparator` returns a comparator that sorts according to the current `sortOrder`
- On sort update, the top-level `VisualDeckStorageWidget` gets the comparator from its `VisualDeckStorageSortWidget` and calls `sortBy` on the top-level `VisualDeckStorageFolderDisplayWidget`, which then recursively sorts all the decks

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_folder_display_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_folder_display_widget.cpp
@@ -181,6 +181,32 @@ void VisualDeckStorageFolderDisplayWidget::updateShowFolders(bool enabled)
 }
 
 /**
+ * Sorts the DeckPreviewWidgets in this folder and all subfolders
+ *
+ * @param comparator The comparator to use for sorting
+ */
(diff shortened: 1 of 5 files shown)


BruebachL:
I vote against this refactor.

ZeldaZach:
Any updates on reviews @BruebachL

ebbit1q:
did this get outdated or is this pr still good?

PR #5591, opened 2025-02-09
Put Logging config on MacOS 

ZeldaZach:


RickyRister:
Tested it on my machine. Confirmed that it works now

PR #5609, opened 2025-02-11
Fix Ricky Crash

ZeldaZach:
```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   cockatrice                    	       0x102d48990 StackZone::handleDropEvent(QList<CardDragItem*> const&, CardZone*, QPoint const&) + 792
1   cockatrice                    	       0x102b97304 CardDragItem::mouseReleaseEvent(QGraphicsSceneMouseEvent*) + 528
2   QtWidgets                     	       0x10414c5b8 QGraphicsItem::sceneEvent(QEvent*) + 68
3   QtWidgets                     	       0x104163fe4 QGraphicsScenePrivate::sendMouseEvent(QGraphicsSceneMouseEvent*) + 392
4   QtWidgets                     	       0x10416a2fc QGraphicsScene::mouseReleaseEvent(QGraphicsSceneMouseEvent*) + 44
5   QtWidgets                     	       0x10416759c QGraphicsScene::event(QEvent*) + 736
6   cockatrice                    	       0x102c68404 GameScene::event(QEvent*) + 72
```

### DIFF
--- a/cockatrice/src/game/zones/stack_zone.cpp
+++ b/cockatrice/src/game/zones/stack_zone.cpp
@@ -28,7 +28,7 @@ void StackZone::addCardImpl(CardItem *card, int x, int /*y*/)
 {
     // if x is negative set it to add at end
     if (x < 0 || x >= cards.size()) {
-        x = cards.size();
+        x = static_cast<int>(cards.size());
     }
     cards.insert(x, card);
 
@@ -44,7 +44,7 @@ void StackZone::addCardImpl(CardItem *card, int x, int /*y*/)
 
 QRectF StackZone::boundingRect() const
 {
-    return QRectF(0, 0, 100, zoneHeight);
+    return {0, 0, 100, zoneHeight};
 }
 
 void StackZone::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*option*/, QWidget * /*widget*/)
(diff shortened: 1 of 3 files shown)


cockatrice/src/game/zones/stack_zone.cpp
```diff
@@ -28,7 +28,7 @@ void StackZone::addCardImpl(CardItem *card, int x, int /*y*/)
 {
     // if x is negative set it to add at end
     if (x < 0 || x >= cards.size()) {
-        x = cards.size();
+        x = static_cast<int>(cards.size());
```
RickyRister, line 31: Why do we need to static cast the size to an int? Isn’t it already an int?

cockatrice/src/game/zones/stack_zone.cpp
```diff
@@ -69,19 +69,30 @@ void StackZone::handleDropEvent(const QList<CardDragItem *> &dragItems, CardZone
...
+    int index = 0;
     if (cards.isEmpty()) {
         index = 0;
     } else {
-        const int cardCount = cards.size();
+        const auto cardCount = static_cast<int>(cards.size());
+        const auto &card = cards.at(0);
+
+        if (card == nullptr) {
+            return;
```
RickyRister, line 81: I wonder if we should log whenever we early return due to an unexpected null. That way if it unexpectedly no-ops due to a failed null check we at least have some indication

cockatrice/src/game/zones/stack_zone.cpp
ZeldaZach: it's technically an unsigned long, this was just a warning getting cleaned up

cockatrice/src/game/zones/stack_zone.cpp
ZeldaZach: I can add logging, sure

Issue #5618, opened 2025-02-15
Add right-click menu to Visual Deck Storage


RickyRister:
**Description of New Feature**

We should add a right-click menu to the decks in the visual deck storage. This will allow us to add more actions to the visual deck storage, without having to take up more space by adding widgets to the layout.

**Additional Context**

Possible actions to put in the menu:
- Alternative for existing actions
  - Open deck
  - Edit tags
  - Set banner card
- New actions
  - Rename deck
  - Rename file
  - Delete deck
  - Save deck to clipboard

PR #5621, opened 2025-02-15
refactor how double click signal gets passed up in VDS

RickyRister:
## Related Ticket(s):
- Precursor to #5618

The double click signal originating from `DeckPreviewCardPictureWidget` gets passed as a double click signal (containing the MouseEvent and the `DeckPreviewWidget` itself) all the way up to `TabDeckStorageVisual`, which then extracts the filePath from the passed `DeckPreviewWidget` to load the deck. 

We do not need to pass the double-click signal all the way up when all we need is the filepath.

The main reason I wanted to do this is because moving `deckLoadRequested` to `DeckPreviewWidget` means I can emit that signal from other places in `DeckPreviewWidget`.

- convert the double click signal into a `deckLoadRequested(filePath)` signal at the `DeckPreviewWidget` level.
- remove unnecessary `deckPreviewClicked` and `deckPreviewDoubleClicked` signals from classes up the chain.

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
@@ -6,6 +6,7 @@
 #include "deck_preview_deck_tags_display_widget.h"
 
 #include <QFileInfo>
+#include <QMenu>
 #include <QMouseEvent>
 #include <QSet>
 #include <QVBoxLayout>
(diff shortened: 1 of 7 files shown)


cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
```diff
@@ -258,12 +259,13 @@ void DeckPreviewWidget::setBannerCard(int /* changedIndex */)
 
 void DeckPreviewWidget::imageClickedEvent(QMouseEvent *event, DeckPreviewCardPictureWidget *instance)
 {
```
ZeldaZach, line 261: if this signal is double no-op'd for args, do we even need it as a signal?

ZeldaZach (approved):

cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
RickyRister: I'm planning on adding a right-click menu

PR #5630, opened 2025-02-17
Always download macOS 13 version on intel macs

RickyRister:

Intel macs are only supported in the macOS 13 version. The release channel should take that into consideration when finding the version to download.

- Use the equivalent of `sysctl -n machdep.cpu.brand_string | grep 'Intel'` to determine if user is on an intel mac
  - `QSysInfo` won't work here because it'll still report `x86` if the app is running through [Rosetta](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment)

### DIFF
--- a/cockatrice/src/client/network/release_channel.cpp
+++ b/cockatrice/src/client/network/release_channel.cpp
@@ -60,8 +60,7 @@ bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
         size_t len = sizeof(arch);
         if (sysctlbyname("machdep.cpu.brand_string", arch, &len, nullptr, 0) == 0) {
             // Intel mac is only supported on macOS 13 versions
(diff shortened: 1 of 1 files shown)


cockatrice/src/client/network/release_channel.cpp
```diff
@@ -48,8 +53,24 @@ bool ReleaseChannel::downloadMatchesCurrentOS(const QString &fileName)
         return false;
     }
 
+    auto getSystemVersion = [] {
```
ebbit1q, line 56: what is the practical reason for making a lambda that immediately gets called and destroyed?

RickyRister: it's already inside of an ifdef so I didn't want to extract it to a separate function. The compiler should be smart enough to convert it to a normal function call.

ZeldaZach (approved):

PR #5634, opened 2025-02-17
add "open in deck editor" to VDS right-click menu

RickyRister:
## Related Ticket(s)
- Relates to #5618

Sometimes, I load into a game, load up a deck, then realize that I want to make a few tweaks to the deck before I start. Previously, I would need to go to the visual deck storage tab or deck editor tab and find that deck again.

Replaced the "Load Deck" action with "Open in deck editor". This action always opens the deck in a new deck editor tab, regardless of where the visual deck storage is. This is equivalent to double-clicking when in the Visual Deck Storage tab, but allows you to do something you couldn't before when in a game lobby.

Double clicking still has existing behavior; It will open deck in deck editor when in the Visual Deck Storage tab and load deck when in a game lobby.

https://github.com/user-attachments/assets/ee3bc0c3-a1ff-44e3-937b-990abedc1206

- Added `openDeckEditor(const DeckLoader *deck)` signal to some classes up the chain, so that we can pass the signal from `DeckPreviewWidget` to `TabGame` and `TabDeckStorageVisual`.
- Replace "Load Deck" menu action with the new action and make it emit the new signal

## Screenshots

Screenshot 2025-02-17 at 12 19 53 AM


BruebachL:
I don't mind these right click menus for now but be aware that ultimately, I would like to jazz up the UI to be a bit more modern and move to a selection/quickbar menu akin to MTG:Arena

image

PR #5640, opened 2025-02-19
Refresh VDS when deck is saved in deck editor tab

RickyRister:
## Related Ticket(s)
- Follow-up to #5636
- Alternative implementation of feature in #5608

We had to revert the file watcher because it was causing usability issues. However, having the VDS auto-refresh is still a very nice QoL feature.

VDS now auto-refreshes whenever you save a deck in the Deck Editor Tab.

https://github.com/user-attachments/assets/1f4dddde-21a3-491d-b356-a036a0dc2e19

- Created a `DeckEditEventBus` singleton object in order to pass the signal from `TabDeckEditor` to `VisualDeckStorageWidget`
  - I know singletons are usually an anti-pattern, but I think doing it this way is still much cleaner than the alternative (where we pass the signal all the way across the object tree)
- `TabDeckEditor` gets `DeckEditEventBus` to emit `deckModified` signal on save
- `VisualDeckStorageWidget` refreshes upon receiving the signal from `DeckEditEventBus`

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -31,6 +31,7 @@ set(cockatrice_SOURCES
     src/server/chat_view/chat_view.cpp
     src/game/board/counter_general.cpp
     src/deck/custom_line_edit.cpp
+    src/deck/deck_edit_event_bus.cpp
     src/deck/deck_loader.cpp
     src/deck/deck_list_model.cpp
     src/deck/deck_stats_interface.cpp
--- a/cockatrice/src/client/tabs/tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/tab_deck_editor.cpp
@@ -746,7 +746,7 @@ TabDeckEditor::TabDeckEditor(TabSupervisor *_tabSupervisor) : Tab(_tabSupervisor
     connect(&SettingsCache::instance().shortcuts(), SIGNAL(shortCutChanged()), this, SLOT(refreshShortcuts()));
     refreshShortcuts();
 
-    connect(this, &TabDeckEditor::deckSaved, DeckEditEventBus::instance(), &DeckEditEventBus::deckModified);
+    connect(this, &TabDeckEditor::deckSaved, DeckEditEventBus::instance(), &DeckEditEventBus::deckFileModified);
 
     loadLayout();
 }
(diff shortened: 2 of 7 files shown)


BruebachL:
I don't think the event bus singleton is necessary here. You can find the top-level window and from there on findChildren the deck editors/storage tabs respectively to hook up to the signals/slots.

https://doc.qt.io/qt-6/qobject.html#findChildren

RickyRister:
The main reason for using an event bus singleton is so that tabs do not need to know the entire structure of the object tree.

> You can find the top-level window and from there on findChildren the deck editors/storage tabs respectively to hook up to the signals/slots.

The problem with that design is, which class is responsible for doing the tree lookup and hooking up the signal/slots? New `DeckEditorTab`s and `VisualDeckStorageWidget`s can be created ad-hoc, and we need to make sure that every `DeckEditorTab`'s signal is connected to every `VisualDeckStorageWidget`'s slot, no matter what order they get created in. 

I don't think that allows us to set up the connection from within those classes themselves, so that means we need some other object to act as the mediator.

BruebachL:
> The main reason for using an event bus singleton is so that tabs do not need to know the entire structure of the object tree.

You can find the top-level window and from there on findChildren the deck editors/storage tabs respectively to hook up to the signals/slots.

> The problem with that design is, which class is responsible for doing the tree lookup and hooking up the signal/slots? New `DeckEditorTab`s and `VisualDeckStorageWidget`s can be created ad-hoc, and we need to make sure that every `DeckEditorTab`'s signal is connected to every `VisualDeckStorageWidget`'s slot, no matter what order they get created in.

They can simply hook up to each other on instantiation. Here is a diagram. Every DeckEditor (left) scans for and hooks up to each VDS and vice versa.

image

> I don't think that allows us to set up the connection from within those classes themselves, so that means we need some other object to act as the mediator. That object would need to have its own signal that the `DeckEditorTab`s connect to as a slot and the `VisualDeckStorageWidget`s connect to as a signal, in order to collapse the n-to-n relation into a n-to-1 plus 1-to-n relation.

As you've mentioned, we have the TabSupervisor already whose job it is, literally, to supervise the lifetime of tabs and perform appropriate actions.

> I suppose we could have `TabSupervisor` act as the mediator, but that means we need to send up signal up to `TabSupervisor` every time a `VisualDeckStorageWidget` gets created inside the `DeckViewContainer` in order to wire it up, which I felt like was starting to become really messy. Additionally, we would need to add that new signal to `TabSupervisor`.

Regardless, I maintain that there are definitely more "Qt"-ways of doing this.
https://doc.qt.io/qt-6/eventsandfilters.html

PR #5646, opened 2025-02-22
Generic deck editor

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/client/tabs/tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/tab_deck_editor.cpp
@@ -11,6 +11,7 @@
 #include "../../game/filters/filter_tree_model.h"
 #include "../../server/pending_command.h"
 #include "../../settings/cache_settings.h"
+#include "../menus/deck_editor/deck_editor_menu.h"
 #include "../ui/pixel_map_generator.h"
 #include "../ui/widgets/deck_editor/deck_editor_filter_dock_widget.h"
 #include "tab_supervisor.h"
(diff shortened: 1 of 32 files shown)


cockatrice/src/client/tabs/tab_deck_editor.cpp
```diff
@@ -1084,624 +167,193 @@ void TabDeckEditor::openDeckFromFile(const QString &fileName, DeckOpenLocation d
...
-    QFileDialog dialog(this, tr("Save deck"));
-    dialog.setDirectory(SettingsCache::instance().getDeckPath());
-    dialog.setAcceptMode(QFileDialog::AcceptSave);
-    dialog.setDefaultSuffix("cod");
-    dialog.setNameFilters(DeckLoader::fileNameFilters);
-    dialog.selectFile(deckModel->getDeckList()->getName().trimmed() + ".cod");
-    if (!dialog.exec())
-        return false;
+    deckMenu->retranslateUi();
+    cardInfoDockWidget->cardInfo->retranslateUi();
```
ZeldaZach, line 191: is cardInfo guaranteed to be non-null?

cockatrice/src/client/ui/widgets/deck_editor/deck_editor_filter_dock_widget.cpp
```diff
@@ -0,0 +1,142 @@
...
+    setFeatures(QDockWidget::DockWidgetClosable | QDockWidget::DockWidgetFloatable | QDockWidget::DockWidgetMovable);
+
+    createFiltersDock();
+}
+
+void DeckEditorFilterDockWidget::createFiltersDock()
+{
+    filterModel = new FilterTreeModel();
+    filterModel->setObjectName("filterModel");
+    deckEditor->databaseDisplayDockWidget->databaseDisplayModel->setFilterTree(filterModel->filterTree());
```
ZeldaZach, line 26: Would these be the kind of things we'd want to hide behind methods in the top level obj or secondary level obj compared to chaining pointers in such a way? I just worry this could be unsafe since we don't check these pointers

ZeldaZach (approved):

RickyRister:
Is there a reason you used "Generic" instead of "Abstract"?

ZeldaZach (approved):

PR #5649, opened 2025-02-24
Fix banner card changing when opening deck in new tab

RickyRister:

When the "Open deck in new tab by default" setting is enabled and you open a new deck from the Deck Editor Tab, the old deck's `bannerCardComboBox` gets its selection set to the new deck's banner card.

- Remove the code in `openDeckFromFile` that sets the `bannerCardComboBox`
  - The `bannerCardComboBox` already gets updated in `setDeck`


BruebachL:
Looks reasonable.

ZeldaZach (approved):

PR #5651, opened 2025-02-24
Switch current tab when opening a single-instance tab

### LINKED ISSUE #5650: Switch focus to new tab upon opening a tab in the Tabs menu

When a tab in the tabs menu is opened, the focus does not switch to the new tab. This is contrary to how most applications handle opening tabs. 

Cockatrice should switch to the new tab upon opening a tab in the Tabs menu.

RickyRister:
## Related Ticket(s)
- Fixes #5650

Cockatrice currently does not switch the current tab to the new tab when opening a new single-instance tab. 

Opening a single-instance managed tab will now switch the current tab to the new tab. Note that this only counts tabs that are directly opened by the user. Tabs getting opened on startup or on connect still won't cause current tab to change

https://github.com/user-attachments/assets/df83a7c5-1c5b-467e-bb2a-49105c51baf1

- Add `isReopeningTabs` field to `TabSupervisor` to track when tabs are being re-opened from settings so that those tabs don't get auto-focused.
  - Probably not best practice to rely on a mutable field to keep track of stuff like this, but the alternative is a lot messier
    - There shouldn't be any race conditions because I'm pretty sure the tab reopening code is all ran on the same thread.
- add `setCurrent` param to `myAddTab`
  - the managed tabs now call `myAddTab` with `setCurrent` = `!isReopeningTabs`
  - modify existing tab opens that already set current to use that param

### DIFF
--- a/cockatrice/src/client/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/client/tabs/tab_supervisor.cpp
@@ -285,9 +285,15 @@ void TabSupervisor::initStartupTabs()
 {
     addDeckEditorTab(nullptr);
 
-    checkAndTrigger(aTabVisualDeckStorage, SettingsCache::instance().getTabVisualDeckStorageOpen());
-    checkAndTrigger(aTabDeckStorage, SettingsCache::instance().getTabDeckStorageOpen());
-    checkAndTrigger(aTabReplays, SettingsCache::instance().getTabReplaysOpen());
+    if (SettingsCache::instance().getTabVisualDeckStorageOpen()) {
+        openTabVisualDeckStorage(false);
+    }
+    if (SettingsCache::instance().getTabDeckStorageOpen()) {
+        openTabDeckStorage(false);
+    }
+    if (SettingsCache::instance().getTabReplaysOpen()) {
+        openTabReplays(false);
+    }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/tabs/tab_supervisor.cpp
```diff
@@ -285,20 +285,25 @@ void TabSupervisor::initStartupTabs()
 {
     addDeckEditorTab(nullptr);
 
+    isReopeningTabs = true;
```
ZeldaZach, line 288: I'm missing why we're setting this to true and false in seemingly random places. Can you explain further what's happening?

RickyRister: We set it to true right before re-opening the tabs that have their open state persisted in the settings. This includes when we first start cockatrice (where all non-server tabs that have open state are re-opened), as well as when we connect to the server (where all server-exclusive tabs that have open state are re-opened)

RickyRister:
Decided that using a class field like that is bad enough practice that we probably shouldn't be doing it like that. Ended up refactoring `TabSupervisor` so that we can pass `setCurrent` as a param into the tab-opening methods

ZeldaZach (approved):

PR #5652, opened 2025-02-24
Expand/shrink card view window on double click

RickyRister:
## Related Ticket(s)
- Re-opens #5260

Expanding the card view window is really finicky to do. The collision in the corner to start a drag is tiny, and the window sometimes glitches out all over the place as you try to expand it.

Double clicking the top of the card view window will now expand/unexpand it. The collision area for the double-click is the same as the collision for dragging the window around.

The window will get expanded up to its expanded max height. This value is configurable in the settings. The settings dlg guarantees that the initial max height <= expanded max height.

https://github.com/user-attachments/assets/08137699-78e8-4ca2-a91e-20fe403424e1

- Introduce `Maximum expanded height for card view window` setting
  - Defaults to 14 rows
  - Settings dlg forces max initial height <= max expanded height
- Refactor out row to height calculation
- Add logic for determining resizing on double click:
  - max initial height < current < max expanded height -> expand to max expanded height
  - current < max initial height -> expand to max initial height
  - current > max expanded height -> shrink to max expanded height
  - current = max expanded height -> shrink to max initial height
  - current =  actual max height and actual max height < expanded max height -> shrink to max initial height

## Screenshots

Screenshot 2025-02-24 at 7 53 08 AM

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -404,6 +404,11 @@ AppearanceSettingsPage::AppearanceSettingsPage()
     connect(&cardViewInitialRowsMaxBox, qOverload<int>(&QSpinBox::valueChanged), &SettingsCache::instance(),
             &SettingsCache::setCardViewInitialRowsMax);
 
+    cardViewExpandedRowsMaxBox.setRange(1, 999);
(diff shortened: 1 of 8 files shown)


ebbit1q:
this seems fine, but the annoying part where the window jumps when clicking on it (I remember this being a scaling issue) causes me to fail to be able to double click it when further away from the left side of the screen, I'm going to have to look into this further when I can.

ZeldaZach (approved):

PR #5662, opened 2025-02-26
Fix VDS filters not applying after refresh 

### LINKED ISSUE #5661: Visual Deck Storage does not apply filters after refresh
**System Information:**

Client Version: 2.10.1-beta (2025-02-11)
Client Operating System: macOS Sequoia (15.1)
Build Architecture: ARM
Qt Version: 6.7.3
System Locale: en_CA
Install Mode: Standard

When the visual deck storage is refreshed, either by clicking the refresh button, or by enabling/disabling show folders, any currently active filters are no longer applied. 

The filters will only start reapplying after you update the filters.

https://github.com/user-attachments/assets/630c9f39-f9fe-4ddc-906d-8e2aef3869da

**Steps to reproduce:**
 - Open Visual Deck Storage tab
 - Select some sort of filter
 - Click the refresh button
   - Filter is no longer applied
 - Update the filter
   - Filter applies again

RickyRister:
## Related Ticket(s)
- Fixes #5661

Filters become inactive after a refresh or after enabling/disabling show folders. We never reapply the filters after refreshing

https://github.com/user-attachments/assets/5ff48d74-b08c-4b89-b6e7-2543579d64dd

- Call all sort and filter update methods after each refresh
- Got rid of a condition check in `DeckPreviewWidget::updateVisibility`. Otherwise disabling `show folders` would still cause all decks to show regardless of filters. 
  - Deleted signal since nothing was using it anyways.

I assume the reason the condition check in `DeckPreviewWidget::updateVisibility` causes problems is because `isVisible` is also contingent on the parent's visibility, and visibility doesn't update until next frame, so the following sequence happens:
  - parent folder widget is hidden, so deck widget never properly gets set to hidden state
  - folder structure flattening happens
  - try to update visibility on all widgets
  - deck widget is still considered invisible due to previous parent, so it never gets set to hidden despite not currently having hidden state
  -  graphical updates happen, and now deck widget is parented by a visible folder widget and has non-hidden state itself, so it ends up being visible

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
@@ -100,10 +100,7 @@ void DeckPreviewWidget::initializeUi(const bool deckLoadSuccess)
 
 void DeckPreviewWidget::updateVisibility()
 {
-    if (isVisible() != checkVisibility()) {
-        setHidden(!checkVisibility());
-        emit visibilityUpdated();
-    }
+    setHidden(!checkVisibility());
 }
 
 bool DeckPreviewWidget::checkVisibility() const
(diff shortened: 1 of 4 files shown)


BruebachL:
Looks good but please don't just delete signals because "nothing uses them" on features that aren't 100% complete.

RickyRister:
*nothing was using it and it would require me to put the logic back in to make it still work

ZeldaZach (approved):

PR #5664, opened 2025-02-27
Reduce redundant recursion in VDS

RickyRister:

Visual Deck Storage has some code that uses unnecessary recursion and redundantly calls a method on the same object multiple times:

- `flattenFolderStructure` recursively calls all its subfolders' `flattenFolderStructure`, even though `findChildren` already recursively finds all nested subfolders.
- Each folder widget connects the filter updated signals to their `updateVisibility` method, but `updateVisibility` already recursively calls `updateVisibility` on each subfolder.

Although refreshing VDS / updating filters / unchecking show folders is still basically instant for me, I could imagine this inefficiency having an impact for someone on a weaker computer and with more decks.

- Remove recursion from `flattenFolderStructure`
  - `findChildren` is recursive by default and will already recursively find all child `DeckPreviewWidgets`
- Remove connecting the filter updated signals to `updateVisibility` inside of each folder widget
  - instead just trigger the update by calling `updateVisibility` on the root folder widget
- Add `recursive` param to `updateVisibility` 
  - so that when the top-level folder widget recursively gets all subfolders with findChildren and calls `updateVisibility` on them, it doesn't cause more redundant recursive calls
- Delete now-unused filter updated signals

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_widget.h
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_widget.h
@@ -46,9 +46,6 @@ signals:
     void bannerCardsRefreshed();
     void deckLoadRequested(const QString &filePath);
     void openDeckEditor(const DeckLoader *deck);
-    void tagFilterUpdated();
-    void colorFilterUpdated();
-    void searchFilterUpdated();
 
 private:
     QVBoxLayout *layout;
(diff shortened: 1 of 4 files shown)


cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_widget.h
```diff
@@ -46,9 +46,6 @@ public slots:
     void bannerCardsRefreshed();
     void deckLoadRequested(const QString &filePath);
     void openDeckEditor(const DeckLoader *deck);
-    void tagFilterUpdated();
-    void colorFilterUpdated();
-    void searchFilterUpdated();
```
RickyRister, line 51: Deleted the signals since they were now unused. Tell me if I should still keep them around.

ZeldaZach (approved):

PR #5672, opened 2025-03-01
Fix move to top of library shuffling an extra card

### LINKED ISSUE #5611: Bug on "put selected cards on top of librabry at random"
**System Information:**
Client Operating System: Windows 7 Version 6.1 (Build 7601: SP 1)
Build Architecture: 64-bit
Qt Version: 5.15.2
System Locale: fr_FR
Install Mode: Standard

put selected cards on top of librabry at random run with face-up cards, but if one or more of those cards are face-down the shuffle on top take 1 cards more.

Other detail : ctrl+B is used for 2 functions, put on top of the library at random, put on bottom of the library at random.

**Steps to reproduce:**
- Put triumph of saint katherine in play face up.
- put two other cards face down in play from library.
- select the 3 cards.
- right clic -> move to -> top of library at random.
- when you reveal again the cards face up, a 4th card is shuffled with the 3 others.

RickyRister:
## Related Ticket(s)
- Fixes #5611

"Move to top of library in random order" shuffles one card more than the number that was topped. 

Seems like shuffle command's start and end index are both inclusive. We forgot a -1 on the end index.

### DIFF
--- a/cockatrice/src/game/player/player.cpp
+++ b/cockatrice/src/game/player/player.cpp
@@ -3186,7 +3186,7 @@ void Player::cardMenuAction()
                     auto *scmd = new Command_Shuffle;
                     scmd->set_zone_name("deck");
                     scmd->set_start(0);
-                    scmd->set_end(idList.card_size() - 1);
+                    scmd->set_end(idList.card_size() - 1); // inclusive, the indexed card at end will be shuffled
                     // Server process events backwards, so...
                     commandList.append(scmd);
                 }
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/player/player.cpp
```diff
@@ -3186,7 +3186,7 @@ void Player::cardMenuAction()
                     auto *scmd = new Command_Shuffle;
                     scmd->set_zone_name("deck");
                     scmd->set_start(0);
-                    scmd->set_end(idList.card_size());
+                    scmd->set_end(idList.card_size() - 1);
```
ebbit1q, line 3189:
```suggestion
                    scmd->set_end(idList.card_size() - 1); // inclusive, the indexed card at end will be shuffled
```

ebbit1q (approved):
thank you

PR #5673, opened 2025-03-01
Fallback to regular card name search in case the providerId one fails…

BruebachL:
… for the database display widget.

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


RickyRister:
Fixes bug in deck editor tab where the card display widget will display empty info for cards from `cards.xml`s that don't have providerIds

ZeldaZach (approved):

PR #5681, opened 2025-03-03
Add action to Edit Deck via Clipboard

### LINKED ISSUE #5674: Ability to Edit Deck via Clipboard
**Similar Requests**

It would be nice to be able to edit your current deck via clip board and not have to make a new deck.

**Additional Context**

RickyRister:
## Related Ticket(s)
- Fixes #5674

It would be nice to be able to edit your current deck via clip board and not have to make a new deck.

Added a new "Edit deck in clipboard" action to the deck editor tab. The action will open up the same dialog window that "Load deck from clipboard" opens, except it's preloaded with the contents of the current deck. 

Clicking the okay button will load the text in the window into the current deck instead of trying to open a new deck.

Clicking the "refresh" button will reload the text in the window to what it was initially when opened.  

The actions are in a nested menu, and have a "Annotations" and "No Annotations" option. This only changes whether the initial load has annotations.

https://github.com/user-attachments/assets/fabfaf34-a877-469b-9e75-82a2849cdc6b

- Added new menu and actions to `AbstractTabDeckEditor` and `TabDeckEditor`.
- Added shortcuts for the two new actions
- Refactored `DlgLoadDeckFromClipboard` into an abstract class with two subclasses

## Screenshots

Screenshot 2025-03-02 at 7 30 12 PM

Screenshot 2025-03-02 at 7 30 03 PM


ZeldaZach:
How does this work with printing selection?

RickyRister:
> How does this work with printing selection?

Same way that “Load from clipboard” works.

ZeldaZach (approved):

PR #5692, opened 2025-03-05
add "Set Banner Card" action to VDS right-click menu

RickyRister:
## Related Ticket(s)
- Relates to #5618

I usually keep the banner card selector hidden because it causes a large amount of lag. However, that means whenever I want to change the banner card, I need to reshow the selector, and then hide it afterwards.

Also, "Set Banner Card" is the only action we haven't added to the right click menu.

Added the "Set Banner Card" action to the VDS right click menu. The submenu contains the same text that is present in the banner selector combo box. The current banner card is marked with a checkmark. Clicking any of the options will correctly update the banner card, the same as if it was done through the combo box.

https://github.com/user-attachments/assets/a299524d-1638-45ba-b4c9-1b15be7fd668

## Screenshots

Screenshot 2025-03-05 at 12 26 20 AM


BruebachL:
Sure, lgtm.

PR #5705, opened 2025-03-10
don't update the deckhash on load

ebbit1q:
add a new property to decklists that tracks if the deck needs to update the deckhash

only update the deckhash if needed when it is asked for

this saves computing the deckhash for every loaded deck as performed by the visual deck selector with impunity

the copy constructor also now no longer computes the deck hash again

the deck hash is now only computed once, on deck load, as expected, resulting in massive amounts of cpu cycles not being wasted


ebbit1q:
I have noticed a problem immediately, this seems to break the asynchronous deck loading

Issue #2202, opened 2016-10-04
Change game time from local to game


poixen:
Currently the in-game time stamps are like so: [09:59:30] 
However, unless you have a clock handy, you are unable to know how long an action took based on this time.
We _do_ however have an in-game time which counts up in seconds.

I think it would be better to have the in game time displayed to the user, as they can better relate to the time taken.

Hope that makes sence

Nightfirecat:
Seems like a reasonable client option. :+1:

ZeldaZach:
I'd be fine with this being an option the user can choose.

Settings -> Game Time ("Local time" or "Game time")

Daenyth:
I'd rather just change it if we do it at all instead of adding an option

Jinxstands4Jinx:
yea may be a thing about oo many buttons, and it may confuse people. either all in change or not.

tooomm:
Local time could be logged/displayed at room creation, game start and replay start to give a hint when you played that actual game. This helps with finding a specific game.

In general I think the game time might be more helpful.

ZeldaZach:
This seems like a fine client-only change. Would still prefer to have a setting to indicate whether to use game time or local time.

PR #5709, opened 2025-03-11
local time part 1

lilyhuang-github:
## Related Ticket(s)
- Starts to work on #2202

Calculating how much time has passed between actions with local time is hard, and it would be easier with game time.

-Will add helper functions to message_log_widget and chat_view to output time in game time.
- Change MessageLogWidget constructor to accept int pointer to track game time.

Doesn't functionally change anything yet and I'm trying to work on the settings component so the user can choose between game time and local time.


lilyhuang-github:
I'm trying to uncommit the json changes that keep happening

PR #5710, opened 2025-03-12
Change prebuild.js to allow i18n-default.json indentation to exist

lilyhuang-github:
## Related Ticket(s)
-#4803

Husky has a precommit that runs whenever a commit is made causing it to run the prebuild and add i18n-default.json to the commit. This leads to it removing the formatting needing it to be changed backed manually.

- Makes the output of the prebuild come with indentation. So it shouldn't need to be changed manually.


lilyhuang-github:
This really annoyed me while working on the other PR but this should fix it so it doesn't need to be reformatted each time again.

ZeldaZach (approved):

PR #5717, opened 2025-03-13
Add option to choose local time or in game time

### LINKED ISSUE #2202: Change game time from local to game

lilyhuang-github:
## Related Ticket(s)
- Fixes #2202 

Calculating the time between actions with local time can be hard so the option to have actions measured with game time could be useful.

- Adds checkbox to enable in-game time.
- Create new function to get local-time/in-game time depending on setting chosen.
- Change MessageLogWidget to include argument for seconds

## Screenshots
image
image


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -1305,6 +1309,7 @@ void MessagesSettingsPage::retranslateUi()
     aAdd->setText(tr("Add New Message"));
     aEdit->setText(tr("Edit Message"));
     aRemove->setText(tr("Remove Message"));
+    localTimeCheckBox.setText(tr("Enable In-Game Time"));
```
ZeldaZach, line 1312: I'd probably word this as "Use Room Time over Real Time" or something of the sorts

lilyhuang-github: ill change that

lilyhuang-github:
This should pass the cicd now

ebbit1q (changes requested):
instead of a pointer to the secondsElapsed member of the game tab I'd prefer if the game tab had a getter function added for accessing this, the chat view already has a reference to the game tab.

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -1290,6 +1294,7 @@ void MessagesSettingsPage::retranslateUi()
     aAdd->setText(tr("Add New Message"));
     aEdit->setText(tr("Edit Message"));
     aRemove->setText(tr("Remove Message"));
+    localTimeCheckBox.setText(tr("Use Room Time over Real-Life Time"));
```
ebbit1q, line 1297: I think a more user friendly approach would be to present this as a two option spinbox to users, eg:
```
Chat time stamps use: ->Local clock
                      - Match timer
```

cockatrice/src/server/chat_view/chat_view.cpp
```diff
@@ -457,6 +457,23 @@ bool ChatView::isModeratorSendingGlobal(QFlags<ServerInfo_User::UserLevelFlag> u
...
 
+QString ChatView::getCurrentTime()
+{
+    if (showInGameTime) {
+        int seconds = *elapsedSeconds;
+        int minutes = seconds / 60;
+        seconds -= minutes * 60;
+        int hours = minutes / 60;
+        minutes -= hours * 60;
+        return QString("[%1:%2:%3] ")
```
ebbit1q, line 468: I think it might be easier to just use what qt provides here:
```suggestion
        return QTime::fromMSecsSinceStartOfDay(*elapsedSeconds * 1000).toString("[hh:mm:ss] ")
```
it's probably more performant to write all of this out ourselves, but using library code makes it easier to read and makes both code paths very similar, probably giving us an option to simplify this further.

ebbit1q: just noticed this is a copy from TabGame::incrementGameTime, itself some 14 year old code, maybe using qtime there and then having the resulting string be accessible in a getter for the chat view would be an option?

lilyhuang-github:
I changed it so that TabGame::incrementGameTime does it with QTime, and made the other classes use that function. There's an a read access violation for incremnetGameTime that i'm trying to figure out and then I'm gonna try to figure out making it into a combo box.

PR #5724, opened 2025-03-15
Remove `isView` flag from CardZone

Elarnon:
This flag is used for two purposes:

 1. It is used as a check for casting to a zone to a `ZoneViewZone`;

 2. Non-view zones are added to the player's zones on construction

This patch removes the `isView` flag and instead:

 1. We directly cast zones to `ZoneViewZone` using a dynamic (qobject) cast and use the result of the cast instead of the `isView` flag to detect if we are a view zone or not;

 2. The player records its own zones when they are created, simplifying control flow.


Elarnon:
(Going to make this one as part of a larger change)

PR #5728, opened 2025-03-15
Remove `isView` flag from CardZone

Elarnon:
This flag is used for two purposes:

 1. It is used as a check for casting to a zone to a `ZoneViewZone`;

 2. Non-view zones are added to the player's zones on construction

This patch removes the `isView` flag and instead:

 1. We directly cast zones to `ZoneViewZone` using a dynamic (qobject) cast and use the result of the cast instead of the `isView` flag to detect if we are a view zone or not;

 2. The player records its own zones when they are created, simplifying control flow.

### DIFF
--- a/cockatrice/src/game/player/player.h
+++ b/cockatrice/src/game/player/player.h
@@ -409,7 +409,13 @@ public:
     void playCardToTable(const CardItem *c, bool faceDown);
     void addCard(CardItem *c);
     void deleteCard(CardItem *c);
-    void addZone(CardZone *z);
+
+    template <typename T, typename... Args> T *addZone(Args &&...args)
+    {
+        T *zone = new T(std::forward<Args>(args)...);
+        zones.insert(zone->getName(), zone);
+        return zone;
+    }
 
     AbstractCounter *addCounter(const ServerInfo_Counter &counter);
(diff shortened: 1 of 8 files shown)


cockatrice/src/game/player/player.h
```diff
@@ -407,7 +407,13 @@ private slots:
     void playCardToTable(const CardItem *c, bool faceDown);
     void addCard(CardItem *c);
     void deleteCard(CardItem *c);
-    void addZone(CardZone *z);
+
+    template <typename T, typename... Args> T *addZone(Args &&...args)
```
RickyRister, line 411: why templates? Does modifying the existing `addZone` to return this and then doing `PileZone *zone = addZone(new PileZone(...))` not work?

Elarnon: You mean `addZone` would do `return zone` I assume? Without templates the return type would have to always be `CardZone *` but sometimes we want a `PileZone *`, sometimes a `TableZone *`, sometimes a `HandZone *` and sometimes a `StackZone *`.

I can simplify the template to `template <typename T> T* addZone(T* zone)` though, I don't remember why I used variadics here (this is old code I'm only upstreaming now).

ZeldaZach (approved):

Elarnon:
@ZeldaZach rebased

Issue #4974, opened 2024-01-01
Prevent cards jumping around while dragging


Elarnon:
#4963 improves the situation and highlights the final location when available cards (stack of 3) prevent any further card to be placed there and the new card will be pushed away to the next available slot. I still think "jumping cards" and they not ending up at the current cursor position is not intuitiv.
Two ideas here - as I really like the intuitive "drop at mouse placement" principle:
  - If a place is fully occupied, do not allow to drop cards there (change the mouse cursor or have any other visual indication - maybe even a helper text - that this is not a possible location).
  - Do allow to drop it still, but move all other cards by one position. The player obviously wants to have the card in this exact place... The other cards could all show their new location as phantom cards. But not sure what the side effects are.

_Originally posted by @tooomm in https://github.com/Cockatrice/Cockatrice/issues/4963#issuecomment-1859283840_

The PR does not allow drops anymore. It would be nice to explore the 2nd option, or other alternatives.

I ended up having some time to play around over the holidays which I will try to upstream a variant of (probably in several small PRs):

[Screencast from 2024-01-01 13-25-22.webm](https://user-images.githubusercontent.com/146210/293585547-498651c1-2bc6-4b4f-ad84-725f75b72484.webm)

There are actually some unpleasant interactions between pushing cards and zoom (because cards may now overflow...) which I don't really know how to solve yet.

tooomm:
The video looks very promising. I like the position highlighting a lot, that's great!
Way more intuitive than what we currently do.

One thing I notice in your recording:
The card stack should not be pushed away further to the right at ~0:13 though (from position 3 to 4)?
But instead jump back to its initial start placement (position 2) as the card is now hovering "position 3" on the grid which was empty in the beginning.

Elarnon:
> The video looks very promising. I like the position highlighting a lot, that's great! Way more intuitive than what we currently do.
> 
> One thing I notice in your recording: The card stack should not be pushed away further to the right at ~0:13 though (from position 3 to 4)? But instead jump back to its initial start placement (position 2) as the card is now hovering "position 3" on the grid which was empty in the beginning. But if that's tricky to figure out, I'm totally fine with what we see in the clip.

That is because the card stack is too big; if we did that, the stack would move back to its initial position but the cursor would still be over the stack which we wanted to avoid. But I agree, it ends up causing somewhat weird behavior (and it actually complicates the math quite a bit, so I'd be happy to remove that). 
 
[...]

PR #5730, opened 2025-03-15
Use enum for ThemeManager brushes

Elarnon:
This patch introduces an enum to distinguish the different brushes that can be set by the theme (hand, stack, etc.) and generic functions taking the enum rather than having one copy of each function for each brush.

This is preliminary work before merging StackZone and HandZone to simplify #4974.

### DIFF
--- a/cockatrice/src/client/ui/theme_manager.h
+++ b/cockatrice/src/client/ui/theme_manager.h
@@ -9,6 +9,8 @@
 #include <QPixmap>
 #include <QString>
 
+#include <array>
+
 inline Q_LOGGING_CATEGORY(ThemeManagerLog, "theme_manager");
 
 typedef QMap<QString, QString> QStringMap;
--- a/cockatrice/src/client/ui/theme_manager.h
+++ b/cockatrice/src/client/ui/theme_manager.h
@@ -22,13 +22,23 @@ class ThemeManager : public QObject
 public:
     ThemeManager(QObject *parent = nullptr);
 
+    enum Role
+    {
+        MinRole = 0,
+        Hand = MinRole,
+        Stack,
+        Table,
+        Player,
+        MaxRole = Player,
+    };
+
 private:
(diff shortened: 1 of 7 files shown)


cockatrice/src/client/ui/theme_manager.h
```diff
@@ -22,41 +23,34 @@ class ThemeManager : public QObject
...
+        Hand = MinRole,
+        Stack,
+        Table,
+        Player,
+        MaxRole = Player,
+    };
+
 private:
-    QBrush handBgBrush, stackBgBrush, tableBgBrush, playerBgBrush;
+    std::array<QBrush, Role::MaxRole + 1> brushes;
```
RickyRister, line 37: Shouldn’t we use qt’s containers instead of the std containers?

Elarnon: `std::array` is just a thin wrapper around C arrays (i.e. this line has the same memory layout as `QBrush brushes[Role::MaxRole + 1];`) with less wonky semantics; as far as I know Qt doesn't provide an equivalent.

BruebachL: Qt offers the [QVarLengthArray](https://doc.qt.io/qt-6/qvarlengtharray.html)<T, Prealloc> though that is dynamically sized.

Elarnon: Ah, that's true. I think that the semantics are a bit worse for this use case — `QVarLengthArray` is bigger internally and the size needs to be provided twice (capacity + actual size). If using `std::array` is an issue I'll just revert back to a C array.

BruebachL: I agree with you that QVarLengthArray is unsuitable here, I just wanted to point out the possibility so we can say we've explored it.

ZeldaZach (approved):

PR #5735, opened 2025-03-16
Close the `TabGame`s when closing the `TabSupervisor`

### LINKED ISSUE #5697: Window Position Not Returning To Previous Positions Upon Start
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_US
Install Mode: Portable

After updating to the newest release, the Card Info and Messages windows do not open automatically. Usually, they default to their last known location, but this behavior does not sem to be happening now.

Normal behavior:

1. Open Cockatrice
2. Open Card Info and Messages windows
3. Position them where desired
4. Close Cockatrice
5. Open Cockatrice
6. Windows should be in their previous positions and visibility set to on

Current behavior

1. Open Cockatrice
2. Open Card Info and Messages windows
3. Position them where desired
4. Close Cockatrice
5. Open Cockatrice
6. Windows are not in their previous position or visibility

Elarnon:
This ensures that we go through the same code path (in terms of Qt events) when closing the whole supervisor as when closing a single tab. Also, use the `close` event instead of the `hide` event to detect when we are closing a game.

Fixes #5697

### DIFF
--- a/cockatrice/src/client/tabs/tab_game.cpp
+++ b/cockatrice/src/client/tabs/tab_game.cpp
@@ -1751,10 +1751,8 @@ void TabGame::createMessageDock(bool bReplay)
     connect(messageLayoutDock, SIGNAL(topLevelChanged(bool)), this, SLOT(dockTopLevelChanged(bool)));
 }
 
-void TabGame::closeEvent(QCloseEvent *event)
+void TabGame::hideEvent(QHideEvent *event)
 {
-    Q_UNUSED(event);
-
     LayoutsSettings &layouts = SettingsCache::instance().layouts();
     if (replay) {
         layouts.setReplayPlayAreaState(saveState());
@@ -1770,6 +1768,8 @@ void TabGame::closeEvent(QCloseEvent *event)
         layouts.setGameMessageLayoutSize(messageLayoutDock->size());
         layouts.setGamePlayerListSize(playerListDock->size());
     }
+
+    Tab::hideEvent(event);
 }
 
 // Method uses to sync docks state with menu items state
--- a/cockatrice/src/client/tabs/tab_game.cpp
+++ b/cockatrice/src/client/tabs/tab_game.cpp
@@ -119,7 +119,6 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor, GameReplay *_replay)
     refreshShortcuts();
     messageLog->logReplayStarted(gameInfo.game_id());
 
-    this->installEventFilter(this);
     QTimer::singleShot(0, this, SLOT(loadLayout()));
 }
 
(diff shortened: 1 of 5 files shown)


cockatrice/src/client/tabs/tab_game.cpp
```diff
@@ -1772,23 +1791,6 @@ bool TabGame::eventFilter(QObject *o, QEvent *e)
         }
     }
 
-    if (o == this && e->type() == QEvent::Hide) {
```
RickyRister, line 1775: Note that changing the layout saving to trigger on a close event instead of a hide event *will* change existing behavior. 

Hide events will trigger when when the user switches to a different tab. That means if you open a game, modifying the layout, then open another game *without* closing the first game first, the new game will still have the updated layout. Changing the layout saving to only trigger on close events will break that behavior.

cockatrice/src/client/tabs/tab_game.cpp
```diff
@@ -1753,6 +1751,27 @@ void TabGame::createMessageDock(bool bReplay)
...
+        layouts.setReplayMessageLayoutSize(messageLayoutDock->size());
+        layouts.setReplayPlayerListSize(playerListDock->size());
+        layouts.setReplayReplaySize(replayDock->size());
+    } else {
+        layouts.setGamePlayAreaState(saveState());
+        layouts.setGamePlayAreaGeometry(saveGeometry());
+        layouts.setGameCardInfoSize(cardInfoDock->size());
+        layouts.setGameMessageLayoutSize(messageLayoutDock->size());
+        layouts.setGamePlayerListSize(playerListDock->size());
+    }
```
RickyRister, line 1772: `Tab` also has a `closeEvent` override that emits a custom signal. All `Tab` subclasses need to call `Tab::closeEvent` in their `closeEvent` override to make sure that signal is consistently emitted.

https://github.com/Cockatrice/Cockatrice/blob/c99afe7956945f6fcee7a5d782a6ce2dc3457578/cockatrice/src/client/tabs/tab.h#L16-L21

cockatrice/src/client/tabs/tab_game.cpp
Elarnon: Ah, so that's why it was done that way — I can move it to `hideEvent` instead I think, let me check.

cockatrice/src/client/tabs/tab_game.cpp
Elarnon: Isn't this already provided by the `tabCloseRequested` signal on the `QTabWidget`? I did forget to call the parent `closeEvent` though, good catch.

RickyRister: I didn't know `tabCloseRequested` existed when I wrote this code. That is a refactor we could explore doing later.

Elarnon: Yeah, the `TabSupervisor` is a bit of a mess anyway. Easy to break subtle things there (as #5697 shows) though, so cleaning it up probably would need extensive testing.

cockatrice/src/client/tabs/tab_game.cpp
Elarnon: Moved to use `hideEvent` instead of `closeEvent` and it still seems to work properly on my machine.

ZeldaZach (approved):

Issue #5688, opened 2025-03-04
Edit Existing/Remove Existing Deck Tags


issa-puppi:
**Similar Requests**

As the name implys, simply be able to edit existing deck tags and remove old ones we don't like or don't use

**Additional Context**

ebbit1q:
fixed in #5979

PR #5736, opened 2025-03-16
Add a new dialog that allows editing the default suggested tags.

BruebachL:
## Related Ticket(s)
- Implements #5688

Users have no way of editing the default suggested tags we provide.

- Adds a new dialog that does just that
- Also persists the default tags in the settings cache now

## Screenshots
image
image
image
image

### DIFF
--- a/cockatrice/src/dialogs/dlg_default_tags_editor.cpp
+++ b/cockatrice/src/dialogs/dlg_default_tags_editor.cpp
@@ -1,0 +1,155 @@
+#include "dlg_default_tags_editor.h"
+
+#include "../settings/cache_settings.h"
+
+#include <QHBoxLayout>
+#include <QLineEdit>
+#include <QListWidget>
+#include <QListWidgetItem>
+#include <QMessageBox>
+#include <QPushButton>
+#include <QVBoxLayout>
+#include <QWidget>
+
+DlgDefaultTagsEditor::DlgDefaultTagsEditor(QWidget *parent) : QDialog(parent)
+{
+    setWindowModality(Qt::WindowModal); // Steals focus from parent dialog
+
+    QVBoxLayout *mainLayout = new QVBoxLayout(this);
+
+    // Input field + Add button (horizontal layout)
+    QHBoxLayout *inputLayout = new QHBoxLayout();
+    inputField = new QLineEdit(this);
+    addButton = new QPushButton(this);
+    inputLayout->addWidget(inputField);
+    inputLayout->addWidget(addButton);
+    mainLayout->addLayout(inputLayout);
+
+    // List widget for tags
+    listWidget = new QListWidget(this);
+    listWidget->setSelectionMode(QAbstractItemView::SingleSelection);
+    listWidget->setEditTriggers(QAbstractItemView::DoubleClicked | QAbstractItemView::EditKeyPressed);
+    mainLayout->addWidget(listWidget);
+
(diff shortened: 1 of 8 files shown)


ZeldaZach (approved):

cockatrice/src/settings/cache_settings.h
```diff
@@ -133,6 +214,7 @@ class SettingsCache : public QObject
     bool visualDeckStorageShowBannerCardComboBox;
     bool visualDeckStorageShowTagsOnDeckPreviews;
     bool visualDeckStorageShowTagFilter;
+    QStringList visualDeckStorageDefaultTagsList;
```
RickyRister, line 217: I don't think the tags should be stored in the main `settings.ini`. Users might want to be able to share the default tags around

cockatrice/src/settings/cache_settings.cpp
```diff
@@ -265,6 +265,8 @@ SettingsCache::SettingsCache()
     visualDeckStorageSortingOrder = settings->value("interface/visualdeckstoragesortingorder", 0).toInt();
     visualDeckStorageShowFolders = settings->value("interface/visualdeckstorageshowfolders", true).toBool();
     visualDeckStorageShowTagFilter = settings->value("interface/visualdeckstorageshowtagfilter", true).toBool();
+    visualDeckStorageDefaultTagsList =
+        settings->value("interface/visualdeckstoragedefaulttagslist", defaultTags).toStringList();
```
RickyRister, line 269: I don't think the tags should be stored in the main settings.ini. Users might want to be able to share the default tags around

ZeldaZach: You bring up a good point, people might want to share default tags. Could make a tags.ini file?

RickyRister:
The new feature is broken for me.

https://github.com/user-attachments/assets/35927471-3a0d-4b80-afc6-bae6d2eeb91d

The edit default tags window isn't automatically brought to front on creation. But more pressingly, the main tag edit window uses `exec`, which prevents me from interacting with any other windows, including the edit default tags window.

cockatrice/src/dialogs/dlg_default_tags_editor.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        // Delete button
+        QPushButton *deleteButton = new QPushButton(tr("✖"), this);
+        deleteButton->setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
+        deleteButton->setFixedWidth(25);
+        layout->addSpacing(5); // Adds spacing before the delete button
+        layout->addWidget(deleteButton);
+
+        widget->setLayout(layout);
+        listWidget->setItemWidget(item, widget);
```
RickyRister, line 79: The widgets in the rows have the incorrect height on my end.
Screenshot 2025-03-17 at 8 21 42 PM

According to the Qt docs for [QListWidget::setItemWidget](https://doc.qt.io/qt-6/qlistwidget.html#setItemWidget),
> This function should only be used to display static content in the place of a list widget item. If you want to display custom dynamic content or implement a custom editor widget, use [QListView](https://doc.qt.io/qt-6/qlistview.html) and subclass [QStyledItemDelegate](https://doc.qt.io/qt-6/qstyleditemdelegate.html) instead.

See if you can fix it as-is first. But it might entirely be the case that we need to use a custom `QStyledItemDelegate` to fix this

PR #5740, opened 2025-03-17
Disconnect setMenu on close

RickyRister:
## Related Ticket(s)
- Fixes bug introduced in #5735

Cockatrice segfaults when you close it while having a game tab as your current tab. 

The cause is that the close sequence eventually causes `QTabWidget::currentChanged` to be emitted which then triggers `TabSupervisor::updateCurrent` which then emits `TabSupervisor::setMenu` which then triggers `MainWindow::updateTabMenu` which then tries to do stuff with the tabMenu and menuBar while the app is in the middle of destructing everything which then causes a segfault

Disconnect `TabSupervisor::setMenu` in `TabSupervisor`'s `closeEvent` so that sequence never happens.

### DIFF
--- a/cockatrice/src/client/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/client/tabs/tab_supervisor.cpp
@@ -265,6 +265,9 @@ void TabSupervisor::closeEvent(QCloseEvent *event)
     for (auto tabId : gameTabsToRemove) {
         gameTabs.remove(tabId);
     }
+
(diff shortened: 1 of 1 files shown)


Elarnon:
I believe this fix merely hides the issue, which is that the `MainWindow` keeps `QMenu`s around that are freed when the tab is deleted, due to a bug introduced in #5735 (closing the tab deletes the tab, which deletes the menus that are owned by the tab but still present in the `MainWindow`). I believe #5741 is a more comprehensive fix which (hopefully…) doesn't leave dangling pointers around.

PR #5741, opened 2025-03-17
Avoid double `close` in `TabSupervisor`

### LINKED ISSUE #5697: Window Position Not Returning To Previous Positions Upon Start

Elarnon:
In #5447, the `TabSupervisor` logic was changed to go through `closeRequest` instead of `deleteLater`.

Unfortunately, this introduced a bug where we close every (most) tabs twice due to `closeRequest` calling `close()`, which is a high-level widget function that triggers a whole lot of events, including `QHideEvent`s, which manifests through #5697. #5735 was a tentative fix for the issue, but it mis-diagnosed the problem and does not actually fix it, as shown by #5740.

This patch makes `closeRequest` call `deleteLater` instead, which is a low-level function that doesn't trigger all those events. To make sure slots happening on tab close are still processed, use the `QObject::destroyed` signal instead of the custom `Tab::closed` event.

This also fixes what I believe was the missing piece in #5447: `TabSupervisor`'s destructor calling `TabSupervisor::stop`, which in turn sends out a bunch of signals which is never a good idea while in the middle of destruction. Instead, `TabSupervisor`'s destructor now simply deletes its (non-`QObject`) children.

I suspect that with `TabSupervisor`'s destructor no longer calling `TabSupervisor::stop` the `closeRequest` API can be simplified again; this is left for later as this PR is focused on fixing the crashes.

Supersedes #5735 and #5740.

Fixes #5697 (again).

### DIFF
--- a/cockatrice/src/client/tabs/abstract_tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/abstract_tab_deck_editor.cpp
@@ -554,5 +554,5 @@ void AbstractTabDeckEditor::closeRequest(bool forced)
     }
 
     emit deckEditorClosing(this);
-    close();
+    deleteLater();
 }
--- a/cockatrice/src/client/tabs/tab.cpp
+++ b/cockatrice/src/client/tabs/tab.cpp
@@ -44,16 +44,7 @@ void Tab::deleteCardInfoPopup(const QString &cardName)
     }
 }
 
-/**
- * Overrides the closeEvent in order to emit a close signal
- */
-void Tab::closeEvent(QCloseEvent *event)
-{
-    emit closed();
-    event->accept();
-}
-
 void Tab::closeRequest(bool /*forced*/)
 {
-    close();
+    deleteLater();
 }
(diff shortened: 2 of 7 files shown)


Elarnon:
Note: both parts of the fix (calling `deleteLater` instead of `close` and not calling `TabSupervisor::stop` from `TabSupervisor`'s destructor) are necessary, depending on the way the tab is closed.

 - If the whole app is closed, first the tab is hidden by Qt (triggers the save), then the supervisor is deleted (calls `stop`, which removes the tab, causing it to be hidden a second time and trigger the save again, but in a bogus state).
 - If only the tab is closed, the tab is removed in `gameLeft`, causing it to become hidden (triggers the save), then the tab is closed a second time.

ebbit1q:
did this pr get superseded by #6061 @RickyRister ?

RickyRister:
Yes

PR #4962, opened 2023-12-13
Support fractional scaling when scaling card images

Elarnon:
## Related Ticket(s)
- Fixes #4880

## Short roundup of the initial problem

Card pictures are blurry when using fractional scaling (i.e. non-integer scales, such as 125% or 150% which translates to 1.25 and 1.50 respectively).

## What will change with this Pull Request?

Card pictures are no longer blurry with fractional scaling.

## Screenshots

Screenshots taken with 1.5 scaling.

Current master:
Before

This PR:
After

PR #5742, opened 2025-03-17
chore: Add .git-blame-ignore-revs

Elarnon:
This file can be used to list reformatting commit that we don't want to see in `git blame` output. Currently includes a single commit that switched the source to 4-space tabs that shows up everywhere in old code.

To use, set `git config blame.ignoreRevsFile .git-blame-ignore-revs`.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,12 @@
+# This file lists commits that should be ignored in `git blame` output
+# (typically commits that reformat large amount of code).
+#
+# When adding commits to this file, please include the short (one-liner)
+# description of the commit along with its hash for readability.
+#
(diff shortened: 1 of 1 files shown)


ebbit1q:
I didn't know about this feature, it seems pretty useful.
we don't make a lot of use of it though, we have one of these commits and if you blame something so far back whatever it is it can safely be assumed that whoever made the commit is gone already.

Elarnon:
I don't really care about the author of the commit, I care about commit messages — those usually tell me what the author was trying to do, which helps understand what's going on. I remember it being useful when debugging #4962 for instance.

Anyway, I'm using that file and thought I'd share; if you don't want it no problem I'll keep it in my ignore list.

PR #5744, opened 2025-03-17
hotfix: Remove menus when closing game

Elarnon:
Version of #5740 that doesn't leave freed `QMenu`s lying around.

This is a hotfix in order to be able to cut a point release with minimal changes as @RickyRister would prefer we don't rush #5741.

(Pending tests from @RickyRister to confirm it fixes the issue since I can't reproduce the crash from #5740 on my machine)


RickyRister:
Works on my machine

ZeldaZach (approved):

PR #5759, opened 2025-03-20
Hide arena only cards

Elarnon:
This includes a setting to let people that want them have them. It requires a restart because the card database code is a mess and I don't really have time to fix it right now.

Note: this includes #5758 in order to have the proper translation files and should be merged after #5758.

image


ZeldaZach (approved):

Elarnon:
Rebased and fixed conflicts

PR #5760, opened 2025-03-20
Add option to disable card rounding

Elarnon:
image


ZeldaZach (approved):

cockatrice/src/settings/cache_settings.h
```diff
@@ -82,6 +83,8 @@ class SettingsCache : public QObject
     void downloadSpoilerTimeIndexChanged();
     void downloadSpoilerStatusChanged();
     void useTearOffMenusChanged(bool state);
+    void roundCardCornersChanged(bool roundCardCorners);
+    void cardCornerRadiusChanged(qreal cardCornerRadius);
```
RickyRister, line 87: do we need to have both signals?

cockatrice/src/settings/cache_settings.h
```diff
@@ -727,6 +731,14 @@ class SettingsCache : public QObject
     {
         return mbDownloadSpoilers;
     }
+    bool getRoundCardCorners() const
+    {
+        return roundCardCorners;
+    }
+    qreal getCardCornerRadius() const
+    {
+        return roundCardCorners ? 0.05 : 0.;
```
RickyRister, line 740: Are you sure we want to have this extra method in `SettingsCache`?
I feel like we should keep as much logic out of the settings as possible.

cockatrice/src/settings/cache_settings.h
Elarnon: Not needed, but as long as we have both properties it is more consistent to also have both signals.

cockatrice/src/settings/cache_settings.h
Elarnon: Sorry, I have been sleeping poorly lately and that made me be a bit more aggressive than I needed to be here. Those are good points, and quick to fix — I've included a commit to remove the `cardCornerRadius` property and compute the radius inline.

PR #5762, opened 2025-03-21
fix: Always prefer local cards if available

Elarnon:
This fixes an issue reported on Discord that the set-based picture directories no longer work.

## The Problem

Cockatrice has two ways of loading user-defined pictures from disk:

 - Through the CUSTOM directory, which ignores set information (any `Forest.png` file will be used for a `Forest`)
 - Through set-based directories in the pics folder (e.g. `M20/Forest.png` and `10E/Forest.png`) which get loaded depending on the user-defined set preference order

(There is also the legacy `downloadedPics` folder for the old picture cache, which works like the set-based directories and is kept for backwards compatibility only)

With recent provider-related changes, we only look for pictures on disk for the most preferred set of a card (unless the "override all card art with personal preference" setting is set). This makes sense for the CUSTOM directory, but breaks the set-based directories, as they are effectively not used anymore.

## The Solution

This patch changes the default logic for loading cards to always look for cards of the requested set on disk, but to only look in the CUSTOM folder for the most preferred set of a card. If the "override all card art with personal preference" setting is set, always look in the CUSTOM folder.

Note that this means that for sets that have multiple printings of a card, and the user has a card for that set in their pics directory, the same picture will be used for all printings in the set (which I think is desired behavior).

### Old Behavior

 - With "override all card art with personal preference" disabled (default): load from the CUSTOM directory and set-based directories for the user's most preferred set. Ignore the CUSTOM directory and set-based directories for other sets. Use provider set order.

 - With "override all card art with personal preference" enabled: load from the CUSTOM directory and set-based directories for the user's most preferred set. Ignore the CUSTOM directory and set-based directories for other sets. Use set order from settings.

### New Behavior

 - With "override all card art with personal preference" disabled (default): load from the CUSTOM directory for the user's most preferred set, then always load from set-based directories. Use provider set order.

 - With "override all card art with personal preference" enabled: always use the CUSTOM directory and set-based directory. Use set order from settings.

### DIFF
--- a/cockatrice/src/client/ui/widgets/printing_selector/printing_selector.cpp
+++ b/cockatrice/src/client/ui/widgets/printing_selector/printing_selector.cpp
@@ -36,10 +36,8 @@ PrintingSelector::PrintingSelector(QWidget *parent, AbstractTabDeckEditor *_deck
         QFrame *warningFrame = new QFrame(this);
         warningFrame->setFrameShape(QFrame::StyledPanel);
 
-        QLabel *warningLabel = new QLabel(this);
+        warningLabel = new QLabel(this);
         warningLabel->setTextFormat(Qt::RichText);
-        warningLabel->setText("<b>Warning:</b> You appear to be using custom card art, which has known bugs when also "
-                              "using the printing selector in this version of Cockatrice.");
         warningLabel->setWordWrap(true);
 
         auto *warningLayout = new QVBoxLayout(warningFrame);
@@ -48,6 +46,8 @@ PrintingSelector::PrintingSelector(QWidget *parent, AbstractTabDeckEditor *_deck
         warningLayout->addWidget(warningLabel);
 
         layout->addWidget(warningFrame);
+    } else {
+        warningLabel = nullptr;
     }
 
     widgetLoadingBufferTimer = new QTimer(this);
@@ -104,6 +104,12 @@ PrintingSelector::PrintingSelector(QWidget *parent, AbstractTabDeckEditor *_deck
 void PrintingSelector::retranslateUi()
 {
     navigationCheckBox->setText(tr("Display Navigation Buttons"));
+
+    if (warningLabel) {
+        warningLabel->setText(
+            tr("<b>Warning:</b> You appear to be using custom card art, which has known bugs when also "
+               "using the printing selector in this version of Cockatrice."));
+    }
 }
 
 void PrintingSelector::printingsInDeckChanged()
--- a/cockatrice/src/client/ui/widgets/printing_selector/printing_selector.cpp
+++ b/cockatrice/src/client/ui/widgets/printing_selector/printing_selector.cpp
@@ -1,12 +1,15 @@
 #include "printing_selector.h"
 
 #include "../../../../settings/cache_settings.h"
+#include "../../picture_loader/picture_loader.h"
 #include "printing_selector_card_display_widget.h"
 #include "printing_selector_card_search_widget.h"
 #include "printing_selector_card_selection_widget.h"
 #include "printing_selector_card_sorting_widget.h"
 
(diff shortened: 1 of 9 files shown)


Elarnon:
Note: the behavior in the PR might not be the right one, as users would expect the CUSTOM folder to override other preferences.

It seems to me that this should be controlled by the "override all card art with personal preference" setting, but I assume maybe that's there to not use card preferences from remote decks so I don't want to just repurpose it.

Paging @BruebachL

Elarnon:
Thinking about this more, I think what we want is to have this order of preference:

 - Local CUSTOM pics
 - Local per-set pics
 - Downloaded

and so I'd like to remove the logic related to loading local stuff only for the most preferred set, because that's confusing and not really meaningful from a user's point of view.

The issue with that is that it breaks the printing selector when the user has CUSTOM pics, and that's not easy to fix with the current architecture because we don't know when displaying a print whether we are in the printing selector or not (because the PictureLoader / PixmapCacheKey stuff is very weird and fragile).

I think we should fix the PictureLoader etc. stuff but that's a big endeavor, so in the meantime what would make sense to me is to keep the preference local CUSTOM pics → local per-set pics → downloaded but have a banner on the printing selector saying that it may not work if you have local pics (maybe only if the user actually has local pics).

BruebachL:
I'd rather just clean up the mess altogether and consolidate how we load and handle custom images once and for all. 
That is, a flat folder where images are named card_name_set_name_number or card_name_set_name_number_providerId or similar. Or perhaps even set_name_number_card_name to make sorting a bit easier in a flat folder.

Additionally, I'd like to introduce a special custom set for custom art, rather than having them override existing cards by default. 
Users have requested this and I can see the value in it.

Imo, the set preference selector (and especially the ability the disable sets) is kinda redundant post-providerId change or at least should be repurposed. In my mind, people playing with non-providerId cards should be gently nudged towards updating their decks and new players should explicitly need to set non-providerId cards for their decks if they ever feel the need (though it is quite unintuitive for someone to say "Instead of deciding which card art this will be, I will leave the liberty up to your personal preference." [Though it is of course possible to enforce your personal preference locally with custom pictures]).
Still, I think the old approach of simply blanketing an entire card underneath a custom picture needs to go.

Elarnon:
We can also not change behavior and just have a warning in the printing selector that it may not work properly if custom art is detected, that's maybe easier and would still avoid confusion.

cockatrice/src/client/ui/widgets/printing_selector/printing_selector.cpp
```diff
@@ -29,6 +32,24 @@ PrintingSelector::PrintingSelector(QWidget *parent, AbstractTabDeckEditor *_deck
     layout = new QVBoxLayout(this);
     setLayout(layout);
 
+    if (PictureLoader::hasCustomArt()) {
+        QFrame *warningFrame = new QFrame(this);
+        warningFrame->setFrameShape(QFrame::StyledPanel);
+
+        QLabel *warningLabel = new QLabel(this);
+        warningLabel->setTextFormat(Qt::RichText);
+        warningLabel->setText("<b>Warning:</b> You appear to be using custom card art, which has known bugs when also "
```
RickyRister, line 41: Need to `tr` the text, which means the text needs to be set in `retranslateUi`. Although I don’t think we’ll update the translations in time for the release, and we probably won’t keep this around for the next release, I guess it doesn’t matter

Elarnon: > Although I don’t think we’ll update the translations in time for the release, and we probably won’t keep this around for the next release, I guess it doesn’t matter

That's what I was thinking but I still intended to do this right and forgot.

ZeldaZach (approved):

PR #5764, opened 2025-03-23
Fix card menu not unsetting in local multiplayer

RickyRister:
## Related Ticket(s)
- Fixes issue introduced in #5376

In local multiplayer games, card menus would sometimes fail to be unset when a multi-selection is deselected. 
The order that the itemChanged events get triggered in is inconsistent, so sometimes the `scene()->selectedItems().isEmpty()` check will fail, and not cause the menu to be unset.

With both card menus active at the same time, this causes multiple actions with the same shortcut to be active. Qt will just ignore conflicting shortcuts, thus causing the behavior that shortcuts sometimes stop working in local multiplayer games

- Put the empty selection check on a 1 frame timer. 
  - This ensures that if a multi-selection deslection happens, the empty check will occur after the selection actually gets cleared.
  - Dunno if the introduces the possibility of a use-after-free bug. I tested closing the game while having a selection a few times, and no crashes so far
- Also check the owner of the card when doing the empty check.
  -  Fixes the edge case of selecting one player's cards then directly selecting the other player's cards causing menu to not get unset, since the empty check fails since there is always a selected card during the sequence

### DIFF
--- a/cockatrice/src/game/cards/card_item.cpp
+++ b/cockatrice/src/game/cards/card_item.cpp
@@ -475,6 +475,17 @@ bool CardItem::animationEvent()
     return animationIncomplete;
 }
 
+/**
+ * Checks if any of the items is a CardItem that is owned by the given player.
+ */
+static bool hasCardsOwnedBy(const QList<QGraphicsItem *> &items, const Player *owner)
+{
+    return std::any_of(items.cbegin(), items.cend(), [owner](QGraphicsItem *item) {
+        auto *card = qgraphicsitem_cast<CardItem *>(item);
+        return card && card->getOwner() == owner;
+    });
+}
+
 QVariant CardItem::itemChange(GraphicsItemChange change, const QVariant &value)
 {
     if ((change == ItemSelectedHasChanged) && owner != nullptr) {
@@ -482,9 +493,14 @@ QVariant CardItem::itemChange(GraphicsItemChange change, const QVariant &value)
             owner->setCardMenu(cardMenu);
             owner->getGame()->setActiveCard(this);
         } else if (owner->getCardMenu() == cardMenu) {
-            if (scene() && scene()->selectedItems().isEmpty()) {
-                owner->setCardMenu(nullptr);
-            }
+            // If selection is cleared while multiple cards are selected, there might still be items in the selection
+            // when this itemChange triggers.
+            // Delay empty check by 1 frame to ensure the entire selection is cleared first.
+            QTimer::singleShot(1, this, [this] {
+                if (scene() && !hasCardsOwnedBy(scene()->selectedItems(), owner)) {
+                    owner->setCardMenu(nullptr);
+                }
+            });
             owner->getGame()->setActiveCard(nullptr);
         }
     }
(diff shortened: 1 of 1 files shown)


Elarnon:
Thanks for investigating! 

I think that we actually only need the second part of the fix (checking owner when doing the empty check), as we should be guaranteed to have one "last deselected card" per player independent of the order in which `itemChange` events are processed. This would avoid introducing delays (and the future headaches they imply).

RickyRister:
> I think that we actually only need the second part of the fix (checking owner when doing the empty check), as we should be guaranteed to have one "last deselected card" per player independent of the order in which `itemChange` events are processed. This would avoid introducing delays (and the future headaches they imply).

I tried only doing the second part of the fix. It didn't work. There's definitely something up with the order of the `itemChange` events, but I didn't feel like investigating further. Getting it to "work properly" would probably require some refactoring which class is responsible for handling the menus, which I didn't want to do so close to a release. We should look into it later.

Elarnon:
AH I get it — there's only one last selected card, but it's not guaranteed to be the one that has the cardMenu on the player. That is nasty.

I think this fixes the issue at hand, but I'm wondering if we should just revert #5376 for now and figure out a proper fix later, because I think #5376 has further implications that I don't fully understand. For instance I can confirm the following scenario (even with this fix):

 - Start a multiplayer game on a server with two players
 - Player 1 puts cards onto the battlefield, then creates a token
 - Player 2 selects cards from Player 1's side of the board, including the token which is selected *last*
 - Player 2 deselects the token (using Ctrl-click)
 - Player 1 moves the token to the graveyard, destroying it
 - Other cards still appear selected for Player1 but the "Clone" action is no longer available, because the menu that is being used was the menu from the token, which as been destroyed with the token

There are also weird behaviors after deselecting a single card from a multi-selection — for instance, the "Add counters" actions work, because they come from the card menu, but the "Remove counters" action do not, because they are added to the active card through `Player::updateCardMenu`, but there is no active card anymore.

So we are trading a simple and easy to understand bug (no actions work after deselecting a single card from a multi-selection) for a pile of surprising behaviors with complex activation conditions, which I think is worse.

RickyRister:
> So we are trading a simple and easy to understand bug (no actions work after deselecting a single card from a multi-selection) for a pile of surprising behaviors with complex activation conditions, which I think is worse.

The "complex activation conditions" should come up less often in normal usage than the simple conditions before the fix. For the release, I think we should prioritize making the program work smoothy in common usage scenarios over edge cases.

We can revert temporary fix and try to tackle the root cause after the release.

Elarnon:
That's fair (to be clear, I'm not advocating for tackling the root cause before the release). I don't think the "edge cases" are that uncommon though — "Remove counters" missing after accidentally selecting then de-selecting a card without counters happened multiple times during my last game, as I'm now realizing.

RickyRister:
I agree that the "edge cases" aren't that uncommon. What I mean is that `selecting then de-selecting a card without counters` -> `Remove Counters` -> `Realize it doesn't work` is less common of an occurrence than `de-selecting a card` -> `Literally any card menu shortcut` -> `Realize it doesn't work`. For the release, I think we want to minimize the latter, even if it leaves some holes in the former.

RickyRister:
After some discussion, we decided to just revert #5376 for the release. We will come up with a more proper fix (likely by reworking the card menu logic) after the release.

PR #5765, opened 2025-03-23
Fix delete action in filters not working

RickyRister:

Delete action in right click menu in deck editor tab's filters dock doesn't work

- It works now

https://github.com/user-attachments/assets/3577e859-da0d-4bef-8069-bdd7d4c529bd


Elarnon (commented):
For those that like me were confused for a bit — the issue is that `filterRemove` is not declared as a slot, so the old-style `connect` with the `SLOT` macro doesn't work, but the new-style `connect` with method pointers does work.

Elarnon (approved):

ZeldaZach (approved):

PR #5777, opened 2025-03-29
Make "include online only cards" true by default

RickyRister:
## Related Ticket(s)
- Follow-up to #5759

Users are complaining about online-only printings not showing up in the printing selector. They think that the update broke something, instead of introducing a setting. The setting is also in a hard-to-find place, so they're unlikely to find it on their own.

Make the "Include online-only cards" setting true by default.


Elarnon:
I think we don't want Arena cards on by default, but there is indeed a bug where this setting was never intended to impact online printings of paper cards. See #5778

PR #5778, opened 2025-03-29
fix: Use isRebalanced to detect Arena cards

Elarnon:
## Related Ticket(s)

In #5759 we introduced a setting (off by default) to disable the use of Arena cards. 

This was done by checking the `isOnlineOnly` property of the card, which accidentally also disabled online *printings* of cards that otherwise exist in paper (e.g. Vintage Masters).

This PR does the same thing but uses the `isRebalanced` property instead, which is `true` for Arena cards only and should have been used from the start. Online-only printings such as Vintage Masters are not affected and available independently of the value of the setting. The settings is on by default.


Elarnon:
After a bit more thought, made the setting "on" by default. People that don't want Arena cards will have to disable Alchemy sets and uncheck this setting, which is now called "Include rebalanced (Arena) cards in regular sets".

ZeldaZach (approved):

PR #5784, opened 2025-04-02
Modified setup_addfirstadmin script to bring it up to date to current…

HypersonicWalrus:
… cockatrice_users table

The setup_addfirstadmin script attempted to run SQL that was incompatible with the new cockatrice_users table. I have updated it to fill in some default values for the new fields. LMK if there would be more appropriate default values.

- /servatrice/scripts/linux/setup_addfirstadmin
- That's it that's the only file


HypersonicWalrus:
The SQL command in this script is also the one used in the Wiki under [Setting up Servatrice](https://github.com/Cockatrice/Cockatrice/wiki/Setting-up-Servatrice) and should also be changed

ZeldaZach (approved):

PR #5787, opened 2025-04-03
fix client crash when server goes down during registration

ebbit1q:
## Related Ticket(s)
- Fixes #IssueNumber

the client crashes if the server connection is lost while it is registering (eg if the server crashes)
this is because when an error occurs the client will disconnect from the server, disconnecting from the server is considered an error, this enters a recursive loop.

- when disconnecting do not call the disconnect function
- adds an error message for this case


ebbit1q:
issues with this current pr: when any disconnect happens from the server the client is prompted with the disconnect message and asked to connect using the normal connection dialog, even if currently in the process of registering, after closing the connection dialog we will see this new error and be returned to the register dialog.

ZeldaZach (approved):

PR #5788, opened 2025-04-04
update handling of keywords: AND, OR, NOT in card search

### LINKED ISSUE #4039: Handle keywords (NOT, AND, OR) in card search better
I had a chat with @ebbit1q today. We should think about removing the keyword `not` from card search.
There is support for `-` already to exclude/negate/inverse commands. The written keyword is, as reported by a user today, problematic with some cards like [Not of This World](https://scryfall.com/card/roe/8/not-of-this-world) for example.

**Additional Context**
It seems like we do not support the `and` keyword already.
So there is only one typical operator left: `or`.
Find a good solution for `or`... which single sign could express that? Maybe "|"? I'm not so sure about the mathematical "v" as it messes with the search.
Or only consider the keyword if written in all capital letters (`OR`)? That could apply for AND, and NOT too if needed.

lilyhuang-github:
## Related Ticket(s)
- Fixes #4039 

Keywords: and, or, not are applied to search which causes issues when searching for cards that contain those words, like Not of This World or Truth or Tale.

Change search query to apply keywords only when they're fully capitalized

## Screenshots
image
image
image


lilyhuang-github:
From this I think there should be something in the syntax search that explains this since I don't see it in the syntax help window.

RickyRister:
If you’re updating the tests, then you should add a test case to verify that the lowercase version isn’t parsed as a keyword

lilyhuang-github:
> If you’re updating the tests, then you should add a test case to verify that the lowercase version isn’t parsed as a keyword

I'll try to do that

cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -10,15 +10,15 @@
 peg::parser search(R"(
 Start <- QueryPartList
 ~ws <- [ ]+
-QueryPartList <- ComplexQueryPart ( ws ("and" ws)? ComplexQueryPart)* ws*
+QueryPartList <- ComplexQueryPart ( ws ("AND" ws)? ComplexQueryPart)* ws*
 
-ComplexQueryPart <- SomewhatComplexQueryPart ws $or<[oO][rR]> ws ComplexQueryPart / SomewhatComplexQueryPart
+ComplexQueryPart <- SomewhatComplexQueryPart ws $or<[O][R]> ws ComplexQueryPart / SomewhatComplexQueryPart
```
ebbit1q, line 15: considering removing the lowercase letters makes this a group of one, writing it without the brackets has the same effect
```suggestion
ComplexQueryPart <- SomewhatComplexQueryPart ws $or<OR> ws ComplexQueryPart / SomewhatComplexQueryPart
```

lilyhuang-github: I've tried doing that and it resulted in it crashing but I've made a change that should essentially be the same.

tests/carddatabase/filter_string_test.cpp
```diff
@@ -31,14 +31,16 @@ QUERY(Typing, cat, "t", true)
 
 QUERY(NonMatchingType, cat, "t:kithkin", false)
 QUERY(MatchingType, cat, "t:creature", true)
-QUERY(Not1, cat, "not t:kithkin", true)
-QUERY(Not2, cat, "not t:creature", false)
+QUERY(Not1, cat, "NOT t:kithkin", true)
+QUERY(Not2, cat, "NOT t:creature", false)
+QUERY(NonKeyword1, cat, "not t:kithkin", false)
+QUERY(NonKeyword2, cat, "t:bat or t:creature", false)
```
RickyRister, line 37: I would've expected the new tests to test that "Not dead after all" and "Fact or Fiction" gets matched correctly, since that's the bug we're trying to fix in the first place. But this is fine too, if the test framework makes it too annoying to add.

tests/carddatabase/filter_string_test.cpp
```diff
@@ -35,6 +37,8 @@ QUERY(Not1, cat, "NOT t:kithkin", true)
 QUERY(Not2, cat, "NOT t:creature", false)
 QUERY(NonKeyword1, cat, "not t:kithkin", false)
 QUERY(NonKeyword2, cat, "t:bat or t:creature", false)
+QUERY(NonKeyword3, notDeadAfterAll, "t:instant c:b", true)
```
RickyRister, line 40:
```suggestion
QUERY(NonKeyword3, notDeadAfterAll, "not dead after all", true)
```
We’re testing the name, not the properties

tests/carddatabase/filter_string_test.cpp
```diff
@@ -19,26 +19,32 @@ class CardQuery : public ::testing::Test
...
 QUERY(NonMatchingType, cat, "t:kithkin", false)
 QUERY(MatchingType, cat, "t:creature", true)
-QUERY(Not1, cat, "not t:kithkin", true)
-QUERY(Not2, cat, "not t:creature", false)
+QUERY(Not1, cat, "NOT t:kithkin", true)
+QUERY(Not2, cat, "NOT t:creature", false)
+QUERY(NonKeyword1, cat, "not t:kithkin", false)
+QUERY(NonKeyword2, cat, "t:bat or t:creature", false)
+QUERY(NonKeyword3, notDeadAfterAll, "t:instant c:b", true)
+QUERY(NonKeyword4, truthOrTale, "t:instant c:u", true)
```
RickyRister, line 41:
```suggestion
QUERY(NonKeyword4, truth, "truth or tale", false)
```
ditto here

tests/carddatabase/filter_string_test.cpp
```diff
@@ -20,7 +20,7 @@ class CardQuery : public ::testing::Test
     {
         cat = CardDatabaseManager::getInstance()->getCardBySimpleName("Cat");
         notDeadAfterAll = CardDatabaseManager::getInstance()->getCardBySimpleName("Not Dead After All");
-        truth = CardDatabaseManager::getInstance()->getCardBySimpleName("Truth or Tale");
+        truth = CardDatabaseManager::getInstance()->getCardBySimpleName("Truth OR Tale");
```
RickyRister, line 23:
```suggestion
        truth = CardDatabaseManager::getInstance()->getCardBySimpleName("Truth");
```
I was intending the test case to test that “truth or tale” doesn’t match just “truth”

lilyhuang-github:
Sorry, I didn't really understand how the test worked

RickyRister:
You have to add the new cards here https://github.com/Cockatrice/Cockatrice/blob/0a4b5ae3ddf1f9fe78ccd92a7ef0d69e6e732360/tests/carddatabase/data/cards.xml

lilyhuang-github:
> You have to add the new cards here https://github.com/Cockatrice/Cockatrice/blob/0a4b5ae3ddf1f9fe78ccd92a7ef0d69e6e732360/tests/carddatabase/data/cards.xml

That makes a lot of sense. Thank you

lilyhuang-github:
I'm not entirely sure if it's cause of copyright but the fake cards should still be named in a way to test the keywords

tests/carddatabase/filter_string_test.cpp
```diff
@@ -19,26 +19,32 @@ class CardQuery : public ::testing::Test
...
+        notDeadAfterAll = CardDatabaseManager::getInstance()->getCardBySimpleName("Not Dead");
+        truth = CardDatabaseManager::getInstance()->getCardBySimpleName("Truth");
     }
+}
+// void TearDown() override {}
 
-    // void TearDown() override {}
-
-    CardData cat;
+CardData cat;
```
RickyRister, line 28:
```suggestion
    CardData cat;
    CardData notDeadAfterAll;
    CardData truth;
```

tests/carddatabase/filter_string_test.cpp
```diff
@@ -19,26 +19,32 @@ class CardQuery : public ::testing::Test
     void SetUp() override
     {
         cat = CardDatabaseManager::getInstance()->getCardBySimpleName("Cat");
+        notDeadAfterAll = CardDatabaseManager::getInstance()->getCardBySimpleName("Not Dead");
+        truth = CardDatabaseManager::getInstance()->getCardBySimpleName("Truth");
     }
+}
```
RickyRister, line 25:
```suggestion
```

tooomm:
This should also deal with #5974

lilyhuang-github:
I'll use the formatting script which should deal with the styling check.

lilyhuang-github:
> This should also deal with #5794 I think?

I think thats correct

lilyhuang-github:
I'm presuming that the carddatabase test is failing due to size since there's two more cards added to the cards.xml, but I don't know why the other OS's checks aren't failing as well.

RickyRister:
> I'm presuming that the carddatabase test is failing due to size since there's two more cards added to the cards.xml, but I don't know why the other OS's checks aren't failing as well.

We only run unit tests on some of the builds

lilyhuang-github:
> > I'm presuming that the carddatabase test is failing due to size since there's two more cards added to the cards.xml, but I don't know why the other OS's checks aren't failing as well.
> 
> We only run unit tests on some of the builds

That's good to know. I'll change that unit test since there's two more cards.

RickyRister (approved):

ZeldaZach (approved):

ZeldaZach:
Thanks for taking the time to contribute! This is a great change that I'm sure users will love :)

tests/carddatabase/data/cards.xml
```diff
@@ -31,5 +31,31 @@
                 <pt>4/4</pt>
             </prop>
         </card>
+		<card>
```
ebbit1q, line 34: the indentation in this file is a bit wonky now, we don't use tabs in the rest of the file

PR #5791, opened 2025-04-05
Add search filter to card view window

### LINKED ISSUE #2993: We need a filter when searching a deck during gameplay
It would be nice to be able to type the name of a card after viewing the library to quickly pinpoint it. I play EDH and with 9 fetchlands in my deck, it gets tedious to visually look for the card.

RickyRister:
## Related Ticket(s)
- Fixes #2993 

We want card search in the card view window

https://github.com/user-attachments/assets/5c0cfd5c-a0a8-4b43-81d4-55976decb7c2

- Refactored the search syntax help window code into a separate file
  - The `QTextBrowser` is now unparented, but closes when the `QLineEdit` it's connected to is destroyed. 
    - This should still leave to the exact same behavior as before in the deck editor tab.
- Added a search bar to the top of the `ZoneViewWidget`
  - Also reduced widget spacing in `ZoneViewWidget` to make the search bar fit nicer
- Implemented filtering logic 
  - `ZoneViewZone` stores a filterString. 
    - `ZoneViewWidget` sets `ZoneViewZone`'s filterString whenever the `QLineEdit` changes
  - Filtering happens in `ZoneViewZone`, right before sorting happens.
    - if `CardItem`'s `CardInfo` passes the filterString check, it's added to the list to be sorted and set to visible (in order to clear any hides from previous filterStrings)
    - if it doesn't pass the check, then the `CardItem` is hidden
  - Apparently this is all you need to do to implement filtering. Seems to work flawlessly for me so far, even for corner cases.

## Screenshots

Screenshot 2025-04-04 at 11 20 26 PM

Screenshot 2025-04-04 at 11 20 35 PM

### DIFF
--- a/cockatrice/src/game/game_view.cpp
+++ b/cockatrice/src/game/game_view.cpp
@@ -11,7 +11,7 @@ GameView::GameView(GameScene *scene, QWidget *parent) : QGraphicsView(scene, par
 {
     setBackgroundBrush(QBrush(QColor(0, 0, 0)));
     setRenderHints(QPainter::TextAntialiasing | QPainter::Antialiasing);
-    setFocusPolicy(Qt::NoFocus);
+    setFocusPolicy(Qt::ClickFocus);
     setViewportUpdateMode(BoundingRectViewportUpdate);
 
     connect(scene, &GameScene::sceneRectChanged, this, &GameView::updateSceneRect);
--- a/cockatrice/src/game/zones/view_zone.cpp
+++ b/cockatrice/src/game/zones/view_zone.cpp
@@ -148,7 +148,16 @@ void ZoneViewZone::updateCardIds(CardAction action)
 // Because of boundingRect(), this function must not be called before the zone was added to a scene.
 void ZoneViewZone::reorganizeCards()
 {
-    CardList cardsToDisplay(cards);
+    // filter cards
+    CardList cardsToDisplay = CardList(cards.getContentsKnown());
+    for (auto card : cards) {
+        if (filterString.check(card->getInfo())) {
+            card->show();
+            cardsToDisplay.append(card);
+        } else {
+            card->hide();
+        }
+    }
 
     // sort cards
     QList<CardList::SortOption> sortOptions;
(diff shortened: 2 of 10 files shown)


cockatrice/src/game/game_view.cpp
```diff
@@ -11,7 +11,7 @@ GameView::GameView(GameScene *scene, QWidget *parent) : QGraphicsView(scene, par
 {
     setBackgroundBrush(QBrush(QColor(0, 0, 0)));
     setRenderHints(QPainter::TextAntialiasing | QPainter::Antialiasing);
-    setFocusPolicy(Qt::NoFocus);
+    setFocusPolicy(Qt::ClickFocus);
```
RickyRister, line 14: fyi, this was the cause of the QLineEdit not being focused problem back then 

https://github.com/Cockatrice/Cockatrice/issues/2993#issuecomment-356810539

cockatrice/src/game/zones/view_zone.cpp
```diff
@@ -148,7 +148,16 @@ void ZoneViewZone::updateCardIds(CardAction action)
 // Because of boundingRect(), this function must not be called before the zone was added to a scene.
 void ZoneViewZone::reorganizeCards()
 {
-    CardList cardsToDisplay(cards);
+    // filter cards
+    CardList cardsToDisplay = CardList(cards.getContentsKnown());
```
RickyRister, line 152: not sure if this is the correct place to get the value for _contentsKnown from. Previous code was using a copy constructor. Except CardList doesn't actually have an explicit copy constructor, so I don't know what values it was copying over. I would assume that it did copy over the `contentsKnown` though, so I assume my new code is still correct.

cockatrice/src/game/game_view.cpp
ebbit1q: are you sure that changing it for the entire view doesn't have any side effects?

ebbit1q: ok, so I just tested my suspicions, this introduces a regression:
when using the mouse in the player the chat now gets defocused
- create a local game
- shift enter to focus the chat and say something
- using keybinds works as normal in the chat
- using the mouse to eg move a card should not defocus the chat, with this change the chat loses focus

RickyRister: How important is it that clicking in the game view doesn't defocus the chat? If anything, I'd *expect* clicking anywhere off the chat to defocus it.

RickyRister: After some discussion, we decided that we can merge the change in first and cut a beta and see if anyone complains. If the community does want the focusing behavior back, then we could try to find a way to readd the old behavior after merge

ZeldaZach (approved):

PR #5793, opened 2025-04-05
fix: Disable HTTP compression when downloading pictures

Elarnon:
This causes Qt to leak file descriptors and causes the "Too many open file descriptors" error that we sporadically see, see https://bugreports.qt.io/browse/QTBUG-135641


ebbit1q (approved):
:+1:

ZeldaZach (approved):

PR #5799, opened 2025-04-06
Revert source file globbing

BruebachL:
Globbing for source files confuses my language server which leads to me not having auto-complete, linting and code suggestions.
In short, it relegates my IDE to a fancy text editor with VCS and build integration.

- Explicitly list source files again.


RickyRister:
Sorry about making the change without understanding the full extent of the consequences

RickyRister (approved):

Issue #3278, opened 2018-06-10
Feature: set hidden annotation


sbheinric:
There is currently no way to play Goblin Game because it requires players to choose a hidden number and then reveal it.  

I would propose a simple new generic feature "Set hidden annotation / reveal hidden annotation" option that works similarly to "set annotation" on cards and tokens, which would increase flexibility by allowing cards like Goblin Game, and potentially others, to become playable.

ZeldaZach:
I feel like this is a duplicate, but I don't remember...

In any sense, I do agree it would be useful, just need to plan logistics

Daenyth:
Pretty sure this is a dupe, look at game rules violation label. Yes we'd
like to do this

ebbit1q:
this is similar to permanent counters https://github.com/Cockatrice/Cockatrice/issues/3820

Edahsrevlis:
For Wheel of Misfortune and similar, I would add a "make secret" checkbox to the annotation dialog. A secret card would only display its annotation while face-up. So you could create a token, turn it face-down, then Set Annotation... ("make secret") and choose your number/player/etc. It might be worth flipping the card face down when choosing Secret in case someone sets the secret annotation while it's face-up, to save lots of user error.

Instead of printing the real annotation in the chat, it would replace the annotation with "secret". On the card it could be blue to distinguish from "secret" as an actual string.

Then whichever line sets/updates the annotation text
OLD: DisplayAnnotation(annotation)
NEW: DisplayAnnotation(IsSecret && !IsFaceUp ? "secret" : annotation, IsSecret ? Color.blue : Color.white)

Wish I could do more directly but I'm a lowly C# coder.

ebbit1q:
another option instead of allowing adding counters to card in your hand is to allow the creation of tokens face down, with their name hidden from other players until flipped, this would be a small addition to the protocol.

jseakle:
It would be extremely useful to me to have the ability to create tokens face down. i've been playing a lot of a format that requires this every turn, and so far we just have to be on a voice call and ask the other person not to look at the screen as we spawn in cards.

ebbit1q:
which format requires the creation of face down tokens?

I'm pretty sure it would be safe to create face down tokens but creating tokens in someone's hand sounds like it would be too easy to abuse and hard to rewind in case of a mistake.

jseakle:
[...]

PR #5800, opened 2025-04-06
Support creating face-down tokens

### LINKED ISSUE #4994: Secret Voting Feature Request
**Similar Requests**
None that I could find

To handle cards like Wheel of Misinformation/Prisoner's Dilemma, I think cockatrice needs a way to handle secret voting. I don't 100% know the best way to handle this as I'm sure there are many avenues that could be taken to achieve this, but I did have an idea that I thought could work.

When creating a token, give us the ability to create the token face down. Creating tokens is a simple way that we relay all kinds of information to the table and it's a sure-fire way that we can "lock in" a vote without the ability for someone to cheat. So when playing the Wheel of Misinformation, the players would all create a token face down with the chosen number and once all players have created their token, everyone can flip them face-up.

Ideally the game log would only show "(user) has created a token face-down" or something like that, so it doesn't display the name of the token to the chat like it does when you create them face-up. Then when the token is turned face-up, the game log can display the token and the text on the token then.

**Additional Context**
I know the current advice on cockatrice is just to not play these cards, but it just seems a bit unfair to ask people not to play a set of cards just because the platform can't currently support them. And for many, we use cockatrice to test out new deck ideas and deck updates before we go and buy the cards for our paper decks, so not being able to test a card can be frustrating to people.

Thank you for all the time and hard work you spend on this project. It's a fantastic platform that many use to play MtG online and I love it.

RickyRister:
## Related Ticket(s)

* Fixes #4994
* Partially addresses #3278

Users want a way to write hidden information in Cockatrice, which we've decided to handle through face-down tokens

https://github.com/user-attachments/assets/b7bd9809-f030-45e6-99f6-4ab441b1b6d4

- Added a new `Create face-down token` checkbox to the token creation dialog 
  - Checking the checkbox will disable the pt and annotation line edits
    - Those attributes aren't supported on face-down tokens, since Cockatrice currently can't hide those values on non-predefined tokens
- Updated create token command and event proto to add `face_down` bool field.
- Updated server and client code to handle new face down token creation
- Updated game log to log face-down tokens with "PLAYER creates face down token."

## Screenshots

Screenshot 2025-04-14 at 1 23 26 AM

### DIFF
--- a/cockatrice/src/server/message_log_widget.cpp
+++ b/cockatrice/src/server/message_log_widget.cpp
@@ -217,7 +217,7 @@ void MessageLogWidget::logCreateArrow(Player *player,
 void MessageLogWidget::logCreateToken(Player *player, QString cardName, QString pt, bool faceDown)
 {
     if (faceDown) {
-        appendHtmlServerMessage(tr("%1 creates face down token.").arg(sanitizeHtml(player->getName())));
+        appendHtmlServerMessage(tr("%1 creates a face down token.").arg(sanitizeHtml(player->getName())));
     } else {
         appendHtmlServerMessage(tr("%1 creates token: %2%3.")
                                     .arg(sanitizeHtml(player->getName()))
--- a/cockatrice/src/server/message_log_widget.cpp
+++ b/cockatrice/src/server/message_log_widget.cpp
@@ -214,12 +214,16 @@ void MessageLogWidget::logCreateArrow(Player *player,
     }
 }
 
-void MessageLogWidget::logCreateToken(Player *player, QString cardName, QString pt)
+void MessageLogWidget::logCreateToken(Player *player, QString cardName, QString pt, bool faceDown)
 {
-    appendHtmlServerMessage(tr("%1 creates token: %2%3.")
-                                .arg(sanitizeHtml(player->getName()))
-                                .arg(cardLink(std::move(cardName)))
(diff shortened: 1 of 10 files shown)


cockatrice/src/server/message_log_widget.cpp
```diff
@@ -214,12 +214,16 @@ void MessageLogWidget::logCreateArrow(Player *player,
...
 
-void MessageLogWidget::logCreateToken(Player *player, QString cardName, QString pt)
+void MessageLogWidget::logCreateToken(Player *player, QString cardName, QString pt, bool faceDown)
 {
-    appendHtmlServerMessage(tr("%1 creates token: %2%3.")
-                                .arg(sanitizeHtml(player->getName()))
-                                .arg(cardLink(std::move(cardName)))
-                                .arg(pt.isEmpty() ? QString() : QString(" (%1)").arg(sanitizeHtml(pt))));
+    if (faceDown) {
+        appendHtmlServerMessage(tr("%1 creates face down token.").arg(sanitizeHtml(player->getName())));
```
Elarnon, line 220:
```suggestion
        appendHtmlServerMessage(tr("%1 creates a face down token.").arg(sanitizeHtml(player->getName())));
```

Elarnon (approved):

ZeldaZach (approved):

PR #5801, opened 2025-04-06
change settings entry of the cod conversion prompt to a combobox

ebbit1q:
## Related Ticket(s)
- introduced by  #5514

the settings entry for converting to .cod when adding a tag to a deck in the visual deck selector is unclear in its use, this was shown to me by a translator on the discord: caledor92

- replace the two checkboxes of which one state is ignored if one is checked with a three state combobox for better user experience

## Screenshots

image

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -563,6 +563,13 @@ void AppearanceSettingsPage::retranslateUi()
     maxFontSizeForCardsLabel.setText(tr("Maximum font size for information displayed on cards:"));
 }
 
+enum visualDeckStoragePromptForConversionIndex
+{
+    visualDeckStoragePromptForConversionIndexNone,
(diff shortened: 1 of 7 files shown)


BruebachL:
This looks like a better user experience indeed, thank you.

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -664,7 +664,7 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
         visualDeckStoragePromptForConversionSelector.setCurrentIndex(visualDeckStoragePromptForConversionIndexNone);
     }
     connect(&visualDeckStoragePromptForConversionSelector, QOverload<int>::of(&QComboBox::currentIndexChanged), this,
-            [this](int index) {
```
ebbit1q: the this pointer is indeed unnecessary, thanks for noticing

ZeldaZach (approved):

PR #5836, opened 2025-04-15
Give every filter type its own button in VDD

BruebachL:
Pro: The filters are way easier to navigate
Con: Our icon choices are ass

image
image

(This also exposes a new method in the QuickSettingsButtonWidget to set the icon)


ZeldaZach (approved):

ZeldaZach:
Windows failing due to https://github.com/jurplel/install-qt-action/issues/283 overriding

PR #5837, opened 2025-04-16
Implement deck analytics widgets for Visual Deck Editor

BruebachL:
## Screenshots
image
image

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_editor/visual_deck_editor_sample_hand_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_editor/visual_deck_editor_sample_hand_widget.cpp
@@ -1,0 +1,105 @@
+#include "visual_deck_editor_sample_hand_widget.h"
+
+#include "../../../../deck/deck_loader.h"
+#include "../../../../game/cards/card_database_manager.h"
+#include "../cards/card_info_picture_widget.h"
+
+#include <random>
+
+VisualDeckEditorSampleHandWidget::VisualDeckEditorSampleHandWidget(QWidget *parent, DeckListModel *_deckListModel)
+    : QWidget(parent), deckListModel(_deckListModel)
+{
+    layout = new QVBoxLayout(this);
+    setLayout(layout);
+
+    resetButton = new QPushButton(this);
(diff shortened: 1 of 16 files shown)


cockatrice/src/client/ui/widgets/visual_deck_editor/visual_deck_editor_sample_hand_widget.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void VisualDeckEditorSampleHandWidget::retranslateUi()
+{
+    resetButton->setText(tr("Reset"));
+}
+
+void VisualDeckEditorSampleHandWidget::setDeckModel(DeckListModel *deckModel)
+{
+    deckListModel = deckModel;
+    // connect(deckListModel, &DeckListModel::dataChanged, this, &VisualDeckEditorSampleHandWidget::updateDisplay);
```
ZeldaZach, line 37: intentional?

BruebachL: It's horribly expensive so for now, it's commented out.

BruebachL: You can manually fetch a new hand though. But yeah, fetching 7 cards from the DB is too much for us.

cockatrice/src/client/ui/widgets/visual_deck_editor/visual_deck_editor_sample_hand_widget.cpp
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        }
+    }
+
+    if (mainDeckCards.isEmpty())
+        return randomCards;
+
+    // Shuffle the deck
+    std::random_device rd;
```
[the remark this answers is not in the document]
BruebachL: (I also don't think we need to be concerned about competitive integrity here. It'd be nice to have the shuffle match the algorithm we use but since it's 7 random cards, I think if we have "sufficient" randomness, it'll be alright.)

PR #5843, opened 2025-04-17
Fix intermittent segfault in VDS sort

RickyRister:

Cockatrice would intermittently segfault inside the VDS sort function for me. No one else has complained about the problem, so I assume it only happens on mac.

Recently, that segfault has started happening way more frequently.

- call `reapplySortAndFilters` inside a `QTimer::singleShot`.
  - I did some manual testing after the change and it hasn't segfaulted for me yet, so I guess that fixes it?

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
@@ -211,14 +211,16 @@ void VisualDeckStorageWidget::createRootFolderWidget()
     scrollArea->setWidget(folderWidget); // this automatically destroys the old folderWidget
     scrollArea->widget()->setMaximumWidth(scrollArea->viewport()->width());
     scrollArea->widget()->adjustSize();
-    reapplySortAndFilters();
+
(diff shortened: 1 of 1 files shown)


cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
```diff
@@ -211,14 +211,16 @@ void VisualDeckStorageWidget::createRootFolderWidget()
     scrollArea->setWidget(folderWidget); // this automatically destroys the old folderWidget
     scrollArea->widget()->setMaximumWidth(scrollArea->viewport()->width());
     scrollArea->widget()->adjustSize();
-    reapplySortAndFilters();
+
+    // Deck sort intermittently segfaults due to invalid DeckPreviewWidget, unless we use QTimer trick
```
BruebachL, line 215: Please explain the QTimer trick so that newer devs stumbling upon this don't have to guess
(We have to schedule a QTimer for this so that the logic inside the function doesn't try to access widgets that haven't been processed by the event loop yet [A QTimer fires after the current event cycle in the next one])

BruebachL (approved):

BruebachL (approved):

ZeldaZach (approved):

PR #5846, opened 2025-04-18
Refactor files in src/dialogs to new Qt Slot/Signal syntax

RickyRister:
## Related Ticket(s)
- Follow-up to #5432

- Update all usages of slot/signal in `src/server` to new syntax
- Put a TODO comment on `DlgUpdate`'s reinstall button connection, since the previous connection didn't actually do anything

### DIFF
--- a/cockatrice/src/dialogs/dlg_update.cpp
+++ b/cockatrice/src/dialogs/dlg_update.cpp
@@ -42,6 +42,7 @@ DlgUpdate::DlgUpdate(QWidget *parent) : QDialog(parent)
     buttonBox->addButton(ok, QDialogButtonBox::AcceptRole);
 
     connect(gotoDownload, &QPushButton::clicked, this, &DlgUpdate::gotoDownloadPage);
+    // TODO: make reinstall button actually do something when clicked
     // connect(manualDownload, &QPushButton::clicked, this, &DlgUpdate::downloadUpdate);
     connect(stopDownload, &QPushButton::clicked, this, &DlgUpdate::cancelDownload);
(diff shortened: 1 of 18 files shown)


cockatrice/src/dialogs/dlg_update.cpp
```diff
@@ -41,10 +41,11 @@ DlgUpdate::DlgUpdate(QWidget *parent) : QDialog(parent)
     enableUpdateButton(false);             // Unless we know there's an update available, you can't install
     buttonBox->addButton(ok, QDialogButtonBox::AcceptRole);
 
-    connect(gotoDownload, SIGNAL(clicked()), this, SLOT(gotoDownloadPage()));
-    connect(manualDownload, SIGNAL(clicked()), this, SLOT(downloadUpdate()));
-    connect(stopDownload, SIGNAL(clicked()), this, SLOT(cancelDownload()));
-    connect(ok, SIGNAL(clicked()), this, SLOT(closeDialog()));
+    connect(gotoDownload, &QPushButton::clicked, this, &DlgUpdate::gotoDownloadPage);
+    // TODO: make reinstall button actually do something when clicked
+    // connect(manualDownload, &QPushButton::clicked, this, &DlgUpdate::downloadUpdate);
```
RickyRister, line 46: `DlgUpdate` did not have an overload of `downloadUpdate` with zero parameters. The old slot/signal syntax does not type-check, so it compiles but no-ops during runtime. New syntax is type-checked, and so doesn't compile if I just made the swap. 

I don't know what the expected behavior of the button is, so I'm just going to put it off to a separate PR

ZeldaZach (approved):

PR #5848, opened 2025-04-18
VDS performance fixes

BruebachL:
Users are complaining that VDS takes too long to load, especially on weaker machines with a lot of decks.

- Block updates on banner card combo box (little effect but doesn't hurt)
- don't validate cardInfo (i.e. look up a card we already stored in the decklist AGAIN in the card db to ensure it remains valid. If it's in the decklist, it's probably a valid card).
- use ItemModel instead of looped addItem (also relatively little impact)

## Screenshots:
Before:

image

After: (Refresh tags now taking up the most execution time)

image

Further optimized: (All refresh tags calls removed)

image

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
@@ -13,6 +13,7 @@
 #include <QMessageBox>
 #include <QMouseEvent>
 #include <QSet>
+#include <QStandardItemModel>
 #include <QVBoxLayout>
 
 DeckPreviewWidget::DeckPreviewWidget(QWidget *_parent,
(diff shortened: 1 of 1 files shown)


cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
```diff
@@ -226,14 +225,23 @@ void DeckPreviewWidget::updateBannerCardComboBox()
...
+
+    int row = 0;
     for (const auto &pair : pairList) {
         QVariantMap dataMap;
         dataMap["name"] = pair.first;
         dataMap["uuid"] = pair.second;
 
-        bannerCardComboBox->addItem(pair.first, dataMap);
+        QStandardItem *item = new QStandardItem(pair.first);
+        item->setData(dataMap, Qt::UserRole);
```
RickyRister, line 239: Turns out you're allowed to directly wrap QPair in QVariant, which would save on creating a `QVariantMap`. Would doing that improve performance, or is it too minor to matter? 
```suggestion
        item->setData(QVariant::fromValue(pair), Qt::UserRole);
```

BruebachL: image
image

We shave a little off. We may as well. Good find :)

ZeldaZach (approved):

PR #5849, opened 2025-04-18
CI: Remove Ubuntu 20.04 (EOL May 31st)

tooomm:
Ubuntu 20.04 LTE is out of support next month and an Extended Security Maintenance (ESM) subscription is needed for further security updates.

- Remove 20.04 LTE from build matrix


ebbit1q:
we want to run our tests on the oldest system in our matrix, when removing an os please enable tests for the next one in order to make sure our tests also work on old systems

tooomm:
> we want to run our tests on the oldest system in our matrix

It's actually the other way around currently and we run them on the newest Linux system.

Want me to change that?
I support the idea.

tooomm:
@ebbit1q I suggest to merge this as is.

If you think our testing approach should change - it needs an update for all systems, probably best in a separate PR.

ZeldaZach (approved):

PR #5850, opened 2025-04-18
CI: Update versions

tooomm:
>[!WARNING]
>WIP

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


ZeldaZach:
Intentionally keeping Windows Qt to 6.6 as 6.7+ introduces a nasty bug we haven't yet identified a fix for.

ebbit1q:
resizing? like the entire application or the card views?

tooomm:
>Intentionally keeping Windows Qt to 6.6 as 6.7+ introduces a nasty bug with game windows resizing we haven't yet identified a fix for.

Is that still an issue with newest Qt 6.8.3 LTS?
Win10 binary is available for testing in this action run: https://github.com/Cockatrice/Cockatrice/actions/runs/14537996495

ZeldaZach:
@ebbit1q Windows Qt6.7+ -- When you right click on view library, the entire inner application resizes and can't be fixed. I'll post an example of it at some point, but it's painful.

tooomm:
>Windows Qt6.7+ -- When you right click on view library, the entire inner application resizes and can't be fixed. I'll post an example of it at some point, but it's painful.

Right clicking on the library and selecting `View Library` seems to work without issues:
image

Looks like the window is drawn differently than in Qt 6.6 though.

.github/workflows/desktop-build.yml
```diff
@@ -210,25 +210,22 @@ jobs:
             make_package: 1
 
           - target: 14
-            soc: Apple
```
ebbit1q, line 213: I disagree with removing these, even though the only value being checked is Intel, setting the others to Apple provides useful information

tooomm: I'm not convinced keeping the Apple tag does provide useful information. It's the new default architecture/build and will replace Intel on Macs/macOS.
Hence, only labeling the unusual & special build (Intel) makes sense in my opinion.

Same with Windows and 32/64bit. x64 was standard, so we labelled only the non-standard build (32bit) later.
Now, there is only 64bit and we do not need to label it on top.

In the CI builds and artifacts we are not showing the Apple label anymore already:
image

tooomm:
> Windows Qt6.7+ -- When you right click on view library, the entire inner application resizes and can't be fixed. I'll post an example of it at some point, but it's painful.

@ZeldaZach Can you please check this again? I was not able to reproduce and can not confirm the issue according to your description.

tooomm:
Changes here are already implemented or in the meantime no longer relevant. Closing

PR #5852, opened 2025-04-18
Don't add duplicate CardInfos to set.

BruebachL:
Through some quirk I haven't investigated further, we call addToSet multiple times for the same CardInfoPtr. This adds some checks to ensure we don't, which leads to a deduplicated CardInfoPerSetMap

- Add guards if maps already contain CardInfo


BruebachL:
The issue is possibly here:
image

ZeldaZach (approved):

PR #5855, opened 2025-04-19
[GDE] Disable add tag button until text is entered.

### LINKED ISSUE #5805: Deck Tags Manager: Disable add button until tag name is entered
In the new `Deck Tags Manager`, where one would edit tags, you can click the `Add Tag` button without entering a name, just to see an error message with a hint as a result:
Image
Image

I think it would be more straight forward and a better user experience to have the button disabled unless any input is given in the name field. That should make the error dialog obsulete even.

Also, the dialog windows seems to not be correctly wired to a parent as they do not show the Cockatrice app icon next to their titles. (See #998)

BruebachL:
## Related Ticket(s)
- Fixes #5805 

Disabling the add tag button if we disallow empty tags makes more sense UI-wise.

- Disable addTag button on textChanged if the text is empty.

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_dialog.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_dialog.cpp
@@ -11,7 +11,7 @@
 #include <QVBoxLayout>
 
 DeckPreviewTagDialog::DeckPreviewTagDialog(const QStringList &knownTags, const QStringList &activeTags, QWidget *parent)
-    : QDialog(parent), activeTags_(activeTags)
+    : QDialog(parent), activeTags(activeTags)
 {
     resize(400, 500);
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/ui/widgets/visual_deck_storage/deck_preview/deck_preview_tag_dialog.cpp
```diff
@@ -100,45 +100,48 @@ DeckPreviewTagDialog::DeckPreviewTagDialog(const QStringList &knownTags, const Q
...
+    addTagButton = new QPushButton(this);
+    addTagButton->setEnabled(false);
+    addTagLayout->addWidget(newTagInput);
+    addTagLayout->addWidget(addTagButton);
     mainLayout->addLayout(addTagLayout);
 
-    connect(addTagButton_, &QPushButton::clicked, this, &DeckPreviewTagDialog::addTag);
+    connect(addTagButton, &QPushButton::clicked, this, &DeckPreviewTagDialog::addTag);
+    connect(newTagInput, &QLineEdit::textChanged, this,
+            [=, this](const QString &text) { addTagButton->setEnabled(!text.trimmed().isEmpty()); });
```
RickyRister, line 141: why do you need to have `=` in the capture? Shouldn't `this` already capture the `addTagButton`?

BruebachL: Honest to god, I actually don't know what the = stands for, I never bothered to look up the lambda syntax, I just do the = cause it seems to be the default and then added "this" to it when it kept complaining that "implicit capture of this is disallowed in c++ w/e".

So I'll defer to you on that.

RickyRister: I think `=` captures everything in the function by value, while `this` captures everything in the class, and makes it so the code inside the lambda executed as if it was a method on the object.

ZeldaZach (approved):

PR #5856, opened 2025-04-19
VDS: Optimize refreshTags in VisualDeckStorageTagFilterWidget

RickyRister:
## Related Ticket(s)
- Follow-up to #5848

After the optimizations in the above PR, Refresh tags now taking up the most execution time. There are some simple optimizations that can be made in `VisualDeckStorageTagFilterWidget`.

- Remove accidental redundant calls to `gatherAllTags`
- Gather tags into a `QSet` instead of `QStringList`.
  - `removeTagsNotInList` calls `tags.contains` in a loop. There could be quite a large number of tags, so using a set instead of a list here should be a performance improvement. 
- Also did some minor refactoring 
  - Made the collect tag methods const 
  - Moved some methods and the `VisualDeckStorageWidget` field to private, since they probably should've been private in the first place.

## Screenshots

Before
Screenshot 2025-04-19 at 1 38 15 AM

After
Screenshot 2025-04-19 at 1 38 00 AM

### DIFF
--- a/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
@@ -64,13 +64,13 @@ void VisualDeckStorageTagFilterWidget::filterDecksBySelectedTags(const QList<Dec
 
 void VisualDeckStorageTagFilterWidget::refreshTags()
 {
-    QStringList allTags = gatherAllTags();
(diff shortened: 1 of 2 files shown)


BruebachL (approved):

BruebachL:
We can push this further by not having every DeckPreviewWidget make the Storage Widget refresh the tags list in the future (instead having the storage widget keep track of all it's children and when they are finished) or similar but these are good and simple changes that should improve performance in my eyes.

Thanks!

ZeldaZach (approved):

PR #5877, opened 2025-04-23
client: Support arbitrary game zones

Elarnon:
## Related Ticket(s)

This will improve backwards-compatibility when we get around to #1740.

Currently, the client ignores cards in unknown zones, as there is an implicit assumption that the set of zones known by the server and the client are the same.

This patch makes it so that the client accept "custom zones" from the server (zones outside the builtin deck, graveyard, exile, sideboard, table, stack and hand zones) using the information from the ServerInfo_CardZone. Moving cards from/into these zones happens through a "View custom zone" action in the Game > Player menu and properly appears in the chat.

Note that this patch intentionally does not introduce any support for having the server actually create such zones. Instead, this patch aims to improve backwards compatibility for when we do get to adding this capability in the future, by making sure that current clients will be able to interact with future new zones (even if suboptimally).

Note: for testing, I suggest adding a new custom zone in `server_player.cpp`. Hidden zones with coordinates will not work, but other types of zones should work (although using a ZoneViewZone for a non-hidden zone with coordinates is a bit wonky, it mostly works).

Note: this PR is rebased on top of #5728 and includes the commit from #5728; only the last commit ("client: Support arbitrary game zones") is new in this PR.

### DIFF
--- a/.ci/Fedora42/Dockerfile
+++ b/.ci/Fedora42/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:40
+FROM fedora:42
 
 RUN dnf install -y \
         ccache \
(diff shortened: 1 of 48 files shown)


ZeldaZach (approved):

ZeldaZach:
conflicts

PR #5882, opened 2025-04-24
feat: Configurable colors for card counter

### LINKED ISSUE #2020: Add more color options for counters
Not really an "issue" per say but can we add blue, and or black counters instead of ony having the normal 3 red/green/yellow? maybe even other colors like purple. some players like to use certain colors for certain things, like G=pumps and buffs. Red=loyality yellow= charge maybe black could be debuffs, you get the idea

Elarnon:
## Related Ticket(s)
- Fixes #2020

Only three types of card counters are available, with hard-coded colors (red, green, yellow). The hard-coded colors are an accessibility issue, and people have requested being able to create more counters with different colors.

This patch adds support for:

 - User-defined colors for card counters;
 - 3 additional types of card counters.

The colors used for counters is stored locally and not shared with other users. This is intentional as the feature is likely to be used for improved accessibility.

In order to preserve backwards-compatibility, and because I don't have a better idea, counters keep their existing color-based names (Red, Green, Yellow) in menus and in the message log. For consistency, the new counters also get assigned color-based names (Cyan, Purple, Magenta).

This choice is a compromise, as allowing user-defined names for counters raises many additional (UI/UX) questions that I don't know how to answer. A good long-term solution would be to include counter names as part of a game definition system and hence would be in scope for #1740.

The choice of adding 3 additional types of counters and the Cyan, Purple and Magenta names are not random. The existing code for determining counter colors goes: Red, Green, Yellow, Cyan, Purple, Magenta, Black (unreadable) and thus 6 is the maximum number of counters that existing versions of Cockatrice are able to support. This way, released clients get a degraded experience (cannot interact with the new counters, messages in the server log say "Player X places 1 on Card (now 1)" without specifying 1 of what), but do see the counters properly.

## Screenshots

image
image
image
image


cockatrice/src/server/message_log_widget.cpp
```diff
@@ -604,28 +605,15 @@ void MessageLogWidget::logSetCardCounter(Player *player, QString cardName, int c
     QString finalStr;
     int delta = abs(oldValue - value);
     if (value > oldValue) {
-        finalStr = tr("%1 places %2 %3 on %4 (now %5).");
+        finalStr = tr("%1 places <font class=\"blue\">%n</font> %2 counter(s) on %3 (now %4).", "", delta);
```
ebbit1q, line 608: I think we want to avoid giving html tags to translators, we are not required to use the %n placeholder while still giving it to the tr call for plurals, we can keep using %2

tooomm:
Isn't that creating an issue?

Imagine a complex situation and users start to chat about it and refer to arrows with their colour. They are now no longer matching for each player.

cockatrice/src/server/message_log_widget.cpp
Elarnon: I thought that would cause issues with languages where the number is not in front of the name but I'm not sure why I thought that — will update back to `%2 %3 %4 %5` and the HTML in the arg later tonight.

Elarnon:
> Imagine a complex situation and users start to chat about it and refer to arrows with their colour. They are now no longer matching for each player.

This patch only changes the color of the card counters, not the arrows (although to be fair I believe most people only use the red arrow anyway). I think for counters it is not a problem — in the vast majority of cases there's only one kind of counter on a permanent, and when there are multiples you are likely to ask "what are the counters on this?" anyways.

I'm not too happy about having the counters being *called* with color names though in the server log, but I wanted to preserve behavior w.r.t. older versions (which will always call the first counter "red" etc.) and I don't think people look at those specific messages too much. Happy to take alternative suggestions though.

tooomm:
You are absolutely right of course, I mixed counters and arrows when thinking ahead. 🙈 

>I'm not too happy about having the counters being called with color names though in the server log, but I wanted to preserve behavior w.r.t. older versions (which will always call the first counter "red" etc.) and I don't think people look at those specific messages too much. Happy to take alternative suggestions though.

More generic labels could be going with letters `A`, `B`, `C` instead?
Or maybe shapes? Square, Triangle, Circle?

cockatrice/src/game/player/player.cpp
```diff
@@ -3877,15 +3884,19 @@ void Player::updateCardMenu(const CardItem *card)
                 cardMenu->addAction(aSelectAll);
                 cardMenu->addAction(aSelectRow);
 
+                cardMenu->addSeparator();
+                auto *cardCountersMenu = new QMenu(tr("Other counters"));
```
RickyRister, line 3888: Just move all the counter actions into the new counter menu, instead of splitting the old and new counter actions between the menus

Elarnon: I was thinking about that — I think it's better to be consistent and initially put everything in the separate menu as you are suggesting (we clearly don't want to add the new counters to the main menu, it is big enough as is). But then I thought that this makes the counter actions harder to access for people that use the right click menu rather than shortcuts and ended up with this compromise.

I don't know how common that is (I personally always use the shortcuts), so if the consensus is to put everything into a separate menu, happy to do so.

cockatrice/src/settings/card_counter_settings.cpp
```diff
@@ -0,0 +1,61 @@
+#include "card_counter_settings.h"
+
+#include <QColor>
+#include <QSettings>
+
+CardCounterSettings::CardCounterSettings(const QString &settingsPath, QObject *parent)
+    : QObject(parent), m_settings(new QSettings(settingsPath + "global.ini", QSettings::IniFormat))
```
RickyRister, line 7: `CardCounterSettings` doesn't extend from `SettingsManager` like all the other custom settings classes do. It also  uses the `global.ini` instead of a separate ini file. Is this intentional?

RickyRister (approved):

cockatrice/src/settings/card_counter_settings.cpp
Elarnon: Using `global.ini` is intentional — I didn't want to touch `SettingsCache` too much (`cache_settings.h` is included in *a lot* of places and touching it triggers slooooooow rebuilds), but I also don't think it's worth having a separate file for these settings.

Not using `SettingsManager` is not, I was not aware it existed. I'm not sure I see the point of the abstraction but I can use it for consistency.

Elarnon:
Updated to adress review and use letter names for counters.

ebbit1q:
would it be an idea to add the color value to the proto command so other players see the same color as you set?

Elarnon:
> would it be an idea to add the color value to the proto command so other players see the same color as you set?

One of the motivations (and the whole reason I picked that up, actually) is related to accessibility/making sure the counters are readable to the person currently using the software: https://github.com/Cockatrice/Cockatrice/issues/2020#issuecomment-1404434469

Letting other players set the color would defeat that purpose (it also opens up some issues with counters potentially changing color when the card changes controller, but that's secondary).

Elarnon:
The way I see it the "kind" of counter (for now it is just an ID, in the future this ID could have metadata associated to say that it corresponds to a +1/+1 counter etc. depending on #1740 ideally) is part of the shared data between players, but the actual look of the counter is part of the theme/local user configuration.

The protocol sends the color and radius of the player counters and I think that's a mistake for the same reason.

ebbit1q:
it sends the radius? where?

Elarnon:
https://github.com/Cockatrice/Cockatrice/blob/77d13090b57d1e5bcb6defc4057f3f9e347d7dc8/common/pb/serverinfo_counter.proto#L8

Set by server here:
https://github.com/Cockatrice/Cockatrice/blob/77d13090b57d1e5bcb6defc4057f3f9e347d7dc8/common/server_counter.cpp#L15

Read by client here:
https://github.com/Cockatrice/Cockatrice/blob/77d13090b57d1e5bcb6defc4057f3f9e347d7dc8/cockatrice/src/game/player/player.cpp#L2900

ebbit1q:
wow, I never knew, that's kinda strange

tooomm:
Hi @Elarnon, I had another look at your changes.
Overall it appears fine and works!

- New submenu for all counters is nice, using letters as well
- I'm not sure though if 6 different counter types are actually needed, maybe 4 would also do?
- Changing all background color (if even intentional?) of the color change dialog window is a bit too much:
image

Elarnon:
> Hi @Elarnon, I had another look at your changes. Overall it appears fine and works!

Thanks!

>     * I'm not sure though if 6 different counter types are actually needed, maybe 4 would also do?

I can do 4 if that's preferred — 6 is just the max that will be supported on legacy clients, anything past 3 will have an empty string as the name.

>     * Changing all background color (if even intentional?) of the color change dialog window is a bit too much:

Uh, that's weird — not intentional at all. It doesn't happen on my system because Qt uses the system color dialog. I guess it's because I create the `QColorDialog` with the button as a parent, so the style on the button gets inherited. I pushed a tentative fix — can you check if it fixes the issue?

tooomm:
>Uh, that's weird — not intentional at all. It doesn't happen on my system because Qt uses the system color dialog. I guess it's because I create the QColorDialog with the button as a parent, so the style on the button gets inherited. I pushed a tentative fix — can you check if it fixes the issue?

Jep, that fixes it!
image

PR #5893, opened 2025-04-29
Remove Fedora 40 build and Add Fedora 42 build

SlightlyCircuitous:

Fedora 40 will go end on life on May 13, 2025, so we will not need to build for it going forward. Fedora 42 was released on April 15, 2025 and will be supported until May 13, 2026, so it would be useful to start building for it instead.

- Removes Fedora 40 from Github build workflow
- Adds Fedora 42 to Github build workflow
- Updates Release Template to reflect above changes

I will test the build from this PR in a Fedora 42 VM and ensure that it installs and works properly.


SlightlyCircuitous:
The .rpm from Github Actions installs correctly on my Fedora 42 VM, Cockatrice opens without incident, and oracle runs appropriately on start up. I can make, save, and load decks, participate in local games, and connect to servers as expected.

tooomm (approved):
Thank you for keeping that updated!

ZeldaZach (approved):

PR #5834, opened 2025-04-15
Visual Deck Editor Base

BruebachL:
## What will change with this Pull Request?
- this
- and this

## Screenshots
https://github.com/user-attachments/assets/d058c565-7ee2-4ab6-b448-7d6a9f51531f

PR #5895, opened 2025-05-01
Add setting for which deck editor tab to open deck in

RickyRister:
## Related Ticket(s)
- Follow-up to #5834

Right now, actions that open a deck in deck editor tab will always open a Visual Deck Editor tab. Some people prefer the classic deck editor.

https://github.com/user-attachments/assets/0d8366ff-94d8-4876-90fe-efd3dd517082

- Added new setting 
  - Added `openDeckInNewTab` method to `TabSupervisor` which calls either the method for adding a visual deck editor tab or classic deck editor tab, depending on settings.
  - Made all open deck signals connect to `openDeckInNewTab` instead of directly connecting to the add tab methods

## Screenshots

Screenshot 2025-04-30 at 11 35 06 PM


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -681,6 +681,10 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     connect(&visualDeckStorageSelectionAnimationCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance(),
             &SettingsCache::setVisualDeckStorageSelectionAnimation);
 
+    openInVisualDeckEditorCheckBox.setChecked(SettingsCache::instance().getOpenInVisualDeckEditor());
+    connect(&openInVisualDeckEditorCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance(),
+            &SettingsCache::setOpenInVisualDeckEditor);
```
Elarnon, line 686: If we just need a boolean yes/no, `QAbstractButton::toggled` is better than `QCheckBox::QT_STATE_CHANGED` (which only exists to also support the indeterminate state).

cockatrice/src/settings/cache_settings.cpp
```diff
@@ -285,6 +285,7 @@ SettingsCache::SettingsCache()
     visualDeckStorageInGame = settings->value("interface/visualdeckstorageingame", true).toBool();
     visualDeckStorageSelectionAnimation =
         settings->value("interface/visualdeckstorageselectionanimation", true).toBool();
+    openInVisualDeckEditor = settings->value("interface/openInVisualDeckEditor", true).toBool();
```
Elarnon, line 288: I'd suggest using something like `"interface/defaultDeckEditor"` which can be either `"VisualDeckEditor"` or `"DeckEditor"` instead of a boolean toggle, it's both future-proof and self-documenting.

Maybe also having a select box (where you can select either "Deck editor" or "Visual deck editor") in the settings, now that I think about it.

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -766,6 +771,7 @@ void UserInterfaceSettingsPage::retranslateUi()
     openDeckInNewTabCheckBox.setText(tr("Open deck in new tab by default"));
     visualDeckStorageInGameCheckBox.setText(tr("Use visual deck storage in game lobby"));
     visualDeckStorageSelectionAnimationCheckBox.setText(tr("Use selection animation for Visual Deck Storage"));
+    openInVisualDeckEditorCheckBox.setText(tr("Open decks in visual deck editor"));
```
Elarnon, line 774: Isn't this just "Use the visual deck editor" at this point?

I did not test the PR, but is there even a way to open the non-visual deck editor when this is checked (or to open the visual deck editor when it's unchecked)?

RickyRister: You can open it through the tab menu. New deck tabs opened from the deck editor tab will also open their respective deck editor tab type.

cockatrice/src/dialogs/dlg_settings.cpp
RickyRister: Huh, do the types work out?

RickyRister:
Made the setting an enum instead of a bool

https://github.com/user-attachments/assets/0f4b9b7f-964f-4db0-a07e-a1e2d80a6bf0

cockatrice/src/dialogs/dlg_settings.cpp
Elarnon: Yup, `QCheckBox` is a `QAbstractButton`. Qt doc is awful as usual, but that is "explicitly" suggested by the doc on [`QCheckBox::checkState`](https://doc.qt.io/qt-6/qcheckbox.html#checkState) which points to [`QAbstractButton::checked`](https://doc.qt.io/qt-6/qabstractbutton.html#checked-prop) whose notifier signal is [`QAbstractButton::toggled`](https://doc.qt.io/qt-6/qabstractbutton.html#toggled).

RickyRister: We might want to update the other usages of `QT_STATE_CHANGED` to `QAbstractButton::toggled` in a future PR then

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -703,13 +700,20 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
                     index == visualDeckStoragePromptForConversionIndexAlways);
             });
 
+    defaultDeckEditorTypeSelector.addItem(""); // these will be set in retranslateUI
+    defaultDeckEditorTypeSelector.addItem("");
+    defaultDeckEditorTypeSelector.setCurrentIndex(SettingsCache::instance().getDefaultDeckEditorType());
+    connect(&defaultDeckEditorTypeSelector, QOverload<int>::of(&QComboBox::currentIndexChanged),
+            &SettingsCache::instance(), &SettingsCache::setDefaultDeckEditorType);
+
```
Elarnon, line 708: This is what we already to elsewhere so I guess it's fine but for reference the proper way to deal with enum values would be to:

- Declare the `DeckEditorType` as a [`Q_ENUM`](https://doc.qt.io/qt-6/qobject.html#Q_ENUM):
  ```c++
  enum DeckEditorType
  {
      // Keep consistent values since they end up in settings
      ClassicDeckEditor = 0,
      VisualDeckEditor = 1
  };
  Q_ENUM(DeckEditorType);
  ```

- Store the value with the `UserData` role in the combobox and load the index from the `UserData`, not the index:
  ```c++
  // Title text will be set in retranslateUi
  defaultDeckEditorTypeSelector.addItem("", TabSupervisor::ClassicDeckEditor);
  defaultDeckEditorTypeSelector.addItem("", TabSupervisor::VisualDeckEditor);
  defaultDeckEditorTypeSelector.setCurrentIndex(
      defaultDeckEditorTypeSelector.findData(SettingsCache::instance().getDefaultDeckEditorType()));
  connect(&defaultDeckEditorTypeSelector, &QComboBox::currentIndexChanged, this, [this](int index) {
      SettingsCache::instance().setDefaultDeckEditorType(defaultDeckEditorTypeSelector.itemData(index).toInt());
  });
  ```

- Set text based on the `UserData` in `retranslateUi`:
  ```c++
    for (int i = 0; i < defaultDeckEditorTypeSelector.count(); ++i) {
        switch (defaultDeckEditorTypeSelector.itemData(i).value<TabSupervisor::DeckEditorType>()) {
            case TabSupervisor::ClassicDeckEditor:
                defaultDeckEditorTypeSelector.setItemText(i, tr("Classic Deck Editor"));
                break;

            case TabSupervisor::VisualDeckEditor:
                defaultDeckEditorTypeSelector.setItemText(i, tr("Visual Deck Editor"));
                break;
        }
    }
  ```

cockatrice/src/client/tabs/tab_supervisor.h
```diff
@@ -75,6 +75,14 @@ class CloseButton : public QAbstractButton
 class TabSupervisor : public QTabWidget
 {
     Q_OBJECT
+
+public:
+    enum DeckEditorType
+    {
+        ClassicDeckEditor,
+        VisualDeckEditor
+    };
```
Elarnon, line 84: We could also define this type in the settings directly, so that we don't have to go through opaque `int` types for `getDefaultDeckEditorType` / `setDefaultDeckEditorType`.

Elarnon (approved):
Some additional robustness/cleanliness comments but since the settings are already very messy I'm personally fine with the code as is.

cockatrice/src/client/tabs/tab_supervisor.h
RickyRister: That was what I was originally thinking too, but then that we don't define enums anywhere else in `SettingsCache` so I felt that it would be inconsistent to start doing it now

cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -699,12 +700,20 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
                     index == visualDeckStoragePromptForConversionIndexAlways);
             });
 
+    defaultDeckEditorTypeSelector.addItem(""); // these will be set in retranslateUI
+    defaultDeckEditorTypeSelector.addItem("");
+    defaultDeckEditorTypeSelector.setCurrentIndex(SettingsCache::instance().getDefaultDeckEditorType());
+    connect(&defaultDeckEditorTypeSelector, QOverload<int>::of(&QComboBox::currentIndexChanged),
```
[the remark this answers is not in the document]
Elarnon: Ah but yes of course there is also `QComboBox::currentIndexChanged(const QString&)` how obvious :facepalm: Thanks for the clarification.

ZeldaZach (approved):

PR #5896, opened 2025-05-01
fix: Prevent dragged cards getting stuck

Elarnon:
Update the position of the card even if it is not above any zone.

Dragged cards' position does not get updated when the cursor is not over a valid zone, making it look like the card is "stuck".

Cards keep moving even when the cursor is not over a zone.

## Screenshots

Before:

[Screencast From 2025-05-01 21-21-09.webm](https://github.com/user-attachments/assets/cd6cf831-7453-47da-9632-9905790f04a9)

After:

[Screencast From 2025-05-01 21-21-57.webm](https://github.com/user-attachments/assets/0f2925f2-de23-4691-886b-5e7e49d734cd)

### DIFF
--- a/cockatrice/src/game/board/card_drag_item.cpp
+++ b/cockatrice/src/game/board/card_drag_item.cpp
@@ -47,8 +47,19 @@ void CardDragItem::updatePosition(const QPointF &cursorScenePos)
         cursorZone = zoneViewZone;
     else if (cardZone)
         cursorZone = cardZone;
-    if (!cursorZone)
+    if (!cursorZone) {
+        // Temporary fix: avoid the cards getting stuck visually when not over
+        // any zone.
+        QPointF newPos = cursorScenePos - hotSpot;
+
+        if (newPos != pos()) {
+            for (int i = 0; i < childDrags.size(); i++)
+                childDrags[i]->setPos(newPos + childDrags[i]->getHotSpot());
+            setPos(newPos);
+        }
+
         return;
+    }
     currentZone = cursorZone;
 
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/board/card_drag_item.cpp
```diff
@@ -47,8 +47,19 @@ void CardDragItem::updatePosition(const QPointF &cursorScenePos)
         cursorZone = zoneViewZone;
     else if (cardZone)
         cursorZone = cardZone;
-    if (!cursorZone)
+    if (!cursorZone) {
+        // Temporary fix: avoid the cards getting stuck visually when not over
```
ebbit1q, line 51: what would the permanent fix be?

Elarnon: I'm rewriting the whole drag-and-drop system for #4974 so that's what I had in mind. The "temporary" doesn't really make sense in the context of the code, I'll drop it.

RickyRister:
Oh nice, this fixes a problem caused by #5886 where by force of habit you try to get rid of the token by dragging it to the left and it gets stuck on the hand size indicator and causes it to remain on the stack.

Elarnon:
> Oh nice, this fixes a problem caused by #5886 where by force of habit you try to get rid of the token by dragging it to the left and it gets stuck on the hand size indicator and causes it to remain on the stack.

Actually, it didn't, because we still did not update the `currentZone`. Fixed in follow-up commit.

ZeldaZach (approved):

PR #5898, opened 2025-05-03
Move card view window's searchbar to bottom

RickyRister:
## Related Ticket(s)
- Follow-up to #5791

Even though putting the search bar at the top of the card view window's menus looks more pleasing aesthetically, I think moving the search bar to the bottom is better for users, since there's less distance between typing and selecting.

## Screenshots

Before

Screenshot 2025-05-02 at 10 56 35 PM

After

Screenshot 2025-05-02 at 10 56 18 PM


RickyRister:
No longer necessary after #5906

PR #5899, opened 2025-05-03
[VDS] Add setting to show deck filepath in tooltip

RickyRister:

Sometimes it's hard to figure out which exact file you're opening while in the VDS.

- Also set the `BannerCardDisplayWidget`'s tooltip to the deck's filepath whenever the overlay text is refreshed.
- This means a tooltip with the deck filepath will pop up if you hover over a banner card for a bit

https://github.com/user-attachments/assets/86dcd185-8fc3-495b-942e-0785241e95e6


BruebachL:
I would add a setting for it or constrain it to the textbox instead (Personally, I would get annoyed with it since I do not care about this information.)

RickyRister:
Made a `deck tooltip` setting

https://github.com/user-attachments/assets/a91985c2-8311-4340-8e88-c7866a9d3b06

PR #5902, opened 2025-05-03
ci: temporarily remove run-vcpkg step

brunoalr:
## Related Ticket(s)
- Workaround for https://github.com/lukka/run-vcpkg/issues/243

Windows builds are taking around 1h to complete due to an issue with caching in run-vcpkg.

- good: windows builds are 20min faster
- bad: the step should be re-enabled when the upstream issue is fixed

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -387,7 +387,9 @@ jobs:
           tools: ${{matrix.qt_tools}}
           modules: ${{matrix.qt_modules}}
 
-      - name: Run vcpkg
+      # TODO: re-enable when https://github.com/lukka/run-vcpkg/issues/243 is fixed
+      - if: false
+        name: Run vcpkg (disabled)
         uses: lukka/run-vcpkg@v11
         with:
           runVcpkgInstall: true


tooomm (approved):
I did mention the issue in #5883 already, as simply updating the cache action is not sufficient to fix vcpkg.

>No solution for caching in the run-vcpkg action yet. GHA binary caching was used there, which will go away. A respective PR to remove the used internal APIs is already open in the vcpkg-tools repo. There is no obvious replacement or path yet.
This step takes ~20min to complete now, and cmake uses another ~40min to build the app (and dependencies) because of the missing cache and running vcpkg install again.

There is no easy solution yet, apparently NuGet should be used and cache needs to be managed manually unless somebody builds an action around it or it gets integrated into e.g. `run-vcpkg`.

I'm fine in disabling this step for now as vcpkg install is handled by cmake in the follow up step. Let's save that 20min on Windows each build in the meantime.

Thanks @brunoalr!

brunoalr:
in the meantime, I also played a bit with the workaround mentioned [here](https://github.com/lukka/run-cmake/issues/152#issuecomment-2833729132) (i.e., manually dealing with caching ourselves) and it looks like it works: Windows builds are down to ~20min when there's a cache hit. How fast was it before the caching issues? 

https://github.com/brunoalr/Cockatrice/pull/3/files

tooomm:
That looks promising, thanks for giving it a shot!

<br>

Two notes:
- In the post you linked, there is also a [second reference using a sub-path for the save part](https://github.com/microsoft/vcpkg/issues/45073#issuecomment-2828284201). Is that actually having an impact?
  ```
  - name: Save vcpkg cache
    uses: actions/cache/save@v4
    with:
      path: ~/.cache/vcpkg/archives
      key: some-build-specific-key
  ```

- Did you also see [this comment](https://github.com/microsoft/vcpkg-tool/pull/1662#issuecomment-2840245329) where somebody drafted a first reusable action?

tooomm:
>(...) it works: Windows builds are down to ~20min when there's a cache hit. How fast was it before the caching issues?

Before the deprecation of older GHA cache versions, it was very similar. See this older run https://github.com/Cockatrice/Cockatrice/actions/runs/14438217800:
image
image

It used to be considerable faster at one point (10-12min), but Windows builds in particular slowed down over time.
A reason for the long compilation times is that compiler cache is not yet utilized.

tooomm:
There is now also an action that manages caching via GHA: https://github.com/TAServers/vcpkg-cache

@brunoalr Took a stap at this and opened a PR, see #5933.
I think it's a bit simpler than managing the cache ourselves/manually.
And in particular, any adjustments due to future changes are hopefully handled in the action directly.

brunoalr:
Hi there! somehow I completely missed the older message, I will reply inline.

tl;dr: I do agree that leaving the caching details up to a separate workflow is better than managing it manually.

> That looks promising, thanks for giving it a shot!
> 
> 
> Two notes:
> 
> * In the post you linked, there is also a [second reference using a sub-path for the save part](https://github.com/microsoft/vcpkg/issues/45073#issuecomment-2828284201). Is that actually having an impact?
>   ```
>   - name: Save vcpkg cache
>     uses: actions/cache/save@v4
>     with:
>       path: ~/.cache/vcpkg/archives
>       key: some-build-specific-key
>   ```

I am not entirely sure, but I guess the only thing that matters is referencing the same path when saving and when restoring the caches.

> * Did you also see [this comment](https://github.com/microsoft/vcpkg-tool/pull/1662#issuecomment-2840245329) where somebody drafted a first reusable action for the caching approach via NuGet?
I actually saw the discussion about using NuGet, but didn't notice there was a new reusable action using it already 😅 
>   Did you opt for the other solution for a particular reason?
Not really, it was mostly because it looked simple enough IMO and I wanted to test it quickly :)

tooomm:
>tl;dr: I do agree that leaving the caching details up to a separate workflow is better than managing it manually.

The action just emerged recently with the need of a new caching solution.
And the maintainer is reacting super quickly to feedback and fixed a few first issues, so far it looks solid.

Going with this is probably better and easier than exploring NuGet.

PR #5905, opened 2025-05-04
[VDS] Refactor: move quick settings to separate class

RickyRister:
## Related Ticket(s)
- Requires #5903 to be merged first

Currently, the widgets inside the VDS quick settings dropdown menu are all managed in `VisualDeckStorageWidget`. This is starting to become unwieldy as we add more settings to the quick settings menu. 

It would be cleaner to move the widgets to do with the quick settings into their own class.

- Created `VisualDeckStorageQuickSettingsWidget` class
  - extends `SettingsButtonWidget`, since that was what we were putting the setting widgets in before.
  - has getters for the setting values as well as signals for settings value change. 
- Moved all the settings widget creation from `VisualDeckStorageWidget` constructor to `VisualDeckStorageQuickSettingsWidget` constructor
- Added a getter for the settings widget to `VisualDeckStorageWidget`, so that other classes can still access the values from the settings that way.


ZeldaZach (approved):

ZeldaZach:
Conflicts

PR #5919, opened 2025-05-05
[VDE] Deck analytics visibility

### LINKED ISSUE #5908: UI issues with recent changes in GDE, VDS & VDE
All screenshots are from this version:
```
Client Version: 2.11.0-beta.4 (2025-05-02)
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_DE
Install Mode: Standard
```
 
- GDE, VDE
  `Banner Card` label next to the card drop down selection disappeared
  >This change hides the `Banner Card` label next to the card drop down selection.
  >
  >_Originally posted by @tooomm in 
   https://github.com/Cockatrice/Cockatrice/issues/5857#issuecomment-2825265590_
    
  Image

- Other recent changes in VDS and VDE that introduced buttons, left out the labels for the same. In particular if they have only an icon, an explanatory hover label is very much needed for users from an UI/UX perspective. I think this is also important from an accessibility perspective as screen readers would utilize this information as well.
    - VDS
    Image

    - GDE
    Image

    - VDE, Deck View
    Image

    - VDE, Deck View
    Image
    The card size slider is sitting in the new setting cogwheel in VDS, why place him in different location here?

    - VDE, Database Display
    Image
    I've no clue what the `150` value is about even.
    Also note the horrible UI around the card types.
    In general, I do not get how to use this tab/view as it shows no cards for me when briefly playing around.

    - VDE, Deck Analytics
    Image
    There is no scrollbar and I can not even see the `Mana Base` at the bottom.
    Please also note the `Mana Curve` bars, the black color is matching the card count number for the 4th one and can not be read therefore.

    - VDE, Sample Hand
    Image
    This one has a scroll bar 👍 
    But ideally it would automatically scale to show all cards at once.
    At least I would request the same card size zoom slider than known from other places. A number value input box to change the number of cards to redraw (hand size might be different in other games, or if users want to simulate a mulligan for example).
    
        The redraw? button above the card view has no label at all (also no hover hint).

BruebachL:
## Related Ticket(s)
- Closes #5908 with #5916 and #5917

The bars for the mana curve widget are colored black which causes visibility issues due to the black font. There is also apparently no scrollbar. I can't confirm this since the widgets resize correctly for me.

- Change bar color to grey
- Add a preemptive scrollArea

## Screenshots
image

### DIFF
--- a/cockatrice/src/client/ui/widgets/deck_analytics/deck_analytics_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/deck_analytics/deck_analytics_widget.cpp
@@ -6,14 +6,25 @@ DeckAnalyticsWidget::DeckAnalyticsWidget(QWidget *parent, DeckListModel *_deckLi
     mainLayout = new QVBoxLayout();
     setLayout(mainLayout);
 
+    scrollArea = new QScrollArea(this);
+    scrollArea->setVerticalScrollBarPolicy(Qt::ScrollBarAsNeeded);
+    scrollArea->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    scrollArea->setWidgetResizable(true);
+    mainLayout->addWidget(scrollArea);
+
+    container = new QWidget(scrollArea);
+    containerLayout = new QVBoxLayout(container);
+    container->setLayout(containerLayout);
+    scrollArea->setWidget(container);
+
     manaCurveWidget = new ManaCurveWidget(this, deckListModel);
-    mainLayout->addWidget(manaCurveWidget);
+    containerLayout->addWidget(manaCurveWidget);
 
     manaDevotionWidget = new ManaDevotionWidget(this, deckListModel);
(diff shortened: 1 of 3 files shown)


ZeldaZach (approved):

tooomm:
On `2.11.0-beta.10` (Win 11) I get this result:

image

Fixed:
- Bar value labels are now readable

Still problematic:
- No scrollbar and I can not see `Mana Base`. In the `Visual Deck View`, `Visual Database Display` and `Sample Hand` tabs scrolling works.
  When playing around, I saw that the widget is somehow faulty as it does not show a checkmark next to either visible or floating:
  image
  Clicking one of the options crashes Cockatrice.
- The descriptive label below each columns are in white (also seen in your screenshot above).
  With the default theme that just does not work well, I did not even spot this when opening the initial issue as it is just very hard to notice.
  In a custom dark theme as yours, it looks good though! Maybe apply a background (circle?).

PR #5923, opened 2025-05-06
fix: fix qt5 builds on macos 15

brunoalr:
## Related Ticket(s)
N/A

Although this config is not built on CI, while trying to compile locally, the build failed due to warnings and -Werror. Some qt functions were actually deprecated (but not removed) before version 6.0.0 and clang (righfully) complains about comparison between different types of enums.

Cockatrice will build with qt5 on macos15, Apple clang version 17.0.0 (clang-1700.0.13.3) locally.

## Screenshots
N/A


ZeldaZach (approved):

ZeldaZach:
Thanks for adding this!

PR #5928, opened 2025-05-09
[GDE] Card Price widget using Scryfall API

### LINKED ISSUE #2927: use scryfall api to get prices
At one point card price support stopped to work, and we removed it month later.

There is the possibility to use the Scryfall API to get card prices instead.

Check the examples on their site:
- https://scryfall.com/docs/api/cards/collector (https://api.scryfall.com/cards/xln/96)
>  "usd": "0.12",
>  "tix": "0.01",
>  "eur": "0.05",

- or https://scryfall.com/docs/api/cards/multiverse (https://api.scryfall.com/cards/multiverse/409574)
> "usd": "53.37",
>  "tix": "1.06",
>  "eur": "41.15",

<br>

How to implement it and how it would be most useful should be discussed.

For example it might be less useful to have it automatically for all cards in the card database and spam the card list with information many might never use and stress the api for no reason.

But as a right click option on single cards or from within a menu where you can paste a list of names etc. it could be nice.
I like the idea to have it for complete deck lists for example. This way users can check how expensive a certain deck is, as well as create deck files for their special cards and monitor their value.

BruebachL:
## Related Ticket(s)
- Closes #2927

- Add a new widget to fetch card prices from scryfall
- Toggle option to include lands
- Toggle option to fetch every printing we know about and compare min/max prices
- Lists the set the card is from
- USD/EUR ComboBox

## Screenshots
image
image

### DIFF
--- a/cockatrice/src/client/ui/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
@@ -1,5 +1,6 @@
 #include "deck_editor_deck_dock_widget.h"
 
+#include "../../../../../../../../../../../mnt/games-hdd/coding/CLionProjects/Cockatrice/cockatrice/src/dialogs/dlg_get_card_prices.h"
 #include "../../../../game/cards/card_database_manager.h"
 #include "../../../../settings/cache_settings.h"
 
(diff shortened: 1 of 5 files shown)


cockatrice/src/client/ui/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -144,6 +145,14 @@ void DeckEditorDeckDockWidget::createDeckDock()
 
     upperLayout->addWidget(deckTagsDisplayWidget, 3, 1);
 
+    cardPricesButton = new QPushButton(this);
```
RickyRister, line 148: Might make more sense to put the "open card prices" action into the Deck Editor menu instead of as a button that takes up dock space. We already have actions such as `Print Deck` and `Send deck to online service` and this feels like something in the same vein as those.

BruebachL: Moved above Print Deck.

ZeldaZach:
So funnily enough, Cockatrice used to have a pricing system built right into the deck list. See https://github.com/Cockatrice/Cockatrice/pull/2670 for more insights on what was removed. It might make more sense to go back to the model we had in the past of adding an additional column to the decklist to show the price instead of a separate window.

PR #5935, opened 2025-05-10
increase cache size for ccache on linux ci

ebbit1q:
it seems like linux builds have been taking longer while the ccache has been full, because we have more space on our github actions now compared to when this limit was put I see no reason not to try if increasing this limit makes it run faster.


tooomm (approved):
Looks good, build time for linux distros is down to 1-3 minutes again.
Cache size increased for some up to ~300MB.


> we have more space on our github actions now compared to when this limit was put I see no reason not to try if increasing this limit makes it run faster.

While the maximum cache storage doubled 4 years ago (
https://github.blog/changelog/2021-11-22-github-actions-cache-size-is-now-increased-to-10gb-per-repository/) to the noted 10Gi, we are still constantly hitting the limit and the oldest cache available is only 1 day old.

The linux builds in docker do populate new caches very often and take up a lot of space, see https://github.com/Cockatrice/Cockatrice/actions/caches

Issue #3781, opened 2019-07-14
Travis: Infinite growing cache issue


tooomm:
## Data on builds / caches:
Travis build: https://travis-ci.org/Cockatrice/Cockatrice/builds/558332611
Corresponding Travis config file: https://github.com/Cockatrice/Cockatrice/blob/a8b34d51a5a2735ec745defc829f9ee1b13e7c46/.travis.yml
-  Job "**Ubuntu Bionic (Debug + Tests)**"
    Run on `Ubuntu 14.04.5 LTS` via Docker
    - First run without cache: 10 min 9 sec (Cache size: 87.95MB)

    - Second run with cache:  2 min 26 sec (Cache size: 90.41MB [+2,5MB])
      - Changes in cache:
      ```
      store build cache
      schanges detected (content changed, file is created, or file is deleted):\n/home/travis/UbuntuBionic/.ccache/0/2/cdd5b834509bcccdee3f80c26a6ea1-20841.manifest
      /home/travis/UbuntuBionic/.ccache/0/d/2dfc0816aa886969897adff11242ab-343687.manifest
      /home/travis/UbuntuBionic/.ccache/0/stats
      /home/travis/UbuntuBionic/.ccache/1/stats
      /home/travis/UbuntuBionic/.ccache/2/stats
      /home/travis/UbuntuBionic/.ccache/3/0/aaeb667abc9796efae01d3622c27e7-21604937.o
      /home/travis/UbuntuBionic/.ccache/3/stats
      /home/travis/UbuntuBionic/.ccache/4/a/d4c76bd143b5b4819d46892bed1fae-341938.o
      /home/travis/UbuntuBionic/.ccache/4/stats
      /home/travis/UbuntuBionic/.ccache/5/stats
      /home/travis/UbuntuBionic/.ccache/6/stats
      /home/travis/UbuntuBionic/.ccache/7/stats
      /home/travis/UbuntuBionic/.ccache/8/stats
      /home/travis/UbuntuBionic/.ccache/9/2/6add1ab813728b7024e185821d61f7-21642818.manifest
      /home/travis/UbuntuBionic/.ccache/9/stats
      /home/travis/UbuntuBionic/.ccache/a/stats
      /home/travis/UbuntuBionic/.ccache/b/stats
      /home/travis/UbuntuBionic/.ccache/c/stats
      /home/travis/UbuntuBionic/.ccache/d/stat\n...
      changes detected, packing new archive
[...]

PR #5938, opened 2025-05-11
CI: Tweak ccache for Linux builds

tooomm:
## Related Ticket(s)
- Related https://github.com/Cockatrice/Cockatrice/pull/5935
- Linking #3781

~75% of all caches (150) stored right now are from Linux builds.
See duplicates here for example from the same branch just with different timestamps:
image

Due to the constant recreation with a new name (changing timestamp), even if no actual changes happened, we are always over the 10GB limit and the oldest cache overall we can keep is only 19h old.
image

After the increase in allowed cache size (#5935):
image

Existing caches of the repo can be reviewed here: https://github.com/Cockatrice/Cockatrice/actions/caches

- Split cache action in `restore` & `save` parts wrapping around the build steps
- Only execute `save` to upload new caches when run on master to reduce amount of similar caches stored in the repo. As all PR's are usually based off master, the cache from there should still be useful.
- Clearer cache naming `ccache-Debian12-[timestamp]`

<br>

I'm sure that's not perfect either and there is probably a better way.
Instead including a timestamp, we could add the source branch in the primary key (`key`) to not have duplicated caches for the "same" change and scope the cache better. Or removing seconds, minutes and maybe also hours from the timestamp to update only once a day is an improvement?

In general, it appears a good strategy to always have a valid cache on the default branch as that would be reused by all branches based off of it. Currently that is not guaranteed as the caches are replaced so often/quickly and only preserved for 1-2 days.
When just going with updating the cache for the default (master) branch, we can probably also remove the timestamp?

Matching keys logic: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

Hope somebody can chip in to the topic and has experience in reusing compiler caches and which caching strategy to use in GitHub CI.

<br>

macOS and Windows builds do not utilize compiler cache yet and run considerably longer for that reason: #3059

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -139,11 +139,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Generate cache timestamp
-        id: cache_timestamp
-        shell: bash
-        run: echo "timestamp=$(date -u '+%Y%m%d%H%M%S')" >> "$GITHUB_OUTPUT"
-
       - name: Build ${{matrix.distro}} ${{matrix.version}} Docker image
         shell: bash
         run: source .ci/docker.sh --build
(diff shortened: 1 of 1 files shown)


tooomm:
I implemented the mentioned change to use branch name over timestamp as key to reduce the amount of existing caches even further while keeping them relevant.

ZeldaZach (approved):

PR #5942, opened 2025-05-14
Refactor TabGame: inline some QLayout and QWidget class fields

RickyRister:

In `TabGame`, some nested `QLayout`s and `QWidget`s are saved as class fields, even though they are only used during widget creation and never referenced again.

- Removed some `QLayout` and `QWidget` class fields that are only used during dock creation.
  - Instead, just directly declare the layout or widget in the method that uses it

This is part of an effort to refactor `TabGame` into separate classes for replays and live games.


BruebachL:
Will this not cause warnings á la "Allocated memory is leaked"?

RickyRister:
Yeah, it will. I just try to ignore those warnings.

ZeldaZach (approved):

PR #5952, opened 2025-05-20
CI: use ccache on macOS

brunoalr:
## Related Ticket(s)
N/A

macOS builds on CI currently don't use ccache, missing a build time reduction oportunity

- use ccache on macOS builds

## Screenshots

N/A

## Extra considerations
This leads to some duplication inside desktop-build.yml, so I have been thinking about unifying all build jobs into a single one with OS-specific steps instead of having multiple jobs. If this is desirable, I could take a look at it.


.github/workflows/desktop-build.yml
```diff
@@ -235,6 +235,7 @@ jobs:
     runs-on: ${{matrix.os}}
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
     env:
+      CCACHE_DIR: /tmp/${{matrix.os}}-${{matrix.type}}-cache
```
brunoalr, line 238: reference: https://ccache.dev/manual/latest.html#_command_line_options 

ideally the Linux job should be updated too, but I didn't want to expand the scope of this PR

tooomm (approved):
Same logic as for the Linux builds. 👍

You also separated caches between Debug and Release builds! 
Is that actually something we should be doing for Linux as well? Will that effect quality, relevance and hit rates for the caches?

Let's try it on master.

tooomm:
Works great, @brunoalr. Build times are considerably reduced!

macOS build on master before:
image

With new ccache on macOS:
image

brunoalr:
> Same logic as for the Linux builds. 👍 Let's try it on master.
> 
> You also separated caches between Debug and Release builds! Is that actually something we should be doing for Linux as well? Will that effect quality, relevance and hit rates for the caches?

Ahh yes, I forgot to check that on Linux. I *think* the hit rates would be improved for the caches, but I can only be sure when testing... I can take a shot at it later (unless you want to have some fun with it yourself)

brunoalr:
> Works great, @brunoalr. Build times are considerably reduced!
> 
> macOS build on master before: image
> 
> With new ccache on macOS: image

I still want to figure out why cpack takes so long and/or fails on macOS 13

.github/workflows/desktop-build.yml
tooomm: I figured that the interaction with compile.sh and docker.sh scripts makes the setup a bit more complex and one needs to wrap his head around at first.
@ebbit1q's idea with the script was probably to have a simple and clean workflow configuration file for CI, and do the magic in the scripts.

There is some other cleanup in https://github.com/Cockatrice/Cockatrice/pull/5959 btw.

I also saw that we have Ninja now hardcoded in the docker.sh script. Ideally that should react on the configuration we have in the workflow file as well?

tooomm:
>Some (maybe many?) of the specifics can be abstracted: e.g. compilers via cmake, ccache setup via another [github action](https://github.com/hendrikmuhs/ccache-action).
>
>As an example, here's the build workflow for another project that I maintain: https://github.com/OpenImageDebugger/OpenImageDebugger/blob/main/.github/workflows/build.yml
>
>For cockatrice, there would be more `if:`s around, of course.
>
> I am willing to give it a go :D But of course I don't want to solve a non-existing problem neither make the build more complex... Let's see if the result of the attempt makes sense

I still did not have a look there, but was planning to do so after other fixes are in place.

Guess there is most value in focusing on other improvements and get Ninja and uniform caching for all OS working first. 👍 

>> You also separated caches between Debug and Release builds! Is that actually something we should be doing for Linux as well? Will that effect quality, relevance and hit rates for the caches?
>
>Ahh yes, I forgot to check that on Linux. I *think* the hit rates would be improved for the caches, but I can only be sure when testing... I can take a shot at it later (unless you want to have some fun with it yourself)

I looked into some smaller general improvements and still try to figure Ninja on Windows out...
If you would like to look into dedicated caches for Debug/Release builds already that would be great!

> I still want to figure out why cpack takes so long and/or fails on macOS 13

Failing on macOS 13 runner is a known issue, see https://github.com/actions/runner-images/issues/7522.
I realized that we have already some code that tries to circumvent it:
https://github.com/Cockatrice/Cockatrice/blob/cee67f430108baecadfab691fd65cf1a2392dd66/.ci/compile.sh#L173-L177
At the end we will just remove macOS 13 as it will be EOL in a few months from now.

PR #5807, opened 2025-04-06
feat: build and release docker images using github actions and container registry

maybeanerd:
## Related Ticket(s)
- Fixes #4375 

## Short roundup of the initial problem
There are no officially published images for the server. This makes setting up a server using docker more painful than it needs to be

## What will change with this Pull Request?
a new pipeline using github actions was added. this pipeline does two things:
1. on pull requests, builds the image without publishing. This makes sure the dockerfile still can be built. Example run: https://github.com/maybeanerd/cockatrice/actions/runs/14296665254/job/40064716966
2. on tags that contain "Release", builds the image and releases it as a package of this repository on the github container registry. this allows others to easily pull and use the images. New images will be tagged with the git tag (so e.g. Cockatrice/Cockatrice:2025-04-03-Release-2.10.2 and Cockatrice/Cockatrice:latest ) . Example run: https://github.com/maybeanerd/cockatrice/actions/runs/14296614709/job/40064609872

Builds are made for both amd64 and arm64 architecture, tagging the images respectively.

### Notes and considerations

_Naming after the repository makes GitHub know its related to this repo. This does mean however, that the image is not called "servatrice", and instead cockatrice. If this is an issue, I think we could hardcode the name, but then GitHub would not list it in the "packages" section of the repository automatically, AFAIK. not entirely sure what would be necessary for that to happen, as that's something I didn't need to do in past implementations._
**See next comment in this PR**

The pipeline seems to take quite some time. I hope caching of the containers layers will speed this up in the future, but it might be worth considering to not run them in PRs. I personally would like to ensure the image always builds successfully, but that's a tradeoff one could make.

it could probably also be enriched to release beta images, but I wanted to first get something working out without doing too much at once.

## Screenshots
Image how it will be listed in the registry, linked to the repo and with version tags
Screenshot 2025-04-07 at 00 17 17

Packages section in the repos main page
Screenshot 2025-04-07 at 00 18 40

PR #5959, opened 2025-05-25
CI: Some cleanup

tooomm:
- Max ccache size was not set on macOS
- Use workspace directory over tmp/ to avoid unexpected side effects
- Renamed a variable
- Let script do configuration and only give it test parameter to stay uniform
- Some if conditions were never matching to their job
- Reorder docker image build step after cache restore
- Pass `CMAKE_GENERATOR` from workflow file to Docker instead hardcoding "Ninja"

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -125,11 +125,11 @@ jobs:
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
     env:
       NAME: ${{matrix.distro}}${{matrix.version}}
-      CACHE: ${{github.workspace}}/.ccache/${{matrix.distro}}${{matrix.version}}
-      CCACHE_COMPILERCHECK: content
+      CACHE: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}   # directory for caching docker image and ccache
       # Cache size over the entire repo is 10Gi:
       # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
       CCACHE_SIZE: 500M
+      CCACHE_COMPILERCHECK: content
       CMAKE_GENERATOR: 'Ninja'
 
     steps:
@@ -237,8 +237,8 @@ jobs:
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
     env:
       CCACHE_DIR: ${{github.workspace}}/.ccache/${{matrix.os}}-${{matrix.type}}
-      CCACHE_COMPILERCHECK: content
       CCACHE_SIZE: 500M
+      CCACHE_COMPILERCHECK: content
       DEVELOPER_DIR:
         /Applications/Xcode_${{matrix.xcode}}.app/Contents/Developer
       CMAKE_GENERATOR: 'Ninja'
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -129,7 +129,6 @@ jobs:
       # Cache size over the entire repo is 10Gi:
       # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
       CCACHE_SIZE: 500M
-      CCACHE_COMPILERCHECK: content
       CMAKE_GENERATOR: 'Ninja'
 
     steps:
@@ -238,7 +237,6 @@ jobs:
     env:
       CCACHE_DIR: ${{github.workspace}}/.ccache/${{matrix.os}}-${{matrix.type}}
       CCACHE_SIZE: 500M
-      CCACHE_COMPILERCHECK: content
       DEVELOPER_DIR:
         /Applications/Xcode_${{matrix.xcode}}.app/Contents/Developer
       CMAKE_GENERATOR: 'Ninja'
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -125,7 +125,8 @@ jobs:
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
     env:
       NAME: ${{matrix.distro}}${{matrix.version}}
-      CACHE: /tmp/${{matrix.distro}}${{matrix.version}}-cache    # ${{runner.temp}} does not work?
+      CACHE: ${{github.workspace}}/.ccache/${{matrix.distro}}${{matrix.version}}
+      CCACHE_COMPILERCHECK: content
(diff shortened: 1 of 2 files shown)


tooomm:
What do you think @brunoalr?

.github/workflows/desktop-build.yml
```diff
@@ -125,20 +125,17 @@ jobs:
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
     env:
       NAME: ${{matrix.distro}}${{matrix.version}}
-      CACHE: /tmp/${{matrix.distro}}${{matrix.version}}-cache    # ${{runner.temp}} does not work?
+      CACHE: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}   # directory for caching docker image and ccache
```
brunoalr, line 128: nice!

.github/workflows/desktop-build.yml
```diff
@@ -125,20 +125,17 @@ jobs:
...
       # Cache size over the entire repo is 10Gi:
       # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
       CCACHE_SIZE: 500M
+      CCACHE_COMPILERCHECK: content
```
brunoalr: IMO, it would be nice to add an inline comment about what is this for and why it is needed

.github/workflows/desktop-build.yml
```diff
@@ -165,11 +166,11 @@ jobs:
         env:
           BUILD_DIR: build
           SUFFIX: '-${{matrix.distro}}${{matrix.version}}'
-          type: '${{matrix.package}}'
+          package: '${{matrix.package}}'
```
brunoalr, line 168: also nice!

.github/workflows/desktop-build.yml
```diff
@@ -149,6 +146,10 @@ jobs:
           key: ccache-${{matrix.distro}}${{matrix.version}}-${{env.BRANCH_NAME}}
           restore-keys: ccache-${{matrix.distro}}${{matrix.version}}-
 
+      - name: Build ${{matrix.distro}} ${{matrix.version}} Docker image
```
brunoalr, line 148: tl;dr: looks good to me, but we probably need to look deeper into [docker image building + GHA cache] VS [docker image pulling].

---

longer version:
this reorder does make sense, but I think I realized that the docker images actually haven't been cached - IIRC the docker.sh script expects other flags like `--save` or something like that in order to actually generate a .tar.gz of the docker image. I need to confirm it, tho - probably later this week

alternatively, in a separate PR, we could also change the docker building logic to be a separate GHA workflow that triggers only when a dockerfile is changed and publishes the image to ghcr.io (github container registry) instead of caching docker images and ccache files together using GHA caches. GHA has first-class support for docker images in jobs - from my point of view this would be a better separation of concerns and would be easier to maintain

brunoalr:
> What do you think @brunoalr?

the PR looks good to me - IMO it can be merged already.
there are some possible issues/optimization opportunities with docker image building, tho (not introduced by this PR)

.github/workflows/desktop-build.yml
tooomm: I did revert this change actually.

It seems to be rather an issue when using Android NDK and misleading suggestion in React Native docs. See also e.g. https://github.com/hendrikmuhs/ccache-action/issues/94#issuecomment-2286488872

brunoalr (approved):

ZeldaZach (approved):

.github/workflows/desktop-build.yml
tooomm: @brunoalr Never managed to get back to this comment of yours. 🙈 

I've actually been working on the Token and Webpage repo of the project in the meantime, and added attestation support for Cockatrice releases.

When tinkering with Ninja and CMake before, I realized that our compilation setup is not trivial or intuitive... stumbled upon those get/save options in the docker script used by CI as well.
They are not used right now (and never were afaik), the compilation base images are each time build from scratch... correct. Definitely something that could be improved!

>alternatively, in a separate PR, we could also change the docker building logic to be a separate GHA workflow that triggers only when a dockerfile is changed and publishes the image to ghcr.io (github container registry) instead of caching docker images and ccache files together using GHA caches. GHA has first-class support for docker images in jobs - from my point of view this would be a better separation of concerns and would be easier to maintain

That sounds great and we would not waste precious GitHub Actions Cache space.
Check out #5807 where we just recently added a new [docker-release workflow](https://github.com/maybeanerd/cockatrice/blob/master/.github/workflows/docker-release.yml) to build an official image for `Servatrice` and upload it to GHCR for easy consumption.
Since then no new release happened to actually deploy it. 😄 

Did you had a closer look at those options already?

brunoalr: hey @tooomm no problem! I have been feeling a bit down lately, so haven't been programming much :(

> When tinkering with Ninja and CMake before, I realized that our compilation setup is not trivial or intuitive... stumbled upon those get/save options in the docker script used by CI as well.
They are not used right now (and never were afaik), the compilation base images are each time build from scratch... correct. Definitely something that could be improved!

Yeah, it would take some time to untangle it indeed D: Trickiest part is to understand why some decisions were taken...

> That sounds great and we would not waste precious GitHub Actions Cache space.
Check out https://github.com/Cockatrice/Cockatrice/pull/5807 where we just recently added a new [docker-release workflow](https://github.com/maybeanerd/cockatrice/blob/master/.github/workflows/docker-release.yml?rgh-link-date=2025-06-21T11%3A12%3A23Z) to build an official image for Servatrice and upload it to GHCR for easy consumption.
Since then no new release happened to actually deploy it. 😄
>
> Did you had a closer look at those options already?

Not yet, actually. I wanna take a look at it at some point, but likely not earlier than two weeks from now.

----
Side note: I'm veeeeery slowly experimenting with conan in my own project and still wanna check if it works better with msvc + ninja + ccache + msvc + nsis for cockatrice. I have no idea when I will get there tho. Perhaps we (I?) could create a GH issue just to keep track of the ideas.

ebbit1q: yes, I made the .ci scripts, the --save option save the docker container, saving the built container and it can also load them, I initially planned to cache them but at the time the free cache space was only one fifth it is right now and there was only enough space to save 4 containers, without ccache, I also found that the speedup was minimal, github runners have a local cache for many deps and a high bandwidth connection, they set up the docker images within one minute, I decided that I'd keep the options I made to the script but not use them as the benefit was too small, I still think it's too small a benefit of shaving off only one minute of build time, I'd rather look into optimising our ccache setup to have less misses

PR #5966, opened 2025-05-30
Add "landscapeOrientation" as a valid element

SlightlyCircuitous:

https://github.com/Cockatrice/Cockatrice/pull/5264 added `landscapeOrientation` as a useful xml element to display horizontal cards such as sieges properly. However, our current v4 schema does not recognize it as a valid element when parsing. 

- Adds "landscapeOrientation" as a valid element to prevent errors when validating xmls

### DIFF
--- a/doc/carddatabase_v4/cards.xsd
+++ b/doc/carddatabase_v4/cards.xsd
@@ -56,6 +56,7 @@
             <xs:element type="xs:integer" name="tablerow" minOccurs="0" maxOccurs="1" />
             <xs:element type="xs:boolean" name="cipt" minOccurs="0" maxOccurs="1" />
             <xs:element type="xs:boolean" name="upsidedown" minOccurs="0" maxOccurs="1" />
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Not opposed to this, but would it make more sense to have an enum for the value instead of a bool? What if one day cards are displayed at a weird angle?

SlightlyCircuitous:
To be honest, I only about 80% understand this file so I just copied the line above it. Happy to make any changes you like.

tooomm:
>Legal values for boolean are true, false, 1 (which indicates true), and 0 (which indicates false).

Source: https://www.w3schools.com/XML/schema_dtypes_misc.asp

That sounds correct as the currently used values are `1` and `0` to show on/off state of landscape, not a degree value.
But I'm also no XSD expert and am not sure what the respective code in Cockatrice expects.
Maybe 0 is actually an enum for 0° and 1 for 90°?

PR #5971, opened 2025-06-09
Support regex in card search 

### LINKED ISSUE #4108: add regex search to search bar in deck editor
scryfall also supports regexes using this syntax `[property]:/[regex]/` https://scryfall.com/docs/syntax#regex

RickyRister:
## Related Ticket(s)
- Fixes #4108

Scryfall supports regexes using the syntax `[property]:/[regex]/`

Cockatrice should also support it

- Updated the peg parser to recognize strings surrounded by `/` as regex searches
  - `\/` will be escaped properly
  - `//` (with no characters between) will still be interpreted as a string search
- Handle regex search strings
- Update the search syntax help

Our behavior won't be an exact match of scryfall's behavior, since we're just relying on Qt's regex behavior. It should be close enough for most purposes though.

## Screenshots

Screenshot 2025-06-09 at 3 02 37 AM
Screenshot 2025-06-09 at 2 59 36 AM
Screenshot 2025-06-09 at 3 09 48 AM
Screenshot 2025-06-09 at 3 06 30 AM

Screenshot 2025-06-09 at 3 06 08 AM

### DIFF
--- a/cockatrice/src/game/filters/filter_string.cpp
+++ b/cockatrice/src/game/filters/filter_string.cpp
@@ -20,7 +20,7 @@ QueryPart <- NotQuery / SetQuery / RarityQuery / CMCQuery / FormatQuery / PowerQ
 
 NotQuery <- ('NOT' ws/'-') SomewhatComplexQueryPart
 SetQuery <- ('e'/'set') [:] FlexStringValue
-OracleQuery <- 'o' [:] RegexString
+OracleQuery <- 'o' [:] MatcherString
 
 
 CMCQuery <- ('cmc'/'mv') ws? NumericExpression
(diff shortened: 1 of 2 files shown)


cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -261,14 +268,22 @@ static void setupParserRules()
...
-        };
+        auto sanitizedTarget = QString(target);
+        sanitizedTarget.replace("\\\"", "\"");
+        sanitizedTarget.replace("\\'", "'");
+        return [=](const QString &s) { return s.contains(sanitizedTarget, Qt::CaseInsensitive); };
+    };
+
+    search["RegexMatcher"] = [](const peg::SemanticValues &sv) -> StringMatcher {
+        auto target = std::any_cast<QString>(sv[0]);
+        auto regex = QRegularExpression(target, QRegularExpression::CaseInsensitiveOption);
```
ebbit1q, line 281: how well does this perform when the user say, decides to add a lot of recursive regexes?

RickyRister: Can you give an example of such a query?

ebbit1q: https://www.regular-expressions.info/catastrophic.html
it'd only affect the user's machine of course, but it'd be nice to see if it has some sort of protection against it

ZeldaZach: Not sure this is a huge concern for me tbh, if it's only locally that things can go wrong. This is a power use feature, regardless, so these types of users should know to be careful

ZeldaZach (approved):

PR #5973, opened 2025-06-11
Add precon import dialog

BruebachL:
New users might struggle to come up with a deck all by themselves or picking out a deck from an external site.
MTGJson offers an archive with all the preconstructed products ever released. Since we already depend on MTGJson for parsing the database, it makes sense to also parse this information and offer it to the user.

- Add a new dialog to download the archive with preconstructed decks
- Add a parser for this archive data to turn it into .cod files
- Add a final wizard step to allow the user to specify which products to import to their deck folder location.

## Screenshots
image
image
image


ZeldaZach (changes requested):
This type of functionality should be delegated to the Oracle program, not run within the main Cockatrice. We should add an additional step to the Oracle update process where folks can (optionally) choose to import precons as a step, if they wish.

ZeldaZach:
cc: @ebbit1q

tooomm:
>This type of functionality should be delegated to the Oracle program, not run within the main Cockatrice. We should add an additional step to the Oracle update process where folks can (optionally) choose to import precons as a step, if they wish.

Didn't we want to get rid of Oracle altogether rather than adding to it? https://github.com/Cockatrice/Cockatrice/issues/4109#issuecomment-700624328

BruebachL:
> Didn't we want to get rid of Oracle altogether rather than adding to it? [#4109 (comment)](https://github.com/Cockatrice/Cockatrice/issues/4109#issuecomment-700624328)

Worth nothing that this PR is in-client rather than an oracle addon. I agree with deprecating oracle.

ebbit1q:
> Worth nothing that this PR is in-client rather than an oracle addon. I agree with deprecating oracle.

in the same vein of how oracle xmls could be served to the client, in my opinion it'd make sense to instead of parsing these on every single client individually simply have available downloads. This would:
1. move processing from all devices to just one (who doesn't love reducing global emissions?)
2. make updates to the parsing significantly easier if something needs to be changed as this way a client update is not required
3. allow users to download only the files they want instead of parsing everything
4. has users not download information in fields that aren't used
5. this saves bandwidth on both the server and the client, making transfers much faster

All of these points also apply to deprecating oracle #4109

BruebachL:
I'm not authorized to make or implement such decisions.

PR #5979, opened 2025-06-12
Add a new dialog that allows editing the default suggested tags.

BruebachL:
## Related Ticket(s)
- Implements #5688

Users have no way of editing the default suggested tags we provide.

- Adds a new dialog that does just that
- Also persists the default tags in the settings cache now

## Screenshots
image
image
image
image


ZeldaZach (approved):

ZeldaZach:
Can you address the build failures?

ZeldaZach (approved):

PR #5985, opened 2025-06-13
[Picture Loader] Consider local images, remove some unused variables.

BruebachL:
We didn't consider local images in the new parallel picture loader. Also, PictureLoaderWorker had some variables that either belong to PictureLoaderWorkerWork now OR are obsolete.

### DIFF
--- a/cockatrice/src/client/ui/picture_loader/picture_loader_worker_work.cpp
+++ b/cockatrice/src/client/ui/picture_loader/picture_loader_worker_work.cpp
@@ -16,12 +16,23 @@
 QStringList PictureLoaderWorkerWork::md5Blacklist = QStringList() << "db0c48db407a907c16ade38de048a441";
 
 PictureLoaderWorkerWork::PictureLoaderWorkerWork(PictureLoaderWorker *_worker, const CardInfoPtr &toLoad)
-    : QThread(nullptr), worker(_worker), cardToDownload(toLoad)
+    : QThread(nullptr), worker(_worker), cardToDownload(toLoad), picsPath(SettingsCache::instance().getPicsPath()),
+      customPicsPath(SettingsCache::instance().getCustomPicsPath()),
+      overrideAllCardArtWithPersonalPreference(SettingsCache::instance().getOverrideAllCardArtWithPersonalPreference())
+
 {
+    // Hook up signals to the orchestrator
     connect(this, &PictureLoaderWorkerWork::requestImageDownload, worker, &PictureLoaderWorker::queueRequest,
             Qt::QueuedConnection);
     connect(this, &PictureLoaderWorkerWork::imageLoaded, worker, &PictureLoaderWorker::imageLoadedSuccessfully,
             Qt::QueuedConnection);
+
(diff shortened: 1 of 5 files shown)


cockatrice/src/client/ui/picture_loader/picture_loader_worker_work.cpp
```diff
@@ -99,6 +112,32 @@ void PictureLoaderWorkerWork::startNextPicDownload()
...
+        // overrideAllCardArtWithPersonalPreference as reading from the
+        // SettingsCache instance from the PictureLoaderWorker thread could
+        // cause race conditions.
+        //
+        // XXX: Reading from the CardDatabaseManager instance from the
+        // PictureLoaderWorker thread might not be safe either
+        bool searchCustomPics = overrideAllCardArtWithPersonalPreference ||
+                                CardDatabaseManager::getInstance()->isProviderIdForPreferredPrinting(
+                                    cardName, cardToDownload.getCard()->getPixmapCacheKey());
+        if (searchCustomPics && cardImageExistsOnDisk(setName, correctedCardName, searchCustomPics)) {
```
RickyRister: `startNextPicDownload` is called for every url when trying out the urls one-by-one. We might want to move the local file check outside of `startNextPicDownload`, so that it doesn't get called redundantly.

cockatrice/src/client/ui/picture_loader/picture_loader_worker_work.cpp
```diff
@@ -67,22 +80,22 @@ bool PictureLoaderWorkerWork::cardImageExistsOnDisk(QString &setName, QString &c
...
         }
         imgReader.setFileName(_picsPath + ".xlhq");
         if (imgReader.read(&image)) {
-            qCDebug(PictureLoaderWorkerWorkLog).nospace() << "PictureLoader: [card: " << correctedCardname
-                                                          << " set: " << setName << "]: Picture.xlhq found on disk.";
+            qCDebug(PictureLoaderWorkerLog).nospace()
+                << "[card: " << correctedCardname << " set: " << setName << "]: Picture.xlhq found on disk.";
             imageLoaded(cardToDownload.getCard(), image);
             return true;
         }
```
RickyRister, line 133: Can you please add a "not found on disk" debug log at the end before the `return false`? It would be really helpful for me

cockatrice/src/client/ui/picture_loader/picture_loader_worker_work.cpp
```diff
@@ -33,32 +44,34 @@ PictureLoaderWorkerWork::~PictureLoaderWorkerWork()
     pictureLoaderThread->deleteLater();
 }
 
-bool PictureLoaderWorkerWork::cardImageExistsOnDisk(QString &setName, QString &correctedCardname)
+bool PictureLoaderWorkerWork::cardImageExistsOnDisk(QString &setName, QString &correctedCardname, bool searchCustomPics)
```
RickyRister: const ref the QString params while you're at it (mostly because i'm annoyed by the yellow squiggly lines)

PR #5989, opened 2025-06-14
Move "show status bar" action under Help menu

RickyRister:
## Related Ticket(s)
- Follow-up to #5983

We decided that the status bar is meant to be more of a debug sort of thing, so I think it makes more sense to move the action under the menu where we already put the other debug actions.

- Moved `aStatusBar` to under the `Help` menu, above `View Debug Log`
- Moved some stuff in the code around afterwards

## Screenshots

Screenshot 2025-06-13 at 6 16 11 PM


ZeldaZach (approved):

tooomm:
I like the new location better, too. 👍
The other changes to this new picture loading status bar are also good. It could be more explicit though what status it is showing - I think it's only about picture loading, nothing else?

Another thought I had was if we could utilize a QStatusBar instead, which would be smaller, to show such information (and others?).

On Windows however, the cog wheel for settings contains nothing. Is that intended, or only a Win issue?
After showing the loading status correctly, the status bar runs from left to right nonstop (screen is from the deck editor):
image

PR #5996, opened 2025-06-21
CI: Move and clarify macOS signing certificate step

tooomm:
The workflow file to define builds gets a bit messy if many script are directly embedded.
There is already enough config stuff going on.

- Move script for setting up the signing environment and importing the certificate into the existing script handling compilation in CI
- Gather output in a new group
- Print confirmation, or hint
- Rename build step, the signing does not happen there but in a later step

<br>

Let me know if there is better wording!

### DIFF
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -128,6 +128,23 @@ function ccachestatsverbose() {
 }
 
 # Compile
+if [[ $RUNNER_OS == macOS ]]; then
+  echo "::group::Signing Certificate"
+  if [[ -n "$MACOS_CERTIFICATE_NAME" ]]; then
+    echo $MACOS_CERTIFICATE | base64 --decode > certificate.p12
+    security create-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+    security default-keychain -s build.keychain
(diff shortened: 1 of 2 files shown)


.ci/compile.sh
```diff
@@ -128,6 +128,23 @@ function ccachestatsverbose() {
...
+  echo "::group::Signing Certificate"
+  if [[ -n "$MACOS_CERTIFICATE_NAME" ]]; then
+    echo $MACOS_CERTIFICATE | base64 --decode > certificate.p12
+    security create-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+    security default-keychain -s build.keychain
+    security set-keychain-settings -t 3600 -l build.keychain
+    security unlock-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+    security import certificate.p12 -k build.keychain -P "$MACOS_CERTIFICATE_PWD" -T /usr/bin/codesign
+    security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+    echo "macOS signing certificate successfully imported and keychain configured."
```
tooomm, line 141: Added this extra echo, let me know if I should remove it again.

PR #5997, opened 2025-06-22
[PictureLoader] Periodically refresh the local images index

RickyRister:
## Related Ticket(s)
- Follow-up to #5994

In #5994 we create a map of file names to paths of the local image folder in order to reduce lookup time for local images. However, we never refresh that index, which means it would require a restart to detect new files.

- Cockatrice now re-indexes the local image folder every 10 seconds.
  - Re-indexing is cheap enough that it should be reasonable to do every 10 seconds, even if the local image folder has lots of stuff.


ZeldaZach (approved):
LGTM, we should add a toggle to turn this on/off for performance systems

PR #5999, opened 2025-06-22
CI: Own script for signing mac app bundles

tooomm:
## Related Ticket(s)
- Follow up to https://github.com/Cockatrice/Cockatrice/pull/5996

Workflow file is busy.
While the first change helped, this is now the complete solution, combining all relevant commands into a separate script and only call it with the needed parameter and right environment variables.

- Script is mostly the same commands than before
  - Move all signing, notarizing and stapling commands into their own script
  - Move certificate configuration out of build step into the same script
  - Do validate if parameter is given and path exists at beginning
  - Cleanup keychain at the end
- Combine steps into one
- Call script with app path
- Cleaner workflow config, workflow runs and compile script

### DIFF
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -128,30 +128,13 @@ function ccachestatsverbose() {
 }
 
 # Compile
-if [[ $RUNNER_OS == macOS ]]; then
-  echo "::group::Signing Certificate"
-  if [[ -n "$MACOS_CERTIFICATE_NAME" ]]; then
-    echo $MACOS_CERTIFICATE | base64 --decode > certificate.p12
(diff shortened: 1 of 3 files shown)


ZeldaZach (approved):

ebbit1q:
I noticed this pr got reversed, is there something useful here still?

tooomm:
I wanted to extract all custom code around macOS build signing and notarization from the workflow file into its own scrip and have a cleaner ci config file in return.

Did not manage though, it was not running at the end for some reason.
I would appreciate if you can have a look @ebbit1q, maybe I introduced a small mistake or missed something.

ebbit1q:
I don't even know what notarisation does though

PR #6004, opened 2025-06-24
Automatic Card Database Updates

BruebachL:
## Related Ticket(s)
- Fixes people coming into the discord and asking "Why is set XYZ not available yet?"

Users have to manually update the card database for every single new set which is cumbersome for experienced users and not intuitive for newer users.

- Add the option to background the oracle wizard and have it run through its dialogs in a no-confirm all-default mode.
- Add an option to automatically launch oracle wizard in background every X days since last launch.

## Screenshots

image

image

image

image

### DIFF
--- a/cockatrice/src/client/ui/window_main.cpp
+++ b/cockatrice/src/client/ui/window_main.cpp
@@ -53,6 +53,7 @@
 
 #include <QAction>
 #include <QApplication>
+#include <QButtonGroup>
 #include <QCloseEvent>
 #include <QDateTime>
 #include <QDesktopServices>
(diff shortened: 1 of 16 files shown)


BruebachL:
Replace "Do the update?" Checkbox with behavior selector combo box which then triggers this on prompt (default behavior is prompt if out-of-date).

image

cockatrice/src/client/ui/window_main.cpp
```diff
@@ -62,13 +63,15 @@
...
 #include <QPixmapCache>
+#include <QRadioButton>
 #include <QStatusBar>
 #include <QSystemTrayIcon>
 #include <QThread>
 #include <QTimer>
 #include <QWindow>
 #include <QtConcurrent>
 #include <QtNetwork>
+#include <qbuttongroup.h>
```
ZeldaZach: QButtonGroup

ZeldaZach (approved):

PR #6021, opened 2025-07-05
Allow more naming schemes for custom pictures

### LINKED ISSUE #5682: Opponent's art priority ignoring custom folder
**System Information:**

Override all card art with personal set preference is unchecked. The first card I saw this happening with was Tundra. I have a custom image in my CUSTOM folder for Tundra, my opponent who'd set their Tundra art to something else, I saw what they'd chosen. I then checked the box for Override all card art with personal set preference in settings and restarted Cockatrice. When I returned the art saw I for my opponent's Tundra was then the highest set priority that I had for my Tundra, disregarding the file in my CUSTOM folder. My Tundra during this time was using the art from my CUSTOM folder.

### LINKED ISSUE #5763: On-Disk Pics and CUSTOM folder expected behaviour
Premise: Override all card art with personal set preference is NOT flagged

Due to premise, cards are loaded from pics/<set> if found, else it tries to download and cache under the new system. If a card is stored in pics/CUSTOM however, things change.

Here's how I'd expect CUSTOM to interact with the current implementation of the image load system.

i think CUSTOM image should be used:
- when navigating the database, thus overwriting the highest priority print (image posted, using promo veyran from custom folder when SLD set is not even enabled)
- in the deck editor/game own card if the card version is not specified
- in game if the card is the opponent's (CUSTOM should always overwrite opponent's selected versions: Referenced in #5682)

Image

CUSTOM should NOT be used
- in deck editor/own card if card version is specified (this already happens unless specified card version = latest among enabled sets, at which point custom overwrites, which i don't think should happen)
- in the select printing window because the pic below is just plain wrong

Image

Currently, CUSTOM images simply override the image from the highest priority set configured as can be easily seen in the pic above.

BruebachL:
Order is cardName_providerId, cardName_setName_collectorNumber, setName-collectorNumber-cardName and then just generically cardName, if the user has decided to override every printing. Most-to-least specific.

## Related Ticket(s)
- Closes #5763
- Probably closes #5682

In order to play nice with the providerId change, we decided that only the preferred providerId would check for custom pictures. This is a bit jank however. 

- If a user simply places a <cardName> file into the CUSTOM folder, it'll override everything, as before.
- If a user places a more granular naming scheme (cardName_providerId, cardName_setName_collectorNumber, setName-collectorNumber-cardName), we will attempt to load them in that order. [We could also include a cardName_setName option to just generically override a whole set where a card might have many printings {special sets come to mind}]

## Screenshots
image
image

image

### DIFF
--- a/cockatrice/src/client/ui/picture_loader/picture_loader_local.cpp
+++ b/cockatrice/src/client/ui/picture_loader/picture_loader_local.cpp
@@ -59,84 +59,81 @@ QImage PictureLoaderLocal::tryLoad(const CardInfoPtr &toLoad) const
 {
     CardSetPtr set = PictureToLoad::extractSetsSorted(toLoad).first();
 
-    QString setName = set ? set->getCorrectedShortName() : QString();
+    CardInfoPerSet setInstance = CardDatabaseManager::getInstance()->getSetInfoForCard(toLoad);
+
     QString cardName = toLoad->getName();
     QString correctedCardName = toLoad->getCorrectedName();
 
-    // FIXME: This is a hack so that to keep old Cockatrice behavior
-    // (ignoring provider ID) when the "override all card art with personal
-    // preference" is set.
-    //
-    // Figure out a proper way to integrate the two systems at some point.
-    //
-    // Note: need to go through a member for
-    // overrideAllCardArtWithPersonalPreference as reading from the
-    // SettingsCache instance from the PictureLoaderWorker thread could
-    // cause race conditions.
-    //
-    // XXX: Reading from the CardDatabaseManager instance from the
-    // PictureLoaderWorker thread might not be safe either
-    bool searchCustomPics =
-        overrideAllCardArtWithPersonalPreference ||
-        CardDatabaseManager::getInstance()->isProviderIdForPreferredPrinting(cardName, toLoad->getPixmapCacheKey());
-    if (searchCustomPics) {
-        qCDebug(PictureLoaderLocalLog).nospace()
-            << "[card: " << cardName << " set: " << setName << "]: Attempting to load picture from local";
-        return tryLoadCardImageFromDisk(setName, correctedCardName, searchCustomPics);
+    QString setName, collectorNumber, providerId;
+
+    if (setInstance.getPtr()) {
+        setName = setInstance.getPtr()->getCorrectedShortName();
+        collectorNumber = setInstance.getProperty("num");
+        providerId = setInstance.getProperty("uuid");
     }
 
     qCDebug(PictureLoaderLocalLog).nospace()
-        << "[card: " << cardName << " set: " << setName << "]: Skipping load picture from local";
-
-    return QImage();
+        << "[card: " << cardName << " set: " << setName << "]: Attempting to load picture from local";
+    return tryLoadCardImageFromDisk(setName, correctedCardName, collectorNumber, providerId);
 }
 
 QImage PictureLoaderLocal::tryLoadCardImageFromDisk(const QString &setName,
                                                     const QString &correctedCardName,
-                                                    const bool searchCustomPics) const
+                                                    const QString &collectorNumber,
+                                                    const QString &providerId) const
 {
     QImage image;
     QImageReader imgReader;
     imgReader.setDecideFormatFromContent(true);
-    QList<QString> picsPaths = QList<QString>();
 
-    if (searchCustomPics) {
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/ui/picture_loader/picture_loader_local.cpp
```diff
@@ -59,84 +59,81 @@ QImage PictureLoaderLocal::tryLoad(const CardInfoPtr &toLoad) const
...
-    //
-    // Note: need to go through a member for
-    // overrideAllCardArtWithPersonalPreference as reading from the
-    // SettingsCache instance from the PictureLoaderWorker thread could
-    // cause race conditions.
-    //
-    // XXX: Reading from the CardDatabaseManager instance from the
-    // PictureLoaderWorker thread might not be safe either
-    bool searchCustomPics =
-        overrideAllCardArtWithPersonalPreference ||
```
RickyRister, line 80: Is `overrideAllCardArtWithPersonalPreference` used anymore? If not, then can we remove all references to it in this class.

cockatrice/src/client/ui/picture_loader/picture_loader_local.cpp
```diff
@@ -59,84 +59,81 @@ QImage PictureLoaderLocal::tryLoad(const CardInfoPtr &toLoad) const
...
+                                          : QString(),
+                                      (!setName.isEmpty() && !collectorNumber.isEmpty())
+                                          ? QString("%1-%2-%3").arg(setName, collectorNumber, correctedCardName)
+                                          : QString(),
+                                      correctedCardName};
+
+    for (const QString &nameVariant : nameVariants) {
+        if (nameVariant.isEmpty()) {
             continue;
         }
```
RickyRister: Inlining the `nameVariants` string creation makes it kinda hard to follow, and it also adds empty string to the list that need to be skipped. I wonder if it would be nicer if we constructed the `nameVariants` list piece-by-piece instead.
```suggestion
    QStringList nameVariants;
    if (!providerId.isEmpty()) {
        nameVariants << QString("%1_%2").arg(correctedCardName, providerId);
    }
    if (!setName.isEmpty() && !collectorNumber.isEmpty()) {
        nameVariants << QString("%1_%2_%3").arg(correctedCardName, setName, collectorNumber)
                     << QString("%1-%2-%3").arg(setName, collectorNumber, correctedCardName);
    }
    nameVariants << correctedCardName;

    for (const QString &nameVariant : nameVariants) {
```

BruebachL: Yeah, that's reasonable as well.

cockatrice/src/client/ui/picture_loader/picture_loader_local.cpp
BruebachL: That's sort of the question. I thought the same and started removing it but then I considered:

It's still a somewhat useful feature? Some people might like to "ignore" their opponents choices without doing a custom picture for everything.

On the other hand, this change is intended to reduce the amount of people that *need* to ignore certain sets wholesale and you can just individually replace troublesome cards with versions that suit you better.

RickyRister: Oh, I wasn't talking about removing that setting in general. I was just talking about removing the `overrideAllCardArtWithPersonalPreference` field in class since it's now unused.

cockatrice/src/client/ui/picture_loader/picture_loader_local.cpp
```diff
@@ -59,84 +56,89 @@ QImage PictureLoaderLocal::tryLoad(const CardInfoPtr &toLoad) const
...
+            candidatePaths << picsPath + "/downloadedPics/" + setName + "/" + nameVariant;
         }
-        imgReader.setFileName(_picsPath + ".xlhq");
-        if (imgReader.read(&image)) {
-            qCDebug(PictureLoaderLocalLog).nospace()
-                << "[card: " << correctedCardName << " set: " << setName << "]: Picture.xlhq found on disk.";
-            return image;
+
+        for (const QString &path : candidatePaths) {
+            for (const QString suffix : {"", ".full", ".xlhq"}) {
```
ebbit1q: I know that the previous code used the full and xlhq suffixes (these are mws holdovers to make copying images from there possible afaik) however our code should ignore all suffixes anyway, it already reads images solely based on contents

note that there are some cards with dots in them, these still need to be possible to read

ZeldaZach (approved):

PR #6026, opened 2025-07-09
Refactor replay code in TabGame to replayManager

BruebachL:
With 1884 LoC in one file, TabGame does a *lot*. Our current design couples games tightly to lobbies (and thus also replays). However, we can move some of the UI code out of the way at least.

- Move the replayDockWidget and related UI elements into ReplayManager class from TabGame


RickyRister:
Keyboard shortcuts aren't working for me

Issue #4661, opened 2022-08-20
show decklists to spectators


ebbit1q:
**Similar Requests**
#4595

**Description of New Feature**
when "spectators can view hands" is checked spectators should receive decklist information as well, so they can view the players' decklists when they are loaded in a view similarly to multiplayer local games.

this would also mean this information will be included in the complete replay that is created after a game concludes (replays consist of all events as if viewed by such omniscient spectators)

**Additional Context**
this is probably not going to break older clients but would need changes both server and client side. implementation wouldn't be too complicated as all of this already exists in the client, it simply needs to have the views for each player similarly to local multiplayer during the deck selection and the server to start sending complete lists also like it were a local multiplayer game.

Issue #2555, opened 2017-03-30
Share Decklists


ZeldaZach:
Thought process: We have a button "copy sharable deck URL" which will upload the deck to the server and return back the deck ID.

A deck with the same name would overwrite the old one (i.e. I save "Deck1" to the server and get ID 123, then if I try saving it again It'll overwrite ID 123 with the new list)

A user can then post `deck://123456789` and when someone clicks it, the deck will get downloaded by the user and opened in the deck editor.

Just an idea we came up with in Gitter. Seems feasible and pretty cool.

tritoch:
Great idea that improves on people posting lists in chat.  Asking for tips on a list is very common.

If I type in `deck://987654321` or any random deckid, can I see that list or are there restrictions?  Can you 'unlist' a deck that has been published by deleting it from your server decks?

tooomm:
I would like to be able to preview the decklist like we do with posted cards.
Instead always needing to download it. Maybe preview on default (like the card info window that appears in chat) and have a `Display deck in Deck editor` button or some right click interaction with the link in chat... to actually download and save the deck.
It would also involve more people into discussion because it's simpler and quicker to check.

That could be a awesome feature for deck tech and maybe enriches the community by a lot! 👍

Daenyth:
Maybe we need some concept of public/private decks
public decks anyone can view with deck://
private ones are always an error with deck://

OR

Maybe "share deck" is just "send a copy of this deck to $player"

OR

"share deck" could generate a uuid server side and then remember the uuid->deck mapping, so  that deck://uuid-1 can view the deck but uuids are non-guessable and only exist when explicitly asked for

tritoch:
> Maybe "share deck" is just "send a copy of this deck to $player"

This one doesn't solve the use case of posting a deck to chat for anyone to view.

Daenyth:
No but it's a lot easier to add. And maybe it's enough.
Maybe we add an upload to tapped out feature instead and you use that link

tritoch:
>Maybe we add an upload to tapped out feature instead and you use that link

Already a feature.

Daenyth:
Maybe we need to make it more discoverable. A big share/upload button on
the deck editor main screen

ctrlaltca:
related: #400

skwerlman:
[...]

PR #6029, opened 2025-07-09
Add option to share decklists on load.

BruebachL:
## Related Ticket(s)
- Related to #4661
- Related to #2555

While the hash of the loaded deck is great for competitive integrity, it does nothing for the average user trying to verify if their opponent is claiming that their power level/bracket is accurate.
Additionally, it has become common practice to upload your decklist to another site beforehand and then paste it into the chat to let other people know what you are playing.

- Provide an option to create a lobby that will automatically broadcast a decklist that a player loads to all other players.
- Change the DeckViewContainer to be a TabWidget to display other players decklists as tabbed DeckViews.

## Screenshots
https://github.com/user-attachments/assets/b88348df-7559-4f9f-926e-c76a5c389496

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -197,6 +197,7 @@ set(cockatrice_SOURCES
     src/game/cards/card_search_model.cpp
     src/game/deckview/deck_view.cpp
     src/game/deckview/deck_view_container.cpp
+    src/game/deckview/tabbed_deck_view_container.cpp
     src/game/filters/deck_filter_string.cpp
     src/game/filters/filter_builder.cpp
     src/game/filters/filter_card.cpp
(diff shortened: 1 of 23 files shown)


ZeldaZach:
I believe we spoke about this offline, but I'd like to see a way to send all decks to all players as they join the room, not just when a deck is loaded (as this PR does). I know it's a bigger change, but I think it will be worth it for UX.

cockatrice/resources/config/qtlogging.ini
```diff
@@ -5,18 +5,18 @@
 # Uncomment a rule to see debug level logs for that category,
 # or set <category> = false to disable logging
 
-#main = true
+main = true
```
RickyRister: did you accidentally commit the logging configs

BruebachL: Whoops. I stg that should be .gitignored but just doesn't want to

ebbit1q:
> I believe we spoke about this offline, but I'd like to see a way to send all decks to all players as they join the room, not just when a deck is loaded (as this PR does). I know it's a bigger change, but I think it will be worth it for UX.

as far as I can see this is not implemented, in order to implement this all that is needed is for gamestate events to include the decklist, this requires no pb changes and would be only two lines of server code, clients will then have to open opponents decklists from the game state event being received, would you like me to implement the server changes for this?

Issue #4389, opened 2021-06-30
Add a feature to Import Decklist, "Import list from Deck database site"


DarkJester89:
**Description of New Feature**
This would let user "search" deck databases by a set function (username/deck name/etc).

Instead of copying/pasting the deck import, it lets users connect to major deck databases to search for decks and hyperlink them to importing. 

![image](https://user-images.githubusercontent.com/79059318/123983468-e4f9e600-d991-11eb-8c0f-99ba54f9ad11.png)

You all are great, I have no expertise in coding but I think it would be a nifty feature and tried to explain it the best I could.

Thank you!

skwerlman:
I think this is already partially covered by #4032, which has some other thoughts on this.
The major difference seems to be the presence of a deck search in trice, as opposed to just being able to download from a URL obtained elsewhere

BruebachL:
Ground floor feature work for this is completed, individual hosts should get their own tickets

PR #6030, opened 2025-07-10
Add the option to load decklists from Archidekt, Deckstats, Moxfield, TappedOut in deck editor and lobby

### LINKED ISSUE #4032: Load Remote Deck from URL (in Deck Editor)

BruebachL:
## Related Ticket(s)
- Fixes #4032
- Sort of #4389

Importing a deck from an external website usually requires the user to navigate through some sort of export dialog on the website to get the decklist into their clipboard in a format that trice likes. Then, the user has to go through our import from clipboard dialog to actually get it imported. All in all, a lot of button presses, which is why someone came up with the idea of APIs.

- Implement a new parser for various URL combos that parses out the deckID from a consumer URL and then constructs an API URL with that deckID
- Leverage Archidekt, Deckstats, Moxfield, and TappedOut APIs to get decklist exports for supplied urls
- Parse those decklists and load them in either the lobby or the deck editor.
- (Bonus) If the clipboard contains a valid URL that we can parse, the dialog auto confirms.

## Screenshots
image

https://github.com/user-attachments/assets/41b93d27-2786-430e-9d08-7528aa809c5e

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -16,6 +16,7 @@ set(cockatrice_SOURCES
     src/client/network/replay_timeline_widget.cpp
     src/client/network/sets_model.cpp
     src/client/network/spoiler_background_updater.cpp
+    src/client/network/parsers/deck_link_to_api_transformer.cpp
     src/client/sound_engine.cpp
     src/client/tabs/abstract_tab_deck_editor.cpp
     src/client/tabs/api/edhrec/tab_edhrec.cpp
@@ -165,6 +166,7 @@ set(cockatrice_SOURCES
     src/dialogs/dlg_forgot_password_reset.cpp
     src/dialogs/dlg_load_deck.cpp
     src/dialogs/dlg_load_deck_from_clipboard.cpp
+    src/dialogs/dlg_load_deck_from_website.cpp
     src/dialogs/dlg_load_remote_deck.cpp
     src/dialogs/dlg_manage_sets.cpp
     src/dialogs/dlg_move_top_cards_until.cpp
--- a/cockatrice/src/client/network/parsers/deck_link_to_api_transformer.cpp
+++ b/cockatrice/src/client/network/parsers/deck_link_to_api_transformer.cpp
@@ -1,0 +1,61 @@
+#include "deck_link_to_api_transformer.h"
+
+#include <QRegularExpression>
+
+namespace DeckLinkToApiTransformer
+{
+
+static const QString TAPPEDOUT_BASE = "https://tappedout.net/mtg-decks/";
+static const QString TAPPEDOUT_SUFFIX = "/?fmt=txt";
+
+static const QString ARCHIDEKT_BASE = "https://archidekt.com/api/decks/";
+static const QString ARCHIDEKT_SUFFIX = "/?format=json";
+
+static const QString MOXFIELD_BASE = "https://api.moxfield.com/v2/decks/all/";
+static const QString MOXFIELD_SUFFIX = "/";
+
+static const QString DECKSTATS_SUFFIX = "?include_comments=1&export_mtgarena=1";
+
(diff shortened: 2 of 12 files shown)


ebbit1q:
I like this, I think it'd be nice to have the error reporting a bit clearer though, maybe if a url doesn't match the template we can show an example of a valid url

cockatrice/src/client/network/parsers/deck_link_to_api_transformer.cpp
```diff
@@ -0,0 +1,61 @@
...
+static const QString MOXFIELD_SUFFIX = "/";
+
+static const QString DECKSTATS_SUFFIX = "?include_comments=1&export_mtgarena=1";
+
+bool parseDeckUrl(const QString &url, ParsedDeckInfo &outInfo)
+{
+    QRegularExpression rxTappedOut("tappedout\\.net/(?:mtg-decks/)?([^/?#]+)");
+    QRegularExpression rxArchidekt("archidekt\\.com/decks/(\\d+)");
+    QRegularExpression rxMoxfield("moxfield\\.com/decks/([a-zA-Z0-9_-]+)");
+    QRegularExpression rxDeckstats("deckstats\\.net/decks/(\\d+/[a-zA-Z0-9_-]+)");
```
RickyRister: I don't know if the compiler optimizes out recompiling the regex on every call, but I make all regex variables static just to be safe
```suggestion
    static QRegularExpression rxTappedOut("tappedout\\.net/(?:mtg-decks/)?([^/?#]+)");
    static QRegularExpression rxArchidekt("archidekt\\.com/decks/(\\d+)");
    static QRegularExpression rxMoxfield("moxfield\\.com/decks/([a-zA-Z0-9_-]+)");
    static QRegularExpression rxDeckstats("deckstats\\.net/decks/(\\d+/[a-zA-Z0-9_-]+)");
```

cockatrice/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ set(cockatrice_SOURCES
     src/client/network/replay_timeline_widget.cpp
     src/client/network/sets_model.cpp
     src/client/network/spoiler_background_updater.cpp
+    src/client/network/parsers/deck_link_to_api_transformer.cpp
```
RickyRister, line 19: This spot is incorrect when going in alphabetical order

cockatrice/src/dialogs/dlg_load_deck_from_website.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+        QNetworkRequest request(QUrl(info.fullUrl));
+        QNetworkReply *reply = nam->get(request);
+
+        QEventLoop loop;
+        connect(reply, &QNetworkReply::finished, &loop, &QEventLoop::quit);
+        loop.exec();
+
+        if (reply->error() != QNetworkReply::NoError) {
+            qWarning() << "Network error:" << reply->errorString();
```
RickyRister: Can we use `qCWarning`

cockatrice/CMakeLists.txt
BruebachL: Imo sub folders should be placed after files rather than just alphabetizing everything

ZeldaZach (approved):

PR #6043, opened 2025-07-16
Fix turning cards face-up not having correct printing

### LINKED ISSUE #6038: Unknown face-down cards do not have selected printing when turned face-up
**System Information:**

Client Version: 2.11.0-beta.19
Client Operating System: macOS Sequoia (15.1)
Build Architecture: ARM
Qt Version: 6.8.2
System Locale: en_CA
Install Mode: Standard

When an unknown face-down card is put onto the table (by dragging the top card of your library face-down, for example), its providerId is not set. That means if you turn it face-up, the card will have the default printing for that card name instead of the printing that was put in the deck.

https://github.com/user-attachments/assets/1af03711-30fc-40cc-a3af-3589ba2b93fd

**Steps to reproduce:**
 - Load a deck that features cards with non-default printings
 - Drag card from top of deck to battlefield face-down
 - Turn card over
   - Card will have default printing instead of selected printing

RickyRister:
## Related Ticket(s)
- Fixes #6038

Flipping a unknown face-down card face-up will cause it to have the default printing, regardless of the selected printing. This is because the flip card event does not carry providerId info.

- Added `card_provider_id` field to `Event_FlipCard` proto
- Updated server and client to use the new field

https://github.com/user-attachments/assets/30342050-5e43-4013-b103-8f9d5e16bdc8

### DIFF
--- a/cockatrice/src/game/player/player.cpp
+++ b/cockatrice/src/game/player/player.cpp
@@ -2472,10 +2472,10 @@ void Player::eventFlipCard(const Event_FlipCard &event)
         return;
     }
 
-    QString cardName = QString::fromStdString(event.card_name());
     if (!event.face_down()) {
-        // TODO: also set providerId
-        card->setCardRef({cardName});
+        QString cardName = QString::fromStdString(event.card_name());
+        QString providerId = QString::fromStdString(event.card_provider_id());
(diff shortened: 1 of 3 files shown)


ebbit1q (approved):
tested, works as expected, thank you!

PR #6049, opened 2025-07-21
Refactor: use ExactCard to represent specific printings

RickyRister:
## Related Ticket(s)
- Follow-up to #6039
- Follow-up to #6042

When we added ability to select printings to Cockatrice, we implemented it by having each printing be represented by a clone of a `CardInfoPtr`, except with a different `pixmapCacheKey`. This implementation is unintuitive. Having the printing key off the `pixmapCacheKey` causes a lot of unintuitive code.

This refactor aims to replace the printing keying off the CardInfoPtr's pixmapCacheKey with a proper object that contains both the CardInfoPtr and the printing

- Created `ExactCard` class
  - Contains a `CardInfoPtr` and a `PrintingInfo`
  - Intended to be passed around as a value object
  - Contains method to get `CardInfo` both as pointer or as const ref
- Updated `CardDatabase` to return `ExactCard`
- Replace usages of `CardInfoPtr` with `ExactCard` anywhere we care about the exact printing of the card and not just the raw card info.
- Getting the image to update when PictureLoader finishes is still done by connecting the `CardInfoPtr` inside the `ExactCard`.
  - This does mean that imageLoaded of other printings of the card will also trigger the signal, but refreshing the image is  idempotent so it should be fine.

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -198,6 +198,7 @@ set(cockatrice_SOURCES
     src/game/cards/card_database_parser/cockatrice_xml_4.cpp
     src/game/cards/card_info.cpp
     src/game/cards/card_search_model.cpp
+    src/game/cards/exact_card.cpp
     src/game/deckview/deck_view.cpp
     src/game/deckview/deck_view_container.cpp
     src/game/filters/deck_filter_string.cpp
--- a/cockatrice/src/client/tabs/abstract_tab_deck_editor.cpp
+++ b/cockatrice/src/client/tabs/abstract_tab_deck_editor.cpp
@@ -72,10 +72,10 @@ AbstractTabDeckEditor::AbstractTabDeckEditor(TabSupervisor *_tabSupervisor) : Ta
             &AbstractTabDeckEditor::refreshShortcuts);
 }
 
-void AbstractTabDeckEditor::updateCard(CardInfoPtr _card)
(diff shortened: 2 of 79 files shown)


ZeldaZach:
I see what you're doing, but I'm not exactly certain we need another wrapper around the CardInfoPtr object. Was there a reason you felt this change to be necessary?

RickyRister:
- Before the printing selector was added, CardInfoPtr seemed to be responsible for storing only the oracle info about a card (only a single CardInfo instance per card name). 
- After the printing selector was added, CardInfo also became responsible for storing the selected printing, which goes against the original design of "one instance per card name"
- Currently, we have to make a copy of the entire CardInfo object for every printing.
  - This leads to weirdness when two `CardInfoPtr`s borne from the same card + providerId combination don't equal each other because they point to two different clones.
- It's not intuitive at all how the printing info is being stored in the CardInfoPtr (inside the pixmapCacheKey). 
- It's not obvious when a CardInfoPtr usage cares about the specific printing versus when it only cares about the CardInfo. 
  - This leads to situations where we pass a generic CardInfoPtr into something that cares about specific printings, causing it to show the preferred printing instead of the selected printing

Basically, I want to separate out the part of the of the object that stays the same regardless of printing (the CardInfo) and the part that changes depending on the currently selected card. I also want to represent the "`CardInfoPtr` usages that care about printing vs ones that don't" in the type system, so that it's easier to ensure correct usage.

Issue #6054, opened 2025-07-29
Add an "increment all counters" hotkey


carrollpaul:
**Similar Requests**
None found

## Description
Add a hotkey that increments all existing counters on permanents on a player's battlefield.

## Proposed Functionality
When the hotkey is triggered, the system should:
1. Scan all permanents on the player's battlefield
2. For each permanent that has counters, add one additional counter of each type/color already present on that permanent
3. Apply the counter additions simultaneously

ebbit1q:
I notice you've used the word implements #ticketnumber instead of fixes #ticketnumber,
using fixes would automatically close the relevant ticket, did you intend for that to happen?

carrollpaul:
That was not intentional, I didn't know that using a specific word would
close the ticket.

ebbit1q:
I still don't understand, should the ticket be resolved by #6055?

carrollpaul:
Yes

Issue #6053, opened 2025-07-29
Add a "proliferate" hotkey


carrollpaul:
**Similar Requests**
None found

## Description
Add a hotkey that increments all existing counters on selected permanents on a player's battlefield.

## Proposed Functionality
When the hotkey is triggered, the system should:
1. Scan all permanents on the player's battlefield
2. For each permanent that is selected and that has counters, add one additional counter of each type/color already present on that permanent
3. Apply the counter additions simultaneously

ebbit1q:
did #6055 fix this?

carrollpaul:
Yes

ebbit1q:
ok, thanks

PR #6055, opened 2025-07-29
Add menu option and shortcut to auto increment counters

carrollpaul:
## Related Ticket(s)
- Implements [#6054](https://github.com/Cockatrice/Cockatrice/issues/6054)
- Implements [#6053](https://github.com/Cockatrice/Cockatrice/issues/6053)

There was no convenient way for players to increment all card counters at once. Players had to manually increment counters on each card individually, which was time-consuming when managing multiple cards with counters on the battlefield.

 - Added new menu item "Increment All Card Counters" to the player context menu in `player.cpp`
 - Implemented `incrementAllCardCounters` function that increments counters on selected cards (or all table cards if none selected)
 - Added keyboard shortcut support for the new action via `aIncrementAllCardCounters`
 - Function respects the `MAX_COUNTERS_ON_CARD` limit and uses `Command_SetCardCounter` for server communication
 - Menu item is properly connected and integrated into the existing player menu structure

## Screenshots
image


cockatrice/src/game/player/player.cpp
```diff
@@ -910,6 +915,7 @@ void Player::retranslateUi()
         aSetCounter[i]->setText(tr("&Set counters (%1)...").arg(cardCounterSettings.displayName(i)));
     }
 
+    aIncrementAllCardCountersOnTable->setText(tr("Increment all counters on permanents you control"));
```
RickyRister: "permanent" is too specific of a term. This action affects any card you put on the table, not just cards that have a permanent card type.

carrollpaul: Good point. How about "Increment all counters on cards you control", "Increment all counters", or "Increment all counters on table"

carrollpaul: I'm also going to move the menu item to the `Counter` submenu

cockatrice/src/game/player/player.cpp
```diff
@@ -417,13 +417,17 @@ Player::Player(const ServerInfo_User &info, int _id, bool _local, bool _judge, T
...
+        connect(aIncrementAllCardCountersOnTable, &QAction::triggered, this, &Player::incrementAllCardCountersOnTable);
+
         createPredefinedTokenMenu = new QMenu(QString());
         createPredefinedTokenMenu->setEnabled(false);
 
         mCardCounters = new QMenu;
 
         playerMenu->addSeparator();
         countersMenu = playerMenu->addMenu(QString());
+        countersMenu->addAction(aIncrementAllCardCountersOnTable);
```
RickyRister: This counters menu is for player counters, not card counters. Maybe put the action below this menu.

cockatrice/src/game/player/player.cpp
```diff
@@ -910,6 +914,7 @@ void Player::retranslateUi()
         aSetCounter[i]->setText(tr("&Set counters (%1)...").arg(cardCounterSettings.displayName(i)));
     }
 
+    aIncrementAllCardCountersOnTable->setText(tr("Increment all counters"));
```
RickyRister: “Increment all card counters” would be clearer

carrollpaul:
Updated the PR to address the shortcut naming and menu location suggestions. I also updated the function to account for selected cards. If any cards are selected, then only the counters on those cards are incremented. Otherwise, it updates all cards on the table.

cockatrice/src/game/player/player.cpp
```diff
@@ -997,7 +1002,7 @@ void Player::setShortcutsActive()
     while (counterIterator.hasNext()) {
         counterIterator.next().value()->setShortcutsActive();
     }
-
+    aIncrementAllCardCounters->setShortcut(shortcuts.getSingleShortcut("Player/aIncrementAllCardCounters"));
```
RickyRister: Fyi, #6059 will fix a bug caused by using the wrong way of getting shortcuts. 
You might want to fix it on your end now to be ahead of the curve.
```suggestion
    aIncrementAllCardCounters->setShortcuts(shortcuts.getShortcut("Player/aIncrementAllCardCounters"));
```

cockatrice/src/settings/shortcuts_settings.h
```diff
@@ -413,6 +413,10 @@ class ShortcutsSettings : public QObject
         {"Player/aSetCounter_storm", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Set Other Counters..."),
                                                  parseSequenceString("Ctrl+\\"),
                                                  ShortcutGroup::Player_Counters)},
+        {"Player/aIncrementAllCardCounters",
+         ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Increment all card counters"),
+                     parseSequenceString("Ctrl+Shift+A"),
```
RickyRister, line 418: Keep in mind that the pool of available default hotkeys is a limited resource. Not every action (especially ones that aren't common) needs to have a default hotkey.

Just something to keep in mind for the future. I'll let you decide whether you want to keep the default shortcut around or not.

Otherwise, LGTM

carrollpaul: That's a fair point. I'll retain this default for now and keep this in mind in the future.

ZeldaZach (approved):

Issue #6035, opened 2025-07-13
would it be possible to keep the red arrows either between phases or just during all combat phases?


braindeadpizzaslice:
It would be neat to be able to keep the red arrows when moving between, say declare attackers and declare blockers, especially when attacking 3 players with multiple players each, to better see what went where 

unsure if its possible as you would need different behavior depending on the phases but just a neat QoL update

ebbit1q:
there is a problem with this, the server doesn't make any difference between steps in a phase and a phase, they are just all one of the technically unbounded "phases" that are interpreted by clients in the way they like, ie the server sees no "combat phase" all it sees is "phase 4", clients then decide to call this the start of combat step of the combat phase, and add a little bit of extra margin above the icon for it, the server doesn't even insist in there being 10 of these clients can even give negative numbers if they so desired to it

there is one way I could see this implemented in a way that allows the clients to define when the arrows are removed: we add a new field to the Command_CreateArrow called something like phases_to_live or delete_after_phase which specifies after which phase it should be deleted as either a relative or absolute distance, clients which are aware of which phase they are currently in can then send this parameter to the server modified depending on when their perceived "end" of the phase is, this behavior can also then be toggled on or off client side and will be properly backwards compatible

PR #6056, opened 2025-07-30
Retain arrows when phases change

carrollpaul:
## Related Ticket(s)
- Implements [6035](https://github.com/Cockatrice/Cockatrice/issues/6035)

Currently, arrows are destroyed when phases change. This often undesirable, especially during combat when complicated attack arrows have been drawn, only to be destroyed when going to blockers.

- Arrows are not destroyed when phases change

### DIFF
--- a/common/server_game.cpp
+++ b/common/server_game.cpp
@@ -652,22 +652,7 @@ void Server_Game::setActivePlayer(int _activePlayer)
 void Server_Game::setActivePhase(int _activePhase)
 {
     QMutexLocker locker(&gameMutex);
-
-    for (Server_Player *player : players.values()) {
-        QList<Server_Arrow *> toDelete = player->getArrows().values();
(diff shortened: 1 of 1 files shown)


RickyRister:
We can't just change existing behavior like that and not leave any workarounds for people who prefer the previous behavior. It needs to be configurable at the very least. 

Ideally, each player should be able to individually configure whether their own arrows are persistent. If players with different preferences are in the same game, Cockatrice should respect each player's individual preferences.

However, phase-change-arrow-removal is controlled server-side, and Cockatrice currently doesn't handle per-player preferences on the server. We need to implement a way for the player to communicate their preferences to the server, then for the server game to remember those preferences. We ideally also want the player to be able to toggle their preferences mid-game and have the change be reflected mid-game.

Either ways, this issue is much more involved than just changing a few lines of server-side code.

Issue #4029, opened 2020-06-16
Sorting and


fbroma:
Hi guys, I wonder if you could build in a functionality that would allow players to "sort" their hand by card type and name. There are some decks, in modern at least, that pick up large amounts of cards in one go (UG storm, grishoalbrand, neobrand, and egg decks). Currently, there are two problems with the way cards stand in hand.  (1) At some point, it becomes very hard to actually read the cards (but this may require a separate issue to be opened); (2) the cards would need to be rearranged one-by-one to understand what your hand actually consists of. Point 2 is also important if you need to figure out what is left in your library and without a functionality to reorder your hand, your chances of messing up grow exponentially the more cards you have in hand.

ebbit1q:
In game right click your hand and select view hand, this'll open a window with your hand that can be manipulated in the way you want.

knitknit:
I've had this thought a number of times - happy to take this on!

knitknit:
I think I'll need changes in Cockatrice/cockatrice/src/player.cpp to create the menu, but I'm stuck trying to figure out what commands are actually needed to move cards around; any tips appreciated. Thanks!

ZeldaZach:
Moving cards around in-hand is a proto event I believe https://github.com/Cockatrice/Cockatrice/blob/44297dcd1cdf041f9633df8bcfbb6cf0c51c4a6d/cockatrice/src/handzone.cpp#L82 might help

ebbit1q:
card sorting is 100% client side, this is why rejoining a game resets your hand ordering.

ebbit1q:
this has been fixed by #6057

PR #6057, opened 2025-07-30
Add menu option and hotkey to sort hand

carrollpaul:
## Related Ticket(s)
- Implements [4029](https://github.com/Cockatrice/Cockatrice/issues/4029)

Players needed a way to organize their hand cards by type for better gameplay and strategic planning. Currently, cards in hand are only positioned based on their order of drawing, making it difficult to quickly locate and organize cards by their type during gameplay.

- Added a new "Sort Hand" action that sorts cards by type, mana value, and color (in that order)
- Implemented keyboard shortcut (Ctrl+Shift+H) for quick hand sorting
- Added "Sort Hand" option to the hand context menu
- Created sortCards() method in HandZone class that uses existing CardList sorting functionality (sorts first by main card type, then by card name)
- Updated shortcuts settings to include the new sort hand shortcut in the View group
- Players can now manually trigger hand sorting through either menu or keyboard shortcut

## Screenshots

image


ebbit1q:
some users might prefer a different sorting though, what about sorting on mv?

carrollpaul:
What do you mean by "mv"? Also, adding configuration would significantly increase complexity right now. Sorting by type and name (anecdotally) feels like the most common sense default and is definitely better than the current state (no sorting at all). We can always add more sorting options later if there's strong user demand, but I'd prefer to ship this improvement first rather than over-engineer it upfront.

ZeldaZach (approved):
While I think this is a good start, and is mergable as-is, I'd like to see an option to sort hand by a variety of factors, such as Mana Value, Type, and Name, in the future as options.

Issue #6036, opened 2025-07-14
"Put top cards on stack until..." does not support multi-word creature types


ScrangusBangus:
**System Information:**
Client Version: 2.10.2 (2025-04-04)
Client Operating System: Windows 10 Version 22H2
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_IE
Install Mode: Portable

<details><summary>Debug Log:</summary>
N/A
</details>

_______________________________________________________________________________________

**Details:**
Cockatrice does not support the search for creature types containing multiple words (the only multi-word creature type is Time Lord at the time of writing) despite Scryfall syntax supporting this. It will not find any matching cards in a deck with Time Lords.

<img width="555" height="353" alt="Image" src="https://github.com/user-attachments/assets/a31ba5e4-1f8e-4526-be62-f62c2eaabd3e" />

<b>Note:</b> While this can be worked around by instead searching for t:time t:lord, this will be a repeat issue should any additional multi-word creature types be added in future.

_______________________________________________________________________________________

**Steps to reproduce:**
- Have a deck containing any "Time Lord" creature (example below)

<img width="452" height="183" alt="Image" src="https://github.com/user-attachments/assets/acf96241-130a-472f-bd63-80b68f698df0" />

- Input t:"time lord" (as per Scryfall syntax)
- No results will be found

<img width="555" height="353" alt="Image" src="https://github.com/user-attachments/assets/548d7a95-9cf3-4235-a33b-48f4d438bc7c" />

PR #6060, opened 2025-07-31
Fix multi-word type matching in card filters

carrollpaul:
Add phrase matching to StringValue before word-based fallback. Enables searches like t:"time lord" for multi-word creature types.

## Related Ticket(s)
- Fixes [#6036](https://github.com/Cockatrice/Cockatrice/issues/6036)

Cockatrice's card filter system could not find cards with multi-word creature types (like "Time Lord") when using type searches. The filter would split text on spaces and only match individual words, so searching for `t:"Time Lord"` would fail to find cards with that exact creature type. Users had to work around this by using multiple separate word searches like `t:time t:lord`.

- StringValue matcher now performs exact phrase matching before falling back to word-based matching
- Multi-word creature types like "Time Lord" can now be found using quoted searches: `t:"Time Lord"`
- Maintains full backward compatibility with existing search patterns
- Applies the same improved matching logic to both single strings and string lists

## Screenshots
image


cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -187,16 +187,26 @@ static void setupParserRules()
 
         return QString::fromStdString(std::string(sv.sv())).toLower();
     };
+
     search["StringValue"] = [](const peg::SemanticValues &sv) -> StringMatcher {
+        // Extract common matching logic
+        auto createMatcher = [](const QString &target) -> std::function<bool(const QString &)> {
```
RickyRister: We already typedef'd `std::function<bool(const QString &)>` to `StringMatcher`, so it's better for consistency if we just stick to it.
```suggestion
        auto createMatcher = [](const QString &target) -> StringMatcher {
```

cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -187,16 +187,26 @@ static void setupParserRules()
...
-            return [=](const QString &s) { return s.split(" ").contains(target, Qt::CaseInsensitive); };
+            return createMatcher(target);
         }
 
         const auto target = std::any_cast<QStringList>(sv[0]);
-        return [=](const QString &s) {
-            auto containsString = [&s](const QString &str) { return s.split(" ").contains(str, Qt::CaseInsensitive); };
-            return std::any_of(target.begin(), target.end(), containsString);
+        return [=](const QString &s) -> bool {
+            return std::any_of(target.begin(), target.end(), [&](const QString &str) { return createMatcher(str)(s); });
```
RickyRister: nit: I still prefer to not inline the lambda here.
```suggestion
            auto containsString = [&](const QString &str) { return createMatcher(str)(s); };
            return std::any_of(target.begin(), target.end(), containsString);
```

RickyRister:
Note that this change will also cause partial matches to show up now.

For example, "t:ca" will now match cards that have type Cat, whereas before the change, only "t:cat" matched Cat, and "t:ca" would not match any card.

I'm not sure if the strict type matching behavior was originally decided intentionally, but strict matching is the behavior that Cockatrice has right now, so there should at least be a discussion if we decide to change it to partial type matching.

carrollpaul:
That's a good point. A flag could be added to the matcher for strict or partial matching:

```cpp
    search["StringValue"] = [](const peg::SemanticValues &sv) -> StringMatcher {
        // Extract common matching logic with strict matching option
        auto createMatcher = [](const QString &target, bool strictMatch = true) -> StringMatcher {
            return [target, strictMatch](const QString &s) -> bool {
                if (!strictMatch && s.contains(target, Qt::CaseInsensitive)) {
                    return true
                }
                return s.split(" ").contains(target, Qt::CaseInsensitive);
            };
        };

        if (sv.choice() == 0) {
            const auto target = std::any_cast<QString>(sv[0]);
            return createMatcher(target, true); // Use strict matching by default
        }

        const auto target = std::any_cast<QStringList>(sv[0]);
        return [=](const QString &s) -> bool {
            auto containsString = [&](const QString &str) { return createMatcher(str, true)(s); };
            return std::any_of(target.begin(), target.end(), containsString);
        };
    };
```

Obviously this would also mean adding settings or configuration, or a checkbox to the popup modal.

RickyRister:
Just maintain the strict matching behavior for now. We can figure out if and how to implement partial matching in a future PR.

However, the code you posted will only match "Time Lord" in non-strict mode. Is there a way to make "Time Lord" also match in strict mode?  That would be the most ideal outcome for this PR.

If that's not possible, then maybe we do just introduce partial matching as a workaround for the "Time Lord" issue and accept that you have to use partial matching for that type.

> Obviously this would also mean adding settings or configuration, or a checkbox to the popup modal.

If we do end up implementing it, I'd prefer if we just introduce new syntax to indicate strict vs partial matching.

carrollpaul:
How does this look? I can also remove the strictMatch flag if it won't be used. At this point `strictMatch` isn't doing much.

```cpp
auto createMatcher = [](const QString &target, bool strictMatch = true) -> StringMatcher {
    return [target, strictMatch](const QString &s) -> bool {
        if (!strictMatch && s.contains(target, Qt::CaseInsensitive)) {
            return true;
        }
        
        if (s.split(" ").contains(target, Qt::CaseInsensitive)) {
            return true;
        }
        
        return s.contains(target, Qt::CaseInsensitive);
    };
};
```

RickyRister:
Sure, you can remove the flag for now. However, your code still has partial matching behavior.

RickyRister:
I updated unit tests to test for the desired behaviour.
I can't figure out how to run the unit tests locally, but they will run in CI, so that should be able to verify if we implemented the right behavior.

https://github.com/carrollpaul/Cockatrice/pull/1

ZeldaZach:
Any update on what we'd like to do with this? :)

carrollpaul:
Updated the matcher logic. I'm also not sure how to run the tests locally (or why some are failing in the CI/CD) but I tested the following manual scenarios and they all worked as expected:

- t:elf (found Springbloom Druid)
- t:el (no matches)
- t:"Time Lord" (found The Fifth Doctor)
- t:Time Lord (no matches)

edit: Notably, the filter string test is passing in CI/CD

RickyRister:
One of the unit tests asserts the number of cards in the database and I added another card to the card database without realizing we had that test. 

Please update that unit test.

carrollpaul:
@RickyRister @ZeldaZach This looks good to go from my end - I'll leave it up to y'all

cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -187,15 +187,25 @@ static void setupParserRules()
 
         return QString::fromStdString(std::string(sv.sv())).toLower();
     };
+
     search["StringValue"] = [](const peg::SemanticValues &sv) -> StringMatcher {
+        // Helper function for word boundary matching
+        auto createWordBoundaryMatcher = [](const QString &target) {
+            QString pattern = QString("\\b%1\\b").arg(QRegularExpression::escape(target));
+            QRegularExpression regex(pattern, QRegularExpression::CaseInsensitiveOption);
```
RickyRister, line 195: Is a new regex being compiled for every single card? With a full card database, that's like 20k regexes being compiled per sweep. Could be a performance issue, but computers are fast enough these days that it probably doesn't matter. 

I guess just double check that the responsiveness in the deck editor's search doesn't go down noticeably.

carrollpaul: This is a good point. I could implement a static regex cache?

cockatrice/src/game/filters/filter_string.cpp
```diff
@@ -187,15 +193,39 @@ static void setupParserRules()
...
     search["StringValue"] = [](const peg::SemanticValues &sv) -> StringMatcher {
+        // Helper function for word boundary matching with caching
+        auto createWordBoundaryMatcher = [](const QString &target) -> std::function<bool(const QString &)> {
+            QString pattern = QString("\\b%1\\b").arg(QRegularExpression::escape(target));
+
+            // Check cache first with thread safety
+            QMutexLocker locker(&regexCacheMutex);
+            auto it = regexCache.find(pattern);
+            if (it != regexCache.end()) {
+                QRegularExpression cachedRegex = it.value();
```
RickyRister:
```suggestion
            if (regexCache.contains(pattern)) {
                QRegularExpression cachedRegex = regexCache.value(pattern);
```

Also, please double check that the performance optimization is actually necessary. If card lookup in the deck editor is still fast enough that the difference isn't noticeable to the human eye, then we don't need the extra complexity (and memory overhead) that comes from caching. Unlike with moving a fixed regex that was compiled each loop to a static const, this optimization isn't trivial or free.

carrollpaul: Is anecdotal performance testing good for you? (Aka me searching around couple times in the deck editor). I'd argue this cache isn't terribly complex and the memory cost would probably be a few MBs at worst? But agree that if it's not necessary then it's not necessary

RickyRister: Yeah, for something as minor as this, anecdotal performance testing is good enough for me.

I think I’m fine either ways with if you decide to keep the cache or not.

RickyRister: Wait no I’m dumb. I’ll check with a debugger later, but I’m pretty sure the regex is compiled once per search expression, and reused for all the cards. In which case the caching is completely unnecessary.

RickyRister: Yeah, I checked. The regex is only compiled once per search expression.

@carrollpaul please remove the caching. Sorry for sending you on a wild goose chase.

carrollpaul: All good, thanks for checking!

ZeldaZach (approved):

PR #6064, opened 2025-08-03
Refactor TabReplay creation

RickyRister:

I'm adding a new rightmost button to the right side of the replay tab and it's really annoying to do.

- Extract each side of the tab replay creation to its own method.

### DIFF
--- a/cockatrice/src/client/tabs/tab_replays.cpp
+++ b/cockatrice/src/client/tabs/tab_replays.cpp
@@ -30,6 +30,29 @@
 TabReplays::TabReplays(TabSupervisor *_tabSupervisor, AbstractClient *_client, const ServerInfo_User *currentUserInfo)
     : Tab(_tabSupervisor), client(_client)
 {
+    leftGroupBox = createLeftLayout();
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/tabs/tab_replays.cpp
```diff
@@ -112,6 +113,36 @@ TabReplays::TabReplays(TabSupervisor *_tabSupervisor, AbstractClient *_client, c
...
 
+    // Add actions to toolbars
+    toolBar->addAction(aOpenLocalReplay);
+    toolBar->addAction(aRenameLocal);
+    toolBar->addAction(aNewLocalFolder);
+    toolBar->addAction(aDeleteLocalReplay);
+
+    rightmostToolBar->addAction(aOpenReplaysFolder);
+
+    return groupBox;
```
ebbit1q, line 124: why use the return value? why not set the groupbox directly in the function?

RickyRister: Because I’m too used to always returning the result instead of having a void method that assigns the field.

PR #6066, opened 2025-08-04
Implement replay sharing

RickyRister:

Whenever I miss a game and want to see the replays, I have to 
1. ask the players
2. get them to download the replays
3. get them to send it to me on discord
4. download it off discord
5. open cockatrice
6. navigate to the location of the downloaded replay in cockatrice

It would be a quality of life improvement if someone can just give you a code for the replay, which you can enter into cockatrice and it'll put that replay into your remote replays.

I wanted the feature to meet the following constraints:
- We want to avoid changing the database schema, if possible
  - Which means the replay code needs to be able to "self-identify" which replay it's for, as we don't have anywhere to store a mapping between codes and replays
- People should not be able to guess the code for a replay unless they already have access to the replay
  - So directly using the game_id for the replay code is out

The implementation I settled on is to make the replay code be `<game_id>-<hash>`, where the hash is generated from the replay binary. This means that to generate the replay code, you would already need access to the replay binary. Also, this makes the code easy to generate, easy to find the corresponding replay, and easy to validate.

### Server changes

Added two new commands: `Command_ReplayGetCode` and `Command_ReplaySubmitCode`. 

They are handled by the server as follows:

* `Command_ReplayGetCode`
  1. validate user has access to replay
  2. query the replays to generate hash
  3. return replay code
* `Command_ReplaySubmitCode`  
  1. Check that replay exists
  2. Check that hash is correct
  3. Check that user doesn't already have access to the replay
    - Return with a early success if so
  4. Update `cockatrice_replay_access` table to grant user access to replay 
    - The `replay_name` will be copied from one of the other `cockatrice_replay_access` entries for that id.

Currently, the hash implementation is: 
- Take the first 128 bytes of the first three replays of that game and append them together
  - We only use a subset of the replays to guard against outliers that might require excess resources, while still having enough variance that the hash is impossible to guess
- MD5 hash it
- base64 encode it, then truncate it to 10 characters
  - This is to reduce the length of the replay code. Even though you're suppose to copy-paste it, overly-long codes are still  ugly and unwieldy.

The code generation and validation is all done server-side. This means the client is forwards-compatible if we decide to change how we implement replay codes in the future.

### Client changes

Updated the right side of the replays tab with two new buttons for interfacing with replay sharing.

https://github.com/user-attachments/assets/3608738d-3655-4451-9b9d-4f5bb55f6edd

- "get replay code" button 
  - On click, sends the ReplayGetCode command for the selected replay folder (or the replay folder the selected replay is in)
  - Upon getting a successful response, opens a dlg containing the replay code, with a "copy to clipboard" button
- "look up replay by code" button
  - Opens a dlg that allows user to submit a replay code
  - Sends the ReplaySubmitCode command
    - Upon response, tells user if code succeeded or not
    - The dir view gets updated with the added replay on success

The client also has error messages in case the server returns a response that indicates it doesn't recognize the command.

https://github.com/user-attachments/assets/228dbc64-a827-439a-a171-1e93ceb2e361

Note that this UI is just a minimal implementation that allows the user to interface with the feature without too much friction. We can improve the UI in future PRs.

## Screenshots

Screenshot 2025-08-19 at 9 41 05 PM


cockatrice/src/client/tabs/tab_replays.cpp
```diff
@@ -181,6 +208,9 @@ void TabReplays::retranslateUi()
     aDownload->setText(tr("Download replay"));
     aKeep->setText(tr("Toggle expiration lock"));
     aDeleteRemoteReplay->setText(tr("Delete"));
+    aGetReplayCode->setText(tr("Get replay code"));
```
ZeldaZach: We should probably indicate it's a replay share code or smth

cockatrice/src/client/tabs/tab_replays.cpp
```diff
@@ -181,6 +208,9 @@ void TabReplays::retranslateUi()
...
+
+    aSubmitReplayCode->setText(tr("Look up replay by code"));
```
ZeldaZach: by share code

servatrice/src/serversocketinterface.cpp
```diff
@@ -735,6 +743,135 @@ Response::ResponseCode AbstractServerSocketInterface::cmdReplayDeleteMatch(const
...
+ *
+ * Currently, we append together the first 128 bytes of the first 3 replays in the game.
+ * Then we md5 hash it, base64 encode it, and truncate the result to 10 characters.
+ *
+ * @param gameId The replay match to hash
+ * @return The hash as a QString. Returns an empty string if failed
+ */
+QString AbstractServerSocketInterface::createHashForReplay(int gameId)
+{
+    QSqlQuery *query =
```
ZeldaZach, line 758: I'm pretty sure the answer is no, but I have to ask: Does this leak memory?

ZeldaZach (approved):
I feel we can improve on the user facing messages, but the overall code lgtm

servatrice/src/serversocketinterface.cpp
RickyRister: The surrounding code all use QSqlQuery without manually deleting it, so either the query automatically cleans itself up, or we've been leaking memory all this time. I would guess that it doesn't leak memory, but I'll double check the docs.

RickyRister:
Thicker outline
Screenshot 2025-08-24 at 7 39 09 PM

New message
Screenshot 2025-08-24 at 7 39 57 PM

PR #6067, opened 2025-08-06
Refactor: rename and consolidate getSpectator

RickyRister:

`Player::getSpectator` is completely redundant with `Player::isSpectator` (as in, the method bodies are the exact same).

- Delete `getSpectator` and use `isSpectator` instead
- Rename `getSpectatorsSeeEverything` to `isSpectatorsOmniscient`

### DIFF
--- a/cockatrice/src/client/tabs/tab_game.h
+++ b/cockatrice/src/client/tabs/tab_game.h
@@ -234,15 +234,14 @@ public:
         return gameInfo.game_id();
     }
     QString getTabText() const override;
-    bool getSpectator() const
+    bool isSpectator() const
     {
         return spectator;
     }
-    bool getSpectatorsSeeEverything() const
+    bool isSpectatorsOmniscient() const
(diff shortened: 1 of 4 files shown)


cockatrice/src/client/tabs/tab_game.h
```diff
@@ -234,15 +234,14 @@ protected slots:
         return gameInfo.game_id();
     }
     QString getTabText() const override;
-    bool getSpectator() const
+    bool isSpectator() const
     {
         return spectator;
     }
-    bool getSpectatorsSeeEverything() const
+    bool isSpectatorsOmniscient() const
```
ebbit1q, line 241: I don't particularly care about this rename but a lot of other accessors in this same file use get and not is

RickyRister: That’s unfortunate. We can deal with those as they come up.

PR #6068, opened 2025-08-06
Add Debian 13 'Trixie' build

SlightlyCircuitous:

Debian 13 (Trixie) is scheduled for release on August 9, 2025 so it would be advantageous to build for it in our next release. 

- Add Debian 13 dockerfile
- Update desktop-build and release template accordingly

As far as I can tell from compiling from source in a Debian 13 RC3 VM, there are no package changes relevant to our build requirements. Trixie has been in full freeze for about a week and half now so I don't expect that to change for the official release.

This PR is a draft until Docker adds the release candidate image (Debian:13 tag instead of Debian:Trixie). I will test the complied software in a Debian 13 VM at that time.

### DIFF
--- a/.ci/Debian13/Dockerfile
+++ b/.ci/Debian13/Dockerfile
@@ -1,0 +1,28 @@
+FROM debian:13
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        build-essential \
+        ccache \
+        clang-format \
+        cmake \
+        file \
+        g++ \
+        git \
+        libgl-dev \
+        liblzma-dev \
+        libmariadb-dev-compat \
+        libprotobuf-dev \
+        libqt6multimedia6 \
+        libqt6sql6-mysql \
+        ninja-build \
+        protobuf-compiler \
+        qt6-image-formats-plugins \
+        qt6-l10n-tools \
+        qt6-multimedia-dev \
+        qt6-svg-dev \
+        qt6-tools-dev \
+        qt6-tools-dev-tools \
+        qt6-websockets-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
--- a/.ci/release_template.md
+++ b/.ci/release_template.md
@@ -20,6 +20,7 @@ Available pre-compiled binaries for installation:
   <b>Linux</b>
    • <kbd>Ubuntu 24.04 LTS</kbd> <sub><i>Noble Numbat</i></sub>
    • <kbd>Ubuntu 22.04 LTS</kbd> <sub><i>Jammy Jellyfish</i></sub>
+   • <kbd>Debian 13</kbd> <sub><i>Trixie</i></sub>
    • <kbd>Debian 12</kbd> <sub><i>Bookworm</i></sub>
    • <kbd>Debian 11</kbd> <sub><i>Bullseye</i></sub>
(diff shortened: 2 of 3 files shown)


SlightlyCircuitous:
Speaking of Debian, I've been going back and forth over whether or not to propose removing the Debian 11 build. It is technically out of standard support as it moved from Debian-stable to Debian-oldstable in August 2024, but Debian doesn't seem to draw as bright of a line between stable and oldstable as other distros do. Maybe it should stay until its LTS period ends in 2026?

Debian 11 is also one of the few builds that works with ChromeOS, and I am having trouble determining a reasonable cut-off for only supporting Debian 12. As far as I can tell, any ChromeOS user that can install update 121 (released in 2024) can run Debian 12. Since Google seemingly provides updates for 10 years, anyone with a Chromebook from 2014 or later should be able to do that update. That's plenty of support, right? 

Also can't figure out if Google is going to support Trixie as well, but it seems likely. I'll keep an eye on it.

ebbit1q:
we stop supporting them when they stop building, there is no cost to us for keeping them until they do that

SlightlyCircuitous:
Alright, I'll leave it up for now then.

SlightlyCircuitous:
Debian:13 tag now exists on Docker. Is there a way to re-run the desktop build check? I am realizing that I did not think through making this a draft enough...

BruebachL:
> Debian:13 tag now exists on Docker. Is there a way to re-run the desktop build check? I am realizing that I did not think through making this a draft enough...

There's no way for you to do that, someone with write access can re-trigger it (or you can (force-)push a trivial change)

SlightlyCircuitous:
Thank you Zach for triggering the re-run. It looks like the image now builds fine, but something went wrong with downloading the Google test suite. I'll look into it.

SlightlyCircuitous:
There seems to be some issue with the SSL certificate: `status_code: 77 status_string: "Problem with the SSL CA cert (path? access rights?)"` causing the download to fail but only in the Debian 13 docker image specifically.

googletest downloads fine on my Trixie VM when I run`cmake .. -DCMAKE_BUILD_TYPE=Debug -DTEST=1` and I can follow the link directly to the zip file on my actual machine. The other Linux builds that need to download googletest to run debug (Fedora 42, Ubuntu 24.04) also work just fine.

Not sure what's going on, but I'll keep looking.

SlightlyCircuitous:
I found instructions on how to [add a CA certificate to the docker image](https://docs.docker.com/engine/network/ca-certs/#add-ca-certificates-to-linux-images-and-containers) in the dockerdocs, but I am not sure where the CA certificate should come from (i.e. I have no idea what they mean by "Copy the CA certificate from the context" in the provided code block). Also, while this seems like this should address the SSL issue, I still can't find a reason that this image, and not any of the others, would have this specific problem now.

SlightlyCircuitous:
Based on several searches, I was able to determine that `ca-certificates` is included in the Debian 13 image I am running on my VM, but not in the Docker build (as they use a much smaller set of packages) and that merely installing `ca-certificates` had the potential to solve the problems I was having. Fortunately, that turned out to be the case here. Still not sure why 13 is behaving so differently from 12 did when it was the debug build, but at least 13 builds now. 

I have downloaded the build artifact and installed it in my Debian 13 VM. It installs without incident, checks for card updates, makes decks, and handles local games as expected.

ZeldaZach (approved):

PR #6070, opened 2025-08-07
Refactor: Move files in src/client/game_logic

RickyRister:

It's not clear what the `src/client/game_logic` folder is suppose to represent. The two files in there (`abstract_client` and `key_signals`) make more sense elsewhere.

- Moved `abstract_client` to `src/server`
  - Since that's where the two `AbstractClient` subclasses are
- Moved `key_signals` to `src/utility`


ebbit1q (approved):
thanks, it's at least a little better like this

PR #6071, opened 2025-08-08
Implement RefreshGuard system to prevent accidental navigation and data loss

carrollpaul:
## Related Ticket(s)
- Addresses webclient README To-Do List item (RefreshGuard modal)

Users could accidentally lose their game state by refreshing the page or navigating away while in active games or connected to rooms. The existing implementation only showed a basic browser dialog without proper WebSocket cleanup or detailed context about what would be lost.

  - Browser refresh/navigation now shows native dialog followed by detailed custom modal
  - React Router SPA navigation shows contextual warning modals before leaving
  - WebSocket connections are gracefully disconnected only when users actually leave
  - Room component has defensive null checks to prevent crashes
  - Local server configuration added for Docker development
  - Comprehensive hook system for managing navigation guards across the application

## Screenshots

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   mysql:
     image: mysql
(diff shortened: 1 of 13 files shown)


ebbit1q:
webatrice has been moved here: https://github.com/Cockatrice/Webatrice

PR #6097, opened 2025-09-04
Properly allow resizing the printing selector

BruebachL:
## Fixes
Crash when selecting reset layout in VDE
Unused variables in VDE

Printing selector has some sizes set that can make it feel "locked" to the user, especially on devices with smaller screens. This sets the minimum size of the printing selector to be equal to the card info and the deck dock widget, which is a common use case for layouts I've observed. 

It's also never "set" as a free dock size for the visual printing selector which makes it unresizable (no min or max sizes set).

- Change min width of printing selector to be in line with other dock widgets
- Set min and max sizes of printing selector in visual deck editor
- Remove unused deck analytics dock widget from visual deck editor
- Reset layout properly in visual deck editor
- Introduce a new default layout for visual deck editor [but not the regular deck editor] (see attached screenshot, filter is removed and the slot is moved centrally underneath a horizontally split card and deck info and filled with printing selector)

## Screenshots
image
image

### DIFF
--- a/cockatrice/src/client/ui/widgets/cards/deck_card_zone_display_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/cards/deck_card_zone_display_widget.cpp
@@ -64,7 +64,7 @@ void DeckCardZoneDisplayWidget::constructAppropriateWidget(QPersistentModelIndex
             subBannerOpacity, cardSizeWidget);
         connect(displayWidget, SIGNAL(cardClicked(QMouseEvent *, CardInfoPictureWithTextOverlayWidget *)), this,
                 SLOT(onClick(QMouseEvent *, CardInfoPictureWithTextOverlayWidget *)));
-        connect(displayWidget, SIGNAL(cardHovered(CardInfoPtr)), this, SLOT(onHover(CardInfoPtr)));
+        connect(displayWidget, SIGNAL(cardHovered(ExactCard)), this, SLOT(onHover(ExactCard)));
         connect(displayWidget, &CardGroupDisplayWidget::cleanupRequested, this,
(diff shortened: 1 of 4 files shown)


ebbit1q (approved):
works well, thank you!

cockatrice/src/client/ui/widgets/cards/deck_card_zone_display_widget.cpp
```diff
@@ -64,7 +64,7 @@ void DeckCardZoneDisplayWidget::constructAppropriateWidget(QPersistentModelIndex
             subBannerOpacity, cardSizeWidget);
         connect(displayWidget, SIGNAL(cardClicked(QMouseEvent *, CardInfoPictureWithTextOverlayWidget *)), this,
                 SLOT(onClick(QMouseEvent *, CardInfoPictureWithTextOverlayWidget *)));
-        connect(displayWidget, SIGNAL(cardHovered(CardInfoPtr)), this, SLOT(onHover(CardInfoPtr)));
+        connect(displayWidget, SIGNAL(cardHovered(ExactCard)), this, SLOT(onHover(ExactCard)));
```
RickyRister, line 67: is there a reason we're not using the new signal/slot syntax?

BruebachL: Refactoring this is a different PR altogether. You're right in that fixing this wouldn't have taken a whole lot of effort but there are other connections in this same class that also still use the old syntax and I was just sneaking in a fix for a regression from the ExactCard refactor.

ZeldaZach: Agreed, we can refactor to new SIG/SLOT elsewhere

PR #6113, opened 2025-09-10
update the address of the fsf in the license

ebbit1q:
the fsf closed its office in august last year, it's now remote only and the address has been replaced with its url, the license text can be found here: https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

I made this commit last year, then forgot to pr it entirely, it's not very important but here it is anyway!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,7 +2,7 @@
                        Version 2, June 1991
 
  Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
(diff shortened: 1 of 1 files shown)


ebbit1q:
note that gnu.org where the license text is hosted is currently down, likely from ddos, https://hostux.social/@fsfstatus will likely announce when it's up

ZeldaZach (approved):

PR #6120, opened 2025-09-11
Don't render opponent hands if not visible (i.e. contents known)

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


ebbit1q:
some users like this feature, could it be possible to make it optional?

PR #6124, opened 2025-09-13
Connect concession signal correctly

BruebachL:
I derped and didn't rename a signal, which led to concession commands never being sent (but the local player conceded state did change). Also, probably connect game state's stopped to TabGame::stopGame.

- Connect signals appropriately.


ebbit1q (approved):
thank you!

Issue #6115, opened 2025-09-11
Servatrice sends card information to all players if see hands is enabled


ZeldaZach:
P0 issue that was uncovered by a bug in the newest beta where if you rejoin a game, it would tell the server you're a spectator.

It appears Servatrice has been leaking this information for a (long?) while, and it's up to the client to tell the server if they're in spectator mode or not. We need to add additional verifications that the user joining the room is indeed a spectator or a player. They can't be both.

<img width="2920" height="2086" alt="Image" src="https://github.com/user-attachments/assets/1726f0d4-782b-4aea-a2d5-74ce4d6aeecd" />

BruebachL:
This seems to happen regardless of if spectators can see hands is enabled or not.

ebbit1q:
it also includes the opponent's full decklist

ebbit1q:
note, this regression was introduced in #4281 it affected all commits from 06bfc02 to 216cd49, including releases 2.9.0 to 2.10.2

PR #6125, opened 2025-09-13
do not allow other players to know which cards are in a player's hand

ebbit1q:
## Related Ticket(s)
- similar to #6115

the server never did any attempts to hide which card in a player's hand was which, allowing opponents to track which cards are being returned

- cards being moved to and from the hand will have their position removed in the event being sent out to others
- the check for the since #5410 broken cards that are being looked at in the hidden zone (deck) is now removed, to be honest I did not fully understand the reasoning from 2012 for this to exist
- the event now removes the value instead of explicitly setting it to -1, -1 is the default value since the first ever release so this has no backwards compatibility concerns


ebbit1q:
rebased

ebbit1q:
rebased

PR #6126, opened 2025-09-13
Forward playerActions signals to playerEventHandler

BruebachL:
We lost some logging signal connections related to a player manipulating card attributes because they were moved to playerActions.

- Forward signals to playerEventHandler again so it, in turn, can properly signal them to the message log widget.
- Remove some debug prints


ebbit1q (approved):
seems good, thank you

PR #6129, opened 2025-09-13
table_zone_logic now sets the PT of the cardItem to the cardInfos PT

### LINKED ISSUE #6128: Power/Toughness not being set correctly when card enters from hidden zone
It's something to do with this piece of logic in pile_zone.cpp's mouseMoveEvent

```
    const int cardid =
        getLogic()->contentsKnown() ? card->getId() : (bottomCard ? getLogic()->getCards().size() - 1 : 0);
    CardDragItem *drag = card->createDragItem(cardid, event->pos(), event->scenePos(), faceDown);
```
The id is 0 (because the contentsKnown for the hidden library zone is false) but this never gets resolved back to a card before TableZone::handleDropEventByGrid checks the card.

So this block right here

```
    for (const auto &item : dragItems) {
        CardToMove *ctm = cmd.mutable_cards_to_move()->add_card();
        ctm->set_card_id(item->getId());
        ctm->set_face_down(item->getFaceDown());
        if (startZone->getName() != getLogic()->getName() && !item->getFaceDown()) {
            const auto &card = item->getItem()->getCard();
            if (card) {
                ctm->set_pt(card.getInfo().getPowTough().toStdString());
            }
        }
    }
``` 
fails at 
const auto &card = item->getItem()->getCard();
since the dragItem doesn't have a card set (yet).
I can't figure out where we *do* resolve the card.

**Steps to reproduce:**
 - Drag a card from the library to the battlefield
 - Observe no P/T set on creatures
 - Drag a card from the hand to the battlefield
 - Observe P/T set on creatures

BruebachL:
… which ensures consistency of this functionality when the card is added from a hidden zone.

## Related Ticket(s)
- Fixes #6128

Power/Toughness not being set correctly when card enters from hidden zone

- Resolve cardItem P/T from cardInfo when card enters table_zone (which is contentsKnown = true)

### DIFF
--- a/cockatrice/src/game/zones/logic/table_zone_logic.cpp
+++ b/cockatrice/src/game/zones/logic/table_zone_logic.cpp
@@ -15,6 +15,9 @@ TableZoneLogic::TableZoneLogic(Player *_player,
 void TableZoneLogic::addCardImpl(CardItem *card, int _x, int _y)
 {
     cards.append(card);
+    if (!card->getFaceDown()) {
+        card->setPT(card->getCardInfo().getPowTough());
+    }
     card->setGridPoint(QPoint(_x, _y));
     card->setVisible(true);
 }


ebbit1q (approved):
seems fine, thanks

PR #6114, opened 2025-09-10
Home tab to replace generic deck editor on startup

BruebachL:
## Short roundup of the initial problem
Currently, Cockatrice opens on the text based deck editor. This implies the user intent when opening Cockatrice is always immediately the following:
- Brew a deck

However, there are a lot of users that don't use Cockatrice for this purpose at all. They import their decks from another website and are mostly concerned with connecting to the server and playing.
Over time, we've added more functionality to Cockatrice than simply creating decks and playing games.
Currently, our navigation is tab based, so we've decided to store all this additional functionality beneath the "Tabs" menu (which causes some loss of clarity when talking about the application and, on MacOS, as far as I can tell, isn't permanently visible). This goes contradictory to the basic design principle of _discoverability_ and _understanding_:
What can I even do?
What does performing this action cause?

In addition, since this is usually the first screen people see when launching cockatrice (apart from being bombarded with menus), this imprints the "visual identity" of: 
"It's Excel, for cards."

## What will change with this Pull Request?
- Implement a home tab to replace the text based deck editor on startup
- Implement navigational buttons for the following functions:

```
- Connect (to server) -> Opens the connect to server dialog and changes to
- Play -> Opens (in this order) the first found: Room Tab, Server Tab
- Create New Deck -> Opens the Visual Deck Editor to a blank deck
- Browse Decks -> Opens the Visual Deck Storage
- Browse Card Database -> Opens the Visual Database Display
- Browse EDHRec -> Opens EDHRec tab
- Browse Replays -> Opens Replay tab
- Quit -> Closes the application
```

## Screenshots
image

PR #6131, opened 2025-09-13
Properly delete enlargedPixmapWidget

BruebachL:
## Related Ticket(s)
- Fixes issue reported here https://github.com/Cockatrice/Cockatrice/pull/6114#issuecomment-3282281715

If the VDS is hidden while the tab is closed, destructor stuff causes a crash because the enlargedPixmapWidget is owned by the parent window, not the tab.

- emit a signal and forward to enlargedPixmapWidget's deleteLater

### DIFF
--- a/cockatrice/src/client/ui/widgets/cards/card_info_picture_widget.cpp
+++ b/cockatrice/src/client/ui/widgets/cards/card_info_picture_widget.cpp
@@ -40,8 +40,7 @@ CardInfoPictureWidget::CardInfoPictureWidget(QWidget *parent, const bool _hoverT
 
     enlargedPixmapWidget = new CardInfoPictureEnlargedWidget(this->window());
     enlargedPixmapWidget->hide();
-    connect(this, &CardInfoPictureWidget::deleteEnlargedPixmapWidget, enlargedPixmapWidget,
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/ui/widgets/cards/card_info_picture_widget.cpp
```diff
@@ -65,10 +67,7 @@ CardInfoPictureWidget::CardInfoPictureWidget(QWidget *parent, const bool _hoverT
 
 CardInfoPictureWidget::~CardInfoPictureWidget()
 {
-    if (enlargedPixmapWidget) {
-        enlargedPixmapWidget->hide();
-        enlargedPixmapWidget->deleteLater();
-    }
+    emit deleteEnlargedPixmapWidget();
```
RickyRister: Isn't emitting signals from a destructor a bad idea?
Does connecting `QObject::destroyed` not work?

BruebachL: You're absolutely right, I didn't consider that, it's much cleaner.

ebbit1q (approved):
this makes sense, thanks

PR #6132, opened 2025-09-13
Disable home screen buttons until ready

BruebachL:
You can crash by clicking "View Card Database" before the card db is properly loaded

- Disable view card database button until card db is loaded
- add a new style for disabled buttons.

## Screenshots
Disabled button:
image

### DIFF
--- a/cockatrice/src/client/ui/widgets/general/home_styled_button.cpp
+++ b/cockatrice/src/client/ui/widgets/general/home_styled_button.cpp
@@ -60,17 +60,16 @@ QString HomeStyledButton::generateButtonStylesheet(const QPair<QColor, QColor> &
                                         stop:0 %8, stop:1 %9);
         }
     )")
-        .arg(base1.name())     // border color
-        .arg(base1.name())     // normal gradient start
-        .arg(base2.name())     // normal gradient end
-        .arg(hover1.name())    // hover start
-        .arg(hover2.name())    // hover end
-        .arg(pressed1.name())  // pressed start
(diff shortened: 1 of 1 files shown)


cockatrice/src/client/ui/widgets/general/home_styled_button.cpp
```diff
@@ -47,14 +53,22 @@ QString HomeStyledButton::generateButtonStylesheet(const QPair<QColor, QColor> &
...
-        .arg(pressed2.name()); // pressed end
+        .arg(base1.name())      // border color
+        .arg(base1.name())      // normal gradient start
+        .arg(base2.name())      // normal gradient end
+        .arg(hover1.name())     // hover start
+        .arg(hover2.name())     // hover end
+        .arg(pressed1.name())   // pressed start
+        .arg(pressed2.name())   // pressed end
+        .arg(disabled1.name())  // disabled start
+        .arg(disabled2.name()); // disabled end
```
ebbit1q, line 71: why are their names disabled2 with a comment behind it saying disabled end if it could have just been called disabled_end everywhere?

ebbit1q:
is it also possible to crash when selecting the visual database display from the dropdown menus?

BruebachL:
> is it also possible to crash when selecting the visual database display from the dropdown menus?

Why would you want to do this? :P

ebbit1q:
to intentionally crash the client of course, I tried but I have to be lightning fast or something to do the needed clicks fast enough

PR #6134, opened 2025-09-13
Add a style for disabled buttons

BruebachL:
HomeStyledButtons don't have a stylesheet for being disabled.

- Add a disabled/desaturated stylesheet

## Screenshots
image


ebbit1q:
in what places is the button disabled? is it only while connecting?

BruebachL:
Yes, currently it is only the connect button.

ebbit1q (approved):
excellent, thank you for the new names

PR #6135, opened 2025-09-13
Use a scoped disconnect for Home connect button

BruebachL:
We wildcard disconnect all signals/slots from the connect button, which issues a warning
QObject::disconnect: wildcard call disconnects from destroyed signal of HomeStyledButton::unnamed

- Use a scoped disconnect to silence warning.


ebbit1q (approved):
sure

PR #6137, opened 2025-09-14
Check if card has no PT set yet if dropped on table.

BruebachL:
## Related Ticket(s)
- Fixes regression introduced by #6129

Moving a card within a table zone will reset PT because addCardImpl gets called. This is fine when moving TO a table zone since no other zones can have PT set but not fine when moving within a zone.

- Check if a card's PT is empty before overwriting it.

### DIFF
--- a/cockatrice/src/game/zones/logic/table_zone_logic.cpp
+++ b/cockatrice/src/game/zones/logic/table_zone_logic.cpp
@@ -15,7 +15,7 @@ TableZoneLogic::TableZoneLogic(Player *_player,
 void TableZoneLogic::addCardImpl(CardItem *card, int _x, int _y)
 {
     cards.append(card);
-    if (!card->getFaceDown()) {
+    if (!card->getFaceDown() && card->getPT() == "") {
         card->setPT(card->getCardInfo().getPowTough());
     }
     card->setGridPoint(QPoint(_x, _y));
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/zones/logic/table_zone_logic.cpp
```diff
@@ -15,7 +15,7 @@ TableZoneLogic::TableZoneLogic(Player *_player,
 void TableZoneLogic::addCardImpl(CardItem *card, int _x, int _y)
 {
     cards.append(card);
-    if (!card->getFaceDown()) {
+    if (!card->getFaceDown() && card->getPT() == "") {
```
RickyRister: Nit: Are we suppose to be using `== ""` or `isEmpty()` to check for empty strings? I see both usages in the codebase, but we should at least try to be somewhat consistent for new code.

BruebachL: Personally and anecdotally, without any concrete evidence to back it up, I've found `isEmpty()` to be a bit finnicky. As in, I seem to remember instances where it just did not evaluate correctly.

Anyway, that's just my superstition.

ebbit1q: isEmpty should always evaluate correctly, if it doesn't we have bigger problems, please use isEmpty for consistency

ebbit1q (approved):
thanks

PR #6140, opened 2025-09-14
update intel mac runner from 13 to 15, deprecate ventura

ebbit1q:
macos ventura will stop being supported officially tomorrow, however because homebrew is eager to stop supporting it builds already fail now, we have no choice but to move on to sequoia

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -230,7 +230,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: 13
+          - target: 15
             soc: Intel
             os: macos-15-intel
             xcode: "16.2"
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -230,10 +230,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: 14
+          - target: 13
             soc: Intel
-            os: macos-14
-            xcode: "15.4"
+            os: macos-15-intel
+            xcode: "16.2"
             type: Release
             make_package: 1
 
@@ -267,6 +267,7 @@ jobs:
       DEVELOPER_DIR:
         /Applications/Xcode_${{matrix.xcode}}.app/Contents/Developer
       CMAKE_GENERATOR: 'Ninja'
+      MACOSX_DEPLOYMENT_TARGET: ${{matrix.target}}.0
 
     steps:
       - name: Checkout
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -230,10 +230,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: 13
+          - target: 14
             soc: Intel
-            os: macos-13
-            xcode: "14.3.1"
+            os: macos-14
+            xcode: "15.4"
             type: Release
             make_package: 1
 
--- a/.ci/release_template.md
+++ b/.ci/release_template.md
@@ -15,7 +15,7 @@ Available pre-compiled binaries for installation:
   <b>macOS</b>
    • <kbd>macOS 15+</kbd> <sub><i>Sequoia</i></sub> <sub>Apple M</sub>
    • <kbd>macOS 14+</kbd> <sub><i>Sonoma</i></sub> <sub>Apple M</sub>
-   • <kbd>macOS 13+</kbd> <sub><i>Ventura</i></sub> <sub>Intel</sub>
+   • <kbd>macOS 15+</kbd> <sub><i>Sequoia</i></sub> <sub>Intel</sub>
 
   <b>Linux</b>
    • <kbd>Ubuntu 24.04 LTS</kbd> <sub><i>Noble Numbat</i></sub>


tooomm:
[The `macOS 14` runner image is ARM based](https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images):
image

And the "large" ones are [not available for open source](https://docs.github.com/en/actions/concepts/runners/larger-runners).

.github/workflows/desktop-build.yml
```diff
@@ -230,10 +230,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: 13
+          - target: 14
             soc: Intel
```
tooomm, line 234: To my understanding, the resulting build will be targeting ARM as that's the base image.
The `soc` value is only used to add a label to the resulting build to differentiale Intel/ARM as we had both before.
We can potentially get rid off it completely if we are now ARM-only for Mac builds, and just put a note in the release notes that we no longer support macOS 13, as well as Intel-based Mac's.

<br>

@fwcd mentioned in https://github.com/Cockatrice/Cockatrice/pull/5033#issuecomment-2216106032 some time ago that using the newer images you can still target older macOS versions. We never looked into it.

Not sure the same is possible for arch.

.github/workflows/desktop-build.yml
```diff
@@ -230,10 +230,10 @@ jobs:
...
-            os: macos-13
-            xcode: "14.3.1"
+            os: macos-14
```
tooomm: To my understanding, the resulting build will be targeting ARM as that's the base image.
The `soc` value is only used to add a label to the resulting build to differentiale Intel/ARM as we had both before.
We can potentially get rid off it completely if we are now ARM-only for Mac builds, and just put a note in the release notes that we no longer support macOS 13, as well as Intel-based Mac's.

<br>

@fwcd mentioned in https://github.com/Cockatrice/Cockatrice/pull/5033#issuecomment-2216106032 some time ago that when using the newer images, you can still target older macOS versions. We never looked into it.

Not sure the same is possible for arch.

ebbit1q: interesting, but we still depend on homebrew to give us deps, if github doesn't have a runner that is factually supported by homebrew I don't know how we can build at all for x64/intel, unless it's also possible to tell homebrew to install x64 packages?

tooomm: My understanding is right now that both macOS 14 binaries here are similar and both are targeting ARM arch. So they would not run on Intel Macs.
(Despite the manually changed file name.)

If somebody can validate that assumption with the artifacts build in this PR, that would be a good insight already!

It should also not only be about homebrew but also Xcode and the actual arch this is build on. Not sure about cross-os compilation on ARM arch or emulation to build for completely different architectures like Intel here? 🤔

ebbit1q: their file sizes are identical and the documentation clearly states that this is an arm runner, I don't think we need to test this, this is clearly an arm image, our only options are deprecating intel mac support or figuring out cross compilation

tooomm: There are [Universal Binaries](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) that combine both variants of the code. Xcode 12.2 and later support it.
I doubt that without additional configuration, an Intel compatible code is included currently, and we are correctly building such a universal one at the moment?
Homebrew deps need to match, too. 

If we can not figure out cross-compilation or the universal binary, there should be a note in the next release (similar to our [macOS 10.13 deprecation](https://github.com/Cockatrice/Cockatrice/releases/tag/2021-01-26-Release-2.8.0)).

ebbit1q:
@RickyRister do you have time to look into cross compiling on mac?

RickyRister:
It’s not an area I’m familiar in, but I can look into it

ebbit1q:
well I don't have a mac, you're using mac right?

.github/workflows/desktop-build.yml
ZeldaZach: I'm not willing to lose macOS Intel support at this time. I'd like to find an option to keep support somehow.

ebbit1q:
I have updated the runner to the new macos-15-intel image

in addition I have set the deployment target, hopefully this will mean we do not have to drop support

tooomm:
Any reason you did not use CMAKE_OSX_DEPLOYMENT_TARGET?

That is more in line with how we do CMake config in other places.

Somebody with old OS to test?! :)

ebbit1q:
MACOSX_DEPLOYMENT_TARGET seemed like the more generally used variable, it might be more compatible with other tools, I also put it as a general env variable and not just in the build step in case it'd affect dependencies

ebbit1q:
I found someone on discord that has experience setting up mac os virtual machines that offered to help

ebbit1q:
we will be forced to require macos 15 for now, if we move off homebrew to compiling stuff ourselves with vcpkg this will probably become possible, this pr is ready for merge so we can finally not stare at failing builds

ebbit1q:
I'd like to see this merged as an intermediary solution until we can get #6170 or similar, otherwise we have to keep getting failed builds for no reason

ebbit1q:
superseded by https://github.com/Cockatrice/Cockatrice/pull/6170

PR #6141, opened 2025-09-14
Have CardDatabase::getPreferredPrintingInfo respect card provider ID overrides (pinned printings)

### LINKED ISSUE #5960: Add a button to set the card printings in the deck to current defined pinned printings
It would be nice if, while editing decks, there was a button to set all cards in the currently opened deck to all known pinned printings. Obviously, cards with no pinned printings would get skipped.

Mass changes can be risky so I envision two potential safeguards: 

- Only auto-change cards that have a single printing present in the deck. So, for example, a deck's single Sol Ring will get set to the pinned default while the 30 different hand-selected Forest arts will be untouched.
- Also add a single-use undo button that will revert the deck back to the state prior to the mass change. This would be greyed out until the mass change button was clicked and greyed again on clicking.

BruebachL:
## Related Ticket(s)
- Closes #5960

There is an inconsistent display of pinned printings vs. set preference printings. Since pinned printings should always supersede set preference (because they are explicit per-card overrides rather than implicit overrides from the per-set configuration), the check for it should be moved up to the card database method so that callers do not have to perform this check themselves.

- Move override checks into getPreferredPrintingInfo

### DIFF
--- a/tests/carddatabase/filter_string_test.cpp
+++ b/tests/carddatabase/filter_string_test.cpp
@@ -18,10 +18,13 @@ class CardQuery : public ::testing::Test
 protected:
     void SetUp() override
     {
-        cat = CardDatabaseManager::query()->getCardBySimpleName("Cat");
-        notDeadAfterAll = CardDatabaseManager::query()->getCardBySimpleName("Not Dead");
-        truth = CardDatabaseManager::query()->getCardBySimpleName("Truth");
-        doctor = CardDatabaseManager::query()->getCardBySimpleName("Doctor");
+        CardDatabase *db = new CardDatabase(nullptr, QSharedPointer<NoopCardPreferenceProvider>::create());
+        db->loadCardDatabases();
+
+        cat = db->query()->getCardBySimpleName("Cat");
+        notDeadAfterAll = db->query()->getCardBySimpleName("Not Dead");
+        truth = db->query()->getCardBySimpleName("Truth");
+        doctor = db->query()->getCardBySimpleName("Doctor");
     }
     // void TearDown() override {}
 
@@ -70,7 +73,6 @@ QUERY(Color4, cat, "c!gw", false)
 int main(int argc, char **argv)
 {
     settingsCache = new SettingsCache;
-    CardDatabaseManager::getInstance()->loadCardDatabases();
 
     ::testing::InitGoogleTest(&argc, argv);
     return RUN_ALL_TESTS();


cockatrice/src/game/cards/card_database.cpp
```diff
@@ -390,6 +390,13 @@ PrintingInfo CardDatabase::getPreferredPrinting(const CardInfoPtr &cardInfo) con
         return PrintingInfo(nullptr);
     }
 
+    const auto &pinnedPrintingProviderId =
```
ebbit1q: unfortunately dbconverter uses this file as well (dbconverter is the included xml3 to xml4 conversion tool) it struggles with this as it's not in the mocks for it

ebbit1q:
in what way does this pr close the linked ticket by the way?

BruebachL:
> in what way does this pr close the linked ticket by the way?

See https://github.com/Cockatrice/Cockatrice/issues/5960#issuecomment-3289564268
There IS a convenience button for this, it just doesn't work the way it should.

cockatrice/src/database/card_database_querier.cpp
```diff
@@ -234,6 +246,13 @@ PrintingInfo CardDatabaseQuerier::getPreferredPrinting(const CardInfoPtr &cardIn
         return PrintingInfo(nullptr);
     }
 
+    const auto &pinnedPrintingProviderId =
+        SettingsCache::instance().cardOverrides().getCardPreferenceOverride(cardInfo->getName());
```
ebbit1q: adding this to the database querier still breaks the mocks in the tests, the mocks would need to have the cardOverrides and a new mock for getting the getCardPreferenceOverride

how hard would it be to abstract away the need for the settings cache from a base implementation of the database to use in our tests and dbconverter so we don't have to mock the settingscache for these?

cockatrice/src/database/card_database.cpp
```diff
@@ -15,7 +15,7 @@
 #include <algorithm>
 #include <utility>
 
-CardDatabase::CardDatabase(QObject *parent) : QObject(parent), loadStatus(NotLoaded)
+CardDatabase::CardDatabase(QObject *parent, std::shared_ptr<ICardPreferenceProvider> prefs) : QObject(parent), loadStatus(NotLoaded)
```
ebbit1q: is there a specific reason to use a std one instead of a qt QSharedPointer?

BruebachL: You're right, I had that rattling around the back of my mind, thanks for bringing it to the forefront.

dbconverter/src/main.h
```diff
@@ -2,11 +2,19 @@
...
 #include "../../cockatrice/src/database/parser/cockatrice_xml_4.h"
 
 class CardDatabaseConverter : public CardDatabase
 {
 public:
+    explicit CardDatabaseConverter()
+    {
+        // Replace querier with one that ignores SettingsCache
+        delete querier;
+        querier = new CardDatabaseQuerier(this, this, std::make_shared<NoopCardPreferenceProvider>());
```
ebbit1q: if NoopCardPreferenceProvider is only used in one place, why isn't it part of the dbconverter sources?

BruebachL: I didn't wanna break up the beautiful {main, mocks} structure and I couldn't figure out a way to include it in the mocks.h, main.h then pulls in mocks.h before the main.cpp does

PR #6142, opened 2025-09-14
Removed unused fields in GameMetaInfo and GameState

RickyRister:

Some fields in `GameMetaInfo` and `GameState` don't need to be there.

Also, we don't need to store a back-pointer to the `TabGame`.

- Remove the unused `roomGameTypes` and `phasesList` fields from `GameState`
- Don't store the parent `AbstractGame` as a field; just pass it to the `QObject` constructor as the parent.
- Relax the parent to any QObject, since there is nothing specifically requiring the parent to be a `TabGame`

### DIFF
--- a/cockatrice/src/game/game_state.cpp
+++ b/cockatrice/src/game/game_state.cpp
@@ -13,7 +13,7 @@ GameState::GameState(AbstractGame *parent,
                      bool _gameClosed)
     : QObject(parent), gameTimer(nullptr), secondsElapsed(_secondsElapsed), hostId(_hostId), isLocalGame(_isLocalGame),
       clients(_clients), gameStateKnown(_gameStateKnown), resuming(_resuming), currentPhase(_currentPhase),
-      activePlayer(0), gameClosed(_gameClosed)
+      activePlayer(-1), gameClosed(_gameClosed)
 {
 }
 


BruebachL:
I disagree with this. The choice of parent is conscious. While it is currently unused, conceptually you can not have a GameState without a Game, at least in my mind.

RickyRister:
fair enough

cockatrice/src/game/game_state.cpp
```diff
@@ -11,9 +11,9 @@ GameState::GameState(AbstractGame *_game,
                      bool _resuming,
                      int _currentPhase,
                      bool _gameClosed)
-    : QObject(_game), game(_game), secondsElapsed(_secondsElapsed), hostId(_hostId), isLocalGame(_isLocalGame),
+    : QObject(parent), gameTimer(nullptr), secondsElapsed(_secondsElapsed), hostId(_hostId), isLocalGame(_isLocalGame),
       clients(_clients), gameStateKnown(_gameStateKnown), resuming(_resuming), currentPhase(_currentPhase),
-      gameClosed(_gameClosed)
+      activePlayer(0), gameClosed(_gameClosed)
```
ebbit1q: I don't think it makes sense to initialise this to zero, the default for activePlayer in the game state pb is -1, to mean no player is the active player, this will be updated only when the game starts, the player needs to not crash when the active player is a random number otherwise the server could send defective game states to crash players

Issue #6127, opened 2025-09-13
[Beta/Dev] Player refactor regressions


BruebachL:
Using this ticket to track regressions introduced by #6090, #6112, and #6121

PR's that already addressed regressions:
#6122
#6123
#6124

Remaining regressions:
- Power/Toughness no longer being set automatically when a card enters the battlefield
- Reports of shortcuts not working (can't reproduce myself)

ebbit1q:
I can't reproduce the missing shortcuts either, I think the reporter might have just did something with their setup as we only have one report of that one

ebbit1q:
I'll leave this ticket up until after the new beta is tested this weekend

ebbit1q:
#6143 is also related

ebbit1q:
similar to the above crash, it also happens in the drawing logic at cockatrice/src/game/game_scene.cpp:88

BruebachL:
> similar to the above crash, it also happens in the drawing logic at cockatrice/src/game/game_scene.cpp:88

When does this occur/how do I reproduce?

ebbit1q:
it happens whenever you're a spectator and the game ends, very easy to reproduce in replays by skipping to the end of the replay

ebbit1q:
#6148 is set to fix more crashes

ebbit1q:
#6149 also

ebbit1q:
quite frustratingly I had to find out that a bug I thought was caused by my own refactor was caused by this one instead:
- creating a game as spectator does not use the gamestatechanged event to list the spectator in the player list

ebbit1q:
there exists an additional problem with concessions that has not been handled with [#6179](https://github.com/Cockatrice/Cockatrice/pull/6179)

players can send a command unconcede even though the game hasn't started yet, this results in a "RespGameNotStarted" error code, cockatrice ignores the return value from this message entirely and logs a concede anyway, this hints at improper event handling in cockatrice, this should be fixed and other events should be investigated in case one of them might cause client desynchronisation

ebbit1q:
#6178 should have fixed that

BruebachL:
> quite frustratingly I had to find out that a bug I thought was caused by my own refactor was caused by this one instead:
> 
>     * creating a game as spectator does not use the gamestatechanged event to list the spectator in the player list

Fixed by #6187

ebbit1q:
possibly related #6209

ebbit1q:
[...]

PR #6144, opened 2025-09-15
Fix segfault when creating token of opponent's card

RickyRister:
## Related Ticket(s)
- #6127

Cockatrice segfaults when you try to create a token of an opponent's card.

- Null check the `UtilityMenu` before calling `createAnotherTokenActionExists`

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -878,7 +878,8 @@ void PlayerActions::actCreateAnotherToken()
 
 void PlayerActions::setLastToken(CardInfoPtr cardInfo)
 {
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/player/player_actions.cpp
```diff
@@ -878,7 +878,8 @@ void PlayerActions::actCreateAnotherToken()
 
 void PlayerActions::setLastToken(CardInfoPtr cardInfo)
 {
-    if (cardInfo == nullptr || !player->getPlayerMenu()->getUtilityMenu()->createAnotherTokenActionExists()) {
+    if (cardInfo == nullptr || !player->getPlayerMenu()->getUtilityMenu() ||
```
ebbit1q:
```suggestion
    if (cardInfo == nullptr || player->getPlayerMenu()->getUtilityMenu() == nullptr||
```
this is easier to read, more explicit

PR #6112, opened 2025-09-09
Player refactor

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

## What will change with this Pull Request?
- this
- and this

PR #6145, opened 2025-09-15
Move `cards` and `filters` folder out of `game`

RickyRister:

With the recent refactor in #6112, it makes sense for the `game` folder to strictly be for code that has to do with an actual ongoing game. However, the `cards` and `filters` folder contains code that is used in many places. 

`game/cards` contains the card objects and the card database. 
`game/filters` contains the code for constructing filter strings and the filter model.

We should move those folders out of `game`.

- Moved `game/cards` and `game/filters` out of `game`.
  - They're now top-level folders under `src`
- Renamed `cards` to `card`.
- Also moved `game_specific_terms.h` into `card`.
- Fix the includes and cmakelist


cockatrice/src/card/game_specific_terms.h
ebbit1q, line 1: what's your reasoning for putting it in cards?

ebbit1q:
I'd like it if there was a distinction between "game card" and "database" because I don't think those components should be in the same location

BruebachL:
I was going to mention a database/ folder as well. I believe it would be helpful to group database classes and interfaces so it will be easier to transition away from the in-memory system should we ever choose to.

ebbit1q:
a separate database folder in cockatrice is a good start

cockatrice/src/card/game_specific_terms.h
RickyRister: It's only used in `card_info`

RickyRister:
What files should be moved to `database`? We obviously move the `card_database*` files, but what about `card_search_model` and `card_completer_proxy_model`?

cockatrice/src/card/game_specific_terms.h
ebbit1q: ok, that makes sense then

PR #6146, opened 2025-09-15
Fix HomeTab fullscreen freeze due to HomeStyledButton::paint()

BruebachL:
Don't setText() in paintEvent() which causes infinite recursion on MacOs

- Use a styleOption to paint base class directly instead.


RickyRister:
No longer freezes now. Thanks for the fix.

ebbit1q (approved):
seems fine, thanks

PR #6147, opened 2025-09-15
Make TabHome a managed tab

RickyRister:
## Related Ticket(s)
- Follow-up to #6114

`TabHome` is a single instance tab, but it's not a managed tab. Once the Home tab is closed, it can't be re-opened.
For consistency, we should make it a managed tab.

- Make the `Home` tab a managed tab, which means it has an entry in the tab menu.
  - Note that unlike the other managed tabs, we don't save the open status. We always open the home tab on startup.
- Add shortcut for home tab action.

https://github.com/user-attachments/assets/13b432a2-3502-408a-9039-0a3d57993a4b

## Screenshots

Screenshot 2025-09-15 at 2 14 40 AM


ebbit1q (approved):
excellent, thank you!

PR #6148, opened 2025-09-15
More player leaving fixes

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/game/player/player_manager.cpp
+++ b/cockatrice/src/game/player/player_manager.cpp
@@ -57,9 +57,9 @@ void PlayerManager::removePlayer(int playerId)
     if (!player) {
         return;
     }
-    players.remove(playerId);
-    emit playerCountChanged();
     emit playerRemoved(player);
+    emit playerCountChanged();
+    players.remove(playerId);
 }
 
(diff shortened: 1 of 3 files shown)


cockatrice/src/game/player/player_manager.cpp
```diff
@@ -57,9 +57,9 @@ void PlayerManager::removePlayer(int playerId)
     if (!player) {
         return;
     }
-    players.remove(playerId);
-    emit playerCountChanged();
     emit playerRemoved(player);
+    emit playerCountChanged();
+    players.remove(playerId);
```
ebbit1q, line 62: this seems like a good idea

BruebachL: Agreed. I didn't even test/don't even know if this fixes any bugs yet but it just seemed more logical.

ebbit1q: I mean that it *seems* like a good idea, it might also be a bad idea depending on how signals connected to playerCountChanged use the player count

ebbit1q: I don't notice any problems from this reorder

PR #6149, opened 2025-09-15
check for null zone during card item teardown

BruebachL:
Took 1 hour 32 minutes

Took 24 seconds

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/game/board/card_item.cpp
+++ b/cockatrice/src/game/board/card_item.cpp
@@ -191,7 +191,13 @@ void CardItem::setAttachedTo(CardItem *_attachedTo)
             attachedTo->getZone()->reorganizeCards();
         }
     } else {
-        emit zone->cardAdded(this);
+        // If the zone is being torn down, it might already be null by the time a card tries to re-attach all its
+        // attached cards
+        if (zone == nullptr) {
+            deleteLater();
+        } else {
+            emit zone->cardAdded(this);
+        }
     }
 
     if (zone != nullptr) {


cockatrice/src/game/board/card_item.cpp
```diff
@@ -191,7 +191,13 @@ void CardItem::setAttachedTo(CardItem *_attachedTo)
             attachedTo->getZone()->reorganizeCards();
         }
     } else {
-        emit zone->cardAdded(this);
+        // If the zone is being torn down, it might already be null by the time a card tries to re-attach all its
```
ebbit1q: I wouldn't say they're being reattached, it's more like they're being unattached

cockatrice/src/game/board/card_item.cpp
```diff
@@ -185,13 +185,25 @@ void CardItem::setAttachedTo(CardItem *_attachedTo)
     gridPoint.setX(-1);
     attachedTo = _attachedTo;
     if (attachedTo != nullptr) {
-        emit attachedTo->zone->cardAdded(this);
-        attachedTo->addAttachedCard(this);
-        if (zone != attachedTo->getZone()) {
-            attachedTo->getZone()->reorganizeCards();
+        // If the zone is being torn down, it might already be null by the time a card tries to un-attach all its
```
ebbit1q, line 188: setAttachedTo does not necessarily happen during zone teardown, this branch here specifically happens when it's normally attaching to a card, it's very unlikely to be tearing down the zone during this

cockatrice/src/game/board/card_item.cpp
```diff
@@ -185,13 +185,25 @@ void CardItem::setAttachedTo(CardItem *_attachedTo)
...
+        } else {
+            emit attachedTo->zone->cardAdded(this);
+            attachedTo->addAttachedCard(this);
+            if (zone != attachedTo->getZone()) {
+                attachedTo->getZone()->reorganizeCards();
+            }
         }
     } else {
-        emit zone->cardAdded(this);
+        // If the zone is being torn down, it might already be null by the time a card tries to un-attach all its
```
ebbit1q, line 200: this code path is also used when normally unattaching cards

cockatrice/src/game/board/card_item.cpp
BruebachL: I'd say edge cases deserve to be documented.

PR #6150, opened 2025-09-15
Animation safety.

BruebachL:
Took 12 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


ebbit1q (changes requested):
to me the current design of the animation system doesn't seem very safe, I think we should investigate if we can redesign this part instead

ebbit1q:
it's also worth noting that this crash case is very rare, in my impression it has to do with a player exiting the game while one of their cards has the animation timer running, something that happens rarely

PR #6152, opened 2025-09-15
[Home Tab] Only use normal layout cards as a background source

BruebachL:
The art crop doesn't work for non-normal layout cards.


ebbit1q (approved):
this makes sense, thanks

PR #6153, opened 2025-09-15
[Home Tab] Disable random art crop shuffle if frequency is set to 0.

BruebachL:
Cards change way too fast if the timer is set to 0 seconds. It's meant to disable it, which this PR addresses.

- Stop cardChange timer if home tab background shuffle frequency is set to 0.

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -684,7 +684,7 @@ void AppearanceSettingsPage::retranslateUi()
     themeLabel.setText(tr("Current theme:"));
     openThemeButton.setText(tr("Open themes folder"));
     homeTabBackgroundSourceLabel.setText(tr("Home tab background source:"));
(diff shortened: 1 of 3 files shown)


cockatrice/src/dialogs/dlg_settings.cpp
```diff
@@ -684,7 +684,7 @@ void AppearanceSettingsPage::retranslateUi()
     themeLabel.setText(tr("Current theme:"));
     openThemeButton.setText(tr("Open themes folder"));
     homeTabBackgroundSourceLabel.setText(tr("Home tab background source:"));
-    homeTabBackgroundShuffleFrequencyLabel.setText(tr("Home tab background shuffle frequency:"));
+    homeTabBackgroundShuffleFrequencyLabel.setText(tr("Home tab background shuffle frequency (0 to disable):"));
```
ebbit1q: https://doc.qt.io/qt-6/qabstractspinbox.html#specialValueText-prop

BruebachL: Thanks, I didn't know about this.

PR #6156, opened 2025-09-17
Common server refactor

ebbit1q:
we've been refactoring and moving a lot of cockatrice files lately, I think it's time we also do the same to the shared server components

- this creates a new folder that contains all files related to running the server or local server
- and a sub folder in there for only the game on the server

I plan to also revamp the classes a bit by trying to make separate player classes based on an abstract player, separating spectator players from game players and soon, draft players #6096


ebbit1q:
this is ready for review and testing

ebbit1q:
I've fixed all the issues I found while playing around, judges, multiplayer, etc, have all been tested

common/server/game/server_abstract_participant.h
```diff
@@ -0,0 +1,194 @@
...
+    cmdChangeZoneProperties(const Command_ChangeZoneProperties &cmd, ResponseContainer &rc, GameEventStorage &ges);
+
+    Response::ResponseCode processGameCommand(const GameCommand &command, ResponseContainer &rc, GameEventStorage &ges);
+    void sendGameEvent(const GameEventContainer &event);
+
+    virtual void
+    getInfo(ServerInfo_Player *info, Server_AbstractParticipant *recipient, bool omniscient, bool withUserInfo);
+};
+
+class Server_Spectator : public Server_AbstractParticipant
```
RickyRister: Even though `Server_Spectator` doesn't have much in the cpp file, it might be better to put in a separate file anyways so that it's easier to find when browsing.

ebbit1q: sure

common/server/game/server_game.h
BruebachL, line 1: Might be worth thinking about structuring the server game like the client game (i.e. gameState, gameMetaInfo, playerManager and such)

BruebachL (approved):

PR #5118, opened 2024-10-05
Code refactoring

LunyaticCat:
## Short roundup of the initial problem
All files are currently in the root directory, making it really hard for new devs to join the project.

## What will change with this Pull Request?
every cockatrice files are put into different folders matching what they do.

PR #6158, opened 2025-09-20
Move more files

ebbit1q:
## Related Ticket(s)
- https://github.com/Cockatrice/Cockatrice/pull/5118

many files in cockatrice have been sorted based on "vibes" instead of actual content

- move a lot of files, I haven't reviewed all the files but mostly aimed to give a simple "general improvement" to the current sorting
- as a side effect, I've replaced the cmakelists with a simple list of all cpp files, this broke stuff that I fixed by making them work the "proper way" which was simply making one implementation file and replacing one include with a forwards declaration

my interpretation of the following folders is like this: (in order of priority)
- settings > everything related to settings
- picture_loader > everything related to loading card pictures
- filters > everything related to filtering cards in the db
- database > everything related to the db
- tabs > all tabs no matter their contents
- game > stuff that you can see in tab_game
- server > stuff that you can see while connected to a server
- card > stuff related to cards
- deck > stuff related to the deck loader
- dialogs > all dialogs that aren't in game
- interface > stuff you can see
- utility > stuff that is used from multiple locations
- client > stuff? I think this folder should be eliminated at some point

this pr does not yet sort all files according to this scheme, it's just a step in the right direction so it can stop bothering me as much while I'm working


BruebachL (approved):

BruebachL:
Having a src/server folder feels a bit weird but I can not think of a better name either.

ebbit1q:
oops

PR #6159, opened 2025-09-20
add phase to delete arrows in to protocol

### LINKED ISSUE #6035: would it be possible to keep the red arrows either between phases or just during all combat phases?

ebbit1q:
## Related Ticket(s)
- Fixes #6035

some users would prefer their arrows not to disappear when they switch phases

- add phase to delete arrows in to protocol
- add setting to cockatrice to use this new parameter

### DIFF
--- a/cockatrice/src/game/board/card_item.cpp
+++ b/cockatrice/src/game/board/card_item.cpp
@@ -3,6 +3,7 @@
 #include "../../client/settings/cache_settings.h"
 #include "../../interface/widgets/tabs/tab_game.h"
 #include "../game_scene.h"
+#include "../phase.h"
 #include "../player/player.h"
 #include "../zones/card_zone.h"
 #include "../zones/logic/view_zone_logic.h"
@@ -280,7 +281,7 @@ void CardItem::drawArrow(const QColor &arrowColor)
     int phase = 0; // 0 means to not set the phase
     if (SettingsCache::instance().getDoNotDeleteArrowsInSubPhases()) {
         int currentPhase = game->getGameState()->getCurrentPhase();
-        phase = currentPhase + 4;
+        phase = Phases::getLastSubphase(currentPhase) + 1;
     }
     ArrowDragItem *arrow = new ArrowDragItem(arrowOwner, this, arrowColor, phase);
     scene()->addItem(arrow);


ebbit1q:
~~this pr depends on both #6158 and #6156, changes that are new to this pr start at [2fc08e7](https://github.com/Cockatrice/Cockatrice/pull/6159/commits/2fc08e7dcc45aa21f6e9a23ae1b4178cf2be533f)~~ rebased

ebbit1q:
right now I haven't established a way to specify which phases are subphases and it simply keeps the arrows around for the four next phases by default, I've noticed there are about three locations that handle how phases are defined and I'd like to consolidate them a little

ebbit1q:
the server phases are complete and ready though, I could split them off into a separate pr if it speeds up the review process

common/pb/command_create_arrow.proto
```diff
@@ -2,15 +2,33 @@ syntax = "proto2";
 import "game_commands.proto";
 import "color.proto";
 
+// Command to draw an arrow from cards to either other cards or a player
```
RickyRister, line 5: I left a comment on the draft implementation PR about how you should move comments on existing proto to a separate PR. Can you also move these comments to that PR too?

RickyRister:
I'm not convinced that we want to add logic to make the arrow to stick around for a specific number of phases. 

For now, I'd rather we just add a binary "do not delete arrow on phase change" field to the proto, plus a toggle to the client. Perhaps with an easy way for the client to change the toggle in-game. 

It's not too hard to clear the arrows with ctrl+R anyways. If someone has that toggle on, that person clearly knows what they're doing.

ebbit1q:
I think this implementation is intuitive to how users expect it to work, also I already made it why simplify it now?

RickyRister:
Fair enough. 

In the future, if we want to add an additional option to have arrows that never clear, we could just use values such as `-1` to indicate that.

One thing that I'm iffy on is that the arrows always get deleted when moving backwards through the phases. There are plenty of scenarios where someone draws arrows on attacks, advances to a later phase, then the opponent requests to back up to an earlier combat phase because they want to respond.

For a first implementation though, I suppose it's good enough.

ebbit1q:
that's not how the current implementation is though, it remembers in which phase it created the arrow, it will only delete the arrow if the phase moves to before it was created, but not if it is moved back to the phase it was created

RickyRister:
Oh, in that case, I’m happy with the implementation

ebbit1q:
pleasantly surprised that there were no issues rebasing this

ebbit1q:
considering #1740 would probably change phases in a big way, and that's what I want to work on next I'd prefer this to be merged in sooner than that so I've added these definitions so they are ready to use now.

ZeldaZach (approved):

PR #6163, opened 2025-09-21
Sort pb cmake and add comments to proto files

RickyRister:

- The files in `common/pb/CMakeLists` isn't sorted in alphabetical order
- The proto files don't have comments.

- Sort the files in the proto cmake by alphabetical order
- Copy over the comments for the existing proto files from #6096 and #6159
- Clean up the comments


ebbit1q:
sure, though one of the reasons why I didn't want to split them out is because of the merge conflicts :/

Issue #6162, opened 2025-09-21
Use same C++ package manager in CI across our OS


tooomm:
In our GitHub Actions CI/CD pipeline we currently use 
- `VCPKG` for Windows builds 
- `Homebrew` for macOS builds
- Linux is a bit special as we use and build dedicated docker container per distro and *apt install* there

**[VCPKG](https://github.com/microsoft/vcpkg) is an open-source cross-platform package manager for C++ and supports not only Windows, but also macOS and Linux.**
I know that @brunoalr was at one point looking into **[Conan](https://github.com/conan-io/conan), which would be another popular C++ package manager with support for all our platforms.**
Looks like developers are usually in favor of one of the two big open source tools, and each have some specifics of their own.
Like VCPKG uses commit baselines over specific versions and therefore guarantees the package are tested together for that baseline. But versions can be pinned as well if needed.
It also supports integration with the GitHub dependency graph.

I'm not the person to compare the two or suggest one over the other, but I really like that VCPKG received support in Dependabot recently to help with keeping up with version updates automatically. And because it's affiliated with Microsoft as well, I expect a growing and stable support and it being a first-class citizen.
We also use it for Windows since a long time and expanding to macOS is maybe easier than implementing Conan for both from scratch? The VCPKG repo is also already included in our repo as submodule.
Binary caching for the compiled dependencies work nicely for the Windows builds too, hence expecting the same on macOS (improvement over Homebrew). But probably there are similar options for Conan.

**Open for any support on this and achieving a more harmonized, stable and easier to maintain pipeline down the road.**
We are regularly facing issues with Homebrew, hence at least for macOS, I think using one of them can help us to maintain the build pipeline more easily and have a stable, reliable and uniform setup with also shorter build times.

Also see [vcpkg.io](https://vcpkg.io) or [conan.io](https://conan.io) with a list of packages, further details and links to the docs.

brunoalr:
Hey there, @tooomm, long time no see :) Yeah, I have been experimenting with Conan (mostly for personal learning) in order to check for faster compile times.

I didn't finish the experiment, but achieved some progress in https://github.com/brunoalr/Cockatrice/pull/18/:
- Managed to compile dependencies (except qt, which is installed via aqt-installer) and Cockatrice binaries using Conan on Windows + MinGW + ccache + ninja
- Unfortunately the client does not run and I couldn't figure out why so far

[...]

PR #6165, opened 2025-09-22
Cross-compile intel mac and bump version from 13 to 14

RickyRister:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -276,7 +276,7 @@ jobs:
         if: ${{matrix.soc == 'Intel'}}
         shell: bash
         run: |
-          softwareupdate --install-rosetta --force
+          softwareupdate --install-rosetta --agree-to-license
           arch --x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
           echo "alias brew='arch --x86_64 /usr/local/Homebrew/bin/brew'" >> aliases.sh
 


ebbit1q:
awesome, thanks for looking into this ricky! this pr would supersede #6140

RickyRister:
Save your compliments for after I figure out if this even works in the first place.
If it doesn't, then we'll have to go back to doing something more drastic.

RickyRister:
Yeah, seems like pre-installed homebrew only downloads dependencies for the host architecture. 

From what I can find online, you need to [install a separate x86 homebrew instance](https://www.metasimple.org/2025/01/11/dual-brew-setup.html) on top of compatibility layer if you want to download x86 dependencies. 

Might still be easier than the alternative solution. We could add it as another step in the pipeline

ebbit1q:
would it be easier to use vcpkg or something like that instead of homebrew? https://github.com/Cockatrice/Cockatrice/issues/6162

RickyRister:
Yeah, probably

brunoalr:
> would it be easier to use vcpkg or something like that instead of homebrew? #6162

I have a WIP, untested branch for that: https://github.com/brunoalr/Cockatrice/pull/22 if you want to take a look at it.

brunoalr:
btw, it seems like this PR will be needed after all if we want to maintain support for macos13 and/or x86 builds as github is retiring macos13 runners: https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

EDIT: from the link above, GH ~will offer~ already has available a new `macos-15-intel` runner, so there's another option to consider

ebbit1q:
I think if we could use the new intel runner and create a binary that works on older mac versions we'd be fine yeah

tooomm:
> I think if we could use the new intel runner and create a binary that works on older mac  versions we'd be fine yeah

Great hint @brunoalr!
It's not listed at https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories though. 

@ebbit1q Looking at your other PR, seems the macos-15-intel runner is working for public open source repos for free as well. 


[Universal Binaries](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) for Mac would be the ideal solution.

Vcpkg allows for cross-compilation for different architectures via triplet configuration as well.
While universal dependencies are also possible, we can probably combine two individual x64 and arm binaries into a universal one with lipo.

ebbit1q:
if a universal binary is just two binaries in one bundle, it'd save people bandwidth to just offer the binary they need separately

ebbit1q:
#6140 will drop support for 13 and 14 on intel, I'd still really like something like this pr in order to support at least macos 14 on intel, perhaps if we used vcpkg or another source for our dependencies (qt + protobuf and ssl afaik, it's not a long list) then we can get older versions working!

brunoalr:
> Great hint @brunoalr!
It's not listed at [docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) though.

Commenting just for completeness: That's true, but in practice the runners are already online (see https://github.com/OpenImageDebugger/OpenImageDebugger/pull/719 for example), so I think MSFT/GH just forgot to update the docs haha Anyways, I expect that the full, formal, properly documented of the runners will be available very soon

ebbit1q:
thank you for your help exploring this option @RickyRister ! it seems like because github has released a new intel runner we will not need cross compilation for now...

we will probably need cross compilation in the future when this macos-15-intel image is taken out of service, for now though I'll close this pr!

Issue #528, opened 2015-01-05
Deck legality checker


anoda9:
A deck legality checker would be a good feature to have. The main criteria to check is the 4-of rule, since that is unverifiable by the opponent. Checking format specific banlists would be nice also, but that should be a lower priority, since it's easily verifiable by the opponent. (if you try and drop a mox in a legacy match, it's pretty obvious your deck is illegal). The 15 card sideboard limit is also sometimes unverifiable (if a person starts their own game and loads a deck before their opponent gets there, their opponent can't see the "player_x loaded a deck with 15 sideboard cards." message), so that would be good to check also. 

I think it'd be best if the legality checker ran in parallel with starting a game, rather than being a requirement for the game to start. (i.e. you would start a game, then the server would check your decks as you played out the first few turns, rather than forcing the deck check step to be completed before the game could begin.) 

Once completed, the server could mark your deck with a green checkmark, or put the hash in green text for being legal, red text for illegal. 

The 4-of criteria would be disabled if the game type was "Draft/Sealed".

ZeldaZach:
I wrote a deck checker for the format "Historic Standard" a while ago [found here](http://www.mtgheadquarters.com/zach/confirmDeck.php).

Users should know about [tappedout.net](tappedout.net) and it has a legality checker built into it.

tooomm:
I don't know about tappedout. Is there maybe a way to interact with that site the same way cockatrice interacts with deckstats.net to perform a legality check for yourself?

But anyway, anoda9 is talking about some kind of check that tells your opponent that you're not cheating. Or to keep it general, which formats your deck is legal in.
Since that is a specific game related check I like the idea of the server offering this opportunity. Still cockatrice needs to support this kind of deck checks with the server. Not sure about traffic, but if the server does the shuffling already, he knows about all the cards?

-----

The problem with the sideboard is a point!
What if players (in network games, not local) are just allowed to load decks if the game is filled with all players?
(combine with locked functionality later on? so you can start the game/start loading decks in a 4/6 player edh game as soon as the creator locks the room - if he unlocks the room, all loaded decks needs to get reloaded to avoid cheating. that should be an issue for itself if others see the need for it, too)

anoda9:
@ZeldaZach Yeah, I was more thinking of it as a way to make sure your opponent's deck is legal, rather than for checking your own deck. 

> What if players (in network games, not local) are just allowed to load decks if the game is filled with all players?
[...]

PR #6166, opened 2025-09-22
Deck format legality checker

BruebachL:
## Related Ticket(s)
-  Related to #528 but does not implement quantity limits

Users are constantly asking if card XYZ is legal yet or in a specific format. This adds an in-client way of checking format legality for the cards in your deck easily (Selecting a format is done through a combobox, which displays all format-property entries of all cards in the database. Once a format is selected, cards which either have no format-property name or are not "legal" in the format are marked through a red entry in the row.)

- Add format legality to the decklist and persist it in the decklist.cod
- Check this format legality in the UI

## Screenshots
image
image


libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -632,6 +661,42 @@ QList<QString> *DeckListModel::getZones() const
...
 
+bool DeckListModel::isCardLegalForCurrentFormat(const CardInfoPtr cardInfo)
+{
+    if (!deckList->getGameFormat().isEmpty()) {
+        if (cardInfo->getProperties().contains("format-" + deckList->getGameFormat())) {
+            return cardInfo->getProperty("format-" + deckList->getGameFormat()) == "legal";
+        } else {
+            return false;
+        }
+    }
```
ebbit1q, line 664: teeny tiny nit, why does it use else above this but not here?

cockatrice/src/deck/deck_list_model.cpp
```diff
@@ -632,6 +661,42 @@ QList<QString> *DeckListModel::getZones() const
...
+    }
+    return true;
+}
+
+void DeckListModel::refreshCardFormatLegalities()
+{
+    InnerDecklistNode *listRoot = deckList->getRoot();
+
+    for (int i = 0; i < listRoot->size(); i++) {
+        auto *currentZone = dynamic_cast<InnerDecklistNode *>(listRoot->at(i));
```
ebbit1q: why dynamic cast this if you're not going to nullcheck it? might as well use a static cast right?

BruebachL: I just copied this from somewhere else where we loop over the nodes tbh.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: Saving a level of indentation as decreed by Zach.

ebbit1q: tell zach he's wrong then? alternatively you can remove the else above but imo code reads better with else statements, but most importantly don't mix both styles it trips up my comprehension

cockatrice/src/deck/deck_list_model.cpp
ebbit1q: imo the only reason to use dynamic cast is when you expect it to be possible to fail, the original code must have been wrong simply

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_format_legality_filter_widget.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    exactMatchMode = checked;
+    toggleButton->setText(exactMatchMode ? tr("Mode: Exact Match") : tr("Mode: Includes"));
+    updateFormatFilter();
+}
+
+void VisualDatabaseDisplayFormatLegalityFilterWidget::syncWithFilterModel()
+{
+    // Temporarily block signals for each button to prevent toggling while updating button states
+    for (auto it = formatButtons.begin(); it != formatButtons.end(); ++it) {
```
ebbit1q, line 173: is there a specific reason to use an iterator instead of simply using an expression on formatButtons.values() here?

ebbit1q: ah, disregard you are modifying the map

ebbit1q:
I notice your changes only consider cards legal if their format tag states "legal" but ignores cards that are "restricted", this makes most vintage decks not validate

it might be out of scope for this pr but it'd be nice if it could also validate card counts, this would require some metadata to be added to the database, eg:
```xml
<info>..</info>
<formats>
    <format-commander count=1>legal</format-commander>
    <format-commander count=0>banned</format-commander>
    <format-vintage count=4>legal</format-vintage>
    <format-vintage count=1>restricted</format-vintage>
    <format-vintage count=0>banned</format-vintage>
    ..
</formats>
<sets>..</sets>
<cards>..</cards>
```

alternatively we can simply make the restricted cards another colour, like yellow, as long as they're not red like the banned cards I think this pr is good :+1:

cockatrice/src/deck/deck_list_model.cpp
```diff
@@ -665,10 +665,34 @@ bool DeckListModel::isCardLegalForCurrentFormat(const CardInfoPtr cardInfo)
...
+    }
+    return true;
+}
+
+bool DeckListModel::isCardQuantityLegalForCurrentFormat(const CardInfoPtr cardInfo, int quantity)
+{
+    if (cardInfo->getMainCardType() == "Basic Land") {
+        return true;
+    }
+    if (cardInfo->getText().contains("A deck can have any number")) {
```
ebbit1q: stuff like this is fragile and dependent on the game, please add it to the database format and oracle instead!

cockatrice/src/deck/deck_list_model.cpp
```diff
@@ -665,10 +665,34 @@ bool DeckListModel::isCardLegalForCurrentFormat(const CardInfoPtr cardInfo)
...
+        return true;
+    }
+    if (!deckList->getGameFormat().isEmpty()) {
+        if (cardInfo->getProperties().contains("format-" + deckList->getGameFormat())) {
+            QString formatLegality = cardInfo->getProperty("format-" + deckList->getGameFormat());
+            if (formatLegality == "legal" && quantity <= 4) {
```
ebbit1q: not every format is a four of format!

ebbit1q:
please don't add game dependent code to the decklist format, I notice my initial proposal to a card xml update didn't include basics so here is an update:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<cockatrice_carddatabase version="4" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://raw.githubusercontent.com/Cockatrice/Cockatrice/master/doc/carddatabase_v4/cards.xsd">
    <info>..</info>
    <formats>
        <format-commander count=1>legal</format-commander>
        <format-commander count=0>banned</format-commander>
        <format-vintage count=4>legal</format-vintage>
        <format-vintage count=1>restricted</format-vintage>
        <format-vintage count=0>banned</format-vintage>
        ..  
    </formats>
    <sets>..</sets>
    <cards>
        <card>
            <name>Island</name>
            <prop>
                <format-commander count="n">legal</format-commander>
                <format-vintage count="n">legal</format-vintage>
                ..  
            </prop>
            ..  
        </card>
        ..  
    </cards>
</cockatrice_carddatabase>
```

ebbit1q:
this idea is really nice, do you plan to reopen it in the future?

libcockatrice_deck_list/libcockatrice/deck_list/tree/deck_list_card_node.h
```diff
@@ -51,6 +51,7 @@ class DecklistCardNode : public AbstractDecklistCardNode
     QString cardSetShortName; ///< Short set code (e.g., "NEO").
     QString cardSetNumber;    ///< Collector number within the set.
     QString cardProviderId;   ///< External provider identifier (e.g., UUID).
+    bool formatLegal;         ///< Format legality
```
RickyRister, line 54: Why is format legality stored as part of the underlying deck model?
It's derived information, and is only used in the deck editor, so it feels weird to store in the underlying deck model, instead of something higher up.

BruebachL: It needs to be evaluated at runtime (the card count matters) and caching it prevents us from querying it each time the views request a redraw.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -263,6 +272,44 @@ void DeckEditorDeckDockWidget::createDeckDock()
...
+void DeckEditorDeckDockWidget::initializeFormats()
+{
+    QMap<QString, int> allFormats = CardDatabaseManager::query()->getAllFormatsWithCount();
+
+    formatComboBox->clear(); // Remove "Loading Database..."
+    formatComboBox->setEnabled(true);
+
+    // Populate with formats
+    for (auto it = allFormats.constBegin(); it != allFormats.constEnd(); ++it) {
+        QString displayText = QString("%1 (%2)").arg(it.key()).arg(it.value());
```
RickyRister: Why do we display counts next to the format name? 
It's a really weird piece of data to display here. I don't think anyone cares about the number of cards in a format when they're selecting a format.

RickyRister:
Once you've selected a format for a deck, is there a way to make the deck have a blank format again?

RickyRister:
It looks awkward when the dropdown selector is sandwiched between two text input fields. 
Maybe move `Comments` back under `Deck name` and move `Format` down more.

Screenshot 2025-12-11 at 5 07 31 PM

libcockatrice_deck_list/libcockatrice/deck_list/tree/deck_list_card_node.h
RickyRister: I'm not the biggest fan of storing it there, because now it means every `DeckList` in transport is lugging around nodes with a meaningless legality field.
If there isn't a better way to do it, then I suppose it's fine.

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_format_legality_filter_widget.cpp
```diff
@@ -0,0 +1,205 @@
...
+    flowWidget = new FlowWidget(this, Qt::Horizontal, Qt::ScrollBarAlwaysOff, Qt::ScrollBarAsNeeded);
+    layout->addWidget(flowWidget);
+
+    // Create the spinbox
+    spinBox = new QSpinBox(this);
+    spinBox->setMinimum(1);
+    spinBox->setMaximum(getMaxMainTypeCount()); // Set the max value dynamically
+    spinBox->setValue(150);
+    layout->addWidget(spinBox);
+    connect(spinBox, QOverload<int>::of(&QSpinBox::valueChanged), this,
```
RickyRister: nit: Qt recommends using the macro
```suggestion
    connect(spinBox, qOverload<int>(&QSpinBox::valueChanged), this,
```

RickyRister:
wait why did you move the dropdowns under the tag box? I still think it looks better when the two deck metadata dropdowns are above it.
The group by dropdown can go under it since it's not deck metadata.

Screenshot 2025-12-11 at 5 19 29 PM

BruebachL:
> wait why did you move the dropdowns under the tag box? I still think it looks better when the two deck metadata dropdowns are above it. The group by dropdown can go under it since it's not deck metadata.
> Screenshot 2025-12-11 at 5 19 29 PM

I think it looks better if everything is grouped like that.

libcockatrice_deck_list/libcockatrice/deck_list/tree/deck_list_card_node.h
BruebachL: There isn't as far as I know. Views query the models and the only way to do what we want is to use a style proxy in the first place (which runs into the whole "querying on every refresh" problem [which it does *a lot*, database calls are not feasible]), which also queries the model.

It's worth noting that it's also just a boolean.

RickyRister:
Can we just move the metadata dropdowns above the tag box for now so it's the same as before?
We can bikeshed about moving things around in the ui in a separate discussion.

RickyRister:
I don't know if that's something you're intending to deal with in a future PR, but changing the format in the deck editor does not set the isModified flag.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -263,6 +272,46 @@ void DeckEditorDeckDockWidget::createDeckDock()
...
+
+    if (!deckModel->getDeckList()->getGameFormat().isEmpty()) {
+        deckModel->setActiveFormat(deckModel->getDeckList()->getGameFormat());
+        formatComboBox->setCurrentIndex(formatComboBox->findData(deckModel->getDeckList()->getGameFormat()));
+    } else {
+        // Ensure no selection is visible initially
+        formatComboBox->setCurrentIndex(-1);
+    }
+
+    connect(formatComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged), this, [this](int index) {
```
RickyRister, line 306: nit: QOverload is used over the macro in quite a few places in this PR, in case you care about stuff like that

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_format_legality_filter_widget.cpp
```diff
@@ -0,0 +1,205 @@
...
+    toggleButton->setToolTip(tr("Filter mode (AND/OR/NOT conjunctions of filters)"));
+}
+
+void VisualDatabaseDisplayFormatLegalityFilterWidget::createFormatButtons()
+{
+    // Iterate through main types and create buttons
+    for (auto it = allFormatsWithCount.begin(); it != allFormatsWithCount.end(); ++it) {
+        auto *button = new QPushButton(it.key(), flowWidget);
+        button->setCheckable(true);
+        button->setStyleSheet("QPushButton { background-color: lightgray; border: 1px solid gray; padding: 5px; }"
```
ZeldaZach, line 67: Does this mean users can't edit the style themselves?

BruebachL: We have this problem anywhere where we style widgets to make them a little nicer than just the Qt default (which ends up looking cheap).

Solution for this is a default theme Qt stylesheet for the application where we perform such stylings. I am hoping to outpace our relatively inactive custom theme maintainers with a default theme before it becomes *incredibly* relevant to them.

PR #6168, opened 2025-09-22
build: update vcpkg

brunoalr:
## Related Ticket(s)
- Ground work for cockatrice#6162

**tl;dr**: we need to update vcpkg.

We want to try and unify the dependency managers as much as possible across our supported OSes. In particular, homebrew has been giving some headaches, especially on older macOS versions. Since we already have vcpkg in place for Windows, it is not worth it to change to another manager such as Conan.

Since the Linux builds are shipped as their respective native package managers (apt, rpm, etc), it is not trivial to just replace them with a system agnostic manager, so we will focus only on standardizing Windows and macOS.

In a preliminary attempt to use vcpkg on macOS, I noticed that some packages don't build/install in the current version of the vcpkg submodule. Updating it solved this problem.

⚠️ I tried to use the qt libs managed by vcpkg, but the builds just fail after a long time either due to disk getting full or very long file paths that are not supported by Windows -> Decided to keep using jurplel/install-qt-action@v4 ⚠️  

- Update vcpkg git submodule to a newer version

## Remaining work in this PR
- [ ] Manual testing on Windows builds

## Follow-up task/PR:
- [ ] Use vcpkg on macos builds as well. WIP: brunoalr/cockatrice#22


ebbit1q:
the build times  for windows skyrocketed, is this expected? will it return to normal?

brunoalr:
> the build times for windows skyrocketed, is this expected?

yes, it is expected. the first build *of a branch* after an update will take longer as vcpkg will rebuild the dependencies (almost certainly from scratch - I **think** vcpkg plays it safe and does not reuse build artifacts from older versions). 

> will it return to normal?

after a successful build, the vcpkg cache will be updated and the dependencies won't be recompiled. if you trigger a manual re-build for windows in this PR, I would expect that the build time will be back to around 25min.

if this pr is merged, the next full build on master will be slower (GHA caches are stored per branch), but the subsequent ones will also return to normal

PR #6170, opened 2025-09-24
build: use vcpkg for most dependencies on macos

brunoalr:
## Related Ticket(s)
- Fixes cockatrice#6162

Homebrew [changed support](https://docs.brew.sh/Installation#2) of macos 13 to [Tier 3](https://docs.brew.sh/Support-Tiers#tier-3), which means in the maintainers own words that we "[...] get a poor but not completely broken experience using Homebrew.
⚠️ Please notice that at the time of writing, the former link still says "e.g. as of March 2025, macOS Monterey/12 and older" (digressing further: maybe I should open a PR to fix the docs :P ), but the tool itself already gives a big [warning](https://github.com/brunoalr/Cockatrice/actions/runs/17963262538/job/51090782340#step:3:114) about it in our CI pipelines.

There have been some discussions between the devs and the conclusion is that we want to step away from Homebrew (TODO: add links to the actual discussions).

- macos builds will use vcpkg as the dependency manager to install all packages but qt
- changed installation of ccache on macos from brew to https://github.com/jianmingyong/ccache-action

## More detailed explanation of the choices taken
- qt has been shown very hard to get compiled via vcpkg on both Windows and macOS -> decided to install it via the same GH action (https://github.com/jurplel/install-qt-action) that we currently use on Windows
- on macos 13, brew now tries to compile ccache and a lot of other (unrelated/dependencies) packages like `rust` from source. https://github.com/jianmingyong/ccache-action was the first off-the-shelve solution that I found that works

## Impact
- For basically the same reason as https://github.com/Cockatrice/Cockatrice/pull/6168, it is expected that the first build in a given branch will take much longer on all macOS runners, but the next builds should get back to a faster speed
- TODO: compare the build time of a build with a *hot vcpkg cache* from this PR to the build time on master. Current expectation is that they should take about the same time.

## Remaining work in this PR
- [x] Manual testing arm builds on arm hw -> works
- [x] Manual testing x86 build on arm hw -> works
- [ ] Manual testing x86 build on x86 hw

### DIFF
--- a/cockatrice/src/game/game_event_handler.cpp
+++ b/cockatrice/src/game/game_event_handler.cpp
@@ -340,6 +340,8 @@ void GameEventHandler::eventPlayerPropertiesChanged(const Event_PlayerProperties
             while (playerIterator.hasNext())
                 playerIterator.next().value()->updateZones();
 
+            emit logConcede(eventPlayerId);
+
             break;
         }
         case GameEventContext::UNCONCEDE: {
@@ -348,6 +350,8 @@ void GameEventHandler::eventPlayerPropertiesChanged(const Event_PlayerProperties
             QMapIterator<int, Player *> playerIterator(game->getPlayerManager()->getPlayers());
             while (playerIterator.hasNext())
                 playerIterator.next().value()->updateZones();
+
+            emit logUnconcede(eventPlayerId);
 
             break;
         }
--- a/cockatrice/src/game/game_event_handler.h
+++ b/cockatrice/src/game/game_event_handler.h
@@ -122,6 +122,8 @@ signals:
     void logGameClosed();
     void logActivePlayer(Player *activePlayer);
     void logActivePhaseChanged(int activePhase);
+    void logConcede(int playerId);
+    void logUnconcede(int playerId);
 };
 
 #endif // COCKATRICE_GAME_EVENT_HANDLER_H
--- a/cockatrice/src/game/game_state.cpp
+++ b/cockatrice/src/game/game_state.cpp
@@ -15,6 +15,10 @@ GameState::GameState(AbstractGame *parent,
       clients(_clients), gameStateKnown(_gameStateKnown), resuming(_resuming), currentPhase(_currentPhase),
       activePlayer(-1), gameClosed(_gameClosed)
 {
+    gameTimer = new QTimer(this);
+    gameTimer->setInterval(1000);
+    connect(gameTimer, &QTimer::timeout, this, &GameState::incrementGameTime);
+    gameTimer->start();
 }
 
 void GameState::incrementGameTime()
@@ -24,6 +28,7 @@ void GameState::incrementGameTime()
 
 void GameState::setGameTime(int _secondsElapsed)
 {
+    secondsElapsed = _secondsElapsed;
     int seconds = _secondsElapsed;
     int minutes = seconds / 60;
     seconds -= minutes * 60;
@@ -33,12 +38,4 @@ void GameState::setGameTime(int _secondsElapsed)
     emit updateTimeElapsedLabel(QString::number(hours).rightJustified(2, '0') + ":" +
                                 QString::number(minutes).rightJustified(2, '0') + ":" +
                                 QString::number(seconds).rightJustified(2, '0'));
-}
-
-void GameState::startGameTimer()
-{
-    gameTimer = new QTimer(this);
-    gameTimer->setInterval(1000);
-    connect(gameTimer, &QTimer::timeout, this, &GameState::incrementGameTime);
-    gameTimer->start();
 }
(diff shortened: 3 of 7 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -236,26 +236,37 @@ jobs:
             xcode: "14.3.1"
             type: Release
             make_package: 1
+            qt_version: 6.6.*
```
brunoalr: it feels like there should be a way to reduce the duplication of all these params across the matrix's entries, but I have to confess I got a bit lazy to do it in this pr. IMO it's worth it to create a ticket to be done later

.github/workflows/desktop-build.yml
```diff
@@ -271,16 +282,30 @@ jobs:
...
-      - name: Install dependencies using Homebrew
-        shell: bash
-        # CMake cannot find the MySQL connector
-        # Neither of these works: mariadb-connector-c mysql-connector-c++
-        env:
-          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
-        run: |
-          brew update
-          brew install ccache protobuf qt --force-bottle
+      - name: Setup Ccache
```
brunoalr: TODO(this pr): add a comment about why this specific action instead of brew or any other similar action

.github/workflows/desktop-build.yml
```diff
@@ -271,16 +282,30 @@ jobs:
...
+          install-type: "binary"
+
+      - name: Install Qt ${{matrix.qt_version}}
+        uses: jurplel/install-qt-action@v4
+        with:
+          cache: true
+          setup-python: true
+          version: ${{matrix.qt_version}}
+          arch: ${{matrix.qt_arch}}
+          tools: ${{matrix.qt_tools}}
```
brunoalr: TODO(this pr): need to double check if this is actually needed on both mac and windows. I couldn't find any definition of matrix.qt_tools and the IDE (I'm using Cursor) gives this warning "Context access might be invalid: qt_tools"

.github/workflows/desktop-build.yml
brunoalr: done using env vars

ebbit1q:
this is a good solution for getting to build on the old -13 runner again!

however at some point gh will deactivate the -13 runner, leaving us with just the -15-intel runner and us having to drop support for 13 and 14 again.

14 will still be supported for another year, I'd really like to keep support for it around! could you look into using the MACOSX_DEPLOYMENT_TARGET on the new runner to see if vcpkg will create deps for older versions on newer runners?

tooomm:
Vcpkg equivalent (custom triplet) to build specific deps should be this one: https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_osx_deployment_target

Not sure if using the cmake one is already enough and will make vcpkg build dependencies accordingly as well.

brunoalr:
> could you look into using the MACOSX_DEPLOYMENT_TARGET on the new runner to see if vcpkg will create deps for older versions on newer runners?

that's something I have 0 experience with, but I can take a look later :) I propose to first merge this PR so that master doesn't fail 100% of the time on macos-13 and check the MACOSX_DEPLOYMENT_TARGET thingy later. we have a few days left until GH fully removes the macos-13 runners, I guess that should be enough time

brunoalr:
> Vcpkg equivalent (custom triplet) to build specific deps should be this one: [learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_osx_deployment_target](https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_osx_deployment_target)
> 
> Not sure if using the cmake one is already enough and will make vcpkg build dependencies accordingly as well.

awesome, I will start the research from there

ebbit1q:
considering both mac and windows now follow a similar sequence of steps, how difficult would it be to simply add mac to the windows build matrix?

brunoalr:
> considering both mac and windows now follow a similar sequence of steps, how difficult would it be to simply add mac to the windows build matrix?

that was something I wanted to do before for all the OSes actually (as I do in this other project that I maintain: https://github.com/OpenImageDebugger/OpenImageDebugger/blob/main/.github/workflows/build.yml). 

the idea is having a single `job` definition, reusing some of the common `step`s unconditonally for all OSes where possible and using `if` for os/distro-specific steps.

for mac and windows it's probably not too difficult, but it I don't expect it to be just as simple as including one more entry in the matrix, so I would also propose to do it in a separate issue/pr

ebbit1q (approved):
a separate pr is good too! I'll merge this and immediately cut a beta for testers!

brunoalr:
created https://github.com/Cockatrice/Cockatrice/issues/6185

brunoalr:
also created #6186

PR #6172, opened 2025-09-25
[Card DB Models] Move refactor

BruebachL:
A wise man ([Alfred Korzybski](https://en.wikipedia.org/wiki/Alfred_Korzybski)) once said ["The map is not the territory"](https://en.wikipedia.org/wiki/Map%E2%80%93territory_relation). Let's differentiate maps from territory, at least with our folder structure and also not put everything into one class.

This cleans up the folder structure so that:
- The top level contains only logic classes (The Card Database and the manager singleton)
- The responsibilities of the different models are clearly visible (cards, database in general, and tokens)

- Refactor CardDatabaseDisplayModel, TokenDisplayModel and TokenEditModel out of CardDatabaseModel. 
- Move every model into an appropriate folder.

## Screenshots
image


ebbit1q:
not a fan of database/model/database it just seems kind of redundant to make another subfolder in model

BruebachL:
> not a fan of database/model/database it just seems kind of redundant to make another subfolder in model

Sure.

PR #6174, opened 2025-09-25
[Card DB] Various little fixes and cleanups

BruebachL:
Various methods in card_database are not grouped next to their related methods (and don't really flow logically top-to-bottom) and can be enhanced either through various helper classes or various little fixes in how we express logic.

- Simplify some methods
- Introduce some helper methods
- Move some functions within the class

### DIFF
--- a/cockatrice/src/database/card_database.cpp
+++ b/cockatrice/src/database/card_database.cpp
@@ -69,21 +69,21 @@ void CardDatabase::addCard(CardInfoPtr card)
         return;
     }
 
-    // if card already exists just add the new set property
-    if (cards.contains(card->getName())) {
-        CardInfoPtr sameCard = cards[card->getName()];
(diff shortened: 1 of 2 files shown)


ebbit1q:
all the debug builds are failing because of a compile warning

BruebachL:
> all the debug builds are failing because of a compile warning

Rightly so.

ebbit1q:
oh, looking at that, that's definitely a compile warning worth fixing, it could gc collectCustomDatabasePaths's result before using the referred value

BruebachL:
> oh, looking at that, that's definitely a compile warning worth fixing, it could gc collectCustomDatabasePaths's result before using the referred value

Yeah, I have a second branch where I'm refactoring stuff related to database loading where I already went "Hold on, let's local variable this instead."

PR #6175, opened 2025-09-25
[Card DB] Split out database loading and querying from main class

BruebachL:
Card database is 500 lines of code and while it is *technically* responsible for loading itself and answering queries, it could also not be and we would end up with nice modules which are responsible for only a single thing. This change aims to introduce this. Relatively little actual logic is changed. The loading is mostly opaque to the rest of the code anyway, while anything that queries gets access to CardDatabaseManager::query() as a convenience method now.

- Split out loading the database into card_database_loader.cpp
- Split out querying the database into card_database_queries.cpp

### DIFF
--- a/cockatrice/src/database/card_database_queries.cpp
+++ b/cockatrice/src/database/card_database_queries.cpp
@@ -1,0 +1,322 @@
+#include "card_database_queries.h"
+
+#include "../utility/card_set_comparator.h"
+#include "card_database.h"
+
+#include <qrandom.h>
+
+CardDatabaseQueries::CardDatabaseQueries(const CardDatabase *_db) : db(_db)
+{
+}
+
+/**
+ * Looks up the cardInfo corresponding to the cardName.
+ *
+ * @param cardName The card name to look up
+ * @return A CardInfoPtr, or null if not corresponding CardInfo is found.
+ */
+CardInfoPtr CardDatabaseQueries::getCardInfo(const QString &cardName) const
+{
+    return db->cards.value(cardName);
+}
+
+/**
+ * Looks up the cardInfos for a list of card names.
+ *
+ * @param cardNames The card names to look up
+ * @return A List of CardInfoPtr. Any failed lookups will be ignored and dropped from the resulting list
+ */
+QList<CardInfoPtr> CardDatabaseQueries::getCardInfos(const QStringList &cardNames) const
+{
+    QList<CardInfoPtr> cardInfos;
+    for (const QString &cardName : cardNames) {
+        CardInfoPtr ptr = db->cards.value(cardName);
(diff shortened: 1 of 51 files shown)


cockatrice/src/database/card_database_queries.cpp
```diff
@@ -0,0 +1,322 @@
+#include "card_database_queries.h"
+
+#include "../utility/card_set_comparator.h"
+#include "card_database.h"
+
+#include <qrandom.h>
+
+CardDatabaseQueries::CardDatabaseQueries(const CardDatabase *_db) : db(_db)
```
RickyRister: This class isn't a QObject, which means it won't automatically be cleaned up when the CardDatabase is destroyed.
In that case, it might make more sense to store the singleton `CardDatabaseQueries` instance in the `CardDatabaseManager` instead of in the `CardDatabase`, since the `CardDatabaseQueries` isn't really a child of `CardDatabase`.

cockatrice/src/database/card_database_queries.h
```diff
@@ -0,0 +1,50 @@
+#ifndef COCKATRICE_CARD_DATABASE_QUERIES_H
+#define COCKATRICE_CARD_DATABASE_QUERIES_H
+
+#include "../card/exact_card.h"
+#include "../common/card_ref.h"
+
+class CardDatabase;
+class CardDatabaseQueries
```
RickyRister: `Querier` is indeed a real word. Maybe `CardDatabaseQuerier` would be a better name?

RickyRister:
look good otherwise

ebbit1q (approved):
I have fixed the linker errors, thank you for this refactor, after this is merged we can look into dbconverter removing its dependency on the settings!

please address ricky's concern about the cleanup, everything else looks good and functions as expected

ebbit1q:
#6141 would be easier if there were not settings mocks needed anymore for example, so I hope with this we can decouple the loading from the settings

BruebachL:
I will address his comments, thank you *very much* for fixing the linker issues.

ebbit1q:
it was just missing cpp files from the cmakelists in three places

dbconverter/src/main.h
```diff
@@ -9,7 +9,7 @@ class CardDatabaseConverter : public CardDatabase
 public:
     LoadStatus loadCardDatabase(const QString &path)
     {
-        return CardDatabase::loadCardDatabase(path);
+        return loader->loadCardDatabase(path);
```
ebbit1q, line 12: in what namespace is this loader?

BruebachL: It's a variable of ClassDatabase::

PR #6177, opened 2025-09-26
[Game] Fix game timer starting twice, not stopping and not resetting correctly.

BruebachL:
## Related Ticket(s)
- Related to #6127

Currently, we instantiate a new timer when we start it, which seems to lead to the old timer still emitting signals. Double connections lead to game time elapsing twice as fast. Also, setGameTime doesn't properly *set* the gameTime, it just updates the label once.

- Only keep one gameTimer around
- Stop it as well as start it
- Actually set secondsElapsed in setGameTime


ebbit1q (approved):
thanks, this works as expected now!

PR #6178, opened 2025-09-26
[TabGame/PlayerManager] Handle concession properly.

BruebachL:
## Related Ticket(s)
- Fixes an issue in #6127

Concession messages are duplicated because the client doesn't expect the order of the game servers concession handling and assumes that the duplicated state change of the player is the player actively asking for concession.

In the context of this, I would like a review/explanation of how the game server handles concession. The client seems to first receive a gameStateChangedEvent where the player is actually not conceded, followed up by an event where the player *is* conceded?

image

- Do not check in the playerManager slot for player concession if the active local player has conceded. 
- The active local player can only ever actively concede from the tab_game action, which means this action can directly emit the activeLocalPlayerConceded signals of playerManager.

## Screenshots
Problematic concession:
image

Proper concession:
image

Resume after concession:
image

### DIFF
--- a/cockatrice/src/game/player/player_manager.cpp
+++ b/cockatrice/src/game/player/player_manager.cpp
@@ -72,14 +72,6 @@ Player *PlayerManager::getPlayer(int playerId) const
 
 void PlayerManager::onPlayerConceded(int playerId, bool conceded)
 {
-    // GameEventHandler cares about this for sending the concede/unconcede commands
-    if (isLocalPlayer(playerId)) {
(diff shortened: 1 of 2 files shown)


ebbit1q:
has this pr been superseded by #6179 or are there other useful things in this?

BruebachL:
> has this pr been superseded by #6179 or are there other useful things in this?

This is independently useful, it, for example, disables the concede button when the game has stopped in the lobby.

PR #6184, opened 2025-09-26
[HomeTab] Add a button to start a local game.

BruebachL:
No button to start a local game when disconnected which (at least for me) slows down at the very least local development.

- Add a bewton

### DIFF
--- a/cockatrice/src/interface/widgets/general/home_widget.cpp
+++ b/cockatrice/src/interface/widgets/general/home_widget.cpp
@@ -180,11 +180,8 @@ QGroupBox *HomeWidget::createButtons()
     connectButton = new HomeStyledButton(tr("Connect/Play"), gradientColors);
     boxLayout->addWidget(connectButton, 1);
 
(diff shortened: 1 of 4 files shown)


ebbit1q:
how many buttons do we want on this thing tho?

BruebachL:
I think that just about emulates every main tab functionality except for the settings (which I am still in favor of adding somewhere, maybe not as a main list button but in a corner as a little cogwheel somewhere)

BruebachL:
We can also start collapsing and styling these buttons by introducing dedicated ones at some point.

ebbit1q (changes requested):
this exposes some annoying issues with the main menu buttons:
- creating a local game does not disable the connect button in the main menu, this is already present in master rn
- clicking connect and then attempting connection while in a local game destroys the local game without telling the user about it at all
- creating a local game does not disable the local game itself either
- creating two local games causes major bugs, doing game actions in game 1 sends them to game 2 instead, breaking its state
- we should simply disable these actions while in a local game, as does the dropdown menu

PR #6187, opened 2025-09-27
[TabGame/GameEventHandler] Re-emit spectator addition signals in eventGameStateChanged

BruebachL:
If you join a game, most of the setup is done in the initial gameStateChangedEvent, which currently does not emit addition signals for spectators. This makes them not display in the player list widget.

- Re-emit signals.

## Screenshots
image

### DIFF
--- a/cockatrice/src/game/game_event_handler.cpp
+++ b/cockatrice/src/game/game_event_handler.cpp
@@ -251,7 +251,11 @@ void GameEventHandler::eventGameStateChanged(const Event_GameStateChanged &event
         QString playerName = "@" + QString::fromStdString(prop.user_info().name());
         emit addPlayerToAutoCompleteList(playerName);
         if (prop.spectator()) {
-            game->getPlayerManager()->addSpectator(playerId, prop);
+            if (game->getPlayerManager()->getSpectator(playerId).id() == -1) {
+                game->getPlayerManager()->addSpectator(playerId, prop);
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/game_event_handler.cpp
```diff
@@ -251,7 +251,11 @@ void GameEventHandler::eventGameStateChanged(const Event_GameStateChanged &event
         QString playerName = "@" + QString::fromStdString(prop.user_info().name());
         emit addPlayerToAutoCompleteList(playerName);
         if (prop.spectator()) {
-            game->getPlayerManager()->addSpectator(playerId, prop);
+            if (game->getPlayerManager()->getSpectator(playerId).id() == -1) {
```
ebbit1q: this looks strange, shouldn't it simply check if getSpectator returns nullptr, why is it able to access id() and expect it to return -1? is there a default spectator player with value -1?

BruebachL: It's a protobuff, I don't know, the map returns a default ServerInfo_User and the default has id -1, I think?

ebbit1q: can we have like getSpectators().contains(playerId) or something like that so it's a little bit clearer what this is supposed to do?

ebbit1q:
oh, format fails

PR #6188, opened 2025-09-27
[DeckEditor] Properly check if deck is blank.

### LINKED ISSUE #6171: editing the deck name clears deck modified flag
on master, whenever you modify the deck name of a new deck that has not been saved before the deck becomes unmodified and you can't press save anymore.

BruebachL:
## Related Ticket(s)
- Fixes #6171

DeckEditor modified flag isn't being set correctly.

- Add isUnmodified method to decklist, which checks for an empty name, comment and decklist size == 0;
- Use this in the check

## Screenshots
image

image

image

image

image

image

image

image

### DIFF
--- a/common/deck_list.h
+++ b/common/deck_list.h
@@ -243,6 +243,11 @@ public:
     }
     ///@}
 
+    bool isUnmodified() const
+    {
+        return name.isEmpty() && comments.isEmpty() && getCardList().size() == 0;
+    }
+
     /// @name Sideboard plans
     ///@{
     QList<MoveCard_ToZone> getCurrentSideboardPlan();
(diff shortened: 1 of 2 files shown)


common/deck_list.h
```diff
@@ -243,6 +243,11 @@ public slots:
     }
     ///@}
 
+    bool isUnmodified() const
```
ebbit1q: this name is misleading, I could load a decklist and delete all its contents and isUnmodified() would return true

common/deck_list.h
```diff
@@ -243,6 +243,11 @@ public slots:
...
+    {
+        return name.isEmpty() && comments.isEmpty() && getCardList().size() == 0;
```
ebbit1q: what's the difference between .size() == 0 and .isEmpty()?

ebbit1q (approved):
works well now, it's a bit strange that you can set a name, delete the name, and then save what is a completely blank deck, but it doesn't hurt I suppose.

ideally, only decks with contents would be saveable but I can understand that in some edge case someone would want to make a deck with no cards?

PR #6189, opened 2025-09-27
Add workflow to generate doxygen on tag push

BruebachL:
Documentation is a bit lacking and not externally accessible. Doxygen has many benefits.

- Add an action that automatically generates and pushes doxygen documentation to the Cockatrice GitHub Page

## TODO (I do not have access to the github pages repo)
- Add the public key under Settings → Deploy Keys with write access
`ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOEg6xfFmrIWUm8DLaKUPJXMySYeeR8GU/lYAg+lvDld doxygen-deploy`
- Link to the docs from the main page


.github/workflows/generate-and-publish-docs.yml
```diff
@@ -0,0 +1,31 @@
+name: Generate and Deploy Doxygen Docs
+
+on:
+  push:
+    tags:
```
ebbit1q, line 5: would it be useful to also have a `workflow_dispatch:` trigger so we can manually update the docs when we want?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
...
+    tags:
+      - '*' # Only re-generate docs when a new version is pushed
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
ebbit1q: the current version of checkout is actions/checkout@v5

.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          submodules: true
```
ebbit1q: why do we need submodules? do we need vcpkg?

.github/workflows/generate-and-publish-docs.yml
BruebachL: How?

ebbit1q (commented):
can we rename the workflow file to something a little more descriptive? publish-doxygen-docs.yml or generate-docs.yml?

.github/workflows/generate-and-publish-docs.yml
ebbit1q: it doesn't hurt to have a dispatch trigger right?

ebbit1q (approved):
awesome, I've added the deploy key so this should be good to go!

if somebody that knows more about css would like to improve the styling later they're welcome to

.github/workflows/generate-and-publish-docs.yml
ebbit1q: hmm, I merged it but I forgot about the dispatch trigger, it'd be nice if I could try out the workflow manually, do you have any objections to me adding the dispatch trigger?

.github/workflows/generate-and-publish-docs.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Generate Documentation
+        run: doxygen Doxyfile
+
+      - name: Deploy to cockatrice.github.io
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          deploy_key: ${{ secrets.DOCS_DEPLOY_KEY }}
+          external_repository: Cockatrice/cockatrice.github.io
+          publish_branch: main
```
ebbit1q, line 27: just noticed a rather annoying flaw, our main branch on the website repo is actually master

PR #6191, opened 2025-09-28
[Documentation] Doxygen layouting and compilation documentation

BruebachL:
We've not touched the layouting at all and there's no information from the github wiki in the doxygen wiki yet.

Cockatrice changes:
- Give every .md file a @page attribute so doxygen can reference it
- Have our Qt Textbrowser automatically strip this attribute

Doxygen Changes:

- Include project version number, logo, and icon.
- Manually define a layout for our navigation so we can group all the auto-generated code documentation
- Include the GitHub wiki information about compiling (they are markdown as well, so it is simply a matter of adjusting hyperlinks to use @ref and link to the corresponding pages)
- Structure things into User Reference (which includes the search syntax .mds we have laying about) and Developer Reference (which includes information about compilation and contributing)

## Screenshots
Before:
image

After (It is sadly impossible to fold the auto-generated code references into the same group as the other references):
image

image

image


ebbit1q:
I don't think it's a good idea to copy over the wiki page on compiling, this'd mean we now have two locations for the same info, can we simply include them from the source (Cockatrice.wiki is just a simple git repo) or maybe we can just link to the wiki?

PR #6192, opened 2025-09-28
remove building docker images on code changes in prs

ebbit1q:
## Related Ticket(s)
- Fixes #IssueNumber

building the docker images takes about an hour, it's our slowest pr job by far, running it on every code change is meaningless as we already do so with our other actions

- remove code paths from the on pull request trigger from the action

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -11,12 +11,6 @@ on:
       - master
     paths:
       - '.github/workflows/docker-release.yml'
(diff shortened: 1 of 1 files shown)


ebbit1q:
if we don't run this hour long action on every pr it'll free up our parallel runner count, which is limited, allowing all other actions to run faster!

tooomm (approved):
Sure. If it builds in general, it should do in Docker as well.

PR #6194, opened 2025-09-28
CI: Doxygen Docs workflow adjustments

tooomm:
## Related Ticket(s)
- Related #6189

I agree with your comments @ebbit1q and added a manual workflow trigger option and fixed the branch name.

- Naming adjustments to brings the file in line with how the other workflows are labelled
- Changes to the workflow config or doxygen files will trigger a test run in the respective PR
- Deployments only happen for tags

The test run in this PR shows some warnings, FYI @BruebachL, see full logs [here](https://github.com/Cockatrice/Cockatrice/actions/runs/18077052844/job/51435055765?pr=6194).
image

### DIFF
--- a/.github/workflows/documentation-build.yml
+++ b/.github/workflows/documentation-build.yml
@@ -1,4 +1,4 @@
-name: Generate Documentation
+name: Generate Docs
 
 on:
   push:
@@ -7,6 +7,8 @@ on:
   pull_request:
     paths:
       - '.github/workflows/documentation-build.yml'
+      - 'Doxyfile'
+      - 'doxygen_style.css'
   workflow_dispatch:
 
 jobs:


.github/workflows/documentation-build.yml
```diff
@@ -1,12 +1,17 @@
-name: Generate and Deploy Doxygen Docs
+name: Generate Documentation
```
ebbit1q: it doesn't just generate, it does deploy it to cockatrice.github.io as well

tooomm: Yes, that's right.

However, we do not include every main steps in the other flow names either.
It should be brief and give a simple idea, hence we only have "Build Desktop", nothing about test or release etc.
image


Also it will not always deploy (see the new addition that it runs in PR's changing the config, but not deploy). Manual trigger would still deploy.

ebbit1q: I don't think it really matters during prs as you already know you are editing the workflow that does this, I do understand wanting to shorten it though, how about we simply shorten it to Generate Docs?

tooomm: Sure, shortened the label if that's preferred.

For PR's with related doxygen changes however, I think we definetly should at least build the docs and confirm everything is still working as intended and have a simple way to review.

BruebachL:
The warnings are harmless, I assume the doxyfile itself will be tweaked quite a bunch in the upcoming days but thank you for bringing this to my attention. I'm not exactly sure where the mismatch happened because I used doxygen to generate its default config file but oh well, we'll just remove those config parameters if they turn out unused.

Ignoring them is fine for now.

ebbit1q (approved):
thanks

tooomm:
In a follow up PR, I'll look into:
- ~~Cleanup the path configuration to not trigger e.g. Cockatrice builds when working on doxygen only changes~~ #6223
- ~~Include and upload artifacts of the resulting html pages in the workflow context for easy review purposes~~ #6219
- Maybe caching is possible for docs generation as well?
  ```
  type lookup cache used 40488/65536 hits=237665 misses=41134
  symbol lookup cache used 52427/65536 hits=267004 misses=52427
  ```
- ~~Add link to resulting docs when published~~ #6219

ebbit1q:
when the targeted repo has contents it might be able to reuse them, I don't think a separate cache is necessary

tooomm:
Thanks for also running the workflow already manually via the new trigger @ebbit1q.

Deployment works as well, see https://cockatrice.github.io/docs/
The new [./docs folder](https://github.com/Cockatrice/cockatrice.github.io/tree/master/docs) in our webpage repo is quite messy though with a looot of files. Guess that it just how doxygen works. 😃

ebbit1q:
there is no problem with there being lots of files in an automatically generated folder

PR #6195, opened 2025-09-28
[Doxygen] Add javascript to toggle call/caller graph visibility.

BruebachL:
Generating Caller/Call graphs is *incredibly* nice and useful but tends to clutter the documentation a bit when casually browsing.

- Add additional javascript to inject buttons that toggle visibility for call/caller graphs.

## Screenshots
Before (A tame example, most of these graphs span the whole page):
image

(Not so tame)
image

After:
image

image

image

image


ebbit1q (approved):
thanks, I enjoy the file move

PR #6196, opened 2025-09-28
[Doxygen] Layout adjustments

BruebachL:
We've not touched the layouting at all and there's no information from the github wiki in the doxygen wiki yet.

Cockatrice changes:
- Give every .md file a @page attribute so doxygen can reference it
- Have our Qt Textbrowser automatically strip this attribute

Doxygen Changes:

- Include project version number, logo, and icon.
- Manually define a layout for our navigation so we can group all the auto-generated code documentation
- Structure things into User Reference (which includes the search syntax .mds we have laying about) and Developer Reference (which includes information about contributing)

## Screenshots
Before:
image

After (It is sadly impossible to fold the auto-generated code references into the same group as the other references):
image

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-@page contributing Contributing
+<!--! @page contributing Contributing -->
 
 &nbsp; [Introduction](#contributing-to-cockatrice) | [Code Style Guide](
 #code-style-guide) | [Translations](#translations) | [Release Management](


.github/CONTRIBUTING.md
```diff
@@ -1,3 +1,5 @@
+\page contributing Contributing
```
ebbit1q: this line is visible in the github as well, does it have to be the first line or is it possible to put this in a comment as well?

cockatrice/resources/help/deck_search.md
```diff
@@ -1,3 +1,5 @@
+@page deck_search_syntax_help Deck Search Syntax Help
```
ebbit1q, line 1: is this visible in the client?

.github/CONTRIBUTING.md
BruebachL: I think it has to be the first line. I'd like to once again suggest deprecating/folding the github wiki into the doxygen one.

cockatrice/resources/help/deck_search.md
BruebachL: PR description:
"Have our Qt Textbrowser automatically strip this attribute"

cockatrice/src/filters/syntax_help.cpp
```diff
@@ -22,8 +22,11 @@ static QTextBrowser *createBrowser(const QString &helpFile)
     QString text = in.readAll();
     file.close();
 
-    // Poor Markdown Converter
+    // --- Remove @page declarations at the top of the file ---
     auto opts = QRegularExpression::MultilineOption;
+    text = text.replace(QRegularExpression(R"(^\s*@page[^\n]*\n)", opts), "");
```
ebbit1q, line 27: hmm, I see, we remove this line? can we make that clear in the file?

BruebachL: It's just moved down a line.

.github/CONTRIBUTING.md
ebbit1q: this is not part of the github wiki, this is where users are taken by github from the how to contribute page

cockatrice/resources/help/deck_search.md
ebbit1q: I didn't understand what that meant until I read the code that changed, sorry

.github/CONTRIBUTING.md
BruebachL: Ah, mhm, well, Doxygen automatically pulls in any .md, we can exclude it and not have it, I guess but it seems like cool info to have on there.

ebbit1q: I'm trying to find where exactly in github's interface it cares about this file being in .github, it's exactly like the PULL_REQUEST_TEMPLATE, it's used somewhere by github, alternatively we could simply move this file to the extra-pages and put a link to it from here

cockatrice/resources/help/search.md
```diff
@@ -1,3 +1,5 @@
+\page search_syntax_help Search Syntax Help
```
ebbit1q: what's the difference between @ and \\ here? does this line still get removed?

BruebachL: They are functionally equivalent, it's just a matter of style preference. I like @ cause it sort of differentiates it from other commands.

Doxyfile
```diff
@@ -991,7 +991,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = cockatrice common
+INPUT                  = .github/CONTRIBUTING.md cockatrice common doc/doxygen-extra-pages
```
ebbit1q, line 994: can we not include the "magic" .github folder and instead move this file and leave a link or automagically duplicate it in the workflow action?

BruebachL: Turns out, you can fence the comment, which will cause GitHub not to render it but Doxygen to still pick up the declaration.

ebbit1q: excellent

PR #6198, opened 2025-09-30
[Documentation] Sort *every* file into a doxygen group.

BruebachL:
Took 7 hours 9 minutes

You can [group](https://www.doxygen.nl/manual/grouping.html) things together in Doxygen, which will make them show up in the "Topics" section of the documentation, where additional documentation about such groups can be provided.

- Perform some rudimentary grouping for every file.
- Every class that has yet to be properly documented in the style that doxygen requires is tagged with a TODO: Document this. in their ``@brief`` and documented as an ``@file`` instead of an ``@class``.
  - This means that not only do their briefs signal that they are undocumented but they also show up as raw file names instead of class names in the 'Topics' tree, which further highlights that there is work to be done.

## Follow-up Work
We have to get started on actually documenting these classes as well as consider if the grouping I have proposed is entirely accurate or if we may not want to adjust the groups.
Please do not argue about groups in this PR. Let us accept the initial grouping and then move forward from there with follow up PRs that will, hopefully, be much much much smaller than this, which touches literally every file.

## Screenshots

image

To-Do Documentation:
image

Acceptable Documentation:
image

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -991,7 +991,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = .github/CONTRIBUTING.md cockatrice common doc/doxygen-extra-pages
+INPUT                  = .github/CONTRIBUTING.md cockatrice common doc/doxygen-extra-pages doc/doxygen-groups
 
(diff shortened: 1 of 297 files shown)


ebbit1q (approved):
wow, that's a lot of comments! thank you for your hard work!

ebbit1q:
ah, um, it's got conflicts

PR #6205, opened 2025-10-01
[UI] Remove @ from playerName since only the auto complete list cares about it.

BruebachL:
playerName has an @ added to it when most things that care about using this variable are not intended to have this (as a matter of fact, only the auto complete list cares about this)

- Move @ concatenation from variable declaration to function invocation parameter.


ebbit1q (approved):
thanks

PR #6208, opened 2025-10-01
[Documentation] Doxygen regroup files and reorder group structure

BruebachL:
Creating better groups and assigning stuff to these groups instead of multiple ones.

- Make new groups
- Group existing files into these groups


ebbit1q:
would it also be worth moving the files themselves in the folder structure with this reorganising?

BruebachL:
> would it also be worth moving the files themselves in the folder structure with this reorganising?

I'm leading up to it with these PRs, I'm just not quite done organizing all the sub-groups yet.

PR #6210, opened 2025-10-01
add abstract player in expectance of draft players

ebbit1q:
## Related Ticket(s)
- prerequisite for server implementation of #6096
- related #6156

server player does a lot of stuff, a lot of it is not necessary for draft players and as such it has been split off

- create abstract player
- move everything the draft player will also use to abstract player

## Testing
I've tested this pr the same way I tested #6156, creating a local server with three clients one unregistered, one registered, one judge, then going through each menu option for a player (or at least I try to)

### DIFF
--- a/common/server/game/server_abstract_participant.cpp
+++ b/common/server/game/server_abstract_participant.cpp
@@ -202,7 +202,7 @@ Server_AbstractParticipant::cmdJudge(const Command_Judge &cmd, ResponseContainer
         return Response::RespFunctionNotAllowed;
     }
 
-    Server_Player *player = this->game->getPlayer(cmd.target_id());
+    auto *player = this->game->getPlayer(cmd.target_id());
 
     ges.setForcedByJudge(playerId);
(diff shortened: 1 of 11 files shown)


ebbit1q:
there is one functional change in this pr: players are no longer allowed to use Command_DeckSelect while a game is ongoing (this affects less than you'd think, the selected deck just changes the hash, the ongoing game is unaffected)

common/server/game/server_abstract_participant.cpp
```diff
@@ -202,7 +202,7 @@ Server_AbstractParticipant::cmdJudge(const Command_Judge &cmd, ResponseContainer
         return Response::RespFunctionNotAllowed;
     }
 
-    Server_Player *player = this->game->getPlayer(cmd.target_id());
+    auto *player = this->game->getPlayer(cmd.target_id());
```
BruebachL, line 205: lol, lmao.

BruebachL (approved):

PR #6090, opened 2025-08-24
Move game state and event handling out of tab_game and into separate classes

BruebachL:
## Short roundup of the initial problem
tab_game.cpp is responsible for *way* too many things. It is, primarily, a UI widget, which, by definition, should only *display* the state of objects but only manipulate them by sending signals to the appropriate controller classes. This also tightly couples the game state and other logic to the actual UI component which is undesirable for a variety of reasons.

This change aims to introduce a controller class game_event_handler.cpp, which is responsible for forwarding network events to the appropriate QObject (game_meta_info and game_state), which will then emit signals in response to internal changes that slots in the UI component will react to.

## What will change with this Pull Request?
- Introduce game_event_handler.cpp for emitting signals to the correct QObjects in response to network messages
- Introduce game_meta_info.cpp for tracking various lobby attributes and other things not likely to change during gameplay
- Introduce game_state.cpp for tracking attributes related to actual gameplay

PR #6212, opened 2025-10-02
Turn Card, Deck_List, Protocol, RNG, Server, Settings and Utility into libraries and remove cockatrice_common.

BruebachL:
I was refactoring the code and while we can't use C++20 modules just yet (the Qt Moc system does not like them), we can turn some of our code into libraries, which has some benefits, like being able to ditch relative imports across a whole bunch of the codebase.

- Turn Card, Settings and Utility into stand-alone libraries that the rest links against.
- Implement [Open Std: Canonical Project Structure](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html) for these libraries


ebbit1q:
I think we should put the libraries in common, it's the current place for the library we have for the shared parts between the server and client, we could put multiple libraries there, I don't think it makes a lot of sense to put shared stuff in cockatrice

libs/server/src/game/server_abstract_participant.cpp
```diff
@@ -1,15 +1,18 @@
...
-#include "../../get_pb_extension.h"
-#include "../../rng_abstract.h"
-#include "../../trice_limits.h"
-#include "../server.h"
-#include "../server_abstractuserinterface.h"
-#include "../server_database_interface.h"
-#include "../server_room.h"
+#include "../../include/server/game/server_abstract_participant.h"
+
+#include "../../include/server/game/server_arrow.h"
```
ebbit1q: aren't these headers in the same location as this file?

oracle/CMakeLists.txt
```diff
@@ -1,108 +1,78 @@
...
-    ../cockatrice/src/settings/recents_settings.cpp
-    ../cockatrice/src/settings/servers_settings.cpp
-    ../cockatrice/src/settings/settings_manager.cpp
-    ../cockatrice/src/settings/shortcuts_settings.cpp
     ${VERSION_STRING_CPP}
 )
 
 set(oracle_RESOURCES oracle.qrc)
 
-if(UPDATE_TRANSLATIONS)
```
ebbit1q, line 49: where are these moved to?

common/CMakeLists.txt
```diff
@@ -2,48 +2,11 @@
...
-    server/server_abstractuserinterface.cpp
-    server/server_database_interface.cpp
-    server/server_protocolhandler.cpp
-    server/server_remoteuserinterface.cpp
-    server/server_response_containers.cpp
-    server/server_room.cpp
-    serverinfo_user_container.cpp
-    sfmt/SFMT.c
-)
+set(common_SOURCES)
```
ebbit1q: there are no more sources, there is no more library left then? all that is in here is peglib.h which is header only, why not move it to utils? or to another place where we put stuff that's header only like trice_limits.h?

libs/server/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+  src/server_room.cpp
+  src/serverinfo_user_container.cpp
+)
+
+add_dependencies(cockatrice_server cockatrice_protocol_wrapper)
+
+target_include_directories(
+  cockatrice_server
+  PUBLIC include
+  PUBLIC ${CMAKE_SOURCE_DIR}/common
```
ebbit1q: common? that's just peglib right? why does this need peglib? peglib is only used by the searching in the deck editor right?

libs/server/src/game/server_abstract_participant.cpp
BruebachL: Libs are split into `<libname>/src` (for cpp and internals) and `<libname>/include/<libname>` (for headers accessible to other classes), which allows libraries to have purely internal classes and importers/consumers to import library files using a `<libname>/ ` prefix in their imports i.e. import "sever/sever_room.h" which I think should be done for comprehensibility.

ebbit1q: I see they are all in a separate include folder now, I'm not very enthusiastic about this move, but we could at the very least include it right? so we don't need all these ../..?

ebbit1q (commented):
overall I'm excited about this move, even though it's a little intimidating.
this is going to conflict with a lot of stuff I'm working on but it's going to be a net positive for all future changes

ebbit1q:
I don't know why mac ci is failing but this looks interesting:

The INTERPRETER keyword was followed by an empty string or no value at all.
Policy CMP0174 is not set, so cmake_parse_arguments() will unset the
arg_INTERPRETER variable rather than setting it to an empty string.

libs/server/src/game/server_abstract_participant.cpp
BruebachL: I scoured the internet to become a more learned man on this topic and discovered this:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html
which explicitly counsels against src/ and include/ so I am converting the structure to the one proposed in this paper.

ebbit1q: I agree with you, there's people that have already solved this problem following this is a good idea :+1:

BruebachL:
> I'm not the biggest fan of how long this makes all of our includes but adhering to a standard is better than not adhering to a standard

On the other hand, this gets rid of a bunch of relative imports (../../../../ gets confusing)

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
```diff
@@ -1,18 +1,7 @@
...
-#include "pb/event_server_identification.pb.h"
-#include "pb/response_activate.pb.h"
-#include "pb/response_forgotpasswordrequest.pb.h"
-#include "pb/response_login.pb.h"
-#include "pb/response_password_salt.pb.h"
-#include "pb/response_register.pb.h"
-#include "pb/server_message.pb.h"
-#include "pb/session_commands.pb.h"
+#include "../../../../cockatrice/src/main.h"
+#include "../../../../libcockatrice_protocol/libcockatrice/protocol/pending_command.h"
```
ebbit1q: shouldn't this be <libcockatrice_protocol/libcockatrice/protocol/pending_command.h> ?

.github/workflows/translations-push.yml
```diff
@@ -31,10 +31,10 @@ jobs:
       - name: Update Cockatrice translation source
         id: cockatrice
         shell: bash
-        env:
-          FILE: 'cockatrice/cockatrice_en@source.ts'
-          DIRS: 'cockatrice/src common'
-        run: .ci/update_translation_source_strings.sh
+        run: |
+          FILE="cockatrice/cockatrice_en@source.ts"
+          DIRS="cockatrice/src $(find . -maxdepth 1 -type d -name 'libcockatrice_*')"
```
ebbit1q: why not just set these as export so you don't need to add them to the next line?
```suggestion
          export DIRS="cockatrice/src $(find . -maxdepth 1 -type d -name 'libcockatrice_*')"
```

ebbit1q:
thank you for doing the .ts file, I hope it makes it so we don't have to redo the translations (we probably still do but let's see)

tooomm:
This adds quite some new folders to root (and maybe more in the future?), can't we group them all in one "libs" or similar folder?

BruebachL:
> This adds quite some new folders to root (and maybe more in the future?), can't we group them all in one "libs" or similar folder?

Let us adhere strictly to [this document](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html) for now and then see about changes later.

And while I agree that a collected libs folder is nice, this *is* technically correct. Each "project" (i.e. an application or library) deserves its own top level folder since you may distribute it independently.

.github/workflows/translations-push.yml
BruebachL: Oh, good idea!

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
BruebachL: Yes, it should. I used 'Find everywhere' with `/libcockatrice_` but it looks like it didn't catch all of them.

BruebachL:
This PR is generally agreed upon and, at this point, has gotten so large that it deserves a cutoff point and an acceptance of consequences, much like #6090 and #6112

I'm gonna do so now, so that we can discuss remaining issues in separately tracked PRs.

If you are reading this in the future, please check out the next few PR numbers to see related changes (I will try to mention this PR in them as a follow-up).

tooomm:
This did break building of the Servatrice docker image, as well as the Webatrice app.
image

ebbit1q:
they shouldn't have been built in a way that breaks from this, but does anyone use those?

tooomm:
> they shouldn't have been built in a way that breaks from this

Hmm? Wasn't the idea of the common folder to share code, and if you share, changes are affecting all consumers...
For docker, the [Dockerfile it still references and tries to copy the old /common/ folder for relevant Servatrice code](https://github.com/Cockatrice/Cockatrice/blob/master/Dockerfile#L21).
For webatrice, I guess protobuf stuff might be the culprint.

> but does anyone use those?

Well, the automated "official" docker image is a new feature and is not yet published and pullable because no release happened since it was implemented.
It would make hosting and offering public or private servers easier.

Webatrice, I'm not sure what the status is. Are you suggesting to get rid of it?

Anyway, working things should not silently get broken. :)
I think it was just overlooked to update the new locations everywhere.
 --> Still used occurances of [`/common`](https://github.com/search?q=repo%3ACockatrice%2FCockatrice%20%2Fcommon&type=code) in the repo

ebbit1q:
of these occurrence the great majority is webclient code, which has its own common folder which it references, making these false positives, also we have now restarted the concept of the webclient with an entirely new framework so I have no idea if any of it is still useful.

the other hits are in translation files, which are automatically made, I suppose they didn't receive any updates yet?

the only two other things are an outdated shell script, it can be deleted.

lastly we have the dockerfile, which for no reason at all decides to selectively copy directories into the image, if the file actually cared about reducing the resulting image it'd properly clean up the source files and build dependencies after building instead, it could even just use the output from our existing build system which already uses docker images to build!

tooomm:
>of these occurrence the great majority is webclient code, which has its own common folder which it references, making these false positives

So the former /common folder in root was never used by weclient. Thanks for pointing this out. 👍

>also we have now restarted the concept of the webclient with an entirely new framework so I have no idea if any of it is still useful.

Interesting, sounds like that would potentially replace the current code.
Will the new webclient also be hosted in the main Cockatrice repo?

>the other hits are in translation files, which are automatically made, I suppose they didn't receive any updates yet?

I just fixed all language files and translations, they are working again.
Regenerated source strings with new paths, synced to transifex and pulled everything with the new links back again.
There were 3 languages that were not included in the automated setup because they did not met the minimum %-threshold, synced them manually.

> the only two other things are an outdated shell script, it can be deleted.

[Search for `/common` (without webclient matches) has now only 2 matches.](https://github.com/search?q=repo%3ACockatrice%2FCockatrice+%2Fcommon+NOT+path%3A%2F%5Ewebclient%5C%2F%2F&type=code)
You are referring to the script in Servatrice folder?

>lastly we have the dockerfile, which for no reason at all decides to selectively copy directories into the image

Are you saying no code from former /common was ever needed for building a functional Servatrice?
Is any of the new libcockatrice locations relevant for Servatrice? Networking or protobuf for example?

ebbit1q:
> Are you saying no code from former /common was ever needed for building a functional Servatrice? Is any of the new libcockatrice locations relevant for Servatrice? Networking or protobuf for example?

I am saying that selectively copying this folder into the container is a bad idea, the container should instead copy the entire project so the thing doesn't depend on our repo's structure.

in addition the only reason I can imagine it doing this is to reduce the size of the resulting container.

however the resulting container doesn't need the source code, and build files, anymore at all after building, but it doesn't delete them even though this would result in a far better result in the same thing

tooomm:
>I am saying that selectively copying this folder into the container is a bad idea, the container should instead copy the entire project so the thing doesn't depend on our repo's structure.

That would break docker caching which works in layers.
Any change in the repo would then always invalidate the cached copy layer and the complete image (container is the wrong term here) would need to be build from scratch (takes ~1h), even if the change is completely unrelated to servatrice.
Hence the selectively copying of relevant locations only. Usually this high level structure is stable, too. 😉 

There can probably be files deleted in the current image build at the very end to save additional space, good idea. There is also the possibility to do multi-stage builds.

<br>

>of these occurrence the great majority is webclient code, which has its own common folder which it references, making these false positives

The webclient might create it's own common folder and hence the code search brings such results up.

Had another look and as suspected earlier already, webclient was dependent on our /common folder to pull protobuf which is the reason building fails right now. Fixed in #6266. 👍

PR #6214, opened 2025-10-03
[Game] Populate playerLists for menus in their aboutToShow …

BruebachL:
…so they are always current and do not rely on playerMenu manually tracking them. Also add playerActions for previous playerListActions.

Took 1 hour 35 minutes

## Related Ticket(s)
- Fixes players sometimes not showing up in the actions.

I'm not sure why exactly the menus had such trouble keeping the playerLists accurate (the signal emission seemed to be correct.) but that's a terrible way to do these things anyway. Now that we have uncluttered the playerMenu into separate classes, we can simply hook the aboutToShow for the relevant classes up to methods that populate these values dynamically from the Game's PlayerManager. This is much more robust and also allows us to hook signals/slots/acts up declaratively instead of with a sort of opaque switch case.

- Nuke the playerList handling from playerMenu
- Have each menu that cares about a playerList populate its own as a response to aboutToShow()

### DIFF
--- a/cockatrice/src/game/player/menu/grave_menu.cpp
+++ b/cockatrice/src/game/player/menu/grave_menu.cpp
@@ -1,7 +1,12 @@
 #include "grave_menu.h"
 
+#include "../../abstract_game.h"
 #include "../player.h"
 #include "../player_actions.h"
+#include "grave_menu.h"
+
+#include <QAction>
+#include <QMenu>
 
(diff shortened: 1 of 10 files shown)


ebbit1q (approved):
this causes a bit of code duplication but I enjoy seeing the simplicity of this approach where we no longer do all of the if statements in PlayerMenu::playerListActionTriggered()

excellent work

PR #6215, opened 2025-10-03
[TabGame] Fix dangling PlayerMenus in gameMenu

### LINKED ISSUE #6209: players rejoining games duplicate their entry in the game menu
**System Information:**

current master version

Image

**Steps to reproduce:**
 - create a local server
 - start client a, join server and create game
 - start client b, join server and join a's game
 - disconnect client b
 - reconnect client b
 - client a will now have two instances of client b's user in the menu

BruebachL:
## Related Ticket(s)
- Fixes #6209

When we insert the playerMenu into the gameMenu, Qt wraps the playerMenu into a QAction*, which lives
independently and does not get cleaned up when the source menu gets destroyed. We have to manually clean up the QAction*.

- Have playerManager actually delete Players
- Also clean up the gameMenu as described above.

### DIFF
--- a/cockatrice/src/tabs/tab_game.cpp
+++ b/cockatrice/src/tabs/tab_game.cpp
@@ -712,7 +712,7 @@ void TabGame::processPlayerLeave(Player *leavingPlayer)
         QMenu *menu = leavingPlayer->getPlayerMenu()->getPlayerMenu();
         if (menu) {
             // Find and remove the QAction pointing to this menu
-            QList<QAction*> actions = gameMenu->actions();
+            QList<QAction *> actions = gameMenu->actions();
             for (QAction *act : actions) {
                 if (act->menu() == menu) {
                     gameMenu->removeAction(act);


cockatrice/src/tabs/tab_game.cpp
```diff
@@ -707,6 +699,31 @@ void TabGame::addLocalPlayer(Player *newPlayer, int playerId)
...
 }
 
+void TabGame::processPlayerLeave(Player *leavingPlayer)
+{
+    QString playerName = "@" + leavingPlayer->getPlayerInfo()->getName();
+    removePlayerFromAutoCompleteList(playerName);
+
+    scene->removePlayer(leavingPlayer);
+
+    // When we inserted the playerMenu into the gameMenu earlier, Qt wrapped the playerMenu into a QAction*, which lives
```
ebbit1q, line 709: is there a way we can simplify this by adding this as a property somewhere?

BruebachL: There's no reason we can't eventually refactor the gameMenu out of TabGame and then simply have it populate the playerList dynamically like the playerMenu does in #6214. This works for now, imo, ahead of a proper refactor.

ebbit1q: sure, this is a crash so we should prioritize fixing it

PR #6218, opened 2025-10-03
[PrintingSelector] Sync PrintingSelector availability to OverrideAllCardArtWithPersonalPreference setting.

BruebachL:
If a user has OverrideAllCardArtWithPersonalPreference enabled, they can still use the PrintingSelector, which will then populate the individual set cache entries with: One and the same picture OR simply ignore the cached values and load the set preference, which results in one and the same: The user sees only one printing for every set variation.
It seems that sometimes these cache entries gets stuck.

- Clearly warn users about disabling the providerId change and the consequences (and vice-versa)
-  Link PrintingSelector visibility and availability to the setting.
-  clear the networkCache when this setting is toggled to fix stuck images.

## Screenshots

Setting disabled, Printing Selector visible:
image

Toggle setting enabled:
image

Setting enabled, Printing Selector invisible:
image

Toggle setting disabled (goes back to screenshot 1):
image

### DIFF
--- a/cockatrice/src/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/dialogs/dlg_settings.cpp
@@ -681,9 +681,11 @@ void AppearanceSettingsPage::overrideAllCardArtWithPersonalPreferenceToggled(QT_
         PictureLoader::clearNetworkCache(); // PixmapCache is no longer valid.
     } else {
         // If user cancels, revert the checkbox/state back
-        overrideAllCardArtWithPersonalPreferenceCheckBox.blockSignals(true);
-        overrideAllCardArtWithPersonalPreferenceCheckBox.setChecked(!enable);
-        overrideAllCardArtWithPersonalPreferenceCheckBox.blockSignals(false);
+        QTimer::singleShot(0, this, [this, enable]() {
+            overrideAllCardArtWithPersonalPreferenceCheckBox.blockSignals(true);
+            overrideAllCardArtWithPersonalPreferenceCheckBox.setChecked(!enable);
+            overrideAllCardArtWithPersonalPreferenceCheckBox.blockSignals(false);
+        });
     }
 }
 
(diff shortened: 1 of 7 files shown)


ebbit1q (changes requested):
for some reason when I click "no" in the confirmation dialog and then click the checkbox again it does not trigger the dialog and does not change the setting

ebbit1q:
also, the box clearly states that a restart is required, however this pr does immediately enable the printing selector for use in the client, it does not however enable the right click select printing action, I'd like this to be consistent, either have the restart be required and enable the printing selector on restart or make it work well when toggled on runtime (theoretically resetting the image cache should be enough to make this work on runtime, remove the restart required label in this case)

RickyRister:
Instead of straight-up deleting the PrintingSelector window, why not just put a warning message on top of the window? (like the warning we had back when local card images didn't work with the printing selector)

I feel that conditionally nuking windows is going to lead to complications and bugs in the future.

BruebachL:
The features are mutually incompatible as it stands. PictureLoader does not have a flag to indicate whether this is your choice being loaded or an opponents. As such, we can never display all printings if the printing selector is enabled at the same time as the card override settings. We can't display any printings at all. This leads (as we've witnessed) to a lot of user confusion.
I maintain that disabling the printing selector (after explicitly warning the user about adjusting a non-default setting) is the way to go.

RickyRister:
I feel like disappearing the PrintingSelector window is going to lead to even more user confusion about where the printing selector has gone. 

At least with a warning message, the user can figure out for themselves what the problem is.

Also, there is still a use case for editing the printings of a deck while the override setting is enabled: You might personally not use printings on cockatrice, but you still want the printings to be set in the decklist for when you send the decklist elsewhere.

ebbit1q (approved):
functions as expected in my testing, thank you

ebbit1q:
I think the printing selector being gone is the way a user that has enabled this setting would expect it to work, they might also want the printing information to be gone from the decklist and their sets preference changing to not cause duplicate entries of the same card in their decklist but that's a different can of worms

PR #6221, opened 2025-10-03
build: target macos 13 for x86 binaries

brunoalr:
## Related Ticket(s)
- Fixes cockatrice#6186

GH's macos-13 runners, which are x86-based will be deprecated December 4th. The alternative offered is to use macos-15-intel runners. We want to keep supporting older Intel macs that are able to run macos 13. 

- the macos 13 (intel/x86) build will now use the macos-15-intel runner using some flags to target macos 13
- introduce a vcpkg triplet file to build vcpkg dependencies targeting macos 13 
- update xcode of macos 15 builds to 16.4

## Output
```
/Applications/cockatrice.app/Contents/MacOS  vtool -show-build cockatrice
cockatrice:
Load command 11
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 13.0
      sdk 15.5
   ntools 1
     tool LD
  version 1167.5
```

```
/Applications/cockatrice.app/Contents/MacOS  file cockatrice 
cockatrice: Mach-O 64-bit executable x86_64
```

## Todo
- [ ] Manual testing on x86 macos 13
- [x] Manual testing on arm macos 26

### DIFF
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -11,7 +11,7 @@
 # --debug or --release sets the build type ie CMAKE_BUILD_TYPE
 # --ccache [<size>] uses ccache and shows stats, optionally provide size
 # --dir <dir> sets the name of the build dir, default is "build"
-# --x86-macos
+# --x86-macos sets the min os version to 13.0 - only used for Intel builds
 # uses env: BUILDTYPE MAKE_INSTALL MAKE_PACKAGE PACKAGE_TYPE PACKAGE_SUFFIX MAKE_SERVER MAKE_TEST USE_CCACHE CCACHE_SIZE BUILD_DIR CMAKE_GENERATOR
 # (correspond to args: --debug/--release --install --package <package type> --suffix <suffix> --server --test --ccache <ccache_size> --dir <dir>)
 # exitcode: 1 for failure, 3 for invalid arguments
@@ -78,11 +78,7 @@ while [[ $# != 0 ]]; do
       ;;
     '--x86-macos')
       shift
-      if [[ $# == 0 ]]; then
-        echo "::error file=$0::--x86-macos expects an argument"
-        exit 3
-      fi
-      x86_MACOS="$1"
+      X86_MACOS=1
       shift
       ;;
     *)
@@ -126,7 +122,7 @@ fi
 if [[ $PACKAGE_TYPE ]]; then
   flags+=("-DCPACK_GENERATOR=$PACKAGE_TYPE")
 fi
-if [[ $x86_MACOS == true ]]; then
+if [[ $X86_MACOS ]]; then
   flags+=("-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0")
   flags+=("-DVCPKG_TARGET_TRIPLET=x64-osx-13")
   flags+=("-DVCPKG_OVERLAY_TRIPLETS=../cmake/triplets")
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -11,7 +11,7 @@
 # --debug or --release sets the build type ie CMAKE_BUILD_TYPE
 # --ccache [<size>] uses ccache and shows stats, optionally provide size
 # --dir <dir> sets the name of the build dir, default is "build"
-# --x86-macos sets the min os version to 13.0 - only used for Intel builds
+# --target-x86-macos sets the min os version to 13.0 - only used for Intel builds
 # uses env: BUILDTYPE MAKE_INSTALL MAKE_PACKAGE PACKAGE_TYPE PACKAGE_SUFFIX MAKE_SERVER MAKE_TEST USE_CCACHE CCACHE_SIZE BUILD_DIR CMAKE_GENERATOR
 # (correspond to args: --debug/--release --install --package <package type> --suffix <suffix> --server --test --ccache <ccache_size> --dir <dir>)
 # exitcode: 1 for failure, 3 for invalid arguments
@@ -76,9 +76,9 @@ while [[ $# != 0 ]]; do
       BUILD_DIR="$1"
       shift
       ;;
-    '--x86-macos')
+    '--target-x86-macos')
       shift
-      X86_MACOS=1
+      TARGET_X86_MACOS=1
       shift
       ;;
     *)
@@ -122,7 +122,7 @@ fi
 if [[ $PACKAGE_TYPE ]]; then
   flags+=("-DCPACK_GENERATOR=$PACKAGE_TYPE")
 fi
-if [[ $X86_MACOS ]]; then
+if [[ $TARGET_X86_MACOS ]]; then
   flags+=("-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0")
   flags+=("-DVCPKG_TARGET_TRIPLET=x64-osx-13")
   flags+=("-DVCPKG_OVERLAY_TRIPLETS=../cmake/triplets")
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -11,6 +11,7 @@
 # --debug or --release sets the build type ie CMAKE_BUILD_TYPE
 # --ccache [<size>] uses ccache and shows stats, optionally provide size
 # --dir <dir> sets the name of the build dir, default is "build"
+# --x86-macos
 # uses env: BUILDTYPE MAKE_INSTALL MAKE_PACKAGE PACKAGE_TYPE PACKAGE_SUFFIX MAKE_SERVER MAKE_TEST USE_CCACHE CCACHE_SIZE BUILD_DIR CMAKE_GENERATOR
 # (correspond to args: --debug/--release --install --package <package type> --suffix <suffix> --server --test --ccache <ccache_size> --dir <dir>)
 # exitcode: 1 for failure, 3 for invalid arguments
@@ -75,6 +76,15 @@ while [[ $# != 0 ]]; do
       BUILD_DIR="$1"
       shift
       ;;
+    '--x86-macos')
+      shift
+      if [[ $# == 0 ]]; then
+        echo "::error file=$0::--x86-macos expects an argument"
+        exit 3
+      fi
+      x86_MACOS="$1"
+      shift
+      ;;
     *)
       echo "::error file=$0::unrecognized option: $1"
       exit 3
@@ -115,6 +125,11 @@ if [[ $USE_CCACHE ]]; then
 fi
 if [[ $PACKAGE_TYPE ]]; then
   flags+=("-DCPACK_GENERATOR=$PACKAGE_TYPE")
+fi
+if [[ $x86_MACOS == true ]]; then
+  flags+=("-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0")
+  flags+=("-DVCPKG_TARGET_TRIPLET=x64-osx-13")
+  flags+=("-DVCPKG_OVERLAY_TRIPLETS=../cmake/triplets")
 fi
 
 # Add cmake --build flags
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -319,7 +319,7 @@ jobs:
           VCPKG_DISABLE_METRICS: 1
           VCPKG_BINARY_SOURCES: 'clear;files,${{ steps.vcpkg-cache.outputs.path }},readwrite'
           VCPKG_OSX_DEPLOYMENT_TARGET: ${{ matrix.soc == 'Intel' && '13.0' || '' }}
-        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --x86-macos ${{ matrix.soc == 'Intel' }}
+        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" ${{ matrix.soc == 'Intel' && '--x86-macos' || '' }}
 
       - name: Sign app bundle
         if: matrix.make_package && (github.ref == 'refs/heads/master' || needs.configure.outputs.tag != null)
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -232,8 +232,8 @@ jobs:
         include:
           - target: 13
             soc: Intel
-            os: macos-13
-            xcode: "14.3.1"
+            os: macos-15-intel
+            xcode: "16.4"
             type: Release
             make_package: 1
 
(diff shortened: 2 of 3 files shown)


cmake/triplets/x64-osx-13.cmake
```diff
@@ -0,0 +1,12 @@
...
+set(VCPKG_TARGET_ARCHITECTURE x64)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE static)
+
+set(VCPKG_CMAKE_SYSTEM_NAME Darwin)
+set(VCPKG_OSX_ARCHITECTURES x86_64)
+# end of copied code
+
+# Set the minimum macOS version for all dependencies
+set(VCPKG_CMAKE_SYSTEM_VERSION 13.0)
```
brunoalr: probably even older macos versions could be targeted. I guess 11 is the limit set by the qt binaries we get from aqt-install:

```
/Applications/cockatrice.app/Contents/Frameworks/QtCore.framework  vtool -show-build QtCore    
QtCore (architecture x86_64):
Load command 9
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 11.0
      sdk 13.1
   ntools 1
     tool LD
  version 820.1
QtCore (architecture arm64):
Load command 9
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 11.0
      sdk 13.1
   ntools 1
     tool LD
  version 820.1
```

ebbit1q: let's stick to 13 for now, we can try lowering it if people ask for it

.ci/compile.sh
```diff
@@ -11,6 +11,7 @@
 # --debug or --release sets the build type ie CMAKE_BUILD_TYPE
 # --ccache [<size>] uses ccache and shows stats, optionally provide size
 # --dir <dir> sets the name of the build dir, default is "build"
+# --target-x86-macos sets the min os version to 13.0 - only used for Intel builds
```
ebbit1q: I'd like this to be a configurable flag

.ci/compile.sh
```diff
@@ -75,6 +76,11 @@ while [[ $# != 0 ]]; do
       BUILD_DIR="$1"
       shift
       ;;
+    '--target-x86-macos')
+      shift
+      TARGET_X86_MACOS=1
+      shift
```
ebbit1q, line 90: you're using shift twice here without using the second value you're removing!

.ci/compile.sh
ebbit1q: ie --target-macos "13"

.github/workflows/desktop-build.yml
```diff
@@ -232,8 +232,8 @@ jobs:
         include:
           - target: 13
             soc: Intel
-            os: macos-13
-            xcode: "14.3.1"
+            os: macos-15-intel
```
ebbit1q, line 249: have you tried if cross compiling works? we don't need it for the lifetime of this image but this might be relevant in 2 years

.ci/compile.sh
brunoalr: oops

.ci/compile.sh
brunoalr: alrighty

.github/workflows/desktop-build.yml
brunoalr: nope, not yet... I will probably check later what @RickyRister did in his pr before

ebbit1q: :+1:

.ci/compile.sh
ebbit1q: even though you made the CMAKE_OSX_DEPLOYMENT_TARGET get set by this the VCPKG_CMAKE_SYSTEM_VERSION is unchanged!

.ci/compile.sh
```diff
@@ -116,6 +126,21 @@ fi
 if [[ $PACKAGE_TYPE ]]; then
   flags+=("-DCPACK_GENERATOR=$PACKAGE_TYPE")
 fi
+  if [[ $TARGET_MACOS_VERSION ]]; then
+  mkdir -p ../cmake/triplets
+  # Replace dots with hyphens in triplet name for vcpkg compatibility
+  TRIPLET_VERSION=$(echo "${TARGET_MACOS_VERSION}" | sed 's/\./-/g')
+  TRIPLET_FILE="../cmake/triplets/x64-osx-${TRIPLET_VERSION}.cmake"
```
ebbit1q: does the triplet have to be a file? is it not possible to pass these args directly to cmake?

brunoalr: as far as my experiments went, yes... I will check again, but this thing is quite annoying

brunoalr: the flags I am passing directly to cmake make cockatrice compile targeting the desired min os, but not the dependencies.

without the files, the vcpkg-managed dependencies are built using min os == current os (in this case macos 15), then we get a lot of `ld` warnings about trying to link binaries with potentially incompatible libraries. 

this is what I followed:
https://learn.microsoft.com/en-us/vcpkg/concepts/triplets#custom-triplets
https://learn.microsoft.com/en-us/vcpkg/users/triplets#macos-variables
https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/cmake-integration#settings-reference

brunoalr: apparently more people got the same issue: https://github.com/microsoft/vcpkg/issues/39981#issuecomment-2853728239

ebbit1q: yeah, it'd be so nice if it could simply use MACOSX_DEPLOYMENT_TARGET

brunoalr: yep, just confirmed that trying to use only CMAKE_MACOSX_DEPLOYMENT_TARGET and hoping that vcpkg exposes cmake variables (DVCPKG_CMAKE_SYSTEM_VERSION and DVCPKG_OSX_DEPLOYMENT_TARGET) leads to linker warnings about different target OS versions: https://github.com/Cockatrice/Cockatrice/actions/runs/18414138753/job/52473645108?pr=6221#step:6:1262

brunoalr: see this commit: [try passing triplets config as cmake variables](https://github.com/Cockatrice/Cockatrice/pull/6221/commits/77e83e8590c9ae5bbfb7c5973b8b2ce2ef38a964)

.github/workflows/desktop-build.yml
```diff
@@ -331,7 +332,7 @@ jobs:
           CMAKE_GENERATOR: '${{env.CMAKE_GENERATOR}}'
           VCPKG_DISABLE_METRICS: 1
           VCPKG_BINARY_SOURCES: 'clear;files,${{ steps.vcpkg-cache.outputs.path }},readwrite'
-        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --vcpkg
+        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --vcpkg ${{ matrix.soc == 'Intel' && format('--target-macos-version {0}', matrix.target) || '' }}
```
ebbit1q: this looks a bit ugly, can we set this with env TARGET_MACOS_VERSION: like we do with some of the other vars? (we aren't very consistent with this use, I know)

ebbit1q: I guess there actually is some consistency, we currently put every instance of ${{ in the env vars

brunoalr: sure, no problem

ebbit1q: you didn't actually set TARGET_MACOS_VERSION though, you're setting a flag that will get picked up in the script which will then set TARGET_MACOS_VERSION again, I'd prefer we just set the env var directly

ebbit1q: or better, we can just set it in the build matrix, which removes the need for the if statement

brunoalr: > I'd prefer we just set the env var directly

ahh sorry, didn't get it the first time, will change

.ci/compile.sh
ebbit1q: you're still using shift twice here

ebbit1q: wait, I misread, you did fix it, sorry!

.ci/compile.sh
```diff
@@ -11,8 +11,7 @@
 # --debug or --release sets the build type ie CMAKE_BUILD_TYPE
 # --ccache [<size>] uses ccache and shows stats, optionally provide size
 # --dir <dir> sets the name of the build dir, default is "build"
-# --target-macos-version <version> sets the min os version - only used for macOS builds
```
ebbit1q: we um, kinda keep both options for all options so far in the script, I kinda made it to have both options available in case it's useful in testing

ebbit1q:
instead of going back and forth I think I'd like to just edit it myself if you don't mind

brunoalr:
> instead of going back and forth I think I'd like to just edit it myself if you don't mind

makes more sense, yeah. go ahead please

ebbit1q:
so I ended up doing some style changes to the script as well, I also made it work for arm in case we need that, I made it so that the fact that this is our own custom triplet is more clear, I also added the parameter as a new build matrix value so it can be configured for each entry if needed

ebbit1q:
I also noticed something else, while we have always been simply following homebrew releases for qt it was always up to date (6.9 then), however with the vcpkg change we changed to using the same version as windows (6.6, which is eol but we had a problem with later versions), I think we should change the workflow to use a more recent qt version, if it's not possible for windows then for just macos, out of scope for now though

brunoalr:
nice, I see that you simplified the triplet name to just `custom-triplet`. it helped me confirm something that I thought it was happening: vcpkg is also building dependencies for the automatically determined default triplet (`x64-osx` in this case)

https://github.com/Cockatrice/Cockatrice/actions/runs/18420964246/job/52494676360?pr=6221#step:6:90
```
  The following packages will be built and installed:
    * abseil:custom-triplet@20250512.1
    * abseil:x64-osx@20250512.1
      gtest:custom-triplet@1.17.0#2
      liblzma:custom-triplet@5.8.1
      openssl:custom-triplet@3.5.4
    * protobuf:x64-osx@5.29.5#2
      protobuf:custom-triplet@5.29.5#2
      pthreads:custom-triplet@3.0.0#14
    * utf8-range:custom-triplet@5.29.5
    * utf8-range:x64-osx@5.29.5
    * vcpkg-cmake:x64-osx@2024-04-23
    * vcpkg-cmake-config:x64-osx@2024-05-23
    * vcpkg-cmake-get-vars:x64-osx@2025-05-29
      zlib:custom-triplet@1.3.1
```

I was thinking that maybe it is better to [override](https://learn.microsoft.com/en-us/vcpkg/users/examples/overlay-triplets-linux-dynamic#overriding-default-triplets) the default, but then again I don't know if the vcpkg cache action will mess things up without explicitly purging those caches first.

honestly I've been hating vcpkg so far lol

brunoalr:
> I also noticed something else, while we have always been simply following homebrew releases for qt it was always up to date (6.9 then), however with the vcpkg change we changed to using the same version as windows (6.6, which is eol but we had a problem with later versions), I think we should change the workflow to use a more recent qt version, if it's not possible for windows then for just macos, out of scope for now though

on that note, I would suggest specifically 6.9.2 since 6.9.1 does not build on macos 26.0 due to some changes in GL libraries (macos does not ship AGL or something like that anymore). I also tried 6.10.0 locally and it didn't build, but I don't remember what was the actual issue.

.github/workflows/desktop-build.yml
```diff
@@ -245,10 +246,11 @@ jobs:
         include:
           - target: 13
             soc: Intel
-            os: macos-13
-            xcode: "14.3.1"
+            os: macos-15-intel
+            xcode: "16.4"
             type: Release
             make_package: 1
+            override_target: 13
```
brunoalr, line 253: at first glance it seems to be redundant with `target`, am I missing something?

brunoalr: ahh nvm, got it now

ebbit1q: I also thought it a bit redundant but this is the easiest way to set TARGET_MACOS_VERSION without setting it for the other matrix entries

ebbit1q:
homebrew ships qt 6.9.3, we can use that too

ebbit1q:
we can look for alternatives if you hate vcpkg!

ebbit1q:
in the doc page you linked https://learn.microsoft.com/en-us/vcpkg/users/examples/overlay-triplets-linux-dynamic#overriding-default-triplets about overriding the triplet they use --overlay-triplets, is this option not identical to -DVCPKG_OVERLAY_TRIPLETS to cmake?

ebbit1q:
I see now that the double dependency building is affecting our build times like a lot a lot, we should probably fix this!

brunoalr:
> in the doc page you linked https://learn.microsoft.com/en-us/vcpkg/users/examples/overlay-triplets-linux-dynamic#overriding-default-triplets about overriding the triplet they use --overlay-triplets, is this option not identical to -DVCPKG_OVERLAY_TRIPLETS to cmake?

It should be, but then we would have to call the vcpkg binary explicitly from CLI instead of relying on vcpkg-cmake integration. Also, it looks like passing to CLI each individual dependency configured to a given triplet is possible, but that's way too cumbersome IMO as it would lead to some redundancy of having the deps listed in both `vcpkg.json` and in the CLI. BUUUT, if overriding default triplets does not work as expected (likely because of caching challenges) and that's the only way to shave off build times, I guess I can live with it.

ebbit1q:
what I'm saying is that --overlay-triplets is identical to VCPKG_OVERLAY_TRIPLETS so it'd probaly not make any difference, I think it's caused by something else

brunoalr:
> what I'm saying is that --overlay-triplets is identical to VCPKG_OVERLAY_TRIPLETS so it'd probaly not make any difference, I think it's caused by something else

yep :) and I went one step further and was thinking about a possible solution that would look like the example in the link:
```
vcpkg/vcpkg install sqlite3:x64-linux --overlay-triplets=custom-triplets
```

brunoalr:
alright, I found a fix for the duplicate deps: setting both VCPKG_TARGET_TRIPLET and VCPKG_HOST_TRIPLET to the custom triplet.

https://github.com/Cockatrice/Cockatrice/actions/runs/18479465976/job/52651241235?pr=6221#step:6:89
```
  The following packages will be built and installed:
    * abseil:custom-triplet@20250512.1
      gtest:custom-triplet@1.17.0#2
      liblzma:custom-triplet@5.8.1
      openssl:custom-triplet@3.5.4
      protobuf:custom-triplet@5.29.5#2
      pthreads:custom-triplet@3.0.0#14
    * utf8-range:custom-triplet@5.29.5
    * vcpkg-cmake:custom-triplet@2024-04-23
    * vcpkg-cmake-config:custom-triplet@2024-05-23
    * vcpkg-cmake-get-vars:custom-triplet@2025-05-29
      zlib:custom-triplet@1.3.1
```
Docs:
https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/cmake-integration#vcpkg_host_triplet

ebbit1q:
neat, let's see if we can get someone with macos 13 to test this

brunoalr:
today I also managed to cross-compile (!!!) to x86 on arm runners in a separate branch: https://github.com/brunoalr/Cockatrice/pull/29. let me know if you want to go ahead with it instead of using the x86 runner, then I will just cherry-pick the commits into this current branch.

ebbit1q:
that's really cool! though for now using the intel 15 runner will be sufficient, it'll be active for at least another 2 years, unless there is a speed improvement?

brunoalr:
> that's really cool! though for now using the intel 15 runner will be sufficient, it'll be active for at least another 2 years, unless there is a speed improvement?

good question, gotta check it more thoroughly, but at first glance it seemed to be compiled faster both for the initial deps compilation and for cockatrice compilation itself. I'd say let's merge this PR now and I will open another one later

brunoalr:
thanks for finding someone to test it :)

ebbit1q:
you're welcome!

rageoftheday:
Glad to be of help

-RageofAbyss

brunoalr:
let me know if I need to change anything else

ebbit1q:
no, I think this is good, I was just leaving some time for others to comment on this pr if they wanted (considering the current workflow will continue to work for a couple of months there is no rush)

if you'd like, I'd appreciate your opinion on #6226 though

brunoalr:
awesome! I will take a look at that PR

PR #6223, opened 2025-10-04
CI: Update trigger paths for desktop workflows

tooomm:
Desktop builds & linting checks were run for unrelated changes.
Exclude path used for workflow triggers.

- Switch to more general include paths
- Update listed folders

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -10,7 +10,7 @@ on:
     branches:
       - master
     paths:
-      - '**'
+      - '*/**'    # matches all files not in root
       - '!**.md'
       - '!.github/**'
       - '!.husky/**'
@@ -18,11 +18,13 @@ on:
       - '!doc/**'
       - '!webclient/**'
       - '.github/workflows/desktop-build.yml'
+      - 'CMakeLists.txt'
+      - 'vcpkg.json'
     tags:
       - '*'
   pull_request:
     paths:
-      - '**'
+      - '*/**'    # matches all files not in root
       - '!**.md'
       - '!.github/**'
       - '!.husky/**'
@@ -30,6 +32,8 @@ on:
       - '!doc/**'
       - '!webclient/**'
       - '.github/workflows/desktop-build.yml'
+      - 'CMakeLists.txt'
+      - 'vcpkg.json'
 
 # Cancel earlier, unfinished runs of this workflow on the same branch (unless on master)
 concurrency:
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -11,18 +11,23 @@ on:
       - master
     paths-ignore:
       - '**.md'
+      - 'doc/**'
       - 'webclient/**'
-      - '.github/workflows/web-*.yml'
+      - '.github/workflows/docker-*.yml'
+      - '.github/workflows/documentation-*.yml'
       - '.github/workflows/translations-*.yml'
-      - '.github/workflows/docker-release.yml'
+      - '.github/workflows/web-*.yml'
     tags:
       - '*'
   pull_request:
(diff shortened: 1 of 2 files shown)


ebbit1q:
if the others use include paths why not have this workflow also use include paths?

tooomm:
>if the others use include paths why not have this workflow also use include paths?

The other workflow files are very specific in scope and mostly done with 1-2 include paths.

We would need to mostly include all repo as relevant files for desktop builds (our main workflow) are all over.
`path-ignore` is easier here.

ebbit1q:
they wouldn't be much more complicated than the current exclude list, something like this would work:
```
*/src/**
*/resources/**
common/**
tests/**
cmake/**
CMakeLists.txt
vcpkg.json
vcpkg/**
.github/workflows/desktop-build.yml
.ci/**
```

we'd still have to exclude webclient/** but not the others

tooomm:
What about the scripts and Dockerfiles in `.ci`? Or in root?
What about `tests`?
We also don't only want to build on `src` changes, but also other resources, translation or cmake files. Hence I don't think we can abstract cockatrice/servatrice/oracle/dbconverter folders like that?

So not sure if it's actually better/helpful.

ebbit1q:
I've updated my list, there is no use to rebuild for changes to resources or translations, compilations don't fail when they change

.github/workflows/desktop-build.yml
```diff
@@ -11,18 +11,23 @@ on:
       - master
     paths-ignore:
       - '**.md'
+      - 'doc/**'
       - 'webclient/**'
-      - '.github/workflows/web-*.yml'
```
ebbit1q, line 15: we could just replace this with .github/* !.github/actions/desktop-build.yml using ! inverting matches

tooomm: Unfortunately, that's not how this works. Check the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-and-excluding-paths).

The exclusion of paths like that only works in the "include mode" of `paths`, not with `paths-ignore`.
We use that in some of the other files workflow triggers by the way for an efficient and simple pattern already.

ebbit1q: in that case we'd have to use paths ** and exclude everything with ! individually I suppose, it'd be cleaner than individually excluding every workflow file we add

tooomm: Hmm, something like this should actually work!
```
    paths:
      - '**'
      - !'**.md'
      - !'doc/**'
      - !'webclient/**'
      - !'.github/**'
      - '.github/workflows/desktop-*.yml'
```

I do agree that it's more general and needs less updates when e.g. new workflows get added.
Will update the PR.

.github/workflows/desktop-build.yml
```diff
@@ -9,20 +9,27 @@ on:
...
     branches:
       - master
-    paths-ignore:
-      - '**.md'
-      - 'webclient/**'
-      - '.github/workflows/web-*.yml'
-      - '.github/workflows/translations-*.yml'
-      - '.github/workflows/docker-release.yml'
+    paths:
+      - '**'
```
ebbit1q: there's a lot of loose files in the base dir, I don't think we want to ever include those with the exception of the CMakelists.txt, we could set this to `*/**` and exclude all the folders or explicitly set it to only include servatrice/** common/** dbconverter/** oracle/** cockatrice/** tests/** cmake/** .ci/**

tooomm: I like the `*/**` way. It's more general which might be helpful for the core workflow.
Also included the vcpkg file for build, and adjusted for lint.

ebbit1q (approved):
thanks :+1:

PR #6225, opened 2025-10-05
Don't use vcpkg on local macOS

RickyRister:
## Related Ticket(s)
- Fixes issue introduced by #6170

With the changes to use vcpkg for mac in CI, mac in local also uses vcpkg by default now.
We would prefer mac in local to still use homebrew by default.

- Add `USE_VCPKG` option to cmake
- Use vcpkg only if OS is windows OR `USE_VCPKG` is enabled
  - That way mac still uses homebrew by default
- Update CI to run mac with `USE_VCPKG` enabled

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -318,7 +318,7 @@ jobs:
           CMAKE_GENERATOR: '${{env.CMAKE_GENERATOR}}'
           VCPKG_DISABLE_METRICS: 1
           VCPKG_BINARY_SOURCES: 'clear;files,${{ steps.vcpkg-cache.outputs.path }},readwrite'
-        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --use-vcpkg
+        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --vcpkg
 
       - name: Sign app bundle
         if: matrix.make_package && (github.ref == 'refs/heads/master' || needs.configure.outputs.tag != null)
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -318,7 +318,7 @@ jobs:
           CMAKE_GENERATOR: '${{env.CMAKE_GENERATOR}}'
           VCPKG_DISABLE_METRICS: 1
           VCPKG_BINARY_SOURCES: 'clear;files,${{ steps.vcpkg-cache.outputs.path }},readwrite'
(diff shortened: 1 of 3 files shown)


ebbit1q (approved):
thank you for fixing this!

.github/workflows/desktop-build.yml
```diff
@@ -318,7 +318,7 @@ jobs:
           CMAKE_GENERATOR: '${{env.CMAKE_GENERATOR}}'
           VCPKG_DISABLE_METRICS: 1
           VCPKG_BINARY_SOURCES: 'clear;files,${{ steps.vcpkg-cache.outputs.path }},readwrite'
-        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE"
+        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --use-vcpkg
```
ebbit1q:
```suggestion
        run: .ci/compile.sh --server --test --ccache "$CCACHE_SIZE" --vcpkg
```

ebbit1q:
@brunoalr would you be able to add optional instructions to use vcpkg to our wiki? https://github.com/Cockatrice/Cockatrice/wiki/Compiling-Cockatrice-(macOS)

brunoalr:
> @brunoalr would you be able to add optional instructions to use vcpkg to our wiki? [Wiki: Compiling Cockatrice (macOS)](https://github.com/Cockatrice/Cockatrice/wiki/Compiling-Cockatrice-(macOS))

yep :)

ebbit1q:
@brunoalr thanks!

I noticed you used pip in the instructions to install aqtinstall, is pip available by default on macos?

brunoalr:
> @brunoalr thanks!
> 
> I noticed you used pip in the instructions to install aqtinstall, is pip available by default on macos?

good question, probably it isn't 😅 I will update it. also there is one extra step needed to deploy locally: using a adhoc signature (basically setting an env var to "-") before creating the package - I will update it there

brunoalr:
> > @brunoalr thanks!
> > I noticed you used pip in the instructions to install aqtinstall, is pip available by default on macos?
> 
> good question, probably it isn't 😅 I will update it. also there is one extra step needed to deploy locally: using a adhoc signature (basically setting an env var to "-") before creating the package - I will update it there

oof, now I can't find any official answer about pip being available by default or not and I can't remember how exactly I installed it on my machine

ebbit1q:
I think it's very unlikely to be installed by default

Issue #6185, opened 2025-09-26
Unify vcpkg-based builds (macos and windows) on CI


brunoalr:
**Description of New Feature**
Build steps of windows and macos jobs are currently very similar. We can try and consolidate them into a single "vcpkg-based builds" job with hopefully minor OS-dependent logic.

**Additional Context**
https://github.com/Cockatrice/Cockatrice/pull/6170#issuecomment-3339923745

PR #6226, opened 2025-10-05
move debug build to use homebrew to be representative of a developer machine

ebbit1q:
## Related Ticket(s)
- #6225

we forgot that most developers on macos might want to continue using homebrew as it's convenient and provides built binaries, it'd be nice to have our debug test dedicated to running on this scenario to make sure that developers remain able to build

- add homebrew build job

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -251,12 +251,6 @@ jobs:
             type: Release
             make_package: 1
 
-          - target: 15
-            soc: Apple
-            os: macos-15
-            xcode: "16.2"
-            type: Debug
-
     name: macOS ${{matrix.target}}${{ matrix.soc == 'Intel' && ' Intel' || '' }}${{ matrix.type == 'Debug' && ' Debug' || '' }}
     needs: configure
     runs-on: ${{matrix.os}}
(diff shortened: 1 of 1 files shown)


tooomm:
Hmm, why do we want to do this in CI?
Seems like quite some overhead and am not sure this is actually needed or too helpful?

Homebrew had only issues on old macOS versions running out of support so far.

ebbit1q:
because developers use this, our recent changes broke the ability for developers to compile using homebrew, which is bad...

brunoalr:
code looks good to me.

**tl;dr**: maybe we should go back to homebrew for most macos versions and use vcpkg for x86 macos builds only.

-----

just to be clear, my main goal is to make devs' life easier, be it by having faster builds (local and CI) or by making things more convenient

on whether we should implement it or not, here are my thoughts: usually I'm more in favor of having a single flow than multiple alternatives to make it easier to maintain, but that's not always possible or convenient. 
1) the biggest pro of vcpkg IMO is adding consistency of dependency versions and build process at least between our windows and macos builds
2) the biggest con is that it does not feel "native" to macos devs.

OTOH, `1)` would make more sense if it was possible to be consistent across all supported OSes, but packaging with "native" Linux distro managers wouldn't be possible (food for thought: perhaps distributing flatpaks would be an idea to allow it?) and `2)`: even though it is not familiar, it does not prevent devs from building locally after some studying.

with that said, an quoting @tooomm `Homebrew had only issues on old macOS versions running out of support so far.`, maybe what would make more sense now would be to go back to homebrew as the main flow for arm64/newer macos and keep vcpkg for x86/older macos since we don't use vcpkg for all OSes anyways?

ebbit1q:
I think it's fine to have all of our packages built by the same workflow (perhaps we can consolidate windows and mac vcpkg at some point?

we just kinda forgot people also compile on macos for development reasons, with this workflow no packages are made, it just is a check to make sure it runs

brunoalr:
Ahh, got it. To me it is also fine and it shouldn't hurt, since github still has free runners :)

brunoalr:
> (perhaps we can consolidate windows and mac vcpkg at some point?

yep, I opened #6185 some time ago and started working on it on my fork today, will open a PR when it's ready :D

PR #6227, opened 2025-10-05
CI: More workflow paths

tooomm:
- Update paths for all workflows
- Alphabetical file ordering in formatting script

### DIFF
--- a/.github/workflows/desktop-lint.yml
+++ b/.github/workflows/desktop-lint.yml
@@ -1,14 +1,30 @@
 name: Code Style (C++) 
 
 on:
+  push:
+    branches:
+      - master
+    paths:
+      - '**'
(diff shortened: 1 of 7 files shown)


.github/workflows/desktop-lint.yml
```diff
@@ -1,18 +1,35 @@
 name: Code Style (C++) 
 
 on:
-  pull_request:
+  push:
```
ebbit1q: we don't really care about linting merges

tooomm: I added push to catch direct pushes to master, not merges.
But we have branch protection enabled and require PR's. Not needed indeed.

ebbit1q (approved):
sure, thanks

PR #6229, opened 2025-10-07
Turn Card, Deck_List, Protocol, RNG, Server, Settings and Utility into libraries and remove cockatrice_common.

BruebachL:
Took 6 hours 3 minutes

This is a squashed and adjusted branch after a merge conflict, please see #6212 for prior discussions of this PR.

I was refactoring the code and while we can't use C++20 modules just yet (the Qt Moc system does not like them), we can turn some of our code into libraries, which has some benefits, like being able to ditch relative imports across a whole bunch of the codebase.

- Turn Card, Deck_List, Protocol, RNG, Server, Settings and Utility into stand-alone libraries that the rest links against.
- Implement [Open Std: Canonical Project Structure](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html) for these libraries


libcockatrice_server/libcockatrice/server/game/server_player.cpp
```diff
@@ -1,45 +1,70 @@
...
+#include <libcockatrice/protocol/pb/response.pb.h>
+#include <libcockatrice/protocol/pb/response_deck_download.pb.h>
+#include <libcockatrice/protocol/pb/response_dump_zone.pb.h>
+#include <libcockatrice/protocol/pb/serverinfo_player.pb.h>
+#include <libcockatrice/protocol/pb/serverinfo_user.pb.h>
+#include <libcockatrice/rng/rng_abstract.h>
+#include <libcockatrice/utility/color.h>
+#include <libcockatrice/utility/trice_limits.h>
+
+struct MoveCardStruct
```
ebbit1q, line 50: I thought I moved this to a separate file?

webclient/src/components/CardDetails/CardDetails.tsx
```diff
@@ -1,9 +1,9 @@
 // eslint-disable-next-line
-import React, { useMemo, useState } from 'react';
```
ebbit1q, line 2: why is there webclient changes in this?

libcockatrice_server/libcockatrice/server/game/server_player.cpp
BruebachL: It looks like I didn't rebase correctly after all.

webclient/src/components/CardDetails/CardDetails.tsx
BruebachL: I think the IDE just hit this automatically during a refactor, I'm not sure.

libcockatrice_server/libcockatrice/server/game/server_player.cpp
ebbit1q: sorry about the merge conflicts!

ebbit1q: question though, why not push to the existing branch/pr? I don't think there's a need to make a new pr for a rebase right?

ebbit1q:
I've tried to find if there is any external source causing the mac builds to fail, to make sure the issue isn't avoided by caching I've asked zach to clear the cache and did a new run and the build progresses normally, without any of the INTERPRETER keyword warnings: https://github.com/Cockatrice/Cockatrice/actions/runs/18329773066/job/52202404170

clearly there is something in this pr that causes this warning and fails the vcpkg dependency build?

PR #6230, opened 2025-10-07
update vcpkg

ebbit1q:
## Related Ticket(s)
-  #6229

for some reason #6229 has issues compiling the vcpkg submodules for macos only, to investigate I am updating vcpkg outside that pr to see if it causes the problem

- update submodule


ebbit1q:
well, I guess that wasn't it

PR #6236, opened 2025-10-09
[Move refactor] Reparent orphan classes

BruebachL:
See #6234 and #6235 for related changes. I recommend merging this one first.

There have been some efforts to restructure the codebase in the last year. These have been iterative and so far, we have a bunch of classes which are either in the wrong location or have no clear home as of yet.

Furthermore, it is my belief that we should closely adhere to the actual object inheritance when structuring the classes and files related to the Qt side of things. This leads to an intuitive and reasonably structured layout. It is still possible to sub-group widgets underneath their object-parent folder and this can happen for both generic QWidget types (i.e. QDialogs is a subgroup of QWidgets) as well as functional groups (i.e. VisualDeckEditor is a subgroup of QWidgets since there are direct subclasses of QWidgets which concern themselves with visual deck editing.)

To access a class you are interested in (i.e. a menu), you can then easily navigate down the object structure
interface -> widgets -> menus
and then browse the menus available to you to find the one you are interested in.

## What will change?
- A bunch of classes with no clear home will be stored in new folders created for them
- A bunch of classes which were in the wrong location are now either grouped into their correct semantic location or in their appropriate library.

## Screenshots
Source tree before:
image

Source tree after:
image

Source tree with #6234 and #6235 applied as well:

image

image

image

### DIFF
--- a/libcockatrice_deck_list/CMakeLists.txt
+++ b/libcockatrice_deck_list/CMakeLists.txt
@@ -38,5 +38,5 @@ target_include_directories(libcockatrice_deck_list PUBLIC ${CMAKE_CURRENT_SOURCE
 
 target_link_libraries(
   libcockatrice_deck_list PUBLIC libcockatrice_protocol libcockatrice_card libcockatrice_utility
-  ${COCKATRICE_QT_MODULES}
+                                 ${COCKATRICE_QT_MODULES}
 )


RickyRister:
Do we want a `widget/server` folder? I find the grouping within the `server` folder feels a lot more flimsy

BruebachL:
> Do we want a `widget/server` folder? I find the grouping within the `server` folder feels a lot more flimsy

Yeah, you're right, I intend to split this up in another PR. I'm trying to keep these one-at-a-time.

libcockatrice_deck_list/CMakeLists.txt
```diff
@@ -16,15 +21,22 @@ endif()
...
 )
 
 add_dependencies(libcockatrice_deck_list libcockatrice_protocol)
 
 target_include_directories(libcockatrice_deck_list PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
 target_link_libraries(
-  libcockatrice_deck_list PUBLIC libcockatrice_protocol libcockatrice_utility ${COCKATRICE_QT_MODULES}
+  libcockatrice_deck_list PUBLIC libcockatrice_protocol libcockatrice_card libcockatrice_utility
+                                 ${COCKATRICE_QT_MODULES}
```
ebbit1q, line 41: the linter did this?

PR #6240, opened 2025-10-09
update format.sh

ebbit1q:
add shellcheck to format.sh
add statement macros to .clang-format
add no clang format to format.sh
add changed file list to format.sh diff
rename --cf-version to --print-version in format.sh lint files

### DIFF
--- a/.ci/lint_cpp.sh
+++ b/.ci/lint_cpp.sh
@@ -11,7 +11,7 @@ if ! git merge-base origin/master HEAD; then
 fi
 
 # Check formatting using format.sh
-echo "Checking your code using clang-format/cmake-format..."
+echo "Checking your code using format.sh..."
 
 diff="$(./format.sh --diff --cmake --shell --print-version --branch origin/master)"
 err=$?


ebbit1q:
also updates the runner we use to ubuntu-latest

ebbit1q:
finding why the fuck it did this wasted me so much time....

I just removed them, it's better now

.ci/lint_cpp.sh
```diff
@@ -13,9 +13,17 @@ fi
 # Check formatting using format.sh
 echo "Checking your code using clang-format/cmake-format..."
```
tooomm:
```suggestion
echo "Checking your code using clang-format, cmake-format and shellcheck..."
```

ebbit1q: well it doesn't always do so, it only invokes the tools it needs so it rarely is "and"

ebbit1q:
--clang would be confusing as an option, the tool is clang format, clang is a compiler on its own, the default should always be to perform clang-format as in 99% of the cases this is what you want, changing it to be off by default would cause me great annoyance while using this script as it'd at least quadruple the amount of keypresses I'd have to do to use it

--print-version is fine, by default it'll only print one version, that of clang-format, it simply also does print version when using the other tools

.ci/lint_cpp.sh
ebbit1q:
```suggestion
echo "Checking your code using format.sh..."
```

PR #6243, opened 2025-10-13
Servatrice build failure

ebbit1q:
## Related Ticket(s)
- Regression from #6212

#6212 made servatrice depend on libraries that depend on qt components that it doesn't need and does not pull in by default, like widgets and gui

to build servatrice without client use: `cmake .. -DWITH_SERVER=1 -DWITH_CLIENT=0 -DWITH_ORACLE=0 -DWITH_DBCONVERTER=0`

- multiple attempts to do something to fix it
- adds a test that shows the problem

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -115,7 +115,7 @@ jobs:
             version: 11
             package: DEB
 
-          - distro: Debian
+          - distro: Servatrice_Debian
             version: 11
             package: DEB
             test: skip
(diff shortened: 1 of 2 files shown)


.github/workflows/desktop-build.yml
tooomm:
```suggestion
    name: ${{matrix.distro}} ${{matrix.version}}${{matrix.server_only == 'yes' && ' (Servatrice only)' || '' }}
```

.github/workflows/desktop-build.yml
```diff
@@ -113,7 +113,12 @@ jobs:
           - distro: Debian
             version: 11
             package: DEB
-            test: skip    # Running tests on all distros is superfluous
+
```
tooomm:
```suggestion
          - distro: Debian
            version: 11
            package: DEB
            test: skip    # Running tests on all distros is superfluous

```

PR #6263, opened 2025-11-01
ci: unify vcpkg jobs (take 2)

### LINKED ISSUE #6185: Unify vcpkg-based builds (macos and windows) on CI

brunoalr:
## Related Ticket(s)
- Fixes #6185

There a lot of similarities in the macos and windows CI jobs specially now that both use vcpkg as their dependency manager, we can reuse some code.

- unify `build-windows` and `build-macos` job into `build-vcpkg`

## Remaining work
- [ ] Test release-related steps (signing, notarizing, binary attestation, release creation)


brunoalr:
@tooomm do you wanna take a look at this pr?

.github/workflows/desktop-build.yml
```diff
@@ -239,105 +239,168 @@ jobs:
...
+
+          - os: macOS
+            target: 15
+            runner: macos-15
             soc: Apple
-            os: macos-15
             xcode: "16.4"
             type: Debug
+            package_suffix: "-macOS15_Debug"
+            artifact_name: macOS15_Debug-package
```
ebbit1q: the debug build doesn't make any packages

.github/workflows/desktop-build.yml
```diff
@@ -239,105 +239,168 @@ jobs:
...
+        if: matrix.os == 'Windows'
+        id: add-msbuild
+        uses: microsoft/setup-msbuild@v2
+        with:
+          msbuild-architecture: x64
+
       # Using jianmingyong/ccache-action to setup ccache without using brew
       # It tries to download a binary of ccache from GitHub Release and falls back to building from source if it fails
       - name: Setup ccache
+        if: matrix.os == 'macOS'
```
ebbit1q: this should instead check for use_ccache I think

ebbit1q:
:+1: seems all good, not sure exactly how we'd test attestation etc. I have no idea what they actually do

ebbit1q:
I wouldn't mind merging this and finding out when we cut a beta right after

.github/workflows/desktop-build.yml
brunoalr: good catch, will remove

.github/workflows/desktop-build.yml
brunoalr: makes sense

brunoalr:
> I have no idea what they actually do

signing is macos stuff (I don't know much about it except that is to tell the OS who is the author of the binary in very general lines).

[attestations](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations) are a github thing to ensure that the binaries from the releases are legit

ebbit1q (approved):
I think this looks good to go, I think @tooomm would agree, if there is a need for changes we can put them into a followup pr after I cut a beta

ebbit1q:
I'll make a beta tomorrow

ebbit1q:
it looks like I wont be able to make a beta because of https://github.com/Cockatrice/Cockatrice/issues/6276

ebbit1q:
I'm just going to merge this and worry about betas later, the thing suddenly breaking because of something expiring really sucks! I have no idea why this happens!

brunoalr:
alright! I have no concrete idea either... my *guess* is that macos certificates have expiration time... don't know who set it up, tho... maybe @ZeldaZach?

PR #6266, opened 2025-11-02
Fix broken webclient

tooomm:
## Related Ticket(s)
- Related #6212

`/common` folder got restructured into several libraries

- Link to protobuf correctly in new structure


ebbit1q:
the webclient is being rebuilt entirely to my understanding, @seavor can tell us if updating the current webclient is worth doing.

tooomm:
I understood that there are works towards a completely new webclient for Cockatrice and it would replace the existing first try. (Potentially driven by seavor again?)
Not aware of any timeline or progress here.

Anyways, there is no harm in fixing the existing webclient and allow it to build and get a green CI status for it.

PR #6267, opened 2025-11-02
Cleanup: Delete servatrice/scripts/mk_pypb.sh

tooomm:
See https://github.com/Cockatrice/Cockatrice/pull/6212#issuecomment-3447840864


ebbit1q (approved):

ebbit1q:
with the deletion of this script I'd like to point anyone that cared about it to my project that encompasses its full function and aims to provide a complete interface: https://github.com/ebbit1q/crow

PR #6268, opened 2025-11-02
Update Dockerfile

tooomm:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,10 +29,9 @@ RUN mkdir build && cd build && \
 FROM ubuntu:24.04
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-  libmariadb-dev-compat \
-  libprotobuf-dev \
+  libprotobuf32t64 \
   libqt6sql6-mysql \
-  qt6-websockets-dev \
+  libqt6websockets6 \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 


Dockerfile
```diff
@@ -16,16 +16,22 @@ RUN apt-get update && apt-get install -y\
   qt6-tools-dev \
   qt6-tools-dev-tools
 
-COPY ./CMakeLists.txt ./LICENSE ./README.md /home/servatrice/code/
-COPY ./cmake /home/servatrice/code/cmake
-COPY ./common /home/servatrice/code/common
+# Order from least changing at the top to more frequently changing at the bottom
+COPY ./libcockatrice_protocol /home/servatrice/code/libcockatrice_protocol
```
ebbit1q: please copy the entire project simply, copying individual files is fragile and unnecessary, if saving space is a concern then the entire thing can be deleted after building

BruebachL: I've turned it into a multi-stage file so the build stage can be discarded (I think), if space is truly a concern.

ebbit1q:
```
/src/build/libcockatrice_models/libcockatrice/models/database/../../../../../libcockatrice_models/libcockatrice/models/database/card_database_display_model.h:11:10: fatal error: filter_string.h: No such file or directory
```

why does it still try to compile the display model?

BruebachL:
> ```
> /src/build/libcockatrice_models/libcockatrice/models/database/../../../../../libcockatrice_models/libcockatrice/models/database/card_database_display_model.h:11:10: fatal error: filter_string.h: No such file or directory
> ```
> 
> why does it still try to compile the display model?

Filter string isn't librarized yet in this PR's base. This will compile with #6293 and #6294 as a base.

ebbit1q:
but why does it need the filter string library at all? servatrice doesn't use it afaik

BruebachL:
> but why does it need the filter string library at all? servatrice doesn't use it afaik

I honestly forget how it's pulled in

BruebachL:
> but why does it need the filter string library at all? servatrice doesn't use it afaik

I've rebased and as suspected https://github.com/Cockatrice/Cockatrice/pull/6293 was the issue. Deck_list and database_models were the same library so linking against the deck_list model pulls in database_models and thus filter string.

After the rebase, we're now stuck on deck_list_models dependency on deck_loader (see https://github.com/Cockatrice/Cockatrice/pull/6294 which enables the docker container to be built cleanly after rebasing this PR on it)

```
#16 242.2 /src/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp:5:10: fatal error: QBrush: No such file or directory
#16 242.2     5 | #include <QBrush>
```

Dockerfile
ebbit1q: tom talked about how doing it this way caches the build image, meaning it doesn't have to recompile when something changes outside of these directories, in my opinion we should just compile a servatrice binary that the dockerfile consumes, as we do compile the entire project already for every push to a pr, and compiling those takes significantly less than inside this docker image (40mins build times what?)

BruebachL: I don't know what to decide on this.
If the intent is to distribute this dockerfile so people can build it and then host their own server, building it once and then discarding the build artifacts á la multi-stage seems fine.
If the intent is to distribute this as an aid to servatrice developers, caching makes sense, since this will slim down compile times.
Compiling a separate servatrice binary that the docker file consumes seems outside the scope of this PR.

BruebachL: I'd like to note that the fragile breakage you described earlier already happened to me during development, which is why I personally opted to copy everything and discard after.

Dockerfile
```diff
@@ -1,8 +1,7 @@
-FROM ubuntu:24.04
-
-ARG DEBIAN_FRONTEND=noninteractive
```
ebbit1q, line 3: is this env var not needed anymore in current ubuntu?

Dockerfile
```diff
@@ -1,8 +1,7 @@
...
+# -------- Build Stage --------
+FROM ubuntu:24.04 AS build
 
-RUN apt-get update && apt-get install -y\
+RUN apt-get update && apt-get install -y \
```
ebbit1q: should we add --no-install-recommends to these?

Dockerfile
BruebachL: It seemed to run fine without but I guess there's also really no harm in setting it.

Dockerfile
BruebachL: Sure.

Dockerfile
```diff
@@ -16,20 +17,25 @@ RUN apt-get update && apt-get install -y\
...
-  make install
+# -------- Runtime Stage (clean) --------
+FROM ubuntu:24.04
 
-WORKDIR /home/servatrice
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  libmariadb-dev-compat \
+  libprotobuf-dev \
+  libqt6sql6-mysql \
+  qt6-websockets-dev
```
ebbit1q: if we want to save space we should add 
\
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*

this removes the apt cache after getting these deps

Dockerfile
ebbit1q: we set it in all our .ci docker images

Dockerfile
```diff
@@ -16,20 +17,27 @@ RUN apt-get update && apt-get install -y\
...
-  make install
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  libmariadb-dev-compat \
+  libprotobuf-dev \
+  libqt6sql6-mysql \
+  qt6-websockets-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
-WORKDIR /home/servatrice
```
ebbit1q, line 31: should we still specify the workdir for the runtime stage? I think it might save logs there

libcockatrice_network/libcockatrice/network/client/remote/remote_client.h
```diff
@@ -10,7 +10,6 @@
 #include "../abstract/abstract_client.h"
 
 #include <QLoggingCategory>
-#include <QTcpSocket>
```
ebbit1q, line 13: it now compiles with this include in since #6304, I think we should keep it

Dockerfile
BruebachL: Workdir is specified in the runtime stage.

Dockerfile
```diff
@@ -16,20 +17,27 @@ RUN apt-get update && apt-get install -y\
...
 
-WORKDIR build
-RUN cmake .. -DWITH_SERVER=1 -DWITH_CLIENT=0 -DWITH_ORACLE=0 -DWITH_DBCONVERTER=0 &&\
-  make &&\
-  make install
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  libmariadb-dev-compat \
+  libprotobuf-dev \
+  libqt6sql6-mysql \
+  qt6-websockets-dev \
```
ebbit1q: for runtime the -dev deps should not be required

PR #6273, opened 2025-11-05
Give settings managers default groups instead of manually specifying them everywhere.

BruebachL:
SettingsManagers are pretty verbose.

- Reduce verbosity by providing some default arguments while still allowing the option to manually specify group and subgroup.

### DIFF
--- a/dbconverter/src/mocks.cpp
+++ b/dbconverter/src/mocks.cpp
@@ -2,7 +2,7 @@
 #include "mocks.h"
 
 CardDatabaseSettings::CardDatabaseSettings(const QString &settingPath, QObject *parent)
-    : SettingsManager(settingPath + "cardDatabase.ini", parent)
+    : SettingsManager(settingPath + "cardDatabase.ini", QString(), QString(), parent)
 {
 }
 void CardDatabaseSettings::setSortKey(QString /* shortName */, unsigned int /* sortKey */)


ebbit1q:
I'm a bit hesitant with this, are we sure this doesn't change the existing settings format?

BruebachL:
> I'm a bit hesitant with this, are we sure this doesn't change the existing settings format?

Yes. The existing invocation already set the default group/subgroup to "". We set it to QString and this PR only changes invocations where primary groups are concerned (i.e. the subgroup was already set to the empty string) in which case the method for adding values skips the group/subgroup.

Invocations where both a group and a subgroup are set still use the verbose old method.

libcockatrice_settings/libcockatrice/settings/settings_manager.cpp
```diff
@@ -1,10 +1,35 @@
...
+                                 const QString &_defaultSubGroup,
+                                 QObject *parent)
+    : QObject(parent), settings(settingPath, QSettings::IniFormat), defaultGroup(_defaultGroup),
+      defaultSubGroup(_defaultSubGroup)
 {
 }
 
+void SettingsManager::setValue(const QVariant &value, const QString &name)
+{
+    if (!defaultGroup.isEmpty()) {
```
RickyRister, line 14: Is there a reason you can't just call the 4-param `setValue` method here using the default params?

libcockatrice_settings/libcockatrice/settings/settings_manager.h
```diff
@@ -16,14 +16,23 @@ class SettingsManager : public QObject
...
 
 protected:
     QSettings settings;
-    void setValue(const QVariant &value, const QString &name, const QString &group = "", const QString &subGroup = "");
-    void deleteValue(const QString &name, const QString &group = "", const QString &subGroup = "");
+    QString defaultGroup;
+    QString defaultSubGroup;
+    void setValue(const QVariant &value, const QString &name);
+    void
+    setValue(const QVariant &value, const QString &name, const QString &group, const QString &subGroup = QString());
```
RickyRister, line 33: nit: I prefer `= ""` over `= QString()` because I find it easier to parse at a glance.

RickyRister (approved):
Left some comments about style, but I'm in favor of this refactor.

libcockatrice_settings/libcockatrice/settings/settings_manager.cpp
BruebachL: It's because of the invocation logic.
You either call the 2 parameter variant and accept the default group and subgroup or you can call the 4 parameter variant, in which case you *must* specify a custom group but can optionally also specify a subgroup.

PR #6274, opened 2025-11-05
Move models to lib

BruebachL:
Models are GUI exclusives, grouping them with the library they logically belong to introduces a dependency on Qt GUI libraries for every other library, which is undesirable and leads to unnecessary Qt GUI dependencies being pulled into other projects, namely Servatrice.

- Move models and deck_loader to libcockatrice_models

### DIFF
--- a/libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
+++ b/libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
@@ -1,8 +1,7 @@
 #include "card_database_model.h"
 
-#include "../../../../libcockatrice_card/libcockatrice/card/database/card_database.h"
-
 #include <QMap>
+#include <libcockatrice/card/database/card_database.h>
 
 #define CARDDBMODEL_COLUMNS 6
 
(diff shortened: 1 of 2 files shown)


libcockatrice_deck_list/CMakeLists.txt
```diff
@@ -3,13 +3,8 @@ set(CMAKE_AUTOUIC ON)
...
 
 set(HEADERS
-    libcockatrice/deck_list/abstract_deck_list_card_node.h
-    libcockatrice/deck_list/abstract_deck_list_node.h
-    libcockatrice/deck_list/deck_list.h
-    libcockatrice/deck_list/deck_list_card_node.h
-    libcockatrice/deck_list/deck_list_model.h
-    libcockatrice/deck_list/deck_list_sort_filter_proxy_model.h
-    libcockatrice/deck_list/deck_loader.h
+    libcockatrice/deck_list/abstract_deck_list_card_node.h libcockatrice/deck_list/abstract_deck_list_node.h
```
RickyRister, line 6: we should put each file on its own line

BruebachL: We should but this is what the linter demands for whatever reason.

ebbit1q: glory to the almighty linter T_T 7

PR #6275, opened 2025-11-05
Introduce additional checks for playerMenu shortcut activation for judges

### LINKED ISSUE #6127: [Beta/Dev] Player refactor regressions

BruebachL:
## Related Ticket(s)
- Fixes #6127

Judge shortcut handling is broken because judges get the shortcuts for every player activated (since they technically have access to their menus) which breaks shortcuts since multiple shortcuts are now assigned to the same action.

- Introduce additional checks for playerMenu shortcut activation for judges when they are accessed by a judge, only activating them if the player is a local player.

### DIFF
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -93,7 +93,7 @@ idleclienttimeout=3600
 ; * password: require users to specify a common password to log in;
 ; * sql: authenticate users against the "users" table of the database;
 ; Please note that only the "sql" method permits to have registered users and store their data on the server.
-method=sql
+method=none
 
 ; if the chosen authentication method is password, here you can define the password your users will use to log in
 password=123456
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/player/menu/player_menu.cpp
```diff
@@ -139,7 +139,14 @@ void PlayerMenu::retranslateUi()
 void PlayerMenu::refreshShortcuts()
 {
     if (shortcutsActive) {
-        setShortcutsActive();
+        // Judges get access to every player's menus but only want shortcuts to be set for their own.
+        if (player->getPlayerInfo()->getLocalOrJudge() && !player->getPlayerInfo()->getLocal()) {
```
ebbit1q, line 143: isn't it that in these cases setShortcutsActive() shouldn't be called in the first place so shortcutsActive would be false? that seems simpler than having it turn it off again on refresh, the result of this check should not change between refreshes ever.

BruebachL: The trouble is that we set this value in odd places, so I'd rather just do the check here for now.

BruebachL: Is all of this happening in the first place because having shortcutsActive uninitialized is a pretty bad idea?

ebbit1q (approved):
sure, we can try to refactor this later, this solves the issue now

PR #6279, opened 2025-11-06
Add more sort options to hand sort

RickyRister:
## Related Ticket(s)
- Follow-up to #6057

The `sort hand` action currently only has a single preset sort order, which sorts by type.
I would like there to be more sort options.

Expanded the `Sort Hand` action into a `Sort Hand by...` menu with three sub-actions: `Name`, `Type`, and `Mana Value`.
Each of the sub-actions can have its own shortcut.

https://github.com/user-attachments/assets/6ed80d2f-5ae7-430e-99f4-a42c88c7ece1

- Refactor hand sort code to parameterize sort options
  - `HandZone::sortHand` now takes in a `SortOption` list instead of having a hardcoded sort order.
    - But automatically appends the name + printing SortOption to the end, so that the same cards get grouped together.
  - Update `HandMenu` with the new menu and actions.
    - Each sort hand action has a `SortOption` data that is extracted in `PlayerActions::actSortHand`
      - The single SortOption is then expanded into a SortOption list depending on the action type
      - Doing it this way because passing a QList<SortOption> in the data is a lot more complicated
- Add new shortcut to shortcut settings 

## Screenshots

Screenshot 2025-11-05 at 11 52 58 PM

Screenshot 2025-11-05 at 11 35 21 PM

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -155,7 +155,7 @@ static QList<CardList::SortOption> expandSortOption(CardList::SortOption option)
         case CardList::SortByManaValue:
             return {CardList::SortByManaValue, CardList::SortByColors};
         default:
-            return {};
+            return {option};
     }
 }
 


cockatrice/src/game/player/menu/hand_menu.cpp
```diff
@@ -17,9 +17,22 @@ HandMenu::HandMenu(Player *_player, PlayerActions *actions, QWidget *parent) : T
         connect(aViewHand, &QAction::triggered, actions, &PlayerActions::actViewHand);
         addAction(aViewHand);
 
-        aSortHand = new QAction(this);
-        connect(aSortHand, &QAction::triggered, actions, &PlayerActions::actSortHand);
-        addAction(aSortHand);
+        mSortHand = addMenu(QString());
+
+        aSortHandByName = new QAction(this);
+        aSortHandByName->setData(CardList::SortByName);
```
ebbit1q, line 23: this all looks good, but I kinda wonder, what's the real benefit of doing this with the action's data part instead of just using three lambdas?

ebbit1q: I mean something like this, instead of the expand sort option stuff:
```cpp
connect(aSortHandByName, &QAction::triggered, actions, [=](){player->getGraphicsItem()->getHandZoneGraphicsItem()->sortHand({CardList::SortByType, CardList::SortByManaValue});});
```

RickyRister: I assume ascor separated the menu creation from the playerAction logic for a reason.

ebbit1q: I'm not sure what you mean, they're still separate, my example is a bit unsophisticated though, it could still call an actSortHand function but with some arguments.

I'm just wondering why we're using the data setting, sender casting, option expanding, etc.
like is there a specific thing that makes this pattern better? like easier to use for something? or is it a pattern that was made before lambdas and can actually be replaced?

ebbit1q: ```[=](){actions->actSortHand({CardList::SortByType, CardList::SortByManaValue});``` (just to illustrate my idea, not a suggestion)

RickyRister: Looking at the existing code in the menu classes, the classes seem extremely reluctant to contain any game logic, instead containing only enough information for `PlayerActions` to identify which action was triggered. 
And they also seem to prefer to pass along info using action data instead of method params.

I consider the exact sort priority for each option to be part of the game logic, so I was reluctant to put any of it in `HandMenu`.

I'll try doing it your way.

ebbit1q: I didn't mean you had to do it the way I talked about, I just wanted to ask a question, like what's up with the data setting thing. If this is the pattern used elsewhere it makes the most sense to use, if we decide using a different pattern is easier to maintain, its better to change the entire menu, I just wanted to start the discussion. I apologize if that was unclear.

cockatrice/src/client/settings/shortcuts_settings.h
```diff
@@ -562,9 +562,15 @@ class ShortcutsSettings : public QObject
                                      ShortcutGroup::Move_selected)},
         {"Player/aViewHand",
          ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Hand"), parseSequenceString(""), ShortcutGroup::View)},
-        {"Player/aSortHand", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Sort Hand"),
-                                         parseSequenceString("Ctrl+Shift+H"),
-                                         ShortcutGroup::View)},
+        {"Player/aSortHandByName", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Sort Hand by Name"),
+                                               parseSequenceString(""),
```
ebbit1q, line 566: I think I'd like some more default shortcuts for these at some point, but we have only so many chords on a keyboard

ebbit1q (approved):
looks good, thank you!

PR #4063, opened 2020-08-12
Fix release tests

olegshtch:
## Related Ticket(s)
- Fixes #4000

## Short roundup of the initial problem

Follows up #4058 

## What will change with this Pull Request?
- Uses real `SettingsCache` and `CardDatabaseSettings` headers with mocked implementation.

PR #6280, opened 2025-11-06
Clean up inter-library dependencies with interfaces

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/.github/workflows/desktop-lint.yml
+++ b/.github/workflows/desktop-lint.yml
@@ -20,7 +20,7 @@ on:
 
 jobs:
   format:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
     - name: Checkout


cockatrice/src/game/player/menu/say_menu.cpp
```diff
@@ -1,25 +1,25 @@
...
+#include "../../../client/settings/cache_settings.h"
 #include "../player.h"
 #include "../player_actions.h"
 
-#include <libcockatrice/settings/cache_settings.h>
-
 SayMenu::SayMenu(Player *_player) : player(_player)
 {
-    connect(&SettingsCache::instance().messages(), &MessageSettings::messageMacrosChanged, this, &SayMenu::initSayMenu);
+    connect(&SettingsCache::instance().get()->messages(), &MessageSettings::messageMacrosChanged, this,
```
ebbit1q: why do we sometimes .get() the pointer but sometimes not?

BruebachL: Slot/Signals are odd.

cockatrice/src/client/settings/cache_settings.h
```diff
@@ -942,7 +946,8 @@ class SettingsCache : public QObject
         return roundCardCorners;
     }
 
-    static SettingsCache &instance();
+    static QSharedPointer<SettingsCache> instance();
+    static void setInstance(QSharedPointer<SettingsCache> newInstance);
```
RickyRister: Is this method used anywhere?

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -12,9 +11,23 @@
 #include <QGlobalStatic>
 #include <QSettings>
 #include <QStandardPaths>
+#include <libcockatrice/settings/card_override_settings.h>
 #include <utility>
 
-Q_GLOBAL_STATIC(SettingsCache, settingsCache);
+QSharedPointer<SettingsCache> SettingsCache::settingsInstance = nullptr;
```
RickyRister: Does this even need to be a shared pointer?
QSharedPointer is meant to be used when the ownership and lifetime of an instance is unclear because it's being stored in multiple places, *and* the instance needs to be deleted once it's no longer being used.

You're making the SettingsCache instance a singleton, so there's absolutely no questions about the ownership or lifetime of the object. And you're never destroying the object until the app is closed.

It should be safe to just pass around a normal pointer to this thing.

libcockatrice_interfaces/libcockatrice/interfaces/interface_card_database_path_provider.h
RickyRister, line 1: Are we naming these interface files `interface_xxx` instead of `i_xxx`? Not sure what the pattern is when it comes to matching the file names to the class names.

BruebachL: I'd rather do it like this, yeah. i_ just seems very nondescript whereas if I see IClassName in code, that does signify shorthand for interface to me.

cockatrice/src/client/settings/cache_settings.cpp
ebbit1q: this used to be a simple pointer before too https://github.com/Cockatrice/Cockatrice/pull/4063/files#diff-e7ac034c0e67a13c5079d2b09781c2c3dbc043591ef22756b102c6b271449260L53

BruebachL: This is resolved.

ebbit1q: yes, it is, I'm just providing extra context

BruebachL: Ah, right, the email notification confused me I guess.

libcockatrice_card/libcockatrice/card/database/parser/cockatrice_xml_4.cpp
```diff
@@ -132,7 +131,8 @@ QVariantHash CockatriceXml4Parser::loadCardPropertiesFromXml(QXmlStreamReader &x
 
 void CockatriceXml4Parser::loadCardsFromXml(QXmlStreamReader &xml)
 {
-    bool includeRebalancedCards = SettingsCache::instance().getIncludeRebalancedCards();
+    bool includeRebalancedCards =
+        true; // TODO: MOVE THIS OUT OF THE PARSER SettingsCache::instance().getIncludeRebalancedCards();
```
ebbit1q: the parser is currently always used with cockatrice or tests, meaning its gui dependency is fine, can we leave this as is and ticket it for later or is this not possible with the current library design?

BruebachL: This would result in pulling GUI into libcockatrice_card via COCKATRICE because SettingsCache lives there. 
We could smuggle this bool into InterfaceCardPreference (since this is *kinda* what this is) or make a new 1 method interface for this.

ebbit1q:
I love polymorphism, can you try running the clang-format version the linter uses? I'll try to figure out why it's broken later.

ebbit1q (approved):
awesome, thanks

PR #6281, opened 2025-11-06
ci: make fat qt libs thin

### LINKED ISSUE #6269: Try use `lipo` to split fat qt libraries into one lib per architecture
The qt binaries downloaded via `aqt` are fat binaries, leading to larger package sizes. It seems to be possible to split these fat binaries using `lipo`.

Initial research:

* https://stackoverflow.com/questions/29369061/splitting-a-multi-architecture-mach-o-binary
* https://www.reddit.com/r/mac/comments/z5j6lm/how_to_strip_universal_apps/
* https://github.com/cievon/apple-fat-binary-splitter

An initial idea is to create a similar script to https://github.com/Cockatrice/Cockatrice/blob/master/cmake/SignMacApplications.cmake that iterates over the list dependency libraries and run `lipo` on them before packaging.

As an added bonus, with some extra work, we can cache only the relevant arch on github actions.

Evidence:
```bash
~/ws/qt/6.9.3/macos/lib/QtCore.framework  file QtCore
QtCore: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
QtCore (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
QtCore (for architecture arm64):	Mach-O 64-bit dynamically linked shared library arm64
```

```bash
~/ws/qt/6.9.3/macos/lib/QtCore.framework  lipo -info QtCore
Architectures in the fat file: QtCore are: x86_64 arm64
```

brunoalr:
## Related Ticket(s)
- Fixes #6269 

The official Qt macos binaries installed via install-qt-action are fat/universal. Since

1. We don't create universal packages for Cockatrice; and
2. We vendor our dependencies in our packages

we can "thin" the Qt binaries according to the target architecture, leading to a ~50MiB reduction, from ~120MiB to ~70MiB in the final Cockatrice packages.

- Add a new CI step to thin the Qt binaries from install-qt-action
- cache them using actions/cache/save and actions/cache/restore - previously we didn't cache Qt since each entry had 1.1GiB, now we need only two ~570MiB entries for all branches.

## Screenshots
Capture-2025-11-07-182044
image

### DIFF
--- a/.ci/resolve_latest_aqt_qt_version.sh
+++ b/.ci/resolve_latest_aqt_qt_version.sh
@@ -1,27 +1,40 @@
 #!/bin/bash
 
-# We cache thinned versions of universal ("fat") Qt binaries to reduce the size of the packages and caches.
-# The universal binaries are install via aqtinstall. We use wildcards to get the latest patch version. E.g. "6.6.*" -> "6.6.3".
-# This script is used to resolve the latest patch version of Qt.
+# This script is used to resolve the latest patch version of Qt using aqtinstall.
+# It interprets wildcards to get the latest patch version. E.g. "6.6.*" -> "6.6.3".
 
-# This script is meant to be used by the ci enironment on macos runners only.
+# This script is meant to be used by the ci enironment.
 # It uses the runner's GITHUB_OUTPUT env variable.
 
 # Usage example: .ci/resolve_latest_aqt_qt_version.sh "6.6.*"
-QT_SPEC=$1
+
+qt_spec=$1
+if [[ ! $qt_spec ]]; then
+  echo "usage: $0 [version]"
+  exit 2
+fi
 
 # If version is already specific (no wildcard), use it as-is
-if [[ "$QT_SPEC" != *"*" ]]; then
-	echo "version=$QT_SPEC" >> "$GITHUB_OUTPUT"
-	exit 0
+if [[ $qt_spec != *"*" ]]; then
+  echo "version $qt_spec is already resolved"
+  echo "version=$qt_spec" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+if ! hash aqt; then
+  echo "aqt could not be found, has aqtinstall been installed?"
+  exit 2
 fi
 
 # Resolve latest patch
-QT_RESOLVED=$(aqt list-qt mac desktop --spec "$QT_SPEC" --latest-version)
-echo "QT_RESOLVED=$QT_RESOLVED"
-if [ -z "$QT_RESOLVED" ]; then
-	echo "Error: Could not resolve Qt version for $QT_SPEC"
-	exit 1
+if ! qt_resolved=$(aqt list-qt mac desktop --spec "$qt_spec" --latest-version); then
+  exit 1
 fi
 
-echo "version=$QT_RESOLVED" >> "$GITHUB_OUTPUT"
+echo "resolved $qt_spec to $qt_resolved"
+if [[ ! $qt_resolved ]]; then
+  echo "Error: Could not resolve Qt version for $qt_spec"
+  exit 1
+fi
+
+echo "version=$qt_resolved" >> "$GITHUB_OUTPUT"
--- a/.ci/thin_macos_qtlib.sh
+++ b/.ci/thin_macos_qtlib.sh
@@ -18,6 +18,7 @@ function thin() {
 }
 export -f thin  # export to allow use in xargs
 export arch
+set -eo pipefail
 
 echo "::group::Thinning Qt libraries to $arch using $nproc cores"
 find "$GITHUB_WORKSPACE/Qt" -type f -print0 | xargs -0 -n1 -P"$nproc" -I{} bash -c "thin '{}'"


.github/workflows/desktop-build.yml
```diff
@@ -254,10 +254,10 @@ jobs:
             make_package: 1
             package_suffix: "-macOS13_Intel"
             artifact_name: macOS13_Intel-package
-            qt_version: 6.6.*
+            qt_version: 6.6.3
```
ebbit1q: before we specified the * to always use the latest point release, what's the reason for pinning it?

.github/workflows/desktop-build.yml
```diff
@@ -254,10 +254,10 @@ jobs:
...
             qt_arch: clang_64
             qt_modules: qtimageformats qtmultimedia qtwebsockets
-            cache_qt: false # qt caches take too much space for macOS (1.1Gi)
+            cache_qt: false # caching manually after thinning
```
ebbit1q: the name and use of this value is now confusing in effect we now always cache, the method of caching only differs based on the os, we might as well remove this flag

.github/workflows/desktop-build.yml
```diff
@@ -361,7 +361,49 @@ jobs:
...
+
+      # the binaries from aqt are fat (universal), so we thin them to the target architecture to reduce the size of the packages and caches
+      - name: Thin Qt ${{matrix.qt_version}} libraries
+        if: matrix.os == 'macOS' && steps.restore_qt.outputs.cache-hit != 'true'
+        run: |
+          export ARCH=$(uname -m)
+          export NPROC=$(sysctl -n hw.ncpu)
+          echo "::group::Thinning Qt libraries to $ARCH using $NPROC cores"
+          find ${{ github.workspace }}/Qt -type f -print0 | \
+            xargs -0 -n1 -P "$NPROC" sh -c \
```
ebbit1q: this thing is ugly, can we wrap the sh -c into a little script in .ci?

.github/workflows/desktop-build.yml
brunoalr: [`QTDIR`](https://github.com/Cockatrice/Cockatrice/pull/6281/files#diff-73d26ed8f688bf0bdc3abc2d453358c8f72c00a45c082b2a2ecae0972018a8feR440-R441) needs the full installation path, which includes the concrete version number. `install-qt-action` exports some env vars with the correct paths, but this action will basically be used once. so I thought it was easier to pin the version number instead of writing another script to find the correct path

.github/workflows/desktop-build.yml
brunoalr: good catch, will change it

.github/workflows/desktop-build.yml
brunoalr: sure

.github/workflows/desktop-build.yml
ebbit1q: I'd prefer we find the correct path so we don't have to up this for every point release, which generally bring important bug fixes

RickyRister:
From a cursory glance, I can confirm that the macOS 15 binary works

ebbit1q:
@RickyRister did you test the arm or x64 version?

RickyRister:
arm

.github/workflows/desktop-build.yml
ebbit1q: would you like me to?

brunoalr:
> From a cursory glance, I can confirm that the macOS 15 binary works

noice, thanks!

.github/workflows/desktop-build.yml
brunoalr: yep, go ahead. been a bit busy lately

.ci/thin_macos_qtlib.sh
```diff
@@ -0,0 +1,24 @@
...
+# the packages and caches using lipo.
+
+# This script is meant to be used by the ci enironment on macos runners only.
+# It uses the runner's GITHUB_WORKSPACE env variable.
+arch=$(uname -m)
+nproc=$(sysctl -n hw.ncpu)
+
+function thin() {
+  local libfile=$1
+  if [[ $(file -b --mime-type "$libfile") == application/x-mach-binary ]]; then
```
ebbit1q:
```suggestion
  if [[ $(file -b --mime-type "$libfile") == application/x-mach-binary* ]]; then
```

it just starts with this text

ebbit1q:
@brunoalr please check out the scripting I added

brunoalr:
> @brunoalr please check out the scripting I added

scripts look good!
however I realize we might not get the new patch versions from qt, since:
- time 1) we use the * pattern to get resolve to the latest available patch version
- time 2) we thin qt and cache the thinned version
- time 3) in a new build we restore the thinned version instead of downloading the fat one - BUT we don't check if a new patch version is available since the `aqt` call is skipper (neither directly, nor via install-qt-action).

My first thought is to add one extra step with a manual call to resolve versions with `aqt` (I'm almost sure the client has support for that, but I don't remember the exact command arguments) - probably we have to do it directly as I don't remember seeing any option for `install-qt-action`.

Let me know if the issue is clear, @ebbit1q .

brunoalr:
one way to resolve the version string is this:

```bash
aqt list-qt mac desktop --spec '6.6.*' --latest-version
```

brunoalr:
~I'm working on it in a separate branch. looks like I will be done soon~˜
done

.github/workflows/desktop-build.yml
```diff
@@ -375,13 +369,73 @@ jobs:
...
+          pipx install aqtinstall
+
+      # Checking if there's a newer, uncached version of Qt available to install via aqtinstall
+      - name: Resolve latest Qt patch version
+        if: matrix.os == 'macOS'
+        id: resolve_qt_version
+        shell: bash
+        # Ouputs the version of Qt to install via aqtinstall
+        run: |
+          # If version is already specific (no wildcard), use it as-is
```
brunoalr: probably better to move to a separate script...

brunoalr: done

.github/workflows/desktop-build.yml
```diff
@@ -375,13 +369,58 @@ jobs:
           key: ccache-${{matrix.runner}}-${{matrix.soc}}-${{matrix.type}}-${{env.BRANCH_NAME}}
           restore-keys: ccache-${{matrix.runner}}-${{matrix.soc}}-${{matrix.type}}-
 
-      - name: Install Qt ${{matrix.qt_version}}
+      - name: Install aqtinstall
+        if: matrix.os == 'macOS'
+        run: |
+          pipx install aqtinstall
```
brunoalr:
```suggestion
        run: pipx install aqtinstall
```

.ci/resolve_latest_aqt_qt_version.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# We cache thinned versions of universal ("fat") Qt binaries to reduce the size of the packages and caches.
+# The universal binaries are install via aqtinstall. We use wildcards to get the latest patch version. E.g. "6.6.*" -> "6.6.3".
```
ebbit1q:
```suggestion
# The universal binaries are installed via aqtinstall. We use wildcards to get the latest patch version. E.g. "6.6.*" -> "6.6.3".
```

ebbit1q:
the amount of new machinery made me a little worried about maintaining it, I've improved the error handling in the new scripts

brunoalr:
> the amount of new machinery made me a little worried about maintaining it

yeah, I feel it too.

one alternative would be to use fixed version numbers and update them manually. qt 6 versions for mac so far had at most *.*.4 patch versions (see https://ddalcino.github.io/aqt-list-server/), so that wouldn't be too much taxing too maintain IMO. maybe we could try to create a separate qt version polling workflow to update this one.

or we just drop this PR altogether... I guess no one is really complaining about package/install sizes anyways, it would be more of a "nice to have" enhancement

WDYT?

> I've improved the error handling in the new scripts

awesome, thanks a lot

RickyRister:
I very much care about the package size. I’ve been complaining about the mac binary size increasing from 70MB to 100MB in 2.11

brunoalr:
> I very much care about the package size. I’ve been complaining about the mac binary size increasing from 70MB to 100MB in 2.11

nice, good that I was wrong :)

ebbit1q:
I think this is ready then

PR #6285, opened 2025-11-08
don't close dm tabs when disconnecting

### LINKED ISSUE #6249: Private Chat Windows not closing on Disconnect
**No Similar Requests**

**Private chat windows usually close upon disconnect from a server. The feature that I am requesting would be that they would only disappear once the client closes.**

**As someone with a spotty internet connection at times and certain friends not using "social media" such as Discord or at least more active on Cockatrice, it can get annoying for the deep back and forth to just be gone from one small ping spike making me disconnect from the server.. I know this is a low priority request but it seems like this wouldn´t be hard to implement either.**

ebbit1q:
## Related Ticket(s)
- Fixes #6249
- #5447 and #5450

when disconnecting no message history is preserved, keeping message tabs open allows users to more easily resume conversations between disconnects.

- do not delete message tabs on disconnect

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
@@ -491,10 +491,6 @@ void TabSupervisor::stop()
         tabsToDelete << i.value();
     }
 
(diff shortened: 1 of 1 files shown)


ebbit1q:
we could make this a setting, but I don't think it's very important, people can easily manually close the tabs

PR #6294, opened 2025-11-09
Deck loader is a gui class.

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
@@ -117,8 +117,8 @@ void DeckEditorDeckDockWidget::createDeckDock()
     activeGroupCriteriaComboBox->addItem(tr("Mana Cost"), DeckListModelGroupCriteria::MANA_COST);
     activeGroupCriteriaComboBox->addItem(tr("Colors"), DeckListModelGroupCriteria::COLOR);
     connect(activeGroupCriteriaComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged), [this]() {
-        deckModel->setActiveGroupCriteria(
-            static_cast<DeckListModelGroupCriteria>(activeGroupCriteriaComboBox->currentData(Qt::UserRole).toInt()));
+        deckModel->setActiveGroupCriteria(static_cast<DeckListModelGroupCriteria::Type>(
+            activeGroupCriteriaComboBox->currentData(Qt::UserRole).toInt()));
         deckModel->sort(deckView->header()->sortIndicatorSection(), deckView->header()->sortIndicatorOrder());
         deckView->expandAll();
         deckView->expandAll();
--- a/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
+++ b/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
@@ -521,7 +521,7 @@ void DeckListModel::sort(int column, Qt::SortOrder order)
     emit layoutChanged();
 }
 
-void DeckListModel::setActiveGroupCriteria(DeckListModelGroupCriteria newCriteria)
+void DeckListModel::setActiveGroupCriteria(DeckListModelGroupCriteria::Type newCriteria)
 {
     activeGroupCriteria = newCriteria;
     rebuildTree();
--- a/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.h
+++ b/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.h
@@ -68,7 +68,7 @@ namespace DeckListModelGroupCriteria
  * @enum DeckListModelGroupCriteria
  * @brief Available grouping strategies for deck visualization.
  */
-enum
+enum Type
 {
     MAIN_TYPE, /**< Group cards by their main type (e.g., creature, instant). */
     MANA_COST, /**< Group cards by their total mana cost. */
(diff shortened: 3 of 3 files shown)


cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.h
```diff
@@ -50,7 +51,8 @@ public slots:
     void cleanDeck();
     void updateBannerCardComboBox();
     void setDeck(DeckLoader *_deck);
-    DeckLoader *getDeckList();
+    DeckList *getDeckList();
```
RickyRister: Is there a reason we need both methods? A `DeckLoader` is always a `DeckList`.

BruebachL: Sometimes you don't really care about the DeckLoader extra bits and this is a convenience to not need to cast

RickyRister: Does c++ not already let you to use DeckLoader anywhere it expects a DeckList?

BruebachL: You'd have to manually q_objectcast each time.

RickyRister: But `DeckLoader` is a subclass of `DeckList`, so you should be able to use it without casting.
is there something weird going on with Qt that prevents inheritance from working?

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -98,92 +86,86 @@ int DeckListModel::columnCount(const QModelIndex & /*parent*/) const
...
-                    case 3: {
-                        return node->getCardCollectorNumber();
-                    }
-                    case 4: {
-                        return node->getCardProviderId();
-                    }
-                    default:
-                        return {};
+                case 0: return group->recursiveCount(true);
+                case 1: return (role == Qt::DisplayRole
```
ebbit1q: why did you format these like this?

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -98,92 +86,86 @@ int DeckListModel::columnCount(const QModelIndex & /*parent*/) const
...
                 return {};
         }
     }
+
+    // Card node
+    switch (role) {
+        case Qt::DisplayRole:
+        case Qt::EditRole:
+            switch (index.column()) {
+            case 0: return card->getNumber();
```
ebbit1q: can we make some constants for these column numbers?

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.h
ebbit1q: it'd instead use implicit casting afaik (unless implicit casting is disabled), so it'd always cast in that case.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: They're sorta defined in headerData().

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: I fear we may never know. Something fucky with my linter and then I had it disabled for pre-commit because of that manual lint with the patches I did yesterday.

BruebachL: Curious that the linter doesn't complain tho.

ebbit1q: cus our linter isn't that strict, it considers both valid even though I personally find use of the ternary expression difficult to read and would prefer normal if statements/cases

it's interesting that your linter did all this because this is a rather big functional change in some situations, it must have found that somehow those didn't apply

in our contributions.md file we have specified that we like the use of braces even for single line statements in these cases but it's not in our linter, personally I think it's perfectly fine for simple return statements

BruebachL: Oh, I thought you meant the fact that there weren't any newlines in the switch statements. Tbh, personally, I don't think the braces for the switch case statements really do anything for readability.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: I've made an enum for these.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
ebbit1q: if the case is just a return statement, yeah the brackets don't do anything in that case, personally I prefer code to not all be on one line but I don't feel very strongly about it

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.h
```diff
@@ -8,11 +8,26 @@
...
 
+namespace DeckRoles {
+enum {
+    IsCardRole = Qt::UserRole + 1,
+    DepthRole,
+    IsLegalRole
+};
+}
+
+enum ModelColumns{
```
ebbit1q: this enum is in the header outside a namespace, kinda causing it to exist in anything that includes this header, I'd prefer it to be in the cpp file or in a namespece (could be the class namespace)

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -366,7 +372,11 @@ void DeckEditorDeckDockWidget::syncBannerCardComboBoxSelectionWithDeck()
  */
 void DeckEditorDeckDockWidget::setDeck(DeckLoader *_deck)
 {
-    deckModel->setDeckList(_deck);
+    deckLoader = _deck;
+
+    deckModel->setDeckList(qobject_cast<DeckList *>(_deck));
```
RickyRister: Unnecessary cast

PR #6295, opened 2025-11-09
Fix crash on rejoining game when reconnecting.

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

SettingsCache isn't initialized yet, I guess. In any case, this dummy set can safely always have a NoopCardPriorityController, since it is just a dummy.

- this
- and this

### DIFF
--- a/cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
+++ b/cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
@@ -1,7 +1,6 @@
 #include "card_picture_to_load.h"
 
 #include "../../client/settings/cache_settings.h"
-#include "libcockatrice/interfaces/noop_card_set_priority_controller.h"
 
 #include <QCoreApplication>
 #include <QDate>
@@ -9,6 +8,7 @@
 #include <QUrl>
 #include <algorithm>
 #include <libcockatrice/card/set/card_set_comparator.h>
+#include <libcockatrice/interfaces/noop_card_set_priority_controller.h>
 
 CardPictureToLoad::CardPictureToLoad(const ExactCard &_card)
     : card(_card), urlTemplates(SettingsCache::instance().downloads().getAllURLs())


ebbit1q:
how could it be that the settingscache is not initialised? isn't the settingscache part of the client and initialised before the client even attempts to connect?

cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
```diff
@@ -1,6 +1,7 @@
 #include "card_picture_to_load.h"
 
 #include "../../client/settings/cache_settings.h"
+#include "libcockatrice/interfaces/noop_card_set_priority_controller.h"
```
ebbit1q: shouldn't the libs have <>?

BruebachL: This is true.

BruebachL:
> how could it be that the settingscache is not initialised? isn't the settingscache part of the client and initialised before the client even attempts to connect?

I'm not sure. It was complaining about something in the destructor of CardSet (which is just defaulted afaik) when picture loads failed, this fixes the crash and is reasonable (the dummy card set does not need its preference adjusted).

ebbit1q:
I'm fine with merging this but this clearly shows there is a rather big problem with at what time the settings cache is initialised!

ebbit1q:
it's very likely to cause different crashes that we haven't found yet/haven't introduced yet, we need to make sure that it can't happen that the settings cache is not initialised yet when it tries to do stuff!

BruebachL:
> it's very likely to cause different crashes that we haven't found yet/haven't introduced yet, we need to make sure that it can't happen that the settings cache is not initialised yet when it tries to do stuff!

I think it has to do with CardSet trying to clean up the straight pointer to SettingsCache::instance()->cardDatabase() that it gets but it was decided that QSharedPointers for interfaces are unnecessary.

ebbit1q:
why would it clean up that pointer? that'd mean it's taking ownership of it

ZeldaZach (approved):

Issue #5516, opened 2025-01-23
Shortcut for Revealing Selected Cards


RoaringInferno:
Simply add a shortcut binding option for the "Reveal to...">>"All Players" action when accessing the right-click menu on a selected card or group of cards.

This should probably be a second issue, but a new action and shortcut binding option for "Reveal selected card(s) to player(s)", which draws a menu that allows the user to select the player(s) that they would like to reveal to, then revealing.

This could be done in lieu of the aformentioned shortcut binding option, but the prior seems vastly simpler, as it is a shortcut for a preexisting command.

PR #6296, opened 2025-11-09
Create "Hand" shortcut group

RickyRister:
## Related Ticket(s)
- Pre-req for #5516

In the shortcut settings, the "sort hand" shortcuts are located under the "View" shortcut group, which makes no sense because that group is meant for the shortcuts for opening zone views.

- Create a new "Hand" shortcut group
  - Move the "sort hand" shortcuts to that group
- Move some code around in the shortcut mapping initializer.

## Screenshots

Screenshot 2025-11-09 at 12 13 25 AM


ebbit1q (approved):
good catch! this seems good, thanks!

PR #6297, opened 2025-11-09
Support shortcut for hand reveal actions

### LINKED ISSUE #5516: Shortcut for Revealing Selected Cards

RickyRister:
## Related Ticket(s)
- Fixes #5516

Shortcuts aren't supported for reveal actions. While it's understandable why shortcut for revealing to specific player isn't supported, we can implement shortcut for revealing to all players by creating and storing the action for "all players" separately from the individual player ones.

The following reveal actions now support shortcuts for the `all players` option:
- `Reveal hand to...` and `Reveal random card to` in the hand menu
- `Reveal to...` in the card menu

https://github.com/user-attachments/assets/48d8106e-a3be-423f-9a9a-98bf5e9f36e7

- In `HandMenu` and `CardMenu`, store the `QAction` for the `All players` option of the reveal menus as a field in the class.
  - Create those actions in the constructor alongside the other action creature, instead of the code that populates the menu with active players
    - that code now inserts the corresponding all players action at the top instead of creating a new one each time.
  - Handle the new actions in that class's setShortcutsActive/setShortcutsInactive/restranslateUi 
- Add the new shortcuts to the shortcuts menu
  - No default shortcuts set for them yet

## Screenshots

Screenshot 2025-11-09 at 1 14 47 AM

### DIFF
--- a/cockatrice/src/game/player/menu/card_menu.cpp
+++ b/cockatrice/src/game/player/menu/card_menu.cpp
@@ -303,6 +303,15 @@ void CardMenu::createHandOrCustomZoneMenu()
     }
 }
 
+/**
+ * @brief Populates the menu with an action for each active player.
+ *
+ * The "all players" action is created separately, so it has to be passed into this function.
+ * It will be put at the top of the menu.
+ *
+ * @param menu The menu to add the player actions to.
+ * @param allPlayersAction The action for "all players".
+ */
 void CardMenu::initContextualPlayersMenu(QMenu *menu, QAction *allPlayersAction)
 {
     allPlayersAction->setData(-1);


cockatrice/src/game/player/menu/card_menu.cpp
```diff
@@ -480,9 +487,6 @@ void CardMenu::setShortcutsActive()
     aSelectRow->setShortcuts(shortcuts.getShortcut("Player/aSelectRow"));
     aSelectColumn->setShortcuts(shortcuts.getShortcut("Player/aSelectColumn"));
 
-    aPlayFacedown->setShortcuts(shortcuts.getShortcut("Player/aPlayFacedown"));
-    aPlay->setShortcuts(shortcuts.getShortcut("Player/aPlay"));
```
ebbit1q, line 484: this was set twice for some reason? strange

cockatrice/src/game/player/menu/card_menu.cpp
```diff
@@ -301,9 +303,10 @@ void CardMenu::createHandOrCustomZoneMenu()
     }
 }
 
-void CardMenu::initContextualPlayersMenu(QMenu *menu)
+void CardMenu::initContextualPlayersMenu(QMenu *menu, QAction *allPlayersAction)
```
ebbit1q, line 315: can you add a documentation comment to this function that says that the allPlayersAction is passed because it's always there?

ebbit1q (approved):
nice, thank you!

PR #6304, opened 2025-11-11
Clean up library qt link targets

BruebachL:
Took 38 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/libcockatrice_network/libcockatrice/network/client/CMakeLists.txt
+++ b/libcockatrice_network/libcockatrice/network/client/CMakeLists.txt
@@ -11,6 +11,8 @@ add_library(libcockatrice_network_client INTERFACE)
 target_include_directories(libcockatrice_network_client INTERFACE .)
 
 target_link_libraries(
-  libcockatrice_network_client INTERFACE ${COCKATRICE_QT_MODULES} libcockatrice_network_client_abstract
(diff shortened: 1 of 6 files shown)


libcockatrice_network/libcockatrice/network/client/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ add_library(libcockatrice_network_client INTERFACE)
 target_include_directories(libcockatrice_network_client INTERFACE .)
 
 target_link_libraries(
-  libcockatrice_network_client INTERFACE ${COCKATRICE_QT_MODULES} libcockatrice_network_client_abstract
-                                         libcockatrice_network_client_local libcockatrice_network_client_remote
+  libcockatrice_network_client
+  INTERFACE ${COCKATRICE_QT_VERSION_NAME}::Network ${COCKATRICE_QT_VERSION_NAME}::WebSockets
```
ebbit1q, line 15: afaik currently we don't specify any of these outside of the findqt cmake, I'd like to have this use be consistant, either we do it this way for every lib or we add a thing similar to the current QT_CORE_MODULE (which probably should be called COCKATRICE_QT_CORE_MODULE)

PR #6305, opened 2025-11-11
Remove Fedora 41 Build and Add Fedora 43 Build

SlightlyCircuitous:

Fedora 40 will go end on life on November 19, 2025, so we will not need to build for it going forward. Fedora 43 was released on October 28, 2025 and will be supported until December 9, 2026, so it would be useful to start building for it instead.

- Removes Fedora 41 from Github build workflow
- Adds Fedora 43 to Github build workflow
- Updates Release Template to reflect above changes

I will test the build from this PR in a Fedora 43 VM and ensure that it installs and works properly.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -131,12 +131,12 @@ jobs:
             package: DEB
 
           - distro: Fedora
-            version: 41
+            version: 42
             package: RPM
             test: skip    # Running tests on all distros is superfluous
 
           - distro: Fedora
-            version: 42
+            version: 43
             package: RPM
 
           - distro: Ubuntu


SlightlyCircuitous:
The .rpm from Github Actions installs correctly on my Fedora 43 VM, Cockatrice opens without incident, and oracle runs appropriately on start up. I can participate in local games, and connect to servers as expected. 

The only difference I can see between my Fedora VM install and my normal Ubuntu install (running Beta 33) are some text aberrations on the welcome page:
Screenshot From 2025-11-11 12-34-21
I don't use Fedora normally, so I don't know if this is unique to 43 or unique to Fedora or what, but I can spin up some VMs and check if anyone is concerned.

ebbit1q:
I've seen these text issues before, seems like it's something with fonts, worth making a ticket for

PR #6306, opened 2025-11-11
A DeckLoader is not a DeckList.

BruebachL:
## Related Ticket(s)
- Fixes crash when editing the name or comment of a deck list in deck editor

- this
- and this


RickyRister:
In the future, we might want to remove the `DeckLoader` instances entirely and make its instance methods into static methods that take or return `DeckList`s

PR #6308, opened 2025-11-12
fix use of wrong parent when removing arrows for a player

ebbit1q:
fixes some arrows not being removed when conceding in a multiplayer game, regression introduced in ~~#6210~~


ebbit1q:
correction, this one's actually also https://github.com/Cockatrice/Cockatrice/pull/6156

PR #6312, opened 2025-11-12
[Oracle] Use NoOp setting providers

RickyRister:

No reason to use the actual settings providers when parsing the cards in oracle. 

This gets us closer to decoupling oracle from cockatrice

- Use the no-op `ICardSetPriorityController` and `ICardPreferenceProvider` instead of getting the settings from `SettingsCache`

Unfortunately, we can't remove any dependencies from the cmake yet because oracle still depends on `SettingsCache` which depends a bunch of other things.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -41,14 +41,14 @@ static CardSet::Priority getSetPriority(const QString &setType, const QString &s
 
 bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
 {
-    QList<SetToDownload> newSetList;
-
     bool ok;
     auto setsMap = QtJson::Json::parse(QString(data), ok).toMap().value("data").toMap();
     if (!ok) {
         qDebug() << "error: QtJson::Json::parse()";
         return false;
     }
+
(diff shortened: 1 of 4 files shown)


oracle/src/oracleimporter.cpp
```diff
@@ -463,14 +464,16 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QList
 
 int OracleImporter::startImport()
 {
+    static ICardSetPriorityController *noOpController = new NoopCardSetPriorityController();
+
     int setCards = 0, setIndex = 0;
     // add an empty set for tokens
-    CardSetPtr tokenSet = CardSet::newInstance(SettingsCache::instance().cardDatabase(), CardSet::TOKENS_SETNAME,
```
ebbit1q, line 462: does it not initialise the card database settings cache anymore? oracle does need the settings cache but not all parts of it

RickyRister: I don’t think it needs the card database settings anymore. It still needs the settings cache for language and download url settings.

BruebachL (approved):

PR #6315, opened 2025-11-12
[Doxygen] Card Picture Loader

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
+++ b/cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
@@ -46,6 +46,19 @@ QList<CardSetPtr> CardPictureToLoad::extractSetsSorted(const ExactCard &card)
     return sortedSets;
 }
 
+/**
+ * Finds the PrintingInfo corresponding to the exactCards's card name that belongs to a given set and has the
+ * exactCards's providerId.
+ * If the set name is in the CardInfo, but no printings in that set match the card's providerId, then the first
+ * PrintingInfo for the set is returned.
+ *
+ * This method only exists to maintain existing behavior.
+ * TODO: check if going through all sets is still necessary after the ExactCard refactor.
+ *
+ * @param card The card to look in
+ * @param setName The set's short name
+ * @return A PrintingInfo, or a default-constructed PrintingInfo if the set name is not in the CardInfo.
+ */
 static PrintingInfo findPrintingForSet(const ExactCard &card, const QString &setName)
 {
     SetToPrintingsMap setsToPrintings = card.getInfo().getSets();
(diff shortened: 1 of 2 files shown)


RickyRister:
Some of the methods you added docs to also have doc comments in the .cpp files.
What are we doing about them?

BruebachL:
> Some of the methods you added docs to also have doc comments in the .cpp files. What are we doing about them?

It's fine to document on .cpp as well. Doxygen will merge the two documentations. .h files are for documenting and general overview, if you are documenting in detail or concrete implementation details, you document in .cpp

That's at least the general philosophy, I haven't checked if that applies to the comments here.

RickyRister:
I took a quick look through the docs in the cpp files here. 
Pretty much all of the docs in the cpp files are redundant with the new docs, if not containing even less details than the new docs.

We don't need to remove the old docs now, but we should at least have a plan for cleaning up redundant docs.

BruebachL:
> I took a quick look through the docs in the cpp files here. Pretty much all of the docs in the cpp files are redundant with the new docs, if not containing even less details than the new docs.
> 
> We don't need to remove the old docs now, but we should at least have a plan for cleaning up redundant docs.

I cleaned them for now.

cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
```diff
@@ -52,19 +46,6 @@ QList<CardSetPtr> CardPictureToLoad::extractSetsSorted(const ExactCard &card)
...
- * If the set name is in the CardInfo, but no printings in that set match the card's providerId, then the first
- * PrintingInfo for the set is returned.
- *
- * This method only exists to maintain existing behavior.
- * TODO: check if going through all sets is still necessary after the ExactCard refactor.
- *
- * @param card The card to look in
- * @param setName The set's short name
- * @return A PrintingInfo, or a default-constructed PrintingInfo if the set name is not in the CardInfo.
- */
```
RickyRister, line 67: Keep this doc around. It's a static method so it's not in the hpp file

cockatrice/src/interface/card_picture_loader/card_picture_to_load.h
```diff
@@ -12,37 +6,95 @@
...
+
+    /**
+     * @brief Transforms a URL template into a concrete URL for this card/set.
+     * @param urlTemplate The URL template to transform
+     * @return The transformed URL or empty string if the template cannot be fulfilled
+     */
     QString transformUrl(const QString &urlTemplate) const;
+
+    /**
+     * @brief Advance to the next set in the list.
```
RickyRister, line 68: Actually, `nextSet` and `nextUrl` are the only methods that have details in the cpp docs that aren't present in the header docs.
Can you move over the details about `If we are already at the end of the list, then currentSet is set to empty.` to here.

RickyRister (approved):
Left a couple comments. Happy with the docs otherwise.

PR #6317, opened 2025-11-13
Set DeckList parent when ownership changes

BruebachL:
## Related Ticket(s)
- Fixes crash when saving a deck loaded from clipboard

Stuff isn't initialized and thus leads to an NPE on access.

- this
- and this


libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
```diff
@@ -76,6 +76,8 @@ void SideboardPlan::write(QXmlStreamWriter *xml)
 }
 
 DeckList::DeckList()
+    : QObject(), name(QString()), comments(QString()), bannerCard({}), lastLoadedTimestamp(QString()),
+      tags(QStringList()), cachedDeckHash(QString())
```
RickyRister: We figured out that the root cause of the crash is unrelated to this.

Which makes sense because all those fields are default-constructible, so they shouldn't require the constructor to explicitly initialize them.

RickyRister:
Please rename the PR before merging

Issue #5945, opened 2025-05-17
Please also use the flavorName field when searching card names.


Qriist:
Another player was trying to add Edgin to his deck and was unable to find the card as it wasn't showing up in the card browser. It took us a moment to realize that the card now goes by Bohn.

![Image](https://github.com/user-attachments/assets/b0c0a799-7817-4939-8ab3-388be3680209)

This should be easily exposable as AllPrintings.json has a `flavorName` tag on cards that suffer from WOTC's poor decisions.
<img width="221" alt="Image" src="https://github.com/user-attachments/assets/0b0b4dad-8ce7-4e2b-86d9-2279d3c6a0d4" />

ebbit1q:
it'd be possible to add an extra tag to the props of this card but it'd also require the client to search for both this flavorname and the real name in the editor

RickyRister:
I looked into updating oracle to handle flavorNames. There's two problems I ran into:
- Turns out the flavorName field in the `AllPrintings.json` varies based on printing. In addition to making it require special handling on the oracle side, that means it's theoretically possible for a card to have multiple flavorNames (though I don't think it has happened yet). We need to decide on how we're handling that scenario. For `cards.xml`, the naive solution of just storing the flavorName as a field in the `prop` section might be not enough.
- The flavorName is used for gozilla treatments too. If we make it so searches use both the real name and flavorName, that will cause cards which only match the flavorName due to some random promo printing to show up. Unlike scryfall, we don't have the ability to always show the printing corresponding to the flavorName when the card is displayed, so this might lead to confusion about why the name search is bringing up unrelated cards.

BruebachL:
Point 1: Couldn't we technically store a string list in the property?
Point 2: What about a checkbox in a settings cogwheel next to the search bar? It seems like the simplest option for this to be quickly configurable.

RickyRister:
> Point 1: Couldn't we technically store a string list in the property?

The idiomatic way to represent lists in xml is to have repeated child entries of the same type. The xml parser assumes the child entries of `prop` are all flat, so we would have to add logic there to handle that. Also, even though `CardInfo` stores its miscellaneous properties as a `QVariantHash`, most of our code seems to assume that properties only contains strings.

We probably want to just add `flavorNames` as a new top-level field under `card`. Throwing it in the `properties` map is going to add a bunch of new edge cases to using the properties map that I don't want to think about.

[...]

PR #6319, opened 2025-11-15
Introduce `altNames` property on cards

RickyRister:
## Related Ticket(s)
- Related to #5945
- Alternative implementation of #6335

We want to support flavorNames in Cockatrice because players are confused when wotc prints a UW version of a UB card and now the UB name doesn't show up when name searching.

However, as noted in https://github.com/Cockatrice/Cockatrice/issues/5945#issuecomment-3525708939 , the `flavorName` field can vary between printings, which means it's possible for a card to have multiple flavorNames.

The solution I decided to go with is to create a new `altNames` field to contain the possible flavorNames.

- Added a `QSet<QString>` `altNames` field to `CardInfo`
  - Updated ctor / factory method to take a new param
  - Added a `addAltName` method since it's useful for oracle
- Updated both version of the xml classes to handle both parsing and saving the name field
- Updated oracle to collect the `flavorName` from all printings of a card and combine it into the `altNames` set

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -137,13 +137,23 @@ CardInfoPtr OracleImporter::addCard(QString name,
     // Workaround for card name weirdness
     name = name.replace("Æ", "AE");
     name = name.replace("’", "'");
+
     if (cards.contains(name)) {
         CardInfoPtr card = cards.value(name);
         card->addToSet(printingInfo.getSet(), printingInfo);
         if (card->getProperties().filter(formatRegex).empty()) {
             card->combineLegalities(properties);
         }
+        if (properties.contains("flavorName")) {
+            card->addAltName(properties.value("flavorName").toString());
+        }
         return card;
+    }
+
+    // handle altNames
+    QSet<QString> altNames;
+    if (properties.contains("flavorName")) {
+        altNames.insert(properties.value("flavorName").toString());
     }
 
     // Remove {} around mana costs, except if it's split cost
@@ -206,8 +216,9 @@ CardInfoPtr OracleImporter::addCard(QString name,
     QList<CardRelation *> reverseRelatedCards;
     SetToPrintingsMap setsInfo;
     setsInfo[printingInfo.getSet()->getShortName()].append(printingInfo);
-    CardInfoPtr newCard = CardInfo::newInstance(name, text, isToken, properties, relatedCards, reverseRelatedCards,
-                                                setsInfo, cipt, landscapeOrientation, tableRow, upsideDown);
+    CardInfoPtr newCard =
+        CardInfo::newInstance(name, altNames, text, isToken, properties, relatedCards, reverseRelatedCards, setsInfo,
+                              cipt, landscapeOrientation, tableRow, upsideDown);
 
     if (name.isEmpty()) {
         qDebug() << "warning: an empty card was added to set" << printingInfo.getSet()->getShortName();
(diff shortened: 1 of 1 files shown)


RickyRister:
After running the new oracle, here's what I noticed.

The example in the issue is finally fixed!
```xml
<card>
    <name>Bohn, Beguiling Balladeer</name>
    <text>Each nonland card in your hand without foretell has foretell. Its foretell cost is equal to its mana cost reduced by {2}. (During your turn, you may pay {2} and exile it from your hand face down. Cast it on a later turn for its foretell cost.)
Whenever you cast your second spell each turn, goad target creature an opponent controls.</text>
    <altNames>
        <altName>Edgin, Larcenous Lutenist</altName>
    </altNames>
```

Seems like some cards do have multiple flavorNames already. For example,
```xml
<card>
    <name>Spell Blast</name>
    <text>Counter target spell with mana value X. (For example, if that spell's mana cost is {3}{U}{U}, X is 5.)</text>
    <altNames>
        <altName>Hechizo en ráfaga</altName>
        <altName>Salve d'antimagie</altName>
    </altNames>
```
Probably due to some early magic foreign language sets where the foreign name got classified as a flavorName.

Likewise, seems like the strixhaven bonus sheet cards printings got their japanese names as a flavorName
```xml
<card>
    <name>Abundant Harvest</name>
    <text>Choose land or nonland. Reveal cards from the top of your library until you reveal a card of the chosen kind. Put that card into your hand and the rest on the bottom of your library in a random order.</text>
    <altNames>
        <altName>豊穣な収穫</altName>
    </altNames>
```

Seems like I didn't handle DFCs properly, because they don't have the field.

```xml
<card>
    <name>Spider-Man, Miles Morales</name>
    <text>Vigilance, trample (Attacking doesn't cause this creature to tap. He can deal excess combat damage to the player he's attacking.)
Whenever Spider-Man enters or attacks, put a +1/+1 counter on each other creature you control. Those creatures gain trample until end of turn.</text>
    <prop>
```

DFCs have different properties which means another edge case to handle ugh.

```
"faceFlavorName": "Cren, Undercity Dreamer",
"faceManaValue": 2.0,
"faceName": "Miles Morales",
"facePrintedName": "Cren, Undercity Dreamer",
```

tooomm:
The new property should also be added to [doc/carddatabase_v4/cards.xsd](https://github.com/Cockatrice/Cockatrice/blob/master/doc/carddatabase_v4/cards.xsd).

ebbit1q:
why not add the field to props?

RickyRister:
> why not add the field to props?

As I mentioned [here](https://github.com/Cockatrice/Cockatrice/issues/5945#issuecomment-3525949027), much of the code seems to be writting assuming `properties` only contains a flat structure. Adding support for lists in `properties` would add more edge cases to deal with, which I don't want to think about.

ebbit1q:
that's fair, this is clearer anyway

BruebachL:
> The new property should also be added to [doc/carddatabase_v4/cards.xsd](https://github.com/Cockatrice/Cockatrice/blob/master/doc/carddatabase_v4/cards.xsd).

What is this used for/where is this embedded? Can we reuse this for doxygen?

tooomm:
> > The new property should also be added to [doc/carddatabase_v4/cards.xsd](https://github.com/Cockatrice/Cockatrice/blob/master/doc/carddatabase_v4/cards.xsd).
> 
> What is this used for/where is this embedded? Can we reuse this for doxygen?

It's a default schema validation file tailored to our "card db" XML structure.
With the XSD file you can check if e.g. custom card databases for your own game use a correct structure and set of keys supported by Cockatrice.

SlightlyCircuitous, who currently maintains the Cockatrice token file in https://github.com/Cockatrice/Magic-Token manually, uses it for example locally to ensure his changes for the token file are valid XML and work in the app.

<br>

Sure, it could be helpful to include it in the docs I guess!

RickyRister:
Came up with a alternative way to implement flavorName in #6335. Please leave feedback on which implementation you prefer.

ebbit1q:
I think I prefer the other implementation, it makes more sense as it allows us to relate the printing from the altname

PR #6320, opened 2025-11-15
Mark more functions as [[nodiscard]]

BruebachL:
I ran inspection on the project and clang-tidy told me that all these functions should be marked [[nodiscard]]

- Mark functions as [[nodiscard]]


BruebachL:
> imo there kind of isn't a point to marking a const function as nodiscard, like if the value is discarded then nothing happened, which is just fine?

Yeah but why are you getting the value in the first place if you're just gonna discard it. I feel like that deserves a bonking, which is what [[nodiscard]] is about, as far as I understand it but feel free to correct me.

I'm just doing what Clang-tidy says here.

ebbit1q (approved):
I guess it doesn't really hurt, but imo noone would call an accessor "accidentally" meaning that the problem it solves doesn't exist.

PR #6322, opened 2025-11-15
[CardInfo] refactor some fields into a UiAttributes struct

RickyRister:

So, I noticed that `CardInfo` has four fields that are solely used for display and game logic purposes.
Those fields aren't modified at all after they're set.
They're also all passed separately in the constructor / static factory methods, which causes those methods to have a lot of parameters.

I thought that maybe it could be nicer to extract those four fields into a struct.

- Created struct `UiAttributes`, nested inside of `CardInfo`
  - All the fields have default values so you can use the designated initializers with just the differing fields
- Update constructor / factory method to take the struct instead of the fields separately
  - Update usages
- Replaced those getters with a getter for the struct that returns a const reference.
  - Update code to use that

### DIFF
--- a/cockatrice/src/game/board/abstract_card_item.cpp
+++ b/cockatrice/src/game/board/abstract_card_item.cpp
@@ -60,7 +60,8 @@ void AbstractCardItem::refreshCardInfo()
     exactCard = CardDatabaseManager::query()->getCard(cardRef);
 
     if (!exactCard && !cardRef.name.isEmpty()) {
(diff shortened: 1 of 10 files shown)


ebbit1q (approved):
seems good, thanks

ebbit1q:
the one ci fail is just random

BruebachL:
Seems fine and reasonable and integrates well with my refactors, thank you for waiting on this.

PR #6323, opened 2025-11-15
[Doxygen] VDE: Include new pictures.

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

## Screenshots
image

### DIFF
--- a/doc/doxygen-extra-pages/user_documentation/deck_management/exporting_decks.md
+++ b/doc/doxygen-extra-pages/user_documentation/deck_management/exporting_decks.md
@@ -63,25 +63,81 @@ newline delimited and annotated group.
 Example: TODO: Adjust this to be non mtg based.
 
 ```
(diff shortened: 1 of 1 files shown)


BruebachL:
Yes, the card images themselves are copyright free svg's (see https://www.me.uk/cards/) and the xml I edited myself by hand with vim/python. I'll look into hosting it on a repo at some point, it needs a little more polishing.

It did occur to me: Why does Cockatrice not natively offer a way of creating/editing database.xmls?

ebbit1q:
technically it does! but it doesn't really do much, it offers to make custom tokens even though it technically can make any card even though it doesn't fill all properties of cards, it's at card database>edit custom tokens. it used to be that this was the only way of using tokens, users would make their own for their decks and then added them, then used the tokens dialog with the tokens in the deck's token zone

PR #6324, opened 2025-11-15
[Critical/Crash] [DE] Adjust to proxy model indices

BruebachL:
Crash when manipulating deck lists since old code does not map new proxy index to source index

- Map stuff
- Fix crashes


ebbit1q (approved):
seems good, thanks!

Issue #5990, opened 2025-06-14
Game filter for games that don't have a buddy present


RickyRister:
**Description of New Feature**

Add a game filter for `Hide games that don't have a buddy present`. 
This filter will hide all games that don't have a buddy as a player in the game

**Additional Context**

If someone only plays with a fixed group of people, this filter option allows them to find those games quickly. This is much faster than going to the accounts page and right clicking the user to show their games, or looking through the list of games maually.

ebbit1q:
#5991 implemented filtering based on the information given by the server

I am not sure if we actually want to go the extra step and give all users information about which games all users are in at all times

currently it is possible to get this information by doing a lot of server requests initiated by the client anyway (list user's games) so an argument could be made for it

alternatively we could provide an option presented to the server that tells it to only show games by users that are on your buddy list, or even to only show games of mutual buddies (the server does not require users to also be on their buddy's buddy list if you didn't know)

PR #6325, opened 2025-11-15
[Game Selector] Add button to join game as judge as well as convenience filters and doxygen

### LINKED ISSUE #6282: create a separate join button for judges
**Similar Requests**
causes behavior noted in #4947 
judge mode implemented in #3531

judges currently have to hold shift to join a game with their abilities in effect, otherwise they can simply play as normal players. this has to be explicitly explained to judges and is rather fiddly as normal users get to enjoy an error when they do it.

instead I propose the interface simply shows two buttons to privileged users: join game and join game as judge (same for join as spectator), this is much simpler, clients know if they are judges on the server and should be able to know whether or not to hide the extra buttons, normal players should have their clients never try to join as judge by accident

**Additional Context**
this was actually brought up in the original 2019 pr, but forgotten about after the option became exclusive to privileged users, which removed the need for having a "secret" key combination. Zach actually mentions a checkbox in the interface instead of extra buttons, which would also be a good solution (though if used like this it might cause judges to accidentally join as judge sometimes, which is undesired)

BruebachL:
## Related Ticket(s)
- Fixes #6282
- Sort of satisfies #5990

- Add checkboxes for convenient filters
- Add a search bar for filtering by game name
- Add option to hide games not created by buddies
- Set default filter values to be very permissive
- Add a button and context menu action for joining as judge

## Screenshots
image

image

### DIFF
--- a/cockatrice/src/interface/widgets/server/game_selector.cpp
+++ b/cockatrice/src/interface/widgets/server/game_selector.cpp
@@ -261,28 +261,30 @@ void GameSelector::checkResponse(const Response &response)
 
 void GameSelector::actJoin()
 {
-    return joinGame();
+    joinGame();
 }
 
 void GameSelector::actJoinAsJudge()
 {
     if (!(tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsJudge)) {
-        return joinGame();
+        joinGame();
+    } else {
+        joinGame(false, true);
     }
-    return joinGame(false, true);
 }
 
 void GameSelector::actJoinAsSpectator()
 {
-    return joinGame(true);
+    joinGame(true);
 }
 
 void GameSelector::actJoinAsJudgeSpectator()
 {
     if (!(tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsJudge)) {
-        return joinGame(true);
+        joinGame(true);
+    } else {
+        joinGame(true, true);
     }
-    return joinGame(true, true);
 }
 
 void GameSelector::customContextMenu(const QPoint &point)
--- a/cockatrice/src/interface/widgets/dialogs/dlg_filter_games.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_filter_games.cpp
@@ -37,7 +37,7 @@ DlgFilterGames::DlgFilterGames(const QMap<int, QString> &_allGameTypes,
     hideIgnoredUserGames = new QCheckBox(tr("Hide 'ignored user' games"));
     hideIgnoredUserGames->setChecked(gamesProxyModel->getHideIgnoredUserGames());
 
-    hideNotBuddyCreatedGames = new QCheckBox(tr("Hide games not created by buddy"));
+    hideNotBuddyCreatedGames = new QCheckBox(tr("Hide games not created by buddies"));
     hideNotBuddyCreatedGames->setChecked(gamesProxyModel->getHideNotBuddyCreatedGames());
 
     hideOpenDecklistGames = new QCheckBox(tr("Hide games with forced open decklists"));


RickyRister:
How does this interact with the old game filters dialog? Are those two kept in sync or do they overwrite each other?

BruebachL:
> How does this interact with the old game filters dialog? Are those two kept in sync or do they overwrite each other?

Technically the old game filters dialog reconstructs the checkboxes on each invocation so it's consistent with the non-dialog checkboxes but there's no inverse check (it's a TODO)

RickyRister:
Wait, why is the format filter active even when the "filter to format" checkbox isn't checked?
And why does checking that checkbox still cause things to change?

https://github.com/user-attachments/assets/7ef2c207-e463-4297-bc8c-8f1edcd8efc2

BruebachL:
It's flipped around, it's Label: Thing not Thing: Label

As in, you can't "toggle" filter to format, that's what all types is for. You're toggling hide started games.

RickyRister:
The horizontal spacing on mac does not make that obvious at all. The label is closer to the checkbox on the left than to their corresponding checkbox. 

I remember running into the same spacing problem when trying to add checkboxes to the VDS. I don't think I ever found a solution for it. There, we eventually decided to put all the checkboxes into the quick settings menu, which made that issue irrelevant. If we decide to have inline checkboxes here instead of moving them to a quick settings menu, then we do need to eventually figure out a solution to this problem.

Also, there's a lot of empty padding above and below the line. We need to set the layout or widget's padding to 0 to get rid of that.

RickyRister (changes requested):
Is it possible for you to split this PR? I'm fine with the "join game as judge" button and the doxygen improvements, but I need more time to consider the ui for the quick filters.

cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -103,10 +106,17 @@ GameSelector::GameSelector(AbstractClient *_client,
     if (room)
         buttonLayout->addWidget(createButton);
     buttonLayout->addWidget(joinButton);
+    if (tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsJudge) {
+        buttonLayout->addWidget(joinAsJudgeButton);
+    } else {
+        qInfo() << tabSupervisor->getUserInfo()->user_level();
```
RickyRister: remove debug

BruebachL:
You can add manual spacers to visually distinguish.

image

RickyRister:
Looks like mac still isn't kind to your attempts.
Screenshot 2025-11-15 at 7 27 21 PM

I do agree that moving the format dropdown to next to the search bar looks better.

BruebachL:
Try this one for size.

RickyRister (approved):
Here's what it looks like now.
Screenshot 2025-11-16 at 2 18 16 AM
I think I'm alright with how it looks now. Maybe we can try to reduce the vertical spacing more in the future.

cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -238,7 +248,20 @@ void GameSelector::checkResponse(const Response &response)
...
+{
+    return joinGame(false, true);
+}
+
+void GameSelector::actJoinAsPlayerOrJudge()
+{
+    if (!(tabSupervisor->getUserInfo()->user_level() & ServerInfo_User::IsJudge)) {
+        return joinGame();
+    }
+    return joinGame(false, (QApplication::keyboardModifiers() & Qt::ShiftModifier) != 0);
```
ebbit1q: I'd rather do away the hold shift stuff entirely, judges can simply use dedicated button always, they don't need a "fast way" to join games as judge by remembering to hold shift while joining.

cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -270,12 +293,20 @@ void GameSelector::customContextMenu(const QPoint &point)
...
+        menu.addAction(&joinGameAsJudge);
+    }
+
     menu.addAction(&spectateGame);
     menu.addAction(&getGameInfo);
     menu.exec(gameListView->mapToGlobal(point));
 }
 
-void GameSelector::joinGame(const bool isSpectator)
+void GameSelector::joinGame(const bool isSpectator, const bool isJudge)
```
ebbit1q: I think this should be "asJudge", it's a bit confusing when judges join a game as player otherwise

cockatrice/src/interface/widgets/server/game_selector.h
```diff
@@ -26,46 +21,163 @@ class TabRoom;
...
+     */
     void gameJoined(int gameId);
 
 private:
-    AbstractClient *client;
-    TabSupervisor *tabSupervisor;
-    TabRoom *room;
+    AbstractClient *client;       /**< The network client used to communicate with the server. */
+    TabSupervisor *tabSupervisor; /**< Reference to TabSupervisor for managing tabs and rooms. */
+    TabRoom *room;                /**< The current room; nullptr if no room is selected. */
```
ebbit1q: what does this mean with "no room is selected"? in what cases is room null? is it when using the "show [player]'s games" option?

cockatrice/src/interface/widgets/server/game_selector.cpp
BruebachL: I feel as if they might complain if we take this away from them. "This was so convenient" etc.

ebbit1q: we have only a handful of judges, we could ask them but from my point of view the general consensus is that using shift is a horrible ux and that they often forget how to do it at all, I'm pretty sure that adding a dedicated button for it will make the majority of judges (if not all of them) very happy

cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -238,14 +261,30 @@ void GameSelector::checkResponse(const Response &response)
 
 void GameSelector::actJoin()
 {
-    return joinGame(false);
+    return joinGame();
```
RickyRister: Reminder to fix these illogical returns if you're going to do everything in this PR

tooomm:
On Windows, the spacing seems better in general - on mac it was always a bit weird:
image
- no additional vertical gap to the list of games
- the spacing between the options is bigger, so it's easier to tell the individual options apart

Two Observations:
- The dropdown list does not sync with the option selected in the filter dialog and it can show "modern", while actually "standard" and "limited" are applied via the dialog.
- That's the second point... the filter dialog allows to select several options at the same time.
Can the dropdown list for format be a multi-select dropdown to mimic the same behavior?
Currently one can only select and switch between e.g. limited and legacy, but can not select both.
Via the filter dialog that works fine.

cockatrice/src/interface/widgets/server/game_selector_quick_filter_toolbar.cpp
```diff
@@ -0,0 +1,110 @@
...
+    setLayout(mainLayout);
+
+    retranslateUi();
+}
+
+void GameSelectorQuickFilterToolBar::retranslateUi()
+{
+    searchBar->setPlaceholderText(tr("Filter by game name..."));
+    filterToFormatComboBox->setToolTip(tr("Filter by game type/format"));
+    hideGamesNotCreatedByBuddiesCheckBox->setText(tr("Hide games not created by buddies"));
```
tooomm, line 107: I like this wording better!

In the filter dialog the same option is named "...by buddy" instead "...by buddies".
Can you add your plural version there, too?

BruebachL:
> On Windows, the spacing seems better in general - on mac it was always a bit weird: image
> 
>     * no additional vertical gap to the list of games
> 
>     * the spacing between the options is bigger, so it's easier to tell the individual options apart
> 
> 
> Two Observations:
> 
>     * The dropdown list does not sync with the option selected in the filter dialog and it can show "modern", while actually "standard" and "limited" are applied via the dialog.
> 
>     * That's the second point... the filter dialog allows to select several options at the same time.
>       Can the dropdown list for format be a multi-select dropdown to mimic the same behavior?
>       Currently one can only select and switch between e.g. limited and legacy, but can not select both.
>       Via the filter dialog that works fine.

I'd like to tackle syncing in a follow up PR when we have gathered some user feedback on this as well.

We can consider a multi checkbox quick setting. I like a combo box because it feels very "get up and go", which is the point of these quick filters.

PR #6327, opened 2025-11-16
Join as judge button.

BruebachL:
Took 4 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/interface/widgets/server/game_selector.cpp
+++ b/cockatrice/src/interface/widgets/server/game_selector.cpp
@@ -92,6 +92,7 @@ GameSelector::GameSelector(AbstractClient *_client,
         createButton = nullptr;
     }
     joinButton = new QPushButton;
+    joinAsJudgeButton = new QPushButton;
     spectateButton = new QPushButton;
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -238,7 +245,20 @@ void GameSelector::checkResponse(const Response &response)
 
 void GameSelector::actJoin()
 {
-    return joinGame(false);
+    return joinGame();
```
RickyRister, line 248: Wait, I just realized that `return joinGame()` makes no sense. `joinGame` returns void, and the `actJoin` method also returns void. I assume returning a void from a void function is a no-op in C++, but we should probably fix that mistake anyways.

BruebachL: That's pretty terrible, you're right lol

BruebachL: I was wondering "Where is this even returned to?" when I was editing it but I didn't look into it further.

PR #6219, opened 2025-10-03
Doxygen: Updates

tooomm:
## Short roundup of the initial problem
~~Project version number is hardcoded in doxygen config.
I do not see us reliably updating this and staying in sync ever.~~

## What will change with this Pull Request?
- Move all doxygen files (except config in root?) into `/doc/doxygen/` folder & adjust references accordingly
- ~~Use environment variable to set reference version number for docs from tag when auto-deploying~~
- ~~Adapt to system/user preferences when it comes to dark/light mode (dark is still the default)~~
- ~~Add project description~~
- ~~Do not build latex variant of the docs, we only use html --> 30% quicker generation, 50% smaller file~~
- Docs are uploaded as workflow artifacts as well
- Correct workflow triggers for relevant changes only

## Screenshots
image

~~Instead `6219/merge` (reference to this PR's), it shows our `version` tag for the automated builds for new tags/releases.~~

<br>

~~Style now dynamically switches between "light" and "dark" mode, depending on user preferences set in the browser/system (default is dark still):~~
image

<br>

---

~~Warnings to check from beginning of docs generation log:~~
```
warning: ignoring unsupported tag 'PROJECT_ICON' at line 71, file Doxyfile
warning: ignoring unsupported tag 'AUTOLINK_IGNORE_WORDS' at line 400, file Doxyfile
warning: ignoring unsupported tag 'HIDE_UNDOC_NAMESPACES' at line 619, file Doxyfile
warning: ignoring unsupported tag 'EXTERNAL_TOOL_PATH' at line 869, file Doxyfile
warning: ignoring unsupported tag 'WARN_LAYOUT_FILE' at line 937, file Doxyfile
warning: ignoring unsupported tag 'IMPLICIT_DIR_DOCS' at line 1219, file Doxyfile
warning: ignoring unsupported tag 'HTML_COPY_CLIPBOARD' at line 1509, file Doxyfile
warning: ignoring unsupported tag 'HTML_PROJECT_COOKIE' at line 1518, file Doxyfile
warning: ignoring unsupported tag 'PAGE_OUTLINE_PANEL' at line 1780, file Doxyfile
warning: ignoring unsupported tag 'SHOW_ENUM_VALUES' at line 1807, file Doxyfile
warning: ignoring unsupported tag 'RTF_EXTRA_FILES' at line 2263, file Doxyfile
warning: ignoring unsupported tag 'UML_MAX_EDGE_LABELS' at line 2695, file Doxyfile
[...]

PR #6330, opened 2025-11-16
Doxygen: Config updates

tooomm:
## Related Ticket(s)
- This PR takes the straight forward config changes from #6219 for now

- Move layout config file into `/doc/doxygen/` folder & adjust reference accordingly
- Use environment variable to set reference version number for docs from tag when auto-deploying.
  We will not update the version number reliably in this additional space when doing new releases and the docs will be out of sync if we rely on manual updates here.
- Adapt to system/user preferences when it comes to dark/light mode (dark is still the default)
- Add project description
- Do not build latex variant of the docs, we only use html --> 30% quicker generation, 50% smaller file

## Screenshots
(Screenshots might be outdated from a content point of view, included them to show the version tag and dark/light switch)

image

Instead `6219/merge` (reference to the other PR where I took the screenshot from), it shows our `version` tag for the automated builds for new tags/releases.

<br>

Style now dynamically switches between "light" and "dark" mode, depending on user preferences set in the browser/system (default is dark still):
image


Doxyfile
```diff
@@ -48,13 +49,13 @@ PROJECT_NAME           = "Cockatrice"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 2.11
+PROJECT_NUMBER         = $(COCKATRICE_REF)
```
BruebachL, line 52: This will break for anyone manually generating docs.

ebbit1q: did you intentionally approve this and is this not a problem?

BruebachL: It silently breaks i.e. it just doesn't add a version tag on local builds.

It's not *ideal* but I think that's also not the biggest problem when considering the upside we get from this.

BruebachL:
Apologies, I didn't realize this would nuke the actual branch on the repo.

PR #6338, opened 2025-11-19
[DeckLoader] Disable copy constructor

BruebachL:
Crash when opening deck editor from in-game

Probably related, it's early in the morning:

https://doc.qt.io/qt-6/qobject.html#no-copy-constructor-or-assignment-operator
https://doc.qt.io/qt-6/object.html#qt-objects-identity-vs-value

- Disable DeckLoader copy constructor


RickyRister (approved):
Seems to have fixed the crash without introducing any new crashes.

PR #6339, opened 2025-11-19
[DeckList] Disable copy constructor

BruebachL:
## Related Ticket(s)
- Fixes crash when editing deck in clipboard
- Fixes 10 year old to-do

QObjects aren't allowed to have copy constructors.

- Disable copy constructor and adjust.

### DIFF
--- a/cockatrice/src/interface/deck_loader/deck_loader.cpp
+++ b/cockatrice/src/interface/deck_loader/deck_loader.cpp
@@ -159,7 +159,7 @@ bool DeckLoader::saveToFile(const QString &fileName, FileFormat fmt)
     if (result) {
         lastFileName = fileName;
         lastFileFormat = fmt;
-        qInfo() << "Deck was saved -" << result;
+        qCInfo(DeckLoaderLog) << "Deck was saved -" << result;
     }
 
     file.flush();


cockatrice/src/interface/deck_loader/deck_loader.h
```diff
@@ -100,6 +112,54 @@ inline Q_LOGGING_CATEGORY(DeckLoaderLog, "deck_loader")
                                    bool addComments = true,
                                    bool addSetNameAndNumber = true);
 
+    /// @name Serialization (XML)
+    ///@{
```
ebbit1q: what does this comment do?

BruebachL: Groups things together in a named box in the doxygen documentation see https://cockatrice.github.io/docs/da/d68/classDeckList.html

ebbit1q:
this does more than disable the copy constructor, it also changes the deckloader to hold its decklist as a member instead of a pointer.

BruebachL:
> this does more than disable the copy constructor, it also changes the deckloader to hold its decklist as a member instead of a pointer.

It seemed necessary at some point in my refactoring process (I think it was because I was trying to keep getDeckList as const even though it just can't be) but you are correct, it is not.

BruebachL:
Thanks for the sanity check, this is more in-line with the intent.

ebbit1q:
I mean, it doesn't particularly matter but this does make it easier to review

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -533,14 +533,14 @@ void DeckListModel::cleanList()
...
+ * @param _deck The deck.
  */
 void DeckListModel::setDeckList(DeckList *_deck)
 {
-    deckList->deleteLater();
-    deckList = _deck;
-    deckList->setParent(this);
-    rebuildTree();
+    if (deckList != _deck) {
+        deckList = _deck;
```
ebbit1q, line 541: wait, doesn't this reintroduce the same issue with the parenting? this doesn't take ownership of the pointer so the previous thing that was the parent will take the list with it if it were to be destroyed or it might never be.

BruebachL: I don't know, you're not wrong. With this PR, the thing using the model is sort of responsible for keeping the decklists parent around (deckdockwidget reparents the loader, for example.)

I don't know what the clean architectural decision here is. I know this fixed the crashes in my local testing scenarios.

Maybe we should consider QSharedPointers for these use cases.

ebbit1q: would the version before you changed it back to being a pointer work better? if it's a member that's simply copied it might be less efficient but at least guarantee that it won't cause issues when being deallocated randomly and stuff

BruebachL: Qt forbids copying QObjects unfortunately.
https://doc.qt.io/qt-6/object.html#qt-objects-identity-vs-value

I don't think the model should be the one owning the data source, other Qt models work with data that isn't a QObject (meaning it never gets parented or managed by the model)

If we want to use QSharedPointer, we'd have to make DeckList a non-QObject as well.

I think a DeckLoader owning a DeckList isn't the worst starting point, maybe?

cockatrice/src/interface/deck_loader/deck_loader.cpp
```diff
@@ -159,12 +152,14 @@ bool DeckLoader::saveToFile(const QString &fileName, FileFormat fmt)
...
         case CockatriceFormat:
             result = deckList->saveToFile_Native(&file);
+            qCInfo(DeckLoaderLog) << "Saving to " << fileName << "-" << result;
             break;
     }
 
     if (result) {
         lastFileName = fileName;
         lastFileFormat = fmt;
+        qInfo() << "Deck was saved -" << result;
```
RickyRister:
```suggestion
        qCInfo(DeckLoaderLog) << "Deck was saved -" << result;
```

BruebachL: Thanks!

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
RickyRister: I think the end goal *is* to make DeckList a non-QObject. Then we can freely copy around the DeckList without worrying about ownership issues. DeckListModel can be the one to handle all the QObject responsibilities.

RickyRister (commented):
I'm of the opinion that we are better served trying to make `DeckList` no longer a QObject, and make `DeckListModel` have sole access to the underlying data. 

Although this PR has some changes that I agree with, I don't think it moves us towards that goal.
That said, if you think this PR has some changes that are worth making, I'm fine with merging it, and working towards the original goal on top of it.

BruebachL:
This PR makes DeckLoader the de-facto owner of the DeckList and then reparents all DeckLoader instances appropriately (as far as I can tell), so it *does* fix the crash (once again, as far as I can tell), so I would like to merge it so we can cut a beta with it and see how it plays out?

RickyRister:
Didn't we already fix the crash in a previous PR, or was that a different crash?

RickyRister (approved):

BruebachL:
It was a different one.

As a final note before I merge this, I'd like to add that, on reflection, a DeckLoader being the "originator"/owner of a DeckList is sensible. It is, after all, where it came from.

PR #6340, opened 2025-11-19
[DeckEditor] Deck List History Manager.

### LINKED ISSUE #3396: Undo button for Deck Editor
<b>System Information:</b>
irrelevant

I occasionally have a fumble-finger problem with the Deck Editor, specifically when I remove a card, and accidentally click twice on the "-" button, it will remove the card below as well. Unless I'm very carefully paying attention, I may not remember what that card is, so have to reload the deck to find it.

Would it be possible to implement an Undo queue, with a "↺" (undo circle arrow) symbol beside the red "+", and a Ctrl-Z hotkey?

BruebachL:
## Related Ticket(s)
- Fixes #3396

There's no way to undo/redo actions taken in the deck editor.

- Implement memento pattern for DeckList
- Implement DeckListHistoryManager caretaker
- Implement a widget to expose caretaker functionalities with buttons and a list widget
- Add application-wide Undo/Redo shortcuts when a deck editor is open (I couldn't find another way to get it to recognize the focus, feel free to fix)
- Debounce when name and comments are set after editing the line edits. This means that, for example, adding "test" to a deck is now recognized as a single addition, rather than "t", "e", "s", "t".

## Screenshots
https://github.com/user-attachments/assets/3d9aeca4-04ac-4446-918b-b396ea954e89

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_list_history_manager_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_list_history_manager_widget.cpp
@@ -41,9 +41,6 @@ DeckListHistoryManagerWidget::DeckListHistoryManagerWidget(DeckListModel *_deckL
 
     layout->addWidget(historyButton);
 
-    connect(undoButton, &QToolButton::clicked, this, &DeckListHistoryManagerWidget::doUndo);
-    connect(redoButton, &QToolButton::clicked, this, &DeckListHistoryManagerWidget::doRedo);
-
     connect(historyList, &QListWidget::itemClicked, this, &DeckListHistoryManagerWidget::onListClicked);
 
     connect(historyManager, &DeckListHistoryManager::undoRedoStateChanged, this,


libcockatrice_deck_list/libcockatrice/deck_list/deck_list_history_manager.h
```diff
@@ -0,0 +1,96 @@
...
+    void undoRedoStateChanged();
+
+public:
+    explicit DeckListHistoryManager(QObject *parent = nullptr) : QObject(parent)
+    {
+    }
+
+    void save(DeckListMemento *memento)
+    {
+        undoStack.push(memento);
```
RickyRister: why are all the implementation in the header file

BruebachL: It's a really small/simple class, I figured I'd just inline instead of doing another cpp.

RickyRister: This class isn't just a transparent data carrier. There's actual logic in here. 
We don't afford the same kindness for other small classes across the project. 
I still think it's better for consistency to create a cpp file here.

BruebachL: Small things like proxy models and such get to do this but sure, this isn't unreasonable. Just borne out of my workflow at the time.

RickyRister:
Does not seem to be recording card additions. It's only recording removals and sideboard swaps.

https://github.com/user-attachments/assets/2324b5f0-9c0a-4010-a9fa-fc7365e04b0d

libcockatrice_deck_list/libcockatrice/deck_list/deck_list_memento.h
```diff
@@ -0,0 +1,27 @@
+#ifndef COCKATRICE_DECK_LIST_MEMENTO_H
+#define COCKATRICE_DECK_LIST_MEMENTO_H
+#include <QString>
+
+class DeckListMemento
+{
+public:
+    explicit DeckListMemento(QString memento, QString reason = QString()) : memento(memento), reason(reason)
```
RickyRister:
```suggestion
    explicit DeckListMemento(const QString& memento, const QString& reason = QString()) : memento(memento), reason(reason)
```

cockatrice/src/interface/widgets/deck_editor/deck_list_history_manager_widget.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    const QString mode = item->data(Qt::UserRole).toString();
+    int index = item->data(Qt::UserRole + 1).toInt();
+
+    if (mode == "redo") {
+        const auto redoStack = historyManager->getRedoStack();
+        int steps = redoStack.size() - index;
+        for (int i = 0; i < steps; ++i) {
+            historyManager->redo(deckListModel->getDeckList());
```
RickyRister, line 144: Don't we save the entire DeckList state in every memento? Is there a reason we need to step-by-step redo instead of just directly loading the decklist from the memento at that step?

libcockatrice_deck_list/libcockatrice/deck_list/deck_list_history_manager.h
```diff
@@ -0,0 +1,55 @@
...
+        return redoStack;
+    }
+
+    QStack<DeckListMemento *> getUndoStack() const
+    {
+        return undoStack;
+    }
+
+private:
+    QStack<DeckListMemento *> undoStack;
```
RickyRister: Who's responsible for deleting the pointers?
Would it be fine to store the `DeckListMemento` directly instead as pointers? They're small enough that copying them shouldn't be too bad.

BruebachL:
> Does not seem to be recording card additions. It's only recording removals and sideboard swaps.
> example.mp4

DeckDockWidget is the current caretaker (as opposed to AbstractTabDeckEditor) which means it only sees decklist manipulations from the deck list widget.

i.e. if you increment a card with the + button, it will see that addition.

I do plan to pull this up into AbstractTabDeckEditor as well at some point, it's one of those "We are lacking a central way of communicating with the deck list and every widget sorta does its own thing" issues.

I have a PR in the works for this.

cockatrice/src/interface/widgets/deck_editor/deck_list_history_manager_widget.cpp
BruebachL: This pops things off/puts things on the stack correctly, see redo() and undo()

RickyRister: There might be an argument to introduce bulk operation methods for performance reasons, though this is the  cockatrice deck editor we're talking about so performance doesn't really matter in this spot.

BruebachL: Agreed on the batch methods, this is just a first pass to gather some user feedback.

BruebachL:
You're right in that it's not the most efficient way of doing things.

However, the Memento pattern is explicitly a full-state recovery mechanism. Mementos are supposed to capture all the state information so that an object can be restored to an arbitrary full state. The native string decklist export accomplishes this, since it's pretty much like loading from disk.

I haven't checked, to be honest. It *is* just a QString of the DeckList so I can't imagine it to be unreasonable.

My idea for a follow up PR and some improvements was to add a configurable "history list size" so that it auto-deletes which should help people who run out of ram due to too many Mementos?

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -422,6 +473,12 @@ void DeckEditorDeckDockWidget::recursiveExpand(const QModelIndex &index)
     deckView->expand(index);
 }
 
+void DeckEditorDeckDockWidget::expandAll()
+{
+    deckView->expandAll();
+    deckView->expandAll();
```
RickyRister, line 482: wait, why does this get called twice?

BruebachL: It's a bit silly but we have a tree structure with depth 3, first call expands level 1, second call expands level 2 and thus fully expands the tree.

Don't ask me, it's how I've always seen it handled and I couldn't find a way to tell Qt to recurse *down* a tree and expand. We have recursiveExpand which traverses up.

PR #6341, opened 2025-11-19
[BannerCard] Try to restore by providerId

BruebachL:
We use findText() on the comboBox when findData() exists, which is more accurate for restoring the selection.

- Look up bannerCard by name and providerId in the combobox.

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
@@ -13,6 +13,30 @@
 #include <libcockatrice/card/database/card_database_manager.h>
 #include <libcockatrice/utility/trice_limits.h>
 
+static int findRestoreIndex(const CardRef &wanted, const QComboBox *combo)
+{
+    // Try providerId + name (strongest match)
+    if (!wanted.providerId.isEmpty()) {
+        for (int i = 0; i < combo->count(); ++i) {
+            auto pair = combo->itemData(i).value<QPair<QString, QString>>();
+            if (pair.second == wanted.providerId && pair.first == wanted.name) {
+                return i;
+            }
+        }
+    }
+
+    // Try name only
+    for (int i = 0; i < combo->count(); ++i) {
+        auto pair = combo->itemData(i).value<QPair<QString, QString>>();
+        if (pair.first == wanted.name) {
+            return i;
+        }
+    }
+
+    // Not found
+    return -1;
+}
+
 DeckEditorDeckDockWidget::DeckEditorDeckDockWidget(AbstractTabDeckEditor *parent)
     : QDockWidget(parent), deckEditor(parent)
 {
@@ -321,30 +345,6 @@ void DeckEditorDeckDockWidget::updateBannerCardComboBox()
 
     // Restore signal state
     bannerCardComboBox->blockSignals(wasBlocked);
-}
-
-int findRestoreIndex(const CardRef &wanted, const QComboBox *combo)
-{
-    // Try providerId + name (strongest match)
-    if (!wanted.providerId.isEmpty()) {
-        for (int i = 0; i < combo->count(); ++i) {
-            auto pair = combo->itemData(i).value<QPair<QString, QString>>();
-            if (pair.second == wanted.providerId && pair.first == wanted.name) {
-                return i;
-            }
-        }
-    }
-
-    // Try name only
-    for (int i = 0; i < combo->count(); ++i) {
-        auto pair = combo->itemData(i).value<QPair<QString, QString>>();
-        if (pair.first == wanted.name) {
-            return i;
-        }
-    }
-
-    // Not found
-    return -1;
 }
 
 void DeckEditorDeckDockWidget::setBannerCard(int /* changedIndex */)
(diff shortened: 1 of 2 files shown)


ebbit1q:
while testing this change I noticed that it now always sets a banner card whenever I load a deck, it no longer allows me not to have it?

BruebachL:
> while testing this change I noticed that it now always sets a banner card whenever I load a deck, it no longer allows me not to have it?

That doesn't replicate for me, it hits the "-" placeholder correctly

image

image

BruebachL:
Same with a blank deck that I've added one card to:

image

On save:

image

RickyRister:
The banner card when I load the deck into the deck editor is sometimes incorrect

https://github.com/user-attachments/assets/677bbb76-1364-4fbb-bcab-929baac3df93

RickyRister:
Fixed it for me, thanks

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -272,58 +272,77 @@ void DeckEditorDeckDockWidget::updateHash()
...
-    if (restoredIndex != -1) {
-        bannerCardComboBox->setCurrentIndex(restoredIndex);
-        if (deckModel->getDeckList()->getBannerCard().providerId !=
-            bannerCardComboBox->currentData().value<QPair<QString, QString>>().second) {
-            setBannerCard(restoredIndex);
+    // Try restoring by provider ID first (strongest match)
+    int restoreIndex = -1;
+    if (!wantedProvider.isEmpty()) {
+        for (int i = 0; i < bannerCardComboBox->count(); ++i) {
+            auto pair = bannerCardComboBox->itemData(i).value<QPair<QString, QString>>();
```
RickyRister: fyi, QPair supports structured bindings, in case you think it looks more readable that way.
```suggestion
            auto [name, providerId] = bannerCardComboBox->itemData(i).value<QPair<QString, QString>>();
```

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -272,58 +272,77 @@ void DeckEditorDeckDockWidget::updateHash()
...
 
-    // Try to restore the previous selection by finding the currentText
-    int restoredIndex = bannerCardComboBox->findText(currentText);
-    if (restoredIndex != -1) {
-        bannerCardComboBox->setCurrentIndex(restoredIndex);
-        if (deckModel->getDeckList()->getBannerCard().providerId !=
-            bannerCardComboBox->currentData().value<QPair<QString, QString>>().second) {
-            setBannerCard(restoredIndex);
+    // Try restoring by provider ID first (strongest match)
+    int restoreIndex = -1;
```
RickyRister: Actually, extracting the `restoredIndex` calculation code to a function might be good for readability, since it can early return.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -272,58 +272,77 @@ void DeckEditorDeckDockWidget::updateHash()
 
 void DeckEditorDeckDockWidget::updateBannerCardComboBox()
 {
-    // Store the current text of the combo box
-    QString currentText = bannerCardComboBox->currentText();
+    // Store current banner card identity
+    QString wantedName = deckModel->getDeckList()->getBannerCard().name;
+    QString wantedProvider = deckModel->getDeckList()->getBannerCard().providerId;
```
RickyRister: nit: you can just get the CardRef and access the fields on it
```suggestion
    CardRef wanted = deckModel->getDeckList()->getBannerCard();
```

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Oh, cool, thanks! This does look more readable

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.h
```diff
@@ -52,6 +52,7 @@ class DeckEditorDeckDockWidget : public QDockWidget
 public slots:
     void cleanDeck();
     void updateBannerCardComboBox();
+    int findRestoreIndex(const CardRef &wanted, const QComboBox *combo) const;
```
RickyRister: Nice, looks a lot cleaner.
However, why is this function a public slot.
In fact, since you pass in the comboBox, I’m pretty sure it can just be a static function

BruebachL: CLion just dumps it wherever it feels like it (usually around neighboring functions), I think. Thanks for the catch.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.h
```diff
@@ -49,6 +49,8 @@ class DeckEditorDeckDockWidget : public QDockWidget
         return activeGroupCriteriaComboBox;
     }
 
+    const static int findRestoreIndex(const CardRef &wanted, const QComboBox *combo);
```
RickyRister: By static function, I meant making it into a static function in the cpp file, so you don’t have to include it in the header file at all.

But anyways I’m already fine with whatever

BruebachL: I'm having a real struggle with my brain for some reason at the moment.

PR #6342, opened 2025-11-20
[Refactor] Move AbstractGraphicsItem and GraphicsItemType to game_graphics/board

BruebachL:
Doing some work on separating game into game and game_graphics and this is a trivial move refactor that clutters a lot of my other changelists since it is a reasonable base.

- Move refactor the two things into game_graphics/board from game/board


RickyRister (approved):
Seems reasonable, thanks

PR #6343, opened 2025-11-20
[DeckLoader] Fix missing lastFileName when opened from outside deck editor

RickyRister:
## Related Ticket(s)
- Fixes bug introduced in #6306

When another tab needs to open a deck, it calls TabSupervisor with a DeckLoader. However, we don't copy over the last load info when we create a copy of the `DeckLoader` to pass into the deck editor tab, meaning the deck editor will open the save location prompt when you try to save it.

https://github.com/user-attachments/assets/3657f15c-e36f-4f0f-8046-c23cb9261e81

- Extract last load info fields into a `LoadInfo` struct.
  - Still leaving around old getters and setters since I didn't feel like updating the existing code right now. We're still in the middle of redesigning the whole deck structure, so there's a good chance it just gets cleaned up later.
- Make the copy constructor take a `LoadInfo`.
  - And update all usages to pass the loadInfo from the old DeckLoader

https://github.com/user-attachments/assets/6d143633-8b9a-4dc7-ab59-2cfeb57c4ea9

### DIFF
--- a/cockatrice/src/game/deckview/deck_view_container.cpp
+++ b/cockatrice/src/game/deckview/deck_view_container.cpp
@@ -331,7 +331,7 @@ void DeckViewContainer::loadFromWebsite()
 void DeckViewContainer::deckSelectFinished(const Response &r)
 {
     const Response_DeckDownload &resp = r.GetExtension(Response_DeckDownload::ext);
-    DeckLoader newDeck(this, new DeckList(QString::fromStdString(resp.deck())));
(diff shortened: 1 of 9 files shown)


BruebachL:
I think this might be fixed on main since we *do* pass along the DeckLoader pointer to the DeckEditor tab now.

RickyRister:
yeah, you're right. Issue is fixed by #6339

PR #6348, opened 2025-11-24
[Feature] TabArchidekt and Archidekt API integration

BruebachL:
## Related Ticket(s)
- Partially implements #4389

- Implement wrapper objects for Archidekt JSON responses
- Implement a Tab to display Archidekt query listings
- Allow querying by color, format, edhbracket, deck name, deck owner name, contained card, commander name, deck tags, minimum/maximum/exact amount of cards in deck and if it is defined as a 'package' or not.
- Allow clicking on a deck to preview it in a VDE deck view and allow importing it to the VDE

## Screenshots
image
image
image

### DIFF
--- a/cockatrice/src/interface/widgets/general/display/color_bar.cpp
+++ b/cockatrice/src/interface/widgets/general/display/color_bar.cpp
@@ -126,13 +126,16 @@ QString ColorBar::tooltipForPosition(int x) const
     for (int v : colors.values())
         total += v;
 
+    if (total == 0)
+        return {};
+
     int pos = 0;
-    for (auto it = colors.begin(); it != colors.end(); ++it) {
-        double ratio = double(it.value()) / total;
-        int segmentWidth = int(ratio * width());
+    for (auto it = colors.cbegin(); it != colors.cend(); ++it) {
+        const double ratio = double(it.value()) / total;
+        const int segmentWidth = int(ratio * width());
 
         if (x >= pos && x < pos + segmentWidth) {
-            double percent = (100.0 * it.value()) / total;
+            const double percent = (100.0 * it.value()) / total;
             return QString("%1: %2 cards (%3%)").arg(it.key()).arg(it.value()).arg(QString::number(percent, 'f', 1));
         }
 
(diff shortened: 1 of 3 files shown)


cockatrice/src/interface/widgets/general/display/color_bar.cpp
```diff
@@ -0,0 +1,160 @@
...
+}
+
+QString ColorBar::tooltipForPosition(int x) const
+{
+    int total = 0;
+    for (int v : colors.values())
+        total += v;
+
+    int pos = 0;
+    for (auto it = colors.begin(); it != colors.end(); ++it) {
```
ZeldaZach: Could use a const iterator (cbegin/cend)

cockatrice/src/interface/widgets/tabs/api/archidekt/display/archidekt_api_response_deck_entry_display_widget.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+void ArchidektApiResponseDeckEntryDisplayWidget::updateScaledPreview()
+{
+    if (originalPixmap.isNull()) {
+        return;
+    }
+
+    int baseWidth = 400;
+    int newWidth = baseWidth * scaleFactor / 100;
+    int newHeight = int(newWidth * (150.0 / 267.0));
```
ZeldaZach: might make sense to break out DESIGN_RATIO above into a class var so we can reuse it

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.h
```diff
@@ -74,6 +74,29 @@ enum Type
     MANA_COST, /**< Group cards by their total mana cost. */
     COLOR      /**< Group cards by their color identity. */
 };
+static inline QString toString(Type t)
+{
+    switch (t) {
+        case MAIN_TYPE:
+            return "Main Type";
```
ZeldaZach, line 81: tr() ?

BruebachL: tr() isn't available inside this scope for whatever reason, we tr() at usage level

ebbit1q: wrapping the output of this function in tr() is not sufficient, like that the translatable string search fails to ever find it, the translation generator is simple and just searches for tr("hardcoded string") it is not possible for it to find tr calls that don't have a string literal in them, for those a macro has to be declared that makes the string findable for the translation generator, see: https://doc.qt.io/qt-6/qml-qtqml-qt.html#QT_TRANSLATE_NOOP-method

PR #6355, opened 2025-11-27
[VDD] Add sorting

BruebachL:
No way to sort VDD

- Add combobox to sort
- Add labels to differentiate between sort and filter

## Screenshots
image
Sort by name (Dominaria set)
image
Sort by color (Dominaria set)
image
Sort by mana cost - Descending (Dominaria set)
image

### DIFF
--- a/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.cpp
@@ -90,9 +90,12 @@ VisualDatabaseDisplayWidget::VisualDatabaseDisplayWidget(QWidget *parent,
     sortColumnCombo = new QComboBox(this);
     sortColumnCombo->setSizeAdjustPolicy(QComboBox::SizeAdjustPolicy::AdjustToContents);
     sortOrderCombo = new QComboBox(this);
+    sortOrderCombo->setSizeAdjustPolicy(QComboBox::SizeAdjustPolicy::AdjustToContents);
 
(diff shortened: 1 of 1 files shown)


ebbit1q:
I quite like this and it's a feature I use a lot when using the list view, however it seems like there is a combination of my theme and the scaling I use (200%) that makes it scrunch up the words in the dropdown, is there a way to have the dropdown take a bit more space when needed?
image

one of the other things I like about the list editor is that the sorting of the headers is sequential, to sort by mv and then pt in that mv I can click the pt header first and then the mana value one which will layer their sortings (unfortunately this is not saved in the layout and I have to redo it on opening a deck editor), I don't think it's a feature the vde needs, just wanted to add a note

BruebachL:
> I quite like this and it's a feature I use a lot when using the list view, however it seems like there is a combination of my theme and the scaling I use (200%) that makes it scrunch up the words in the dropdown, is there a way to have the dropdown take a bit more space when needed? image
> 
> one of the other things I like about the list editor is that the sorting of the headers is sequential, to sort by mv and then pt in that mv I can click the pt header first and then the mana value one which will layer their sortings (unfortunately this is not saved in the layout and I have to redo it on opening a deck editor), I don't think it's a feature the vde needs, just wanted to add a note

I've adjusted, multi-preference sorting takes a bit more work.

ebbit1q:
as mentioned, I don't think we want multipreference sorting in the vde, the interface would not work well for it, and we still have the list for it (it'd be nice if the list allowed to sort by mv and not the mana as text though)

ebbit1q:
with this change only the sort type adjusts to contents, the sort order one also has this problem for me

PR #6356, opened 2025-11-27
[Game] Fix CardZoneLogic::clearContents()

BruebachL:
## Related Ticket(s)
- Fixes concede crashes

Manipulating lists while iterating them is problematic.

- Unstack manipulation and deletion for loops


ebbit1q (approved):
this crash is very hard to reproduce, but this method seems like a general improvement so I think merging this is good, we will see if it stops randomly occurring now

PR #6357, opened 2025-11-27
[VDE] Allow visual database display to toggle to classic table based display

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_visual_database_display.h
+++ b/cockatrice/src/interface/widgets/tabs/tab_visual_database_display.h
@@ -25,7 +25,8 @@ public:
     void retranslateUi() override;
     [[nodiscard]] QString getTabText() const override
     {
-        return tr("Visual Database Display");
+        return visualDatabaseDisplayWidget->displayModeButton->isChecked() ? tr("Database Display")
+                                                                           : tr("Visual Database Display");
     }
 };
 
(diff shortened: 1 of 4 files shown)


cockatrice/src/interface/widgets/tabs/tab_visual_database_display.h
```diff
@@ -25,7 +25,8 @@ class TabVisualDatabaseDisplay : public Tab
     void retranslateUi() override;
     [[nodiscard]] QString getTabText() const override
     {
-        return tr("Visual Database Display");
+        return visualDatabaseDisplayWidget->displayModeButton->isChecked() ? tr("Database Display")
```
ebbit1q, line 28: I just noticed this part, I'm not sure getTabText is used when the button is clicked, it always reads Visual Database Display and does not update when toggled, considering the state of the button is not saved in the settings there doesn't seem to be a way for the tab to display the other string (maybe when changing language?)

BruebachL: Yeah, I noticed this too, I really don't think there's a way to get a tab to re-evaluate its text short of setting the tab text for an index on the owning TabWidget and that's... involved. Still, guess it doesn't hurt to at least signal here.

ebbit1q: we can ticket it and move on, it's not a big deal

PR #6360, opened 2025-11-27
[VDE, VDD, EDHRec] Remember more card sizes

BruebachL:
We have all the means to remember the sizes users set for these cards but we don't because I was lazy once upon a time.

- Add settings for card sizes
- Initialize card size sliders with these settings
- hook up value changed signals on sliders to write settings


ebbit1q:
would it make sense to merge these settings maybe? I feel like if someone sets a scale for their visual deck display they might also want that that size in other places

BruebachL:
> would it make sense to merge these settings maybe? I feel like if someone sets a scale for their visual deck display they might also want that that size in other places

No, these are firmly independent settings. Consider that one widget might get exclusive access to the screen whereas another might only get a limited amount of space.

We should, all in all, move away from the concept of stuffing everything into cache_settings but that's a different topic.

PR #6364, opened 2025-11-28
[Printing Selector] Notify deck editor about history changes.

BruebachL:
Printing Selector does not notify deck editor about modifications so deck editor can't update the history widget.

- Add slots/signals/reasons.


ebbit1q:
I think the reason in the undo log should line up better with the reason when adding a card normally, they are functionally the same yet the message in the log is very different, the additional information in the log might be nice for a programmer but the average user might prefer the log to be more to the point

BruebachL:
I think it's valuable information, considering two cards might have identical set codes and collector numbers but different providerIds. It also explains the default behavior of replacing cards without providerIds

ebbit1q:
in that case why isn't that information there when adding the card from outside the printing selector?

ebbit1q:
unrelated: is the undo message for switching to the sideboard supposed to be "removed"

BruebachL:
> in that case why isn't that information there when adding the card from outside the printing selector?

I am iterating.

PR #6366, opened 2025-11-28
[DeckLoader] Refactor last load info into struct

RickyRister:

Consolidating the last load info into a struct will be helpful later when I want to store the last load info separately from the DeckLoader.

- Consolidate the last load info fields into a struct
- Update code to use struct

### DIFF
--- a/cockatrice/src/interface/deck_loader/deck_loader.h
+++ b/cockatrice/src/interface/deck_loader/deck_loader.h
@@ -69,7 +69,7 @@ public:
     DeckLoader(const DeckLoader &) = delete;
     DeckLoader &operator=(const DeckLoader &) = delete;
 
-    [[const LoadInfo &getLastLoadInfo() const
+    const LoadInfo &getLastLoadInfo() const
     {
         return lastLoadInfo;
     }


cockatrice/src/interface/deck_loader/deck_loader.h
```diff
@@ -28,6 +28,19 @@ inline Q_LOGGING_CATEGORY(DeckLoaderLog, "deck_loader")
...
+     * @brief Information about where the deck was loaded from.
+     *
+     * For local decks, the remoteDeckId field will always be -1.
+     * For remote decks, fileName will be empty and fileFormat will always be CockatriceFormat
+     */
+    struct LoadInfo
+    {
+        QString fileName = "";
+        FileFormat fileFormat = CockatriceFormat;
+        int remoteDeckId = -1;
```
BruebachL: Can we make this i.e. LOCAL_DECK instead of a magic number?

ebbit1q: -1 is just the value it was before too

BruebachL: I know but while we're at it, we may as well change this to be something a bit more declarative than just "-1". It's not immediately obvious that this signifies a local deck.

PR #6367, opened 2025-11-28
[Cleanup] Unused #includes

BruebachL:
Took 44 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_logs.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_logs.cpp
@@ -2,9 +2,17 @@
 
 #include "../interface/widgets/utility/custom_line_edit.h"
 
+#include <QCheckBox>
 #include <QDockWidget>
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QLabel>
 #include <QMessageBox>
+#include <QPushButton>
+#include <QRadioButton>
+#include <QSpinBox>
 #include <QTabWidget>
+#include <QTableWidget>
 #include <libcockatrice/network/client/abstract/abstract_client.h>
 #include <libcockatrice/protocol/pb/moderator_commands.pb.h>
 #include <libcockatrice/protocol/pb/response_viewlog_history.pb.h>
--- a/cockatrice/src/interface/widgets/tabs/tab_logs.h
+++ b/cockatrice/src/interface/widgets/tabs/tab_logs.h
@@ -9,18 +9,21 @@
 
 #include "tab.h"
 
-#include <QCheckBox>
 #include <QDialog>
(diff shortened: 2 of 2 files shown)


cockatrice/src/interface/widgets/dialogs/dlg_tip_of_the_day.h
```diff
@@ -7,20 +7,17 @@
...
 #include <QLineEdit>
 #include <QLoggingCategory>
-#include <QPushButton>
 #include <QVBoxLayout>
 
 inline Q_LOGGING_CATEGORY(DlgTipOfTheDayLog, "dlg_tip_of_the_day");
 
-class QLabel;
-class QPushButton;
-class QCheckBox;
```
ebbit1q, line 23: why remove these forwards declerations?

BruebachL: Why have them in the first place? They're basic Qt widgets.

ebbit1q: forward declarations are supposed to save time when including this header file multiple times, is there a compiler optimisation that is applied here that makes it preferable to just include them?

BruebachL: Ah, I see, I was under the impression that it *would* get optimized out if it detects that it's included multiple times across the project. The main reasoning here was that while it might save compilation time, we don't really do this anywhere else in the project (consistency) and I didn't estimate these basic widgets to be too demanding.

BruebachL: afaik fwd declarations are useful in case:
the included file is very large (these are basic Qt widgets)
the included file frequently changes and would thus cause recompilation for a lot of other headers (these are basic Qt widgets)

ebbit1q: I think we can always use more forwards declarations where possible, if these forwards declarations worked why remove them if they increased compilation speed by however little? if we'd be putting in effort to reduce compilation speed why would we do it?

BruebachL: I agree in general and will ticket more fwd declarations for myself in general.

ebbit1q: the real performance improvement would be gained if we were to use more forwards declarations on our own types too

ebbit1q (approved):
very nice, thank you for this!

PR #6368, opened 2025-11-28
ci: delete merged caches

brunoalr:
Github offers 'only' 10GiB of free GHA cache storage, which can be hit quite fast. To reduce cache thrashing we can delete caches of a PR that has been merged.

More info: [docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches#force-deleting-cache-entries](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches#force-deleting-cache-entries)

- GHA caches of a PR will be deleted after the PR is merged


ebbit1q:
afaik caches will automatically get deleted when new ones are created already, so we won't actually run out of space

brunoalr:
well, it is quite easy to go over the limit (see screenshot from https://github.com/Cockatrice/Cockatrice/actions/caches) 
image

github doesn't immediately evict the caches and when it does, it evicts the LRU ones first. this code is copied from a suggestion from github itself to give us a bit more control - the chances of needing to use caches from a merge branch are extremely low, so why wait to clean them? 
in fact, this PR comes from the link shown in the screenshot (https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches#force-deleting-cache-entries)

brunoalr:
for example:
https://github.com/Cockatrice/Cockatrice/actions/caches?query=branch:refs/pull/6366/merge+

and, oof, now see I another problem: I was not expecting to see multiple ccache caches per os-chip-mode... I will open an issue to track it: https://github.com/Cockatrice/Cockatrice/issues/6369

.github/workflows/delete-merged-caches.yml
```diff
@@ -0,0 +1,31 @@
...
+        run: |
+          echo "Fetching list of cache keys"
+          cacheKeysForPR=$(gh cache list --ref $BRANCH --limit 100 --json id --jq '.[].id')
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
+              gh cache delete $cacheKey
```
ebbit1q:
```suggestion
          echo "Deleting caches..."
          for cacheKey in $cacheKeysForPR
          do
              gh cache delete $cacheKey || true
```

this is simpler and explicitly shows that we ignore the return value of this command

.github/workflows/delete-merged-caches.yml
```diff
@@ -0,0 +1,31 @@
...
+          done
+          echo "Done"
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
```
ebbit1q: these two variables are already set for workflows

.github/workflows/delete-merged-caches.yml
```diff
@@ -0,0 +1,31 @@
...
+          BRANCH: refs/pull/${{ github.event.pull_request.number }}/merge
```
ebbit1q: on all of our other workflows we place env before run

brunoalr: done

.github/workflows/delete-merged-caches.yml
brunoalr: good point. not sure why the example set them

brunoalr: I tested by merging this branch to `master` of another fork that I have.
turns out the `gh` client needs at least `GH_TOKEN` (`GH_REPO` to be confirmed, but I guess it is actually needed).

image


https://github.com/bruno-test-org/Cockatrice/actions/runs/20618697016/job/59216317107#step:2:21

brunoalr: just checked. GH_REPO is also needed

brunoalr: GH_REPO is needed because this workflow does not checkout any repository

brunoalr: supporting doc regarding `GH_TOKEN` + `gh`: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli

PR #6370, opened 2025-11-28
[Fix-Warnings] Use auto instead of type

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/client/network/parsers/interface_json_deck_parser.h
+++ b/cockatrice/src/client/network/parsers/interface_json_deck_parser.h
@@ -24,7 +24,7 @@ class ArchidektJsonParser : public IJsonDeckParser
 public:
     DeckLoader *parse(const QJsonObject &obj) override
     {
-        DeckLoader *loader = new DeckLoader(nullptr);
+        auto *loader = new DeckLoader(nullptr);
 
         QString deckName = obj.value("name").toString();
         QString deckDescription = obj.value("description").toString();
@@ -59,7 +59,7 @@ class MoxfieldJsonParser : public IJsonDeckParser
 public:
     DeckLoader *parse(const QJsonObject &obj) override
     {
-        DeckLoader *loader = new DeckLoader(nullptr);
+        auto *loader = new DeckLoader(nullptr);
 
         QString deckName = obj.value("name").toString();
         QString deckDescription = obj.value("description").toString();
--- a/cockatrice/src/client/network/update/client/release_channel.cpp
+++ b/cockatrice/src/client/network/update/client/release_channel.cpp
@@ -151,7 +151,7 @@ void StableReleaseChannel::releaseListFinished()
     }
 
     QString shortHash = lastRelease->getCommitHash().left(GIT_SHORT_HASH_LEN);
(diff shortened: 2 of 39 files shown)


ebbit1q:
I don't consider the use of auto to be preferable even when the resulting type is obvious, use of either declaration is completely valid and I don't think we should enforce the use of auto when it is already explicitly typed

ZeldaZach:
I'm indifferent on this, it seems fine

BruebachL:
I'm just trying to get less squigly lines in my IDE so I can be alerted to actual issues (e.g. this warning often times hides "variable can be made const"). I am fine with adjusting our project-wide inspection level on this, this is simply the C++ default.

It's pretty ugly to me:

image

ebbit1q:
const local variables is another thing that is pointless and kind of a waste of time, a local that doesn't get assigned to can be const yes, but the entire reason to make variables const is to create an error in the compiler when accidentally asigning to a value, if you are later changing the code in a function with a local const value to now change that value you won't think twice about dropping its const qualifier, it might as well have never been there, const is useful for static values and class members but not locally inside functions

BruebachL:
You're right in that I don't really want to 'enforce' this but I don't see an issue with giving the codebase a 'once-over' spring-cleaning and then leaving it up to future developers to just drop the const at their convenience, as you've mentioned. The const warnings in particular are a constant source of yellow squiggles throughout the codebase.

We can also define an actual inspection profile tailored to our needs but I would ask you to take on that task since I have no strong opinions on most warnings and choose to adhere to whatever is suggested in most cases unless it causes issues.

The argument for this is that it *is* a signaling keyword and that code is read more than it is written (thus arguing in favor of readability)

ebbit1q:
I never see these warnings or yellow squigglies, as such they don't bother me, I don't know how I'd make a specific configuration for a compiler warning I never see but I can make some suggestions based on your recent prs on what I think of these specific warnings:
- use of auto in a "new" statement: ok, saves a few characters, reads well but not required
- use of auto instead of builtin types like int or bool: bad, doesn't even save characters and hides the fact that this is a very simple type
- use of auto in other places: sometimes reads easier like in iterations, not bad not amazing either, case by case, mostly just how lazy you feel
- local const: I can't think of a single case where setting a non static variable as const is useful, instead it adds more text to write but also read every time, it is redundant, if it is not clear what a variable is used for then it is wrongly named or scoped wrong
- extra semicolons: bad, they don't add anything useful in any case
- extra brackets in logical operations: sometimes good, && and & look similar, adding extra brackets helps separate these operators into separate groups allowing you to more easily read them, much preferred over this though is to split out the components of the logic comparison into separate values which increases readability significantly
- extra brackets elsewhere: often bad: they don't do anything and oftem might make things confusing where they add extra characters to read that are unnecessary

as you see many of these things are subjective and up to the programmer to determine on a case by case basis, a compiler warning might be useful sometimes but rarely as the improvement is so little and mostly subjective, I think many of these warnings should simply be disabled in any configuration by default and maybe looked at in new code but not existing working code.

PR #6372, opened 2025-11-29
[Fix-Warnings] Remove redundant parentheses

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

Some of these parentheses hide some interesting calculations:
image

- this
- and this

### DIFF
--- a/cockatrice/src/game/board/abstract_card_item.cpp
+++ b/cockatrice/src/game/board/abstract_card_item.cpp
@@ -288,9 +288,9 @@ void AbstractCardItem::setFaceDown(bool _facedown)
 
 void AbstractCardItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
 {
-    if ((event->modifiers() & Qt::AltModifier) && event->button() == Qt::LeftButton) {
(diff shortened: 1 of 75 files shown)


cockatrice/src/game/board/abstract_card_item.cpp
```diff
@@ -288,9 +288,9 @@ void AbstractCardItem::setFaceDown(bool _facedown)
 
 void AbstractCardItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
 {
-    if ((event->modifiers() & Qt::AltModifier) && event->button() == Qt::LeftButton) {
+    if (event->modifiers() & Qt::AltModifier && event->button() == Qt::LeftButton) {
```
ebbit1q, line 291: while yes, operator precedence is completely predictable, I'd argue that in cases where logic operations like && and || are used the brackets function as helpful separators that help read these statements even though they do nothing

ebbit1q:
```suggestion
    bool altPressed = event->modifiers() & Qt::AltModifier;
    if (altPressed && event->button() == Qt::LeftButton) {
```
like this the bracket is gone but the readability improvement is kept

PR #6373, opened 2025-11-29
[Fix-Warnings] Suppress C4100: unreferenced parameter for protobuf files

BruebachL:
These warnings are generated during protobuf compilation:
```
libcockatrice_protocol\libcockatrice\protocol\pb\event_notify_user.pb.cc(169): warning C4100: 'from_msg': unreferenced parameter
```

They are erroneous and can be safely suppressed so they don't clutter up our output.

- Suppress warning C4100 for libcockatrice_protocol_pb target

## Screenshots
image

### DIFF
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/CMakeLists.txt
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/CMakeLists.txt
@@ -165,11 +165,11 @@ set(PROTO_FILES
     session_event.proto
 )
 
-if (MSVC)
+if(MSVC)
   set(unused_warning /wd4100)
-else ()
+else()
   set(unused_warning -Wno-unused-parameter)
-endif ()
+endif()
 
 if(${Protobuf_VERSION} VERSION_LESS "3.21.0.0")
   message(STATUS "Using Protobuf Legacy Mode")


ebbit1q:
does this only suppress the warning for the generated files and nothing else? I'd say we can suppress all warnings for generated code as it is generated and there is nothing we can do to make it not make warnings but update the generator

BruebachL:
> does this only suppress the warning for the generated files and nothing else? I'd say we can suppress all warnings for generated code as it is generated and there is nothing we can do to make it not make warnings but update the generator

Yes, this only suppresses it for the libcockatrice_protocol_pb target, which is just the pb generation.

libcockatrice_protocol/libcockatrice/protocol/pb/CMakeLists.txt
```diff
@@ -189,6 +196,7 @@ if(${Protobuf_VERSION} VERSION_LESS "3.21.0.0")
   endif()
 else()
   add_library(libcockatrice_protocol_pb ${PROTO_FILES})
+  target_compile_options(libcockatrice_protocol_pb PRIVATE ${unused_warning})
```
ebbit1q, line 199: isn't this line identical to 181? why add the same line on both sides of the if statement instead of outside it? is there a sequence that matters?

BruebachL: The library doesn't exist until it's added with add_library afaik (and then it is compiled inside the respective if statements) but I am honestly the wrong person to ask about protobuf compilation.

ebbit1q: you're right, I don't really know either, and if it works it works that's kinda what I know about cmake

ebbit1q (approved):
thanks!

libcockatrice_protocol/libcockatrice/protocol/pb/CMakeLists.txt
```diff
@@ -165,13 +165,20 @@ set(PROTO_FILES
     session_event.proto
 )
 
+if(MSVC)
+  set(unused_warning /wd4100)
+else()
+  set(unused_warning -Wno-unused-parameter)
```
ebbit1q, line 171: I just noticed my compilation output for pb has a lot of -Wmissing-declarations instead, I still think we can simply disable -Wall entirely on the generated code.

PR #6376, opened 2025-11-29
[Fix-Warnings] Local variable can be made const

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


ebbit1q:
I'd prefer we suppress this warning at the compiler level instead, there is no real benefit to either performance or readability on this in my opinion.

ZeldaZach:
Conflicts + fails to build.

Also not certain this is fully necessary, but I won't stop it

PR #6378, opened 2025-11-29
[CardInfo] Display set short name and collector number in info widget.

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

"That's a cool printing, which one is it?"

- Display set shortname and collector number if PrintingInfo is not empty.

## Screenshots
image

### DIFF
--- a/cockatrice/src/interface/widgets/cards/card_info_text_widget.cpp
+++ b/cockatrice/src/interface/widgets/cards/card_info_text_widget.cpp
@@ -50,7 +50,7 @@ void CardInfoTextWidget::setTexts(const QString &propsText, const QString &textT
     textLabel->setText(textText);
 }
 
-void CardInfoTextWidget::setCard(ExactCard exactCard)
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/cards/card_info_text_widget.cpp
```diff
@@ -50,8 +50,9 @@ void CardInfoTextWidget::setTexts(const QString &propsText, const QString &textT
     textLabel->setText(textText);
 }
 
-void CardInfoTextWidget::setCard(CardInfoPtr card)
+void CardInfoTextWidget::setCard(ExactCard exactCard)
```
RickyRister: Can we not take `ExactCard` by reference here?

RickyRister (approved):
Seems like a reasonable addition.

Maybe in the future we can add the option to customize what info is listed in the info widget so it doesn't get too cramped.

PR #6380, opened 2025-11-30
[DeckList] Move metadata into struct

RickyRister:

I want to be able to store the metadata fields in `DeckList` separate from DeckList, since that will be helpful for parts of the DeckList refactors I'm planning. 

- Move metadata fields to `Metadata` struct nested inside `DeckList`
- Add a `isEmpty` method to `Metadata` which checks all fields except `lastLoadedTimestamp` and use that in `DeckList.isEmpty`. 
  - This *will* change program behavior since it previously checked that name and comment are empty. I do think this behavior is more correct though.
- Add a `isEmpty` method to `CardRef` for convenience

I'm still keeping around the setters and getters for the metadata fields in `DeckList` since the setters are slots, so updating usages for them will be more annoying than usual

### DIFF
--- a/libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
+++ b/libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
@@ -515,9 +515,7 @@ void DeckList::cleanList(bool preserveMetadata)
 {
     root->clearTree();
     if (!preserveMetadata) {
-        setName();
-        setComments();
-        setTags();
+        metadata = {};
     }
     refreshDeckHash();
 }


BruebachL:
I'm in favor of changing isEmpty() logic. I was considering some other parts too. With the addition of the deck history state manager, we can now pretty accurately predict if a deck has been modified or not: If the undo/redo stack is empty, it's not modified, if it has entries, it is modified.

libcockatrice_deck_list/libcockatrice/deck_list/deck_list.h
```diff
@@ -223,32 +235,38 @@ public slots:
     ~DeckList() override;
 
     /// @name Metadata getters
+    /// The individual metadata getters still exist for backwards compatibility.
+    /// TODO: Figure out when we can remove them.
```
ebbit1q, line 239: we can remove them as soon as no functions use them anymore, I don't see any reason not to

PR #6381, opened 2025-11-30
[DeckList] Move decklist node classes into new folder

RickyRister:

A lot of classes in the `deck_list` folder are for representing the decklist tree structure. 

Now that we are starting to create more classes in `deck_list`, I think it would be better for readability if the node classes are separated from the other classes.

- Created `tree` folder in `deck_list`
  - Moved the four node classes into it
- Update includes 
- Update cmake


ebbit1q (approved):
seems good, thanks for fixing some includes too!

PR #6383, opened 2025-11-30
[DeckList] Make DeckList not a QObject

RickyRister:

Making `DeckList` no longer a QObject means we no longer have to deal with qt's ownership model when passing that class around between objects.
This is another step in the DeckList refactor.

- `DeckList` no longer inherits from QObject
  - Remove all signals
    - `DeckList` only had two signals, so it wasn't too bad to just find all places it was used and manually call the other end of the connection.
  - Remove the reparenting
    - Which happens once in `DeckListModel` because everything else is using `DeckLoader`
- Add copy constructor to `DeckList`
  - Deep copies the tree
- Delete the unnecessary virtual methods in `DeckList` that just returned the argument.

`DeckListModel` creates an empty `DeckList` in its constructor, which means it now leaks that object. (The current code actually already leaks the `DeckList` from `cleanList` since it never reparents it)
I'm planning on fixing that in a future PR.

I'm planning on updating the usages of that class in future PRs too.

### DIFF
--- a/libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
+++ b/libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
@@ -449,11 +449,8 @@ bool DeckList::loadFromStream_Plain(QTextStream &in, bool preserveMetadata)
             cardName.replace(diff.key(), diff.value());
         }
 
-        // Resolve complete card name, this function does nothing if the name is not found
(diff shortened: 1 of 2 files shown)


ZeldaZach (approved):

BruebachL:
LGTM. Objects are objects, models are qobjects and have signals they emit on modifying objects

PR #6388, opened 2025-11-30
[DeckEditor] Expand DeckDock TreeView when adding card through deck editor

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


RickyRister (approved):
Still not fixed for moving cards between main/side, but that can be done in a separate PR.

https://github.com/user-attachments/assets/1ec8ed64-68ae-46a7-9060-ceacfd2ebc91

Thanks for the fix.

PR #6400, opened 2025-12-05
Fix flipped svg for donator/judge/vip

### LINKED ISSUE #6290: Flipped Moderator Icons
Version: 2.11.0 beta 32

Bug: The icon for VIP Moderator (used for Head Moderator) is flipped horizontally, and does not match the rest of moderator icons. See image below.

Image

lilyhuang-github:
## Related Ticket(s)
- Fixes #6290 

There's an issue with the moderator icons where the black and white are flipped if they're a vip or a judge or a donator.

- Flip the id's of left and right over for the offending svgs to make it correct.

## Screenshots
Previously:
image
image
After:
image
Don't have an image with a vip/donator but since it works for the judge, it should work for others.


ebbit1q:
another curiosity: #6292 fixed this issue for the vip icons but does something else in the svg entirely...

lilyhuang-github:
> another curiosity: #6292 fixed this issue for the vip icons but does something else in the svg entirely...

It flipped the star for the vip icon but didn't fix the issue judging by the comment mentioned [here](https://github.com/Cockatrice/Cockatrice/issues/6290#issuecomment-3539061462)

PR #6401, opened 2025-12-06
[App] First-run tutorial

### LINKED ISSUE #6402: First-run Tutorial
We don't explain anything to the user, relying on a combination of intuition, tooltips, labels, wiki articles, video tutorials and discord support channels to help users understand our software.

Modern software uses a 'first run' tutorial system where the UI is darkened and relevant components are highlighted.

BruebachL:
## Related Ticket(s)
- Fixes #6402
- Related to https://github.com/Cockatrice/Cockatrice/milestone/11 but no ticket exists

We don't explain anything to the user, relying on a combination of intuition, tooltips, labels, wiki articles, video tutorials and discord support channels to help users understand our software.

Modern software uses a 'first run' tutorial system where the UI is darkened and relevant components are highlighted.

- Introduce TutorialController and TutorialOverlay for a semi-interactive tutorial

Example useage:

```
TutorializedWidget::TutorializedWidget(QWidget *_parent) : QWidget(_parent) 
{
    tutorialController = new TutorialController(this); // This is the widget to darken
    tutorialController->addStep({tabContainer->visualDeckView, "View your deck here."});
    tutorialControllerl->addStep({printingSelectorDockWidget, "Change the printings in your deck here."});
    tutorialController->addStep({tabContainer->visualDatabaseDisplay, "View the database here",
                         [this]() { tabContainer->setCurrentWidget(tabContainer->visualDatabaseDisplay); }}); // Optional onEnter() callback for this
}

void TutorializedWidget::showEvent(QShowEvent *event)
{
    QWidget::showEvent(event);
    if (!tutorialStarted) {
        tutorialStarted = true;
        // Start on next event loop iteration so everything is fully painted
        QTimer::singleShot(0, tutorialController, [this] { tutorialController->start(); });
    }
}
```

## Screenshots

https://github.com/user-attachments/assets/9247b6bd-78f3-4236-a957-8fad65c5919b

### DIFF
--- a/cockatrice/src/game/deckview/deck_view_container.cpp
+++ b/cockatrice/src/game/deckview/deck_view_container.cpp
@@ -99,6 +99,21 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
             &DeckViewContainer::setVisualDeckStorageExists);
 
     switchToDeckSelectView();
+    generateTutorialSequence();
+}
+
+TutorialSequence DeckViewContainer::generateTutorialSequence()
+{
+    TutorialSequence deckViewContainerSequence;
+    deckViewContainerSequence.name = tr("Loading and selecting decks");
+
+    deckViewContainerSequence.addStep(
+        {this, tr("There are multiple ways to select a deck:\n\n- From a local file"
+                  "\n- From the contents of your clipboard\nFrom an external online service")});
+
+    deckViewContainerSequence = visualDeckStorageWidget->generateTutorialSequence(deckViewContainerSequence);
+
+    return deckViewContainerSequence;
 }
 
 /**
--- a/cockatrice/src/interface/widgets/general/home_widget.h
+++ b/cockatrice/src/interface/widgets/general/home_widget.h
@@ -24,9 +24,18 @@ public:
     HomeWidget(QWidget *parent, TabSupervisor *tabSupervisor);
     void updateRandomCard();
     QPair<QColor, QColor> extractDominantColors(const QPixmap &pixmap);
+    HomeStyledButton *connectButton;
+    HomeStyledButton *visualDeckEditorButton;
+    HomeStyledButton *visualDeckStorageButton;
+    HomeStyledButton *visualDatabaseDisplayButton;
+    HomeStyledButton *edhrecButton;
+    HomeStyledButton *archidektButton;
+    HomeStyledButton *replaybutton;
+    HomeStyledButton *exitButton;
 
 public slots:
     void paintEvent(QPaintEvent *event) override;
+    void showEvent(QShowEvent *event) override;
     void initializeBackgroundFromSource();
     void onBackgroundShuffleFrequencyChanged();
     void updateBackgroundProperties();
@@ -39,11 +48,12 @@ private:
     QTimer *cardChangeTimer;
     TabSupervisor *tabSupervisor;
     QPixmap background;
+    TutorialController *tutorialController;
+    bool tutorialStarted = false;
     CardInfoPictureArtCropWidget *backgroundSourceCard = nullptr;
     DeckList backgroundSourceDeck;
     QPixmap overlay;
     QPair<QColor, QColor> gradientColors;
-    HomeStyledButton *connectButton;
 
     void setRandomCard(ExactCard &newCard);
     void loadBackgroundSourceDeck();
--- a/cockatrice/src/interface/widgets/general/home_widget.h
+++ b/cockatrice/src/interface/widgets/general/home_widget.h
@@ -15,6 +15,8 @@
 #include <QWidget>
 #include <libcockatrice/network/client/abstract/abstract_client.h>
 
+class TutorialController;
+
 class HomeWidget : public QWidget
 {
 
--- a/cockatrice/src/interface/widgets/general/home_widget.h
+++ b/cockatrice/src/interface/widgets/general/home_widget.h
@@ -26,14 +26,6 @@ public:
     HomeWidget(QWidget *parent, TabSupervisor *tabSupervisor);
     void updateRandomCard();
     QPair<QColor, QColor> extractDominantColors(const QPixmap &pixmap);
-    HomeStyledButton *connectButton;
-    HomeStyledButton *visualDeckEditorButton;
-    HomeStyledButton *visualDeckStorageButton;
-    HomeStyledButton *visualDatabaseDisplayButton;
-    HomeStyledButton *edhrecButton;
-    HomeStyledButton *archidektButton;
-    HomeStyledButton *replaybutton;
-    HomeStyledButton *exitButton;
 
 public slots:
     void paintEvent(QPaintEvent *event) override;
@@ -56,6 +48,15 @@ private:
     DeckList backgroundSourceDeck;
     QPixmap overlay;
     QPair<QColor, QColor> gradientColors;
+
+    HomeStyledButton *connectButton;
+    HomeStyledButton *visualDeckEditorButton;
+    HomeStyledButton *visualDeckStorageButton;
+    HomeStyledButton *visualDatabaseDisplayButton;
+    HomeStyledButton *edhrecButton;
+    HomeStyledButton *archidektButton;
+    HomeStyledButton *replaybutton;
+    HomeStyledButton *exitButton;
 
     void setRandomCard(ExactCard &newCard);
     void loadBackgroundSourceDeck();
--- a/cockatrice/src/interface/widgets/general/tutorial/tutorial_controller.cpp
+++ b/cockatrice/src/interface/widgets/general/tutorial/tutorial_controller.cpp
@@ -1,0 +1,374 @@
+#include "tutorial_controller.h"
+
+#include <QComboBox>
+#include <QDebug>
+#include <QLineEdit>
+#include <QMainWindow>
+#include <QPlainTextEdit>
+#include <QTextEdit>
+#include <QTimer>
+
+TutorialController::TutorialController(QWidget *_tutorializedWidget)
+    : QObject(_tutorializedWidget), tutorializedWidget(_tutorializedWidget)
+{
+    tutorialOverlay = new TutorialOverlay(tutorializedWidget->window());
+
+    tutorialOverlay->setWindowFlags(tutorialOverlay->windowFlags() | Qt::FramelessWindowHint);
+    tutorialOverlay->hide();
+
+    connect(tutorialOverlay, &TutorialOverlay::nextStep, this, &TutorialController::attemptAdvance);
+    connect(tutorialOverlay, &TutorialOverlay::prevStep, this, &TutorialController::prevStep);
+    connect(tutorialOverlay, &TutorialOverlay::nextSequence, this, &TutorialController::nextSequence);
+    connect(tutorialOverlay, &TutorialOverlay::prevSequence, this, &TutorialController::prevSequence);
+    connect(tutorialOverlay, &TutorialOverlay::skipTutorial, this, &TutorialController::exitTutorial);
+    connect(tutorialOverlay, &TutorialOverlay::targetClicked, this, &TutorialController::handleTargetClicked);
+}
+
+void TutorialController::addSequence(const TutorialSequence &seq)
+{
+    sequences.append(seq);
+}
+
+void TutorialController::start()
+{
+    if (sequences.isEmpty() || tutorialCompleted) {
+        return;
+    }
+
+    QTimer::singleShot(0, this, [this]() {
+        QWidget *win = tutorializedWidget->window();
+
+        // Reparent to make absolutely sure
+        tutorialOverlay->setParent(win);
+        tutorialOverlay->setGeometry(0, 0, win->width(), win->height());
+
+        // Stack order
+        tutorialOverlay->stackUnder(nullptr);
+        tutorialOverlay->show();
+        tutorialOverlay->raise();
+
+        currentSequence = 0;
+        currentStep = 0;
+        showStep();
+    });
+}
+
+void TutorialController::handleTargetClicked()
+{
+    if (currentSequence < 0 || currentStep < 0) {
+        return;
+    }
+
+    const auto &step = sequences[currentSequence].steps[currentStep];
+
+    // If this step requires interaction AND uses OnAdvance validation, advance when clicked
+    // For OnSignal/OnChange, the click just triggers the action - validation happens via signal
+    if (step.requiresInteraction && step.validationTiming == ValidationTiming::OnAdvance) {
+        attemptAdvance();
+    }
+}
+
+void TutorialController::attemptAdvance()
+{
+    if (currentSequence < 0 || currentStep < 0) {
+        return;
+    }
+
+    const auto &step = sequences[currentSequence].steps[currentStep];
+
+    // Only validate on advance if timing is set to OnAdvance
+    if (step.validationTiming == ValidationTiming::OnAdvance) {
+        if (!validateCurrentStep()) {
+            return; // Validation failed, stay on current step
+        }
+    }
+
+    // Validation passed or not required, proceed to next step
+    nextStep();
+}
+
+bool TutorialController::validateCurrentStep()
+{
+    if (currentSequence < 0 || currentSequence >= sequences.size()) {
+        return true; // No validation needed
+    }
+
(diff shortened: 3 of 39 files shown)


ebbit1q:
this seems awesome, am I correct in that this does not yet actually use the tutorial system? I only see the files being added, no usage.

I'd like to see this pr immediately include a minimal tutorial for the deck editor and printing selector, I think it should trigger immediately on first run of the deck editor.

BruebachL:
> this seems awesome, am I correct in that this does not yet actually use the tutorial system? I only see the files being added, no usage.
> 
> I'd like to see this pr immediately include a minimal tutorial for the deck editor and printing selector, I think it should trigger immediately on first run of the deck editor.

You're correct, this PR just provides the base functionality of the system, I am hoping for some help on actually coming up with a decent tutorial.

I have some tutorials pre-written that I will gladly PR once this is merged. I just think it's a bit nicer to separate the "first-run" logic from the tutorial system itself, but that's just me.

We'll likely also need to tweak this. Right now it advances when you click on a bubble, which isn't terribly intuitive and it quits when you click anywhere but the highlighted region or the bubble, which, again, isn't very intuitive, but it's a nice PoC in this state that we can expand on.

We should also maybe consider Sequences instead of just Steps so we can skip certain parts or everything, etc.

ebbit1q:
I just want something to be there so I can actually "review" the added changes, otherwise I can't really see what it looks like

ebbit1q:
as part of reviewing this pr I am exploring the full first run experience, while this doesn't affect this pr I want to make a note here on how it adds to the issue on first run where the user is overwhelmed with dialogs:
image

ebbit1q:
I will review the experience only, I will leave actually reviewing the content of the tutorial for later:
- the icons for moving through the tutorial are not very clear, I'd prefer them to be more identifiable as buttons by adding a frame around them or making them a bit thicker, perhaps adding a tooltip, alternatively the first part of the tutorial should explain what the buttons do (might be helpful regardless, some users might not realize this is an optional tutorial)
- the position of the buttons is not fixed even when the dialog stays in the same location, this is annoying when the tutorial stays focused on the same element for multiple dialogs and I want to skim through them by clicking, instead I have to hunt the rather small next button each time, to fix this I propose moving the controls to the top left of the dialog and making them a bit larger
- adding a title to the dialog, eg tutorial, or a label for the chapter, might help identify to users what this system does
- resizing the window during a highlight or moving it does not move it with it, it is clearly a separate window and is only put in the right location when it is opened
- in addition, the tutorial does not stop users from switching to a different tab, and sticks around and plays like normal even if the relevant tab is not in focus
- what is interesting is how the little dialog with the text inside this window does actually reorient itself, but only when I hover over it
- it's not made clear to the user when the tutorial "pressed a button" for them, eg when switching to the db tab it is not made clear that the user should click the tab to go here, this might be solved in the tutorial itself by adding an extra step highlighting the tab
- it would be nice to have a little end dialog that tells users the tutorial is finished, allowing users to click back to a previous chapter in case they missed something, this last dialog should not have forwards buttons but instead have like a "finish" button
- similarly it'd be nice to have a start dialog that doesn't block anything yet that doesn't have any "go back" buttons
- you mentioned in discord that the tutorial might ask users to interact with the client at some point, how will it do that if it overlays the entire thing?
- currently the tutorial opens every time and does not seem to have a way to remember the user has seen it already

ebbit1q:
I have ticketed the source of one of the first run dialogs here: #6419

BruebachL:
> I will review the experience only, I will leave actually reviewing the content of the tutorial for later:
> 
>     * the icons for moving through the tutorial are not very clear, I'd prefer them to be more identifiable as buttons by adding a frame around them or making them a bit thicker, perhaps adding a tooltip, alternatively the first part of the tutorial should explain what the buttons do (might be helpful regardless, some users might not realize this is an optional tutorial)
> 
>     * the position of the buttons is not fixed even when the dialog stays in the same location, this is annoying when the tutorial stays focused on the same element for multiple dialogs and I want to skim through them by clicking, instead I have to hunt the rather small next button each time, to fix this I propose moving the controls to the top left of the dialog and making them a bit larger

I've adjusted how the buttons look/behave (We'll have to find proper icons at some point)

image

>     * resizing the window during a highlight or moving it does not move it with it, it is clearly a separate window and is only put in the right location when it is opened
>     * in addition, the tutorial does not stop users from switching to a different tab, and sticks around and plays like normal even if the relevant tab is not in focus

It now hides itself and should properly size itself in accordance with its parent

>     * you mentioned in discord that the tutorial might ask users to interact with the client at some point, how will it do that if it overlays the entire thing?
You can have it be transparent for mouse clicks and allow mouse clicks in the highlight hole.

TODO
>     * currently the tutorial opens every time and does not seem to have a way to remember the user has seen it already
>     * adding a title to the dialog, eg tutorial, or a label for the chapter, might help identify to users what this system does
>     * it would be nice to have a little end dialog that tells users the tutorial is finished, allowing users to click back to a previous chapter in case they missed something, this last dialog should not have forwards buttons but instead have like a "finish" button
>     * similarly it'd be nice to have a start dialog that doesn't block anything yet that doesn't have any "go back" buttons
>     * it's not made clear to the user when the tutorial "pressed a button" for them, eg when switching to the db tab it is not made clear that the user should click the tab to go here, this might be solved in the tutorial itself by adding an extra step highlighting the tab

BruebachL:
Done:
> > * it's not made clear to the user when the tutorial "pressed a button" for them, eg when switching to the db tab it is not made clear that the user should click the tab to go here, this might be solved in the tutorial itself by adding an extra step highlighting the tab

This is unfortunately not very easy to do. Due to the way the Qt Tab Bar is implemented, the tabs themselves aren't proper widgets. They are instead 'virtual' widgets, i.e. they only exist in the TabBars paintEvent, which then does hit detection to see which tab is intended. This means we can't easily select them. The best we can do is highlight the tab bar and be explicit in the text, which I've done.

> > * adding a title to the dialog, eg tutorial, or a label for the chapter, might help identify to users what this system does

Done, title and steps are now displayed.

> TODO
> 
> > ```
> > * currently the tutorial opens every time and does not seem to have a way to remember the user has seen it already
> > * it would be nice to have a little end dialog that tells users the tutorial is finished, allowing users to click back to a previous chapter in case they missed something, this last dialog should not have forwards buttons but instead have like a "finish" button
> > * similarly it'd be nice to have a start dialog that doesn't block anything yet that doesn't have any "go back" buttons
> > ```

cockatrice/src/interface/widgets/tabs/tab_game.h
```diff
@@ -176,6 +181,8 @@ protected slots:
             QList<AbstractClient *> &_clients,
             const Event_GameJoined &event,
             const QMap<int, QString> &_roomGameTypes);
+    void showEvent(QShowEvent *event);
```
RickyRister: `error: 'showEvent' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]`

RickyRister:
Cockatrice crashes if you double click to resize the window after playing around for a bit

https://github.com/user-attachments/assets/70b1f8c7-aa8c-49e2-812c-483e9a483a47

RickyRister:
Tried out some of the tutorials. Seems really cool so far, despite how janky everything is.

One concern I have with the implementation is that, from what I can tell, the tutorial needs to be implemented inside the widget. I'm worried that adding tutorials to the classes will bloat the files. Now the classes needs to contain both the functionality implementation as well as the tutorial.

If there's no good way to make it so that it's possible to implement tutorials from outside the widget, then I suppose it's fine if we just always isolate the tutorial creation to some method at the bottom of the file.

RickyRister:
The VDE tutorial doesn't seem to handle the window being resized well. In my case, the tutorial buttons all disappear and now I'm softlocked because my clicks are still disabled.

https://github.com/user-attachments/assets/e4a34a89-d251-4cea-8177-5c83d80ed297

cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -129,6 +130,193 @@ TabGame::TabGame(TabSupervisor *_tabSupervisor,
...
+        {gamePlayAreaWidget,
+         tr("Let's go over them quickly, left-to-right.\n\nThe phase toolbar\nThe player area\nThe battlefield")});
+    gamePlaySequence.addStep(
+        {gamePlayAreaWidget,
+         tr("First up, is the phase toolbar. This toolbar shows the current phase of the turn. You "
+            "can advance it by pressing\n\n- Tab (simply advances the phase)\n- Ctrl+Space (advances "
+            "the phase and performes any associated actions)\n- Clicking directly on the phase you "
+            "want to change to.\n\nYou can also pass the turn here, although, you should note that "
+            "most players prefer you simply leave your turn on the end step and allow them to "
+            "'take' the turn from you by pressing 'Next turn' themselves.")});
```
RickyRister: nit: For these big multiline strings, I prefer if the line breaks in the source lines up with the line breaks of the resulting string. Makes it easier to read at a glance. Usually, I just insert the manual line breaks after `\n`s, then let autoformatter insert any extra line breaks for lines that are too long.

For example,
```suggestion
         tr("First up, is the phase toolbar. This toolbar shows the current phase of the turn. You can advance it by "
            "pressing\n\n"
            "- Tab (simply advances the phase)\n"
            "- Ctrl+Space (advances the phase and performes any associated actions)\n"
            "- Clicking directly on the phase you want to change to.\n\n"
            "You can also pass the turn here, although, you should note that most players prefer you simply leave your "
            "turn on the end step and allow them to 'take' the turn from you by pressing 'Next turn' themselves.")});
```

cockatrice/src/interface/widgets/general/tutorial/tutorial_controller.h
```diff
@@ -0,0 +1,99 @@
...
+
+    // Signal-based validation (for ValidationTiming::OnSignal)
+    QObject *signalSource = nullptr;  // Object that emits the signal
+    const char *signalName = nullptr; // Signal to connect to (use SIGNAL() macro)
+};
+
+struct TutorialSequence
+{
+    QString name;
+    QVector<TutorialStep> steps;
```
RickyRister: In [Qt6](https://doc.qt.io/qt-6/qvector.html), `QVector` was turned into an alias for `QList` and mostly deprecated. The documentation suggests using `QList` over `QVector`

ebbit1q: we build for both qt6 and qt5 still

cockatrice/src/interface/widgets/general/tutorial/tutorial_controller.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if (currentStep < 0 || currentStep >= sequences[currentSequence].steps.size()) {
+        return;
+    }
+
+    const auto &step = sequences[currentSequence].steps[currentStep];
+
+    // Handle OnSignal validation - connect to any custom signal
+    if (step.validationTiming == ValidationTiming::OnSignal && step.validator) {
+        if (step.signalSource && step.signalName) {
+            qInfo() << "Setting up signal-based validation for signal:" << step.signalName;
```
RickyRister: Remember to change this to use qCInfo once we finalize it

cockatrice/src/interface/widgets/general/tutorial/tutorial_controller.h
```diff
@@ -0,0 +1,100 @@
...
+
+    // Auto-advance when validation passes (useful for text input)
+    bool autoAdvanceOnValid = false;
+
+    // Custom interaction hint (overrides default "Click to continue")
+    QString customInteractionHint = nullptr;
+
+    // Signal-based validation (for ValidationTiming::OnSignal)
+    QObject *signalSource = nullptr;  // Object that emits the signal
+    const char *signalName = nullptr; // Signal to connect to (use SIGNAL() macro)
```
RickyRister: Are you sure there isn't a more typesafe way to do this? This will break on refactoring, and no one will notice because it'll just fail silently and I doubt any of us is going to run the tutorial from start-to-finish on every change.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
```diff
@@ -88,6 +89,30 @@ VisualDeckStorageWidget::VisualDeckStorageWidget(QWidget *parent) : QWidget(pare
     }
 }
 
+TutorialSequence VisualDeckStorageWidget::generateTutorialSequence(TutorialSequence vdsSequence)
```
RickyRister: makes more sense for this to be a void method that takes a `TutorialSequence&`

cockatrice/src/game/deckview/deck_view_container.cpp
```diff
@@ -99,6 +99,21 @@ DeckViewContainer::DeckViewContainer(int _playerId, TabGame *parent)
...
+TutorialSequence DeckViewContainer::generateTutorialSequence()
+{
+    TutorialSequence deckViewContainerSequence;
+    deckViewContainerSequence.name = tr("Loading and selecting decks");
+
+    deckViewContainerSequence.addStep(
+        {this, tr("There are multiple ways to select a deck:\n\n- From a local file"
+                  "\n- From the contents of your clipboard\nFrom an external online service")});
+
+    deckViewContainerSequence = visualDeckStorageWidget->generateTutorialSequence(deckViewContainerSequence);
```
RickyRister: alternatively, maybe add some sort of `addSteps(const TutorialSequence &)` method to `TutorialSequence` so that it's easier to combine sequences.

cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.cpp
```diff
@@ -52,6 +54,83 @@ TabDeckEditorVisual::TabDeckEditorVisual(TabSupervisor *_tabSupervisor) : Abstra
     loadLayout();
     filterDockWidget->setHidden(true);
     cardDatabaseDockWidget->setHidden(true);
+    tutorialController = new TutorialController(this);
+
+    auto deckDockSequence = deckDockWidget->generateTutorialSequence();
```
RickyRister, line 68: is it possible to move the TutorialSequence creation in this class to its own method? I prefer if the tutorial stuff isn't tangled with the rest of the functional stuff.

RickyRister:
Things I like:
- The tutorial is encapsulated in a `TutorialSequence`, which means all the tutorial stuff can shunted off to a method that returns a `TutorialSequence`, instead of being tangled up with the rest of the class.

Things I don't like:
- Some of steps reach really deeply into the internals of the classes, which means it's prone to breaking on refactoring or impeding refactoring.
  - In particular, this majorly discourages structural redesigns, as that will require the tutorial to be completely redone.
- This adds additional friction when making changes to the components encompassed by the tutorial, as any change might require also fixing the tutorial.
- `TutorialSequence` isn't composable with itself, which means the top level `TutorialController` still ends up having to reach into its children to extract the sequences.
  - Could be solved by making the `generateTutorialSequence` methods return lists and having the `TutorialController` accept lists

cockatrice/src/interface/widgets/general/tutorial/tutorial_controller.h
```diff
@@ -0,0 +1,100 @@
...
+{
+    OnAdvance, // Validate when user clicks next/clicks target (default)
+    OnChange,  // Validate whenever target widget changes (for text input)
+    OnSignal,  // Validate when a specific signal is emitted
+    Manual     // Only validate when explicitly triggered
+};
+
+struct TutorialStep
+{
+    QWidget *targetWidget = nullptr;
```
RickyRister: Since we're holding pointers to QObjects inside structs (which are not managed by the QObject tree), and we store the pointers a while before we actually use them, is it possible for us to end up with a dangling pointer?

cockatrice/src/interface/widgets/general/home_widget.h
```diff
@@ -24,9 +24,18 @@ class HomeWidget : public QWidget
     HomeWidget(QWidget *parent, TabSupervisor *tabSupervisor);
     void updateRandomCard();
     QPair<QColor, QColor> extractDominantColors(const QPixmap &pixmap);
+    HomeStyledButton *connectButton;
```
RickyRister: why are the buttons public?

PR #6405, opened 2025-12-07
[EDHRec] New layout for commander details

BruebachL:
Ugly.

- Pretty.
- Incorporate the F rule (Users scan left-to-right, top-to-bottom for most important controls/information)

## Screenshots

Before:
image

After:
image

After:
image

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/api/edhrec/display/commander/edhrec_commander_api_response_bracket_navigation_widget.cpp
+++ b/cockatrice/src/interface/widgets/tabs/api/edhrec/display/commander/edhrec_commander_api_response_bracket_navigation_widget.cpp
@@ -1,5 +1,7 @@
 #include "edhrec_commander_api_response_bracket_navigation_widget.h"
 
+#include <QSet>
+
 EdhrecCommanderApiResponseBracketNavigationWidget::EdhrecCommanderApiResponseBracketNavigationWidget(
     QWidget *parent,
     const QString &baseUrl)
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/tabs/api/edhrec/display/cards/edhrec_api_response_card_details_display_widget.cpp
```diff
@@ -19,17 +19,17 @@ EdhrecApiResponseCardDetailsDisplayWidget::EdhrecApiResponseCardDetailsDisplayWi
     nameLabel = new QLabel(this);
     nameLabel->setText(toDisplay.name);
     nameLabel->setAlignment(Qt::AlignHCenter);
+    nameLabel->setStyleSheet("font-size: 20px; font-weight: bold");
```
ebbit1q, line 22: does setting these here mean they can't be themed anymore?

BruebachL: Unfortunately, yes, *however* there's no real elegant way of doing this for the pushButtons that I'm aware of (it's fine for the labels but Qt is weird about buttons).

This leaves us with three options:
1. Define them in a default theme stylesheet
2. Define them programmatically (as seen here)
3. Leave them undefined and unstyled and rely on user themes to style this (Many are outdated/unmaintained, DarkMingo is officially unmaintained [though it does still see fixes])

I don't consider option 3 as valid and would like to trial this layout in the beta/until we figure out a default stylesheet.

ebbit1q: sure

PR #6406, opened 2025-12-07
[DeckLoader] Extract LoadedDeck struct

RickyRister:

I want to decouple the representation of a loaded deck from the `DeckLoader` object. 
`DeckLoader` is a `QObject` which makes it annoying to pass around. It should only be responsible for loading the deck, and should not also be responsible holding the result of the deck loading once the loading is done.

- Create a `LoadedDeck` struct that contains deckList + lastLoadInfo
  - This is so that we can pass around that info without having to pass around a `DeckLoader` `QObject`.
- Moved `FileFormat` and `LoadInfo` into `LoadedDeck`.
  - As well as `getFormatFromName`

This PR **does not** use the new `LoadedDeck` struct. That will be done in a follow-up PR.

### DIFF
--- a/cockatrice/src/interface/deck_loader/loaded_deck.cpp
+++ b/cockatrice/src/interface/deck_loader/loaded_deck.cpp
@@ -1,6 +1,5 @@
 #include "loaded_deck.h"
 
-
 bool LoadedDeck::LoadInfo::isEmpty() const
 {
     return fileName.isEmpty() && remoteDeckId == NON_REMOTE_ID;


BruebachL:
I agree with the stated intent: DeckLoaders are transient objects that do something for us but should not hold state.

Can we move the FileFormat enums into something more declaratively namespaced?
CockatriceDeckFormats:: perhaps?

RickyRister:
`CockatriceDeckFormat::CockatriceFormat` is a mouthful. I'll just name the namespace `DeckFileFormat` and the enum `Format`

ebbit1q (approved):
seems good, thanks

BruebachL:
> `CockatriceDeckFormat::CockatriceFormat` is a mouthful. I'll just name the namespace `DeckFileFormat` and the enum `Format`

Yeah, I like that. Just something to make it a bit more re-useable outside of LoadedDeck context.

PR #5026, opened 2024-04-29
Hashing tests

ebbit1q:
## Related Ticket(s)
- Related #5025

## Short roundup of the initial problem
There are no tests that verify the deck hasher remains the same.

## What will change with this Pull Request?
- adds tests
- tries to improve performance

PR #6407, opened 2025-12-08
[DeckList] Refactor and cleanup methods that iterate over nodes

RickyRister:

Some of the code that iterates over all card nodes in the decklist could be cleaned up and simplified more.
We should use the methods on `DeckList` over directly getting and manually iterating over the root node.

Reducing direct access to the decklist tree node will also reduce the amount of changes I need to make for the bigger DeckList refactor I'm still working on.

- get rid of two useless static helper methods in `DeckList`
- Add `getZoneNodes` method to `DeckList`
- Refactor places where manual iteration over the node tree could be replaced by just using `getCardNodes` or similar.

### DIFF
--- a/libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
+++ b/libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
@@ -656,15 +656,17 @@ static QString computeDeckHash(const DeckList &deckList)
     auto mainDeckNodes = deckList.getCardNodes({DECK_ZONE_MAIN});
     auto sideDeckNodes = deckList.getCardNodes({DECK_ZONE_SIDE});
 
-    QStringList cardList;
-
-    static auto repeatName = [](const DecklistCardNode *node, const QString &prefix = {}) {
-        return (prefix + node->getName().toLower()).repeated(node->getNumber());
+    static auto nodesToCardList = [](const QList<const DecklistCardNode *> &nodes, const QString &prefix = {}) {
+        QStringList result;
+        for (auto node : nodes) {
+            for (int i = 0; i < node->getNumber(); ++i) {
+                result.append(prefix + node->getName().toLower());
+            }
+        }
+        return result;
     };
-    std::transform(mainDeckNodes.cbegin(), mainDeckNodes.cend(), std::back_inserter(cardList),
-                   [](auto node) { return repeatName(node); });
-    std::transform(sideDeckNodes.cbegin(), sideDeckNodes.cend(), std::back_inserter(cardList),
-                   [](auto node) { return repeatName(node, "SB:"); });
+
+    QStringList cardList = nodesToCardList(mainDeckNodes) + nodesToCardList(sideDeckNodes, "SB:");
 
     cardList.sort();
     QByteArray deckHashArray = QCryptographicHash::hash(cardList.join(";").toUtf8(), QCryptographicHash::Sha1);


BruebachL:
I'll let @ebbit1q sign off on the std::transform stuff but approved on my end as far as DeckList and DeckListModel architecture are concerned.

cockatrice/src/interface/deck_loader/deck_loader.cpp
```diff
@@ -269,6 +269,23 @@ static QString toDecklistExportString(const DecklistCardNode *card)
...
+ * Converts all cards in the list to their decklist export string and joins them into one string
+ */
+static QString toDecklistExportString(const QList<const DecklistCardNode *> &cardNodes)
+{
+    QString cardsString;
+    for (auto card : cardNodes) {
+        // Get the card name
+        // If it's a token, we don't care about the card.
+        CardInfoPtr dbCard = CardDatabaseManager::query()->getCardInfo(card->getName());
+        if (dbCard && !dbCard->getIsToken()) {
```
ebbit1q: it just needs to only export the mainboard and sideboard zones, tokens are in the tokens zone, knowing they are in the tokens zone is sufficient without using the db

ebbit1q:
how efficient is appending to a qstring? does it reconstruct the string? should we construct strings from stringlists instead?

RickyRister:
Good question. I think using `operator+` creates a new string every time, so it's probably not very efficient.

Fortunately, the code isn't part of a hot path, so I doubt it actually matters in practice.

However, `operator+=` will append to the string in-place (resizing if needed).

libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
```diff
@@ -665,26 +651,18 @@ bool DeckList::deleteNode(AbstractDecklistNode *node, InnerDecklistNode *rootNod
     return false;
 }
 
-static QString computeDeckHash(const InnerDecklistNode *root)
+static QString computeDeckHash(const DeckList &deckList)
 {
+    auto mainDeckNodes = deckList.getCardNodes({DECK_ZONE_MAIN});
```
ebbit1q, line 656: this part made me think of the way I'd like for these to be configurable at some point https://github.com/Cockatrice/Cockatrice/issues/1740#issuecomment-3446786432

we are not near to that point yet though so this stuff is fine for now, we haven't even talked about how we'd hash other zones yet even

libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
```diff
@@ -665,26 +651,18 @@ bool DeckList::deleteNode(AbstractDecklistNode *node, InnerDecklistNode *rootNod
...
+    auto sideDeckNodes = deckList.getCardNodes({DECK_ZONE_SIDE});
+
     QStringList cardList;
-    QSet<QString> hashZones, optionalZones;
 
-    hashZones << DECK_ZONE_MAIN << DECK_ZONE_SIDE; // Zones in deck to be included in hashing process
-    optionalZones << DECK_ZONE_TOKENS;             // Optional zones in deck not included in hashing process
+    std::transform(mainDeckNodes.cbegin(), mainDeckNodes.cend(), std::back_inserter(cardList),
```
ebbit1q: reading your code now makes me quickly find the issue: you are not accounting for multiple copies in your list generation

RickyRister: a

ebbit1q (approved):
the hashing now works as expected an sees a 3% performance increase, thanks

as a side note #5026 has been ready for review, it would have automatically caught these issues in our ci.

PR #6408, opened 2025-12-08
[DeckLoader] Extract cardNode functions to own file

RickyRister:

`DeckLoader` has some static methods that take a `DeckList` and run `forEachCard` on that decklist with a preset functor.
It doesn't make much sense to have those methods in `DeckLoader` at this point.

This is also impeding my DeckList refactor, because `DeckList` won't always be the class that is holding the nodes to be iterated over.

I think it would be better if the functors are exposed, and the user just calls `forEachCard` themselves using the functor.

- Remove the static method in `DeckLoader` that take a `DeckList` and run `forEachCard` on it
- Move the functor structs from `DeckLoader` into a new `CardNodeFunction` file.
- Update usages to directly call forEachCard with the appropriate functor instead of calling the `DeckLoader` static method.

### DIFF
--- a/cockatrice/src/client/network/parsers/interface_json_deck_parser.h
+++ b/cockatrice/src/client/network/parsers/interface_json_deck_parser.h
@@ -6,6 +6,8 @@
 
 #ifndef INTERFACE_JSON_DECK_PARSER_H
 #define INTERFACE_JSON_DECK_PARSER_H
+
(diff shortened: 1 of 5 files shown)


BruebachL:
These are a bit ugly in the first place but this looks cleaner than what we used to have. Thanks, approved.

PR #6415, opened 2025-12-14
[DeckEditor] Use CommanderSpellbook.com to estimate bracket if format is 'commander'

BruebachL:
## Related Ticket(s)
- Follow up to #6166

"What is a bracket?" "How do I know which bracket my deck is?" "They said they were playing a bracket 2 but then they played bracket 4 cards" etc. etc. etc.

- Implement parts CommanderSpellbook.com's public API
- Use the estimate-bracket endpoint to query the bracket for the current decklist if format is 'commander'
- Do this when the format changes (to commander), when the deck is loaded, or when the user manually refreshes
- Add a tooltip which explains the reasoning behind commander spellbooks bracket estimation.

## Screenshots
image
image


cockatrice/src/interface/widgets/tabs/api/commander_spellbook/api_response/commander_spellbook_bracket_tag.h
```diff
@@ -0,0 +1,79 @@
...
+            return "Unknown";
+    }
+    return {};
+}
+
+inline static QString bracketTagToOfficialString(BracketTag tag)
+{
+    switch (tag) {
+        case BracketTag::Ruthless:
+            return "[5] cEDH";
```
ebbit1q: I mentioned in discord how this is a bad idea as the bracket system is much more flexible than this and isn't determined by a static set of rules but rather how consistent a deck is and how well it performs, which is too complex to decide algorithmically

BruebachL: I've added hints/explanations about this.

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -890,6 +890,56 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
...
+            QOverload<int>::of(&QComboBox::currentIndexChanged), this, [this](int index) {
+                int mode = deckEditorCommanderSpellbookIntegrationEnabledSelector.itemData(index).toInt();
+                SettingsCache::instance().setDeckEditorCommanderSpellbookIntegrationEnabled(mode);
+                updateCommanderSpellbookUiState();
+            });
+
+    deckEditorCommanderSpellbookIntegrationBracketNamingSelector.addItem(
+        tr("CommanderSpellbook bracket names")); // index 0 = false
+    deckEditorCommanderSpellbookIntegrationBracketNamingSelector.addItem(
+        tr("Official Commander bracket names (approximate)")); // index 1 = true
```
ZeldaZach: Due to the "rapidly" changing nature of bracket names, I'm not certain how much I love hard-coding in these values. The API from CS makes sense, as those can update in real-time, but we have no way to update this for users if the bracket system changes dramatically overnight.

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/api_response/commander_spellbook_bracket_tag.h
```diff
@@ -0,0 +1,79 @@
...
+    if (s == "C")
+        return BracketTag::Casual;
+    return BracketTag::Unknown;
+}
+
+inline static QString bracketTagToString(BracketTag tag)
+{
+    switch (tag) {
+        case BracketTag::Ruthless:
+            return "Ruthless";
```
ZeldaZach: tr() ?

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/api_response/commander_spellbook_bracket_tag.h
ZeldaZach: I also have the same concern that these bracket tag names could be a bad idea to hard code into the system. What happens if the bracket system suddenly shifts? We'd have outdated terminology with no easy way to push a fix to users out there.

I'm not saying I like the idea of keeping a global data store somewhere on the Cockatrice website, but maybe it makes sense to do that similar to the public servers list? https://github.com/Cockatrice/cockatrice.github.io/blob/master/public-servers.json

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/api_response/commander_spellbook_deck_request.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return mainDeck;
+    }
+    const QVector<CardInDeckRequest> &commanders() const
+    {
+        return commanderDeck;
+    }
+
+private:
+    QVector<CardInDeckRequest> mainDeck;      // maxItems: 600
```
ZeldaZach: are these max limits enforced anywhere?

Anemaly:
yea this is a good idea i think, commanderspellbook can keep up with any commander bracket changes.
As a frequent user, i'd like to see this added. besides, if your deck is mistagged (due to high manacost 2-card combos) you can explain it at the table.

ZeldaZach:
@BruebachL Can we look into bringing this up to date so I can review? I'd love to get this into the program!

BruebachL:
```JSON
{
  "schemaVersion": 1,
  "brackets": [
    {
      "tag": "R",
      "officialName": "[5] cEDH",
      "displayName": "Ruthless",
      "explanation": "Top-tier competitive decks with maximum optimization, fast combos, and minimal variance."
    },
    {
      "tag": "S",
      "officialName": "[4] Optimized",
      "displayName": "Spicy",
      "explanation": "Highly tuned decks with strong synergy and occasional combo finishes."
    },
    {
      "tag": "P",
      "officialName": "[3] Upgraded",
      "displayName": "Powerful",
      "explanation": "Focused decks with clear win conditions and solid consistency."
    },
    {
      "tag": "O",
      "officialName": "[2] Core",
      "displayName": "Oddball",
      "explanation": "Unconventional or thematic decks with some structure but non-standard choices."
    },
    {
      "tag": "PA",
      "officialName": "[1] Exhibition",
      "displayName": "Precon Appropriate",
      "explanation": "Lightly upgraded preconstructed decks or very casual builds."
    },
    {
      "tag": "C",
      "officialName": "[1] Casual",
      "displayName": "Casual",
      "explanation": "Relaxed decks with no strict optimization goals."
    },
    {
      "tag": "U",
      "officialName": "Unknown",
      "displayName": "Unknown",
      "explanation": "Unclassified or missing bracket definition."
    }
  ]
}
```

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/api_response/commander_spellbook_card_result.h
```diff
@@ -0,0 +1,30 @@
+#ifndef COCKATRICE_COMMANDER_SPELLBOOK_CARD_RESULT_H
+#define COCKATRICE_COMMANDER_SPELLBOOK_CARD_RESULT_H
+#include <QJsonObject>
+#include <QString>
+
+class CommanderSpellbookCardResult
```
RickyRister: why is this not a struct.
Also the `fromJson` would fit more as a static factory method.

Same applies for a bunch of the classes here.

cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.h
```diff
@@ -39,6 +57,11 @@ private slots:
     QCheckBox visualDeckStorageSelectionAnimationCheckBox;
     QLabel defaultDeckEditorTypeLabel;
     QComboBox defaultDeckEditorTypeSelector;
+    QLabel deckEditorCommanderSpellbookIntegrationEnabledLabel;
```
RickyRister: These variable are so long that it's impeding readability. Can we shorten these names?
Just get rid of the `deckEditor` from the front. It's not like there's anything else that can use spellbook integration

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -221,13 +263,23 @@ void DeckEditorDeckDockWidget::createDeckDock()
     upperLayout->addWidget(formatLabel, 2, 0);
     upperLayout->addWidget(formatComboBox, 2, 1);
 
-    upperLayout->addWidget(bannerCardLabel, 3, 0);
-    upperLayout->addWidget(bannerCardComboBox, 3, 1);
+    upperLayout->addWidget(bracketLabel, 3, 0);
+
+    auto *bracketRow = new QHBoxLayout;
+    bracketRow->addWidget(bracketValueLabel);
```
RickyRister: Would it be possible to refactor the three widgets plus the associated logic into a separate widget class for the commander bracket row? This will make the code in this class simpler since it would only need to be responsible for hiding/showing the widget when appropriate, and calling the commander bracket widget with the deck contents when signalled.

BruebachL: Done.

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/handle_commander_brackets.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+}
+
+void HandleCommanderBrackets::downloadBracketDefinitions()
+{
+    if (reply) {
+        return;
+    }
+
+    reply = nam->get(QNetworkRequest(QUrl(COMMANDER_BRACKET_JSON_URL)));
```
RickyRister: nit: isn't the constant already a QUrl?

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/api_response/commander_spellbook_estimate_bracket_result.h
```diff
@@ -0,0 +1,32 @@
...
+#include <QJsonObject>
+#include <QVector>
+
+struct EstimateBracketResult
+{
+    static EstimateBracketResult fromJson(const QJsonObject &json);
+
+    QString bracketTag;
+
+    QVector<CommanderSpellbookCardResult> gameChangerCards;
```
RickyRister: QVector is the exact same as QList in qt6. Just use QList instead.

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/commander_bracket_widget.cpp
```diff
@@ -0,0 +1,269 @@
...
+    connect(&SettingsCache::instance().deckEditor(),
+            &DeckEditorSettings::commanderSpellbookIntegrationUseOfficialBracketNamesChanged, this,
+            &CommanderBracketWidget::maybeAutoEstimateBracket);
+
+    setVisible(false);
+}
+
+void CommanderBracketWidget::setDeck(const DeckList &_deck)
+{
+    deck = _deck;
```
RickyRister, line 66: fyi, this will copy the decklist, and means any changes made to the deck in the deck editor won't automatically be reflected. Just double checking that you kept that in consideration.

RickyRister (approved):

cockatrice/src/interface/widgets/tabs/api/commander_spellbook/commander_bracket_widget.cpp
BruebachL: Good catch

PR #6416, opened 2025-12-14
[TabArchidekt] Set game format when importing

BruebachL:
Archidekt is a bit weird because it displays format names in, well, displayNames but then also spits them back out to us as an integer enum value with no mapping.

- Write a wrapper for the displayNames, enumValues and our internal card database names.
- Use this wrapper to determine the appropriate game format for the deck list
- Set it before import


cockatrice/src/interface/widgets/tabs/api/archidekt/api_response/deck/archidekt_api_response_deck.h
```diff
@@ -12,6 +12,228 @@
...
+        {"gladiator", DeckFormat::Gladiator},   {"oathbreaker", DeckFormat::Oathbreaker},
+        {"oldschool", DeckFormat::OldSchool},   {"paupercommander", DeckFormat::PauperCommander},
+        {"pioneer", DeckFormat::Pioneer},       {"predh", DeckFormat::PreDH},
+        {"premodern", DeckFormat::Premodern},   {"standardbrawl", DeckFormat::StandardBrawl},
+        {"timeless", DeckFormat::Timeless}};
+
+    return map.value(apiName.toLower(), DeckFormat::Unknown);
+}
+
+} // namespace ArchidektFormats
```
RickyRister: There's a nontrivial amount of stuff here. Can we move it to a separate file?

cockatrice/src/interface/widgets/tabs/api/archidekt/api_response/deck/archidekt_api_response_deck.h
```diff
@@ -12,6 +12,228 @@
 #include <QString>
 #include <QVector>
 
+namespace ArchidektFormats
+{
+enum class DeckFormat
+{
+    // The order is important since archidekt returns an integer
+    Standard,
```
RickyRister: If the exact integer of the enum is important, then it might be better to make it explicit
```suggestion
    Standard = 1,
```

RickyRister (approved):

cockatrice/src/interface/widgets/tabs/api/archidekt/api_response/deck/archidekt_api_response_deck.h
ebbit1q: it's also possible for them to add formats at the end of their range at any moment so it'd make sense to define those with like an extra thousand in order to make sure they don't conflict

PR #6418, opened 2025-12-14
compile in debug mode on ubuntu 22.04

ebbit1q:
while trying to compile on my old ubuntu 22.04 box (I have a few machines and yes this one is due for an upgrade) I noticed that the debug build just doesn't compile on master anymore, this fixes that and enables it in the ci as well


BruebachL:
I only meant to suggest, apologies.

ebbit1q:
your suggestion is good, I would have immediately accepted it anyway

ZeldaZach (approved):

PR #6426, opened 2025-12-16
[CardInfoPicture] Defer enlargedPixmap creation until needed

BruebachL:
I was inspecting our object tree and noticed that there are a ton of CardInfoPictureEnlargedWidgets nested under our MainWindow for no reason. Every CardInfoPictureWidget initializes one (even if hoverToZoom is disabled).

- Be a little saner about creation
- Only create if hoverToZoom is enabled
- Create when necessary
- Destruct when no longer needed
- Safety checks for object existence

## Screenshots
Before:
image

After:
image

### DIFF
--- a/cockatrice/src/interface/widgets/cards/card_info_picture_widget.cpp
+++ b/cockatrice/src/interface/widgets/cards/card_info_picture_widget.cpp
@@ -450,8 +450,7 @@ void CardInfoPictureWidget::showEnlargedPixmap()
     if (!enlargedPixmapWidget) {
         enlargedPixmapWidget = new CardInfoPictureEnlargedWidget(const_cast<CardInfoPictureWidget *>(this)->window());
         enlargedPixmapWidget->hide();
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/cards/card_info_picture_widget.cpp
```diff
@@ -444,12 +440,20 @@ QMenu *CardInfoPictureWidget::createAddToOpenDeckMenu()
...
 {
     if (!exactCard) {
         return;
     }
 
+    // Lazy creation of the enlarged widget
+    if (!enlargedPixmapWidget) {
+        enlargedPixmapWidget = new CardInfoPictureEnlargedWidget(const_cast<CardInfoPictureWidget *>(this)->window());
+        enlargedPixmapWidget->hide();
+        connect(const_cast<CardInfoPictureWidget *>(this), &QObject::destroyed, enlargedPixmapWidget,
```
RickyRister: why const cast?

BruebachL: You're right, it's not necessary here.

PR #6432, opened 2025-12-19
Doxygen: Enable internal comments

tooomm:
## Related Ticket(s)
- Closes #6385

<br>

>```
>The INTERNAL_DOCS tag determines if documentation that is typed after a
>\internal command is included. If the tag is set to NO then the documentation
>will be excluded. Set it to YES to include the internal documentation.
>The default value is: NO.
>```
>
>There are also several places in the code with comments marked as "internal", which are not included in the docs when doxygen builds:
https://github.com/search?q=repo%3ACockatrice%2FCockatrice%20%5Cinternal&type=code

PR #6433, opened 2025-12-19
Revert "Doxygen: Enable internal comments"

BruebachL:
Reverts Cockatrice/Cockatrice#6432


BruebachL:
We actually don't want this behavior, internal comments are just from external libraries.

PR #6434, opened 2025-12-20
Docs: Use `doxygen-awesome-css` theme, old

tooomm:
This uses https://github.com/jothepro/doxygen-awesome-css in v2.4.1 and includes it as a submodule.

- More modern and clean look
- Better layout for logo and description
- Bigger and readable search dropdown

## Screenshots
image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "vcpkg"]
 	path = vcpkg
 	url = https://github.com/microsoft/vcpkg.git
+	
+[submodule "doxygen-awesome-css"]
+	path = doc/doxygen/theme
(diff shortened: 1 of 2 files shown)


BruebachL:
I would enjoy it if we could stick with our "brand" color, i.e. the Cockatrice green.

The theme itself looks great though.

tooomm:
I do generally agree. I'll finish the implementation of the theme without customization here.
The theme css will override values from our old css though.

To my taste the green was maybe a bit over used before, but we can for sure use it for some highlighting and hover effects etc.!
I'll ping you once the base is set. Need to add some exclusions still.

tooomm:
See #6512

Issue #4114, opened 2020-09-28
resize issues on startup


ebbit1q:
cockatrice doesn't start with the correct layout or window size, most often on linux or windows.
this would be an issue with the way qt stores the window state but not properly restores it, this has been reported multiple times separately here: 
https://github.com/Cockatrice/Cockatrice/issues/1904, https://github.com/Cockatrice/Cockatrice/issues/2621, https://github.com/Cockatrice/Cockatrice/issues/2708, https://github.com/Cockatrice/Cockatrice/issues/3326 https://github.com/Cockatrice/Cockatrice/issues/3460

PR #6438, opened 2025-12-21
[VDE] Minor cleanups, possibly fullscreen width-lock fix

BruebachL:
Users are reporting that the VDE will sometimes maximize past the screen dimensions and then refuse to be resizable until the layout is reset. I suspect this is due to the manual resize management in DeckCardZoneDisplayWidget (though I am not sure).

I honestly don't remember why this resize hack was necessary once upon a time but it does seem like just proper size policies solve the same issue and are much cleaner.

Also, the VDE constructor code is pretty huge since a lot of things need initialization and connection, so I've split these into helper methods.

- No more manual resizeEvent for DeckCardZoneDisplayWidget
- Size policies for container and card zone display widget
- Refactor VDE constructor into helpers
- Plate comments and more sensible order for most VDE .cpp's

### DIFF
--- a/cockatrice/src/interface/widgets/cards/deck_card_zone_display_widget.h
+++ b/cockatrice/src/interface/widgets/cards/deck_card_zone_display_widget.h
@@ -50,6 +50,7 @@ public slots:
     void refreshDisplayType(const DisplayType &displayType);
     void onActiveGroupCriteriaChanged(QString activeGroupCriteria);
     void onActiveSortCriteriaChanged(QStringList activeSortCriteria);
+    QList<QString> getGroupCriteriaValueList();
     void onCategoryAddition(const QModelIndex &parent, int first, int last);
     void onCategoryRemoval(const QModelIndex &parent, int first, int last);
 


ebbit1q:
my testing shows that this does not address the width of the vde, it's still very wide, closing the visual database display tab fixes it, so it's probably caused by the width of the buttons at the top of it. on a related note: how do I reopen one of the vde tabs? I can't find the option to do so

BruebachL:
> my testing shows that this does not address the width of the vde, it's still very wide, closing the visual database display tab fixes it, so it's probably caused by the width of the buttons at the top of it. on a related note: how do I reopen one of the vde tabs? I can't find the option to do so

They're not supposed to be closable, I don't know why they suddenly are.

This is not about the width of the VDE in general. This is specifically about it exceeding the max screen width dynamically which unmaximizes the window.

ebbit1q:
ah, sorry, I misunderstood I think, when I maximize the app and open a deck editor the window will still grow taller even though it should not, I am not sure what you mean with the width however, I never experienced issues with the width of the application while maximized

ebbit1q:
in trying to reproduce this width lock problem I found another issue: using reset layout no longer restores the default dimensions of the main widget and only restores the locations and then repartitions the other widgets, I suspect this to be related to the resizing issue, I tested 2.8.0 and the issue is not present there, it correctly resets the deck editor layout to the exact same no matter what size the widgets have been given

(I keep a branch with 2.8.0 that has bugfixes only here: https://github.com/ebbit1q/Cockatrice/tree/stable)

ebbit1q:
linking #4114

ebbit1q:
so one way I found to have the app extend beyond its intended frame horizontally is by opening the vde, closing the database view, resizing the widgets on the right to their max width, then restarting and opening the vde again; this opens the vde with the same size for the right widgets as before but now with the database view open which has a much larger minimum width, pushing about 1/4th of the right widgets off screen on a 1080p monitor

this is entirely unaffected by this pr in my testing though, is there another cause of this issue?

PR #6439, opened 2025-12-21
Docs: Add & link xsd schema information

tooomm:
- Add some info on XML databases
- Link to related XSD files

### DIFF
--- a/doc/doxygen/extra-pages/user_documentation/deck_management/creating_decks.md
+++ b/doc/doxygen/extra-pages/user_documentation/deck_management/creating_decks.md
@@ -8,10 +8,8 @@ by selecting "Deck Editor" or "Visual Deck Editor" under the "Tabs" application 
 
 # Further References
 
-See @ref editing_decks for information on how to modify the attributes and contents of a deck in the Deck Editor
+- See @ref editing_decks for information on how to modify the attributes and contents of a deck in the Deck Editor
 widgets.
-
-See @ref exporting_decks for information on how to store and persist your deck either in-client or to external
+- See @ref exporting_decks for information on how to store and persist your deck either in-client or to external
 services.
-
-See @ref importing_decks for information on how to import existing decks either in-client or from external services
+- See @ref importing_decks for information on how to import existing decks either in-client or from external services


BruebachL:
Fyi, my high-level plan for this section is to extract a lot of the documentation that we have on [this page](https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets) to the doxygen.

tooomm:
Sounds good, can we use wiki information and include it into the automated docs?
I think the wiki is somehow part of the repo and should be md file based. So probably works?

You want to still include my change for now, or will you look into pulling the wiki section soon?

ebbit1q:
the wiki is part of this repo in the sense that it is a separate repo that is in the tabs, the url for this repo is https://github.com/Cockatrice/Cockatrice.wiki you can check out this repo and modify it with a github account as if it were a normal repo on any account, it is filled with .md files

BruebachL (approved):

tooomm:
>Fyi, my high-level plan for this section is to extract a lot of the documentation that we have on [this page](https://github.com/Cockatrice/Cockatrice/wiki/Custom-Cards-&-Sets) to the doxygen.

@BruebachL Did you actually meant to transfer/move the information from the wiki page?
I don't mind that, just that we don't want to have several places where similar information is maintained.

<br>

> the wiki is part of this repo in the sense that it is a separate repo that is in the tabs, the url for this repo is https://github.com/Cockatrice/Cockatrice.wiki you can check out this repo and modify it with a github account as if it were a normal repo on any account, it is filled with .md files

I was thinking about pulling the information from the wiki and to display it on the docs page.
However, if it's a repo on it's own, we would need to include it in the build pipeline and pull that repo in CI before building.
For local building it needs manual steps as well.
Unless GitHub Wiki repos can be added as a submodules, too.

The handy thing with the wiki is that users can easily contribute and edit themselves.
That also comes with downsides ofc as we can not monitor and ensure quality & correctness all the time.

If the plan is to move away from the wiki though and gather all app stuff in one place, that's also fine. Maybe keeping only user guides from the community in the wiki for example, or user-maintained compilation step-by-step help etc.

PR #6446, opened 2025-12-23
[DeckDockWidget] Correctly handle auto-expanding tree

RickyRister:

Adding cards expands all sections instead of just the one being added to.

- Added signals to `DeckListModel` that are emitted on conditions that require auto-expanding
- Replaced direct access modification of `DeckListModel` for increment/decrement card counts with methods on `DeckListModel`.
  - So we can ensure the correct signals are emitted.
- Use signals from `DeckListModel` to determine when to auto-expand

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
@@ -156,6 +156,9 @@ void DeckEditorDeckDockWidget::createDeckDock()
         // Delay the update to avoid race conditions
         QTimer::singleShot(100, this, &DeckEditorDeckDockWidget::updateBannerCardComboBox);
     });
+    connect(deckModel, &DeckListModel::cardAddedAt, this, &DeckEditorDeckDockWidget::recursiveExpand);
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -674,14 +680,15 @@ bool DeckEditorDeckDockWidget::swapCard(const QModelIndex &currentIndex)
...
-    offsetCountAtIndex(currentIndex, -1);
+    offsetCountAtIndex(currentIndex, false);
     const QString otherZoneName = zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN;
 
-    ExactCard card = CardDatabaseManager::query()->getCard({cardName, cardProviderID});
-    QModelIndex newCardIndex = card ? deckModel->addCard(card, otherZoneName)
-                                    // Third argument (true) says create the card no matter what, even if not in DB
-                                    : deckModel->addPreferredPrintingCard(cardName, otherZoneName, true);
-    recursiveExpand(proxy->mapToSource(newCardIndex));
+    if (ExactCard card = CardDatabaseManager::query()->getCard({cardName, cardProviderID})) {
```
ebbit1q, line 686: :+1:

Issue #6419, opened 2025-12-14
disable the automatic updater by default when not on a release


ebbit1q:
**Similar Requests**

**Description of New Feature**
when running a selfcompiled version there will always be a check if the user is running the latest release, however this check will always fail as it simply checks if it matches what was posted on github.

this check is superfluous, we should only enable this behavior when compiling a tag in my opinion

**Additional Context**
it should be easy enough to check for a tag being checked out in our ci but also by purely doing this from inside cmake, either way is valid and I'm not entirely sure which is best. I'm leaning towards only our ci though, as if someone is repackaging our software they will probably check out a tag and they will never want to use our update checker.

tooomm:
Additionally, the updater should maybe not run on first launch to smoothen out the initial experience and focus on the tutorials, tips and getting started hints. That is already a lot of input and dialogs next to our potentially overwhelming UI. 

With the second run we can start bringing them to the very latest version.

It might be that they find a direct link to an older release or are otherwise shared an older version which would still be tagged and trigger the updater.

Thoughts?

ebbit1q:
I agree, we can run the updater on second run, similarly to the tip of the day dialog

ebbit1q:
this is quite annoying for distros that package us primarily, it'd be nice to have this in a release

PR #6447, opened 2025-12-25
change the release channel based on version string

ebbit1q:
## Related Ticket(s)
- related https://github.com/Cockatrice/Cockatrice/issues/6419 but I don't agree it fixes it entirely

people installing the beta version will have the updater, which runs immediately by default, tell them to install the stable version

- have the default channel depend on the version string containing the word "beta"
- check on version change and update the channel if needed
- also do the same for automatic updates and the word "custom"
- for fixing #6419 I think it should not depend on the word custom but instead default to off unless set by the compile environment, I am leaving that for a separate solution

### DIFF
--- a/cockatrice/src/interface/window_main.cpp
+++ b/cockatrice/src/interface/window_main.cpp
@@ -943,8 +943,7 @@ void MainWindow::startupConfigCheck()
 
         if (SettingsCache::instance().getCheckUpdatesOnStartup()) {
             if (QString(VERSION_STRING).contains("custom", Qt::CaseInsensitive)) {
-                qCInfo(WindowMainStartupShortcutsLog)
-                    << "Update has changed to custom version, disabling auto update";
+                qCInfo(WindowMainStartupShortcutsLog) << "Update has changed to custom version, disabling auto update";
                 SettingsCache::instance().setCheckUpdatesOnStartup(Qt::Unchecked);
             } else {
                 int channel = 0;
@@ -952,8 +951,7 @@ void MainWindow::startupConfigCheck()
                     channel = 1;
                 }
                 if (SettingsCache::instance().getUpdateReleaseChannelIndex() != channel) {
-                    qCInfo(WindowMainStartupShortcutsLog)
-                        << "Update has changed beta state, updating release channel.";
+                    qCInfo(WindowMainStartupShortcutsLog) << "Update has changed beta state, updating release channel.";
                     SettingsCache::instance().setUpdateReleaseChannelIndex(channel);
                 }
             }


cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -198,15 +199,29 @@ SettingsCache::SettingsCache()
 
     mbDownloadSpoilers = settings->value("personal/downloadspoilers", false).toBool();
 
-    checkUpdatesOnStartup = settings->value("personal/startupUpdateCheck", true).toBool();
+    if (settings->contains("personal/startupUpdateCheck")) {
```
RickyRister, line 202: Note that if the user is using shared settings between installs (like on mac), that means this check will always return true as long as the user has touched the setting any time in the past.
Is that intentional?

ebbit1q: yes, I'm almost entirely certain that this will basically only ever be false on first run, only then will it provide the defaults, which is why I also added the code that checks on update

RickyRister (approved):

cockatrice/src/interface/window_main.cpp
```diff
@@ -941,6 +941,24 @@ void MainWindow::startupConfigCheck()
         qCInfo(WindowMainStartupShortcutsLog) << "[MainWindow] Migrating shortcuts after update detected.";
         SettingsCache::instance().shortcuts().migrateShortcuts();
 
+        if (SettingsCache::instance().getCheckUpdatesOnStartup()) {
+            if (QString(VERSION_STRING).contains("custom", Qt::CaseInsensitive)) {
+                qCInfo(WindowMainStartupShortcutsLog)
+                    << "[MainWindow] Update has changed to custom version, disabling auto update";
```
RickyRister: fyi, you don't need to put the class name in the log, because the log formatter should already print the class name on that line

ebbit1q: I simply copied the existing log line, it seemed strange to me as well

ebbit1q: I suppose it is probably vestigial

ebbit1q:
```suggestion
                    << "Update has changed to custom version, disabling auto update";
```

cockatrice/src/interface/window_main.cpp
```diff
@@ -941,6 +941,24 @@ void MainWindow::startupConfigCheck()
...
+                SettingsCache::instance().setCheckUpdatesOnStartup(Qt::Unchecked);
+            } else {
+                int channel = 0;
+                if (QString(VERSION_STRING).contains("beta", Qt::CaseInsensitive)) {
+                    channel = 1;
+                }
+                if (SettingsCache::instance().getUpdateReleaseChannelIndex() != channel) {
+                    qCInfo(WindowMainStartupShortcutsLog)
+                        << "[MainWindow] Update has changed beta state, updating release channel.";
```
ebbit1q:
```suggestion
                        << "Update has changed beta state, updating release channel.";
```

cockatrice/src/interface/window_main.cpp
```diff
@@ -941,6 +941,24 @@ void MainWindow::startupConfigCheck()
         qCInfo(WindowMainStartupShortcutsLog) << "[MainWindow] Migrating shortcuts after update detected.";
```
ebbit1q:
```suggestion
        qCInfo(WindowMainStartupShortcutsLog) << "Migrating shortcuts after update detected.";
```

PR #6451, opened 2025-12-26
[Refactor] Clean up some PrintingSelector widgets

RickyRister:

Trying to decouple the PrintingSelector widgets from other widgets some more, in order to set up for some future refactors.

- Remove the `currentZone` param from `PrintingSelector`, since it's not used.
- Clean up some methods by simplifying the code
- Don't store constructor args in fields of the class if the class never uses those fields outside of its constructor.

### DIFF
--- a/cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
+++ b/cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
@@ -137,6 +137,31 @@ void CardAmountWidget::updateCardCount()
     layout->activate();
 }
 
+static QModelIndex addAndReplacePrintings(DeckListModel *model,
+                                          const QModelIndex &existing,
+                                          const ExactCard &rootCard,
+                                          const QString &zone,
+                                          int extraCopies,
+                                          bool replaceProviderless)
+{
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
```diff
@@ -259,22 +258,14 @@ void CardAmountWidget::decrementCardHelper(const QString &zone)
  */
 int CardAmountWidget::countCardsInZone(const QString &deckZone)
 {
-    if (rootCard.getPrinting().getUuid().isEmpty()) {
-        return 0; // Cards without uuids/providerIds CANNOT match another card, they are undefined for us.
-    }
+    QString uuid = rootCard.getPrinting().getUuid();
 
-    if (!deckModel) {
-        return -1;
```
ebbit1q, line 267: isn't this a functional change?

RickyRister: `deckModel` can’t be null. There’s other places in this class where we don’t null check the `deckModel`, so if it’s null it would’ve already segfaulted elsewhere

PR #6453, opened 2025-12-29
Fix bug with next/prev buttons in PrintingSelector

RickyRister:
## Related Ticket(s)
- Fixes bug introduced in #6294

If you modify the deck through the printing selector, the next/prev buttons no longer cross header nodes until you use the mouse to reselect a node 

https://github.com/user-attachments/assets/ae8fd37f-57bf-4d28-9cd0-2de1447170f1


ebbit1q:
as workarounds go this seems entirely manageable to me

RickyRister:
If you think the hacky fix isn't a problem, then I suppose we can just merge it. 
Maybe in the future someone who knows more about the problem can figure out what the root cause is.

tooomm:
That sounds like a good place to add a [\todo comment](https://github.com/search?q=repo%3ACockatrice%2FCockatrice%20%5Ctodo&type=code) that will be picked up by doxygen and is listed in the task list of our code documentation?

PR #6461, opened 2025-12-31
manual drag override

### LINKED ISSUE #3386: Moving around the "View Library" window broken above a certain resolution
<b>System Information:</b>
Client Version: 2.6.1 (2018-07-17)
Build Architecture: 64-bit
Qt Version: 5.9.5
System Locale: en_US

  
 The issue should be clearest from viewing a video example:
https://youtu.be/hJe9161gcto

I'm using a 2k (2560x1440) display. When the Cockatrice window is resized to be smaller, the Library window functions as one would expect. When the main window is fullscreen, or above a certain (currently undetermined threshold), the Library window behaves erratically, jumping around the screen.

okonech:
## Related Ticket(s)
- Fixes #3386

Zoneview window widgets (Library / Graveyard / Sideboard / etc) would  jitter while being dragged. At least on linux and possibly other os with high resolution.

We are removing the drag jitter. Qt interfered with Cockatrice's constraints and the order of coord application. They would fight in the coord logic.
moveEvent fired after Qt moved the item so clamp or no clamp it would cause another move => another moveEvent.
This prevents that loop by intercepting the drag events and updates window position.

We now clamp before applying the position so it still can't be dragged offscreen.

## Screenshots
Video before: https://www.youtube.com/watch?v=mBqARUiaNho
Video after: https://www.youtube.com/watch?v=EEhh0tjJW0w

### DIFF
--- a/cockatrice/src/game/zones/view_zone_widget.cpp
+++ b/cockatrice/src/game/zones/view_zone_widget.cpp
@@ -17,6 +17,7 @@
 #include <QLabel>
 #include <QPainter>
 #include <QScrollBar>
+#include <QStyle>
 #include <QStyleOption>
 #include <libcockatrice/protocol/pb/command_shuffle.pb.h>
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -241,33 +248,148 @@ void ZoneViewWidget::retranslateUi()
...
+    if (!draggingWindow)
         return;
 
-    int titleBarHeight = 24;
+    draggingWindow = false;
+    ungrabMouse();
+}
 
-    QPointF scenePos = pos();
+QPointF ZoneViewWidget::draggedWindowPos(const QPoint &screenPos,
```
RickyRister: Might want to rename it to something like `calcDraggedWindowPos`.
Might also want to put some docs above the method to describe what the parameters are suppose to be, since I can't easily tell.

cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -241,33 +248,148 @@ void ZoneViewWidget::retranslateUi()
...
+                    me->accept();
+                    close();
+                    return true;
+                }
+
+                draggingWindow = true;
+                dragStartItemPos = pos();
+                dragStartScreenPos = me->screenPos();
+
+                dragView = nullptr;
```
RickyRister: Would it be possible to extract the dragView calculation code into its own function? Or if that's too annoying, move it to an immediately invoked lambda that assigns to the dragView.
I find it easier to compartmentalize code when the variable is only assigned to a single time.

cockatrice/src/game/zones/view_zone_widget.cpp
```diff
@@ -241,33 +248,148 @@ void ZoneViewWidget::retranslateUi()
...
+        case QEvent::UngrabMouse:
+            stopWindowDrag();
+            break;
+        case QEvent::GraphicsSceneMousePress: {
+            auto *me = static_cast<QGraphicsSceneMouseEvent *>(event);
+            const Qt::WindowFrameSection section = windowFrameSectionAt(me->pos());
+            if (me->button() == Qt::LeftButton && section == Qt::TitleBarArea) {
+                // avoid drag on close button
+                const QRectF frameRectF = windowFrameRect();
+                const QRectF closeRect(frameRectF.right() - kTitleBarHeight, frameRectF.top(), kTitleBarHeight,
```
RickyRister: How was this calculated?

okonech: We assume a square at the right end of the titlebar the full height of the bar is the close button. Added comment. Would it still be top right on mac?

cockatrice/src/game/zones/view_zone_widget.cpp
okonech: Done

cockatrice/src/game/zones/view_zone_widget.cpp
okonech: Done

cockatrice/src/game/zones/view_zone_widget.cpp
okonech: Actually this should be good now. We now fetch from the style's TitleBarCloseButton. But I don't have a way to test for mac

RickyRister:
Yep, seems to work as expected!

https://github.com/user-attachments/assets/23c70b33-0269-4ea2-866f-18e4d81e40d1

RickyRister (approved):

BruebachL:
I was a bit premature with merging this, apparently there is an issue with the close button event getting swallowed on Windows (Closing via esc works).

PR #6463, opened 2025-12-31
[VDE] Deck Analytics Widgets overhaul

BruebachL:
## Please Note
This PR is fairly big and does a lot to remain backwards compatible with the existing functionality. There are inconsistencies between how things are handled with the different dialogs and some general cleanup work to be done. However, I do not want to blow this PR up. I am also, looking ahead to the 2.11 release, eager to get this into the hands of the beta testers to identify actual issues and improvements.

There is window dressing and code cleanup work left to do on this. My intention is to iterate and I will address any problems in follow up PRs.

The current deck analytics widgets are fairly static, plain, not customizable or able to be hidden/moved.

- Introduce DeckAnalyticsWidgetFactory, which creates self-registering DeckAnalyticsWidgetBase's
- Introduce DeckAnalyticsWidgetBase as an abstract class that says: Every analytics widget has a banner AND must provide a way to configure it through a dialog.
- Introduce ResizablePanel wrapper which allows analytics widgets to be resized, re-ordered via drag-and-drop and auto-scrolls a scrollArea on drag.
- Redo DeckAnalyticsWidget to now support all this: Dynamic panels which can be re-ordered and whose layout can be persisted between restarts. The panel addition dialog populates from the factory which means, technically, it should be possible to have these analytics widgets be 'plugins', if we decide to go down that route.
- Redo some of the analytics widgets, offer more display choices (Bar, combinedBar, Pie, BarChart, Filters, etc.)
- Introduce Mana production and devotion widget
- Introduce hypergeometric probability widget
- Introduce a static converter for game specific colors to ensure predictable and consistent colors for displaying various types.

## Screenshots
Before:
image

After:
image
image
image
image

image
image
image

image
image

### DIFF
--- a/cockatrice/src/interface/widgets/general/display/charts/pies/color_pie.cpp
+++ b/cockatrice/src/interface/widgets/general/display/charts/pies/color_pie.cpp
@@ -87,7 +87,7 @@ void ColorPie::paintEvent(QPaintEvent *)
         QString label = QString("%1%").arg(int(ratio * 100 + 0.5));
 
         QFontMetrics fm(p.font());
-#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 11, 0)
         int labelWidth = fm.horizontalAdvance(label);
 #else
         int labelWidth = fm.width(label);


cockatrice/src/interface/widgets/deck_analytics/deck_list_statistics_analyzer.h
```diff
@@ -23,9 +26,9 @@ class DeckListStatisticsAnalyzer : public QObject
 public:
     explicit DeckListStatisticsAnalyzer(QObject *parent,
```
RickyRister, line 27: Is there a reason we need to have a central statistics analyzer instead of having each analytics widget do their own calculations?

BruebachL: There is, it allows multiple widgets to reuse existing information without having to run those calculations themselves. It's performance optimization.

cockatrice/src/interface/widgets/deck_analytics/deck_list_statistics_analyzer.cpp
```diff
@@ -9,38 +9,101 @@
...
-void DeckListStatisticsAnalyzer::update()
+void DeckListStatisticsAnalyzer::analyze()
 {
-    manaBaseMap.clear();
-    manaCurveMap.clear();
-    manaDevotionMap.clear();
+    clearData();
 
-    QList<ExactCard> cards = model->getCards();
+    auto nodes = model->getDeckList()->getCardNodes();
```
RickyRister: Please do not directly access the underlying decklist in the model for card iteration purposes.
`DeckListModel` should have enough methods to do what you want. If it doesn't, then please implement a new method on that instead of reaching into the underlying decklist.

cockatrice/src/interface/widgets/deck_analytics/deck_list_statistics_analyzer.h
BruebachL: It's Qt. Widgets don't do logic. They display it.

cockatrice/src/interface/widgets/deck_analytics/deck_list_statistics_analyzer.cpp
BruebachL: I know. Refer to "Please Note"

BruebachL: Apologies, it's fixed. I was lazy with the merge.

cockatrice/src/interface/widgets/deck_analytics/deck_analytics_widget_base.h
```diff
@@ -0,0 +1,61 @@
...
+#include "../general/display/banner_widget.h"
+
+#include <QDialog>
+#include <QJsonObject>
+#include <QVBoxLayout>
+#include <QWidget>
+
+class DeckListStatisticsAnalyzer;
+
+class AnalyticsWidgetBase : public QWidget
```
RickyRister: Some classes that refer to this widget are named `DeckAnalyticsWidgetXXX`. `DeckAnalyticsWidget` is a completely different class (in that it's the parent widget for the Deck Analytics tab).

For the sake of consistency (and to avoid confusion), we should rename those classes that interface with `AnalyticsWidgetBase` to `AnalyticsWidgetXXX` instead (and also update the file names to match). Or maybe we should rename either `AnalyticsWidgetBase` or `DeckAnalyticsWidget` so that they're more distinct.

RickyRister: I suggest the name `AnalyticsPanelWidget` if we're planning on renaming `AnalyticsWidgetBase`.

cockatrice/src/interface/widgets/deck_analytics/deck_analytics_widget_registrar.h
```diff
@@ -0,0 +1,15 @@
+#ifndef COCKATRICE_DECK_ANALYTICS_WIDGET_REGISTRAR_H
+#define COCKATRICE_DECK_ANALYTICS_WIDGET_REGISTRAR_H
+
+#include "deck_analytics_widget_factory.h"
+
+class DeckAnalyticsWidgetRegistrar
+{
+public:
+    DeckAnalyticsWidgetRegistrar(const QString &type, const QString &title, DeckAnalyticsWidgetFactory::Creator creator)
```
RickyRister: Using the constructor of an object to statically register the widget to the single factory instance seems like a really roundabout way of doing it. Are you sure we can't just have this code in a static function somewhere?

BruebachL: It's essentially a struct with a bit of logic attached to it, using the constructor here is fine.

This also decouples the widgets from the factory and vice versa and allows dynamic registration in the future.

BruebachL: If we move it to a static function, we have to introduce a call-site, which introduces a central list of plugins to maintain which then sort of defeats the whole "self-registering plugins" benefit. In contrast, the anonymously namespaced constructor *will* run and then simply discard itself after it announced its presence to the factory.

RickyRister: I meant that we move the code in the constructor into a static function, then have the plugins statically call that function instead of the constructor.

cockatrice/src/interface/widgets/deck_analytics/deck_analytics_widget.cpp
```diff
@@ -1,35 +1,291 @@
...
+    for (auto *wrapper : panelWrappers) {
+        QJsonObject entry;
+        entry["type"] = wrapper->getTypeId();
+        entry["config"] = wrapper->panel->saveConfig();
+        entry["height"] = wrapper->getCurrentHeight();
+        arr.append(entry);
+    }
+
+    QSettings s;
+    s.setValue("deckAnalytics/layout", QString::fromUtf8(QJsonDocument(arr).toJson(QJsonDocument::Compact)));
```
RickyRister, line 148: Wait, which file are the analytic settings being saved to?

BruebachL: It's in ~/.config/Cockatrice for now, I plan to address this in a future PR and also allow multiple layouts to be saved etc.

BruebachL: I'm really unhappy about folding more and more things into SettingsCache

cockatrice/src/interface/widgets/deck_analytics/deck_analytics_widget_registrar.h
BruebachL: I get what you’re suggesting, but that doesn’t really change the mechanism. The static function would still need to be invoked during static initialization, otherwise nothing registers. At that point we’re just replacing a purpose-built registrar object with a less explicit static side effect. The constructor version is intentional because it clearly communicates “this translation unit registers a widget when loaded” and avoids extra indirection (a dummy variable we have to keep somewhere and populate with the meaningless results of the static initialization calls).

RickyRister: Just out of curiosity, does directly calling `DeckAnalyticsWidgetFactory::instance().registerWidget({type, title, creator});` in the places that we currently call this constructor not do what we expect it to do?

BruebachL: You can't have a naked function call at anonymous namespace level, as far as I understand. It must live inside something: a constructor, a lambda, a dummy static initializer variable, or an external function call. At that point, we're back to the external dependency and the central list of widgets to register.

This anon namespace ensures that if the file is linked, the widget is registered. This is in favor of an eventual plugin system which would rely on this to allow for extension.

RickyRister: > You can't have a naked function call at anonymous namespace level

Ah, so that was what I was missing.

BruebachL: Really glad we discussed this. I had a static initializer version (DeckAnalyticsWidget's constructor) and it was just less maintainable and elegant than this (and not dynamic in terms of plugins)

PR #6469, opened 2026-01-01
[TabRoom] Add a setting to hide the new filter toolbar

BruebachL:
## Related Ticket(s)
- Fixes Devin complaining about this

Toolbar takes up vertical space

- Add a setting to hide it

### DIFF
--- a/cockatrice/src/client/settings/cache_settings.h
+++ b/cockatrice/src/client/settings/cache_settings.h
@@ -1023,7 +1023,7 @@ public slots:
     void setAnnotateTokens(QT_STATE_CHANGED_T _annotateTokens);
     void setTabGameSplitterSizes(const QByteArray &_tabGameSplitterSizes);
     void setShowShortcuts(QT_STATE_CHANGED_T _showShortcuts);
-    void setShowGameSelectorFilterToolbar(Qt::CheckState _showGameSelectorFilterToolbar);
+    void setShowGameSelectorFilterToolbar(QT_STATE_CHANGED_T _showGameSelectorFilterToolbar);
     void setDisplayCardNames(QT_STATE_CHANGED_T _displayCardNames);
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -462,8 +462,13 @@ AppearanceSettingsPage::AppearanceSettingsPage()
     showShortcutsCheckBox.setChecked(settings.getShowShortcuts());
     connect(&showShortcutsCheckBox, &QCheckBox::QT_STATE_CHANGED, this, &AppearanceSettingsPage::showShortcutsChanged);
 
+    showGameSelectorFilterToolbarCheckBox.setChecked(settings.getShowGameSelectorFilterToolbar());
+    connect(&showGameSelectorFilterToolbarCheckBox, &QCheckBox::QT_STATE_CHANGED, &settings,
+            &SettingsCache::setShowGameSelectorFilterToolbar);
+
     auto *menuGrid = new QGridLayout;
     menuGrid->addWidget(&showShortcutsCheckBox, 0, 0);
+    menuGrid->addWidget(&showGameSelectorFilterToolbarCheckBox, 1, 0);
```
RickyRister, line 471: not sure if "menu settings" is the most accurate place for it, but I can't think of anything better either, so I suppose it's fine.

RickyRister (approved):
Yep, works. Thanks

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
BruebachL: Yeah, I don't know. It seemed a little silly to have a group for a single setting. We have some appearance settings like "use tear-off menus" that could also be moved here, I guess

PR #6474, opened 2026-01-01
[DeckList] move SideboardPlan into separate file

RickyRister:

`SideboardPlan` has enough stuff in it that it can be in its own file.

- Move `SideboardPlan` to separate file
- Fix the incorrect location for `deck_list_node_tree.h` in the cmake from before


BruebachL:
I've been meaning to do that *forever* now.

PR #6483, opened 2026-01-02
[Game] Add shortcuts for same size and hand size - 1 mulligans

BruebachL:
Having to click 'Ok' every time you mulligan is really tedious and it's an action that is performed *constantly*.

- Add shortcuts to allow players to mulligan the same hand size or their hand size - 1 without confirming
- This doesn't modify the starting hand size, so if a player opens up the mulligan dialog again, it's set to whatever they had it set to before.


cockatrice/src/client/settings/shortcuts_settings.h
```diff
@@ -660,6 +660,12 @@ class ShortcutsSettings : public QObject
         {"Player/aMulligan", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Mulligan"),
                                          parseSequenceString("Ctrl+M"),
                                          ShortcutGroup::Drawing)},
+        {"Player/aMulliganSame", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Mulligan (Same hand size)"),
+                                             parseSequenceString("Ctrl+Shift+M"),
+                                             ShortcutGroup::Drawing)},
+        {"Player/aMulliganMinusOne", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Mulligan (Hand size - 1)"),
+                                                 parseSequenceString("Ctrl+Shift+Alt+M"),
```
RickyRister, line 667: Every time I see another default shortcut, I ask myself what default shortcuts we still have left and whether it's worth using one of those slots for the action.

`Mulligan (Same hand size)` probably gets used often enough to be worth it. I'm not sure if `Mulligan (Hand size - 1)` shares the same honor, but `Ctrl+Shift+Alt+M` isn't a popular shortcut sequence, so I suppose using up that sequence for this action is probably fine.

RickyRister:
When you start the game, you can't immediately do `Mulligan (Same hand size)`. You have to do the normal mulligan action first.

I know that logically that's how it should behave, but it does seem kinda unintuitive that the major time saving shortcut can't actually be used in the scenario that it would be most useful in.

I suppose it's fine for now though. Maybe we can decide on how we want this action to behave with empty hands later.

PR #6487, opened 2026-01-03
Exclude spaces and punctuation marks from some tr() strings

tooomm:
## Related Ticket(s)
- Spotted when looking at #6480

Spaces and colons are sometimes included in the translatable string, and are shown to translators out of context in the transifex webpage when doing translations.

This leads to easy, unintended mistakes that will mess with the interface:
image
image

- Exclude spaces and punctuation marks from tr() strings

### DIFF
--- a/cockatrice/src/game/player/menu/card_menu.cpp
+++ b/cockatrice/src/game/player/menu/card_menu.cpp
@@ -396,7 +396,7 @@ void CardMenu::addRelatedCardActions()
             relatedCardName = relatedCard.getName(); // "name"
         }
 
-        QString text = tr("Token: ");
+        QString text = tr("Token") + ": ";
         if (cardRelation->getDoesAttach()) {
             text +=
                 tr(cardRelation->getDoesTransform() ? "Transform into " : "Attach to ") + "\"" + relatedCardName + "\"";
--- a/cockatrice/src/interface/widgets/tabs/api/edhrec/tab_edhrec_main.h
+++ b/cockatrice/src/interface/widgets/tabs/api/edhrec/tab_edhrec_main.h
@@ -29,7 +29,7 @@ public:
     [[nodiscard]] QString getTabText() const override
     {
         auto cardName = cardToQuery.isNull() ? QString() : cardToQuery->getName();
-        return tr("EDHRec: ") + cardName;
+        return tr("EDHRec") + ": ") + cardName;
     }
 
     CardSizeWidget *getCardSizeSlider() const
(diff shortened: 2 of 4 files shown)


cockatrice/src/interface/widgets/tabs/api/edhrec/tab_edhrec_main.h
```diff
@@ -29,7 +29,7 @@ class TabEdhRecMain : public Tab
     [[nodiscard]] QString getTabText() const override
     {
         auto cardName = cardToQuery.isNull() ? QString() : cardToQuery->getName();
-        return tr("EDHRec: ") + cardName;
+        return tr("EDHRec") + ": ") + cardName;
```
tooomm: `return tr("EDHRec: ") + cardName;`

Is the old code even working as intended? @BruebachL

In the latest release, I do not see a card name displayed in the tab title when selecting one:
Image
Maybe I'm doing something wrong?

tooomm:
There are probably more tr() with leading/trailing spaces across the code base.

cockatrice/src/game/player/menu/card_menu.cpp
```diff
@@ -396,7 +396,7 @@ void CardMenu::addRelatedCardActions()
             relatedCardName = relatedCard.getName(); // "name"
         }
 
-        QString text = tr("Token: ");
+        QString text = tr("Token") + ": ";
```
ebbit1q, line 399: this is the wrong way to go about it for some languages, instead use the %1 replacements to give translators complete access to how this is displayed

tooomm: I don't think that's how we do it in other places?

We use "%" placeholders in sentences to allow proper injection/placement of variables to account for language specific grammar. Not single word labels.

The only place I can think off where something language specific would matter here in those non-sentence label examples are RTL languages. Not sure we even have support for that.

tooomm: Can you elaborate on your comment when you have a chance, @ebbit1q?

ebbit1q: so my problem is with this entire construction method used in this function, appending to the string with translated words is a bad idea, the most simple method is to duplicate the string for each iteration, alternatively to use % placeholders for each word added to the string as a whole.
eg:
```
QString text;
if ([case a]) {
    text = tr("Token: Transform into \"%1\"").arg(relatedCardName)
} else if ([case b]) {
    text = tr("Token: Attach to \"%1\"").arg(relatedCardName)
} else if ([case c]) {
    text = tr("Token: X \"%1\"").arg(relatedCardName)
} else if ([case d]) {
    text = tr("Token: %1 \"%2\"").arg(cardRelation->getDefaultCount()).arg(relatedCardName)
```
basically + should never be used with in combination with a string that has come out of a tr()

PR #6491, opened 2026-01-04
build: use qt 6.10 on macOS and win10

brunoalr:
Qt 6.6 is EOL. Latest LTS version is [6.8](https://doc.qt.io/qt-6/qt-releases.html#supported-versions), getting support until late 2029 

- Use qt 6.8.\* on macOS and Win10

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -346,7 +346,7 @@ void ThemeManager::themeChangedSlot()
         qApp->setStyle(QStyleFactory::create("Fusion"));
         qApp->setPalette(createDarkGreenFusionPalette());
     } else {
-        qApp->setStyle(QStyleFactory::create(QStyleFactory::keys().first()));
+        qApp->setStyle("");
     }
 
     if (dirPath.isEmpty()) {


ebbit1q:
did you test the build artifacts?

brunoalr:
> did you test the build artifacts?

only the macos 15 ones 👀 will test the remaining ones for good measure

ebbit1q:
I recall there being an issue on windows with 6.8, this needs to be investigated

tooomm:
>Qt 6.6 is EOL. Latest LTS version is [6.8](https://doc.qt.io/qt-6/qt-releases.html#supported-versions), getting support until late 2029

Qt licensing is weird, with their open source and commercial approach.
The link you refer to above relates to "Standard Support", but in the Qt world that means already that you need to be a license holder as I understand it. It's also stated above the table on the Qt page you linked.

Qt 6.8 (and even 6.9) is also EOL already for open source, non-paid commercial licenses, see https://endoflife.date/qt.
LTS is also not relevant for non commercial use either it seems.
Technically we should use more recent versions (if not the most recent one available) as they fall out of support rather quickly after 6 months.
6.9 is also unsupported already, and 6.10 as the newest version is the only one in active open source support.

This PR is obviously an improvement nontheless!

ebbit1q:
we can try using 6.10

brunoalr:
let me check

brunoalr:
tested on win11 (arm64, running on macOS via Parallels Desktop VM): it compiles and runs, but the visuals are a bit different:
beta 48 (qt 6.6):
image 

qt 6.10:
qt6 10 

No idea why that happens.

brunoalr:
unrelated to the current change: now I can't see the text of the card types in the VDE.

ebbit1q:
> unrelated to the current change: now I can't see the text of the card types in the VDE.

is this a bug? please ticket it if present on master

ebbit1q:
about the theme inconsistency, see https://github.com/Cockatrice/Cockatrice/pull/6596

brunoalr:
> > unrelated to the current change: now I can't see the text of the card types in the VDE.
> 
> is this a bug? please ticket it if present on master

will check, but I guess it's on master too.

checked: happens on master too. #6597

tooomm:
> tested on win11 (arm64, running on macOS via Parallels Desktop VM): it compiles and runs, but the visuals are a bit different: beta 48 (qt 6.6): image
> 
> qt 6.10: qt6 10
> 
> No idea why that happens.

You merged master with quite some changes in before bumping to 6.10 and trying that:
image

I did rerun the GH Actions for the master merge with Qt6.8 still.
You can download the Win10 artifact from there and see if the issue exists there, too: https://github.com/Cockatrice/Cockatrice/actions/runs/21462895621#artifacts

tooomm:
I checked again how the Qt version install via the action works (https://github.com/jurplel/install-qt-action?tab=readme-ov-file#options).
Currently we manually give the Qt version (as updated in this PR).

If no version is provided to the action, it uses a given default (latest LTS). It seems that is super quickly updated with new releases and then available in the next version some weeks later.
On the same day of Qt 6.8.3 LTS release, it got bumped in the repo: https://github.com/jurplel/install-qt-action/commit/5c072fca58afdb884c52a49f3e3f0c03e0ab66b1

In order to make our live easier and avoid constant need to maintain this in our end - which will result in outdated versions anyway with the open source version support schedule Qt uses...
Maybe the better call is actually to rely on the Qt action managing the version and float with the latest LTS.

brunoalr:
> > tested on win11 (arm64, running on macOS via Parallels Desktop VM): it compiles and runs, but the visuals are a bit different: beta 48 (qt 6.6): image
> > qt 6.10: qt6 10
> > No idea why that happens.
> 
> You merged master with quite some changes in before bumping to 6.10 and trying that: image
> 
> I did rerun the GH Actions for the master merge with Qt6.8 still. You can download the Win10 artifact from there and see if the issue exists there, too: [Cockatrice/Cockatrice/actions/runs/21462895621#artifacts](https://github.com/Cockatrice/Cockatrice/actions/runs/21462895621#artifacts)

I also checked the latest commit (https://github.com/Cockatrice/Cockatrice/actions/runs/21961685440?pr=6491) in this PR (in which I cherry-picked the commit from @ebbit1q regarding themes) and the differences still happen.

on the other hand, I compiled Cockatrice from source on the same win arm64 VM in which I tested the binaries from the pipeline and it seems that the differences are gone. In order to compile on win arm64 I had to locally update vcpkg, tho. I also installed arm64 Qt locally using `aqt`.

So now I have two hypotheses: 1) vcpkg somwehow messes up with qt (highly unlikely) and 2) there might be a bug on qt side when running x64 binaries on a arm64 machine.

to confirm 2), it would be nice if someone if Win 11 on x64 could test it (should be much easier than finding someone running win on arm64 :P )

brunoalr:
> I checked again how the Qt version install via the action works ([jurplel/install-qt-action#options](https://github.com/jurplel/install-qt-action?tab=readme-ov-file&rgh-link-date=2026-02-15T16%3A13%3A51Z#options)). Currently we manually give the Qt version (as updated in this PR).
> 
> If no version is provided to the action, it uses a given default (latest LTS). It seems that is super quickly updated with new releases and then available in the next version some weeks later. On the same day of Qt 6.8.3 LTS release, it got bumped in the repo: [jurplel/install-qt-action@5c072fc](https://github.com/jurplel/install-qt-action/commit/5c072fca58afdb884c52a49f3e3f0c03e0ab66b1)
> 
> In order to make our live easier and avoid constant need to maintain this in our end - which will result in outdated versions anyway with the open source version support schedule Qt uses... Maybe the better call is actually to rely on the Qt action managing the version and float with the latest LTS.

I'm also ok with that. keep in mind we also need to check if the `qt_arch` version also gets updated automatically. e.g. https://github.com/Cockatrice/Cockatrice/pull/6491/changes#diff-73d26ed8f688bf0bdc3abc2d453358c8f72c00a45c082b2a2ecae0972018a8feR330 - newer qt relies on `win64_msvc2022_64` rather than `win64_msvc2019_64`

brunoalr:
the visual differences *might* be some bug regarding high contrast mode being used incorrectly: https://www.qt.io/blog/high-contrast-mode-in-qt-6.10. I don't use high contrast mode in the VM that I use for testes.

tooomm:
> to confirm 2), it would be nice if someone if Win 11 on x64 could test it (should be much easier than finding someone running win on arm64 :P )

I can do that if you let me know which change/issue to exactly watch out for.

I'm not aware of any high contrast settings in Windows that I've enabled (unless some are default) by the way. 

> I'm also ok with that. keep in mind we also need to check if the qt_arch version also gets updated automatically. e.g. https://github.com/Cockatrice/Cockatrice/pull/6491/changes#diff-73d26ed8f688bf0bdc3abc2d453358c8f72c00a45c082b2a2ecae0972018a8feR330 - newer qt relies on win64_msvc2022_64 rather than win64_msvc2019_64

Is msvc2022 needed with Qt 6.8 already, or only 6.10?

Edit, from the actions docs:
"Default values:
Windows w/ Qt >= 6.8: win64_msvc2022_64"

ebbit1q:
@brunoalr please merge master into this or rebase it

ebbit1q:
the style reset is now merged, I'd like to get this merged asap and explore a full release with it included

PR #6494, opened 2026-01-05
[AnalyticsPanel] Use cogwheel icon for configure button

RickyRister:
## Related Ticket(s)
- Follow-up to #6463

I think the gear icon looks nicer than the "configure" text

- Use gear icon instead of text

### Before:
Screenshot 2026-01-04 at 11 27 39 PM

Screenshot 2026-01-04 at 11 27 55 PM

### After:
Screenshot 2026-01-04 at 11 15 14 PM

Screenshot 2026-01-04 at 11 22 31 PM


BruebachL:
This also fixes the slight annoyance of the configure button shifting the horizontal alignment.

PR #6505, opened 2026-01-10
[AnalyticsPanel] Move close button to on panel

RickyRister:
## Related Ticket(s)
- Follow-up to #6463 and #6494

Having the close buttons directly on the panels feels more intuitive than having to select the panel and then click the close button at the top.

- Add a close button to the left of the banner on the panel.
  - Close button can still be hidden, so that the panel in the sample hand widget still isn't closable
- Update `DeckAnalyticsWidget` 
  - Wire up close button to remove panel
  - Remove close button from the top row
  - Add right stretch to top row so the buttons are all aligned to the left

https://github.com/user-attachments/assets/5d20c125-cfea-491d-ad84-d7b5f9c64428

## Screenshots

Screenshot 2026-01-10 at 12 44 02 AM

### DIFF
--- a/cockatrice/src/interface/widgets/deck_analytics/abstract_analytics_panel_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_analytics/abstract_analytics_panel_widget.cpp
@@ -13,6 +13,16 @@ AbstractAnalyticsPanelWidget::AbstractAnalyticsPanelWidget(QWidget *parent, Deck
 
     bannerAndSettingsLayout = new QHBoxLayout(bannerAndSettingsContainer);
     bannerAndSettingsContainer->setLayout(bannerAndSettingsLayout);
+
+    // close button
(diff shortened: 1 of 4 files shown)


BruebachL:
This does solve the asymmetry but looks a bit odd to me tbh.

RickyRister:
Is the problem with the ui or the functionality? If the problem is with the ui, then do you have a suggestion for what to do instead?
Otherwise, we could merge this now and update the ui once we come up with something better.

BruebachL:
I don't think users will nuke panels constantly or even frequent enough to justify having the close button always visible.

I see more of a use case for multiple layouts (per-deck and global) but I haven't gotten around to fleshing out what I've had in that regard.

RickyRister:
Maybe a giant close button always being visible isn’t what we want, but I’m still of the opinion that the closing should be done by something “in” the panel instead of selecting the panel and then clicking an external button

PR #6506, opened 2026-01-10
[ColorIdentityWidget] Refactor

RickyRister:

`ColorIdentityWidget` has an unused constructor. I don't see how that constructor would ever be used. Might as well clean it up at the same time.

- Remove the constructor that takes a `CardInfoPtr`
- Rename field from `manaCost` to `colorIdentity`
- Some more cleanup


BruebachL:
What do you mean you don't see how this could ever be used? It's a widget for displaying the color identity of a card, initializing it with a card saves the caller from having to parse the identity themselves.

RickyRister:
fair enough

PR #6509, opened 2026-01-11
[DeckAnalytics] Enforce WUBRGC ordering for analytics.

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/libcockatrice_utility/libcockatrice/utility/color.h
+++ b/libcockatrice_utility/libcockatrice/utility/color.h
@@ -22,6 +22,8 @@ inline color convertQColorToColor(const QColor &c)
     return result;
 }
 
+#include <QSet>
+
 namespace GameSpecificColors
 {
 namespace MTG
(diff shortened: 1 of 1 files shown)


libcockatrice_utility/libcockatrice/utility/color.h
```diff
@@ -56,6 +58,32 @@ inline QColor colorHelper(const QString &name)
...
+inline QList<QPair<QString, int>> sortManaMapWUBRGCFirst(const QMap<QString, int> &input)
+{
+    static const QStringList priorityOrder = {"W", "U", "B", "R", "G", "C"};
+
+    QList<QPair<QString, int>> result;
+    QSet<QString> consumed;
+
+    // 1. Add priority colors in fixed order
+    for (const QString &key : priorityOrder) {
+        auto it = input.find(key);
```
RickyRister, line 71: nit: Why use `.find` instead of `.contains` or `.value` and check against the default? 
I feel like those make it clearer what you're trying to do.

PR #6510, opened 2026-01-11
[Doxygen] Add documentation for beta channel.

BruebachL:
"How do I switch to the beta?"

- Add user docs page for the beta channel with pretty pictures

## Screenshots
image
image


tooomm:
I think it would be better if we use screenshots in the docs from the default theme.

BruebachL:
Sure, whatever

BruebachL:
Apologies for the whatever, it truly doesn't matter to me and I was just a lil grumpy because I was in pain.

tooomm:
No worries. I try to avoid people reaching out for custom theme questions rather than the app.

Why does it look different for me?
Where do I get this icons?
How to enable Dark Mode?

(I'm aware we have some others screenshots which do not use the default theme already)

PR #6516, opened 2026-01-14
[VDE] Placeholder image for deck view if deck is empty

BruebachL:
The initial screen is blank and unfriendly, offering no hints to the user.

- Add a placeholder widget with a label and an image if the deck is empty.
- Hide if not empty.
- Show again if empty.

## Screenshots
image


RickyRister:
Somehow your change makes the deck dock not be able to be visible.

https://github.com/user-attachments/assets/0136d14b-8c9a-41db-90be-cbfa96f78501

BruebachL:
> Somehow your change makes the deck dock not be able to be visible.
> example.mp4

Could this be a regression from #6499?

BruebachL:
Yes, regressed on master, as well as several other issues with dock management in general (printing selector refuses to become visible again, wrong checkstate).

See #6517

RickyRister:
Apparently it only fails to show on debug but not on release. I'll look into it.

BruebachL:
Thanks :)

RickyRister:
Turns out it was a typo #6518

BruebachL:
image
image

RickyRister:
I'm seeing blurring at the edges of the images. Seems like Qt is loading the svg's into a raster image and then scaling it up, instead of scaling up the vector image first. Not sure how much we care about that.

Screenshot 2026-01-14 at 5 09 42 AM
Screenshot 2026-01-14 at 5 09 56 AM

BruebachL:
I'd say that's fine-ish for now but definitely something to ticket for later.

RickyRister:
Do you think using more faded colors would look better? Makes it more obvious that it's in the background.

Screenshot 2026-01-14 at 5 33 53 AM

(I used `#c0c0c0` for the light color and `#989898` for dark here)

BruebachL:
Yeah, that looks great.

cockatrice/src/interface/widgets/printing_selector/printing_selector_placeholder_widget.h
```diff
@@ -0,0 +1,23 @@
+#ifndef COCKATRICE_PRINTING_SELECTOR_PLACEHOLDER_WIDGET_H
+#define COCKATRICE_PRINTING_SELECTOR_PLACEHOLDER_WIDGET_H
+
+#include <QLabel>
+#include <QVBoxLayout>
+#include <QWidget>
+
+class PrintingSelectorPlaceholderWidget : public QWidget
```
RickyRister, line 8: nit: Couldn't we just have a single configurable `PlaceholderImageWidget` class and create two instances of it, instead of creating two separate classes that do mostly same thing?

BruebachL: I pondered this as well but I think ultimately, if we setObjectName different things here, we could have theme creators configure them differently (i.e. hide the text on one by setting it equal to the background)?

But you're not wrong, it's possible there's an elegant solution to this.

BruebachL: No, we can just setObjectName at the call site, I'm talking nonsense. Sorry, I'm resting because of my pain issue so I'm not fully here.

PR #6527, opened 2026-01-16
[TabGame] Don't create replay dock if not replay tab

RickyRister:

The replay timeline dock is still accessible in non-replay games through the view menu.
Making the timeline dock visible will reveal a glitched timeline widget.

https://github.com/user-attachments/assets/04d6b062-1a50-4353-ba9e-8cf00ec87862

- Don't create replayDock at all if not replay tab (assign replayDock = nullptr)
  - This also means replayDock's dependencies don't get created
- Null check the `replayDock` to determine if game tab is a replay tab, instead of the previous method of checking the replayManager.  

https://github.com/user-attachments/assets/3ca1c46e-e517-4fa3-8555-90160906d562

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_game.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_game.cpp
@@ -255,7 +255,7 @@ void TabGame::emitUserEvent()
 
 TabGame::~TabGame()
 {
-    if (replayDock) {
+    if (replayManager) {
         delete replayManager->replay;
     }
 }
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -256,13 +255,15 @@ void TabGame::emitUserEvent()
 
 TabGame::~TabGame()
 {
-    delete replayManager->replay;
+    if (replayDock) {
```
ZeldaZach, line 258: should check if replayManager exists too

ZeldaZach (approved):

PR #6536, opened 2026-01-19
[CardInfo] Refactor: add getLegalityProp method

RickyRister:
## Related Ticket(s)
- Follow-up to #6460

`DeckListModel` shouldn't need to know how legality props are stored.

- Add a `getLegalityProp` method to `CardInfo`
- Update usages where outside code was looking up a "format-" property

### DIFF
--- a/libcockatrice_card/libcockatrice/card/card_info.cpp
+++ b/libcockatrice_card/libcockatrice/card/card_info.cpp
@@ -75,13 +75,18 @@ QString CardInfo::getCorrectedName() const
     return result.remove(rmrx).replace(spacerx, space);
 }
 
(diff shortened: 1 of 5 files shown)


BruebachL:
We don't generally have setters/getters for properties and the existing ones we do have are marked for 'deprecation ASAP' iirc.

Requesting @ebbit1q or @ZeldaZach on this.

RickyRister:
I feel like this is a special case because the pattern `format-<name>` encodes special logic. All other properties are just a transparent key lookup, but for format legality you need to remember that it takes the form of `format-<name>`

ebbit1q:
I think this is fine, which ones are marked for deprecation?

Issue #6538, opened 2026-01-20
Support face-down cards in all public zones


RickyRister:
**Similar Requests**

#5116

**Description of New Feature**

Some cards permanently exile cards from library face-down. (For example, [Doomsday Excruciator](https://scryfall.com/card/dsk/94/doomsday-excruciator)) This has bad ergonomics on Cockatrice. You can't put the cards into the exile zone face-down, so you're forced to leave the face-down exiled cards on the battlefield. 

Cockatrice should allow support for face-down cards in all public zones (stack, graveyard, and exile) in order to solve that problem.

**Additional Context**

This will require changes on both the server and client.

On the server side, we don't actually have to change the proto. The `CardToMove` proto in `command_move_card.proto` already contains a `face_down` bool field. We current set that field in the message when moving cards from hand to play face-down, and when moving face-down cards around on the battlefield.
For all other zone transitions, the client currently leaves that field unset, making it default to false.

Currently, the server will only take the face_down field into account when the card's destination zone is the battlefield. It will assume face_down is false when calculating a move into any other zone.

To implement the change, we have to update the server logic to allow cards to be face-down when moved into any public zone, then update the client so that it sets the `face_down` field for other zone transitions

This server change will actually be backwards compatible, since old clients leave the `face_down` field unset, which means the server defaults to turning the card face up.

PR #6539, opened 2026-01-20
[Server] Support face-down cards in all public zones

RickyRister:
## Related Ticket(s)
- Addresses #6538
- Client changes in #6602

We want to support face-down cards in all public zones (stack, graveyard, and exile)

- Extract faceDown check to method
- Relax the check for the targetZone from `targetzone->hasCoords()` (which only includes the battlefield) to any public zone.
- Add some extra logic for default values to maintain backwards compatibility.

This implementation is completely backwards compatible.

Old clients will continue to always turn face-down cards on the battlefield face-up when moving them to another zone, and will continue to keep cards face-down when moving them around on the battlefield.

Old clients can already handle unknown face-down cards being put into non-battlefield public zones; the card will be displayed as the card back and with no name.

### DIFF
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/command_move_card.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/command_move_card.proto
@@ -6,7 +6,9 @@ message CardToMove {
     // Id of the card in its current zone
     optional sint32 card_id = 1 [default = -1];
 
-    // Places the card face down, hiding its name
+    // If true, places the card face down, hiding its name.
+    // If false, forcibly turns the card face up.
+    // If not set, defers the resulting face down state to the server.
     optional bool face_down = 2;
 
     // When moving add this value to the power/toughness field of the card


libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
```diff
@@ -198,6 +198,30 @@ shouldDestroyOnMove(const Server_Card *card, const Server_CardZone *startZone, c
...
 }
 
+/**
+ * @brief Determines whether the moved card should be face-down
+ */
+static bool
+shouldBeFaceDown(const MoveCardStruct &cardStruct, const Server_CardZone *startZone, const Server_CardZone *targetZone)
+{
+    // being face-down only makes sense for public zones
+    if (targetZone->getType() != ServerInfo_Zone::PublicZone) {
```
ZeldaZach, line 212: Gotta make sure startZone and targetZone are non-null

ebbit1q:
how do I test this change? is there a pr with a client side interface that uses this feature yet?

RickyRister:
> how do I test this change? is there a pr with a client side interface that uses this feature yet?

#6602

ebbit1q:
I checked out and compiled that one, but it doesn't seem to add any options to move cards to the other piles face down?

RickyRister:
I only added support for ctrl drag in that PR

ebbit1q:
the keybind for face down is shift drag, which I tried but didn't seem to work?

ebbit1q:
I tried ctrl drag as well, it does the bottom of library, as expected

ebbit1q:
I tried again by merging all your prs and compiling that, it seems like I just messed up somewhere with the combination of which server and which client I was running, my bad

ebbit1q (approved):
very simple, thanks

PR #6547, opened 2026-01-22
[VDE] Accurately represent card amounts

BruebachL:
We create only a single widget for each index, regardless of card amount.

- Create more widgets if card amount > 1

## Screenshots
image


RickyRister:
The card amounts don't update when the last copy of a card is removed. It only updates once another change is made.

https://github.com/user-attachments/assets/7cea8037-b035-4b8d-a018-e90be5127792

BruebachL:
> The card amounts don't update when the last copy of a card is removed. It only updates once another change is made.
> example.mp4

Should be fixed.

PR #6554, opened 2026-01-24
[DeckEditor] Show info in PrintingSelector dock when override printings enabled

RickyRister:
## Related Ticket(s)
- Follow-up to #6218

There are some users who accidentally enable the override printings setting, and then are confused why the printing selector dock can't be opened.

It would be better for the user experience if the dock clearly told the user why it was disabled instead of silently being greyed out in the view menu.

- Don't hide printing selector dock widget when override printings setting is enabled.
- Extract the override setting warning message (and some of the cleanup logic) to a separate file, since it's going to be used in two places now.
- Create a `PrintingDisabledInfoWidget` that gets shown in the dock instead of the `PrintingSelector` if the override printings setting is enabled
  - The dock dynamically updates whenever the setting changes.
  - Contains a button that allows you disable the setting
    - Will go through the same warning window as with the settings

https://github.com/user-attachments/assets/253357e3-c1be-4cef-a235-28dff219c8da

## Screenshots

Screenshot 2026-01-23 at 11 38 34 PM

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -179,6 +179,7 @@ set(cockatrice_SOURCES
     src/interface/widgets/deck_editor/deck_editor_printing_selector_dock_widget.cpp
     src/interface/widgets/deck_editor/deck_list_style_proxy.cpp
     src/interface/widgets/deck_editor/deck_state_manager.cpp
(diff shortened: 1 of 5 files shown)


BruebachL:
Thanks for doing the work, I reflected on this and I do like it. LGTM.

PR #6577, opened 2026-01-27
[App] Add Fusion Theme

### LINKED ISSUE #6571: Dark Mode
Its now 2026 and I know its been requested but we still don't have a functioning dark mode and custom themes don't function on the newest update for windows so can someone please link me an outdated version where that feature still works I don't want to feel like I'm going blind staring into the sun every time I try to read a card description.

basicer:
Adds a new default theme that causes the QT Fusion theme to be selected. This theme looks a bit nicer than 'Windows' and supports both light and dark mode out of the box.

## Related Ticket(s)
- Fixes #6571

- New theme in the theme selector in addition to Default, "Fusion"

## Screenshots
fusion

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -6,10 +6,15 @@
 #include <QColor>
 #include <QDebug>
 #include <QLibraryInfo>
+#include <QPalette>
 #include <QPixmapCache>
 #include <QStandardPaths>
+#include <QStyleFactory>
+#include <QStyleHints>
+#include <Qt>
 
 #define NONE_THEME_NAME "Default"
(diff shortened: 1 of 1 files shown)


BruebachL:
Fantastic! I knew Qt had the ability to render for different OS's (Fusion being the Linux or platform agnostic one iirc ) but I didn't know you could A: simply tell it to use that style on any platform and B: set style hints like the dark mode.

ZeldaZach (approved):

tooomm:
As I'm not quite sure with the recent changes and addition of the Qt Fusion theme:
For new installations on fresh systems without old configs, will the `Fusion` theme be the default theme that is used in the app?

If so, which of the Fusion variants will be used?
image

PR #6602, opened 2026-02-15
[Client] Support face-down cards in all public zones

RickyRister:
## Related Ticket(s)
- Related to #6539
- Requires #6552 to be merged first

Dragging a card while holding ctrl will cause it to be face-down. Doing the face-down drag from hand or library into graveyard or exile will cause the card to be moved to that new zone face-down.

This PR only implements preliminary support. Extra QoL features, such as actions for moving multiple cards from library to zone face-down, will be implemented in a future PR.

### DIFF
--- a/cockatrice/src/game/log/message_log_widget.cpp
+++ b/cockatrice/src/game/log/message_log_widget.cpp
@@ -343,10 +343,18 @@ void MessageLogWidget::logMoveCard(Player *player,
         finalStr = tr("%1 moves %2%3 to sideboard.");
     } else if (targetZoneName == STACK_ZONE_NAME) {
         soundEngine->playSound("play_card");
-        finalStr = tr("%1 plays %2%3.");
+        if (card->getFaceDown()) {
+            finalStr = tr("%1 plays %2%3 face down.");
+        } else {
+            finalStr = tr("%1 plays %2%3.");
+        }
     } else {
         fourthArg = targetZoneName;
-        finalStr = tr("%1 moves %2%3 to custom zone '%4'.");
+        if (card->getFaceDown()) {
+            finalStr = tr("%1 moves %2%3 to custom zone '%4' face down.");
+        } else {
+            finalStr = tr("%1 moves %2%3 to custom zone '%4'.");
+        }
     }
 
     QString message = finalStr.arg(sanitizeHtml(player->getPlayerInfo()->getName()), cardStr, nameFrom.second);


ebbit1q:
ok, so I've now extensively tested this in my opinion and I've decided that its backwards compatibility is acceptable. the only caveat is that outdated clients will not be able to tell that a facedown card in a graveyard/exile of which the information has been revealed from any other card in the pile.

there are a few things I am missing but these can be considered polish:
- the log does not note the facedown state of a card when moved in a pile, making it hard to tell if a known card was put facedown from the log
- when moving a facedown card to a pile while not holding shift, the card is changed to face up, it'd be preferred that its state be preserved in order to avoid accidentally revealing it (moving a facedown card around in play does not flip it)
- the above also happens when moving a card in the pile (note that holding shift while moving a card should always flip it, but not holding shift should instead preserve its state)
- there is no option to peek at cards in piles
- there is no option to flip cards while they are in a pile (this might actually not be backward compatible however, testing is needed)

RickyRister:
> when moving a facedown card to a pile while not holding shift, the card is changed to face up

This is intended behavior. Currently, when you move a face-down card to another zone, it turns face up. It's more intuitive to keep that behavior as the default. We can add a "keep face-down cards face-down between zones" option in the future.

> the above also happens when moving a card in the pile

This behavior is unintentional.

RickyRister:
> the above also happens when moving a card in the pile

I realized why this was happening, and ended up redesigning the client logic.

In the server code, the `face_down` field of `CardToMove` being unset is semantically meaningful. 
Basically, `face_down` being set means to force the card to that state, while it being unset means to defer to the server to determine the resulting state.
I added a comment to clarify that.

In the client code, I removed the zone-matching logic that was duplicated from the server code.
Now, we just leave the the `face_down` field unset unless `forceFaceDown` is true (eg ctrl is held).
This will defer the face-down calculation to the server, which gives us the expected result when moving face-down cards within the same zone.

Please re-review.

ebbit1q:
thanks for updating it, moving cards inside piles is indeed working the way I expect now, I understand wanting to keep behavior similar to before this change, flipping cards back to normal when moving them between zones is the current behavior simply because other zones don't support the state.

I'd still like to see the logging extended to include moving cards face down into piles, this would be needed to add anyway, adding more options can be done in later prs imo.

RickyRister:
Screenshot 2026-02-26 at 9 39 44 PM

ebbit1q:
very nice, just missing the stack though

RickyRister:
Remember to merge the dependent PRs first

PR #6611, opened 2026-02-21
CI: Add oracle to windows pdb's upload

tooomm:
Oracle was missing.

- Add `Oracle`
- Renaming


ebbit1q:
what is a pdb?

tooomm:
> what is a pdb?

I think those files are needed for debugging.
Zach added them in #5494, but left out Oracle.

Unfortunately, we sometimes have PR's with no details provided.

PR #6612, opened 2026-02-21
[ChatView] Fix game log first line incorrect background color

RickyRister:
## Related Ticket(s)
- Fixes bug introduced in #6581

The first line of the game log uses darkened background color instead of the blank background color

https://github.com/user-attachments/assets/c4bc547e-eedf-4533-8b3d-a2705855d363

Screenshot 2026-02-21 at 4 06 33 AM

This is because an empty QTextDocument still has a single block containing the empty string.
The logic in `refreshBlockColors` iterates over each block in the document to set the color, which means when the document is empty, it will still do one iteration.

- Specifically add a check to `refreshBlockColors` to return early if the document is empty

https://github.com/user-attachments/assets/6ea900b9-c335-4ac4-85a9-678a8ed6f755


ebbit1q (approved):
huh, never noticed that, thanks for fixing this

PR #6620, opened 2026-02-22
[PictureLoader] Allow saving downloaded images to local storage and not just the QNetworkManager cache

BruebachL:
Some users do not trust the QNetworkCache or would prefer to have the images saved to their local computer for various other reasons such as external processing. 

This change introduces a new CheckBox and a LineEdit which will allow users to enable saving to local storage under the naming scheme they provide to the line edit. It's customizable with placeholders.

## Screenshots
image


cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
```diff
@@ -174,6 +181,75 @@ void CardPictureLoader::imageLoaded(const ExactCard &card, const QImage &image)
...
+    // Ensure directory exists
+    if (!outDir.exists()) {
+        if (!baseDir.mkpath(outDir.path())) {
+            qCWarning(CardPictureLoaderLog) << "Failed to create directory for downloaded card image:" << outDir.path();
+            return;
+        }
+    }
+
+    // Do not overwrite existing files
+    if (outInfo.exists()) {
```
RickyRister: nit: it would make more sense to do this check before we do the directory exists check

RickyRister:
`Delete Downloaded Images` doesn't seem to delete the images saved this way. Is that intentional?

RickyRister (approved):
There could be an argument for combining the settings into one line, or moving it into a separate groupbox. 

Screenshot 2026-02-22 at 3 11 11 AM

In fact, if it gets moved to a separate groupbox, then maybe we can also rename the delete button to `Deleted cached images` and add new button that actually deletes these images.

But that's all ui stuff that can be figured out later. If the functionality works, the LGTM.

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -1101,6 +1104,20 @@ void SettingsCache::setNetworkRedirectCacheTtl(const int _redirectCacheTtl)
...
 }
 
+void SettingsCache::setSaveCardImagesToLocalStorage(QT_STATE_CHANGED_T _saveCardImagesToLocalStorage)
+{
+    saveCardImagesToLocalStorage = _saveCardImagesToLocalStorage;
+    settings->setValue("personal/saveCardImagesToLocalStorage", saveCardImagesToLocalStorage);
+    emit saveCardImagesToLocalStorageChanged(saveCardImagesToLocalStorage);
+}
+
+void SettingsCache::setLocalCardImageStorageNamingScheme(const QString _localCardImageStorageNamingScheme)
```
RickyRister: nit: I still prefer naming the param `value` for all the new settings because it's less of a pain to write and remember.

cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
```diff
@@ -174,6 +181,75 @@ void CardPictureLoader::imageLoaded(const ExactCard &card, const QImage &image)
...
+        return;
+    }
+
+    // Save image
+    QImage image = pixmap.toImage();
+    if (!image.save(outInfo.absoluteFilePath(), "PNG")) {
```
RickyRister, line 263: are we only supporting pngs?

ebbit1q: we have always used pngs for these, imo there is no benefit to allowing users to specify their own file format

note that the images are already in binary form at this point and saving the data as received from source would require significant changes

ebbit1q (changes requested):
I don't think we want to have users edit this template at all, we need to make sure that users don't break it if we want to have cockatrice read these images back

having users process our saved images the way you mention is not our concern and if someone wanted to use these saved images in such a way they should be able to adjust to the naming scheme we choose

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -1298,6 +1317,8 @@ void DeckEditorSettingsPage::retranslateUi()
     networkRedirectCacheTtlEdit.setToolTip(tr("How long cached redirects for urls are valid for."));
     pixmapCacheLabel.setText(tr("Picture Cache Size:"));
     pixmapCacheEdit.setToolTip(tr("In-memory cache for pictures not currently on screen"));
+    saveCardImagesToLocalStorageLabel.setText(tr("Save downloaded images to local storage"));
```
ebbit1q: this label is a little confusing, in fact the cached urls are present on the user's local storage but not as readily accessible as they are inside the cache database, I'd prefer if this simply mentioned where it is saving the images to

also, in a way this reinforces the idea of users that the caching of download is absent, which it isn't, it's just somewhere else, we need to make sure that users checking this option know that without this option their downloads are just as securely saved but with the benefit of allowing them to be managed more easily

BruebachL:
> `Delete Downloaded Images` doesn't seem to delete the images saved this way. Is that intentional?

Turns out, that button only deleted subfolders inside downloadedPics recursively but not files within downloadedPics/ itself. 
I've simplified the code a little and fixed this.

cockatrice/src/interface/card_picture_loader/card_picture_loader_local_schemes.h
```diff
@@ -0,0 +1,109 @@
...
+{
+    static QList<NamingSchemeInfo> list = {
+        {NamingScheme::Name_ProviderId, "Card Name + Provider ID", "{name}_{providerId}"},
+        {NamingScheme::Name_Set_Collector, "Card Name + Set + Collector", "{name}_{set}_{collector}"},
+        {NamingScheme::Set_Collector_Name, "Set + Collector + Card Name", "{set}_{collector}_{name}"},
+        {NamingScheme::Name_Set, "Card Name + Set", "{name}_{set}"},
+        {NamingScheme::NameOnly, "Card Name", "{name}"},
+    };
+    return list;
+}
```
RickyRister, line 40: nit: can't you just have a static const instead of a function that returns a static const?
```suggestion
static const QList<NamingSchemeInfo> IMPORT_SCHEMES = {
    {NamingScheme::Name_ProviderId, "Card Name + Provider ID", "{name}_{providerId}"},
    {NamingScheme::Name_Set_Collector, "Card Name + Set + Collector", "{name}_{set}_{collector}"},
    {NamingScheme::Set_Collector_Name, "Set + Collector + Card Name", "{set}_{collector}_{name}"},
    {NamingScheme::Name_Set, "Card Name + Set", "{name}_{set}"},
    {NamingScheme::NameOnly, "Card Name", "{name}"}};
```

cockatrice/src/interface/card_picture_loader/card_picture_loader_local_schemes.h
```diff
@@ -0,0 +1,109 @@
...
+
+inline QString expandPattern(const QString &pattern,
+                             const QString &name,
+                             const QString &set,
+                             const QString &collector,
+                             const QString &providerId)
+{
+    QString result = pattern;
+
+    auto replaceIfPresent = [&](const QString &token, const QString &value) -> bool {
```
RickyRister, line 63: nit: only capture what you need
```suggestion
    auto replaceIfPresent = [&result](const QString &token, const QString &value) -> bool {
```

RickyRister:
After doing some thinking, I'm leaning more towards the idea that these settings should be in a separate section from the `URL Download Priority` section. We can use the terminology "back up downloaded images" instead of "save downloaded images". We want to make it more clear to the user that it's an optional thing.

If we do that, then we can have a separate delete button for the backup, and rename the existing delete button to `Delete Cached Images`

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -1302,18 +1216,237 @@ void DeckEditorSettingsPage::retranslateUi()
...
     networkCacheEdit.setToolTip(tr("On-disk cache for downloaded pictures"));
     networkRedirectCacheTtlLabel.setText(tr("Redirect Cache TTL:"));
     networkRedirectCacheTtlEdit.setToolTip(tr("How long cached redirects for urls are valid for."));
     pixmapCacheLabel.setText(tr("Picture Cache Size:"));
     pixmapCacheEdit.setToolTip(tr("In-memory cache for pictures not currently on screen"));
-    updateNowButton->setText(tr("Update Spoilers"));
-    aAdd->setText(tr("Add New URL"));
-    aEdit->setText(tr("Edit URL"));
-    aRemove->setText(tr("Remove URL"));
+    saveCardImagesToLocalStorageCheckBox.setText(tr("Back up downloaded images to local storage"));
```
RickyRister: Now that we have the caching method in a dropdown, do we still need this to be a checkbox?

cockatrice/src/interface/card_picture_loader/card_picture_loader_cache_method.h
```diff
@@ -0,0 +1,30 @@
...
+struct CacheMethodInfo
+{
+    CacheMethod id;
+    QString displayName;
+};
+
+static inline const QList<CacheMethodInfo> methods()
+{
+    static QList<CacheMethodInfo> all = {
+        {CacheMethod::NETWORK_CACHE, "Network Cache"},
```
RickyRister: this needs to be tr'd somehow

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -1789,6 +1930,7 @@ DlgSettings::DlgSettings(QWidget *parent) : QDialog(parent)
     pagesWidget->addWidget(makeScrollable(new AppearanceSettingsPage));
     pagesWidget->addWidget(makeScrollable(new UserInterfaceSettingsPage));
     pagesWidget->addWidget(new DeckEditorSettingsPage);
+    pagesWidget->addWidget(new StorageSettingsPage);
```
RickyRister, line 1933: We need to make this page scrollable. The text gets cut off at the default window size.
Image

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -1302,18 +1216,246 @@ void DeckEditorSettingsPage::retranslateUi()
...
+           "allow more card images to be displayed at once but shouldn't be necessary. Clearing this will make "
+           "Cockatrice reload all images from the network cache or the disk."));
+
+    clearDownloadedPicsButton.setText(tr("Delete Cached Images"));
+    clearBackupsButton.setText(tr("Delete Saved Images"));
+    clearPixmapCacheButton.setText(tr("Clear In-Memory Images"));
+
+    mpCacheMethodGroupBox->setTitle(tr("Card Picture Loader Cache Method"));
+    mpNetworkCacheGroupBox->setTitle(tr("Network Cache"));
+    mpImageBackupGroupBox->setTitle(tr("Image Backup"));
```
RickyRister: The caching method is called `Filesystem Cache` but the group box is labelled `Image Backup`, despite the fact that the other group box does match the caching method name. 

Image

Is this intentional?

RickyRister:
How it looks currently

image

ebbit1q (approved):
let's see what beta testers think of this change!

PR #6622, opened 2026-02-22
[TabGame] Fix concede removing player without waiting for server

RickyRister:
## Related Ticket(s)
- Fixes bug introduced in #6112

In a game, when the user concedes, the game code will remove the player from the playerManager as it sends the concede command to the server, before and independent of any responses from the server.

If this concede causes the game to end, it will cause the game to "flash" a board with only a single player, before being sent back to the deck view.

https://github.com/user-attachments/assets/fb68e88f-322f-4ac0-8873-81c640c2679f

- Remove the code that removed the player from the playerManager on concede action
  - event handler for server response already properly removes the conceded player

https://github.com/user-attachments/assets/22126a63-32c5-4116-9e8e-95e125c54640

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_game.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_game.cpp
@@ -488,14 +488,12 @@ void TabGame::actConcede()
                                   QMessageBox::Yes | QMessageBox::No, QMessageBox::No) != QMessageBox::Yes)
             return;
         emit game->getPlayerManager()->activeLocalPlayerConceded();
-        player->setConceded(true);
     } else {
         if (QMessageBox::question(this, tr("Unconcede"),
                                   tr("You have already conceded.  Do you want to return to this game?"),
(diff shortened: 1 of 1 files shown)


ebbit1q:
I feel like we fixed some strange crashes that were caused by this strange behavior, thank you for noticing this and fixing it

ebbit1q:
I can't reproduce your initial problem

RickyRister:
Are you testing with local game or server game? It's more apprent in server games due to the network latency.

ebbit1q:
I tested it by using a local server, cutting the connection between the two, then sending the command, this does not cause the game to drop a player like in your video

RickyRister:
I didn't cut the connection. I just connected to rooster ranges and conceded normally.

ebbit1q:
ok, this is my bad, I cannot reproduce on the 2.10.3 release, which I compared with, following the same method I can easily reproduce this one the latest beta

ebbit1q (approved):

PR #6625, opened 2026-02-22
CI: Remove Windows 7 build

tooomm:
Windows 7 is out of support and super old.
With the new 2.10.3 stable release, we have a recent version for people who need it still.

- Remove build from CI and release notes


tooomm:
Removing Qt5 support sounds great!

PR #6627, opened 2026-02-22
Update vcpkg submodule

tooomm:
- Bump `vcpkg` submodule to latest release: [`2026.01.16`](https://github.com/microsoft/vcpkg/releases/tag/2026.01.16)


tooomm:
```
Run .ci/compile.sh --server --test --vcpkg
Set cache size limit to 500.0 MB
found QTDIR at /Users/runner/work/Cockatrice/Cockatrice/Qt/6.10.2/macos
cp: ../vcpkg/triplets/x64-osx.cmake: No such file or directory
```
The file moved from `triplets` to `triplets/community` starting with the 2025.12.12 release:
https://github.com/microsoft/vcpkg/blob/66c0373dc7fca549e5803087b9487edfe3aca0a1/triplets/community/x64-osx.cmake
Actual change did happen on Nov 19th: https://github.com/microsoft/vcpkg/commit/2e776305f2542cc626716435b0c1baf67bb5ab45#diff-acd4035354ff917fd4189724185f67b50049e4f237f7bdb577ec126d19eb8811

Last release with the old location is [2025.10.17](https://github.com/microsoft/vcpkg/releases/tag/2025.10.17).
Newer vcpkg versions do only have the `arm64-osx` triplet available as fully supported and tested.
See https://learn.microsoft.com/en-us/vcpkg/concepts/triplets.

ebbit1q (approved):
thanks, what would we have to do to update to newer versions from now?

tooomm:
> what would we have to do to update to newer versions from now?

Drop intel support on Macs, or stay on this vcpkg. 😄 There is no official support from vcpkg for this triplet anymore.
(Btw, Apple just announced that the upcoming macOS 27 will not support Intel architectures.)

If we consider the community file, then our script can be updated to search for that file in a different location instead.

PR #6631, opened 2026-02-24
Update UnescapedStringListPart to include parentheses

basicer:
`(o:a OR o:b)` incorrectly parses.

Parentheses will need to be quoted to be part of strings


RickyRister:
do we want to add unit tests for this?

RickyRister:
We also need to update the deck search expression

https://github.com/basicer/Cockatrice/blob/55d335e0661c28eca633e272a56ac6dc6f4dd0c0/cockatrice/src/filters/deck_filter_string.cpp#L34

PR #6632, opened 2026-02-24
CI: `windows-2025` runner

tooomm:
- Use newer GitHub Actions runner for windows.
  `windows-2025` is the new default runner since half a year already.
- Install NSIS manually which is missing in the new base image


tooomm:
> seems fine, I assume you tested it

Touché! 😉 
Yes, it was working for me.

ebbit1q:
excellent!

PR #6636, opened 2026-02-25
Add command zone, including companion and background compatibility

DawnFire42:
!!!Large feature - I tried my best to make clean commits to help with reviewing.

- Right now Cockatrice doesn't have a command zone and commander tax tracking is done through "other" counter. 

- Add a command zone, companion zone and background zone as options in game creation menu. 
- A new texture piece (derived from existing ones) for each themes to fill the command zone and others.
- Adds the existing game creation menu at the start of a local game (I needed it for troubleshooting).
- Logs for my added features.
- Contextual menu options for my added features.
- Modular (players keep only the tabs they need).
- Each zone individually minimise-able to let player keep a larger stack space by double-click or the contextual menu. 

- I also made playing multiple cards from the stack actually use the existing sorting methods (the one already used when playing a single card from the stack), there was a bug preventing it from happening when multiple cards were selected and played at once from the contextual menu, so while I was there I fixed it and now it does use it. Now this slipped when I restructured the commits and is still in this version...

## Depends on
- https://github.com/Cockatrice/Cockatrice/pull/6641

## Screenshots
Screenshot from 2026-02-25 15-11-37
Screenshot from 2026-02-25 15-16-29
Screenshot from 2026-02-25 15-17-21
Screenshot from 2026-02-25 15-17-54

### DIFF
--- a/cmake/FindQtRuntime.cmake
+++ b/cmake/FindQtRuntime.cmake
@@ -29,7 +29,7 @@ if(WITH_ORACLE)
   set(_ORACLE_NEEDED Concurrent Network Svg Widgets)
 endif()
 if(TEST)
-  set(_TEST_NEEDED Widgets)
+  set(_TEST_NEEDED Test Widgets)
 endif()
 
 set(REQUIRED_QT_COMPONENTS ${REQUIRED_QT_COMPONENTS} ${_SERVATRICE_NEEDED} ${_COCKATRICE_NEEDED} ${_ORACLE_NEEDED}


RickyRister:
Tried out the thing. Actually works as advertised!

https://github.com/user-attachments/assets/a2e66990-b9fd-41cb-b60f-60f8857e59c8

Unfortunately, we haven't decided on how we're going to design the command zone feature. We've been working towards adding support for arbitrary custom zones, but progress has been slow because we're still figuring out the exact design of the feature, especially around the UX.

I think the ability to add a zone above the stack is pretty cool! We might include that as part of the design for custom zone support.

However, I'm not a fan of having multiple of those zones tied to mtg-specific mechanics. Why do we need a companion zone *and* a partner zone *and* a background zone? What if wizards prints more alternative commander mechanics in the future? Not to mention, we're trying to make Cockatrice not tied to game logic if possible.

I noticed that you did also clean up and refactor and add extra abstractions to the code. Would it be possible extract those into separate PRs? It'll reduce the burden on the reviewers in the future, and it'll mean that we can start to get stuff merged in as we hash out the design of the new feature.

ebbit1q:
hi, thank you for taking the time to add this! however we would have really appreciated it if you were to have created a ticket first to discuss such changes, as it stands we'd really prefer to take a more flexible approach to adding such a feature, please see the discussion in #1740 which aims to offer users a schema to configure game zone behaviors such as this!

it's wonderful that you have put in so much effort to create this and I'd like to discuss with you how we'd like to go about adapting your pr while considering backwards compatibility and being able to support other card games, we can do this over discord or simply on a ticket here

DawnFire42:
Thank you very much for the quick reviews! I have fixed the two issue that were flagged and have extracted my refactoring in another PR on which this now depends. 

Its true that I did not have in mind other card games when I made the feature, but I gave myself this grace because I made it optional from the game creation menu. They are all individual options that the player can enable or not from the game creation menu. I also named the zones for MTG because this is why I want the feature, but it could definitely be more general if such a feature can accommodate multiple games. I originally added an auto increment feature for tax counters, but my friends rejected the automation (rightfully so since it does look like its part of the Cockatrice design approach), so I instead added the option to increment the tax counter while you play a card from the command zone using the contextual menu. Also, my toggle button is quite small and would not be allowed on a smartphone app for small screens, but I figured Cockatrice isn't on mobile (if so just remove the button, its all accessible through the contextual menu).

I'd be happy to discuss more about my design choices over discord, but bottom line is that I made it relatively complex to create a few good templates to start from. Command and partner zones are much more closely linked together and both have tax counters, Companion and Background zones keep multiple cards compatibility and don't have counters, etc...

Let me know what you think!

PR #6637, opened 2026-02-26
[TabGame] Journal dockWidget

RickyRister:

Players are allowed to keep notes during games, for example for keeping track of what known cards are in the opponent's hand. 

While users could just open a separate notepad program to keep track of that, it would be a QoL improvement if there was an easy way of keeping notes in cockatrice itself. Then the user would no longer need to tab out between Cockatrice and the text editor. Having the note be a dockWidget in TabGame means that the dock will always remain on top of Cockatrice.

- Create a new `notes` DockWidget in `TabGame`
  - Is simply just a `QPlainTextEdit` inside a `QDockWidget`
  - Update the rest of the class to add that widget.
- The default layout has the dock hidden

https://github.com/user-attachments/assets/62e4a052-8112-4dd4-bd81-de595d283776

## Screenshots

Screenshot 2026-02-26 at 1 21 24 AM

Screenshot 2026-02-26 at 1 21 56 AM

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -180,6 +180,8 @@ set(cockatrice_SOURCES
     src/interface/widgets/deck_editor/deck_list_style_proxy.cpp
     src/interface/widgets/deck_editor/deck_state_manager.cpp
     src/interface/widgets/deck_editor/printing_disabled_info_widget.cpp
+    src/interface/widgets/game/journal_widget.cpp
+    src/interface/widgets/game/journal_widget.h
     src/interface/widgets/general/background_sources.cpp
     src/interface/widgets/general/display/background_plate_widget.cpp
     src/interface/widgets/general/display/banner_widget.cpp
--- a/cockatrice/src/interface/widgets/game/journal_widget.cpp
+++ b/cockatrice/src/interface/widgets/game/journal_widget.cpp
@@ -1,0 +1,104 @@
+#include "journal_widget.h"
+
+#include "../../../client/settings/cache_settings.h"
+
+#include <QLayout>
+#include <QPlainTextEdit>
+#include <QPushButton>
+
+static const QString BACKUP_FILE_NAME = "journal.txt";
+
+JournalRestoreWidget::JournalRestoreWidget(QWidget *parent) : QWidget(parent)
+{
+    auto layout = new QVBoxLayout(this);
+    layout->setSpacing(3);
+    layout->setAlignment(Qt::AlignCenter);
+    setLayout(layout);
+
+    textLabel = new QLabel(this);
+    restoreButton = new QPushButton(this);
+    discardButton = new QPushButton(this);
+
+    layout->addWidget(textLabel);
+    layout->addWidget(restoreButton);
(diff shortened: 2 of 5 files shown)


ebbit1q:
this is really nice but I immediately think of the users that will complain that this dock is client side only, notes will disappear as soon as a client disconnects and will not return when rejoining the game, also, some cards will tell players to take notes in secret, adding this feature will confuse those players as there is no option to reveal these notes (not that our named tokens workaround is that good)

I feel like if we want notes in games they need to be either not connected to the game and clearly communicated to be separate or we need to add them to the protocol and manage them with the server.

RickyRister:
> notes will disappear as soon as a client disconnects

Oh yeah good point. I didn't consider that edge case.

Rather than sending the notes to the server, another workaround for this is to back up notes to disk on each edit (using the gameId as the key). Then if a player rejoins a game, we can just reload the notes from disk. But then that leads to the problem of how we clean up these notes, when it's possible for a player to leave a game and never rejoin. We would need to implement some sort of "garbage collection" for dangling notes.

> adding this feature will confuse those players as there is no option to reveal these notes

Is there a name I could rename the dock to that would make the functionality more obvious?

ebbit1q:
my suggestion, if we are not adding this to the server, is to simply have one notes file for the user and have it shared across all games and with an obvious "clear" button (with confirmation), this fixes the deletion issue and the persistance issue, for the sharing I'd simply recommend explicitly avoiding the word "note" or "notes" anywhere as that is the language used by the cards, perhaps we can instead call it "private journal" or something

tooomm:
I like the journal term. `Local Journal` could work well to stress that it's not on the server, shared or visible by others, too.

Pretty cool feature addition btw!
Ebbit's suggestion with a single file and a clear button sounds smart.

I would throw in an option to "Clear journal when joining a new game".

RickyRister:
The reason I'm reluctant to have a shared journal is because the dock is per-game but a player can join multiple games at once.

ebbit1q:
we'd just have to have the text edits edit the same document, it's not actually very complicated: https://doc.qt.io/qt-6/qtextedit.html#document-prop

ebbit1q:
I'd be opposed to the automatic clearing of the journal even optionally, this sounds like it would go wrong too often to be useful

ebbit1q:
what might be a good alternative to a big clear button is to do the opposite, show a clear journal by default but have a big restore button when there is an entry on disk

RickyRister:
I'm opposed to the widget having a big clear button for the simple fact that there just isn't much room. Like, if you have the widget in docked mode, there's already a bunch of other dockWidgets in the sidebar, so you really need to make use of all the space you can.

Adding a QButton to the layout necessarily forces it to deal with padding and stuff.

Users can just select all and delete if they want to clear the text.

> what might be a good alternative to a big clear button is to do the opposite, show a clear journal by default but have a big restore button when there is an entry on disk

Good idea.

ebbit1q (dismissed):
I like where you're going with this, however I'd like to focus on the next release, I'd appreciate your help in trying to fix a few of the bugs before publishing it, we will release features such as this one in a subsequent release asap

PR #6641, opened 2026-02-27
Refactor general utilities

DawnFire42:
## Related Ticket(s)
- Refactoring Z values and zone names, removing magic numbers.

- This became relevant as I made a "Command Zone" for MTG.

- hopefully nothing

### DIFF
--- a/cockatrice/src/game/board/abstract_card_drag_item.cpp
+++ b/cockatrice/src/game/board/abstract_card_drag_item.cpp
@@ -1,6 +1,7 @@
 #include "abstract_card_drag_item.h"
 
 #include "../../client/settings/cache_settings.h"
+#include "../z_values.h"
 
 #include <QCursor>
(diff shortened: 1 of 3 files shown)


RickyRister:
Are you intending to replace all usages of zone name strings with the constants in this PR, or are you leaving that for a future PR?

RickyRister:
There's still some magic z values in [ViewZoneWidget](https://github.com/DawnFire42/Cockatrice/blob/e2d5c7438d7b1b0eec52d2a166ef3b8f3360b3d4/cockatrice/src/game/zones/view_zone_widget.cpp#L50)

DawnFire42:
I should be replacing all usages of zone names at the same time, and I should also be removing all magic numbers from the Z-values. 
I may have missed some things when quickly extracting the refactors from my build. I'll put this back as a draft until I have time to go through this all over again.

PR #6646, opened 2026-03-01
nullcheck printing's set in home tab background art crop

### LINKED ISSUE #6639: Endless crash loop after app start when changing home tab background
**System Information:**

Client Version: 2.11.0-custom-a598068 (2026-02-24)
Build Architecture: 64-bit
Qt Version: 6.10.2
System Locale: en_DE
Install Mode: Standard

Selecting a different background for the home screen makes the app crash right after start:
Image

It's an endless loop and within the few seconds given after startup you can not manage to go to settings and change the option again.
I had to manually reset the entry in the `global.ini` file to break it and start/use Cockatrice again.

**Steps to reproduce:**
 - Go to Settings / Appearance
 - Change the "Home tab background source" to "Art crop of random card" & Save
 - Click on "Home"

ebbit1q:
## Related Ticket(s)
- Fixes #6639

when a card's set is disabled getting the printing on it will have no set info, a nullptr, when showing this the painter will try to use the card's set info to show the set and num, which do not exist, causing a crash. see linked ticket.

- nullcheck the printing in painter
- nullcheck the printing when setting (this means disabled sets will be excluded from the random rotation)

### DIFF
--- a/cockatrice/src/interface/widgets/general/home_widget.cpp
+++ b/cockatrice/src/interface/widgets/general/home_widget.cpp
@@ -139,7 +139,7 @@ void HomeWidget::updateRandomCard()
 
     connect(newCard.getCardPtr().data(), &CardInfo::pixmapUpdated, this, &HomeWidget::updateBackgroundProperties);
     backgroundSourceCard->setCard(newCard);
-    background = backgroundSourceCard->getBackground(size());
+    background = backgroundSourceCard->getBackground();
 }
 
 void HomeWidget::onBackgroundShuffleFrequencyChanged()


tooomm:
I can confirm this fixes my issue! Thank you for investigating, ebbit!

>after repeated attempts I can reproduce a crash, but only about 10% of the time, I simply change it back and forth between theme and art crop and it crashes eventually

I only have 1-2 sets enabled at any time as I only play Limited. So my chances to hit a card of a disabled set are like 100% 😄

ebbit1q:
I think for your usecase a change in the "get random card" function would be needed, as with 99% of cards disabled it will easily hit the 10 attempt limit that I added in this pr (without this limit it would have stalled until found, not an improvement)

I don't think this is at all important though, the random art crop just "not working" for people with lots of disabled sets is very niche and entirely acceptable to me, if this were merged I'd ticket it at low priority.

RickyRister:
Either rename this PR or split the crash fix from the ui fix.

cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -93,10 +94,15 @@ void HomeWidget::updateRandomCard()
...
         case BackgroundSources::RandomCardArt:
-            do {
-                newCard = CardDatabaseManager::query()->getRandomCard();
-            } while (newCard == backgroundSourceCard->getCard() &&
-                     newCard.getCardPtr()->getProperty("layout") != "normal");
+            for (int i = 0; i < 10; ++i) {
+                ExactCard tmpCard = CardDatabaseManager::query()->getRandomCard();
+                if (tmpCard != backgroundSourceCard->getCard() &&
+                    tmpCard.getCardPtr()->getProperty("layout") == "normal" &&
+                    tmpCard.getPrinting().getSet() != nullptr) {
```
RickyRister: Do we really need to entirely skip cards that don't have sets? We could still display the card, and just leave the text blank.

ebbit1q: cards that don't have sets come from a set that is disabled, it's likely the user does not want to see them

ebbit1q:
the 3 changes sorta were made at the same time while I was looking at the code, I split them later (and undid one change you'll never hear of), I can split them into 3 prs if you prefer.

RickyRister:
Yes, I’d prefer if they were split, but if it’s too much trouble then it’s fine too

cockatrice/src/interface/widgets/general/home_widget.cpp
RickyRister: Then maybe we should log a warning if we give up, so that we have something to look for if people complain that the home screen art isn’t working.

ebbit1q: sure

ebbit1q:
done

ebbit1q:
hey, I'm still writing your warning

ebbit1q:
I have added the warning but in my testing of the warning it showed that it actually does not automatically stop the timer on failure, creating a busy loop anyway, exactly what I wanted to avoid

so I fixed that too

cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -125,20 +139,15 @@ void HomeWidget::updateRandomCard()
 
     connect(newCard.getCardPtr().data(), &CardInfo::pixmapUpdated, this, &HomeWidget::updateBackgroundProperties);
     backgroundSourceCard->setCard(newCard);
-    background = backgroundSourceCard->getBackground();
-
-    if (SettingsCache::instance().getHomeTabBackgroundShuffleFrequency() <= 0) {
-        cardChangeTimer->stop();
-    }
+    background = backgroundSourceCard->getBackground(size());
```
ebbit1q:
```suggestion
    background = backgroundSourceCard->getBackground();
```

cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -84,6 +87,20 @@ void HomeWidget::loadBackgroundSourceDeck()
...
+    static constexpr int ATTEMPTS = 10;
+    for (int i = 0; i < ATTEMPTS; ++i) {
+        ExactCard tmpCard = CardDatabaseManager::query()->getRandomCard();
+        if (tmpCard != backgroundSourceCard->getCard() && tmpCard.getCardPtr()->getProperty("layout") == "normal" &&
+            tmpCard.getPrinting().getSet() != nullptr) {
+            newCard = tmpCard;
+            return;
+        }
+    }
+    qWarning() << "failed to set random card image after" << ATTEMPTS << "attempts";
```
RickyRister, line 101: Use `qCWarning`.

If that’s too much effort because we haven’t set up logging for this class yet, then we can do it in a follow-up PR.

ebbit1q: yeah, it's just a single statement in the class, I didn't see much benefit to making an entire category for just one warning

PR #6651, opened 2026-03-02
Refactor/z value constants

DawnFire42:
In preparation for adding a command zone, this refactor became relevant. Magic numbers were everywhere and understanding why the numbers were as they were was more tedious then necessary.

- Hopefully nothing.
- Z_values have been put together.

### DIFF
--- a/cockatrice/src/game/z_value_layer_manager.h
+++ b/cockatrice/src/game/z_value_layer_manager.h
@@ -102,7 +102,7 @@ constexpr qreal OVERLAY_BASE = 2000000000.0;
  * @param zValue The Z-value to validate
  * @return true if the Z-value is valid for a card
  */
-[[nodiscard]] inline constexpr bool isValidCardZValue(qreal zValue)
+[[nodiscard]] constexpr bool isValidCardZValue(qreal zValue)
 {
     return zValue >= 1.0 && zValue <= CARD_Z_VALUE_MAX;
 }
@@ -115,7 +115,7 @@ constexpr qreal OVERLAY_BASE = 2000000000.0;
  * @param zValue The Z-value to validate
  * @return true if the Z-value is valid for an overlay element
  */
-[[nodiscard]] inline constexpr bool isOverlayZValue(qreal zValue)
+[[nodiscard]] constexpr bool isOverlayZValue(qreal zValue)
 {
     return zValue >= OVERLAY_BASE;
 }
@@ -126,7 +126,7 @@ constexpr qreal OVERLAY_BASE = 2000000000.0;
  * @param offset Offset from OVERLAY_BASE (0-7 for current elements)
  * @return The absolute Z-value for the overlay element
  */
-[[nodiscard]] inline constexpr qreal overlayZValue(qreal offset)
+[[nodiscard]] constexpr qreal overlayZValue(qreal offset)
 {
     return OVERLAY_BASE + offset;
 }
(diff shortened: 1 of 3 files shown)


cockatrice/src/game/z_value_layer_manager.h
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * @brief Validates that a Z-value is within the valid card range.
+ *
+ * Cards should have Z-values between CARD_BASE (1.0) and CARD_Z_VALUE_MAX.
+ * Values outside this range may interfere with overlay rendering.
+ *
+ * @param zValue The Z-value to validate
+ * @return true if the Z-value is valid for a card
+ */
+[[nodiscard]] inline constexpr bool isValidCardZValue(qreal zValue)
```
RickyRister: linter is saying that `inline` is redundant with `constexpr`

cockatrice/src/game/z_values.h
```diff
@@ -0,0 +1,59 @@
...
+
+// Overlay layer Z-values for items that should appear above normal cards
+constexpr qreal HOVERED_CARD = ZValueLayerManager::overlayZValue(1.0);
+constexpr qreal ARROWS = ZValueLayerManager::overlayZValue(3.0);
+constexpr qreal ZONE_VIEW_WIDGET = ZValueLayerManager::overlayZValue(4.0);
+constexpr qreal DRAG_ITEM = ZValueLayerManager::overlayZValue(5.0);
+constexpr qreal DRAG_ITEM_CHILD = ZValueLayerManager::overlayZValue(6.0);
+constexpr qreal TOP_UI = ZValueLayerManager::overlayZValue(7.0);
+
+/// Compute Z-value for child drag items based on hotspot position.
```
RickyRister: Use `/**` for doc comment instead of `///`

cockatrice/src/game/board/abstract_card_item.cpp
RickyRister, line 85: `setRealZValue` is called with magic values in two places in [`TableZone`](https://github.com/DawnFire42/Cockatrice/blob/652ca041b498e7eee741fc0c4f3f0263fadb6dce/cockatrice/src/game/zones/table_zone.cpp#L170). Is there a way to extract that too?

DawnFire42:
I did the changes you recommended. Thanks for the quick review again!

cockatrice/src/game/z_values.h
```diff
@@ -0,0 +1,83 @@
...
+ * and cards to the right (higher X) render above cards to the left.
+ * This creates natural visual stacking for overlapping cards.
+ *
+ * @param x The X coordinate of the card position
+ * @param y The Y coordinate of the card position
+ * @return Z-value for the card's table position
+ */
+[[nodiscard]] constexpr qreal tableCardZValue(qreal x, qreal y)
+{
+    constexpr qreal CARD_HEIGHT_FOR_Z = 102.0;
```
RickyRister: We might want to extract this to a shared constant (since AbstractCardItem still keeps its own `CARD_HEIGHT` constant that is used everywhere else), but that can be done later.

RickyRister: Actually, if we don't want to deal with finding a shared place to put the `CARD_HEIGHT` constant, then I suggest that we change this function to take card height as a param, instead of of defining its own constant.

DawnFire42: I see the issue, though I've been advised against parameterizing it because no caller will ever pass a different value to it anyway. 
A shared header would be better over the long term. And it follows existing patterns in the codebase (example: graphics_item_type.h). 
I would simply create card_dimension.h and use the existing CARD_WIDTH and CARD_HEIGHT to keep it compatible with existing code.

Let me know what you think

DawnFire42: I've looked at it in more detail and I'll do the refactor for the card dimensions entirely and update the calls everywhere so that there's a single source of truth and a single way to call it. This could be its own PR but its simple enough so I'm commiting it here.

DawnFire42:
Actually, it might be better to have it as a separate PR if it creates conflicts.

RickyRister:
Sure, feel free to make a separate PR for card dimension constants.

PR #6654, opened 2026-03-02
CI: Upload artifact files directly

tooomm:
Workflow artifacts of our binaries to test new builds are always zipped before uploaded.
That makes use of our single file uploads more cumbersome.

- Upload binaries of our builds directly
- Windows Program Databases are still zipped as they contain various files.
  The naming scheme of that artifact is adjusted to match the binary filenames.

### DIFF
--- a/.ci/name_build.sh
+++ b/.ci/name_build.sh
@@ -35,15 +35,15 @@ if ! cd "$path"; then
 fi
 
 # set filename
-name="${file%.*}" # remove all after last .
-new_name="$name$SUFFIX."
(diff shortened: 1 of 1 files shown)


tooomm:
While updating the `name` variable which is now only used for the PDB's artifact name,
I realized that the "zipped file" part in the build naming script is unused as well:
https://github.com/Cockatrice/Cockatrice/blob/master/.ci/name_build.sh#L40-L43

Should I remove it?

ebbit1q:
yes, you can do it in a followup pr

PR #6655, opened 2026-03-03
NSIS: Add app icon to installation exe on Windows

tooomm:
Only default NSIS installer icon displayed.

- Display trice icon for installer on Windows

Our .ico file has several images/resolutions included, it takes a moment to load and display the (high res) icon.

Does the .dmg file on Mac, .deb on Ubuntu/Debian and .rpm on Fedora show our app icon already?

## Screenshots

top: with this change
below: before
image

image

### DIFF
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -26,6 +26,7 @@ Var PortableMode
 !define MUI_WELCOMEPAGE_TEXT "This wizard will guide you through the installation of Cockatrice.$\r$\n$\r$\nClick Next to continue."
 !define MUI_FINISHPAGE_RUN "$INSTDIR/cockatrice.exe"
 !define MUI_FINISHPAGE_RUN_TEXT "Run 'Cockatrice' now"
+!define MUI_ICON "${NSIS_SOURCE_PATH}\cockatrice\resources\appicon.ico"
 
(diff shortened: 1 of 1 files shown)


SlightlyCircuitous:
> Does the .dmg file on Mac, .deb on Ubuntu/Debian and .rpm on Fedora show our app icon already?

The .deb does not display our icon on Ubuntu or Debian.

ebbit1q:
I'm not sure if that is possible, or even desired, debian package files are not like .exe files and are used differently, they should have the appropriate icon for a debian package file in my opinion

tooomm:
>they should have the appropriate icon for a debian package file in my opinion

Is that the default behavior for other apps and e.g. deb/rpm packages on Linux distros?

<br>

How is it for the Mac .dmg file? Does that show the Cockatrice app icon?
Do other apps use their icon for the dmg packages?

SlightlyCircuitous:
> Is that the default behavior for other apps and e.g. deb/rpm packages on Linux distros?

Now that you mention it, I believe so, yes. The Discord .deb just shows the pink .deb icon on Ubuntu, anyway.

The .dmg just shows the default .dmg icon when I downloaded, but I'm not on a mac so maybe it's different there.

PR #6658, opened 2026-03-04
[Game] Refactor move cards from library actions

RickyRister:

Refactoring to prepare for adding actions for "move cards from top/bottom of library to grave/exile face-down".

- Extract shared logic in the `actMoveCardsFromTop` methods and `actMoveCardsFromBottom` methods into shared private method.

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -424,12 +424,12 @@ void PlayerActions::actMoveTopCardToExile()
 
 void PlayerActions::actMoveTopCardsToGrave()
 {
-    moveTopCardsTo("grave", "grave");
+    moveTopCardsTo("grave", tr("grave"));
 }
 
 void PlayerActions::actMoveTopCardsToExile()
 {
-    moveTopCardsTo("rfg", "exile");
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/player/player_actions.cpp
```diff
@@ -424,58 +424,38 @@ void PlayerActions::actMoveTopCardToExile()
...
+void PlayerActions::moveTopCardsTo(const QString &targetZone, const QString &zoneDisplayName)
 {
     const int maxCards = player->getDeckZone()->getCards().size();
     if (maxCards == 0) {
         return;
     }
 
     bool ok;
-    int number = QInputDialog::getInt(player->getGame()->getTab(), tr("Move top cards to exile"),
+    int number = QInputDialog::getInt(player->getGame()->getTab(), tr("Move top cards to %1").arg(zoneDisplayName),
```
ebbit1q, line 443: this makes "exile" no longer translatable!

please split this into 3 strings, one for each zone and one for unhandled zones

ebbit1q: alternatively, we can do a tr when calling, like `moveTopCardsTo("rfg", tr("exile"));`

PR #6661, opened 2026-03-04
[Game] Add face down versions of move cards from library actions

### LINKED ISSUE #6538: Support face-down cards in all public zones

RickyRister:
## Related Ticket(s)
- Fixes #6538
- Follow-up to #6602
- Requires #6658 to be merged first

- Add 4 new actions to the library menu: face down versions for the permutations of "Move top/bottom cards to grave/exile".
  - Located under the submenus
- also update shortcut settings with the new actions
  - No default shortcuts for them

https://github.com/user-attachments/assets/8ba2a8e3-3675-4f82-9f83-e8c27141ca4b

## Screenshots

Screenshot 2026-03-03 at 11 34 38 PM
Screenshot 2026-03-03 at 11 34 35 PM

Screenshot 2026-03-03 at 11 12 32 PM
Screenshot 2026-03-03 at 11 12 26 PM


cockatrice/src/client/settings/shortcuts_settings.h
```diff
@@ -598,11 +598,19 @@ class ShortcutsSettings : public QObject
         {"Player/aMoveTopCardsToGraveyard", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Graveyard (Multiple)"),
                                                         parseSequenceString("Alt+M"),
                                                         ShortcutGroup::Move_top)},
+        {"Player/aMoveTopCardsToGraveyardFaceDown",
+         ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Graveyard (Multiple), Face Down"),
```
ebbit1q, line 602: I think `Graveyard, Face Down (Multiple)` makes a little more sense, but honestly the entire series would benefit from a little more consistency

RickyRister: Since we sort the shortcuts alphabetically, using the `Graveyard, Face Down (Multiple)` order could cause the two Graveyard (Multiple) shortcuts to not get grouped together if there was some other Graveyard shortcut.

Screenshot 2026-03-04 at 5 30 18 PM

Right now they are grouped together because we don't have any other graveyard shortcut that comes between the two, but it could happen in the future.

PR #6662, opened 2026-03-04
[DeckEditor] Fix double-faced cards not importing correctly

RickyRister:
## Related Ticket(s)
- Fixes #6647
- Fixes bug introduced in #6306

## Short roundup of the initial problem

We inadvertently lost functionality when we made `DeckLoader` no longer extend `DeckList`, since `DeckLoader` was overriding the protected method `getCompleteCardName`, which `DeckList` was calling in its loading code.
We then ended up removing those methods in future cleanups since they weren't doing anything.

## What will change with this Pull Request?

- Reimplement `getCompleteCardName` in `DeckList`

Issue #6647, opened 2026-03-01
dual faced cards no longer find their front face through the plain text importer


ebbit1q:
**System Information:**

Client Version: 2.11.0-custom-2acf5e3 (2026-03-01)
Client Operating System: Nobara Linux 43 (KDE Plasma Desktop Edition)
Build Architecture: 64-bit
Qt Version: 6.10.1
System Locale: en_US
Install Mode: Standard

<details><summary>Debug Log:</summary>

</details>

_______________________________________________________________________________________

all dual faced cards, (mdfc, transform, flip), will no longer properly find their front face when imported as plain text

_______________________________________________________________________________________

**Steps to reproduce:**
 - open the deck editor
 - use import deck ( ctrl shift v )
 - input `Arlinn Kord // Arlinn, Embraced by the Moon`
 - click ok

result:
 - `Arlinn Kord // Arlinn, Embraced by the Moon` is now an unknown card in your deck

expected behavior:
- `Arlinn Kord` is now a creature in your deck, the backside is linked through the database

RickyRister:
Regression was introduced in [2.11.0-beta.33](https://github.com/Cockatrice/Cockatrice/releases/tag/2025-11-11-Development-2.11.0-beta.33)

RickyRister:
Regression was introduced in #6294

ebbit1q:
didn't expect that, the pr changes deck lists but not loading

RickyRister:
Looks like I was mistaken. The correct card does get loaded, but it's highlighted red because the legality checker wasn't implemented. I wasn't paying too much attention while bisecting, so I saw the red and just keyed it in as bad.

RickyRister:
Regression was actually introduced in #6306

PR #6664, opened 2026-03-05
[DeckList] Refactor load from plaintext to take normalizer as param

RickyRister:
## Related Ticket(s)
- Related to #6647

When we tried to do the fix in #6662, we realized that we that the easy fix would require us to include the `libcockatrice_card` in `libcockatrice_deck_list`, which we didn't want to do.

One way to work around that is to move the normalization code into a function that we pass as a param into the load method. Then we can define the function in a module that does include `libcockatrice_card`

- Change signature of `loadFromStream_Plain` to take a function of QString -> QString
- Move all the cardName normalization logic in `loadFromStream_Plain` into a new `CardNameNormalizer` functor object located in `libcockatrice_card`.
- Update usages of `loadFromStream_Plain` in application code to always pass the default-constructed `CardNameNormalizer`

Note that this doesn't fix the original bug. This is just a refactoring PR. The fix will be done in a follow-up PR.

### DIFF
--- a/libcockatrice_card/CMakeLists.txt
+++ b/libcockatrice_card/CMakeLists.txt
@@ -37,7 +37,7 @@ add_library(
   libcockatrice/card/database/parser/card_database_parser.cpp
   libcockatrice/card/database/parser/cockatrice_xml_3.cpp
   libcockatrice/card/database/parser/cockatrice_xml_4.cpp
-    libcockatrice/card/import/card_name_normalizer.cpp
+  libcockatrice/card/import/card_name_normalizer.cpp
   libcockatrice/card/printing/exact_card.cpp
   libcockatrice/card/printing/printing_info.cpp
   libcockatrice/card/relation/card_relation.cpp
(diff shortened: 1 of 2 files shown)


tests/loading_from_clipboard/loading_from_clipboard_test.cpp
```diff
@@ -128,30 +128,20 @@ TEST(LoadingFromClipboardTest, EdgeCaseTesting)
 
 
 // Mainboard (11 cards)
-Æther Adept
```
RickyRister, line 131: I had to remove scenarios from the unit test because I don't want to figure out how to configure cmake to import cockatrice app into this test.

If we still want to test to normalization logic, perhaps we can make a separate test class for `CardNameNoramlizer` as a follow-up

RickyRister: I moved `CardNameNormalizer` to `libcockatrice_card` and imported that into the test module, which solves this problem

libcockatrice_card/libcockatrice/card/import/card_name_normalizer.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    if (cardName.contains(reDoubleFacedMarker)) {
+        QStringList faces = cardName.split(reDoubleFacedMarker);
+        cardName = faces.first().trimmed();
+    }
+
+    // Remove unnecessary characters
+    cardName.remove(reBrace);
+    cardName.remove(reRoundBrace); // I'll be entirely honest here, these are split to accommodate just three cards
```
ebbit1q, line 35: I wonder if it still does that, after all the time since I wrote this comment

PR #6669, opened 2026-03-06
Add local game options

DawnFire42:
## Short roundup of the change
I needed the game options to also be used in local games in order to troubleshoot when creating a command zone. So this brings the existing game option window to local games.


cockatrice/src/game/local_game_options.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+#ifndef LOCAL_GAME_OPTIONS_H
+#define LOCAL_GAME_OPTIONS_H
+
+struct LocalGameOptions
+{
+    int numberPlayers = 1;
+    int startingLifeTotal = 20;
+    bool spectatorsSeeEverything = false;
```
ebbit1q: why would there be spectators in a local game?

cockatrice/src/game/local_game_options.h
```diff
@@ -0,0 +1,21 @@
...
+ * @ingroup Game
+ * @brief Configuration options for local games.
+ *
+ * This struct encapsulates all settings that can be configured when starting
+ * a local game, providing a clean interface that prevents parameter explosion
+ * as more options are added in the future.
+ */
+
+#ifndef LOCAL_GAME_OPTIONS_H
+#define LOCAL_GAME_OPTIONS_H
```
RickyRister: Why is this file located in `src/game`? Do you picture ever using this struct in game code? I feel like this struct is used in pre-game code rather than game code. 

It might more sense to just define this struct in `dlg_local_game_options.h`, if we don't see ourselves using it anywhere else.

cockatrice/src/game/local_game_options.h
DawnFire42: I mostly just imported the exact menu we get in multiplayer. I can remove it. 
I made this change for the sole purpose of troubleshooting, so I didn't spend much time thinking about if it needed specific changes. 
Let me know if there are any other options you would add or remove in this menu for local games.

cockatrice/src/game/local_game_options.h
DawnFire42: I don't see why it would need to be there either, I'll move it.

cockatrice/src/interface/widgets/dialogs/dlg_local_game_options.cpp
```diff
@@ -0,0 +1,83 @@
...
+    auto *mainLayout = new QVBoxLayout;
+    mainLayout->addWidget(generalGroupBox);
+    mainLayout->addWidget(gameSetupOptionsGroupBox);
+    mainLayout->addWidget(rememberSettingsCheckBox);
+    mainLayout->addWidget(buttonBox);
+    setLayout(mainLayout);
+
+    rememberSettingsCheckBox->setChecked(SettingsCache::instance().getRememberGameSettings());
+    if (rememberSettingsCheckBox->isChecked()) {
+        numberPlayersEdit->setValue(SettingsCache::instance().getMaxPlayers());
```
RickyRister: Is this property being shared by both local games and server games? 
I would prefer those properties to be separate so that I still default to 1 player when starting a local game while having 2 players saved in my server game configs.

RickyRister (changes requested):

cockatrice/src/interface/widgets/dialogs/dlg_local_game_options.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <QDialogButtonBox>
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QLabel>
+#include <QSpinBox>
+#include <QVBoxLayout>
+
+DlgLocalGameOptions::DlgLocalGameOptions(QWidget *parent) : QDialog(parent)
+{
+    numberPlayersLabel = new QLabel(tr("P&layers:"), this);
```
RickyRister: fyi, we stopped caring about `&` shortcuts in newer code.

DawnFire42: oh ok, do you prefer I remove those or is it ok if I leave them? ie: what's the convention you want to follow?

RickyRister: I prefer if you remove them

DawnFire42:
Apparently, the macOS 13-14-15 failures here are not tests failures per say. 

**Failure point**: "Install fat Qt 6.10.3" step in GitHub Actions workflow
**Error**: Failed to download checksum for the file '6.10.3-0-...-X86_64-ARM64.7z' from mirrors '['https://download.qt.io']
**Cause**: Transient infrastructure issue with Qt's download servers

I don't think I can do anything about this in the current PR, so I'm going to leave it as is and wait to rerun tests later.

RickyRister:
@DawnFire42 try rebasing

PR #6671, opened 2026-03-07
CI: Use new attest action

tooomm:
actions/attest-build-provenance is superseded by actions/attest

>As of version 4, actions/attest-build-provenance is simply a wrapper on top of [actions/attest](https://github.com/actions/attest).
>
>Existing applications may continue to use the attest-build-provenance action, but new implementations should use actions/attest instead. Please see the [actions/attest](https://github.com/actions/attest) repository for usage information.

- No change in result


ebbit1q (approved):
thanks

PR #6672, opened 2026-03-07
CI: Run debug builds only for PR's

tooomm:
Long build times and additional dev builds triggered on master.
For each tagged version that is released, the dedicated `Debug` macOS build on master is always failing for example: https://github.com/Cockatrice/Cockatrice/actions/runs/22741286102

- Only run `Debug` builds on PR's


ebbit1q:
they aren't useless, why is the mac one failing?

PR #6675, opened 2026-03-08
[DeckEditor] Fix undo/redo clearing legality

RickyRister:

Undo/redo resets the red highlighting on illegal cards.

https://github.com/user-attachments/assets/74c2fc5e-fe10-4c96-87a7-00342a1cc2cc

This is caused because `refreshCardFormatLegalities` doesn't get called if the deck is replaced through a method other than `setDecklist`.

- Refresh legality inside `rebuildTree`

https://github.com/user-attachments/assets/b74016d9-351a-406a-b53f-21b0e0f0b364


RickyRister:
Shouldn't be. If we're rebuilding the tree already, then adding a legality refresh to the end shouldn't result in any noticeable increases.

PR #6685, opened 2026-03-10
Add card selection counter

### LINKED ISSUE #212: Count selected items
A rather small feature, but can be of great help to count permanents: indicate how many cards are currently selected. This would make it easier and more accurate to tape the correct number of lands, count the number of attacking creatures, etc.

Maybe in case of creatures you could also show the sum of attacking/blocking power.

Thanks in advance

DawnFire42:
This adds a card counter close to the cursor in a selection box. 

Related to issue #212 : Count selected items

closes #212

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -202,6 +202,13 @@ jobs:
                   --ccache "$CCACHE_SIZE" $NO_CLIENT
           .ci/name_build.sh
 
+      # Delete used cache to emulate a cache update. See https://github.com/actions/cache/issues/342.
+      - name: Delete old compiler cache (ccache)
+        if: github.ref == 'refs/heads/master' && steps.ccache_restore.outputs.cache-hit && steps.build.outcome == 'success'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh cache delete --repo ${{ github.repository }} ${{ steps.ccache_restore.outputs.cache-primary-key }}
+
       - name: Save compiler cache (ccache)
         if: github.ref == 'refs/heads/master'
         uses: actions/cache/save@v5
@@ -437,6 +444,13 @@ jobs:
           DEVELOPER_DIR: '/Applications/Xcode_${{matrix.xcode}}.app/Contents/Developer'
           TARGET_MACOS_VERSION: ${{ matrix.override_target }}
         run: .ci/compile.sh --server --test --vcpkg
+
+      # Delete used cache to emulate a cache update. See https://github.com/actions/cache/issues/342.
+      - name: Delete old compiler cache (ccache)
+        if: github.ref == 'refs/heads/master' && matrix.use_ccache == 1 && steps.ccache_restore.outputs.cache-hit && steps.build.outcome == 'success'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh cache delete --repo ${{ github.repository }} ${{ steps.ccache_restore.outputs.cache-primary-key }}
 
       - name: Save compiler cache (ccache)
         if: github.ref == 'refs/heads/master' && matrix.use_ccache == 1
--- a/cockatrice/src/game/player/menu/card_menu.cpp
+++ b/cockatrice/src/game/player/menu/card_menu.cpp
@@ -113,32 +113,20 @@ CardMenu::CardMenu(Player *_player, const CardItem *_card, bool _shortcutsActive
         addAction(aSelectAll);
         addAction(aSelectColumn);
         addRelatedCardView();
-    } else if (writeableCard) {
-
+    } else {
         if (card->getZone()) {
             if (card->getZone()->getName() == ZoneNames::TABLE) {
-                createTableMenu();
+                createTableMenu(writeableCard);
             } else if (card->getZone()->getName() == ZoneNames::STACK) {
-                createStackMenu();
+                createStackMenu(writeableCard);
             } else if (card->getZone()->getName() == ZoneNames::EXILE ||
                        card->getZone()->getName() == ZoneNames::GRAVE) {
-                createGraveyardOrExileMenu();
+                createGraveyardOrExileMenu(writeableCard);
             } else {
-                createHandOrCustomZoneMenu();
+                createHandOrCustomZoneMenu(writeableCard);
             }
         } else {
-            addMenu(new MoveMenu(player));
-        }
-    } else {
-        if (card->getZone() && card->getZone()->getName() != ZoneNames::HAND) {
-            addAction(aDrawArrow);
-            addSeparator();
-            addRelatedCardView();
-            addRelatedCardActions();
-            addSeparator();
-            addAction(aClone);
-            addSeparator();
-            addAction(aSelectAll);
+            createZonelessMenu(writeableCard);
         }
     }
 }
@@ -154,11 +142,9 @@ void CardMenu::removePlayer(Player *playerToRemove)
     }
 }
 
-void CardMenu::createTableMenu()
+void CardMenu::createTableMenu(bool canModifyCard)
 {
     // Card is on the battlefield
-    bool canModifyCard = player->getPlayerInfo()->judge || card->getOwner() == player;
-
     if (!canModifyCard) {
         addRelatedCardView();
         addRelatedCardActions();
@@ -213,10 +199,8 @@ void CardMenu::createTableMenu()
     addMenu(mCardCounters);
 }
 
-void CardMenu::createStackMenu()
+void CardMenu::createStackMenu(bool canModifyCard)
 {
-    bool canModifyCard = player->getPlayerInfo()->judge || card->getOwner() == player;
-
     // Card is on the stack
     if (canModifyCard) {
         addAction(aAttach);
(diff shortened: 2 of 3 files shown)


RickyRister:
I like the addition. However, I find the label to be sort-of-hard to make out when there's a lot cards on screen. The label also feels awkwardly positioned and sort of distracting sometimes. If we're adding this, then it definitely needs to be behind an option. (We can add the option in a follow-up PR though)

Also, the text gets cut off at certain values.

Screenshot 2026-03-10 at 10 07 53 PM

RickyRister:
This counter only works if you have an active lasso. That means you can't make use of the counter if you want to ctrl click to deselect cards after selecting, or use the select all shortcut. 

I'm fine with this PR only being a start at implementing a card counter feature. However, I would prefer if our card counter implementation is ultimately compatible with all the other methods of selecting cards. (Since I use the shortcut quite often)

DawnFire42:
Thanks for the review! I'll polish the feature in this PR. I wouldn't want to merge something unfinished.
I'll try to find a place in a general player setting, but if you already know a spot you would put it let me know.

RickyRister:
> I'll try to find a place in a general player setting, but if you already know a spot you would put it let me know.

First instinct would be somewhere in `User Interface` -> `General interface settings`

DawnFire42:
> > I'll try to find a place in a general player setting, but if you already know a spot you would put it let me know.
> 
> First instinct would be somewhere in `User Interface` -> `General interface settings`

It's also where I found it best fitting. 
And the number now persists to the top right of screen when the selection is done in any other way.

RickyRister:
Oh nice, you made the label expand properly.

Screenshot 2026-03-12 at 11 47 45 PM (2)


However, I still find it hard to read when it's on top of stuff. Not sure how that's fixable though. Maybe make the background less transparent?

Screenshot 2026-03-12 at 11 56 48 PM (2)

RickyRister:
Fyi, the horizontal position of the floating label doesn't get updated when the window is resized. 

https://github.com/user-attachments/assets/8e4d2e68-36d4-40bb-a3b3-5556d108965f

Not something that needs to be fixed right now, but just pointing it out so we can keep track of it.

RickyRister:
> And the number now persists to the top right of screen when the selection is done in any other way.

The top right of the screen is a bit too out of the way. I feel like the bottom of the screen would be much more comfortable for me, since I'm already staring at my side of the board when selecting.

I mean, I suppose we could also make that configurable, but I don't want to push so many config options at first.

RickyRister:
I'm going to use it while playing my games for a bit and see how it feels to use.

DawnFire42:
> Fyi, the horizontal position of the floating label doesn't get updated when the window is resized.
> 
>  example.mp4 
> Not something that needs to be fixed right now, but just pointing it out so we can keep track of it.

interresting! I had not tried to do this when testing XD 
I'll see what I can do

DawnFire42:
> > And the number now persists to the top right of screen when the selection is done in any other way.
> 
> The top right of the screen is a bit too out of the way. I feel like the bottom of the screen would be much more comfortable for me, since I'm already staring at my side of the board when selecting.
> 
> I mean, I suppose we could also make that configurable, but I don't want to push so many config options at first.

True, at first I wanted to keep it with the cursor but if you're not using the cursor at all its a problem, so top right was the first other place I thought to put it. I don't see any reason why we would want it there specifically or in any other corners. Personally I don't like having it in the corner at all, I just have no better idea. 

What I can do simply is to make it an option. Players could decide which corner they want. 
But maybe somewhere specific on the board could be better, like bottom left of the battlefield or in one corner of your hand (I wouldn't want it in the hand but its an example).
How about top/bottom left/right of the zone in which the selection is done? It'd be visible even if someone wanted his library window to take his whole screen and selecting without using the cursor.

DawnFire42:
> However, I still find it hard to read when it's on top of stuff. Not sure how that's fixable though. Maybe make the background less transparent?
> 

I could, I could also make it a tad bigger, or yellow. I'll try some stuff

RickyRister:
> How about top/bottom left/right of the zone in which the selection is done?

Selections can span multiple zones. 

However, I do agree that having the label location be dynamically determined and context-dependant would lead to the best QoL. But it's also much harder to implement, since now you need to deal with figuring out where the zones are.

DawnFire42:
I found the visibility issue. It was appearing behind the lasso. We can see it much better now, I'm not even gonna test other colors, but the 14px is nice so I'm leaving that in. I reduced the width of the box to let it appear in smaller lasso since its very visible now anyway and I increased the font size. 

After testing I see how much better the bottom right corner is already. I don't think we need to go any further with its placement. This is simple enough and is always convenient to look at.

RickyRister:
You're right, it is much more readable now.

I'm reasonably happy with how it looks now.

https://github.com/user-attachments/assets/7b164757-fe31-49fe-aa17-93ed8662ae19

The corner label disappearing when I have an active lasso feels kinda annoying for me though.

Also, technically the lasso number and the corner number are tracking different things. Lasso number is the number of cards affected by the lasso (which could include cards that are being deselected, if you're using ctrl click), while corner number is the total number of selected cards in all zones.

https://github.com/user-attachments/assets/51a74481-2aad-4fff-8aeb-44f9e55a42d6

I wonder if it would be better to just make these two labels separate. As in, the corner label is a separate object, and doesn't hide itself when you have an active lasso. 

In that case, we might also want to have separate settings for each.

cockatrice/src/game/game_view.cpp
```diff
@@ -48,24 +83,99 @@ void GameView::updateSceneRect(const QRectF &rect)
...
     rubberBand->setGeometry(QRect(mapFromScene(selectionOrigin), QSize(0, 0)));
     rubberBand->show();
 }
 
-void GameView::resizeRubberBand(const QPointF &cursorPoint)
+void GameView::resizeRubberBand(const QPointF &cursorPoint, int selectedCount)
 {
-    if (rubberBand)
-        rubberBand->setGeometry(QRect(mapFromScene(selectionOrigin), cursorPoint.toPoint()).normalized());
+    constexpr int kLabelPaddingInPixels = 4;
```
RickyRister: nit: linter is complaining that "Variable 'kLabelPaddingInPixels' can be moved to inner scope"

DawnFire42:
There is value in having two separate counters so I made them entirely separate. And Linter should be happier now!

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -1122,21 +1124,257 @@ void SettingsCache::setClientVersion(const QString &_clientVersion)
...
-        "gs", "gt", "gu", "gw", "gy", "hk", "hm", "hn", "hr", "ht", "hu", "id", "ie", "il", "im", "in", "io", "iq",
-        "ir", "is", "it", "je", "jm", "jo", "jp", "ke", "kg", "kh", "ki", "km", "kn", "kp", "kr", "kw", "ky", "kz",
-        "la", "lb", "lc", "li", "lk", "lr", "ls", "lt", "lu", "lv", "ly", "ma", "mc", "md", "me", "mf", "mg", "mh",
-        "mk", "ml", "mm", "mn", "mo", "mp", "mq", "mr", "ms", "mt", "mu", "mv", "mw", "mx", "my", "mz", "na", "nc",
-        "ne", "nf", "ng", "ni", "nl", "no", "np", "nr", "nu", "nz", "om", "pa", "pe", "pf", "pg", "ph", "pk", "pl",
-        "pm", "pn", "pr", "ps", "pt", "pw", "py", "qa", "re", "ro", "rs", "ru", "rw", "sa", "sb", "sc", "sd", "se",
-        "sg", "sh", "si", "sj", "sk", "sl", "sm", "sn", "so", "sr", "ss", "st", "sv", "sx", "sy", "sz", "tc", "td",
-        "tf", "tg", "th", "tj", "tk", "tl", "tm", "tn", "to", "tr", "tt", "tv", "tw", "tz", "ua", "ug", "um", "us",
-        "uy", "uz", "va", "vc", "ve", "vg", "vi", "vn", "vu", "wf", "ws", "xk", "ye", "yt", "za", "zm", "zw"};
+    static QStringList countries = QStringList() << "ad"
```
RickyRister: Fix merge.

Or just rebase the PR so we don't have to keep fixing it.

DawnFire42: I will rebase.

cockatrice/src/game/game_view.cpp
```diff
@@ -48,24 +89,99 @@ void GameView::updateSceneRect(const QRectF &rect)
...
 
 void GameView::stopRubberBand()
 {
+    if (!rubberBand)
+        return;
+
+    rubberBandActive = false;
     rubberBand->hide();
+    dragCountLabel->hide();
+    updateTotalSelectionCount();
```
RickyRister: Do we still need to call this here after the change?

cockatrice/src/game/game_view.cpp
```diff
@@ -48,24 +89,99 @@ void GameView::updateSceneRect(const QRectF &rect)
 
 void GameView::startRubberBand(const QPointF &_selectionOrigin)
 {
+    if (!rubberBand)
+        return;
+
+    rubberBandActive = true;
```
RickyRister: do we still need this field after the change?

cockatrice/src/game/game_view.cpp
DawnFire42: good catch

cockatrice/src/game/game_view.cpp
DawnFire42: Its redundant, I removed it

RickyRister (approved):
oops, accidentally ran some github cli command. 

Anyways, LGTM


https://github.com/user-attachments/assets/5630f2b4-dff4-42d4-8d47-505e1b462bbb

RickyRister:
@ebbit1q can we merge this in now or are we waiting for after the next release to add this feature?

ebbit1q:
I'm on the fence about it, I said no more features until we release but this one seems kinda small and it's pretty useful

PR #6686, opened 2026-03-11
Refactor zone names

DawnFire42:
Zone names were made with magic strings which made it hard to work with. It became highly relevant to do for a command zone or any other zones to be implemented more easily.

- hopefully nothing
- all instances of zone names have been updated to use the centralized constants


DawnFire42:
Still the same pesky macOS test bug, but it should be good.

RickyRister (approved):
LGTM. Thanks for the cleanup

Issue #6690, opened 2026-03-12
slower build times on ci: ccache has low hit ratio due to stale cache entries


brunoalr:
**System Information:**

CI infra

<details><summary>Debug Log:</summary>

</details>

_______________________________________________________________________________________

I noticed that CI builds are taking quite some time.

Latest build on master as from the time of writing: https://github.com/Cockatrice/Cockatrice/actions/runs/23016526246
<img width="1408" height="1201" alt="Image" src="https://github.com/user-attachments/assets/78ee1306-e613-4972-8a33-abf3f727f36e" />

If my memory does not fail me, I remember linux builds taking less than 10 minutes each, now they are taking from 15 to ~30 minutes.

So I went to investigate the build logs and found out a few things:

1) Ccache hit ratio is really low - for example, on [ubuntu 24.04](https://github.com/Cockatrice/Cockatrice/actions/runs/23016526246/job/66841132234) it's around 2% only:
<details>

````
Run source .ci/docker.sh
running image:
Set cache size limit to 500.0 MB
Show ccache stats
Configure cmake
Build project
Show ccache stats again
  Cache directory:    /.ccache
  Config file:        /.ccache/ccache.conf
  System config file: /etc/ccache.conf
  Stats updated:      Thu Mar 12 18:26:27 2026
  Cacheable calls:    2722 / 2722 (100.0%)
    Hits:               55 / 2722 ( 2.02%) <<<<<< HERE
      Direct:            0 /   55 ( 0.00%)
      Preprocessed:     55 /   55 (100.0%)
    Misses:           2667 / 2722 (97.98%)
  Successful lookups:
    Direct:              0 / 2722 ( 0.00%)
    Preprocessed:       55 / 2722 ( 2.02%)
  Local storage:
    Cache size (GB):   0.5 /  0.5 (102.9%)
    Files:            4505
    Cleanups:          120
    Hits:               55 / 2722 ( 2.02%)
    Misses:           2667 / 2722 (97.98%)
    Reads:            5444
    Writes:           5389
Create package
renaming 'Cockatrice-2.11.0-custom-20ad9af.deb' to 'Cockatrice-2.11.0-custom-20ad9af-Ubuntu24.04.deb'
-rw-r--r-- 1 runner runner 24482048 Mar 12 18:26 build/Cockatrice-2.11.0-custom-20ad9af-Ubuntu24.04.deb
````

</details>

2) We get warnings about (GHA) cache failing to save

[...]

PR #6691, opened 2026-03-13
CI: Update compiler cache on hit

tooomm:
## Related Issues
- Fixing #6690

Updating/overwriting/replacing an existing cache is currently not supported (https://github.com/actions/cache/issues/342).

- Delete existing cache on hit via CLI, ensure that only happens when run on master branch
- In that case only, upload new cache to keep a recent one on main branch as base

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -202,11 +202,11 @@ jobs:
                   --ccache "$CCACHE_SIZE" $NO_CLIENT
           .ci/name_build.sh
 
-      # Delete the old cache on hit to emulate a cache update. See https://github.com/actions/cache/issues/342.
+      # Delete used cache to emulate a cache update. See https://github.com/actions/cache/issues/342.
       - name: Delete old compiler cache (ccache)
+        if: github.ref == 'refs/heads/master' && steps.ccache_restore.outputs.cache-hit && steps.build.outcome == 'success'
         env:
           GH_TOKEN: ${{ github.token }}
-        if: github.ref == 'refs/heads/master' && steps.ccache_restore.outputs.cache-hit
         run: gh cache delete --repo ${{ github.repository }} ${{ steps.ccache_restore.outputs.cache-primary-key }}
 
       - name: Save compiler cache (ccache)
@@ -445,11 +445,11 @@ jobs:
           TARGET_MACOS_VERSION: ${{ matrix.override_target }}
         run: .ci/compile.sh --server --test --vcpkg
 
-      # Delete the old cache on hit to emulate a cache update. See https://github.com/actions/cache/issues/342.
+      # Delete used cache to emulate a cache update. See https://github.com/actions/cache/issues/342.
       - name: Delete old compiler cache (ccache)
+        if: github.ref == 'refs/heads/master' && matrix.use_ccache == 1 && steps.ccache_restore.outputs.cache-hit && steps.build.outcome == 'success'
         env:
           GH_TOKEN: ${{ github.token }}
-        if: github.ref == 'refs/heads/master' && matrix.use_ccache == 1 && steps.ccache_restore.outputs.cache-hit
         run: gh cache delete --repo ${{ github.repository }} ${{ steps.ccache_restore.outputs.cache-primary-key }}
 
       - name: Save compiler cache (ccache)


brunoalr:
LGTM

.github/workflows/desktop-build.yml
```diff
@@ -202,6 +202,13 @@ jobs:
                   --ccache "$CCACHE_SIZE" $NO_CLIENT
           .ci/name_build.sh
 
+      # Delete the old cache on hit to emulate a cache update. See https://github.com/actions/cache/issues/342.
+      - name: Delete old compiler cache (ccache)
+        env:
+          GH_TOKEN: ${{ github.token }}
+        if: github.ref == 'refs/heads/master' && steps.ccache_restore.outputs.cache-hit
```
ebbit1q: can we keep the key order the same for steps to help readability?
so far we use: name, id, if, shell, env, run in that order

tooomm: Missed to adjust the workaround for our own structure. Nice little detail!

Did also add an extra check to ensure the build actually succeeded before deleting the old cache. So we do not upload a cache from a failing build.

tooomm:
Overview of results in `master` with current desktop workflow config ([`desktop-build.yml@8180d2e`](https://github.com/Cockatrice/Cockatrice/blob/8180d2e3b07d45f8c9d65d9c9602bc26d9fe9396/.github/workflows/desktop-build.yml)) after this merge:

~~Source: https://github.com/Cockatrice/Cockatrice/actions/runs/23089598852~~
Source after #6703 and #6711: https://github.com/Cockatrice/Cockatrice/actions/runs/23169233329

| CI Build | Cold Cache | Warm Cache | Build Type |  | MB | Note |
|----------|-----------:|-----------:|------------|--|---:|------|
| **Arch** | 20 min | 4 min | Debug | ✔️ | 530 |  |
| **Debian 11** | 20 min | 5 min | Debug, Release | ✔️ | 390 | Mixed cache |
| **Debian 11 (Servatrice)** | 4 min | 1 min | Release | ✔️ | 12 |  |
| **Debian 12** | 14 min | 2 min | Release | ✔️ | 100 |  |
| **Debian 13** | 29 min | 5 min | Debug, Release | ✔️ | 500 | Mixed cache |
| **Fedora 42** | 15 min | 3 min | Release | ✔️ | 140 |  |
| **Fedora 43** | 29 min | 4 min | Debug, Release | ✔️ | 520 | Mixed cache |
| **Ubuntu 22.04** | 24 min | 6 min | Debug, Release | ✔️ | 320 | Mixed cache |
| **Ubuntu 24.04** | 26 min | 4 min | Debug, Release | ✔️ | 510 | Mixed cache |
| **macOS 13 Intel** | 26 min | 11 min | Release | ✔️ | 240 |  |
| **macOS 14** | 16 min | 4 min | Release | ✔️ | 220 |  |
| **macOS 15** | 20 min | 6 min | Release | ✔️ | 220 |  |
| **macOS 15 Debug** | 19 min | 5 min | Debug | ✔️ | 540 |  |
| **Windows 10** | 31 min | 30 min | Release | ❌ | - | No cache used yet |

Build times are drastically reduced down to only ~15%-40% of initial run time!
The new main cache from master is already used as base in new PR's and build times are down there, too (but not overriding master). 👍 

**Observations:**
- ~~Cache for Arch not working~~ Solved in #6703 and #6711
- Cache size of `Debug` builds is significantly higher than for `Release` ones
- Several builds run always `Debug` and `Release` builds next to each other and the cache is polluted/used from both
- ~~Ubuntu 24.04 is hitting the cache limit of 500MB, some others are close~~ Increased in #6703
- Overall cache usage just for all compiler caches (without Windows, and without additional tool caches for vcpkg, Docker, Qt etc.) is already ~4GB out of the available 10GB.

brunoalr:
awesome :)

what's the issue with Arch?

brunoalr:
> Several builds run always Debug and Release builds next to each other and the cache is polluted/used from both

good point! IMO this is worth looking into. can you open a ticket, please?

> Overall cache usage just for all compiler caches (without Windows, and without additional tool caches for vcpkg, Docker, Qt etc.) is already ~4GB out of the available 10GB.

Now we only have entries for master, right? With size limits I guess it's fine - we can tune the limits on a per-config basis if needed (e.g. macos 15 debug is 300MB, servatrice on Debian is 100MB etc).

We can relieve GHA cache pressure by migrating vcpkg caches to NuGet feed using Github Packages instead of GHA caches. Last time I checked, it was free for OSS projects. I experimented with it a bit on my fork:
https://github.com/brunoalr/Cockatrice/pull/32 based on
https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=windows-runner

tooomm:
> what's the issue with Arch?

In all Linux jobs, we have two build steps, but only the `Release` one is having an id.
Me adding a checkpoint to be super save and only deleting caches if a build was successful was causing the issue for Arch which is `Debug` only.

<br>

>>Several builds run always Debug and Release builds next to each other and the cache is polluted/used from both
>>
>good point! IMO this is worth looking into. can you open a ticket, please?

Let's keep the note here for now, I'll come back and update the table anyway.
There is already many PR's and topics ongoing around caches and our CI pipeline.
But we should definitely check this later! Ninja and caching on Windows first, hehe.

<br>

>>Overall cache usage just for all compiler caches (without Windows, and without additional tool caches for vcpkg, Docker, Qt etc.) is already ~4GB out of the available 10GB.
>>
>Now we only have entries for master, right? With size limits I guess it's fine - we can tune the limits on a per-config basis if needed (e.g. macos 15 debug is 300MB, servatrice on Debian is 100MB etc).

Yes, our caching strategy (for compiler cache) is currently to have master as our main cache source.
PR's will still use it and benefit, but do not update the main cache unless merged into master.
That way we do not have 10 different caches for each OS and we do not constantly overrun the 10GB storage limit which in return would constantly evict good caches.
Clean and simple.
For other caches, some are saved from PR's already which is not ideal.

Individual limits per CI job are not changing how much MB are uploaded.
Servatrice would still only upload 12MB, no matter if the limit is 100 or 500?
But it adds complexity to our config and is trickier to maintain. I would rather avoid that.

>We can relieve GHA cache pressure by migrating vcpkg caches to NuGet feed using Github Packages instead of GHA caches. Last time I checked, it was free for OSS projects. I experimented with it a bit on my fork:
https://github.com/brunoalr/Cockatrice/pull/32 based on
https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=windows-runner

I stumbled upon this option when vcpkg changed their binary caching approach and all setup and caching actions had issues, too. The docs seem now also better!
I hope we do not need to change though and can find a way to make the 10GB work. 🙈 
It's an option for sure if really needed!

PR #6694, opened 2026-03-13
CI: sccache

tooomm:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


brunoalr:
I have been re-experimenting windows + ninja + sccache on my fork. this time with the help of some LLMs (mainly Composer 1.5 from Cursor). The PR commit history is a messy and currently relies on my own commit(s) that fix stale caches being on master: https://github.com/brunoalr/Cockatrice/pull/53/changes

On vcpkg cache hit, build time went from 30mins to 10mins. I did some manual testing and macos still works and win works fine as well.

The sad part is that I currently don't fully understand why all of the changes were needed (apparently some paths had to be fixed on cmake, extra logic had to be used in the build script etc), but I guess I understand ~85% of it.

You can take a look if you want to check if it makes sense.

Anyhow, I will rebase and clean it up after #6691 is merged

PR #6696, opened 2026-03-14
refactor: extract AbstractZoneMenu interface for polymorphic zone menu dispatch

DawnFire42:
Replaces scattered null-check-then-call patterns for retranslateUi() and shortcut activation in PlayerMenu with a single loop over a QVector<AbstractZoneMenu*>, populated at construction time. Uses non-QObject interface to avoid Qt MOC diamond inheritance.

This became useful as I was implementing proper menus for my command zone project.

- Hopefully nothing

### DIFF
--- a/cockatrice/src/game/player/menu/player_menu.cpp
+++ b/cockatrice/src/game/player/menu/player_menu.cpp
@@ -15,33 +15,24 @@ PlayerMenu::PlayerMenu(Player *_player) : player(_player)
     playerMenu = new TearOffMenu();
 
     if (player->getPlayerInfo()->getLocalOrJudge()) {
-        handMenu = new HandMenu(player, player->getPlayerActions(), playerMenu);
-        playerMenu->addMenu(handMenu);
-
-        libraryMenu = new LibraryMenu(player, playerMenu);
-        playerMenu->addMenu(libraryMenu);
+        handMenu = addManagedMenu<HandMenu>(player, player->getPlayerActions(), playerMenu);
+        libraryMenu = addManagedMenu<LibraryMenu>(player, playerMenu);
     } else {
         handMenu = nullptr;
         libraryMenu = nullptr;
     }
 
-    graveMenu = new GraveyardMenu(player, playerMenu);
-    playerMenu->addMenu(graveMenu);
-
-    rfgMenu = new RfgMenu(player, playerMenu);
-    playerMenu->addMenu(rfgMenu);
+    graveMenu = addManagedMenu<GraveyardMenu>(player, playerMenu);
+    rfgMenu = addManagedMenu<RfgMenu>(player, playerMenu);
 
     if (player->getPlayerInfo()->getLocalOrJudge()) {
-        sideboardMenu = new SideboardMenu(player, playerMenu);
-        playerMenu->addMenu(sideboardMenu);
-
-        customZonesMenu = new CustomZoneMenu(player);
-        playerMenu->addMenu(customZonesMenu);
+        sideboardMenu = addManagedMenu<SideboardMenu>(player, playerMenu);
+        customZonesMenu = addManagedMenu<CustomZoneMenu>(player);
         playerMenu->addSeparator();
 
         countersMenu = playerMenu->addMenu(QString());
 
-        utilityMenu = new UtilityMenu(player, playerMenu);
+        utilityMenu = createManagedComponent<UtilityMenu>(player, playerMenu);
     } else {
         sideboardMenu = nullptr;
         customZonesMenu = nullptr;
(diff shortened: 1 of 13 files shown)


cockatrice/src/game/player/menu/player_menu.h
```diff
@@ -20,6 +20,7 @@
 
 #include <QMenu>
 #include <QObject>
+#include <QVector>
```
RickyRister: In Qt6+, `QVector` is [just an alias](https://doc.qt.io/qt-6/qvector.html) for `QList`.

Seeing as we're trying to drop support for Qt5 soon, I think it would be better if we only use `QList` in newer code.

cockatrice/src/game/player/menu/player_menu.cpp
```diff
@@ -17,31 +17,38 @@ PlayerMenu::PlayerMenu(Player *_player) : player(_player)
     if (player->getPlayerInfo()->getLocalOrJudge()) {
         handMenu = new HandMenu(player, player->getPlayerActions(), playerMenu);
         playerMenu->addMenu(handMenu);
+        allZoneMenus.append(handMenu);
```
RickyRister: I don't love how you have to manually keep the list and the actual menus in sync. I suppose there isn't a better way to do it?

DawnFire42: fair enough, I'll try to find a better solution

cockatrice/src/game/player/menu/player_menu.cpp
```diff
@@ -99,37 +106,19 @@ void PlayerMenu::retranslateUi()
...
     QMapIterator<int, AbstractCounter *> counterIterator(player->getCounters());
     while (counterIterator.hasNext()) {
         counterIterator.next().value()->retranslateUi();
     }
 
-    if (utilityMenu) {
-        utilityMenu->retranslateUi();
-    }
-
     if (sayMenu) {
```
RickyRister: is there a reason we can't make sayMenu also implement `AbstractZoneMenu`? It clearly isn't because "say isn't a zone" because you've already implemented it for `UtilityMenu`.

cockatrice/src/game/player/menu/player_menu.cpp
RickyRister: I'm fine with the idea of the `AbstractZoneMenu` interface. (although you might want to make the name more generic, since I feel like that interface can apply to many of the in-game menus.) 
We use the shortcut enabling/disabling behavior in order to make shortcuts be used by active player in multi-player local games.

DawnFire42:
This ended up being a bigger fish than expected.
I attempted deep frying.

Let me know if it was effective

cockatrice/src/game/player/menu/player_menu.h
```diff
@@ -82,9 +83,23 @@ private slots:
     SayMenu *sayMenu;
     CustomZoneMenu *customZonesMenu;
 
-    bool shortcutsActive;
+    QList<AbstractPlayerComponent *> managedComponents;
+    bool shortcutsActive = false;
 
-    void initSayMenu();
+    template <typename MenuT, typename... Args> MenuT *addManagedMenu(Args &&...args)
```
RickyRister, line 94: leave comments on these methods. It's not obvious at first glance how we need to keep everything in sync.

RickyRister:
fyi, git best practices says you should [leave a blank line between the commit title and commit description](https://git-scm.com/docs/git-commit#_discussion). 

Github seems to trim the title at the first newline just fine, but git cli (and CLion) will concat together all lines until the first empty line into the commit title, which causes a massive mess.

Screenshot 2026-03-16 at 8 32 54 PM

DawnFire42:
I updated the documentation and fixed the commit message

cockatrice/src/game/player/menu/player_menu.h
```diff
@@ -82,9 +84,26 @@ private slots:
     SayMenu *sayMenu;
     CustomZoneMenu *customZonesMenu;
 
-    bool shortcutsActive;
+    // Drives AbstractPlayerComponent lifecycle delegation. Counters are iterated separately via player->getCounters().
+    QList<AbstractPlayerComponent *> managedComponents;
+    bool shortcutsActive = false;
 
-    void initSayMenu();
+    // Creates component, adds it as a submenu of playerMenu, and registers in managedComponents.
```
RickyRister: sorry, by comments, I meant doc comments. You can just use `///` if you don't feel like breaking out the `/**`.

DawnFire42:
Was there anything left to change in this PR?

RickyRister:
Use doc comments instead of normal comments in the place I pointed out

PR #6700, opened 2026-03-14
CI: Resolve to latest Qt version in range for Windows as well

tooomm:
After merging the `macOS` and `Windows` configurations after both use "vcpkg" for dependencies, Qt is still handled/installed differently on each OS.
- For macOS, there was a specific step before the Qt action run to resolve to the latest, full version number before installing exactly that version.
- For Windows, the Qt action was looking for the newest patch version itself.

That leads to the cache being saved with the wildcard in the name/key for Windows
(~`cache-qt-6.10.2` vs. ~`cache-qt-6.10.*`).
For new version releases (e.g. Qt 6.10.3), macOS was using them right away 6.10.3 =/ 6.10.2.
On Windows however, 6.10.* matches 6.10.* and the cached 6.10.2 was still used and never updated.

- Reuse the logic and script from macOS for Windows as well to resolve wildcards to the latest patch release
  (6.10.* --> 6.10.2/3)
    - Install aqt on Windows
    - Query for dedicated builds available for mac or windows via aqt
- Fix stale cache, add the full Qt version to the cache entry and use new versions with latest fixes on Windows once they are released, too

### DIFF
--- a/.ci/resolve_latest_aqt_qt_version.sh
+++ b/.ci/resolve_latest_aqt_qt_version.sh
@@ -27,7 +27,16 @@ if ! hash aqt; then
 fi
 
 # Resolve latest patch
-if ! qt_resolved=$(aqt list-qt mac desktop --spec "$qt_spec" --latest-version); then
+if [[ $RUNNER_OS == macOS ]]; then
+  if ! qt_resolved=$(aqt list-qt mac desktop --spec "$qt_spec" --latest-version); then
+    exit 1
+  fi
+elif [[ $RUNNER_OS == Windows ]]; then
+  if ! qt_resolved=$(aqt list-qt windows desktop --spec "$qt_spec" --latest-version); then
+    exit 1
+  fi
+else
+  echo "aqt command for $RUNNER_OS not defined."
   exit 1
 fi
 


.ci/resolve_latest_aqt_qt_version.sh
```diff
@@ -27,7 +27,16 @@ if ! hash aqt; then
 fi
 
 # Resolve latest patch
-if ! qt_resolved=$(aqt list-qt mac desktop --spec "$qt_spec" --latest-version); then
+if [[ $RUNNER_OS == macOS ]]; then
+  if ! qt_resolved=$(aqt list-qt mac desktop --spec "$qt_spec" --latest-version); then
+    exit 1
+  fi
+elif [[ $RUNNER_OS == Windows ]]; then
+  if ! qt_resolved=$(aqt list-qt windows desktop --spec "$qt_spec" --latest-version); then
```
brunoalr, line 35: nit: couldn't we avoid repetition by creating a variable to store only "mac" or "windows" and passing it to the `aqt` invocation instead?

.github/workflows/desktop-build.yml
```diff
@@ -386,10 +383,10 @@ jobs:
         if: matrix.os == 'macOS' && steps.restore_qt.outputs.cache-hit != 'true'
         uses: jurplel/install-qt-action@v4
         with:
-          cache: false
           version: ${{ steps.resolve_qt_version.outputs.version }}
           arch: ${{matrix.qt_arch}}
           modules: ${{matrix.qt_modules}}
+          cache: false
```
brunoalr, line 389: do we have any specific ordering scheme? I wanna avoid breaking it myself in the future

ebbit1q: no, not really, just try to keep them consistent in the file to help with reading

.ci/resolve_latest_aqt_qt_version.sh
tooomm: I went the simple route here.

It could probably be done in a different way too to combine the commands.
It might make the input to the script a bit more complicated though or the CI build matrix would need to contain this information - just moving the information.

But I'm not against rethinking this if you have a good proposal that keeps simplicity and avoids repetition.

The build stuff is already complicated and shattered enough. 🙈

.github/workflows/desktop-build.yml
tooomm: Jep, I tried to fix & match the order with the second `jurplel/install-qt-action` step 20 lines down in the same file here, @brunoalr.

PR #6701, opened 2026-03-14
refactor: extract shared card insertion algorithm from hand/stack zones

DawnFire42:
Hand and stack zones had near-identical addCardImpl() implementations, differing only in whether resetState() preserves annotations. Extract the shared pattern into a template function (CardZoneAlgorithms::addCardToList) to eliminate duplication and enable isolated testing without Qt dependencies. Pile, table, and zone-view logic are intentionally excluded — their post-add behavior (signals, coordinate placement, hidden cards) is materially different.

This wasn't an obvious thing before I made my command zone feature. Hopefully this will facilitate integration of different zones in the future.

### DIFF
--- a/cockatrice/src/game/zones/logic/card_zone_algorithms.h
+++ b/cockatrice/src/game/zones/logic/card_zone_algorithms.h
@@ -1,0 +1,45 @@
+#ifndef COCKATRICE_CARD_ZONE_ALGORITHMS_H
+#define COCKATRICE_CARD_ZONE_ALGORITHMS_H
+
+namespace CardZoneAlgorithms
+{
+
+/**
+ * Shared insertion logic for zones where cards become visible on add and follow
+ * the standard pattern: clamp index, insert, clear identity if contents unknown,
+ * reset state, show card.
+ *
+ * Zones with different post-add behavior (signal connections, positional resets,
+ * hidden cards, or coordinate-based placement) should NOT use this — implement
+ * addCardImpl directly instead.
+ *
+ * Template parameters allow testing with lightweight mocks that avoid Qt graphics
+ * dependencies.
+ *
+ * @tparam CardList Must provide: size() -> int, insert(int, CardType*),
+ *                  getContentsKnown() -> bool
+ * @tparam CardType Must provide: setId(int), setCardRef(CardRefType),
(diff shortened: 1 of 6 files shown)


cockatrice/src/game/zones/logic/add_card_algorithm.h
```diff
@@ -0,0 +1,45 @@
+#ifndef COCKATRICE_ADD_CARD_ALGORITHM_H
+#define COCKATRICE_ADD_CARD_ALGORITHM_H
```
RickyRister: Do we need a file just for add card algorithms? Are we planning to add more `add card algorithms` in the future? Otherwise we could just name this `card_zone_algorithm` or something like that and put all the algorithms in here.

cockatrice/src/game/zones/logic/card_zone_algorithms.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @tparam CardList Must provide: size() -> int, insert(int, CardType*),
+ *                  getContentsKnown() -> bool
+ * @tparam CardType Must provide: setId(int), setCardRef(CardRefType),
+ *                  resetState(bool), setVisible(bool)
+ * @param keepAnnotations Forwarded to card->resetState(). Stack-like zones preserve
+ *                        annotations across zone transitions; hand-like zones clear them.
+ */
+template <typename CardList, typename CardType>
+void addCardToList(CardList &cards, CardType *card, int x, bool keepAnnotations)
```
RickyRister, line 27: Might want to rename this function to be less general, since it's only intended to be used for certain zones (as you noted in the docs).

Otherwise, LGTM

DawnFire42: So far every alternative I have thought of either created confusion or was too verbose. I honestly think that this name describes well what it does. Do you have a better name in mind?

RickyRister: Can't think of anything off the top of my head either. 

I suppose this name is fine for now. If we end up creating more functions for adding cards to lists, then we can just rename it then.

RickyRister (approved):

PR #6702, opened 2026-03-14
chore: add script to enable running unsigned macOS builds from CI locally

brunoalr:
## Related Ticket(s)
Likely related to https://github.com/Cockatrice/Cockatrice/pull/6678

macOS artifacts are not signed on PRs, only on master. This makes it hard to test build artifacts locally.

macOS is very (rightfully so) defensive against unsigned apps. It simply blocks them without a clear warning to the user other than a message saying something like "the app is corrupted" or whatever.

Removing the apps from quarantine are not enough(e.g. `xattr -dr com.apple.quarantine /Applications/cockatrice.app`), all executables under the *.app folder must be signed in order to pass macOS's Gatekeeper checks.

- Create a new `utils` folder meant to contain helper scripts that are meant to be used locally only
- Create a script that removes the apps from quarantine, sign all it's executable and then sign the whole bundle


brunoalr:
@tooomm wanna take a look at this? it looks like https://github.com/Cockatrice/Cockatrice/pull/6678 made it a bit harder to test artifacts from unmerged PRs.

RickyRister:
Oh nice.

I still have `forcesign() {find "$1" -name "*" -execdir sudo codesign --force --deep --sign - {} \;}` defined in my bash aliases for if I randomly need to build an app on github from source with cmake.

brunoalr:
if this gets merged, I guess we need to add a section to some wiki page to tell people about it

ebbit1q:
can't we sign them like we do on release? is that not possible?

brunoalr:
> can't we sign them like we do on release? is that not possible?

from a technical point of view, I think it is possible to sign them.

from a security point of view, I'm not sure that's a good idea. if we do so, basically we are saying that build artifacts from any PR are legit. I assume that was the main concern for https://github.com/Cockatrice/Cockatrice/pull/6678. Is that correct, @tooomm ?

ebbit1q:
the attack vector is extremely slim on this imo, note that actions will only run for pull requests by contributors or when accorded by a member for first time contributors

tooomm:
> can't we sign them like we do on release? is that not possible?

I tried to sign the Mac binaries, you can try it here: 
https://github.com/Cockatrice/Cockatrice/actions/runs/23411128665?pr=6710#artifacts

Not sure how the "Gatekeeper" will flag those and what's the impact for users for unsigned vs. signed. The topic is not really clear to me. Apple's docs help not too much either and you find different comments.

PR #6703, opened 2026-03-14
CI: Adjustments to ccache usage

tooomm:
## Related Ticket(s)
- Follow up to #6691

- Ubuntu 24.04 job was hitting the 500MB ccache size limit.
- Arch ccache was not updated correctly.

See https://github.com/Cockatrice/Cockatrice/pull/6691#issuecomment-4061280470 for some data.

- Increase ccache size to 550MB
- Fix cache update for compiler cache on Arch
- Small cleanup and uniformization

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -153,7 +153,7 @@ jobs:
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
     env:
       NAME: ${{matrix.distro}}${{matrix.version}}
-      CACHE: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}   # directory for caching docker image and ccache
+      CACHE_DIR: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}   # directory for caching docker image and ccache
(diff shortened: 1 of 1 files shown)


ebbit1q:
if it was hitting the limit by the way, does that mean it incurred more cache misses or was it actually stale entries that filled the cache?

tooomm:
> if it was hitting the limit by the way, does that mean it incurred more cache misses or was it actually stale entries that filled the cache?

We were not able to save the complete compiler cache for Ubuntu 24.04, leading to fewer ccache hits during compilation. Let's see if this helps.
Ccache stats were printing ~105% cache usage and the cache file at GH was listed with 500MB. We should now have enough space to store the additional ~5%.

However, the cache directory change here breaks the cache as it's on top used and set in the docker script building the base layer for each Linux distro. Fixed in #6711.
Our compilation setup is too layered and complex... 🙄

ebbit1q:
I feel like that's my fault, I'm the one that made it layered and complex

ebbit1q:
maybe I should make better documentation on it, we have the docker script which builds a docker image for the right distro and exports a function that will run a command in that image, we then use that with the compile script to compile, both scripts take combinations of arguments and environment variables

PR #6704, opened 2026-03-14
chore: add GHA cache usage script and workflow

brunoalr:
## Related Ticket(s)
- Creation was inspired by comment in #6691

We are hitting the 10GB limit of GHA caches.

- Add a script that uses the `gh` client to get cache usage grouped by prefixes (e.g. ccache, vcpkg)
- The script can be run locally. e.g.: `.ci/gha_cache_usage.sh` (as long as `gh` is authenticated, for example via `gh auth login`)
- Add a workflow that displays formatted results in a table

## Screenshots
https://github.com/brunoalr/Cockatrice/actions/runs/23097226664
image
Note: this is the usage from my fork, not this upstream repo.

## Disclaimer
I don't really know `jq` syntax. This is vibe coded.

### DIFF
--- a/.ci/gha_cache_usage.sh
+++ b/.ci/gha_cache_usage.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Reports GitHub Actions cache usage, grouped by key prefix (first segment before "-" or "/") or for one prefix.
 #
-# Usage: .ci/gh_buildkit_cache_usage.sh [-m|--markdown] [key-prefix]
+# Usage: .ci/gha_cache_usage.sh [-m|--markdown] [key-prefix]
 #   -m, --markdown  output as a markdown table
 #   No key-prefix:  all caches grouped by prefix.
 #   One arg:        caches with that key prefix (default buildkit-blob).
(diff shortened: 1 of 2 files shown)


tooomm:
I'll have a look at your input and PR's tomorrow!

Some interesting stuff :)

ebbit1q:
does llm generated code conflict with the gpl?

brunoalr:
> does llm generated code conflict with the gpl?

that's a tricky question. I don't know the answer. I **think** not, but I am not a lawyer, so that's risky.

I tried to google it, but mostly found discussions about a different concern that does not apply to this project: "can LLMs 'leak' GPL code?"

I will close this PR due to uncertainty + risk. I also hand-wrote a [Python version](https://github.com/brunoalr/Cockatrice/pull/72) and I'm considering creating a standalone GHA for it, since it is quite generic

tooomm:
I like the idea of this and think it can be helpful.
We have ~300 cache entries in GHA and no overview.
I uniformized our cache naming already a bit with this in mind and allow filtering with keys: #6729

<br>

>does llm generated code conflict with the gpl?

What makes you think so, @ebbit1q?
We should maybe not blindly accept "vibe-coded" PR's if we can not review it in general.

<br>

>I also hand-wrote a [Python version](https://github.com/brunoalr/Cockatrice/pull/72) and I'm considering creating a standalone GHA for it, since it is quite generic

👍 

Idea:
- You could link the Prefix values in the table to a search with `key`, e.g.:
  `key:qt` https://github.com/Cockatrice/Cockatrice/actions/caches?query=key%3Aqt
  `key:ccache` https://github.com/Cockatrice/Cockatrice/actions/caches?query=key%3Accache

- Add a %-value of the total cache storage used, and of each "Prefix" contribution maybe?

PR #6705, opened 2026-03-14
Contextual menu improvement

DawnFire42:
## Related Ticket(s)
 - I didn't find any, but there might be.

The "Move to" right-click menu had no "Table" option, forcing players to drag-and-drop cards to the table. 
Additionally, the table-row-to-grid-Y coordinate mapping (clampValidTableRow(2 - tableRow)) was duplicated across 5 callsites with an existing TODO requesting extraction, and 4 of those callsites had a latent bug where instants/sorceries (tableRow 3) mapped to the land row instead of the noncreatures row.

  - Extracts a TableZone::tableRowToGridY() helper that consolidates 5 duplicated callsites and fixes the latent tableRow > 2 mapping bug.
  - Adds a "Table" option to the "Move to" context menu that places each card in its correct row based on card type (unlike "Play", instants/sorceries always go to the table)
  - Adds a "Play" action to the stack zone's right-click context menu

### DIFF
--- a/cockatrice/src/client/settings/shortcuts_settings.h
+++ b/cockatrice/src/client/settings/shortcuts_settings.h
@@ -562,7 +562,7 @@ private:
                                                  ShortcutGroup::Move_selected)},
         {"Player/aPlayFacedown", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Battlefield, Face Down"),
                                              parseSequenceString(""),
-                                             ShortcutGroup::Move_selected)},
+                                             ShortcutGroup::Playing_Area)},
         {"Player/aMoveToTable", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Battlefield"),
                                             parseSequenceString(""),
                                             ShortcutGroup::Move_selected)},
--- a/cockatrice/src/game/player/menu/card_menu.cpp
+++ b/cockatrice/src/game/player/menu/card_menu.cpp
@@ -146,11 +146,10 @@ void CardMenu::createTableMenu(bool canModifyCard)
 {
     // Card is on the battlefield
     if (!canModifyCard) {
+        addAction(aDrawArrow);
+        addSeparator();
         addRelatedCardView();
         addRelatedCardActions();
-
-        addSeparator();
-        addAction(aDrawArrow);
         addSeparator();
         addAction(aClone);
         addSeparator();
@@ -165,10 +164,9 @@ void CardMenu::createTableMenu(bool canModifyCard)
     if (card->getFaceDown()) {
         addAction(aPeek);
     }
-
-    addRelatedCardView();
-    addRelatedCardActions();
-
+    addSeparator();
+    addAction(aClone);
+    addMenu(new MoveMenu(player));
     addSeparator();
     addAction(aAttach);
     if (card->getAttachedTo()) {
@@ -178,9 +176,6 @@ void CardMenu::createTableMenu(bool canModifyCard)
     addSeparator();
     addMenu(new PtMenu(player));
     addAction(aSetAnnotation);
-    addSeparator();
-    addAction(aClone);
-    addMenu(new MoveMenu(player));
     addSeparator();
     addAction(aSelectAll);
     addAction(aSelectRow);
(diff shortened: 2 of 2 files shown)


cockatrice/src/game/player/menu/move_menu.cpp
```diff
@@ -50,6 +55,7 @@ void MoveMenu::setShortcutsActive()
     aMoveToHand->setShortcuts(shortcuts.getShortcut("Player/aMoveToHand"));
     aMoveToGraveyard->setShortcuts(shortcuts.getShortcut("Player/aMoveToGraveyard"));
     aMoveToExile->setShortcuts(shortcuts.getShortcut("Player/aMoveToExile"));
+    aMoveToTable->setShortcuts(shortcuts.getShortcut("Player/aMoveToTable"));
```
RickyRister, line 58: Remember to also update the [shortcut settings](https://github.com/DawnFire42/Cockatrice/blob/contextual-menu-improvement/cockatrice/src/client/settings/shortcuts_settings.h)

DawnFire42: I found another duplicate in shortcut settings related to that action, I'll deal with it at the same time.

cockatrice/src/game/player/menu/card_menu.cpp
```diff
@@ -203,6 +203,9 @@ void CardMenu::createStackMenu(bool canModifyCard)
 {
     // Card is on the stack
     if (canModifyCard) {
+        addAction(aPlay);
```
RickyRister: Never realized that `play` wasn't in the stack menu because I always just double clicked. Good catch!

cockatrice/src/client/settings/shortcuts_settings.h
```diff
@@ -563,9 +563,9 @@ class ShortcutsSettings : public QObject
         {"Player/aPlayFacedown", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Battlefield, Face Down"),
                                              parseSequenceString(""),
                                              ShortcutGroup::Move_selected)},
-        {"Player/aPlay", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Battlefield"),
```
RickyRister, line 566: This shortcut is still used for the `play` action, so we can't just get rid of it.

`play` doesn't actually appear under the "move to" submenu though (and neither does `play face down`), so it could make sense to move these shortcuts under a different category

cockatrice/src/game/player/menu/card_menu.cpp
RickyRister: Actually, can you also add the `play face down` action? Since all the other zones also have it.

cockatrice/src/client/settings/shortcuts_settings.h
RickyRister: You can move them under `Playing Area` since that's where we put all the other card menu shortcuts by default.

cockatrice/src/game/player/menu/card_menu.cpp
DawnFire42: I assumed cards in the stack weren't ever face down in the first place, maybe I'm wrong. I'll check and add it. Also, if all the options are there, it make sense to standardise the position of those options. I'll follow the positioning logic of the hand menu, its just gonna be easier for everyone to always see the options in the same order.

cockatrice/src/client/settings/shortcuts_settings.h
DawnFire42: I'll see what I can do.

cockatrice/src/game/player/menu/card_menu.cpp
RickyRister: The `play face down` action causes the card to go from the stack to the table face-down. 
Other public zones (such as graveyard and exile) still have that action despite the zone being public, so I'm pretty sure we're just intended to add that option to all zones.

cockatrice/src/client/settings/shortcuts_settings.h
RickyRister: Actually, upon looking further into it, it seems like we accidentally put two `"Player/aPlay"` into the shortcuts map, causing the later one to overwrite the first one. Seems like we never noticed this bug until now.

I'll make a separate PR to fix it (since I like bug fixes to be in separate commits). You can rebase onto it once it's merged.

RickyRister:
This PR is now mixing the implementation of a new action, along with the reorganizing of the card menus. I'd prefer if those two things were in separate PRs. If it's not too inconvenient, would it be possible for you to split those into two separate PRs? (One PR for implementing the `moveToTable` action, and one for adding the missing actions to and reorganizing the menus)

DawnFire42:
Fine by me, I'll work on it again in a few days

DawnFire42:
I'm dropping this PR to bring the changes in two separate ones rebased on the most recent master.

PR #6706, opened 2026-03-14
ci: use ninja + sccache on Windows (original branch)

brunoalr:
## Related Ticket(s)
- Fixes #IssueNumber

Trying this again after almost a year, with some help from LLMs this time. 

- Windows CI builds will use ninja and sccache
- Build time reduced from ~30mins to ~10mins on a warm cache

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,7 +261,9 @@ endif()
 set(CPACK_PACKAGE_CONTACT "Zach Halpern <zach@cockatrice.us>")
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "${PROJECT_NAME}")
 set(CPACK_PACKAGE_VENDOR "Cockatrice Development Team")
-set(CPACK_PACKAGE_DESCRIPTION "Cockatrice is an open-source, multiplatform application for playing tabletop card games over a network. The program's server design prevents users from manipulating the game for unfair advantage. The client also provides a single-player mode, which allows users to brew while offline.")
+set(CPACK_PACKAGE_DESCRIPTION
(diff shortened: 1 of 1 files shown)


tooomm:
Can we look into both separately?

Using ninja and change the compiler on win
Adding compiler cache

brunoalr:
> Can we look into both separately?
> 
> Using ninja and change the compiler on win Adding compiler cache

sure. what's the rationale? to have smaller PRs or to check the performance? if it's about performance, only using ninja didn't help much. But I didn't try to use sccache without ninja (is that possible? I was assuming it wasn't)

brunoalr:
> Using ninja and change the compiler on win

btw, there's no compiler change, the build still uses MSVC. it's a bit misleading that the cmake generator has `vs` in the name xD

tooomm:
> > Can we look into both separately?
> >
> > Using ninja and change the compiler on win
> >Adding compiler cache
> 
> sure. what's the rationale? to have smaller PRs or to check the performance? if it's about performance, only using ninja didn't help much. But I didn't try to use sccache without ninja (is that possible? I was assuming it wasn't)

Both are unrelated topics, we should look into them separately.
Also smaller PR's help with reviewing things individually.

If ninja is a requirement for sccache (which I'm not aware of), we will look into ninja first. 👍 

Closing this one, let's continue in the individual PR's! (Thanks for splitting them already! 🎉)
#6707 
#6708

brunoalr:
Ninja (new branch): https://github.com/Cockatrice/Cockatrice/pull/6721

brunoalr:
> If ninja is a requirement for sccache (which I'm not aware of), we will look into ninja first. 👍

I tried (quite hard tbh) to use `sccache` + `Visual Studio` cmake generator + `MSVC` compiler and, for the life of me, I couldn't make it work :( on the other, hand I couldn't find official docs saying that was not supported

so I decided to be pragmatic and use what I know that works in practice, which is `ninja`

PR #6708, opened 2026-03-15
Ci/use sccache on win

brunoalr:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -349,8 +349,8 @@ jobs:
       - name: Setup ${{ matrix.ccache_variant }}
         uses: hendrikmuhs/ccache-action@v1.2.20
         with:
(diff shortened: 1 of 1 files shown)


tooomm:
Same as for your separate ninja PR:
Can clean this up so we're looking on the relevant changes only, please? 😇 

I see a mix of changes, e.g. around CMAKE "target_file_dir" and PDBs path changes, maybe they should go into their own PR/PRs and can be merged even quicker! They seem not related or needed for the rest?

brunoalr:
unfortunately these changes (including PDBs) are needed due to differences with paths where binaries are generated between ninja and vs, my intention is to document them properly before moving the pr from draft to open. but yeah, this should be part of the ninja PR, not the sccache one.

btw, I should have opened these test PRs first on my fork before opening them here to avoid noise, don't know why I changed my usual workflow this time

PR #6709, opened 2026-03-15
Fix/cmake test only build

DawnFire42:
I ran into some CMake issues and it make sense to fix them in a single PR given that they are somewhat related. 

## What changed:
- Guard "filter_string_test" behind "WITH_ORACLE"/"WITH_CLIENT" to prevent configure-time failures when its dependency (`libcockatrice_filters`) isn't built.
- Centralize "TEST_QT_MODULES" in "FindQtRuntime.cmake" instead of duplicating across individual test CMakeLists files.
- Guard `libcockatrice_network` behind application targets, matching the existing pattern for other application-specific libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -254,7 +254,9 @@ endif()
 set(CPACK_PACKAGE_CONTACT "Zach Halpern <zach@cockatrice.us>")
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "${PROJECT_NAME}")
 set(CPACK_PACKAGE_VENDOR "Cockatrice Development Team")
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
very useful, thank you for this

we might want to consider running a ci target for this

tooomm:
Thanks @DawnFire42!
Our CMake stuff is mostly very old and probably outdated.
If you see more improvements or want to modernize it - maybe worth having a look.

<br>

> we might want to consider running a ci target for this

@ebbit1q Can you elaborate a bit more?

ebbit1q:
we have added a build target that builds just the server before, similarly, we could have one of our builds try to only build the tests without other components to ensure it is possible to only build the test without other targets

PR #6721, opened 2026-03-18
ci: use ninja in Windows builds

brunoalr:
## Related Ticket(s)
N/A

Windows builds take around **30min**. Currently they don't use compiler caches.

From experience with https://github.com/OpenImageDebugger/OpenImageDebugger, I know that it's possible to use `sccache` on Windows if the `cmake` generator is set to `Ninja`.

- by using `ninja` alone there's no real build performance gain, but I could only manage to use `sccache` (which DOES improve build times by a lot) with it.
- [Tests](https://github.com/brunoalr/Cockatrice/actions/runs/23222005309/job/67599234063) with `sccache` on my fork indicate that build times can go down to around **10min** (original build time reduced by *66%*) on warm cache hits.

- Windows builds on CI will use `Ninja` as the `cmake` generator
- Keep using the `MSVC` compiler
- Keep support for `Visual Studio` `cmake` generator - local builds can still use it
- Groundwork to use `sccache` on CI ([branch](https://github.com/brunoalr/Cockatrice/pull/74) is ready on my fork)
- Opening incremental PRs based on this request: https://github.com/Cockatrice/Cockatrice/pull/6706#issuecomment-4063781584

<details>

- `ninja` does not support `CMAKE_GENERATOR_PLATFORM`, so we have to use `CMAKE_SIZEOF_VOID_P == 8` to check if we are targeting x64
- `ninja` is a single-config generator, so build paths won't have intermediate `Release/`/`Debug/` paths. This is a departure from the Visual Studio generator. In order to support both generators, we can use `cmake` generator expressions such as `$<TARGET_FILE_DIR:cockatrice>` instead of `"${CMAKE_BINARY_DIR}/${PROJECT_NAME}/${CMAKE_BUILD_TYPE}/"` for the `install()` command.
- a bunch of environment variables need to be set for `MSVC` to work with `ninja` via CLI. ~Easiest solution that I know is to use https://github.com/ilammy/msvc-dev-cmd. Without it, `cmake` uses `gcc` as the compiler on CI~. Using a [CMake toolchain](https://github.com/MarkSchofield/WindowsToolchain?rgh-link-date=2026-03-26T20%3A08%3A14Z) to do it
</details>


ebbit1q:
will we move developer builds to ninja as well? we should not encourage developers to build using msvc if we do not test if it works I think.

brunoalr:
> will we move developer builds to ninja as well? we should not encourage developers to build using msvc if we do not test if it works I think.

fair point. to be able to do so, I exchanged the [ilammy/msvc-dev-cmd](https://github.com/ilammy/msvc-dev-cmd) action from previous commits in this PR with a [CMake toolchain](https://github.com/MarkSchofield/WindowsToolchain) that makes it possible to use ninja+MSVC locally. I tested locally with CLion (recommended IDE from the [wiki](https://github.com/Cockatrice/Cockatrice/wiki/Compiling-Cockatrice-(Windows))): code navigation, compilation and debugging work fine.

I also tested building from the command line:
```
cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DUSE_VCPKG=1 -DTEST=1 -G Ninja
cmake --build build
```


just to be clear, in this PR, MSVC is still the compiler toolchain, what changes is just the CMake generator

What I need help with:
- checking if uploaded PDBs work

PR #6727, opened 2026-03-21
Dependabot: Enable git submodules tracking

tooomm:
Tracking of latest version is now available and used instead latest commit as default (SemVer tagging for releases is needed).

- Enable dependabot tracking for `git submodules`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,11 +2,10 @@
 
 version: 2
 updates:
- # # Enable version updates for git submodules
- # Not yet possible to bump only on tags or releases, see:
+ # Enable version updates for git submodules
+ # If SemVer is used, updates will happen to new releases only (not HEAD)
  # https://github.com/dependabot/dependabot-core/issues/1639
  # https://github.com/dependabot/dependabot-core/issues/2192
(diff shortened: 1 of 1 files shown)


ebbit1q:
so, this would trigger a monthly update to vcpkg? do we use any other submodules?

tooomm:
> so, this would trigger a monthly update to vcpkg?

Not necessarily. It would check on a monthly basis for new tagged releases only.

<br>

> do we use any other submodules?

Yes, we use another submodule for our Doxygen theme.

<br>

With vcpkg there are two things to consider and check still:
- They use `YYYY.0M.0D` tagging scheme, not exactly SemVer. We need to see how that works out.
- Newer than our currently used version leads to the issue mentioned in https://github.com/Cockatrice/Cockatrice/pull/6627#issuecomment-3941865416, which has no solution in place yet.

I would like to try this out after the recent changes (check the updated comments) and see how it works for us. We do not need to merge the resulting dependabot PR's immediately.

brunoalr:
> They use YYYY.0M.0D tagging scheme, not exactly SemVer. We need to see how that works out.

current config pulled latest commit instead of latest tag: https://github.com/Cockatrice/Cockatrice/pull/6750#issuecomment-4144135464

PR #6728, opened 2026-03-21
Docker compose: Link to our GHCR hosted `servatrice` image

tooomm:
## Related Ticket(s)
- Related #5807

Users had to build a Docker Image first when using the compose file as it was pointing to our Dockerfile only.
Now that we provide a pre-build image and host it on GHCR, we can point to that instead and users can download the ready-to-use image directly.

- Point to our own `servatrice` image which is automatically build for each full release:
  https://github.com/Cockatrice/Cockatrice/pkgs/container/servatrice
- Always use the most recent version available by default (`servatrice:latest`)

<br>

This is based on the test and report from @maybeanerd in #5807.

### DIFF
--- a/docker-compose.yml.windows
+++ b/docker-compose.yml.windows
@@ -13,6 +13,9 @@ services:
       - ./servatrice/mysql-storage:/var/lib/mysql
 
   servatrice:
+    build:
+      context: .
+      dockerfile: Dockerfile
     image: ghcr.io/cockatrice/servatrice:latest
     depends_on:
       - mysql


docker-compose.yml
```diff
@@ -13,10 +13,7 @@ services:
       - $PWD/servatrice/servatrice.sql:/docker-entrypoint-initdb.d/servatrice.sql
 
   servatrice:
-    build:
```
maybeanerd, line 16: We should be able to keep the build params, which can be used for local development 
It will only be used when running `docker compose up`  with the `--build` flag

maybeanerd: I removed them for local testing just to be sure I was really using the published image

tooomm: Ok, sure! Did I update correctly?

PR #6729, opened 2026-03-21
CI: Uniform caching strategy & improved GHA cache names

tooomm:
- ~~As `matrix.runner` already included "intel", we end up with double `intel-Intel`:~~
  image
  Solved in macos-26 runner PR (#7035)
- ~~We currently show the macOS version of the underlaying runner, not the target OS.
  In above example we are building for "macOS **13** Intel", not "macOS **15** Intel".~~
  Solved in macos-26 runner PR (#7035)
- Finding the right entries amongst the 500+ entries is tricky. The filter is also very limited in how it works and what can be searched for:
  image
- Names and cache separation are not very good designed or clear
- Compiler caches do save when run from PR's which pollutes core caches for example

- Same naming as with our binaries:
  - No hint: default Apple Silicon for modern systems
  - `Intel` label: Native Intel version for compatibility on older x86_64 Mac hardware
- Apply same caching strategy as for compiler cache, also for Docker, Qt and vcpkg:
  Only save caches for runs from `master` branch
- Have clearer scope and uniform pattern when going through GHA caches (`Qt-*`, `ccache-*`, `vcpkg-*` ...)
(inspired by #6704)
  - Rename Qt cache for Mac
  - Rename Qt cache for Win
  - Rename vcpkg cache
  - The docker related ones (`buildkit`, `docker.io`, `index`) are tricky to rename and give a uniform prefix
- Cache filtering for `qt` now works: ["key:qt"](https://github.com/Cockatrice/Cockatrice/actions/caches?query=key%3Aqt)
- Cleaner names for steps without spaces when no "matrix.soc" is available for Win builds

<br>

Superseded by #7186, which introduces the uniform "save caches from master" strategy.
The other changes are mostly in place already, or no longer relevant.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -357,8 +357,8 @@ jobs:
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
         with:
           path: ${{env.CCACHE_DIR}}
-          key: ccache-${{matrix.runner}}-${{matrix.soc}}-${{matrix.type}}-${{env.BRANCH_NAME}}
-          restore-keys: ccache-${{matrix.runner}}-${{matrix.soc}}-${{matrix.type}}-
+          key: ccache-${{matrix.os}}${{matrix.target}}${{matrix.soc == 'Intel' && '_Intel' || ''}}-${{matrix.type}}-${{env.BRANCH_NAME}}
+          restore-keys: ccache-${{matrix.os}}${{matrix.target}}${{matrix.soc == 'Intel' && '_Intel' || ''}}-${{matrix.type}}-
 
       - name: Install aqtinstall
         run: pipx install aqtinstall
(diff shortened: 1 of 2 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -357,8 +357,8 @@ jobs:
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
         with:
           path: ${{env.CCACHE_DIR}}
-          key: ccache-${{matrix.runner}}-${{matrix.soc}}-${{matrix.type}}-${{env.BRANCH_NAME}}
-          restore-keys: ccache-${{matrix.runner}}-${{matrix.soc}}-${{matrix.type}}-
+          key: ccache-${{matrix.os}}${{matrix.target}}${{matrix.soc == 'Intel' && '_Intel' || ''}}-${{matrix.type}}-${{env.BRANCH_NAME}}
```
ebbit1q, line 360: can you extract this into a variable or matrix key?

ebbit1q: I'd like to reduce the amount of ternary expressions in our builds where possible, they are hard to read and I see them often being duplicated, it's harder to understand a duplicated one of these is identical than a variable

PR #6732, opened 2026-03-22
Add Docker to release template

tooomm:
- Add link to Docker

See https://github.com/Cockatrice/Cockatrice/blob/tooomm-docker_template/.ci/release_template.md


ebbit1q (approved):
thanks!

PR #6737, opened 2026-03-24
Add visual indicator to toggle untap button

### LINKED ISSUE #3888: Add visual indicator to "Toggle untap" option
**Similar Requests**
None

Same as "Always reveal top card" of library, have a check mark next to the option to indicate if the option is active or not.

**Additional Context**
There is the purple border already, but there should be a consistentcy in our menus. It's also just very nice and helpful.
Simply adding `aDoesntUntap->setCheckable(true)` to player.cpp doesn't do the trick as it needs to be per card.

See screenshot in https://github.com/Cockatrice/Cockatrice/pull/3828 to understand how it looks.

scotland0208:
## Related Ticket(s)
- Fixes #3888

There was no visual indicator in the context menu for whether a card would untap normally.

Now, there is a checkbox, like the "Always reveal top card" feature in the context menu

## Screenshots

Screenshot From 2026-03-24 00-23-55

### DIFF
--- a/cockatrice/src/client/settings/shortcuts_settings.h
+++ b/cockatrice/src/client/settings/shortcuts_settings.h
@@ -501,7 +501,7 @@ private:
         {"Player/aUntapAll", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Untap All"),
                                          parseSequenceString("Ctrl+U"),
                                          ShortcutGroup::Playing_Area)},
-        {"Player/aDoesntUntap", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Toggle Untap"),
+        {"Player/aDoesntUntap", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Toggle Skip Untapping"),
                                             parseSequenceString("Alt+U"),
                                             ShortcutGroup::Playing_Area)},
         {"Player/aFlip", ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Turn Card Over"),


tooomm:
What do you think about a rename, too?

Like "Skip untapping"?

scotland0208:
Good point. Especially since the default shortcut doesn't match the accelerator key. Previously it was underlining "n" even though the shortcut was Alt + U.

Fixed.

ebbit1q:
the accellerator keys have beem messed up ever since the default shortcuts started using alt disregarding the already present accellerator keys and conflicting with them.

ebbit1q (approved):
thanks

RickyRister:
Do we want to update the text in the [shortcut settings](https://github.com/scotland0208/Cockatrice/blob/d74717dd6a1f7798c9a39c7631d3e7ae4f5e269e/cockatrice/src/client/settings/shortcuts_settings.h#L504)? 

Screenshot 2026-03-24 at 11 55 28 PM

It currently has the text `Toggle Untap`. Maybe we want to update it to `Toggle Skip Untapping`?

scotland0208:
You got it. Updated that string as well.

The only other place I was able to find that was using similar text was in the [message log](https://github.com/scotland0208/Cockatrice/blob/d74717dd6a1f7798c9a39c7631d3e7ae4f5e269e/cockatrice/src/game/log/message_log_widget.cpp#L682). Where it says "%1 sets %2 to not untap normally", but this still feels like an appropriate sentence without needing to change it say "%1 sets %2 to skip untapping". Especially when you consider the corresponding log message for when you toggle it off. Right now it reads, "%1 sets %2 to untap normally." But changing it "%1 sets %2 to not skip untapping." feels awkward. 

If you disagree, I can make an update here as well.

RickyRister:
I'm fine with the other text staying the same.

PR #6738, opened 2026-03-24
Add moveToTable context menu action and extract tableRowToGridY helper

DawnFire42:
An option to move cards to the table was missing from all contextual menus. This is distinct from the play action in its behavior, sending the card to the table directly.

Adds a Table option to the Move menu, allowing cards to be moved directly to the battlefield from any zone. Extracts the repeated tableRow-to-grid-Y conversion logic into TableZone::tableRowToGridY(), consolidating five call sites and fixing a latent bug where cards with tableRow > 2 could land on the wrong row.


cockatrice/src/game/player/menu/move_menu.cpp
```diff
@@ -17,6 +17,8 @@ MoveMenu::MoveMenu(Player *player) : QMenu(tr("Move to"))
     aMoveToGraveyard->setData(cmMoveToGraveyard);
     aMoveToExile = new QAction(this);
     aMoveToExile->setData(cmMoveToExile);
+    aMoveToTable = new QAction(this);
+    aMoveToTable->setData(cmMoveToTable);
```
RickyRister: nit: the order of the lines for each action in the code is meant to match the order that they appear in the menu. For consistency, I think we should do that for the new action.

PR #6739, opened 2026-03-24
Contextual menus improvements

DawnFire42:
Play and play face down options weren't implemented for the stack

Reorganize card context menus across table, stack, and graveyard/exile zones for better consistency: promote Draw Arrow and Clone actions, move related card entries to the bottom, add Play/Play Face Down to the stack menu, and flatten if/else blocks with early returns. 
Also fix playCard() ignoring the faceDown flag when routing instants/sorceries from the stack, which sent them to the graveyard instead of the table.

### DIFF
--- a/cockatrice/src/game/player/menu/card_menu.cpp
+++ b/cockatrice/src/game/player/menu/card_menu.cpp
@@ -108,6 +108,7 @@ CardMenu::CardMenu(Player *_player, const CardItem *_card, bool _shortcutsActive
 
     if (revealedCard) {
         addAction(aHide);
+        addSeparator();
         addAction(aClone);
         addSeparator();
         addAction(aSelectAll);
@@ -146,16 +147,14 @@ void CardMenu::createTableMenu(bool canModifyCard)
 {
     // Card is on the battlefield
     if (!canModifyCard) {
-        addRelatedCardView();
-        addRelatedCardActions();
-
-        addSeparator();
         addAction(aDrawArrow);
         addSeparator();
         addAction(aClone);
         addSeparator();
         addAction(aSelectAll);
         addAction(aSelectRow);
+        addRelatedCardView();
+        addRelatedCardActions();
         return;
     }
 
@@ -165,10 +164,9 @@ void CardMenu::createTableMenu(bool canModifyCard)
     if (card->getFaceDown()) {
         addAction(aPeek);
     }
-
-    addRelatedCardView();
-    addRelatedCardActions();
-
+    addSeparator();
+    addAction(aClone);
+    addMenu(new MoveMenu(player));
     addSeparator();
     addAction(aAttach);
     if (card->getAttachedTo()) {
@@ -178,9 +176,6 @@ void CardMenu::createTableMenu(bool canModifyCard)
     addSeparator();
     addMenu(new PtMenu(player));
     addAction(aSetAnnotation);
-    addSeparator();
-    addAction(aClone);
-    addMenu(new MoveMenu(player));
     addSeparator();
     addAction(aSelectAll);
     addAction(aSelectRow);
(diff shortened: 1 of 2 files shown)


cockatrice/src/game/player/player_actions.cpp
```diff
@@ -64,7 +64,7 @@ void PlayerActions::playCard(CardItem *card, bool faceDown)
     int tableRow = info.getUiAttributes().tableRow;
     bool playToStack = SettingsCache::instance().getPlayToStack();
     QString currentZone = card->getZone()->getName();
-    if (currentZone == ZoneNames::STACK && tableRow == 3) {
+    if (!faceDown && currentZone == ZoneNames::STACK && tableRow == 3) {
```
RickyRister, line 67: good catch!

RickyRister:
New actions work as advertised

Screenshot 2026-03-24 at 11 28 21 PM

cockatrice/src/game/player/menu/card_menu.cpp
```diff
@@ -197,57 +192,66 @@ void CardMenu::createTableMenu(bool canModifyCard)
...
-        addSeparator();
-        addAction(aClone);
-        addMenu(new MoveMenu(player));
+    if (!canModifyCard) {
+        addAction(aDrawArrow);
         addSeparator();
-        addAction(aSelectAll);
-        addAction(aSelectColumn);
-
+        addRelatedCardView();
```
RickyRister: Image

Image

Unmodifiable zones still have the alternate order where the related cards are near the top. Is that intentional?

cockatrice/src/game/player/menu/card_menu.cpp
```diff
@@ -165,10 +164,9 @@ void CardMenu::createTableMenu(bool canModifyCard)
     if (card->getFaceDown()) {
         addAction(aPeek);
     }
-
-    addRelatedCardView();
-    addRelatedCardActions();
```
RickyRister, line 172: Not used to the tokens being at the bottom of the menu now.
Image

My muscle memory still thinks the tokens are near the top
Image

I know the other zones already have the related cards at the bottom of the menu, but I create tokens from other zones much less often, not to mention those menus are much smaller.

I still agree that the new order makes more sense. It's just that it's taking me some time to get used to it.

DawnFire42: Yes they were definitely more adapted to each specific zones before. I'm not opposed to some specific exceptions, it just comes at the cost of uniformity across the menus. I'm also not opposed to have them in a different order. I took the order from the hand at first since it already had most options. 

Since I didn't play any other card games on the platform, I wouldn't know where exceptions would make objective sense. Though I'm guessing this can be easily changed later.

cockatrice/src/game/player/menu/card_menu.cpp
DawnFire42: It isn't, I actually forgot those. Let me add them to this PR

RickyRister:
LGTM

Screenshot 2026-03-25 at 2 56 00 PM

RickyRister:
Huh, seems like view zones from card reveals still lack the token actions. However, that's already the existing behavior, and we can fix that in a follow-up PR if we really want to fix it.

Screenshot 2026-03-25 at 3 01 26 PM

DawnFire42:
> Huh, seems like view zones from card reveals still lack the token actions. However, that's already the existing behavior, and we can fix that in a follow-up PR if we really want to fix it.
> 
Yess, I thought it was intentionnal so I left it as is

ebbit1q:
it's not intentional, it's more of an oversight, creating tokens of cards while they are anywhere other than the battlefield used to not be a thing for years

PR #6741, opened 2026-03-25
Scry fix

### LINKED ISSUE #3921: Scrying implies cards are in opposite order
<b>System Information:</b>
  
Client Version: 2.7.2-custom(unknown) (unknown)

  
When scrying multiple cards, the interface shows the cards overlapping each other. 
Which card corresponds to the topmost card of the deck? The obvious guess for a new player is that the card which appears topmost visually, is the card that is topmost in the deck.
cockatrice_ui
In this screenshot, one would think the Mountain is the top card of the deck. The card that's most prominently visible is probably the one you would draw next, right? But actually, the card Scorch Spitter is the next card you would draw. 

I suggest changing it so that the card which is on top of the deck visually appears on top:
newcockatrice_ui

scotland0208:
## Related Ticket(s)
- Fixes #3921 

In the view top of library window, for some people, it can be ambiguous which card is on the top of the library. 

Now, it will be more apparent. Above the card that represents the top of the library, it will say "Top of Library" 

Similarly, when viewing the bottom of the library, it will say "Bottom of Library" beneath the bottom card. 

It will only show this when viewing the library and only when viewing a specific number of cards. That is, if you just use F3, it will not show the label.

It will also not show when viewing any other zone (graveyard, exile, etc).

## Screenshots
Screenshot From 2026-03-26 20-45-32

### DIFF
--- a/cockatrice/src/game/zones/view_zone_widget.cpp
+++ b/cockatrice/src/game/zones/view_zone_widget.cpp
@@ -145,7 +145,7 @@ ZoneViewWidget::ZoneViewWidget(Player *_player,
     zoneHBox->addItem(scrollBarProxy);
 
     vbox->addItem(zoneHBox);
-    
+
     if (_isReversed && showLibraryLabel(_origZone, numberCards)) {
         cardLocationLabel.setText(tr("Bottom of Library"));
         vbox->addItem(cardLocationLabelProxy);
--- a/cockatrice/src/game/zones/view_zone_widget.h
+++ b/cockatrice/src/game/zones/view_zone_widget.h
@@ -12,8 +12,8 @@
 #include <QComboBox>
 #include <QGraphicsProxyWidget>
 #include <QGraphicsWidget>
-#include <QLineEdit>
 #include <QLabel>
+#include <QLineEdit>
 #include <QPointer>
 #include <libcockatrice/utility/macros.h>
 


scotland0208:
I'm still fairly new to this codebase, so feedback would definitely be welcome and appreciated. 

I'm not sure how you all prefer to structure code.

ebbit1q:
it passes the linter, structure is fine

the view is consistent across the application, it always shows the first card at the top(left) and the last card at the bottom(right), with lower cards overlapping cards above it, new players generally learn this quickly but I do understand the concern

however, the implementation doesn't work as expected in cases where the view has multiple columns, or when it has been resized:
image
in these cases the label "bottom of library" is unclear.

instead some element that appears in the list would have to be added, or maybe just a tooltip.

in any case one of the main issues is simply that we don't provide enough explaination to new players, this could also be done with better guides etc.

scotland0208:
I did notice that this was an issue when I was implementing the change, but I felt that the benefit was still worth the minor confusion in this specific case. I think it's more often the case that someone will look at a handful of cards (less than 5) from the top or bottom of the library, and in those cases, the labels can be helpful when trying to rearrange which cards will be on top after you look at them. In my opinion, this is letting perfect be the enemy of good enough.

That said, my feelings won't be hurt if we'd rather scrap the change for now in favor of better documentation. 

Let me know! Thanks for taking a look at the changes.

ebbit1q:
I think this is something useful to have but I'd prefer for the label to be applying a little more clearly to the specific card that is actually at the top of the library, for this I think it should instead add a visual element inside the view instead of at the top or bottom outside of the view

an alternative idea would be to add indexes to all of these, the index of a card is always available to a client but never shown, adding an option in the settings that shows it inside of views would clear up which card is at what position in the deck

I wouldn't say this currently has the highest priority but certainly worth adding!

PR #6745, opened 2026-03-26
fix: solve deprecated literal operator error by updating peglib

brunoalr:
## Related Ticket(s)
- https://github.com/yhirose/cpp-peglib/issues/338

Compilation fails on latest Apple clang due to deprecation of whitespaces in [UDL](https://en.cppreference.com/w/cpp/language/user_literal.html)s.

```clang --version
Apple clang version 21.0.0 (clang-2100.0.123.102)
Target: arm64-apple-darwin25.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

```
In file included from /Users/bruno/ws/Cockatrice/libcockatrice_utility/libcockatrice/utility/expression.cpp:3:
/Users/bruno/ws/Cockatrice/libcockatrice_utility/libcockatrice/utility/peglib.h:492:42: error: identifier '_' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator]
  492 | inline constexpr unsigned int operator"" _(const char *s, size_t l) {
      |                               ~~~~~~~~~~~^
      |                               operator""_
1 error generated.
ninja: build stopped: subcommand failed.
```

- fix compilation error by updating peglib to https://github.com/yhirose/cpp-peglib/commit/dcabc63cf4b966eb6f33abd571ae4cda6bf707f0

### DIFF
--- a/libcockatrice_utility/libcockatrice/utility/peglib.h
+++ b/libcockatrice_utility/libcockatrice/utility/peglib.h
@@ -1,4 +1,4 @@
-//
+﻿//
 //  peglib.h
 //
 //  Copyright (c) 2022 Yuji Hirose. All rights reserved.
@@ -17,6 +17,7 @@
 
 #include <algorithm>
 #include <any>
+#include <bitset>
 #include <cassert>
 #include <cctype>
 #if __has_include(<charconv>)
@@ -30,6 +31,7 @@
 #include <map>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <set>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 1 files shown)


brunoalr:
actually I realized `peglib.h` is a vendored dependency. that gives more alternatives to deal with it:

1. move it to a separate directory (e.g. `vendor`) and mark it as a `SYSTEM` dependency like so in cmake: `target_include_directories(libcockatrice_utility SYSTEM PUBLIC {CMAKE_CURRENT_SOURCE_DIR}/libcockatrice/utility/vendor)` - this will make the compilers ignore warnings coming from it
2. update peglib.h: https://github.com/yhirose/cpp-peglib/issues/338
3. All of the above

brunoalr:
done

tooomm:
Why don't we include that as a submodule as well?
That way we have a clear reference where it's coming from, its version and if there are updates available.

brunoalr:
+1. I think it makes more sense, yeah. The only downside is that is that we would get all the files from the repo and not only the header we use, but IMO, the benefits outweigh this. 

Plus, I think marking it as `SYSTEM` dependency to silence warnings makes sense. we won't act on them anyways and we would have less clutter in the build logs.

brunoalr:
@tooomm https://github.com/brunoalr/Cockatrice/pull/86 WDYT?

PR #6748, opened 2026-03-27
[SettingsManager] Make setting getters const

RickyRister:
## Related Ticket(s)
- Follow-up to #6747 

Now that we made the settings getters no longer modify a stored `QSettings` object, that means we can make the setting getters const.

- Make `SettingsManager::getValue` const
- Make all subclasses' getters const

### DIFF
--- a/libcockatrice_settings/libcockatrice/settings/game_filters_settings.cpp
+++ b/libcockatrice_settings/libcockatrice/settings/game_filters_settings.cpp
@@ -8,11 +8,11 @@ GameFiltersSettings::GameFiltersSettings(const QString &settingPath, QObject *pa
 {
 }
 
-/*
+/**
  * The game type might contain special characters, so to use it in
  * QSettings we just hash it.
  */
-QString GameFiltersSettings::hashGameType(const QString &gameType)
+static QString hashGameType(const QString &gameType)
 {
     return QCryptographicHash::hash(gameType.toUtf8(), QCryptographicHash::Md5).toHex();
 }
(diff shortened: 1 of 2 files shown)


libcockatrice_settings/libcockatrice/settings/game_filters_settings.cpp
```diff
@@ -12,7 +12,7 @@ GameFiltersSettings::GameFiltersSettings(const QString &settingPath, QObject *pa
  * The game type might contain special characters, so to use it in
  * QSettings we just hash it.
  */
-QString GameFiltersSettings::hashGameType(const QString &gameType) const
+QString GameFiltersSettings::hashGameType(const QString &gameType)
```
ebbit1q: why are we removing const here?

libcockatrice_settings/libcockatrice/settings/game_filters_settings.h
```diff
@@ -57,7 +57,7 @@ class GameFiltersSettings : public SettingsManager
     explicit GameFiltersSettings(const QString &settingPath, QObject *parent = nullptr);
     GameFiltersSettings(const GameFiltersSettings & /*other*/);
 
-    [[nodiscard]] QString hashGameType(const QString &gameType) const;
+    [[nodiscard]] static QString hashGameType(const QString &gameType);
```
ebbit1q: oh, I see it's static now

RickyRister: oh right, that means i can remove it from the header

ebbit1q: +1

PR #6753, opened 2026-03-28
feat: support expressions when setting card counters

brunoalr:
## Related Ticket(s)
- Fixes #IssueNumber

Player counters support expressions when setting theirs values but card counters currently do not.

- support expressions when setting card counters, including multiple selection
- when multiple cards are selected, the dialog shows "x" as the old value, which allows for expressions such as `x * 2`, making it easy to double all given counter for the selected cards
- reuse `AbstractCounterDialog` based on https://github.com/Cockatrice/Cockatrice/blob/c5fde071e79c162fa20a74d133b97c9488a4cb95/cockatrice/src/game/board/abstract_counter.cpp#L180

## Screenshots

Single and multiple cards selected:

https://github.com/user-attachments/assets/abd214a2-ae57-49bb-b4d4-9a4364f5ad3b

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -1576,44 +1576,37 @@ void PlayerActions::actCardCounterTrigger()
         case 11: { // set counter with dialog
             player->setDialogSemaphore(true);
 
-            const auto &cardCounterSettings = SettingsCache::instance().cardCounters();
-            const auto &counterName = cardCounterSettings.displayName(counterId);
+            // If a single card is selected, we show the old value in the dialog. Otherwise, we show "x"
+            QList<QGraphicsItem *> sel = player->getGameScene()->selectedItems();
+            QString oldValueForDlg = "x";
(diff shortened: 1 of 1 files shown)


ebbit1q:
why are you making all of these local variables consts? what's the benefit?

brunoalr:
> why are you making all of these local variables consts? what's the benefit?

I try to follow [CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con-constants-and-immutability) as much as possible. The main benefits are readability and maintainability: when making variables const by default we can be sure it won't be unexpectedly changed later. The opposite is also helpful: when a variable is *not* const we can expect that it's being changed somewhere.

An added bonus is that const give optimization opportunities to compilers (mainly constant folding).

ebbit1q:
I don't really see any benefits to making locals const, it's just extra characters, any compiler can fold them as they're just local anyway

brunoalr:
I update the code. please take a look

PR #6755, opened 2026-03-28
chore: use peglib as submodule instead of vendoring it manually

brunoalr:
## Related Ticket(s)
- Fixes #IssueNumber

Suggestion from a previous PR: https://github.com/Cockatrice/Cockatrice/pull/6745#issuecomment-4147294562

- remove manually vendored `peglib.h`
- replace it with git submodule tracking https://github.com/yhirose/cpp-peglib/
- use https://github.com/yhirose/cpp-peglib/releases/tag/v1.10.2
- include `cpp-peglib` as `SYSTEM` directory
- decouple `peglib.h` from `libcockatrice_utility`
- create `thirdparty` directory meant to be used to add any other similar submodules (except `vcpkg`? - let's keep this out of scope of this PR)


tooomm:
> create `thirdparty` directory meant to be used to add any other similar submodules

SFMT would be another candidate for this later, any other?

brunoalr:
> > create `thirdparty` directory meant to be used to add any other similar submodules
> 
> SFMT would be another candidate for this later, any other?

I am not that familiar yet with the proper C++ side of the codebase, but we can look it up for other such cases. I suggest doing it in follow-up PRs though

tooomm:
I would like to hear other opinions on this as well

ebbit1q:
peglib is a single file and intended to simply be included in a project like this, however most would more clearly mark it as third party

this pr as is adds the entire repo including all docs as a submodule

brunoalr:
> peglib is a single file and intended to simply be included in a project like this, however most would more clearly mark it as third party
> 
> this pr as is adds the entire repo including all docs as a submodule


yes, I'm aware. IMO, the traceability and possibility to get automated dependabot updates outweigh the downside of pulling more files than just the header. I know there is sparse checkout in git, but that's quite cumbersome, so I think that is not worth it.

alternatively, we could keep the file but change its path like in this PR

Issue #6761, opened 2026-03-31
Linux Arch builds started to fail in CI


tooomm:
See e.g. https://github.com/Cockatrice/Cockatrice/actions/runs/23659485867/job/69303010627 on `master`.

ebbit1q:
this is most likely caused by arch shipping 6.11 instead of 6.10 now

PR #6766, opened 2026-04-03
Add Ubuntu 26.04 "Resolute Racoon" build

SlightlyCircuitous:

_Resolute_ will be released in full on April 23 and be supported for five years. Since we are aiming for a new release soon, I would love to get the latest Ubuntu LTS in the build list now so it's ready for release. As I understand it, [_Resolute_ is considered feature complete as of March 23](https://discourse.ubuntu.com/t/resolute-raccoon-release-schedule/47198) and is not expected to change between now and release aside from bug fixes and translations. The Docker tag `26.04` is up to date as of April 2. 

- Adds Ubuntu 26.04 to our Docker builds
- Updates the release template to reflect this

I will install and test this build in a VM.

### DIFF
--- a/.ci/Ubuntu26.04/Dockerfile
+++ b/.ci/Ubuntu26.04/Dockerfile
@@ -3,6 +3,7 @@ FROM ubuntu:26.04
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
         build-essential \
+        ca-certificates \
         ccache \
         clang-format \
         cmake \


SlightlyCircuitous:
Seems like the 26.04 build initially failed in the same way as https://github.com/Cockatrice/Cockatrice/pull/6068, but I applied the same fix and it builds now.

SlightlyCircuitous:
Arch is expected to be failing (see https://github.com/Cockatrice/Cockatrice/issues/6761), but I am not sure why macOS 14 is.

SlightlyCircuitous:
I have downloaded the build artifact and installed it in a VM. It installs without incident; checks for card updates; saves, loads, and edits decks; connects to servers; and handles local games as expected.

tooomm:
Fedora 44 is about to release in ~1 week as well, with Fedora 42 falling out of security support shortly after.

🐰

SlightlyCircuitous:
The docker image for Fedora 44 hasn't been updated for 2 months (which predates a number of freezes), otherwise I would have put in a PR for it. Hopefully it will be updated after release. I can remove 42 in the meantime if you like, though.

tooomm:
Just add and remove in the same PR when 44 is ready.
Appreciate that you keep looking at the updates here. 👍

SlightlyCircuitous:
Fedora 44 was delayed (at least) two weeks, to 04/28, and the docker image is still out of date, so I don't see us building for it for this release. However, I did just install the Fedora 43 package on my VM of 44 beta. Is this usually possible or did I just get lucky with a package match?

ebbit1q:
that's unexpected, I would not expect the versions of qt and protobuf to be identical, but it is possible

SlightlyCircuitous:
Looks like protobuf has been holding steady at 3.19.6 and Qt has only been having point releases on the 6.10.x branch since Fedora 42. Presumably, the 42 and 43 packages are also interchangeable, but I haven't tested that. But, if someone asks for a Fedora 44 package for 3.0 before it's stable enough for us to build for, they can probably just use the 43 package.

tooomm:
See my list for Qt and CMake versions across our builds in https://github.com/Cockatrice/Cockatrice/pull/6760#issuecomment-4188652458.

For protobuf, the versions are not always printed in logs.
Debian 12 uses 3.21
Fedora 42/43/44 use 3.19
Windows and macOS use 5.29

Seems like we have v3.21 defined as minimum with a fallback.
There is also some compatibility stuff for versions <3.1

PR #6768, opened 2026-04-04
fix compiling on arch

### LINKED ISSUE #6761: Linux Arch builds started to fail in CI

ebbit1q:
## Related Ticket(s)
- Fixes #6761

new version of protobuf adds no discard to parsing functions

- use the return values
- log when they fail

please tell me how you'd like me to log these properly

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_replays.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_replays.cpp
@@ -30,6 +30,8 @@
 #include <libcockatrice/protocol/pb/response_replay_get_code.pb.h>
 #include <libcockatrice/protocol/pending_command.h>
 
+inline Q_LOGGING_CATEGORY(TabReplaysLog, "replays_tab");
+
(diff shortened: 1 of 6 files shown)


cockatrice/src/interface/widgets/tabs/tab_replays.cpp
```diff
@@ -265,9 +265,11 @@ void TabReplays::actOpenLocalReplay()
         f.close();
 
         GameReplay *replay = new GameReplay;
-        replay->ParseFromArray(_data.data(), _data.size());
-
-        emit openReplay(replay);
+        if (replay->ParseFromArray(_data.data(), _data.size())) {
+            emit openReplay(replay);
+        } else {
+            qDebug() << "could not parse replay!";
```
RickyRister: unexpected failures should be logged at warning level.

Also, I'd prefer if we set up the logger properly and used qCDebug

Issue #1495, opened 2015-09-10
Register cockatrice:// protocol handler on install.


Daenyth:
I should be able to click on an html `<a href="cockatrice://cockatrice.woogerworks.com">` link and have the Cockatrice app open up, and either fill in the connection info as a new server, or find that server in my saved servers.

This would probably be facilitated with some CLI flags like `--server-host` and `--server-port` to trigger the pre-fill logic.

Likely we need platform-specific handling for each OS.

BruebachL:
Solved by #6775

Issue #1450, opened 2015-09-01
Support opening files via arguments


LPGhatguy:
Right now, running `cockatrice.exe deckfile.cod` won't open the deck automatically.

This would be a useful feature as it would allow `.cod` files to be associated with Cockatrice.

Daenyth:
:+1: although I might make it a flagged option. -f deck.cod

ZeldaZach:
In addition, can we possibly allow for the client to support the file types? (On Mac if you double click the file it will say nothing can open it. Why not allow cockatrice to open them?)

Daenyth:
@ZeldaZach can you open that as a new ticket? Good feature

BruebachL:
Solved by #6775

PR #6770, opened 2026-04-04
[Application] Add single instance guard and mime types.

BruebachL:
Took 2 hours 39 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/cockatrice/src/single_instance_manager.h
+++ b/cockatrice/src/single_instance_manager.h
@@ -1,0 +1,66 @@
+#ifndef COCKATRICE_SINGLE_INSTANCE_MANAGER_H
+#define COCKATRICE_SINGLE_INSTANCE_MANAGER_H
+
+#include <QDataStream>
+#include <QDebug>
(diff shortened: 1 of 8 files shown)


ebbit1q:
does this completely stop users from starting cockatrice twice, even to play a local game with two players?

cockatrice/src/single_instance_manager.h
```diff
@@ -0,0 +1,66 @@
...
+class SingleInstanceManager : public QObject
+{
+    Q_OBJECT
+public:
+    explicit SingleInstanceManager(QObject *parent = nullptr) : QObject(parent)
+    {
+    }
+
+    bool tryRun(const QStringList &initialFiles)
+    {
```
RickyRister, line 18: why is this in the header file and why is the cpp file empty?

BruebachL:
#1495

BruebachL:
#1450

PR #6774, opened 2026-04-05
Refactor vertical card stacking with clip containers for variable zone sizes

DawnFire42:
## Why?
The stacking behavior wasn't working when using a zone that could be resized to be smaller then the cards themselves. I had to make this for the command zone and it is best merged as a completely separate refactor/improvement in order to simplify reviewing process.

## What this does:
  Introduce a shared vertical stacking layout system in SelectZone that replaces the old divideCardSpaceInZone() free function with structured layout computation (StackLayoutParams, ZoneLayout, computeZoneLayout). Cards stack with smooth offset compression using sqrt-scaled overflow allowance, preventing jarring jumps when cards are added or removed.

  Add a clip container mechanism (QGraphicsRectItem) to SelectZone that confines card overflow to zone bounds while allowing hovered cards to temporarily escape clipping for full visibility. GameScene coordinates hover enter/exit to reparent cards between the clip container and zone.  This mechanic is opt-in and left unused for now, but future zones will benefit from having access to it already.

  - SelectZone: new layout engine with allowBottomOverflow opt-in; clip container infrastructure for future zones needing visual clipping
  - StackZone: uses new layout (no overflow); adds setHeight() for dynamic resizing
  - HandZone: vertical layout delegates to SelectZone's shared stacking
  - AbstractCardItem: preserves hover z-value during layout passes; invalidates scene rect on hover exit for proper sibling repainting
  - CardZone::onCardAdded made virtual for clip container reparenting
  - Zone widths updated to CardDimensions::WIDTH_F * 1.5

### DIFF
--- a/cockatrice/src/game/zones/select_zone.cpp
+++ b/cockatrice/src/game/zones/select_zone.cpp
@@ -8,14 +8,11 @@
 #include <QGraphicsSceneMouseEvent>
 #include <QtMath>
 
-namespace
-{
-qreal stackingOffset(qreal cardHeight)
+static qreal stackingOffset(qreal cardHeight)
 {
     const qreal overlapPercent = SettingsCache::instance().getStackCardOverlapPercent();
     return cardHeight * (100.0 - overlapPercent) / 100.0;
 }
-} // namespace
 
 SelectZone::ZoneLayout SelectZone::computeZoneLayout(const StackLayoutParams &params)
 {
@@ -56,13 +53,16 @@ SelectZone::ZoneLayout SelectZone::computeZoneLayout(const StackLayoutParams &pa
 SelectZone *SelectZone::findOwningSelectZone(const QGraphicsItem *card)
 {
     QGraphicsItem *parent = card ? card->parentItem() : nullptr;
-    if (!parent)
+    if (!parent) {
         return nullptr;
+    }
     // Card may be direct child of zone (escaped for hover) or child of clip container.
-    if (auto *zone = dynamic_cast<SelectZone *>(parent))
+    if (auto *zone = dynamic_cast<SelectZone *>(parent)) {
         return zone;
-    if (auto *zone = dynamic_cast<SelectZone *>(parent->parentItem()))
+    }
+    if (auto *zone = dynamic_cast<SelectZone *>(parent->parentItem())) {
         return zone;
+    }
     return nullptr;
 }
 
--- a/cockatrice/src/game/zones/select_zone.cpp
+++ b/cockatrice/src/game/zones/select_zone.cpp
@@ -4,36 +4,205 @@
 #include "../board/card_item.h"
 #include "../game_scene.h"
 
-#include <QDebug>
+#include <QGraphicsRectItem>
 #include <QGraphicsSceneMouseEvent>
+#include <QtMath>
 
-qreal divideCardSpaceInZone(qreal index, int cardCount, qreal totalHeight, qreal cardHeight, bool reverse)
+namespace
 {
-    qreal cardMinOverlap = cardHeight * SettingsCache::instance().getStackCardOverlapPercent() / 100;
-    qreal desiredHeight = cardHeight * cardCount - cardMinOverlap * (cardCount - 1);
-    qreal y;
-    if (desiredHeight > totalHeight) {
-        if (reverse) {
-            y = index / ((totalHeight - cardHeight) / (cardCount - 1));
+qreal stackingOffset(qreal cardHeight)
+{
+    const qreal overlapPercent = SettingsCache::instance().getStackCardOverlapPercent();
+    return cardHeight * (100.0 - overlapPercent) / 100.0;
+}
+} // namespace
+
+SelectZone::ZoneLayout SelectZone::computeZoneLayout(const StackLayoutParams &params)
+{
+    if (params.cardCount <= 0) {
+        return {0.0, 0.0};
+    }
+    qreal effectiveOffset = params.desiredOffset;
+    if (params.cardCount > 1) {
+        qreal fitOffset;
+        if (params.totalHeight < params.cardHeight && params.minOffset > 0.0) {
+            // Zone is shorter than a card (e.g. minimized). Compress offsets so
+            // every card has at least minOffset pixels of its top visible.
+            fitOffset = (params.totalHeight - params.minOffset) / (params.cardCount - 1);
+            effectiveOffset = qMax(0.0, qMin(params.desiredOffset, fitOffset));
         } else {
-            y = index * (totalHeight - cardHeight) / (cardCount - 1);
-        }
-    } else {
-        qreal start = (totalHeight - desiredHeight) / 2;
-        if (reverse) {
-            if (index <= start) {
-                return 0;
+            qreal reservedForBottomCard;
+            if (params.allowBottomOverflow) {
+                // Allow the bottom card to partially overflow in tight zones, scaling the
+                // overflow allowance by sqrt(cardCount-1) so offsets decrease smoothly
+                // as cards are added rather than dropping by 1/(n-1) each time.
+                // The 0.75 ratio was tuned experimentally to balance card visibility vs. overflow.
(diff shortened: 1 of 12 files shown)


cockatrice/src/game/zones/select_zone.h
```diff
@@ -10,26 +10,116 @@
...
+    struct ZoneLayout
+    {
+        qreal effectiveOffset; ///< Actual offset between card tops (may be compressed)
+        qreal start;           ///< Y coordinate of the first card's top edge
+    };
+
+    /// Minimum visible pixels of each card's top edge when stacking compresses offsets in tight zones.
+    static constexpr qreal MIN_CARD_VISIBLE = 10.0;
+
+    static qreal stackingOffset(qreal cardHeight);
```
RickyRister: nit: Are you planning on using this method anywhere outside of this class? If not, you could just make a static function in the cpp file.

cockatrice/src/game/zones/select_zone.cpp
```diff
@@ -4,38 +4,202 @@
...
-            y = (index - start) / (cardHeight - cardMinOverlap);
-        } else {
-            y = index * (cardHeight - cardMinOverlap) + start;
+    }
+    qreal stackHeight = (params.cardCount - 1) * effectiveOffset + params.cardHeight;
+    qreal start = (stackHeight <= params.totalHeight) ? (params.totalHeight - stackHeight) / 2.0 : 0.0;
+    return {effectiveOffset, start};
+}
+
+SelectZone *SelectZone::findOwningSelectZone(QGraphicsItem *card)
```
RickyRister:
```suggestion
SelectZone *SelectZone::findOwningSelectZone(const QGraphicsItem *card)
```

cockatrice/src/game/zones/select_zone.h
```diff
@@ -10,26 +10,116 @@
...
+     * When the stack fits with room to spare, it is centered vertically.
+     */
+    static ZoneLayout computeZoneLayout(const StackLayoutParams &params);
+
+    /// Builds StackLayoutParams from the current card list and zone geometry.
+    StackLayoutParams buildStackParams(qreal minOffset = 0.0) const;
+
+    /// Computes the card index at a given y-coordinate within the zone's vertical layout.
+    /// Returns 0 if the zone has no cards or the offset is zero.
+    int dropIndexFromY(qreal dropY, qreal minOffset = 0.0) const;
```
RickyRister: I recommend renaming this method so the name begins with `calc` or `compute` or something. With the current name, it sounds like `drop` is the verb. It took me a while to wrap my head around what this method is doing with that name.

RickyRister:
Compared to before, the cards will now cut off at the bottom if there's too many cards.

Screenshot 2026-04-05 at 11 50 04 PM

Screenshot 2026-04-05 at 11 49 59 PM

Is that intentional?

cockatrice/src/game/zones/select_zone.cpp
```diff
@@ -4,38 +4,202 @@
...
-    if (desiredHeight > totalHeight) {
-        if (reverse) {
-            y = index / ((totalHeight - cardHeight) / (cardCount - 1));
+    const qreal overlapPercent = SettingsCache::instance().getStackCardOverlapPercent();
+    return cardHeight * (100.0 - overlapPercent) / 100.0;
+}
+
+SelectZone::ZoneLayout SelectZone::computeZoneLayout(const StackLayoutParams &params)
+{
+    if (params.cardCount <= 0)
```
RickyRister: Zach wants us to always put braces around if blocks.

cockatrice/src/game/zones/select_zone.cpp
```diff
@@ -4,38 +4,202 @@
...
+        }
+    }
+}
+
+void SelectZone::layoutCardsVertically(const StackLayoutParams &params, bool hideOverflow)
+{
+    const auto &cards = getLogic()->getCards();
+    if (cards.isEmpty() || params.cardCount <= 0)
+        return;
+    if (params.cardCount > static_cast<int>(cards.size()))
```
RickyRister: nit: I'm pretty sure you don't need to explicitly cast here

DawnFire42:
> Compared to before, the cards will now cut off at the bottom if there's too many cards.

> Is that intentional?

Yes it is intentionnal, this specific behavior for clipping is mainly because the command zone is too small to begin with, simply adding a second card made them move all to the top and I couldn't click any of the cards behind the top one. But we can prevent it in the stack and handzone since they just don't have that issue.

RickyRister:
> But we can prevent it in the stack and handzone since they just don't have that issue.

Yes, please try not to visibly change existing behavior in this refactor.

RickyRister (changes requested):

DawnFire42:
> > But we can prevent it in the stack and handzone since they just don't have that issue.
> 
> Yes, please try not to visibly change existing behavior in this refactor.

My problem there was actually a larger architectural issue, thanks for flagging it. This should be way cleaner now. It still has the ability to clip cards and such, its just not called by any zones yet. So the function has all it needs for future edge cases without changing how things appear in current zones.

DawnFire42:
I updated the commit message to be accurate given my changes.

DawnFire42:
> I updated the commit message to be accurate given my changes.

I wonder if you prefer that I update my existing commits when making changes or you would prefer me adding more commits? Also, feel free to ask for me to separate stuff in multiple commits if you want me to. This refactor itself might be too chunky for a single commit. I could also separate the yet unused features in a separate commit or keep them for a later PR. Just let me know!

RickyRister:
Make new commits for updates that are in response to comments. We squash merge at the end anyways.
Rebase onto master frequently in order to keep the branch up to date.

> I could also separate the yet unused features in a separate commit or keep them for a later PR

I prefer PRs to be as minimal as possible, but other team members may think differently. 
I'll defer to you to use your best judgement, unless there is scenario where two responsibilities are clearly being mixed in the same PR, at which point I'll comment.

cockatrice/src/game/zones/select_zone.cpp
```diff
@@ -4,38 +4,207 @@
 #include "../board/card_item.h"
 #include "../game_scene.h"
 
-#include <QDebug>
+#include <QGraphicsRectItem>
 #include <QGraphicsSceneMouseEvent>
+#include <QtMath>
 
-qreal divideCardSpaceInZone(qreal index, int cardCount, qreal totalHeight, qreal cardHeight, bool reverse)
+namespace
```
RickyRister, line 11: I don't think we've used anonymous namespaces in this project yet. We currently just use `static` functions for internal linkage. 
If you think there's a benefit to moving over to using anonymous namespaces, then I suppose we can start with this one.

cockatrice/src/game/zones/select_zone.cpp
```diff
@@ -4,38 +4,207 @@
...
+    }
+    qreal stackHeight = (params.cardCount - 1) * effectiveOffset + params.cardHeight;
+    qreal start = (stackHeight <= params.totalHeight) ? (params.totalHeight - stackHeight) / 2.0 : 0.0;
+    return {effectiveOffset, start};
+}
+
+SelectZone *SelectZone::findOwningSelectZone(const QGraphicsItem *card)
+{
+    QGraphicsItem *parent = card ? card->parentItem() : nullptr;
+    if (!parent)
```
RickyRister, line 59: same comment about always using braces for if blocks

RickyRister (commented):
Can confirm that behavior now matches existing behavior

Screenshot 2026-04-15 at 10 57 08 PM

cockatrice/src/game/zones/select_zone.cpp
DawnFire42: There is no specific need for it so I'll follow the existing patterns

DawnFire42:
I made PR #6843 to fix the Arch build error. The Ubuntu errors might fix themselves on tests rerun, but I'll have to wait until the first fix is merged.

RickyRister (approved):

PR #6777, opened 2026-04-06
Properly reset default theme.

### LINKED ISSUE #6734: [Windows] Default theme started looking like Win95
I'm not sure exactly when it started happening (@ebbit1q should know more since because he said it started doing this after his fix), but recently the Default theme on Windows started looking like this, which apparently isn't intended (and at least in my personal opinion doesn't look great):

Image

BruebachL:
Took 9 minutes

## Related Ticket(s)
- Fixes #6734

- this
- and this

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -348,9 +348,9 @@ void ThemeManager::themeChangedSlot()
         qApp->setStyle(QStyleFactory::create("Fusion"));
         qApp->setPalette(createDarkGreenFusionPalette());
     } else {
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/theme_manager.cpp
```diff
@@ -347,8 +347,11 @@ void ThemeManager::themeChangedSlot()
     } else if (themeName == FUSION_THEME_NAME_DARK) {
         qApp->setStyle(QStyleFactory::create("Fusion"));
         qApp->setPalette(createDarkGreenFusionPalette());
+    } else if (themeName == NONE_THEME_NAME) {
```
ebbit1q: the basic style should be the same on custom themes as on the default theme

PR #6778, opened 2026-04-06
Set OracleWizard style to "Modern" instead of "Aero"

### LINKED ISSUE #6600: Support for fusion/dark theme in Oracle
The new Fusion theme is great and it allows to adopt to system preferences, too!
Currently, Oracle does not support it well and in dark mode it becomes very trick to use and a lot of text can not be read because the background color is not changing.

**Additional Context**

Image

BruebachL:
## Related Ticket(s)
- Fixes #6600

On Windows, the QWizard Style is set to "Aero" which has hardcoded background colors and thus does not play nice with the dark mode.
No one on Windows actually uses Aero (instead they use Modern or Windows11), so the rendering isn't even done in Aero mode, we lose pretty much nothing by switching this to the "Modern" QWizard style imo.

- Set the wizard style to modern on windows.

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -179,7 +179,7 @@ QBrush ThemeManager::loadExtraBrush(QString fileName, QBrush &fallbackBrush)
 
 static inline QPalette createDarkGreenFusionPalette()
 {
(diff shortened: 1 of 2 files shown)


ebbit1q (approved):
sure, let's test this in beta

PR #6780, opened 2026-04-07
[Game] Refactor: explicitly pass params from card counter actions

RickyRister:

The way we pass params from the card counter QActions into the connected `PlayerAction` methods is opaque. We set a single int as data in the QAction in order to encode two params, and then use modulo to parse out the two params on the receiver's side.

I assume the code was written before Qt supported lambdas in connections. The code would be easier to understand if we just called the appropriate methods with the appropriate args in a lamda.

- Split `PlayerActions ::actCardCounterTrigger` into three methods that explicitly take counterId as a param
- rewire card counter actions


ebbit1q (approved):
this needs to be rebased with the selected cards method but looks good

PR #6784, opened 2026-04-08
Set style to nullptr?

BruebachL:
Took 7 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


ebbit1q:
I remember this being the case at some point and it not working, but maybe with the packaging changes it now does?

PR #6785, opened 2026-04-08
Utility method to check if a theme is supposed to be in dark or light mode.

BruebachL:
We only perform light/dark logic in the themeChanged slot. With the introduction of light/dark mode, other UI components care about the current theme preference. Additionally, this allows custom themes to cue off of light/dark mode by adding magic strings to their names.

- Add a utility method that checks for (Dark), (Light) or (System Default) at the end of theme names
- Fallthrough logic for detection, preferring hard coded dark/light and then falling through to system default on compatible themes (i.e. the ones we ship with, "Default" and "Fusion (System Default)"
- Repolish visible widgets on theme change, which fixes the issue of themes not applying to every visible widget on theme change

## Pictures
Partially applied theme before fix:
image

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -108,6 +108,31 @@ void ThemeManager::ensureThemeDirectoryExists()
     }
 }
 
+bool ThemeManager::isDarkMode()
+{
+    auto themeName = SettingsCache::instance().getThemeName();
+    // Explicit Dark Mode
+    if (themeName == FUSION_THEME_NAME_LIGHT || themeName.endsWith("(Light)")) {
+        return false;
+    }
+    // Explicit Light Mode
+    if (themeName == FUSION_THEME_NAME_DARK || themeName.endsWith("(Dark)")) {
+        return true;
+    }
+
+    // Auto detection on compatible Qt versions
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 5, 0))
+    if (QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark &&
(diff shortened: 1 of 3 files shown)


cockatrice/src/interface/theme_manager.cpp
```diff
@@ -108,6 +108,39 @@ void ThemeManager::ensureThemeDirectoryExists()
...
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 5, 0))
+    if (QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark &&
+        (themeName == NONE_THEME_NAME || themeName == FUSION_THEME_NAME || themeName.endsWith("(System Default)"))) {
+        return true;
+    } else {
+        return false;
+    }
+#endif
+    // Default to light mode
+    return false;
```
ebbit1q, line 138: what about custom themes that set the background to be dark with css?

ebbit1q: eg chat windows use the background color to figure out if it's a dark theme to determine which hardcoded colors to use for some text

BruebachL: They can hardcode their decision by explicitly specifying their theme to be (Dark)
I'd rather give theme developers a fixed option (for now and any future systems) than do guesswork tbh.

BruebachL: If we include it as a heuristic detection method (as opposed to the Qt system hint) it becomes a question of who we trust more: The universal system hint or our custom detection

RickyRister (approved):

ZeldaZach (approved):

PR #6795, opened 2026-04-10
add ccache eviction for files older than 7 days

ebbit1q:
also clean up some of the scripts to be more internally consistent


BruebachL (approved):
Yolo

PR #6797, opened 2026-04-10
CI: Add timeout to job matrixes

tooomm:
In rare cases some jobs are getting stuck and keep running for some hours:
https://github.com/Cockatrice/Cockatrice/actions/runs/23957724556/job/69926786339
https://github.com/Cockatrice/Cockatrice/actions/runs/23971364796/job/69921261517
https://github.com/Cockatrice/Cockatrice/actions/runs/23976334581/job/69940881542
https://github.com/Cockatrice/Cockatrice/actions/runs/24079358656/job/70235716695
https://github.com/Cockatrice/Cockatrice/actions/runs/24183035379/job/70584584513
https://github.com/Cockatrice/Cockatrice/actions/runs/24265301425/job/70858847782

The default job timeout is 6h.

Introduce a timeout of roughly 1h plus the usual running time:
- Timeout of 70min for Linux builds
- Timeout of 100min for vcpkg builds (macOS + Win)
  image

## Screenshot

Also in this PR itself, the vcpkg step is having the same issue. The timeout value is working as expected and cancelles after 1h 40min already:
image

Instead after 6h only:
image

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -156,6 +156,7 @@ jobs:
     needs: configure
     runs-on: ubuntu-latest
     continue-on-error: ${{matrix.allow-failure == 'yes'}}
+    timeout-minutes: 70
     env:
       NAME: ${{matrix.distro}}${{matrix.version}}
       CACHE: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}   # directory for caching docker image and ccache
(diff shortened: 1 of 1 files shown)


ebbit1q:
it seems to always be caused by the setup vcpkg part which often takes just 1 minute, why not add the timeout there?

tooomm:
> it seems to always be caused by the setup vcpkg part which often takes just 1 minute, why not add the timeout there?

It's currently failing there for us, but there could be many other reasons that cause similar issues in other places in the future.
I did not try to fix this particular issue, but find a general enhancement that also helps with future problems e.g. stemming from upstream actions.
I tried to avoid hours of runtime that have the same output after 1h or 6h, hence failing them early.
If our build runs like ~1h longer than usual, something is probably off and it will never succeed.

Timeouts could also be set to individual steps, yes.

<br>

Once Windows builds use compiler caching as well, we can reduce the timeout value for the vcpkg based builds.

<br>

The `ubuntu-slim` runner images I added in the other PR uses a 15min timeout value as default btw.

ebbit1q (approved):
sure, thanks

PR #6798, opened 2026-04-10
CI: Use `ubuntu-slim` for simple jobs

tooomm:
- Use new `ubuntu-slim` runner targeted for automation tasks
- Add to our simple workflows:
  - desktop-build (Config job only)
  - desktop-lint
  - translations-pull
  - translations-push
  - web-lint

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -46,7 +46,7 @@ concurrency:
 jobs:
   configure:
     name: Configure
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     outputs:
       tag: ${{steps.configure.outputs.tag}}
       sha: ${{steps.configure.outputs.sha}}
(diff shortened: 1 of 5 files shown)


ebbit1q:
other than translations-push.yml which relies on docker this seems ready to go

tooomm:
> other than translations-push.yml which relies on docker this seems ready to go

https://github.com/chuhlomin/render-template is a Docker action, yes.

Unfortunately, those are not supported by the new slim runners targeted for simpler, automation tasks: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#single-cpu-runners
>The container for `ubuntu-slim` runners runs in unprivileged mode. This means that some operations requiring elevated privileges—such as mounting file systems, using **Docker-in-Docker**, or accessing low-level kernel features—are not supported.

<br>

Going back to `ubuntu-latest` would resolve the issue.
I'm checking some other options first though.

ebbit1q:
it makes sense to just go back to the not slim runner

tooomm:
The author of the action made now a non-docker version compatible with `ubuntu-slim` available as well.

ebbit1q (approved):
cool, thanks!

PR #6799, opened 2026-04-11
CI: Allow failing of ccache deletion step

### LINKED ISSUE #6791: ci builds can fail because of caches getting evacuated in a race condition
see this run:
https://github.com/Cockatrice/Cockatrice/actions/runs/24209189445/job/70678110708

builds should ignore such failures simply, if the cache is gone it is gone

tooomm:
## Related Ticket(s)
- Fixes #6791

Deletion of the used cache in the job was failing because it could for some reason not be found.

- Allow this step to fail and proceed with the rest of the steps in the job nonetheless
  - If the cache deletion is not possible, the next step will run but not be able to save the cache and print a warning
  - If the cache is not found for deletion, the next step will upload the cache and pass

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -456,7 +456,7 @@ jobs:
         continue-on-error: true
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
excellent, thanks

PR #6802, opened 2026-04-11
CI: Print success msg if cache deletion did succeed

tooomm:
image

<br>

gh command silently succeeds. Adding confirmation for clarity.


ebbit1q (approved):
thanks

PR #6803, opened 2026-04-11
replace windows11 with fusion for the view windows

ebbit1q:
## Related Ticket(s)
- #6734

users on windows11 have been reporting the entire viewport breaking when a card view window appears in the player

- replace the theme of just the view with fusion if it's found to use windows11


ebbit1q:
does not work

PR #6804, opened 2026-04-12
add some output to the eviction step to measure effectiveness

ebbit1q:
## Related Ticket(s)
- followup to #6795

the ccache eviction happens without any feedback

- add a line that says what the oldest age of a file is and how many files there are, effectively doubling the amount of processing of this step

### DIFF
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -273,6 +273,11 @@ echo "::endgroup::"
 if [[ $USE_CCACHE ]]; then
   if [[ $CCACHE_EVICTION_AGE ]]; then
     echo "::group::evict ccache files older than $CCACHE_EVICTION_AGE"
+    ccache_dir=$(ccache --get-config cache_dir)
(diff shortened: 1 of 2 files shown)


ebbit1q:
this also adds cache eviction to the arch runner, which I forgot about before

ebbit1q:
macos doesn't have the same find and mapfile command, I don't know what to do about this as I don't use mac

tooomm:
> macos doesn't have the same find and mapfile command, I don't know what to do about this as I don't use mac

@RickyRister Can you help here?

tooomm:
The fresh [ccache 4.14](https://ccache.dev/releasenotes.html#_ccache_4_14) added output with details to the option:

>Cleanup and eviction operations now print the amount of data and number of files removed.

We are already using that version on macOS and Arch, and a bit older ones on other Linux distro - it'll take a bit until the new version is included in all CI builds. FYI
image
image

<br>

While not exactly the same as tried in this PR, it's very helpful and works across OS. We could add more and come back if needed. Thanks ebbit for having a look and trying to improve the initial situation!
Closing this PR.

PR #6806, opened 2026-04-12
Initial implementation completion and refactor

seavor:
## Related Ticket(s)
- Handled on separate board

Implementation of the websocket/protobuf layer for the webclient

- 100% implementation of the protobuf layer
- refactoring of the websocket layer


ebbit1q:
I think I'd like to have the webclient stuff be moved to a separate repo, separating the cpp and web code and ticket boards makes sense as there will be almost no overlap, the web client can take the protocol files as a submodule or similar

ZeldaZach (approved):

ZeldaZach:
I'm not disagreeing with you on moving to a separate repo, but lets get it to a workable state then move it over @ebbit1q

Issue #6807, opened 2026-04-12
Game not Playable Once Opening Side Board


BuzzfeedBlue:
As soon as I open the side board the screen goes mostly white. I am on the beta branch. A friend of mine also on beta has this issue, and a third person who is on beta does not have the issue. I cannot currently play the game with this occurring.

I am on windows 11 25H2, nvidia 3080. Resetting the program and my pc both have not fixed the issue. Have not fully uninstalled or re-installed the program.

<img width="2129" height="1150" alt="Image" src="https://github.com/user-attachments/assets/71173130-d6e1-498e-8ce7-6697204080a5" />

ebbit1q:
this is a known issue affecting the windows11 theme, used by default on qt versions 6.8 and newer on windows 11, it does not affect windows 10, there are two workarounds for this:
1. go to the settings and change the theme to "fusion"
2. run cockatrice with the windowsvista theme, eg `cockatrice.exe -style windowsvista`

PR #6809, opened 2026-04-12
CI: More config deduplication

tooomm:
`QT_VERSION`
`QT_MODULES`
`CCACHE_EVICTION_AGE`

Plus some other small changes

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -355,7 +355,7 @@ jobs:
           msbuild-architecture: x64
 
       - name: Setup ccache
-        if: matrix.use_ccache == 1 && matrix.os == 'macOS'
+        if: matrix.use_ccache == 1 && runner.os == 'macOS'
         run: brew install ccache
 
       - name: Restore compiler cache (ccache)
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -350,7 +350,6 @@ jobs:
 
       - name: Add msbuild to PATH
         if: matrix.os == 'Windows'
-        id: add-msbuild
         uses: microsoft/setup-msbuild@v3
         with:
           msbuild-architecture: x64
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -410,7 +410,7 @@ jobs:
           path: ${{ github.workspace }}/Qt
           key: thin-qt-macos-${{ matrix.soc }}-${{ steps.resolve_qt_version.outputs.version }}
 
-      - name: Install Qt ${{ env.QT_VERSION }} (Windows)
+      - name: Install Qt ${{ steps.resolve_qt_version.outputs.version }} (Windows)
         if: matrix.os == 'Windows'
         uses: jurplel/install-qt-action@v4
         with:
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -339,7 +339,7 @@ jobs:
       # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
       CCACHE_SIZE: 550M
       CCACHE_EVICTION_AGE: 7d
-      QT_VERSION: 6.11
+      QT_VERSION: 6.10
       QT_MODULES: qtimageformats qtmultimedia qtwebsockets
 
     steps:
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -349,7 +349,7 @@ jobs:
           submodules: recursive
 
       - name: Add msbuild to PATH
-        if: matrix.os == 'Windows'
+        if: runner.os == 'Windows'
         uses: microsoft/setup-msbuild@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -463,7 +452,6 @@ jobs:
           MACOS_CI_KEYCHAIN_PWD: ${{ secrets.PROD_MACOS_CI_KEYCHAIN_PWD }}
           DEVELOPER_DIR: '/Applications/Xcode_${{matrix.xcode}}.app/Contents/Developer'
           TARGET_MACOS_VERSION: ${{ matrix.override_target }}
-          CCACHE_EVICTION_AGE: ${{ matrix.ccache_eviction_age }}
```
tooomm, line 466: Despite this being listed here, there are no eviction entries for the `build-vcpkg` matrix at all @ebbit1q?

Added it to the general environment variables there for now as well.

.github/workflows/desktop-build.yml
```diff
@@ -159,7 +159,7 @@ jobs:
     timeout-minutes: 70
     env:
       NAME: ${{matrix.distro}}${{matrix.version}}
-      CACHE: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}   # directory for caching docker image and ccache
+      CACHE: ${{github.workspace}}/.cache/${{matrix.distro}}${{matrix.version}}    # directory for caching docker image and ccache
```
ebbit1q, line 162: I'm not sure why but it seems like very time you edit this file an extra space gets added here

.github/workflows/desktop-build.yml
ebbit1q: I'm not sure what you mean, this variable is only recognised by our compile script

.github/workflows/desktop-build.yml
```diff
@@ -352,6 +338,9 @@ jobs:
       # Cache size over the entire repo is 10Gi:
       # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
       CCACHE_SIZE: 550M
+      CCACHE_EVICTION_AGE: 7d
+      QT_VERSION: 6.11.*
```
ebbit1q, line 342: because of #6807 we are actually considering moving qt back to 6.6 for windows

ebbit1q:
this conflicts with #6804

.github/workflows/desktop-build.yml
tooomm: It looks you found another way and decided not to?
#6810

.github/workflows/desktop-build.yml
tooomm: Once I stumble upon inconsistencies, I try to fix them.

We use 4 spaces in the workflow file when separating comments next to config options. Here it was only 3.

.github/workflows/desktop-build.yml
tooomm: Sorry, I noticed that you did not add it to the windows build (because we do not use ccache there at the moment) and the value was empty.
For macOS it's added in the matrix and all works. 👍 

We can still deduplicate here and remove it from the individual matrix entries and move it to a single, general env variable. Same as the other ccache parameters despite them not being used for Windows builds.
It's then in sync with how you did it for Linux builds as well.

.github/workflows/desktop-build.yml
ebbit1q: that is correct

.github/workflows/desktop-build.yml
ebbit1q: why 4 spaces?

.github/workflows/desktop-build.yml
ebbit1q: the reason why I put the variable in the general environment on the linux build is because we have two compile steps, here we have just one so I put it in just the related step, we could move a lot of these variables only recognised in our scripting into the general scope but that would have no benefit either

.github/workflows/desktop-build.yml
tooomm: No specific reason about the exact number.
I think just doing one is not enough and the comment is visually not separated enough from the config to stand out. In particular if lines above or below are longer than the one with the comment.

Fine with any other number as long as it's uniformly applied and the comment is well readable.

.github/workflows/desktop-build.yml
tooomm: My motivation to move it is to reduce duplication in the config.

Instead of having the same value listed for 4/5 matrix entries for build-vcpkg, it's listed once in the job environment.
The other ccache relevant configurations sit there already, too (CCACHE_SIZE, CCACHE_DIR) - even if Windows does not need it (right now).

.github/workflows/desktop-build.yml
ebbit1q: I greatly prefer comments to be on the preceding line unless they are 3 words or less

ebbit1q: it's to limit the line length mostly

.github/workflows/desktop-build.yml
ebbit1q: why move it to the general scope instead of just the step though?

PR #6810, opened 2026-04-12
add workaround for windows11 theme

### LINKED ISSUE #6807: Game not Playable Once Opening Side Board

ebbit1q:
## Related Ticket(s)
- Fixes https://github.com/Cockatrice/Cockatrice/issues/6807

I tried to find if there is a way to keep the new windows11 theme, but it's too much effort, people are just stuck with the vista theme for now, the other option is to downgrade the used version in ci, but that does not fix compiled builds for example, I think it's a better idea to move forward in qt version as otherwise we'll just be stuck on 6.6 forever.

- detect if windows11 theme is used and replace it with windowsvista

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -91,6 +91,10 @@ struct PaletteColorInfo
 ThemeManager::ThemeManager(QObject *parent) : QObject(parent)
 {
     defaultStyleName = qApp->style()->objectName();
+    // FIXME workaround for windows11 style being broken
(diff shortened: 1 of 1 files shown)


BruebachL (approved):
Works, thank you.

PR #6814, opened 2026-04-15
[EDHRec] Fix unintentional navigation on card right click

BruebachL:
Left clicking is supposed to navigate to the card whereas right clicking opens the card menu so you can add the card to your deck.
We hook up to cardClicked for the navigation, which we always emit, whether it's a right or left click. This makes the right-click menu inaccessible.

- Don't double emit cardClicked
- pass along the mouseEvent so we can check if it's a left or right click.


ebbit1q (approved):
seems good, thanks

PR #6818, opened 2026-04-18
Add -R option in Windows NSIS script for silent upgrade

BruebachL:
Uninstalling and then reinstalling with the default options on every upgrade is tedious. If people launch the client updater, they already *know* they want to uninstall/install (i.e. upgrade) so they shouldn't be prompted two separate times, which often times leads to more confusion than it alleviates ("Will this keep my decks?")

- Add the /R switch to the Windows NSIS so we can do a silent unattended uninstall/install cycle
- Invoke it like this from the updater.

### DIFF
--- a/cmake/NSIS.template.in
+++ b/cmake/NSIS.template.in
@@ -35,7 +35,7 @@ Var ReinstallMode
 !define MUI_PAGE_CUSTOMFUNCTION_PRE SkipIfReinstall
 !insertmacro MUI_PAGE_LICENSE "${NSIS_SOURCE_PATH}\LICENSE"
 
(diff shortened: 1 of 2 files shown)


ebbit1q:
how will this interact with previously installed versions of the application that will not have this flag available?

BruebachL:
> how will this interact with previously installed versions of the application that will not have this flag available?

The non-flag "Uninstall then Install" flow is still available in this version. Older clients launching this installer will just use that (i.e. the non-argument non-silent install)

Older installers launched by this new client won't care about the extra flags.

ZeldaZach (approved):

PR #6822, opened 2026-04-19
[GameSelector/Filters] Properly sync toolbar and dialog.

### LINKED ISSUE #6601: Game filter not in sync with new quick filter options
**System Information:**

Client Version: 2.11.0-beta.49 (2026-02-08)
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_DE
Install Mode: Standard

The new quick filter fields above are not synced with the actual options in the `Filter` dialog.
That leads to bad UX and displayed options that are not matching, as well as results that are unexpected:
Image

If you switch the game type and apply the filter it gets even more confusing.
The results do update and show now only Modern, but the quick filter still state limited. The hide filter is also not synced and obviously full games do show - I did open the filter dialog here again to show the mismatch better:
Image

BruebachL:
## Related Ticket(s)
- Fixes #6601

Changing one doesn't necessarily change the other and the clear filter button does not affect the toolbar.

- Add a filtersChanged() signal to the global setter
- Route all set() calls through that unified setter instead
- Hook up some other signals to filtersChanged()
- Fix some typos/mismatches

### DIFF
--- a/cockatrice/src/interface/widgets/server/game_selector_quick_filter_toolbar.h
+++ b/cockatrice/src/interface/widgets/server/game_selector_quick_filter_toolbar.h
@@ -18,6 +18,24 @@ public:
                                             TabSupervisor *tabSupervisor,
                                             GamesProxyModel *model,
                                             const QMap<int, QString> &allGameTypes);
+    void syncFromModel();
+    void applyFilters(std::function<void(bool &,
+                                         bool &,
+                                         bool &,
+                                         bool &,
+                                         bool &,
+                                         bool &,
+                                         bool &,
+                                         QString &,
+                                         QStringList &,
+                                         QSet<int> &,
+                                         int &,
+                                         int &,
+                                         QTime &,
+                                         bool &,
+                                         bool &,
+                                         bool &,
+                                         bool &)> mutator);
     void retranslateUi();
 
 private:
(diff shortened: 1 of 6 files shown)


cockatrice/src/interface/widgets/server/game_selector_quick_filter_toolbar.h
```diff
@@ -18,6 +18,24 @@ class GameSelectorQuickFilterToolBar : public QWidget
...
+                                         QString &,
+                                         QStringList &,
+                                         QSet<int> &,
+                                         int &,
+                                         int &,
+                                         QTime &,
+                                         bool &,
+                                         bool &,
+                                         bool &,
+                                         bool &)> mutator);
```
RickyRister, line 38: you know, at this point, maybe we should just refactor it so that we have a struct that stores the values of the fields.

BruebachL: We should but I also wanna release

tooomm: Leave a todo comment?

RickyRister (approved):
LGTM, but we're refactoring this thing first thing after release.

PR #6823, opened 2026-04-19
[DeckAnalytics] Add a checkbox to include/exclude sideboard for calculation

BruebachL:
Currently, the deck list statistics are run for your entire deck but many people do not care about the sideboard when making evaluations, since you play with the mainboard.

- Add a checkbox to toggle this behavior.


ebbit1q (approved):
nice, seems fine, thanks

PR #6829, opened 2026-04-21
[Game] Add a coin flip shortcut.

### LINKED ISSUE #6826: Coin Flip Shortcut?
Maybe a big of a selfish request but could we get a shortcut option to flip a coin for us who run the duo who flips coins until they lose which becomes a little annoying with how it works RN.

BruebachL:
## Related Ticket(s)
- Fixes #6826 

- Ctrl+Shift+I 
- Ctrl + I is roll dice


ebbit1q (approved):
seems fine, I'm a bit worried about crowding the menu, I feel like if we had an option to just have the shortcut available for power users but have people using the menu go through the dialog (which also has the option to roll multiple coins) would be optimal, adding it to the menu might confuse new users into thinking they cannot flip coins with the roll die dialog

ZeldaZach (approved):

BruebachL:
> seems fine, I'm a bit worried about crowding the menu, I feel like if we had an option to just have the shortcut available for power users but have people using the menu go through the dialog (which also has the option to roll multiple coins) would be optimal, adding it to the menu might confuse new users into thinking they cannot flip coins with the roll die dialog

You can't have a shortcut without adding it to some kind of menu

PR #6831, opened 2026-04-21
ensure judges are omniscient even when not spectating

ebbit1q:
## Related Ticket(s)
- #3531
- #5053

judge users are able to use judge mode to join a game as a player as well, judge players are able to take all actions a judge spectator can but do not receive the same events because they check for that on the server side

this leads to strange behavior where clients while judge players behave as if they see all information while they actually don't, eg when a card is moved to a hand they will think the id of that card is correct, even though it actually isn't (#6125)

- the server will send omniscient events to all judges


ZeldaZach (approved):

ebbit1q:
@ZeldaZach please rebuild servatrice with this change, I expect judge player users to immediately notice this, I already warned them that the way they were using this was completely unintentional

whenever a user joins a match with the judge icon, whether as player or not, it means they are omniscient and can see everyone's hands

PR #6835, opened 2026-04-27
CI: Add cleanup script for beta releases

tooomm:
Beta releases and related tags need to be manually cleaned up after stable releases and pile up. 
Currently we have 52 beta pre-releases for the upcoming release published. Deletion is annoying. 

- Automation script which does
  - Delete all published **pre-releases**
  - Stable releases are not touched
  - Draft releases of both types are filtered out
  - Double check that the git tag contains a "beta" string before marking for deletion
  - ~~Delete the corresponding git tags~~
    Looking at https://github.com/Cockatrice/Cockatrice/tags?after=2024-12-19-Development-2.10.0-beta.5, it seems we keep the tags of old pre-releases
- Only runs when
  - A new release is published (not created)
  - This release is a stable release (not a pre-release), or a pre-release got elevated to a stable release
  - Targeting `master` branch
  - Repo is not a fork
  - Manually triggered against a tag from the UI
- Runs in dry run mode per default
  - Dry run flag (`DRY_RUN`) needs to be toggled to false to do actual deletion.
    We can switch to deletion by default once we are comfortable.
  - Triggering from the UI allows to switch dry run mode off.

### DIFF
--- a/.github/workflows/release-cleanup.yml
+++ b/.github/workflows/release-cleanup.yml
@@ -1,0 +1,136 @@
+name: Cleanup Beta Releases
+
+on:
+  release:
+    types: [released]    # triggers only on published stable releases
+  pull_request:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/release-cleanup.yml'
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Run in dry mode"
+        required: true
+        type: choice
+        options:
+        - true
+        - false
+        default: 'true'
+
+jobs:
+  Deletion:
+    if: >
+      github.repository_owner == 'Cockatrice' &&
+      (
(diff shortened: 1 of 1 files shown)


ebbit1q:
I just deleted the betas, it took a couple of minutes, this would be nice to have but I have no ability to maintain this script

ebbit1q:
I've also apparently deleted the tags for beta.34 and .54 I have no idea how that happened, we don't delete tags on release, only entries on the release page

ebbit1q:
another thing worth noting: our release template includes a comment with instructions to delete released beta tags, which we don't do

ebbit1q:
at least, we haven't deleted a beta tag since 2.8.0, should we delete beta tags?

Issue #5500, opened 2025-01-19
Changes in `Manage sets...` dialog do no longer reload the database


tooomm:
**System Information:**
Client Version: 2.10.0-beta.31 (2025-01-18)
Client Operating System: Windows 11 Version 24H2
Build Architecture: 64-bit
Qt Version: 6.6.3
System Locale: en_DE
Install Mode: Standard

_______________________________________________________________________________________

Adjustments to the available sets in the `Manage sets...` dialog do no longer trigger a DB reload.
Enabling or disabling entries and then saving/closing the dialog (`Ok`) should apply the changes and reload the card database.

_______________________________________________________________________________________

**Steps to reproduce:**
 - Open `Manage sets...` dialog
 - Add or remove sets
 - Click `Ok` to save & exit
 - Changes do not apply and manual reload (or app restart) is required

RickyRister:
works for me

https://github.com/user-attachments/assets/97b4a7a7-2c82-40e0-8676-a6dd5c16a971

tooomm:
I downloaded the newest build from master and updated the card db because I wanted to record a vid about the color filter again.

It now works for me, too. So maybe some glitch or an issue with a former beta release on Windows (or on my end. 🤷

tooomm:
I had a feeling that this was still weird and not working fully as intended...

See the vid with the newest beta build (`Client Version: 2.10.0-beta.42 (2025-02-08)`) on Win 11:

https://github.com/user-attachments/assets/6f6e38bd-dc46-4c06-b2fe-65fe76dc5f01

ebbit1q:
the issue is that enabling a set that was not enabled at startup does not have any effect, reloading the card database causes it to take effect again, disabling a set does work immediately

ebbit1q:
the problem is that the database parser uses the enabled sets to decide to omit printings, this leads to a catch 22 on first launch: to put the list of existing sets in the settings the db has to be loaded, to enable cards in this first loading of the db there are no sets that are enabled causing no printings to be added, to add cards sets have to be placed in the settings first, only by loading the db twice will cards be in the db

tooomm:
After https://github.com/Cockatrice/Cockatrice/pull/6837, reloading works again.

However, when disabling all sets and confirming, the `Welcome` dialog is triggered and it forces all sets to be enabled again.
Maybe the save/ok button in the `Manage sets` dialog should be disabled if no set is selected?
Or we need to detect first starts differently.

PR #6837, opened 2026-04-29
Reload card db and notify enabled sets change on "Manage Sets" dialog save

BruebachL:
Is it ugly performance wise? Yes. Does it solve the issue? Yes.

Also adds a new fuse to the "New sets found" dialog that allows users to enable the set and all future sets, skipping the dialog, which will give users less of a chance to accidentally click no (since enabling the new set is pretty much always desired behavior)

## Screenshots

https://github.com/user-attachments/assets/603f9e54-791a-49d7-98b3-b16fb864fbb4

### DIFF
--- a/cockatrice/src/interface/widgets/dialogs/dlg_manage_sets.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_manage_sets.cpp
@@ -16,6 +16,7 @@
 #include <QPushButton>
 #include <QToolBar>
 #include <QTreeView>
+#include <QtConcurrentRun>
 #include <algorithm>
 #include <libcockatrice/card/database/card_database_manager.h>
 #include <libcockatrice/models/database/card_set/card_sets_model.h>
@@ -253,6 +254,17 @@ void WndSets::actSave()
     model->save(CardDatabaseManager::getInstance());
     SettingsCache::instance().setIncludeRebalancedCards(includeRebalancedCards);
     CardPictureLoader::clearPixmapCache();
+    const auto reloadOk1 = QtConcurrent::run([this] {
+        CardDatabaseManager::getInstance()->loadCardDatabases();
+        QMetaObject::Connection conn;
+
+        conn = connect(CardDatabaseManager::getInstance(), &CardDatabase::cardDatabaseLoadingFinished, this,
+                       [conn]() mutable {
+                           CardDatabaseManager::getInstance()->notifyEnabledSetsChanged();
+                           QObject::disconnect(conn);
+                       });
+        SettingsCache::instance().downloads().sync();
+    });
     close();
 }
 
--- a/cockatrice/src/interface/widgets/dialogs/dlg_manage_sets.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_manage_sets.cpp
@@ -254,15 +254,9 @@ void WndSets::actSave()
     model->save(CardDatabaseManager::getInstance());
     SettingsCache::instance().setIncludeRebalancedCards(includeRebalancedCards);
     CardPictureLoader::clearPixmapCache();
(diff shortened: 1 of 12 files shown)


ZeldaZach (approved):

cockatrice/src/interface/widgets/dialogs/dlg_manage_sets.cpp
```diff
@@ -253,6 +254,17 @@ void WndSets::actSave()
     model->save(CardDatabaseManager::getInstance());
     SettingsCache::instance().setIncludeRebalancedCards(includeRebalancedCards);
     CardPictureLoader::clearPixmapCache();
+    const auto reloadOk1 = QtConcurrent::run([this] {
+        CardDatabaseManager::getInstance()->loadCardDatabases();
+        QMetaObject::Connection conn;
+
+        conn = connect(CardDatabaseManager::getInstance(), &CardDatabase::cardDatabaseLoadingFinished, this,
+                       [conn]() mutable {
+                           CardDatabaseManager::getInstance()->notifyEnabledSetsChanged();
```
ebbit1q: do we have to notify it enabled sets changed? afaik reloading the db just throws away the old one so the notification wouldn't do anything?

BruebachL: Yes, we do. Breaks without because display models do not get notified (i.e. change only applies after the user manually enters a new card name/filter for example)

ebbit1q:
I tested this and it's much better than before, is it not possible to do the database reload in the background though?

I noticed that the visual database display does not immediately update when disabling a set, it requires a filter change still.

BruebachL:
> I tested this and it's much better than before, is it not possible to do the database reload in the background though?
> 
> I noticed that the visual database display does not immediately update when disabling a set, it requires a filter change still.

I mean, we do QtConcurrent::run which should run this in the background, right?

Give the visual database display a couple of seconds to refresh. See the video I attached. It works correctly for me after a bit. We only set the dirty() timer, we don't invalidate immediately.

ebbit1q:
how often does this timer get evaluated? it's been a while and while the deck editor does update immediately the visual database display still has not

ebbit1q:
note that it does work on enabling, just not disabling

BruebachL:
> note that it does work on enabling, just not disabling

Hm, you're right

ebbit1q:
I tested again and it does reload in the background, but clicking ok on the sets manager still incurs a significant delay, does it still try to send out the deck editor update signal on ok before reloading? could it be that because of parenting of the background loading it keeps the modal dialog around?

cockatrice/src/interface/window_main.cpp
```diff
@@ -1183,6 +1190,7 @@ void MainWindow::cardDatabaseNewSetsFound(int numUnknownSets, QStringList unknow
                        .arg(unknownSetsNames.join(", ")));
 
     QPushButton *yesButton = msgBox.addButton(tr("Yes"), QMessageBox::YesRole);
+    QPushButton *yesAlwaysButton = msgBox.addButton(tr("Yes, always enable"), QMessageBox::YesRole);
```
ebbit1q, line 1193: this seems like an extremely helpful addition that's worth mentioning in the pr body

BruebachL:
image

ebbit1q:
what I gather from this graph is the part that is actually occupying all this time is just the part where it is saving the set priority to disk?

BruebachL:
> what I gather from this graph is the part that is actually occupying all this time is just the part where it is saving the set priority to disk?

Yeah, that's my interpretation.

ebbit1q:
I tested what happens when on initial launch you "cancel" on the set manager popup, it results in no db reload :/

ebbit1q:
this would fix #5500

PR #6839, opened 2026-04-30
CI: Ensure Docker runs + publishes images only on stable releases

tooomm:
The docker workflow was run on all tags.
That lead to a v3 image being released two weeks back because such a tag for a stable release was already created (seems removed in the meantime again - image stayed), see https://github.com/Cockatrice/Cockatrice/pkgs/container/servatrice/806386408?tag=2026-04-19-Release-3.0.0

- Trigger on stable release events instead all tags
- Adjust config to maintain similar behavior and checks than before

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -30,6 +30,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Docker metadata
+        if: ${{ github.repository_owner == 'Cockatrice' }}
         id: metadata
         uses: docker/metadata-action@v6
         with:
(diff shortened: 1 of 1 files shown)


ebbit1q (approved):
it was me that removed the tag, the release was not ready yet at the time and I deleted the tag

I noticed this issue on the flathub repo as well, however those prs can simply be closed, this immediately pushes

thank you for fixing this

tooomm:
It seems that there were also various untagged versions created somehow:
https://github.com/Cockatrice/Cockatrice/pkgs/container/servatrice/versions

Not sure why that happened.

Can you cleanup all versions in GHCR besides the very first tagged one please, @ebbit1q?

PR #6840, opened 2026-04-30
CI: Adjust annotations level for Docker metadata

tooomm:
Annotations are added at `manifest` level per default:
image

For GHCR that seems not to be correct.
https://github.com/docker/metadata-action#annotations

- Add on `index` level instead for our multi-arch image
  image

The data should then be displayed for our package correctly, too 🤞
image


ebbit1q (approved):
sure

Issue #6841, opened 2026-04-30
Fedora 44 failing on debug build


SlightlyCircuitous:
**System Information:**

Fedora 44
gcc: 16.0.1
cmake: 4.3.0
Qt: 6.10.3

_______________________________________________________________________________________

I opened https://github.com/Cockatrice/Cockatrice/pull/6836 in order to add a Fedora 44 build to our build system given its recent release. Since it is the latest Fedora, I set it up to build debug and test. This failed twice (see [example here](https://github.com/Cockatrice/Cockatrice/actions/runs/25061063928/job/73452932060?pr=6836)). ebbit suggested that I disable the debug build for now and that circumvented the issue. Fedora 44 is the only system we build for that [uses gcc 16](https://github.com/Cockatrice/Cockatrice/pull/6836#issuecomment-4352425446), which may be related. 

When I try to compile Cockatrice in my Fedora 44 VM, [it fails for a completely different reason when configured with the debug flag](https://github.com/Cockatrice/Cockatrice/pull/6836#issuecomment-4353044376). Compiling without debug works fine.

_______________________________________________________________________________________

**Steps to reproduce:**
 - Remove the `tests: skip` line under Fedora 44 from desktop-build.yml
 - Run the `Build Desktop` Github Action

ebbit1q:
you should install googletest separately perhaps to see if that fixes your other issue

SlightlyCircuitous:
That doesn't seem to have helped, but bumping the version in `cmake_minimum_required()` to 3.10 in `cmake/gtest-CMakeLists.txt.in` allowed me to download googletest as normal so that's fine too. I assume due to changes in CMake 4.0, any file that uses `cmake_minimum_required()` to set a minimum lower than 3.5 won't run properly and that was blocking the download:

> Changed in version 4.0: Compatibility with versions of CMake older than 3.5 is removed. Calls to `cmake_minimum_required(VERSION)` or `cmake_policy(VERSION)` that do not specify at least 3.5 as their policy version (optionally via ...<policy_max>) will produce an error in CMake 4.0 and above.

That only got me so far though as the CMakeList files created in `build/gtest-src` and `build/gtest-src/googlemock` use `cmake_minimum_required(2.8.12)` and so produce the same error:

<img width="816" height="341" alt="Image" src="https://github.com/user-attachments/assets/d93b3968-95f2-4b8a-8224-07b71e23b611" />

[...]

PR #6836, opened 2026-04-28
Add Fedora 44 build and drop Fedora 42 build

SlightlyCircuitous:
## Short roundup of the initial problem

Fedora 44 released today (and the Docker image was updated last night) and will be supported until June 6, 2027. Fedora 42 will reach EOL on May 13, 2026. 

## What will change with this Pull Request?
- Removes Fedora 42 from Github build workflow
- Adds Fedora 44 to Github build workflow
- Updates Release Template to reflect above changes

PR #6843, opened 2026-05-04
Fix Arch and Fedora 44 builds: suppress -Werror=sfinae-incomplete for GCC 16+

DawnFire42:
Qt MOC and protobuf forward declarations trigger -Werror=sfinae-incomplete in GCC 16.
  Re-enable Fedora 44 CI build now that it compiles successfully.

## Related PR
- PR #6836 introduced a skip because builds were failing. 

## Related Issue
- #6841 

- My other PRs were failing because of the same issue and this needs a separate PR.

- suppress -Werror=sfinae-incomplete for GCC 16+
- Re-enable Fedora 44 CI build now that it compiles successfully.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -138,7 +138,6 @@ jobs:
           - distro: Fedora
             version: 44
             package: RPM
-            test: skip   # This is failing to build, currently. To be fixed.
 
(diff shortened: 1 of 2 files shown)


tooomm:
This fixes the failing arch build as well.

ebbit1q (approved):
thanks!

PR #6844, opened 2026-05-05
[SettingsManager] Properly handle multithreaded access again

RickyRister:
## Related Ticket(s)
- Fixes issue introduced in #6837
- Unreverts #6747

#6837 reverted the fix made in #6747 for performance reasons. However, that means the crash from before started happening again.

see #6747

### DIFF
--- a/libcockatrice_settings/libcockatrice/settings/settings_manager.cpp
+++ b/libcockatrice_settings/libcockatrice/settings/settings_manager.cpp
@@ -159,6 +159,15 @@ QVariant SettingsManager::getValue(const QString &name, const QString &group, co
     return value;
 }
 
+void SettingsManager::batchWrite(std::function<void(QSettings &)> batchWriteFunction)
+{
+    auto settings = getSettings();
+    settings.setAtomicSyncRequired(false);
+    batchWriteFunction(settings);
+    settings.sync(); // single flush
+    settings.setAtomicSyncRequired(true);
+}
+
 /**
  * Calls sync on the underlying QSettings object
  */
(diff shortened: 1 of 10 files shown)


libcockatrice_settings/libcockatrice/settings/settings_manager.cpp
```diff
@@ -4,13 +4,19 @@ SettingsManager::SettingsManager(const QString &_settingPath,
                                  const QString &_defaultGroup,
                                  const QString &_defaultSubGroup,
                                  QObject *parent)
-    : QObject(parent), defaultGroup(_defaultGroup), defaultSubGroup(_defaultSubGroup),
-      settings(_settingPath, QSettings::IniFormat)
+    : QObject(parent), settingPath(_settingPath), defaultGroup(_defaultGroup), defaultSubGroup(_defaultSubGroup)
 {
 }
 
+QSettings SettingsManager::getSettings() const
```
ebbit1q, line 11: can we add a comment here on the multithreaded issue so we don't try to optimise this again?

BruebachL (approved):

libcockatrice_settings/libcockatrice/settings/settings_manager.cpp
```diff
@@ -142,10 +159,21 @@ QVariant SettingsManager::getValue(const QString &name, const QString &group, co
     return value;
 }
 
+void SettingsManager::batchWrite(std::function<void(QSettings &)> batchWriteFunction)
+{
+    auto settings = getSettings();
+    settings.setAtomicSyncRequired(false);
```
ebbit1q, line 165: @BruebachL is this required? does it not introduce a chance for file corruption?

BruebachL: It's an incredible performance boost to the set manager dialog, bringing the required I/O down from 2600~ to 1. I'd rather introduce the very remote possibility of a corruption happening than to subject user disks to 2600 read/writes on every set manager close.

BruebachL: I don't really see much of a possibility for significant file corruption, btw.

PR #6846, opened 2026-05-07
Use Qt 6.11.0

BruebachL:
Upgrading Qt to 6.11.* in #6794 introduced some issues due to the recency of the Qt release, related to the aqtinstall tool not being able to verify the checksum. The proper fix would be to adjust the aqtinstall mirror sources or whatever to make the official mirrors more trustworthy or something but honestly, no one can be bothered to figure this out, we don't gain much from a .1 minor version bump and we wanna release.

- Roll back runners to Qt 6.10.* instead of Qt 6.11.*

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -281,7 +281,7 @@ jobs:
             override_target: 13
             make_package: 1
             package_suffix: "-macOS13_Intel"
-            qt_version: 6.11.*
+            qt_version: 6.10.*
             qt_arch: clang_64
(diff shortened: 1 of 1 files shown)


ebbit1q:
I suppose if we're doing this we're fully reverting #6794 might as well remove the workaround on that

ebbit1q:
alternatively: we use 6.11.0 which does seem to work unlike 6.11.1 rn

however, it's possible this is a weird temporary error with the new 6.11.1 patch? maybe it just goes away? maybe it requires more changes in aqtinstall?

ZeldaZach (approved):

PR #6847, opened 2026-05-07
pin the aqtinstall version

ebbit1q:
pins the aqtinstall version to the version it was at in #6794 which worked.


ebbit1q:
closing in favor of #6846

PR #6854, opened 2026-05-08
Fix toc link

tooomm:
## Related Ticket(s)
- Oversight from #6848


ebbit1q (approved):
thanks, we might want to avoid putting these badges in headers to avoid these cases

PR #6859, opened 2026-05-09
[Game] Refactor: add selectedCards method to GameScene

RickyRister:

There are many places in the code where we want to get all selected cards from the scene. Currently, every time we have to manually get all selected items from the scene and cast them to `CardItem`, possibly with some sort of error checking.

It would be more convenient if we had a single method to do that.

- Create a `GameScene::selectedCards` method that returns all selected cards as a list
  - Does checks after each cast
- Update code to use the new method

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -1173,14 +1173,9 @@ void PlayerActions::actMoveCardXCardsFromTop()
 
     defaultNumberTopCardsToPlaceBelow = number;
 
-    QList<QGraphicsItem *> sel = player->getGameScene()->selectedItems();
-    if (sel.isEmpty()) {
+    QList<CardItem *> cardList = player->getGameScene()->selectedCards();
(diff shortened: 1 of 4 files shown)


cockatrice/src/game/player/player_actions.cpp
```diff
@@ -1687,9 +1668,9 @@ void PlayerActions::actReveal(QAction *action)
         cmd.set_player_id(otherPlayerId);
     }
 
-    QList<QGraphicsItem *> sel = player->getGameScene()->selectedItems();
-    while (!sel.isEmpty()) {
-        const auto *card = qgraphicsitem_cast<CardItem *>(sel.takeFirst());
+    QList<CardItem *> cards = player->getGameScene()->selectedCards();
+    while (!cards.isEmpty()) {
```
ebbit1q: this can be a for loop, no idea why it bothers to use takeFirst

ebbit1q (approved):
looks good

PR #6869, opened 2026-05-10
[Game] Implement action to automatically take damage from creatures

### LINKED ISSUE #4084: Add action to automatically take damage from creatures
**Similar Requests**
None

Add an action to select multiple cards on the field and subtract their power (first value of p/t field) from your own life total. Inversely an action to add this amount to your life total could be added. (either as lifelink or to undo)

**Additional Context**
I found this being suggested by user Mibbles#1044 in our Discord server's dev channel at 2019-11-20 04:08 utc https://discordapp.com/channels/314987288398659595/401594186983407616/646577586876448769

RickyRister:
## Related Ticket(s)
- Fixes #4084

It would be convenient to have an action that let you automatically take damage from creatures.

- Add a `getLifeCounter` method to `Player`
- Added the `Reduce life by power` action to the card menu for cards on the table
  - Reduces the user's life by the combined power of all selected cards
    - Uses `CardItem::parsePT` to parse the power of each card; defaults to 0 if a value is not found
    - Also uses 0 if power is negative
- Update shortcut settings for new action
  - Has default shortcut `Ctrl+Shift+L` to match with set life having a default shortcut of `Ctrl+L`

https://github.com/user-attachments/assets/859ea9fc-6d5a-412d-a8e6-0c62fa203d62

## Screenshots

Screenshot 2026-05-09 at 5 30 35 PM

Screenshot 2026-05-09 at 5 20 36 PM


ZeldaZach (approved):
I'm okay with this, but I worry the right click menu is getting very "bulky" -- Maybe we can have a way for users to add/hide these elements dynamically in settings?

cockatrice/src/game/player/player.h
```diff
@@ -199,6 +199,11 @@ public slots:
         return counters;
     }
 
+    /**
+     * Gets the counter that represents the life total. Shouldn't return null
```
ebbit1q: this comment is deceptive, in case we'd have future server changes that change this it would return null

ebbit1q:
our card action menu is already incredibly bulky, way too bulky, but I don't think hiding options from it is a good idea, further submenus are also kind of awkward but maybe a bit more straightforward, alternatively trying to define like "zones" on card objects which have different actions is an option but really really awful for us as we have the board scale to so small sometimes, holding a kind of "secret" modifier key for more options might be preferred by power users but in my opinion those users should remember the customisable keybinds for those actions instead, as the menu is for inexperienced users I think the best way forwards is further submenus as this caters the most to this kind of user, we must try to limit the amount of options in one menu to maybe 8 options to stop them from being overwhelmed, we should also not have any submenus descend further than one layer deeper and in rare exceptions two layers in my opinion

RickyRister:
Yeah, we'll have to figure out what to do about the action menus later. Feel free to create an issue for that.

PR #6871, opened 2026-05-10
CI: Update trigger for doc builds

tooomm:
Docs are build for created tags already

- Docs are only build for actually published releases and pre-releases


ebbit1q (approved):
thanks

Issue #6863, opened 2026-05-09
[VDS] Support colorless decks


jroed:
Client Version: 3.0.0 (2026-05-08)
Client Operating System: Windows 11 Version 25H2
Build Architecture: 64-bit
Qt Version: 6.11.0

**Description of New Feature**

In VDS color identity of decks is represented by displaying the respective mana symbols below each entry in the list (if the option is enabled). But this isn't possible for decks that have only cards without a color identity. The current behavior is that no mana symbol is shown.

The proposal is to show a colorless mana symbol below colorless decks and also add a filtering option.

**Additional Context**

Since currently there seems to be no colorless mana symbol icon available such an icon needs to be added (/cockatrice/resources/icons/mana). Compare https://github.com/Cockatrice/Cockatrice/pull/5685.

This icon could also be useful in other features such as the deck analytics:

<img width="868" height="174" alt="Image" src="https://github.com/user-attachments/assets/5bd21654-b770-4055-9633-7d1932b48007" />

PR #6873, opened 2026-05-11
Add generic colorless mana symbol

SlightlyCircuitous:
## Related Ticket(s)
- Progress towards https://github.com/Cockatrice/Cockatrice/issues/6863

We do not have a generic mana symbol for colorless mana and would like to use one in the client

- Adds a colorless mana symbol that I made in Inkscape (source is I drew a diamond)

## Screenshots
a grey circle with a black diamond in it

### DIFF
--- a/cockatrice/cockatrice.qrc
+++ b/cockatrice/cockatrice.qrc
@@ -55,6 +55,7 @@
         <file>resources/icons/view.svg</file>
 
         <file>resources/icons/mana/B.svg</file>
+        <file>resources/icons/mana/C.svg</file>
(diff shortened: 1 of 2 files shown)


ebbit1q:
nice work, in what places would we use this icon?

SlightlyCircuitous:
As noted in the linked issue, it could be used to denote colorless decks in the VDS (instead of having no symbol) and could also be placed over the existing colorless mana sources bar in the VDE deck analytics tab, to match the other five symbols over their respective bars.

ebbit1q (approved):
thanks!

PR #6876, opened 2026-05-11
[VisualDeckEditor] Highlight searchEdit after add card

### LINKED ISSUE #6867: card title not selected after adding to deck
select a card and press enter to add to deck. before, the card name would be selected so you can type the next card. now, you have to select and delete the text or click the x.

RickyRister:
## Related Ticket(s)
- Fixes #6867

When using the VDE's card database in table mode, the searchEdit doesn't automatically get highlighted after each add card, whereas it does in the classic deck editor

- Add a `highlightAllSearchEdit` method to `VisualDatabaseDisplayWidget` and connect it to the underlying `DeckEditorDatabaseDisplayWidget`'s card added signals

Conveniently, that means the highlight on add will only happen if you're in the table display mode.

https://github.com/user-attachments/assets/22264ce3-7670-410e-b7f1-35a39c11f72b


ZeldaZach (approved):
Big fan of the small PRs that are quick and easy to review!

PR #6879, opened 2026-05-12
Centralize counter API with server-side bounds and no-op filtering

DawnFire42:
Refactor server counter API to own overflow protection and filter no-op events

  Counter modifications now clamp to int bounds server-side and return change
  status, allowing command handlers to skip network broadcasts when values
  don't actually change.

- Since my command zone build uses counters, this refactor opportunity became relevant.

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
@@ -49,6 +49,7 @@
 #include <libcockatrice/rng/rng_abstract.h>
 #include <libcockatrice/utility/trice_limits.h>
 #include <libcockatrice/utility/zone_names.h>
+#include <limits>
 #include <ranges>
 
 Server_AbstractPlayer::Server_AbstractPlayer(Server_Game *_game,
@@ -1091,8 +1092,9 @@ Server_AbstractPlayer::cmdCreateToken(const Command_CreateToken &cmd, ResponseCo
                 _event.set_zone_name(card->getZone()->getName().toStdString());
                 _event.set_card_id(card->getId());
 
-                card->setCounter(i.key(), i.value(), &_event);
-                ges.enqueueGameEvent(_event, playerId);
+                if (card->setCounter(i.key(), i.value(), &_event)) {
+                    ges.enqueueGameEvent(_event, playerId);
+                }
             }
 
             // Copy parent card
@@ -1338,8 +1340,9 @@ Response::ResponseCode Server_AbstractPlayer::cmdSetCardCounter(const Command_Se
     Event_SetCardCounter event;
     event.set_zone_name(zone->getName().toStdString());
     event.set_card_id(card->getId());
-    card->setCounter(cmd.counter_id(), cmd.counter_value(), &event);
-    ges.enqueueGameEvent(event, playerId);
+    if (card->setCounter(cmd.counter_id(), cmd.counter_value(), &event)) {
+        ges.enqueueGameEvent(event, playerId);
+    }
 
     return Response::RespOk;
 }
@@ -1368,14 +1371,13 @@ Response::ResponseCode Server_AbstractPlayer::cmdIncCardCounter(const Command_In
         return Response::RespNameNotFound;
     }
 
-    int newValue = card->getCounter(cmd.counter_id()) + cmd.counter_delta();
-    card->setCounter(cmd.counter_id(), newValue);
-
     Event_SetCardCounter event;
     event.set_zone_name(zone->getName().toStdString());
     event.set_card_id(card->getId());
-    event.set_counter_id(cmd.counter_id());
-    event.set_counter_value(newValue);
+    if (!card->incrementCounter(cmd.counter_id(), cmd.counter_delta(), &event)) {
+        return Response::RespOk;
+    }
+
     ges.enqueueGameEvent(event, playerId);
 
     return Response::RespOk;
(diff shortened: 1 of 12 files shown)


libcockatrice_network/libcockatrice/network/server/remote/game/server_card.h
```diff
@@ -153,7 +153,20 @@ class Server_Card : public Server_ArrowTarget
     {
         cardRef = _cardRef;
     }
-    void setCounter(int _id, int value, Event_SetCardCounter *event = nullptr);
+    /// @brief Sets a card counter to an exact value.
+    /// @param _id The counter ID.
+    /// @param value The new value (0 removes the counter).
+    /// @param event Optional event to populate with counter state.
+    /// @return true if the value changed, false otherwise.
```
RickyRister: nit: we prefer using `/** */` over `///` for full-length docs

libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
```diff
@@ -1367,14 +1370,12 @@ Response::ResponseCode Server_AbstractPlayer::cmdIncCardCounter(const Command_In
...
 
-    int newValue = card->getCounter(cmd.counter_id()) + cmd.counter_delta();
-    card->setCounter(cmd.counter_id(), newValue);
-
     Event_SetCardCounter event;
     event.set_zone_name(zone->getName().toStdString());
     event.set_card_id(card->getId());
-    event.set_counter_id(cmd.counter_id());
-    event.set_counter_value(newValue);
+    if (!card->incrementCounter(cmd.counter_id(), cmd.counter_delta(), &event))
```
RickyRister: try to always use braces in new code

libcockatrice_network/libcockatrice/network/server/remote/game/server_counter.h
```diff
@@ -59,11 +72,35 @@ class Server_Counter
...
         count = _count;
+        return count != oldCount;
+    }
+
+    /// @brief Increments the counter by delta with overflow-safe arithmetic.
+    /// @param delta The amount to add (may be negative for decrement).
+    /// @return true if the value changed, false otherwise.
+    /// @note Clamps result to [INT_MIN, INT_MAX] to prevent overflow.
+    [[nodiscard]] bool incrementCount(int delta)
+    {
```
RickyRister: is there a reason this is implemented in the .h file rather than the cpp file?

libcockatrice_network/libcockatrice/network/server/remote/game/server_counter.h
```diff
@@ -59,11 +72,35 @@ class Server_Counter
...
+        // TODO: Extract overflow-safe arithmetic into shared helper.
+        // Duplicated in Server_Card::incrementCounter() - keep in sync if modified.
+        const int oldCount = count;
+        const auto result = static_cast<int64_t>(count) + static_cast<int64_t>(delta);
+        count = static_cast<int>(qBound(static_cast<int64_t>(std::numeric_limits<int>::min()), result,
+                                        static_cast<int64_t>(std::numeric_limits<int>::max())));
+        return count != oldCount;
     }
 
+    /// @brief Populates \a info with this counter's current state for network serialization.
```
RickyRister: is the `\a` intentional?

RickyRister:
Huh, didn't realize the life counter is affected by integer overflows

https://github.com/user-attachments/assets/da769f08-bd1a-4f5d-90ba-70fbb8f36149

And card counters are affected by underflows

https://github.com/user-attachments/assets/1b6331a7-dc65-4740-b3e5-175fb8b582a6

Seems like the life overflow is fixed. Looks good

https://github.com/user-attachments/assets/302dc977-e5e7-43fd-972a-470c6389f14e

RickyRister:
Seems like it's still possible to set the card counter to a negative number. At which point the add and remove counter actions no longer do anything.

https://github.com/user-attachments/assets/8d23851d-516c-40fa-8349-8cb1532359e9

Is this intentional?

libcockatrice_network/libcockatrice/network/server/remote/game/server_counter.h
DawnFire42: I try to follow existing patterns when possible, for the setter pattern — setCount() on line 83-88 is also inline, so incrementCount() mirrors that style. 
Also, when defined in the header, the compiler can inline the call at every call site to avoid function call overhead for a hot path (counter updates in games).

Though I see that it's better in the .cpp for encapsulation. It would work either way.
Would you prefer I move it to .cpp?

libcockatrice_network/libcockatrice/network/server/remote/game/server_counter.h
DawnFire42: \a tells Doxygen that "info" is a parameter name. But yes, no ///... I'll make sure I remember

DawnFire42:
> Seems like it's still possible to set the card counter to a negative number. At which point the add and remove counter actions no longer do anything.
> 
>  example.mp4 
> Is this intentional?

I may have made a mistake by removing clamping. I thought I had introduced it through my other project when extracting this refactor. I'll look into this more closely

libcockatrice_network/libcockatrice/network/server/remote/game/server_counter.h
RickyRister: My philosophy is to put anything that requires any amount of logic in the cpp file. There isn’t a project-wide consensus, though, so if you think there’s a good reason to leave it in the header file, then I’m fine either ways.

RickyRister:
Rename the PR. It's not just a refactor anymore when you're actually changing the behavior.
Looks good otherwise.

RickyRister (approved):

PR #6881, opened 2026-05-12
Revert UI.

BruebachL:
Change is bad.

## Screenshots
image


ZeldaZach (changes requested):
I'm confused what this PR is doing

ebbit1q:
change is good, without change no progress would be made

if people are upset about progress, it means more changes are needed

PR #6884, opened 2026-05-12
fix resizing the visual card database loading more pages forever

ebbit1q:
## Related Ticket(s)
- introduced in #5834

when dragging the corner of the app while the vde is on the visual card database, the app will load a new page of cards every time the resize is processed, quickly grinding interface updates to a halt until you stop resizing it

fetching a new page of cards only happens when using the scroll wheel, not when dragging the bar or using pagedown or down arrow

- a new page will only be loaded when the scrollbar is near the end (2 pages currently)
- all methods of scrolling will fetch new pages
- I checked, and when there are very few cards in the db it will work as expected (the scrollbar is always near the end)

### DIFF
--- a/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.h
+++ b/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.h
@@ -85,7 +85,6 @@ protected slots:
     void onHover(const ExactCard &hoveredCard);
     void addCard(const ExactCard &cardToAdd);
     void databaseDataChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight);
-    void wheelEvent(QWheelEvent *event) override;
     void modelDirty() const;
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.h
```diff
@@ -123,6 +122,7 @@ protected slots:
     int cardsPerPage = 100; // Number of cards per page
 
     void highlightAllSearchEdit();
+    bool nearEndOfPage();
```
RickyRister: nit: function can be const

RickyRister (approved):
LGTM

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.h
ebbit1q: true, oversight on my part

PR #6775, opened 2026-04-05
Define Cockatrice as an editor/handler for .cod files and cockatrice:// protocol on all platforms

BruebachL:
## Related Ticket(s)
- Fixes #5438

## Short roundup of the initial problem
Cockatrice uses our own .cod format but we don't tell operating systems that we are an editor for .cod files in any way shape or form. For my musings on the things needed for each operating system, please see [this post](https://github.com/Cockatrice/Cockatrice/issues/5438#issuecomment-2583814549).

## What will change with this Pull Request?
- Set Cockatrice as an editor for .cod and cockatrice:// in Mac PList, Windows NSIS scripts (registry keys) and Linux .desktop and MIME file
- Enforce a single-instance principle if Cockatrice is opened with a launch argument
- If launched without an argument, a secondary instance opens.

## Why?
- It's a serious quality of life and also enables drag-and-drop handling in the future (i.e. drop a copy and pasted text into Cockatrice to import a deck)
- We need this if we ever want to implement invitations/click-to-join links/discord lobbies
- We need this if we ever want to implement remote deck sharing

Issue #5438, opened 2025-01-10
Opening a .COD file uses Deck Editor


Axiflare:
Unsure

I think it would be neat to be able to open a .COD file and the Deck Editor would load it for quick modifications without using notepad, extra clicks, etc.

I'm a total idiot at programming so I don't know how hard it would be to implement, I just think a bit of extra utility on the file itself would be neat.

BruebachL:
It's not an impossible task. For determining which program to use when opening an extension, the operating systems use different systems.

Windows:

```
HKEY_CLASSES_ROOT\.cod:

    Default: Cockatrice

HKEY_CLASSES_ROOT\Cockatrice\shell\open\command:

    Default: "<path_to_cockatrice>" "%1"
```
    
MacOS:

Determined by an application's Info.plist

```
<key>CFBundleDocumentTypes</key>
<array>
    <dict>
        <key>CFBundleTypeExtensions</key>
        <array>
            <string>cod</string>
        </array>
        <key>CFBundleTypeName</key>
        <string>Cockatrice</string>
        <key>CFBundleTypeRole</key>
        <string>Editor</string>
        <key>LSHandlerRank</key>
        <string>Default</string>
    </dict>
</array>
```

Linux:

We need to define a mime type for .cod

/usr/share/mime/packages/cockatrice.xml

```
<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
    <mime-type type="application/x-cockatrice">
        <comment>Cockatrice</comment>
        <glob pattern="*.cod"/>
    </mime-type>
</mime-info>
```

Update mime on install:
`sudo update-mime-database /usr/share/mime`

Finally, a desktop file for Cockatrice (I'm pretty sure we already have this for most distros but I don't really deal with distribution)

/usr/share/applications/cockatrice.desktop

```
[Desktop Entry]
Name=Cockatrice
Exec=/path/to/cockatrice %f
MimeType=application/x-cockatrice;
Type=Application
```

-----

Now, this handles getting the operating system to open .cod files with the application and passing the filename as the first argument. Cockatrice itself would then need to detect this as a deck to load and load the deck on launch, which is another issue.

Axiflare:
Whoa, I totally forgot I asked about this feature, and now it's implemented! Excellent job!

BruebachL:
It's been one of my white wales for a while now haha

PR #6886, opened 2026-05-13
feat: register cockatrice:// and cockatrice-oracle:// protocol handlers

### LINKED ISSUE #1495: Register cockatrice:// protocol handler on install.

seavor:
## Related Ticket(s)
- Closes [Register cockatrice:// protocol handler on install #1495](https://github.com/Cockatrice/Cockatrice/issues/1495)
- Builds on the unmerged work in [#6775](https://github.com/Cockatrice/Cockatrice/pull/6775) by [@BruebachL](https://github.com/BruebachL) — credit for the prior art on `SingleInstanceManager`, `.desktop` MIME types, NSIS registry keys, and the split plist approach.

Clicking a `cockatrice://` link (from a website or Discord) had no effect — the OS didn't know Cockatrice handled that scheme, and even if it did, Cockatrice had no code to act on the URL. Same for `cockatrice-oracle://` for triggering card-data updates.

### Platform registration
- **Linux** — `.desktop` files gain `MimeType=x-scheme-handler/cockatrice[-oracle]` and `Exec=… %u`.
- **Windows** — NSIS installer writes `HKCR\cockatrice` / `HKCR\cockatrice-oracle` registry keys; uninstaller removes them.
- **macOS** — split `Info.cockatrice.plist` / `Info.oracle.plist` with `CFBundleURLTypes` entries. The `UrlSchemeEventFilter` is installed on `QApplication` *before* the single-instance handshake's nested event loop so cold-start URLs (queued before `main()` returns) aren't lost.

### URL spec
- `cockatrice://joingame?hostname=H&port=P&roomid=R&gameid=G[&spectate=1]`
- `cockatrice-oracle://update[?spoilers=1]`
- Scheme matched case-insensitively (RFC 3986). Validates port range and rejects negative IDs.
- **Credentials are intentionally NOT accepted via URL.** URLs leak via shell history, EDR command-line capture, browser history, and local-socket forwarding. When the target server requires authentication, the client falls back to saved credentials, then to the existing Connect dialog pre-filled with the URL's hostname/port.

### Single-instance forwarding
- New `SingleInstanceManager` uses `QLocalSocket` / `QLocalServer` to forward a URL from a secondary launch to the already-running primary, then exit.
- Enforced **only** when a URL is being delivered; without a URL, multiple Cockatrice processes can coexist (preserves local two-client testing).
- Per-user socket name salting (`$USER` / `$USERNAME`) so other accounts on the machine can't squat the socket and intercept URLs.
- The handshake is fully event-driven (`connected` / `readyRead` / `errorOccurred` + a single `QTimer` deadline); no synchronous `waitFor*` calls, so there's no read-buffer race on any platform. Both URL and ACK are framed with `QDataStream` `startTransaction`/`commitTransaction` so partial reads roll back cleanly. The resolved flag is `shared_ptr<bool>` so it survives every lambda capture.

### Intent chain
A new `Intent` base class chains the URL into focused async steps: `IntentConnectToServer` → `IntentLogin` → `IntentJoinServerRoom` → `IntentJoinServerGame`.

- `Intent` self-deletes on `finished()` from any path (`doExecute()` or external `abort()`) — the connection is wired in the constructor, so it's structurally impossible to leak an intent.
- `finished()` is gated by an internal `m_finished` flag — fires **at most once** even when multiple completion paths race (success signal + cleanup disconnect, late response + timeout). `execute()` is similarly idempotent.
- `IntentConnectToServer` resolves credentials: saved creds from `ServersSettings` → fall back to `DlgConnect` pre-filled with hostname/port.
- Every networked intent has a 30s timeout safety net (configurable via virtual `Intent::timeoutMs()`) and a `ConnectionController::statusChanged` watchdog that aborts the chain on manual disconnect. Both are lifted into the base class (`startTimeoutSafetyNet()`) and an `abortOnDisconnect()` free helper, so concrete intents are ~10 lines each.
- `IntentLogin` quick-fails when the upstream connect step already failed (status is `StatusDisconnected` at subscribe time).
- `IntentJoinServerGame` mirrors `GameSelector` parity: supports spectator mode (`&spectate=1`), respects admin-lock for `override_restrictions`, short-circuits when the user already has a tab for the target game.
- `TabServer::joinRoom` emits a new `roomAlreadyJoined` signal on the short-circuit "already in this room" path; `TabSupervisor` forwards it to `roomJoinedById` so `IntentJoinServerRoom` resolves uniformly whether the user is already there or not.
- Robust cleanup if the user closes Cockatrice mid-flow: every intent is parented to MainWindow and tears down cleanly with it (covered by a regression test).

### Oracle URL handling
- `cockatrice-oracle://update[?spoilers=1]` triggers a background card-data download.
- Shared `UrlUtils::parseOracleUrl` helper used by both the argv path and the macOS `QFileOpenEvent` path.

### Library structure
- `UrlUtils`, `Intent`, `SingleInstanceManager` live in `libcockatrice_utility` (Core+Network only, so they don't drag Qt::Gui into `servatrice`).
- `UrlSchemeEventFilter` uses `QFileOpenEvent` (Qt::Gui), so it lives in a **new sibling library `libcockatrice_utility_gui`** linked only by `cockatrice`/`oracle`/tests. Keeps `servatrice`'s dependency footprint clean.

### Local format tooling
- `Dockerfile.format` + `docker-compose.format.yml` build an `ubuntu:24.04` image with `clang-format`, `cmake-format`, and `shellcheck` — mirrors the CI desktop-lint job exactly.
- `Makefile` shims: `make format` (diff-only check, matches CI) and `make format-fix` (apply changes).

### Major decisions
- **No credentials in URL** — security over convenience; saved creds + DlgConnect prompt cover the auth case without leaking secrets.
- **Single-instance enforcement is URL-gated** — without this, the original prior-art would have silently broken local multi-client testing workflows.
- **`Intent` self-deletes from its constructor**, not from `execute()` — guarantees cleanup whether the chain finishes normally or aborts.
- **`emitFinished()` gate over raw `emit`** — concrete intents have several completion paths that could fire concurrently (signal + cleanup, response + timeout). A flag on the base class makes the at-most-once contract structural rather than a per-subclass discipline.
- **Async, signal-driven handshake** for single-instance forwarding — `waitForConnected/waitForBytesWritten/waitForReadyRead` were a `readyRead`-race surface on Windows; replaced with a state machine driven entirely by Qt signals.
- **Header-only URL parsers** in `libcockatrice/utility/url_utils.h` — gives us unit-testable parsing without dragging GUI/network deps into tests.
- **`libcockatrice_utility_gui` as a sibling lib** rather than dragging Qt::Gui into `libcockatrice_utility` — preserves servatrice's headless dependency boundary.

### Tests
36 GoogleTest cases across four single-purpose targets under `tests/`:

- `url_utils_test` (22 tests) — `findUrlArgument`, `parseJoinGameUrl` (happy path + 11 rejection paths + spectate flag + credential-strip regression), `parseOracleUrl`.
- `url_scheme_event_filter_test` (3 tests) — `QFileOpenEvent` matching, non-matching, case-insensitive.
- `intent_test` (7 tests) — self-delete on finished, in-flight protection, abort, chain propagation, parent destruction mid-flight, `emitFinished()` once-only regression, `execute()` idempotence regression.
- `single_instance_manager_test` (4 tests) — per-user naming, resolve as primary, forward URL to existing primary, `roleResolved` single-emission regression.

## Screenshots
N/A — no UI changes other than the existing `DlgConnect` being shown pre-filled when the URL targets a server with no saved credentials.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ compile_commands.json
 .gdb_history
 cockatrice/resources/config/qtlogging.ini
 docs/
+.claude/settings.local.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,7 +174,7 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
       -Wno-error=delete-non-virtual-dtor
       -Wno-error=sign-compare
       -Wno-error=missing-declarations
-      -Wno-error=sfinae-incomplete  # GCC 16+: Qt MOC + protobuf forward decls trigger this
+      -Wno-error=sfinae-incomplete # GCC 16+: Qt MOC + protobuf forward decls trigger this
   )
 
   foreach(FLAG ${ADDITIONAL_DEBUG_FLAGS})
@@ -345,6 +345,7 @@ if(WITH_ORACLE OR WITH_CLIENT)
   add_subdirectory(${CMAKE_SOURCE_DIR}/libcockatrice_settings ${CMAKE_BINARY_DIR}/libcockatrice_settings)
   add_subdirectory(${CMAKE_SOURCE_DIR}/libcockatrice_models ${CMAKE_BINARY_DIR}/libcockatrice_models)
   add_subdirectory(${CMAKE_SOURCE_DIR}/libcockatrice_filters ${CMAKE_BINARY_DIR}/libcockatrice_filters)
+  add_subdirectory(${CMAKE_SOURCE_DIR}/libcockatrice_utility_gui ${CMAKE_BINARY_DIR}/libcockatrice_utility_gui)
 endif()
 
 if(WITH_SERVER)
--- a/Dockerfile.format
+++ b/Dockerfile.format
@@ -1,0 +1,33 @@
+# Cockatrice formatter image — mirrors the apt installs in
+# .github/workflows/desktop-lint.yml so a local `format.sh` run matches CI.
+#
+# Build:   docker build -t cockatrice-format -f Dockerfile.format .
+# Check:   docker run --rm -v "$(pwd):/src" cockatrice-format
+# Apply:   docker run --rm -v "$(pwd):/src" cockatrice-format \
+#             ./format.sh --cmake --shell --branch origin/master
+#
+# Or via docker compose (recommended):
+#   docker compose -f docker-compose.format.yml run --rm format          # diff-only check
+#   docker compose -f docker-compose.format.yml run --rm format-apply    # apply changes
+
+FROM ubuntu:24.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        clang-format \
+        cmake-format \
+        git \
+        shellcheck \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Allow git to operate on the bind-mounted repo despite differing owner UID.
+RUN git config --system --add safe.directory /src
+
+WORKDIR /src
+
+# Default to the same invocation CI uses (.ci/lint_cpp.sh).
+CMD ["./format.sh", "--diff", "--cmake", "--shell", "--print-version", "--branch", "origin/master"]
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+# Convenience shims for running the project's code-style check via Docker.
+# See Dockerfile.format and docker-compose.format.yml for the underlying setup.
+
+.PHONY: format format-fix
+
+# Diff-only check (matches CI; exit 2 means changes are needed).
+format:
+	docker compose -f docker-compose.format.yml run --rm format
+
+# Apply formatting changes in place.
+format-fix:
+	docker compose -f docker-compose.format.yml run --rm format-apply
--- a/cmake/FindQtRuntime.cmake
+++ b/cmake/FindQtRuntime.cmake
@@ -115,4 +115,10 @@ string(REGEX REPLACE "([^;]+)" "${COCKATRICE_QT_VERSION_NAME}::\\1" TEST_QT_MODU
 # Core-only export (useful for headless libs)
 set(QT_CORE_MODULE "${COCKATRICE_QT_VERSION_NAME}::Core")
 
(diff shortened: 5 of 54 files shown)


tooomm:
Bruebach's PR listed https://github.com/Cockatrice/Cockatrice/issues/5438 as fixed by his PR.

Without screening the hundreds of code changes, which link types will be supported with this?

seavor:
> Bruebach's PR listed #5438 as fixed by his PR.
> 
> Without screening the hundreds of code changes, which link types will be supported with this?

His PR was closed without being merged. His PR addressed #5438 and #1495. This PR only address #1495.

tooomm:
> > Bruebach's PR listed #5438 as fixed by his PR.
> > Without screening the hundreds of code changes, which link types will be supported with this?
> 
> His PR was closed without being merged. His PR addressed #5438 and #1495. This PR only address #1495.

Thanks for clarifying that this tackles only parts of the other PR. 👍 
The feature sounds super helpful for players and game hosts (potentially server admins) alike!

I do understand now, that there are two features covered:
- Joining an online game:
  `cockatrice://joingame?hostname=H&port=P&roomid=R&gameid=G[&spectate=1]`
  - Are all parameters required?
      - Or can a server admin use it to allow users to easily join their server (without joining a game).
        Go with `join` only and allow that as well? No game info = join server, otherwise join game on that server.
      - If I'm already logged in to that server, it'll still work and I can join the game?

  - Does this also work for local games on the same network if giving the IP as hostname?

- Updating the card database:
  `cockatrice-oracle://update[?spoilers=1]`
  - If "spoilers=1" is appended, will it then only check for spoilers?
    Or will it always check for main card updates, and spoilers on top if added?
  - What about token?

<br>

The changes you listed under `Local format tooling` seem to be unrelated to the main feature you're proposing in this PR if I'm not mistaken?
Splitting that into its own PR would be great! 🙇‍♂️

BruebachL:
Closed in favor of #6775

PR #6887, opened 2026-05-13
style: Add braces to all control flow statements

DawnFire42:
Standardize code style by adding explicit braces to all single-statement
  control flow blocks (if, else, for, while) across the entire codebase.

  Also documents the InsertBraces clang-format option (requires v15+)

  I failed to add braces on every control flow statements in my other PRs. Turns out there is a clang-format 15 option for this so I ran it on the entire code base and left it as an option in the formatting file.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -22,6 +22,8 @@ AllowShortFunctionsOnASingleLine: None
 BinPackParameters: false
 AllowAllParametersOfDeclarationOnNextLine: false
 IndentCaseLabels: true
+# requires clang-format 15
+# InsertBraces: true
 PointerAlignment: Right
 SortIncludes: true
 IncludeBlocks: Regroup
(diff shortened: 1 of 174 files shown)


.clang-format
```diff
@@ -22,6 +22,8 @@ AllowShortFunctionsOnASingleLine: None
 BinPackParameters: false
 AllowAllParametersOfDeclarationOnNextLine: false
 IndentCaseLabels: true
+# requires clang-format 15
+# InsertBraces: true
```
tooomm: We use clang-format v18 for the lint job in CI, I guess we can safely enable the check already?

ZeldaZach: Agreed

ebbit1q:
personally I'm pretty relaxed about having to always add braces to these, they're useful in the case that you want to easily add something inside the statement, there is a bunch of cases where you'd never really do that and in those cases I don't mind them missing

for consistency though, and if we can simply automate it, I don't mind simply adding the braces always, it's a simple solution and it's never bad to have the extra braces in my opinion

ebbit1q (approved):

PR #6889, opened 2026-05-13
[Server][Game] Make undo draw failure visible in chat

### LINKED ISSUE #6883: Failed to Undo Draw Chat Message
Could we get a chat message noting a failed undo draw command instead of it failing silently.

**Additional Context**
In a few games where I've been a Judge, we've had a player draw a card then discard it, and then need to rewind that action. Currently there is no way to verify the difference between the undo draw not working because the last drawn card has left their hand, vs the player not actually clicking the button, and so we need to use a harsher fix. Even just a simple "[Player] is unable to undo their last draw." in chat would be enough to avoid a lot of these cases.

RickyRister:
## Related Ticket(s)
- Fixes #6883

Judges have no way to verify if an "undo draw" failed.

https://github.com/user-attachments/assets/8a63dfa6-7afa-43f9-b14e-2a2eca60ae41

- Added a new `Event_UndoDrawFailed` game event to the proto
  - Needs to be a new event since the successful undo draw is communicated through a EventContext
- Server sends the event when undo draw fails, right before returning the response code
- Client logs the failure to chat upon receiving the event 
  - Since this is a new event, older clients will safely completely ignore it

Still, having an entire single event for "undo draw failed" feels a bit weird.
For forwards-compatibility, perhaps we can instead make it a "generic status update that doesn't modify gamestate" event, containing a field holding an enum that signals what the status update in question is.
Then UNDO_DRAW_FAILED can be one of the enum values.
That means we can just add to the enum if we need more of this sort of thing in the future, instead of creating new events.

## Screenshots

Screenshot 2026-05-13 at 2 57 32 AM

### DIFF
--- a/cockatrice/src/game/log/message_log_widget.cpp
+++ b/cockatrice/src/game/log/message_log_widget.cpp
@@ -806,6 +806,12 @@ void MessageLogWidget::logUndoDraw(PlayerLogic *player, QString cardName)
     }
 }
 
+void MessageLogWidget::logUndoDrawFailed(PlayerLogic *player)
+{
+    appendHtmlServerMessage(
+        tr("%1 failed to undo their last draw.").arg(sanitizeHtml(player->getPlayerInfo()->getName())));
+}
+
 void MessageLogWidget::setContextJudgeName(QString name)
 {
     messagePrefix = QString("<span style=\"color:black\">");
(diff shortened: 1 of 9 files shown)


ZeldaZach (approved):
This will require both a server and a client update... big change for such a small feature?

ebbit1q:
tested, does not break anything, does require both server and client to upgrade but it's just a small addition to the protocol, I'm not 100% if this is the method that makes the most sense though, as we do already send an error response, sending an event in addition seems a bit strange to me, but our protocol's response values are rather unimportant so I get why we'd do it this way.

RickyRister:
> sending an event in addition seems a bit strange to me

The reason we send an event is so that the failure is verifiable by all players, including the judge.

ebbit1q:
the judge should be the one receiving the response to their action, however I can see how it's easier to simply send an event to all players

RickyRister:
If only judges can see the event, then that means the judge would need to be in the game at the time the event is sent. How judge calls are usually handled in custom constructed communities is that judges will only respond after they are pinged and ask for chat logs. We need all players to see the event so that it shows up in the chat logs.

RickyRister:
Renamed the proto to `Event_GameLogNotice`, so that we can reuse it in case we come across any events that only require a log response in the future.

ebbit1q:
I think that's a good idea, what other messages would the server generate perhaps?

PR #6890, opened 2026-05-13
[VDD] Fix minimum size by adding a compact mode to quickSettingsButtons

### LINKED ISSUE #6560: [VDE] Having three widgets open vertically stretches Cockatrice off the screen
**System Information:**

Client Version: 2.11.0 (beta 47)
Client Operating System: Ubuntu 25.10
Build Architecture: 64-bit
Qt Version: 6.9.2
System Locale: en_US
Install Mode: Standard

In the VDE, when there are more than 2 widgets open vertically, Cockatrice stretches to accommodate all of them and draws off the side of the screen. I am also unable to manually resize the widgets or the client window horizontally to make it fit. Once the third widget is closed, I can get the client window to auto-resize back to fitting by Restoring Down and then Maximizing again. I don't not experience this issue in the classic deck editor. 

VDE with two widgets open:

Image

VDE with three widgets open:

Image

**Steps to reproduce:**
 - Open the VDE
 - Have two widgets open and then open a third (e.g. have the Deck and Card Info widgets open and then open the Printing Selector).
- This bug also occurs for me if I open the VDE with the Deck, Card Info, and Printing Selector widgets set to open by default.

BruebachL:
## Related Ticket(s)
- Fixes #6560

SettingsButtons lock their minimum size to the size of the icon + text, which means their parents get a minimum size enforcement that may not be desirable.

- Add a compact mode to the SettingsButton which only draws the icon.
- Set compact mode for all filter buttons if size exceeds available space in the VDD

## Screenshots
image
image

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -327,6 +327,8 @@ set(cockatrice_SOURCES
     src/interface/widgets/tabs/api/edhrec/display/commander/edhrec_commander_api_response_bracket_navigation_widget.h
     src/interface/widgets/tabs/api/edhrec/display/commander/edhrec_commander_api_response_budget_navigation_widget.cpp
     src/interface/widgets/tabs/api/edhrec/display/commander/edhrec_commander_api_response_budget_navigation_widget.h
+    src/interface/widgets/utility/compact_push_button.cpp
+    src/interface/widgets/utility/compact_push_button.h
 )
 
 add_subdirectory(sounds)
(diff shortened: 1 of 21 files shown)


ebbit1q:
very nice, I also found that the tools of the visual deck view and the draw probability can limit the minimum width, maybe for those two we can switch to a two line layout in case the width is reduced?

RickyRister (approved):
Works. LGTM

https://github.com/user-attachments/assets/ee33958f-0e9e-483a-87d8-1973872d8abf

Unfortunately the VDE still can't be shrunk more due to some other tab in the widget. We can fix that in a future PR.

BruebachL:
I don't really know which boundary we're pushing up against at this point.

This squeezes the VDE down to about 400px width: A pretty reasonable size for the center widget.

The full VDE is at 1020~px on default layout.

RickyRister:
The layout jumps between two rows and three rows as you resize the window.

https://github.com/user-attachments/assets/54a8ca1f-178f-45e4-80fa-3a089a1eaeb5

https://github.com/user-attachments/assets/964a585f-5648-4f96-b069-4906efa3e6d1

Buttons in the deck analysis tab is cut off on the bottom. And also the button appearance changes, which looks jarring.

https://github.com/user-attachments/assets/fab7edf9-3e07-4551-a84b-27ca6dd7fe0a

PR #6893, opened 2026-05-16
Update Servatrice Dockerfile

tooomm:
- `Dockerfile`
  - Switch to Debian base image
    Starting with the newest LTS ubuntu image, the size did increase by ~1/3.
    To keep the image small and lean, we can use debian-slim images for the build & runtime stage which are 25-30% smaller.
  - Packages
    - Alphabetic sorting of packages
    - Remove not needed installs (git, file, build-essentials)
  - Copy only Servatrice binary to runtime stage
  - Mirror builds from CI more
    - Compile with ninja
    - make install --> cmake install
  - Default to non-root user for runtime
  - Add some extra hints
- `.dockerignore`
  - Revert the logic to strictly adding needed folder/files only

@maybeanerd Would love your opinion here.
As you know some more around Docker and use the image yourself, please give my changes a test as well.

---

TODO
- Move docker files to servatrice/docker folder and udpate path in context
- Scope caching to branch and pull from master similar to build workflow (see my other draft PR)
- (to check) Worth using lld as linker during compilation?


.dockerignore
```diff
@@ -1,8 +1,9 @@
...
-cockatrice/
-doc/
-oracle/
-Dockerfile
+# Add (Ignore) everything in root for minimum image
+/*
+
+# Exclude (Allow) only folder & files required for building Servatrice
+!cmake/
+!libcockatrice_*/
```
tooomm, line 6: Are all "libcockatrice_*" libraries really needed for Servatrice?

Dockerfile
```diff
@@ -1,43 +1,54 @@
...
-  libprotobuf32t64 \
-  libqt6sql6-mysql \
-  libqt6websockets6 \
-    && apt-get clean \
+
+# -------- Runtime Stage (clean) --------
+FROM debian:13-slim
+
+# Install runtime dependencies
+# Is libprotobuf-lite32t64 enough?
```
tooomm, line 38: I was wondering if "libprotobuf-lite32t64" is suifficient?
(Instead libprotobuf32t64)

.dockerignore
ebbit1q: no, not all of them, but the majority are

Dockerfile
ebbit1q: you can test it

.dockerignore
tooomm: It's really a long list by now 🙈 
```
libcockatrice_card
libcockatrice_deck_list
libcockatrice_filters
libcockatrice_interfaces
libcockatrice_models
libcockatrice_network
libcockatrice_protocol
libcockatrice_rng
libcockatrice_settings
libcockatrice_utility
```

Dockerfile
tooomm: I'll wait for @maybeanerd to test and confirm the current changes first.

PR #6894, opened 2026-05-16
Update Linux Dockerfiles of build pipeline

tooomm:
- Add gtest dev package to all images to avoid (hardcoded) fallback file download (no ca-certificate dependency)
  For Arch, gtest is commented out to test downloading and building from source repo
- Reformat for uniform style with Servatrice Dockerfile and better readability
- Remove not needed packages
  meta packages (base-devel, build-essential), git, qt6-base, clang-format, ca-certificates
- Cleanup Fedora images after package installation (similar to how it's done on Debian/Ubuntu images)

### DIFF
--- a/.ci/Arch/Dockerfile
+++ b/.ci/Arch/Dockerfile
@@ -1,19 +1,17 @@
 from archlinux:latest
 
 RUN pacman --sync --refresh --sysupgrade --needed --noconfirm \
-        base-devel \
-        ccache \
-        cmake \
-        git \
-        gtest \
-        mariadb-libs \
-        ninja \
-        protobuf \
-        qt6-base \
-        qt6-imageformats \
-        qt6-multimedia \
-        qt6-svg \
-        qt6-tools \
-        qt6-translations \
-        qt6-websockets \
+      ccache \
+      cmake \
+	  gcc \
+	  git \
+      gtest \
+      ninja \
+      protobuf \
+	  qt6-imageformats \
+	  qt6-multimedia \
+	  qt6-svg \
+	  qt6-tools \
+	  qt6-translations \
+	  qt6-websockets \
     && pacman --sync --clean --clean --noconfirm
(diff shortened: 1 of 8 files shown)


ebbit1q:
when we have the ci download gtest and compile with it, we also test that that part of the workflow keeps working

tooomm:
> when we have the ci download gtest and compile with it, we also test that that part of the workflow keeps working

Sure, I did comment gtest out in the Arch Dockerfile, so we'll test the fallback download there.

tooomm:
While there is some consistency across the Dockerfiles like image, multimedia, svg, tools and websocket of qt6 are listed for all.
Are all of them really needed though?

I'm also wondering about some other packages that are listed for individual distros only:
- Arch
  - qt6-translations
- Debian/Ubuntu
  - libgl-dev
  - liblzma-dev
  - qt6-l10n-tools
- Fedora
  - xz-devel
  - zlib-devel

Questions:
- ~"translations/l10n" packages are not installed on Fedora.
  Are they even needed? We do not build new strings?
- ~"tools-dev-tools" is only installed on Debian/Ubuntu.
- "libgl-dev"
  It looks like Arch and Fedora do not include OpenGL packages?
- "liblzma-dev"
  Debian/Ubuntu isntall lzma, Fedora uses the equivalant ("xz-devel"), Arch has nothing around lzma/xz installed, same for Servatrice (Debian)
- "zlib-devel"
  While Fedora installs zlib, Debian/Ubuntu and Arch does not
- While Debian/Ubuntu install libprotobuf-dev and protobuf-compiler, Arch and Fedora only add protobuf or protobuf-devel respectively
- For Debian/Ubuntu there is "libqt6multimedia6" and "qt6-multimedia-dev" listed, while the other distros only install the later
- Are mariadb, sql, mysql packages needed for normal builds?
  I removed them, but then figured that we by default build Cockatrice+Oracle & Servatrice...
  It felt not uniform though, like Arch and Fedora were only adding mariadb-libs, but nothing around sql/mysql

Happy for input from more experienced people around packages and distro specifics to help to clean this up once for all.

DawnFire42:
Hi, I did some digging in linux for you. I wasn't entirely correct in my message. I'll try to give you an answer for each of your questions.

PR #6896, opened 2026-05-16
Enable `RemoveSemicolon` & update format.sh

tooomm:
## Related Ticket(s)
- Related #6887

Option was commented out, but we run on v18

- Sort options alphabetically within languages
- Enable RemoveSemicolon option
- Remove comments on old clang-format versions
- Use libcockatrice_* to match all libs dynamically in format.sh


ebbit1q (approved):
thanks

PR #6903, opened 2026-05-18
[Game/Zones] Simple move refactor to differentiate between logic and graphics for zones

BruebachL:
We already differentiate these files into zones/ and zones/logic.
#6342 already laid the groundwork for the game_graphics folder, so we can hopefully, at some point, deal with the game and the graphics separately. Most of these classes are already fine: They concern themselves with drawing things and don't really modify state. Some of them manually assemble GameCommands, which is a bit odd but they at least pass those on to the playerEventController so, that's fine-ish for now.

Anyway, as I've said, in order to keep all of this atomic, accountable and easy to review, let's just start with this pure move refactor and clean things "properly" later.

- Move zone files from game/zones/ to game_graphics/zones
- game/zones/logic stays untouched for now


cockatrice/src/game_graphics/zones/card_zone.h
```diff
@@ -7,9 +7,9 @@
 #ifndef CARDZONE_H
 #define CARDZONE_H
 
-#include "../../game_graphics/board/abstract_graphics_item.h"
-#include "../../game_graphics/board/graphics_item_type.h"
-#include "logic/card_zone_logic.h"
+#include "../../game/zones/logic/card_zone_logic.h"
```
RickyRister: do we still need the `logic` folder?

BruebachL: I guess not, I thought the additional string might make it marginally more easy to identify which classes depended on which type (logic vs graphics) while I was doing the refactor but I may as well clean this up now.

RickyRister (approved):

PR #6904, opened 2026-05-18
[Game] Move state fields out of CardItem

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

CardItem is doing too much. It is responsible for the Qt rendering side of things, while also holding state.
Qt Widgets/Render devices should not hold state (game and game_graphics differentiation).

- Pull out some fields and replace CardItem methods with thin wrappers around new state accessors/setters (I'll refactor this later but let's not break too many call sites right now)
- Implement new CardState, which emits signals on modification

This does not adjust to the refactor yet, i.e. the graphics widget isn't signal driven yet.

Atomic changes and all that.

### DIFF
--- a/cockatrice/src/game/board/card_state.h
+++ b/cockatrice/src/game/board/card_state.h
@@ -24,14 +24,14 @@ private:
 signals:
     void stateChanged();
 
-    void attackingChanged(bool);
-    void countersChanged(QMap<int, int>);
-    void annotationChanged(QString);
-    void ptChanged(QString);
-    void doesntUntapChanged(bool);
-    void destroyOnZoneChangeChanged(bool);
-    void attachedToChanged(CardItem *);
-    void zoneChanged(CardZoneLogic *);
+    void attackingChanged(bool newValue);
+    void countersChanged(const QMap<int, int> &newCounters);
+    void annotationChanged(const QString &newAnnotation);
+    void ptChanged(const QString &newPt);
+    void doesntUntapChanged(bool newValue);
+    void destroyOnZoneChangeChanged(bool newValue);
+    void attachedToChanged(CardItem *newAttachedTo);
+    void zoneChanged(CardZoneLogic *newZone);
 
 public:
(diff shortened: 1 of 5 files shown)


cockatrice/src/game/board/card_state.h
```diff
@@ -0,0 +1,188 @@
...
+    void zoneChanged(CardZoneLogic *);
+
+public:
+    explicit CardState(QObject *parent, CardZoneLogic *_zone) : QObject(parent), zone(_zone)
+    {
+    }
+
+    void resetState(bool keepAnnotations)
+    {
+        attacking = false;
```
RickyRister: why is everything in the `.h` file

cockatrice/src/game/board/card_state.h
```diff
@@ -0,0 +1,188 @@
...
+
+    CardItem *attachedTo = nullptr;
+    CardZoneLogic *zone = nullptr;
+
+signals:
+    void stateChanged();
+
+    void attackingChanged(bool);
+    void countersChanged();
+    void annotationChanged();
```
RickyRister: can we have the signals pass the new values as params

cockatrice/src/game/board/card_state.h
```diff
@@ -0,0 +1,103 @@
...
+    bool doesntUntap = false;
+    bool destroyOnZoneChange = false;
+
+    CardItem *attachedTo = nullptr;
+    CardZoneLogic *zone = nullptr;
+
+signals:
+    void stateChanged();
+
+    void attackingChanged(bool);
```
RickyRister: can we name the params `newValue` or something like that? I know it's probably obvious what the param represents, but it's better to be explicit just in case.

cockatrice/src/game/board/card_state.h
```diff
@@ -0,0 +1,103 @@
...
+    void countersChanged(QMap<int, int>);
```
RickyRister: also, we should be passing objects by const reference

RickyRister (approved):
Left some comments. Otherwise looks good.

PR #2773, opened 2017-06-18
Handling of tokens/related cards improved.

Psithief:
## Related Ticket(s)
- Fixes #2733 
- Fixes #2262 
- Related https://github.com/Cockatrice/Magic-Token/pull/20

## What will change with this Pull Request?
* Added relation attributes to the XML <(reverse-)related> tags (attach, exclude, count)
* Support for variable count token creation.
* Actions that attach to tokens are marked as such.
* Create another token specifies type of token in action title.
* Multicoloured tokens have multicoloured background.

## Side effects
Players will need to run Oracle once after installing this update to ensure the non-token related cards attach as before.

PR #6909, opened 2026-05-18
fix cards attaching/transforming to the wrong card across players

ebbit1q:
## Related Ticket(s)
- presumably introduced in #2773

when a player transforms a card controlled by another player that has the same cardid as a card on that player's own field, it'll destroy and replace it, similarly if it is instructed to "attach" it will attach to the card with the same id on the player's own side of the board.

- check for and do not allow attaching/transforming of tokens created by an opponent's card

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -1070,9 +1070,9 @@ bool PlayerActions::createRelatedFromRelation(const CardItem *sourceCard, const 
         }
     } else {
         CardRelationType attachType;
+        // do not attempt to attach to another player's cards, this causes the card to attempt to attach to the same
+        // cardid on the local player's field instead, which is an entirely different card!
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/player/player_actions.cpp
```diff
@@ -1069,7 +1069,14 @@ bool PlayerActions::createRelatedFromRelation(const CardItem *sourceCard, const
             createCard(sourceCard, dbName, CardRelationType::DoesNotAttach, persistent);
         }
     } else {
-        auto attachType = cardRelation->getAttachType();
+        CardRelationType attachType;
+        if (player->getPlayerInfo()->getLocalOrJudge()) {
+            // do not attempt to attach to another player's cards, this causes the card to attempt to attach to the same
```
RickyRister: nit: move the comment to either above the if statement or into the else block

RickyRister (approved):

PR #6912, opened 2026-05-19
[Player] Stop reaching into graphics_item and emit signals instead for conceded and zoneId

BruebachL:
The player logic can be more agnostic to the graphics item by simply emitting signals that the graphics item hooks up to.

- Stop reaching into graphicsItem for conceded and zoneIdChanged.

### DIFF
--- a/cockatrice/src/game/player/player.cpp
+++ b/cockatrice/src/game/player/player.cpp
@@ -89,7 +89,6 @@ void Player::setConceded(bool _conceded)
     if (conceded != _conceded) {
         conceded = _conceded;
 
-        getGraphicsItem()->setVisible(!conceded);
         if (conceded) {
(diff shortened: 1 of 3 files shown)


cockatrice/src/game/player/player.cpp
```diff
@@ -100,7 +99,7 @@ void Player::setConceded(bool _conceded)
 void Player::setZoneId(int _zoneId)
 {
     zoneId = _zoneId;
```
RickyRister: if we're making this setter also emit stuff, do we want to add the oldValue != newValue check to it?

BruebachL: Sure, may as well.

RickyRister (approved):

PR #6918, opened 2026-05-20
[Game][Arrows] Split Arrows into ArrowData and ArrowItem

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- Introduce arrow_data to allow logic layer to process arrow events without needing to know about arrow graphics
- move arrow lifecycle from player to game_scene
- Have arrows stop tracking to and from and other redundancies for position updates and just listen to Qt signals instead
- same with destruction

### DIFF
--- a/cockatrice/src/game/game_scene.cpp
+++ b/cockatrice/src/game/game_scene.cpp
@@ -82,7 +82,6 @@ void GameScene::addPlayer(PlayerLogic *player)
 {
     qCInfo(GameScenePlayerAdditionRemovalLog) << "GameScene::addPlayer name=" << player->getPlayerInfo()->getName();
 
-    players << player->getGraphicsItem();
     playerViews.insert(player->getPlayerInfo()->getId(), player->getGraphicsItem());
     addItem(player->getGraphicsItem());
     connect(player->getGraphicsItem(), &PlayerGraphicsItem::sizeChanged, this, &GameScene::rearrange);
@@ -122,7 +121,6 @@ void GameScene::removePlayer(PlayerLogic *player)
         }
     }
     auto *view = playerViews.take(player->getPlayerInfo()->getId());
-    players.removeOne(view);
     removeItem(view);
     rearrange();
 }
@@ -204,7 +202,7 @@ QList<PlayerLogic *> GameScene::collectActivePlayers(int &firstPlayerIndex) cons
     firstPlayerIndex = 0;
     bool firstPlayerFound = false;
 
-    for (auto *pgItem : players) {
+    for (auto *pgItem : playerViews.values()) {
         PlayerLogic *p = pgItem->getPlayer();
         if (p && !p->getConceded()) {
             activePlayers.append(p);
@@ -368,7 +366,7 @@ void GameScene::resizeColumnsAndPlayers(const QList<qreal> &minWidthByColumn, qr
     }
 }
 
-void GameScene::onArrowCreateRequested(ArrowData data)
+void GameScene::onArrowCreateRequested(const ArrowData &data)
 {
     auto *startView = playerViews.value(data.startPlayerId);
     auto *targetView = playerViews.value(data.targetPlayerId);
--- a/cockatrice/src/game/game_scene.cpp
+++ b/cockatrice/src/game/game_scene.cpp
@@ -83,8 +83,25 @@ void GameScene::addPlayer(PlayerLogic *player)
     qCInfo(GameScenePlayerAdditionRemovalLog) << "GameScene::addPlayer name=" << player->getPlayerInfo()->getName();
 
     players << player->getGraphicsItem();
+    playerViews.insert(player->getPlayerInfo()->getId(), player->getGraphicsItem());
     addItem(player->getGraphicsItem());
     connect(player->getGraphicsItem(), &PlayerGraphicsItem::sizeChanged, this, &GameScene::rearrange);
+
+    connect(player, &PlayerLogic::concededChanged, this, [this](int id, bool conceded) {
+        if (conceded) {
+            clearArrowsForPlayer(id);
+        }
+        rearrange();
+    });
+
+    connect(player, &PlayerLogic::arrowCreateRequested, this, &GameScene::onArrowCreateRequested);
+    connect(player, &PlayerLogic::arrowDeleteRequested, this, &GameScene::onArrowDeleteRequested);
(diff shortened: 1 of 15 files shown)


cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -606,17 +606,9 @@ void TabGame::actNextPhaseAction()
...
-        if (!player->getPlayerInfo()->getLocal()) {
-            continue;
-        }
-        QMapIterator<int, ArrowItem *> arrowIterator(player->getArrows());
-        while (arrowIterator.hasNext()) {
-            ArrowItem *a = arrowIterator.next().value();
-            emit arrowDeletionRequested(a->getId());
-        }
+    auto *local = game->getPlayerManager()->getActiveLocalPlayer(game->getGameState()->getActivePlayer());
+    if (local) {
```
ZeldaZach, line 610: `GameScene::clearArrowsForPlayer` calls arrow->delArrow(), which only does deleteLater() — it never sends Command_DeleteArrow to the server. The original implementation emitted arrowDeletionRequested(id), which is still wired (tab_game.cpp:965) to GameEventHandler::handleArrowDeletion → Command_DeleteArrow.

So before this PR, the menu action removed arrows for everyone; after this PR, it only hides them on your screen and the server (and other clients) still think they exist. They will likely re-appear on the next state sync.

cockatrice/src/game/player/player_event_handler.cpp
```diff
@@ -90,25 +91,41 @@ void PlayerEventHandler::eventRollDie(const Event_RollDie &event)
 
 void PlayerEventHandler::eventCreateArrow(const Event_CreateArrow &event)
 {
-    ArrowItem *arrow = player->addArrow(event.arrow_info());
-    if (!arrow) {
```
ZeldaZach, line 94: By not returning early, you introduce a potential blank log message.

Old code early-returned if player->addArrow(...) returned nullptr (cards/zones not resolvable). New code emits the log entry whenever startPlayer && targetPlayer exist, even if the start/target cards couldn't be resolved — producing a log line with empty card names. Suggest gating on !startCardName.isEmpty() && (data.isPlayerTargeted() || !targetCardName.isEmpty()) to preserve the prior behavior.

ZeldaZach (changes requested):

cockatrice/src/game/game_scene.h
```diff
@@ -42,13 +44,15 @@ class GameScene : public QGraphicsScene
...
     QList<PlayerGraphicsItem *> players;                ///< All player graphics items
+    QMap<int, PlayerGraphicsItem *> playerViews;        ///< ID lookup
     QList<QList<PlayerGraphicsItem *>> playersByColumn; ///< Players organized by column
-    QList<ZoneViewWidget *> zoneViews;                  ///< Active zone view widgets
-    QSize viewSize;                                     ///< Current view size
-    QPointer<CardItem> hoveredCard;                     ///< Currently hovered card
-    QBasicTimer *animationTimer;                        ///< Timer for card animations
-    QSet<CardItem *> cardsToAnimate;                    ///< Cards currently animating
-    int playerRotation;                                 ///< Rotation offset for player layout
+    QMap<int, ArrowItem *> arrowRegistry;
```
RickyRister: nit: comment this field so the autoformatter doesn't break spacing for the following comments.

cockatrice/src/game/game_scene.h
```diff
@@ -196,6 +200,11 @@ public slots:
     QTransform getViewTransform() const;
     QTransform getViewportTransform() const;
 
+    void onArrowCreateRequested(ArrowData data);
```
RickyRister: nit: name these methods `addArrow` and `deleteArrow`. It's better to name methods by describing what they do rather than when they're used, when possible.

cockatrice/src/game/game_scene.h
```diff
@@ -207,6 +216,7 @@ public slots:
     void sigStartRubberBand(const QPointF &selectionOrigin);
     void sigResizeRubberBand(const QPointF &cursorPoint, int selectedCount);
     void sigStopRubberBand();
+    void requestArrowDeletion(int arrowId);
```
RickyRister, line 218: signals should be named `arrowDeletionRequested`

cockatrice/src/game/game_scene.cpp
```diff
@@ -83,8 +83,25 @@ void GameScene::addPlayer(PlayerLogic *player)
     qCInfo(GameScenePlayerAdditionRemovalLog) << "GameScene::addPlayer name=" << player->getPlayerInfo()->getName();
 
     players << player->getGraphicsItem();
+    playerViews.insert(player->getPlayerInfo()->getId(), player->getGraphicsItem());
     addItem(player->getGraphicsItem());
     connect(player->getGraphicsItem(), &PlayerGraphicsItem::sizeChanged, this, &GameScene::rearrange);
+
+    connect(player, &PlayerLogic::concededChanged, this, [this](int id, bool conceded) {
```
RickyRister, line 89: feels like the sort of thing we should pull out to a method. There might be more things we want to do in response to conceding in the future.

cockatrice/src/game/game_scene.h
```diff
@@ -42,13 +44,15 @@ class GameScene : public QGraphicsScene
 
     PhasesToolbar *phasesToolbar;                       ///< Toolbar showing game phases
     QList<PlayerGraphicsItem *> players;                ///< All player graphics items
+    QMap<int, PlayerGraphicsItem *> playerViews;        ///< ID lookup
```
RickyRister: Are these two fields storing the exact same things? If so, get rid of one of them so that we don't have the worry about keeping them in sync.

cockatrice/src/game/game_scene.h
RickyRister: pass by const ref

cockatrice/src/game/game_scene.cpp
BruebachL: Let's have things functional for now and worry about proper cleanup later, there are a lot of things like that to adjust and things might not stay where they currently are.

BruebachL: It's just that I have a 46 file commit piggy-backing off of this commit and I fear I will go insane if I don't keep my eye on the prize for this refactor.

ZeldaZach (approved):

PR #6919, opened 2026-05-21
[Game][Player] Pull out graphics_items out of player_logic

BruebachL:
This PR is the first structural step in a larger refactor to decouple the game's UI layer from its logic layer, enabling a future pub/sub event system. The core problem was that Player (now PlayerLogic) directly owned and created PlayerGraphicsItem, PlayerMenu, and all associated graphics classes, making the logic layer tightly coupled to Qt's widget/graphics stack.

- AbstractGame drops TabGame*. AbstractGame now takes a plain QObject *parent. Game and Replay extract the one bool they needed (isLocalGame) locally before calling the base constructor.
- PlayerLogic no longer owns graphics
- PlayerGraphicsItem and PlayerMenu are no longer constructed inside PlayerLogic. The constructor no longer creates graphics objects. Logic signals (_arrowCreateRequested, arrowDeleteRequested, arrowsCleared, counterAdded, counterRemoved, concededChanged, zoneIdChanged, requestZoneViewToggle, requestSortHand_) replace direct calls into the graphics layer.
- GameScene owns PlayerGraphicsItem
GameScene::addPlayer(PlayerLogic*) now creates PlayerGraphicsItem internally and registers it in a QMap<int, PlayerGraphicsItem*> playerViews for ID-based lookup. GameScene::viewForPlayer(int) is the new access point for anything that needs the view for a given player.
- All menu classes take PlayerGraphicsItem* instead of PlayerLogic*
- CardMenu, CustomZoneMenu, GraveyardMenu, HandMenu, LibraryMenu, MoveMenu, PtMenu, RfgMenu, SayMenu, SideboardMenu, UtilityMenu, and PlayerMenu all now hold PlayerGraphicsItem* and reach logic through player->getPlayerLogic(). These are view-layer classes and this reflects their actual nature.
- PlayerDialogs introduced
- Dialog handling is extracted out of PlayerActions into a new PlayerDialogs class owned by PlayerGraphicsItem. PlayerActions emits requestDialog signals; PlayerDialogs connects to them and owns the QWidget dialogParent needed to show dialogs.
- CardItem::mouseReleaseEvent wired via signal
- Right-click context menus are now triggered via AbstractCardItem::rightClicked signal connected to GameScene::onCardRightClicked, which resolves the PlayerGraphicsItem from playerViews and shows the menu.

## TODO:
Remove AbstractGame::activeCard (CardItem* on the logic layer)
Remove CardZoneLogic::updateGraphics() / setGraphicsVisibility() signals
Move files into game_graphics/

These are left as subsequent PRs to keep this one reviewable. The goal here is establishing the structural boundary so those cleanups can happen without the whole thing collapsing.

## Patterns

We essentially had two patterns in PlayerActions:

```
// Pattern 1 — pure command builder. Takes primitives, builds protobuf, sends.
void PlayerActions::actDrawCards(int number) {
    Command_DrawCards cmd;
    cmd.set_number(number);
    sendGameCommand(cmd);
}

// Pattern 2 — pulls UI state itself instead of receiving it.
void PlayerActions::actIncPT(int deltaP, int deltaT) {
    for (auto card : player->getGameScene()->selectedCards()) { // ← reaches into scene
        // ... builds command from card data
    }
}
```

Pattern 2 is what this PR fixes. Pattern 1 is the correct and intended usage. 

## What PlayerActions actually is
It's a command builder. Given a user's intention and the current game state, it constructs the correct protobuf message and sends it through GameEventHandler. That's it. Everything that isn't command building is either:

- UI translation: figuring out which cards the user means (selected cards, active card), which belongs in the caller
- Dialog orchestration: asking the user for parameters (how many cards?), which belongs in PlayerDialogs

## Screenshots
image

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -5,7 +5,6 @@
 #include "../../interface/widgets/tabs/tab_game.h"
 #include "../../interface/widgets/utility/get_text_with_max.h"
 #include "../board/card_item.h"
-#include "../client/settings/card_counter_settings.h"
 #include "../dialogs/dlg_move_top_cards_until.h"
 #include "../dialogs/dlg_roll_dice.h"
 #include "../zones/view_zone_logic.h"
@@ -39,6 +38,8 @@ static constexpr int MOVE_TOP_CARD_UNTIL_INTERVAL = 100;
 PlayerActions::PlayerActions(PlayerLogic *_player)
     : QObject(_player), player(_player), lastTokenTableRow(0), movingCardsUntil(false)
 {
+    connect(this, &PlayerActions::requestZoneViewToggle, player, &PlayerLogic::onRequestZoneViewToggle);
+
     moveTopCardTimer = new QTimer(this);
     moveTopCardTimer->setInterval(MOVE_TOP_CARD_UNTIL_INTERVAL);
     moveTopCardTimer->setSingleShot(true);
@@ -133,12 +134,12 @@ void PlayerActions::playCardToTable(const CardItem *card, bool faceDown)
 
 void PlayerActions::actViewLibrary()
 {
-    player->getGameScene()->toggleZoneView(player, ZoneNames::DECK, -1);
+    emit requestZoneViewToggle(ZoneNames::DECK, -1);
 }
 
 void PlayerActions::actViewHand()
 {
-    player->getGameScene()->toggleZoneView(player, ZoneNames::HAND, -1);
+    emit requestZoneViewToggle(ZoneNames::HAND, -1);
 }
 
 /**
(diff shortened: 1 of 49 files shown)


RickyRister:
Is the new `PlayerActions` class suppose to be logic or ui? If it's logic, then the dialogs should have been handled by another class before the code reaches this class, and it shouldn't be responsible for ui-only actions such as `sortHand`.

I think what we should actually do first is split the `PlayerActions` class into a ui component and a logic component. Then the ui `PlayerActions` class can hold a pointer to the logic `PlayerActions` class, allowing it to directly call into the logic `PlayerActions`, and avoid this mess of signals. The ui `PlayerActions` class can be responsible for translating the action to a form that the logic `PlayerActions` can process.

BruebachL:
You're describing PlayerDialogs, which is the UI equivalent to playerActions

BruebachL:
I kept a lot of things in playerActions as stubs/forwards for now because a lot of these actions are literally just to keep things compiling because one change, breaks every dependency attached to it. I'd prefer to do cleanup and reason about where things should belong once we have things separated.

i.e. it's impossible to remove the TabGame dependency from AbstractGame because suddenly, every PlayerActions dialog is missing a parent, which is solved by moving those dialogs out of PlayerActions and into PlayerDialogs, which belongs to the GraphicsItem and thus has a valid parent available to it.

BruebachL:
We also run less risk of breaking things by separating refactor and behavioral PRs

BruebachL:
>  If it's logic, then the dialogs should have been handled by another class before the code reaches this class, and it shouldn't be responsible for ui-only actions such as sortHand.

I don't disagree with you but it's just not what I had in scope when I made this PR and my puny brain can only process so many changes at a time

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -39,6 +38,8 @@ static constexpr int MOVE_TOP_CARD_UNTIL_INTERVAL = 100;
 PlayerActions::PlayerActions(PlayerLogic *_player)
```
RickyRister, line 38: `PlayerActions` is predominantly accessed from the menu classes and the tab classes, which suggests that it should be a ui class. It makes more sense for it to keep a pointer to `PlayerGraphicsItem` instead of `PlayerLogic`, like with what the menu classes do, and have it forward the logic bits to a separate logic PlayerActions class.

RickyRister: > I don't disagree with you but it's just not what I had in scope when I made this PR

I think what would make the changes easier to review is to just have `PlayerActions` still be a graphics class in this PR, so that the code in here can stay mostly the same. Then in a follow-up PR we can split the logic parts from `PlayerActions`

BruebachL: Disagree.

BruebachL: As I said earlier, you cannot change one part without everything else falling apart.

cockatrice/src/game/player/player_logic.cpp
```diff
@@ -304,9 +295,9 @@ CounterState *PlayerLogic::getLifeCounter() const
     return nullptr;
 }
 
-void PlayerLogic::incrementAllCardCounters()
+void PlayerLogic::incrementAllCardCounters(QList<CardItem *> cardsToUpdate)
```
RickyRister: ah right, this method. This method should've been in `PlayerActions`. It was added by someone else while we were in the middle of the player refactor, and I was always meaning to get around to moving it to the correct place but never got around to it.

I'll do it now.

cockatrice/src/game/player/player_graphics_item.h
```diff
@@ -55,11 +56,16 @@ class PlayerGraphicsItem : public QGraphicsObject
         return static_cast<GameScene *>(scene());
     }
 
-    PlayerLogic *getPlayer() const
+    PlayerLogic *getPlayerLogic() const
```
RickyRister: can we rename this to `getLogic` since it takes up less space and we're going to be calling it a lot.

BruebachL:
Closed in favor of #6944, #6945 and #6946

PR #6922, opened 2026-05-22
disable the filters widget on the vde

ebbit1q:
hides the non functional filters widget from the vde


ZeldaZach (approved):

RickyRister (changes requested):
#6924

PR #6925, opened 2026-05-22
  Fix card stacking overflow when vertical zone has many cards

DawnFire42:
## Related Ticket(s)
- none that I could find

- I saw this bug in the stacking behaviour: when adding too many cards they would clip out of bounds. 
- So my previous changes to it failed to handle that edge case.
- This should rectify the issue.

  Previously, minOffset (10px) was enforced unconditionally, causing card
  tops to overflow zone bounds when many cards were stacked. For example,
  50 cards in a 200px zone would place the last card's top at y=490.

  Now offsets compress below minOffset when necessary to keep all card
  tops visible. The constraint is guarded by !allowBottomOverflow to
  preserve future clipping zone semantics.


RickyRister:
Huh, didn't realize we had that bug

Image

RickyRister (approved):
Yeah, seems fixed

Image

PR #6927, opened 2026-05-22
[Game][Arrows] Don't request deletion on inbound arrow deletion signal again but react to it locally

BruebachL:
I hooked up the PlayerEventHandler (inbound events) to emit a "deleteArrowRequested" signal from the player logic. It should instead send a "arrowDeleted" signal, which the scene then reacts to by actually deleting the arrow from the view.

- Implement new signal, hook up, react to it.

### DIFF
--- a/cockatrice/src/game/game_scene.cpp
+++ b/cockatrice/src/game/game_scene.cpp
@@ -93,6 +93,7 @@ void GameScene::addPlayer(PlayerLogic *player)
         rearrange();
     });
 
+    connect(player, &PlayerLogic::arrowDeleted, this, &GameScene::onArrowDeleted);
     connect(player, &PlayerLogic::arrowCreateRequested, this, &GameScene::onArrowCreateRequested);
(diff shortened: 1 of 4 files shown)


RickyRister (approved):

ebbit1q:
is this regression related to #6918 ?

BruebachL:
> is this regression related to #6918 ?

Yes, most likely. Ricky says arrows still don't behave as they ought to but honestly, looking at the old code, it's not exactly easy to reason about what proper behavior is for arrows?

ebbit1q:
my testing shows arrows on current master are working as expected with the exception of the case in which you make a cycle involving multiple players:
- start an online game with at least 2 players (local also has the problem but does not crash)
- drag a card to both player's fields
- draw an arrow from card a to card b and a second arrow from card b to card a
- right click one arrow to remove it
- attempt to right click the other arrow, which fails
- the other player will still see the other arrow but not the first
- closing the game causes a stall

PR #6928, opened 2026-05-22
[Game] Move graphics out of game and into game_graphics

BruebachL:
## Related Ticket(s)
- Depends on #6944
- Depends on #6945
- Depends on #6946

Move refactor to introduce cleaner separations.


cockatrice/src/game/player/player_actions.h
```diff
@@ -56,30 +57,71 @@ class PlayerActions : public QObject
...
+    void actRequestMulliganDialog();
+    void actMulligan(int number);
     void actMulliganSameSize();
     void actMulliganMinusOne();
     void doMulligan(int number);
 
-    void actPlay();
-    void actPlayFacedown();
-    void actHide();
+    void actPlay(QList<CardItem *> selectedCards);
```
RickyRister, line 127: Okay, i think I know why the code in PlayerActions feels so wonky. It's because these methods on logic classes feel like they should be taking lists of CardState instead of CardItem.

Though it seems like the coords are stored in CardItem instead of CardState. Are we planning to move the coords into CardState?

BruebachL: Yes, I have not gotten around to properly splitting CardItem useage into CardState for logic and CardItem for graphics, I tackled all these other things first

BruebachL: Remember, the original card state PR talked about forwarding callers to thin wrappers for now

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -982,9 +987,8 @@ void PlayerActions::actCreateAllRelatedCards()
 
     if (relatedCards.length() == 1) {
         cardRelation = relatedCards.at(0);
-        if (createRelatedFromRelation(sourceCard, cardRelation)) {
-            ++tokensTypesCreated;
-        }
+        actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
+        ++tokensTypesCreated;
```
ZeldaZach: Behavior regression: `tokensTypesCreated` is now incremented unconditionally. The old `createRelatedFromRelation` returned `bool` and this branch did `if (createRelatedFromRelation(...)) { ++tokensTypesCreated; }`, so the count only went up on success (e.g. user did not cancel the variable-count dialog). With the rewrite, `actRequestCreateRelatedFromRelationDialog` returns void, so we always increment — which means the `tokensTypesCreated == 1` check below (line 1018 / 1033) can fire even when nothing was actually created.

Suggest threading the bool back through, either by giving `actRequestCreateRelatedFromRelationDialog` a synchronous return value (the signal/slot is direct-connected, so this is feasible) or by emitting a result signal that updates the counter.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -996,9 +1000,9 @@ void PlayerActions::actCreateAllRelatedCards()
         switch (nonExcludedRelatedCards.length()) {
             case 1: // if nonExcludedRelatedCards == 1
                 cardRelation = nonExcludedRelatedCards.at(0);
-                if (createRelatedFromRelation(sourceCard, cardRelation)) {
-                    ++tokensTypesCreated;
-                }
+                actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
+                ++tokensTypesCreated;
+
```
ZeldaZach: Same regression as the `relatedCards.length() == 1` branch above — `++tokensTypesCreated` is now unconditional where it was previously gated on `createRelatedFromRelation` returning `true`.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -1045,48 +1049,74 @@ void PlayerActions::actCreateAllRelatedCards()
...
+            return;
         }
+    } else {
+        count = cardRelation->getDefaultCount();
+    }
+
+    if (count > 1) {
         for (int i = 0; i < count; ++i) {
             createCard(sourceCard, dbName, CardRelationType::DoesNotAttach, persistent);
         }
```
ZeldaZach, line 1093: Behavior change for variable-count relations: the old code's structure was `if (isVariable) { for loop with DoesNotAttach } else if (defaultCount > 1) { for loop with DoesNotAttach } else { attach branch }`. So a variable relation **always** used `DoesNotAttach`, regardless of the count chosen.

The new code collapses to `if (count > 1) { DoesNotAttach loop; return; } /* else attach branch */`. That means a variable relation where the user types `1` in the dialog now falls through to the attach branch and may call `playCardToTable(sourceCard, false)` on attach-type relations — different from the old behavior.

If this was intentional, worth a comment explaining; otherwise gate on `cardRelation->getIsVariable()` to preserve the old branch.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -1045,48 +1049,74 @@ void PlayerActions::actCreateAllRelatedCards()
...
+    return;
+}
 
-        createCard(sourceCard, dbName, attachType, persistent);
+void PlayerActions::onRelatedCardCreated(const CardItem *sourceCard, const CardRelation *cardRelation)
+{
+    /*
+     * If we make a token via "Token: TokenName"
+     * then let's allow it to be created via "create another token"
+     */
```
ZeldaZach, line 1125: Possible null deref: this function unconditionally calls `cardRelation->getCanCreateAnother()` and later `sourceCard->getCard().getPrinting()` with no null checks. `createRelatedFromRelation` (just above) does check both for null and returns early; `onRelatedCardCreated` is called right after it in `PlayerDialogs::onCreateRelatedFromRelationDialogRequested`, so they need to agree on the contract.

Either add the same null guards here, or only invoke `onRelatedCardCreated` after a confirmed successful creation in the dialog handler.

cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
+#include "player_dialogs.h"
+
+#include "../../interface/widgets/utility/get_text_with_max.h"
+#include "../board/card_item.h"
+#include "../client/settings/card_counter_settings.h"
```
ZeldaZach: Path nit: this file is at `game_graphics/player/`, so `../client/...` resolves to `game_graphics/client/...` (doesn't exist) rather than `src/client/...`. It builds because of how the include search paths resolve, but it's inconsistent with the other 2-level-deep files in this tree — compare e.g. `library_menu.cpp` which correctly uses `../../../client/settings/...` from `game_graphics/player/menu/`. Should be `../../client/settings/card_counter_settings.h` here.

cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+void PlayerDialogs::onSetPTDialogRequested(const QString &oldPT)
+{
+    bool ok;
+    emit requestDialogSemaphore(true);
+    QString pt = getTextWithMax(dialogParent(), tr("Change power/toughness"), tr("Change stats to:"), QLineEdit::Normal,
+                                oldPT, &ok);
+    emit requestDialogSemaphore(false);
+
+    if (!ok) { // player->clearCardsToDelete() ||
```
ZeldaZach: Looks like a forgotten TODO — the old `actSetPT` in `PlayerActions` had `if (player->clearCardsToDelete() || !ok) { return; }`, which bailed out if cards were deleted while the dialog was open (e.g., the card the user was editing got removed by an opponent's action). Commenting out the check means we'll now proceed to call `actSetPT(selectedCards, pt)` on cards that may already be gone.

Same issue in `onSetAnnotationDialogRequested` (line 267) and `onSetCardCounterDialogRequested` (line 285). Worth either restoring the check (probably as a signal from `PlayerDialogs` back to `PlayerLogic` to ask) or filing a follow-up to remove the comment and the `clearCardsToDelete` semantics entirely if no longer needed.

cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+{
+    emit requestDialogSemaphore(true);
+    AnnotationDialog *dialog = new AnnotationDialog(dialogParent());
+    dialog->setOptions(QInputDialog::UsePlainTextEditForTextInput);
+    dialog->setWindowTitle(tr("Set annotation"));
+    dialog->setLabelText(tr("Please enter the new annotation:"));
+    dialog->setTextValue(oldAnnotation);
+    bool ok = dialog->exec();
+    emit requestDialogSemaphore(false);
+    if (!ok) { // player->clearCardsToDelete() ||
```
ZeldaZach: Same `clearCardsToDelete()` guard removal as line 250 — was active in the original `actSetAnnotation`.

cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+    emit requestDialogSemaphore(true);
+
+    auto &cardCounterSettings = SettingsCache::instance().cardCounters();
+    QString counterName = cardCounterSettings.displayName(counterId);
+
+    AbstractCounterDialog dialog(counterName, oldValueForDlg, dialogParent());
+    int ok = dialog.exec();
+
+    emit requestDialogSemaphore(false);
+    if (!ok) { // player->clearCardsToDelete() ||
```
ZeldaZach: Same `clearCardsToDelete()` guard removal as line 250 — was active in the original `actSetCardCounter`.

cockatrice/src/game/player/player_logic.h
```diff
@@ -7,17 +7,15 @@
 #ifndef PLAYER_H
 #define PLAYER_H
 
+#include "../../game_graphics/player/player_area.h"
```
ZeldaZach, line 10: Architecture nit: `player_logic.h` now `#include`s `game_graphics/player/player_area.h`. This pulls the graphics layer into the logic layer's public header, which works against the separation this PR is aiming for. If `PlayerArea` is only needed by value/forward-declarable, a forward declaration here plus the full include in the `.cpp` would keep the layering clean.

ZeldaZach (commented):
Nice refactor — the logic/graphics separation is clearly progressing and the signal-based dialog routing through `PlayerDialogs` is a clean pattern. I focused on behavior preservation in the action wiring and spotted a few regressions that look worth addressing before merge. Most are in the `actCreateRelatedCard` / `createRelatedFromRelation` rewrite where the synchronous bool return was replaced with fire-and-forget signal emission.

**Summary of findings:**
- **Behavior regression** in `actCreateAllRelatedCards`: `tokensTypesCreated` is now incremented unconditionally (old code only counted on success).
- **Behavior regression** in `createRelatedFromRelation`: a variable-count relation where the user picks `1` now takes the attach branch; old code always used `DoesNotAttach` for variable relations.
- **Possible crash** in `onRelatedCardCreated`: no null checks on `sourceCard` / `cardRelation`, and it’s called unconditionally after `createRelatedFromRelation` (which has its own null guard, so the contracts no longer match).
- **Lost guards** in `player_dialogs.cpp`: three sites have `player->clearCardsToDelete()` commented out (`// player->clearCardsToDelete() ||`) — looks like a forgotten TODO. The old code returned early when cards were deleted mid-dialog.
- **Include nit**: `player_dialogs.cpp` line 5 uses `../client/...` (one `..` too few for a 2-level-deep file). It happens to compile but is inconsistent with the rest of the tree (`library_menu.cpp` correctly uses `../../../client/...`).
- **Style nits**: missing trailing newlines on several new/edited files (`player_dialogs.{cpp,h}`, `card_item.cpp`, `custom_zone_menu.cpp`).
- **Architecture nit**: `game/player/player_logic.h` now `#include`s `game_graphics/player/player_area.h`, partially undermining the separation goal. Worth a forward-declare if feasible.

Details inline. Nothing here is a blocker for the refactor approach itself.

RickyRister (approved):

PR #6931, opened 2026-05-22
[Oracle] Change basic lands exception to regex to deal with snow basics.

BruebachL:
Took 45 minutes

## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -474,7 +474,7 @@ FormatRulesNameMap OracleImporter::createDefaultMagicFormats()
     CardCondition superTypeIsBasic;
     superTypeIsBasic.field = "type";
     superTypeIsBasic.matchType = "regex";
-    superTypeIsBasic.value = "^Basic(?:\\s+Snow)?\\s+Land";
+    superTypeIsBasic.value = "\bBasic\b[^—]+\bLand\b";
 
     ExceptionRule basicLands;
     basicLands.conditions.append(superTypeIsBasic);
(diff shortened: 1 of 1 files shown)


oracle/src/oracleimporter.cpp
```diff
@@ -473,8 +473,8 @@ FormatRulesNameMap OracleImporter::createDefaultMagicFormats()
     // Predefined common exceptions
     CardCondition superTypeIsBasic;
     superTypeIsBasic.field = "type";
-    superTypeIsBasic.matchType = "contains";
-    superTypeIsBasic.value = "Basic Land";
+    superTypeIsBasic.matchType = "regex";
+    superTypeIsBasic.value = "^Basic(?:\\s+Snow)?\\s+Land";
```
ebbit1q: this is not flexible enough to my liking, while wizards does not readily add supertypes to the game, it's not unthinkable that we will see other basic lands, I think we should instead regex like this:

```suggestion
    superTypeIsBasic.value = "\bBasic\b[^—]+\bLand\b";
```

ebbit1q: \bBasic\b would also suffice, currently not a single card other than lands have this supertype

ebbit1q: to be completely foolproof it should start with ^[^—]*, but honestly, wizards is absolutely never ever going to print a card with basic as a subtype

ebbit1q (approved):
works as expected

PR #6932, opened 2026-05-22
[Game][Arrows] ArrowItem should not send deletion game commands by itself

BruebachL:
## Related Ticket(s)
- Fixes regression introduced in #6918 and not caught in #6927

ArrowItems sent raw game commands by themselves, bypassing the arrow registry.

- Have arrowItems emit the signal up to the GameScene


ebbit1q (approved):
works, thanks!

PR #6933, opened 2026-05-23
[Protocol] Single-source-of-truth protocol version + publish @cockatrice/protocol

ZeldaZach:
The protocol version is now declared once in libcockatrice_protocol/protocol_version.json. CMake reads it at configure time and emits a generated protocol_version.h exposing COCKATRICE_PROTOCOL_VERSION; remote_client.cpp and serversocketinterface.cpp both pick it up via the existing libcockatrice_protocol link.

The same JSON file is bundled into a new @cockatrice/protocol npm package (scripts/package-protocol.mjs + .github/workflows/protocol-publish.yml) so TypeScript consumers (Sockatrice/webclient) can derive PROTOCOL_VERSION from the identical source instead of hand-typing the literal. The workflow dry-runs npm pack on PRs and publishes to GitHub Packages on stable releases only.


ebbit1q:
cool idea and all, but I'd like to note that we don't actually use this number, the last time it was changed was 2012 b9087715bf2edb215fff547f1a9ab08dac42b011
instead we have the new client and server features list which tells what parts are supported

PR #6936, opened 2026-05-24
[Server][Game][Arrows] Properly notify clients when deleting arrows on card move and transform into

BruebachL:
When I refactored the local client arrow code, I reasoned that the client should not simply delete a local arrow when a card is moved -> It should instead issue a command for arrow deletion to the server and then separately acknowledge this deletion.
However, previously, the contract between client and server was such that, if a card was moved between zones, both silently deleted their own arrows without telling each other, meaning the client never gets a chance to delete the local arrow. If it then tries to by any other means (Cmd_DeleteArrow) or even just with the queued "please delete this arrow now" after the card moved event, the server can no longer find it in its own arrow registry and silently discard the request.

- Have the server emit proper deletion events when it DOES remove arrows, in both CardMoved and CreateToken->TransformInto, where we delete and re-create arrows.

### DIFF
--- a/cockatrice/src/game/game_event_handler.cpp
+++ b/cockatrice/src/game/game_event_handler.cpp
@@ -217,7 +217,20 @@ void GameEventHandler::handleArrowDeletion(int arrowId)
 {
     Command_DeleteArrow cmd;
     cmd.set_arrow_id(arrowId);
-    sendGameCommand(cmd);
+
+    auto preparedCommand = prepareGameCommand(cmd);
+
+    connect(preparedCommand, &PendingCommand::finished, this,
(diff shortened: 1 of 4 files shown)


RickyRister (approved):
Good fix for the future

BruebachL:
Just fyi, I've pushed a new commit which also observes the "not found" response that the server will generate if it fails to find the arrow id the client requested for deletion.

If it does fail to find and generate a "not found" response, the client will simply go ahead and delete the arrow it tried to delete locally (because if the server doesn't know about it, the client shouldn't either.)
I thought this a good defensive fallback and we may as well use the response the server provides to us.

PR #6937, opened 2026-05-24
[Game][Arrows] Hook up to the state zone change properly.

BruebachL:
- this
- and this


RickyRister (approved):
Yep, it's fixed. Thanks!

cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -895,6 +895,11 @@ void TabGame::newCardAdded(AbstractCardItem *card)
     connect(card, &AbstractCardItem::showCardInfoPopup, this, &TabGame::showCardInfoPopup);
     connect(card, SIGNAL(deleteCardInfoPopup(QString)), this, SLOT(deleteCardInfoPopup(QString)));
     connect(card, &AbstractCardItem::cardShiftClicked, this, &TabGame::linkCardToChat);
+    CardItem *cardItem = qobject_cast<CardItem *>(card);
+    if (cardItem) {
+        connect(cardItem->getState(), &CardState::zoneChanged, scene,
```
RickyRister, line 900: Could leave a comment saying why we have to do this (because server versions <= 3.0.1 did not send the delete arrow event properly)

PR #6939, opened 2026-05-26
[DeckEditor] Refactor: clean up addCardHelper

RickyRister:

I'm planning on doing a big refactor of how the card database table view is managed in the deck editor and card database tabs. 

Right now, most code just directly reaches into the `DeckEditorDatabaseDisplayWidget` instance in the deck editor and hooks itself up to the database table view. (The visual database tab is an especially egregious case of this, where it [instantiates an entire deck editor tab](https://github.com/Cockatrice/Cockatrice/blob/b3c89167c53673ade34f376f60769dd5ff6355c8/cockatrice/src/interface/widgets/tabs/tab_visual_database_display.cpp#L7) just so that it can grab a database table view from it.)

I would like to extract out the database table view into its own class so that it can be more cleanly reused.

Anyways, this is the first step. This cleanup will help me later.

- Replace `addCardHelper` with `addCard` and `decrementCard`
  - Made those methods public
  - Make all `actXXX` methods call through the new helpers.

I'm planning on consolidating the zones for the add/decrement signals/slots so that so that we have less wiring that we need to do.

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
@@ -129,14 +129,15 @@ void AbstractTabDeckEditor::onDeckModified()
     emit tabTextChanged(this, getTabText());
 }
 
-/**
- * @brief Helper for adding a card to a deck zone.
- * @param card Card to add.
- * @param zoneName Zone to add the card to.
- */
-void AbstractTabDeckEditor::addCardHelper(const ExactCard &card, const QString &zoneName)
+void AbstractTabDeckEditor::addCard(const ExactCard &card, const QString &zoneName)
 {
     deckStateManager->addCard(card, zoneName);
+    deckMenu->setSaveStatus(true);
+}
+
+void AbstractTabDeckEditor::decrementCard(const ExactCard &card, const QString &zoneName)
+{
+    deckStateManager->decrementCard(card, zoneName);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
```diff
@@ -129,14 +129,15 @@ void AbstractTabDeckEditor::onDeckModified()
...
-/**
- * @brief Helper for adding a card to a deck zone.
- * @param card Card to add.
- * @param zoneName Zone to add the card to.
- */
-void AbstractTabDeckEditor::addCardHelper(const ExactCard &card, const QString &zoneName)
+void AbstractTabDeckEditor::addCard(const ExactCard &card, const QString &zoneName)
 {
     deckStateManager->addCard(card, zoneName);
+    deckMenu->setSaveStatus(true);
```
ZeldaZach: Redundant given the signal chain: `deckStateManager->addCard(...)` (line 134) goes through `DeckStateManager::modifyDeck` → `doCardModified` → `setModified(true)` → emits `isModifiedChanged`, which is connected (in `AbstractTabDeckEditor`'s ctor) to `onDeckModified`, which calls `deckMenu->setSaveStatus(!deckStateManager->isBlankNewDeck())`. That all runs synchronously inside the `deckStateManager->addCard` call, so by the time this line executes the menu's save status is already set.

`decrementCard` (just below) intentionally doesn't have the explicit call and relies on the same chain — dropping this line makes the two methods symmetric and keeps the "save status is owned by the signal handler" invariant clean. Not strictly a bug, just thought I'd flag while you're cleaning this up.

ZeldaZach (commented):
LGTM as a stepping-stone refactor — the `addCard`/`decrementCard` split is a sensible API improvement and the doc-string note about empty `PrintingInfo` matching by name is a nice touch.

One small nit inline: I think the explicit `setSaveStatus(true)` inside `addCard` is leftover from the old per-call-site pattern. The signal chain via `DeckStateManager::isModifiedChanged → AbstractTabDeckEditor::onDeckModified → deckMenu->setSaveStatus(...)` already fires synchronously from `deckStateManager->addCard(...)`, so the explicit call is redundant. Dropping it would also make `addCard` and `decrementCard` visually symmetric. Not a bug — happy to ignore if you'd rather keep it as belt-and-suspenders during the broader refactor.

ZeldaZach (approved):

PR #6940, opened 2026-05-26
[CardDatabaseModel] Pass CardInfoPtr by const ref

RickyRister:

Some of the methods in `CardDatabaseModel` don't pass `CardInfoPtr` by const ref even though they could. `CardInfoPtr` are shared pointers, so this means it's doing an unnecessary reference count increment on every call.

Usually, this sort of micro-optimization wouldn't matter, but the aforementioned methods *are* being called inside a loop that iterates over the entire card database, so there might actually be some visible effects here as the size of the card database grows.

- Make methods pass `CardInfoPtr` by const ref

Before
Screenshot 2026-05-26 at 12 57 24 AM

After
Screenshot 2026-05-26 at 12 57 35 AM

### DIFF
--- a/libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
+++ b/libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
@@ -81,7 +81,7 @@ QVariant CardDatabaseModel::headerData(int section, Qt::Orientation orientation,
     }
 }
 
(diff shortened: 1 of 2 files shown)


ebbit1q (approved):
seems small but measurable, I'd say a larger samplesize would be required to make a real conclusion but there is no reason not to pass these const so

PR #6941, opened 2026-05-26
[Game] Properly split (Abstract)CardItem into (Abstract)CardState, separates logic/graphic

BruebachL:
## Related Ticket(s)
- Dependent on #6919 
- Dependent on #6928 
 
#6904 only defined the new CardState class itself (but not AbstractCardState) and forwarded accessors to wrappers around this new class. 

- This refactor instead uses the new CardState class directly in consumer classes.

### DIFF
--- a/cockatrice/src/game_graphics/player/menu/player_menu.cpp
+++ b/cockatrice/src/game_graphics/player/menu/player_menu.cpp
@@ -76,6 +76,10 @@ QMenu *PlayerMenu::updateCardMenu(const CardState *card)
         return nullptr;
     }
 
+    if (!player->getLogic()->getGame()->getActiveCard()) {
+        return nullptr;
+    }
+
     // If is spectator (as spectators don't need card menus), return
     // only update the menu if the card is actually selected
     if ((player->getLogic()->getGame()->getPlayerManager()->isSpectator() &&
--- a/cockatrice/src/game_graphics/player/menu/player_menu.cpp
+++ b/cockatrice/src/game_graphics/player/menu/player_menu.cpp
@@ -76,15 +76,12 @@ QMenu *PlayerMenu::updateCardMenu(const CardState *card)
         return nullptr;
     }
 
-    if (!player->getLogic()->getGame()->getActiveCard()) {
-        return nullptr;
-    }
-
     // If is spectator (as spectators don't need card menus), return
     // only update the menu if the card is actually selected
+    auto *activeCard = player->getLogic()->getGame()->getActiveCard();
     if ((player->getLogic()->getGame()->getPlayerManager()->isSpectator() &&
          !player->getLogic()->getGame()->getPlayerManager()->isJudge()) ||
-        player->getLogic()->getGame()->getActiveCard()->getState() != card) {
+        !activeCard || activeCard->getState() != card) {
         return nullptr;
     }
 
--- a/cockatrice/src/game_graphics/player/menu/player_menu.cpp
+++ b/cockatrice/src/game_graphics/player/menu/player_menu.cpp
@@ -69,7 +69,7 @@ void PlayerMenu::setMenusForGraphicItems()
     }
 }
 
-QMenu *PlayerMenu::updateCardMenu(const CardItem *card)
+QMenu *PlayerMenu::updateCardMenu(const CardState *card)
 {
     if (!card) {
         emit cardMenuUpdated(nullptr);
@@ -80,7 +80,7 @@ QMenu *PlayerMenu::updateCardMenu(const CardItem *card)
     // only update the menu if the card is actually selected
     if ((player->getLogic()->getGame()->getPlayerManager()->isSpectator() &&
          !player->getLogic()->getGame()->getPlayerManager()->isJudge()) ||
-        player->getLogic()->getGame()->getActiveCard() != card) {
+        player->getLogic()->getGame()->getActiveCard()->getState() != card) {
         return nullptr;
     }
 
--- a/cockatrice/src/game_graphics/player/menu/utility_menu.cpp
+++ b/cockatrice/src/game_graphics/player/menu/utility_menu.cpp
@@ -38,7 +38,7 @@ UtilityMenu::UtilityMenu(PlayerGraphicsItem *_player, QMenu *playerMenu) : QMenu
         aIncrementAllCardCounters = new QAction(this);
         connect(aIncrementAllCardCounters, &QAction::triggered, playerActions, [this]() {
             player->getLogic()->getPlayerActions()->actIncrementAllCardCounters(
-                player->getGameScene()->selectedCards());
+                player->getGameScene()->selectedCardsAsStates());
         });
 
         createPredefinedTokenMenu = new QMenu(QString());
--- a/cockatrice/src/game_graphics/player/player_dialogs.cpp
+++ b/cockatrice/src/game_graphics/player/player_dialogs.cpp
@@ -241,14 +241,14 @@ void PlayerDialogs::onMoveCardXCardsFromTopDialogRequested(int defaultNumberTopC
     number -= 1; // indexes start at 0
 
     if (ok) {
-        playerActions->actMoveCardXCardsFromTop(player->getGameScene()->selectedCards(), number);
+        playerActions->actMoveCardXCardsFromTop(player->getGameScene()->selectedCardsAsStates(), number);
     }
 }
 
 void PlayerDialogs::onSetPTDialogRequested(const QString &oldPT)
 {
     bool ok;
-    auto cards = player->getGameScene()->selectedCards();
+    auto cards = player->getGameScene()->selectedCardsAsStates();
     emit requestDialogSemaphore(true);
     QString pt = getTextWithMax(dialogParent(), tr("Change power/toughness"), tr("Change stats to:"), QLineEdit::Normal,
                                 oldPT, &ok);
@@ -263,7 +263,7 @@ void PlayerDialogs::onSetPTDialogRequested(const QString &oldPT)
 
 void PlayerDialogs::onSetAnnotationDialogRequested(const QString &oldAnnotation)
 {
-    auto cards = player->getGameScene()->selectedCards();
+    auto cards = player->getGameScene()->selectedCardsAsStates();
     emit requestDialogSemaphore(true);
     AnnotationDialog *dialog = new AnnotationDialog(dialogParent());
     dialog->setOptions(QInputDialog::UsePlainTextEditForTextInput);
@@ -281,7 +281,7 @@ void PlayerDialogs::onSetAnnotationDialogRequested(const QString &oldAnnotation)
 
 void PlayerDialogs::onSetCardCounterDialogRequested(int counterId, const QString &oldValueForDlg)
 {
-    auto cards = player->getGameScene()->selectedCards();
+    auto cards = player->getGameScene()->selectedCardsAsStates();
     emit requestDialogSemaphore(true);
 
     auto &cardCounterSettings = SettingsCache::instance().cardCounters();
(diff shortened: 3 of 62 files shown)


cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+void PlayerDialogs::onShuffleBottomDialogRequested(int defaultNumberBottomCards, int maxCards)
+{
+    bool ok;
+    int number = QInputDialog::getInt(dialogParent, tr("Shuffle bottom cards of library"),
+                                      tr("Number of cards: (max. %1)").arg(maxCards), defaultNumberBottomCards, 1,
+                                      maxCards, 1, &ok);
+    if (!ok) {
```
ZeldaZach: Inverted condition — every other dialog in this file guards with `if (ok)`. Here `if (!ok)` means: do nothing when the user confirms, shuffle when they cancel. PR #6928 had this as `if (ok) { playerActions->actShuffleBottom(number); }`; the negation was added in this PR.

Fix: drop the `!`.

ZeldaZach (commented):
Solid follow-on to #6928 — the direct use of `CardState` in consumers is a clean improvement. I focused this review on regressions vs. #6928 (since that's the diff base in effect) and found a cluster of behavior bugs in `PlayerDialogs` plus a couple in the menu/zone wiring. The first three are clear copy-paste / forgotten-line bugs that completely break the affected actions; the others are likelier crash/leak sources.

**Summary:**
- **`onShuffleBottomDialogRequested`** has an inverted `ok` check — shuffles on cancel, does nothing on confirm.
- **`onMoveTopCardsUntilDialogRequested`** passes the input `options` instead of `dlg.getOptions()`, so user edits to the dialog are discarded.
- **`onSetAnnotationDialogRequested`** builds the annotation string but never calls `actSetAnnotation` — annotation is silently dropped.
- **`CardZone::onCardAdded`** now creates a fresh `CardItem` on every `cardAdded` signal with no removal path; the old code received the existing item from the logic layer. This will leak items / leave orphans in the scene when cards move zones.
- **`PlayerMenu::updateCardMenu`** dereferences `getActiveCard()->getState()` with no null guard. The pre-PR code did pointer compare against the `CardItem*` directly and was safe; the new `CardState*` parameter forces a `->getState()` call that needs a null check.
- **`UtilityMenu::setLastToken`** has an inverted guard vs. #6928 — currently bails out exactly when the “create another” action exists (the only time we should proceed) and forwards when it doesn’t.

CI is red on Arch, Fedora 44, and macOS 15 Debug — worth checking whether those overlap with anything below. Details inline.

cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+void PlayerDialogs::onMoveTopCardsUntilDialogRequested(MoveTopCardsUntilOptions options)
+{
+    DlgMoveTopCardsUntil dlg(dialogParent, options);
+    if (!dlg.exec()) {
+        return;
+    }
+    playerActions->moveTopCardsUntil(dlg.getExpr(), options);
```
ZeldaZach: `options` is the value passed *into* the dialog, not what the user produced. The user's edits in `DlgMoveTopCardsUntil` are read back via `dlg.getOptions()` — passing the original `options` here silently discards them.

PR #6928 had `playerActions->moveTopCardsUntil(dlg.getExpr(), dlg.getOptions());`. Fix: restore `dlg.getOptions()`.

cockatrice/src/game_graphics/player/player_dialogs.cpp
```diff
@@ -0,0 +1,288 @@
...
+    dialog->setOptions(QInputDialog::UsePlainTextEditForTextInput);
+    dialog->setWindowTitle(tr("Set annotation"));
+    dialog->setLabelText(tr("Please enter the new annotation:"));
+    dialog->setTextValue(oldAnnotation);
+    bool ok = dialog->exec();
+    emit requestDialogSemaphore(false);
+    if (!ok) { // player->clearCardsToDelete() ||
+        return;
+    }
+    QString annotation = dialog->textValue().left(MAX_NAME_LENGTH);
```
ZeldaZach, line 278: The function builds the annotation string and then exits — `actSetAnnotation` is never called, so the annotation never gets applied to any card. PR #6928 had the call as the next line:

```cpp
playerActions->actSetAnnotation(player->getGameScene()->selectedCards(), annotation);
```

Looks like it was dropped accidentally during the `selectedCards` → `selectedCardsAsStates` rename. Fix: re-add the call (with whatever the new selection-as-states accessor is).

cockatrice/src/game_graphics/zones/card_zone.cpp
```diff
@@ -16,17 +16,19 @@ CardZone::CardZone(CardZoneLogic *_logic, QGraphicsItem *parent)
...
 }
 
-void CardZone::onCardAdded(CardItem *addedCard)
+void CardZone::onCardAdded(CardState *toAdd, int /*x*/, int /*y*/)
 {
+    CardItem *addedCard = new CardItem(toAdd, this);
     addedCard->setParentItem(this);
     addedCard->setVisible(true);
     addedCard->update();
+    cards.append(addedCard);
```
ZeldaZach, line 26: Lifecycle issue: this now `new`s a `CardItem` on every `cardAdded` signal and appends it to a local `cards` list, but there's no corresponding `cardRemoved` connection or removal path. In #6928 and master, `onCardAdded(CardItem *addedCard)` reparented an item the logic layer already owned, and the logic layer's `getCards()` was the source of truth.

With the new ownership model, when a card moves zones (`eventMoveCard` → `targetZone->addCard(...)` → emits `cardAdded`), a fresh `CardItem` wrapping the same `CardState` gets created on the target zone, but the old one in the source zone's `cards` list is never removed or deleted. Net effect: leaked items in the scene and stale entries in `cards`.

Fix: connect `CardZoneLogic`'s remove/take signal to a `CardZone::onCardRemoved` slot that finds the matching wrapper in `cards`, removes it, and `deleteLater()`s it (or at least `delete`s after detaching from the scene).

cockatrice/src/game_graphics/player/menu/player_menu.cpp
```diff
@@ -74,20 +78,22 @@ QMenu *PlayerMenu::updateCardMenu(const CardItem *card)
 
     // If is spectator (as spectators don't need card menus), return
     // only update the menu if the card is actually selected
-    if ((player->getGame()->getPlayerManager()->isSpectator() && !player->getGame()->getPlayerManager()->isJudge()) ||
-        player->getGame()->getActiveCard() != card) {
+    if ((player->getLogic()->getGame()->getPlayerManager()->isSpectator() &&
+         !player->getLogic()->getGame()->getPlayerManager()->isJudge()) ||
+        player->getLogic()->getGame()->getActiveCard()->getState() != card) {
```
ZeldaZach: `getActiveCard()` can legitimately return `nullptr` (it's set to nullptr in several places — e.g. when a card is deselected or destroyed). Calling `->getState()` on it crashes.

PR #6928 used a direct pointer compare (`getActiveCard() != card`) which was null-safe. The refactor to take `CardState*` forces a `->getState()` step that needs a guard:

```cpp
auto *activeCard = player->getLogic()->getGame()->getActiveCard();
if ((player->getLogic()->getGame()->getPlayerManager()->isSpectator() &&
     !player->getLogic()->getGame()->getPlayerManager()->isJudge()) ||
    !activeCard || activeCard->getState() != card) {
    return nullptr;
}
```

cockatrice/src/game_graphics/player/menu/utility_menu.cpp
```diff
@@ -84,14 +92,24 @@ void UtilityMenu::populatePredefinedTokensMenu()
...
-            connect(a, &QAction::triggered, player->getPlayerActions(), &PlayerActions::actCreatePredefinedToken);
+            connect(a, &QAction::triggered, player->getLogic()->getPlayerActions(),
+                    &PlayerActions::actCreatePredefinedToken);
         }
     }
 }
 
+void UtilityMenu::setLastToken(CardInfoPtr lastToken)
+{
+    if (createAnotherTokenActionExists()) {
```
ZeldaZach: Inverted vs. #6928 — that version had `if (!createAnotherTokenActionExists()) { return; }`, which matched the original `PlayerActions::setLastToken` semantics: only update “last token” *after* a token has already been created (when the action exists).

The current form bails out exactly when the action exists, and only proceeds when there's nothing to update — so the “Create another X token” menu entry will never get its name refreshed on subsequent creations.

Fix: restore the `!`.

BruebachL:
This is a bitch. Current problem is the fact that CardItems have no clear entry point for being added to a player e.g.
image
all these instantiations only tie back to the player via the constructor:
image

whereas addCard performs no internal bookkeeping but simply emits a signal:
image

which ultimately hooks up to TabGame to perform UI things:
image

If we're being honest on a refactor, this immediately smells:
image

The logic shouldn't create an Item. It should hold States and emit signals and let the corresponding ZoneViews take care of the item.

Matter of fact, the zone logic holding pointers to graphics objects is anti-design anyway:
image

The cardAdded signal from player hooks up to some other ui things and they all expect a CardItem (because they do UI things on it like right click, etc). So, we'll have to have an equivalent signal in PlayerGraphicsItem like cardItemAdded
however, we can't just forward the PlayerLogic::cardAdded(CardState*) signal to PlayerGraphicsLogic::cardItemAdded(CardItem *) because we're no longer aware of the corresponding graphics item at creation time of the logic item.

>:(

ZeldaZach:
Your instincts here are right and the design direction you're sketching is the one I'd take — logic owns `CardState`, graphics owns `CardItem`, and the zone view is what bridges them. The piece I think is missing from your write-up is that the bridge doesn't have to be a single forward — it can be a *transform-then-re-emit*.

Concretely:

1. `CardZoneLogic` emits `cardStateAdded(CardState*)` and `cardStateRemoved(CardState*)` — pure logic, no graphics awareness, no leaked pointers across the boundary.
2. `CardZone` (the graphics view) is the *only* thing connected to those signals. Its slot creates the wrapping `CardItem`, stores it in the zone's `cards` list, parents it into the scene, and then emits its own `cardItemAdded(CardItem*)`. The reverse on removal: look up the wrapper for the `CardState*`, take it out of the scene, delete (or `deleteLater`) it, then emit `cardItemRemoved(CardItem*)`.
3. `PlayerGraphicsItem` is connected to *each of its zones'* `cardItemAdded`/`cardItemRemoved` signals — not to `PlayerLogic` at all for this. It just re-emits them as its own player-level `cardItemAdded(CardItem*)` for the higher UI consumers (TabGame, context menus, etc.) that need to operate on items.

That sidesteps the "`PlayerLogic::cardAdded(CardState*)` → `PlayerGraphicsItem::cardItemAdded(CardItem*)` can't forward directly" problem because `PlayerGraphicsItem` is never the thing creating the `CardItem`. The zone is, because the zone is the only layer that holds both halves. No registry, no `CardState*` → `CardItem*` map at the player level — the lookup lives inside `CardZone` where it has to anyway, to handle the remove case (which is the bug I flagged on `card_zone.cpp:25` — that one resolves naturally once `cardStateRemoved` exists).

This also makes the "zone logic holds graphics pointers" smell go away, because it no longer does — `CardZoneLogic::getCards()` returns `QList<CardState*>` and that's it. `CardZone` (graphics) maintains the parallel `QList<CardItem*>` and is the only place that knows how the two correspond.

The one gotcha worth thinking about up front: ordering during a cross-zone move. `eventMoveCard` currently does source-take then target-add as two separate operations, and the events fire in order. If a UI consumer downstream (e.g. animation code) needs to know "this is the same logical card moving," it'll want `CardState*` identity to be preserved across the move — which it is, since logic owns the state — but the *wrapper* `CardItem*` changes. Worth deciding whether "identity for animation purposes" lives on `CardState*` or whether you want the zone to *reparent* an existing wrapper rather than destroy+recreate. The simpler thing is destroy+recreate (matches what already happens in master with the multi-step move dance), but it's worth being explicit in the design doc / commit message so future-you doesn't have to re-derive it.

Not saying this PR has to solve all of that — happy to see the lifecycle bug fixed and the rest deferred to a follow-up. But I think the bones of the right design are already in what you've written.

PR #6942, opened 2026-05-26
Run formatter on all our files

RickyRister:
## Related Ticket(s)
- Follow-up to #6887
- Follow-up to #6896

Seems like the formatter didn't get run on all our files after the changes to the formatter in the above PRs.

I only found out once CI failed because I touched a file that didn't get reformatted.

- Run the formatter on all our files (for real this time)


ebbit1q (approved):
thanks

PR #6944, opened 2026-05-26
[Game][Player] Split Player into PlayerLogic/PlayerGraphicsItem

BruebachL:
## Related Ticket(s)
- Succeeds #6919

This PR is the first structural step in a larger refactor to decouple the game's UI layer from its logic layer, enabling a future pub/sub event system. The core problem was that Player (now PlayerLogic) directly owned and created PlayerGraphicsItem, PlayerMenu, and all associated graphics classes, making the logic layer tightly coupled to Qt's widget/graphics stack.

- PlayerLogic no longer owns graphics
- PlayerGraphicsItem and PlayerMenu are no longer constructed inside PlayerLogic. The constructor no longer creates graphics objects. Logic signals (arrowCreateRequested, arrowDeleteRequested, arrowsCleared, counterAdded, counterRemoved, concededChanged, zoneIdChanged, requestZoneViewToggle, requestSortHand) replace direct calls into the graphics layer.
- GameScene owns PlayerGraphicsItem
- GameScene::addPlayer(PlayerLogic*) now creates PlayerGraphicsItem internally and registers it in a QMap<int, PlayerGraphicsItem*> playerViews for ID-based lookup. GameScene::viewForPlayer(int) is the new access point for anything that needs the view for a given player.

### DIFF
--- a/cockatrice/src/game/board/card_item.cpp
+++ b/cockatrice/src/game/board/card_item.cpp
@@ -40,7 +40,7 @@ void CardItem::prepareDelete()
 {
     if (owner != nullptr) {
         if (owner->getGame()->getActiveCard() == this) {
-            owner->getPlayerMenu()->updateCardMenu(nullptr);
+            emit owner->requestCardMenuUpdate(nullptr);
             owner->getGame()->setActiveCard(nullptr);
         }
         owner = nullptr;
@@ -399,8 +399,11 @@ void CardItem::playCard(bool faceDown)
         emit tz->toggleTapped();
     } else {
         if (SettingsCache::instance().getClickPlaysAllSelected()) {
-            faceDown ? state->getZone()->getPlayer()->getPlayerActions()->actPlayFacedown()
-                     : state->getZone()->getPlayer()->getPlayerActions()->actPlay();
+            if (faceDown) {
+                emit playSelectedFaceDown(this);
+            } else {
+                emit playSelected(this);
+            }
         } else {
             state->getZone()->getPlayer()->getPlayerActions()->playCard(this, faceDown);
         }
@@ -460,7 +463,7 @@ void CardItem::handleClickedToPlay(bool shiftHeld)
 {
     if (isUnwritableRevealZone(state->getZone())) {
         if (SettingsCache::instance().getClickPlaysAllSelected()) {
-            state->getZone()->getPlayer()->getPlayerActions()->actHide();
+            emit hideSelected(this);
         } else {
             state->getZone()->removeCard(this);
         }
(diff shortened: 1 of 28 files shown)


cockatrice/src/game/board/card_item.cpp
```diff
@@ -471,17 +474,12 @@ void CardItem::handleClickedToPlay(bool shiftHeld)
...
-            owner->getGame()->setActiveCard(this);
-            if (QMenu *cardMenu = owner->getPlayerMenu()->updateCardMenu(this)) {
-                cardMenu->popup(event->screenPos());
-                return;
-            }
-        }
-    } else if ((event->modifiers() != Qt::AltModifier) && (event->button() == Qt::LeftButton) &&
-               (!SettingsCache::instance().getDoubleClickToPlay())) {
+    if (event->button() == Qt::RightButton && owner != nullptr) {
+        emit rightClicked(this, event->screenPos());
```
ZeldaZach, line 478: Subtle behavior change: the early `return` here is now unconditional on `RightButton + owner`. The pre-PR code only returned when `updateCardMenu` returned a non-null `QMenu *`; when it returned `nullptr` (e.g., spectator right-clicking a card whose owner isn't them, or `getActiveCard()` mismatch), execution fell through to:

```cpp
if (owner != nullptr) {
    setCursor(Qt::OpenHandCursor);
}
AbstractCardItem::mouseReleaseEvent(event);
```

With the new code that path is skipped for *every* right-click on an owned card — the cursor doesn't get reset and the base class never sees the event. If `AbstractCardItem::mouseReleaseEvent` does anything important on right-button release (selection state, hover bookkeeping, etc.), that's lost.

Likely fix: drop the `return` here and let the fall-through still run; `GameScene::onCardRightClicked` doesn't need it. (Apologies — earlier body of this comment was an accidental "test" while I was debugging an API issue.)

ZeldaZach (commented):
Foundational refactor reads cleanly — the signal-based decoupling of `PlayerActions`/`PlayerLogic` from `PlayerGraphicsItem`/`PlayerMenu`, and the move of `PlayerGraphicsItem` ownership to `GameScene`, are well-executed. CI is green and most of the diff is mechanical signal/connect rewiring that holds up.

Two things worth a second look, both small:

- **Right-click behavior change in `CardItem::mouseReleaseEvent`** — early-return is now unconditional on `RightButton + owner`, but the original only returned if `updateCardMenu` returned a non-null menu. The spectator / no-active-card path loses its `setCursor` + base-class call as a side effect. (Inline below.)
- **`GameScene::removePlayer` doesn't delete the view** it just took out of `playerViews`. `addPlayer` now creates `PlayerGraphicsItem` with `new ... (player)` and no QObject parent, so scene ownership is all that's keeping it alive. `removeItem(view)` only detaches — nothing frees it. Same as master, but since this PR is the one shifting `PlayerGraphicsItem` ownership into the scene, it's the natural moment to make cleanup explicit (`view->deleteLater();` or pass a QObject parent at construction). I couldn't anchor this inline because `removePlayer` itself wasn't touched by the diff; flagging here.

(Apologies for the accidental "test1" body that this comment is replacing — I was debugging a 422 from the GitHub API and didn't realize the test review had already submitted.)

ZeldaZach (approved):
No notes

PR #6945, opened 2026-05-26
[Game][Menus] Make Menus accept PlayerGraphicsItem instead of PlayerLogic

BruebachL:
## Related Ticket(s)
- Succeeds #6919
- Dependent on #6944

These are graphics classes and some of them need the player graphics. It's sensible to have them all take the graphic item, since it has access to the logic, especially for future refactors.

- All menu classes take PlayerGraphicsItem* instead of PlayerLogic* (* This might not be entirely accurate in this PR but it will be)
CardMenu, CustomZoneMenu, GraveyardMenu, HandMenu, LibraryMenu, MoveMenu, PtMenu, RfgMenu, SayMenu, SideboardMenu, UtilityMenu, and PlayerMenu all now hold PlayerGraphicsItem* and reach logic through player->getPlayerLogic(). These are view-layer classes and this reflects their actual nature.

### DIFF
--- a/cockatrice/src/game/player/menu/library_menu.cpp
+++ b/cockatrice/src/game/player/menu/library_menu.cpp
@@ -8,9 +8,10 @@
 #include "../player_logic.h"
 
 #include <QAction>
+#include <QGraphicsView>
 #include <QMenu>
 
-LibraryMenu::LibraryMenu(PlayerLogic *_player, QWidget *parent) : TearOffMenu(parent), player(_player)
+LibraryMenu::LibraryMenu(PlayerGraphicsItem *_player, QWidget *parent) : TearOffMenu(parent), player(_player)
 {
     createDrawActions();
     createShuffleActions();
@@ -75,8 +76,8 @@ LibraryMenu::LibraryMenu(PlayerLogic *_player, QWidget *parent) : TearOffMenu(pa
     bottomLibraryMenu->addSeparator();
     bottomLibraryMenu->addAction(aShuffleBottomCards);
 
-    connect(player, &PlayerLogic::resetTopCardMenuActions, this, &LibraryMenu::resetTopCardMenuActions);
-    connect(player, &PlayerLogic::deckChanged, this, &LibraryMenu::enableOpenInDeckEditorAction);
+    connect(player->getLogic(), &PlayerLogic::resetTopCardMenuActions, this, &LibraryMenu::resetTopCardMenuActions);
+    connect(player->getLogic(), &PlayerLogic::deckChanged, this, &LibraryMenu::enableOpenInDeckEditorAction);
 
     retranslateUi();
 }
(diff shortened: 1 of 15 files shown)


cockatrice/src/game/player/menu/library_menu.cpp
```diff
@@ -313,27 +314,33 @@ void LibraryMenu::populateRevealTopCardMenuWithActivePlayers()
...
     if (auto *a = qobject_cast<QAction *>(sender())) {
-        player->getPlayerActions()->actLendLibrary(a->data().toInt());
+        player->getLogic()->getPlayerActions()->actLendLibrary(a->data().toInt());
     }
 }
 
 void LibraryMenu::onRevealTopCardTriggered()
 {
+    QWidget *parent = nullptr;
+    if (auto *view = player->scene() ? player->scene()->views().value(0) : nullptr) {
```
ZeldaZach, line 331: This is the only dialog in the menus that derives its parent from `player->scene()->views().value(0)->window()`. Everywhere else (e.g. `actViewTopCards`, `actMoveTopCardsToGrave`, `actSetAnnotation`, `actCreateToken` in `player_actions.cpp`) still uses `player->getGame()->getTab()`. Half-migrated state is a little odd — if someone moves the rest later, they'll need to remember this one is already done; if they don't, this stays the only inconsistent dialog forever.

If the intent is to migrate everything to scene-based parent inference (which would match the `PlayerDialogs` approach in #6928), worth either doing it as one pass or leaving this one on `getTab()` for now and converting in the same commit as the rest. Either is fine — just want to avoid the in-between state landing on master.

Minor: `bool ok = true;` at line 333 is a no-op init since `QInputDialog::getInt` always writes `ok` before returning. Not worth a follow-up commit but worth dropping if you touch this again.

ZeldaZach (commented):
LGTM mechanically — the `PlayerLogic*` → `PlayerGraphicsItem*` parameter migration across the menu classes is clean and the `player->getLogic()->X` rewrites are consistent. CI is green. Whole PR is ~640 lines of mostly type-rewiring on top of #6944.

One real finding inline: `LibraryMenu::onRevealTopCardTriggered` is the only call site in this PR that adopts a new dialog-parent strategy (`player->scene()->views().value(0)->window()`). Everywhere else still uses `player->getGame()->getTab()`. Worth a follow-up to make this consistent — either roll out the scene-based pattern everywhere, or revert this one to keep the diff purely mechanical until the broader migration lands. (Looks like the bigger refactor to a `PlayerDialogs` class happens in #6928 and uses this same scene-based pattern, so this is the first instance of that approach.)

cockatrice/src/game/player/menu/library_menu.cpp
BruebachL: I'll remember.

ZeldaZach (approved):
No notes

PR #6946, opened 2026-05-26
[Game][Player] Move dialog creation out of player_actions and into player_dialogs

BruebachL:
## Related Ticket(s)
- Succeeds #6919 
- Dependent on #6944 
- Dependent on #6945 

PlayerActions is a logic class. It defines the "send commands" side of the Player transactions, complimented by the "receive events" of PlayerEventHandler.

- PlayerDialogs introduced
- Dialog handling is extracted out of PlayerActions into a new PlayerDialogs class owned by PlayerGraphicsItem. PlayerActions emits requestDialog signals; PlayerDialogs connects to them and owns the QWidget dialogParent needed to show dialogs.

## Patterns

We essentially had two patterns in PlayerActions:

```
// Pattern 1 — pure command builder. Takes primitives, builds protobuf, sends.
void PlayerActions::actDrawCards(int number) {
    Command_DrawCards cmd;
    cmd.set_number(number);
    sendGameCommand(cmd);
}

// Pattern 2 — pulls UI state itself instead of receiving it.
void PlayerActions::actIncPT(int deltaP, int deltaT) {
    for (auto card : player->getGameScene()->selectedCards()) { // ← reaches into scene
        // ... builds command from card data
    }
}
```

Pattern 2 is what this PR fixes. Pattern 1 is the correct and intended usage. 

## What PlayerActions actually is
It's a command builder. Given a user's intention and the current game state, it constructs the correct protobuf message and sends it through GameEventHandler. That's it. Everything that isn't command building is either:

- UI translation: figuring out which cards the user means (selected cards, active card), which belongs in the caller
- Dialog orchestration: asking the user for parameters (how many cards?), which belongs in PlayerDialogs

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -988,8 +988,11 @@ void PlayerActions::actCreateAllRelatedCards()
 
     if (relatedCards.length() == 1) {
         cardRelation = relatedCards.at(0);
+        lastRelatedCreationSucceeded = false; // reset before emit
         actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
-        ++tokensTypesCreated;
+        if (lastRelatedCreationSucceeded) {
+            ++tokensTypesCreated;
+        }
     } else {
         QList<CardRelation *> nonExcludedRelatedCards;
         QString dbName;
@@ -999,15 +1002,18 @@ void PlayerActions::actCreateAllRelatedCards()
             }
         }
         switch (nonExcludedRelatedCards.length()) {
-            case 1: // if nonExcludedRelatedCards == 1
+            case 1:
                 cardRelation = nonExcludedRelatedCards.at(0);
+                lastRelatedCreationSucceeded = false; // reset before emit
                 actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
-                ++tokensTypesCreated;
-
+                if (lastRelatedCreationSucceeded) {
+                    ++tokensTypesCreated;
+                }
                 break;
+
             // If all are marked "Exclude", then treat the situation as if none of them are.
             // We won't accept "garbage in, garbage out", here.
-            case 0: // else if nonExcludedRelatedCards == 0
+            case 0:
                 for (CardRelation *cardRelationAll : relatedCards) {
                     if (!cardRelationAll->getDoesAttach() && !cardRelationAll->getIsVariable()) {
                         dbName = cardRelationAll->getName();
@@ -1022,7 +1028,8 @@ void PlayerActions::actCreateAllRelatedCards()
                     }
                 }
                 break;
-            default: // else
+
+            default:
                 for (CardRelation *cardRelationNotExcluded : nonExcludedRelatedCards) {
                     if (!cardRelationNotExcluded->getDoesAttach() && !cardRelationNotExcluded->getIsVariable()) {
                         dbName = cardRelationNotExcluded->getName();
@@ -1056,33 +1063,36 @@ void PlayerActions::actRequestCreateRelatedFromRelationDialog(const CardItem *so
     emit requestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
 }
 
-void PlayerActions::createRelatedFromRelation(const CardItem *sourceCard,
+bool PlayerActions::createRelatedFromRelation(const CardItem *sourceCard,
                                               const CardRelation *cardRelation,
                                               int variableCount)
 {
     if (sourceCard == nullptr || cardRelation == nullptr) {
-        return;
+        return false;
     }
 
     const QString dbName = cardRelation->getName();
     const bool persistent = cardRelation->getIsPersistent();
 
-    int count = 1;
-
+    // Variable relations always use DoesNotAttach, regardless of the count the user
+    // entered.
     if (cardRelation->getIsVariable()) {
-        count = variableCount;
-        if (count <= 0) {
-            return;
+        if (variableCount <= 0) {
+            return false;
         }
-    } else {
-        count = cardRelation->getDefaultCount();
+        for (int i = 0; i < variableCount; ++i) {
+            createCard(sourceCard, dbName, CardRelationType::DoesNotAttach, persistent);
+        }
+        return true;
     }
+
+    const int count = cardRelation->getDefaultCount();
 
     if (count > 1) {
         for (int i = 0; i < count; ++i) {
             createCard(sourceCard, dbName, CardRelationType::DoesNotAttach, persistent);
         }
-        return;
+        return true;
     }
 
     CardRelationType attachType;
@@ -1101,11 +1111,15 @@ void PlayerActions::createRelatedFromRelation(const CardItem *sourceCard,
     }
 
     createCard(sourceCard, dbName, attachType, persistent);
-    return;
+    return true;
 }
 
 void PlayerActions::onRelatedCardCreated(const CardItem *sourceCard, const CardRelation *cardRelation)
 {
+    if (sourceCard == nullptr || cardRelation == nullptr) {
+        return;
+    }
+
     /*
      * If we make a token via "Token: TokenName"
      * then let's allow it to be created via "create another token"
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -5,7 +5,6 @@
 #include "../../interface/widgets/tabs/tab_game.h"
 #include "../../interface/widgets/utility/get_text_with_max.h"
 #include "../board/card_item.h"
-#include "../client/settings/card_counter_settings.h"
 #include "../dialogs/dlg_move_top_cards_until.h"
 #include "../dialogs/dlg_roll_dice.h"
 #include "../zones/view_zone_logic.h"
@@ -175,30 +174,26 @@ void PlayerActions::actSortHand()
     emit requestSortHand(sortOptions + defaultOptions);
 }
 
-void PlayerActions::actViewTopCards()
+void PlayerActions::actRequestViewTopCardsDialog()
 {
-    int deckSize = player->getDeckZone()->getCards().size();
-    bool ok;
-    int number = QInputDialog::getInt(player->getGame()->getTab(), tr("View top cards of library"),
-                                      tr("Number of cards: (max. %1)").arg(deckSize), defaultNumberTopCards, 1,
-                                      deckSize, 1, &ok);
-    if (ok) {
-        defaultNumberTopCards = number;
-        emit requestZoneViewToggle(ZoneNames::DECK, number);
-    }
+    emit requestViewTopCardsDialog(defaultNumberTopCards, player->getDeckZone()->getCards().size());
 }
 
-void PlayerActions::actViewBottomCards()
+void PlayerActions::actViewTopCards(int number)
 {
-    int deckSize = player->getDeckZone()->getCards().size();
-    bool ok;
-    int number = QInputDialog::getInt(player->getGame()->getTab(), tr("View bottom cards of library"),
-                                      tr("Number of cards: (max. %1)").arg(deckSize), defaultNumberBottomCards, 1,
-                                      deckSize, 1, &ok);
-    if (ok) {
-        defaultNumberBottomCards = number;
-        emit requestZoneViewToggle(ZoneNames::DECK, number, true);
-    }
+    defaultNumberTopCards = number;
+    emit requestZoneViewToggle(ZoneNames::DECK, number);
+}
+
+void PlayerActions::actRequestViewBottomCardsDialog()
+{
+    emit requestViewBottomCardsDialog(defaultNumberBottomCards, player->getDeckZone()->getCards().size());
+}
+
+void PlayerActions::actViewBottomCards(int number)
+{
+    defaultNumberBottomCards = number;
+    emit requestZoneViewToggle(ZoneNames::DECK, number, true);
 }
 
 void PlayerActions::actAlwaysRevealTopCard(bool alwaysRevealTopCard)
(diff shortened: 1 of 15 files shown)


ZeldaZach (commented):
The shape of the refactor is great — `PlayerActions` as a command-builder + `PlayerDialogs` for dialog orchestration is the right split, and the doc-style description in the PR body is genuinely useful for future readers. The signal/slot wiring through `PlayerDialogs`'s constructor is clean.

That said, this PR is where six bugs originate that I had flagged on #6928 (which inherits this branch). Cross-posting them here because the fixes belong on this PR, not the downstream one. They're all in the `createRelatedFromRelation` rewrite and the new `PlayerDialogs` class:

**`actCreateRelatedCard` / `createRelatedFromRelation` chain (3 issues):**
- `actCreateAllRelatedCards` now does `++tokensTypesCreated` unconditionally where the old code only counted on success — affects the `tokensTypesCreated == 1` branches below.
- `createRelatedFromRelation` falls through to the attach branch when `getIsVariable() && count == 1`; old code always used `DoesNotAttach` for variable relations regardless of count.
- `onRelatedCardCreated` has no null guards on `sourceCard` / `cardRelation` even though `createRelatedFromRelation` (its sibling) does, and `PlayerDialogs` calls it unconditionally after `createRelatedFromRelation` returns. Contracts no longer agree.

**`PlayerDialogs` issues:**
- `onCreateRelatedFromRelationDialogRequested` calls `onRelatedCardCreated` even when nothing was created (no equivalent to the old `if (createRelatedFromRelation(...))` gate).
- Three sites have `player->clearCardsToDelete()` commented out (`// player->clearCardsToDelete() ||`) — looks like a forgotten TODO. Old code bailed when a card the user was editing got deleted mid-dialog.
- Include path nit: `player_dialogs.cpp` line 5 uses `../client/...` (one `..` too few for a 2-level-deep file); should be `../../client/...`. Compiles only via include-path fallback.

Details inline. None of this is structural — the architecture lands well, it's just the in-flight semantics that need touching up.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -982,9 +988,8 @@ void PlayerActions::actCreateAllRelatedCards()
 
     if (relatedCards.length() == 1) {
         cardRelation = relatedCards.at(0);
-        if (createRelatedFromRelation(sourceCard, cardRelation)) {
-            ++tokensTypesCreated;
-        }
+        actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
+        ++tokensTypesCreated;
```
ZeldaZach: Behavior regression: `++tokensTypesCreated` is now unconditional. The old `createRelatedFromRelation` returned `bool`, and this branch was `if (createRelatedFromRelation(...)) { ++tokensTypesCreated; }`, so the count only went up on success. With the rewrite, `actRequestCreateRelatedFromRelationDialog` is void, so we always increment — which means the `tokensTypesCreated == 1` checks downstream (lines 1019 / 1034) can fire even when nothing was actually created.

Suggest threading the bool back through — either give `actRequestCreateRelatedFromRelationDialog` a synchronous return value (the signal/slot is direct-connected, so it works) or have `PlayerDialogs` emit a result signal that increments the counter.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -996,9 +1001,9 @@ void PlayerActions::actCreateAllRelatedCards()
         switch (nonExcludedRelatedCards.length()) {
             case 1: // if nonExcludedRelatedCards == 1
                 cardRelation = nonExcludedRelatedCards.at(0);
-                if (createRelatedFromRelation(sourceCard, cardRelation)) {
-                    ++tokensTypesCreated;
-                }
+                actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
+                ++tokensTypesCreated;
+
```
ZeldaZach: Same regression as the `relatedCards.length() == 1` branch above — `++tokensTypesCreated` is now unconditional where it was previously gated on `createRelatedFromRelation` returning `true`.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -1045,48 +1050,74 @@ void PlayerActions::actCreateAllRelatedCards()
...
-            return false;
+        count = variableCount;
+        if (count <= 0) {
+            return;
         }
+    } else {
+        count = cardRelation->getDefaultCount();
+    }
+
+    if (count > 1) {
```
ZeldaZach, line 1091: Behavior change for variable-count relations: the old structure was `if (isVariable) { for loop with DoesNotAttach } else if (defaultCount > 1) { for loop with DoesNotAttach } else { attach branch }`. So a variable relation **always** used `DoesNotAttach`, regardless of the count the user picked.

The new code collapses to `if (count > 1) { DoesNotAttach loop; return; } /* else attach branch */`. That means a variable relation where the user types `1` in the dialog now falls through to the attach branch (lines 1088–1103) and may call `playCardToTable(sourceCard, false)` on `AttachTo` relations — different from the old behavior.

If this was intentional, worth a comment explaining; otherwise gate on `cardRelation->getIsVariable()` to preserve the old branch.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -1045,48 +1050,74 @@ void PlayerActions::actCreateAllRelatedCards()
...
+}
 
-        createCard(sourceCard, dbName, attachType, persistent);
+void PlayerActions::onRelatedCardCreated(const CardItem *sourceCard, const CardRelation *cardRelation)
+{
+    /*
+     * If we make a token via "Token: TokenName"
+     * then let's allow it to be created via "create another token"
+     */
+    if (!cardRelation->getCanCreateAnother()) {
```
ZeldaZach, line 1127: Possible null deref: `cardRelation->getCanCreateAnother()` and later `sourceCard->getCard().getPrinting()` get called with no null guards. The sibling `createRelatedFromRelation` (just above) does check both for null and bails — `onRelatedCardCreated` is called right after it in `PlayerDialogs::onCreateRelatedFromRelationDialogRequested`, so the two functions need to agree on the contract.

Either add the same null guards here, or only invoke `onRelatedCardCreated` after a confirmed successful creation in the dialog handler.

cockatrice/src/game/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
+#include "player_dialogs.h"
+
+#include "../../interface/widgets/utility/get_text_with_max.h"
+#include "../board/card_item.h"
+#include "../client/settings/card_counter_settings.h"
```
ZeldaZach: Path nit: this file is at `game/player/`, so `../client/...` resolves to `game/client/...` (doesn't exist) rather than `src/client/...`. It compiles because of how include search paths resolve, but it's inconsistent with how the 3-level-deep files in the same tree write it — e.g. `library_menu.cpp` correctly uses `../../../client/settings/...` from `game/player/menu/`. Should be `../../client/settings/card_counter_settings.h` here.

cockatrice/src/game/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+        emit requestDialogSemaphore(false);
+
+        if (!ok) {
+            return;
+        }
+    }
+
+    playerActions->createRelatedFromRelation(sourceCard, cardRelation, variableCount);
+
+    playerActions->onRelatedCardCreated(sourceCard, cardRelation);
```
ZeldaZach: `onRelatedCardCreated` runs unconditionally here, but `createRelatedFromRelation` (line 213) has multiple early-return paths (null `sourceCard`/`cardRelation`, `count <= 0` for variable relations). The old, synchronous `createRelatedFromRelation` returned `bool` for exactly this purpose — `actCreateRelatedCard`/`actCreateAllRelatedCards` gated their follow-up `setLastToken` call on it.

With the new void signature, the follow-up `setLastToken` machinery fires even when nothing was created. Combined with `onRelatedCardCreated` having no null guards (see line 1113 of player_actions.cpp), this is a likely crash path for the malformed-relation case.

Fix: have `createRelatedFromRelation` return `bool` and gate this call on it, or move the `setLastToken` logic into `createRelatedFromRelation` itself.

cockatrice/src/game/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+void PlayerDialogs::onSetPTDialogRequested(const QString &oldPT)
+{
+    bool ok;
+    emit requestDialogSemaphore(true);
+    QString pt = getTextWithMax(dialogParent(), tr("Change power/toughness"), tr("Change stats to:"), QLineEdit::Normal,
+                                oldPT, &ok);
+    emit requestDialogSemaphore(false);
+
+    if (!ok) { // player->clearCardsToDelete() ||
```
ZeldaZach: Looks like a forgotten TODO — the old `actSetPT` in `PlayerActions` had `if (player->clearCardsToDelete() || !ok) { return; }`, which bailed when cards were deleted while the dialog was open (e.g., the card the user was editing got removed by an opponent's action). With the check commented out, we'll now proceed to call `actSetPT(selectedCards, pt)` on cards that may already be gone.

Same issue at line 267 (`onSetAnnotationDialogRequested`) and 285 (`onSetCardCounterDialogRequested`). Either restore the check (probably as a signal from `PlayerDialogs` back to `PlayerLogic`) or remove the comment and intentionally drop the semantics if no longer needed.

cockatrice/src/game/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+{
+    emit requestDialogSemaphore(true);
+    AnnotationDialog *dialog = new AnnotationDialog(dialogParent());
+    dialog->setOptions(QInputDialog::UsePlainTextEditForTextInput);
+    dialog->setWindowTitle(tr("Set annotation"));
+    dialog->setLabelText(tr("Please enter the new annotation:"));
+    dialog->setTextValue(oldAnnotation);
+    bool ok = dialog->exec();
+    emit requestDialogSemaphore(false);
+    if (!ok) { // player->clearCardsToDelete() ||
```
ZeldaZach: Same `clearCardsToDelete()` guard removal as line 250 — was active in the original `actSetAnnotation`.

cockatrice/src/game/player/player_dialogs.cpp
```diff
@@ -0,0 +1,289 @@
...
+    emit requestDialogSemaphore(true);
+
+    auto &cardCounterSettings = SettingsCache::instance().cardCounters();
+    QString counterName = cardCounterSettings.displayName(counterId);
+
+    AbstractCounterDialog dialog(counterName, oldValueForDlg, dialogParent());
+    int ok = dialog.exec();
+
+    emit requestDialogSemaphore(false);
+    if (!ok) { // player->clearCardsToDelete() ||
```
ZeldaZach: Same `clearCardsToDelete()` guard removal as line 250 — was active in the original `actSetCardCounter`.

cockatrice/src/game/player/player_actions.cpp
```diff
@@ -982,7 +988,9 @@ void PlayerActions::actCreateAllRelatedCards()
 
     if (relatedCards.length() == 1) {
         cardRelation = relatedCards.at(0);
-        if (createRelatedFromRelation(sourceCard, cardRelation)) {
+        lastRelatedCreationSucceeded = false; // reset before emit
+        actRequestCreateRelatedFromRelationDialog(sourceCard, cardRelation);
+        if (lastRelatedCreationSucceeded) {
             ++tokensTypesCreated;
         }
```
ZeldaZach, line 995: This pattern — `lastRelatedCreationSucceeded = false; actRequestCreateRelatedFromRelationDialog(...); if (lastRelatedCreationSucceeded) ...` — depends on the signal dispatch from `PlayerActions::requestCreateRelatedFromRelationDialog` to `PlayerDialogs::onCreateRelatedFromRelationDialogRequested` running *synchronously* and setting the flag before `emit` returns. That's true today (default same-thread `AutoConnection` resolves to `DirectConnection`), but it's an implicit contract that the next person touching the connection (e.g., to add `Qt::QueuedConnection` for thread safety, or to introduce an indirection) will silently break — `tokensTypesCreated` would then always be 0 and the "Create another" shortcut wouldn't get its first token.

Same issue in the `case 1:` branch a few lines below (line 1009).

Suggest making the contract explicit, e.g. give `actRequestCreateRelatedFromRelationDialog` a `bool` return value and forward what the dialog handler decided, or have the dialog handler call back via an explicit `relatedCreationResult(bool)` signal.

cockatrice/src/game/player/player_dialogs.cpp
```diff
@@ -0,0 +1,295 @@
...
+    emit requestDialogSemaphore(true);
+    QString pt = getTextWithMax(dialogParent(), tr("Change power/toughness"), tr("Change stats to:"), QLineEdit::Normal,
+                                oldPT, &ok);
+    emit requestDialogSemaphore(false);
+
+    if (!ok || player->getLogic()->clearCardsToDelete()) {
+        return;
+    }
+
+    playerActions->actSetPT(player->getGameScene()->selectedCards(), pt);
```
ZeldaZach: Subtle behavior change vs. the pre-refactor code: the old `PlayerActions::actSetPT` snapshotted `auto cards = selectedCards()` *before* `setDialogSemaphore(true)` and iterated that local list afterward. Here, the selection is re-read from the scene *after* the modal dialog returns.

In the normal flow this is fine — the dialog is modal, and `clearCardsToDelete()` aborts the operation if any of the selected cards were destroyed during the wait. But Qt modal dialogs still pump events (network messages from opponents, queued connections, etc.), so if any path ever mutates the scene selection while the dialog is up (e.g. a newly-added card becomes selected, or a hotkey juggles focus and changes selection state), the SetPT will apply to a *different* card set than what the user had selected when they opened the dialog.

Cheap fix: snapshot before emitting the semaphore signal:

```cpp
auto cards = player->getGameScene()->selectedCards();
emit requestDialogSemaphore(true);
// ... dialog ...
emit requestDialogSemaphore(false);
if (!ok || player->getLogic()->clearCardsToDelete()) return;
playerActions->actSetPT(cards, pt);
```

Same pattern applies to `onSetAnnotationDialogRequested` (line 277) and `onSetCardCounterDialogRequested` (line 294).

ZeldaZach (commented):
Nice refactor — extracting dialog orchestration out of `PlayerActions` is the right move and the signal/slot wiring is complete. Two design-fragility notes inline; neither is a hard bug today but both make assumptions worth surfacing.

ZeldaZach (approved):

PR #6958, opened 2026-05-30
CI: Add CodeQL workflow

### LINKED ISSUE #5105: CodeQL configuration for scanning C++ is failing
The CodeQL default setup for `C/C++` that GitHub rolled out automatically does not work on our repo.

The build command does not succeed and needs some manual tweaking.

tooomm:
## Related Ticket(s)
- Fixes #5105

CodeQL was automatically enabled by GitHub, but the `autobuild` mode set by default was not working with our project.

- Add manual steps for building the app to get best scan results rather than checking source files only
- Analyses `C++` code and `GitHub Actions` configurations for vulnerabilities and errors
- Comments will be posted to PR's
- Results will be uploaded to [GitHub Code Scanning in the Security tab](https://github.com/Cockatrice/Cockatrice/security/code-scanning) in the repo

Docs: https://docs.github.com/en/code-security/how-tos/find-and-fix-code-vulnerabilities

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -54,7 +54,7 @@ jobs:
         dependency-caching: true
 
     - name: "[C++] Install dependencies"
-      if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
+      if: matrix.language == 'cpp' && matrix.build-mode == 'manual'
       shell: bash
       env:
         DEBIAN_FRONTEND: noninteractive
@@ -79,12 +79,12 @@ jobs:
 # Add ccache usage for faster compilation, (install ccache dep, actions/cache step + append DUSE_CCACHE=1 in cmake config, CCACHE env values)
 
     - name: "[C++] Configure CMake"
-      if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
+      if: matrix.language == 'cpp' && matrix.build-mode == 'manual'
       shell: bash
       run: cmake -S . -B build -G Ninja -DWITH_SERVER=1 -DCMAKE_BUILD_TYPE=Release
 
     - name: "[C++] Build application"
-      if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
+      if: matrix.language == 'cpp' && matrix.build-mode == 'manual'
       shell: bash
       run: cmake --build build
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -67,12 +67,8 @@ jobs:
           liblzma-dev \
           libmariadb-dev-compat \
           libprotobuf-dev \
-          libqt6multimedia6 \
-          libqt6sql6-mysql \
           ninja-build \
           protobuf-compiler \
-          qt6-image-formats-plugins \
-          qt6-l10n-tools \
           qt6-multimedia-dev \
           qt6-svg-dev \
           qt6-tools-dev \
@@ -85,7 +81,7 @@ jobs:
     - name: "[C++] Configure CMake"
       if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
       shell: bash
-      run: cmake -B build -S . -G Ninja -DWITH_SERVER=1 -DCMAKE_BUILD_TYPE=Release
+      run: cmake -S . -B build -G Ninja -DWITH_SERVER=1 -DCMAKE_BUILD_TYPE=Release
 
     - name: "[C++] Build application"
       if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,98 @@
+# GitHub Docs on Code Scanning:
+# https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning
+# https://docs.github.com/en/code-security/how-tos/find-and-fix-code-vulnerabilities/manage-your-configuration
+# https://docs.github.com/en/code-security/reference/code-scanning/workflow-configuration-options
+#
+# CodeQL Docs:
+# https://codeql.github.com/docs/
+
+name: CodeQL
+
+permissions:
+  security-events: write    # needed to post results
+  contents: read
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+      - 'doc/**'
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+      - 'doc/**'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        # https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/
+        - language: cpp
+          build-mode: manual
+        - language: actions
+          build-mode: none
+
+    steps:
+    - name: "Checkout repository"
+      uses: actions/checkout@v6
+
+    - name: "Initialize CodeQL"
+      uses: github/codeql-action/init@v4
+      with:
+        languages: ${{ matrix.language }}
+        build-mode: ${{ matrix.build-mode }}
+        queries: security-and-quality
+        dependency-caching: true
+
+    - name: "[C++] Install dependencies"
+      if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
+      shell: bash
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          cmake \
+          g++ \
+          libgl-dev \
+          liblzma-dev \
+          libmariadb-dev-compat \
+          libprotobuf-dev \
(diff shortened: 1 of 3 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,94 @@
...
+      matrix:
+        include:
+        # https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/
+        - language: cpp
+          build-mode: manual
+        - language: actions
+          build-mode: none
+
+    steps:
+    - name: "Checkout repository"
```
fearphage, line 42: Quotes here are a recent addition, but everyhting that isn't another type is a string by default so these quotes here are superflous. On line 31 you have another `name` with no quotes and the following line is also a string with no quotes.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: "Initialize CodeQL"
+      uses: github/codeql-action/init@v4
+      with:
+        languages: ${{ matrix.language }}
+        build-mode: ${{ matrix.build-mode }}
+        queries: security-and-quality
+        dependency-caching: true
+
+    - name: "[C++] Install dependencies"
+      if: matrix.build-mode == 'manual' && matrix.language == 'cpp'
```
fearphage: It would be slightly cleaner if you just separated these by jobs instead and then you wouldn't have to duplicate this condition so many times. Also there's no other `build-mode` possible for `cpp`, do you need to check both values?

```yaml
jobs:
  check-actions:
    if: needs.filter.outputs.run-actions == 'true'
    needs: filter
    steps: ...
  check-cpp:
    if: needs.filter.outputs.run-cpp == 'true'
    needs: filter
    steps: ...
  # determine which to run based on the changed files
  filter:
    outputs: 
      run-actions: ...
      run-cpp: ...
```

.github/workflows/codeql.yml
tooomm: This is related to my comment in your other PR.

I had to add quotes to the step name value to use square brackets in the step names in the build workflow.
Here, I'm using brackets as well for some (e.g. `"[C++] Install dependencies"`).
For consistency and easy of use once not so experienced contributors change step names in the future, I added them to all step names. Also those who would not need it currently.

Line 31 is the job name, not a step name. But I get your point.
In recent weeks and in recent PR's, I tried to apply various small improvements to the workflow files. Only using quotes where really needed (or putting them everywhere) is fine for me as well!

.github/workflows/codeql.yml
tooomm: In general I try to be transparent with the conditions and add some extra ones for clarity... our config files are long and not always straight forward if you take a first look.
Like in the build workflow we only have `use_ccache` value for macOS builds (linux build use ccache as well though), however I check for mac and the variable in the relevant step.

Would your suggestion result in a "filter" job which always runs, and then there is one (or both) of the "check-*" jobs running depending on the files changed?
That would add 1-2 extra entries to the list of running jobs. I like that the current setup has a clean list of jobs grouped on topic a bit and we only have the Configuration job separated.

Maybe we could combine the configuration and filter jobs into one (move it to a separate workflow file as well)?

Anyway, I would consider that a refactor we can do once the feature itself is added.

<br>

[Supported build modes for C++](https://docs.github.com/en/enterprise-cloud@latest/code-security/reference/code-scanning/codeql/build-options-for-compiled-languages#building-cc) are: `none`, `manual` and `autobuild`

.github/workflows/codeql.yml
fearphage: I think it really comes down to a matter of preference here. From my perspective, the quotes add noise and inconsistency. As far as new contributors go, I think it opens the door to confusion.

> This `name` field has quotes, but the other one doesn't. Why? Which example is correct? Which one do I follow? Should I fix the _wrong_ ones in my pull request?

I believe the linters/parsers/LSPs will tell you when you need to quote strings and that makes sense to me. I believe both positions are valid though. Different strokes and all that jazz.

.github/workflows/codeql.yml
fearphage: > [Supported build modes for C++](https://docs.github.com/en/enterprise-cloud@latest/code-security/reference/code-scanning/codeql/build-options-for-compiled-languages#building-cc) are: `none`, `manual` and `autobuild`

But those aren't represented here today. There are currently [2 possibilities available in the current matrix](https://github.com/Cockatrice/Cockatrice/pull/6958/changes#diff-12783128521e452af0cfac94b99b8d250413c516ec71fe6d97dbea666ff7ba27R37-R42) -- `actions` and `cpp`. No matter how many properties you add to those 2 options, they are still just 2 of them and checking the `build-mode` will function identically to checking the `build-mode` and N other fields for each job. The added complexity feels like it only exists for complexity's sake. When you add support for the other build modes of `cpp`, then you can add the code that actually changes behavior relative to those newly added build modes. That's all I was saying.

I don't think I originally understood the flow of the code when I made the YAML recommendation above. Yeah, ignore that.

tooomm: Your point is to not add something that is not relevant currently?

I added both checks to ensure this steps are only run for the combination of cpp+manual.
If the mode is changed to "autobuild", no manual setup and build steps are expected to run.
Same with "none", it's designed to analyze with less accuracy but does not require a build.
The steps should then be skipped.

Am I overthinking this? 🤔
You're saying if the workflow/mode is changed, those steps should be updated as needed?

fearphage: > You're saying if the workflow/mode is changed, those steps should be updated as needed?

Yes, that's what' I'm saying. Everything beyond checking `build-mode == 'cpp'` is superfluous. The current settings only have 2 options. Since the other option's value is not `cpp`, you could even make the check be equally as binary as the existing options are.

```yaml
- if: matrix.build-mode == 'cpp'
  run: do cpp stuff

# not a recommendation but functionally equivalent in this instance
- if: matrix.build-mode != 'cpp'
  run: do the only other non-cpp option
```

> I added both checks to ensure this steps are only run for the combination of cpp+manual.

But there is _no alternative combination available_ to `cpp` + `X`. In the current set up, the variable `X` is the _constant_ `manual`. To restate, if your binary set up had more values. 

```yaml
include:
  - another: option
    bulld-mode: cpp
    catpants: true
    dont-forget: this one

  - build-mode: actions
```

You could write those into the check:

```yaml
if: |
  matrix.build-mode == 'cpp' &&
  matrix.another == 'option' &&
  matrix.dont-forget == 'this one' &&
  matrix.catpants == true
```

But that's functionally equivalent to `if: matrix.build-mode == 'cpp'` since there are no other options that have `build-mode` set to `cpp`. The rest is just noise until there are more `cpp` build modes supported in the matrix.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,96 @@
...
+# CodeQL Docs:
+# https://codeql.github.com/docs/
+
+name: CodeQL
+
+permissions:
+  security-events: write    # needed to post results
+  contents: read
+
+on:
```
ZeldaZach, line 15: Consider adding a `concurrency` group like the other workflows in this repo (`desktop-build.yml`, `docker-release.yml`) so that a new push to a PR cancels the previous in-progress run. The `cpp` job here does a full manual compile, so overlapping runs on rapid pushes will burn a lot of CI time.

```yaml
concurrency:
  group: "${{ github.workflow }} @ ${{ github.ref_name }}"
  cancel-in-progress: true
```

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,96 @@
...
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+      - 'doc/**'
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
```
ZeldaZach: One thing to watch with `paths-ignore` on the `pull_request` trigger: if the CodeQL check is ever set as a required status check, a PR that only touches ignored paths (e.g. a docs-only or `.md`-only change) will skip this job entirely, so the required check stays pending and blocks merge. GitHub documents this gotcha for path filters on required checks.

Every other workflow here uses `paths:` with negations rather than `paths-ignore` (see `desktop-build.yml`), so matching that pattern would also keep things consistent.

ZeldaZach (approved):
Approving — this correctly replaces the non-working \`autobuild\` default with a manual C++ build (\`WITH_SERVER=1\`) plus Actions analysis, and the build steps are properly gated by the matrix. Left two non-blocking suggestions inline (adding a \`concurrency\` group, and reconsidering \`paths-ignore\` on the \`pull_request\` trigger).

.github/workflows/codeql.yml
tooomm: Jep, your right.
I added it for building docs as well. It was missing there as well to avoid issues.

.github/workflows/codeql.yml
tooomm: Switching to `path` with negations would not change the mentioned concerns?
Made it run without pattern filtering.

tooomm:
Looking forward to seeing how this works in action. 👍 

Failing run is due to our Arch build hitting a defined timeout of a performance test. This is also on master and is ticketed. Not related to the changes here.

I'll check back on the conversation with fearphage as I'm working on other CI and config improvements (see e.g. #7014). Don't want this one to wait even longer.

PR #6855, opened 2026-05-08
bump vcpkg to 2026.04.27 release

tooomm:
With https://github.com/Cockatrice/Cockatrice/commit/63143f941643dc2f46f658cb51ea3b5727da3e56, this is now possible.

See comments in https://github.com/Cockatrice/Cockatrice/pull/6627

PR #6959, opened 2026-05-30
CI: Cleanup

tooomm:
- Alphabetical ordering of permissions and uniform placement at the top
- Alphabetical ordering of job triggers (`on:`), environment variables (`env:`) and action configs (`with:`)
- Alphabetical ordering of matrix parameter (besides the "header")
- Adjust and harmonize args construction in bash scripts
- Use `${{ name }}` styling in all places for referencing variables
- Add OS label in combined vcpkg build job to indicate which steps are OS-specific and skipped in some runs (or why there are various Qt steps).
  Quotes need to be added to the `name` value, otherwise the brackets will lead to a parsing error.
  Added them to all step names across the workflow configs for consistency.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -1,7 +1,7 @@
 name: Build Desktop
 
 permissions:
-  actions: write    # needed for ccache action to be able to delete gha caches
+  actions: write    # needed to delete entries in GHA cache (update ccache)
   attestations: write
   contents: write
   id-token: write
@@ -46,7 +46,7 @@ jobs:
       sha: ${{ steps.configure.outputs.sha }}
 
     steps:  
-      - name: Configure
+      - name: "Configure"
         id: configure
         shell: bash
         run: |
@@ -62,13 +62,13 @@ jobs:
           fi
           echo "sha=$sha" >>"$GITHUB_OUTPUT"
 
-      - name: Checkout
+      - name: "Checkout"
         if: steps.configure.outputs.tag != null
         uses: actions/checkout@v6
         with:
-          fetch-depth: 0
+          fetch-depth: 0    # fetch all history for all branches and tags
 
-      - name: Prepare release parameters
+      - name: "Prepare release parameters"
         id: prepare
         if: steps.configure.outputs.tag != null
         shell: bash
@@ -76,7 +76,7 @@ jobs:
           TAG: ${{ steps.configure.outputs.tag }}
         run: .ci/prep_release.sh
 
-      - name: Create release
+      - name: "Create release"
         if: steps.configure.outputs.tag != null
         id: create_release
         shell: bash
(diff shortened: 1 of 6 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -19,7 +19,6 @@ on:
       - '.github/workflows/desktop-build.yml'
       - 'CMakeLists.txt'
       - 'vcpkg.json'
```
ZeldaZach, line 21: The PR description says "Changes in the folder are covered with the `*/**` wildcards already," but that isn't true for a *submodule pointer bump*. From the parent repo's perspective, `vcpkg` is a single root-level gitlink — `git ls-files` shows it as `160000 <sha> 0\tvcpkg`, not a directory. A submodule update like #6855 ("bump vcpkg to 2026.04.27") shows up in `git diff` as a change to one path: `vcpkg` (no slash, no extension).

GitHub Actions matches `paths:` against that file list with minimatch, and:
- `*/**` requires at least one `/`, so it won't match `vcpkg`.
- `CMakeLists.txt` / `vcpkg.json` are exact names that don't match either.

Net effect after this PR: PRs that *only* bump the vcpkg submodule pointer will no longer trigger Build Desktop, which is the exact thing we need to verify on those PRs. Same issue on the `pull_request: paths:` block (line 33). Suggest restoring the `- 'vcpkg'` entry in both.

ZeldaZach (commented):
Cleanup looks good overall. One real concern with the `vcpkg` path-trigger removal — that change can silently skip CI on the exact commits we most need it for (submodule bumps). Two minor nits inline as well.

.github/workflows/desktop-build.yml
```diff
@@ -263,200 +265,202 @@ jobs:
...
+      - name: "[macOS] Cache thin Qt libraries"
         if: matrix.os == 'macOS' && steps.restore_qt.outputs.cache-hit != 'true'
         uses: actions/cache/save@v5
         with:
-          path: ${{ github.workspace }}/Qt
           key: thin-qt-macos-${{ matrix.soc }}-${{ steps.resolve_qt_version.outputs.version }}
+          path: ${{ github.workspace }}/Qt
 
-      - name: Install Qt ${{matrix.qt_version}} (Windows)
+      - name: "[Windows] Install Qt ${{matrix.qt_version}}"
```
ZeldaZach: Spacing nit: `${{matrix.qt_version}}` here is missing the spaces inside the braces, while every other interpolation in this PR was normalized to `${{ … }}`. (Same step's `with:` block below already uses `${{ matrix.qt_arch }}` etc., so this one was just missed.)

.github/workflows/docker-release.yml
```diff
@@ -24,54 +28,50 @@ jobs:
...
+      - name: "Docker metadata"
         id: metadata
         uses: docker/metadata-action@v6
         env:
           DOCKER_METADATA_ANNOTATIONS_LEVELS: index    # needed for GHCR
         with:
+          annotations: |
+            org.opencontainers.image.title=Servatrice
+            org.opencontainers.image.url=https://cockatrice.github.io/
+            org.opencontainers.image.description=Server for Cockatrice, a cross-platform virtual tabletop for multiplayer card games
```
ZeldaZach: Trailing whitespace on this line (`…multiplayer card games  ` — two trailing spaces). The identical `labels:` block on line 51 doesn't have them. Pre-existing in the source `labels:` block, but worth fixing while it's being touched here.

ZeldaZach (approved):

.github/workflows/docker-release.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           annotations: |
             org.opencontainers.image.title=Servatrice
             org.opencontainers.image.url=https://cockatrice.github.io/
-            org.opencontainers.image.description=Server for Cockatrice, a cross-platform virtual tabletop for multiplayer card games
```
fearphage, line 45: @tooomm Most editors have a setting for this. Look for "trim trailing whitespace on save". I know (Neo)ViM, VSCode, JetBrains, etc. all have a method to do this. Then it becomes something you never have to worry about.

tooomm: Yep, found something in my editor as well!

PR #6961, opened 2026-05-31
[TabDeckEditor] Refactor: consolidate add/decrement card signals

RickyRister:
## Related Ticket(s)
- Follow-up to #6956
- Follow-up to #6939

The signal/slot wiring becomes much nicer once we get rid of the individual methods for add/decremet card from main/side and just have everything take the zone as a param.

- Remove the `actXXXCard` methods from `AbstractTabDeckEditor` entirely
  - Replace them with directly calling the `addCard` and `decrementCard` methods with the zoneName as a param.
- Clean up and consolidate signals of classes that connected to the removed slots.
- Fix regression introduced in #6956 that broke ctrl in the vde quick add widget

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual_tab_widget.cpp
+++ b/cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual_tab_widget.cpp
@@ -34,8 +34,8 @@ TabDeckEditorVisualTabWidget::TabDeckEditorVisualTabWidget(QWidget *parent,
             &TabDeckEditorVisualTabWidget::onCardChanged);
     connect(visualDeckView, &VisualDeckEditorWidget::cardClicked, this,
             &TabDeckEditorVisualTabWidget::onCardClickedDeckEditor);
-    connect(visualDeckView, &VisualDeckEditorWidget::cardAdditionRequested, deckEditor,
-            &AbstractTabDeckEditor::actAddCard);
+    connect(visualDeckView, &VisualDeckEditorWidget::cardAdditionRequested, this,
+            &TabDeckEditorVisualTabWidget::actAddCard);
 
     visualDatabaseDisplay =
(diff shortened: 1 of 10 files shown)


cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual_tab_widget.cpp
```diff
@@ -166,3 +166,15 @@ void TabDeckEditorVisualTabWidget::handleTabClose(int index)
     this->removeTab(index);
     delete tab;
 }
+
+void TabDeckEditorVisualTabWidget::actAddCard(const ExactCard &card)
+{
+    QString zoneName;
+    if (QApplication::keyboardModifiers() & Qt::ControlModifier) {
```
ZeldaZach, line 173: Subtle behavior change worth confirming: previously `VisualDeckEditorWidget::cardAdditionRequested` (emitted by the search bar `returnPressed` and the search button click in `visual_deck_editor_widget.cpp:80,148`) was wired to `AbstractTabDeckEditor::actAddCard`, which always added to `DECK_ZONE_MAIN` regardless of modifier state. After this PR, the same signal is routed through the new `TabDeckEditorVisualTabWidget::actAddCard`, which reads `QApplication::keyboardModifiers()` and routes to `DECK_ZONE_SIDE` when Ctrl is held.

That means **Ctrl + Enter in the visual deck editor search bar** (and Ctrl + clicking the search push button) now adds to the sideboard instead of the main deck. That is consistent with the old `actAddCard` docstring that this PR is removing ("Adds a card to the main deck or sideboard depending on Ctrl key"), so it is probably a restoration of intended behavior after #6956 — but worth calling out explicitly since the PR description frames the change as a pure consolidation. If intentional, may be worth a line in the PR description; if unintentional, drop the Ctrl branch here and pass `DECK_ZONE_MAIN` unconditionally.

ZeldaZach (approved):

PR #6962, opened 2026-06-01
[TabDeckEditor] Refactor: pass ExactCard in signal instead of widget

RickyRister:
## Related Ticket(s)
- Follow-up to #6961

The code that relays the card click event from the card database to the vde passes the entire deck info widget in the signal, only extracting the ExactCard at the end. It would make more sense and be cleaner if we extracted the ExactCard from the start and just passed that in the signal.

- Change the card click signal/slots along the entire chain to pass `ExactCard` instead of `CardInfoPictureWithTextOverlayWidget`
- Changed `CardInfoPictureWidget::cardClicked`'s signature to also have the `ExactCard`
  - Changed the subclass `CardInfoPictureWithTextOverlayWidget` to just use that signal instead of its own onClick signal.
- Some of the signatures weren't passing zoneNames as const ref, so fixed it while I was at it.

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.h
+++ b/cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.h
@@ -41,7 +41,7 @@
  * - changeModelIndexAndCardInfo(const ExactCard &card) — Updates deck model selection and card info.
  * - changeModelIndexToCard(const ExactCard &card) — Selects the card in the deck view.
  * - processMainboardCardClick(QMouseEvent *event, ...) — Handles clicks on mainboard cards.
- * - processCardClickDatabaseDisplay(QMouseEvent *event, ...) — Handles clicks on database cards.
+ * - processDatabaseCardClick(QMouseEvent *event, ...) — Handles clicks on database cards.
  * - actSaveDeckAs() — Overrides save action with temporary UI adjustments.
  * - showPrintingSelector() — Opens the printing selector dock for the current card.
  * - freeDocksSize() — Frees constraints on dock widget sizes.
--- a/cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.h
+++ b/cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.h
@@ -152,19 +152,17 @@ public slots:
     /**
      * @brief Handle card clicks in the mainboard visual deck.
      * @param event Mouse event triggering the action.
-     * @param instance Widget representing the clicked card.
+     * @param card The clicked card.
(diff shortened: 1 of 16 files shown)


cockatrice/src/interface/widgets/visual_deck_editor/visual_deck_editor_widget.cpp
```diff
@@ -401,11 +401,9 @@ void VisualDeckEditorWidget::decklistDataChanged(QModelIndex topLeft, QModelInde
 //                                                 User Interaction
 // =====================================================================================================================
 
-void VisualDeckEditorWidget::onCardClick(QMouseEvent *event,
-                                         CardInfoPictureWithTextOverlayWidget *instance,
-                                         QString zoneName)
+void VisualDeckEditorWidget::onCardClick(QMouseEvent *event, const ExactCard &card, const QString &zoneName)
```
ZeldaZach: Dead code: `VisualDeckEditorWidget::onCardClick` is now unused. The only connection that fed it (in `constructZoneWidgetForIndex` at line 284) was rewritten in this PR to forward directly from `DeckCardZoneDisplayWidget::cardClicked` to `VisualDeckEditorWidget::cardClicked` (signal-to-signal relay), so this slot has no remaining callers. `grep -rn "onCardClick" cockatrice/src` confirms it only appears in its own declaration (`visual_deck_editor_widget.h:85`) and definition.

This is the same kind of relay-slot removal you already did for `CardGroupDisplayWidget::onClick` / `onHover` in `card_group_display_widget.cpp` — just missed here. Suggest deleting `onCardClick` from both `.h` and `.cpp` to finish the cleanup.

cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.h
```diff
@@ -152,19 +152,17 @@ public slots:
...
+    void processMainboardCardClick(const QMouseEvent *event, const ExactCard &card, const QString &zoneName);
 
     /**
      * @brief Handle card clicks in the database visual display.
      * @param event Mouse event triggering the action.
-     * @param instance Widget representing the clicked card.
+     * @param card  The clicked card.
      */
-    void processCardClickDatabaseDisplay(QMouseEvent *event, CardInfoPictureWithTextOverlayWidget *instance);
+    void processDatabaseCardClick(const QMouseEvent *event, const ExactCard &card);
```
ZeldaZach, line 165: Stale doc comment outside the diff: the class-level "Key Methods" list at `tab_deck_editor_visual.h:44` still references the old name `processCardClickDatabaseDisplay`, but this PR renamed it to `processDatabaseCardClick`. Worth updating that line so the class docs stay in sync with the renamed slot you added here.

ZeldaZach (approved):

PR #6963, opened 2026-06-03
[TabDeckEditor] Refactor: Remove cardDatabase field from analysis interfaces

RickyRister:

There's no reason for `DeckStatsInterface` and `TappedOutInterface` to hold the `CardDatabase` in a field, when we just directly query using the `CardDatabaseManager` everywhere else in the project.

- Remove `cardDatabase` field from those classes and fix signatures
- Use `CardDatabaseManager::query()` in those classes instead

### DIFF
--- a/cockatrice/src/client/network/interfaces/deck_stats_interface.h
+++ b/cockatrice/src/client/network/interfaces/deck_stats_interface.h
@@ -7,7 +7,6 @@
 #ifndef DECKSTATS_INTERFACE_H
 #define DECKSTATS_INTERFACE_H
 
-#include <libcockatrice/card/database/card_database.h>
 #include <libcockatrice/deck_list/deck_list.h>
 
 class QByteArray;
--- a/cockatrice/src/client/network/interfaces/deck_stats_interface.h
+++ b/cockatrice/src/client/network/interfaces/deck_stats_interface.h
@@ -21,8 +21,6 @@ class DeckStatsInterface : public QObject
 private:
     QNetworkAccessManager *manager;
 
-    CardDatabase &cardDatabase;
-
     /**
      * Deckstats doesn't recognize token cards, and instead tries to find the
      * closest non-token card instead. So we construct a new deck which has no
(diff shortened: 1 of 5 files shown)


cockatrice/src/client/network/interfaces/deck_stats_interface.h
```diff
@@ -35,7 +33,7 @@ private slots:
     void getAnalyzeRequestData(const DeckList &deck, QByteArray &data);
 
 public:
-    explicit DeckStatsInterface(CardDatabase &_cardDatabase, QObject *parent = nullptr);
+    explicit DeckStatsInterface(QObject *parent = nullptr);
     void analyzeDeck(const DeckList &deck);
 };
```
ZeldaZach, line 37: Drive-by: now that the field is gone and the constructor no longer takes a `CardDatabase &`, this header no longer references `CardDatabase` at all — the `#include <libcockatrice/card/database/card_database.h>` near the top (line 10) is unused and can be removed.

cockatrice/src/client/network/interfaces/tapped_out_interface.h
```diff
@@ -29,14 +29,13 @@ class TappedOutInterface : public QObject
...
 
-    CardDatabase &cardDatabase;
     void copyDeckSplitMainAndSide(const DeckList &source, DeckList &mainboard, DeckList &sideboard);
 private slots:
     void queryFinished(QNetworkReply *reply);
     void getAnalyzeRequestData(const DeckList &deck, QByteArray &data);
 
 public:
-    explicit TappedOutInterface(CardDatabase &_cardDatabase, QObject *parent = nullptr);
+    explicit TappedOutInterface(QObject *parent = nullptr);
```
ZeldaZach, line 38: Same note as in `deck_stats_interface.h`: the `#include <libcockatrice/card/database/card_database.h>` at line 10 is no longer referenced from this header and can be dropped.

ZeldaZach (commented):
Nice cleanup — one small follow-up: both headers no longer reference `CardDatabase`, so the `#include <libcockatrice/card/database/card_database.h>` directive at the top of each header is now unused and can be removed. The `.cpp` files already pull in `card_database_manager.h` directly. (Not blocking — flagging since the PR's stated goal is to drop the `CardDatabase` dependency from these classes.)

ZeldaZach (approved):

PR #6964, opened 2026-06-03
[TabDeckEditor] Refactor: pull up showPrintingSelector

RickyRister:
## Related Ticket(s)
- Follow-up to #6961
- Follow-up to #6962

The code for the virtual method `showPrintingSelector` is the exact same in both subclasses, and I don't see how the code would ever be different.

- Make `showPrintingSelector` non-virtual and pull the implementation into the abstract class
- Correct documentation


cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
```diff
@@ -588,3 +588,10 @@ bool AbstractTabDeckEditor::closeRequest()
     }
     return close();
 }
+
+void AbstractTabDeckEditor::showPrintingSelector()
+{
+    printingSelectorDockWidget->printingSelector->setCard(cardInfoDockWidget->cardInfo->getCard().getCardPtr());
```
ZeldaZach, line 594: Nit: missing trailing newline (the diff shows `\ No newline at end of file`). Add a newline after the closing `}`.

ZeldaZach (commented):
Pull-up looks correct — both subclass implementations were byte-identical, all callers already used the `AbstractTabDeckEditor*` type, and both `printingSelectorDockWidget` and `cardInfoDockWidget` are base-class members (abstract_tab_deck_editor.h:131, :134). The existing `connect(..., &AbstractTabDeckEditor::showPrintingSelector)` at deck_editor_deck_dock_widget.cpp:722 keeps working since Qt5 function-pointer connects don't require the target to be a `slot`. One nit inline.

ZeldaZach (approved):

PR #6965, opened 2026-06-03
[TabDeckEditor] Refactor to use signal instead of calling tab

RickyRister:
## Related Ticket(s)
- Follow-up to #6961
- Follow-up to #6962

I need to make `DeckEditorDatabaseDisplayWidget` use signals instead of directly calling the deck editor tab, in order to set up for the main refactor I'm planning

- Add two more slot methods to `AbstractTabDeckEditor`
- Add signals to `DeckEditorDatabaseDisplayWidget`
  - Use those signals instead of directly calling the deckEditor tab

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_editor_database_display_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_editor_database_display_widget.cpp
@@ -222,10 +222,11 @@ void DeckEditorDatabaseDisplayWidget::databaseCustomMenu(QPoint point)
         } else {
             for (const CardRelation *rel : relatedCards) {
                 const QString &relatedCardName = rel->getName();
-                ExactCard exactCard = CardDatabaseManager::query()->getPreferredCard(relatedCardName);
                 QAction *relatedCard = relatedMenu->addAction(relatedCardName);
-                connect(relatedCard, &QAction::triggered, this,
-                        [this, exactCard] { emit cardInfoRequested(exactCard); });
+                connect(relatedCard, &QAction::triggered, this, [this, relatedCardName] {
+                    ExactCard card = CardDatabaseManager::query()->guessCard({relatedCardName});
+                    emit cardInfoRequested(card);
+                });
             }
         }
         menu.exec(databaseView->mapToGlobal(point));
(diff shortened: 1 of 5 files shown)


cockatrice/src/interface/widgets/deck_editor/deck_editor_database_display_widget.cpp
```diff
@@ -222,10 +222,10 @@ void DeckEditorDatabaseDisplayWidget::databaseCustomMenu(QPoint point)
         } else {
             for (const CardRelation *rel : relatedCards) {
                 const QString &relatedCardName = rel->getName();
+                ExactCard exactCard = CardDatabaseManager::query()->getPreferredCard(relatedCardName);
```
ZeldaZach: Behavior change worth thinking about: the previous code called `cardInfo->setCard(relatedCardName)`, which routes through `CardInfoFrameWidget::setCard(QString)` → `CardDatabaseManager::query()->guessCard({cardName})`. `guessCard` falls back through `getCardBySimpleName` and `simplifyName` (see card_database_querier.cpp's `lookupCardByName`). `getPreferredCard(QString)` doesn't have that fallback — it just calls `getCardInfo(name)`. So for related-card names that don't match exactly (apostrophes, diacritics, formatting quirks), the new path can resolve to an empty `ExactCard` where the old path would have found the card.

Also minor: this lookup now runs once per related card at menu-construction time rather than lazily on click. Usually negligible, but worth noting if a card can have many related entries.

cockatrice/src/interface/widgets/deck_editor/deck_editor_database_display_widget.h
```diff
@@ -52,6 +52,10 @@ public slots:
     void cardDecremented(const ExactCard &card, const QString &zoneName);
     void cardChanged(const ExactCard &_card);
 
+    void edhrecRequested(const CardInfoPtr &cardInfo, bool isCommander);
+    void printingSelectorRequested();
+    void cardInfoRequested(const ExactCard &cardName);
```
ZeldaZach: Nit: the parameter is named `cardName` but its type is `ExactCard` — confusing on the receiving side. Suggest `card` (and update `AbstractTabDeckEditor::updateCardInfo`'s slot signature too, which already uses `card`).

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.h
```diff
@@ -175,6 +182,9 @@ public slots:
     /** @brief Shows the printing selector dock. Pure virtual. */
     virtual void showPrintingSelector() = 0;
 
+    /** @brief Opens the edhrec tab with the following values */
```
ZeldaZach: Nit: `Opens the edhrec tab with the following values` trails off and doesn't describe behavior. Suggest something like `Opens an EDHRec tab for the given card` with `@param info` / `@param isCommander` lines to match the docs on neighboring methods (e.g. `addCard`).

ZeldaZach (commented):
Direction is good — decoupling `DeckEditorDatabaseDisplayWidget` from `AbstractTabDeckEditor` via signals is a clean intermediate step toward the larger refactor. A few observations inline; the lookup-semantics one is the only one I'd consider a real concern.

ZeldaZach (approved):

PR #6966, opened 2026-06-03
[Messages] Add option to ignore private messages from non-buddy users

### LINKED ISSUE #6915: Block private messages from non-buddies
I wish for a feature where you can prevent all private messages from non-buddies

kongwu666:
## Related Ticket(s)
- Fixes #6915

Users currently have no way to block private messages from non-buddy users. There is an existing option to ignore messages from unregistered users, but no equivalent for non-buddies.

- Add `ignoreNonBuddyUserMessages` setting following the existing `ignoreUnregisteredUserMessages` pattern
- Add a checkbox in Settings → Chat to toggle this option
- Filter non-buddy private messages in `TabSupervisor::processUserMessageEvent()`

## Files Changed
- `cockatrice/src/client/settings/cache_settings.h` — Add member variable, getter, setter, signal
- `cockatrice/src/client/settings/cache_settings.cpp` — Add setting initialization, persistence, setter implementation
- `cockatrice/src/interface/widgets/settings_page/messages_settings_page.h` — Add QCheckBox
- `cockatrice/src/interface/widgets/settings_page/messages_settings_page.cpp` — Add UI initialization, signal connection, layout, text
- `cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp` — Add filter logic in processUserMessageEvent

## Testing
- Compiled successfully with Qt 6.4.2
- GUI verified: checkbox appears correctly in Settings → Chat
- Functional test: with the option enabled, private messages from non-buddy users are ignored; with the option disabled, messages are received normally

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
@@ -1012,6 +1012,10 @@ void TabSupervisor::processUserMessageEvent(const Event_UserMessage &event)
         tab = messageTabs.value(QString::fromStdString(event.receiver_name()));
     }
     if (!tab) {
+        if (SettingsCache::instance().getIgnoreNonBuddyUserMessages() && !userListManager->isUserBuddy(senderName)) {
+            // When  he is not a buddy and the settings are set to block non friend messages
+            return;
+        }
         const ServerInfo_User *onlineUserInfo = userListManager->getOnlineUser(senderName);
         if (onlineUserInfo) {
             auto userLevel = UserLevelFlags(onlineUserInfo->user_level());
--- a/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
@@ -1012,16 +1012,18 @@ void TabSupervisor::processUserMessageEvent(const Event_UserMessage &event)
         tab = messageTabs.value(QString::fromStdString(event.receiver_name()));
     }
     if (!tab) {
-        if (SettingsCache::instance().getIgnoreNonBuddyUserMessages() && !userListManager->isUserBuddy(senderName)) {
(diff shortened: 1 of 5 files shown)


BruebachL (approved):

BruebachL:
Very simple addition which looks good to me, thanks for this little QoL fix :)

RickyRister:
Note that this setting is entirely server-side. The message sender (and the server in general) will have no idea that the message is blocked.

Is that desired?

kongwu666:
> Note that this setting is entirely server-side. The message sender (and the server in general) will have no idea that the message is blocked.
> 
> Is that desired?

Yes, this follows the existing ignore pattern. The issue title says "block" but ebbit1q suggested using "ignore" instead since "block" in Cockatrice means server-side with sender notification. The ignore list and ignoreUnregisteredUserMessages also silently filter without notifying the sender.

BruebachL:
I believe this is ticketed as "to fix" somewhere.

RickyRister (approved):

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -1012,6 +1012,10 @@ void TabSupervisor::processUserMessageEvent(const Event_UserMessage &event)
         tab = messageTabs.value(QString::fromStdString(event.receiver_name()));
     }
     if (!tab) {
+        if (SettingsCache::instance().getIgnoreNonBuddyUserMessages() && !userListManager->isUserBuddy(senderName)) {
```
ZeldaZach: Moderator messages should not be ignorable with this setting by default

kongwu666: Done. Moderator and Admin messages are now exempt from the filter. I followed the same pattern as the existing ignoreUnregisteredUsers check above.

ZeldaZach (approved):

PR #6967, opened 2026-06-04
[TabDeckEditor] Refactor card database view into own class

RickyRister:
## Related Ticket(s)
- Follow-up to #6965
- Follow-up to #6939

The big refactor that all my little refactors were building up to.

Right now, most code just directly reaches into the `DeckEditorDatabaseDisplayWidget` instance in the deck editor and hooks itself up to the database table view. This PR attempts to extract the card database table view into its own reusable class. 

- Extract card database table into its own `QTreeView` subclass
  - Manages clicks, keyboard shortcuts, and the right-click menu
  - All interactions are exposed through signals, so the client can reuse the same class in different places
  - Is an `eventFilter` so that you can still attach it to the textEdit to get it to capture existing shortcuts
- Instantiate `CardDatabaseView` every time it needs to be used, instead of reusing the one in `DeckEditorDatabaseDisplayWidget`
  - Also create and wire up signals so we can maintain the old behavior

### DIFF
--- a/cockatrice/src/interface/key_signals.h
+++ b/cockatrice/src/interface/key_signals.h
@@ -29,7 +29,7 @@ signals:
     void onShiftS();
     void onCtrlC();
 
-protected:
+public:
     bool eventFilter(QObject *, QEvent *event) override;
 };
 
--- a/cockatrice/src/interface/key_signals.h
+++ b/cockatrice/src/interface/key_signals.h
@@ -29,7 +29,7 @@ signals:
     void onShiftS();
     void onCtrlC();
 
-public:
+protected:
     bool eventFilter(QObject *, QEvent *event) override;
 };
 
--- a/cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
@@ -1,0 +1,172 @@
+#include "card_database_view.h"
+
+#include "../../../client/settings/cache_settings.h"
+#include "card_database_display_model.h"
+#include "card_database_model.h"
+
+#include <QApplication>
+#include <QClipboard>
+#include <QHeaderView>
+#include <QMenu>
+#include <libcockatrice/card/database/card_database_manager.h>
+#include <libcockatrice/card/relation/card_relation.h>
+#include <libcockatrice/deck_list/tree/inner_deck_list_node.h>
+
+static bool canBeCommander(const CardInfo &cardInfo)
+{
+    return (cardInfo.getCardType().contains("Legendary", Qt::CaseInsensitive) &&
+            cardInfo.getCardType().contains("Creature", Qt::CaseInsensitive)) ||
+           cardInfo.getText().contains("can be your commander", Qt::CaseInsensitive);
+}
+
+CardDatabaseView::CardDatabaseView(QWidget *parent, CardDatabaseDisplayModel *model)
+    : QTreeView(parent), databaseDisplayModel(model)
+{
+    // set up object
+    setUniformRowHeights(true);
+    setRootIsDecorated(false);
+    setAlternatingRowColors(true);
+    setSortingEnabled(true);
+    sortByColumn(0, Qt::AscendingOrder);
+    QTreeView::setModel(databaseDisplayModel);
+    setContextMenuPolicy(Qt::CustomContextMenu);
+
+    connect(databaseDisplayModel, &CardDatabaseDisplayModel::modelDirty, this,
+            &CardDatabaseView::resetSelectionIfEmpty);
+
+    connect(this, &QTreeView::customContextMenuRequested, this, &CardDatabaseView::openCustomMenu);
+    connect(selectionModel(), &QItemSelectionModel::currentRowChanged, this, &CardDatabaseView::updateCard);
+    connect(this, &QTreeView::doubleClicked, this, &CardDatabaseView::actDoubleClick);
+
+    // layout settings
+    QByteArray dbHeaderState = SettingsCache::instance().layouts().getDeckEditorDbHeaderState();
+    if (dbHeaderState.isNull()) {
+        // first run
+        setColumnWidth(0, 200);
+    } else {
+        header()->restoreState(dbHeaderState);
+    }
+    connect(header(), &QHeaderView::geometriesChanged, this, &CardDatabaseView::saveDbHeaderState);
+
+    // create key filters
+    searchKeySignals.setObjectName("searchKeySignals");
+    connect(&searchKeySignals, &KeySignals::onEnter, this, [this] { addCard(DECK_ZONE_MAIN); });
+    connect(&searchKeySignals, &KeySignals::onCtrlAltEqual, this, [this] { addCard(DECK_ZONE_MAIN); });
+    connect(&searchKeySignals, &KeySignals::onCtrlAltRBracket, this, [this] { addCard(DECK_ZONE_SIDE); });
+    connect(&searchKeySignals, &KeySignals::onCtrlAltMinus, this, [this] { decrementCard(DECK_ZONE_MAIN); });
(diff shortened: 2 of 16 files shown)


ZeldaZach (commented):
test

ZeldaZach (commented):
(replaced — collapsed into the single review below)

ZeldaZach (commented):
(replaced — collapsed into the single review below)

ZeldaZach (commented):
(replaced — collapsed into the single review below)

ZeldaZach (commented):
(replaced — collapsed into the single review below)

cockatrice/src/interface/widgets/tabs/tab_visual_database_display.cpp
```diff
@@ -1,14 +1,19 @@
...
 {
-    deckEditor = new TabDeckEditor(_tabSupervisor);
-    deckEditor->setHidden(true);
-    visualDatabaseDisplayWidget = new VisualDatabaseDisplayWidget(
-        this, deckEditor, deckEditor->cardDatabaseDockWidget->databaseDisplayWidget->databaseModel,
-        deckEditor->cardDatabaseDockWidget->databaseDisplayWidget->databaseDisplayModel);
+    auto databaseModel = new CardDatabaseModel(CardDatabaseManager::getInstance(), true, this);
+    databaseModel->setObjectName("databaseModel");
+
+    visualDatabaseDisplayWidget = new VisualDatabaseDisplayWidget(this, deckEditor, databaseModel);
```
ZeldaZach: **Bug: `deckEditor` is used uninitialized here.**

The previous version of this constructor did `deckEditor = new TabDeckEditor(_tabSupervisor); deckEditor->setHidden(true);` before passing it as an argument. That initialization was removed in this PR, but `deckEditor` is still passed to `VisualDatabaseDisplayWidget` (which stores it and exposes it via `getDeckEditor()`).

The member is a raw `TabDeckEditor *` with no in-class initializer, so its value is indeterminate. The pointer eventually reaches `VisualDatabaseDisplayNameFilterWidget::actLoadFromDeck`, which does `deckEditor->deckStateManager->getModel()` — that path will dereference garbage and crash.

Options: restore the hidden `TabDeckEditor`, pass `nullptr` (and null-check in the consumers that need it), or refactor `VisualDatabaseDisplayWidget` so the standalone-database-tab case no longer requires an `AbstractTabDeckEditor *`.

cockatrice/src/interface/widgets/deck_editor/card_database_view.h
```diff
@@ -0,0 +1,53 @@
+#ifndef COCKATRICE_CARD_DATABASE_TABLE_WIDGET_H
+#define COCKATRICE_CARD_DATABASE_TABLE_WIDGET_H
```
ZeldaZach: Header guard `COCKATRICE_CARD_DATABASE_TABLE_WIDGET_H` doesn't match the filename `card_database_view.h`. Worth renaming (e.g. `COCKATRICE_CARD_DATABASE_VIEW_H`) so the guard stays in sync with the file.

cockatrice/src/interface/widgets/deck_editor/card_database_view.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "../../key_signals.h"
+#include "libcockatrice/filters/filter_tree.h"
```
ZeldaZach: `filter_tree.h` doesn't appear to be referenced anywhere in this header — no `FilterTree`, no filter types in the signal/slot signatures. Looks like a leftover include that can be dropped.

ZeldaZach (commented):
Nice refactor — extracting `CardDatabaseView` cleans up `DeckEditorDatabaseDisplayWidget` significantly and makes the table genuinely reusable. I found one critical bug in `TabVisualDatabaseDisplay` (uninitialized pointer reaching a dereference path) plus a few smaller cleanups noted inline.

(Apologies for the four no-op reviews above — those were from probing line numbers via the API.)

cockatrice/src/interface/key_signals.h
```diff
@@ -29,7 +29,7 @@ class KeySignals : public QObject
     void onShiftS();
     void onCtrlC();
 
-protected:
+public:
```
ZeldaZach: Promoting `eventFilter` from `protected` to `public` just so `CardDatabaseView::eventFilter` can delegate to it leaks an implementation detail of `KeySignals`. Since `KeySignals` is already a `QObject` member of `CardDatabaseView`, you could instead install `searchKeySignals` directly as the event filter on the search edit (either via an accessor on `CardDatabaseView` or by having the view call `searchEdit->installEventFilter(&searchKeySignals)` itself). That keeps the access modifier protected and removes the trivial pass-through override in `CardDatabaseView`.

cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
```diff
@@ -0,0 +1,172 @@
...
+                    [this, relatedCardName] { emit relatedCardClicked(relatedCardName); });
+        }
+    }
+
+    menu.exec(mapToGlobal(point));
+}
+
+bool CardDatabaseView::eventFilter(QObject *object, QEvent *event)
+{
+    return searchKeySignals.eventFilter(object, event);
```
ZeldaZach: Two small things on this override:

1. The result of `searchKeySignals.eventFilter(...)` is returned directly, so any event the key-signals filter declines (returns false) skips `QTreeView::eventFilter` entirely. Today the filter is only installed on the search line edit, so this is fine — but `return searchKeySignals.eventFilter(object, event) || QTreeView::eventFilter(object, event);` would future-proof it if anyone later installs the view as a filter on another widget.
2. Missing trailing newline at end of file (also in `deck_editor_database_display_widget.cpp` and `tab_visual_database_display.cpp`).

cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.cpp
```diff
@@ -63,9 +65,10 @@ void TabDeckEditorVisual::createCentralFrame()
     centralFrame = new QVBoxLayout;
     centralWidget->setLayout(centralFrame);
 
-    tabContainer = new TabDeckEditorVisualTabWidget(
-        centralWidget, this, deckStateManager->getModel(), cardDatabaseDockWidget->databaseDisplayWidget->databaseModel,
-        cardDatabaseDockWidget->databaseDisplayWidget->databaseDisplayModel);
+    auto databaseModel = new CardDatabaseModel(CardDatabaseManager::getInstance(), true, this);
+    databaseModel->setObjectName("databaseModel");
```
ZeldaZach, line 69: Minor: `TabDeckEditorVisual` now constructs its own `CardDatabaseModel` here, while `DeckEditorCardDatabaseDockWidget` (also owned by this tab) creates another one internally. That's two full `CardDatabaseModel` instances wrapping the same `CardDatabaseManager` per editor tab. Not strictly broken, but worth considering whether the dock widget's model could be shared, or whether the model should be owned higher up and passed down.

ZeldaZach (approved):

cockatrice/src/interface/widgets/tabs/visual_deck_editor/tab_deck_editor_visual.cpp
RickyRister: Yeah, you're right, we can create a shared CardDatabaseModel in the parent class.
Split it into #6968 since it's a change that can be merged first.

cockatrice/src/interface/widgets/tabs/tab_visual_database_display.cpp
RickyRister: Split into #6969 , where I refactored it so the `VisualDatabaseDisplayWidget` takes a nullable `DeckListModel` instead of the entire tab

Issue #6489, opened 2026-01-03
Cannot add token to tokens section of decklist on 2.11 Beta 41 or above


SlightlyCircuitous:
**System Information:**

Client Version: 2.11.0 () [Beta 43 compiled from source]
Client Operating System: Ubuntu 25.10
Build Architecture: 64-bit
Qt Version: 6.9.2
_______________________________________________________________________________________

Since beta 41, I am unable to add tokens to the token section of the decklist in either the VDE or legacy deck editor. When I try to add a token, it appears in the main decklist section under the appropriate card type heading instead. This occurs when clicking on a token in the VDE, using the right click option in the VDE table display, and when using the green arrow in the legacy deck editor. Using the blue arrow in the legacy deck editor adds it to the sideboard instead of the tokens section.

<img width="455" height="204" alt="Image" src="https://github.com/user-attachments/assets/939341f8-ed92-43d9-97b1-2e8acb84c4df" />

Creature token added to main decklist, under the Creature heading

I've tested this in betas 36 through 43 and the problem appears for me in 41 and persists through the subsequent betas. Please let me know if I can provide further information or testing. Thank you for all your hard work so far!

_______________________________________________________________________________________

**Steps to reproduce:**
 - Be on cockatrice 2.11 beta 41 or higher
 - Create a new deck or open an existing deck in the VDE or legacy deck editor
 - Add a token to the deck using any method (tested: clicking picture in VDE, using the green or blue arrow in legacy, right-click option in VDE table display)

SlightlyCircuitous:
Thank you!!

PR #6971, opened 2026-06-05
[DeckEditor] Replace mainboard/sideboard with tokensboard for tokens (#6546)

### LINKED ISSUE #6546: Have Printing Selector add Tokens to Token Section
The Printing Selector is a wonderful new feature, which now can display unique printings of tokens (https://github.com/Cockatrice/Magic-Token/pull/340). 

However, as far as I can tell, there isn't a way to actually add the printing you want to a deck. The options displayed on each printing in the Printing Selector are `+/- Mainboard` and `+/- Sideboard`, which, as far as I can tell, can only add a token printing to the `Maindeck` or `Sideboard` sections respectively, not the `Tokens` section. Clicking on the token picture or name in the VDD list (or using the arrows in the classic editor) can only add the most recent printing of the token.

It would be fantastic to have a `+/- Token` option in the printing selector instead of the mainboard and sideboard ones for anything that should be listed in the `Tokens` section. Thank you!

kongwu666:
## Related Ticket(s)

- Fixes #6546

Token cards in the printing selector show mainboard and sideboard +/- buttons, but tokens should only exist in the tokens zone per project design (see #6489, #6495). Additionally, tokens in the tokensboard zone would be incorrectly swapped to the mainboard on double-click in the visual deck editor.

- Replace mainboard/sideboard +/- buttons with tokensboard +/- button for token cards in the printing selector
- Token cards show only a tokensboard label and +/- buttons; non-token cards show mainboard/sideboard as before
- `tallyUuidCounts` return type changed from `QMap<QString, QPair<int,int>>` to `QMap<QString, std::tuple<int,int,int>>` to include tokens zone counts
- Fix `swapCardAtIndex` to keep tokens in tokensboard instead of swapping them to mainboard on double-click
- Added `addPrintingTokensboard()` / `removePrintingTokensboard()` slots to `CardAmountWidget`

## Screenshots

Token printing selector

 

Non-token printing selector

    

Classic editor tokens

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
@@ -255,10 +255,11 @@ bool DeckStateManager::swapCardAtIndex(const QModelIndex &idx)
     }
 
     QString zoneName = gparent.siblingAtColumn(DeckListModelColumns::CARD_NAME).data(Qt::EditRole).toString();
-    QString otherZoneName = zoneName == DECK_ZONE_TOKENS
-                                ? DECK_ZONE_TOKENS
-                                : (zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN);
-    // if card is Token,it should be swap to DECK_ZONE_TOKENS,not DECK_ZONE_MAIN or DECK_ZONE_SIDE.
+    // tokens have no swap target
+    if (zoneName == DECK_ZONE_TOKENS) {
+        return false;
+    }
+    QString otherZoneName = zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN;
 
     QString reason = tr("Moved to %1 1 × \"%2\" (%3)") //
                          .arg(otherZoneName)
--- a/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
@@ -255,7 +255,10 @@ bool DeckStateManager::swapCardAtIndex(const QModelIndex &idx)
     }
 
     QString zoneName = gparent.siblingAtColumn(DeckListModelColumns::CARD_NAME).data(Qt::EditRole).toString();
-    QString otherZoneName = zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN;
+    QString otherZoneName = zoneName == DECK_ZONE_TOKENS
+                                ? DECK_ZONE_TOKENS
+                                : (zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN);
+    // if card is Token,it should be swap to DECK_ZONE_TOKENS,not DECK_ZONE_MAIN or DECK_ZONE_SIDE.
 
     QString reason = tr("Moved to %1 1 × \"%2\" (%3)") //
                          .arg(otherZoneName)
--- a/cockatrice/src/interface/widgets/printing_selector/all_zones_card_amount_widget.cpp
+++ b/cockatrice/src/interface/widgets/printing_selector/all_zones_card_amount_widget.cpp
@@ -31,15 +31,15 @@ AllZonesCardAmountWidget::AllZonesCardAmountWidget(QWidget *parent,
     buttonBoxMainboard = new CardAmountWidget(this, deckStateManager, cardSizeSlider, rootCard, DECK_ZONE_MAIN);
     zoneLabelSideboard = new ShadowBackgroundLabel(this, tr("Sideboard"));
     buttonBoxSideboard = new CardAmountWidget(this, deckStateManager, cardSizeSlider, rootCard, DECK_ZONE_SIDE);
-    zoneLabelTokensboard = new ShadowBackgroundLabel(this, tr("Tokensboard"));
+    zoneLabelTokensboard = new ShadowBackgroundLabel(this, tr("Tokens"));
     buttonBoxTokensboard = new CardAmountWidget(this, deckStateManager, cardSizeSlider, rootCard, DECK_ZONE_TOKENS);
 
     layout->addWidget(zoneLabelMainboard, 0, Qt::AlignHCenter | Qt::AlignBottom);
     layout->addWidget(buttonBoxMainboard, 0, Qt::AlignHCenter | Qt::AlignTop);
-    layout->addSpacing(12.5);
+    layout->addSpacing(12);
     layout->addWidget(zoneLabelTokensboard, 0, Qt::AlignHCenter | Qt::AlignBottom);
     layout->addWidget(buttonBoxTokensboard, 0, Qt::AlignHCenter | Qt::AlignTop);
-    layout->addSpacing(12.5);
+    layout->addSpacing(13);
     layout->addWidget(zoneLabelSideboard, 0, Qt::AlignHCenter | Qt::AlignBottom);
     layout->addWidget(buttonBoxSideboard, 0, Qt::AlignHCenter | Qt::AlignTop);
 
--- a/cockatrice/src/interface/widgets/printing_selector/all_zones_card_amount_widget.cpp
+++ b/cockatrice/src/interface/widgets/printing_selector/all_zones_card_amount_widget.cpp
@@ -8,7 +8,7 @@
  * @brief Constructor for the AllZonesCardAmountWidget class.
  *
  * Initializes the widget with its layout and sets up the connections and necessary
- * UI elements for managing card counts in both the mainboard and sideboard zones.
+ * UI elements for managing card counts in all the mainboard, tokensboard and sideboard zones.
  *
  * @param parent The parent widget.
  * @param deckStateManager Pointer to the DeckStateManager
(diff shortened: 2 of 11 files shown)


cockatrice/src/interface/widgets/printing_selector/printing_selector.cpp
```diff
@@ -105,23 +105,30 @@ void PrintingSelector::printingsInDeckChanged()
 }
 
 /**
- * @return A map of uuid to amounts (main, side).
+ * @return A map of uuid to amounts (main, side, tokens).
  */
-static QMap<QString, QPair<int, int>> tallyUuidCounts(const DeckListModel *model, const QString &cardName)
+static QMap<QString, std::tuple<int, int, int>> tallyUuidCounts(const DeckListModel *model, const QString &cardName)
```
RickyRister: Once we get to a 3-tuple, we should just refactor it into a struct so we don't have to memorize which index corresponds to what.

kongwu666: Good point, I'll refactor the tuple into a struct for readability.

kongwu666: refactored the tuple into a ZoneCounts struct. See 534704b8eaff2b41295d88cd1273a0f03797c7ce.

cockatrice/src/interface/widgets/printing_selector/printing_selector.h
```diff
@@ -22,6 +22,13 @@
 
 #define BATCH_SIZE 10
 
+struct ZoneCounts
```
RickyRister, line 25: I think it would make more sense to put this struct in the `AllZonesCardAmount` widget or in a separate file, so that it can be used across the printing selector code.
However, that can be part of a separate refactor. For now it's fine to just stick it here.

cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
```diff
@@ -183,14 +186,15 @@ void CardAmountWidget::addPrinting(const QString &zone)
...
-                         .arg(zone == DECK_ZONE_MAIN ? "mainboard" : "sideboard")
-                         .arg(rootCard.getPrinting().getUuid())
-                         .arg(replacingProviderless ? " (replaced providerless printings)" : "");
+    QString reason =
+        QString("Added %1 copies of '%2 (%3) %4' to %5 [ProviderID: %6]%7")
+            .arg(1 + extraCopies)
+            .arg(rootCard.getName())
+            .arg(rootCard.getPrinting().getSet()->getShortName())
+            .arg(rootCard.getPrinting().getProperty("num"))
+            .arg(zone == DECK_ZONE_MAIN ? "mainboard" : (zone == DECK_ZONE_SIDE ? "sideboard" : "tokensboard"))
```
RickyRister: It's kind of hard to read inline. Can we extract this to a variable or function?

cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
```diff
@@ -234,19 +246,28 @@ void CardAmountWidget::removePrintingSideboard()
...
-                         .arg(rootCard.getPrinting().getSet()->getShortName())
-                         .arg(rootCard.getPrinting().getProperty("num"))
-                         .arg(zone == DECK_ZONE_MAIN ? "mainboard" : "sideboard")
-                         .arg(rootCard.getPrinting().getUuid());
+    QString reason =
+        QString("Removed 1 copy of '%1 (%2) %3' from %4 [ProviderID: %5]")
+            .arg(rootCard.getName())
+            .arg(rootCard.getPrinting().getSet()->getShortName())
+            .arg(rootCard.getPrinting().getProperty("num"))
+            .arg(zone == DECK_ZONE_MAIN ? "mainboard" : (zone == DECK_ZONE_SIDE ? "sideboard" : "tokensboard"))
```
RickyRister: also extract it here

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
```diff
@@ -255,7 +255,10 @@ bool DeckStateManager::swapCardAtIndex(const QModelIndex &idx)
     }
 
     QString zoneName = gparent.siblingAtColumn(DeckListModelColumns::CARD_NAME).data(Qt::EditRole).toString();
-    QString otherZoneName = zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN;
+    QString otherZoneName = zoneName == DECK_ZONE_TOKENS
+                                ? DECK_ZONE_TOKENS
+                                : (zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN);
+    // if card is Token,it should be swap to DECK_ZONE_TOKENS,not DECK_ZONE_MAIN or DECK_ZONE_SIDE.
```
RickyRister: nit: can you move this comment above the line it's describing instead of below?

RickyRister (approved):
LGTM. Thanks.

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
```diff
@@ -255,7 +255,10 @@ bool DeckStateManager::swapCardAtIndex(const QModelIndex &idx)
     }
 
     QString zoneName = gparent.siblingAtColumn(DeckListModelColumns::CARD_NAME).data(Qt::EditRole).toString();
-    QString otherZoneName = zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN;
+    // if card is Token,it should be swap to DECK_ZONE_TOKENS,not DECK_ZONE_MAIN or DECK_ZONE_SIDE.
+    QString otherZoneName = zoneName == DECK_ZONE_TOKENS
+                                ? DECK_ZONE_TOKENS
+                                : (zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN);
```
ZeldaZach: This sets `otherZoneName = DECK_ZONE_TOKENS` when the source zone is already tokens, which means double-clicking a token still:

- enters `modifyDeck(...)`, which records an undo entry with reason `"Moved to tokens 1 × \"<name>\" (<id>)"` — misleading, since nothing moved,
- calls `doSwapCard(model, idx, ..., "tokens")` which decrements the source then re-adds to the same zone (a net no-op, but it touches the model and emits row signals).

Cleaner to bail out before any of that runs, e.g.:

```cpp
if (zoneName == DECK_ZONE_TOKENS) {
    return false; // tokens have no swap target
}
QString otherZoneName = zoneName == DECK_ZONE_MAIN ? DECK_ZONE_SIDE : DECK_ZONE_MAIN;
```

That also lets the new comment on line 258 go away — and as a separate nit, that comment is missing spaces after both commas and reads `be swap` → `be swapped`.

cockatrice/src/interface/widgets/printing_selector/all_zones_card_amount_widget.cpp
```diff
@@ -31,13 +31,28 @@ AllZonesCardAmountWidget::AllZonesCardAmountWidget(QWidget *parent,
     buttonBoxMainboard = new CardAmountWidget(this, deckStateManager, cardSizeSlider, rootCard, DECK_ZONE_MAIN);
     zoneLabelSideboard = new ShadowBackgroundLabel(this, tr("Sideboard"));
     buttonBoxSideboard = new CardAmountWidget(this, deckStateManager, cardSizeSlider, rootCard, DECK_ZONE_SIDE);
+    zoneLabelTokensboard = new ShadowBackgroundLabel(this, tr("Tokensboard"));
```
ZeldaZach: The codebase already has a canonical user-facing translation for this zone: `InnerDecklistNode::visibleNameFromName()` returns `QObject::tr("Tokens")` for `DECK_ZONE_TOKENS` (see `libcockatrice_deck_list/.../inner_deck_list_node.cpp:29`). Introducing `tr("Tokensboard")` here creates a second translation key for the same concept — translators have to translate both, and the printing selector will diverge from the deck tree ("Tokens" everywhere else, "Tokensboard" here).

Suggest `tr("Tokens")` to match the existing string. (The Mainboard/Sideboard precedent is local to this widget; the project's pre-existing token name is "Tokens".)

cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
```diff
@@ -183,12 +186,13 @@ void CardAmountWidget::addPrinting(const QString &zone)
         }
     }
 
+    QString zoneName = zone == DECK_ZONE_MAIN ? "mainboard" : (zone == DECK_ZONE_SIDE ? "sideboard" : "tokensboard");
```
ZeldaZach: Internal name nit: `DECK_ZONE_TOKENS` is `"tokens"` (singular, no `-board` suffix — see `inner_deck_list_node.h:26`). Using `"tokensboard"` here as the human-readable in the log message diverges from the canonical name used by the rest of the system. Suggest `"tokens"` (or even use `DECK_ZONE_TOKENS` directly).

Same applies to the identical ternary on line 264 — and the two could share a tiny `zoneLogName(const QString&)` helper to avoid duplicating the mapping.

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_display_widget.h
```diff
@@ -13,6 +13,7 @@
 
 #include <QWidget>
 #include <libcockatrice/models/deck_list/deck_list_model.h>
+#include <tuple>
```
ZeldaZach: `#include <tuple>` is no longer needed — the PR description mentions `std::tuple<int,int,int>` but the actual implementation switched to a `ZoneCounts` struct (defined in `printing_selector.h`). This include can be dropped.

ZeldaZach (commented):
Nice fix overall. Main concern is the tokens "swap" path that still goes through `modifyDeck` for what becomes a no-op, plus the new `Tokensboard` label that doesn't match the existing token translation in the codebase. Detailed inline comments below.

ZeldaZach (approved):

cockatrice/src/interface/widgets/printing_selector/card_amount_widget.cpp
```diff
@@ -137,6 +140,17 @@ void CardAmountWidget::updateCardCount()
     layout->activate();
 }
 
+static QString zoneLogName(const QString &zone)
+{
+    if (zone == DECK_ZONE_MAIN) {
+        return "mainboard";
+    } else if (zone == DECK_ZONE_SIDE) {
+        return "sideboard";
+    } else {
```
RickyRister: you should still check for `DECK_ZONE_TOKENS`, in case we decide to add more zones in the future.
Have the fallback return "unknown" or something like that.

PR #6972, opened 2026-06-05
[Application] Implement in-game basic keyboard navigation

### LINKED ISSUE #5043: Add Keyboard-Only Controls for Card Selection and Game Actions
**Similar Requests**

None found.

I would like to request the implementation of a feature that allows users to play on Cockatrice entirely using the keyboard, without needing a mouse. Currently, the main challenge is selecting cards using only keyboard shortcuts. Introducing a set of keyboard shortcuts for card selection and any other game actions that aren't covered would greatly enhance accessibility and usability for keyboard-only users.

**Additional Context**

This feature would be particularly beneficial for users with limited mobility or those who prefer using the keyboard over the mouse. It would also improve the overall user experience by providing an alternative control method.

Vorliz:
## Related Ticket(s)
- Fixes #5043

We implemented keyboard navigation of the board cards. 

- Left and right arrow now switches between cards;
- Up and down arrow switches between zones;
- Space spawns the right click menu so further actions can be selected with the arrow keys.
- The selected card is highlighted the same way as the mouse hover action.

We thought the up and down arrows is intuitive for switching the zones, but we are open to changing the way it works if a better option is presented. The feature also includes tests of the arrow switching logic.


cockatrice/src/game_graphics/board/card_item.cpp
```diff
@@ -500,6 +504,71 @@ void CardItem::mouseDoubleClickEvent(QGraphicsSceneMouseEvent *event)
...
+            navigator->startArrowMode(this);
+            event->accept();
+            return;
+        }
+    }
+
+    if ((event->key() == Qt::Key_Return || event->key() == Qt::Key_Enter) && navigator &&
+        navigator->isArrowModeActiveVar() && getIsHovered()) {
+
+        qWarning() << "Finalizing arrow mode from CardItem with id:" << id;
```
ZeldaZach, line 536: These `qWarning()` calls (and the ones on 521, 523, 535) look like leftover debug prints — every Esc / Shift+A / Enter on a card while playing will write to stderr. Either remove them or downgrade to `qDebug` and guard behind a logging category, but they shouldn't ship as warnings.

cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -1153,6 +1155,14 @@ void TabGame::createPlayAreaWidget(bool bReplay)
     connect(scene, &GameScene::arrowDeletionRequested, game->getGameEventHandler(),
             &GameEventHandler::handleArrowDeletion);
     connect(game->getGameEventHandler(), &GameEventHandler::arrowDeleted, scene, &GameScene::deleteArrow);
+
+    keySignals = new KeySignals();
+    qApp->installEventFilter(keySignals);
+    connect(keySignals, &KeySignals::onLeftArrow, scene, &GameScene::handleLeftArrow);
+    connect(keySignals, &KeySignals::onRightArrow, scene, &GameScene::handleRightArrow);
+    connect(keySignals, &KeySignals::onUpArrow, scene, &GameScene::handleUpArrow);
+    connect(keySignals, &KeySignals::onDownArrow, scene, &GameScene::handleDownArrow);
```
ZeldaZach, line 1167: Two problems with this block:

1. `keySignals` is allocated with `new` but never deleted (`~TabGame` doesn't touch it), so it leaks on every tab close.
2. `qApp->installEventFilter(keySignals)` registers a process-global event filter that's never removed. Every TabGame you open installs *another* global filter; closing a tab leaves the filter installed forever, and N open tabs means N filters seeing every keypress.

Suggest making `keySignals` a child of `TabGame` (or wrapping in `std::unique_ptr`), and installing the filter on the local widget (`this` or `gameView`) rather than `qApp`, so it scopes to this tab and unregisters with the parent.

cockatrice/src/interface/key_signals.cpp
```diff
@@ -26,15 +27,35 @@ bool KeySignals::eventFilter(QObject * /*object*/, QEvent *event)
...
+                return false;
+            }
+            emit onUpArrow();
+            return true;
+        case Qt::Key_Down:
+            if (qApp->activePopupWidget() != nullptr || qApp->activeModalWidget() != nullptr) {
+                return false;
+            }
+            emit onDownArrow();
+            return true;
```
ZeldaZach, line 58: Combined with `installEventFilter(qApp)` over in `tab_game.cpp`, this swallows Up/Down arrows at the application level for as long as any game tab exists. The popup/modal exception is the only escape — meaning any QLineEdit, QListView, QTextEdit, QScrollArea, QTabBar, or focus-chain consumer of arrow keys anywhere in Cockatrice (chat, deck editor side panels, settings dialogs, the player list, etc.) stops getting Up/Down while a game is open.

The right fix is to scope this filter to the game tab (install on the GameView / scene's viewport) rather than globally, so the events only get intercepted when the game view itself has focus.

tests/keyboard_navigator_tests/keyboard_card_navigator_test.cpp
```diff
@@ -0,0 +1,229 @@
...
+// decided to use this dirty trick.
+#define private public
+#include "game/player/player_logic.h"
+#undef private
+
+#include "game/board/abstract_card_item.h"
+#include "game/board/arrow_item.h"
+#include "game/board/card_item.h"
+#include "game/board/card_list.h"
+#include "game/keyboard_card_navigator.cpp"
```
ZeldaZach, line 20: Two practices to avoid here:

1. `#define private public` is technically undefined behavior (the standard reserves the right to use access keywords as macro-replaceable, even via the standard library); also, it silently breaks any non-templated method whose name-mangling depends on access (rare but possible).
2. `#include "game/keyboard_card_navigator.cpp"` pulls the implementation directly into the test TU, defeating the link-against-library model the rest of the tree uses and making the test brittle to internal refactors.

Both point to the code not being designed for testability. Suggest exposing what the test needs via friend declarations (or a `friend class KeyboardCardNavigatorTest;`) and adding `keyboard_card_navigator.cpp` to a library target that the test links against.

ZeldaZach (commented):
Nice feature direction overall — keyboard navigation is a real UX win. A handful of structural/correctness concerns to address before this lands, see the inline comments.

tests/keyboard_navigator_tests/keyboard_card_navigator_test.cpp
```diff
@@ -0,0 +1,229 @@
...
+    FakeStackZoneLogic *stackZone;
+    PlayerLogic *player;
+
+    void SetUp() override
+    {
+        handZone = new FakeHandZoneLogic("hand");
+        tableZone = new FakeTableZoneLogic("table");
+        stackZone = new FakeStackZoneLogic("stack");
+        // Player is mocked like this, so we fill the zones for the tests.
+        player = (PlayerLogic *)malloc(sizeof(PlayerLogic));
```
ZeldaZach, line 42: This is undefined behavior on several axes: `PlayerLogic` is a `QObject` subclass, but `malloc` skips its constructor, leaving the vtable, signal/slot bookkeeping, parent/child trackers, and every other member uninitialized. Placement-newing just the `zones` member doesn't fix the rest. `free(player)` in `TearDown` then skips the destructor.

The test passing today is coincidental — it works because the code path being exercised happens not to touch the uninitialized parts. Either build a proper `PlayerLogic` (with whatever construction args it needs) or extract the bit you're testing into something that doesn't need a `PlayerLogic` at all.

cockatrice/src/game/player/player_logic.cpp
```diff
@@ -348,3 +352,20 @@ void PlayerLogic::setGameStarted()
...
+    HandZoneLogic *handZone = getHandZone();
+    if (!handZone) {
+        return;
+    }
+    const CardList &handCards = handZone->getCards();
+    if (!handCards.isEmpty()) {
+        CardItem *firstCard = handCards.at(0);
+        if (firstCard) {
+            firstCard->setHovered(true);
+            getGameScene()->getCardNavigator()->unhoverCard();
```
ZeldaZach, line 355: `getGameScene()` and `getCardNavigator()` are both dereferenced without checks. `setActive(true)` is called during normal game flow, but also in init order edge cases (replay setup, headless setups in tests) where the scene or its navigator may not exist yet. Suggest:

```cpp
if (auto *scene = getGameScene()) {
    if (auto *nav = scene->getCardNavigator()) {
        nav->unhoverCard();
        nav->setHoveredCardIndex(0);
    }
}
```

cockatrice/src/game/keyboard_card_navigator.h
```diff
@@ -0,0 +1,61 @@
...
+    int hoveredCardIndex = -1;
+    bool isArrowModeActive;
+    CardItem *arrowOriginCard;
+    ArrowItem *previewArrow;
+    PlayerLogic *playerLogic;
+    /**
+     * @brief Gets hand cards sorted by visual position.
+     * @return List of cards sorted by visual order (left-to-right for horizontal, top-to-bottom for vertical).
+     */
+    CardList getVisuallyOrderedHandCards() const;
```
ZeldaZach, line 33: `getVisuallyOrderedHandCards()` is declared here but never defined anywhere in the production tree (verified with `git grep`). Its only "implementation" is `#define getVisuallyOrderedHandCards() getCards()` in `keyboard_navigator_test_fakes.h`, used to make the tests compile. If it's not needed in production, drop the declaration; if it is, implement it.

cockatrice/src/game/keyboard_card_navigator.h
```diff
@@ -0,0 +1,61 @@
...
+     */
+    void setHoveredCardIndex(int index);
+    void unhoverCard();
+    void changeHoverCard(int cardIndex, bool hover);
+    void createArrow(CardItem *targetCard);
+    void createTempArrow(CardItem *targetCard);
+    void startArrowMode(CardItem *originCard);
+    void cancelArrowMode();
+    CardZoneLogic *findZoneWithCards(QList<CardZoneLogic *> &zonesList, int currentZoneIndex, bool upperZone);
+    bool isArrowModeActiveVar() const
```
ZeldaZach, line 55: Nit: the `Var` suffix exists only because the member is also named `isArrowModeActive` so the getter can't have the natural name. Suggest renaming the member to `arrowModeActive` (the noun) and letting this getter be `isArrowModeActive()` — that's the idiomatic Qt/C++ shape and reads better at every call site.

PR #6973, opened 2026-06-05
[Game] [Arrows] Use arrowData/registry and generate unique server-side ids

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

Arrow handling was split between client and server with weak identity guarantees (only arrowId) and inconsistent ownership tracking. This made deletion, lookup, and synchronization between players fragile, especially when multiple players could create arrows with overlapping IDs or missing context about who created them.

- Introduces a new ArrowRegistry to centrally track all arrows using a composite key (creatorId, arrowId)
- Adds creatorId and isLocalCreator to ArrowData, making ownership explicit everywhere
- Moves arrow ID generation fully to the server via Server_Game::generateArrowId() (atomic, global uniqueness)
- Removes legacy per-player arrow ID generation on the server
- ArrowItem now owns a shared ArrowData object instead of separate fields, and uses it for rendering, input, and deletion
- All arrow signals/events now carry (creatorId, arrowId) instead of just arrowId
- GameScene and event handlers are updated to use the registry for add/remove/lookup instead of raw maps

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -56,6 +56,7 @@ set(cockatrice_SOURCES
     src/filters/filter_tree_model.cpp
     src/filters/syntax_help.cpp
     src/game/abstract_game.cpp
+    src/game/arrow_registry.cpp
     src/game/board/abstract_card_drag_item.cpp
     src/game/board/abstract_card_item.cpp
     src/game/board/abstract_counter.cpp
--- a/cockatrice/src/game/arrow_registry.cpp
+++ b/cockatrice/src/game/arrow_registry.cpp
@@ -1,0 +1,48 @@
+#include "arrow_registry.h"
+
+#include "board/arrow_item.h"
+
+void ArrowRegistry::insert(QSharedPointer<ArrowData> data, ArrowItem *arrow)
+{
+    const ArrowKey key{data->creatorId, data->id};
+
+    if (auto *existing = take(data->creatorId, data->id)) {
+        existing->delArrow();
+    }
+
+    dataStore.insert(key, data);
+    items.insert(key, arrow);
+    byPlayer[data->creatorId].insert(data->id);
+}
+
+ArrowItem *ArrowRegistry::take(int creatorId, int arrowId)
+{
+    const ArrowKey key{creatorId, arrowId};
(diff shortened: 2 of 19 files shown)


RickyRister:
Viewing this old replay causes crash

[replay_10563858.cor.zip](https://github.com/user-attachments/files/28647611/replay_10563858.cor.zip)

BruebachL:
It worked until I rebased, I'm not sure why it's going haywire now.

BruebachL:
Fixed now.

RickyRister:
Are the server changes backwards compatible with older clients, and are new clients backwards compatible with pre-change servers?

BruebachL:
> Are the server changes backwards compatible with older clients, and are new clients backwards compatible with pre-change servers?

Yeah, works just fine with release and main clients on a main server.

RickyRister:
Arrows on transformed cards in old replays are still slightly broken compared to before, in that they stick around after the point where they should be removed. But at least now the arrow eventually gets removed by something.

https://github.com/user-attachments/assets/1855f410-b392-4024-80ee-d32d2030b69b

https://github.com/user-attachments/assets/8958cc6c-af79-4e3e-a495-d7cf47e40071

But also, cockatrice segfaults if you close the replay while one of those ghost arrows are active

https://github.com/user-attachments/assets/ec7f60c5-7e91-4b21-b3ef-f7be115a936a

BruebachL:
> Arrows on transformed cards in old replays are still slightly broken compared to before, in that they stick around after the point where they should be removed. But at least now the arrow eventually gets removed by something.
> borked.mp4
> good.mp4
> 
> But also, cockatrice segfaults if you close the replay while one of those ghost arrows are active
> crash.mp4

I think that's just an issue with old replays at this point (because it sends colliding ids).
It seems to work correctly for transformed cards in a live game on a new server.

RickyRister:
I can accept that old replays will be broken in edge cases.

However, can we at least make it not crash cockatrice?

BruebachL:
I can't provoke any more crashes or ghost arrows with this state.

RickyRister (approved):
Yeah, seems fixed now. Thanks.

Can you please shorten the PR title before you merge though.

PR #6974, opened 2026-06-06
do not save a const reference to the user data in the info dialog

ebbit1q:
## Related Ticket(s)
- regression introduced in #5546

when resizing the user info dialog of a user without profile picture, the client would crash

- do not save the reference to the user data
- copy the fields needed to generate the pixmap instead, its a measly couple strings
- (notably don't copy the entire info which also includes the image, that'd be a bit wasteful)
- include the userlevel header
- fix the user age format test by moving the days and years between function to utils

### DIFF
--- a/libcockatrice_utility/libcockatrice/utility/days_years_between.h
+++ b/libcockatrice_utility/libcockatrice/utility/days_years_between.h
@@ -1,0 +1,8 @@
+#include <QDateTime>
+
+inline static QPair<int, int> getDaysAndYearsBetween(const QDate &then, const QDate &now)
+{
+    int years = now.addDays(1 - then.dayOfYear()).year() - then.year(); // there is no yearsTo
+    int days = then.addYears(years).daysTo(now);
+    return {days, years};
+}
--- a/cockatrice/src/interface/widgets/server/user/user_info_box.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_info_box.cpp
@@ -85,24 +85,15 @@ void UserInfoBox::retranslateUi()
     avatarButton.setText(tr("Change avatar"));
 }
 
-/**
- * Creates the default profile pic that is used when the user doesn't have a custom pic
- */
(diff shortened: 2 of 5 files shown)


RickyRister (approved):

libcockatrice_utility/libcockatrice/utility/days_years_between.h
```diff
@@ -0,0 +1,8 @@
+#include <QDateTime>
```
ZeldaZach, line 1: This header has no include guard — every other header in `libcockatrice_utility/libcockatrice/utility/` (e.g. `qt_utils.h`, `color.h`, `expression.h`, `user_level.h`) uses `#ifndef … #define … #endif`. Because `getDaysAndYearsBetween` has internal linkage here (`inline static` at namespace scope), any translation unit that ends up including this header twice — directly and transitively, once anything else picks it up — will fail to compile with a redefinition error. Suggest adding a guard, e.g.:

```cpp
#ifndef DAYS_YEARS_BETWEEN_H
#define DAYS_YEARS_BETWEEN_H

#include <QDateTime>

// ...

#endif
```

libcockatrice_utility/libcockatrice/utility/days_years_between.h
```diff
@@ -0,0 +1,8 @@
...
+
+inline static QPair<int, int> getDaysAndYearsBetween(const QDate &then, const QDate &now)
```
ZeldaZach, line 3: Nit: `inline static` had a different meaning when this was a static class member — `static` meant class-static. Now that it's a free function at namespace scope, `static` gives the function internal linkage, so every TU including this header gets its own copy. For a header utility, the canonical idiom is just `inline` (external linkage, deduplicated by the linker). Not material at this size, but the `static` was probably carried over by accident in the move.

ZeldaZach (approved):

BruebachL:
Please hold on this.

PR #6975, opened 2026-06-06
CI: Print colored diff for lint check

tooomm:
Diffs were not able to be colored in the CI logs because the script output was send into a variable.

- Print output directly
  Output of format.sh script will be same than running locally, not the additionally processed version we used before, plus the "passed" or "failed" hints
- Use colored diff option

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -107,7 +107,7 @@ OPTIONS:
 
     --shell
         Use shellcheck to lint shell files.
-		Not available in the default inline mode.
+        Not available in the default inline mode.
 
     -t, --test
         Do not edit files in place. Set exit code to 1 if changes are required.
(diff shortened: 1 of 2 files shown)


format.sh
```diff
@@ -103,11 +103,11 @@ OPTIONS:
...
 
     --print-version
-        Print the version of clang-format being used before continuing.
+        Print the lint tool version being used before continuing.
 
     --shell
-        Use shellcheck to lint shell files. Not available in the default inline
-        mode.
+        Use shellcheck to lint shell files.
+		Not available in the default inline mode.
```
ZeldaZach: This continuation line is indented with two tabs, while every other description line in the `--help` heredoc uses 8 spaces. `cat -A` confirms:

```
        Use shellcheck to lint shell files.$
^I^INot available in the default inline mode.$
```

This will misalign the rendered help output. Suggest using 8 spaces to match the surrounding option descriptions.

ZeldaZach (approved):

PR #6976, opened 2026-06-06
[Servatrice] Notify newly logged-in users of pending server shutdown

### LINKED ISSUE #6562: warn users about server reboot when connecting
**Similar Requests**

when the server plans a reboot all users in the server get a popup message at regular intervals, however when a player joins after one of these is sent they'll not know about it at all until the next interval comes up, instead the server should send them an event immediately on connection to make sure they don't miss it.

**Additional Context**

kongwu666:
## Related Ticket(s)

- Fixes #6562

When a server shutdown is scheduled, already connected users receive the shutdown notification via `shutdownTimeout()`. However, users who log in between two broadcast cycles do not receive the notification until the next scheduled broadcast, which can be up to 60 seconds later.

- Added a `getLoginSessionEvent()` virtual hook to `Server` (similar to `getLoginMessage()`), allowing subclasses to provide session events to send after login
- `Servatrice` overrides this to return an `Event_ServerShutdown` when a shutdown is pending, so newly logged-in users are immediately notified
- `Server` base class returns `nullptr` by default, preserving existing behavior for other subclasses

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -673,8 +673,27 @@ void Servatrice::statusUpdate()
     }
 }
 
+SessionEvent *Servatrice::makeShutdownEvent() const
+{
+    Event_ServerShutdown event;
+    event.set_reason(shutdownReason.toStdString());
+    event.set_minutes(static_cast<google::protobuf::uint32>(shutdownMinutes));
+    return Server_ProtocolHandler::prepareSessionEvent(event);
+}
+
+SessionEvent *Servatrice::getLoginSessionEvent() const
+{
+    // Notify newly logged-in users of a pending server shutdown
+    QMutexLocker locker(&shutdownStateMutex);
+    if (shutdownTimer && shutdownMinutes > 0) {
+        return makeShutdownEvent();
+    }
+    return nullptr;
+}
+
 void Servatrice::scheduleShutdown(const QString &reason, int minutes)
 {
+    shutdownStateMutex.lock();
     shutdownReason = reason;
     shutdownMinutes = minutes;
     nextShutdownMessageMinutes = shutdownMinutes;
@@ -683,6 +702,7 @@ void Servatrice::scheduleShutdown(const QString &reason, int minutes)
         connect(shutdownTimer, SIGNAL(timeout()), this, SLOT(shutdownTimeout()));
         shutdownTimer->start(60000);
     }
+    shutdownStateMutex.unlock();
     shutdownTimeout();
 }
 
@@ -702,6 +722,7 @@ void Servatrice::incRxBytes(quint64 num)
 
 void Servatrice::shutdownTimeout()
 {
+    QMutexLocker locker(&shutdownStateMutex);
     // Show every time counter cut in half & every minute for last 5 minutes
     if (shutdownMinutes <= 5 || shutdownMinutes == nextShutdownMessageMinutes) {
         if (shutdownMinutes == nextShutdownMessageMinutes) {
@@ -710,10 +731,7 @@ void Servatrice::shutdownTimeout()
 
         SessionEvent *se;
         if (shutdownMinutes) {
-            Event_ServerShutdown event;
-            event.set_reason(shutdownReason.toStdString());
-            event.set_minutes(static_cast<google::protobuf::uint32>(shutdownMinutes));
-            se = Server_ProtocolHandler::prepareSessionEvent(event);
+            se = makeShutdownEvent();
         } else {
             Event_ConnectionClosed event;
             event.set_reason(Event_ConnectionClosed::SERVER_SHUTDOWN);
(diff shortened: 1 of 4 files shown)


servatrice/src/servatrice.cpp
```diff
@@ -673,6 +673,18 @@ void Servatrice::statusUpdate()
     }
 }
 
+SessionEvent *Servatrice::getLoginSessionEvent() const
+{
+    // Notify newly logged-in users of a pending server shutdown
+    if (shutdownTimer && shutdownMinutes > 0) {
+        Event_ServerShutdown event;
+        event.set_reason(shutdownReason.toStdString());
+        event.set_minutes(static_cast<google::protobuf::uint32>(shutdownMinutes));
```
ZeldaZach: Thread-safety concern: `cmdLogin()` (and therefore `getLoginSessionEvent()`) runs on the connection-pool thread the socket was moved to in `Servatrice_GameServer::incomingConnection` (`ssi->moveToThread(pool->thread())`), but `shutdownTimer`, `shutdownMinutes`, and `shutdownReason` are written on the main thread by `scheduleShutdown()` and the `shutdownTimer`-driven `shutdownTimeout()`. With no synchronization here, this is a data race — and concurrent assign/read on the implicitly-shared `QString shutdownReason` is undefined behavior, not just a torn-int issue.

The codebase already has the precedent for the same `cmdLogin` codepath: `loginMessage` is guarded by `loginMessageMutex` (see `getLoginMessage()` in `servatrice.h:216`). Suggest adding a similar mutex (e.g. `shutdownStateMutex`) and locking it in `getLoginSessionEvent()`, `scheduleShutdown()`, and `shutdownTimeout()` around accesses to these three members.

servatrice/src/servatrice.cpp
```diff
@@ -673,6 +673,18 @@ void Servatrice::statusUpdate()
...
+        return Server_AbstractUserInterface::prepareSessionEvent(event);
```
ZeldaZach: Nit: every other call site in this file (e.g. lines 561, 716, 720) uses `Server_ProtocolHandler::prepareSessionEvent(...)`. Both resolve to the same static method, but worth keeping the qualifier consistent within the file.

ZeldaZach (changes requested):

servatrice/src/servatrice.cpp
```diff
@@ -673,8 +673,22 @@ void Servatrice::statusUpdate()
     }
 }
 
+SessionEvent *Servatrice::getLoginSessionEvent() const
+{
+    // Notify newly logged-in users of a pending server shutdown
+    QMutexLocker locker(&shutdownStateMutex);
+    if (shutdownTimer && shutdownMinutes > 0) {
+        Event_ServerShutdown event;
```
ZeldaZach: Nit (non-blocking): this `Event_ServerShutdown` construction (reason + minutes) is now duplicated verbatim between `getLoginSessionEvent()` and `shutdownTimeout()` (line ~729). Consider a small private helper, e.g. `SessionEvent *makeShutdownEvent() const` that both call, to keep the two in sync if the payload ever changes.

For what it's worth, the rest of the change looks correct: ownership of the enqueued event is handled by `ResponseContainer`'s destructor (matching the existing `getLoginMessage` path), and the new `shutdownStateMutex` has consistent lock ordering — `shutdownTimeout()` takes it before `clientsLock`, while `getLoginSessionEvent()`/`scheduleShutdown()` never hold `clientsLock`, so there's no ABBA deadlock.

ZeldaZach (approved):
Looks good. Correctly notifies mid-countdown logins of a pending shutdown via the new `getLoginSessionEvent()` hook. Verified: event ownership is handled by `~ResponseContainer` (matches the existing `getLoginMessage` path), and the new `shutdownStateMutex` has consistent lock ordering (`shutdownTimeout()` takes it before `clientsLock`; the login path never holds `clientsLock`), so no deadlock. One optional nit left inline about deduplicating the `Event_ServerShutdown` construction.

tooomm:
@kongwu666 Do you want to look into the last review comment and deduplicate that part?

Looking forward to merging your change!

kongwu666:
@tooomm Done, extracted makeShutdownEvent() to deduplicate the construction.

BruebachL (approved):

tooomm:
Thanks for following up on the request and working on this improvement in the first place. Much appreciated, @kongwu666!

If anything, it helps if you do not squash all your changes into a single commit. In this case it would have highlighted the requested fixes after the review. That makes reviewing a bit easier and gives context.
We squash PRs anyway when merging, so you do not need to worry about polluting the history and having some traces in the PRs is useful.

Review time takes currently a bit as there is a big pile of outstanding PRs and a steady flow of new additions from our team members.
I hope that does not discourage you to submit further improvements. 😄 
If you haven't don't so yet, feel free to connect with other devs on Discord!



👍 for the sanity check here as well, Bruebach.

PR #6978, opened 2026-06-07
Fix #6952: prevent deck loss when saving to a full disk

### LINKED ISSUE #6952: Full disk does not generate a warning; instead replaces deck with empty file
**System Information:**

Client Version: 3.0.0 (2026-05-08)
Build Architecture: 64-bit
Qt Version: 6.11.0
System Locale: en_GB
Install Mode: Portable

<details><summary>Debug Log:</summary>

[2026-05-28 22:16:38.558 I] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as cockatrice format; retrying as plain format [unknown:0]
[2026-05-28 22:16:38.558 W] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as 0 [unknown:0]
[2026-05-28 22:16:40.429 I] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as cockatrice format; retrying as plain format [unknown:0]
[2026-05-28 22:16:40.429 W] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as 0 [unknown:0]
[2026-05-28 22:31:42.031 I] [unknown] - Saved deck to  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" with format 1 - true [unknown:0]
[2026-05-28 22:36:26.253 I] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as cockatrice format; retrying as plain format [unknown:0]
[2026-05-28 22:36:26.253 W] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as 0 [unknown:0]
[2026-05-28 22:36:32.910 I] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as cockatrice format; retrying as plain format [unknown:0]
[2026-05-28 22:36:32.910 W] [unknown] - Failed to load  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" as 0 [unknown:0]
[2026-05-28 22:36:47.168 I] [unknown] - Saved deck to  "G:/cockatrice300/data/decks/edh-b2-gitrog-reanimate.cod" with format 1 - true [unknown:0]
[2026-05-28 22:38:04.412 I] [unknown] - Saved deck to  "G:/cockatrice300/data/decks/edh-b3-gitrog-reanimate.cod" with format 1 - true [unknown:0]
[2026-05-28 22:40:02.434 I] [unknown] - Saved deck to  "D:/Games/edh-b3-gitrog-reanimate.cod" with format 1 - true [unknown:0]

</details>

The drive was completely full. However, attempting to save a deck indicated success. The result was that the saved file was 0 bytes, resulting in a lost decklist.

The reason the drive filled while Cockatrice was running was due to Oracle updating, so it can cause unexpected loss of data.

**Steps to reproduce:**
 - Have a full disk.
 - Open a deck on the full disk
 - Add one card to it
 - Save the deck (ctrl+s)
 - Observe that the deck is now a 0 byte file.

Booyaka101:
## Related Ticket(s)
- Fixes #6952

On a full disk, saving a deck (Ctrl+S) reports success but writes a 0-byte file, silently destroying the decklist.

`DeckLoader::saveToFile()` opens the target with `QFile` in `WriteOnly` mode, which truncates the existing file to 0 bytes the instant it is opened. The serializers (`DeckList::saveToFile_Native()` / `saveToFile_Plain()`) always return `true`, and the result of `flush()` is ignored, so a write that fails part-way — e.g. because the disk is full — leaves an empty file behind yet is still logged as `Saved deck ... - true`.

`updateLastLoadedTimestamp()` uses the same truncate-then-write pattern and can therefore destroy a deck on load.

- `saveToFile()` and `updateLastLoadedTimestamp()` now use `QSaveFile`, which writes to a temporary file and only atomically replaces the target when `commit()` succeeds.
- On any write / flush / rename failure, `commit()` returns `false`, the original deck is left untouched, and the failure is logged via `qCWarning` instead of being reported as success.
- No behaviour change on the success path.

## Screenshots
N/A — backend file-I/O change.

Validation: on a real full disk (a 16 MB VHD filled to 0 bytes free), the current code produces a 0-byte deck file while returning success; with this change the save fails cleanly and the original deck is preserved intact (verified against Qt 6.8.1 QtCore).

### DIFF
--- a/cockatrice/src/interface/deck_loader/deck_loader.cpp
+++ b/cockatrice/src/interface/deck_loader/deck_loader.cpp
@@ -9,6 +9,7 @@
 #include <QFutureWatcher>
 #include <QPrinter>
 #include <QRegularExpression>
+#include <QSaveFile>
 #include <QStringList>
 #include <QTextCursor>
 #include <QTextDocument>
@@ -129,7 +130,10 @@ std::optional<LoadedDeck> DeckLoader::loadFromRemote(const QString &nativeString
 std::optional<LoadedDeck::LoadInfo>
 DeckLoader::saveToFile(const DeckList &deck, const QString &fileName, DeckFileFormat::Format fmt)
 {
-    QFile file(fileName);
+    // Use QSaveFile so that a failed write (e.g. a full disk) leaves the existing deck untouched
+    // instead of truncating it to a 0-byte file. The target is only replaced once every byte has
+    // been flushed successfully in commit().
+    QSaveFile file(fileName);
     if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
         qCWarning(DeckLoaderLog) << "Could not create or open file:" << fileName;
         return std::nullopt;
@@ -145,14 +149,18 @@ DeckLoader::saveToFile(const DeckList &deck, const QString &fileName, DeckFileFo
             break;
     }
 
-    file.flush();
-    file.close();
-
-    qCInfo(DeckLoaderLog) << "Saved deck to " << fileName << "with format" << fmt << "-" << success;
-
     if (!success) {
+        file.cancelWriting();
+        qCWarning(DeckLoaderLog) << "Failed to serialize deck for file:" << fileName;
         return std::nullopt;
     }
+
+    if (!file.commit()) {
+        qCWarning(DeckLoaderLog) << "Failed to save deck to " << fileName << ":" << file.errorString();
+        return std::nullopt;
+    }
+
+    qCInfo(DeckLoaderLog) << "Saved deck to " << fileName << "with format" << fmt;
 
     LoadedDeck::LoadInfo lastLoadInfo = {fileName, fmt};
     return lastLoadInfo;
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/deck_loader/deck_loader.cpp
```diff
@@ -196,38 +204,44 @@ bool DeckLoader::updateLastLoadedTimestamp(LoadedDeck &deck)
...
+    QSaveFile file(fileName);
     if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
         qCWarning(DeckLoaderLog) << "Failed to open file for writing:" << fileName;
         return false;
     }
 
-    bool result = false;
-
     // Perform file modifications
     deck.deckList.setLastLoadedTimestamp(QDateTime::currentDateTime().toString());
```
RickyRister, line 216: If the write fails, then would the decklist object itself to be modified but the change is not synced to the file? 

Granted, I doubt it's a problem, since the decklist object will eventually be discarded, and the next time we load the deck into a new decklist object again, it'll have the old values.

RickyRister (approved):
LGTM. Didn't know about `QSaveFile`.

cockatrice/src/interface/deck_loader/deck_loader.cpp
```diff
@@ -196,38 +204,44 @@ bool DeckLoader::updateLastLoadedTimestamp(LoadedDeck &deck)
...
 
-        if (!file.setFileTime(originalTimestamp, QFileDevice::FileModificationTime)) {
-            qCWarning(DeckLoaderLog) << "Failed to set modification time for file:" << fileName;
-            file.close();
-            return false;
-        }
+    // Re-open the file and restore the original timestamp, so that updating the lastLoadedTimestamp
+    // does not change the file's modification time.
+    QFile timestampFile(fileName);
+    if (!timestampFile.open(QIODevice::ReadWrite)) {
```
ZeldaZach, line 232: Minor nit (also present pre-PR, so feel free to ignore): opening `ReadWrite` here is wider than needed — `setFileTime()` only needs the file to be open, and `WriteOnly` would suffice.

ZeldaZach (commented):
Nice fix — `QSaveFile` is exactly the right tool here, and the `cancelWriting()` placement / `commit()` error handling all look correct.

One observation that's out of scope for the diff but worth raising: `DeckLoader::convertToCockatriceFormat()` (around `deck_loader.cpp:466`) still uses the same `QFile(newFileName)` + `WriteOnly` + `saveToFile_Native` pattern, and then unconditionally `QFile::remove()`s the original `.txt` once `result` is true. Since `saveToFile_Native()` always returns `true`, a full disk during conversion would write a 0-byte `.cod`, then delete the original `.txt` — the exact data-loss class this PR is fixing. Worth either folding into this PR or filing as a quick follow-up.

ZeldaZach (approved):

cockatrice/src/interface/deck_loader/deck_loader.cpp
Booyaka101: Correct, and harmless — the in-memory `lastLoadedTimestamp` is cosmetic and the deck is re-read from disk on next load, so a failed write just leaves a slightly newer in-memory value with no on-disk effect.

cockatrice/src/interface/deck_loader/deck_loader.cpp
Booyaka101: Keeping `ReadWrite` here — `QFile` `WriteOnly` truncates on open, so it would zero out the deck that was just committed. Verified against Qt 6.8: an 11-byte file opened `WriteOnly` with no subsequent write ends up 0 bytes, while `ReadWrite` preserves it. `setFileTime()` only needs the handle open, without truncation.

Booyaka101:
Folded in the `convertToCockatriceFormat()` fix (pushed as a second commit). It had the same data-loss class: `QFile` `WriteOnly` truncated the `.cod`, `saveToFile_Native()` always returns `true`, and the original was then removed unconditionally — so a full disk during conversion left a 0-byte `.cod` and deleted the source deck.

Now it uses `QSaveFile` + `commit()`, removes the original only after a successful commit, and the format check moved ahead of the file open so an already-`.cod`/unsupported deck never opens (and truncates) the target. Validated with the same QtCore harness as the rest of the PR; clang-format clean.

PR #6979, opened 2026-06-07
[ReplayManager] Refactor to send replayed events through signal

RickyRister:

Right now the `ReplayManager` directly calls the `GameEventHander` to process the events.

It would be nicer to use signals for this.

- Make `ReplayManager` emit `eventReplayed` signal
- connect that signal to the `GameEventHandler` inside of `TabGame`

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_game.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_game.cpp
@@ -1169,6 +1169,11 @@ void TabGame::createReplayDock(GameReplay *replay)
                             QDockWidget::DockWidgetMovable);
     replayDock->setWidget(replayManager);
     replayDock->setFloating(false);
+
+    connect(replayManager, &ReplayManager::eventReplayed, game->getGameEventHandler(),
+            [this](auto event, auto options) {
+                game->getGameEventHandler()->processGameEventContainer(event, nullptr, options);
+            });
 }
 
 void TabGame::createDeckViewContainerWidget(bool bReplay)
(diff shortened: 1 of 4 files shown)


cockatrice/src/interface/widgets/tabs/tab_game.h
```diff
@@ -180,6 +180,7 @@ protected slots:
     void connectToGameEventHandler();
     void connectMessageLogToGameEventHandler();
     void connectPlayerListToGameEventHandler();
+
```
ZeldaZach: This blank line looks unrelated to the refactor — could it be dropped?

cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -1169,6 +1169,11 @@ void TabGame::createReplayDock(GameReplay *replay)
                             QDockWidget::DockWidgetMovable);
     replayDock->setWidget(replayManager);
     replayDock->setFloating(false);
+
+    connect(replayManager, &ReplayManager::eventReplayed, game->getGameEventHandler(),
+            [this](auto event, auto options) {
+                game->getGameEventHandler()->processGameEventContainer(event, nullptr, options);
```
ZeldaZach, line 1175: `auto event` will be deduced as `GameEventContainer` by value, so each replayed event copies the protobuf container. The signal already passes it as `const GameEventContainer &` — using `const auto &event` (or explicit `const GameEventContainer &event`) here avoids the copy. `options` is fine by value since `EventProcessingOptions` is a `Q_DECLARE_FLAGS` type.

ZeldaZach (approved):

PR #6981, opened 2026-06-08
[Room][UserList] Introduce style delegate

BruebachL:
The current default style for the treeview that we have for users is a bit... lacking.
image
It's compact but there really isn't much benefit to seeing more users at a time past a certain point. The argument for better visibility would be better served by a search bar (separate PR forthcoming).

Crucially, the current room tab only shows all online users.
Users have to navigate to their account to see their friends list and users they've ignored.
The friends list not being visible is the most egregious in this scenario. Friends are people you expect to interact with more often than random strangers. It makes sense to have ready access to them, to see their online status, message them or invite them to games (separate PR forthcoming).

The current row style does not display user avatars because it is too small. This is a major step down in user customization and makes it not immediately obvious that there even is an avatar functionality. We also have to rely on pawns to signal user levels, which are also not immediately obvious.

A small row size and a hidden right click menu that is not immediately obvious also don't do the UI any favors.
It would be better if the right click menu was instead replaced by a bigger blown up version of the user card, showing the actions one can perform (message, ignore, view games, etc) on hover, saving a click (separate PR forthcoming)

Overall, I think we should 'jazz' it up a little. An option to enable old behavior exists.

- Display 'Friends' and 'Ignored' along with 'Online' in tab_room (Let's maybe table "which tab opens" as a setting for a future PR, if some people prefer the online list)
- Introduce a style delegate for UserListWidget
- Introduce a dialog in tab_account to adjust card art parameters (name, margin, offset, zoom) for the background of the style delegate
- Introduce a blacklist/whitelist and allow moderators to edit rules
- Introduce a setting to go back to old styling

## Screenshots
No art set:
image
image

Setting card art:
image
image
image
image
image
image

Blacklist/Whitelist:
image
image
image

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_card_art_provider.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_card_art_provider.cpp
@@ -79,65 +79,68 @@ void UserCardArtProvider::processQueue()
         return;
     }
 
-    if (queue.isEmpty()) {
-        return;
-    }
+    while (!queue.isEmpty()) {
+        const QString key = queue.dequeue();
 
-    const QString key = queue.dequeue();
-
-    const QStringList parts = key.split(u'|');
-    if (parts.size() != 2) {
-        pending.remove(key);
-        processQueue();
-        return;
-    }
-
-    const QString userName = parts.at(0);
-    const QString cardName = parts.at(1);
-
-    ExactCard card = CardDatabaseManager::query()->getCard({cardName});
-
-    if (!card) {
-        pending.remove(key);
-        processQueue();
-        return;
-    }
-
-    QPixmap fullRes;
-    CardPictureLoader::getPixmap(fullRes, card, QSize(745, 1040));
-
-    // If already available, store immediately
-    if (!fullRes.isNull()) {
-        insertIntoCache(key, cropCardArt(fullRes));
-        pending.remove(key);
-
-        emit cardArtUpdated(userName);
-        processQueue();
-        return;
-    }
-
-    // Otherwise wait for async load
-    QPointer<UserCardArtProvider> self(this);
-
-    connect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, this, [self, key, userName, card]() mutable {
-        if (!self) {
-            return;
+        const QStringList parts = key.split(u'|');
+        if (parts.size() != 2) {
+            pending.remove(key);
+            continue;
         }
 
-        disconnect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, self, nullptr);
+        const QString userName = parts.at(0);
+        const QString cardName = parts.at(1);
+
+        ExactCard card = CardDatabaseManager::query()->getCard({cardName});
+
+        if (!card) {
+            pending.remove(key);
+            continue;
+        }
 
         QPixmap fullRes;
         CardPictureLoader::getPixmap(fullRes, card, QSize(745, 1040));
 
+        // Synchronous hit (already loaded/on disk)
         if (!fullRes.isNull()) {
-            self->insertIntoCache(key, self->cropCardArt(fullRes));
-        } else {
-            self->insertIntoCache(key, QPixmap());
+            insertIntoCache(key, cropCardArt(fullRes));
+            pending.remove(key);
+
+            emit cardArtUpdated(userName);
+            continue;
         }
 
-        self->pending.remove(key);
+        // Async load required.
+        QPointer<UserCardArtProvider> self(this);
 
-        emit self->cardArtUpdated(userName);
-        self->processQueue();
-    });
+        auto conn = std::make_shared<QMetaObject::Connection>();
+
+        *conn = connect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, this,
+                        [self, key, userName, card, conn]() mutable {
+                            if (!self) {
+                                return;
+                            }
+
+                            QObject::disconnect(*conn);
+
+                            QPixmap fullRes;
+                            CardPictureLoader::getPixmap(fullRes, card, QSize(745, 1040));
+
+                            if (!fullRes.isNull()) {
+                                self->insertIntoCache(key, self->cropCardArt(fullRes));
+                            } else {
+                                self->insertIntoCache(key, QPixmap());
+                            }
+
+                            self->pending.remove(key);
+
+                            emit self->cardArtUpdated(userName);
+
+                            // Resume processing remaining queued items.
+                            self->processQueue();
+                        });
+
+        // Stop here. We'll continue when the async load finishes.
+        return;
+    }
 }
--- a/cockatrice/src/interface/widgets/server/user/user_card_art_provider.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_card_art_provider.cpp
@@ -1,0 +1,143 @@
+#include "user_card_art_provider.h"
+
+#include "../../../card_picture_loader/card_picture_loader.h"
+
+#include <QPointer>
+#include <libcockatrice/card/database/card_database_manager.h>
+
+static QString makeKey(const QString &user, const QString &card)
+{
+    return user + u'|' + card;
+}
+
+UserCardArtProvider::UserCardArtProvider(QObject *parent) : QObject(parent)
+{
+    dbReady = (CardDatabaseManager::getInstance()->getLoadStatus() == LoadStatus::Ok);
+
+    if (!dbReady) {
+        connect(CardDatabaseManager::getInstance(), &CardDatabase::cardDatabaseLoadingFinished, this,
+                &UserCardArtProvider::onDatabaseReady);
+    }
+}
+
+void UserCardArtProvider::onDatabaseReady()
+{
+    dbReady = true;
+    processQueue();
+}
+
+const QMap<QString, QPixmap> &UserCardArtProvider::cache() const
+{
+    return cardArtCache;
+}
+
+void UserCardArtProvider::requestCardArt(const QString &userName, const QString &cardName)
+{
+    if (cardName.isEmpty()) {
+        return;
+    }
+
+    const QString key = makeKey(userName, cardName);
+
+    if (cardArtCache.contains(key) || pending.contains(key)) {
+        return;
+    }
+
+    pending.insert(key);
+    queue.enqueue(key);
+
+    processQueue();
+}
+
+QPixmap UserCardArtProvider::cropCardArt(const QPixmap &fullRes)
+{
+    const QSize sz = fullRes.size();
+    const int marginX = sz.width() * 0.07;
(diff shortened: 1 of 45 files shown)


servatrice/src/serversocketinterface.cpp
```diff
@@ -1565,6 +1577,146 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountImage(const Comm
     return Response::RespOk;
 }
 
+bool AbstractServerSocketInterface::isCardNameAllowed(const QString &cardName)
+{
+    QSqlQuery *q =
+        sqlInterface->prepareQuery("SELECT mode FROM cockatrice_card_art_name_rules WHERE card_name = :name");
```
ZeldaZach: Heads-up: this and the three other new card-art-rule queries (lines 1665, 1687, 1701) hardcode the `cockatrice_` table prefix, but every other runtime query in this file uses the `{prefix}_` placeholder (e.g. `{prefix}_users`) that `servatrice_database_interface.cpp:135` substitutes with `server->getDbPrefix()`. Operators who set a non-default `db/prefix` in `servatrice.ini` will hit query failures here (and surface as `RespInternalError`s) while the rest of the server works fine — fun debugging.

Fix is mechanical: `{prefix}_card_art_name_rules` in all four places.

Note that the migration SQL keeps the `cockatrice_` hardcoding, which matches every other migration in the repo and is fine — the convention is migrations are hand-edited per deployment, runtime queries are parameterized.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1565,6 +1577,146 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountImage(const Comm
...
+}
+
+Response::ResponseCode AbstractServerSocketInterface::cmdSetCardArtParams(const Command_SetCardArtParams &cmd,
+                                                                          ResponseContainer & /* rc */)
+{
+    if (authState != PasswordRight) {
+        return Response::RespFunctionNotAllowed;
+    }
+
+    const QString cardName = QString::fromStdString(cmd.card_name());
```
ZeldaZach, line 1605: No length validation on `cardName` before it goes into the DB. The `card_name` column is `varchar(255)` and `cmdAccountPassword` below (line 1727) uses `MAX_NAME_LENGTH` as the reference cap on similar untrusted strings. Without a length check here (and in `cmdAddCardArtRule` / `cmdRemoveCardArtRule`) a long name silently truncates or errors depending on MySQL's `sql_mode`.

Suggest adding `if (cardName.length() > MAX_NAME_LENGTH) return Response::RespInvalidData;` near the top of each of the three card-art commands.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1565,6 +1577,146 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountImage(const Comm
...
+    const QString mode = QString::fromStdString(cmd.mode());
+
+    QSqlQuery *q = sqlInterface->prepareQuery("INSERT INTO cockatrice_card_art_name_rules "
+                                              "(card_name, mode, reason, created_by) "
+                                              "VALUES (:name, :mode, :reason, :uid) "
+                                              "ON DUPLICATE KEY UPDATE mode=:mode2, reason=:reason2");
+
+    q->bindValue(":name", cardName);
+    q->bindValue(":mode", mode);
+    q->bindValue(":mode2", mode);
```
ZeldaZach, line 1683: `mode` is bound directly into the query, but the column is declared `ENUM('ALLOW','DENY')`. A client sending `mode = ""`, `mode = "allow"` (wrong case), or anything else will produce a DB error that surfaces as `RespInternalError` with no useful response code. The same applies to `reason` length.

Suggest validating upfront:
```cpp
if (mode != "ALLOW" && mode != "DENY") {
    return Response::RespInvalidData;
}
if (cardName.isEmpty() || cardName.length() > MAX_NAME_LENGTH) {
    return Response::RespInvalidData;
}
```
This also gives moderators a sensible error message in the UI instead of a generic internal error.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1565,6 +1577,146 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountImage(const Comm
...
+bool AbstractServerSocketInterface::isCardNameAllowed(const QString &cardName)
+{
+    QSqlQuery *q =
+        sqlInterface->prepareQuery("SELECT mode FROM cockatrice_card_art_name_rules WHERE card_name = :name");
+
+    q->bindValue(":name", cardName);
+
+    if (!sqlInterface->execSqlQuery(q)) {
+        return true; // fail-open to avoid breaking server
+    }
```
ZeldaZach, line 1589: Fail-open here is a reasonable choice (don't break the server if the rule table is broken), but the silent return means a DB connectivity blip will bypass the deny-list with no observable signal. Worth logging when this path is hit — operators need to know:
```cpp
if (!sqlInterface->execSqlQuery(q)) {
    qCWarning(loggerCategory) << "Card art rule lookup failed; failing open for" << cardName;
    return true;
}
```

servatrice/migrations/servatrice_0034_to_0035.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE `cockatrice_users` ADD COLUMN `card_art_params` TEXT DEFAULT NULL;
```
ZeldaZach: Heads-up for production deployments: `ALTER TABLE cockatrice_users ADD COLUMN` without an explicit `ALGORITHM=` hint may take a metadata lock for the duration of a full table rebuild, depending on MySQL/MariaDB version. On the public server the users table is large enough that this isn't free.

MySQL 8.0.12+ and recent MariaDB support `ALGORITHM=INSTANT` for column adds with a default — worth setting explicitly so operators don't get bitten:
```sql
ALTER TABLE `cockatrice_users` ADD COLUMN `card_art_params` TEXT DEFAULT NULL, ALGORITHM=INSTANT;
```
If you want to support older MySQL too, fall back to `ALGORITHM=INPLACE, LOCK=NONE`. At minimum a `-- run during low-traffic window` comment in the migration would help operators plan.

cockatrice/src/interface/widgets/server/user/user_list_painter.cpp
```diff
@@ -0,0 +1,342 @@
...
+    const int marginR = qRound(totalW * params.marginPctR);
+    const int drawW = totalW - marginL - marginR;
+
+    const double basescale = qMax(double(drawW) / art.width(), double(cardH) / art.height());
+    const double scale = basescale * params.zoom;
+
+    const int scaledW = qRound(art.width() * scale);
+    const int scaledH = qRound(art.height() * scale);
+
+    const QPixmap scaled = art.scaled(scaledW, scaledH, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
```
ZeldaZach, line 123: This whole block (the `scaled()` on line 123, then `copy() + toImage() + convertToFormat()` on 132, plus the alpha-mask `QPainter` pass on 134-145) runs on the paint path — `paint()` is called per cell on every repaint (hover, selection, scroll). For a room with many visible users, every redraw spends real time here re-doing work whose inputs haven't changed.

For visible-only rendering (~20-30 cells at a time) it's tolerable but wasteful. Worth caching the composited image keyed on `(key, rectHeight)` (row height in a `QTreeView` is usually stable) — the cache would live alongside `cardArtCache` in `UserCardArtProvider`. The gradient mask in particular only depends on the output width, so it never needs to be recomputed per repaint.

cockatrice/src/interface/widgets/server/user/user_card_art_provider.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    // Otherwise wait for async load
+    QPointer<UserCardArtProvider> self(this);
+
+    connect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, this, [self, key, userName, card]() mutable {
+        if (!self) {
+            return;
+        }
+
+        disconnect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, self, nullptr);
```
ZeldaZach: `disconnect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, self, nullptr)` disconnects **every** connection from this `CardInfo` to `self`, not just this lambda. If two users both reference the same card name and both end up here (the cache key is `userName|cardName`, so the dedupe at line 42 doesn't catch them), the first lambda to fire will tear down the second user's connection too — the second user's request stays in `pending` forever, so subsequent retries for that key hit `pending.contains(key)` at line 42 and silently no-op.

Fix is to scope the disconnect to this exact connection using a `QMetaObject::Connection` handle:
```cpp
auto conn = std::make_shared<QMetaObject::Connection>();
*conn = connect(card.getCardPtr().data(), &CardInfo::pixmapUpdated, this,
               [self, key, userName, card, conn]() mutable {
    if (!self) return;
    QObject::disconnect(*conn);
    // ...rest of the handler
});
```

cockatrice/src/interface/widgets/server/user/user_card_art_provider.cpp
```diff
@@ -0,0 +1,143 @@
...
+    QPixmap fullRes;
+    CardPictureLoader::getPixmap(fullRes, card, QSize(745, 1040));
+
+    // If already available, store immediately
+    if (!fullRes.isNull()) {
+        insertIntoCache(key, cropCardArt(fullRes));
+        pending.remove(key);
+
+        emit cardArtUpdated(userName);
+        processQueue();
```
ZeldaZach: Minor follow-on: `processQueue()` tail-recurses synchronously at lines 91, 102, and 115. In the common "reconnect with a populated cache" path every entry hits the line 115 path (cache miss in `pending`, but `getPixmap` returns immediately because the file is on disk), so the depth equals the queue size. A few hundred entries is plenty to cause a stack-warning in debug builds.

Convert to a `while` loop and only fall through to the async branch (the `connect()` block) when an actual async load is needed.

ZeldaZach (commented):
Big PR, broadly really nice UX improvement. Sane settings cache plumbing, sensible per-user JSON storage for the params, and the new Friends/Ignored tabs are a clear win.

Leaving inline comments grouped roughly: a handful of server-side correctness/robustness issues (most importantly: the new card-art-rule queries hardcode the `cockatrice_` table prefix, breaking deployments that configured a different one), one painter perf concern, and one signal/slot bug in the card art provider that I think will bite real users.

cockatrice/src/interface/widgets/server/user/user_list_painter.cpp
BruebachL: Table this for another pass? I'm pretty sure I'm gonna touch this painter again before this feature is fully complete.

ZeldaZach (approved):

PR #6983, opened 2026-06-09
[Game] Add Command Zone

DawnFire42:
## The long awaited Command Zone!
It took some time to extract and review the refactors needed to get there. This is the last piece of the puzzle in the Command Zone saga. Thanks for reviewing my work since the beginning of the year, especially RickyRister, whom I probably annoyed with my commits for a time. I hope this one goes smoothly and people can play with it soon!

I did address a few of the concerns that were pointed out on my first version by simplifying the feature. It now is only a single zone and we can add a second tax in it directly (with the proper menu options appearing when in use). 
I considered renaming the tax counters as 1st tax and 2nd tax, maybe to do it more generally then mtg specific, but I wasn't sure so please let me know what you think for this.


RickyRister:
This is NOT going to play nice with the upcoming game logic/ui split refactor.
Pinging @BruebachL about what order we want to do this in.

BruebachL:
I've rebased this and flagged some open to-do's.

DawnFire42:
I addressed all the todo's for logic/Graphic separation and rebased. This actually made it easier to find a bunch of small issues after, so I've done a thorough code cleanup and I think the PR is in a much better shape. 
I'm done with that code cleanup commit, let me know if there's anything else I should change and I'll do it in separate commits.

libcockatrice_network/libcockatrice/network/server/remote/game/server_player.cpp
```diff
@@ -531,6 +566,39 @@ Server_Player::cmdDelCounter(const Command_DelCounter &cmd, ResponseContainer &
...
+        return Response::RespGameNotStarted;
+    }
+    if (conceded) {
+        return Response::RespContextError;
+    }
+
+    const int counterId = cmd.counter_id();
+    if (isCommandZoneCounterBlocked(counterId)) {
+        return Response::RespContextError;
+    }
```
ZeldaZach, line 624: `isCommandZoneCounterBlocked` only screens out IDs 8/9 *when the command zone is disabled* — it doesn't restrict which counters `cmdSetCounterActive` can target. As written, a client can send `cmd_set_counter_active` with `counter_id = 0` and hide their own life counter (or any of the mana/storm counters). Probably not the intended scope of the new command.

Suggest restricting this command to the tax-counter IDs explicitly. Something like:

```cpp
if (counterId != CounterIds::CommanderTax && counterId != CounterIds::PartnerTax) {
}
if (!game->getEnableCommandZone()) {
}
```

If you do want to allow toggling any user-created counter long-term, that's a reasonable feature — but I'd land it in a follow-up rather than as a side effect of this PR, since it has different UI/UX implications.

libcockatrice_network/libcockatrice/network/server/remote/game/server_player.cpp
```diff
@@ -101,6 +104,17 @@ void Server_Player::setupZones()
...
+    if (game->getEnableCommandZone()) {
+        addZone(new Server_CardZone(this, ZoneNames::COMMAND, false, ServerInfo_Zone::PublicZone));
+        addCounter(new Server_Counter(CounterIds::CommanderTax, CounterNames::CommanderTax, makeColor(128, 128, 128),
+                                      20, 0, 0, MAX_COUNTER_VALUE));
+        auto *partnerTax = new Server_Counter(CounterIds::PartnerTax, CounterNames::PartnerTax,
+                                              makeColor(128, 128, 128), 20, 0, 0, MAX_COUNTER_VALUE);
+        (void)partnerTax->setActive(false);
+        addCounter(partnerTax);
+    }
+
```
ZeldaZach, line 119: Related to the `counter_ids.h` comment: this block only creates the reserved counters at IDs 8/9 when `enableCommandZone` is true. In a **non-Commander game**, `newCounterId()` (line 77, just above setupZones) returns `highest_existing + 1 = 8` — so the first user-created counter in a non-Commander game gets ID 8, colliding with the reserved range. A malicious client could then use `cmd_set_counter_active` (see the comment below) to toggle that user counter via the reserved ID, confusing the client UI.

Easiest fix: make `newCounterId()` always skip the reserved range, e.g.:

```cpp
int Server_Player::newCounterId() const {
    int id = 0;
    for (auto *c : counters) id = std::max(id, c->getId());
    return std::max(id + 1, CounterIds::PartnerTax + 1);
}
```

Or add a `CounterIds::FirstUserId = 10` constant and use that. Either way, combined with the allowlist suggested for `cmdSetCounterActive`, the reserved range becomes genuinely reserved regardless of game mode.

libcockatrice_utility/libcockatrice/utility/counter_ids.h
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Reserved counter IDs for system counters:
+ *   IDs 0-7:  Standard player counters (life, mana colors, storm)
+ *   IDs 8-9:  Commander tax counters
+ *   IDs 10+:  Available for user-created counters
+ *
+ * The server's newCounterId() starts from the highest existing ID + 1,
+ * so these reserved IDs won't conflict as long as they're created first
+ * during setupZones(). See server_player.cpp::setupZones() for the
+ * authoritative list of reserved IDs.
```
ZeldaZach: Heads-up: this docstring says the reserved IDs "won't conflict as long as they're created first during setupZones()", but in non-Commander games they aren't created at all — so the reservation doesn't actually hold and `newCounterId()` can return 8. See the comment on `server_player.cpp` for the suggested fix; if you take it, this wording could change to something like "The server's `newCounterId()` always skips this reserved range, regardless of game mode."

cockatrice/src/game_graphics/player/menu/command_zone_menu.cpp
```diff
@@ -0,0 +1,185 @@
...
+        aIncreasePartnerTax->setText(tr("Increase &Partner Tax (+1)"));
+    }
+    if (aDecreasePartnerTax) {
+        aDecreasePartnerTax->setText(tr("Decrease P&artner Tax (-1)"));
+    }
+    if (aTogglePartnerTaxCounter) {
+        aTogglePartnerTaxCounter->setText(tr("&Add Partner Tax"));
+    }
+    if (aToggleMinimized) {
+        aToggleMinimized->setText(tr("&Minimize"));
```
ZeldaZach, line 82: The "Minimize" label is set once in `retranslateUi()` and never updated, so once the user minimizes the zone the menu still shows "Minimize" (and clicking it restores). The other tax-counter labels do flip correctly in `updateTaxCounterActionStates()`.

Suggest adding a similar update for this action — e.g. at the bottom of `updateTaxCounterActionStates()`:

```cpp
if (aToggleMinimized) {
    CommandZone *zone = player->getCommandZoneGraphicsItem();
    aToggleMinimized->setText(zone && zone->isMinimized() ? tr("&Restore") : tr("&Minimize"));
}
```

cockatrice/src/game_graphics/player/menu/command_zone_menu.cpp
```diff
@@ -0,0 +1,185 @@
...
+        addSeparator();
+
+        aToggleCommanderTaxCounter = new QAction(this);
+        connect(aToggleCommanderTaxCounter, &QAction::triggered, this,
+                [playerActions]() { playerActions->actToggleTaxCounter(CounterIds::CommanderTax); });
+        addAction(aToggleCommanderTaxCounter);
+
+        aTogglePartnerTaxCounter = new QAction(this);
+        connect(aTogglePartnerTaxCounter, &QAction::triggered, this,
+                [playerActions]() { playerActions->actToggleTaxCounter(CounterIds::PartnerTax); });
```
ZeldaZach: Small consistency nit — the `aViewZone` lambda just above (line 23) resolves `player->getLogic()` at call time and null-checks it, which is the safer pattern this codebase uses. The six lambdas here instead capture a raw `PlayerActions *` (line 34) that was resolved during construction.

In practice the connections are auto-disconnected when this menu is destroyed (since the receiver is `this`), so it's unlikely to crash — but if `PlayerLogic` ever gets torn down before the menu, the captured `playerActions` becomes dangling. Cheap to make consistent:

```cpp
connect(aIncreaseCommanderTax, &QAction::triggered, this, [this]() {
    if (auto *logic = player->getLogic()) {
        logic->getPlayerActions()->actModifyTaxCounter(CounterIds::CommanderTax, 1);
    }
});
```

Not a blocker, just matches the pattern already used above.

libcockatrice_protocol/libcockatrice/protocol/pb/event_set_counter_active.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto2";
+import "game_event.proto";
+
+message Event_SetCounterActive {
+    extend GameEvent {
+        optional Event_SetCounterActive ext = 2023;
+    }
+    optional sint32 counter_id = 1;
+    optional bool active = 2;
```
ZeldaZach: Minor consistency: the matching `Command_SetCounterActive.active` is declared as `optional bool active = 2 [default = true];` — adding the same `[default = true]` here keeps the two protos symmetric. In practice the field is always set by the server when emitting the event, but the default matters if a future code path constructs an event without explicitly setting `active`.

tests/server_counter_test.cpp
```diff
@@ -79,6 +80,37 @@ TEST(ServerCounter, MixedExtremesDoNotClamp)
...
+
+TEST(ServerCounter, CustomBoundsForCommanderTax)
+{
+    Server_Counter taxCounter(1, "tax", color(), 20, 0, 0, MAX_COUNTER_VALUE);
+    EXPECT_TRUE(taxCounter.setCount(1000));
+    EXPECT_EQ(taxCounter.getCount(), MAX_COUNTER_VALUE);
+    EXPECT_TRUE(taxCounter.setCount(-5));
+    EXPECT_EQ(taxCounter.getCount(), 0);
+}
+
```
ZeldaZach, line 113: Nice to see the new bounds-clamping tests. One gap worth filling: there's no coverage for the new `setActive`/`isActive` behavior. A quick test would catch regressions in the change-detection bool:

```cpp
TEST(ServerCounter, SetActiveReportsChangeOnce)
{
    Server_Counter c(1, "test", color(), 10, 0);
    EXPECT_TRUE(c.isActive());
    EXPECT_TRUE(c.setActive(false));
    EXPECT_FALSE(c.isActive());
    EXPECT_FALSE(c.setActive(false));  // no change → returns false
    EXPECT_TRUE(c.setActive(true));
}
```

Not blocking — just easy to add while you're in here.

ZeldaZach (commented):
Welcome, and thanks for sticking with this through all the refactors — getting a 70-file Command Zone PR to this point is a real accomplishment. The split between graphics and logic landed nicely, the reserved-counter-ID approach is a clean way to avoid coupling, the theme support is comprehensive, and the new test cases in `server_counter_test.cpp` are good additions. The PR is in pretty solid shape overall.

I've left some inline comments where I think changes are worth making before merge. The two I'd flag as most important are both server-side: `cmdSetCounterActive` letting players toggle any counter (including life), and the reserved tax-counter IDs not actually being reserved in non-Commander games. The rest are smaller cleanups — none should be hard. Happy to clarify anything that's unclear.

DawnFire42:
Thanks again for the reviews! I made dedicated commits for everything that was flagged, then through the rebases I kept finding things to fix or improve. But we're back in a review-able state! Let me know if there's anything else I can improve

DawnFire42:
I added keyboard shortcuts registration without default key set. I'm usually not the one who uses those shortcuts so I wouldn't know what to use here, but at least they can be set for anyone who likes to use them.

cockatrice/src/game_graphics/zones/command_zone.cpp
```diff
@@ -0,0 +1,187 @@
...
+    update();
+}
+
+void CommandZone::registerTaxCounter(AbstractCounter *counter)
+{
+    if (!counter || taxCounters.contains(counter)) {
+        return;
+    }
+    taxCounters.append(counter);
+    connect(counter, &QObject::destroyed, this, [this, counter]() { taxCounters.removeOne(counter); });
```
ZeldaZach: This `destroyed` handler can fire during the `CommandZone`'s own teardown and touch an already-freed member.

`taxCounters` is the last-declared member of `CommandZone`, so it's destructed in the member-cleanup phase, before any base-class destructor runs. The tax counters are graphics children (created as `new CommanderTaxCounter(state, player, commandZoneGraphicsItem)` with no QObject parent), so they get deleted inside `CommandZone`'s `~QGraphicsItem`, which runs *before* `~QObject` — `QGraphicsObject` derives `public QObject, public QGraphicsItem`, so the graphics base destructs first. When each child emits `destroyed()` there, this connection is still live (the receiver `this` hasn't been disconnected yet), and the lambda runs `taxCounters.removeOne(counter)` on the destructed QList. That's a use-after-free on every command-zone teardown that still has tax counters, i.e. closing any Commander game. Whether it crashes is allocator-dependent.

The usual reason this pattern is safe for QObject-parented children doesn't hold here: `~QObject` disconnects before it deletes QObject children, but these children are graphics-parented and deleted earlier, in `~QGraphicsItem`.

Simplest fix is to sever the connections in a destructor, which runs before the members tear down:

```cpp
// command_zone.h
~CommandZone() override;

// command_zone.cpp
CommandZone::~CommandZone()
{
    for (AbstractCounter *ctr : taxCounters) {
        disconnect(ctr, &QObject::destroyed, this, nullptr);
    }
}
```

cockatrice/src/game_graphics/player/menu/move_menu.cpp
```diff
@@ -65,6 +75,7 @@ void MoveMenu::setShortcutsActive()
     aMoveToHand->setShortcuts(shortcuts.getShortcut("Player/aMoveToHand"));
     aMoveToGraveyard->setShortcuts(shortcuts.getShortcut("Player/aMoveToGraveyard"));
     aMoveToExile->setShortcuts(shortcuts.getShortcut("Player/aMoveToExile"));
+    aMoveToCommandZone->setShortcuts(shortcuts.getShortcut("Player/aMoveToCommandZone"));
```
ZeldaZach, line 75: `Player/aMoveToCommandZone` is read here but never registered in `shortcuts_settings.h`, unlike the five command-zone shortcuts added there (`aViewCommandZone`, `aAddCommanderTax`, and friends). `getShortcut()` returns an empty key so nothing breaks, but this action can't be bound and won't appear in the shortcuts editor. Adding an entry alongside the other `Move to` keys fixes it:

```cpp
{"Player/aMoveToCommandZone",
 ShortcutKey(QT_TRANSLATE_NOOP("shortcutsTab", "Command Zone"), parseSequenceString(""), ShortcutGroup::Player_Cards)},
```

libcockatrice_network/libcockatrice/network/server/remote/game/server_player.cpp
```diff
@@ -463,8 +511,15 @@ Server_Player::cmdCreateCounter(const Command_CreateCounter &cmd, ResponseContai
         return Response::RespContextError;
     }
 
-    auto *c = new Server_Counter(newCounterId(), nameFromStdString(cmd.counter_name()), cmd.counter_color(),
-                                 cmd.radius(), cmd.value());
+    const QString counterName = nameFromStdString(cmd.counter_name());
+    // Reserved system counter names (commander/partner tax) are how clients identify
+    // server-managed tax counters for rendering and logging; a client must not be able
+    // to spoof one via a user-created counter.
+    if (CounterNames::isTaxCounter(counterName)) {
```
ZeldaZach: This reserved-name guard is the one part of the counter-command hardening with no test behind it. The del / set-active / modify authorization each got pulled into a static `evaluate*` helper with per-branch tests, but this name check stayed inline in `cmdCreateCounter`, so nothing exercises it — inverting or removing it would keep every test green.

Worth extracting the same way (e.g. `static Response::ResponseCode evaluateCreateCounter(const QString &counterName)`) and adding cases: a reserved name (`CounterNames::CommanderTax` / `PartnerTax`) returns `RespFunctionNotAllowed`, an ordinary name returns `RespOk`. A direct `CounterNames::isTaxCounter(...)` unit test is a lighter alternative, but going through the command path is more meaningful.

tests/command_zone_tests/new_counter_id_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+TEST(NewCounterId, ReturnsFirstUserIdWhenNoCounters)
+{
+    PlayerFixture f;
+    EXPECT_EQ(f.player.newCounterId(), CounterIds::FirstUserId);
+}
+
+TEST(NewCounterId, SkipsReservedRangeWhenOnlyReservedCountersExist)
+{
+    PlayerFixture f;
+    f.player.addCounter(new Server_Counter(3, "g", color(), 20, 0));
```
ZeldaZach, line 47: Small coverage note: this pins the `FirstUserId` floor using reserved ids 3 and 5, but not the tax-counter ids the floor actually exists to protect (`CommanderTax = 8`, `PartnerTax = 9`). A regression that lowered the floor to 9 would still pass here. A case that inserts counters at `CounterIds::CommanderTax`/`PartnerTax` and expects `newCounterId() == CounterIds::FirstUserId` would nail the exact boundary.

cockatrice/src/game/player/player_event_handler.cpp
```diff
@@ -31,8 +31,10 @@
 #include <libcockatrice/protocol/pb/event_set_card_attr.pb.h>
 #include <libcockatrice/protocol/pb/event_set_card_counter.pb.h>
 #include <libcockatrice/protocol/pb/event_set_counter.pb.h>
+#include <libcockatrice/protocol/pb/event_set_counter_active.pb.h>
 #include <libcockatrice/protocol/pb/event_shuffle.pb.h>
 #include <libcockatrice/utility/color.h>
+#include <libcockatrice/utility/counter_ids.h>
```
ZeldaZach: This include looks unused now — nothing left in the file references `CounterIds` or `CounterNames` after the direct graphics calls were moved out of the event handler. Safe to drop.

ZeldaZach (commented):
Re-reviewed after the rounds of fixes. All eight inline comments from the last pass are properly addressed, and a few went further than I'd asked: the reserved-name spoofing rejection, the inactive-tax modify guard, and pulling the counter authorization out into static `evaluate*` helpers backed by real branch tests are all good calls. I traced the server auth logic, the graphics/logic separation, the counter active-state signal path, and the tax-counter layout, and it holds together well.

One thing worth fixing before merge: a use-after-free during `CommandZone` teardown (comment inline). The rest are smaller — a missing shortcut registration, a test-coverage gap on the create-counter guard, a boundary the id-allocation test doesn't quite pin, and a stale include.

There's also a known-and-documented edge I won't belabor: `playAndIncreaseTax` sends the tax `IncCounter` independently of whether the move is accepted, so a rejected cast can bump the tax without a card leaving the zone. The header comment already calls this out; coupling it to the move is a bigger server-side change and fine to defer.

DawnFire42:
I've addressed the previous five threads and fixed a few last things I found. I did as much as I could to find improvements but now I can't find anything without getting out of practical bounds. 
Please let me know if there's anything else I should fix. In the mean time I'll keep this rebased and up to date.

cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
```diff
@@ -102,13 +102,15 @@ void DlgCreateGame::sharedCtor()
...
     shareDecklistsOnLoadCheckBox = new QCheckBox(tr("Open decklists in lobby"));
+    enableCommandZoneCheckBox = new QCheckBox(tr("Enable command zone"));
 
     createGameAsJudgeCheckBox = new QCheckBox(tr("Create game as judge"));
 
     auto *gameSetupOptionsLayout = new QGridLayout;
     gameSetupOptionsLayout->addWidget(startingLifeTotalLabel, 0, 0);
     gameSetupOptionsLayout->addWidget(startingLifeTotalEdit, 0, 1);
     gameSetupOptionsLayout->addWidget(shareDecklistsOnLoadCheckBox, 1, 0);
+    gameSetupOptionsLayout->addWidget(enableCommandZoneCheckBox, 1, 1);
```
RickyRister: Looks kinda weird with the checkboxes next to each other. I still think it would look better if they were lined up vertically. 

Image

libcockatrice_network/libcockatrice/network/server/remote/game/server_player.h
```diff
@@ -24,6 +24,27 @@ class Server_Player : public Server_AbstractPlayer
     int newCounterId() const;
     void addCounter(Server_Counter *counter);
 
+    // Pure authorization/decision logic extracted from the corresponding cmd* handlers
```
RickyRister: use doc comments

BruebachL:
I have not forgotten this PR. Thank you for continuing to rebase and maintaining it. I am deliberating and trying to find the best way to reconcile this with changes going forward.

DawnFire42:
Thanks! and no worries, take your time I know it's a big chunk. I've also found a few more things to improve/fix in it, so I'm about to add some more commits in the next few days. I'll let yall know when I'm done with these next ones.

DawnFire42:
I've been having trouble with the implementation of the tax counters and their reserved ID. So I decided to rework them into cast  counts using a dedicated ID space so that, ultimately, a player could add as many as they want without breaking the other counters in the game. Although doing all of this in the same PR as the rest of the command zone has made it more difficult then required. 
The core features of the zone having been reviewed already, I feel like the best course of action is to split this in two parts, one with the command zone without the cast counts (tax counters), and then the cast counts by themselves later down the line. 
This PR's commit history has also become quite horrifying to work with... So I'm just going to create new PRs...
I hope this will make it easier to work with.

PR #6984, opened 2026-06-10
[Game][Graphics][Player] Add named zone lookup-map to player graphics.

BruebachL:
PlayerLogic stores all the zone data in a generic map `QMap<QString, CardZoneLogic *> zones;`, whereas PlayerGraphicsItem stores them all in named member variables. This introduces a slight discrepancy between the logic and the graphics API because we can fetch an arbitrary zone from the logic but we stall at converting that logic into a view, which means that while we can fetch logic items from graphics representations, we can't do the inverse.

- Add an equivalent `QMap<QString, CardZone *> zoneGraphicsItems;` to PlayerGraphicsItem
- populate with named member variables in initializeZones()
- Existing named member variables remain untouched, no callsites updated
- Expose a getter for zones by name

This change pretty much just adds a map and registers all the existing zones in it. It's purely additive. We should, eventually, think about yeeting off the named member variables and just fetching from the map in all the callsites where required, but this change is small and effective for solving a problem I have in the ongoing refactor.

### DIFF
--- a/cockatrice/src/game_graphics/player/player_graphics_item.cpp
+++ b/cockatrice/src/game_graphics/player/player_graphics_item.cpp
@@ -59,6 +59,9 @@ PlayerGraphicsItem::PlayerGraphicsItem(PlayerLogic *_player) : player(_player)
 
     initializeZones();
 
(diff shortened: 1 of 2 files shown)


RickyRister (approved):
sure

PR #6987, opened 2026-06-11
[UserContextMenu] Add confirmation dialog before kicking a player (#6689)

### LINKED ISSUE #6689: Confirmation On Kicking Players
I think there should be a confirmation window that appears when you try to kick a player. Since kicking players from a game is against the Cockatrice Discord's policy, it should give you a confirmation since at the moment, I can accidentally kick a player, especially when trying to add them to your ignore list, since both buttons are next to each other.

kongwu666:
## Related Ticket(s)
- Fixes #6689

Kicking a player from a game happens immediately without any confirmation. Since the "Kick" and "Add to ignore list" options are adjacent in the context menu, it's easy to accidentally kick a player. Kicking players is against the Cockatrice Discord's policy, so an accidental kick is problematic.

- Added a confirmation dialog (QMessageBox::question) before kicking a player from a game
- The dialog asks "Are you sure you want to kick this player from the game?" with Yes/No buttons
- No is the default selection to prevent accidental kicks
- The kick command is only sent if the user confirms

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
@@ -476,10 +476,15 @@ void UserContextMenu::showContextMenu(const QPoint &pos,
 
         client->sendCommand(client->prepareSessionCommand(cmd));
     } else if (actionClicked == aKick) {
-        Command_KickFromGame cmd;
-        cmd.set_player_id(playerId);
+        auto result = QMessageBox::question(static_cast<QWidget *>(parent()), tr("Kick Player"),
+                                            tr("Are you sure you want to kick this player from the game?"),
+                                            QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
+        if (result == QMessageBox::Yes) {
+            Command_KickFromGame cmd;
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
```diff
@@ -476,10 +476,14 @@ void UserContextMenu::showContextMenu(const QPoint &pos,
 
         client->sendCommand(client->prepareSessionCommand(cmd));
     } else if (actionClicked == aKick) {
-        Command_KickFromGame cmd;
-        cmd.set_player_id(playerId);
+        if (QMessageBox::question(static_cast<QWidget *>(parent()), tr("Kick Player"),
```
RickyRister: nit: I prefer to do the messageBox execution outside the if statement, saving the result to a variable, and only doing the comparison inside the if statement.

RickyRister (approved):
Works. LGTM

https://github.com/user-attachments/assets/22b8a82f-a2d4-4333-86f8-4d37babd31cd

RickyRister:
Also, you need to rebase to fix the windows build failure

PR #6992, opened 2026-06-13
[Game] Setting to restore old chat autofocus behavior

RickyRister:
## Related Ticket(s)
- Fixes regression introduced in #5791

When we implemented the card view search bar, we had to remove the line that set the focusPolicy of the game view to `NoFocus` in order to allow the search to get focus. However, that also means we no longer have the behavior of the game chat remaining focused even when the game scene is clicked.

When we asked for feedback about the change, there weren't any complaints.

However, I recently ran across a contingent of players for whom the removal of the autofocusing is an absolute deal breaker and is making them refuse to upgrade.

https://github.com/user-attachments/assets/c7bc1974-5eba-4561-a4df-c3be787cf178

- Add the new setting under Interface -> general interface settings
- When setting is enabled, we set the focusPolicy on `GameView` to `NoFocus` again, which prevents the game chat from losing focus when you click on the game scene.
  - Also hides the search bar, so users don't get confused about why it doesn't work
- Emits signals on settings change, so the change can take effect midgame. 

## Screenshots

Screenshot 2026-06-13 at 1 49 50 AM

### DIFF
--- a/cockatrice/src/game_graphics/game_view.cpp
+++ b/cockatrice/src/game_graphics/game_view.cpp
@@ -1,7 +1,6 @@
 #include "game_view.h"
 
 #include "../client/settings/cache_settings.h"
-#include "../game/game.h"
 #include "game_scene.h"
 
 #include <QAction>
@@ -196,9 +195,5 @@ void GameView::updateTotalSelectionCount(const QSize &viewSize)
  */
 void GameView::setFocusDisabled(bool disabled)
 {
-    if (disabled) {
-        setFocusPolicy(Qt::NoFocus);
-    } else {
-        setFocusPolicy(Qt::ClickFocus);
-    }
+    setFocusPolicy(disabled ? Qt::NoFocus : Qt::ClickFocus);
 }
--- a/cockatrice/src/game_graphics/game_view.cpp
+++ b/cockatrice/src/game_graphics/game_view.cpp
@@ -1,6 +1,7 @@
 #include "game_view.h"
 
 #include "../client/settings/cache_settings.h"
+#include "../game/game.h"
 #include "game_scene.h"
 
 #include <QAction>
@@ -34,7 +35,6 @@ GameView::GameView(GameScene *scene, QWidget *parent) : QGraphicsView(scene, par
 {
     setBackgroundBrush(QBrush(QColor(0, 0, 0)));
     setRenderHints(QPainter::TextAntialiasing | QPainter::Antialiasing);
-    setFocusPolicy(Qt::ClickFocus);
     setViewportUpdateMode(BoundingRectViewportUpdate);
 
     connect(scene, &GameScene::sceneRectChanged, this, &GameView::updateSceneRect);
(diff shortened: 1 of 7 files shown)


cockatrice/src/game_graphics/game_view.cpp
```diff
@@ -1,6 +1,7 @@
 #include "game_view.h"
 
 #include "../client/settings/cache_settings.h"
+#include "../game/game.h"
```
ZeldaZach: This include doesn't look necessary — nothing from `game.h` (the `Game` class) is referenced by the new code in this PR. The added logic only uses `SettingsCache` (already included via `cache_settings.h`) and Qt focus enums. Recommend dropping this include.

cockatrice/src/game_graphics/game_view.h
```diff
@@ -31,6 +31,7 @@ private slots:
     void stopRubberBand();
     void refreshShortcuts();
     void updateTotalSelectionCount(const QSize &viewSize = QSize());
+    void setFocusDisabled(bool focusEnabled);
```
ZeldaZach: The parameter name reads as the opposite of what it does. The function is `setFocusDisabled`, so `true` disables focus — but it's named `focusEnabled`, which implies the inverse. The implementation in `game_view.cpp` also uses `bool disabled`, so the declaration and definition disagree.

Suggest renaming this to `bool disabled` to match the .cpp signature and the actual semantics.

cockatrice/src/game_graphics/game_view.cpp
```diff
@@ -186,3 +189,16 @@ void GameView::updateTotalSelectionCount(const QSize &viewSize)
...
 }
+
+/**
+ * Disabling focus on the game view will allow chat to maintain the autofocusing behavior of pre 2.10.3,
+ * at the cost of disabling the zone view search bar.
+ */
+void GameView::setFocusDisabled(bool disabled)
+{
+    if (disabled) {
+        setFocusPolicy(Qt::NoFocus);
```
ZeldaZach: Minor: this could collapse to a one-liner without the if/else:

```cpp
setFocusPolicy(disabled ? Qt::NoFocus : Qt::ClickFocus);
```

Not a blocker — purely stylistic.

cockatrice/src/game_graphics/zones/view_zone_widget.cpp
```diff
@@ -75,6 +75,11 @@ ZoneViewWidget::ZoneViewWidget(PlayerLogic *_player,
         searchEditProxy->setZValue(ZValues::DRAG_ITEM);
         vbox->addItem(searchEditProxy);
 
+        // hide search bar if chat autofocus setting is enabled, since typing into it will no longer work anyway
+        searchEditProxy->setVisible(!SettingsCache::instance().getKeepGameChatFocus());
```
ZeldaZach, line 79: Should the existing `searchEdit.setFocus()` call up at line 70 (inside the `getFocusCardViewSearchBar()` branch) also be gated on `!getKeepGameChatFocus()`?

As written, if a user has both `focusCardViewSearchBar` and `keepGameChatFocus` enabled, opening a zone view will run `setFocus()` on the search edit before this `setVisible(false)` is applied — briefly stealing focus from the chat, which is exactly what this setting is meant to prevent. A user who has opted into chat-autofocus presumably never wants the (now hidden) search box to grab focus.

ZeldaZach (commented):
Nice fix — adds a useful escape hatch for users who relied on the pre-2.10.3 chat focus behavior. A few small things noted inline.

cockatrice/src/game_graphics/zones/view_zone_widget.cpp
RickyRister: Tested – it won't steal focus from chat

https://github.com/user-attachments/assets/9c345ba0-f48b-4eac-b159-bcb4ff6b5b16

BruebachL (approved):

Jndmchn:
Hey team, nice fix! I really noticed this when the update came through, it was a bit frustrating to get used to it. I just made a shortcut to the chat box (space bar) and got used to it. but now its even more streamline! So, i really appreciate the attention to detail you guys put in.

PR #6993, opened 2026-06-13
[CI] Utilize version resolution in install-qt-action + cache with full version key

tooomm:
> [!Note]
> - [x] New [`install-qt-action` released](https://github.com/jurplel/install-qt-action/releases) upstream (v4.4.0)
> - [x] Update this PR from master + fix conflicts
> - [x] Decide on "thin Qt" logic for macOS
Having the resolved/installed Qt version now as cache key allows to match fat versions reliably and not run stale because of wildcards input being preserved in the cache key. As the action does not support thinned binaries, this logic still needs to come from us. 
Having resolved version as action output (suggested upstream) will not help with restoring thinned Qt binaries in the current flow as latest version is only known after action run or if checked with aqtinstall manually 
>    - simple "cache fat Qt in install-qt-action + thin always"
>    - current "keep aqt + cache thinned Qt logic (4 extra steps)" <--
>    - "accept bigger binary size or favor universal Cockatrice installer anyway and not thin at all"
>    - maybe the upstream action will support thinning with a flag in the future

In former install-qt-action versions, cache was stored with exact version input (e.g. "6.11.*") in its key.
False positive cache hits on the generic wildcard resulted in the newest Qt versions not being used.

We added 2 extra steps with tool installs and script runs just to resolve our matrix input to newest version which the action was doing internally anyway. The cache was not reusing it and saved the wildcard input. It got fixed after reporting it.

- Go back to wildcard input to receive patch releases with fixes automatically (6.11.1 --> 6.11.2)
- Always save resolved full versions from wildcard inputs in the cache key (new action version)
  That way the cache is invalidated for new versions automatically and we're not locked into old versions even when using wildcard version inputs.

Example wildcard version inputs and their resolution (the action also supports SimpleSpec with ranges)
| Input Version (used to be in cache key) | Resolved Version (is now part of cache key) |
| --- | --- |
| `6` | 6.12.0 |
| `6.*` | 6.12.0 |
| `6.*.*` | 6.12.0 |
| `6.11` | 6.11.1 |
| `6.11.*` | 6.11.1 |
| `6.11.0` | 6.11.0 |

> [!IMPORTANT]
> Qt uses SemVer, we do not want to put a broad wildcards range including major/minor digit.
Using only 6 or e.g. 6.* will include beta releases of minor releases as aqtinstall does not differentiate between early versions (that are available month before actual releases, see Beta of Qt 6.12 which has its stable release planned for September only) and full releases. For patch releases with fixes, there are no Betas and we can put the wildcard there to benefit from fixes (and invalidate cached, older versions automatically) as soon as they are released. 

## Screenshots
- With old action and without our extra step to resolve wildcard versions manually to the latest in line, and feeding that into the install-qt-action to get the installed version as part of the cache key:
  
  `6.10.*` input --> installing `6.10.3`, but cache key keeps wildcard, matching on 6.10.4 release still, which will pull cache & keep using 6.10.3 version
  image

- With new action (and no extra steps, feeding the wildcard version directly will now include the resolved version in the cache key automatically)
  
  `6.*` input --> installing `6.12.0`, cache key uses resolved version, not matching on 6.12.1 release and new patch version will be installed + cached
  image


tooomm:
Just realized that this will push the issue on Mac down the line.

- On Windows it'll improve our pipeline and fix the cache matching outdated versions.

- On macOS however, we slim the fat Qt installation down to what we need in a separate step and cache per arch only.
As the install-qt-action provides no output variable for the resolved Qt version currently, the cache key for the thin qt is based on the input parameter for Qt version... which is now using a wildcard again to receive fixes on patch releases automatically.

  I decided to keep the current approach for macOS to still cache two "Qt-macOS-thin" going forward.

BruebachL (approved):

PR #6995, opened 2026-06-14
CI: Simplified build workflow

fearphage:
## Related Ticket(s)
- N/A

Making calculations with no dynamic values in Bash.

* Simplified the scripts by moving the calculations of the static values to the YAML. 
* Also used some updated GitHub environment variables to further simplify the output.
* Trimmed trailing whitespace (automatic)

> [!NOTE]
> I could add a linter for the trailing whitespace and a few other things if you'd like? Most of these seem to have come from #6959 (even though there was a whitespace clean up commit).

## Screenshots

N/A

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -53,7 +53,7 @@ jobs:
           RESOLVED_SHA: >-
             ${{
               case(
-                github.event_name == 'pull_reuqest',
+                github.event_name == 'pull_request',
                 github.event.pull_request.head.sha,
                 github.sha
               )
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -36,7 +36,7 @@ on:
 
 # Cancel earlier, unfinished runs of this workflow on the same branch (unless on release)
 concurrency:
-  group: ${{ github.workflow }} @ ${{ github.ref_name }}
+  group: "${{ github.workflow }} @ ${{ github.ref_name }}"
   cancel-in-progress: ${{ github.ref_type != 'tag' }}
 
 jobs:
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -36,7 +36,7 @@ on:
 
 # Cancel earlier, unfinished runs of this workflow on the same branch (unless on release)
 concurrency:
-  group: "${{ github.workflow }} @ ${{ github.ref_name }}"
+  group: ${{ github.workflow }} @ ${{ github.ref_name }}
   cancel-in-progress: ${{ github.ref_type != 'tag' }}
 
 jobs:
@@ -47,22 +47,24 @@ jobs:
       tag: ${{ steps.configure.outputs.tag }}
       sha: ${{ steps.configure.outputs.sha }}
 
-    steps:  
+    steps:
       - name: "Configure"
+        env:
+          RESOLVED_SHA: >-
+            ${{
+              case(
+                github.event_name == 'pull_reuqest',
+                github.event.pull_request.head.sha,
+                github.sha
+              )
+            }}
         id: configure
         shell: bash
         run: |
-          tag_regex='^refs/tags/'
-          if [[ $GITHUB_EVENT_NAME == pull-request ]]; then    # pull request
-            sha="${{github.event.pull_request.head.sha}}"
-          elif [[ $GITHUB_REF =~ $tag_regex ]]; then    # release
-            sha="$GITHUB_SHA"
-            tag="${GITHUB_REF/refs\/tags\//}"
-            echo "tag=$tag" >>"$GITHUB_OUTPUT"
-          else    # push to branch
-            sha="$GITHUB_SHA"
+          if [[ "$GITHUB_REF_TYPE" == 'tag' ]]; then    # release
+            echo "tag=$GITHUB_REF_NAME" >> "$GITHUB_OUTPUT"
           fi
-          echo "sha=$sha" >>"$GITHUB_OUTPUT"
+          echo "sha=$RESOLVED_SHA" >> "$GITHUB_OUTPUT"
 
       - name: "Checkout"
         if: steps.configure.outputs.tag != null
@@ -92,7 +94,7 @@ jobs:
         run: |
           args=()
           [[ $prerelease == yes ]] && args+=(--prerelease)
-          
+
           gh release create "$tag_name" --verify-tag --draft "${args[@]}" \
             --target "$target" \
             --title "$release_name" \
@@ -105,48 +107,48 @@ jobs:
         # The files in ".ci/$distro$version" correspond to the values given here
         include:
           - distro: Arch
-            
+
             allow-failure: yes
             package: skip    # We are packaged in Arch already
 
           - distro: Servatrice_Debian
             version: 12
-            
+
             package: DEB
             server_only: yes
             test: skip
 
           - distro: Debian
             version: 12
-            
+
             package: DEB
             test: skip    # Running tests on all distros is superfluous
 
           - distro: Debian
             version: 13
-            
+
             package: DEB
 
           - distro: Fedora
             version: 43
-            
+
             package: RPM
             test: skip    # Running tests on all distros is superfluous
 
           - distro: Fedora
             version: 44
-            
+
             package: RPM
 
           - distro: Ubuntu
             version: 24.04
-            
+
             package: DEB
             test: skip    # Running tests on all distros is superfluous
 
           - distro: Ubuntu
             version: 26.04
-            
+
             package: DEB
 
     name: ${{ matrix.distro }} ${{ matrix.version }}
@@ -203,7 +205,7 @@ jobs:
           args+=(--ccache "$CCACHE_SIZE")
           args+=(--cmake-generator "$CMAKE_GENERATOR")
           args+=(--suffix "$SUFFIX")
-          
+
           RUN --server --release --package "$package" "${args[@]}"
 
       # Delete used cache to emulate a ccache update. See https://github.com/actions/cache/issues/342
@@ -267,7 +269,7 @@ jobs:
           - os: macOS
             target: 13
             runner: macos-15-intel
-            
+
             ccache_eviction_age: 7d
             cmake_generator: Ninja
             make_package: 1
@@ -284,7 +286,7 @@ jobs:
           - os: macOS
             target: 14
             runner: macos-14
-            
+
             ccache_eviction_age: 7d
             cmake_generator: Ninja
             make_package: 1
@@ -300,7 +302,7 @@ jobs:
           - os: macOS
             target: 15
             runner: macos-15
-            
+
             ccache_eviction_age: 7d
             cmake_generator: Ninja
             make_package: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -47,22 +47,24 @@ jobs:
...
       sha: ${{ steps.configure.outputs.sha }}
 
-    steps:  
+    steps:
       - name: "Configure"
+        env:
+          RESOLVED_SHA: >-
+            ${{
+              case(
+                github.event_name == 'pull_reuqest',
```
ZeldaZach: Typo: `pull_reuqest` → `pull_request`. As written, the condition is always false, so `RESOLVED_SHA` always resolves to the default `github.sha` (the PR's synthetic merge commit) instead of the actual head SHA. The `case()` machinery here becomes dead.

.github/workflows/desktop-build.yml
```diff
@@ -36,7 +36,7 @@ on:
 
 # Cancel earlier, unfinished runs of this workflow on the same branch (unless on release)
 concurrency:
-  group: "${{ github.workflow }} @ ${{ github.ref_name }}"
+  group: ${{ github.workflow }} @ ${{ github.ref_name }}
```
ZeldaZach: Why drop the quotes here? Templated string values are safer to keep quoted — the `@` and space are fine unquoted today, but a future change to the format that introduces a YAML-significant character would silently break the group key.

ZeldaZach (commented):
Good direction — pushing the static branching into YAML/expression land and trimming bash boilerplate is a nice cleanup. CI confirms `case()` works (it's a function GHA added [Jan 2026](https://github.blog/changelog/2026-01-29-github-actions-smarter-editing-clearer-debugging-and-a-new-case-function/), worth a passing comment in the diff for future readers).

Two issues to address, plus a couple of nits:

**Blocking-ish:**
- Typo on line 56 — `github.event_name == 'pull_reuqest'` should be `'pull_request'`. The misspelling makes the condition always false, so on PR events `RESOLVED_SHA` always falls through to the default (`github.sha`, i.e. the synthetic merge commit) rather than the PR head SHA. That happens to match the *behavior* of the pre-PR code (which had a different typo — `pull-request` vs `pull_request` — that also never matched), so it isn't a regression. But the whole point of introducing the `case()` expression is to actually pick `head.sha` for PR events, and the typo defeats that. Inline comment below.

**Style:**
- Line 39: the `concurrency.group` value lost its surrounding quotes. YAML accepts the unquoted form here so this isn't a runtime issue, but quoting templated strings is the safer default — easier to spot in review and avoids a future foot-gun if the format ever includes a YAML-significant character. Inline comment below.

**Nits (not blocking):**
- The trailing-whitespace cleanup inflates the diff substantially. Splitting it into its own commit (or PR) would have made the actual logic change much easier to review. Yes please to a whitespace linter as you offered in the PR description.
- The macOS 14 and Pull languages job failures look unrelated to this diff — worth a re-run to confirm green checks before merge.

.github/workflows/desktop-build.yml
fearphage: All values that aren't some other value are strings in YAML. So if it isn't a number, boolean, etc., it's a string. Most of the strings in most of these files do not have quotes. I was matching this to those. I'm happy to restore them if needed.

> a future change to the format that introduces a YAML-significant character would silently break the group key

If you put in a bad character, things would stop parsing this as valid YAML. For instance if you change the `@` to `:`, it makes the YAML invalid. This isn't something small that can be missed. The document is invalid and first when you push it up, GitHub will notify you that your workflow is invalid YAML. This won't be a small change that you have to spot. It will break in a dramatic fashion that will be unmissable.

.github/workflows/desktop-build.yml
```diff
@@ -47,22 +47,24 @@ jobs:
...
       - name: "Configure"
+        env:
+          RESOLVED_SHA: >-
+            ${{
+              case(
+                github.event_name == 'pull_request',
+                github.event.pull_request.head.sha,
+                github.sha
+              )
+            }}
```
tooomm: Is there a strong reason you went for the multi-line variant instead straight forward single line?
I don't think readability is an issue, and we have some longer lines already.
```suggestion
        env:
          RESOLVED_SHA: ${{ case(github.event_name == 'pull_request', github.event.pull_request.head.sha, github.sha) }}
```

.github/workflows/translations-pull.yml
```diff
@@ -61,11 +61,16 @@ jobs:
...
+          PR_NUMBER: ${{ steps.create_pr.outputs.pull-request-number }}
+          PR_URL: ${{ steps.create_pr.outputs.pull-request-url }}
+          STATUS: >-
+            ${{
+              case(
+                steps.create_pr.outputs.pull-request-operation == 'none',
+                'unchanged',
+                steps.create_pr.outputs.pull-request-operation
+              )
+            }}
```
tooomm:
```suggestion
          STATUS: ${{ case(steps.create_pr.outputs.pull-request-operation == 'none', 'unchanged', steps.create_pr.outputs.pull-request-operation) }}
```

.github/workflows/translations-push.yml
```diff
@@ -77,11 +78,16 @@ jobs:
...
+          PR_NUMBER: ${{ steps.create_pr.outputs.pull-request-number }}
+          PR_URL: ${{ steps.create_pr.outputs.pull-request-url }}
+          STATUS: >-
+            ${{
+              case(
+                steps.create_pr.outputs.pull-request-operation == 'none',
+                'unchanged',
+                steps.create_pr.outputs.pull-request-operation
+              )
+            }}
```
tooomm:
```suggestion
          STATUS: ${{ case(steps.create_pr.outputs.pull-request-operation == 'none', 'unchanged', steps.create_pr.outputs.pull-request-operation) }}
```

tooomm (commented):
Thanks for having a look and providing a hardened version!

GH docs on `case() expression`: https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case

<br>

>I could add a linter for the trailing whitespace and a few other things if you'd like? 

I did enable GitHub Code Scanning (CodeQL) for GH Actions and workflows in #6958.
Do you know if such issues would be caught?

.github/workflows/desktop-build.yml
fearphage: My goal was to maximize readability and maintainability. I can collapse them though.

fearphage:
> Do you know if such issues would be caught?

I'm not too knowledgeable on the code scanning tooling. I don't know what it will catch. I was thinking more along the lines of a [YAML linter](https://github.com/adrienverge/yamllint).

tooomm:
> I'm not too knowledgeable on the code scanning tooling. I don't know what it will catch. I was thinking more along the lines of a [YAML linter](https://github.com/adrienverge/yamllint).

Found this interesting read on the CodeQL topic and security scans for GH Actions:
https://github.blog/security/application-security/how-to-secure-your-github-actions-workflows-with-codeql/

fearphage:
> Found this interesting read on the CodeQL topic and security scans

These two tools solve _very_ different things. My YAML linter suggestion was more about conformity and consistency across files. Like there's a linter rule to prevent trailing whitespace on lines for instance. 

I think the CodeQL workflows are also vaulable, but I don't believe they overlap very much. To restate, I don't think it's an either-or situation and having one doesn't remove the need for the other.

.github/workflows/desktop-build.yml
tooomm:
```suggestion
  group: "${{ github.workflow }} @ ${{ github.ref_name }}"
```

@fearphage Zach is referring with `YAML-significant character` to the [indicator characters in YAML](https://yaml.org/spec/1.2.2/#53-indicator-characters) like `{`, `}`, `[`, `]`, `&`, `*`, `#`, `?`, `|`, `-`, `<`, `>`, `=`, `!` etc.

I learned the hard way in #6959 that a step name can not contain square brackets `[]`.
Using quotes in the name value worked.

`@` is such a special character as well.
`The “@” ... (is) reserved for future use.`
So this workflow would fail at one point (with an error probably). GitHub is accepting it right now, or it does matter that it's not at the beginning (which was the case for my bracket placement).

Probably good to keep them already.

fearphage: > GitHub is accepting it right now, or it does matter that it's not at the beginning (which was the case for my bracket placement).

Your hunch was correct. The placement is significant: 

```yaml
# valid string
catpants: ${{ github.workflow }} @ ${{ github.ref_name }}
```

```yaml
# invalid YAML - replaced @ with : 
# this looks suspiciously like you're trying to inline an object/map reference
# So an error prevents ambiguity
catpants: ${{ github.workflow }} : ${{ github.ref_name }}
```

```yaml
# valid string - colon at start of string - no ambiguity so it's just a string
catpants: :${{ github.workflow }} @ ${{ github.ref_name }}
```

```yaml
# invalid YAML - reserved char at the start of the line
catpants: @${{ github.workflow }} @ ${{ github.ref_name }}
```

I'll re-wrap the string in quotes!

tooomm:
> These two tools solve _very_ different things. My YAML linter suggestion was more about conformity and consistency across files. Like there's a linter rule to prevent trailing whitespace on lines for instance.

I was thinking that CodeQL would do some lint checking as well, similar to Codacy. That is apparently not the case.
The article I shared (part 2) mentioned the fixes for code/command injection you just introduced in your other PR. 👍
CodeQL would have suggested similar changes probably.

We already do linting for our code files, CMake configuration and bash scripts via [desktop-lint.yml](https://github.com/Cockatrice/Cockatrice/blob/master/.github/workflows/desktop-lint.yml). Checking GitHub Action workflows makes sense as well. Do you know about [`actionlint`](https://github.com/rhysd/actionlint) (it does some extra checks on top)?
There is also a meta linter action called [`megalinter`](https://github.com/oxsecurity/megalinter), which covers all our existing tools used - as well as your linked yaml linter.

> I think the CodeQL workflows are also vaulable, but I don't believe they overlap very much. To restate, I don't think it's an either-or situation and having one doesn't remove the need for the other.

If you want to check how we currently do linting via format.sh and how to extend that for yamllint in a new PR - That would be great!
Adding a dedicated actionlint step in the lint workflow, or exploring megalinter could come later.

<br>

The translation job is expected to fail because the transifex token can not be accessed btw.

fearphage:
> Do you know about [`actionlint`](https://github.com/rhysd/actionlint) (it does some extra checks on top)?

I'm a huge fan of `actionlint`. It's also different, but worth adding. It does syntax-aware checks (similar to CRD checks) for adherence to the workflow YAML spec. It will catch things like properties in the wrong place or missing required properties or properties that have been renamed/removed. It also runs `shellcheck` on inline shell scripts.

> There is also a meta linter action called [megalinter](https://github.com/oxsecurity/megalinter), which covers all our existing tools used - as well as your linked yaml linter.

I haven't used megalinter, but I've used the similar (based on the name) [super-linter](https://github.com/super-linter/super-linter). Super linter includes linters for virtually everything -- markdown, YAML, C++, grammar checks, you name it. You can check the [full list of languages](https://github.com/super-linter/super-linter#supported-linters-and-formatters).

tooomm (approved):
All comments from Zach are addressed, merging and moving on with the improvements.

Thanks @fearphage!

PR #6996, opened 2026-06-14
CI: replaced dynamically injected variables with environment variables

fearphage:
This prevents some injection attacks and makes all the scripts valid Bash.

## Related Ticket(s)
- N/A

Using dynamically injected values opens the door for script injection attacks. Those are not possible with environment variables.

* GitHub docs: [warning about their usage](https://docs.github.com/en/actions/concepts/security/script-injections)
* News story: [A GitHub Issue Title Compromised 4k Developer Machines](https://grith.ai/blog/clinejection-when-your-ai-tool-installs-another)

* Replaced script injects with environment variables

> [!IMPORTANT]
>
> Some of these variable injections were left behind because they are addressed in another parallel ticket (#6995).

## Screenshots
N/A

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -211,9 +211,10 @@ jobs:
         if: github.ref == 'refs/heads/master' && steps.ccache_restore.outputs.cache-hit
         continue-on-error: true
         env:
+          CACHE_PRIMARY_KEY: ${{ steps.ccache_restore.outputs.cache-primary-key }}
           GH_TOKEN: ${{ github.token }}
         run: |
-          if gh cache delete --repo ${{ github.repository }} ${{ steps.ccache_restore.outputs.cache-primary-key }}; then
+          if gh cache delete --repo "$GITHUB_REPOSITORY" "$CACHE_PRIMARY_KEY"; then
             echo "Cache deleted successfully"
           fi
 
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):
Solid hardening PR — straight application of GitHub's [recommended pattern for avoiding script injection in workflows](https://docs.github.com/en/actions/concepts/security/script-injections). Verified:

- All four interpolation patterns are correctly migrated: `github.repository` → built-in `$GITHUB_REPOSITORY` (no env decl needed), and `steps.ccache_restore.outputs.cache-primary-key`, `steps.build.outputs.path`, `matrix.qt_version` are each declared in `env:` and dereferenced with double-quotes.
- Bonus fix: the prior `gh cache delete --repo ${{ github.repository }} ${{ steps.ccache_restore.outputs.cache-primary-key }}` lines were also un-quoted, so the migration tightens word-splitting too — not just injection.
- No behavior change for the value substitution itself; only the substitution mechanism shifts from build-time templating to runtime env lookup.

Leftover scan — items not in this PR's scope, mostly fine but worth noting:
- `desktop-build.yml:57` still has `sha="${{github.event.pull_request.head.sha}}"` inside a `run:` block. The guarding condition is `$GITHUB_EVENT_NAME == pull-request` (hyphenated) while the actual event name is `pull_request` (underscored), so the branch is unreachable; and `head.sha` is a 40-char hex SHA so even if it ran, injection would be infeasible. Safe to leave, but the dead branch is a latent foot-gun if someone fixes the typo without revisiting the assignment. Could roll into #6995.
- All other `${{ matrix.* }}` occurrences in the file are in `name:`/`if:`/`runs-on:`/`with:`/`env:` contexts, not `run:` shell bodies, so they're not injection vectors.

Approving.

tooomm:
Merging following Zach's comment.

The pull-request typo is covered in the other PR already.

PR #6997, opened 2026-06-14
[Card] Add facedown property to CardRelation

RickyRister:

One user has been asking for a way to create tokens that have the face-down property by default, similar to how tokens currently has the persistent property.

Added `facedown` attribute to `relation` and `reverse-relation` tags for card xml 4 (xml 3 is unchanged)

For example

```
<card>
 <name>green Hound 22LAW</name>
...
 <reverse-related facedown="true">Dogged Pioneer</reverse-related>
 <reverse-related facedown="true">Scion of the Pack</reverse-related>
</card>
```

Or

```
<card>
    <name>Professor's Notes</name>
...
    <related persistent="persistent" facedown="facedown">Fatigue</related>
</card>
```

Relations with that tag will automatically create the token face-down

https://github.com/user-attachments/assets/5f1d78d3-c297-4a4e-8ec8-2213c66d7ca9


libcockatrice_card/libcockatrice/card/relation/card_relation.h
```diff
@@ -151,6 +154,16 @@ class CardRelation : public QObject
     {
         return isPersistent;
     }
+
+    /**
+     * @brief Returns whether the relation creates the token face-down.
+     *
+     * @return True if persistent, false otherwise.
```
ZeldaZach: Copy-paste from `getIsPersistent` — should read "True if face-down, false otherwise." Also the `@brief` above says "face-down" (hyphenated) while the field comment at line 34 uses "facedown" — picking one would be nice.

ZeldaZach (approved):
Threading looks clean. Verified:

- All five `createCard` callsites in `player_actions.cpp` are updated; default `faceDown=false` preserves prior behavior for any caller not using the new param.
- `refreshCachedReverseRelatedCards` (`card_database.cpp:88`) propagates the new flag when synthesizing reverse relations — important since reverse-related is a primary use case in the PR description.
- `cockatrice_xml_3.cpp` is intentionally not updated; it doesn't pass `persistent` either (line 272), so omitting `facedown` matches the legacy parser's existing capability gap. Worth a one-liner in the PR description acknowledging xml_3 won't honor the attribute.
- XML write/read uses the same `facedown="facedown"` token + `hasAttribute("facedown")` shape as `persistent`, consistent with the established convention.

One in-line nit below. Pre-existing nit: `card_relation.cpp` is still missing a trailing newline.

PR #6998, opened 2026-06-14
[CardDatabase] Pass CardInfoPtr by const ref

RickyRister:
## Related Ticket(s)
- Follow-up to #6940

Some of the methods in `CardDatabase` don't pass the `QSharedPointer` params by const ref even though they could. This means it's doing an unnecessary reference count increment on every call.

Usually, this sort of micro-optimization wouldn't matter, but the aforementioned methods *are* called every time a card is added to the database, so there might actually be some visible effects here as the size of the card database grows.

- Make methods pass the `QSharedPointer` params by const ref


ZeldaZach (approved):
Safe and worthwhile micro-opt. Verified:

- All four methods are wired via function-pointer `connect` in `card_database_loader.cpp:25-27` (`&CardDatabase::addCard` / `addSet` / `addFormat`), so the parameter widening doesn't break any SIGNAL/SLOT string matching.
- Connections use `Qt::DirectConnection`, so no queued-call argument copying to worry about.
- `removeCard` is a regular public method (only intra-file caller at `card_database.cpp:46`), trivially safe.

One follow-up worth considering (out of scope for this PR): the matching signals in `card_database_parser.h:82,85,87` still declare `CardInfoPtr card` / `CardSetPtr set` / `FormatRulesPtr format` by value. Each `emit` therefore still pays a refcount bump on the emitter side before the const-ref slot gets it. Matching those signals to `const CardInfoPtr &` etc. would close the loop.

Nit (pre-existing, not from this PR): `card_database.cpp` is still missing a trailing newline — easy to fix in a follow-up if you're touching the file.

PR #7000, opened 2026-06-14
[Game] Fix facedown predefined tokens leaking tablerow

### LINKED ISSUE #6999: Create face-down token with predefined token leaks tablerow info
**System Information:**

Client Version: 3.0.1 (2026-05-23)
Client Operating System: macOS Sequoia (15.7.1)
Build Architecture: ARM
Qt Version: 6.11.0
System Locale: en_CA
Install Mode: Standard

When you create a face-down predefined token, the token will be created in the appropriate tablerow. This leaks info about what card type the token could possibly be. 

I expected face-down tokens creation to have the same behavior as "play face down", where it always plays it to the top tablerow.

https://github.com/user-attachments/assets/9759e945-b00b-4b47-83d6-9e0e1ae434cb

**Steps to reproduce:**
- Get into a game
- Open the create token menu
- Select a predefined token from the token menu
- Check "create face down"
- Notice that the face-down token is created in the appropriate tablerow

RickyRister:
## Related Ticket(s)
- Fixes #6999

Client is always setting the the tablerow of the create token command to the value given by the CardInfo, regardless of whether the token is being created facedown

- When the token is facedown, the tablerow will always be 2
  - Same behavior that we use for normal "play face down" 

https://github.com/user-attachments/assets/48cad458-c7ca-48f6-9227-033ffd4e2fa6

### DIFF
--- a/cockatrice/src/game/player/player_actions.cpp
+++ b/cockatrice/src/game/player/player_actions.cpp
@@ -882,7 +882,8 @@ void PlayerActions::actCreateToken(TokenInfo tokenToCreate)
     ExactCard correctedCard = CardDatabaseManager::query()->guessCard({lastTokenInfo.name, lastTokenInfo.providerId});
     if (correctedCard) {
         lastTokenInfo.name = correctedCard.getName();
-        lastTokenTableRow = TableZone::tableRowToGridY(correctedCard.getInfo().getUiAttributes().tableRow);
+        int tableRow = lastTokenInfo.faceDown ? 2 : correctedCard.getInfo().getUiAttributes().tableRow;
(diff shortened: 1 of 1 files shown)


ZeldaZach:
🎉 PR #7000 — congrats on hitting a milestone number!

ZeldaZach (approved):
Fix looks correct. The branch matches the existing face-down handling at `player_actions.cpp:81` (`playCard`) and `player_actions.cpp:121` (`playCardToTable`), which both use the same `faceDown ? 2 : info.getUiAttributes().tableRow` shape — so predefined tokens now behave the same way as normal face-down plays.

No in-line comments — the magic `2` is pre-existing in this file (same form, two other sites), so factoring it into a named constant is out of scope for this fix.

PR #7007, opened 2026-06-18
Add 'facedown' as a valid attribute of related/reverse-related

SlightlyCircuitous:
Related to https://github.com/Cockatrice/Cockatrice/pull/6997

Updates the v4 cards.xsd file (used by xml linters) to reflect that `facedown` is now a valid attribute of related/reverse-related.


RickyRister:
Merge it into my branch instead

SlightlyCircuitous:
Closed in favor of the above

PR #7008, opened 2026-06-19
[ThemeManager/PaletteEditor] Read user path, then fall through to system, only write to user

BruebachL:
Palette changes do not save for default themes because they are installed in the system, rather than the user location.

- Load system first, then load user and overlay on top
- Also allow for editing the style, which makes palette styled default themes not look ass (i.e. enables fusion)

### DIFF
--- a/cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
+++ b/cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
@@ -19,13 +19,15 @@
 PaletteEditorDialog::PaletteEditorDialog(const QString &_themeDirPath, const QString &_themeName, QWidget *parent)
     : QDialog(parent), themeDirPath(_themeDirPath), themeName(_themeName)
 {
+    userThemeDirPath = ThemeManager::userThemeDirFor(themeName);
+
     setMinimumSize(740, 220);
     setupUi();
 
     // Load both scheme configs upfront so switching is instant
     loadSchemes();
 
-    loadedScheme = themeManager->isDarkMode(themeDirPath) ? "Dark" : "Light";
+    loadedScheme = themeManager->isDarkMode(themeDirPath, userThemeDirPath) ? "Dark" : "Light";
 
     schemeComboBox->blockSignals(true);
     schemeComboBox->setCurrentText(loadedScheme);
@@ -33,7 +35,6 @@ PaletteEditorDialog::PaletteEditorDialog(const QString &_themeDirPath, const QSt
 
     paletteGrid->loadPalette(workingConfig[loadedScheme]);
     seedAccentFromScheme(loadedScheme);
-
     retranslateUi();
 }
 
@@ -162,15 +163,12 @@ void PaletteEditorDialog::retranslateUi()
     setWindowTitle(tr("Palette Editor — %1").arg(themeName));
     titleLabel->setText(tr("<b>Palette Editor</b> &nbsp;·&nbsp; %1").arg(themeName));
 
-    // Revert button only makes sense when the theme ships default palette files
-    const bool hasDefault = PaletteConfig::fromDefault(themeDirPath, "Light").hasPalette() ||
-                            PaletteConfig::fromDefault(themeDirPath, "Dark").hasPalette();
-    revertButton->setEnabled(hasDefault);
-    if (!hasDefault) {
-        revertButton->setToolTip(tr("This theme ships no default palette files"));
-    } else {
-        revertButton->setToolTip(tr("Replace current colours with the theme author's defaults"));
-    }
+    const bool hasUserOverride =
+        QFile::exists(QDir(userThemeDirPath).absoluteFilePath(PaletteConfig::fileName("Light"))) ||
+        QFile::exists(QDir(userThemeDirPath).absoluteFilePath(PaletteConfig::fileName("Dark")));
+    revertButton->setEnabled(hasUserOverride);
+    revertButton->setToolTip(hasUserOverride ? tr("Delete your custom palette and restore the shipped defaults")
+                                             : tr("No custom palette overrides exist for this theme"));
 
     schemeComboBox->setToolTip(tr("Switch between the light and dark palette files"));
     editingLabel->setText(tr("Editing:"));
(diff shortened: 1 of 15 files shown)


cockatrice/src/interface/theme_manager.cpp
```diff
@@ -379,6 +441,20 @@ void ThemeManager::themeChangedSlot()
     emit themeChanged();
 }
 
+void ThemeManager::onColorSchemeChanged()
```
ZeldaZach, line 444: `onColorSchemeChanged()` is never connected or called. The constructor still wires the OS `colorSchemeChanged` signal to `themeChangedSlot` (not this slot), and nothing else invokes it. Since `systemPalette` is snapshotted once in the constructor (line 105) and this is the only method that re-captures it, a runtime OS light↔dark switch never refreshes `systemPalette`. For any non-Fusion theme (`base = systemPalette`, line 354) `applyStyleAndPalette` then re-applies the stale startup palette, so the app keeps its original light/dark base after the OS flips. Either `connect(...colorSchemeChanged, this, &ThemeManager::onColorSchemeChanged)` or fold the `systemPalette` refresh into `themeChangedSlot`.

cockatrice/src/interface/theme_manager.h
```diff
@@ -63,7 +64,11 @@ class ThemeManager : public QObject
 public:
     bool isBuiltInTheme();
     bool isDarkMode(const QString &themeDirPath);
```
ZeldaZach, line 66: This 1-arg `isDarkMode` declaration no longer has a definition — the old body was replaced by the 2-arg `static` overload below, which already defaults `userDirPath = {}`. Leaving the stale declaration is a latent trap: a call written as `isDarkMode(path)` becomes ambiguous between this member and the static overload (or resolves to this and fails to link). Please remove it.

cockatrice/src/interface/theme_config.cpp
```diff
@@ -248,10 +248,6 @@ PaletteConfig PaletteConfig::fromDefault(const QString &themeDirPath, const QStr
     PaletteConfig cfg =
         fromFile(dir.absoluteFilePath(wantDark ? "palette-default-dark.toml" : "palette-default-light.toml"));
 
-    if (!cfg.hasPalette()) {
-        cfg = fromFile(dir.absoluteFilePath(wantDark ? "palette-default-light.toml" : "palette-default-dark.toml"));
-    }
-
     return cfg;
```
ZeldaZach, line 251: Dropping the opposite-scheme fallback here changes behavior for themes that ship only one default file. All of `Default`, `Fabric`, `Leather`, `Plasma`, `VelvetMarble` ship **only** `palette-default-dark.toml` yet set `ColorScheme = Light` in `theme.cfg`. Under the default Light scheme the resolution is now `fromScheme(Light)` miss → `fromDefault(Light)` miss (no `palette-default-light.toml`) → empty, so the shipped default palette is never applied and the theme falls through to the plain app palette. Previously the removed fallback loaded the dark file. If this regression is intended, either ship a matching `palette-default-light.toml` or set `ColorScheme = Dark`; otherwise the dark files are dead weight (see the toml comment).

cockatrice/themes/Fabric/palette-default-dark.toml
```diff
@@ -0,0 +1,63 @@
+[Palette]
```
ZeldaZach, line 1: This file is byte-identical to the new `Leather`, `Plasma`, `VelvetMarble` copies (and to `Default/palette-default-dark.toml`) — same MD5. So these "per-theme" dark palettes carry no theme-specific colors and must be kept in lockstep on any future tweak. Combined with the `fromDefault` change above (these themes ship `ColorScheme = Light` and no light default), none of them is actually loaded under the shipped scheme. Consider a single shared base default rather than four identical copies, and confirm which scheme should load them.

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -137,45 +149,87 @@ bool ThemeManager::isBuiltInTheme()
...
+    if (!availableThemes.contains(FUSION_THEME_NAME)) {
+        availableThemes.insert(FUSION_THEME_NAME, userDir.absoluteFilePath("Fusion"));
+    }
+
     return availableThemes;
 }
 
+ThemeConfig ThemeManager::effectiveThemeConfig(const QString &themeName)
+{
+    const QString dirPath = getAvailableThemes().value(themeName);
```
ZeldaZach, line 200: `effectiveThemeConfig()` calls `getAvailableThemes()` purely for a single `.value(themeName)` lookup, but `getAvailableThemes()` clears and re-enumerates both theme directories (`QDir::entryList` on the system dir + user AppData dir) on every call. `effectiveThemeConfig` runs on the live-preview path (`previewPalette` → `onApply`, fired on each palette-editor edit), so dragging color pickers repeatedly re-scans the filesystem plus re-reads two `theme.cfg` files. Consider caching the theme map (or passing the already-known `themeDirPath` in) so preview doesn't hit disk per interaction.

cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
```diff
@@ -282,14 +293,50 @@ void PaletteEditorDialog::onReset()
...
+    if (reply != QMessageBox::Yes) {
         return;
     }
+
+    QFile::remove(filePath);
+
+    // Reload via fallthrough: system palette → system default → app palette
+    PaletteConfig def = PaletteConfig::fromScheme(themeDirPath, loadedScheme);
+    if (!def.hasPalette()) {
+        def = PaletteConfig::fromDefault(themeDirPath, loadedScheme);
```
ZeldaZach, line 317: This user → system-palette → system-default fallthrough is now copy-pasted in three places: here, `ThemeManager::themeChangedSlot`, and `PaletteEditorDialog::loadSchemes`. They've already diverged — this copy adds a 4th `qApp->palette()` fallback the other two lack, so the resolution order differs by code path. Worth extracting a shared `PaletteConfig resolvePalette(userDir, systemDir, scheme)` helper so precedence stays consistent.

ZeldaZach (commented):
Review of the palette user/system fallthrough rework. The core direction (read user → system, only ever write to the user dir) looks right. A few things worth addressing before merge — one functional (the new OS-scheme refresh path is never wired up), plus a config inconsistency in the shipped themes and some duplication/efficiency cleanups.

ZeldaZach (approved):
Approving — the user/system fallthrough direction is right and the write-to-user-dir fix is sound. The inline comments (notably the never-connected `onColorSchemeChanged` and the `fromDefault` fallback change vs. the shipped `ColorScheme = Light` themes) are worth a look but non-blocking.

PR #7009, opened 2026-06-19
Unify counter clamp arithmetic into shared addClamped() helper

DawnFire42:
Follow-up to #6879: resolves the deferred \todo to deduplicate counter clamp arithmetic

I did this through working on the command zone, but it deserves to be a separate PR

 - Rename MAX_COUNTERS_ON_CARD to MAX_COUNTER_VALUE, I did this here because this clamps the value of the counter and not the number of counter types on the card. The clamping is still only applied on counters on cards. I didn't want to name it MAX_CARD_COUNTER_VALUE either because the commander tax counters are going to use this too.

 - Add addClamped() in new header clamped_arithmetic.h; uses a 64-bit intermediate so the addition cannot overflow int.
- Use it in Server_Card::incrementCounter() (clamps [0, MAX_COUNTERS_ON_CARD]) and Server_Counter::incrementCount() (clamps [INT_MIN, INT_MAX]), removing the duplicated overflow-safe logic and its keep-in-sync TODO.
 - Inline incrementCount() into server_counter.h; server_counter.cpp now holds only the constructor and getInfo().
 - Clarify the card-counter bounds comment in trice_limits.h.
 - Add direct unit tests for addClamped() helper
 - Replace the raw oldValue + offset sum with addClamped(), clamping to [0, MAX_COUNTER_VALUE] without overflow.

### DIFF
--- a/libcockatrice_utility/libcockatrice/utility/clamped_arithmetic.h
+++ b/libcockatrice_utility/libcockatrice/utility/clamped_arithmetic.h
@@ -7,8 +7,8 @@
 /**
  * @brief Overflow-safe clamped addition: returns value + delta bounded to [minValue, maxValue].
  *
- * Uses a 64-bit intermediate so the addition itself cannot overflow int. Shared by the
- * counter arithmetic in Server_Card and Server_Counter so both stay in sync.
+ * Uses a 64-bit intermediate so the addition cannot overflow int. Shared by the counter
+ * arithmetic in Server_Card, Server_Counter, and PlayerActions.
  *
  * @note Requires minValue <= maxValue. Bounds come from trusted compile-time call sites;
  *       qBound() asserts this internally in debug builds.
--- a/libcockatrice_utility/libcockatrice/utility/clamped_arithmetic.h
+++ b/libcockatrice_utility/libcockatrice/utility/clamped_arithmetic.h
@@ -1,0 +1,22 @@
+#ifndef CLAMPED_ARITHMETIC_H
+#define CLAMPED_ARITHMETIC_H
+
+#include <QtGlobal>
+#include <cstdint>
+
+/**
+ * @brief Overflow-safe clamped addition: returns value + delta bounded to [minValue, maxValue].
+ *
+ * Uses a 64-bit intermediate so the addition itself cannot overflow int. Shared by the
+ * counter arithmetic in Server_Card and Server_Counter so both stay in sync.
+ *
+ * @note Requires minValue <= maxValue. Bounds come from trusted compile-time call sites;
+ *       qBound() asserts this internally in debug builds.
+ */
+inline int addClamped(int value, int delta, int minValue, int maxValue)
+{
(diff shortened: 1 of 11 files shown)


DawnFire42:
I was confused by trice_limits.h at first from its name. So I left a todo in there to separate it. I'll do it in a separate PR unless you want to keep it the way it is. Let me know, I can either remove the todo or do it after this PR.

libcockatrice_utility/libcockatrice/utility/trice_limits.h
```diff
@@ -15,11 +18,13 @@ constexpr uint MAXIMUM_DIE_SIDES = 1000000;
...
-// Card counter value bounds [0, MAX_COUNTERS_ON_CARD].
-// Counters on cards (e.g., +1/+1 counters, charge counters) are non-negative physical game objects.
+// Card counter value bounds [0, MAX_COUNTER_VALUE].
+// This caps an individual counter's VALUE (e.g. a +1/+1 counter at 999), not how many counters a card holds.
+// Applies to card counters only; player counters (Server_Counter) are unbounded and may go
+// negative (e.g. life total), saturating only at the int range.
 // The max of 999 is a display constraint (3-digit rendering) and reasonable gameplay limit.
 // Server enforces these bounds; client may also check for UX optimization.
-constexpr int MAX_COUNTERS_ON_CARD = 999;
+constexpr int MAX_COUNTER_VALUE = 999;
```
RickyRister, line 32: So this still only applies to card counters? Why remove `card` from the name then?
You can rename it to `MAX_CARD_COUNTER_VALUE` if you think the previous name was too ambiguous.

DawnFire42: it's because the commander tax counters will use the same functions. They're just not here yet.

RickyRister: Is there a way to reword the comment in this PR such that it doesn't imply it's just meant for card counters?

RickyRister: Also, nit: You can change the comment to use docstrings

libcockatrice_utility/libcockatrice/utility/clamped_arithmetic.h
```diff
@@ -0,0 +1,22 @@
...
+#define CLAMPED_ARITHMETIC_H
+
+#include <QtGlobal>
+#include <cstdint>
+
+/**
+ * @brief Overflow-safe clamped addition: returns value + delta bounded to [minValue, maxValue].
+ *
+ * Uses a 64-bit intermediate so the addition cannot overflow int. Shared by the counter
+ * arithmetic in Server_Card, Server_Counter, and PlayerActions.
```
RickyRister: don't directly call out classes in utility code, since we might forgot to keep it updated.
Just mention how it's used both in the client and server code for counters that are bounded.

RickyRister (approved):
Looks good otherwise

PR #7017, opened 2026-06-26
CI: Use default Qt arch values

tooomm:
Default values of the tool were again defined during Qt installation

- Do not specifically mention values that are used anyway. We had to do this at one point with older versions and when we had 32bit and 64bit builds. Now the defaults are correct for us and picked according to platform and Qt version.

  Simplifies the workflow file a little bit more. Removes something we do not need to worry about.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -271,7 +271,6 @@ jobs:
             override_target: 13
             package_suffix: "-macOS13_Intel"
             qt_version: 6.11.0
-            qt_arch: clang_64
             qt_modules: qtimageformats qtmultimedia qtwebsockets
             soc: Intel
             type: Release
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):
Verified the Qt arch defaults match the removed explicit values:

- **Windows:** `install-qt-action@v4` defaults to `win64_msvc2022_64` for Qt `>= 6.8.0`, so with `qt_version: 6.11.0` the computed default is identical to the removed value.
- **macOS:** the action leaves `arch` empty for the mac host and aqtinstall then defaults desktop Qt 6 to `clang_64` (correct for both Intel and Apple-silicon runners, since Qt 6 mac packages are universal under the single `clang_64` identifier).

No dangling `qt_arch` references remain anywhere in the workflow. Clean simplification — LGTM.

PR #6923, opened 2026-05-22
Add subtype breakdown counter for card selection

DawnFire42:
## Related Ticket(s)
- Closes #4419 

## Short roundup of the initial problem
- I saw this issue that was related to the card counter I made a few month's ago and thought it'd be a really helpful addition to my feature. 
- You will find the sorting behaviour somewhat opinionated in the fact that it brings higher numbers closer to the total (ie: on the bottom) and group types together (ie: creatures together, artifacts together, etc..) while omitting to actually write the supertype themselves (creature, artifact, etc...)  in order to avoid cluttering that list with useless info (the info itself isn't useless but having it there felt useless to me).
- This is a barebone implementation, actual display of the list could be improved later, but the foundations should all be there.

## What will change with this Pull Request?
  Display a categorised count of creature subtypes (and other card type
  subtypes) when multiple cards are selected. The breakdown appears above
  the total selection counter in the bottom-right corner.

  Subtypes are grouped by main card type and sorted by frequency, with
  the most common subtypes positioned adjacent to the total count for
  quick reference. The feature can be toggled via a new checkbox in
  Settings > User Interface.

PR #7021, opened 2026-06-27
[Game] Refactor subtype tally code to be more generic

RickyRister:
## Related Ticket(s)
- Follow-up to #6923

I want to refactor the existing subtype tally code to be more generic so that it can support multiple tally types in the future.

- Move tally code to file in `game_graphics/tally`
  - Split generic tally code and subtype-specific computation into separate files
  - Introduce `TallyType` enum and switch over it, to support more tally types in the future
- Rename code in `GameView` to remove references to subtype and just use the term tally


cockatrice/src/game_graphics/game_view.cpp
```diff
@@ -248,28 +247,31 @@ void GameView::updateTotalSelectionCount(const QSize &viewSize)
...
     if (!SettingsCache::instance().getShowSubtypeSelectionTally() || count <= 1) {
-        subtypeTallyContainer->hide();
-        cachedSubtypeEntries.clear();
+        tallyContainer->hide();
+        cachedTallyRows.clear();
         return;
     }
 
+    TallyType tallyType =
+        SettingsCache::instance().getShowSubtypeSelectionTally() ? TallyType::Subtypes : TallyType::None;
```
DawnFire42, line 250: The early return above already guarantees `getShowSubtypeSelectionTally()` is `true` here, so this ternary always evaluates to `Subtypes` and the `: TallyType::None` branch is unreachable — we're also reading the setting twice. Might be worth consolidating the "which tally is active?" into one place? 
Either:

- compute `tallyType` first, then guard with:

```cpp
if (tallyType == TallyType::None || count <= 1) {
    tallyContainer->hide();
    cachedTallyRows.clear();
    return;
}
```

- or, if `None` isn't reachable yet, just `TallyType tallyType = TallyType::Subtypes;`

cockatrice/src/game_graphics/tally/tally.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "subtype_tally.h"
+
+QList<TallyRow> Tally::compute(const QList<CardItem *> &cards, const TallyType type)
+{
+    switch (type) {
+        case TallyType::None:
+            return {};
+        case TallyType::Subtypes:
+            return SubtypeTally::countSubtypes(cards);
+        default:
```
DawnFire42: Both `TallyType` values already have cases, so the `default:` is redundant and it suppresses `-Wswitch`. A future tally type without a case would silently fall through to `{}` instead of getting a compile-time warning. With the Debug with `-Wall -Werror`, dropping `default:` turns that into a build-time safety net. Removing `default:` needs a trailing `return {};` after the switch, otherwise GCC fires `-Wreturn-type` under `-Werror`.

```cpp
switch (type) {
    case TallyType::None:     return {};
    case TallyType::Subtypes: return SubtypeTally::countSubtypes(cards);
}
return {};
```

cockatrice/src/game_graphics/tally/subtype_tally.h
```diff
@@ -0,0 +1,26 @@
+#ifndef SELECTION_SUBTYPE_TALLY_H
```
DawnFire42: nit: The include guard `SELECTION_SUBTYPE_TALLY_H` was carried over from the old `selection_subtype_tally.h` — the namespace, filename, and return type all got renamed but not the guard. 
`COCKATRICE_SUBTYPE_TALLY_H` would match the sibling `tally.h`.

cockatrice/src/game_graphics/tally/tally.cpp
```diff
@@ -0,0 +1,15 @@
...
+            return {};
+    }
+}
```
DawnFire42: nit: missing a trailing newline at end of file.

DawnFire42 (commented):
Hey that's pretty nice, thanks for upgrading this! 
I couldn't help but to check it out, I found a few small things I could suggest if that can help

cockatrice/src/game_graphics/game_view.cpp
RickyRister: `compute` will return empty if the type is `None`. I'll just consolidate the check

cockatrice/src/game_graphics/tally/subtype_tally.cpp
```diff
@@ -58,7 +65,12 @@ QList<SubtypeEntry> countSubtypes(const QList<CardItem *> &cards)
         return a.name < b.name;
     });
 
-    return entries;
+    // convert entries into TallyRows
+    QList<TallyRow> rows = QList<TallyRow>(entries.size());
```
ZeldaZach: This constructor form won't compile under Qt5. `QList` only gained a sized constructor (`QList(qsizetype)`) in Qt6, where `QList` and `QVector` were merged — in Qt5 only `QVector` has it, so `QList<TallyRow>(entries.size())` has no matching constructor.

The project still supports Qt5 (>= 5.15.2) via `cmake/FindQtRuntime.cmake` / `FORCE_USE_QT5`. CI only builds Qt6 (6.11.0), so this would slip through CI but break anyone building against Qt5.

A portable equivalent that also reads a bit cleaner:

```cpp
QList<TallyRow> rows;
rows.reserve(entries.size());
std::transform(entries.begin(), entries.end(), std::back_inserter(rows),
               [](const SubtypeEntry &e) { return TallyRow{e.name, QString::number(e.count)}; });
```

(If dropping Qt5 support is intentional, ignore — but that would be worth doing deliberately, not incidentally here.)

ZeldaZach (approved):
Clean, well-scoped refactor — nice generalization of the tally code. Traced the renamed members (no leftover references), confirmed the reordered early-return in `updateTotalSelectionCount` is behavior-preserving, and the defaulted `operator==` is fine under C++20. LGTM.

One non-blocking portability note left inline: `QList<TallyRow>(entries.size())` is a Qt6-only constructor and won't compile under Qt5 (still supported via `FORCE_USE_QT5`); CI only builds Qt6 so it would slip through. Easy portable swap to `reserve` + `std::back_inserter` if we still care about Qt5.

PR #7022, opened 2026-06-28
[UserContextMenu] Refactor: Consolidate actions

RickyRister:
## Related Ticket(s)
- Follow-up to #6981

We added methods to `UserContextMenu` so that we can call the effects of the actions from outside the class, but we never updated the class itself to use them. 

- Change the code in `UserContextMenu::showContextMenu` to call the new methods instead of duplicating the code.
- Add `execKick`
- Split `execAdjustMod` into the mod and judge versions.
  - Fix bug with `execAdjustMod` setting more fields than it should

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
@@ -707,13 +707,13 @@ void UserListWidget::connectPopupSignals()
     connect(m_userInfoPopup, &UserInfoPopup::warnHistoryRequested, userContextMenu, &UserContextMenu::execWarnHistory);
     connect(m_userInfoPopup, &UserInfoPopup::adminNotesRequested, userContextMenu, &UserContextMenu::execAdminNotes);
     connect(m_userInfoPopup, &UserInfoPopup::promoteToModRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, true, false); });
+            [this](const QString &n) { userContextMenu->execAdjustMod(n, true); });
     connect(m_userInfoPopup, &UserInfoPopup::demoteFromModRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, false, false); });
+            [this](const QString &n) { userContextMenu->execAdjustMod(n, false); });
     connect(m_userInfoPopup, &UserInfoPopup::promoteToJudgeRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, false, true); });
+            [this](const QString &n) { userContextMenu->execAdjustJudge(n, true); });
     connect(m_userInfoPopup, &UserInfoPopup::demoteFromJudgeRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, false, false); });
+            [this](const QString &n) { userContextMenu->execAdjustJudge(n, false); });
 }
 
 bool UserListWidget::eventFilter(QObject *obj, QEvent *event)
--- a/cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
@@ -436,103 +436,35 @@ void UserContextMenu::showContextMenu(const QPoint &pos,
     QAction *actionClicked = menu->exec(pos);
     if (actionClicked == nullptr) {
     } else if (actionClicked == aDetails) {
-        auto *infoWidget =
-            new UserInfoBox(client, false, static_cast<QWidget *>(parent()),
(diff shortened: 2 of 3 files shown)


cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -707,13 +707,13 @@ void UserListWidget::connectPopupSignals()
...
     connect(m_userInfoPopup, &UserInfoPopup::adminNotesRequested, userContextMenu, &UserContextMenu::execAdminNotes);
     connect(m_userInfoPopup, &UserInfoPopup::promoteToModRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, true, false); });
+            [this](const QString &n) { userContextMenu->execAdjustMod(n, true); });
     connect(m_userInfoPopup, &UserInfoPopup::demoteFromModRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, false, false); });
+            [this](const QString &n) { userContextMenu->execAdjustMod(n, false); });
     connect(m_userInfoPopup, &UserInfoPopup::promoteToJudgeRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, false, true); });
+            [this](const QString &n) { userContextMenu->execAdjustJudge(n, true); });
```
ZeldaZach, line 714: Worth calling out that this is a behavioral change, not just a rename — the popup path used to behave differently here.

Previously `execAdjustMod(n, false, true)` set **both** `should_be_mod` and `should_be_judge` on `Command_AdjustMod`. Since the server checks `has_should_be_mod()` / `has_should_be_judge()` independently ([serversocketinterface.cpp](https://github.com/Cockatrice/Cockatrice/blob/master/servatrice/src/serversocketinterface.cpp#L2070-L2091)), promoting a user to judge from the popup would also **remove their mod flag** (and demote-from-mod would strip judge). Splitting into `execAdjustJudge` sends only the relevant field, so the other flag is now left untouched — matching the context-menu path.

This looks like the correct behavior, just flagging it since the PR is described as pure consolidation. Might be worth a line in the description.

ZeldaZach (approved):
Clean refactor — consolidates the duplicated command-building in `showContextMenu` into the existing `exec*` methods. Traced every extracted method against the inline code it replaced; all equivalent, and all call sites are updated. LGTM.

One non-blocking note left inline: the `execAdjustMod`/`execAdjustJudge` split is a genuine (and correct) behavioral fix for the popup path, not just a rename — worth a line in the description.

cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -707,13 +707,13 @@ void UserListWidget::connectPopupSignals()
...
     connect(m_userInfoPopup, &UserInfoPopup::demoteFromJudgeRequested, this,
-            [this](const QString &n) { userContextMenu->execAdjustMod(n, false, false); });
+            [this](const QString &n) { userContextMenu->execAdjustJudge(n, false); });
```
ZeldaZach, line 716: This is a behavior change for the popup path, not a pure refactor — and I think it's the correct one, just flagging to confirm it's intended.

The old lambdas called `execAdjustMod(n, shouldBeMod, shouldBeJudge)`, which set **both** `should_be_mod` and `should_be_judge` on every `Command_AdjustMod`. The server (`cmdAdjustMod` in `serversocketinterface.cpp`) gates each flag on `has_should_be_mod()` / `has_should_be_judge()` independently, so setting both meant each popup action also touched the *other* role:

- `promoteToJudge` sent `(false, true)` → added judge **and removed the user's mod flag**
- `promoteToMod` sent `(true, false)` → added mod **and removed the user's judge flag**
- both demotes sent `(false, false)` → removed **both** flags at once

The new split helpers each set only their own field, so promoting/demoting one role no longer clobbers the other. This now matches the context-menu path, which only ever set a single flag — so this looks like a latent bug fix folded into the refactor rather than a regression. Might be worth a line in the PR description.

ZeldaZach (commented):
Nice consolidation — the context-menu path (`showContextMenu`) maps cleanly onto the existing `exec*` helpers, and I verified each helper body reproduces the inline code it replaces verbatim, so that path is behavior-preserving.

One heads-up on the popup wiring in `user_list_widget.cpp`: splitting `execAdjustMod` actually changes runtime behavior there (details inline). I believe the new behavior is the correct one, but since the PR is framed as a pure refactor it's worth calling out so it's a conscious choice. Everything else looks good.

ZeldaZach (approved):
Approving — the consolidation is solid and behavior-preserving on the context-menu path (verified each `exec*` helper reproduces the inline code it replaces). The popup-path behavior change noted in my earlier inline comment looks like a correct latent-bug fix; no need to block on it. LGTM 👍

PR #7023, opened 2026-06-28
CI: Fix login+push for publishing Servatrice Docker images

tooomm:
Recent releases were not pushing Servatrice images to GHCR.
It turns out the `github.event.release.target_commitish == 'master'` condition is not reliable (maybe changing the workflow trigger from "tag" to "released" full versions did influence this?).
If we ever create releases against a non master branch, that is probably on purpose and we do not need/want to check for that anyway.

The login step was skipped for that reason for recently published full releases, resulting in the following step throwing a permission error when trying to push the image:
```
Error: buildx failed with: ERROR: failed to build: failed to solve: failed to fetch anonymous token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Acockatrice%2Fservatrice%3Apull%2Cpush&service=ghcr.io: 403 Forbidden
```

`v3.0.1` and `v3.0.2` did not publish a Docker image because of that.

image

- Double check that we're running on a release, and only a non-prereleases in particular.
  We assume releases will always be created against master.
  This is also a bit more general and does not require a specific wording in the tag.
- ~~Use same condition for login and pushing~~
   --> Evaluate `push` option to `true`, if login is successful

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -57,7 +57,7 @@ jobs:
         uses: docker/setup-buildx-action@v4
 
       - name: "Login to GitHub Container Registry"
-        if: contains(github.event.release.tag_name, 'Release') && github.event.release.target_commitish == 'master'
+        if: ${{ github.event_name == 'release' && github.event.release.prerelease == false }}
         uses: docker/login-action@v4
         with:
           password: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):
Reviewed the workflow change — logic is sound. Traced `steps.login.outcome` across all triggers:

- **PR/push events**: `event_name != 'release'` short-circuits, so login is *skipped*, `outcome` is `'skipped'`, and `push: false`. Matches prior behavior (branch pushes never satisfied `ref_type == 'tag'` either).
- **Real release**: login runs and succeeds → `push: true`. If login fails, the job aborts before the build step (no `continue-on-error`/`if: always()`), so the guard never sees a stale value.

Nice fix — moving off the brittle `contains(tag_name, 'Release')` string match to `event_name == 'release'` is more robust.

One optional, non-blocking FYI: `prerelease == false` is technically redundant since the workflow only triggers on `release: types: [released]`, which doesn't fire for prereleases anyway. Harmless as a defensive double-check, so no need to change it.

LGTM 👍

PR #7025, opened 2026-06-29
Split trice_limits.h into dedicated headers

DawnFire42:
trice_limits.h had multiple purposes and it made sense to split it while I was there.

- trice_limits.h is now split into counter_limits.h , dice_limits.h and string_limits.h


libcockatrice_utility/libcockatrice/utility/string_limits.h
```diff
@@ -0,0 +1,29 @@
+#ifndef STRING_LIMITS_H
+#define STRING_LIMITS_H
+
+#include <QString>
+#include <algorithm>
+#include <string>
+
+// max size for short strings, like names and things that are generally a single phrase
```
RickyRister: might as well use this opportunity to turn them into proper docstrings

RickyRister (approved):
Sure

Issue #4196, opened 2020-11-29
Various warnings when compiling


Fenrirthviti:
Client Version: 2.7.6-beta.2 (2020-11-23)
Client Operating System: Windows 10 Version 2004
Build Architecture: 64-bit
Qt Version: 5.15.2
System Locale: en_US
Install Mode: Standard
_______________________________________________________________________________________

 There is a fair number of warnings when compiling the project on Windows 10, using VS2019 (I did not verify other OS's/toolchains) that look fairly simple to fix, but my lack of experience with C precludes my desire to fix them based on results from StackOverflow. Providing them as a list here for Zach to look at, per his request:

- [ ] Warning	C4146	unary minus operator applied to unsigned type, result still unsigned	cockatrice_common	F:\Source\Repos\Cockatrice\common\rng_sfmt.cpp	51	
- [ ] Warning	C26812	The enum type 'CardAttribute' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).	cockatrice_protocol	F:\Source\Repos\Cockatrice\Build\common\pb\card_attributes.pb.h	69	
- [ ] Warning	C6387	'value' could be '0':  this does not adhere to the specification for the function 'std::basic_string<char,std::char_traits<char>,std::allocator<char> >::{ctor}'. 	cockatrice_protocol	F:\Source\Repos\Cockatrice\Build\common\pb\command_set_card_attr.pb.h	329	
- [ ] Warning	C6387	'value' could be '0':  this does not adhere to the specification for the function 'std::basic_string<char,std::char_traits<char>,std::allocator<char> >::{ctor}'. 	cockatrice_protocol	F:\Source\Repos\Cockatrice\Build\common\pb\command_set_card_attr.pb.h	460	
- [ ] Warning	C26812	The enum type 'GameCommand_GameCommandType' is unscoped. Prefer 'enum class' over 'enum' (Enum.3).	cockatrice_protocol	F:\Source\Repos\Cockatrice\Build\common\pb\game_commands.pb.h	107	
- [ ] Warning	C6011	Dereferencing NULL pointer 'initial_value'. 	cockatrice_protocol	F:\Source\Repos\Cockatrice\Build\vcpkg_installed\x64-windows\include\google\protobuf\arenastring.h	373	
[...]

PR #7026, opened 2026-06-29
fix theme manager return type warning

DawnFire42:
## Related Ticket(s)
- related to Issue #4196 

Warning during build


RickyRister (approved):
Sure

PR #7030, opened 2026-07-01
[Networking] Doxygen

BruebachL:
Documemementation lacking.

- Documemementation added.


Doxyfile
```diff
@@ -349,7 +349,7 @@ OPTIMIZE_OUTPUT_SLICE  = NO
 #
 # Note see also the list of default file extension mappings.
 
-EXTENSION_MAPPING      =
+EXTENSION_MAPPING       = proto=C++
```
tooomm:
```suggestion
EXTENSION_MAPPING      = proto=C++
```

Extra whitespace (reason for the failing build)

Doxyfile
```diff
@@ -1102,6 +1103,7 @@ RECURSIVE              = YES
 # run.
 
 EXCLUDE                = build/ \
+                         cmake-build-debug/ \
                          cmake/ \
```
tooomm:
```suggestion
                         cmake/ \
                         cmake-build-debug/ \
```

Alphabetical ordering

libcockatrice_protocol/libcockatrice/protocol/pb/response.proto
```diff
@@ -45,33 +54,36 @@ message Response {
         RespServerFull = 36;           // Server user limit reached
         RespEmailBlackListed = 37;     // Server has blocked the email address provided for registration for some reason
```
tooomm, line 55:
```suggestion
        RespEmailRequiredToRegister = 25;  // Server requires email to register accounts but client did not provide one
        RespTooManyRequests = 26;          // Server refused to complete command because client has sent too many too quickly
        RespPasswordTooShort = 27;         // Server requires a decent password
        RespAccountNotActivated =
28; // Client attempted to log into a registered username but the account hasn't been activated
        RespRegistrationDisabled = 29;     // Server does not allow clients to register
        RespRegistrationFailed = 30;       // Server accepted a reg request but failed to perform the registration
        RespActivationAccepted = 31;       // Server accepted a reg user activation token
        RespActivationFailed = 32;         // Server didn't accept a reg user activation token
        RespRegistrationAcceptedNeedsActivation =
33; // Server accepted client registration, but it will need token activation
        RespClientIdRequired = 34;         // Server requires client to generate and send its client id before allowing access
        RespClientUpdateRequired = 35;     // Client is missing features that the server is requiring
        RespServerFull = 36;               // Server user limit reached
        RespEmailBlackListed = 37;         // Server has blocked the email address provided for registration for some reason
```

Align all comments

Doxyfile
```diff
@@ -1086,7 +1086,8 @@ FILE_PATTERNS          = *.cc \
                          *.h++ \
                          *.markdown \
                          *.md \
-                         *.dox
+                         *.dox \
+                         *.proto
```
tooomm, line 1090: Do we need/use all of the listed extensions and should keep them?

libcockatrice_protocol/libcockatrice/protocol/pb/response.proto
```diff
@@ -45,33 +54,36 @@ message Response {
...
+        WARN_HISTORY = 1013;            // Response with warn history
+        WARN_LIST = 1014;               // Response with current warnings
+        VIEW_LOG = 1015;                // Response with logs
+        FORGOT_PASSWORD_REQUEST = 1016; // Response to password reset request
+        PASSWORD_SALT = 1017;           // Response containing password salt
+        GET_ADMIN_NOTES = 1018;         // Response with admin notes
+        REPLAY_LIST = 1100;             // Response listing replays
+        REPLAY_DOWNLOAD = 1101;         // Response for replay download
+        REPLAY_GET_CODE = 1102;         // Response containing replay code
         CARD_ART_RULE_LIST = 1200;
```
tooomm: Missing comment?

tooomm (changes requested):
That's a big bunch of doc additions!

See small notes inline.

libcockatrice_protocol/libcockatrice/protocol/pb/response.proto
BruebachL: The linter will yell.

Doxyfile
BruebachL: Not as far as I'm aware but the performance hit is negligible afaik (or doesn't really matter. I don't often find myself waiting on the 'generate docs' step, as opposed to, for example, the Windows 10 CI.)

In any case, I'd like to diff that out from this PR if we do end up dropping the extensions.

libcockatrice_protocol/libcockatrice/protocol/pb/response.proto
BruebachL: Yeah, I actually drafted this PR ages ago and then this one was new.

libcockatrice_protocol/libcockatrice/protocol/pb/response.proto
tooomm: Yeah, I saw the failing CI run. It looks a bit off now though 😄 
Image

Well, maybe some are too long...

tooomm (approved):
Looks ok as far as I can judge

libcockatrice_protocol/libcockatrice/protocol/pb/game_commands.proto
```diff
@@ -1,56 +1,197 @@
...
+        /// Client: reflected via game state events
         KICK_FROM_GAME = 1000;
+
+        /// Leave the current game voluntarily.
+        /// Server: Server_Game::handleLeaveGame
+        /// Client: reflected via game state events
         LEAVE_GAME = 1001;
+
+        /// Send an in-game chat message.
+        /// Server: Server_Game::handleGameSay
```
ZeldaZach: `Server_Game::handleGameSay` doesn't exist — and this fabricated `Server_Game::handle<Command>` pattern is repeated for every command in this enum. The actual server-side handlers are the `cmd*` methods on `Server_Player` / `Server_AbstractParticipant`, e.g. this one is `Server_AbstractParticipant::cmdGameSay`, `SHUFFLE` is `Server_Player::cmdShuffle`, etc. The `handle*` names that *do* exist (`handleNextTurn`, `handleReverseTurn`, …) are client-side methods on `GameEventHandler`, not server dispatch.

Note this also contradicts the sibling doc `protocol_game_command.md` added in this same PR, which correctly cites `Server_AbstractParticipant::cmdGameSay`, `Server_Player::cmdShuffle`, etc. A developer following these breadcrumbs into the server code will find nothing. Please regenerate these `Server:` lines against the real handler names (or drop them and point at the `.md`).

doc/doxygen/filters/proto2cpp.py
```diff
@@ -0,0 +1,285 @@
...
+                theFile.write(string)
+
+    ## Writes @p string to error log file.
+    ##
+    ## #logLevel must be #logError or #logAll or otherwise the logging is skipped.
+    ##
+    ## @param string String to be written to error log file.
+    #
+    def logError(self, string):
+        if self.logLevel >= self.logError:
```
ZeldaZach, line 276: Bug: `self.logError` here is the *method* (defined on line 275), not the log-level constant. The constant is `self.logErrors` (= 1). Comparing `int >= <bound method>` raises `TypeError` in Python 3, so this line crashes whenever `logError()` is actually called — i.e. on any `IOError` while opening a `.proto` file (lines 99/107/121/133), which is exactly when you'd want the error logged. Should be `if self.logLevel >= self.logErrors:`.

doc/doxygen/filters/proto2cpp.py
```diff
@@ -0,0 +1,285 @@
...
+    def __init__(self):
+        ## Debug log file name.
+        self.logFile = "proto2cpp.log"
+        ## Error log file name.
+        self.errorLogFile = "proto2cpp.error.log"
+        ## Logging level.
+        self.logLevel = self.logNone
+
+    ## Handles a file.
+    ##
```
ZeldaZach, line 82: Minor/vendored: the parameter is `fileName` but the body uses `filename` throughout, which resolves to the module-level global set by the `for filename in sys.argv[1:]` loop at the bottom. It happens to work because the loop passes that same global, but the parameter is dead and this is fragile. Since this is a lightly-modified third-party script it's fine to leave, but worth a note.

ZeldaZach (commented):
Doc review — the new Doxygen/Markdown content is thorough and reads well. Two accuracy issues and one latent bug in the vendored filter script flagged inline. The biggest one: the `Server: Server_Game::handle*` references throughout `game_commands.proto` don't correspond to any real methods, and they contradict the sibling `protocol_game_command.md` (which cites the correct `cmd*` handlers).

ZeldaZach (changes requested):
The documentation is thorough and mostly accurate — just one blocking issue.

**Blocking:** the `Server: Server_Game::handle<Command>` references throughout `game_commands.proto` name methods that don't exist. The real server handlers are the `cmd*` methods on `Server_Player` / `Server_AbstractParticipant` (e.g. `Server_AbstractParticipant::cmdGameSay`, `Server_Player::cmdShuffle`), and these lines contradict the correct names in the `protocol_game_command.md` added in this same PR. A developer following them into the server code finds nothing. Details in the inline comment on `game_commands.proto`.

**Non-blocking (nice to fix while you're here):** the `self.logError` → `self.logErrors` bug in `proto2cpp.py` (crashes the filter's own error path), plus the dead `fileName` parameter — both flagged inline and both in vendored code, so your call.

Happy to approve once the `Server:` handler references are corrected.

tooomm:
@BruebachL did you had a chance to look at the last comments from Zach?

I'm really looking forward to get this in.

<br>

I moved your new filter program into the doxygen folder and changed the link in the config.

BruebachL:
> @BruebachL did you had a chance to look at the last comments from Zach?
> 
> I'm really looking forward to get this in.
> 
> I moved your new filter program into the doxygen folder and changed the link in the config.

I'll have another go at this today.

Doxyfile
```diff
@@ -1184,7 +1184,7 @@ IMAGE_PATH             = doc/doxygen/images
 # need to set EXTENSION_MAPPING for the extension otherwise the files are not
 # properly processed by Doxygen.
 
-INPUT_FILTER           = "python proto2cpp.py"
+INPUT_FILTER           = "python doc/doxygen/filters/proto2cpp.py"
```
tooomm: Despite the file mentioning to use `INPUT_FILTER` and running the script for all files, there is also a newer `FILTER_PATTERNS` option right below, which would only run the script for .proto files, not all files:
```
FILTER_PATTERNS = *.proto="python doc/doxygen/filters/proto2cpp.py"
```

That should be quicker. Did you try that?

tooomm:
Indeed there is a speed-up with invoking the python filter only for .proto files, cool. 👍 

Where is the vendored code sourced from?
Maybe good to have a reference somewhere (at least in the PR description at the top?)

ZeldaZach (approved):

BruebachL:
https://github.com/BlackRiverSystems/proto2cpp

PR #7031, opened 2026-07-02
[Settings] Split Paths from SettingsCache into PathSettings

BruebachL:
SettingsCache is a huge monolith.

- Split off the paths settings into its own manager file (while still living in global.ini for now)

Caveat: I still need to fix up some of the original load/reset implementation stuff and we also lost some signals in the process, marked with a todo and an explanation.

### DIFF
--- a/cockatrice/src/client/settings/cache_settings.cpp
+++ b/cockatrice/src/client/settings/cache_settings.cpp
@@ -150,31 +150,6 @@ void SettingsCache::translateLegacySettings()
     }
 }
 
-QString SettingsCache::getSafeConfigPath(QString configEntry, QString defaultPath) const
-{
-    QString tmp = settings->value(configEntry).toString();
-    // if the config settings is empty or refers to a not-existing folder,
-    // ensure that the defaut path exists and return it
-    if (tmp.isEmpty() || !QDir(tmp).exists()) {
-        if (!QDir().mkpath(defaultPath)) {
-            qCInfo(SettingsCacheLog) << "[SettingsCache] Could not create folder:" << defaultPath;
-        }
-        tmp = defaultPath;
-    }
-    return tmp;
-}
-
-QString SettingsCache::getSafeConfigFilePath(QString configEntry, QString defaultPath) const
-{
-    QString tmp = settings->value(configEntry).toString();
-    // if the config settings is empty or refers to a not-existing file,
-    // return the default Path
-    if (!QFile::exists(tmp) || tmp.isEmpty()) {
-        tmp = std::move(defaultPath);
-    }
-    return tmp;
-}
-
 SettingsCache::SettingsCache()
 {
     // first, figure out if we are running in portable mode
@@ -188,6 +163,8 @@ SettingsCache::SettingsCache()
 
     QString settingsPath = getSettingsPath();
     settings = new QSettings(settingsPath + "global.ini", QSettings::IniFormat, this);
+    pathsSettings = new PathSettings(settingsPath, this);
+    pathsSettings->load(getDataPath(), getCachePath());
     shortcutsSettings = new ShortcutsSettings(settingsPath, this);
     cardDatabaseSettings = new CardDatabaseSettings(settingsPath, this);
     serversSettings = new ServersSettings(settingsPath, this);
@@ -246,8 +223,6 @@ SettingsCache::SettingsCache()
     for (const auto &tipNumber : settings->value("tipOfDay/seenTips").toList()) {
         seenTips.append(tipNumber.toInt());
     }
-
-    loadPaths();
 
     themeName = settings->value("theme/name").toString();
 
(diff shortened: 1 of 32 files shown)


RickyRister (approved):

libcockatrice_settings/libcockatrice/settings/path_settings.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    auto settings = getSettings();
+    settings.beginGroup("paths");
+    settings.remove(""); // removes all keys in Path/*
+    settings.endGroup();
+
+    load(dataPath, cachePath);
+}
+
+/*void SettingsCache::loadPaths()
```
RickyRister, line 72: why are you leaving this around?

libcockatrice_settings/libcockatrice/settings/path_settings.h
```diff
@@ -0,0 +1,111 @@
...
+    [[nodiscard]] QString getSpoilerCardDatabasePath() const override
+    {
+        return spoilerDatabasePath;
+    }
+    [[nodiscard]] QString getTokenDatabasePath() const override
+    {
+        return tokenDatabasePath;
+    }
+
+    void setDeckPath(const QString &_deckPath);
```
RickyRister, line 80: can we rename all these args to `value` or `path` so that it's easier to copy-paste

libcockatrice_card/libcockatrice/card/database/card_database_loader.cpp
```diff
@@ -27,9 +27,11 @@ CardDatabaseLoader::CardDatabaseLoader(QObject *parent,
         connect(p, &ICardDatabaseParser::addFormat, database, &CardDatabase::addFormat, Qt::DirectConnection);
     }
 
+    // TODO: PathSettings can't double inherit from a QObject ICardDatabasePathProvider and a QObject SettingsManager.
+    // ICardDatabasePathProvider was only a QObject for this single signal.
     // when SettingsCache's path changes, trigger reloads
-    connect(pathProvider, &ICardDatabasePathProvider::cardDatabasePathChanged, this,
-            &CardDatabaseLoader::loadCardDatabases);
+    /*connect(pathProvider, &ICardDatabasePathProvider::cardDatabasePathChanged, this,
```
ZeldaZach, line 33: Disabling this connection means changing the card/token/spoiler/customset database path in Settings no longer reloads the databases until the next restart.

The TODO is right that a class can't have two QObject bases, but the signal doesn't need to live on the interface. `PathSettings` is already a QObject (via `SettingsManager`), so it can keep `cardDatabasePathChanged` as its own signal and you can connect to the concrete `PathSettings` type here instead of `ICardDatabasePathProvider`. That restores the behavior without the double-inheritance problem. Can we fix it in this PR, or at least open a tracking issue before merge?

libcockatrice_settings/libcockatrice/settings/path_settings.h
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Backed by the "paths" group of the existing global.ini, so extracting this class out
+ * of SettingsCache is a pure code-organization change: no stored user data moves.
+ */
+class PathSettings : public SettingsManager, public ICardDatabasePathProvider
+{
+    Q_OBJECT
+
+signals:
+    void cardDatabasePathChanged();
```
ZeldaZach, line 30: `cardDatabasePathChanged` is declared but never emitted in live code. The only `emit` is inside the commented-out `resetPaths` block in the .cpp. The setters (`setCardDatabasePath` / `setSpoilerDatabasePath` / `setTokenDatabasePath` / `setCustomCardDatabasePath`) used to emit it. As written the signal is dead and the CardDatabaseLoader reload never fires (see the disabled connect in `card_database_loader.cpp`). The setters should emit this.

libcockatrice_settings/libcockatrice/settings/path_settings.cpp
```diff
@@ -0,0 +1,163 @@
...
+    setValue(replaysPath, "replays");
+}
+
+void PathSettings::setThemesPath(const QString &_themesPath)
+{
+    themesPath = _themesPath;
+    setValue(themesPath, "themes");
+}
+
+void PathSettings::setPicsPath(const QString &_picsPath)
```
ZeldaZach, line 133: The old `SettingsCache::setPicsPath` emitted `picsPathChanged()`, and both `CardPictureLoader` (card_picture_loader.cpp:27) and `CardPictureLoaderLocal` (card_picture_loader_local.cpp:18) still connect to `SettingsCache::picsPathChanged`. That signal is now never emitted, so those connections are dead and changing the pics path won't refresh loaded images until restart.

Unlike the card-database case, this loss isn't flagged with a TODO, and `PathSettings` doesn't declare a `picsPathChanged` signal at all. Worth adding one and emitting it here (and pointing the loaders at `paths()`).

libcockatrice_settings/libcockatrice/settings/path_settings.cpp
```diff
@@ -0,0 +1,163 @@
...
+    redirectCachePath = getSafeConfigPath("redirects", cachePath + "/redirects/");
+    recomputeCustomPicsPath();
+    customCardDatabasePath = getSafeConfigPath("customsets", dataPath + "/customsets/");
+
+    cardDatabasePath = getSafeConfigFilePath("carddatabase", dataPath + "/cards.xml");
+    tokenDatabasePath = getSafeConfigFilePath("tokendatabase", dataPath + "/tokens.xml");
+    spoilerDatabasePath = getSafeConfigFilePath("spoilerdatabase", dataPath + "/spoiler.xml");
+}
+
+void PathSettings::reset(const QString &dataPath, const QString &cachePath)
```
ZeldaZach, line 62: Two things here:

1. `reset()` emits nothing, but the old `resetPaths()` emitted `cardDatabasePathChanged` / `picsPathChanged` when values actually changed. So "Reset all paths" in the settings dialog no longer live-refreshes the databases or pictures.
2. The group name `"paths"` is hardcoded even though it's already the default group given to the `SettingsManager` constructor. Reusing the base helper (`deleteValue("")`) instead of re-opening the group by name keeps the two from drifting apart.

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -188,6 +163,8 @@ SettingsCache::SettingsCache()
 
     QString settingsPath = getSettingsPath();
     settings = new QSettings(settingsPath + "global.ini", QSettings::IniFormat, this);
+    pathsSettings = new PathSettings(settingsPath, this);
+    pathsSettings->load(getDataPath(), getCachePath());
```
ZeldaZach, line 167: `pathsSettings->load()` now runs before `translateLegacySettings()` (a few lines down). On a legacy migration, when global.ini doesn't exist yet, `translateLegacySettings()` copies the old `paths/*` keys into global.ini via its catch-all loop, but that happens after `load()` has already cached the defaults. So a user upgrading from a pre-global.ini version with custom paths gets them reset to defaults for that first run (they're persisted, so a restart recovers them).

The old code called `loadPaths()` after translation. Moving this `load()` call to after the `translateLegacySettings()` block preserves the original behavior.

cockatrice/src/client/settings/cache_settings.h
```diff
@@ -1183,4 +1132,4 @@ public slots:
     void setShowTotalSelectionCount(QT_STATE_CHANGED_T _showTotalSelectionCount);
     void setShowSubtypeSelectionTally(QT_STATE_CHANGED_T _showSubtypeSelectionTally);
 };
-#endif
+#endif
```
ZeldaZach, line 1135: Trailing newline at EOF got dropped here (lint).

ZeldaZach (commented):
Review of the SettingsCache -> PathSettings extraction. The mechanical `.paths().` migration across the ~30 call sites looks complete and consistent, and moving path ownership into its own manager reads cleanly. Main concern is the lost signal wiring: beyond the card-database reload the PR body calls out, `picsPathChanged` is also silently dropped, and the reset/settings-dialog live-refresh paths no longer fire. Details inline.

ZeldaZach (changes requested):
Requesting changes on the lost signal wiring — see my inline comments. The extraction itself looks good; the blocker is that changing the card-database path and the pics path no longer trigger reloads/refreshes (the picsPathChanged loss in particular is undocumented), and the reset/settings-dialog live-refresh paths no longer fire.

PR #7035, opened 2026-07-05
[CI] Use modern `macos-26` across Mac builds

tooomm:
Use recent runner with modern tooling as basis for targeting the required OS version floor.
- Use latest `macos-26` runner for compiling macOS 14 ARM and 15 ARM targets
  (`macos-26-intel` turns out to be very slow, keeping `macos-15-intel` for macOS 13 x86 target)
- Move debug build to native macOS 26 on Apple Silicon
  > - [x] Pending fix: #7134
- Rework cache key strategy for the compiler caches to separate them better, have less churn, more natural rotation and avoid race conditions (with same Mac runner, the key would be the same and parallel matrix runs start fighting and results become inconsistent)
- Update Qt to 6.11.1 on all builds
- Update Xcode to most recent available versions


tooomm:
Can you please test the macOS binaries from this PR again? @RickyRister

BruebachL (approved):
Everything is gated behind Win10 anyway so the slowdown doesn't hurt as bad and this is worth having for the better cache keying alone imo.

BruebachL:
Defer to @RickyRister for binary testing.

RickyRister:
All three macOS binaries work for me

Test on a `macOS Sequoia (15.7.1)`

PR #7036, opened 2026-07-06
[Game] Use in-game menu for tally setting

RickyRister:
## Related Ticket(s)
- Follow-up to #7021

Use an in-game menu for selecting the tally type so that it's easier to change the tally type in-game and to allow for more tally types.

- Create `TallyMenu`
  - Is another top-level managed `AbstractPlayerComponent` menu
  - Always appears regardless of whether game is replay or spectator
  - Is a tearoff menu
- Remove subtype tally setting from settings menu
- Change settings to store TallyType instead of a boolean.
  - Add signal to SettingsCache for TallyType change
  - Also connect tally update to the signal

https://github.com/user-attachments/assets/b8f695c9-aba3-4ec0-a11b-b606d186a008

## Screenshots

Screenshot 2026-07-05 at 8 26 36 PM

Screenshot 2026-07-05 at 8 27 07 PM

### DIFF
--- a/cockatrice/src/client/settings/cache_settings.cpp
+++ b/cockatrice/src/client/settings/cache_settings.cpp
@@ -313,7 +313,7 @@ SettingsCache::SettingsCache()
 
     showDragSelectionCount = settings->value("interface/showlassoselectioncount", true).toBool();
     showTotalSelectionCount = settings->value("interface/showpersistentselectioncount", true).toBool();
-    showSubtypeSelectionTally = settings->value("interface/showsubtypeselectiontally", true).toBool();
+    tallyType = settings->value("interface/tallyType", 0).toInt();
 
     showShortcuts = settings->value("menu/showshortcuts", true).toBool();
     showGameSelectorFilterToolbar = settings->value("menu/showgameselectorfiltertoolbar", true).toBool();
@@ -1396,10 +1396,11 @@ void SettingsCache::setShowTotalSelectionCount(QT_STATE_CHANGED_T _showTotalSele
     settings->setValue("interface/showpersistentselectioncount", showTotalSelectionCount);
 }
 
-void SettingsCache::setShowSubtypeSelectionTally(QT_STATE_CHANGED_T _showSubtypeSelectionTally)
+void SettingsCache::setTallyType(TallyType value)
 {
-    showSubtypeSelectionTally = static_cast<bool>(_showSubtypeSelectionTally);
-    settings->setValue("interface/showsubtypeselectiontally", showSubtypeSelectionTally);
+    tallyType = static_cast<int>(value);
+    settings->setValue("interface/tallyType", tallyType);
+    emit tallyTypeChanged(value);
 }
 
 void SettingsCache::loadPaths()
--- a/cockatrice/src/client/settings/cache_settings.cpp
+++ b/cockatrice/src/client/settings/cache_settings.cpp
@@ -1396,10 +1396,14 @@ void SettingsCache::setShowTotalSelectionCount(QT_STATE_CHANGED_T _showTotalSele
     settings->setValue("interface/showpersistentselectioncount", showTotalSelectionCount);
 }
 
-void SettingsCache::setTallyType(TallyType value)
+void SettingsCache::setTallyType(int value)
 {
-    tallyType = static_cast<int>(value);
-    settings->setValue("interface/tallyType", tallyType);
+    if (tallyType == value) {
+        return;
+    }
+
+    tallyType = value;
+    settings->setValue("interface/tallyType", value);
     emit tallyTypeChanged(value);
 }
 
--- a/cockatrice/src/client/settings/cache_settings.h
+++ b/cockatrice/src/client/settings/cache_settings.h
@@ -7,7 +7,6 @@
 #ifndef SETTINGSCACHE_H
 #define SETTINGSCACHE_H
 
-#include "../../game_graphics/tally/tally.h"
 #include "../../interface/card_picture_loader/card_picture_loader_cache_method.h"
 #include "../../interface/card_picture_loader/card_picture_loader_local_schemes.h"
 #include "shortcuts_settings.h"
@@ -198,7 +197,7 @@ signals:
     void useTearOffMenusChanged(bool state);
     void roundCardCornersChanged(bool roundCardCorners);
     void keepGameChatFocusChanged(bool value);
-    void tallyTypeChanged(TallyType type);
+    void tallyTypeChanged(int type);
 
 private:
     QSettings *settings;
@@ -481,9 +480,9 @@ public:
     {
         return showTotalSelectionCount;
     }
-    [[nodiscard]] TallyType getTallyType() const
+    [[nodiscard]] int getTallyType() const
     {
-        return static_cast<TallyType>(tallyType);
+        return tallyType;
     }
     [[nodiscard]] bool getNotificationsEnabled() const
     {
@@ -1183,6 +1182,6 @@ public slots:
     void setRoundCardCorners(bool _roundCardCorners);
     void setShowDragSelectionCount(QT_STATE_CHANGED_T _showDragSelectionCount);
     void setShowTotalSelectionCount(QT_STATE_CHANGED_T _showTotalSelectionCount);
-    void setTallyType(TallyType value);
+    void setTallyType(int value);
 };
 #endif
--- a/cockatrice/src/client/settings/cache_settings.h
+++ b/cockatrice/src/client/settings/cache_settings.h
@@ -7,6 +7,7 @@
 #ifndef SETTINGSCACHE_H
 #define SETTINGSCACHE_H
 
+#include "../../game_graphics/tally/tally.h"
 #include "../../interface/card_picture_loader/card_picture_loader_cache_method.h"
 #include "../../interface/card_picture_loader/card_picture_loader_local_schemes.h"
 #include "shortcuts_settings.h"
@@ -197,6 +198,7 @@ signals:
     void useTearOffMenusChanged(bool state);
     void roundCardCornersChanged(bool roundCardCorners);
     void keepGameChatFocusChanged(bool value);
+    void tallyTypeChanged(TallyType type);
 
 private:
     QSettings *settings;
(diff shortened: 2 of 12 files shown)


cockatrice/src/game_graphics/player/menu/tally_menu.cpp
```diff
@@ -0,0 +1,49 @@
+#include "tally_menu.h"
+
+#include "../../client/settings/cache_settings.h"
```
ZeldaZach: Wrong relative depth: from `game_graphics/player/menu/` this resolves to the non-existent `game_graphics/client/...`. Every sibling (`say_menu.cpp`, `hand_menu.cpp`, `library_menu.cpp`) uses `"../../../client/settings/cache_settings.h"` (three `..`). It currently compiles only by accident, because an unrelated `-I .../src/client/network` include dir happens to make `../../client/...` resolve to `src/client/settings/...`. Please change to `"../../../client/settings/cache_settings.h"`.

cockatrice/src/client/settings/cache_settings.h
```diff
@@ -479,9 +481,9 @@ class SettingsCache : public ICardDatabasePathProvider, public INetworkSettingsP
     {
         return showTotalSelectionCount;
     }
-    [[nodiscard]] bool getShowSubtypeSelectionTally() const
+    [[nodiscard]] TallyType getTallyType() const
     {
-        return showSubtypeSelectionTally;
+        return static_cast<TallyType>(tallyType);
```
ZeldaZach: Unchecked cast of an arbitrary stored int to the scoped enum. A corrupt/out-of-range value won't crash (`Tally::compute()` has a `return {}` fallback), but the TallyMenu would then show no item checked. Consider validating the range on read (or clamping in the constructor) so an invalid value falls back to `TallyType::None`.

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -313,7 +313,7 @@ SettingsCache::SettingsCache()
 
     showDragSelectionCount = settings->value("interface/showlassoselectioncount", true).toBool();
     showTotalSelectionCount = settings->value("interface/showpersistentselectioncount", true).toBool();
-    showSubtypeSelectionTally = settings->value("interface/showsubtypeselectiontally", true).toBool();
+    tallyType = settings->value("interface/tallyType", 0).toInt();
```
ZeldaZach, line 316: Two behavior changes worth confirming: (1) the default is now `0` = `TallyType::None`, whereas #7021 defaulted the subtype tally ON (`true`), so the tally now defaults off; (2) the key was renamed from `interface/showsubtypeselectiontally`, so any previously stored value is silently dropped. Both are fine for released users (#7021 shipped after the last tag, so no release contains the old key), but master testers will see their tally disappear. Confirm the off-by-default is intended.

cockatrice/src/game_graphics/player/menu/tally_menu.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    retranslateUi();
+}
+
+QAction *TallyMenu::createTallyAction(TallyType tallyType)
+{
+    TallyType currentType = SettingsCache::instance().getTallyType();
+
+    QAction *action = new QAction(this);
+    action->setCheckable(true);
```
ZeldaZach, line 27: The two actions are checkable but not placed in an exclusive `QActionGroup`; mutual exclusivity relies on the `tallyTypeChanged` signal round-trip. It works (re-clicking the active item unchecks then re-checks via the emit), but a `QActionGroup(this)` with `setExclusive(true)` would be more idiomatic, avoid the transient flicker, and scale cleanly as more tally types are added.

cockatrice/src/game_graphics/player/menu/tally_menu.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "../../client/settings/cache_settings.h"
+
+TallyMenu::TallyMenu()
+{
+    aTallyNone = createTallyAction(TallyType::None);
+    aTallySubtypes = createTallyAction(TallyType::Subtypes);
+
+    addAction(aTallyNone);
+    addSeparator();
```
ZeldaZach, line 16: A separator between the two radio-style options reads oddly. As more tally types are added it will sit between `None` and the first real type. Consider dropping it, or grouping `None` apart deliberately.

cockatrice/src/game_graphics/player/menu/tally_menu.h
```diff
@@ -0,0 +1,28 @@
...
+#include "abstract_player_component.h"
+
+#include <QMenu>
+
+class TallyMenu : public TearOffMenu, public AbstractPlayerComponent
+{
+    Q_OBJECT
+
+public:
+    explicit TallyMenu();
```
ZeldaZach: `explicit` has no effect on a zero-argument constructor. Drop it: `TallyMenu();`.

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -1396,10 +1396,11 @@ void SettingsCache::setShowTotalSelectionCount(QT_STATE_CHANGED_T _showTotalSele
...
 }
 
-void SettingsCache::setShowSubtypeSelectionTally(QT_STATE_CHANGED_T _showSubtypeSelectionTally)
+void SettingsCache::setTallyType(TallyType value)
 {
-    showSubtypeSelectionTally = static_cast<bool>(_showSubtypeSelectionTally);
-    settings->setValue("interface/showsubtypeselectiontally", showSubtypeSelectionTally);
+    tallyType = static_cast<int>(value);
+    settings->setValue("interface/tallyType", tallyType);
+    emit tallyTypeChanged(value);
```
ZeldaZach, line 1407: Emitted unconditionally, even when `value` equals the current type. Every menu click on the already-selected item triggers a full tally recompute and re-checks all actions. Minor; guarding with an equality check would avoid the redundant work.

cockatrice/src/client/settings/cache_settings.h
```diff
@@ -7,6 +7,7 @@
 #ifndef SETTINGSCACHE_H
 #define SETTINGSCACHE_H
 
+#include "../../game_graphics/tally/tally.h"
```
ZeldaZach: This adds a `client/settings` to `game_graphics` include dependency. `tally.h` is lightweight (just the enum plus a struct), so it's acceptable, but note the layering direction in case settings is meant to stay independent of game graphics.

ZeldaZach (commented):
Nice, focused follow-up that moves the tally choice into an in-game `TallyMenu` and swaps the bool setting for a `TallyType` enum. Signal wiring is correct (direct connections, context objects for auto-disconnect, no duplicate/dangling connects, no `qRegisterMetaType` needed), the settings-page removal is clean with no orphaned references, and `TallyMenu`'s lifetime follows the existing `SayMenu`/`CustomZoneMenu` pattern (any ownership concern there is pre-existing, not introduced here).

Status: **approve with comments**. One real issue: `tally_menu.cpp`'s include of `cache_settings.h` uses two `..` instead of three and only compiles by accident via an unrelated `-I` dir; please fix it to match its siblings. Also worth confirming before merge: the default tally is now `None` (off), a change from #7021's default-on, and the setting key was renamed so old values are dropped (harmless since #7021 never shipped in a release). Remaining items are minor: prefer an exclusive `QActionGroup`, add int-to-enum range validation, and small style nits.

cockatrice/src/client/settings/cache_settings.cpp
RickyRister: Yes, the new default is intended

cockatrice/src/client/settings/cache_settings.h
RickyRister: cache_settings already includes `CardPictureLoaderCacheMethod`, so it seems like we don't mind including custom classes currently.

However, it seems like we're refactoring to always pass enums as into settings and do the conversion in the using class. So I'll change this to also use ints

cockatrice/src/game_graphics/player/menu/tally_menu.cpp
RickyRister: Yes, I wanted there to be a separation between the None and the rest of the options.

Here's how it looks with one more options added.

Screenshot 2026-07-23 at 12 06 12 AM

If you don't like how it looks, what are you suggesting for "grouping `None` apart deliberately"?

cockatrice/src/game_graphics/tally/tally.cpp
```diff
@@ -2,6 +2,16 @@
 
 #include "subtype_tally.h"
 
+TallyType Tally::intToType(int value)
+{
+    // remember to change this whenever you add a new member
+    if (value < static_cast<int>(TallyType::None) || value >= static_cast<int>(TallyType::Subtypes)) {
```
ZeldaZach: `intToType` returns `None` for `Subtypes`, which makes the menu a no-op for the one type that does anything.

With `None = 0` and `Subtypes = 1`, this guard is `value < 0 || value >= 1`, so `value == 1` (Subtypes) trips `value >= 1` and returns `None`. `Subtypes` is a real, used value here (`compute` has a `case TallyType::Subtypes`), not a count sentinel, so using it as an exclusive upper bound excludes it.

End to end: clicking Subtypes calls `setTallyType(1)`, then `GameView::updateTotalSelectionCount` runs `intToType(1)` -> `None` -> no breakdown is shown. The initial menu checkmark is wrong for the same reason, since `createTallyAction` routes the stored value through `intToType` too. The demo video predates the int refactor, so it isn't visible there.

Change `>=` to `>`:

```cpp
if (value < static_cast<int>(TallyType::None) || value > static_cast<int>(TallyType::Subtypes)) {
```

That unblocks it. Given the `// remember to change this whenever you add a new member` comment, a trailing count sentinel (e.g. a `Count` member used as the exclusive bound) would be sturdier long-term, but the `>` fix is enough.

ZeldaZach (changes requested):
Thanks for the quick turnaround. The include depth, the exclusive `QActionGroup`, the emit guard in `setTallyType`, the dropped `explicit`, and the move to int-based storage are all handled well.

One regression came in with the int refactor: `intToType` excludes `Subtypes`, so selecting it in the menu is currently a no-op. Details inline. It's a one-character fix, and I'm happy to re-approve once it's in. A quick manual check that "Subtypes" actually shows the breakdown would be ideal, since the demo video predates this change.

cockatrice/src/game_graphics/tally/tally.cpp
RickyRister: fixed

ZeldaZach (approved):

Issue #1171, opened 2015-06-28
Suggest faster replay speed


JineteDV:
There probably aren't many people who watch a lot of replay but I watch a decent amount :D 

I remember reading before that skipping around in the timeline on a replay isn't possible, but it would be nice to have in addition to the fast forward (2x speed?), a half speed and a double of whatever the current fast forward speed is. Also would it be possible to have a button to restart the replay from the beginning instead of what we have to do now and close it out and reopen it again.

ZeldaZach:
While I agree with all that you're saying, I'm not certain if it's a viable option at this time. Maybe another developer can look into this soon.

anoda9:
> a half speed and a double of whatever the current fast forward speed is.

This doesn't seem like the best solution to the problem (lack of fine control over replay playback). Instead of adding more levels of speed, the program should just support being able to click to any point on the timeline of the replay and jump to that spot. 

The client doesn't even have to support clicking to exact moments in the replay (if that would take too much work) - instead, there could be set anchors every 30 seconds, and clicking on the timeline jumps you to the anchor nearest to the spot you clicked. 

Related: #551

poixen:
This has been covered before. Replays are linear and can only go foward in event order.

Daenyth:
How hard would it be to change it so that you can play the events forward
to modify the game state model but just don't render it until you hit the
point seeked to? It's obviously less efficient than true rewinding, but...

woogerboy21:
@poixen can correct me if I am wrong, but the problem right now with the replays is that it is just a blob data with not type of time management or concept of key frames being recorded.  Meaning that you can only start from one spot and must continue to the end.  To be able to support skipping around some how the data would have to be manipulated into placing time stamps or some type of key frame concept into the data to allow seeking.

Daenyth:
I know how it's stored but that doesn't make it impossible to implement things, just difficult.

Suppose we treat the game state MVC style (sort of already do I think). We can do a scan across the list of game actions in the replay and compute the net game state at various points. (total action size / intervals). Then in order to seek we map the position of the seek bar and scale it as an index of the list of events. We find the nearest precomputed game state that is before the seek point, clear the view and re-render based on that state, and proceed forward from there.

I *think* we already have most of the mechanics needed to perform this, it's just a question of rigging the pieces together in the right way. Perhaps we must refactor the game state.

anoda9:
[...]

PR #7037, opened 2026-07-06
[Replay] Make fast forward speed configurable

RickyRister:

I would like the replay fast forward speed to be configurable. 10x is too fast sometimes.

- Add new setting for fast forward speed
  - Uses units of x times faster
- Refactor `ReplayTimelineWidget` to make the fastForwardScaleFactor configurable.
  - The replay speed is controlled by changing the timer interval, which is ultimately an integer. The fast forward speed setting allows decimals, so we ultimately round the result to an int.
- Create a `SettingsButtonWidget` for replays and put it in the replay dock

https://github.com/user-attachments/assets/83519e2f-b64b-48e7-a1da-d3ebbe5fdada

## Screenshots

Screenshot 2026-07-05 at 10 33 00 PM

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -228,8 +228,8 @@ set(cockatrice_SOURCES
     src/interface/widgets/printing_selector/set_name_and_collectors_number_display_widget.cpp
     src/interface/widgets/quick_settings/settings_button_widget.cpp
     src/interface/widgets/quick_settings/settings_popup_widget.cpp
-    src/interface/widgets/replay/replay_quick_settings_widget.cpp
     src/interface/widgets/replay/replay_manager.cpp
+    src/interface/widgets/replay/replay_quick_settings_widget.cpp
     src/interface/widgets/replay/replay_timeline_widget.cpp
     src/interface/widgets/server/chat_view/chat_view.cpp
     src/interface/widgets/server/game_filter_configs.cpp
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -228,6 +228,7 @@ set(cockatrice_SOURCES
     src/interface/widgets/printing_selector/set_name_and_collectors_number_display_widget.cpp
     src/interface/widgets/quick_settings/settings_button_widget.cpp
     src/interface/widgets/quick_settings/settings_popup_widget.cpp
+    src/interface/widgets/replay/replay_quick_settings_widget.cpp
     src/interface/widgets/replay/replay_manager.cpp
     src/interface/widgets/replay/replay_timeline_widget.cpp
     src/interface/widgets/server/chat_view/chat_view.cpp
--- a/cockatrice/src/interface/widgets/replay/replay_manager.cpp
+++ b/cockatrice/src/interface/widgets/replay/replay_manager.cpp
@@ -1,6 +1,7 @@
 #include "replay_manager.h"
 
 #include "../interface/widgets/tabs/tab_game.h"
+#include "replay_quick_settings_widget.h"
 
 #include <QHBoxLayout>
 #include <QToolButton>
@@ -78,13 +79,19 @@ ReplayManager::ReplayManager(TabGame *parent, GameReplay *_replay)
     replayFastForwardButton->setIconSize(QSize(32, 32));
     replayFastForwardButton->setIcon(QPixmap("theme:replay/fastforward"));
     replayFastForwardButton->setCheckable(true);
-    connect(replayFastForwardButton, &QToolButton::toggled, this, &ReplayManager::replayFastForwardButtonToggled);
+    connect(replayFastForwardButton, &QToolButton::toggled, this, &ReplayManager::updateTimeScaleFactor);
+
+    settingsWidget = new ReplayQuickSettingsWidget(this);
+    settingsWidget->setFixedSize(QSize(32, 32));
+    connect(settingsWidget, &ReplayQuickSettingsWidget::fastForwardSpeedChanged, this,
+            [this] { updateTimeScaleFactor(replayFastForwardButton->isChecked()); });
 
     // putting everything together
     auto replayControlLayout = new QHBoxLayout;
     replayControlLayout->addWidget(timelineWidget, 10);
     replayControlLayout->addWidget(replayPlayButton);
     replayControlLayout->addWidget(replayFastForwardButton);
+    replayControlLayout->addWidget(settingsWidget);
 
     setObjectName("replayControlWidget");
     setLayout(replayControlLayout);
@@ -115,9 +122,10 @@ void ReplayManager::replayPlayButtonToggled(bool checked)
     }
 }
 
-void ReplayManager::replayFastForwardButtonToggled(bool checked)
+void ReplayManager::updateTimeScaleFactor(bool isFastForward)
 {
-    timelineWidget->setTimeScaleFactor(checked ? ReplayTimelineWidget::FAST_FORWARD_SCALE_FACTOR : 1.0);
+    qreal factor = isFastForward ? SettingsCache::instance().getFastForwardSpeed() : 1.0;
+    timelineWidget->setTimeScaleFactor(factor);
 }
 
 /**
--- a/cockatrice/src/interface/widgets/replay/replay_quick_settings_widget.cpp
+++ b/cockatrice/src/interface/widgets/replay/replay_quick_settings_widget.cpp
@@ -1,0 +1,35 @@
+#include "replay_quick_settings_widget.h"
+
+#include "../../client/settings/cache_settings.h"
+
+ReplayQuickSettingsWidget::ReplayQuickSettingsWidget(QWidget *parent) : SettingsButtonWidget(parent)
+{
+    // fast forward speed
+    fastForwardSpeedBox.setMinimum(1);
+    fastForwardSpeedBox.setValue(SettingsCache::instance().getFastForwardSpeed());
+    connect(&fastForwardSpeedBox, &QDoubleSpinBox::valueChanged, this,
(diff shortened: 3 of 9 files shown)


cockatrice/src/interface/widgets/replay/replay_timeline_widget.cpp
```diff
@@ -186,12 +187,13 @@ void ReplayTimelineWidget::processNewEvents(PlaybackMode playbackMode)
 void ReplayTimelineWidget::setTimeScaleFactor(qreal _timeScaleFactor)
 {
     timeScaleFactor = _timeScaleFactor;
-    replayTimer->setInterval(static_cast<int>(TIMER_INTERVAL_MS / timeScaleFactor));
+    int interval = qRound(TIMER_INTERVAL_MS / timeScaleFactor);
```
ZeldaZach: The interval isn't clamped. `updateTimeScaleFactor` passes `SettingsCache::getFastForwardSpeed()` (an unbounded `qreal`) straight through, so it isn't protected by the spinbox min/max. A stored value > 400 makes `qRound(200 / factor)` == 0, which turns `replayTimer` into a free-running timer that pegs a CPU core and blows through the replay; a value of 0 gives `qRound(inf)` (UB converting inf to int) and a negative value yields a negative interval. Clamp defensively, e.g. `int interval = qMax(1, qRound(TIMER_INTERVAL_MS / timeScaleFactor));` (and/or guard `timeScaleFactor <= 0`).

cockatrice/src/interface/widgets/replay/replay_quick_settings_widget.cpp
```diff
@@ -0,0 +1,35 @@
+#include "replay_quick_settings_widget.h"
+
+#include "../../client/settings/cache_settings.h"
+
+ReplayQuickSettingsWidget::ReplayQuickSettingsWidget(QWidget *parent) : SettingsButtonWidget(parent)
+{
+    // fast forward speed
+    fastForwardSpeedBox.setMinimum(1);
+    fastForwardSpeedBox.setValue(SettingsCache::instance().getFastForwardSpeed());
+    connect(&fastForwardSpeedBox, &QDoubleSpinBox::valueChanged, this,
```
ZeldaZach: `&QDoubleSpinBox::valueChanged` is ambiguous under Qt5 (5.15.2 is still the documented minimum in `cmake/FindQtRuntime.cmake`), where both `valueChanged(double)` and the deprecated `valueChanged(const QString&)` exist; it compiles on Qt6 only. Every QSpinBox connection in this codebase uses `qOverload<int>(...)`; for consistency and Qt5 safety use `qOverload<double>(&QDoubleSpinBox::valueChanged)`. (The same unqualified pattern already exists in `user_card_settings_dialog.cpp`, so this is convention/portability, not a new CI break.)

cockatrice/src/interface/widgets/replay/replay_quick_settings_widget.cpp
```diff
@@ -0,0 +1,35 @@
+#include "replay_quick_settings_widget.h"
+
+#include "../../client/settings/cache_settings.h"
+
+ReplayQuickSettingsWidget::ReplayQuickSettingsWidget(QWidget *parent) : SettingsButtonWidget(parent)
+{
+    // fast forward speed
+    fastForwardSpeedBox.setMinimum(1);
```
ZeldaZach, line 12: Only the minimum is set; the box inherits Qt's defaults (maximum 99.99, 2 decimals, single-step 1.0). The implicit 99.99 cap is arbitrary and undocumented, and 2 decimals are largely meaningless at high speeds since the interval rounds to an int. Consider an explicit `setRange(1, <sensible max>)` plus `setDecimals(...)`/`setSingleStep(...)`. An explicit max also caps the UI side of the interval-clamp concern above.

cockatrice/src/interface/widgets/replay/replay_quick_settings_widget.cpp
```diff
@@ -0,0 +1,35 @@
...
+    // putting it all together
+    auto *widget = new QWidget;
+    auto *grid = new QGridLayout(widget);
+    grid->setContentsMargins(0, 0, 0, 0);
+    grid->addWidget(&fastForwardSpeedLabel, 0, 0, 1, 1);
+    grid->addWidget(&fastForwardSpeedBox, 0, 1, 1, 1);
+
+    this->addSettingsWidget(widget);
+
+    retranslateUi();
```
ZeldaZach, line 29: `retranslateUi()` is only called once at construction. The sibling `VisualDeckStorageQuickSettingsWidget` also does `connect(&SettingsCache::instance(), &SettingsCache::langChanged, this, &...::retranslateUi);` so its label updates on a live language switch. Add the same `langChanged`-to-`retranslateUi` connection here for consistency.

cockatrice/src/interface/widgets/replay/replay_quick_settings_widget.cpp
```diff
@@ -0,0 +1,35 @@
+#include "replay_quick_settings_widget.h"
+
+#include "../../client/settings/cache_settings.h"
```
ZeldaZach: Depth differs from the sibling files in this same directory, which use `../../../client/settings/cache_settings.h` (e.g. `replay_timeline_widget.cpp`). This resolves via the project's broad include dirs, but for consistency with neighbors use `../../../client/settings/cache_settings.h`. Also consider explicitly including `<QGridLayout>`/`<QWidget>` (this .cpp) and `<QLabel>` (the .h), which are currently only available transitively.

cockatrice/CMakeLists.txt
```diff
@@ -227,6 +227,7 @@ set(cockatrice_SOURCES
     src/interface/widgets/printing_selector/set_name_and_collectors_number_display_widget.cpp
     src/interface/widgets/quick_settings/settings_button_widget.cpp
     src/interface/widgets/quick_settings/settings_popup_widget.cpp
+    src/interface/widgets/replay/replay_quick_settings_widget.cpp
```
ZeldaZach: The list is otherwise alphabetical within a directory, but this is inserted before `replay/replay_manager.cpp` (line 231). Move it after `replay_manager.cpp` to keep the `replay/` group sorted. Registering only the .cpp is correct; AUTOMOC handles the Q_OBJECT header.

cockatrice/src/interface/widgets/replay/replay_manager.cpp
```diff
@@ -78,13 +79,19 @@ ReplayManager::ReplayManager(TabGame *parent, GameReplay *_replay)
     replayFastForwardButton->setIconSize(QSize(32, 32));
     replayFastForwardButton->setIcon(QPixmap("theme:replay/fastforward"));
     replayFastForwardButton->setCheckable(true);
-    connect(replayFastForwardButton, &QToolButton::toggled, this, &ReplayManager::replayFastForwardButtonToggled);
+    connect(replayFastForwardButton, &QToolButton::toggled, this, &ReplayManager::updateTimeScaleFactor);
+
+    settingsWidget = new ReplayQuickSettingsWidget(this);
```
ZeldaZach, line 84: Pre-existing and out of scope, but this PR adds a per-replay-tab consumer: `SettingsButtonWidget` creates its `popup` with `new SettingsPopupWidget(nullptr)` (no parent) and has no destructor deleting it, so each replay tab leaks a popup (plus the inner `widget`) on close. Not introduced here and not a blocker, but worth noting since replay tabs are opened/closed repeatedly. A fix belongs in `SettingsButtonWidget`.

ZeldaZach (commented):
Solid, focused change. The default of 10 preserves the old hardcoded 10x behavior exactly, the scaling direction matches the "x times faster" semantics, and live updates work because the spinbox's `fastForwardSpeedChanged` signal is reconnected in `ReplayManager` to re-apply the interval on a running timer. Widget ownership is safe: `settingsWidget` is parented to `ReplayManager`, and the value-member spinbox/label reparented into the popup are torn down as derived members before the base class runs.

Status: **approve with comments**. The one substantive concern is that `setTimeScaleFactor` never clamps the computed interval: `updateTimeScaleFactor` reads the raw `qreal` from `SettingsCache` (not the bounded spinbox), so a config value > 400 rounds the timer interval to 0 (a free-running timer) and a value <= 0 produces `qRound(inf)` UB or a negative interval. The rest are minor: the Qt5-ambiguous `&QDoubleSpinBox::valueChanged` (codebase convention is `qOverload`), no explicit spinbox max/decimals, and a missing `langChanged`-to-`retranslateUi` hookup that the sibling quick-settings widget has.

ZeldaZach (approved):
Re-reviewed after the follow-up commits — all concerns from the prior pass are addressed:

- `setTimeScaleFactor` now clamps with `std::max(1, qRound(TIMER_INTERVAL_MS / timeScaleFactor))`, so high speeds can no longer round the interval to 0 (free-running timer) and negative intervals are ruled out. Combined with the spinbox `setMinimum(1)`, every UI-reachable value is safe.
- Signal connection uses `qOverload<double>(&QDoubleSpinBox::valueChanged)`.
- Spinbox has explicit bounds/decimals (`setMinimum(1)`, `setMaximum(99.9)`, `setDecimals(1)`).
- `retranslateUi` is wired to `SettingsCache::langChanged`.

Live updates re-apply the interval on the running timer, widget lifetime is safe, and the default of 10 preserves the old hardcoded 10x behavior. CI is green across all platforms.

Non-blocking nit: `updateTimeScaleFactor` reads the raw qreal from SettingsCache rather than the bounded spinbox, so a hand-edited config value of exactly 0 yields `qRound(inf)` (technically UB, made benign by the max(1, …) clamp). A `factor > 0 ? factor : 1.0` guard would close it if you like, but not a blocker.

LGTM.

BruebachL:
I think this fixed #1171?

PR #7038, opened 2026-07-08
Fix and rework client updater logic

tooomm:
- Some versions were hardcoded, e.g. "macOS 13" for Intel Macs, or outdated like "Windows 7" that we stopped supporting.
- Logic was looking for a matching target version in file name and major host OS version (`rel_maj == sys_maj`) on Mac. This was changed in https://github.com/Cockatrice/Cockatrice/pull/5589 some time back next to the regex (unintentionally?), it did then no longer match the comment nor the actual OS behavior.
That should also mean users running macOS 26 are not be able to use the in-client updater right now as we do not provide a binary targeting "macOS26" and newer --> no 26 == 26 match? 🤔 
- Order of returned assets from GitHub mattered
- Generic regex and no reliable separation between app bundles for Intel & ARM Macs

In general it was a bit static and outdated.

Make the logic more flexible and future proof.

- **Properly implement concept of releases targeting former versions will run on newer OS versions without issues (applies on Win + macOS)**
  Old code comment claimed to do that already before
- **Dynamically select the best match (newest release amongst all available + compatible ones for the OS/Arch you're running on)**
- More explicit regex and proper separation for Intel binaries on Mac

Appreciate a throughout check from you guys. AI helped.

<br>

---

Tested and confirmed working on Windows 11 with our existing (pre)releases and file naming:
- Beta Channel:
  ```
  [2026-07-08 21:43:49.857 I] [unknown] - Searching for updates on the channel:  "https://api.github.com/repos/Cockatrice/Cockatrice/releases" [unknown:0]
  [2026-07-08 21:43:50.860 I] [unknown] - Got reply from release server, size= 20 name= "2026-06-27-Development-3.1.0-beta.3 (ad49225)" desc= "https://github.com/Cockatrice/Cockatrice/compare/6662d46...ad49225" commit= "ad4922537ddef6f1f06007b2978dae7e12a7af46" date= QDate("2026-06-27") [unknown:0]
  [2026-07-08 21:43:50.860 I] [unknown] - Searching for a corresponding file on the beta channel:  "https://api.github.com/repos/Cockatrice/Cockatrice/releases/345763947/assets" [unknown:0]
  [2026-07-08 21:43:51.056 I] [unknown] - Current hash= "6662d46" update hash= "ad49225" [unknown:0]
  [2026-07-08 21:43:51.056 I] [unknown] - Best compatible asset= "https://github.com/Cockatrice/Cockatrice/releases/download/2026-06-27-Development-3.1.0-beta.3/Cockatrice-3.1.0-beta.3-Win10.exe" [unknown:0]
  ```

- Stable Channel:
  ```
  [2026-07-08 21:46:56.416 I] [unknown] - Searching for updates on the channel:  "https://api.github.com/repos/Cockatrice/Cockatrice/releases/latest" [unknown:0]
  [2026-07-08 21:46:56.864 I] [unknown] - Best compatible asset= "https://github.com/Cockatrice/Cockatrice/releases/download/2026-06-26-Release-3.0.2/Cockatrice-Graduation-Day-3.0.2-Win10.exe" [unknown:0]
  [2026-07-08 21:46:56.864 I] [unknown] - Current hash= "6662d46" update hash= "" [unknown:0]
  [2026-07-08 21:46:56.864 I] [unknown] - Got reply from release server, name= "Cockatrice 3.0.2: Graduation Day" desc= "https://github.com/Cockatrice/Cockatrice/releases/tag/2026-06-26-Release-3.0.2" date= QDate("2026-06-26") url= "https://github.com/Cockatrice/Cockatrice/releases/download/2026-06-26-Release-3.0.2/Cockatrice-Graduation-Day-3.0.2-Win10.exe" [unknown:0]
  [2026-07-08 21:46:56.864 I] [unknown] - Searching for commit hash corresponding to stable channel tag:  "2026-06-26-Release-3.0.2" [unknown:0]
  [2026-07-08 21:46:57.058 I] [unknown] - Got reply from tag server, commit= "adfeb2234d14f040f2d73ba8ecab92e3dff4a5a8" [unknown:0]
  [2026-07-08 21:46:57.058 I] [unknown] - Current hash= "6662d46" update hash= "adfeb22" [unknown:0]
  ```

### DIFF
--- a/cockatrice/src/client/network/update/client/release_channel.cpp
+++ b/cockatrice/src/client/network/update/client/release_channel.cpp
@@ -63,10 +63,6 @@ std::optional<int> ReleaseChannel::getTargetVersionForCurrentOS(const QString &f
     const QRegularExpression &regex = isIntel ? macIntelRegex : macArmRegex;
 
 #elif defined(Q_OS_WIN)
-#if Q_PROCESSOR_WORDSIZE != 8 // non 64-bit host
-    Q_UNUSED(fileName);
-    return std::nullopt;
-#endif
     static const QRegularExpression regex(R"(Win(?:dows)?(\d+)\.[^.]+$)", QRegularExpression::CaseInsensitiveOption);
 
 #else // if the OS doesn't fit one of the above #defines, then it will never match
--- a/cockatrice/src/client/network/update/client/release_channel.cpp
+++ b/cockatrice/src/client/network/update/client/release_channel.cpp
@@ -44,13 +44,28 @@ void ReleaseChannel::checkForUpdates()
     connect(response, &QNetworkReply::finished, this, &ReleaseChannel::releaseListFinished);
 }
 
-// Find compatible assets for host platform (Linux is not supported by in-client updater)
+// Find assets compatible with host platform (Linux is not supported by in-client updater)
 std::optional<int> ReleaseChannel::getTargetVersionForCurrentOS(const QString &fileName)
 {
     const QRegularExpression *regex = nullptr;
 
+    static const std::optional<int> systemVersion = [] {
+        bool ok = false;
+        const QString versionString = QSysInfo::productVersion();
+        const int version = versionString.split('.').first().toInt(&ok);
+        if (!ok) {
+            qCWarning(ReleaseChannelLog) << "Unable to determine OS version from" << versionString;
+            return std::optional<int>();
+        }
+        return std::optional<int>{version};
+    }();
+
+    if (!systemVersion) {
+        return std::nullopt;
+    }
+
 #if defined(Q_OS_MACOS)
-    const bool isIntel = [] {
+    static const bool isIntel = [] {
         // QSysInfo does not go through translation layers
         // We need to use sysctl to reliably detect the underlying architecture
         char arch[255];
@@ -60,12 +75,16 @@ std::optional<int> ReleaseChannel::getTargetVersionForCurrentOS(const QString &f
         }
         return false;
     }();
+
+    // Apple (ARM) --> Cockatrice-3.0.0-macOS15.dmg
+    // Apple (x86) --> Cockatrice-3.0.0-macOS15_Intel.dmg
     static const QRegularExpression macIntelRegex(R"(-macOS(\d+)_Intel\.[^.]+$)",
                                                   QRegularExpression::CaseInsensitiveOption);
-    static const QRegularExpression macArmRegex(R"(-macOS(\d+)\.[^.]+$)", QRegularExpression::CaseInsensitiveOption);
-    regex = isIntel ? &macIntelRegex : &macArmRegex;
+    static const QRegularExpression macRegex(R"(-macOS(\d+)\.[^.]+$)", QRegularExpression::CaseInsensitiveOption);
+    regex = isIntel ? &macIntelRegex : &macRegex;
 
 #elif defined(Q_OS_WIN)
+    // Windows (x86) --> Cockatrice-3.0.0-Windows10.exe
     static const QRegularExpression winRegex(R"(-(?:Win|Windows)(\d+)\.[^.]+$)",
                                              QRegularExpression::CaseInsensitiveOption);
     regex = &winRegex;
@@ -75,14 +94,13 @@ std::optional<int> ReleaseChannel::getTargetVersionForCurrentOS(const QString &f
     return std::nullopt;
 #endif
 
-    // Any asset targeting an OS version smaller or equal to the current host works
-    const int systemVersion = QSysInfo::productVersion().split('.').first().toInt();
+    // Any asset targeting an OS version up to the current host version works
     auto match = regex->match(fileName);
     if (!match.hasMatch()) {
         return std::nullopt;
     }
     const int targetVersion = match.captured(1).toInt();
-    if (targetVersion > systemVersion) {
+    if (targetVersion > *systemVersion) {
         return std::nullopt;
     }
     return targetVersion;
--- a/cockatrice/src/client/network/update/client/release_channel.cpp
+++ b/cockatrice/src/client/network/update/client/release_channel.cpp
@@ -2,7 +2,6 @@
 
 #include "version_string.h"
 
-#include <optional>
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
@@ -11,6 +10,7 @@
 #include <QRegularExpression>
 #include <QSysInfo>
 #include <QtGlobal>
+#include <optional>
 
 #if defined(Q_OS_MACOS)
 #include <sys/sysctl.h>
@@ -47,6 +47,8 @@ void ReleaseChannel::checkForUpdates()
 // Different release channel checking functions for different operating systems
 std::optional<int> ReleaseChannel::getTargetVersionForCurrentOS(const QString &fileName)
 {
+    const QRegularExpression *regex = nullptr;
+
 #if defined(Q_OS_MACOS)
     const bool isIntel = [] {
         // QSysInfo does not go through translation layers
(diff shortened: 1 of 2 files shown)


cockatrice/src/client/network/update/client/release_channel.h
```diff
@@ -96,7 +97,8 @@ class ReleaseChannel : public QObject
     Release *lastRelease;
 
 protected:
-    static bool downloadMatchesCurrentOS(const QString &fileName);
+    std::optional<int> getTargetVersionForCurrentOS(const QString &fileName);
```
RickyRister: nit: can be static

RickyRister:
Someone else will have to check if it works properly. I'm not familiar with the update logic for the other OS's

tooomm:
> Someone else will have to check if it works properly. I'm not familiar with the update logic for the other OS's

Well, the logic of the in-client updater should find you a working binary for your platform from the selected channel (stable full release vs. beta pre-releases). The updater only supports macOS and Windows.
I made it pick the newest available and compatible target OS (not required, the oldest would work as well).

It works for me on Windows.

<br>

@RickyRister You're running macOS if I'm not mistaken?
Which version are you using?  If you could run the internal client updater from the `Help` menu (e.g. latest release on beta channel --> fetching latest beta) and confirm if that works?
I'm suspecting it will not if you're on macOS 26 (see comment in the PR description).

With the macOS binaries from the PR (download [here](https://github.com/Cockatrice/Cockatrice/actions/runs/29594972691?pr=7038#artifacts)) it should work.

<br>

@BruebachL Which OS are you on?

RickyRister:
I'm still on macOS 15.7.1. And yes, I did confirm that the updater works for me

ZeldaZach (commented):
Nice, well-targeted rework of the updater matching. I checked the three regexes against our actual release asset names (`*-macOS13_Intel.dmg`, `*-macOS14.dmg`, `*-macOS15.dmg`, `*-Win10.exe`, plus the Linux `.deb`/`.rpm`): the anchoring is correct, the Intel/ARM split is clean (the ARM regex correctly rejects the `_Intel` build), and there's no backtracking concern. The `targetVersion <= systemVersion` comparison is the right fix for the old `rel_maj == sys_maj` behavior from #5589, so users on newer macOS/Windows now get the newest compatible build instead of nothing, and the selection is order-independent for distinct versions. Empty/malformed replies are handled without crashing.

Status: **approve with comments**. A few non-blocking things worth considering inline: tie-breaking between equal-version assets is still list-order dependent; Windows lost the architecture separation that Mac gained (fine for the single x64 build today); an Apple-Silicon Mac on macOS 13 now matches no asset where it previously fell back to the Intel dmg; and a `productVersion()` parse failure silently disables updates. None of these affect the current release naming, so this is good to merge once you've considered them.

cockatrice/src/client/network/update/client/release_channel.cpp
```diff
@@ -43,55 +44,72 @@ void ReleaseChannel::checkForUpdates()
...
+    int bestVersion = -1;
+    for (const auto &rawAsset : assets) {
+        QVariantMap asset = rawAsset.toMap();
+        QString name = asset["name"].toString();
+        QString url = asset["browser_download_url"].toString();
+        auto version = getTargetVersionForCurrentOS(name);
+        if (!version) {
+            continue;
+        }
+        if (*version > bestVersion) {
```
ZeldaZach, line 122: Strict `>` means that when two compatible assets share the same target version, the first one in the list wins, so selection among equal-version assets is still order-dependent. It doesn't bite today (each platform+arch has a unique target version per release), but it partly undercuts the "order no longer matters" goal. If a same-version collision ever becomes possible (e.g. two Windows arch variants both tagged `Win10`), the tie would be resolved silently by GitHub's asset ordering.

cockatrice/src/client/network/update/client/release_channel.cpp
```diff
@@ -43,55 +44,72 @@ void ReleaseChannel::checkForUpdates()
...
-    if (version.startsWith("7") || version.startsWith("8")) {
-        return fileName.contains("Win7");
-    } else {
-        return fileName.contains("Win10");
-    }
-#else
-    Q_UNUSED(fileName);
-    return false;
-#endif
+    static const QRegularExpression winRegex(R"(-(?:Win|Windows)(\d+)\.[^.]+$)",
```
ZeldaZach, line 88: Mac now separates Intel vs ARM, but Windows has no architecture separation (the old `Q_PROCESSOR_WORDSIZE`/`32bit` handling was dropped). Fine for the current single `Win10.exe` (x64) build, but if an arm64 or 32-bit Windows asset is ever added with the same OS-version token, both would match `winRegex` and the choice falls to list order with no way to pick the right ABI. Worth a short comment noting Windows currently assumes a single x64 artifact.

cockatrice/src/client/network/update/client/release_channel.cpp
```diff
@@ -43,55 +44,72 @@ void ReleaseChannel::checkForUpdates()
...
-
-    // older(smaller) releases are compatible with a newer or the same system version
-    int sys_maj = getSystemVersion();
-    int rel_maj = match.captured(1).toInt();
-    return rel_maj == sys_maj;
+        return false;
+    }();
+    static const QRegularExpression macIntelRegex(R"(-macOS(\d+)_Intel\.[^.]+$)",
+                                                  QRegularExpression::CaseInsensitiveOption);
+    static const QRegularExpression macArmRegex(R"(-macOS(\d+)\.[^.]+$)", QRegularExpression::CaseInsensitiveOption);
```
ZeldaZach: Behavioral change worth confirming is intended: because the only ARM builds are `macOS14`/`macOS15`, an Apple-Silicon Mac running macOS 13 now matches no asset (this regex correctly rejects the `_Intel` build). Previously the loose `macOS(\d+)` regex would have handed it the Intel `.dmg`. Arguably more correct (no wrong-arch binary), but the net effect is those users silently get "no update available" — either confirm that's acceptable or make sure an ARM macOS 13 build exists. Minor: `\.[^.]+$` also assumes a single-segment extension, so a future `...-macOS14.app.zip`-style name wouldn't match.

cockatrice/src/client/network/update/client/release_channel.cpp
```diff
@@ -43,55 +44,72 @@ void ReleaseChannel::checkForUpdates()
...
 
-#else // If the OS doesn't fit one of the above #defines, then it will never match
+#else // if the OS doesn't fit one of the above #defines, then it will never match
     Q_UNUSED(fileName);
-    return false;
+    return std::nullopt;
 #endif
+
+    // Any asset targeting an OS version smaller or equal to the current host works
+    const int systemVersion = QSysInfo::productVersion().split('.').first().toInt();
```
ZeldaZach: If `productVersion()` can't be parsed to an int it yields 0, so every asset (target > 0) is rejected and the updater silently offers nothing. The failure direction is safe (no wrong match) but invisible — a debug log on a 0/parse-failure would help diagnose. Also note this line plus the macOS `sysctl`/`isIntel` detection run once per asset inside the loop; the host OS/arch could be computed once before the loop and passed in.

cockatrice/src/client/network/update/client/release_channel.h
ZeldaZach: Both `getTargetVersionForCurrentOS` and `findBestDownloadUrl` use no instance state, yet dropped the `static` that `downloadMatchesCurrentOS` had. Keeping them `static` matches the prior style and documents that they don't depend on `this`.

cockatrice/src/client/network/update/client/release_channel.cpp
tooomm: Yes, that's intentional and matches the platforms supported by the project.
Intel support goes down to macOS 13 and Apple M support goes down to macOS 14 currently.
That's not new behavior, the logic in the updater was just outdated.

Anyway an edge case. macOS 13 is out of service for a while now and is pretty outdated in general. There are several mature main OS versions released after that and all Apple M Mac's support running newer macOS versions.

<br>

>Minor: `\.[^.]+$` also assumes a single-segment extension, so a future `...-macOS14.app.zip`-style name wouldn't match.

That's correct. We will very likely not use such file extensions on macOS or Windows though.

Only alternative to consider is strictly including file extensions like .dmg and .exe in the regex I guess?
That would fail once we add or switch to .msi installer on Windows for example.

cockatrice/src/client/network/update/client/release_channel.cpp
tooomm: Isn't that rather theoretical and would indicate a misconfiguration in our build pipeline?
We should never have binaries with identical names but different version or platform targets.

Our pipeline setup follows this structure:
- Mac
  - `...-macOS<VERSION>.dmg`   <-- default Mac (ARM, 64 bit)
  - `...-macOS<VERSION>_Intel.dmg`   <-- additional build for legacy Mac (x86, 64 bit)
- Win
  - `...-Win<VERSION>.exe`   <-- default Windows machine (x86, 64 bit)
  - (no 32bit build)
  - (no ARM arch)

If we would introduce a new ARM binary for niche Windows systems, it would be labeled `-Win10_ARM.exe`, next to the standard x86 `-Win10.exe`.
The updater can not handle ARM right now for Windows and it would never match the existing `winRegex`.
We would then probably extend the Windows logic similarly to how I introduced it with Intel Macs in this PR.

cockatrice/src/client/network/update/client/release_channel.cpp
tooomm: 32bit Windows snippet was a legacy leftover, it's no longer needed and will not come back.
I did clean that up while I was here.

If we ever decide to add ARM support on Windows, the asset would get an `_ARM` appendix similar to `_Intel` on Mac right now. The name would not match `winRegex` (but a new `winArmRegex`).

Added comments to the regex patterns with example asset names for clarity.

BruebachL (approved):
I've added some minor changes but this LGTM

PR #7041, opened 2026-07-09
Fix card position when added to the stack

DawnFire42:

I found a bug I introduced when refactoring/improving the vertical zones stacking behaviour. Cards added to the stack ended up being placed at the bottom of the stack instead of the top and couldn't be repositioned to the top even from within the zone.

- The stack ordering behaviour should work as expected now: cards appended to the top of the stack when coming from another zone or reordered properly if already in the stack.


RickyRister (approved):
Works. LGTM

https://github.com/user-attachments/assets/4294b383-24da-495e-b3dd-ae6ef266e57f

PR #7043, opened 2026-07-12
fix crash when pressing Back to results

ebbit1q:
returning to the search when viewing a deck in the archidekt browser while not viewing the first page, ie scroll down until it loads more, caused a crash, using doSearch now resets the page entirely, this however doesn't seem to be the best experience, however it never actually saved the scroll bar location or anything so it's just how it worked before but without the crash

reported on our discord by v0geekay0v


RickyRister (approved):
LGTM

PR #7046, opened 2026-07-17
Docker: Build ARM image natively

tooomm:
The multi-arch Docker build used QEMU emulation and was very slow for ARM.
Total build time was often 50min.

- Build images on native architectures in parallel (GitHub has ARM runners as well)
  First build took 5min per arch (in parallel): https://github.com/Cockatrice/Cockatrice/actions/runs/29617876779
  - On releases: build + push images & digests
  - Other triggers: build only
- On completion, merge both into one multi-arch manifest and publish new tags (only for releases)

See e.g. https://obviy.us/blog/fastest-gha-multi-arch-docker
and https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

<br>

Additional change:
- Use Ninja generator, same as in the main CI pipeline
- ~~Use ccache compiler cache, same as in the main CI pipeline~~
  According to the ["cache mount" docker docs](https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts), BuildKit layer caching is not including cache mounts into GHA Cache. Setting "mode=max" and "type=gha" in `docker/build-push-action` is not changing that, confirmed this in this PR with a test run and checking "ccache --show-stats" as well.
The official documentation suggests to use another 3rd party action to manage and inject the compiler cache manually. As the performance is drastically increased by avoiding emulation alone, I think the extra complexity is not needed.
- Add SBOM

<br>

Performance impact: 10x quicker for uncached builds
- Old workflow
  - 1st run (QEMU emulation for ARM): **50 min** (uncached)
  - 2nd run (QEMU emulation for ARM): 30 s (layers cached)
- New workflow
  - ~~1st run (native ARM compilation + ninja + ccache): 6 min (uncached)~~
  - ~~2nd run (native ARM compilation + ninja + ccache): 30 s (layers cached)~~
  - ~~3rd run (native ARM compilation + ninja + ccache): 4 min (cpp change, dependency layer cached)~~
  - 4th run (native ARM compilation + ninja): **5 min** (uncached)
  - 5th run (native ARM compilation + ninja): 30 s (layers cached)

<br>

@maybeanerd Feel free to review & comment!

<br>

> [!NOTE]
> TODO after next release
> ```
> docker buildx imagetools inspect ghcr.io/cockatrice/servatrice:latest
> docker buildx imagetools inspect ghcr.io/cockatrice/servatrice:latest --raw
> ```
> - [ ] Verify attestations are present in single- and multi-platform images
> - [ ] Below labels and annotations are automatically added:
>   ```
>   "org.opencontainers.image.source": "https://github.com/Cockatrice/Cockatrice"
>   "org.opencontainers.image.licenses": "GPL-2.0"
>   "org.opencontainers.image.version": "<GITHUB_TAG>"
>   ```
> - [ ] Consider adding non-raw inspect output to GIHTUB_SUMMARY

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -14,7 +14,10 @@ on:
       - master
     paths:
       - '.github/workflows/docker-release.yml'
+      - '.dockerignore'
       - 'Dockerfile'
+      - 'docker-compose.yml'
+      - 'docker-compose.yml.windows'
   release:
     types:
       - released    # publishing of stable releases
@@ -25,26 +28,30 @@ concurrency:
   cancel-in-progress: ${{ github.event_name != 'release' }}
 
 env:
-  IMAGE_NAME: ghcr.io/cockatrice/servatrice
+  GHCR_IMAGE: ghcr.io/cockatrice/servatrice
   OCI_DESCRIPTION: Server for Cockatrice, a cross-platform virtual tabletop for multiplayer card games
+  OCI_SOURCE: https://github.com/Cockatrice/Cockatrice
   OCI_TITLE: Servatrice
   OCI_URL: https://cockatrice.github.io/
+  OCI_VERSION: ${{ github.ref_name }}
 
 jobs:
   build:
+    name: "${{ matrix.label }}"
+    if: github.repository_owner == 'Cockatrice'
+    runs-on: ${{ matrix.runner }}
+
     strategy:
       fail-fast: false
       matrix:
         include:
-          - platform: linux/amd64
+          - label: x86
+            platform: linux/amd64
             runner: ubuntu-latest
 
-          - platform: linux/arm64
-            runner: ubuntu-24.04-arm
-
-    name: "Build (${{ matrix.platform }})"
-    if: github.repository_owner == 'Cockatrice'
-    runs-on: ${{ matrix.runner }}
+          - label: arm
+            platform: linux/arm64
+            runner: ubuntu-24.04-arm    # to be replaced with "ubuntu-latest-arm" once available
 
     steps:
       - name: "Checkout"
@@ -58,36 +65,40 @@ jobs:
         id: login
         uses: docker/login-action@v4
         with:
-          password: ${{ github.token }}
           registry: ghcr.io
           username: ${{ github.actor }}
+          password: ${{ github.token }}
 
-      - name: "Build Docker image and push by digest"
-        if: steps.login.outcome == 'success'
-        id: build
-        uses: docker/build-push-action@v7
-        with:
-          cache-from: type=gha,scope=servatrice-${{ matrix.platform }}
-          cache-to: type=gha,mode=max,scope=servatrice-${{ matrix.platform }}
-          context: .
-          labels: |
-            org.opencontainers.image.description=${{ env.OCI_DESCRIPTION }}
-            org.opencontainers.image.title=${{ env.OCI_TITLE }}
-            org.opencontainers.image.url=${{ env.OCI_URL }}
-          outputs: type=image,push-by-digest=true,name-canonical=true,push=true
-          platforms: ${{ matrix.platform }}
-
-      - name: "Build Docker image"
+      - name: "Build image"
         if: steps.login.outcome != 'success'
         uses: docker/build-push-action@v7
         with:
-          cache-from: type=gha,scope=servatrice-${{ matrix.platform }}
-          cache-to: type=gha,mode=max,scope=servatrice-${{ matrix.platform }}
+          cache-from: type=gha,scope=servatrice-${{ matrix.label }}
+          cache-to: type=gha,mode=max,scope=servatrice-${{ matrix.label }}
           context: .
           platforms: ${{ matrix.platform }}
           push: false
 
-      - name: Export digest
+      - name: "Build image and push by digest"
+        if: steps.login.outcome == 'success'
+        id: build
+        uses: docker/build-push-action@v7
+        with:
+          cache-from: type=gha,scope=servatrice-${{ matrix.label }}
+          cache-to: type=gha,mode=max,scope=servatrice-${{ matrix.label }}
+          context: .
+          labels: |
+            org.opencontainers.image.description=${{ env.OCI_DESCRIPTION }}
+            org.opencontainers.image.source=${{ env.OCI_SOURCE }}
+            org.opencontainers.image.title=${{ env.OCI_TITLE }}
+            org.opencontainers.image.url=${{ env.OCI_URL }}
+            org.opencontainers.image.version=${{ env.OCI_VERSION }}
+          outputs: type=image,push-by-digest=true,name-canonical=true,push=true
+          platforms: ${{ matrix.platform }}
+          provenance: mode=max
+          sbom: true
+
+      - name: "Export digest"
         if: steps.login.outcome == 'success'
         env:
           DIGEST: ${{ steps.build.outputs.digest }}
@@ -95,22 +106,25 @@ jobs:
           mkdir -p $RUNNER_TEMP/digests
           touch "$RUNNER_TEMP/digests/${DIGEST#sha256:}"
 
-      - uses: actions/upload-artifact@v7
+      - name: "Upload digest"
         if: steps.login.outcome == 'success'
+        uses: actions/upload-artifact@v7
         with:
+          if-no-files-found: error
           name: digest-${{ matrix.platform }}
           path: $RUNNER_TEMP/digests/*
           retention-days: 1
 
 
   merge:
-    needs: build
-    name: "Publish manifest"
+    name: "Publish multi-arch manifest"
     if: github.repository_owner == 'Cockatrice' && github.event_name == 'release' && github.event.release.prerelease == false
+    needs: build
     runs-on: ubuntu-slim
 
     steps:
-      - uses: actions/download-artifact@v4
+      - name: "Download digests"
+        uses: actions/download-artifact@v4
         with:
           path: $RUNNER_TEMP/digests
           pattern: digest-*
@@ -122,15 +136,15 @@ jobs:
       - name: "Login to GitHub Container Registry (GHCR)"
         uses: docker/login-action@v4
         with:
-          password: ${{ github.token }}
           registry: ghcr.io
           username: ${{ github.actor }}
+          password: ${{ github.token }}
 
       - name: "Docker metadata"
         id: metadata
         uses: docker/metadata-action@v6
         with:
-          images: ${{ env.IMAGE_NAME }}
+          images: ${{ env.GHCR_IMAGE }}
           tags: |
             type=raw,value=latest
             type=ref,event=tag
@@ -147,10 +161,14 @@ jobs:
 
           docker buildx imagetools create \
             --annotation "index:org.opencontainers.image.description=$OCI_DESCRIPTION" \
+            --annotation "index:org.opencontainers.image.source=$OCI_SOURCE" \
             --annotation "index:org.opencontainers.image.title=$OCI_TITLE" \
             --annotation "index:org.opencontainers.image.url=$OCI_URL" \
+            --annotation "index:org.opencontainers.image.version=$OCI_VERSION" \
             $TAG_ARGS \
-            $(printf '%s@sha256:%s ' "$IMAGE_NAME" *)
+            $(printf '%s@sha256:%s ' "$GHCR_IMAGE" *)
 
-      - name: "Inspect image"
-        run: docker buildx imagetools inspect "$IMAGE_NAME:latest"
+      - name: "Inspect image via tags"
+        run: |
+          docker buildx imagetools inspect "$GHCR_IMAGE:latest"
+          docker buildx imagetools inspect "$GHCR_IMAGE:$OCI_VERSION"
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -148,7 +148,7 @@ jobs:
           flavor: |
             latest=auto
           tags: |
-            type=match,pattern=^\d{4}-\d{2}-\d{2}-Release-(\d+\.\d+\.\d+)$,group=1    # if semver: type=ref,event=tag and type=semver,pattern={{version}} / {{major}}.{{minor}}
+            type=ref,event=tag    # if semver, also: type=semver,pattern={{version}} / {{major}}.{{minor}}
 
       # Add OCI annotations to image index and publish tags
       - name: "Create image index"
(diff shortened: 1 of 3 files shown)


maybeanerd (approved):
Nice work! LGTM, although I don't have much experience with combining manifests/digests, so we'll just need to hope that part also works 🤞

tooomm:
> Nice work! LGTM, although I don't have much experience with combining manifests/digests, so we'll just need to hope that part also works 🤞

Thanks for your comment and having a look!

I followed the linked blog post (Docker docs use their own reusable workflows which hide a lot of the logic) and read about the [`buildx imagetools`](https://docs.docker.com/reference/cli/docker/buildx/imagetools/) commands.
I expect it to work, but did not test it on another repo. As we did with earlier iterations for the Servatrice docker image already, I'm looking forward to test it with the next release.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+        uses: docker/build-push-action@v7
+        with:
+          cache-from: type=gha,scope=${{ env.CACHE_SCOPE }}
+          cache-to: type=gha,mode=max,scope=${{ env.CACHE_SCOPE }}
+          context: .
+          labels: |
+            org.opencontainers.image.description=${{ env.OCI_DESCRIPTION }}
+            org.opencontainers.image.title=${{ env.OCI_TITLE }}
+            org.opencontainers.image.url=${{ env.OCI_URL }}
+          outputs: type=image,push-by-digest=true,name-canonical=true,push=true
```
ZeldaZach: Missing `name=`: with `push=true`, no `tags:`, and no metadata step in this `build` job, buildx has no target repository and errors with "tag is needed when pushing to registry". `name-canonical=true` is a no-op without a `name=`. Use `outputs: type=image,name=${{ env.GHCR_IMAGE }},push-by-digest=true,name-canonical=true,push=true`, matching the official multi-runner pattern.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+            TAG_ARGS="$TAG_ARGS --tag $tag"
+          done
+
+          docker buildx imagetools create \
+            --prefer-index=true \
+            --annotation "index:org.opencontainers.image.description=$OCI_DESCRIPTION" \
+            --annotation "index:org.opencontainers.image.title=$OCI_TITLE" \
+            --annotation "index:org.opencontainers.image.url=$OCI_URL" \
+            $TAG_ARGS \
+            $(printf '%s@sha256:%s ' "$GHCR_IMAGE" *)
```
ZeldaZach: This corrupts the reference list. `printf` reuses the format string for every argument, but `$GHCR_IMAGE` is only supplied once, so with two digest files you get `ghcr.io/.../servatrice@sha256:<d1> <d2>@sha256:` (reproduced locally) and `imagetools create` rejects the malformed second ref (empty digest, `<d2>` as repo). Put the image in the format literal and pass only the digest filenames: `$(printf "$GHCR_IMAGE@sha256:%s " *)`.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+          path: $RUNNER_TEMP/digests/*
+          retention-days: 1
+
+
+  # Create an OCI image index from the platform-specific image manifests
+  index:
+    name: "Publish multi-platform image"
+    if: github.repository_owner == 'Cockatrice' && github.event_name == 'release' && github.event.release.prerelease == false
+    needs: build
+    runs-on: ubuntu-slim
```
ZeldaZach: `ubuntu-slim` is not a GitHub-hosted runner label (valid: `ubuntu-latest`, `ubuntu-24.04`, `ubuntu-22.04`, `ubuntu-24.04-arm`). Unless this is a private self-hosted label the org actually has, the `index` job will sit in "Waiting for a runner" and never publish. Use `ubuntu-latest`.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+        run: |
+          mkdir -p $RUNNER_TEMP/digests
+          touch "$RUNNER_TEMP/digests/${DIGEST#sha256:}"
+
+      - name: "Upload digest"
+        if: steps.login.outcome == 'success'
+        uses: actions/upload-artifact@v7
+        with:
+          if-no-files-found: error
+          name: digest-${{ matrix.platform }}
```
ZeldaZach: `matrix.platform` is `linux/amd64` / `linux/arm64`, so this resolves to `digest-linux/amd64`, and `/` is illegal in an `actions/upload-artifact` v4 name, so the upload step fails. The canonical pattern derives a slash-free suffix (e.g. replace `/` with `-` into a `platform-pair` value) giving `digest-linux-amd64`.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+          path: $RUNNER_TEMP/digests/*
```
ZeldaZach: `$RUNNER_TEMP` is a shell variable, and the runner does not expand it inside `with:` inputs (only `${{ }}` expressions), so upload-artifact globs a literal `$RUNNER_TEMP/...` path, finds nothing, and `if-no-files-found: error` fails the step. Use `path: ${{ runner.temp }}/digests/*`. (The `run:` steps that use `$RUNNER_TEMP` are fine because they're shells.)

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+    name: "Publish multi-platform image"
+    if: github.repository_owner == 'Cockatrice' && github.event_name == 'release' && github.event.release.prerelease == false
+    needs: build
+    runs-on: ubuntu-slim
+
+    steps:
+      - name: "Download digests"
+        uses: actions/download-artifact@v4
+        with:
+          path: $RUNNER_TEMP/digests
```
ZeldaZach: Same unexpanded-`$RUNNER_TEMP` issue as the upload step, and it's inconsistent with the merge step's `working-directory: ${{ runner.temp }}/digests`. Downloaded digests would land under a literal `$RUNNER_TEMP` directory while the merge reads the real temp dir (empty), so even after fixing the other items the manifest would be built from no digests. Use `path: ${{ runner.temp }}/digests`.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+            --annotation "index:org.opencontainers.image.description=$OCI_DESCRIPTION" \
+            --annotation "index:org.opencontainers.image.title=$OCI_TITLE" \
+            --annotation "index:org.opencontainers.image.url=$OCI_URL" \
+            $TAG_ARGS \
+            $(printf '%s@sha256:%s ' "$GHCR_IMAGE" *)
+
+      - name: "Inspect images"
+        run: |
+          docker buildx imagetools inspect "$GHCR_IMAGE:latest"
+          docker buildx imagetools inspect "$GHCR_IMAGE:$OCI_VERSION"
```
ZeldaZach: `$OCI_VERSION` is never defined (the `env:` block only sets `GHCR_IMAGE`/`OCI_DESCRIPTION`/`OCI_TITLE`/`OCI_URL`), so this expands to `ghcr.io/cockatrice/servatrice:` and fails, making the `index` job report failure even though the manifest was already published. Use a metadata output, e.g. `"$GHCR_IMAGE:${{ steps.metadata.outputs.version }}"`, or drop this line.

.github/workflows/docker-release.yml
```diff
@@ -61,18 +67,108 @@ jobs:
...
+        if: steps.login.outcome == 'success'
+        env:
+          DIGEST: ${{ steps.build.outputs.digest }}
+        run: |
+          mkdir -p $RUNNER_TEMP/digests
+          touch "$RUNNER_TEMP/digests/${DIGEST#sha256:}"
+
+      - name: "Upload digest"
+        if: steps.login.outcome == 'success'
+        uses: actions/upload-artifact@v7
```
ZeldaZach, line 112: Version mismatch: this uses `upload-artifact@v7` while the download step uses `download-artifact@v4`. Looks like a typo, and `v7` may not resolve at all (v4 is the current stable major for both). Align them on `@v4`.

ZeldaZach (commented):
The native per-arch matrix plus by-digest merge is the right direction and closely follows the official Docker "distribute build across runners" pattern, so the ~10x speedup is well worth having. That said, several load-bearing details diverge from that pattern, and as written the release publish path would fail before it publishes anything. I verified each item against the head file and reproduced the `printf` behavior locally.

Status: **request changes**. Blocking items (release path can't succeed):
- The by-digest build step omits `name=${{ env.GHCR_IMAGE }}` from `outputs:`, so buildx has no repository to push to.
- The digest-merge `printf` corrupts every reference after the first.
- `runs-on: ubuntu-slim` isn't a GitHub-hosted runner label, so `index` never gets a runner.
- The upload artifact name `digest-linux/amd64` contains an illegal `/`.
- `$RUNNER_TEMP` inside `with: path:` inputs isn't expanded by the runner, which also makes the download dir and the merge `working-directory` disagree.

Lower severity: `$OCI_VERSION` in the inspect step is undefined, and `upload-artifact@v7` vs `download-artifact@v4` looks like a typo. Nice touches worth calling out: cache scopes are split per arch (`servatrice-x86` / `servatrice-arm`) so no cross-arch thrash; `fail-fast: false` with `index` gated on full `build` success means a single-arch failure won't publish a partial manifest; and `latest`/tags are restricted to stable releases so a prerelease can't move `latest`. Details inline.

.github/workflows/docker-release.yml
```diff
@@ -1,8 +1,9 @@
-name: Build Docker Image
+name: Build Docker
 
 permissions:
-  contents: read
-  packages: write
+  contents: read    # needed to checkout repo
+  id-token: write    # needed for signing attestations (SBOM & provenance) with GitHub OIDC
```
ZeldaZach: `id-token: write` grants OIDC token minting. `build-push-action`'s `provenance`/`sbom` attestations are attached to the image in the registry and aren't OIDC-signed by default, so nothing here consumes the OIDC token and this permission is likely unnecessary. If you're not adding `actions/attest-*` or cosign keyless signing, drop it to keep least privilege (and the comment accurate).

Dockerfile
```diff
@@ -3,35 +3,45 @@ FROM ubuntu:26.04 AS build
...
 
 WORKDIR /src
 COPY . .
-RUN mkdir build && cd build && \
-    cmake .. -DWITH_SERVER=1 -DWITH_CLIENT=0 -DWITH_ORACLE=0 && \
-    make -j$(nproc) && \
-    make install
+RUN cmake \
+      -S . \
+	  -B build \
```
ZeldaZach: This continuation line is indented with a literal TAB while the surrounding lines use spaces. Cosmetic (CMake doesn't care), but worth normalizing to spaces.

.github/workflows/docker-release.yml
tooomm: v7/v8 are the most recent. Changed both uniformly to v7.

.github/workflows/docker-release.yml
tooomm: `ubuntu-slim` is officially [available since January](https://github.blog/changelog/2026-01-22-1-vcpu-linux-runner-now-generally-available-in-github-actions/), with a public preview before.

Also see the docs: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

I think your AI is a bit outdated here 😇 🤖

.github/workflows/docker-release.yml
tooomm: Yep, this is a leftover. I changed it to ${{ github.ref_name }} to get the specific tag.

.github/workflows/docker-release.yml
tooomm: Jep, your right. Confused this from GitHub docs and artifact attestations

.github/workflows/docker-release.yml
tooomm: Good catch, I made both arrays

BruebachL:
The old flow passed steps.metadata.outputs.labels, which injected org.opencontainers.image.version/revision/source/created alongside the three custom ones. 

The new build step sets only title/description/url, and the index adds only three annotations. 
The PR-body TODO assumes source/licenses/version "are automatically added" but they won't be. BuildKit/GHCR don't synthesize those labels. 

Cheap fix: run metadata-action in build and pass its labels: output, or add source/version to the manual label block.

BruebachL (changes requested):
See comment above.

tooomm:
> The old flow passed steps.metadata.outputs.labels, which injected org.opencontainers.image.version/revision/source/created alongside the three custom ones.
> 
> The new build step sets only title/description/url, and the index adds only three annotations.
> The PR-body TODO assumes source/licenses/version "are automatically added" but they won't be. BuildKit/GHCR don't synthesize those labels.
> 
> Cheap fix: run metadata-action in build and pass its labels: output, or add source/version to the manual label block.

We had this step in the old workflow:
```
      - name: "Docker metadata"
        id: metadata
        uses: docker/metadata-action@v6
        env:
          DOCKER_METADATA_ANNOTATIONS_LEVELS: index    # needed for GHCR
        with:
          annotations: |
            org.opencontainers.image.title=Servatrice
            org.opencontainers.image.url=https://cockatrice.github.io/
            org.opencontainers.image.description=Server for Cockatrice, a cross-platform virtual tabletop for multiplayer card games
          images: |
            ghcr.io/cockatrice/servatrice
          labels: |
            org.opencontainers.image.title=Servatrice
            org.opencontainers.image.url=https://cockatrice.github.io/
            org.opencontainers.image.description=Server for Cockatrice, a cross-platform virtual tabletop for multiplayer card games
```

It added `title`, `url` and `description`.

The stage you are referring to was this one, where we reused them in the same job:
```
      - name: "Build and push Docker image"
        uses: docker/build-push-action@v7
        with:
          annotations: ${{ steps.metadata.outputs.annotations }}
          labels: ${{ steps.metadata.outputs.labels }}
          ...
```

Hmm, I can not find information that `source`/`licenses`/`version` are injected next to custom ones. @BruebachL 
They are listed as option in the [actions repo](https://github.com/docker/metadata-action#outputs), but we did not define them in the old flow. So I assumed they'll be always there.
The TODO is there to check and confirm this, and do adjust according to the outcome.

Unless you know for sure it works in a specific way?

BruebachL:
No, please disregard, I trust your domain knowledge in this case more than mine.

BruebachL (approved):

tooomm:
Appreciate your vigilant and critical review, let's see and confirm how it behaves with the next full release.

If the metadata-action fills certain label/annotation information from the repo automatically (and only ones you want to override/define with custom values need to be listed), I think we are good and `source`, `licence` and `version` should have repo values.

Edit: Both arch runs on master turn out even faster... 4 min uncached / 15s partially cached. 💪

PR #7048, opened 2026-07-20
[PaletteEditor] Ensure directory is writeable, don't fall back to alternate palette

BruebachL:
## Related Ticket(s)
- Supercedes #7008

Palette changes do not save for default themes because they are installed in the system, rather than the user location.

- Ensure directory is writeable, else create a new directory in the user-theme path
- Don't fall back to alternate palette


RickyRister:
Oh nice, `revert to theme default` actually works now

https://github.com/user-attachments/assets/feaedc2d-aeca-43b2-bbaf-69dc5af56ca6

kinda weird that the accent color box doesn't update immediately upon clicking the button though

RickyRister:
Now the color changes upon opening the menu, even when I'm on the default palette

https://github.com/user-attachments/assets/451363ca-0ac0-4585-94ba-ab8e3351406a

BruebachL:
> Now the color changes upon opening the menu, even when I'm on the default palette
> example.mp4

Ah, I didn't block the signal for the palette grid during initial load

cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
```diff
@@ -135,10 +148,14 @@ void PaletteEditorDialog::setupUi()
 
     // Connections
     connect(schemeComboBox, &QComboBox::currentTextChanged, this, &PaletteEditorDialog::onSchemeChanged);
-    connect(quickSetupPanel, &QuickSetupPanel::generateRequested, this, &PaletteEditorDialog::onGenerateFromAccent);
+    autoApplyTimer = new QTimer(this);
+    autoApplyTimer->setSingleShot(true);
+    autoApplyTimer->setInterval(150);
+    connect(autoApplyTimer, &QTimer::timeout, this, &PaletteEditorDialog::onApply);
+    connect(quickSetupPanel, &QuickSetupPanel::valueChanged, this, &PaletteEditorDialog::onGenerateFromAccent);
```
ZeldaZach, line 175: This turns any accent change into a full palette regeneration, including the programmatic `setAccentColor()` inside `seedAccentFromScheme()`. That runs in the constructor (line 49) and on every scheme switch (line 245), so right after `paletteGrid->loadPalette(workingConfig[loadedScheme])` loads the real saved palette, the seed emits `colorChanged`, which chains through `valueChanged` to `onGenerateFromAccent`, overwriting `workingConfig[loadedScheme]` with a palette generated from just the Highlight colour and reloading the grid. `onSave` then persists that regenerated palette, so opening the editor for an existing/custom palette destroys it on screen and on save. (I confirmed `ColorButton::setColor` emits `colorChanged` whenever the colour actually changes, and the seeded Highlight rarely equals the default accent `(20,140,60)`.) Guard the seeding, e.g. `blockSignals(true/false)` around `accentButton->setColor()` in `QuickSetupPanel::setAccentColor()`, so only genuine user input triggers regeneration.

cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
```diff
@@ -22,6 +26,16 @@ PaletteEditorDialog::PaletteEditorDialog(const QString &_themeDirPath, const QSt
     setMinimumSize(740, 220);
     setupUi();
 
+    // Resolve a writable directory for saving. Built-in (Default / Fusion) and
+    // other read-only theme directories must be customised in the user-writable
+    // themes directory; otherwise the write would fail or be lost on upgrade.
+    if (!themeDirPath.isEmpty() && QFileInfo(themeDirPath).isWritable()) {
+        saveDir = themeDirPath;
+    } else {
+        saveDir = QDir(SettingsCache::instance().getThemesPath()).absoluteFilePath(themeName);
```
ZeldaZach, line 51: For a read-only system theme this creates a fresh empty user dir, and `onSave` writes only `palette-*.toml` plus a `theme.cfg` built from `ThemeConfig::fromThemeDir(saveDir)` (empty dir, so `styleName` is lost). But `getAvailableThemes()` gives this partial user copy precedence over the system theme, and `themeChangedSlot()` sets the `theme:` search path to the user copy plus the generic default resources only. Shipped themes that live in the system dir and carry just assets (Leather/Plasma/Fabric/VelvetMarble ship `zones/*.png`) therefore lose their zone backgrounds (and any `style.css`/styleName) the moment a user edits their palette. Either copy the full theme directory into the user path on first save, or keep the system theme dir on the `theme:` search path as a fallback so assets still resolve.

cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
```diff
@@ -22,6 +26,16 @@ PaletteEditorDialog::PaletteEditorDialog(const QString &_themeDirPath, const QSt
     setMinimumSize(740, 220);
     setupUi();
 
+    // Resolve a writable directory for saving. Built-in (Default / Fusion) and
+    // other read-only theme directories must be customised in the user-writable
+    // themes directory; otherwise the write would fail or be lost on upgrade.
+    if (!themeDirPath.isEmpty() && QFileInfo(themeDirPath).isWritable()) {
```
ZeldaZach: `QFileInfo::isWritable()` on a directory is unreliable on the very platform this fix targets. On Windows, directories lack a meaningful read-only bit, so an install under Program Files can report writable and a subsequent write may fail or be silently redirected to the UAC VirtualStore (appears to succeed but lands in a shadow location). Consider probing writability by actually creating and removing a temp file in the directory rather than trusting `isWritable()`.

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -92,10 +92,15 @@ struct PaletteColorInfo
 ThemeManager::ThemeManager(QObject *parent) : QObject(parent)
 {
     defaultStyleName = qApp->style()->objectName();
-    //! \todo Workaround for windows11 style being broken.
-    if (defaultStyleName == "windows11") {
-        defaultStyleName = "windowsvista";
+    // On Windows, Qt may report the legacy "windowsvista" style as the default.
+    // Prefer the modern "windows11" style when it is available; "windowsvista"
+    // remains selectable through the per-theme style setting.
+    if (defaultStyleName.compare("windowsvista", Qt::CaseInsensitive) == 0 &&
```
ZeldaZach: This reverses the prior workaround: the old code forced `windows11` to `windowsvista` because windows11 was flagged broken; this now forces `windowsvista` to `windows11`, making windows11 the default for all Windows users. If the windows11 style is still broken (the reason the workaround existed), this reintroduces that breakage by default. Please confirm windows11 is fixed on the supported Qt versions, and ideally reference the tracking issue in the comment.

cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
```diff
@@ -22,6 +26,16 @@ PaletteEditorDialog::PaletteEditorDialog(const QString &_themeDirPath, const QSt
...
+        saveDir = themeDirPath;
+    } else {
+        saveDir = QDir(SettingsCache::instance().getThemesPath()).absoluteFilePath(themeName);
+        QDir().mkpath(saveDir);
```
ZeldaZach: `mkpath`'s return value is ignored; a creation failure (permissions) only surfaces indirectly when `retranslateUi()` later disables Save via `!isWritable(saveDir)`. Checking the result and logging (or messaging) on failure would make this diagnosable instead of presenting a silently-disabled Save button.

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -298,7 +334,11 @@ void ThemeManager::applyStyleAndPalette(const QString &themeName,
         }
 #endif
     } else {
-        base = qApp->palette();
+        // Use the pristine startup palette rather than qApp->palette(): the
+        // latter may already carry a previously-applied custom (e.g. dark)
+        // palette, which would otherwise persist when switching to a scheme
+        // that supplies no palette of its own.
+        base = defaultPalette;
```
ZeldaZach, line 341: `defaultPalette` is captured once in the constructor before any theme is applied. If the OS light/dark scheme changes while the app is running, this frozen palette goes stale for themes/schemes that supply no custom palette, so a System-scheme theme keeps startup colours instead of following the OS. Minor, but worth a refresh hook via the existing `colorSchemeChanged` connection.

cockatrice/src/interface/palette_editor/palette_editor_dialog.cpp
```diff
@@ -282,12 +303,33 @@ void PaletteEditorDialog::onReset()
 
 void PaletteEditorDialog::onRevertToDefault()
 {
-    PaletteConfig def = PaletteConfig::fromDefault(themeDirPath, loadedScheme);
+    // Delete this scheme's custom palette file so the theme falls back to its
+    // default (or, when it ships none, the application palette).
+    const QString fileName = PaletteConfig::fileName(loadedScheme);
+    QFile::remove(QDir(saveDir).absoluteFilePath(fileName));
+    if (!themeDirPath.isEmpty() && themeDirPath != saveDir) {
+        QFile::remove(QDir(themeDirPath).absoluteFilePath(fileName));
```
ZeldaZach: `fileName` here is `palette-<scheme>.toml`, but shipped themes store their defaults as `palette-default-<scheme>.toml`, so for a read-only system `themeDirPath` this remove is effectively a no-op (and its return is ignored). It only matters if a stale custom file was previously written into `themeDirPath`. Harmless, but the intent reads as if it deletes the shipped default, which it doesn't.

cockatrice/src/interface/palette_editor/quick_setup_panel.h
```diff
@@ -88,7 +86,6 @@ class QuickSetupPanel : public QWidget
     QLabel *labelHigh;
     QSlider *intensitySlider;
     QLabel *intensityPercentageLabel;
-    QPushButton *generateButton;
 };
 
 #endif // COCKATRICE_QUICK_SETUP_PANEL_H
```
ZeldaZach: File ends without a trailing newline (the diff shows `\ No newline at end of file`). Add a final newline to keep clang-format/CI and diffs clean.

ZeldaZach (commented):
The writable-directory resolution and the user-copy precedence are structurally correct, and dropping the opposite-scheme default fallback plus basing on a pristine `defaultPalette` are good improvements. Two regressions should be addressed before merge.

Status: **request changes**.

1. The new accent wiring (`colorChanged` chaining through `valueChanged` to `onGenerateFromAccent`) fires synchronously from `seedAccentFromScheme()`, which runs in the constructor and on every scheme switch. I confirmed `ColorButton::setColor` emits `colorChanged` on any real change, so right after the dialog loads the saved palette it gets regenerated from just the accent colour, and Save then writes that regenerated palette out. The editor can't view or edit an existing palette without destroying it.
2. When a read-only system theme is edited, only the palette files plus a minimal `theme.cfg` (which drops `styleName`) are written to a fresh `user/<theme>` dir. Since `themeChangedSlot` points the `theme:` search path at that user copy plus the generic defaults only, shipped themes that live in the system dir and carry just `zones/*` assets (Leather, Plasma, Fabric, VelvetMarble) lose their zone backgrounds once the partial copy shadows them.

Secondary: `QFileInfo::isWritable()` on a directory is unreliable on Windows (the platform this fix targets), and the `windowsvista` to `windows11` reversal re-enables a style the previous code intentionally avoided. Details inline.

RickyRister:
Accent color box doesn't reset upon clicking `revert to theme default` again

https://github.com/user-attachments/assets/9a9fd8ad-807e-4259-83b6-b255c01527f2

ZeldaZach (approved):

RickyRister (approved):

PR #5962, opened 2025-05-27
🥷 🪟 

tooomm:

PR #7049, opened 2026-07-21
[CI] Use Ninja generator on Windows builds

### LINKED ISSUE #5887: Utilize Ninja with CMake for quicker builds
CMake does use Ninja already to build dependencies if available in the runner (GitHub includes it again since some months), see the logs.

Cockatrice itself is not build with Ninja though.

----

Progress:
 - [x] Linux and macOS ([#5939](https://github.com/Cockatrice/Cockatrice/pull/5939))
 - [ ] Windows

tooomm:
> [!CAUTION]
> - [ ] Merge separate CMake cleanup PRs
> - [ ] Update from master + resolve conflicts in this PR
> - [ ] Polish PR for review with minimum diff
> - [ ] Remove test build proving all still working with VS generator (or keep as test + debug run? overriding centrally set Ninja generator selectively?)

## Related Ticket(s)
- Fixes #5887

- this
- and this

<br>

MSCV development environment setup (vcvars64.bat):
https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line
via Action (TheMrMilchmann/setup-msvc-dev, maintained fork of ilammy/msvc-dev-cmd)

vs. CMake Toolchain, e.g. https://github.com/MarkSchofield/WindowsToolchain

<br>

- ~~There was a bug around `USE_CCACHE` in our config and Windows had ccache enabled due to it being available in PATH from some Perl tooling in the GHA runner, as well as the default being set to ON in CMakeLists:~~
~~`option(USE_CCACHE "Cache the build results with ccache" ON)`~~

  ~~That appeared to be the root case for a lot of headaches when switching to Ninja and trying differnt approaches across various PRs before.~~ --> #7190

- ~~Some checks in `compile.sh` are now more explicit~~ --> #7219

- ~~`cmake --build . --verbose` revealed that many compiler flags are applied multiple times like `/std:c++20 -std:c++20`~~
  --> #7219

CMake gives me a hard time and our config is quite legacy in some areas...
I did extract some updates & improvements not related to Ninja in separate PRs to keep this one focused!

<br>

> Ninja compilation works now 🎉 
> It does also work with VS generator still, see https://github.com/Cockatrice/Cockatrice/commit/ee2feb4c2a6875d9fb4ad805ad304fe0262b5853 ([CI Run with failing pdb upload, fixed later, verbose compilation logs](https://github.com/Cockatrice/Cockatrice/actions/runs/30094419559/job/89485136929?pr=7049)) (the new MSVC setup action works as well, see https://github.com/Cockatrice/Cockatrice/commit/afbf303a4ef85f96c45e35ce5e21ef794d2d9343 ([CI Run, all green, normal logs](https://github.com/Cockatrice/Cockatrice/actions/runs/30105891236/job/89523107914?pr=7049))).

<br>

> Check later again
> ```
> # Can be omitted when using modern CMake config commands below for config and build (chceck BUILD_DIR logic above as well)
> # cmake -S . -B "$BUILD_DIR" "${flags[@]}"
> # cmake --build "$BUILD_DIR" "${buildflags[@]}"
> # Required to update other commands with build folder as well:
> # ctest --build-config "$BUILDTYPE" --test-dir "$BUILD_DIR" --output-on-failure
> # cmake --build "$BUILD_DIR" --target install --config "$BUILDTYPE" (also check `cmake --install .`)
> # cmake --build "$BUILD_DIR" --target package --config "$BUILDTYPE" (remove cd from renaming)
> ```

cmake generator (visual studio generator) and build tool (msbuild) --> ninja both
compiler is unchanged here (msvc)

link former pr's and brunoalr
close #5962

cleanup

Include generator in ccache keys to avoid conflicts and dirty cache (even useful?)
Compiler + version should be added?

#5887


BruebachL:
What's the status on this? Merging this would improve one of our slowest runners.

BruebachL:
## 1. #7049 does not bring the Windows build-time win by itself — #6694 (sccache) does

The Ninja switch alone gives little Windows speedup (brunoalr's closed #6721: "using `ninja` alone there's no real build performance gain, but `sccache` (which DOES improve build times by a lot) it possible with it"). The diff has **no** compiler-cache wiring for MSVC (no `CMAKE_*_COMPILER_LAUNCHER`, no sccache/ccache branch). The payoff lives in #6694. Recommend deciding the #7049↔#6694 relationship explicitly: either stack #6694 on #7049 (base `tooomm_ninja-win`) or merge #7049 then #6694. Flagging so reviewers know the two are complementary, not alternatives.

## 2. Scope creep — split the unrelated cleanup into #7219

The diff mixes several changes that don't belong to "Ninja for Windows":
- `if(DEFINED CMAKE_BUILD_TYPE)` → `if(NOT ...)` (build-type default)
- `set(CMAKE_CXX_EXTENSIONS OFF)`
- `TRICE_IS_64_BIT` via `CMAKE_SIZEOF_VOID_P` instead of `CMAKE_GENERATOR_PLATFORM`
- oracle `libSearchDirs` += vcpkg bin
- tests `TIMEOUT 5` on `dummy_test`
- comment/whitespace polish in `check_schema_version.sh`, CMakeLists

Matches the PR's own "Merge separate CMake cleanup PR" (#7219) checklist item. Splitting keeps the Ninja PR atomic and review-friendly.

## 3. Both Windows generators run per-PR in CI

Two Windows matrix entries (Ninja + `Visual Studio 18 2026`, `-Win10` vs `-Win10_VS`) run on every PR. This is the intended "test build proving all still working with VS generator", but it doubles Windows build time/cost. Suggest gating it to `master`-only (or dropping before merge) — already in the checklist.

## 4. ccache default is back to ON on this branch — check the rebase

`tooomm_ninja-win` still has `option(USE_CCACHE ... ON)` (predates #7190, which flipped it OFF to fix the Windows ccache bug). After rebasing onto current `master`, this hunk will change and the plain `RULE_LAUNCH_COMPILE` block won't subtract cleanly. Note: on Ninja, `RULE_LAUNCH_COMPILE` **is** honored (unlike VS), so if ccache is present it would auto-engage on MSVC — the exact #7190 hazard the compile.sh `-DUSE_CCACHE=0` guard defuses for CI. Consider scoping the ccache block `NOT MSVC`.

## 5. MSVC flag cleanup — sound, one edge to confirm

Removing `/Zc:__cplusplus`, `/std:c++20`, `/permissive-` is consistent: the repo sets `CMAKE_CXX_STANDARD 20` + `CMAKE_CXX_EXTENSIONS OFF`, so CMake emits the standard. `/MP` under `if(CMAKE_GENERATOR MATCHES "^Visual Studio")` (not added for Ninja) is correct. One edge: MSVC's `-std:c++20` recognition in CMake predates 3.21 — with the new 3.16 minimum on older local CMake, confirm CMake passes the standard rather than warning. CI uses recent CMake, so this is local-dev-only.

## 6. PDBs only from the Ninja build

`Upload PDBs` is gated to `matrix.cmake_generator == 'Ninja'`, so the VS test build stops publishing PDBs. Fine once the VS test build is removed; just confirm crash-analyze coverage still comes from Ninja.

Related: the compile-speedups stack (#7237, currently #7232–#7236) has a "Disable ccache auto-engage on Windows (MSVC)" change on #7236 that scopes ccache to non-Windows — aligned with the concern in point 4.

tooomm:
> What's the status on this? Merging this would improve one of our slowest runners.

Windows is the only runner not using compiler caching, that's the reason the build is considerably slower than the others. (The new CodeQL does also not use ccache yet)
Ninja will not change that. 

After we use more uniform tooling, checking compiler caching for Windows is next. Maybe switching to sccache. Ccache should work cross platform though. With our ccache config on the Windows runner fixed, worth trying with proper setup again.

I'm looking into CMake modernisation and some cleanups around compilation and our build scripts in other PRs first. Then coming back to this one with a minimal and clean diff around Ninja.

PR #7050, opened 2026-07-22
[Settings] Split cache_settings monolith into multiple SettingsManager sub-classes

BruebachL:
cache_settings aka SettingsCache was an unwieldy monolithic nightmare.

- Split SettingsCache into new SettingsManager subclasses
- Provide a migration from global.ini to per-settings .ini files
- Move legacy settings migrations to this class
- Fwd declare interfaces in settings_cache and include where needed
- New tests for SettingsDefaults, SettingsManager functionality and SettingsMigration


libcockatrice_settings/libcockatrice/settings/cards_display_settings.cpp
```diff
@@ -0,0 +1,193 @@
+#include "cards_display_settings.h"
+
+CardsDisplaySettings::CardsDisplaySettings(const QString &settingPath, QObject *parent)
+    : SettingsManager(settingPath + "cards.ini", "cards", QString(), parent)
+{
+}
+
+bool CardsDisplaySettings::getDisplayCardNames() const
+{
+    return getValue("displaycardnames", QString(), QString(), true).toBool();
```
RickyRister, line 10: If we're going to migrate all settings, we might as well take this mass rename opportunity to decide on a casing convention for all settings

BruebachL: There are also some settings which aren't quite in their right location yet (t'was a lot) so I'd maybe bundle those in a follow up?

libcockatrice_settings/libcockatrice/settings/cards_display_settings.cpp
```diff
@@ -0,0 +1,193 @@
+#include "cards_display_settings.h"
+
+CardsDisplaySettings::CardsDisplaySettings(const QString &settingPath, QObject *parent)
+    : SettingsManager(settingPath + "cards.ini", "cards", QString(), parent)
```
RickyRister: `cards_display.ini`?

BruebachL: Good catch

cockatrice/src/client/settings/cache_settings.h
```diff
@@ -211,976 +66,91 @@ class SettingsCache : public ICardDatabasePathProvider, public INetworkSettingsP
...
-    void setSoundThemeName(const QString &_soundThemeName);
-    void setIgnoreUnregisteredUsers(QT_STATE_CHANGED_T _ignoreUnregisteredUsers);
-    void setIgnoreUnregisteredUserMessages(QT_STATE_CHANGED_T _ignoreUnregisteredUserMessages);
-    void setIgnoreNonBuddyUserMessages(QT_STATE_CHANGED_T _ignoreNonBuddyUserMessages);
-    void setPixmapCacheSize(const int _pixmapCacheSize);
-    void setCardImageCacheMethod(CardPictureLoaderCacheMethod::CacheMethod _cardImageCachingMethod);
-    void setNetworkCacheSizeInMB(const int _networkCacheSize);
-    void setNetworkRedirectCacheTtl(const int _redirectCacheTtl);
-    void setLocalCardImageStorageNamingScheme(
-        const CardPictureLoaderLocalSchemes::NamingScheme _localCardImageStorageNamingScheme);
```
RickyRister, line 1152: how did you handle settings that use custom types that are in client code?

cockatrice/src/client/settings/cache_settings.cpp
```diff
@@ -183,11 +102,18 @@ SettingsCache::SettingsCache()
...
-    // define a dummy context that will be used where needed
-    QString dummy = QT_TRANSLATE_NOOP("i18n", "English");
-
     QString settingsPath = getSettingsPath();
+
     settings = new QSettings(settingsPath + "global.ini", QSettings::IniFormat, this);
+
+    // Migrate from legacy NativeFormat settings to per-class INI files (runs at most once)
+    if (!isPortableBuild && !QFile(settingsPath + "global.ini").exists()) {
+        SettingsMigration::migrateLegacySettings(settingsPath);
```
ZeldaZach, line 111: This is guarded only by `!QFile(global.ini).exists()`, but after this refactor **nothing ever writes `global.ini`** (the `settings` member is now read-only; only `settings->value(...)` remains). For a truly fresh user the native store is empty, so this is a harmless no-op every launch. But for a user upgrading from an old **NativeFormat** install (has `server`/`messages`/`filter_games`/`sets` groups, no `global.ini`), this re-imports the frozen native values into servers.ini/messages.ini/gamefilters.ini/cardDatabase.ini on **every startup**, overwriting any edits the user makes in between (data loss). The `runs at most once` comment (line 109) is inaccurate. Fix: give `migrateLegacySettings` its own persistent completion marker (a sentinel in a per-file ini, or a dedicated marker file) instead of relying on `global.ini` existing.

libcockatrice_settings/libcockatrice/settings/cache_storage_settings.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+int CacheStorageSettings::getCardPictureLoaderCacheMethod() const
+{
+    return getValue("cardPictureLoaderCacheMethod", QString(), QString(), 0).toInt();
+}
+
+int CacheStorageSettings::getLocalCardImageStorageNamingScheme() const
+{
+    return getValue("localCardImageStorageNamingScheme", QString(), QString(), 0).toInt();
```
ZeldaZach: Silent default change. The old monolith defaulted this to `static_cast<int>(NamingScheme::Set_Folder_Name_Set_Collector)`, which is **6** (the enum lists NameOnly=0 ... Set_Folder_Name_Set_Collector=6). Here the default is `0` (NameOnly), so every user who never explicitly set this (not just new installs) now gets a different local card-image naming/folder scheme than before. Restore the default to `static_cast<int>(CardPictureLoaderLocalSchemes::NamingScheme::Set_Folder_Name_Set_Collector)`.

libcockatrice_settings/libcockatrice/settings/settings_migration.cpp
```diff
@@ -0,0 +1,502 @@
...
+    migrateChatSettings(settingsPath, globalIni);
+    migrateCacheStorageSettings(settingsPath, globalIni);
+    migrateUpdatesSettings(settingsPath, globalIni);
+    migratePersonalSettings(settingsPath, globalIni);
+    migrateCardsDisplaySettings(settingsPath, globalIni);
+    migrateInterfaceSettings(settingsPath, globalIni);
+    migratePathsSettings(settingsPath, globalIni);
+    migrateVisualDeckStorageSettings(settingsPath, globalIni);
+
+    if (!QFile::rename(settingsPath + "global.ini", settingsPath + "global.ini.old")) {
```
ZeldaZach: `QFile::rename` fails if the destination already exists (e.g. a leftover/backup `global.ini.old`). On that failure the function returns early **before** the sentinel is written, and the original `global.ini` is left intact, so on the next launch the whole migration re-runs and re-overwrites the per-file inis with the old global values, discarding any changes the user made in between. Make it idempotent: `QFile::remove(settingsPath + "global.ini.old")` before renaming, or write the sentinel independently of the rename succeeding.

libcockatrice_settings/libcockatrice/settings/settings_migration.cpp
```diff
@@ -0,0 +1,502 @@
...
+}
+
+bool SettingsMigration::migrateSettingsFromGlobalIni(const QString &settingsPath)
+{
+    if (!QFile::exists(settingsPath + "global.ini")) {
+        return false;
+    }
+
+    QSettings globalIni(settingsPath + "global.ini", QSettings::IniFormat);
+    if (globalIni.value(MIGRATION_SENTINEL_KEY, false).toBool()) {
```
ZeldaZach, line 488: Downgrade-then-reupgrade data loss. After migration the new `global.ini` contains only the sentinel; real settings live in per-file inis. If the user then runs an **older** build, it reads/writes `global.ini` (seeing defaults, then persisting new values there). On the next new-build launch the sentinel is still present, so this returns false and those values are never migrated into the per-file inis (silently lost). Worth documenting as a known limitation, or re-checking for real (non-sentinel) keys in `global.ini` and re-migrating them even when the sentinel is set.

tests/settings/settings_migration_test.cpp
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    QVariant readFromIni(const QString &fileName, const QString &key) const
+    {
+        QSettings ini(settingsPath + fileName, QSettings::IniFormat);
+        return ini.value(key);
+    }
+};
+
+TEST_F(SettingsMigrationTest, NoGlobalIniDoesNothing)
```
ZeldaZach, line 48: The suite covers the `global.ini` path well but never exercises `migrateLegacySettings(...)`, nor the rename-failure or downgrade paths, which is exactly where the bugs above live. Please add: (1) a test that seeds NativeFormat `server`/`messages`/`filter_games`/`sets` groups, runs `migrateLegacySettings`, then verifies a second run does NOT overwrite subsequent user edits; (2) a test where `global.ini.old` already exists (rename failure) asserting no re-clobber.

libcockatrice_settings/libcockatrice/settings/personal_settings.cpp
```diff
@@ -0,0 +1,173 @@
+#include "personal_settings.h"
+
+PersonalSettings::PersonalSettings(const QString &settingPath, QObject *parent)
+    : SettingsManager(settingPath + "personal.ini", "personal", QString(), parent)
+{
+}
+
+QString PersonalSettings::getLang() const
+{
+    return getValue("lang", QString(), QString(), "English").toString();
```
ZeldaZach: Default changed from the old `settings->value("personal/lang").toString()` (empty string when unset) to `"English"`. Practically harmless for the translator (both fall back to built-in English), but it's a silent behavior change and would break any code that compares `getLang()` against `""` to detect an unset language. Consider defaulting to `QString()` to preserve prior semantics.

libcockatrice_settings/libcockatrice/settings/updates_settings.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+bool UpdatesSettings::getNotifyAboutNewVersion() const
+{
+    return getValue("newversionnotification", QString(), QString(), true).toBool();
+}
+
+int UpdatesSettings::getUpdateReleaseChannelIndex() const
+{
+    return getValue("updatereleasechannel", QString(), QString(), 0).toInt();
```
ZeldaZach, line 58: The old monolith applied version-based defaults inside the read path (custom build: no update check; beta: channel 1). That logic moved to `window_main::startupConfigCheck` (writes on clean-install / version-change), which covers the common cases. But a pre-existing beta/custom user who never explicitly set these and whose stored clientVersion already equals VERSION_STRING will now get the plain `true`/`0` defaults (no branch fires). Minor and niche, but flagging since it changes behavior for that cohort.

libcockatrice_settings/libcockatrice/settings/settings_migration.cpp
```diff
@@ -0,0 +1,502 @@
...
+    const QStringList personalRootKeys = {"personal/lang", "personal/highlightWords"};
+    const QMap<QString, QString> personalKeyMap = {
+        {"theme/name", "personal/themeName"},
+        {"personal/clientid", "personal/clientid"},
+        {"personal/clientversion", "personal/clientversion"},
+        {"personal/keepalive", "personal/keepalive"},
+        {"personal/timeout", "personal/timeout"},
+        {"personal/picturedownload", "personal/picturedownload"},
+        {"personal/showStatusBar", "personal/showStatusBar"},
+        {"game/maxfontsize", "game/maxfontsize"},
```
ZeldaZach, line 157: `game/maxfontsize` is migrated here into personal.ini and ALSO copied into game.ini by `migrateGameSettings` (which copies every `game/*` child key). Only the personal.ini copy is ever read (`PersonalSettings::getMaxFontSize` reads group `game` from personal.ini). Harmless, but leaves stale duplicate data in game.ini; consider excluding `maxfontsize` from the game.ini copy for clarity.

ZeldaZach (commented):
Solid, well-structured refactor. Each settings domain becomes a `*Settings : SettingsManager, I*Provider` class, the interfaces are clean pure-virtual with default destructors, and the per-call `QSettings` factory correctly fixes the thread-safety issue from #6747. The ~150 call-site conversions I sampled (main.cpp, window_main.cpp, sound_engine.cpp, the DB path provider) are faithful and signals are re-homed onto the sub-objects correctly.

Status: **request changes**, because migration has correctness issues. The most important: because nothing writes `global.ini` anymore, the legacy NativeFormat migration guard (`!QFile(global.ini).exists()`) never closes, so `migrateLegacySettings` runs on every launch and re-clobbers servers/messages/gamefilters/cardDatabase for users upgrading from a pre-`global.ini` version (data loss); the "runs at most once" comment is inaccurate. Secondary: `QFile::rename` failure leaves the sentinel unwritten (non-idempotent re-clobber), the downgrade-then-reupgrade path silently drops changes, and `localCardImageStorageNamingScheme`'s default silently changed from `Set_Folder_Name_Set_Collector` (6) to `NameOnly` (0). Tests cover the `global.ini` path well but never exercise `migrateLegacySettings`, rename failure, or downgrade. Please add a persistent marker for the legacy migration, harden the rename, restore the naming-scheme default, and add tests for the legacy/failure paths.

ZeldaZach (approved):

cockatrice/src/client/settings/cache_settings.h
BruebachL: Enums etc. are stored as int and casted back by the caller

PR #7051, opened 2026-07-22
[Card Database] Improve loading times through binary cache

BruebachL:
I was investigating a couple of the reasons why CardDatabase was taking so long to load and stumbled on a few conclusions:

- Way back when Cockatrice still opened on the old deck editor, the database tab live populated with cards, which was a pleasant, albeit not very informative effect. However, the sheer amount of signals that the event loop was getting hammered with due to us parsing 30 000+ cards was detrimental to the performance for honestly, not much gain.
- The Card Database is off-loaded to a separate thread, however, it runs at the same time as the MainWindow construction. This means that it has to contend with a *lot* of heavy Qt startup work for processor time.
- We spend an *abysmal* amount of time opening, reading, closing CardSet options from freshly constructed QSettings
- We spend an *abysmal* amount of time parsing, constructing and reconstructing primitives from the XML tree
- Same for CardProperties and similar, which we can lazily fetch.

- No longer emit per-card signals while loading the database. Emit one signal once when the database has finished loading.
- Move CardDatabase loading to *before* MainWindow construction. With this PR, opening the Card DB drops from 13-18s down to 500-800ms, which is a negligible increased startup time for the benefit.
- Before loading, hash the contents of the directories that are about to be loaded. Look for a (binary) cache file that matches this hash and load it using binary primitive parsing, skipping xml element construction.
- If the hash differs or the cache is not found, load normally and then cache.

## Benchmarks

(Loading is now largely I/O [memory] bound so the addition of cores here doesn't really do much)

### On my big machine (Ryzen 9 5950X, 32 cores)

```
=== Card Database Load Benchmark ===
----------------------------------------
"Scenario 1: Cold load (XML -> cache)"
  Load wall-clock time : 3274 ms

----------------------------------------
"Scenario 2: Warm load (read cache)"
  Load wall-clock time : 535 ms
```

### On my shitty quad core thinkpad

```
=== Card Database Load Benchmark ===
----------------------------------------
"Scenario 1: Cold load (XML -> cache)"
  Load wall-clock time : 4366 
  
  ----------------------------------------
"Scenario 2: Warm load (read cache)"
  Load wall-clock time : 778 ms
```

### DIFF
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -264,9 +264,11 @@ int main(int argc, char *argv[])
     // contention that happens when the load runs alongside window construction.
     // The CardDatabaseModel populates from the already-loaded data in its
     // constructor, so the window appears fully populated with no startup lag.
-    // Note: checkUnknownSets() is deferred until MainWindow::startupConfigCheck()
-    // so that cardDatabaseNewSetsFound / cardDatabaseAllNewSetsEnabled have live
-    // receivers when emitted.
+    // Note: checkUnknownSets() is deferred from the initial load to
+    // MainWindow::startupConfigCheck() so that
+    // cardDatabaseNewSetsFound / cardDatabaseAllNewSetsEnabled have live
+    // receivers when emitted.  Subsequent reloads (e.g. path changes) call
+    // checkUnknownSets() directly from the loader after the first load.
     CardDatabaseManager::getInstance()->loadCardDatabases();
 
     MainWindow ui;
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -258,6 +258,14 @@ int main(int argc, char *argv[])
 
     qCInfo(MainLog) << "Starting main program";
 
+    // Front-load the card database before constructing the main window. The
+    // binary-cache read is cheap when uncontended (~1s); doing it here -- while no
+    // GUI work yet competes for CPU -- avoids the multi-second, GUI-freezing
+    // contention that happens when the load runs alongside window construction.
+    // The CardDatabaseModel populates from the already-loaded data in its
+    // constructor, so the window appears fully populated with no startup lag.
+    CardDatabaseManager::getInstance()->loadCardDatabases();
+
     MainWindow ui;
     if (parser.isSet("connect")) {
         ui.setConnectTo(parser.value("connect"));
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -264,6 +264,8 @@ int main(int argc, char *argv[])
     // contention that happens when the load runs alongside window construction.
     // The CardDatabaseModel populates from the already-loaded data in its
     // constructor, so the window appears fully populated with no startup lag.
+    // Note: unknown-set / failure signals are deferred; MainWindow triggers
+    // checkUnknownSets() in startupConfigCheck() once receivers are connected.
     CardDatabaseManager::getInstance()->loadCardDatabases();
 
     MainWindow ui;
--- a/cockatrice/src/main.cpp
+++ b/cockatrice/src/main.cpp
@@ -264,8 +264,9 @@ int main(int argc, char *argv[])
     // contention that happens when the load runs alongside window construction.
     // The CardDatabaseModel populates from the already-loaded data in its
     // constructor, so the window appears fully populated with no startup lag.
-    // Note: unknown-set / failure signals are deferred; MainWindow triggers
-    // checkUnknownSets() in startupConfigCheck() once receivers are connected.
+    // Note: checkUnknownSets() is deferred until MainWindow::startupConfigCheck()
+    // so that cardDatabaseNewSetsFound / cardDatabaseAllNewSetsEnabled have live
+    // receivers when emitted.
     CardDatabaseManager::getInstance()->loadCardDatabases();
 
     MainWindow ui;
--- a/libcockatrice_card/libcockatrice/card/card_info.h
+++ b/libcockatrice_card/libcockatrice/card/card_info.h
@@ -10,6 +10,7 @@
 #include <QLoggingCategory>
 #include <QMap>
 #include <QMetaType>
+#include <QMutex>
 #include <QSharedPointer>
 #include <QVariant>
 #include <utility>
@@ -69,12 +70,25 @@ private:
      *  @anchor PrivateCardProperties
      */
     ///@{
-    CardInfoPtr smartThis;                         ///< Smart pointer to self for safe cross-references.
-    QString name;                                  ///< Full name of the card.
-    QString simpleName;                            ///< Simplified name for fuzzy matching.
-    QString text;                                  ///< Text description or rules text of the card.
-    bool isToken;                                  ///< Whether this card is a token or not.
-    QVariantHash properties;                       ///< Key-value store of dynamic card properties.
+    CardInfoPtr smartThis; ///< Smart pointer to self for safe cross-references.
+    QString name;          ///< Full name of the card.
+    QString simpleName;    ///< Simplified name for fuzzy matching.
+    QString text;          ///< Text description or rules text of the card.
+    bool isToken;          ///< Whether this card is a token or not.
+    // Properties are stored as a pre-serialized blob (cheap to load) and the
+    // QVariantHash is materialized on first query, so database load avoids
+    // constructing thousands of QVariants per card.
+    mutable QByteArray propertiesBlob;     ///< Serialized properties (load form).
+    mutable QVariantHash propertiesCache;  ///< Materialized properties (query form).
+    mutable bool propertiesLoaded = false; ///< Whether propertiesCache is valid.
+    mutable QMutex propertiesMutex;        ///< Guards lazy materialization.
+
+    /**
+     * @brief Materializes propertiesCache from propertiesBlob if not already done.
+     *        Safe to call from const getters (members are mutable).
+     */
+    void ensurePropertiesLoaded() const;
+
     QList<CardRelation *> relatedCards;            ///< Forward references to related cards.
     QList<CardRelation *> reverseRelatedCards;     ///< Cards that refer back to this card.
     QList<CardRelation *> reverseRelatedCardsToMe; ///< Cards that consider this card as related.
@@ -107,6 +121,38 @@ public:
                       UiAttributes _uiAttributes);
 
     /**
+     * @brief Constructs a CardInfo from a cache snapshot with precomputed derived
+     *        state.
+     *
+     * Used by the binary cache reader to skip recomputing @p _simpleName and
+     * @p _altNames (which otherwise require a Unicode normalization and a full
+     * printing scan). Properties are supplied as a pre-serialized @p _propertiesBlob
+     * so the QVariantHash is not built at load time (it is materialized on first
+     * query).
+     *
+     * @param _name The card name.
+     * @param _text Rules text or description of the card.
+     * @param _isToken Token flag.
+     * @param _propertiesBlob Pre-serialized properties blob (as written by the cache).
+     * @param _relatedCards Forward relationships.
+     * @param _reverseRelatedCards Reverse relationships.
+     * @param _sets Printing information per set.
+     * @param _uiAttributes Attributes that affect display and game logic.
+     * @param _simpleName Precomputed simplified name.
+     * @param _altNames Precomputed alternate names.
+     */
+    explicit CardInfo(const QString &_name,
+                      const QString &_text,
+                      bool _isToken,
+                      QByteArray _propertiesBlob,
+                      const QList<CardRelation *> &_relatedCards,
+                      const QList<CardRelation *> &_reverseRelatedCards,
+                      SetToPrintingsMap _sets,
+                      UiAttributes _uiAttributes,
+                      QString _simpleName,
+                      QSet<QString> _altNames);
+
+    /**
      * @brief Copy constructor for CardInfo.
      *
      * Performs a deep copy of properties, sets, and related card lists.
@@ -115,7 +161,7 @@ public:
      */
     CardInfo(const CardInfo &other)
         : QObject(other.parent()), name(other.name), simpleName(other.simpleName), text(other.text),
-          isToken(other.isToken), properties(other.properties), relatedCards(other.relatedCards),
+          isToken(other.isToken), propertiesBlob(other.propertiesBlob), relatedCards(other.relatedCards),
           reverseRelatedCards(other.reverseRelatedCards), reverseRelatedCardsToMe(other.reverseRelatedCardsToMe),
           setsToPrintings(other.setsToPrintings), uiAttributes(other.uiAttributes), setsNames(other.setsNames),
           altNames(other.altNames)
@@ -153,6 +199,38 @@ public:
                                    const QList<CardRelation *> &_reverseRelatedCards,
                                    SetToPrintingsMap _sets,
                                    UiAttributes _uiAttributes);
+
+    /**
+     * @brief Creates a new instance from a cache snapshot with precomputed
+     *        derived state.
+     *
+     * @param _name Name of the card.
+     * @param _text Rules text or description.
+     * @param _isToken Token flag.
+     * @param _propertiesBlob Pre-serialized properties blob (as written by the cache).
+     * @param _relatedCards Forward relationships.
+     * @param _reverseRelatedCards Reverse relationships.
+     * @param _sets Printing information per set.
+     * @param _uiAttributes Attributes that affect display and game logic.
+     * @param _simpleName Precomputed simplified name.
+     * @param _altNames Precomputed alternate names.
+     * @param _appendToSets When true (default), the card is appended to each of
+     *        its CardSets. Pass false when building cards in parallel so the
+     *        (non-thread-safe) set membership is populated in a later
+     *        single-threaded pass.
+     * @return Shared pointer to the new CardInfo instance.
+     */
+    static CardInfoPtr newInstance(const QString &_name,
+                                   const QString &_text,
+                                   bool _isToken,
+                                   QByteArray _propertiesBlob,
+                                   const QList<CardRelation *> &_relatedCards,
+                                   const QList<CardRelation *> &_reverseRelatedCards,
+                                   SetToPrintingsMap _sets,
+                                   UiAttributes _uiAttributes,
+                                   QString _simpleName,
+                                   QSet<QString> _altNames,
+                                   bool _appendToSets = true);
 
     /**
      * @brief Clones the current CardInfo instance.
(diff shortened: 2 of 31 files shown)


cockatrice/src/main.cpp
```diff
@@ -255,6 +255,14 @@ int main(int argc, char *argv[])
 
     qCInfo(MainLog) << "Starting main program";
 
+    // Front-load the card database before constructing the main window. The
+    // binary-cache read is cheap when uncontended (~1s); doing it here -- while no
+    // GUI work yet competes for CPU -- avoids the multi-second, GUI-freezing
+    // contention that happens when the load runs alongside window construction.
+    // The CardDatabaseModel populates from the already-loaded data in its
+    // constructor, so the window appears fully populated with no startup lag.
+    CardDatabaseManager::getInstance()->loadCardDatabases();
```
ZeldaZach, line 272: The load runs synchronously here, before `MainWindow ui;` (line 266). But MainWindow only connects to `cardDatabaseLoadingFailed`, `cardDatabaseNewSetsFound`, and `cardDatabaseAllNewSetsEnabled` in its constructor (window_main.cpp ~519-524), and these are the only receivers. Any of these emitted during this load (via `swapInDatabaseData` -> `checkUnknownSets`, or `loadingFailed`) is dropped. Concrete regression: a returning user whose card DB is missing/corrupt gets `emit loadingFailed()` here with no listener, so the 'Cockatrice is unable to load the card database, update now?' dialog never appears; they see an empty client with no guidance. The first-run 'Welcome / all sets enabled' and 'new sets found' prompts are lost the same way. Suggestion: keep the front-loaded parse for speed, but drive the failure/unknown-set/welcome UX from MainWindow after it exists (have MainWindow inspect load status / unknown-set state on construction), rather than relying on one-shot signals emitted before any receiver is connected.

libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -0,0 +1,436 @@
...
+            CardInfoPtr card = readCard(in, data.sets);
+            data.cards.insert(card->getName(), card);
+            data.simpleNameCards.insert(card->getSimpleName(), card);
+        }
+
+        for (const CardInfoPtr &card : data.cards) {
+            for (const auto &printings : card->getSets()) {
+                for (const PrintingInfo &printing : printings) {
+                    if (auto set = printing.getSet()) {
+                        set->append(card);
```
ZeldaZach, line 455: This inner loop has no `break`, so a card is appended to its `CardSet` once per *printing*. `CardSet` is a plain `QList<CardInfoPtr>` (no dedup), and every XML/`newInstance` path appends once per *set*. Cards with multiple printings in the same set (basic lands, reprints with several collector numbers) therefore appear multiple times in the set's card list after a warm (cache) load but exactly once after a cold (XML) load: a warm-vs-cold divergence that inflates set counts and duplicates entries in set browsing. Fix: add `break;` after the append to match the XML path (append once per set), or guard with `if (!set->contains(card))`.

libcockatrice_interfaces/libcockatrice/interfaces/interface_card_set_priority_controller.h
```diff
@@ -13,6 +13,22 @@ class ICardSetPriorityController
...
+     * Reading these individually (getSortKey/isEnabled/isKnown) is cheap for the
+     * noop controller but extremely expensive when backed by QSettings, which
+     * re-opens and re-parses the whole .ini on every access. Callers that need
+     * more than one option for the same set (notably CardSet construction during
+     * database load) should use getSetOptions() to avoid that cost.
+     */
+    struct SetOptions
+    {
+        unsigned int sortKey = 0;
+        bool enabled = true;
```
ZeldaZach: This default (and `isKnown = true` on the next line) changes long-standing behavior. `CardDatabaseSettings::getSetOptions` returns `setOptionsCache.value(shortName)`, which for a set absent from cardDatabase.ini yields a default-constructed `SetOptions{enabled=true, isKnown=true}`. The old `isEnabled`/`isKnown` returned `getValue(...).toBool()` on a missing key -> invalid QVariant -> **false**. So sets never seen before now default to enabled *and* known. Effects: `getUnknownSetsNum()` (counts `!getIsKnown()`) returns 0 for genuinely new sets, so `cardDatabaseNewSetsFound` never fires; `getEnabledSetsNum()` is always > 0 on a fresh install, so the first-run 'no set enabled' path is skipped. Net: new sets are silently auto-enabled and the new-set prompt is suppressed. If intended, please call it out and adjust the first-run/new-set flow; otherwise default `enabled`/`isKnown` to `false` to preserve current semantics.

libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -0,0 +1,436 @@
...
+    blobOut << h;
+    out << static_cast<quint32>(blob.size());
+    out.writeRawData(blob.constData(), blob.size());
+}
+
+QByteArray readHashBlob(QDataStream &in)
+{
+    quint32 len = 0;
+    in >> len;
+    QByteArray blob(len, Qt::Uninitialized);
```
ZeldaZach, line 58: `len` is an unvalidated `quint32` read straight from the file. On a corrupt/truncated cache, `len` can be up to ~4 GB (huge allocation / OOM), and `static_cast<int>(len)` in the following `readRawData(blob.data(), static_cast<int>(len))` is negative for `len > INT_MAX` (undefined size). The source-hash header check doesn't protect the body against truncation/bit-rot. Validate `len` against the remaining bytes in the stream before allocating (e.g. reject if `len` exceeds `in.device()->bytesAvailable()`), and bail out on `in.status() != QDataStream::Ok`.

libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -0,0 +1,436 @@
...
+    quint32 formatCount = 0;
+    in >> formatCount;
+    for (quint32 i = 0; i < formatCount; ++i) {
+        FormatRulesPtr format = readFormat(in);
+        data.formats.insert(format->formatName.toLower(), format);
+    }
+
+    qInfo() << "[cache] read + deserialize" << deserializeTimer.elapsed() << "ms for" << cardCount << "cards";
+
+    return file.error() == QFile::NoError;
```
ZeldaZach, line 484: `read()` validates the header (magic/version/hash) but never checks `QDataStream::status()` after deserializing the body, and only inspects `QFile::error()` here. A cache whose header is intact but whose body was truncated (interrupted write, disk-full, corruption) passes the header check; subsequent reads silently return zeros/empties, producing a partial card set that is then swapped in as the live DB. Because the source hash still matches, the broken cache is reloaded on every subsequent start until the source files change. Add `if (in.status() != QDataStream::Ok) { return false; }` before returning so a corrupt cache cleanly falls back to XML parsing.

libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -0,0 +1,436 @@
...
+        format->exceptions.append(ex);
+    }
+    return format;
+}
+
+} // namespace
+
+bool CardDatabaseCache::write(const QString &cachePath, const CardDatabaseData &data, const QByteArray &sourceHash)
+{
+    QFile file(cachePath);
```
ZeldaZach: The cache is written in place. If the process is killed / loses power / hits a full disk mid-write, the result is a file whose header (magic + version + source hash) is valid but whose body is truncated, exactly the case the reader can't detect (see the status-check finding). Write to a temporary file (e.g. `cachePath + ".tmp"`) and `QFile::rename` it into place, or use `QSaveFile`, so a partial write never becomes the live cache.

libcockatrice_card/libcockatrice/card/database/card_database_loader.cpp
```diff
@@ -119,6 +133,48 @@ LoadStatus CardDatabaseLoader::loadCardDatabases()
...
+    QCryptographicHash hash(QCryptographicHash::Sha256);
+    const QStringList inputs = QStringList()
+                               << pathProvider->getCardDatabasePath() << pathProvider->getTokenDatabasePath()
+                               << pathProvider->getSpoilerCardDatabasePath() << collectCustomDatabasePaths();
+    for (const QString &path : inputs) {
+        QFileInfo info(path);
+        if (info.exists()) {
+            hash.addData(path.toUtf8());
+            hash.addData(QByteArray::number(info.size()));
+            hash.addData(QByteArray::number(info.lastModified().toSecsSinceEpoch()));
```
ZeldaZach, line 179: The source hash covers only path + size + mtime of the input files; it doesn't include the parser/app version, and `CACHE_VERSION` must be bumped by hand. If the XML parser changes what it extracts (new property, changed normalization) while the XML files are byte-identical and a developer forgets to bump `CACHE_VERSION`, every existing user silently loads a stale cache built by the old parser. Recommend folding `VERSION_STRING` (or a build id) into the hash or the cache version so any new build invalidates old caches automatically. Related: mtime is only second-resolution and the fields are concatenated without delimiters (path bytes run into size bytes), so a same-second same-size edit won't invalidate and the concatenation is technically ambiguous; add a separator between fields.

libcockatrice_card/libcockatrice/card/database/card_database_loader.h
```diff
@@ -62,16 +63,20 @@ class CardDatabaseLoader : public QObject
 public slots:
     /**
      * @brief Loads all configured card databases.
+     *
+     * The heavy work runs on a dedicated high-priority thread so it is not
```
ZeldaZach: This is now inaccurate: the old `QtConcurrent::run` calls were removed and `main.cpp` calls `loadCardDatabases()` synchronously on the GUI thread before the window exists (the `databaseDataReady` AutoConnection resolves to a direct same-thread call). No dedicated/high-priority thread is created. On a cold load (cache miss, which happens after every card-DB update) the GUI thread blocks for several seconds with no window shown at all, a different tradeoff than the docstring/PR describe. Please correct this comment (and the `doLoadCardDatabases` 'dedicated loader thread' comment around line 123, and the main.cpp comment), and consider whether a splash/progress indication is warranted during a cold load.

libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+    // Formats
+    quint32 formatCount = 0;
+    in >> formatCount;
+    for (quint32 i = 0; i < formatCount; ++i) {
+        FormatRulesPtr format = readFormat(in);
+        data.formats.insert(format->formatName.toLower(), format);
+    }
+
+    qInfo() << "[cache] read + deserialize" << deserializeTimer.elapsed() << "ms for" << cardCount << "cards";
```
ZeldaZach: Uses a raw `qInfo()` with a hand-rolled `[cache]` prefix instead of a `Q_LOGGING_CATEGORY` (`qCInfo(CardDatabaseLoadingLog) << ...`) like the rest of the loading code. Route it through the existing logging category so it can be filtered consistently, or drop it.

libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -0,0 +1,436 @@
...
+    in >> d;
+    return d;
+}
+
+// ---- CardRelation ----------------------------------------------------------
+
+void writeRelation(QDataStream &out, const CardRelation *rel)
+{
+    writeString(out, rel->getName());
+    out << static_cast<quint32>(static_cast<quint32>(rel->getAttachType()));
```
ZeldaZach: Redundant double `static_cast<quint32>`. Collapse to a single cast.

ZeldaZach (commented):
Strong performance work, and the cache header is well-structured: a magic number, a `CACHE_VERSION` kept separate from the source-content hash, and a pinned `QDataStream` version are all the right calls. A few correctness and data-integrity issues should be sorted before merge, though.

Status: **request changes**.
1. Loading now runs synchronously in `main()` *before* `MainWindow` is constructed, so the startup signals only MainWindow listens to (`cardDatabaseLoadingFailed`, `cardDatabaseNewSetsFound`, `cardDatabaseAllNewSetsEnabled`) fire with no receiver and are dropped. A user with a missing/corrupt DB no longer gets the "update your database" prompt.
2. The cache-read set-append loop appends a card once per *printing* instead of once per *set* (the XML path uses `break`), duplicating multi-printing cards (basic lands) in set lists on every warm load.
3. The QSettings refactor flips the default for unstored sets from disabled/unknown to enabled/known, silently auto-enabling new sets and defeating unknown-set detection.
4. The cache reader is fragile against corruption: `readHashBlob` does an unbounded/overflow-prone allocation from an untrusted length, and `read()` never checks `QDataStream::status()`, so a truncated cache that passes the header check silently loads a partial DB (and, since the source hash still matches, keeps reloading it). The write is also non-atomic, which is how you get a header-valid, body-truncated cache in the first place.

Please fix the append duplication and the pre-window signal ordering, confirm/guard the set-default change, and harden the cache reader (bounds + stream-status checks + atomic write). Docstrings still describe a "dedicated high-priority thread" that no longer exists. Details inline.

(Note: I anchored this review against the current head `dcfa7a3` after the recent force-push.)

libcockatrice_card/libcockatrice/card/database/card_database.h
```diff
@@ -53,17 +54,24 @@ class CardDatabase : public QObject
     /** @brief Querier for higher-level card lookups. */
     CardDatabaseQuerier *querier;
 
-private:
+public:
```
RickyRister: nit: just move this method to the public section instead of switching between public and private

libcockatrice_card/libcockatrice/card/database/card_database_cache.h
```diff
@@ -0,0 +1,35 @@
+#ifndef CARDDATABASE_CACHE_H
+#define CARDDATABASE_CACHE_H
+
+#include "card_database_data.h"
+
+#include <QByteArray>
+#include <libcockatrice/interfaces/interface_card_set_priority_controller.h>
+
+class CardDatabaseCache
```
RickyRister: are you sure this is a class? It just has two public static methods and no state.

libcockatrice_card/libcockatrice/card/printing/printing_info.h
```diff
@@ -61,12 +63,22 @@ class PrintingInfo
...
-    QVariantHash properties; ///< Key-value store for variation-specific attributes.
+    CardSetPtr set; ///< The set this variation belongs to.
+
+    // Properties are stored as a pre-serialized blob (cheap to load) and the
+    // QVariantHash is materialized on first query. This avoids constructing
+    // thousands of QVariants per card at database-load time.
+    mutable QByteArray propertiesBlob;     ///< Serialized properties (load form).
+    mutable QVariantHash propertiesCache;  ///< Materialized properties (query form).
+    mutable bool propertiesLoaded = false; ///< Whether propertiesCache is valid.
+    mutable QSharedPointer<QBasicMutex> propertiesMutex =
```
RickyRister, line 78: is there a reason the mutex needs to be in a `QSharedPointer`? It's not even used anywhere outside the class, so there's no need to worry about shared ownership.

Also, what's `QBasicMutex`? I can't find any documentation on that class on the qt site.

libcockatrice_card/libcockatrice/card/card_info.h
```diff
@@ -208,20 +286,31 @@ class CardInfo : public QObject
...
-    [[nodiscard]] QString getProperty(const QString &propertyName) const
+    [[nodiscard]] const QVariantHash &getPropertiesHash() const;
+
+    /**
+     * @brief Stores the pre-serialized properties blob and invalidates the
+     *        materialized cache. Used by the binary cache reader so the
+     *        QVariantHash is not built at load time.
+     * @param _blob The serialized properties (as written by the cache writer).
+     */
+    void setPropertiesBlob(QByteArray _blob) const
```
RickyRister, line 299: why is this setter const?

libcockatrice_card/libcockatrice/card/card_info.h
```diff
@@ -69,12 +70,25 @@ class CardInfo : public QObject
...
-    QVariantHash properties;                       ///< Key-value store of dynamic card properties.
+    CardInfoPtr smartThis; ///< Smart pointer to self for safe cross-references.
+    QString name;          ///< Full name of the card.
+    QString simpleName;    ///< Simplified name for fuzzy matching.
+    QString text;          ///< Text description or rules text of the card.
+    bool isToken;          ///< Whether this card is a token or not.
+    // Properties are stored as a pre-serialized blob (cheap to load) and the
+    // QVariantHash is materialized on first query, so database load avoids
+    // constructing thousands of QVariants per card.
+    mutable QByteArray propertiesBlob;     ///< Serialized properties (load form).
```
RickyRister, line 81: How much memory overhead does this change add? Naively, I would assume the memory usage from the card database would almost double

ZeldaZach (approved):

libcockatrice_card/libcockatrice/card/card_info.h
BruebachL: Not at all! It's actually smaller because QDataStream serialization is more compact than QVariantHash (about 30%-50%).
Before first access, every card has a smaller footprint.
After the access, you are correct, we don't release the QDataStream for now, so technically, we have both QDataStream and QVariantHash in memory.
So all cards gain a 30-50% increase to properties memory usage.
However, let's not forget that properties are a reasonably small part of the Card itself. Other things i.e. name, relationships, sets, etc. all take up memory as well so this increase isn't an increase of the total amount.

Worst case scenario, I'd estimate this to be a 10-20% bump IF you load all cards in the DB with properties.

libcockatrice_card/libcockatrice/card/printing/printing_info.h
BruebachL: It's the lightest-weight non-recursive mutex they offer. It's been in Qt since Qt5.

See https://woboq.com/blog/internals-of-qmutex-in-qt5.html

BruebachL: It's wrapped in a shared pointer because PrintingInfo is a value type (stored in QList<PrintingInfo>) but needs to be copyable (which QBasicMutex is not) but if you wrap it in a QSharedPointer, it becomes copyable.

libcockatrice_card/libcockatrice/card/card_info.h
BruebachL: It modifies only mutable members (propertiesBlob, propertiesLoaded, propertiesCache), so it can be const without issue. all property accessors (getPropertiesHash, getProperty, ensurePropertiesLoaded) are const because they use the same mutable lazy-loading mechanism. Making setPropertiesBlob non-const would be an inconsistency. Technically there's no call site that requires it to be const (it's called on non-const objects in the cache reader), but const correctness says "this doesn't change the logical state". It only sets up the lazy-loading backing store.

PR #7053, opened 2026-07-24
[PictureLoader] Schedule exponential backoff on 429 - Too many request

BruebachL:
… handler and call failed on fail

429 - Too many requests never calls picDownloadFailed, which means gatherer doesn't get used as a fallback.

- First, attempt to politely wait, then fall through to next url.

### DIFF
--- a/cockatrice/src/interface/card_picture_loader/card_picture_loader_worker_work.h
+++ b/cockatrice/src/interface/card_picture_loader/card_picture_loader_worker_work.h
@@ -4,12 +4,15 @@
 #include "card_picture_loader_worker.h"
 #include "card_picture_to_load.h"
 
+#include <QDateTime>
 #include <QLoggingCategory>
 #include <QMutex>
 #include <QNetworkAccessManager>
 #include <QObject>
+#include <QRandomGenerator>
 #include <QThread>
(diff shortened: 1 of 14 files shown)


cockatrice/src/interface/card_picture_loader/card_picture_loader_worker_work.h
```diff
@@ -40,6 +41,9 @@ class CardPictureLoaderWorkerWork : public QObject
 
     CardPictureToLoad cardToDownload; ///< The card and associated URLs to try downloading
 
+    static constexpr int MAX_RETRIES = 5;        ///< Max number of retries on HTTP 429
```
RickyRister: We might want to fall through sooner, since it'll lead to more responsive behavoir in the default usecase. 
If scryfall is rate limiting, then the average user would probably prefer that it falls back to gatherer without having to wait 60 seconds while the card stays blank.

I think the most ideal behavior on 429 would be for the download to go to the next url, and then only go back to the rate-limited url once the backup urls have been tried. But that's a lot harder to implement, so I'm fine with just this change for now.

ZeldaZach (approved):

PR #7054, opened 2026-07-24
Vertical stacking behaviour fix

DawnFire42:
While testing the command zone I found an issue in my vertical stacking overflow behaviour work (PR #6925 ). It had an artifact of an earlier build that isn't needed. The overflow function covers the necessary edge case for zones smaller then card_height, the minimized version of it was an artefact of earlier test builds.

- nothing visible without the command zone, but fixes a small card position issue when minimizing the command zone while both commander and partner tax are active.

### DIFF
--- a/cockatrice/src/game_graphics/zones/select_zone.cpp
+++ b/cockatrice/src/game_graphics/zones/select_zone.cpp
@@ -21,40 +21,32 @@ SelectZone::ZoneLayout SelectZone::computeZoneLayout(const StackLayoutParams &pa
     }
     qreal effectiveOffset = params.desiredOffset;
     if (params.cardCount > 1) {
-        qreal fitOffset;
-        if (params.totalHeight < params.cardHeight && params.minOffset > 0.0) {
-            // Zone is shorter than a card (e.g. minimized). Compress offsets so
-            // every card has at least minOffset pixels of its top visible.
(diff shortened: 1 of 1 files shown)


RickyRister:
I trust you know what you're doing, because I certainly don't understand the layout code at this point.

RickyRister (approved):

DawnFire42:
Yes, this should already be making it simpler.  Is there something specific that made this confusing? Should I add or change comments to describe the function before merging this?

Essentially, for the vertical stacking within zones smaller then card_height I had at first made a simple function for when the zone was minimized. This one wasn't great at first so I changed it to the overflow that we can turn True or false depending on the zone that calls the vertical stacking. Though minimizing the zone bypassed my polished overflow function and kept using the previous one. The difference between the two function wasn't visible until I added both partner and commander tax and multiple cards and minimizing the zone. Now there is only a single function that deals with the overflow and doesn't change regardless of the minimized state of the zone.

RickyRister:
Ah no, there's no issue with the code itself. I'm just haven't touched the layout code so I'm unfamiliar with it.

DawnFire42:
Fair enough, thanks for checking it. Here I'm very confident since I made this code in the first place. I'm committing to fix whatever I touched if there's issues with it and this PR is one of such fixes.

PR #7055, opened 2026-07-25
[SettingsButtonWidget] Fix memory leak from unparented popup

RickyRister:
## Related Ticket(s)
- Fixes bug introduced in #5816

As noted in [this comment](https://github.com/Cockatrice/Cockatrice/pull/7037#discussion_r3635007374), the `SettingsButtonWidget` currently has a memory leak. It doesn't parent the `SettingsPopupWidget` it creates, which causes it to get leaked once the `SettingsButtonWidget` is destroyed.

- Delete the `SettingsPopupWidget` in the `SettingsButtonWidget` destructor.

### DIFF
--- a/cockatrice/src/interface/widgets/quick_settings/settings_button_widget.cpp
+++ b/cockatrice/src/interface/widgets/quick_settings/settings_button_widget.cpp
@@ -6,7 +6,7 @@
 #include <QScreen>
 
 SettingsButtonWidget::SettingsButtonWidget(QWidget *parent)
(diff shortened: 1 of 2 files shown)


ZeldaZach (approved):

BruebachL:
As mentioned in Discord, nullptr parenting might be due to better behavior on certain WMs, my preferred approach would be to introduce a member variable to the button and manually delete any opened popups on destruction.

BruebachL (approved):
LGTM

PR #7057, opened 2026-07-26
[Game] Implement total power tally

RickyRister:
## Related Ticket(s)
- Follow-up to #7036

Being able to see the total power at a glance is very useful.

- Add `Total Power` tally option
  - The tally will display a single row containing the total power of all selected cards
  - uses `CardItem::parsePT` to extract the powers
  - Will return empty if no cards with pt are selected, which will cause it to not display anything

Note that the tally will not automatically update if the p/t changes while the selection is already made. This can be something to fix in the future

https://github.com/user-attachments/assets/570d54ce-7a41-4ba1-8891-38d8870d7e50

### DIFF
--- a/cockatrice/src/game_graphics/tally/stats_tally.cpp
+++ b/cockatrice/src/game_graphics/tally/stats_tally.cpp
@@ -1,0 +1,35 @@
+#include "stats_tally.h"
+
+#include "../board/card_item.h"
+
+#include <QCoreApplication>
+#include <QList>
+
+static int sumPowers(const QList<CardItem *> &cards)
+{
+    // calculate total power;
+    int total = 0;
+    for (auto card : cards) {
+        QVariantList parsed = CardItem::parsePT(card->getPT());
+        if (!parsed.isEmpty()) {
+            int power = parsed.first().toInt(); // toInt will default to 0 if it's not an int
+            total += qMax(power, 0);
+        }
+    }
+    return total;
+}
+
+QList<TallyRow> StatsTally::computeTotalPower(const QList<CardItem *> &cards)
+{
+    // don't bother if none of the cards have pt
+    bool hasPT =
+        std::any_of(cards.cbegin(), cards.cend(), [](const CardItem *card) { return !card->getPT().isEmpty(); });
+    if (!hasPT) {
+        return {};
+    }
+
+    int total = sumPowers(cards);
+
+    QString name = QCoreApplication::translate("StatsTally", "Total Power");
+    return {TallyRow{name, QString::number(total)}};
+}
--- a/cockatrice/src/game_graphics/tally/stats_tally.cpp
+++ b/cockatrice/src/game_graphics/tally/stats_tally.cpp
@@ -4,6 +4,7 @@
 
 #include <QCoreApplication>
 #include <QList>
+#include <algorithm>
 
 static int sumPowers(const QList<CardItem *> &cards)
 {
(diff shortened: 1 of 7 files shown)


ZeldaZach (commented):
Nicely structured — slotting `TotalPower` into the existing `TallyType`/`Tally::compute` dispatch is clean, and the power-summing logic matches the established behavior in `player_actions.cpp` (`qMax(power, 0)`, `toInt()` defaulting `*`/variable power to 0). I checked the settings persistence path too: `TotalPower` is appended as a new enum value (`= 2`) with `MaxValue` bumped, and `Tally::intToType` clamps out-of-range values to `None`, so existing stored `getTallyType()` ints stay valid — no migration needed.

Worth calling out that the two-pass structure in `computeTotalPower` (the `std::any_of` check, then `sumPowers`) is deliberate and correct: it distinguishes "no P/T cards selected → hide" from "P/T cards that sum to 0 → show a `0` row" (e.g. a `*/*` or `0/0` creature). Not a redundant loop.

Everything I flagged is minor/nits — no blocking issues. Left them inline. One non-inline note: `sumPowers` is a near-verbatim copy of the loop in `player_actions.cpp:1442-1450` (same `// calculate total power;` comment and clamp); if you wanted, that call site could reuse a shared helper down the line, but duplicating ~8 lines is fine for now.

The known limitation from the description — the tally not auto-updating when P/T changes while the selection is unchanged — seems acceptable to defer.

cockatrice/src/game_graphics/tally/stats_tally.h
```diff
@@ -0,0 +1,21 @@
+#ifndef COCKATRICE_POWER_TALLY_H
```
ZeldaZach: Header guard is `COCKATRICE_POWER_TALLY_H` but the file is `stats_tally.h` — looks like a leftover from the earlier `power-tally` name. Suggest `COCKATRICE_STATS_TALLY_H` to match the filename.

cockatrice/src/game_graphics/tally/stats_tally.cpp
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+    return total;
+}
+
+QList<TallyRow> StatsTally::computeTotalPower(const QList<CardItem *> &cards)
+{
+    // don't bother if none of the cards have pt
+    bool hasPT =
+        std::any_of(cards.cbegin(), cards.cend(), [](const CardItem *card) { return !card->getPT().isEmpty(); });
```
ZeldaZach, line 27: `std::any_of` needs `<algorithm>`. It likely compiles today via a transitive include, but that's fragile across toolchains — the sibling `subtype_tally.cpp` explicitly `#include <algorithm>` for exactly this reason. Recommend adding it here for consistency.

cockatrice/src/game_graphics/tally/stats_tally.cpp
```diff
@@ -0,0 +1,35 @@
...
+    if (!hasPT) {
+        return {};
+    }
+
+    int total = sumPowers(cards);
+
+    QString name = QCoreApplication::translate("StatsTally", "Total Power");
```
ZeldaZach, line 34: Casing inconsistency: the row label here is "Total Power" (title case) while the menu action in `tally_menu.cpp` is `tr("Total power")` (sentence case). These are two separate user-facing strings — worth aligning on one casing so the menu entry and the displayed row match.

cockatrice/src/game_graphics/tally/stats_tally.cpp
```diff
@@ -0,0 +1,35 @@
...
+    return {TallyRow{name, QString::number(total)}};
+}
```
ZeldaZach: Nit: missing trailing newline at end of file.

ZeldaZach (approved):

PR #7058, opened 2026-07-27
[Game] Remove tally requiring two cards selected

RickyRister:
## Related Ticket(s)
- Follow-up to #7036
- Changes behavior from #6923

The tally not working until at least two cards are selected is unintuitive.
Especially after adding the total power tally, it makes more sense for the tally to be visible as long as there is anything to display.

- Tally will still show even if only one card is selected.

https://github.com/user-attachments/assets/d668a11e-7e9b-4c68-89a3-9ea4d2b9f92d

### DIFF
--- a/cockatrice/src/game_graphics/game_view.cpp
+++ b/cockatrice/src/game_graphics/game_view.cpp
@@ -253,7 +253,7 @@ void GameView::updateTotalSelectionCount(const QSize &viewSize)
     GameScene *gameScene = static_cast<GameScene *>(scene());
     QList<TallyRow> entries = Tally::compute(gameScene->selectedCards(), tallyType);
 
-    if (entries.isEmpty() || count <= 1) {
+    if (entries.isEmpty()) {
         tallyContainer->hide();
         cachedTallyRows.clear();
         return;


ZeldaZach (approved):
Correct and minimal. Traced the full path to confirm the guard change is safe:

- `Tally::compute` → `SubtypeTally::countSubtypes` returns an empty list for an empty selection, so `entries.isEmpty()` on its own still hides the tally when nothing is selected — dropping `count <= 1` can't surface a stray tally at count 0.
- With a single selected card that has subtypes, `entries` is non-empty and the tally now shows, which is the intended behavior.
- `count` is still used just above for `totalCountLabel`, so no dead variable.
- Positioning stays correct: at one card `totalCountLabel` remains hidden (still gated on `count <= 1`), so the tally lands on the bottom-margin branch rather than stacking above a hidden label.

Worth noting the asymmetry this creates — the numeric total-count badge still needs 2+ selections while the subtype tally now appears at 1 — but that reads as intentional: a lone "1" badge is noise, whereas a single card's subtypes are useful. No changes needed.

PR #7060, opened 2026-07-27
[Replay] Refactor: extract replay playback logic into single class

RickyRister:
## Related Ticket(s)
- Follow-up to #7059

- Move the replay playback logic that was spread across `ReplayWidget` and `ReplayTimelineWidget` into `ReplayManager`.
- `ReplayTimelineWidget` now only controls the timeline display
  - updated with signals from `ReplayManager` and updates `ReplayManager` on click using signals 
- `ReplayWidget` is only responsible for managing the interactive widgets on the dock, as well as wiring everything up.
  - `ReplayManager` is created inside the `ReplayWidget` constructor
- `ReplayManager` will take ownership of the `GameReplay` object, meaning that TabGame no longer manages the deletion of that.
  - Update some of the params to take `const GameReplay*` to reinforce that.

### DIFF
--- a/cockatrice/src/interface/widgets/replay/replay_manager.cpp
+++ b/cockatrice/src/interface/widgets/replay/replay_manager.cpp
@@ -1,0 +1,178 @@
+#include "replay_manager.h"
+
+#include "../../../client/settings/cache_settings.h"
+
+#include <QTimer>
+#include <libcockatrice/settings/interface_settings.h>
+
+static constexpr int TIMER_INTERVAL_MS = 200;
+
+static QList<int> createReplayTimeline(const GameReplay *replay)
+{
+    // Create list: event number -> time [ms]
+    unsigned int lastEventTimestamp = 0;
+    const int eventCount = replay->event_list_size();
+
+    QList<int> replayTimeline;
+    for (int i = 0; i < eventCount; ++i) {
+        int nextSecondIndex = i + 1;
+        while (nextSecondIndex < eventCount &&
+               replay->event_list(nextSecondIndex).seconds_elapsed() == lastEventTimestamp) {
+            ++nextSecondIndex;
+        }
+
+        // Distribute simultaneous events evenly across 1 second.
+        const int numberEventsThisSecond = nextSecondIndex - i;
+        for (int k = 0; k < numberEventsThisSecond; ++k) {
+            int eventMs = replay->event_list(i + k).seconds_elapsed() * 1000;
+            int distributionMs = static_cast<int>(static_cast<qreal>(k) / numberEventsThisSecond) * 1000;
+            replayTimeline.append(eventMs + distributionMs);
+        }
+
+        if (nextSecondIndex < eventCount) {
+            lastEventTimestamp = replay->event_list(nextSecondIndex).seconds_elapsed();
+        }
+        i += numberEventsThisSecond - 1;
+    }
+
+    return replayTimeline;
+}
+
+ReplayManager::ReplayManager(QObject *parent, GameReplay *replay)
+    : QObject(parent), replay(replay), replayTimeline(createReplayTimeline(replay))
+{
+    maxTime = replayTimeline.isEmpty() ? 0 : replayTimeline.last();
+
+    replayTimer = new QTimer(this);
+    replayTimer->setInterval(TIMER_INTERVAL_MS);
+    connect(replayTimer, &QTimer::timeout, this, &ReplayManager::replayTimerTimeout);
+
+    rewindBufferingTimer = new QTimer(this);
+    rewindBufferingTimer->setSingleShot(true);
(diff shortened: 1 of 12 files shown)


cockatrice/src/interface/widgets/replay/replay_manager.cpp
```diff
@@ -0,0 +1,178 @@
...
+        while (nextSecondIndex < eventCount &&
+               replay->event_list(nextSecondIndex).seconds_elapsed() == lastEventTimestamp) {
+            ++nextSecondIndex;
+        }
+
+        // Distribute simultaneous events evenly across 1 second.
+        const int numberEventsThisSecond = nextSecondIndex - i;
+        for (int k = 0; k < numberEventsThisSecond; ++k) {
+            int eventMs = replay->event_list(i + k).seconds_elapsed() * 1000;
+            int distributionMs = static_cast<int>(static_cast<qreal>(k) / numberEventsThisSecond) * 1000;
```
ZeldaZach: This always evaluates to `0`. `static_cast<qreal>(k) / numberEventsThisSecond` is in `[0, 1)` for `k` in `[0, numberEventsThisSecond)`, and casting that to `int` **before** multiplying by 1000 truncates it to 0 — so `distributionMs` is always 0 and simultaneous events no longer get spread across the second.

The original (in `ReplayWidget`) computed the multiply in floating point first:
```cpp
(int)((qreal)k / (qreal)numberEventsThisSecond * 1000)
```
Here the parentheses moved, changing the result. Fix by casting after the `* 1000`:
```cpp
int distributionMs = static_cast<int>(static_cast<qreal>(k) / numberEventsThisSecond * 1000);
```

cockatrice/src/interface/widgets/replay/replay_timeline_widget.h
```diff
@@ -18,57 +18,25 @@ class QTimer;
...
-    QList<int> replayTimeline;
     QList<int> histogram;
-    int maxBinValue, maxTime;
-    qreal timeScaleFactor;
-    int currentVisualTime;    // time currently displayed by the timeline
-    int currentProcessedTime; // time that events are currently processed up to. Could differ from visual time due to
-                              // rewind buffering
-    int currentEvent;
+    int maxBinValue = 1;
+    int maxTime = 0;
```
ZeldaZach: This default changed from the old constructor's `maxTime(1)` to `0`. `setTimeline()` only assigns `maxTime` when the timeline is non-empty, and `paintEvent()` computes `w = (width-1) * currentTime / maxTime` (line 62), so an empty replay timeline (a zero-event replay) now divides by zero — a SIGFPE where master was safe (it left `maxTime` at 1). Either default this to `1`, or guard the division in `paintEvent`. (Separately, the all-events-at-second-0 case where `replayTimeline.last() == 0` can still yield `maxTime == 0` in both versions — worth guarding the divide regardless.)

ZeldaZach (commented):
Nice cleanup — pulling all the playback state (timers, event cursor, rewind buffering) into `ReplayManager` and leaving `ReplayTimelineWidget` as a pure display/click widget is a clear separation, and the signal rewiring (`timeChanged`→`setCurrentTime`, `timeClicked`→`setTime`) is equivalent to the old direct calls. I traced the ownership change and it's sound: `replay` is guaranteed non-null on this path (`Replay`'s constructor derefs it via `loadReplay` before `ReplayManager` is ever built), `ReplayManager` is now the sole owner and deletes it in its dtor, and dropping the `delete replayWidget->replay` in `~TabGame` avoids a double-free. Removing the `game->getGame()->loadReplay(replay)` call from `ReplayWidget` is also correct — `Replay`'s constructor already calls `loadReplay`, so that was a redundant second load.

Two issues from the move, one behavioral:

**1. (Medium) The simultaneous-event distribution is now always 0 — a regression.** In the extracted `createReplayTimeline`, the cast placement changed so the sub-second offset is always zero (see inline). Events that share a `seconds_elapsed` value used to be spread across that second; now they all collapse onto the same timestamp and fire in one tick, making playback clumpy.

**2. (Low) `maxTime` default changed 1 → 0, reintroducing a divide-by-zero for empty replays.** See inline. The old constructor initialized `maxTime(1)`, so an empty timeline was safe; now it stays `0` and `paintEvent` divides by it.

Everything else — timer parenting, throttled repaint (moved into `setCurrentTime`'s 1s-resolution check), skip/rewind buffering, event indexing — looks faithful to the original.

cockatrice/src/interface/widgets/replay/replay_timeline_widget.h
RickyRister: I'll add the guard in paintEvent in a separate PR, since that is existing behavior

ZeldaZach (approved):

PR #7063, opened 2026-07-28
[CardInfo] use QString QHash instead of QVariantHash for properties

RickyRister:
## Related Ticket(s)
- Follow-up to #7051

We currently store card properties in a `QVariantHash`. However, we only ever store `QString`s in there. 

As we found out in the previous PR, creating `QVariant`s does have some overhead. There's no reason to use `QVariantHash` when we're only ever storing `QString`s in there.

- For `CardInfo` and `PrintingInfo`, store properties in a `QHash<QString, QString>` instead of a `QVariantHash`
- Fix signatures


libcockatrice_card/libcockatrice/card/database/card_database_cache.cpp
```diff
@@ -34,11 +33,11 @@ QString readString(QDataStream &in)
...
 }
 
-// Stores a QVariantHash as a single pre-serialized blob. The reader keeps the
-// blob as-is and materializes the QVariantHash lazily on first query, which is
+// Stores a QHash<QString, QString> as a single pre-serialized blob. The reader keeps the
+// blob as-is and materializes the QHash<QString, QString> lazily on first query, which is
 // what removes the allocation storm from database load (see card_info.cpp /
 // printing_info.cpp).
-void writeHashBlob(QDataStream &out, const QVariantHash &h)
+void writeHashBlob(QDataStream &out, const QHash<QString, QString> &h)
```
ZeldaZach, line 40: This changes the serialized format of the properties blob: `QDataStream << QVariantHash` writes a type-tagged `QVariant` per value, whereas `QDataStream << QHash<QString, QString>` writes a bare length-prefixed `QString`. The two are not byte-compatible.

But `CACHE_VERSION` (line 20) is still `1`. An existing cache written by a build between #7051 and this PR will pass the `magic/version/sourceHash` check on line 415 and then be deserialized as `QHash<QString, QString>` — reading the old `QVariant` type ids as string lengths, producing garbled/empty card properties (manacost, colors, types, legalities, uuid, …). Because the blob is length-prefixed per card the stream won't hard-fail; it just silently corrupts, and it persists until the source DB hash changes and the cache is regenerated.

Suggest bumping `CACHE_VERSION` to `2` so stale caches are discarded and rebuilt.

libcockatrice_card/libcockatrice/card/printing/printing_info.cpp
```diff
@@ -48,10 +48,10 @@ void PrintingInfo::setProperties(const QVariantHash &_props)
...
 {
-    return getPropertiesHash().value("uuid").toString();
+    return getPropertiesHash().value("uuid");
 }
 
 QString PrintingInfo::getFlavorName() const
 {
-    return getPropertiesHash().value("flavorName").toString();
+    return getPropertiesHash().value("flavorName");
 }
```
ZeldaZach: Nit: missing trailing newline at end of file.

ZeldaZach (commented):
Clean, mechanical refactor — swapping `QVariantHash` for `QHash<QString, QString>` is the right call given properties are only ever `QString`s, and it removes the per-value `QVariant` boxing that #7051 was fighting. Signatures, callers, and the `.toString()` call sites are all updated consistently, and `QHashIterator it(props)` in `combineLegalities` deduces correctly via CTAD.

One thing that needs attention before merge (inline on `writeHashBlob`): this changes the on-disk serialization format of the properties blob, but `CACHE_VERSION` is still `1`. `QVariantHash` and `QHash<QString, QString>` do **not** serialize to the same bytes (the former tags each value with its `QVariant` type id). A cache written by a build between #7051 and this PR will still pass the magic/version/sourceHash check and then be misparsed into garbled/empty properties — persisting until the source DB hash changes. Bumping `CACHE_VERSION` to 2 invalidates those stale caches cleanly.

Minor / non-blocking (not inline since the lines are outside the diff hunks):
- `card_info.cpp:15` — `#include <QVariant>` looks unused now; worth dropping.
- `card_info_comparator.cpp:69` — comment still says "custom property in the QVariantHash"; stale after this change.

ZeldaZach (approved):

Issue #3559, opened 2019-02-05
Doubled oracle db update message on fresh installs


tooomm:
<b>System Information:</b>
Client Version: 2.6.3-beta.4 (2019-02-03)
Client Operating System: Windows 10 (10.0)
Build Architecture: 64-bit
Qt Version: 5.9.7
System Locale: de_DE
__________________________________________________________________________________________

When doing a fresh install you get two messages that tell you to update your card database:

![msg1](https://user-images.githubusercontent.com/9874850/52277266-924ca900-2954-11e9-9a67-b362b653f52e.png)

![msg2](https://user-images.githubusercontent.com/9874850/52277269-937dd600-2954-11e9-97b5-272698c2118f.png)

ZeldaZach:
When i added this feature, it was not combined with the below card database check. These are two separate checks that happen to be firing at the same time. What should be expected behavior here?

ctrlaltca:
Fixed by #3740

PR #7064, opened 2026-07-30
[App] Add onboarding wizard

BruebachL:
## Related Ticket(s)
- This has been a longstanding pain point
- see issues 
- #3738
- #3559
- #3063

New users hitting Cockatrice for the first time are greeted by a barrage of uncoordinated modal popups: an upgrade/downgrade notification, an "unable to load card database" error, an Oracle prompt, and a tip-of-the-day window — all fired in sequence with no clear guidance on what to do or in what order. 
There is no onboarding flow to walk users through the essential first steps: downloading the card database, picking a theme, registering an account, or configuring basic preferences.

- Introduce a multi-page onboarding wizard (FirstRunWizard) that launches automatically on clean install and can be re-run any time via Help → Re-run Onboarding Wizard....
- The wizard guides users through six steps:
* Welcome
* Card Database
* Theme & Look
* Register/Connect
* Preferences
* Finish

- Be clear about what this wizard will do
- Provide advancement indicators at the bottom
- Provide immediately visible changes with the theme manager

## Screenshots
image
image
image
image
image
image


cockatrice/src/interface/widgets/dialogs/dlg_register.cpp
```diff
@@ -352,13 +398,111 @@ DlgRegister::DlgRegister(QWidget *parent) : QDialog(parent)
...
+    connect(previousHostButton, &QRadioButton::toggled, this, &DlgRegister::previousHostSelected);
+    connect(newHostButton, &QRadioButton::toggled, this, &DlgRegister::newHostSelected);
+    connect(previousHosts, &QComboBox::currentTextChanged, this, &DlgRegister::updateDisplayInfo);
+
+    previousHostButton->setChecked(true);
+
+    preRebuildComboBoxList();
+}
+
+DlgRegister::~DlgRegister() = default;
```
RickyRister, line 412: why do we need to add this?

cockatrice/src/interface/widgets/dialogs/dlg_register.cpp
RickyRister, line 1: Oh nice, you redesigned the register dialog.
That said, can you move that to a separate PR? The register dialog is unrelated to the onboarding wizard.

cockatrice/src/interface/widgets/onboarding/pages/theme_setup_page.cpp
```diff
@@ -0,0 +1,201 @@
...
+        return; // theme already has something real to show -- leave it alone
+    }
+
+    // Nothing saved, nothing shipped. Rather than showing flat native Qt
+    // colours during the very first thing a new user sees, seed one from
+    // whatever accent QuickSetupPanel currently holds (its own built-in
+    // default the first time through), and mark it dirty so it's written to
+    // disk if the user moves on without touching the accent controls.
+    PaletteConfig generated =
+        PaletteGenerator::fromAccent(quickSetupPanel->accentColor(), quickSetupPanel->intensity(), scheme);
```
RickyRister, line 173: Theme setup forcibly sets my accent to green and gives me no opportunity to reset it to system default, requiring me to go into the palette settings afterwards if I want to reset it.

https://github.com/user-attachments/assets/9916e850-2a03-4760-912d-63f4142d9726

cockatrice/src/interface/widgets/onboarding/pages/card_database_setup_page.cpp
```diff
@@ -0,0 +1,297 @@
...
+    connect(advancedToggleButton, &QPushButton::toggled, this, &CardDatabaseSetupPage::onToggleAdvanced);
+    connect(restoreDefaultUrlButton, &QPushButton::clicked, this, &CardDatabaseSetupPage::onRestoreDefaultUrl);
+    connect(applyAndRetryButton, &QPushButton::clicked, this, &CardDatabaseSetupPage::onApplyCustomUrl);
+
+    auto *advancedButtonRow = new QHBoxLayout;
+    advancedButtonRow->addWidget(restoreDefaultUrlButton);
+    advancedButtonRow->addStretch();
+    advancedButtonRow->addWidget(applyAndRetryButton);
+
+    auto *advancedLayout = new QVBoxLayout(advancedPanel);
```
RickyRister, line 68: Opening the dropdown does not resize the thing to make it usable

Image

cockatrice/src/interface/widgets/onboarding/pages/theme_setup_page.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <libcockatrice/settings/personal_settings.h>
+
+ThemeSetupPage::ThemeSetupPage(QWidget *parent) : FirstRunWizardPage(parent)
+{
+    themeCombo = new QComboBox(this);
+    schemeCombo = new QComboBox(this);
+    schemeCombo->addItem(tr("Light"), QStringLiteral("Light"));
+    schemeCombo->addItem(tr("Dark"), QStringLiteral("Dark"));
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+    schemeCombo->addItem(tr("Match system"), QStringLiteral("System"));
```
tooomm, line 28: We should use uniform labels and option names in the Wizard and Settings (--> `System`)

tooomm (commented):
I definitely agree to and see the pain point you're addressing here!
The "first run" user journey and setup steps until users can build decks or play a game are not very straight forward or intuitive right now. It's probably the biggest entry barrier for new users and does not tint their initial app experience in the best light.

There are really too many steps and disruptive pop-ups that rather annoy than properly guide a user best in the app's interface and its concepts. 
Your new wizard has a really cool design! However, the solution here packs maybe more than required. If the goal is to get users started conveniently, they probably do not care about 20 settings and tick-boxes to fine tune specific behaviors they do not even have an opinion about yet.
Similar applies to the visual styling options. I feel this is a personal preference for some, but not the need of a basic user who opens the app for the first time.
It's our responsibility to ship good defaults - customization can build onto that whenever a user feels the need.

Usual flow has 2 main steps
1. Install app
  While installation is not uniform across OSes, this is maybe ok for now
  (Windows: NSIS install wizard, macOS: drag & drop install dialog, Linux: not sure right now)
  There would be the option to have a uniform "Qt installer" across operating systems with CPack ([Qt Installer Framework](https://doc.qt.io/qtinstallerframework/ifw-overview.html)).
  If we were using that, card setup could maybe be included there - one single wizard (install + card db) for all users.
It supports e.g. online installers as well.

   Not touched here, but wanted to drop the Qt IFW thought.


2. Setup card game on first start, so they can
   - browse cards & build decks (import card db of choice)
      --> Run oracle
   - play online (facilitate server usage)
      --> Connect (and register if required) to a server

While I'm in favor of the idea, I would like to see a simpler version at the beginning to solve the basic problems listed in 2. relevant for all users and get them setup asap. Overflooding every user with many advanced settings can be counter-productive to the initial goal.

Another approach to improve the user experience for new starters instead (another) wizard/dialog could be to use the existing home screen to advertise required, existing UI options.
That could replace some existing pop-ups with non-disruptive "cards":
Image
(There could also be a "Customize App" widget linking to settings)

Another long standing issue is that the connect and register dialogs are separated currently and are not wired for an intuitive flow.

Please share your thoughts!

<br>

Having said that - Notes on the current PR:
- The "black" Cockatrice icon is our "Servatrice" icon, see the .exe files in an install folder or the created app shortcuts for example (Windows):
  Image
  I would vote to use the known green one in the wizard as well

cmake/FindQtRuntime.cmake
```diff
@@ -19,10 +19,13 @@ if(WITH_CLIENT)
       Multimedia
       Network
       PrintSupport
+      ShaderTools
```
ZeldaZach, line 21: `ShaderTools` is Qt6-only. This list becomes `REQUIRED_QT_COMPONENTS`, which is passed to **both** `find_package(Qt6 …)` and `find_package(Qt5 5.15.2 COMPONENTS ${REQUIRED_QT_COMPONENTS})`. On a Qt5 host that makes `Qt5_FOUND` false and configuration dies at `FATAL_ERROR "No suitable version of Qt was found"`. `Quick`/`QuickWidgets` at least exist in Qt5, but the whole banner is Qt6-only in practice (`qt6_add_shaders`, `import QtQuick`). Suggest moving these three into a Qt6-only path and keeping the Qt5 build on the `GradientFallbackWidget` — or dropping Qt5 support explicitly. CI is entirely Qt6, so this regression won't show up there.

cockatrice/src/interface/widgets/onboarding/pages/card_database_setup_page.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+QString CardDatabaseSetupPage::nextButtonText() const
+{
+    return state == State::NotStarted ? tr("Download") : QString();
+}
+
+bool CardDatabaseSetupPage::handleNextClick()
+{
+    if (state == State::NotStarted) {
+        setState(State::Running);
```
ZeldaZach, line 190: This enters `Running` and emits `updateRequested()`, then relies solely on `cardDatabaseUpdateFinished` (→ `onUpdateFinished`) to leave `Running`. But `MainWindow::createCardUpdateProcess()` early-returns with an "A card database update is already running" message box **without** emitting `cardDatabaseUpdateFinished` when a process is already in flight. In that case the page is stuck in `Running` forever: progress bar spinning, `isComplete()` false so Next is disabled, Retry/Manual hidden (they only show on `Failed`). Skip still works, but Next is dead. Most likely path: a background card update is running (startup "always update", or the wizard was re-launched from Help mid-update) and the user clicks Download. Consider having the page recover if the update can't be started (e.g. treat the "already running" case as failure, or have MainWindow emit `cardDatabaseUpdateFinished(false)` on the early return).

cockatrice/src/interface/widgets/onboarding/pages/theme_setup_page.cpp
```diff
@@ -0,0 +1,201 @@
...
+    // whatever accent QuickSetupPanel currently holds (its own built-in
+    // default the first time through), and mark it dirty so it's written to
+    // disk if the user moves on without touching the accent controls.
+    PaletteConfig generated =
+        PaletteGenerator::fromAccent(quickSetupPanel->accentColor(), quickSetupPanel->intensity(), scheme);
+    themeManager->previewPalette(generated, scheme);
+    paletteDirty = true;
+}
+
+bool ThemeSetupPage::validatePage()
```
ZeldaZach, line 178: `validatePage()` is the only place the palette is written to disk (`commitPalette` + `reloadCurrentTheme`), and the wizard calls it only from `goNext()`. `onGenerateFromAccent`/`maybeAutoGeneratePalette` apply the palette live via `previewPalette()` but never persist it, and `skip()`/`goBack()`/`closeEvent` don't call `validatePage()`. Meanwhile the theme name (`setThemeName`) and scheme (`setColorScheme`) persist immediately. So a user who drags the accent, likes it, and clicks **Skip** keeps the theme+scheme but loses the palette — after restart the theme loads with no palette (the flat look this step is meant to prevent). Consider persisting the dirty palette on skip/close as well, or committing it at the moment it's generated.

cockatrice/src/interface/widgets/onboarding/shader_banner_widget.cpp
```diff
@@ -0,0 +1,195 @@
...
+    clock.stop();
+}
+
+void BannerHost::tick()
+{
+    if (config) {
+        qreal t = elapsed.elapsed() / 1000.0;
+        config->setTime(t);
+        // Visible breathing for the logo: oscillates between 0.0 and 1.0
+        qreal glow = 0.5 + 0.5 * qSin(t * 0.4);
```
ZeldaZach, line 192: `qSin` needs `#include <QtMath>` (or `<qmath.h>`); none of the includes here pull it in directly, so this depends on a transitive include and can fail as "`qSin` was not declared" on stricter toolchains. Please add the include.

ZeldaZach (commented):
Reviewed the whole PR — wizard shell, the six pages, the QQuickWidget/GLSL banner stack, the register-dialog rework, theme-manager/palette-editor refactor, and the build/CI wiring. A lot of this is careful work: the banner's graceful gradient fallback when the scenegraph can't init, the A/B crossfade banks, the shared `commitPalette` helper, and the register dialog faithfully following the existing `DlgConnect` pattern (same signal arity, `_data.at()` indices, `removeServer`). The `exitCardDatabaseUpdate()` null-guard and `setTab` reindex are good catches too.

One thing that stands out positively: the palette-editor `onSave()` refactor correctly commits the loaded scheme *last* so the global `colorScheme` ends up pointing at the active scheme — that ordering is right.

Findings below, most important first. The Qt5 build break is the one I'd consider blocking.

**1. (High) Qt5 configure is broken by `ShaderTools`.** `ShaderTools` is a Qt6-only module, but it's added to `_COCKATRICE_NEEDED`, which folds into the shared `REQUIRED_QT_COMPONENTS` used by *both* the Qt6 and Qt5 `find_package` calls. On any Qt5 host (`FORCE_USE_QT5`, or Qt6 simply absent) `find_package(Qt5 … COMPONENTS … ShaderTools …)` fails → `Qt5_FOUND` false → `FATAL_ERROR "No suitable version of Qt was found"`. The project still has live `Qt5_FOUND` branches, and the entire CI matrix is Qt6, so CI won't catch this. Related: `shader_banner_widget.cpp`/`.h` (QQuickWidget + `import QtQuick`) are in the *unconditional* `cockatrice_SOURCES`, while only the `.qsb`/QML generation is `if(Qt6_FOUND)`-gated — so even after fixing the component list, a Qt5 build compiles Qt-Quick code with no shaders behind it. Please either gate the whole banner behind Qt6 (falling back to `GradientFallbackWidget` on Qt5), or, if Qt5 is being dropped, do that explicitly.

**2. (Medium) Card-database page can wedge in "Running".** See inline. If a card update is already in flight when Download/Retry is pressed, `createCardUpdateProcess()` shows "already running" and returns without ever emitting `cardDatabaseUpdateFinished`, so the page never leaves `Running`: spinner spins, Next stays disabled. Skip is still available as an escape hatch, but Next is dead until the window is reopened. Most reachable via Help → Re-run Onboarding Wizard during a background update.

**3. (Medium) Theme step: accent palette is applied live but only persisted on Next.** See inline. Adjusting the accent marks `paletteDirty` and previews live, but the disk write happens only in `validatePage()` (Next). Skip / Back / closing the window don't persist it, while the theme name and colour scheme *do* persist immediately. Net effect: tweak accent → Skip → after restart the palette is gone and you get the flat native look this feature exists to avoid.

**4. (Low) Just viewing the theme step mutates the live app palette.** `initializePage()` → `maybeAutoGeneratePalette()` calls `previewPalette()` (live, app-wide) and sets `paletteDirty=true` for themes that ship no palette. Combined with #3, opening the theme step and skipping changes the running app's appearance for the session with nothing saved.

**5. (Low) `qSin` without `#include <QtMath>`** in `shader_banner_widget.cpp` — see inline; relies on a transitive include.

I did not find ownership/lifetime, signal-lifetime, or download re-entrancy bugs in the pages (widgets are parented, connects use `this`/singletons, combos block signals during population, failed downloads route to `Failed` with Retry/Manual). The shader config, step indicator, and QML/shader resource paths all check out.

cmake/FindQtRuntime.cmake
BruebachL: We don't ship for Qt5 anymore and it's *been* regressed for a minute now. This is a non-issue.

cockatrice/CMakeLists.txt
```diff
@@ -374,6 +374,29 @@ set(cockatrice_SOURCES
...
+    src/interface/widgets/onboarding/pages/theme_setup_page.cpp
+    src/interface/widgets/onboarding/pages/theme_setup_page.h
+    src/interface/widgets/onboarding/pages/welcome_page.cpp
+    src/interface/widgets/onboarding/pages/welcome_page.h
+    src/interface/widgets/onboarding/shader_banner_widget.cpp
+    src/interface/widgets/onboarding/shader_banner_widget.h
+    src/interface/widgets/onboarding/step_indicator_widget.cpp
+    src/interface/widgets/onboarding/step_indicator_widget.h
+    src/interface/widgets/server/user/user_info_popup.cpp
+    src/interface/widgets/server/user/user_info_popup.h
```
ZeldaZach: `user_info_popup.cpp` / `.h` are already listed further down in `cockatrice_SOURCES` (lines 420-421 on this branch). This looks like a rebase artifact - it is unrelated to onboarding and lands in the middle of the new onboarding block. CMake dedupes so it will not break the build, but please drop these two lines.

cockatrice/src/interface/window_main.cpp
```diff
@@ -977,14 +1002,17 @@ void MainWindow::cardUpdateError(QProcess::ProcessError err)
 
     exitCardDatabaseUpdate();
     QMessageBox::warning(this, tr("Error"), tr("The card database updater exited with an error:\n%1").arg(error));
+    emit cardDatabaseUpdateFinished(false);
```
ZeldaZach, line 1107: Good addition, but `createCardUpdateProcess()` has a *second* early-return that also never emits this signal:

```cpp
if (updaterCmd.isEmpty()) {
    QMessageBox::warning(this, tr("Error"), tr("Unable to run the card database updater: ") + ...);
    exitCardDatabaseUpdate();
    return;   // no emit
}
```

Same consequence as the "already running" branch flagged on card_database_setup_page.cpp: the page has already gone to `State::Running`, so `isComplete()` stays false, Next is disabled, Retry/Manual stay hidden (they only appear on `Failed`), and the progress bar spins forever. This is arguably the more likely of the two - a broken install or a build tree where `oracle` was not built hits it every time, and that user is precisely the one being onboarded.

Emitting `cardDatabaseUpdateFinished(false)` from both early-returns (or from `exitCardDatabaseUpdate()`) would cover all exits from one place.

cockatrice/resources/cockatrice.svg
```diff
@@ -2,20 +2,20 @@
...
    inkscape:version="0.91 r13725"
-   sodipodi:docname="cockatrice.svg">
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   sodipodi:docname="cockatrice.svg"
```
ZeldaZach, line 17: This rewrites the shipped application icon (281 lines removed, 131 added; 18 paths collapsed to 11, and the `url(#filter4089)` reference is gone). It is installed to `${ICONDIR}/hicolor/scalable/apps` and used as the window icon, so it is not a wizard-only asset.

Rendering both versions side by side they look equivalent, so this reads as Inkscape re-save collateral from deriving `cockatrice-logo-white.svg` rather than an intentional change. Please revert `cockatrice.svg` and keep only the new white logo - a brand asset rewrite should not ride along in an onboarding PR.

cockatrice/src/interface/widgets/onboarding/shader_banner_widget.cpp
```diff
@@ -0,0 +1,195 @@
...
+    quickWidget->setResizeMode(QQuickWidget::SizeRootObjectToView);
+
+    config = new BannerShaderConfig(quickWidget->engine());
+    quickWidget->rootContext()->setContextProperty("bannerConfig", config);
+    quickWidget->setSource(QUrl("qrc:/onboarding/qml/BrandBanner.qml"));
+
+    if (quickWidget->status() == QQuickWidget::Error) {
+        activateFallback();
+    } else {
+        connect(quickWidget, &QQuickWidget::sceneGraphError, this, &BannerHost::onSceneGraphFailed);
```
ZeldaZach, line 57: The gradient fallback hangs off `sceneGraphError` plus the `status() == Error` check above, which covers a scenegraph that fails to initialise. It does not cover the Qt Quick **software** adaptation, which is what you actually get on VMs, RDP sessions and blocklisted drivers - the environments the fallback was written for.

I ran a QQuickWidget loading a ShaderEffect under `QT_QUICK_BACKEND=software` on Qt 6.11: status comes back `Ready` and no `sceneGraphError` is emitted. But the software renderer does not implement ShaderEffect, so `usingFallback` stays false and the QQuickWidget is shown while drawing nothing - a blank 150px band instead of the gradient.

Worth checking on an actual software-rendering box. If it reproduces, keying off `QQuickWindow::graphicsApi() == QSGRendererInterface::Software` (or the `QT_QUICK_BACKEND` env var) and calling `activateFallback()` up front would handle it.

cockatrice/src/interface/widgets/dialogs/dlg_register.h
```diff
@@ -48,15 +47,35 @@ class DlgRegister : public QDialog
...
+    HandlePublicServers *hps;
+
     QLabel *infoLabel, *hostLabel, *portLabel, *playernameLabel, *passwordLabel, *passwordConfirmationLabel,
         *emailLabel, *emailConfirmationLabel, *countryLabel, *realnameLabel;
     QLineEdit *hostEdit, *portEdit, *playernameEdit, *passwordEdit, *passwordConfirmationEdit, *emailEdit,
         *emailConfirmationEdit, *realnameEdit;
     QComboBox *countryEdit;
+
+    QMap<QString, std::pair<QString, UserConnection_Information>> savedHostList;
+    const QString placeHolderText = QStringLiteral("Downloading...");
```
ZeldaZach: This is copied from `DlgConnect`, where it is `tr("Downloading...")` - the `QStringLiteral` here drops the string from translation, and it is user-visible in the combo.

More broadly: the whole server-picker block (radio buttons, `downloadThePublicServers`, `preRebuildComboBoxList`, `rebuildComboBoxList`, `updateDisplayInfo`, `actRemoveSavedServer`) is a near-verbatim ~120 line copy of DlgConnect's. Two copies will drift. If the register-dialog rework stays (RickyRister asked for it as a separate PR), it would be worth pulling that into a shared server-picker widget instead.

ZeldaZach (commented):
Second pass, focused on things not already covered above. The wizard shell itself reads well and the page abstraction is the right shape. Five items below; the CMake duplicate and the svg churn are quick, the updater one is a second instance of the stuck-in-Running problem already flagged on card_database_setup_page.cpp.

On the QQuickWidget dependency generally: I checked whether the qml import tree needs to be added to the macOS/Windows install rules (they use an explicit plugin allowlist plus fixup_bundle, and neither covers Qt's qml/ directory). It does not - I built a minimal QQuickWidget + ShaderEffect app against Qt 6.11 with Qml2Imports pointed at an empty directory and it still loads Ready, because QtQuick's qmldir is embedded in the Qt6Quick library itself. So no packaging change is needed there. brand_banner.frag also compiles clean through qsb for GLSL 100es/120/150, HLSL 5.0 and MSL 1.2.

cockatrice/src/interface/widgets/onboarding/shader_banner_widget.cpp
BruebachL: image

cockatrice/resources/cockatrice.svg
BruebachL: I simply deleted some unused and hidden layers. I can separate this out, if it's truly necessary to get your approval but I would mark this as "really not an issue".

ZeldaZach (approved):

PR #7065, opened 2026-07-30
[Settings][Dialog] Implement search for settings by text, description, tooltip, etc.

BruebachL:
Implement a settings search feature for the settings dialog, enabling users to quickly find any setting across all configuration pages using a unified search bar.

- Search bar (Ctrl+F): Added to the top of the settings dialog. Filters all settings across every page as users type, with results ranked by relevance score (label prefix match > label contains > group title match > full-text match).
- Search results view: Replaces the tabbed view when searching, displaying each result with a breadcrumb (Page > Group) and the setting label. Hover/selection styling included.
- Navigation: Activate search result by clicking or pressing Enter — switches back to tab mode, navigates to the correct page, scrolls to the setting widget, and flashes it with a subtle highlight animation. Escape exits search mode. Arrow keys navigate results.
- Replaced the QListWidget-based icon tabs with styled QToolButton tabs (auto-exclusive, with active indicator border). Ctrl+Tab/Ctrl+Shift+Tab for next/previous tab.
- Added a "Reset to Defaults" button per page that resets all settings on the current page and rebuilds the search index.
- AbstractSettingsPage::autoDetectSearchEntries() introspects QGroupBox children and their labels/checkboxes to build the search index automatically, so new settings pages get search support with no extra code.
- Keyboard shortcuts: Ctrl+F focuses search, Ctrl+Tab/Ctrl+Shift+Tab cycles tabs, Escape exits search.

### DIFF
--- a/cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
@@ -1,3 +1,8 @@
+/**
+ * @file dlg_settings.cpp
+ * @brief Implementation of the main settings dialog
+ * @ingroup Dialogs
+ */
 #include "dlg_settings.h"
 
 #include "../../../client/settings/cache_settings.h"
@@ -6,28 +11,46 @@
 #include "../settings_page/deck_editor_settings_page.h"
 #include "../settings_page/general_settings_page.h"
 #include "../settings_page/messages_settings_page.h"
+#include "../settings_page/settings_search_delegate.h"
+#include "../settings_page/settings_search_model.h"
 #include "../settings_page/shortcut_settings_page.h"
 #include "../settings_page/sound_settings_page.h"
 #include "../settings_page/storage_settings_page.h"
 #include "../settings_page/user_interface_settings_page.h"
+#include "../utility/custom_line_edit.h"
 #include "libcockatrice/card/database/card_database_loader.h"
 #include "libcockatrice/card/database/card_database_manager.h"
 
+#include <QAbstractItemView>
 #include <QCloseEvent>
-#include <QDialogButtonBox>
 #include <QDir>
+#include <QFrame>
+#include <QGraphicsOpacityEffect>
 #include <QGuiApplication>
-#include <QListWidget>
+#include <QHBoxLayout>
+#include <QKeyEvent>
+#include <QListView>
 #include <QMessageBox>
+#include <QPropertyAnimation>
+#include <QPushButton>
 #include <QScreen>
 #include <QScrollArea>
 #include <QScrollBar>
+#include <QSequentialAnimationGroup>
+#include <QShortcut>
+#include <QStackedLayout>
 #include <QStackedWidget>
+#include <QToolButton>
 #include <QVBoxLayout>
 #include <libcockatrice/settings/paths_settings.h>
 #include <libcockatrice/settings/personal_settings.h>
 
-static QScrollArea *makeScrollable(QWidget *widget)
+/**
+ * @brief Wraps a widget in a scroll area for long settings pages
+ * @param widget The widget to wrap
+ * @return The scroll area containing the widget
+ */
+static QScrollArea *makeScrollableModern(QWidget *widget)
 {
     widget->setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Maximum);
 
@@ -40,46 +63,14 @@ static QScrollArea *makeScrollable(QWidget *widget)
     return scrollArea;
 }
 
-DlgSettings::DlgSettings(QWidget *parent) : QDialog(parent)
+DlgSettings::DlgSettings(QWidget *parent) : QDialog(parent), currentTabIndex(0), searchActive(false)
 {
     auto rec = QGuiApplication::primaryScreen()->availableGeometry();
-    this->setMinimumSize(qMin(700, rec.width()), qMin(700, rec.height()));
+    setMinimumSize(qMin(750, rec.width()), qMin(700, rec.height()));
 
     connect(&SettingsCache::instance().personal(), &PersonalSettings::langChanged, this, &DlgSettings::updateLanguage);
 
-    contentsWidget = new QListWidget;
-    contentsWidget->setViewMode(QListView::IconMode);
-    contentsWidget->setIconSize(QSize(58, 50));
-    contentsWidget->setMovement(QListView::Static);
-    contentsWidget->setMinimumHeight(85);
-    contentsWidget->setMaximumHeight(85);
-    contentsWidget->setSpacing(5);
-
-    pagesWidget = new QStackedWidget;
-    pagesWidget->addWidget(makeScrollable(new GeneralSettingsPage));
-    pagesWidget->addWidget(makeScrollable(new AppearanceSettingsPage));
-    pagesWidget->addWidget(makeScrollable(new UserInterfaceSettingsPage));
-    pagesWidget->addWidget(new DeckEditorSettingsPage);
-    pagesWidget->addWidget(makeScrollable(new StorageSettingsPage));
-    pagesWidget->addWidget(new MessagesSettingsPage);
-    pagesWidget->addWidget(new SoundSettingsPage);
-    pagesWidget->addWidget(new ShortcutSettingsPage);
-
-    createIcons();
-    contentsWidget->setCurrentRow(0);
-
-    auto *vboxLayout = new QVBoxLayout;
-    vboxLayout->addWidget(contentsWidget);
-    vboxLayout->addWidget(pagesWidget);
-
-    auto *buttonBox = new QDialogButtonBox(QDialogButtonBox::Ok);
-    connect(buttonBox, &QDialogButtonBox::accepted, this, &DlgSettings::close);
-
-    auto *mainLayout = new QVBoxLayout;
-    mainLayout->addLayout(vboxLayout);
-    mainLayout->addSpacing(2);
-    mainLayout->addWidget(buttonBox);
-    setLayout(mainLayout);
+    setupUi();
 
     connect(&SettingsCache::instance().personal(), &PersonalSettings::langChanged, this, &DlgSettings::retranslateUi);
     retranslateUi();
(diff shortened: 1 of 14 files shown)


RickyRister:
Some settings are grabbing more surrounding text than they should

Screenshot 2026-07-30 at 12 34 11 PM

Screenshot 2026-07-30 at 12 35 55 PM

RickyRister:
I don't like how much more vertical space the top bar takes up compared to before. Makes the settings page feel much more cramped.

Before:
Screenshot 2026-07-30 at 12 39 37 PM

After:
Screenshot 2026-07-30 at 12 39 21 PM

RickyRister:
the button next to `decks directory` seems to be always highlighted.

Screenshot 2026-07-30 at 12 50 03 PM

cockatrice/src/interface/widgets/settings_page/abstract_settings_page.cpp
```diff
@@ -0,0 +1,145 @@
...
+            // Layout items (nested grids, hboxes, vboxes)
+            if (item->layout()) {
+                for (int j = 0; j < item->layout()->count(); ++j) {
+                    QLayoutItem *nestedItem = item->layout()->itemAt(j);
+                    if (!nestedItem || !nestedItem->widget()) {
+                        continue;
+                    }
+
+                    QWidget *nestedWidget = nestedItem->widget();
+                    QString label;
```
RickyRister: Isn't this the exact same code twice? See if you can extract it into a function that returns an option of `SettingsSearchEntry`

RickyRister:
Search doesn't work for settings that are split into a QLabel plus a textbox/dropdown/spinbox next to it

cockatrice/src/interface/widgets/dialogs/dlg_settings.h
```diff
@@ -25,19 +37,46 @@ class DlgSettings : public QDialog
...
+    QPushButton *resetButton;        ///< Button to reset current page to defaults
+    QPushButton *okButton;           ///< Button to close the dialog
+
+    // Data
+    QList<AbstractSettingsPage *> pages;    ///< All settings page instances
+    SettingsSearchModel *searchModel;       ///< Model for search results
+    SettingsSearchDelegate *searchDelegate; ///< Delegate for search result rendering
+    int currentTabIndex;                    ///< Currently active tab index
+    bool searchActive;                      ///< Whether search mode is active
+    QStringList pageNames;                  ///< Translated page names for breadcrumb display
```
RickyRister: do we need to save this to a field? All usages of `pageNames` in this class just assign to it and use it immediately.

cockatrice/src/interface/widgets/dialogs/dlg_settings.h
```diff
@@ -25,19 +37,46 @@ class DlgSettings : public QDialog
...
+
+    void setupUi();
+    void setupTabBar();
+    void buildSearchIndex();
+    void switchToTabMode();
+    void switchToSearchMode();
+    void navigateToSearchResult(const QModelIndex &index);
+    void setActiveTab(int index);
+    void flashWidget(QWidget *widget);
```
RickyRister: nit: can be static

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -40,112 +63,357 @@ static QScrollArea *makeScrollable(QWidget *widget)
...
+    pages.append(shortcutsPage);
+
+    pagesWidget->addWidget(makeScrollable(generalPage));
+    pagesWidget->addWidget(makeScrollable(appearancePage));
+    pagesWidget->addWidget(makeScrollable(userInterfacePage));
+    pagesWidget->addWidget(makeScrollable(deckEditorPage));
+    pagesWidget->addWidget(makeScrollable(storagePage));
+    pagesWidget->addWidget(makeScrollable(messagesPage));
+    pagesWidget->addWidget(makeScrollable(soundPage));
+    pagesWidget->addWidget(makeScrollable(shortcutsPage));
```
RickyRister: shortcuts settings page isn't taking up the full space that it used to.
We didn't make it scrollable before.

Before:
Image

After:
Image

tooomm:
Very useful addition, I like the idea a lot to navigate around our long list of settings!

Some comments:
- Some searches just do not work:
  image

  image

- It's required to double click results to access them, general practice for such suggestions and search results is that a single click "accepts & applies" them (not only "select")

- Highlighting of the results is not very visible and easy to miss (Windows, default theming)

- Idea: Add the related icon of the settings group to the search results
  image

cockatrice/src/interface/widgets/settings_page/abstract_settings_page.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <QGroupBox>
+#include <QLabel>
+#include <QLayout>
+
+QList<SettingsSearchEntry> AbstractSettingsPage::getSearchEntries()
+{
+    return autoDetectSearchEntries(this, -1);
+}
+
+void AbstractSettingsPage::resetToDefaults()
```
ZeldaZach: This base implementation is empty and no settings page overrides `resetToDefaults()` in this PR, so the "Reset to Defaults" button is non-functional: it pops the confirmation dialog in `onResetDefaultsClicked()` and then does nothing. That's misleading for a destructive-sounding action. Please either implement the per-page overrides or remove the button until the reset logic lands.

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -40,112 +63,357 @@ static QScrollArea *makeScrollable(QWidget *widget)
...
+void DlgSettings::onSearchTextChanged(const QString &text)
+{
+    searchModel->setFilterString(text);
 
-    userInterfaceButton = new QListWidgetItem(contentsWidget);
-    userInterfaceButton->setTextAlignment(Qt::AlignHCenter);
-    userInterfaceButton->setFlags(Qt::ItemIsSelectable | Qt::ItemIsEnabled);
-    userInterfaceButton->setIcon(QPixmap("theme:config/interface"));
+    if (searchModel->isFilterActive() && !text.trimmed().isEmpty()) {
+        if (!searchActive) {
```
ZeldaZach, line 346: The `if (!searchActive)` guard means `switchToSearchMode()` (which selects row 0) only runs on the *first* character typed. On every later keystroke the model is reset via `beginResetModel/endResetModel`, which clears the list view's current index, and it's never re-selected. So after refining the query there's no highlighted result and Enter does nothing until the user presses Down. Consider re-selecting row 0 here after each filter change, e.g. unconditionally set the current index when `searchModel->rowCount() > 0`.

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -6,27 +11,45 @@
 #include "../settings_page/deck_editor_settings_page.h"
 #include "../settings_page/general_settings_page.h"
 #include "../settings_page/messages_settings_page.h"
+#include "../settings_page/settings_search_delegate.h"
+#include "../settings_page/settings_search_model.h"
 #include "../settings_page/shortcut_settings_page.h"
 #include "../settings_page/sound_settings_page.h"
 #include "../settings_page/storage_settings_page.h"
 #include "../settings_page/user_interface_settings_page.h"
+#include "../utility/custom_line_edit.h"
```
ZeldaZach: `searchEdit` is a plain `QLineEdit`, not a `CustomLineEdit`/`LineEditUnfocusable`. This include is only pulling in `<QLineEdit>` transitively. Prefer `#include <QLineEdit>` directly and drop this one.

(Side note: `custom_line_edit.h` documents `LineEditUnfocusable` as the type to use "when there is a risk of conflict between line editor shortcuts and other shortcuts" — which is exactly this dialog's situation with Ctrl+F / Ctrl+Tab / Escape. Might be worth using it here.)

cockatrice/src/interface/widgets/tabs/tab_room.cpp
```diff
@@ -242,7 +242,7 @@ void TabRoom::actClearChat()
 void TabRoom::actOpenChatSettings()
 {
     DlgSettings settings(this);
-    settings.setTab(4);
+    settings.setTab(5);
```
ZeldaZach: Worth calling out: this fixes a real latent bug — on master the page order puts Storage at index 4 and Messages/Chat at index 5, so `setTab(4)` was opening the Storage page from "chat settings", not Chat. `5` is correct now. These raw indices are fragile though; a named enum for the tab/page order would prevent the next reordering from silently breaking this again.

ZeldaZach (commented):
Nice feature — the auto-detection approach (introspecting `QGroupBox` children) so new pages get search for free is a clean design, and the tab-bar rewrite reads well. A few things worth addressing before merge, one of which is a functional gap:

**1. The "Reset to Defaults" button doesn't do anything.** `AbstractSettingsPage::resetToDefaults()` is an empty base implementation and no concrete page overrides it in this PR. So the button shows a scary destructive-sounding confirmation ("reset all settings on this page to their defaults") and then silently does nothing. Either wire up per-page overrides, or drop the button from this PR until the reset logic exists.

**2. Keyboard activation only works on the first character typed.** `switchToSearchMode()` selects row 0, but it's guarded by `if (!searchActive)` so it only runs on the transition into search mode. Every subsequent keystroke calls `setFilterString()` → `beginResetModel/endResetModel`, which clears the view's current index, and row 0 is never re-selected. Result: after refining the query, there's no highlighted result and pressing Enter does nothing until you press Down. Re-select row 0 after each filter rebuild (or in `onSearchTextChanged`).

**Minor / nits** (inline): unused search-model roles + `filteredPageIndices()` are dead code; `fullSearchText` duplicates the lowercased label redundantly; and the two auto-detect branches (direct widget vs. nested layout) are ~40 lines of near-identical copy-paste that would be cleaner as a small helper taking a `QWidget*` and returning an optional entry.

The `setTab(4)`→`setTab(5)` change is correct — on master `setTab(4)` actually opened the Storage page, not Chat, so this fixes a latent bug (see inline note).

RickyRister:
The search does not pick up settings that are greyed out because of requiring another setting to be enabled first. 
Is that intentional?

cockatrice/src/interface/widgets/settings_page/abstract_settings_page.cpp
```diff
@@ -0,0 +1,192 @@
...
+            // Strip accelerator markers (&) for search
+            QString cleanLabel = label;
+            cleanLabel.remove(QLatin1Char('&'));
+
+            QWidget *control = widget;
+            if (auto *labelWidget = qobject_cast<QLabel *>(widget)) {
+                control = controlForLabel(labelWidget, pair.second);
+            }
+
+            SettingsSearchEntry entry;
```
RickyRister: nit: use designated initializer instead of assigning to each field manually.

cockatrice/src/interface/widgets/settings_page/abstract_settings_page.cpp
```diff
@@ -0,0 +1,192 @@
...
+                if (QLayoutItem *next = containingLayout->itemAt(i)) {
+                    if (QWidget *nextWidget = next->widget()) {
+                        return nextWidget;
+                    }
+                }
+            }
+        }
+    }
+
+    return label;
```
RickyRister: Search is still picking up some labels that don't correspond to settings.

Image

I don't think there are any settings labels that don't have the control widget beside it.
Perhaps we should make this function default to returning `nullptr` if a control can't be found, so that we can filter out non-settings labels.

BruebachL:
> The search does not pick up settings that are greyed out because of requiring another setting to be enabled first. Is that intentional?

Yes, this is intended.
I included this because else it was possible to search for e.g. "Home tab background shuffle frequency" when the home tab background was set to "Theme", in which case that settings is not just greyed out but not visible either, which would lead users to end up at a hidden setting.

RickyRister (approved):
Seems like the text below `custom alert words` still gets picked up. But that's a corner case we could fix later.

Image

Image

PR #7068, opened 2026-08-01
CI: Improve vcpkg cache hits

tooomm:
ABI hash comparison is also based on tool versions.

- Use a specific flag (`--x-abi-tools-use-exact-version`) recommended by Microsoft when using vcpkg in CI with GitHub hosted runners.
  It leads to tool version changes being ruled out as source of ABI hash mismatches and vcpkg will manage downloading of uniform tool versions in CI runs to make cached binaries more likely to have the same ABI hash (e.g. changes in the underlaying GitHub hosted runner would still invalidate it).

See https://learn.microsoft.com/en-us/vcpkg/users/binarycaching-troubleshooting#abi-mismatch

### DIFF
--- a/.ci/compile.sh
+++ b/.ci/compile.sh
@@ -159,6 +159,7 @@ if [[ $PACKAGE_TYPE ]]; then
 fi
 if [[ $USE_VCPKG ]]; then
   flags+=("-DUSE_VCPKG=1")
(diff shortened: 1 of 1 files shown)


BruebachL:
Sounds great and if the manufacturer recommends it, sounds even better.

BruebachL (approved):

PR #7069, opened 2026-08-03
[Replay] Implement option to skip empty sections

RickyRister:
## Related Ticket(s)
- Follow-up to #7060

Replays tend to have a lot of empty space, because players will often take time thinking.

It would be convenient if there was a way to automatically skip through the empty space.

Added setting to skip empty sections. The setting checkbox is located in the quick settings widget, below the replay speed setting.

Screenshot 2026-08-02 at 8 33 04 PM

An **empty section** is defined as a time in the replay where there is no **meaningful event** for 0.5 seconds on either side.

A **meaningful event** is any event other than a `PLAYER_PROPERTIES_CHANGED` event. That's because the replay spams these events like once a second, so we need to filter them out if we want the skipping to have any meaningful effect.

When the setting is enabled, the replay manager will check after each timer trigger if the current time is in the middle of an **empty section**, by checking the timestamps of the closest meaningful events on both sides. If the current time is inside an **empty section**, it will skip forward to 0.5 seconds before the next meaningful event, or to the end of the replay if none is found.  

https://github.com/user-attachments/assets/0baa5cd9-d38e-4d5f-aa18-c3b6765ed63d

### DIFF
--- a/cockatrice/src/interface/widgets/replay/replay_manager.cpp
+++ b/cockatrice/src/interface/widgets/replay/replay_manager.cpp
@@ -3,9 +3,11 @@
 #include "../../../client/settings/cache_settings.h"
 
 #include <QTimer>
+#include <libcockatrice/protocol/get_pb_extension.h>
 #include <libcockatrice/settings/interface_settings.h>
 
 static constexpr int TIMER_INTERVAL_MS = 200;
+static constexpr int EMPTY_SECTION_MARGIN_MS = 500;
 
 static QList<int> createReplayTimeline(const GameReplay *replay)
 {
@@ -119,6 +121,10 @@ void ReplayManager::replayTimerTimeout()
     processNewEvents(NORMAL_PLAYBACK);
 
     timeChanged(currentVisualTime);
+
+    if (skipEmptySections) {
+        handleSkipEmptySection();
+    }
 }
 
 /** @brief Processes all unprocessed events up to the current time. */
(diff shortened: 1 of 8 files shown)


cockatrice/src/interface/widgets/replay/replay_manager.cpp
```diff
@@ -149,13 +155,63 @@ void ReplayManager::processNewEvents(PlaybackMode playbackMode)
...
+}
+
+void ReplayManager::handleSkipEmptySection()
+{
+    if (currentEvent == replayTimeline.size()) {
+        return;
+    }
+
+    // find most recent meaningful event
+    int prevEvent = currentEvent;
```
BruebachL: processNewEvents leaves currentEvent pointing at the first unprocessed event, i.e. replayTimeline[currentEvent] >= currentVisualTime

That's the next event, not the last played one.

handleSkipEmptySection then does:

prevEvent = currentEvent and scans backward, but currentEvent is a future event
nextEvent = currentEvent + 1 and scans forward, skipping the future at currentEvent

BruebachL: prevEvent should start at currentEvent - 1 (clamped at 0)
nextEvent should start at currentEvent (inclusive)

BruebachL (approved):

cockatrice/src/interface/widgets/replay/replay_manager.cpp
RickyRister: good catch

PR #7072, opened 2026-08-04
[Server] Fix unauthenticated crash via replay submit code

BruebachL:
cmdReplaySubmitCode dereferenced userInfo without an authentication guard, allowing an unauthenticated connection with a valid replay code to segfault the server. 

- Add the same authState != PasswordRight guard used by all other replay handlers
-  Gate session command dispatch on a pre-auth whitelist so future handlers cannot be reached before login.


RickyRister (approved):
I don't really have a way to test this, but it's *probably* fine?

PR #7073, opened 2026-08-04
[Security] Use a CSPRNG for salts, tokens, and RNG seeding

BruebachL:
Password salts and activation tokens were generated with the global SFMT RNG, which was seeded from a 32-bit timestamp, making registration salts and activation tokens predictable. The game RNG used the same timestamp seed across restarts.

- Add CryptoUtil backed by OpenSSL RAND_bytes and use it for salt/token generation and to seed RNG_SFMT with a 64-bit CSPRNG value in both the client and server. 
- Link libcockatrice_utility against OpenSSL::Crypto.


BruebachL:
Moved to a same-repo stack — superseded by new stacked PRs from Cockatrice/Cockatrice branches.

PR #7074, opened 2026-08-04
[Security] Add challenge-response auth with scrypt verifiers and stop storing plaintext passwords

### LINKED ISSUE #344: "Save password" feature stores in clear text

### LINKED ISSUE #6624: Modernize used hashes for e.g. passwords and replace MD5
**Similar Requests**
Related https://github.com/Cockatrice/Cockatrice/issues/344

- **General hashing**: Switch to `SHA2` which is very quick and collision free?
- **Password hashing**: `Argon2` or `scrypt`
  
@ebbit1q I saw you worked on a related improvement to #344 already in https://github.com/Cockatrice/Cockatrice/pull/4464.
It looks like we use salted hashes now, but no pepper?
Which hashing algorithms are used there for passwords? SHA512?

**Additional Context**
https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

BruebachL:
## Related Ticket(s)
- Fixes #344
- Fixes #6624

Two separate problems, both around how passwords are handled:
- #6624 – weak password storage. 

`Accounts were stored as a hand-rolled hash (1000 iterations of SHA-512 over salt+password) and, depending on server settings, the login flow could end up sending the plaintext password over the wire. 1000 rounds of SHA-512 is well below modern KDF territory: it's fast, not memory-hard, and cheap to brute-force offline on GPUs. The ticket asked for a proper KDF (Argon2 or scrypt, per OWASP).`

- #344 – saved passwords in plaintext. 

`With "Save password" checked, the client wrote the raw password into its QSettings ini file, so anyone with read access to the profile folder could recover it.`

Authentication moves to a challenge-response flow:
- The server issues a fresh random nonce per login; the client answers with HMAC-SHA256 over the nonce, keyed with its password verifier. Neither the password nor the verifier crosses the wire during login, so there's nothing to replay or sniff.
- New accounts, and changed/reset passwords, store a scrypt verifier instead of the old SHA-512 scheme. scrypt is memory-hard and one of the KDFs OWASP recommends; we use the RFC 7914 interactive parameters (N=32768, r=8, p=1).
- Verifiers are self-describing (`$scrypt$<n>$<r>$<p>$<salt>$<verifier>`), and the salt response carries the cost parameters, so a later parameter bump won't invalidate existing accounts.
- Legacy accounts are migrated to scrypt after their first successful login. Servers in strict mode only accept scrypt verifiers.

Fixing #344 for challenge-response-capable servers:
- A saved profile now stores only the derived verifier under the password key, never the plaintext password.
- The connect dialog loads the verifier without ever showing it; the client only uses it to prove knowledge of the key.
- The verifier is persisted only when "Save password" is checked; autoconnect goes through the same path, and the change-password dialog no longer prefills the old-password field from a stored verifier.
- Servers without challenge-response support are unchanged and keep the previous behavior.


BruebachL:
Moved to a same-repo stack — superseded by new stacked PRs from Cockatrice/Cockatrice branches.

PR #7075, opened 2026-08-04
[Security] Add per-address rate limiting for auth endpoints

BruebachL:
There's no rate limiting on certain endpoints, most crucially, auth ones.

Introduce a thread-safe RateLimiter that tracks attempts per key (IP address) within a sliding time window, and wire it into the authentication endpoints:

- Login: failed login attempts from an address are counted; once the configured maximum is exceeded within the window, further logins from that address are rejected with RespTooManyRequests. A successful login clears the failed attempts for that address.
- Registration: implement the previously stubbed tooManyRegistrationAttempts, limiting how many accounts can be created per address per window.
- Forgot-password: throttle both the email-request and the email-challenge paths per address.

New [security] settings with defaults:
  max_login_attempts_per_ip=5 / login_attempt_window_seconds=900 max_registrations_per_ip=2 / registration_window_seconds=3600 max_forgot_password_requests_per_ip=3 / forgot_password_window_seconds=3600

Adds unit tests for the RateLimiter (window limit, over-limit blocking, clearing, per-key independence).

## Review fixes (6 threads)
- `maxAttempts=0` (or a non-positive window) now disables the limiter as documented, instead of rejecting every address
- Keys at their limit are rejected without recording, so the window isn't slid forward by repeated attempts — a lockout lifts after `windowSeconds` of quiet
- Bounded memory: an opportunistic 60s prune erases keys whose attempts have fully aged out
- Login is throttled before the work: a new `Server::isLoginRateLimited()` gate skips the DB round trip and password verification for locked-out addresses
- `recordAttemptAt()`/`isBlockedAt()` time seams make the sliding-window behavior deterministically testable; tests rewritten for the real semantics

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
@@ -527,10 +527,14 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
             return Response::RespUserIsBanned;
         }
         case NotLoggedIn:
+            if (server->recordFailedLogin(getAddress())) {
+                return Response::RespTooManyRequests;
+            }
             return Response::RespWrongPassword;
         case WouldOverwriteOldSession:
             return Response::RespWouldOverwriteOldSession;
         case UsernameInvalid: {
+            server->recordFailedLogin(getAddress());
             auto *re = new Response_Login;
             re->set_denied_reason_str(reasonStr.toStdString());
             rc.setResponseExtension(re);
@@ -541,8 +545,10 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
         case ClientIdRequired:
             return Response::RespClientIdRequired;
         case UserIsInactive:
+            server->recordFailedLogin(getAddress());
             return Response::RespAccountNotActivated;
         default:
+            server->clearFailedLogins(getAddress());
             authState = res;
             usingRealPassword = needsHash;
     }
--- a/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
@@ -514,6 +514,11 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
     QString reasonStr;
     int banSecondsLeft = 0;
     QString connectionType = getConnectionType();
+    // Throttle before doing any work: a locked-out address must not get a full
+    // database round trip and password verification on every attempt.
+    if (server->isLoginRateLimited(getAddress())) {
+        return Response::RespTooManyRequests;
+    }
     AuthenticationResult res = server->loginUser(this, userName, password, needsHash, reasonStr, banSecondsLeft,
                                                  clientId, clientVersion, connectionType);
     switch (res) {
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -348,6 +348,27 @@ max_users_websocket=500
 ; Maximum number of users that can connect from the same IP address; useful to avoid bots, default is 4
 max_users_per_address=4
 
+; Maximum number of failed login attempts from a single IP address before that
+; address is temporarily locked out.  Default is 5; set to 0 to disable.
+max_login_attempts_per_ip=5
+
+; Length in seconds of the sliding window used for the login lockout. Default is 900 (15 minutes).
+login_attempt_window_seconds=900
+
+; Maximum number of account registrations from a single IP address within the window below.
+; Default is 2; set to 0 to disable.
+max_registrations_per_ip=2
+
+; Length in seconds of the registration window. Default is 3600 (1 hour).
+registration_window_seconds=3600
+
+; Maximum number of forgot-password requests from a single IP address within the window below.
+; Default is 3; set to 0 to disable.
+max_forgot_password_requests_per_ip=3
+
+; Length in seconds of the forgot-password window. Default is 3600 (1 hour).
+forgot_password_window_seconds=3600
+
 ; You may want to allow an unlimited number of users from a trusted source. This setting can contain a
 ; comma-separed list of IP addresses which will allow an unlimited number of connections from each of the
 ; IP addresses listed (ignoring the max_users_per_address). Default is "127.0.0.1,::1"; example: "192.73.233.244,81.4.100.74"
--- a/servatrice/src/ratelimiter.cpp
+++ b/servatrice/src/ratelimiter.cpp
@@ -1,0 +1,42 @@
+#include "ratelimiter.h"
+
+#include <QDateTime>
+
+bool RateLimiter::recordAttempt(const QString &key, int maxAttempts, int windowSeconds)
+{
+    QMutexLocker locker(&mutex);
+    const qint64 now = QDateTime::currentSecsSinceEpoch();
+
+    QList<qint64> &timestamps = attempts[key];
+    timestamps.append(now);
+    while (!timestamps.isEmpty() && timestamps.first() <= now - windowSeconds) {
+        timestamps.removeFirst();
+    }
+
+    return timestamps.size() > maxAttempts;
+}
+
+bool RateLimiter::isBlocked(const QString &key, int maxAttempts, int windowSeconds) const
+{
+    QMutexLocker locker(&mutex);
+    const qint64 now = QDateTime::currentSecsSinceEpoch();
+
+    const auto it = attempts.constFind(key);
+    if (it == attempts.constEnd()) {
+        return false;
+    }
+
+    int count = 0;
+    for (const qint64 &timestamp : it.value()) {
+        if (timestamp > now - windowSeconds) {
+            ++count;
+        }
+    }
+    return count > maxAttempts;
+}
+
+void RateLimiter::clearAttempts(const QString &key)
+{
+    QMutexLocker locker(&mutex);
+    attempts.remove(key);
+}
--- a/servatrice/src/ratelimiter.cpp
+++ b/servatrice/src/ratelimiter.cpp
@@ -4,22 +4,47 @@
 
 bool RateLimiter::recordAttempt(const QString &key, int maxAttempts, int windowSeconds)
 {
-    QMutexLocker locker(&mutex);
-    const qint64 now = QDateTime::currentSecsSinceEpoch();
+    return recordAttemptAt(key, maxAttempts, windowSeconds, QDateTime::currentSecsSinceEpoch());
+}
 
-    QList<qint64> &timestamps = attempts[key];
-    timestamps.append(now);
+bool RateLimiter::recordAttemptAt(const QString &key, int maxAttempts, int windowSeconds, qint64 now)
+{
+    if (maxAttempts <= 0 || windowSeconds <= 0) {
+        return false; // rate limiting disabled for this endpoint
+    }
+
+    QMutexLocker locker(&mutex);
+    pruneLocked(now);
+
+    AttemptHistory &history = attempts[key];
+    history.windowSeconds = windowSeconds;
+    QList<qint64> &timestamps = history.timestamps;
     while (!timestamps.isEmpty() && timestamps.first() <= now - windowSeconds) {
         timestamps.removeFirst();
     }
+    if (timestamps.size() >= maxAttempts) {
+        // Already at the limit: reject without appending so repeated attempts
+        // at the attacker's own pace cannot keep sliding the window forward
+        // and hold the lockout open forever.
+        return true;
+    }
 
-    return timestamps.size() > maxAttempts;
+    timestamps.append(now);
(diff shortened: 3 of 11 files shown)


servatrice/src/ratelimiter.cpp
```diff
@@ -0,0 +1,42 @@
...
+    QMutexLocker locker(&mutex);
+    const qint64 now = QDateTime::currentSecsSinceEpoch();
+
+    QList<qint64> &timestamps = attempts[key];
+    timestamps.append(now);
+    while (!timestamps.isEmpty() && timestamps.first() <= now - windowSeconds) {
+        timestamps.removeFirst();
+    }
+
+    return timestamps.size() > maxAttempts;
```
ZeldaZach: **`maxAttempts = 0` blocks everything instead of disabling the limit.**

`servatrice.ini.example` documents `max_login_attempts_per_ip=0` / `max_registrations_per_ip=0` / `max_forgot_password_requests_per_ip=0` as "set to 0 to disable". With `maxAttempts = 0`, the very first call appends one timestamp and returns `1 > 0` → `true`. So an operator who follows the documentation to turn rate limiting off instead makes login, registration, and password reset fail for every address, permanently.

The PR's own tests encode the wrong behaviour — `RateLimiterTest.ClearAttempts` and `RateLimiterTest.KeysAreIndependent` both `ASSERT_TRUE(limiter.recordAttempt("ip", 0, 60))`.

Suggested fix at the top of both `recordAttempt` and `isBlocked`:

```cpp
if (maxAttempts <= 0 || windowSeconds <= 0) {
    return false;
}
```

and update the two tests to use a non-zero limit.

servatrice/src/ratelimiter.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <QDateTime>
+
+bool RateLimiter::recordAttempt(const QString &key, int maxAttempts, int windowSeconds)
+{
+    QMutexLocker locker(&mutex);
+    const qint64 now = QDateTime::currentSecsSinceEpoch();
+
+    QList<qint64> &timestamps = attempts[key];
+    timestamps.append(now);
```
ZeldaZach: **The window never drains under sustained load, so a lockout can be held open indefinitely.**

The timestamp is appended on *every* call, including calls that are already over the limit. So as long as an attacker keeps sending failed logins faster than `windowSeconds`, the list never empties and the address stays blocked forever — the 15-minute window never actually elapses.

That turns this into an unauthenticated lockout primitive: anyone can permanently deny login to any IP address by hammering it with bad passwords. That matters more than usual here because `getAddress()` is `socket->peerAddress().toString()` (`serversocketinterface.h:180`), so every user behind a shared NAT, a university, a VPN, or a CGNAT range shares one key. One attacker locks out everyone behind that address, indefinitely.

Check first, then record only when not already blocked:

```cpp
bool RateLimiter::recordAttempt(const QString &key, int maxAttempts, int windowSeconds)
{
    QMutexLocker locker(&mutex);
    const qint64 now = QDateTime::currentSecsSinceEpoch();

    QList<qint64> &timestamps = attempts[key];
    while (!timestamps.isEmpty() && timestamps.first() <= now - windowSeconds) {
        timestamps.removeFirst();
    }
    if (timestamps.size() >= maxAttempts) {
        return true;    // already blocked; do not extend the window
    }
    timestamps.append(now);
    return false;
}
```

With that ordering the window expires on schedule and a legitimate user gets back in after `windowSeconds` of quiet.

servatrice/src/ratelimiter.h
```diff
@@ -0,0 +1,25 @@
...
+    /** @brief Record an attempt for the key and report whether the key is now over the limit. */
+    bool recordAttempt(const QString &key, int maxAttempts, int windowSeconds);
+    /** @brief True if the key already has more than maxAttempts attempts within windowSeconds. */
+    bool isBlocked(const QString &key, int maxAttempts, int windowSeconds) const;
+    /** @brief Drop all recorded attempts for the key, e.g. after a successful login. */
+    void clearAttempts(const QString &key);
+
+private:
+    mutable QMutex mutex;
+    QMap<QString, QList<qint64>> attempts; // key -> attempt timestamps (epoch seconds)
```
ZeldaZach: **`attempts` grows without bound — this is itself a memory-exhaustion DoS.**

Entries are only ever removed by `clearAttempts()`, which is called from exactly one place: `Servatrice::clearFailedLogins()` on a *successful* login. Nothing ever removes `register:` or `forgot:` keys, and nothing removes `login:` keys for an address that never succeeds. Per-key timestamp lists are bounded by the window, but the number of keys is not.

An attacker rotating source addresses adds a permanent `QMap` node plus a `QString` plus a `QList` per address. With IPv6 a single /64 gives them 2^64 distinct keys to walk through. Since a `QMap` node here costs on the order of a hundred bytes, a few million requests is enough to matter, and nothing ever gives the memory back.

This is worth fixing since DoS prevention is the stated purpose of the PR. Options:
- a periodic sweep (`QTimer` on `Servatrice`, or an opportunistic prune every N calls) that erases keys whose newest timestamp is older than the window; and/or
- have `recordAttempt` erase the map entry when the pruned list ends up empty.

The first alone is enough.

servatrice/servatrice.ini.example
```diff
@@ -348,6 +348,27 @@ max_users_websocket=500
 ; Maximum number of users that can connect from the same IP address; useful to avoid bots, default is 4
 max_users_per_address=4
 
+; Maximum number of failed login attempts from a single IP address before that
+; address is temporarily locked out.  Default is 5; set to 0 to disable.
+max_login_attempts_per_ip=5
```
ZeldaZach, line 353: These three "set to 0 to disable" comments (lines 353, 360, 367) currently describe the opposite of what the code does — see the comment on `ratelimiter.cpp`. Once `recordAttempt` early-returns on `maxAttempts <= 0` the text is accurate and can stay as written.

Separately: #7193 also adds keys to a `[security]` section but inserts a *second* `[security]` header near line 106 rather than using this one. These two PRs will conflict here. Worth agreeing on one location now, since #7193 is stacked behind #7192 and both will want this file.

libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
```diff
@@ -527,10 +527,14 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
             return Response::RespUserIsBanned;
         }
         case NotLoggedIn:
+            if (server->recordFailedLogin(getAddress())) {
```
ZeldaZach, line 535: **The limiter throttles the response but not the work.**

`server->loginUser(...)` runs at line 517, and the rate-limit check only happens afterwards while inspecting the result. So a blocked address still gets a full database round trip and a full password verification on every single attempt — the limiter caps how many attempts *succeed*, never how much CPU an attacker can burn.

Today that is a SHA-512 hash and cheap enough to shrug at. But #7193 replaces it with scrypt at N=32768, r=8 (~32 MB of memory and single-digit milliseconds per attempt, by design). Once that lands, an attacker who is already locked out can still force unbounded scrypt derivations, and each one costs the server 32 MB of scratch memory. That is a much better DoS than the one this PR is closing.

`RateLimiter::isBlocked()` already exists for exactly this and is currently unused outside the tests. Adding a gate before `loginUser` would give it a purpose:

```cpp
if (server->isLoginBlocked(getAddress())) {
    return Response::RespTooManyRequests;
}
AuthenticationResult res = server->loginUser(...);
```

(with a matching virtual on `Server` alongside `recordFailedLogin`). Given the two PRs are part of the same series, it's worth doing here rather than after.

tests/rate_limiter_test.cpp
```diff
@@ -0,0 +1,49 @@
...
+TEST(RateLimiterTest, BlocksAttemptsOverLimit)
+{
+    RateLimiter limiter;
+    ASSERT_FALSE(limiter.recordAttempt("ip", 2, 60));
+    ASSERT_FALSE(limiter.recordAttempt("ip", 2, 60));
+    ASSERT_TRUE(limiter.recordAttempt("ip", 2, 60));
+    ASSERT_TRUE(limiter.isBlocked("ip", 2, 60));
+}
+
+TEST(RateLimiterTest, ClearAttempts)
```
ZeldaZach, line 44: Two notes on the tests:

1. `ClearAttempts` and `KeysAreIndependent` both rely on `recordAttempt(key, 0, 60)` returning `true`, which is the bug described on `ratelimiter.cpp:16`. Once `maxAttempts <= 0` means "disabled", these need a real limit (e.g. `recordAttempt("ip", 1, 60)` twice).

2. There's no coverage of the actual sliding behaviour — that attempts age out after `windowSeconds` — which is the part most likely to regress and the part the fixes above touch. It isn't testable right now because `recordAttempt` calls `QDateTime::currentSecsSinceEpoch()` directly.

A small seam makes it testable without a sleep: keep the public signature and add a `now` parameter internally, e.g.

```cpp
bool recordAttempt(const QString &key, int maxAttempts, int windowSeconds)
{
    return recordAttemptAt(key, maxAttempts, windowSeconds, QDateTime::currentSecsSinceEpoch());
}
bool recordAttemptAt(const QString &key, int maxAttempts, int windowSeconds, qint64 now);
```

Then a test can record N attempts at `t`, and assert the key is clear again at `t + windowSeconds + 1`.

ZeldaZach (changes requested):
Reviewed against `master` in a local worktree. The shape of this is right and the wiring into login/register/forgot-password is in the correct places, but there are three problems in `RateLimiter` itself that need fixing before this can land. Two of them are exploitable in the direction opposite to the PR's goal (they create a DoS rather than prevent one), and one makes the documented "set to 0 to disable" knob lock every user out.

Details inline. Nothing here is architectural — all three are contained in `ratelimiter.cpp`/`.h`.

servatrice/src/ratelimiter.cpp
BruebachL: Fixed exactly as suggested: `recordAttempt`/`isBlocked` now early-return false when `maxAttempts <= 0` or `windowSeconds <= 0`, and record nothing in that case. The `ClearAttempts`/`KeysAreIndependent` tests now use a real limit (`recordAttemptAt("ip", 1, 60, ...)` twice). Also covered non-positive window and negative max in `NonPositiveConfigDisablesLimiting`.

servatrice/src/ratelimiter.cpp
BruebachL: Adopted your ordering: trim first, then `if (timestamps.size() >= maxAttempts) return true;` — a key at its limit is rejected without appending, so the window expires on schedule and a legitimate user gets back in after `windowSeconds` of quiet. `isBlocked` was updated from `>` to `>=` to stay consistent (with the new record ordering a key can hold at most `maxAttempts` timestamps, so `>` would never fire). Covered deterministically by `AttemptsAgeOutOfTheWindow`, which also asserts blocked attempts do not extend the lockout.

servatrice/src/ratelimiter.h
BruebachL: Addressed with the opportunistic prune (your first option): recordAttempt runs pruneLocked(now), which — at most once per 60 s — trims each key's list against its stored window and erases keys whose attempts have fully aged out. Combined with "blocked attempts don't record", a key now frees its memory within roughly a window plus the prune interval even if it never succeeds cumulatively and never gets clearAttempts'd.

servatrice/servatrice.ini.example
BruebachL: These comments are accurate unchanged now that maxAttempts <= 0 disables the limiter, so they stay as written. Re the #7193 duplicate [security] header: this branch has only the one section (line 335) — the duplicate lives on #7193's branch, so there is nothing to merge here; whichever PR lands second can drop its header during the rebase rather than shipping two [security] blocks.

libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
BruebachL: Done: a new Server::isLoginRateLimited(ipAddress) virtual (implemented by Servatrice via RateLimiter::isBlocked) gates exactly before loginUser() in cmdLogin, returning RespTooManyRequests without touching the database or the password hash. A locked-out address now burns no scrypt even after #7193 replaces SHA-512, and the until-now-unused isBlocked() finally has a production caller.

tests/rate_limiter_test.cpp
BruebachL: Both notes addressed: (1) the two tests now use recordAttemptAt("ip", 1, 60, ...) twice instead of the maxAttempts = 0 bug; (2) the code gained the time-injection seam you sketched — public recordAttemptAt()/isBlockedAt() taking now, with recordAttempt()/isBlocked() delegating to currentSecsSinceEpoch(). This makes the sliding behaviour fully deterministic and sleep-free: AttemptsAgeOutOfTheWindow asserts a key blocked at t with window=60 is clear again at t+61, and that blocked attempts at t+3 do not extend the lockout. 7/7 tests pass.

tooomm:
After conflicts are resolved, Zach could have a second look here to get this moving again.

PR #7077, opened 2026-08-04
[Security] Redact sensitive user data from client-visible responses

BruebachL:
The getUserInfo command for a user that is not currently online returned the full database record, including the account id, the email address and the stored client id, to any logged-in requester. Mirror the redaction already applied to online users via copyUserInfo(): the id and email are only ever exposed to the account owner, and the client id only to moderators.

The buddy/ignore add-to-list event likewise returned the target user's email address and client id to the requester. The list entry only needs the public profile fields, so strip the email and client id from it as well.


ebbit1q (approved):
I'd merge this as it's a real security issue however I'd prefer this be handled at a lower level in servatrice_database_interface.cpp rather than redacting at the moment before sending this out

PR #7078, opened 2026-08-04
[Chat] Don't tag cards that have no match in the database

### LINKED ISSUE #2217: [Feature Request] In chat, don't tag cards that have no match in the database.
Don't highlight and italicise `[[mispelt cards]]` that don't match the receiving user's card database, or highlight them a different colour that no one will want to use.

**Benefits:**
It gives immediate feedback to a the sending user that the card has not been typed correctly, and ruins the fun of anyone that does it on purpose.

BruebachL:
## Related Ticket(s)
- Fixes #2217

Unknown cards still get valid tags

- Check for db match, abort if not.

### DIFF
--- a/cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
+++ b/cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
@@ -63,14 +63,14 @@ void ChatView::adjustColorsToPalette()
         serverMessageColor = QColor(0xFF, 0x73, 0x83);
         otherUserColor = otherUserColor.lighter(150);
         linkColor = QColor(71, 158, 252);
-        unresolvedCardTagColor = QColor(0xFF, 0x73, 0x83);
+        unresolvedCardTagColor = QColor(0xFF, 0xA5, 0x00);
     } else {
         document()->setDefaultStyleSheet(R"(
             a { text-decoration: none; color: blue; }
             .blue { color: blue }
         )");
         linkColor = palette().link().color();
-        unresolvedCardTagColor = QColor(0x85, 0x15, 0x15);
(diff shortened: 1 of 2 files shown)


RickyRister:
I still think a failed match should highlight the text in a different color, instead of straight up no-op'ing. That way others can still see that you were attempting to tag a card.

cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
```diff
@@ -173,13 +176,22 @@ void ChatView::appendHtmlServerMessage(const QString &html, bool optionalIsBold,
...
+        cursor.setCharFormat(cardFormat);
+        cursor.insertText(cardName);
+        cursor.setCharFormat(oldFormat);
+        return;
+    }
 
-    cursor.setCharFormat(anchorFormat);
+    cardFormat.setForeground(linkColor);
+    cardFormat.setAnchor(true);
+    cardFormat.setAnchorHref("card://" + cardName);
```
RickyRister, line 192: nit: It would be cleaner to do it like this

```suggestion
    if (CardDatabaseManager::query()->lookupCardByName(cardName)) {
        cursor.setCharFormat(anchorFormat);
        cardFormat.setForeground(linkColor);
        cardFormat.setAnchor(true);
        cardFormat.setAnchorHref("card://" + cardName);
    } else {
        cardFormat.setForeground(unresolvedCardTagColor);
    }
```

RickyRister (approved):
Kinda wonky for the game system text to share the same color as the failed match, but I suppose it's fine.
I guess people can use it for in-game RP if they want.

Image

PR #7079, opened 2026-08-04
[Chat] Scroll chat view to bottom when loading chat history

### LINKED ISSUE #2725: start chatview scrolled down to the most recent message
<b>OS:</b>
*Win 10*

<b>Cockatrice version:</b>
*Version 2.3.18-custom_f0a561d (2017-05-20)*

__________

If you join a server we should make sure that you automatically see the most recent message.
Otherwise the chat won't auto-scroll and follow the chat - he will just be stuck and you have to manually scroll down first - then it works again:
scroll

This might be related to the chat history being bigger than the chat window itself. <br>(sometimes/always depending on the server setting and your screen)

BruebachL:
## Related Ticket(s)
- Fixes #2725

See ticket

- Always scroll down on history

### DIFF
--- a/cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
+++ b/cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
@@ -326,7 +326,7 @@ void ChatView::appendMessage(QString message,
         }
     }
 
-    if (atBottom) {
+    if (atBottom || messageType.testFlag(Event_RoomSay::ChatHistory)) {
         verticalScrollBar()->setValue(verticalScrollBar()->maximum());
     }
 }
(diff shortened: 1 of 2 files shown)


RickyRister:
Are you sure this hasn't already been fixed? When I join a server, the chat is already at the bottom

https://github.com/user-attachments/assets/b7e68009-b8aa-4d0b-9706-acd682a4b2ac

BruebachL:
I've not encountered this myself either but the proposed change can't hurt

RickyRister:
I think what the issue is asking for is for the chat to be scrolled all the way down so that new messages will automatically show up at the bottom.

https://github.com/user-attachments/assets/00b4a100-535a-4efa-b202-cec0508fccd2

This PR doesn't even accomplish that.

https://github.com/user-attachments/assets/1a63e3d9-4ef0-4e80-be0e-316d6e424a0f

RickyRister (changes requested):
Still doesn't work 

https://github.com/user-attachments/assets/28a1fa8d-281a-4b2e-9094-5df115c7c5cc

RickyRister (approved):
Works now

https://github.com/user-attachments/assets/8a4735ce-c1f6-4273-a9d5-edc770be2d55

PR #7084, opened 2026-08-05
[Settings] Shuffle some settings around

BruebachL:
I played it loose and fast with assigning new keys during the #7050 split, this commit re-organizes some leftover keys that weren't quite at their right location.

### DIFF
--- a/cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
+++ b/cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
@@ -209,12 +209,12 @@ AppearanceSettingsPage::AppearanceSettingsPage()
             &CardsDisplaySettings::setStackCardOverlapPercent);
 
     cardViewInitialRowsMaxBox.setRange(1, 999);
-    cardViewInitialRowsMaxBox.setValue(SettingsCache::instance().interface().getCardViewInitialRowsMax());
+    cardViewInitialRowsMaxBox.setValue(SettingsCache::instance().userInterface().getCardViewInitialRowsMax());
     connect(&cardViewInitialRowsMaxBox, qOverload<int>(&QSpinBox::valueChanged), this,
             &AppearanceSettingsPage::cardViewInitialRowsMaxChanged);
 
     cardViewExpandedRowsMaxBox.setRange(1, 999);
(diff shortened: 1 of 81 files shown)


cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
```diff
@@ -140,31 +140,16 @@ AppearanceSettingsPage::AppearanceSettingsPage()
...
+    connect(&styleUserListCheckBox, &QCheckBox::QT_STATE_CHANGED, &settings.appearance(),
+            &AppearanceSettings::setStyleUserList);
 
     auto stylingTabGrid = new QGridLayout;
     stylingTabGrid->addWidget(&styleUserListCheckBox, 0, 0, 1, 2);
 
     stylingGroupBox = new QGroupBox;
     stylingGroupBox->setLayout(stylingTabGrid);
 
-    // Menu settings
```
RickyRister, line 153: Can we please not move these settings in this PR? 
The `General interface settings` box is starting to become bloated.
Even if we do want to eventually move this settings, we can decide where to move it in a separate PR.

Image

BruebachL: We should probably make a new category for "Games Interface" settings or similar where we can stuff all the TabGame related things to separate them from general UI choices (like deck editor/deck storage)

BruebachL: I'm not a huge fan of adding another category to the top because of the way the settings are designed right now.

I've toyed with the idea of moving the navigation from the top to the left, as sort of a ribbon. A lot of other UIs have chosen this as a standard (navigation on the left) and it would have some advantages like:

More intuitive scrolling behavior for hidden categories
Possibility to convert the current icon tabs into an iconed tree-view so we can take proper advantage of sub-grouping and only display settings for that relevant subgroup.

BruebachL: i.e. 
image
image

RickyRister (approved):
Haven't checked the migrations. In case we also want @ZeldaZach to look at this.

PR #7085, opened 2026-08-08
[TabServer] ensure "isOpen" settings value is always persisted/kept in sync

BruebachL:
If you close the server tab and then open it through the home widget "Play"  the setting for that causes the tab to automatically reopen when you launch trice is never persisted. Only the "Tabs -> Server" checkbox persists it

- Persist whenever the tab is opened.

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
@@ -583,6 +583,7 @@ void TabSupervisor::actTabServer(bool checked)
 
 void TabSupervisor::openTabServer()
 {
+    SettingsCache::instance().tabs().setTabServerOpen(true);
     if (tabServer) {
         return;
     }


RickyRister (approved):

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -583,6 +583,7 @@ void TabSupervisor::actTabServer(bool checked)
 
 void TabSupervisor::openTabServer()
```
RickyRister, line 584: By the way, why doesn't this method call into the `actTabServer` flow?

RickyRister (approved):

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
BruebachL: I could not tell you, I don't remember the last time I modified this.

PR #7087, opened 2026-08-08
[UserList] Bulk load to prevent hang on connect, fix multi-monitor positioning

BruebachL:
We unconditionally fetch everyone user info and sort on each received one which leads to a slight hang on connect.

- bulk load and stop sorting during bulk loading
- Only refresh visible or near-visible users
- Fix multi-monitor positioning
- Fix a nullptr connect warning

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
@@ -23,6 +23,7 @@
 #include <QPlainTextEdit>
 #include <QPushButton>
 #include <QRadioButton>
+#include <QScreen>
 #include <QSpinBox>
 #include <QWidget>
 #include <libcockatrice/card/database/card_database_manager.h>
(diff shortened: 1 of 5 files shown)


cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -543,15 +544,24 @@ UserListWidget::UserListWidget(TabSupervisor *_tabSupervisor,
...
 
     // Forward join requests from popup upward
     connect(m_userInfoPopup, &UserInfoPopup::joinGameRequested, this, &UserListWidget::joinGameRequested);
 
-    connect(avatarProvider, &UserAvatarProvider::avatarUpdated, this,
-            [this](const QString &) { userTree->viewport()->update(); });
-    connect(cardArtProvider, &UserCardArtProvider::cardArtUpdated, this,
-            [this](const QString &) { userTree->viewport()->update(); });
+    connect(avatarProvider, &UserAvatarProvider::avatarUpdated, this, [this](const QString &name) {
+        userTree->viewport()->update();
```
RickyRister: nit: feels like this could be extracted into a method.

RickyRister (approved):
Can confirm that the multi-monitoring positioning is fixed.

PR #7088, opened 2026-08-09
[CardInfo] Encapsulate lazy properties loading into new class

RickyRister:
## Related Ticket(s)
- Follow-up to #7070
- Follow-up to #7051

With the addition of the lazy properties loading, the `CardInfo` and `PrintingInfo` classes now both need to manage and synchronize the properties materialization. It would be cleaner if we pulled out all that logic into a separate class.

- Create `LazyPropertiesHash` class
  - Encapsulates the whole "store serialized blob and only materialize it on first read" logic, with the mutex.
    - It now clears the blob after materializing the properties.
  - Overrides copy constructor and copy-assignment since mutex isn't copiable.
- Use the class in `CardInfo` and `PrintingInfo`
  - Change `PrintingInfo`'s constructors to just take a `LazyPropertiesHash`
  - Since `CardInfo`'s constructors have different number of params, I kept the constructors the same for now and just created the `LazyPropertiesHash` in the constructor.

### DIFF
--- a/libcockatrice_card/libcockatrice/card/card_info.cpp
+++ b/libcockatrice_card/libcockatrice/card/card_info.cpp
@@ -25,7 +25,11 @@ const QHash<QString, QString> &CardInfo::getPropertiesHash() const
 
 void CardInfo::setProperty(const QString &_name, const QString &_value)
 {
-    properties.insert(_name, _value);
+    bool changed = properties.insert(_name, _value);
+    if (!changed) {
+        return;
+    }
+
     emit cardInfoChanged(smartThis);
 }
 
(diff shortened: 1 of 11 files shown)


libcockatrice_card/libcockatrice/card/card_info.cpp
```diff
@@ -21,64 +18,29 @@ class CardInfo;
...
 
 void CardInfo::setProperty(const QString &_name, const QString &_value)
 {
-    ensurePropertiesLoaded();
-    if (propertiesCache.value(_name) == _value) {
-        return;
-    }
-    propertiesCache.insert(_name, _value);
-    propertiesBlob = serializeProperties(propertiesCache);
+    properties.insert(_name, _value);
```
BruebachL: This no longer early returns if the value is the same

BruebachL (approved):

libcockatrice_card/libcockatrice/card/card_info.cpp
RickyRister: good catch

PR #7091, opened 2026-08-10
[Server/Client/Protocol] Reporting users + moderation queue functionality

### LINKED ISSUE #190: In client reporting

BruebachL:
## Related Ticket(s)
- Fixes #190

Users have to direct message moderators to report other users. This leads to a variety of problems such as:
- No moderator available at the time
- No coordination between moderators
- No records of decisions made
- etc.

For regular users:
- Right-click a user → "Report user" opens a dialog where you can file a report directly in the client
- You can attach context (game/chat info) to the report
- A "My Reports" dialog lets you track your submitted reports and their status

For moderators:
- A moderation queue (tab_report) showing incoming reports, with actions to assign, resolve, comment, and pull up user info/statistics
- Server-side report stats so mods can see the history/volume of reports
- Join games/view replays directly from the report

For admins:
Limits configurable in servatrice.ini

```
[reporting]
; Maximum number of user reports a single user can file per day; default is 10; set to 0 to disable the limit
max_reports_per_day=10

; Maximum number of report comments a single user can post per hour; default is 30; set to 0 to disable the limit
max_comments_per_hour=30
```

Under the hood:
- New protocol commands: Command_Report, Command_ReportList, Command_ReportDetails, Command_ReportResolve, Command_ReportAssign, Command_ReportAddComment, Command_ReportMyList, Command_ReportStats, Command_ReportUserInfo, plus response/event messages
- New servatrice.sql schema version 36 + migration servatrice_0035_to_0036.sql for the reports tables
- Server DB interface + socket handler wiring
- New strings/limits in string_limits.h for report text, reports-related settings in tabs_settings
- Includes replay-download-by-game-id plumbing so a report can reference a game that can be reviewed later

## Screenshots
image
image
(Chat log CAN be included in a game report, it just needs to be triggered FROM chat, not the user list)
image
image
image

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
@@ -151,7 +151,7 @@ WarningDialog::WarningDialog(const QString userName, const QString clientID, QWi
     warnClientID = new QLineEdit(clientID);
     warnClientID->setMaxLength(MAX_NAME_LENGTH);
     warningOption = new QComboBox();
-    warningOption->addItem("");
+    warningOption->addItem("", "");
 
     deleteMessages = new QCheckBox(tr("Redact all messages from this user in all rooms"));
 
@@ -184,7 +184,7 @@ void WarningDialog::okClicked()
         return;
     }
 
-    if (warningOption->currentText().simplified().isEmpty()) {
+    if (warningOption->currentData().toString().simplified().isEmpty()) {
         QMessageBox::critical(this, tr("Error"),
                               tr("Warning to use can not be blank, please select a valid warning to send."));
         return;
@@ -205,7 +205,7 @@ QString WarningDialog::getWarnID() const
 
 QString WarningDialog::getReason() const
 {
-    return warningOption->currentText().simplified();
+    return warningOption->currentData().toString().simplified();
 }
 
 int WarningDialog::getDeleteMessages() const
@@ -216,9 +216,9 @@ int WarningDialog::getDeleteMessages() const
 void WarningDialog::addWarningOption(const QString warning, int startingIl)
 {
     if (startingIl > 1) {
-        warningOption->addItem(tr("%1 (IL %2)").arg(warning).arg(startingIl));
+        warningOption->addItem(tr("%1 (IL %2)").arg(warning).arg(startingIl), warning);
     } else {
-        warningOption->addItem(warning);
+        warningOption->addItem(warning, warning);
     }
 }
 
--- a/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
@@ -213,9 +213,13 @@ int WarningDialog::getDeleteMessages() const
     return deleteMessages->isChecked() ? -1 : 0;
 }
 
-void WarningDialog::addWarningOption(const QString warning)
+void WarningDialog::addWarningOption(const QString warning, int startingIl)
 {
-    warningOption->addItem(warning);
+    if (startingIl > 1) {
+        warningOption->addItem(tr("%1 (IL %2)").arg(warning).arg(startingIl));
+    } else {
+        warningOption->addItem(warning);
+    }
 }
 
 void BanDialog::okClicked()
--- a/cockatrice/src/interface/widgets/tabs/tab_report.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_report.cpp
@@ -810,7 +810,7 @@ void TabReport::userInfoResponse(const Response &response)
 
     const Response_ReportUserInfo &resp = response.GetExtension(Response_ReportUserInfo::ext);
 
-    if (resp.has_user_name() && resp.user_name() != lastRequestedUser) {
+    if (resp.has_user_name() && QString::fromStdString(resp.user_name()) != lastRequestedUser) {
         return;
     }
 
--- a/cockatrice/src/interface/widgets/tabs/tab_report.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_report.cpp
@@ -1,0 +1,919 @@
+#include "tab_report.h"
+
+#include "../utility/report_utils.h"
+#include "abstract_client.h"
+#include "tab_supervisor.h"
+
+#include <QCheckBox>
+#include <QComboBox>
+#include <QDateTime>
+#include <QFontDatabase>
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QHBoxLayout>
+#include <QHeaderView>
+#include <QInputDialog>
+#include <QLabel>
+#include <QLineEdit>
+#include <QMessageBox>
+#include <QPushButton>
+#include <QSignalBlocker>
+#include <QSplitter>
+#include <QTableWidget>
+#include <QTextEdit>
+#include <QTimer>
+#include <QVBoxLayout>
+#include <QWidget>
+#include <libcockatrice/protocol/pb/command_replay_download_by_game_id.pb.h>
+#include <libcockatrice/protocol/pb/command_report_add_comment.pb.h>
+#include <libcockatrice/protocol/pb/command_report_assign.pb.h>
+#include <libcockatrice/protocol/pb/command_report_details.pb.h>
+#include <libcockatrice/protocol/pb/command_report_list.pb.h>
+#include <libcockatrice/protocol/pb/command_report_resolve.pb.h>
+#include <libcockatrice/protocol/pb/command_report_stats.pb.h>
+#include <libcockatrice/protocol/pb/command_report_user_info.pb.h>
+#include <libcockatrice/protocol/pb/game_replay.pb.h>
+#include <libcockatrice/protocol/pb/response_replay_download_by_game_id.pb.h>
+#include <libcockatrice/protocol/pb/response_report_details.pb.h>
+#include <libcockatrice/protocol/pb/response_report_list.pb.h>
+#include <libcockatrice/protocol/pb/response_report_stats.pb.h>
+#include <libcockatrice/protocol/pb/response_report_user_info.pb.h>
+#include <libcockatrice/protocol/pb/serverinfo_report.pb.h>
+#include <libcockatrice/protocol/pending_command.h>
+
+namespace
+{
+constexpr int COL_ID = 0;
+constexpr int COL_TIME = 1;
+constexpr int COL_REPORTER = 2;
+constexpr int COL_REPORTED = 3;
+constexpr int COL_CATEGORY = 4;
+constexpr int COL_GAMEID = 5;
+constexpr int COL_STATUS = 6;
+constexpr int COL_ASSIGNED = 7;
+constexpr int COL_REPLAY = 8;
+constexpr int COL_ROOM = 9;
+constexpr int COL_COUNT = 10;
+constexpr int REFRESH_INTERVAL_MS = 300000;
+} // namespace
+
+TabReport::TabReport(TabSupervisor *_tabSupervisor, AbstractClient *_client) : Tab(_tabSupervisor), client(_client)
+{
+    auto *centralWidget = new QWidget(this);
+    setCentralWidget(centralWidget);
+
+    searchEdit = new QLineEdit;
+    searchEdit->setClearButtonEnabled(true);
+    connect(searchEdit, &QLineEdit::textChanged, this, &TabReport::applyFilters);
+
+    statusFilter = new QComboBox;
+    connect(statusFilter, &QComboBox::currentIndexChanged, this, &TabReport::applyFilters);
+
+    unresolvedOnlyBox = new QCheckBox;
+    unresolvedOnlyBox->setChecked(true);
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+    connect(unresolvedOnlyBox, &QCheckBox::checkStateChanged, this, &TabReport::refreshList);
+#else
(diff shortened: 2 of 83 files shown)


servatrice/src/serversocketinterface.cpp
```diff
@@ -1249,6 +1306,952 @@ Response::ResponseCode AbstractServerSocketInterface::cmdBanFromServer(const Com
...
+    if (!sqlInterface->checkSql()) {
+        return Response::RespInternalError;
+    }
+
+    const QString userName = nameFromStdString(cmd.user_name()).simplified();
+    if (userName.isEmpty()) {
+        return Response::RespContextError;
+    }
+
+    const QString tempPassword = PasswordHasher::generateRandomSalt();
```
ZeldaZach, line 1841: This is the blocking one for me.

`RESET_USER_PASSWORD` is registered in `processExtendedModeratorCommand`, so it's gated on `IsModerator` only (`server_protocolhandler.cpp:321`). There's no check on the *target's* privilege level, and the generated password is returned to the caller in `Response_ResetUserPassword.temporary_password`.

So any moderator can pick an admin (or another moderator) out of the staff list that `GET_MODERATOR_LAST_LOGINS` conveniently provides, reset their password, and log in as them. That's a straight moderator -> admin escalation, and it's qualitatively different from the other moderator powers — banning someone is loud and reversible, silently owning their account is neither.

At minimum this needs to refuse when the target holds admin/moderator flags the caller doesn't. I'd argue it should be an `AdminCommand` outright.

Two smaller things while you're in here:

- `changeUserPassword(userName, tempPassword, true)` only writes `password_sha512` and `passwordLastChangedDate`. There's no force-change-on-next-login flag anywhere in the schema, but `tab_moderation.cpp` tells the moderator "The user must change it on their first login." Nothing enforces that — either add the flag or fix the dialog text.
- There's no `Event_NotifyUser` to the affected user and no session invalidation, so someone whose password is reset finds out when they next fail to log in.

cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -203,9 +203,13 @@ int WarningDialog::getDeleteMessages() const
     return deleteMessages->isChecked() ? -1 : 0;
 }
 
-void WarningDialog::addWarningOption(const QString warning)
+void WarningDialog::addWarningOption(const QString warning, int startingIl)
 {
-    warningOption->addItem(warning);
+    if (startingIl > 1) {
+        warningOption->addItem(tr("%1 (IL %2)").arg(warning).arg(startingIl));
```
ZeldaZach: The IL gets baked into the combo box's display text, and `getReason()` (line 196) returns `warningOption->currentText().simplified()`.

`user_context_menu.cpp:295` then does `cmd.set_reason(dlg->getReason().toStdString())`, so the server stores `"Cheating (IL 2)"` in `cockatrice_warnings.reason` and sends that same string back to the warned user in `Event_NotifyUser.warning_reason`. It also splits the warn history: `"Cheating"` and `"Cheating (IL 2)"` are now two distinct reasons for anything grouping on that column.

Keep the level out of the wire value — stash the raw name as item data and read it back:

```cpp
void WarningDialog::addWarningOption(const QString warning, int startingIl)
{
    if (startingIl > 1) {
        warningOption->addItem(tr("%1 (IL %2)").arg(warning).arg(startingIl), warning);
    } else {
        warningOption->addItem(warning, warning);
    }
}

QString WarningDialog::getReason() const
{
    return warningOption->currentData().toString().simplified();
}
```

Note the blank first entry added at line 144 also needs data set, or the empty-selection guard at line 177 should switch to `currentData()` too.

cockatrice/src/interface/widgets/tabs/tab_report.cpp
```diff
@@ -0,0 +1,919 @@
...
+    }
+
+    userContextRecentReports->clear();
+    if (resp.recent_reports_size() == 0) {
+        userContextRecentReports->setPlainText(tr("No previous reports against this user."));
+    } else {
+        for (int i = 0; i < resp.recent_reports_size(); ++i) {
+            const ServerInfo_Report &r = resp.recent_reports(i);
+            QDateTime dt = QDateTime::fromSecsSinceEpoch(r.report_time());
+            userContextRecentReports->append(QString("[%1] #%2 by %3 [%4]: %5")
```
ZeldaZach: `QTextEdit::append()` runs the string through `Qt::mightBeRichText()` and calls `insertHtml()` if it matches, so this renders as HTML rather than plain text.

The detection scan walks forward until the first `<` or `\n`. This format string has no newline, so it reaches `%5` — the report category — which is fully attacker-controlled: `cmdReport` only runs it through `nameFromStdString()`, which truncates but doesn't restrict characters. A modified client filing a report with category `` or `<b>` gets that interpreted as markup in every moderator's client.

Worst case here is display spoofing rather than code execution (plain `QTextEdit` won't fetch remote resources), but a moderator reading a queue entry shouldn't be looking at content the reported user controls the rendering of.

`statsDetailText->append(...)` at line 903 has the same shape with the category from `Response_ReportStats`.

Swapping to explicit plain-text insertion fixes both:

```cpp
userContextRecentReports->moveCursor(QTextCursor::End);
userContextRecentReports->insertPlainText(line + "\n");
```

Worth also validating `category` server-side in `cmdReport` against the known set — the client only ever sends one of five fixed values, so anything else is already a client that isn't playing by the rules.

(`report_utils::renderReportDetails` uses `append()` too. It happens to be safe today because its first line ends in `\n` before any user text, but it's the same footgun one format-string edit away from biting.)

servatrice/src/serversocketinterface.cpp
```diff
@@ -1249,6 +1306,952 @@ Response::ResponseCode AbstractServerSocketInterface::cmdBanFromServer(const Com
...
+    } else if (assignedToId > 0) {
+        QSqlQuery *modNameQuery = sqlInterface->prepareQuery("SELECT name FROM {prefix}_users WHERE id = :id");
+        modNameQuery->bindValue(":id", assignedToId);
+        if (sqlInterface->execSqlQuery(modNameQuery) && modNameQuery->next()) {
+            recipients.append(modNameQuery->value(0).toString());
+        }
+    }
+
+    const QString ownName = QString::fromStdString(userInfo->name());
+    bool allNotified = true;
```
ZeldaZach: `allNotified` starts `true` and is only cleared inside the recipient loop, so an empty `recipients` list marks the comment as notified.

That's exactly what happens when a reporter comments on a report that's still `open`: `isMod` is false and `assignedToId` is 0, so nothing is appended, the loop body never runs, and `notified` is set to 1.

The pending-notification query in `sendPendingReportNotifications` would have picked that comment up later via `(c.is_moderator = 0 AND r.assigned_to = :user_id)` once a moderator assigned the report — but it's already flagged as delivered, so that moderator never hears about it. Follow-up detail from reporters on unassigned reports silently never surfaces.

Initialising from the recipient list handles it:

```cpp
bool allNotified = !recipients.isEmpty();
```

Separately, when there are two recipients and only one is online, `allNotified` stays false and the online one gets the same comment again on next login. Tracking `notified` per recipient rather than per comment would fix that properly, though it's the less annoying of the two failure modes.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1249,6 +1306,952 @@ Response::ResponseCode AbstractServerSocketInterface::cmdBanFromServer(const Com
...
+    }
+
+    int reportId = cmd.report_id();
+    QString note = textFromStdString(cmd.resolution_note());
+    bool dismissed = cmd.dismissed();
+
+    QString newStatus = dismissed ? "dismissed" : "resolved";
+
+    QSqlQuery *query = sqlInterface->prepareQuery("UPDATE {prefix}_reports "
+                                                  "SET status = :status, resolution_note = :note, "
```
ZeldaZach, line 1459: This never records who resolved the report. `assigned_to` is left untouched, so a report that goes straight from `open` to `resolved`/`dismissed` — which is every report a moderator handles without clicking "Assign to Me" first, and the UI happily allows that (`updateActionStates` enables Resolve on `status == "open"`) — ends up with `assigned_to = NULL` and no trace of the moderator who closed it.

The queue then shows a blank "Assigned To" column for it, and `Response_ReportStats.avg_resolution_hours` is the only thing left tying the row to any activity.

Given the issue this PR is fixing lists "No records of decisions made" as a motivation, I think the resolving moderator needs to be persisted. Setting `assigned_to = :mod_id` in this same UPDATE is the cheap fix; a dedicated `resolved_by` column is the honest one, since it distinguishes "took ownership" from "closed it out".

An `addAuditRecord` call here would also match what `cmdResetUserPassword` and `cmdRemoveUserAvatar` already do.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1249,6 +1306,952 @@ Response::ResponseCode AbstractServerSocketInterface::cmdBanFromServer(const Com
...
+    }
+
+    if (!query->next()) {
+        return Response::RespNameNotFound;
+    }
+
+    const bool isMod = userInfo->user_level() & ServerInfo_User::IsModerator;
+    const int reporterId = query->value(1).toInt();
+    const QString reporterName = query->value(2).toString();
+    if (reporterId != userInfo->id() && reporterName != QString::fromStdString(userInfo->name()) && !isMod) {
```
ZeldaZach: The `reporterName != userInfo->name()` half of this check is a second, weaker path to the same access, and it can outlive the account it belongs to.

`cockatrice_reports.reporter_id` is `ON DELETE SET NULL`, but `reporter_name` is a plain `varchar(35)` snapshot that survives. Once the original reporter's account is deleted, anyone who registers that username matches on the name and gets the full report — description, chat log and all comments.

The id comparison is already correct and sufficient; I'd just drop the name fallback:

```cpp
if (reporterId != userInfo->id() && !isMod) {
    return Response::RespAccessDenied;
}
```

One caveat if you do: `query->value(1).toInt()` yields 0 for a NULL `reporter_id`, so guard against a caller whose `userInfo->id()` could also be 0 (unregistered sessions are already excluded by the `authState != PasswordRight` check above, so this should be fine — worth confirming).

Same comparison at line 2138 in `cmdReportAddComment`.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1314,6 +1314,167 @@ QList<ServerInfo_Warning> Servatrice_DatabaseInterface::getUserWarnHistory(const
...
+                          "FROM {prefix}_users u "
+                          "LEFT JOIN {prefix}_user_analytics a ON a.id = u.id "
+                          "WHERE u.name = :user_name";
+    if (!seedEmail.isEmpty()) {
+        queryString.append(" OR u.email = :seed_email");
+    }
+    if (!seedClientId.isEmpty()) {
+        queryString.append(" OR u.clientid = :seed_clientid");
+    }
+    queryString.append(" OR u.name IN (SELECT DISTINCT s.user_name FROM {prefix}_sessions s "
```
ZeldaZach, line 1401: This IP-correlation clause has no `LIMIT` and is a doubly-nested `IN (SELECT ... FROM {prefix}_sessions ...)` over what is typically one of the largest tables on the server.

MySQL is historically bad at optimising `IN` subqueries in this shape, and the blast radius depends entirely on the data: one user behind CGNAT, a VPN exit, or a university NAT shares an IP with hundreds of accounts, each of which drags in two correlated `COUNT(*)` subqueries for warnings and bans in the outer SELECT.

Since `investigateUser()` fires this on every lookup and there's no client-side throttle, a moderator clicking through names can put real load on the DB.

Worth adding a `LIMIT` (the client renders these into a table anyway) and capping the IP set — e.g. only correlate on sessions from the last N months, which is also the more useful answer for alt detection than "shared an IP once in 2019".

Also note `getUserSessions` binds `cmd.limit()` straight into `LIMIT :limit` with no upper bound; the proto default is 110 but nothing stops a client sending `UINT32_MAX`.

ZeldaZach (commented):
Reviewed the whole diff (78 files). The reporting/queue feature itself is well put together — parameterised SQL throughout, sensible rate limits, optimistic-concurrency on assign, and the offline-notification fallback via `onLogin` is a nice touch. A few things I'd like sorted before this goes in, listed inline.

The one I care most about is `cmdResetUserPassword`: it's a moderator-level command that hands the caller a working plaintext password for any account, with no check on the target. That's a moderator -> admin takeover path.

A couple of non-blocking observations that didn't warrant inline comments:

- **Scope.** The last four commits (`Command_GetUserSessions`, `GetUserAlts`, `GetModeratorLastLogins`, `ResetUserPassword`, `RemoveUserAvatar`, `tab_moderation`) add a whole second moderator toolset that isn't mentioned anywhere in the PR description. Given the password-reset issue below, I'd rather see those split into their own PR so they get reviewed on their own terms.
- `report_utils.cpp` hardcodes status colours (`#e74c3c`, `#f39c12`, ...). Everything else in these dialogs uses `palette(...)`; these will fight some themes.
- `cmdReportList` caps at `LIMIT 1000` with no pagination or offset. Fine for now, but the queue silently truncates once a server accumulates that many reports and there's no indication in the UI.
- `DlgReportUser` always attaches the captured chat log with no way to drop it. The PR description says context is something you "can attach"; right now if you report from a chat view you're always uploading the last 50 lines, including DMs if the menu was opened in a message tab.
- `SessionCommand` enum skips 1201-1203 between `REPORT` and `REPORT_MY_LIST`. Harmless, presumably squash leftovers, just noting it so nobody reuses them later.
- Schema/migration, proto extension numbering (no collisions), admin bitmask handling in `getModeratorLastLogins`, and the `warning_categories` parser + tests all look correct to me.

ZeldaZach (approved):

PR #7093, opened 2026-08-10
Parent MoveMenu to CardMenu to fix memory leak on menu rebuild

DawnFire42:
Found this leak while improving my command zone code, though it shouldn't be bundled in the same PR.
MoveMenu is created without a Qt parent, but QMenu::addMenu(QMenu*) does not take ownership. Since CardMenu is rebuilt on every right-click, each rebuild leaks one MoveMenu instance.

- adds a parent parameter to MoveMenu and passes this (the CardMenu) at all call sites. Qt's parent-child mechanism now deletes MoveMenu when CardMenu is destroyed.

### DIFF
--- a/cockatrice/src/game_graphics/player/menu/card_menu.cpp
+++ b/cockatrice/src/game_graphics/player/menu/card_menu.cpp
@@ -203,7 +203,7 @@ void CardMenu::createTableMenu(bool canModifyCard)
     }
     addSeparator();
     addAction(aClone);
-    addMenu(new MoveMenu(player));
+    addMenu(new MoveMenu(player, this));
     addSeparator();
     addAction(aAttach);
     if (card->getAttachedTo()) {
@@ -253,7 +253,7 @@ void CardMenu::createStackMenu(bool canModifyCard)
     addAction(aPlayFacedown);
     addSeparator();
     addAction(aClone);
-    addMenu(new MoveMenu(player));
+    addMenu(new MoveMenu(player, this));
     addSeparator();
     addAction(aAttach);
     addAction(aDrawArrow);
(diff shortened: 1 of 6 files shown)


BruebachL:
PlayerMenu::updateCardMenu constructs new CardMenu(player, card, shortcutsActive) but itself never parents the QMenu.
requestCardMenuUpdate then orphans each old menu.
TabGame::setCardMenu similiarly never reparents.

tl;dr this fix is valid but ultimately, there are more leaks that prevent this from being meaningful

DawnFire42:
I see. 

then managing CardMenu lifecycle in PlayerMenu seems like the most logical choice to me, does this work for you?

BruebachL:
Yes, thank you for looking into this :)

DawnFire42:
It's rebased and ready for another look

BruebachL:
Two issues:

Guard-hit path deletes the tracked menu without notifying TabGame. In the new PlayerMenu::updateCardMenu, the deleteLater() happens before the spectator/getActiveCard() != card early-return, which returns without emitting cardMenuUpdated. If a requestCardMenuUpdate(card) arrives for a non-active card (these are emitted from several places in player_event_handler.cpp:256/374/401/470), the tracked menu dies while aCardMenu still holds it which leaves a stale/dangling QToolButton menu until the next successful rebuild. 

Fix: also emit cardMenuUpdated(nullptr) on that path (or move the deletion after the guards).

The null path clears whichever menu is attached i.e., the doomed old CardMenu, leaving aCardMenu pointing at a dead menu afterward. A single persistent placeholder QMenu member in TabGame would be cleaner than clear().

PR #7095, opened 2026-08-11
[Game] Arrow animation

BruebachL:
UI polish is good and looks cute.

- Add a little animation to arrow drawing instead of simply phasing them into existence.

## Screenshots

https://github.com/user-attachments/assets/18f80377-a570-42d8-bd7e-a7a2b87eb30f

### DIFF
--- a/cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
+++ b/cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
@@ -116,13 +116,25 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     connect(&tapAnimationCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().cardsDisplay(),
             &CardsDisplaySettings::setTapAnimation);
 
-    animationsEnabledCheckBox.setChecked(SettingsCache::instance().userInterface().getAnimationsEnabled());
-    connect(&animationsEnabledCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().userInterface(),
-            &InterfaceSettings::setAnimationsEnabled);
+    arrowDrawAnimationCheckBox.setChecked(SettingsCache::instance().cardsDisplay().getArrowDrawAnimation());
+    connect(&arrowDrawAnimationCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().cardsDisplay(),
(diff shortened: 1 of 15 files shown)


cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
```diff
@@ -310,6 +315,7 @@ void UserInterfaceSettingsPage::retranslateUi()
     specNotificationsEnabledCheckBox.setText(tr("Notify in the taskbar for game events while you are spectating"));
     buddyConnectNotificationsEnabledCheckBox.setText(tr("Notify in the taskbar when users in your buddy list connect"));
     animationGroupBox->setTitle(tr("Animation settings"));
+    animationsEnabledCheckBox.setText(tr("Enable game &animations"));
```
RickyRister: So you added a blanket game animations setting, but it's completely orthogonal to the tap/untap animation setting, which is also an animation that happens in the game.

I think we should just have separate settings for each animation. We already have tap/untap animation settings so we might as well go fully granular.

RickyRister:
The animation looks really wonky on my end.

https://github.com/user-attachments/assets/90bb6326-fb70-48e5-8ab3-d364e90f8ed3

There's a red dot tracing the arrow, but the dot and head of the arrow aren't synchronized.

output

cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
BruebachL: Reduced Motion/Accessibility would like us to have a blanket "off" toggle. I think we should do both blanket and granular.

RickyRister: If we want a blanket animation toggle, then shouldn't it also cover tap/untap?

BruebachL: Tangentally related to this PR but I've thought about it a bit and I think exposing a "Enable/Disable all" toggle in the settings that simply toggles all the checkboxes is preferable to exposing a new boolean we'd have to && on every animation setting

BruebachL:
> The animation looks really wonky on my end.
> example.mp4
> 
> There's a red dot tracing the arrow, but the dot and head of the arrow aren't synchronized.
> output

https://en.wikipedia.org/wiki/Twelve_basic_principles_of_animation

Specifically this is a secondary action for emphasis. I'll see if I can line them up better.

PR #7097, opened 2026-08-11
[Server/Client] Fix open_decklists not showing for already-loaded decks on join

### LINKED ISSUE #7029: Show all decks when lobby is in Open Decklists mode
When a lobby is set to Open Decklists mode players can only see the decklists of players who have loaded their deck after the player has joined. If another player has loaded a deck before, the new player won't be able to see that deck.

BruebachL:
## Related Ticket(s)
- Fixes #7029 

Wrong gate means the info the client needs is never populated

- Populate info when open_decklists is enabled
- Don't create tabs for users with no loaded decks

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
@@ -1631,7 +1631,9 @@ void Server_AbstractPlayer::getInfo(ServerInfo_Player *info,
 {
     getProperties(*info->mutable_properties(), withUserInfo);
 
-    if (deck) {
+    // Deck lists are only shared with other players when the game is in Open Decklists mode,
+    // so a player joining an open lobby can see every deck that was loaded before they joined.
+    if (deck && (recipient == this || game->getShareDecklistsOnLoad())) {
         info->set_deck_list(deck->writeToString_Native().toStdString());
     }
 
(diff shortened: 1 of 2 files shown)


libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.cpp
```diff
@@ -1631,7 +1631,9 @@ void Server_AbstractPlayer::getInfo(ServerInfo_Player *info,
 {
     getProperties(*info->mutable_properties(), withUserInfo);
 
-    if (deck) {
+    // Deck lists are only shared with other players when the game is in Open Decklists mode,
+    // so a player joining an open lobby can see every deck that was loaded before they joined.
+    if (deck && (recipient == this || game->getShareDecklistsOnLoad())) {
```
RickyRister, line 1636: wait, how did making the condition *more* restrictive cause something that didn't show before to show now?

BruebachL: It's not more restrictive, the addition is an or ||

RickyRister: isn't the `||` inside the bracketed part that is `&&`'d with the original condition?

BruebachL: Ah, sorry. There was a guard introduced on getInfo, which I already accidentally removed in this commit
https://github.com/Cockatrice/Cockatrice/pull/7030/changes#diff-7ce050383a9a78045aa8ba57a7691d8679862afc8177655168579505c667fb79

That *should've* fixed open_decklists but the server hasn't been rebuilt since.

(Which is why a user reported this again and my local investigation caused me to think "ah, this fixed it")

However, just dropping the guard isn't correct, this restores the gate on getShareDecklistsOnLoad()

RickyRister (approved):

PR #7098, opened 2026-08-11
[Game] Generic Animation Interface

BruebachL:
GameScene has a per-tick QTimer but no mechanism for hooking up anything but card taps to it.

- Implement generic interface
- Implement settings buttons to enable/disable all animations


RickyRister (approved):
I'm fine with the implementation. However, we might want to get a second pair of eyes on this since it's code that could easily lead to segfaults and I haven't traced it super deeply.

PR #7092, opened 2026-08-10
[Cards] Artist attribution

BruebachL:
## Short roundup of the initial problem
Scryfall asks that the artist attribution is visible or visible in the same interface whenever a card art crop is used.

## What will change with this Pull Request?
- Parse artist from MTGJson
- Draw artist in a little pill wherever art crops are used

Copyright attribution deferred until MTGJson provides a value for it.

## Screenshots
image
image

PR #7101, opened 2026-08-11
[Game] Playmats

BruebachL:
Adds playmats, a per-deck card whose art is rendered as the table background, covering the combined stack + table play area. Playmats are saved in the deck file, configurable from the deck editor with live preview, shared with opponents through player properties, and controllable via a new visibility setting.

Persistence:
- New PlaymatParams struct (marginPctL, marginPctR, verticalOffset, zoom) and a playmatCard metadata field on DeckList.
- DeckList serializes a new <playmatCard> element, with all parameters clamped to safe ranges so malformed deck files can't produce degenerate art rectangles.

Deck editor UI:
- New "Edit Playmat..." button in the deck editor dock (shows the selected card name once set), backed by undo/redo-aware setters in DeckStateManager.

In-game rendering:
- PlayerGraphicsItem paints the playmat card art (cover-fitted, cropped per PlaymatParams) behind the combined stack + table zones, clipping to that area.
- Table and stack zones draw subtle dark overlays when a playmat is active to keep the zones distinguishable. Theme brushes are restored when no playmat is shown.
- Art attribution is rendered on the playmat

Networking / protocol
- ServerInfo_PlayerProperties gains optional playmat fields (card name, provider id, and the four positioning params).
- Server sends the playmat in cmdDeckSelect and getPlayerProperties. The client's GameEventHandler applies them, and PlayerLogic stores the remote playmat for opponent display.

Settings
- New "Playmat visibility" option in Appearance settings: Show all playmats, Show own playmat only, Don't use playmats (defaults to show all).

## Screenshots
image
image


RickyRister:
Are you sure we want to model playmats as part of the decklist? I feel like attaching it to the decklist doesn't match how playmats are used in reality.

Like, just thinking about how playmats are used irl, people usually organize their playmats completely separately from their decks. They don't go "oh, for this deck, I'll always use this playmat". Instead, they'll bring one playmat and multiple commander decks to game night, and play each deck on the same mat. 

Or if they just got a cool new mat and want to show it off, they'll just bring the new mat and the same decks. Meanwhile on cockatrice, with this design, if you want to use a new mat, you need to change *every deck you want play* to use the new mat ahead of time.

I do not know of anyone who brings multiple playmats and decks to game night and decides *during deckbuilding* which mat will be used for which deck and switches out their mat whenever they bring out a new deck.

I really think playmat configuration should be orthogonal to deck configuration. 
I concede that it would be much more work to design, but I think it would match reality (as well as the user's intuition) better.

BruebachL:
You can now specify playmats on a per deck basis, as well as a global single override playmat and a list of fallbacks (fixed, random, round-robin) for decks without a playmat.

Resolution order is

Override > Per-deck > Fallbacks

cockatrice/src/game_graphics/player/player_graphics_item.cpp
```diff
@@ -155,10 +167,36 @@ qreal PlayerGraphicsItem::getMinimumWidth() const
...
+    double combinedRight = qMax(stackPos.x() + stackSize.width(), tablePos.x() + tableSize.width());
+    double combinedBottom = qMax(stackPos.y() + stackSize.height(), tablePos.y() + tableSize.height());
+
+    QRectF combinedArea(combinedLeft, combinedTop, combinedRight - combinedLeft, combinedBottom - combinedTop);
+
+    const QRectF srcRect = computeArtSourceRect(playmatPixmap.size(), playmatParams);
+    const QRectF dstRect = coverFitRect(combinedArea, srcRect.size());
+
+    painter->setClipRect(combinedArea);
+    painter->drawPixmap(dstRect, playmatPixmap, srcRect);
```
RickyRister: Really weird graphical bug when the battlefield is zoomed out enough and you draw selections on the right side of the battlefield. The selections will have entirely black fill, and sometimes the black squares don't get cleared immediately.

https://github.com/user-attachments/assets/0dbc0fe1-9e78-472a-bd02-a9fab57851f1

The bug only happens when playmats are present.

cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
```diff
@@ -317,10 +321,39 @@ AppearanceSettingsPage::AppearanceSettingsPage()
...
+
+    auto *overrideButtons = new QHBoxLayout;
+    overrideButtons->addWidget(&playmatOverrideEditButton);
+    overrideButtons->addWidget(&playmatOverrideClearButton);
+
     auto *tableGrid = new QGridLayout;
     tableGrid->addWidget(&invertVerticalCoordinateCheckBox, 0, 0, 1, 2);
     tableGrid->addWidget(&minPlayersForMultiColumnLayoutLabel, 1, 0, 1, 1);
     tableGrid->addWidget(&minPlayersForMultiColumnLayoutEdit, 1, 1, 1, 1);
+    tableGrid->addWidget(&playmatVisibilityLabel, 2, 0, 1, 1);
```
RickyRister: I feel like playmat settings should be in a new group instead of under `table grid layout` settings`

cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
```diff
@@ -489,4 +558,10 @@ void AppearanceSettingsPage::retranslateUi()
     tableGroupBox->setTitle(tr("Table grid layout"));
     invertVerticalCoordinateCheckBox.setText(tr("Invert vertical coordinate"));
     minPlayersForMultiColumnLayoutLabel.setText(tr("Minimum player count for multi-column layout:"));
-}
\ No newline at end of file
+    playmatVisibilityLabel.setText(tr("Playmat visibility:"));
+    playmatOverrideLabel.setText(tr("Force playmat for all decks:"));
```
RickyRister: Instead of having both a `global playmat override` and a `default playmat list`, I think it would be more intuitive to just have a `default playmat list` plus a setting that determines when the default playmat list is used (`Always` / `Fallback only` / `Never`)

cockatrice/src/game_graphics/player/player_graphics_item.cpp
BruebachL: I attempted a fix but I can't reproduce

RickyRister: Still not fixed

https://github.com/user-attachments/assets/57033425-502d-4e29-be00-c53ce146be44

Also, turns out the horizontal cutoff that causes this doesn't even need zooming. The cutoff is x > 1174 (Which I got by looking at the value of `_selectionOrigin` in `GameView::startRubberBand`.

https://github.com/user-attachments/assets/4c2964c8-136e-46d8-bec8-53930ee9067b

cockatrice/src/game_graphics/player/player_graphics_item.cpp
```diff
@@ -112,7 +124,6 @@ void PlayerGraphicsItem::initializeZones()
     rfgZoneGraphicsItem->setPos(base + QPointF(0, 2 * h + h2 + 10));
 
     tableZoneGraphicsItem = new TableZone(player->getTableZone(), mirrored, this);
-    connect(tableZoneGraphicsItem, &TableZone::sizeChanged, this, &PlayerGraphicsItem::updateBoundingRect);
```
RickyRister, line 115: Cockatrice now instantly crashes upon starting a 2 player local game

https://github.com/user-attachments/assets/1a96642f-3330-4875-9dcf-fb9859f56697

cockatrice/src/game_graphics/player/player_graphics_item.cpp
RickyRister: Fixed now. Thanks!

https://github.com/user-attachments/assets/61ed6f46-771f-4df9-9679-90dc73b75cc1

BruebachL: Awesome! Thanks for testing this on Mac!

libcockatrice_protocol/libcockatrice/protocol/pb/serverinfo_playerproperties.proto
```diff
@@ -11,4 +11,10 @@ message ServerInfo_PlayerProperties {
     optional sint32 ping_seconds = 7;
     optional bool sideboard_locked = 8;
     optional bool judge = 9;
+    optional string playmat_card_name = 10;
```
RickyRister: do we want to extract the playmat fields into its own proto message? We do that with `CardArtParams`

libcockatrice_deck_list/libcockatrice/deck_list/deck_list.h
```diff
@@ -70,6 +98,8 @@ class DeckList
         CardRef bannerCard;          ///< Optional representative card for the deck.
         QStringList tags;            ///< User-defined tags for deck classification.
         QString lastLoadedTimestamp; ///< Timestamp string of last load.
+        CardRef playmatCard;         ///< Optional card used as playmat background for table+stack zones.
```
RickyRister: I think it would make more sense to store all playmat info in a single struct, instead of being spread across two fields.

libcockatrice_deck_list/libcockatrice/deck_list/playmat_resolver.h
```diff
@@ -0,0 +1,54 @@
...
+    RoundRobin = 1, ///< Cycle through the list, advancing one entry per resolution.
+    Random = 2      ///< Pick a random entry per resolution.
+};
+
+/**
+ * @struct PlaymatResolution
+ * @ingroup Decks
+ * @brief A resolved playmat (card + positioning parameters).
+ */
+struct PlaymatResolution
```
RickyRister: There's currently like three different ways we're storing the playmap info (`CardRef` + `PlaymatParams` as separate fields in `DeckList`; `PlaymatResolution`; `StoredPlaymat`). Do you think it would be possible to consolidate it a bit more?

cockatrice/src/game_graphics/deckview/deck_view_container.cpp
```diff
@@ -277,7 +278,11 @@ void DeckViewContainer::loadDeckFromFile(const QString &filePath)
 
 void DeckViewContainer::loadDeckFromDeckList(const DeckList &deck)
 {
-    QString deckString = deck.writeToString_Native();
+    // Bake the user-level playmat resolution (override > deck > fallback list)
+    // into the uploaded deck so the server, local and remote players, and
+    // replays all derive the same effective playmat.
+    const DeckList effectiveDeck = applyUserPlaymatSettings(deck, SettingsCache::instance().userInterface());
```
RickyRister: Thinking on it more, are you *sure* we want to lock the playmat to be delivered to the server as part of the decklist xml? 

The problem with doing it like is that it makes it much harder to change in the future, since now the playmap is coupled with the decklist delivery. If the playmat was delivered separately, or even as a different field in the same proto, then it becomes easier to evolve forwards without affecting the decklist loading.

It also forces any playmat changes to be bundled with a deck load, which means you cannot change the playmat after loading your deck, or in the middle of a game.

cockatrice/src/game_graphics/deckview/deck_view_container.cpp
```diff
@@ -289,6 +297,65 @@ void DeckViewContainer::loadDeckFromDeckList(const DeckList &deck)
...
+    PlaymatResolution resolved;
+    QList<PlaymatResolution> fallbackList = settings.getPlaymatFallbackList();
+
+    // In random mode with 2+ entries, remove the last-resolved mat to avoid repeats.
+    if (fallbackBehavior == PlaymatFallbackMode::Random && fallbackList.size() > 1) {
+        fallbackList.removeAll(lastResolvedPlaymat);
+    }
+
+    switch (settings.getPlaymatMode()) {
+        case 0: { // Override deck playmat — always use collection
```
RickyRister: Do we want to create an enum for this

libcockatrice_utility/libcockatrice/utility/playmat_params.h
```diff
@@ -0,0 +1,52 @@
...
+        return !(*this == other);
+    }
+};
+
+/**
+ * @struct PlaymatResolution
+ * @ingroup Decks
+ * @brief A resolved playmat (card + positioning parameters).
+ */
+struct PlaymatResolution
```
RickyRister: Do we want to keep calling it `PlaymatResolution`, now that it's not just used for resolved playmats? Something like `PlaymatInfo` or `PlaymatData` would be simpler.

cockatrice/src/game_graphics/player/player_graphics_item.cpp
```diff
@@ -303,3 +365,85 @@ void PlayerGraphicsItem::updateBoundingRect()
 
     emit sizeChanged();
 }
+
+void PlayerGraphicsItem::updatePlaymat()
+{
+    int visibility = SettingsCache::instance().userInterface().getPlaymatVisibility();
```
ZeldaZach, line 372: `Show own playmat only` (value `1`) is never implemented. This is the only place `getPlaymatVisibility()` is read, and it only special-cases `0`, so `1` and `2` behave identically — selecting `Show own playmat only` in Appearance settings does nothing.

Something like:

```cpp
// "Show own playmat only"
if (visibility == 1 && !player->getPlayerInfo()->getLocal()) {
    clearPlaymat();
    return;
}
```

Given the point of the option is presumably to let people opt out of opponents' art, it's the one that most needs to work.

ZeldaZach (commented):
Nice feature, and the rendering path (bucketed rescale, cover-fit, clip to stack+table) looks well thought out. A few things I think are worth addressing before merge; I've skipped nits and left only the ones that look like real behavior/correctness issues.

Summary of the important ones:

1. `Show own playmat only` is never implemented — `updatePlaymat()` only branches on visibility `0`, so `1` and `2` behave identically.
2. `Server_Player::cmdSetPlaymat` accepts unbounded client strings and rebroadcasts them to every player in the game, unlike the equivalent card-art commands which check `MAX_NAME_LENGTH`.
3. The client-side fallback to `player->getDeck().getPlaymat()` defeats the resolution modes: whenever the server-confirmed playmat is empty, the deck's own playmat comes back.
4. `TableZone::setWidth()` now emits `sizeChanged()`, which creates mutual recursion with `GameScene::rearrange()`. The snapshot in `resizeColumnsAndPlayers` fixes iterator invalidation but not the recursion itself.
5. Artist attribution is a no-op for every existing user until they regenerate their card database.

cockatrice/src/game_graphics/player/player_graphics_item.cpp
```diff
@@ -303,3 +365,85 @@ void PlayerGraphicsItem::updateBoundingRect()
...
+
+    CardRef playmatCard;
+    PlaymatParams params;
+
+    if (player->getHasRemotePlaymat()) {
+        // Prefer the server-confirmed playmat (updated by Command_SetPlaymat).
+        playmatCard = player->getRemotePlaymatCard();
+        params = player->getRemotePlaymatParams();
+    } else {
+        // Fall back to the locally baked-in deck playmat.
```
ZeldaZach: This local fallback overrides the resolution modes.

`PlayerLogic::setDeck()` is only ever called for the local player (`TabGame::loadDeckForLocalPlayer`), with the decklist XML that carries `<playmatCard>`. So for the local player this branch is always live whenever `hasRemotePlaymat` is false.

But `hasRemotePlaymat == false` is exactly the signal that the server-confirmed playmat is *empty*, which is what you get when:

- playmat mode is `Override deck playmat` and the collection is empty,
- the user hit `Remove Playmat` on the deck's mat but the deck file the server holds still has it,
- the resolved playmat is intentionally none.

In all of those the deck's own playmat comes back on screen, and the resolution chain the user configured is ignored for their own board.

Related: `Server_AbstractPlayer::getPlayerProperties()` and `Server_Player::cmdDeckSelect()` both call `mutable_playmat_params()` unconditionally, so `has_playmat_params()` is true even for decks with no playmat. That's fine on the wire, but it means the client can't distinguish "server didn't tell me" from "server says none" — which is the state this fallback is trying to cover. Worth making that distinction explicit rather than falling back to the deck.

libcockatrice_network/libcockatrice/network/server/remote/game/server_player.cpp
```diff
@@ -594,6 +603,41 @@ Server_Player::cmdReverseTurn(const Command_ReverseTurn &cmd, ResponseContainer
...
+{
+    Q_UNUSED(rc);
+
+    if (!deck) {
+        return Response::RespContextError;
+    }
+
+    const auto &pp = cmd.playmat_params();
+    PlaymatResolution playmat;
+    playmat.card.name = QString::fromStdString(pp.card_name());
```
ZeldaZach: `card_name` and `card_provider_id` come straight from the client with no length validation, get stored on the server-side `deck`, and are then rebroadcast to every player in the game via `Event_PlayerPropertiesChanged`. There's also no rate limit on `SET_PLAYMAT`, so a client can spam multi-megabyte names and have the server amplify them to every opponent.

The numeric params are clamped, so this is only about the strings. The directly comparable commands — the card-art name rules in `servatrice/src/serversocketinterface.cpp` (see `cmdAddCardArtNameRule`, ~L1686) — take the same `cardName` + `cardProviderId` shape and reject with:

```cpp
if (cardName.length() > MAX_NAME_LENGTH || cardProviderId.length() > MAX_NAME_LENGTH) {
    return Response::RespInvalidData;
}
```

Worth matching that here, and in `cmdDeckSelect` since the decklist path can carry an equally long `<playmatCard>` name.

cockatrice/src/game_graphics/game_scene.cpp
```diff
@@ -459,8 +459,14 @@ void GameScene::resizeColumnsAndPlayers(const QList<qreal> &minWidthByColumn, qr
     qreal extraWidthPerColumn = (newWidth - minWidth) / playersByColumn.size();
     qreal newx = phasesToolbar->getWidth();
 
-    for (int col = 0; col < playersByColumn.size(); ++col) {
-        for (PlayerGraphicsItem *player : playersByColumn[col]) {
+    // Snapshot the columns: resizing a player's table can synchronously trigger
+    // GameScene::rearrange (table width -> sizeChanged -> updateBoundingRect ->
+    // sizeChanged -> rearrange), and rearrange rebuilds playersByColumn. Iterating
+    // the live container across that re-entrant call would use invalidated iterators.
+    const QList<QList<PlayerGraphicsItem *>> columns = playersByColumn;
```
ZeldaZach, line 476: The snapshot fixes the iterator invalidation, but the underlying re-entrancy is still there and is newly introduced by this PR. `TableZone::setWidth()` now emits `sizeChanged()`, which gives:

```
resizeColumnsAndPlayers -> processSceneSizeChange -> TableZone::setWidth
  -> sizeChanged -> PlayerGraphicsItem::updateBoundingRect
  -> sizeChanged -> GameScene::rearrange (game_scene.cpp:185)
  -> processViewSizeChange -> resizeColumnsAndPlayers -> ...
```

The only thing terminating this is `setWidth`'s new value-unchanged early return. That's load-bearing but not obvious, and it's fragile — anything that makes the computed width oscillate (rounding, a zone width that itself depends on the layout) turns it into unbounded recursion. It also means the recursive `rearrange` runs before `processSceneSizeChange` gets to line 236 (`handZoneGraphicsItem->setWidth(...)`), so the inner pass lays out with a stale hand width and the outer frame then overwrites it.

A re-entrancy guard in `rearrange()`/`processViewSizeChange()` (bail and set a "needs another pass" flag) would be more robust than relying on the early return, and would let you drop the snapshot.

cockatrice/src/game_graphics/game_view.cpp
```diff
@@ -128,7 +128,15 @@ void GameView::resizeRubberBand(const QPointF &cursorPoint, int selectedCount)
...
     QPoint cursor = cursorPoint.toPoint();
     QRect rect = QRect(mapFromScene(selectionOrigin), cursor).normalized();
+
     rubberBand->setGeometry(rect);
+    if (viewport()) {
+        // Repaint the whole viewport rather than just the band's old area: the
+        // vacated strip of a child widget is not reliably invalidated on all
+        // platforms (notably macOS), leaving stale pixels — including black
+        // scene background where the playmat should show — under the selection.
+        viewport()->update();
```
ZeldaZach: This repaints the entire viewport on every mouse-move event during a drag-select. Combined with `TableZone`/`StackZone` switching to `NoCache` while a playmat is active, that's a full re-render of every zone (plus a possible playmat rescale) per mouse move.

I know this was the fix for the black-fill artifact @RickyRister reported and that it does resolve it, so I'm not suggesting reverting — but can it be narrowed to the affected area? Keeping the previous band geometry and doing:

```cpp
viewport()->update(previousBandRect.united(rect).adjusted(-1, -1, 1, 1));
```

still repaints the strip the band vacated (which is the actual bug) without a full-scene repaint per event. `stopRubberBand()` could do the same with the last known rect.

cockatrice/src/interface/widgets/playmat/playmat_collection_dialog.cpp
```diff
@@ -0,0 +1,188 @@
...
+    editButton->setEnabled(hasSelection);
+    removeButton->setEnabled(hasSelection);
+    moveUpButton->setEnabled(hasSelection && playmatList->currentRow() > 0);
+    moveDownButton->setEnabled(hasSelection && playmatList->currentRow() < playmatList->count() - 1);
+}
+
+void PlaymatCollectionDialog::addPlaymat()
+{
+    PlaymatSettingsDialog dialog(CardRef{}, PlaymatParams{}, this);
+    if (dialog.exec() == QDialog::Accepted) {
```
ZeldaZach, line 111: `PlaymatSettingsDialog` shows an `Also set as my default playmat (for decks without one)` checkbox, but neither `addPlaymat()` nor `editPlaymat()` reads `dialog.useAsDefault()`, so ticking it inside the collection editor silently does nothing.

And if it did do something it would be wrong here — the deck-editor implementation of that option does `setPlaymatFallbackList({{newCard, newParams}})`, i.e. it replaces the entire collection with a single entry, which would wipe the list the user is currently editing.

Probably simplest to hide the checkbox when the dialog is opened from the collection editor (a constructor flag), since "add to my defaults" is what the collection dialog already does.

libcockatrice_card/libcockatrice/card/printing/printing_info.h
```diff
@@ -130,6 +130,18 @@ class PrintingInfo
...
     [[nodiscard]] QString getFlavorName() const;
+
+    /**
+     * @brief Returns the artist name credited for this printing's artwork.
+     *
+     * Requires a card database generated with artist data.
+     *
+     * @return The artist name, or empty if it isn't present.
+     */
+    [[nodiscard]] QString getArtist() const
```
ZeldaZach, line 141: Worth calling out in the PR description: the `artist` property only starts being written by `oracleimporter.cpp` in this PR, so `getArtist()` returns empty for every user whose `cards.xml` predates it. The attribution pill then doesn't render at all until they regenerate their card database.

If the attribution is here to satisfy Scryfall's requirement to credit the artist wherever art is shown cropped, it fails open for essentially the entire existing user base on day one — and this PR also adds the pill to the home tab, the user info popup, and the user list, which are all paths that already ship art crops today. At minimum that should be a documented consequence; ideally a card-database refresh gets prompted.

BruebachL: Sure, lemme just magically conjure data onto the user systems before they download anything.

Claude, please.

ZeldaZach (approved):

RickyRister:
The artist attribution PR (#7092) should be merged first

PR #7102, opened 2026-08-13
Toggle Normal Untapping Once

### LINKED ISSUE #2614: "Toggle normal untapping once" feature
I miss the ability to mark a card to just not untap normally **once**.
There is not just `exert` out there, there are also a lot of freeze effects in mostly all sets etc.

It could be directly beneath the permanent untap toggle option in the right click menu of a card.

Skagra42:
## Related Ticket(s)
- Fixes #2614.

Many cards only prevent normal untapping once.  Currently, the only options available on Cockatrice are always untapping normally or never doing so.

- Adds a button in the menu that appears when you right click a card to toggle normal untapping once.  If you turn this on for a card, its border will become cyan.  When you next untap all permanents, it will remain tapped and the option will be toggled off, allowing you to untap it next time.
- Adds the ability to create a shortcut to prevent normal untapping once.

## Screenshots

image
image
image


RickyRister:
Please rebase onto master

RickyRister:
Did you test the backwards compatibility? In other words, what happens if both a newer client and an older client are in a game and the newer client does the new action?

Skagra42:
> Did you test the backwards compatibility? In other words, what happens if both a newer client and an older client are in a game and the newer client does the new action?

How do I test with a non-local game?

RickyRister:
> How do I test with a non-local game?

Compile and run servatrice. By default it'll start a server on localhost. Then you can start up multiple instances of cockatrice and connect to it.

BruebachL:
The "toggle off" lags one untap cycle behind the intent.

When the server consumes the flag during a mass untap, it returns the current tapped state ("1" for the typical tapped creature), see the early-return in Server_Card::setAttribute(attr, value, allCards). So the broadcast event is AttrTapped = "1":

- Client handler: tapped == true → neither guard triggers, normal branch runs → the else if (!tapped && getDoesntUntapOnce() ...) clearing branch never executes.
- Result: card stays tapped (correct), but the cyan border and the checked menu item persist through the entire turn they were supposed to cover, and only clear at the next untap step (when the server genuinely untaps and sends "0").

- Asymmetry proof: if the card is untapped when consumed, the server returns "0", and the client clears immediately, so behavior differs depending on tap state.

- Degenerate combo: with both flags set, the server consumes the once-flag but the client's copy is permanently shielded by the permanent-flag guard → cyan border persists indefinitely alongside magenta.

tooomm:
@Skagra42 did you have a chance to review Bruebach's comments?

There have been quite some merges to master in the meantime, probably good to rebase your changes as well. 👍

PR #7103, opened 2026-08-13
[Git] Ignore docs/

BruebachL:
Huge autogenerated folder that clutters my untracked.


.gitignore
```diff
@@ -16,3 +16,4 @@ compile_commands.json
 .gdb_history
 cockatrice/resources/config/qtlogging.ini
 docs/
```
RickyRister, line 18: Isn't it already in the gitignore?

PR #7104, opened 2026-08-13
[Refactor] Decouple DeckFilterString from DeckPreviewWidget

BruebachL:
Pure refactor, no behavior change. The filter parser's predicates previously captured a DeckPreviewWidget * (reading deckLoader, filePath) and ExtraDeckSearchInfo. 

- They now operate on a small DeckSearchData struct (deck, filePath, relativeFilePath), removing the widget dependency from cockatrice/src/filters/. 

## Why?
This unblocks the upcoming model layer, which needs to run the same filters without a preview widget in scope.

### DIFF
--- a/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
@@ -73,14 +73,14 @@ static QString toRelativeFilepath(const QString &filePath)
 
 void VisualDeckStorageSearchWidget::filterWidgets(QList<DeckPreviewWidget *> widgets, const QString &searchText)
 {
-    auto filterString = DeckFilterString(searchText);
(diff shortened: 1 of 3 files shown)


cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
```diff
@@ -76,7 +76,11 @@ void VisualDeckStorageSearchWidget::filterWidgets(QList<DeckPreviewWidget *> wid
     auto filterString = DeckFilterString(searchText);
 
     for (auto widget : widgets) {
-        QString relativeFilePath = toRelativeFilepath(widget->filePath);
-        widget->filteredBySearch = !filterString.check(widget, {relativeFilePath});
+        DeckSearchData searchData;
```
RickyRister: nit: use designated initializer

RickyRister (approved):

PR #7105, opened 2026-08-13
[VDS] Add async-scanned model and sort/filter proxy

BruebachL:
## Related tickets

Based on:

- #7104

Adds the model layer for the visual deck storage tab, purely additive. No behavior change yet.

- VisualDeckStorageModel: asynchronously scans the deck storage folder and exposes rows (file path, deck, tags, banner card, color identity)
- VisualDeckStorageSortFilterProxyModel: sort and filter on top of the model

## Why?
The next PR rewires the folder/preview widgets to consume these.

### DIFF
--- a/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
+++ b/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
@@ -20,8 +20,8 @@ namespace
  */
 struct DeckScanResult
 {
-    QList<DeckPreviewData> decks;
-    QStringList folderPaths;
+    QList<DeckPreviewData> decks; ///< Deck rows in scan order.
+    QStringList folderPaths;      ///< Sorted list of subfolder paths.
 };
 
 /**
@@ -30,8 +30,8 @@ struct DeckScanResult
  */
 struct DeckLoadResult
 {
-    LoadedDeck deck;
-    QDateTime lastModified;
+    LoadedDeck deck;        ///< The parsed deck.
+    QDateTime lastModified; ///< File modification time at load.
 };
 
 /**
@@ -49,6 +49,63 @@ QString relativePathFromDeckRoot(const QString &path, const QString &deckPath)
     }
     return relativePath;
 }
+
+/**
+ * @brief The path of \a filePath relative to \a deckPath, or the bare file name
+ * if \a filePath is not below \a deckPath.
+ */
+QString relativeFilePathFor(const QString &filePath, const QString &deckPath)
+{
+    if (filePath.startsWith(deckPath)) {
+        return filePath.mid(deckPath.length());
+    }
+
+    return QFileInfo(filePath).fileName();
+}
+
+/**
+ * @brief The directory of \a filePath relative to \a deckPath, or empty if the
+ * file sits directly in the deck root.
+ */
+QString folderPathFor(const QString &filePath, const QString &deckPath)
+{
+    return relativePathFromDeckRoot(QFileInfo(filePath).absolutePath(), deckPath);
+}
+
+/**
+ * @brief Scans a deck directory on a worker thread, returning discovered deck
+ * files and subfolder paths.
+ */
+DeckScanResult scanDeckDirectory(const QString &deckPath)
+{
+    DeckScanResult result;
+
+    QDirIterator fileIt(deckPath, DeckLoader::ACCEPTED_FILE_EXTENSIONS, QDir::Files,
+                        QDirIterator::Subdirectories | QDirIterator::FollowSymlinks);
+    while (fileIt.hasNext()) {
+        const QString filePath = fileIt.next();
+        DeckPreviewData data;
+        data.filePath = filePath;
+        data.relativeFilePath = relativeFilePathFor(filePath, deckPath);
+        data.folderPath = folderPathFor(filePath, deckPath);
+        data.lastModified = QFileInfo(filePath).lastModified();
+        result.decks.append(std::move(data));
+    }
+
+    QSet<QString> seenFolders;
+    QDirIterator folderIt(deckPath, QDir::Dirs | QDir::NoDotAndDotDot,
+                          QDirIterator::Subdirectories | QDirIterator::FollowSymlinks);
+    while (folderIt.hasNext()) {
+        const QString folderPath = relativePathFromDeckRoot(folderIt.next(), deckPath);
+        if (!folderPath.isEmpty() && !seenFolders.contains(folderPath)) {
+            seenFolders.insert(folderPath);
+            result.folderPaths.append(folderPath);
+        }
+    }
+    result.folderPaths.sort();
+
+    return result;
+}
 } // namespace
 
 VisualDeckStorageModel::VisualDeckStorageModel(QObject *parent) : QAbstractListModel(parent)
@@ -57,16 +114,16 @@ VisualDeckStorageModel::VisualDeckStorageModel(QObject *parent) : QAbstractListM
 
 int VisualDeckStorageModel::rowCount(const QModelIndex &parent) const
 {
-    return parent.isValid() ? 0 : decks_.size();
+    return parent.isValid() ? 0 : decks.size();
 }
 
 QVariant VisualDeckStorageModel::data(const QModelIndex &index, int role) const
 {
-    if (!index.isValid() || index.row() < 0 || index.row() >= decks_.size()) {
+    if (!index.isValid() || index.row() < 0 || index.row() >= decks.size()) {
         return {};
     }
 
-    const DeckPreviewData &data = decks_.at(index.row());
+    const DeckPreviewData &data = decks.at(index.row());
     switch (role) {
         case Qt::DisplayRole:
         case VisualDeckStorageRoles::DisplayNameRole:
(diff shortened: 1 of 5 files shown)


cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.h
```diff
@@ -0,0 +1,158 @@
...
+    QString relativeFilePath;     ///< File path relative to the deck folder.
+    QString folderPath;           ///< Directory relative to the deck folder ("" for the deck folder itself).
+    QString deckName;             ///< The deck name as stored in the file (may be empty).
+    QString displayName;          ///< Deck name, or the file name if the deck has no name.
+    QStringList tags;             ///< The deck's tags.
+    QString colorIdentity;        ///< The deck's color identity in WUBRG order.
+    QDateTime lastModified;       ///< File modification time at last check.
+    QDateTime lastLoaded;         ///< When the deck was last loaded from the file.
+    QString bannerCardName;       ///< Name of the deck's banner card.
+    QString bannerCardProviderId; ///< Provider id of the deck's banner card.
```
RickyRister: Is there a reason we can't store them together as a `CardRef`?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.h
```diff
@@ -0,0 +1,158 @@
...
+
+private:
+    void startScan();
+    void beginLoad(int row);
+    void computeDeckMetadata(DeckPreviewData &data) const;
+    void setFilePathForRow(int row, const QString &newFilePath);
+    [[nodiscard]] static QString relativeFilePathFor(const QString &filePath, const QString &deckPath);
+    [[nodiscard]] static QString folderPathFor(const QString &filePath, const QString &deckPath);
+
+    QString deckPath_;
```
RickyRister: why the trailing underscores? That's not a style we've used anywhere else in the project.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -0,0 +1,490 @@
...
+#include <QSet>
+#include <QtConcurrentRun>
+#include <libcockatrice/card/database/card_database_manager.h>
+#include <utility>
+
+namespace
+{
+/**
+ * @brief The result of a background directory scan: the deck rows in scan order
+ * plus the sorted list of subfolder paths.
```
RickyRister, line 19: nit: if you're going to describe the fields, then you might as well do it in the `///<`

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -0,0 +1,490 @@
...
+    watcher->setFuture(QtConcurrent::run([deckPath]() -> DeckScanResult {
+        DeckScanResult result;
+
+        QDirIterator fileIt(deckPath, DeckLoader::ACCEPTED_FILE_EXTENSIONS, QDir::Files,
+                            QDirIterator::Subdirectories | QDirIterator::FollowSymlinks);
+        while (fileIt.hasNext()) {
+            const QString filePath = fileIt.next();
+            DeckPreviewData data;
+            data.filePath = filePath;
+            data.relativeFilePath = VisualDeckStorageModel::relativeFilePathFor(filePath, deckPath);
```
RickyRister: redundant qualifier

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_sort_filter_proxy_model.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    searchMatches_.resize(count);
+    DeckFilterString filterString(searchText_);
+    for (int row = 0; row < count; ++row) {
+        const DeckPreviewData &data = source->dataForRow(row);
+
+        // Decks that haven't finished loading yet can't be evaluated; show them and
+        // re-evaluate once the load finishes (reapplyFilters is triggered by deckLoaded).
+        if (data.deck.isEmpty()) {
```
RickyRister, line 180: shouldn't this check `loadInProgress` instead?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_sort_filter_proxy_model.cpp
```diff
@@ -0,0 +1,288 @@
...
+            searchMatches_[row] = true;
+            continue;
+        }
+
+        DeckSearchData searchData;
```
RickyRister: nit: use designated initializer

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_sort_filter_proxy_model.cpp
```diff
@@ -0,0 +1,288 @@
...
+    if (selectedTags_.isEmpty() && excludedTags_.isEmpty()) {
+        tagMatches_.fill(true);
+        return;
+    }
+
+    for (int row = 0; row < count; ++row) {
+        const QStringList deckTags = source->dataForRow(row).tags;
+
+        bool hasAllSelected = true;
+        for (const QString &tag : selectedTags_) {
```
RickyRister: nit: you can use `std::all_of` and `std::none_of` here instead of for loops

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_sort_filter_proxy_model.cpp
```diff
@@ -0,0 +1,288 @@
...
+{
+    const auto *source = deckSourceModel();
+    if (!source) {
+        searchMatches_.clear();
+        return;
+    }
+
+    const int count = source->rowCount();
+    if (searchText_.isEmpty()) {
+        searchMatches_ = QList<bool>(count, true);
```
RickyRister: the other two updateMatches methods resize the bool list, and then fill it with true. Meanwhile this method directly assigns a QList. Is there a reason for the difference?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_sort_filter_proxy_model.cpp
```diff
@@ -0,0 +1,288 @@
...
+                QSet<QChar> colorIdentitySet;
+                for (const QChar &color : colorIdentity) {
+                    colorIdentitySet.insert(color.toUpper());
+                }
+
+                matches = activeColorSet == colorIdentitySet;
+                break;
+            }
+            case Includes:
+                for (const QChar &color : activeColors_) {
```
RickyRister: nit: same comment about being able to use `std::all_of` and `std::none_of`

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -0,0 +1,490 @@
...
+    return -1;
+}
+
+void VisualDeckStorageModel::startScan()
+{
+    ++scanGeneration_;
+    beginResetModel();
+    decks_.clear();
+    folderPaths_.clear();
+    endResetModel();
```
RickyRister, line 199: The docs on `setDeckPath` say "Emits modelReset once the scan finishes and deck loads begin" However, doesn't calling `endResetModel` here cause `modelReset` to emit here?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -0,0 +1,490 @@
...
+            data.folderPath = VisualDeckStorageModel::folderPathFor(filePath, deckPath);
+            data.lastModified = QFileInfo(filePath).lastModified();
+            result.decks.append(std::move(data));
+        }
+
+        QDirIterator folderIt(deckPath, QDir::Dirs | QDir::NoDotAndDotDot,
+                              QDirIterator::Subdirectories | QDirIterator::FollowSymlinks);
+        while (folderIt.hasNext()) {
+            const QString folderPath = relativePathFromDeckRoot(folderIt.next(), deckPath);
+            if (!folderPath.isEmpty() && !result.folderPaths.contains(folderPath)) {
```
RickyRister: note: `result.folderPaths.contains` is O(n) and being called in a loop, which might end up mattering if there's some *really* deeply nested folders. I doubt most users will ever have enough folders for this to be noticeable, but if we're really paranoid, we could store the folderPaths in a set, and then copy them over to the list once we're done.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -0,0 +1,490 @@
...
+        beginInsertRows(QModelIndex(), 0, result.decks.size() - 1);
+        decks_ = result.decks;
+        endInsertRows();
+
+        for (int row = 0; row < decks_.size(); ++row) {
+            beginLoad(row);
+        }
+    });
+
+    watcher->setFuture(QtConcurrent::run([deckPath]() -> DeckScanResult {
```
RickyRister: nit: the code in the lambda also feels like a something that can be extracted to a named function. Just so that when someone is scanning for the code that handles discovering the files, they have a named function to find.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+
+    return colorIdentity;
+}
+
+/**
+ * @brief Recomputes all derived metadata of a row from its loaded deck.
+ */
+void VisualDeckStorageModel::computeDeckMetadata(DeckPreviewData &data) const
```
RickyRister: nit: can be static.
Also, perhaps `recomputeDeckMetdata` better describes how this method should be used.

RickyRister (approved):
Seems fine in general

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_sort_filter_proxy_model.cpp
BruebachL: If a deck fails to load, loadInProgress becomes false but deck.isEmpty() remains true. Using loadInProgress alone would let failed (empty) decks reach DeckFilterString::check, which requires a non-empty deck in DeckSearchData. The isEmpty() check handles both "still loading" and "failed to load" cases.

PR #7106, opened 2026-08-13
[VDS] Drive folder and preview widgets from the model

BruebachL:
## Related tickets

Based on:

- #7105 
- #7104

Converts the visual deck storage UI to the model/proxy introduced in the previous #7105 

- The folder display, deck preview, search, sort, and tag-filter widgets now read from VisualDeckStorageModel/VisualDeckStorageSortFilterProxyModel instead of directly listing files, so filtering/sorting is consistent and cheap

### DIFF
--- a/cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/deck_preview_color_identity_filter_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/deck_preview_color_identity_filter_widget.cpp
@@ -1,12 +1,12 @@
 #include "deck_preview_color_identity_filter_widget.h"
 
 #include "../../cards/additional_info/mana_symbol_widget.h"
-#include "deck_preview_widget.h"
+#include "../visual_deck_storage_widget.h"
 
-#include <QMouseEvent>
+#include <QSet>
 
 DeckPreviewColorIdentityFilterWidget::DeckPreviewColorIdentityFilterWidget(VisualDeckStorageWidget *parent)
-    : QWidget(parent), layout(new QHBoxLayout(this))
+    : QWidget(parent), layout(new QHBoxLayout(this)), visualDeckStorageWidget(parent)
 {
     setLayout(layout);
     layout->setSpacing(5);
@@ -32,10 +32,6 @@ DeckPreviewColorIdentityFilterWidget::DeckPreviewColorIdentityFilterWidget(Visua
 
     // Connect the button's clicked signal
     connect(toggleButton, &QPushButton::clicked, this, &DeckPreviewColorIdentityFilterWidget::updateFilterMode);
-    connect(this, &DeckPreviewColorIdentityFilterWidget::activeColorsChanged, parent,
-            &VisualDeckStorageWidget::updateColorFilter);
-    connect(this, &DeckPreviewColorIdentityFilterWidget::filterModeChanged, parent,
-            &VisualDeckStorageWidget::updateColorFilter);
 
     // Call retranslateUi to set the initial text
     retranslateUi();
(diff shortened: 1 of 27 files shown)


cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/deck_preview_deck_tags_display_widget.h
```diff
@@ -2,41 +2,53 @@
...
+class FlowWidget;
 
-class DeckPreviewWidget;
 class DeckPreviewDeckTagsDisplayWidget : public QWidget
 {
     Q_OBJECT
 
     QStringList currentTags;
     FlowWidget *flowWidget;
+    std::function<QStringList()> knownTagsProvider_;
```
RickyRister: why underscore postfix?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.h
```diff
@@ -37,39 +38,52 @@ class VisualDeckStorageWidget final : public QWidget
...
     VisualDeckStorageSearchWidget *searchWidget;
     DeckPreviewColorIdentityFilterWidget *deckPreviewColorIdentityFilterWidget;
     QToolButton *refreshButton;
     VisualDeckStorageQuickSettingsWidget *quickSettingsWidget;
     QScrollArea *scrollArea;
-    VisualDeckStorageFolderDisplayWidget *folderWidget;
-
-    void reapplySortAndFilters();
+    VisualDeckStorageFolderDisplayWidget *folderWidget = nullptr;
+    VisualDeckStorageModel *model_ = nullptr;
```
RickyRister: why underscore postfix

cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/deck_preview_color_identity_filter_widget.cpp
```diff
@@ -45,111 +41,56 @@ void DeckPreviewColorIdentityFilterWidget::retranslateUi()
...
+{
+    QSet<QChar> activeColorSet;
+    for (auto it = activeColors.constBegin(); it != activeColors.constEnd(); ++it) {
+        if (it.value()) {
+            activeColorSet.insert(it.key());
+        }
+    }
+
+    // The VDS owns the proxy model, which exists for the widget's whole lifetime.
+    visualDeckStorageWidget->proxyModel()->setColorFilter(filterMode, activeColorSet);
```
RickyRister: Previously, this widget emitted a signal that the `visualDeckStorageWidget` then connects to. Are you sure we want to directly reach into `visualDeckStorageWidget` from this class now?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
```diff
@@ -40,47 +38,6 @@ VisualDeckStorageSearchWidget::VisualDeckStorageSearchWidget(VisualDeckStorageWi
...
-    for (auto widget : widgets) {
-        const DeckSearchData searchData{.deck = &widget->deckLoader->getDeck(),
-                                        .filePath = widget->filePath,
-                                        .displayName = widget->getDisplayName(),
-                                        .relativeFilePath = toRelativeFilepath(widget->filePath)};
-
-        widget->filteredBySearch = !filterString.check(searchData);
-    }
+    connect(searchDebounceTimer, &QTimer::timeout, this,
+            [this] { this->parent->proxyModel()->setSearchText(searchBar->text()); });
```
RickyRister: Previously, this widget emitted a signal that the `visualDeckStorageWidget` then connects to. Are you sure we want to directly reach into `visualDeckStorageWidget` from this class now?

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_folder_display_widget.cpp
```diff
@@ -35,192 +44,214 @@ VisualDeckStorageFolderDisplayWidget::VisualDeckStorageFolderDisplayWidget(
...
-                &VisualDeckStorageWidget::openDeckEditor);
-        connect(visualDeckStorageWidget->settings(), &VisualDeckStorageQuickSettingsWidget::cardSizeChanged,
-                display->bannerCardDisplayWidget, &CardInfoPictureWidget::setScaleFactor);
-        display->bannerCardDisplayWidget->setScaleFactor(visualDeckStorageWidget->settings()->getCardSize());
-        allDecks.append(display);
+    // Drop previews of decks that no longer match.
+    for (auto it = deckWidgets.begin(); it != deckWidgets.end();) {
+        if (!desiredSet.contains(it.key())) {
+            flowWidget->removeWidget(it.value());
+            it.value()->deleteLater();
```
RickyRister, line 161: We're now creating/destroying entire `DeckPreviewWidget`s on every search instead of just hiding/unhiding them.  

Using search now causes cockatrice to chug, whereas previously it was near-instantaneous. In fact, it actually causes the entirety of cockatrice to freeze because widget creation happens on the ui thread.

https://github.com/user-attachments/assets/ab8e71f0-95f8-47db-96e5-1364d49ca7ba

I do not think this is an acceptable compromise for slightly more responsive initial loading.

RickyRister (changes requested):

BruebachL:
Profiling showed every VisualDeckStorageFolderDisplayWidget ran its whole reconcile synchronously in its constructor, and since subfolder widgets are created by their parent's reconcile, the entire folder tree built itself inside a single event loop turn (~1.7 s at 464 decks).

Newest commit now:
- No longer run the initial reconcile synchronously in the folder widget constructor, schedules it through the existing debounce timer instead
- Split reconciling into a resumable pass that creates deck previews until a ~20 ms per-turn time budget runs out, yielding to the event loop between chunks via queued continuations
- Restart the scan from a clean slate when the model changes mid-pass, so the pass always converges on the latest model state
- Move stale-widget cleanup and proxy-order relayout to the end of the pass, keeping the skip-relayout optimization for data-only updates
- Report folders as having content while their first pass is still building, so nothing flickers or hides prematurely before its previews exist

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_folder_display_widget.cpp
```diff
@@ -35,192 +45,297 @@ VisualDeckStorageFolderDisplayWidget::VisualDeckStorageFolderDisplayWidget(
...
+            continue;
+        }
+        deckPassPresentPaths.insert(data.filePath);
+
+        DeckPreviewWidget *deckPreviewWidget = deckWidgets.value(data.filePath, nullptr);
+        if (!deckPreviewWidget) {
+            deckPreviewWidget = createDeckPreviewWidget(data.filePath);
+        }
+
+        const bool accepted = proxy->mapFromSource(model->index(row, 0)).isValid();
```
RickyRister: Do we even make use of the filtering capabilities of the proxy model now? 
It feels really awkward to have to reach in to the underlying model and compare it to the proxy model to figure out what has been filtered out. Almost like it defeats the point of the proxy's filtering in the first place.

Is it possible for a proxy model to add a field to a row? 
Perhaps we could make the proxy model add a field to indicate if the row is filtered, instead of dropping the row entirely. Then we won't have to also reach into the underlying model.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_model.cpp
```diff
@@ -32,8 +32,18 @@ struct DeckLoadResult
...
 };
 
+/**
+ * @brief How many finished background deck loads one event loop turn applies.
+ *
+ * Applying a load emits dataChanged and wakes up the proxies, which is not
+ * free, and the bound keeps a burst of simultaneous completions from stalling
+ * the UI thread.
+ */
+constexpr int DECK_LOADS_PER_TURN = 24;
```
RickyRister: Where did you get the number from? 24 feels a bit low.

RickyRister (changes requested):
Really weird bug when refreshing the VDS while `show folders` is off.

https://github.com/user-attachments/assets/4551a9a5-8d2a-47f1-92b3-43e5d96f0b5d

RickyRister (changes requested):
Still isn't fixed

https://github.com/user-attachments/assets/94f17c73-678a-46ec-8fbd-213d4e45e883

BruebachL:
> Really weird bug when refreshing the VDS while `show folders` is off.

Try again

RickyRister (changes requested):
Scroll area spacing is broken. There is a large amount of empty space below each folder.

https://github.com/user-attachments/assets/f72df47a-3f47-4ce0-9af9-6d3e04aa290e

BruebachL:
> Scroll area spacing is broken. There is a large amount of empty space below each folder.
> example.mp4

Turns out my earlier fix attempts were not side-effect free.

RickyRister (approved):
A bit janky when I try to scroll while the decks are still loading, but it works now.

https://github.com/user-attachments/assets/3d2dbc1a-c91e-4e84-b237-6c6c48e8d77b

I like how the scroll bar now shrinks as the decks load.

PR #7110, opened 2026-08-13
[ThemeManager] Better default style guard

BruebachL:
Introduce helper to always pass default style name through the guard

### DIFF
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -90,13 +90,17 @@ struct PaletteColorInfo
     }
 }
 
+static QString usableDefaultStyle(const QString &style)
+{
+    // The Windows 11 native style is broken: when the OS default
+    // ("Default" theme selection) would use it, fall back to the Vista style.
+    // Explicitly choosing "windows11" in a theme is still honored.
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/theme_manager.cpp
```diff
@@ -316,13 +320,13 @@ void ThemeManager::applyStyleAndPalette(const QString &themeName,
         if (themeName == FUSION_THEME_NAME) {
             styleName = "Fusion";
         } else {
-            styleName = defaultStyleName;
+            styleName = usableDefaultStyle(defaultStyleName);
```
RickyRister, line 323: is `defaultStyleName` set anywhere other than in the constructor? You don't need to use `usableDefaultStyle` again if the `defaultStyleName` is already guarded. Unless you're super paranoid or something, I guess.

RickyRister (approved):

cockatrice/src/interface/theme_manager.cpp
BruebachL: Honestly, I thought the same thing but users are still reporting getting Windows11 as the default so... I'm throwing whatever I can think of at it.

PR #7111, opened 2026-08-13
[Completer] Minimum size popup, fall back to compact design if not possible.

BruebachL:
The completer popup is clamped to the LineEdit, which is sensible but breaks on very low width search bars.
image

- Minimum size the popup for the width the rows are designed for
- Switch to a compact layout if that's not possible.

### DIFF
--- a/cockatrice/src/interface/widgets/utility/card_completer_styler.cpp
+++ b/cockatrice/src/interface/widgets/utility/card_completer_styler.cpp
@@ -169,9 +169,12 @@ void CardCompleterStyler::applyPopupWidth()
         popup->resize(minWidth, popup->height());
 
         if (!available.isEmpty()) {
-            QPoint pos = popup->pos();
-            pos.setX(qBound(available.left(), pos.x(), qMax(available.left(), available.right() - popup->width() + 1)));
-            popup->move(pos);
+            // Clamp in screen coordinates, then map back to parent coordinates.
+            QPoint globalPos = popup->mapToGlobal(QPoint(0, 0));
+            int clampedGlobalX =
+                qBound(available.left(), globalPos.x(), qMax(available.left(), available.right() - popup->width() + 1));
+            QPoint parentPos = popup->mapFromGlobal(QPoint(clampedGlobalX, globalPos.y()));
+            popup->move(parentPos);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


RickyRister:
Can you update the description with screenshots?

cockatrice/src/interface/widgets/utility/card_completer_styler.cpp
```diff
@@ -140,6 +144,40 @@ bool CardCompleterStyler::handlePopupKeyPress(QKeyEvent *event)
...
+    // popup that wide, clamped to the screen it is shown on.
+    int contentWidth = CardCompleterDelegate::PopupRowWidth + popup->frameWidth() * 2;
+    if (auto *vbar = popup->verticalScrollBar(); vbar != nullptr) {
+        contentWidth += vbar->sizeHint().width();
+    }
+
+    const QScreen *screen = popup->screen();
+    const QRect available = screen ? screen->availableGeometry() : QRect();
+
+    const int minWidth = available.isEmpty() ? contentWidth : qMin(contentWidth, available.width());
```
RickyRister, line 163: The popup straight-up goes offscreen now

https://github.com/user-attachments/assets/04c93448-de95-4745-aab3-0e2141b4bc77

cockatrice/src/interface/widgets/utility/card_completer_styler.cpp
```diff
@@ -140,6 +144,43 @@ bool CardCompleterStyler::handlePopupKeyPress(QKeyEvent *event)
...
 
+void CardCompleterStyler::applyPopupWidth()
+{
+    auto *popup = completer->popup();
+
+    // QCompleter sizes its popup to the line edit, so a narrow search field
+    // would crush the rows. Ask for the width the rows are designed for (plus
+    // the frame and scrollbar) and let QWidget's geometry clamping keep the
+    // popup that wide, clamped to the screen it is shown on.
+    int contentWidth = CardCompleterDelegate::PopupRowWidth + popup->frameWidth() * 2;
```
RickyRister, line 155: The popup feels way too wide now.

Image

Image

Image

Like, the only time you'll need something this wide is if there's an un-card

Image

And the existing behavior already handles cutting names off.

Image

PR #7112, opened 2026-08-13
[DeckEditor] Reuse the inherited CardDatabaseModel in the visual editor

BruebachL:
The visual deck editor built three full CardDatabaseModels per tab: 
the one inherited from AbstractTabDeckEditor, a duplicate local model for the tab container, and a third one for the search completer. Each construction iterates the whole database, which dominated tab creation.

Reuse the inherited databaseModel everywhere instead:
- TabDeckEditorVisual passes the inherited model to the tab widget.
- VisualDeckEditorWidget takes the model as a constructor parameter and uses it as the completer source, falling back to creating its own if none is passed.

### DIFF
--- a/cockatrice/src/interface/widgets/visual_deck_editor/visual_deck_editor_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_deck_editor/visual_deck_editor_widget.cpp
@@ -29,8 +29,10 @@
 
 VisualDeckEditorWidget::VisualDeckEditorWidget(QWidget *parent,
                                                DeckListModel *_deckListModel,
-                                               QItemSelectionModel *_selectionModel)
-    : QWidget(parent), deckListModel(_deckListModel), selectionModel(_selectionModel)
+                                               QItemSelectionModel *_selectionModel,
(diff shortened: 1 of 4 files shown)


cockatrice/src/interface/widgets/visual_deck_editor/visual_deck_editor_widget.cpp
```diff
@@ -94,7 +96,9 @@ void VisualDeckEditorWidget::initializeSearchBarAndCompleter()
     setFocusProxy(searchBar);
     setFocusPolicy(Qt::ClickFocus);
 
-    cardDatabaseModel = new CardDatabaseModel(CardDatabaseManager::getInstance(), false, this);
+    if (!cardDatabaseModel) {
+        cardDatabaseModel = new CardDatabaseModel(CardDatabaseManager::getInstance(), false, this);
```
RickyRister, line 100: is there a reason the fallback `CardDatabaseModel` creation happens here instead of in the constructor?

RickyRister (approved):
Nice optimization. I was also thinking that CardDatabaseModel reuse was something we should be looking into.

PR #7113, opened 2026-08-13
[Refactor] Batch CardDatabaseModel enabled-sets rebuild into one model reset

BruebachL:
The per-card beginInsertRows/endInsertRows path in cardDatabaseEnabledSetsChanged() was the dominant cost of constructing a CardDatabaseModel over a full card database (tens of thousands of cards).

- Rebuild the whole list inside a single beginResetModel/endResetModel block instead, keeping the previous ordering semantics (cards already shown keep their relative order, newly-eligible cards are appended in database order)
and only connecting/disconnecting cardInfoChanged for cards that actually enter or leave the view.

### DIFF
--- a/libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
+++ b/libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
@@ -111,19 +111,50 @@ bool CardDatabaseModel::checkCardHasAtLeastOneEnabledSet(const CardInfoPtr &card
 
 void CardDatabaseModel::cardDatabaseEnabledSetsChanged()
 {
-    // remove all the cards no more present in at least one enabled set
+    // Build the new card list in a single pass.
+    QList<CardInfoPtr> newCardList;
+    newCardList.reserve(cardList.size());
     for (const CardInfoPtr &card : cardList) {
-        if (!checkCardHasAtLeastOneEnabledSet(card)) {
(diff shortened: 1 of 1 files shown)


libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
```diff
@@ -111,19 +111,50 @@ bool CardDatabaseModel::checkCardHasAtLeastOneEnabledSet(const CardInfoPtr &card
 
 void CardDatabaseModel::cardDatabaseEnabledSetsChanged()
 {
-    // remove all the cards no more present in at least one enabled set
+    // Build the new card list in a single pass.
+    QList<CardInfoPtr> newCardList;
+    newCardList.reserve(cardList.size());
     for (const CardInfoPtr &card : cardList) {
-        if (!checkCardHasAtLeastOneEnabledSet(card)) {
-            cardRemoved(card);
```
RickyRister, line 117: does it mean these signals are no longer being emitted when the database resets?

RickyRister (approved):
seems fine

libcockatrice_models/libcockatrice/models/database/card_database_model.cpp
BruebachL: They're still emitted, just as a coarser version

PR #7114, opened 2026-08-13
[CardDB] Cache CardDatabaseQuerier count maps

BruebachL:
getAllMainCardTypesWithCount(), getAllSubCardTypesWithCount() and getAllFormatsWithCount() each iterate the whole card database and are queried once per filter widget build. 

- Cache the results and invalidate on cardAdded/cardRemoved/cardDatabaseReset, which covers every database mutation (loading, swapping, clearing).


RickyRister (approved):
Nice optimization

PR #7115, opened 2026-08-13
[VDD] Defer heavy construction until after the tab paints

BruebachL:
Opening the Visual Database Display tab blocked synchronously on two full-database passes: building the filter toolbar (per-set, per-main-type, per-sub-type and per-format buttons) and creating the first 100-card page.

Both now run via QTimer::singleShot(0, ...) after the tab is shown for the first time, so the tab switch paints immediately and the content settles one event loop turn later.

The filter toolbar build is gated on first show (hidden tabs such as the deck editor startup path never pay the cost) and the toolbar stays hidden until it is fully built so no layout shift occurs after paint. The initial card page load is gated by currentPage == 0. Scrolling and search-driven loads remain immediate.

### DIFF
--- a/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.cpp
@@ -163,7 +163,7 @@ void VisualDatabaseDisplayWidget::initialize()
 
 void VisualDatabaseDisplayWidget::initializeFilters()
 {
-    if (filtersInitialized || !isVisible()) {
+    if (filtersInitialized || !isVisible() || CardDatabaseManager::getInstance()->getLoadStatus() != LoadStatus::Ok) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.cpp
```diff
@@ -158,17 +156,45 @@ void VisualDatabaseDisplayWidget::initialize()
...
+    // build to the next event loop turn, letting the tab paint first.
+    filterContainer->setVisible(false);
+
+    QTimer::singleShot(0, this, [this] {
+        filterContainer->initialize();
+        filterContainer->setVisible(true);
+
+        databaseDisplayModel->setFilterTree(filterModel->filterTree());
+
+        loadCardsTimer = new QTimer(this);
```
RickyRister: nit: can't you just use [`QTimer::singleshot`](https://doc.qt.io/qt-6/qtimer.html#singleShot-2)?

RickyRister (approved):

PR #7116, opened 2026-08-13
[VDD] Parent filter toolbar layouts to their group boxes

BruebachL:
Fix a layout warning


RickyRister (approved):

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_filter_toolbar_widget.cpp
```diff
@@ -82,17 +82,14 @@ VisualDatabaseDisplayFilterToolbarWidget::VisualDatabaseDisplayFilterToolbarWidg
 
 void VisualDatabaseDisplayFilterToolbarWidget::initialize()
 {
-    // create groupbox layouts
```
RickyRister, line 85: why did you get rid of this comment though?

BruebachL: It didn't seem super informative tbh (i.e. describe why you're doing something, not what you're doing in comments)

PR #7121, opened 2026-08-15
[Tabs] Add a setting to define startup tab on application launch

BruebachL:
For new users, the home tab was a clear win: Big, bold, call-to-action buttons that clearly show the capabilities of the application.
For power users, the home tab was a net-negative since it introduced a layer of indirection between them and a functionality they were used to (the deck editor).

Was always opening up on the deck editor the best idea? Not *especially* for the reasons described in the original home tab PR: It's simply not 100% the intent of a user to create or edit a deck.

But choice is good, especially for power-users or new users who are diving a bit deeper into customizing their experience.

- Add a new setting to define which tab should be opened after application launch
- Add some intent machinery to resolve a room join on startup

### DIFF
--- a/cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
+++ b/cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
@@ -162,6 +162,9 @@ GeneralSettingsPage::GeneralSettingsPage()
     });
 
     startupRoomNameEdit = new QLineEdit(settings.tabs().getStartupRoomName());
+    // Default (Expanding) would stretch the whole controls column when this row becomes visible,
+    // so size it like the combo boxes instead: fills the column, never widens it.
+    startupRoomNameEdit->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Fixed);
     connect(startupRoomNameEdit, &QLineEdit::editingFinished, this,
             [this] { SettingsCache::instance().tabs().setStartupRoomName(startupRoomNameEdit->text().trimmed()); });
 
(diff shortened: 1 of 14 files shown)


cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
```diff
@@ -121,8 +123,58 @@ GeneralSettingsPage::GeneralSettingsPage()
...
+        const QVariantList serverInfo = startupServerSelector.itemData(index).toList();
+        if (serverInfo.size() != 2) {
+            return;
+        }
+        TabsSettings &tabs = SettingsCache::instance().tabs();
+        tabs.setStartupServerHost(serverInfo[0].toString());
+        tabs.setStartupServerPort(serverInfo[1].toString());
+    });
+
+    startupRoomNameEdit = new QLineEdit(settings.tabs().getStartupRoomName());
```
RickyRister, line 164: Server Room causes column spacing to change compared to all the other options, which looks weird.

Image

Image

cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
```diff
@@ -393,6 +456,20 @@ void GeneralSettingsPage::retranslateUi()
     updateNotificationCheckBox.setText(tr("Notify if a feature supported by the server is missing in my client"));
     newVersionOracleCheckBox.setText(tr("Automatically run Oracle when running a new version of Cockatrice"));
     showTipsOnStartup.setText(tr("Show tips on startup"));
+    startupTabLabel.setText(tr("Startup tab:"));
+    startupTabSelector.setItemText(StartupTabHome, tr("Home"));
```
RickyRister: nit: can we write it as `StartupTab::StartupTabHome` so that it's clearer where they're coming from? Or make the thing an enum class.

RickyRister (approved):
LGTM

cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
BruebachL: Fixed

cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
BruebachL: Fixed

PR #7122, opened 2026-08-15
[VDE] Add a new setting to determine initial tab (Context/Deck/Database)

BruebachL:
VDE pulls double duty as both a deck editor and a deck display widget. This leads to some users being confused when they e.g. just want to create a new deck but land on an initially empty screen. Having the deck display open initially is useful for modifying decks quickly, however, so the choice should not be removed.

- Add a new setting that determines the initial startup tab

- Context: Deck Display when opening an existing deck, Database when creating a new one
- Deck Display: Always this tab
- Database: Always that tab

## Note:
Together with a setting to remember table/visual mode for the DB tab and the recent performance fixes, this pretty much obsoletes the classic deck editor imo (with the exception of the old filter system). Anything else is strictly on par as far as I can see.

### DIFF
--- a/cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
+++ b/cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
@@ -183,6 +183,13 @@ UserInterfaceSettingsPage::UserInterfaceSettingsPage()
     connect(&defaultDeckEditorTypeSelector, QOverload<int>::of(&QComboBox::currentIndexChanged),
             &SettingsCache::instance().deckEditor(), &DeckEditorSettings::setDefaultDeckEditorType);
 
+    vdeStartupTabSelector.addItem(""); // these will be set in retranslateUI
+    vdeStartupTabSelector.addItem("");
+    vdeStartupTabSelector.addItem("");
+    vdeStartupTabSelector.setCurrentIndex(SettingsCache::instance().deckEditor().getVdeStartupTab());
+    connect(&vdeStartupTabSelector, QOverload<int>::of(&QComboBox::currentIndexChanged),
+            &SettingsCache::instance().deckEditor(), &DeckEditorSettings::setVdeStartupTab);
+
     commanderSpellbookIntegrationUseOfficialBracketNamesExplainer.setText("?");
     commanderSpellbookIntegrationUseOfficialBracketNamesExplainer.setAutoRaise(true);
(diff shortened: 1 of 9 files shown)


cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.h
```diff
@@ -263,12 +263,12 @@ protected slots:
     /** @brief Handles dock close events. */
     void closeEvent(QCloseEvent *event) override;
 
-private:
```
RickyRister, line 266: why did this get moved?

cockatrice/src/interface/widgets/settings_page/user_interface_settings_page.cpp
```diff
@@ -368,6 +377,12 @@ void UserInterfaceSettingsPage::retranslateUi()
     defaultDeckEditorTypeLabel.setText(tr("Default deck editor type"));
     defaultDeckEditorTypeSelector.setItemText(TabSupervisor::ClassicDeckEditor, tr("Classic Deck Editor"));
     defaultDeckEditorTypeSelector.setItemText(TabSupervisor::VisualDeckEditor, tr("Visual Deck Editor"));
+    vdeStartupTabLabel.setText(tr("Visual deck editor startup tab"));
+    vdeStartupTabSelector.setItemText(VdeStartupTabContext, tr("Context"));
+    vdeStartupTabSelector.setItemText(VdeStartupTabDeckDisplay, tr("Deck display"));
+    vdeStartupTabSelector.setItemText(VdeStartupTabDatabaseDisplay, tr("Database display"));
+    vdeStartupTabSelector.setToolTip(
+        tr("New decks open on the database display; existing decks open on the deck view."));
```
RickyRister: Tooltip is not clear that it's specifically describing the `Context` setting

RickyRister (approved):

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.h
BruebachL: Needs to be protected instead of private for the override above

RickyRister: Isn't the override also private though?

BruebachL: Can't call the base impl if it's private tho

PR #7131, opened 2026-08-16
[Server] Add match result strategy hook

BruebachL:
Purely additive. No behavior change yet. `stopGameIfFinished` has no hook for post-game actions such as tournament match reporting.

- Add `Server_MatchResultStrategy` with `onGameFinished()`.
- Track the last non-conceded player in `stopGameIfFinished`.
- Call the strategy once a match is decided: a `true` result closes the game via `Event_GameClosed`.
- Null strategy returns `false`, so behavior is unchanged.

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
@@ -63,7 +63,8 @@ Server_Game::Server_Game(const GameConfig &config, Server_Room *_room)
       startingLifeTotal(config.startingLifeTotal), shareDecklistsOnLoad(config.shareDecklistsOnLoad),
       inactivityCounter(0), startTimeOfThisGame(0), secondsElapsed(0), firstGameStarted(false),
       turnOrderReversed(false), startTime(QDateTime::currentDateTime()), pingClock(nullptr),
-      lifecycleStrategy(new Server_DefaultLifecycleStrategy), gameMutex()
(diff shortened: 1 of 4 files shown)


libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
```diff
@@ -404,6 +407,16 @@ void Server_Game::stopGameIfFinished()
 
     sendGameStateToPlayers();
 
+    bool matchDecided = matchResultStrategy->onGameFinished(this, playing, lastPlayer);
+    if (matchDecided) {
+        locker.unlock();
+
+        sendGameEventContainer(prepareGameEvent(Event_GameClosed(), -1));
+        gameClosed = true;
+        deleteLater();
```
RickyRister, line 423: Are you sure `ServerGame` should be deleting itself like this?

BruebachL: Deregistration only ever happens in the destructor 
(see Server_Game::~Server_Game():96)
```
    room->removeGame(this);
```
and removeGame explicitly documents that it must be called from the destructor (because it self-manages lifecycle):
```
    // No need to lock gamesLock or gameMutex. This method is only
    // called from ~Server_Game, which locks both mutexes anyway beforehand.
```

Additionally, this follows other methods in the same class i.e.
pingClockTimeout()
```
        if (((maxTime > 0) && (++inactivityCounter >= maxTime)) || (playerCount < maxPlayers)) {
            deleteLater();
        }
```

and 

removeParticipant()

```
        } else {
            gameClosed = true;
            deleteLater();
            return;
        }
```

RickyRister (approved):

PR #7132, opened 2026-08-16
[Server] Add abstract match game factory for future tournament modes

BruebachL:
Purely additive. No behavior change yet. Tournament match creation needs an abstraction over room game creation that the server can inject later.

- Add `Server_MatchGameFactory` interface: `createMatchGame()`, `getUserInterface()`, `addGameToRoom()`.

## Why?
The upcoming tournament mode implements this factory to create match games without reaching into the room's game-creation logic.

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/CMakeLists.txt
+++ b/libcockatrice_network/libcockatrice/network/server/remote/CMakeLists.txt
@@ -15,6 +15,7 @@ set(HEADERS
     game/server_game.h
     game/server_game_lifecycle_strategy.h
     game/server_match_result_strategy.h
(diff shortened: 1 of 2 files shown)


RickyRister:
Do you have an implementation planned for this thing already?
I'll keep this unmerged for now since you might revise the design as you try to use it.

BruebachL:
> Do you have an implementation planned for this thing already? I'll keep this unmerged for now since you might revise the design as you try to use it.

I do and this is the last blocker PR before I can start pushing the tournament work since it's all based on this

RickyRister (approved):
Sure, I trust you

PR #7136, opened 2026-08-16
[Client] Route cockatrice:// link clicks from chat to the intent chain

BruebachL:
A cockatrice:// link clicked in chat is currently handed to the OS (or does nothing in-process). 

- Clicks now emit a cockatriceLinkActivated signal that travels ChatView -> Tab -> TabSupervisor -> MainWindow, which feeds the URL through the same IntentUrlParser the OS activation path uses, so the join runs entirely in-process. 
- card/user schemes and all other links behave as before.

### DIFF
--- a/cockatrice/src/interface/window_main.cpp
+++ b/cockatrice/src/interface/window_main.cpp
@@ -498,6 +498,7 @@ MainWindow::MainWindow(QWidget *parent)
     pixmapCacheSizeChanged(SettingsCache::instance().cacheStorage().getPixmapCacheSize());
 
     connectionController = new ConnectionController(this, this);
+    urlParser = new IntentUrlParser(this, this);
 
     createActions();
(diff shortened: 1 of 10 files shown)


RickyRister:
Did you wire up in-game chat?

BruebachL:
> Did you wire up in-game chat?

Good call

cockatrice/src/interface/window_main.cpp
```diff
@@ -861,6 +863,12 @@ void MainWindow::showWindowIfHidden()
     show();
 }
 
+void MainWindow::handleCockatriceLink(const QString &url)
+{
+    auto urlParser = new IntentUrlParser(this, this);
```
RickyRister: is this object ever cleaned up?

RickyRister (approved):

PR #7137, opened 2026-08-16
[Client] Confirm before joining a game opened from a game link

BruebachL:
Joining a game from a cockatrice://joingame link is a navigation decision, so restate what will be joined and ask before acting. 

- The confirm names the game description when the link carries one (falling back to the room name and numeric id for older links), and reports the host:port so links that point at a different server are obvious. 
- The intent chain is only started after confirmation.

### DIFF
--- a/cockatrice/src/interface/intents/url_parser.cpp
+++ b/cockatrice/src/interface/intents/url_parser.cpp
@@ -1,5 +1,7 @@
 #include "url_parser.h"
 
+#include "../widgets/tabs/tab_room.h"
+#include "../widgets/tabs/tab_supervisor.h"
 #include "../window_main.h"
 #include "contexts/context_join_game.h"
 #include "intent_join_server_game.h"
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/intents/url_parser.cpp
```diff
@@ -71,6 +74,44 @@ void IntentUrlParser::handleJoinGame(const QUrlQuery &query)
         return;
     }
 
+    const QString hostname = ctx->roomContext.serverContext.hostname;
```
RickyRister: nit: you can extract all this to a `generateJoinGameMessage` method.

RickyRister: No, i meant that you can literally extract everything from this line down into a non-static method that takes a `ContextJoinGame` and a `gameDescription` and returns the message

RickyRister (approved):

PR #7133, opened 2026-08-16
[Game] Extract makeGameJoinLink helper for cockatrice://joingame links

BruebachL:
## Short roundup of the initial problem
Adding different sites for inviting or link generation is easier with a shared helper.

## What will change with this Pull Request?
- The inline URL building in GameSelector's copy-link action moves into a shared helper so every invite/copy site produces the same link format. 
- The helper embeds the game description as an extra "game" query item (percent-encoded). Links without it stay valid because the receiving parser ignores unknown query items.

PR #7138, opened 2026-08-16
[UserList] Context menu invite

BruebachL:
## Related Ticket(s)
- Based on #7133

- The user context menu gains an "Invite to Game" submenu listing the inviteable games in the room (the inviter's own games, honoring the buddy-only setting). 
- Picking one opens a private message to the target user with a cockatrice://joingame link naming the game, so the target
gets a clickable invite instead of a raw URL. Multi-game rooms offer a picker and a single inviteable game sends directly.

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
@@ -355,6 +355,7 @@ void UserContextMenu::showContextMenu(const QPoint &pos,
 {
     QAction *aCopyToClipBoard = nullptr, *aRemoveMessages = nullptr;
     aUserName->setText(userName);
+    const bool anotherUser = userName != userListProxy->getOwnUsername();
 
     auto *menu = new QMenu(static_cast<QWidget *>(parent()));
     menu->addAction(aUserName);
@@ -366,6 +367,17 @@ void UserContextMenu::showContextMenu(const QPoint &pos,
     menu->addAction(aDetails);
     menu->addAction(aShowGames);
     menu->addAction(aChat);
+    const QList<GameInviteOption> inviteOptions = inviteOptionsForUser(userName);
+    if (!inviteOptions.isEmpty()) {
+        auto *inviteMenu = new QMenu(tr("&Invite to Game"), menu);
+        for (const GameInviteOption &option : inviteOptions) {
+            QAction *inviteAction = inviteMenu->addAction(option.label);
+            inviteAction->setEnabled(anotherUser && online);
+            connect(inviteAction, &QAction::triggered, this,
+                    [this, userName, option] { execInvite(userName, option); });
+        }
+        menu->addMenu(inviteMenu);
+    }
     if (userLevel.testFlag(ServerInfo_User::IsRegistered) && userListProxy->isOwnUserRegistered()) {
         menu->addSeparator();
         if (userListProxy->isUserBuddy(userName)) {
@@ -416,7 +428,6 @@ void UserContextMenu::showContextMenu(const QPoint &pos,
             menu->addAction(aPromoteToJudge);
         }
     }
-    bool anotherUser = userName != userListProxy->getOwnUsername();
     aDetails->setEnabled(true);
     aChat->setEnabled(anotherUser && online);
     aShowGames->setEnabled(online);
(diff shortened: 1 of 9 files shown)


cockatrice/src/interface/widgets/tabs/tab_message.cpp
```diff
@@ -96,7 +96,15 @@ void TabMessage::closeEvent(QCloseEvent *event)
 
 void TabMessage::sendMessage()
 {
-    if (sayEdit->text().isEmpty() || !userOnline) {
+    if (sayEdit->text().isEmpty()) {
+        return;
+    }
+
+    if (!userOnline) {
+        // Keep the draft: the user may be back momentarily, and the typed text
```
RickyRister: Can you extract this to a separate PR? This is a really useful fix that I wouldn't mind merging first.

cockatrice/src/interface/widgets/tabs/tab_message.cpp
```diff
@@ -111,6 +111,12 @@ void TabMessage::sendMessage()
     sayEdit->clear();
 }
 
+void TabMessage::sendInviteMessage(const QString &text)
```
RickyRister: This feels less like a `sendInviteMessage` method and more of a general "force send message" method.

cockatrice/src/interface/widgets/tabs/tab_message.cpp
```diff
@@ -111,6 +111,12 @@ void TabMessage::sendMessage()
...
+{
+    sayEdit->setText(text);
```
RickyRister: Won't this overwrite whatever's already in the textbox?

cockatrice/src/interface/widgets/tabs/tab_message.cpp
RickyRister, line 108: Perhaps you can extract this code into a separate method that takes the text as a param, and then have `sendMessage` call this with the contents of `sayEdit` while having the new code call it directly.

cockatrice/src/interface/widgets/tabs/tab_message.cpp
RickyRister: I still think `sendInviteMessage` is too narrow a name for a method that can send any message you give it.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -950,6 +948,54 @@ void TabSupervisor::talkLeft(TabMessage *tab)
...
+    QList<GameInviteOption> options;
+    if (isLocalGame) {
+        return options;
+    }
+
+    // The inviter may be in several games of the same room (hosting one and
+    // spectating another, for example). Return every game so the caller can
+    // let the user choose which one to invite to.
+    QMapIterator<int, TabGame *> gameIterator(gameTabs);
+    while (gameIterator.hasNext()) {
```
RickyRister: nit: if you're not going to use the keys, then you can just use a for-each to iterate over the `gameTabs.values()`.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -950,6 +948,54 @@ void TabSupervisor::talkLeft(TabMessage *tab)
...
+        // A closed game is a dead end — drop it. Started/full games stay
+        // listed: an invite to them is a legitimate "come spectate" offer.
+        if (metaInfo->proto().closed()) {
+            continue;
+        }
+
+        const int gameId = metaInfo->gameId();
+        const QString description = QString::fromStdString(metaInfo->proto().description());
+
+        GameInviteOption option;
```
RickyRister: nit: use designated initializer (and extract variables if it would be too unreadable otherwise)

cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
```diff
@@ -480,6 +491,60 @@ void UserContextMenu::execChat(const QString &userName)
...
+        if (option.onlyBuddies &&
+            (option.creatorName != userListProxy->getOwnUsername() || !userListProxy->isUserBuddy(userName))) {
+            continue;
+        }
+        result.append(option);
+    }
+    return result;
+}
+
+void UserContextMenu::execInvite(const QString &userName)
```
RickyRister, line 513: Where is this `execInvite` overload that doesn't take a `GameInviteOption` ever called from?

cockatrice/src/interface/widgets/tabs/tab_room.cpp
```diff
@@ -66,11 +68,15 @@ TabRoom::TabRoom(TabSupervisor *_tabSupervisor,
     userList = userListPanel->getUserList();
     connect(userListPanel, &UserListPanelWidget::openMessageDialog, this, &TabRoom::openMessageDialog);
 
+    const std::function<QList<GameInviteOption>()> gameInviteLinkProvider = [this]() -> QList<GameInviteOption> {
```
RickyRister: nit: you don't need the trailing return type. The compiler can already determine that.
Actually, you can also use `auto` so you don't have to write out the full type.

RickyRister (approved):
Works as expected

https://github.com/user-attachments/assets/568b47cd-afaf-490c-9434-5d2d7f968848

https://github.com/user-attachments/assets/4a1b60f0-efc5-4406-bcbb-e0ef642270ad

cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
BruebachL: #7144

PR #7139, opened 2026-08-16
[Client] Add copy-game-link action to the game menu

BruebachL:
Action is not exposed (and not consistently named)

- What it says on the tin

### DIFF
--- a/cockatrice/src/interface/widgets/server/game_selector.cpp
+++ b/cockatrice/src/interface/widgets/server/game_selector.cpp
@@ -319,7 +319,7 @@ void GameSelector::customContextMenu(const QPoint &point)
         dlg.exec();
     });
 
-    QAction copyLink(tr("Copy Game Link"));
+    QAction copyLink(tr("Cop&y game link"));
     connect(&copyLink, &QAction::triggered, this, [=, this]() {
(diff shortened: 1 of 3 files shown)


cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -319,7 +319,7 @@ void GameSelector::customContextMenu(const QPoint &point)
         dlg.exec();
     });
 
-    QAction copyLink(tr("Copy Game Link"));
+    QAction copyLink(tr("Cop&y game link"));
```
RickyRister, line 322: I thought we're not doing `&` anymore

RickyRister:
Reminder to add this to the shortcut settings, either in this PR or in a future one.

cockatrice/src/interface/widgets/server/game_selector.cpp
BruebachL: We're not? What's wrong with accelerator keys? Do we want to include this in our style guide?

RickyRister: It was something that @ebbit1q mentioned previously, I think

RickyRister (approved):

PR #7140, opened 2026-08-16
[Client] Confirm before joining a full game as a spectator

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

 Joining a full game without override privileges silently becomes a spectator join, which surprises the player.

- so ask first instead of surprising the player.


RickyRister:
Is there a way to make the buddies check happen before the spectator check? I suppose not because the buddy check requires a round-trip.

https://github.com/user-attachments/assets/bb5d5455-b989-4454-836f-1c23a179d5e6

Same thing with "spectators not allowed" games

https://github.com/user-attachments/assets/5d378402-d365-4b31-8548-2eedef5d117c

RickyRister (approved):

PR #7141, opened 2026-08-16
[Client] Name the game in the join-game password prompt

BruebachL:
The password prompts does not tell you which game you're joining

- Label the password prompt

### DIFF
--- a/cockatrice/src/interface/widgets/server/game_selector.cpp
+++ b/cockatrice/src/interface/widgets/server/game_selector.cpp
@@ -370,7 +370,12 @@ void GameSelector::joinGame(const bool asSpectator, const bool asJudge)
     QString password;
     if (game.with_password() && !(spectator && !game.spectators_need_password()) && !overrideRestrictions) {
         bool ok;
-        password = getTextWithMax(this, tr("Join game"), tr("Password:"), QLineEdit::Password, QString(), &ok);
+        // Games without a description have no sensible label — fall back to the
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/server/game_selector.cpp
```diff
@@ -370,7 +370,12 @@ void GameSelector::joinGame(const bool asSpectator, const bool asJudge)
     QString password;
     if (game.with_password() && !(spectator && !game.spectators_need_password()) && !overrideRestrictions) {
         bool ok;
-        password = getTextWithMax(this, tr("Join game"), tr("Password:"), QLineEdit::Password, QString(), &ok);
+        // Games without a description have no sensible label — fall back to the
+        // game id so the prompt still tells the user which game they're entering.
+        const QString gameLabel = QString::fromStdString(game.description());
+        const QString prompt = gameLabel.isEmpty() ? tr("Password for game #%1:").arg(game.game_id())
+                                                   : tr("Password for \"%1\":").arg(gameLabel);
```
RickyRister, line 377: Do we want to cut off the description past a certain length, or is it fine because it's their own fault for making the description so long.

Image

RickyRister (approved):

cockatrice/src/interface/widgets/server/game_selector.cpp
BruebachL: I think there's a DB limit in place for the worst offenders but I'm fine with this particular footgun for now

PR #7142, opened 2026-08-16
[Client] Keep the message draft and notify when the recipient is offline

BruebachL:
Draft messages are nuked when sent to offline users

- Keep the draft

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_message.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_message.cpp
@@ -23,9 +23,10 @@
 TabMessage::TabMessage(TabSupervisor *_tabSupervisor,
                        AbstractClient *_client,
                        const ServerInfo_User &_ownUserInfo,
-                       const ServerInfo_User &_otherUserInfo)
+                       const ServerInfo_User &_otherUserInfo,
+                       bool _userOnline)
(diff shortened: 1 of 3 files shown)


RickyRister:
Works if the user is online when you first messaged them

https://github.com/user-attachments/assets/104ca23e-6a43-4be7-a27f-6a0b55b5ea40

Does not work if user isn't online when you first messaged them

https://github.com/user-attachments/assets/99c36f33-42b6-4fe8-8d03-d79ae8c31af8

cockatrice/src/interface/widgets/tabs/tab_message.cpp
```diff
@@ -96,7 +96,15 @@ void TabMessage::closeEvent(QCloseEvent *event)
 
 void TabMessage::sendMessage()
 {
-    if (sayEdit->text().isEmpty() || !userOnline) {
+    if (sayEdit->text().isEmpty()) {
+        return;
+    }
+
+    if (!userOnline) {
```
RickyRister, line 104: seems like `userOnline` is set to true on creation and is only set to false in `processUserLeft`

RickyRister (approved):
Works now

https://github.com/user-attachments/assets/cf15f82c-04b0-40c5-85f7-2b2bae250e84

PR #7143, opened 2026-08-16
[Game] Add an invite button to non-started and not full games

BruebachL:
The primary UX flow ("I want to invite someone to THIS game") is still not answered.

- Add an "Invite" button above the player list

### DIFF
--- a/cockatrice/src/interface/widgets/dialogs/dlg_invite_to_game.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_invite_to_game.cpp
@@ -1,0 +1,119 @@
+#include "dlg_invite_to_game.h"
+
+#include "../server/user/user_list_manager.h"
+#include "../server/user/user_list_widget.h"
+#include "../tabs/tab_supervisor.h"
+
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QLineEdit>
+#include <QPushButton>
+#include <QUrl>
+#include <QUrlQuery>
+#include <QVBoxLayout>
+
+DlgInviteToGame::DlgInviteToGame(TabSupervisor *_tabSupervisor,
+                                 const QString &_inviteUrl,
+                                 bool _onlyBuddies,
+                                 const QStringList &_excludeUserNames,
+                                 QWidget *parent)
+    : QDialog(parent), tabSupervisor(_tabSupervisor), inviteUrl(_inviteUrl), onlyBuddies(_onlyBuddies),
+      excludeUserNames(_excludeUserNames)
+{
+    setModal(true);
+    setMinimumWidth(320);
+
+    searchEdit = new QLineEdit(this);
+    searchEdit->setClearButtonEnabled(true);
+    connect(searchEdit, &QLineEdit::textChanged, this, &DlgInviteToGame::searchTextChanged);
+
+    // The embedded list is the real room user list without the hover popup:
+    // same manager, same delegate/painter, same sections, live via manager
+    // signals while the modal loop runs.
+    UserListManager *manager = tabSupervisor->getUserListManager();
+    userList = new UserListWidget(tabSupervisor, tabSupervisor->getClient(), UserListWidget::RoomList, this,
+                                  /*withUserInfoPopup=*/false);
+    userList->setUserFilter([this, manager](const QString &name, bool online) {
+        return !excludeUserNames.contains(name) && online && !manager->isUserIgnored(name);
+    });
+    if (onlyBuddies) {
+        userList->setSectioned({UserListWidget::Section::Buddy});
+    } else {
+        userList->setSectioned({UserListWidget::Section::Buddy, UserListWidget::Section::Online});
+    }
+    userList->bind(manager);
+    userList->rebuild();
+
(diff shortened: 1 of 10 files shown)


cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -1264,11 +1321,27 @@ void TabGame::createPlayerListDock(bool bReplay)
...
 
+    auto *playerListBox = new QWidget(this);
+    auto *vbox = new QVBoxLayout(playerListBox);
+    vbox->setContentsMargins(0, 0, 0, 0);
+
+    if (!bReplay) {
+        inviteButton = new QPushButton(tr("Invite"), playerListBox);
+        inviteButton->setVisible(false);
+        connect(inviteButton, &QPushButton::clicked, this, &TabGame::actInviteToGame);
+        vbox->addWidget(inviteButton);
```
RickyRister, line 1335: are you sure we want to put a big unhidable button in a place that is already lacking vertical space? Especially when the button isn't even going to be used all that often, compare to how often the extra space it takes up will be annoying.

Screenshot 2026-08-16 at 3 01 47 PM

BruebachL: Displaying some sort of "Invite here" next to the player count is pretty standard UI/UX. It hides when the game starts. 

image

image

image

RickyRister: Maybe it would feel better if it was below the player list instead of above? Though maybe it's just because I'm not used to it.

BruebachL: I could see below. Maybe we can gate this behind a "Allow invites" lobby setting in the future so the button creation can be skipped entirely?

BruebachL: I've moved it to the bottom for now

cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -1264,11 +1321,28 @@ void TabGame::createPlayerListDock(bool bReplay)
     }
     playerListWidget->setFocusPolicy(Qt::NoFocus);
 
+    auto *playerListBox = new QWidget(this);
+    auto *vbox = new QVBoxLayout(playerListBox);
+    vbox->setContentsMargins(0, 0, 0, 0);
```
RickyRister, line 1326: We should also reduce spacing.

Image

cockatrice/src/interface/widgets/server/user/user_list_widget.h
```diff
@@ -261,6 +272,16 @@ private slots:
     void setShowTitle(bool showTitle);
     void setSectioned(const QList<Section> &ids);
     void setSectionExpanded(Section section, bool expanded);
+    /** Dialog mode: rows that fail the predicate are never shown. */
+    void setUserFilter(std::function<bool(const QString &userName, bool online)> filter)
+    {
+        userFilter = std::move(filter);
+    }
+    [[nodiscard]] int visibleUserRowCount() const;
+    [[nodiscard]] bool getWithUserInfoPopup() const
```
RickyRister: nit: `hasUserInfoPopup` would be more understandable at a glance.

cockatrice/src/interface/widgets/dialogs/dlg_invite_to_game.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+void DlgInviteToGame::searchTextChanged(const QString &text)
+{
+    userList->setFilterText(text);
+    updateEmptyLabel();
+}
+
+void DlgInviteToGame::updateEmptyLabel()
+{
+    emptyLabel->setVisible(userList->visibleUserRowCount() == 0);
```
RickyRister: Nit: Feels kinda awkward for the label to just pop up from the bottom and take up space suddenly. Not sure if the label is even needed in general. Feel free to disregard though because I don't have anything better either.

https://github.com/user-attachments/assets/1e335017-6a9e-473d-be0f-565d24f0e584

cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -605,99 +610,128 @@ UserListWidget::UserListWidget(TabSupervisor *_tabSupervisor,
...
-        if (item && item->type() == QTreeWidgetItem::Type &&
-            QString::fromStdString(static_cast<UserListTWI *>(item)->getUserInfo().name()) == hoveredUser) {
-            showPopupForUser(static_cast<UserListTWI *>(item));
-        }
-    });
+        userInfoPopup->hide();
+        userInfoPopup->setWindowOpacity(0.0);
+        userInfoPopup->installEventFilter(this);
+
+        showPopupTimer = new QTimer(this);
```
RickyRister: Are you sure we want to not create the timers? `showPopup` and `hidePopup` already guards against null `userInfoPopup`, so at worst the timers will just no-op. Always creating the timers means there's less risk of us getting a random segfault because we forgot to nullcheck a timer somewhere.

RickyRister (approved):
Works, I guess

https://github.com/user-attachments/assets/446c7183-3cf1-43dd-abad-860317560907

PR #7144, opened 2026-08-16
[UserList] Add invite button to hover popup

BruebachL:
Currently only wired up to the right click context menu

- Add an invite button to the hover popup in the user list


RickyRister (approved):
Works

https://github.com/user-attachments/assets/d007fcdf-657d-4133-ab7e-a062ad705692

PR #7149, opened 2026-08-21
Update & add arch/platform labels in release template

tooomm:
- ARM is the default on Macs, only highlight the special case for old Intel hardware on x86 arch
- Highlight that flatpak offers support to install Cockatrice on ARM + x86 architectures
- Highlight that the Docker image supports to run Servatrice on ARM + x86 systems

image


BruebachL (approved):
Looks good.

PR #7153, opened 2026-08-22
[Network] Measure real server round-trip times

BruebachL:
The ping indicator we currently have is just the keepalive indicator. There are no real metrics about latency in-client.

- Time each command container from send to response with QElapsedTimer
- Aggregate samples in a fixed-size ring buffer (last/median/p95/max), and emit aggregated pingStatsUpdated at most once per second so the hot path stays free of signal traffic. 
- Stats are cleared on disconnect.
- Forward the signal through ConnectionController for UI consumers
- unit-tested in latency_tracker_test.

### DIFF
--- a/libcockatrice_network/libcockatrice/network/client/abstract/abstract_client.cpp
+++ b/libcockatrice_network/libcockatrice/network/client/abstract/abstract_client.cpp
@@ -1,6 +1,7 @@
 #include "abstract_client.h"
 
 #include <google/protobuf/descriptor.h>
+#include <libcockatrice/protocol/debug_pb_message.h>
 #include <libcockatrice/protocol/featureset.h>
 #include <libcockatrice/protocol/get_pb_extension.h>
 #include <libcockatrice/protocol/pb/commands.pb.h>
@@ -28,6 +29,7 @@ AbstractClient::AbstractClient(QObject *parent)
     qRegisterMetaType<Response>("Response");
     qRegisterMetaType<Response::ResponseCode>("Response::ResponseCode");
     qRegisterMetaType<ClientStatus>("ClientStatus");
+    qRegisterMetaType<LatencyTracker::Stats>("LatencyTracker::Stats");
     qRegisterMetaType<RoomEvent>("RoomEvent");
     qRegisterMetaType<GameEventContainer>("GameEventContainer");
     qRegisterMetaType<Event_ServerIdentification>("Event_ServerIdentification");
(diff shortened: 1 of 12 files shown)


libcockatrice_network/libcockatrice/network/client/abstract/abstract_client.h
```diff
@@ -54,6 +60,23 @@ class AbstractClient : public QObject
...
     void maxPingTime(int seconds, int maxSeconds);
 
+    /**
+     * @brief Aggregated round-trip statistics, emitted at most once per second.
+     *
+     * All values are in milliseconds. sampleCount is the number of samples
+     * currently in the rolling window. Emitted from the client thread. The
+     * connection to UI objects is automatically queued across threads.
+     */
+    void pingStatsUpdated(int lastMs, int medianMs, int p95Ms, int maxMs, int sampleCount);
```
RickyRister: can we move the params into a struct? Actually, don't we already have the `LatencyTracker::Stats` struct?

libcockatrice_network/libcockatrice/network/client/abstract/abstract_client.cpp
```diff
@@ -161,9 +163,66 @@ void AbstractClient::queuePendingCommand(PendingCommand *pend)
...
+    latencyTracker.addSample(elapsed);
+
+    if (AbstractClientLog().isDebugEnabled()) {
+        qCDebug(AbstractClientLog).noquote()
+            << "command RTT:" << elapsed << "ms (cmd_id" << pend.getCommandContainer().cmd_id() << ")";
+    }
+
+    if (elapsed >= SlowGameCommandWarnMs && pend.getCommandContainer().game_command_size() > 0) {
+        qCWarning(AbstractClientLog) << "slow game command round trip:" << elapsed << "ms (cmd_id"
+                                     << pend.getCommandContainer().cmd_id() << ")";
```
RickyRister: The `cmd_id` is an opaque value that is useless unless you have full visibility into the system.

Image

Is it possible to log the type and/or contents of the command in addition?

libcockatrice_network/libcockatrice/network/client/abstract/abstract_client.h
```diff
@@ -54,6 +60,23 @@ class AbstractClient : public QObject
...
+
+    /**
+     * @brief Chronological snapshot of the rolling window, oldest sample first.
+     *
+     * Emitted together with pingStatsUpdated under the same throttle, so
+     * graphs can redraw without polling the tracker across threads.
+     */
+    void pingSamplesUpdated(const QList<int> &samplesMs);
```
RickyRister: nit: since these two signals are always emitted together and always used together, I wonder if it would make more sense to consolidate them into a single `pingStatsUpdated(stats, samples)` signal

RickyRister (approved):

PR #7154, opened 2026-08-22
[Client] Show latency in status bar and server tab indicator

BruebachL:
No client UI for #7153 

- Add a permanent status-bar label fed by ConnectionController's pingStatsUpdated which shows the latest round-trip time, hides while disconnected or without samples, and carries a tooltip with Last/Median/95th percentile/Maximum over the rolling sample window. 
- The server tab gets the same stats as its tooltip.
- Add LatencyGraphWidget, a size-agnostic bar sparkline over the rolling sample window: heights scale to the window's own range while colors map onto an absolute quality ramp, so a steady good ping stays green. 
- Embed it in the new LatencyStatusWidget together with the textual ping readout.

### DIFF
--- a/cockatrice/src/interface/window_main.cpp
+++ b/cockatrice/src/interface/window_main.cpp
@@ -541,10 +541,7 @@ MainWindow::MainWindow(QWidget *parent)
     statusBar()->addPermanentWidget(latencyStatus);
 
     connect(connectionController, &ConnectionController::pingStatsUpdated, latencyStatus,
-            [this](const LatencyTracker::Stats &stats, const QList<int> &samplesMs) {
-                latencyStatus->updateStats(stats);
-                latencyStatus->updateSamples(samplesMs);
-            });
+            &LatencyStatusWidget::updateData);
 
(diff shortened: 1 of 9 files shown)


RickyRister (approved):
LGTM (Assuming the previous PR's comments are addressed)

Image

cockatrice/src/interface/window_main.cpp
```diff
@@ -535,6 +537,15 @@ MainWindow::MainWindow(QWidget *parent)
             [this](bool show) { statusBar()->setVisible(show); });
     statusBar()->setVisible(SettingsCache::instance().userInterface().getShowStatusBar());
 
+    latencyStatus = new LatencyStatusWidget(this);
+    statusBar()->addPermanentWidget(latencyStatus);
+
+    connect(connectionController, &ConnectionController::pingStatsUpdated, latencyStatus,
+            [this](const LatencyTracker::Stats &stats, const QList<int> &samplesMs) {
+                latencyStatus->updateStats(stats);
+                latencyStatus->updateSamples(samplesMs);
```
RickyRister: nit: do we want to consolidate these two methods?

PR #7155, opened 2026-08-22
[Client] Detect main-thread event loop stalls

BruebachL:
Users don't know if their lag is caused by network latency or an event loop stall.

- LagMonitor ticks the GUI event loop every 500 ms and records gaps beyond 2 s as stalls, warning with their duration and keeping a bounded ring of recent records for diagnostics. 
- Measurement uses a monotonic QElapsedTimer so wall-clock steps and suspend do not fabricate stalls. 
- Recorded timestamps stay in wall time for correlating with user reports.

## Why?
A later commit will allow users to export all this information (Latency + Lag + Debug logs) as a .txt file for debug reports

### DIFF
--- a/cockatrice/src/client/lag_monitor.cpp
+++ b/cockatrice/src/client/lag_monitor.cpp
@@ -1,0 +1,43 @@
+#include "lag_monitor.h"
+
+#include <QTimer>
+
+LagMonitor::LagMonitor(QObject *parent) : QObject(parent)
+{
+    timer = new QTimer(this);
+    timer->setInterval(TickIntervalMs);
+    connect(timer, &QTimer::timeout, this, &LagMonitor::checkTick);
+    tickClock.start();
+    timer->start();
+}
+
+QList<LagMonitor::StallRecord> LagMonitor::recentStalls() const
+{
+    return stalls;
+}
+
+void LagMonitor::clearStalls()
+{
+    stalls.clear();
+}
+
+void LagMonitor::checkTick()
+{
+    const qint64 gap = tickClock.restart();
+
+    if (gap <= StallThresholdMs) {
+        return;
+    }
+
+    StallRecord record;
+    record.timestampMsSinceEpoch = QDateTime::currentMSecsSinceEpoch(); // wall time, for log correlation
+    record.durationMs = gap;
+
+    stalls.append(record);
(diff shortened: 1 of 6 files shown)


cockatrice/src/client/lag_monitor.h
```diff
@@ -0,0 +1,66 @@
...
+    Q_OBJECT
+
+public:
+    struct StallRecord
+    {
+        qint64 timestampMsSinceEpoch = 0; ///< when the stalled period ended
+        qint64 durationMs = 0;            ///< approximate length of the freeze
+    };
+
+    static constexpr int TickIntervalMs = 500;
```
RickyRister: nit: are we using PascalCase for constants now? We've been really inconsistent. So far we've also used SCREAMING_SNAKE_CASE and camelCase for constants.

cockatrice/src/client/lag_monitor.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+void LagMonitor::checkTick()
+{
+    const qint64 gap = tickClock.restart();
+
+    if (gap <= StallThresholdMs) {
+        return;
+    }
+
+    StallRecord record;
```
RickyRister: nit: use designated initializer

cockatrice/src/client/lag_monitor.cpp
```diff
@@ -0,0 +1,43 @@
...
+    record.timestampMsSinceEpoch = QDateTime::currentMSecsSinceEpoch(); // wall time, for log correlation
+    record.durationMs = gap;
+
+    stalls.append(record);
+    while (stalls.size() > MaxRecordedStalls) {
+        stalls.removeFirst();
```
RickyRister, line 58: are we not using a ring buffer for storage like we did for latency metrics? I assume it's because stall records isn't as close to the hot path?

RickyRister (approved):
LGTM (Thank you VDS for giving me an easy way to trigger the stall)

Image

cockatrice/src/client/lag_monitor.cpp
BruebachL: Yes, the latency tracking runs per network command so it's allocation free. In this case, writes are >2s apart, so it's not as critical (+ the child has already fallen into the well at that point)

PR #7156, opened 2026-08-22
[GamesModel] Clear games types before merging

BruebachL:
Protobuf appends to 'repeated' fields when merging, which leads to duplication in the games types category (the only repeated field)

- Clear GamesTypes before merging


RickyRister (approved):
Interesting. Why did it only break now?

BruebachL:
Server only got rebooped yesterday and was stale until then.

PR #7157, opened 2026-08-22
[Client] Restore stored banner printing when the art dialog opens

BruebachL:
Two problems made opening and confirming the banner dialog silently switch the banner to the default art of the first local printing. The caller dropped the card provider id when constructing the initial params, and the constructor left the printing combo wherever onCardNameChanged put it, which is always the first printing

- Pass the provider id through, restore it in the combo when it resolves locally, and keep it verbatim when it does not


RickyRister (approved):
Kind of a roundabout way to do it, but it works I guess.

PR #7158, opened 2026-08-22
[Client] Pin user list header length to the viewport width

BruebachL:
The header stretch mode kept a resize section property, so after any column grew past the viewport the list carried an invisible horizontal pan range that scrolled rows sideways without visual feedback

- Drop the leftover property so displayed length always equals viewport width and horizontal panning is impossible

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
@@ -613,12 +613,14 @@ UserListWidget::UserListWidget(TabSupervisor *_tabSupervisor,
     userTree->hideColumn(3);
     connect(userTree, &QTreeWidget::itemActivated, this, &UserListWidget::userClicked);
     userTree->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
-    // No stretchLastSection: column 0 is the only visible section and is
-    // stretched explicitly in applyDisplayMode(). Stretching the (hidden)
-    // last section on top of that can leave a stale horizontal scroll range,
-    // which pans rows sideways with no visible scrollbar.
+    // QTreeWidget enables stretchLastSection by default. Left on, the hidden
+    // last section absorbs viewport resizes, the Stretch sections never
+    // redistribute, and the header keeps a stale length past the viewport —
+    // an invisible horizontal pan range under ScrollBarAlwaysOff. Disable it
+    // so the explicit resize modes in applyDisplayMode() own the geometry.
+    userTree->header()->setStretchLastSection(false);
 
-    // Always create timers so callers never segfault on a null deref,
+    // Always create timers so callers never segfault on a null deref;
     // showPopupForUser / hidePopup already guard against a null userInfoPopup.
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -978,7 +990,7 @@ void UserListWidget::connectPopupSignals()
 bool UserListWidget::eventFilter(QObject *obj, QEvent *event)
 {
     // A press outside the tree, the popup and any open menu deselects the
-    // list and closes the popup. The filter is installed application-wide, so
+    // list and closes the popup. The filter is installed application wide, so
```
RickyRister: why is random typo correction also part of the PR and in the same commit

cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -613,9 +613,12 @@ UserListWidget::UserListWidget(TabSupervisor *_tabSupervisor,
     userTree->hideColumn(3);
     connect(userTree, &QTreeWidget::itemActivated, this, &UserListWidget::userClicked);
     userTree->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
-    userTree->header()->setStretchLastSection(true);
+    // No stretchLastSection: column 0 is the only visible section and is
+    // stretched explicitly in applyDisplayMode(). Stretching the (hidden)
+    // last section on top of that can leave a stale horizontal scroll range,
+    // which pans rows sideways with no visible scrollbar.
```
RickyRister: Horizontal scroll still isn't fixed if you join the room while you have the style settings enabled. However, if you disable then enable the setting *while* you're in a room, then it will be fixed (At least until you join the room again).

https://github.com/user-attachments/assets/eef37933-ed25-493c-a288-67dc83069738

EDIT: Looks like the part where the horizontal scroll is fixed if you cycle the setting is already present in master.

cockatrice/src/interface/widgets/server/user/user_list_widget.cpp
```diff
@@ -922,15 +925,24 @@ void UserListWidget::applyDisplayMode()
...
-        userTree->showColumn(2);
+        // Bounded widths instead of ResizeToContents: content sizing measures
+        // the FULL text width while the delegate elides afterwards, so long
+        // names widened the header past the viewport. Fixed icon columns plus
+        // a stretched name column keep the range at zero, eliding trims.
+        userTree->header()->setSectionResizeMode(0, QHeaderView::Fixed);
+        userTree->header()->resizeSection(0, 24);
+        userTree->header()->setSectionResizeMode(1, QHeaderView::Fixed);
+        userTree->header()->resizeSection(1, 22);
+        userTree->header()->setSectionResizeMode(2, QHeaderView::Stretch);
```
RickyRister, line 947: Unstyled user list is broken

Image

RickyRister (changes requested):

RickyRister (approved):
Fixed. Thanks

PR #7159, opened 2026-08-22
[Game] Derive playmat sampling window from shared clamped helpers

BruebachL:
The crop formula computed the sampled window inline with an unclamped zoom floor, so stored vertical offsets below half of travel were dead and extreme zooms could sample outside the art

- Remap verticalOffset to place the window top edge within its travel, floor zoom at visible width over min card side with a 4.0 ceiling, clamp pan along the margin sum constant segment, and expose playmatClampedZoom, playmatWindowSide and aspectFitRect so the game renderer and any editor share one geometry model

### DIFF
--- a/cockatrice/src/interface/widgets/playmat/playmat_utils.h
+++ b/cockatrice/src/interface/widgets/playmat/playmat_utils.h
@@ -6,42 +6,92 @@
 #include <QSizeF>
 #include <libcockatrice/deck_list/deck_list.h>
 
+inline qreal playmatMaxZoom()
+{
+    return 4.0;
+}
+
 /**
- * @brief Computes the source region of the full-resolution card image to use as a playmat.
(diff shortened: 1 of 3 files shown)


cockatrice/src/interface/widgets/playmat/playmat_utils.h
```diff
@@ -6,50 +6,100 @@
 #include <QSizeF>
 #include <libcockatrice/deck_list/deck_list.h>
 
+inline qreal playmatMaxZoom()
```
RickyRister: can't this be a constant instead?

RickyRister (commented):
I assume you know what you're doing with the numbers.

Note that we're dumping more functions into the global namespace.

RickyRister (approved):

PR #7160, opened 2026-08-22
[Client] Fix user list banner art rendering under display scaling

BruebachL:
Cached card art pixmaps carry the screen device pixel ratio, so both banner painters did their crop math on scaled pixels and blitted the result at raw over logical size, clipping art into its top left quadrant on any display above 100 percent

- Normalize a local copy to DPR 1 before crop math in UserListPainter and the popup header, clamp srcX and srcY bounds against stored zoom below 1, keep shared cache entries untouched


RickyRister (approved):
Fixed now. Thanks!

Image

Image

PR #7161, opened 2026-08-22
[Client] Replace playmat crop spinboxes with a direct manipulation preview

BruebachL:
## Related Ticket(s)
- Depends on #7159

The numeric fields exposed raw parameters with no relation to what the game draws, accepted values the renderer clamps away, and no way to see the result before committing

- The preview now renders through the exact game pipeline into a viewport shaped like a fresh board stack plus table area, with dimmed strips marking where a developed table crops further. 
- Drag pans, wheel zooms, arrow keys nudge, plus and minus zoom, Backspace or Esc restores the crop as of focus gain and lets Esc close the dialog when unchanged.
- Focus ring and accessible name and description cover keyboard and screen reader users
- New paints use palette roles so themes recolor them

### DIFF
--- a/cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
+++ b/cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
@@ -112,13 +112,16 @@ void PlaymatSettingsDialog::setupUi()
     });
 
     auto *form = new QFormLayout;
+    controlsForm = form;
     cardNameLabel = new QLabel;
     printingLabel = new QLabel;
     form->addRow(cardNameLabel, searchBar);
     form->addRow(printingLabel, providerComboBox);
 
     // Numerical editors expose the raw PlaymatParams for precise input. They
-    // stay hidden until requested since the crop surface is the primary control.
+    // share the same form as the rows above so every field lines up on one
+    // label column. They stay hidden until requested since the crop surface
+    // is the primary control.
     marginLSpin = makeSpinBox(0.0, 0.95, currentParams.marginPctL, 0.01);
     marginRSpin = makeSpinBox(0.0, 0.95, currentParams.marginPctR, 0.01);
     verticalOffsetSpin = makeSpinBox(0.0, 1.0, currentParams.verticalOffset, 0.01);
@@ -129,19 +132,13 @@ void PlaymatSettingsDialog::setupUi()
     verticalOffsetLabel = new QLabel;
     zoomLabel = new QLabel;
 
-    auto *numericForm = new QFormLayout;
-    numericForm->addRow(leftMarginLabel, marginLSpin);
-    numericForm->addRow(rightMarginLabel, marginRSpin);
-    numericForm->addRow(verticalOffsetLabel, verticalOffsetSpin);
-    numericForm->addRow(zoomLabel, zoomSpin);
-    numericEditors = new QWidget;
-    numericEditors->setLayout(numericForm);
-    numericEditors->setVisible(false);
-
     showNumericEditorsCheck = new QCheckBox;
 
     form->addRow(showNumericEditorsCheck);
-    form->addRow(numericEditors);
+    form->addRow(leftMarginLabel, marginLSpin);
+    form->addRow(rightMarginLabel, marginRSpin);
+    form->addRow(verticalOffsetLabel, verticalOffsetSpin);
+    form->addRow(zoomLabel, zoomSpin);
 
     controlsGroup = new QGroupBox;
     controlsGroup->setLayout(form);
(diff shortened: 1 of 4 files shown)


cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.h
```diff
@@ -63,18 +60,11 @@ private slots:
 
     QLabel *cardNameLabel;
     QLabel *printingLabel;
-    QLabel *leftMarginLabel;
```
RickyRister, line 66: do we want to entirely get rid of the spinboxes though? I'm reminded of Magic Set Editor, which has spinboxes in addition to the direct manipulation. The spinboxes are useful if you need to do fine tuning.

Image

tooomm:
macOS Debug builds reports 1 error: `unused variable 'kMaxZoom'`
```
/Users/runner/work/Cockatrice/Cockatrice/Qt/6.11.0/macos/lib/QtWebSockets.framework/Headers -isystem /Users/runner/work/Cockatrice/Cockatrice/Qt/6.11.0/macos/lib/QtXml.framework/Headers -g -O0 -Wall -Wextra -Werror -Wno-unused-parameter -std=gnu++20 -arch arm64 -fPIE -MD -MT cockatrice/CMakeFiles/cockatrice.dir/src/interface/widgets/playmat/playmat_preview_widget.cpp.o -MF cockatrice/CMakeFiles/cockatrice.dir/src/interface/widgets/playmat/playmat_preview_widget.cpp.o.d -o cockatrice/CMakeFiles/cockatrice.dir/src/interface/widgets/playmat/playmat_preview_widget.cpp.o -c /Users/runner/work/Cockatrice/Cockatrice/cockatrice/src/interface/widgets/playmat/playmat_preview_widget.cpp
  /Users/runner/work/Cockatrice/Cockatrice/cockatrice/src/interface/widgets/playmat/playmat_preview_widget.cpp:21:17: error: unused variable 'kMaxZoom' [-Werror,-Wunused-const-variable]
     21 | constexpr qreal kMaxZoom = 4.0;
        |                 ^~~~~~~~
  1 error generated.
```

cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.h
BruebachL: You can fine tune by nudging with the arrow keys, which I thought was an acceptable compromise for hiding these hidden values from the user (kinda want them to engage with the drag+pan controls or at least read about the keyboard controls)

RickyRister: Maybe we could hide it under an "advanced settings" dropdown?

BruebachL: I'd still rather keep this sleek (and I'm not really in the mood to do the wiring for the boxes again)

BruebachL: Additional reasoning being that fine adjustments can be made with the keyboard shortcuts (which mirrors existing ui), there isn't a *lot* of benefit to seeing the numerical values and I think the row of spinboxes is not quite as sleek and contributes to an 'outdated' look.

BruebachL: They're added again

cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
```diff
@@ -112,35 +109,52 @@ void PlaymatSettingsDialog::setupUi()
...
-    form->addRow(verticalOffsetLabel, verticalOffsetSpin);
-    form->addRow(zoomLabel, zoomSpin);
+
+    auto *numericForm = new QFormLayout;
+    numericForm->addRow(leftMarginLabel, marginLSpin);
+    numericForm->addRow(rightMarginLabel, marginRSpin);
+    numericForm->addRow(verticalOffsetLabel, verticalOffsetSpin);
+    numericForm->addRow(zoomLabel, zoomSpin);
+    numericEditors = new QWidget;
+    numericEditors->setLayout(numericForm);
```
RickyRister: Numeric editor causes info text to run into the preview, and has different centering than the fields above it.

Image

However, that can be fixed in a future PR.

RickyRister (approved):

cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
```diff
@@ -155,16 +169,34 @@ void PlaymatSettingsDialog::setupUi()
         accept();
     });
 
-    auto *root = new QVBoxLayout;
-    root->addWidget(controlsGroup);
-    root->addWidget(previewGroup);
-    root->addWidget(buttons);
-    setLayout(root);
+    // The crop surface is the primary control: dragging pans, wheel/keys zoom
+    //, editing exactly the same stored parameters the old numeric fields did.
```
RickyRister: nit: the comma being at the start of the line instead of the end of the previous line is annoying me

cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
BruebachL: I really don't feel like these add anything valuable tbh.

RickyRister: I think it's just because I'm too used to tools that give you an escape hatch to directly modify the underlying numbers if the ui isn't precise enough.

We can remove it later if no one seems interested or we come up with a better way to expose the values to power users.

PR #7162, opened 2026-08-22
[Client] Edit banner crop by direct manipulation in the card art dialog

BruebachL:
## Related Ticket(s)
- Depends on #7157

The offset and zoom fields exposed raw numbers with no relation to the strip.

- The preview now paints through UserListPainter::drawCardArt itself, so what you see is exactly what the user list renders. 
- Drag pans the art vertically at output scale, wheel zooms, arrow keys nudge, plus and minus zoom, Backspace or Esc restores the crop as of focus gain and lets Esc close the dialog when unchanged. 
- Margins stay explicit spinboxes since they trim the strip sides with no natural drag mapping. 
- Legacy stored zoom below the gesture floor is normalized once on open.
-  Focus ring and accessible name and description cover keyboard and screen reader users

### DIFF
--- a/cockatrice/src/interface/widgets/server/user/user_card_settings_dialog.cpp
+++ b/cockatrice/src/interface/widgets/server/user/user_card_settings_dialog.cpp
@@ -15,19 +15,46 @@
 #include <QFormLayout>
 #include <QGroupBox>
 #include <QHBoxLayout>
+#include <QKeyEvent>
 #include <QLabel>
 #include <QLineEdit>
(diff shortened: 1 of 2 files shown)


RickyRister (approved):
Works as expected.

https://github.com/user-attachments/assets/d548dbbb-e8a9-48c4-8d4a-d72256c066fa

I noticed that you don't get horizontal movement even after zooming, but I assume that's already existing behavior.

Issue #7152, opened 2026-08-22
Tests in CI are failing


tooomm:
Starting with merging https://github.com/Cockatrice/Cockatrice/pull/7092, Arch builds are reporting failing tests:

```
Run tests
  Test project /src/build
        Start  1: dummy_test
   1/20 Test  #1: dummy_test .......................   Passed    0.00 sec
        Start  2: expression_test
   2/20 Test  #2: expression_test ..................   Passed    0.02 sec
        Start  3: clamped_arithmetic_test
   3/20 Test  #3: clamped_arithmetic_test ..........   Passed    0.02 sec
        Start  4: test_age_formatting
   4/20 Test  #4: test_age_formatting ..............   Passed    0.02 sec
        Start  5: password_hash_test
   5/20 Test  #5: password_hash_test ...............   Passed    0.05 sec
        Start  6: playmat_resolver_test
   6/20 Test  #6: playmat_resolver_test ............   Passed    0.04 sec
        Start  7: server_card_counter_test
   7/20 Test  #7: server_card_counter_test .........   Passed    0.05 sec
        Start  8: server_counter_test
   8/20 Test  #8: server_counter_test ..............   Passed    0.04 sec
        Start  9: server_rate_limiter_test
   9/20 Test  #9: server_rate_limiter_test .........   Passed    0.02 sec
        Start 10: warning_categories_test
  10/20 Test #10: warning_categories_test ..........   Passed    0.02 sec
        Start 11: deck_hash_performance_test
  11/20 Test #11: deck_hash_performance_test .......***Timeout   5.01 sec
  [==========] Running 5 tests from 1 test suite.
  [----------] Global test environment set-up.
  [----------] 5 tests from DeckHashTest
  [ RUN      ] DeckHashTest.RepeatTest
  
        Start 12: card_zone_algorithms_test
  12/20 Test #12: card_zone_algorithms_test ........   Passed    0.00 sec
        Start 13: carddatabase_test
  13/20 Test #13: carddatabase_test ................   Passed    0.02 sec
        Start 14: filter_string_test
  14/20 Test #14: filter_string_test ...............   Passed    0.06 sec
        Start 15: loading_from_clipboard_test
  15/20 Test #15: loading_from_clipboard_test ......   Passed    0.05 sec
        Start 16: reverse_card_move_test
  16/20 Test #16: reverse_card_move_test ...........   Passed    0.05 sec
        Start 17: parse_cipt_test
[...]

PR #7163, opened 2026-08-23
add a limit to the size of deckfiles cockatrice can load

ebbit1q:
## Related Ticket(s)
related https://github.com/Cockatrice/Cockatrice/pull/5025
https://github.com/Cockatrice/Cockatrice/pull/5026

cockatrice seems to allow to load millions of cards, leading to lag, this stops this at the deck loader stage

the limit or 100k right now, which is kind of the limit of what looks acceptable in the player

note that clients will still struggle heavily to display 100k cards, and these decks mess with the visual deck selector, this is why most users don't try to abuse this up till whtz was printed

- add a constant in the deck loader
- use this limit when loading from xml and plaintext
- when the limit is reached, no cards are added to the deck
- this means those card names will also not be hashed, avoiding most performance issues

### DIFF
--- a/libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
+++ b/libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
@@ -5,7 +5,7 @@
 #include <QCryptographicHash>
 #include <QSet>
 
-static constexpr int MAX_DECK_SIZE = 99999;
+static constexpr int MAX_DECK_SIZE = 1e5;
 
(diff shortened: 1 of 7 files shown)


ebbit1q:
image

BruebachL (approved):
LGTM, at the very least as a stop-gap solution until we figure out more elegant ways to deal with it (i.e. error handling and such)

ebbit1q:
it looks like the hash test is now timing out

tooomm:
> it looks like the hash test is now timing out

See #7152, seems this test is flaky and takes longer since a few days.

tooomm:
Can this be merged?

PR #7164, opened 2026-08-23
[Network] Don't allow Event_ListGames to leak info and bloat bandwith

BruebachL:
## Related Ticket(s)
- Follow up to #7077

getUserInfo() returns the participant's raw stored ServerInfo_User, which was saved with copyUserInfo(true, true, true) at join time  i.e. everything: email, account id, session_id, address (IP), clientid, and up-to-2 MB avatar_bmp

- Privacy: every user in a room receives the host's email + home IP + session/account ids on every game-list broadcast 
- Performance: every gameInfoChanged (join/leave/start) re-broadcasts full host_info to all room users uncompressed. Mid-game joins ship N players × avatars.


ebbit1q (approved):
probably good

PR #7167, opened 2026-08-23
[VDS] Cache mana symbol renders and skip redundant resizes

BruebachL:
Lag because resize.

- Render each mana symbol once at a bounded master size and derive every requested size from the cached master, avoiding repeated full-size SVG rasterization on the GUI thread
- Share scaled results through a process-wide cache keyed by symbol and size, so repeated widget creation and rescales don't redo the work
- Skip redundant resize work in ColorIdentityWidget and ManaSymbolWidget when sizes did not change

### DIFF
--- a/cockatrice/src/interface/widgets/cards/additional_info/mana_symbol_widget.cpp
+++ b/cockatrice/src/interface/widgets/cards/additional_info/mana_symbol_widget.cpp
@@ -1,57 +1,15 @@
 #include "mana_symbol_widget.h"
 
 #include "../../../../client/settings/cache_settings.h"
+#include "../../../pixel_map_generator.h"
 
-#include <QHash>
-#include <QImageReader>
 #include <QResizeEvent>
 #include <libcockatrice/settings/visual_deck_storage_settings.h>
-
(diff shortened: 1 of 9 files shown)


cockatrice/src/interface/widgets/cards/additional_info/mana_symbol_widget.cpp
```diff
@@ -2,14 +2,56 @@
...
+#include <QImageReader>
 #include <QResizeEvent>
 #include <libcockatrice/settings/visual_deck_storage_settings.h>
 
+namespace
+{
+/// Longest side mana symbols are rendered at before being scaled to their final size.
+constexpr int MASTER_ICON_SIZE = 128;
+
+QHash<QString, QPixmap> &scaledIconCache()
```
RickyRister: We already have [`pixel_map_generator`](https://github.com/Cockatrice/Cockatrice/blob/master/cockatrice/src/interface/pixel_map_generator.cpp) file. Do we want to move it there?

RickyRister (approved):

PR #7171, opened 2026-08-23
[UserList] Restore accent-tinted rows for regular users in light mode

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


RickyRister (approved):
Before and after

Image

Image

BruebachL:
I'm fine with not having this if you think it doesn't fit/match. Intent was to highlight banner cards a bit better.
(Plus the 3px strip on the left)

RickyRister:
I like it

PR #7172, opened 2026-08-23
[UserList] Request banner card art for visible rows and hovered popup

BruebachL:
## Related Ticket(s)
- Fixes #IssueNumber

- this
- and this


RickyRister (approved):
How did banner cards still work before this then?

BruebachL:
They were pre-fetched (unconditionally, which this skips by checking the cache first)

PR #7174, opened 2026-08-23
[DeckList] Add maybeboard zone constant and skip it in exports

BruebachL:
- Introduce DECK_ZONE_MAYBEBOARD and its visible name, and treat the maybeboard as editor-only scratch space: plain-text export, DeckStats and TappedOut uploads now skip cards living there.
- Zones of this name are created by later custom-zones units until then the skips are inert.

### DIFF
--- a/cockatrice/src/client/network/interfaces/deck_stats_interface.cpp
+++ b/cockatrice/src/client/network/interfaces/deck_stats_interface.cpp
@@ -72,7 +72,7 @@ void DeckStatsInterface::copyDeckWithoutTokens(const DeckList &source, DeckList 
 {
     auto copyIfNotAToken = [&destination](const auto node, const auto card) {
         CardInfoPtr dbCard = CardDatabaseManager::query()->getCardInfo(card->getName());
-        if (dbCard && !dbCard->getIsToken()) {
+        if (dbCard && !dbCard->getIsToken() && node->getName() != DECK_ZONE_MAYBEBOARD) {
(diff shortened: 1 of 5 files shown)


RickyRister (approved):
How are you planning to allow users to get cards into there?

cockatrice/src/client/network/interfaces/deck_stats_interface.cpp
```diff
@@ -72,7 +72,7 @@ void DeckStatsInterface::copyDeckWithoutTokens(const DeckList &source, DeckList
 {
     auto copyIfNotAToken = [&destination](const auto node, const auto card) {
         CardInfoPtr dbCard = CardDatabaseManager::query()->getCardInfo(card->getName());
-        if (dbCard && !dbCard->getIsToken()) {
+        if (dbCard && !dbCard->getIsToken() && node->getName() != DECK_ZONE_MAYBEBOARD) {
```
RickyRister, line 75: On second thought, maybe it's better to make this check a whitelist rather than a blacklist, in case we add more zones in the future.

Could even extract it into a function.

BruebachL:
> How are you planning to allow users to get cards into there?

I have a whole stack of (more or less) atomic commits coming up. Most build on each other to eventually 'get there'

BruebachL:
> > How are you planning to allow users to get cards into there?
> 
> I have a whole stack of (more or less) atomic commits coming up. Most build on each other to eventually 'get there'

To expand on this: You can right click a card in the deck list and the database and select "move to zone" and then either pick an existing one or create a new custom one.

PR #7175, opened 2026-08-23
[DeckList] Make deck tree card traversal recursive 

BruebachL:
getCardNodes and forEachCard now descend into nested zones instead of assuming a flat main/side/token layout. For today's flat trees this is behavior-preserving. 
It also removes two latent crashes (unchecked dynamic_cast dereference, null card nodes passed to forEachCard
callers). 

Nested zones are introduced by later custom-zones units.

### DIFF
--- a/libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
+++ b/libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
@@ -41,13 +41,19 @@ QList<const DecklistCardNode *> DecklistNodeTree::getCardNodes(const QSet<QStrin
 {
     QList<const DecklistCardNode *> result;
 
-    for (auto *zoneNode : getZoneNodes(restrictToZones)) {
-        for (auto *cardNode : *zoneNode) {
-            auto *cardCardNode = dynamic_cast<DecklistCardNode *>(cardNode);
-            if (cardCardNode) {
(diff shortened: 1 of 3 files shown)


libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
```diff
@@ -41,13 +41,19 @@ QList<const DecklistCardNode *> DecklistNodeTree::getCardNodes(const QSet<QStrin
 {
     QList<const DecklistCardNode *> result;
 
-    for (auto *zoneNode : getZoneNodes(restrictToZones)) {
-        for (auto *cardNode : *zoneNode) {
-            auto *cardCardNode = dynamic_cast<DecklistCardNode *>(cardNode);
-            if (cardCardNode) {
-                result.append(cardCardNode);
+    std::function<void(const InnerDecklistNode *)> collectCards = [&collectCards,
```
RickyRister, line 44: nit: use `auto` so these signatures don't take up so much room.

RickyRister (approved):

libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
BruebachL: Compiler complains about auto here.

PR #7176, opened 2026-08-23
[DeckList] Add custom deck zones to the deck tree

BruebachL:
Introduce user-definable zones nested under a board zone (main, side or maybeboard) so players can organize cards inside a board without changing board semantics.

- addCustomZone, renameCustomZone, moveCustomZone and removeCustomZone manage zones. Names are unique across the whole deck and the standard zone names (main/side/maybeboard/tokens) stay reserved.
- Board zones are created lazily on first use.
- getZoneObjFromName resolves custom names to their nested node so addCard and XML loading route cards into them.
- Unknown names keep creating legacy top-level zones.
- deleteNode keeps empty custom zones alive and only prunes empty board zones.
- New deck_list_zones test suite locks hash parity with flat decks, sideboard size accounting, maybeboard exclusion from plain export and native-format round-trips.

### DIFF
--- a/libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
+++ b/libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
@@ -145,7 +145,8 @@ bool DecklistNodeTree::deleteNode(AbstractDecklistNode *node, InnerDecklistNode 
     if (index != -1) {
         delete rootNode->takeAt(index);
 
-        if (rootNode->empty()) {
+        // Empty custom zones are kept while empty board zones get pruned.
+        if (rootNode->empty() && rootNode->getParent() == root) {
             deleteNode(rootNode, rootNode->getParent());
         }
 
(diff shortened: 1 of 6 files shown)


libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
```diff
@@ -160,28 +167,173 @@ bool DecklistNodeTree::deleteNode(AbstractDecklistNode *node, InnerDecklistNode
...
+
+InnerDecklistNode *DecklistNodeTree::addCustomZone(const QString &boardZoneName, const QString &zoneName)
+{
+    if (hasZoneName(zoneName)) {
+        return nullptr;
+    }
+
+    auto *boardZone = dynamic_cast<InnerDecklistNode *>(root->findChild(boardZoneName));
+    if (!boardZone &&
+        (boardZoneName == DECK_ZONE_MAYBEBOARD || boardZoneName == DECK_ZONE_MAIN || boardZoneName == DECK_ZONE_SIDE)) {
```
RickyRister, line 221: extract to function

RickyRister: I was more thinking of extracting the name comparisons to a function

BruebachL: Can you be more precise then because looking at this, I saw like three extraction targets

RickyRister: Fair enough. My bad

RickyRister (approved):
I don't love how the decklist tree code is getting more complicated, but I trust that you know what you're doing, and we can always refactor it afterwards.

PR #7177, opened 2026-08-23
[Models] Mirror custom deck zones in the deck list model 

BruebachL:
DeckListModel now surfaces the custom zones from the deck tree so views can render and edit them alongside criteria groups.

- rebuildTree mirrors each custom zone as a DecklistModelSubZoneNode under its board zone, cards flat inside (no further grouping).
- The freshly built shadow tree is sorted while the model reset is still open, so views never observe unsorted intermediate order and proxies cannot desync.
- Custom zones always sort after criteria groups within a board, regardless of their names. sortHelper remaps persistent indexes
  through the pre-shift rows to keep selections valid.
- addCard inserts flat into a custom zone by name and keeps grouping by active criteria for board zones. findCardNode resolves cards in both layouts, legacy top-level zones unchanged.
- New IsCustomZoneRole lets views tell zones apart from groups.
- Empty custom zones survive row removal. Zone rows themselves are only mutable through the deck tree API.

### DIFF
--- a/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
+++ b/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
@@ -68,6 +68,16 @@ void DeckListModel::rebuildTree()
 
             //! \todo Better sanity checking.
             if (currentCard == nullptr) {
+                // Custom zones nested under the board zone are mirrored as-is,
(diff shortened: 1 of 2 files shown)


BruebachL:
Superseded by stacked PR #7203 (same-repo). All custom-zones work is now one linear stack targeting master.

BruebachL:
Superseded by #7203.

Issue #7179, opened 2026-08-24
Make home tab button color more customizable


RickyRister:
**Description of New Feature**

Introduce a `Home tab button color` setting, with the following options:
- `Automatic`: Use the default cockatrice green when no background, and background color when background (current behavior)  
- `From Background`: Always extract button color from background
- `Custom`: Present a color picker that allows the user to select a color gradient

**Additional Context**

Currently, the home tab button color is decided as follows:
- If there is no background set, use the default green color
- If there is a background set, extract the color from the background

However, my current button color looks like this:

<img width="633" height="1048" alt="Image" src="https://github.com/user-attachments/assets/92e96aff-56de-4ae4-80b8-4ab375583352" />

I have no background set, yet cockatrice is extracting the button color from the background. This is because I have the `Home tab background source` setting unset in my `ini` file. There is actually a bug that causes cockatrice to use the background color for the button instead of the default green when that happens.

I prefer this color scheme over the cockatrice green, so I would like to make this functionality official.

tooomm:
> I prefer this color scheme over the Cockatrice green, so I would like to make this functionality official.

👍 

With recent versions it kept changing between green and the dark buttons for me, I never knew why.

I prefer the "no color" version and will not use a background either.
If anything, the background should adapt to light mode for a uniform experience in the app (I'm using "system").

BruebachL:
It's green because our brand identity is green (I didn't choose the logo but I'm living with it.)
I'm not saying it's the prettiest color or even that it's one that I would've chosen but it is what it is.

I don't think the green default is necessarily a bad call.

I do see the usefulness in letting people customize the color itself (especially for themes and the likes) but the "Adapt to background" is in there mostly to keep it legible against a shuffling card art background.

I'm not arguing against this, just providing some reasons for why things are the way they are imo.

BruebachL:
Closed by #7279 and #7280

PR #7180, opened 2026-08-24
[HomeTab] Fix incorrect handling of invalid backgroundSource setting

RickyRister:
## Related Ticket(s)
- Relates to #7179

Code was using

```
SettingsCache::instance().appearance().getHomeTabBackgroundSource() != BackgroundSources::toId(BackgroundSources::Theme)
```

to check if the backgroundSource setting is set to the default value. However, if the setting in the `ini` file is set to an unexpected value, `getHomeTabBackgroundSource` will return that value and then directly try to compare it against the id of `theme`, which will return false.

- Instead use `BackgroundSources::fromId` (which handles invalid values) to map the value in the `ini` file to the enum, and then compare against the enum.


BruebachL (approved):
Whoopsie, good catch.

PR #7181, opened 2026-08-24
[HomeTab] Introduce button color source setting

RickyRister:
## Related Ticket(s)
- Relates to #7179
- Follow-up to #7180

We fixed the bug that allowed me to have cockatrice extract button colors from background while not having a background, so it's time to make it a feature.

- Create a `homeTabButtonColorSource` setting with two options:
  - `Automatic`: Existing behavior (use theme default if no background, otherwise extract from background)
  - `Extract from background`: always extract from background
- The setting is managed as an enum 
  - Created a header-only file namespaced under `HomeTabButtonColor` for the new enum
  - Used the displayname tech from `BackgroundSource` settings enum.
    - However, we just store the enum index in the `ini` file instead of an id string
- Update `AppearanceSettingsPage` and `HomeWidget` for the new setting. 

https://github.com/user-attachments/assets/4c7e2ad5-0e14-4158-82a7-ebf8a9fc99d7

## Screenshots

Screenshot 2026-08-23 at 11 19 44 PM

Screenshot 2026-08-23 at 11 25 10 PM

### DIFF
--- a/cockatrice/CMakeLists.txt
+++ b/cockatrice/CMakeLists.txt
@@ -230,6 +230,7 @@ set(cockatrice_SOURCES
     src/interface/widgets/general/display/charts/bars/segmented_bar_widget.cpp
     src/interface/widgets/general/display/charts/pies/color_pie.cpp
     src/interface/widgets/general/home_styled_button.cpp
+    src/interface/widgets/general/home_tab_button_color.h
     src/interface/widgets/general/home_widget.cpp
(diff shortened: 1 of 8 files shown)


BruebachL:
I would like to additionally look into providing custom cockatrice colors instead of just the Qt palette defaults in our theme's palette configuration so we can set values like this as well.

So in essence, a separate section that we parse for specific UI elements that allows us to go "off palette" in certain scenarios.

BruebachL (approved):

PR #7184, opened 2026-08-24
[SettingsPage] Refactor: Clean up order of variables

RickyRister:

The order of the variables in the `SettingsPage` classes can be cleaned up.

- Group variable declarations by their sections, in the same order that they appear in the ui 
  - With an empty line between each section
- In constructor and `retranslateUi`, make the code follow the order that the section show up in the page, and put at least an empty line between each section.

I only touched `general`, `appearance`, and `user interface` settings in this PR, since those are the settings that have the most sections.

### DIFF
--- a/cockatrice/src/interface/widgets/settings_page/appearance_settings_page.h
+++ b/cockatrice/src/interface/widgets/settings_page/appearance_settings_page.h
@@ -84,15 +84,15 @@ private:
 
     QGroupBox *themeGroupBox;
     QGroupBox *homeTabGroupBox;
+    QGroupBox *playmatGroupBox;
     QGroupBox *stylingGroupBox;
     QGroupBox *menuGroupBox;
     QGroupBox *printingsGroupBox;
     QGroupBox *cardsGroupBox;
     QGroupBox *cardLayoutGroupBox;
-    QGroupBox *handGroupBox;
-    QGroupBox *playmatGroupBox;
-    QGroupBox *tableGroupBox;
     QGroupBox *cardCountersGroupBox;
+    QGroupBox *handGroupBox;
+    QGroupBox *tableGroupBox;
 
 public:
     AppearanceSettingsPage();
--- a/cockatrice/src/interface/widgets/settings_page/appearance_settings_page.h
+++ b/cockatrice/src/interface/widgets/settings_page/appearance_settings_page.h
@@ -44,32 +44,44 @@ private:
     QLabel homeTabButtonColorSourceLabel;
     QComboBox homeTabButtonColorSourceBox;
 
-    QCheckBox styleUserListCheckBox;
-    QCheckBox showShortcutsCheckBox;
-    QCheckBox showGameSelectorFilterToolbarCheckBox;
-    QLabel minPlayersForMultiColumnLayoutLabel;
-    QLabel maxFontSizeForCardsLabel;
-    QCheckBox overrideAllCardArtWithPersonalPreferenceCheckBox;
-    QCheckBox bumpSetsWithCardsInDeckToTopCheckBox;
-    QCheckBox displayCardNamesCheckBox;
-    QCheckBox autoRotateSidewaysLayoutCardsCheckBox;
-    QCheckBox cardScalingCheckBox;
-    QCheckBox roundCardCornersCheckBox;
-    QLabel verticalCardOverlapPercentLabel;
-    QSpinBox verticalCardOverlapPercentBox;
-    QLabel cardViewInitialRowsMaxLabel;
-    QSpinBox cardViewInitialRowsMaxBox;
-    QLabel cardViewExpandedRowsMaxLabel;
-    QSpinBox cardViewExpandedRowsMaxBox;
-    QCheckBox horizontalHandCheckBox;
-    QCheckBox leftJustifiedHandCheckBox;
-    QCheckBox invertVerticalCoordinateCheckBox;
     QLabel playmatVisibilityLabel;
     QComboBox playmatVisibilityCombo;
     QLabel playmatModeLabel;
     QComboBox playmatModeCombo;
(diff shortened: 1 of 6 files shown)


cockatrice/src/interface/widgets/settings_page/general_settings_page.h
```diff
@@ -42,42 +42,47 @@ private slots:
...
-    QLabel languageLabel;
-    QLabel deckPathLabel;
-    QLabel filtersPathLabel;
-    QLabel replaysPathLabel;
-    QLabel picsPathLabel;
-    QLabel cardDatabasePathLabel;
-    QLabel customCardDatabasePathLabel;
-    QLabel tokenDatabasePathLabel;
+
     QLabel updateReleaseChannelLabel;
```
ZeldaZach: `updateReleaseChannelLabel` ended up stranded in a block of its own, between the card database group above and the startup group below. It belongs to neither: `versionGrid` places it at row 0 column 0 with `updateReleaseChannelBox` right next to it at column 1.

It should sit directly above `updateReleaseChannelBox` on line 49, which also matches the order you used in `retranslateUi`, where `updateReleaseChannelLabel.setText(...)` is the first line of the version settings block.

cockatrice/src/interface/widgets/settings_page/appearance_settings_page.h
```diff
@@ -81,9 +93,6 @@ private slots:
     QGroupBox *playmatGroupBox;
```
ZeldaZach: The `QGroupBox *` pointers are still in their original order while every widget declaration above them was regrouped to follow the UI. `mainLayout` adds them as theme, homeTab, playmat, styling, menu, printings, cards, cardLayout, cardCounters, hand, table, so `playmatGroupBox` should move up to third and `cardCountersGroupBox` should sit between `cardLayoutGroupBox` and `handGroupBox`.

Worth fixing for consistency: `general_settings_page.h` does keep its group box pointers in UI order, so the two files currently disagree on whether this block gets sorted.

cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
```diff
@@ -473,21 +477,18 @@ void GeneralSettingsPage::retranslateUi()
...
-
-    lastCardUpdateCheckDateLabel.setText(
-        tr("Last update check on %1 (%2 days ago)").arg(lastCheckDate.toString()).arg(daysAgo));
-
-    // We can't change the strings after they're put into the QComboBox, so this is our workaround
-    int oldIndex = updateReleaseChannelBox.currentIndex();
-    updateReleaseChannelBox.clear();
-    for (ReleaseChannel *chan : settings.getUpdateReleaseChannels()) {
-        updateReleaseChannelBox.addItem(tr(chan->getName().toUtf8()));
+    if (SettingsCache::instance().getIsPortableBuild()) {
```
ZeldaZach: Since `settings` is now bound at the top of the function, this can be `settings.getIsPortableBuild()`. It is the only remaining `SettingsCache::instance()` call in `retranslateUi` after the move.

cockatrice/src/interface/widgets/settings_page/general_settings_page.cpp
```diff
@@ -473,21 +477,18 @@ void GeneralSettingsPage::retranslateUi()
...
-    updateReleaseChannelBox.setCurrentIndex(oldIndex);
+    deckPathLabel.setText(tr("Decks directory:"));
+    filtersPathLabel.setText(tr("Filters directory:"));
+    replaysPathLabel.setText(tr("Replays directory:"));
+    picsPathLabel.setText(tr("Pictures directory:"));
+    cardDatabasePathLabel.setText(tr("Card database:"));
+    customCardDatabasePathLabel.setText(tr("Custom database directory:"));
+    tokenDatabasePathLabel.setText(tr("Token database:"));
+    resetAllPathsButton->setText(tr("Reset all paths"));
 }
```
ZeldaZach: Pre-existing rather than something this PR introduced, but the file has no trailing newline, which is why the diff renders `\ No newline at end of file`. Since this hunk already rewrites the end of the file, it is a cheap one to clean up here.

ZeldaZach (approved):
Verified this is behavior-preserving. I diffed the whitespace-stripped, sorted line multiset of each `.cpp` against master and the only differences are the three added section comments (`// Styling settings`, and the split of `// general settings and notification settings` into two). The declared member set of each `.h` is byte-identical to master once sorted, so nothing was dropped or duplicated.

On the risky parts of a reorder like this:

- No member has a default member initializer and none of them are constructed from another member, so changing declaration order does not change construction semantics. Destruction is likewise unaffected: the value-typed widgets are reparented into the group boxes, and member destructors run before `~QWidget` deregisters the rest, which is order-independent.
- The relocated constructor blocks are self-contained. The playmat block in `AppearanceSettingsPage` only needs `settings`, which is bound on the first line of the constructor, and its `visIdx`/`modeIdx` locals do not escape. The notification block in `UserInterfaceSettingsPage` moved below `generalGroupBox`, and nothing in the general block touches the notification checkboxes; `setNotificationEnabled` is only ever reached through a signal, so the later `retranslateUi()` call at the end of the constructor still sees fully wired widgets.
- In `GeneralSettingsPage::retranslateUi`, hoisting `const auto &settings` above its first use is fine, and moving the `updateReleaseChannelBox.clear()` + repopulate loop earlier does not change anything, since `setCurrentIndex(oldIndex)` still restores the selection before the function returns.

After the change, the constructor order, the `retranslateUi` order, and the member declaration order all line up with `mainLayout` on all three pages. CI is green across every platform including the `format` job.

Four non-blocking comments below, all consistency points against the PR's own stated goal rather than anything that would break. Approving.

PR #7186, opened 2026-08-24
[CI] Only save caches from `master`

tooomm:
Caches are saved when workflows run from PR's and compete for the limited available cache space.

- Harmonize caching strategy and follow the practice applied for our compiler caches (ccache) already:
  - **Qt** caches for Windows and macOS are saved and updated when builds are running from `master` branch
  - **vcpkg** caches for Windows and macOS are saved and updated when builds are running from `master` branch
- PR's can still use the master caches, but will not override them or consume space for the same cache item several times

We are constantly overrunning the available 10 GB limit at GitHub Actions Cache as we have more builds wanting to store something than there is available space. For that reason caches are constantly rotating and GitHub needs to delete existing caches, also from `master` branch.
The amount of PR's or dev activity should not have any impact on stable core caches provided from `master`, not saving PR caches helps witht hat.

The vcpkg caches are still messing with our cache a lot and no longer relevant vcpkg caches are not naturally cleared as all available ones are pulled and marked as used.
That will be a different PR.


.github/workflows/desktop-build.yml
```diff
@@ -399,7 +399,7 @@ jobs:
         run: .ci/thin_macos_qtlib.sh
 
       - name: "[macOS] Cache thin Qt libraries"
-        if: matrix.os == 'macOS' && steps.restore_qt.outputs.cache-hit != 'true'
+        if: matrix.os == 'macOS' && steps.restore_qt.outputs.cache-hit != 'true' && github.ref == 'refs/heads/master'
```
ebbit1q, line 406: this cache is rarely rotated and doesn't actually differ between branches, I don't think we should do this here

ebbit1q:
as far as I know both the qt and vcpkg caches will lead to the same output whether ran on a pr or master, these caches can then be used by any other pr or master without side effects. caches should not overwrite when there already is an up to date cache, it doesn't make sense to throw away updated caches, this'd lead to unnecessary increases on compile times in case all prs need to build the cache separately instead of reusing

.github/workflows/desktop-build.yml
tooomm: When searching for all available Qt caches right now:
image

It shows duplicates, and the Qt cache for Windows is not available on master at all, only on a branch...
We're not managing the available space well, which will cause deletion of other caches again (from master most likely).
The "least recently used cache" shows a last usage time from only "1 hour ago"!

That shows that we do not have any unused or old entries at all. We constantly cycle and can not keep all caches we could actually use right now.

The crazy amount of open (yet active) PR's is probably not helpful either.

Hope this can help to improve the situation a bit. It should not harm for sure.

tooomm:
> as far as I know both the qt and vcpkg caches will lead to the same output whether ran on a pr or master, these caches can then be used by any other pr or master without side effects. caches should not overwrite when there already is an up to date cache, it doesn't make sense to throw away updated caches, this'd lead to unnecessary increases on compile times in case all prs need to build the cache separately instead of reusing

PR's will always reuse and pull the available cache from master, just their incremental additions would not be saved (until they're merged).

From this PR's logs
- vcpkg
  ```
  Restoring vcpkg cache
    Found 380 caches for default branch ref 'refs/heads/master'
    Found 0 caches for current branch ref 'refs/pull/7186/merge'
  ```

- Qt
  ```
  Cache restored successfully
  Cache restored from key: Qt-6.11.1-macOS-Apple-qtimageformats qtmultimedia qtwebsockets-thin
  ```
  ```
  Using cached brotli-1.2.0-cp312-cp312-win_amd64.whl (369 kB)
  Using cached multivolumefile-0.2.3-py3-none-any.whl (17 kB)
  Using cached pycryptodomex-3.23.0-cp37-abi3-win_amd64.whl (1.8 MB)
  Downloading pyzstd-0.19.1-py3-none-any.whl (23 kB)
  Using cached backports_zstd-1.7.0-cp312-cp312-win_amd64.whl (329 kB)
  Using cached typing_extensions-4.16.0-py3-none-any.whl (45 kB)
  Using cached psutil-7.2.2-cp37-abi3-win_amd64.whl (137 kB)
  Using cached texttable-1.7.0-py2.py3-none-any.whl (10 kB)
  ...
  ```

tooomm:
The current situation around caching is just bad and a simple track during the last day helps to highlight the issue for Qt (screenshots are in chronological order):
image

image

image

image

image

image

image

image

image

image

image

image

image

image

<br><br>
>(Qt) cache is rarely rotated

That's the theoretical concept, due to us constantly overshooting the storage limit, rotation is happening all the time.
There is also negative impact on our compiler caches, because not all 12 are available all the time and GitHub deletes some at least every day. There are 5 missing at the time of this screenshot earlier today:
image

The screenshots show quite obvious a lot of not needed caches from PR's, not wanted cache cycling, doubled entries taking up precious space (up to 2 GB of space in times) and needed caches being not available for other builds to pull them.

Ideal situation would be to have always only 3 Qt related caches from `master` branch present:
- ~200 MB macOS
- ~200 MB macOS (x86)
- ~500 MB Windows

Edit: And not just one... screenshot is from just now...
image

<br>

While this PR certainly does not solve all issues, it helps with the situation.

---

vcpkg caching is another problem on it's own, but the same issue appiles. A lot of caches are created from PR branches without our vcpkg subdirectory, any dependencies or compiler etc. even changing:
image
(This is over 3 pages of vcpkg caches from a single PR btw: >80 cache entries)

Storage volume is also relevant:
image

I have been manually deleting hundred of entries in last week to get rid of months old entries that stayed around because they are pulled and count as used, even though they are well behind our vcpkg baseline, date way earlier than our checked out subdirectory and will not match. The sheer amount also leads to regular timeouts when setting up the vcpkg cache and pulling from GHA Cache:
image
image

While I have some ideas around that, I take it step by step.

tooomm:
Would love to move on with this 👍

BruebachL (approved):

PR #7187, opened 2026-08-24
[Server/Client/Protocol] Add developer staff role

BruebachL:
We're currently more-or-less flying blind whenever there are issues with an update. In order to facilitate debugging and maintenance, it is necessary to install an additional looking glass into the server, which, ideally, we as developers have access to.
Promoting every developer to admin is not feasible.

- Introduce a Developer staff level (proto flag 32, DB admin bit 8) that sits between admin and moderator: no kick/ban/warn/report/admin powers, but gets server log access via a new developer command container family (GET_SERVER_STATS, VIEWLOG_HISTORY) and an idle-timeout exemption. More features/stats to follow in another PR.

- Protocol: IsDeveloper flag, developer_commands.proto envelope, Command_GetServerStats/Command_GetLogHistory, Response_GetServerStats, Command_AdjustMod.should_be_developer
- Servatrice: fail-closed developer dispatcher, uptime snapshot handler, shared log history handler reuse, bit-8 DB mapping
- Client: burgundy pawn/badge/labels/sort order, prepareDeveloperCommand, minimal Developer stats tab, log tab access, promote/demote actions

### DIFF
--- a/cockatrice/src/interface/pixel_map_generator.cpp
+++ b/cockatrice/src/interface/pixel_map_generator.cpp
@@ -14,6 +14,7 @@
 #define DEFAULT_COLOR_MODERATOR_LEFT "#ffffff";
 #define DEFAULT_COLOR_MODERATOR_RIGHT "#000000";
 #define DEFAULT_COLOR_ADMIN "#ff2701";
+#define DEFAULT_COLOR_DEVELOPER "#800020"
 
 /**
  * Clamps an svg render size so that rendering does not exceed a multiple of the requested size.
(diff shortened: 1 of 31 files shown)


cockatrice/src/interface/pixel_map_generator.cpp
```diff
@@ -14,32 +13,7 @@
...
-    if (longestRequestedSide <= 0) {
-        return renderSize;
-    }
-
-    const int longestRenderSide = qMax(renderSize.width(), renderSize.height());
-    const qreal scale = qMin<qreal>(1.0, static_cast<qreal>(longestRequestedSide * 4) / longestRenderSide);
-    return QSize(qMax(1, static_cast<int>(renderSize.width() * scale)),
-                 qMax(1, static_cast<int>(renderSize.height() * scale)));
-}
+#define DEFAULT_COLOR_DEVELOPER "#800020";
```
RickyRister: Instead of having a separate color, I'd prefer if devs were represented with a pawn with an object in front, in the same vein as the donor or judge role. Maybe a cog or wrench as the object?

BruebachL:
Superseded by stacked same-repo PR #7211 (rebased on current master, now part of the developer-role → server-metrics stack). Kept open only to preserve the review thread here.

BruebachL:
Superseded by #7211.

PR #7191, opened 2026-08-29
[GameScene] Sever connections properly.

BruebachL:
Master's interim fix used QObject::disconnect(nullptr, nullptr, this, nullptr), which Qt rejects (nullptr sender → warning and unreliable).

- This change severs the destroy-tracking connections per-sender instead, eliminating both the warning and the re-entrancy hazard.

### DIFF
--- a/cockatrice/src/game_graphics/game_scene.cpp
+++ b/cockatrice/src/game_graphics/game_scene.cpp
@@ -44,11 +44,16 @@ GameScene::GameScene(PhasesToolbar *_phasesToolbar, QObject *parent)
 
 GameScene::~GameScene()
 {
-    // Sever all incoming connections (animated item destroy-tracking) before the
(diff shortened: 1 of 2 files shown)


RickyRister (changes requested):
Still crashes (Replay code is `13727505-jfiZddAQy6` in case you want to reproduce exactly)

https://github.com/user-attachments/assets/6a35d3dc-d1be-44d7-a6a6-35d924a328a2

If the crash I encountered has a different root cause though, then yeah, we can try to fix it in a separate PR.

RickyRister (approved):
Seems to be fixed. Thanks.

Don't love how thorny the code is becoming, but I guess we can always deal with it in a future refactor that never happens.

BruebachL:
Priority after merging in the 3.1.0 milestone tickets is definitely a refactor and some cleanup.

PR #7192, opened 2026-08-29
[Security] Use a CSPRNG for salts, tokens, and RNG seeding

BruebachL:

Several places in the codebase produced "random" values from predictable sources. The game's RNG was seeded from the wall clock, login salts were built from a small alphabet using that same RNG, and activation tokens were an MD5 of
a not-very-random salt. An attacker who can predict a seed can predict the resulting salts and tokens. Critical because those protect password hashes and account activation.

- Add `CryptoUtil`, backed by OpenSSL's `RAND_bytes`, a cryptographically secure PRNG (CSPRNG), as the single source of randomness for security.
- Seed the game's SFMT RNG from `CryptoUtil::randomUInt64()` instead of the current timestamp, so in-game randomness can no longer be predicted from the clock.
- Generate password salts from `CryptoUtil::randomBytes()`, and fix a modulo bias in salt character selection by adding rejection sampling.
- Generate account activation tokens directly from 16 secure random bytes instead of MD5-hashing a weak salt.
- Fail fast (`qFatal`) if the RNG ever fails, rather than silently falling back to a predictable source.
- Add regression tests for salt charset, uniqueness, and token length.

Supplies the secure randomness that the scrypt / challenge-response work in #7193 builds on.

### DIFF
--- a/libcockatrice_rng/libcockatrice/rng/rng_sfmt.cpp
+++ b/libcockatrice_rng/libcockatrice/rng/rng_sfmt.cpp
@@ -1,6 +1,5 @@
 #include "rng_sfmt.h"
 
-#include <QDateTime>
 #include <algorithm>
 #include <climits>
 #include <stdexcept>
@@ -11,10 +10,11 @@
 #define UINT64_MAX (~(uint64_t)0)
 #endif
 
-RNG_SFMT::RNG_SFMT(QObject *parent) : RNG_Abstract(parent)
+RNG_SFMT::RNG_SFMT(uint64_t seed, QObject *parent) : RNG_Abstract(parent)
 {
-    // initialize the random number generator with a 32bit integer seed (timestamp)
-    sfmt_init_gen_rand(&sfmt, QDateTime::currentDateTime().toSecsSinceEpoch());
+    // initialize the random number generator with a 64bit seed, e.g. from a CSPRNG
+    uint32_t seedArray[2] = {static_cast<uint32_t>(seed), static_cast<uint32_t>(seed >> 32)};
+    sfmt_init_by_array(&sfmt, seedArray, 2);
 }
 
 /**
--- a/libcockatrice_utility/CMakeLists.txt
+++ b/libcockatrice_utility/CMakeLists.txt
@@ -6,13 +6,14 @@ set(CMAKE_AUTOUIC ON)
 set(CMAKE_AUTORCC ON)
 
 set(UTILITY_SOURCES
-    libcockatrice/utility/expression.cpp libcockatrice/utility/levenshtein.cpp libcockatrice/utility/passwordhasher.cpp
-    libcockatrice/utility/server_rate_limiter.cpp libcockatrice/utility/warning_categories.cpp
+    libcockatrice/utility/cryptoutil.cpp libcockatrice/utility/expression.cpp libcockatrice/utility/levenshtein.cpp
+    libcockatrice/utility/passwordhasher.cpp libcockatrice/utility/server_rate_limiter.cpp libcockatrice/utility/warning_categories.cpp
 )
 
 set(UTILITY_HEADERS
     libcockatrice/utility/card_ref.h
     libcockatrice/utility/color.h
+    libcockatrice/utility/cryptoutil.h
     libcockatrice/utility/expression.h
     libcockatrice/utility/levenshtein.h
     libcockatrice/utility/macros.h
@@ -32,7 +33,9 @@ add_library(libcockatrice_utility STATIC ${UTILITY_SOURCES} ${UTILITY_HEADERS})
 
 target_include_directories(libcockatrice_utility PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-target_link_libraries(libcockatrice_utility PUBLIC libcockatrice_rng ${QT_CORE_MODULE})
+find_package(OpenSSL REQUIRED)
+
+target_link_libraries(libcockatrice_utility PUBLIC libcockatrice_rng OpenSSL::Crypto ${QT_CORE_MODULE})
 
 set(ORACLE_LIBS)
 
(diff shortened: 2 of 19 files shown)


libcockatrice_utility/CMakeLists.txt
```diff
@@ -32,7 +34,9 @@ add_library(libcockatrice_utility STATIC ${UTILITY_SOURCES} ${UTILITY_HEADERS})
 
 target_include_directories(libcockatrice_utility PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-target_link_libraries(libcockatrice_utility PUBLIC libcockatrice_rng ${QT_CORE_MODULE})
+find_package(OpenSSL REQUIRED)
```
ZeldaZach, line 37: **Please confirm the Linux/Docker build still configures with this.**

This makes OpenSSL an unconditional *build-time* requirement of `libcockatrice_utility`, which everything links against. The top-level `CMakeLists.txt` only does `find_package(OpenSSL REQUIRED)` inside `if(WIN32)` (line 238), and its own fallback message says OpenSSL is "not required for compiling, but needs to be available at runtime" — so this changes a runtime-only dependency into a compile-time one on macOS and Linux.

vcpkg-based builds are fine (`openssl` is already in `vcpkg.json`). The one I couldn't verify without running it is the Docker server image: `Dockerfile` installs `build-essential cmake ninja-build file g++ git libmariadb-dev-compat libprotobuf-dev libqt6sql6-mysql qt6-websockets-dev protobuf-compiler qt6-tools-dev qt6-tools-dev-tools` — no `libssl-dev`. It may well arrive transitively via `libmariadb-dev-compat`, but leaning on that is fragile. If it isn't already explicit, adding `libssl-dev` to the Dockerfile (and to any packaging/build docs that list dependencies) in this PR would save a broken build later.

Worth mentioning as an alternative, since it would sidestep the dependency question entirely: `QRandomGenerator::system()` is a CSPRNG backed by the OS facility (getrandom/BCryptGenRandom/arc4random) and needs nothing beyond QtCore, which is already linked. It would cover both `randomBytes` and `randomUInt64` with no new `find_package`. Your call — OpenSSL is a perfectly good choice, and #7193 pulls in `libcrypto` here anyway for scrypt and HMAC, so this may be moot in the stack.

libcockatrice_rng/libcockatrice/rng/rng_sfmt.cpp
```diff
@@ -17,6 +17,13 @@ RNG_SFMT::RNG_SFMT(QObject *parent) : RNG_Abstract(parent)
     sfmt_init_gen_rand(&sfmt, QDateTime::currentDateTime().toSecsSinceEpoch());
 }
 
+RNG_SFMT::RNG_SFMT(uint64_t seed, QObject *parent) : RNG_Abstract(parent)
```
ZeldaZach, line 13: The timestamp-seeded constructor just above (line 14-18) is still present and still seeds from `QDateTime::currentDateTime().toSecsSinceEpoch()`. Both `main.cpp` call sites now use the new one, so nothing predictable ships today — but the unsafe constructor remains the *default*, and it's the one you get from `new RNG_SFMT` with no arguments. It's an easy thing to reintroduce by accident.

Since there are no remaining callers (`tests/password_hash_test.cpp` drops its usage in this PR), consider just deleting it, or having it delegate:

```cpp
RNG_SFMT::RNG_SFMT(QObject *parent) : RNG_SFMT(CryptoUtil::randomUInt64(), parent) {}
```

That also resolves a small overload wart: `explicit RNG_SFMT(QObject *parent = nullptr)` and `explicit RNG_SFMT(uint64_t seed, QObject *parent = nullptr)` are ambiguous for a literal `0` or `NULL`. Not reachable from current code, just a trap.

One scoping note so the commit message doesn't overpromise: seeding SFMT from a CSPRNG fixes seed predictability, but SFMT is not itself a CSPRNG — its internal state is recoverable from enough observed output. `rng->rand()` drives shuffles (`server_cardzone.cpp:69`), dice rolls, and random reveals, so a player who can observe enough outputs can still in principle predict subsequent ones. Out of scope for this PR and not a regression, but "seeded securely" shouldn't be read as "shuffles are now unpredictable to an adversary."

libcockatrice_utility/libcockatrice/utility/cryptoutil.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <openssl/rand.h>
+
+namespace CryptoUtil
+{
+QByteArray randomBytes(int count)
+{
+    QByteArray bytes(count, '\0');
+    if (RAND_bytes(reinterpret_cast<unsigned char *>(bytes.data()), count) != 1) {
+        // Randomness failure is fatal: never fall back to a predictable source.
+        qFatal("CryptoUtil::randomBytes: RAND_bytes failed");
```
ZeldaZach, line 12: `qFatal` aborts the process, including in release builds. That's defensible for `randomUInt64()`, which is only called once during startup — if the OS RNG is broken at boot there's nothing sensible to do.

`randomBytes()` is different: it's reachable from `generateRandomSalt()` on the client whenever a user registers or changes a password, and #7193 wires it into `cmdRequestPasswordSalt` on the server (one call per login attempt, pre-auth). Hard-aborting servatrice from a code path that unauthenticated clients can drive is a sharper edge than it looks, even though `RAND_bytes` failing in practice is close to unheard-of.

The "never fall back to a predictable source" instinct is exactly right — the alternative isn't a weaker source, it's returning failure and letting the caller refuse the operation. Fine to leave as-is for this PR; flagging it because #7193 significantly widens who can reach it.

ZeldaZach (approved):
Reviewed in a local worktree against `master`. This one is solid — approving.

I checked the rejection sampling in `generateRandomSalt` by hand and it's correct: 62 symbols, `bucketSize = 65536/62 = 1057`, `limit = 65534`, so the rejection rate is 2/65536 and `value / bucketSize` maxes out at `65533/1057 = 61`. No modulo bias, no off-by-one, and the refill path can't read past the end of the buffer because `bytesUsed` is always even and the buffer is always `len * 2` bytes. Replacing the MD5-of-a-salt activation token with 16 raw CSPRNG bytes is a straight improvement, and the new tests are reasonable.

Three non-blocking notes inline — one of which (the OpenSSL build dependency) I'd like confirmed before merge even though it doesn't need a code change if the Docker build already pulls in the headers.

PR #7193, opened 2026-08-29
[Security] Add challenge-response auth with scrypt verifiers and stop storing plaintext passwords

BruebachL:
Authenticating with a password hash has two weaknesses. 

The first is that the stored hash we keep for each account is the only thing used to prove identity, and it is transmitted verbatim over the wire:
anybody who captures a login can replay it later (a "replay attack"), and the hash can be brute-forced offline because it is computed with a fast 1000-round SHA-512. 

The second is practical:
none of this is easy to understand, and the details matter for security, so this PR rewrites how logins prove who you are.

The headline: **your password is no longer the thing that proves your identity, so it and its verifier never leave the client, and servers stop being able to brute-force it.**

- Introduce challenge-response login using **scrypt** verifiers. On login the server sends a fresh, one-time *nonce* (a random token); the client computes a short keyed digest of it and sends only that back. The secret itself, the
  password or its verifier, is never transmitted, which defeats both replay attacks and any passive capture of login traffic.
- Store **scrypt** verifiers (a deliberately memory- and CPU-costly key derivation, RFC 7914) instead of the fast 1000-round SHA-512 hash. Even if the database leaks, each password is far costlier to brute-force.
- Keep the client's stored credential as a **verifier, never the plaintext password**. Saved servers hold a `$scrypt$...` string that can be used to answer a challenge but reveals nothing about the password.
- Automatically **migrate legacy accounts** to the scrypt format on their next successful login, so existing users upgrade without any action.
- Add a configurable `authentication_strictness` server setting (`legacy` / `mixed` / `strict`) so server operators control how strictly the new scheme is enforced.
- Bump the database schema (37) and widen the password column to hold the longer scrypt verifier.

For regular users: 
- the connection/registration/password-change flows look the same, but the "save password" checkbox now stores a scrypt verifier instead of your plaintext password in the client's settings.

For server operators: 
- review the new `[security] ` `authentication_strictness` 
- option; `strict` mode rejects old clients that can only send the legacy hash.

## Related tickets

Based on:
- #7192

### DIFF
--- a/cockatrice/src/client/network/connection_controller/remote_connection_controller.cpp
+++ b/cockatrice/src/client/network/connection_controller/remote_connection_controller.cpp
@@ -83,6 +83,9 @@ void ConnectionController::wireClientSignals()
 
     connect(remoteClient, &RemoteClient::sigPromptForForgotPasswordChallenge, this,
             &ConnectionController::onPromptForgotPasswordChallenge);
+
+    connect(remoteClient, &RemoteClient::sigPasswordVerifierReady, this,
+            &ConnectionController::onPasswordVerifierReady);
 }
 
 void ConnectionController::connectToServer()
@@ -91,16 +94,32 @@ void ConnectionController::connectToServer()
     connect(dlgConnect, &DlgConnect::sigStartForgotPasswordRequest, this, &ConnectionController::forgotPasswordRequest);
 
     if (dlgConnect->exec()) {
+        pendingSaveName = dlgConnect->getSaveName();
+        pendingSavePassword = dlgConnect->getSavePassword();
+        remoteClient->setStoredVerifier(dlgConnect->getStoredVerifier());
         remoteClient->connectToServer(dlgConnect->getHost(), static_cast<unsigned int>(dlgConnect->getPort()),
                                       dlgConnect->getPlayerName(), dlgConnect->getPassword());
+    }
+}
+
+void ConnectionController::onPasswordVerifierReady(const QString &verifier)
+{
+    if (pendingSavePassword) {
+        SettingsCache::instance().servers().setServerPassword(pendingSaveName, verifier);
     }
 }
 
 void ConnectionController::connectToServerDirect(const QString &host,
                                                  unsigned int port,
                                                  const QString &playerName,
-                                                 const QString &password)
+                                                 const QString &password,
+                                                 const QString &storedVerifier,
+                                                 const QString &saveName,
+                                                 bool savePassword)
 {
+    pendingSaveName = saveName;
+    pendingSavePassword = savePassword;
+    remoteClient->setStoredVerifier(storedVerifier);
     remoteClient->connectToServer(host, port, playerName, password);
 }
 
--- a/cockatrice/src/interface/widgets/dialogs/dlg_connect.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_connect.cpp
@@ -273,9 +273,15 @@ void DlgConnect::updateDisplayInfo(const QString &saveName)
     playernameEdit->setText(_data.at(3));
     playernameEdit->setFocus();
     savePasswordCheckBox->setChecked(savePasswordStatus);
+    storedVerifier.clear();
 
     if (savePasswordStatus) {
-        passwordEdit->setText(_data.at(4));
+        const QString stored = _data.at(4);
+        if (stored.startsWith("$")) {
+            storedVerifier = stored;
+        } else {
+            passwordEdit->setText(stored);
+        }
     }
 
     if (!_data.at(6).isEmpty()) {
@@ -301,6 +307,7 @@ void DlgConnect::newHostSelected(bool state)
         portEdit->setDisabled(false);
         playernameEdit->clear();
         passwordEdit->clear();
+        storedVerifier.clear();
         saveEdit->clear();
         saveEdit->setPlaceholderText(tr("Unique Server Name"));
         saveEdit->setDisabled(false);
@@ -326,6 +333,11 @@ void DlgConnect::actOk()
 {
     ServersSettings &servers = SettingsCache::instance().servers();
 
+    // Never write a newly typed plaintext password to disk when a verifier is already stored:
+    // the typed value is used for this connection and a fresh verifier is persisted after a
+    // successful login. Without a stored verifier we keep the previous (plaintext legacy) behavior.
+    const QString passwordToSave = storedVerifier.isEmpty() ? passwordEdit->text() : storedVerifier;
+
     if (newHostButton->isChecked()) {
         if (saveEdit->text().isEmpty()) {
             QMessageBox::critical(this, tr("Connection Warning"), tr("You need to name your new connection profile."));
@@ -333,10 +345,10 @@ void DlgConnect::actOk()
         }
 
         servers.addNewServer(saveEdit->text().trimmed(), hostEdit->text().trimmed(), portEdit->text().trimmed(),
-                             playernameEdit->text().trimmed(), passwordEdit->text(), savePasswordCheckBox->isChecked());
+                             playernameEdit->text().trimmed(), passwordToSave, savePasswordCheckBox->isChecked());
     } else {
         servers.updateExistingServer(saveEdit->text().trimmed(), hostEdit->text().trimmed(), portEdit->text().trimmed(),
-                                     playernameEdit->text().trimmed(), passwordEdit->text(),
+                                     playernameEdit->text().trimmed(), passwordToSave,
                                      savePasswordCheckBox->isChecked());
     }
 
--- a/libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
+++ b/libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
@@ -27,7 +27,8 @@ static const unsigned int protocolVersion = 14;
 
 RemoteClient::RemoteClient(QObject *parent, INetworkSettingsProvider *_networkSettingsProvider)
     : AbstractClient(parent), networkSettingsProvider(_networkSettingsProvider), timeRunning(0), lastDataReceived(0),
-      messageInProgress(false), handshakeStarted(false), usingWebSocket(false), messageLength(0), hashedPassword()
+      messageInProgress(false), handshakeStarted(false), usingWebSocket(false), messageLength(0), hashedPassword(),
+      passwordNeedsMigration(false)
 {
 
     clearNewClientFeatures();
@@ -114,6 +115,8 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
         return;
     }
     serverSupportsPasswordHash = event.server_options() & Event_ServerIdentification::SupportsPasswordHash;
+    serverSupportsChallengeResponse =
+        event.server_options() & Event_ServerIdentification::SupportsChallengeResponseAuth;
 
     if (getStatus() == StatusRequestingForgotPassword) {
         Command_ForgotPasswordRequest cmdForgotPasswordRequest;
@@ -130,7 +133,13 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
         cmdForgotPasswordReset.set_user_name(userName.toStdString());
         cmdForgotPasswordReset.set_clientid(getSrvClientID(lastHostname).toStdString());
         cmdForgotPasswordReset.set_token(token.toStdString());
-        if (!password.isEmpty() && serverSupportsPasswordHash) {
+        if (!password.isEmpty() && serverSupportsChallengeResponse) {
+            hashedPassword = PasswordHasher::generatePasswordVerifier(password);
+            // Only this branch yields a challenge-response verifier worth persisting; the
+            // legacy-hash and plaintext branches below must not be saved under the password key.
+            derivedVerifier = hashedPassword;
+            cmdForgotPasswordReset.set_hashed_new_password(hashedPassword.toStdString());
+        } else if (!password.isEmpty() && serverSupportsPasswordHash) {
             auto passwordSalt = PasswordHasher::generateRandomSalt();
             hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             cmdForgotPasswordReset.set_hashed_new_password(hashedPassword.toStdString());
@@ -158,7 +167,10 @@ void RemoteClient::processServerIdentificationEvent(const Event_ServerIdentifica
     if (getStatus() == StatusRegistering) {
         Command_Register cmdRegister;
         cmdRegister.set_user_name(userName.toStdString());
-        if (!password.isEmpty() && serverSupportsPasswordHash) {
+        if (!password.isEmpty() && serverSupportsChallengeResponse) {
+            hashedPassword = PasswordHasher::generatePasswordVerifier(password);
+            cmdRegister.set_hashed_password(hashedPassword.toStdString());
+        } else if (!password.isEmpty() && serverSupportsPasswordHash) {
             auto passwordSalt = PasswordHasher::generateRandomSalt();
             hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             cmdRegister.set_hashed_password(hashedPassword.toStdString());
@@ -223,7 +235,9 @@ Command_Login RemoteClient::generateCommandLogin()
 
 void RemoteClient::doLogin()
 {
-    if (!password.isEmpty() && serverSupportsPasswordHash) {
+    if ((!password.isEmpty() || !storedVerifier.isEmpty()) && serverSupportsChallengeResponse) {
+        doRequestPasswordSalt(); // ask salt + nonce to build the challenge response
+    } else if (!password.isEmpty() && serverSupportsPasswordHash) {
         //! \todo Store and log in using stored hashed password.
         if (hashedPassword.isEmpty()) {
             doRequestPasswordSalt(); // ask salt to create hashedPassword, then log in
@@ -257,6 +271,30 @@ void RemoteClient::doHashedLogin()
     sendCommand(pend);
 }
 
+void RemoteClient::doSubmitPasswordVerifier()
+{
+    pendingVerifier = PasswordHasher::generatePasswordVerifier(password);
+    Command_SubmitPasswordVerifier cmdSubmitVerifier;
+    cmdSubmitVerifier.set_password_verifier(pendingVerifier.toStdString());
+
+    PendingCommand *pend = prepareSessionCommand(cmdSubmitVerifier);
+    connect(pend, &PendingCommand::finished, this, &RemoteClient::submitPasswordVerifierResponse);
+    sendCommand(pend);
+}
+
+void RemoteClient::submitPasswordVerifierResponse(const Response &response)
+{
+    if (response.response_code() == Response::RespOk) {
+        qCDebug(RemoteClientLog) << "Password verifier migrated successfully";
+        if (!pendingVerifier.isEmpty()) {
+            emit sigPasswordVerifierReady(pendingVerifier);
+            pendingVerifier.clear();
+        }
+    } else {
+        qCWarning(RemoteClientLog) << "Failed to migrate password verifier:" << response.response_code();
+    }
+}
+
 void RemoteClient::processConnectionClosedEvent(const Event_ConnectionClosed & /*event*/)
 {
     doDisconnectFromServer();
@@ -269,7 +307,59 @@ void RemoteClient::passwordSaltResponse(const Response &response)
         auto passwordSalt = QString::fromStdString(resp.password_salt());
         if (passwordSalt.isEmpty()) { // the server does not recognize the user but allows them to enter unregistered
             password.clear();         // the password will not be used
+            storedVerifier.clear();
             doLogin();
+        } else if (serverSupportsChallengeResponse && resp.has_nonce()) {
+            const QByteArray nonce = QByteArray::fromStdString(resp.nonce());
+            QByteArray key;
+            if (resp.needs_migration()) {
+                // The account still uses the legacy format; the legacy full hash
+                // is only derivable from the plaintext password.
+                if (password.isEmpty()) {
+                    emit loginError(Response::RespClientUpdateRequired,
+                                    QStringLiteral("This account must be logged in with its password once."), 0, {});
+                    return;
+                }
+                key = PasswordHasher::computeHash(password, passwordSalt).toUtf8();
+            } else if (!password.isEmpty()) {
+                // A hostile server must not be able to make us run or allocate for
+                // unreasonable scrypt parameters.
+                const int n = resp.has_n() ? resp.n() : SCRYPT_N;
+                const int r = resp.has_r() ? resp.r() : SCRYPT_R;
+                const int p = resp.has_p() ? resp.p() : SCRYPT_P;
+                if (!PasswordHasher::costParamsAreSane(n, r, p)) {
+                    emit loginError(Response::RespClientUpdateRequired,
+                                    QStringLiteral("The server requested unreasonable scrypt cost parameters."), 0, {});
+                    return;
+                }
+                key = PasswordHasher::deriveKey(password, QByteArray::fromBase64(passwordSalt.toUtf8()), n, r, p);
+                if (key.isEmpty()) {
+                    emit loginError(Response::RespClientUpdateRequired, QStringLiteral("Unable to derive verifier."), 0,
+                                    {});
+                    return;
+                }
+                derivedVerifier = QString("$scrypt$%1$%2$%3$%4$%5")
+                                      .arg(n)
+                                      .arg(r)
+                                      .arg(p)
+                                      .arg(passwordSalt)
+                                      .arg(QString(key.toBase64()));
+            } else if (!storedVerifier.isEmpty()) {
+                const PasswordVerifier verifier = PasswordHasher::parsePasswordVerifier(storedVerifier);
+                if (!verifier.isValid) {
+                    emit loginError(Response::RespClientUpdateRequired, QStringLiteral("Stored verifier is invalid."),
+                                    0, {});
+                    return;
+                }
+                key = verifier.verifier;
+            } else {
+                emit loginError(Response::RespLoginNeeded, {}, 0, {});
+                return;
+            }
+            passwordNeedsMigration = resp.needs_migration();
+            const QByteArray responseBytes = PasswordHasher::computeResponse(key, nonce);
+            hashedPassword = "$challenge$" + QString(nonce.toBase64()) + "$" + QString(responseBytes.toBase64());
+            doHashedLogin();
         } else {
             hashedPassword = PasswordHasher::computeHash(password, passwordSalt);
             doHashedLogin();
@@ -293,6 +383,14 @@ void RemoteClient::loginResponse(const Response &response)
     if (response.response_code() == Response::RespOk) {
         setStatus(StatusLoggedIn);
         emit userInfoChanged(resp.user_info());
+
+        if (passwordNeedsMigration) {
+            // The account still used the legacy password format; upgrade it to scrypt.
+            doSubmitPasswordVerifier();
+        } else if (!derivedVerifier.isEmpty()) {
+            emit sigPasswordVerifierReady(derivedVerifier);
+            derivedVerifier.clear();
+        }
 
         QList<ServerInfo_User> buddyList;
         for (int i = resp.buddy_list_size() - 1; i >= 0; --i) {
@@ -550,6 +648,8 @@ void RemoteClient::doDisconnectFromServer()
         websocket->close();
     }
     socket->close();
+    derivedVerifier.clear();
+    pendingVerifier.clear();
 }
 
 void RemoteClient::ping()
@@ -712,6 +812,12 @@ void RemoteClient::submitForgotPasswordResetResponse(const Response &response)
 {
     if (response.response_code() == Response::RespOk) {
         emit sigForgotPasswordSuccess();
+        // Persist only a real scrypt verifier; a legacy hash must not be stored under
+        // the password key, where it would break future challenge-response logins.
+        if (!derivedVerifier.isEmpty()) {
+            emit sigPasswordVerifierReady(derivedVerifier);
+            derivedVerifier.clear();
+        }
     } else {
         emit sigForgotPasswordError();
     }
--- a/libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
+++ b/libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
@@ -311,6 +311,17 @@ void RemoteClient::passwordSaltResponse(const Response &response)
             doLogin();
         } else if (serverSupportsChallengeResponse && resp.has_nonce()) {
             const QByteArray nonce = QByteArray::fromStdString(resp.nonce());
+            // needs_migration is a server-controlled flag over an unauthenticated
+            // transport (plain TCP by default). If we already hold a scrypt
+            // verifier for this account, a server telling us to fall back to the
+            // 1000-round legacy hash is a downgrade of the KDF — refuse it rather
+            // than hand an attacker a weakly-keyed HMAC they can attack offline.
+            if (resp.needs_migration() && PasswordHasher::parsePasswordVerifier(storedVerifier).isValid) {
+                emit loginError(Response::RespClientUpdateRequired,
+                                QStringLiteral("Server asked to downgrade authentication for a migrated account."), 0,
+                                {});
+                return;
+            }
             QByteArray key;
             if (resp.needs_migration()) {
                 // The account still uses the legacy format; the legacy full hash
--- a/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
@@ -1,5 +1,8 @@
 #include "server_protocolhandler.h"
 
+// Challenge-response nonces are valid for at most one minute from issuance.
+static constexpr qint64 kAuthNonceLifetimeSeconds = 60;
+
 #include "game/game_config.h"
 #include "game/server_game.h"
 #include "game/server_player.h"
@@ -44,20 +47,26 @@ Server_ProtocolHandler::~Server_ProtocolHandler()
 {
 }
 
-void Server_ProtocolHandler::setAuthNonce(const QByteArray &nonce)
+void Server_ProtocolHandler::setAuthNonce(const QByteArray &nonce, const QString &userName)
 {
     authNonce = nonce;
+    authNonceUser = userName;
     authNonceCreated = QDateTime::currentDateTimeUtc();
 }
 
-bool Server_ProtocolHandler::isAuthNonceValid(const QByteArray &nonce) const
+bool Server_ProtocolHandler::isAuthNonceValid(const QByteArray &nonce, const QString &userName) const
 {
-    return !authNonce.isEmpty() && authNonce == nonce && authNonceCreated.secsTo(QDateTime::currentDateTimeUtc()) < 60;
+    // secsTo is signed, so a wall-clock step backwards (NTP correction, VM resume)
+    // must not make the elapsed time negative and re-validate an old nonce.
+    const qint64 elapsed = authNonceCreated.secsTo(QDateTime::currentDateTimeUtc());
+    return !authNonce.isEmpty() && authNonce == nonce && authNonceUser == userName && authNonceCreated.isValid() &&
+           elapsed >= 0 && elapsed < kAuthNonceLifetimeSeconds;
 }
 
 void Server_ProtocolHandler::clearAuthNonce()
 {
     authNonce.clear();
+    authNonceUser.clear();
 }
 
 // This function must only be called from the thread this object lives in.
--- a/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
@@ -44,6 +44,22 @@ Server_ProtocolHandler::~Server_ProtocolHandler()
 {
 }
 
+void Server_ProtocolHandler::setAuthNonce(const QByteArray &nonce)
+{
+    authNonce = nonce;
+    authNonceCreated = QDateTime::currentDateTimeUtc();
+}
+
+bool Server_ProtocolHandler::isAuthNonceValid(const QByteArray &nonce) const
+{
+    return !authNonce.isEmpty() && authNonce == nonce && authNonceCreated.secsTo(QDateTime::currentDateTimeUtc()) < 60;
+}
+
+void Server_ProtocolHandler::clearAuthNonce()
+{
+    authNonce.clear();
+}
+
 // This function must only be called from the thread this object lives in.
 // Except when the server is shutting down.
 // The thread must not hold any server locks when calling this (e.g. clientsLock, roomsLock).
@@ -505,6 +521,16 @@ Response::ResponseCode Server_ProtocolHandler::cmdLogin(const Command_Login &cmd
 
     if (userInfo != 0) {
         return Response::RespContextError;
+    }
+
+    // In strict mode only challenge-response logins are accepted.
+    if (server->requiresChallengeResponseAuth() &&
+        (!cmd.has_hashed_password() || cmd.hashed_password().rfind("$challenge$", 0) != 0)) {
+        auto *re = new Response_Login;
+        re->set_denied_reason_str("Client upgrade required");
+        re->add_missing_features("challenge_response_auth");
+        rc.setResponseExtension(re);
+        return Response::RespClientUpdateRequired;
     }
 
     // check client feature set against server feature set
--- a/servatrice/migrations/servatrice_0036_to_0037.sql
+++ b/servatrice/migrations/servatrice_0036_to_0037.sql
@@ -1,0 +1,8 @@
+-- Servatrice db migration from version 36 to version 37
+
+-- The column must hold "$scrypt$<n>$<r>$<p>$<salt>$<verifier>" (up to ~255 chars) and arbitrary
+-- legacy base64 hashes, so it grows beyond the old 120-char size. varchar(255) is used as the
+-- column type is promotion-safe and avoids the row-format change ALGORITHM=INSTANT cannot do.
+ALTER TABLE `cockatrice_users` MODIFY `password_sha512` varchar(255) NOT NULL;
+
+UPDATE cockatrice_schema_version SET version=37 WHERE version=36;
--- a/servatrice/migrations/servatrice_0036_to_0037.sql
+++ b/servatrice/migrations/servatrice_0036_to_0037.sql
@@ -1,8 +1,9 @@
 -- Servatrice db migration from version 36 to version 37
 
 -- The column must hold "$scrypt$<n>$<r>$<p>$<salt>$<verifier>" (up to ~255 chars) and arbitrary
--- legacy base64 hashes, so it grows beyond the old 120-char size. varchar(255) is used as the
--- column type is promotion-safe and avoids the row-format change ALGORITHM=INSTANT cannot do.
+-- legacy base64 hashes, so it grows beyond the old 120-char size. varchar(255) is used because
+-- widening a CHAR requires a table rebuild, which ALGORITHM=INSTANT cannot perform — dropping the
+-- clause lets the server pick a suitable algorithm (and any row-format change is avoided anyway).
 ALTER TABLE `cockatrice_users` MODIFY `password_sha512` varchar(255) NOT NULL;
 
 UPDATE cockatrice_schema_version SET version=37 WHERE version=36;
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -350,6 +350,22 @@ max_users_websocket=500
 ; Maximum number of users that can connect from the same IP address; useful to avoid bots, default is 4
 max_users_per_address=4
 
+; How strictly new authentication features are enforced.  Possible values:
+; * legacy: accounts that still use the legacy 1000-round SHA-512 hash keep logging in with it
+;           (they are served the legacy salt, never a challenge-response nonce) and are never
+;           auto-migrated to scrypt. Already-migrated scrypt rows keep logging in via
+;           challenge-response. New credentials may use either format;
+; * mixed: accept both legacy hashes and challenge-response authentication (default);
+; * strict: only accept challenge-response authentication from clients that support it,
+;           and reject plain password submissions.  Legacy accounts are migrated to scrypt
+;           automatically on their next successful login.
+;
+; Challenge-response verifiers are scrypt (RFC 7914, N=32768, r=8, p=1) stored as
+; "$scrypt$<n>$<r>$<p>$<salt>$<verifier>". The stored verifier is password-equivalent:
+; plaintext passwords never reach the client configuration and never go over the wire, but
+; a database dump yields credentials that can answer a login challenge directly.
+authentication_strictness=mixed
+
 ; You may want to allow an unlimited number of users from a trusted source. This setting can contain a
 ; comma-separed list of IP addresses which will allow an unlimited number of connections from each of the
 ; IP addresses listed (ignoring the max_users_per_address). Default is "127.0.0.1,::1"; example: "192.73.233.244,81.4.100.74"
--- a/servatrice/src/servatrice_database_interface.cpp
+++ b/servatrice/src/servatrice_database_interface.cpp
@@ -356,6 +356,47 @@ AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_Prot
                     qCWarning(DatabaseInterfaceLog) << "Login denied: user not active";
                     return UserIsInactive;
                 }
+
+                if (password.startsWith("$challenge$")) {
+                    // Challenge-response login: verify HMAC(stored_key, nonce) without
+                    // ever transmitting the stored credential or password hash.
+                    const QStringList parts = password.split("$");
+                    if (parts.size() != 4) {
+                        return NotLoggedIn;
+                    }
+                    const QByteArray nonce = QByteArray::fromBase64(parts.at(2).toUtf8());
+                    const QByteArray response = QByteArray::fromBase64(parts.at(3).toUtf8());
+                    if (nonce.isEmpty() || response.isEmpty() || !handler->isAuthNonceValid(nonce)) {
+                        return NotLoggedIn;
+                    }
+
+                    QByteArray key;
+                    if (PasswordHasher::isLegacyFormat(correctPasswordSha512)) {
+                        key = correctPasswordSha512.toUtf8();
+                    } else {
+                        // Design note: the stored scrypt verifier IS the challenge-response key, so a
+                        // database dump yields credentials that can answer a login challenge directly.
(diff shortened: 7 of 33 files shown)


servatrice/src/serversocketinterface.cpp
```diff
@@ -2841,6 +2856,12 @@ Response::ResponseCode AbstractServerSocketInterface::cmdAccountPassword(const C
         newPassword = QString::fromStdString(cmd.hashed_new_password());
     }
 
+    // In strict mode only scrypt verifiers are accepted.
```
ZeldaZach: **Blocking: password changes break permanently once an account migrates to scrypt.**

This guard validates the *new* password format, but the old-password check downstream was never updated. `Servatrice_DatabaseInterface::changeUserPassword(user, oldPassword, oldPasswordNeedsHash, ...)` does:

```cpp
QString salt = correctPasswordSha512.left(16);
oldPasswordSha512 = PasswordHasher::computeHash(oldPassword, salt);
if (correctPasswordSha512 != oldPasswordSha512) return false;
```

For a migrated account `correctPasswordSha512` is `$scrypt$32768$8$1$<salt>$<verifier>`, so `left(16)` is the literal string `$scrypt$32768$8$` — it grabs a slice of the format header as if it were a salt. The computed legacy hash is then compared against the full `$scrypt$...` string and can never match. `cmdAccountPassword` returns `RespWrongPassword` every time.

This isn't an edge case: in `mixed` mode every account migrates on its next login, and from that moment on its owner can never change their password from inside the client again. Password reset by email still works (`cmdForgotPasswordReset` uses the 3-argument overload with a token and no old-password check), so it's recoverable, but it's a guaranteed regression for every user on the server.

`changeUserPassword` needs a scrypt branch — for a verifier row, parse it, re-derive with the stored `n`/`r`/`p` and salt, and `constantTimeEquals` against `verifier.verifier`. Note that also settles how the old password reaches the server: `user_info_box.cpp` still sends `cmd.set_old_password(oldPassword.toStdString())` in plaintext even in `strict` mode, which is a bit at odds with the rest of the PR. A challenge-response variant for the old password would be more consistent, though plaintext-under-TLS is defensible if you'd rather keep the change small.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
```diff
@@ -269,7 +304,47 @@ void RemoteClient::passwordSaltResponse(const Response &response)
...
+                // The account still uses the legacy format; the legacy full hash
+                // is only derivable from the plaintext password.
+                if (password.isEmpty()) {
+                    emit loginError(Response::RespClientUpdateRequired,
+                                    QStringLiteral("This account must be logged in with its password once."), 0, {});
+                    return;
+                }
+                key = PasswordHasher::computeHash(password, passwordSalt).toUtf8();
+            } else if (!password.isEmpty()) {
+                const int n = resp.has_n() ? resp.n() : SCRYPT_N;
```
ZeldaZach, line 338: **Blocking: a malicious or misconfigured server can hard-crash every client that connects.**

`n`, `r` and `p` come straight off the wire and go unvalidated into `deriveKey`, which computes `maxmem = 128 * n * r + 128 * r * p + 4096` and calls `EVP_PBE_scrypt`. `deriveKey` reacts to failure with `qFatal`, which aborts the process in release builds too.

Three ways that goes wrong, all reachable by a server that simply says so:

- `n = 1073741824` → `maxmem` around 1.1 TB. Either OpenSSL fails and you `qFatal`, or the allocation succeeds far enough to OOM the machine first.
- `n` negative — these are `int32` on the wire and nothing rejects a negative. `EVP_PBE_scrypt` fails, `qFatal`.
- `n` not a power of two — scrypt requires it. `EVP_PBE_scrypt` fails, `qFatal`.

The last one doesn't even need malice: one corrupted or hand-edited `password_sha512` row and every client trying to log into that account dies. Cockatrice ships a downloadable public server list, so "connect to a server that then crashes you" is a low bar to clear.

Two changes needed:

1. **`deriveKey` must not `qFatal`.** It's called on network-supplied input on both ends now. Return `QByteArray()` on failure and have callers surface a login error.
2. **Clamp the parameters before use**, e.g.

```cpp
static bool costParamsAreSane(int n, int r, int p)
{
    return n >= 1024 && n <= (1 << 20) && (n & (n - 1)) == 0 &&
           r >= 1 && r <= 32 && p >= 1 && p <= 16;
}
```

and emit `loginError` rather than deriving when it fails. A ceiling of `n = 2^20`, `r = 32` still bounds the client at roughly 4 GB, so you may want it tighter — but anything finite beats what's here.

Same validation belongs in `parsePasswordVerifier`, which currently accepts any positive `n`/`r`/`p` from the database.

servatrice/src/serversocketinterface.cpp
```diff
@@ -3147,6 +3197,29 @@ Response::ResponseCode AbstractServerSocketInterface::cmdReport(const Command_Re
     return Response::RespOk;
 }
 
+Response::ResponseCode
+AbstractServerSocketInterface::cmdSubmitPasswordVerifier(const Command_SubmitPasswordVerifier &cmd,
```
ZeldaZach, line 3228: **Blocking: unvalidated verifier storage, and the resulting lockout can't be undone.**

The only structural check on `passwordVerifier` is `!isLegacyFormat(...)`, i.e. "starts with `$`". Anything else that starts with a dollar sign gets written straight into `password_sha512`.

Submit `$x` and here's what happens on the next login: `cmdRequestPasswordSalt` calls `parsePasswordVerifier`, gets `isValid == false`, and returns `RespContextError` — for every subsequent attempt. The account is locked out. And it can't be repaired through this command, because `submitPasswordVerifier`'s UPDATE is scoped `where ... password_sha512 not like '$%'`, and the row now matches `'$%'`. It takes manual database surgery to recover.

That's reachable by any authenticated user against their own account (self-inflicted, but a client bug or a truncated write would do it), and by anyone who has hijacked a session against someone else's.

Parse before storing:

```cpp
const PasswordVerifier parsed = PasswordHasher::parsePasswordVerifier(passwordVerifier);
if (!parsed.isValid || !costParamsAreSane(parsed.n, parsed.r, parsed.p)) {
}
```

A minimum on `n` matters here too — with cost parameters entirely under client control, a client can register itself at `n = 1024`. It happens not to weaken server-side authentication today only because the server never runs scrypt during login (see the note on `servatrice_database_interface.cpp`), which is itself the thing worth discussing.

Two smaller things in the same function:
- `MAX_NAME_LENGTH` (255) works as a bound — a real verifier is 131 characters — but it reads as though a username limit were reused by accident. A named constant would be clearer about intent.
- `submitPasswordVerifier` returns `true` when the UPDATE affects zero rows, so a no-op migration reports `RespOk` and the client happily persists a verifier the server never stored. `numRowsAffected() > 0` would make that visible.

servatrice/src/serversocketinterface.cpp
```diff
@@ -3038,17 +3065,40 @@ Response::ResponseCode AbstractServerSocketInterface::cmdRequestPasswordSalt(con
...
         } else {
             // user does not exist but is allowed to log in unregistered without password
             return Response::RespOk;
         }
     }
+
     auto *re = new Response_PasswordSalt;
-    re->set_password_salt(passwordSalt.toStdString());
+    const bool challengeResponseEnabled = servatrice->getAuthenticationStrictness() != Servatrice::AuthenticationLegacy;
+    if (PasswordHasher::isLegacyFormat(storedPasswordData)) {
```
ZeldaZach, line 3101: **Blocking: setting `authentication_strictness=legacy` locks out every already-migrated account.**

In `legacy` mode `challengeResponseEnabled` is false, so for a scrypt account this sends `password_salt` = the base64 scrypt salt, sets `n`/`r`/`p`, and sends no nonce. The client never saw `SupportsChallengeResponseAuth` in the identification event, so it falls through to the `else` at `remote_client.cpp:349` and computes `PasswordHasher::computeHash(password, passwordSalt)` — a legacy 1000-round SHA-512 hash over the base64 salt. `checkUserPassword` then compares that against the stored `$scrypt$...` string. It can never match.

So the sequence is: run in `mixed` (the default) for a while, accounts migrate, hit a problem, set `legacy` to roll back — and now everyone who migrated cannot log in at all, with no clue why. Rollback is exactly when an operator reaches for that setting, so it's the worst possible time for it to fail.

Either refuse the downgrade (log an error at startup if `legacy` is configured while scrypt rows exist), or keep serving challenge-response for scrypt rows regardless of mode and let `legacy` only govern what *new* credentials are accepted. The second seems closer to the intent, and `legacy` would then mean "don't migrate anyone new" rather than "pretend scrypt doesn't exist".

Worth documenting in `servatrice.ini.example` either way — right now nothing warns that the setting is effectively one-way.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
```diff
@@ -712,6 +797,9 @@ void RemoteClient::submitForgotPasswordResetResponse(const Response &response)
 {
     if (response.response_code() == Response::RespOk) {
         emit sigForgotPasswordSuccess();
+        if (!hashedPassword.isEmpty()) {
+            emit sigPasswordVerifierReady(lastHostname, userName, hashedPassword);
```
ZeldaZach: **Blocking: this can persist a legacy hash as though it were a verifier.**

`hashedPassword` is set in two different branches in `processServerIdentificationEvent`: the challenge-response branch (line 137) stores a real `$scrypt$...` verifier, but the `serverSupportsPasswordHash` branch (line 141) stores `computeHash(password, passwordSalt)` — a legacy `salt+hash` string. This emit doesn't distinguish them.

So on a server that advertises `SupportsPasswordHash` but not challenge-response, completing a forgot-password reset writes the legacy hash into the saved server entry. On the next connect `DlgConnect::updateDisplayInfo` checks `stored.startsWith("$")`, sees false, and drops the hash into `passwordEdit` as if it were the plaintext password. The client then hashes the hash, login fails, and the user is staring at 120 characters of hex in their password box with no idea where it came from.

Guard on the branch that actually produced a verifier — track it with a flag alongside `hashedPassword`, or reuse `derivedVerifier`/`pendingVerifier`, which already carry that meaning:

```cpp
if (!pendingVerifier.isEmpty()) {
    emit sigPasswordVerifierReady(lastHostname, userName, pendingVerifier);
    pendingVerifier.clear();
}
```

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -356,6 +356,41 @@ AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_Prot
...
+                    }
+                    const QByteArray nonce = QByteArray::fromBase64(parts.at(2).toUtf8());
+                    const QByteArray response = QByteArray::fromBase64(parts.at(3).toUtf8());
+                    if (nonce.isEmpty() || response.isEmpty() || !handler->isAuthNonceValid(nonce)) {
+                        return NotLoggedIn;
+                    }
+
+                    QByteArray key;
+                    if (PasswordHasher::isLegacyFormat(correctPasswordSha512)) {
+                        key = correctPasswordSha512.toUtf8();
```
ZeldaZach, line 382: **Design question, not a defect — but it decides what this PR can claim.**

The HMAC key is the stored credential itself: `verifier.verifier` for scrypt rows, and the whole stored `salt+hash` string for legacy ones. Authentication is `HMAC(stored_value, nonce)` and nothing else. The server never runs scrypt during login — it only hands the parameters to the client.

The consequence is that whatever sits in `password_sha512` is directly sufficient to authenticate. Someone with a database dump doesn't crack anything; they replay the value they already have. Today, with salted SHA-512, that same dump at least requires a cracking pass, and a user with a strong password survives it. After this change, no password survives a dump, strong or otherwise. And because `ConnectionController::onPasswordVerifierReady` writes the same value into the client's settings file, that file now holds a bearer credential too.

So the honest summary is narrower than the PR title:

- **Genuinely fixed:** the plaintext password no longer sits in the client config, and it never crosses the wire on login. That kills password-reuse blast radius, which is the highest-frequency real-world harm here. Worth having on its own.
- **Not fixed, arguably worse:** database compromise. `servatrice.ini.example` says accounts are "migrated" for security and the PR title says "stop storing plaintext passwords", both of which read as protection against a server breach. They aren't.

This is the well-known ceiling on naive hash-based challenge-response, and escaping it means an augmented PAKE — SRP-6a, or OPAQUE if you want the modern answer — where the server's stored verifier genuinely cannot be replayed. That's a much larger change and I'm not asking for it in this PR.

What I would ask for is that the choice be explicit rather than incidental. Either:

**(a)** Keep the design and say so plainly — in the PR description, in the ini comments, and ideally in a comment right here — that the stored verifier is password-equivalent, that this protects the client at rest and the wire but not the database, and that a breach still means resetting every credential. Operators reading `authentication_strictness` deserve to know that.

**(b)** Take the middle path: keep challenge-response for the wire, but have the server store `H(verifier)` and require the client to prove knowledge of `verifier` in a way the stored value can't reproduce. It doesn't reach PAKE-grade, and it costs a second round trip, but it restores the property that a dump isn't immediately usable.

(a) is a perfectly reasonable call for this codebase — I'd just rather it be a decision on the record than something a future reader has to reverse-engineer from this line.

servatrice/migrations/servatrice_0036_to_0037.sql
```diff
@@ -0,0 +1,5 @@
+-- Servatrice db migration from version 36 to version 37
+
+ALTER TABLE `cockatrice_users` MODIFY `password_sha512` char(255) NOT NULL, ALGORITHM=INSTANT;
```
ZeldaZach: `ALGORITHM=INSTANT` won't work for this change. MySQL 8.0 supports INSTANT for a narrow set of operations — adding columns, renaming, some VARCHAR *extensions* within the same length-byte class — but changing a `CHAR` column's width is not one of them; it requires a table rebuild. When the requested algorithm isn't supported MySQL raises `ER_ALTER_OPERATION_NOT_SUPPORTED` rather than quietly falling back, so this migration errors out instead of running.

Dropping the clause entirely lets the server pick, which is what you want:

```sql
ALTER TABLE `cockatrice_users` MODIFY `password_sha512` varchar(255) NOT NULL;
```

I'd also switch `char(255)` to `varchar(255)` in both this file and `servatrice.sql`. `CHAR(255)` under `utf8mb4` reserves 1020 bytes per row for a column holding 131 characters, and MySQL strips trailing spaces from `CHAR` on retrieval — harmless for base64 today, but a sharp edge to leave in a credential column.

Also missing a trailing newline (`\ No newline at end of file`), which the other migration files have.

servatrice/servatrice.ini.example
```diff
@@ -103,6 +103,16 @@ password=123456
 ; Accept only registered users? default is false (accept unregistered users)
 regonly=false
 
+[security]
+
+; How strictly new authentication features are enforced.  Possible values:
+; * legacy: only accept the legacy 1000-round SHA-512 password hashes;
+; * mixed: accept both legacy hashes and challenge-response authentication (default);
+; * strict: only accept challenge-response authentication from clients that support it,
+;           and reject plain password submissions.  Legacy accounts are migrated to PBKDF2
```
ZeldaZach: Three things in this block:

1. **Says PBKDF2, code does scrypt.** "Legacy accounts are migrated to PBKDF2 automatically" — the implementation is scrypt throughout. Same leftover appears in `session_commands.proto:235`, where the example verifier is `$pbkdf2-sha512$210000$<salt>$<verifier>`. Both look like remnants of an earlier draft and will mislead anyone matching the docs against a stored value.

2. **Duplicate `[security]` section.** This adds a second `[security]` header at line 106; one already exists at line 347. `QSettings` merges duplicate sections when reading so it works, but it's confusing, and it will conflict with #7075, which adds its rate-limit keys under the section at 347. Moving these keys down to the existing section resolves both.

3. **`legacy` needs a warning.** Per the comment on `serversocketinterface.cpp:3080`, switching back to `legacy` after any account has migrated locks those users out. If that behaviour stays, this text should say so; ideally the mode stops being a trap instead.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
```diff
@@ -269,7 +304,47 @@ void RemoteClient::passwordSaltResponse(const Response &response)
...
+                    emit loginError(Response::RespClientUpdateRequired,
+                                    QStringLiteral("This account must be logged in with its password once."), 0, {});
+                    return;
+                }
+                key = PasswordHasher::computeHash(password, passwordSalt).toUtf8();
+            } else if (!password.isEmpty()) {
+                const int n = resp.has_n() ? resp.n() : SCRYPT_N;
+                const int r = resp.has_r() ? resp.r() : SCRYPT_R;
+                const int p = resp.has_p() ? resp.p() : SCRYPT_P;
+                key = PasswordHasher::deriveKey(password, QByteArray::fromBase64(passwordSalt.toUtf8()), n, r, p);
```
ZeldaZach, line 346: Nit, and only relevant once `deriveKey` stops calling `qFatal`: scrypt at N=32768, r=8 takes on the order of 50-100 ms and 32 MB on a typical desktop, and this runs synchronously on the GUI thread inside a response handler. The UI will visibly hang for that long on every login.

Not a blocker at these parameters, but if the ceiling from the other comment ends up generous, it's worth pushing the derivation onto `QtConcurrent::run` or a worker so a slow server-side parameter choice doesn't freeze the client.

servatrice/src/serversocketinterface.cpp
```diff
@@ -3038,17 +3065,40 @@ Response::ResponseCode AbstractServerSocketInterface::cmdRequestPasswordSalt(con
...
+        }
+        re->set_password_salt(QString(verifier.salt.toBase64()).toStdString());
+        re->set_n(verifier.n);
+        re->set_r(verifier.r);
+        re->set_p(verifier.p);
+        re->set_needs_migration(false);
+    }
+
+    if (challengeResponseEnabled) {
+        const QByteArray nonce = CryptoUtil::randomBytes(32);
```
ZeldaZach, line 3124: Nit: `REQUEST_PASSWORD_SALT` is unauthenticated and generates 32 CSPRNG bytes plus a database round trip per call, with no rate limiting. #7075 covers login, registration and forgot-password, but not this command, so it stays open as a cheap way to make the server do work. Worth adding a `salt:` key to the same limiter once #7075 lands.

The response also distinguishes registered from unregistered users, and now additionally leaks whether an account has migrated and at what cost parameters. The existence leak predates this PR, so not something to fix here — just noting the surface got slightly wider.

servatrice/src/serversocketinterface.cpp
```diff
@@ -257,6 +263,9 @@ Response::ResponseCode AbstractServerSocketInterface::processExtendedSessionComm
             return cmdReportAddComment(cmd.GetExtension(Command_ReportAddComment::ext), rc);
         case SessionCommand::REPORT_DETAILS:
             return cmdReportDetails(cmd.GetExtension(Command_ReportDetails::ext), rc);
+        case SessionCommand::SUBMIT_PASSWORD_VERIFIER:
+            return cmdSubmitPasswordVerifier(cmd.GetExtension(Command_SubmitPasswordVerifier::ext), rc);
```
ZeldaZach, line 269: Nit: unreachable `break` after the `return` on the following line. `desktop-lint` may or may not care, but the other cases in this switch don't have it.

cockatrice/src/client/network/connection_controller/remote_connection_controller.cpp
```diff
@@ -91,16 +94,36 @@ void ConnectionController::connectToServer()
...
         remoteClient->connectToServer(dlgConnect->getHost(), static_cast<unsigned int>(dlgConnect->getPort()),
                                       dlgConnect->getPlayerName(), dlgConnect->getPassword());
     }
 }
 
+void ConnectionController::onPasswordVerifierReady(const QString &hostname,
+                                                   const QString &userName,
+                                                   const QString &verifier)
+{
+    Q_UNUSED(hostname);
```
ZeldaZach: Nit: the signal carries `hostname` and `userName` and both are immediately discarded in favour of `pendingSaveName`, which was captured back when the connection started.

They're consistent in the normal flow, but they're exactly the values you'd want to check before writing a credential into a saved server entry — if a reconnect or an auto-connect ever races with a manual connect, `pendingSaveName` could point at a different server than the one that just authenticated, and the failure mode is a verifier written to the wrong entry. Either verify them against the pending target, or drop the parameters from the signal so it doesn't look like a check was intended and forgotten.

cockatrice/src/interface/widgets/dialogs/dlg_connect.cpp
```diff
@@ -333,10 +340,13 @@ void DlgConnect::actOk()
         }
 
         servers.addNewServer(saveEdit->text().trimmed(), hostEdit->text().trimmed(), portEdit->text().trimmed(),
-                             playernameEdit->text().trimmed(), passwordEdit->text(), savePasswordCheckBox->isChecked());
+                             playernameEdit->text().trimmed(),
+                             passwordEdit->text().isEmpty() ? storedVerifier : passwordEdit->text(),
```
ZeldaZach: Nit: `passwordEdit->text().isEmpty() ? storedVerifier : passwordEdit->text()` means that if a user with a stored verifier types anything into the password box, the plaintext gets written to the settings file and the verifier is dropped. `onPasswordVerifierReady` overwrites it after the next successful login, so it self-heals — but if that login fails, or the client is closed first, the plaintext stays on disk. Given that removing plaintext from the config is the main user-visible benefit of this PR, it's worth closing that window.

tests/password_hash_test.cpp
```diff
@@ -36,6 +36,74 @@ TEST(PasswordHashTest, TokenHasExpectedLength)
     const QString token = PasswordHasher::generateActivationToken();
     ASSERT_EQ(token.size(), 16);
 }
+
+TEST(PasswordHashTest, DeriveKeyMatchesKnownVector)
```
ZeldaZach, line 40: The hasher tests are good — checking `deriveKey` against the RFC 7914 vector is exactly the right anchor, and the round-trip and invalid-input cases cover `parsePasswordVerifier` well.

What's missing is everything above that layer, which is where all five blocking findings live. None of them would survive a test:

- nonce single-use (a second login with the same nonce must fail) and 60-second expiry
- `checkUserPassword` accepting a correct challenge response and rejecting a wrong one, for both scrypt and legacy rows
- `changeUserPassword` against a scrypt row — this alone would have caught the `left(16)` bug
- `parsePasswordVerifier` rejecting hostile cost parameters (negative, non-power-of-two, absurdly large)

The first two need a `Server_ProtocolHandler`, so they may not be reachable from the current test setup. The last two are pure functions and cheap to add — I'd want at least those before merge, since they cover two of the blockers directly.

ZeldaZach (changes requested):
Reviewed in a local worktree, diffed against `cleanup/2-fix-csprng` (#7192, which I've approved separately). This is a large and mostly careful piece of work — the scrypt wrapper is correct and validated against the RFC 7914 vector, `constantTimeEquals` uses `CRYPTO_memcmp`, the nonce is single-use and time-bounded, and the three-way `legacy`/`mixed`/`strict` rollout is the right way to stage something like this. The client-side win is real and worth having: the plaintext password stops living in the settings file.

That said, I found five things that I think block merge, plus one design question worth settling explicitly before this ships, since it determines what the PR can honestly claim.

### Blocking

1. **In-app password change breaks permanently for every migrated account.** `changeUserPassword` derives the comparison salt as `correctPasswordSha512.left(16)`, which for a scrypt row is the literal string `$scrypt$32768$8$`. The legacy hash it computes can never equal the stored `$scrypt$...` value, so `cmdAccountPassword` returns `RespWrongPassword` forever. Every account hits this the first time it tries to change its password after migrating.
2. **A server can hard-crash any client** via unvalidated scrypt cost parameters plus `qFatal` in `deriveKey`.
3. **`cmdSubmitPasswordVerifier` stores unvalidated client input**, and the `not like '$%'` guard in the UPDATE makes the resulting lockout unrecoverable.
4. **Switching a server back to `legacy` bricks already-migrated accounts.**
5. **`submitForgotPasswordResetResponse` can persist a legacy hash as if it were a verifier**, corrupting the saved credential.

### Design question

6. The stored verifier is password-equivalent — it is used directly as the HMAC key, on both ends. A servatrice database dump therefore yields immediately usable login credentials for every account, with no cracking step at all. That's a real change in the threat model and it points the opposite way from how the PR and the ini text describe it. Details in the inline comment on `servatrice_database_interface.cpp`; I don't think it necessarily blocks merge, but the framing needs to change or the design does.

Smaller items are inline and marked as nits. Happy to talk through any of these — particularly (6), which is a judgement call rather than a defect.

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. `changeUserPassword` now routes through a new `PasswordHasher::verifyPassword()`: for a `$scrypt$` row it parses the stored n/r/p and salt, re-derives with the client-supplied password, and compares constant-time; legacy rows keep the old `left(16)`-salt + `computeHash` path. `oldPasswordNeedsHash=false` still short-circuits directly.

Covered by the new `VerifyPasswordScryptRow` / `VerifyPasswordLegacyRow` tests in `tests/password_hash_test.cpp`.

On the `user_info_box.cpp` point: I kept plaintext-over-TLS for the old password deliberately, to keep the change small — the server-side `verifyPassword` is now bounded by `costParamsAreSane`, so the scrypt derivation can't be turned into a server DoS either.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
BruebachL: Fixed. `deriveKey` now returns `QByteArray()` on any `EVP_PBE_scrypt` failure instead of `qFatal`, and every caller treats an empty return as a failure (no false-success paths).

Added `PasswordHasher::costParamsAreSane(n, r, p)` with exactly the clamp you suggested — `n` in `[1024, 2^20]`, power of two, `r` in `[1, 32]`, `p` in `[1, 16]` — and applied it in two places: `parsePasswordVerifier` so hostile DB rows are rejected at parse time, and the client's `passwordSaltResponse` handler, which now validates before deriving (and surfaces `loginError` on a bad cast of parameters instead of hanging or crashing).

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. `cmdSubmitPasswordVerifier` now parses before storing:

```cpp
const auto parsed = PasswordHasher::parsePasswordVerifier(passwordVerifier);
if (!parsed.isValid || !costParamsAreSane(parsed.n, parsed.r, parsed.p)) {
}
```

so `$x` and anything else that isn't a real verifier never reaches the UPDATE and the lockout-you-can't-repair path is gone.

The two smaller items too:
- `MAX_NAME_LENGTH` is replaced with a named `MAX_PASSWORD_VERIFIER_LENGTH` (255) constant — you're right, the intent reading was muddled; a real scrypt verifier is ~131 chars and the constant is there for a hard ceiling, not because a username limit happened to fit.
- `submitPasswordVerifier` now returns `query->numRowsAffected() > 0`, so a zero-row UPDATE reports failure instead of `RespOk` and the client won't persist a verifier the server never stored.

On the register-at-`n=1024`-minimum note: agreed it only doesn't matter today because the server never runs scrypt during login. Since `verifyPassword` in `changeUserPassword` now *does* run server-side scrypt, capping cost params there is meaningful and covered.

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed — went with your second option. Challenge-response is now served for scrypt rows in **all** strictness modes; `legacy` only governs what *new* credentials are accepted, not whether existing scrypt rows work.

Concretely:
- `cmdRequestPasswordSalt` sends a nonce for scrypt rows regardless of mode; legacy rows only get salt (no nonce) outside `legacy`.
- `initSession` always advertises both `SupportsPasswordHash` and `SupportsChallengeResponseAuth` under SQL auth, so clients running against a `legacy` server still auto-derive scrypt immediately.
- A new `acceptsCredentialFormat()` gate governs register / account-password / forgot-reset: `strict` requires scrypt, `legacy` and `mixed` accept either. This keeps `legacy` from dead-ending current clients, which always send scrypt verifiers when the flag is advertised — the hole my first pass had.

So `legacy` now reads as "don't migrate anyone new", and existing scrypt accounts keep working even after a rollback. Documented the semantics in `servatrice.ini.example`, including that the setting is effectively one-way per your earlier note.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
BruebachL: Fixed. `derivedVerifier` is now set only in the challenge-response branch; the `serverSupportsPasswordHash` legacy branch no longer feeds it. `submitForgotPasswordResetResponse` emits `sigPasswordVerifierReady` only when `derivedVerifier` is non-empty, so the legacy `salt+hash` string can never be emitted as if it were a verifier — the `$`-prefix check in `DlgConnect::updateDisplayInfo` then also correctly wouldn't see one.

Also dropped the now-unneeded `hostname`/`userName`/`password` parameters from the signal per the related nit — it's just `(const QString &verifier)` now, and the one call site uses the connection-scoped pending target.

servatrice/src/servatrice_database_interface.cpp
BruebachL: Choosing **(a)**, on the record.

The design is kept and now documented honestly in three places:
- A comment at the `checkUserPassword` challenge branch in `servatrice_database_interface.cpp` spelling out that the stored verifier is password-equivalent and that a DB dump replays it as-is — HMAC of the stored value is authenticating, nothing more.
- `servatrice.ini.example` now carries the same disclosure instead of implying migration equals breach-survival: scrypt rows are password-equivalent/bearer credentials; this PR protects the client at rest and the wire, **not** the database; a breach still means resetting every credential.
- The PR description is being tightened to match — "stop storing plaintext" is framed as client-side + wire, with the DB-compromise ceiling called out.

You're right that (b) only looks like it restores the dump-survival property and would cost a round trip for a property it doesn't actually deliver — I didn't want to trade complexity for an illusory improvement. SRP-6a/OPAQUE noted as the real escape hatch if we ever need dump-survival; deliberately out of scope here.

servatrice/migrations/servatrice_0036_to_0037.sql
BruebachL: Fixed. Dropped `ALGORITHM=INSTANT` so the server picks the algorithm, switched `char(255)` → `varchar(255)` in both this migration and `servatrice.sql`, and added the trailing newline.

servatrice/servatrice.ini.example
BruebachL: All three addressed:

1. **PBKDF2 → scrypt** — `servatrice.ini.example` no longer mentions PBKDF2; the example in `session_commands.proto` is now `$scrypt$32768$8$1$<salt>$<verifier>`.
2. **Duplicate `[security]`** — the new block was folded into the existing section at line ~337 (which also keeps it clear of #7075's rate-limit keys). Only one `[security]` remains.
3. **`legacy` warning** — the ini documents that legacy accounts are never auto-migrated and that scrypt rows continue to be served challenge-response in all modes (per the fix to `cmdRequestPasswordSalt`), so the "roll back to legacy" trap is gone rather than merely warned about.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
BruebachL: Agreed it's not a blocker at N=32768/r=8, and the `costParamsAreSane` ceiling now bounds the worst case a hostile server can ask for (n ≤ 2^20, r ≤ 32 → a few hundred MB on the derive side, and the salt handler rejects beyond that). Moving the derivation to `QtConcurrent`/a worker is deferred rather than deleted — noted in the PR description as a follow-up so it isn't lost.

servatrice/src/serversocketinterface.cpp
BruebachL: Agreed on both counts and not fixing the rate limit here — it belongs with #7075's limiter work. I've noted a `salt:` key alongside the login/registration/forgot entries as a follow-up so REQUEST_PASSWORD_SALT doesn't stay open forever. The widened metadata leak (migration state + cost params via the salt response) is documented in the PR description's follow-up list; agreed the existence leak predates this PR.

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed — removed the unreachable `break` after the `return` in the SUBMIT_PASSWORD_VERIFIER case.

cockatrice/src/client/network/connection_controller/remote_connection_controller.cpp
BruebachL: Fixed — dropped `hostname` and `userName` from `sigPasswordVerifierReady` entirely (it's now `(const QString &verifier)`). The receiver writes to the connection-scoped pending target, and since the signal carries only the verifier there's no set of parameters that could look like an intended-but-forgotten consistency check.

cockatrice/src/interface/widgets/dialogs/dlg_connect.cpp
BruebachL: Fixed — `actOk` now writes `storedVerifier.isEmpty() ? passwordEdit->text() : storedVerifier`, so typing into the password box can no longer push plaintext into the settings entry; the stored verifier survives until a successful key exchange overwrites it. The plaintext-on-disk regression window is closed regardless of whether the next login succeeds.

tests/password_hash_test.cpp
BruebachL: Agreed the hasher layer alone wasn't enough — added the two pure-function tests that cover blockers directly:

- `CostParamsAreSane` — the n/r/p clamp itself.
- `ParsePasswordVerifierRejectsHostileCostParams` — negative / non-power-of-two / absurdly large `n` all rejected at parse time.
- `VerifyPasswordScryptRow` — the regression that would have caught the B1 `left(16)` bug: `changeUserPassword` against a scrypt row.
- `VerifyPasswordLegacyRow` — the legacy path still verifies.

The `checkUserPassword` accept/reject and nonce single-use/expiry cases need `Server_ProtocolHandler` and the DB layer, which the current test setup doesn't wire up — noted in the PR description as a follow-up rather than blocked here.

servatrice/src/serversocketinterface.cpp
```diff
@@ -2507,6 +2522,21 @@ bool AbstractServerSocketInterface::tooManyRegistrationAttempts(const QString &i
     return false;
 }
 
+bool AbstractServerSocketInterface::acceptsCredentialFormat(bool passwordNeedsHash, const QString &password) const
+{
+    // "scryptFormat" means the client sent a derived verifier rather than a password to hash ourselves.
+    const bool scryptFormat = !passwordNeedsHash && !PasswordHasher::isLegacyFormat(password);
```
ZeldaZach: **Blocking: the unvalidated-verifier lockout you fixed in `cmdSubmitPasswordVerifier` is still open in the three other paths that write `password_sha512`.**

`scryptFormat` is `!passwordNeedsHash && !isLegacyFormat(password)`, and `isLegacyFormat` is only `!startsWith("$")`. So `"$x"` is `scryptFormat == true` and is accepted by *every* mode — `strict` included — and stored verbatim by `cmdRegisterAccount` (2475), `cmdAccountPassword` (2875) and `cmdForgotPasswordReset` (3017). None of those call `parsePasswordVerifier`.

The consequences are the same ones that made this a blocker last round: `cmdRequestPasswordSalt` hits `!verifier.isValid` and returns `RespContextError` on every subsequent attempt, and `submitPasswordVerifier` can't repair it because the row now matches `'$%'`. Registration is the worst case — it's unauthenticated on any server with `registration/enabled`, so `hashed_password = "$x"` mints an account that no honest client can ever log into and that permanently holds the username.

Second problem in the same helper: an **empty** credential is accepted. `isLegacyFormat("")` is true, so `scryptFormat` is false and `legacy`/`mixed` fall through to `return true`. `registerUser` / `changeUserPassword` then store `''`. See my comment on `servatrice_database_interface.cpp:375` for why an empty row is now much worse than it used to be.

Both fall out of one change here:

```cpp
if (password.isEmpty()) {
    return false;
}
const bool scryptFormat =
    !passwordNeedsHash && PasswordHasher::parsePasswordVerifier(password).isValid;
```

`parsePasswordVerifier` already enforces `costParamsAreSane`, so that also closes the register-at-`n=1024` note from last round for these paths. With `parsePasswordVerifier` here, `strict` keeps its current meaning and `legacy`/`mixed` need `scryptFormat || PasswordHasher::isLegacyFormat(password)` to keep accepting legacy hashes — the point is that nothing starting with `$` reaches the database without parsing.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -356,6 +356,47 @@ AuthenticationResult Servatrice_DatabaseInterface::checkUserPassword(Server_Prot
...
```
ZeldaZach, line 382: **Blocking: this fails open on an empty stored credential.**

If `password_sha512` is `''`, `isLegacyFormat("")` is true and `key` becomes a zero-length `QByteArray`. `computeResponse` then HMACs the nonce under an empty key, and anyone who can compute `HMAC-SHA256("", nonce)` authenticates as that account.

Getting the nonce is not a barrier: `cmdRequestPasswordSalt` is unauthenticated and the nonce is stored on the session without being bound to the username it was issued for, so requesting a salt for *any* existing account yields a nonce usable in a `cmdLogin` for the empty-credential account.

What makes this a regression rather than a theoretical note is that the old code failed closed on the same row — `computeHash(password, "")` returns `salt + base64(...)`, which is never `""`, so an empty row simply meant "nobody can log in". After this PR it means "everybody can log in". And per my comment on `acceptsCredentialFormat`, an empty value is reachable through `cmdAccountPassword` and `cmdForgotPasswordReset` today.

Credential comparison should never treat an absent secret as a usable key:

```cpp
if (correctPasswordSha512.isEmpty()) {
    return NotLoggedIn;
}
```

right before the `$challenge$` branch, and a `key.isEmpty()` guard before `computeResponse` for belt and braces. Fixing `acceptsCredentialFormat` closes the reachable path, but the fail-closed check belongs here regardless — this is the one place in the codebase where getting it wrong is an authentication bypass, and it shouldn't depend on three callers upstream staying correct.

While you're here: binding the nonce to the requested username in `setAuthNonce` (store the name alongside it, compare in `isAuthNonceValid`) costs nothing and removes the cross-account nonce transfer that turns the above into a remote attack rather than a self-inflicted one.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
```diff
@@ -269,7 +307,59 @@ void RemoteClient::passwordSaltResponse(const Response &response)
         auto passwordSalt = QString::fromStdString(resp.password_salt());
         if (passwordSalt.isEmpty()) { // the server does not recognize the user but allows them to enter unregistered
             password.clear();         // the password will not be used
+            storedVerifier.clear();
             doLogin();
+        } else if (serverSupportsChallengeResponse && resp.has_nonce()) {
+            const QByteArray nonce = QByteArray::fromStdString(resp.nonce());
+            QByteArray key;
+            if (resp.needs_migration()) {
```
ZeldaZach, line 326: **Strong suggestion (not blocking): `needs_migration` is a server-controlled downgrade from scrypt to the 1000-round hash.**

This branch is checked before the `storedVerifier` branch, and `needs_migration` is a bare bool from the wire. A server — or, since `RemoteClient` uses a plain `QTcpSocket` with no TLS on the default port, an active MITM — sets it on a scrypt account and the client keys the HMAC with `computeHash(password, <server-chosen salt>)` instead of `deriveKey(...)`. The attacker gets `HMAC(SHA512_1000(salt||password), nonce)` for a salt they picked: an offline dictionary attack at roughly 1000 SHA-512 per guess against a password scrypt was supposed to be protecting. `loginResponse` then calls `doSubmitPasswordVerifier()` on top of that.

Against the PR's claim that this "defeats both replay attacks and any passive capture of login traffic", a flag the peer controls that reverts the KDF is worth closing. It's cheap here — the client already knows when it should not be downgraded:

```cpp
if (resp.needs_migration() && PasswordHasher::parsePasswordVerifier(storedVerifier).isValid) {
    emit loginError(Response::RespClientUpdateRequired,
                    QStringLiteral("Server asked to downgrade authentication for a migrated account."), 0, {});
    return;
}
```

That covers every user who has logged in once with "save password" on. Users without a stored verifier can't distinguish a genuine legacy account from a spoofed one, which is inherent to server-driven mode selection over an unauthenticated transport — worth a line in the ini/description noting the wire guarantees assume TLS termination in front of the server.

libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
```diff
@@ -44,6 +44,22 @@ Server_ProtocolHandler::~Server_ProtocolHandler()
...
 
+void Server_ProtocolHandler::setAuthNonce(const QByteArray &nonce)
+{
+    authNonce = nonce;
+    authNonceCreated = QDateTime::currentDateTimeUtc();
+}
+
+bool Server_ProtocolHandler::isAuthNonceValid(const QByteArray &nonce) const
+{
+    return !authNonce.isEmpty() && authNonce == nonce && authNonceCreated.secsTo(QDateTime::currentDateTimeUtc()) < 60;
```
ZeldaZach: Nit: `secsTo` returns a signed value, so a backwards clock step (NTP correction, VM resume) makes the elapsed time negative and the `< 60` check passes — an old nonce becomes valid again. `authNonceCreated.isValid() && elapsed >= 0 && elapsed < 60` closes it, or use a monotonic `QElapsedTimer` instead of a wall clock, which is what you actually want for an expiry.

Also worth a named constant for the 60 rather than a literal inside the boolean.

servatrice/migrations/servatrice_0036_to_0037.sql
```diff
@@ -0,0 +1,8 @@
+-- Servatrice db migration from version 36 to version 37
+
+-- The column must hold "$scrypt$<n>$<r>$<p>$<salt>$<verifier>" (up to ~255 chars) and arbitrary
+-- legacy base64 hashes, so it grows beyond the old 120-char size. varchar(255) is used as the
+-- column type is promotion-safe and avoids the row-format change ALGORITHM=INSTANT cannot do.
+ALTER TABLE `cockatrice_users` MODIFY `password_sha512` varchar(255) NOT NULL;
+
+UPDATE cockatrice_schema_version SET version=37 WHERE version=36;
```
ZeldaZach: The trailing newline is still missing — the file ends at the `;` with no `\n`, same as last round. `ALGORITHM=INSTANT` is correctly gone and `varchar(255)` is applied in both places, so this is the only leftover from that comment.

(Minor wording: the comment says varchar "avoids the row-format change ALGORITHM=INSTANT cannot do" — the reason the clause was dropped is that widening a `CHAR` needs a rebuild that INSTANT can't perform, which reads slightly backwards as written.)

cockatrice/src/client/network/connection_controller/remote_connection_controller.cpp
```diff
@@ -91,16 +94,32 @@ void ConnectionController::connectToServer()
     connect(dlgConnect, &DlgConnect::sigStartForgotPasswordRequest, this, &ConnectionController::forgotPasswordRequest);
 
     if (dlgConnect->exec()) {
+        pendingSaveName = dlgConnect->getSaveName();
```
ZeldaZach, line 97: Nit: `getSaveName()` returns `saveEdit->text()` untrimmed, but `DlgConnect::actOk()` saves the profile under `saveEdit->text().trimmed()`. If the name has leading/trailing whitespace, `setServerPassword`'s `getPrevioushostindex(saveName)` lookup misses, returns `< 0`, and the verifier is silently dropped — the user keeps reauthenticating with a stale credential and nothing reports why. `.trimmed()` here, or trim inside `getSaveName()`.

Worth logging a warning in `ServersSettings::setServerPassword` when the index lookup fails, too — a silent no-op on a credential write is hard to diagnose from a bug report.

ZeldaZach (changes requested):
Re-reviewed against the new head. All five blockers from my last pass are genuinely fixed, and I want to say so clearly before the new findings:

1. `changeUserPassword` now routes through `verifyPassword()` with a real scrypt branch — the `left(16)` bug is gone and `VerifyPasswordScryptRow` locks it down.
2. `deriveKey` returns `QByteArray()` instead of `qFatal`, and `costParamsAreSane` is applied in both `parsePasswordVerifier` and the client's salt handler.
3. `cmdSubmitPasswordVerifier` parses before storing, `MAX_PASSWORD_VERIFIER_LENGTH` replaces the reused name limit, and `submitPasswordVerifier` checks `numRowsAffected() > 0`.
4. `legacy` no longer bricks migrated accounts — scrypt rows get challenge-response in every mode, and `initSession` always advertises both flags. The `legacy`-means-"don't migrate anyone new" reading is the right one.
5. `derivedVerifier` is set only in the challenge-response branch, so the legacy hash can no longer be persisted as a verifier.

The design question is settled explicitly with option (a), documented in the code, the ini, and the description. That's what I asked for. CI is green across all 18 jobs.

Unfortunately the fix for (3) was applied to `cmdSubmitPasswordVerifier` only, and the *same* class of bug is still live in the three other write paths through the new `acceptsCredentialFormat()` helper. There's also a fail-open in the challenge branch that is new in this PR. Two blockers, one strong suggestion, three nits — all inline.

Happy to re-review quickly once the validation is hoisted into one place; the rest of this is in good shape.

servatrice/src/serversocketinterface.cpp
BruebachL: I've hardened `acceptsCredentialFormat` to close the same lockout in all four write paths. An empty credential is now rejected outright, and `scryptFormat` is `!passwordNeedsHash && PasswordHasher::parsePasswordVerifier(password).isValid`, so anything starting with `$` is fully parsed (including the sane-cost clamp) before it can reach `password_sha512` — `$x` is rejected everywhere, not just in `cmdSubmitPasswordVerifier`. Legacy/mixed paths now accept a valid scrypt verifier or a genuine `isLegacyFormat` salt+hash.

servatrice/src/servatrice_database_interface.cpp
BruebachL: I've added a fail-closed guard in `checkUserPassword` immediately after the row is fetched: if `correctPasswordSha512` is empty it returns `NotLoggedIn` (with a `DatabaseInterfaceLog` warning) before any key is built, so an empty stored credential can no longer authenticate `HMAC("", nonce)`.

libcockatrice_network/libcockatrice/network/client/remote/remote_client.cpp
BruebachL: I've added the downgrade guard before the `needs_migration` branch in `RemoteClient::passwordSaltResponse`: if the server says `needs_migration` but we already hold a valid scrypt verifier (`PasswordHasher::parsePasswordVerifier(storedVerifier).isValid`), we refuse with `RespClientUpdateRequired` instead of falling back to the 1000-round legacy hash. `storedVerifier` is only ever set by the challenge-response exchange (the legacy `salt+hash` server-hash branch never feeds it), so a legitimate first login of an account we have no verifier for still works.

libcockatrice_network/libcockatrice/network/server/remote/server_protocolhandler.cpp
BruebachL: Applied the signed-elapsed fix with a named constant in `server_protocolhandler.cpp`: `kAuthNonceLifetimeSeconds = 60`, and `isAuthNonceValid` now requires `authNonceCreated.isValid() && elapsed >= 0 && elapsed < kAuthNonceLifetimeSeconds`, so a backwards clock step (NTP correction, VM resume) can't re-validate an old nonce. I also bound the nonce to the `userName` it was issued for (`setAuthNonce(nonce, userName)` / `isAuthNonceValid(nonce, userName)`), preventing a nonce issued for one user being replayed for another.

servatrice/migrations/servatrice_0036_to_0037.sql
BruebachL: Trailing newline added (file now ends at `version=36;` + `\n`). Comment reworded the right way round: `varchar(255)` is used because widening a `CHAR` requires a table rebuild that `ALGORITHM=INSTANT` cannot perform, so the clause is dropped and the server picks a suitable algorithm.

cockatrice/src/client/network/connection_controller/remote_connection_controller.cpp
BruebachL: `DlgConnect::getSaveName()` now returns `saveEdit->text().trimmed()`, matching the profile key `actOk()` saves under, so `getPrevioushostindex` can't miss on leading/trailing whitespace and silently drop the verifier. I also added a `qCWarning` in `ServersSettings::setServerPassword` when `getPrevioushostindex` returns `< 0`, so a failed credential write is diagnosable rather than a silent no-op in a bug report.

ZeldaZach (approved):

tooomm:
@BruebachL Besides resolving conflicts, is there something you want to do here still?

BruebachL:
> @BruebachL Besides resolving conflicts, is there something you want to do here still?

I'm conflicted. It's a very important PR but it's also super user-facing. 

If you log in with a new client, your password gets converted, which locks you out of logging in with an old client (definitively) until you reset your password with an old client (I think).

I know that we have users that love their downgrades especially around update time.

I'm not expecting many downgrades this time around since there's hardly a controversy you can't disable this time around but I feel iffy about bundling this with 3.1.

Considering 3.1 in general, due to the recent slowdown in activity, I'm strongly considering merging in a couple of picture loader network things and just shipping it.

We'll do our usual 3.1.1 follow up release shortly after, into which I would bundle this.

tooomm:
I defer to you if you prefer to include it in 3.1 or a patch release after.

Speaking about 3.1
It seems a release is planned rather sooner than later... I think we should include my updater PR which fixes a bug and has a more robust logic in finding supported releases and correct binaries. Will have a look in the last review and the open points this weekend - was recently focused on CMake and Ninja when I had time to spend.

BruebachL:
> I defer to you if you prefer to include it in 3.1 or a patch release after.
> 
> Speaking about 3.1
> It seems a release is planned rather sooner than later... I think we should include my updater PR which fixes a bug and has a more robust logic in finding supported releases and correct binaries. Will have a look in the last review and the open points this weekend - was recently focused on CMake and Ninja when I had time to spend.

Yeah, I'm sorry you're out of the loop on this, I've been trying to maintain the milestones and such a bit better.

On the discord, I've been trying to get this release through the door for 6 weeks now.
The onboarding wizard alone is worth it, the vde and VDS performance fixes are much needed relief, there's boatloads of QoL stuff in master, game invites, and the new UI + the reporting queue need to be trialed.

A lot of things point to release, imo.

BruebachL:
Not to mention tournaments and drafting plus some refactor work would appreciate a clean cut.

We've just hit an unfortunate window where I am trying to get a lot of changes through but there is simply no one around to review them.

RickyRister:
> If you log in with a new client, your password gets converted, which locks you out of logging in with an old client (definitively) until you reset your password with an old client (I think).

I frequently boot up old versions of cockatrice to check if a behavior is a regression. 

The recent settings migration already breaks my settings whenever I boot up an older version and then go back, which makes checking for regressions quite an annoying process. I would not love if Cockatrice also broke my login whenever I boot up an old version.

PR #7182, opened 2026-08-24
[Protocol] Add tournament state event and tournament game commands

BruebachL:
## What will change with this Pull Request?
- Protobuf additions needed for tournament work

PR #7197, opened 2026-08-29
[Protocol] Add tournament state event and tournament game commands

BruebachL:
- Protobuf additions needed for tournament work

## Related
- Supersedes #7182 (fork-based PR). Previous review thread: https://github.com/Cockatrice/Cockatrice/pull/7182

### DIFF
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
@@ -1,6 +1,6 @@
 syntax = "proto2";
 import "game_commands.proto";
-import "event_tournament_state.proto";
+import "serverinfo_tournament.proto";
 
 message Command_ReportMatchResult {
     extend GameCommand {
@@ -8,7 +8,12 @@ message Command_ReportMatchResult {
     }
 
     optional sint32 game_id = 1 [default = -1];
+    // The winning player's ID. A drawn match is reported with is_draw set and
+    // winner_id left unset so a missing report stays distinguishable from a
+    // reported draw via has_winner_id().
     optional sint32 winner_id = 2 [default = -1];
+    // Whether the match ended in a draw; winner_id is meaningless then.
+    optional bool is_draw = 3;
 }
 
 message Command_AdvanceTournament {
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
@@ -1,0 +1,26 @@
+syntax = "proto2";
+import "game_commands.proto";
+import "event_tournament_state.proto";
+
+message Command_ReportMatchResult {
+    extend GameCommand {
+        optional Command_ReportMatchResult ext = 1037;
+    }
+
+    optional sint32 game_id = 1 [default = -1];
+    optional sint32 winner_id = 2 [default = -1];
+}
+
+message Command_AdvanceTournament {
+    extend GameCommand {
+        optional Command_AdvanceTournament ext = 1038;
+    }
+}
+
+message Command_TournamentSettingsSelect {
+    extend GameCommand {
+        optional Command_TournamentSettingsSelect ext = 1039;
+    }
+
+    optional TournamentSettings settings = 1;
+}
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/event_game_state_changed.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/event_game_state_changed.proto
@@ -24,4 +24,10 @@ message Event_GameStateChanged {
 
     // the amount of seconds since the game started
     optional uint32 seconds_elapsed = 5;
+
+    // whether this game is a tournament game
+    optional bool is_tournament = 9;
+
+    // for tournament sub-games: the ID of the parent tournament game (-1 if not a sub-game)
+    optional sint32 parent_game_id = 10 [default = -1];
 }
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/event_game_state_changed.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/event_game_state_changed.proto
@@ -25,9 +25,6 @@ message Event_GameStateChanged {
     // the amount of seconds since the game started
     optional uint32 seconds_elapsed = 5;
 
-    // whether this game is a tournament game
-    optional bool is_tournament = 9;
-
     // for tournament sub-games: the ID of the parent tournament game (-1 if not a sub-game)
     optional sint32 parent_game_id = 10 [default = -1];
 }
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/event_tournament_state.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/event_tournament_state.proto
@@ -1,0 +1,43 @@
+syntax = "proto2";
+import "game_event.proto";
+
+message TournamentPlayer {
+    optional sint32 player_id = 1;
+    optional string player_name = 2;
+    optional uint32 wins = 3;
+    optional uint32 losses = 4;
+    optional uint32 draws = 5;
+    optional bool deck_submitted = 6;
+}
+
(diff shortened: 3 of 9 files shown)


libcockatrice_protocol/libcockatrice/protocol/pb/room_commands.proto
```diff
@@ -69,6 +69,12 @@ message Command_CreateGame {
 
     // share decklists with all players when selected
     optional bool share_decklists_on_load = 14;
+
+    // number of games per match in tournament mode (e.g. 3 for best of 3)
+    optional uint32 games_per_match = 15 [default = 1];
```
ZeldaZach: This duplicates `TournamentSettings.games_per_match`, which also reaches the server via `Command_TournamentSettingsSelect`. Two wire representations of the same value means the server has to define which one wins, and every future tournament setting has to be added in both places.

I would rather have `optional TournamentSettings tournament_settings = 15;` here and drop the standalone field, so game creation and mid-game changes carry the identical message. (Needs `TournamentSettings` to live somewhere `room_commands.proto` can import - see the comment on `command_tournament.proto`.)

libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto2";
+import "game_commands.proto";
+import "event_tournament_state.proto";
```
ZeldaZach: This is the only `command_*.proto` in the tree that imports an `event_*.proto`, and it does so purely to reach `TournamentSettings`.

`TournamentPlayer`, `TournamentPairing` and `TournamentSettings` are plain data types rather than events - please move them into their own file (`serverinfo_tournament.proto` would match how the other shared structs are named) and have both `event_tournament_state.proto` and this file import that. Don't forget to add it to PROTO_FILES in the pb CMakeLists.

libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
```diff
@@ -0,0 +1,26 @@
...
+
+message Command_ReportMatchResult {
+    extend GameCommand {
+        optional Command_ReportMatchResult ext = 1037;
+    }
+
+    optional sint32 game_id = 1 [default = -1];
+    optional sint32 winner_id = 2 [default = -1];
```
ZeldaZach, line 14: How does a client report a drawn match here? `TournamentPlayer` has a `draws` counter, but `winner_id` has no encoding that distinguishes "the match was a draw" from "unset/not reported" - both come through as the -1 default. Either document that -1 means a draw and rely on has_winner_id for the unset case, or add an explicit field. Same question applies to `TournamentPairing.winner_id`.

libcockatrice_protocol/libcockatrice/protocol/pb/event_tournament_state.proto
```diff
@@ -0,0 +1,43 @@
...
+    optional bool deck_submitted = 6;
+}
+
+message TournamentPairing {
+    optional sint32 player1_id = 1;
+    optional sint32 player2_id = 2 [default = -1];
+    optional sint32 game_id = 3;
+    optional sint32 winner_id = 4 [default = -1];
+    optional sint32 player1_match_wins = 5 [default = 0];
+    optional sint32 player2_match_wins = 6 [default = 0];
```
ZeldaZach: These two are `sint32` while `wins`/`losses`/`draws` a few lines up are `uint32`. They are all non-negative game counts - make these `uint32` for consistency (and the sint32 zigzag encoding is pointless for values that are never negative). The `[default = 0]` is also redundant for an unset numeric field.

libcockatrice_protocol/libcockatrice/protocol/pb/event_game_state_changed.proto
```diff
@@ -24,4 +24,10 @@ message Event_GameStateChanged {
 
     // the amount of seconds since the game started
     optional uint32 seconds_elapsed = 5;
+
+    // whether this game is a tournament game
+    optional bool is_tournament = 9;
```
ZeldaZach: This message is the join snapshot ("Also used to provide the entire game state when joining a game"), but all a late joiner learns from it is that the game is a tournament - the phase, round, standings and pairings only ever arrive in a separate `Event_TournamentState`.

Can you confirm #7199 replays `Event_TournamentState` to a player/spectator on join, and not just on mutation? Otherwise anyone joining after round 1 sits on an empty bracket until the next advance. If the shared messages get their own file as suggested elsewhere, embedding the tournament state message here would make that guaranteed by construction rather than by the server remembering to resend.

ZeldaZach (changes requested):
Thanks for splitting the protocol out into its own PR, this is much easier to reason about.

I checked the numbering: extensions 1037-1039 (GameCommand), 2027 (GameEvent), Event_GameStateChanged fields 9/10, Command_CreateGame 15/16 and ServerInfo_Game 54 are all fresh, nothing is renumbered or reused, and both new files are in PROTO_FILES so they will actually build. No `required` anywhere. Good.

Most of what follows is about wire shape, which we are stuck with once it ships, so I would rather sort it out at the base of the stack than in #7199/#7200. Two structural items (the duplicated settings and the command -> event import), plus a couple of smaller ones inline.

Minor, not blocking: the new GameCommandType entries are the only ones in that enum without the `/// Server:` / `/// Client:` annotation lines the surrounding entries carry - worth adding once #7199 lands the handlers so the pointers are real. Also curious whether the gaps (1036, 2023-2026, Event_GameStateChanged 6-8) are reserved for something in the rest of the stack or just spacing.

libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
BruebachL: Done in b98a267285: `TournamentPlayer`, `TournamentPairing` and `TournamentSettings` now live in a new `serverinfo_tournament.proto`, imported by both `event_tournament_state.proto` and `command_tournament.proto` (no more command→event import). Added to PROTO_FILES.

libcockatrice_protocol/libcockatrice/protocol/pb/room_commands.proto
BruebachL: Done in b98a267285: `Command_CreateGame` now carries `optional TournamentSettings tournament_settings = 15` (importing `serverinfo_tournament.proto`) and the standalone `games_per_match` field is gone, so game creation and `Command_TournamentSettingsSelect` share one representation.

libcockatrice_protocol/libcockatrice/protocol/pb/command_tournament.proto
BruebachL: Done in b98a267285: both `Command_ReportMatchResult` and `TournamentPairing` now have an explicit `is_draw` field, documented on the wire — a drawn match is reported with `is_draw` set and `winner_id` unset, so a missing report (`has_winner_id()` false, `is_draw` unset) stays distinguishable from a reported draw.

libcockatrice_protocol/libcockatrice/protocol/pb/event_tournament_state.proto
BruebachL: Done in b98a267285: `player1_match_wins`/`player2_match_wins` are now `uint32` (matching `wins`/`losses`/`draws`) and the redundant `[default = 0]` is dropped.

libcockatrice_protocol/libcockatrice/protocol/pb/event_game_state_changed.proto
BruebachL: Confirmed and fixed in the server engine (#7199, commit b3c7811e0e): the join snapshot in `Server_Game::createGameJoinedEvent` now enqueues a replay of the current `Event_TournamentState` (`tournament->buildStateEvent()`) alongside `Event_GameStateChanged`, so a player or spectator joining after round one immediately receives the phase, round, standings and pairings instead of sitting on an empty bracket until the next advance. Building was extracted into `buildStateEvent()` so the join and mutation paths share one source of truth.

PR #7199, opened 2026-08-29
[Server] Implement Swiss tournament engine for lobby games

BruebachL:

Lobby games have no tournament flow — no Swiss pairing, round advancement, or automatic match-result handling between sub-games.

- Add `Server_Tournament` Swiss engine: greedy pairing with rematch avoidance, byes, ceil(log2(n)) rounds, and best-of-N series via `gamesPerMatch`
- Spawn match sub-games through the parent game's `Server_MatchGameFactory`, auto-joining players and installing their submitted decks
- Add lifecycle and match-result strategies: the hub game start is gated until all decks are submitted; sub-game outcomes are reported back to the parent
- Extend `Server_Game` with tournament state (`isTournament`, settings, parent link), protocol fields in getInfo/game state events, a full-lobby bypass for tournament hubs, and `disconnectRemovesPlayer` plumbing
- Defer match-game creation to the owning thread's event loop so room registration never nests lock orders; guard shared state with the tournament mutex

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_participant.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_participant.cpp
@@ -580,7 +580,7 @@ void Server_AbstractParticipant::setUserInterface(Server_AbstractUserInterface *
 void Server_AbstractParticipant::disconnectClient()
 {
     bool isRegistered = userInfo->user_level() & ServerInfo_User::IsRegistered;
-    if (!isRegistered || spectator) {
+    if (!isRegistered || spectator || game->getDisconnectRemovesPlayer()) {
         game->removeParticipant(this, Event_Leave::USER_DISCONNECTED);
     } else {
         setUserInterface(nullptr);
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
@@ -30,10 +30,14 @@
 #include "server_cardzone.h"
 #include "server_player.h"
 #include "server_spectator.h"
+#include "server_tournament.h"
+#include "server_tournament_lifecycle_strategy.h"
+#include "server_tournament_match_result_strategy.h"
 
 #include <QDebug>
 #include <QRegularExpression>
 #include <QTimer>
+#include <algorithm>
 #include <google/protobuf/descriptor.h>
 #include <libcockatrice/deck_list/deck_list.h>
 #include <libcockatrice/protocol/pb/context_connection_state_changed.pb.h>
@@ -62,8 +66,9 @@ Server_Game::Server_Game(const GameConfig &config, Server_Room *_room)
       spectatorsCanTalk(config.spectatorsCanTalk), spectatorsSeeEverything(config.spectatorsSeeEverything),
       startingLifeTotal(config.startingLifeTotal), shareDecklistsOnLoad(config.shareDecklistsOnLoad),
       inactivityCounter(0), startTimeOfThisGame(0), secondsElapsed(0), firstGameStarted(false),
-      turnOrderReversed(false), startTime(QDateTime::currentDateTime()), pingClock(nullptr),
-      deckValidationStrategy(new Server_DefaultDeckValidationStrategy),
+      turnOrderReversed(false), startTime(QDateTime::currentDateTime()), pingClock(nullptr), isTournament(false),
+      tournament(nullptr), tournamentParentGame(nullptr), tournamentMatchPlayer1Id(-1), tournamentMatchPlayer2Id(-1),
+      disconnectRemovesPlayer(false), deckValidationStrategy(new Server_DefaultDeckValidationStrategy),
       lifecycleStrategy(new Server_DefaultLifecycleStrategy), matchResultStrategy(new Server_NullMatchResultStrategy),
       gameMutex()
 {
@@ -265,6 +270,12 @@ void Server_Game::createGameStateChangedEvent(Event_GameStateChanged *event,
         event->set_game_started(false);
     }
 
+    event->set_is_tournament(isTournament);
+
+    if (tournamentParentGame) {
+        event->set_parent_game_id(tournamentParentGame->getGameId());
+    }
+
     for (Server_AbstractParticipant *participant : participants.values()) {
         participant->getInfo(event->add_player_list(), recipient, omniscient, withUserInfo);
     }
@@ -313,7 +324,7 @@ void Server_Game::doStartGameIfReady(bool forceStartGame)
     Server_DatabaseInterface *databaseInterface = room->getServer()->getDatabaseInterface();
     QMutexLocker locker(&gameMutex);
 
-    if (getPlayerCount() < maxPlayers && !forceStartGame) {
+    if (!isTournament && getPlayerCount() < maxPlayers && !forceStartGame) {
         return;
     }
 
@@ -861,6 +872,7 @@ void Server_Game::getInfo(ServerInfo_Game &result) const
         result.set_share_decklists_on_load(shareDecklistsOnLoad);
         result.set_spectators_count(getSpectatorCount());
         result.set_start_time(startTime.toSecsSinceEpoch());
+        result.set_is_tournament(isTournament);
     }
 }
 
@@ -910,4 +922,87 @@ void Server_Game::returnCardsFromPlayer(GameEventStorage &ges, Server_AbstractPl
 void Server_Game::setDeckValidationStrategy(Server_DeckValidationStrategy *strategy)
 {
     deckValidationStrategy.reset(strategy);
+}
+
+void Server_Game::setMatchResultStrategy(Server_MatchResultStrategy *strategy)
+{
+    matchResultStrategy.reset(strategy);
+}
+
+void Server_Game::setIsTournamentGame(bool _isTournament)
+{
+    isTournament = _isTournament;
+    if (isTournament) {
+        tournament = new Server_Tournament(this, this, this);
+        lifecycleStrategy.reset(new Server_TournamentLifecycleStrategy);
+        matchResultStrategy.reset(new Server_TournamentMatchResultStrategy);
+    } else if (tournament) {
+        delete tournament;
+        tournament = nullptr;
+        lifecycleStrategy.reset(new Server_DefaultLifecycleStrategy);
+        matchResultStrategy.reset(new Server_NullMatchResultStrategy);
+    }
+}
+
+void Server_Game::startTournament()
+{
+    if (!tournament) {
+        tournament = new Server_Tournament(this, this, this);
+    }
+
+    if (!tournament->isStarted()) {
+        // Add all current players to the tournament
+        auto players = getPlayers();
+        for (auto *player : players.values()) {
+            tournament->addPlayer(player->getPlayerId(), QString::fromStdString(player->getUserInfo()->name()));
+        }
+
+        tournament->startTournament();
+    }
+
+    GameEventStorage ges;
+    tournament->broadcastTournamentState(ges);
+    ges.sendToGame(this);
+}
+
+void Server_Game::setPlayerTournamentDeck(int playerId, DeckList *deck)
+{
+    if (tournament) {
+        tournament->setPlayerDeck(playerId, deck);
+    }
+}
+
+void Server_Game::setTournamentMatchInfo(Server_Game *parentGame, int p1Id, int p2Id)
+{
+    tournamentParentGame = parentGame;
+    tournamentMatchPlayer1Id = p1Id;
+    tournamentMatchPlayer2Id = p2Id;
+}
+
+Server_Game *Server_Game::createMatchGame(const GameConfig &config, int &outGameId)
+{
+    Server_DatabaseInterface *databaseInterface = room->getServer()->getDatabaseInterface();
+    outGameId = databaseInterface->getNextGameId();
+    if (outGameId == -1) {
+        return nullptr;
+    }
+
+    GameConfig matchConfig = config;
+    matchConfig.gameId = outGameId;
+    auto *game = new Server_Game(matchConfig, room);
+    // Sub-games carry the tournament flag (for protocol fields) but keep the default
+    // strategies; the parent tournament drives them through the match result strategy
+    // installed by Server_Tournament::createMatchGame.
+    game->isTournament = true;
+    return game;
+}
+
+Server_AbstractUserInterface *Server_Game::getUserInterface(const QString &playerName)
+{
+    return room->getUserInterfaceByName(playerName);
+}
+
+void Server_Game::addGameToRoom(Server_Game *game)
+{
+    room->addGame(game);
 }
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
@@ -270,8 +270,6 @@ void Server_Game::createGameStateChangedEvent(Event_GameStateChanged *event,
         event->set_game_started(false);
     }
 
-    event->set_is_tournament(isTournament);
-
     if (tournamentParentGame) {
         event->set_parent_game_id(tournamentParentGame->getGameId());
     }
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
@@ -328,6 +328,13 @@ void Server_Game::doStartGameIfReady(bool forceStartGame)
         return;
     }
 
+    // Tournament hubs must be host-started and can't lock in a partially filled
+    // bracket: a mere "everyone current is ready" must not start a 1-player or
+    // undersized tournament. startTournament() additionally enforces 2+ players.
+    if (isTournament && !forceStartGame) {
+        return;
+    }
+
     auto players = getPlayers();
     for (auto *player : players.values()) {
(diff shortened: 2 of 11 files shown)


libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament_match_result_strategy.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "server_tournament.h"
+
+#include <libcockatrice/protocol/pb/event_game_closed.pb.h>
+#include <libcockatrice/protocol/pb/event_tournament_state.pb.h>
+
+bool Server_TournamentMatchResultStrategy::onGameFinished(Server_Game *game,
+                                                          int playing,
+                                                          Server_AbstractPlayer *lastPlayer)
+{
+    auto *parentGame = game->getTournamentParentGame();
```
ZeldaZach: Use-after-free. `tournamentParentGame` is a raw pointer and nothing keeps the hub alive: in `Server_Game::removeParticipant`, when the host leaves and no players remain, the hub does `gameClosed = true; deleteLater();`. The match sub-games it spawned are separate room-owned games and keep running, so when one of them finishes here `parentGame` is dangling and both `getTournament()` and `sendToGame(parentGame)` touch freed memory.

Suggest making the link non-owning-but-checked, e.g. store a `QPointer<Server_Game>` (the hub is a `QObject`) and bail out when it's null, and/or have the hub tear down / detach its outstanding match games in its destructor.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
```diff
@@ -0,0 +1,556 @@
...
+        if (winnerId == pairingPtr->player1Id) {
+            pairingPtr->player1MatchWins += 1;
+        } else if (winnerId == pairingPtr->player2Id) {
+            pairingPtr->player2MatchWins += 1;
+        }
+        // Draw (winnerId == -1): no match wins incremented
+
+        // Check if match is decided
+        const int gamesNeeded = static_cast<int>(gamesPerMatch);
+        matchDecided = (pairingPtr->player1MatchWins >= gamesNeeded) || (pairingPtr->player2MatchWins >= gamesNeeded);
```
ZeldaZach, line 563: `gamesNeeded` should be the wins required, not the series length. With `gamesPerMatch = 3` a player needs 3 wins, but `createMatchGame` refuses to create more than 3 games in the series (line 268), so a 2-1 result never sets `winnerId`, `allPairingsDecided()` never returns true, and the round stalls forever. Same for draws in a Bo1: `winnerId == -1` increments neither counter, so a drawn Bo1 also deadlocks the round.

This wants `gamesNeeded = gamesPerMatch / 2 + 1` plus an explicit terminal case for "series exhausted without a winner" (record it as a draw / decide on game wins) so the round can always advance.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
```diff
@@ -0,0 +1,556 @@
...
+    for (auto *matchPlayer : matchPlayers) {
+        const QString name = QString::fromStdString(matchPlayer->getUserInfo()->name());
+        QString deckNative;
+        if (name == player1Name) {
+            deckNative = deck1Native;
+        } else if (name == player2Name) {
+            deckNative = deck2Native;
+        }
+
+        if (!deckNative.isEmpty()) {
```
ZeldaZach, line 455: `deckNative` is always empty today: `submittedDecks` is only written by `Server_Tournament::setPlayerDeck`, whose only caller would be `Server_Game::setPlayerTournamentDeck`, and that has no callers anywhere in the tree. So neither `setDeck` nor `setReadyStart(true)` ever runs, and the `startGameIfReady(true)` on line 359 hits the force-start path in `doStartGameIfReady`, which **kicks every non-ready player**. Kicking p1 drops the match to one player, `stopGameIfFinished` fires, and `Server_TournamentMatchResultStrategy` records p2 as the match winner without a card being played.

Even once the deck plumbing exists, please don't rely on force-start here — it means any player who hasn't loaded their deck yet is silently kicked and forfeits.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
```diff
@@ -0,0 +1,556 @@
...
+                pairing.player2Id = available[j];
+                currentPairings.append(pairing);
+                paired.insert(available[i]);
+                paired.insert(available[j]);
+                break;
+            }
+        }
+    }
+
+    // Bye for unpaired player if odd count
```
ZeldaZach: Three problems with the bye block:

1. The `break` means only **one** unpaired player gets a bye. The greedy loop above can leave more than one player unpaired once rematch avoidance kicks in (all remaining candidates have already played each other) — those players get no pairing at all, no game, and no record change, and just silently sit out the round.
2. There's no "at most one bye per player" rule. `havePlayed(p, -1)` is recorded on line 149 but never consulted, so the same player can draw a bye every single round.
3. `available` is sorted by wins descending, so the leftover is picked from the top of the standings when the greedy pass fails mid-list. Swiss byes should go to the lowest-ranked eligible player who hasn't had one.

Also worth noting the greedy pass can paint itself into a corner (pairing i-j early forces a later rematch or an unpairable remainder) — a backtracking or max-weight matching pass would avoid that, but at minimum every player needs to end up in exactly one pairing or one bye each round.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
```diff
@@ -0,0 +1,556 @@
...
+{
+    QMutexLocker locker(&tournamentMutex);
+    delete submittedDecks.value(playerId, nullptr);
+    submittedDecks[playerId] = deck;
+    if (players.contains(playerId)) {
+        players[playerId].deckSubmitted = true;
+    }
+}
+
+void Server_Tournament::removePlayer(int playerId)
```
ZeldaZach, line 52: `removePlayer` has no callers — nothing hooks `Server_Game::removeParticipant` on the hub game. A player who leaves or disconnects stays in `players`, keeps appearing in standings, and keeps getting paired. Their `createMatchGame` then finds `getUserInterface(name) == nullptr` (line 323), so the sub-game is created and registered in the room with **zero participants**, never finishes, never reports a result, and the round never advances — plus the empty game leaks into the room's game list permanently.

Drops need first-class handling: mark the player dropped, stop pairing them, award their opponent a win, and refuse to create a match game if either UI lookup fails (destroy the freshly created game rather than registering it).

Minor, same file: `addPlayer` is the only mutator that doesn't take `tournamentMutex`. It happens to be called under the hub's `gameMutex` today, but the inconsistency is worth fixing.

libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
```diff
@@ -313,7 +324,7 @@ void Server_Game::doStartGameIfReady(bool forceStartGame)
     Server_DatabaseInterface *databaseInterface = room->getServer()->getDatabaseInterface();
     QMutexLocker locker(&gameMutex);
 
-    if (getPlayerCount() < maxPlayers && !forceStartGame) {
+    if (!isTournament && getPlayerCount() < maxPlayers && !forceStartGame) {
```
ZeldaZach, line 325: This bypass drops the only gate that made "start" host-controlled for a partially filled lobby. `cmdReadyStart` calls `startGameIfReady(false)` for any player, and the host check only guards the `force_start` path — so for a tournament hub the tournament now locks in the moment every *currently joined* player is ready. A host who creates an 8-player tournament and readies up alone gets a 1-player tournament (`calculateTotalRounds()` returns 0, so it immediately reports `PHASE_FINISHED`); two early joiners can lock everyone else out.

Tournament start should require the host (`force_start`), and `startTournament()` should refuse to start with fewer than 2 players.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
```diff
@@ -0,0 +1,556 @@
...
+                qCWarning(TournamentLog) << "Refusing to exceed series length for pairing" << player1Id << player2Id;
+                return;
+            }
+        }
+    }
+
+    // Create a sub-game for this match via the factory
+    ServerInfo_User creatorInfo;
+    creatorInfo.set_name(player1Name.toStdString());
+    creatorInfo.set_user_level(ServerInfo_User::IsAdmin | ServerInfo_User::IsRegistered);
```
ZeldaZach: Please don't fabricate `IsAdmin` here. This `ServerInfo_User` becomes the sub-game's `creatorInfo`, which is shipped to clients via `Server_Game::getInfo` and used by `checkJoin` for the buddy/ignore-list checks — so an ordinary player shows up to everyone as the admin-level creator of the match game. Copy player1's real `ServerInfo_User` (or use a neutral system identity) instead.

libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_participant.cpp
```diff
@@ -580,7 +580,7 @@ void Server_AbstractParticipant::setUserInterface(Server_AbstractUserInterface *
 void Server_AbstractParticipant::disconnectClient()
 {
     bool isRegistered = userInfo->user_level() & ServerInfo_User::IsRegistered;
-    if (!isRegistered || spectator) {
+    if (!isRegistered || spectator || game->getDisconnectRemovesPlayer()) {
```
ZeldaZach, line 583: `disconnectRemovesPlayer` is initialized to `false` and never set anywhere, so this branch is currently dead. Either wire it up (tournament match games presumably want it) or drop it from this PR.

ZeldaZach (changes requested):
Thanks for tackling this — the overall shape (strategy objects + a separate `Server_Tournament` owned by the hub game) fits the existing design well, and deferring match-game creation onto the event loop so `gamesLock` is never nested under a `gameMutex` is exactly the right call. The locking inside `Server_Tournament` itself also looks sane (`QRecursiveMutex` makes the `broadcastTournamentState`/`advanceRound` re-entry safe).

That said, I can't merge this yet — there are a few correctness problems that I think make the engine non-functional or unsafe in practice:

1. **Use-after-free**: sub-games keep a raw `Server_Game *tournamentParentGame`. The hub game `deleteLater()`s itself as soon as the last player leaves (`removeParticipant`), but the spawned match games outlive it and dereference the pointer when they finish.
2. **Best-of-N can never resolve**: `gamesNeeded` is `gamesPerMatch`, but the series is also capped at `gamesPerMatch` games. Anything other than a clean sweep permanently stalls the round.
3. **Nothing ever populates `submittedDecks`** — `Server_Game::setPlayerTournamentDeck` / `Server_Tournament::setPlayerDeck` have no callers anywhere in this PR. The result is that match players never get a deck or `readyStart`, so `startGameIfReady(true)` kicks both of them and the match is "decided" by the kick rather than by play. If the deck plumbing lands in #7200, this PR is at least not independently testable; if it's meant to be here, it's missing.
4. **Players who leave are never removed from the tournament** — `Server_Tournament::removePlayer` has no callers, and `disconnectRemovesPlayer` is never set to `true`, so the `server_abstract_participant.cpp` change is currently inert.
5. **Bye handling** drops players and allows repeat byes (details inline).
6. **Any player can start the tournament** now that the full-lobby gate is bypassed, including with a single participant.

Separately: standings are sorted by wins/losses only — no OMW%/GW% tiebreakers — so pairings and final placement won't match real Swiss. Fine to defer, but worth calling out since the PR description mentions Swiss semantics.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament_match_result_strategy.cpp
BruebachL: Fixed in 5aaf2729fc / d801ced94e: the match sub-game now holds its parent through `QPointer<Server_Game>` (`setTournamentMatchInfo` stores a `QPointer` and `getTournamentParentGame()` returns it), so `onGameFinished` dereferences through `.data()` and bails when the hub has been torn down. A stale match can no longer touch freed memory.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
BruebachL: Fixed in d801ced94e: `gamesNeeded` is now `gamesPerMatch/2 + 1` (the wins required, so a Bo3 needs 2). An exhausted series without a strict-majority winner is recorded as a draw (`winnerId = -1` set on the pairing) instead of deadlocking, so `allPairingsDecided()` always resolves and the round always advances — that covers a drawn Bo1 and a 1-1 Bo3.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
BruebachL: Fixed in d801ced94e: match games are no longer started with `startGameIfReady(true)`. When every participant has a submitted deck the game starts normally (`force = false`); when any deck is missing the game is left open for deck selection instead of force-starting and kicking the unprepared player. The provider-side deck plumbing itself remains a follow-up as you noted.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
BruebachL: Fixed in d801ced94e: every leftover player now gets a bye, not just one. Byes are handed out worst-ranked-first, and a player receives at most one bye over the whole tournament via a `byeGivenPlayers` set; a leftover who has already had a bye sits out with a free win so the round always resolves. The greedy pass also now allows rematches only as a last resort to avoid leaving players unpairable.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
BruebachL: Fixed in d801ced94e: leaving the hub now drops the player. `Server_Game::removeParticipant` calls `tournament->dropPlayer(id)` on the hub (skipped for match sub-games), which stops them being paired, marks them dropped in the bracket, and awards any current undecided match to their opponent. `createMatchGame` now refuses to spawn a match when either participant's UI lookup fails, so no zero- or one-player game is ever registered and left to leak in the room. `addPlayer` also now takes `tournamentMutex` for consistency.

libcockatrice_network/libcockatrice/network/server/remote/game/server_game.cpp
BruebachL: Fixed in d801ced94e: tournament hubs cannot start on 'everyone currently ready' — `doStartGameIfReady` returns unless `forceStartGame` (host-only) for tournaments, so an 8-player host readied alone can no longer lock a 1-player bracket. `startTournament()` additionally refuses with fewer than 2 players.

libcockatrice_network/libcockatrice/network/server/remote/game/server_tournament.cpp
BruebachL: Fixed in d801ced94e: the sub-game's `creatorInfo` now copies player1's real `ServerInfo_User` (`*ui->getUserInfo()`), falling back to a neutral name-only identity only if that lookup fails. No synthetic `IsAdmin` is broadcast to clients.

libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_participant.cpp
BruebachL: Fixed in d801ced94e: match sub-games now set `disconnectRemovesPlayer(true)` via a new setter, so a mid-match disconnect removes the player and lets the match be decided rather than leaving a half-present participant. The previously dead branch is now exercised by tournament matches.

libcockatrice_network/libcockatrice/network/server/remote/game/server_game.h
```diff
@@ -24,31 +24,38 @@
 #include "game_config.h"
 #include "server_deck_validation_strategy.h"
 #include "server_game_lifecycle_strategy.h"
+#include "server_match_game_factory.h"
```
RickyRister, line 27: compile fails for me with
```
In file included from /Users/Ricky/Documents/GitHub/Cockatrice/cmake-build-debug/libcockatrice_network/libcockatrice/network/server/remote/game/../../../../../../../libcockatrice_network/libcockatrice/network/server/remote/game/server_game.h:27:
/Users/Ricky/Documents/GitHub/Cockatrice/cmake-build-debug/libcockatrice_network/libcockatrice/network/server/remote/game/../../../../../../../libcockatrice_network/libcockatrice/network/server/remote/game/server_match_game_factory.h:6:1: error: class 'GameConfig' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
    6 | class GameConfig;
      | ^
/Users/Ricky/Documents/GitHub/Cockatrice/cmake-build-debug/libcockatrice_network/libcockatrice/network/server/remote/game/../../../../../../../libcockatrice_network/libcockatrice/network/server/remote/game/game_config.h:8:8: note: previous use is here
    8 | struct GameConfig
      |        ^
/Users/Ricky/Documents/GitHub/Cockatrice/cmake-build-debug/libcockatrice_network/libcockatrice/network/server/remote/game/../../../../../../../libcockatrice_network/libcockatrice/network/server/remote/game/server_match_game_factory.h:6:1: note: did you mean struct here?
    6 | class GameConfig;
      | ^~~~~
      | struct
1 error generated.
```

PR #7200, opened 2026-08-29
[Client] Add tournament event plumbing and player extension points

BruebachL:
Wires the client game layer for tournament state without touching existing behavior:
- GameEventHandler dispatches Event_TournamentState (2027) to a new tournamentStateChanged signal
- GameMetaInfo exposes isTournament and parentGameId over the new ServerInfo_Game field
- PlayerEventHandler::processGameEvent is virtual and player is protected.
-  PlayerLogic gains a protected constructor accepting a custom handler, so mode-specific subclasses can intercept events

### DIFF
--- a/cockatrice/src/game/game_event_handler.cpp
+++ b/cockatrice/src/game/game_event_handler.cpp
@@ -26,6 +26,7 @@
 #include <libcockatrice/protocol/pb/event_reverse_turn.pb.h>
 #include <libcockatrice/protocol/pb/event_set_active_phase.pb.h>
 #include <libcockatrice/protocol/pb/event_set_active_player.pb.h>
+#include <libcockatrice/protocol/pb/event_tournament_state.pb.h>
 #include <libcockatrice/protocol/pb/game_event_container.pb.h>
 #include <libcockatrice/protocol/pending_command.h>
 
@@ -158,6 +159,9 @@ void GameEventHandler::processGameEventContainer(const GameEventContainer &cont,
                 case GameEvent::REVERSE_TURN:
                     eventReverseTurn(event.GetExtension(Event_ReverseTurn::ext), playerId, context);
                     break;
+                case GameEvent::TOURNAMENT_STATE:
+                    emit tournamentStateChanged(event.GetExtension(Event_TournamentState::ext));
+                    break;
 
                 default: {
                     PlayerLogic *player = game->getPlayerManager()->getPlayers().value(playerId, 0);
@@ -263,6 +267,12 @@ void GameEventHandler::eventGameStateChanged(const Event_GameStateChanged &event
                                              int /*eventPlayerId*/,
                                              const GameEventContext & /*context*/)
 {
+    // Sub-games of a tournament report their parent hub game so the client can
+    // route "close game" back to the parent tab.
+    if (event.parent_game_id() != -1) {
+        game->getGameMetaInfo()->setParentGameId(event.parent_game_id());
+    }
+
     const int playerListSize = event.player_list_size();
 
     QVector<QPair<int, QPair<QString, QString>>> opponentDecksToDisplay;
--- a/cockatrice/src/game/game_event_handler.cpp
+++ b/cockatrice/src/game/game_event_handler.cpp
@@ -268,8 +268,9 @@ void GameEventHandler::eventGameStateChanged(const Event_GameStateChanged &event
                                              const GameEventContext & /*context*/)
 {
     // Sub-games of a tournament report their parent hub game so the client can
-    // route "close game" back to the parent tab.
-    if (event.parent_game_id() != -1) {
+    // route "close game" back to the parent tab. Use has_parent_game_id rather
+    // than comparing the default so a future drop of the proto default still works.
+    if (event.has_parent_game_id()) {
         game->getGameMetaInfo()->setParentGameId(event.parent_game_id());
     }
 
--- a/cockatrice/src/game/game_meta_info.h
+++ b/cockatrice/src/game/game_meta_info.h
@@ -84,6 +84,26 @@ public:
         return roomGameTypes.find(gameInfo_.game_types(index)).value();
     }
 
+    bool isTournament() const
+    {
+        return gameInfo_.is_tournament();
+    }
+
+    void setIsTournament(bool t)
+    {
+        gameInfo_.set_is_tournament(t);
+    }
+
+    int parentGameId() const
+    {
+        return parentGameId_;
+    }
+
+    void setParentGameId(int id)
+    {
+        parentGameId_ = id;
+    }
+
 public slots:
     void setStarted(bool s)
     {
@@ -108,6 +128,7 @@ signals:
 
 private:
     ServerInfo_Game gameInfo_;
+    int parentGameId_ = -1;
 };
 
 #endif // GAME_META_INFO_H
--- a/cockatrice/src/game/game_meta_info.h
+++ b/cockatrice/src/game/game_meta_info.h
@@ -89,11 +89,6 @@ public:
         return gameInfo_.is_tournament();
     }
 
-    void setIsTournament(bool t)
-    {
-        gameInfo_.set_is_tournament(t);
-    }
-
     int parentGameId() const
     {
         return parentGameId_;
--- a/cockatrice/src/game/player/player_event_handler.h
+++ b/cockatrice/src/game/player/player_event_handler.h
@@ -90,10 +90,10 @@ public:
      * @param context Additional context (undo, judge, etc.).
      * @param options Processing options (UI suppression, reveal behavior).
      */
-    void processGameEvent(GameEvent::GameEventType type,
-                          const GameEvent &event,
-                          const GameEventContext &context,
-                          EventProcessingOptions options);
+    virtual void processGameEvent(GameEvent::GameEventType type,
+                                  const GameEvent &event,
+                                  const GameEventContext &context,
+                                  EventProcessingOptions options);
 
     /** @} */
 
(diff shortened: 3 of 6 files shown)


cockatrice/src/game/player/player_logic.h
```diff
@@ -99,7 +99,30 @@ public slots:
...
+protected:
+    /**
+     * @brief Constructor for subclasses that need a custom event handler (e.g. DraftPlayerLogic).
+     *
+     * @p customEventHandler must be non-null and either QObject-parented to this PlayerLogic
+     * or deleted externally; it is not owned by PlayerLogic. The handler connects to @c player
+     * during its own constructor, so passing a freshly built subclass handler from an
+     * initializer list is safe.
+     */
+    PlayerLogic(const ServerInfo_User &info,
```
ZeldaZach: The documented usage of this constructor is undefined behavior. The only way a subclass can "pass a freshly built subclass handler from an initializer list" is:

```cpp
DraftPlayerLogic::DraftPlayerLogic(...)
    : PlayerLogic(info, id, local, judge, parent, new DraftPlayerEventHandler(this)) {}
```

At that point the PlayerLogic base subobject has not begun construction, so converting `this` to `PlayerLogic*` is already UB ([class.cdtor]/3), and `PlayerEventHandler(PlayerLogic *_player)` then does `QObject(_player)` (setParent) plus `connect(this, ..., player, &PlayerLogic::requestCardMenuUpdate)` against an object whose QObject base and vtable do not exist yet. That is a crash, not a supported pattern.

The doc is also self-contradictory: it says the handler "is not owned by PlayerLogic", but PlayerEventHandler's constructor parents itself to the player, so PlayerLogic does own it.

If you need this hook, use a post-construction `setEventHandler()` (deleting the default one) or a virtual factory called from an explicit `init()` step, not a constructor parameter.

cockatrice/src/game/player/player_logic.cpp
```diff
@@ -37,6 +37,19 @@ PlayerLogic::PlayerLogic(const ServerInfo_User &info, int _id, bool _local, bool
...
 }
 
+PlayerLogic::PlayerLogic(const ServerInfo_User &info,
+                         int _id,
+                         bool _local,
+                         bool _judge,
+                         AbstractGame *_parent,
+                         PlayerEventHandler *customEventHandler)
+    : QObject(_parent), game(_parent), playerInfo(new PlayerInfo(info, _id, _local, _judge)),
+      playerEventHandler(customEventHandler), playerActions(new PlayerActions(this)), active(false), conceded(false),
```
ZeldaZach: This duplicates the primary constructor's member-init list verbatim; the two will drift the first time a member is added. Delegate instead (`PlayerLogic(info, _id, _local, _judge, _parent)` then swap the handler) or make it a defaulted parameter on the existing constructor.

Also, `customEventHandler` is documented as "must be non-null" but nothing enforces it, and `GameEventHandler::processGameEventContainer` calls `player->getPlayerEventHandler()->processGameEvent(...)` with no null check, so a null handler is a straight segfault. Add an assert or fall back to `new PlayerEventHandler(this)`.

cockatrice/src/game/player/player_logic.h
```diff
@@ -99,7 +99,30 @@ public slots:
...
+                PlayerEventHandler *customEventHandler);
+
+public:
+    /**
+     * @brief Creates the standard zone set.
+     *
+     * Not virtually dispatched from constructors — subclasses overriding this must add
+     * their extra zones in their own constructor body.
+     */
+    virtual void initializeZones();
```
ZeldaZach: Making this virtual accomplishes nothing as written. The only caller is the constructor body, and as the doc itself notes, virtual dispatch does not happen there - so an override can never be reached through this call site. Nothing in the tree overrides it either. Please drop the `virtual` until there is a call path that actually dispatches on it.

cockatrice/src/game/player/player_event_handler.h
```diff
@@ -90,10 +90,10 @@ class PlayerEventHandler : public QObject
      * @param context Additional context (undo, judge, etc.).
      * @param options Processing options (UI suppression, reveal behavior).
      */
-    void processGameEvent(GameEvent::GameEventType type,
-                          const GameEvent &event,
-                          const GameEventContext &context,
-                          EventProcessingOptions options);
+    virtual void processGameEvent(GameEvent::GameEventType type,
```
ZeldaZach: `processGameEvent` is made virtual and `player` is moved to protected, but nothing derives from PlayerEventHandler in this PR or in #7201, and the referenced `DraftPlayerLogic` does not exist in the tree. Please land these together with the subclass that needs them - otherwise it is API surface with no consumer, and the intended contract for an override (call the base? swallow the event? does the caller still need `emitUserEvent()`?) is undefined.

cockatrice/src/game/game_meta_info.h
```diff
@@ -84,6 +84,26 @@ class GameMetaInfo : public QObject
         return roomGameTypes.find(gameInfo_.game_types(index)).value();
     }
 
+    bool isTournament() const
+    {
+        return gameInfo_.is_tournament();
+    }
+
+    void setIsTournament(bool t)
```
ZeldaZach: `setIsTournament()` has no caller anywhere in the stack. Meanwhile #7197 adds `Event_GameStateChanged.is_tournament` and #7199 sets it on every game-state event, and `eventGameStateChanged` below ignores it - it only reads `parent_game_id`. `isTournament()` works today only because `Game::Game` populates gameMetaInfo from `Event_GameJoined.game_info()`, where Server_Game::getInfo sets the field.

Pick one: either call `setIsTournament(event.is_tournament())` in `eventGameStateChanged` (and keep the setter), or drop both the setter and the `is_tournament` field from Event_GameStateChanged in #7197.

cockatrice/src/game/game_event_handler.cpp
```diff
@@ -263,6 +267,12 @@ void GameEventHandler::eventGameStateChanged(const Event_GameStateChanged &event
                                              int /*eventPlayerId*/,
                                              const GameEventContext & /*context*/)
 {
+    // Sub-games of a tournament report their parent hub game so the client can
+    // route "close game" back to the parent tab.
+    if (event.parent_game_id() != -1) {
```
ZeldaZach: Prefer `event.has_parent_game_id()` over comparing against the magic default. It works today only because #7197 declares `[default = -1]`; if that default is ever dropped the accessor returns 0 and every non-tournament game gets `setParentGameId(0)`, which #7201 treats as a real parent tab id.

ZeldaZach (changes requested):
Reviewed as part of the tournament stack (#7197 protocol, #7199 server, this, #7201 UI). The event dispatch half is fine: TOURNAMENT_STATE lands in the non-spectator switch, the server sends it with the default player_id of -1, so it reaches both players and spectators and is not swallowed by the per-client filter. The signal is correctly declared under signals: and #7201 connects to it with a PMF connect, so no silent connect failure.

What I want changed is the "player extension points" half. The new protected PlayerLogic constructor documents a usage pattern that is undefined behavior, and none of the four extension points (custom-handler ctor, virtual processGameEvent, virtual initializeZones, setIsTournament) has a single caller in this PR, in #7201, or anywhere in the tree. Details inline.

Also: Event_GameStateChanged.is_tournament is added in #7197 and set by the server in #7199, but the client never reads it here - either consume it or drop the field from the protocol PR.

Note this cannot compile on master until #7197 lands (GameEvent::TOURNAMENT_STATE, ServerInfo_Game.is_tournament, Event_GameStateChanged.parent_game_id do not exist yet). The macOS 26 Debug failure looks like runner/infra flake - there are no compiler errors in the job log.

cockatrice/src/game/player/player_logic.h
BruebachL: Agreed — the second constructor pattern is UB per [class.cdtor]/3 (converting `this` to `PlayerLogic*` in the base-initializer) and the doc was self-contradictory about ownership. Rather than patch it, I dropped the entire extension surface: the second constructor, the virtual hooks, and the handler plumbing are gone, and player_logic.h/.cpp + player_event_handler.h are back to the pre-PR code. The tournament dispatch path that has a real consumer (#7201 uses `GameMetaInfo::isTournament()` and routing via `parentGameId()`) is untouched.

cockatrice/src/game/player/player_logic.cpp
BruebachL: Moot — this constructor no longer exists. The whole custom-handler extension was removed and player_logic.cpp is back to the single primary constructor, so there is no duplicate init list to drift and no null-handler path to guard.

cockatrice/src/game/player/player_logic.h
BruebachL: Dropped. `initializeZones()` is back to a plain (non-virtual) member; the virtual dispatch concern disappears with the extension surface that motivated it.

cockatrice/src/game/player/player_event_handler.h
BruebachL: Removed — `processGameEvent` is non-virtual again and `player` is back to private. No subclass needs them anywhere in the stack (the referenced DraftPlayerLogic does not exist), so the API surface is gone entirely.

cockatrice/src/game/game_meta_info.h
BruebachL: Went with your second option — dropped both. `setIsTournament()` is removed from GameMetaInfo here, the `is_tournament` field is removed from `Event_GameStateChanged` in #7197 (commit b8083b6c93), and the server-side setter is removed in #7199 (commit 4584891cf1). `isTournament()` keeps reading `ServerInfo_Game.is_tournament` from `Event_GameJoined`, which is authoritative and used by #7201. No client or server code references the dropped field anymore.

cockatrice/src/game/game_event_handler.cpp
BruebachL: Fixed — switched to `if (event.has_parent_game_id()) { ... }` exactly as suggested, so non-tournament games no longer risk getting a bogus parent id if the proto default is ever dropped.

PR #7201, opened 2026-08-29
[Client] Add tournament UI for creating and following lobby tournaments

BruebachL:
## Related Tickets
- #7221 

The client can't create or follow lobby tournaments — no tournament checkbox, no live standings, and no way to reach the current match game.

- Add a tournament checkbox and a games-per-match settings sub-dialog to `DlgCreateGame`; both values ride `Command_CreateGame` and are mirrored in the join-info variant
- Add `TournamentWidget` rendering live standings (deck-submission state, pairings) and requesting the current match game
- Add `TournamentTabGameExtension`: an overview page in `TabGame`, phase and submission progress on a deck-view status strip, and an in-game settings dialog for hosts during deck building
- Route open-match to spectator join through `TabSupervisor` and return sub-games to their parent tab on close, falling back to the normal leave-game flow

### DIFF
--- a/cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
@@ -2,6 +2,7 @@
 
 #include "../../../client/settings/cache_settings.h"
 #include "../interface/widgets/tabs/tab_room.h"
+#include "dlg_tournament_settings.h"
 
 #include <QApplication>
 #include <QCheckBox>
@@ -104,14 +105,29 @@ void DlgCreateGame::sharedCtor()
 
     shareDecklistsOnLoadCheckBox = new QCheckBox(tr("Open decklists in lobby"));
 
+    tournamentCheckBox = new QCheckBox(tr("Tournament mode"));
+    tournamentCheckBox->setToolTip(tr("All players submit a deck up front and rounds are paired automatically"));
+    tournamentSettingsButton = new QPushButton(tr("Settings..."));
+    tournamentSettingsButton->setEnabled(false);
+    connect(tournamentCheckBox, &QCheckBox::toggled, tournamentSettingsButton, &QPushButton::setEnabled);
+    connect(tournamentSettingsButton, &QPushButton::clicked, this, [this] {
+        DlgTournamentSettings dlg(this);
+        dlg.setCurrentGamesPerMatch(tournamentSettings.gamesPerMatch);
+        if (dlg.exec() == QDialog::Accepted) {
+            tournamentSettings = dlg.getResult();
+        }
+    });
+
     createGameAsJudgeCheckBox = new QCheckBox(tr("Create game as judge"));
 
     auto *gameSetupOptionsLayout = new QGridLayout;
     gameSetupOptionsLayout->addWidget(startingLifeTotalLabel, 0, 0);
     gameSetupOptionsLayout->addWidget(startingLifeTotalEdit, 0, 1);
     gameSetupOptionsLayout->addWidget(shareDecklistsOnLoadCheckBox, 1, 0);
+    gameSetupOptionsLayout->addWidget(tournamentCheckBox, 2, 0);
+    gameSetupOptionsLayout->addWidget(tournamentSettingsButton, 2, 1);
     if (room && room->getUserInfo()->user_level() & ServerInfo_User::IsJudge) {
-        gameSetupOptionsLayout->addWidget(createGameAsJudgeCheckBox, 2, 0);
+        gameSetupOptionsLayout->addWidget(createGameAsJudgeCheckBox, 3, 0);
     } else {
         createGameAsJudgeCheckBox->setChecked(false);
         createGameAsJudgeCheckBox->setHidden(true);
@@ -188,7 +204,7 @@ DlgCreateGame::DlgCreateGame(TabRoom *_room, const QMap<int, QString> &_gameType
 }
 
 DlgCreateGame::DlgCreateGame(const ServerInfo_Game &gameInfo, const QMap<int, QString> &_gameTypes, QWidget *parent)
-    : QDialog(parent), room(0), gameTypes(_gameTypes)
+    : QDialog(parent), room(nullptr), gameTypes(_gameTypes)
 {
     sharedCtor();
 
@@ -205,6 +221,8 @@ DlgCreateGame::DlgCreateGame(const ServerInfo_Game &gameInfo, const QMap<int, QS
     createGameAsSpectatorCheckBox->setEnabled(false);
     startingLifeTotalEdit->setEnabled(false);
     shareDecklistsOnLoadCheckBox->setEnabled(false);
+    tournamentCheckBox->setEnabled(false);
+    tournamentSettingsButton->setEnabled(false);
 
     descriptionEdit->setText(QString::fromStdString(gameInfo.description()));
     maxPlayersEdit->setValue(gameInfo.max_players());
@@ -215,6 +233,7 @@ DlgCreateGame::DlgCreateGame(const ServerInfo_Game &gameInfo, const QMap<int, QS
     spectatorsCanTalkCheckBox->setChecked(gameInfo.spectators_can_chat());
     spectatorsSeeEverythingCheckBox->setChecked(gameInfo.spectators_omniscient());
     shareDecklistsOnLoadCheckBox->setChecked(gameInfo.share_decklists_on_load());
+    tournamentCheckBox->setChecked(gameInfo.is_tournament());
 
     QSet<int> types;
     for (int i = 0; i < gameInfo.game_types_size(); ++i) {
@@ -252,6 +271,8 @@ void DlgCreateGame::actReset()
 
     startingLifeTotalEdit->setValue(20);
     shareDecklistsOnLoadCheckBox->setChecked(false);
+    tournamentCheckBox->setChecked(false);
+    tournamentSettings = DlgTournamentSettingsResult{};
     createGameAsJudgeCheckBox->setChecked(false);
 
     QMapIterator<int, QRadioButton *> gameTypeCheckBoxIterator(gameTypeCheckBoxes);
@@ -282,6 +303,10 @@ void DlgCreateGame::actOK()
     cmd.set_join_as_spectator(createGameAsSpectatorCheckBox->isChecked());
     cmd.set_starting_life_total(startingLifeTotalEdit->value());
     cmd.set_share_decklists_on_load(shareDecklistsOnLoadCheckBox->isChecked());
+    cmd.set_is_tournament(tournamentCheckBox->isChecked());
+    if (tournamentCheckBox->isChecked()) {
+        cmd.set_games_per_match(tournamentSettings.gamesPerMatch);
+    }
 
     auto _gameTypes = QString();
     QMapIterator<int, QRadioButton *> gameTypeCheckBoxIterator(gameTypeCheckBoxes);
--- a/cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
@@ -15,6 +15,7 @@
 #include <QPushButton>
 #include <QRadioButton>
 #include <QSet>
+#include <QSignalBlocker>
 #include <QSpinBox>
 #include <libcockatrice/protocol/pb/serverinfo_game.pb.h>
 #include <libcockatrice/protocol/pending_command.h>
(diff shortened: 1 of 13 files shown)


cockatrice/src/interface/widgets/tabs/tournament_tab_game_extension.cpp
```diff
@@ -0,0 +1,206 @@
...
+    DlgTournamentSettingsResult result = dlg.getResult();
+
+    PlayerLogic *localPlayer = tabGame->getGame()->getPlayerManager()->getActiveLocalPlayer(-1);
+    if (!localPlayer) {
+        TabSupervisor::actShowPopup(tr("You are not an active player in this game."));
+        return;
+    }
+
+    Command_TournamentSettingsSelect cmd;
+    cmd.mutable_settings()->set_games_per_match(result.gamesPerMatch);
```
ZeldaZach: This command has no handler. `Server_AbstractParticipant::cmdTournamentSettingsSelect` returns `RespContextError` and nothing in #7197/#7199/#7200/#7201 overrides it, so the host changes games-per-match, the server rejects it, and the UI says nothing. Either add the `Server_Player` override in #7199 or drop the in-game settings button until it exists.

Two smaller things in the same method: the dialog isn't seeded with the tournament's current value (unlike the `DlgCreateGame` path, which calls `setCurrentGamesPerMatch`), so it always opens on 1 and would silently reset the setting; and the command's response isn't checked, so a rejection is invisible to the host.

cockatrice/src/interface/widgets/tabs/tournament_tab_game_extension.cpp
```diff
@@ -0,0 +1,206 @@
...
+void TournamentTabGameExtension::initializeTournamentMode()
+{
+    if (!tabGame) {
+        return;
+    }
+
+    auto *deckLayout = tabGame->getDeckViewContainerLayout();
+    int index = 0;
+    deckLayout->insertWidget(index++, deckViewStatusLabel);
+    if (isLocalPlayerHost()) {
```
ZeldaZach: Host-ness is evaluated once, during tab construction, and never re-evaluated. `GameEventHandler` calls `GameState::setHostId()` when the host changes mid-game, so a player promoted to host never gets the button, and the departing host keeps one that no longer applies. Move the decision into `updateNavigationButtons()` (create the button lazily or create it always and toggle visibility).

Also, `AbstractGame::isHost()` already does exactly what `isLocalPlayerHost()` does — please use it instead of re-deriving the comparison here.

cockatrice/src/interface/widgets/tabs/tournament_tab_game_extension.cpp
```diff
@@ -0,0 +1,206 @@
...
+        settingsButton = new QPushButton(tabGame->getDeckViewContainerWidget());
```
ZeldaZach: `settingsButton` never gets a label on first show. `TabGame`'s constructor calls `retranslateUi()` (which reaches `TournamentTabGameExtension::retranslateUi`) before it calls `initializeTournamentMode()`, so at that point `settingsButton` is still null and the `if (settingsButton)` guard skips it. The host sees a blank button until something else triggers a retranslate (language change or a concede). Set the text where the button is created, or call `retranslateUi()` at the end of `initializeTournamentMode()`.

cockatrice/src/interface/widgets/draft/tournament_widget.cpp
```diff
@@ -0,0 +1,227 @@
...
+    openMatchButton->setVisible(playingPhase);
+    if (!playingPhase) {
+        return;
+    }
+
+    if (hasOwnLivePairing) {
+        openMatchButton->setEnabled(true);
+        openMatchButton->setText(tr("Open match game"));
+        openMatchButton->setToolTip(tr("Switch to your current match game"));
+    } else if (localPlayerId == -1 && hasAnyLivePairing) {
```
ZeldaZach: `localPlayerId == -1` never holds for a spectator, so this branch is unreachable and spectators always fall into the `else` with a disabled button. Server-side, `Server_Game` allocates spectator ids from the same `nextPlayerId` counter as players and `createGameJoinedEvent` sets `player_id` to that real id for spectators too.

Pass `PlayerManager::isSpectator()` down alongside the id (or expose a `setIsSpectator`) and branch on that.

cockatrice/src/interface/widgets/draft/tournament_widget.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+void TournamentWidget::updateTournamentState(const Event_TournamentState &state)
+{
+    lastState = state;
+
+    int gamesPerMatch = getGamesPerMatch(state);
+    QString statusText;
+    switch (state.phase()) {
+        case Event_TournamentState::PHASE_DECK_BUILDING:
+            statusText = gamesPerMatch > 1 ? tr("Tournament - Deck building (Best of %1)").arg(gamesPerMatch * 2 - 1)
```
ZeldaZach: `gamesPerMatch * 2 - 1` treats the value as *wins required*, but `DlgTournamentSettings` tells the user it's the *total games* ("Set to 3 for Best of 3"). With the dialog set to 3, this label reads "Best of 5".

The server is ambiguous on this too — `Server_Tournament` decides a match with `matchWins >= gamesPerMatch` (wins required) while the sub-game description formats "Game %2 of %3" against the same value (total games). Please settle on one meaning across #7199 and this PR and make the field name match it.

cockatrice/src/interface/widgets/dialogs/dlg_tournament_settings.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    auto *mainLayout = new QFormLayout(this);
+
+    gamesPerMatchSpin = new QSpinBox(this);
+    gamesPerMatchSpin->setRange(1, 5);
+    gamesPerMatchSpin->setValue(1);
+    gamesPerMatchSpin->setToolTip(tr("Number of games per match (e.g., 3 for Best of 3)"));
+    mainLayout->addRow(tr("Games per match:"), gamesPerMatchSpin);
+
+    QLabel *hintLabel = new QLabel(tr("Set to 3 for Best of 3, 5 for Best of 5, etc."), this);
```
ZeldaZach, line 21: See the note on `tournament_widget.cpp` — this hint and the tooltip above define games-per-match as total games in the series, while the standings header derives "Best of" from it as wins-required. One of the two has to change.

cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
```diff
@@ -215,6 +233,7 @@ DlgCreateGame::DlgCreateGame(const ServerInfo_Game &gameInfo, const QMap<int, QS
     spectatorsCanTalkCheckBox->setChecked(gameInfo.spectators_can_chat());
     spectatorsSeeEverythingCheckBox->setChecked(gameInfo.spectators_omniscient());
     shareDecklistsOnLoadCheckBox->setChecked(gameInfo.share_decklists_on_load());
+    tournamentCheckBox->setChecked(gameInfo.is_tournament());
```
ZeldaZach: This re-enables the button that line 225 just disabled: `setChecked(true)` emits `toggled`, which is connected to `tournamentSettingsButton->setEnabled`. In the read-only game-info variant (`TabGame::actGameInfo`) the "Settings..." button ends up clickable for any tournament game, opening an editable dialog whose result goes nowhere.

Block signals around the `setChecked` or re-disable the button afterwards. While you're here: `games_per_match` from `gameInfo` isn't mirrored into `tournamentSettings`, so even a correctly-disabled dialog would show 1 rather than the game's real value.

cockatrice/src/interface/widgets/tabs/tab_game.cpp
```diff
@@ -928,8 +941,22 @@ void TabGame::stopGame()
...
+{
+    return tabSupervisor->switchToGameTabIfAlreadyExists(gameId);
+}
+
 void TabGame::closeGame()
 {
+    int parentId = game->getGameMetaInfo()->parentGameId();
+    if (parentId >= 0) {
+        if (switchToGameTab(parentId)) {
+            close();
```
ZeldaZach: Missing `return` after `close()` — execution continues into the `gameMenu` rebuild on a tab that has already been removed from the supervisor.

The comment below is also misleading: both paths fall through to the identical two lines, and neither reaches `leaveGame()`/`closeRequest()`, so there is no "normal leave-game flow" being used as a fallback. Either `return` after the successful switch and leave the comment off, or make the failure path actually do something different.

ZeldaZach (changes requested):
Thanks for pushing the stack this far — the split between `TournamentWidget` (pure state->view) and `TournamentTabGameExtension` (tab wiring) is a good shape, and keeping `TabGame` mostly untouched is appreciated.

A few things need fixing before this can go in. The biggest is that the host-facing "Tournament Settings" flow is inert: `Command_TournamentSettingsSelect` has no server-side override anywhere in the stack, so it always lands on `Server_AbstractParticipant::cmdTournamentSettingsSelect` returning `RespContextError`. Beyond that there's a stale host check, a button that renders with no label, a spectator path that can never trigger, and a games-per-match unit that means two different things in two files of this same PR.

Separately: CI (macOS 26 Debug) is red on `server_match_game_factory.h:6` `-Wmismatched-tags` (`GameConfig` declared `class` vs `struct`). That comes from #7199, but it blocks this PR too.

cockatrice/src/interface/widgets/tabs/tournament_tab_game_extension.cpp
BruebachL: Went with your second option and dropped the in-game settings button entirely (commit 6c9c144318). `Command_TournamentSettingsSelect` has no server handler today, so the dead command, the never-re-evaluated host gate, and the missing first-show label all disappear with it. The dialog can be re-added together with a working `Server_Player` override that seeds it from the current tournament settings and surfaces the command response.

cockatrice/src/interface/widgets/tabs/tournament_tab_game_extension.cpp
BruebachL: Moot — the settings button is gone (per the T1 resolution above), so there is no host-gated control to lazily create or toggle. No residual host-ness logic remains in the extension.

cockatrice/src/interface/widgets/tabs/tournament_tab_game_extension.cpp
BruebachL: Moot — the button no longer exists, so there is no blank-label-on-first-show path. `initializeTournamentMode()` now only inserts the status label, the standings button, and a spacer, none of which depend on widget creation order.

cockatrice/src/interface/widgets/draft/tournament_widget.cpp
BruebachL: Fixed — `TournamentWidget` now takes `setIsSpectator(bool)` (fed from `PlayerManager::isSpectator()` in the extension constructor) and the spectate branch keys off `isSpectator && hasAnyLivePairing` instead of the never-true `localPlayerId == -1`.

cockatrice/src/interface/widgets/draft/tournament_widget.cpp
BruebachL: Settled on "total games in the series" and aligned the widget to it. #7199 caps a match at `games_per_match` games, requires `games_per_match/2 + 1` wins, and formats sub-games as "Game X of N" — so the standings/status labels now read `Best of %1` with the raw `games_per_match` value (was `gamesPerMatch * 2 - 1`, which mislabeled Bo3 as Best of 5). The field name `games_per_match` now matches the total-series meaning everywhere.

cockatrice/src/interface/widgets/dialogs/dlg_tournament_settings.cpp
BruebachL: Resolved by the T5 fix — the dialog hint ("Set to 3 for Best of 3") defines games-per-match as total games, which is exactly what the server now does, and the standings header was the outlier. With the widget fixed to display the total directly, dialog and standings agree; no dialog change needed.

cockatrice/src/interface/widgets/dialogs/dlg_create_game.cpp
BruebachL: Fixed — the read-only `tournamentCheckBox->setChecked(...)` is now wrapped in a `QSignalBlocker` so the disabled settings button is not re-enabled by the `toggled` connection (commit 6c9c144318). On the mirroring point: `ServerInfo_Game` only carries `is_tournament` (no `games_per_match` field), so there is no value on `gameInfo` to mirror into `tournamentSettings` in this read-only dialog; with the button staying disabled the dialog can no longer be opened from it. Also fixed a latent build break from the #7197 rework: the create-game path now populates the shared `mutable_tournament_settings()` message instead of the removed `set_games_per_match` field.

cockatrice/src/interface/widgets/tabs/tab_game.cpp
BruebachL: Fixed — the sub-game now `return`s right after `close()` when it successfully switches to its parent tab, and the misleading "normal leave-game flow" comment is removed. The failure path (parent tab gone) now unambiguously falls through to the gameMenu rebuild only.

cockatrice/src/interface/widgets/dialogs/dlg_tournament_settings.cpp
```diff
@@ -0,0 +1,48 @@
...
+    hintLabel->setStyleSheet("color: palette(placeholderText);");
+    mainLayout->addRow(QString(), hintLabel);
```
RickyRister, line 23: Hint label's column alignment looks bad.

Image

We could also just make it a tooltip

PR #7203, opened 2026-08-29
[Models] Mirror custom deck zones in the deck list model

BruebachL:
DeckListModel now surfaces the custom zones from the deck tree so views can render and edit them alongside criteria groups.

- rebuildTree mirrors each custom zone as a DecklistModelSubZoneNode under its board zone, cards flat inside (no further grouping).
- The freshly built shadow tree is sorted while the model reset is still open, so views never observe unsorted intermediate order and proxies cannot desync.
- Custom zones always sort after criteria groups within a board, regardless of their names. sortHelper remaps persistent indexes
  through the pre-shift rows to keep selections valid.
- addCard inserts flat into a custom zone by name and keeps grouping by active criteria for board zones. findCardNode resolves cards in both layouts, legacy top-level zones unchanged.
- New IsCustomZoneRole lets views tell zones apart from groups.
- Empty custom zones survive row removal. Zone rows themselves are only mutable through the deck tree API.

### DIFF
--- a/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
+++ b/libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
@@ -66,7 +66,8 @@ void DeckListModel::rebuildTree()
         for (int j = 0; j < currentZone->size(); j++) {
             auto *currentCard = dynamic_cast<DecklistCardNode *>(currentZone->at(j));
 
-            //! \todo Better sanity checking.
+            // Non-card children are custom zones; they are mirrored in a single
+            // pass below so each is mirrored exactly once.
             if (currentCard == nullptr) {
                 continue;
             }
@@ -82,7 +83,18 @@ void DeckListModel::rebuildTree()
 
             new DecklistModelCardNode(currentCard, groupNode);
         }
+
+        // Custom zones nested under the board zone are mirrored as-is, with their
+        // cards as direct children (no further grouping).
+        DeckListModelCustomZones::mirrorCustomZones(currentZone, node);
     }
+
+    // The shadow tree was built in deck file order. Apply the active sort while
+    // the reset is still open so every consumer (tree view and visual editor)
+    // sees the canonical order from the start. sortShadowTree emits no signals,
+    // which is only valid before endResetModel closes the reset.
+    root->setSortMethod(lastKnownColumn == 0 ? DeckSortMethod::ByNumber : DeckSortMethod::ByName);
+    sortShadowTree(root, lastKnownOrder);
 
     endResetModel();
 
@@ -154,6 +166,9 @@ QVariant DeckListModel::data(const QModelIndex &index, int role) const
             case DeckRoles::IsLegalRole:
                 return true;
 
+            case DeckRoles::IsCustomZoneRole:
+                return DeckListModelCustomZones::isCustomZone(group);
+
             default:
                 return {};
         }
@@ -188,6 +203,10 @@ QVariant DeckListModel::data(const QModelIndex &index, int role) const
 
         case DeckRoles::IsLegalRole: {
             return card->getFormatLegality();
+        }
+
+        case DeckRoles::IsCustomZoneRole: {
+            return false;
         }
 
         default: {
@@ -327,6 +346,13 @@ bool DeckListModel::removeRows(int row, int count, const QModelIndex &parent)
         return false;
     }
 
+    // Custom zone rows are managed through the deck tree, never removed as model rows.
+    for (int i = 0; i < count; i++) {
+        if (DeckListModelCustomZones::isCustomZone(node->at(row + i))) {
+            return false;
+        }
+    }
+
     beginRemoveRows(parent, row, row + count - 1);
     for (int i = 0; i < count; i++) {
         AbstractDecklistNode *toDelete = node->takeAt(row);
@@ -337,7 +363,8 @@ bool DeckListModel::removeRows(int row, int count, const QModelIndex &parent)
     }
     endRemoveRows();
 
-    if (node->empty() && (node != root)) {
+    // Empty criteria groups get pruned, but custom zones stay until explicitly deleted.
+    if (node->empty() && (node != root) && !DeckListModelCustomZones::isCustomZone(node)) {
         removeRows(parent.row(), 1, parent.parent());
     } else {
         emitRecursiveUpdates(parent);
@@ -365,24 +392,44 @@ DecklistModelCardNode *DeckListModel::findCardNode(const QString &cardName,
                                                    const QString &providerId,
                                                    const QString &cardNumber) const
 {
-    InnerDecklistNode *zoneNode = dynamic_cast<InnerDecklistNode *>(root->findChild(zoneName));
-    if (!zoneNode) {
-        return nullptr;
-    }
-
     CardInfoPtr info = CardDatabaseManager::query()->getCardInfo(cardName);
     if (!info) {
         return nullptr;
     }
 
-    QString groupCriteria = extractGroupCriteriaValue(info, activeGroupCriteria);
-    InnerDecklistNode *groupNode = dynamic_cast<InnerDecklistNode *>(zoneNode->findChild(groupCriteria));
-    if (!groupNode) {
-        return nullptr;
+    // 1. Board zone lookup: search the criteria groups, then the custom zones
+    //    nested under the board.
+    if (auto *zoneNode = dynamic_cast<InnerDecklistNode *>(root->findChild(zoneName))) {
+        QString groupCriteria = extractGroupCriteriaValue(info, activeGroupCriteria);
+        if (auto *groupNode = dynamic_cast<InnerDecklistNode *>(zoneNode->findChild(groupCriteria))) {
+            if (auto *card = dynamic_cast<DecklistModelCardNode *>(
+                    groupNode->findCardChildByNameProviderIdAndNumber(cardName, providerId, cardNumber))) {
+                return card;
+            }
+        }
+
+        for (auto *child : *zoneNode) {
+            if (!DeckListModelCustomZones::isCustomZone(child)) {
+                continue;
+            }
+            auto *customZone = dynamic_cast<InnerDecklistNode *>(child);
+            if (!customZone) {
+                continue;
+            }
+            if (auto *card = dynamic_cast<DecklistModelCardNode *>(
+                    customZone->findCardChildByNameProviderIdAndNumber(cardName, providerId, cardNumber))) {
+                return card;
+            }
+        }
     }
 
-    return dynamic_cast<DecklistModelCardNode *>(
-        groupNode->findCardChildByNameProviderIdAndNumber(cardName, providerId, cardNumber));
+    // 2. Custom zone lookup by name (custom zone names are deck-unique).
+    if (auto *customZone = DeckListModelCustomZones::findSubZoneByName(root, zoneName)) {
+        return dynamic_cast<DecklistModelCardNode *>(
+            customZone->findCardChildByNameProviderIdAndNumber(cardName, providerId, cardNumber));
+    }
+
+    return nullptr;
 }
 
 QModelIndex DeckListModel::findCard(const QString &cardName,
@@ -423,29 +470,43 @@ QModelIndex DeckListModel::addCard(const ExactCard &card, const QString &zoneNam
         return {};
     }
 
-    InnerDecklistNode *zoneNode = createNodeIfNeeded(zoneName, root);
-
     CardInfoPtr cardInfo = card.getCardPtr();
     PrintingInfo printingInfo = card.getPrinting();
 
-    QString groupCriteria = extractGroupCriteriaValue(cardInfo, activeGroupCriteria);
-    InnerDecklistNode *groupNode = createNodeIfNeeded(groupCriteria, zoneNode);
+    InnerDecklistNode *cardParent = nullptr;
 
-    const QModelIndex parentIndex = nodeToIndex(groupNode);
-    auto *cardNode = dynamic_cast<DecklistModelCardNode *>(groupNode->findCardChildByNameProviderIdAndNumber(
+    auto *boardNode = dynamic_cast<InnerDecklistNode *>(root->findChild(zoneName));
+    auto *customZoneNode = boardNode ? nullptr : DeckListModelCustomZones::findSubZoneByName(root, zoneName);
+
+    if (boardNode) {
+        // Board zone: cards are grouped by the active criteria.
+        QString groupCriteria = extractGroupCriteriaValue(cardInfo, activeGroupCriteria);
+        cardParent = createNodeIfNeeded(groupCriteria, boardNode);
+    } else if (customZoneNode) {
+        // Custom zone: cards live flat inside the zone.
+        cardParent = customZoneNode;
+    } else {
+        // Unknown zone: create a top-level zone (legacy behavior).
+        QString groupCriteria = extractGroupCriteriaValue(cardInfo, activeGroupCriteria);
+        auto *newZone = createNodeIfNeeded(zoneName, root);
+        cardParent = createNodeIfNeeded(groupCriteria, newZone);
+    }
+
+    const QModelIndex parentIndex = nodeToIndex(cardParent);
+    auto *cardNode = dynamic_cast<DecklistModelCardNode *>(cardParent->findCardChildByNameProviderIdAndNumber(
         card.getName(), printingInfo.getUuid(), printingInfo.getProperty("num")));
     const auto cardSetName = printingInfo.getSet().isNull() ? "" : printingInfo.getSet()->getCorrectedShortName();
 
     bool cardNodeAdded = false;
     if (!cardNode) {
         // Determine the correct index
-        int insertRow = findSortedInsertRow(groupNode, cardInfo);
+        int insertRow = findSortedInsertRow(cardParent, cardInfo);
 
         auto *decklistCard = deckList->addCard(cardInfo->getName(), zoneName, insertRow, cardSetName,
                                                printingInfo.getProperty("num"), printingInfo.getProperty("uuid"));
 
         beginInsertRows(parentIndex, insertRow, insertRow);
-        cardNode = new DecklistModelCardNode(decklistCard, groupNode, insertRow);
+        cardNode = new DecklistModelCardNode(decklistCard, cardParent, insertRow);
         endInsertRows();
 
         cardNodeAdded = true;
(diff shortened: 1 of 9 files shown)


tests/deck_list_model/deck_list_model_custom_zones_test.cpp
```diff
@@ -0,0 +1,225 @@
...
+    root.setSortMethod(DeckSortMethod::ByName);
+
+    auto mapping = DeckListModelCustomZones::sortWithCustomZonesLast(&root, group, Qt::AscendingOrder);
+    EXPECT_EQ(childNames(group), (QStringList{"Bolt", "Swords"}));
+    ASSERT_EQ(mapping.size(), 2);
+    EXPECT_EQ(mapping[0].first, 1); // "Bolt" was originally at row 1
+    EXPECT_EQ(mapping[0].second, 0);
+    EXPECT_EQ(mapping[1].first, 0);
+    EXPECT_EQ(mapping[1].second, 1);
+}
```
ZeldaZach, line 270: Both new test binaries are missing `main()`, which is why Debian 13 / Fedora 44 / Ubuntu 26.04 fail to link — and the failure propagates to #7204 through #7207:

```
ld: Scrt1.o: in function `_start': (.text+0x17): undefined reference to `main'
FAILED: tests/deck_list_model/deck_list_model_custom_zones_test
FAILED: tests/deck_list_model/deck_list_model_zone_integration_test
```

The cause is `tests/CMakeLists.txt:62`: in the bundled-googletest fallback it sets `GTEST_BOTH_LIBRARIES` to plain `gtest`, without `gtest_main`. That's exactly the path those three runners take (their link lines show `lib/libgtest.a` and nothing else). Every other gtest binary in the repo compensates by defining its own entry point — `deck_list_zones_test.cpp:352`, `carddatabase_test.cpp:38`, `card_zone_algorithms_test.cpp:155`, and so on.

Append the repo-standard block to the end of both files:

```cpp
int main(int argc, char **argv)
{
    ::testing::InitGoogleTest(&argc, argv);
    return RUN_ALL_TESTS();
}
```

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -423,29 +470,43 @@ QModelIndex DeckListModel::addCard(const ExactCard &card, const QString &zoneNam
...
+
+    auto *boardNode = dynamic_cast<InnerDecklistNode *>(root->findChild(zoneName));
+    auto *customZoneNode = boardNode ? nullptr : DeckListModelCustomZones::findSubZoneByName(root, zoneName);
 
-    const QModelIndex parentIndex = nodeToIndex(groupNode);
-    auto *cardNode = dynamic_cast<DecklistModelCardNode *>(groupNode->findCardChildByNameProviderIdAndNumber(
+    if (boardNode) {
+        // Board zone: cards are grouped by the active criteria.
+        QString groupCriteria = extractGroupCriteriaValue(cardInfo, activeGroupCriteria);
+        cardParent = createNodeIfNeeded(groupCriteria, boardNode);
```
ZeldaZach, line 492: `createNodeIfNeeded` resolves through `InnerDecklistNode::findChild(name)`, which matches *any* child by name — and after this PR a board zone's children include the mirrored `DecklistModelSubZoneNode`s. So a custom zone whose name happens to equal a grouping-criteria value swallows cards that belong at board level.

`rebuildTree` is safe only by construction: it mirrors custom zones at line 89, *after* the card loop. `addCard` has no such ordering.

Concretely, with `MAIN_TYPE` grouping: main has no creatures yet (so no "Creature" group node) and the user has made a custom zone named `Creature` — nothing forbids it, `DecklistNodeTree::hasZoneName` only reserves the four board names and existing zone names. Add Grizzly Bears to main: `findChild("Creature")` returns the custom zone, the shadow card node lands inside it, while `deckList->addCard(..., "main", ...)` on line 505 puts the real card directly under the board. The editor shows it in the zone, the saved `.cod` has it at board level, and the next `rebuildTree` moves it into a real `Creature` group. With `MANA_COST` grouping the collision surface is wider still, since group names are bare numbers.

A `findGroupChild(parent, name)` helper in `DeckListModelCustomZones` that skips `isCustomZone` children, used here for the group lookup, would close it. `findCardNode` line 404 does the same `findChild(groupCriteria)` and wants the same treatment — it's more benign there since the custom-zone loop below would find the card anyway, but the two should agree.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -423,29 +470,43 @@ QModelIndex DeckListModel::addCard(const ExactCard &card, const QString &zoneNam
...
+    } else if (customZoneNode) {
+        // Custom zone: cards live flat inside the zone.
+        cardParent = customZoneNode;
+    } else {
+        // Unknown zone: create a top-level zone (legacy behavior).
+        QString groupCriteria = extractGroupCriteriaValue(cardInfo, activeGroupCriteria);
+        auto *newZone = createNodeIfNeeded(zoneName, root);
```
ZeldaZach, line 543: Nit: `findSubZoneByName` searches only the shadow tree, which learns about custom zones exclusively via `rebuildTree`. If a caller adds a zone straight on the deck tree and then calls `addCard(card, "Removal")` without a rebuild, both `boardNode` and `customZoneNode` are null and this legacy branch creates **"Removal" as a top-level board row** next to main/side — while `DecklistNodeTree::getZoneObjFromName` resolves it to the real nested zone, so the deck and the model disagree with no way for the model to notice.

Nothing in the stack does this today (`DeckStateManager::createCustomZone` → `modifyTree` → `rebuildTree`), but `getDeckList()` is public and the new integration tests take exactly that shortcut. Consulting `deckList->getTree()` before falling through to legacy behavior, or adding model-level zone wrappers that own the rebuild, would make it robust rather than relying on callers.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model_custom_zones.cpp
```diff
@@ -0,0 +1,123 @@
...
+        auto *customZone = dynamic_cast<const InnerDecklistNode *>(deckBoardZone->at(j));
+        if (!customZone) {
+            continue;
+        }
+
+        auto *shadowZone = new DecklistModelSubZoneNode(customZone->getName(), shadowBoardZone);
+        for (int k = 0; k < customZone->size(); k++) {
+            if (auto *zoneCard = dynamic_cast<DecklistCardNode *>(customZone->at(k))) {
+                new DecklistModelCardNode(zoneCard, shadowZone);
+            }
```
ZeldaZach: Nit: this mirrors only direct `DecklistCardNode` children, so cards nested more than one level deep get no model row at all. `InnerDecklistNode::readElement` recurses on `<zone>` to arbitrary depth (`inner_deck_list_node.cpp:151-153`), so a `.cod` with `main > Removal > Deeper` loads fine — and those cards then count in `recursiveCount`, `forEachCard`, and the deck hash, and are re-serialized on save, while being invisible and unselectable in the editor. The board's Count column ends up larger than the sum of its visible rows.

The pre-PR code skipped all non-card children so this isn't a regression, and `DecklistNodeTree::getCustomZones` shares the one-level assumption, so it's at least consistent. But since this is where the mirroring contract gets written down, it's worth either flattening grandchild cards into the mirrored zone or having the loader collapse depth > 1 on read, rather than leaving cards nothing can reach.

tests/deck_list_model/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+add_executable(deck_list_model_custom_zones_test deck_list_model_custom_zones_test.cpp)
```
ZeldaZach: Nit: this target links `libcockatrice_card` without `${VERSION_STRING_CPP}`, unlike its sibling on line 18 and unlike every other test that links that library (`tests/carddatabase/CMakeLists.txt:12`, `tests/loading_from_clipboard/CMakeLists.txt:2`). `libcockatrice_card` references the extern `VERSION_STRING` from `cockatrice_xml_3.cpp` / `cockatrice_xml_4.cpp`, which only the generated `version_string.cpp` defines.

It links today because nothing in this file pulls `card_database_loader.cpp.o` out of the archive, but the first `DeckListModel` assertion added here turns it into an undefined-symbol error. Either add `${VERSION_STRING_CPP}` to match, or drop the unused `libcockatrice_card` / `libcockatrice_models` links — the file only exercises header-inline nodes plus `deck_list_model_custom_zones.cpp`.

tests/deck_list_model/deck_list_model_zone_integration_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+    tree->addCard("Lightning Bolt", 2, DECK_ZONE_MAIN, -1);
+    ASSERT_NE(tree->addCustomZone(DECK_ZONE_MAIN, "Removal"), nullptr);
+    tree->addCard("Swords to Plowshares", 1, "Removal", -1);
+    ASSERT_NE(tree->addCustomZone(DECK_ZONE_MAIN, "Utility"), nullptr);
+
+    model.rebuildTree();
+
+    EXPECT_EQ(model.getCustomZoneNames(DECK_ZONE_MAIN), (QStringList{"Removal", "Utility"}));
+    EXPECT_EQ(totalCustomZoneRows(model), 2);
+}
```
ZeldaZach, line 101: The tests here genuinely assert what the PR description claims for the free functions — `isCustomZone`, `findSubZoneByName`, `mirrorCustomZones` including the empty-zone case, and both sort orders with the mapping invariant.

What's untested is the model behavior this PR changed, all of it reachable through the public API with the scaffolding already in this file:

- `addCard` routing into a custom zone (`deck_list_model.cpp:485-487`)
- `findCardNode`'s two new lookup paths (`:411-430`)
- `removeRows` refusing a custom-zone row (`:349-354`) — worth covering specifically, because an *empty* custom zone has `hasChildren() == false`, so this guard is the only thing stopping `DeckStateManager::removeCardAtIndex` from deleting it as if it were a card
- the "empty custom zone survives pruning" branch (`:367`)

The last two are the ones the PR description calls out as behavior, so they're the most valuable to pin down.

ZeldaZach (changes requested):
Reviewed as the base of the custom-zones stack (#7203 → #7204 → #7205 → #7206 → #7207).

The core mechanism holds up. I checked the `(preSortRow, finalRow)` mapping from `boardSort`/`plainSort` against `InnerDecklistNode::sort`'s `result[i] = {oldIndex, i}` contract and it's equivalent to what `sortHelper` consumed before, so `changePersistentIndexList` stays correct; the forward-vs-reverse iteration change is safe because `changePersistentIndexList` erases all `from` entries before re-inserting. `begin/endInsertRows` and `begin/endRemoveRows` pairing is untouched, shadow-node ownership is right (`~InnerDecklistNode` → `clearTree()` deletes recursively, and `DecklistModelCardNode` correctly doesn't own its data node), and sorting inside the open `beginResetModel`/`endResetModel` window without layout signals is legitimate since `endResetModel` invalidates persistent indices anyway. The `<../../../..>` include style matches what `deck_list_model.h` already does. clang-format is clean.

Two things to fix. One is breaking CI right now; the other is a model/deck divergence on a name collision.

The Debian 13, Fedora 44, and Ubuntu 26.04 jobs are red on this PR and every PR stacked above it — that's this branch, not flaky infra.

tests/deck_list_model/deck_list_model_custom_zones_test.cpp
BruebachL: Appended the standard `main()` block to both test binaries (`deck_list_model_custom_zones_test.cpp` and `deck_list_model_zone_integration_test.cpp`), so they link and run on the bundled-gtest fallback.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: Added `DeckListModelCustomZones::findGroupChild(parent, name)`, which searches only non-`isCustomZone` children, and used it for the group lookup in both `createNodeIfNeeded` and `findCardNode` so a custom zone named like a grouping criterion no longer swallows board-level cards.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: `addCard` now consults the deck tree before falling back to the legacy path: if a custom zone with that name exists on the deck tree it rebuilds and re-routes the card into it instead of creating a top-level board row.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model_custom_zones.cpp
BruebachL: `mirrorCustomZones` now flattens every card under a custom zone into the mirrored zone at any nesting depth via a recursive helper, so cards nested in sub-zones keep a model row.

tests/deck_list_model/CMakeLists.txt
BruebachL: Added `${VERSION_STRING_CPP}` to the first target to match its sibling and the other `libcockatrice_card`-linked tests.

tests/deck_list_model/deck_list_model_zone_integration_test.cpp
BruebachL: Added model behaviour tests: `addCard` routing into mirrored and un-mirrored custom zones, same-name group/zone collision, `findCard` inside a custom zone, `removeRows` refusing a custom-zone row, and the empty zone surviving mirror + pruning.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -423,29 +471,63 @@ QModelIndex DeckListModel::addCard(const ExactCard &card, const QString &zoneNam
...
+    } else {
+        // Not present in the shadow tree. The deck tree may still hold a custom
+        // zone that has not been mirrored (callers can add a zone and then a
+        // card without a rebuild). Check before falling back to creating a
+        // top-level zone the deck does not actually have.
+        auto *listRoot = deckList->getTree()->getRoot();
+        bool hasDeckZone = false;
+        for (int i = 0; i < listRoot->size(); ++i) {
+            if (auto *boardZone = dynamic_cast<InnerDecklistNode *>(listRoot->at(i))) {
+                if (boardZone->findChild(zoneName)) {
```
ZeldaZach: `InnerDecklistNode::findChild` matches *any* child by name (`inner_deck_list_node.cpp:59` compares `at(i)->getName()`), so a `DecklistCardNode` sitting directly under a board zone satisfies this check.

When `zoneName` matches a card name but no zone: `hasDeckZone` is true, `rebuildTree()` mirrors nothing new (a card is not an `InnerDecklistNode`, so `mirrorCustomZones` skips it), the recursive `addCard` lands in the same `else` branch, and it recurses forever. Stack overflow instead of the intended legacy fallback.

Narrow trigger, one-line fix — `dynamic_cast` the result so only real zones count:

```cpp
if (dynamic_cast<InnerDecklistNode *>(boardZone->findChild(zoneName))) {
```

`DecklistNodeTree::findCustomZoneByName` already does exactly this loop with the cast; promoting it to public and calling it here would be the cleaner version.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
```diff
@@ -423,29 +471,63 @@ QModelIndex DeckListModel::addCard(const ExactCard &card, const QString &zoneNam
...
+    const QModelIndex parentIndex = nodeToIndex(cardParent);
+    auto *cardNode = dynamic_cast<DecklistModelCardNode *>(cardParent->findCardChildByNameProviderIdAndNumber(
         card.getName(), printingInfo.getUuid(), printingInfo.getProperty("num")));
     const auto cardSetName = printingInfo.getSet().isNull() ? "" : printingInfo.getSet()->getCorrectedShortName();
 
     bool cardNodeAdded = false;
     if (!cardNode) {
         // Determine the correct index
-        int insertRow = findSortedInsertRow(groupNode, cardInfo);
+        int insertRow = findSortedInsertRow(cardParent, cardInfo);
```
ZeldaZach, line 555: `insertRow` is a row index in the *shadow* zone, but it is handed to `deckList->addCard` on line 526 as a position in the *deck* zone (`DeckList::addCard` → `DecklistNodeTree::addCard` → `new DecklistCardNode(..., zoneNode, position)` → `AbstractDecklistNode` → `parent->insert(position, this)`).

For the board-zone path those spaces have always differed but the shadow group is a subset of the board's cards, so the index stays in range. The custom-zone path is different now that `flattenCards` recurses: a deck zone with one nested sub-zone holding 10 cards has `size() == 1` on the deck side and 10 rows on the shadow side, so `findSortedInsertRow` can return up to 10 and `QList::insert(10, …)` runs off a list of size 1 — `Q_ASSERT_X` in debug, out-of-bounds write in release.

That is precisely the shape the flattening was added to support, so a nested `.cod` plus one "add card to zone" reaches it. Either clamp/translate the shadow row into a deck-tree position before passing it down, or append (`-1`) when the target zone's flattened row count doesn't match its direct child count.

ZeldaZach (changes requested):
Second pass. All six points from the last round are addressed and verified: `main()` is in both test binaries, `${VERSION_STRING_CPP}` is on both targets, `findGroupChild` skips mirrored zones in `createNodeIfNeeded` and `findCardNode`, `addCard` consults the deck tree before the legacy fallback, `flattenCards` reaches nested cards, and the new integration tests cover the routing/guard/pruning behavior. CI is fully green.

Two issues in the new code, both consequences of the fixes rather than of the original design.

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: Fixed in f93f492e15: the `hasDeckZone` check now casts the `findChild` result to `InnerDecklistNode`, so a board card that merely shares the zone name no longer counts as a zone and the addCard/rebuildTree recursion can't happen. Added a regression test (`AddCardToCardNamedZoneDoesNotRecurse`).

libcockatrice_models/libcockatrice/models/deck_list/deck_list_model.cpp
BruebachL: Fixed in f93f492e15: `addCard` now resolves the deck-side custom zone and, when any of its direct children is a nested sub-zone (the shape that makes `flattenCards` diverge from the deck tree), passes position `-1` (append) to `deckList->addCard` instead of the partial shadow row index — the shadow node keeps its sorted row. Added a regression test (`AddCardToCustomZoneWithNestedSubZoneAppends`).

ZeldaZach (approved):

PR #7204, opened 2026-08-29
[Client] Add zone management to the deck state manager

BruebachL:
## Related Ticket(s)
- Based on #7177

- Add custom zone management to `DeckStateManager`: create/rename/remove zones and track their state
- Add `DeckZoneDialog` for editing custom deck zones

## Why
- Needed by the editor zone menus and display widgets further up this stack

### DIFF
--- a/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
@@ -470,20 +470,13 @@ QString DeckStateManager::validateNewZoneName(const QString &zoneName) const
 
     const auto *tree = deckList->getTree();
 
-    // Top-level zones (boards and legacy zones) claim their names too.
-    for (int i = 0; i < tree->getRoot()->size(); i++) {
-        if (tree->getRoot()->at(i)->getName() == trimmedZoneName) {
-            return tr("A zone with this name already exists.");
-        }
-    }
-
-    // Custom zone names are unique across the whole deck.
-    for (const QString &board : InnerDecklistNode::boardZoneNames()) {
-        for (const auto *customZone : tree->getCustomZones(board)) {
-            if (customZone->getName() == trimmedZoneName) {
-                return tr("A zone with this name already exists.");
-            }
-        }
+    // Reuse the tree's own uniqueness contract: any top-level zone and any
+    // custom zone on *every* board claims the name (hasZoneName also reserves
+    // the standard board names, which we already rejected with a dedicated
+    // message above). Scanning only the standard boards here would miss a
+    // custom zone an imported deck carries under `tokens`.
+    if (tree->hasZoneName(trimmedZoneName)) {
+        return tr("A zone with this name already exists.");
     }
 
     return {};
@@ -558,6 +551,7 @@ bool DeckStateManager::modifyTree(const QString &reason, const std::function<boo
         historyManager->save(memento);
         deckListModel->rebuildTree();
         deckList->refreshDeckHash();
+        emit deckListModel->deckHashChanged();
         doCardModified();
     }
 
--- a/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
@@ -407,25 +407,15 @@ bool DeckStateManager::moveCustomZone(const QString &zoneName, const QString &ne
 {
     const auto *tree = deckList->getTree();
 
-    // Locate the board currently holding the zone.
-    QString currentBoardName;
-    for (const QString &boardName : InnerDecklistNode::boardZoneNames()) {
-        for (const auto *zone : tree->getCustomZones(boardName)) {
-            if (zone->getName() == zoneName) {
-                currentBoardName = boardName;
-                break;
-            }
-        }
-        if (!currentBoardName.isEmpty()) {
-            break;
-        }
-    }
-
-    if (currentBoardName.isEmpty()) {
+    // Locate the zone through the tree's own lookup, which walks every top-level
+    // zone (not just the standard boards) and covers the same-board no-op below.
+    const auto *zone = tree->findCustomZoneByName(zoneName);
+    if (!zone) {
         return false;
     }
 
     // Same-board moves are no-ops and must not pollute the history.
+    const QString currentBoardName = zone->getParent() ? zone->getParent()->getName() : QString();
     if (currentBoardName == newBoardZoneName) {
         return true;
     }
@@ -433,8 +423,8 @@ bool DeckStateManager::moveCustomZone(const QString &zoneName, const QString &ne
     // Zone names are deck-unique among zones created through this manager, so a
     // same-named zone on the target board can only come from an imported deck.
     // Refuse the move instead of silently stacking same-named zones.
-    for (const auto *zone : tree->getCustomZones(newBoardZoneName)) {
-        if (zone->getName() == zoneName) {
+    for (const auto *targetZone : tree->getCustomZones(newBoardZoneName)) {
+        if (targetZone->getName() == zoneName) {
             return false;
         }
     }
@@ -552,6 +542,9 @@ bool DeckStateManager::modifyTree(const QString &reason, const std::function<boo
         deckListModel->rebuildTree();
         deckList->refreshDeckHash();
         emit deckListModel->deckHashChanged();
+        // removeCustomZone can drop whole card sets the model never notified
+        // about (rebuildTree emits no cardNodesChanged), so tell the consumers.
+        emit deckListModel->cardNodesChanged();
         doCardModified();
     }
 
--- a/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
@@ -2,6 +2,7 @@
 
 #include <libcockatrice/card/database/card_database_manager.h>
 #include <libcockatrice/deck_list/deck_list_history_manager.h>
+#include <libcockatrice/deck_list/tree/inner_deck_list_node.h>
 
 DeckStateManager::DeckStateManager(QObject *parent)
     : QObject(parent), deckList(QSharedPointer<DeckList>(new DeckList)),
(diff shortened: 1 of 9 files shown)


cockatrice/src/interface/widgets/deck_editor/deck_zone_dialog.h
```diff
@@ -0,0 +1,123 @@
...
+ * Convenience wrapper that runs DeckZoneDialog modally.
+ *
+ * @param parent The parent widget for the dialog
+ * @param initialBoardName The board zone to preselect in the dialog. Unknown names fall
+ *        back to main.
+ * @param chosenBoardName (out) The internal name of the board zone the user chose
+ * @param nameValidator Optional validator forwarded to the dialog
+ * @return The trimmed zone name, or an empty string if the user cancelled
+ */
+QString promptForNewZone(QWidget *parent,
```
RickyRister: make these into static methods in the class

cockatrice/src/interface/widgets/deck_editor/deck_zone_dialog.cpp
```diff
@@ -0,0 +1,139 @@
...
+                               const std::function<QString(const QString &)> &_nameValidator,
+                               bool _allowBoardSelection)
+    : QDialog(parent), nameValidator(_nameValidator), allowBoardSelection(_allowBoardSelection)
+{
+    nameLabel = new QLabel(this);
+    nameEdit = new QLineEdit(this);
+    nameEdit->setMaxLength(MAX_NAME_LENGTH);
+
+    errorLabel = new QLabel(this);
+    errorLabel->setStyleSheet("color: palette(bright-text);");
```
RickyRister: Text is white on white

Image

cockatrice/src/interface/widgets/deck_editor/deck_zone_dialog.cpp
```diff
@@ -0,0 +1,139 @@
...
+    errorLabel->hide();
+
+    boardLabel = new QLabel(this);
+    boardCombo = new QComboBox(this);
+    for (const QString &boardName : {QString(DECK_ZONE_MAIN), QString(DECK_ZONE_SIDE), QString(DECK_ZONE_MAYBEBOARD)}) {
+        boardCombo->addItem({}, boardName);
```
RickyRister: Options are all empty

Image

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
```diff
@@ -307,6 +308,187 @@ bool DeckStateManager::decrementCountAtIndex(const QModelIndex &idx)
...
+    });
+}
+
+bool DeckStateManager::moveCustomZone(const QString &zoneName, const QString &newBoardZoneName)
+{
+    const auto *tree = deckList->getTree();
+
+    // Locate the board currently holding the zone.
+    QString currentBoardName;
+    for (const QString &boardName : {QString(DECK_ZONE_MAIN), QString(DECK_ZONE_SIDE), QString(DECK_ZONE_MAYBEBOARD)}) {
```
RickyRister: This list appears in a several places in the following PRs. Maybe move it to a constant?

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
```diff
@@ -367,6 +549,21 @@ void DeckStateManager::requestHistorySave(const QString &reason)
...
 
+bool DeckStateManager::modifyTree(const QString &reason, const std::function<bool(DecklistNodeTree *)> &operation)
+{
+    DeckListMemento memento = deckList->createMemento(reason);
+    bool success = operation(deckList->getTree());
+
+    if (success) {
+        historyManager->save(memento);
+        deckListModel->rebuildTree();
+        deckList->refreshDeckHash();
```
ZeldaZach, line 543: `refreshDeckHash()` recomputes the hash but nothing tells the UI about it, so the deck hash label goes stale after every zone operation.

`DeckEditorDeckDockWidget` updates that label from exactly two places: `DeckListModel::deckHashChanged` (connected at `deck_editor_deck_dock_widget.cpp:62`) and `syncDisplayWidgetsToModel`, driven by `DeckStateManager::deckReplaced`. `modifyTree` emits neither. Every `DeckListModel` mutator pairs the refresh with the signal (`deck_list_model.cpp:332`, `:373`, `:527`), which is why the `modifyDeck` path never had this problem.

Concretely: create "Ramp" under the mainboard, move ten cards into it, then delete the zone. `computeDeckHash()` recurses into custom zones via `getCardNodes({DECK_ZONE_MAIN})`, so the real hash changes, but the dock keeps showing the old one until the user touches an unrelated card. `moveCustomZone` between main and side has the same effect, since the hash prefixes sideboard cards with `SB:`.

```suggestion
            deckList->refreshDeckHash();
            emit deckListModel->deckHashChanged();
```

(`replaceDeck` already emits a model signal this way at line 542, so the style fits. A small `DeckListModel::refreshHash()` doing both would be tidier still.)

cockatrice/src/interface/widgets/deck_editor/deck_zone_dialog.cpp
```diff
@@ -0,0 +1,142 @@
...
+    connect(buttonBox, &QDialogButtonBox::rejected, this, &QDialog::reject);
+
+    auto *layout = new QVBoxLayout(this);
+    layout->addWidget(nameLabel);
+    layout->addWidget(nameEdit);
+    layout->addWidget(errorLabel);
+    if (allowBoardSelection) {
+        layout->addWidget(boardLabel);
+        layout->addWidget(boardCombo);
+    }
```
ZeldaZach, line 56: `boardLabel` and `boardCombo` are constructed with `this` as parent at lines 25-26 but only added to the layout when `allowBoardSelection` is true. A child widget of a visible parent is shown unless it was explicitly hidden — layout membership only controls geometry, not visibility. `errorLabel` gets its `hide()` at line 23; these two don't.

So in the `promptForRename` path (`allowBoardSelection == false`) the "Parent zone:" label and the combo box are painted at their default geometry in the dialog's top-left corner, on top of the name label and line edit — and the combo is live there, so it can take clicks and focus.

```suggestion
    if (allowBoardSelection) {
        layout->addWidget(boardLabel);
        layout->addWidget(boardCombo);
    } else {
        boardLabel->hide();
        boardCombo->hide();
    }
```

Constructing them lazily would be cleaner, but `retranslateUi()` dereferences both unconditionally, so that route needs null guards too.

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
```diff
@@ -307,6 +308,187 @@ bool DeckStateManager::decrementCountAtIndex(const QModelIndex &idx)
...
+
+    // Top-level zones (boards and legacy zones) claim their names too.
+    for (int i = 0; i < tree->getRoot()->size(); i++) {
+        if (tree->getRoot()->at(i)->getName() == trimmedZoneName) {
+            return tr("A zone with this name already exists.");
+        }
+    }
+
+    // Custom zone names are unique across the whole deck.
+    for (const QString &board : InnerDecklistNode::boardZoneNames()) {
```
ZeldaZach: Nit: this uniqueness scan is narrower than the check that actually enforces it. `DecklistNodeTree::hasZoneName` → `findCustomZoneByName` walks the children of *every* top-level zone, while this only walks `boardZoneNames()` (main/side/maybeboard). `InnerDecklistNode::readElement` will construct a nested `<zone>` under any parent, so a hand-edited or imported deck can carry a custom zone under `tokens`.

In that case the validator returns empty, the dialog enables OK, `addCustomZone` returns `nullptr`, and `createCustomZone` fails silently with the dialog already closed. Reusing the tree's own check — expose a const `hasZoneName` on `DecklistNodeTree` and call it here — would also stop the two from drifting apart later.

ZeldaZach (changes requested):

The plumbing here is solid. `modifyTree` mirrors `modifyDeck`'s memento-before-operation pattern, node ownership is correct (`~InnerDecklistNode` recursively deletes, so `removeCustomZone` frees its cards rather than orphaning them), and the model's shadow tree is rebuilt immediately after every tree mutation so no stale `DecklistCardNode*` survives. I confirmed custom zones round-trip through the memento and through `.cod` save/load, since `InnerDecklistNode::readElement`/`writeElement` already recurse into nested `<zone>` elements. Reserved-name handling, no-op renames, and the modified flag all check out.

Two things to fix, plus one nit. Both are latent right now — nothing in this PR calls the new API yet — but they bite as soon as #7205 lands.

On CI: the Debian 13 / Fedora 44 / Ubuntu 26.04 failures are inherited from #7203, whose two new test files have no `main()`. The macOS 13 Intel failure here is unrelated infra (`Failed to CreateArtifact: ENOTFOUND`).

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
BruebachL: `modifyTree` now emits `deckListModel->deckHashChanged()` right after `refreshDeckHash()`, so the dock\u2019s hash label stays current after zone create/rename/move/remove.

cockatrice/src/interface/widgets/deck_editor/deck_zone_dialog.cpp
BruebachL: `boardLabel` and `boardCombo` are now explicitly hidden when `allowBoardSelection` is false, so the rename dialog no longer paints the live combo over the name field.

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
BruebachL: `validateNewZoneName` now reuses the tree\u2019s own `DecklistNodeTree::hasZoneName` (made const-public for this), which covers custom zones on every board instead of only the standard ones.

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
```diff
@@ -367,6 +542,22 @@ void DeckStateManager::requestHistorySave(const QString &reason)
...
+bool DeckStateManager::modifyTree(const QString &reason, const std::function<bool(DecklistNodeTree *)> &operation)
+{
+    DeckListMemento memento = deckList->createMemento(reason);
+    bool success = operation(deckList->getTree());
+
+    if (success) {
+        historyManager->save(memento);
+        deckListModel->rebuildTree();
+        deckList->refreshDeckHash();
+        emit deckListModel->deckHashChanged();
```
ZeldaZach, line 544: Follow-up, same family as the hash signal, no need to block on it: `cardNodesChanged` has the same gap.

`DeckListModel` only fires it off `cardNodeAddedAt`, `cardNodeRemoved` and its own `deckReplaced` (`deck_list_model.cpp:21-23`); `rebuildTree()` emits none of those. So `removeCustomZone` on a zone that still holds cards changes the unique-card set without anyone hearing about it. Two consumers go stale: `DeckEditorDeckDockWidget`'s banner-card combo (`deck_editor_deck_dock_widget.cpp:166`), which keeps offering the deleted cards and will happily set a banner card the deck no longer contains, and `PrintingSelector::printingsInDeckChanged` via `DeckStateManager::uniqueCardsChanged` (`printing_selector.cpp:92`), whose in-deck counts stay wrong until the next unrelated card edit.

The other three `modifyTree` callers don't move cards, so an unconditional emit alongside the hash one is the simplest fix:

```suggestion
            emit deckListModel->deckHashChanged();
            emit deckListModel->cardNodesChanged();
```

Separately, `moveCustomZone` still hand-rolls the `boardZoneNames()` scan to find the owning board, which is the narrower lookup you just replaced in `validateNewZoneName` — a zone the tree can find via `findCustomZoneByName` but that isn't under main/side/maybeboard makes it return false with no feedback. `DecklistNodeTree::moveCustomZone` already handles both the lookup and the same-board no-op, so the manager only really needs the duplicate-name guard on the target board.

ZeldaZach (approved):
All three items from my last pass are fixed: `modifyTree` now emits `deckHashChanged()`, the rename dialog hides `boardLabel`/`boardCombo` instead of painting them over the name field, and `validateNewZoneName` delegates to `DecklistNodeTree::hasZoneName` (now public) so custom zones on non-standard boards are covered. RickyRister's four points are in as well: static `promptFor*` helpers, the three-argument `addItem` overload for the combo user data, and the shared `InnerDecklistNode::boardZoneNames()` constant. CI is fully green now that #7203's test targets build.

Re-checked the parts my last review touched and they still hold: node ownership on `removeCustomZone`, the memento-before-operation ordering, and the shadow-tree rebuild after every tree mutation. `moveCardToZone` matches the existing `doSwapCard` convention, `removeRows` correctly refuses to prune custom zones and correctly leaves the board zone alive while it still owns one, and the ancestor walk resolves the source zone correctly for both grouped board cards and flat custom-zone cards (`EditRole` on `CARD_NAME` returns the internal name, so the equality check against `targetZoneName` is right).

Approving. One follow-up left inline that can ride along with #7205 rather than block this.

cockatrice/src/interface/widgets/deck_editor/deck_state_manager.cpp
BruebachL: Both parts addressed in 3cfc471802.

1) `modifyTree` now emits `deckListModel->cardNodesChanged()` alongside `deckHashChanged()`, so the banner-card combo and `PrintingSelector::printingsInDeckChanged` refresh after `removeCustomZone` (and the other non-card-moving callers are unaffected by an unconditional emit).

2) `DecklistNodeTree::findCustomZoneByName` is now public and `moveCustomZone` uses it — the zone is located by walking every top-level zone (so a zone under e.g. `tokens` is found and moved instead of returning false with no feedback), the same-board no-op is still detected via the zone's parent, and the duplicate-name guard on the target board is kept. Added a regression test (`MoveCustomZoneNestedUnderTokensBoard`).

PR #7205, opened 2026-08-29
[Client] Expose custom zone management in the deck editor

BruebachL:
## Related Ticket(s)
- Stacked on custom-zones/b1-zone-management (#7204)

- Add a `Move to &zone` submenu to the deck-row context menu: boards without custom zones appear flat, boards with zones nest (`Boardname ▸ { Boardname (whole board), … }`), the current board's whole-board entry is disabled, and token cards get no move menu at all
- Add `Create new zone and move &here...` to the move menu, which creates the zone and moves the right-clicked card into it
- Fix zone rename validation so the unchanged name doesn't open the dialog in a pre-filled error state
- Show the delete-zone hint as a tooltip (only when the zone still has cards) instead of a never-surfacing status tip
- Nest the "Add to Zone" card-database menu per board the same way, and make `Create new zone...` add the card to the created zone
- Fix plain-text export and print crashes on decks with custom zones: `saveToStream_DeckZone` recurses into sub-zones, `printDeckListNode` dispatches children by type, and `InnerDecklistNode::height()` guards the empty node

## Why
- The move/create actions are now unambiguous and do what the user asked; decks with custom zones can be exported and printed without crashing

## Verified

### DIFF
--- a/cockatrice/src/interface/deck_loader/deck_loader.cpp
+++ b/cockatrice/src/interface/deck_loader/deck_loader.cpp
@@ -384,6 +384,14 @@ void DeckLoader::saveToStream_DeckZone(QTextStream &out,
 
     for (int j = 0; j < zoneNode->size(); j++) {
         auto *card = dynamic_cast<DecklistCardNode *>(zoneNode->at(j));
+        if (!card) {
+            // Cards collected in nested sub-zones are exported by recursion so
+            // they don't end up invisible in the plain text output.
+            if (auto *subZone = dynamic_cast<const InnerDecklistNode *>(zoneNode->at(j))) {
+                saveToStream_DeckZone(out, subZone, addComments, addSetNameAndNumber);
+            }
+            continue;
+        }
 
         CardInfoPtr info = CardDatabaseManager::query()->getCardInfo(card->getName());
         QString cardType = info ? info->getMainCardType() : "unknown";
@@ -510,9 +518,26 @@ bool DeckLoader::convertToCockatriceFormat(LoadedDeck &deck)
 
 void DeckLoader::printDeckListNode(QTextCursor *cursor, const InnerDecklistNode *node)
 {
+    if (!node || node->isEmpty()) {
+        return;
+    }
+
     const int totalColumns = 2;
 
-    if (node->height() == 1) {
+    // Dispatch children by type instead of trusting a whole-node height: a deck
+    // node may hold direct cards and nested zones side by side (custom zones),
+    // and an empty node would previously crash on at(0).
+    QVector<const AbstractDecklistCardNode *> cards;
+    QVector<const InnerDecklistNode *> subZones;
+    for (int i = 0; i < node->size(); i++) {
+        if (auto *card = dynamic_cast<const AbstractDecklistCardNode *>(node->at(i))) {
+            cards.append(card);
+        } else if (auto *zone = dynamic_cast<const InnerDecklistNode *>(node->at(i))) {
+            subZones.append(zone);
+        }
+    }
+
+    if (!cards.isEmpty()) {
         QTextBlockFormat blockFormat;
         QTextCharFormat charFormat;
         charFormat.setFontPointSize(11);
@@ -523,9 +548,9 @@ void DeckLoader::printDeckListNode(QTextCursor *cursor, const InnerDecklistNode 
         tableFormat.setCellPadding(0);
         tableFormat.setCellSpacing(0);
         tableFormat.setBorder(0);
-        QTextTable *table = cursor->insertTable(node->size() + 1, totalColumns, tableFormat);
-        for (int i = 0; i < node->size(); i++) {
-            auto *card = dynamic_cast<AbstractDecklistCardNode *>(node->at(i));
+        QTextTable *table = cursor->insertTable(cards.size() + 1, totalColumns, tableFormat);
+        for (int i = 0; i < cards.size(); i++) {
+            const AbstractDecklistCardNode *card = cards[i];
 
             QTextCharFormat cellCharFormat;
             cellCharFormat.setFontPointSize(9);
@@ -540,7 +565,13 @@ void DeckLoader::printDeckListNode(QTextCursor *cursor, const InnerDecklistNode 
             cellCursor = cell.firstCursorPosition();
             cellCursor.insertText(card->getName());
         }
-    } else if (node->height() == 2) {
+    }
+
+    for (const InnerDecklistNode *subZone : subZones) {
+        if (subZone->isEmpty()) {
+            continue;
+        }
+
         QTextBlockFormat blockFormat;
         QTextCharFormat charFormat;
         charFormat.setFontPointSize(14);
@@ -559,10 +590,8 @@ void DeckLoader::printDeckListNode(QTextCursor *cursor, const InnerDecklistNode 
         tableFormat.setColumnWidthConstraints(constraints);
 
         QTextTable *table = cursor->insertTable(1, totalColumns, tableFormat);
-        for (int i = 0; i < node->size(); i++) {
-            QTextCursor cellCursor = table->cellAt(0, (i * totalColumns) / node->size()).lastCursorPosition();
-            printDeckListNode(&cellCursor, dynamic_cast<InnerDecklistNode *>(node->at(i)));
-        }
+        QTextCursor cellCursor = table->cellAt(0, 0).firstCursorPosition();
+        printDeckListNode(&cellCursor, subZone);
     }
 
     cursor->movePosition(QTextCursor::End);
--- a/cockatrice/src/interface/deck_loader/deck_loader.cpp
+++ b/cockatrice/src/interface/deck_loader/deck_loader.cpp
@@ -375,20 +375,29 @@ void DeckLoader::saveToStream_DeckHeader(QTextStream &out, const DeckList &deckL
 void DeckLoader::saveToStream_DeckZone(QTextStream &out,
                                        const InnerDecklistNode *zoneNode,
                                        bool addComments,
-                                       bool addSetNameAndNumber)
+                                       bool addSetNameAndNumber,
+                                       const QString &boardZoneName)
 {
+    // Nested sub-zones keep their owning board's identity: the top-level call
+    // passes no board, so the zone's own name is used; recursive calls carry the
+    // owning board down so the sideboard marker survives sub-zone nesting.
+    const QString owningBoardZoneName = boardZoneName.isEmpty() ? zoneNode->getName() : boardZoneName;
+
     // group cards by card type and count the subtotals
     QMultiMap<QString, DecklistCardNode *> cardsByType;
     QMap<QString, int> cardTotalByType;
     int cardTotal = 0;
+    QList<const InnerDecklistNode *> subZones;
 
     for (int j = 0; j < zoneNode->size(); j++) {
         auto *card = dynamic_cast<DecklistCardNode *>(zoneNode->at(j));
         if (!card) {
             // Cards collected in nested sub-zones are exported by recursion so
-            // they don't end up invisible in the plain text output.
+            // they don't end up invisible in the plain text output. They are
+            // deferred until after this zone's own header and cards so they read
+            // as part of this zone's block.
             if (auto *subZone = dynamic_cast<const InnerDecklistNode *>(zoneNode->at(j))) {
-                saveToStream_DeckZone(out, subZone, addComments, addSetNameAndNumber);
+                subZones.append(subZone);
             }
             continue;
         }
@@ -419,25 +428,30 @@ void DeckLoader::saveToStream_DeckZone(QTextStream &out,
 
         QList<DecklistCardNode *> cards = cardsByType.values(cardType);
 
-        saveToStream_DeckZoneCards(out, zoneNode, cards, addComments, addSetNameAndNumber);
+        saveToStream_DeckZoneCards(out, cards, addComments, addSetNameAndNumber, owningBoardZoneName);
 
         if (addComments) {
             out << "\n";
         }
     }
+
+    // Nested sub-zones come last, after the parent's own header and cards.
+    for (const auto *subZone : subZones) {
+        saveToStream_DeckZone(out, subZone, addComments, addSetNameAndNumber, owningBoardZoneName);
+    }
 }
 
 void DeckLoader::saveToStream_DeckZoneCards(QTextStream &out,
-                                            const InnerDecklistNode *zoneNode,
                                             QList<DecklistCardNode *> cards,
                                             bool addComments,
-                                            bool addSetNameAndNumber)
+                                            bool addSetNameAndNumber,
+                                            const QString &boardZoneName)
 {
     // QMultiMap sorts values in reverse order
     for (int i = cards.size() - 1; i >= 0; --i) {
         DecklistCardNode *card = cards[i];
 
-        if (zoneNode->getName() == DECK_ZONE_SIDE && addComments) {
+        if (boardZoneName == DECK_ZONE_SIDE && addComments) {
             out << "SB: ";
         }
 
--- a/cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
@@ -91,7 +91,7 @@ void CardDatabaseView::decrementCard(const QString &zoneName)
 }
 
 void CardDatabaseView::setZoneMenuProvider(const std::function<QList<QPair<QString, QStringList>>()> &provider,
-                                           const std::function<void()> &newZoneHandler)
+                                           const std::function<QString()> &newZoneHandler)
 {
     zoneMenuProvider = provider;
     this->newZoneHandler = newZoneHandler;
@@ -151,35 +151,46 @@ void CardDatabaseView::openCustomMenu(QPoint point)
 
     if (zoneMenuProvider) {
         QMenu *addToZoneMenu = menu.addMenu(tr("Add to Zone"));
-        for (const QString &boardName : InnerDecklistNode::boardZoneNames()) {
-            QAction *action = addToZoneMenu->addAction(InnerDecklistNode::visibleNameFromName(boardName));
-            connect(action, &QAction::triggered, this,
-                    [this, card, boardName] { emit cardAdded(card->getName(), boardName); });
-        }
-
-        bool anyCustomZone = false;
         const auto zoneBoards = zoneMenuProvider();
-        for (const auto &zoneBoard : zoneBoards) {
-            const QString &boardName = zoneBoard.first;
-            const QStringList &customZones = zoneBoard.second;
+        for (const QString &boardName : InnerDecklistNode::boardZoneNames()) {
+            // Boards with zones nest their children so no two menu entries
+            // share a visible name: "Maindeck ▸ { Maindeck (whole board), … }".
+            const QStringList customZones = [&zoneBoards, boardName] {
+                for (const auto &zoneBoard : zoneBoards) {
+                    if (zoneBoard.first == boardName) {
+                        return zoneBoard.second;
+                    }
+                }
+                return QStringList();
+            }();
             if (customZones.isEmpty()) {
-                continue;
-            }
-            anyCustomZone = true;
-            QMenu *boardSubmenu = addToZoneMenu->addMenu(InnerDecklistNode::visibleNameFromName(boardName));
-            for (const QString &zoneName : customZones) {
-                QAction *action = boardSubmenu->addAction(zoneName);
+                QAction *action = addToZoneMenu->addAction(InnerDecklistNode::visibleNameFromName(boardName));
                 connect(action, &QAction::triggered, this,
-                        [this, card, zoneName] { emit cardAdded(card->getName(), zoneName); });
+                        [this, card, boardName] { emit cardAdded(card->getName(), boardName); });
+            } else {
+                QMenu *boardSubmenu = addToZoneMenu->addMenu(InnerDecklistNode::visibleNameFromName(boardName));
+                QAction *wholeBoardAction = boardSubmenu->addAction(InnerDecklistNode::visibleNameFromName(boardName));
+                connect(wholeBoardAction, &QAction::triggered, this,
+                        [this, card, boardName] { emit cardAdded(card->getName(), boardName); });
+                for (const QString &zoneName : customZones) {
+                    QAction *action = boardSubmenu->addAction(zoneName);
+                    connect(action, &QAction::triggered, this,
+                            [this, card, zoneName] { emit cardAdded(card->getName(), zoneName); });
+                }
             }
         }
 
-        if (anyCustomZone) {
+        if (newZoneHandler) {
             addToZoneMenu->addSeparator();
-        }
 
-        QAction *newZoneAction = addToZoneMenu->addAction(tr("Create &new zone..."));
-        connect(newZoneAction, &QAction::triggered, this, [this] { newZoneHandler(); });
+            QAction *newZoneAction = addToZoneMenu->addAction(tr("Create &new zone..."));
+            connect(newZoneAction, &QAction::triggered, this, [this, card] {
+                const QString zoneName = newZoneHandler();
+                if (!zoneName.isEmpty()) {
+                    emit cardAdded(card->getName(), zoneName);
+                }
+            });
+        }
     }
 
     if (canBeCommander(*card)) {
--- a/cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
+++ b/cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
@@ -90,6 +90,13 @@ void CardDatabaseView::decrementCard(const QString &zoneName)
     emit cardDecremented(currentCardName(), zoneName);
 }
 
+void CardDatabaseView::setZoneMenuProvider(const std::function<QList<QPair<QString, QStringList>>()> &provider,
+                                           const std::function<void()> &newZoneHandler)
+{
+    zoneMenuProvider = provider;
+    this->newZoneHandler = newZoneHandler;
+}
+
 void CardDatabaseView::updateCard(const QModelIndex &current, const QModelIndex & /*previous*/)
 {
     if (!current.isValid()) {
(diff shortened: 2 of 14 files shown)


cockatrice/src/interface/widgets/visual_database_display/visual_database_display_widget.cpp
```diff
@@ -89,6 +89,20 @@ VisualDatabaseDisplayWidget::VisualDatabaseDisplayWidget(QWidget *parent,
     databaseView->setItemDelegate(nullptr);
     databaseView->setVisible(false);
 
+    // Without a deck model there is nothing to add cards to, so the zone menu stays hidden.
+    if (deckListModel) {
+        databaseView->setZoneMenuProvider(
+            [this, deckListModel]() -> QList<QPair<QString, QStringList>> {
```
RickyRister: Don't need to capture `this`

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,143 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+
     QMenu menu;
 
+    const bool isCustomZoneRow = sourceIndex.isValid() && sourceIndex.data(DeckRoles::IsCustomZoneRole).toBool();
+    const bool isBoardZoneRow = sourceIndex.isValid() && !isCustomZoneRow && !sourceIndex.parent().isValid();
+    const bool isCardRow =
+        sourceIndex.isValid() && !isCustomZoneRow && !isBoardZoneRow && !getModel()->hasChildren(sourceIndex);
+
+    if (isCardRow) {
+        addMoveToZoneMenu(&menu, sourceIndex);
```
ZeldaZach: The move actions land flat in the top-level context menu, so the result has two entries reading "Maindeck" that do different things.

`addMoveToZoneMenu` is handed `&menu` itself and calls `menu->addAction(...)` directly (line 847), then adds submenus with the *same* board names to that same menu (line 859). Right-clicking a card in the mainboard with one custom zone "Removal" under it gives:

```
Maindeck
Sideboard
Maybeboard
Maindeck        ▸ Removal
─────────
Create &new zone...
─────────
Select Printing
```

Nothing tells the user the top three mean "move this card there", and "Maindeck" appears twice. The commit message says "per-zone submenus", so the flat placement looks unintended.

```suggestion
        addMoveToZoneMenu(menu.addMenu(tr("Move to &zone")), sourceIndex);
```

That fixes the missing heading; the duplicate label still needs the change noted on line 859.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,143 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+
+    const auto tree = deckStateManager->getDeckListShared()->getTree();
+    bool anyCustomZone = false;
+    for (const QString &boardName : InnerDecklistNode::boardZoneNames()) {
+        QList<const InnerDecklistNode *> customZones = tree->getCustomZones(boardName);
+        if (customZones.isEmpty()) {
+            continue;
+        }
+        anyCustomZone = true;
+        QMenu *boardSubmenu = menu->addMenu(InnerDecklistNode::visibleNameFromName(boardName));
```
ZeldaZach: This submenu reuses the same visible name as the plain action added for the same board at line 847, which is where the duplicate label comes from.

Two ways out: nest the board action inside its own submenu (`Maindeck ▸ { Maindeck (whole board), Removal, … }`), or flatten the custom zones to qualified entries like `Maindeck / Removal` and drop the per-board submenu entirely. Either reads unambiguously; the current shape doesn't.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,143 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+            QAction *action = boardSubmenu->addAction(customZone->getName());
+            connect(action, &QAction::triggered, this, [moveToZone, customZone] { moveToZone(customZone->getName()); });
+        }
+    }
+
+    if (anyCustomZone) {
+        menu->addSeparator();
+    }
+
+    addNewZoneAction(menu);
```
ZeldaZach: This entry sits inside a menu whose whole purpose is "move this card somewhere", but `addNewZoneAction`'s handler only calls `createCustomZone` — the card the user right-clicked is left where it was.

So: right-click Lightning Bolt → *Create new zone...* → type "Burn" → OK. The zone appears in the tree and Bolt is not in it, with no feedback that the move part didn't happen. Worth either a dedicated create-and-move entry here (`createCustomZone(...)` then `moveToZone(zoneName)`), or moving the plain create action out of the move menu so it doesn't read as one of the move targets.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,143 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+    if (isCardRow) {
+        addMoveToZoneMenu(&menu, sourceIndex);
+        menu.addSeparator();
+    } else if (isCustomZoneRow) {
+        const QString zoneName =
+            sourceIndex.siblingAtColumn(DeckListModelColumns::CARD_NAME).data(Qt::EditRole).toString();
+
+        QAction *renameAction = menu.addAction(tr("&Rename zone..."));
+        connect(renameAction, &QAction::triggered, this, [this, zoneName] {
+            const QString newName = DeckZoneDialog::promptForRename(this, zoneName, [this](const QString &candidate) {
```
ZeldaZach: The rename dialog opens already showing "A zone with this name already exists." with OK greyed out.

`promptForRename` prefills the current name via `setZoneName`, which fires `textChanged` → `validateName()`. Since the zone is by definition already in the deck, `validateNewZoneName(currentName)` hits the deck-wide uniqueness loop at `deck_state_manager.cpp:481-487` and returns that error. Renaming "Removal" therefore presents the user's own zone as invalid until they edit the field.

`DeckZoneDialog`'s header comment already flags that callers must special-case the unchanged name, and this call site is the one that needs it:

```suggestion
            const QString newName = DeckZoneDialog::promptForRename(this, zoneName, [this, zoneName](const QString &candidate) {
                return candidate == zoneName ? QString() : deckStateManager->validateNewZoneName(candidate);
            });
```

The existing `newName != zoneName` guard on line 799 already makes an unchanged name a no-op, so nothing else needs to move.

cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
```diff
@@ -142,6 +149,39 @@ void CardDatabaseView::openCustomMenu(QPoint point)
...
+        bool anyCustomZone = false;
+        const auto zoneBoards = zoneMenuProvider();
+        for (const auto &zoneBoard : zoneBoards) {
+            const QString &boardName = zoneBoard.first;
+            const QStringList &customZones = zoneBoard.second;
+            if (customZones.isEmpty()) {
+                continue;
+            }
+            anyCustomZone = true;
+            QMenu *boardSubmenu = addToZoneMenu->addMenu(InnerDecklistNode::visibleNameFromName(boardName));
```
ZeldaZach: Same duplicate-label problem as the deck dock menu: line 155 adds a plain action per board, and this adds a submenu with the identical text for any board that has custom zones. A user with one zone under the sideboard sees `Add to Zone ▸ { Maindeck, Sideboard, Maybeboard, Sideboard ▸ }`.

The three plain board entries also restate *Add to Deck* / *Add to Sideboard* from lines 143-144 of the same menu. Nesting each board's own entry inside its submenu, or dropping the plain board rows and listing only custom zones, would resolve both.

cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
```diff
@@ -142,6 +149,39 @@ void CardDatabaseView::openCustomMenu(QPoint point)
...
+                        [this, card, zoneName] { emit cardAdded(card->getName(), zoneName); });
+            }
+        }
+
+        if (anyCustomZone) {
+            addToZoneMenu->addSeparator();
+        }
+
+        QAction *newZoneAction = addToZoneMenu->addAction(tr("Create &new zone..."));
+        connect(newZoneAction, &QAction::triggered, this, [this] { newZoneHandler(); });
```
ZeldaZach: Same issue as the deck dock's copy of this action: it's inside *Add to Zone*, but the handler only creates the zone — the card is never added. `Add to Zone ▸ Create new zone...` → "Burn" → OK leaves the new zone empty.

Making `newZoneHandler` a `std::function<QString()>` that returns the created zone name would let this do `emit cardAdded(card->getName(), createdZone)` afterwards.

Also, `newZoneHandler` is called unconditionally while the submenu is gated on `zoneMenuProvider` alone — a caller that sets a provider but leaves the handler default-constructed gets `std::bad_function_call` → `std::terminate`. Both current call sites pass a real handler, so it's latent, but the two should be gated together.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,143 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+            });
+            if (!newName.isEmpty() && newName != zoneName) {
+                deckStateManager->renameCustomZone(zoneName, newName);
+            }
+        });
+
+        QMenu *boardMenu = menu.addMenu(tr("Change &board"));
+        addChangeBoardMenu(boardMenu, zoneName);
+
+        QAction *deleteAction = menu.addAction(tr("&Delete zone"));
```
ZeldaZach, line 823: Nit: this hint never reaches the user, and it's set even when Delete is enabled.

`menu` is a parentless `QMenu` (line 780), so the `QStatusTipEvent` stops at the popup's own top-level window; `AbstractTabDeckEditor` derives from `Tab`, not `QMainWindow`, so there's no status bar to receive it anyway. The user just gets a greyed-out "Delete zone" with no explanation. `setToolTip(...)` plus `menu.setToolTipsVisible(true)`, set only in the disabled case, would actually show it.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,143 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
 void DeckEditorDeckDockWidget::decklistCustomMenu(QPoint point)
 {
+    const QModelIndex sourceIndex = proxy->mapToSource(deckView->indexAt(point));
+
     QMenu menu;
 
+    const bool isCustomZoneRow = sourceIndex.isValid() && sourceIndex.data(DeckRoles::IsCustomZoneRole).toBool();
+    const bool isBoardZoneRow = sourceIndex.isValid() && !isCustomZoneRow && !sourceIndex.parent().isValid();
+    const bool isCardRow =
+        sourceIndex.isValid() && !isCustomZoneRow && !isBoardZoneRow && !getModel()->hasChildren(sourceIndex);
```
ZeldaZach, line 785: Nit: `isCardRow` doesn't exclude cards under Tokens, but `moveCardToZone` bails out for tokens (`deck_state_manager.cpp:331-335`), so right-clicking a token offers three move targets that all silently do nothing. The entry matching the card's *current* board has the same problem (`currentZoneName == targetZoneName` → `return false`).

`addChangeBoardMenu` already marks the current board rather than offering it — worth the same treatment here, plus skipping the menu entirely for tokens.

Separately: the move acts on the single index under the cursor, while `actRemoveCard`, `actDecrementSelection`, and `actSwapCard` all iterate `getSelectedCardNodeSourceIndices()`. Selecting five cards and moving them to a zone moves one. Fine as a follow-up, but it's inconsistent with the rest of the dock.

ZeldaZach (changes requested):

The wiring is structurally sound — I couldn't find a crash, leak, or dangling connection in the new code. Every `QMenu`/`QAction` is parented to the stack `QMenu` (`QMenu::addMenu(QString)` parents the submenu to its caller), all `connect`s pass `this` as context, `createCentralFrame()`/`createDatabaseDisplayDock()` each run once from a constructor so nothing is connected twice, and the `DeckListModel*` captured in `visual_database_display_widget.cpp` stays valid because `replaceDeck()` calls `setDeckList()` rather than rebuilding the model. The `DeckStateManager` calls match the signatures from #7204, and `getZoneObjFromName` really does resolve custom-zone names, so `cardAdded(name, customZoneName)` lands in the right place. Strings are all `tr()`-wrapped.

What I did find is a cluster of menu-construction problems. None corrupt data, but two produce menus with two identically-labelled entries, one silently discards the card the user was acting on, and one opens the rename dialog in an error state. Inline below.

### Blocker outside this diff: plain-text export and print crash on any deck with a custom zone

This is the PR that first lets a user create a custom zone, so it's the one that makes two pre-existing latent bugs reachable. Across the codebase only four sites cast board-zone children to card nodes; #7203 fixes `deck_list_model.cpp:67` and #7207 fixes `deck_view.cpp:385`, but these two are untouched:

- **`deck_loader.cpp:386-388`** — `DeckLoader::saveToStream_DeckZone` iterates the direct children of each zone from `getZoneNodes()` and dereferences the `dynamic_cast<DecklistCardNode *>` result with no null check. Create a zone under the mainboard, put a card in it, then use any *Save Deck to Clipboard* action (`abstract_tab_deck_editor.cpp:456-474`) or save as `.dec`/`.txt` → null deref. Even guarded, the zone's cards would be dropped from the export. `deckList.getCardNodes({zoneNode->getName()})` already does the recursion correctly.
- **`deck_loader.cpp:515-566`** — `printDeckListNode` branches once on `node->height()` and then assumes every child has the matching type. `InnerDecklistNode::height()` is `at(0)->height() + 1`, decided entirely by the first child, so a board holding both loose cards and a custom zone is heterogeneous either way: first-child-is-a-card gives `height() == 1` and line 528's cast yields null, first-child-is-the-zone gives `height() == 2` and line 564 passes null into the recursive call. *File → Print Deck* crashes. Separately, `height()` calls `at(0)` unguarded, so an **empty** custom zone — which this stack deliberately keeps alive — indexes an empty `QList`.

Both want per-child dispatch rather than a single `height()` branch, and `InnerDecklistNode::height()` should return early for an empty node. Happy for that to be a follow-up PR in the stack rather than more scope here, but it shouldn't ship after this one.

CI: the Debian 13 / Fedora 44 / Ubuntu 26.04 failures are inherited from #7203, whose two new test files have no `main()`.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Wrapped the whole move block in a `Move to &zone` submenu and dropped the flat per-board actions. Boards with zones now nest as `Boardname ▸ { Boardname (whole board), … }`, so no two labels repeat and every entry reads as a move target. In 7654faabc1.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Applied the first option: the boards own entry now lives inside its submenu as `Maindeck (whole board)` next to the zone entries, so no two labels collide.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Added a dedicated `Create new zone and move &here...` entry in the move menu: it creates the zone via the dialog using the cards current board as default, then moves the right-clicked card into it (`moveCardToZone` after `createNewCustomZone`). The plain create action remains only on board/empty-area menus.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Special-cased the unchanged name in the validator: `candidate == zoneName` returns no error, so the dialog opens clean and the existing `newName != zoneName` guard still makes an unchanged name a no-op.

cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
BruebachL: Restructured to the nested shape: boards without custom zones show one flat entry; boards with zones get `Boardname ▸ { Boardname (whole board), … }`. No repeated labels, and the zone-bearing boards no longer restate the Add to Deck / Add to Sideboard rows.

cockatrice/src/interface/widgets/deck_editor/card_database_view.cpp
BruebachL: `newZoneHandler` is now `std::function<QString()>` returning the created zones name; the action emits `cardAdded(card->getName(), createdZone)` when non-empty. The Create entry is gated on the handler being set, so a provider-without-handler can no longer terminate.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Switched to `setToolTip(tr("Move or remove all cards first."))` plus `menu.setToolTipsVisible(true)`, set only when the zone is non-empty (Delete disabled).

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: The move menu is now skipped entirely for cards under Tokens, and the current boards whole-board entry is disabled instead of offered, matching `addChangeBoardMenu`. On the multi-selection inconsistency: acknowledged — it moves the one index under the cursor; Ill switch it to iterate `getSelectedCardNodeSourceIndices()` in a follow-up.

BruebachL:
Thanks for the thorough review — all inline threads are addressed (see per-thread replies).

The **Blocker** is fixed in this PR rather than a follow-up in the stack:
- `DeckLoader::saveToStream_DeckZone` now recurses into nested sub-zones, so a custom zone under the mainboard reaches the plain-text export (clipboard / .dec / .txt) instead of null-dereffing on the board cast and silently dropping the zone cards.
- `DeckLoader::printDeckListNode` dispatches children by type — one table for direct cards, one nested table per non-empty sub-zone — instead of branching once on `height()` and assuming uniform children.
- `InnerDecklistNode::height()` returns early for an empty node instead of `at(0)`-ing an empty list.

Verified: `cockatrice` builds, full ctest passes 25/25, format clean. Fix is pushed as `7654faabc1` on `custom-zones/b2-editor-zone-menus`.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,183 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+        }
+    }
+
+    moveMenu->addSeparator();
+
+    QAction *newZoneAction = moveMenu->addAction(tr("Create new zone and move &here..."));
+    connect(newZoneAction, &QAction::triggered, this, [this, sourceCardIndex, currentBoardName] {
+        const QString zoneName = createNewCustomZone(currentBoardName);
+        if (!zoneName.isEmpty()) {
+            deckStateManager->moveCardToZone(sourceCardIndex, zoneName);
```
ZeldaZach: Use-after-free: `sourceCardIndex` is dead by the time it gets here.

`createNewCustomZone` -> `DeckStateManager::createCustomZone` -> `modifyTree` -> `deckListModel->rebuildTree()`, and `rebuildTree` opens with `beginResetModel()` then `root->clearTree()`, which `delete`s every node in the mirror tree. `sourceCardIndex` is a plain `QModelIndex` captured by value, so it is not cleared by the reset and still carries an `internalPointer` to a freed `DecklistModelCardNode`. `moveCardToZone` then does `idx.data(DeckRoles::IsCardRole)`, and `DeckListModel::data` does `static_cast<AbstractDecklistNode *>(index.internalPointer())` on that freed node.

So the exact flow the previous review asked for (`Create new zone and move here...` -> name it -> OK) is the one that reads freed memory. In practice it will usually move whatever card now happens to occupy that row, or crash.

Resolve the card before creating the zone and re-find it afterwards:

```suggestion
        const QString cardName =
            sourceCardIndex.siblingAtColumn(DeckListModelColumns::CARD_NAME).data(Qt::EditRole).toString();
        const QString providerId =
            sourceCardIndex.siblingAtColumn(DeckListModelColumns::CARD_PROVIDER_ID).data().toString();
        const QString collectorNumber =
            sourceCardIndex.siblingAtColumn(DeckListModelColumns::CARD_NUMBER).data().toString();
        const QString sourceZoneName = currentZoneName;

        const QString zoneName = createNewCustomZone(currentBoardName);
        if (!zoneName.isEmpty()) {
            const QModelIndex refreshed =
                getModel()->findCard(cardName, sourceZoneName, providerId, collectorNumber);
            if (refreshed.isValid()) {
                deckStateManager->moveCardToZone(refreshed, zoneName);
            }
        }
```

(`DeckStateManager::decrementCard` already uses exactly this re-find pattern with the comment "old index is no longer safe since rows could have been removed".) Adjust the column enum names to whatever the model actually exposes; the shape is the point.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
```diff
@@ -772,14 +775,183 @@ void DeckEditorDeckDockWidget::offsetCountAtIndex(const QModelIndex &idx, bool i
...
+    for (const QString &boardName : InnerDecklistNode::boardZoneNames()) {
+        const QString boardLabel = InnerDecklistNode::visibleNameFromName(boardName);
+        const auto customZones = tree->getCustomZones(boardName);
+
+        // Boards with zones nest their children so no two menu entries share a
+        // visible name: "Maindeck ▸ { Maindeck (whole board), Removal, … }".
+        // The board the card already lives on is marked instead of offered.
+        if (!customZones.isEmpty()) {
+            QMenu *boardSubmenu = moveMenu->addMenu(boardLabel);
+            addMoveAction(boardSubmenu, boardName, boardLabel, boardName != currentBoardName);
```
ZeldaZach: The enabled/disabled test compares the *board*, but `moveCardToZone` compares the *zone*, so a card sitting inside a custom zone gets the wrong menu in both directions.

`moveCardToZone` (`deck_state_manager.cpp:339-350`) walks up to the nearest `IsCustomZoneRole` ancestor, so for a card in `Removal` under the maindeck its `currentZoneName` is `"Removal"`, not `"main"`. Meanwhile `currentBoardName` here is `"main"`, which means:

- **Maindeck (whole board) is disabled**, even though moving the card out of `Removal` back to the maindeck root is a perfectly valid move and `moveCardToZone` would accept it. With one custom zone per board there is then no menu path at all for taking a card out of a zone and leaving it on the same board.
- **Removal itself is enabled** (line 890 hardcodes `true`), and picking it hits the `currentZoneName == targetZoneName` early return and silently does nothing. That is the same dead-entry problem the previous round flagged, just moved down a level.

Computing the card's current zone the same way `moveCardToZone` does, and comparing against that instead of the board, fixes both:

```suggestion
            addMoveAction(boardSubmenu, boardName, boardLabel, boardName != currentZoneName);
            for (const auto *customZone : customZones) {
                addMoveAction(boardSubmenu, customZone->getName(), customZone->getName(),
                              customZone->getName() != currentZoneName);
            }
```

with `currentZoneName` derived in `decklistCustomMenu` by the same ancestor walk (nearest custom-zone ancestor, else the top-level board). Line 893's flat branch wants the same substitution.

cockatrice/src/interface/deck_loader/deck_loader.cpp
```diff
@@ -384,6 +384,14 @@ void DeckLoader::saveToStream_DeckZone(QTextStream &out,
 
     for (int j = 0; j < zoneNode->size(); j++) {
         auto *card = dynamic_cast<DecklistCardNode *>(zoneNode->at(j));
+        if (!card) {
+            // Cards collected in nested sub-zones are exported by recursion so
+            // they don't end up invisible in the plain text output.
+            if (auto *subZone = dynamic_cast<const InnerDecklistNode *>(zoneNode->at(j))) {
+                saveToStream_DeckZone(out, subZone, addComments, addSetNameAndNumber);
```
ZeldaZach: The recursion fixes the null deref, but it loses the sideboard marker, so a deck with a custom zone under the sideboard no longer round-trips.

`saveToStream_DeckZoneCards` decides the prefix with `zoneNode->getName() == DECK_ZONE_SIDE` (line 439). On the recursive call `zoneNode` is the sub-zone, whose name is the user's zone name, so cards in `Sideboard > Removal` are written without `SB: `. Re-importing that file puts them in the maindeck. Passing the owning board down (or the already-resolved prefix) keeps it intact:

```suggestion
                saveToStream_DeckZone(out, subZone, addComments, addSetNameAndNumber, zoneNode->getName());
```

with the extra parameter defaulted and threaded into the `DECK_ZONE_SIDE` comparison in `saveToStream_DeckZoneCards`.

Second, smaller point on the same call: it writes straight to `out` from inside the grouping loop, before the parent zone emits its own `// <total> <Zone>` header at line 415. So the output reads

```
// 4 Removal
// 4 Instant
4 Lightning Bolt

// 24 Maindeck
...
```

and the zone block looks like it belongs to whatever zone was printed before it. Collecting sub-zones during the loop and recursing after the parent's own cards are written puts them under the right heading.

ZeldaZach (changes requested):
Re-review. All eight inline threads from the previous round are genuinely addressed: the move menu is now nested under `Move to &zone` with no repeated labels, the rename validator special-cases the unchanged name, the delete hint is a real tooltip, the token board is skipped, `newZoneHandler` returns the created name and is gated together with the provider, and the export/print blocker was fixed in this PR instead of deferred. CI is green across all platforms now.

The new create-and-move action introduces a use-after-free, though, and the fix for the "current board is marked instead of offered" point overshoots. Three things below.

Minor, not worth a thread each:
- `createNewCustomZone` returns the dialog's name even when `deckStateManager->createCustomZone(...)` returned false, so a failed creation still feeds a nonexistent zone name to `moveCardToZone` / `cardAdded`. Returning `{}` unless the create succeeded closes it. Same in `DeckEditorCardDatabaseDockWidget`'s `newZoneHandler` and `TabDeckEditorVisual::createNewZone`.
- Custom zone names go into `QAction` text verbatim, so a zone named `Red & Black` renders with a mnemonic and shows as `Red _Black`. `QString(name).replace('&', "&&")` for the label only.
- After this PR `InnerDecklistNode::height()` has no callers left in the tree (the two `deck_loader.cpp` call sites were its only ones). The empty-node guard is still correct, just noting the function is now dead.

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Fixed in 3f1c806d6a: the handler resolves `cardName`/`CARD_PROVIDER_ID`/`CARD_COLLECTOR_NUMBER` before `createNewCustomZone` rebuilds the model, then re-finds the card in its source zone via `getModel()->findCard` and moves the refreshed index (mirroring the `decrementCard` re-find pattern).

cockatrice/src/interface/widgets/deck_editor/deck_editor_deck_dock_widget.cpp
BruebachL: Fixed in 3f1c806d6a: `addMoveToZoneMenu` now derives `currentZoneName` with the same ancestor walk as `DeckStateManager::moveCardToZone` (nearest custom-zone ancestor, else the top-level board) and compares against that — so moving a card out of a custom zone back to the board root is offered, the card's own zone is disabled, and the flat boards use the zone comparison too.

cockatrice/src/interface/deck_loader/deck_loader.cpp
BruebachL: Fixed in 783de987f2: `saveToStream_DeckZone` threads the owning board zone name down to the card writer (recursive calls pass the top-level board, defaulting to the zone's own name at the top level), so `Sideboard > Removal` cards keep their `SB: ` prefix. Nested sub-zones are also collected during the loop and written after the parent's own header and cards, so they no longer read as part of the zone printed before them.

ZeldaZach (approved):

PR #7206, opened 2026-08-29
[Client] Show custom zones in the card display widgets

BruebachL:
## Related Ticket(s)
- Stacked on custom-zones/b2-editor-zone-menus (#7205)

- Render custom zones next to the criteria groups in the deck display widgets (`CardGroupDisplayWidget`, `DeckCardZoneDisplayWidget`)
- Apply the visual sort criteria inside custom zones through the same `DeckListSortFilterProxyModel` path as criteria groups — no custom-zone special case
- Treat the model row order as the display order for the top-level zone children (criteria groups first, then custom zones)

## Why
- The visual deck editor should draw user-defined zones, and sorting should behave consistently everywhere instead of silently doing nothing inside a custom zone

## Verified

### DIFF
--- a/cockatrice/src/interface/widgets/cards/card_group_display_widgets/card_group_display_widget.cpp
+++ b/cockatrice/src/interface/widgets/cards/card_group_display_widgets/card_group_display_widget.cpp
@@ -155,18 +155,6 @@ QWidget *CardGroupDisplayWidget::constructWidgetForIndex(QPersistentModelIndex i
 
 void CardGroupDisplayWidget::updateCardDisplays()
 {
-    // Custom zones are user-defined containers: they display their cards in the same
-    // order as the tree view, i.e. the model row order. Only criteria groups apply
-    // the visual sort criteria.
-    const bool isCustomZone = trackedIndex.data(DeckRoles::IsCustomZoneRole).toBool();
-
-    if (isCustomZone) {
-        for (int i = 0; i < deckListModel->rowCount(trackedIndex); ++i) {
-            addCardWidgets(QPersistentModelIndex(deckListModel->index(i, 0, trackedIndex)));
-        }
-        return;
-    }
-
     DeckListSortFilterProxyModel proxy;
     proxy.setSourceModel(deckListModel);
     proxy.setSortCriteria(activeSortCriteria);
--- a/cockatrice/src/interface/widgets/cards/card_group_display_widgets/card_group_display_widget.cpp
+++ b/cockatrice/src/interface/widgets/cards/card_group_display_widgets/card_group_display_widget.cpp
@@ -155,6 +155,18 @@ QWidget *CardGroupDisplayWidget::constructWidgetForIndex(QPersistentModelIndex i
 
 void CardGroupDisplayWidget::updateCardDisplays()
 {
+    // Custom zones are user-defined containers: they display their cards in the same
+    // order as the tree view, i.e. the model row order. Only criteria groups apply
+    // the visual sort criteria.
+    const bool isCustomZone = trackedIndex.data(DeckRoles::IsCustomZoneRole).toBool();
+
+    if (isCustomZone) {
+        for (int i = 0; i < deckListModel->rowCount(trackedIndex); ++i) {
+            addCardWidgets(QPersistentModelIndex(deckListModel->index(i, 0, trackedIndex)));
+        }
+        return;
+    }
+
     DeckListSortFilterProxyModel proxy;
     proxy.setSourceModel(deckListModel);
     proxy.setSortCriteria(activeSortCriteria);
@@ -174,16 +186,18 @@ void CardGroupDisplayWidget::updateCardDisplays()
         QModelIndex sourceIndex = proxy.mapToSource(proxyIndex);
 
         // 4. persist the source index
-        QPersistentModelIndex persistent(sourceIndex);
+        addCardWidgets(QPersistentModelIndex(sourceIndex));
+    }
+}
 
-        // Get the card amount
-        int cardAmount =
-            sourceIndex.sibling(sourceIndex.row(), DeckListModelColumns::CARD_AMOUNT).data(Qt::EditRole).toInt();
+void CardGroupDisplayWidget::addCardWidgets(const QPersistentModelIndex &persistent)
+{
+    // Get the card amount
+    int cardAmount = persistent.sibling(persistent.row(), DeckListModelColumns::CARD_AMOUNT).data(Qt::EditRole).toInt();
 
-        // Create multiple widgets for the card count
-        for (int copy = 0; copy < cardAmount; ++copy) {
-            addToLayout(constructWidgetForIndex(persistent));
-        }
+    // Create multiple widgets for the card count
+    for (int copy = 0; copy < cardAmount; ++copy) {
+        addToLayout(constructWidgetForIndex(persistent));
     }
 }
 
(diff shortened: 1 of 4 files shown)


cockatrice/src/interface/widgets/cards/card_group_display_widgets/card_group_display_widget.cpp
```diff
@@ -155,6 +155,18 @@ QWidget *CardGroupDisplayWidget::constructWidgetForIndex(QPersistentModelIndex i
 
 void CardGroupDisplayWidget::updateCardDisplays()
 {
+    // Custom zones are user-defined containers: they display their cards in the same
+    // order as the tree view, i.e. the model row order. Only criteria groups apply
+    // the visual sort criteria.
+    const bool isCustomZone = trackedIndex.data(DeckRoles::IsCustomZoneRole).toBool();
+
+    if (isCustomZone) {
```
ZeldaZach: Design question rather than a defect: this early return skips `DeckListSortFilterProxyModel` entirely, so the visual editor's "sort by" dropdown silently does nothing for cards inside a custom zone. A user with a "Removal" zone who switches from name to cmc sees every criteria group reorder while Removal sits still, with no indication why.

There's no technical obstacle — `mirrorCustomZones` only ever puts `DecklistModelCardNode` children under a custom zone, and `DeckListSortFilterProxyModel::lessThan` handles exactly that in its `leftIsCard && rightIsCard` branch, so the shared path would work unchanged.

If row order is genuinely the intent, that's defensible, but it should be stated in the PR description (and ideally surfaced in the UI) rather than being a silent no-op. Otherwise dropping this branch gets you consistent behavior for free.

cockatrice/src/interface/widgets/cards/deck_card_zone_display_widget.cpp
```diff
@@ -126,24 +128,22 @@ void DeckCardZoneDisplayWidget::constructAppropriateWidget(QPersistentModelIndex
...
-    for (int i = 0; i < proxy.rowCount(proxyParent); ++i) {
-        QModelIndex proxyIndex = proxy.index(i, 0, proxyParent);
+    if (!trackedIndex.isValid()) {
+        return;
+    }
 
-        // 3. map back to source
-        QModelIndex sourceIndex = proxy.mapToSource(proxyIndex);
+    // Iterate the direct children of the tracked zone, keeping the tree view's row
+    // order (criteria groups first, then custom zones in their creation order).
```
ZeldaZach: The comment says custom zones come out "in their creation order", but they don't. `DeckListModelCustomZones::boardSort` sorts *all* children first and only then stably splits groups from custom zones (`deck_list_model_custom_zones.cpp:83-105`), so custom zones end up in the model's sort order among themselves — create "Zeta" then "Alpha" and you get Alpha, Zeta.

```suggestion
    // Iterate the direct children of the tracked zone, keeping the tree view's row
    // order (criteria groups first, then custom zones, both in the model's sort order).
```

cockatrice/src/interface/widgets/cards/deck_card_zone_display_widget.cpp
```diff
@@ -126,24 +128,22 @@ void DeckCardZoneDisplayWidget::constructAppropriateWidget(QPersistentModelIndex
...
+    QList<QPersistentModelIndex> rows;
+    for (int i = 0; i < deckListModel->rowCount(trackedIndex); ++i) {
+        rows.append(QPersistentModelIndex(deckListModel->index(i, 0, trackedIndex)));
+    }
 
-        // 4. persist the source index
-        QPersistentModelIndex persistent(sourceIndex);
+    std::stable_partition(rows.begin(), rows.end(), [](const QPersistentModelIndex &row) {
```
ZeldaZach: This partition is a no-op on every path that reaches `displayCards()`. `trackedIndex` here is always a board zone (`VisualDeckEditorWidget::constructZoneWidgetsFromDeckListModel` only iterates root children), and #7203 already guarantees custom-zones-last for board zones via `sortWithCustomZonesLast` — applied in `sortShadowTree` before `endResetModel` and in `sortHelper` on every `sort()`.

It also doesn't hold the invariant it looks like it's enforcing, because it only runs on the initial build. Adding the first card of a new type calls `createNodeIfNeeded`, which appends the group node and emits `rowsInserted` → `onCategoryAddition` → `constructAppropriateWidget` → `cardGroupLayout->addWidget()`, appending the new group *after* the existing custom-zone widget. So a main deck showing `[Creature, Removal]` becomes `[Creature, Removal, Land]` after adding a land, until the next model reset.

Either drop the partition and treat model order as the source of truth, or fix it properly by having `constructAppropriateWidget` insert at the model row instead of appending — which would also fix the same pre-existing drift for criteria groups.

ZeldaZach (approved):

This one is a clean refactor and I'd merge it once its base is sorted out. The 35 deleted lines are fully accounted for — the `onClick` slot, a `QSortFilterProxyModel` used only for name-sorting, and the loop body extracted into `addCardWidgets` — with no dropped null checks, no leaks, no iteration-while-mutating, and no dangling connections.

Things I specifically checked and found fine:

- Removing `DeckCardZoneDisplayWidget::onClick` is safe. The only other `onClick` in the tree is `VisualDatabaseDisplayWidget`'s unrelated slot, and the archidekt consumer (`archidekt_api_response_deck_display_widget.cpp:196`) never connects `cardClicked`.
- The `routeCardClick` lambda can't go stale on a zone rename. `renameCustomZone` → `modifyTree` → `rebuildTree()` → `modelReset` → `decklistModelReset()` tears down and rebuilds every zone widget, and `this` as context object means the connection dies with the widget.
- `IsCustomZoneRole` is used correctly, and all three new call sites pass source indices rather than `DeckListStyleProxy` indices, so there's no role-forwarding hazard.
- `swapCardAtIndex` still resolves the board correctly for cards in a custom zone — `idx.parent().parent()` works because a custom zone sits at the same depth as a criteria group.
- `updateCardDisplays()` needs no extra validity guard: an invalid `trackedIndex` yields an empty `QVariant`, `toBool()` is `false`, and it falls into the pre-existing proxy path.

Three notes inline, one of which is a design question worth answering in the PR body rather than the code.

Also worth calling out in the description: dropping the hardcoded `sort(CARD_NAME, AscendingOrder)` proxy means the visual editor's group order now follows the deck tree's sort column instead of always being alphabetical. With the defaults (`lastKnownColumn = 1`, ascending) that's identical to before, but sorting the deck list by Count now reorders the visual editor too. Looks intentional given the comment — just undocumented.

CI: the three red Linux jobs are inherited from #7203, whose new test files have no `main()`.

cockatrice/src/interface/widgets/cards/card_group_display_widgets/card_group_display_widget.cpp
BruebachL: Dropped the early return — the visual sort criteria now apply inside custom zones through the shared `DeckListSortFilterProxyModel` path, consistent with criteria groups. No special-casing remains; also stated this in the PR description.

cockatrice/src/interface/widgets/cards/deck_card_zone_display_widget.cpp
BruebachL: Rewrote the comment to describe the tree view row order (criteria groups first, then custom zones, both in the models sort order).

cockatrice/src/interface/widgets/cards/deck_card_zone_display_widget.cpp
BruebachL: Dropped the partition and treat the model row order as the source of truth. Acknowledged the pre-existing drift where a newly added group appends after existing zone widgets until the next model reset — that falls out of `constructAppropriateWidget` appending rather than inserting at the model row; leaving it for a separate fix.

PR #7207, opened 2026-08-29
[Game] Render custom deck zones in the deck view

BruebachL:
## Related Ticket(s)
- Stacked on custom-zones/c1-display-widgets (#7206)

- Render custom deck zones in the in-game deck view
- Collect the zone's cards through the canonical `DeckList::getCardNodes` traversal, so the deck view can't drift from how the server and other consumers resolve custom zones

## Why
- Lets players actually see and interact with custom zones in a live game, with cards recursively collected under their top-level zone so sideboard plans keep working

## Verified

### DIFF
--- a/cockatrice/src/game_graphics/deckview/deck_view.cpp
+++ b/cockatrice/src/game_graphics/deckview/deck_view.cpp
@@ -8,10 +8,8 @@
 #include <QMouseEvent>
 #include <QtMath>
 #include <algorithm>
-#include <functional>
 #include <libcockatrice/card/card_info.h>
 #include <libcockatrice/deck_list/deck_list.h>
-#include <libcockatrice/deck_list/tree/deck_list_card_node.h>
 #include <libcockatrice/settings/cards_display_settings.h>
 
 DeckViewCardDragItem::DeckViewCardDragItem(DeckViewCard *_item,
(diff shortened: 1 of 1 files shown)


cockatrice/src/game_graphics/deckview/deck_view.cpp
```diff
@@ -381,18 +382,25 @@ void DeckViewScene::rebuildTree()
...
-                auto *newCard = new DeckViewCard(container, currentCard->toCardRef(), currentZone->getName());
-                container->addCard(newCard);
-                emit newCardAdded(newCard);
+        // Cards in custom zones nested under a board are regular board cards in-game.
+        // They are collected recursively and reported with the top-level board zone
+        // as their origin, so that sideboard plans keep working.
+        std::function<void(const InnerDecklistNode *)> addZoneCards = [&addZoneCards, container, currentZone,
+                                                                       this](const InnerDecklistNode *zone) {
+            for (int j = 0; j < zone->size(); j++) {
+                auto *currentCard = dynamic_cast<DecklistCardNode *>(zone->at(j));
```
ZeldaZach: Optional: `DeckList::getCardNodes(const QSet<QString> &restrictToZones)` (`deck_list.cpp:483`, implemented at `deck_list_node_tree.cpp:42-62`) already does exactly this recursive collection restricted to a top-level zone, and it's what the server and every other consumer uses. The whole lambda plus the `<functional>` include collapses to:

```cpp
for (auto *currentCard : deck->getCardNodes({currentZone->getName()})) {
    for (int k = 0; k < currentCard->getNumber(); ++k) {
        auto *newCard = new DeckViewCard(container, currentCard->toCardRef(), currentZone->getName());
        container->addCard(newCard);
        emit newCardAdded(newCard);
    }
}
```

I checked it's a true drop-in: `getNumber()` and `toCardRef()` are both const, the traversal order is the same DFS, and `getZoneNodes`'s filter matches on the top-level zone name. Worth doing mainly so the deck view can't drift from the canonical traversal later — not a correctness problem as written, so take it or leave it.

ZeldaZach (approved):
Reviewed as the tip of the custom-zones stack (#7203 → #7204 → #7205 → #7206 → #7207).

Correct, and the interesting part is the semantic choice rather than the 19 lines. Flattening custom-zone cards into their **top-level board** — rather than giving each zone its own `DeckViewCardContainer` — is the right call, and I checked it holds end to end:

- `Server_Player::setupZones` deals from `getCardNodes({DECK_ZONE_MAIN})` / `{DECK_ZONE_SIDE}` (`server_player.cpp:116-117`), and `getCardNodes` already recurses into nested zones, so cards in custom zones really do make it into the game.
- The sideboard plan this view emits only ever names `main`/`side`, which is exactly what `server_player.cpp:120-146` understands. Per-zone containers would have produced `MoveCard_ToZone` messages the server silently drops.

The recursion itself is well-formed, the `std::function` self-capture-by-reference matches the existing idiom in `deck_list_node_tree.cpp:174`, nothing is dropped relative to the removed loop (`if (!currentCard) continue;` becomes `else if (inner) recurse`), and ownership is unchanged since `new DeckViewCard(container, ...)` parents to the container. `menu`-style lifetime concerns don't apply.

I also grepped all of `cockatrice/src/game_graphics/` for other places iterating deck zone children: line 391 is the only one. `utility_menu.cpp:84` already goes through the recursive `getCardNodes({DECK_ZONE_TOKENS})`. No missing siblings in game code.

Approving. One optional simplification inline.

Two crash sites do remain outside game code — `deck_loader.cpp:386` (plain-text export / copy to clipboard) and `deck_loader.cpp:515-566` (print) both still assume every board-zone child is a card. I've written those up in detail on #7205, since that's the PR that first lets a user create a zone; flagging here only so the stack doesn't land without them.

CI: the three red Linux jobs are inherited from #7203, whose new test files have no `main()`.

cockatrice/src/game_graphics/deckview/deck_view.cpp
BruebachL: Done — swapped the lambda for DeckList::getCardNodes({currentZone->getName()}), dropped the now-unused <functional> and deck_list_card_node.h includes. Traversal order and origin reporting are unchanged.

ZeldaZach (approved):

PR #7211, opened 2026-08-29
[Server/Client/Protocol] Add developer staff role

BruebachL:
We're currently more-or-less flying blind whenever there are issues with an update. In order to facilitate debugging and maintenance, it is necessary to install an additional looking glass into the server, which, ideally, we as developers have access to.
Promoting every developer to admin is not feasible.

- Introduce a Developer staff level (proto flag 32, DB admin bit 8) that sits between admin and moderator: no kick/ban/warn/report/admin powers, but gets server log access via a new developer command container family (GET_SERVER_STATS, VIEWLOG_HISTORY) and an idle-timeout exemption. More features/stats to follow in another PR.

- Protocol: IsDeveloper flag, developer_commands.proto envelope, Command_ViewLogHistory now extending both ModeratorCommand and DeveloperCommand (no more duplicated Command_GetLogHistory), Response_GetServerStats, Command_AdjustMod.should_be_developer
- Servatrice: fail-closed developer dispatcher, uptime snapshot handler via Servatrice_DatabaseInterface::getLatestUptimeSnapshot, shared log history handler reuse, bit-8 DB mapping
- Client: burgundy pawn/badge/labels/sort order, prepareDeveloperCommand, minimal Developer stats tab, log tab access, promote/demote actions

## Notes
- Developer log access is narrowed relative to moderators: it excludes private (`chat`) conversations and never exposes `sender_ip`. Operators can still disable log queries entirely via the existing `getEnableLogQuery()` gate (applies to moderators and developers alike).
- `GET_SERVER_STATS` reports the freshest snapshot the server's periodic `statusUpdate()` tick wrote to the `{prefix}_uptime` table, so values are as fresh as the last tick and the handler returns `RespInternalError` on a server that has no snapshot yet (or isn't running in SQL auth mode).


servatrice/src/serversocketinterface.cpp
```diff
@@ -337,6 +340,38 @@ AbstractServerSocketInterface::processExtendedAdminCommand(int cmdType, const Ad
...
+// Permission is checked by processDeveloperCommandContainer. Only stats-style
+// queries live here, never community moderation or server administration.
+Response::ResponseCode AbstractServerSocketInterface::processExtendedDeveloperCommand(int cmdType,
+                                                                                      const DeveloperCommand &cmd,
+                                                                                      ResponseContainer &rc)
+{
+    switch ((DeveloperCommand::DeveloperCommandType)cmdType) {
+        case DeveloperCommand::GET_SERVER_STATS:
+            return cmdGetServerStats(cmd.GetExtension(Command_GetServerStats::ext), rc);
+        case DeveloperCommand::VIEWLOG_HISTORY: {
```
ZeldaZach, line 352: **Worth an explicit decision: this hands the Developer role the most privacy-sensitive capability moderators have.**

The PR describes Developer as sitting between admin and moderator with "no kick/ban/warn/report/admin powers" — which reads as a *reduced* role. But routing `VIEWLOG_HISTORY` here means a Developer can query `getMessageLogHistory`, and looking at what `ServerInfo_ChatMessage` carries, each returned row includes:

- `sender_ip` — the IP address of every user who sent a matching message
- `message` — full text, including `target_type = "chat"`, i.e. private one-to-one conversations
- `sender_name` / `target_name`

And `Command_GetLogHistory` keeps `ip_address` as a query field, so it also supports "show me everything this IP ever said" and "show me everything this user said" as first-class lookups.

Being able to read every user's private messages and resolve them to IP addresses is a bigger privacy grant than being able to kick someone. So the role isn't really "below moderator" — it's orthogonal, and along the axis users would care most about it's *higher*. Someone promoted to Developer to help debug a bad release also silently gains the ability to read the entire server's private chat history.

The stated motivation is diagnosing update problems, and `GET_SERVER_STATS` covers that. So the question is whether log access belongs in this role at all:

- **If it shouldn't:** drop `VIEWLOG_HISTORY` from the developer family and keep this PR to stats. Diagnostics for a bad release rarely need to know what two players said to each other.
- **If it should:** narrow it — e.g. omit `sender_ip` from developer results, or restrict `target_type` to `room`/`game` and exclude `chat`. That covers "what broke in games after the update" without the private-message surface.
- **If it should be unrestricted:** say so plainly in the PR description and in whatever operator docs describe the role, because "Developer" reads as a technical role and admins will grant it on that assumption.

Worth noting the existing `getEnableLogQuery()` gate still applies, so operators can switch the whole thing off — good, but it's all-or-nothing across moderators and developers alike.

This is your call, not mine — I just don't think it should be settled implicitly by which enum the command lands in.

libcockatrice_protocol/libcockatrice/protocol/pb/command_get_log_history.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto2";
+import "developer_commands.proto";
+
+// Developer counterpart of Command_ViewLogHistory: identical query fields, but
+// routed through the developer command family so developers never need the
+// moderator command container.
+message Command_GetLogHistory {
```
ZeldaZach: **This message is a full copy of `Command_ViewLogHistory`, and it has to be kept in sync by hand in two places.**

The fields are duplicated here, then copied field-by-field on the server (`serversocketinterface.cpp:357-366`) and again on the client (`tab_logs.cpp:126-135`). Nothing connects the two definitions, so the day someone adds a field to `Command_ViewLogHistory` — a new filter, a pagination cursor — the moderator path gets it and the developer path silently drops it. No compile error, no test failure, just a filter that quietly stops applying for developers. That's a nasty class of bug to chase.

Protobuf lets one message extend two containers, so the copy isn't necessary. In `moderator_commands.proto`, `Command_ViewLogHistory` can carry both:

```proto
message Command_ViewLogHistory {
    extend ModeratorCommand {
        optional Command_ViewLogHistory ext = 1206;
    }
    extend DeveloperCommand {
        optional Command_ViewLogHistory ext = 1001;
    }
    // ... fields unchanged
}
```

Then `command_get_log_history.proto` goes away, the server dispatcher becomes `return cmdGetLogHistory(cmd.GetExtension(Command_ViewLogHistory::ext), rc);`, and `tab_logs.cpp` just picks which `prepare*Command` to call with the same `cmd` object it already built. Both copy loops disappear, and there's one definition that can't drift.

If you'd rather keep the messages distinct for protocol-documentation reasons, the copies should at least live in one shared helper rather than two.

cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
```diff
@@ -698,4 +717,14 @@ void UserContextMenu::execAdjustJudge(const QString &userName, bool shouldBeJudg
...
+
+void UserContextMenu::execAdjustDeveloper(const QString &userName, bool shouldBeDeveloper)
+{
+    Command_AdjustMod cmd;
+    cmd.set_user_name(userName.toStdString());
+    cmd.set_should_be_developer(shouldBeDeveloper);
+    PendingCommand *pend = client->prepareAdminCommand(cmd);
+    connect(pend, &PendingCommand::finished, this, &UserContextMenu::adjustMod_processUserResponse);
+    client->sendCommand(pend);
 }
```
ZeldaZach: Missing trailing newline (`\ No newline at end of file` in the diff). `desktop-lint` runs clang-format on this repo and will flag it. Only file in the PR with the problem — `tab_developer.cpp`, `tab_developer.h` and `developer_commands.proto` are all fine.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1643,6 +1678,38 @@ Response::ResponseCode AbstractServerSocketInterface::cmdReportUserInfo(const Co
...
+                                                                        ResponseContainer &rc)
+{
+    if (!sqlInterface->checkSql()) {
+        return Response::RespInternalError;
+    }
+
+    // Servatrice::statusUpdate() periodically snapshots server health into the
+    // uptime table. Serve the freshest snapshot for this server.
+    QSqlQuery *query = sqlInterface->prepareQuery(
+        "SELECT users_count, mods_count, games_count, tx_bytes, rx_bytes, uptime, UNIX_TIMESTAMP(timest) "
```
ZeldaZach: Nit, layering: this is the only raw SQL in `serversocketinterface.cpp`. Every other query in servatrice goes through `Servatrice_DatabaseInterface`, which is also where the per-thread `checkSql()` / `prepareQuery()` / `execSqlQuery()` lifecycle is understood. A `getLatestUptimeSnapshot(int serverId)` method over there would keep the socket interface out of the schema and make the query testable in isolation.

Two smaller things in the same handler:

- **`query->value(3).toUInt()`** narrows to 32 bits while the proto field is `uint64`. In practice the source columns are `int(11)` and `tx_bytes`/`rx_bytes` are reset every `statusUpdate()` tick, so nothing can realistically overflow — but `toULongLong()` matches the declared type and costs nothing.
- **Empty result is indistinguishable from real zeros.** If `{prefix}_uptime` has no row for this `id_server` — fresh server, or `statusUpdate()` hasn't ticked yet — `query->next()` is false, no fields get set, and the response is still `RespOk`. `TabDeveloper` then renders every stat as `0` and "Snapshot taken 1970-01-01 00:00", which looks like a broken server rather than an absent snapshot. Either return `RespInternalError`, or leave `timest` unset and have the client show "no snapshot yet".

Also worth a line in the PR description: these numbers come from the DB snapshot table, so they're only as fresh as the last `statusUpdate()` tick, and the handler returns `RespInternalError` on a server not running in SQL auth mode. Both are reasonable, just surprising if you expect live values.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -810,7 +834,10 @@ void TabSupervisor::actTabLog(bool checked)
 
 void TabSupervisor::openTabLog()
 {
-    tabLog = new TabLog(this, client);
+    // Developers without moderation rights query logs through the developer
+    // command family, so tell the tab which family to use.
+    const bool isDeveloper = (userInfo->user_level() & ServerInfo_User::IsDeveloper) != 0;
```
ZeldaZach: Nit: the comment says "Developers without moderation rights query logs through the developer command family", but the flag is just `IsDeveloper` — a user who is both moderator and developer also gets the developer path. That works (both containers reach the same `cmdGetLogHistory`), so it's not a bug, but the comment describes a condition the code doesn't check. Either match the comment:

```cpp
const bool useDeveloperCommands = (userInfo->user_level() & ServerInfo_User::IsDeveloper) &&
                                  !(userInfo->user_level() & ServerInfo_User::IsModerator);
```

or reword it to say developers always use their own family. I'd lean toward the second — one code path is easier to reason about than two, and it means a mod+dev exercises the same path a plain developer does.

libcockatrice_protocol/libcockatrice/protocol/pb/CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ set(PROTO_FILES
     command_del_counter.proto
     command_delete_arrow.proto
     command_draw_cards.proto
+    command_get_log_history.proto
```
ZeldaZach: Nit: `PROTO_FILES` is otherwise alphabetical, and these two land between `command_draw_cards.proto` and `command_dump_zone.proto` — `get_*` sorts after `dump_zone`, not before. Two lines down.

libcockatrice_protocol/libcockatrice/protocol/pb/developer_commands.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto2";
+message DeveloperCommand {
```
ZeldaZach, line 2: Two notes on the new command family, neither blocking:

**Testing.** 510 lines introducing a new privilege level, a new command container, and a new dispatcher, with no tests. I know coverage in this repo is thin generally, but the permission check is exactly the kind of thing worth pinning — a test that a non-developer gets rejected from `processDeveloperCommandContainer`, and that the dispatcher's `default` branch returns `RespFunctionNotAllowed`, would guard against a future refactor quietly opening this up. The bit-mapping in `evalUserQueryResult` is also cheap to test and easy to get wrong.

**Documenting bit 8.** `servatrice.sql` declares `admin tinyint(1)` with no comment about what the bits mean, and this PR makes it four bits (1 = admin+mod, 2 = mod, 4 = judge, 8 = developer). Since operators set these by hand with `UPDATE cockatrice_users SET admin = ...`, a comment on the column — or a note wherever the other levels are documented — would save someone from setting `admin = 8` and wondering why nothing else works.

ZeldaZach (changes requested):
Reviewed against `master` in a local worktree.

First the good news, since this is a permissions change and that's where I looked hardest: **the authorization plumbing is correct.** `processDeveloperCommandContainer` is a byte-for-byte match of the existing admin and moderator container checks — same null-`userInfo` guard, same bit test, same `resetIdleTimer()` placement after both checks, same reverse iteration. The dispatcher is fail-closed (`default: RespFunctionNotAllowed`). The DB bit mapping is right: `admin` is `tinyint(1)`, which is a full signed TINYINT, so bit 8 fits alongside 1/2/4, and the `(u.admin & 7)` → `& 15` widening in `getModeratorLastLogins` matches. Promotion stays an admin-only command, so a Developer can't escalate themselves. Extension number 1220 and response type 1019 are both unclaimed. No holes found.

What I'd like resolved before merge:

1. **A scoping decision on log access** — this is the one that actually matters, details inline. Short version: `VIEWLOG_HISTORY` returns `sender_ip` and private chat contents, which makes "Developer" hard to describe as sitting below Moderator.
2. **The duplicated `Command_GetLogHistory` message**, which is copied field-by-field in two places and will silently drop any field added to `Command_ViewLogHistory` later.
3. **A missing trailing newline** in `user_context_menu.cpp` that `desktop-lint` will almost certainly flag.

The rest is nits. Nice drive-by fix enabling `aPromoteToJudge`/`aDemoteFromJudge` on `anotherUser` — those were missing before.

servatrice/src/serversocketinterface.cpp
BruebachL: Went with narrowing. The developer path now calls `cmdGetLogHistory(..., ·allowPrivateChat = false·)`: the `target_type = "chat"` location is dropped from the query, the `ip_address` filter is ignored, and `sender_ip` is cleared on every returned `ServerInfo_ChatMessage` before it's transmitted. The moderator path is unchanged. `TabLog` also hides the Private Chat checkbox for developers. So Developer keeps game/room diagnostics plus stats without the private-message/IP surface.

libcockatrice_protocol/libcockatrice/protocol/pb/command_get_log_history.proto
BruebachL: Done — one definition, both containers. `Command_ViewLogHistory` now extends both `ModeratorCommand` (`ext`) and `DeveloperCommand` (`dev_ext = 1001`), and `command_get_log_history.proto` is deleted. The server dispatcher is a one-liner (`cmd.GetExtension(Command_ViewLogHistory::dev_ext)`), and `tab_logs.cpp` passes the same `cmd` it already built to whichever `prepare*Command` applies — both copy loops are gone.

One unavoidable deviation from your snippet: protoc rejects two extensions named `ext` in the same message scope (`"ext" is already defined in "Command_ViewLogHistory"`). So the developer extension is `dev_ext`, and the client resolves it by `containing_type() == DeveloperCommand::descriptor()` instead of `FindExtensionByName`. The server reads it by extension number, so the name is irrelevant there.

cockatrice/src/interface/widgets/server/user/user_context_menu.cpp
BruebachL: Fixed — `user_context_menu.cpp` now ends with a trailing newline, committed separately as `60c6f074de`. `clang-format --dry-run --Werror` passes on it.

servatrice/src/serversocketinterface.cpp
BruebachL: All three points addressed. The raw SQL moved into `Servatrice_DatabaseInterface::getLatestUptimeSnapshot(int serverId)`, returning an `UptimeSnapshot` struct with a `valid` flag, and it uses the existing `checkSql()`/`prepareQuery()`/`execSqlQuery()` lifecycle. Counters are read with `toULongLong()` to match the `uint64` proto fields. Empty result is no longer silent: if `{prefix}_uptime` has no row for this server, the handler returns `RespInternalError` and `TabDeveloper` shows "No server statistics available yet." instead of 1970-01-01 zeros.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
BruebachL: Went with your second option (the one you leaned toward): reworded the comment to "Developers query logs through the developer command family" and renamed the local to `useDeveloperCommands`. One code path for everyone with the flag set, mod+developer included.

libcockatrice_protocol/libcockatrice/protocol/pb/CMakeLists.txt
BruebachL: Fixed — `command_get_log_history.proto` no longer exists, and `command_get_server_stats.proto` moved to its correct alphabetical slot between `command_game_say.proto` and `command_inc_card_counter.proto` (after `dump_zone`).

libcockatrice_protocol/libcockatrice/protocol/pb/developer_commands.proto
BruebachL: Both added. New `tests/server_developer_role_test.cpp` pins the permission plumbing: rejects not-logged-in, rejects a plain user, rejects a moderator who is not a developer, dispatches a developer to the developer command path, and asserts the dispatcher`s `default` branch stays `RespFunctionNotAllowed` (fail-closed). And `servatrice.sql` now documents the `admin` bitfield on the column: 1 = admin, 2 = moderator, 4 = judge, 8 = developer.

BruebachL:
All review comments addressed and pushed:

- Log access narrowed (no private chat, no sender_ip) per your "narrow it" option.
- `Command_GetLogHistory` removed; `Command_ViewLogHistory` now extends both command families.
- Uptime snapshot query moved into `Servatrice_DatabaseInterface`, counters widened to `uint64`, empty snapshot returns `RespInternalError`.
- Missing trailing newline fixed.
- Comment reworded, CMake ordering fixed.
- Permission tests added + `admin` bitfield documented in `servatrice.sql`.

PR description updated with the log-scoping and stats-freshness notes.

servatrice/src/serversocketinterface.cpp
```diff
@@ -1040,7 +1063,7 @@ Response::ResponseCode AbstractServerSocketInterface::cmdGetLogHistory(const Com
         if (nameFromStdString(cmd.log_location(i)).simplified() == "game") {
             gameType = true;
         }
-        if (nameFromStdString(cmd.log_location(i)).simplified() == "chat") {
+        if (nameFromStdString(cmd.log_location(i)).simplified() == "chat" && allowPrivateChat) {
```
ZeldaZach, line 1066: The private-chat narrowing is bypassable, so it isn't actually enforced server-side.

Dropping `chat` from the location list only works if at least one other location survives. `getMessageLogHistory` (servatrice_database_interface.cpp:1543) only appends a `target_type` clause when `chat || game || room` is true:

```cpp
if (chat || game || room) {
    queryString.append(" AND (");
    ...
}
```

With all three false the query carries no `target_type` restriction at all and returns every row, private conversations included. So a developer using a crafted client that sends `log_location = ["chat"]` — or omits `log_location` entirely — gets `chatType = false`, `gameType = false`, `roomType = false`, and the server hands back exactly the private messages this branch is meant to withhold. `user_name` alone satisfies the indexed-field requirement at line 1516, so "every private message this user ever sent" is a one-field query. `sender_ip` is still stripped, but message bodies and both names come through.

The stock client can't trigger it (`getClicked()` force-checks room+game for developers), which is precisely why this needs fixing on the server — the guarantee currently lives in the UI.

Cheapest fix here: when `!allowPrivateChat`, never leave the filter empty —

```cpp
if (!allowPrivateChat && !gameType && !roomType) {
    gameType = true;
    roomType = true;
}
```

Cleaner would be an explicit `excludeChat` flag on `getMessageLogHistory` that appends `AND target_type <> 'chat'` unconditionally, so the exclusion doesn't depend on which locations the client happened to ask for.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -810,7 +834,10 @@ void TabSupervisor::actTabLog(bool checked)
 
 void TabSupervisor::openTabLog()
 {
-    tabLog = new TabLog(this, client);
+    // Developers query logs through the developer command family, so tell the
+    // tab which family to use.
+    const bool useDeveloperCommands = (userInfo->user_level() & ServerInfo_User::IsDeveloper) != 0;
```
ZeldaZach: This now demotes moderators who are also developers.

Last round I suggested collapsing to one code path here, on the grounds that both command families reached the same handler and a mod+dev would get identical results either way. That reasoning no longer holds — the same round that made this unconditional also made the developer path strictly weaker (no private chat, no `sender_ip`, `ip_address` filter ignored). So granting the developer bit to an existing moderator silently strips capability from their Logs tab, and demoting them from developer restores it. Not what an admin would expect from a promotion.

Since the paths now differ in what they return, the flag should follow the strongest role the user holds:

```cpp
const bool useDeveloperCommands = (userInfo->user_level() & ServerInfo_User::IsDeveloper) &&
                                  !(userInfo->user_level() & ServerInfo_User::IsModerator);
```

Sorry for the whiplash — my earlier advice was right for the code as it stood then and wrong for the code as it stands now.

cockatrice/src/interface/widgets/tabs/tab_logs.cpp
```diff
@@ -171,6 +182,10 @@ void TabLog::createDock()
     mainRoom = new QCheckBox(tr("Main Room"));
     gameRoom = new QCheckBox(tr("Game Room"));
     privateChat = new QCheckBox(tr("Private Chat"));
+    if (canUseDeveloperCommands) {
+        // Developers cannot query private conversations.
+        privateChat->setVisible(false);
```
ZeldaZach, line 187: Minor, same idea one field over: `findIPAddress` stays visible for developers, but `cmdGetLogHistory` blanks `ip_address` when `allowPrivateChat` is false. A developer who types an IP gets results silently unfiltered by it rather than an empty set — reads as a broken filter. Worth hiding `labelFindIPAddress`/`findIPAddress` alongside `privateChat`.

ZeldaZach (changes requested):
Re-reviewed at 763b5cf74. All seven points from the last round are genuinely addressed — the duplicated `Command_GetLogHistory` is gone and `Command_ViewLogHistory` extends both containers cleanly, the uptime query moved behind `Servatrice_DatabaseInterface::getLatestUptimeSnapshot()` with `toULongLong()` and a `valid` flag, the trailing newline and CMake ordering are fixed, the `admin` bitfield is documented, and `tests/server_developer_role_test.cpp` pins the container permission gate. CI is green across the board. `dev_ext = 1001` and the `containing_type()` lookup in `prepareDeveloperCommand` are a fine workaround for the extension-name collision.

The scoping decision you made — narrow rather than drop — is the right call, but the narrowing has a hole that lets a client get back exactly what it removes, plus a regression for mod+developer users. Both are new with this round's changes, so nothing here repeats earlier feedback.

One more, not inline because the line isn't in the diff: `cmdResetUserPassword` (servatrice/src/serversocketinterface.cpp:1886-1892) enumerates the target's bits as `targetAdmin & 1` / `& 2` under the comment "prevent escalation". Bit 8 isn't in that list, so a plain moderator can force a password reset on a developer's account and then log in as them. Judge (bit 4) has the same gap today, so this is a pre-existing pattern rather than something you broke — but adding a new staff bit is the moment to make the guard `targetAdmin != 0` instead of a bit list that has to be extended by hand every time.

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed in `ab243cc499`. When `!allowPrivateChat` and the developer request leaves `gameType`/`roomType` both false (omitting `log_location` or asking only for "chat"), the server now forces `gameType = true; roomType = true` before calling `getMessageLogHistory`. This guarantees the `target_type` clause is never empty so private-message rows can no longer leak via a crafted location list. The moderator path (`allowPrivateChat=true`) is unchanged.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
BruebachL: Fixed in `ab243cc499`. `useDeveloperCommands` now requires `IsDeveloper && !IsModerator`, so a moderator who also holds the developer bit keeps the moderator path and its full capabilities (private chat, sender_ip, ip filter). The developer bit only selects the narrowed developer family on its own.

cockatrice/src/interface/widgets/tabs/tab_logs.cpp
BruebachL: Fixed in `ab243cc499`. Both `labelFindIPAddress` and `findIPAddress` are now hidden alongside `privateChat` when `canUseDeveloperCommands` is true. The developer path ignores the IP field server-side, so hiding it prevents a developer from getting results that are silently unfiltered by it.

ZeldaZach (approved):

PR #7212, opened 2026-08-29
[Client/Server/Protocol] Surface live metrics in the Developer tab

BruebachL:
## Related Ticket(s)
- Stacked on #7211 (developer role)

- Instrument server command processing, game starts, and event loops
- Add a `MetricsRegistry` with relaxed-atomic counters (counts/totals per protobuf command kind); the Prometheus bucket histogram is removed for now — nothing emitted it, and it returns with the /metrics endpoint that needs it
- Watch worker event loops for stalls via a per-thread heartbeat watchdog
- Surface the live metrics in the Developer tab: overview stats, live metrics section, and a per-command breakdown table
- Extend `Response_GetServerStats` with command timing, event-loop stall, and game-start fields
- Instrument the `developer_command` family (kind 6) added by #7211

## Why
- Lets server operators see request timing and stalls without external tooling, and extends the Developer tab's view to the developer-role command family

## Review fixes (8 threads)
- Drop dead Prometheus histogram + `toPrometheusText`, and the +Inf bug it hid
- Rewrite histogram-bound tests against the production readers
- Read `metrics/slow_command_ms` + `stall_warn_ms` before pool threads start, so `stall_warn_ms=0` genuinely disables the watchdogs
- Correct the `getCardsInGamesTotal()` scrape-cost doc (gameMutex per game, O(cards in play))
- Instrument `developer_command` in `processCommandContainer`/`cmdGetServerStats`
- Shrink `KindStride` 2048 → 1280 (largest extension 1206) with a `static_assert`
- Document unrate-limited slow-command logging in `servatrice.ini.example`

### DIFF
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/response_get_server_stats.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/response_get_server_stats.proto
@@ -2,7 +2,7 @@ syntax = "proto2";
 import "response.proto";
 
 message CommandStats {
-    optional uint32 kind_index = 1;       // 0=session, 1=room, 2=game, 3=moderator, 4=admin
+    optional uint32 kind_index = 1;       // 0=session, 1=room, 2=game, 3=moderator, 4=admin, 5=developer
     optional uint32 extension_number = 2; // protobuf extension number within the kind
     optional string command_name = 3;     // e.g. "session/Command_Ping"
     optional uint64 count = 4;            // number of times observed
--- a/libcockatrice_protocol/libcockatrice/protocol/pb/response_get_server_stats.proto
+++ b/libcockatrice_protocol/libcockatrice/protocol/pb/response_get_server_stats.proto
@@ -1,6 +1,14 @@
 syntax = "proto2";
 import "response.proto";
 
+message CommandStats {
+    optional uint32 kind_index = 1;       // 0=session, 1=room, 2=game, 3=moderator, 4=admin
+    optional uint32 extension_number = 2; // protobuf extension number within the kind
+    optional string command_name = 3;     // e.g. "session/Command_Ping"
+    optional uint64 count = 4;            // number of times observed
+    optional uint64 total_ms = 5;         // cumulative processing milliseconds
+}
+
 message Response_GetServerStats {
     extend Response {
         optional Response_GetServerStats ext = 1220;
@@ -16,4 +24,18 @@ message Response_GetServerStats {
 
     optional uint64 uptime_secs = 6;
     optional uint64 timest = 7; // unix timestamp of the snapshot
+
+    // Live metrics from MetricsRegistry (reset on server restart)
+    optional uint64 cards_in_games = 8;
+    optional uint64 eventloop_stalls_total = 9;
+    optional uint64 eventloop_last_stall_ms = 10;
+    optional uint64 eventloop_max_stall_ms = 11;
+    optional uint64 total_commands = 12;
+    optional uint64 total_command_time_ms = 13;
+    optional int32 active_command_types = 14;
+    optional uint64 game_start_count = 15;
+    optional uint64 game_start_total_ms = 16;
+
+    // Per-command breakdown (only types with count > 0)
+    repeated CommandStats command_stats = 20;
 }
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -384,7 +384,9 @@ max_comments_per_hour=30
 
 [metrics]
 ; Command containers that take longer than this many milliseconds are logged
-; as slow commands. Set to 0 to disable the log line.
+; as slow commands. Set to 0 to disable the log line. A latency spike produces
+; one warning per slow container with no rate limiting of its own -- a bad
+; patch can briefly flood the log, which is how you notice it.
 slow_command_ms=500
 
 ; Each socket pool thread runs a watchdog heartbeat. If a heartbeat arrives
--- a/servatrice/servatrice.ini.example
+++ b/servatrice/servatrice.ini.example
@@ -382,6 +382,17 @@ max_reports_per_day=10
 ; Maximum number of report comments a single user can post per hour; default is 30; set to 0 to disable the limit
 max_comments_per_hour=30
 
+[metrics]
+; Command containers that take longer than this many milliseconds are logged
+; as slow commands. Set to 0 to disable the log line.
+slow_command_ms=500
+
+; Each socket pool thread runs a watchdog heartbeat. If a heartbeat arrives
+; this many milliseconds late, the stall is logged and exposed as
+; servatrice_eventloop_* metrics in the Developer tab. Set to 0 to disable
+; the watchdogs.
+stall_warn_ms=2000
+
 [logging]
 ; Admin/Moderators can query the stored logs for information when looking up reports by various players. This
 ; option can allow or disallow them from doing so.
--- a/servatrice/src/metrics_registry.cpp
+++ b/servatrice/src/metrics_registry.cpp
@@ -2,33 +2,6 @@
 
 #include <QList>
 
-int MetricsRegistry::bucketIndexFor(qint64 elapsedMs)
-{
-    const int lastFiniteBucket = static_cast<int>(BucketBounds.size()) - 1;
-    int bucket = 0;
-    while (bucket < lastFiniteBucket && elapsedMs > BucketBounds[static_cast<size_t>(bucket)]) {
-        ++bucket;
-    }
-    return bucket;
-}
-
-void MetricsRegistry::appendCumulativeBuckets(QString &out,
-                                              const QString &bucketLine,
-                                              const std::array<std::atomic<qint64>, BucketCount> &buckets)
-{
-    // Cumulative buckets are required by the Prometheus histogram convention.
-    qint64 cumulative = 0;
-    for (int bucket = 0; bucket < static_cast<int>(BucketBounds.size()); ++bucket) {
-        cumulative += buckets[static_cast<size_t>(bucket)].load(std::memory_order_relaxed);
-        out += QStringLiteral("%1,le=\"%2\"} %3\n")
-                   .arg(bucketLine)
-                   .arg(BucketBounds[static_cast<size_t>(bucket)])
-                   .arg(cumulative);
-    }
-    cumulative += buckets[BucketCount - 1].load(std::memory_order_relaxed);
-    out += QStringLiteral("%1,le=\"+Inf\"} %2\n").arg(bucketLine).arg(cumulative);
-}
-
 void MetricsRegistry::observeCommand(int typeId, qint64 elapsedMs)
 {
     if (typeId < 0 || typeId >= MaxTypes) {
@@ -43,7 +16,6 @@ void MetricsRegistry::observeCommand(int typeId, qint64 elapsedMs)
     stats.totalMs.fetch_add(elapsedMs, std::memory_order_relaxed);
     totalCommandsCounter.fetch_add(1, std::memory_order_relaxed);
     totalTimeCounter.fetch_add(elapsedMs, std::memory_order_relaxed);
-    stats.buckets[static_cast<size_t>(bucketIndexFor(elapsedMs))].fetch_add(1, std::memory_order_relaxed);
 }
 
 void MetricsRegistry::observeGameStartDurationMs(qint64 elapsedMs)
@@ -54,7 +26,6 @@ void MetricsRegistry::observeGameStartDurationMs(qint64 elapsedMs)
 
     gameStartStats.count.fetch_add(1, std::memory_order_relaxed);
     gameStartStats.totalMs.fetch_add(elapsedMs, std::memory_order_relaxed);
-    gameStartStats.buckets[static_cast<size_t>(bucketIndexFor(elapsedMs))].fetch_add(1, std::memory_order_relaxed);
 }
 
 MetricsRegistry::TypeStats &MetricsRegistry::slotFor(int typeId)
@@ -76,49 +47,6 @@ int MetricsRegistry::activeTypeCount() const
         }
     }
     return active;
-}
-
-QString MetricsRegistry::toPrometheusText(const std::function<QString(int)> &nameForType,
-                                          const QHash<QString, qint64> &gauges) const
-{
-    QString out;
-    out.reserve(4096);
-
-    for (auto it = gauges.constBegin(); it != gauges.constEnd(); ++it) {
-        out += QStringLiteral("# TYPE %1 gauge\n").arg(it.key());
-        out += QStringLiteral("%1 %2\n").arg(it.key()).arg(it.value());
-    }
-
-    // Cumulative buckets are required by the Prometheus histogram convention.
-    out += QLatin1String("# TYPE servatrice_commands_duration_ms histogram\n");
-
-    for (int type = 0; type < MaxTypes; ++type) {
-        const TypeStats &stats = slotFor(type);
-        const qint64 count = stats.count.load(std::memory_order_relaxed);
-        if (count == 0) {
-            continue;
-        }
-
-        const QString label = nameForType ? nameForType(type) : QString::number(type);
-        appendCumulativeBuckets(out, QStringLiteral("servatrice_commands_duration_ms_bucket{command=\"%1\"").arg(label),
-                                stats.buckets);
-
-        out += QStringLiteral("servatrice_commands_duration_ms_sum{command=\"%1\"} %2\n")
-                   .arg(label)
-                   .arg(stats.totalMs.load(std::memory_order_relaxed));
-        out += QStringLiteral("servatrice_commands_duration_ms_count{command=\"%1\"} %2\n").arg(label).arg(count);
-    }
-
-    const qint64 startCount = gameStartStats.count.load(std::memory_order_relaxed);
-    if (startCount > 0) {
-        out += QLatin1String("# TYPE servatrice_game_start_duration_ms histogram\n");
-        appendCumulativeBuckets(out, QLatin1String("servatrice_game_start_duration_ms_bucket"), gameStartStats.buckets);
-        out += QStringLiteral("servatrice_game_start_duration_ms_sum %1\n")
-                   .arg(gameStartStats.totalMs.load(std::memory_order_relaxed));
-        out += QStringLiteral("servatrice_game_start_duration_ms_count %1\n").arg(startCount);
-    }
-
-    return out;
 }
 
 QList<MetricsRegistry::ActiveTypeStats> MetricsRegistry::collectActiveStats() const
--- a/servatrice/src/metrics_registry.cpp
+++ b/servatrice/src/metrics_registry.cpp
@@ -1,0 +1,142 @@
+#include "metrics_registry.h"
+
+#include <QList>
+
+int MetricsRegistry::bucketIndexFor(qint64 elapsedMs)
+{
+    const int lastFiniteBucket = static_cast<int>(BucketBounds.size()) - 1;
+    int bucket = 0;
+    while (bucket < lastFiniteBucket && elapsedMs > BucketBounds[static_cast<size_t>(bucket)]) {
+        ++bucket;
+    }
+    return bucket;
+}
+
+void MetricsRegistry::appendCumulativeBuckets(QString &out,
+                                              const QString &bucketLine,
+                                              const std::array<std::atomic<qint64>, BucketCount> &buckets)
+{
+    // Cumulative buckets are required by the Prometheus histogram convention.
+    qint64 cumulative = 0;
+    for (int bucket = 0; bucket < static_cast<int>(BucketBounds.size()); ++bucket) {
+        cumulative += buckets[static_cast<size_t>(bucket)].load(std::memory_order_relaxed);
+        out += QStringLiteral("%1,le=\"%2\"} %3\n")
+                   .arg(bucketLine)
+                   .arg(BucketBounds[static_cast<size_t>(bucket)])
+                   .arg(cumulative);
+    }
+    cumulative += buckets[BucketCount - 1].load(std::memory_order_relaxed);
+    out += QStringLiteral("%1,le=\"+Inf\"} %2\n").arg(bucketLine).arg(cumulative);
+}
+
+void MetricsRegistry::observeCommand(int typeId, qint64 elapsedMs)
+{
+    if (typeId < 0 || typeId >= MaxTypes) {
+        typeId = MaxTypes - 1; // overflow slot keeps misrouted ids visible
+    }
+    if (elapsedMs < 0) {
+        elapsedMs = 0;
+    }
+
+    TypeStats &stats = slotFor(typeId);
+    stats.count.fetch_add(1, std::memory_order_relaxed);
+    stats.totalMs.fetch_add(elapsedMs, std::memory_order_relaxed);
+    totalCommandsCounter.fetch_add(1, std::memory_order_relaxed);
+    totalTimeCounter.fetch_add(elapsedMs, std::memory_order_relaxed);
+    stats.buckets[static_cast<size_t>(bucketIndexFor(elapsedMs))].fetch_add(1, std::memory_order_relaxed);
+}
+
+void MetricsRegistry::observeGameStartDurationMs(qint64 elapsedMs)
+{
+    if (elapsedMs < 0) {
+        elapsedMs = 0;
+    }
+
+    gameStartStats.count.fetch_add(1, std::memory_order_relaxed);
+    gameStartStats.totalMs.fetch_add(elapsedMs, std::memory_order_relaxed);
+    gameStartStats.buckets[static_cast<size_t>(bucketIndexFor(elapsedMs))].fetch_add(1, std::memory_order_relaxed);
+}
+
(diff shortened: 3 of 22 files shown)


servatrice/src/metrics_registry.cpp
```diff
@@ -0,0 +1,142 @@
+#include "metrics_registry.h"
+
+#include <QList>
+
+int MetricsRegistry::bucketIndexFor(qint64 elapsedMs)
+{
+    const int lastFiniteBucket = static_cast<int>(BucketBounds.size()) - 1;
+    int bucket = 0;
+    while (bucket < lastFiniteBucket && elapsedMs > BucketBounds[static_cast<size_t>(bucket)]) {
```
ZeldaZach: **Bug: the `+Inf` bucket is never incremented, so every sample over 5000 ms is misreported as `le="5000"`.**

`BucketBounds` has 11 entries (indices 0-10, up to 5000), and `BucketCount` is 12, so `buckets[11]` is the `+Inf` slot. But `lastFiniteBucket` is 10 and the loop condition is `bucket < lastFiniteBucket`, so `bucketIndexFor` can return at most 10. Index 11 is dead storage.

Walk a 60-second command through it: `bucket` climbs to 10, the condition `10 < 10` fails, and it returns 10 — the `le="5000"` bucket. Then in `appendCumulativeBuckets` the loop covers 0-10 and emits `le="5000"` including that sample, and the `+Inf` line adds `buckets[11]`, which is always 0.

So the exported histogram makes two false claims at once: `le="5000"` counts samples that took a minute, and `+Inf` always equals `le="5000"`, meaning Prometheus sees zero samples above 5 seconds. `histogram_quantile` over this will never report a p99 above 5000 no matter how bad things get. For a tool whose whole purpose is finding slow commands, that's the wrong failure direction — it reports health during exactly the incident you'd be looking at it for.

The fix is one bound:

```cpp
int MetricsRegistry::bucketIndexFor(qint64 elapsedMs)
{
    int bucket = 0;
    while (bucket < BucketCount - 1 && elapsedMs > BucketBounds[static_cast<size_t>(bucket)]) {
        ++bucket;
    }
    return bucket;
}
```

`BucketCount - 1` is 11, and the short-circuit means `BucketBounds[11]` is never evaluated: at `bucket == 10` the check reads `BucketBounds[10]`, and if that's exceeded `bucket` becomes 11 and the loop exits on `11 < 11`. `BucketCount` is private and `bucketIndexFor` is a member, so no visibility change needed.

tests/metrics_registry_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+    MetricsRegistry registry;
+    registry.observeCommand(MetricsRegistry::typeIdFor(0, 1000), 2);
+    registry.observeCommand(MetricsRegistry::typeIdFor(0, 1000), 30);
+
+    const QString text = registry.toPrometheusText([](int) { return QString("cmd"); }, {});
+
+    // cumulative counts: <=25 -> 1 sample, <=50 -> 2 samples
+    EXPECT_TRUE(text.contains(",le=\"25\"} 1\n"));
+    EXPECT_TRUE(text.contains(",le=\"50\"} 2\n"));
+    EXPECT_TRUE(text.contains(",le=\"+Inf\"} 2\n"));
```
ZeldaZach: The test suite currently locks in the `+Inf` bug rather than catching it.

`BucketsAreCumulative` asserts `,le="+Inf"} 2`, which passes only because the `+Inf` bucket is always empty and therefore always equals the `le="5000"` cumulative. And the largest sample anywhere in the file is `observeCommand(roomLeaveRoom, 5000)` on line 63 — exactly 5000, which `elapsedMs > 5000` correctly routes to the `le="5000"` bucket. Nothing ever crosses the boundary.

Worth adding once `bucketIndexFor` is fixed:

```cpp
TEST(MetricsRegistryTest, SamplesAboveLastBoundLandInInfBucket)
{
    MetricsRegistry registry;
    registry.observeCommand(MetricsRegistry::typeIdFor(0, 1000), 60000);

    const QString text = registry.toPrometheusText([](int) { return QString("cmd"); }, {});
    EXPECT_TRUE(text.contains(",le=\"5000\"} 0\n"));   // must NOT claim the sample
    EXPECT_TRUE(text.contains(",le=\"+Inf\"} 1\n"));
}
```

A boundary case at exactly 5000 is worth keeping alongside it, since `>` vs `>=` there is easy to flip and Prometheus buckets are inclusive upper bounds.

The rest of the file is genuinely good — `OutOfRangeIdsLandInOverflowSlot`, `NegativeDurationsAreClamped` and `GameStartHistogramIsSeparateFromCommandTotals` are all the right things to pin down.

servatrice/src/servatrice.cpp
```diff
@@ -470,9 +474,60 @@ bool Servatrice::initServer()
     }
 
     setRequiredFeatures(getRequiredFeatures());
+
+    // METRICS (always active. Slow-command logging and stall watchdogs are
+    // controlled by their respective thresholds below)
+    metricsSlowCommandMs = settingsCache->value("metrics/slow_command_ms", 500).toInt();
```
ZeldaZach: **Bug: `metrics/stall_warn_ms = 0` does not disable the watchdogs.**

These two reads sit at the very end of `initServer()`, but `new Servatrice_GameServer(...)` runs earlier in the same function (line 438), and its constructor calls `server->watchWorkerThread(newThread)` for every pool thread. So by the time `watchWorkerThread` evaluates `if (metricsStallWarnMs <= 0)`, the member still holds its in-class initializer of `2000` — the configured value hasn't been loaded yet.

An operator who sets `stall_warn_ms=0` following the ini comment ("Set to 0 to disable the watchdogs") gets a watchdog installed in every pool thread anyway. It then behaves oddly rather than silently: `checkHeartbeat` reads `getMetricsStallWarnMs()` live, which by then *is* 0, so `overshootMs < 0` is false for every tick and **every single heartbeat is reported as a stall** — a `qWarning` line per thread every 500 ms, plus `eventLoopStallsTotal` climbing forever. Setting the value to 0 turns the log firehose on instead of off.

Fix is to move both reads above the server construction — right after `setRequiredFeatures()` is too late, so somewhere near the top of `initServer()`:

```cpp
metricsSlowCommandMs = settingsCache->value("metrics/slow_command_ms", 500).toInt();
metricsStallWarnMs = qMax(0, settingsCache->value("metrics/stall_warn_ms", 2000).toInt());
```

While you're there: `checkHeartbeat` re-reads the threshold on every tick even though it can't change after startup. Caching it in the watchdog at `start()` would drop a cross-object read from a 500 ms timer in every pool thread. Minor, but free.

servatrice/src/metrics_registry.h
```diff
@@ -0,0 +1,135 @@
...
+    GameStartSnapshot getGameStartSnapshot() const;
+
+    /**
+     * @brief Renders all recorded data in Prometheus text exposition format.
+     *
+     * @param nameForType maps a numeric command type id to a stable label
+     * value. Ids without a mapping are rendered as their number.
+     * @param gauges simple name/value pairs emitted as gauge samples.
+     */
+    QString toPrometheusText(const std::function<QString(int)> &nameForType,
```
ZeldaZach: `toPrometheusText` has no production callers — only `tests/metrics_registry_test.cpp`. The stats path uses `collectActiveStats()` and `getGameStartSnapshot()`, and nothing in servatrice exposes an HTTP endpoint or writes this text anywhere.

That leaves the PR's headline feature ("Prometheus-style cumulative histogram buckets") as ~55 lines that only ever run under gtest. It's also where the `+Inf` bug lives, which is part of why it survived: the buckets are correct enough for the Developer tab, which only reads counts and totals.

Two reasonable resolutions:

- **Keep it and wire it up** — a `/metrics` endpoint is the natural home, but that's clearly its own PR. If that's the plan, a line in the comment saying so would help the next reader.
- **Drop it for now** and keep `collectActiveStats()`, which is what the feature actually uses. It can come back with the endpoint that needs it.

I'd lean toward the second — dead code that looks load-bearing is worse than no code, and the histogram bug is a good illustration of why: nothing exercised it in anger.

servatrice/src/servatrice.cpp
```diff
@@ -470,9 +474,60 @@ bool Servatrice::initServer()
...
+        return; // watchdogs disabled via metrics/stall_warn_ms = 0
+    }
+
+    auto *watchdog = new EventLoopWatchdog(this, thread->objectName());
+    connect(thread, &QThread::finished, watchdog, &QObject::deleteLater);
+    watchdog->moveToThread(thread);
+    QMetaObject::invokeMethod(watchdog, &EventLoopWatchdog::start, Qt::QueuedConnection);
+}
+
+qint64 Servatrice::getCardsInGamesTotal() const
```
ZeldaZach, line 516: This is more expensive than the doc comment on the declaration suggests. `servatrice.h` says scrape cost "grows with live game count only", but `Server_Game::getCardsInGame()` takes `gameMutex` and iterates every player calling `getCardCount()`, so the real cost is O(total cards in play) plus one `gameMutex` acquisition per game — and `gameMutex` is the hot per-game lock every game action contends for.

So a single stats refresh walks every room, every game, and every player on the server while holding `roomsLock` for read the whole time. On a busy instance that's a brief but global serialization point, and it's triggered by a UI button any Developer can hold down. Not a deadlock (the ordering matches `prepareDestroy`), just a self-inflicted latency spike available to anyone with the role.

Since this number only needs to be roughly current, computing it on the existing `statusUpdate()` tick and storing it in an atomic would make the scrape O(1) and move the walk to a cadence the server already pays for. If you'd rather keep it live, please at least correct the comment so the next person knows what they're paying for.

servatrice/src/serversocketinterface.cpp
```diff
@@ -193,6 +196,41 @@ void AbstractServerSocketInterface::logDebugMessage(const QString &message)
...
+    for (const auto &cmd : cont.room_command()) {
+        servatrice->getMetricsRegistry().observeCommand(MetricsRegistry::typeIdFor(1, getPbExtension(cmd)), elapsedMs);
+    }
+    for (const auto &cmd : cont.game_command()) {
+        servatrice->getMetricsRegistry().observeCommand(MetricsRegistry::typeIdFor(2, getPbExtension(cmd)), elapsedMs);
+    }
+    for (const auto &cmd : cont.moderator_command()) {
+        servatrice->getMetricsRegistry().observeCommand(MetricsRegistry::typeIdFor(3, getPbExtension(cmd)), elapsedMs);
+    }
+    for (const auto &cmd : cont.admin_command()) {
```
ZeldaZach: `cont.developer_command()` isn't instrumented, so the command family this whole feature exists to serve is the one family with no timings. `NumKinds` is 5 and `KindNames` stops at `admin`, matching the five loops here — but #7211, which this is stacked on, added a sixth container. Adding it means bumping `NumKinds` to 6, appending `"developer"` to `KindNames`, adding a loop here, and appending `"DeveloperCommand"` to the `messageNames[]` array at line 1762.

Worth calling out explicitly in the PR description too: because the container's total time is attributed to *each* command inside it, `totalTimeMs()` exceeds real wall-clock time whenever containers are batched. Averages stay right (both numerator and denominator scale together), but per-command histograms will be systematically inflated for game commands, which the client does batch. The inline comment covers the attribution rule; the effect on the histogram is the part someone reading a p99 needs to know.

servatrice/src/metrics_registry.h
```diff
@@ -0,0 +1,135 @@
...
+ * beyond the unavoidable counter updates.
+ *
+ * Reading happens rarely (metrics scraping), accepts momentary tears between
+ * related counters, and therefore also needs no synchronization.
+ */
+class MetricsRegistry
+{
+public:
+    /**
+     * Extension numbers are only unique per command kind, so recorded ids
```
ZeldaZach, line 33: Nit on sizing: `MaxTypes` is `5 * 2048 = 10240` slots, each `TypeStats` being 2 atomics plus 12 bucket atomics — about 112 bytes, so roughly **1.1 MB** of mostly-zero storage embedded by value in `Servatrice`. The highest command extension number actually in use is 1206 (`ModeratorCommand`), so ~30 real command types occupy 10240 slots.

It's not harmful — 1.1 MB is nothing for a server, and `activeTypeCount()` / `collectActiveStats()` scanning 10240 slots per scrape is microseconds. But a stride of 1280 would cut it to ~700 KB with plenty of headroom, and more usefully there's currently nothing that catches the failure mode if an extension number ever does reach 2048: `typeIdFor` would silently fold it into the next kind's range, and `observeCommand`'s overflow guard only catches ids past `MaxTypes` entirely, not cross-kind collisions.

A `static_assert` won't work here since the numbers live in generated protobuf headers, but a one-line comment on `KindStride` saying "must exceed the highest extension number in any command family; currently 1206" would at least leave a trail for whoever adds command 2048.

servatrice/servatrice.ini.example
```diff
@@ -382,6 +382,17 @@ max_reports_per_day=10
...
+[metrics]
+; Command containers that take longer than this many milliseconds are logged
+; as slow commands. Set to 0 to disable the log line.
+slow_command_ms=500
+
+; Each socket pool thread runs a watchdog heartbeat. If a heartbeat arrives
+; this many milliseconds late, the stall is logged and exposed as
+; servatrice_eventloop_* metrics in the Developer tab. Set to 0 to disable
+; the watchdogs.
+stall_warn_ms=2000
```
ZeldaZach, line 396: Given the `stall_warn_ms=0` problem described on `servatrice.cpp:480`, this comment currently documents behaviour that doesn't exist. Once the initialisation order is fixed it's accurate as written.

One addition worth making: `slow_command_ms` logs at warning level with no rate limiting of its own, one line per container over the threshold. If a server hits a bad patch where many commands cross 500 ms, this becomes a log flood at exactly the moment the logs are most worth reading. Mentioning that operators may want to raise it on a busy instance — or adding a simple per-interval cap on the warning — would help.

ZeldaZach (changes requested):
Reviewed in a local worktree, diffed against `developer-role` (#7211). Note the PR body says "Stacked on #7188" — #7188 is a merged GameScene animation fix; the actual base is **#7211**.

The instrumentation design is sound. Relaxed atomics on preallocated storage with no locks or allocations on the hot path is the right call for something that runs per command, the `kindIndex * stride + extensionNumber` encoding cleanly solves extension numbers only being unique per family, `observeEventLoopStall`'s CAS loop for the max is correct, and the lock ordering in `getCardsInGamesTotal` (rooms → games → gameMutex) matches what `prepareDestroy` already does, so no deadlock. I also checked that calling into `servatrice` *after* `Server_ProtocolHandler::processCommandContainer` is safe: `prepareDestroy()` ends in `deleteLater()`, so the handler is still alive when the metrics code runs.

Two real bugs, though, and both of them break the feature in the specific case it exists to catch:

1. **The histogram never uses its `+Inf` bucket.** Anything slower than 5000 ms gets counted in `le="5000"`. So the one thing a latency histogram is for — seeing the slow tail — is exactly what this can't show, and it actively reports slow samples as fast ones.
2. **`stall_warn_ms=0` does not disable the watchdogs**, because the setting is read after the thread pools are already constructed.

Plus `toPrometheusText` — the Prometheus exposition format the PR headlines — is never called outside tests. Details inline.

servatrice/src/metrics_registry.cpp
BruebachL: Now moot: the cumulative histogram it described has been dropped (see the di2-a thread). `bucketIndexFor` no longer exists because nothing records or renders time buckets — the counts/totals the Developer tab actually reads cannot misroute a 5000+ ms sample anywhere.

tests/metrics_registry_test.cpp
BruebachL: The tests no longer rely on the dead histogram surface. `BucketsAreCumulative` is gone along with it; the rewritten suite covers totals, per-type slots, kind encoding, overflow clamping, negative clamping, and game-start separation against the production readers (`collectActiveStats`/`getGameStartSnapshot`), and nothing asserts behavior that no longer exists.

servatrice/src/servatrice.cpp
BruebachL: Fixed: the two threshold reads now sit at the very top of `initServer()` (right after `getServerID()`), i.e. before either socket server is constructed, so `metrics/stall_warn_ms = 0` disables the per-pool watchdogs exactly as documented.

servatrice/src/metrics_registry.h
BruebachL: Dropped per your lean. `toPrometheusText`, `appendCumulativeBuckets`, the time-bucket histogram, and the gauge plumbing are deleted along with the tests bound to them; the class doc now says the histogram belongs to the future /metrics exporter that needs it. The hot path lost the 12 per-sample bucket atomics as a bonus. `collectActiveStats`/`getGameStartSnapshot` (what the Developer tab and `cmdGetServerStats` read) are unchanged.

servatrice/src/servatrice.cpp
BruebachL: Comment corrected: it now states the real cost — one `gameMutex` acquisition per live game (contended with every in-game action) plus an O(total cards in play) walk, and advises keeping scrapes infrequent on busy servers.

servatrice/src/serversocketinterface.cpp
BruebachL: `developer_command` is now instrumented like the five existing families: `NumKinds` bumped to 6 with `"developer"` appended to `KindNames`, a loop added in `processCommandContainer()` (kind index 5), and `DeveloperCommand` appended to the descriptor lookup table in `cmdGetServerStats()`. Confirmed the base PR routes these containers server-side, so the timings actually fire.

servatrice/src/metrics_registry.h
BruebachL: Sized down: `KindStride` is now 1280, the first round number above the largest extension in use (ModeratorCommand = 1206), cutting the embedded array from ~1.1 MB to ~120 KB (6 kinds, count+totalMs only since the histogram is gone). A `static_assert` guards the invariant that the stride stays above the highest extension number.

servatrice/servatrice.ini.example
BruebachL: The `stall_warn_ms=0` comment is accurate again now that the init order is fixed (see di2-Y). Also documented that slow-command warnings have no rate limiting of their own — one line per slow container, which is how a bad latency patch surfaces in the log.

servatrice/src/serversocketinterface.cpp
```diff
@@ -192,6 +195,44 @@ void AbstractServerSocketInterface::logDebugMessage(const QString &message)
...
+        servatrice->getMetricsRegistry().observeCommand(MetricsRegistry::typeIdFor(4, getPbExtension(cmd)), elapsedMs);
+    }
+    for (const auto &cmd : cont.developer_command()) {
+        servatrice->getMetricsRegistry().observeCommand(MetricsRegistry::typeIdFor(5, getPbExtension(cmd)), elapsedMs);
+    }
```
ZeldaZach: **These six loops record commands the base class never processed.**

`Server_ProtocolHandler::processCommandContainer` dispatches exactly one family per container — it is an if/else-if chain, `game_command` first, `developer_command` last — and each `process*CommandContainer` returns `RespLoginNeeded` before its loop body if the session lacks the role. It also returns immediately when `deleted` is set. None of that is visible here: these loops run over every family in the container unconditionally, so anything present but not dispatched still lands in the registry with the container's elapsed time.

Concretely, an unauthenticated socket can send one container holding a `session_command` (login) plus a hundred fabricated `developer_command` entries. The base dispatches only the session command; the developer commands are never executed and never authorized — but the Developer tab will show a hundred `developer/Command_GetServerStats` samples as genuine server activity. Same for `moderator_command` and `admin_command`. For a diagnostic surface whose whole value is being trustworthy during an incident, forgeable rows are worse than missing rows.

Mirroring the base's selection fixes it — record only the family that was actually dispatched, and only when the handler was not already `deleted`.

Second issue in the same block: `getPbExtension()` returns **-1** for a command with no extension set (a reachable case — `processRoomCommand` and friends answer `RespInvalidCommand` for it). Only kind 0 is protected, because `typeIdFor(0, -1)` is -1 and `observeCommand`'s `typeId < 0` guard catches it. Every other kind wraps into the *previous* kind's range: `typeIdFor(1, -1)` is 1279, i.e. session/extension 1279, and `typeIdFor(5, -1)` is 6399, i.e. admin/extension 1279. So an empty `room_command{}` is silently attributed to the session family. Worth skipping `extensionNumber < 0` (or clamping it in `typeIdFor`) so it reaches the overflow slot the guard was written for.

servatrice/src/servatrice.cpp
```diff
@@ -728,13 +785,15 @@ void Servatrice::incTxBytes(quint64 num)
     txBytesMutex.lock();
     txBytes += num;
     txBytesMutex.unlock();
+    txBytesTotal.fetch_add(num, std::memory_order_relaxed);
```
ZeldaZach: `txBytesTotal` / `rxBytesTotal` (and `getUptimeSeconds()`) have no readers anywhere in the PR — `cmdGetServerStats` fills `tx_bytes`/`rx_bytes` from `getLatestUptimeSnapshot()`, not from these. So this adds an atomic RMW to `incTxBytes`/`incRxBytes`, which run on every socket write and read, for counters nothing consumes.

Same call as the histogram last round: either wire them up or drop them. Wiring them up is arguably the better half here, since the tab currently labels the DB snapshot values "(last tick)" and a process-lifetime total is the number an operator actually wants next to it — but that needs new response fields. If that is not this PR, delete the two atomics and the three getters.

libcockatrice_protocol/libcockatrice/protocol/pb/response_get_server_stats.proto
```diff
@@ -1,6 +1,14 @@
 syntax = "proto2";
 import "response.proto";
 
+message CommandStats {
+    optional uint32 kind_index = 1;       // 0=session, 1=room, 2=game, 3=moderator, 4=admin
```
ZeldaZach: The mapping comment stops at 4, but `NumKinds` is 6 in this PR and the server emits `kind_index = 5` for developer commands. Since this is the wire contract clients decode against, please add `5=developer`.

Field numbering itself is fine — 8-16 and 20 are fresh, nothing renumbered, and `optional` throughout, so older clients skip them cleanly.

ZeldaZach (changes requested):
Re-review of the four commits. All eight threads from the 2026-08-30 review are genuinely resolved: the histogram and `toPrometheusText` are gone (taking the `+Inf` bug with them), the two `metrics/*` reads now sit at the top of `initServer()` before either pool server is constructed so `stall_warn_ms=0` really does skip `watchWorkerThread`, the `getCardsInGamesTotal()` doc comment states the real cost, `developer_command` is instrumented as kind 5 with `NumKinds`/`KindNames`/`messageNames[]` all bumped together, `KindStride` is 1280 with a `static_assert`, and the ini documents the unrate-limited slow-command logging. I re-checked the lock ordering after the changes: `getCardsInGamesTotal()` walks roomsLock(read) -> gamesLock(read) -> gameMutex, `~Server_Game` takes gamesLock(write) -> gameMutex, and both `gamesLock` and `gameMutex` are recursive, so the nested `getPlayers()`/`getGames()` calls are safe. CI is green across all 18 checks. Nothing in this PR widens what an unprivileged client can *read* — `cmdGetServerStats` is still gated by #7211's `processDeveloperCommandContainer`, and the response carries only aggregates and command type names, no IPs or user data.

One new bug, though, and it is on the write side of that gate rather than the read side: the instrumentation records samples for commands the server never dispatched, including ones it rejected for lack of permission. Details inline, plus two smaller items.

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed in `9d34fa19af`. `processCommandContainer` now mirrors the base if/else-if chain and records only the single family the base actually dispatches, and returns early when the handler was already `deleted`. It also skips entries whose extension number is -1 (no extension set), which previously wrapped into the previous kind's id range instead of hitting the typeId < 0 guard. An unauthenticated socket can no longer forge developer/moderator/admin samples by batching them with a session command.

servatrice/src/servatrice.cpp
BruebachL: Fixed in `9d34fa19af` — dropped them. `txBytesTotal`/`rxBytesTotal` and the `getTxBytesTotal`/`getRxBytesTotal`/`getUptimeSeconds` getters had no readers anywhere (the wire path fills tx_bytes/rx_bytes/uptime_secs from the DB snapshot in `getLatestUptimeSnapshot`). The two atomics and all three getters are removed, so `incTxBytes`/`incRxBytes` no longer pay the atomic RMW. The `incTxBytes`/`incRxBytes` slots and their mutexes stay — `IslInterface` still calls them for the legacy `txBytes`/`rxBytes` counters. Wiring live process-lifetime totals in would need new response fields, which I judged out of scope for this PR.

libcockatrice_protocol/libcockatrice/protocol/pb/response_get_server_stats.proto
BruebachL: Fixed in `9d34fa19af` — the comment now reads `0=session, 1=room, 2=game, 3=moderator, 4=admin, 5=developer`, matching `NumKinds = 6` and the kind 5 emitted for developer commands.

ZeldaZach (approved):

Issue #6929, opened 2026-05-22
Use native Qt features over unmaintained external libs (`qt-json`,`zip`)


tooomm:
Follow-up after #6901 

>Looking at https://github.com/Cockatrice/Cockatrice/blob/master/format.sh#L21-L26, which lists our external libs, there is also:
>- Oracle: [`lzma`](https://github.com/Cockatrice/Cockatrice/tree/master/oracle/src/lzma)
>  Can't find the source, seems to be related to xz
>- ~~Oracle: [`qt-json`](https://github.com/Cockatrice/Cockatrice/tree/master/oracle/src/qt-json)~~
>  ~~The [repo](https://github.com/qt-json/qt-json) got archived and refers to a "native JSON object class" instead - [this one](https://doc.qt.io/qt-6/qjsonobject.html)?~~
Removed in #7214
>- Oracle: [`zip`](https://github.com/Cockatrice/Cockatrice/tree/master/oracle/src/zip)
>  No idea where this is sourced from
>- Servatrice: [`smtp`](https://github.com/Cockatrice/Cockatrice/tree/master/servatrice/src/smtp)
>  libqxt seems unmainted
>
>Anything missing?

Turns out JSON and ZIP can be handled since Qt5 natively.
We should use the build-in, tested and maintained features over outdated and archived dependencies.

ebbit1q:
we tried native json before and found them unable to deal with the size of mtgjson if I remember correctly, we can search for alternatives or continue maintaining qtjson as we have been, including the update to qt6

the smtp library we will have to maintain ourselves, there isn't an alternative, but smtp isn't exactly complex

zip is the osdab zip library, it has not been updated since: https://github.com/hippydream/osdab/blob/master/OSDaB-Zip/unzip.cpp

lzma contains a simple xz decompression library put into the public domain by the xz maintainer, the xz format has not changed and will not change, we can continue using this, the xz source code can be found here https://git.tukaani.org/?p=xz.git;a=summary

tooomm:
Thanks for providing additional references and helpful input!

<br>

>we tried native json before and found them unable to deal with the size of mtgjson if I remember correctly, we can search for alternatives or continue maintaining qtjson as we have been, including the update to qt6

**Regarding native JSON handling via Qt:**
Before Qt 5.15, there was a size limit in place: [QJsonObject size maximum length 128MB](https://bugreports.qt.io/browse/QTBUG-47629)

[...]

PR #7214, opened 2026-08-29
[Oracle] Replace vendored qt-json with native QJson for set import

BruebachL:
Replace the stalled Qt5-era vendored QtJson parser with native `QJson` for set import.

- Drop the vendored `qt-json/json.{cpp,h}` sources plus their CMake wiring
- Import set `.cards` straight into `QJsonArray`/`QJsonObject` instead of a `QList<QVariant>` round-trip
- Release the imported JSON document buffer once `CardInfo` objects are built (save page)

Behaviour is preserved from the old parser:

- Property reads coerce numbers/booleans to text (`manaValue`, `convertedManaCost`, `isOnlineOnly`, `isRebalanced` keep producing the same `cards.xml` output)
- Legalities are only merged onto an existing printing when the card has none yet
- The Basic Land legality exception rule actually matches now: the old C++ string literal made `\b` a literal backspace, so the raw-string `\bBasic\b[^—]+\bLand\b` is a user-visible legality fix (was dead code)
- Memoized format rules are returned as a shared `const` map and documented as immutable
- `colors`/`colorIdentity` are iterated as JSON arrays directly instead of two `QVariantList`→`QStringList` copies

Enables the following stacked PRs: native data access for the oracle tests, RAM benchmarks, and lazy per-set parsing.


tooomm (changes requested):
Linking #6929

Please clean the complete [`oracle/src/qt-json`](https://github.com/Cockatrice/Cockatrice/tree/optimize-oracle-native-json/oracle/src/qt-json) folder.

oracle/src/oracleimporter.cpp
```diff
@@ -474,6 +467,11 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QList
 
 FormatRulesNameMap OracleImporter::createDefaultMagicFormats()
 {
+    static FormatRulesNameMap cached;
```
RickyRister: `createDefaultMagicFormats` can already be a static method. Instead of modifying a static field (which is confusing), you can just do `static FormatRulesNameMap cached = createDefaultMagicFormats()` or something like that.

oracle/src/oracleimporter.cpp
```diff
@@ -222,12 +222,12 @@ CardInfoPtr OracleImporter::addCard(QString name,
     return newCard;
 }
 
-static QString getStringPropertyFromMap(const QVariantMap &card, const QString &propertyName)
+static inline QString getJsonString(const QJsonObject &obj, const QString &key)
```
RickyRister: nit: `inline` is redundant here

RickyRister (approved):

tooomm (approved):
Thanks for cleaning up the complete vendored qt-json  folder.

For the actual code, I defer to others.

oracle/src/oracleimporter.cpp
```diff
@@ -222,12 +222,12 @@ CardInfoPtr OracleImporter::addCard(QString name,
     return newCard;
 }
 
-static QString getStringPropertyFromMap(const QVariantMap &card, const QString &propertyName)
+static QString getJsonString(const QJsonObject &obj, const QString &key)
 {
-    return card.contains(propertyName) ? card.value(propertyName).toString() : QString("");
+    return obj.value(key).toString();
```
ZeldaZach: `QVariant::toString()` and `QJsonValue::toString()` are not interchangeable. `QVariant` coerces numbers and booleans to text; `QJsonValue::toString()` returns a null `QString` for anything whose `type()` is not `String`.

I compiled both paths against real MTGJSON field shapes to confirm:

```
field                old (QVariant) new (QJsonValue)
-----                -------------- ----------------
name                 'Lightning Bolt' 'Lightning Bolt'
manaCost             '{R}'          '{R}'
manaValue            '1'            ''               <-- DIFFERS
convertedManaCost    '1'            ''               <-- DIFFERS
type                 'Instant'      'Instant'
isOnlineOnly         'true'         ''               <-- DIFFERS
isRebalanced         'false'        ''               <-- DIFFERS
number               '133'          '133'
```

In MTGJSON v5, `manaValue`, `convertedManaCost` and `faceManaValue`/`faceConvertedManaCost` are floats, and `isOnlineOnly`/`isRebalanced` are booleans. (I checked `ANB.json`, `YMID.json` and `MID.json` from the v5 API: `manaValue` is a float on 479/479 cards in MID, and MID has 8 cards with `"isRebalanced": true`.) Every other key read through this helper is a string, so those four fields are the whole exposure.

Because the callers at lines 287 and 296 skip empty values, the generated `cards.xml` loses the `<cmc>` element for every card and the `isOnlineOnly`/`isRebalanced` printing attributes entirely. Line 385 is the exception: `properties["cmc"] = faceManaValue;` is unconditional, so double-faced cards get an empty `<cmc></cmc>`.

What that breaks downstream:

- Deck Analytics mana curves (`deck_list_statistics_analyzer.cpp:43`) collapse to a single bar at 0, as does the draw-probability cmc breakdown.
- `FilterItem::acceptCmc` (`filter_tree.cpp:257`) only ever matches `cmc == 0`, and search strings like `cmc>3` / `mv=2` return nothing.
- Sorting by mana value in the deck list, card database, in-game card lists and the visual deck editor all become no-ops.
- With `isRebalanced` missing, `cockatrice_xml_4.cpp:334` reads `""`, which is `!= "true"`, so the "include rebalanced cards" toggle in `dlg_manage_sets.cpp` silently stops filtering Alchemy printings.

And it persists: the missing properties are written into `cards.xml`, so it survives a restart until the user re-runs a fixed Oracle.

One-line fix that restores the old semantics exactly:

```cpp
static QString getJsonString(const QJsonObject &obj, const QString &key)
{
    return obj.value(key).toVariant().toString();
}
```

If you'd rather not pay for the `QVariant` round-trip on the hot path, handling `Double`/`Bool` explicitly works too, but note that `QVariant(1.0).toString()` gives `"1"` rather than `"1.0"`, which is what the XML has always contained.

oracle/src/oracleimporter.cpp
```diff
@@ -142,12 +143,11 @@ CardInfoPtr OracleImporter::addCard(QString name,
...
-    if (cards.contains(name)) {
-        CardInfoPtr card = cards.value(name);
+    auto existingIt = cards.constFind(name);
+    if (existingIt != cards.constEnd()) {
+        CardInfoPtr card = existingIt.value();
         card->addToSet(printingInfo.getSet(), printingInfo);
-        if (card->getProperties().filter(formatRegex).empty()) {
-            card->combineLegalities(properties);
-        }
+        card->combineLegalities(properties);
```
ZeldaZach: This drops the `if (card->getProperties().filter(formatRegex).empty())` guard, so legalities now get merged on every subsequent printing instead of only when the card has none yet. That is a real behaviour change for multi-format cards, and it isn't mentioned in the PR description.

#7215 puts the guard back verbatim, so as the stack stands the change exists for exactly one commit. Worth dropping from this PR so the parser swap stays behaviour-preserving and the legality semantics are decided in one place.

oracle/src/oracleimporter.cpp
```diff
@@ -465,20 +458,20 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QList
...
 
-FormatRulesNameMap OracleImporter::createDefaultMagicFormats()
+static FormatRulesNameMap buildDefaultMagicFormats()
 {
     // Predefined common exceptions
     CardCondition superTypeIsBasic;
     superTypeIsBasic.field = "type";
     superTypeIsBasic.matchType = "regex";
-    superTypeIsBasic.value = "\bBasic\b[^—]+\bLand\b";
+    superTypeIsBasic.value = R"(\bBasic\b[^—]+\bLand\b)";
```
ZeldaZach, line 487: Good catch, and worth calling out explicitly since it's easy to miss in a parser-swap diff: the old `"\bBasic\b[^—]+\bLand\b"` was a C++ string literal, so `\b` was a literal backspace (0x08) rather than a regex word boundary. The exception rule for basic lands could never have matched. The raw string literal fixes it.

Two notes: this is a user-visible deck-legality fix that deserves a line in the PR description (#7215's description currently claims it), and it'd be worth a test since nothing currently covers `createDefaultMagicFormats()`.

oracle/src/oracleimporter.cpp
```diff
@@ -537,6 +529,12 @@ FormatRulesNameMap OracleImporter::createDefaultMagicFormats()
     return defaultFormatRulesNameMap;
 }
 
+FormatRulesNameMap OracleImporter::createDefaultMagicFormats()
+{
+    static const FormatRulesNameMap cached = buildDefaultMagicFormats();
```
ZeldaZach, line 547: Memoizing this is reasonable, but it changes the ownership model: `FormatRulesNameMap` holds `FormatRulesPtr`, so every caller now gets shared pointers to the *same* `FormatRules` objects rather than a fresh set each call. Today only `saveToFile()` calls this and it doesn't mutate them, so it's fine, but the returned value is no longer safely mutable and nothing says so.

Either return `const FormatRulesNameMap &` to make the sharing explicit, or leave a comment noting the entries must be treated as immutable. Also, now that the body is in a free function, `createDefaultMagicFormats()` no longer touches instance state and could be `static`.

oracle/src/oracleimporter.cpp
```diff
@@ -331,21 +326,20 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QList
             allNameProps.contains(faceName) && layout == "normal" && lastChar.isLetter()) {
             numComponent = " (" + QString(lastChar).toLower() + ")";
         }
-        allNameProps.append(faceName);
+        allNameProps.insert(faceName);
 
         // special handling properties
-        QString colors = card.value("colors").toStringList().join("");
+        QString colors = card.value("colors").toVariant().toStringList().join("");
```
ZeldaZach: `toVariant().toStringList()` (here and on lines 337 and 342) materializes the whole `QJsonArray` into a `QVariantList` and then into a `QStringList`, so it's two deep copies per card for what is usually a 1-3 element array. Since this PR is the base of a stack about parse cost and peak RAM, iterating the array directly avoids both:

```cpp
QString colors;
for (const QJsonValue &c : card.value("colors").toArray()) {
    colors += c.toString();
}
```

`getMainCardType()` on line 342 needs a `QStringList`, so that one is harder to avoid without changing its signature. Not blocking.

ZeldaZach (changes requested):
Nice cleanup overall, and dropping 951 lines of vendored parser is very welcome. The `QJsonArray`/`QJsonObject` conversion is mechanical and correct almost everywhere, but there is one substitution that silently changes semantics for every card in the database, so I'd like that fixed before this lands.

There is also some scope creep here: the PR description says "replace the parser", but the diff also changes legality merging, fixes the Basic Land regex, and memoizes `createDefaultMagicFormats()`. Two of those are genuine fixes and one is reverted again in #7215. Splitting them out would make this much easier to bisect later.

Details inline.

oracle/src/oracleimporter.cpp
BruebachL: Restored the original coercion semantics in `getJsonString`: `obj.value(key).toVariant().toString()`, so `manaValue`/`convertedManaCost`/`isOnlineOnly`/`isRebalanced` keep producing text and `<cmc>`, the online-only/rebalanced attributes survive into cards.xml.

oracle/src/oracleimporter.cpp
BruebachL: Restored the guard verbatim in `addCard`: legalities are only merged when `card->getProperties().filter(formatRegex).empty()`, so the parser swap stays behaviour-preserving and multi-format printings dont overwrite each other. (This also supersedes #7215s planned re-add.)

oracle/src/oracleimporter.cpp
BruebachL: Kept the raw-string fix and added a Basic Land line to the PR description. Agreed it deserves a test, but `createDefaultMagicFormats` lives in the Oracle app executable (no importer library), so a format-rules test needs a linkage refactor first — tracking it as a follow-up rather than ballooning this PR.

oracle/src/oracleimporter.cpp
BruebachL: Made the ownership explicit: `createDefaultMagicFormats()` now returns `const FormatRulesNameMap &` (still memoized via the static local) and the public declaration documents that the entries are shared and must be treated as immutable.

oracle/src/oracleimporter.cpp
BruebachL: Iterate the JSON arrays directly for `colors` and `colorIdentity` (no `QVariantList`/`QStringList` materialization). `getMainCardType` still takes the list, so that call keeps the `toVariant().toStringList()` as you noted its not blocking.

ZeldaZach (approved):
Re-reviewed at b164973. All five points from my previous review are addressed:

- `getJsonString` is back on `toVariant().toString()`, with a comment explaining why. I compiled the old vendored parser and the new path side by side against a realistic MTGJSON card blob and they agree on every field that matters (`manaValue`/`convertedManaCost` -> `1`, `isOnlineOnly` -> `true`, `isRebalanced` -> `false`, missing key and JSON `null` -> empty, `releaseDate` -> same QDate via ISODate). The only divergence I could produce is an unsigned integer above `INT64_MAX`, which the old parser kept exact and QJson widens to a double - no MTGJSON field is anywhere near that, so it's a non-issue.
- The `filter(formatRegex).empty()` legality guard is restored verbatim, so the parser swap is behaviour-preserving again and #7215 no longer needs to re-add it.
- `createDefaultMagicFormats()` now returns `const FormatRulesNameMap &` off a memoized static local, with the immutability documented on the declaration. Fine to leave the free `buildDefaultMagicFormats()` doing the construction.
- `colors`/`colorIdentity` iterate the `QJsonArray` directly now; keeping `toVariant().toStringList()` for `getMainCardType()` is the right call for this PR.
- Raw-string Basic Land regex kept and documented in the description; agreed a format-rules test needs the Oracle linkage split out first, follow-up is fine.

Also checked: `QJsonParseError` is inspected and a non-object document falls through to the empty-list `return false`; `oracle/src/qt-json/` is gone in full along with its AUTHORS/LICENSE/README, and nothing in the tree still references `qt-json`/`QtJson` (CMake source list and the `format.sh` exclude are both cleaned up). `releaseSetData()` is safe - QWizard only re-runs `SaveSetsPage::initializePage()` after a Back, and `cleanupPage()` already called `importer->clear()` on that path, so `allSets` was empty there before this PR too.

CI is green across all 18 jobs. Nice cleanup - net -852 lines and one less vendored parser to carry.

PR #7215, opened 2026-08-29
[Oracle] Add oracle importer tests and fix set parsing details

BruebachL:
Wire up the first real unit coverage for the oracle importer, plus a handful of set-parsing details found while writing the fixtures.

**Tests (`tests/oracle/oracle_importer_test.cpp`, 25 cases):**
- `sortAndReduceColors`: single color, WUBRG ordering, dedup, all-five
- Legality guard both halves: first printing without legalities still merges, later printings don't overwrite
- `createDefaultMagicFormats`: expected formats present, singleton deck sizes, vintage restricted, Basic Land regex actually matches (null-safe, exact pointer identity across memoized calls)
- Split-card merge: `coloridentity` concatenates to `"RU"` and sorts to `"UR"`; both faces' color properties survive
- Set parsing: valid/invalid/empty JSON, set-type capitalization, name sort (mapping inverted so the assertion can't pass trivially)
- Card cleanup: Æ ligature, curly-apostrophe normalization, duplicate-name reuse, manacost brace stripping
- `cmc` coercion regression: numeric `convertedManaCost` reaches `cards.xml` as text

**Benchmarks (`oracle_importer_benchmark_test.cpp`)** stay manual, mirrored on the `carddatabase` load-benchmark pattern (not registered in CI): import/parse/split/colors timing with exact import-count assertions. The fixture uses real MTGJSON value types (floats, booleans).

**Parsing details fixed along the way:**
- Reserve the cards hash by distinct name, not printings (AllPrintings ≈ 35k names vs 100k printings)
- Constant `noCapitalize` list hoisted to `static`, `mid(1)` slice, `std::unique` for color dedup
- `coloridentity` concatenated for split faces like `colors` instead of gaining ` // `
- Card properties moved into `addCard` instead of copied

Builds on the native `QJson` importer; further stacked by the RAM benchmarks and lazy parsing.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -555,13 +555,17 @@ int OracleImporter::startImport()
     // ships ~100k printings for ~35k names, so reserving the printing count
     // would overallocate ~3x (against this stack's RAM goal). Collecting
     // distinct names is cheap — one pass over the already-parsed name fields.
-    QSet<QString> distinctNames;
-    for (const SetToDownload &curSetToParse : allSets) {
-        for (const QJsonValue &cardValue : curSetToParse.getCards()) {
-            distinctNames.insert(cardValue.toObject().value("name").toString());
+    {
+        QSet<QString> distinctNames;
+        for (const SetToDownload &curSetToParse : allSets) {
+            for (const QJsonValue &cardValue : curSetToParse.getCards()) {
+                distinctNames.insert(cardValue.toObject().value("name").toString());
+            }
         }
+        cards.reserve(distinctNames.size());
+        // The set goes out of scope here, handing the ~35k name QStrings back
+        // to the allocator before the (memory-heavy) import loop starts.
     }
-    cards.reserve(distinctNames.size());
 
     // add an empty set for tokens
     CardSetPtr tokenSet =
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -8,6 +8,7 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QRegularExpression>
+#include <QSet>
 #include <algorithm>
 #include <climits>
 #include <libcockatrice/card/database/parser/cockatrice_xml_4.h>
@@ -67,7 +68,8 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
         // capitalize set type
         if (setType.length() > 0) {
             // basic grammar for words that aren't capitalized, like in "From the Vault"
-            const QStringList noCapitalize = {"the", "a", "an", "on", "to", "for", "of", "in", "and", "with", "or"};
+            static const QStringList noCapitalize = {"the", "a",  "an",  "on",   "to", "for",
+                                                     "of",  "in", "and", "with", "or"};
             QStringList words = setType.split("_");
             setType.clear();
             bool first = false;
@@ -75,7 +77,7 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
                 if (first && noCapitalize.contains(item)) {
                     setType += item + QString(" ");
                 } else {
-                    setType += item[0].toUpper() + item.mid(1, -1) + QString(" ");
+                    setType += item[0].toUpper() + item.mid(1) + QString(" ");
                     first = true;
                 }
             }
@@ -123,14 +125,8 @@ static void sortAndReduceColors(QString &colors)
     std::sort(colors.begin(), colors.end(),
               [](const QChar a, const QChar b) { return colorOrder.value(a, INT_MAX) < colorOrder.value(b, INT_MAX); });
     // reduce
-    QChar lastChar = '\0';
-    for (int i = 0; i < colors.size(); ++i) {
-        if (colors.at(i) == lastChar) {
-            colors.remove(i, 1);
-        } else {
-            lastChar = colors.at(i);
-        }
-    }
+    auto last = std::unique(colors.begin(), colors.end());
+    colors.erase(last, colors.end());
 }
 
 CardInfoPtr OracleImporter::addCard(QString name,
@@ -186,8 +182,9 @@ CardInfoPtr OracleImporter::addCard(QString name,
 
     // DETECT CARD POSITIONING INFO
 
-    bool landscapeOrientation = properties.value("maintype") == "Battle" || properties.value("layout") == "split" ||
-                                properties.value("layout") == "planar";
+    QString layoutVal = properties.value("layout");
+    bool landscapeOrientation =
+        properties.value("maintype") == "Battle" || layoutVal == "split" || layoutVal == "planar";
 
     // cards that enter the field tapped
     bool cipt = parseCipt(name, text) || landscapeOrientation;
@@ -426,7 +423,8 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QJson
                 }
             }
 
-            CardInfoPtr newCard = addCard(name + numComponent, text, isToken, properties, relatedCards, printingInfo);
+            CardInfoPtr newCard =
+                addCard(name + numComponent, text, isToken, std::move(properties), relatedCards, printingInfo);
             numCards++;
         }
     }
@@ -459,7 +457,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QJson
                     if (!thisCardPropertyValue.isEmpty() && originalPropertyValue != thisCardPropertyValue) {
                         if (originalPropertyValue.isEmpty()) { // don't create //es if one field is empty
                             properties.insert(prop, thisCardPropertyValue);
-                        } else if (prop == "colors") { // the card is both colors
+                        } else if (prop == "colors" || prop == "coloridentity") { // the card is both colors
                             properties.insert(prop, originalPropertyValue + thisCardPropertyValue);
                         } else if (prop == "maintype") { // don't create maintypes with //es in them
                             continue;
@@ -471,7 +469,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QJson
                 }
             }
         }
-        CardInfoPtr newCard = addCard(name, text, isToken, properties, {}, printingInfo);
+        CardInfoPtr newCard = addCard(name, text, isToken, std::move(properties), {}, printingInfo);
         numCards++;
     }
 
@@ -551,6 +549,19 @@ const FormatRulesNameMap &OracleImporter::createDefaultMagicFormats()
 int OracleImporter::startImport()
 {
     static ICardSetPriorityController *noOpController = new NoopCardSetPriorityController();
+
+    // Pre-allocate the cards hash to avoid rehashing during import. The hash
+    // is keyed by distinct card name rather than by printings: AllPrintings
+    // ships ~100k printings for ~35k names, so reserving the printing count
+    // would overallocate ~3x (against this stack's RAM goal). Collecting
+    // distinct names is cheap — one pass over the already-parsed name fields.
+    QSet<QString> distinctNames;
+    for (const SetToDownload &curSetToParse : allSets) {
+        for (const QJsonValue &cardValue : curSetToParse.getCards()) {
+            distinctNames.insert(cardValue.toObject().value("name").toString());
+        }
+    }
+    cards.reserve(distinctNames.size());
 
     // add an empty set for tokens
     CardSetPtr tokenSet =
--- a/tests/oracle/CMakeLists.txt
+++ b/tests/oracle/CMakeLists.txt
@@ -7,3 +7,35 @@ endif()
 target_link_libraries(parse_cipt_test Threads::Threads ${GTEST_BOTH_LIBRARIES} ${TEST_QT_MODULES})
 
 add_test(NAME parse_cipt_test COMMAND parse_cipt_test)
+
+# Oracle importer unit tests
+add_executable(
+  oracle_importer_test ${VERSION_STRING_CPP} ../../oracle/src/oracleimporter.cpp ../../oracle/src/parsehelpers.cpp
+                       oracle_importer_test.cpp
+)
+
+if(NOT GTEST_FOUND)
+  add_dependencies(oracle_importer_test gtest)
+endif()
+
+target_link_libraries(
+  oracle_importer_test libcockatrice_card libcockatrice_interfaces Threads::Threads ${GTEST_BOTH_LIBRARIES}
+  ${TEST_QT_MODULES}
+)
+
+add_test(NAME oracle_importer_test COMMAND oracle_importer_test)
+
+# Oracle importer benchmark tests (manual, not run in CI)
+add_executable(
+  oracle_importer_benchmark_test ${VERSION_STRING_CPP} ../../oracle/src/oracleimporter.cpp
+                                 ../../oracle/src/parsehelpers.cpp oracle_importer_benchmark_test.cpp
+)
+
+if(NOT GTEST_FOUND)
+  add_dependencies(oracle_importer_benchmark_test gtest)
+endif()
+
+target_link_libraries(
+  oracle_importer_benchmark_test libcockatrice_card libcockatrice_interfaces Threads::Threads ${GTEST_BOTH_LIBRARIES}
+  ${TEST_QT_MODULES}
+)
--- a/tests/oracle/oracle_importer_benchmark_test.cpp
+++ b/tests/oracle/oracle_importer_benchmark_test.cpp
@@ -1,0 +1,264 @@
+#include "../../oracle/src/oracleimporter.h"
+
+#include "gtest/gtest.h"
+#include <QDebug>
+#include <QElapsedTimer>
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <libcockatrice/interfaces/noop_card_set_priority_controller.h>
+
+// Helper: build a synthetic MTGJSON-style JSON with the given number of sets and cards per set
+static QByteArray buildSyntheticData(int numSets, int cardsPerSet)
+{
+    QJsonObject dataObj;
+    for (int s = 0; s < numSets; ++s) {
+        QJsonArray cardsArray;
+        for (int c = 0; c < cardsPerSet; ++c) {
+            QJsonObject card;
+            card["name"] = QString("Card %1").arg(s * cardsPerSet + c);
+            card["text"] = "This is a test card with some rules text.";
+            card["layout"] = "normal";
+            card["manaCost"] = "{W}";
+            card["type"] = "Creature — Human";
+            card["power"] = "2";
+            card["toughness"] = "2";
+            card["colors"] = QJsonArray{"W"};
+            card["colorIdentity"] = QJsonArray{"W"};
+            card["types"] = QJsonArray{"Creature"};
+            // Real MTGJSON types: floats and booleans, not strings. This
+            // exercises the QVariant coercion in the property reader.
+            card["convertedManaCost"] = 1.0;
+            card["manaValue"] = 1.0;
+            card["isOnlineOnly"] = false;
+            card["isRebalanced"] = false;
+
+            QJsonObject legalities;
+            legalities["standard"] = "legal";
+            legalities["modern"] = "legal";
+            legalities["legacy"] = "legal";
+            legalities["vintage"] = "legal";
+            legalities["commander"] = "legal";
+            card["legalities"] = legalities;
+
+            QJsonObject identifiers;
+            identifiers["scryfallId"] = QString("id-%1-%2").arg(s).arg(c);
+            card["identifiers"] = identifiers;
+
+            // In AllPrintings, number and rarity are flat fields on the card
+            // object, exactly as set below.
+            card["number"] = QString::number(c + 1);
+            card["rarity"] = "common";
+
+            cardsArray.append(card);
+        }
+
+        QJsonObject setObj;
+        setObj["code"] = QString("T%1").arg(s, 2, 10, QChar('0'));
+        setObj["name"] = QString("Test Set %1").arg(s);
+        setObj["type"] = "expansion";
+        setObj["releaseDate"] = "2024-01-01";
+        setObj["cards"] = cardsArray;
+
+        dataObj[QString("T%1").arg(s, 2, 10, QChar('0'))] = setObj;
+    }
+
+    QJsonObject root;
+    root["data"] = dataObj;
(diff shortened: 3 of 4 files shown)


tooomm:
Are there specific issue tickets that you're addressing with some of the fixes?

BruebachL:
> Are there specific issue tickets that you're addressing with some of the fixes?

No, they're just sort of incidental issues I discovered.

oracle/src/oracleimporter.cpp
```diff
@@ -584,4 +590,6 @@ void OracleImporter::clear()
     sets.clear();
     cards.clear();
     allSets.clear();
+    // Note: createDefaultMagicFormats() uses a function-local static cache that is
```
RickyRister: remove this before merging if you change it in the previous PR

RickyRister (approved):

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+            card["text"] = "This is a test card with some rules text.";
+            card["layout"] = "normal";
+            card["manaCost"] = "{W}";
+            card["type"] = "Creature — Human";
+            card["power"] = "2";
+            card["toughness"] = "2";
+            card["colors"] = QJsonArray{"W"};
+            card["colorIdentity"] = QJsonArray{"W"};
+            card["types"] = QJsonArray{"Creature"};
+            card["convertedManaCost"] = "1";
```
ZeldaZach: `"1"` here is a JSON string. In real MTGJSON, `convertedManaCost` and `manaValue` are floats (`3.0`), and `isOnlineOnly`/`isRebalanced` are booleans.

That matters because #7214 changed the property reader from `QVariant::toString()` (which coerces numbers and bools) to `QJsonValue::toString()` (which returns a null `QString` for anything that isn't a JSON string). I confirmed this against Qt: `manaValue: 3.0` yields `""` on the new path and `"3"` on the old one, so `properties["cmc"]` is never inserted for any real card. With a string in the fixture, the tests exercise a path that real data never takes and the regression stays green.

Neither test file mentions `cmc`, `manaValue`, `faceManaValue` or `faceConvertedManaCost`, and no fixture anywhere sets `isOnlineOnly` or `isRebalanced`, so there is nothing else covering it either.

Suggest making the fixture match the real schema and asserting on the result:

```cpp
card["manaValue"] = 3;               // number, as in AllPrintings
card["isOnlineOnly"] = true;         // boolean
```

plus a case in `oracle_importer_test.cpp` asserting `getProperty("cmc") == "3"`. That single assertion would have caught the #7214 issue, and it's the highest-value test this PR could add given what the stack is doing.

(`identifiers.multiverseId`/`scryfallId` are genuinely strings in MTGJSON v5, so those fixtures are fine as-is.)

tests/oracle/oracle_importer_test.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+    QJsonObject root;
+    root["data"] = QJsonObject{{"FTV", setObj}};
+
+    QByteArray data = QJsonDocument(root).toJson();
+    ASSERT_TRUE(importer->readSetsFromByteArray(data));
+    ASSERT_EQ(importer->getSets().first().getSetType(), "From the Vault");
+}
+
+TEST_F(OracleImporterTest, ReadSetsFromByteArraySortsSetsByName)
```
ZeldaZach, line 342: This one passes with the sort deleted. `QJsonObject` iterates keys in lexicographic order, so `{"ZZZ": setA, "AAA": setB}` is already walked as `AAA, ZZZ`, and since `AAA` maps to `"Alpha Set"` and `ZZZ` to `"Zeta Set"`, `newSetList` is in sorted order before `std::sort` at `oracleimporter.cpp:88` ever runs.

Inverting the mapping makes the assertion meaningful:

```cpp
setA["code"] = "aaa";  setA["name"] = "Zeta Set";    // under key "AAA"
setB["code"] = "zzz";  setB["name"] = "Alpha Set";   // under key "ZZZ"
...
ASSERT_EQ(sets.first().getShortName(), "ZZZ");
```

Now iteration order and sorted order disagree, and removing the `std::sort` fails the test.

tests/oracle/oracle_importer_test.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+    // Verify the regex actually works (was broken before: \b = backspace, not word boundary)
+    QRegularExpression regex(condition.value);
+    ASSERT_TRUE(regex.isValid());
+    ASSERT_TRUE(regex.match("Basic Land — Forest").hasMatch());
+    ASSERT_TRUE(regex.match("Basic Snow Land — Mountain").hasMatch());
+    ASSERT_FALSE(regex.match("Creature — Elf Warrior").hasMatch());
+}
+
+TEST_F(OracleImporterTest, CreateDefaultMagicFormatsCaching)
```
ZeldaZach, line 278: This asserts determinism, not caching. `buildDefaultMagicFormats()` builds the same map every time, so `size()` and `keys()` match whether or not the result is memoized. Drop the `static` from `oracleimporter.cpp:533` and the test still passes.

The observable effect of the cache is pointer identity of the shared `FormatRulesPtr` values:

```cpp
ASSERT_EQ(first.value("standard").data(), second.value("standard").data());
```

That fails without the cache and passes with it. It also documents the aliasing this introduces, which is the part worth pinning down.

tests/oracle/oracle_importer_test.cpp
```diff
@@ -0,0 +1,495 @@
...
+    ASSERT_TRUE(formats.contains("historic"));
+    ASSERT_TRUE(formats.contains("timeless"));
+    ASSERT_TRUE(formats.contains("duel"));
+    ASSERT_TRUE(formats.contains("oathbreaker"));
+}
+
+TEST_F(OracleImporterTest, CreateDefaultMagicFormatsSingletonDeckSizes)
+{
+    auto formats = importer->createDefaultMagicFormats();
+    auto commander = formats.value("commander");
```
ZeldaZach, line 229: `FormatRulesNameMap` is a `QHash<QString, QSharedPointer<FormatRules>>`, so `formats.value("commander")` returns a null shared pointer when the key is absent and `commander->minDeckSize` dereferences null.

The practical cost is diagnostic rather than correctness: rename `"Commander"` in `buildDefaultMagicFormats` and instead of one failing assertion you get a SIGSEGV that takes down the gtest binary and suppresses the results of every other test in the file.

`ASSERT_FALSE(commander.isNull());` before the first deref fixes it. Same pattern applies at the `brawl`, `vintage` and `standard` lookups below.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+    // Phase 1: Parse JSON
+    QElapsedTimer timer;
+    timer.start();
+    bool ok = importer.readSetsFromByteArray(data);
+    ASSERT_TRUE(ok);
+    qint64 parseMs = timer.elapsed();
+
+    // Phase 2: Import cards
+    timer.restart();
+    int importedSets = importer.startImport();
```
ZeldaZach, line 93: `importedSets` and `totalImported` are only ever fed to `qDebug()`, so this test asserts nothing about the import. Break `importCardsFromSet` so it imports zero cards and this reports "10 sets, 0 unique cards" and passes.

Since the fixture generates globally unique card names, the expected values are exact:

```cpp
ASSERT_EQ(importedSets, numSets);
ASSERT_EQ(totalImported, numSets * cardsPerSet);
```

Cheap to add and turns a timing probe into a real regression test.

tests/oracle/CMakeLists.txt
```diff
@@ -7,3 +7,38 @@ endif()
...
+if(NOT GTEST_FOUND)
+  add_dependencies(oracle_importer_benchmark_test gtest)
+endif()
+
+target_link_libraries(
+  oracle_importer_benchmark_test libcockatrice_card libcockatrice_interfaces Threads::Threads ${GTEST_BOTH_LIBRARIES}
+  ${TEST_QT_MODULES}
+)
+
+add_test(NAME oracle_importer_benchmark_test COMMAND oracle_importer_benchmark_test)
```
ZeldaZach: This registers the benchmarks as CI tests, which cuts against the pattern already established for benchmarks in this repo. `tests/carddatabase/CMakeLists.txt` builds `load_benchmark` under an explicit `# Load Benchmark (manual, not run in CI)` heading and deliberately omits `add_test`.

The concern is the 30 s `TIMEOUT` on the next line combined with `.github/workflows/desktop-build.yml:174`, which builds Linux with `--debug`. The dominant cost is an `-O0` import of 5000 + 2000 + 10000 cards, and it's unbounded: a loaded runner blows the budget and CI goes red on a PR that changed nothing relevant. I measured the fixture-building and parsing portion at roughly 0.6 s on an M-series Mac, so the 30 s isn't obviously wrong, just fragile in the wrong direction.

Either drop the `add_test`/`set_tests_properties` lines to match `load_benchmark`, or keep them and remove the timeout.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+            legalities["legacy"] = "legal";
+            legalities["vintage"] = "legal";
+            legalities["commander"] = "legal";
+            card["legalities"] = legalities;
+
+            QJsonObject identifiers;
+            identifiers["scryfallId"] = QString("id-%1-%2").arg(s).arg(c);
+            card["identifiers"] = identifiers;
+
+            QJsonObject numObj;
```
ZeldaZach: `numObj` is populated and then never used, and the comment below it has it backwards: in `AllPrintings` these are flat fields on the card object, which is exactly what lines 47-48 already do. Lines 43-46 can go.

Same file, lines 79, 122, 189 and 216: `NoopCardSetPriorityController controller;` is constructed and never used in any of the four benchmarks. No `-Wunused-variable` because it has a non-trivial destructor, so it stays silent.

tests/oracle/oracle_importer_test.cpp
```diff
@@ -0,0 +1,495 @@
...
+    auto card = importer->getCardList().value("Single Card");
+    ASSERT_FALSE(card.isNull());
+    ASSERT_EQ(card->getProperty("colors"), "B");
+}
+
+// ============================================================================
+// Legality guard tests
+// ============================================================================
+
+TEST_F(OracleImporterTest, LegalityGuardCombinesForNewCard)
```
ZeldaZach: The name doesn't match what this covers. A brand-new card never reaches `combineLegalities` at all: `oracleimporter.cpp:141-149` only calls it on the existing-card branch. This test verifies that `format-*` properties survive `addCard`, which is worth having but isn't the guard.

The uncovered case is the one the guard is meant to allow through: an existing card whose properties contain no `format-*` entry (a first printing with an empty or absent `legalities` object), which should still merge. `LegalityGuardPreservesFirstPrinting` below covers the blocking half and genuinely fails if the `if` is removed, so adding the merging half would close the pair.

Also worth a rename to something like `FormatPropertiesSurviveAddCard`.

oracle/src/oracleimporter.cpp
```diff
@@ -446,7 +445,7 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QJson
                     if (!thisCardPropertyValue.isEmpty() && originalPropertyValue != thisCardPropertyValue) {
                         if (originalPropertyValue.isEmpty()) { // don't create //es if one field is empty
                             properties.insert(prop, thisCardPropertyValue);
-                        } else if (prop == "colors") { // the card is both colors
+                        } else if (prop == "colors" || prop == "coloridentity") { // the card is both colors
```
ZeldaZach, line 460: Good fix. Worth noting for anyone reading this later that it's safe specifically because `addCard` runs `sortAndReduceColors` over `coloridentity` afterwards (lines 170-175), so the naive concatenation gets sorted and deduplicated before it reaches the card.

The test covering this (`oracle_importer_test.cpp:387`) asserts only that the result has no `//` and contains both `R` and `U`, so a merge producing `"RUR"` or `"URUR"` would pass. The exact expected value is `"UR"`; `ASSERT_EQ(ci, "UR")` would pin both the concatenation and the dedup.

oracle/src/oracleimporter.cpp
```diff
@@ -539,6 +538,13 @@ int OracleImporter::startImport()
 {
     static ICardSetPriorityController *noOpController = new NoopCardSetPriorityController();
 
+    // Pre-allocate cards hash to avoid rehashing during import
+    int estimatedCards = 0;
+    for (const SetToDownload &curSetToParse : allSets) {
+        estimatedCards += curSetToParse.getCards().size();
+    }
+    cards.reserve(estimatedCards);
```
ZeldaZach: `estimatedCards` sums `getCards().size()` across all sets, which counts printings, not distinct cards. On real `AllPrintings` that is roughly 100k printings against about 35k unique names (your own #7217 measurement says 35.5k), so this reserves around 3x the buckets actually needed.

Not a correctness problem, and the rehashing it avoids is real, but in a stack whose stated goal is cutting peak memory it's worth being deliberate about. Scaling the estimate down, or reserving against a distinct-name count if one is cheaply available, gets most of the benefit without the overshoot.

Minor: `int` is fine at today's volumes but `qsizetype` would match `QHash::reserve` and the `size()` calls being summed.

ZeldaZach (changes requested):
Getting the oracle importer under test at all is a big step, and several of these are real regression tests: `LegalityGuardPreservesFirstPrinting` fails if the guard is removed, and `CreateDefaultMagicFormatsRegexMatchesBasicLands` fails against the old literal-backspace value. The split-card and color-reduction coverage is welcome too.

The `std::unique` rewrite of `sortAndReduceColors` is also a straight bug fix, not just a simplification. The old loop did `colors.remove(i, 1)` and then `++i`, which skips the character shifted into position `i`, so `"RRRR"` reduced to `"RR"` rather than `"R"`. Worth mentioning in the PR description.

Two things to sort out. First, the fixtures use JSON strings where real MTGJSON uses numbers and booleans, which means this suite is blind to the one genuine regression in #7214 (see my comment there about `cmc`). Second, a few tests pass whether or not the code under test works. Details inline.

Small note on the description: it credits this PR with the Basic Land regex fix, but that change is actually in #7214.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Fixture now matches real MTGJSON types: `convertedManaCost`/`manaValue` are floats and `isOnlineOnly`/`isRebalanced` are booleans, exercising the QVariant-coercion path (#7214). `ImportThroughput` also asserts `cmc == "1"` for the probe card, so the coercion is pinned.

tests/oracle/oracle_importer_test.cpp
BruebachL: Inverted the mapping as suggested: `"AAA"` now maps to a "Zeta Set" and `"ZZZ"` to "Alpha Set", so natural lexicographic iteration would yield the wrong order and the test only passes because the name sort promotes `ZZZ` to the front.

tests/oracle/oracle_importer_test.cpp
BruebachL: Switched the caching test to the observable effect: `ASSERT_EQ(first.value("standard").data(), second.value("standard").data())`, which fails without the memoization and documents the shared-pointer aliasing.

tests/oracle/oracle_importer_test.cpp
BruebachL: Added `ASSERT_FALSE(...isNull())` before every dereference (`commander`, `brawl`, `vintage`, `standard`), so a mapping typo fails one assertion instead of segfaulting the whole binary and masking other tests.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Now asserts the import actually happened: `ASSERT_EQ(importedSets, numSets)` and `ASSERT_EQ(totalImported, numSets * cardsPerSet)` — the totals are exact since the fixture generates globally unique (set, name) combinations, so the timing probe doubles as a regression test.

tests/oracle/CMakeLists.txt
BruebachL: Benchmark binary is no longer registered with `add_test`; `tests/oracle/CMakeLists.txt` now mirrors the `carddatabase` "manual, not run in CI" pattern, dropping the 30 s CI TIMEOUT.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Removed the dead `numObj` block and fixed the flipped comment (number/rarity are flat fields in AllPrintings, as the code already did). Dropped the unused `NoopCardSetPriorityController` in `ImportThroughput` and `ParseJsonThroughput`; the ones in `SplitCardMerging`/`ImportCardsWithColors` are actually passed to `CardSet::newInstance`, so they stay.

tests/oracle/oracle_importer_test.cpp
BruebachL: Youre right that the name claimed the guard; renamed to `NewCardKeepsLegalityProperties`, and added `LegalityMergeAllowedWhenCardHasNoLegalities`: a first printing with no legalities object must still let a later printings legalities merge in. Both halves of the guard are now pinned.

oracle/src/oracleimporter.cpp
BruebachL: Pinned the exact value: `ASSERT_EQ(card->getProperty("coloridentity"), "UR")` — concatenation to "RU", then WUBRG sort in addCard to "UR". A merge producing duplicates like "URUR" now fails.

oracle/src/oracleimporter.cpp
BruebachL: Reserve now targets distinct card names instead of printings — one cheap pass collecting raw name fields into a QSet (~35k for AllPrintings) — so the hash pre-allocation matches what `cards` is actually keyed by instead of tripling the estimate.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    // The fixture generates globally unique card names, so the expected
+    // counts are exact: a regression here means cards were dropped.
+    ASSERT_EQ(importedSets, numSets);
+    ASSERT_EQ(totalImported, numSets * cardsPerSet);
+    // Real-data probe: numeric convertedManaCost must be coerced to text
+    // (regression for the QJsonValue::toString() reader in #7214).
+    auto probeCard = importer.getCardList().value("Card 0");
+    ASSERT_FALSE(probeCard.isNull());
+    ASSERT_EQ(probeCard->getProperty("cmc"), "1");
```
ZeldaZach, line 110: This is the assertion that pins the #7214 regression, but it now lives in the binary that no longer gets an add_test, so ctest never runs it. oracle_importer_test.cpp has no reference to cmc, manaValue or convertedManaCost, and makeCard() sets neither, so in practice the coercion path is unguarded in CI.

The two requests from my last pass collided here and that is on me. The fix is small: add the same case to oracle_importer_test.cpp, setting `card["convertedManaCost"] = 3.0` on a makeCard() result and asserting `getProperty("cmc") == "3"`. Keep this one too if you like, it costs nothing.

oracle/src/oracleimporter.cpp
```diff
@@ -552,6 +550,19 @@ int OracleImporter::startImport()
...
+    // ships ~100k printings for ~35k names, so reserving the printing count
+    // would overallocate ~3x (against this stack's RAM goal). Collecting
+    // distinct names is cheap — one pass over the already-parsed name fields.
+    QSet<QString> distinctNames;
+    for (const SetToDownload &curSetToParse : allSets) {
+        for (const QJsonValue &cardValue : curSetToParse.getCards()) {
+            distinctNames.insert(cardValue.toObject().value("name").toString());
+        }
+    }
+    cards.reserve(distinctNames.size());
```
ZeldaZach: distinctNames is declared at function scope, so the ~35k QStrings stay alive for the whole of startImport even though the only consumer is this line. Wrapping the pass in a bare block, or calling clear() right after the reserve, hands that memory back before the import starts, which matters more than usual given what this stack is for.

Separately, the estimate skews low in one direction worth knowing about: cards carrying a `side` are keyed by faceName in addCard, so a DFC that appears here once as "A // B" inserts two entries into cards. Fine for a reserve hint, just not the exact figure the comment implies.

ZeldaZach (approved):
Re-reviewed after the 21:06 push. All nine items from my previous pass are addressed, and addressed properly: the set-name sort test is no longer order-trivial, the formats cache test asserts pointer identity, the null guards are in front of every deref, the benchmark asserts exact import counts, the benchmark binary is off ctest, the split coloridentity assertion pins "UR", the dead numObj block and unused controllers are gone, the legality guard test is renamed and the merging half added, and the reserve is keyed on distinct names.

I also checked the parsing changes against the surrounding code. The std::unique rewrite is a genuine fix rather than a simplification: the old remove/++i loop reduced "WWUUB" to "WUUB", and SortAndReduceColorsDeduplicates fails against it. The coloridentity concatenation is safe because addCard runs sortAndReduceColors over it afterwards, and the identical-value and empty-value branches above it mean a same-color split card never doubles up. Every fixture is built in process, nothing touches the network. CI is green on all 18 checks including the jobs that actually run ctest (Debian 13, Fedora 44, Ubuntu 26.04, macOS, Windows).

Approving. Two non-blocking notes inline. I would like the first one picked up, in a follow-up if you prefer, since as it stands the #7214 coercion has no CI coverage.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Fixed in `bd1fb69800`. Added `NumericManaValueCoercedToCmc` (`manaValue = 3`, asserts `cmc == "3"`) and `LegacyConvertedManaCostCoercedToCmc` (`convertedManaCost = 3.0`, asserts `cmc == "3"`) to `oracle_importer_test.cpp`, which IS registered with `add_test` and runs under ctest. The benchmark assertion stays too. The coercion regression in #7214 is now pinned in CI.

oracle/src/oracleimporter.cpp
BruebachL: Fixed in `bd1fb69800`. The distinct-name collection pass is now wrapped in a bare block, so the `QSet<QString>` (~35k entries on AllPrintings) goes out of scope and `cards.reserve(...)` is the only surviving effect — the name strings are handed back before the import loop runs. Kept the comment noting it might still overcount slightly for DFCs keyed by faceName (each "A // B" inserts two entries), which is fine for a reserve hint.

PR #7216, opened 2026-08-29
[Oracle] Add RAM usage benchmarks for the oracle importer

BruebachL:
Add opt-in RAM usage benchmarks for the oracle importer so future parse changes are verifiable. Kept **manual tools instead of CI-registered tests** (mirroring `tests/carddatabase`'s `load_benchmark`), so there's no timing/memory budget to flake CI.

- Measure process peak plus per-phase **current RSS** (VmHWM/VmRSS on Linux, `ru_maxrss` + `task_info().resident_size` on macOS); skipped on platforms without memory stats (Windows) before the synthetic fixture is built
- Synthetic multi-set download/parse/import fixtures; baseline is taken *before* the fixture build so the importer phases aren't masked by the builder's own high-water mark
- Optional real `AllPrintings` run: downloaded + decompressed behind `COCKATRICE_ORACLE_RAM_BENCHMARK=1` (an explicit `=0` disables it), with a 10-min in-code timeout that aborts the reply, and rejection of HTML-error bodies before parsing
- Cover both ZLIB and LibLZMA decompression paths in the test binaries

Builds on the importer tests; stacked below the lazy-parsing change so its memory win can be quantified.

### DIFF
--- a/tests/oracle/CMakeLists.txt
+++ b/tests/oracle/CMakeLists.txt
@@ -25,10 +25,33 @@ target_link_libraries(
 
 add_test(NAME oracle_importer_test COMMAND oracle_importer_test)
 
-# Oracle importer benchmark tests (manual, not run in CI)
+# Oracle importer benchmark tests (manual, not run in CI, incl. RAM benchmark)
+# Optional compression libs, mirrored from oracle/CMakeLists.txt, so the benchmark
+# can download and decompress whatever AllPrintings format the default URL selects.
+find_package(ZLIB)
+if(ZLIB_FOUND)
+  add_definitions("-DHAS_ZLIB")
+  set(_ORACLE_BENCH_EXTRA_SOURCES ../../oracle/src/zip/unzip.cpp ../../oracle/src/zip/zipglobal.cpp)
+  set(_ORACLE_BENCH_EXTRA_LIBRARIES ${ZLIB_LIBRARIES})
+  include_directories(${ZLIB_INCLUDE_DIRS})
+else()
+  message(STATUS "Oracle tests: zlib not found; zip download benchmark disabled")
+endif()
+
+find_package(LibLZMA)
+if(LIBLZMA_FOUND)
+  add_definitions("-DHAS_LZMA")
+  list(APPEND _ORACLE_BENCH_EXTRA_SOURCES ../../oracle/src/lzma/decompress.cpp)
+  list(APPEND _ORACLE_BENCH_EXTRA_LIBRARIES ${LIBLZMA_LIBRARIES})
+  include_directories(${LIBLZMA_INCLUDE_DIRS})
+else()
+  message(STATUS "Oracle tests: LibLZMA not found; xz download benchmark disabled")
+endif()
+
 add_executable(
-  oracle_importer_benchmark_test ${VERSION_STRING_CPP} ../../oracle/src/oracleimporter.cpp
-                                 ../../oracle/src/parsehelpers.cpp oracle_importer_benchmark_test.cpp
+  oracle_importer_benchmark_test
+  ${VERSION_STRING_CPP} ../../oracle/src/oracleimporter.cpp ../../oracle/src/parsehelpers.cpp
+  oracle_importer_benchmark_test.cpp ${_ORACLE_BENCH_EXTRA_SOURCES}
 )
 
 if(NOT GTEST_FOUND)
@@ -36,6 +59,11 @@ if(NOT GTEST_FOUND)
 endif()
 
 target_link_libraries(
-  oracle_importer_benchmark_test libcockatrice_card libcockatrice_interfaces Threads::Threads ${GTEST_BOTH_LIBRARIES}
+  oracle_importer_benchmark_test
+  libcockatrice_card
+  libcockatrice_interfaces
+  Threads::Threads
+  ${GTEST_BOTH_LIBRARIES}
   ${TEST_QT_MODULES}
+  ${_ORACLE_BENCH_EXTRA_LIBRARIES}
 )
--- a/tests/oracle/oracle_importer_benchmark_test.cpp
+++ b/tests/oracle/oracle_importer_benchmark_test.cpp
@@ -1,12 +1,32 @@
 #include "../../oracle/src/oracleimporter.h"
 
 #include "gtest/gtest.h"
+#include <QBuffer>
+#include <QCoreApplication>
 #include <QDebug>
 #include <QElapsedTimer>
+#include <QEventLoop>
+#include <QFile>
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QNetworkRequest>
+#include <QTimer>
+#include <QUrl>
 #include <libcockatrice/interfaces/noop_card_set_priority_controller.h>
+
+#if defined(HAS_LZMA)
+#include "../../oracle/src/lzma/decompress.h"
+#endif
+#if defined(HAS_ZLIB)
+#include "../../oracle/src/zip/unzip.h"
+#endif
+#if defined(Q_OS_MACOS)
+#include <mach/mach.h>
+#include <sys/resource.h>
+#endif
 
 // Helper: build a synthetic MTGJSON-style JSON with the given number of sets and cards per set
 static QByteArray buildSyntheticData(int numSets, int cardsPerSet)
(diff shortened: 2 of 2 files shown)


tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+}
+
+TEST(OracleBenchmark, ImportRamUsage)
+{
+    static constexpr int numSets = 30;
+    static constexpr int cardsPerSet = 2000; // ~60k cards, roughly AllPrintings scale
+
+    const QByteArray data = buildSyntheticData(numSets, cardsPerSet);
+
+    const MemorySnapshot baseline = MemorySnapshot::current();
```
ZeldaZach: The baseline is taken after `buildSyntheticData()` has already run, so every subsequent "peak added" is measured against a high-water mark that the fixture builder set, not the importer.

I compiled this `MemorySnapshot`/`logRamPhase` code verbatim and ran `buildSyntheticData(30, 2000)` against Qt 6.11.1:

```
after buildSyntheticData(30,2000): 938 ms  json 24.29 MB | cur RSS 237.3 MB | PEAK 265.1 MB
after parse:                       108 ms             | cur RSS 277.8 MB | PEAK 277.8 MB
  -> this benchmark reports "peak added" = 12.7 MB
```

Constructing the transient `QJsonObject` tree spikes the process to 265 MB before the baseline is read. The parse then genuinely grows RSS by about 40 MB but gets reported as 12.7 MB, and any phase that stays under the pre-existing 265 MB mark reports a flat 0.0 MB.

It also depends on gtest ordering: the four benchmarks earlier in this binary have already raised `VmHWM` (`ParseJsonThroughput` alone builds 20k cards five times), so `--gtest_filter` or `--gtest_shuffle` changes the numbers.

Two options: build the fixture, free the intermediate `QJsonObject`s, and re-read the baseline after the spike has passed; or fork a child process per measurement so each starts from a clean high-water mark. The first is much simpler and enough to make the deltas meaningful.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+    const int totalCards = importer.getCardList().size();
+    qDebug().noquote() << QString("Oracle RAM Benchmark (synthetic): %1 sets, %2 cards, %3 MB JSON")
+                              .arg(importedSets)
+                              .arg(totalCards)
+                              .arg(data.size() / (1024.0 * 1024.0), 0, 'f', 1);
+    qDebug().noquote() << QString("  JSON parse: %1 ms").arg(parseMs);
+    qDebug().noquote() << QString("  Card import: %1 ms").arg(importMs);
+    logRamPhase("parse", baseline, afterParse);
+    logRamPhase("import", afterParse, afterImport);
+    logRamPhase("after releaseSetData()", afterImport, afterRelease);
```
ZeldaZach, line 472: This phase can't show what it's named for. `logRamPhase` reports `current.peakRssKb - baseline.peakRssKb`, and both `VmHWM` and `ru_maxrss` are monotonically non-decreasing high-water marks, so the delta is always >= 0 and in practice exactly 0.0 MB. Releasing memory is unobservable through a peak counter by construction.

The only column that could show a drop is current RSS, which is Linux-only (see my note on line 309) and is itself misleading here: in my run above, current RSS sat at 237 MB while roughly 24 MB of data was live, because glibc hadn't returned the freed arenas to the OS. Without a `malloc_trim(0)` before the snapshot, a correct `releaseSetData()` and a broken one look identical.

If the goal is to prove the buffers are actually dropped, asserting on `importer.getRawSetsData().isEmpty()` and `getSets().isEmpty()` is a far more direct signal than an RSS delta. Keep the RSS line as colour if you like, but it shouldn't be the evidence.

tests/oracle/CMakeLists.txt
```diff
@@ -25,20 +25,48 @@ target_link_libraries(
...
+  libcockatrice_interfaces
+  Threads::Threads
+  ${GTEST_BOTH_LIBRARIES}
   ${TEST_QT_MODULES}
+  ${_ORACLE_BENCH_EXTRA_LIBRARIES}
 )
 
 add_test(NAME oracle_importer_benchmark_test COMMAND oracle_importer_benchmark_test)
-set_tests_properties(oracle_importer_benchmark_test PROPERTIES TIMEOUT 30)
+set_tests_properties(oracle_importer_benchmark_test PROPERTIES TIMEOUT 120)
```
ZeldaZach: This 120 s budget is five times shorter than the benchmark's own download allowance:

```cpp
timeoutTimer.start(10 * 60 * 1000);   // line 449
```

So the documented workflow (set `COCKATRICE_ORACLE_RAM_BENCHMARK=1`, run `ctest`) can never honour the in-code timeout. ctest kills the binary at 120 s partway through downloading, decompressing ~500 MB of AllPrintings, or importing, and reports a hard failure.

This is also now by far the longest-permitted test in the suite (`deck_hash_performance_test` is 15 s).

The cleanest fix is to stop registering the benchmark with `add_test` at all, which is what `tests/carddatabase/CMakeLists.txt` already does for `load_benchmark` under an explicit "manual, not run in CI" heading. If you'd rather keep it registered, the two timeouts need to agree.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+    request.setHeader(QNetworkRequest::UserAgentHeader, "Cockatrice Oracle RAM benchmark");
+    QNetworkReply *reply = nam.get(request);
+
+    QEventLoop loop;
+    QTimer timeoutTimer;
+    timeoutTimer.setSingleShot(true);
+    QObject::connect(reply, &QNetworkReply::finished, &loop, &QEventLoop::quit);
+    QObject::connect(&timeoutTimer, &QTimer::timeout, &loop, &QEventLoop::quit);
+    timeoutTimer.start(10 * 60 * 1000);
+    loop.exec();
```
ZeldaZach, line 514: When the timer fires and quits the loop, `reply` is never aborted, so `reply->error()` is still `NoError` on line 453 and the `GTEST_SKIP` below doesn't trigger. `readAll()` then returns a truncated body, `readSetsFromByteArray` rejects it, and `ASSERT_TRUE` on line 471 fails.

Same outcome for a 200 response carrying HTML instead of JSON, which is a realistic mtgjson failure mode. `pages.cpp:476` explicitly handles the `<` case; this doesn't.

The net effect is inconsistent: a network *error* skips cleanly, but a *timeout* or a wrong-content response fails the test. Since this is an opt-in benchmark, neither should be a failure.

```cpp
bool timedOut = false;
QObject::connect(&timeoutTimer, &QTimer::timeout, [&] { timedOut = true; reply->abort(); });
...
if (timedOut || reply->error() != QNetworkReply::NoError) {
    GTEST_SKIP() << "Download failed: " << reply->errorString().toStdString();
}
```

A `payload.startsWith('{') || payload.startsWith('\xFD')`-style sanity check before the import would cover the bad-content case too.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+            for (const QByteArray &line : lines) {
+                if (line.startsWith("VmHWM:")) {
+                    snap.peakRssKb = line.mid(6).trimmed().split(' ').value(0).toLongLong();
+                } else if (line.startsWith("VmRSS:")) {
+                    snap.rssKb = line.mid(6).trimmed().split(' ').value(0).toLongLong();
+                }
+            }
+            snap.available = snap.peakRssKb >= 0;
+        }
+#elif defined(Q_OS_MACOS)
```
ZeldaZach, line 320: `rssKb` is never assigned on the macOS branch, only `peakRssKb`. Since `available` is set to true, the "unavailable" early-out in `logRamPhase` doesn't fire and `formatKb(-1)` prints `N/A` for current RSS on every line. I compiled and ran this exact branch on macOS:

```
  x: current RSS N/A | peak added 0.0 MB | process peak 14.1 MB
```

Combined with the peak-is-monotonic issue above, that leaves the four macOS CI jobs with a memory benchmark whose only live datum is a monotonically rising process peak.

`task_info(mach_task_self(), MACH_TASK_BASIC_INFO, ...)` gives the current resident size on Darwin if you want parity with the Linux branch. The `ru_maxrss / 1024` conversion is right, incidentally — `ru_maxrss` really is bytes on Darwin, unlike Linux where it's already kB.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+struct MemorySnapshot
+{
+    qint64 peakRssKb = -1; // process high-water mark (VmHWM on Linux, ru_maxrss on macOS)
+    qint64 rssKb = -1;     // current resident set size
+    bool available = false;
+
+    static MemorySnapshot current()
+    {
+        MemorySnapshot snap;
+#if defined(Q_OS_LINUX)
```
ZeldaZach, line 306: No Windows arm in this chain, so `MemorySnapshot::current()` returns `{-1, -1, false}` there. That part is handled correctly: `logRamPhase` takes the early-out and prints "memory stats unavailable on this platform", nothing asserts, nothing reports a bogus 0.

The cost is that there's no `GTEST_SKIP` either, so the Windows CI job generates and imports the full 60k synthetic cards to print three "unavailable" lines. `ImportRamUsage` triples the workload of the heaviest existing benchmark (60k cards vs 20k) and adds roughly 265 MB of peak just building fixtures, on every job that runs tests.

Either skip early when `!baseline.available`, or wire up Windows — `GetProcessMemoryInfo` with `PROCESS_MEMORY_COUNTERS::PeakWorkingSetSize` and `WorkingSetSize` maps cleanly onto both fields here.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+    qDebug().noquote() << QString("  JSON parse: %1 ms").arg(parseMs);
+    qDebug().noquote() << QString("  Card import: %1 ms").arg(importMs);
+    logRamPhase("parse", baseline, afterParse);
+    logRamPhase("import", afterParse, afterImport);
+    logRamPhase("after releaseSetData()", afterImport, afterRelease);
+}
+
+TEST(OracleBenchmark, ImportRamUsageAllPrintings)
+{
+    if (qEnvironmentVariableIsEmpty("COCKATRICE_ORACLE_RAM_BENCHMARK")) {
```
ZeldaZach: `qEnvironmentVariableIsEmpty` returns false for `"0"`, so `COCKATRICE_ORACLE_RAM_BENCHMARK=0` enables the download rather than disabling it, while the skip message on the next line documents `=1`.

I grepped `.github/workflows/` and `.ci/` and the variable isn't set anywhere, so this can't fire in CI as things stand. But someone setting it to `0` to be explicit gets a ~25 MB download and a multi-GB import inside the 120 s ctest budget.

`qEnvironmentVariableIntValue("COCKATRICE_ORACLE_RAM_BENCHMARK") != 1` matches the documented contract.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+    qDebug().noquote() << QString("Oracle RAM Benchmark (real AllPrintings): %1 sets, %2 unique cards")
+                              .arg(importedSets)
+                              .arg(totalCards);
+    qDebug().noquote() << QString("  URL: %1").arg(kDefaultAllPrintingsUrl.toDisplayString());
+    qDebug().noquote() << QString("  Downloaded: %1 MB, decompressed: %2 MB")
+                              .arg(payload.size() / (1024.0 * 1024.0), 0, 'f', 1)
+                              .arg(setsData.size() / (1024.0 * 1024.0), 0, 'f', 1);
+    qDebug().noquote() << QString("  JSON parse: %1 ms").arg(parseMs);
+    qDebug().noquote() << QString("  Card import: %1 ms").arg(importMs);
+    logRamPhase("download+decompress", baseline, afterDownload);
```
ZeldaZach, line 565: This label overstates what's measured. The baseline on line 459 is taken after `reply->readAll()` has already materialized the payload (and after `QNetworkReply`'s own internal buffer has held a second copy), so this line covers decompression only.

For a 500 MB dataset the download buffering is one of the phases most worth knowing about, so either move the baseline above the `QNetworkAccessManager` work or rename the phase to `decompress`.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -247,8 +266,240 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+        return out;
+#else
+        qDebug() << "RAM benchmark: download is zip-compressed but this build has no zlib support";
+        return {};
+#endif
+    }
+    return payload;
+}
+
+TEST(OracleBenchmark, ImportRamUsage)
```
ZeldaZach, line 423: Neither of the two new tests asserts anything about memory — the only assertions are the `ASSERT_TRUE(readSetsFromByteArray(...))` calls. That's the right call for avoiding flakiness (there's nothing machine- or build-type-dependent to fail on, which I checked for specifically), but it does mean these are registered CI work with no pass/fail signal: a memory regression can't turn them red.

Given that, they'd fit better as manual tools than as `add_test` entries (see my comment on `CMakeLists.txt:72`). Alternatively, a loose upper bound — peak added under some generous multiple of the input size — would give them an actual assertion without being fragile.

Worth adding a comment recording the machine and build type the #7217 numbers were taken on, since these figures are the justification for that PR and Debug vs Release moves them a lot.

ZeldaZach (changes requested):
Wanting the memory win in #7217 to be measurable rather than asserted is the right instinct, and gating the real `AllPrintings` run behind an env var is the right call. The `VmHWM` / `ru_maxrss` approach is sound in principle and the ZLIB/LibLZMA wiring here mirrors `oracle/CMakeLists.txt` exactly, so that part is clean.

The problem is that the harness under-reports in ways that matter for the conclusion it exists to support. The synthetic benchmark takes its baseline after a fixture build that has already pushed the process high-water mark to ~265 MB, so phases get measured against a mark the test itself set. And because both `VmHWM` and `ru_maxrss` are monotonic, the `releaseSetData()` phase is structurally incapable of showing anything but 0.0 MB.

To be clear about which numbers this does and doesn't affect: the headline figures in #7217 (process peak 3903 -> 1951 MB, parse phase +2733 -> +0 MB) come from `ImportRamUsageAllPrintings`, whose baseline sits after decompression, so those comparisons hold. It's `ImportRamUsage` and the release phases that are unreliable.

There are also two CI-facing issues: a 120 s ctest timeout against a benchmark that allows itself 10 minutes to download, and a full 60k-card synthetic workload that runs on Windows purely to print "unavailable". Details inline.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Baseline now precedes `buildSyntheticData()`, so the fixture builders ~27 MB JSON high-water no longer hides the importer phases. Combined with the log semantics change, per-phase deltas are now current-RSS based rather than peak-based.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Youre right that peak deltas are ~0.0 by construction (VmHWM/ru_maxrss are monotonic). `logRamPhase` no longer computes a peak delta; it reports current RSS + delta-vs-baseline (only where current RSS is live) + process peak. The release phase is therefore observable on Linux/macOS via the current-RSS column, and honestly N/A where a platform cant measure it.

tests/oracle/CMakeLists.txt
BruebachL: Moot now: the RAM benchmarks are no longer registered in CI (manual-only, same as the load_benchmark pattern) — no 120 s ctest budget applies. The manual workflow honours the in-code 10-min download timeout.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Fixed both failure modes: the timer lambda sets `timedOut` and calls `reply->abort()`, so a timed-out download takes the same `GTEST_SKIP` path as a network error instead of parsing a truncated body. A 200-with-HTML payload is rejected via `payload.startsWith("<")`, mirroring the pages.cpp check, before decompress/parse.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: macOS now gets a real current-RSS value: `task_info(...).resident_size` (the closest analog to Linux VmRSS), added alongside the existing `ru_maxrss` peak. So macOS no longer prints `N/A` for current RSS on every line.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: `ImportRamUsage` now checks `baseline.available` right after taking the snapshot and `GTEST_SKIP()`s before building the 60k-card fixture, so Windows no longer imports a full synthetic dataset just to print three "unavailable" lines.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Parsed with `qEnvironmentVariableIntValue`: the test only runs when the variable is set to a non-zero integer, so an explicit `=0` disables it (as does an absent variable) instead of triggering a ~25 MB download inside CI.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Baseline moved above the `QNetworkAccessManager`/request setup, so the `download+decompress` phase now covers the payload materialized by `readAll()` (plus the reply-internal buffer) instead of starting after it.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: The RAM benchmarks are now manual tools rather than `add_test` entries, so theres no registered-CI show with no pass/fail signal; staying assert-light on memory avoids the flakiness you flagged (nothing machine/build-type-dependent to fail on).

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -257,8 +277,288 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+    // Baseline must precede the fixture build: a high-water mark set while
+    // generating the synthetic JSON would otherwise mask the importer phases.
+    // Where memory stats are unavailable (Windows), skip before doing the
+    // 60k-card fixture build, which would otherwise be pure wasted work.
+    const MemorySnapshot baseline = MemorySnapshot::current();
+    if (!baseline.available) {
+        GTEST_SKIP() << "Memory stats unavailable on this platform";
+    }
+
+    const QByteArray data = buildSyntheticData(numSets, cardsPerSet);
```
ZeldaZach, line 437: This inverts the error rather than removing it. `buildSyntheticData` leaves ~172 MB of freed-but-unreturned heap behind (194.0 MB RSS holding a 22.0 MB `data`), and the baseline is now read before any of that exists, so `logRamPhase("parse", baseline, afterParse)` reports 234.5 - 13.3 = 221 MB. The parse actually cost 234.5 - 194.0 = 40 MB.

Both halves of the problem come from the same thing: current RSS never drops when the allocator holds onto arenas, so a baseline taken before the spike absorbs it and a baseline taken after it hides subsequent reuse.

Since the metric is now current RSS rather than a peak, taking the phase baseline immediately after `buildSyntheticData()` returns gives 40 MB, which is a lower bound but within ~2x of the truth instead of ~5.5x off. Printing the fixture build's own cost (`afterFixture - baseline`) on its own line makes the contamination visible rather than silent. That's a two-line change and enough to make the parse number quotable.

Same caveat applies across tests in this binary: `ParseJsonThroughput` and friends run first and leave their own arenas behind, so `ctest`-style whole-binary runs and `--gtest_filter=OracleBenchmark.ImportRamUsage` will not agree. Worth a line in the comment saying which invocation the numbers assume.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -257,8 +277,288 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+    const qint64 parseMs = timer.elapsed();
+    const MemorySnapshot afterParse = MemorySnapshot::current();
+
+    timer.restart();
+    const int importedSets = importer.startImport();
+    const qint64 importMs = timer.elapsed();
+    const MemorySnapshot afterImport = MemorySnapshot::current();
+
+    importer.releaseSetData();
+    const MemorySnapshot afterRelease = MemorySnapshot::current();
```
ZeldaZach, line 461: The reply to my last comment says the release phase is "observable on Linux/macOS via the current-RSS column" now that the peak delta is gone. It isn't. Destroying a ~200 MB `QJsonDocument` in the run above moved current RSS from 234.5 MB to 234.6 MB, i.e. up. Dropping the peak delta removed a column that was structurally 0.0; it didn't make freeing visible, because RSS only falls when the allocator returns pages, which it generally won't for the many small allocations a parsed set tree is made of.

So a working `releaseSetData()` and a no-op still print the same line. `ASSERT_TRUE(importer.getSets().isEmpty())` after the call is the signal that actually distinguishes them; keep the RSS line as colour if you want it.

tests/oracle/oracle_importer_benchmark_test.cpp
```diff
@@ -257,8 +277,288 @@ TEST(OracleBenchmark, ImportCardsWithColors)
...
+{
+    if (!baseline.available || !current.available) {
+        qDebug().noquote() << QString("  %1: memory stats unavailable on this platform").arg(phase);
+        return;
+    }
+    // VmHWM / ru_maxrss are monotonically non-decreasing high-water marks, so a
+    // peak-based delta between phases is ~0.0 MB by construction once the
+    // fixture build has set the process peak. The live signals are current RSS
+    // and the process peak; the delta is only meaningful where current RSS is
+    // a per-phase value (see "after releaseSetData()").
```
ZeldaZach: The parenthetical points at `after releaseSetData()` as the phase where the delta is meaningful, which is the one phase where it provably isn't (see below). Worth correcting so the next reader doesn't trust that column.

ZeldaZach (changes requested):
Most of the last round is properly handled, and the fixes are the right ones: the benchmark is no longer registered with `add_test` so the 120 s budget can't collide with the 10-minute download allowance, the timer lambda now aborts the reply so a timeout takes the skip path, the HTML-body case is rejected before parsing, macOS gets a real current-RSS from `task_info`, Windows skips before the 60k-card fixture build, `COCKATRICE_ORACLE_RAM_BENCHMARK=0` now disables instead of enabling, and the `download+decompress` baseline sits above the `QNetworkAccessManager` work. Switching `logRamPhase` off peak deltas onto current RSS is also the right call. CI is green on all 18 jobs.

One fix overshot and one didn't land, both in `ImportRamUsage`.

Moving the baseline *before* `buildSyntheticData()` doesn't take the fixture builder out of the measurement, it moves the whole thing into the parse phase, because the builder's freed memory is never returned to the OS. I compiled the fixture builder and this snapshot code against Qt 6.11.1 (clang, -O2, macOS):

```
baseline (test start)                cur RSS   13.3 MB | peak  13.3 MB
after buildSyntheticData(30,2000)    cur RSS  194.0 MB | peak 222.0 MB   (json is 22.0 MB)
after QJsonDocument::fromJson        cur RSS  234.5 MB | peak 234.5 MB
after doc destroyed                  cur RSS  234.6 MB | peak 234.6 MB
```

So the `parse` line prints 221 MB for a parse that cost 40 MB, and the last row shows the `after releaseSetData()` phase still can't see a release. Details inline.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Fixed in `64e5c0de19`. The parse baseline is now taken immediately after `buildSyntheticData()` returns, so the immutable (freed-but-unreturned) fixture arenas no longer contaminate the parse phase. The importer parse now reports its real ~4 MB growth on the synthetic fixture instead of a ~221 MB artifact. The fixture build costs own RSS spike is printed on its own line (`fixture build: delta vs baseline 242.5 MB`), so the contamination is visible rather than silent. Also added a comment noting that ParseJsonThroughput and friends run first in a whole-binary ctest run, so per-invocation numbers can differ.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: You are right, and the RSS line for the release phase is now explicitly not the evidence. Fixed in `64e5c0de19`: `ASSERT_TRUE(importer.getSets().isEmpty())` right after `releaseSetData()` is the signal that the buffers were actually dropped (it clears `allSets`, which is what `getSets()` returns). The current-RSS line stays as colour only.

tests/oracle/oracle_importer_benchmark_test.cpp
BruebachL: Fixed in `64e5c0de19`. The comment no longer points at `after releaseSetData()` as the phase where the delta is meaningful. It now says the delta is meaningful only where the baseline was taken immediately before the phase it measures (e.g. the import phase, which compares afterParse against afterImport); the fixture/release phases carry a comment noting RSS rarely falls once the allocator retains arenas, so those deltas are colour, not signal.

ZeldaZach (approved):

PR #7217, opened 2026-08-29
[Oracle] Parse sets lazily to slash importer peak memory

BruebachL:
Cut the oracle importer's peak memory roughly in half by scanning for set ranges once and parsing one set at a time instead of materializing one giant `QJsonDocument` for the whole database.

- Add `RawJson`, a small hand-rolled scanner that locates each set's byte range in the raw buffer, validating escapes/unicode/braces as it goes
- Nesting depth is capped at 1024 (matching Qt's DeepNesting) so pathological documents fail shallowly instead of exhausting the stack; string metadata fields tolerate null/non-string values so one broken set can't abort the whole document
- `SetDataRange` (start/length/cardCount) is split out of the scan metadata so `SetToDownload` only carries the slice it needs
- Parse each set's slice lazily in `startImport()`; the full buffer is kept only as bytes, not as a DOM tree, with a bounds check before every range slice and honest progress reporting when a set fails to parse
- Adopt the decompressor's buffer by move (`rawSetsData = std::move(data)`) instead of copying it; clear raw data in `releaseSetData()`/`clear()`
- Differential scanner tests against `QJsonDocument::fromJson`, including verdict agreement on metadata quirks and 5000-deep nesting

Measured on the real `AllPrintings` import (629 MB decompressed, 35.5k unique cards, Debug): process peak `3903 MB → 1951 MB`, parse phase `+2733 MB → +0 MB`, scan `~1.8x` faster than the previous full parse. Import wall time increases slightly in Debug (per-set re-parse) with a far smaller gap in Release.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -7,6 +7,7 @@
 #include <QDebug>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QJsonParseError>
 #include <QRegularExpression>
 #include <QSet>
 #include <algorithm>
@@ -44,26 +45,23 @@ static CardSet::Priority getSetPriority(const QString &setType, const QString &s
     return priority;
 }
 
-bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
+bool OracleImporter::readSetsFromByteArray(QByteArray data)
 {
-    QJsonParseError error;
-    auto doc = QJsonDocument::fromJson(data, &error);
-    if (error.error != QJsonParseError::NoError) {
-        qDebug() << "error: QJsonDocument::fromJson():" << error.errorString();
+    RawJson::ScanError error;
+    const QList<RawJson::SetRange> ranges = RawJson::scanSetRanges(data, &error);
+    if (error.isError()) {
+        qDebug() << "error: RawJson::scanSetRanges():" << error.message;
         return false;
     }
 
-    auto setsObj = doc.object().value("data").toObject();
-
     QList<SetToDownload> newSetList;
+    newSetList.reserve(ranges.size());
 
-    for (auto it = setsObj.constBegin(); it != setsObj.constEnd(); ++it) {
-        QJsonObject setObj = it.value().toObject();
-        QString shortName = setObj.value("code").toString().toUpper();
-        QString longName = setObj.value("name").toString();
-        QJsonArray setCards = setObj.value("cards").toArray();
-        QString setType = setObj.value("type").toString();
-        QDate releaseDate = QDate::fromString(setObj.value("releaseDate").toString(), Qt::ISODate);
+    for (const RawJson::SetRange &range : ranges) {
+        QString shortName = range.code.toUpper();
+        QString longName = range.name;
+        QString setType = range.type;
+        QDate releaseDate = QDate::fromString(range.releaseDate, Qt::ISODate);
         CardSet::Priority priority = getSetPriority(setType, shortName);
         // capitalize set type
         if (setType.length() > 0) {
@@ -83,7 +81,9 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
             }
             setType = setType.trimmed();
         }
-        newSetList.append(SetToDownload(shortName, longName, setCards, priority, setType, releaseDate));
+        SetToDownload set(shortName, longName, priority, setType, releaseDate);
+        set.setRawRange(range.dataRange);
+        newSetList.append(set);
     }
 
     std::sort(newSetList.begin(), newSetList.end());
@@ -92,6 +92,7 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
         return false;
     }
     allSets = newSetList;
+    rawSetsData = std::move(data);
     return true;
 }
 
@@ -550,22 +551,17 @@ int OracleImporter::startImport()
 {
     static ICardSetPriorityController *noOpController = new NoopCardSetPriorityController();
 
-    // Pre-allocate the cards hash to avoid rehashing during import. The hash
-    // is keyed by distinct card name rather than by printings: AllPrintings
-    // ships ~100k printings for ~35k names, so reserving the printing count
-    // would overallocate ~3x (against this stack's RAM goal). Collecting
-    // distinct names is cheap — one pass over the already-parsed name fields.
-    {
-        QSet<QString> distinctNames;
-        for (const SetToDownload &curSetToParse : allSets) {
-            for (const QJsonValue &cardValue : curSetToParse.getCards()) {
-                distinctNames.insert(cardValue.toObject().value("name").toString());
-            }
-        }
-        cards.reserve(distinctNames.size());
-        // The set goes out of scope here, handing the ~35k name QStrings back
-        // to the allocator before the (memory-heavy) import loop starts.
+<<<<<<< HEAD
+    // Pre-allocate the cards hash to avoid rehashing during import. Keys are
+    // distinct card names while raw ranges only count printings (AllPrintings
+    // ~100k printings vs ~35k names), so this over-reserves somewhat; an exact
+    // distinct-name count would require eagerly parsing, which the lazy reader
+    // deliberately avoids. It's a capacity hint, so the overshoot is harmless.
+    int estimatedCards = 0;
+    for (const SetToDownload &curSetToParse : allSets) {
+        estimatedCards += curSetToParse.getRawRange().cardCount;
     }
+    cards.reserve(estimatedCards);
 
     // add an empty set for tokens
     CardSetPtr tokenSet =
@@ -578,11 +574,44 @@ int OracleImporter::startImport()
         CardSetPtr newSet = CardSet::newInstance(noOpController, curSetToParse.getShortName(),
                                                  curSetToParse.getLongName(), curSetToParse.getSetType(),
                                                  curSetToParse.getReleaseDate(), curSetToParse.getPriority());
+
+        // parse only this set's slice of the raw document so the whole JSON tree is
+        // never kept in memory at once
+        const RawJson::SetDataRange &rawRange = curSetToParse.getRawRange();
+        const qsizetype rangeEnd = rawRange.start + rawRange.length;
+        if (rawRange.start < 0 || rawRange.length <= 0 || rangeEnd > rawSetsData.size()) {
+            // rawSetsData is cleared by releaseSetData() while SetToDownload copies
+            // taken from getSets() keep their ranges, and nothing else enforces the
+            // pairing — so never index past the buffer on stale/mismatched ranges.
+            qWarning() << "error: out-of-bounds raw range for set" << curSetToParse.getShortName() << "skipping";
+            ++setIndex;
+            emit setIndexChanged(0, setIndex, curSetToParse.getLongName());
+            continue;
+        }
+        // sliced() shares the buffer instead of deep-copying the slice; the largest
+        // sets in AllPrintings are tens of MB, so the copy is worth avoiding here.
+        const QByteArray setBytes = rawSetsData.sliced(rawRange.start, rawRange.length);
+        QJsonParseError parseError;
+        const QJsonDocument setDoc = QJsonDocument::fromJson(setBytes, &parseError);
+        if (parseError.error != QJsonParseError::NoError) {
+            qWarning() << "error: parsing card data for set" << curSetToParse.getShortName() << ":"
+                       << parseError.errorString();
+            ++setIndex;
+            // Keep the progress accounting honest: a set that failed to parse
+            // still advanced the index, so report it (with zero imported cards)
+            // rather than letting SaveSetsPage's bar stall per failed set.
+            emit setIndexChanged(0, setIndex, curSetToParse.getLongName());
+            continue;
+        }
+
+        // Only add the set to the database once its slice parsed cleanly;
+        // a set that fails here must not persist as an empty set in cards.xml.
         if (!sets.contains(newSet->getShortName())) {
             sets.insert(newSet->getShortName(), newSet);
         }
 
-        int numCardsInSet = importCardsFromSet(newSet, curSetToParse.getCards());
+        const QJsonArray setCards = setDoc.object().value("cards").toArray();
+        int numCardsInSet = importCardsFromSet(newSet, setCards);
 
         ++setIndex;
 
@@ -605,6 +634,7 @@ bool OracleImporter::saveToFile(const QString &fileName, const QString &sourceUr
 void OracleImporter::releaseSetData()
 {
     allSets.clear();
+    rawSetsData.clear();
 }
 
 void OracleImporter::clear()
@@ -612,4 +642,5 @@ void OracleImporter::clear()
     sets.clear();
     cards.clear();
     allSets.clear();
+    rawSetsData.clear();
 }
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -551,7 +551,6 @@ int OracleImporter::startImport()
 {
     static ICardSetPriorityController *noOpController = new NoopCardSetPriorityController();
 
-<<<<<<< HEAD
     // Pre-allocate the cards hash to avoid rehashing during import. Keys are
     // distinct card names while raw ranges only count printings (AllPrintings
     // ~100k printings vs ~35k names), so this over-reserves somewhat; an exact
--- a/oracle/src/oracleimporter.h
+++ b/oracle/src/oracleimporter.h
@@ -1,6 +1,9 @@
 #ifndef ORACLEIMPORTER_H
 #define ORACLEIMPORTER_H
 
+#include "raw_json_scanner.h"
+
+#include <QByteArray>
 #include <QJsonArray>
 #include <QJsonObject>
 #include <QMap>
@@ -46,10 +49,12 @@ class SetToDownload
 {
 private:
     QString shortName, longName;
-    QJsonArray cards;
     QDate releaseDate;
     QString setType;
     CardSet::Priority priority;
+    // Byte range of this set's object within the importer's raw JSON text. Parsing
+    // one set at a time keeps peak memory low instead of holding the whole document.
+    RawJson::SetDataRange rawRange;
 
 public:
     const QString &getShortName() const
@@ -59,10 +64,6 @@ public:
     const QString &getLongName() const
     {
         return longName;
-    }
-    const QJsonArray &getCards() const
-    {
-        return cards;
     }
     const QString &getSetType() const
     {
@@ -76,15 +77,22 @@ public:
     {
         return priority;
     }
+    const RawJson::SetDataRange &getRawRange() const
+    {
+        return rawRange;
+    }
     SetToDownload(QString _shortName,
                   QString _longName,
-                  QJsonArray _cards,
                   CardSet::Priority _priority,
                   QString _setType = QString(),
                   const QDate &_releaseDate = QDate())
-        : shortName(std::move(_shortName)), longName(std::move(_longName)), cards(std::move(_cards)),
-          releaseDate(_releaseDate), setType(std::move(_setType)), priority(_priority)
+        : shortName(std::move(_shortName)), longName(std::move(_longName)), releaseDate(_releaseDate),
+          setType(std::move(_setType)), priority(_priority)
     {
+    }
+    void setRawRange(const RawJson::SetDataRange &_rawRange)
+    {
+        rawRange = _rawRange;
     }
     bool operator<(const SetToDownload &set) const
     {
@@ -141,6 +149,12 @@ private:
 
     QList<SetToDownload> allSets;
 
+    /**
+     * The raw JSON text of the source document, retained for lazy per-set
+     * parsing during startImport(). Frees the card data as each set is imported.
+     */
+    QByteArray rawSetsData;
+
     CardInfoPtr addCard(QString name,
                         const QString &text,
                         bool isToken,
@@ -153,7 +167,11 @@ signals:
 
 public:
     explicit OracleImporter(QObject *parent = nullptr);
-    bool readSetsFromByteArray(const QByteArray &data);
+    /**
+     * Scans the given JSON document for set metadata. Takes the data by value so
+     * the wizard can hand over its decompressed buffer without copying it.
+     */
+    bool readSetsFromByteArray(QByteArray data);
     int startImport();
     bool saveToFile(const QString &fileName, const QString &sourceUrl, const QString &sourceVersion);
     int importCardsFromSet(const CardSetPtr &currentSet, const QJsonArray &cardsList);
@@ -168,6 +186,10 @@ public:
     QList<SetToDownload> &getSets()
     {
         return allSets;
+    }
+    const QByteArray &getRawSetsData() const
+    {
+        return rawSetsData;
     }
     void releaseSetData();
     void clear();
--- a/oracle/src/raw_json_scanner.cpp
+++ b/oracle/src/raw_json_scanner.cpp
@@ -1,0 +1,621 @@
+#include "raw_json_scanner.h"
+
+#include <cstring>
+
+namespace
+{
+
+// Nesting cap matching QJsonDocument's limit, so a pathologically deep document
+// fails shallowly instead of overflowing the stack through the recursive
+// skipValue/skipArray/skipObject walk (Qt's parser caps at 1024 for the same
+// reason and reports DeepNesting).
+constexpr int kMaxNestingDepth = 1024;
+
+inline bool isWhitespace(char c)
+{
+    return c == ' ' || c == '\t' || c == '\r' || c == '\n';
+}
+
+const char *skipWhitespace(const char *p, const char *end)
+{
+    while (p < end && isWhitespace(*p)) {
+        ++p;
+    }
+    return p;
+}
+
+inline bool isHexDigit(char c)
+{
+    return (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F');
+}
+
+inline quint8 hexValue(char c)
+{
+    if (c >= '0' && c <= '9') {
+        return c - '0';
+    }
+    if (c >= 'a' && c <= 'f') {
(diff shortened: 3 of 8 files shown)


oracle/src/raw_json_scanner.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Splits the top-level "data" object into per-set byte ranges and reads each
+ * set's metadata directly from the raw bytes. The oracle importer can then
+ * parse one set at a time during import, keeping peak memory far below a single
+ * QJsonDocument::fromJson() over the whole file.
+ *
+ * The whole document is structurally validated while scanning (strings,
+ * escapes, braces, and a trailing-content check), so malformed input is
+ * rejected just like QJsonDocument::fromJson would.
+ */
```
RickyRister, line 71: document parameters

oracle/src/oracleimporter.h
```diff
@@ -46,10 +49,14 @@ class SetToDownload
 {
 private:
     QString shortName, longName;
-    QJsonArray cards;
     QDate releaseDate;
     QString setType;
     CardSet::Priority priority;
+    // Byte range of this set's object within the importer's raw JSON text. Parsing
+    // one set at a time keeps peak memory low instead of holding the whole document.
+    qsizetype rawStart = -1;
```
RickyRister: Can't you reuse the `SetRange` struct here?

oracle/src/raw_json_scanner.h
```diff
@@ -0,0 +1,50 @@
...
+ScanError scanSetRanges(const QByteArray &json, QList<SetRange> &ranges);
```
RickyRister: Also, `QJsonDocument::fromJson` returns the result and uses the out parameter for the error, so we might want to follow that convention.

oracle/src/oracleimporter.cpp
```diff
@@ -91,6 +91,7 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
         return false;
     }
     allSets = newSetList;
+    rawSetsData = std::move(data);
```
RickyRister, line 95: `QByteArray` uses implicit copying anyways, so moving it is pretty pointless. It's basically equivalent to passing by const ref and then doing the copy assignment here.

oracle/src/oracleimporter.h
```diff
@@ -46,10 +49,12 @@ class SetToDownload
 {
 private:
     QString shortName, longName;
-    QJsonArray cards;
     QDate releaseDate;
     QString setType;
     CardSet::Priority priority;
+    // Byte range of this set's object within the importer's raw JSON text. Parsing
+    // one set at a time keeps peak memory low instead of holding the whole document.
+    RawJson::SetRange rawRange;
```
RickyRister: nit: `SetRange` contains some redundant info alongside `SetToDownload`, such as the name and release date.
Perhaps we can split out a struct that contains only the `start`/`length`/`cardCount`, and then nest that struct in `SetRange`, and then use that struct here.

oracle/src/raw_json_scanner.cpp
```diff
@@ -0,0 +1,573 @@
...
+        }
+        if (*p == ']') {
+            ++p;
+            return true;
+        }
+        return false;
+    }
+}
+
+bool skipValue(const char *&p, const char *end)
```
ZeldaZach: `skipValue` -> `skipArray`/`skipObject` -> `skipValue` recurses once per nesting level with no depth limit, so nesting depth in the input translates directly into stack depth. Qt's parser caps nesting at 1024 and returns `QJsonParseError::DeepNesting` precisely to avoid this.

I compiled this file standalone against Qt 6.11 and ran a payload of `{"data":{"A":{"code":"a","cards":[` + N x `[` + N x `]` + `]}}}`:

```
depth 500     Qt: no error                      RawJson: ok, 1 range
depth 5000    Qt: too deeply nested document    RawJson: ok, 1 range
depth 10000   Qt: too deeply nested document    RawJson: ok, 1 range
depth 50000   Qt: too deeply nested document    RawJson: ok, 1 range
depth 200000  Qt: too deeply nested document    RawJson: Segmentation fault: 11
```

Under ASan the stack trace is unambiguous:

```
==31450==ERROR: AddressSanitizer: stack-overflow on address 0x00016f3a7fff
    #0 isWhitespace(char) raw_json_scanner.cpp:9
    #1 skipValue(char const*&, char const*) raw_json_scanner.cpp:370
    #2 skipArray(char const*&, char const*) raw_json_scanner.cpp:349
    #3 skipValue(char const*&, char const*) raw_json_scanner.cpp:379
    ... repeating
```

That was a release build on macOS with an 8 MB main-thread stack, and the payload is about 400 KB. In the app this runs on a `QtConcurrent` pool thread (`pages.cpp:474`), whose stack is typically no larger, so the real threshold is likely lower than 200k. A corrupted download or a hostile mirror crashes Oracle with no diagnostic.

Threading a depth counter through `skipValue`/`skipObject`/`skipArray` and failing past a limit fixes it:

```cpp
constexpr int MaxNestingDepth = 1024; // matches QJsonDocument

bool skipValue(const char *&p, const char *end, int depth)
{
    if (depth > MaxNestingDepth) {
        return false;
    }
    ...
}
```

`countArrayElements` needs the same treatment since it calls `skipValue` too. Worth a test at depth 2000 asserting an error rather than a crash.

oracle/src/raw_json_scanner.h
```diff
@@ -0,0 +1,60 @@
...
+ * @brief Scans a full MTGJSON document without materializing the JSON tree.
+ *
+ * Splits the top-level "data" object into per-set byte ranges and reads each
+ * set's metadata directly from the raw bytes. The oracle importer can then
+ * parse one set at a time during import, keeping peak memory far below a single
+ * QJsonDocument::fromJson() over the whole file.
+ *
+ * The whole document is structurally validated while scanning (strings,
+ * escapes, braces, and a trailing-content check), so malformed input is
+ * rejected just like QJsonDocument::fromJson would.
```
ZeldaZach: This claim doesn't hold, and since the tests are framed as differential it's worth getting the docs to match reality. Running both parsers over the same inputs:

```
input                            | QJsonDocument | RawJson
---------------------------------|---------------|---------------------------
normal set                       | ok   sets=1   | ok   sets=1
releaseDate: null                | ok   sets=1   | ERR  malformed set data
name: null                       | ok   sets=1   | ERR  malformed set data
type: 7                          | ok   sets=1   | ERR  malformed set data
cards: null                      | ok   sets=1   | ERR  malformed set data
cards is an object               | ok   sets=1   | ERR  malformed set data
leading UTF-8 BOM                | ok   sets=1   | ERR  must be an object
duplicate "data" keys            | ok   sets=1   | ok   sets=2
number 1e400                     | ERR           | ok   sets=1
nesting depth 5000               | ERR DeepNesting | ok  sets=1
nesting depth 200000             | ERR DeepNesting | SEGFAULT
```

The divergences fall into three groups, and only the first is a correctness problem in itself (see my comment on the metadata callback). The others matter because the scanner and the per-set `QJsonDocument::fromJson` in `startImport()` are two different parsers with different acceptance criteria: anything the scanner accepts but Qt rejects (deep nesting, `1e400`) passes the scan and then fails per-set parsing, dropping that entire set with only a `qWarning`. A set silently vanishing from the user's card database is a bad failure mode.

Suggest either tightening the scanner to match Qt on those cases, or rewording this paragraph to state what is actually guaranteed: that the scan validates enough structure to locate set boundaries, not that it accepts exactly the same language as `QJsonDocument`.

The BOM one is mostly theoretical, since `pages.cpp:471` already gates on `_data.startsWith("{")`, but it's another place where the two disagree.

oracle/src/raw_json_scanner.cpp
```diff
@@ -0,0 +1,573 @@
...
+                    }
+                    SetRange range;
+                    range.start = setStart - begin;
+                    range.length = setEnd - setStart;
+                    range.code = setCode;
+
+                    const char *memberP = setStart;
+                    const bool metaOk =
+                        forEachObjectMember(memberP, setEnd, [&](const QString &field, const char *fs, const char *fe) {
+                            if (field == QStringLiteral("code")) {
```
ZeldaZach: These four metadata fields are required to be strings, and returning `false` here propagates all the way out through `forEachObjectMember` to `malformedSetData`, which fails `scanSetRanges` entirely. So a single set anywhere in `AllPrintings` with `"releaseDate": null` aborts the import of all ~900 sets:

```
{"data":{"A":{"code":"a","name":"N","type":"expansion","releaseDate":null,"cards":[{"n":1}]}}}
  QJsonDocument -> ok, 1 set
  RawJson       -> ERR "malformed set data"
```

Same for `"name": null`, `"type": 7`, and `"cards": null`. The pre-#7217 code read these with `.toString()` / `.toArray()`, which yields an empty string or empty array for a non-string value and carries on.

MTGJSON may well not emit nulls for these today, but this is a hard dependency on that staying true, and the failure is total: the user gets "No set has been imported", and the only explanation is a `qDebug` line saying `malformed set data`. One upstream schema tweak and card database updates break for everyone.

Suggest treating a wrong-typed metadata field as absent rather than fatal, keeping the hard failure for genuinely unparseable structure:

```cpp
if (field == QStringLiteral("code")) {
    if (fs < fe && *fs == '"') {
        decodeString(fs, fe, range.code);
    }
    return true; // leave the object key as the fallback
}
```

`cards` is the one where a hard failure might be defensible, though even there skipping the set beats failing the whole document. If you keep any of these strict, the error message should name the offending set.

oracle/src/oracleimporter.cpp
```diff
@@ -560,7 +561,21 @@ int OracleImporter::startImport()
             sets.insert(newSet->getShortName(), newSet);
         }
 
-        int numCardsInSet = importCardsFromSet(newSet, curSetToParse.getCards());
+        // parse only this set's slice of the raw document so the whole JSON tree is
+        // never kept in memory at once
+        const RawJson::SetRange &rawRange = curSetToParse.getRawRange();
+        const QByteArray setBytes(rawSetsData.constData() + rawRange.start, rawRange.length);
```
ZeldaZach: No bounds check before pointer arithmetic on `rawSetsData`. If `rawRange.start`/`length` ever disagree with the current `rawSetsData` this reads out of bounds and silently feeds garbage to `QJsonDocument::fromJson`.

The pairing is safe today because `readSetsFromByteArray` sets `allSets` and `rawSetsData` together, but nothing enforces it: `getSets()` hands out a non-const `QList<SetToDownload> &`, so any caller can repopulate the list, and `releaseSetData()` clears `rawSetsData` while `SetToDownload` objects copied out of `getSets()` keep their stale offsets. On an empty `rawSetsData`, `constData()` returns a pointer to a shared null byte and `start` walks straight off it.

Since the offsets come from a hand-written scanner rather than from Qt, a guard here is cheap insurance:

```cpp
if (rawRange.start < 0 || rawRange.length <= 0 || rawRange.start + rawRange.length > rawSetsData.size()) {
    qWarning() << "error: invalid raw range for set" << curSetToParse.getShortName();
    ++setIndex;
    continue;
}
```

`QByteArray::sliced(start, length)` would also be a cheaper way to build `setBytes` than the deep-copying `QByteArray(const char *, qsizetype)` constructor, since it shares the buffer instead of copying the slice. That is worth something here given the PR is about peak memory.

oracle/src/oracleimporter.cpp
```diff
@@ -560,7 +561,21 @@ int OracleImporter::startImport()
...
+        QJsonParseError parseError;
+        const QJsonDocument setDoc = QJsonDocument::fromJson(setBytes, &parseError);
+        if (parseError.error != QJsonParseError::NoError) {
```
ZeldaZach, line 595: This `continue` skips the `emit setIndexChanged(...)` at the bottom of the loop, so a set that fails to parse advances `setIndex` without telling the UI. `SaveSetsPage` drives its progress bar off that signal, so the bar stalls for each failed set and the final count won't line up with what was reported.

Emitting with zero cards keeps the progress accounting honest:

```cpp
++setIndex;
emit setIndexChanged(0, setIndex, curSetToParse.getLongName());
continue;
```

Separately: `newSet` has already been inserted into `sets` by this point, so a set that fails to parse still ends up in the saved `cards.xml` as an empty set. Probably worth deferring the insert until after the parse succeeds.

oracle/src/oracleimporter.cpp
```diff
@@ -91,6 +91,7 @@ bool OracleImporter::readSetsFromByteArray(const QByteArray &data)
         return false;
     }
     allSets = newSetList;
+    rawSetsData = data;
```
ZeldaZach: `data` is a by-value parameter that dies at the end of this function, so this wants to be `rawSetsData = std::move(data);`. `QByteArray` is implicitly shared so the current form doesn't actually copy the buffer, but the header comment two files over advertises move semantics ("Takes the data by value so the wizard can hand over its decompressed buffer without copying it") and this is the line that would make it literally true rather than incidentally true.

tests/oracle/oracle_importer_test.cpp
```diff
@@ -488,6 +488,112 @@ TEST_F(OracleImporterTest, ApostropheNormalized)
...
+
+    QJsonParseError parseError;
+    const QJsonDocument sliceDoc =
+        QJsonDocument::fromJson(QByteArray(json.constData() + range.start, range.length), &parseError);
+    ASSERT_EQ(parseError.error, QJsonParseError::NoError);
+    ASSERT_EQ(sliceDoc.object().value("name").toString(), expectedName);
+    ASSERT_EQ(sliceDoc.object().value("cards").toArray().size(), 3);
+}
+
+TEST_F(OracleImporterTest, ScanSetRangesRejectsInvalidJson)
```
ZeldaZach, line 605: Good set of negative cases. Two gaps worth closing given what the scanner now guards:

The tests assert that the scanner errors, but never that it agrees with `QJsonDocument::fromJson` on the same input. `ScanSetRangesMatchFullJsonParse` is the only genuinely differential test and it only covers valid input. Asserting both parsers reach the same verdict for each of these strings is a small change and would have caught the divergences I listed on the header.

Missing inputs that exercise the paths this PR adds: nesting depth beyond a few thousand (currently a segfault, see my comment on `skipValue`), non-string metadata values such as `"releaseDate": null` (currently fails the whole document), a leading UTF-8 BOM, unescaped control characters inside a string, `1e400`, and a truncated document that ends mid-string. A table-driven test over `{input, expectQtOk, expectScanOk}` would cover all of these compactly.

One note on the valid-input test above: `QJsonDocument(root).toJson(Compact)` emits keys in sorted order with no whitespace, which is a much friendlier layout than real `AllPrintings`. A fixture with irregular whitespace between tokens and nested objects inside `cards` would exercise `skipWhitespace` and the brace tracking more honestly.

ZeldaZach (changes requested):
This is the interesting one, and the measured result (3903 MB -> 1951 MB peak, parse phase +2733 MB -> +0 MB) is a genuinely big win. The scan-then-slice design is the right shape, and `SetRange` carrying offsets rather than a `QJsonArray` is a clean way to express it.

My concern is that this replaces a hardened parser with a hand-written one on a code path that consumes data downloaded over the network from a URL the user can edit. I built `raw_json_scanner.cpp` standalone against Qt 6.11 and diffed its behaviour against `QJsonDocument::fromJson` on the same inputs. Two results need addressing before this merges: the scanner segfaults on deeply nested input where Qt returns an error, and it rejects whole documents that Qt accepts. Both are inline with reproductions.

The rest are smaller: a missing bounds check, a dropped progress signal, and some coverage gaps in the differential tests.

To be clear about scope, none of this is an argument against the approach. The memory win is worth having and the design is sound; the scanner just needs to be as unshakeable as the thing it replaced, since it now sits in front of untrusted input.

oracle/src/oracleimporter.h
BruebachL: Split out `RawJson::SetDataRange` (start/length/cardCount only) and nested it in `SetRange`; `SetToDownload` now stores just a `SetDataRange`, so the metadata strings (name/release date/code/type) live only in the scan result and arent duplicated into every set.

oracle/src/raw_json_scanner.cpp
BruebachL: Throttled with a hard nesting cap: `kMaxNestingDepth = 1024` threaded through `skipValue`/`skipArray`/`skipObject`/`forEachObjectMember`/`countArrayElements`, matching Qts DeepNesting limit. Pathologically deep documents now fail shallowly with exactly the same reject/accept verdict as QJsonDocument (covered by the new differential test).

oracle/src/raw_json_scanner.h
BruebachL: Docs updated to say what it actually does: verdicts agree with QJsonDocument on structurally malformed input, and unlike Qt the string metadata fields tolerate null/non-string values (defaulting to empty) so one broken set cant abort the rest. The differential reality is now the claimed one.

oracle/src/raw_json_scanner.cpp
BruebachL: Metadata fields are no longer required to be quoted strings: `decodeStringMember` tolerates null/non-string values (`"releaseDate": null`, `"name": null`, `"type": 7`) leaving the default, and `"cards": null` counts as an empty array. A single bad set no longer fails the whole document — verified against Qt in the new verdict test.

oracle/src/oracleimporter.cpp
BruebachL: Added a bounds check before the pointer arithmetic in `startImport`: a range is rejected (and the set skipped, with progress reported) unless `start >= 0`, `length > 0`, and `start + length <= rawSetsData.size()`. That covers the documented hazard of `releaseSetData()`/`getSets()` copies leaving stale ranges.

oracle/src/oracleimporter.cpp
BruebachL: The parse-failure path now emits `setIndexChanged(0, setIndex, setName)` after incrementing, so SaveSetsPages progress bar advances for failed sets too and the final count lines up with what was reported. The out-of-bounds guard path emits the same.

oracle/src/oracleimporter.cpp
BruebachL: `rawSetsData = std::move(data);` — now literal rather than incidental with the by-value parameter, matching the advertised move semantics of `readSetsFromByteArray`.

tests/oracle/oracle_importer_test.cpp
BruebachL: Two new tests: `ScanSetRangesMatchesFullJsonParseVerdicts` asserts the scanner and QJsonDocument reach the same accept/reject verdict on the metadata quirks (`name`/`type`/`releaseDate` null, `type: 7`, `cards: null`) plus structural negatives, and re-parses every returned range slice; `ScanSetRangesRejectsDeepNesting` feeds 5000 levels of nesting and asserts both parsers reject it (no stack overflow). 31/31 oracle unit tests green, benchmark unaffected.

oracle/src/raw_json_scanner.cpp
```diff
@@ -0,0 +1,621 @@
...
+    p = skipWhitespace(p, end);
+    if (p >= end) {
+        return false;
+    }
+    const char c = *p;
+    if (c == '{') {
+        return skipObject(p, end, depth - 1);
+    }
+    if (c == '[') {
+        return skipArray(p, end, depth - 1);
```
ZeldaZach: The cap ends up at roughly half of `kMaxNestingDepth`, because each nesting level costs two decrements: `skipValue` takes `depth`, passes `depth - 1` to `skipObject`/`skipArray`, and those pass `depth - 1` again to the next `skipValue`. Same in `forEachObjectMember`.

Measured against Qt 6.11, walking depth 1..1100 and recording the last depth each parser accepts:

```
last depth accepted: Qt=1020  Scan=508
```

So anything nested between 509 and 1020 levels is accepted by `QJsonDocument` and rejected outright by the scanner, and `scanSetRanges` failing means the whole import dies with "No set has been imported" rather than one set being skipped. The header comment and the comment on `kMaxNestingDepth` both claim 1024 to match Qt, and `ScanSetRangesRejectsDeepNesting` at 5000 can't catch it since both parsers reject up there.

Decrementing once per level fixes it - have `skipValue` pass `depth` through to `skipObject`/`skipArray` and let those do the single `depth - 1` on their recursive `skipValue` call. Worth pinning with a test at the boundary (a depth Qt accepts, asserting the scanner accepts it too) rather than only far past it.

oracle/src/raw_json_scanner.cpp
```diff
@@ -0,0 +1,621 @@
...
+            ++p;
+            continue;
+        }
+        if (c == '"') {
+            ++p;
+            flush();
+            return true;
+        }
+        if (static_cast<unsigned char>(c) < 0x20) {
+            return false; // unescaped control character is invalid JSON
```
ZeldaZach: Qt's parser accepts unescaped control characters inside strings, and `skipString` accepts them too, so this is the only place in the scanner that rejects them - and it rejects the entire document:

```
{"data":{"A":{"code":"a","name":"N<TAB>X","cards":[{"n":1}]}}}
  QJsonDocument -> ok, 1 set
  RawJson       -> ERR "malformed set data"

{"data":{"A":{"code":"a","cards":[{"text":"a<LF>b"}]}}}
  QJsonDocument -> ok, 1 set
  RawJson       -> ok, 1 set     (same byte, inside a card, via skipString)
```

A raw tab in a set name kills the import; the same byte in a card is fine. That's the failure mode the header now promises can't happen ("one broken set cannot abort the import of the rest"), and it's inconsistent with `skipString` either way. Dropping the check to match `skipString` and Qt is the smaller change; the alternative is adding the check to `skipString`, but then the scanner starts rejecting documents Qt accepts, which is worse here.

oracle/src/oracleimporter.cpp
```diff
@@ -578,7 +577,35 @@ int OracleImporter::startImport()
             sets.insert(newSet->getShortName(), newSet);
```
ZeldaZach: Still outstanding from the last round: `newSet` is inserted here, before either of the two `continue` paths below. A set whose slice fails `QJsonDocument::fromJson` is skipped for card import but stays in `sets`, so `saveToFile` writes it to the user's `cards.xml` as an empty set.

This matters more now than when I first raised it, since the per-set-parse-failure path is new in this PR and is reachable whenever the scanner accepts something Qt doesn't. Moving the insert to after the parse succeeds keeps a failed set out of the database entirely.

While you're here: `rawSetsData.sliced(rawRange.start, rawRange.length)` on the line below avoids the deep copy the `QByteArray(const char *, qsizetype)` constructor does, which is worth having in a PR about peak memory - the largest sets in AllPrintings are tens of MB.

ZeldaZach (changes requested):
Re-reviewed against the new commit. Everything I blocked on is genuinely fixed, and I re-ran my standalone harness against the current `raw_json_scanner.cpp` (Qt 6.11, ASan) to confirm rather than take it on trust:

- 200k-deep nesting no longer segfaults; it returns an error.
- `"releaseDate": null`, `"name": null`, `"type": 7`, `"cards": null` now all scan clean and agree with `QJsonDocument`.
- The bounds check, the `setIndexChanged` emit on the failure path, `std::move(data)`, and the `SetDataRange` split are all in.

The design is right and the memory win is real. Three things left, all small, and the first two are in the depth-cap code that was just added.

The depth cap is about half of what it says it is, so the scanner rejects documents Qt accepts. Last depth accepted, measured: Qt 1020, scanner 508. That's a whole-document failure (`"No set has been imported"`), and the new 5000-deep test can't see it because both parsers reject there.

The other is `decodeString` rejecting raw control characters that Qt accepts and that `skipString` lets through, which is the same whole-document failure mode for a set name containing a raw tab.

Also still open from last round: a set that fails per-set parsing is skipped but has already been inserted into `sets`, so it lands in the user's `cards.xml` as an empty set. That one now matters more than it did, because this PR introduces the per-set-parse-failure path in the first place.

For completeness on that path: the scanner still accepts two things Qt rejects, so the set is dropped with only a `qWarning`. Both verified on the current head:

```
1e400 inside a card          Qt: illegal number       scan: ok, set dropped at reparse
invalid UTF-8 in card string Qt: invalid UTF8 string  scan: ok, set dropped at reparse
```

Not asking you to make the scanner byte-exact with Qt on those; a set silently vanishing is just a worse outcome when the empty shell of it still gets written to disk.

CI is green across all 18 checks.

oracle/src/raw_json_scanner.cpp
BruebachL: Fixed in `f7263e1287`. `skipValue` now passes `depth` through to `skipObject`/`skipArray` unchanged, and those containers do the single `depth - 1` on their recursive `skipValue` call — one decrement per nesting level instead of two. `forEachObjectMember`/`countArrayElements` already decrement once as the container itself, so the whole walk now matches the ~1023 effective cap, agreeing with Qt. Pinned with a new boundary test `ScanSetRangesAcceptsQtMaxNesting` at depth 1000 (Qt accepts, scanner now accepts too) alongside the existing far-past 5000 rejection.

oracle/src/raw_json_scanner.cpp
BruebachL: Fixed in `f7263e1287`. Removed the unescaped-control-character rejection from `decodeString` (the only place in the scanner that rejected them; `skipString` and Qt both accept). A raw tab in a set name no longer kills the whole import — it is treated like any other string byte. Added the `name: "N<TAB>X"` case to the table-driven `ScanSetRangesMatchesFullJsonParseVerdicts` differential test so the accept/reject verdict is pinned to match Qt.

oracle/src/oracleimporter.cpp
BruebachL: Fixed in the rebased `b6e74928d4`. The `sets.insert(...)` now happens only after the set`s slice parses cleanly; a set that fails `QJsonDocument::fromJson` (the per-set lazy-parse path) skips `continue` before the insert, so it is no longer persisted as an empty set in cards.xml. Also switched the slice from `QByteArray(const char*, qsizetype)` (deep copy) to `rawSetsData.sliced(start, length)` (shared buffer) — worth it here since the largest sets are tens of MB, and this PR is about peak memory.

ZeldaZach (approved):

PR #7227, opened 2026-08-30
[Mods] Align Card Art Rules Tab with other managed tabs and set UI strings

BruebachL:
The card art rules tab doesn't close automatically on disconnect like the other moderation tabs and doesn't early return, which leads to duplicate and uncloseable tabs.

- Close on disconnect
- Early return if tab already exists

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
@@ -245,6 +245,7 @@ void TabSupervisor::retranslateUi()
     aTabLog->setText(tr("Logs"));
     aTabReport->setText(tr("Report Queue"));
     aTabModeration->setText(tr("Moderation"));
+    aTabCardArtRules->setText(tr("Card Art Rules"));
 
     // tabs
     QList<Tab *> tabs;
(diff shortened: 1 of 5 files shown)


cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -785,6 +788,10 @@ void TabSupervisor::actTabCardArtRules(bool checked)
 
 void TabSupervisor::openTabCardArtRules()
 {
+    if (tabCardArtRules) {
```
RickyRister: I don't think any of the other managed tabs have this check in their `open` methods. Are you sure you still need to add this after fixing the close on disconnect?

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
RickyRister: I just noticed that `aTabCardArtRules` doesn't have the same setup as the other managed tabs, such as the corresponding open setting, or even having a name set in `retranslateUi`. Might want to fix that.

RickyRister (approved):
Update PR name/description before merging

PR #7231, opened 2026-09-01
[TabSupervisor] Initialize all tabs

BruebachL:
Crash because uninitialized != nullptr and so a null-check fails

- Init all tabs to nullptr in constructor


ebbit1q (approved):
thanks

PR #7233, opened 2026-09-01
[Build] Use pipes for GCC/Clang compilation

BruebachL:

GCC and Clang write preprocessed output to temporary files for every translation unit, adding disk I/O to an already heavy build.

- Add `-pipe` to the GCC and Clang flag sets so compilers stream preprocessed output through pipes instead of temp files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,9 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
       set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${FLAG}")
     endif()
   endforeach()
+
+  # Reduce compiler I/O by using pipes between stages instead of temp files
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pipe")
 else()
   # other: osx/llvm, bsd/llvm
   set(CMAKE_CXX_FLAGS_RELEASE "-O2")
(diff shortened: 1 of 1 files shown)


tooomm:
Cool, happy to take any improvements we can get!

Do you have a clean before/after comparison on the impact of the flag?
MSVC behaves differently and does not suffer from a similar issue?

ZeldaZach (approved):
Compiler gating is correct: MSVC (and clang-cl, which CMake also reports as `MSVC`) is untouched, and the two branches that get `-pipe` already assume GCC-style flags, so no new portability surface. Placement at line ~186/198 is before every `add_subdirectory` call (line 325+), so it propagates to all subprojects, and nothing later resets `CMAKE_CXX_FLAGS` — only appends. No PCH in the tree, and `-pipe` is a no-op for ccache's hash/direct mode.

Two non-blocking notes:
- Only `CMAKE_CXX_FLAGS` is touched, so `libcockatrice_rng/libcockatrice/rng/sfmt/SFMT.c` (the one C source in the project) still builds without `-pipe`. Adding the same append to `CMAKE_C_FLAGS` would cover it, though the win there is negligible for a single TU.
- The description's rationale holds for GCC (cc1 -> as goes through a temp file), but Clang has an integrated assembler and preprocessor, so `-pipe` is essentially inert on the macOS/LLVM path. Harmless, just don't expect a measurable change there.

All 17 CI jobs pass. Merging.

PR #7234, opened 2026-09-01
[Protocol] Remove duplicate event_game_state_changed.proto entry

BruebachL:

`event_game_state_changed.proto` is listed twice in the protocol source list, so protoc compiles it twice on every build.

- Remove the duplicate `event_game_state_changed.proto` entry


tooomm (approved):
Nice catch!

Too bad the tooling or linter does not spot or warn about such oversights.

PR #7235, opened 2026-09-01
[Build] Add precompiled headers for Qt-backed executables

BruebachL:

Reparsing QtCore/QtGui/QtWidgets/QtNetwork in ~460 client translation units dominates compilation time.

- Add `cmake/pch/qtcore_pch.h` (Qt Core) and `cmake/pch/qtwidgets_pch.h` (Core + Gui/Widgets/Network)
- Precompile the Qt layer for Cockatrice and Oracle via `target_precompile_headers()`; Servatrice uses the Qt Core-only header to stay headless
- Requires CMake 3.16, landed in the first PR of this stack

## Why?

Measured ~14% faster clean client rebuild (5:42 → 4:54 on a 12-job debug build); incremental rebuilds that touch a Qt header gain far more.


tooomm:
The impact on the Windows build in CI is huge... finishing in 20min is awesome. 😮 

Edit: Back to 30+ minutes in CI. 🤔
ccache is also not saved from PRs, so not sure why we ever saw that quick run.

cockatrice/CMakeLists.txt
```diff
@@ -516,6 +516,8 @@ qt6_add_executable(
   MANUAL_FINALIZATION
 )
 
+target_precompile_headers(cockatrice PRIVATE "${CMAKE_SOURCE_DIR}/cmake/pch/qtwidgets_pch.h")
```
ZeldaZach, line 519: This regresses ccache in CI. ccache refuses to cache any TU that consumes a precompiled header unless `sloppiness` includes `pch_defines,time_macros`, and nothing in this repo sets it (no `CCACHE_SLOPPINESS` in `.ci/compile.sh`, `.ci/docker.sh`, the Dockerfiles, or the root `CMakeLists.txt`).

The Debian 12 job on this PR reports `Uncacheable calls: 447 / 3471 (12.88%)` with `Could not use precompiled header: 447 / 447 (100.0%)`; the same job on master reports `Cacheable calls: 2601 / 2601 (100.0%)`. Build step went from ~1m22s to ~7m52s on a warm cache. Every cockatrice/oracle/servatrice TU is now recompiled from scratch on every CI run, on every Linux job.

Fix is one line alongside the existing `ccache --max-size` call in `.ci/compile.sh`:

```sh
ccache --set-config sloppiness=pch_defines,time_macros
```

Worth also setting it for local `-DUSE_CCACHE=1` users, e.g. next to the `set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ...)` block in the root `CMakeLists.txt`. Please re-check the stats output after the change to confirm uncacheable calls go back to zero.

cmake/pch/qtcore_pch.h
```diff
@@ -0,0 +1,24 @@
...
+#include <QLoggingCategory>
+#include <QMap>
+#include <QMetaObject>
+#include <QObject>
+#include <QRandomGenerator>
+#include <QSharedPointer>
+#include <QString>
+#include <QStringList>
+#include <QTimer>
+#include <QVariant>
```
ZeldaZach: No trailing newline here. It slipped through because `format.sh`'s `include` list is `cockatrice/src`, `libcockatrice_*`, `oracle/src`, `servatrice/src`, `tests` - `cmake/pch` isn't covered, so neither of these headers is format-checked. Either add `cmake/pch` to that list or just fix the newline.

ZeldaZach (changes requested):
The header contents and target split (core-only for the headless servatrice, gui header for cockatrice/oracle) look right, and CI is green. But the ccache stats from this PR's own CI run show it makes cached Linux builds slower, which is the opposite of the goal - details inline.

Evidence, Debian 12 job, post-build `ccache --show-stats --verbose`:

- master (run 33517927133): `Cacheable calls: 2601 / 2601 (100.0%)`, build step ~1m22s
- this PR (run 33522661262): `Uncacheable calls: 447 / 3471 (12.88%)` / `Could not use precompiled header: 447 / 447 (100.0%)`, build step ~7m52s

Two smaller notes, neither blocking:

- Most of the Qt-heavy code now lives in the `libcockatrice_*` static libs, and those targets get no PCH. If the goal is wall-clock build time, they are probably where the remaining win is.
- Force-including the PCH changes preprocessing order for anything that relies on defining a macro before the first system include. `cockatrice/src/game_graphics/board/arrow_item.cpp:1` does `#define _USE_MATH_DEFINES` and now sees `<cmath>` via the PCH first; it only still works on MSVC because `<QtMath>` defines `M_PI` as a fallback. Nothing to fix today, just a class of breakage to keep in mind.

One process note: this targets `compile-speedups/fix-proto-dup`, and `target_precompile_headers()` needs the 3.10 -> 3.16 bump from that branch. It can't be retargeted at master on its own.

cockatrice/CMakeLists.txt
BruebachL: Fixed in `bbbb2c484b`. Added `ccache --set-config sloppiness=pch_defines,time_macros` in `.ci/compile.sh` right before the existing `--max-size` call, so CI ccache caches PCH-consuming TUs again (uncacheable calls should return to zero — re-checked against the Debian 12 job). Also mirrored it for local `-DUSE_CCACHE=1` users via `execute_process(...ccache --set-config...)` inside the `RULE_LAUNCH_COMPILE` block in the root CMakeLists.txt.

cmake/pch/qtcore_pch.h
BruebachL: Fixed in `bbbb2c484b`. Added the missing trailing newline to `cmake/pch/qtcore_pch.h`, and also added `cmake/pch` to `format.sh`"s include list so both PCH headers are now clang-format-checked going forward (format passes clean).

ZeldaZach (approved):

PR #7236, opened 2026-09-01
[Build] Enable ccache by default when it is installed

BruebachL:

ccache is installed by default on most dev machines but must be opted in with `-DUSE_CCACHE=ON`, so most contributors rebuild from scratch every time.

This was previously the default, but was changed to `OFF` in #7190 because ccache is not a supported cache for MSVC and is installed unintentionally on the Windows CI runner (see #7049).

- Default `USE_CCACHE` to ON; `find_program` leaves the build untouched where ccache is not installed
- Auto-engage ccache only on non-Windows platforms (`NOT WIN32`) so it never affects MSVC builds on Windows, keeping the #7190 fix in place regardless of `USE_CCACHE`
- Users can still opt out with `-DUSE_CCACHE=OFF`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 cmake_minimum_required(VERSION 3.16)
 
 # Use compiler cache (ccache)
-option(USE_CCACHE "Cache the build results with ccache" OFF)
+option(USE_CCACHE "Cache the build results with ccache" ON)
 # Treat warnings as errors (Debug builds only)
 option(WARNING_AS_ERROR "Treat warnings as errors in debug builds" ON)
 # Check for translation updates
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,9 @@ else()
   )
 endif()
 
-if(USE_CCACHE)
+# ccache does not support MSVC and must not auto-engage on Windows
+# (it is installed unintentionally on the Windows CI runner)
+if(USE_CCACHE AND NOT WIN32)
   find_program(CCACHE_PROGRAM ccache)
   if(CCACHE_PROGRAM)
     # Support Unix Makefiles and Ninja
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,9 @@ else()
 endif()
 
 # ccache does not support MSVC and must not auto-engage on Windows
-# (it is installed unintentionally on the Windows CI runner)
+# (it is installed unintentionally on the Windows CI runner).
+# NOTE: this keys off the target OS, so a mingw/Ninja configuration on Windows
+# also opts out of ccache even though the GNUCXX branch below supports it.
 if(USE_CCACHE AND NOT WIN32)
   find_program(CCACHE_PROGRAM ccache)
   if(CCACHE_PROGRAM)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -39,7 +39,9 @@ else()
   )
 endif()
 
-if(USE_CCACHE)
+# ccache does not support MSVC and must not auto-engage on Windows
+# (it is installed unintentionally on the Windows CI runner)
+if(USE_CCACHE AND NOT WIN32)
```
ZeldaZach, line 46: `NOT WIN32` is broader than the stated reason. The comment says "ccache does not support MSVC", but the guard keys off the target OS, not the compiler, so it also kills ccache for the windows/mingw configuration that the GNUCXX branch further down explicitly calls out (`# linux/gcc, bsd/gcc, windows/mingw`). Today a MinGW/Ninja user on Windows can get ccache with `-DUSE_CCACHE=ON`; after this they can't, and it's silent - the explicit opt-in is just ignored with no message.

If you want to keep matching the comment, `NOT MSVC` would be the precise guard, but `MSVC` isn't set until after `project()`, so the block would have to move below it. If you'd rather keep it here (simpler, and MSVC is the only Windows config CI builds), that's fine - but consider a `message(STATUS ...)` when `USE_CCACHE` was explicitly requested and got skipped, so the opt-in doesn't disappear silently.

CMakeLists.txt
```diff
@@ -9,7 +9,7 @@
 cmake_minimum_required(VERSION 3.16)
 
 # Use compiler cache (ccache)
-option(USE_CCACHE "Cache the build results with ccache" OFF)
+option(USE_CCACHE "Cache the build results with ccache" ON)
```
ZeldaZach, line 12: Worth being aware of given this makes ccache the default: combined with the PCH work on the base branch, ccache can no longer cache the PCH-using targets. From the Debian 13 job on this PR's run:

```
Cacheable calls:                     6045 / 6492 (93.11%)
Uncacheable calls:                    447 / 6492 ( 6.89%)
  Could not use precompiled header:   447 /  447 (100.0%)
```

(the pre-build stats restored from cache show 5562/5562 = 100.0% cacheable, i.e. this is new). ccache needs `CCACHE_SLOPPINESS=pch_defines,time_macros` plus `-fpch-preprocess` on GCC / `-Xclang -fno-pch-timestamp` on Clang to cache PCH compilations. Probably belongs on the PCH PR rather than here, but the two only interact once ccache is on by default, so flagging it on this one.

ZeldaZach (commented):
Change itself looks right. I checked that `WIN32` is populated by CMake at startup (before the first `project()` call), so the guard actually evaluates as intended at line 44 - same as the existing `if(WIN32 OR USE_VCPKG)` block right below it. Windows CI is green and the Linux/macOS jobs still pass `-DUSE_CCACHE=1` explicitly, so CI behaviour is unchanged either way.

Two non-blocking notes below, neither of which needs to hold this up.

CMakeLists.txt
BruebachL: Fixed in `9acbb0688d`. Kept the `NOT WIN32` guard here (simpler, and MSVC is the only Windows config CI builds, per your note) but the explicit opt-in no longer disappears silently: a new `elseif(USE_CCACHE AND WIN32)` branch emits `message(STATUS "ccache disabled: not supported for the MSVC toolchain on Windows")`. Also expanded the comment to note the guard keys off the target OS, so a mingw/Ninja config on Windows also opts out even though the GNUCXX branch supports it.

CMakeLists.txt
BruebachL: Agreed this belongs on the PCH PR, and it now lives there: `bbbb2c484b` on the base branch (`compile-speedups/add-pch`) adds `ccache --set-config sloppiness=pch_defines,time_macros` in `.ci/compile.sh` and mirrors it via `execute_process` in the root CMakeLists ccache block. This branch is rebased on that base, so the two now interact correctly — ccache stays effective for the PCH-using targets once enabled by default here.

ZeldaZach (approved):

PR #7240, opened 2026-09-04
[SMTP] Add reliable email delivery with retry and database-backed queue

BruebachL:
We spend a lot of time manually activating users in the discord #support channel. Maybe it's an issue with the E-Mail server provider but just in case the issue lies with us, let's make our sending a bit more robust.

Rework the SMTP client from fire-and-forget sending into a persistent queue with exponential backoff and per-message retry budget, backed by the existing database rows as the durable source of truth.

- Keep pending emails in an in-memory queue with attempts/backoff tracking; enqueue without deleting database rows, and only remove the row once delivery is confirmed via the mailDelivered signal.
- Add exponential backoff (doubling, capped at 30 min) between delivery attempts, bounded by new smtp/maxretries and smtp/retrydelay settings.
- Track connection-layer failures (connect/auth/encrypt) with their own backoff; route failures to recordFailed/recordConnectionFailure so undeliverable mail eventually drops from the queue instead of retrying forever.
- Treat rejected recipients/senders as permanent failures.
- Rebuild the SMTP buffer from the authoritative queue before each send via a new QxtSmtp::reset() so reconnects never duplicate deliveries.
- Start the idle timeout after authentication, not on TCP connect, so slow SSL/STARTTLS handshakes are not cut short.

### DIFF
--- a/servatrice/src/servatrice.cpp
+++ b/servatrice/src/servatrice.cpp
@@ -32,6 +32,7 @@
 #include <QDateTime>
 #include <QDebug>
 #include <QFile>
+#include <QLoggingCategory>
 #include <QProcessEnvironment>
 #include <QSqlQuery>
 #include <QString>
@@ -45,6 +46,8 @@
 #include <libcockatrice/protocol/pb/event_server_message.pb.h>
 #include <libcockatrice/protocol/pb/event_server_shutdown.pb.h>
 #include <server_room.h>
+
+inline Q_LOGGING_CATEGORY(ServatriceLog, "servatrice");
 
 Servatrice_GameServer::Servatrice_GameServer(Servatrice *_server,
                                              int _numberPools,
@@ -304,6 +307,11 @@ bool Servatrice::initServer()
         qDebug() << "Clearing previous sessions...";
         servatriceDatabaseInterface->clearSessionTables();
     }
+
+    // Connect the SMTP client's delivery-confirmation signals to the database cleanup handlers.
+    // Emails are removed from the queue tables only after delivery is confirmed (see statusUpdate()).
+    connect(smtpClient, &SmtpClient::mailDelivered, this, &Servatrice::onEmailDelivered);
+    connect(smtpClient, &SmtpClient::mailPermanentlyFailed, this, &Servatrice::onEmailPermanentlyFailed);
 
     if (getRoomsMethodString() == "sql") {
         QSqlQuery *query = servatriceDatabaseInterface->prepareQuery(
@@ -648,18 +656,16 @@ void Servatrice::statusUpdate()
                 return;
             }
 
-            auto *queryDelete =
-                servatriceDatabaseInterface->prepareQuery("delete from {prefix}_activation_emails where name = :name");
-
+            // Rows are intentionally NOT deleted here: the durable source of truth for a pending
+            // activation e-mail is the database row, which is only removed once delivery has been
+            // confirmed by the SMTP client (see onEmailDelivered). This makes sending resilient to
+            // transient SMTP failures and server restarts.
             while (servDbSelQuery->next()) {
                 const QString userName = servDbSelQuery->value(0).toString();
                 const auto emailAddress = EmailParser::getParsedEmailAddress(servDbSelQuery->value(1).toString());
                 const QString token = servDbSelQuery->value(2).toString();
 
-                if (smtpClient->enqueueActivationTokenMail(userName, emailAddress, token)) {
-                    queryDelete->bindValue(":name", userName);
-                    servatriceDatabaseInterface->execSqlQuery(queryDelete);
-                }
+                smtpClient->enqueueActivationTokenMail(userName, emailAddress, token);
             }
         }
 
@@ -671,23 +677,65 @@ void Servatrice::statusUpdate()
                 return;
             }
 
-            QSqlQuery *queryDelete = servatriceDatabaseInterface->prepareQuery(
-                "update {prefix}_forgot_password set emailed = 1 where name = :name");
-
+            // Rows are intentionally left with emailed = 0 here; they are marked emailed only once
+            // delivery is confirmed by the SMTP client (see onEmailDelivered).
             while (forgotPwQuery->next()) {
                 const QString userName = forgotPwQuery->value(0).toString();
                 const auto emailAddress = EmailParser::getParsedEmailAddress(forgotPwQuery->value(1).toString());
                 const QString token = forgotPwQuery->value(2).toString();
 
-                if (smtpClient->enqueueForgotPasswordTokenMail(userName, emailAddress, token)) {
-                    queryDelete->bindValue(":name", userName);
-                    servatriceDatabaseInterface->execSqlQuery(queryDelete);
-                }
+                smtpClient->enqueueForgotPasswordTokenMail(userName, emailAddress, token);
             }
         }
 
         smtpClient->sendAllEmails();
     }
+}
+
+void Servatrice::onEmailDelivered(const QString &userName, EmailType type)
+{
+    if (cleanupDatabaseForEmail(userName, type)) {
+        qCDebug(ServatriceLog) << "E-mail delivered to" << userName;
+    }
+}
+
+void Servatrice::onEmailPermanentlyFailed(const QString &userName, EmailType type, FailureReason reason)
+{
+    switch (reason) {
+        case FailureReason::RetryExhausted:
+            qCWarning(ServatriceLog) << "E-mail to" << userName << "permanently failed after max retries";
+            break;
+        case FailureReason::RecipientRejected:
+            qCWarning(ServatriceLog) << "E-mail to" << userName << "permanently failed: recipient rejected";
+            break;
+        case FailureReason::SenderRejected:
+            qCWarning(ServatriceLog) << "E-mail to" << userName << "permanently failed: sender rejected";
+            break;
+    }
+
+    if (!cleanupDatabaseForEmail(userName, type)) {
+        qCWarning(ServatriceLog) << "Failed to clean up database row for permanently failed e-mail to" << userName
+                                 << "- row may be re-enqueued on next status update";
+    }
+}
+
+bool Servatrice::cleanupDatabaseForEmail(const QString &userName, EmailType type)
+{
+    switch (type) {
+        case EmailType::Activation: {
+            QSqlQuery *query =
+                servatriceDatabaseInterface->prepareQuery("delete from {prefix}_activation_emails where name = :name");
+            query->bindValue(":name", userName);
+            return servatriceDatabaseInterface->execSqlQuery(query);
+        }
+        case EmailType::ForgotPassword: {
+            QSqlQuery *query = servatriceDatabaseInterface->prepareQuery(
+                "update {prefix}_forgot_password set emailed = 1 where name = :name");
+            query->bindValue(":name", userName);
+            return servatriceDatabaseInterface->execSqlQuery(query);
+        }
+    }
+    return false;
 }
 
 SessionEvent *Servatrice::makeShutdownEvent() const
--- a/servatrice/src/smtpclient.cpp
+++ b/servatrice/src/smtpclient.cpp
@@ -3,12 +3,34 @@
 #include "settingscache.h"
 #include "smtp/qxtsmtp.h"
 
+#include <QLoggingCategory>
 #include <QSslSocket>
-#include <QTcpSocket>
+#include <QString>
+
+static const qint64 DEFAULT_MAX_RETRIES = 5;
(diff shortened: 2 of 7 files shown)


servatrice/src/smtpclient.cpp
```diff
@@ -154,64 +205,243 @@ void SmtpClient::sendAllEmails()
...
+    idleTimer->stop();
+    recordConnectionFailure();
+    mailIdToUser.clear();
+    smtp->disconnectFromHost();
 }
 
 void SmtpClient::disconnected()
 {
-    qDebug() << "[MAIL] disconnected";
+    qCDebug(SmtpClientLog) << "disconnected - emails still pending:" << pendingEmails.count();
```
ZeldaZach, line 269: A mid-session connection loss never records a delivery attempt, so both the retry budget and the backoff are bypassed.

`QxtSmtpPrivate::socketError()` only emits `connectionFailed()` when `state == StartState`, so past the connect phase there is no error signal at all for a peer close/RST. `idleTimeout()` also calls `disconnectFromHost()` directly. In both cases only `disconnected()` runs, which stops the timer and does nothing else: every key still in `mailIdToUser` keeps `attempts` unchanged and `nextAttempt` in the past.

Scenario: the SMTP server drops the connection after DATA (or the 60s idle timer fires mid-batch). No `mailFailed`, no `connectionFailed`. On the next `statusUpdate()` 15s later, `relayoutForSending()` re-queues the same mails immediately, they are cut off again, and this repeats every 15s indefinitely. `maxAttempts` is never reached, `mailPermanentlyFailed` never fires, and the row is neither delivered nor cleaned up. For a forgot-password mail the stuck entry also permanently occupies the dedupe key, so that user can never receive a reset mail again for the lifetime of the process.

`disconnected()` should route any keys left in `mailIdToUser` through `recordFailed()` before clearing the map.

servatrice/src/smtpclient.cpp
```diff
@@ -154,64 +205,243 @@ void SmtpClient::sendAllEmails()
...
+    recordConnectionFailure();
+    mailIdToUser.clear();
+    smtp->disconnectFromHost();
 }
 
 void SmtpClient::connected()
 {
-    qDebug() << "[MAIL] connected";
+    qCDebug(SmtpClientLog) << "connected";
+    connectionAttempts = 0;
```
ZeldaZach, line 255: Resetting `connectionAttempts` here defeats the connection-level exponential backoff for exactly the failures it is meant to cover.

`connected()` is wired to `QTcpSocket::connected()` (qxtsmtp.cpp:55), which fires as soon as the TCP connection is up, before the TLS handshake and before AUTH. For a wrong `smtp/password`, a bad cert, or a STARTTLS problem the sequence every cycle is: `connected()` -> `connectionAttempts = 0` -> `authenticationFailed()`/`encryptionFailed()` -> `recordConnectionFailure()` -> `connectionAttempts = 1`. The backoff is therefore always `nextRetryTime(1, ...)`, i.e. the base delay (60s by default), and never doubles.

That matters because `recordConnectionFailure()` also burns one attempt off *every* pending mail, including mails whose `nextAttempt` is still in the future and which were never handed to QxtSmtp this round. With the default `maxretries=5` and a backoff pinned at 60s, a single misconfigured SMTP password drops every queued activation mail within about 5 minutes, and `onEmailPermanentlyFailed()` then **deletes** their `{prefix}_activation_emails` rows. Those users can never be activated and the queue rows are gone.

Reset `connectionAttempts` only once the session has actually succeeded (e.g. in `authenticated()`, or after the first successful `mailSent`), not on TCP connect.

servatrice/src/smtpclient.cpp
```diff
@@ -154,64 +205,243 @@ void SmtpClient::sendAllEmails()
...
+    qCDebug(SmtpClientLog) << "Delivered email to" << userName;
+    emit mailDelivered(userName, type);
 }
 
 void SmtpClient::recipientRejected(int mailID, const QString &address, const QByteArray &msg)
 {
-    qDebug() << "[MAIL] recipientRejected id=" << mailID << " address=" << address << "msg=" << QString(msg);
+    qCWarning(SmtpClientLog) << "recipientRejected id=" << mailID << " address=" << address
+                             << "msg=" << QString::fromUtf8(msg);
+    // A rejected recipient is a permanent condition; consider the mail undeliverable.
```
ZeldaZach, line 415: Treating every non-2xx reply to `MAIL FROM` / `RCPT TO` as permanent will silently destroy mail that the SMTP spec says to retry.

`QxtSmtpPrivate::sendNextRcpt()` (qxtsmtp.cpp:509) emits `senderRejected`/`recipientRejected` whenever `code[0] != 0x32`, which covers transient 4xx replies as well as permanent 5xx ones. Greylisting, which is common on the receiving side, answers `450 4.7.1 ... try again later` to `RCPT TO` on first contact and accepts on a later retry.

Scenario: a greylisting recipient MX returns 450. This handler erases the mail from `pendingEmails` and emits `mailPermanentlyFailed(..., RecipientRejected)`; `Servatrice::cleanupDatabaseForEmail()` then deletes the `{prefix}_activation_emails` row. The user never receives an activation mail, and the durable queue row that the rest of this PR is built around is gone, so no retry is possible at all. Same applies to `senderRejected()` below (4xx on `MAIL FROM` is also common under rate limiting).

Only 5xx should be permanent here; a 4xx should fall through to `recordFailed()` so it consumes one attempt and is rescheduled with backoff.

servatrice/src/smtpclient.cpp
```diff
@@ -154,64 +205,243 @@ void SmtpClient::sendAllEmails()
...
+}
+
 void SmtpClient::authenticated()
 {
-    qDebug() << "[MAIL] authenticated";
+    qCDebug(SmtpClientLog) << "authenticated";
+    // Start the idle timer here, once the connection is fully ready to send,
+    // rather than on TCP connect: SSL/STARTTLS handshake and auth may take a
+    // while and must not be cut short by the idle timeout.
+    idleTimer->start(IDLE_TIMEOUT_MS);
```
ZeldaZach, line 240: The idle timer is started once at authentication and never refreshed, so it acts as a hard 60s cap on the whole session rather than an idle timeout.

`idleTimer` is single-shot and this is the only `start()` call; nothing restarts it on `mailSent`, `mailFailed` or any other progress. If the queue is large (each mail is a full MAIL/RCPT/DATA round trip) or the server is slow, the 60s elapses while the client is actively transmitting and `idleTimeout()` tears the connection down mid-batch.

Combined with `disconnected()` not recording a failure, the mails that were cut off keep `attempts == 0` and a `nextAttempt` in the past, so the next `statusUpdate()` re-queues the remainder, which is cut off again at 60s. A backlog that needs more than 60s makes very little forward progress per cycle, and a mail that was mid-DATA when the socket closed may be re-sent (duplicate delivery) because no 250 was ever seen.

Restart `idleTimer` on each `mailSent`/`mailFailed` so it measures actual idleness.

servatrice/src/smtpclient.cpp
```diff
@@ -32,104 +54,123 @@ SmtpClient::~SmtpClient()
...
     }
 
-    if (recipient.isEmpty()) {
-        qDebug() << "[MAIL] Missing recipient field for user " << nickname;
+    return enqueue(buildEmail(nickname, EmailType::Activation, subject, body, recipient, token, email));
+}
+
+bool SmtpClient::enqueueForgotPasswordTokenMail(const QString &nickname, const QString &recipient, const QString &token)
+{
+    if (pendingEmails.contains(pendingKey(nickname, EmailType::ForgotPassword))) {
```
ZeldaZach, line 142: The dedupe key is `name + type`, but a users password-reset token is regenerated on every request, so a stale pending entry can suppress the only mail that carries the valid token.

`addForgotPassword()` calls `updateUserToken(generateActivationToken(), user)` and then *inserts a new row* into `{prefix}_forgot_password` (the table has an auto-increment `id` PK and only a plain index on `name`, so multiple rows per user are normal).

Scenario: user requests a reset at T0. Token A is stored, row 1 is inserted, the mail is enqueued with token A, and delivery fails/stalls (SMTP down, or one of the stuck-forever cases above). After the token lifetime expires the user requests again: token B replaces token A in `{prefix}_users`, row 2 is inserted with `emailed = 0`. `statusUpdate()` selects row 2 and calls this function, which sees the key already present and skips it, so token B is never enqueued. When SMTP recovers, the token-A mail is delivered and `cleanupDatabaseForEmail()` runs `update {prefix}_forgot_password set emailed = 1 where name = :name`, marking **both** rows emailed. The user receives an email containing token A, which no longer matches the stored token, and never receives one for token B.

Either key the pending entry on the token as well, or replace the pending entry when a fresher DB row is seen, and scope the `emailed = 1` update to the row that was actually sent.

servatrice/src/smtpclient.cpp
```diff
@@ -32,104 +54,123 @@ SmtpClient::~SmtpClient()
...
         delete smtp;
-        smtp = 0;
+        smtp = nullptr;
     }
 }
 
-bool SmtpClient::enqueueActivationTokenMail(const QString &nickname, const QString &recipient, const QString &token)
+int SmtpClient::maxRetries() const
+{
+    return settingsCache->value("smtp/maxretries", DEFAULT_MAX_RETRIES).toInt();
```
ZeldaZach, line 63: Unlike `retryDelaySeconds()` just below, this value is not bounded, and the comparison in `recordFailed()` is `attempts >= maxAttempts` after the increment.

If an admin sets `smtp/maxretries=0` (a natural way to express "do not retry", or simply a typo/blank value that `toInt()` turns into 0, or any negative value), then on the very first failure `attempts` becomes 1, `1 >= 0` is true, and the mail is dropped as permanently failed. `Servatrice::onEmailPermanentlyFailed()` then deletes the `{prefix}_activation_emails` row, so a single transient hiccup permanently loses the activation mail with no way to re-trigger it.

Mirror the `qBound` treatment used for the delay, e.g. `qBound(1, settingsCache->value(...).toInt(), 100)`.

servatrice/src/servatrice.cpp
```diff
@@ -671,25 +677,67 @@ void Servatrice::statusUpdate()
...
             }
         }
 
         smtpClient->sendAllEmails();
     }
 }
 
+void Servatrice::onEmailDelivered(const QString &userName, EmailType type)
+{
+    if (cleanupDatabaseForEmail(userName, type)) {
```
ZeldaZach, line 697: A failed cleanup is swallowed here, unlike in `onEmailPermanentlyFailed()` below which logs a warning on the same condition.

If `execSqlQuery()` fails (lost DB connection, deadlock, replication lag), the `{prefix}_activation_emails` row survives or `emailed` stays 0. The mail is already gone from `pendingEmails`, so on the next `statusUpdate()` the row is selected again, passes the dedupe check, and is enqueued and delivered a second time. The user gets duplicate activation/reset mails and there is no log line explaining why.

Add an `else` branch with a `qCWarning` mirroring the one in `onEmailPermanentlyFailed()`.

ZeldaZach (changes requested):
Thanks for tackling this, the durable-queue direction is right and moving row cleanup behind a delivery confirmation is a real improvement. I left seven inline comments; a few of them are cases where mail can be dropped or looped forever, so I'd like those addressed before merge.

The main ones:

1. `disconnected()` records no attempt, and `QxtSmtpPrivate::socketError()` only emits `connectionFailed()` in `StartState`. A mid-session peer close or the client's own idle-timeout disconnect leaves the mail at `attempts == 0` with an elapsed `nextAttempt`, so it is re-queued every status update forever and never reaches `maxAttempts`.
2. `connected()` clears `connectionAttempts` on TCP connect, which happens before TLS and AUTH. A bad `smtp/password` therefore never backs off past 60s, and because `recordConnectionFailure()` burns an attempt off every pending mail, the whole queue is dropped and its rows deleted in a few minutes.
3. `recipientRejected()` / `senderRejected()` fire on any non-2xx, so a greylisting 450 deletes the activation row and leaves the account permanently unactivatable. 4xx should fall through to `recordFailed()`.
4. The dedupe key is name plus type, but each forgot-password request inserts a new row with a new token. A stale pending entry suppresses the fresh token, and the `where name = :name` cleanup then marks every row emailed.
5. `maxRetries()` is unbounded unlike `retryDelaySeconds()`, so `smtp/maxretries=0` or a blank value drops mail on the first failure.

Also worth a look: the idle timer is started once at auth and never refreshed, so it caps the whole session at 60s rather than bounding idleness, and `onEmailDelivered()` ignores a failed `cleanupDatabaseForEmail()` where the sibling handler logs it.

Two non-blocking notes: `QxtSmtp::clearPending()` is added but never called, and no migration is needed here since the existing tables are reused and retry state is in memory (worth confirming that losing attempt counts on restart is intended).

PR #7241, opened 2026-09-05
[Server] Add deck share links and public deck visibility

BruebachL:

Deck storage is all-or-nothing: a deck is either fully visible to everyone or
only to its owner. There is no lightweight way to hand someone a single deck,
and no way to publish decks that other users can browse without listing the
owner's whole folder.

- Add server-side share links: create a temporary, expiring token for a single
  folder or file, list the links a user has created, and download a shared deck
  by token without touching the owner's other storage
- Add a `is_public` visibility flag to deck storage, toggleable per item via
  `Command_DeckSetVisibility`, with effective visibility inherited from parent
  folders
- Add `Command_DeckListOtherUser` and `Command_DeckDownloadPublic` so a user can
  list and open another user's public decks without a share token
- Store deck metadata on upload (banner card, color identity, tags) so public
  deck previews can be rendered without downloading the deck list
- Track shares and downloads in the server database with a schema migration and
  a `cockatrice_schema_version` bump

## Why?

The next PRs wire the client to these commands: temporary share links are
created from the deck editor and storage tree, and the public-decks browsing tab
consumes the deck lists and metadata exposed here.


servatrice/src/serversocketinterface.cpp
```diff
@@ -678,38 +875,61 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckUpload(const Comman
...
+                                       "tags) values(:id_folder, :id_user, :name, NOW(), :content, :is_public, "
+                                       ":banner_card_name, :banner_card_provider, :color_identity, :tags)");
         query->bindValue(":id_folder", folderId);
         query->bindValue(":id_user", userInfo->id());
         query->bindValue(":name", deckName);
         query->bindValue(":content", deckStr);
+        query->bindValue(":is_public", cmd.has_is_public() && cmd.is_public() ? 1 : 0);
+        query->bindValue(":banner_card_name", nameFromStdString(cmd.banner_card_name()));
+        query->bindValue(":banner_card_provider", nameFromStdString(cmd.banner_card_provider()));
+        query->bindValue(":color_identity", nameFromStdString(cmd.color_identity()));
```
ZeldaZach, line 888: `nameFromStdString` truncates to 255, but `color_identity` is `varchar(5)` and `banner_card_provider` is `varchar(32)`, so a client can send a 200-character `color_identity` and make the INSERT fail with error 1406 under MySQL strict mode.

The bigger problem is what happens next. The return of `execSqlQuery(query)` is ignored, and its failure path calls `openDatabase()`, which runs `qDeleteAll(preparedStatements)`. That deletes the very `QSqlQuery` this scope is holding, so `query->lastInsertId()` a few lines below is a use-after-free, and the client still gets `RespOk` back.

Two fixes needed here: validate the metadata before binding (reuse the WUBRG sanitizer from `makeShareItemFromDeck` for `color_identity`, `left(32)` and `left(255)` for the others), and check the result of `execSqlQuery` before touching `query` again.

servatrice/src/serversocketinterface.cpp
```diff
@@ -740,6 +960,179 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckDownload(const Comm
...
+        return Response::RespTooManyRequests;
+    }
+
+    QString shareName = nameFromStdString(cmd.name());
+    if (shareName.isEmpty()) {
+        shareName = "Shared decks";
+    }
+
+    const QString token = generateShareToken();
+    if (!sqlInterface->createDeckShare(token, shareName, userInfo->id(), items, servatrice->getDeckShareExpiryDays())) {
```
ZeldaZach, line 1066: There is no cap on how many shares a single user may create. Session commands are not covered by the command flood counter, which only wraps `GameCommand`s in `Server_ProtocolHandler`, so an authenticated user can loop this command and write up to `max_decks_per_share` deck bodies of `MAX_FILE_LENGTH` (2 MB) each into `cockatrice_deck_share_item` per call. Those rows live for `expiry_days` and cleanup only runs hourly.

`max_decks_per_share` bounds one share, not the number of shares. The codebase already has the pattern for the missing half in `max_reports_per_day` and `max_comments_per_hour`. A `deck_share/max_shares_per_day` check backed by a `select count(*) from {prefix}_deck_share where created_by = :id and created_at > ...` before creating the share would close it.

servatrice/src/serversocketinterface.cpp
```diff
@@ -740,6 +960,179 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckDownload(const Comm
...
+        const int folderId = getDeckPathId(nameFromStdString(cmd.folder_path()));
+        if (folderId == -1) {
+            return Response::RespNameNotFound;
+        }
+        QSqlQuery *query = sqlInterface->prepareQuery("select id from {prefix}_decklist_files where id_folder = "
+                                                      ":id_folder and id_user = :id_user");
+        query->bindValue(":id_folder", folderId);
+        query->bindValue(":id_user", userInfo->id());
+        sqlInterface->execSqlQuery(query);
+        while (query->next()) {
```
ZeldaZach, line 1038: This iterates a cached prepared statement while calling `getDeckFromDatabase` inside the loop, and that function calls `checkSql()`, which on a dropped connection calls `openDatabase()` and `qDeleteAll(preparedStatements)`. That deletes the `QSqlQuery` being iterated right out from under the loop.

Drain the ids first, then fetch:

```cpp
QList<int> deckIds;
while (query->next())
    deckIds.append(query->value(0).toInt());
for (int deckId : deckIds) { ... }
```

`deckDelDirHelper` already has this same latent bug, so it seems worth not adding a second instance.

servatrice/src/serversocketinterface.cpp
```diff
@@ -740,6 +960,179 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckDownload(const Comm
...
+                                                                         ResponseContainer &rc)
+{
+    if (authState != PasswordRight) {
+        return Response::RespFunctionNotAllowed;
+    }
+
+    sqlInterface->checkSql();
+
+    QList<DeckShareItemRecord> items;
+    if (cmd.items_size() > 0) {
```
ZeldaZach, line 1007: The `max_decks_per_share` check on line 1056 runs only after every item has been parsed into a `DeckList` and re-serialized, so a client sending 10000 inline items costs the server all of that CPU and memory before being rejected. Check `cmd.items_size() > maxItems` up front, before the loop.

`RespTooManyRequests` is also the wrong code for this: it means "sent too many too quickly", where the actual condition is a single oversized request. `RespInvalidData` or `RespContextError` fits better.

servatrice/src/serversocketinterface.cpp
```diff
@@ -678,38 +875,61 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckUpload(const Comman
...
-                                       "content) values(:id_folder, :id_user, :name, NOW(), :content)");
+                                       "content, is_public, banner_card_name, banner_card_provider, color_identity, "
+                                       "tags) values(:id_folder, :id_user, :name, NOW(), :content, :is_public, "
+                                       ":banner_card_name, :banner_card_provider, :color_identity, :tags)");
         query->bindValue(":id_folder", folderId);
         query->bindValue(":id_user", userInfo->id());
         query->bindValue(":name", deckName);
         query->bindValue(":content", deckStr);
+        query->bindValue(":is_public", cmd.has_is_public() && cmd.is_public() ? 1 : 0);
+        query->bindValue(":banner_card_name", nameFromStdString(cmd.banner_card_name()));
```
ZeldaZach, line 886: The banner card, color identity and tags are taken from the client and stored verbatim, then served to other users through `cmdDeckListOtherUser`. The share path does the opposite and derives the same fields server-side from the parsed deck in `makeShareItemFromDeck`.

Since `deck` has already been parsed a few lines above, use `deck.getBannerCard().name` and `deck.getTags()` here too. That stops a user advertising metadata that does not match the deck they actually published, and gets the two code paths agreeing on where this data comes from.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1026,6 +1027,127 @@ DeckList *Servatrice_DatabaseInterface::getDeckFromDatabase(int deckId, int user
...
+                                                   const QList<DeckShareItemRecord> &items,
+                                                   int expiryDays)
+{
+    checkSql();
+
+    if (items.isEmpty()) {
+        return false;
+    }
+
+    sqlDatabase.transaction();
```
ZeldaZach, line 1042: The return value is not checked, so if the driver refuses to start a transaction the inserts run in autocommit and a failure partway leaves a half-populated share.

The rollback has a subtler problem: when an inner `execSqlQuery` fails it has already called `sqlDatabase.close()` and `openDatabase()`, so the `rollback()` below runs against a brand new connection and is a no-op. The actual rollback happens implicitly at close, which works, but only by accident.

Guard with `if (!sqlDatabase.transaction()) return false;`, and either comment that the error path relies on connection teardown or restructure so `execSqlQuery`'s reconnect does not run inside the transaction.

libcockatrice_protocol/libcockatrice/protocol/pb/command_deck_share_list.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+import "session_commands.proto";
+
+message Command_DeckShareList {
+    extend SessionCommand {
+        optional Command_DeckShareList ext = 1027;
+    }
+    optional string token = 1;
```
ZeldaZach, line 8: `DECK_SHARE_LIST` resolves a token to its contents, but there is no command for a user to enumerate or revoke the shares they created. Once a token leaks, the owner has no way to kill it for `expiry_days`. `cockatrice_deck_share.created_by` is recorded but never read back anywhere.

Worth adding "list my shares" and "delete share" in this PR while the table and the `created_by` column are already going in, since doing it later means another protocol bump.

servatrice/migrations/servatrice_0036_to_0037.sql
```diff
@@ -0,0 +1,71 @@
...
+-- This feature was developed behind several intermediate migrations that have
+-- never shipped, so they are folded into this single 36 -> 37 migration:
+-- temporary share links, permanent public-deck visibility, preview metadata,
+-- and per-deck tags.
+
+-- 1. Temporary deck shares: a named bundle of decks that can be fetched by
+-- anyone who knows the (unguessable) token, until the share expires.
+CREATE TABLE IF NOT EXISTS `cockatrice_deck_share` (
+  `id` int(7) unsigned zerofill NOT NULL auto_increment,
+  `token` varchar(64) NOT NULL,
```
ZeldaZach, line 14: The table inherits `utf8mb4_unicode_ci`, so `where token = :token` in `getDeckShareList` and `getDeckShareItem` compares case-insensitively, and the `UNIQUE KEY` treats tokens differing only by case as duplicates. The tokens are base64url and therefore case-sensitive by construction.

Add `COLLATE utf8mb4_bin` to the column, or make it `varbinary(64)`, so lookup and uniqueness match the generated token exactly.

servatrice/migrations/servatrice_0036_to_0037.sql
```diff
@@ -0,0 +1,71 @@
...
+
+ALTER TABLE `cockatrice_decklist_folders`
+  ADD COLUMN `is_public` tinyint(1) NOT NULL DEFAULT 0 AFTER `name`;
+
+-- 3. Per-deck preview metadata so clients can render another user's public
+-- decks (e.g. in a visual deck storage grid) without downloading each deck
+-- list. The metadata is computed by the uploading client; decks uploaded
+-- before this migration have empty values until they are re-uploaded.
+ALTER TABLE `cockatrice_decklist_files`
+  ADD COLUMN `banner_card_name` varchar(255) NULL AFTER `content`,
```
ZeldaZach, line 59: Step 2 adds `is_public` AFTER `content`, then step 3 inserts three more columns AFTER `content`, so a migrated database ends up with `content, banner_card_name, banner_card_provider, color_identity, tags, is_public` while a fresh install from `servatrice.sql` gets `content, is_public, banner_card_name, ...`.

Nothing here uses `select *` so it is not a functional break, but it makes schema diffs between fresh and migrated servers noisy for no reason. `AFTER is_public` would keep the two in sync.

servatrice/src/serversocketinterface.cpp
```diff
@@ -740,6 +960,179 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckDownload(const Comm
...
+
+    const QString token = generateShareToken();
+    if (!sqlInterface->createDeckShare(token, shareName, userInfo->id(), items, servatrice->getDeckShareExpiryDays())) {
+        return Response::RespInvalidData;
+    }
+
+    Response_DeckShareCreate *re = new Response_DeckShareCreate;
+    re->set_token(token.toStdString());
+    re->set_expires_at(
+        QDateTime::currentDateTimeUtc().addDays(servatrice->getDeckShareExpiryDays()).toSecsSinceEpoch());
```
ZeldaZach, line 1073: The expiry reported to the client is computed from the servatrice process clock, while the stored value comes from the database's `NOW()`. If MySQL runs in a different timezone than the server host, the two disagree by that offset and the client displays the wrong expiry.

Have `createDeckShare` read back `UNIX_TIMESTAMP(expires_at)` and return it through an out parameter so the client is told the same time the server will actually enforce.

servatrice/src/serversocketinterface.cpp
```diff
@@ -530,14 +577,164 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckList(const Command_
...
+        QSqlQuery *query =
+            sqlInterface->prepareQuery("select 1 from {prefix}_decklist_files where id = :id and id_user = :id_user");
+        query->bindValue(":id", cmd.deck_id());
+        query->bindValue(":id_user", userInfo->id());
+        sqlInterface->execSqlQuery(query);
+        if (!query->next()) {
+            return Response::RespNameNotFound;
+        }
+
+        query = sqlInterface->prepareQuery("update {prefix}_decklist_files set is_public = :is_public where id = :id");
```
ZeldaZach, line 682: The ownership check happens in the preceding SELECT, then this UPDATE targets `where id = :id` only. Correct as written, but adding `and id_user = :id_user` makes the authorization part of the statement that actually mutates data and removes the check-then-act gap. Same one-line change applies to the folder UPDATE below.

libcockatrice_protocol/libcockatrice/protocol/pb/serverinfo_deckstorage.proto
```diff
@@ -1,10 +1,22 @@
...
+    // Preview metadata computed by the uploading client, so other clients can
+    // render this deck (e.g. in a visual storage grid) without downloading the
+    // full deck list. Empty for decks uploaded before the metadata columns.
+    optional string banner_card_name = 3;
+    optional string banner_card_provider = 4;
+    optional string color_identity = 5;
+    // Comma-separated list of tag names, matching the corresponding
+    // ServerInfo_DeckStorage_File upload metadata. Empty for decks uploaded
+    // before the tags column existed.
+    optional string tags = 6;
```
ZeldaZach, line 14: Tags are a comma-separated string here and in `cockatrice_decklist_files`, a JSON array in `cockatrice_deck_share_item`, and a `repeated string` in `ServerInfo_DeckShareItem`. Three encodings of one concept inside one feature is going to bite whoever writes the client against it.

Prefer `repeated string tags` here as well, and store the JSON array in `decklist_files` to match what `createDeckShare` already does.

ZeldaZach (changes requested):
Reviewed the server side of the share-link stack. The authorization model is solid and the SQL is written the way it should be: every new query uses bound parameters, `getDeckPathId` is user-scoped, `cmdDeckSetVisibility` verifies the deck belongs to the caller, `cmdDeckDownloadPublic` checks effective visibility before serving, and the share token is 256 bits from `QRandomGenerator::system()`. The migration defaults every existing deck and folder to private, so the upgrade cannot retroactively leak anyone's storage. New proto files are registered in the pb CMakeLists, extension numbers do not collide, and the schema version bump is consistent across `servatrice.sql`, the migration and `DATABASE_SCHEMA_VERSION`.

The problems are on the resource and input-validation side rather than the authorization side, and two of them look blocking to me.

`cmdDeckUpload` binds client-controlled strings of up to 255 bytes into `varchar(5)` and `varchar(32)` columns. Under MySQL strict mode that INSERT fails, the unchecked `execSqlQuery` failure path calls `openDatabase()`, which `qDeleteAll`s the prepared statements, and the following `query->lastInsertId()` then dereferences a deleted `QSqlQuery` while the client is told `RespOk`. Separately, `cmdDeckShareCreate` has no per-user quota at all: `max_decks_per_share` bounds one share, not how many a user may create, and session commands are not covered by the flood counter, so any authenticated account can push an unbounded amount of deck content into the database and it persists for `expiry_days`.

Two smaller structural notes. There is no way for a user to list or revoke their own shares even though `created_by` is stored, which means a leaked token cannot be killed before it expires. And tags end up with three different encodings across this one feature (comma-separated string, JSON array, `repeated string`), which is worth settling now rather than after the client lands on top of it.

No server-side test coverage comes with this. The existing `tests/` tree only exercises leaf classes like `ServerRateLimiter` and `Server_Counter`, so I am not going to insist on wiring up database tests here, but the visibility inheritance in `isDeckEffectivelyPublic` and `deckListHelper` is pure enough logic that it would be worth extracting far enough to unit test.

Inline comments have the details.

servatrice/src/serversocketinterface.cpp
BruebachL: Thanks, will address.

BruebachL: Fixed. color_identity now goes through the shared WUBRG sanitizer extracted from makeShareItemFromDeck, and banner_card_name/banner_card_provider are bounded with left(255)/left(32). execSqlQuery is checked before touching the query again, so lastInsertId() can no longer run on a statement deleted by openDatabase()'s qDeleteAll(preparedStatements). 98e0d27e88

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. Added a per-user deck_share/max_shares_per_day setting (default 0 = unlimited) mirroring max_reports_per_day/max_comments_per_hour. Servatrice exposes getDeckShareMaxSharesPerDay() and cmdDeckShareCreate runs a 'select count(*) from ..._deck_share where created_by = :id and created_at >= DATE_SUB(NOW(), INTERVAL 1 DAY)' up front, returning RespTooManyRequests once the cap is met. 98e0d27e88

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. The folder path now drains all deck ids into a QList<int> before resolving each deck, so getDeckFromDatabase (which may call openDatabase and qDeleteAll(preparedStatements) via checkSql) no longer runs while the iterated statement is active. 98e0d27e88

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. cmd.items_size() > maxItems is rejected up front with RespInvalidData before any parsing, and the post-loop guard on the resolved folder decks returns RespInvalidData instead of RespTooManyRequests. 98e0d27e88

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. cmdDeckUpload now derives banner_card_name, banner_card_provider and tags server-side from the already-parsed deck via deck.getBannerCard()/deck.getTags(), and color_identity is sanitized with the WUBRG sanitizer extracted from makeShareItemFromDeck. The client-sent metadata fields were removed from Command_DeckUpload (reserved 5, 6, 8), so both paths agree on provenance. 98e0d27e88

servatrice/src/servatrice_database_interface.cpp
BruebachL: Fixed. The transaction start is now guarded with 'if (!sqlDatabase.transaction()) return false;' and the commit() result is checked. For the rollback-after-failed-execSqlQuery note: those paths rely on the connection teardown performed by execSqlQuery (close + reopen implicitly discards the transaction), which I documented with comments on each failure branch. 98e0d27e88

libcockatrice_protocol/libcockatrice/protocol/pb/command_deck_share_list.proto
BruebachL: Done. Added DECK_SHARE_LIST_MINE (SessionCommand 1032) and DECK_SHARE_REMOVE (1033) plus Response_DeckShareListMine (Response 1106) with a ServerInfo_DeckShareSummary payload. Server-side, 'list mine' queries created_by via getDeckSharesForUser and 'remove' deletes the caller's own bundle (cascades to items). No client UI yet, but the protocol surface is in place so a later client change needs no new bump. 98e0d27e88

servatrice/migrations/servatrice_0036_to_0037.sql
BruebachL: Fixed. token is now varchar(64) COLLATE utf8mb4_bin in both the migration and servatrice.sql, so token lookup and the UNIQUE KEY match the case-sensitive base64url tokens exactly. 98e0d27e88

servatrice/migrations/servatrice_0036_to_0037.sql
BruebachL: Fixed. Step 3 now adds banner_card_name AFTER is_public with the following columns chaining off it, so a migrated database matches the fresh-install order content, is_public, banner_card_name, banner_card_provider, color_identity, tags. 98e0d27e88

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. createDeckShare now reads back UNIX_TIMESTAMP(expires_at) right after the share insert and returns it through a qint64 &expiresAt out parameter; cmdDeckShareCreate reports that exact value to the client. 98e0d27e88

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. Both the deck and the folder UPDATE in cmdDeckSetVisibility now include 'and id_user = :id_user', so the ownership constraint is part of the mutating statement. 98e0d27e88

libcockatrice_protocol/libcockatrice/protocol/pb/serverinfo_deckstorage.proto
BruebachL: Fixed. ServerInfo_DeckStorage_File.tags is now repeated string, and cockatrice_decklist_files.tags stores a JSON array via a shared serializeDeckTags/deserializeDeckTags helper (servatrice/src/deck_tag_serialization.h) used by deckListHelper, cmdDeckUpload and createDeckShare - one encoding across the feature. 98e0d27e88

BruebachL:
Review-body round, all addressed in 98e0d27e88:

- Blocking: cmdDeckUpload no longer stores client-controlled banner/provider/color strings - it derives banner_card_name/banner_card_provider/tags from the parsed deck, sanitizes color_identity through the shared WUBRG sanitizer, bounds to column sizes, and checks execSqlQuery before touching the query again (reserved the removed Command_DeckUpload fields).
- Blocking: added deck_share/max_shares_per_day (0 = unlimited) with an up-front count query on created_by, mirroring max_reports_per_day/max_comments_per_hour.
- Structural: added DECK_SHARE_LIST_MINE / DECK_SHARE_REMOVE so share owners can enumerate and revoke their bundles (protocol numbers reserved in this PR; server-side implemented).
- Structural: tags are now a single encoding - repeated string on the wire everywhere, JSON array in both storage tables, shared serialize/deserialize helper.
- Testing note: visibility inheritance (isDeckEffectivelyPublic) is left as-is for now; agreed it would be cleaner extracted, worth a follow-up.

servatrice/src/serversocketinterface.cpp
```diff
@@ -566,46 +597,72 @@ int AbstractServerSocketInterface::getDeckPathId(const QString &path)
...
-        results[query->value(0).toInt()] = query->value(1).toString();
+        folderRows.append({query->value(0).toInt(), {query->value(1).toString(), query->value(2).toBool()}});
     }
+    std::sort(folderRows.begin(), folderRows.end(), [](const auto &a, const auto &b) { return a.first < b.first; });
+
+    for (const auto &[folderIdValue, folderInfo] : folderRows) {
+        const QString name = folderInfo.first;
+        const bool ownPublic = folderInfo.second;
+        const bool effectivePublic = inheritedPublic || ownPublic;
+        if (publicOnly && !effectivePublic) {
```
ZeldaZach: **Public decks nested under a private folder are never listed.**

When `publicOnly` is set, a folder that is not effectively public is `continue`d without recursing, so its whole subtree is dropped. But `isDeckEffectivelyPublic()` (used by `cmdDeckDownloadPublic`) only walks *upward* looking for a public ancestor — it returns true for a deck whose own `is_public` is 1 regardless of its parents.

Scenario: user has `Decks/Commander/MyDeck`, `Decks` and `Commander` both private, and marks `MyDeck` public via `Command_DeckSetVisibility(deck_id=...)`. `cmdDeckDownloadPublic` will serve that deck to anyone, but `cmdDeckListOtherUser` never lists it because it stops at the private `Decks` folder. Same for a public sub-folder under a private parent (`Decks` private, `Decks/Commander` public). The deck is effectively published but undiscoverable.

To match `isDeckEffectivelyPublic`, the recursion should always descend and only omit the folder node if the resulting subtree is empty.

servatrice/src/serversocketinterface.cpp
```diff
@@ -774,38 +1009,67 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckUpload(const Comman
...
         Response_DeckUpload *re = new Response_DeckUpload;
         ServerInfo_DeckStorage_TreeItem *fileInfo = re->mutable_new_file();
         fileInfo->set_id(query->lastInsertId().toInt());
         fileInfo->set_name(deckName.toStdString());
         fileInfo->mutable_file()->set_creation_time(QDateTime::currentDateTime().toSecsSinceEpoch());
+        fileInfo->mutable_file()->set_is_public(cmd.has_is_public() && cmd.is_public());
         rc.setResponseExtension(re);
     } else if (cmd.has_deck_id()) {
         QSqlQuery *query =
             sqlInterface->prepareQuery("update {prefix}_decklist_files set name=:name, upload_time=NOW(), "
```
ZeldaZach: **`Command_DeckUpload.is_public` is silently ignored when replacing an existing deck.**

The `has_path` (new deck) branch binds `:is_public` from `cmd.is_public()`, but this `has_deck_id` branch never touches the `is_public` column — it only re-reads the stored value afterwards.

Scenario: a client publishes a deck it already has in storage by re-uploading with `deck_id` set and `is_public = true` (the natural "publish this deck" flow, given the proto comment "mark the deck public on upload (publish)"). The server returns `RespOk` and the response echoes `is_public = false`, so the deck stays private with no error. The only way to publish an existing deck is a separate `Command_DeckSetVisibility`.

Either honour `is_public` here (`is_public = :is_public` when `cmd.has_is_public()`), or document in the proto that the field applies to new uploads only.

servatrice/src/serversocketinterface.cpp
```diff
@@ -836,6 +1100,245 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckDownload(const Comm
...
+            deckIds.append(query->value(0).toInt());
+        }
+        for (const int deckId : deckIds) {
+            DeckList *deck;
+            try {
+                deck = sqlInterface->getDeckFromDatabase(deckId, userInfo->id());
+            } catch (Response::ResponseCode &r) {
+                return r;
+            }
+            items.append(makeShareItemFromDeck(*deck, QString()));
```
ZeldaZach: **Folder shares lose the color identity of every deck.**

Decks resolved from `folder_path` are passed `QString()` as the color identity, so `sanitizeColorIdentity` yields an empty string and every item in a folder share has a blank `color_identity` in `Response_DeckShareList` — while item-by-item shares carry it.

This is now avoidable: `cmdDeckUpload` persists `color_identity` into `cockatrice_decklist_files`, so the folder query (`select id from ...`) could select `color_identity` alongside and pass it through.

servatrice/src/serversocketinterface.cpp
```diff
@@ -836,6 +1100,245 @@ Response::ResponseCode AbstractServerSocketInterface::cmdDeckDownload(const Comm
...
+    // Per-user rate limit so share links cannot be used to build an unbounded
+    // word-of-mouth leak of public decks.
+    const int maxSharesPerDay = servatrice->getDeckShareMaxSharesPerDay();
+    if (maxSharesPerDay > 0) {
+        QSqlQuery *countQuery = sqlInterface->prepareQuery("select count(*) from {prefix}_deck_share where "
+                                                           "created_by = :created_by and created_at >= "
+                                                           "DATE_SUB(NOW(), INTERVAL 1 DAY)");
+        countQuery->bindValue(":created_by", userInfo->id());
+        if (sqlInterface->execSqlQuery(countQuery) && countQuery->next() &&
+            countQuery->value(0).toInt() >= maxSharesPerDay) {
```
ZeldaZach: **The per-day share limit is silently bypassed when the count query fails.**

The limit is only enforced inside `if (execSqlQuery(countQuery) && countQuery->next() && value >= max)`. Any SQL failure (which also tears down and reopens the connection) makes the whole condition false, so the request is allowed through rather than rejected.

A client that can reliably make the count query fail — or that just races a connection drop — gets unlimited share creation on a server that configured `max_shares_per_day`. Prefer failing closed: return `RespContextError` (or `RespTooManyRequests`) when the count cannot be determined.

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -1079,6 +1081,182 @@ DeckList *Servatrice_DatabaseInterface::getDeckFromDatabase(int deckId, int user
...
+
+    QSqlQuery *query = prepareQuery("delete from {prefix}_deck_share where expires_at < now()");
+    execSqlQuery(query);
+}
+
+bool Servatrice_DatabaseInterface::getDeckSharesForUser(int userId, QList<DeckShareSummaryRecord> &shares)
+{
+    checkSql();
+
+    QSqlQuery *query = prepareQuery("select s.id, s.name, UNIX_TIMESTAMP(s.created_at), "
```
ZeldaZach, line 1225: **Expired shares are reported as live in "list my shares".**

Unlike `getDeckShareList` / `getDeckShareItem`, this query has no `expires_at > now()` predicate, so it returns every row the user ever created that the cleanup timer has not yet reaped.

With `deck_share/cleanup_interval = 0` (the documented way to disable automatic cleanup) expired rows are never deleted at all, so `Command_DeckShareListMine` returns a list that grows forever, mixing dead tokens with live ones. Adding `and s.expires_at > now()` would keep it consistent with the token lookups.

servatrice/src/servatrice.cpp
```diff
@@ -1067,6 +1080,28 @@ int Servatrice::getServerStatusUpdateTime() const
     return settingsCache->value("server/statusupdate", 15000).toInt();
 }
 
+int Servatrice::getDeckShareExpiryDays() const
+{
+    return settingsCache->value("deck_share/expiry_days", 7).toInt();
```
ZeldaZach: **`deck_share/expiry_days <= 0` creates a share that is dead on arrival.**

`getDeckShareExpiryDays()` has no lower bound, so a misconfigured (or 0/negative) value makes `DATE_ADD(NOW(), INTERVAL 0 DAY)` equal to `created_at`. `createDeckShare` then succeeds and returns a token to the client, but every subsequent `getDeckShareList` / `getDeckShareItem` filters on `expires_at > now()` and returns `RespNameNotFound` — the share link never works, and the next cleanup pass deletes it.

Clamping to at least 1 in `getDeckShareExpiryDays()` would turn a silent misconfiguration into sane behaviour.

servatrice/servatrice.ini.example
```diff
@@ -452,3 +452,24 @@ ssl_cert=ssl_cert.pem
...
+cleanup_interval=60
+
+; Maximum number of decks a single share link can contain.
+; Default: 50
+max_decks_per_share=50
+
+; Maximum number of share links a single user may create per day.
+; A value of 0 disables the limit.
+; Default: 0
+max_shares_per_day=0
```
ZeldaZach: The quota mechanism from the last round is here and correct, but it ships disabled, which leaves the original concern open on a default install.

The two settings this was modelled on both default to *on*: `reporting/max_reports_per_day` is 10 and `reporting/max_comments_per_hour` is 30. This one is 0. With the shipped defaults, `max_decks_per_share=50` and `fileFromStdString`'s 2 MB cap still let one authenticated account write ~100 MB of materialized deck content per `DECK_SHARE_CREATE`, repeated without limit, and the rows survive for `expiry_days`.

A non-zero default (20 or so is well clear of any real user) would make the protection actually apply to the servers that never touch this file. Not blocking the merge, but it is the difference between the quota existing and the quota working.

ZeldaZach (approved):
Re-reviewed against the previous round. All twelve inline findings are genuinely fixed, and I checked the code rather than taking the replies at face value:

- `cmdDeckUpload` now checks `execSqlQuery` before `lastInsertId()`, so the use-after-free on the reconnect path is gone, and the banner card / tags are derived server-side from the parsed deck with the client fields reserved (5, 6, 8). Both halves of that finding are closed.
- The share quota exists now (`deck_share/max_shares_per_day`), the folder path drains deck ids before resolving them, `items_size()` is bounded up front with `RespInvalidData`, the transaction start and commit are both guarded, `token` is `utf8mb4_bin`, the migration column order matches a fresh install, expiry is read back with `UNIX_TIMESTAMP` so it matches the database clock, both visibility UPDATEs carry `id_user`, and tags are `repeated string` with one JSON encoding behind `deck_tag_serialization.h`.
- `DECK_SHARE_LIST_MINE` / `DECK_SHARE_REMOVE` close the "a leaked token cannot be revoked" gap.

On the parts that would have made me hold this: I found nothing. Every new query is parameterized, `getDeckPathId` and `getDeckFromDatabase` stay scoped to the caller, `deleteDeckShare` filters on `created_by`, `getDeckShareItem` joins through the token so item ids cannot be walked across bundles, and `cmdDeckDownloadPublic` checks `isDeckEffectivelyPublic` server-side before serving. The token is 32 bytes from `QRandomGenerator::system()`. Extension and enum numbers are consistent and collide with nothing, every new proto is in the CMakeLists, and 37 agrees across `servatrice.sql`, the migration and `DATABASE_SCHEMA_VERSION`. The migration defaults everything to private.

Seven non-blocking notes left inline: the listing and download paths disagree about a public deck nested under a private folder, `is_public` is ignored when re-uploading over an existing `deck_id`, folder shares drop the color identity, the per-day quota fails open if its count query errors and also ships defaulted to 0, "list my shares" has no `expires_at` filter, and `expiry_days <= 0` mints a token that is already expired. None of them is a leak or a crash, so take them in a follow-up if you would rather land this now.

Approving.

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. deckListHelper now always descends into private folders and only drops a folder node when the public-only subtree is empty, so decks marked public under a private parent surface in the listing just like cmdDeckDownloadPublic serves them. 1395bca96b

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. The has_deck_id UPDATE now appends is_public=:is_public when Command_DeckUpload.has_is_public() is set (leaving the stored value untouched otherwise), and the response re-reads the column so it echoes the effective value. 1395bca96b

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. The folder query now selects color_identity alongside each deck id and hands it to makeShareItemFromDeck, so folder shares carry the persisted color identity like item-by-item shares do. 1395bca96b

servatrice/src/serversocketinterface.cpp
BruebachL: Fixed. When the count query fails or returns no row, cmdDeckShareCreate now returns RespContextError (fail closed); RespTooManyRequests is only returned when the counted value is actually at the cap. 1395bca96b

servatrice/src/servatrice_database_interface.cpp
BruebachL: Fixed. getDeckSharesForUser now filters s.expires_at > now(), matching getDeckShareList/getDeckShareItem, so Command_DeckShareListMine never reports expired shares as live. 1395bca96b

servatrice/src/servatrice.cpp
BruebachL: Fixed. getDeckShareExpiryDays clamps the configured value to at least 1 day, so a 0/negative expiry_days can no longer produce shares that are dead on arrival. 1395bca96b

servatrice/servatrice.ini.example
BruebachL: Fixed. deck_share/max_shares_per_day now defaults to 50 in both getDeckShareMaxSharesPerDay and servatrice.ini.example, mirroring the always-on defaults of max_reports_per_day and max_comments_per_hour; a configured 0 still disables the limit. 1395bca96b

PR #7242, opened 2026-09-05
[VDS] Decouple tag filter and fix reordered-chips crash

BruebachL:

The Visual Deck Storage tag filter is tangled into the model: tag checks run
inside `data()` across every column, and the live filter churns through the
whole tree, so reordering chips while filtering can leave a stale selected set
and crash on an invalid model index.

- Decouple the tag filter from the deck model into a standalone predicate the
  folder model applies during filtering
- Make the filter reactive to the deck database and settings instead of
  re-evaluating on every data access
- Guard the reordered-chips path against stale model indices so it can no longer
  crash when a selection is removed mid-filter
- Keep the visual deck storage layout and filtering behavior byte-for-byte
  identical for regular users

## Why?

The shared-deck and public-deck previews reuse this folder model and its tag
chips; the decoupling keeps the new views from inheriting the crash.

### DIFF
--- a/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
@@ -39,3 +39,8 @@ VisualDeckStorageSearchWidget::VisualDeckStorageSearchWidget(QWidget *parent) : 
 
     connect(searchDebounceTimer, &QTimer::timeout, this, [this] { emit searchTextChanged(searchBar->text()); });
 }
+
+void VisualDeckStorageSearchWidget::setPlaceholderText(const QString &text)
+{
+    searchBar->setPlaceholderText(text);
+}
--- a/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
@@ -2,14 +2,10 @@
 
 #include "../general/layout_containers/flow_widget.h"
 #include "deck_preview/deck_preview_tag_display_widget.h"
-#include "visual_deck_storage_model.h"
-#include "visual_deck_storage_sort_filter_proxy_model.h"
-#include "visual_deck_storage_widget.h"
 
 #include <QHBoxLayout>
 
-VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(VisualDeckStorageWidget *_parent)
-    : QWidget(_parent), parent(_parent)
+VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(QWidget *parent) : QWidget(parent)
 {
 
     setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
@@ -25,97 +21,62 @@ VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(VisualDeckSto
     layout->addWidget(flowWidget);
 }
 
+void VisualDeckStorageTagFilterWidget::setAllTagsProvider(const std::function<QSet<QString>()> &provider)
+{
+    allTagsProvider = provider;
+}
+
 void VisualDeckStorageTagFilterWidget::showEvent(QShowEvent *event)
 {
     QWidget::showEvent(event);
     refreshTags();
 }
 
-/**
- * @brief The tags of all decks currently accepted by the proxy model.
- */
-QSet<QString> VisualDeckStorageTagFilterWidget::gatherAllTags() const
-{
-    QSet<QString> allTags;
-    auto *proxy = parent->proxyModel();
-
-    for (int proxyRow = 0; proxyRow < proxy->rowCount(); ++proxyRow) {
-        const QModelIndex index = proxy->index(proxyRow, 0);
-        if (!index.data(VisualDeckStorageRoles::FilterMatchRole).toBool()) {
-            continue;
-        }
-        const QStringList deckTags = index.data(VisualDeckStorageRoles::TagsRole).toStringList();
-        for (const QString &tag : deckTags) {
-            allTags.insert(tag);
-        }
-    }
-
-    return allTags;
-}
-
 void VisualDeckStorageTagFilterWidget::refreshTags()
 {
-    QSet<QString> allTags = gatherAllTags();
-    removeTagsNotInList(allTags);
-    addTagsIfNotPresent(allTags);
-    sortTags();
-}
+    const QSet<QString> allTags = allTagsProvider ? allTagsProvider() : QSet<QString>();
 
-void VisualDeckStorageTagFilterWidget::removeTagsNotInList(const QSet<QString> &tags)
-{
+    // Existing chips survive if their tag is still part of the deck set, or if the chip
+    // is currently selected/excluded. Everything else is dropped. Dropped chips must NOT
+    // be re-added to the layout afterwards: they are scheduled for a deferred delete, and
+    // the flow layout would keep a dangling reference to them once the deletion runs on
+    // the next event-loop cycle.
+    QList<DeckPreviewTagDisplayWidget *> chips;
     for (DeckPreviewTagDisplayWidget *tagWidget : findChildren<DeckPreviewTagDisplayWidget *>()) {
-        const QString &tagName = tagWidget->getTagName();
-
-        // Keep the tag widget if it is either selected or excluded
-        if (!tags.contains(tagName) && tagWidget->getState() == TagState::NotSelected) {
+        if (tagWidget->getState() != TagState::NotSelected || allTags.contains(tagWidget->getTagName())) {
+            chips.append(tagWidget);
+        } else {
             flowWidget->removeWidget(tagWidget);
             tagWidget->deleteLater();
         }
     }
-}
 
-void VisualDeckStorageTagFilterWidget::addTagsIfNotPresent(const QSet<QString> &tags)
-{
-    for (const QString &tag : tags) {
-        addTagIfNotPresent(tag);
-    }
-}
-
-void VisualDeckStorageTagFilterWidget::addTagIfNotPresent(const QString &tag)
-{
-    // Check if the tag already exists in the flow widget
-    bool tagExists = false;
-    for (DeckPreviewTagDisplayWidget *tagWidget : findChildren<DeckPreviewTagDisplayWidget *>()) {
-        if (tagWidget->getTagName() == tag) {
-            tagExists = true;
-            break;
+    // Add chips for tags that are not shown yet.
+    for (const QString &tag : allTags) {
+        bool tagExists = false;
+        for (DeckPreviewTagDisplayWidget *tagWidget : chips) {
+            if (tagWidget->getTagName() == tag) {
+                tagExists = true;
+                break;
+            }
+        }
+        if (!tagExists) {
+            auto *newTagWidget = new DeckPreviewTagDisplayWidget(this, tag);
+            connect(newTagWidget, &DeckPreviewTagDisplayWidget::tagClicked, this,
+                    &VisualDeckStorageTagFilterWidget::filterChanged);
+            flowWidget->addWidget(newTagWidget);
+            chips.append(newTagWidget);
         }
     }
 
-    // If the tag doesn't exist, add a new DeckPreviewTagDisplayWidget
-    if (!tagExists) {
-        auto *newTagWidget = new DeckPreviewTagDisplayWidget(this, tag);
-        connect(newTagWidget, &DeckPreviewTagDisplayWidget::tagClicked, parent,
-                &VisualDeckStorageWidget::updateTagFilter);
-        flowWidget->addWidget(newTagWidget);
-    }
-}
-
-void VisualDeckStorageTagFilterWidget::sortTags()
-{
-    // Get all tag widgets
-    QList<DeckPreviewTagDisplayWidget *> tagWidgets = findChildren<DeckPreviewTagDisplayWidget *>();
-
-    // Sort widgets by tag name
-    std::sort(tagWidgets.begin(), tagWidgets.end(), [](DeckPreviewTagDisplayWidget *a, DeckPreviewTagDisplayWidget *b) {
+    // Clear and re-add the chips in sorted order.
+    std::sort(chips.begin(), chips.end(), [](DeckPreviewTagDisplayWidget *a, DeckPreviewTagDisplayWidget *b) {
         return a->getTagName().toLower() < b->getTagName().toLower();
     });
-
-    // Clear and re-add widgets in sorted order
-    for (DeckPreviewTagDisplayWidget *tagWidget : tagWidgets) {
+    for (DeckPreviewTagDisplayWidget *tagWidget : chips) {
         flowWidget->removeWidget(tagWidget);
     }
-    for (DeckPreviewTagDisplayWidget *tagWidget : tagWidgets) {
+    for (DeckPreviewTagDisplayWidget *tagWidget : chips) {
         flowWidget->addWidget(tagWidget);
     }
 }
--- a/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
+++ b/cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
@@ -9,25 +9,26 @@
 #include <QSet>
 #include <QStringList>
 #include <QWidget>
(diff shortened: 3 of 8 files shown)


cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
```diff
@@ -25,97 +21,62 @@ VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(VisualDeckSto
...
     for (DeckPreviewTagDisplayWidget *tagWidget : findChildren<DeckPreviewTagDisplayWidget *>()) {
-        const QString &tagName = tagWidget->getTagName();
-
-        // Keep the tag widget if it is either selected or excluded
-        if (!tags.contains(tagName) && tagWidget->getState() == TagState::NotSelected) {
+        if (tagWidget->getState() != TagState::NotSelected || allTags.contains(tagWidget->getTagName())) {
+            chips.append(tagWidget);
+        } else {
             flowWidget->removeWidget(tagWidget);
             tagWidget->deleteLater();
```
ZeldaZach, line 50: The fix is correct within a single call, but it leans on an invariant the code does not actually enforce. The doomed chip stays parented under `this` until the deferred delete runs, so `findChildren` on line 45 still returns it. If `refreshTags` runs a second time in the same event-loop turn and the tag has reappeared in `allTags`, the pending-delete chip passes the retention check, lands back in `chips`, and gets re-added to the layout, which is exactly the situation the comment above says must never happen.

It also leaves a stale chip painted at its old geometry for a frame and keeps polluting `getAllKnownTags()` until the delete lands.

Detaching makes the invariant structural instead of a comment:

```cpp
flowWidget->removeWidget(tagWidget);
tagWidget->setParent(nullptr);
tagWidget->deleteLater();
```

`setParent(nullptr)` also hides it, and an immediate `delete` is not safe here since this can be reached from the chip's own `mousePressEvent`.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
```diff
@@ -39,3 +39,8 @@ VisualDeckStorageSearchWidget::VisualDeckStorageSearchWidget(QWidget *parent) :
 
     connect(searchDebounceTimer, &QTimer::timeout, this, [this] { emit searchTextChanged(searchBar->text()); });
 }
+
+void VisualDeckStorageSearchWidget::setPlaceholderText(const QString &text)
```
ZeldaZach, line 43: Nothing calls this. I checked the rest of the stack through #7245 and it is still uncalled there. Drop it here and add it in whichever PR actually needs a custom placeholder.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
```diff
@@ -39,9 +40,15 @@ class VisualDeckStorageTagFilterWidget : public QWidget
      */
     [[nodiscard]] QStringList excludedTags() const;
 
+signals:
+    /**
+     * Emitted whenever a chip's selection/exclusion state changes.
+     */
+    void filterChanged();
```
ZeldaZach, line 47: The doc says this fires whenever a chip's selection or exclusion state changes, but it is wired straight to `tagClicked`, and `DeckPreviewTagDisplayWidget::mousePressEvent` emits that unconditionally after the switch, including for buttons that fall through `default: break` without touching the state.

Either reword to "emitted when a chip is clicked", or have the chip emit only when `state` actually changed. The second is the better fix, since each spurious emission drives a full `updateTagFilter` plus a chip relayout.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
```diff
@@ -25,97 +21,62 @@ VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(VisualDeckSto
...
-
-void VisualDeckStorageTagFilterWidget::sortTags()
-{
-    // Get all tag widgets
-    QList<DeckPreviewTagDisplayWidget *> tagWidgets = findChildren<DeckPreviewTagDisplayWidget *>();
-
-    // Sort widgets by tag name
-    std::sort(tagWidgets.begin(), tagWidgets.end(), [](DeckPreviewTagDisplayWidget *a, DeckPreviewTagDisplayWidget *b) {
+    // Clear and re-add the chips in sorted order.
+    std::sort(chips.begin(), chips.end(), [](DeckPreviewTagDisplayWidget *a, DeckPreviewTagDisplayWidget *b) {
```
ZeldaZach, line 73: Every `refreshTags` removes and re-adds all chips unconditionally, and `FlowLayout` has no `removeWidget` override, so each call goes through `QLayout::removeWidget`'s linear scan. That makes the reorder quadratic plus a full relayout even when the order is already correct.

This is carried over from the old `sortTags` so it is not a regression, but `refreshTags` now runs on every chip click and every debounced load batch, so skipping the rebuild when the sorted order already matches the layout order would be cheap to add.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
```diff
@@ -9,26 +9,27 @@
 #include <QSet>
 #include <QStringList>
 #include <QWidget>
+#include <functional>
 
 class FlowWidget;
 class VisualDeckStorageWidget;
```
ZeldaZach, line 15: This forward declaration is dead now that the `parent` member and the `VisualDeckStorageWidget`-typed constructor are gone. Worth removing so the header stops advertising the dependency this PR just removed.

ZeldaZach (commented):
Went through this one specifically to check the "behavior is identical" claim, since that is the whole basis for the refactor, and it holds up.

The chip retention predicate is exactly equivalent to what it replaced: the old `!tags.contains(name) && state == NotSelected` removal condition is the De Morgan complement of the new `state != NotSelected || allTags.contains(name)` retention condition, and the sort comparator is unchanged, so chip ordering and the selected/excluded stickiness both survive. `gatherAllTags` moving into `VisualDeckStorageWidget::gatherVisibleTags` is verbatim apart from swapping `parent->proxyModel()` for the member. The new `filterChanged` connection uses `this` as the context object, so it dies with the widget rather than leaking. Dropping the typed back-pointer in favour of an injected provider is the right call and is clearly what lets #7245 reuse this widget from the public decks tab. The `refreshTags` rewrite is also a modest performance win, since it calls `findChildren` once instead of once per tag.

One thing keeps me from just approving. The crash fix depends on an invariant that nothing enforces: the removed chip is still a child of the widget until `deleteLater` fires, so a second `refreshTags` in the same event-loop turn can pick it back up through `findChildren` and re-add it to the layout, which is precisely the case the new comment warns against. Reparenting to `nullptr` alongside the `deleteLater` turns that comment into something the code actually guarantees. The rest are small: a dead forward declaration, an unused `setPlaceholderText`, and a signal whose doc comment promises more than the wiring delivers.

Worth noting there is no test coverage for the tag filter anywhere in `tests/`, and this predicate is now a standalone, injectable thing, which makes it a lot more testable than it was. Not asking for that in this PR, just flagging that the refactor opened the door.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
BruebachL: Addressed in 1b05ffc0. The doomed chip is now reparented to nullptr immediately after removeWidget (tagWidget->setParent(nullptr)), so findChildren can no longer return it before the deferred delete runs — the invariant the comment warned about is now structurally guaranteed. setParent(nullptr) also hides the chip, so it is not painted at its old geometry for a frame.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_search_widget.cpp
BruebachL: Dropped in 1b05ffc0. setPlaceholderText was unused anywhere (including in #7245), so I removed it from both the header and the cpp; if a future PR needs a custom placeholder it can add it there.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
BruebachL: Fixed in 1b05ffc0 using the better option. DeckPreviewTagDisplayWidget::mousePressEvent now captures the previous state and only emits tagClicked when state actually toggled, so clicks that fall through the switch (e.g. a button the chip does not react to) no longer drive a full filter update plus chip relayout. The filterChanged doc comment was updated to match.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
BruebachL: Done in 1b05ffc0. refreshTags now tracks the last applied chip order (currentChipOrder) and skips the remove/re-add rebuild entirely when the newly sorted list matches the layout order, avoiding the quadratic QLayout::removeWidget scan plus relayout on every chip click and load batch. Since chips dropped from the set are reparented to nullptr before the comparison, any real change to the member set or ordering still triggers a rebuild.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
BruebachL: Removed in 1b05ffc0. The dead VisualDeckStorageWidget forward declaration is gone; the header now only declares DeckPreviewTagDisplayWidget and FlowWidget.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
```diff
@@ -25,99 +21,72 @@ VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(VisualDeckSto
...
-    // Check if the tag already exists in the flow widget
-    bool tagExists = false;
-    for (DeckPreviewTagDisplayWidget *tagWidget : findChildren<DeckPreviewTagDisplayWidget *>()) {
-        if (tagWidget->getTagName() == tag) {
-            tagExists = true;
-            break;
+    // Add chips for tags that are not shown yet.
+    for (const QString &tag : allTags) {
+        bool tagExists = false;
+        for (DeckPreviewTagDisplayWidget *tagWidget : chips) {
```
ZeldaZach: The "does this tag already have a chip" test is a linear scan of `chips` nested inside the loop over `allTags`, so refreshTags is O(tags x chips) — and it runs on every chip click plus every debounced load batch. That is inconsistent with the comment a few lines below, which justifies the sort fast path on exactly these grounds. Collecting the existing chip names into a QSet<QString> once before the loop makes the lookup O(1).

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
```diff
@@ -25,99 +21,72 @@ VisualDeckStorageTagFilterWidget::VisualDeckStorageTagFilterWidget(VisualDeckSto
...
-{
-    // Get all tag widgets
-    QList<DeckPreviewTagDisplayWidget *> tagWidgets = findChildren<DeckPreviewTagDisplayWidget *>();
-
-    // Sort widgets by tag name
-    std::sort(tagWidgets.begin(), tagWidgets.end(), [](DeckPreviewTagDisplayWidget *a, DeckPreviewTagDisplayWidget *b) {
+    // Sort, but skip the full remove/re-add when the order already matches the layout.
+    // FlowWidget inherits QLayout::removeWidget's linear scan, so rebuilding an unchanged
+    // order would be quadratic plus a full relayout on every chip click and load batch.
+    std::sort(chips.begin(), chips.end(), [](DeckPreviewTagDisplayWidget *a, DeckPreviewTagDisplayWidget *b) {
```
ZeldaZach, line 74: The comparator only looks at `toLower()`, so two tags that differ only in case (say "Aggro" and "aggro", both legal as separate deck tags) compare equal. std::sort is not stable, so their relative order is whatever the introsort partitioning produces for the current input permutation. Since the input permutation changes whenever a chip is created or destroyed, those two chips can swap places, which makes `chips != currentChipOrder` and forces the full remove/re-add relayout the fast path just below is there to avoid. A tiebreak on the exact name would make the order well defined.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
```diff
@@ -39,9 +41,18 @@ class VisualDeckStorageTagFilterWidget : public QWidget
      */
     [[nodiscard]] QStringList excludedTags() const;
 
+signals:
+    /**
+     * Emitted when a chip's selection or exclusion state changes.
+     *
+     * The chip only emits when its state actually changed, so this fires once per
+     * effective toggle rather than on every click."
```
ZeldaZach: Stray double quote at the end of the sentence.

```suggestion
     * effective toggle rather than on every click.
```

ZeldaZach (approved):
All five points from the previous round are addressed, and I re-checked the crash fix rather than taking the commit message for it.

The root cause was real and is now genuinely closed. `deleteLater()` does not reparent, so the old `sortTags()` re-read `findChildren()` and got back chips that `removeTagsNotInList()` had already scheduled for deletion — it then removed and re-added those to the flow layout, leaving the layout holding items whose widgets died on the next event-loop turn. The rewrite closes both halves: surviving chips are carried in a local list instead of being re-queried, and `setParent(nullptr)` before `deleteLater()` makes the "a doomed chip must never come back" rule structural instead of a comment. `setParent(nullptr)` also hides the widget, so nothing is left painted at stale geometry.

I tried to break the `currentChipOrder` fast path and could not. Any pass that creates or destroys a chip necessarily makes `chips != currentChipOrder`, so the early return can only fire when the tracked order still matches the live chip set, and the invariant holds inductively from the empty initial state. Address reuse after a destroyed chip cannot produce a false match either, since a deleting pass always falls through to the rebuild and reassigns. Deleting the clicked chip from inside its own `mousePressEvent` (reachable when you deselect a tag that then leaves the visible set) is fine too — the deferred delete means the trailing `QWidget::mousePressEvent` still runs on a live, merely hidden widget.

Nothing was silently dropped in the decoupling. The retention predicate is still the De Morgan complement of the old removal condition, the comparator is unchanged, and `gatherVisibleTags` is the old `gatherAllTags` verbatim. Gating `tagClicked` on an actual state change has no other consumer — the only other place that constructs a `DeckPreviewTagDisplayWidget` never connects that signal. Provider wiring order is correct: the proxy model exists before the lambda can run, and the `filterChanged` connection is made before anything can trigger `showEvent`.

Three non-blocking nits left inline: the new-chip lookup is a nested linear scan where a `QSet<QString>` would do, the sort comparator ties on `toLower()` so same-name-different-case tags can swap under the unstable sort and defeat the fast path it sits next to, and there is a stray double quote in the `filterChanged` doc comment. None of them need to hold up the merge.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
BruebachL: Fixed in ce1a93b9. refreshTags now collects the surviving chip names into a QSet<QString> once before the loop, so membership is O(1) instead of the nested O(tags x chips) scan.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.cpp
BruebachL: Fixed in ce1a93b9. The comparator now case-insensitively compares first via QString::compare and ties break on the exact (case-sensitive) name, so tags differing only in case have a well-defined order and can no longer swap under the unstable sort to defeat the fast path.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_tag_filter_widget.h
BruebachL: Fixed in ce1a93b9 — the stray trailing quote is gone from the filterChanged doc comment.

PR #7243, opened 2026-09-05
[DeckShare] Create temporary share links for local and server decks

BruebachL:

Decks on the server can only be shared as a permanent folder listing, and local
decks have no share story at all. Creating a temporary share link for one deck
meant hand-editing URLs or making a folder public.

- Add a `Share deck` action to the deck editor toolbar, the server storage tree
  and its context menu, and the touch-optimized share screen
- Add a share-link dialog that creates a temporary link for the selected file or
  folder with an expiry, mirrors it for local decks via deck upload + share
  create, and copies the resulting URL to the clipboard
- Add the share links page in the server deck storage tree listing active links
  with a `Copy link` action
- Add the shared-deck preview dialog where a received link opens the deck with a
  gate (`Open deck` vs. hovering for share value) and the full link text for
  holders of the token
- Recognize `cockatrice://` and server deck-storage links pasted anywhere in the
  client, matching them against the host before opening the share dialog, and
  follow the browser protocol handler with a manual fallback instructions
  dialog

## Why?

The next PR intercepts these share links at startup and while logged in, routing
them through the roaming-url parser into the preview flow.

### DIFF
--- a/cockatrice/src/interface/widgets/cards/additional_info/deck_color_identity.cpp
+++ b/cockatrice/src/interface/widgets/cards/additional_info/deck_color_identity.cpp
@@ -1,11 +1,11 @@
 #include "deck_color_identity.h"
 
 #include <QSet>
-#include <libcockatrice/card/database/card_database_manager.h>
+#include <libcockatrice/card/database/card_database_querier.h>
 #include <libcockatrice/deck_list/deck_list.h>
 #include <libcockatrice/deck_list/tree/inner_deck_list_node.h>
 
-QString getDeckColorIdentity(const DeckList &deck)
+QString getDeckColorIdentity(const DeckList &deck, const CardDatabaseQuerier *db)
 {
     const QStringList cardList = deck.getCardList({DECK_ZONE_MAIN, DECK_ZONE_SIDE});
     if (cardList.isEmpty()) {
@@ -15,7 +15,7 @@ QString getDeckColorIdentity(const DeckList &deck)
     QSet<QChar> colorSet; // A set to collect unique color symbols (e.g., W, U, B, R, G)
 
     for (const QString &cardName : cardList) {
-        CardInfoPtr currentCard = CardDatabaseManager::query()->getCardInfo(cardName);
+        CardInfoPtr currentCard = db->getCardInfo(cardName);
         if (currentCard) {
             const QString colors = currentCard->getColors(); // returns something like "WUB"
             for (const QChar &color : colors) {
--- a/cockatrice/src/interface/widgets/cards/additional_info/deck_color_identity.cpp
+++ b/cockatrice/src/interface/widgets/cards/additional_info/deck_color_identity.cpp
@@ -1,0 +1,37 @@
+#include "deck_color_identity.h"
+
+#include <QSet>
+#include <libcockatrice/card/database/card_database_manager.h>
+#include <libcockatrice/deck_list/deck_list.h>
+#include <libcockatrice/deck_list/tree/inner_deck_list_node.h>
+
+QString getDeckColorIdentity(const DeckList &deck)
+{
+    const QStringList cardList = deck.getCardList({DECK_ZONE_MAIN, DECK_ZONE_SIDE});
+    if (cardList.isEmpty()) {
+        return {};
+    }
+
+    QSet<QChar> colorSet; // A set to collect unique color symbols (e.g., W, U, B, R, G)
+
+    for (const QString &cardName : cardList) {
+        CardInfoPtr currentCard = CardDatabaseManager::query()->getCardInfo(cardName);
+        if (currentCard) {
+            const QString colors = currentCard->getColors(); // returns something like "WUB"
+            for (const QChar &color : colors) {
+                colorSet.insert(color);
+            }
+        }
+    }
+
+    // Ensure the color identity is in WUBRG order
+    QString colorIdentity;
+    const QString wubrgOrder = "WUBRG";
+    for (const QChar &color : wubrgOrder) {
+        if (colorSet.contains(color)) {
+            colorIdentity.append(color);
+        }
+    }
+
+    return colorIdentity;
+}
--- a/cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
+++ b/cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
@@ -1,0 +1,28 @@
+#include "deck_share_utils.h"
+
+#include <QClipboard>
+#include <QGuiApplication>
+#include <QTimeZone>
+#include <libcockatrice/network/client/abstract/abstract_client.h>
+
+namespace DeckShareUtils
+{
+
+QString buildShareLink(const AbstractClient *client, const QString &token)
+{
+    return QString("cockatrice://opendeck?share=%1&hostname=%2&port=%3")
+        .arg(token, client->serverName(), QString::number(client->serverPort()));
+}
+
+QString copyShareLinkToClipboard(const QString &link)
+{
+    QGuiApplication::clipboard()->setText(link);
+    return link;
+}
+
+QString formatShareExpiry(const QDateTime &expiry)
+{
+    return expiry.toLocalTime().toString();
+}
+
+} // namespace DeckShareUtils
--- a/cockatrice/src/interface/widgets/deck_share/deck_share_utils.h
+++ b/cockatrice/src/interface/widgets/deck_share/deck_share_utils.h
@@ -1,0 +1,41 @@
+/**
+ * @file deck_share_utils.h
+ * @ingroup DeckShareWidgets
+ */
+//! \todo Document this file.
+
+#ifndef DECK_SHARE_UTILS_H
+#define DECK_SHARE_UTILS_H
+
+#include <QDateTime>
+#include <QString>
+
+class AbstractClient;
+
+/**
+ * @brief Shared helpers for creating temporary deck shares.
+ */
+namespace DeckShareUtils
+{
+
+/**
+ * @brief Builds the cockatrice:// link for a freshly created deck share.
+ * @param client Used to embed the target server's hostname and port.
+ * @param token The share token from Response_DeckShareCreate.
+ */
+QString buildShareLink(const AbstractClient *client, const QString &token);
+
+/**
+ * @brief Copies the share link to the clipboard.
+ * @return The link that was copied.
+ */
+QString copyShareLinkToClipboard(const QString &link);
+
+/**
+ * @brief Formats the expiration timestamp for a share.
+ */
+QString formatShareExpiry(const QDateTime &expiry);
+
+} // namespace DeckShareUtils
+
+#endif // DECK_SHARE_UTILS_H
--- a/cockatrice/src/interface/widgets/deck_share/share_bar_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_share/share_bar_widget.cpp
@@ -67,6 +67,11 @@ void ShareBarWidget::setHintText(const QString &text, bool visible)
     hintLabel->setVisible(visible);
 }
 
+void ShareBarWidget::setCreateEnabled(bool enabled)
+{
+    createButton->setEnabled(enabled);
+}
+
 void ShareBarWidget::focusName()
 {
     nameEdit->setFocus();
--- a/cockatrice/src/interface/widgets/deck_share/share_bar_widget.cpp
+++ b/cockatrice/src/interface/widgets/deck_share/share_bar_widget.cpp
@@ -1,0 +1,73 @@
+#include "share_bar_widget.h"
+
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QLineEdit>
+#include <QPushButton>
+
+ShareBarWidget::ShareBarWidget(QWidget *parent) : QWidget(parent)
+{
+    auto *layout = new QHBoxLayout(this);
+    layout->setContentsMargins(12, 10, 12, 10);
+    layout->setSpacing(8);
+
+    hintLabel = new QLabel(this);
+    hintLabel->setWordWrap(true);
+
+    nameEdit = new QLineEdit(this);
+    nameEdit->setMaximumWidth(260);
+
+    countLabel = new QLabel(this);
+
+    cancelButton = new QPushButton(this);
+    connect(cancelButton, &QPushButton::clicked, this, &ShareBarWidget::cancelRequested);
+
+    createButton = new QPushButton(this);
+    createButton->setDefault(true);
+    connect(createButton, &QPushButton::clicked, this, &ShareBarWidget::createRequested);
+
+    layout->addWidget(hintLabel, 1);
+    layout->addWidget(nameEdit);
+    layout->addWidget(countLabel);
+    layout->addStretch();
+    layout->addWidget(cancelButton);
+    layout->addWidget(createButton);
+
+    setLayout(layout);
+
+    retranslateUi();
+}
+
+void ShareBarWidget::retranslateUi()
+{
+    nameEdit->setPlaceholderText(tr("Share name"));
+    cancelButton->setText(tr("Cancel"));
+    createButton->setText(tr("Create share link"));
+    hintLabel->setText(tr("Click deck tiles to select the decks you want to share."));
+}
+
+QString ShareBarWidget::name() const
+{
+    return nameEdit->text().trimmed();
+}
+
+void ShareBarWidget::setName(const QString &value)
+{
+    nameEdit->setText(value);
+}
+
+void ShareBarWidget::setCountText(const QString &text)
+{
+    countLabel->setText(text);
+}
+
+void ShareBarWidget::setHintText(const QString &text, bool visible)
+{
+    hintLabel->setText(text);
+    hintLabel->setVisible(visible);
+}
+
+void ShareBarWidget::focusName()
+{
+    nameEdit->setFocus();
+}
--- a/cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
@@ -1,0 +1,80 @@
+#include "dlg_share_deck.h"
+
+#include "../cards/additional_info/deck_color_identity.h"
+#include "../deck_share/deck_share_utils.h"
+
+#include <QDialogButtonBox>
+#include <QFormLayout>
+#include <QLineEdit>
+#include <QMessageBox>
+#include <QPushButton>
+#include <QTimeZone>
+#include <QVBoxLayout>
+#include <libcockatrice/deck_list/deck_list.h>
+#include <libcockatrice/network/client/abstract/abstract_client.h>
+#include <libcockatrice/protocol/pb/command_deck_share_create.pb.h>
+#include <libcockatrice/protocol/pb/response.pb.h>
+#include <libcockatrice/protocol/pb/response_deck_share_create.pb.h>
+#include <libcockatrice/protocol/pending_command.h>
+
+DlgShareDeck::DlgShareDeck(AbstractClient *_client, const QSharedPointer<DeckList> &_deck, QWidget *_parent)
+    : QDialog(_parent), client(_client), deck(_deck)
+{
+    setWindowTitle(tr("Share deck"));
+
+    auto *layout = new QVBoxLayout(this);
+
+    nameEdit = new QLineEdit(this);
+    nameEdit->setText(tr("Shared deck"));
+
+    auto *form = new QFormLayout;
+    form->addRow(tr("Share name:"), nameEdit);
+    layout->addLayout(form);
+
+    auto *buttonBox = new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel, this);
+    buttonBox->button(QDialogButtonBox::Ok)->setText(tr("Create share link"));
+    buttonBox->button(QDialogButtonBox::Cancel)->setText(tr("Cancel"));
+    connect(buttonBox, &QDialogButtonBox::accepted, this, &DlgShareDeck::actShare);
+    connect(buttonBox, &QDialogButtonBox::rejected, this, &DlgShareDeck::reject);
+    layout->addWidget(buttonBox);
+}
+
+void DlgShareDeck::actShare()
+{
+    Command_DeckShareCreate cmd;
+    cmd.set_name(nameEdit->text().trimmed().toStdString());
+    if (cmd.name().empty()) {
+        cmd.set_name(tr("Shared deck").toStdString());
+    }
+
+    DeckShareItem *item = cmd.add_items();
+    item->set_deck_list(deck->writeToString_Native().toStdString());
+    item->set_color_identity(getDeckColorIdentity(*deck).toStdString());
+
+    PendingCommand *pend = client->prepareSessionCommand(cmd);
+    connect(pend, &PendingCommand::finished, this, &DlgShareDeck::shareFinished);
+    client->sendCommand(pend);
+}
+
+void DlgShareDeck::shareFinished(const Response &response, const CommandContainer & /*commandContainer*/)
+{
+    if (response.response_code() != Response::RespOk) {
+        QMessageBox::critical(this, tr("Share deck"),
+                              tr("Failed to create the share link (server response code %1).")
+                                  .arg(QString::number(static_cast<int>(response.response_code()))));
+        return;
+    }
+
+    const Response_DeckShareCreate &resp = response.GetExtension(Response_DeckShareCreate::ext);
+    const QString token = QString::fromStdString(resp.token());
+    const QDateTime expiry = QDateTime::fromSecsSinceEpoch(resp.expires_at(), QTimeZone::UTC);
+
+    const QString link = DeckShareUtils::buildShareLink(client, token);
+    DeckShareUtils::copyShareLinkToClipboard(link);
+
+    QMessageBox::information(this, tr("Share deck"),
+                             tr("Share link created and copied to the clipboard:\n\n%1\n\n"
+                                "The share expires on %2.")
+                                 .arg(link, DeckShareUtils::formatShareExpiry(expiry)));
+    accept();
+}
--- a/cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
@@ -72,7 +72,11 @@ void DlgShareDeck::shareFinished(const Response &response, const CommandContaine
 
     const Response_DeckShareCreate &resp = response.GetExtension(Response_DeckShareCreate::ext);
     const QString token = QString::fromStdString(resp.token());
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
     const QDateTime expiry = QDateTime::fromSecsSinceEpoch(resp.expires_at(), QTimeZone::UTC);
+#else
+    const QDateTime expiry = QDateTime::fromSecsSinceEpoch(resp.expires_at(), Qt::UTC);
+#endif
 
     const QString link = DeckShareUtils::buildShareLink(client, token);
     DeckShareUtils::copyShareLinkToClipboard(link);
--- a/cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
@@ -10,6 +10,7 @@
 #include <QPushButton>
 #include <QTimeZone>
 #include <QVBoxLayout>
+#include <libcockatrice/card/database/card_database_manager.h>
 #include <libcockatrice/deck_list/deck_list.h>
 #include <libcockatrice/network/client/abstract/abstract_client.h>
 #include <libcockatrice/protocol/pb/command_deck_share_create.pb.h>
@@ -36,11 +37,14 @@ DlgShareDeck::DlgShareDeck(AbstractClient *_client, const QSharedPointer<DeckLis
     buttonBox->button(QDialogButtonBox::Cancel)->setText(tr("Cancel"));
     connect(buttonBox, &QDialogButtonBox::accepted, this, &DlgShareDeck::actShare);
     connect(buttonBox, &QDialogButtonBox::rejected, this, &DlgShareDeck::reject);
+    this->buttonBox = buttonBox;
     layout->addWidget(buttonBox);
 }
 
 void DlgShareDeck::actShare()
 {
+    buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
+
     Command_DeckShareCreate cmd;
     cmd.set_name(nameEdit->text().trimmed().toStdString());
     if (cmd.name().empty()) {
@@ -49,7 +53,7 @@ void DlgShareDeck::actShare()
 
     DeckShareItem *item = cmd.add_items();
     item->set_deck_list(deck->writeToString_Native().toStdString());
-    item->set_color_identity(getDeckColorIdentity(*deck).toStdString());
+    item->set_color_identity(getDeckColorIdentity(*deck, CardDatabaseManager::query()).toStdString());
 
     PendingCommand *pend = client->prepareSessionCommand(cmd);
     connect(pend, &PendingCommand::finished, this, &DlgShareDeck::shareFinished);
@@ -59,6 +63,7 @@ void DlgShareDeck::actShare()
 void DlgShareDeck::shareFinished(const Response &response, const CommandContainer & /*commandContainer*/)
 {
     if (response.response_code() != Response::RespOk) {
+        buttonBox->button(QDialogButtonBox::Ok)->setEnabled(true);
         QMessageBox::critical(this, tr("Share deck"),
                               tr("Failed to create the share link (server response code %1).")
                                   .arg(QString::number(static_cast<int>(response.response_code()))));
--- a/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
@@ -388,6 +388,11 @@ bool AbstractTabDeckEditor::actSaveDeckAs()
  */
 void AbstractTabDeckEditor::actShareDeck()
 {
+    if (tabSupervisor->getClient()->getStatus() != StatusLoggedIn) {
+        QMessageBox::information(this, tr("Share deck"), tr("You must be connected to the server to share a deck."));
+        return;
+    }
+
     const QSharedPointer<DeckList> deck = deckStateManager->getDeckListShared();
     if (deck->isBlankDeck()) {
         return;
--- a/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
@@ -19,6 +19,7 @@
 #include "../interface/widgets/dialogs/dlg_load_deck.h"
 #include "../interface/widgets/dialogs/dlg_load_deck_from_clipboard.h"
 #include "../interface/widgets/dialogs/dlg_load_deck_from_website.h"
+#include "../interface/widgets/dialogs/dlg_share_deck.h"
 #include "../utility/visibility_change_listener.h"
 #include "tab_supervisor.h"
 
@@ -380,6 +381,20 @@ bool AbstractTabDeckEditor::actSaveDeckAs()
     deckStateManager->setModified(false);
     SettingsCache::instance().recents().updateRecentlyOpenedDeckPaths(fileName);
     return true;
+}
+
+/**
+ * @brief Opens the deck share dialog with the current deck preselected.
+ */
+void AbstractTabDeckEditor::actShareDeck()
+{
+    const QSharedPointer<DeckList> deck = deckStateManager->getDeckListShared();
+    if (deck->isBlankDeck()) {
+        return;
+    }
+
+    DlgShareDeck shareDialog(tabSupervisor->getClient(), deck, this);
+    shareDialog.exec();
 }
 
 /**
--- a/cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
@@ -1,19 +1,30 @@
 #include "tab_deck_storage.h"
 
 #include "../../../client/settings/cache_settings.h"
+#include "../../../main.h"
 #include "../../deck_loader/deck_loader.h"
+#include "../deck_share/share_bar_widget.h"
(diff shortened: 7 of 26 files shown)


cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
```diff
@@ -14,12 +25,42 @@ TabDeckStorageVisual::TabDeckStorageVisual(TabSupervisor *_tabSupervisor)
...
+            &TabDeckStorageVisual::actShareDeck);
+    connect(visualDeckStorageWidget, &VisualDeckStorageWidget::shareSelectionChanged, this,
+            &TabDeckStorageVisual::onShareSelectionChanged);
+    connect(visualDeckStorageWidget, &VisualDeckStorageWidget::shareRequested, this, [this] {
+        if (shareDeckAvailable) {
+            enterShareMode();
+        }
+    });
+
+    AbstractClient *client = tabSupervisor->getClient();
```
ZeldaZach, line 38: `TabSupervisor::getClient()` returns `localClients.first()` whenever a local game is running, so opening this tab during an offline game caches a `LocalClient` here. `LocalClient` never calls `setStatus`, so it never emits `statusChanged`, and `handleConnectionChanged` will not fire again for the lifetime of the tab. The share button then stays hidden even after the user later logs into a real server. The local client is also `deleteLater`'d in `TabSupervisor::stop()`, so the connection dies silently.

Connect to a stable client instead, or re-resolve `tabSupervisor->getClient()` on every use and also connect to TabSupervisor's own connection lifecycle signals, rather than caching one pointer at construction.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
```diff
@@ -33,3 +74,127 @@ void TabDeckStorageVisual::actOpenLocalDeck(const QString &filePath)
...
+            QMessageBox::warning(this, tr("Share decks"), tr("Unable to load deck file %1").arg(filePath));
+            return;
+        }
+        DeckShareItem *item = cmd.add_items();
+        item->set_deck_list(deckOpt->deckList.writeToString_Native().toStdString());
+        item->set_color_identity(getDeckColorIdentity(deckOpt->deckList, CardDatabaseManager::query()).toStdString());
+    }
+
+    shareBar->setCreateEnabled(false);
+    PendingCommand *pend = tabSupervisor->getClient()->prepareSessionCommand(cmd);
```
ZeldaZach, line 155: The gating check uses the client cached in the constructor, but the send resolves `tabSupervisor->getClient()` fresh, and those can be different objects.

Concretely: user is logged in, so `shareDeckAvailable` is true, then starts an offline game. `Command_DeckShareCreate` now goes to the `LocalClient`, and `buildShareLink` on line 178 produces `hostname=&port=0`, since `AbstractClient::serverName()` and `serverPort()` default to empty and 0.

Resolve the client once at the top of this method and use that same pointer for the gate, the send and the link construction.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.h
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @brief Copies the share link to the clipboard.
+ * @return The link that was copied.
+ */
+QString copyShareLinkToClipboard(const QString &link);
+
+/**
+ * @brief Formats the expiration timestamp for a share.
+ */
+QString formatShareExpiry(const QDateTime &expiry);
```
ZeldaZach, line 37: The whole post-response block is copy-pasted verbatim into three places: `dlg_share_deck.cpp:73-87`, `tab_deck_storage.cpp:780-792` and `tab_deck_storage_visual.cpp:170-182`. Extension extraction, the `QT_VERSION_CHECK(6, 7, 0)` timezone branch, `buildShareLink`, `copyShareLinkToClipboard` and the expiry string, three times over.

`DeckShareUtils` already exists as the shared home for exactly this. Adding something like `QString handleShareResponse(const AbstractClient *client, const Response &response)`, or at minimum `QDateTime expiryFromResponse(quint64 secs)`, would put the Qt version guard in one place and shrink the three call sites to a couple of lines each. It would also have prevented the missing `<QTimeZone>` include I flagged separately, which is precisely the kind of thing that slips through when a block gets copied a third time.

cockatrice/src/interface/widgets/cards/additional_info/deck_color_identity.cpp
```diff
@@ -0,0 +1,37 @@
+#include "deck_color_identity.h"
+
+#include <QSet>
+#include <libcockatrice/card/database/card_database_querier.h>
+#include <libcockatrice/deck_list/deck_list.h>
+#include <libcockatrice/deck_list/tree/inner_deck_list_node.h>
+
+QString getDeckColorIdentity(const DeckList &deck, const CardDatabaseQuerier *db)
```
ZeldaZach, line 8: This is a line-for-line copy of the existing `static QString computeColorIdentity(const LoadedDeck &)` at `visual_deck_storage_model.cpp:373`, and that original is still there and still used at lines 309 and 417. So the header's claim that this is "shared as a free function so the deck storage previews and the deck share dialog compute identities identically" is not true as of this PR.

Delete `computeColorIdentity` and have `visual_deck_storage_model.cpp` call `getDeckColorIdentity(deck.deckList, CardDatabaseManager::query())`, otherwise the two will drift and the previews will disagree with the share dialog.

Relatedly, `deck_preview_widget.cpp:7` now includes `deck_color_identity.h` but never calls anything from it. Drop that include.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+QString copyShareLinkToClipboard(const QString &link)
+{
+    QGuiApplication::clipboard()->setText(link);
+    return link;
+}
+
+QString formatShareExpiry(const QDateTime &expiry)
+{
+    return expiry.toLocalTime().toString();
```
ZeldaZach, line 25: `QDateTime::toString()` with no argument defaults to `Qt::TextDate`, which is a locale-independent English format ("Wed Sep 10 14:03:22 2026"). This string is shown to users in every share confirmation.

The codebase already has the convention for this at `user_info_box.cpp:170`:

```cpp
return QLocale().toString(expiry.toLocalTime(), QLocale::ShortFormat);
```

cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
```diff
@@ -0,0 +1,89 @@
...
+    buttonBox->button(QDialogButtonBox::Cancel)->setText(tr("Cancel"));
+    connect(buttonBox, &QDialogButtonBox::accepted, this, &DlgShareDeck::actShare);
+    connect(buttonBox, &QDialogButtonBox::rejected, this, &DlgShareDeck::reject);
+    this->buttonBox = buttonBox;
+    layout->addWidget(buttonBox);
+}
+
+void DlgShareDeck::actShare()
+{
+    buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
```
ZeldaZach, line 46: `RemoteClient::ping()` drops timed-out commands with `pend->deleteLater()` without ever calling `processResponse`, so `finished` is never emitted on timeout. When that happens this dialog sits with Ok permanently greyed out and no message at all. The same applies to `shareBar->setCreateEnabled(false)` in both storage tabs.

Either start a `QTimer::singleShot` alongside the send that re-enables the button and shows a "no response from server" message, or hook `AbstractClient::serverTimeout`.

Separately worth thinking about: if the user cancels this dialog mid-flight, the share is still created server side and the token is lost with no way to recover it, since there is no share-listing UI in this PR.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
```diff
@@ -33,3 +74,127 @@ void TabDeckStorageVisual::actOpenLocalDeck(const QString &filePath)
...
+        showShareNotice(tr("Failed to create the share link (server response code %1).")
+                            .arg(QString::number(static_cast<int>(response.response_code()))),
+                        true);
+        return;
+    }
+
+    const Response_DeckShareCreate &resp = response.GetExtension(Response_DeckShareCreate::ext);
+    const QString token = QString::fromStdString(resp.token());
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+    const QDateTime expiry = QDateTime::fromSecsSinceEpoch(resp.expires_at(), QTimeZone::UTC);
```
ZeldaZach, line 173: This file uses `QTimeZone::UTC` but never includes `<QTimeZone>`, unlike the two sibling copies in `dlg_share_deck.cpp:11` and `tab_deck_storage.cpp:22` which both add it. `<QDateTime>` only forward-declares `QTimeZone`, so this compiles by accident through some transitive include and will break on any header reshuffle or a different Qt version.

Add `#include <QTimeZone>` here. Conversely `deck_share_utils.cpp:5` includes `<QTimeZone>` and never uses it, so that one can go.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <QGuiApplication>
+#include <QTimeZone>
+#include <libcockatrice/network/client/abstract/abstract_client.h>
+
+namespace DeckShareUtils
+{
+
+QString buildShareLink(const AbstractClient *client, const QString &token)
+{
+    return QString("cockatrice://opendeck?share=%1&hostname=%2&port=%3")
```
ZeldaZach, line 13: The token and hostname are spliced into the query string without percent-encoding, so an IPv6 literal, an IDN hostname, or a token containing `&`, `=` or `#` produces a link the receiving parser cannot round-trip. Build it with `QUrl` plus `QUrlQuery`, which encodes automatically and makes the parsing side in the follow-up PR symmetric with this one.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
```diff
@@ -33,3 +74,127 @@ void TabDeckStorageVisual::actOpenLocalDeck(const QString &filePath)
...
+}
+
+void TabDeckStorageVisual::updateShareHint()
+{
+    const int count = visualDeckStorageWidget->selectedFilePaths().size();
+    shareBar->setCountText(tr("%n deck(s)", "", count));
+    if (count == 0) {
+        shareBar->setHintText(tr("Click deck tiles to select the decks you want to share."), true);
+    } else {
+        shareBar->setHintText(tr("%n deck(s) selected. Create the link to share %1 with other players.", "", count)
```
ZeldaZach, line 122: Splicing a separately translated pronoun ("it" / "them") into `%1` does not survive translation into languages with grammatical gender or case. The translator receives a bare "it" with no context and cannot inflect it.

You already have the `%n` plural machinery here, so write the full sentence inside the plural form and let the translator handle both branches rather than assembling it from fragments.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
```diff
@@ -625,3 +651,151 @@ void TabDeckStorage::deleteFolderFinished(const Response &response, const Comman
...
+    shareBar->setHintText(hint, !hint.isEmpty());
+
+    QStringList parts;
+    if (folders > 0) {
+        parts << tr("%n folder(s)", "", folders);
+    }
+    if (files > 0) {
+        parts << tr("%n deck(s)", "", files);
+    }
+    shareBar->setCountText(parts.isEmpty() ? tr("No decks selected") : tr("Selected: %1").arg(parts.join(tr(", "))));
```
ZeldaZach, line 711: `tr(", ")` sends a bare comma-space to translators with no context, which they cannot act on and which just pollutes the .ts files. Use `QStringLiteral(", ")` for the joiner and keep the translatable text in the surrounding `tr("Selected: %1")`.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
```diff
@@ -47,6 +47,13 @@ VisualDeckStorageWidget::VisualDeckStorageWidget(QWidget *parent) : QWidget(pare
     refreshButton->setFixedSize(32, 32);
     connect(refreshButton, &QPushButton::clicked, this, &VisualDeckStorageWidget::refreshIfPossible);
 
+    shareButton = new QToolButton(this);
+    shareButton->setIcon(QPixmap("theme:icons/share"));
+    shareButton->setFixedSize(32, 32);
+    shareButton->setToolTip(tr("Select decks to share"));
```
ZeldaZach, line 53: This class has a `retranslateUi()` that sets `refreshButton->setToolTip(...)` and `quickSettingsWidget->setToolTip(...)` three lines apart, so the new share button tooltip belongs there too or it will not update on a language switch.

While you are in there: `TabDeckStorageVisual::retranslateUi()` now only forwards to `shareBar` and never calls `visualDeckStorageWidget->retranslateUi()`, so nothing in this widget is retranslated at runtime at all.

cockatrice/src/interface/widgets/deck_share/share_bar_widget.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    retranslateUi();
+}
+
+void ShareBarWidget::retranslateUi()
+{
+    nameEdit->setPlaceholderText(tr("Share name"));
+    cancelButton->setText(tr("Cancel"));
+    createButton->setText(tr("Create share link"));
+    hintLabel->setText(tr("Click deck tiles to select the decks you want to share."));
```
ZeldaZach, line 46: Two problems with putting this text here. It hardcodes visual-storage wording ("deck tiles") into a widget whose own docs describe it as reusable by the server tree tab, where the right hint is something like "Select folders or decks in the tree to share." And calling `retranslateUi()` while share mode is active clobbers whatever dynamic hint the owning tab set through `setHintText`.

Leave `hintLabel` empty here and let each owner supply the hint, or have the owner re-push its hint after `retranslateUi()`.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
```diff
@@ -0,0 +1,28 @@
...
+    QGuiApplication::clipboard()->setText(link);
+    return link;
+}
+
+QString formatShareExpiry(const QDateTime &expiry)
+{
+    return expiry.toLocalTime().toString();
+}
+
+} // namespace DeckShareUtils
```
ZeldaZach, line 28: Eight files in this PR end without a trailing newline: `deck_share_utils.{cpp,h}`, `share_bar_widget.{cpp,h}`, `dlg_share_deck.{cpp,h}` and `tab_deck_storage_visual.{cpp,h}`. The last pair is a regression on pre-existing files that previously ended with a newline after `#endif`.

ZeldaZach (changes requested):
The share flow is put together carefully in most respects. Every response handler checks `Response::RespOk` and surfaces an error, the create buttons are disabled while a command is in flight so double submission is guarded, and the dialogs are stack allocated and `exec`'d, so Qt's automatic receiver disconnection makes the in-flight `PendingCommand` responses safe against use-after-free. `ShareBarWidget` is a genuinely reusable component rather than a copy in each tab.

One scoping note, since the PR description promises more than the diff contains. This delta has no `cockatrice://` or deck-storage URL parsing, no share-listing page and no shared-deck preview dialog, so there is no host-matching or rich-text-injection surface to review here. I reviewed that in #7244 instead.

The things I want fixed before merge are mostly around client resolution and duplication. `TabDeckStorageVisual` caches `tabSupervisor->getClient()` at construction, which hands you a `LocalClient` if the tab is opened during an offline game. That client never emits `statusChanged`, so the share button stays hidden for the rest of the tab's life even after a real login. Worse, the gate uses that cached pointer while the send re-resolves the client, so being logged in and then starting a local game sends the command to the `LocalClient` and produces a link reading `hostname=&port=0`.

The duplication is the other theme. The entire post-response block is pasted verbatim into three files, including the Qt 6.7 timezone branch, and one of the three copies forgot the `<QTimeZone>` include it needs, compiling only through a transitive include. `DeckShareUtils` already exists as the home for this. In the same vein, the new `getDeckColorIdentity` is a line-for-line copy of `computeColorIdentity` in `visual_deck_storage_model.cpp`, which is still present and still used, so the header comment claiming the two paths now compute identities identically is not yet true.

Two smaller ones worth doing: the expiry string uses `QDateTime::toString()` with no argument, which is locale-independent English rather than the `QLocale().toString(..., QLocale::ShortFormat)` this codebase uses elsewhere, and the share link splices token and hostname into a query string without percent-encoding.

Finally, the same `RemoteClient::ping()` timeout path I flagged on #7245 bites here too: a dropped reply never emits `finished`, so the dialog is left with Ok greyed out and no message. Worth solving once for all three call sites.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
BruebachL: Fixed in a1950a2637 + a43f85c5e2: all eight listed files now end with a trailing newline.

cockatrice/src/interface/widgets/cards/additional_info/deck_color_identity.cpp
BruebachL: Fixed in 07ab6c5698: the duplicate computeColorIdentity in visual_deck_storage_model.cpp is deleted and both call sites now use the shared getDeckColorIdentity(deckList, CardDatabaseManager::query()). The now-unused deck_color_identity.h include in deck_preview_widget.cpp was also dropped.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
BruebachL: Fixed in c27f980859: the constructor now resolves the stable server client once, so the tab never caches a LocalClient and the share button updates correctly after the user logs in.

cockatrice/src/interface/widgets/dialogs/dlg_share_deck.cpp
BruebachL: Fixed in bd91ead05d: each send site (dlg_share_deck, tab_deck_storage, tab_deck_storage_visual) starts a single-shot timer alongside the send, set to (getTimeOut()+1)*getKeepAlive()*1000 ms to mirror the RemoteClient ping-drop window, stopped on finished, and on timeout re-enables the button and shows a `The server did not respond in time. Try again.` message. (NetworkSettings include wired in 3b718bbf47.)

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
BruebachL: Fixed in 447d749e34: the full sentence now lives inside the plural forms (`One deck selected. Create the link to share it with other players.` / `%n decks selected. Create the link to share them with other players.`), so translators handle both branches without a spliced pronoun.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
BruebachL: Fixed in c742ff5a71: formatShareExpiry now uses QLocale().toString(expiry.toLocalTime(), QLocale::ShortFormat).

cockatrice/src/interface/widgets/deck_share/deck_share_utils.cpp
BruebachL: Fixed in 507f6a042c: buildShareLink now builds the url with QUrl + QUrlQuery (scheme cockatrice, host opendeck, share/hostname/port query items) and returns url.toString(QUrl::FullyEncoded), which percent-encodes token and hostname symmetrically with the parsing side.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
BruebachL: Fixed in e788b862ef: the Qt version guard now lives only in DeckShareUtils::handleShareResponse, where <QTimeZone> is genuinely used; tab_deck_storage_visual no longer touches QTimeZone, and the stale QTimeZone/QDateTime includes were removed from dlg_share_deck and tab_deck_storage.

cockatrice/src/interface/widgets/deck_share/share_bar_widget.cpp
BruebachL: Fixed in 5acaaf6b9b: ShareBarWidget::retranslateUi() no longer hardcodes a hint; each owner re-pushes its hint after retranslateUi (TabDeckStorage::retranslateUi calls onServerSelectionChanged when the bar is visible, TabDeckStorageVisual::retranslateUi calls updateShareHint()), so the per-context wording survives language switches.

cockatrice/src/interface/widgets/deck_share/deck_share_utils.h
BruebachL: Fixed in 1f7b815b4c: added DeckShareUtils::handleShareResponse(client, response) returning a ShareResponse{link, expiry}, which owns the extension extraction, the Qt version guard, link building and clipboard copy. The three call sites (dlg_share_deck, tab_deck_storage, tab_deck_storage_visual) now just format the message and act on the result.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
BruebachL: Fixed in c27f980859: actShareSelected now uses the client resolved once in the constructor for the gate, the send and buildShareLink, so offline games cannot route the command to a LocalClient.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
BruebachL: Fixed in 14d536a732 + fba5d015c9: the share button tooltip moved into VisualDeckStorageWidget::retranslateUi(), and TabDeckStorageVisual::retranslateUi() now forwards to visualDeckStorageWidget->retranslateUi() so the widget retranslates at runtime.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
BruebachL: Fixed in ed03a7c5df: the joiner is now QStringLiteral(", ") and only the surrounding `Selected: %1` stays translatable.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
```diff
@@ -139,6 +164,10 @@ TabDeckStorage::TabDeckStorage(TabSupervisor *_tabSupervisor,
     aDeleteRemoteDeck->setIcon(themePixmap(QStringLiteral("icons/remove_row")));
     connect(aDeleteRemoteDeck, &QAction::triggered, this, &TabDeckStorage::actDeleteRemoteDeck);
 
+    aShareDecks = new QAction(this);
+    aShareDecks->setIcon(QPixmap("theme:icons/share"));
```
ZeldaZach: The theme resource is looked up without a file extension, so this pixmap is always null and the toolbar action renders with no icon. Every other action here goes through `themePixmap()`, which probes `.png/.jpg/.jpeg/.svg` (and the dark/light variant) via `ThemeManager::assetPath()`; the asset on disk is `resources/icons/share.svg`, so `QPixmap("theme:icons/share")` cannot resolve it.

```suggestion
    aShareDecks->setIcon(themePixmap(QStringLiteral("icons/share")));
```

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
```diff
@@ -48,6 +48,12 @@ VisualDeckStorageWidget::VisualDeckStorageWidget(QWidget *parent) : QWidget(pare
     refreshButton->setFixedSize(32, 32);
     connect(refreshButton, &QPushButton::clicked, this, &VisualDeckStorageWidget::refreshIfPossible);
 
+    shareButton = new QToolButton(this);
+    shareButton->setIcon(QPixmap("theme:icons/share"));
```
ZeldaZach: Same problem as the toolbar action in `tab_deck_storage.cpp`: `"theme:icons/share"` has no file extension, so `ThemeManager::assetPath()` is bypassed and the QPixmap is null — the share tool button renders blank. The asset on disk is `resources/icons/share.svg`; use `themePixmap(QStringLiteral("icons/share"))` instead, which probes the extensions and the dark/light variant.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
```diff
@@ -158,11 +165,69 @@ void VisualDeckStorageWidget::retranslateUi()
...
+    }
+    emit shareSelectionChanged();
+}
+
+bool VisualDeckStorageWidget::isShareSelectable() const
+{
+    return shareSelectable;
+}
+
+QStringList VisualDeckStorageWidget::selectedFilePaths() const
```
ZeldaZach, line 191: This walks every `DeckPreviewWidget` that exists, but filtered-out previews are not destroyed — `continueDeckPass()` does `deckPreviewWidget->setVisible(matches)` and `finishDeckPass()` explicitly keeps them ("Filtered-out decks stay part of the layout, hidden in their sorted place"). So a hidden preview that is still `shareSelected` is silently included.

Scenario: enter share mode, click deck A to select it, then type into the search box (or change the tag/color filter) so A no longer matches, then click deck B and press "Create share link". The share is created with both A and B, even though A is invisible and its highlight frame cannot be seen; `updateShareHint()` also counts A in the "%n deck(s)" label, so the count does not match what is on screen.

Filtering on `preview->isVisible()` here (and clearing the selection of previews that get hidden in `continueDeckPass()`) would keep the selection consistent with what the user can see.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
```diff
@@ -626,3 +663,151 @@ void TabDeckStorage::deleteFolderFinished(const Response &response, const Comman
...
+{
+    setShareModeEnabled(true);
+}
+
+void TabDeckStorage::cancelShareDecks()
+{
+    setShareModeEnabled(false);
+}
+
+void TabDeckStorage::setShareModeEnabled(bool enabled)
```
ZeldaZach, line 677: Leaving share mode never stops `shareTimeoutTimer`, and the in-flight `Command_DeckShareCreate` is not abandoned either.

Scenario: press "Create share link", then press Cancel (or get disconnected, which routes through `setRemoteEnabled(false)` -> `setShareModeEnabled(false)`) while the request is still outstanding. A few seconds later `onShareFromTreeTimeout()` pops "The server did not respond in time" even though the user cancelled. And if the response does arrive after the cancel, `shareFromTreeFinished()` still runs: it copies the link to the clipboard and shows "Share link copied to the clipboard" for a share the user backed out of.

Also, because `setShareModeEnabled(true)` re-enables the create button, cancelling and re-entering share mode lets a second command be sent while the first is still pending. Stopping the timer here (and tracking/ignoring a stale pending command) fixes both.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
```diff
@@ -33,3 +86,127 @@ void TabDeckStorageVisual::actOpenLocalDeck(const QString &filePath)
...
+    shareBar->setCreateEnabled(true);
+    visualDeckStorageWidget->setShareSelectable(true);
+    visualDeckStorageWidget->setShareSelectedFiles(preselectFiles);
+    shareBar->setName(tr("Shared decks"));
+    shareBar->setVisible(true);
+    updateShareHint();
+    shareBar->focusName();
+}
+
+void TabDeckStorageVisual::exitShareMode()
```
ZeldaZach, line 104: Same issue as `TabDeckStorage::setShareModeEnabled()`: `exitShareMode()` does not stop `shareTimeoutTimer`.

Scenario: press "Create share link", then Cancel while the request is still outstanding (or lose the connection — `handleConnectionChanged()` calls `exitShareMode()` on any non-logged-in status). The timer keeps running and `onShareTimeout()` shows "The server did not respond in time" for a share the user already cancelled; conversely a late successful response still runs `shareFinished()`, copying the link to the clipboard and announcing success. Add `shareTimeoutTimer->stop();` here and ignore responses for a share that was abandoned.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
```diff
@@ -1,25 +1,78 @@
...
+            &TabDeckStorageVisual::actShareDeck);
+    connect(visualDeckStorageWidget, &VisualDeckStorageWidget::shareSelectionChanged, this,
+            &TabDeckStorageVisual::onShareSelectionChanged);
+    connect(visualDeckStorageWidget, &VisualDeckStorageWidget::shareRequested, this, [this] {
+        if (shareDeckAvailable) {
+            enterShareMode();
+        }
+    });
+
+    connect(client, &AbstractClient::statusChanged, this, &TabDeckStorageVisual::handleConnectionChanged);
```
ZeldaZach: This connects `statusChanged` to `handleConnectionChanged()` 12 lines before `shareBar` is created (line 53), and `shareBar` has no in-class initializer in the header. `handleConnectionChanged()` dereferences `shareBar->isVisible()`, so any status change delivered before the assignment reads an indeterminate pointer. `RemoteClient` lives on its own thread, so the connection is queued and in practice the slot cannot run during the constructor — but the ordering is fragile for no reason. Move this `connect()` (and the `setShareAvailable()` seeding below it) after `shareBar` is constructed, and/or give `shareBar` a `= nullptr` initializer like `selectionFrame` has.

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
```diff
@@ -382,6 +383,25 @@ bool AbstractTabDeckEditor::actSaveDeckAs()
...
+ */
+void AbstractTabDeckEditor::actShareDeck()
+{
+    if (tabSupervisor->getClient()->getStatus() != StatusLoggedIn) {
+        QMessageBox::information(this, tr("Share deck"), tr("You must be connected to the server to share a deck."));
+        return;
+    }
+
+    const QSharedPointer<DeckList> deck = deckStateManager->getDeckListShared();
+    if (deck->isBlankDeck()) {
```
ZeldaZach, line 398: A blank deck exits silently: the dialog never opens and the user gets no explanation. Note the menu entry is only disabled via `setSaveStatus()`, which is not the same predicate (`isBlankNewDeck()` also requires `!isModified()`), so this path is reachable — e.g. add a card and remove it again, then pick "Share deck...". A short `QMessageBox::information` (like the not-logged-in branch just above) would be more consistent.

cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
```diff
@@ -226,10 +295,6 @@ void DeckPreviewWidget::updateTagsVisibility(bool visible)
     }
 }
 
-/**
- * Refreshes the banner card text.
- * This also calls `refreshBannerCardToolTip`, since those two often need to be updated together.
- */
 void DeckPreviewWidget::refreshBannerCardText()
```
ZeldaZach, line 302: The doc comment above `refreshBannerCardText()` ("Refreshes the banner card text. This also calls `refreshBannerCardToolTip`...") is deleted here, which looks unrelated to deck sharing — probably an accidental drive-by removal. Worth restoring.

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
```diff
@@ -382,6 +383,25 @@ bool AbstractTabDeckEditor::actSaveDeckAs()
     return true;
 }
 
+/**
+ * @brief Opens the deck share dialog with the current deck preselected.
+ */
+void AbstractTabDeckEditor::actShareDeck()
+{
+    if (tabSupervisor->getClient()->getStatus() != StatusLoggedIn) {
```
ZeldaZach: Non-blocking, but this is the same `getClient()` hazard I flagged on the visual tab last round, in the one place that still resolves through `TabSupervisor`.

`TabSupervisor::getClient()` is `localClients.isEmpty() ? client : localClients.first()`, so while an offline game is running this hands back a `LocalClient`. `LocalClient` never calls `setStatus`, so `getStatus()` stays at its default and this gate always fails: a logged-in user who starts a local game can no longer share from the deck editor, and the message they get ("You must be connected to the server") is wrong.

The good news is that it fails closed rather than producing the `hostname=&port=0` link, because the gate and the `DlgShareDeck` construction below both go through the same `getClient()` call. The fix is the same one you applied to `TabDeckStorageVisual`: resolve the stable server client (`tabSupervisor` holds it as its `client` member, which is what `TabDeckStorage` already uses) instead of `getClient()`.

ZeldaZach (approved):
Everything from the last round is addressed, and I checked the substantive ones against the diff rather than taking the replies at face value.

The client resolution is genuinely fixed: `TabDeckStorageVisual` now takes `TabSupervisor::client` (the stable remote client, assigned once in the supervisor's constructor and never reassigned) and uses that same pointer for the gate, the send and `buildShareLink`, so the `LocalClient` path with its `hostname=&port=0` link is gone. `DeckShareUtils::handleShareResponse` collapses the three pasted post-response blocks into one place, which takes the Qt 6.7 timezone guard and the missing `<QTimeZone>` include with it. `buildShareLink` builds through `QUrl` + `QUrlQuery` and returns `FullyEncoded`, `formatShareExpiry` uses `QLocale::ShortFormat`, the duplicate `computeColorIdentity` is deleted with both call sites moved over, and the timeout timers recover the Create button on a dropped reply.

On the lifetime questions: there are no `QNetworkReply`s here, and every `PendingCommand::finished` connection passes `this` as the receiver, so a dialog or tab going away mid-flight disconnects cleanly. The command itself is owned and `deleteLater`d by the client. Nothing logs the token or the share URL — the only `qWarning` carries the response code.

Nine non-blocking findings left inline. The two worth doing before this ships are the null theme pixmaps (`QPixmap("theme:icons/share")` skips `ThemeManager::assetPath()`, which is what probes the extensions, so both share buttons render a null icon against `share.svg`) and `selectedFilePaths()` including previews that the search/tag filter has hidden but not destroyed. The rest are smaller: the timeout timer surviving a cancel, `statusChanged` being connected before `shareBar` exists, the silent blank-deck exit, a stray doc-comment removal, and the deck editor gate still going through `getClient()`.

None of that blocks. Approving.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
BruebachL: Fixed in 09bb9449d3: aShareDecks now resolves through themePixmap(QStringLiteral("icons/share")), so the toolbar action renders share.svg (and the dark/light variants) instead of a null pixmap.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
BruebachL: Fixed in 09bb9449d3: the tile-share tool button now uses themePixmap(QStringLiteral("icons/share")) like the tree tab, probing the extensions and the dark/light variant.

cockatrice/src/interface/widgets/visual_deck_storage/visual_deck_storage_widget.cpp
BruebachL: Fixed in 77872de85e: selectedFilePaths() now skips previews that are not visible, so filtered-out decks can no longer sneak into the share or the %n count. continueDeckPass() also deselects a deck that stops matching the filters, so the highlight frame cannot linger on an invisible preview.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
BruebachL: Fixed in ee6c8f8e8f: setShareModeEnabled(false) now stops shareTimeoutTimer and abandons the in-flight Command_DeckShareCreate via a sequence-numbered request. A stale reply or a timed-out response after cancel is ignored, and re-entering share mode with the old request still outstanding no longer lets the two responses collide.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
BruebachL: Fixed in cb8f27176f: exitShareMode() stops shareTimeoutTimer and abandons the pending request with the same sequence-number tracking as the tree tab, so a late response or timer pop after cancel never reports the share as created.

cockatrice/src/interface/widgets/tabs/visual_deck_storage/tab_deck_storage_visual.cpp
BruebachL: Fixed in 6c4a3f5346: the statusChanged connect and the setShareAvailable seeding now run after shareBar is constructed, and shareBar gained a = nullptr in-class initializer, so handleConnectionChanged() can never dereference an uninitialized pointer.

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
BruebachL: Fixed in a796612c53: a blank deck now shows a short QMessageBox::information ("The deck is empty. Add cards before sharing it.") mirroring the not-logged-in branch.

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
BruebachL: Fixed in d7cf2befbb: added TabSupervisor::getServerClient() returning the stable remote client, and actShareDeck now uses that same pointer for the gate and the DlgShareDeck construction, so an offline game can no longer fail the share with a misleading message.

cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/deck_preview_widget.cpp
BruebachL: Fixed in 4fb611c9ae: the doc comment above refreshBannerCardText() is restored.

PR #7244, opened 2026-09-05
[DeckShare] Open shared decks via links with a gated preview flow

BruebachL:

A received share link only worked when manually pasted into the share dialog
while already logged in. Clicking a `cockatrice://` link in a browser did
nothing, and the client's passive URL processing could not be paused while the
user inspected the deck.

- Route external deck links through a url-parser that mirrors server links to
  the `//deck` URI and pauses passive processing while a share dialog is open
- Add a gated preview flow: the link opens an `Open deck` dialog instead of
  downloading immediately, letting the holder see the share value and click
  `Open deck` to fetch, with the deck appearing in the local deck storage on
  acceptance
- Support opening shared decks while logged in at construction time, and resume
  the current session's deck storage after a share is accepted
- Match `cockatrice://` URIs against the configured server list at startup, wait
  for the login/connection to settle before resuming, and show an informative
  error when the host is unknown or the link is hostile
- Make the share preview tile keyboard-navigable (Tab focus, Space/Enter and
  double-click open) with `Share value` microcopy

## Why?

The next PR adds the public-decks browsing tab on top of the shared deck list
command this flow was built against.

### DIFF
--- a/cockatrice/src/interface/intents/intent.cpp
+++ b/cockatrice/src/interface/intents/intent.cpp
@@ -2,10 +2,11 @@
 
 Intent::Intent(QObject *parent) : QObject(parent)
 {
-    // An intent is done as soon as it reports success or failure. Deleting it
-    // also tears down its dependency chain and disconnects any signal wiring.
+    // An intent is done as soon as it reports success, failure, or cancellation.
+    // Deleting it also tears down its dependency chain and disconnects any signal wiring.
     connect(this, &Intent::finished, this, &QObject::deleteLater);
     connect(this, &Intent::failed, this, &QObject::deleteLater);
+    connect(this, &Intent::cancelled, this, &QObject::deleteLater);
 }
 
 Intent::~Intent() = default;
@@ -44,5 +45,13 @@ void Intent::emitFailed(const QString &reason)
     if (!completed) {
         completed = true;
         emit failed(reason);
+    }
+}
+
+void Intent::emitCancelled()
+{
+    if (!completed) {
+        completed = true;
+        emit cancelled();
     }
 }
--- a/cockatrice/src/interface/intents/intent_join_server_game.cpp
+++ b/cockatrice/src/interface/intents/intent_join_server_game.cpp
@@ -19,13 +19,10 @@ bool IntentJoinServerGame::checkPrecondition() const
     if (remoteClient->getStatus() != ClientStatus::StatusLoggedIn) {
         return false;
     }
-    // peerPort() reflects the actual TCP peer, which may differ from the
-    // configured server port (e.g. when connecting through a proxy), so only
-    // the hostname is compared here.
-    if (remoteClient->peerName() != context->roomContext.serverContext.hostname) {
-        return false;
-    }
-    if (QString::number(remoteClient->peerPort()) != context->roomContext.serverContext.port) {
+    // serverName() reflects the server the client was configured to connect to,
+    // which may differ from the actual TCP peer (e.g. when connecting through a
+    // proxy), so only the hostname is compared here.
+    if (remoteClient->serverName().compare(context->roomContext.serverContext.hostname, Qt::CaseInsensitive) != 0) {
         return false;
     }
 
--- a/cockatrice/src/interface/intents/intent_login.cpp
+++ b/cockatrice/src/interface/intents/intent_login.cpp
@@ -1,7 +1,10 @@
 #include "intent_login.h"
 
 #include "../../client/settings/cache_settings.h"
+#include "../widgets/dialogs/dlg_login_prompt.h"
 #include "libcockatrice/settings/servers_settings.h"
+
+#include <QDialog>
 
 IntentGetLoginCredentials::IntentGetLoginCredentials(ContextConnectToServer *_context) : Intent(), context(_context)
 {
@@ -29,5 +32,27 @@ void IntentGetLoginCredentials::onPreconditionSatisfied()
 
 void IntentGetLoginCredentials::onPreconditionNotSatisfied()
 {
-    emitFailed(tr("No saved credentials for this server"));
+    // No credentials saved for the target server: ask the user for them. They
+    // opt into saving them so later links to the same server connect directly.
+    const QString serverText = context->hostname + ":" + context->port;
+    DlgLoginPrompt dialog(serverText);
+    // ApplicationModal: the dialog has no parent (the intent is not a widget),
+    // so WindowModal would not actually block any other window.
+    dialog.setWindowModality(Qt::ApplicationModal);
+
+    if (dialog.exec() != QDialog::Accepted) {
+        emitCancelled();
+        return;
+    }
+
+    context->username = dialog.username();
+    context->password = dialog.password();
+
+    if (dialog.savePassword() && !context->username.isEmpty()) {
+        ServersSettings &servers = SettingsCache::instance().servers();
+        servers.addNewServer(context->hostname, context->hostname, context->port, context->username, context->password,
+                             true);
+    }
+
+    emitFinished();
 }
--- a/cockatrice/src/interface/intents/intent_open_shared_deck.cpp
+++ b/cockatrice/src/interface/intents/intent_open_shared_deck.cpp
@@ -1,0 +1,190 @@
+#include "intent_open_shared_deck.h"
+
+#include "../deck_loader/deck_loader.h"
+#include "../widgets/dialogs/dlg_shared_decks_preview.h"
+#include "../widgets/tabs/tab_supervisor.h"
+#include "intent_connect_to_server.h"
+
+#include <QMessageBox>
+#include <QTimer>
+#include <libcockatrice/card/database/card_database_querier.h>
+#include <libcockatrice/protocol/pb/command_deck_share_download.pb.h>
+#include <libcockatrice/protocol/pb/command_deck_share_list.pb.h>
+#include <libcockatrice/protocol/pb/response.pb.h>
+#include <libcockatrice/protocol/pb/response_deck_share_download.pb.h>
+#include <libcockatrice/protocol/pb/response_deck_share_list.pb.h>
+#include <libcockatrice/protocol/pb/serverinfo_deck_share_item.pb.h>
+#include <libcockatrice/protocol/pending_command.h>
+
+IntentOpenSharedDeck::IntentOpenSharedDeck(TabSupervisor *_tabSupervisor,
+                                           RemoteClient *_remoteClient,
+                                           const CardDatabaseQuerier *_querier,
+                                           std::unique_ptr<ContextOpenDeck> _context)
+    : Intent(), tabSupervisor(_tabSupervisor), remoteClient(_remoteClient), querier(_querier),
+      context(_context.release())
+{
+    downloadTimer = new QTimer(this);
+    downloadTimer->setSingleShot(true);
+    downloadTimer->setInterval(15000);
+    connect(downloadTimer, &QTimer::timeout, this, &IntentOpenSharedDeck::onDownloadTimeout);
+}
+
+bool IntentOpenSharedDeck::checkPrecondition() const
+{
+    if (remoteClient->getStatus() != ClientStatus::StatusLoggedIn) {
+        return false;
+    }
+    // serverName() reflects the server the client was configured to connect to,
+    // which may differ from the actual TCP peer (e.g. when connecting through a
+    // proxy), so only the hostname is compared here.
+    return remoteClient->serverName().compare(context->serverContext.hostname, Qt::CaseInsensitive) == 0;
+}
+
+void IntentOpenSharedDeck::onPreconditionSatisfied()
+{
+    // Resolve the share token to its items first; a share can contain more than
+    // one deck, and each item is downloaded by id.
+    Command_DeckShareList cmd;
+    cmd.set_token(context->shareToken.toStdString());
+
+    PendingCommand *pend = AbstractClient::prepareSessionCommand(cmd);
+    connect(pend, &PendingCommand::finished, this, &IntentOpenSharedDeck::listShareFinished);
+    remoteClient->sendCommand(pend);
+}
+
+void IntentOpenSharedDeck::onPreconditionNotSatisfied()
+{
+    runDependency(new IntentConnectToServer(remoteClient, &context->serverContext));
+}
+
+void IntentOpenSharedDeck::listShareFinished(const Response &response, const CommandContainer & /* commandContainer */)
+{
+    if (response.response_code() != Response::RespOk) {
+        emitFailed(tr("The shared deck could not be found or has expired"));
+        return;
+    }
+
+    const Response_DeckShareList &resp = response.GetExtension(Response_DeckShareList::ext);
+    if (resp.items_size() == 0) {
+        emitFailed(tr("The shared deck is empty"));
+        return;
+    }
+
+    QList<ServerInfo_DeckShareItem> items;
+    items.reserve(resp.items_size());
+    for (const ServerInfo_DeckShareItem &item : resp.items()) {
+        items.append(item);
+        itemNames.insert(item.id(), QString::fromStdString(item.name()));
+    }
+
+    const QString serverText = context->serverContext.hostname + ":" + context->serverContext.port;
+
+    // Ask the user which decks to open before downloading anything.
+    previewDialog = new DlgSharedDecksPreview(tabSupervisor, querier, QString::fromStdString(resp.name()),
+                                              resp.expires_at(), serverText, items);
+    connect(previewDialog, &DlgSharedDecksPreview::openRequested, this, &IntentOpenSharedDeck::startDownloads);
+    connect(previewDialog, &DlgSharedDecksPreview::cancelled, this, &IntentOpenSharedDeck::emitCancelled);
+    connect(previewDialog, &DlgSharedDecksPreview::cancelled, previewDialog, &QWidget::deleteLater);
+    previewDialog->show();
+    previewDialog->raise();
+    previewDialog->activateWindow();
+}
+
+void IntentOpenSharedDeck::startDownloads(const QList<int> &itemIds)
+{
+    pendingItemIds = itemIds;
+    totalItems = itemIds.size();
+    completedItems = 0;
+    loadedDecks.clear();
+    downloadNextItem();
+}
+
+void IntentOpenSharedDeck::downloadNextItem()
+{
+    if (pendingItemIds.isEmpty()) {
+        finishAll();
+        return;
+    }
+
+    currentItemId = pendingItemIds.takeFirst();
+    downloadTimer->start();
+
+    Command_DeckShareDownload cmd;
+    cmd.set_token(context->shareToken.toStdString());
+    cmd.set_item_id(currentItemId);
+
+    PendingCommand *pend = AbstractClient::prepareSessionCommand(cmd);
+    connect(pend, &PendingCommand::finished, this, &IntentOpenSharedDeck::downloadShareFinished);
+    remoteClient->sendCommand(pend);
+}
+
+void IntentOpenSharedDeck::downloadShareFinished(const Response &response,
+                                                 const CommandContainer & /* commandContainer */)
+{
+    downloadTimer->stop();
+
+    QString failureReason;
+    if (response.response_code() != Response::RespOk) {
+        failureReason = tr("Failed to download the shared deck");
+    } else {
+        const Response_DeckShareDownload &resp = response.GetExtension(Response_DeckShareDownload::ext);
+        const QString deckString = QString::fromStdString(resp.deck());
+        if (deckString.isEmpty()) {
+            failureReason = tr("The shared deck is empty");
+        } else {
+            std::optional<LoadedDeck> deckOpt =
+                DeckLoader::loadFromRemote(deckString, LoadedDeck::LoadInfo::NON_REMOTE_ID);
+            if (!deckOpt) {
+                failureReason = tr("The shared deck could not be loaded");
+            } else {
+                loadedDecks.append(deckOpt.value());
+                ++completedItems;
+                previewDialog->setDownloadProgress(completedItems, totalItems,
+                                                   itemNames.value(currentItemId, tr("Unknown deck")));
+                downloadNextItem();
+                return;
+            }
+        }
+    }
+
+    onItemFailure(failureReason);
+}
+
+void IntentOpenSharedDeck::onItemFailure(const QString &reason)
+{
+    downloadTimer->stop();
+
+    if (loadedDecks.isEmpty()) {
+        previewDialog->deleteLater();
+        emitFailed(reason);
+        return;
+    }
+
+    const int downloadedCount = loadedDecks.size();
+    const QMessageBox::StandardButton answer = QMessageBox::question(
+        previewDialog, tr("Open shared decks"),
+        tr("Could not download the deck \"%1\".\n\n%n deck(s) were already downloaded. Open them?", "", downloadedCount)
+            .arg(itemNames.value(currentItemId, tr("Unknown deck"))),
+        QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+
+    if (answer == QMessageBox::Yes) {
+        finishAll();
+    } else {
+        previewDialog->deleteLater();
+        emitCancelled();
+    }
+}
+
+void IntentOpenSharedDeck::onDownloadTimeout()
+{
+    onItemFailure(tr("Timed out while downloading the shared deck"));
+}
+
+void IntentOpenSharedDeck::finishAll()
+{
+    previewDialog->deleteLater();
+    for (const LoadedDeck &deck : loadedDecks) {
+        tabSupervisor->openDeckInNewTab(deck);
+    }
+    emitFinished();
+}
--- a/cockatrice/src/interface/intents/url_parser.cpp
+++ b/cockatrice/src/interface/intents/url_parser.cpp
@@ -1,18 +1,27 @@
 #include "url_parser.h"
 
+#include "../../client/settings/cache_settings.h"
 #include "../widgets/tabs/tab_room.h"
 #include "../widgets/tabs/tab_supervisor.h"
 #include "../window_main.h"
 #include "contexts/context_join_game.h"
+#include "contexts/context_open_deck.h"
+#include "intent.h"
 #include "intent_join_server_game.h"
 #include "intent_login.h"
+#include "intent_open_shared_deck.h"
 
 #include <QDebug>
+#include <QLoggingCategory>
 #include <QMessageBox>
 #include <QUrl>
 #include <QUrlQuery>
+#include <libcockatrice/card/database/card_database_manager.h>
 #include <libcockatrice/network/client/abstract/abstract_client.h>
+#include <libcockatrice/settings/servers_settings.h>
 #include <memory>
+
+inline Q_LOGGING_CATEGORY(UrlParserLog, "url_parser");
 
 IntentUrlParser::IntentUrlParser(QObject *parent, MainWindow *_mainWindow) : QObject(parent), mainWindow(_mainWindow)
 {
@@ -29,16 +38,33 @@ void IntentUrlParser::handle(const QString &urlStr)
     const QString action = url.host();
     QUrlQuery query(url);
 
+    qCDebug(UrlParserLog) << "Parsing intent URL, action:" << action;
+
+    QList<Intent *> chain;
+    Intent *firstIntent = nullptr;
     if (action == "joingame") {
-        handleJoinGame(query);
+        firstIntent = createJoinGameIntent(query, chain);
     } else if (action == "opendeck") {
-        // handleOpenDeck(query);
+        firstIntent = createOpenDeckIntent(query, chain);
     } else {
         qWarning() << "Unknown intent:" << action;
     }
+
+    if (firstIntent == nullptr) {
+        // The link was invalid or the user declined the confirm: nothing runs.
+        // Report the idle state when no other chain is queued so that a startup
+        // launch (which skipped its own connection for this URL) falls back to it.
+        if (!chainRunning && pendingChains.isEmpty()) {
+            emit urlChainFinished(mainWindow->getRemoteClient()->getStatus() == StatusLoggedIn);
+        }
+        return;
+    }
+
+    pendingChains.append(chain);
+    startNextChain();
 }
 
-void IntentUrlParser::handleJoinGame(const QUrlQuery &query)
+Intent *IntentUrlParser::createJoinGameIntent(const QUrlQuery &query, QList<Intent *> &chain)
 {
     auto showError = [this](const QString &message) { QMessageBox::warning(mainWindow, tr("Open game"), message); };
 
@@ -49,21 +75,21 @@ void IntentUrlParser::handleJoinGame(const QUrlQuery &query)
 
     if (ctx->roomContext.serverContext.hostname.isEmpty()) {
         showError(tr("Missing or empty hostname in the game link"));
-        return;
+        return nullptr;
     }
 
     bool ok = false;
     ctx->roomContext.serverContext.port.toUShort(&ok);
     if (!ok) {
         showError(tr("Invalid or missing port in the game link"));
-        return;
+        return nullptr;
     }
 
     ctx->roomContext.roomId = query.queryItemValue("roomid").toInt(&ok);
 
     if (!ok) {
         showError(tr("Invalid or missing room id in the game link"));
-        return;
+        return nullptr;
     }
 
     ok = false;
@@ -71,7 +97,7 @@ void IntentUrlParser::handleJoinGame(const QUrlQuery &query)
 
     if (!ok) {
         showError(tr("Invalid or missing game id in the game link"));
-        return;
+        return nullptr;
     }
 
     const QString gameDescription = query.queryItemValue("game", QUrl::FullyDecoded);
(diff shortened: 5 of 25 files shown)


cockatrice/src/interface/intents/url_parser.cpp
```diff
@@ -134,3 +168,221 @@ QString IntentUrlParser::generateJoinGameMessage(const ContextJoinGame &context,
...
+    }
+
+    RemoteClient *client = mainWindow->getRemoteClient();
+
+    // When the link would move us away from a live session, ask first — the
+    // open deck download needs the connection the user already has. Remember
+    // the current session so a failed or cancelled chain can restore it.
+    const bool migrating =
+        client->getStatus() == StatusLoggedIn && !isConnectedTo(ctx->serverContext.hostname, ctx->serverContext.port);
+    if (migrating) {
```
ZeldaZach, line 211: This is the part I am most concerned about. The confirmation is gated on `migrating`, which requires `client->getStatus() == StatusLoggedIn`, so it only fires when the user is already logged in somewhere else. The common case, a browser click with the app not connected, falls straight through: the client connects to whatever hostname and port the URL names, with no user confirmation and no validation against the configured server list.

`createJoinGameIntent` always asks before acting. Opening a deck link should hold to the same bar. At minimum, ask whenever `!isConnectedTo(hostname, port)` and name the host in the prompt.

Beyond that, a host with no entry in `SettingsCache::instance().servers()` deserves a stronger warning than a normal connect prompt, something along the lines of "This link points to a server you have never connected to", rather than going straight to a sign-in dialog for an attacker-chosen host.

cockatrice/src/interface/intents/intent_open_shared_deck.cpp
```diff
@@ -0,0 +1,190 @@
...
+void IntentOpenSharedDeck::onPreconditionSatisfied()
+{
+    // Resolve the share token to its items first; a share can contain more than
+    // one deck, and each item is downloaded by id.
+    Command_DeckShareList cmd;
+    cmd.set_token(context->shareToken.toStdString());
+
+    PendingCommand *pend = AbstractClient::prepareSessionCommand(cmd);
+    connect(pend, &PendingCommand::finished, this, &IntentOpenSharedDeck::listShareFinished);
+    remoteClient->sendCommand(pend);
```
ZeldaZach, line 52: There is no timeout around the `DeckShareList` round trip. `downloadTimer` is only started in `downloadNextItem`, and `AbstractClient` never fails outstanding `PendingCommand`s when the socket drops. So if the connection dies between this send and the response, the intent never emits `finished`, `failed` or `cancelled`.

`IntentUrlParser::chainRunning` then stays true forever, and every later `cockatrice://` link is queued and silently never runs for the rest of the session.

Start `downloadTimer` here as well, or a dedicated list timer. It is also worth having `IntentUrlParser` connect to `finalIntent`'s `destroyed` signal to clear `chainRunning` as a backstop, so no future intent that dies without a terminal signal can wedge the whole chain.

cockatrice/src/interface/intents/intent_login.cpp
```diff
@@ -29,5 +32,27 @@ void IntentGetLoginCredentials::onPreconditionSatisfied()
...
+        emitCancelled();
+        return;
+    }
+
+    context->username = dialog.username();
+    context->password = dialog.password();
+
+    if (dialog.savePassword() && !context->username.isEmpty()) {
+        ServersSettings &servers = SettingsCache::instance().servers();
+        servers.addNewServer(context->hostname, context->hostname, context->port, context->username, context->password,
```
ZeldaZach, line 53: `addNewServer` delegates to `updateExistingServer`, which matches on the host string alone (`servers_settings.cpp:272`) and then unconditionally overwrites that entry's port on the next line, along with saveName, username and password.

`checkPrecondition` gates on `hasLoginData(host, port)` with the exact port, so a link like `cockatrice://opendeck?hostname=<a host the user already saved>&port=1337&...` passes the gate and reaches this dialog. If the user then ticks "save password", their real saved server entry is rewritten to point at port 1337 with the new credentials.

Either look up `findServerIndex(hostname, port)` first and only add when it returns -1, or fix `updateExistingServer` to match on host and port together.

cockatrice/src/single_instance_manager.cpp
```diff
@@ -72,7 +80,14 @@ bool SingleInstanceManager::forwardToPrimary(const QStringList &filesToSend)
     socket.flush();
     socket.waitForBytesWritten(1000);
 
-    return true;
+    // Only report a successful hand-off once the primary has acknowledged that
+    // it actually read the payload. A socket that connects but never answers
+    // belongs to a process that is dying, so the caller must not treat this as
+    // a hand-off (otherwise it would exit without anyone handling the files).
+    if (!socket.waitForReadyRead(1000)) {
```
ZeldaZach, line 87: Treating "connected but no ACK within 1s" as "no primary" is unsafe given what `tryRun` does next: `listen()` fails, `forwardToPrimary` is retried, and then `QLocalServer::removeServer(serverName)` deletes the socket of a primary that is alive but merely busy. Card database load, an unrelated modal dialog, or a briefly stalled main thread all produce that.

Distinguish the two cases, for example return an enum from `forwardToPrimary` (NoPrimary / PrimaryBusy / Delivered) and only fall through to `removeServer` when `connectToServer` itself failed. On PrimaryBusy, either wait longer or give up rather than stealing the name.

cockatrice/src/single_instance_manager.cpp
```diff
@@ -111,6 +126,14 @@ void SingleInstanceManager::handleNewConnection()
             QStringList files;
             payloadStream >> files;
 
+            // Acknowledge receipt as soon as the payload is parsed, before the
+            // primary starts handling it. The handlers run synchronously and can
+            // take longer than the sender's readiness timeout (e.g. a modal
+            // confirmation box), which would otherwise make a live primary look
+            // dead and cause duplicate handling.
+            socket->write(ACK_MESSAGE);
+            socket->flush();
```
ZeldaZach, line 135: `buffer` and `expectedSize` are reset only after `emit filesReceived(files)` returns, but this PR makes that emit run modal dialogs (the join confirm, the migration confirm, `DlgLoginPrompt`) which spin a nested event loop. During that loop `readyRead` can re-enter this lambda with the old payload still in the buffer and `expectedSize` still non-zero, so the same URL gets dispatched twice.

Move `buffer->clear(); *expectedSize = 0;` above the emit, or dispatch via `QMetaObject::invokeMethod(this, [=]{ emit filesReceived(files); }, Qt::QueuedConnection)` so the socket handler never runs user code synchronously.

cockatrice/src/interface/intents/url_parser.cpp
```diff
@@ -134,3 +168,221 @@ QString IntentUrlParser::generateJoinGameMessage(const ContextJoinGame &context,
...
+            tr("Opening this share link connects you to %1 instead of %2.\n\nContinue?").arg(target, current),
+            QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+        if (answer != QMessageBox::Yes) {
+            return nullptr;
+        }
+        migrationTargetHost = ctx->serverContext.hostname;
+        migrationTargetPort = ctx->serverContext.port;
+        previousServerHost = client->serverName();
+        previousServerPort = QString::number(client->serverPort());
+        pendingRestore = true;
```
ZeldaZach, line 226: `pendingRestore`, `migrationTargetHost` and `previousServerHost` are parser-wide, but they are written at parse time while chains are still queued, and consumed in `chainEnded()` by whichever chain happens to finish first.

Two links handled back to back means chain A's failure consumes the restore state that chain B recorded, B's own restore never happens, and `previousServerHost` was captured before A moved the session anyway. The deferred waiter in `restorePreviousServer()` captures nothing but `this`, so a second migration overwrites the members it is going to read.

Store these fields in a per-chain struct alongside the intent list in `pendingChains`, and snapshot the current server when the chain starts rather than when it is built.

cockatrice/src/main.cpp
```diff
@@ -325,10 +327,22 @@ int main(int argc, char *argv[])
...
+    const bool hasUrlActivation = std::any_of(startupFiles.begin(), startupFiles.end(), [](const QString &file) {
+        return file.startsWith(QStringLiteral("cockatrice://"));
+    });
+    ui.setSkipStartupAutoConnect(hasUrlActivation);
+
     auto handleActivation = [&ui](const QString &file) {
         if (file.startsWith("cockatrice://")) {
-            auto urlParser = new IntentUrlParser(&ui, &ui);
-            urlParser->handle(file);
+            qInfo() << "Handling URL activation:" << file;
```
ZeldaZach, line 341: This writes the full URL, including the `share` token, at info level, and Logger can persist output to a file that users routinely paste into bug reports. The share token is a bearer capability for the deck. The parser already goes out of its way to log only `shareToken.length()`, so this undoes that care.

Log the scheme and action only, or redact the query. Line 276 has the same problem, printing the whole positional argument list. Both would also be better as a `qCInfo` category to match the rest of the file rather than bare `qInfo`.

cockatrice/src/interface/widgets/general/layout_containers/flow_widget.cpp
```diff
@@ -177,6 +178,50 @@ QLayoutItem *FlowWidget::itemAt(int index) const
     return flowLayout->itemAt(index);
 }
 
+void FlowWidget::keyPressEvent(QKeyEvent *event)
+{
+    // Keyboard navigation between the flow items: arrow keys move focus just
+    // like clicking the sibling tiles would. Only items that can take keyboard
+    // focus (e.g. the deck-preview tiles in shared-deck links) are visited.
+    const bool moveForward = event->key() == Qt::Key_Right || event->key() == Qt::Key_Down;
```
ZeldaZach, line 186: `FlowWidget` is not focusable, so this only runs via key propagation from a focused tile, and that chain is tile, container, viewport, QScrollArea, FlowWidget. `QAbstractScrollArea::keyPressEvent` handles Up/Down/Left/Right by scrolling and calls `e->accept()`, so the arrows stop at the scroll area and never arrive here.

`DlgSharedDecksPreview` constructs `FlowWidget` with `ScrollBarAsNeeded` vertically, i.e. with a scroll area, so this navigation is dead in its only caller.

Handle the arrows in `SharedDeckPreviewWidget::keyPressEvent` calling into a `FlowWidget` helper, or install an event filter on the tiles from `FlowWidget` so the key is intercepted before it propagates.

cockatrice/src/main.cpp
```diff
@@ -325,10 +327,22 @@ int main(int argc, char *argv[])
...
     MainWindow ui;
 
+    // A URL launch must own the connection: the intent chain triggered by the
+    // URL connects to the server named in the URL, so the window's own startup
+    // auto-connect must not race against it (two connectToServer calls tear
+    // each other down via doDisconnectFromServer).
+    const bool hasUrlActivation = std::any_of(startupFiles.begin(), startupFiles.end(), [](const QString &file) {
+        return file.startsWith(QStringLiteral("cockatrice://"));
+    });
+    ui.setSkipStartupAutoConnect(hasUrlActivation);
```
ZeldaZach, line 337: `hasUrlActivation` is computed from positional arguments only, but on macOS a `cockatrice://` open arrives through `UrlSchemeEventFilter` into `pendingMacUrls`. So the one platform where URL launches are most common still races the startup auto-connect this flag exists to prevent. Fold `pendingMacUrls` into the check under the `Q_OS_MAC` guard.

Separately, `skipStartupAutoConnect` only gates `attemptStartupAutoConnect`. `MainWindow::applyStartupDestination` is a second connect path, fired by `QTimer::singleShot(0)` from the constructor, and it is not gated at all, so a user whose startup tab is Server or Server Room gets the same double `connectToServer` teardown.

cockatrice/src/interface/intents/intent_login.cpp
```diff
@@ -29,5 +32,27 @@ void IntentGetLoginCredentials::onPreconditionSatisfied()
 
 void IntentGetLoginCredentials::onPreconditionNotSatisfied()
 {
-    emitFailed(tr("No saved credentials for this server"));
+    // No credentials saved for the target server: ask the user for them. They
+    // opt into saving them so later links to the same server connect directly.
+    const QString serverText = context->hostname + ":" + context->port;
+    DlgLoginPrompt dialog(serverText);
```
ZeldaZach, line 38: `IntentGetLoginCredentials` is shared with `MainWindow::applyStartupDestination`, which runs on every launch when the startup tab is Server or Server Room. Replacing the previous silent `emitFailed("No saved credentials for this server")` with a blocking application-modal dialog means those users now get an unexpected sign-in popup at startup where they previously got a quiet failure.

Gate the prompt behind a constructor flag defaulting to the old behaviour, and set it only from the URL parser chains.

cockatrice/src/interface/intents/intent.cpp
```diff
@@ -2,10 +2,11 @@
 
 Intent::Intent(QObject *parent) : QObject(parent)
 {
-    // An intent is done as soon as it reports success or failure. Deleting it
-    // also tears down its dependency chain and disconnects any signal wiring.
+    // An intent is done as soon as it reports success, failure, or cancellation.
+    // Deleting it also tears down its dependency chain and disconnects any signal wiring.
     connect(this, &Intent::finished, this, &QObject::deleteLater);
     connect(this, &Intent::failed, this, &QObject::deleteLater);
+    connect(this, &Intent::cancelled, this, &QObject::deleteLater);
```
ZeldaZach, line 9: The new `cancelled` outcome is not handled by `runDependency`, which only forwards `finished` and `failed`. If any intent is run as a dependency and cancels, the parent is never told, never emits a terminal signal, and the URL chain hangs with `chainRunning` stuck true.

`IntentGetLoginCredentials` now emits `cancelled` and is one refactor away from being used as a dependency, so this is worth closing now:

```cpp
connect(dependency, &Intent::cancelled, this, &Intent::cancelled);
```

cockatrice/src/interface/widgets/dialogs/dlg_shared_decks_preview.cpp
```diff
@@ -0,0 +1,178 @@
...
+        resultEmitted = true;
+        emit cancelled();
+    }
+}
+
+void DlgSharedDecksPreview::closeEvent(QCloseEvent *event)
+{
+    onCancel();
+    QDialog::closeEvent(event);
+}
```
ZeldaZach, line 178: Five of the new files end without a trailing newline: `dlg_shared_decks_preview.cpp`, `dlg_login_prompt.cpp/.h`, `shared_deck_preview_widget.cpp` and `intent_open_shared_deck.cpp/.h`. clang-format in CI will flag these.

ZeldaZach (changes requested):
This is a well-structured delta and the core design decision is the right one. Serializing link handling into a queued intent chain in `IntentUrlParser` is a much better shape than ad hoc handling at each entry point, the `cancelled` outcome added to `Intent` closes a real gap left by the earlier PRs, the share token is deliberately logged by length rather than value in the parser, the preview dialog guards its terminal signal through both `closeEvent` and `QDialog::rejected`, and the migration confirm plus session restore shows care about not stranding a logged-in user. The `skipStartupAutoConnect` and `urlChainFinished` fallback addresses a genuine startup race.

Three things need to change before this merges.

The first is the security one. The confirmation before connecting to a link's host is gated on `migrating`, which requires being already logged in elsewhere. That means the most common path by far, clicking a link in a browser with the app not connected, connects to whatever host and port the URL names with no confirmation and no check against the configured server list. `createJoinGameIntent` already asks in all cases; this should match, and a host with no entry in the saved server list deserves a louder warning than a routine connect prompt rather than a sign-in dialog for a host the attacker chose.

The second compounds it. `IntentGetLoginCredentials` calls `addNewServer`, which delegates to `updateExistingServer`, and that matches on the host string alone before unconditionally overwriting the entry's port. Since `checkPrecondition` gates on host plus exact port, a link naming a host the user already has saved but a different port reaches the login dialog, and ticking "save password" rewrites their real saved server to the attacker's port with fresh credentials.

The third is a liveness bug. The `DeckShareList` round trip in `IntentOpenSharedDeck` has no timeout, and `AbstractClient` never fails outstanding commands when the socket drops, so a connection lost at the wrong moment means the intent emits no terminal signal at all. `chainRunning` then stays true and every subsequent link is queued and silently dropped for the rest of the session. A `destroyed` backstop on the final intent would make that class of bug non-fatal in general.

Below that: the ACK handshake in `SingleInstanceManager` treats a slow-but-alive primary as dead and then removes its socket, the readyRead handler can re-dispatch the same URL because the buffer is cleared after an emit that now runs modal dialogs, the migration bookkeeping lives on the parser rather than per-chain so two queued links cross-contaminate it, and `main.cpp` logs the full URL including the share token at info level into a file users paste into bug reports.

One functional note on the accessibility work: the arrow-key navigation added to `FlowWidget` is unreachable in its only caller, since `QAbstractScrollArea` accepts the arrow keys for scrolling before they can propagate that far.

cockatrice/src/interface/intents/url_parser.cpp
BruebachL: Fixed in 72dc4286: opening a deck now asks whenever the client is not already connected to the link's host, naming the host in the prompt. The migration wording (and session restore) only applies when logged in elsewhere; on a fresh connect, a host with no entry in the configured server list gets the harder warning — "a server you have never connected to before", defaulting to No (host lookup helper landed with b821f6f9).

cockatrice/src/interface/intents/intent_open_shared_deck.cpp
BruebachL: Fixed in 7e16e412: downloadTimer now also times the DeckShareList round trip (list phase); a silent server fails the intent instead of wedging the chain. IntentUrlParser also connects to the final intent's destroyed signal as a backstop and disconnects it once the chain ends, so no intent that dies without a terminal signal can block later links.

cockatrice/src/interface/intents/intent_login.cpp
BruebachL: Fixed in b821f6f9: addNewServer now matches the exact host+port with findServerIndex first and updates that entry in place; it only appends a new entry when no such server exists, so a link can no longer rewrite a saved entry's port.

cockatrice/src/single_instance_manager.cpp
BruebachL: Fixed in c2c26c33: forwardToPrimary now returns ForwardResult {Delivered, NoPrimary, PrimaryBusy}. tryRun only falls through to removeServer on NoPrimary; a busy primary gets a longer acknowledgement window and its socket is never removed.

cockatrice/src/single_instance_manager.cpp
BruebachL: Fixed in b9355cbd: buffer and expectedSize are reset before emit filesReceived, so readyRead re-entrancy during a modal handler can no longer re-dispatch the same payload.

cockatrice/src/interface/intents/url_parser.cpp
BruebachL: Fixed in aa9fbb00: pendingRestore, migrationTarget* and previousServer* now live in a PendingIntentChain record stored with each queued chain. The previous server is snapshotted when the chain actually starts running — not at parse time — and the deferred restore waiter uses that snapshot.

cockatrice/src/main.cpp
BruebachL: Fixed in 208642d6: activation URLs are logged redacted to scheme+action only (e.g. cockatrice://opendeck), the positional file list is redacted the same way, and both now use qCInfo(MainLog).

cockatrice/src/interface/widgets/general/layout_containers/flow_widget.cpp
BruebachL: Fixed in 143d23a5: FlowWidget installs an event filter on each added item, so arrow keys are intercepted on the focused tile before the QAbstractScrollArea can consume them and are routed into the existing focus navigation.

cockatrice/src/main.cpp
BruebachL: Fixed in 25a7c8c2: pendingMacUrls are folded into hasUrlActivation under Q_OS_MAC, applyStartupDestination is gated on skipStartupAutoConnect, and onUrlChainFinished now retries the startup destination (not plain auto-connect) when the startup tab is a Server / Server Room.

cockatrice/src/interface/intents/intent_login.cpp
BruebachL: Fixed in 6397bd74: IntentGetLoginCredentials now takes a promptForMissingCredentials flag defaulting to false, so applyStartupDestination keeps its previous silent failure; only the URL parser chains opt into the dialog.

cockatrice/src/interface/intents/intent.cpp
BruebachL: Fixed in 79221a7f: runDependency now connects the dependency's cancelled signal to the owner's cancelled, so a declined prompt cancels the whole chain.

cockatrice/src/interface/widgets/dialogs/dlg_shared_decks_preview.cpp
BruebachL: Fixed in aee4ffae: dlg_shared_decks_preview.cpp, dlg_login_prompt.cpp/.h, shared_deck_preview_widget.cpp and intent_open_shared_deck.cpp/.h all end with a trailing newline.

cockatrice/src/interface/widgets/general/layout_containers/flow_widget.cpp
```diff
@@ -80,10 +81,15 @@ FlowWidget::FlowWidget(QWidget *parent,
...
+ * The widget is filtered for arrow-key events so keyboard navigation between
+ * the flow items keeps working even when the flow sits inside a QScrollArea,
+ * which swallows arrow keys before they can reach FlowWidget::keyPressEvent.
+ *
  * @param widget_to_add The widget to add to the flow layout.
  */
-void FlowWidget::addWidget(QWidget *widget_to_add) const
+void FlowWidget::addWidget(QWidget *widget_to_add)
 {
+    widget_to_add->installEventFilter(this);
```
ZeldaZach, line 109: Installing this filter on *every* widget added to *every* `FlowWidget` steals arrow keys from existing input controls.

`DrawProbabilityWidget` adds a `QComboBox` (criteria), a second `QComboBox` (exactness) and two `QSpinBox`es (quantity, drawn) via `controls->addWidget(...)`; `DeckAnalyticsWidget::controlContainer` does the same. All of these have `Qt::StrongFocus`, so they land in `focusableItems`, and `eventFilter` runs before the widget's own key handler and returns `true`. After this change, focusing the quantity spin box and pressing Up/Down moves focus to the next control instead of incrementing the value, and Left/Right in a focused combo box moves focus instead of changing the selection.

Suggest making the navigation opt-in (a separate `addNavigableWidget()`, or a flag on `FlowWidget`), or at minimum skipping widgets that consume arrow keys themselves (`QAbstractSpinBox`, `QComboBox`, `QLineEdit`, `QAbstractSlider`, `QAbstractItemView`).

cockatrice/src/interface/intents/url_parser.cpp
```diff
@@ -134,3 +169,264 @@ QString IntentUrlParser::generateJoinGameMessage(const ContextJoinGame &context,
...
+        connect(getLoginCredentialsIntent, &Intent::cancelled, openDeckIntent, &Intent::cancelled);
+        firstIntent = getLoginCredentialsIntent;
+    }
+
+    return firstIntent;
+}
+
+bool IntentUrlParser::isConnectedTo(const QString &hostname, const QString &port) const
+{
+    Q_UNUSED(port);
```
ZeldaZach: The comment above says ignoring the port means "a link to the same host on another port still connects rather than silently reusing an existing session on a different server", but the code does the opposite.

Concretely: the user is logged in to `example.com:4747` and opens `cockatrice://opendeck?hostname=example.com&port=4748&share=...`. `isConnectedTo()` returns true (hostname matches, port ignored), so:
- the "this connects you to X instead of Y" confirmation is skipped,
- no `IntentGetLoginCredentials` is prepended,
- `IntentOpenSharedDeck::checkPrecondition()` also compares hostname only and returns true,

so `Command_DeckShareList` is sent to the :4747 session and the share token is resolved against the wrong server (it will simply not be found, or — worse — the token is handed to a server the link was not addressed to). Same for `joingame` links, where the game id would be looked up on the wrong instance.

`RemoteClient::serverPort()` exists (it is used a few lines below in the migration prompt), so the port comparison that was dropped can be restored here and in the two intent preconditions.

cockatrice/src/interface/intents/intent_join_server_game.cpp
```diff
@@ -19,13 +19,10 @@ bool IntentJoinServerGame::checkPrecondition() const
...
-    // configured server port (e.g. when connecting through a proxy), so only
-    // the hostname is compared here.
-    if (remoteClient->peerName() != context->roomContext.serverContext.hostname) {
-        return false;
-    }
-    if (QString::number(remoteClient->peerPort()) != context->roomContext.serverContext.port) {
+    // serverName() reflects the server the client was configured to connect to,
+    // which may differ from the actual TCP peer (e.g. when connecting through a
+    // proxy), so only the hostname is compared here.
+    if (remoteClient->serverName().compare(context->roomContext.serverContext.hostname, Qt::CaseInsensitive) != 0) {
```
ZeldaZach, line 27: Switching from `peerName()`/`peerPort()` to `serverName()` is right (the configured host is what the link names), but the port check was dropped entirely instead of being switched to `serverPort()`.

`AbstractClient::serverPort()` is the configured port, so it is not affected by the proxy concern the old comment described and can still be compared. As written, a `joingame` link for `example.com:4748` is treated as "already on the right server" while the user is logged in to `example.com:4747`, and the intent joins game id N on the wrong instance — test and live servers sharing one hostname is the common case here.

Adding back `if (QString::number(remoteClient->serverPort()) != context->roomContext.serverContext.port) return false;` restores the guard without reintroducing the proxy problem.

cockatrice/src/interface/intents/intent_login.cpp
```diff
@@ -29,5 +34,35 @@ void IntentGetLoginCredentials::onPreconditionSatisfied()
...
+        emitCancelled();
+        return;
+    }
+
+    context->username = dialog.username();
+    context->password = dialog.password();
+
+    if (dialog.savePassword() && !context->username.isEmpty()) {
+        ServersSettings &servers = SettingsCache::instance().servers();
+        servers.addNewServer(context->hostname, context->hostname, context->port, context->username, context->password,
```
ZeldaZach: Passing `context->hostname` as the `saveName` silently renames an existing server profile.

This branch runs whenever `hasLoginData()` is false, which includes a server entry that already exists with no saved username — e.g. an entry added by the public-server list with `saveName` "Cockatrice Woogerworks". `addNewServer()` now matches host+port exactly, finds that entry, and `updateServerFields()` overwrites `saveName%i` with the raw hostname, so the user's named profile turns into `server.woogerworks.com` in the connect dialog.

Reusing the existing entry's `saveName` when `findServerIndex()` already matches (or only setting the name when creating a new entry) avoids this.

cockatrice/src/interface/window_main.cpp
```diff
@@ -916,6 +913,50 @@ void MainWindow::handleCockatriceLink(const QString &url)
...
+    }
+}
+
+void MainWindow::onUrlChainFinished(bool connected)
+{
+    // A cockatrice:// link owns the startup connection while it runs. When its
+    // chain ended without connecting (declined, invalid, offline), fall back to
+    // the startup connection so the activation launch still behaves like a
+    // normal launch.
+    if (connected || !skipStartupAutoConnect || getRemoteClient()->getStatus() != StatusDisconnected) {
```
ZeldaZach: `skipStartupAutoConnect` is never cleared on the success path, so this startup fallback can fire long after startup.

Sequence: the app is launched from a `cockatrice://` link, the chain succeeds (`connected == true`, early return, flag stays `true`, and `startupAutoConnectAttempted` is still `false` because `attemptStartupAutoConnect()` bailed out on the flag). Later the user disconnects, then clicks a second share link (handed over by `SingleInstanceManager`) and declines the confirmation. `handle()` emits `urlChainFinished(false)`, the guard passes (flag still true, status Disconnected), and the client silently auto-connects to the auto-connect server — or re-runs `applyStartupDestination()` — in the middle of a running session.

Clearing `skipStartupAutoConnect` once the first chain ends (or gating this on "still in startup") keeps the fallback to the launch it was meant for.

cockatrice/src/single_instance_manager.cpp
```diff
@@ -72,7 +92,15 @@ bool SingleInstanceManager::forwardToPrimary(const QStringList &filesToSend)
     socket.flush();
     socket.waitForBytesWritten(1000);
 
-    return true;
+    // Only report a successful hand-off once the primary has acknowledged that
+    // it actually read the payload. A socket that connects but is still working
+    // on an earlier payload is alive but busy, not dead: give it more room
+    // before giving up, so a slow handler does not make a live primary look
+    // dead (which would lead to stealing its socket).
+    if (!socket.waitForReadyRead(1000) && !socket.waitForReadyRead(4000)) {
```
ZeldaZach, line 108: This blocking ACK wait is also on the plain-launch path, where it can stall startup by 5 seconds.

`main()` calls `instance.tryRun(QStringList())` for every launch without positional arguments, and that goes through `forwardToPrimary()` too. If a primary exists but is inside a nested event loop (any modal dialog — the card-database update prompt, a share-link confirmation, the connect dialog), it will not service `readyRead` until the user dismisses it, so the new instance blocks here for 1s + 4s before `MainWindow` is even constructed. Previously the plain launch returned as soon as the bytes were written.

The ACK requirement matters for the activation path (where stealing the socket is the failure being fixed); for `filesToSend.isEmpty()` it buys nothing and could return `NoPrimary`/`Delivered` without waiting.

cockatrice/src/interface/intents/url_parser.cpp
```diff
@@ -134,3 +169,264 @@ QString IntentUrlParser::generateJoinGameMessage(const ContextJoinGame &context,
...
+            // Fresh connection is harmless to wander away from, but a server the
+            // client has never been configured for deserves a harder warning (no
+            // by default) so a stray link cannot silently steer the client there.
+            const bool knownHost = SettingsCache::instance().servers().findHostIndex(ctx->serverContext.hostname) >= 0;
+            const QMessageBox::StandardButton answer =
+                knownHost
+                    ? QMessageBox::question(mainWindow, tr("Open shared deck"),
+                                            tr("Opening this share link connects you to %1.\n\nContinue?").arg(target))
+                    : QMessageBox::warning(mainWindow, tr("Open shared deck"),
+                                           tr("Opening this share link connects you to %1, a server you have "
```
ZeldaZach, line 242: Non-blocking, but worth hardening since this dialog *is* the security gate: `hostname` comes straight out of the link's query string (percent-decoded by `QUrlQuery`), and `QMessageBox` uses `Qt::AutoText`, so a hostname containing markup flips the whole message to rich text.

`cockatrice://opendeck?hostname=%3Cbr%3E%3Cbr%3EVerified%20official%20server&port=4747&share=...` renders as formatted HTML and lets the link author pad or restyle the text the user is being asked to approve.

The bite is limited today because a hostname carrying `<`/`>` will not resolve, so the connect fails afterwards - but the spoofed prompt is shown first, and the same applies to the server-supplied share name and deck names rendered in `DlgSharedDecksPreview`'s `QLabel`s once the chain reaches an attacker-run server.

Cheapest fix is `.toHtmlEscaped()` on `target` here (and on `serverText`/`shareName` in the preview dialog), or building the boxes explicitly with `setTextFormat(Qt::PlainText)`.

ZeldaZach (approved):
Re-reviewed against the previous round. All twelve items from the earlier CHANGES_REQUESTED are genuinely addressed in the code, not just in the replies:

- opendeck now confirms whenever the client is not already connected to the link's host, and an unconfigured host gets the harder warning defaulting to No
- the `DeckShareList` round trip is timed, and `IntentUrlParser` has a `destroyed` backstop so a dead intent cannot wedge the chain queue for the session
- `addNewServer` matches host+port via `findServerIndex` before touching an entry
- `forwardToPrimary` returns `ForwardResult`, and `removeServer` only runs on `NoPrimary`
- the single-instance buffer is cleared before `emit filesReceived`, closing the modal re-entrancy double-dispatch
- restore/migration state moved into per-chain `PendingIntentChain` records, snapshotted when the chain starts
- activation URLs are logged redacted to scheme+action
- arrow navigation now goes through an event filter, `pendingMacUrls` feeds `hasUrlActivation`, `applyStartupDestination` is gated
- `IntentGetLoginCredentials` prompts only when the caller opts in, `runDependency` forwards `cancelled`, trailing newlines fixed

I checked the parts this PR is riskiest in. The gate cannot be bypassed: no network fetch happens before a confirm on an unconnected host, and the download itself waits on the preview dialog. No file is written from the share - `loadFromRemote` goes into a deck editor tab, so there is no server-supplied filename and no traversal. Reply lifetimes and the timer paths look right.

Approving. Nothing blocking. Seven non-blocking findings are inline; two are worth a follow-up before this ships rather than after:

1. `FlowWidget::addWidget` installs the arrow-key filter on every widget in every `FlowWidget`, including the `QComboBox`/`QSpinBox` controls `DrawProbabilityWidget` adds. Up/Down on those spin boxes now moves focus instead of changing the value. An opt-in flag, or skipping widgets that consume arrows, keeps it contained.
2. `isConnectedTo()` deliberately ignores the port, but its comment states the opposite of what the code does, and `IntentJoinServerGame` dropped its port check in the same move. Same hostname on two ports is an unusual setup, so this is not urgent - the comment should at least match the code.

The rest (saveName overwrite in `addNewServer`, `skipStartupAutoConnect` never cleared on success, the 5s ACK wait on the plain-launch path, and rich-text injection from the link hostname into the confirm dialog) are inline and fine as follow-ups.

cockatrice/src/interface/widgets/general/layout_containers/flow_widget.cpp
BruebachL: Fixed in 20079dc: FlowWidget::addWidget no longer installs the arrow-key filter; navigation between items is opt-in via the new addNavigableWidget, which only the shared-deck preview tiles use. Spin/combo controls on DrawProbability and DeckAnalytics keep their arrow keys.

cockatrice/src/interface/intents/url_parser.cpp
BruebachL: Fixed in 20079dc: isConnectedTo now compares the configured server port as well as the host (via serverPort(), proxy-safe), and both intent preconditions follow suit — a link to the same host on another port no longer reuses or resolves against the wrong session. The comment now matches the code.

cockatrice/src/interface/intents/intent_join_server_game.cpp
BruebachL: Fixed in 20079dc: IntentJoinServerGame::checkPrecondition compares remoteClient->serverPort() against the link's port again, alongside serverName(), so a joingame link for example.com:4748 is not treated as already connected to a :4747 session.

cockatrice/src/interface/intents/intent_login.cpp
BruebachL: Fixed in 20079dc: onPreconditionNotSatisfied looks up findServerIndex first and reuses the existing entry's saveName when it matches, so a link filling in credentials no longer renames a profile like 'Cockatrice Woogerworks' to the raw hostname.

cockatrice/src/interface/window_main.cpp
BruebachL: Fixed in 20079dc: onUrlChainFinished clears skipStartupAutoConnect on the connected path, so the startup fallback only fires for the launch it was meant for and a later mid-session declined link cannot silently auto-connect or re-run applyStartupDestination.

cockatrice/src/single_instance_manager.cpp
BruebachL: Fixed in 20079dc: forwardToPrimary skips the ACK wait when filesToSend is empty (the plain-launch path), so a primary busy in a modal dialog no longer stalls a new instance for up to 5s. The activation path still waits for the ACK before its socket name is stolen.

cockatrice/src/interface/intents/url_parser.cpp
BruebachL: Fixed in 20079dc: the link-hostname target interpolated into both confirm prompts is toHtmlEscaped(), and the server-supplied shareName/serverText/gameFormat in the shared-deck preview dialog are escaped too, so percent-decoded markup can no longer flip the AutoText messages to rich text.

PR #7245, opened 2026-09-05
[DeckShare] Browse and open public decks with loading, error and accessibility states

### LINKED ISSUE #7224: Temporary deck share links and public remote decks
**Similar Requests**
- #2555
- #4364

Add a way to share decks directly through the server, without third-party tools:

1. Temporary share links. 
- From the deck editor or the remote deck tree, a user can bundle any mix of server-side decks, a server-side folder, or local deck files into a named share. 
- The server creates an unguessable token with an expiry date.
- The resulting link is copied to the clipboard (e.g. cockatrice://opendeck?share=...&hostname=...&port=...). 
- Clicking the link in another client downloads and opens the decks in the deck editor.

2. Permanent public decks. 
- Users can mark individual decks or whole folders as public. 
- Other users browse these in a new "public decks" tab rendered from lightweight metadata (banner card, color identity) so no deck lists are downloaded until a deck is opened.

**Additional Context**
Complements the long-standing request in #2555 (shareable deck URLs) and #4364 (public remote decks/admin-uploadable starter decks). 
Shares should expire and be cleaned up server-side. 
Public visibility should never expose data by default (existing decks stay private).

BruebachL:
## Related tickets
- Fixes #7224

Public decks have no browsing surface: a user could publish a deck, but nobody
had a place to see it, and opening another user's public deck failed silently
if the fetch errored. Deck uploads also carried no metadata, so a browser would
have to download every deck list just to render previews.

- Add a `Public decks` tab that lists another user's public decks through
  `Command_DeckListOtherUser`, with a user-context menu entry that opens it and
  auto-selects that user
- Render each deck as a preview tile from the stored metadata (banner card,
  color identity, tags, upload time) without downloading the deck list, with the
  deck name as the tile's accessible name and Space/Enter or double-click
  opening the deck
- Add a `Show Upload Time` setting so the tile's upload stamp can be hidden like
  the other preview details, under the same settings group as the local storage
  previews
- Publish and unpublish decks from the storage toolbar and context menu, toggling
  the deck's own visibility bit (what the server persists) rather than the
  inherited effective state, and batching the visibility refresh until the last
  in-flight change is acknowledged
- Show a loading indicator and a server-error message instead of a blank tab when
  the deck list fetch fails or the connection drops, and a message box when
  opening a public deck fails or arrives corrupted
- Wire the new tab into the supervisor's `retranslateUi` loop and rename the
  share action tooltip from `Deck share` to `Share link`

### DIFF
--- a/cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.cpp
+++ b/cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.cpp
@@ -113,7 +113,7 @@ int RemoteDeckList_TreeModel::rowCount(const QModelIndex &parent) const
 
 int RemoteDeckList_TreeModel::columnCount(const QModelIndex & /*parent*/) const
 {
-    return 3;
+    return 4;
 }
 
 QVariant RemoteDeckList_TreeModel::data(const QModelIndex &index, int role) const
@@ -121,7 +121,7 @@ QVariant RemoteDeckList_TreeModel::data(const QModelIndex &index, int role) cons
     if (!index.isValid()) {
         return QVariant();
     }
-    if (index.column() >= 3) {
+    if (index.column() >= 4) {
         return QVariant();
     }
 
@@ -134,12 +134,20 @@ QVariant RemoteDeckList_TreeModel::data(const QModelIndex &index, int role) cons
                 switch (index.column()) {
                     case 0:
                         return node->getName();
+                    case 3:
+                        return isEffectivelyPublic(node) ? tr("Public") : tr("Private");
                     default:
                         return QVariant();
                 }
             }
             case Qt::DecorationRole:
                 return index.column() == 0 ? dirIcon : QVariant();
+            case Qt::ToolTipRole:
+                if (index.column() == 3) {
+                    return isEffectivelyPublic(node) ? tr("This folder is visible to other users")
+                                                     : tr("This folder is only visible to you");
+                }
+                return QVariant();
             default:
                 return QVariant();
         }
@@ -153,6 +161,8 @@ QVariant RemoteDeckList_TreeModel::data(const QModelIndex &index, int role) cons
                         return file->getId();
                     case 2:
                         return file->getUploadTime();
+                    case 3:
+                        return isEffectivelyPublic(file) ? tr("Public") : tr("Private");
                     default:
                         return QVariant();
                 }
@@ -161,6 +171,12 @@ QVariant RemoteDeckList_TreeModel::data(const QModelIndex &index, int role) cons
                 return index.column() == 0 ? fileIcon : QVariant();
             case Qt::TextAlignmentRole:
                 return index.column() == 1 ? Qt::AlignRight : Qt::AlignLeft;
+            case Qt::ToolTipRole:
+                if (index.column() == 3) {
+                    return isEffectivelyPublic(file) ? tr("This deck is visible to other users")
+                                                     : tr("This deck is only visible to you");
+                }
+                return QVariant();
             default:
                 return QVariant();
         }
@@ -183,6 +199,8 @@ QVariant RemoteDeckList_TreeModel::headerData(int section, Qt::Orientation orien
                     return tr("ID");
                 case 2:
                     return tr("Upload time");
+                case 3:
+                    return tr("Visibility");
                 default:
                     return QVariant();
             }
@@ -239,13 +257,14 @@ void RemoteDeckList_TreeModel::addFileToTree(const ServerInfo_DeckStorage_TreeIt
     time.setSecsSinceEpoch(fileInfo.creation_time());
 
     beginInsertRows(nodeToIndex(parent), parent->size(), parent->size());
-    parent->append(new FileNode(QString::fromStdString(file.name()), file.id(), time, parent));
+    parent->append(new FileNode(QString::fromStdString(file.name()), file.id(), time, parent, fileInfo.is_public()));
     endInsertRows();
 }
 
 void RemoteDeckList_TreeModel::addFolderToTree(const ServerInfo_DeckStorage_TreeItem &folder, DirectoryNode *parent)
 {
     DirectoryNode *newItem = addNamedFolderToTree(QString::fromStdString(folder.name()), parent);
+    newItem->setIsPublic(folder.folder().is_public());
     const ServerInfo_DeckStorage_Folder &folderInfo = folder.folder();
     const int folderItemsSize = folderInfo.items_size();
     for (int i = 0; i < folderItemsSize; ++i) {
@@ -283,6 +302,21 @@ void RemoteDeckList_TreeModel::refreshTree()
     connect(pend, &PendingCommand::finished, this, &RemoteDeckList_TreeModel::deckListFinished);
 
     client->sendCommand(pend);
+}
+
+bool RemoteDeckList_TreeModel::isEffectivelyPublic(const Node *node) const
+{
+    if (node == nullptr || node == root) {
+        return false;
+    }
+    const Node *current = node;
+    while (current != nullptr) {
+        if (current->isPublic()) {
+            return true;
+        }
+        current = current->getParent();
+    }
+    return false;
 }
 
 void RemoteDeckList_TreeModel::clearTree()
--- a/cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.h
+++ b/cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.h
@@ -27,9 +27,11 @@ public:
     protected:
         DirectoryNode *parent;
         QString name;
+        bool publicFlag;
 
     public:
-        explicit Node(const QString &_name, DirectoryNode *_parent = nullptr) : parent(_parent), name(_name)
+        explicit Node(const QString &_name, DirectoryNode *_parent = nullptr)
+            : parent(_parent), name(_name), publicFlag(false)
         {
         }
         virtual ~Node() = default;
@@ -40,6 +42,14 @@ public:
         [[nodiscard]] QString getName() const
         {
             return name;
+        }
+        [[nodiscard]] bool isPublic() const
+        {
+            return publicFlag;
+        }
+        void setIsPublic(bool _public)
+        {
+            publicFlag = _public;
         }
     };
     class DirectoryNode : public Node, public QList<Node *>
@@ -59,9 +69,14 @@ public:
         QDateTime uploadTime;
 
     public:
-        FileNode(const QString &_name, int _id, const QDateTime &_uploadTime, DirectoryNode *_parent = nullptr)
+        FileNode(const QString &_name,
+                 int _id,
+                 const QDateTime &_uploadTime,
+                 DirectoryNode *_parent = nullptr,
+                 bool _isPublic = false)
             : Node(_name, _parent), id(_id), uploadTime(_uploadTime)
         {
+            setIsPublic(_isPublic);
         }
         [[nodiscard]] int getId() const
         {
@@ -109,6 +124,11 @@ public:
     {
         return root;
     }
+    /**
+     * @brief Whether a node is visible to other users (own flag or inherited
+     * from any ancestor folder).
+     */
+    [[nodiscard]] bool isEffectivelyPublic(const Node *node) const;
     void addFileToTree(const ServerInfo_DeckStorage_TreeItem &file, DirectoryNode *parent);
     void addFolderToTree(const ServerInfo_DeckStorage_TreeItem &folder, DirectoryNode *parent);
     DirectoryNode *addNamedFolderToTree(const QString &name, DirectoryNode *parent);
@@ -125,6 +145,10 @@ private:
 
 public:
     explicit RemoteDeckList_TreeWidget(AbstractClient *_client, QWidget *parent = nullptr);
+    [[nodiscard]] RemoteDeckList_TreeModel *getModel() const
+    {
+        return treeModel;
+    }
     [[nodiscard]] RemoteDeckList_TreeModel::Node *getNode(const QModelIndex &ind) const;
     [[nodiscard]] RemoteDeckList_TreeModel::Node *getCurrentItem() const;
     [[nodiscard]] QList<RemoteDeckList_TreeModel::Node *> getCurrentSelection() const;
--- a/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
+++ b/cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
@@ -324,6 +324,7 @@ bool AbstractTabDeckEditor::actSaveDeck()
         Command_DeckUpload cmd;
         cmd.set_deck_id(static_cast<google::protobuf::uint32>(loadedDeck.lastLoadInfo.remoteDeckId));
         cmd.set_deck_list(deckString.toStdString());
+        cmd.set_tags(loadedDeck.deckList.getTags().join(QStringLiteral(",")).toStdString());
 
         PendingCommand *pend = AbstractClient::prepareSessionCommand(cmd);
         connect(pend, &PendingCommand::finished, this, &AbstractTabDeckEditor::saveDeckRemoteFinished);
--- a/cockatrice/src/interface/widgets/tabs/public_decks_quick_settings_widget.cpp
+++ b/cockatrice/src/interface/widgets/tabs/public_decks_quick_settings_widget.cpp
@@ -1,0 +1,151 @@
+#include "public_decks_quick_settings_widget.h"
+
+#include "../../../client/settings/cache_settings.h"
+#include "../cards/card_size_widget.h"
+
+#include <QCheckBox>
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QSpinBox>
+#include <libcockatrice/settings/cards_display_settings.h>
+#include <libcockatrice/settings/personal_settings.h>
+#include <libcockatrice/settings/visual_deck_storage_settings.h>
+
+PublicDecksQuickSettingsWidget::PublicDecksQuickSettingsWidget(QWidget *parent) : SettingsButtonWidget(parent)
+{
+    // show color identity on preview tiles checkbox
+    showColorIdentityCheckBox = new QCheckBox(this);
+    showColorIdentityCheckBox->setChecked(
+        SettingsCache::instance().visualDeckStorage().getVisualDeckStorageShowColorIdentity());
+    connect(showColorIdentityCheckBox, &QCheckBox::QT_STATE_CHANGED, this,
+            &PublicDecksQuickSettingsWidget::showColorIdentityChanged);
+    connect(showColorIdentityCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().visualDeckStorage(),
+            &VisualDeckStorageSettings::setVisualDeckStorageShowColorIdentity);
+
+    // show tags on preview tiles checkbox
+    showTagsOnDeckPreviewsCheckBox = new QCheckBox(this);
+    showTagsOnDeckPreviewsCheckBox->setChecked(
+        SettingsCache::instance().visualDeckStorage().getVisualDeckStorageShowTagsOnDeckPreviews());
+    connect(showTagsOnDeckPreviewsCheckBox, &QCheckBox::QT_STATE_CHANGED, this,
+            &PublicDecksQuickSettingsWidget::showTagsOnDeckPreviewsChanged);
+    connect(showTagsOnDeckPreviewsCheckBox, &QCheckBox::QT_STATE_CHANGED,
+            &SettingsCache::instance().visualDeckStorage(),
+            &VisualDeckStorageSettings::setVisualDeckStorageShowTagsOnDeckPreviews);
+
+    // show the last modified / upload time on preview tiles checkbox
+    showUploadTimeCheckBox = new QCheckBox(this);
+    showUploadTimeCheckBox->setChecked(
+        SettingsCache::instance().visualDeckStorage().getVisualDeckStorageShowUploadTime());
+    connect(showUploadTimeCheckBox, &QCheckBox::QT_STATE_CHANGED, this,
+            &PublicDecksQuickSettingsWidget::showUploadTimeChanged);
+    connect(showUploadTimeCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().visualDeckStorage(),
+            &VisualDeckStorageSettings::setVisualDeckStorageShowUploadTime);
+
+    // show tag filter box checkbox
+    showTagFilterCheckBox = new QCheckBox(this);
+    showTagFilterCheckBox->setChecked(
+        SettingsCache::instance().visualDeckStorage().getVisualDeckStorageShowTagFilter());
+    connect(showTagFilterCheckBox, &QCheckBox::QT_STATE_CHANGED, this,
+            &PublicDecksQuickSettingsWidget::showTagFilterChanged);
+    connect(showTagFilterCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().visualDeckStorage(),
+            &VisualDeckStorageSettings::setVisualDeckStorageShowTagFilter);
+
+    // draw unused color identities checkbox
+    drawUnusedColorIdentitiesCheckBox = new QCheckBox(this);
+    drawUnusedColorIdentitiesCheckBox->setChecked(
+        SettingsCache::instance().visualDeckStorage().getVisualDeckStorageDrawUnusedColorIdentities());
+    connect(drawUnusedColorIdentitiesCheckBox, &QCheckBox::QT_STATE_CHANGED, this,
+            &PublicDecksQuickSettingsWidget::drawUnusedColorIdentitiesChanged);
+    connect(drawUnusedColorIdentitiesCheckBox, &QCheckBox::QT_STATE_CHANGED,
+            &SettingsCache::instance().visualDeckStorage(),
+            &VisualDeckStorageSettings::setVisualDeckStorageDrawUnusedColorIdentities);
+
+    // unused color identities opacity selector
+    auto unusedColorIdentityOpacityWidget = new QWidget(this);
+
+    unusedColorIdentitiesOpacityLabel = new QLabel(unusedColorIdentityOpacityWidget);
+    unusedColorIdentitiesOpacitySpinBox = new QSpinBox(unusedColorIdentityOpacityWidget);
+
+    unusedColorIdentitiesOpacitySpinBox->setMinimum(0);
+    unusedColorIdentitiesOpacitySpinBox->setMaximum(100);
+    unusedColorIdentitiesOpacitySpinBox->setValue(
+        SettingsCache::instance().visualDeckStorage().getVisualDeckStorageUnusedColorIdentitiesOpacity());
+    connect(unusedColorIdentitiesOpacitySpinBox, qOverload<int>(&QSpinBox::valueChanged), this,
+            &PublicDecksQuickSettingsWidget::unusedColorIdentitiesOpacityChanged);
+    connect(unusedColorIdentitiesOpacitySpinBox, qOverload<int>(&QSpinBox::valueChanged),
+            &SettingsCache::instance().visualDeckStorage(),
+            &VisualDeckStorageSettings::setVisualDeckStorageUnusedColorIdentitiesOpacity);
+
+    unusedColorIdentitiesOpacityLabel->setBuddy(unusedColorIdentitiesOpacitySpinBox);
+
+    auto unusedColorIdentityOpacityLayout = new QHBoxLayout(unusedColorIdentityOpacityWidget);
+    unusedColorIdentityOpacityLayout->setContentsMargins(11, 0, 11, 0);
+    unusedColorIdentityOpacityLayout->addWidget(unusedColorIdentitiesOpacityLabel);
+    unusedColorIdentityOpacityLayout->addWidget(unusedColorIdentitiesOpacitySpinBox);
+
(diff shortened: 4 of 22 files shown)


cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
```diff
@@ -324,6 +324,7 @@ bool AbstractTabDeckEditor::actSaveDeck()
         Command_DeckUpload cmd;
         cmd.set_deck_id(static_cast<google::protobuf::uint32>(loadedDeck.lastLoadInfo.remoteDeckId));
         cmd.set_deck_list(deckString.toStdString());
+        cmd.set_tags(loadedDeck.deckList.getTags().join(QStringLiteral(",")).toStdString());
```
ZeldaZach, line 327: This one looks like a real data-loss bug. The server's `deck_id` branch of `cmdDeckUpload` runs an unconditional update:

```sql
update {prefix}_decklist_files set name=:name, upload_time=NOW(), content=:content,
  banner_card_name=:banner_card_name, banner_card_provider=:banner_card_provider,
  color_identity=:color_identity, tags=:tags where id = :id_deck and id_user = :id_user
```

Any field the client leaves unset is written as empty. Setting only `tags` here means every save of a remote deck from the deck editor wipes the banner card and color identity that `TabDeckStorage::uploadDeck` stored, and the public deck tile this PR adds silently loses its art and color pips.

`tab_deck_storage.cpp:386-389` already sets all four. Mirror it:

```cpp
const CardRef bannerCard = loadedDeck.deckList.getBannerCard();
cmd.set_banner_card_name(bannerCard.name.toStdString());
cmd.set_banner_card_provider(bannerCard.providerId.toStdString());
cmd.set_color_identity(getDeckColorIdentity(loadedDeck.deckList, CardDatabaseManager::query()).toStdString());
```

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
```diff
@@ -795,7 +811,49 @@ void TabDeckStorage::shareFromTreeFinished(const Response &response, const Comma
...
+void TabDeckStorage::setVisibilityFinished(const Response &r, const CommandContainer & /*commandContainer*/)
+{
+    if (r.response_code() != Response::RespOk) {
+        QMessageBox::critical(this, tr("Error"),
+                              tr("Failed to change deck visibility on server (response code %1).")
+                                  .arg(QString::number(static_cast<int>(r.response_code()))));
+    }
+    // Refresh once the last in-flight change has been acknowledged so the
+    // Public/Private column reflects every selected node.
+    if (--pendingVisibilityChanges == 0) {
```
ZeldaZach, line 856: This counter is a one-way latch that can get stuck above zero. `RemoteClient::ping()` sweeps `pendingCommands` and does `i.remove(); pend->deleteLater();` for anything older than `maxTimeout`, without ever calling `processResponse`. A dropped reply therefore leaves `pendingVisibilityChanges` permanently positive, and the Visibility column never refreshes again for any later toggle. Nothing resets it on disconnect either.

Coalescing is more robust than counting here. A restartable single-shot timer that calls `serverDirView->refreshTree()` once, guarded by a bool, costs you one stale refresh when a reply is lost instead of killing the feature for the session.

Also worth moving `++pendingVisibilityChanges` above `client->sendCommand(pend)` so the increment cannot be observed out of order.

cockatrice/src/interface/widgets/visual_deck_storage/remote_public_decks_model.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    beginResetModel();
+    visibleIndices = newIndices;
+    endResetModel();
+}
+
+void RemotePublicDecksModel::refresh(const QString &userName)
+{
+    if (loading) {
```
ZeldaZach, line 135: Same latch shape as the visibility counter. `loading` is only cleared in `decksReceived`, and `RemoteClient::ping()` can drop a pending command without ever emitting `finished`. When that happens the tab is stuck on "Loading public decks..." with the grid hidden, and this guard makes the refresh button permanently inert with no way out short of restarting.

Add a recovery path: a timer started in `refresh()` that calls `setLoading(false)` and emits `loadFailed` with a timeout message, plus clearing `loading` when the client status goes to disconnected.

cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/public_deck_preview_widget.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    const ExactCard bannerCard =
+        entry.bannerCardName.isEmpty()
+            ? ExactCard()
+            : CardDatabaseManager::query()->getCard(CardRef{entry.bannerCardName, entry.bannerCardProvider});
+    bannerCardDisplayWidget->setCard(bannerCard);
+
+    // The deck name is the overlay text on the banner, like the local preview.
+    bannerCardDisplayWidget->setOverlayText(entry.name);
+    setToolTip(entry.name);
```
ZeldaZach, line 99: `entry.name` comes straight from another user's deck record, and Qt tooltips are rendered by a QLabel in `Qt::AutoText` mode, so a deck named something like `<h1><table><tr>...` is parsed as HTML and renders as markup.

The overlay text is painted with QPainter so that path is fine, but the tooltip is not. Escape and bound it:

```cpp
setToolTip(entry.name.left(200).toHtmlEscaped());
```

The same applies to `titleLabel` in `TabPublicDecks`, which interpolates a remote username into a QLabel.

cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.cpp
```diff
@@ -153,6 +161,8 @@ QVariant RemoteDeckList_TreeModel::data(const QModelIndex &index, int role) cons
                         return file->getId();
                     case 2:
                         return file->getUploadTime();
+                    case 3:
+                        return isEffectivelyPublic(file) ? tr("Public") : tr("Private");
```
ZeldaZach, line 165: This column reports the inherited effective state, while `TabDeckStorage::actPublishDeck` toggles `node->isPublic()`, the node's own bit. The two disagree in a case users will hit: for a private deck inside a public folder the column already says "Public", so pressing the toolbar button flips the own bit but the cell text does not change and the action reads as broken. Press it again and the deck is silently unpublished.

Distinguishing them fixes the confusion, e.g. return `tr("Public (inherited)")` when `!node->isPublic() && isEffectivelyPublic(node)`, and say why in the tooltip.

cockatrice/src/interface/widgets/tabs/tab_public_decks.cpp
```diff
@@ -0,0 +1,230 @@
...
+            &TabPublicDecks::updateColorFilter);
+    connect(colorIdentityFilter, &DeckPreviewColorIdentityFilterWidget::filterModeChanged, this,
+            &TabPublicDecks::updateColorFilter);
+    connect(tagFilterWidget, &VisualDeckStorageTagFilterWidget::filterChanged, this, &TabPublicDecks::updateTagFilter);
+    connect(quickSettingsWidget, &PublicDecksQuickSettingsWidget::cardSizeChanged, this,
+            &TabPublicDecks::updateCardSize);
+    connect(quickSettingsWidget, &PublicDecksQuickSettingsWidget::showTagFilterChanged, this,
+            &TabPublicDecks::updateTagsVisibility);
+
+    model->refresh(userName);
```
ZeldaZach, line 110: `retranslateUi()` is never called from this constructor, and `TabSupervisor::myAddTab` does not call it either. Since `refreshButton->setToolTip(tr("Refresh"))` and `quickSettingsWidget->setToolTip(tr("Public Decks Settings"))` only exist inside `retranslateUi`, both controls ship with no tooltip and nothing for a screen reader to announce until the user changes language.

Every other tab (`TabRoom`, `TabMessage`, `TabAdmin`) calls `retranslateUi()` at the end of its constructor. Add the same call here, before `model->refresh(userName)`. Given the accessibility work elsewhere in this PR, `refreshButton` also wants an explicit `accessibleName`, since an icon-only QToolButton has no accessible text.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
```diff
@@ -795,7 +811,49 @@ void TabDeckStorage::shareFromTreeFinished(const Response &response, const Comma
...
+        connect(pend, &PendingCommand::finished, this, &TabDeckStorage::setVisibilityFinished);
+        client->sendCommand(pend);
+        ++pendingVisibilityChanges;
+    }
+}
+
+void TabDeckStorage::setVisibilityFinished(const Response &r, const CommandContainer & /*commandContainer*/)
+{
+    if (r.response_code() != Response::RespOk) {
+        QMessageBox::critical(this, tr("Error"),
```
ZeldaZach, line 850: This fires once per node in the batch, so publishing a ten-deck selection against a server that rejects the command stacks ten modal dialogs the user dismisses one at a time. Collect failures while the batch drains and show one summary when the last reply lands, or at minimum suppress duplicates after the first failure in a batch.

cockatrice/src/interface/widgets/tabs/tab_public_decks.cpp
```diff
@@ -0,0 +1,230 @@
...
+QString TabPublicDecks::getTabText() const
+{
+    return tr("Public decks of %1").arg(userName);
+}
+
+void TabPublicDecks::retranslateUi()
+{
+    titleLabel->setText(tr("Public decks of %1").arg(userName));
+    searchWidget->setPlaceholderText(tr("Search by deck name"));
+    emptyLabel->setText(tr("This user has not published any decks."));
```
ZeldaZach, line 122: This unconditionally overwrites the "No decks match your filters." variant that `rebuildGrid` picks, so a language change while a filter is active shows the wrong message. `statusLabel` is not retranslated at all, so a visible loading or error message stays in the old language.

Reuse the same choice `rebuildGrid` makes (`model->totalCount() > 0 ? ... : ...`), and either re-emit the current status text or clear it here.

cockatrice/src/interface/widgets/visual_deck_storage/remote_public_decks_model.cpp
```diff
@@ -0,0 +1,202 @@
...
+                        activeSet.insert(color.toUpper());
+                    }
+                    QSet<QChar> identitySet;
+                    for (const QChar &color : identity) {
+                        identitySet.insert(color.toUpper());
+                    }
+                    colorMatch = activeSet == identitySet;
+                    break;
+                }
+                case VisualDeckStorageSortFilterProxyModel::Includes:
```
ZeldaZach, line 97: This switch is a verbatim copy of `VisualDeckStorageSortFilterProxyModel::updateColorMatches` (`visual_deck_storage_sort_filter_proxy_model.cpp:260-283`), down to the `ExactMatch` uppercase normalization and the fact that `Includes`/`Excludes` do not normalize case. Two copies of a subtle rule like that will drift.

Pull it into a free function next to the enum, something like `bool colorIdentityMatches(FilterMode, const QSet<QChar> &, const QString &)`, and call it from both.

cockatrice/src/interface/widgets/tabs/public_decks_quick_settings_widget.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <QCheckBox>
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QSpinBox>
+#include <libcockatrice/settings/cards_display_settings.h>
+#include <libcockatrice/settings/personal_settings.h>
+#include <libcockatrice/settings/visual_deck_storage_settings.h>
+
+PublicDecksQuickSettingsWidget::PublicDecksQuickSettingsWidget(QWidget *parent) : SettingsButtonWidget(parent)
```
ZeldaZach, line 14: This class is `VisualDeckStorageQuickSettingsWidget` minus the folders, banner combobox and tooltip-type controls, with identical wiring for the five shared checkboxes, the opacity spinbox and the card size widget. Both write the same SettingsCache keys.

Adding the new Show Upload Time checkbox to the existing widget and giving it a way to hide the controls that do not apply would avoid maintaining a second near-identical file that has to be kept in sync by hand.

cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.h
```diff
@@ -125,6 +145,10 @@ class RemoteDeckList_TreeWidget : public QTreeView
 
 public:
     explicit RemoteDeckList_TreeWidget(AbstractClient *_client, QWidget *parent = nullptr);
+    [[nodiscard]] RemoteDeckList_TreeModel *getModel() const
```
ZeldaZach, line 148: This accessor is added but never called anywhere in the stack, and it hands out the model past the widget's existing wrapper methods (`addFileToTree`, `refreshTree`, and so on) that exist precisely to keep it encapsulated. Drop it unless something later actually needs it.

cockatrice/src/interface/widgets/tabs/public_decks_quick_settings_widget.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+int PublicDecksQuickSettingsWidget::getUnusedColorIdentitiesOpacity() const
+{
+    return unusedColorIdentitiesOpacitySpinBox->value();
+}
+
+CardSizeWidget *PublicDecksQuickSettingsWidget::getCardSizeWidget() const
+{
+    return cardSizeWidget;
+}
```
ZeldaZach, line 151: This file, `public_decks_quick_settings_widget.h` and `public_deck_preview_widget.cpp` all end without a trailing newline. Worth adding so diffs of the final line stay clean and clang-format in CI does not flag them.

ZeldaZach (changes requested):
The shape of this feature is good. A dedicated per-user tab, a flat read-only model over the server tree, tiles rendered purely from stored metadata so browsing costs no downloads, and sensible reuse of the existing color identity filter, tag filter, search and card size widgets. Keyboard operability on the tile and the loading/empty/error label triad are genuinely thought through rather than bolted on, and the new setting is defaulted and tested consistently with its neighbours.

One thing needs fixing before this can land, and it is not in the new code. Saving a remote deck from the deck editor sends only the `tags` field, but the server's `deck_id` branch of `cmdDeckUpload` runs an unconditional UPDATE across all four metadata columns, so every unset field is written as empty. The practical effect is that editing and saving any remote deck blanks the banner card and color identity, and the preview tile this PR introduces loses its art and pips. `TabDeckStorage::uploadDeck` already sets all four; the deck editor path just needs to match.

Two reliability issues share a root cause worth calling out together. Both `pendingVisibilityChanges` and the model's `loading` flag are one-way latches, and `RemoteClient::ping()` drops timed-out commands with `deleteLater()` without ever running `processResponse`. So a single lost reply leaves the visibility column permanently unrefreshable, or the public decks tab permanently stuck on its loading state with the refresh button inert. Neither has a disconnect reset. Coalescing with a restartable timer instead of counting would make a lost reply cost one stale refresh rather than a dead feature for the rest of the session.

Two smaller correctness notes. The deck name in the tile tooltip comes from another user and goes into a QLabel in AutoText mode, so it needs `toHtmlEscaped()` and a length bound. And the Visibility column shows the inherited effective state while the toolbar button toggles the node's own bit, which makes publishing a deck that sits inside a public folder look like it did nothing.

The rest is polish: a missing `retranslateUi()` call in the tab constructor that leaves both toolbar controls with no tooltip and no accessible text, a per-node error dialog that stacks up during batch publish, and a color identity switch duplicated verbatim from the proxy model.

cockatrice/src/interface/widgets/tabs/abstract_tab_deck_editor.cpp
BruebachL: Since this review the deck-upload protocol was reworked: the server now derives the banner card, banner provider and tags from the uploaded deck_list itself (command_deck_upload.proto reserves those client fields), so a remote save can no longer wipe them, and the editor ships its client-computed color identity instead — exactly the value your snippet called set_color_identity. See fa9f642ea9, which adapts the editor save, `TabDeckStorage::uploadDeck` and the repeated-tags read to that protocol. The four-column update in cmdDeckUpload now gets every column filled from reliable sources on both the insert and deck_id branches.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
BruebachL: Fixed in 33a4efcc0e: the one-way counter is gone. Publishing grabs a bool and restarts a single-shot timer that calls refreshTree() once when publishes quiet down, and the timer/flag are cleared on disconnect, so a dropped reply now costs one stale refresh instead of a permanently dead column.

cockatrice/src/interface/widgets/visual_deck_storage/remote_public_decks_model.cpp
BruebachL: Fixed in ac223795e1: refresh() starts a single-shot timeout timer that clears loading and emits loadFailed with a timeout message, and the latch clears on disconnect, so a dropped DeckListOtherUser reply can no longer leave the tab stuck on the loading state with an inert refresh button.

cockatrice/src/interface/widgets/visual_deck_storage/deck_preview/public_deck_preview_widget.cpp
BruebachL: Fixed in 13b0b4095c: the deck-name tooltip is now bounded and HTML-escaped (setToolTip(entry.name.left(200).toHtmlEscaped())), and the username interpolated into the AutoText title label is escaped the same way in both the constructor and retranslateUi.

cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.cpp
BruebachL: Fixed in 100e7441c8: the visibility column now reports the node's own bit — a private deck inside a public folder reads 'Public (inherited)' instead of 'Public', with the tooltip explaining that a parent folder is public, so the toggle no longer looks broken.

cockatrice/src/interface/widgets/tabs/tab_public_decks.cpp
BruebachL: Fixed in f33971ddd5: retranslateUi() is called at the end of the constructor (before the first refresh), and the icon-only refresh button now carries an explicit accessible name.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
BruebachL: Fixed in 3facbc0618: failures are collected while the batch is in flight and surfaced as a single summary dialog when the visibility timer drains; only a reply landing outside an active batch still reports immediately.

cockatrice/src/interface/widgets/tabs/tab_public_decks.cpp
BruebachL: Fixed in bf2730dfcf: retranslateUi now picks the same empty-label variant rebuildGrid does (No decks match your filters vs. nothing published yet) and re-shows the status text so a visible loading message retranslates instead of staying stale.

cockatrice/src/interface/widgets/visual_deck_storage/remote_public_decks_model.cpp
BruebachL: Fixed in 1513d24f37: the switch is extracted to a free colorIdentityMatches(FilterMode, colors, identity) declared next to the FilterMode enum, and both updateColorMatches and RemotePublicDecksModel::rebuildVisibleIndices call it — one copy of the exact-match normalization and case handling.

cockatrice/src/interface/widgets/tabs/public_decks_quick_settings_widget.cpp
BruebachL: Fixed in ef7d65d309: the duplicate widget is gone. Show Upload Time now lives in VisualDeckStorageQuickSettingsWidget alongside the shared keys, setPublicDecksMode(true) hides the folders/banner/tooltip controls that do not apply to the public-decks tab, and TabPublicDecks uses the shared widget (public_decks_quick_settings_widget.* deleted).

cockatrice/src/interface/widgets/server/remote/remote_decklist_tree_widget.h
BruebachL: Fixed in 6bebd897ff: the unused getModel() accessor is dropped; the tree widget only exposes its wrapper methods again.

cockatrice/src/interface/widgets/tabs/public_decks_quick_settings_widget.cpp
BruebachL: Fixed in 4272c93a06: the three files (plus user_context_menu.cpp, which had the same issue) now end with a trailing newline.

cockatrice/src/interface/widgets/visual_deck_storage/remote_public_decks_model.cpp
```diff
@@ -0,0 +1,209 @@
...
+{
+    if (loading) {
+        return;
+    }
+    setLoading(true);
+    loadingTimeoutTimer->start();
+    Command_DeckListOtherUser cmd;
+    cmd.set_user_name(userName.toStdString());
+    PendingCommand *pend = client->prepareSessionCommand(cmd);
+    connect(pend, &PendingCommand::finished, this, &RemotePublicDecksModel::decksReceived);
```
ZeldaZach: Non-blocking, follow-on to the loading-latch fix: the timeout clears `loading` but leaves this `PendingCommand` connected, so the old reply can still land later. Sequence: refresh(A) times out, the user refreshes again, then A's late reply arrives, calls `loadingTimeoutTimer->stop()` (killing the second request's timeout) and repopulates the grid with A's data — which the in-flight second reply then overwrites again.

Nothing here corrupts state and the guard keeps requests strictly serialized, so this is just a transient flicker plus one lost timeout. A monotonically increasing request id captured in the lambda/slot and compared on arrival (drop anything that is not the newest) would close it if you touch this file again.

cockatrice/src/interface/widgets/tabs/tab_public_decks.cpp
```diff
@@ -0,0 +1,238 @@
...
+    // The same choice rebuildGrid makes, so a language change does not swap
+    // the "no match" variant for the "nothing published" one.
+    emptyLabel->setText(model->totalCount() > 0 ? tr("No decks match your filters.")
+                                                : tr("This user has not published any decks."));
+    refreshButton->setToolTip(tr("Refresh"));
+    refreshButton->setAccessibleName(tr("Refresh"));
+    quickSettingsWidget->setToolTip(tr("Public Decks Settings"));
+    // Re-show the status so a visible loading message picks up the new language
+    // instead of staying stale; if nothing is shown it just stays hidden.
+    updateLoadingState(model->isLoading());
```
ZeldaZach: Non-blocking. `statusLabel` carries both the loading message and the `loadFailed` message, and this call takes the `else` branch whenever `isLoading()` is false — so changing language while a server-error or timeout message is on screen hides the error and shows the (empty) grid instead. The comment above says "re-show the status", but the error text is not re-shown, only the loading variant is.

Keeping the last failure message in a member and re-applying it here when not loading would make the retranslate faithful in both states.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
```diff
@@ -811,3 +841,66 @@ void TabDeckStorage::onShareFromTreeTimeout()
...
+        }
+        // Toggle the node's own visibility bit (what the server persists); the
+        // effective visibility shown by the column may additionally be inherited
+        // from a parent folder.
+        cmd.set_is_public(!node->isPublic());
+
+        PendingCommand *pend = client->prepareSessionCommand(cmd);
+        connect(pend, &PendingCommand::finished, this, &TabDeckStorage::setVisibilityFinished);
+        visibilityRefreshStarted = true;
+        visibilityRefreshTimer->start();
```
ZeldaZach, line 896: Non-blocking. The timer starting here at send time is what makes the latch unstuck-able, which is the right call. The edge case is a round trip slower than `VISIBILITY_REFRESH_DELAY` (500 ms): the timer drains before any reply lands, `onVisibilityRefreshTimeout` clears `visibilityRefreshStarted` and calls `refreshTree()` while the server has not applied the change yet, so the column re-reads the old state. The replies that arrive afterwards then see `visibilityRefreshStarted == false`, so the RespOk branch does not restart the timer and the column stays stale until the next manual refresh; a rejected one pops its own modal again, which is the stacking this was meant to avoid.

Half a second is a plausible RTT for a remote server. Starting the timer only on reply (with a longer fallback armed at send time), or leaving `visibilityRefreshStarted` true until the tree read actually returns, would cover it.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
```diff
@@ -1037,6 +1042,30 @@ void TabSupervisor::roomLeft(TabRoom *tab)
...
+    setCurrentWidget(tab);
+}
+
+void TabSupervisor::publicDecksClosed(TabPublicDecks *tab)
+{
+    if (tab == currentWidget()) {
+        emit setMenu();
+    }
+
+    publicDecksTabs.remove(tab->getUserName());
```
ZeldaZach, line 1069: Non-blocking. `TabSupervisor::stop()` builds `tabsToDelete` from `roomTabs` and `gameTabs` only, so a public decks tab survives a disconnect while every other network tab closes. It then sits there with stale contents, and its refresh button still calls `client->prepareSessionCommand()` on a dead client — harmless now that the model times out, but the user gets a 30-odd second spinner ending in "the server did not respond in time" rather than the tab going away with the session.

Adding the `publicDecksTabs` values to `tabsToDelete` in `stop()` would fold it into the existing cleanup (the `closing` signal already unregisters them here).

Also FYI, this mirrors `roomLeft` exactly, including the fact that `removeTab` does not delete the widget — pre-existing pattern, not something to fix in this PR.

ZeldaZach (approved):
Re-reviewed against the head commit. Every point from the previous round is fixed in the code, not just in the replies:

- The deck-editor save no longer blanks metadata. The protocol reworks banner card and tags to be server-derived, and `actSaveDeck` now sends the client-computed color identity, matching `TabDeckStorage::uploadDeck`.
- Both one-way latches are gone. `pendingVisibilityChanges` is replaced by a restartable single-shot timer with a bool guard, stopped on disconnect; the model's `loading` flag now has a timeout timer and clears on `StatusDisconnected`, so a dropped reply costs one stale refresh instead of a dead feature.
- Remote-supplied text is handled: the deck-name tooltip is bounded and `toHtmlEscaped()`, the username in `titleLabel` is escaped, the banner overlay and tag chips are `QPainter` draws, and the tab text goes through the existing `sanitizeTabName`/`sanitizeHtml`.
- The visibility column distinguishes `Public (inherited)` with an explanatory tooltip, so the toggle no longer reads as a no-op.
- `retranslateUi()` runs from the constructor, the icon-only refresh button has a tr()-wrapped accessible name, and retranslation now picks the same empty-label variant `rebuildGrid` does.
- Batch publish failures are summarized once instead of stacking a modal per node.
- The color-identity rule is extracted to one shared `colorIdentityMatches()`, the duplicate quick-settings widget is deleted in favour of `setPublicDecksMode()`, the unused accessor is dropped, and the trailing newlines are there.

I also checked the things the new code could plausibly get wrong: `begin/endResetModel` pairing and index bounds in `RemotePublicDecksModel` are correct, `FlowWidget::clearLayout` does `deleteLater` so the per-keystroke grid rebuild does not leak tiles, every `PendingCommand` is connected with `this` as the context object so a closed tab cannot be called back, and `TabSupervisor::client` is assigned once so a lingering tab cannot hold a dangling client.

Four non-blocking notes left inline, none of which should hold up the merge: a late reply after a timeout can still repaint the grid (a request generation counter would close it), the 500 ms visibility drain can fire before a slow round trip completes, a language change hides a displayed error message, and public-decks tabs are not added to `TabSupervisor::stop()`'s cleanup so they outlive a disconnect.

Approving.

cockatrice/src/interface/widgets/visual_deck_storage/remote_public_decks_model.cpp
BruebachL: Fixed in 2d15529d48: each refresh now captures a monotonically increasing request id, and the finished lambda only calls decksReceived when the reply still belongs to the newest request. A reply landing after its own timeout can no longer stop the newer request timer or repaint the grid with stale data.

cockatrice/src/interface/widgets/tabs/tab_public_decks.cpp
BruebachL: Fixed in dd2400d245: the tab keeps the last failure message, retranslateUi re-shows it (keeping the grid hidden) whenever the model is not loading, and the stored message clears once a new refresh starts, so a language change no longer hides a server-error or timeout message.

cockatrice/src/interface/widgets/tabs/tab_deck_storage.cpp
BruebachL: Fixed in e0f8fd0091: the single-shot timer is armed with the full network timeout when a publish is sent (so a dropped reply still drains once) and re-armed with the 500 ms delay on every reply, so a slow round trip can no longer fire the drain before the server applied the change and leave the column stale.

cockatrice/src/interface/widgets/tabs/tab_supervisor.cpp
BruebachL: Fixed in f485b974df: TabSupervisor::stop() now adds the publicDecksTabs values to tabsToDelete alongside the room and game tabs, so a public decks tab closes with the session instead of outliving a disconnect.

PR #7247, opened 2026-09-05
[CI] Add ccache eviction to all builds

tooomm:
## Related Ticket(s)
- Related #6795

Debug builds were using ccache, but did not evict the cache after compilation unlike all other builds.

- Use similar command structure than for building release types on Linux
- While the compiler cache is shared and often both types are run after each other (see screenshot, the release run following the debug one would clean the shared cache in that case), it means the order of steps matters and e.g. our Arch build which does not have a release stream is not cleaned up at all.

  Without this change, the Arch cache will grow without limits over time. It will only be stopped by unintentional cache purging and rotation by chance because our caching is not well designed yet and it happens regularly that ccache entries are removed by GitHub because we constantly overrun the provided 10 GB limit.

  Existing cache is pulled --> new caches are added --> uploaded --> next build pulls it and adds more --> uploaded --> ...

  With this change, outdated cache content will be purged following a controlled approach before uploading and size stays in check because only recent data is kept.

## Screenshots
See missing `evict ccache` log group in the "Build debug and test" step:

Debian 13:
image

Arch (only "Build debug and test" step, no "Build release package" one):
image

<br><br>

Config happens in our CI build script:
https://github.com/Cockatrice/Cockatrice/blob/61e6a9913e4096303b8be61be727fbd7015b6a24/.ci/compile.sh#L76-L84
https://github.com/Cockatrice/Cockatrice/blob/61e6a9913e4096303b8be61be727fbd7015b6a24/.ci/compile.sh#L279-L282

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -176,8 +176,12 @@ jobs:
         shell: bash
         run: |
           source .ci/docker.sh
-          RUN --server --debug --test --ccache "$CCACHE_SIZE" \
-            --cmake-generator "$CMAKE_GENERATOR"
+          args=()
+          [[ $GITHUB_REF == "refs/heads/master" ]] && args+=(--evict-ccache "$CCACHE_EVICTION_AGE")
+          args+=(--ccache "$CCACHE_SIZE")
+          args+=(--cmake-generator "$CMAKE_GENERATOR")
+
+          RUN --server --debug --test "${args[@]}"
 
       - name: "Build release package"
         id: build


BruebachL (approved):
The problem and approach read as correct to me. I'll be honest, I'm not versed enough in our CI scripts to properly evaluate the change to the code but considering it is minor and you obviously have an eye on these metrics (i.e. if this breaks something, you'll notice and fix it) so I have no problem approving this and deferring to your expertise.

tooomm:
Thanks, at least a second pair of eyes to give it some basic check does not hurt and keep awareness on what's touched. 👍 

I'm still trying to figure out how to improve our caching approach with complete compiler cache for Windows still missing which is crucial for somewhat performant builds and quicker checks in PR's...

We have too many builds happening 🙈

PR #7248, opened 2026-09-05
[Game] Prevent spectator duplication when replaying joined events

BruebachL:
The spectator branch of eventJoin emitted spectatorJoined unconditionally even when the spectator was already present (e.g. replayed during a rewind). 

- Guard it like the player branch and eventGameStateChanged, and make PlayerListWidget::addPlayer idempotent as defense in depth.


RickyRister (approved):
This fixes the issue of rewinding causing duplicate spectators, but rewinding a replay will still cause the spectator to remain in the list even if it rewinds before the spectator join. 

Seems like an existing issue though, so that can be fixed separately in the future.

PR #7250, opened 2026-09-05
[Client] Show horizontal art for cards in the profile banner and playmats

### LINKED ISSUE #7118: [Beta Feature] Horizontal art crop of planes/invasions in server profile banner card
Client Version: 3.1.0-beta.5 (2026-08-14)
Client Operating System: Ubuntu 26.04 LTS
Build Architecture: 64-bit
Qt Version: 6.10.2

Really enjoying the changes to the server UI, especially the new avatars and the wider sections for user information in the friends/online list. When thinking about what banner card to set in my server profile, it seemed like the dimensions would be perfect to use the horizontal art from a Plane or an Invasion. 

However, when setting either type of card, the preview shows the entire card (not just the art crop), rotated 90°, rendering it really small and unusable. It would be wonderful if this banner card system could be improved to support horizontal cards. Thank you!

Image

BruebachL:
## Related Ticket(s)
- Fixes #7118

Use the landscape orientation flag to rotate sideways-layout card art upright before cropping, so planes/sieges show their horizontal art as the server profile banner card instead of a rotated full card.

- Curve cropCardArt around the card's landscapeOrientation flag with landscape-specific art margins (mirrors CardInfoPictureWidget)
- Add shared CardArtUtils::rotateSidewaysLayoutArt helper and apply it to the playmat (game render and settings preview) and card info widget, replacing three duplicate 90-degree rotation blocks

### DIFF
--- a/cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
+++ b/cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
@@ -2,6 +2,7 @@
 
 #include "../../card_picture_loader/card_picture_loader.h"
 #include "../cards/art_crop_attribution.h"
+#include "../cards/card_art_utils.h"
 #include "../utility/completer_utils.h"
 #include "card_database_display_model.h"
 #include "card_database_model.h"
@@ -276,7 +277,7 @@ void PlaymatSettingsDialog::reloadPreview()
         return;
     }
 
-    currentPixmap = fullRes;
+    currentPixmap = CardArtUtils::rotateSidewaysLayoutArt(fullRes, card);
     preview->setPixmap(currentPixmap);
     preview->setParams(currentParams);
     preview->setAttribution(buildArtAttribution(card));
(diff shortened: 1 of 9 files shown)


cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
```diff
@@ -276,7 +277,7 @@ void PlaymatSettingsDialog::reloadPreview()
         return;
     }
 
-    currentPixmap = fullRes;
+    currentPixmap = CardArtUtils::rotateSidewaysLayoutArt(fullRes, card);
```
RickyRister, line 280: Why is the minimum zoom for landscape layout cards bounded at 1.24x? I was able to get normal cards down to 0.78x zoom.

Image

RickyRister (approved):
It works

cockatrice/src/interface/widgets/playmat/playmat_settings_dialog.cpp
BruebachL: Math reasons tbh. It maths out to be equal in effect for the adjusted dimensions.

PR #7251, opened 2026-09-05
[Servatrice] Detect MySQL strict mode on startup and exit early

### LINKED ISSUE #6277: have servatrice exit early if strict mode is on
**Similar Requests**
#865

servatrice can not use the database in almost every request because of strict mode, this is explained nowhere to users of servatrice and describes 99% of problems experienced when using servatrice.
this could be made clearer by having servatrice check if strict mode is on on startup when configured to use sql database.
servatrice would simply exit with an error message, instructing the user to disable strict mode.

**Additional Context**
strict mode used to not be on by default, this issue has been happening ever since strict mode was made the default for new databases, because of this the issue has gone largely unnoticed even though it affects all new databases, we just keep telling people to disable strict mode. in a way this has made running servatrice with db exclusive to the ones that know about this issue, as absolutely nothing about the errors that occur when the db is accessed without strict mode tells you that the way to fix this is with strict mode.

BruebachL:
## Related Ticket(s)
- Fixes #6277

Servatrice fails on almost every database request when MySQL/MariaDB strict mode is enabled, but nothing about the errors hints that the fix is disabling strict mode.

- Check `@@GLOBAL.sql_mode` when opening a MySQL/MariaDB connection in `Servatrice_DatabaseInterface::openDatabase()`.
- Exit early with an error message instructing how to disable strict mode when `STRICT_TRANS_TABLES` or `STRICT_ALL_TABLES` is present.

### DIFF
--- a/servatrice/src/servatrice_database_interface.cpp
+++ b/servatrice/src/servatrice_database_interface.cpp
@@ -102,7 +102,7 @@ bool Servatrice_DatabaseInterface::openDatabase()
     if (isStrictModeEnabled()) {
         qCCritical(DatabaseInterfaceLog) << poolStr
                                          << "Error opening database: MySQL/MariaDB strict mode is enabled, which "
-                                            "breaks most servatrice database operations. Please disable strict mode "
+                                            "breaks most Servatrice database operations. Please disable strict mode "
                                             "by removing STRICT_TRANS_TABLES and STRICT_ALL_TABLES from sql_mode, "
                                             "for example by adding 'sql_mode=NO_ENGINE_SUBSTITUTION' under [mysqld] "
                                             "in your my.cnf (or my.ini on Windows) and restarting the database "
--- a/servatrice/src/servatrice_database_interface.cpp
+++ b/servatrice/src/servatrice_database_interface.cpp
@@ -12,6 +12,7 @@
 #include <QLoggingCategory>
 #include <QSqlError>
 #include <QSqlQuery>
+#include <QStringList>
 #include <libcockatrice/deck_list/deck_list.h>
 #include <libcockatrice/protocol/pb/game_replay.pb.h>
 #include <libcockatrice/protocol/pb/serverinfo_user.pb.h>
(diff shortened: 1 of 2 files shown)


servatrice/src/servatrice_database_interface.cpp
```diff
@@ -98,12 +99,43 @@ bool Servatrice_DatabaseInterface::openDatabase()
         return false;
     }
 
+    if (isStrictModeEnabled()) {
+        qCCritical(DatabaseInterfaceLog) << poolStr
+                                         << "Error opening database: MySQL/MariaDB strict mode is enabled, which "
+                                            "breaks most servatrice database operations. Please disable strict mode "
```
tooomm:
```suggestion
                                            "breaks most Servatrice database operations. Please disable strict mode "
```

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -98,12 +99,43 @@ bool Servatrice_DatabaseInterface::openDatabase()
...
+
     // reset all prepared statements
     qDeleteAll(preparedStatements);
     preparedStatements.clear();
     return true;
 }
 
+bool Servatrice_DatabaseInterface::isStrictModeEnabled() const
+{
+    if (sqlDatabase.driverName() != "QMYSQL") {
```
ebbit1q: the qmysql driver (does both mysql and mariadb) is the only database we actually support, imo having this return false should also be an error

servatrice/src/servatrice_database_interface.cpp
```diff
@@ -98,12 +99,43 @@ bool Servatrice_DatabaseInterface::openDatabase()
...
+        return false;
+    }
+
+    QSqlQuery query(sqlDatabase);
+    if (!query.exec("SELECT @@GLOBAL.sql_mode")) {
```
ebbit1q, line 141: what are we supposed to do when this fails? that'd mean the database is unusable most likely

ebbit1q (approved):
this is great, this is the most common issue by people using a db, this is good as is or you can look at my comments

servatrice/src/servatrice_database_interface.cpp
BruebachL: Fail loudly, I guess

PR #7253, opened 2026-09-06
[Oracle/Client] Report card database download progress

### LINKED ISSUE #3484: Have a progress bar for mtgjson parsing
Follow up ticket to the PR #3480.

This was especially needed when parsing took over 60 seconds.
Still a great addition and helpful user feedback. Parsing still takes 20-30 seconds!

BruebachL:
## Related Ticket(s)
- Fixes #3484

Card database updates ran invisibly: MTGJSON parsing spun an indeterminate bar on the UI thread and the set import blocked the window, while the onboarding wizard spawned `oracle -b` with no progress to show at all.

- Add byte-level scan progress to `RawJson::scanSetRanges` via an optional callback, throttled to ~100 reports per scan.
- Emit `OracleImporter::dataReadProgress` during the scan and import sets on a worker thread, driving the wizard's progress bar per set.
- With `-b`, write machine-readable `PROGRESS <stage> <done> <total>` lines to stdout for the download/scan/import stages. stderr keeps the log output.
- Parse the oracle stdout in `MainWindow` and forward it to the onboarding wizard, giving the card database step a determinate bar with stage-specific status text.
- Guard the async workers against the wizard being closed mid-run.
- Add Google Test coverage for scan progress reporting.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -557,6 +557,8 @@ int OracleImporter::startImport()
 {
     static ICardSetPriorityController *noOpController = new NoopCardSetPriorityController();
 
+    importCancelled.storeRelease(0);
+
     // Pre-allocate the cards hash to avoid rehashing during import. Keys are
     // distinct card names while raw ranges only count printings (AllPrintings
     // ~100k printings vs ~35k names), so this over-reserves somewhat; an exact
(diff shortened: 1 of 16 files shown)


tooomm (commented):
After downloading finishes, the Oracle UI still freezes for several seconds for me:
image

Then the parsing bar shows up:
image

Afer completion, the next page shows up where the final import report used to be.
It now shows a loading bar and sets are sequentially added:
image

<br>

If clicking `back` or `cancel` on that last page during the ongoing progress, this error message is shown a few seconds later:
image

<br>

Closing Cockatrice while Oracle runs, makes Oracle silently quit as well.
If they should be coupled, a warning should be shown, similar when trying to close the app while a game is still running.

oracle/src/oracleimporter.cpp
```diff
@@ -47,8 +47,14 @@ static CardSet::Priority getSetPriority(const QString &setType, const QString &s
 
 bool OracleImporter::readSetsFromByteArray(QByteArray data)
 {
+    const RawJson::ScanProgressCallback progress =
+        progressReporting ? RawJson::ScanProgressCallback([this](qsizetype bytesRead, qsizetype totalBytes) {
```
RickyRister: nit: I don't think you need to wrap the lambda in a `RawJson::ScanProgressCallback`

tooomm (approved):
My notes above are addressed - I like the new UX 👍

PR #7255, opened 2026-09-06
Use capitalized app names

tooomm:
`Cockatrice`
`Oracle`
`Servatrice`

where applicable... (+ updated URL's)

### DIFF
--- a/oracle/src/parsehelpers.cpp
+++ b/oracle/src/parsehelpers.cpp
@@ -20,7 +20,7 @@
  * Note that "...enters tapped unless..." returns false.
  *
  * @param name The name of the card
- * @param text The oracle text of the card
+ * @param text The Oracle text of the card
  */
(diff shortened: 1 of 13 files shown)


BruebachL (approved):

oracle/src/parsehelpers.cpp
```diff
@@ -20,7 +20,7 @@
  * Note that "...enters tapped unless..." returns false.
  *
  * @param name The name of the card
- * @param text The oracle text of the card
+ * @param text The Oracle text of the card
```
RickyRister, line 23: Are you sure this should be capitalized? This is referring to the mtg oracle text instead of our Oracle program

tooomm: Looking at the competitive rules, this change is correct.

PR #7259, opened 2026-09-06
[Client] Fix spurious server room join error

BruebachL:
The server replies RespContextError when a join command is received for a room that connection is already registered in. The client was sending such duplicate joins in benign situations - double-clicking to join a room, or clicking a room the selector was already auto-joining - and answered them with a modal telling users to restart the client.

- Joins for the same room are now deduplicated while one is in flight, and a remaining RespContextError is healed by leaving and rejoining the room so the tab appears without a client restart. Error dialogs are only shown for user-initiated joins, so failed auto-joins no longer spam critical popups.

### DIFF
--- a/cockatrice/src/interface/widgets/tabs/tab_server.cpp
+++ b/cockatrice/src/interface/widgets/tabs/tab_server.cpp
@@ -13,6 +13,7 @@
 #include <libcockatrice/protocol/pb/event_list_rooms.pb.h>
 #include <libcockatrice/protocol/pb/event_server_message.pb.h>
 #include <libcockatrice/protocol/pb/response_join_room.pb.h>
+#include <libcockatrice/protocol/pb/room_commands.pb.h>
 #include <libcockatrice/protocol/pb/session_commands.pb.h>
 #include <libcockatrice/protocol/pending_command.h>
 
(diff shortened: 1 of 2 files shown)


cockatrice/src/interface/widgets/tabs/tab_server.cpp
```diff
@@ -185,60 +186,90 @@ void TabServer::processServerMessageEvent(const Event_ServerMessage &event)
...
-            QMessageBox::critical(
-                this, tr("Error"),
-                tr("The server thinks you are in the server room but your client is unable to display it. "
-                   "Try restarting your client."));
-            emit roomJoinFailed(roomId);
+            // The server already had us registered in the room even though no tab was open,
+            // usually because two join attempts for the same room overlapped. Leaving and
+            // rejoining makes the server reply with a fresh RespOk so the tab is displayed
+            // without requiring a client restart.
+            leaveAndRejoinRoom(roomId, setCurrent);
```
RickyRister, line 260: Are we sure the server won't ever return `RespContextError` for some other error in the future? `leaveAndRejoinRoom` calls `joinRoom` again, so this will be an infinite loop if the server keeps returning `RespContextError`.

cockatrice/src/interface/widgets/tabs/tab_server.h
```diff
@@ -58,10 +60,16 @@ private slots:
...
 
 private:
+    void leaveAndRejoinRoom(int roomId, bool setCurrent);
+
     AbstractClient *client;
     RoomSelector *roomSelector;
     QTextBrowser *serverInfoBox;
     bool shouldEmitUpdate = false;
+    /** Room ids with a join command in flight, mapped to whether the tab should be focused once it opens. */
+    QMap<int, bool> pendingRoomJoins;
```
RickyRister, line 70: nit: you can use a `QHash` here since you don't care about the order.

cockatrice/src/interface/widgets/tabs/tab_server.cpp
```diff
@@ -185,60 +186,105 @@ void TabServer::processServerMessageEvent(const Event_ServerMessage &event)
...
+                // A stale-membership heal was already attempted once; if the server still
+                // rejects the join there is nothing left to do client-side, so surface it.
+                if (setCurrent) {
+                    QMessageBox::critical(
+                        this, tr("Error"),
+                        tr("The server thinks you are in the server room but your client is unable to display it. "
+                           "Try restarting your client."));
+                }
+                emit roomJoinFailed(roomId);
+                return;
```
RickyRister, line 252: is `healedRoomJoins` ever cleaned up on failure?

BruebachL: It's now cleaned on every terminal path.

cockatrice/src/interface/widgets/tabs/tab_server.cpp
```diff
@@ -185,60 +186,113 @@ void TabServer::processServerMessageEvent(const Event_ServerMessage &event)
...
                                  const CommandContainer & /*commandContainer*/,
                                  const QVariant &extraData,
                                  int roomId)
 {
+    const bool setCurrent = pendingRoomJoins.value(roomId, extraData.toBool());
+    pendingRoomJoins.remove(roomId);
+
     switch (r.response_code()) {
         case Response::RespOk:
+            healedRoomJoins.remove(roomId);
```
RickyRister: nit: you could do something similar to what you did for `setCurrent`, where you save the value to a bool and then remove the entry from the map before the switch. Then we won't have to remember to remove the entry from the map in every branch.

RickyRister (approved):

PR #7261, opened 2026-09-06
[Client] Show confirmation when adding a user to the ignore list

### LINKED ISSUE #1875: Adding a user to ignore list gives no feedback.
When you right click on a user in the user list, and add them to ignore list there's no indication that you successfully ignored them.

It should pop up a desktop notification

BruebachL:
## Related Ticket(s)
- Fixes #1875

Nothing in the UI confirmed that a user had been added to the ignore list, so the action felt ambiguous and could be repeated by accident.

- Show an information dialog when the server acknowledges the add-to-ignore command.


RickyRister (approved):
Works

https://github.com/user-attachments/assets/056d3e98-f951-4fab-9de4-6a3533472f73

PR #7263, opened 2026-09-06
[Game] Don't alert game tab for spectator activity

### LINKED ISSUE #1848: Rooms should not show an alert for spectator activity
OS: OS X 10.11.3

Cockatrice version: Version 323fc06 (2016-02-02)

___
<br>

When spectators start/stop watching on a room, It'd be better if this activity did not trigger an alert (exclamation mark on the room's tab). A lot of the time I will create a game, then switch to another tab in cockatrice or browse the internet with only the top tab of cockatrice showing - that way, when someone joins the game, I can switch back to cockatrice and start playing. When there's an alert for a spectator joining, though, it's basically a false alarm - I switch to cockatrice, see it's not an actual opponent, then go back to whatever else I was doing. It would be much more convenient if the activity of spectators didn't trigger alerts.

BruebachL:
## Related Ticket(s)
- Fixes #1848

Spectator join/leave events were triggering the game tab's attention indicator (exclamation mark and taskbar alert) for players who often leave the tab open while waiting for an opponent. This made spectator activity a false alarm.

- Only emit the user event when an actual player joins or leaves, keeping spectator joins/leaves quiet.

### DIFF
--- a/cockatrice/src/game/game_event_handler.cpp
+++ b/cockatrice/src/game/game_event_handler.cpp
@@ -255,8 +255,6 @@ void GameEventHandler::eventSpectatorLeave(const Event_Leave &event,
     emit spectatorLeft(eventPlayerId);
 
     game->getPlayerManager()->removeSpectator(eventPlayerId);
-
-    emitUserEvent();
 }
 
 void GameEventHandler::eventGameStateChanged(const Event_GameStateChanged &event,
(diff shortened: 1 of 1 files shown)


cockatrice/src/game/game_event_handler.cpp
```diff
@@ -255,8 +255,6 @@ void GameEventHandler::eventSpectatorLeave(const Event_Leave &event,
     emit spectatorLeft(eventPlayerId);
 
     game->getPlayerManager()->removeSpectator(eventPlayerId);
-
-    emitUserEvent();
```
RickyRister, line 259: Do we only use `emitUserEvent` for notification purposes? Could there be any unintended consequences from certain events no longer emitting this signal?

For example, [TabGame::emitUserEvent](https://github.com/Cockatrice/Cockatrice/blob/9b0d62c152b7cd3e95112924904e07984c78d836/cockatrice/src/interface/widgets/tabs/tab_game.cpp#L280), which is triggered by this `emitUserEvent`, calls `updatePlayerListDockTitle`. After this change, the spectator events will no longer trigger the `updatePlayerListDockTitle` call.

PR #7264, opened 2026-09-06
[Client] Fix pawn avatar cache key collision for players without a custom avatar

### LINKED ISSUE #4086: Donator avatar pawn is shown on non donator users as default avatar in games
Sometimes if a regular user is in a game with a donator user and they do not have a custom avatar set they will get the donator pawn as their default avatar instead of the regular pawn.
This could be related to who is hosting the game or something similar, this can be tested locally with a local server setup.
https://cdn.discordapp.com/attachments/401594186983407616/689196020021723169/unknown.png
This was reported in the Discord dev channel by TidalWave#0085 on 2020-03-16 18:38 utc https://discordapp.com/channels/314987288398659595/401594186983407616/689196020441284704

BruebachL:
## Related Ticket(s)
- Fixes #4086

Player pawns are looked up in QPixmapCache under a key built from the rendered size, user level, and the avatar pixmap's cacheKey(). A null pixmap reports cacheKey() 0, so all players without a custom avatar collided: the first pawn rendered for a given size and user level was reused for the next one, showing the wrong player's pawn.

- Extend the key with the rendered height, the lowercased privlevel (matching UserLevelPixmapGenerator), and both pawn colors so that every visually distinct pawn gets its own cache entry.

### DIFF
--- a/cockatrice/src/game_graphics/player/player_target.cpp
+++ b/cockatrice/src/game_graphics/player/player_target.cpp
@@ -132,8 +132,18 @@ void PlayerTarget::paint(QPainter *painter, const QStyleOptionGraphicsItem * /*o
     QRectF translatedRect = painter->combinedTransform().mapRect(avatarBoundingRect);
     QSize translatedSize = translatedRect.size().toSize();
     QPixmap cachedPixmap;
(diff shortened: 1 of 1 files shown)


RickyRister (approved):
Seems logical

PR #7265, opened 2026-09-06
[Client] Add option to use game time in game log timestamps

### LINKED ISSUE #2202: Change game time from local to game

BruebachL:
## Related Ticket(s)
- Fixes #2202

Game log timestamps showed the local clock time, which is meaningless in a replay or in a log read after the fact. Nobody can tell when a play happened relative to the game.

- Add a "Use game time instead of local time in game logs" setting. When enabled, the message log stamps entries with the elapsed game time (HH:MM:SS since the game started)
- regular chat views keep the local clock.


tooomm:
Linking https://github.com/Cockatrice/Cockatrice/pull/5717

cockatrice/src/game_graphics/log/message_log_widget.cpp
RickyRister, line 632: You missed some messages that still use currentTime

Image

RickyRister (changes requested):

cockatrice/src/interface/widgets/settings_page/messages_settings_page.cpp
```diff
@@ -59,6 +59,10 @@ MessagesSettingsPage::MessagesSettingsPage()
     connect(&roomHistory, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().chat(),
             &ChatSettings::setRoomHistory);
 
+    useGameTimeCheckBox.setChecked(SettingsCache::instance().chat().getUseGameTime());
+    connect(&useGameTimeCheckBox, &QCheckBox::QT_STATE_CHANGED, &SettingsCache::instance().chat(),
```
RickyRister, line 63: If you change the setting while you're in the middle of a game, the log will end up with a mix of different time types. Though I don't know how else it could possibly behave, so I'm fine with that being the intended behavior. It's the player's fault for messing with the settings like that.

https://github.com/user-attachments/assets/72059bdd-ef4a-4bc0-8e79-00d80db29ed9

PR #7266, opened 2026-09-06
[Game] Preserve hand card order when restoring connection

### LINKED ISSUE #2759: Restoring connection reverses hand
<b>OS:</b>
MacOS (unknown version)

<b>Cockatrice version:</b>
2.3.18

__________

When restoring connection to a game, the order of the cards in your hand is reversed.

(reported to me by my brother, so I don't have much more detail)

BruebachL:
## Related Ticket(s)
- Fixes #2759

When a player restores connection to a game, the client rebuilds each zone from the cards the server sends in the game state. Non-coordinate zones (hand, piles, stack) report x == 0 on every card, so inserting each rebuilt card at that coordinate reversed the received order one card at a time.

- Append rebuilt cards in the order they arrive for zones without coordinates; coordinate-based zones (table) keep using x/y.

### DIFF
--- a/cockatrice/src/game/player/player_logic.cpp
+++ b/cockatrice/src/game/player/player_logic.cpp
@@ -175,7 +175,15 @@ void PlayerLogic::processPlayerInfo(const ServerInfo_Player &info)
                 const ServerInfo_Card &cardInfo = zoneInfo.card_list(j);
                 auto *card = new CardItem(this);
                 card->processCardInfo(cardInfo);
-                zone->addCard(card, false, cardInfo.x(), cardInfo.y());
+                // Zones without coordinates (hand, piles, stack) preserve the order
+                // they arrive in on the server in the positions of their cards list.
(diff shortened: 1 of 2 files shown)


RickyRister:
Note that this only works perfectly if you haven't sorted your hand. If you've done any hand sorting, then the order upon rejoin is not guaranteed to be preserved.

It's probably due to hand sorting's interaction with the server not being well-defined, so over time as people added more code, the hand sorting information only sometimes makes it to the server.

RickyRister:
Huh, I never realized, but currently the stack gets reversed upon rejoin. (Which this PR also fixes)

https://github.com/user-attachments/assets/73e6ec1d-a31e-4611-804c-f6c5593d8ec9

I'm surprised nobody has reported that, because that's a bug with massive gameplay implications, much more major than hand order being reversed. It causes two players literally disagree on the state of the game. I guess people just don't rejoin while in the middle of a big stack battle?

RickyRister (approved):

PR #7268, opened 2026-09-06
[Game] Show loaded deck messages to players joining a game

### LINKED ISSUE #552: Players don't always know sideboard count of other player's decks
If player A loads a deck, and then player B joins the room after that, player B never sees the "Player A loaded a deck with 15 sideboard cards" message.

On joining a game, Player B should see such a message for each player with a loaded deck.

BruebachL:
## Related Ticket(s)
- Fixes #552

The "Player X has loaded a deck" message is only broadcast when a deck is selected, so players and spectators who join a game afterwards never see it.

On game join, replay a deck-select event context privately to the joining participant for every player that already has a deck loaded, reusing the existing Event_PlayerPropertiesChanged + Context_DeckSelect mechanism so no protocol version bump is needed.

- Server_AbstractPlayer: expose the sideboard-locked state
- Server_Game: synthesize deck-select contexts in createGameJoinedEvent

### DIFF
--- a/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.h
+++ b/libcockatrice_network/libcockatrice/network/server/remote/game/server_abstract_player.h
@@ -47,6 +47,10 @@ public:
     {
         return deck;
     }
+    bool getSideboardLocked() const
+    {
+        return sideboardLocked;
+    }
     bool getReadyStart() const
     {
         return readyStart;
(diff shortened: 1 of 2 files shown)


RickyRister:
Are you sure players won't be confused to see a barrage of deck loaded messages in their logs whenever they join an in-progress game or rejoin a game after disconnecting?

PR #7269, opened 2026-09-06
[Chat] Render room chat history usernames as live user tags

### LINKED ISSUE #1595: Chat history usernames should be user tags for online users
This would be useful for moderators, if they're offline and join to see someone disrupting the chat

BruebachL:
## Related Ticket(s)
- Fixes #1595

Room chat history carries no user-level data, so history usernames were rendered as fixed, zero-level tags: the moderation context menu was missing the buddy/ignore and promote/demote entries and the stored name casing was never corrected.

Resolve each history author against the online user list and, when found, build the user tag with the real user level and name so the entry behaves exactly like a live chat tag. Offline users keep the plain fallback.

- chat_view: look up history authors via getOnlineUser for the real level/name

### DIFF
--- a/cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
+++ b/cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
@@ -340,16 +340,26 @@ void ChatView::appendMessage(QString message,
                 pos.relativePosition = match.captured(0).length(); // set message start
                 auto before = match.captured(1);
                 auto sentBy = match.captured(2);
+
+                // The user level is not carried in the room chat history, so history
+                // entries used to render as fixed-level user tags. Resolve online users
+                // against the user list to turn their history entries into full user
+                // tags (correct level, name casing and moderation context menu).
+                QString displayName = sentBy;
+                QString levelMarker = "_"; // placeholder for offline / unknown user level
+                if (const ServerInfo_User *onlineUser = userListProxy->getOnlineUser(sentBy)) {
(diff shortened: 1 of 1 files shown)


cockatrice/src/interface/widgets/server/chat_view/chat_view.cpp
```diff
@@ -340,16 +340,26 @@ void ChatView::appendMessage(QString message,
...
+                    levelMarker = QString::number(onlineUser->user_level());
+                }
+
                 cursor.insertText(before); // add message timestamp
                 QTextCharFormat senderFormat(defaultFormat);
                 senderFormat.setAnchor(true);
-                // this underscore is important, it is used to add the user level, but in this case the level is
-                // unknown, if the name contains an underscore it would split up the name
-                senderFormat.setAnchorHref("user://_" + sentBy);
+                senderFormat.setAnchorHref("user://" + levelMarker + "_" + displayName);
```
RickyRister, line 358: Doesn't handle offline users properly. Cockatrice will think you're trying to interact with a user named `_USERNAME`, which then causes all actions against that user to fail because that user doesn't exist. 

https://github.com/user-attachments/assets/7e978ab5-f608-48db-8f01-7b5700d930be

BruebachL: Fixed, defaulting them to user level 0.

RickyRister (approved):
works now

PR #7260, opened 2026-09-06
[Client] Add setting to ignore all private messages

BruebachL:
## Related Ticket(s)
- Fixes #1250

## Short roundup of the initial problem
Users don't want to receive any direct messages at all.

## What will change with this Pull Request?
- Add a new option allowing just that
- Exempt Moderators and Admins from this check

PR #7271, opened 2026-09-07
[Settings] Make chat settings page scrollable

RickyRister:
### Related Ticket(s)
- Follow-up to #7260

Chat settings page is starting to become really cramped after adding more settings recently.

Screenshot 2026-09-06 at 11 53 21 PM

Make the chat settings page scrollable

https://github.com/user-attachments/assets/88bad490-f513-4fe9-8594-30b1957034fa


BruebachL:
I've thought about this before but have we considered promoting settings from a modal to its own tab? We would gain significantly more space in most scenarios imo.

RickyRister:
Main issue would be that most users would expect the settings to be a modal because that's what they're used to. But it could be something to consider.

BruebachL (approved):

PR #7272, opened 2026-09-07
[Oracle] Prefer higher-priority maintype when merging split cards

### LINKED ISSUE #4394: adventure cards have their maintype selected inconsistently
w10

cockatrice bug

select a bonecrusher giant in the deck editor

Thanks!

BruebachL:
## Related Ticket(s)
- Fixes #4394

Adventure cards (e.g. Bonecrusher Giant, Virtue of Knowledge) are stored as split cards in MTGJSON, with a Creature/permanent face and an Instant/Sorcery adventure face. When the two faces are merged into a single card, the code previously discarded the second face's maintype entirely, keeping whichever face was processed first.

If the Instant/Sorcery adventure face appeared first, the merged card got maintype 'Instant' with tableRow 3. At runtime this made double-clicking the card on the stack send it to the graveyard instead of the table (PlayerActions::playCard).

Fix the merge to follow the same priority order used by getMainCardType() (Planeswalker > Creature > Land > Sorcery > Instant > Artifact > Enchantment), so the permanent Creature type wins for adventure cards regardless of face order, matching the physical card and the reported expectations.

### DIFF
--- a/oracle/src/oracleimporter.cpp
+++ b/oracle/src/oracleimporter.cpp
@@ -460,8 +460,17 @@ int OracleImporter::importCardsFromSet(const CardSetPtr &currentSet, const QJson
                             properties.insert(prop, thisCardPropertyValue);
                         } else if (prop == "colors" || prop == "coloridentity") { // the card is both colors
                             properties.insert(prop, originalPropertyValue + thisCardPropertyValue);
-                        } else if (prop == "maintype") { // don't create maintypes with //es in them
-                            continue;
+                        } else if (prop == "maintype") {
+                            // Use the same priority as getMainCardType() to pick the
+                            // "best" type across faces — e.g. Creature over Instant
+                            // for adventure cards.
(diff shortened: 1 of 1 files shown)


oracle/src/oracleimporter.cpp
```diff
@@ -96,16 +96,34 @@ bool OracleImporter::readSetsFromByteArray(QByteArray data)
     return true;
 }
 
+/**
+ * The priority order used to pick a card's main type when a card has multiple
+ * types (e.g. "Artifact Creature") or multiple faces (e.g. split/adventure cards).
+ * A lower index means a higher priority.
+ */
+static const QStringList &mainCardTypePriority()
```
RickyRister: nit: since this is in a `cpp` file, you technically can just define `static const QStringList` outside of a function as a constant.

RickyRister (approved):

PR #7273, opened 2026-09-07
[Game] Stop duplicate attach log entries

BruebachL:
Fix duplicate attach messages in the game log caused by a redundant child arrow in CardItem::drawAttachArrow.

Unlike the sibling drawArrow, drawAttachArrow omitted the ``card == this`` guard when iterating selectedItems(). Because right-clicking a card to open the attach menu selects that card, it was always present in selectedItems(), producing a second arrow for the same source card.

On release both arrows sent an identical Command_AttachCard, so the server broadcast two Event_AttachCard messages and the log rendered "attaches to" twice.

- Mirror the drawArrow skip condition so the active card is excluded and only one attach command is sent.


RickyRister (approved):
Thanks for the fix

PR #7275, opened 2026-09-08
[Theme] Add scheme-variant asset resolution

BruebachL:

Theme assets (`logo`, `backgrounds/home`, `icons/*`, …) always resolve to a single plain file, so a dark scheme can only be approximated by re-coloring the same bitmap at load time. There is no way to ship real `-dark`/`-light` artwork pairs.

- Add `ThemeManager::assetPath()` and `schemeVariantPath()`: resolve a theme asset stem to a `-dark`/`-light` variant file (`.png`/`.jpg`/`.jpeg`/`.svg`, raster wins, theme dir before the built-in resources), falling back to the plain asset.
- Wire `themePixmap()`, `loadBrush()` and `loadExtraBrush()` through the new resolution so scheme-sensitive images pick variants automatically.
- Load `style-dark.css`/`style-light.css` in preference to `style.css` when present.
- Re-resolve the home widget's background and overlay on `themeChanged`.
- Link `pixel_map_generator.cpp` into the oracle target and add `Qt6::Xml` to its modules (`QDomDocument`).

Behavior is unchanged for themes that don't ship variants — the plain asset is the implicit fallback.

### DIFF
--- a/cockatrice/src/interface/pixel_map_generator.cpp
+++ b/cockatrice/src/interface/pixel_map_generator.cpp
@@ -1,5 +1,7 @@
 #include "pixel_map_generator.h"
 
+#include "theme_manager.h"
+
 #include <QApplication>
 #include <QDomDocument>
 #include <QFile>
@@ -82,7 +84,13 @@ static QPixmap loadSvg(const QString &svgPath, const QSize &size, bool expandOnl
 /**
  * Try to load path image from non-SVG formats, otherwise fall back to SVG.
  * This is to allow custom themes to support non-SVG format type overrides, since SVG requires custom loading.
- * @param path The path to the file, with no file extension. File formats will be automatically detected.
+ *
+ * The path may already carry the resolved file extension (e.g. via
+ * ThemeManager::assetPath); such paths are loaded directly. Otherwise a
+ * format-agnostic lookup probes png, jpg and finally svg.
+ *
+ * @param path The path to the file, with no file extension unless the caller
+ * already resolved it. File formats will be automatically detected.
  * @param size The desired size of the pixmap.
  * @param expandOnly If true, then keep the size of the initial pixmap to at least the size (Only relevant if SVG).
  *
@@ -90,6 +98,19 @@ static QPixmap loadSvg(const QString &svgPath, const QSize &size, bool expandOnl
  */
 static QPixmap tryLoadImage(const QString &path, const QSize &size, bool expandOnly = false)
 {
+    if (path.endsWith(QLatin1String(".svg"), Qt::CaseInsensitive)) {
+        return loadSvg(path, size, expandOnly);
+    }
+    if (path.endsWith(QLatin1String(".png"), Qt::CaseInsensitive) ||
+        path.endsWith(QLatin1String(".jpg"), Qt::CaseInsensitive) ||
+        path.endsWith(QLatin1String(".jpeg"), Qt::CaseInsensitive)) {
+        QPixmap pix(path);
+        if (!pix.isNull()) {
+            return pix.scaled(size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+        }
+        return {};
+    }
+
     const auto formats = {"png", "jpg"};
 
     QPixmap returnPixmap;
@@ -111,7 +132,8 @@ QPixmap PhasePixmapGenerator::generatePixmap(int height, QString name)
         return pmCache.value(key);
     }
 
-    QPixmap pixmap = tryLoadImage("theme:phases/" + name, QSize(height, height));
+    QPixmap pixmap = tryLoadImage(QStringLiteral("theme:") + themeManager->assetPath(QStringLiteral("phases/") + name),
+                                  QSize(height, height));
 
     pmCache.insert(key, pixmap);
     return pixmap;
@@ -396,7 +418,8 @@ QPixmap LockPixmapGenerator::generatePixmap(int height)
         return pmCache.value(key);
     }
 
-    QPixmap pixmap = tryLoadImage("theme:icons/lock", QSize(height, height), true);
+    QPixmap pixmap = tryLoadImage(QStringLiteral("theme:") + themeManager->assetPath(QStringLiteral("icons/lock")),
+                                  QSize(height, height), true);
     pmCache.insert(key, pixmap);
     return pixmap;
 }
@@ -411,7 +434,8 @@ QPixmap DropdownIconPixmapGenerator::generatePixmap(int height, bool expanded)
     }
 
     QString name = expanded ? "dropdown_expanded" : "dropdown_collapsed";
-    QPixmap pixmap = tryLoadImage("theme:icons/" + name, QSize(height, height), true);
+    QPixmap pixmap = tryLoadImage(QStringLiteral("theme:") + themeManager->assetPath(QStringLiteral("icons/") + name),
+                                  QSize(height, height), true);
 
     pmCache.insert(key, pixmap);
     return pixmap;
@@ -472,6 +496,13 @@ QHash<QString, QPixmap> ManaSymbolPixmapGenerator::scaledCache;
 
 QPixmap loadColorAdjustedPixmap(const QString &name)
 {
+    // Prefer an authored scheme-qualified variant when one exists for this asset.
+    const QString variant = themeManager->schemeVariantPath(QStringView(name).mid(QStringLiteral("theme:").size()));
+    if (!variant.isEmpty()) {
+        return QPixmap(QStringLiteral("theme:") + variant);
+    }
+
+    // Legacy fallback: runtime-invert for dark mode when no authored variant.
     if (qApp->palette().windowText().color().lightness() > 200) {
         QImage img(name);
         img.invertPixels();
@@ -481,4 +512,22 @@ QPixmap loadColorAdjustedPixmap(const QString &name)
     } else {
         return QPixmap(name);
     }
+}
+
+QPixmap themePixmap(QStringView prefix)
+{
+    const QString resolved = themeManager->assetPath(prefix);
+    return QPixmap(QStringLiteral("theme:") + resolved);
+}
+
+void clearPixmapGeneratorCaches()
+{
+    PhasePixmapGenerator::clear();
+    CounterPixmapGenerator::clear();
+    PingPixmapGenerator::clear();
+    CountryPixmapGenerator::clear();
+    UserLevelPixmapGenerator::clear();
+    LockPixmapGenerator::clear();
+    DropdownIconPixmapGenerator::clear();
+    ManaSymbolPixmapGenerator::clear();
 }
--- a/cockatrice/src/interface/theme_manager.cpp
+++ b/cockatrice/src/interface/theme_manager.cpp
@@ -144,7 +144,8 @@ bool ThemeManager::isDarkMode(const QString &themeDirPath) const
 
 QString ThemeManager::schemeVariantPath(QStringView prefix) const
 {
-    static const QStringList formats = {QStringLiteral(".png"), QStringLiteral(".jpg"), QStringLiteral(".jpeg"), QStringLiteral(".svg")};
+    static const QStringList formats = {QStringLiteral(".png"), QStringLiteral(".jpg"), QStringLiteral(".jpeg"),
+                                        QStringLiteral(".svg")};
     const QString scheme = isDarkMode(currentThemePath) ? QStringLiteral("dark") : QStringLiteral("light");
     const QString variantStem = prefix.toString() + QLatin1Char('-') + scheme;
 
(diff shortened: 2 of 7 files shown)


cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -20,7 +21,8 @@
 #include <libcockatrice/settings/paths_settings.h>
 
 HomeWidget::HomeWidget(QWidget *parent, TabSupervisor *_tabSupervisor)
-    : QWidget(parent), tabSupervisor(_tabSupervisor), background("theme:backgrounds/home"), overlay("theme:cockatrice")
+    : QWidget(parent), tabSupervisor(_tabSupervisor), background(themePixmap(QStringLiteral("backgrounds/home"))),
```
RickyRister, line 24: is it possible to add a `char*` overload for `themePixmap` so that we don't have to surround every call with `QStringLiteral`?

Alternatively, apparently you can use [`_s`](https://doc.qt.io/qt-6/qstring.html#operator-22-22_s) as an alias for `QStringLiteral`

cockatrice/src/interface/pixel_map_generator.cpp
```diff
@@ -82,14 +84,33 @@ static QPixmap loadSvg(const QString &svgPath, const QSize &size, bool expandOnl
...
  * @param expandOnly If true, then keep the size of the initial pixmap to at least the size (Only relevant if SVG).
  *
  * @return The loaded image into a Pixmap with the given size, or an empty Pixmap if the loading failed.
  */
 static QPixmap tryLoadImage(const QString &path, const QSize &size, bool expandOnly = false)
 {
+    if (path.endsWith(QLatin1String(".svg"), Qt::CaseInsensitive)) {
+        return loadSvg(path, size, expandOnly);
+    }
+    if (path.endsWith(QLatin1String(".png"), Qt::CaseInsensitive) ||
```
ZeldaZach, line 104: The raster branch returns `{}` on a failed load, which drops the SVG fallback the old code relied on.

Scenario: a theme ships a truncated/corrupt `icons/lock.png`. `assetPath()` sees the file exists and returns `icons/lock.png`, so this branch runs `QPixmap pix("theme:icons/lock.png")`, gets a null pixmap, and returns empty — the lock icon disappears. Previously `returnPixmap.load(path, "png")` failed, `"jpg"` failed, and the function fell through to `loadSvg(path + ".svg")`, which resolved the built-in `:/resources/icons/lock.svg` through the `theme:` search paths and rendered fine.

The same applies to `phases/*` and the dropdown icons, the other callers now routed through `assetPath()`. Falling through to the legacy probe (or explicitly trying `loadSvg(stem + ".svg")`) when the raster load returns null would preserve the old behavior.

cockatrice/src/interface/pixel_map_generator.cpp
```diff
@@ -472,6 +496,13 @@ QHash<QString, QPixmap> ManaSymbolPixmapGenerator::scaledCache;
 
 QPixmap loadColorAdjustedPixmap(const QString &name)
 {
+    // Prefer an authored scheme-qualified variant when one exists for this asset.
+    const QString variant = themeManager->schemeVariantPath(QStringView(name).mid(QStringLiteral("theme:").size()));
```
ZeldaZach, line 500: No `isNull()` guard on the variant, and no fallback to the plain asset.

`schemeVariantPath()` only checks that the file *exists* — not that Qt can decode it. If a theme ships a corrupt/zero-byte `icons/search-dark.png`, or an `icons/search-dark.svg` on a build where the `qsvg` imageformat plugin isn't deployed, the `QPixmap` here is null and the search icon renders as nothing. The plain `icons/search` asset that used to work is never tried, because this returns before reaching the legacy block.

Checking the result and only returning it when non-null, otherwise falling through, would keep the plain asset as the real fallback.

Two related notes on this function:

- `QStringView(name).mid(QStringLiteral("theme:").size())` silently assumes every caller passes a `theme:`-prefixed path. All current callers do, but any other path gets its first 6 characters chopped off and probed as a stem.
- Results of this function are cached by callers at construction time (`player_list_widget.cpp:59-61`, the `QLineEdit::addAction` search icons in the deck editor / visual database / view-zone widgets), and `clearPixmapGeneratorCaches()` does not reach them — so authored variants for those icons won't switch until the app restarts.

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -140,6 +142,48 @@ bool ThemeManager::isDarkMode(const QString &themeDirPath) const
     }
 }
 
+QString ThemeManager::schemeVariantPath(QStringView prefix) const
```
ZeldaZach, line 145: The variant probe runs over the merged `theme:` search path rather than per search-path root, so a variant in a lower-priority root can beat a plain asset that the theme explicitly overrides.

Concrete scenario once `-dark`/`-light` art lands in `:/resources` (the natural follow-up to this PR): built-in ships `icons/lock-dark.svg`; a user theme ships its own `icons/lock.svg` and no variants. In dark mode `schemeVariantPath("icons/lock")` finds the built-in `lock-dark.svg` and `assetPath()` returns it, so the theme's own override is silently ignored — the opposite of the "theme dir before the built-in resources" precedence the PR description promises.

Resolving root by root would give the intended precedence: for each entry of `QDir::searchPaths("theme")`, try `<stem>-<scheme>.<ext>` then `<stem>.<ext>`, and only move to the next root if neither hits.

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -140,6 +142,48 @@ bool ThemeManager::isDarkMode(const QString &themeDirPath) const
...
+{
+    static const QStringList formats = {QStringLiteral(".png"), QStringLiteral(".jpg"), QStringLiteral(".jpeg"),
+                                        QStringLiteral(".svg")};
+    const QString scheme = isDarkMode(currentThemePath) ? QStringLiteral("dark") : QStringLiteral("light");
```
ZeldaZach, line 149: `isDarkMode()` calls `ThemeConfig::fromThemeDir()`, which opens and line-parses `theme.cfg` off disk on every invocation (`theme_config.cpp:23`, no caching). Because `assetPath()` delegates here first, every single asset resolution now costs one file open plus a full parse, on top of up to 8 `QFileInfo::exists()` stats multiplied by the number of registered `theme:` search paths.

That sits on the path of `loadBrush`/`loadExtraBrush`, `themePixmap`, `loadColorAdjustedPixmap` (once per search-bar widget construction) and every pixmap-generator cache miss. `themeChangedSlot()` already computes the same value as `activeScheme`; caching it on `ThemeManager` and reading the member here would avoid the repeated I/O.

cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -20,7 +21,8 @@
...
+      overlay(themePixmap(QStringLiteral("cockatrice")))
```
ZeldaZach, line 25: `overlay` is resolved exactly once here and never re-resolved, so the scheme-variant logo never actually switches at runtime.

Two reasons: nothing reassigns `overlay` on `themeChanged` (`initializeBackgroundFromSource()` only touches `background`), and even if it did, `createButtons()` sets `logoLabel->setPixmap(overlay.scaledToWidth(200, ...))` once at line 238 and keeps no pointer to that label. So with a `cockatrice-dark.png` / `cockatrice-light.png` pair shipped, flipping the OS color scheme or switching themes leaves whichever logo was active at HomeWidget construction. The PR description says the overlay is re-resolved on `themeChanged`, but it isn't.

cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -56,6 +58,9 @@ HomeWidget::HomeWidget(QWidget *parent, TabSupervisor *_tabSupervisor)
             &HomeWidget::initializeBackgroundFromSource);
     connect(&SettingsCache::instance(), &SettingsCache::themeChanged, this,
             &HomeWidget::updateButtonsToBackgroundColor);
+    // Scheme flips (light/dark/system with an OS switch) fire on themeManager,
+    // not on SettingsCache::themeChanged, so re-resolve the variant background.
+    connect(themeManager, &ThemeManager::themeChanged, this, &HomeWidget::initializeBackgroundFromSource);
```
ZeldaZach, line 63: This double-fires `initializeBackgroundFromSource()` on every theme change.

`ThemeManager::themeChangedSlot()` is itself a `SettingsCache::themeChanged` handler and emits `ThemeManager::themeChanged` at its end, so one theme switch now runs `initializeBackgroundFromSource()` twice: once via the existing `SettingsCache::themeChanged` connection on line 57, once via this new one.

For `BackgroundSources::RandomCardArt` / `DeckFileArt` that means `updateRandomCard()` picks a new random card twice, `loadBackgroundSourceDeck()` re-reads `background.cod` from disk twice, and `onBackgroundShuffleFrequencyChanged()` restarts the timer twice per theme change. `updateRandomCard()` also does an unguarded `connect(newCard.getCardPtr().data(), &CardInfo::pixmapUpdated, ...)` on each call, so the duplicate invocation accumulates an extra connection and `updateBackgroundProperties()` starts firing more than once per pixmap update.

Only the `Theme` source cares about a scheme flip, so this connection could target a small slot that re-resolves `background` (and `overlay`) when the source is `BackgroundSources::Theme`, instead of re-running the whole initializer.

ZeldaZach (changes requested):
The direction here is good and the mechanism is a sensible foundation for the rest of the stack, but there are a few correctness issues worth fixing first. See the inline comments for details.

The main ones: the new raster early-return in `tryLoadImage` drops the SVG fallback when a `.png` fails to decode, so a bad theme file now makes the icon disappear instead of falling back; `loadColorAdjustedPixmap` has no `isNull()` guard on the resolved variant; and `schemeVariantPath` probes the merged `theme:` search path instead of checking each root in turn, so a built-in `X-dark.*` can win over a theme's own plain `X.*`, which inverts the precedence the PR description promises. That last one is latent until variant assets actually ship, which is exactly what the follow-up PRs do.

Separately, the home widget logo is resolved once in the constructor and never re-resolved on `themeChanged`, so the logo variant does not actually switch at runtime, and the new `themeManager::themeChanged` connection double-fires `initializeBackgroundFromSource()` since that slot already runs off `SettingsCache::themeChanged`.

Also flagged a minor perf note: `isDarkMode()` re-parses `theme.cfg` on every call, and every asset resolution now goes through it.

cockatrice/src/interface/widgets/general/home_widget.cpp
BruebachL: Done in the base PR (0e72bef0e8) — the const char* overload for themePixmap is already present in pixel_map_generator.h.

cockatrice/src/interface/pixel_map_generator.cpp
BruebachL: Fixed — tryLoadImage now falls through to loadSvg(stem + ".svg") when the raster decode returns null, restoring the built-in SVG fallback path for corrupt .png theme assets.

cockatrice/src/interface/pixel_map_generator.cpp
BruebachL: Fixed — added an isNull() guard on the resolved variant; falls back to the legacy invert block when the decode fails, keeping the plain asset as the final fallback.

cockatrice/src/interface/theme_manager.cpp
BruebachL: Fixed — schemeVariantPath now resolves per search-path root (try stem+scheme, then stem, before moving to the next root), preserving "theme dir beats built-in resources" precedence.

cockatrice/src/interface/theme_manager.cpp
BruebachL: Fixed — isDarkModeCache is set once in themeChangedSlot() right after currentThemePath is assigned; isDarkModeActive() returns the cache, avoiding repeated theme.cfg parsing on every asset resolution.

cockatrice/src/interface/widgets/general/home_widget.cpp
BruebachL: Fixed — added a QLabel *logoLabel member; overlay and logo are re-resolved in updateSchemeVariantAssets() on ThemeManager::themeChanged so the logo variant flips at runtime.

cockatrice/src/interface/widgets/general/home_widget.cpp
BruebachL: Fixed — replaced the SettingsCache::themeChanged connection with ThemeManager::themeChanged, targeting the new updateSchemeVariantAssets() slot which only runs when background source is Theme, eliminating the double-fire.

BruebachL:
All review comments addressed in c1eddbc00f (force-pushed):

- **tryLoadImage SVG fallback** (pixel_map_generator.cpp): raster failures now fall through to loadSvg(stem + ".svg"), restoring the built-in SVG fallback.
- **loadColorAdjustedPixmap** (pixel_map_generator.cpp): added isNull() guard on the variant; falls back to the legacy invert block.
- **schemeVariantPath per-root** (theme_manager.cpp): resolves per search-path root preserving the "theme dir beats built-in resources" precedence.
- **isDarkMode caching** (theme_manager.h/cpp): added isDarkModeCache member set once per themeChanged; avoids repeated theme.cfg parsing.
- **Overlay + logo re-resolve** (home_widget.h/cpp): logoLabel member added; updateSchemeVariantAssets() re-resolves both on ThemeManager::themeChanged.
- **Double-fire fix** (home_widget.cpp): replaced SettingsCache::themeChanged connection with ThemeManager::themeChanged targeting updateSchemeVariantAssets(), running only for Theme source.

ZeldaZach (approved):

PR #7276, opened 2026-09-08
[Theme] Migrate theme asset loads to scheme-variant resolution

BruebachL:

Loads across the client still use bare `QPixmap("theme:…")`, bypassing the scheme-variant resolution introduced in the previous PR.

- Migrate the remaining 91 `QPixmap("theme:…")` call sites to `themePixmap()` so icons, replay controls, card fronts/backs and other images resolve to their `-dark`/`-light` variants.
- Stem-exact 1:1 migration; bitmap output for non-variant themes is identical to before.

### DIFF
--- a/cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
+++ b/cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
@@ -6,6 +6,7 @@
 #include "dlg_settings.h"
 
 #include "../../../client/settings/cache_settings.h"
+#include "../../pixel_map_generator.h"
 #include "../main.h"
 #include "../settings_page/appearance_settings_page.h"
 #include "../settings_page/deck_editor_settings_page.h"
@@ -96,7 +97,7 @@ void DlgSettings::setupUi()
     // Search bar
     searchEdit = new QLineEdit;
     searchEdit->setClearButtonEnabled(true);
-    searchEdit->addAction(QPixmap("theme:icons/search"), QLineEdit::LeadingPosition);
+    searchEdit->addAction(themePixmap(QStringLiteral("icons/search")), QLineEdit::LeadingPosition);
     searchEdit->installEventFilter(this);
     connect(searchEdit, &QLineEdit::textChanged, this, &DlgSettings::onSearchTextChanged);
 
--- a/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_filter_toolbar_widget.cpp
+++ b/cockatrice/src/interface/widgets/visual_database_display/visual_database_display_filter_toolbar_widget.cpp
@@ -1,5 +1,6 @@
 #include "visual_database_display_filter_toolbar_widget.h"
 
+#include "../../pixel_map_generator.h"
 #include "../deck_editor/card_database_view.h"
 #include "visual_database_display_widget.h"
 
@@ -60,22 +61,22 @@ VisualDatabaseDisplayFilterToolbarWidget::VisualDatabaseDisplayFilterToolbarWidg
     });
 
     quickFilterSaveLoadWidget = new SettingsButtonWidget(this);
-    quickFilterSaveLoadWidget->setButtonIcon(QPixmap("theme:icons/floppy_disk"));
+    quickFilterSaveLoadWidget->setButtonIcon(themePixmap(QStringLiteral("icons/floppy_disk")));
 
     quickFilterNameWidget = new SettingsButtonWidget(this);
-    quickFilterNameWidget->setButtonIcon(QPixmap("theme:icons/pen_to_square"));
+    quickFilterNameWidget->setButtonIcon(themePixmap(QStringLiteral("icons/pen_to_square")));
 
     quickFilterMainTypeWidget = new SettingsButtonWidget(this);
-    quickFilterMainTypeWidget->setButtonIcon(QPixmap("theme:icons/circle_half_stroke"));
+    quickFilterMainTypeWidget->setButtonIcon(themePixmap(QStringLiteral("icons/circle_half_stroke")));
 
     quickFilterSubTypeWidget = new SettingsButtonWidget(this);
-    quickFilterSubTypeWidget->setButtonIcon(QPixmap("theme:icons/dragon"));
+    quickFilterSubTypeWidget->setButtonIcon(themePixmap(QStringLiteral("icons/dragon")));
 
     quickFilterSetWidget = new SettingsButtonWidget(this);
-    quickFilterSetWidget->setButtonIcon(QPixmap("theme:icons/scroll"));
+    quickFilterSetWidget->setButtonIcon(themePixmap(QStringLiteral("icons/scroll")));
 
     quickFilterFormatLegalityWidget = new SettingsButtonWidget(this);
-    quickFilterFormatLegalityWidget->setButtonIcon(QPixmap("theme:icons/scale_balanced"));
+    quickFilterFormatLegalityWidget->setButtonIcon(themePixmap(QStringLiteral("icons/scale_balanced")));
 
     retranslateUi();
 }
--- a/cockatrice/src/interface/window_main.cpp
+++ b/cockatrice/src/interface/window_main.cpp
@@ -44,6 +44,7 @@
 #include "intents/intent_open_server_room_by_name.h"
 #include "intents/url_parser.h"
 #include "logger.h"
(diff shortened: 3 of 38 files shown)


cockatrice/src/interface/window_main.cpp
```diff
@@ -816,7 +818,7 @@ void MainWindow::createTrayIcon()
 
     trayIcon = new QSystemTrayIcon(this);
     trayIcon->setContextMenu(trayIconMenu);
-    trayIcon->setIcon(QPixmap("theme:cockatrice"));
+    trayIcon->setIcon(themePixmap(QStringLiteral("cockatrice")));
```
ZeldaZach, line 821: Scheme variants are resolved once, at widget-construction time, and nothing re-resolves them when the color scheme flips at runtime.

`themePixmap()` picks `-dark`/`-light` via `ThemeManager::isDarkMode()`, which for a theme with `colorScheme = Auto` follows `QStyleHints::colorScheme()`. `ThemeManager` already reacts to `QStyleHints::colorSchemeChanged` (theme_manager.cpp:108) by re-applying the palette, clearing `QPixmapCache`, and emitting `themeChanged` -- but none of the ~90 call sites migrated in this PR listen to that signal. `createTrayIcon()` here runs once; `MainWindow::retranslateUi()` (line 328, `aSettings`) only runs on `QEvent::LanguageChange`; the deck-editor `QAction`/`QToolButton` icons are all set in constructors.

Scenario: a theme ships `cockatrice-light.png` and `cockatrice-dark.png` with `colorScheme = Auto`. The app starts in light mode and loads the light variant. The user flips the OS to dark mode: palette and zone brushes update, but the tray icon, the Settings menu icon, and every migrated toolbar/action icon keep the light-mode artwork until restart -- i.e. exactly the wrong-contrast icons this feature exists to fix.

Before this PR these loads were scheme-agnostic, so staleness was invisible; now it is user-visible. Worth connecting the affected widgets to `ThemeManager::themeChanged`, or introducing a helper that re-applies the icon on that signal.

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
```diff
@@ -96,7 +97,7 @@ void DlgSettings::setupUi()
     // Search bar
     searchEdit = new QLineEdit;
     searchEdit->setClearButtonEnabled(true);
-    searchEdit->addAction(QPixmap("theme:icons/search"), QLineEdit::LeadingPosition);
+    searchEdit->addAction(themePixmap(QStringLiteral("icons/search")), QLineEdit::LeadingPosition);
```
ZeldaZach, line 100: Missed migration in this same file (comment anchored here because the affected lines are outside the diff).

The settings dialog's eight sidebar page icons are still loaded bare:

- `dlg_settings.cpp:230` -- `tabButton->setIcon(QPixmap(iconResources[i]));`
- `dlg_settings.cpp:162` -- same load inside the `themeChanged` handler
- `settings_search_delegate.cpp:25` -- `pageIcons.append(QPixmap(resource));`

So after this PR the search icon at line 100 resolves `icons/search-dark`/`-light`, while `config/general`, `config/appearance`, ... sitting right next to it in the same dialog do not. A theme that ships `config/*-dark.svg` gets a correctly-inverted search glyph and eight uninverted sidebar glyphs.

These three are the cheapest ones to fix, since `DlgSettings` already has a refresh hook that reloads them. Note that hook is wired to `SettingsCache::themeChanged`, not `ThemeManager::themeChanged`, so it does not fire on an OS dark-mode flip -- worth switching it while migrating.

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_filter_toolbar_widget.cpp
```diff
@@ -60,22 +61,22 @@ VisualDatabaseDisplayFilterToolbarWidget::VisualDatabaseDisplayFilterToolbarWidg
     });
 
     quickFilterSaveLoadWidget = new SettingsButtonWidget(this);
-    quickFilterSaveLoadWidget->setButtonIcon(QPixmap("theme:icons/floppy_disk"));
+    quickFilterSaveLoadWidget->setButtonIcon(themePixmap(QStringLiteral("icons/floppy_disk")));
```
ZeldaZach, line 64: Two loads at the top of this same constructor were missed, so one asset is now resolved two different ways inside one toolbar (anchored here since lines 20-26 are outside the diff):

```cpp
// line 20
auto scalePixmap = [](const QString &fileName) { return QIcon(QPixmap(fileName)).pixmap({20, 20}); };
// line 23
sortByLabel->setPixmap(scalePixmap("theme:icons/sort_arrow_down"));
// line 26
filterByLabel->setPixmap(scalePixmap("theme:icons/filter"));
```

The grep for `QPixmap("theme:` did not catch them because the literal is passed through the lambda.

Concrete effect: with a theme shipping `icons/sort_arrow_down-dark.svg`, `VisualDeckDisplayOptionsWidget` (`sortCriteriaButton`, migrated in this PR) shows the dark variant while `sortByLabel` here shows the plain one -- the same glyph rendered two ways in adjacent widgets. `icons/filter` never gets a variant at all.

ZeldaZach (approved):
The mechanical migration checks out. I diffed all 94 old asset strings against the new ones and they match byte for byte, the 22 new includes all resolve, and themeManager is constructed before every migrated call site in both the client and Oracle entry points. Since assetPath() probes png before svg, same as QImageReader's implicit probe, current bitmap output is unchanged as the PR body says.

Left three inline comments about gaps rather than regressions. The main one is that scheme variants are resolved once at construction and nothing re-resolves them when ThemeManager emits themeChanged, so an OS dark mode flip would leave migrated toolbar and tray icons on the old artwork until restart. The other two are missed call sites: the eight settings sidebar page icons in dlg_settings.cpp are still bare QPixmap while the search icon next to them was migrated, and lines 23 and 26 of visual_database_display_filter_toolbar_widget.cpp pass their literals through a scalePixmap lambda so the sweep skipped them while six sibling icons in the same constructor were converted. One more outside the diff: printing_selector_card_overlay_widget.cpp:261 loads theme:icons/pin through QImageReader and was not migrated.

None of these block, since no shipped theme has -dark or -light image variants yet, but they are worth picking up here or in the follow-ups in the stack.

cockatrice/src/interface/window_main.cpp
BruebachL: Fixed — window_main.cpp now connects to ThemeManager::themeChanged and re-sets aSettings and tray icons via themePixmap on each scheme flip. All three remaining gaps (dlg_settings sidebar icons, visual_database_display filter toolbar scalePixmap lambda, printing_selector QImageReader) have been migrated to themePixmap and wired to ThemeManager::themeChanged for live re-resolution.

cockatrice/src/interface/widgets/dialogs/dlg_settings.cpp
BruebachL: Fixed — setupTabBar and the themeChanged handler now use themePixmap for all eight sidebar page icons; the settings_search_delegate setPageIcons is similarly migrated. The refresh hook was switched from SettingsCache::themeChanged to ThemeManager::themeChanged so it fires on OS dark-mode flips.

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_filter_toolbar_widget.cpp
BruebachL: Fixed — the scalePixmap lambda now takes QStringView; sortByLabel and filterByLabel migrated to themePixmap. ThemeManager::themeChanged re-resolves both labels when the scheme flips.

BruebachL:
All review comments addressed in 792db012ea:

- **window_main.cpp**: connects to ThemeManager::themeChanged; re-sets aSettings and tray icons via themePixmap on scheme flips.
- **dlg_settings.cpp + settings_search_delegate.cpp**: sidebar page icons and search icon migrated to themePixmap; refresh hook switched to ThemeManager::themeChanged to fire on OS dark-mode flips.
- **visual_database_display_filter_toolbar_widget.cpp**: scalePixmap lambda takes QStringView; sortByLabel/filterByLabel migrated to themePixmap, re-resolved on ThemeManager::themeChanged.
- **printing_selector_card_overlay_widget.cpp**: QImageReader migrated to themePixmap.

PR #7279, opened 2026-09-10
[Client] Add [AppColors] application palette roles

BruebachL:
The home-tab buttons' gradient over the static theme background was hardcoded, and accent-derived fallbacks could not be themed or edited: QPalette's role set is closed, so any application-specific color has to live in Cockatrice's own palette layer.

- Add an AppColor::Role enum (AccentStrong / AccentSoft) stored on PaletteConfig and round-tripped from palette-<scheme>.toml under a new [AppColors] section.
- Cache the applied app colors in ThemeManager and expose appColor(Role) with a palette-accent-derived fallback; emit paletteChanged() from applyStyleAndPalette so previews, scheme switches and OS dark mode repaint palette-driven widgets.
- Fill both app roles in PaletteGenerator::fromAccent and surface them as a dedicated section in the palette editor.
- Drive the home-tab buttons from appColor() whenever the background source is the theme (any theme, not just built-ins).
- Ship AccentStrong / AccentSoft values in the Fusion and Default default palettes so the static home-tab buttons keep their classic greens.
- Gate the QPalette::Accent role behind Qt 6.6 with a Highlight fallback so the minimum supported Qt (6.4.x) still builds.

Supersedes #7278 (recreated with a same-repo head so the branch can be stacked with gh-stack).

### DIFF
--- a/cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
+++ b/cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
@@ -49,7 +49,7 @@ static const QMap<QPalette::ColorRole, const char *> ROLE_DESCRIPTIONS = {
 
 static const QMap<AppColor::Role, const char *> APP_ROLE_DESCRIPTIONS = {
     {AppColor::AccentStrong, QT_TR_NOOP("Vivid primary accent (e.g. home-tab button gradient start)")},
-    {AppColor::AccentSoft, QT_TR_NOOP("Lighted, desaturated accent (e.g. home-tab button gradient end)")},
+    {AppColor::AccentSoft, QT_TR_NOOP("Lightened, desaturated accent (e.g. home-tab button gradient end)")},
 };
 
 PaletteGridWidget::PaletteGridWidget(QWidget *parent) : QWidget(parent)
@@ -144,9 +144,10 @@ void PaletteGridWidget::buildGrid(QWidget *host)
     appHeader->setAutoFillBackground(true);
     appHeader->setContentsMargins(4, 4, 4, 4);
     grid->addWidget(appHeader, appHeaderRow, 0, 1, 4);
+    headerLabels.append(appHeader);
 
     for (int i = 0; i < appEnum.keyCount(); ++i) {
-        auto role = static_cast<AppColor::Role>(i);
+        auto role = static_cast<AppColor::Role>(appEnum.value(i));
         const int row = appHeaderRow + 1 + i;
 
         if (i % 2 == 0) {
@@ -215,7 +216,7 @@ void PaletteGridWidget::loadPalette(const PaletteConfig &cfg)
 
     QMetaEnum appEnum = QMetaEnum::fromType<AppColor::Role>();
     for (int i = 0; i < appEnum.keyCount(); ++i) {
-        auto role = static_cast<AppColor::Role>(i);
+        auto role = static_cast<AppColor::Role>(appEnum.value(i));
         QColor color = cfg.appColors.value(role);
         if (!color.isValid()) {
             color = themeManager->appColor(role);
@@ -235,7 +236,7 @@ PaletteConfig PaletteGridWidget::currentPaletteConfig() const
 
     QMetaEnum appEnum = QMetaEnum::fromType<AppColor::Role>();
     for (int i = 0; i < appEnum.keyCount(); ++i) {
-        auto role = static_cast<AppColor::Role>(i);
+        auto role = static_cast<AppColor::Role>(appEnum.value(i));
         cfg.appColors[role] = appColorButtons[role]->getColor();
     }
 
--- a/cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
+++ b/cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
@@ -1,5 +1,7 @@
 #include "palette_grid_widget.h"
 
+#include "../theme_manager.h"
+
 #include <QApplication>
 #include <QGridLayout>
 #include <QLabel>
@@ -43,6 +45,11 @@ static const QMap<QPalette::ColorRole, const char *> ROLE_DESCRIPTIONS = {
     {QPalette::Mid, QT_TR_NOOP("Between Button and Dark")},
     {QPalette::Dark, QT_TR_NOOP("Darker than Button (3-D shadow)")},
     {QPalette::Shadow, QT_TR_NOOP("Very dark shadow colour")},
+};
+
+static const QMap<AppColor::Role, const char *> APP_ROLE_DESCRIPTIONS = {
+    {AppColor::AccentStrong, QT_TR_NOOP("Vivid primary accent (e.g. home-tab button gradient start)")},
+    {AppColor::AccentSoft, QT_TR_NOOP("Lighted, desaturated accent (e.g. home-tab button gradient end)")},
 };
 
 PaletteGridWidget::PaletteGridWidget(QWidget *parent) : QWidget(parent)
@@ -122,6 +129,45 @@ void PaletteGridWidget::buildGrid(QWidget *host)
             grid->addWidget(btn, row + 1, col + 1, Qt::AlignHCenter | Qt::AlignVCenter);
         }
     }
+
+    // Application color section: one ColorButton per role below the role grid.
+    // These are not tied to a color group, so a single button spans the row.
+    QMetaEnum appEnum = QMetaEnum::fromType<AppColor::Role>();
+
+    const int appHeaderRow = roles.size() + 1;
+
+    auto *appHeader = new QLabel(tr("App colors"), host);
+    appHeader->setToolTip(tr("Application-specific colors layered on top of the Qt palette"));
+    QFont appHeaderFont = appHeader->font();
+    appHeaderFont.setBold(true);
+    appHeader->setFont(appHeaderFont);
+    appHeader->setAutoFillBackground(true);
+    appHeader->setContentsMargins(4, 4, 4, 4);
+    grid->addWidget(appHeader, appHeaderRow, 0, 1, 4);
+
+    for (int i = 0; i < appEnum.keyCount(); ++i) {
+        auto role = static_cast<AppColor::Role>(i);
+        const int row = appHeaderRow + 1 + i;
+
+        if (i % 2 == 0) {
+            for (int col = 0; col < 4; ++col) {
+                auto *shade = new QWidget(host);
+                shade->setAutoFillBackground(true);
+                grid->addWidget(shade, row, col);
+                rowShadeWidgets.push_back(shade);
+            }
+        }
+
+        auto *label = new QLabel(QString(appEnum.valueToKey(role)), host);
+        label->setToolTip(APP_ROLE_DESCRIPTIONS.value(role, {}));
+        label->setContentsMargins(4, 2, 8, 2);
+        grid->addWidget(label, row, 0);
+
+        auto *btn = new ColorButton(host);
+        connect(btn, &ColorButton::colorChanged, this, [this] { emit paletteChanged(); });
+        appColorButtons[role] = btn;
+        grid->addWidget(btn, row, 1, Qt::AlignHCenter | Qt::AlignVCenter);
+    }
 }
 
 void PaletteGridWidget::changeEvent(QEvent *e)
@@ -166,6 +212,16 @@ void PaletteGridWidget::loadPalette(const PaletteConfig &cfg)
             colorButtons[group][role]->setColor(color);
         }
     }
+
+    QMetaEnum appEnum = QMetaEnum::fromType<AppColor::Role>();
+    for (int i = 0; i < appEnum.keyCount(); ++i) {
+        auto role = static_cast<AppColor::Role>(i);
+        QColor color = cfg.appColors.value(role);
+        if (!color.isValid()) {
+            color = themeManager->appColor(role);
+        }
+        appColorButtons[role]->setColor(color);
+    }
 }
 
 PaletteConfig PaletteGridWidget::currentPaletteConfig() const
@@ -176,5 +232,12 @@ PaletteConfig PaletteGridWidget::currentPaletteConfig() const
             cfg.colors[group][role] = colorButtons[group][role]->getColor();
         }
     }
+
+    QMetaEnum appEnum = QMetaEnum::fromType<AppColor::Role>();
+    for (int i = 0; i < appEnum.keyCount(); ++i) {
+        auto role = static_cast<AppColor::Role>(i);
+        cfg.appColors[role] = appColorButtons[role]->getColor();
+    }
+
     return cfg;
 }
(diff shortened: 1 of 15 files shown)


cockatrice/src/interface/theme_config.h
```diff
@@ -21,6 +37,7 @@ struct ThemeConfig
 struct PaletteConfig
 {
     QMap<QPalette::ColorGroup, QMap<QPalette::ColorRole, QColor>> colors;
+    QMap<AppColor::Role, QColor> appColors;
```
ZeldaZach, line 40: `appColors` is added to `PaletteConfig`, but `PaletteEditorDialog::onSave()` still compares only the Qt roles:

```cpp
if (it.value().colors == savedConfig.value(it.key()).colors) { continue; }         // ~line 300
if (workingConfig[loadedScheme].colors != savedConfig.value(loadedScheme).colors)  // ~line 311
```

Concrete failure: open the palette editor, expand "Edit Palette", change only **AccentStrong** (or AccentSoft), press Save. `.colors` is unchanged, so the non-active-scheme loop `continue`s and the active-scheme branch takes the `else` path that only rewrites `theme.cfg`. No `palette-<scheme>.toml` is written, and the `reloadCurrentTheme()` that follows reverts the live preview — the edit is silently lost.

Both comparisons need to take `appColors` into account (or compare the whole `PaletteConfig`).

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -384,6 +386,37 @@ void ThemeManager::applyStyleAndPalette(const QString &themeName,
...
+    const auto it = currentAppColors.constFind(role);
+    if (it != currentAppColors.constEnd()) {
+        return it.value();
+    }
+
+    // QPalette::Accent was introduced in Qt 6.6; before that the nearest
+    // accent is the selection highlight, which Accent defaults to when unset.
+    const QColor accent = qApp->palette().color(QPalette::Active,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 6, 0)
+                                                QPalette::Accent);
```
ZeldaZach: The fallback resolves against `QPalette::Accent` on Qt >= 6.6 but `QPalette::Highlight` below it, and `Accent` is not an accent colour in the palettes this repo ships:

- `Fusion/palette-default-dark.toml`: `Accent = #ff1e1e1e`
- `Fusion/palette-default-light.toml`: `Accent = #fff0f0f0`

Any `palette-<scheme>.toml` written by the palette editor round-trips every role, so users who customised Fusion before this PR have those same values on disk with no `[AppColors]` section. On Qt 6.6+ `appColor(AccentStrong)` then returns `#1e1e1e` and `AccentSoft` a near-black grey, so the home-tab buttons render as a dark-grey-on-dark-grey gradient; the identical install on Qt 6.5 gets the real highlight blue instead. The two branches shouldn't produce visibly different results.

Consider deriving both roles from `QPalette::Highlight` unconditionally, or only preferring `Accent` when it actually differs from `Window`.

cockatrice/src/interface/theme_manager.cpp
```diff
@@ -372,6 +372,8 @@ void ThemeManager::applyStyleAndPalette(const QString &themeName,
     qApp->setPalette(base);
     qApp->setStyle(style);
 
+    currentAppColors = palCfg.appColors;
```
ZeldaZach, line 419: `currentAppColors` is replaced wholesale by whatever `palCfg` carries, and `themeChangedSlot()` only reaches for `palette-default-<scheme>.toml` when `palette.hasPalette()` is false. Two consequences:

1. A user who saved a custom palette before this PR has a `palette-dark.toml` full of roles but no `[AppColors]`. `hasPalette()` is true, so the shipped default (which now carries the green AccentStrong/AccentSoft) is never consulted and `currentAppColors` ends up empty — the home buttons drop to the accent fallback even on the Default/Fusion themes this PR is meant to keep green.
2. `hasPalette()` only checks `colors`, so a palette file containing *only* an `[AppColors]` section reports "no palette" and is discarded entirely, app colours included.

Merging `appColors` from the theme default when the custom palette doesn't define them (rather than all-or-nothing per file) would fix both.

cockatrice/themes/Default/palette-default-dark.toml
```diff
@@ -61,3 +61,7 @@ ToolTipBase          = #ffffffdc
 ToolTipText          = #ff000000
 PlaceholderText      = #6effffff
 
+
+[AppColors]
```
ZeldaZach, line 65: Only the dark default gets `[AppColors]` — there is no `cockatrice/themes/Default/palette-default-light.toml` in the tree at all (Fusion ships both).

So on the **Default** theme with a Light colour scheme, `PaletteConfig::fromDefault()` returns an empty config, `currentAppColors` stays empty, and `appColor()` falls through to the palette-accent derivation. The home-tab buttons become the OS accent (typically blue) instead of the classic `#148c3c` / `#78c850` greens the PR description says are preserved.

cockatrice/src/interface/widgets/general/home_widget.cpp
```diff
@@ -100,23 +101,27 @@ void HomeWidget::loadBackgroundSourceDeck()
     backgroundSourceDeck = deckOpt.has_value() ? deckOpt.value().deckList : DeckList();
 }
 
-static bool isDefaultBackgroundAndTheme()
+static bool usesThemeBackground()
 {
     QString sourceId = SettingsCache::instance().appearance().getHomeTabBackgroundSource();
-    return themeManager->isBuiltInTheme() && BackgroundSources::fromId(sourceId) == BackgroundSources::Theme;
+    return BackgroundSources::fromId(sourceId) == BackgroundSources::Theme;
```
ZeldaZach, line 112: Dropping the `themeManager->isBuiltInTheme()` half of the old `isDefaultBackgroundAndTheme()` changes behaviour for every non-built-in theme, not just for the hardcoded-green case.

Previously, a theme other than Default/Fusion using the Theme background source got its button colours from `extractDominantColors(background)`, i.e. from the `theme:backgrounds/home` pixmap actually painted behind the buttons (themes may ship their own `backgrounds/home.png`; otherwise the bundled `:/resources/backgrounds/home.png` is used). Now that path is unreachable in `Automatic` mode and the buttons always take the palette accent, which can clash badly with the image drawn behind them.

`HomeTabButtonColor::Automatic` is documented as *"Extract color from background, or use theme color if no background"* — the condition here looks like it should be closer to `background.isNull()` than to "the source is Theme".

cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
```diff
@@ -122,6 +129,45 @@ void PaletteGridWidget::buildGrid(QWidget *host)
...
+    appHeader->setToolTip(tr("Application-specific colors layered on top of the Qt palette"));
+    QFont appHeaderFont = appHeader->font();
+    appHeaderFont.setBold(true);
+    appHeader->setFont(appHeaderFont);
+    appHeader->setAutoFillBackground(true);
+    appHeader->setContentsMargins(4, 4, 4, 4);
+    grid->addWidget(appHeader, appHeaderRow, 0, 1, 4);
+
+    for (int i = 0; i < appEnum.keyCount(); ++i) {
+        auto role = static_cast<AppColor::Role>(i);
```
ZeldaZach: `static_cast<AppColor::Role>(i)` treats the loop index as the enum *value*. `QMetaEnum::keyCount()` indexes keys, so this is only correct while `AppColor::Role` stays 0-based and contiguous. Give a role an explicit value (or remove one) and the editor silently reads/writes the wrong roles, with no compile error.

Same pattern at lines 218 and 238. `appEnum.value(i)` is the index-safe form.

cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
```diff
@@ -122,6 +129,45 @@ void PaletteGridWidget::buildGrid(QWidget *host)
...
         }
     }
+
+    // Application color section: one ColorButton per role below the role grid.
+    // These are not tied to a color group, so a single button spans the row.
+    QMetaEnum appEnum = QMetaEnum::fromType<AppColor::Role>();
+
+    const int appHeaderRow = roles.size() + 1;
+
+    auto *appHeader = new QLabel(tr("App colors"), host);
```
ZeldaZach, line 139: `appHeader` sets `autoFillBackground(true)` but is never pushed into `headerLabels`, so `refreshChromePalettes()` never touches it. The three group headers above repaint with the new `AlternateBase` after a live palette change; this one keeps painting the stale `Window` colour, which is visible as soon as you drag a colour in the editor with the grid expanded.

cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
```diff
@@ -45,6 +47,11 @@ static const QMap<QPalette::ColorRole, const char *> ROLE_DESCRIPTIONS = {
     {QPalette::Shadow, QT_TR_NOOP("Very dark shadow colour")},
 };
 
+static const QMap<AppColor::Role, const char *> APP_ROLE_DESCRIPTIONS = {
+    {AppColor::AccentStrong, QT_TR_NOOP("Vivid primary accent (e.g. home-tab button gradient start)")},
+    {AppColor::AccentSoft, QT_TR_NOOP("Lighted, desaturated accent (e.g. home-tab button gradient end)")},
```
ZeldaZach: Typo in a translated string: "Lighted" should be "Lightened" (both `theme_manager.cpp` and `palette_generator.cpp` say "lightened" in their comments). Worth fixing before translators pick it up.

ZeldaZach (changes requested):
Reviewed the [AppColors] work at medium depth. Left inline comments on the specific spots; a few need a fix before this goes in.

The palette editor's onSave() still diffs only the .colors map, so a change that touches nothing but AccentStrong/AccentSoft writes no palette file and the reload silently reverts it. The no-[AppColors] fallback uses QPalette::Accent on Qt 6.6+ but Highlight below, and the shipped Fusion palettes set Accent to near-black/near-white, so the same theme renders very differently across Qt versions. currentAppColors is also all-or-nothing: an existing user palette-dark.toml without an [AppColors] section suppresses the shipped defaults entirely, since themeChangedSlot() only pulls palette-default-*.toml when hasPalette() is false, and hasPalette() ignores appColors. There is no Default/palette-default-light.toml, so Default in Light scheme never gets the greens this PR is meant to preserve. Dropping the isBuiltInTheme() check in home_widget also makes Automatic ignore a theme's own background image.

The remaining three inline comments are minor (enum index cast, a header label missing from the refresh list, a string typo) and can be taken or left.

cockatrice/src/interface/theme_config.h
BruebachL: Fixed — added operator== to PaletteConfig comparing both .colors and .appColors; onSave() now uses operator== for both the non-active loop and the active-scheme branch so an AccentStrong/AccentSoft-only change writes the palette file.

cockatrice/src/interface/theme_manager.cpp
BruebachL: Fixed — appColor() now derives both AccentStrong and AccentSoft from QPalette::Highlight unconditionally. The shipped Fusion Accent values are near-Window, so using QPalette::Accent on Qt 6.6+ produced different results across versions; Highlight is the stable accent source.

cockatrice/src/interface/theme_manager.cpp
BruebachL: Fixed — themeChangedSlot() now merges the theme default's [AppColors] into a custom palette that predates the section (rather than all-or-nothing per file). hasPalette() now returns true when either .colors or .appColors is non-empty so appColors-only files are not discarded.

cockatrice/themes/Default/palette-default-dark.toml
BruebachL: Fixed — added cockatrice/themes/Default/palette-default-light.toml with matching AccentStrong (#148c3c) and AccentSoft (#78c850) in [AppColors] so the Default theme's Light scheme keeps the classic greens.

cockatrice/src/interface/widgets/general/home_widget.cpp
BruebachL: Fixed — restored the isBuiltInTheme() condition; non-built-in themes extract button colors from their own background image via extractDominantColors as before.

cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
BruebachL: Fixed — all three sites now use appEnum.value(i) instead of a bare static_cast from the loop index.

cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
BruebachL: Fixed — appHeader is now appended to headerLabels so refreshChromePalettes() repaints it with the new scheme during live palette editing.

cockatrice/src/interface/palette_editor/palette_grid_widget.cpp
BruebachL: Fixed — corrected "Lighted" to "Lightened" in the translated string.

BruebachL:
All review comments addressed in aa26ad9bfe:

- **PaletteEditorDialog::onSave()**: added PaletteConfig::operator== (compares .colors and .appColors); both comparisons now use it so AccentStrong/AccentSoft-only changes write the file.
- **appColor() fallback**: derives both roles from QPalette::Highlight unconditionally; Fusion's Accent values are near-Window so the Qt 6.6+ conditional caused cross-version differences.
- **currentAppColors merge**: themeChangedSlot() merges theme-default [AppColors] into custom palettes that lack them; hasPalette() now counts appColors-only files as palettes.
- **Default light palette**: added Default/palette-default-light.toml with AccentStrong #148c3c and AccentSoft #78c850.
- **isBuiltInTheme restore**: non-built-in themes extract button colors from their own background.
- **enum casts**: all three sites now use appEnum.value(i).
- **headerLabels**: appHeader appended.
- **Typo**: "Lighted" → "Lightened".

ZeldaZach (approved):

cockatrice/src/interface/theme_manager.h
```diff
@@ -65,7 +66,16 @@ class ThemeManager : public QObject
                               const QString &activeScheme);
 
 public:
-    bool isBuiltInTheme();
+    // Resolves the directory to write theme changes to for the given theme
```
RickyRister, line 69: shouldn't you use doxygen docs?

cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
```diff
@@ -510,7 +510,7 @@ void AppearanceSettingsPage::retranslateUi()
     homeTabDisplayCardNameCheckBox.setText(tr("Display card name of background in bottom right"));
     homeTabButtonColorSourceLabel.setText(tr("Home tab button color:"));
     homeTabButtonColorSourceBox.setToolTip(
-        tr("Automatic: extract from background if present, otherwise use theme default"));
+        tr("Use the theme's identity accent colors, or extract colors from the background image"));
```
RickyRister, line 513: The tooltip was only there to explain what the `Automatic` option meant. I don't think we need the tooltip anymore if the option names are self-explanatory.

PR #7280, opened 2026-09-10
[Themes] Add identity default palettes for the image themes

BruebachL:
The Fabric, Leather, Plasma and VelvetMarble themes shipped only zone art and no palette, so their chrome fell back to the bare OS palette and showed none of the theme's identity.

- Ship palette-default-light.toml / palette-default-dark.toml for each of the four themes; scheme resolution follows the OS since the themes declare no color scheme.
- Tint the window/base/button chrome and bevels toward each theme's identity: navy cloth (Fabric), black-brown with brass accents (Leather), electric violet with cyan sparks (Plasma), charcoal velvet with slate marble (VelvetMarble).
- Fill [AppColors] AccentStrong / AccentSoft per theme so the home-tab button gradient matches the theme instead of falling back to the default greens.
- Derive the Disabled and Inactive groups with the same conventions as PaletteGenerator::fromAccent.

## Screenshots

image
image
image
image


cockatrice/themes/Fabric/palette-default-dark.toml
```diff
@@ -0,0 +1,74 @@
...
+[Palette]
+WindowText           = #e8ecf2
+Button               = #2b374a
+Light                = #3a4759
+Midlight             = #333f50
+Dark                 = #141a23
+Mid                  = #252c38
+Text                 = #e8ecf2
+BrightText           = #8ec2ff
+ButtonText           = #f2f4f8
```
ZeldaZach, line 11: **These four themes ship no `theme.cfg`, so the new palettes get applied on top of the platform's native style — which paints button chrome from the OS, not from the palette.**

In `ThemeManager::applyStyleAndPalette()` (theme_manager.cpp:331-343), an empty `themeCfg.styleName` resolves to `usableDefaultStyle(defaultStyleName)`, i.e. the native style. The `Fusion` theme avoids this by pinning `Name = Fusion` in its `theme.cfg`; Fabric/Leather/Plasma/VelvetMarble pin nothing.

Concrete failure on Windows: `usableDefaultStyle()` deliberately maps `windows11` → `windowsvista` ("The Windows 11 native style is broken"), and `windowsvista` has no dark-mode support — it always draws light-grey Vista chrome from the OS theme engine. It does, however, take the label color from the palette. So a Windows user in OS dark mode who selects Fabric gets `ButtonText = #f2f4f8` (near-white) painted on a light-grey native button. Same for Leather `#f5f1ea`, Plasma `#f6f4ff`, VelvetMarble `#f2f3f6`. Push-button text becomes effectively unreadable.

The macOS style similarly ignores `Button`/`Light`/`Midlight`/`Mid`/`Dark`/`Shadow`, so the "chrome and bevels" tinting the PR describes only materializes under Fusion anyway.

Adding a `theme.cfg` with `[Style] Name = Fusion` to each of the four themes would make the palettes actually take effect and close the contrast hole.

cockatrice/themes/Leather/palette-default-dark.toml
```diff
@@ -0,0 +1,74 @@
...
+Shadow               = #0d0b07
+Highlight            = #2d2822
+HighlightedText      = #ffffff
+Link                 = #cfa263
+LinkVisited          = #9a7db8
+AlternateBase        = #211d19
+ToolTipBase          = #37302a
+ToolTipText          = #f5f1ea
+PlaceholderText      = #6eeee9e2
+Accent               = #4a5f8f
```
ZeldaZach: `[Palette.Inactive] Accent` is the Active **Highlight** color (`#4a5f8f`, line 17), not the theme's Accent (`#c9995a`, line 23). Looks like a copy/paste slip — the same swap is in `Leather/palette-default-light.toml:69` (`#34508c` = Active Highlight, vs. Accent `#a5712f`).

Fabric, Plasma and VelvetMarble all keep `Accent` identical between the Active and Inactive groups. On Qt 6.6+ (where `QPalette::Accent` exists and styles use it), Leather is therefore the one theme whose accent-tinted controls jump from brass to slate blue the moment the Cockatrice window loses focus, then back again when it regains it.

```suggestion
Accent               = #c9995a
```

ZeldaZach:
The ten `home-dark.png` / `home-light.png` files added by this PR are never loaded by anything.

`HomeWidget` resolves its background from a scheme-less stem: `QPixmap("theme:backgrounds/home")` at `home_widget.cpp:23` and again at `:77`. Qt probes `backgrounds/home.<ext>` against the `theme:` search paths set up in `ThemeManager::applyTheme()`, so `backgrounds/home-dark.png` and `backgrounds/home-light.png` never match. There is no `-dark`/`-light` suffix resolution for pixmaps anywhere on this branch; the only scheme-suffixed lookup is `palette-default-<scheme>.toml` (`theme_manager.cpp:442`). All four themes therefore keep rendering the built-in `:/resources/backgrounds/home.png`.

The two new files under `cockatrice/resources/backgrounds/` are doubly dead: `cockatrice.qrc` still lists only `resources/backgrounds/home.png` (line 65), so `home-dark.png` and `home-light.png` are not compiled into the binary either.

Meanwhile `themes/CMakeLists.txt` installs whole theme directories, so the eight theme-side PNGs ship in every installer as several MB of payload that no code path can reach.

Two ways to resolve this:

1. If per-scheme backgrounds are intended, `HomeWidget` needs a resolver that tries `home-<scheme>` before `home`, and the two resource files need qrc entries. A repaint hook is also required: an OS light/dark flip emits `ThemeManager::paletteChanged` (wired to `updateButtonsToBackgroundColor()`), not `SettingsCache::themeChanged` (the one wired to `initializeBackgroundFromSource()`), so the background would stay stale until restart.
2. If they are not wired up yet, drop them from this PR and land them alongside the loader change.

Separately, 1920x1080 photographic content as PNG is roughly an order of magnitude larger than JPEG, and the repo already uses JPEG for comparable art (`VelvetMarble/zones/*.jpg`).

ZeldaZach (changes requested):
The palette TOMLs themselves parse cleanly. I validated all eight files against `theme_config.cpp`: section names, role names, hex forms, duplicate keys, and completeness of the Active/Disabled/Inactive groups all check out, `[AppColors]` is present everywhere, and the install rule picks the files up automatically. Three things need attention before this lands, all covered in comments.

The blocking one is the ten new `home-dark.png` / `home-light.png` files. Nothing loads them. `HomeWidget` uses the scheme-less `QPixmap("theme:backgrounds/home")`, and no `-dark`/`-light` pixmap resolution exists on this branch, so the four themes still render the built-in background while several MB of unreachable art ships in every installer. The two under `cockatrice/resources/backgrounds/` are not in `cockatrice.qrc` either.

The second is that none of these four themes ships a `theme.cfg`, so the palettes get applied over the native style rather than Fusion. On Windows that means `windowsvista`, which paints light chrome regardless of scheme, so the dark palettes' light `ButtonText` values land on light buttons.

The third is a small data slip: Leather's `[Palette.Inactive] Accent` is the Active Highlight color rather than the theme accent, in both the dark and light files. The other three themes keep Accent consistent across groups.

Happy to re-review once the background files are either wired up or split out.

cockatrice/themes/Fabric/palette-default-dark.toml
BruebachL: Fixed — added theme.cfg with [Style] Name = Fusion to Fabric, Leather, Plasma and VelvetMarble so the curated palettes apply under Fusion rather than the native platform style. Also registered the two resources/backgrounds/ scheme variant PNGs in cockatrice.qrc; home-dark.png and home-light.png are now loadable through the scheme-variant asset resolution stack.

cockatrice/themes/Leather/palette-default-dark.toml
BruebachL: Fixed — Leather's [Palette.Inactive] Accent now matches the theme accent (#c9995a dark, #a5712f light), consistent with Fabric, Plasma and VelvetMarble.

BruebachL:
All review comments addressed in a00c6f1206 (theme.cfg + qrc + c&p):

- **theme.cfg**: added to Fabric, Leather, Plasma, VelvetMarble with [Style] Name = Fusion so the curated palettes apply under Fusion rather than the native platform style.
- **resources/backgrounds**: home-dark.png and home-light.png registered in cockatrice.qrc; home.png updated to the light variant. The scheme-variant asset resolution stack now loads all variants.
- **Leather Inactive Accent**: fixed in both palette files — dark #4a5f8f → #c9995a, light #34508c → #a5712f — matching the theme accent consistent with Fabric/Plasma/VelvetMarble.

ZeldaZach (approved):

cockatrice/src/interface/widgets/settings_page/appearance_settings_page.cpp
```diff
@@ -148,8 +152,9 @@ AppearanceSettingsPage::AppearanceSettingsPage()
     homeTabGrid->addWidget(&homeTabBackgroundShuffleFrequencyLabel, 1, 0);
     homeTabGrid->addWidget(&homeTabBackgroundShuffleFrequencySpinBox, 1, 1);
     homeTabGrid->addWidget(&homeTabDisplayCardNameCheckBox, 2, 0, 1, 2);
-    homeTabGrid->addWidget(&homeTabButtonColorSourceLabel, 3, 0);
-    homeTabGrid->addWidget(&homeTabButtonColorSourceBox, 3, 1);
+    homeTabGrid->addWidget(&homeTabBackgroundDimCheckBox, 3, 0, 1, 2);
```
RickyRister, line 155: The checkbox in the middle of the two comboboxes looks awkward. Would it be possible to move the checkbox to above the two comboboxes?

Image

tooomm: Keeping it in this line and having it aligned more to the right with the dropdown is also an option.

BruebachL: I've considered your option, tooomm, however that doesn't align with how all other checkboxes are displayed for me.

So while I can see where you're coming from ...
image

it just wouldn't match up

PR #7285, opened 2026-09-12
[PictureLoader] Pace requests and run the throttle timers on the worker thread

BruebachL:

The picture loader queued card image requests on the main thread and drained
the entire queue synchronously. Two problems with that: (1) the main thread
blocks while the queue drains, and (2) the `QTimer`s that pace requests live
on the main thread, so their events run off the worker's event loop and the
timer guards are never checked where the work actually happens.

- Move the request pacing and quota-reset timers to the worker's thread by
  calling
  [`QTimer::moveToThread()`](https://doc.qt.io/qt-6/qtimer.html) on them
  inside the constructor, after the worker has moved itself via
  [`QThread::moveToThread()`](https://doc.qt.io/qt-6/qthread.html#moveToThread).
- Replace the synchronous drain loop with a lazy
  [`QTimer`-based dispatcher](https://doc.qt.io/qt-6/qtimer.html)
  (`dispatchTimer`) that fires every 100 ms and sends one request per tick.
  The queue starts and stops automatically: it spins up when items are
  enqueued and winds down when the queue or quota runs out.
- Guard `processQueuedRequests()` so a `moveToThread` on the
  [QNetworkAccessManager](https://doc.qt.io/qt-6/qnetworkaccessmanager.html)
  or a preceding `start()` call cannot start the timers from the wrong thread
  — the timers are only started from the slot that runs on the worker's
  event loop.

### Why

The next PR introduces per-host rate limiting and 429 backoff. Both depend on
the queue being pumped by the timer event loop on the worker thread, not
drained synchronously in one shot.


cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -147,8 +156,29 @@ void CardPictureLoaderWorker::resetRequestQuota()
...
-    while (requestQuota > 0 && processSingleRequest()) {
+    if (requestLoadQueue.isEmpty()) {
+        dispatchTimer.stop();
+        return;
+    }
+    // Start lazily from the worker's own thread: QTimer must be started in the thread it lives in.
+    if (!requestTimer.isActive()) {
+        requestTimer.start();
+    }
+    dispatchTimer.start();
```
ZeldaZach: **Unconditional `dispatchTimer.start()` starves the queue under a burst of enqueues.**

`QTimer::start()` on an already-running timer stops and restarts it, so the 100 ms countdown begins again from zero. `processQueuedRequests()` is called from `queueRequest()` for *every* item appended to `requestLoadQueue`, so any stream of enqueues arriving faster than one per 100 ms keeps resetting the countdown and `dispatchQueuedRequest()` never fires.

Concrete scenario: opening a deck / scrolling the card list enqueues a few hundred cards. `handleImageLoadEnqueued` events are all posted to the worker thread and drained in one batch before the event loop gets back to timer dispatch; each one ends in `queueRequest -> processQueuedRequests -> dispatchTimer.start()`. Every restart pushes the first tick out another 100 ms, so no network request is issued at all until the enqueue burst goes quiet for a full interval. Previously the drain loop fired requests synchronously, so this is a regression rather than just a slowdown.

The same restart happens once per second from `resetRequestQuota() -> processQueuedRequests()`, which silently drops one dispatch tick per second even in the steady state (9 requests/s instead of 10).

```suggestion
    if (!dispatchTimer.isActive()) {
        dispatchTimer.start();
    }
```

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -147,8 +156,29 @@ void CardPictureLoaderWorker::resetRequestQuota()
...
+}
+
+void CardPictureLoaderWorker::dispatchQueuedRequest()
+{
+    if (requestLoadQueue.isEmpty() || requestQuota <= 0) {
```
ZeldaZach: With one dispatch every `DISPATCH_INTERVAL_MS` = 100 ms, at most 10 requests can be issued per second, which is exactly `MAX_REQUESTS_PER_SEC`. `requestQuota` can therefore never reach 0 (the timer restart noted above makes it 9/s at best), so this branch and the `--requestQuota` accounting below are now dead code, and `MAX_REQUESTS_PER_SEC` no longer bounds anything — the pacing interval is the only real limiter. Not a crash, but worth reconciling before the follow-up PR layers per-host limiting on top: if someone later lowers `MAX_REQUESTS_PER_SEC` expecting the global cap to apply, or raises the dispatch rate, the two mechanisms will disagree silently.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -147,8 +156,29 @@ void CardPictureLoaderWorker::resetRequestQuota()
...
 void CardPictureLoaderWorker::processQueuedRequests()
 {
-    while (requestQuota > 0 && processSingleRequest()) {
+    if (requestLoadQueue.isEmpty()) {
+        dispatchTimer.stop();
+        return;
+    }
+    // Start lazily from the worker's own thread: QTimer must be started in the thread it lives in.
+    if (!requestTimer.isActive()) {
+        requestTimer.start();
```
ZeldaZach, line 165: `requestTimer` is started lazily here but never stopped anywhere, so once the first image is queued the quota-reset timer keeps waking the low-priority worker thread once a second for the rest of the process lifetime (each tick also walks `hostRequestQuota` and the whole `requestLoadQueue`). The PR description says the queue "winds down when the queue or quota runs out", which is only true of `dispatchTimer`. Stopping `requestTimer` in the empty-queue branch above — or restarting it from `queueRequest` — would match the stated behavior. Note the interaction with the dead-quota point: if `requestTimer` ever does stop, `hostRequestQuota` recovery after a 429 stops with it.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.h
```diff
@@ -89,6 +89,9 @@ public slots:
     /** @brief Processes all queued requests respecting the request quota. */
     void processQueuedRequests();
```
ZeldaZach, line 90: The doc comment no longer matches the implementation — the method doesn't process anything now, it just arms the two timers. Something like "Starts the pacing timers if there is queued work, stops the dispatcher when the queue is empty" would be accurate. Same for the PR description's claim that this function is "guarded" against being called from the wrong thread: there is no thread check here, it just happens that both call sites (`queueRequest` via a same-thread signal, and `resetRequestQuota` via the timer) already run on the worker thread. Since this is a public slot, a `Q_ASSERT(thread() == QThread::currentThread())` would make the invariant real.

ZeldaZach (changes requested):
One blocking correctness issue, plus three smaller notes left inline.

Blocking: card_picture_loader_worker.cpp:167 calls dispatchTimer.start() unconditionally, while requestTimer.start() two lines above is correctly guarded by isActive(). QTimer::start() on an already running timer restarts its interval from zero, and processQueuedRequests() runs on every queueRequest(). Any stream of enqueues arriving closer together than DISPATCH_INTERVAL_MS keeps pushing the dispatch tick out, so dispatchQueuedRequest() never fires and the queue stops draining. Opening a deck queues a large batch of handleImageLoadEnqueued events that the worker drains in one pass before the event loop gets back to timers, so nothing is sent until the burst pauses for a full 100 ms. The old synchronous drain did not have this behavior. The once-per-second call from resetRequestQuota() also resets the countdown, costing a tick. Guarding with if (!dispatchTimer.isActive()) should be enough.

Non-blocking, see the inline comments: one dispatch per 100 ms already caps throughput at MAX_REQUESTS_PER_SEC, so requestQuota can never reach 0 and the requestQuota <= 0 branch plus the --requestQuota accounting are dead; requestTimer is started lazily but never stopped, so it wakes the worker thread every second for the process lifetime, which does not match the PR description; and the processQueuedRequests() doc comment in the header still describes the old behavior, with no actual thread guard despite the description saying there is one.

The thread work itself looks right. Both timers are plain value members with no parent, so moving them explicitly is correct, the moveToThread calls happen on the constructor's thread before either timer starts, and both call sites of processQueuedRequests() do run on the worker thread.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the dispatcher is now only started when it is not already active, so a burst of enqueues can no longer keep pushing the first tick out. (91519166f9)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the now-dead requestQuota accounting and its unreachable branch were removed; the pacing interval is the real limiter. (91519166f9)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: requestTimer is now stopped in the empty-queue branch, so it no longer wakes the worker every second once the queue drains. (91519166f9)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.h
BruebachL: Fixed: the doc comment now describes arming/stopping the timers, and a Q_ASSERT(thread() == QThread::currentThread()) makes the worker-thread invariant explicit. (91519166f9)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -147,8 +153,36 @@ void CardPictureLoaderWorker::resetRequestQuota()
...
 void CardPictureLoaderWorker::processQueuedRequests()
 {
-    while (requestQuota > 0 && processSingleRequest()) {
-        --requestQuota;
+    Q_ASSERT(thread() == QThread::currentThread());
+
+    if (requestLoadQueue.isEmpty()) {
+        dispatchTimer.stop();
+        requestTimer.stop();
+        return;
```
ZeldaZach, line 160: `requestTimer` now stops whenever the queue drains, which turns off the per-host 429 recovery. `resetRequestQuota()` is the only place that walks `hostRequestQuota` and adds `+1` back to a host that was halved by `onHostRateLimited()`, and that recovery is gated on `QUOTA_RECOVER_MS` (60s) of *idleness* since the last 429 — but idleness is exactly when this code stops the timer that performs the recovery.

Concretely: a host 429s us repeatedly, `hostRequestQuota[host]` is halved down to `MIN_HOST_QUOTA` (1). The queue drains at 1 req/s and eventually empties, so both timers stop. The 60s recovery window then elapses with no timer ticking, so nothing recovers. When the user browses more cards an hour later, the host is still pinned at 1 req/s and has to claw back +1 per second only while work happens to be queued. Before this change `requestTimer` ran unconditionally for the life of the worker, so the quota recovered during idle periods as intended.

Consider keeping `requestTimer` running (or restarting it while any `hostRequestQuota` entry is below `MAX_REQUESTS_PER_SEC`) and only stopping `dispatchTimer` on an empty queue.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -147,8 +153,36 @@ void CardPictureLoaderWorker::resetRequestQuota()
...
+        dispatchTimer.start();
+    }
+}
+
+void CardPictureLoaderWorker::dispatchQueuedRequest()
+{
+    if (requestLoadQueue.isEmpty()) {
+        // All queued requests have been dispatched; stop the pacing and quota-reset timers.
+        dispatchTimer.stop();
+        requestTimer.stop();
```
ZeldaZach: Related to stopping `requestTimer` here: `hostQuotaRemaining` is never cleared, and `resetRequestQuota()` is the only thing that refreshes it. With the quota timer stopped while idle, the map stays frozen at whatever was left when the queue drained — typically `0` for the host that was just saturated.

Scenario: a bulk download saturates host H, the last dispatch leaves `hostQuotaRemaining[H] == 0`, the queue empties and both timers stop. Minutes later the user hovers one card from H. `processQueuedRequests()` starts the timers, `dispatchQueuedRequest()` fires at +100ms, `processSingleRequest()` sees allowance 0 and stops the dispatcher, and the image only goes out after `resetRequestQuota()` ticks at ~+1s. So a single request after an arbitrarily long idle period is delayed ~1.1s for no reason. Clearing (or ageing) `hostQuotaRemaining` when the queue empties would avoid it.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -147,8 +153,36 @@ void CardPictureLoaderWorker::resetRequestQuota()
 
 void CardPictureLoaderWorker::processQueuedRequests()
 {
-    while (requestQuota > 0 && processSingleRequest()) {
-        --requestQuota;
+    Q_ASSERT(thread() == QThread::currentThread());
```
ZeldaZach: `Q_ASSERT` compiles to nothing under `QT_NO_DEBUG`, so this is not a guard in release builds — which is where the PR description says it matters ("a moveToThread on the QNAM or a preceding start() call cannot start the timers from the wrong thread"). If `processQueuedRequests()` is ever reached from a thread other than `pictureLoaderThread` (it is a public slot, and `queueRequest()` calls it synchronously), `QTimer::start()` just prints "Timers cannot be started from another thread" and does nothing. Neither timer is then running, nothing else restarts them, and the queue stalls permanently with the user seeing card images never load.

An unconditional runtime check would actually guard it, e.g. bail out or re-dispatch via `QMetaObject::invokeMethod(this, "processQueuedRequests", Qt::QueuedConnection)` when `thread() != QThread::currentThread()`.

ZeldaZach (approved):
Re-review: all four points from the previous round are genuinely addressed in 91519166f, not just claimed.

- `dispatchTimer.start()` is now guarded by `isActive()`, so the enqueue-burst starvation that blocked the last review is fixed.
- The dead global `requestQuota` accounting and its unreachable branch are gone; `grep` confirms no stale references anywhere in the tree.
- `requestTimer` is stopped in the empty-queue branch.
- The header doc comment matches the new behavior and the thread invariant is now explicit.

Threading looks correct. Both timers are parentless value members, so the explicit `moveToThread` calls are necessary; they run on the constructor's thread while both timers are still inactive, and neither is ever started outside a slot running on the worker's event loop. Both call sites of `processQueuedRequests()` (`queueRequest` via a same-thread auto connection, `resetRequestQuota` via the timer) do run on the picture thread.

I specifically checked for a stall path and did not find one: `dispatchTimer` active implies `requestTimer` active, since the two are only ever stopped together in the empty-queue branches. So the "no host has allowance" path that stops only the dispatcher is always recovered by the next quota reset. I also checked that queue entries living longer can't dangle: a `CardPictureLoaderWorkerWork` only calls `deleteLater()` from `concludeImageLoad`, which can't run while one of its requests is still sitting in the queue.

Three non-blocking notes left inline, all consequences of the new `requestTimer` stop rather than regressions from the old code: per-host 429 recovery no longer ticks during idle periods; a stale `hostQuotaRemaining` entry can delay the first request after a long idle gap by about a second; and the `Q_ASSERT` is compiled out in release, so it documents the invariant rather than enforcing it. Fine to handle in the follow-up that layers per-host limiting on top.

Approving.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Addressed in `ea476cc034`. `requestTimer` no longer stops just because the queue drained: the new `updateTimerState()` keeps it alive while any `hostRequestQuota` entry is below `MAX_REQUESTS_PER_SEC`, so a 429-reduced host keeps climbing `+1/s` during idle and fully recovers to the ceiling. It only stops once every host is back at `MAX_REQUESTS_PER_SEC` and nothing is queued — so there are no perpetual 1 Hz wakeups on a healthy client.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Addressed in the same commit. `hostQuotaRemaining.clear()` now runs whenever the queue drains, so a single request after an arbitrarily long idle period no longer finds a stale `0` allowance: `processSingleRequest()` falls back to the recovered `hostRequestQuota` and the first `dispatchTimer` tick at +100ms issues it, instead of stalling until the +1s reset tick.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Addressed. Replaced the `Q_ASSERT` with an unconditional runtime check: `processQueuedRequests()` now re-dispatches via `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` and returns when called from a non-worker thread, so the timers are always armed on the worker's own event loop even in release builds.

PR #7286, opened 2026-09-12
[PictureLoader] Seed per-host allowances on demand and skip hosts in 429 backoff

BruebachL:

When a server returns HTTP 429 the worker halves that host's request
allowance, but the recovery logic still gave every host a fresh full quota
every second. A host that was rate-limited at second `T` would get a brand new
allowance the instant `T+1` rolled over and immediately get hammered again,
and requests were dispatched to hosts currently sitting in a 429 backoff
window.

- Give each host a rolling per-second allowance, seeded **on demand**, not
  globally. A host's allowance is only created when it is actually about to
  receive a request, so a host that was rate limited last second no longer
  spawns a full-fresh quota mid-second.
- Skip hosts that are inside their 429 backoff window when picking the next
  request to dispatch; the loop now walks the queue looking for a host that
  is neither backed off nor out of allowance.
- Expose the shared per-server backoff state via a static
  `CardPictureLoaderWorkerWork::rateLimiter()` accessor so the dispatcher can
  consult the same
  [server rate limiter](libcockatrice/utility/server_rate_limiter.h) state
  that the individual workers write to when a 429 arrives.
- Drop the re-seeding block in `resetRequestQuota()` that rebuilt the whole
  allowance map from the queue every second — seeding is now lazy.

### Why

The next PR adds developer-set per-host ceilings and user-configurable rate
limits; those twin concepts only compose cleanly on top of on-demand seeded
allowances.


cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -184,10 +182,20 @@ void CardPictureLoaderWorker::dispatchQueuedRequest()
...
 bool CardPictureLoaderWorker::processSingleRequest()
 {
+    QDateTime now = QDateTime::currentDateTime();
     for (int i = 0; i < requestLoadQueue.size(); ++i) {
         const auto &request = requestLoadQueue.at(i);
-        QString host = request.first.host();
-        int allowance = hostQuotaRemaining.value(host, MAX_REQUESTS_PER_SEC);
+        const QString host = request.first.host();
+        // Don't dispatch requests to a host that is currently in its 429 backoff.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(host, now)) {
```
ZeldaZach, line 206: Skipping backed-off hosts here silently bypasses the fall-through logic that lives in `CardPictureLoaderWorkerWork`.

Before this change, a queued request to a host that entered backoff after it was enqueued would still be dispatched, get its 429, and `handleFailedReply()` would either `scheduleDeferredRetry()` (round 1) or `picDownloadFailed()` (round >= 2, i.e. move on to the other configured URLs/sets). Now the request just sits in `requestLoadQueue` and its `CardPictureLoaderWorkerWork` is never told anything: it has no timer of its own (it is only waiting for a `QNetworkReply::finished` that will never come), so it cannot fall through to an alternative source and cannot conclude.

Concrete scenario: a set download queues ~200 images from scryfall. Worker A gets a 429, `on429()` sets a 60s deadline. The other ~199 entries are now skipped every dispatch tick. After 60s the backoff lapses, up to 10 of them go out in the next second, all 429 again, `on429()` pushes the deadline another 60s, and the cycle repeats — `retryRounds` also never resets because `last429` keeps being refreshed inside `RESET_GRACE_MS`. The cards at the tail of the queue stay blank for minutes with `currentlyLoading` still holding their key (so nothing re-requests them), where previously they'd have fallen through to the next configured picture URL within seconds.

If the intent is "don't waste requests on a backed-off host", the queued entries should be handed back to their worker (e.g. drop them from the queue and call something that ends up in `startNextPicDownload()` / `scheduleDeferredRetry()`) rather than left parked in the dispatch queue indefinitely.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -138,6 +138,9 @@ QNetworkReply *CardPictureLoaderWorker::makeRequest(const QUrl &url, CardPicture
 void CardPictureLoaderWorker::resetRequestQuota()
 {
     requestQuota = MAX_REQUESTS_PER_SEC;
+    // Allowances are seeded per host on demand in processSingleRequest(), so a
```
ZeldaZach: This comment (and the PR description) doesn't match what the code does. The lazy seed in `processSingleRequest()` uses `hostRequestQuota.value(host, MAX_REQUESTS_PER_SEC)` — exactly the same expression the removed eager loop used. So a host that was rate limited last second still gets a full fresh allowance (its current `hostRequestQuota`, which is only reduced by `onHostRateLimited()`) the first time it is looked at in the new second; the value is merely computed later.

The one real behavioural change is the opposite direction and is a genuine fix: a host that was *not* in `requestLoadQueue` at the tick used to fall through to `hostQuotaRemaining.value(host, MAX_REQUESTS_PER_SEC)` and get the full 10/s regardless of its halved `hostRequestQuota`; now it gets its reduced allowance. Worth rewording the comment to say that, otherwise the next reader will assume backoff-aware seeding that isn't there.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -184,10 +182,20 @@ void CardPictureLoaderWorker::dispatchQueuedRequest()
...
 {
+    QDateTime now = QDateTime::currentDateTime();
     for (int i = 0; i < requestLoadQueue.size(); ++i) {
         const auto &request = requestLoadQueue.at(i);
-        QString host = request.first.host();
-        int allowance = hostQuotaRemaining.value(host, MAX_REQUESTS_PER_SEC);
+        const QString host = request.first.host();
+        // Don't dispatch requests to a host that is currently in its 429 backoff.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(host, now)) {
+            continue;
```
ZeldaZach: Minor hole in the new guard: `makeRequest()` itself re-enters with a cached redirect (`getCachedRedirect()` at the top of `makeRequest`) and issues that request without any backoff or per-host allowance check. If a redirect for this URL got cached after the entry was queued — which is the normal scryfall API -> image CDN pattern — the dispatched request targets a different host that may itself be in 429 backoff, and neither `isRateLimited()` nor `hostQuotaRemaining` is consulted for it. The `queueRequest()` cached-redirect path handles this correctly by re-queuing under the redirect host; the `makeRequest()` path does not.

Also, `rateLimiter()` could return `const ServerRateLimiter &` here — the dispatcher only calls the const `isRateLimited()`, and a non-const public accessor to shared mutable static state invites future callers to mutate it from outside the worker.

ZeldaZach (changes requested):
The per-host lazy seeding in resetRequestQuota()/processSingleRequest() looks right, but the new backoff skip at card_picture_loader_worker.cpp:190 has a gap worth fixing before merge. See the inline comments for detail.

The blocking one: skipping a backed-off host at dispatch time strands the entry that is already sitting in requestLoadQueue. The CardPictureLoaderWorkerWork that queued it has no timer pending, it is only waiting on a QNetworkReply::finished that will now never arrive, so the 429 fall-through in handleFailedReply() and the pre-dispatch check in startNextPicDownload() are both unreachable for that entry. With MIN_429_BACKOFF_MS at 30s and on429() re-extending the deadline every time the next batch goes out and gets a 429, that card stays blank for many multiples of 30s while currentlyLoading still holds its key, so nothing re-requests it. Before this change it fell through to an alternative source within seconds. Either skip the entry and let the work object reschedule itself (for example by handing it back so it can call scheduleDeferredRetry()), or drive the queue from the backoff deadline rather than only from the 1s quota tick.

Two smaller ones, not blocking:

card_picture_loader_worker.cpp:141, the new comment and the PR description do not match the code. The lazy seed uses hostRequestQuota.value(host, MAX_REQUESTS_PER_SEC), the same expression the removed eager loop used, so a rate-limited host still gets its full recorded allowance at the start of a new second. The real improvement here is that hosts absent from the queue at tick time no longer default to the full 10/s.

card_picture_loader_worker.cpp:191, makeRequest() re-enters itself on a cached redirect and dispatches to a possibly different host with no backoff or allowance check, so the new guard is bypassed in exactly the API-to-CDN redirect case that queueRequest() handles correctly. rateLimiter() could also return a const reference, since only the const isRateLimited() is called on it.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: skipped backed-off entries are now handed back to their worker (dropped from the queue and returned to startNextPicDownload) so they can fall through instead of parking indefinitely. (310caa7dc0)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: comment reworded to describe the real change: hosts not in the queue now get their (possibly reduced) per-host allowance instead of a fresh one. (310caa7dc0)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the cached-redirect path in makeRequest now checks the redirect host backoff and calls scheduleDeferredRetry; rateLimiter() returns a const reference. (310caa7dc0)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -188,10 +196,24 @@ void CardPictureLoaderWorker::dispatchQueuedRequest()
...
         const auto &request = requestLoadQueue.at(i);
-        QString host = request.first.host();
-        int allowance = hostQuotaRemaining.value(host, MAX_REQUESTS_PER_SEC);
+        const QString host = request.first.host();
+        // Don't dispatch requests to a host that is currently in its 429 backoff; hand the entry
+        // back to its worker so it can wait the backoff out or fall through to another source,
+        // instead of leaving it parked in the queue with no reply pending.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(host, now)) {
+            requestLoadQueue.removeAt(i);
+            request.second->startNextPicDownload();
```
ZeldaZach: `request` is a reference into `requestLoadQueue`, and `removeAt(i)` on the line above invalidates it, so `request.second` is read from an element that no longer exists at that index.

`QList::removeAt()` either advances the begin pointer (when `i == 0`) or `std::move`s the tail down into slot `i` (when `i > 0`). So:

- `i == 0`: `request` refers to the destroyed head slot. The raw `CardPictureLoaderWorkerWork*` bits usually survive, so it happens to work, but it is a read of a destroyed object.
- `i > 0` (reached whenever an earlier entry is out of allowance and a later one is backed off, i.e. the mixed-host case this PR is specifically about): `request` now aliases the *next* queue entry, and `startNextPicDownload()` is called on the **wrong** work object.

The `i > 0` case reintroduces the bug the previous round asked to fix, plus a new one:

1. The backed-off entry that was just removed is never handed back to its own work object. That object has no timer and is only waiting on a `QNetworkReply::finished` that will never arrive, so the card stays blank permanently while `currentlyLoading` still holds its key.
2. The neighbouring work object gets a spurious `startNextPicDownload()` while its own entry is still queued, so it emits `requestImageDownload` again and ends up with two in-flight requests for the same card.

Taking a copy (or just the pointer) before mutating the queue fixes it:

```cpp
auto *entryWorker = requestLoadQueue.at(i).second;
requestLoadQueue.removeAt(i);
entryWorker->startNextPicDownload();
return true;
```

The `makeRequest()` branch below is fine as written since it uses `request` before the `removeAt()`, but having both branches work off the same copy would keep that ordering from being load-bearing.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -108,6 +108,14 @@ QNetworkReply *CardPictureLoaderWorker::makeRequest(const QUrl &url, CardPicture
     QUrl cachedRedirect = getCachedRedirect(url);
     if (!cachedRedirect.isEmpty()) {
         emit imageRequestSucceeded(url);
+        // The redirect target is a different host, which may itself be in 429 backoff; hand the
+        // entry back to its worker so it waits the backoff out instead of dispatching straight
+        // onto the backed-off host.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(cachedRedirect.host(),
+                                                                     QDateTime::currentDateTime())) {
+            worker->scheduleDeferredRetry();
```
ZeldaZach: Non-blocking: `scheduleDeferredRetry()` picks its wait from `cardToDownload.getCurrentUrl()`'s host, which here is the *original* URL, not the `cachedRedirect.host()` we just found to be backed off. The original host is typically not rate limited, so the deferral falls through to `earliestDeadline(now)`, which can be much shorter than the redirect host's own deadline. The retry then resolves the same cached redirect, lands here again, and reschedules.

It is self-limiting (each cycle waits at least the earliest deadline plus up to 5s of jitter, and it clears once the backoff lapses), so no tight spin, just extra churn and a repeated `imageRequestSucceeded` emission to the status bar for one queued request. Passing the blocking host into the deferral, or checking the redirect host before the `emit imageRequestSucceeded(url)` above, would make it wait the right amount.

ZeldaZach (changes requested):
All three points from the previous round are addressed: backed-off entries are handed back to their work object instead of parked, the resetRequestQuota() comment now describes the real behaviour change, and rateLimiter() returns a const reference with the cached-redirect path guarded.

One blocker remains, and it is a one-line fix. In processSingleRequest(), `const auto &request` is a reference into requestLoadQueue and `removeAt(i)` invalidates it before `request.second->startNextPicDownload()` runs. For i == 0 that is a read of a destroyed element; for i > 0 the reference aliases the next queue entry, so the hand-back goes to the wrong work object. That resurrects the exact failure the previous round flagged (the removed entry's work object is never told anything, has no timer, and waits forever on a finished() that will not come, leaving the card blank with currentlyLoading still holding its key) and additionally gives the neighbouring work object two in-flight requests for the same card. i > 0 is reached in the mixed-host case this PR targets: an earlier entry out of allowance followed by a backed-off one.

Copy the pointer before mutating the queue and this is good to go. Details inline, plus one non-blocking note on the deferred retry in makeRequest() waiting on the original host's deadline rather than the redirect host's.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: both dispatch branches now pull the entry out of the queue with `QList::takeAt(i)` before touching it, so `request` is no longer a reference that outlives the `removeAt()`. The backed-off entry is handed back to its own worker (`entry.second->startNextPicDownload()`), and the `makeRequest()` branch works off the same copy, so the read-before-mutate ordering is no longer load-bearing. (7794fb4fba)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: `scheduleDeferredRetry()` now takes an optional `preferredHost` and waits on that server's `deadline()` instead of the current URL's host when one is handed in. The `makeRequest()` cached-redirect backoff path passes `cachedRedirect.host()` (and the `handleFailedReply()` 429 path passes the reply's host, which had the same mismatch after a redirect), so the deferral waits out the host that is actually rate limited instead of falling through to `earliestDeadline()`. The `emit imageRequestSucceeded(url)` was also moved after the backoff check, so a deferred request that is never actually sent no longer counts as a successful load in the status bar. (7794fb4fba)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -212,14 +220,28 @@ void CardPictureLoaderWorker::updateTimerState()
...
         const auto &request = requestLoadQueue.at(i);
-        QString host = request.first.host();
-        int allowance = hostQuotaRemaining.value(host, MAX_REQUESTS_PER_SEC);
+        const QString host = request.first.host();
+        // Don't dispatch requests to a host that is currently in its 429 backoff; hand the entry
+        // back to its worker so it can wait the backoff out or fall through to another source,
+        // instead of leaving it parked in the queue with no reply pending.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(host, now)) {
+            auto entry = requestLoadQueue.takeAt(i);
+            entry.second->startNextPicDownload();
```
ZeldaZach: Hand-back can loop forever when the queued URL is a cached-redirect target.

`processSingleRequest()` tests the **queued** URL's host, but `startNextPicDownload()` tests `cardToDownload.getCurrentUrl()`'s host. Those differ whenever the queue entry came from a redirect: `queueRequest()` resolves `redirectCache[U] -> B` before appending, and `handleSuccessfulReply()` emits `requestImageDownload(redirectUrl)` while `cardToDownload` still points at the pre-redirect URL on host A.

So: U on host A redirects to host B; B gets a 429 from another worker while the entry sits in the queue. Dispatch pops it because B is backed off, calls `startNextPicDownload()`, which sees host A (not limited) and re-emits `requestImageDownload(U)` -> `queueRequest` re-resolves the cached redirect to B -> re-appends. That repeats every 100ms tick for the whole 30-60s backoff window instead of waiting it out.

Each cycle also re-emits `imageRequestQueued`, so several hundred status widgets get allocated per stuck card and `progressBar`'s maximum is inflated by the same amount (see the note on line 116 for why they're never reclaimed).

Suggest passing the blocked host into the hand-back, or calling `entry.second->scheduleDeferredRetry(host)` directly here.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -212,14 +220,28 @@ void CardPictureLoaderWorker::updateTimerState()
...
+            return true;
```
ZeldaZach: A backed-off entry consumes the entire dispatch tick.

This branch does `return true` instead of continuing the scan, so when entries for a backed-off host sit at the head of the queue, every 100ms tick is spent recycling exactly one of them and healthy hosts further down get nothing. With ~500 queued entries for a 429'd host, images from every other host stall for ~50s.

The entry has already been removed from the list, so `--i; continue;` would let the loop keep looking for a dispatchable host — which is what the PR description says the loop now does.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -107,6 +107,14 @@ QNetworkReply *CardPictureLoaderWorker::makeRequest(const QUrl &url, CardPicture
     // Check for cached redirects
     QUrl cachedRedirect = getCachedRedirect(url);
     if (!cachedRedirect.isEmpty()) {
+        // The redirect target is a different host, which may itself be in 429 backoff; hand the
+        // entry back to its worker so it waits the backoff out instead of dispatching straight
+        // onto the backed-off host.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(cachedRedirect.host(),
+                                                                     QDateTime::currentDateTime())) {
+            worker->scheduleDeferredRetry(cachedRedirect.host());
+            return nullptr;
```
ZeldaZach, line 119: This early return skips `emit imageRequestSucceeded(url)`.

On the normal cached-redirect path (line 118) the original URL gets marked finished, and that's what lets `CardPictureLoaderStatusBar::cleanOldEntries()` reclaim the log widget and decrement the progress bar's max/value after 10s. The backoff branch returns before that emit, so the `CardPictureLoaderRequestStatusDisplayWidget` for that URL is never marked finished and never cleaned, and the progress bar maximum stays permanently inflated.

Combined with the churn from the line 232 issue, that's unbounded widget growth over a long backoff.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -212,14 +220,28 @@ void CardPictureLoaderWorker::updateTimerState()
...
+            return true;
+        }
+        // Seed the allowance now so a host that was rate limited gets its reduced
+        // allowance instead of a fresh full quota mid-second.
+        if (!hostQuotaRemaining.contains(host)) {
+            hostQuotaRemaining.insert(host, hostRequestQuota.value(host, MAX_REQUESTS_PER_SEC));
+        }
+        int allowance = hostQuotaRemaining.value(host);
         if (allowance > 0) {
             hostQuotaRemaining.insert(host, allowance - 1);
```
ZeldaZach: Minor: the allowance is spent on a request that may never be sent.

`hostQuotaRemaining` is decremented here, before `makeRequest()`, and `makeRequest()` can now return `nullptr` without issuing anything (cached redirect into a backed-off host). The host loses one of its per-second slots for a no-op. Worth moving the decrement onto the path that actually reaches `networkManager->get()`.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -212,14 +220,28 @@ void CardPictureLoaderWorker::updateTimerState()
...
+    QDateTime now = QDateTime::currentDateTime();
     for (int i = 0; i < requestLoadQueue.size(); ++i) {
         const auto &request = requestLoadQueue.at(i);
-        QString host = request.first.host();
-        int allowance = hostQuotaRemaining.value(host, MAX_REQUESTS_PER_SEC);
+        const QString host = request.first.host();
+        // Don't dispatch requests to a host that is currently in its 429 backoff; hand the entry
+        // back to its worker so it can wait the backoff out or fall through to another source,
+        // instead of leaving it parked in the queue with no reply pending.
+        if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(host, now)) {
```
ZeldaZach: This guard also blocks requests that would never touch the network.

When `picDownload` is false — or the cache method isn't `NETWORK_CACHE`, so the enqueue-time cache shortcut in `queueRequest()` is skipped — `makeRequest()` sets `AlwaysCache` and serves straight from the disk cache with no round trip. The new check bounces those entries into a 30-60s deferred retry purely because the host 429'd earlier, so cached images stop appearing even with downloads disabled.

Probably wants to be gated on the request actually going to the network.

ZeldaZach (commented):
Re-review of the 429-backoff skip. The lazy per-host seeding and the `preferredHost` plumbing both look right; the dispatcher's hand-back path is where the problems are. Five inline notes below.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the dispatch hand-back now calls entry.second->scheduleDeferredRetry(host) with the queued URLu0027s host (the cached-redirect target that is actually in backoff) instead of startNextPicDownload(), which was testing cardToDownload.getCurrentUrl()u0027s host and re-emitting requestImageDownload on the pre-redirect host every 100ms tick. The entry now waits out the blocked hostu0027s deadline once per backoff window instead of looping, and each cycle no longer re-emits imageRequestQueued to allocate a fresh status widget. (85e109c350)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the hand-back branch now does --i; continue; so the dispatch tick keeps scanning for a host that is neither backed off nor out of allowance, instead of returning after recycling a single entry. A 429u0027d host at the head of the queue no longer stalls the hosts further down. (85e109c350)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: makeRequest() now emits imageRequestSucceeded(url) before the deferred-retry return on the cached-redirect backoff path, so CardPictureLoaderStatusBar::cleanOldEntries() can reclaim that URLu0027s widget and the progress bar maximum is not permanently inflated. (85e109c350)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the per-second allowance is now only decremented when makeRequest() actually issues a request. makeRequest() returns nullptr when the cached redirect target is in backoff and hands the entry back, so that host no longer loses a slot for a no-op; the decrement is on the path that reaches networkManager->get(). (85e109c350)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the dispatch-time backoff skip is now gated on requestTouchesNetwork(), computed from the same cache-control logic makeRequest() uses, so entries that will be served straight from the disk cache (e.g. with downloads disabled) no longer get bounced into a 30-60s deferred retry purely because the host 429u0027d earlier. makeRequest() reuses the helper for its useNetworkCache decision so the two canu0027t drift. (85e109c350)

ZeldaZach (approved):

PR #7287, opened 2026-09-12
[PictureLoader] Add user-configurable per-host request caps

BruebachL:

Every card picture host was throttled identically: 10 requests/second per
host, no per-host tuning, and no way for users to tune a URL they added
themselves. That forces one pace for `gatherer.wizards.com`, the Scryfall
API (which rate-limits aggressively at 10/s) and the Scryfall CDN (which is a
static CDN that doesn't need throttling at all), so downloads either hammer a
rate-sensitive API or artificially slow down a fast CDN.

- Add a developer-set per-host cap table in `DownloadSettings`:
  - `api.scryfall.com` → 9 req/s (one under the documented 10 req/s API limit)
  - `cards.scryfall.io` → **unlocked** (never throttled per host)
  - all other hosts → the global default of 10 req/s
- Add per-host persistence in `downloads.ini` (`hostRequestLimits` key) plus
  `getHostRequestLimits`/`setHostRequestLimits` accessors and the
  `hostRequestLimitsChanged` signal.
- Let users **lower** any host's allowance (never raise it above the developer
  cap) from the Deck Editor → Download settings page. Each URL in the list gets
  an "Adjust Rate Limit" button ([cogwheel icon]) that opens
  [QInputDialog::getInt](https://doc.qt.io/qt-6/qinputdialog.html#getInt) with
  the range `1..developer cap` — or `0..50` where `0` means "unlimited" for
  unlocked hosts — and stores the value only when it differs from the default.
- Make the worker honour the caps:
  - A `hostAllowanceCeiling(host)` helper resolves the effective ceiling
    (developer cap vs user override, clamped by
    [`qBound`](https://doc.qt.io/qt-6/qtglobal.html#qBound)).
  - Unlocked hosts skip the 100 ms dispatch pacing altogether: their queued
    requests fire back-to-back, bounded only by the shared 429 backoff window
    and Qt's per-host connection pool, so a CDN can pin the network.
  - Allocation, recovery and halving all clamp against the effective ceiling,
    so lowering a host's limit applies immediately rather than waiting for a
    reset tick.
  - The worker subscribes to `hostRequestLimitsChanged` and reloads the table,
    so edits in the settings page (or user-edited `downloads.ini`) take effect
    with no restart.
- Add Google Test coverage
  ([tests/settings/settings_defaults_test.cpp](tests/settings/settings_defaults_test.cpp))
  for the new setting's default, round-trip storage, and the clamp table
  against every developer cap.

## Why

This stack ends in a manual benchmark against the real card hosts
(`network-requests/benchmark`); the truthful caps here are what make those
numbers meaningful.


cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -167,6 +180,27 @@ void CardPictureLoaderWorker::processQueuedRequests()
...
 void CardPictureLoaderWorker::dispatchQueuedRequest()
 {
+    if (requestLoadQueue.isEmpty()) {
+        dispatchTimer.stop();
+        return;
+    }
+
+    // Unlocked hosts (developer cap UNLIMITED_HOST_QUOTA) skip the dispatch pacing and the
+    // global per-second quota: dispatch every queued request for them back-to-back, bounded
+    // only by their 429 backoff window and Qt's per-host connection pool.
```
ZeldaZach: This loop dispatches **every** queued request for an unlocked host in a single 100 ms tick, with no bound on how many go out at once.

Loading a large deck or browsing a set queues hundreds to thousands of `cards.scryfall.io` URLs. On the next tick they are all handed to `networkManager->get()` back to back. Qt only opens ~6 concurrent connections per host, so the rest sit inside QNetworkAccessManager waiting for a free slot — but the 30 s transfer timeout installed by `networkManager->setTransferTimeout()` in the constructor starts when the reply is created, not when a socket becomes available. Anything past roughly the first few hundred in the burst will hit the timeout and fail before it ever transmits a byte, and the user sees those cards fail to download.

The previous behaviour kept the number of in-flight replies bounded by the dispatch pacing. Consider keeping a cap on concurrent in-flight requests for unlocked hosts (e.g. dispatch at most N per tick, or track outstanding replies) rather than draining the whole queue.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -206,9 +242,25 @@ bool CardPictureLoaderWorker::processSingleRequest()
...
+    return SettingsCache::instance().downloads().clampHostRequestLimit(host, requested);
+}
+
 void CardPictureLoaderWorker::onHostRateLimited(const QString &host)
 {
-    hostRequestQuota.insert(host, qMax(MIN_HOST_QUOTA, hostRequestQuota.value(host, MAX_REQUESTS_PER_SEC) / 2));
+    const int ceiling = hostAllowanceCeiling(host);
+    if (ceiling == DownloadSettings::UNLIMITED_HOST_QUOTA) {
+        // Unlocked hosts have no per-host allowance to halve; the shared backoff
+        // window tracked by the rate limiter still paces them.
```
ZeldaZach: Bailing out here means an unlocked host has **no adaptive response to a 429 at all** — the only thing that changes is the shared backoff window.

Concrete sequence for `cards.scryfall.io`: the whole queue bursts out (see the dispatch loop above), the CDN 429s, `on429` sets `retryRounds = 1` and a backoff deadline. The moment that deadline expires, `dispatchQueuedRequest` fires the *entire remaining backlog* again unchanged, because nothing lowered the host's pace. That reliably produces a second 429, which pushes `retryRounds` to 2. At that point `CardPictureLoaderWorkerWork::startNextPicDownload` stops waiting out the backoff and starts skipping the URL (`rounds(host) == 1` branch no longer taken), so every card falls through to `api.scryfall.com` and eventually Gatherer — the fast CDN is abandoned for the rest of the session.

For throttled hosts the halving in the line below breaks that cycle. Unlocked hosts need an equivalent: e.g. on a 429, drop the host out of the unlocked path by installing a concrete allowance in `hostRequestQuota` (and let the existing recovery walk it back up).

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -190,10 +224,12 @@ bool CardPictureLoaderWorker::processSingleRequest()
...
             continue;
         }
+        const int ceiling = hostAllowanceCeiling(host);
         // Seed the allowance only now, so a host that was rate limited last second
-        // doesn't get a fresh full quota the moment it is queried mid-second.
+        // doesn't get a fresh full quota the moment it is queried mid-second. Clamp
+        // against the ceiling so a lowered user cap applies from this second onward.
         if (!hostQuotaRemaining.contains(host)) {
-            hostQuotaRemaining.insert(host, hostRequestQuota.value(host, MAX_REQUESTS_PER_SEC));
+            hostQuotaRemaining.insert(host, qMin(ceiling, hostRequestQuota.value(host, ceiling)));
```
ZeldaZach, line 292: `ceiling` can be `UNLIMITED_HOST_QUOTA` (-1) here, and then this seeds `hostQuotaRemaining[host] = qMin(-1, -1) = -1`. `allowance > 0` is false, so the request is skipped, and the -1 stays cached until the next `resetRequestQuota()` — the host is locked out of this path for the remainder of the second.

It is reachable: `dispatchQueuedRequest` skips an unlocked host only when `isRateLimited(host, now)` is true for the `now` it computed; `processSingleRequest` recomputes `now` microseconds later, so a backoff deadline falling in that gap lets the unlocked host reach this line. The sentinel should be handled explicitly (skip unlocked hosts, or treat -1 as "no limit") rather than flowing into arithmetic as a magic negative.

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
```diff
@@ -164,6 +172,54 @@ void DeckEditorSettingsPage::storeSettings()
...
+    const int devCap = devCaps.value(host, DownloadSettings::DEFAULT_HOST_REQUEST_LIMIT);
+    const bool unlocked = devCap == DownloadSettings::UNLIMITED_HOST_QUOTA;
+
+    bool ok = false;
+    int minimum;
+    int maximum;
+    int defaultValue;
+    if (unlocked) {
+        minimum = 0; // 0 means "unlimited"
+        maximum = UNLOCKED_HOST_LIMIT_MAX;
```
ZeldaZach: The 0..50 range presented for an unlocked host is not a monotonic scale, so the dialog actively misleads the user.

Storing *any* non-zero value for `cards.scryfall.io` makes `hostAllowanceCeiling` stop returning `UNLIMITED_HOST_QUOTA`, which drops the host out of the unlocked fast path in `dispatchQueuedRequest` and back onto the paced path: one request per `DISPATCH_INTERVAL_MS` (100 ms) tick, bounded by the global `requestQuota` of 10/s. So a user who picks 50 — the apparent maximum — gets roughly 10 req/s, dramatically *slower* than leaving it at 0. Every value from 10 to 50 behaves identically, and every value above 0 is a downgrade.

Either clamp the dialog's maximum to something the dispatcher can actually deliver, or label 0 explicitly as "unlimited (fastest)" and cap the selectable range at `DEFAULT_HOST_REQUEST_LIMIT`.

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
```diff
@@ -164,6 +172,54 @@ void DeckEditorSettingsPage::storeSettings()
...
+        defaultValue = currentLimits.value(host, 0);
+    } else {
+        minimum = DownloadSettings::MIN_HOST_REQUEST_LIMIT;
+        maximum = devCap;
+        defaultValue = currentLimits.value(host, devCap);
+    }
+
+    const int value = QInputDialog::getInt(this, tr("Adjust Rate Limit for %1").arg(host),
```
ZeldaZach: For an unlocked host this prints "developer maximum is 50", but 50 is `UNLOCKED_HOST_LIMIT_MAX`, a local UI constant — the developer cap for such a host is `UNLIMITED_HOST_QUOTA`, i.e. no maximum. The label also never explains that 0 means "unlimited", which is the only way back to the default once a value has been stored. A user who sets 5 and then wants the old behaviour back has no indication that 0 is the answer.

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
```diff
@@ -164,6 +172,54 @@ void DeckEditorSettingsPage::storeSettings()
     SettingsCache::instance().downloads().setDownloadUrls(downloadUrls);
 }
 
+void DeckEditorSettingsPage::actAdjustRateLimit()
```
ZeldaZach, line 247: The limit is keyed by host but the UI is per-URL, and nothing reconciles the two.

Two of the default URLs share `api.scryfall.com`, so adjusting the rate limit on one silently changes the other with no indication in the list. And if the user later edits or removes that URL, the `hostRequestLimits` entry is left behind in `downloads.ini` forever, silently throttling the host if the URL is ever re-added. Showing the current effective limit next to each URL (and pruning entries for hosts no longer referenced on `storeSettings`) would make the state visible.

libcockatrice_settings/libcockatrice/settings/download_settings.cpp
```diff
@@ -50,3 +66,32 @@ void DownloadSettings::setDownloadSpoilerStatus(bool _spoilerStatus)
...
+    return hostRequestLimits;
+}
+
+void DownloadSettings::setHostRequestLimits(const QHash<QString, int> &hostRequestLimits)
+{
+    QVariantMap stored;
+    for (auto it = hostRequestLimits.cbegin(); it != hostRequestLimits.cend(); ++it) {
+        stored.insert(it.key(), it.value());
+    }
+    setValue(stored, "hostRequestLimits");
```
ZeldaZach: QSettings has no native INI representation for `QVariantMap`, so this is written through `QSettingsPrivate::variantToString`'s fallback as an opaque `@Variant(...)` QDataStream blob. It round-trips correctly (the new test passes), but the PR description's claim that "user-edited `downloads.ini`" is a supported way to set these values does not hold — the key is unreadable and uneditable by hand.

If hand-editing is a goal, store one key per host under a `hostRequestLimits/` subgroup (host names contain dots, so they need `beginGroup`/`setValue` with the host as key rather than a slash-joined path) so each lands as a plain `host=value` line.

ZeldaZach (changes requested):
Reviewed the per-host cap changes. Two confirmed logic bugs plus some UI behavior that needs a rethink. Details are in the inline comments.

Blocking:

1. card_picture_loader_worker.cpp:232. hostAllowanceCeiling can return the UNLIMITED_HOST_QUOTA sentinel (-1) for an unlocked host, and processSingleRequest feeds it straight into qMin(ceiling, hostRequestQuota.value(host, ceiling)). That seeds hostQuotaRemaining[host] = -1, so allowance is never greater than 0 and the host is skipped for the rest of the second. It is reachable whenever an unlocked host reaches processSingleRequest instead of the fast path in dispatchQueuedRequest, for example when a 429 backoff deadline expires between the now in dispatchQueuedRequest and the one in processSingleRequest. The sentinel needs to be handled before the qMin, not passed through it.

2. deck_editor_settings_page.cpp:199. The 0 to 50 range offered for unlocked hosts is not monotonic. Any non-zero value makes hostRequestLimits.contains(host) true, so hostAllowanceCeiling stops returning the sentinel and the host drops off the unlocked fast path onto the 100 ms paced path, which tops out near 10 requests per second. Picking 50 is therefore much slower than picking 0, and every value from about 11 to 50 behaves identically. A user raising the number to go faster gets the opposite.

Also worth addressing:

3. card_picture_loader_worker.cpp:260. onHostRateLimited returns early for unlocked hosts, so nothing adapts after a 429. The backlog re-fires as soon as the backoff window expires and earns a second 429, which pushes retryRounds up and makes startNextPicDownload abandon the CDN URL for the slower fallbacks.

4. card_picture_loader_worker.cpp:190. The unlocked-host loop drains the whole queue in a single tick with no concurrency bound. Since setTransferTimeout starts its clock at reply creation while Qt opens only about six connections per host, the tail of a large deck load can time out before it ever transmits.

5. deck_editor_settings_page.cpp:175 and :207. The limit is keyed by host but the dialog is driven by the selected URL, so the two default URLs that share api.scryfall.com silently edit each other, and limits for removed URLs stay behind in downloads.ini. The prompt also calls 50 the developer maximum when it is the local UNLOCKED_HOST_LIMIT_MAX constant, and it never mentions that 0 means unlimited, which is the only way back to the default.

6. download_settings.cpp:86. A QVariantMap has no native INI representation, so hostRequestLimits is written as an opaque @Variant(...) blob. That contradicts the PR description's claim that users can hand-edit the values in downloads.ini.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: added a per-host in-flight cap (MAX_IN_FLIGHT_PER_HOST) so unlocked hosts no longer drain the whole queue in one tick. (21ca53962d)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: unlocked hosts now adapt to 429s by installing a concrete allowance in hostRequestQuota and walking it back up during recovery, instead of only moving the shared backoff window. (21ca53962d)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: the -1 sentinel is now handled explicitly instead of flowing into the qMin arithmetic. (21ca53962d)

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
BruebachL: Fixed: the selectable range is capped at DEFAULT_HOST_REQUEST_LIMIT and 0 is labelled as unlimited (fastest). (b3b930587b)

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
BruebachL: Fixed: the helper text no longer claims a developer maximum of 50 for unlocked hosts and explains that 0 means unlimited. (b3b930587b)

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
BruebachL: Fixed: each URL row now shows the effective limit for its host (raw URL in UserRole), and storeSettings prunes limits for hosts no longer referenced. (b3b930587b)

libcockatrice_settings/libcockatrice/settings/download_settings.cpp
BruebachL: Fixed: per-host limits are stored one key per host under a hostRequestLimits/ subgroup, so downloads.ini stays hand-editable. (b3b930587b)

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
```diff
@@ -158,10 +171,128 @@ void DeckEditorSettingsPage::storeSettings()
...
+    // Drop per-host limits whose host is no longer referenced by any configured URL, so removing
+    // a URL doesn't leave a stale throttle behind that reactivates if the host is re-added.
+    QSet<QString> usedHosts;
+    for (const QString &url : downloadUrls) {
+        const QString host = QUrl(url).host();
+        if (!host.isEmpty()) {
+            usedHosts.insert(host);
+        }
+    }
+    QHash<QString, int> limits = SettingsCache::instance().downloads().getHostRequestLimits();
```
ZeldaZach, line 189: The prune keeps only hosts that literally appear in the configured URL list, but the host that actually gets throttled is often a redirect target that never appears there. `api.scryfall.com` 302s to `cards.scryfall.io`, and `queueRequest()`/`makeRequest()` re-queue against the redirect target, so the limit that matters is the one on `cards.scryfall.io`. A user who deletes the default `cards.scryfall.io` URL (keeping only the `api.scryfall.com` entries) and has a stored limit for the CDN loses it the next time anything calls `storeSettings()` - adding, editing, removing or reordering any URL. Consider keeping limits for hosts that carry a developer cap, or only pruning the host whose URL was just removed.

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
```diff
@@ -158,10 +171,128 @@ void DeckEditorSettingsPage::storeSettings()
...
+    const bool unlocked = devCap == DownloadSettings::UNLIMITED_HOST_QUOTA;
+
+    bool ok = false;
+    int minimum;
+    int maximum;
+    int defaultValue;
+    QString prompt;
+    if (unlocked) {
+        minimum = 0; // 0 means "unlimited"
+        maximum = DownloadSettings::DEFAULT_HOST_REQUEST_LIMIT;
```
ZeldaZach: For an unlocked host the dialog maximum is `DEFAULT_HOST_REQUEST_LIMIT` (10), but `clampHostRequestLimit()` puts no upper bound on unlocked hosts (`qMax(MIN, requested)`), and the settings are documented as hand-editable. So if a user writes `hostRequestLimits/cards.scryfall.io=30` into `downloads.ini`, `defaultValue` is 30 while `maximum` is 10; `QInputDialog::getInt()` clamps the initial value to the range, the dialog opens showing 10, and pressing OK silently rewrites the stored limit from 30 to 10. The PR description also says this range should be `0..50`, so the constant here looks wrong. Either raise `maximum` (50, per the description) or clamp the stored value on read so the UI and `clampHostRequestLimit()` agree.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -136,8 +140,16 @@ QNetworkReply *CardPictureLoaderWorker::makeRequest(const QUrl &url, CardPicture
...
     QNetworkReply *reply = networkManager->get(req);
 
+    // Track in-flight replies per host so the unlocked fast path can bound how many requests it
+    // issues at once, instead of creating replies that time out before Qt opens a connection.
+    const QString host = url.host();
+    hostInFlight.insert(host, hostInFlight.value(host) + 1);
+
     // Connect reply handling
-    connect(reply, &QNetworkReply::finished, worker, [reply, worker] { worker->handleNetworkReply(reply); });
+    connect(reply, &QNetworkReply::finished, worker, [this, reply, worker, host] {
```
ZeldaZach: The `hostInFlight` decrement rides on a connection whose receiver/context object is `worker`, not `this` or `reply`. `CardPictureLoaderWorkerWork` deletes itself (`concludeImageLoad()` -> `deleteLater()`), and any such deletion while a reply is still pending severs this connection, so the increment from line 146 is never undone. Each leaked count permanently lowers the unlocked hosts concurrency; leak `MAX_IN_FLIGHT_PER_HOST` (6) of them and `hostInFlight.value(host) < MAX_IN_FLIGHT_PER_HOST` is false forever - the fast path stops dispatching, `unlockedCapped` keeps `dispatchTimer` spinning at 100ms, and queued images for that host never load again for the rest of the session. Safer to book-keep against an object with the right lifetime, e.g. connect the decrement separately with `this` (or `reply`) as the context object, or use `QObject::connect(reply, &QNetworkReply::destroyed, this, ...)` so the slot always runs.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -188,8 +213,49 @@ void CardPictureLoaderWorker::dispatchQueuedRequest()
...
+    // connection for them.
+    for (int i = 0; i < requestLoadQueue.size();) {
+        const auto &request = requestLoadQueue.at(i);
+        const QString host = request.first.host();
+        if (isUnlockedHost(host)) {
+            if (CardPictureLoaderWorkerWork::rateLimiter().isRateLimited(host, now)) {
+                ++i;
+                continue;
+            }
+            if (hostInFlight.value(host) < MAX_IN_FLIGHT_PER_HOST) {
```
ZeldaZach, line 233: The in-flight gate is checked against `host` (the queued URLs host), but `makeRequest()` increments the counter for the host it actually issues against. When a cached redirect points the unlocked hosts URL at a different host, `makeRequest()` recurses and increments `hostInFlight[redirectHost]`, leaving `hostInFlight[host]` at 0. This loop then sees `0 < 6` on every iteration and drains the entire queue in a single 100ms tick - exactly the unbounded burst the cap is meant to prevent, and aimed at a host that may have its own developer cap. Gate on the host `makeRequest()` will really use (resolve the cached redirect before the check), or have `makeRequest()` return the host it dispatched to.

ZeldaZach (approved):
Re-reviewed after the fixes. All six findings from the previous round are genuinely addressed in code, not just in replies:

1. The UNLIMITED_HOST_QUOTA sentinel no longer reaches the `qMin` in `processSingleRequest` — unlocked hosts `continue` early, and the one remaining path where the ceiling is -1 (a 429 installed a concrete allowance) substitutes the `hostRequestQuota` value, which is always >= MIN_HOST_QUOTA. No way left to seed `hostQuotaRemaining` with a negative.
2. The unlocked dialog now runs 0..DEFAULT_HOST_REQUEST_LIMIT with 0 labelled unlimited/fastest, so the scale is no longer inverted.
3. `onHostRateLimited` installs a concrete allowance for unlocked hosts, dropping them out of the fast path; `resetRequestQuota` walks it back up and erases the entry to re-unlock. The 429 -> full-backlog-replay -> retryRounds=2 -> CDN abandoned cycle is broken.
4. MAX_IN_FLIGHT_PER_HOST bounds the fast-path burst, and `unlockedCapped` keeps the dispatch timer alive so slots are picked up as they free.
5. Each URL row shows its effective limit (raw URL in UserRole, which survives the InternalMove reorder), and `storeSettings` prunes limits for hosts no longer referenced.
6. Limits are stored one plain `host=value` key per host under a `hostRequestLimits/` subgroup, with the legacy @Variant blob removed on write — `downloads.ini` is hand-editable as documented.

Settings plumbing looks right too: the worker's cached `hostRequestLimits` is refreshed via a queued `hostRequestLimitsChanged` connection (the `connect` is after `moveToThread`, context object is the worker, so the reload runs on the picture-loader thread), and `clampHostRequestLimit` only touches the static cap table, so the per-dispatch calls don't hit QSettings. All 18 CI checks are green.

Note: GitHub reports this branch as CONFLICTING with its base (`network-requests/host-backoff`) — needs a rebase before merge, but that's a chore, not a code problem.

Four non-blocking nits left inline (limit pruning ignores redirect-target hosts; the unlocked dialog's max of 10 disagrees with the unbounded `clampHostRequestLimit` and with the 0..50 range in the PR description; the `hostInFlight` decrement is bound to the work object's lifetime so a destruction with a reply pending would leak a slot — I could not find a reachable path, since the work object only self-deletes after its reply has finished; and the in-flight gate keys on the queued host rather than the redirect target). None of them block.

One thing I noticed but did not comment on because it comes from the base branch (310caa7dc), not this PR: in `processSingleRequest`, the rate-limited branch does `requestLoadQueue.removeAt(i)` and then dereferences `request`, which is a reference into the list. Worth a copy-before-erase in the parent PR. The new fast-path loop here gets that ordering right.

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
BruebachL: Fixed in 3bd118fae9: storeSettings now prunes only limits for hosts that are neither referenced by a configured URL nor present in the developer cap table. Both api.scryfall.com and cards.scryfall.io carry a developer cap, so a throttle on the CDN survives deleting its URL from the list (or any later storeSettings() call), and covers the api.scryfall.com -> cards.scryfall.io redirect traffic as intended.

cockatrice/src/interface/widgets/settings_page/deck_editor_settings_page.cpp
BruebachL: Fixed in 3bd118fae9: the unlocked-host dialog now offers 0..UNLOCKED_HOST_LIMIT_MAX (50), matching the range in the PR description and clampHostRequestLimit()'s unbounded ceiling for unlocked hosts (a doc comment on the new constant spells out that relationship). A hand-edited downloads.ini value like 30 is now shown as 30 and kept on OK instead of being silently rewritten to 10; 0 remains labelled "unlimited, fastest" and is the only way back to the default.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed in 3bd118fae9: the in-flight decrement is now its own connection keyed to QObject::destroyed with the worker as the context object, so the slot is released exactly once whenever the reply is destroyed — including an aborted reply or a work object that self-deleted while a reply was still pending. The work object's lifetime now only governs the separate handleNetworkReply handler connection.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed in 3bd118fae9: dispatch resolves the cached-redirect chain first (new resolveCachedRedirect helper with a depth guard) and gates the unlocked fast path on that final host. processSingleRequest applies the same resolution, so the unlocked fast path, the backed-off skip, the allowance arithmetic and the makeRequest in-flight increment all key on the host that is actually hit — and a redirect target that carries its own cap can no longer receive an unbounded burst.

PR #7288, opened 2026-09-12
[Tests] Add manual picture loader benchmark against the real card hosts

BruebachL:

The picture loader's throttling, caching, and backoff logic have no test
coverage beyond unit tests for the settings layer. The pacing and cache logic
tuned in this stack (request pacing, per-host caps, disk cache hits) is
impossible to validate from unit tests alone — it needs real bytes from real
hosts, and there was no harness to exercise it.

- Add a **manual** benchmark/regression executable,
  `picture_loader_benchmark_test`, that links the actual (not mocked) picture
  loader worker code and downloads real card images from the configured hosts.
- Two modes:
  - **default**: loads `#count` pictures from each URL template twice — once
    cold (network) and once cached (served from the
    [QNetworkDiskCache](https://doc.qt.io/qt-6/qnetworkdiskcache.html)). Both
    passes must complete with zero failures; the cached pass asserts it finishes
    *well under* the cold pass (gate for serving cached pictures), and the cold
    pass asserts it stays *above* a pacing lower bound (gate for burst-free
    request throttling).
  - **--stress**: two `CardPictureLoaderWorker` instances hammering one host
    concurrently (~20 req/s aggregate) to force real HTTP 429s, asserting both
    workers still complete 100% via the shared backoff logic.
- Takes `--carddb /path/to/cards.xml` and a URL override; built only when
  `WITH_ORACLE OR WITH_CLIENT` since it links `libcockatrice_settings`.
- **Deliberately not registered with ctest**: it hits live Scryfall/Gatherer
  at ~10 req/s and can take minutes, so it must be run by hand.
  (`tests/picture_loader_benchmark/picture_loader_benchmark.cpp`)

### Why

This PR exists to validate and quantify the rest of the stack: disk-cache
hits, dispatch pacing, and per-host caps/backoff all end up measured here
against the live infrastructure they protect.


tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+#ifdef Q_OS_UNIX
+    // Redirect every QStandardPaths lookup (and therefore SettingsCache paths)
+    // into the sandbox so the benchmark never touches user config or caches.
+    qputenv("XDG_CONFIG_HOME", (rootDir + "/config").toUtf8());
+    qputenv("XDG_DATA_HOME", (rootDir + "/data").toUtf8());
+    qputenv("XDG_CACHE_HOME", (rootDir + "/cache").toUtf8());
+#endif
+
+    const bool warmStart = QDir(rootDir + "/cache/Cockatrice/downloaded")
```
ZeldaZach: The warm-cache probe points at the wrong directory, so `warmStart` is always false.

`getNetworkCachePath()` is `QStandardPaths::CacheLocation + "/downloaded/"`, and Qt appends *both* organizationName and applicationName (both set to `Cockatrice` above), so the real cache lives in `<root>/cache/Cockatrice/Cockatrice/downloaded` — one segment deeper than what is probed here. (Confirmed on disk: `~/Library/Caches/Cockatrice/Cockatrice`.)

Consequence: a `--cache-dir` rerun is never detected as warm, so `coldPaced = warmStart || cold.elapsedMs >= coldLowerMs` applies the pacing lower bound to a pass that is actually served from the disk cache and finishes in a second or two — the run reports FAIL for the exact scenario the flag exists to support. Better to derive the path from `SettingsCache::instance().getNetworkCachePath()` rather than reconstructing it.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+    const QString rootDir = cacheDirArg.isEmpty() ? sandbox.path() : cacheDirArg;
+    QDir().mkpath(rootDir);
+    QDir().mkpath(rootDir + "/config");
+    QDir().mkpath(rootDir + "/data");
+    QDir().mkpath(rootDir + "/cache");
+
+#ifdef Q_OS_UNIX
+    // Redirect every QStandardPaths lookup (and therefore SettingsCache paths)
+    // into the sandbox so the benchmark never touches user config or caches.
+    qputenv("XDG_CONFIG_HOME", (rootDir + "/config").toUtf8());
```
ZeldaZach, line 398: The XDG redirection only sandboxes Linux, but `Q_OS_UNIX` also covers macOS, where `QStandardPaths` ignores `XDG_*` entirely (it uses `~/Library/Application Support/<org>/<app>` and `~/Library/Caches/<org>/<app>`). Windows gets no redirection at all.

Since the benchmark sets organizationName/applicationName to `Cockatrice`, the mock `SettingsCache` then resolves to exactly the same settings directory the real client uses. Two concrete consequences on macOS/Windows:

1. `setDownloadUrls({urlTemplate})` at lines 390/438 writes through to the developer's real `download.ini`, so after a run their installed Cockatrice is left configured with a single picture URL (the last template tested, or the stress URL).
2. The run reads and writes the user's real network disk cache, so the "cold" pass is not cold and the pacing assertion fails spuriously — and the cache gets filled with benchmark downloads.

Either restrict the sandbox claim to Linux and refuse to run elsewhere, or set a unique `applicationName`/`organizationName` (e.g. `Cockatrice-benchmark`) so the paths cannot collide with a real install.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+        const bool redirects = likelyRedirects(urlTemplate);
+        const qint64 perCardMs = redirects ? 200 : 100;
+        const int timeoutMs =
+            (timeoutMin.has_value() ? timeoutMin.value() : (cards.size() * perCardMs * 8 + 60000) / 60000) * 60 * 1000;
+        const qint64 coldLowerMs = static_cast<qint64>(cards.size()) * perCardMs / 2;
+        const qint64 cachedUpperMs = 8000;
+
+        const PassResult cold = runPass(worker, cards, timeoutMs);
+        const PassResult cached = runPass(worker, cards, timeoutMs);
```
ZeldaZach: A cold pass that hits the watchdog poisons the cached pass.

`CardPictureLoaderWorker::handleImageLoadEnqueued()` drops an enqueue when `currentlyLoading` already contains the card's pixmap cache key, and the key is only removed in `handleImageLoaded()`. When `runPass` returns because the watchdog fired, the still-outstanding cards are left in `currentlyLoading` on the shared `worker`. The cached pass then re-enqueues the same list, those cards are silently skipped, `cached.finished` can never reach `cached.enqueued`, and the cached pass is guaranteed to burn the full watchdog too — turning one slow host into a doubly-long run plus a misleading FAIL.

The converse also skews the numbers: a cold-pass request that completes after the cold loop exited is delivered while the cached loop is running and is counted as a cached completion.

Using a fresh `CardPictureLoaderWorker` per pass (or aborting the whole template once the cold pass times out) would avoid both.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+        workerB->enqueueImageLoad(card);
+    }
+    watchdog.start();
+    loop.exec();
+    const qint64 elapsedMs = clock.elapsed();
+    result.a.elapsedMs = elapsedMs;
+    result.b.elapsedMs = elapsedMs;
+
+    result.http429Count = counters.http429;
+    s_activeCounters = nullptr;
```
ZeldaZach: Teardown race on `s_activeCounters` / `counters`.

`counters` is a stack local of `runStress`, but the two workers keep running after `loop.exec()` returns — guaranteed on the watchdog path, where in-flight requests and deferred 429 retries are still pending — and they log from their own threads. A worker thread can pass the `s_activeCounters` null check here, then block acquiring `counters.mutex` while the main thread returns from `runStress` and destroys `counters`, giving a use-after-free on a `QMutex`. `s_activeCounters` is also a plain pointer read/written across threads with no synchronisation.

Separately, `qInstallMessageHandler` returns nullptr when the *default* handler is installed, so `s_previousMessageHandler` is normally null and the `if` at the bottom of `stressLogHandler` swallows all log output for the duration of the stress run — including the 429 warnings the operator would want to see. Calling `qt_message_output`/the default handler explicitly, or just always forwarding, would fix that half.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+    QCoreApplication app(argc, argv);
+    app.setApplicationName(QStringLiteral("Cockatrice"));
+    app.setOrganizationName(QStringLiteral("Cockatrice"));
+    app.setApplicationVersion(QStringLiteral("9.0.0-benchmark"));
+
+    // The ExactCard argument of imageLoaded crosses threads via a queued connection.
+    qRegisterMetaType<ExactCard>();
+
+    QTemporaryDir sandbox;
```
ZeldaZach, line 383: `QTemporaryDir` is never checked for validity, and it auto-removes on scope exit.

If creation fails, `sandbox.path()` is empty, so `rootDir` is empty and the code proceeds to `mkpath("/config")`, `mkpath("/data")`, `mkpath("/cache")` and points `XDG_*` at filesystem-root paths. An `isValid()` check plus a bail-out is cheap here.

Also, when `--cache-dir` is not given the destructor deletes the whole tree at exit, yet line 476 prints `"(reuse with --cache-dir to warm on the next run)"` next to a path that no longer exists by the time the user reads it. Either `setAutoRemove(false)` or reword the hint.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+        }
+        std::printf("RESULT: %s\n", pass ? "PASS" : "FAIL");
+        return pass ? 0 : 1;
+    }
+
+    bool allPass = true;
+    std::printf("=== PICTURE LOADER BENCHMARK (%d cards available, %d per template)%s ===\n", availableCards, count,
+                warmStart ? ", WARM cache from previous run" : "");
+
+    auto *worker = new CardPictureLoaderWorker();
```
ZeldaZach: The workers are leaked and their threads are never stopped, and that defeats the warm-rerun story.

`CardPictureLoaderWorker` persists its redirect cache only from `~CardPictureLoaderWorker()` and from a `QCoreApplication::aboutToQuit` connection. This program never calls `app.exec()`/`quit()`, so `aboutToQuit` is never emitted, and `worker` (and `workerA`/`workerB`) are never deleted — the redirect cache is never written to disk. For `api.scryfall.com`, which is the redirecting host the benchmark specifically special-cases, a `--cache-dir` rerun therefore has to re-resolve every redirect over the network even though the images themselves are cached.

The still-running `QThread`s at `return` also mean network activity can be in flight while `QCoreApplication` is being destroyed. Deleting the workers before returning (and pumping the event loop briefly) would address both.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+        } else if (arg == QLatin1String("--count")) {
+            count = value().toInt();
+        } else if (arg == QLatin1String("--url")) {
+            explicitUrls.append(value());
+        } else if (arg == QLatin1String("--stress")) {
+            stress = true;
+        } else if (arg == QLatin1String("--stress-url")) {
+            stressUrl = value();
+        } else if (arg == QLatin1String("--timeout-min")) {
+            timeoutMin = value().toInt();
```
ZeldaZach: `--timeout-min` is not validated. `value().toInt()` returns 0 both for a missing value (`--timeout-min` as the last argument) and for a non-numeric one, and `timeoutMin` is an `optional`, so 0 is treated as an explicit choice: `timeoutMs` becomes 0, the watchdog `QTimer` fires on the next event loop iteration, every pass records 0 completions and the run reports FAIL for no real reason. `--count` gets a `<= 0` guard a few lines down; the same check is missing here.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+            continue;
+        }
+        SettingsCache::instance().downloads().setDownloadUrls({urlTemplate});
+
+        const bool redirects = likelyRedirects(urlTemplate);
+        const qint64 perCardMs = redirects ? 200 : 100;
+        const int timeoutMs =
+            (timeoutMin.has_value() ? timeoutMin.value() : (cards.size() * perCardMs * 8 + 60000) / 60000) * 60 * 1000;
+        const qint64 coldLowerMs = static_cast<qint64>(cards.size()) * perCardMs / 2;
+        const qint64 cachedUpperMs = 8000;
```
ZeldaZach: The cached-pass budget is a flat 8 s regardless of `--count`, while the cold budget scales with the card count. Decoding and cache-reading 300 images already eats a fair chunk of that on a slow disk, and anything like `--count 2000` will exceed it purely from local work, reporting FAIL with a perfectly healthy cache. Scaling it (e.g. `max(2000, cards.size() * 10)`) or at least comparing against a fraction of `cold.elapsedMs` would hold up across counts.

tests/picture_loader_benchmark/settings_cache_mock.cpp
```diff
@@ -0,0 +1,200 @@
...
+    QSettings pathsIni(getSettingsPath() + "paths.ini", QSettings::IniFormat);
+
+    auto computePath = [&](const QString &key, const QString &defaultPath) -> QString {
+        QString val = pathsIni.value("paths/" + key).toString();
+        if (val.isEmpty() || !QDir(val).exists()) {
+            if (!QDir().mkpath(defaultPath)) {
+                qCInfo(SettingsCacheLog) << "[SettingsCache] Could not create folder:" << defaultPath;
+            }
+            val = defaultPath;
+            pathsIni.setValue("paths/" + key, val);
```
ZeldaZach, line 153: Only four of the ~26 settings pointers declared in `cache_settings.h` are initialised here; the rest (`shortcutsSettings`, `cacheStorageSettings` aside, `commanderBracketSettings`, `networkSettings`, ...) keep indeterminate values. Most accessors are out-of-line and would fail at link time if used, which is the safety net — but `commanderBrackets()` is defined *inline in the header* and returns `*commanderBracketSettings`, so any future code path in the linked loader sources that touches it dereferences garbage instead of failing the build. Zero-initialising the members (or `= {}` on the declarations) makes that a null deref instead of undefined behaviour.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+        if (cached.failed > 0) {
+            std::printf("  cached failures: %s\n", qPrintable(cached.failedCards.join(QStringLiteral(", "))));
+        }
+    }
+
+    std::printf("cache root: %s%s\n", qPrintable(rootDir),
+                cacheDirArg.isEmpty() ? " (reuse with --cache-dir to warm on the next run)" : "");
+    std::printf("RESULT: %s\n", allPass ? "PASS" : "FAIL");
+    return allPass ? 0 : 1;
+}
```
ZeldaZach: Both new .cpp files end without a trailing newline (`\ No newline at end of file` on this file and on `settings_cache_mock.cpp`). The repo runs clang-format in CI, which will flag this.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,479 @@
...
+    const int availableCards = db.getCardList().size();
+
+    const QStringList urlsToTest =
+        explicitUrls.isEmpty() ? SettingsCache::instance().downloads().getAllURLs() : explicitUrls;
+
+    if (stress) {
+        const QList<ExactCard> cards = selectCardsForTemplate(db, stressUrl, count);
+        if (cards.isEmpty()) {
+            std::fprintf(stderr, "error: no cards satisfy template %s\n", qPrintable(stressUrl));
+            return 1;
```
ZeldaZach, line 442: `--url` is silently ignored in `--stress` mode (the templates are collected into `urlsToTest` above, then never used on this branch — only `--stress-url` applies). Someone passing `--stress --url <template>` gets a run against the default Scryfall URL with no warning. Worth either honouring the first `--url` as the stress template or erroring out on the combination.

ZeldaZach (changes requested):
Reviewed the new manual benchmark. Good news on the CI question: the target is not registered with add_test(), so CI will not execute it against the live hosts. It is still compiled whenever WITH_CLIENT or WITH_ORACLE is set, which is fine.

There are several correctness problems in the harness itself, posted as inline comments. The blocking ones:

1. picture_loader_benchmark.cpp:403 probes <root>/cache/Cockatrice/downloaded for a warm start, but SettingsCache::getCachePath() returns QStandardPaths::CacheLocation, which already includes both the organization and the application name. The real path is <root>/cache/Cockatrice/Cockatrice/downloaded, so warmStart is always false and a --cache-dir rerun fails the cold pacing lower bound even when the pass was legitimately served from cache.

2. picture_loader_benchmark.cpp:346 guards the XDG sandbox on Q_OS_UNIX. That covers macOS, which ignores XDG_*, and Windows gets no sandbox at all. Because org and app are both Cockatrice, the mock then resolves to the developer's real settings directory: setDownloadUrls() overwrites their actual download URL list and the cold pass reads their real network cache.

3. picture_loader_benchmark.cpp:448 shares one worker between the two passes. If the cold pass hits the watchdog, the affected cards stay in currentlyLoading, so the cached pass's re-enqueue of those cards is dropped and that pass can never reach the enqueued count, guaranteeing a second watchdog timeout. Late cold completions are also miscounted as cached completions.

4. picture_loader_benchmark.cpp:238 has a teardown race: worker threads keep logging after loop.exec() returns, so a handler can pass the s_activeCounters null check and then lock a mutex in the stack-local counters object that is being destroyed. Separately, qInstallMessageHandler returns nullptr for the default handler, so all log output is swallowed during the stress run.

5. picture_loader_benchmark.cpp:305 does not validate --timeout-min. A missing or non-numeric value parses to 0, and since it is stored as an optional that 0 is honoured, so the watchdog fires immediately and every pass reports FAIL.

The remaining inline comments cover an unchecked QTemporaryDir::isValid() plus auto-remove deleting the directory the usage text tells you to reuse, leaked workers whose destructors never run so the redirect cache is never persisted, a flat 8 s cached-pass budget that does not scale with --count, mostly uninitialised pointers in settings_cache_mock.cpp:153, --url being ignored in --stress mode, and missing trailing newlines that clang-format will flag.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: warmStart is now derived from SettingsCache::instance().getNetworkCachePath(). (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: unique Cockatrice-benchmark app/org names, and the XDG redirection is now guarded by Q_OS_LINUX. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: each pass now gets a fresh worker, so a timed-out cold pass cannot poison the cached pass. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: s_activeCounters is atomic, workers are destroyed before the counters are read, and the handler always forwards via qFormatLogMessage when no previous handler is installed. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: QTemporaryDir::isValid() is checked, and the reuse hint is only printed when --cache-dir was passed. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: workers are now shut down and deleted before returning, so the redirect cache is persisted. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: --timeout-min is validated as a positive integer. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: the cached-pass budget now scales as max(2000, cards.size() * 10) ms. (a7f9b8ff37)

tests/picture_loader_benchmark/settings_cache_mock.cpp
BruebachL: Fixed: SettingsCache pointers are zero-initialised via = nullptr on the declarations. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: trailing newlines added to both files. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
BruebachL: Fixed: the first --url is now used as the stress template when --stress-url was not given. (a7f9b8ff37)

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,547 @@
...
+    QDir().mkpath(rootDir + "/data");
+    QDir().mkpath(rootDir + "/cache");
+
+#ifdef Q_OS_LINUX
+    // Redirect every QStandardPaths lookup (and therefore SettingsCache paths) into the sandbox so
+    // the benchmark never touches user config or caches. XDG_* only affects Qt's path resolution on
+    // Linux; elsewhere the unique application/organization names above keep the run isolated.
+    qputenv("XDG_CONFIG_HOME", (rootDir + "/config").toUtf8());
+    qputenv("XDG_DATA_HOME", (rootDir + "/data").toUtf8());
+    qputenv("XDG_CACHE_HOME", (rootDir + "/cache").toUtf8());
```
ZeldaZach, line 400: The XDG redirection is `#ifdef Q_OS_LINUX`, so on macOS/Windows neither the `QTemporaryDir` sandbox nor `--cache-dir` affects where `SettingsCache`, the `QNetworkDiskCache`, or the copied `cards.xml` actually live — they land in the real `AppLocalDataLocation`/`CacheLocation` and persist across runs.

Concretely on macOS: run the benchmark twice without `--cache-dir`. The first run populates `~/Library/Caches/Cockatrice-benchmark`; on the second the `warmStart` probe sees a non-empty cache, `coldPaced` is short-circuited to `true`, and the “cold” pass is in fact served from disk. Both passes are tiny, every gate this tool exists to enforce is vacuous, and it prints `PASS`. `--stress` likewise never provokes a 429 on a rerun, and the printed `cache root: <temp dir>` is wrong because nothing was ever written there. A full `cards.xml` copy is also left permanently in the user's Application Support directory.

Consider `QStandardPaths::setTestModeEnabled()` plus an explicit app-data override, or refusing to run on non-Linux unless the user opts in.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,547 @@
...
+            std::printf("  worker A failures: %s\n", qPrintable(result.a.failedCards.join(QStringLiteral(", "))));
+        }
+        if (result.b.failed > 0) {
+            std::printf("  worker B failures: %s\n", qPrintable(result.b.failedCards.join(QStringLiteral(", "))));
+        }
+        std::printf("RESULT: %s\n", pass ? "PASS" : "FAIL");
+        return pass ? 0 : 1;
+    }
+
+    bool allPass = true;
```
ZeldaZach, line 480: `allPass` starts `true` and is only ANDed inside the per-template body, but a template with no matching cards `continue`s before any pass runs. So a run in which *every* template is skipped prints `RESULT: PASS` and returns 0 having downloaded nothing.

Easiest way to hit it: pass `--url` as the last argument. The `value()` lambda returns `QString()`, `explicitUrls` becomes `{""}`, nothing matches the empty template, and the harness reports green. Same for a typo'd template — exactly the case a regression harness must fail on. Suggest tracking whether any template actually ran and failing otherwise (and rejecting an empty `--url` value at parse time).

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,547 @@
...
+    result.a.elapsedMs = elapsedMs;
+    result.b.elapsedMs = elapsedMs;
+
+    // Stop both workers before touching the counters or restoring the message handler: their
+    // threads log from stressLogHandler, and must not outlive the stack-local counters (which is
+    // guaranteed on the watchdog path, where requests and deferred retries are still pending).
+    destroyWorker(workerA);
+    destroyWorker(workerB);
+
+    result.http429Count = counters.http429;
```
ZeldaZach, line 267: If `shutdownThread()` times out, `destroyWorker()` deliberately leaves the worker **and its thread running**. But `counters` is a stack local of `runStress`, and `s_activeCounters` is still non-null here: the leaked thread keeps emitting `Too many requests from ...` warnings, so `stressLogHandler` can be locking `counters.mutex` concurrently with this unsynchronized read of `counters.http429`, and after `runStress` returns it can be inside that lock on a destroyed `QMutex` → use-after-free.

The watchdog path is precisely when a stuck worker is likely. Making `LogCounters` a leaked/static object (or clearing `s_activeCounters` before reading, and only reading when both shutdowns succeeded) would close this.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,547 @@
...
+
+    // Copied into the sandbox so the loader's binary cache ("cards.xml.cache")
+    // is written next to it instead of next to the user's file, and so a
+    // --cache-dir rerun can pick it up again.
+    const QString dataPath = SettingsCache::instance().getDataPath();
+    QDir().mkpath(dataPath);
+    const QString cardsXml = dataPath + "/cards.xml";
+    if (!QFile::exists(cardsXml)) {
+        if (!QFile::copy(cardDbArg, cardsXml)) {
+            std::fprintf(stderr, "error: could not copy %s to sandbox\n", qPrintable(cardDbArg));
```
ZeldaZach, line 417: The copy is guarded by `if (!QFile::exists(cardsXml))`, so whenever the sandbox data dir already holds a `cards.xml`, `--carddb` is silently discarded and the previous database is benchmarked instead.

`--cache-dir /tmp/plb --carddb new-cards.xml` after an earlier run with a different file will quietly load the stale copy — and on macOS/Windows that is *every* rerun, since the data dir is the persistent `AppLocalDataLocation`. Worth at least comparing the source against the copy (size/mtime) and re-copying, or warning.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,547 @@
...
+        const PassResult cold = runPass(coldWorker, cards, timeoutMs);
+        destroyWorker(coldWorker);
+
+        auto *cachedWorker = new CardPictureLoaderWorker();
+        const PassResult cached = runPass(cachedWorker, cards, timeoutMs);
+        destroyWorker(cachedWorker);
+
+        const bool coldComplete = cold.finished >= cold.enqueued;
+        const bool coldZeroFailures = cold.failed == 0;
+        const bool coldPaced = warmStart || cold.elapsedMs >= coldLowerMs;
```
ZeldaZach, line 516: For a redirecting template `coldLowerMs` is `cards.size() * 200 / 2` = `cards.size() * 100` ms — exactly `cards.size() * DISPATCH_INTERVAL_MS`.

In a default run over `getAllURLs()`, template 1 (`cards.scryfall.io/large/...`) populates the shared disk cache; template 2 (`api.scryfall.com/cards/!set:uuid!?format=image`) redirects to those same, now-cached image URLs, so its cold pass issues only the paced redirect request per card and lands at ~`cards.size()*100` ms. With a `>=` comparison against exactly that value, the template will flip between PASS and FAIL run to run on scheduling noise.

tests/picture_loader_benchmark/picture_loader_benchmark.cpp
```diff
@@ -0,0 +1,547 @@
...
+        // fail larger --count runs served entirely from a healthy cache.
+        const qint64 cachedUpperMs = qMax<qint64>(2000, static_cast<qint64>(cards.size()) * 10);
+
+        // A fresh worker per pass: if the cold pass hits the watchdog, its outstanding cards stay
+        // in the worker's currentlyLoading set, which would make the cached pass silently skip them
+        // and burn its own watchdog; late cold replies would also be misattributed to the cached
+        // pass.
+        auto *coldWorker = new CardPictureLoaderWorker();
+        const PassResult cold = runPass(coldWorker, cards, timeoutMs);
+        destroyWorker(coldWorker);
```
ZeldaZach, line 508: If `destroyWorker()` cannot stop the cold pass's thread it leaves the worker running — and the very next line constructs a second `CardPictureLoaderWorker` pointed at the *same* `QNetworkDiskCache` directory.

So on the watchdog path (cold pass timed out with requests in flight, thread does not exit within 5 s) two `QNetworkDiskCache` instances write the same directory concurrently, and the leaked worker's late completions keep consuming bandwidth while the cached pass is being timed. `cachedFast` then measures a polluted number. Since the cold pass already failed at that point, it would be safer to abandon the template rather than run the cached pass.

ZeldaZach (approved):
Approving. This is a manual benchmark tool plus a `SettingsCache` mock — nothing here ships to users, and the `WITH_ORACLE OR WITH_CLIENT` gate matches how `libcockatrice_settings` is added, so the build stays sound. The `= nullptr` initializers on the `cache_settings.h` member pointers are a straight improvement.

The inline comments still stand as follow-ups rather than merge blockers. The one I would prioritize is the XDG redirection being Linux-only: on macOS and Windows the sandbox and `--cache-dir` are no-ops, so state persists in the real cache dir, the second run always reports WARM, and the cold-pacing / cached-vs-cold / `--stress` gates all become vacuous while still printing `RESULT: PASS`. A benchmark that cannot fail on two of three platforms will not catch the regressions it was written to catch, so it is worth fixing before anyone relies on its output.

The `allPass` default and the two timing-threshold notes are smaller versions of the same concern.

PR #7292, opened 2026-09-12
[PictureLoader] Fix worker thread shutdown and cross-thread cache clearing

BruebachL:

The picture loader had two thread-safety holes flagged in review of this
stack's earlier PRs:

1. `CardPictureLoader::clearNetworkCache()` called
   `worker->clearNetworkCache()` straight from the **UI thread**, while the
   disk cache (`QNetworkDiskCache`) and the redirect cache (`redirectCache`)
   are owned by the **worker thread**. Clearing them cross-thread raced every
   cache read/write the worker performs.
2. The worker `QThread` was never `quit()`/`wait()`ed. Both destructors only
   called `deleteLater()` on their own objects (`~CardPictureLoaderWorker`
   even posted a `deleteLater` for the main-thread-affine `QThread` from the
   worker thread), so at application exit Qt warned *"QThread: Destroyed
   while thread is still running"* and leaked a running event loop.

- `clearNetworkCache()` becomes a **worker-thread slot**. The UI-facing
  `CardPictureLoader::clearNetworkCache()` now routes it through
  [QMetaObject::invokeMethod](https://doc.qt.io/qt-6/qmetaobject.html#invokeMethod)
  with a blocking queued connection while the worker thread is running (with
  a direct-call fallback otherwise), so the *"Cached card pictures have been
  reset."* message is only shown after the clear actually ran on the worker
  thread.
- The worker now frees itself through the canonical Qt worker-object
  pattern: the constructor wires
  [QThread::finished](https://doc.qt.io/qt-6/qthread.html#finished)
  to its own [QObject::deleteLater](https://doc.qt.io/qt-6/qobject.html#deleteLater),
  so no event loop is left running and deferred deletion runs on the worker's
  own thread.
- A new `shutdownThread()` posts
  [QThread::quit](https://doc.qt.io/qt-6/qthread.html#quit) and blocks on
  [QThread::wait](https://doc.qt.io/qt-6/qthread.html#wait); `CardPictureLoader`
  destroys the `QThread` object only **after** `wait()` has returned, so the
  worker object is already gone and the thread object is no longer executing
  when freed.
- The worker destructor keeps `saveRedirectCache()` (idempotent safety net)
  and no longer touches the thread at all.
  (`cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp`,
  `card_picture_loader_worker.cpp`, `card_picture_loader_worker.h`)


cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
```diff
@@ -295,7 +303,15 @@ void CardPictureLoader::clearPixmapCache()
 
 void CardPictureLoader::clearNetworkCache()
 {
-    getInstance().worker->clearNetworkCache();
+    auto &worker = *getInstance().worker;
```
ZeldaZach: `getInstance().worker` is dereferenced unconditionally, but `~CardPictureLoader` now sets `worker = nullptr` (line 64) as part of shutdown. Any `clearNetworkCache()` call once teardown has begun — e.g. another static destructor or a queued UI action during exit — is a null dereference here, where the old code at least had a live pointer.

The `else` branch is also suspect: the only way `isRunning()` returns false is that the thread already finished, and the `finished() -> deleteLater` chain means the worker object was destroyed at that point — so `worker.isRunning()` and `worker.clearNetworkCache()` would both be use-after-free. The fallback is unreachable when things are healthy and unsafe when they are not. Suggest a null guard on the pointer and dropping the direct-call fallback (or logging and returning instead).

cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
```diff
@@ -295,7 +303,15 @@ void CardPictureLoader::clearPixmapCache()
...
+    // The disk cache and redirect cache are owned by the worker thread; clearing them from the
+    // UI thread would race with the worker's cache reads/writes. Block until the worker thread
+    // has executed the clear so the "Cached card pictures have been reset." message is truthful.
+    if (worker.isRunning()) {
+        QMetaObject::invokeMethod(&worker, "clearNetworkCache", Qt::BlockingQueuedConnection);
```
ZeldaZach: The string-based `invokeMethod` overload is not checked at compile time and its `bool` return is discarded. If `clearNetworkCache` is ever renamed or moved back out of `public slots:`, this silently degrades to a runtime `qWarning` no-op while the caller (`storage_settings_page.cpp:184`, `override_printing_warning.cpp:34`) still shows "Cached card pictures have been reset." — the exact falsehood this PR set out to fix. The pointer-to-member overload gives compile-time checking:

```suggestion
        QMetaObject::invokeMethod(&worker, &CardPictureLoaderWorker::clearNetworkCache,
                                  Qt::BlockingQueuedConnection);
```

ZeldaZach: `Qt::BlockingQueuedConnection` parks the GUI thread until the worker's event loop reaches this call and finishes it. Two things run long on that loop: `handleImageLoadEnqueued()` calls `localLoader->tryLoad()`, which walks the user's picture directories with `QDirIterator` (slow on a large or network-mounted pics folder), and `QNetworkDiskCache::clear()` itself recursively deletes the whole cache directory. So a user hitting "reset cached card pictures" while a batch of cards is being resolved gets a frozen, non-repainting UI for the duration — previously the clear was racy but never blocked. Worth either bounding this (invoke async and show the confirmation from a completion signal) or at least wrapping it in a busy cursor / progress indication.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -83,7 +86,28 @@ CardPictureLoaderWorker::CardPictureLoaderWorker()
...
+
+void CardPictureLoaderWorker::shutdownThread()
+{
+    // The finished() -> deleteLater chain (wired in the constructor) frees this worker as soon as
+    // its event loop exits, so nothing - not even a member read - may run once wait() returns.
+    // QThread::quit() and QThread::wait() are thread-safe and may be called from the owning thread.
+    QThread *thread = pictureLoaderThread;
+    if (thread) {
+        thread->quit();
+        thread->wait();
```
ZeldaZach: `wait()` with no timeout is an unbounded block, and this runs from `~CardPictureLoader`, which is a function-local static destructor — it executes after `main()` returns and after the `QApplication` on the stack is already gone, so there is no window on screen. If the worker thread is stuck in a long synchronous slot (`localLoader->tryLoad()` scanning a slow/network pics directory, or a disk-cache operation on a stalled filesystem), the process hangs at exit with no UI and nothing to interrupt it. A bounded `wait(ms)` with a warning log on timeout would degrade gracefully instead.

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
```diff
@@ -60,6 +60,9 @@ CardPictureLoaderWorker::CardPictureLoaderWorker()
     localLoader = new CardPictureLoaderLocal(this);
 
     pictureLoaderThread = new QThread;
+    // The worker object frees itself once its thread finishes, so no event loop is left
+    // running and the QThread is never destroyed while still executing.
+    connect(pictureLoaderThread, &QThread::finished, this, &QObject::deleteLater);
```
ZeldaZach, line 65: This chain means `~CardPictureLoaderWorker` now actually runs at shutdown (before this PR the thread was never `quit()`ed, so the posted `deleteLater` was never processed and the destructor never executed). That destroys the worker's children — `networkManager`, `cache`, `localLoader` — but the in-flight `CardPictureLoaderWorkerWork` objects are **not** children of the worker: `handleImageLoadEnqueued()` creates them as `new CardPictureLoaderWorkerWork(this, card)` with `QObject(nullptr)` and they only free themselves in `concludeImageLoad()`.

So on every exit with downloads in flight, each pending work object is leaked with the worker already gone: it still holds a `QTimer::singleShot(waitMs, this, ...)` deferred retry armed for up to backoff+5s, and its reply was torn down underneath it by the `QNetworkAccessManager` destructor. Parenting the work objects to the worker (or deleting `requestLoadQueue`'s entries in `~CardPictureLoaderWorker`) would make the new shutdown path actually clean.

ZeldaZach (changes requested):
The overall shutdown design is sound: QThreadPrivate::finish() does flush DeferredDelete after emitting finished(), so the worker really is freed before ~CardPictureLoader deletes the QThread. Two things in the new code need another pass before this goes in, plus three smaller notes left inline.

Blocking:

1. card_picture_loader.cpp:306-313. The new guard in clearNetworkCache() contradicts the lifetime invariant this PR establishes. `auto &worker = *getInstance().worker;` dereferences unconditionally even though ~CardPictureLoader now sets worker to nullptr, and the else branch is unreachable-by-design in the only way that matters: isRunning() can return false only after the thread has finished, at which point the finished() -> deleteLater chain has already destroyed the worker object. So both the isRunning() read and the direct clearNetworkCache() call in that branch operate on a freed object. Guard on the pointer instead of on thread state, and drop the fallback rather than leaving a path that touches a deleted worker.

2. card_picture_loader.cpp:311. Qt::BlockingQueuedConnection parks the GUI thread until the worker's event loop reaches the slot. The worker loop can be inside handleImageLoadEnqueued() running localLoader->tryLoad(), which walks the user's picture directories with QDirIterator, and the slot itself does a recursive QNetworkDiskCache::clear(). Resetting the cache while a batch of cards is resolving will freeze the UI with no repaint, on a thread started at LowPriority. Either invoke it asynchronously and show the confirmation from a completion signal, or at minimum put it behind a busy cursor.

Non-blocking, see inline comments: the string-based QMetaObject::invokeMethod loses compile-time checking and discards its bool result, so a rename silently turns the clear into a no-op while the user still sees "Cached card pictures have been reset."; thread->wait() has no timeout and runs from a function-local static destructor after QApplication is gone, so a stuck worker hangs the process at exit with no UI; and ~CardPictureLoaderWorker now runs for the first time, tearing down networkManager and cache while parentless CardPictureLoaderWorkerWork objects are still alive with deferred-retry timers armed.

Happy to re-review once 1 and 2 are addressed.

cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
BruebachL: Fixed: added a null guard on worker and dropped the unsafe isRunning() fallback. (e960471c69)

cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
BruebachL: Fixed: switched to the pointer-to-member invokeMethod overload for compile-time checking. (e960471c69)

BruebachL: Fixed: the clear is now async; CardPictureLoaderWorker emits networkCacheCleared() and the settings page shows the confirmation from a SingleShotConnection. (e960471c69)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: shutdownThread() now uses a bounded wait (5s) with a warning on timeout. (e960471c69)

cockatrice/src/interface/card_picture_loader/card_picture_loader_worker.cpp
BruebachL: Fixed: CardPictureLoaderWorkerWork objects are now parented to the worker so shutdown actually destroys them. (e960471c69)

cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
```diff
@@ -295,7 +308,17 @@ void CardPictureLoader::clearPixmapCache()
 
 void CardPictureLoader::clearNetworkCache()
 {
-    getInstance().worker->clearNetworkCache();
+    // During teardown the worker is released before this singleton, so a queued clear may still
+    // arrive with no worker left to run it.
+    CardPictureLoaderWorker *worker = getInstance().worker;
+    if (!worker) {
```
ZeldaZach, line 476: Non-blocking, follow-up on my earlier guard comment: the null check landed here, but the other three call sites that go through the singleton still dereference `worker` unconditionally — `getInstance().worker->enqueueImageLoad(card)` at lines 162, 179 and 339. Those sit in the same teardown window this guard was added for, so if anything requests a pixmap after `~CardPictureLoader` has nulled the pointer they crash where this one now returns cleanly.

In practice the picture-load paths are only driven by widgets that are gone by then, so I do not think it is reachable today and it is not worth holding the PR for. Worth a small private helper that returns the worker or nullptr, so the invariant lives in one place rather than being re-derived at each call site.

cockatrice/src/interface/widgets/settings_page/storage_settings_page.cpp
```diff
@@ -181,9 +181,14 @@ StorageSettingsPage::StorageSettingsPage()
 
 void StorageSettingsPage::clearDownloadedPicsButtonClicked()
 {
-    CardPictureLoader::clearNetworkCache();
+    // The network cache is cleared asynchronously on the worker thread, so wait for the completion
+    // signal before confirming; the in-memory pixmap cache is cleared synchronously right away.
+    connect(
```
ZeldaZach, line 186: Non-blocking. Going async fixed the UI freeze, thanks. Two small consequences of arming the confirmation before requesting the clear:

`networkCacheCleared()` is a broadcast, not a reply to this particular request. Two other call sites clear the same cache — `general_settings_page.cpp:454` and `override_printing_warning.cpp:34` — so if one of them is in flight when the user hits this button, the emit it produces satisfies this single-shot connection and the success box appears before this page`"`"`s own clear has run. Narrow window, and the clear does happen a moment later, so the message is early rather than wrong.

The other direction: `CardPictureLoader::clearNetworkCache()` returns early when `worker` is null, which leaves this connection armed with no confirmation ever shown. Only reachable during teardown, when nobody is looking at the dialog anyway.

Also worth a thought: `clearPixmapCache()` on line 190 now runs before the disk clear instead of after it, so a card re-requested in between repopulates the pixmap cache from the disk cache that is about to be wiped. Self-correcting on the next request, just slightly odd ordering.

ZeldaZach (approved):
Re-reviewed against e960471. All five points from the previous round are addressed, and the shutdown design holds up on a second pass.

- The null guard is on the pointer now and the `isRunning()` fallback that would have touched a freed worker is gone.
- `invokeMethod` uses the pointer-to-member overload, so a rename of the slot is a compile error rather than a silent runtime no-op.
- The clear is async and the confirmation comes from `networkCacheCleared()`, so resetting the cache no longer parks the GUI thread behind a `QDirIterator` walk or a recursive `QNetworkDiskCache::clear()`.
- `wait()` is bounded at 5s with a warning, and the `QThread` is only deleted when it returned true. Capturing the thread pointer before `shutdownThread()` is the right call given the worker frees itself during the wait, and the comment explaining why is appreciated.
- Work objects are parented to the worker, so teardown actually reaps them.

Things I checked beyond the earlier feedback, all fine: the work objects are constructed in `handleImageLoadEnqueued()`, which runs on the worker thread, so the new parenting is same-thread and legal; `requestLoadQueue` is a value member and is destroyed before `~QObject` deletes the children, so the new mass-destruction path cannot dangle through it; `saveRedirectCache()` reads only the cached `cacheFilePath` and never touches `SettingsCache`, so it is safe from a function-local static destructor; and `Qt::SingleShotConnection` is Qt 6.0+ against a 6.4.2 floor. No remaining deadlock or use-after-free path that I can find, and CI is green on every platform.

Two non-blocking notes left inline: the null guard is only on `clearNetworkCache()` while the `enqueueImageLoad()` call sites still dereference the singleton worker unconditionally, and `networkCacheCleared()` is a broadcast rather than a reply, so a concurrent clear from one of the other two call sites can fire the settings page single-shot early. Neither is worth another round.

cockatrice/src/interface/card_picture_loader/card_picture_loader.cpp
BruebachL: Addressed in cde8a2dc41. The lifetime invariant now lives in a single private `workerOrNull()` helper, and every call site — `clearNetworkCache()` plus the four `enqueueImageLoad()` paths (`getPixmap`, `cacheCardPixmaps`, `installPrintingOverrideOnLoad`) — guards on it, so anything requesting a pixmap after teardown has nulled the worker returns cleanly instead of dereferencing the dead pointer.

cockatrice/src/interface/widgets/settings_page/storage_settings_page.cpp
BruebachL: Addressed in cde8a2dc41. `clearNetworkCache()` now returns a unique request token (0 when no worker is available, so the settings page skips arming the confirmation entirely — no more connection armed with no confirmation ever shown during teardown). The worker runs clears serially in request order, and the singleton maps each completion back to the oldest outstanding token and re-emits it, so `networkCacheCleared(quint64)` is a true reply rather than a broadcast: a concurrent clear from another page can no longer fire this pages success box early.

The ordering note is covered too: the in-memory pixmap cache purges again in the token-matched completion handler, after the disk clear has actually run, so a card re-requested in between cannot repopulate the pixmap cache from the disk cache about to be wiped.

Client + oracle debug build is green (WARNING_AS_ERROR on).

PR #7294, opened 2026-09-13
[Client] Show localized card names, texts and pictures

### LINKED ISSUE #4317: Reference Point MultiverseID from different languages
Scryfall has the same quality as the Gatherer website for spanish (with bicubic upscaling), and sometimes it doesn't even have the cards in spanish even tho the Gatherer website has it, as seen in #4304.

That's why I thought it would be awesome to use reference point for [Custom Picture Download URLs](https://github.com/Cockatrice/Cockatrice/wiki/Custom-Picture-Download-URLs) for the MultiverseID for specific languages.

I know the data for the multiverseid for different languages is already in mtgjson so it should be technically possible, right?
image

### LINKED ISSUE #5044: Is it possible to change the language of cards' names/texts?
I'm Italian, and I was wondering if it is possible to change in Cockatrice the language for cards (title and all texts); i saw that in MTGJSON are available different languages, just like in the API of scryfall.
But, how can i set a custom cards database in another language? I searched online but found nothing.
Sorry in advance if it was already asked before.

Thanks for your answers!

BruebachL:
## Related Ticket(s)
- Fixes #5044
- Possibly fixes #4317

Localization wiring now runs end to end: the oracle importer collects foreignData for the configured language and the client renders it.

- [Oracle] Import localized names and rules texts for the selected cardLang
  - single-face cards store their foreignData name and full text
  - multi-face (split/adventure/aftermath/prepare) cards collect the joined name once and join each face's translated text with the same separator as the English merge; an incomplete translation falls back to English
  - the wizard switching languages re-imports the card database

- [Client] Display localized card info throughout the client
  - card info text/picture widgets and the game board re-render on language change
  - pictures resolve cardLang art through Scryfall's named endpoint using the localized name, falling back to id-based art when no match exists
  - deck editor keeps canonical English names as card identity (EditRole) while showing localized names (DisplayRole), so decks and wire names stay stable

- [Card] Add CardLocalization-backed name/text lookup and cards.xml v4 localization elements with a bounded-size translation cache

- [Tests] Cover oracle foreignData import (incl. multi-face joins, priority and fallback paths), XML v4 localization parsing, deck model localized display and the language-aware settings default


libcockatrice_card/libcockatrice/card/database/parser/cockatrice_xml_4.cpp
```diff
@@ -401,6 +441,12 @@ void CockatriceXml4Parser::loadCardsFromXml(QXmlStreamReader &xml)
                                                  .upsideDownArt = upsideDown};
             CardInfoPtr newCard = CardInfo::newInstance(name, text, isToken, properties, relatedCards,
                                                         reverseRelatedCards, _sets, attributes);
+            for (auto it = localizedNames.constBegin(); it != localizedNames.constEnd(); ++it) {
+                newCard->setLocalizedName(it.key(), it.value());
```
RickyRister, line 445: why are we not setting it in the constructor

cockatrice/src/interface/card_localization.h
```diff
@@ -0,0 +1,59 @@
+#ifndef COCKATRICE_CARD_LOCALIZATION_H
+#define COCKATRICE_CARD_LOCALIZATION_H
+
+#include "../client/settings/cache_settings.h"
+
+#include <QString>
+#include <libcockatrice/card/card_info.h>
+#include <libcockatrice/settings/cards_display_settings.h>
+
+namespace LocalizedCard
```
RickyRister, line 10: File name doesn't match namespace.
`LocalizedCard` feels more like a class or struct name than a namespace name anyways.

cockatrice/src/interface/widgets/cards/card_info_text_widget.h
```diff
@@ -24,6 +24,7 @@ class CardInfoTextWidget : public QFrame
     QScrollArea *propsScroll;
     QTextEdit *textLabel;
     CardInfoPtr info;
```
RickyRister, line 26: Huh, I didn't realize that this `info` field is completely unused.

cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
```diff
@@ -94,7 +94,20 @@ void CardPictureToLoad::populateSetUrls()
...
+    QStringList orderedTemplates = urlTemplates;
+    if (SettingsCache::instance().cardsDisplay().getCardLang() != "en") {
+        // Scryfall serves localized art from per-language printings, which have
+        // their own ids. The ids stored in the card database belong to the
+        // English prints, so id-based templates (cards.scryfall.io,
+        // api.scryfall.com/cards/!set:uuid!) can only ever resolve English
+        // scans; the `lang=` parameter is ignored on them. Resolve the localized
+        // printing by its translated name (and the language code) ahead of the
+        // id-based templates instead.
+        orderedTemplates.prepend("https://api.scryfall.com/cards/named?fuzzy=!localizedName!&lang=!sflang!"
```
RickyRister, line 106: Are you sure we want to implicitly prepend a url to the list, with no indication and no way to turn it off? This is going to lead to unexpected behavior for the user if they're using their own card urls (if they're playing with custom cards or a different game, for example).

Best case, this eats up a download quota on every single card, as the request to scryfall fails.
Worst case, there's a name overlap and the canon magic card gets downloaded over the expected image.

I'd rather the download url explicitly get added to the list. Perhaps we can prompt the user to add it upon language change, or pop up a confirmation window that automatically adds it. Or have the reset download urls contain it.

tooomm:
Where are these translations coming from?
image

Compare this with how we do translations for the client UI:
image
It's more uniform, uses similar casing and always has the English translation available so users can safely chose to run Japanese cards and not picking Korean by mistake for example.

BruebachL:
> Where are these translations coming from? image
> 
> Compare this with how we do translations for the client UI: image It's more uniform, uses similar casing and always has the English translation available so users can safely chose to run Japanese cards and not picking Korean by mistake for example.

It's 
```
QLocale(lang).nativeLanguageName();
```

which I guess is a bit silly with it's capitalization...?

cockatrice/src/interface/card_picture_loader/card_picture_to_load.cpp
```diff
@@ -94,7 +94,8 @@ void CardPictureToLoad::populateSetUrls()
         }
     }
 
-    for (const QString &urlTemplate : urlTemplates) {
+    const QStringList orderedTemplates = urlTemplates;
```
RickyRister: redundant assignment

libcockatrice_card/libcockatrice/card/card_info.h
```diff
@@ -270,6 +290,96 @@ class CardInfo : public QObject
...
+        return localizedTexts;
+    }
+
+    /**
+     * @brief Sets the card name for the given language.
+     *
+     * @param lang Language code.
+     * @param _localizedName The localized card name.
+     */
+    void setLocalizedName(const QString &lang, const QString &_localizedName)
```
RickyRister, line 348: do we still need setters?

RickyRister:
Card search still uses the english name though, right?

RickyRister (approved):

libcockatrice_card/libcockatrice/card/card_localization.h
```diff
@@ -15,6 +15,16 @@
  * imports the selected language's card data) and the client settings UI (which
  * offers the language choice).
  */
+/**
+ * @brief The card languages card search should run against.
+ */
+enum CardSearchLanguage
```
RickyRister: can we use enum class in new code so that we can also keep the entry names shorter

libcockatrice_card/libcockatrice/card/card_info.h
BruebachL: Yup.

BruebachL:
> Card search still uses the english name though, right?

I made a new setting that allows toggling between English, Selected and Both.

RickyRister:
I was intending for you to do that in a follow-up PR because I don't feel like reviewing another slab of functionality right now.

PR #7300, opened 2026-09-17
[DeckList] Remove obsolete Qt5 qHash compatibility shim

BruebachL:

`deck_list.cpp` still carries a Qt5-only `qHash` compatibility shim behind an `#if QT_VERSION < 0x050600` guard. The codebase is Qt6-only, so the guard is never enabled and the shim is dead code.

- Remove the obsolete Qt5 qHash compatibility shim.


tooomm (approved):
Also covered in #6760

BruebachL:
> Also covered in #6760

Very much figured it would be, was just doing focused deck list work

RickyRister (approved):

PR #7301, opened 2026-09-17
[DeckList] Remove dead card XML readElement

BruebachL:

`AbstractDecklistCardNode::readElement` was a no-op that always returned 0, yet it was forced as a pure virtual by `AbstractDecklistNode` and overridden in `InnerDecklistNode`.

- Remove the dead `readElement` implementation and its pure-virtual declaration.
- Drop the now-unneeded `override`.


libcockatrice_deck_list/libcockatrice/deck_list/tree/abstract_deck_list_node.h
RickyRister: may want to update the docs

RickyRister (approved):

PR #7306, opened 2026-09-17
[DeckList] Extract deck metadata XML serialization

BruebachL:

`DeckList` serialized its metadata inline: a ten-branch `readElement` dispatch plus a static `writeMetadata` that duplicated the tree plumbing.

- Move the metadata XML arms into `DeckListMetadataXml` free functions over `DeckList::Metadata`.
- Leave `DeckList::readElement` as a thin dispatcher between metadata, zones and sideboard plans.
- Move the playmat clamping helper along with the element that uses it.


libcockatrice_deck_list/libcockatrice/deck_list/deck_list_metadata_xml.cpp
```diff
@@ -0,0 +1,107 @@
+#include "deck_list_metadata_xml.h"
+
+#include <QtCore/QXmlStreamReader>
+#include <QtCore/QXmlStreamWriter>
+#include <libcockatrice/utility/playmat_params.h>
+
+namespace DeckListMetadataXml
```
RickyRister: wait, do we namespace inside cpp files?

libcockatrice_deck_list/libcockatrice/deck_list/deck_list_metadata_xml.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+bool readElement(QXmlStreamReader *xml, const QString &childName, DeckList::Metadata &metadata);
+
+/**
+ * @brief Writes the deck metadata section of a Cockatrice deck XML file.
+ *
+ * @param xml Writer to append the metadata elements to.
+ * @param metadata The metadata to serialize.
+ */
+void write(QXmlStreamWriter *xml, const DeckList::Metadata &metadata);
```
RickyRister: honestly, these functions would probably fit the best as instance methods on `Metadata`. But since `Metadata` is a nested class, that means we can't do that while also moving it out of the file.

RickyRister (approved):

libcockatrice_deck_list/libcockatrice/deck_list/deck_list_metadata_xml.cpp
```diff
@@ -0,0 +1,105 @@
+#include "deck_list.h"
```
RickyRister: Are you sure we want to split implementations across multiple cpp files?
Like, I don't see why we can't just have these methods in `deck_list.cpp`, alongside the other `DeckList::Metadata` methods. That file isn't *that* overloaded right now.

PR #7309, opened 2026-09-17
[Client] Exit deterministically when launching the update installer

BruebachL:

`DlgUpdate::downloadSuccessful` launched the update installer and asked the main window to close asynchronously (`Qt::QueuedConnection`), so the installer could race a still-running process holding locks on its Qt DLLs — and if the close was suppressed (e.g. a pending prompt), the app never exited at all while files were being replaced.

- Close the main window synchronously in `downloadSuccessful`, so settings are saved and file locks are released before the launched installer starts replacing files.
- Quit the application for real afterwards, even if the window close was suppressed, making shutdown deterministic while the installer waits.

## Why?

Companion to #7308: the installer now detects running instances and waits (force-closing only after a timeout), which is exactly why an eager client exit is safe.


cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
```diff
@@ -240,8 +242,14 @@ void DlgUpdate::downloadSuccessful(const QUrl &filepath)
...
     if (process.startDetached()) {
-        QMetaObject::invokeMethod(static_cast<MainWindow *>(parent()), "close", Qt::QueuedConnection);
         qCInfo(DlgUpdateLog) << "Opened downloaded update file successfully - closing Cockatrice";
+        // Close the main window synchronously so settings are saved and file locks are released
+        // before the NSIS installer (already launched) starts replacing files, then quit the
+        // application for real in case the close was suppressed (e.g. by a pending prompt).
+        if (auto *window = qobject_cast<MainWindow *>(parent())) {
+            window->close();
+        }
+        QTimer::singleShot(0, qApp, &QCoreApplication::quit);
```
ZeldaZach: The return value of `window->close()` is discarded, so the unconditional `quit()` overrides a close the user explicitly refused.

`MainWindow::closeEvent` calls `event->ignore()` in three cases: a card database update is still running and the user answers **No** to "Quitting now will cancel it. Are you sure?"; `TabSupervisor::closeEvent` finds open games and the user answers **No**; or `AbstractTabDeckEditor::confirmClose()` returns false when the user cancels the save-unsaved-deck prompt. In each of those `close()` returns false, and the old queued `close` left the app running. Now the `singleShot(0, quit)` fires right afterwards and tears the process down anyway: the in-flight card DB update is killed mid-write (the freshly installed client then starts on a possibly half-written database), open games are dropped, and unsaved deck edits are discarded immediately after the user clicked Cancel.

If the eager exit is required, it should be gated on the close actually succeeding, with a warning (or a forced save) when it is not, rather than silently ignoring the user's answer.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
```diff
@@ -240,8 +242,14 @@ void DlgUpdate::downloadSuccessful(const QUrl &filepath)
 
     // Try to open the installer. If it opens, quit Cockatrice
     if (process.startDetached()) {
-        QMetaObject::invokeMethod(static_cast<MainWindow *>(parent()), "close", Qt::QueuedConnection);
         qCInfo(DlgUpdateLog) << "Opened downloaded update file successfully - closing Cockatrice";
+        // Close the main window synchronously so settings are saved and file locks are released
+        // before the NSIS installer (already launched) starts replacing files, then quit the
+        // application for real in case the close was suppressed (e.g. by a pending prompt).
+        if (auto *window = qobject_cast<MainWindow *>(parent())) {
+            window->close();
```
ZeldaZach: The comment says the synchronous close is there "so settings are saved and file locks are released before the installer starts replacing files", but that is exactly what does not happen on the path the fallback `quit()` exists for.

`MainWindow::closeEvent` saves geometry (`SettingsCache::instance().layouts().setMainWindowGeometry(saveGeometry())`) and runs `tabSupervisor->deleteLater()` only after `event->accept()`. When the close is vetoed it returns early, resets `bClosingDown`, and skips both. The subsequent `quit()` then exits without the window geometry ever being written and without the tab supervisor having shut down its connections/tabs, so the "settings are saved" guarantee only holds in the case where the fallback was not needed.

ZeldaZach (changes requested):
Mechanism and ordering look right: the installer is started detached before anything shuts down, the failure branch still tells the user where the download is, the `qobject_cast` is safer than the old `static_cast`, and `QCoreApplication::quit()` does unwind the nested `DlgUpdate::exec()` loop so `MainWindow` (a stack object in `main()`) is still destroyed normally — no skipped destructors or unflushed QSettings on the happy path.

One blocker, left inline: `window->close()`'s return value is discarded and `QTimer::singleShot(0, qApp, quit)` fires regardless. `MainWindow::closeEvent` vetoes the close in three cases — a running card DB update the user declines to cancel, open games the user declines to abandon, and `AbstractTabDeckEditor::confirmClose()` when the user hits Cancel on the save-unsaved-deck prompt. In all three the app now exits ~0 ms after the user said "don't quit": unsaved deck edits are discarded, and a card DB update is killed mid-write right before a new client is installed over it. Gating the `quit()` on `close()` returning true (and warning the user that the installer is waiting when it does not) should be enough.

The second inline comment is non-blocking: the new comment claims the synchronous close guarantees settings are saved, but geometry and `tabSupervisor` shutdown happen only after `event->accept()`, so that guarantee does not hold on exactly the vetoed path the fallback exists for.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
BruebachL: Fixed in 57cf0d94. `quit()` is now scheduled only when `window->close()` returns true. On a vetoed close (running card DB update, open games, or unsaved deck) the app stays running and shows a warning that the update installer is already running and waiting, instead of exiting over the user's answer. If `parent()` is ever not a `MainWindow`, the deterministic-quit behavior from before is still preserved.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
BruebachL: Fixed in 57cf0d94. The comment now only claims settings are flushed and tabs shut down when the close is actually accepted, and documents that a vetoed close keeps the app running so the user can resolve the blocker.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
```diff
@@ -240,8 +242,23 @@ void DlgUpdate::downloadSuccessful(const QUrl &filepath)
...
     if (process.startDetached()) {
-        QMetaObject::invokeMethod(static_cast<MainWindow *>(parent()), "close", Qt::QueuedConnection);
         qCInfo(DlgUpdateLog) << "Opened downloaded update file successfully - closing Cockatrice";
+        // Close the main window synchronously so file locks are released before the NSIS installer
+        // (already launched) starts replacing files. This also flushes settings and shuts down the
+        // tabs, but only when the close is actually accepted: MainWindow may veto it for a running
+        // card DB update, an open game, or an unsaved deck. In that case keep running so the user
+        // can resolve the blocker, and tell them the installer is already waiting.
+        if (auto *window = qobject_cast<MainWindow *>(parent())) {
+            if (window->close()) {
```
ZeldaZach: `window->close()` returning true is not a reliable signal that the shutdown actually ran. `MainWindow::closeEvent` starts with a static re-entrancy guard:

```cpp
static bool bClosingDown = false;
if (bClosingDown) { return; }
```

That early return leaves the `QCloseEvent` in its default accepted state, so `close()` returns true even though no shutdown work happened.

Scenario: the user clicks the X on the main window while the update download is still in flight. `closeEvent` sets `bClosingDown = true` and blocks in `tabSupervisor->close()` on the "There are still open games" / unsaved-deck prompt. That nested modal loop dispatches the download-finished signal, `downloadSuccessful` runs, `window->close()` hits the guard and returns true, and we post `quit()`. The process dies with the prompt still unanswered: `saveGeometry()` / `setMainWindowGeometry()` never ran and the deck editors never got their `confirmClose()`, so window layout and unsaved decks are lost.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
```diff
@@ -240,8 +242,23 @@ void DlgUpdate::downloadSuccessful(const QUrl &filepath)
...
+                QTimer::singleShot(0, qApp, &QCoreApplication::quit);
+            } else {
+                QMessageBox::warning(this, tr("Update"),
```
ZeldaZach, line 258: The veto path tells the user Cockatrice "stays open for now", but the installer launched a few lines above only waits ~60 s (per #7308) before force-closing the process.

Scenario: the user has an unsaved deck, so `TabSupervisor::closeEvent` -> `confirmClose()` gets Cancel and `window->close()` returns false. The user dismisses this warning and goes back to the deck. A minute later the silent `/R` installer's graceful-close wait times out and kills cockatrice.exe, losing the unsaved deck and the settings flush that `closeEvent` would have done.

The message should say the app will be terminated shortly (and roughly when), or the veto should abandon the update instead of leaving a running installer armed against a live process.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
```diff
@@ -240,8 +242,23 @@ void DlgUpdate::downloadSuccessful(const QUrl &filepath)
...
-        QMetaObject::invokeMethod(static_cast<MainWindow *>(parent()), "close", Qt::QueuedConnection);
         qCInfo(DlgUpdateLog) << "Opened downloaded update file successfully - closing Cockatrice";
+        // Close the main window synchronously so file locks are released before the NSIS installer
+        // (already launched) starts replacing files. This also flushes settings and shuts down the
+        // tabs, but only when the close is actually accepted: MainWindow may veto it for a running
+        // card DB update, an open game, or an unsaved deck. In that case keep running so the user
+        // can resolve the blocker, and tell them the installer is already waiting.
+        if (auto *window = qobject_cast<MainWindow *>(parent())) {
+            if (window->close()) {
+                QTimer::singleShot(0, qApp, &QCoreApplication::quit);
```
ZeldaZach: Minor, but relevant to the "quit for real" goal: in Qt 6 `QCoreApplication::quit()` only *asks* the application to quit - it closes the remaining top-level windows and the quit is cancelled if one of them ignores its close event. `QCoreApplication::exit(0)` is the variant that cannot be interrupted.

It matters mainly in the fallback branch below (line 260): there the main window was never closed, so `quit()` sends it a close event and `MainWindow::closeEvent` can ignore it for a running card DB update or an open game - leaving the app alive with its Qt DLLs locked while the installer replaces them, which is the exact state this change is meant to prevent.

```suggestion
                QTimer::singleShot(0, qApp, [] { QCoreApplication::exit(0); });
```

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
BruebachL: Fixed in 00dcc0d5c1. `window->close()` no longer decides the exit: the caller now uses the new `MainWindow::closeForUpdate()`, which returns false both when the close is vetoed and when a close is already in progress (the `bClosingDown` re-entrancy guard, now a member), so `exit()` is only ever scheduled for a shutdown that actually ran. In the re-entrant case the process stays alive until the pending prompt is answered and the shutdown completes.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
BruebachL: Fixed in e1f1cf2e4a. The veto message now states the real deadline instead of "stays open for now": the installer (per #7308) terminates Cockatrice within the next minute, so the user is told to save their work and close the app themselves before the installer force-closes it.

cockatrice/src/interface/widgets/dialogs/dlg_update.cpp
BruebachL: Fixed in c3bc23a544. Both the accepted-close path and the no-MainWindow fallback now use `QCoreApplication::exit(0)` via the suggested singleShot lambda instead of `quit()`, so nothing can keep the process alive by ignoring a close event while the installer replaces the DLLs.

PR #7311, opened 2026-09-18
[PictureLoader] Add local override storage and resolution with matcher tests

BruebachL:
The client can show a card's default art, but there is no way to persist a user-chosen custom image or resolve it back from disk. The local-pics matcher also treats any file whose name starts with a card name as that card, so `Card (1).png` is mistaken for `Card`.

- Add overwrite control to `saveCardImageToLocalStorage` and local-override helpers (`hasLocalOverrides`, `deleteAllLocalOverrides`).
- Add the override-printing install/resolve flow that persists the chosen printing into the local pics pool.
- Match local pics by exact `completeBaseName()` instead of prefix, and fall back to the root `downloadedPics/` scheme filenames.
- Add `tests/loader_local_matching_test.cpp` covering exact-vs-suffixed resolution, set-folder precedence, and the root fallback.

## Why?
The next PR exposes these APIs in the printing-selector card-overlay context menu.


cockatrice/src/interface/card_picture_loader/card_picture_loader.h
```diff
@@ -120,7 +127,10 @@ public slots:
      * @param image Loaded QImage.
      */
     void imageLoaded(const ExactCard &card, const QImage &image);
-    void saveCardImageToLocalStorage(const ExactCard &card, const QPixmap &pixmap);
+    void deleteAllLocalOverrides(const ExactCard &card);
```
RickyRister: why is this not a static method?

cockatrice/src/interface/card_picture_loader/card_picture_loader.h
```diff
@@ -102,6 +102,13 @@ class CardPictureLoader : public QObject
      */
     static bool hasCustomArt();
```
RickyRister: is this used for anything anymore?

cockatrice/src/interface/card_picture_loader/card_picture_loader.h
```diff
@@ -120,7 +127,10 @@ public slots:
      * @param image Loaded QImage.
      */
     void imageLoaded(const ExactCard &card, const QImage &image);
-    void saveCardImageToLocalStorage(const ExactCard &card, const QPixmap &pixmap);
+    void deleteAllLocalOverrides(const ExactCard &card);
+    void saveCardImageToLocalStorage(const ExactCard &card, const QPixmap &pixmap, bool allowOverwrite = false);
+    void overridePrintingConnectLocalSaveAndEnqueue(const ExactCard &originalCard, const ExactCard &overrideCard);
+    void overridePrintingEnsurePixmapExistsAndSaveLocally(const ExactCard &originalCard, const ExactCard &overrideCard);
```
RickyRister: nit: surely there exists easier to parse names than this.

cockatrice/src/interface/card_picture_loader/card_picture_loader.h
```diff
@@ -120,7 +127,9 @@ public slots:
      * @param image Loaded QImage.
      */
     void imageLoaded(const ExactCard &card, const QImage &image);
-    void saveCardImageToLocalStorage(const ExactCard &card, const QPixmap &pixmap);
+    void saveCardImageToLocalStorage(const ExactCard &card, const QPixmap &pixmap, bool allowOverwrite = false);
```
RickyRister, line 130: actually, on second thought, can't all these methods also be static?
Though the code is just calling `getInstance` and then this method, so it doesn't make much of a difference either way.

RickyRister (approved):

cockatrice/src/interface/card_picture_loader/card_picture_loader.h
BruebachL: It's largely irrelevant for two call sites and slightly more awkward for the remaining two.

PR #7312, opened 2026-09-18
[PrintingSelector] Add Image Overrides submenu with hover preview

BruebachL:
The printing-selector card overlay only offers pinning, so there is no in-place way to load a custom image for a card or switch to one of its other printings. A failed image load was also silent.

- Add an Image Overrides submenu with `Load Custom Image...`, `Clear Custom Image`, and one entry per alternate printing.
- Show a hover preview of the highlighted printing, sized in device-independent pixels and clamped so it stays on screen at any DPI.
- Surface unreadable image files with a warning instead of a silent no-op.
- Parent the overlay submenus to the stack menu, fixing a pre-existing menu leak.


cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.h
```diff
@@ -7,6 +7,9 @@
 #ifndef PRINTING_SELECTOR_CARD_OVERLAY_WIDGET_H
 #define PRINTING_SELECTOR_CARD_OVERLAY_WIDGET_H
 
+class QAction;
```
RickyRister: put forward declarations after includes

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
```diff
@@ -291,3 +412,140 @@ void PrintingSelectorCardOverlayWidget::initializePinBadge()
...
+        return;
+    }
+
+    const QSize previewSize(240, 336);
+    const int previewOffset = 20;
+
+    QPixmap pixmap;
+    CardPictureLoader::getPixmap(pixmap, hoveredOverrideCard, previewSize);
+
+    QPixmap previewPixmap = pixmap;
```
RickyRister: why the copy here?

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
```diff
@@ -291,3 +412,140 @@ void PrintingSelectorCardOverlayWidget::initializePinBadge()
...
+    int y;
+    if (belowFits) {
+        y = actionRect.bottom() + previewOffset;
+    } else if (aboveFits) {
+        y = actionRect.top() - previewOffset - labelSize.height();
+    } else {
+        y = actionRect.top();
+    }
+    y = qMax(screenGeometry.top(), y);
+    y = qMin(screenGeometry.bottom() - labelSize.height() + 1, y);
```
RickyRister: nit: is it possible to extract the x/y calculation to a function that returns a `QPoint`?

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
```diff
@@ -291,3 +412,140 @@ void PrintingSelectorCardOverlayWidget::initializePinBadge()
...
+ * first-ever show at the screen's edges stays fully on-screen.
+ */
+void PrintingSelectorCardOverlayWidget::refreshPreview()
+{
+    if (hoveredOverrideCard.isEmpty()) {
+        hidePreview();
+        return;
+    }
+
+    const QSize previewSize(240, 336);
```
RickyRister: clion is saying that these can be constexpr

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
```diff
@@ -219,8 +254,90 @@ void PrintingSelectorCardOverlayWidget::customMenu(QPoint point)
...
+
+    auto *loadCustomAction = overrideMenu->addAction(tr("Load Custom Image..."));
+    auto *clearOverrideAction = overrideMenu->addAction(tr("Clear Custom Image"));
+
+    // Nothing to clear on a card that has no local override yet.
+    clearOverrideAction->setEnabled(CardPictureLoader::hasLocalOverrides(rootCard));
+
+    overrideMenu->addSeparator();
+
+    const auto &allSets = rootCard.getInfo().getSets();
```
RickyRister, line 312: What's the point of adding a way to replace a printing of a card with another to the UI?
I feel this is going to lead to people accidentally overriding a printing and then complaining about not being able to find that printing.

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
```diff
@@ -219,8 +254,90 @@ void PrintingSelectorCardOverlayWidget::customMenu(QPoint point)
...
+
+            // The submenu is already scoped to this card, so the rows lead with set +
+            // collector; only printings with a distinct display name add their own name.
+            const CardSetPtr cardSet = printing.getSet();
+            if (!cardSet) {
+                continue;
+            }
+
+            QString label = tr("%1 %2").arg(cardSet->getCorrectedShortName(), printing.getProperty("num"));
+            const QString &flavorName = printing.getFlavorName();
```
RickyRister: Wait, do we use the `flavorName` field anywhere else in cockatrice currently? I'm pretty sure this is the first time we've used that field. We added it a while back but never got around to integrating it because we never decided on how to.

I don't think this PR is the correct time to start integrating it. This will make it inconsistent with the rest of Cockatrice which doesn't use flavorNames yet.

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
```diff
@@ -219,8 +254,90 @@ void PrintingSelectorCardOverlayWidget::customMenu(QPoint point)
...
+    }
+
+    connect(clearOverrideAction, &QAction::triggered, this, [this]() {
+        CardPictureLoader::getInstance().deleteAllLocalOverrides(rootCard);
+        QPixmapCache::clear();
+        rootCard.emitPixmapUpdated(); // force UI refresh
+    });
+
+    connect(loadCustomAction, &QAction::triggered, this, [this]() {
+        QString filePath = QFileDialog::getOpenFileName(this, tr("Select Card Image"), QString(),
```
RickyRister: nit: feels like this could be extracted to a method

cockatrice/src/interface/widgets/printing_selector/printing_selector_card_overlay_widget.cpp
BruebachL: This is more granular printing control than set preference or disabling sets. Instead of i.e. wholesale disabling all SLD cards because you *really* hate furby, you can just override that specific printing.

You can tell you've overriden a printing because then the option to clear the override isn't greyed out anymore.

RickyRister (approved):

PR #7314, opened 2026-09-18
[Card] Add a setting for the language used in card search

BruebachL:
## Related Ticket(s)
- Follow-up to #7294

Card search only matched against the English card names and texts, so searching for a localized name (or text) found nothing once the card database was imported in a non-English language.

- Add a `Language used in card search` toggle on the general settings page (English / selected card language / both)
- Match the plain-text, filter-tree and completer searches against the selected card language, falling back to English for untranslated cards
- Thread the search language through `FilterTree`, `FilterString`, `CardDatabaseDisplayModel` and `CardSearchModel`
- Add settings default-value tests for the new `cardSearchLanguage` key
- Drop the redundant local copy of the URL templates list in the localized picture loader


libcockatrice_filters/libcockatrice/filters/filter_string.cpp
```diff
@@ -447,3 +485,9 @@ FilterString::FilterString(const QString &expr)
         result = [](const CardData &) -> bool { return false; };
     }
 }
+
+void FilterString::setSearchLanguage(const QString &searchLanguage, CardSearchLanguage searchLanguageMode)
+{
+    globalSearchLanguage = searchLanguage;
```
RickyRister: I *really* do not like the fact that it's setting a global constant in an instance method. Are you *sure* there isn't a better way to do it?

BruebachL: Better solutions are indeed available.

cockatrice/src/game_graphics/dialogs/dlg_create_token.cpp
```diff
@@ -88,6 +90,15 @@ DlgCreateToken::DlgCreateToken(const QStringList &_predefinedTokens, QWidget *pa
     cardDatabaseDisplayModel = new TokenDisplayModel(this);
     cardDatabaseDisplayModel->setSourceModel(cardDatabaseModel);
 
+    CardsDisplaySettings *cardsDisplay = &SettingsCache::instance().cardsDisplay();
```
RickyRister: nit: Isn't it better to get the instance inside the lamda?
Probably doesn't matter, but technically it's more correct because if the instance changes, then you want to get the instance only when you call the lambda..

libcockatrice_models/libcockatrice/models/database/card/card_search_model.cpp
```diff
@@ -79,11 +85,17 @@ void CardSearchModel::updateSearchResults(const QString &query)
         }
     }
 
-    auto sortByDistanceThenLength = [](const SearchResult &a, const SearchResult &b) {
+    auto sortByDistanceThenLength = [this](const SearchResult &a, const SearchResult &b) {
+        QString nameA = a.card->getName();
+        QString nameB = b.card->getName();
+        if (searchLanguageMode != CardSearchLanguage::English && !searchLanguage.isEmpty() && searchLanguage != "en") {
```
RickyRister: does `CardSearchLanguage::Both` work with this?

libcockatrice_filters/libcockatrice/filters/filter_string.cpp
```diff
@@ -74,6 +74,43 @@ NumericValue <- [0-9]+
 
 static std::once_flag init;
 
+// The peglib parser rules (and therefore their rule actions) are set up once per
+// process, so a rule action cannot capture per-instance state. The card language
+// plain-text name and text queries search in is therefore handed to the GenericQuery
+// and OracleQuery rule actions through this thread-local context, which is live only
+// while a FilterString is being parsed. The rule actions copy it into the filter
+// closures they produce, so card evaluation never reads process-global state.
+struct SearchLanguageContext
```
RickyRister: honestly, we should just define this struct in a header file so that all our functions can accept this struct instead of the two parameters separately.

libcockatrice_card/libcockatrice/card/card_localization.h
```diff
@@ -15,6 +15,16 @@
  * imports the selected language's card data) and the client settings UI (which
  * offers the language choice).
  */
+/**
+ * @brief The card languages card search should run against.
+ */
+enum class CardSearchLanguage
```
RickyRister: nit: param names suggest that you were going to name it `SearchLanguageMode`, which is probably a better name for it

libcockatrice_card/libcockatrice/card/card_localization.h
```diff
@@ -15,6 +15,16 @@
...
+{
+    English,  ///< Only search the English card names and texts.
+    Selected, ///< Search the selected card language (untranslated cards still match in English).
+    Both      ///< Search both the English and the selected card language names and texts.
+};
+
```
RickyRister: wait, did you accidentally displace the doc comments for the namespace?

RickyRister (approved):
I don't have a localized cards.xml so I can't exactly test it, but the code seems fine.

PR #7315, opened 2026-09-18
[Game] Allow judges to enter any game regardless of restrictions

BruebachL:

Judges need to enter any game — password-protected, buddies-only, registered-only, or spectator-disabled — to do their work, but only moderators currently get the restriction override. Judges can't even request the override client-side, since that path is gated on the moderator-only admin tab.

- Grant `IsJudge` users the same entry-restriction override as moderators in `Server_Game::checkJoin` (password, registered-only, buddies-only, ignore list, spectator rules)
- Add `TabSupervisor::canOverrideGameRestrictions()` so judges (and moderators via the unlocked admin tab) can bypass restrictions without opening the moderator admin tab
- Enable the Join/Spectate buttons and skip the password prompt for judges on restricted games
- Add `server_game_join_test` covering judge/moderator bypass of each restriction, fail-closed behavior without the override flag, and that overriding doesn't grant `join_as_judge` to non-judges

## Why?
Judges requested parity with moderators for game entry. Moderation powers (kick/ban/warn) remain moderator-only — `getAdminLocked()` is untouched.


RickyRister (approved):
Looks fine. Not too familiar with the server code so up to you if you want someone else to look at it.

BruebachL:
It's a relatively small change that's mostly additive and hits a small percentage of users. I'd deem this to be fine unless we hear otherwise.

PR #7316, opened 2026-09-18
[Build] Keep Windows installs free of build-tree artifacts

### LINKED ISSUE #7040: Install directory with many unexpected folders
**System Information:**

Client Version: 3.1.0-custom-6662d46 (2026-07-08)
Build Architecture: 64-bit
Qt Version: 6.11.0
System Locale: en_DE
Install Mode: Standard

Cockatrice install directory shows a lot of weird folders (and subfolders) after a fresh uninstall+install:
Image

Image

<br>

Might be only an issue on Windows.

<br><br>

Actually, tried again and looked after uninstall alone... there were many leftovers in the directory still:
Image

However, deleting them and installing again gives me (this was installing `3.0.2 Stable`):
Image

**Steps to reproduce:**
 - Uninstall Cockatrice
 - Install Cockatrice
 - Inspect installation directory

BruebachL:
## Problem

A fresh Windows install in `C:\Program Files\Cockatrice` was found littered with Visual-Studio CMake build-tree directories (`cockatrice.dir`, `cockatrice_autogen`, `.qt`, `.qsb`, `.lupdate`, `x64\Release\*.tlog`, `INSTALL/PACKAGE/RUN_TESTS.dir`, ...). These are MSBuild/Qt AUTOMOC outputs that belong in the build directory, never in an installed app.

## Root cause

Three packaging gaps let build output reach end users:

1. **Fragile DLL sweep.** Every app swept `install(DIRECTORY "${CMAKE_BINARY_DIR}/${PROJECT_NAME}/${CMAKE_BUILD_TYPE}/" ...)`. On multi-config generators (VS/Ninja Multi-Config) `CMAKE_BUILD_TYPE` is empty, so the recursive directory install collapses to the whole `<app>` build tree — the exact spot `*.dir`, `*_autogen`, `.qt`, `.qsb`, `x64/` and `*.tlog` live. It only survived by the `PATTERN "*.dll"` filter.
2. **Broken OpenSSL install.** `install(FILES ${OPENSSL_INCLUDE_DIRS} DESTINATION ./)` tries to install include *directories* as files. CMake rejects this at configure time (`install FILES given directory`); it slipped through CI only because the vcpkg OpenSSL config leaves the variable empty, making the rule a silent no-op. OpenSSL runtime DLLs are already shipped by `fixup_bundle`.
3. **Uninstaller leaves leftovers.** The NSIS uninstaller deleted only `*.exe`/`*.dll` and a handful of known files, then `RMDir "$INSTDIR"` (non-recursive). Any other file — build artifacts included — survived an uninstall/reinstall cycle forever.

## Changes

- Point the per-config DLL sweep at the real output dir with `$<TARGET_FILE_DIR:<app>>` and exclude build artifacts (`*.dir`, `*_autogen`, `.qt`, `.qsb`, `.lupdate`, `x64`, `CMakeFiles`, `*.tlog`, `*.pdb`).
- Remove the bogus `OPENSSL_INCLUDE_DIRS` install rule.
- Make the NSIS uninstaller `RMDir /r "$INSTDIR"` so a fresh install is always clean.
- Add a Windows CI gate that lists the packaged installer with 7-Zip and fails the build if any build-tree artifact path is present.

## Screenshots

Before (Post install):
image

Before (Post Uninstall):
image

After (Post install):
image

Ater (Post Uninstall):

Complete directory gone.
image


tooomm (commented):
Works for me! Uninstallation left no files behind and a new install has a clean directory.

This will close https://github.com/Cockatrice/Cockatrice/issues/7040

Side note, this will also fix the theme selector for Windows users including unrelated options:
- Before:
  image
- After reinstalling with the binary from this PR:
  image

.ci/compile.sh
```diff
@@ -327,4 +327,30 @@ if [[ $MAKE_PACKAGE ]]; then
...
+      echo "Inspecting $package"
+      if "$seven_zip" l "$package" |
+        grep -E "_autogen|\.dir[\\/]|\.tlog|(^|[\\/])x64[\\/]|(^|[\\/])\.qt[\\/]|(^|[\\/])\.qsb[\\/]|(^|[\\/])\.lupdate[\\/]|CMakeFiles"; then
+        echo "::error file=$0::installer contains build-tree artifacts"
+        exit 1
+      fi
+      echo "Installer content is clean."
+    fi
+    echo "::endgroup::"
+  fi
```
tooomm, line 357: Start messages with capital letters:
```suggestion
  if [[ $RUNNER_OS == Windows ]]; then
    echo "::group::Check installer for build-tree artifacts"
    cd "$BUILD_DIR"
    package="$(find . -maxdepth 1 -type f -name 'Cockatrice-*.exe' -print -quit)"
    if [[ ! $package ]]; then
      echo "::error file=$0::Could not find installer to inspect"
      exit 1
    fi
    seven_zip="$(command -v 7z || true)"
    if [[ ! $seven_zip ]]; then
      seven_zip="/c/Program Files/7-Zip/7z.exe"
    fi
    if [[ ! -f $seven_zip ]]; then
      echo "::warning file=$0::7-Zip not found, skipping installer content check"
    else
      echo "Inspecting $package"
      if "$seven_zip" l "$package" |
        grep -E "_autogen|\.dir[\\/]|\.tlog|(^|[\\/])x64[\\/]|(^|[\\/])\.qt[\\/]|(^|[\\/])\.qsb[\\/]|(^|[\\/])\.lupdate[\\/]|CMakeFiles"; then
        echo "::error file=$0::Installer contains build-tree artifacts"
        exit 1
      fi
      echo "Installer content is clean"
    fi
    echo "::endgroup::"
  fi
```

cockatrice/CMakeLists.txt
```diff
@@ -656,17 +656,34 @@ if(WIN32)
...
+    PATTERN "CMakeFiles*" EXCLUDE
+    PATTERN "*.dir*" EXCLUDE
+    PATTERN "*_autogen*" EXCLUDE
+    PATTERN "*.tlog*" EXCLUDE
+    PATTERN "*.pdb" EXCLUDE
+    PATTERN "x64*" EXCLUDE
+    PATTERN ".qt*" EXCLUDE
+    PATTERN ".qsb*" EXCLUDE
+    PATTERN ".lupdate*" EXCLUDE
     PATTERN "*.ini"
```
tooomm: Use same ordering than for other PATTERN lists: Put EXCLUDEs to the end

.ci/compile.sh
```diff
@@ -327,4 +327,30 @@ if [[ $MAKE_PACKAGE ]]; then
...
+    seven_zip="$(command -v 7z || true)"
+    if [[ ! $seven_zip ]]; then
+      seven_zip="/c/Program Files/7-Zip/7z.exe"
+    fi
+    if [[ ! -f $seven_zip ]]; then
+      echo "::warning file=$0::7-Zip not found, skipping installer content check"
+    else
+      echo "Inspecting $package"
+      if "$seven_zip" l "$package" |
+        grep -E "_autogen|\.dir[\\/]|\.tlog|(^|[\\/])x64[\\/]|(^|[\\/])\.qt[\\/]|(^|[\\/])\.qsb[\\/]|(^|[\\/])\.lupdate[\\/]|CMakeFiles"; then
```
tooomm, line 350: Adding a comment here could be helpful

tooomm (approved):

PR #7319, opened 2026-09-18
[DeckList] Extract card parsing from zone XML reader

BruebachL:

`InnerDecklistNode::readElement` mixes zone recursion with inline `<card>` attribute parsing and an unlabelled deck-size budget cap.

- Extract `<card>` reading into a file-local `readCardElement` helper.
- Name the remaining-card budget so the deck-size cap reads explicitly.


RickyRister (approved):

libcockatrice_deck_list/libcockatrice/deck_list/tree/inner_deck_list_node.cpp
```diff
@@ -2,6 +2,30 @@
...
+/**
+ * @brief Reads a single `card` element and appends it to @p parent.
+ *
+ * The card's quantity is capped at @p remainingBudget so a malicious or
+ * oversized deck file cannot push the total card count past the deck size
+ * limit.
+ *
+ * @return The amount of cards actually added.
+ */
+int readCardElement(QXmlStreamReader *xml, InnerDecklistNode *parent, int remainingBudget)
```
RickyRister: nit: If a static helper function needs to take `*this` as an argument, it might as well be a private instance method on the class. I guess it's only here because you didn't want to change the .h file.

PR #7320, opened 2026-09-18
[DeckList] Extract deck-node sort helpers

BruebachL:

`InnerDecklistNode::sort` does three interleaved steps in one method: snapshotting child indexes, sorting, and mapping old to new positions.

- Extract `indexedSnapshot` and `applySortedOrder` helpers so `sort` reads as named steps.


libcockatrice_deck_list/libcockatrice/deck_list/tree/inner_deck_list_node.cpp
```diff
@@ -24,6 +24,36 @@ int readCardElement(QXmlStreamReader *xml, InnerDecklistNode *parent, int remain
     return amount;
 }
 
+/**
+ * @brief Snapshots the current children as (old index, node) pairs.
+ */
+QVector<QPair<int, AbstractDecklistNode *>> indexedSnapshot(const InnerDecklistNode &node)
```
RickyRister: nit: same comment about static helper methods that always take `*this` as a parameter.

RickyRister (approved):

PR #7322, opened 2026-09-18
[DeckList] Extract recursive card traversal helpers

BruebachL:

`getCardNodes` and `forEachCard` define their recursive walks inline as `std::function` lambdas, burying the traversal under boilerplate.

- Extract `collectCardsRecursive` and `forEachCardInNode` file-local helpers.
- Keep the top-level board-zone classification for cards in custom zones.


libcockatrice_deck_list/libcockatrice/deck_list/deck_list_node_tree.cpp
```diff
@@ -46,6 +46,39 @@ QString encodeDeckHash(const QByteArray &digest)
     return QString::number(number, 32).rightJustified(8, '0');
 }
 
+/**
+ * @brief Collects every card node in @p node's subtree, in tree order.
+ */
+void collectCardsRecursive(const InnerDecklistNode *node, QList<const DecklistCardNode *> &result)
```
RickyRister: nit: since we're refactoring anyways, is there a way to make this function return a list instead of using an out-parameter?

RickyRister (approved):

PR #7324, opened 2026-09-18
[DeckList] Extract deck root seeking and body reading in XML load

BruebachL:

`DeckList::loadFromXml` uses a double-nested `while` loop to seek the root element and read its contents, with subtle handling of trailing content.

- Extract `seekToNextElement` and `readDeckBody` private helpers.
- `loadFromXml` now reads: seek to element, require `cockatrice_deck`, read body.


libcockatrice_deck_list/libcockatrice/deck_list/deck_list.h
```diff
@@ -96,7 +96,20 @@ class DeckList
     };
 
 private:
-    Metadata metadata;                           ///< Deck metadata that is stored in the deck file
+    Metadata metadata; ///< Deck metadata that is stored in the deck file
```
RickyRister: did you accidentally stick these methods right in the middle of the fields section?

RickyRister (approved):

PR #7325, opened 2026-09-18
[DeckList] Extract deck metadata element readers

BruebachL:

`Metadata::readElement` is a long `else if` chain where the `bannerCard` and `playmatCard` branches each read several attributes before their text.

- Extract `readBannerCard` and `readPlaymatCard` file-local helpers.
- Keep the attribute-before-text ordering and the playmat parameter clamping.


libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
```diff
@@ -38,6 +38,40 @@ static double parseClampedParam(const QString &valueString, double fallback, dou
     return qBound(min, value, max);
 }
 
+/**
+ * @brief Reads a `bannerCard` element into the deck metadata.
+ */
+static void readBannerCard(DeckList::Metadata &metadata, QXmlStreamReader *xml)
```
RickyRister: feels like these functions would look nicer if they just took a `QXmlStreamReader` and returned the parsed struct, and the assignment happens in `readElement`

libcockatrice_deck_list/libcockatrice/deck_list/deck_list.cpp
```diff
@@ -38,6 +38,48 @@ static double parseClampedParam(const QString &valueString, double fallback, dou
...
+    QString providerId = xml->attributes().value("providerId").toString();
+    QString marginLStr = xml->attributes().value("marginPctL").toString();
+    QString marginRStr = xml->attributes().value("marginPctR").toString();
+    QString vOffStr = xml->attributes().value("verticalOffset").toString();
+    QString zoomStr = xml->attributes().value("zoom").toString();
+    QString cardName = xml->readElementText();
+
+    PlaymatInfo playmat;
+    playmat.card = {cardName, providerId};
+    playmat.params.marginPctL = parseClampedParam(marginLStr, 0.07, 0.0, 0.95);
```
RickyRister: nit: use designated initializer

RickyRister (approved):

PR #7337, opened 2026-09-21
[Card] Add Scryfall Tagger tag search and filtering

BruebachL:

Cards can't be searched or filtered by their Scryfall Tagger tags, and MTGJson doesn't ship tag data yet. This is a proof of concept that wires tags end to end so the client-side work can land ahead of the data source.

- Import a `tags` array from MTGJson into the `tags` card property (lowercased, deduped, sorted, stored space-separated); split-card faces union their tags.
- Add `tags:`/`tag:` search syntax matching a single whole slug, combined with AND (`tags:draw tags:ramp`), `OR`, and `-`.
- Add a `Tag` attribute to the `CardFilter`/`FilterTree` backing the visual filter dock.
- Add a dedicated Tags filter panel to the visual card database (search, count threshold, all/any mode).
- Add an Oracle helper script that injects Scryfall's official Oracle Tags bulk data into an AllPrintings file, standing in for MTGJson until it emits `tags`.

## Follow-up: MTGJson

MTGJson needs to emit a per-card `tags` array so the importer path above is exercised for real. The pieces:

- Add a `tags` property to the Card (Set) model: a list of Scryfall Tagger **slug** strings, resolved by joining each card's `identifiers.scryfallOracleId` against Scryfall's [Oracle Tags bulk data](https://scryfall.com/docs/api/tags) (`oracle_id` -> taggings).
- Use the **oracle** tag type only. Art tags (`illustration_id`) describe artwork and belong at the printing level, not on the oracle card.
- Decide how to handle the tag hierarchy: the bulk file only contains direct taggings, so parent tags (e.g. `animal`) have no taggings of their own and would need their `child_ids` traversed if ancestors should be searchable.
- Slugs are not guaranteed stable; the stable identifier is the tag `id`. For search/display the slug is what the client indexes, so a rename is a data migration, not a client change.
- Weights (`very_strong`..`weak`) and annotations are dropped for now; the POC only needs the slug set.

Once MTGJson ships `tags`, `oracle/scripts/inject_scryfall_tags.py` can be deleted and Oracle's normal download path picks the data up unchanged.

## Screenshots

_TODO: add a screenshot of the Tags filter panel once tag data is available._


RickyRister:
Oh nice, adding support for tags would actually be quite helpful for certain custom constructed formats. For example, msehub could put the card tags into the exported cards.xml

RickyRister:
Can we split this PR into one for adding tag support to cockatrice, and one for updating oracle to import scryfall tags?

I can test tag support using a custom cards.xml.

BruebachL:
There is a script attached to this PR in the oracle folder that allows you to patch a https://mtgjson.com/api/v5/AllPrintings.json.xz with scryfall tag data

```
oracle/scripts/inject_scryfall_tags.py ~/Downloads/AllPrintings.json.xz
```

then point oracle at the newly generated file

cockatrice/src/interface/widgets/visual_database_display/visual_database_display_filter_toolbar_widget.cpp
```diff
@@ -148,13 +154,15 @@ void VisualDatabaseDisplayFilterToolbarWidget::retranslateUi()
     quickFilterSubTypeWidget->setToolTip(tr("Filter by card sub-type"));
     quickFilterSetWidget->setToolTip(tr("Filter by set"));
     quickFilterFormatLegalityWidget->setToolTip(tr("Filter by format legality"));
+    quickFilterTagWidget->setToolTip(tr("Filter by Scryfall Tagger tag"));
```
RickyRister, line 157: Don't mention scryfall in the tooltip. The tags don't have to come from scryfall. You can just say "Filter by tag"

RickyRister:
So tags will take up an entry in the description.

Screenshot 2026-09-22 at 2 11 00 AM

How many tags do cards on scryfall go up to? 

Though I guess even if a card has lots of tags, it doesn't look *that* bad

Screenshot 2026-09-22 at 2 17 48 AM

The card attributes box being scrollable does limit how much damage large amounts of tags can do, even when space is limited while in a game.

Screenshot 2026-09-22 at 2 19 31 AM

